PR #1741, opened 2018-10-04
masks preview initial position

edgardoh:
When an image has distortions applied (including rotation), the initial position of the preview mask for the circle, brush and path, if the mouse is not over the darkroom area, is not the middle of the darkroom area.
It should work the same as in an image with no distortions.

Also, do not reset posx, posy if a mask is being created/edited, to avoid strange behaviours.
(ie: drag a mask outside the darkroom area and it will be displayed at the top-left corner)

### DIFF
--- a/src/develop/masks/masks.c
+++ b/src/develop/masks/masks.c
@@ -1245,6 +1245,12 @@ int dt_masks_events_mouse_leave(struct dt_iop_module_t *module)
   if(darktable.develop->form_gui)
   {
     dt_masks_form_gui_t *gui = darktable.develop->form_gui;
+
+    // if masks are being created or edited don't reset the position
+    if(gui->creation || gui->form_dragging || gui->source_dragging || gui->point_dragging >= 0
+       || gui->feather_dragging >= 0 || gui->seg_dragging >= 0 || gui->point_border_dragging >= 0)
+      return 0;
+
     gui->posx = gui->posy = -1.f;
   }
   return 0;
--- a/src/develop/masks/brush.c
+++ b/src/develop/masks/brush.c
@@ -2228,11 +2228,8 @@ static void dt_brush_events_post_expose(cairo_t *cr, float zoom_scale, dt_masks_
       float xpos, ypos;
       if(gui->posx == -1.f && gui->posy == -1.f)
       {
-        float zoom_x, zoom_y;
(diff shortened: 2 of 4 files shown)


TurboGit:
For me this always place the mask at the center of the picture and not at the center of the current view (piece). So if if the center of the picture is not visible you won't see the mask.

In liquify I have the masks always at the center of the current view.

TurboGit:
Working thanks!

One question I thought you had also implemented the mouse_leave action to draw the current mask on the center of the screen when leaving the darkroom. This is not working anymore...

src/develop/masks/masks.c
```diff
@@ -1245,6 +1245,12 @@ int dt_masks_events_mouse_leave(struct dt_iop_module_t *module)
   if(darktable.develop->form_gui)
   {
     dt_masks_form_gui_t *gui = darktable.develop->form_gui;
+
+    // if masks are being created or edited don't reset the position
```
TurboGit: That's it, so you've disable the move at the center when moving the mouse out of the darkroom. I have done just the opposite in liquify. I find it better to have the mask at the center instead of staying on the border and sometime not very visible.

edgardoh: I've done this so when the mask or a node/segment is dragging and the mouse get out of the darkroom it doesn't move to the left-top corner.
In creation, I like better the preview to stay at the edge instead of jump to the middle, but that's a matter of taste. To change it back I think that the only thing needed is to remove the gui->creation from the if(). Let me know if you want that.

TurboGit: I think consistency if very important. I must say that I have no string opinion on this. The only thing is that we must choose one option had have the masks, the retouch and the liquify works the same. Any opinion on that? We need to agree on one behavior :)

TurboGit: If you really like better to have the preview stays on the border I'm ok with that and will change liquify accordingly. Just to be sure we agree on this.

edgardoh: I agree that consistency is important. And I do like this behaviour better. When I move the mouse outside the darkroom area my eyes are on the destination and I find distracting to see with the corner of my eyes the preview jumping to the center.

TurboGit: Done.

PR #1743, opened 2018-10-07
fix gui modules order

edgardoh:
When moving from one image to another within the darkroom, if the new image has multiple instances of a module, the order of the modules on the module list is not correct.
This is a gui issue only, the dev->iop list is correct.
This PR fixes this, and display multi-instance modules collapsed instead of expanded.

This also fixes some gtk assertions on the same scenario.


TurboGit:
Code looks good but I cannot reproduce the issue. Do you have a step by step way to check this? Thanks.

aurelienpierre:
I think using one of mine's or rawfiner's denoising styles will produce this issue quite consistently. Copy/pasting the editing from one image to another with such modules is almost garanteed to produce it.

TurboGit:
Reproduced. Thanks.

PR #1745, opened 2018-10-10
Darkroom UI refactor

aurelienpierre:
This is what is currently discussed on the dev mailing list.

This PR aims at drawing an unidirectional path between the modules, to show in which order the user has to set them to get better results while avoiding editing in circles (going back on previously set modules because we changed their input). This order is mostly the pixelpipe one.

**basic modules tab:**
every technical module related to dynamic range restoration (+ crop and flip) that needs to be set before the input color correction. Leaving this tab, the image should look dull and unclipped.

**correction modules tab:**
every technical module related to image reconstruction (denoising, highlights recovery). Leaving this tab, the image should look clean and dull.

After the corrections tab, the input color profile is applied to correct the color. This is a breaking point because the input data must be in the same state than the data used to calibrate the sensor and produce the ICC profile, to get good color accuracy. Everything happening before the input color profile should be considered "technical retouch" whereas users get more creative freedom after this breaking point".

**tone modules tab:**
every artistic module related to (local and global) contrast and lightness. Drawback: several modules handle colors as well. Setting the tones will affect the saturation, especially in Lab, so it's better to do it before the colors.

**color modules tab:**
every artistic module related to colors. Not much change here except for the color mapping module here.

**effects module tab:**
every filter used to enhance the picture.

**favorite modules:**
this stays unchanged and allows users to set their favorite modules for quick access.

Benefits:
---

- if you don't know where to begin with 8 different modules that fix the lightness and 3 different modules for tonemapping, now you just have to follow the UI order.
- fixing the dynamic range before the color correction is critical to retain color accuracy. It is not the place in the exposure module to try setting the contrast.
- if you follow this order, you shouldn't have to ever come back on a previously set module. That's especially important when you use parametric blending and complex masking operations.
- it makes the technical part of the editing (reconstruction and recovering) fully separate from the creative part.
- it is like the 10:10 hand position on a steering wheel: you can still drive one-handed if you want, but that's the safest way to do it.
- it's a simple change that does not require a lot of testing and threaten the immediate stability of the software.

Screenshots
---

Note that the modules "scale pixels", "rotate pixels" and "raw black/white point" are hidden by default, and usefull only for some weird sensors.

Vote for (or against) it
---

https://framaforms.org/darktable-ui-refactor-1540510408

Feedback
---

So far, some developpers have been positive about that as well as most of the new users. Long-time users have mixed feelings.


MenfinGIT:
Using Darktable since several years, I think this UI refactor is a great idea. Not only for new users but for the old ones too.
It's a great ergonomic improvement for Darktable

acogent:
Me, I am a new user and I do not master well the "pixel pipe".
So, if I am offered a logical order, I will assimilate much faster !
Thank you Aurelien

TurboGit:
Please see : #1776

This should make everybody happy hopefully :)

aurelienpierre:
So far:
capture du 2018-10-26 16-16-40

blonchkman:
Congratulations on this reorganization proposal to which I subscribe without reservation.

As an amateur and still rather beginner, I renew without ceases this little tango of reorganization of the history. Thus, it seems to me that this proposal to reorganize the modules according to the order imposed by the mathematical logic of their treatment in the stack will help me in my future editing work. Moreover, it forms a logical sequence perfectly understandable for the layman.

I was amazed for a long time to see the order of treatment of modules move when manipulated. To avoid the side effects of these reclassifications, one could have imagined:

- an automatic solution by separating the view of the history (variable by nature), of the order in which the modules are executed (always fixed).
- or manual: by allowing a form of displacement of the order of the modules in the history after its compression.

Aurélien's proposal seems to me to be a perfectly credible compromise with the great advantage of being probably the easiest to implement ;-)

PR #1746, opened 2018-10-12
Sort aspect ratio

TurboGit:
Add a way to sort on the actual aspect ration (final one, after cropping).

I need that from time to time to select a set of picture in a specific aspect (portrait or landscape) for an exhibition.

### DIFF
--- a/src/common/collection.c
+++ b/src/common/collection.c
@@ -373,6 +373,7 @@ gchar *dt_collection_get_sort_query(const dt_collection_t *collection)
         break;
 
       case DT_COLLECTION_SORT_ID:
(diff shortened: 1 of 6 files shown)


TurboGit:
There is a dead-lock if you use the aspect-ratio filter while importing new pictures. To be fixed.

TurboGit:
I'd like some testing, if you read this and build darktable. Please test and report here. Thanks.

TurboGit:
Should be fixed! Thanks for reporting this case.

edgardoh:
All good now.

TurboGit:
Ok, let's merge for a wider audience then. Thanks for the feedback.

PR #1607, opened 2018-01-03
Add scrollbars to lighttable

oexler:
Here is my attempt to add scrollbars for the center.
They show the current position like the previous indicators on the window border.
But you can also use them to change the position.

I have modified the clamping in the zoomable light table view to always show at least one column or row on the left and bottom side.

PR #1747, opened 2018-10-13
Add scrollbars to lighttable #2

oexler:
As the first PR #1607 got merged, but then reverted again due to the problem with the map view, here is the follow-up PR.
It includes the previous commits and the fix for map view and the shifted picture when zooming out.
It also replaces some space around the center drawing area with a margin that gets removed when the scrollbar is visible, so it should't be so ugly anymore.

### DIFF
--- a/src/control/control.c
+++ b/src/control/control.c
@@ -176,7 +176,7 @@ void dt_control_cleanup(dt_control_t *s)
 
 gboolean dt_control_configure(GtkWidget *da, GdkEventConfigure *event, gpointer user_data)
 {
-  darktable.control->tabborder = 8;
+  darktable.control->tabborder = 2;
   const int tb = darktable.control->tabborder;
   // re-configure all components:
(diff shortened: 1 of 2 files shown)


TurboGit:
Still some issues on the diaporama view now. Each time I enter I get 5 times the following message:

(darktable:6174): GLib-CRITICAL **: 23:17:37.145: g_strlcpy: assertion 'src != NULL' failed

I'll try to test the capture view tomorrow. But I suppose it also needs some fixes.

TurboGit:
Just noticed... "diaporama" in my first comment is the French for slideshow :)

oexler:
I couldn't find any reference to the center widget in slideshow.c, so I just compiled darktable without my changes and I also get 6 times
(darktable.exe:20340): GLib-CRITICAL **: g_strlcpy: assertion 'src != NULL' failed
so I don't think this message has anything to do with this pull request.

TurboGit:
I think that this has broken the "zoomable light table". It is not possible to drag&drop the pictures on this view. Can you have a look?

PR #1748, opened 2018-10-14
Fix: Custom image order only in view Lighttable

monsieurmona:
@edgardoh found that dragging the main image darktable view is also possible, if "custom sort" is selected. 

This fix checks the current view and registers drag and drop only in lighttable view.

### DIFF
--- a/src/views/lighttable.c
+++ b/src/views/lighttable.c
@@ -366,11 +366,7 @@ static void _update_collected_images(dt_view_t *self)
                               &lib->statements.main_query, NULL);
 
   _unregister_custom_image_order_drag_n_drop(self);
-
-  if (darktable.collection->params.sort == DT_COLLECTION_SORT_CUSTOM_ORDER)
-  {
-    _register_custom_image_order_drag_n_drop(self);
-  }
+  _register_custom_image_order_drag_n_drop(self);
 
   dt_control_queue_redraw_center();
 }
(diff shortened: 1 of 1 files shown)


TurboGit:
Sounds good! Thanks.

TurboGit:
We still have issues with this, it crashes darktable-generate-cache tool.

And I see here, 2 calls:

_unregister_custom_image_order_drag_n_drop(self);
_register_custom_image_order_drag_n_drop(self);

Why is that needed? You unregister and register the same actions! I have removed those calls on my side the drag&drop seems to be working. Is that really needed?

TurboGit:
I'll revert this part for now, a crash is too bad.

monsieurmona:
Thanks for your feedback. I wonder, this is almost as the same as before (the usual developer excuse). It works on my machine. Do you have a callstack? 

The unregister removes all connections (if there are any) whereas the register, connects only if we are in light table and sort by custom order. 

I'll try to find the trouble. But if you could get me callstack of that crash, it would help me a lot.

monsieurmona:
Got the the callstack

TurboGit:
> The unregister removes all connections (if there are any) whereas the register, connects only
> if we are in light table and sort by custom order.

Ok, but do we need to do that every time we updade the collected images? In all case this should probably only happen in GUI mode, no? This could explain why darktable-generate-cache is crashing?

monsieurmona:
I found the collection is updated every time the sorting was changed. So I put the un/registering there. I did not find any better place. My knowledge of the darktable architecture is still not worth to mention. So I would be happy to get advice. 

I have fixed the crash. Please see https://github.com/darktable-org/darktable/pull/1749

TurboGit:
But my question is why do we need to unregister and register again. This seems independent on the actual sorting, the actions are all for drag&drop on the gui. So what is the scenario where this is needed? I'm asking as after removing the 2 calls to fix the crash, the "custom sort" seems to still be working. I may be missing some corner case...

TurboGit:
I see, so it is probably more appropriate to move those unregister/register in _view_lighttable_collection_listener_callback.

monsieurmona:
I'll have a look in the evening.

PR #1754, opened 2018-10-16
Use middle-click to control clone location.

TurboGit:
This is needed as ctrl-click is already used for selecting sharp edge
with creating path.

WIP: this is work in progress to initiate discussion.

First it is not complete, the same change is needed for cirlcle/ellipse/brush.

We can alternatively:
- use another modifier than GDK_CONTROL_MASK for absolute selection of the clone.

### DIFF
--- a/src/develop/masks/path.c
+++ b/src/develop/masks/path.c
@@ -1020,7 +1020,7 @@ static int dt_path_events_button_pressed(struct dt_iop_module_t *module, float p
   else
     masks_border = MIN(dt_conf_get_float("plugins/darkroom/masks/path/border"), 0.5f);
 
-  if(gui->creation && which == 1 && (state & (GDK_CONTROL_MASK | GDK_SHIFT_MASK)))
+  if(gui->creation && which == 2 && (state & (GDK_CONTROL_MASK | GDK_SHIFT_MASK)))
   {
     // set some absolute or relative position for the source of the clone mask
     if(form->type & DT_MASKS_CLONE) dt_masks_set_source_pos_initial_state(gui, state, pzx, pzy);


TurboGit:
The other modifier can be : GDK_MOD1_MASK (corresponding to ALT key).

edgardoh:
And is getting in the way even when creating the path, not only for the first node. Too bad, I like the cntrl+click for this.

The middle click doesn't feel comfortable to me, maybe lack of use or maybe my mouse.
I did make a quick try with the GDK_MOD1_MASK but it has no effect, maybe the alt key is reserved for something?
If not the alt key maybe we can use the right-click + cntrl | shift, but since the right-click alone is used for cancel the add it can be kind of confusing.
Clearly we are running out of keys!

TurboGit:
Working for me, you need to also add the support for it in masks.c

in dt_masks_set_source_pos_initial_state:
```
  if(state & GDK_MOD1_MASK)
    gui->source_pos_type = DT_MASKS_SOURCE_POS_ABSOLUTE;
  else if(state & GDK_SHIFT_MASK)
    gui->source_pos_type = DT_MASKS_SOURCE_POS_RELATIVE_TEMP;
  else
    fprintf(stderr, "unknown state for setting masks position type\n");
```

edgardoh:
I did, but if it works for you I can investigate later what happen here.

For me is either the Alt or the right-click, what do you think?

edgardoh:
@aurelienpierre , now right-click, with or without modifiers, cancel the add, but that can be fixed. 
That feature has already been requested, and yes,  it would be great, but there's a couple of issues:
-when adding a shape we don't have access to the image, so the calculation should be done asych, not sure how to achieve that.
-the "best" source needs to be a clean area with a texture that is good to cover the destination area, I don't know how to define that, is quite different as how denoise algorithms select a patch.

Yet another option is cntrl+shift, so shift set the relative source and cntrl+shift the absolute one. It can work.

@TurboGit , BTW this was my mistake, if you want I can work on it.

TurboGit:
@edgardoh : ok, I let you handle this and yes the ctrl-shift sounds good. Thanks.

I'm closing this.

PR #1755, opened 2018-10-17
Introduce a new gamma correction for color profiles based on video so…

TurboGit:
…ftware

Works better than legacy gamma for high dynamic range (> 10 EV)

### DIFF
--- a/data/kernels/basic.cl
+++ b/data/kernels/basic.cl
@@ -296,7 +296,7 @@ highlights_1f_lch_xtrans (read_only image2d_t in, write_only image2d_t out, cons
   const int ylid = get_local_id(1);
   const int xgid = get_group_id(0);
   const int ygid = get_group_id(1);
-  
+
   // individual control variable in this work group and the work group size
   const int l = mad24(ylid, xlsz, xlid);
   const int lsz = mul24(xlsz, ylsz);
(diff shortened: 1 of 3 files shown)


TurboGit:
Previous PR was here #1730

TurboGit:
BTW, Aurélien better to no commit the fr.po in a PR as this file is almost fully changed every time and so there is big chance that we get conflicts.

aurelienpierre:
> BTW, Aurélien better to no commit the fr.po in a PR as this file is almost fully changed every time and so there is big chance that we get conflicts.

Ok, I thought it would help.

edgardoh:
I've been playing with this and the new color balance and they work very nice, thanks @aurelienpierre !

The only thing is the color pickers, I assume that they work like this for performance reasons, but it would be nicer if values are updated when the user finish drawing the area. Also the buttons could be state buttons instead of push button, to have a visual clue of what is being defined. This will be more user-friendly and closer to how other modules work.

aurelienpierre:
Thanks @edgardoh !

> but it would be nicer if values are updated when the user finish drawing the area

I don't know, I really don't like having values updated as soon as I draw stuff and before I'm done drawing, especially with huge RAWs. But state buttons are a good idea, I have to check how to do that.

edgardoh:
I think we are saying the same thing, what I want is the values to be
updated only when I have finished drawing the rectangle, but without the
need of pushing the button again.
Now if I'm not happy with the results I have to push the button again, draw
the area again and push the button once again, and so on... what I propose
is to draw the area, values are updated, if I draw another area values are
updated again, etc.

aurelienpierre:
> Now if I'm not happy with the results I have to push the button again, draw the area again and push the button once again, and so on... what I propose is to draw the area, values are updated, if I draw another area values are updated again, etc.

hmm I have no idea how to do that.

edgardoh:
If you don't mind I can give it a try, I'll probably have some time until I
get an answer to my question on pixl.us

aurelienpierre:
> If you don't mind I can give it a try, I'll probably have some time until I get an answer to my question on pixl.us

no problem. Thank you !

PR #1756, opened 2018-10-17
fix set clone masks source position

edgardoh:
As @TurboGit  reported, the cntrl-click gets in the way when creating a path, that uses it to create sharp edges.
Now to set an absolute position on a clone mask source shift+cntrl-click is used.

### DIFF
--- a/src/develop/masks/brush.c
+++ b/src/develop/masks/brush.c
@@ -1280,7 +1280,9 @@ static int dt_brush_events_button_pressed(struct dt_iop_module_t *module, float 
   else
     masks_density = MIN(dt_conf_get_float("plugins/darkroom/masks/brush/density"), 1.0f);
 
-  if(gui->creation && which == 1 && (state & (GDK_CONTROL_MASK | GDK_SHIFT_MASK)))
+  if(gui->creation && which == 1
+     && (((state & (GDK_CONTROL_MASK | GDK_SHIFT_MASK)) == (GDK_CONTROL_MASK | GDK_SHIFT_MASK))
+         || ((state & GDK_SHIFT_MASK) == GDK_SHIFT_MASK)))
   {
     // set some absolute or relative position for the source of the clone mask
     if(form->type & DT_MASKS_CLONE) dt_masks_set_source_pos_initial_state(gui, state, pzx, pzy);
--- a/src/develop/masks/circle.c
+++ b/src/develop/masks/circle.c
@@ -246,7 +246,9 @@ static int dt_circle_events_button_pressed(struct dt_iop_module_t *module, float
     dt_control_queue_redraw_center();
     return 1;
   }
-  else if(gui->creation && (state & (GDK_CONTROL_MASK|GDK_SHIFT_MASK)))
+  else if(gui->creation && which == 1
+          && (((state & (GDK_CONTROL_MASK | GDK_SHIFT_MASK)) == (GDK_CONTROL_MASK | GDK_SHIFT_MASK))
+              || ((state & GDK_SHIFT_MASK) == GDK_SHIFT_MASK)))
   {
     // set some absolute or relative position for the source of the clone mask
     if(form->type & DT_MASKS_CLONE) dt_masks_set_source_pos_initial_state(gui, state, pzx, pzy);
--- a/src/develop/masks/ellipse.c
+++ b/src/develop/masks/ellipse.c
@@ -638,7 +638,9 @@ static int dt_ellipse_events_button_pressed(struct dt_iop_module_t *module, floa
     dt_control_queue_redraw_center();
     return 1;
   }
-  else if(gui->creation && (state & (GDK_CONTROL_MASK | GDK_SHIFT_MASK)))
+  else if(gui->creation && which == 1
+          && (((state & (GDK_CONTROL_MASK | GDK_SHIFT_MASK)) == (GDK_CONTROL_MASK | GDK_SHIFT_MASK))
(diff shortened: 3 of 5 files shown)


edgardoh:
I'm also having a gtk assertion here:
https://github.com/darktable-org/darktable/blob/master/src/iop/rawdenoise.c#L958
not sure why

TurboGit:
Good thanks.

TurboGit:
@edgardoh : I think I have fixed this issue.

edgardoh:
> @edgardoh <https://github.com/edgardoh> : I think I have fixed this issue.
>
Yes, I can't replicate it anymore. Thanks.

aurelienpierre:
I'm testing that right now, how/where is it supposed to work ?

aurelienpierre:
I see now. It's really nice :grinning:

edgardoh:
I forget, with only SHIFT pressed it should set a relative position. From
then, all masks will have the source at the same relative position than the
first one.

aurelienpierre:
I don't know where to dump that (maybe we should enable Github issues tracker), but I have a small problem with the drawn masks in the retouch module, in healing mode:
1. request the brush tool
1. adjust the size of the brush to be really small
1. begin to paint : the size is automatically increased as soon as we click. The only way to get very fine brush lines is to resize the size (Shift + Scroll) after the drawing.

edgardoh:
Works fine for me.

aurelienpierre:
That's strange. I get that output:
```
rt_process_forms: missing form=1867975264 from array
rt_process_forms: missing form=1867975360 from array
rt_process_forms: missing form=1867975808 from array
rt_process_forms: missing form=1867976032 from array
rt_process_forms: missing form=1867976544 from array
rt_process_forms: missing form=1867976992 from array
rt_process_forms: missing form=1867979936 from array
rt_process_forms: missing form=1867980384 from array
rt_process_forms: missing form=1867980704 from array
rt_process_forms: missing form=1867981056 from array
rt_process_forms: missing form=1867981152 from array
rt_process_forms: missing form=1867981344 from array
```

and also:
```
[opencl_retouch] couldn't enqueue kernel! -4
```

but I'm pretty sure it's not related.

aurelienpierre:
Ok, I found the culprit: the brush size gets resized when "smoothing of the brush strokes" is set at medium or high, in the preferences.

aurelienpierre:
But changing back from low to medium seems to solve the issue. Weird. So it was something in the settings.

TurboGit:
I have it set to medium. Maybe related to memory???

TurboGit:
Just saw your reply Aurélien, strange indeed !

aurelienpierre:
speaking of which, @edgardoh, I have another editing stack that makes the whole OS crash:

Picture: https://www.dropbox.com/s/jijut3baqifvhe5/mckayla-keiferhunnifordphotography.NEF?dl=0

XMP:
[mckayla-keiferhunnifordphotography.NEF.zip](https://github.com/darktable-org/darktable/files/2508086/mckayla-keiferhunnifordphotography.NEF.zip)

edgardoh:
You'll have to be more specific on how to duplicate it.

aurelienpierre:
> You'll have to be more specific on how to duplicate it.

Just export the picture in JPEG on the disk, with the XMP stack provided.

edgardoh:
Works fine for me.

aurelienpierre:
It seems that the crash was related to the OpenCL masking thing you mentionned elsewhere and solved by #1770. 

However, on the first retouch module (with gaussian blur patches), I always get:

```
340,530509 [dev_pixelpipe] took 0,165 secs (0,210 CPU) processed `retouche 1' on GPU, blended on GPU [export]
[opencl_retouch] couldn't enqueue kernel! -4
347,314161 [dev_pixelpipe] took 6,784 secs (14,076 CPU) processed `retouche' on CPU, blended on CPU [export]
```

The `retouch 1` instance only contains healing spots.

edgardoh:
The retouch have a printf() to display that error, other modules have
dt_print(DT_DEBUG_OPENCL...), so it may happen on other modules and its not
displayed.
This is just to get some feedback, I'll change it eventually with the
proper dt_print().

But to the point, -4 is out of resources, so it depends on the resources
available and how much is needed by the retouch, nothing I can do about it.
Most modules use a tiling process that consumes less memory, but the
healing cannot be tiled, so that is not an option.

What I would like to do is to tell dt how much memory is needed so if
there's not enough it can fall back to cpu without the error message,
there's something like that for tiling, but I don't know if it can be used
on non-tiled operations.

aurelienpierre:
> But to the point, -4 is out of resources, so it depends on the resources

I know. I wonder if others get the same error on the same picture + XMP. Out of ressource on a Nvidia Quadro with 4 GB RAM seems suspicious to me, that's all.

aurelienpierre:
Ok I have found it.

I run Ubuntu 18.04 with Gnome desktop. Using Gnome + Xorg, Xorg takes at least 1066 MiB of vRAM by itself and dt exports this picture in 23 s, due to the fallback on CPU for retouch and equalizer.

Running Gnome + Wayland takes barely 71 MiB or vRAM, dt doesn't fallback on CPU and exporting this pictures takes 13 s.

So, Xorg + Gnome + Ubuntu 18.04 are a big pile of garbage with a lot of caching and overhead.

moy:
In case you've missed it, @edgardoh, there's an issue with the cross showing the source of the clone in case of rotated image (orientation = portrait or crop and rotate): https://redmine.darktable.org/issues/12424. Other shapes are properly rotated.

edgardoh:
Thanks, I didn't see it.

PR #1757, opened 2018-10-18
Color balance : minor fixes

aurelienpierre:
Improve the RGB -> HSL -> RGB conversion accuracy in UI by moving the luma correction to commit_params(). Now the sliders RGB values are corrected only to make them consistent with the HSL ones.

Improve the performance of the lift/gamma/gain in C & SSE2 codepathes by 2× avoiding to create temp variables in the loops. Now, once the temporary RGB values are computed in RGB[3], we reuse that vector for every computation.

Resize the hard range (from 0 to 25 %) of the saturation sliders which will make it more accurate to set, we can still output values from 0 to 100 % using the keyboard. 25 % is more than enough in real life.

Fix a small labelling issue on the contrast tooltip text.

### DIFF
--- a/src/iop/colorbalance.c
+++ b/src/iop/colorbalance.c
@@ -179,7 +179,7 @@ int legacy_params(dt_iop_module_t *self, const void *const old_params, const int
   }
   return 1;
 }
-/*
+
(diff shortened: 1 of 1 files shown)


TurboGit:
Good thanks!

PR #1760, opened 2018-10-19
Add collection and sorting based on aspect-ratio.

TurboGit:
This is the second version. The first add only a filter, now it is
also possible to collect images based on their finale (cropped)
aspect ratio with all the provided operators on the collect module.

Finaly in this second version we keep the aspect ratio in the
database which makes the sorting/collecting as fast as for other
filters. For old pictures already in the database, the aspect ratio
are updated on the database with a max of 5 seconds at a time.
In case of time out the aspect ratio, the computation will be done
in the next cycle.

Note that on my computer I need a bit less than 20s for computing
the aspect ratio of 50k pictures.

The database format is upgraded and so from this commit it is not
possible to revert.


TurboGit:
Thanks for reporting. Point 1 and 2 should be fixed now. I have also fixed the case where we copy/paste history as we possibly change the aspect ratio too.

For point 3, this is expected I think. Computing the aspect for all pictures currently in the db could take a very long time. This is so done in chunk of 5s. If we time-out when end-up with a mixed order in the lighttable. The next time the remaining pictures will be handled. And this possibly 2, 5 or 10 times. Then when all the aspect ratio are computed we are fine as we update when needed aspect for touched pictures only (leaving darkroom, changing orientation on the lighttable, using copy/paste or discard history on the lighttable).

For the record, and as noted in my log message, on my computer I needed a bit less than 20s to update 50k pictures.

edgardoh:
All good now, thanks.

PR #1761, opened 2018-10-20
Fix: Move images in zoomable light table

monsieurmona:
* Remove drag and drop registration for zoomable light table
* Pan if custom sorting is not required

### DIFF
--- a/src/views/lighttable.c
+++ b/src/views/lighttable.c
@@ -69,6 +69,8 @@ static gboolean go_pgdown_key_accel_callback(GtkAccelGroup *accel_group, GObject
 
 static void _update_collected_images(dt_view_t *self);
 
+static gboolean _is_custom_image_order_required();
+
 /**
  * this organises the whole library:
  * previously imported film rolls..
@@ -1882,7 +1884,7 @@ int button_pressed(dt_view_t *self, double x, double y, double pressure, int whi
   lib->select_offset_x += x;
   lib->select_offset_y += y;
 
-  if (dt_control_get_mouse_over_id() < 0)
+  if (dt_control_get_mouse_over_id() < 0 || !_is_custom_image_order_required(self))
   {
     lib->pan = 1;
   }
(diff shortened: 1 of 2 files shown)


src/views/lighttable.c
```diff
@@ -2501,29 +2503,51 @@ void gui_init(dt_view_t *self)
...
+    // only in file manager
+    // only in light table
+    // only if custom image order is selected
+    dt_view_t *current_view = darktable.view_manager->current_view;
+    if (layout == file_manager_layout &&
+        darktable.collection->params.sort == DT_COLLECTION_SORT_CUSTOM_ORDER &&
+        current_view &&
+        current_view->view(self) == DT_VIEW_LIGHTTABLE)
+    {
+      custom_image_order_required = TRUE;
```
TurboGit: return TRUE;

src/views/lighttable.c
```diff
@@ -2501,29 +2503,51 @@ void gui_init(dt_view_t *self)
...
+    }
+  }
+
+  return custom_image_order_required;
```
TurboGit: return FALSE.

src/views/lighttable.c
```diff
@@ -2501,29 +2503,51 @@ void gui_init(dt_view_t *self)
   g_signal_connect(G_OBJECT(display_profile), "value-changed", G_CALLBACK(display_profile_callback), NULL);
 }
 
+static gboolean _is_custom_image_order_required(dt_view_t *self)
+{
+  gboolean custom_image_order_required = FALSE;
```
TurboGit: can be removed with proposed change below

src/views/lighttable.c
monsieurmona: Done

src/views/lighttable.c
monsieurmona: Done

src/views/lighttable.c
monsieurmona: Done

src/views/lighttable.c
TurboGit: Perfect! Thanks.

monsieurmona:
Thanks for your help!

PR #1762, opened 2018-10-21
Color balance : Fix a bug with the color neutralization

aurelienpierre:
The optimizer was broken after the luminance correction was moved to `commit_param()`
Basically, RGB values were luminance-corrected twice.
Now, the auto-optimizers use the same RGB normalization as user inputs.

Correct the RGB parameters in luminance also for lift/gamma/gain ProPhoto RGB

Improve a bit the accuracy of HSL -> RGB but it's still not perfect…


TurboGit:
Thanks!

PR #1763, opened 2018-10-21
color picker in profile log

edgardoh:
Changed the behaviour of the color picker on the unbreak input profile, now values are updated when:
-the user finish to draw the area
-select another color picker
-select the same color picker


TurboGit:
Thanks!

PR #1764, opened 2018-10-21
Makes the darkroom tools button more visible when active.

TurboGit:
The raw-overexposed, overexposed, gammut check, softproof buttons
are now more visible when activated. This is also user configurable.

I'm maybe too old but I have hard time to see if the buttons on the darkroom are activated or not. This is now configurable via CSS and a default color is proposed to make this more visible.

### DIFF
--- a/data/darktable.css.in
+++ b/data/darktable.css.in
@@ -683,3 +683,8 @@ cell:selected
   /* insensitive */
   color: black;
 }
+
+#darkroom_tools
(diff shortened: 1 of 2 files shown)


TurboGit:
I have a better solution.

TurboGit:
I have a better solution.

TurboGit:
I have a better solution.

PR #1765, opened 2018-10-22
Log profile : fix the luma estimation in the auto-optimizer.

aurelienpierre:
The luma is the energy of the light, computed from the RGB trichromatic signal.
This implies knowing in which RGB space we are, but this module is applied before
the input color profile, so it is a medium-relative RGB un-normalized space.
Previously, we assumed to be in ProPhotoRGB because it
is the largest RGB space. While it gives fair results most of the time, some sensors
behave quite poorly, especially those having a strong green cast in the shadows
at high ISO, because the G channel weighs for 72 % in the luma computation.

As a less precise, but more robust, estimator of the luma, we now take the max of the RGB
channels for each sample (grey, black, white). This doesn't always remap the L values (in Lab)
between [17; 96] (the black is sometimes in [0; 12]), but avoids silly remappings
(black = [25; 60]) that happened with the previous approach.

### DIFF
--- a/src/iop/profile_gamma.c
+++ b/src/iop/profile_gamma.c
@@ -349,9 +349,8 @@ static void apply_auto_grey(dt_iop_module_t *self)
   dt_iop_profilegamma_params_t *p = (dt_iop_profilegamma_params_t *)self->params;
   dt_iop_profilegamma_gui_data_t *g = (dt_iop_profilegamma_gui_data_t *)self->gui_data;
 
-  float XYZ[3];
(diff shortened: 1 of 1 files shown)


aurelienpierre:
Yes, it's normal. When adding contrast, what it does actually is lighting greys > fulcrum and darkening greys < fulcrum. The fulcrum is the grey that is not affected by the correction, and sets the transition point.

TurboGit:
Perfect then, thanks.

PR #1767, opened 2018-10-22
Color balance : 1 fix, 2 improvements

aurelienpierre:
Fix an an error in the color-neutralization optimizer : the HSL sliders for gain were not updated after the optimization (1 missing line)

Improve the luma/color optimizers : when no color patch is defined, automatically use the whole image average, min and max (previously, we returned an error message asking fon initialization). This allows a very fast workflow in 3 clicks when the picture is not too difficult.


edgardoh:
There's an issue with the display mask:

-activate color balance
-with opencl active set a parametric mask on base curve
-display mask

with cpu seems to work fine

TurboGit:
I can reproduce indeed! Really strange the bad interaction between both modules.

PR #1768, opened 2018-10-23
color picker in colorbalance

edgardoh:
Ported the color picker behaviour from  #1763 

GitHub did make a mess with the code, but there's some missing darktable.gui->reset = 1;
see:
static void apply_lift_auto(dt_iop_module_t *self)
...
  dt_bauhaus_slider_set_soft(g->lift_factor, p->lift[CHANNEL_FACTOR] - 1.0f);
...

not sure is intended.

### DIFF
--- a/src/iop/colorbalance.c
+++ b/src/iop/colorbalance.c
@@ -87,6 +87,17 @@ typedef enum _controls_t
   BOTH
 } _controls_t;
 
+typedef enum dt_iop_colorbalance_pickcolor_type_t {
+  DT_PICKCOLBAL_NONE = 0,
+  DT_PICKCOLBAL_HUE_LIFT = 1,
+  DT_PICKCOLBAL_HUE_GAMMA = 2,
+  DT_PICKCOLBAL_HUE_GAIN = 3,
+  DT_PICKCOLBAL_LIFT_FACTOR = 4,
+  DT_PICKCOLBAL_GAMMA_FACTOR = 5,
+  DT_PICKCOLBAL_GAIN_FACTOR = 6,
+  DT_PICKCOLBAL_GREY = 7
+} dt_iop_colorbalance_pickcolor_type_t;
+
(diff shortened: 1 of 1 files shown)


src/iop/colorbalance.c
```diff
@@ -962,78 +810,644 @@ static inline void normalize_RGB_sliders(GtkWidget *R, GtkWidget *G, GtkWidget *
...
+  dt_bauhaus_slider_set_soft(g->gain_factor, p->gain[CHANNEL_FACTOR] - 1.0f);
+
+  dt_dev_add_history_item(darktable.develop, self, TRUE);
+}
+#endif
+
+static int call_apply_picked_color(struct dt_iop_module_t *self, dt_iop_colorbalance_gui_data_t *g)
+{
+  int handled = 0;
+  if(g->which_colorpicker == DT_PICKCOLBAL_HUE_LIFT)
```
TurboGit: I would certainly use a case statement. And possibly with the following scheme:

```
int handled = 1;

switch()
{
     case:
     ...
     default:
        handled = 0;
}
return handled;
```

src/iop/colorbalance.c
```diff
@@ -87,6 +87,17 @@ typedef enum _controls_t
   BOTH
 } _controls_t;
 
+typedef enum dt_iop_colorbalance_pickcolor_type_t {
```
TurboGit: Please { on next line.

src/iop/colorbalance.c
TurboGit: I'll do that in profle_gamma.c

src/iop/colorbalance.c
edgardoh: This is clang-format, on enums it works like this.

TurboGit: Yes, I know but that's not dt style (and that's not a style I have chosen, I wasn't dev at the start of the project, but consistency is important). Will fix :)

PR #1769, opened 2018-10-24
Two Windows specific build fixes

peterbud:
* create_metadata.sh: remove CR from $line in the generated metadata file - I think it should not influence the *nix shell script, but pls check
* Replace hard-coded Graphicsmagick versions with variables


aurelienpierre:
Does it fix the crashes during building on windows ?

peterbud:
No, this PR does not fix the crash, I'm looking into that one as well. These changes just fix the building process.

TurboGit:
I'm no Windows expert, nor do I use Windows but I trust you :) Will merge. Thanks.

PR #1770, opened 2018-10-24
Bugfix: messed-up masks in OpenCL with colorbalance and profilegamma

aurelienpierre:
Leaves the 4th channel of the picture untouched in OpenCL colorbalance and profilegamma modules. 

This avoids messing up the masks. I wasn't aware that the 4th channel was the alpha mask.

Solves the bug described in comments of #1767 by @edgardoh and the one I described in #1756.

### DIFF
--- a/data/kernels/basic.cl
+++ b/data/kernels/basic.cl
@@ -2189,7 +2189,7 @@ profilegamma_log (read_only image2d_t in, write_only image2d_t out, int width, i
 
   if(x >= width || y >= height) return;
 
-  float4 i = read_imagef(in, sampleri, (int2)(x, y));
+  const float4 i = read_imagef(in, sampleri, (int2)(x, y));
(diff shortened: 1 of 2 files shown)


TurboGit:
All good now thanks for the quick fix.

TurboGit:
Indeed, not even a warning.

edgardoh:
Kind of defeats the purpose of using a const.

aurelienpierre:
The original code of the colorbalance used the const that way, and I found that strange too. When you define `const float4 o`, if you later do `o = whatever`, it throws an error. But if you use the channels of `o` like `o.x = whatever` or `o.xyz = whatever`, it's ok. I don't understand either.

Anyway, writing OpenCL is very close to writing pure maths: I have only seen this level of abstraction in linear algebra programs (Matlab or Python Numpy), so I guess C is only an interface to call Assembly or something.

PR #1771, opened 2018-10-24
Enable GCC SSE2 -> AVX2 instructions conversion

aurelienpierre:
This adds `-msse2avx` flag so SSE2 intrinsics are converted to Intel AVX for best performance
See https://www.macieira.org/blog/2012/06/avx-optimised-raster-painting-for-windows-too/

In the equalizer iop, on SSE codepath, with an image at 36 Mpx, it reduces the runtime from 7s to 6s on Intel Xeon.


TurboGit:
The "mavx2" option is supported by GCC version 4.7 and higher. Are we sure all distribs are ready for that? I kind of remember a discussion about this on the mailing-list where some distribs were really lagging behind and I don't remember what is the oldest GCC version dt is supporting.

TurboGit:
Ok, safe on the darktable page here (https://www.darktable.org/install/) it is said that the GCC must be >= 5.0!

TurboGit:
Ok, so let's go for it.

LebedevRI:
You can't arbitrarily enable `-m` flags.
It *forces* the compiler to use those instruction sets
https://godbolt.org/z/JbX5_e
Not, what happens if the cpu does not support that instruction set?

aurelienpierre:
What's the option to politely suggest GCC to use AVX if available ?

LebedevRI:
`-march=native`, which is already specified by default (see `-- Checking for -march=native support` in the cmake output, and in the actually-used buildflags), should already do it.
(unless the compiler authors [intentionally] incorrectly specified the supported cpu features for that cpu)

And both compilers are smart enough to produce avx even when the sse intrinsics were used
https://godbolt.org/z/AeESmf

So i don't know what you were seeing / what is the question.

aurelienpierre:
That's strange, because, from what you say, compiling a `Release` build should already use AVX instructions, and yet, adding the `-mavx2` gives me a 17 % speedboost.

PR #1772, opened 2018-10-25
Revert "Enable GCC SSE2 -> AVX2 instructions conversion"

LebedevRI:
Reverts darktable-org/darktable#1771


TurboGit:
Thanks Roman. So I guess the only option would be to check for the availability on the CPU at configuration time.

cryptomilk:
That's why we created:

https://build.opensuse.org/package/show/home:darix:darktable:master/darktable

The you can choose the right build for your CPU ...

aurelienpierre:
It would be gool to have it on other distros as well.

PR #1773, opened 2018-10-25
Add option to change darkroom background color around the image

cryptomilk:
This implements the suggestions from:

https://discuss.pixls.us/t/neutral-theme-for-darktable-and-gimp/9504/27

### DIFF
--- a/data/darktableconfig.xml.in
+++ b/data/darktableconfig.xml.in
@@ -1079,6 +1079,24 @@
     <longdescription>process the image in darkroom mode with a small border. set to 0 if you don't want any border.</longdescription>
   </dtconfig>
   <dtconfig prefs="gui">
+    <name>plugins/darkroom/ui/background_color</name>
+    <type>
+      <enum>
+        <option>white</option>
+        <option>light gray</option>
+        <option>medium gray</option>
+        <option>spanish gray</option>
+        <option>dim gray</option>
+        <option>jet gray</option>
+        <option>black</option>
+        <option>default</option>
+      </enum>
+    </type>
+    <default>default</default>
+    <shortdescription>background color around the image in darkroom mode</shortdescription>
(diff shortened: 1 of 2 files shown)


TurboGit:
I'm not sure about this one. We don't have a single color option in the preference actually. And the solution is to change the CSS to setup a color. So can you clarify why this is needed here? Why not change from a CSS? I can't imagine a preference entry for every single configurable color :)

cryptomilk:
The background color changes how you perceive the colors in the image. A darker background and your image is brighter, a white background and your image is less colorful. Depending what you are developing for you want to change that background color.

It would be better to have directly as a small button in the darkroom ui to switch it without requiring a restart or editing a file. However my darktable skills are not the best :-)

aurelienpierre:
I have already thought about that, but I think it would be better to ship 2 or 3 different CSS templates (dark, neutral, light), and be able to select them from the preferences, rather than to hack just the background color value from the code. Last year, the hardcoded GTK colors were specifically removed from the source code.

maf:
@cryptomilk thank you for this PR.

I opt for having the configuration in the CSS and choosing theme from UI. However making own theme or ability to modify exiting one using UI (like using GTK+ color picker) is in my opinion *must have*.

@TurboGit Why? Dakrtable falls into color critical category and sits in the color management workfow. This starts with environment preparation (walls, desktop, lightning and so on) and later with calibration and profiling of the devices. Thus one can have custom workplace with specific lightning and may need to adjust background accordingly.  I myself modified theme because I print. Scan the linked thread for details.

TurboGit:
@maf : exactly what I said, it is so important than having preference for a single color is not a good solution. We can do that properly in CSS, or we need to provide a full preference tab to be able to change ALL colors.

cryptomilk:
Closing as this needs to be implemented differently. I'm happy to work on this but I need a helping hand or some pair-programming.

PR #1656, opened 2018-02-21
Added contextual help links

mpaglia0:

PR #1774, opened 2018-10-25
Update context sensitive help

rawfiner:
This pull request update the context sensitive help in the following ways:
- it adds the urls that @mpaglia0 put in https://github.com/darktable-org/darktable/pull/1656 (I had to do a rebase as one of the commits of that pull request was unrelated to this, so I removed it and kept the others). Thanks @mpaglia0 for all this work
- it change the way the cursor behaves. Before: it was a question mark, and became a hand when there was a help link. But in darktable, when going other the right panel, the shape was overrided. Now: it is a cross, and becomes a question mark on widgets that have a link (I think it is clearer like this), and the shape is no more changed by other events.
- a message is displayed to the user if he clicks on something for which we do not have any url
- the expanders have now help urls too. Previously, we had to click inside a module to get help, now we can click on its expander.
- the code gets the ui language, and checks if the usermanual is available in this language.

This PR solves the issue https://redmine.darktable.org/issues/8660

The only problem that remains, and I think it requires a huge GUI code change (at least, I do not see how to solve it without such a change), is that on some elements the cursor is a cross whereas they do have a help url, and the url open successfully on click.
I personally think we can live with that, as this behavior is only here for a small number of the widgets, and considering the amount of changes this may require in the code I do not know if it really is worth it (I already spent quite a lot on time on this, but did not find any solution).

### DIFF
--- a/src/libs/tools/global_toolbox.c
+++ b/src/libs/tools/global_toolbox.c
@@ -240,21 +240,21 @@ static void _main_do_event(GdkEvent *event, gpointer data)
           }
           if(base_url)
           {
-            dt_l10n_language_t *language = (dt_l10n_language_t *)g_list_nth(darktable.l10n->languages, darktable.l10n->selected)->data;
+            dt_l10n_language_t *language
+                = (dt_l10n_language_t *)g_list_nth(darktable.l10n->languages, darktable.l10n->selected)->data;
             char *lang = language->code;
             const int number_of_supported_languages = 4;
-            const char* supported_languages[4] = {"en", "fr", "it", "es"};
+            const char *supported_languages[4] = { "en", "fr", "it", "es" };
             bool is_language_supported = false;
-            for (int i = 0; i < number_of_supported_languages; i++)
+            for(int i = 0; i < number_of_supported_languages; i++)
             {
-              if (0 == strcmp(lang, supported_languages[i]))
+              if(0 == strcmp(lang, supported_languages[i]))
               {
                 is_language_supported = true;
                 break;
               }
             }
-            if (!is_language_supported)
-              lang = "en";
+            if(!is_language_supported) lang = "en";
             char *url = g_build_path("/", base_url, lang, help_url, NULL);
             // TODO: call the web browser directly so that file:// style base for local installs works
 #if GTK_CHECK_VERSION(3, 22, 0)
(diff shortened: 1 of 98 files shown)


src/libs/tools/global_toolbox.c
```diff
@@ -249,8 +264,13 @@ static void _main_do_event(GdkEvent *event, gpointer data)
 #endif
             g_free(base_url);
             g_free(url);
+            dt_control_log(_("Help url opened in web brower"));
           }
         }
+        else
+        {
+          dt_control_log(_("There is no help available for this element"));
```
TurboGit: Likewise, no capital letters on log messages.

src/libs/tools/global_toolbox.c
```diff
@@ -249,8 +264,13 @@ static void _main_do_event(GdkEvent *event, gpointer data)
 #endif
             g_free(base_url);
             g_free(url);
+            dt_control_log(_("Help url opened in web brower"));
```
TurboGit: No capital letters on log messages.

src/libs/tools/global_toolbox.c
```diff
@@ -238,8 +240,21 @@ static void _main_do_event(GdkEvent *event, gpointer data)
           }
           if(base_url)
           {
-            // TODO: try to use the currently set language
-            char *lang = "en";
+            dt_l10n_language_t *language
+                = (dt_l10n_language_t *)g_list_nth(darktable.l10n->languages, darktable.l10n->selected)->data;
+            char *lang = language->code;
+            const int number_of_supported_languages = 4;
```
TurboGit: I would certainly terminate the array with \0 and stop at the end. Having two coupled variables is not good I think.

src/libs/tools/global_toolbox.c
```diff
@@ -238,8 +240,21 @@ static void _main_do_event(GdkEvent *event, gpointer data)
...
+            const char *supported_languages[4] = { "en", "fr", "it", "es" };
+            bool is_language_supported = false;
```
TurboGit: should be gboolean (not sure bool is supported on all GCC version and CLang). I could be wrong, but in dt we do use gboolean.

rawfiner:
Thanks for the code review :-)
I will update the code soon

rawfiner:
These commits add the changes requested, and add a section id for color balance in the user manual so that it can be referenced in the url

rawfiner:
I have pulled master, and rebased the branch on master

PR #1775, opened 2018-10-26
Collect exposure

TurboGit:


TurboGit:
One point that could not work except in some countries is the dot vs coma returned by atof(). In France we use coma, but SQLite3 ask for dot. And as I said there was also precision issues and some pictures were not displayed (10 in the count(*) in the collection and only 8 in the lighttable for example).

PR #1777, opened 2018-10-26
Added ability to collect images by exposure in lighttable

mondoman712:
I'm pretty sure I've fixed the rounding errors this time. With regards to using commas instead of periods, that would require separating what's displayed from the values actually used and in my opinion is a separate issue since the aperture values also use periods.


mondoman712:
nvm apparently the last one was merged

PR #1780, opened 2018-10-28
control: Use gboolean instead of C99 stdbool

cryptomilk:
Looks like this fails on some Fedora systems as stdbool.h is not included. May also depend on some special CFLAGS. However the code uses gboolean everywhere else.

### DIFF
--- a/src/control/control.c
+++ b/src/control/control.c
@@ -75,7 +75,7 @@ void dt_control_init(dt_control_t *s)
   s->dev_zoom_x = 0;
   s->dev_zoom_y = 0;
   s->dev_zoom = DT_ZOOM_FIT;
-  s->lock_cursor_shape = false;
+  s->lock_cursor_shape = FALSE;
 }
 
 void dt_control_key_accelerators_on(struct dt_control_t *s)
@@ -100,12 +100,12 @@ int dt_control_is_key_accelerators_on(struct dt_control_t *s)
 
 void dt_control_forbid_change_cursor()
 {
-  darktable.control->lock_cursor_shape = true;
+  darktable.control->lock_cursor_shape = TRUE;
 }
 
 void dt_control_allow_change_cursor()
 {
-  darktable.control->lock_cursor_shape = false;
+  darktable.control->lock_cursor_shape = FALSE;
 }
 
 void dt_control_change_cursor(dt_cursor_t curs)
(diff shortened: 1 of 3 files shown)


mandree:
This goes a long way but isn't sufficient, see attached a snippet from the attached -> [build log](https://github.com/darktable-org/darktable/files/2522732/build.txt).

mandree:
I need another patch:
[0001-Further-fix-to-use-gboolean-TRUE-FALSE-instead-of-no.patch.txt](https://github.com/darktable-org/darktable/files/2522737/0001-Further-fix-to-use-gboolean-TRUE-FALSE-instead-of-no.patch.txt)
```
From 032ba4010d46f4b2f1aa0dccea5d5bc625fdbcd0 Mon Sep 17 00:00:00 2001
From: Matthias Andree <matthias.andree@gmx.de>
Date: Sun, 28 Oct 2018 12:39:55 +0100
Subject: [PATCH] Further fix to use gboolean TRUE/FALSE instead of
 non-standard true/false.

---
 src/libs/tools/global_toolbox.c | 4 ++--
 1 file changed, 2 insertions(+), 2 deletions(-)

diff --git a/src/libs/tools/global_toolbox.c b/src/libs/tools/global_toolbox.c
index c3526d034..238d52659 100644
--- a/src/libs/tools/global_toolbox.c
+++ b/src/libs/tools/global_toolbox.c
@@ -246,13 +246,13 @@ static void _main_do_event(GdkEvent *event, gpointer data)
             // array of languages the usermanual supports.
             // NULL MUST remain the last element of the array
             const char *supported_languages[] = { "en", "fr", "it", "es", NULL };
-            gboolean is_language_supported = false;
+            gboolean is_language_supported = FALSE;
             int i = 0;
             while(supported_languages[i])
             {
               if(!strcmp(lang, supported_languages[i]))
               {
-                is_language_supported = true;
+                is_language_supported = TRUE;
                 break;
               }
               i++;
-- 
2.17.2

```

mandree:
In case it helps, here is my [CMakeCache.txt](https://github.com/darktable-org/darktable/files/2522739/CMakeCache.txt) that fails on upstream's f0ad51b3382ea9d62df49d1007d59d8ad82beacb

rawfiner:
Thanks for these changes. I tried to fetch this, and it compiled successfully on my linux mint.
This PR solves mistakes I introduced previously due to a bad knowledge of gboolean.
Thank you again for the changes, and my apologies.

TurboGit:
Indeed, I missed that during my review :( We don't want to use bool! Thanks.

PR #1781, opened 2018-10-28
Add SQL script to migrate albums from Capture One Pro

bobobo1618:
I organize my photos by album in Capture One Pro and wanted to preserve that structure when moving to Darktable.

This script takes Darktable's and Capture One Pro's SQLite3 databases and creates a tag for every Capture One Pro collection, then adds images with the tag, matching them by absolute path.

### DIFF
--- a/tools/migrate_capture_one_pro.sql
+++ b/tools/migrate_capture_one_pro.sql
@@ -1,0 +1,46 @@
+ATTACH 'data.db' AS data;
+ATTACH 'library.db' AS library;
+ATTACH 'C1ProCat.cocatalogdb' AS c1cat;
+
+-- Temporary table to store the mapping from Capture One Pro albums and Darktable tags to be looked up when bringing in images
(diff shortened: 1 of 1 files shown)


tools/migrate_capture_one_pro.sql
```diff
@@ -0,0 +1,46 @@
+ATTACH 'data.db' AS data;
```
TurboGit: Can you add a set of comments at the start of the script about the setup before starting the script. From where it should be run? What are the db to copy before starting... etc. A bit of context for first time users.

And also what it does. As I understand it, it creates a tag for each Capture One collection? All this must be clearly described. Thanks.

bobobo1618: I've added a section at the start. Let me know if that's sufficient.

TurboGit:
Good! Thanks.

PR #1785, opened 2018-10-30
profilegamma : add presets

aurelienpierre:
Add generic camera presets for 16, 14, 10, 12, 10, 8 EV which can be mapped and auto-applied depending on the ISO setting.
Typical cameras dynamic range for each ISO range can be found here : https://www.dxomark.com/best-cameras-for-landscape-under-45200-dollars

Change the label "mode" for "curve" to advertise the similarity between what is done here and the basecurve module.


TurboGit:
Looks good. Thanks.

PR #1786, opened 2018-10-30
Added shuffle option to sorting -- Looking for opinions

mondoman712:
I added the ability to shuffle your images in the light table.

### DIFF
--- a/src/common/collection.c
+++ b/src/common/collection.c
@@ -447,6 +447,10 @@ gchar *dt_collection_get_sort_query(const dt_collection_t *collection)
         sq = dt_util_dstrcat(sq, ORDER_BY_QUERY, "aspect_ratio DESC, filename DESC, version DESC");
         break;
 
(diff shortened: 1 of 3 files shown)


TurboGit:
Do you have a usage of this or is that just for the challenge :) ?

mondoman712:
I was just scrolling through my images looking for any pairs of visually similar images from different places and thought this would be a better way to find some, but also just an interesting way to look through a collection of images. Once I came up with the idea I did then just want to see if I could implement it and I thought I'd just do it and make a PR and we can discuss whether anyone else might find it useful here.

TurboGit:
Ok, let's see what others think about this. Not a large piece of code, so probably not disruptive, but yet I do want to be sure it is useful enough to be integrated.

aurelienpierre:
I can actually see a use for that.

TurboGit:
Oki, let's plan this for 2.8.

TurboGit:
manually merged.

PR #1787, opened 2018-10-31
freaky BW

edgardoh:
This is based on GMIC Freaky B&W.
It has 6 modes:
the mix allows to choose the rgb combination to generate the b&w image
the max uses the maximum of rgb channels
the dst use the sine transform to reconstruct from laplacian
the fft use the fourier transform
the plain dst reconstruct from laplacian using the padded image
the dst 2 use the un-padded image

dst and fft uses the fftw3, so a new library must be added.

It is very sensitive to image scale & crop, but I care more about the global effect than the details, so it is not a big deal to me.

It is not for every picture, but the effect is nice, and with all the modes, I always find one that I like.

### DIFF
--- a/src/iop/freaky_bw.c
+++ b/src/iop/freaky_bw.c
@@ -1,0 +1,1217 @@
+/*
+    This file is part of darktable,
+    copyright (c) 2018 edgardo hoszowski.
+
+    darktable is free software: you can redistribute it and/or modify
+    it under the terms of the GNU General Public License as published by
+    the Free Software Foundation, either version 3 of the License, or
+    (at your option) any later version.
+
+    darktable is distributed in the hope that it will be useful,
+    but WITHOUT ANY WARRANTY; without even the implied warranty of
+    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+    GNU General Public License for more details.
+
+    You should have received a copy of the GNU General Public License
+    along with darktable.  If not, see <http://www.gnu.org/licenses/>.
+*/
+
+
+/*
+   based on G'MIC filter "Freaky B&W"
+   http://gmic.eu/
+
+   it also have a dst version in addition to the fft one
+
+#@gui Freaky B&amp;W : fx_freaky_bw, fx_freaky_bw_preview
+#@gui : Strength (%) = float(90,0,100)
+#@gui : Oddness (%) = float(20,0,100)
+#@gui : Brightness (%) = float(0,-100,100)
+#@gui : Contrast (%) = float(0,-100,100)
+#@gui : Gamma (%) = float(0,-100,100)
+#@gui : sep = separator(), Preview type = choice("Full","Forward horizontal","Forward vertical","Backward
+horizontal","Backward vertical","Duplicate top","Duplicate left","Duplicate bottom","Duplicate right")
+#@gui : sep = separator(), note = note("<small>Author: <i>David Tschumperl&#233;</i>.      Latest update:
+<i>09/30/2015</i>.</small>")
+fx_freaky_bw :
+  -repeat $! -l[$>] -split_opacity -l[0]
+    -to_rgb
+
+    # Estimate gradient field of B&W result.
+    --expand_xy 1,0 -channels. 0,4
+    -f. ">if (c!=4,i,
+           Rx = i(x+1,y,0,0) - i(x,y,0,0);
+           Ry = i(x,y+1,0,0) - i(x,y,0,0);
+           Rn = Rx^2 + Ry^2;
+           Gx = i(x+1,y,0,1) - i(x,y,0,1);
+           Gy = i(x,y+1,0,1) - i(x,y,0,1);
+           Gn = Gx^2 + Gy^2;
+           Bx = i(x+1,y,0,2) - i(x,y,0,2);
+           By = i(x,y+1,0,2) - i(x,y,0,2);
+           Bn = Bx^2 + By^2;
+           n = 1e-5 + max(Rn,Gn,Bn)^"{$2%}";
+           val = 0;
+          if (Rn>=Gn && Rn>=Bn,
+            i(x,y,0,3) = Rx/n; val=Ry/n,
+          if (Gn>=Rn && Gn>=Bn,
+            i(x,y,0,3) = Gx/n; val=Gy/n,
+            i(x,y,0,3) = Bx/n; val=By/n));
+          val
+         )"
+    -channels. 3,4
+    -luminance[0] ia={0,ia}
+
+    # Estimate laplacian of final image.
+    -s. c
+    -f.. "i - i(x-1,y,0,0)"
+    -f. "i - i(x,y-1,0,0)"
+    -+[-2,-1]
+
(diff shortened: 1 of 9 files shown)


src/iop/freaky_bw.c
```diff
@@ -0,0 +1,1217 @@
...
+
+const char *name()
+{
+  return _("freaky bw");
+}
+
+int groups()
+{
+  return IOP_GROUP_COLOR;
+}
```
rabauke: Possibly the effects group is more appropriate.

tools/iop_dependencies.py
```diff
@@ -369,6 +369,10 @@ def add_edges(gr):
   # want to change exposure/basecurve after tone mapping
   gr.add_edge(('exposure', 'tonemap'))
   gr.add_edge(('basecurve', 'tonemap'))
+  # fbw goes after tonemap but before exposure/basecurve
+  gr.add_edge(('freaky_bw', 'tonemap'))
```
rabauke: I wonder, if this is the right place in the pixel pipe. Isn't this too early? I had expect to find it much later, e.g., after the 'monochrome' module, in particular after exposure and basecurve.

src/iop/freaky_bw.c
TurboGit: No, we need to call dt_iop_get_group() now to let the possibility to reorganize the IOP.

TurboGit:
When I saw this I said why not! But I'm wondering if we want to introduce some random/artistic effects like this we could have lot and lot of IOP at the end. So I'm wondering if we don't want to have a generic effect like IOP and a combobox to select the effect. Not sure, just starting a discussion in this direction. What would be the pros and the cons?

tools/iop_dependencies.py
edgardoh: I tried several places in the pipe, this is what I find to be the best.

edgardoh:
>  some random/artistic effects
>
 I didn't close my eyes and flip a coin to decide which GMIC filter to
port, much less if share it or not. I've been using it for a while and I
find it very useful, and I'm running under the assumption that there's
someone out there with similar editing styles/taste than me.

> if we don't want to have a generic effect like IOP and a combobox to
> select the effect. Not sure, just starting a discussion in this direction.
> What would be the pros and the cons?
>
Different filters may need different places in the pipe, so only one IOP
won't do it. Also, dt only allow a fixed number of controls, so probably a
redesign will be needed there.

src/iop/freaky_bw.c
edgardoh: I use it mostly for the BW conversion, that's why I place it there, but its really the same to me.

TurboGit:
@edgardoh : sorry I didn't intend to say that you've randomly chosen a GIMIC filter, only that such filter and many in GMIC are kind of random in the sense of you do not use it to achieve some precise goal as you can't really guess what will happen on the picture before hand really. With this module for example you drag some sliders and some effect are coming and passing and even at some point you seem to have globally the same effect with different values of the same slider. That's why I call this random. Hope this clarifies my thinking :)

TurboGit:
@edgardoh : otherwise yes I guess a generic module won't fit the current architecture.

src/iop/freaky_bw.c
```diff
@@ -0,0 +1,1217 @@
...
+  int width_fft;
+  int height_fft;
+  int width_fft_complex;
+  int height_fft_complex;
+
+  dt_pthread_mutex_t *fftw3_lock;
+} fbw_fft_t;
+
+typedef enum dt_iop_fbw_bw_methods_t {
+  dt_iop_fbw_bw_mix_dst = 0,
```
TurboGit: should be all upper-case : DT_IOP_FBW_BW_MIX_DST

edgardoh:
> @edgardoh <https://github.com/edgardoh> : sorry I didn't intend to say
> that you've randomly chosen a GIMIC filter, only that such filter and many
> in GMIC are kind of random in the sense of you do not use it to achieve
> some precise goal as you can't really guess what will happen on the picture
> before hand really. With this module for example you drag some sliders and
> some effect are coming and passing and even at some point you seem to have
> globally the same effect with different values of the same slider. That's
> why I call this random. Hope this clarifies my thinking :)
>
Non taken ;), I'm sure that by now you know that I don't share things just
for the fun of it, I just wanted to be clear for a casual reader, sorry if
it sounded harsh.

In a way you're right, the effect is unpredictable, and it very much
depends on the image (and the crop), that's why I have 6 modes. But the,
let's say, effect of each slider is well defined: given the default effect,
one can anticipate what a change in the rgb sliders will do. The same with
the oddness, once you are familiar with it.

src/iop/freaky_bw.c
edgardoh: Sure, but let's first work on the description of the 6 modes, I really don't have a clue what text to use so it makes some sense for the user, not everyone is familiar with the fourier or sine transform and the laplacian, and the user shouldn't even care about it.

edgardoh:
Here are a couple of examples of the most common use (for me). Please don't share the one with people.

In this I have several images and I need a quick edit so I can share them. Light is very bad and I shoot with very high ISO, so I just care about capturing the right moment. If I get one that I really like I'll take more time later.
One is with the freaky b&w with default parameters and some exposure/contrast, the other same setting but with the monochrome. The difference here is subtle, that depends on the image.
aikido_freaky_bw
aikido_monochrome

In this I sometimes have an image that I like but again very noisy, this transform the noise on some texture that, depending the image & personal taste, is nice.
This two are with and without the freaky b&w and just the base curve.

eh_20180331_dsc1892_dt
eh_20180331_dsc1892_dt_01

edgardoh:
I should also show the difference regarding image scale:




aikido_fbw_full
macro_fbw_full

edgardoh:
I'm having second thoughts about this. 
My first thought was: if its useful for me it may be useful for others, but that is not good enough, it also needs to be a good fit for darktable.

Obviously I have nothing against having artistic filters in dt, but the image scale is a problem, and I don't think that anything can be done here about it.
The only option I can think of is to force all the pipes to process the image at the same size. Maybe is not that difficult and all it takes is a new setting on an existing module or a new iop, but for just one filter it doesn't seem to make sense. If the GMIC plugin is implemented this will be something to think about.

I'm closing this for now, if the time comes when it makes more sense I can always reopen it.

rabauke:
> When I saw this I said why not! But I'm wondering if we want to introduce some random/artistic effects like this we could have lot and lot of IOP at the end. So I'm wondering if we don't want to have a generic effect like IOP and a combobox to select the effect. Not sure, just starting a discussion in this direction. What would be the pros and the cons?

I have done something into this direction by writing a GMIC module for darktable. It implements some selected GMIC filters for darktable (see my posts on the mailing list). I like this extension very much and use it often. However, there is a major problem with such a multi-effects-filter module. The module has a specific place in the pixel pipeline. There are always some effect-filters for which this is not the appropriate place.

Regarding the freaky BW filter: I am unsure if one should include it into darktable as it has a more artistic character. Furthermore, I would expect problems in the preview, in particular when the preview shows only parts of the image as the module utilizes a DFT, which in a non-local operation.

PR #1788, opened 2018-11-01
properly disable color picker on reset

moy:
This applies to "color balance" and "unbreak input profile" (log
mode).

Before this commit, a click on "reset parameters" was disabling the
colorpicker, but keeping the colorpicker icon active. The reason was
that self->request_color_pick was set to DT_REQUEST_COLORPICK_OFF but
g->which_colorpicker was kept to its former value.

Introduce disable_colorpick to reset both variables at the same time,
and add one missing set_colorpick_state call to propagate the
information to the GUI.

### DIFF
--- a/src/iop/colorbalance.c
+++ b/src/iop/colorbalance.c
@@ -1120,6 +1120,13 @@ static void apply_gain_auto(dt_iop_module_t *self)
 }
 #endif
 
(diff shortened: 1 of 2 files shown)


TurboGit:
Sounds good. Thanks.

PR #1789, opened 2018-11-01
mask: give hint about ctrl+click to rotate ellipses

moy:
The hint was given when the mouse was over a point only, but it's
valid everywhere.


TurboGit:
Thanks! That's nice to fix all those small details that make a real difference!

PR #1790, opened 2018-11-02
Colorbalance : fix ASC CDL implementation

aurelienpierre:
I did a mistake in the ASC CDL (slope/offset/power) implementation, using `power = (1 / (user_input + 1))` (as the lift/gamma/gain does).

The standard implementation is `power = user_input`, which is corrected here as `power = (2 - user_input)` to be consistent with the user input scaling (between -1 and 1).

This will affect existing editings and needs to be corrected ASAP, although the visual difference is not dramatic. It also makes the auto-neutralization more accurate and the settings easier to control.


src/iop/colorbalance.c
```diff
@@ -1120,13 +1139,6 @@ static void apply_gain_auto(dt_iop_module_t *self)
 }
 #endif
 
-static void disable_colorpick(struct dt_iop_module_t *self)
```
TurboGit: @aurelienpierre : I suppose all the changes below are not expected. right?

TurboGit: Yes, "git reset -p HEAD^" but this is not on the top commit so you also need to play with git rebase -i I suppose. I can handle this in less than a minute, so I can do it if you don't feel messing with Git. Just let me know.

aurelienpierre: Ok I let you do it. Thank you !

TurboGit: Done, please double check that I haven't mess anything. Thanks.

aurelienpierre: Looks good. I will test later.

PR #1791, opened 2018-11-02
Move crop & rotate to the linear part of the pixelpipe

aurelienpierre:
Rotation is a convolution. Convolutions should happen in a linear space. 

This is what happens when we rotate a gamma-corrected image (classic sRGB gamma 2.4/2.2 depending how blows the wind):
imagen 1
This what happens in a linear space:
imagen
Look at the edges…

For now, crop and rotate is done in the Lab part of the pixelpipe, just before colorbalance. First of all, Lab is by no mean a linear space (you convert from XYZ to Lab using a piece-wise x³ formula). Then, several aesthetic curves are applied in-between (gamma correction, colorchecker, colorbalance) that apply non-linear operation in a non-linear space. So this position makes no sense at all.

This PR moves the crop and rotate IOP just after the perspective correction and ensures the colorbalance  is properly positionned before the colorize IOP (where it was before). That is because the colorbalance position in the pixelpipe was defined according to the position of the crop and rotate IOP (which is a major case of rabbits + towers = fireworks) and not according to the position of other color-relative IOP.

### DIFF
--- a/src/iop/ashift.c
+++ b/src/iop/ashift.c
@@ -4215,7 +4215,7 @@ void init(dt_iop_module_t *module)
   module->params = calloc(1, sizeof(dt_iop_ashift_params_t));
   module->default_params = calloc(1, sizeof(dt_iop_ashift_params_t));
   module->default_enabled = 0;
-  module->priority = 205; // module order created by iop_dependencies.py, do not edit!
+  module->priority = 217; // module order created by iop_dependencies.py, do not edit!
   module->params_size = sizeof(dt_iop_ashift_params_t);
   module->gui_data = NULL;
   dt_iop_ashift_params_t tmp = (dt_iop_ashift_params_t){ 0.0f, 0.0f, 0.0f, 0.0f, DEFAULT_F_LENGTH, 1.0f, 100.0f, 1.0f, ASHIFT_MODE_GENERIC, 0,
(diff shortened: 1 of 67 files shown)


TurboGit:
This need quite a lot of testing and I fear it will break the history. The issue is with the back-transform, if one has put a mask on some IOP it needs to follows the transformation. IIRC we had some issues with the liquify module for example. There is some corner cases where the transform/back-transform is not accurate and then the mask is slightly moved around. I'll try to find the discussion and the example I had at that time.

aurelienpierre:
The crop and rotate will still be after liquify though. That PR is not critical but could make the edge-preservation of pictures better.

cryptomilk:
Rebase needed

TurboGit:
This is a very delicate PR for the transform and back-transform. I had no time to check that all is right with the masks when put before and after in the pixelpipe.

TurboGit:
Ok, I have found back my test image with the issue of transform/back-transform. This was due to the lens correction module which do not properly keep the mask position in monochrome module. I have also tested with the exposure module which comes before lens-correction and crop&rotate (now) and the masks position is "broken" the same with or without this PR.

So, all in all, I'll merge now. Thanks.

TurboGit:
BTW, I'll do a manual merge with conflicts resolved.

PR #1792, opened 2018-11-02
Update contextual help

rawfiner:
This pull request adds some links for the contextual help, adds anchors to some links, and makes the code more robust.
In particular, if the system language is not detected and the language is not set, there is no segmentation fault anymore, as well as if the code is compiled without l10n support.
I tested the code both on linux mint 18 and inside a docker based on fedora.


rawfiner:
I thought it was ready, but it seems it is not ready yet. Closing the PR for the moment, and I will fix everything and reopen it later.

rawfiner:
The problem with usermanual should be fixed by commit 76a8216.

rawfiner:
By the way, issue https://redmine.darktable.org/issues/8660 should be closed

TurboGit:
Thanks!

PR #1793, opened 2018-11-02
Update non local means in denoise profiled

rawfiner:
This pull request contains several changes for the non local means algorithm:
(1) a replacement for the slider "radius" (added recently by @hanatos ), which impacts a lot the performance. Time of execution depends on the square of the radius. The replacement allows to have a similar visual effect without impacting performance by looking at patches more "wide-spread" without changing the number of patches. So there is no more performance issue, and the slider is much more usable in practice.
(2) a change in the algorithm's core that make it possible to use smaller force without having too much "unchanged" pixels. Previously, the central patch was used to denoise itself (which is absurd), thus it had a weight of 1, and smaller weights of other patches were not able to be considered correctly. Now, the central patch is not used. This affects the output little, so it should not be an issue for backward compatibility, but gives a wider "visually usable" range for the force slider.
(3) a change for the presets. I personally think using the non local means in the luminance preset is a bad idea, as non local means is not very consistent: sometimes it gives great outputs, but sometimes a lot of ugly artifacts. Thus, I put a wavelet instead, with a curve fitting the use in "lightness mode" (smooth a little bit less the fine grain noise). In addition, I changed the wavelet in color mode, in order to have a curve to better remove the fine-grain color noise. I think these presets will show the users that playing with the curve can allow them to tune the denoising.

Please note:
I tested the opencl code on CPU, as I don't have a GPU. It should work, but please try to test it on GPU anyway.
Code tested under linux mint 18 and in a docker based on fedora.

Here are some comparisons:
For point (1) (left is with these changes, right is with radius). See the outputs are very similar, whereas with these changes the code was 5 times faster
comparison_radius
For point (2) (left is old, right is new). Some pixels are left undenoised on the left.
comparison_central_patch
For point (3) (left is old, right is new). Output with new presets has less artifacts, and a noise more pleasing to look at.
comparaison_presets

### DIFF
--- a/data/kernels/denoiseprofile.cl
+++ b/data/kernels/denoiseprofile.cl
@@ -229,7 +229,7 @@ denoiseprofile_accu(read_only image2d_t in, global float4* U2, global float* U4,
 
 
 kernel void
-denoiseprofile_finish(read_only image2d_t in, global float4* U2, write_only image2d_t out, const int width, const int height,
+denoiseprofile_finish(read_only image2d_t in, read_only image2d_t in2, global float4* U2, write_only image2d_t out, const int width, const int height,
                              const float4 a, const float4 sigma2)
 {
   const int x = get_global_id(0);
@@ -240,8 +240,11 @@ denoiseprofile_finish(read_only image2d_t in, global float4* U2, write_only imag
 
   float4 u2   = U2[gidx];
   float alpha = read_imagef(in, sampleri, (int2)(x, y)).w;
+  /* in2 is the image after the precondition
+     orig is the pixel value from this (undenoised) image */
+  float4 orig = read_imagef(in2, sampleri, (int2)(x, y));
 
-  float4 px = ((float4)u2.w > (float4)0.0f ? u2/u2.w : (float4)0.0f);
+  float4 px = ((float4)u2.w > (float4)0.0f ? u2/u2.w : orig;
 
   px = (px < (float4)0.5f ? (float4)0.0f : 
     0.25f*px*px + 0.25f*sqrt(1.5f)/px - 1.375f/(px*px) + 0.625f*sqrt(1.5f)/(px*px*px) - 0.125f - sigma2);
--- a/data/kernels/denoiseprofile.cl
+++ b/data/kernels/denoiseprofile.cl
@@ -244,7 +244,7 @@ denoiseprofile_finish(read_only image2d_t in, read_only image2d_t in2, global fl
      orig is the pixel value from this (undenoised) image */
   float4 orig = read_imagef(in2, sampleri, (int2)(x, y));
 
-  float4 px = ((float4)u2.w > (float4)0.0f ? u2/u2.w : orig;
+  float4 px = ((float4)u2.w > (float4)0.0f ? u2/u2.w : orig);
 
   px = (px < (float4)0.5f ? (float4)0.0f : 
     0.25f*px*px + 0.25f*sqrt(1.5f)/px - 1.375f/(px*px) + 0.625f*sqrt(1.5f)/(px*px*px) - 0.125f - sigma2);
--- a/src/iop/denoiseprofile.c
+++ b/src/iop/denoiseprofile.c
@@ -299,11 +299,11 @@ void init_presets(dt_iop_module_so_t *self)
 {
   // these blobs were exported as dtstyle and copied from there:
   add_preset(self, _("chroma (use on 1st instance)"),
-             "gz04eJxjYGiwZ2B44MAAphv2b4vdb1mnYGTFLcZomryXxzRr910TRgYYaLADEkB1DvYQ9RSJURUDAIFFGLo=", 5,
+             "gz03eJxjYGiwZ2B44MAAphv2b4vdb1mnYGTFLcZomryXxzRr910TRgYYaLADEkB1DvYQ9eSLiUf9s+te/s/OjJPV/sA7Y3vZI/X2EHnyMAB8fx/c", 5,
              "gz12eJxjZGBgEGYAgRNODESDBnsIHll8AM62GP8=", 7);
   add_preset(self, _("luma (use on 2nd instance)"),
-             "gz04eJxjYGhwYGB4AMQM9gwMDfu3xe63rFMwsuIWYzRN3stjmrX7rgkDHDTYQdQ5gNTaUyhGVQwAImYYOg==", 5,
-             "gz12eJxjZGBgEGAAgWlODESDBnsIHll8AJKaGMo=", 7);
+             "gz03eJxjYGiwZ4AAIN2w388ywNJ97j7LS3qqpokcz03CM61NGRlgoMEOos4BpNaeQjF7zi3/7XJef7QLvPPUbnHHXag8eRgAb44fKg==", 5,
+             "gz12eJxjZGBgEABiRoYCJwaiQYM9BI8sPgBoshil", 7);
 }
 
 typedef union floatint_t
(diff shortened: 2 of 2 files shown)


TurboGit:
For 3 I found that there is less details in the new version. The image is smoother, less noise but not as sharp.

rawfiner:
Thank you for the feedback. I updated the luma preset so that it better preserves the sharpness.

aurelienpierre:
@rawfiner I'm assuming the central pixel was the top of a gaussian distribution. Since you put it at a null weight, did you normalize the gaussian window so the sum of its elements is still 1 ?

TurboGit:
All good to me and I have tested this on GPU. Thanks.

rawfiner:
@aurelienpierre I am not sure I understand your comment. The patchs are all "equal" regarding to the non local means algorithm: a patch which is more far away accounts as much as a patch which is very close to the central point. The only thing which impacts the patches weights is their similarity to the central patch, which is computed by doing the sum of the squared differences, for each pixel of the patch.

rawfiner:
@TurboGit Thank you for the tests :-)

aurelienpierre:
@rawfiner inside a patch, the pixels don't have a varying weight depending on their position from the center ? Usually the patches are 2D gaussian windows convolved on the picture as a correlation method. Convolutions need normalized kernels. I don't have the brainpower to look at that yet, but I think the patches need to be normalized to account for the loss of the central pixel to avoid numeric instabilities.

TurboGit:
Tested, effect looks good and indeed the time is never a problem with this implementation. Let's go for integration and have more field testing. Thanks for the contribution.

upegelow:
I'm seeing some quite heavy alias-like artifacts in structures with fine detail like foliage. See for example the leaves in this image: [https://www.dropbox.com/s/as0iozppfjalb4e/IMG_2517.CR2?dl=0](url), especially if fully zoomed in.

I see the artifacts with processing on CPU and GPU the like. In addition on the GPU I have intermittent "gaps" in the image when panning the view.

TurboGit:
@rawfiner , this gives lot of troubles I'll revert today until the issue is resolved.

rawfiner:
I am afraid it is normal to get artifacts using non local means, especially
for big patch sizes. That's precisely the reason why I changed the luma
preset to use wavelets instead even for the luma denoising.

I did not have time to test your image yet though. Maybe tomorrow, or
latter in the week.

May I ask what parameters you used?
I personaly use mostly patch sizes of 1 and 2, as 3 and 4 gives more
artifacts.

So, let's see why are we getting artifacts with this algorithm.

A patch is a square of pixels. A patch size of 1 means the patch is a 3x3
square, of 2 means the patch is a 5x5 square, of 3 means it is a 7x7
square, and of 4 means it is a 9x9 square.

Non local means works by doing a weighted average for every pixel,
computing the weights using a patch similarity measure: the more similar a
patch is to the patch centered on the considered pixel, the higher the
weight.
For each pixel of the image, 224 patches are considered to compute the
weighted average.

The problem we have is related to the similarity measure: to compare 2
patches, we do the sum of the squares of the pixel differences (and we
normalize this sum). Let's call this sdiff.
The lower it is, the better the similarity.
With this similarity measure, every pixels of the patches are involved the
same way: pixels which are far from the patch center are used the same way
as pixels which are close to the patch center.

With a big patch size like 4, we have 9x9=81 pixels patch. This is 32
pixels more than a 7x7=49 pixels patch, and these 32 pixels can have A LOT
of influence on the similarity.

Let's take an example. The image "Patch-original.png" represents the patch
we are considering: its center is the pixel we are denoising.
The images "Patch-candidate-1.png" and "Patch-candidate-2.png" are patches
we compare with the original patch.
For a human, it is obvious that candidate 1 is better than candidate 2.
Yet, considering black is 0, white is 1, and grey is 0.5, we get a sdiff of
30 for candidate 1, and 7.75 for candidate 2, thus the candidate 2 is said
to be more similar to the original patch by our similarity measure!
This will give more weight to candidate 2 than to candidate 1 in the
weighted average, and our central pixel will become grey instead of
remaining white.
Whereas if we take a patch size of 3, candidate 1 is perfectly similar to
the original patch (sdiff is 0).

The images "diff" represent these differences: blue is a 0.5 difference
(sdiff of 0.25) and pink is a 1 difference (sdiff of 1).

With this example, we see why we can get artifacts and why bigger patch
size is not necessarily better than smaller ones.

Now, why are we keeping this similarity measure? We could give more weights
to pixels close to the center of the patch than to those who are far away
from it?
Yes, we could do that.
But the fact that all pixels are considered the same way allows some
optimisations which result in a very fast implementation of the algorithm,
much faster that we could have if we had a weighted similarity measure.

Still, it is not impossible to have a better similarity measure and keep
good performances.
I have an idea of one measure that may lead to better results while being
compatible with the efficient implementation, but this won't be ready for
2.6.


Le 4 nov. 2018 7:02 PM, "Ulrich Pegelow" <notifications@github.com> a
écrit :

I'm seeing some quite heavy alias-like artifacts in structures with fine
detail like foliage. See for example the leaves in this image:
https://www.dropbox.com/s/as0iozppfjalb4e/IMG_2517.CR2?dl=0 <http://url>,
especially if fully zoomed in.

I see the artifacts with processing on CPU and GPU the like. In addition on
the GPU I have intermittent "gaps" in the image when panning the view.

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/darktable-org/darktable/pull/1793#issuecomment-435691630>,
or mute the thread

rawfiner:
Seems the images I attached to the email are not here ;-(
I will put them on a drive and give a link

rawfiner:
Ok @Turbogit
I am afraid these problems are related to the core of the algorithm itself. If you want to revert only the "ignoring central patch" fix while keeping the rest (coarse grain noise reduction changes and presets), you can revert these commits:
52b27a8dc067b014b5f25f5378b2a36fba6c9a52
a070f505ea9de47fa7a026eaa424286fc0784f77
eb8092620f84806f96798d9d90879fb2fa297370

hanatos:
i'm also having issues with this. please revert!

it breaks legacy history stacks and the results look like a lot of aliasing indeed. if you change the algorithm you need to provide a way to reproduce the old one, in general. if it's a mere bugfix that always makes pictures better there may be exceptions to this, but in this case i'm not happy with the output.

rawfiner:
Alright, thank you for the tests.

To explain what happens: keeping the central patch acts like a varying opacity that changes pixel per pixel, sometimes low, sometimes high, which results in unprocessed zones when the patches available are not similar enough to the central patch. I considered it was preferable to always denoise the pixel, and let the user choose the global opacity, as this forces the user to increase the force to avoid these unprocessed pixels. But I am ok with keeping current behaviour if you prefer it and for the legacy reason.

@Turbogit thanks for the revert.
Should I reopen a pull request for points (1) and (3) (as only point (2) was the problem)?

TurboGit:
@rawfiner : we need to be more cautious this tine.

In 1 I read "similar visual effect" is that similar or identical? Won't that break also the old history too? If it is faster and identical I would say yes, we want this.

For 3, it seems sound to me to go for it.

rawfiner:
@TurboGit (1) is similar but clearly not identical, there is no direct relation between it and the old radius (both have the effect of considering more widespread patches, but the way they do it is different). Yet, the "radius" slider it is supposed to replaced has never been in a release yet, it has been integrated in master very recently (if it had been in a release, it would have been too late to consider the replacement). So yes, it breaks the old history of people who used the master branch, but it does not break the old history of users of dt 2.4.4.

rawfiner:
(Search radius slider was introduced in: becfd43a79b3387a9d238f93c02c7b53c315ea4e )

TurboGit:
@rawfiner : Ok, I understand that there was no radius slider before. but there was an hard-coded value in the code. My question is then, with your proposal do the default gives the identical result? This was the case with the slider @hanatos has merged recently, it was just exposing the hard coded value with a default being identical.

rawfiner:
@Turbogit , yes, the default gives an identical result as in 2.4.4.
I will redo separated pull requests for this and for the preset (probably tonight), to make testing easier.

PR #1795, opened 2018-11-03
BUGFIX : color picker wrong readings

aurelienpierre:
what was done
---

When using local and global color pickers, the pixelpipe was redirected to a special processing function that disabled the gamma.c IOP to avoid messing up the readings.

problem
---

For some time, the gamma.c IOP does nothing else than converting RGB output to `uint8`. Now the gamma correction is done in the colorout.c IOP. See #1794 

Consequently, the readings of the color picker Lab were always off.

fix
---

In the special processing function without gamma, we select and disable the colorout.c IOP instead of the gamma.c.

warning
---

It does not fix the bug when LittleCMS is used.


TurboGit:
Sounds good. Thanks.

One question, do you foresee a fix for the LittleCMS case?

aurelienpierre:
> One question, do you foresee a fix for the LittleCMS case?

Yes, I'm working on it too, but it won't be that easy.

PR #1796, opened 2018-11-03
Tonecurve presets & GUI improvements

aurelienpierre:
_WIP, help welcomed_

screenshot
---
capture d ecran du 2018-11-03 03-53-46

tl; dr
---

This PR is intended to enable further the logaritmic workflow introduced with the unbreak profile module. It only affects the GUI. No internal data changes.

New features
---

- Add presets for generic image-processing functions such as Log2, Exp2, Gamma 2, Gamma 0.5
- Rename existing presets with more explicit names describing the actual operations (contrast high becomes contrast - high (gamma))
- Existing presets get 7 nodes instead of 5. That will allow to use the curve the same way as the Lightroom equalizer (with blacks, shadows, mid-tones, highlights and whites), with the presets as a base and define filmic curves easily.
- Port the log-log scaling from the basecurve module, and enable semi-log scales. The benefit is, in semi-log (y) mode, when you draw a line, it's actually an exponential (good to revert partially the log profile). The other way around, in semi-log (x), when you draw a line, it's actually a log. The log-log mode is usefull to zoom closer on the shadows. This only affects the display.
- Let the user choose the base of the log scaling, to match the grey level set in unbreak profile. This means you can center the graph to whatever grey level you want, to help you see what you are doing.
- Rename the modes : automatic, manual, auto XYZ and auto RGB were not super explicit. They are renamed "color space" : Lab (bounded channels), Lab (independant channels), RGB (bounded channels), XYZ (ibounded channels)
- Expose the interpolation method : previously, this was implicitely set by applying presets. Now user can choose between 3 splines with different fitting properties.
- Hide the 3 Lab tabs when bounded channels mode are used. They are useless in this setup, and it's misleading to have Lab tabs when you work in RGB or XYZ anyway.
- In RGB and XYZ mode, convert the luminance values of the color samplers to XYZ, so they are properly remapped on the graph. Previously, in RGB and XYZ, you still got Lab luminance values. Again, misleading.
- Remap the color samplers accordingly to the log scale.
- Remap the inset histogram accordingly to the log scale.

What does not work (yet)
---

- global color pickers values conversion to XYZ fails, I don't know why (this Lab -> XYZ conversion is used successfully several times in the 2 other modules I hacked)
- after setting up color pickers, when the cursor leaves and enter back the graph zone, color pickers disappear (GUI glitch)
- the histogram in log-log or semi-log (x) sometimes disappear and values are wrong anyway.

### DIFF
--- a/src/gui/draw.h
+++ b/src/gui/draw.h
@@ -287,7 +287,7 @@ static inline void dt_draw_histogram_8_log(cairo_t *cr, uint32_t *hist, int32_t 
 static inline void dt_draw_histogram_8_log_base(cairo_t *cr, uint32_t *hist, int32_t channels, int32_t channel, float base_log)
 {
   cairo_move_to(cr, 0, 0);
-  for(int k = 0; k < 256; k++) cairo_line_to(cr, CLAMP((int)255.0 * logf(k/255.0 * (base_log - 1.0f) + 1.0f), 0, 255) / logf(base_log), logf(1.0 + hist[channels * k + channel]));
+  for(int k = 0; k < 256; k++) cairo_line_to(cr, CLAMP((int)(255.0 * logf(k/255.0 * (base_log - 1.0f) + 1.0f)), 0, 255) / logf(base_log), logf(1.0 + hist[channels * k + channel]));
   cairo_line_to(cr, 255, 0);
   cairo_close_path(cr);
   cairo_fill(cr);
--- a/src/gui/draw.h
+++ b/src/gui/draw.h
@@ -287,7 +287,12 @@ static inline void dt_draw_histogram_8_log(cairo_t *cr, uint32_t *hist, int32_t 
 static inline void dt_draw_histogram_8_log_base(cairo_t *cr, uint32_t *hist, int32_t channels, int32_t channel, float base_log)
 {
   cairo_move_to(cr, 0, 0);
-  for(int k = 0; k < 256; k++) cairo_line_to(cr, CLAMP((int)(255.0 * logf(k/255.0 * (base_log - 1.0f) + 1.0f)), 0, 255) / logf(base_log), logf(1.0 + hist[channels * k + channel]));
+  for(int k = 0; k < 256; k++)
+  {
+    const float x = (float)k / 255.0f;
+    cairo_line_to(cr, logf(x * (base_log - 1.0f) + 1.0f) / logf(base_log), logf(1.0 + hist[channels * k + channel]));
+  }
+
   cairo_line_to(cr, 255, 0);
   cairo_close_path(cr);
   cairo_fill(cr);
--- a/src/gui/draw.h
+++ b/src/gui/draw.h
@@ -83,7 +83,7 @@ static inline void dt_draw_grid(cairo_t *cr, const int num, const int left, cons
 }
 
 static inline void dt_draw_loglog_grid(cairo_t *cr, const int num, const int left, const int top,
-                                       const int right, const int bottom, const int base)
+                                       const int right, const int bottom, const float base)
 {
   float width = right - left;
   float height = bottom - top;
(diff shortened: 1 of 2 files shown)


TurboGit:
Don't know what the exact fix for the picker, but the wrong code is this one.

1. you do
float Y_picker_min = 0.0f, Y_picker_max = 0.0f, Y_picker_mean = 0.0f;

2. then ONLY if autoscale_ab is s_scale_automatic_xyz or s_scale_automatic_rgb you update the 3 floats above

3. you use the possibly unitialized 3 floats for the picker after:

TurboGit:
Looking at the code I see that raw_* [ch_L] are two order of magnitude larger than the corresponding Y_picker_*.

Also, I'm not sure why you set the raw_[ch_L] but this is what cause the trouble with the picker. What's the point, since the picker will reset those values, no? And you well get them back on next iteration:

    raw_mean = self->picked_color;
    raw_min = self->picked_color_min;
    raw_max = self->picked_color_max;

But I don't know well this part of the code.

TurboGit:
I have I think the explanation for the why:
1. you get the picked L
2. you convert to XYZ and Y is a bit less that L on the raw_*[ch_L]
3. you stuff this lower value into raw_*[ch_L]
4. the mouse move over the tone curve, each mouse-move you redraw
5. you get back to 1, read the now lower L value, redo the computation L becomes lower and lower

At the end it is 0.

You can see that sometime the colorpicker red area on the tone curve gets moved to the left.

So the actual circuitry in your code is wrong. Hope this helps.

aurelienpierre:
> So the actual circuitry in your code is wrong. Hope this helps.
 Yes it does, thank you very much !

src/iop/tonecurve.c
```diff
@@ -1200,17 +1533,46 @@ static gboolean dt_iop_tonecurve_draw(GtkWidget *widget, cairo_t *crf, gpointer
...
+          Y_picker_mean = XYZ[1];
+        }
+
         picker_scale(raw_mean, picker_mean);
         picker_scale(raw_min, picker_min);
         picker_scale(raw_max, picker_max);
 
+        if (autoscale_ab == s_scale_automatic_xyz || autoscale_ab == s_scale_automatic_rgb)
+        {
+          raw_min[ch_L] = Y_picker_min;
```
TurboGit: isn't here picker_* expected instead of the raw_*?

aurelienpierre: picker_* are the global pickers you set in the left panels. raw_* are the local ones you set in the module. We take care of them lines 1482-1503

TurboGit:
@aurelienpierre : what about this branch? do you plan something for 2.6? thanks.

src/gui/draw.h
```diff
@@ -277,7 +277,7 @@ static inline void dt_draw_histogram_8_log(cairo_t *cr, uint32_t *hist, int32_t
 static inline void dt_draw_histogram_8_log_base(cairo_t *cr, uint32_t *hist, int32_t channels, int32_t channel, float base_log)
 {
   cairo_move_to(cr, 0, 0);
-  for(int k = 0; k < 256; k++) cairo_line_to(cr, CLAMP((int)logf(k * (base_log - 1.0f) + 1.0f), 0, 255) / logf(base_log), logf(1.0 + hist[channels * k + channel]));
+  for(int k = 0; k < 256; k++) cairo_line_to(cr, CLAMP((int)255.0 * logf(k/255.0 * (base_log - 1.0f) + 1.0f), 0, 255) / logf(base_log), logf(1.0 + hist[channels * k + channel]));
```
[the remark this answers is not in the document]
aurelienpierre: Done

aurelienpierre:
as of now, the XYZ conversions are fixed, in the graph and in the labels. The histogram are still not log and always Lab even when using XYZ or RGB

cryptomilk:
Someone neeed to retrigger the travis build, it has been stuck pulling a docker image.

aurelienpierre:
Everything should be working now. I have removed the histograms from the view in RGB and XYZ mode since they are valid only for Lab. Also, they won't be shown in log-log and log (x) scales, because they need to be rescaled. But these 2 issues are non-trivial to solve, and the linear modes work as before.

TurboGit:
Again really nice! Thanks Aurélien.

TurboGit:
@aurelienpierre : testing more today I found a quite annoying issue! The colorpicker zone displayed in the tone-curve is wrong and seems to not match the actual selected area. Indeed, changing the curve in the area does not change the color-picker zone. Can you have a look?

TurboGit:
@aurelienpierre : really not sure about this, but here is a possible fix and my thinking to come with that.

#1833

cryptomilk:
If I'm selecting in `Lab, independent channels` then I get the a and b channels. However if you close the tone curve maybe add another and then reopen it again. It only shows the L channel and a and b are hidden. I need to reselect `Lab, independent channels` to get the a and b channel again!

PR #1797, opened 2018-11-03
Bugfix of Bugfix (critical) : gamma.c, cast uint8 properly

aurelienpierre:
Solve the issue raised by @cryptomilk


TurboGit:
Thanks!

cryptomilk:
I can confirm that it is fixed :-)

PR #1798, opened 2018-11-03
Denoise bilateral filter

PaoloAst:
Allowing bilateral filter to be moved in the correct group

### DIFF
--- a/src/iop/bilateral.cc
+++ b/src/iop/bilateral.cc
@@ -69,9 +69,9 @@ const char *name()
   return _("denoise (bilateral filter)");
 }
 
-int groups()
+int groups() 
 {
-  return IOP_GROUP_CORRECT;
+return dt_iop_get_group("input color profile", IOP_GROUP_CORRECT);
 }
 
 int flags()


TurboGit:
Good catch! Thanks.

src/iop/bilateral.cc
```diff
@@ -69,9 +69,9 @@ const char *name()
   return _("denoise (bilateral filter)");
 }
 
-int groups()
+int groups() 
 {
-  return IOP_GROUP_CORRECT;
+return dt_iop_get_group("input color profile", IOP_GROUP_CORRECT);
```
TurboGit: the name was wrong, cut&paste... I have fixed that.

TurboGit: And there was a missing include, seems like you didn't even test to compile the code!

PaoloAst: Actually not, sorry. I'm not  experienced in DT and I didn't want to make something wrong

TurboGit: You should have said so upfront I would have paid more attention to the code. Anyway all under control and working now.

PaoloAst: You're definitely right. I will. I promise

PR #1799, opened 2018-11-04
fix undo delete module

edgardoh:
Fixes the crash when undo a module delete. Also handles better the multi-name.

### DIFF
--- a/src/libs/history.c
+++ b/src/libs/history.c
@@ -180,6 +180,25 @@ static GtkWidget *_lib_history_create_button(dt_lib_module_t *self, int num, con
   return widget;
 }
 
+static dt_iop_module_t *get_base_module(dt_develop_t *dev, const char *op)
+{
+  dt_iop_module_t *result = NULL;
+
+  GList *modules = g_list_first(dev->iop);
+  while(modules)
+  {
+    dt_iop_module_t *mod = (dt_iop_module_t *)modules->data;
+    if(strcmp(mod->op, op) == 0)
+    {
+      result = mod;
+      break;
+    }
+    modules = g_list_next(modules);
+  }
+
+  return result;
+}
+
 static GList *_duplicate_history(GList *hist)
 {
   GList *result = NULL;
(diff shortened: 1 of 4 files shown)


src/libs/history.c
```diff
@@ -193,10 +212,29 @@ static GList *_duplicate_history(GList *hist)
...
+    {
+      dt_iop_module_t *base = get_base_module(darktable.develop, old->op_name);
+      if(base)
+      {
+        params_size = base->params_size;
+      }
+      else
+      {
+        // nothing else to do
+        fprintf(stderr, "[_duplicate_history] can't find base module for %s\n", old->op_name);
```
TurboGit: Why the underscore just after [? Is that a dt style, I never paid attention to that.

src/libs/history.c
```diff
@@ -279,8 +297,12 @@ static void _pop_undo(gpointer user_data, dt_undo_type_t type, dt_undo_data_t *d
       // this fixes the duplicate module when undo: hitem->multi_priority = 0;
       if (hitem->module == NULL)
       {
-        const dt_iop_module_t *base = get_base_module(darktable.develop, hitem->multi_name);
-
+        const dt_iop_module_t *base = get_base_module(darktable.develop, hitem->op_name);
+        if(base == NULL)
+        {
+          fprintf(stderr, "[_pop_undo] can't find base module for %s\n", hitem->op_name);
```
TurboGit: same question here about the underscore

src/libs/history.c
edgardoh: Is the name of the function.
No idea if there's a standard for function naming, woudn't hurt to have one. As far as the printf, if this isn't a standard for errors that just can't happen it should be, it make it easy to see where it comes from when reported.

src/libs/history.c
edgardoh: Idem.

TurboGit:
Perfect! Fixes a recent(?) regression as I'm sure I had tested that not so long ago.

PR #1800, opened 2018-11-04
fix masks and new module crash

edgardoh:
This fixes Bug #11956 and some random crashes when using masks.
https://redmine.darktable.org/issues/11956

Basically it makes dev->iop and dev-forms thread-safe inside the pipe.

To duplicate the masks issue add several heavy masks on retouch and delete them while dt is still processing.

### DIFF
--- a/src/develop/imageop.c
+++ b/src/develop/imageop.c
@@ -564,18 +564,12 @@ static void dt_iop_gui_delete_callback(GtkButton *button, dt_iop_module_t *modul
   // we remove all references in the history stack and dev->iop
   dt_dev_module_remove(dev, module);
 
-  // we recreate the pipe
-  dt_dev_pixelpipe_cleanup_nodes(dev->pipe);
-  dt_dev_pixelpipe_cleanup_nodes(dev->preview_pipe);
-  dt_dev_pixelpipe_create_nodes(dev->pipe, dev);
-  dt_dev_pixelpipe_create_nodes(dev->preview_pipe, dev);
-
   // we cleanup the module
   dt_accel_disconnect_list(module->accel_closures);
   dt_accel_cleanup_locals_iop(module);
   module->accel_closures = NULL;
-  dt_iop_cleanup_module(module);
-  free(module);
+  // don't delete the module, a pipe may still need it
+  dev->alliop = g_list_append(dev->alliop, module);
   module = NULL;
 
   // we update show params for multi-instances for each other instances
--- a/src/develop/masks.h
+++ b/src/develop/masks.h
@@ -275,6 +275,8 @@ int dt_masks_legacy_params(dt_develop_t *dev, void *params, const int old_versio
 
 /** we create a completely new form. */
 dt_masks_form_t *dt_masks_create(dt_masks_type_t type);
+/** retrieve a form from a list of forms with is id */
+dt_masks_form_t *dt_masks_get_from_id_ext(GList *forms, int id);
 /** retrieve a form with is id */
 dt_masks_form_t *dt_masks_get_from_id(dt_develop_t *dev, int id);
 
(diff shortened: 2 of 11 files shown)


src/develop/imageop.c
```diff
@@ -549,10 +549,13 @@ static void dt_iop_gui_delete_callback(GtkButton *button, dt_iop_module_t *modul
   dt_dev_module_remove(dev, module);
 
   // we recreate the pipe
+/*
+  // pipe will be recreated below
```
TurboGit: I don't see where. It is refreshed later in the routine. In any case I would certainly remove those commented out code if it is not necessary. And add a hint where the pipe is recreated.

edgardoh: I left this just for the review, I'll remove it later.
Where it says refreshed, the effect is the same.

TurboGit:
I see hat in redmine #11956 you're talking about a version without copy. What happened to this? Was there still issues with it?

src/develop/imageop.c
TurboGit: Understood. Then better moving the message indeed.

edgardoh:
That was not a good solution, I deleted the branch.

TurboGit:
I'm sure it was not a good solution since you've deleted the branch. I just wanted to know more about this issues. Complex? Not stable? Was not fixing all cases?

TurboGit:
BTW, I'm just asking as I think it could help understand this PR and help the review.

edgardoh:
It was only for the modules and it was working, but it freeze dt while
waiting for the pipe to finish. On heavy edits that may seem forever, but
even on light edits dt seem lees responsive.
This approach is better for me, locks should be the last option, and this
I'm sure covers more scenarios and is less intrusive that adding locks
everywhere.

In order for this to be 100% safe some locks should be added anyway, I get
that, but with this alone I didn't have any crashes since I'm using it.

src/views/darkroom.c
```diff
@@ -1881,6 +1890,12 @@ void leave(dt_view_t *self)
     free(module);
     dev->iop = g_list_delete_link(dev->iop, dev->iop);
   }
+  while(dev->alliop)
+  {
+    dt_iop_cleanup_module((dt_iop_module_t *)dev->alliop->data);
```
TurboGit: can't we have multiple cleanup() for a module? this is done also just in the loop above.

TurboGit:
Forgot to say, there is conflicts to be resolved. Thanks.

src/views/darkroom.c
edgardoh: My bad, I didn't explain it right. dev->alliop don't work exactly like dev->allforms. For the masks, dev->allforms has ALL forms, and they are free() on exit, dev->alliop has only the modules that has been removed from dev->iop, and it assumes that dt_iop_cleanup_module() hasn't been called.
This is done here:
https://github.com/darktable-org/darktable/pull/1800/files#diff-2c39561107e06b0530706f865a77eabaR580

The idea is: the user delete a module, but the pipe may still need it, so, instead of physically delete it, it is removed from dev->iop (that's not an issue since the pipe works on a copy of dev->iop) and is added to dev->alliop, where it is free() on exit.

edgardoh:
Rebased.

src/develop/masks.h
```diff
@@ -328,6 +330,8 @@ void dt_masks_form_remove(struct dt_iop_module_t *module, dt_masks_form_t *grp,
 void dt_masks_form_change_opacity(dt_masks_form_t *form, int parentid, int up);
 void dt_masks_form_move(dt_masks_form_t *grp, int formid, int up);
 int dt_masks_form_duplicate(dt_develop_t *dev, int formid);
+/* duplicate the list of forms, replace item in the list with form is the same formid */
```
TurboGit: can't parse this one. "with form is the same formid" should it be "with form with the same formid"?

src/develop/masks.h
```diff
@@ -275,6 +275,8 @@ int dt_masks_legacy_params(dt_develop_t *dev, void *params, const int old_versio
 
 /** we create a completely new form. */
 dt_masks_form_t *dt_masks_create(dt_masks_type_t type);
+/** retrieve a form from a list of forms with is id */
```
TurboGit: "with is id" -> "with its id"?

Same typo below, you're not the culprit here, but if you can fix it too.

PR #1802, opened 2018-11-04
gamma.c : more robust casting

aurelienpierre:
Implement @cryptomilk suggestion to cast to uint32 first and handle the 8 bits overflows more robustly (and maybe more precisely).


TurboGit:
Indeed. Better thanks.

PR #1803, opened 2018-11-05
profile_gamma IOP : remove the dynamic range remapping

aurelienpierre:
Using the auto-tuner and color-pickers, previously we tried to remap the black and white points to the extreme values of IT8 charts (18 and 96 % in Lab luminance, thus 2.3 and 90 % in XYZ luminance).

However, this is very unreliable in most cases for mathematical reasons : the black point appears at the numerator and at the denominator of the equations we need to solve to get the proper parameters.

This could be easily fixed using a numeric optimizer as the one I have put in the colorbalance IOP. However, the grey value accuracy is not garanteed anymore, and even if having black and white points right is nice, the grey point is the most critical.

With this change, the auto-tuner and color-pickers will remap the black to 0 % and the white to 100 %, ensuring the grey is at 50 % accurately. That is, 75 % from the reading of the global color-picker which is completely wrong (bugfix attempted and reverted).

The new way to give more room to the dynamic range at extreme values is to use the security factor slider. Not great, but still good enough and certainly better than before (more consistent and reproductible).

### DIFF
--- a/src/iop/profile_gamma.c
+++ b/src/iop/profile_gamma.c
@@ -410,7 +410,6 @@ static void apply_auto_black(dt_iop_module_t *self)
   float black = fmax(fmax(self->picked_color_min[0], self->picked_color_min[1]), self->picked_color_min[2]);
   float EVmin = Log2Thres(black / (p->grey_point / 100.0f), noise);
   EVmin *= (1.0f + p->security_factor / 100.0f);
-  EVmin -= 0.0230f * p->dynamic_range;
 
   p->shadows_range = EVmin;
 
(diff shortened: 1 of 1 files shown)


TurboGit:
Works fine. Thanks.

edgardoh:
@aurelienpierre , out of curiosity, when the color picker is called from a module, does it return the value from the input image of the module or the final image? And what is the expected behaviour for the profile log and color balance?

aurelienpierre:
@edgardoh there are 3 color pickers variants : module input, module output, pipe output. In-module default is to take the input of the module, that's what colorbalance and log do.

edgardoh:
Thanks.

PR #1804, opened 2018-11-05
Undo memory leak

edgardoh:
This fixes the undo of a deleted base instance, that was not working, and a memory leak on every undo-delete, to confirm build with asan.

The redo of a deleted instance and undo of a new one still have funny effects, but that is even more tricky than this one, so better to handle it on a new PR.

There's still a memory leak on the undo, but it doesn't seem to be related to this. I'll check that later.

### DIFF
--- a/src/develop/imageop.c
+++ b/src/develop/imageop.c
@@ -545,15 +545,6 @@ static void dt_iop_gui_delete_callback(GtkButton *button, dt_iop_module_t *modul
     dt_iop_gui_cleanup_module(module);
   }
 
-  // we remove all references in the history stack and dev->iop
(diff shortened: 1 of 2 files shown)


edgardoh:
BTW, delete several instances, undo/redo fast, dt will crash.

edgardoh:
Just fixed the last memory leak.

TurboGit:
Thanks fantastic work! I'll test... To tell the truth I'm a bit nervous about this one which is touching some delicate part of the code. We need to test, I'll do that right away.

TurboGit:
Ok, tested and working fine. I did not spot issues on the code, let's merge this for a wider field testing.

PR #1805, opened 2018-11-05
enoise profile update presets

rawfiner:
This PR updates the presets for denoise profiled:
- the chroma preset now results in less isolated green pixels 
- the luma preset tries to be gentle for fine grain noise while denoising the image, and gives less artifacts than using non local means
This has been


rawfiner:
This is usable for testing but not ready for integration yet, I have to update the blobs as the current ones set the radius slider to 1 (this is visible when switching back to non local means).

rawfiner:
Blobs updated ;-)

PR #1806, opened 2018-11-05
denoise profile NLmeans: fast coarse noise reduction

rawfiner:
Replacement for the slider "radius", which impacts a lot the performance. Time of execution depends on the square of the radius. The replacement allows to have a similar visual effect without impacting performance by looking at patches more "wide-spread" without changing the number of patches. So there is no more performance issue, and the slider is much more usable in practice.

Value is set to 0 by default, which corresponds to the old behavior.

Some information here: https://github.com/darktable-org/darktable/pull/1793

### DIFF
--- a/src/iop/denoiseprofile.c
+++ b/src/iop/denoiseprofile.c
@@ -64,7 +64,7 @@ typedef enum dt_iop_denoiseprofile_channel_t
 
 // this is the version of the modules parameters,
 // and includes version information about compile-time dt
-DT_MODULE_INTROSPECTION(5, dt_iop_denoiseprofile_params_t)
+DT_MODULE_INTROSPECTION(6, dt_iop_denoiseprofile_params_t)
 
 typedef struct dt_iop_denoiseprofile_params_v1_t
 {
@@ -84,11 +84,23 @@ typedef struct dt_iop_denoiseprofile_params_v4_t
   float y[DT_DENOISE_PROFILE_NONE][DT_IOP_DENOISE_PROFILE_BANDS]; // values to change wavelet force by frequency
 } dt_iop_denoiseprofile_params_v4_t;
 
+typedef struct dt_iop_denoiseprofile_params_v5_t
+{
+  float radius;                      // patch size
+  float nbhood;                      // search radius
+  float strength;                    // noise level after equalization
+  float a[3], b[3];                  // fit for poissonian-gaussian noise per color channel.
+  dt_iop_denoiseprofile_mode_t mode; // switch between nlmeans and wavelets
+  float x[DT_DENOISE_PROFILE_NONE][DT_IOP_DENOISE_PROFILE_BANDS];
+  float y[DT_DENOISE_PROFILE_NONE][DT_IOP_DENOISE_PROFILE_BANDS]; // values to change wavelet force by frequency
+} dt_iop_denoiseprofile_params_v5_t;
+
 typedef struct dt_iop_denoiseprofile_params_t
 {
   float radius;     // patch size
   float nbhood;     // search radius
   float strength;   // noise level after equalization
+  float scattering; // spread the patch search zone without increasing number of patches
   float a[3], b[3]; // fit for poissonian-gaussian noise per color channel.
   dt_iop_denoiseprofile_mode_t mode; // switch between nlmeans and wavelets
   float x[DT_DENOISE_PROFILE_NONE][DT_IOP_DENOISE_PROFILE_BANDS];
@@ -102,6 +114,7 @@ typedef struct dt_iop_denoiseprofile_gui_data_t
   GtkWidget *radius;
   GtkWidget *nbhood;
   GtkWidget *strength;
+  GtkWidget *scattering;
   dt_noiseprofile_t interpolated; // don't use name, maker or model, they may point to garbage
   GList *profiles;
   GtkWidget *stack;
@@ -126,6 +139,7 @@ typedef struct dt_iop_denoiseprofile_data_t
   float radius;                      // search radius
   float nbhood;                      // search radius
   float strength;                    // noise level after equalization
+  float scattering;                  // spread the search zone without changing number of patches
   float a[3], b[3];                  // fit for poissonian-gaussian noise per color channel.
   dt_iop_denoiseprofile_mode_t mode; // switch between nlmeans and wavelets
   dt_draw_curve_t *curve[DT_DENOISE_PROFILE_NONE];
@@ -212,7 +226,7 @@ int legacy_params(dt_iop_module_t *self, const void *const old_params, const int
     else
       memcpy(&v4, old_params, sizeof(v4)); // was v4 already
 
-    dt_iop_denoiseprofile_params_t *v5 = new_params;
+    dt_iop_denoiseprofile_params_v5_t *v5 = new_params;
     v5->radius = v4.radius;
     v5->strength = v4.strength;
     v5->mode = v4.mode;
@@ -230,6 +244,38 @@ int legacy_params(dt_iop_module_t *self, const void *const old_params, const int
       }
     }
     v5->nbhood = 7; // set to old hardcoded default
+    return 0;
+  }
+  else if(new_version == 6)
+  {
+    dt_iop_denoiseprofile_params_v5_t v5;
+    if(old_version < 5)
+    {
+      // first update to v5
+      if(legacy_params(self, old_params, old_version, &v5, 5)) return 1;
+    }
+    else
+      memcpy(&v5, old_params, sizeof(v5)); // was v5 already
+
+    dt_iop_denoiseprofile_params_t *v6 = new_params;
+    v6->radius = v5.radius;
+    v6->strength = v5.strength;
+    v6->mode = v5.mode;
+    v6->nbhood = v5.nbhood;
+    for(int k = 0; k < 3; k++)
+    {
+      v6->a[k] = v5.a[k];
+      v6->b[k] = v5.b[k];
+    }
+    for(int b = 0; b < DT_IOP_DENOISE_PROFILE_BANDS; b++)
+    {
+      for(int c = 0; c < DT_DENOISE_PROFILE_NONE; c++)
+      {
+        v6->x[c][b] = v5.x[c][b];
+        v6->y[c][b] = v5.y[c][b];
+      }
+    }
+    v6->scattering = 0.0; // no scattering
     return 0;
   }
   return 1;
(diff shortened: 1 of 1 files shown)


rawfiner:
Just a comment about how it works:
- with radius, the algorithm iterated on [-nbhood, nbhood] interval on both axis to compare patches. The number of patches was nbhood*nbhood, thus the execution time increased with the square of nbhood.
- with default radius to 7, this means it iterated in [-7, -6, -5, -4, -3, -2, -1, 0, 1, 2, 3, 4, 5, 6, 7]
- with this PR, the algorithm iterates on [-7*7*nbhood-7, -6*6*nbhood-6, ..., 6*6*nbhood+6, 7*7*nbhood+7]. Note that it is no more a continuous interval, but an interval with gaps. With nbhood=0, we get the same list as previously. With nbhood higher, we look for patches more distant from the center than the "normal" ones. The number of patches is constant, equal to 7*7 and does not depend on nbhood.

src/iop/denoiseprofile.c
```diff
@@ -227,7 +229,7 @@ int legacy_params(dt_iop_module_t *self, const void *const old_params, const int
         v5->y[c][b] = v4.y[c][b];
       }
     }
-    v5->nbhood = 7; // set to old hardcoded default
+    v5->nbhood = 0.0; // set to old hardcoded default
```
TurboGit: Is the comment here still actual?

src/iop/denoiseprofile.c
```diff
@@ -1075,7 +1084,13 @@ static void process_nlmeans(struct dt_iop_module_t *self, dt_dev_pixelpipe_iop_t
   // adjust to zoom size:
   const float scale = fmin(roi_in->scale, 2.0f) / fmax(piece->iscale, 1.0f);
   const int P = ceilf(d->radius * scale); // pixel filter size
-  const int K = ceilf(d->nbhood * scale); // nbhood
+  const int K = ceilf(DT_IOP_DENOISE_PROFILE_NBHOOD_NORMAL_SIZE * scale);
+
+  int coordinate_offsets[DT_IOP_DENOISE_PROFILE_NBHOOD_NORMAL_SIZE];
+  for(int i = 0; i < DT_IOP_DENOISE_PROFILE_NBHOOD_NORMAL_SIZE; i++)
+  {
+    coordinate_offsets[i] = i * i * d->nbhood + i;
```
TurboGit: Ok, let me just try an assertion to check I'm reading this correctly.

- before nbhood was 7 and we are looking at the 7 neighborhood.
- now nbhood is 0, and the coordinate_offsets is "coordinate_offsets[v] = v" so we are also looking at the same 7 neighborhood.
- if nbhood is changed then we have 7 more spread points to look at (for catching coarse grain).

Are all this right?

If so, indeed we just break history with recent change and this is not an issue since it was only on a dev release. So waiting for your comment before merging.

TurboGit: I'm stupid, I just saw your second comment explaining exactly that. So I have my answer, and it is good to see that I got it right :)

TurboGit:
This is not ready. We get again lot of structured lines (grid) when the cursor for coarse-grain changed.

capture d ecran de 2018-11-05 22-25-55

I don't have this with current module. Sure if you push the cursor it may be quite CPU hungry, but at least not artifacts.

rawfiner:
Grid were normal considering the way the patches were chosen. The tooltip text indicated to reduce the slider value in that case.

I updated the code with a different formula to determine which patches are considered, which should remove the grid effect.
The new formula still guarantees that with nbhood==0 we get the behaviour of 2.4.4

For testing, I worked on my portable computer this morning, thus I could not try to compile and run the opencl version.

Yet, I have been able to test both the updated normal and sse versions.

src/iop/denoiseprofile.c
rawfiner: Good point, it is deprecated. Thanks

src/iop/denoiseprofile.c
rawfiner: Yes, you understand it right!
The problem with grids comes from the fact that we have grid gaps between the patches we consider: lines or columns without any patches, and lines or columns with a lot of patches. I updated the formula so that the patches are chosen in a way that makes the patches better distributed.

rawfiner:
Previous formula:
p1000383_05

New formula (same parameters):
p1000383_06

rawfiner:
As you can see from the code, with the new formula it is no more obvious that a couple of coordinates is used only once.
Thus, I made a python script to test the formula: it checks that we never get twice the same couple of coordinates.
It tests this for all nbhood values in [0,1] with a step of 0.001 are tested.
You can find it here if needed:
https://drive.google.com/open?id=1jFx7fdntX3HSiQnpbe4bP162nc1AMXmD

rawfiner:
Another comparison for the formula, maybe easier to see the difference.
Old formula (grid effect is very visible, it gives textures while the squares should not have texture):
dscf9740_01

New formula (same parameters. The grid effect is no more visible. Vertical and horizontal lines are less sharp though)
dscf9740

TurboGit:
Well testing again but we still get some wrong results. See the screen shot. We have two options:
- use current version, it use lot of CPU when moving search radius
- use your version, fast but tends to produce some random results

No local-means, just chroma denoise:
no-denoise

Local-means, zoom at 157%:
denoise-zoom157

Local-means, zoom at 193%:
denoise-zoom193

rawfiner:
@TurboGit, another question to help me understand, if you move slightly in the image, do the artifacts change completely?

TurboGit:
@rawfiner : no moving does not seems to change the issue. I'm on GPU. Yes it is somehow visible at 100%, but a bit more difficult to see.

> I did not understand that the artefacts were a preview issue

I don't understand this one.

rawfiner:
@TurboGit I was wondering if it was a preview issue, or if it was the same thing for export.
Could you try to reproduce with opencl disabled, so that I can know if the issue is with opencl implementation or if it is a general issue?

TurboGit:
This is frightening :) I can't reproduce the issue now with or without OpenCL, but the screen shot are there, I have set the very same value... No grid, no banding issue... Frightening! There is something wrong going on.

rawfiner:
That's weird ;-(
By the way, I am trying to improve the formula to get better a better distribution of the patches.
Currently, the distribution is like this for 0.99 (it looks better for smaller values though):
patch_repartition

rawfiner:
I just changed the formula. The new one behaves better for high values:


This gif shows how the patches positions are affected when we increase the slider by steps of 0.01:
animation

I have to test this on my fedora docker, I did not have time to do it yet. I tested sse and normal version on linux mint though (and the opencl one, but on cpu).

aurelienpierre:
@rawfiner I don't have time to follow all your adjustements (too busy working in my own features, sorry), but non-local means are intended to work on textures (repetable patterns), so providing examples of plain surfaces images doesn't help much to evaluate the efficiency of the code.

rawfiner:
@aurelienpierre The goal of the plain surface images is to check that the modification does not induce grid artifacts, which were very visible for high values of the parameter on plain surface, but much less visible on textured surface. That's why I used examples with plain surface images.

rawfiner:
Here are some tests on a more textured image (with patch size = 4, force = 2)
original with wavelets in color mode:
p1920766_07

denoised without the coarse noise slider, nor the radius slider:
execution time: 1.9s on CPU
p1920766_05

denoised with coarse noise slider set to maximum (note that we should have put it to a smaller value, a value of 0.3 is enough to remove coarse grain noise on this image, I set it to the maximum to try to stress the module, and to see potential effects of having less patches close to the center, see bellow for what we can have with smaller values):
execution time: 2.0s on CPU
p1920766_06

denoised with radius slider, changed the force and radius to get close to the result of what I got with coarse noise slider (radius is 20, force is 1.2)
execution time: 15s on CPU
p1920766_08

denoised with coarse noise slider again, trying to get closer to the result I had with radius slider (coarse slider set to 0.3 and force to 1.9)
execution time: 2s on CPU
p1920766_11

TurboGit:
Ok, tested again with different picture and seems better to me now. I had no grid-like pattern. I want to test it again later today with another picture. If all goes well I'll merge today.

rawfiner:
@TurboGit thanks for the testing.
I still need to test on fedora before the merging

TurboGit:
@rawfiner : this is still not working, on my side I have a large band of blur at the bottom of the image (bottom of the lighttable view). See screenshot above.

capture d ecran de 2018-11-08 17-35-06

At this stage I'm not sure it is safe to have this included into 2.6. Maybe next cycle?

rawfiner:
@TurboGit I will take a look at this. I am ok for a possible integration later than 2.6 in a next cycle

rawfiner:
I have been able to reproduce this band of blur. It seems that it affects only the opencl version. Thank you for reporting this issue

rawfiner:
This bug is already here in the current version
I used patch size 0, radius 20 and force 4 on this image:
degrade

Here is the result without opencl:
degrade_02

Here is the result with opencl:
degrade_03

Here is the difference (enhanced with gimp):
degrade_difference

As you can see, the difference is not only on borders, but in fact it is also visible inside the image.
I will try to find the cause of this. This is maybe caused by the loop on j which goes from -K to 0 on opencl code, it should maybe go from -K to K (did not have time to test this yet. I will test it as soon as possible).

upegelow:
>  it seems more complicated than that. I have to understand fully the opencl code to try to find the issue.

Please read the publication that is quoted in the header of denoiseprofile.cl. The OpenCL algorithm collects data in a different way than the sliding window CPU algorithm. Accumulating rounding errors play a role here and I would not expect bitwise identical results. However, the border issue looks strange indeed.

Concerning the loop: going from -K to 0 takes advantage of the symmetry and is key for GPU performance.

rawfiner:
@upegelow thank you for this information. I will read the publication.
The fact that accumulating rounding errors leads to the small differences in the center perfectly makes sense, and I did not think about this. Then, the problem is only for borders.
This narrows the search space to be explored to find the source of the issue.
Thanks!

rawfiner:
I opened a pull request here to fix this: https://github.com/darktable-org/darktable/pull/1834
Thanks again @TurboGit for finding the issue, and @upegelow for your advices.

rawfiner:
I just rebased the changes, but I did not test it on gpu yet. I will do it soon

rawfiner:
Tested again both on CPU and GPU. Note that testing on GPU will be easier once #2000 is integrated.

TurboGit:
@rawfiner : #2000 is now integrated. let me know if this is ready or not. thanks.

rawfiner:
Thanks @TurboGit
This PR needs to be rebased on master, apart of that it is ready.

TurboGit:
Tested and I cannot reproduce the issues we had before. Seems like this is good to go in. Thanks.

I'll just use  "scattering (coarse-grain noise)" for label instead of "scattering (coarse-grain noise reduction)" which is too long and mangle the values at the end of the slider. Hope this is fine with you.

src/iop/denoiseprofile.c
```diff
@@ -2705,11 +2797,15 @@ void gui_init(dt_iop_module_t *self)
                                          "wavelets work best for `color' blending"));
   gtk_widget_set_tooltip_text(g->radius, _("radius of the patches to match. increase for more sharpness"));
   gtk_widget_set_tooltip_text(g->nbhood, _("emergency use only: radius of the neighbourhood to search patches in. increase for better denoising performance, but watch the long runtimes! large radii can be very slow. you have been warned"));
+  gtk_widget_set_tooltip_text(g->scattering,
+                              _("scattering of the neighbourhood to search patches in. increase for better "
+                                "coarse-grain noise reduction. does not affect execution time."));
```
TurboGit: I just notice that... "does not affect execution time", using this slider I had quickly a large slowdown with values like 15. Is this comment still valid?

TurboGit: Forget this question, I was using another slider! This one indeed does not increase the execution time.

rawfiner: No problem :-)

PR #1807, opened 2018-11-07
upgrade undo

edgardoh:
This modifies the history undo to use the same schema than PR  #1800

Now it works on the snapshot of dev->history and a copy of dev->iop, and when is done it replaces it.

It handles the delete of modules on dev->iop, that have to be done when the user undo a new multi-instance or redo a delete multi-instance, as there will be an extra module on dev->iop.
To do this I assume that any multi_priority > 0 must be in history and there must be only one multi_priority = 0 per operation.
If I find a multi_priority > 0 not in history, I delete it.
If I find two multi_priority = 0 for the same operation, one is in history and the other is not, I delete the one that is not in history.

It also handles the move up/down of a multi-instance, and fixes some small issues.

### DIFF
--- a/src/develop/develop.c
+++ b/src/develop/develop.c
@@ -799,6 +799,7 @@ void dt_dev_pop_history_items(dt_develop_t *dev, int32_t cnt)
     memcpy(module->params, module->default_params, module->params_size);
     memcpy(module->blend_params, module->default_blendop_params, sizeof(dt_develop_blend_params_t));
     module->enabled = module->default_enabled;
+    module->multi_name[0] = '\0';
     modules = g_list_next(modules);
   }
   // go through history and set gui params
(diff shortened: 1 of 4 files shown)


edgardoh:
Travis CI is running for a day now, I don't what's going on...

TurboGit:
I've tested that, it works great on my side. That's a very nice enhancement/fix. Thanks.

edgardoh:
Thanks.

A bit late, but here is my testing:
Add a bunch of multi-instances with parametric mask, reorder them, delete
them all, go crazy with the cntrl+z / cntrl+y
Same with drawn mask and the retouch
Add some style and go back and forward in history.

No crashes and the values are restored ok.

I can't say that I have tested every single scenario, and to keep track of
the values while modules are being reordered, deleted and added is not
easy, but as far as I can tell it works ok.

PR #1808, opened 2018-11-07
cmake: Fix RPATH so that you can execute darktable in build dir

cryptomilk:
This will allow that you can run darktable directly in the build directory:

./src/darktable

will just work! The binaries will be relinked during installation and the installation rpath set.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -286,10 +286,20 @@ else(APPLE)
 endif(APPLE)
 
 if(NOT WIN32)
-  # Windows doesn't know the concept of RPATHs :(
-  set(CMAKE_BUILD_WITH_INSTALL_RPATH TRUE)
-  set(CMAKE_INSTALL_RPATH_USE_LINK_PATH FALSE)
-  set(CMAKE_INSTALL_RPATH ${RPATH_DT}/../${CMAKE_INSTALL_LIBDIR}/darktable)
+    # Windows doesn't know the concept of RPATHs :(
+
+    # use, i.e. don't skip the full RPATH for the build tree
+    set(CMAKE_SKIP_BUILD_RPATH  FALSE)
+
+    # when building, don't use the install RPATH already
+    # (but later on when installing)
+    set(CMAKE_BUILD_WITH_INSTALL_RPATH FALSE)
+
+    set(CMAKE_INSTALL_RPATH ${RPATH_DT}/../${CMAKE_INSTALL_LIBDIR}/darktable)
+
+    # add the automatically determined parts of the RPATH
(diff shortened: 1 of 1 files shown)


TurboGit:
I see, this makes sense but is a very delicate part. I'll merge now, hopefully this wont break the different builds.

LebedevRI:
Looks like the shiny days has passed, and the time of breakage has come.
This broke openSUSE_Leap_42.3 x86_64 build
https://build.opensuse.org/package/live_build_log/graphics:darktable:master/darktable/openSUSE_Leap_42.3/x86_64

I *really* like this, but does this *actually* work though,
with all the loadable plugins and such being loaded from build dir, and not from some install?

TurboGit:
@LebedevRI : as said above I fear that this was a delicate issue and would broke some platforms. if I see no solution I will revert.

TurboGit:
Ok, let's revert this. I don't have Open Suse to test anyway.

cryptomilk:
@LebedevRI What was the error? The build log has been overwritten.

LebedevRI:
Something about RPATH. I don't know where to look for previous build logs.
https://build.opensuse.org/packages/darktable/job_history/graphics:darktable:master/openSUSE_Leap_42.3/x86_64

cryptomilk:
I've applied the patch to the packet but it builds just fine for me:

https://build.opensuse.org/package/show/home:gladiac:branches:graphics:darktable:master/darktable

LebedevRI:
Uh, no?
https://build.opensuse.org/package/live_build_log/home:gladiac:branches:graphics:darktable:master/darktable/openSUSE_Leap_42.3/x86_64
```
[  537s] CMake Error at src/cmake_install.cmake:45 (file):
[  537s]   file RPATH_CHANGE could not write new RPATH:
[  537s] 
[  537s]     $ORIGIN/../lib64/darktable
[  537s] 
[  537s]   to the file:
[  537s] 
[  537s]     /home/abuild/rpmbuild/BUILDROOT/darktable-2.5.0~git852.71efab7eb-494.1.x86_64/usr/lib64/darktable/libdarktable.so
[  537s] 
[  537s]   No valid ELF RPATH or RUNPATH entry exists in the file;
[  537s] Call Stack (most recent call first):
[  537s]   cmake_install.cmake:38 (include)
```

cryptomilk:
Question:

Why is this needed in the first place?
And for which binary is that needed exactly? Only for the darktable executable?

cryptomilk:
I guess we want this:

https://github.com/cryptomilk/darktable/commit/f89a5fb356fb2f9f3697a2c7aa306fe233e17fe2

PR #1809, opened 2018-11-07
blend mask feathering by a guided filter

rabauke:
Feathering adjusts the blend mask by taking into account the image content
such that edges of the mask match those in the image.

Here a short demo. First make a rough selection via drawn or parametric masks:

Then select "feathering" as "mask enhancement" and adjust radius, brightness and contrast such that edges of the mask match those in the image.


TurboGit:
Can you give a bit more information about the way to use it? It will help testing... Thanks.

aurelienpierre:
@TurboGit It comes as an addition to gaussian blur within the blending parameters, and works when you have semi-transparent masks (parametric and/or drawn). Using ligthness, contrast, and blur radius, it allows you to refine masks edges. It's quite awesome actually.

data/kernels/blendop.cl
```diff
@@ -1007,6 +1007,36 @@ blendop_rgb (__read_only image2d_t in_a, __read_only image2d_t in_b, __read_only
   write_imagef(out, (int2)(x, y), o);
 }
 
+__kernel void
+blendop_mask_enhance_contrast(__read_only image2d_t mask_in, __write_only image2d_t mask_out,
+                              const int width, const int height,
+                              const float e, const float brightness) {
```
TurboGit: { should be at the start of a new line

src/develop/blend_gui.c
```diff
@@ -2045,15 +2069,51 @@ void dt_iop_gui_init_blending(GtkWidget *iopw, dt_iop_module_t *module)
...
-
+    bd->brightness_slider = dt_bauhaus_slider_new_with_range(module, -1.0, 1.0, 0.01, 0.0, 2);
+    dt_bauhaus_widget_set_label(bd->brightness_slider, _("blend"), _("mask brightness"));
+    dt_bauhaus_slider_set_format(bd->brightness_slider, "%.2f");
+    gtk_widget_set_tooltip_text(bd->brightness_slider, _("brightness adjustment of blend mask"));
+    g_signal_connect(G_OBJECT(bd->brightness_slider), "value-changed",
+                     G_CALLBACK(_blendop_blendif_brightness_callback), bd);
+
+
+    bd->brightness_slider = dt_bauhaus_slider_new_with_range(module, -1.0, 1.0, 0.01, 0.0, 2);
```
TurboGit: we have two section identical, the code above is the same as this one.

src/develop/blend_gui.c
```diff
@@ -2045,15 +2069,51 @@ void dt_iop_gui_init_blending(GtkWidget *iopw, dt_iop_module_t *module)
                      G_CALLBACK(_blendop_masks_invert_callback), bd);
 
 
+    bd->masks_blur_modes_combo = dt_bauhaus_combobox_new(module);
+    dt_bauhaus_widget_set_label(bd->masks_blur_modes_combo, _("mask enhancement"), _("mask enhancement"));
+
+    dt_bauhaus_combobox_add(bd->masks_blur_modes_combo, _("Gaussian blur"));
```
TurboGit: No capital letters in UI label (G -> g).

TurboGit:
Also as we are very close from the feature freeze, can you describe more the actual algorithm behind this? To be clear my fear is that is not full ready and if we want to improve later we'll break history or we would have to support multiple algorithm on the blend layer. Not something we have at the moment. I would have like to see that earlier to have more field testing. Now if the algorithm are safe or state of the art I think it can go. All in all I'm seeking more background information about this.

rabauke:
The algorithm is described in detail in the references given in the sources, see also http://kaiminghe.com/eccv10/index.html
The suggested mask filtering is based on a so-called guided filter, which is also employed in the haze removal module. More testing and feedback would be highly appreciated.

Kabouik:
I have been waiting for this feature since I saw [this message](https://www.mail-archive.com/darktable-dev@lists.darktable.org/msg02042.html) from @rabauke on the list a long time ago. Eagerly looking forward to trying this when it is merged, and quite happy to see that the project has not been abandoned. Thanks for your work, hope merging won't require too much additional sweat.

TurboGit:
@rabauke : and what about the feathered mode which seems not working with parametric masks? Can you describe the way to use it?

TurboGit:
@Kabouik : if you can test it it would be great.

TurboGit:
@rabauke : I can't make this working... sorry I really need more information about the way it is supposed to be used. Do you have an example?

aurelienpierre:
@TurboGit 

the feathered mode only works with semi-transparent regions (opacity < 100 %).

* the mask blur controls the spread of the edge detection
* the mask brightness controls how much you retain or exclude the outside of the mask
* the mask contrast controls the sharpness/feathering of the edges of the mask.

@rabauke I'm not sure the sliders labels are super clear there. I would to go for:

* mask edge detection radius
* mask edge offset
* mask edge contrast

rabauke:
> @rabauke : and what about the feathered mode which seems not working with parametric masks? Can you describe the way to use it?

There was an issue in the OpenCL code path. Now it works also with parametric masks (and OpenCL).

TurboGit:
Ok, thanks for the example and the code fix! I'll have a look again.

TurboGit:
@rabauke : please can you remove rawspeed update from the following commit: e953a6

TurboGit:
@rabauke : not sure you've seen but I have request some changes.

rabauke:
@turbogit Issues have already been solved, the rawspeed issue will be fixed soon.

TurboGit:
Another question, with the default parameters the blend is 100% identical to previous version. Right?

rabauke:
@TurboGit : There have not been any recent parameter changes. Do you have any specific in mind. Furthermore, I removed the rawspeed sources from the pull request. It was not intended to include any rawspeed stuff.

TurboGit:
@rabauke : my question was more to ensure that the current implementation does not break history and that the default behavior without the two new sliders is 100% equal to the current implementation. That's what I think from reading the code, but just wanted to be sure.

rabauke:
> @rabauke : my question was more to ensure that the current implementation does not break history and that the default behavior without the two new sliders is 100% equal to the current implementation. 

It was always my aim not to break history when implementing this new feature. It's crucial indeed.

TurboGit:
Ok, thanks! Let's go with it and have more field testing. Thanks!

moy:
This seems to work only for modules between the input and output color profile modules, see https://redmine.darktable.org/issues/12429

Cc: @rabauke, in case you're not subscribed to the bugtracker.

rabauke:
@moy Thanks for the bug report. Can reproduce the behavior. Will fix it as soon as possible.

PR #1810, opened 2018-11-08
Fix OpenCL

cryptomilk:
This fixes OpenCL with ROCm for me.


cryptomilk:
@aurelienpierre

TurboGit:
Sound good! We had discussed that with Aurélien some time ago and were surprised that the const was accepted here.

aurelienpierre:
Thanks @cryptomilk

aurelienpierre:
Just out of curiosity, what was broken ? @cryptomilk

TurboGit:
It was a compilation issue.

PR #1811, opened 2018-11-09
New module : Filmic

aurelienpierre:
This is based on the work of @sobotka who wrote the original Blender Filmic plugin and has been a great help for this work. Many thanks to him.

what ?
---

This new module is the log profile + tonecurve boundled at once in order to perform a quick, efficient and powerful dynamic-range compression with contrast addition, allowing to remap the middle grey value to a specified value.

capture d ecran du 2018-11-08 23-36-33

how ?
---

The settings are put in the same order they should be used.

The first group is directly issued from the unbreak profile module. It does a logarithmic remapping of the dynamic range, helped with auto-tuners. Auto-tune will do a good job 85 % of the time. You goal, setting this group, is not to enhance the picture, but adjust the bounds of the log. Don't strive for beauty in there.

The second group is the set of controls that will internally build an S-shaped "filmic curve". The controls are converted in nodes by which we create a curve using the same API as the tone curve module. The curves we try to create are shaped like that:

capture d ecran du 2018-11-07 21-02-06

Good values for the contrast are usually between 1.5 and 2. The contrast is the slope of the linear middle part of the S. The latitude is the length of that linear part. More latitude will give more contrast at the extreme values of the histogram. The latitude is bounded between 25 % and 95 % of the dynamic range.

The balance shadows/highlights will translate the linear part toward the highlights or the shadows so one extreme has more room than the other. This allows to preserve the details in highligts or in shadows.

The main advantage of that module is that, no matter the contrast you set, the grey point will always be remaped to the input value, and the extreme values (black and white) are never affected once properly remapped through the log.

The saturation allows to desaturate if needed, to compensate for the artifacts that contrast manipulations can create in shadows (over-saturation). It can also be used to produce a pleasing black and white, based on the linear luminance.

The interpolation method is an emergency bypass for setups where the interpolation fails (you will see the contrast inverted). Cubic splines can give very pleasant but very unpredictible results, while monotonic splines are robust but sometimes too contrasty. Centripetal splines are good overall but can lack a bit of contrast.

The last group is related to the destination color-spaces. For classic Adobe RGB, sRGB etc. the defauts values work. They are usefull only for exotic gamma 1.8 color spaces and such. Grey and power function of the destination should never be touched in standard use, while the black and white luminance can be used to create a "retro" faded look if you like.

The suggested workflow is:

1. exposure
1. white balance
1. color profiles / LUT
1. filmic (tones)
1. color balance (colors)
1. denoising and enhancements as you please.

what's in ?
---

The modules comes before levels, tone curves, etc. to allow more precise control afterwards. The code comes mostly out of `src/iop/tonecurves.c` and `src//iop/profile_gamma.c`. The module has pure C, SSE and OpenCL versions.

why ?
---

The main benefit of this module is it's really easy and fast to setup. It can also be used with parametric blending and mask feathering as a better replacement than "shadows/highlights". It produces none to very little out-of-gamut colors.

It can replace (global) tonemap, base curves, shadows and highlights. It can be used in conjunction with tone curves, local contrast, etc.

screenshot or it didn't happen !
---

Before:
capture d ecran du 2018-11-09 00-01-34

After:
capture d ecran du 2018-11-09 00-02-36

Before: (picture by @cryptomilk)
capture d ecran du 2018-11-09 00-04-31

After:
capture d ecran du 2018-11-09 00-05-32

After: (use with the awesome guided filter of @rabauke from #1809 )
capture d ecran du 2018-11-09 00-15-20

Before:
capture d ecran du 2018-11-09 00-09-24

After : filmic + 2 instances of color balance with shipped orange/teal presets
capture d ecran du 2018-11-09 00-09-07

Note : the screenshots are produced on a screen with Adobe RGB profile, so colors might be desaturated.

### DIFF
--- a/data/kernels/programs.conf
+++ b/data/kernels/programs.conf
@@ -22,3 +22,4 @@ basecurve.cl            18
 locallaplacian.cl       19
 dwt.cl                  20
 retouch.cl              21
+filmic.cl               22
--- a/src/common/sse.h
+++ b/src/common/sse.h
@@ -16,7 +16,7 @@
 
 #include <xmmintrin.h>
 
-/*
+/**
  * Fast SSE2 implementation of special math functions.
  */
 
@@ -120,6 +120,22 @@ static inline __m128 _mm_pow_ps(__m128 x, __m128 y)
 static inline __m128 _mm_pow_ps1(__m128 x, float y)
 {
   return _mm_exp2_ps(_mm_mul_ps(_mm_log2_ps(x), _mm_set1_ps(y)));
+}
+
+
+/**
+ * Allow access of the content of a SSE vector
+ **/
+
+static inline float _mm_vectorGetByIndex( __m128 V, unsigned int i)
+{
+  union {
+    __m128 v;
+    float a[4];
+  } converter;
+
+  converter.v = V;
+  return converter.a[i];
 }
 
 // modelines: These editor modelines have been set for all relevant files by tools/update_modelines.sh
--- a/src/gui/draw.h
+++ b/src/gui/draw.h
@@ -195,9 +195,19 @@ static inline void dt_draw_curve_calc_values(dt_draw_curve_t *c, const float min
   c->csample.m_outputRes = 0x10000;
   CurveDataSample(&c->c, &c->csample);
   if(x)
+  {
+#ifdef _OPENMP
+#pragma omp parallel for SIMD() default(none) shared(x) schedule(static)
+#endif
     for(int k = 0; k < res; k++) x[k] = k * (1.0f / res);
+  }
   if(y)
+  {
+#ifdef _OPENMP
+#pragma omp parallel for SIMD() default(none) shared(y, c) schedule(static)
+#endif
     for(int k = 0; k < res; k++) y[k] = min + (max - min) * c->csample.m_Samples[k] * (1.0f / 0x10000);
+  }
 }
 
 static inline float dt_draw_curve_calc_value(dt_draw_curve_t *c, const float x)
--- a/src/iop/CMakeLists.txt
+++ b/src/iop/CMakeLists.txt
@@ -123,7 +123,7 @@ add_iop(denoiseprofile "denoiseprofile.c")
 add_iop(defringe "defringe.c")
 add_iop(ashift "ashift.c")
 add_iop(hazeremoval "hazeremoval.c")
-add_iop(toneequalizer "filmic.c" DEFAULT_VISIBLE)
+add_iop(filmic "filmic.c" DEFAULT_VISIBLE)
 
 if(RSVG2_FOUND)
   add_iop(watermark "watermark.c")
--- a/src/iop/CMakeLists.txt
+++ b/src/iop/CMakeLists.txt
@@ -123,7 +123,7 @@ add_iop(denoiseprofile "denoiseprofile.c")
 add_iop(defringe "defringe.c")
 add_iop(ashift "ashift.c")
 add_iop(hazeremoval "hazeremoval.c")
-add_iop(toneequalizer "toneequalizer.c" DEFAULT_VISIBLE)
+add_iop(toneequalizer "filmic.c" DEFAULT_VISIBLE)
 
 if(RSVG2_FOUND)
   add_iop(watermark "watermark.c")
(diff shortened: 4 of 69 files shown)


rabauke:
Also the module has its roots in a program called filmic (as far as I understand) I would prefere to give the module a more significant name which gives the user a hint what the module is good for. What about _tone equalizer_?

aurelienpierre:
@rabauke as you see, tone equalizer was my first idea (and the name of the branch). But it is not really an equalizer (which implies having a setting for each frequency). Filmic curves are a well-known thing in the imaging industry (just google it, 72600 results) so I would prefer to keep it that way.

TurboGit:
@aurelienpierre : the "balance shadows-highlights" seems working on the opposite order. It says give more room to shadows or highlight. When I push to the right to 33%, I would have expected to make more room by 33% to highlight, but it turn out that the shadows was more present in the picture.

aurelienpierre:
@TurboGit if you give more room to highlights, that means you have to compress the shadows, isn't it ? What it does is moving all the control points of the curve on the x axis. Positive % is moving to the left (compress the shadows, decompress the highlights), and the other way around. The effect of the decompression is not obvious if you have set the white point in the log tone-mapping too far (being over-conservative). Also, that's the behaviour you get with monotonic and centripetal splines, but cubic spline blow in your face once in a while, and you get sometimes the opposite of what you asked (that sayed, the contrast is gorgeous when set right).

TurboGit:
Just wonderful work! Thanks a lot Aurélien for your work on this.

LebedevRI:
This broke aarch64 build.
https://build.opensuse.org/package/live_build_log/graphics:darktable:master/darktable/openSUSE_Tumbleweed_ARM/aarch64
Please either fix quickly, or revert.

aurelienpierre:
So I get the error is:

```
[  537s] In file included from /home/abuild/rpmbuild/BUILD/darktable-2.5.0~git844.47c68b30b/src/iop/filmic.c:25,
[  537s]                  from /home/abuild/rpmbuild/BUILD/darktable-2.5.0~git844.47c68b30b/build/src/iop/introspection_filmic.c:51:
[  537s] /home/abuild/rpmbuild/BUILD/darktable-2.5.0~git844.47c68b30b/src/common/sse.h:17:10: fatal error: xmmintrin.h: No such file or directory
[  537s]  #include <xmmintrin.h>
[  537s]           ^~~~~~~~~~~~~
[  537s] compilation terminated.
[  537s] make[2]: *** [src/iop/CMakeFiles/filmic.dir/build.make:71: src/iop/CMakeFiles/filmic.dir/introspection_filmic.c.o] Error 1
[  537s] make[2]: Leaving directory '/home/abuild/rpmbuild/BUILD/darktable-2.5.0~git844.47c68b30b/build'
[  537s] make[1]: *** [CMakeFiles/Makefile2:4190: src/iop/CMakeFiles/filmic.dir/all] Error 2
[  537s] make[1]: *** Waiting for unfinished jobs....
```
The library <xmmintrin.h> seems to be missing on you system, which is super weird because that dependency was introduced in d99df3190f (Tobias Ellinghaus 2017-09-12 22:39:39 +0200  17) and is used also in common/inline_colorspaces.h, which is used multiple times in IOPs.

LebedevRI:
aarch64 is not x86 architecture, and does not have sse/avx, and naturally that header does not exist there.

aurelienpierre:
So how did that not break before ? I didn't add that lib.

aurelienpierre:
ahhhh my bad, precompiler stuff:
```
#ifdef __SSE2__
#include "common/sse.h"
#include <xmmintrin.h>
``` 

I fix that in 5 min.

LebedevRI:
No idea, it's your code, you should know how it works.
https://build.opensuse.org/package/binaries/graphics:darktable:master/darktable/openSUSE_Tumbleweed_ARM
contains `2.5.0~git814.ddddb759c` as the last successful build, that was ddddb759c
The next build was https://build.opensuse.org/package/rdiff/graphics:darktable:master/darktable?linkrev=base&rev=475 `2.5.0~git828.c3d3dce22`
c3d3dce22

LebedevRI:
Sweet, that did it, thanks.

TurboGit:
Great to see this fixed. Thanks.

PR #1813, opened 2018-11-09
faster lighttable

TurboGit:
I'd like to have this re-discussed and tested.

The idea is to skip as much as possible event to make the lighttable faster on 4k (or more) screens. This is NOT a final solution as the best path would be to redesign and rewrite from scratch the lighttable. But this won't happen soon, and we have report that the lighttable is slow on large screen...

All in all I'm wondering if this should not go in for 2.6.

This needs lot of testing. And please if you have tested on 4k (or sup) screen can you tell if this improves the responsiveness of the lighttable or not. I don't have a 4k screen so cannot tell.

### DIFF
--- a/src/views/lighttable.c
+++ b/src/views/lighttable.c
@@ -727,7 +727,7 @@ end_query_cache:
                 while (loop_count--)
                 {
                   // ex shift + down toggle selection on images_in_row images
-                  to_toggle =  idx+(-1*lib->key_jump_offset/abs(lib->key_jump_offset)*loop_count);
+                  to_toggle = idx+(-1*lib->key_jump_offset/abs(lib->key_jump_offset)*loop_count);
                   if (query_ids[to_toggle])
                     dt_selection_toggle(darktable.selection, query_ids[to_toggle]);
                 }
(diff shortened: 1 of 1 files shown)


TurboGit:
I understand, we are very close from feature freeze.

TurboGit:
@cryptomilk : done, make sure to update the faster-lighttable branch as I have fixed a refresh when changing collection. TIA for testing.

TurboGit:
On a 4k display?

TurboGit:
Also, what should be really faster is the mouse over and the follow of the highlighting of the picture under the mouse pointer. If think this is more visible when you have more than 10 images per row.

cryptomilk:
I have an Eizo 4K Monitor (3840 x 2160 resolution)

If I move the mouse (showing 12 images per row) it takes ~1sec till the selection follows.

TurboGit:
Ok, 1s is not good but how much time did it take before? About the same time? If so I'll just close and forget this PR :)

cryptomilk:
If I make the window much smaller, it is almost instant.

cryptomilk:
And it is faster with your patch than with master. However only in a small window. It doesn't feel like there is any improvement if I make it fullscreen.

TurboGit:
Ok, too bad I had hoped :) Closing. Thanks for testing on a 4k.

TurboGit:
Another attempt here: https://github.com/darktable-org/darktable/pull/1973

If someone can test and report the perf.

PR #1814, opened 2018-11-09
filmic: disable color picker icon when 'auto optimized' is clicked

moy:
it's only a cosmetic change (disable the icon), but there's a more serious issue underneath (not fixed by this PR): if the colorpicker was previously active, the old area is used because we don't wait for the pipeline to be reprocessed hence the colorpick area is not updated.

### DIFF
--- a/src/iop/filmic.c
+++ b/src/iop/filmic.c
@@ -481,6 +481,20 @@ static void apply_auto_white_point_source(dt_iop_module_t *self)
   dt_dev_add_history_item(darktable.develop, self, TRUE);
 }
 
+static void disable_colorpick(struct dt_iop_module_t *self)
+{
+  dt_iop_filmic_gui_data_t *g = (dt_iop_filmic_gui_data_t *)self->gui_data;
+  self->request_color_pick = DT_REQUEST_COLORPICK_OFF;
+  g->which_colorpicker = DT_PICKPROFLOG_NONE;
+}
+
+static void set_colorpick_state(dt_iop_filmic_gui_data_t *g, const int which_colorpicker)
+{
+  dt_bauhaus_widget_set_quad_active(g->grey_point_source, which_colorpicker == DT_PICKPROFLOG_GREY_POINT);
+  dt_bauhaus_widget_set_quad_active(g->black_point_source, which_colorpicker == DT_PICKPROFLOG_BLACK_POINT);
+  dt_bauhaus_widget_set_quad_active(g->white_point_source, which_colorpicker == DT_PICKPROFLOG_WHITE_POINT);
+}
+
 static void security_threshold_callback(GtkWidget *slider, gpointer user_data)
 {
   dt_iop_module_t *self = (dt_iop_module_t *)user_data;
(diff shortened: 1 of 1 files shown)


TurboGit:
Thanks!

aurelienpierre:
I think it's the same in the color balance and in the profile_gamma. That's new.

moy:
> I think it's the same in the color balance and in the profile_gamma. That's new.

Indeed, and this probably explains why I had difficulty understanding the behavior of these buttons ;-). Re-testing and now it seems "optimize luma" turns most of my images into plain white unconditionally.

See my message on darktable.fr: I think these buttons should actually be actual color pickers too. If we go this way, this solves all UI problems at once.

aurelienpierre:
The intended use-case of the auto-tune buttons is to have a one-click setting to get a readable picture, while being able to retake full manual control i needed. Having a color-picker for that is redundant with the separate color pickers and defeats the purpose of an all-in-one quick optimizer.

I noticed the issue with the image turning white, that's only since a couple of days, so maybe someting was changed in the color-picker API.

moy:
> Having a color-picker for that is redundant with the separate color pickers and defeats the purpose of an all-in-one quick optimizer.

It's not: the color-picker would chose the whole image and be OK by default. There would be no need (but a possibility) to actually chose an area of the picture. It'd also be a 1-click setting.

Actually, currently it's already a 2-clicks settings since you get "wait for the preview to be updated" on first click.

aurelienpierre:
> It's not: the color-picker would chose the whole image and be OK by default. There would be no need (but a possibility) to actually chose an area of the picture. It'd also be a 1-click setting.

ok then.



> Actually, currently it's already a 2-clicks settings since you get "wait for the preview to be updated" on first click.

I really don't understand what's going on with the `dt_dev_reprocess_all(self->dev)`. Is that an asynchronous function ? Why aren't why able to force it to execute and finish before grabbing the picker's reading ?

moy:
> I really don't understand what's going on with the dt_dev_reprocess_all(self->dev). Is that an asynchronous function ?

In my understanding it is, but I still don't get all the logic behind the API (and I didn't really find much documentation :-().

> Why aren't we able to force it to execute and finish before grabbing the picker's reading ?

In any case, it wouldn't be a good thing if you could. It's a bad idea to do anything non-trivial in a callback synchronously in a GUI: it would freeze the GUI.

Here, what you'd want before doing the auto-tuning is not to re-launch the whole pipeline anyway: best would be to launch the part of the pipeline before the module, then get the result of the colorpicker on the input image, do the autotuning, and finally let the rest of the pipeline execute. I *think* (but may be wrong) this is what happens when using the colorpicker normally.

What I tried yesterday was to let the button's callback just register the colorpicker, and catch the flow later in a callback. I tried putting the rest of the code in `process` but it depends on the pipeline used (SSE Vs normal Vs OpenCL). It should work if you duplicate the call in all the `process` variants. If we switch to a real colorpicker this won't be needed anyway.

aurelienpierre:
>  I tried putting the rest of the code in `process` but it depends on the pipeline used (SSE Vs normal Vs OpenCL). It should work if you duplicate the call in all the `process` variants. If we switch to a real colorpicker this won't be needed anyway.

What code did you put there ? process() functions should not contain any GUI stuff nor parameters settings. process can be called fron darktable-cli with no graphic environnement, and it's not good to set parameters in there.

moy:
I guess this comment was meant for #1818, I'll continue the discussion there.

PR #1815, opened 2018-11-09
Filmic :  Fix the computations for the shadows/offset balance 

aurelienpierre:
Use a proper vector (x, y) translation for the latitude with true cartesian coordinates.
Again, I shouldn't be allowed near a keyboard before solving the equations on paper first. :face_with_head_bandage: 

Also, rename the interpolation methods with the result they tend to produce:

* faded
* contrasted
* linear

instead of cubic, monotonic and centripetal splines (although their names were geeky and cool).

Also, since the cubic interpolation is very contrasty and the monotinic one is rather dull, I added a new mode (optimized) that average both, as a trade-off. The computations are slower, so it's an additional option. 

Finally, remove the grey node which often overconstrains the problem and creates cusps.


TurboGit:
Sounds good! Thanks.

PR #1817, opened 2018-11-10
 BUGFIX : filmic/compilation on non-AVX/SSE systems 

aurelienpierre:
Use the pre-processor magic to not load `sse.h` on non-SSE architectures.


aurelienpierre:
> You really might want to start using `git rebase` :/

I know, but you got your fix in 8 minutes.

PR #1818, opened 2018-11-10
Filmic auto picker

moy:
This turns the "autotune" button into a picker. While doing this I noticed that pickers weren't activated immediately when active, hence this PR changes that too. Also, color pickers are disabled when one of the 3 black/white/grey slider is moved to avoid bad interaction between the picker automatically moving the sliders and manual usage of the sliders. See commit messages for details.

This needs more testing than I did. If this is accepted, the same should be done for "unbreak input profile" and "color balance".

### DIFF
--- a/src/iop/filmic.c
+++ b/src/iop/filmic.c
@@ -100,7 +100,6 @@ typedef struct dt_iop_filmic_params_t
 
 typedef struct dt_iop_filmic_gui_data_t
 {
-  int apply_picked_color;
   int which_colorpicker;
   GtkWidget *white_point_source;
   GtkWidget *grey_point_source;
@@ -534,6 +533,8 @@ static void security_threshold_callback(GtkWidget *slider, gpointer user_data)
 
 static void apply_autotune(dt_iop_module_t *self)
 {
+  if(self->dt->gui->reset) return;
+
   dt_iop_filmic_gui_data_t *g = (dt_iop_filmic_gui_data_t *)self->gui_data;
   dt_iop_filmic_params_t *p = (dt_iop_filmic_params_t *)self->params;
 
@@ -720,6 +721,23 @@ static int call_apply_picked_color(struct dt_iop_module_t *self, dt_iop_filmic_g
   return handled;
 }
 
+static gboolean draw(GtkWidget *widget, cairo_t *cr, dt_iop_module_t *self)
+{
+  if(darktable.gui->reset) return FALSE;
+  dt_iop_filmic_gui_data_t *g = (dt_iop_filmic_gui_data_t *)self->gui_data;
+
+  /* No color picked, or picked color already applied */
+  if(self->picked_color_max[0] < 0.0f) return FALSE;
+
+  call_apply_picked_color(self, g);
+  /* Make sure next call won't re-apply autotune.
+     Needed to avoid infinite loops draw -> autotune -> set_sliders
+     -> draw. */
+  self->picked_color_max[0] = -INFINITY;
+
+  return FALSE;
+}
+
 static int get_colorpick_from_button(GtkWidget *button, dt_iop_filmic_gui_data_t *g)
 {
   int which_colorpicker = DT_PICKPROFLOG_NONE;
@@ -743,55 +761,23 @@ static void color_picker_callback(GtkWidget *button, dt_iop_module_t *self)
 
   if(self->off) gtk_toggle_button_set_active(GTK_TOGGLE_BUTTON(self->off), 1);
 
-  if(self->request_color_pick == DT_REQUEST_COLORPICK_OFF)
+  int clicked_colorpick = get_colorpick_from_button(button, g);
+  if(self->request_color_pick == DT_REQUEST_COLORPICK_OFF || g->which_colorpicker != clicked_colorpick)
   {
-    g->which_colorpicker = get_colorpick_from_button(button, g);
+    g->which_colorpicker = clicked_colorpick;
 
-    if(g->which_colorpicker != DT_PICKPROFLOG_NONE)
-    {
-      dt_iop_request_focus(self);
-      self->request_color_pick = DT_REQUEST_COLORPICK_MODULE;
-      dt_lib_colorpicker_set_area(darktable.lib, 0.99);
+    dt_iop_request_focus(self);
+    self->request_color_pick = DT_REQUEST_COLORPICK_MODULE;
+    dt_lib_colorpicker_set_area(darktable.lib, 0.99);
 
-      g->apply_picked_color = 1;
-
-      dt_dev_reprocess_all(self->dev);
-    }
+    dt_dev_reprocess_all(self->dev);
   }
   else
   {
-    if(self->request_color_pick != DT_REQUEST_COLORPICK_MODULE || self->picked_color_max[0] < 0.0f)
-    {
-      dt_control_log(_("wait for the preview to be updated."));
-      return;
-    }
-    self->request_color_pick = DT_REQUEST_COLORPICK_OFF;
-
-    if(g->apply_picked_color)
-    {
-      call_apply_picked_color(self, g);
-      g->apply_picked_color = 0;
-    }
-
-    const int which_colorpicker = get_colorpick_from_button(button, g);
-    if(which_colorpicker != g->which_colorpicker && which_colorpicker != DT_PICKPROFLOG_NONE)
-    {
-      g->which_colorpicker = which_colorpicker;
-
-      self->request_color_pick = DT_REQUEST_COLORPICK_MODULE;
-      dt_lib_colorpicker_set_area(darktable.lib, 0.99);
-
-      g->apply_picked_color = 1;
-
-      dt_dev_reprocess_all(self->dev);
-    }
-    else
-    {
-      g->which_colorpicker = DT_PICKPROFLOG_NONE;
-    }
+    disable_colorpick(self);
   }
-
   set_colorpick_state(g, g->which_colorpicker);
+  dt_control_queue_redraw();
 }
 
 void gui_focus(struct dt_iop_module_t *self, gboolean in)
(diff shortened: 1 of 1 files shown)


moy:
> What code did you put there ? `process()` functions should not contain any GUI stuff nor parameters settings. process can be called from darktable-cli with no graphic environnement, and it's not good to set parameters in there. If you need to compute intermediate parameters, that should be done in `commit_params()`.

I'm not 100% satisfied with these calls in `process()` functions, but the part accessing the GUI is protected by `if (g && ...)` so it's tolerant to having no GUI.

"Just before starting the processing" is really when you want the picker-related code to be executed, otherwise you need to run the pixelpipe twice (once to get the color, and once to apply the autotune). I was hoping for a picker-specific callback right before `process()` (which would be the right solution IMHO), but didn't find one.

TurboGit:
I think this is the first time I see GUI code in process(). This must be avoided to me it is wrong to do that.

TurboGit:
In exposure() what is done is use a callback on the "draw" signal for the widget to get the picked color if I'm reading the code correctly. Can't this be done here?

TurboGit:
@moy : see #1819 I still have an icon status issue, minor but annoying, but at least nothing done in process(). I'm not sure to be able to fix that today. If you have an idea?

moy:
After experimenting a bit, it seems that:

* The draw signal is received right before the module is applied in the pipeline. This is good for us.

* It's also received in many other places. It shouldn't be much of an issue for us, since the autotuning is relatively fast. At worse, we can protect it with "if nothing changed, don't re-optimize" or so.

* It seems to be called both for the preview pipeline and for the full pipeline. Shouldn't be much of an issue either, but there might be cases where we run either for the preview or the full pipeline depending on which one runs fastest.

moy:
I just force-pushed a version that should be OK:

* No more calls in `process()` functions.

* Picker applied ASAP, i.e. when pushing the button and as the picking area changes

* Code simpler than before (57 insertions(+), 69 deletions(-), and I added several comments :-) ).

I'm still not happy with putting the code in the `draw` signal, it's called far too often and doesn't match what we want conceptually, but it's only called too often, and my code protects against useless calls (setting `picked_color_max` to `-INFINITY`), so it's OK.

Details in the commit messages.

TurboGit:
There is a GUI issue on my side. Click on one picker, click again on the same picker. It says with an active state.

src/iop/filmic.c
```diff
@@ -1244,11 +1212,18 @@ void gui_init(dt_iop_module_t *self)
   gtk_widget_set_tooltip_text(g->security_factor, _("enlarge or shrink the computed dynamic range"));
   g_signal_connect(G_OBJECT(g->security_factor), "value-changed", G_CALLBACK(security_threshold_callback), self);
 
-  g->auto_button = gtk_button_new_with_label(_("auto tune source"));
+  g->auto_button = dt_bauhaus_combobox_new(self);
+  dt_bauhaus_widget_set_label(g->auto_button, NULL, _("auto tune globally"));
```
TurboGit: Please keep old string "auto tune source", sounds better.

TurboGit:
BTW, isn't the callback just before process() the routine commit_params()? Would it work better to use this instead of draw()?

moy:
> BTW, isn't the callback just before process() the routine commit_params()? Would it work better to use this instead of draw()?

Is there a documentation about all this somewhere?

TurboGit:
> Is there a documentation about all this somewhere?

Not that I know.

moy:
Experimentally, it seems `commit_params` is called right before the module, but doesn't have the `picked_color_*`. I guess the colorpicker comes right after that.

TurboGit:
> Both work for me. Can you give more details on what you did so that I can see what's going on? When you say "active", do you mean the picker is active (drag on image draws the rectangle) or just the icon is still white?

Yes, that the picker stays white.

I can reproduce on all pickers:
1. click a picker
2. the area appear on the picture
3. click again on the same picker
4. the area disappear from the picture
5. this picker icon is still white as if selected
6. click on the image and the picker is not white anymore, meaning only the icon was not refreshed

moy:
I sill can't reproduce, but I guess I found the issue (a call to `dt_control_queue_redraw` before `set_colorpick_state`). Just force-pushed a fix. Can you test again?

moy:
Hmm, I guess I can't do more without being able to reproduce locally :-(.

In principle, the second click should execute

```
    disable_colorpick(self);
  }
  set_colorpick_state(g, g->which_colorpicker);
  dt_control_queue_redraw();
```

(at the end of `color_picker_callback`)

Can you check that the code actually goes this way (GDB or printf)?

I really don't see what could go wrong in this piece of code ...

TurboGit:
Yes, this is called, it seems that there is no redraw of the widget. As I said as soon as I click on the image I do have a draw signal (tested by adding a printf() in the draw() callback) and the icon are properly refreshed. So all this is just a refresh issue!

TurboGit:
I have a solution. No time for now. I'll propose a patch over this PR.

moy:
Cool, thanks.

TurboGit:
Merged with my fix. It was a nice case :) Yet I think we should have a better support for colorpicker and code sharing for all modules. It is quite painful to have to deal with all those details in all modules :(

moy:
> Yet I think we should have a better support for colorpicker and code sharing for all modules.

I fully agree. It's less painful now that there's less code to deal with it, but it should still be better factored.

I won't have much time for this in the next few days, but we should:

* Before the release, port this PR to profile_gamma and colorbalance

* Factor the common parts in `libs/colorpicker`

TurboGit:
@moy : I'm on that. Should have something by tonight. The factoring of code is done for filmic.

TurboGit:
@moy : done, if you have a chance to do some testing do not hesitate.

moy:
The "auto tune" button in profile_gamma.c, and "optimize luma" / "neutralize colors" in color balance are still broken the same way as the old "auto tune" in filmic. Using the color picker API there like filmic should solve this.

Changing mode in color balance should probably deactivate the picker. Currently, you can activate the fulcrum picker, change mode to lift/gamma/gain where the picker disappears, and still have the picker activated (i.e. dragging within the picture grabs a color, and it seems to do weird things then). Same for the "color control slider" dropdown.

Other pickers (exposure, levels, perhaps clut and white balance) could benefit from this, but this may wait until after the release (no need to risk breaking them now if they are not broken ...).

THANKS!

@aurelienpierre : just a ping in case you didn't see this discussion (a bit hidden in a closed PR ...).

TurboGit:
> The "auto tune" button in profile_gamma.c, and "optimize luma" / "neutralize colors" in color balance are still broken the same way as the old "auto tune" in filmic. Using the color picker API there like filmic should solve this.

I don't follow, profile_gamma and colorbalance are now using the color picker API.

TurboGit:
> Changing mode in color balance should probably deactivate the picker.

Done. Also reseing picker status when changing controls.

moy:
Hmm, I was using the darktable binary within the build tree (now that it's possible), but apparently I was getting half of the new version (eg. the Git commit id in the version string was the new one), but half of the old version (i.e. not all your changes).

TurboGit:
@moy : yes looks like running from build-tree is not working

moy:
auto-tune levels appears in "unbreak input profile" when the mode is "gamma". I think it should appear only in log mode.

TurboGit:
Right! Good catch, will fix.

moy:
clicking on the "reset" button in "color balance" doesn't disable the color picker icon if the selected icon is one of the "autotune" pickers. Weird.

TurboGit:
I saw that. Just don't know why yet.

moy:
You changed the text from "neutralize colors" to "optimize color". I think the old text was better, it's not really optimizing, but neutralizing.

Actually, in color balance, I think the optimizers are a bit more complex: they were using previously selected areas using the 3 pickers. So it was probably to keep them as normal buttons. No time to investigate more, sorry.

aurelienpierre:
> Actually, in color balance, I think the optimizers are a bit more complex: they were using previously selected areas using the 3 pickers. So it was probably to keep them as normal buttons. No time to investigate more, sorry.

Yes indeed. What have you done to my baby ? :rofl:

moy:
> clicking on the "reset" button in "color balance" doesn't disable the color picker icon if the selected icon is one of the "autotune" pickers. Weird.

Same in profile_gamma, but not in filmic. I don't understand why.

TurboGit:
I think we are facing a Gtk bug. This is just a redraw of the icon issue.

TurboGit:
I have restored the label, sorry a cut&paste error :(

For the colorpicker in color balance I think I have kept the previous behavior (using the selected region if any).

moy:
> For the colorpicker in color balance I think I have kept the previous behavior (using the selected region if any).

OK, the behavior is actually more complex than I thought. If 3 patches have been selected, it takes them, but otherwise it guesses from the whole image. Not sure it's really a good idea from the UI point of view, it's really hard to understand how the buttons work without looking at the code :-(.

This can be improved in several ways IMHO. Not all ideas are good, just thinking aloud:

* If only 1 or 2 of the 3 patches have been selected, then the callback could use the ones selected and do the whole-image guessing only for the missing one.

* If the 3 patches have been selected, then no need to activate the picker: the optimizer could run right away when clicking the button.

* Currently, the first call to the callback disables the picker callback but keeps the picker active. I think the line `self->request_color_pick = DT_REQUEST_COLORPICK_OFF;` in `apply_autoluma` should either be removed or replaced with `dt_iop_color_picker_reset`.

* Having the button be shaped like pickers when the user is more or less supposed to have picked the colors elsewhere before is confusing. I'd suggest using the picker API but not the picker button. If `apply_autoluma` disables the picker, it'll be transparent for the user that it's a picker (except for the fact that it'll draw the picker's rectangle temporarily).

moy:
Actually, on overall it seems the pickers in colorbalance are broken now. I already faced this before the last commits, but very often I use the pickers and get a full white image.

aurelienpierre:
it seems to work on my side.

> Not sure it's really a good idea from the UI point of view, it's really hard to understand how the buttons work without looking at the code :-

that's because there is still no doc on that. The way I designed it was: have a quick fallback for 80 % of the easy cases, then unclutch the auto mode when it fails. So, the button triggers a full-picture guess when user lazy, and gets whatever the user feeds it otherwise.

moy:
> So, the button triggers a full-picture guess when user lazy, and gets whatever the user feeds it otherwise.

The problem is that there's no feedback on what is done: there's no indication about which patch have already been selected, and no indication whether it's a full-picture guess or a 3-patches optimization. 

Perhaps just a `dt_control_log(_("no patch selected, doing a full-picture guess."));` (or "not enough patches selected") in the case of auto-guessing would help already. Ideally there would also be a visual indication on which patch have already been selected, but it'd be a much more involved change.

TurboGit:
Maybe we should change the label of the color picker when the 3 luma patches (or color patches) have been selected from : 
- "optimize luma" to "optimize luma from patches"
- "neutralize colors" to "neutralize colors from patches"

What do you think?

moy:
Good idea.

TurboGit:
Do you want to work on that? Or should I go with this...

moy:
If you have time, go with it. I'm spending too much time on dt and not enough on other urgent stuff these days :-(.

TurboGit:
Done. Also, I think I have fixed the redraw issue. Was due to some missing redraw in the bauhaus widgets.

PR #1820, opened 2018-11-11
Bugfix: darktable-cli --version should return 0, not 1

lastzero:


TurboGit:
Obviously correct to me. Thanks.

PR #1821, opened 2018-11-11
cmake: Only set install RPATH for darktable binaries requiring it

cryptomilk:
Hi,

it looks like this is only needed for the darktable executables and plugins.

A test build with this patch is at:
https://build.opensuse.org/package/show/home:gladiac:branches:graphics:darktable:master/darktable

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -841,6 +841,17 @@ if (WIN32)
   set_target_properties(darktable PROPERTIES LINK_FLAGS "-mconsole -mwindows -Wl,-subsystem,console")
   _detach_debuginfo (darktable bin)
   _detach_debuginfo (lib_darktable bin)
+else()
+    # Note that $ORIGIN is not a variable but has a special meaning at runtime.
+    # The string "$ORIGIN" should end up in the executable as-is.
+    set(RPATH_DT "$ORIGIN")
+    if (APPLE)
+        # The string "@loader_path" should end up in the executable as-is.
+        set(RPATH_DT "@loader_path")
+    endif()
+    set_target_properties(darktable
+                          PROPERTIES
+                          INSTALL_RPATH ${RPATH_DT}/../${CMAKE_INSTALL_LIBDIR}/darktable)
 endif(WIN32)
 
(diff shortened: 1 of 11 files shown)


TurboGit:
@LebedevRI : this looks like it works, are you ok for a second try? We'll revert if it breaks things again.

src/CMakeLists.txt
```diff
@@ -841,6 +841,17 @@ if (WIN32)
   set_target_properties(darktable PROPERTIES LINK_FLAGS "-mconsole -mwindows -Wl,-subsystem,console")
   _detach_debuginfo (darktable bin)
   _detach_debuginfo (lib_darktable bin)
+else()
+    # Note that $ORIGIN is not a variable but has a special meaning at runtime.
+    # The string "$ORIGIN" should end up in the executable as-is.
+    set(RPATH_DT "$ORIGIN")
```
LebedevRI: In general, if you have to repeat the exact same snippet in more than two places, you are doing it wrong.

LebedevRI:
> @LebedevRI : this looks like it works, are you ok for a second try? We'll revert if it breaks things again.

No opinion

TurboGit:
Note that this does not work as expected. Running dt from the build directory still uses the IOP module from the install directory. Any chance to fix that, or should I revert? Thanks.

parafin:
You can check how it's done for relocatable DT on MacOS and Windows:
https://github.com/darktable-org/darktable/blob/master/src/common/file_location.c
But I think there's a problem - the directory structures in build and install dirs aren't the same, so this use-case (running DT from build dir) might require custom code. It's probably not the best idea - why increase the amount of code if you could just install DT to some temporary directory (writable by user) and run it from there?

parafin:
Or maybe one can get away with passing --moduledir and --noiseprofiles arguments (and optionally --localedir) without changing any code.

cryptomilk:
You can put all libraries in one directory in the build dir using:

    set(CMAKE_LIBRARY_OUTPUT_DIRECTORY "${CMAKE_BINARY_DIR}/lib")

However, the code needs to try to first dlopen() the library from the build dir and fall back to the system directory. I guess it would work if we pass the path where to look for modules to:

dt_module_load_modules() and have the build dir locations of the modules in config.h. Then we can first try to load from the build dir and if that fails go to the system dir.

PR #1825, opened 2018-11-12
fix copy masks

edgardoh:
Copy / Paste history is not handling masks very well, in addition to always copy all the masks from the source image it doesn't check for duplicate forms, and when having duplicate forms on an edit they can only be removed by hand, the clear unused masks will not do it.

In order to fix it I'm doing the copy / paste in memory, so only the forms used by the modules being pasted are copied, and if a form already exists it is replaced.

This also will allow to fix the apply style on export.

As usual, I left some of the old code for a better reading on github.

### DIFF
--- a/src/common/history.c
+++ b/src/common/history.c
@@ -814,7 +814,7 @@ static int _history_merge_module_into_history(dt_develop_t *dev_dest, dt_iop_mod
   if(module_added)
   {
     dt_dev_add_history_item_ext(dev_dest, module, FALSE, TRUE);
-    dt_dev_pop_history_items_ext(dev_dest, dev_dest->history_end, TRUE);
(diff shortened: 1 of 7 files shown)


TurboGit:
I have a scenario where this does not work:
- with 3 pictures
- open the first one
- with exposure create two circle mask (push exposure to see the effect)
- go to lighttable
- copy the exposure from this image
- paste it on the second and third image
- open first image remove the two circles from the exposure module
- open the second do the same
- open the third, when you remove the first circle the effect is still on the image

edgardoh:
Confirmed! Thanks for reporting it.

But it also happens on the current version. Seems to be a display thing, if
I go to lighttable things go back to normal.

Do you think is better to fix it here or create a new PR?

TurboGit:
If it happens already let's open a new PR and I'll merge this one. Thanks.

PR #1826, opened 2018-11-13
Filmic : fixes and changes

aurelienpierre:
1. Add the OpenCL version (I forgot to commit in my first PR): ×4/×6 speed-up
2. Add a graph of the curve to diagnose issues and poor parametrization: cusps, clipping
capture d ecran du 2018-11-12 20-29-26
3. Subsequently to the graph display, be less conservative on nodes sanitization
4. Compute the contrast/slope differently to give better control from the UI parameters.
5. Desaturates the output selectively, based on the concavity of the curve. Since the concave part of the curve (toe/shadows) adds saturation and the convex part of the the curve (shoulder/highlights) desaturates, we compute the second order centered differences of the curve (= second order derivative) and use these values as weights to re/de-saturate. On the linear part, the second derivative == 0.0, so the saturation is not affected.

### DIFF
--- a/data/kernels/filmic.cl
+++ b/data/kernels/filmic.cl
@@ -1,0 +1,75 @@
+/*
+    This file is part of darktable,
+    copyright (c) 2018 Aurélien Pierre.
+
+    darktable is free software: you can redistribute it and/or modify
+    it under the terms of the GNU General Public License as published by
+    the Free Software Foundation, either version 3 of the License, or
+    (at your option) any later version.
+
+    darktable is distributed in the hope that it will be useful,
+    but WITHOUT ANY WARRANTY; without even the implied warranty of
+    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+    GNU General Public License for more details.
+
+    You should have received a copy of the GNU General Public License
+    along with darktable.  If not, see <http://www.gnu.org/licenses/>.
+*/
+
+#include "basic.cl"
+
+kernel void
+filmic (read_only image2d_t in, write_only image2d_t out, int width, int height,
+        const float dynamic_range, const float shadows_range, const float grey,
+        read_only image2d_t table, read_only image2d_t diff, const float saturation,
+        const float contrast, const float power)
+{
+  const unsigned int x = get_global_id(0);
+  const unsigned int y = get_global_id(1);
+
+  if(x >= width || y >= height) return;
+
+  float4 i = read_imagef(in, sampleri, (int2)(x, y));
+  float4 o = Lab_to_XYZ(i);
+  o = XYZ_to_prophotorgb(o);
+
+  const float4 noise = pow((float4)2.0f, (float4)-16.0f);
+  const float4 dynamic4 = dynamic_range;
+  const float4 shadows4 = shadows_range;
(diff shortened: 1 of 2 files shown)


TurboGit:
The curve is only a view and cannot be controlled directly. This makes the curve a bit less fun but why not. Otherwise when I change the grey slider the curve does not change as for the black and white slider.

TurboGit:
@aurelienpierre : ok, thanks for the feedback

TurboGit:
Not sure but I find the saturation slider less effective. Even pushed to max the saturation is slightly pushed on the image. I kind of remember that it was more effective in previous version. No?

aurelienpierre:
Just to give some more details, the purpose of this module is to build automatically a tonecurve with parameters precomputed from the output of the log and some "human-readable" inputs in order to remap the grey level to a precise value and compress the bounds of the dynamic range.

> I kind of remember that it was more effective in previous version. No?

It's actually more effective although the effect is less dramatic.  :grinning:  Previously, the saturation affected the whole luminance range, which is bad because we don't want to touch the linear part of the curve (where the saturation is ok). Now, it only affects the extreme luminance, by pushing them towards pure black and pure white, as film does. This is arguably more pleasant, but more importantly avoids any gamut clipping (saturated and very bright/dark areas are the first to clip). So you need to test on very bright skin tones, for example, it will look a bit like Kodak Portra 160 NC.

TurboGit:
Again thanks for the feedback. I'm asking just to be sure there is no issue before merging as this module is very new to me :)

aurelienpierre:
@TurboGit No problem.

TurboGit:
Ok, sounds good!

aurelienpierre:
capture d ecran du 2018-11-13 04-27-06

@TurboGit, on the left, saturation at 1000 % (max). On the right, at 150 % (almost minimum). Notice how the shadow under the chin is less redish and the skin tone more natural (less yellow) in highlights on the right. You would need to apply Adobe RGB profile on the pic to have true colors though.

moy:
I'm getting weird results with the saturation slider on highly saturated dark red. Just the saturation slider changes between the images. Surprisingly, the problematic one is the one with saturation = 100%.

maze
no-maze

I'm still getting the artefact if I turn the image into black-and-white after filmic in the pipe (so it doesn't seem to be an out-of-display-gamut issue). It doesn't disapear when I activate gamut cliping in the input color profile.

I can share the raw privately if needed.

aurelienpierre:
@moy you should try using more dynamic range by pushing the black to the left, or push the saturation te the right to desaturate less. This effect is to be expected sometimes, that's why the desaturation is not automatic.

moy:
The artefact I get are with the default parameters. Just one click to activate the filmic module and the artefacts are there :-(. I can reproduce on an RAW image with only filmic activated (no other module).

It seems to happen quite materialistically on saturated reds (even not-so-saturated, I get it on a child's lips).

I don't understand why this is expected: desaturation shouldn't affect the luma, right?

aurelienpierre:
could you send me the RAW to contact à aurelienpierre point com ?

aurelienpierre:
@moy ok, I have to find a more robust estimator for the second order derivative upper bound. It's a normalization issue.

PR #1674, opened 2018-04-10
watermark: update documentation for tags and ratings

rabauke:
Documentation for the merged PR  #1671.

PR #1827, opened 2018-11-13
guilded filter: rework dt_develop_blend_process and dt_develop_blend_…

rabauke:
I reworked the functions dt_develop_blend_process and dt_develop_blend_process_cl for better readability. There is a clear one-to-one correspondence in both functions now. Furthermore, the global opacity parameter is taken into account in the application of the recently introduced mask tone-curve. This might break edits carried our with recent git master but should in crease usability. E.g., the (drawn & parametric) mask can be easily inverted by inverting the drawn mask and inverting the parametric mask plus inverting the mask brightens slider of the mask tone curve.


TurboGit:
This does not compile:

```
...darktable/src/develop/blend.c: In function ‘dt_develop_blend_process’:
...darktable/src/develop/blend.c:3005:11: error: ‘guide’ not specified in enclosing ‘parallel’
           memcpy(guide + oindex, (float *)ivoid + iindex, sizeof(*guide) * owidth * ch);
           ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
...darktable/src/develop/blend.c:2999:9: error: enclosing ‘parallel’
 #pragma omp parallel for default(none)

```

TurboGit:
Breaking edits with current Git is not a problem. Until 2.6 is not out we can break anything :)

rabauke:
> This does not compile:
> 
> ```
> ...darktable/src/develop/blend.c: In function ‘dt_develop_blend_process’:
> ...darktable/src/develop/blend.c:3005:11: error: ‘guide’ not specified in enclosing ‘parallel’
>            memcpy(guide + oindex, (float *)ivoid + iindex, sizeof(*guide) * owidth * ch);
>            ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
> ...darktable/src/develop/blend.c:2999:9: error: enclosing ‘parallel’
>  #pragma omp parallel for default(none)
> ```

 @TurboGit  Thanks for the hint. Has been fixed now. Because of the `default(none)` declaration I am surprised that it did compile before with any compiler before.

rabauke:
@TurboGit Side note: I plan to implement also two other modifications: 

* Instead of applying either the guided filter _or_ the Gaussian blur filter to the mask it should be possible to apply both.
* The user should be able to choose the input image _or_ the module's output as a guide.

TurboGit:
Note that the feature freeze is tomorrow. Do you plan to have those new features ready tomorrow or will it come at an addition on next development cycle?

rabauke:
@TurboGit  I have also included the mentioned modifications now into this PR.

TurboGit:
Nice! I'll have a look tomorrow.

TurboGit:
Sounds good thanks.

rabauke:
@TurboGit This order has been chosen intentionally. Feathering comes first in the mask refinement, Gaussian blur afterwards. This is to match typical workflows which involve both kinds of mask adjustments. Feathering may yield masks with too hard edges, which may be smoothed by the Gaussian filter.

TurboGit:
@rabauke : understood! all this will become clear with the documentation.

upegelow:
I like the options for mask refinement a lot. However, I am a bit afraid of the large number of controls that are visible all the time. Looks a bit like an airplane cockpit now. Wouldn't it be better to only activate all these controls on request? Maybe a combobox "mask refinement" with yes/no could help to improve clarity.

upegelow:
Another important remark: looks like the recent commits have broken legacy presets. I have my own defined presets with parametric masks defined in several modules. When I activate the preset now I only get the module with blending switched off :(

TurboGit:
This last one is quite bad!

TurboGit:
@upegelow : I agree that there is quite a lot of controls, the idea to have a combo with two options (just blur as before and guider filter).

TurboGit:
@upegelow : I have fixed the blending legacy support, a cut&paste issue!

upegelow:
> @upegelow : I have fixed the blending legacy support, a cut&paste issue!

Hmm, not sure that this has worked out. I still cannot see the parametric masks. Legacy parametric masks seem to be gone also when importing legacy history stacks. I don't want to test it too much with my work as I fear to lose all edits.

TurboGit:
This is strange. I have just rebuilt an old darktable and then with current master I have:

> [imageop_init_presets] updating 'exposure' preset 'testblend' from blendop version 7 to version 8

Also the old edits are now correctly upgraded. The preset testblend is a parametric on hue and with a blur mask, everything is properly upgraded on my side.

aurelienpierre:
> Hmm, not sure that this has worked out. I still cannot see the parametric masks. Legacy parametric masks seem to be gone also when importing legacy history stacks. I don't want to test it too much with my work as I fear to lose all edits.

same here. The drawn masks are in the list, but the modules don't use them and are reset

upegelow:
Blending options for presets are stored in data.db. Looks like darktable automatically tries to update all of them to the latest version. If an error occurs here the blending part of a preset is lost. I have just restored my data.db from a recent backup, and now with current master presets are back to normal. Once again the blendop version of all presets gets automatically updates on first invocation of darktable - this time with correct results. Bad news for those who don't have a backup of data.db.

TurboGit:
@upegelow : do you restored data.db and this time all was properly migrated to latest version? does that means that there was a transient issue?

upegelow:
I assume that there is a problem with blendop version 8. Whatever happened: all settings which have been converted recently to this version are screwed. This seems to affect *all* entries in presets as version updates happens for all presets each time darktable is started. But this is also true for all image edits which have been touched recently. For touching it seems to be sufficient to open an image in darkroom or export an image. All blending is set to off in this case and the module takes effect on the full image. As I use a lot of parametric masks I have lost quite some edits. Not good.

All images that have not been touched and still contain blendop version 7 or earlier seem to be OK if opened with current master.

TurboGit:
Ok, thanks for the feedback, we have a big issue to fix then.

@rabauke : hope you'll find time to work on this.

TurboGit:
> Whatever happened: all settings which have been converted recently to this version are screwed.

I confirm. But I took a data.db backup and started again darktable. All my presets have been migrated properly with the blendop parameters. So to me it looks like a transient issue that has been fixed in current master. That's sad, but always possible for us working on master :( As dev we really need to have a backup, you had one and I had one too.

Yet, we need to double check that all is under control now.

rabauke:
I created fresh installations of darktale 2.4.4 and current git master (commit 0515f4cc8eb4f5184135665d09702e13027101ba). I did some very simple edits involving drawn masks in dt 2.4.4. No problems opening these in current git master. Hope problem has been fixed now. Please report if problem pops up again and describe which versions have been used.

TurboGit:
Ok, here is what is happening:
1.  in c94173f9 we have the first new blend version 8. the legacy is working
2. in 0b073023 the params change but we keep version 8

So every edit done with 1 are lost and preset migration are also broken.

To me all is fine on master. And the issue is a transient one that can occurs on master.

@rabauke : can you confirm that?

edgardoh:
It seems to me that mask_combine is missing from the migration

TurboGit:
Hum... Indeed and this is missing from many migrations :( From v5 and v6 and possible others. Will fix that right away.

upegelow:
> That's sad, but always possible for us working on master :( As dev we really need to have a backup, you had one and I had one too.

True, but I think we should take this as a reminder not to be too permissive when breaking legacy history stacks in central parts of the program. The parametric mask is a feature that is used very intensively by many users and developers. It would have been better in this case to not overwrite the blendop version 8 with a newer one but generate a new one. Even if the algorithm has changed and not all of the version 8 features can be fully converted it would have been the better choice.

I am more relaxed when it comes to the development of a completely new iop, like retouch or filmic. If I use it I know that incompatible changes can happen.

For me it means that I have now a lot of edits where blendop parameters have been converted into the dead-end version 8. Any idea how I could restore them? Maybe a tool to convert back to version 7?

TurboGit:
@upegelow : the best is to use your backup data.db (as you've explained above) to restore the preset. or maybe I did not fully understood the question?

upegelow:
Presets are restored, but I have lost like 70 edits.

cryptomilk:
I just wanted to export an image and and I just got a gray image. When checking the history, I see that all blend modes, drawn masks and parametric masks are gone!

However if I open the XMP file with an editor I can still see mask information in there, but they don't work. The blendop verison is set to 8.

DottieUnderwood:
I created a hacked version of darktable that will convert v8 old to v8 new. Only needs minor edits to blend.c and blend.h. First you temporary "convert" the xmp files from v8 to v1. 
for i in *.xmp; do sed s/blendop_version=\"8\"/blendop_version=\"1\"/g $i>$i; done
Then run the modified darktable to convert "v1" to v8 new.

The new blend features are great btw! Good work @rabauke. Parametric masks are the really **the** killer feature of darktable and they are not even mentioned on the feature page WTF?

Add this to blend.h
/* he he */
typedef struct dt_develop_blend_params8a_t
{
  uint32_t mask_mode;
  uint32_t blend_mode;
  float opacity;
  uint32_t mask_combine;
  uint32_t mask_id;
  uint32_t blendif;
  float radius;
  float contrast;
  float brightness;
  uint32_t mask_blur_mode;
  uint32_t reserved[4];
float blendif_parameters[4 * DEVELOP_BLENDIF_SIZE];
} dt_develop_blend_params8a_t;

blend.c (This replaces the old v1 converter)
  if(old_version == 1 && new_version == 8)
  {
    /* he he */
    if(length != sizeof(dt_develop_blend_params8a_t)) return 1;

    dt_develop_blend_params8a_t *o = (dt_develop_blend_params8a_t *)old_params;
    dt_develop_blend_params_t *n = (dt_develop_blend_params_t *)new_params;
    dt_develop_blend_params_t *d = (dt_develop_blend_params_t *)module->default_blendop_params;

    *n = *d; // start with a fresh copy of default parameters
    n->mask_mode = o->mask_mode;
    n->blend_mode = o->blend_mode;
    n->opacity = o->opacity;
    n->mask_combine = o->mask_combine;
    n->mask_id = o->mask_id;
    n->blur_radius = o->radius;
    n->blendif = o->blendif;
    n->brightness = o->brightness;
    n->contrast = o->contrast;
    memcpy(n->blendif_parameters, o->blendif_parameters, 4 * DEVELOP_BLENDIF_SIZE * sizeof(float));
    return 0;
  }

rabauke:
@DottieUnderwood  Thanks for your contribution! Minor refinement of our great suggestion (untested):

```
if(old_version == 1 && new_version == 8)
  {
    /* he he */
    if(length != sizeof(dt_develop_blend_params8a_t)) return 1;

    dt_develop_blend_params8a_t *o = (dt_develop_blend_params8a_t *)old_params;
    dt_develop_blend_params_t *n = (dt_develop_blend_params_t *)new_params;
    dt_develop_blend_params_t *d = (dt_develop_blend_params_t *)module->default_blendop_params;

    *n = *d; // start with a fresh copy of default parameters
    n->mask_mode = o->mask_mode;
    n->blend_mode = o->blend_mode;
    n->opacity = o->opacity;
    n->mask_combine = o->mask_combine;
    n->mask_id = o->mask_id;
    if (o->mask_blur_mode==1) {
       n->blur_radius = o->radius;
       n->feathering_radius = 0;
    } else {
       n->blur_radius = 0;
       n->feathering_radius = o->radius;
    }
    n->blendif = o->blendif;
    n->brightness = o->brightness;
    n->contrast = o->contrast;
    memcpy(n->blendif_parameters, o->blendif_parameters, 4 * DEVELOP_BLENDIF_SIZE * sizeof(float));
    return 0;
  }
```

cryptomilk:
I will try that soon.

As a reminder, documentation for the new features are needed! For the filmic module we use hackmd for collaborative editing.

https://hackmd.io/s/HyqqZQdaX#

TurboGit:
Nice going! Let's see if @upegelow can get back the 70 edits...

rabauke:
I am going to draft some documentation, possibly I will try mackmd. (Never heard of that before.) Please also consider https://github.com/darktable-org/darktable/pull/1674

cryptomilk:
The recovery from above doesn't work for me at least.

TurboGit:
@rabauke : the documentation has always been handled by @upegelow.

upegelow:
> @rabauke : the documentation has always been handled by @upegelow.

It would be great if someone else could take the lead on the manual this time. I am fully drowned in a new job.

upegelow:
> Nice going! Let's see if @upegelow can get back the 70 edits...

I will give it a try tomorrow evening. Right now I have been able to find a backup of my XMPs originating from the day just before the problematic commit. From there I can reconstruct the changes I made in between.

PR #1828, opened 2018-11-13
colorbalance: better management of user-selected vs autoselected patches

moy:
This PR keeps track of whether each luma/color patch was autoselected or user-selected for two tweaks:

* Don't claim we're doing an optimization "from patch" when patches were auto-selected.
* Don't override user-selected with auto-selection.

Details in commit messages. Yes, I like clean history and long commit messages ;-).

Cc: @aurelienpierre.


TurboGit:
Sounds really good! Thanks.

PR #1829, opened 2018-11-14
Filmic : bug fixes

aurelienpierre:
1. Try to fix a building bug on Windows reported by @phweyland 
2. Change a label and move the white slider before the black (to follow the setting order)
3. Fix the bug reported by @moy regarding noise in low-lights with the desaturation. The selective desaturation no longer use the 2nd order discrete derivative, proven to be too unreliable with curves LUT. Instead, we use now a simple window function built on the logarithm parameters. Now, the desaturation mask looks like that (x being the luminance, y being the weight of the saturation):
image
4. Vectorize more operations in `process()` SSE version

### DIFF
--- a/data/kernels/filmic.cl
+++ b/data/kernels/filmic.cl
@@ -22,7 +22,7 @@ kernel void
 filmic (read_only image2d_t in, write_only image2d_t out, int width, int height,
         const float dynamic_range, const float shadows_range, const float grey,
         read_only image2d_t table, read_only image2d_t diff, const float saturation,
-        const float contrast, const float power, const float max_grad)
+        const float contrast, const float power)
 {
   const unsigned int x = get_global_id(0);
   const unsigned int y = get_global_id(1);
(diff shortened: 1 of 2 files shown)


moy:
I don't get the ugly artefacts anymore, but the "saturation" scale is surprising: the "saturation" slider does almost nothing between 10% and 100% (on the image I sent you privately, the effect starts being really visible at 1% only), and really gets a non-trivial effect only on the first percents. Not necessarily a problem, but as a user it's surprising. Perhaps you get something more useful in the range 10%-100% with a wide gamut display though.

aurelienpierre:
I understand. Film has non-trivial effects for users used to digital. It depends a lot on the image, really. Some have no visible effects until you get below 5-10 %, for some others you have to get around 20-50 %. The most important part is it should follow the log dynamic range and desaturates progressively at the bounds.

TurboGit:
I confirm, probably an issue with the saturation slider. On my side, tested on 2 pictures. The saturation does nothing (tested with a snapshot, no visible diff between 10% and 1000%).

aurelienpierre:
@TurboGit and @moy you have to use color pickers in lowlights and check the a and b values.

TurboGit:
I see, indeed very light change and expected as discussed in a previous PR. The two pictures where I tested before had very little lowligh.

moy:
The maths are probably right, but perhaps the UI can be improved by having a non-linear slider (e.g. apply x^2 or so to the saturation parameter between the UI and the maths) so that the whole range is useful.

I haven't tested on enough pictures (and actually didn't do real-life tests where the goal was actually to get a nice image), so I can't say whether this is really a good idea yet.

PR #1831, opened 2018-11-14
group mask adjustment UI elements by section labels

rabauke:
Darktable comes with a lot of options to modify blend masks. For better usability I added some section labels to group options that belong together. The figure shows an example of the revised UI.

### DIFF
--- a/src/develop/blend_gui.c
+++ b/src/develop/blend_gui.c
@@ -1375,6 +1375,7 @@ void dt_iop_gui_init_blendif(GtkBox *blendw, dt_iop_module_t *module)
     g_signal_connect(G_OBJECT(bd->upper_polarity), "toggled", G_CALLBACK(_blendop_blendif_polarity_callback),
                      bd);
 
+    gtk_box_pack_start(GTK_BOX(bd->blendif_box), dt_ui_section_label_new(_("parametric mask")), TRUE, TRUE, 0);
     gtk_box_pack_start(GTK_BOX(bd->blendif_box), GTK_WIDGET(header), TRUE, FALSE, 0);
     gtk_box_pack_start(GTK_BOX(bd->blendif_box), GTK_WIDGET(uplabel), TRUE, FALSE, 0);
     gtk_box_pack_start(GTK_BOX(bd->blendif_box), GTK_WIDGET(upslider), TRUE, FALSE, 0);
@@ -1534,6 +1535,7 @@ void dt_iop_gui_init_masks(GtkBox *blendw, dt_iop_module_t *module)
     gtk_box_pack_end(GTK_BOX(abox), bd->masks_brush, FALSE, FALSE, 0);
 
 
+    gtk_box_pack_start(GTK_BOX(bd->masks_box), dt_ui_section_label_new(_("drawn mask")), TRUE, TRUE, 0);
     gtk_box_pack_start(GTK_BOX(bd->masks_box), GTK_WIDGET(hbox), TRUE, TRUE, 0);
     gtk_box_pack_start(GTK_BOX(bd->masks_box), GTK_WIDGET(abox), TRUE, TRUE, 0);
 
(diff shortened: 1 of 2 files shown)


TurboGit:
This is better. But still using lot of space.

We have:
```
- feature guide
- feathering radius
- mask blur
- mask brightness
- mask contrast
```

We could have:
- mask tool (combo with blur, full)

When blur selected:
```
- mask tool   blur
- mask blur
```

When full selected:
```
- mask tool   full
- feature guide
- feathering radius
- mask blur
- mask brightness
- mask contrast
```

At the end, in full mode we have one more line :( Maybe we can find something better.

aurelienpierre:
I don't know which commit is messing up here, but I have lost all the masks and blending options done with previous versions of dt.

TurboGit:
Which version? I have fixed that in commit 15ebdaaa.

aurelienpierre:
> Which version? I have fixed that in commit [15ebdaa](https://github.com/darktable-org/darktable/commit/15ebdaaa4f675e0cbef8c13701c02381105ebbc0).

Ah. Seems to work now but I have lost the editing history on one picture, and I'm unable to reimport it from the XMP embedded in the JPEG file.

rabauke:
> We could have:
> 
>     * mask tool (combo with blur, full)

I agree that the UI has become quite a lot of options and (new) users might be overwhelmed by the possibilities. I am not sure if an additional combo box would be the solution. Possibly a step forward. Where would one maintain the value of the toggle box. Would this be an additional blend parameter? Is just a parameter that modifies the appearance of the UI, thus one might store this information somewhere else.

upegelow:
I am clearly in favor of Pascal's suggestion. The section labels are already an improvement but I still see the risk of overwhelming novice users. While they will understand what "blur" means the complete set of options is so much that many will not even dare to test. The most clean place for the toggle is of course in the blendop parameters (migrating to version 9). Take one of the reserved fields to not let the params block grow bigger.

aurelienpierre:
Just my 2 cents here, but we have to draw a clear distinction between an efficient UI (= allowing quick setup) and a clear UI (= allowing easy understanding of what's going on).

Having hidden sliders in drop-down, toogle boxes or sub-menus might seem nice to declutter the interface, however, it adds more steps to access the settings. It's the metaphor of the aircraft cockpit… Yes, you have a lot of knobs but they are immediately accessible. It took me less time to set up an autopilot in an A-340 than what it takes me to set up a consumer GPS.

At the end, I believe in documentation, not in simplification.

upegelow:
> It's the metaphor of the aircraft cockpit… Yes, you have a lot of knobs but they are immediately accessible. It took me less time to set up an autopilot in an A-340 than what it takes me to set up a consumer GPS.
> 

It all ends up to the question if our target group resembles more a trained airplane pilot or the buyer of a piece of consumer electronics. There are probably good reasons for the way how cockpits are designed and how consumer GPS devices are designed.

Reminds me of a discussion we had several years ago: should we offer a general selection between an "novice view" and an "expert view"?

aurelienpierre:
> There are probably good reasons for the way how cockpits are designed and how consumer GPS devices are designed.

Yes, they are designed for lazy fools who don't want to read the manual. So it has to be "intuitive", but intuitive for whom ? I dislike intuitive, it's how you end up with the ribbon menu in Microsoft Office. By trying to make complicated things more simple, they ended up making simple things more complicated.

darktable is complicated, there are lots of options, advanced features, so just make it rational, don't even try to make it simple, and write a full documentation instead. It's Linux, not Windows.

TurboGit:
A difficult discussion. I agree with Aurélien. It would be terrible to simplify the UI to please beginners and to make expert loose time having to open multiple drop-down to find the good option. The balance is delicate to find.

About the novice vs expert view... Why not having a preference for this. A simple boolean that will remove many options like the fusion algorithm, the masks blur and feathering, the log view of the tone curve... etc. This sounds like a better choice than having multiple drop down in my point of view. So maybe, just do nothing for this PR and let's plan something more challenging for 2.8.

PR #1832, opened 2018-11-15
filmic: rename "auto tuning security factor" to "security factor"

moy:
It's misleading to include "auto tuning" in the name, since the slider
acts also when auto-tuning isn't enabled. Mention auto-tuning in the
tooltip text.

I'm not a native but I think we should also `s/security/safety/` here and in profile_gamma.

Cc: @aurelienpierre

### DIFF
--- a/src/iop/filmic.c
+++ b/src/iop/filmic.c
@@ -1279,9 +1279,11 @@ void gui_init(dt_iop_module_t *self)
   // Security factor
   g->security_factor = dt_bauhaus_slider_new_with_range(self, -200., 200., 1.0, p->security_factor, 2);
   dt_bauhaus_widget_set_label(g->security_factor, NULL, _("auto tuning safety factor"));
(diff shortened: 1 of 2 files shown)


TurboGit:
Indeed "safety factor" is maybe more correct. But as you I'm not native English :)

TurboGit:
@moy : ok so let's go with "safety factor" then.

moy:
Force-pushed a version that does security -> safety too.

You may want to omit the last commit (renaming of variables in the code) if you're afraid of conflicts, but I think it's better to be consistent between variable names and UI in the long run.

TurboGit:
I have conflicts!

moy:
I'll rebase.

moy:
Rebased on `origin/master`.

TurboGit:
Perfect thanks!

PR #1833, opened 2018-11-15
tonecurve: do not transform picker sample in RGB or XYZ mode.

TurboGit:
Fix picker sample displayed on the graph.

Since the graph and the curve on the module are not changed when using XYZ or RGB mode (that is they stays in Lab), I do not think we need to change the picker sample.

Really not sure, but it does fix the picker area representation.


TurboGit:
Close, see #1836

PR #1834, opened 2018-11-15
Fix opencl of nlmeans (borders were oversmoothed)

rawfiner:
Before:
    output was very different than with CPU version of nlmeans on borders:
    they were oversmoothed (very visible with high radius).
Now:
    output much closer, and borders are not oversmoothed anymore.

Important warning: I do not have a GPU. I tested this opencl code on CPU. Please, test it on a GPU

Exemple:
With this test image:
black

Using denoise profile in non local means mode with patch size = 0, radius = 30, and force = 4
CPU output:
black_profile_cpu

old opencl output:
black_profile_gpu_old

new opencl output:
black_profile_gpu_new

There is still a very small difference between opencl output and CPU output, but it is much less visible and is probably due to rounding error (thanks @upegelow for pointing me the fact that rounding errors could result in small differences)

This resolves this issue: https://redmine.darktable.org/issues/12400

### DIFF
--- a/data/kernels/denoiseprofile.cl
+++ b/data/kernels/denoiseprofile.cl
@@ -86,7 +86,10 @@ denoiseprofile_dist(read_only image2d_t in, global float* U4, const int width, c
   const int y = get_global_id(1);
   const int gidx = mad24(y, width, x);
 
+  // Reminder: q.x and q.y can be negative
   if(x >= width || y >= height) return;
+  if(x+q.x >= width || y+q.y >= height) return;
+  if(x+q.x < 0 || y+q.y < 0) return;
 
   float4 p1 = read_imagef(in, sampleri, (int2)(x, y));
   float4 p2 = read_imagef(in, sampleri, (int2)(x, y) + q);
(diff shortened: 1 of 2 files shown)


data/kernels/denoiseprofile.cl
```diff
@@ -211,7 +214,22 @@ denoiseprofile_accu(read_only image2d_t in, global float4* U2, global float* U4,
   const int y = get_global_id(1);
   const int gidx = mad24(y, width, x);
 
-  if(x >= width || y >= height) return;
+  if(q.x<0)
+  {
+    if(x-q.x >= width || x<-q.x) return;
```
upegelow: This looks a bit strange to me. Can you please elaborate on the logic?

rawfiner: Yes of course. Thanks for taking time to look at the code.
The idea of all these tests is to prevent accesses to pixels that are outside the limits of the image.
See bellow, there is a memory access at (x,y)+q and another at (x,y)-q
One very important thing is that q.x and q.y can be negative.
Thus, if we have x=3 and q=-5 for instance, x+q will be negative, and thus outside the range of the line we are considering.
The test (x-q.x<width) checks one of the memory bounds for the access (x,y)-q
The test (x<-q.x) is equivalent to (x+q.x<0). It checks one of the bounds for the memory access (x,y)+q

TurboGit: All this sounds ok to me. If @upegelow is ok I'll merge.

upegelow: I tested it and it works fine for me.

TurboGit: Thanks, will merge then.

rawfiner:
Thanks to both of you for the code review and testing

rawfiner:
@TurboGit could you close the bug report on redmine ( https://redmine.darktable.org/issues/12400 )? It seems that I cannot do it, even if I was the person who opened the bug report ;-(

rawfiner:
Unfortunately, it seems that this PR did not solve completely the issue, but made it only less visible... ;-(
The issue is still here, but borders are not oversmoothed anymore, they are undersmoothed. See https://redmine.darktable.org/issues/12400 for some images.

PR #1836, opened 2018-11-16
Tonecurve : fix the color picker errors in XYZ/RGB 

aurelienpierre:
- reset the notebook view to the first tab when the color space is changed
- reset the interpolator GTK combobox when the curve is reset from double click on the graph
- hide the histogram in RGB and XYZ mode, since it is Lab histogram 
- fix the color picker errors in XYZ/RGB (similar to #1833 )


TurboGit:
Working great now! Thanks.

PR #1837, opened 2018-11-16
modulelist: add default setup.

TurboGit:
Easy way to get back to initial configuration when dt was first started.


TurboGit:
manually merged.

PR #1838, opened 2018-11-18
Eliminate assertion failures introduced in PR #1816.

schwerdf:
Running Darktable with my previous pull request #1816 merged in, I noticed that there were a lot of assertion failures coming out of the autocomplete event handlers, because they were being run even when editing on the quick-tag entry box was disabled.

This change adds guard conditions to the handlers to avoid running them in this case, eliminating the assertion failures (and speeding up tagging somewhat).


TurboGit:
Thanks!

PR #1839, opened 2018-11-19
filmic : use RGB channels average as a black exposure guess instead of RGB min

aurelienpierre:
This only affects the guess done in auto mode. The current version almost always puts the black at -16 EV.


TurboGit:
Thanks!

PR #1840, opened 2018-11-19
X-T3 for 2.4.x

levitte:
This is just a backport from master, expressly to get X-T3 into the 2.4.x series.


TurboGit:
You must not change rawspeed.

levitte:
Does this mean that bringing X-T3 to the 2.4.x series is impossible? BTW, all I did was update the submodule revision, does that count as "change rawspeed"?

levitte:
> BTW, all I did was update the submodule revision, does that count as "change rawspeed"?

Yeahok, I guess it does.

rawfiner:
@TurboGit why is it not legit to update rawspeed here? It is the only way to be able to open new raw files in darktable... Or is it supposed to be updated by someone in particular, and other devs should not update it?
Or I am missing something?

TurboGit:
Because rawspeed has probably been updated since then and we will coordinate with Roman to cut a stable branch in rawspeed for the release. So, until then we do not want to update rawspeed submodule.

rawfiner:
Alright, that makes sense. Thanks :-)

LebedevRI:
> Or is it supposed to be updated by someone in particular, and other devs should not update it?

Basically, yeah.

> cut a stable branch in rawspeed for the release. 

So far there were no "stable rawspeed branches".
I *think* in upcoming cycle it will be possible to avoid to branching it too.

@darix did bring up the idea of "`lensfun-update-db` for darktable",
but most of the changes would need to be on dt side.

> So, until then we do not want to update rawspeed submodule.

It probably can be updated, but i bet there are more changes that need
to be cherry-picked on the dt side (or else you will break the 2.4.x).
Mostly because 2.6 is near i'm not sure it's worth it.
'Everyone' should be testing master :)

darix:
if you want test master you can grab a package here:

https://software.opensuse.org/download.html?project=graphics:darktable:master&package=darktable

(we are a few commits behind master due to technical issues, but I hope those can be sorted tomorrow)

PR #1841, opened 2018-11-21
custom module order and masks history

edgardoh:
This actually is two branches merged in one, the masks history adds the masks to the history, the custom module order allows the user to change the order of the modules on the pipe.

**1-masks history**
I added a snapshot of the masks on the history entry and refresh it each time there's an operation with masks.
If the operation is from a module it is added to the module's entry on history.
If it is from the mask manager a new entry 'mask manager' is created.
I have replaced the main.mask table with main.masks_history, I could just add a column, but the name seems better.

**2-custom module order**
The modules can now be re-ordered with the multi-instance menu option or with SHIFT + drag.

I have added a new field (double)iop_order to the module and history, it now controls the order of the pipe, multi_priority is used only to id a multi-instance and is on ascending order (currently is in reverse order), priority is no longer used and it has been removed.

To keep backward compatibility a new field iop_order_version is added to the image. This is needed because there are modules that are enabled but do not show in history, so when history is read there’s no way to know the iop_order for them.
Iop order version is generated dynamically, the first version is a list of the current modules starting at 1 and incrementing by 1, this should never be changed. Next versions are generated by calling functions to add/move modules relative to existent modules (i.e. _ioppr_insert_iop_after(_iop_order_list, history_list, "basicadj", "colorin", dont_move))
Before history is read, a iop order list is generated for that version and the iop_order is assigned to each module. When reading the history the iop_order is assigned from the modules in history, at this point we have the pipe exactly the same as when it was saved. Then modules added after this iop_order_version are assigned with an iop_order checking that there’s no conflicts with the current order, so new modules will still be displayed in the proper order in the pipe.

I have considered a couple of other designs, but I liked this better.

There are two 'rules' that controls how/where the user can move a module:
-a new module flag, IOP_FLAGS_FENCE, that prevents any module to be moved across it.
Demosaic have this flag, so raw and non-raw modules are not mixed:
https://github.com/darktable-org/darktable/pull/1841/files#diff-a001a1125b3a87d818accabc8b2f2751R172

-a list of module pairs, (module1, module2). It says that module1 must always be before module2.
colorin and colorout have this rule:
https://github.com/darktable-org/darktable/pull/1841/files#diff-58dc49e912c0e0c7ecb71cf42c999effR105

When user moves a module or a new module is added/moved on the iop_order_version, the logic to calculate the new iop_order is as follows:
if module1 is moved between module2 and module3, then
module1 = module2 + (module3 - module2) / 2.
iop_order is a double, so there should be enough room to move modules around. 

Since priority is no longer used and modules can be moved everywhere the module’s colorspace do not depend anymore of the position on the pipe, so modules now ask for the input colorspace and informs the output one. This can be dynamic, so Lab modules with an RGB option can ask for an RGB input image and work in the working colorspace instead of having prophoto hard-coded.

**3-Working colorspace**
A new option is added to the color input profile that allows to select the working colorspace. Only RGB colorspaces can be selected, if for some reason the colorspace can’t be generated linear rec 2020 is used.
Image is transformed to the working colorspace by the pipe, so it will be the default for RGB modules between colorin and colorout.
Colorin has been moved right after demosaic and colorout before finalscale, so the entire pipe work on the working colorspace, but those modules can be moved, so the current order can still be used.
Lab modules can also be moved before colorin or after colorout, but the transform will assume that the RGB space is the working colorspace, so the transformation won’t be correct. That won’t be validated and is user responsibility.

**4-Basic adjustments**
A new module ‘basic adjustments’ is added, it works on RGB and adjust exposure, contrast, brightness and saturation and has an auto feature based on RawTherapee’s auto levels. It is mainly to test how a new module behaves with the iop_order_version feature but results are nice and is a good alternative to the base curve.

**5-rgb curve**
Same as the basic adjustments, I've done it to test some features. It has only rgb curves.

There's 6 new dummy modules, dummy1 .. dummy6, that can be used to test how new modules affect the pipe order. To use it they must be enabled on CMakeLists.txt
https://github.com/edgardoh/darktable/blob/mask_iop/src/iop/CMakeLists.txt#L85
and add it somewhere on the priority list
https://github.com/darktable-org/darktable/pull/1841/files#diff-58dc49e912c0e0c7ecb71cf42c999effR69

I have migrated a database with over 10.000 images, exported and edited some with very heavy edits, everything seems fine.
New edits also work as expected, but of course there are endless possible scenarios, so I'll keep using it.


TurboGit:
@edgardoh, sounds like another very nice move forward ! that's of course for 2019, looking forward to a stable PR about this. Something not clear (to me) on the description, is it possible and easy to define a new pipe module order or does that means recompilation? Also does all this means that for current edit the old order is kept to ensure history backward compatibility?

edgardoh:
At the bottom I describe, seems that not very well, how to add new modules
to the pipe. It is easy, but it needs recompilation. Essentially you add a
new version to the priorities, pretty much like for a new module version.
This ensure backwards compatibility, and less important but nice to have,
new modules are on the right order on the pipe, even for old versions.

edgardoh:
I have updated the first post with a couple a features I forget about.

aurelienpierre:
That's very interesting, I was working on the color spaces API to allow the modules to work in ACES RGB (https://github.com/aurelienpierre/darktable/tree/next-gen) so I will merge my branch with yours and continue the color sanitization.

Also, a few notes: 

- gamma-corrected RGB are bad working spaces to push pixels,
- what is the RGB space you name "linear RGB" ?
- RGB spaces should be merged between `colorspaces.c` and `colorspaces_inline_conversions.h` but I propose to take over from here (I have merged white points already).
- I plan on adding ACES P0, P1, ACEScc, ACESgc, and CIE RGB.

also, if you look at my branch, I have made stupid things with Lab that need to be reverted before going further.

edgardoh:
As I stated somewhere else, color is not my thing, so please be gentle.

Its not my goal on this PR to redesign how color is handled by dt, I'm not
saying is not important, just not my goal here. It will take someone with
strong color knowledge, if you are that person then great. And if you are
planning on base your branch on mine, I expect some redesign on my side
will be needed.

What I've done is as follows:
I have split the iop_cs_rgb into iop_cs_linear_rgb and iop_cs_gamma_rgb
https://github.com/edgardoh/darktable/blob/mask_iop/src/develop/imageop.h#L550

So demosaic is returning iop_cs_linear_rgb
colorin expects iop_cs_linear_rgb and returns iop_cs_Lab
colorout expects iop_cs_Lab and returns iop_cs_gamma_rgb
gamma expects iop_cs_gamma_rgb

Each module still have a default colorspace that depends on his relative
default position on the pipe regarding demosaic, colorin and colorout, but
now they can ask for a particular colorspace and inform the output
colorspace, that is done like this:
https://github.com/edgardoh/darktable/blob/mask_iop/src/iop/levels.c#L567
https://github.com/edgardoh/darktable/blob/mask_iop/src/iop/tonecurve.c#L806

The pipe handles the color transformation, so before call the process() it
transform to the expected colorspace by the module, like this:
https://github.com/edgardoh/darktable/blob/mask_iop/src/develop/pixelpipe_hb.c#L1357

Transformation is done here (yes, yes, not the right place for this
function):
https://github.com/edgardoh/darktable/blob/mask_iop/src/common/iop_priorities.c#L1101

color picker and histogram are working fine like this.
A note on histogram, it seem to work with unsigned ints, I didn't look at
it very deep as I'm not planning on doing anything about it, but if that's
so, there should be some rounding errors on the modules that uses histogram
for calculations.

The blend module is another thing, now it expects both input and out images
to have the same colorspace, and it must be the default colorspace for the
parent module, so before calling the blend module I have to transform both
images:
https://github.com/edgardoh/darktable/blob/mask_iop/src/develop/pixelpipe_hb.c#L1444

The blend module should be redesigned:
-to accept any colorspace on both images
-the math should be checked, it seems that it uses the transform function
for linear and gamma rgb on some places.
-It is very difficult to add a new colorspace, something more generic
should be done here.
-all blend modes and blend if should be available on any colorspace, at
least different than raw

I don't know this interact with your feature, by just looking at the code
it seems that there are no conflicts, so let me know if something should be
done on my side.

edgardoh:
I just fixed some minor bugs and added drag & drop for the module list, it works with SHIFT + click.

edgardoh:
Indeed, I like the equalizer for denoising, and it does a far better job before the base curve.

Did you try it on new edits or have you migrated old ones?

rawfiner:
I tried it on new edits. Though, I made these edits on images that I am used to edit, so I was able to compare the results with the older editings I made

edgardoh:
Great, I just wanted to know the scenario you were testing. Testing this won't be easy, so I'll try to keep track of this things.

Thanks for trying it!

rawfiner:
It seems that when copy/pasting and styles do not preserve the custom module order yet, is it normal in current state of development or should the order be already preserved?

edgardoh:
That's complicated, right now I'm preserving the current behavior: on copy
all / paste all with overwrite mode an exact duplicate of the source
image's history is created, on any other case only the last entry in
history for each instance is copied, and is placed right after the last
instance on the destination image.

I haven't found a good definition to what to do to preserve the order that
the modules have on the original image. The order in the pipe is not
absolute, it depends on the relative position of each module to each other,
so let's say on the source image I have 3 instances of the tone curve,
moved after the colorout, and a lowpass after those tone curves. I copy the
lowpass and paste it into an image that have no tone curve. This means that
the tone curve exists on the destination image, but it is disabled and on
the default order, so where do I place the copied lowpass?

Something similar happens with the styles, so until I find a better way it
uses the same definition as the copy/paste.

aurelienpierre:
@edgardoh I will focus on the next release (doc and stuff) for now. As soon as it's done, I will take a look at your work and merge it with mine. 

In an ideal case, colorin and colorout take whatever is fed to them, and could even act as a pass-through (although editing in camera RGB space sounds like a bad idea, I'm still curious to try it).

At some point, it would help to have an option to reset the pipe order as the legacy, for example to apply presets.

> -It is very difficult to add a new colorspace, something more generic should be done here.

Generally speaking, color spaces conversions are matrix products. Except for Lab and derivatives which do a piece-wise x³ function on top. So, RGB and XYZ transformations can use a generic matrix product, using a general API that takes the right matrix as an input, but the Lab conversions will be a pain and likely suck all the performance. I still wonder if BLAS could be a good option to optimize the RGB transforms/matrix products decided at user level (stuff you can't optimize at compilation time).

Any way, it seems very promising.

edgardoh:
El lun., 26 nov. 2018 a las 18:08, Aurélien PIERRE (<
notifications@github.com>) escribió:

> @edgardoh <https://github.com/edgardoh> I will focus on the next release
> (doc and stuff) for now. As soon as it's done, I will take a look at your
> work and merge it with mine.
>
> In an ideal case, colorin and colorout take whatever is fed to them, and
> could even act as a pass-through (although editing in camera RGB space
> sounds like a bad idea, I'm still curious to try it).
>
> At some point, it would help to have an option to reset the pipe order as
> the legacy, for example to apply presets.
>
> -It is very difficult to add a new colorspace, something more generic
> should be done here.
>
> Generally speaking, color spaces conversions are matrix products. Except
> for Lab and derivatives which do a piece-wise x³ function on top. So, RGB
> and XYZ transformations can use a generic matrix product, using a general
> API that takes the right matrix as an input, but the Lab conversions will
> be a pain and likely suck all the performance. I still wonder if BLAS could
> be a good option to optimize the RGB transforms/matrix products decided at
> user level (stuff you can't optimize at compilation time).
>

What I meant here is that in order to add a new colorspace to the blend
module (HSL, XYZ, etc) all the blend modes routines have to be modified. It
would be better if they all have the form:

-transform forward
-do their thing..
-transform backwards

so the new colorspace will be only a parameter.

rawfiner:
@edgardoh thank you for these explanations, I understand the problem now

rawfiner:
Hi,
FYI, I tried to test this again today, but I get these errors when trying to launch darktable:
[dt_ioppr_check_so_iop_priorities] missing priority for module dummy3
[dt_ioppr_check_so_iop_priorities] missing priority for module dummy2
[dt_ioppr_check_so_iop_priorities] missing priority for module dummy1
And darktable does not start after that

edgardoh:
@rawfiner , you can disable the dummy modules on CMakeLists.txt or you can add a priority to them, i.e: you can copy iop_priorities to your config directory
https://github.com/edgardoh/darktable/blob/mask_iop/iop_priorities

rawfiner:
Ok thanks @edgardoh , I will try that :-)

edgardoh:
Not sure I understand. Right now, on master, priority is an int,
hard-coded, and can change between version. In this implementation priority
is an int and is dynamic, what controls the order of the pipe is iop_order,
a float, that behaves like you described, if module1 is moved between
module2 and module3, then module1 = module2 + (module3 - module2) / 2.f
What you are proposing is to have priority changed to float and hard-coded,
I'm not sure if you mean to have all modules hard-coded or just some, and
then calculate the priorities for the rest, and how it relates to the
iop_order.

rawfiner:
Alright. I did not understand that you were using already a mecanism with this formula. You can forget about my proposition then, which seems equivalent to what you have already done ;-)
I think that with such a formula, copy pasting can be done by simply "merging" the modules of the source and the destination based on the iop_priority values? Or am I missing something?

edgardoh:
Right now the iop_order can be different on different versions, so using
just its value can make a big mess. I had a reason to do it this way, but I
can't remember it now.
Your proposal makes sense, so let's say any issues the iop_order may have
can be handled in a different way and I make it to be the same across
versions, then I can use the iop_order value and modules, when pasted, will
keep the order on the pipe. That would be great, I don't like the current
behavior when modules have been moved. But there's still the problem when
there's a conflict with the iop_order, let's say I create 3 instances of
exposure and I paste it into an image with 3 instances of exposure but with
different names, with the current version I'll get:

exposure 3
exposure 2
exposure 1
exposure c
exposure b
exposure a

with the new behavior I'll get:

exposure 3
exposure c
exposure 2
exposure b
exposure 1
exposure a

probably not what the user expects.

Still, I like better the idea that if I have the equalizer near the
demosaic, when pasted it will be near the demosaic, not after the colorin,
so its just a matter of finding a better way to handle the conflicts of the
iop_order.

rawfiner:
Cool, that seems promising! :-)
I haven't thought about the conflict problem that you raise, that's a good point. In such a case, I agree we should check if there are consecutive instances of the same module, and in this case, paste after the latest instance of this sequence.

Concerning the issue across versions, another approach could be to have a version number attached with the iop_priority list, and "convert" the numbers when we find an old list of iop_priorities. But I think this would be much harder to maintain that having frozen values that don't change across versions.

Also, if we use the "frozen values" method, we should think about which values to use so that we keep enough space for future module additions. As such, maybe (probably) all current values should be changed. We should get uniform intervals between all initial values, as we don't know where the new modules will be added (and 0 should be for the beginning of the pipe, not for the first module, as well as the maximum value which should be for the end of the pipe and not for the last module). To get uniform intervals, I think it would be easier to use an integer instead of a float for iop_order: we want the same number of possible values inside each interval.
Let's do a quick computation of what this would imply, to see if having a frozen system is problematic. Currently, I think we have about 50 modules. Let's consider we currently need 64 intervals, for 62 modules and 2 for begin and end. With iop_priority in an uint32_t, this would make it possible to divide by 2 each interval 26 times in worst case (2^32/64). This is already quite a big number I think. If we want to take more margin, using an uint64_t, this would make it possible to divide by 2 each interval 58 times in worst case (2^64/64). So I think having "frozen values" is not a problem for future module additions. 

Anyway, whatever solution you choose, thanks for all the work :-)

src/develop/develop.h
```diff
@@ -272,6 +288,8 @@ void dt_dev_reload_image(dt_develop_t *dev, const uint32_t imgid);
 int dt_dev_is_current_image(dt_develop_t *dev, uint32_t imgid);
 void dt_dev_add_history_item_ext(dt_develop_t *dev, struct dt_iop_module_t *module, gboolean enable, gboolean no_image);
 void dt_dev_add_history_item(dt_develop_t *dev, struct dt_iop_module_t *module, gboolean enable);
+void dev_add_masks_history_item_ext(dt_develop_t *dev, struct dt_iop_module_t *_module, gboolean _enable, gboolean no_image);
```
TurboGit: dt_dev_add...

src/develop/develop.h
```diff
@@ -272,6 +288,8 @@ void dt_dev_reload_image(dt_develop_t *dev, const uint32_t imgid);
...
+void dev_add_masks_history_item(dt_develop_t *dev, struct dt_iop_module_t *_module, gboolean enable);
```
TurboGit: dt_dev_add_

edgardoh:
@rawfiner , I've done only a very basic test but seems to be working. This still is not ideal but for a first version maybe is good enough, I'll use for a while and see how it goes.

rawfiner:
Thank you @edgardoh,  I will test that soon :-)

rawfiner:
I made some tests with it, this is much better, thanks a lot! :smiley:

LebedevRI:
I have twice now typed a comment for this pr, and discarded it. (posted into wrong pr just now)

TLDR: i don't think i get to comment on the feature itself, but i would like to very strongly highlight
that this has the potential to be the third case (previous were masks, multi-instances)
that was merged [potentially too soon] in a horrible/disgusting state (thus discouraging anyone
and everyone from even touching it, fixing bugs in it), as usual with no tests whatsoever,
and the echo in the form of bugs will be there for a *very* long time.
Given the nature of this change, i suspect this may end up being the most destructive of all three...

aurelienpierre:
What would be useful here is to know what in particular could break, and what to specifically test.

LebedevRI:
By _tests_ i didn't meant the dt's standard approach of give it to 1000's of monkeys and see what broke,
but the more proper approach of actually writing tests that are actually automatically run.

Since there aren't any of these (who needs those anyway, everyone's code is perfect, can't have bugs, won't break any pre-existing stuff/features), i'd guess you would want to triple check all of the new [re-]ordering logic (will be sad if it is subtly broken), all of the migration code (actually verify that the module order will still remain correct), try most of the re-orderings to see that nothing obvious breaks.

(obviously all that should be done under asan; i have not built dt under asan for some time,
so i can't tell if there are some changes needed.)

TurboGit:
Hum, how much pictures to update? Maybe some missing index on some table?

edgardoh:
@aurelienpierre , the sql is heavy but I have tested it with +10.000 images and it was in the order of 10 minutes, not hours. It can probably be optimized but that's not a priority right now, first the design must be validated.

aurelienpierre:
I have like 30000 pictures I think, about 3.6 MB of DB. I have retried several times from different backups.

edgardoh:
Probably is not linear, so that seems about right. I don't know what you mean by did not finish. It crashes? Do you have an error message?

aurelienpierre:
No crash, no message, it just went on for a long time with dt running in background but not showing any window.

edgardoh:
That's normal, you won't see any window until the migration finish. Maybe it just need more time, like I said the sql is heavy.

TurboGit:
Ok, I did build it for the first time, I wanted to see if it works on my side before going into a deeper testing. Some issues:

1. open tone curve
2. select RGB independant channels
3. select [b] 

BUG: see that the [g] and [b] channels have no histogram in background (red does)

5. on the [b] tab and move the curve
6. select scale : log-log

BUG: the scale is not changed

The mask history is great! Masks feel first class citizen now! More later...

edgardoh:
I can't duplicate the first issue, all 3 histograms are displayed fine.
Check that the log scale is not active, histograms are not displayed with
log scale.
Confirmed second issue, seems that log scale works only on the first tab,
very reasonable for lab, not so much for rgb.

My goal with the tone curve and the levels was mainly to add the new
functionality so it can be tested. Now modules can ask for a specific
colorspace, so I want to make sure that the transformation is correct (both
modules ask for a prophotoRGB image), the histogram represents the image on
that colorspace on that point of the pipe, the colorpicker return the
correct values and the blend module keep working as usual. If all that
works then I'm happy.

If we look at the modules itself then I'm not so happy. The histogram and
the edit is done in linear rgb, that is not intuitive for the user and
probably not what is expected. If a linear curve (from the presets) is
applied the middle gray is shifted. That's because the middle gray on
linear rgb is around .18 and not .5, so the math seems correct, but not the
functionality.
Not sure what to do about it.

aurelienpierre:
Histogram in log will be a nightmare to implement (you need to log-encode the image before you count the pixels and put them in the bins).

Also, the RGB curves have their LUT built assuming they are Lab (so UI 50 % = middle grey), and then, the LUT are converted to RGB (so middle grey = 18 % internally). Indeed, the behaviour is as expected but the controls are weird. I think the other softwares I have tested put the middle grey at 50 % also for the RGB curves, so I think it works.

TurboGit:
We have two big features in there: mask history and module order.

Do you think it could be possible create 2 PR?

This would certainly help testing and integration.

edgardoh:
I started it as two developments, but at some point they started to overlap, I don't remember exactly where, so I continue as a single development. Right now they cannot be split without breaking things.

rawfiner:
@edgardoh If I understand correctly, colorin would have a fixed place in the pipe?
I am not confident in the fact that denoise profile results would remain the same if denoise profile is done after colorin, thus we should test this and see how to handle backward compatibility if the results are differents.
Apart of that, it seems a good idea to me, even though I am clearly not an expert about colorspaces :-)

edgardoh:
> @edgardoh <https://github.com/edgardoh> If I understand correctly,
> colorin would have a fixed place in the pipe?
>
Yes, so colorout, at least on a first version. Its a limitation, but if
some module gives better results before colorin or after colorout that can
be handled. The user won't be able to move it, but again, is a limitation.

> I am not confident in the fact that denoise profile results would remain
> the same if denoise profile is done after colorin, thus we should test this
> and see how to handle backward compatibility if the results are differents.
>
Backward compatibility will be handled, previous edits will have
colorin/colorout at the same place in the pipe. As for the different
results, that's expected, right now color denoise behaves different
before/after colorin, I can't tell if better or worst, that's probably
subjective.

> Apart of that, it seems a good idea to me, even though I am clearly not an
> expert about colorspaces :-)
>
Me neither, but there's seems to be a consent that is better to do the
edits on the working colorspace, and it should RGB linear, so this is a
first step on that direction.

rawfiner:
Thanks for the answer :-)
Seems good to me.

edgardoh:
I just pushed the working colorspace feature and updated the first post with the new changes.

I didn't lock colorin and colorout so they can still be moved to get the current pipe order, but is user responsability not to move Lab modules outside colorin-colorout range (see the first post).

edgardoh:
Just added the working profile to the histogram profile and the sse and opencl version of the colortransforms, so that's it for the main features, what's left is to test it to the dead.

data/kernels/color_conversion.cl
```diff
@@ -0,0 +1,185 @@
...
+    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+    GNU General Public License for more details.
+
+    You should have received a copy of the GNU General Public License
+    along with darktable.  If not, see <http://www.gnu.org/licenses/>.
+*/
+
+#include "common.h"
+
+// must be in synch with dt_iop_colorspace_type_t in imageop.h
```
TurboGit: I suppose that ideally we'd want to move this definition into imageop_common.h and include this here and in imageop.c.

data/kernels/rgbcurve.cl
```diff
@@ -0,0 +1,90 @@
...
+        ratio = curve_lum / lum;
+      }
+      
+      pixel.x = ratio * pixel.x;
+      pixel.y = ratio * pixel.y;
+      pixel.z = ratio * pixel.z;
+    }
+    else
+    {
+      pixel.x = lookup_unbounded(table_r, pixel.x, coeffs_r);
```
TurboGit: looks suspicious that table_r and coeffs_r is used for the x, y & z values.

src/common/iop_order.h
```diff
@@ -0,0 +1,193 @@
...
+#endif
+
+struct dt_iop_module_t;
+struct dt_develop_t;
+struct dt_dev_pixelpipe_t;
+
+typedef struct dt_iop_order_entry_t
+{
+  double iop_order;
+  char operation[20];
```
TurboGit: current max for an operation name is 16, 20 seems ok but a bit short maybe?

src/common/iop_order.c
```diff
@@ -0,0 +1,2418 @@
...
+ *   modules can be added/moved in the current version
+ * - a module can't be more than once on the same version
+ */
+static int _ioppr_legacy_iop_order_step(GList **_iop_order_list, GList *history_list, const int old_version, const int dont_move)
+{
+  int new_version = -1;
+  
+  if(0)
+  {
+    _ioppr_insert_iop_before(_iop_order_list, history_list, "dummy1", "colorin", dont_move);
```
TurboGit: what is dummy1 and dummy2? was this for debugging? if it is the case shouldn't it be removed?

src/common/iop_order.c
```diff
@@ -0,0 +1,2418 @@
...
+    _ioppr_insert_iop_after(_iop_order_list, history_list, "basicadj", "colorin", dont_move);
+    _ioppr_insert_iop_after(_iop_order_list, history_list, "rgbcurve", "hazeremoval", dont_move);
+    _ioppr_insert_iop_before(_iop_order_list, history_list, "rgblowpass", "lowpass", dont_move);
+    
+    // _ioppr_move_iop_before(_iop_order_list, "lens", "exposure", dont_move);
+    // _ioppr_move_iop_after(_iop_order_list, "tonecurve", "levels", dont_move);
+
+    new_version = 2;
+  }
+/*  // version 2 --> 3
```
TurboGit: maybe remove this commented out code?

src/develop/format.c
```diff
@@ -79,21 +59,38 @@ void default_input_format(dt_iop_module_t *self, dt_dev_pixelpipe_t *pipe, dt_de
...
+  if(dt_ioppr_get_iop_order(pipe->iop_order_list, self->op) >= dt_ioppr_get_iop_order(pipe->iop_order_list, "rawprepare")) return;
 
   if(piece->pipe->dsc.filters)
     dsc->datatype = TYPE_UINT16;
 }
 
+int default_input_colorspace(dt_iop_module_t *self, dt_dev_pixelpipe_t *pipe,
+                     dt_dev_pixelpipe_iop_t *piece)
+{
+  return self->default_colorspace(self, pipe, piece);
```
TurboGit: is that expected that this is self->default_coulorspace and not self->input_colorspace().
likewise for the two others implementations below (->blend_colorspace() and ->output_colorspace())

TurboGit:
Also, rgbcurve is crashing as soon as you try to move the curve.

TurboGit:
I think that the rgbcurve and rgblowpass module should be removed from this PR. If really needed, let's have them part of a new PR.

TurboGit:
In old edits with masks a "mask manager" entry is displayed in the history? Is that expected? I think yes reading the code, but I prefer asking. Also, why is this not needed for new edits?

tools/iop_dependencies.py
```diff
@@ -510,8 +510,13 @@ def add_edges(gr):
   gr.add_edge(('colorbalance', 'vibrance'))
```
TurboGit: We probably want to delete this file since the module->priority has been removed from all iop, right?

src/iop/CMakeLists.txt
```diff
@@ -81,6 +81,15 @@ add_iop(gamma "gamma.c")
 add_iop(temperature "temperature.c" DEFAULT_VISIBLE)
 add_iop(colorcorrection "colorcorrection.c" DEFAULT_VISIBLE)
 add_iop(exposure "exposure.c" DEFAULT_VISIBLE)
+add_iop(basicadj "basicadj.c" DEFAULT_VISIBLE)
+add_iop(rgbcurve "rgbcurve.c" DEFAULT_VISIBLE)
+add_iop(rgblowpass "rgblowpass.c" DEFAULT_VISIBLE)
+#add_iop(dummy1 "dummy1.c" DEFAULT_VISIBLE)
```
TurboGit: I would remove all those commented out lines.

src/develop/pixelpipe_hb.c
```diff
@@ -1301,6 +1295,7 @@ static int dt_dev_pixelpipe_process_rec(dt_dev_pixelpipe_t *pipe, dt_develop_t *
     if(dt_opencl_is_inited() && pipe->opencl_enabled && pipe->devid >= 0)
     {
       int success_opencl = TRUE;
+      int input_cst_cl = input_format->cst;
```
TurboGit: int -> dt_iop_colorspace_type_t ?

TurboGit:
The mask_history table is updated only when switching images or quitting darktable. This means that if a crash occurs we loose precious edits. Maybe we can write the masks history when switching modules? But this is already the behavior for standard history, so maybe better to do that part of a new PR.

src/iop/dummy1.c
```diff
@@ -0,0 +1,142 @@
+/*
```
TurboGit: To be removed. It is commented out in CMakeList.txt.

The same apply to amm dummy?.c

src/iop/rawoverexposed.c
```diff
@@ -75,7 +75,12 @@ int default_group()
 
 int flags()
 {
-  return IOP_FLAGS_ALLOW_TILING | IOP_FLAGS_HIDDEN | IOP_FLAGS_ONE_INSTANCE | IOP_FLAGS_NO_HISTORY_STACK;
+  return IOP_FLAGS_ALLOW_TILING | IOP_FLAGS_HIDDEN | IOP_FLAGS_ONE_INSTANCE | IOP_FLAGS_NO_HISTORY_STACK | IOP_FLAGS_FENCE;
+}
+
+int default_colorspace(dt_iop_module_t *self, dt_dev_pixelpipe_t *pipe, dt_dev_pixelpipe_iop_t *piece)
+{
+  return iop_cs_rgb;
```
TurboGit: this is really rgb and not RAW?

data/kernels/color_conversion.cl
edgardoh: Yes, all color things should be somewhere else, but I'm not sure exactly where, we can discuss it.

data/kernels/rgbcurve.cl
edgardoh: This is the linked mode, all rgb use one curve.

src/common/iop_order.h
edgardoh: this is a
https://github.com/darktable-org/darktable/blob/master/src/control/settings.h#L35

seems that this is new, I'll change it.

src/common/iop_order.c
```diff
@@ -0,0 +1,2418 @@
...
+    fprintf(stderr, "[_ioppr_move_iop_before] next module %s don't exists on iop order list\n", op_next);
+
+  *_iop_order_list = iop_order_list;
+}
+
+// moves op_current after op_prev by updating the iop_order
+// only if dont_move == FALSE
+static void _ioppr_move_iop_after(GList **_iop_order_list, const char *op_current, const char *op_prev, const int dont_move)
+{
+  if(dont_move) return;
```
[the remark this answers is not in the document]
edgardoh: The idea is to have only one line in the caller to hide the implementation. The more logic we have, more prone to errors.

src/common/iop_order.c
```diff
@@ -0,0 +1,2418 @@
...
+    _ioppr_insert_iop_before(&iop_order_list, history_list, op_new, prior_next->operation, check_history);
+
+  *_iop_order_list = iop_order_list;
+}
+
+// moves op_current before op_next by updating the iop_order
+// only if dont_move == FALSE
+static void _ioppr_move_iop_before(GList **_iop_order_list, const char *op_current, const char *op_next, const int dont_move)
+{
+  if(dont_move) return;
```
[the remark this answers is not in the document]
edgardoh: See previous...

src/common/iop_order.c
```diff
@@ -0,0 +1,2418 @@
...
+    _ioppr_insert_iop_before(_iop_order_list, history_list, "dummy1", "colorin", dont_move);
+    _ioppr_insert_iop_after(_iop_order_list, history_list, "dummy2", "colorin", dont_move);
+    _ioppr_move_iop_after(_iop_order_list, "colorin", "demosaic", dont_move);
+    _ioppr_move_iop_before(_iop_order_list, "colorout", "finalscale", dont_move);
+  }
+
+  // version 1 --> 2
+  if(old_version == 1)
+  {
+    _ioppr_move_iop_after(_iop_order_list, "colorin", "demosaic", dont_move);
```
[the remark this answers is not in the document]
edgardoh: Idem.

src/develop/format.c
edgardoh: default_colorspace is mandatory, the other are optional and the default_colorspace is the default (if they are not implemented)

tools/iop_dependencies.py
edgardoh: OK

src/develop/pixelpipe_hb.c
edgardoh: This happens a lot, I'll fix it as I find them.

src/iop/rawoverexposed.c
edgardoh: Yes, this is after the colorout, so the input is an rgb image.

edgardoh:
@TurboGit , all new modules are for testing the new features and while in testing better to keep them, removing them is easy.
The basic adjustments I like and I want to keep it.
The rgb lowpass I don't like and I'll eventually delete it.
The rgb curve I like, we can discuss if keep it or merge it with the tone curve. For me is better to keep it.

I haven't test the rgb curve since the last rebase, there's probably some issues. I'm working on the color zones now, I'll check as soon as I can.

The mask manager entry is used when a mask without a module is modified but also on some other cases where a single module is not available, like the migration, compress history, and a couple more that I don't remember.

TurboGit:
@edgardoh : I'm not against the module, I think the basicadj is nice for begginner and I'd like to have it. But I would really prefer to have all those modules in separate PR.

It will make this PR smaller. If we need to revert this PR temporarily we will keep the iop. And all the dummy for for testing should just do in a specific branches. I don't want to have this in the master.

edgardoh:
> @edgardoh <https://github.com/edgardoh> : I'm not against the module, I
> think the basicadj is nice for begginner and I'd like to have it. But I
> would really prefer to have all those modules in separate PR.
>
> It will make this PR smaller. If we need to revert this PR temporarily we
> will keep the iop. And all the dummy for for testing should just do in a
> specific branches. I don't want to have this in the master.
>
Sure, I'm not saying to promote this with all those modules, just to keep
them while testing, then I'll remove them all and add a new PR for the ones
I care about.

TurboGit:
Ok, so please remove them now. I have tested this a lot, I have make many reviews so I think it is time to bit the bullet and merge this. Better sooner than later for this to be more widely tested now.

edgardoh:
> Ok, so please remove them now. I have tested this a lot, I have make many
> reviews so I think it is time to bit the bullet and merge this. Better
> sooner than later for this to be more widely tested now.
>
Ups, I just pushed a WIP version of the color zones! I'll revert it, do
some cleanup and let you know.

edgardoh:
@TurboGit , should be ready now, please check it.

TurboGit:
Ok, this is now merged. I have added a commit on top of this as the migration was not done after 40 minutes on my side on my full database (~ 55k pictures). With the new index it is done in 10 seconds.

Thanks again for the hard work! This is a really nice move.

edgardoh:
@TurboGit , there's a couple of items on my todo list:

- colorin and colorout can be moved. I'm not really sure about this, I don't see the point of having modules before/after them, but since up till now it was possible I let it like this. The problem is that this generates some special cases and if it won't be used better to have them fixed.

- color routines don't belong in iop_order.c, we should define some library for this, but colorspaces.c doesn't seem the right place either. Something similar with opencl.

- undo is not working in master, I tried to follow it but I have no idea what's going on. Does not seem to be related to this PR, but I can't be sure.

TurboGit:
For undo any procedure to reproduce? I had tested that yesterday and it was working fine a quick test this morning shows that it still works for me.

TurboGit:
I see the undo issue now! ctrl-z does not work until ctrl-y is used, then all seems fine... I'll look at this.

rawfiner:
I think denoiseprofile should always remain before colorin (at least if the user wants accurate profiles).
Otherwise, we would need to make noise profiles for every input profile and every working profile.

moy:
About the "after/before color in/out", dithering should be after color out (it essentially does a clever rounding of float values to integers, so it must come after anything that change the float values). I can't think of any other module that really should be after color out.

edgardoh:
I didn't give too much thoughts to the module order, but there are more than a few that can be changed. The channel mixer should not be last, the denoise(s) should be first, the levels and tone curve should be swapped.
Maybe some rules should be added too, I'm not sure if some distord modules need a specific order.
I'm working on other things now and keeping an eye on the bug reports and I don't want to add any more, but maybe the best way to handle this is to create a PR to centralize the discussion there.
My two cents.

TurboGit:
This has broken the perspective correction, see:
https://github.com/darktable-org/darktable/issues/2201

phweyland:
> modules now ask for the input colorspace and informs the output one. This can be dynamic, so Lab modules with an RGB option can ask for an RGB input image

beside the common `default_colorspace()` we have the 2 routines:
```
- input_colorspace()
- output_colorspace()
```
When are they triggered ?
The background of the question is that, depending on user settings in a module, the needed colorspace can be different. For example in old fashion tonecurve, I want to call RGB only if the user select the rgb mode. For lut3d I have a similar need depending of the type of lut.
How should I do ?

Talking about lut3d, you cover already the sRGB colorspace. Would it be easy to add the the gamma REC709 ? Same question, but I guess it's more difficult, for LOG RGB ? What do you recommend ?

edgardoh:
> modules now ask for the input colorspace and informs the output one. This
> can be dynamic, so Lab modules with an RGB option can ask for an RGB input
> image
>
> beside the common default_colorspace() we have the 2 routines:
>
> - input_colorspace()
> - output_colorspace()
>
> When are they triggered ?
>
Don't really remember, but when the calling program need to know the input
or output colorspace for a module, for sure in the pipe.

> The background of the question is that, depending on user settings in a
> module, the needed colorspace can be different. For example in old fashion
> tonecurve, I want to call RGB only if the user select the rgb mode. For
> lut3d I have a similar need depending of the type of lut.
> How should I do ?
>
You add those functions on the iop, like the default_colorspace(), only
difference is that you'll return the colorspace based on the parameters (I
guess). Notice that pipe and piece can be NULL, so you have to handle that
scenario.

> Talking about lut3d, you cover already the sRGB colorspace. Would it be
> easy to add the the gamma REC709 ? Same question, but I guess it's more
> difficult, for LOG RGB ? What do you recommend ?
>
There's a limitation on the work profile, only those that can be used by
the internal transform function are accepted. This is for performance,
lcms2 is slow. You can check here how this is determined:
https://github.com/darktable-org/darktable/blob/master/src/common/colorspaces.c#L137

This is for profiles from file, the ones in-memory must be added by hand,
on that same colorspaces.c

junkyardsparkle:
So... I just realized that it's now impossible to export the weird Lab pfm format [needed by `darktable-chart`](https://www.darktable.org/usermanual/en/darktable_chart_making_input_images.html)... any ideas for a workaround?

phweyland:
> You add those functions on the iop, like the default_colorspace(), only difference is that you'll return the colorspace based on the parameters

in fact input_colorspace() just let you choose between iop_cs_lab and iop_cs_rgb. In case of iop_cs_rgb it uses the rgb working profile selected in colorin. 
If that's right, a module (like lut3d) cannot choose real time between sRGB, rec709, prophotorgb or whatever. Is my understanding correct ?

Then I've to provide somewhere transforms between prophotorgb and the other rgb colorspaces. Is colorspaces_inline_conversions.h still the right place ?

edgardoh:
That's correct, the input image is either Lab or RGB with the work profile
applied, that's the point of the work profile. If you want a particular RGB
colorspace the module has to do the transformation.

I have no idea what you need to do, but having this transforms hard-coded
doesn't seem a good idea, is there a way to do it dynamic? If you create
in-memory profiles it should be possible (and not too difficult).

phweyland:
> If you create in-memory profiles it should be possible (and not too difficult).

Sorry I don't get that. Could you develop a little bit ?  Is there already an example of that ?

edgardoh:
What I call in-memory profiles are hard-coded profiles defined in
colorspaces.c. The sRGB, linear rec2020, etc that you see in the profiles
list are such profiles. They are added here:

https://github.com/darktable-org/darktable/blob/master/src/common/colorspaces.c#L1185

you can check on the same file how they are defined.

So, if you create your profiles on the same way they can be added to a list
so the user can select them or you can have them hard-coded, depending on
your needs. Then, the same routines that I have created to do the
transforms can be used.

edgardoh:
Have in mind that all this work only if the module is between the colorin
and colorout, as it needs a profile that describes the image (that's what
the colorin does). If you also want your module to work before the colorin
you'll have to define a transform between camera RGB and the colorspace of
your choice.

phweyland:
In colorspaces.c I can see the creation of several colorspaces. They end up into a list of lcms profiles.

Some of them are considered by iop_order.c (colorspaces transform) which prepares the data to transform iop_cs_lab <-> iop_cs_rgb. Pixelpipe uses dt_ioppr_transform_image_colorspace() to execute this transformation.

As from module standpoint we have only the below choices (but only raw, lab and rgb are implemented if I understand properly):
```
/** colorspace enums */
typedef enum dt_iop_colorspace_type_t
{
  iop_cs_NONE = -1,
  iop_cs_RAW = 0,
  iop_cs_Lab = 1,
  iop_cs_rgb = 2,
  iop_cs_LCh = 3,
  iop_cs_HSL = 4
} dt_iop_colorspace_type_t;
```
I cannot use this mechanic (dt_ioppr_transform_image_colorspace()) to transform prophotorgb into gamma rec709 for example.

Could I use instead cmsCreateTransform, cmsDoTransform and cmsDeleteTransform ? I've also seen that in pixelpipe for color picker and histogram stuff. It seems that's the only way to use colorspaces.c lcms color profiles directly from iop module ... but maybe not fast enough ...

Otherwise I'm back to hard-coded transforms in colorspaces_inline_conversions.h...

I'm a bit concerned to have several ways / places to make colorspaces transform.

edgardoh:
You have more function to perform transforms on iop_order.c, the RGB <-->
RGB is already defined, so if you create an in-memory profile you don't
have to worry about that, but first you have to decide if you'll create
such profiles.

phweyland:
> but first you have to decide if you'll create such profiles.

If this is the recommended way I'll do this.

> on iop_order.c, the RGB <--> RGB is already defined

I've missed that ... could you help me ?

edgardoh:
This one transforms the entire image RGB <--> RGB:

https://github.com/darktable-org/darktable/blob/master/src/common/iop_order.h#L145

But it needs the from/to profiles, that can be in-memory or from file
profiles, in both cases the matrix should be available as I described
earlier (not sure if it was in this thread). It can work with lcms2, but it
will be too slow.

The per-pixel version is not exposed, but if you need it I can add it.

I still don't know what you want to do, so I can't recommend anything, but
this way if you want to add a new colorspace you only need to create the
in-memory profile, and it can also work with user defined profiles (from
file).

phweyland:
> iop_order.h

Ooops, I haven't visited this one. Thanks.

> I still don't know what you want to do

For the time being I try to implement the module to apply 3D lut on image. But these luts are valid for a given colorspace, which can be sRGB, gamma REC709 and LOG. So before applying the lut the user has to select the corresponding colorspace (it may not come with the lut).
The LOG is a particular case. In film production it can come out the camera and can be a work profile. So I'll try to place the module before colorin, transform the raw data into log and apply the luts which work in LOG profile. That should work if the module gives as output profile, an existing colorin profile.

edgardoh:
If you place it before colorin and change the colorspace the you'll have to
change the colorin as well, because it expects camera RGB. Or maybe just
disable it, but in any case it should be carefully considered, colorin is a
very sensitive module, and a lot of things are based on that.

If you place it after the colorin, maybe you can use the work profile as
input, and just display the luts that can work with that profile. Or if the
lut just accept one profile then there's no need for the user to select it,
there's only one option. But I don't know the design, so take this just as
wild ideas.

Again, in any case, if you place it after the colorin and change the
colorspace you'll have to set the work profile so the rest of the pipe can
handle it, and for that the profile matrix is needed.

phweyland:
I've some difficulties with profile transform.
So far, process() works fine with native prophotorgb.
But when I try to transform prophotorgb that doesn't work and I don't find my mistake(s). Here a piece of process(). Is there any obvious mistake I cannot see ?

```
  const dt_iop_order_iccprofile_info_t *const dest_profile
      = colorspace == DT_IOP_REC709 ? dt_ioppr_add_profile_info_to_list(self->dev, DT_COLORSPACE_REC709, "", INTENT_PERCEPTUAL)
      : colorspace == DT_IOP_SRGB ? dt_ioppr_add_profile_info_to_list(self->dev, DT_COLORSPACE_SRGB, "", INTENT_PERCEPTUAL)
      : colorspace == DT_IOP_LIN_REC709 ? dt_ioppr_add_profile_info_to_list(self->dev, DT_COLORSPACE_LIN_REC709, "", INTENT_PERCEPTUAL)
      : NULL;
  const dt_iop_order_iccprofile_info_t *const rgb_profile = dt_ioppr_get_iop_work_profile_info(self->dev);

  if (colorspace == DT_IOP_REC709 || colorspace == DT_IOP_SRGB || colorspace == DT_IOP_LIN_REC709)
  {
    dt_ioppr_transform_image_colorspace_rgb(ibuf, obuf, roi_out->width, roi_out->height, rgb_profile, dest_profile, "linrec2020 to specific");
    return;
  }
```
I expect to get a new image with the effect of the colorspace transform. Nothing psychedelic and unstable (not repetitive) ... :-)

image

image

edgardoh:
Seems correct, do you have this on a PR, updated with this changes so I can
play with it?

phweyland:
> do you have this on a PR, updated with this changes so I can play with it?

I've just pushed the current state (PR #2128 )

rabauke:
This has reverted some changes which had been merged before via my pull request https://github.com/darktable-org/darktable/pull/1890 . Consequently, the bug which had been fixed by the mentioned pull request appears again. I do not see how _custom module order and masks history_ require to revert changes to fix issues with module renaming. Probably a merge conflict has been resolved erroneously.

houz:
Dragging modules with shift key breaks the old behaviour that you can expand more than one module at a time. That needs to be fixed.

houz:
I fixed it by assigning ctrl-shift to dragging.

PR #1842, opened 2018-11-23
Update documentation for denoising modules: would like some feedback

rawfiner:
This is a work in progress, for updating the doc for denoising modules.
As a side effect, the channel mixer is impacted too as it is useful when using the RGB curves, thus I wanted to make more clear how to use it for that.
Description of "curves" is the same for rawdenoise and denoiseprofile (no need to bother reading twice the same thing).

I am looking for feedback here: what is fine, and what is not clear (and if possible, how to make things more clear). Also, if you have better sentences for some parts, I take that too, as I am not a native speaker.

Thanks :-)

### DIFF
--- a/doc/usermanual/darkroom/modules/color/channelmixer.xml
+++ b/doc/usermanual/darkroom/modules/color/channelmixer.xml
@@ -47,7 +47,9 @@
 
     <para>
       First select your output channel and then set the amount each input channel feeds into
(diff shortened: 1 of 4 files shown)


upegelow:
Cool. Very well written!

rawfiner:
Thank you for the review!

PR #1844, opened 2018-11-24
Filmic - b&w synchronized with grey

phweyland:
When one changes the grey point, black and white settings are lost. 
This change keeps black and white points where they are when the grey changes, adapting their EV value.


aurelienpierre:
Sounds good to me, thanks !

TurboGit:
Thanks.

PR #1845, opened 2018-11-25
BUGFIX: tonecurve, GTK notebook not updated on GUI reset

aurelienpierre:
Fix a bug where the notebook holding the L/a/b tabs was not reset to the first L tab when resetting the module (reported by @cryptomilk).

Also, reintroduce the Lab histogram in background for all non-log views, since the UI controls are always Lab anyway.


TurboGit:
Thanks!

PR #1846, opened 2018-11-25
Colorbalance: fixes and improvements

aurelienpierre:
1. Some code cleaning on the debug preprocessor instructions
1. Fix the computations for the auto-tuners that went wrong since the last change on ASC CDL algo (on the power parameters)
1. Rework the HSL <-> RGB conversions (in UI sliders only) to make them 100 % reversible (before, there was a small error), now 100 % saturation leads to 50 % RGB primaries and the other way around
1. Remove the preset for the log profile compensation, since it not really recommended to do it there
1. Not sure about the last one: add an additional saturation correction after the CDL (the current one is before), to comply with ASC recommendations. This is arguably a new feature, so, since we are in feature freeze, worse case you can revert that last commit.

The input saturation can be used to desaturate the image to an almost B&W image, and then the CDL can be used as a split-toning. 

When using the CDL to revert parasite color casts, the input saturation can be used to resaturate the input so some of the original color cast is preserved (soften the correction).

The output saturation can be used to make the CDL effect stronger without changing hues (without that parameter, one would have to push the saturation of slope/offset/power channels separately, which may result in changes hues/degradation of the settings).


TurboGit:
Sounds good and to me the last point is more like a fix than an enhancement. I'm one having trouble add saturation and with this new slider I get the result I want easily. Thanks for the work on this Aurélien.

TurboGit:
With this last version the 2 remaining presets are not working. It just does nothing.

aurelienpierre:
Yes, I saw the issue. I will fix that today hopefully

PR #1848, opened 2018-11-26
Update docs

TurboGit:
for 2.6

### DIFF
--- a/doc/usermanual/lighttable/concepts/filtering.xml
+++ b/doc/usermanual/lighttable/concepts/filtering.xml
@@ -56,7 +56,10 @@
     Images in the lighttable view can be displayed in different sort orders, depending on
     <quote>filename</quote>, <quote>time</quote> (when the photo was taken),
     <quote>rating</quote> (i.e. stars), <quote>id</quote> (an internal ordering number of
-    darktable), or <quote>color labels</quote>. You can sort in reversed order by toggling the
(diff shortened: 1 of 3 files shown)


doc/usermanual/lighttable/concepts/filtering.xml
```diff
@@ -56,8 +56,10 @@
     Images in the lighttable view can be displayed in different sort orders, depending on
     <quote>filename</quote>, <quote>time</quote> (when the photo was taken),
     <quote>rating</quote> (i.e. stars), <quote>id</quote> (an internal ordering number of
-    darktable), or <quote>color labels</quote>. You can sort in reversed order by toggling the
-    triangle button to the right of the <quote>sort by</quote> combobox.
+    darktable), <quote>color labels</quote>, <quote>groups</quote>, <quote>full path</quote>,
+    <quote>title</quote>, <quote>description</quote>, <quote>aspect ratio</quote> or
+    <quote>custom order</quote> (user can drag &amp; drop images). You can sort in reversed order
```
upegelow: Playing a bit around with this feature it looks to me that you can only change custom order if the lighttable view is in file manager mode. When in zoomable lighttable mode you only pan the whole set of images.

TurboGit: Right. Will update the doc.

PR #1850, opened 2018-11-27
filmic: add some borders to area to gain so vertical space.

TurboGit:
I find the area in filmic a bit big especially since it is not sensitive. Using a smaller curve help gaining so vertical space. What do you think?


aurelienpierre:
Good idea.

houz:
Hi, I never liked the tiny curve widget, so instead I made it non-square. What do you guys think about it?
see 4e25690c1f65d1ba1618578d1ba39652edca5f16

aurelienpierre:
The benefit of a square graph is the identity line it easy to read : it's the diagonal. I think we lose some of the clarity with a rectangular design. That said, filmic goes from scene-referred EV (x axis) to display-referred % (y axis), so the identity line does not mean much, and there is no neutral setting anyway. 

I think I'm good both ways.

PR #1851, opened 2018-11-27
filmic : improvements

aurelienpierre:
* performance: put all the desaturation mask evaluation out of the pixel loops (×2-×3 speed-up)
* processing: change the desaturation mask to a gaussian window to prevent the transitions issues between desaturated and non-desaturated areas
* processing: add a new "preserve chroma" mode for users who complained filmic desaturated too much. The chroma preservation ensures the color information in linear space is retained between the input and the output. This will however produce over-saturated pictures and needs to be adjusted in color balance with -50 % saturation. This does not ensure the color preservation is perceptualy accurate.
* UI: make the saturation slider logarithmic, so the big changes don't happen only at low values.

### DIFF
--- a/src/iop/filmic.c
+++ b/src/iop/filmic.c
@@ -878,6 +878,7 @@ static void saturation_callback(GtkWidget *slider, gpointer user_data)
   dt_iop_module_t *self = (dt_iop_module_t *)user_data;
   if(self->dt->gui->reset) return;
   dt_iop_filmic_params_t *p = (dt_iop_filmic_params_t *)self->params;
+  dt_iop_filmic_gui_data_t *g = (dt_iop_filmic_gui_data_t *)self->gui_data;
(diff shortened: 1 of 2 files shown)


src/iop/filmic.c
```diff
@@ -169,6 +174,54 @@ int flags()
...
+      int interpolator;
+    } dt_iop_filmic_params_v1_t;
+
+    dt_iop_filmic_params_v1_t *o = (dt_iop_filmic_params_v1_t *)old_params;
+    dt_iop_filmic_params_t *n = (dt_iop_filmic_params_t *)new_params;
+    dt_iop_filmic_params_t *d = (dt_iop_filmic_params_t *)self->default_params;
+
+    *n = *d; // start with a fresh copy of default parameters
+
+    n->grey_point_source = o->grey_point_source;
```
TurboGit: you are missing black_point_source

aurelienpierre: Thanks

TurboGit:
Looks good now. Thanks.

PR #1852, opened 2018-11-27
Add an entry in usermanual for contextual help

rawfiner:
This adds an explanation about contextual help in the usermanual


doc/usermanual/overview/overview.xml
```diff
@@ -1055,6 +1055,20 @@
...
+    <sect2 status="final" id="contextual_help">
+
+      <title>Contextual help</title>
+
+      <para>
+        A button
+        <inlinegraphic fileref="&icon_contextual_help;" scalefit="1" width="2%" align="center" />
+        located in the upper panel allows you to open the online usermanual for various elements
+        of darktable. Click on this button, then click on the element for which you want help.
+        The related usermanual entry will open in your web browser.
```
upegelow: Out of curiosity: in a typical Linux environment where is the web browser defined? In my case for some strange reason this features opens up the eric6 Python IDE :)

rawfiner: I am afraid I don't know... the code calls gtk_show_uri or gtk_show_uri_on_window, which are supposed to open the right app depending on the uri (it should open the file browser if it is the uri of a file, the web browser if the uri is a url, etc...)

upegelow: > Try putting in your `~/.bashrc` something like `export BROWSER=firefox`

Did not help. In the end I needed to edit ~/.local/share/applications/mimeapps.list to get it working with firefox.

PR #1853, opened 2018-11-27
filmic + colorbalance: update presets

aurelienpierre:
colorbalance: remove the presets that are a burden to maintain (assert size error on the mask & blendop)
filmic: update default parameters to more usual values

### DIFF
--- a/src/iop/colorbalance.c
+++ b/src/iop/colorbalance.c
@@ -233,50 +233,6 @@ int legacy_params(dt_iop_module_t *self, const void *const old_params, const int
   return 1;
 }
 
-// taken from denoiseprofile.c
-static void add_preset(dt_iop_module_so_t *self, const char *name, const char *pi, const char *bpi, const int blendop_version)
-{
-  int len, blen;
-  uint8_t *p  = dt_exif_xmp_decode(pi, strlen(pi), &len);
-  assert(len == sizeof(dt_iop_colorbalance_params_t));
-  uint8_t *bp = dt_exif_xmp_decode(bpi, strlen(bpi), &blen);
-  if(blendop_version != dt_develop_blend_version())
-  {
-    // update to current blendop params format
-    void *bp_new = malloc(sizeof(dt_develop_blend_params_t));
-
-    if(dt_develop_blend_legacy_params_from_so(self, bp, blendop_version, bp_new, dt_develop_blend_version(),
-      blen) == 0)
-    {
-      free(bp);
-      bp = bp_new;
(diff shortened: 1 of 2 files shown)


src/iop/colorbalance.c
```diff
@@ -233,50 +233,6 @@ int legacy_params(dt_iop_module_t *self, const void *const old_params, const int
...
-    dt_gui_presets_add_with_blendop(name, self->op, self->version(),
-                                    p, len, bp, 1);
-  free(bp);
-  free(p);
-}
-
-void init_presets(dt_iop_module_so_t *self)
-{
-  // these blobs were exported as dtstyle and copied from there:
-  add_preset(self, _("split-toning teal-orange (2nd instance)"),
```
TurboGit: Too bad :( I've not followed the discussion, but the blendop legacy support does not work?

TurboGit: I mean we have the same circuitry in denoisprofile for example where it seems to work, no?

src/iop/colorbalance.c
```diff
@@ -233,50 +233,6 @@ int legacy_params(dt_iop_module_t *self, const void *const old_params, const int
...
-    else
-    {
-      free(bp);
-      free(bp_new);
-      bp = NULL;
-    }
-  }
-
-  if(p && bp)
-    dt_gui_presets_add_with_blendop(name, self->op, self->version(),
```
TurboGit: self->version() here seems strange, it means that the version of the param is always set to be the latest version of the module. It should be the version of the param at the time the preset param was created I think?

src/iop/colorbalance.c
aurelienpierre: not for hard-coded presets it seems. Everytime the blendop API is updated, the presets become invalid and raise a size error

src/iop/colorbalance.c
aurelienpierre: That might be the problem. Honestly, I just took that code from the denoiseprofile IOP, I didn't give it a lot of thought

TurboGit: The code in denoiseprofile do use version which is passed to the add_preset().

src/iop/colorbalance.c
aurelienpierre: anyway, the concept of having static blobs exported from somewhere else in the code is not the cleanest design.

aurelienpierre:
I will try to find a better option

TurboGit:
@aurelienpierre, see my fix it was indeed the params itself the issue using wrong version and not the blend params which were properly handled.

PR #1854, opened 2018-11-28
filmic: bugfix

aurelienpierre:
Fix a bug in the sanitization of the parameters that led to the balance shadows/highlights being clamped between [0;50] % instead of [-50;50] %, so the shadows lift was actually ineffective.

Make the shoulder/toe nodes calculation more straight-forward (the result should be the same though).

Adjust sliders ranges and defaults params so that defaults are in the middle of the ranges


TurboGit:
Thanks.

PR #1855, opened 2018-11-28
[RFC] liquify: fix off-by-one access

moy:
Found with valgrind. When distance == 0, the code was doing 2 accesses in an array of 1 float.

This fixes the symptom, but I'm not sure whether `distance == 0` is supposed to happen. I guess it can't, and then a proper fix would be at the root of the issue to make sure `distance > 0`, and probably an `assert(distance > 0)` should be added here. @TurboGit, what do you think?


TurboGit:
We may add an assert() sure. I still don't understand how this can happen though :(

PR #1859, opened 2018-11-30
Update contextual help links for new modules/features

rawfiner:
This adds urls for:
- retouch
- filmic
- contextual help button

Note I am anticipating a little here: filmic help is not yet in the usermanual.
I assumed that filmic help will be in a section called "filmic" in a subsection of tone_group, but this pull request should be merged only after the filmic help is added, to be sure of this assumption.


rawfiner:
I think it is better to merge it only after documentation has come. Otherwise, users will have some "link not found" errors on their web browsers

rawfiner:
I removed the commit that adds the url for filmic. I now think it is better to add the urls for the other features that already have a documentation without waiting for the filmic one. As such, even if filmic documentation arrives after 2.6, we will still have the contextual help links for retouch, duplicate, etc for 2.6.
I will make a second pull request for filmic url when filmic documentation will be available.

PR #1862, opened 2018-11-30
exiv2: compatibility with 0.27.0.1

edgomez:
Seems like newer exiv2 versions dropped APIs used in current DT code

Use the equivalent ones that did not get dropped


TurboGit:
In Debian experimentation we still have 0.27 so maybe this can wait for the 2.7 dev release and possibly one of the corrective 2.6.x releases?

edgomez:
Sure I'm not in a hurry. I just happened to try an up to date version to see if one of my new lenses was supported. Then i just found out and fixed this small issue.

 As you're preparing 2.6, i fully understand this can be put on hold. I'm ok

TurboGit:
Ok, it makes sense to integrate this in 2.6 then. I'll check and merge if possible.

TurboGit:
Let's this be in RC1 then. Thanks.

PR #1863, opened 2018-11-30
build: update list of features

edgomez:
Get the list of build.sh features in sync with CMakeLists.txt option()s


TurboGit:
Sounds like a good idea indeed. Thanks.

PR #1864, opened 2018-11-30
colorchecker: add click event to make the color picked match the target color

edgomez:
When using the colorchecker module on a photo where a color chart is visible, it is now possible to  a mod1+click on a patch to let the colorchecker module compute the correction to make the picked color match the target reference color a/b values.

This lets the user match the color target reference by hand picking the color patches on his/her photo manually.

PS: It doesn't try to match the L target reference value as one   can adjust L* using exposure for the whole image.

### DIFF
--- a/src/iop/colorchecker.c
+++ b/src/iop/colorchecker.c
@@ -1159,7 +1159,8 @@ static gboolean checker_motion_notify(GtkWidget *widget, GdkEventMotion *event,
         "click to select\n"
         "double click to reset\n"
         "right click to delete patch\n"
-        "shift-click while color picking to replace patch"),
+        "shift-click while color picking to replace patch\n"
+        "alt-click while color picking to match the color with the patch reference"),
       p->source_L[patch], p->source_a[patch], p->source_b[patch]);
   gtk_widget_set_tooltip_text(g->area, tooltip);
   return TRUE;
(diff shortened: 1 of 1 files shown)


src/iop/colorchecker.c
```diff
@@ -1210,6 +1211,16 @@ static gboolean checker_button_press(GtkWidget *widget, GdkEventButton *event,
     self->gui_update(self);
     return TRUE;
   }
+  else if((event->button == 1) &&
+          ((event->state & GDK_MOD1_MASK) == GDK_MOD1_MASK) &&
+          (self->request_color_pick == DT_REQUEST_COLORPICK_MODULE))
+  {
+    // Make the picked color match target color on its chroma components
+    // NB: we don't match the L channel because the user is still left with
+    //     the option to increase the exposure by himself/herself
```
TurboGit: I'm wondering if this is the right option. Maybe we should also match the L channel to have the correct color and let user adjust if necessary? It seems easier this way than letting the user try to adjust the L channel to match the color.

edgomez: My take on this is that the L* values are only going to match the reference values in very specific cases during the shot and there are probably more risks at trying to adjust L* values differently depending on their chromaticity.

For example, if the target is not lit homogeneously (close speedlight w/ rapid falloff, or target hold w/ an angle to light source), matching L* would mess up L* values for the different patches depending on their physical position in the scene.

Hence my comment about letting users adjust L* globally on the image.

But I have no strong opinion on this. What's yours ?

aurelienpierre: > For example, if the target is not lit homogeneously

If so, the whole colorchecker profiling becomes useless. I think you have to assume in the code that the lightingt has been done properly. That's user's responsibility.

LebedevRI:
I have twice now typed a comment for this pr, and discarded it.

TLDR: i don't think i get to comment on the feature itself, but i would like to very strongly highlight
that this has the potential to be the third case (previous were masks, multi-instances)
that was merged [potentially too soon] in a horrible/disgusting state (thus discouraging anyone
and everyone from even touching it, fixing bugs in it), as usual with no tests whatsoever,
and the echo in the form of bugs will be there for a *very* long time.
Given the nature of this change, i suspect this may end up being the most destructive of all three...

TurboGit:
Any news on this? Thanks.

TurboGit:
No feedback. Closing for now.

PR #1865, opened 2018-12-01
retouch: reorder tools to be consistent with other module.

TurboGit:
the brush is always first in mask, the select tool has been moved last.
also the fill mode is now last, probably less used than blur.

### DIFF
--- a/src/iop/retouch.c
+++ b/src/iop/retouch.c
@@ -2588,13 +2588,14 @@ void gui_init(dt_iop_module_t *self)
                  "shapes are added to the current scale"),
                (char *)NULL);
 
-  g->bt_brush
-      = dtgtk_togglebutton_new(dtgtk_cairo_paint_masks_brush, CPF_STYLE_FLAT | CPF_DO_NOT_USE_BORDER, NULL);
-  g_signal_connect(G_OBJECT(g->bt_brush), "button-press-event", G_CALLBACK(rt_add_shape_callback), self);
-  g_object_set(G_OBJECT(g->bt_brush), "tooltip-text", _("add brush"), (char *)NULL);
-  gtk_toggle_button_set_active(GTK_TOGGLE_BUTTON(g->bt_brush), FALSE);
-  gtk_widget_set_size_request(GTK_WIDGET(g->bt_brush), bs, bs);
-  gtk_box_pack_end(GTK_BOX(hbox_shapes), g->bt_brush, FALSE, FALSE, 0);
+  g->bt_edit_masks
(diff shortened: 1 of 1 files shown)


TurboGit:
@edgardoh ok with this?

upegelow:
From a logical point of view: shouldn't the retouch tools come below the wavelet decompose? In the normal way of working you first consider how to separate an image and then you start to work with tools on specific scales. This also would bring the shape modifiers closer to the shape producers.

TurboGit:
Agreed. Makes senses and will certainly help usability. Will do that right away.

TurboGit:
Done. It indeed makes things clearer to me.

upegelow:
@edgardoh should give his POV on this.

TurboGit:
@edgardoh : the most important point is having the tools and the tools options close to each other. So another option is to move the tools options above the wavelet decompose section. Would this be ok?

For the tools I'm using the brush more often :)

edgardoh:
> @edgardoh <https://github.com/edgardoh> : the most important point is
> having the tools and the tools options close to each other. So another
> option is to move the tools options above the wavelet decompose section.
> Would this be ok?
>
By tool options you mean the shapes properties, so you want to bring those
two sections together? If that's the case, I place it last because the only
algorithms that have a property are the blur and fill with color, and they
are not used very often, so again the order of the sections reflect how
often they are used.

> For the tools I'm using the brush more often :)
>
Well, probably there's no way to make every one happy ;)

I've been using this tool for a long time time now, from very basic sensor
dust clean to heavy retouching and I'm comfortable with the GUI, but that
may very well be because I'm use to it, there's no way for me to tell other
than try something different.

From the use-case perspective, with the current functionality, there's no
much interaction with the GUI, I use the continuous add and focus on the
image, and when I use the wd I focus on the wd bar, so it doesn't really
matters where it is placed. Same goes for the order of the shapes and
algorithm buttons. So if you have a strong conviction that this changes
will make things better, go for it.

I still have the feeling that the order of the sections is OK the way it,
but again, what I say above.

TurboGit:
> If that's the case, I place it last because the only algorithms that have a property are the blur and fill with color

No, there is also always the opacity. And for me I'm using the blur a lot as this is a good way to remove artefacts from the skin on different scales.

edgardoh:
Just one more note, the properties section show/hide some fields based on the selected shape, so if a section is below it, it will be 'moved' each time a shape is selected. If you want the shapes and properties section together, the best option probably is what @upegelow suggested and make the wd section first.

TurboGit:
Ok, so let's go for this. It is actually what I have done in this PR.

PR #1869, opened 2018-12-02
guided filter: add an additional parameter to adjust the effect of th…

rabauke:
…e guiding image for better feathering results in rgb mode

This pull request is to improve feathering results in rgb mode and to fix the issue https://redmine.darktable.org/issues/12429 . The result of the guided filter depends on the relative amplitudes of the input image (the mask) and the guiding image. In LAB mode pixel values are of the order of 100, in RGB mode, however, pixel values are in [0, 1]. Thus, an additional parameter was introduced to compensate this. Now, mask feathering should work in RGB mode as good as in LAB mode.

### DIFF
--- a/src/common/guided_filter.c
+++ b/src/common/guided_filter.c
@@ -40,31 +40,27 @@ typedef struct tile
   int left, right, lower, upper;
 } tile;
 
-typedef float rgb_lab_pixel[3];
-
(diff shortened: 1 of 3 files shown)


TurboGit:
Working fine for me! Thanks.

TurboGit:
Thanks will be part of RC1!

PR #1870, opened 2018-12-03
exif: canon subject distance is in cm.

TurboGit:
This fixes https://redmine.darktable.org/issues/12432

But I'm not sure it is ok for all Canon camera. Better this to be double checked by someone else I'm using Nikon.

### DIFF
--- a/src/common/exif.cc
+++ b/src/common/exif.cc
@@ -648,15 +648,23 @@ static bool dt_exif_read_exif_data(dt_image_t *img, Exiv2::ExifData &exifData)
     }
     else if(Exiv2::testVersion(0,25,0) && FIND_EXIF_TAG("Exif.CanonFi.FocusDistanceUpper"))
     {
-      float FocusDistanceUpper = pos->toFloat();
-      if(FocusDistanceUpper <= 0.0f || FocusDistanceUpper >= 0xffff)
+      const float FocusDistanceUpper = pos->toFloat();
+      if(FocusDistanceUpper <= 0.0f || (int)FocusDistanceUpper >= 0xffff)
       {
         img->exif_focus_distance = 0.0f;
       }
-      else if(FIND_EXIF_TAG("Exif.CanonFi.FocusDistanceLower"))
+      else
       {
-        float FocusDistanceLower = pos->toFloat();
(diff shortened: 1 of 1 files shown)


src/common/exif.cc
```diff
@@ -659,6 +659,10 @@ static bool dt_exif_read_exif_data(dt_image_t *img, Exiv2::ExifData &exifData)
         img->exif_focus_distance = (FocusDistanceLower + FocusDistanceUpper) / 200;
       }
     }
+    else if(FIND_EXIF_TAG("Exif.CanonSi.SubjectDistance"))
```
LebedevRI: This code (and it's position (after the upper/lower)) appears to be correct.

src/common/exif.cc
```diff
@@ -659,6 +659,10 @@ static bool dt_exif_read_exif_data(dt_image_t *img, Exiv2::ExifData &exifData)
         img->exif_focus_distance = (FocusDistanceLower + FocusDistanceUpper) / 200;
```
LebedevRI: This code (that i wrote) is incorrect though.
It should be
```c++
    else if(Exiv2::testVersion(0,25,0) && FIND_EXIF_TAG("Exif.CanonFi.FocusDistanceUpper"))
    {
      float FocusDistanceUpper = pos->toFloat();
      if(FocusDistanceUpper <= 0.0f || FocusDistanceUpper >= 0xffff)
      {
        img->exif_focus_distance = 0.0f;
      }
      else
      {
        img->exif_focus_distance = FocusDistanceUpper / 100.0;
        if(FIND_EXIF_TAG("Exif.CanonFi.FocusDistanceLower"))
        {
          float FocusDistanceLower = pos->toFloat();
          if(FocusDistanceLower > 0.0f && FocusDistanceLower < 0xffff))
          {
            img->exif_focus_distance += FocusDistanceLower / 100.0;
          }
        }
      }
    }
```

TurboGit: Ok, I was indeed asking myself about the / 200 ! I'll integrate your change in this PR. Thanks for the review.

LebedevRI: Actually, wait, this is almost correct.
We still need to average, so 
```c++
img->exif_focus_distance += FocusDistanceLower / 100.0;
img->exif_focus_distance /= 2.0;
```

PR #1872, opened 2018-12-04
update German translation for darktable 2.6 release

mepi0011:
I have translated all fuzzy and untranslated messages. Now all 2501messages (100%) translated for darktable 2.6.


TurboGit:
Thanks.

PR #1873, opened 2018-12-04
Change aspect ratio of GUI curves for rawdenoise and denoiseprofile

rawfiner:
Takes less space like this, and it remains very usable with this new aspect ratio.
I used same aspect ratio as the one of color zones.


TurboGit:
Just perfect now! Thanks!

PR #1874, opened 2018-12-05
filmic: bugfixes & improvements

aurelienpierre:
1. UI:  plot the nodes on the graph so users get a better idea of what's going on with the parameters, and sanitize the nodes less aggressively so errors are more obvious and intuitive (avoids the effect "small parameters changes lead to big curve flips"
1. BUG: catch on exception where the curve interpolation method is not properly saved/imported and raises a SEGFAULT on the next picture opening

### DIFF
--- a/src/iop/filmic.c
+++ b/src/iop/filmic.c
@@ -1025,22 +1025,15 @@ void compute_curve_lut(dt_iop_filmic_params_t *p, float *table, float *table_tem
   // nodes for mapping from log encoding to desired target luminance
   // X coordinates
   float toe_log = grey_log - latitude/dynamic_range * fabsf(black_source/dynamic_range);
-
-  toe_log = CLAMP(toe_log, 0.0f, grey_log);
-
   float shoulder_log = grey_log + latitude/dynamic_range * white_source/dynamic_range;
 
-  shoulder_log = CLAMP(shoulder_log, grey_log, 1.0f);
 
   // interception
   float linear_intercept = grey_display - (contrast * grey_log);
 
   // y coordinates
   float toe_display = (toe_log * contrast + linear_intercept);
-  toe_display = CLAMP(toe_display, 0.0f, grey_display);
-
   float shoulder_display = (shoulder_log * contrast + linear_intercept);
-  shoulder_display = CLAMP(shoulder_display, grey_display, 1.0f);
 
   // Apply the highlights/shadows balance as a shift along the contrast slope
   const float norm = powf(powf(contrast, 2.0f) + 1.0f, 0.5f);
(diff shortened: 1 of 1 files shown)


upegelow:
To me adding nodes to the display goes into the wrong direction. This will increase mismatch between user expectation and function. To me the graph is a pure assist function to visualize the settings of the sliders in a supplementary way (you know my opinion). With nodes this generates even more of an expectation that the graph is meant for direct manipulation. Users will try, they will fail and they will get frustrated.

moy:
I'm not getting the weird behavior I had anymore with this PR. I'm not sure adding the control points is a good idea as @upegelow noticed (no strong opinion on that).

src/iop/filmic.c
```diff
@@ -1169,20 +1192,21 @@ void compute_curve_lut(dt_iop_filmic_params_t *p, float *table, float *table_tem
...
 
     // Compute the cubic spline interpolation
     curve = dt_draw_curve_new(0.0, 1.0, CUBIC_SPLINE);
-    for(int k = 0; k < nodes; k++) (void)dt_draw_curve_add_point(curve, x[k], y[k]);
+    for(int k = 0; k < nodes_data->nodes; k++) (void)dt_draw_curve_add_point(curve, nodes_data->x[k], nodes_data->y[k]);
     dt_draw_curve_calc_values(curve, 0.0f, 1.0f, res, NULL, table);
     dt_draw_curve_destroy(curve);
 
     // Average both LUT
+    /*
```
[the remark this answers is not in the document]
aurelienpierre: Indeed, debugging leftover

aurelienpierre:
The thing is some real users have asked to see the nodes. So I don't know if @upegelow is speaking of real users-expressed expectations or of what he thinks they might expect, but it seems we have a choice to make here. (And I would be very careful when imagining what users want, because we don't have a marketing service to collect user surveys…)

When the parameter are properly set, 5 nodes are available on the graph. When the parameters are inconsistent, the software remodes the 1 to 2 invalid nodes. So, the nodes are yet another debugging tool to understand how the parameters affect the curve.

As for the representation of the nodes, what about filled/plain circles instead of the hollow ones ? So that hollow = edit, plain = read-only.

TurboGit:
Maybe, let's go with plain circle. A bar could be barely visible on some slope of the curve.

moy:
I like the idea of showing the points with a different graphical representation. Cross marks or filled circles both look good to me.

If possible, a tooltip when hovering the curve with the mouse saying e.g. "transformation applied to the image, use the cursors below to edit" would be cool too. I don't think we need much to convince the user that this curve shouldn't be edited, but I can imagine the frustration of a user trying to edit the curve and trying to understand where the "bug" comes from.

src/iop/filmic.c
```diff
@@ -943,10 +950,36 @@ static void interpolator_callback(GtkWidget *widget, dt_iop_module_t *self)
...
   dt_iop_color_picker_reset(&g->color_picker, TRUE);
   const int combo = dt_bauhaus_combobox_get(widget);
-  if(combo == 0) p->interpolator = CUBIC_SPLINE;
-  if(combo == 1) p->interpolator = CATMULL_ROM;
-  if(combo == 2) p->interpolator = MONOTONE_HERMITE;
-  if(combo == 3) p->interpolator = 3; // Optimized
+
+  switch (combo)
+  {
+    case CUBIC_SPLINE:
```
TurboGit: Note that all this is equivalent to:

`p->interpolator = combo;
`

TurboGit: Understood.

TurboGit:
@moy : it is too late for the tooltip for 2.6.0.

TurboGit:
Thanks that's a nice clean-up.I too find the curve welcomed, I also think that such a curve is well-known by the photographers as it is the very same we have on the tone curve for example. So to me the added value is clear. We can always revisit this later.

junkyardsparkle:
I wonder if making the node dots dark (darker than the background grid) and shown as *behind* the plotted curve would make them more obviously not for "grabbing"?

aurelienpierre:
Careful here, you don't know what UI background color are in use. It's bad enough that the colors are hard-coded, overdoing the read/write vs. read--only separation could harm the readibility more than anything else. i think keeping the nodes the same color as the curve is the less bad option. We will add a toolitp for the next minor-release, meanwhile users will have to RTFM or send emails to report "un-intuitive behaviour" until they get it.

junkyardsparkle:
Good point. Could it be implemented by simply giving the dots a slightly lesser opacity than the curve?

PR #1876, opened 2018-12-05
color balance: more meaningful UI

aurelienpierre:
- display the UI controls in % for factors and saturation, and in degree ° for hues. This avoids values with 5 decimals and makes the UI clearer. Internal parameters are not changed, only the UI is affected.
- remove the hard-coded 2.2 normalization in lift/gamma/gain ProPhotoRGB. This was intended to roughly match the output between sRGB and ProPhotoRGB mode using the same parameters, but if we add more color spaces in the future, the hardcoded value will create compatibility problems and ugly workarounds. So users will need to adjust the saturation if they switch between sRGB and ProPhotoRGB.


TurboGit:
Sounds good! Thanks.

PR #1877, opened 2018-12-06
filmic: graph improvements

aurelienpierre:
- remove the grey node to keep only 4 interpolations nodes: after going back and forth about this central node, I have come to the conclusion that it's better to give it up (at the expense of some precision on the grey point remapping) rather than having to deal with the cusps it produces.
- make a (big ?) improvement on the graph so now the grey is always remapped at (50, 50) % on the graph, no matter the parameters. This is only a scaling of the coordinates on the display, it does not change the actual parameters (similar in the logic to the log scaling in tone curve module, but using a different scaling equation). Since we give up the central node, the center of the graph can now be used anytime as a reference for the grey point and the S curve is now always symmetrical. It should give a better readability.

Sorry for the noise, with scattered last-minute PR, but ideas come to me as I use the tools and I believe it can help.

### DIFF
--- a/src/iop/filmic.c
+++ b/src/iop/filmic.c
@@ -1338,15 +1338,15 @@ void init(dt_iop_module_t *module)
   dt_iop_filmic_params_t tmp
     = (dt_iop_filmic_params_t){
                                  .grey_point_source   = 18, // source grey
-                                 .black_point_source  = -10.0,  // source black
-                                 .white_point_source  = 3.0,  // source white
+                                 .black_point_source  = -8.64,  // source black
+                                 .white_point_source  = 2.45,  // source white
                                  .security_factor     = 0.0,  // security factor
                                  .grey_point_target   = 18.0, // target grey
                                  .black_point_target  = 0.0,  // target black
                                  .white_point_target  = 100.0,  // target white
                                  .output_power        = 2.2,  // target power (~ gamma)
                                  .latitude_stops      = 2.0,  // intent latitude
-                                 .contrast            = 1.618,  // intent contrast
+                                 .contrast            = 1.5,  // intent contrast
                                  .saturation          = 100.0,   // intent saturation
                                  .balance             = 0.0, // balance shadows/highlights
                                  .interpolator        = CUBIC_SPLINE, //interpolator
--- a/src/iop/filmic.c
+++ b/src/iop/filmic.c
@@ -1149,19 +1149,19 @@ void compute_curve_lut(dt_iop_filmic_params_t *p, float *table, float *table_tem
   else
   {
     // everything OK
-    nodes_data->nodes = 5;
+    nodes_data->nodes = 4;
 
     nodes_data->x[0] = black_log;
     nodes_data->x[1] = toe_log;
-    nodes_data->x[2] = grey_log,
-    nodes_data->x[3] = shoulder_log;
-    nodes_data->x[4] = white_log;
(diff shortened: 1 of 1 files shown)


TurboGit:
I have something very strange with this PR. Take as an example the image you've used at some point. When I used the auto-tuner I get a curve all against the left side of the display. See screen shot:

capture d ecran de 2018-12-06 09-53-37

And in many cases the image with the auto-tuner is full of strange color. See second screenshot:

capture d ecran de 2018-12-06 09-58-03

aurelienpierre:
that's not weird at all, your parameters are bad, so the curve breaks. With white at + 6EV, black should be at -4 EV top. In this setup, the grey should be adjusted at 4 % or so, in order to center the dynamic range around 0. 

Thinking about it, maybe I should put the max of the black range at - 4EV.

TurboGit:
But that's not *my* parameters. That's the parameters of the auto-tuner :)

aurelienpierre:
Yeah, the auto-tuner is not super reliable. You need a good demosaicing and a good denoising (if needed) to have it work properly. I could add some sanitization checks on that too, but I kind of prefer having some obvious clue that it broke instead to have is silently half-corrected stuff.

Notice that the same auto-tuner works better in profilegamma IOP, probably because it works in non-converted RGB. Filmic works in the Lab pipe, so it converts back to RGB, and that seems destructive at some point, regarding the black/noise level.

TurboGit:
Thanks for the notes. I have tested the auto-tuner on more than 20 pictures. It always works great on landscape pictures. I have some problems as described above but globally it works in more cases to me. So, that's a good move forward and a nice stabilization effort for 2.6. Thanks.

TurboGit:
Yes, in that case you need to manually set the black/middle/white points. The auto-tuner cannot guess all the time the good values. Note that it works fine on HDR images and landscapes for example. On low-key, high-key and noisy pictures it may not be the right tool. Aurélien will correct me if I'm wrong.

junkyardsparkle:
I totally understand about the "reasonable values"... the problem, from the perspective of a naive user not paying attention to the numbers, is that they tend to discover the "reasonable" range by moving sliders *a little bit beyond it*. ;)

Trying to remember/reapply the ratio does seem like it would get messy, though. Is it possible there's a more elegant way to make this case "resilient"? That is, to conditionally *not* have the condition 1 supersede the condition 2 at this point? If going into that range doesn't really produce useful results, then it's purpose is more as a "bumper zone" anyway, right?

Sorry if I'm not making sense; I haven't looked at the code for this at all, still just trying to break it from the perspective of an excited new user... I appreciate that a bullet-proof UI for this kind of thing is really difficult.

aurelienpierre:
>  Is it possible there's a more elegant way to make this case "resilient"? That is, to conditionally _not_ have the condition 1 supersede the condition 2 at this point? If going into that range doesn't really produce useful results, then its purpose is more as a "bumper zone" anyway, right?

I will have to think about that. The problem is users may edit something else than regular camera raw: X-rays, 3D renders/CGI, whatever… I don't want to lock skilled users in an UI that thinks for them and limit them, for the sake of preventing unskilled users from doing bad things. 

So, even if I have not had useful results passed that point, on raw pictures, I'm not confident enough to say it will never give good results on some pictures, and lock the feature for everyone in order to avoid to the majority the trouble to read the doc, and get some sense of what they are doing.

PR #1878, opened 2018-12-06
denoiseprofile: minor fix to increase performance

rabauke:
The result of fmaxf is always positive.


TurboGit:
Indeed posix semantics says that if x is 0 then sqrtf(x) returns 0. But is that true on all platforms?

For example on Windows it does not seems to be the case, at least with a quick search I was not able to find this explicitly documented.

rabauke:
Note that the variable d is always non-negative, see line of code before. No need to double check. The zero argument case is also well defined, even on Windows, see https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/sqrt-sqrtf-sqrtl?view=vs-2017 A sqrt function that is undefined for zero should be considered seriously broken.

TurboGit:
Sorry... sometime I need to sleep more :(

PR #1879, opened 2018-12-07
filmic: auto-tuner fix

aurelienpierre:
- fix the luminance detection in auto-tuners by using the measure in XYZ space (should be way more reliable)
- ensure black level <= - white level any time: that should avoid many problems with bad curves, when the shadows range is smaller than the highlights range. Good curves are produced only when the dynamic range is centered or slided toward the black, and linear camera RGB always put the white around + 2.45 EV, so the black should always be betwen -5.55 and -12.00 EV when the gray is at 18 %.
- resize the slider ranges to safe values, accordingly to the previout point
- prepare presets for the next minor version (commented for now since we are in string freeze)

### DIFF
--- a/src/iop/filmic.c
+++ b/src/iop/filmic.c
@@ -231,7 +231,108 @@ int legacy_params(dt_iop_module_t *self, const void *const old_params, const int
   return 1;
 }
 
+/* UNCOMMENT AFTER DARKTABLE 2.6 IS RELEASED
 
+void init_presets(dt_iop_module_so_t *self)
+{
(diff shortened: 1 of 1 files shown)


rawfiner:
I tested it today, including on complicated pictures. I just wanted to say that I am finding the module much easier (and faster!) to use!
This is awesome. Thanks!

aurelienpierre:
Nice ! Putting together the maths to get the algo working is one thing, but designing an easy interface to control it is a lot more work. It took a fair amount of iterations, but I'm glad it's finally working.

PR #1880, opened 2018-12-07
Replace special character in noiseprofiles.json

peterbud:
The MSYS2 version of `jsonschema `is currently failing validation if there are these special characters in the json file. I hope Damjan would not mind. The best would be to fix `jsonschema `though

### DIFF
--- a/data/noiseprofiles.json
+++ b/data/noiseprofiles.json
@@ -3356,7 +3356,7 @@
           ]
         },
         {
-          "comment": "olympus e-1 contributed by Damjan Vrenčur",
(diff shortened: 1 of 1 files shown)


TurboGit:
Is that a validation only issue or it makes the profiles denoise module unusable?

peterbud:
It's only a stupid validation error, because the latest MSYS2 version of `jsonschema` recently cannot cope with these special characters, and results validation error (hence stopping the build), where the `json` file is perfectly valid. Once you replace these `'ĉ'` with `'c'` validation is successful, and build succeeds.

TurboGit:
Ok, so let's merge this.

PR #1883, opened 2018-12-08
filmic: avoid "staircase function" effect when drawing the curve

moy:
The computation of x when drawing nodes and the curve itself was along
the lines of:

  powf(...) + d

where powf(...) and d were large numbers and the result small for low
middle grey values. As a result, using single precision floats were
resulting in cancellation effect where variatiations of k were not
propagated to x, resulting in points artificially grouped on at the
same abscissa on the graph, i.e. visually a staircase function, or
even all points squashed together at x=0:

stair-multiple-stairs
stair-one-stair
stair-vertical

Fix this by using double precision locally. The effect on performance
is marginal since only a few hundreds of points are computed.

This only affects the way the curve is drawn, the processing of the
image was not affected.

While we're there, fix the coding style.

### DIFF
--- a/src/iop/filmic.c
+++ b/src/iop/filmic.c
@@ -1444,8 +1444,12 @@ static gboolean dt_iop_tonecurve_draw(GtkWidget *widget, cairo_t *crf, gpointer 
 
   for(int k = 0; k < nodes_data->nodes; k++)
   {
-    const float x = powf(2.0f, a * nodes_data->x[k] + b) + d,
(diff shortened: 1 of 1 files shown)


aurelienpierre:
Notice that this should not happen anymore. The straicasing happens only with bad parameters, e.g. with black level > - white level. That case leads to poor curves, and the UI should prevent it to arise by correcting the values straight in the sliders since my last PR on filmic.

PR #1884, opened 2018-12-08
duplicate: disable deleting the first clone.

TurboGit:
We can't leave the darktable opened without a selected picture. And it
should fix the bug #12448.

### DIFF
--- a/src/libs/duplicate.c
+++ b/src/libs/duplicate.c
@@ -254,6 +254,7 @@ static void _lib_duplicate_init_callback(gpointer instance, dt_lib_module_t *sel
   dt_develop_t *dev = darktable.develop;
 
   int first_imgid = -1;
+  int count = 0;
 
   // we get a summarize of all versions of the image
   DT_DEBUG_SQLITE3_PREPARE_V2(dt_database_get(darktable.db), "SELECT i.version, i.id, m.value FROM images AS "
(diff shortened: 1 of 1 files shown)


moy:
I can't get dt to segfault anymore with this PR (i tested with the setup that caused the crash previously, i.e. my old library.db file).

Ideally, we should be able to delete any duplicate, even the first one, until there's only one duplicate left. If one has two versions of the same image, one may want to remove the first version and keep only the second. In the short term I think your PR is good (better a lack of feature than a crash ...), though.

I'm wondering whether it would be better to just remove the button instead of disabling it.

AFAICT, there is now no way to delete the image file from disk from the duplicate module, which is a good thing. But the confirmation message should be reworded from "send X selected images to trash" to "remove X selected images from the library" (cf. the difference between "trash" and "remove" buttons in lighttable). After the release because we're in string freeze obviously.

moy:
I like the idea better, but now I'm getting the segfault again :-(. It's not just "deleting a duplicate when there's only one available" but really deleting the first one which triggers it.

TurboGit:
Do you have an updated backtrace?

moy:
Hmm, obviously, Heisenbug effect, I can't reproduce under gdb anymore :-(.

TurboGit:
I have created multiple clone and deleted always the first one. No crash and all this running under gdb!

TurboGit:
Ok, so let's go with this. Hopefully it is fixed or we will get better bt next time. Thanks.

TurboGit:
Merged with a little edit.

moy:
I just attached a new backtrace. I have a gdb open on the crash if needed.

moy:
It crashes in rawprepare, so I'm assuming that dt triggers the pipeline on the just deleted image, but that's just a guess.

TurboGit:
I'll have a look.

TurboGit:
A guess, a possible race condition. Can you test #1885 ?

PR #1885, opened 2018-12-08
duplicate: rework the way duplicated are removed to avoid race condit…

TurboGit:
…ion.

Should fix the crash reported when deleting a duplicate.


TurboGit:
Good, thanks for testing, those race conditions are very hard to fix as it may or may not be reproduced. As in this case, I don't have crashes.

moy:
Thanks. I'll open a new bug if I get the crashes again in real-life, but hopefully I won't.

PR #1887, opened 2018-12-09
Allow middle-click to be used to open the presets.

TurboGit:
This allow middle-click to open and middle-click to select the
preset to apply on the new instance, so avoiding the left-click to middle
click finger move.

For #12453.

### DIFF
--- a/src/develop/imageop.c
+++ b/src/develop/imageop.c
@@ -1660,20 +1660,25 @@ static void _preset_popup_position(GtkMenu *menu, gint *x, gint *y, gboolean *pu
 }
 #endif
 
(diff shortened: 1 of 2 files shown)


moy:
Looks good to me. Note that it doesn't apply to lightroom modules (like "export selected"). It doesn't matter much because these modules don't have the "multiple instances" thing like darkroom, but since the button has the same shape it may make sense to have exactly the same behavior hence allow middle-click.

TurboGit:
Good point! I'll add that, consistency is a very strong argument to me.

TurboGit:
manually merged.

PR #1888, opened 2018-12-09
Lp fulcrum

phweyland:
The lowpass module amplifies (contrast) the difference between image & blurred image around the image middle grey but doesn’t give the possibility to chose the middle point of this amplification.
Here is a try intoducing in lowpass module a fulcrum contrast slider (by default at 50, Lab L) letting the user set where he wants to make the amplification happen.


aurelienpierre:
You did not update the OpenCL version of the processing. Do you need help on that part ?

phweyland:
In fact I haven't seen any OpenCL piece of code involved.
The fulcrum part is integrated in the ctable calculation (10000 points), that's all. I've understood, but I may be wrong, that's ctable which is the entry for OpenCL.

aurelienpierre:
You are right, my bad.

src/iop/lowpass.c
```diff
@@ -142,55 +155,74 @@ int groups()
...
+    return 0;
+  }
   return 1;
 }
 
 void init_key_accels(dt_iop_module_so_t *self)
 {
   dt_accel_register_slider_iop(self, FALSE, NC_("accel", "radius"));
   dt_accel_register_slider_iop(self, FALSE, NC_("accel", "contrast"));
+  dt_accel_register_slider_iop(self, FALSE, NC_("accel", "cfulcrum"));
```
TurboGit: The preset name should be "contrast fulcrum" or something more readable.

TurboGit:
I'd like also this to be rebased on master to avoid all the merges in the history. Can you please do on your branch:

```
$ git checkout lp-fulcrum
$ git fetch origin
$ git rebase origin/master
$ git push -f phweyland lp-fulcrum
```

Thanks.

phweyland:
I'm sorry. I've done everything except I'm blocked on:
```
> $ git push -f phweyland lp-fulcrum
> fatal: 'phweyland' does not appear to be a git repository
> fatal: Could not read from remote repositor
```
I've tried to understand the password issue but I'm afraid I've only added a bit of mess.

TurboGit:
Ok,not sure what is happening. I'll clean-up before merging.

src/iop/lowpass.c
```diff
@@ -142,55 +155,74 @@ int groups()
...
 {
-  if(old_version == 1 && new_version == 4)
+  if(old_version == 1 && new_version == 5)
   {
     const dt_iop_lowpass_params1_t *old = old_params;
     dt_iop_lowpass_params_t *new = new_params;
     new->order = old->order;
     new->radius = fabs(old->radius);
     new->contrast = old->contrast;
+    new->cfulcrum = 50.0f;
```
TurboGit: default cfulcrum is 50.

src/iop/lowpass.c
```diff
@@ -533,7 +576,7 @@ void commit_params(struct dt_iop_module_t *self, dt_iop_params_t *p1, dt_dev_pix
 #endif
     for(int k = 0; k < 0x10000; k++)
     {
-      float kx2m1 = 2.0f * (float)k / 0x10000 - 1.0f;
+      float kx2m1 = 2.0f * ((float)k / 0x10000 - d->cfulcrum / 100.0f);
```
TurboGit: But then the algorithm does not keep history as here : 50/100 is not equal to 1.

src/iop/lowpass.c
```diff
@@ -519,7 +562,7 @@ void commit_params(struct dt_iop_module_t *self, dt_iop_params_t *p1, dt_dev_pix
   if(fabs(d->contrast) <= 1.0f)
   {
     // linear curve for contrast up to +/- 1
-    for(int k = 0; k < 0x10000; k++) d->ctable[k] = d->contrast * (100.0f * k / 0x10000 - 50.0f) + 50.0f;
+    for(int k = 0; k < 0x10000; k++) d->ctable[k] = d->contrast * (100.0f * k / 0x10000 - d->cfulcrum) + 50.0f;
```
TurboGit: This is ok, but below...

aurelienpierre: It should be ```d->ctable[k] = d->contrast * (100.0f * k / 0x10000 - d->cfulcrum) + d->cfulcrum;```

src/iop/lowpass.c
edgardoh: Also
float kx2m1 = 2.0f * (float)k / 0x10000 - 1.0f;
is not the same as
float kx2m1 = 2.0f * ((float)k / 0x10000 - 1.0f);

src/iop/lowpass.c
TurboGit: Right I missed that!

src/iop/lowpass.c
phweyland: > 
> 
> But then the algorithm does not keep history as here : 50/100 is not equal to 1.

You're right, but multiplied by 2 it does.

phweyland: > float kx2m1 = 2.0f * (float)k / 0x10000 - 1.0f;
> is not the same as
> float kx2m1 = 2.0f * ((float)k / 0x10000 - 1.0f);

You're right too, but
float kx2m1 = 2.0f * ((float)k / 0x10000 - 0.50f);
does.

src/iop/lowpass.c
phweyland: > default cfulcrum is 50.

Yes. Can you explicit a bit more what you mean ?

src/iop/lowpass.c
aurelienpierre: but then your function output is not properly bounded in [0; 1] : https://www.wolframalpha.com/input/?i=1.5+*+(x+-+0.25)+%2B+0.50,+x+%3D+%5B0,+1%5D

src/iop/lowpass.c
edgardoh: You're right, my bad!

src/iop/lowpass.c
aurelienpierre: > Makes the correction proportional to fulcrum, which is not what we look for.

it does not, the fulcrum is an offset. It's proportionnal to the contrast.

phweyland: ...
https://www.wolframalpha.com/input/?i=1.+*+(x+-+0.5)+%2B+0.50,+x+%3D+%5B0,+1%5D

phweyland:
It seems I've wiped a conversation, sorry ...

src/iop/lowpass.c
aurelienpierre: imagine fulcrum = 0.25, the current function as you wrote it is not only adding contrast but also luminance, because it is shifted linearly by 0.25.

phweyland:
> 
it does not, the fulcrum is an offset. It's proportionnal to the contrast. 

No. In the initial formula, the first 50 is the pivot of the amplification ( * contrast).
The second one gives the difference with the grey. It's the pivot of blending mechanism. Below grey the value is substracted from image, above it is added.
The first one is replaced by the fulcrum because we want to change the pivot of the amplification.
The second remains constant, because the blending mechanism still works the same.


----- Mail original -----
De: "Aurélien PIERRE" <notifications@github.com>
À: "darktable-org/darktable" <darktable@noreply.github.com>
Cc: "Philippe" <philippe.weyland@libertysurf.fr>, "Author" <author@noreply.github.com>
Envoyé: Dimanche 9 Décembre 2018 18:42:20
Objet: Re: [darktable-org/darktable] Lp fulcrum (#1888)



@aurelienpierre commented on this pull request. 


In src/iop/lowpass.c : > @@ -519,7 +562,7 @@ void commit_params(struct dt_iop_module_t *self, dt_iop_params_t *p1, dt_dev_pix

   if(fabs(d->contrast) <= 1.0f)

   {

     // linear curve for contrast up to +/- 1

-    for(int k = 0; k < 0x10000; k++) d->ctable[k] = d->contrast * (100.0f * k / 0x10000 - 50.0f) + 50.0f;

+    for(int k = 0; k < 0x10000; k++) d->ctable[k] = d->contrast * (100.0f * k / 0x10000 - d->cfulcrum) + 50.0f; 



Makes the correction proportional to fulcrum, which is not what we look for. 

it does not, the fulcrum is an offset. It's proportionnal to the contrast.

phweyland:
> imagine fulcrum = 0.25, the current function as you wrote it is not only adding contrast but also luminance, because it is shifted linearly by 0.25.

With it initial formula, we get a deviation from grey proportionnal to Contrast and (Luminance - Grey).
With the new one, we get a deviation from grey proportionnal to Contrast and (Luminance - Fulcrum).



----- Mail original -----
De: "Aurélien PIERRE" <notifications@github.com>
À: "darktable-org/darktable" <darktable@noreply.github.com>
Cc: "Philippe" <philippe.weyland@libertysurf.fr>, "Author" <author@noreply.github.com>
Envoyé: Dimanche 9 Décembre 2018 18:53:06
Objet: Re: [darktable-org/darktable] Lp fulcrum (#1888)



@aurelienpierre commented on this pull request. 


In src/iop/lowpass.c : > @@ -519,7 +562,7 @@ void commit_params(struct dt_iop_module_t *self, dt_iop_params_t *p1, dt_dev_pix

   if(fabs(d->contrast) <= 1.0f)

   {

     // linear curve for contrast up to +/- 1

-    for(int k = 0; k < 0x10000; k++) d->ctable[k] = d->contrast * (100.0f * k / 0x10000 - 50.0f) + 50.0f;

+    for(int k = 0; k < 0x10000; k++) d->ctable[k] = d->contrast * (100.0f * k / 0x10000 - d->cfulcrum) + 50.0f; 

imagine fulcrum = 0.25, the current function as you wrote it is not only adding contrast but also luminance, because it is shifted linearly by 0.25.

aurelienpierre:
But the blendings are no concern of yours at this stage of the process. All you do here is increasing/reverting contrast with a transfer function, that happens to add also lightness in your modification. The blending is external of this IOP.

phweyland:
> The blending is external of this IOP.

You are right. As the current lowpass module works, at least for local contrast effect, a pure grey image (constrast = 0) doesn't affect the original image (overlay blending).
The more the grey image goes away from grey (contrast <> 0 and more the luminosity is far from grey), the more the effect is important.
In the current module, this difference is calculated by reference to grey (50).
The fulcrum gives the ability to change the reference to calculate the difference. With fulcrum = 25 for example, all points for which the luminosity is closed to 25 won't be affected. This gives a pivot point in the low light. On the opposite, fulcrum = 75 will not affect the high lights around 75. But in both case the contrast will boost the local difference, either in low or in high lights.
A demo would be easier ... :-)

aurelienpierre:
I'm currently testing your code. If I sample an area (using the color-picker) before the lowpass, and set its average luminance as the fulcrum, using contrast = 1 and lightness = 0, the output average luminance over this area becomes 50 % (as I planned). It should not do that (otherwise, it's not a grey fulcrum), the average luminance should not change when == fulcrum.

The proper behaviour is obtained with:

```
diff --git a/src/iop/lowpass.c b/src/iop/lowpass.c
index 2feadcf89..cc6ff9cc1 100644
--- a/src/iop/lowpass.c
+++ b/src/iop/lowpass.c
@@ -562,7 +562,7 @@ void commit_params(struct dt_iop_module_t *self, dt_iop_params_t *p1, dt_dev_pix
   if(fabs(d->contrast) <= 1.0f)
   {
     // linear curve for contrast up to +/- 1
-    for(int k = 0; k < 0x10000; k++) d->ctable[k] = d->contrast * (100.0f * k / 0x10000 - d->cfulcrum) + 50.0f;
+    for(int k = 0; k < 0x10000; k++) d->ctable[k] = d->contrast * (100.0f * k / 0x10000 - d->cfulcrum) + d->cfulcrum;
   }
   else
   {
@@ -577,7 +577,7 @@ void commit_params(struct dt_iop_module_t *self, dt_iop_params_t *p1, dt_dev_pix
     for(int k = 0; k < 0x10000; k++)
     {
       float kx2m1 = 2.0f * ((float)k / 0x10000 - d->cfulcrum / 100.0f);
-      d->ctable[k] = 50.0f * (contrastscale * kx2m1 / sqrtf(1.0f + contrastm1sq * kx2m1 * kx2m1) + 1.0f);
+      d->ctable[k] = d->cfulcrum * (contrastscale * kx2m1 / sqrtf(1.0f + contrastm1sq * kx2m1 * kx2m1) + 1.0f);
     }
   }
 ```

To deal with the blending thresholds, one should use the lightness correction.

aurelienpierre:
> If constrast = 0 (brightness and saturation = 0), the image should be full grey, right ?
> If you change the fulcrum, there is no reason to affect the image, right ?
> If, in the formula you add fulcrum instead of 50, changing fulcrum will change the image.

A fulcrum means you roll the contrast around that value, without affecting that value. So, for contrast = 0, the image should be == fulcrum everywhere.

phweyland:
> A fulcrum means you roll the contrast around that value, without affecting that value. So, for contrast = 0, the image should be == fulcrum everywhere.

No, I don't agree.  I think you don't get my point.
The "no change" is a full grey image (all points=50).
The driver of the change is the contrast. If contrast is null there is no change, whatever the fulcrum (the fulcrum is only the pivot of that contrast !). Then the image is still full grey. And that that the way I imagined it shoudl work with overlay blending (I haven't tested other modes but I don't see why this wouldn't work the same way).

> The proper behaviour is obtained with:

I don't agree, neither. I've seen the behaviour of this in my tests and that was not what I expected. Nevertheless I'm curious to see the usage we could do of it, you have probably another idea.

edgardoh:
I just tested it with overlay and softlight, it doesn't seem to have much difference than playing with the brightness. I didn't a full test, but by setting brightness = -1 I get almost the same as with fulcrum 75.
Maybe I'm missing something?

aurelienpierre:
No, that does not work. Your are not supposed to change the lightness at all with contrast = 0, no matter the fulcrum. That means, the output lightness is the fulcrum lightness anytime, not an arbitrary 50 % designed to help with a particular blending. As of now, the fulcrum was hard-set at 50%, fine, but if you want to make is dynamic, you hade to stay consistent.

From a Fourier spectrum point of view, "no change" means conservation of energy, and no contrast means you have a single frequency = 0 Hz which intensity is the average input energy. Output means output of the IOP with no blending. Null contrast means no luminance variation in the picture  (no frequency > 0 Hz), that does not trigger any assumption on the average output value in general (the intensity of the 0 Hz frequency).

Again, you are pushing the assumption that the module will always be used blended in overlay mode, where 50 % is the threshold value, so grey = 50 % means indeed the blended output is unchanged (identity). But if you blend in multiply mode, 50 % means 50 %, not darken/lighten threshold, and in this case, your contrast setting is changing the lightness too, and it's bad.

You need to forget the blending here. What this IOP does is a blur, on which several corrections are applied. What happens after the IOP is accessory. For every pixel == fulcrum, output = input for every contrast value. That's what a fulcrum is. That's not what your algo does. What you call fulcrum is a normalization shift that works only if lowpass is blended in overlay.

phweyland:
> For each pixel == fulcrum, output = input for all contrast value. That's what a fulcrum is. That's not what your algo does. What you call fulcrum is a normalization shift

I can agree on this.  

> As of now, the fulcrum was hard-set at 50%,

I'm afraid the lowpass module doesn't work as you describe neither. The hard-set 50% value cannot be a fulcrum because for contrast = 0 the current output is a flat grey image, then the sentence `For each pixel == fulcrum, output = input for all contrast value.` is not verified.

So what about changing the name ? normalization shift ? contrast slider pivot ?

aurelienpierre:
> I'm afraid the lowpass module doesn't work as you describe neither. The hard-set 50% value cannot be a fulcrum because for contrast = 0 the current output is a flat grey image, then the sentence `For each pixel == fulcrum, output = input for all contrast value.` is not verified.

yes it is. For each (input pixel == fulcrum), output pixel = fulcrum. If contrast == 0, **all** output pixels = fulcrum. If contrast != 0, **only** (input pixel == fulcrum) give output pixel = fulcrum. Thus for all (input pixel == fulcrum), input pixel = output pixel == fulcrum.

> So what about changing the name ? normalization shift ? contrast slider pivot ?

Just use the shadows/highlights module, because that shift is exactly what its "compress" slider does, and it blurs the same way with gaussian and bilateral, with separated controls of luminance and saturation for shadows and highlights, and performs a built-in overlay blending.

phweyland:
> For each (input pixel == fulcrum), output pixel = fulcrum. If contrast == 0, **all** output pixels = fulcrum. If contrast != 0, **only** (input pixel == fulcrum) give output pixel = fulcrum. Thus for all (input pixel == fulcrum), input pixel = output pixel == fulcrum.

My bad, that's right.

> use the shadows/highlights module

I'll do it tonight.

phweyland:
Still a doubt. I haven't been able to find the way to use the pure "fulcrum" slider (like with your formula).
Do you have any idea in mind on that way (from usage point of view) ?

phweyland:
> I just tested it with overlay and softlight, it doesn't seem to have much difference than playing with the brightness. I didn't a full test, but by setting brightness = -1 I get almost the same as with fulcrum 75

Not sure to understand what your trying to do.
One point which is sure, the fulcrum slider influences only the behaviour of the contrast slider. It has no effect on brightness and saturation sliders.

aurelienpierre:
> One point which is sure, the fulcrum slider influences only the behaviour of the contrast slider. It has no effect on brightness and saturation sliders.

No, your version of the fulcrum is affecting the lightness as well, that's the point I'm trying to make since yesterday. Whatever value used as a fulcrum gets remapped to 50 % lightness. Just check the maths.

> But this module provides an image of contrast, which by definition is grey for contrast = 0, whatever the intensity.

This module provides a low-pass filter. A such filter retains the energy of the signal, otherwise it's not a frequential filter anymore. I don't know if you have a background in harmonic analysis, but I find it easier to understand from a spectral point of view. Contrast means non-constant image. Constant image does not implies middle grey image (that's only a particular case).

> Therefore “contrast fulcrum” seems to be the correct name as a fulcrum is nothing else than a pivot point, here the intensity pivot point, around which rounds the contrast (delta intensity) but not the intensity itself.

I'm sorry, but I cannot agree with you. You don't only roll around the pivot here, you slide the pivot to 50 % lightness too.

> Still a doubt. I haven't been able to find the way to use the pure "fulcrum" slider (like with your formula).
> Do you have any idea in mind on that way (from usage point of view) ?

When fulcrum = 50 %, with my formula, it changes nothing, compared to the current implementation. When fulcrum < 50 % and you are using an overlay blending (so 50 % L means threshold), you need to increase the lightness to remap the fulcrum value to 50 % (exactly what your formula does without advertising it). The other way around, when fulcrun > 50 %, you need to decrease the lightness.

Again, what you do is ok in a setup where the lowpass is only blended in overlay mode, to perform a dodge and burn. The problem is your contrast fulcrum affects the lightness as well, which is supposed to be controlled separately. My pure fulcrum slider makes them completely independant (as they should), but in the other hand, needs another lightness adjustement. That's why I don't like the lowpass filter to perform dodging and burning : it is a low-level technical filter that has much nicer and more specialized interfaces in shadows/highlights or even in equalizer and local-contrast. I know Shane Milton has a love affair with this module, but I suspect it's only because he doesn't understand what equalizer does (wavelets are not trivial, but they are nothing more than stacked low-pass filters with increasing scales).

If I were you, I would just add the pure contrast fulcrum in lowpass, then add a sigmoidal contrast correction in shadows/highlights with the fulcrum defined by the compression parameter. That way, you have a consistent (yet trickier to set) low-pass for technical users, and a more powerfull shadows/highlights module to do at once what you are doing with 2 instances of low-pass.

phweyland:
> equalizer [] (wavelets are not trivial, but they are nothing more than stacked low-pass filters with increasing scales)

Yes, it works nicely in the way you can choose one or different sizes of patterns to contrast or to smooth. I've also the impression that the constrast is smoothed/amplified around the local luminosity (on the opposite of lowpass and local-constrast which have a fulcrum at 50%). But one cannot chose the lumosity where the effect applies. 
In fact it would be nice to modulate also the local contrast on different luminosity levels (kind of stacked fulcrums ...).

> Just use the shadows/highlights module, because that shift is exactly what its "compress" slider does

I don't understand this. With compress = 100%, ouput = input. With 0%, shadows and highlights effects just overlap/cross over. It seems to limit the luminosity extent of each slided. I don't see the relationship with fulcrum ...

> Or even better, use the exposure module to increase the lightness linearly, with parametric blending in multiply mode to isolate shadows/highlights, and a blurred/feathered mask to reproduce your low-pass.

Sounds interesting but I haven't succeed in getting a pleasant local contrast effect. Would you have any example available ?

PR #1889, opened 2018-12-09
retouch gui

edgardoh:
This PR reverts changes made to the retouch gui, this is the right order for sections and controls.


TurboGit:
Well we've been discussed that! At least Ulrich and I agreed for the current ordering. IIRC you had no strong opinion. So to revert we will need a poll :) We cannot satisfy everyone. I have myself used the new layout extensively recently to develop portraits and must say that I much prefer the current layout. Note that I have a strong opinion for having the brush first for consistency reasons.

So, yes a poll... I cannot go one way and another. But I can assure you that if we have many people asking for this PR to go in with good arguments, it will.

aurelienpierre:
I like the UI as it is but I don't really care, both ways are good with me.

TurboGit:
@edgardoh : I suppose this won't change at this point. I'll close this PR. If you want this to be re discussed fell free to reopen. I'm ready to change my mind if a majority if for the previous layout.

PR #1890, opened 2018-12-09
fix module reaming UI issues

rabauke:
Currently module renaming does not work in some desktop environments. No UI element to enter the new name shows up.  To fix this
```
gtk_widget_show_all(d->floating_window);
```
was replaced by
```
gtk_widget_show(d->floating_window);
gtk_widget_show(entry);
```

Other code changes are to _not_ to destroy the window as soon as it loses focus. This is necessary because  the desktop environment may be configured to give focus to the window below the mouse cursor without clicking into the window. In this case, the window with the input field would immediately disappear unless it is created under the current position of the mouse cursor (which is normally not the case). Side effect of the chance: Input dialog can only be canceled via ESC key.

### DIFF
--- a/src/develop/imageop.c
+++ b/src/develop/imageop.c
@@ -934,12 +934,6 @@ static gboolean _rename_module_key_press(GtkWidget *entry, GdkEventKey *event, d
   return FALSE; /* event not handled */
 }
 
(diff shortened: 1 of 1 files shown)


TurboGit:
> Input dialog can only be canceled via ESC key.

Or enter without entering a value. This works for me and if it fixes the issue on all DE all good. Thanks.

PR #1892, opened 2018-12-10
Add Fujifilm X-T3 white balance presets

cryptomilk:
Add Fujifilm X-T3 white balance presets


darix:
woot woot!

TurboGit:
Sounds good thanks.

PR #1893, opened 2018-12-10
filmic: relax one UI constraint

aurelienpierre:
The constraint over the black exposure (ensuring <= - white exposure), although safe in 100 % cases, leads to faded blacks in some difficult cases. Relaxing this constraint will give less security in the parameters (users will need to be more careful) but will allow to cope with more cases. (This is only on the GUI side).


moy:
Tested on a handful of images, works well for me and does fix https://redmine.darktable.org/issues/12456#change-36418.

Thanks!

TurboGit:
Looks good! Thanks.

PR #1894, opened 2018-12-11
filmic: don't rescale the graph abcissa when grey log < grey destination

aurelienpierre:
This case was avoided with the constraint over the black exposure value (< - white exposure) that has been relaxed in the last PR.
The coordinate remapping equation does not work in this setup, so we just leave the abcissa linear.


TurboGit:
Code is sounds and working fine on my side. Thanks. I really hope we are done with the auto-tuner code stabilization for a maybe final RC2.

aurelienpierre:
I hope so too. It has been quite an adventure.

PR #1895, opened 2018-12-11
fix segfault in denoise(profiled)

### LINKED ISSUE #12458: white balance presets for Fujifilm x-t200
Hi,
Can you please add the white balance presets for the Fujifilm X-t200 camera. here you can find the package that was created with your extract_wb_from_images.sh script
[darktable-whitebalance-20220910.tar.gz](https://github.com/darktable-org/darktable/files/9539925/darktable-whitebalance-20220910.tar.gz)

wpferguson:
Added check for gui_data and params to prevent crash when applying denoise(profiled) to single image when module is not visible in the gui.  Fix #12458


TurboGit:
Perfect thanks!

PR #1896, opened 2018-12-11
filmic: bugfix

aurelienpierre:
auto black color-picker did not update parameters after constraint relaxation


TurboGit:
Working fine thanks.

PR #1897, opened 2018-12-11
lighttable: colors customizable through CSS

moy:
Most colors used to display thumbnails were hardcoded values. For a
white background theme, the colors were rather dark. With this patch,
one can get a light theme for the lighttable with eg.

```
@define-color thumbnail_bg_color #eee;
@define-color thumbnail_selected_bg_color shade(@thumbnail_bg_color, .9);
@define-color thumbnail_hover_bg_color shade(@thumbnail_bg_color, .8);
@define-color thumbnail_outline_color #7f7f7f;
@define-color thumbnail_selected_outline_color shade(@thumbnail_outline_color, .8);
@define-color thumbnail_hover_outline_color shade(@thumbnail_outline_color, .6);
@define-color thumbnail_font_color #aaa;
@define-color thumbnail_selected_font_color shade(@thumbnail_font_color, .8);
@define-color thumbnail_hover_font_color shade(@thumbnail_font_color, .6);
@define-color thumbnail_border_color #222;
@define-color thumbnail_selected_border_color #eee;
```

Screenshot:
light


src/gui/gtk.c
```diff
@@ -1184,6 +1184,41 @@ int dt_gui_gtk_init(dt_gui_gtk_t *gui)
   c[DT_GUI_COLOR_BRUSH_TRACE] = (GdkRGBA){ 0., 0., 0., 0.8 };
   gtk_style_context_lookup_color(ctx, "brush_trace", &c[DT_GUI_COLOR_BRUSH_TRACE]);
 
+  c[DT_GUI_COLOR_THUMBNAIL_BG] = (GdkRGBA){ 0.4, 0.4, 0.4, 1.0 };
```
TurboGit: I don't think this is correct, it may or may not work as gtk_style_context_lookup_color() does not say that the color is not changed if not found. In dt we currently do:
```
if(!gtk_style_context_lookup_color(ctx, "thumbnail_bg_color", &c[DT_GUI_COLOR_THUMBNAIL_BG]))
   c[DT_GUI_COLOR_THUMBNAIL_BG] = (GdkRGBA){ 0.4, 0.4, 0.4, 1.0 };
```

moy: In practice it does work at least for me, but why not be more strict. We'll need to update the similar code right above too (I'm guilty for it too). Actually, I think it's time to introduce a helper function to make this a one-liner. Will do later.

moy: Done.

TurboGit:
Sounds good now! I plan to merge in 2.6.1, ok for you?

moy:
> I plan to merge in 2.6.1, ok for you?

Yes. It should be harmless change, but better resist the temptation to merge it before 2.6, just in case.

moy:
I've pushed a few more commits & rebased. All of them can wait for 2.6.1. The last 3 commits are not directly CSS-related, let me know if you prefer a separate PR.

TurboGit:
conflict resolved and manually merged.

src/gui/gtk.c
```diff
@@ -1166,23 +1166,39 @@ int dt_gui_gtk_init(dt_gui_gtk_t *gui)
...
+    [DT_GUI_COLOR_THUMBNAIL_HOVER_OUTLINE] = { "thumbnail_hover_outline_color", { 0.6, 0.6, 0.6, 1.0 } },
+    [DT_GUI_COLOR_THUMBNAIL_FONT] = { "thumbnail_font_color", { 0.425, 0.425, 0.425, 1.0 } },
+    [DT_GUI_COLOR_THUMBNAIL_SELECTED_FONT] = { "thumbnail_selected_font_color", { 0.5, 0.5, 0.5, 1.0 } },
+    [DT_GUI_COLOR_THUMBNAIL_HOVER_FONT] = { "thumbnail_hover_font_color", { 0.7, 0.7, 0.7, 1.0 } },
+    [DT_GUI_COLOR_THUMBNAIL_BORDER] = { "thumbnail_border_color", { 0.1, 0.1, 0.1, 1.0 } },
+    [DT_GUI_COLOR_THUMBNAIL_SELECTED_BORDER] = { "thumbnail_selected_border_color", { 0.9, 0.9, 0.9, 1.0 } },
+    [DT_GUI_COLOR_FILMSTRIP_BG] = { "filmstrip_bg_color", { 0.2, 0.2, 0.2, 1.0 } },
+  };
+
+  for(int i = 0; i < DT_GUI_COLOR_LAST; i++)
```
TurboGit: Just spotted this mistake. The index i start to 0, first item name is "darkroom_bg_color" but it is set into the c[0] which is actually DT_GUI_COLOR_BG. And since the loop is going up to DT_GUI_COLOR_LAST-1 then we read last element in the struct which is not set.

Can you have a look?

TurboGit: Fixed in 46b1d457e4d5d9

moy:
Thanks for the fix. Looks good to me.

PR #1899, opened 2018-12-12
Fix crash denoise

TurboGit:


wpferguson:
@TurboGit , tested and it works.  And, it's a much more elegant solution than mine.  I thought about other modules being effected, but my understanding of the GUI code is rather limited at this point.

TurboGit:
Thanks for testing. I'll commit that for 2.6.0rc3 as potentially fixing crashes in other modules.

PR #1901, opened 2018-12-14
retouch: fix computation of roi_in.

TurboGit:
for non clone we want the area and not the source area.

Fixes #12463.


aurelienpierre:
Works fine with me, soooooo much quicker… Many thanks @TurboGit !

TurboGit:
Thanks! Merging then. Just ready for rc2!

PR #1902, opened 2018-12-14
equalizer: fix bug #12461 and fix radius in tiling_callback

rabauke:
The equalizer module causes segmentation faults in the SSE2 and the CPU code paths when the maximal radius of the wavelet decomposition is larger than the height or the width of the roi. For a _valid_ decomposition 

(2 * mult) + (2 * mult) 

must be smaller than the height/width. See the three main loops in `eaw_decompose` and `eaw_decompose_sse2`.

### DIFF
--- a/src/iop/atrous.c
+++ b/src/iop/atrous.c
@@ -31,6 +31,7 @@
 #include "iop/iop_api.h"
 #include "common/iop_group.h"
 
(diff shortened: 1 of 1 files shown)


TurboGit:
Sounds good to me! Thanks.

PR #1903, opened 2018-12-14
colorbalance : update teal/orange preset for blendop v8 and mask feat…

aurelienpierre:
…hering

Preset now optimized to be used on linearly encoded RGB (with filmic, instead of log profile previously) and to take advantage of the feathering feature


TurboGit:
Perfect! Safe and looks a bit more pronounced effect now with presets.

PR #1906, opened 2018-12-16
remove unreachable code and if-clauses which evaluate always to true/false

rabauke:
I applied some static code analysis tools to the darktable sourcs and found, among others, unreachable code and if-clauses which evaluate always to true/false.

### DIFF
--- a/src/common/imageio_rawspeed.cc
+++ b/src/common/imageio_rawspeed.cc
@@ -254,7 +254,6 @@ dt_imageio_retval_t dt_imageio_open_rawspeed(dt_image_t *img, const char *filena
         break;
       default:
         return DT_IMAGEIO_FILE_CORRUPTED;
-        break;
     }
 
     // dimensions of uncropped image
@@ -399,7 +398,7 @@ dt_imageio_retval_t dt_imageio_open_rawspeed_sraw(dt_image_t *img, RawImage r, d
       }
     }
   }
-  else if(cpp == 3 || cpp == 4)
+  else // case cpp == 3 or 4
   {
 /*
  * standard 3-ch image
--- a/src/iop/clipping.c
+++ b/src/iop/clipping.c
@@ -1182,17 +1182,11 @@ void commit_params(struct dt_iop_module_t *self, dt_iop_params_t *p1, dt_dev_pix
     d->all_off = 0;
     d->crop_auto = 0;
   }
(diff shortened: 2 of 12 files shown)


src/common/imageio_rawspeed.cc
```diff
@@ -399,7 +398,7 @@ dt_imageio_retval_t dt_imageio_open_rawspeed_sraw(dt_image_t *img, RawImage r, d
       }
     }
   }
-  else if(cpp == 3 || cpp == 4)
+  else
```
TurboGit: ok, but I would keep the comment for the readers:

   else // case cpp == 3 or 4

src/libs/live_view.c
```diff
@@ -524,13 +524,7 @@ void gui_post_expose(dt_lib_module_t *self, cairo_t *cr, int32_t width, int32_t
...
     if(!img) img = dt_image_cache_get(darktable.image_cache, imgid, 'r');
 
-    int zoom = 1;
-    float imgwd = 0.90f;
-    if(zoom == 1)
-    {
-      imgwd = .97f;
-    }
-
+    float imgwd = 0.97f;
```
TurboGit: Indeed, looks like a code after refactoring :)

TurboGit: And as we are at it, make imgwd a const to make it clear that it does not change.

rabauke:
Requested changes applied, plus some further minor fixes.

src/iop/clipping.c
```diff
@@ -1187,12 +1187,6 @@ void commit_params(struct dt_iop_module_t *self, dt_iop_params_t *p1, dt_dev_pix
     d->all_off = 1;
     d->k_apply = 0;
   }
-  else
```
TurboGit: Are you sure about this one?

We have:

```
if(p->k_type == 4)
else if(p->k_type >= 0 && p->k_apply == 1)
else if(p->k_type == 0)
else
```

So the else part can be taken if p->k_type > 0 and p->k_apply == 0.

upegelow: Out of own experience I'd be very cautious with any changes to that module shortly before a release. IMHO it's better to allow for intensive testing in the development branch.

TurboGit:
Merged manually. Better sooner than later, but this new version seems correct to me. Thanks.

PR #1909, opened 2018-12-17
Fix mask source

TurboGit:
This would also have been fixing the retouch slowness recently fixed. Seems important to properly handle this generic case too.


TurboGit:
manually merged.

PR #1910, opened 2018-12-17
Update pt_BR translation

jpellegrini:
This is an update of the pt_BR translation by Frido Claudino and Jeronimo Cordoni Pellegrini, based on previous work by Guilherme Brondani Torri.


TurboGit:
Thanks!

PR #1916, opened 2018-12-18
filmic: bugfix (UI)

aurelienpierre:
Fix the black exposure correction in UI when the grey slider is modified, so that the mid-tones are affected accordingly.

Before, moving the grey did:

image

So it was a no-op.

Now, it does:

image
image

So the bounds are preserved, but the mid-tones are raised, which is expected from the grey value.


TurboGit:
Do we want this in 2.6?

TurboGit:
This sounds very safe and UI only, so I'd say yes.

PR #1917, opened 2018-12-18
color contrast: BUGFIX

aurelienpierre:
Fix a bug (https://redmine.darktable.org/issues/12368) that arised when using OpenMP SIMD codepath (OpenCL disabled): the input L and alpha channels were not copied in the output of the IOP.

Also, optimize the SSE path by moving the constants definitions out of the loop.

Also, enable the GTK accels because they are available now.


TurboGit:
Sounds like we also want this in 2.6.

TurboGit:
manually merged.

PR #1918, opened 2018-12-19
colorbalance: BUGFIX

aurelienpierre:
In pure C version, the output saturation was corrected while CDL was applied. Now, CDL is applied, then the output luminance is computed, and finally the saturation is corrected.

Also, some perf optimizations : 
* perform checks out of inner loops 
* combine (i, j) loops to (k) loops improve auto-vectorization
* replace `_mm_store_ps` with `_mm_stream_ps` for better cache management

Now, the auto-vectorized OpenMP SIMD codepath runs roughly at the same speed as the pure SSE2.


TurboGit:
manually merged.

PR #1920, opened 2018-12-19
Correct function signature for focus-in-event handler

### LINKED ISSUE #1915: windows: cannot enter text in "collect images"
In the entry of the collect image module (and under WIndows only) we cannot enter text. It is possible to clear the selection in the entry with backspace, but no character can be entered for whatever mode is selected (tag, filename, film roll, folders...)

peterbud:
Fixes #1915 
Interestingly it caused problem only in release builds, debug version were working.

### DIFF
--- a/src/libs/collect.c
+++ b/src/libs/collect.c
@@ -1601,7 +1601,7 @@ int position()
   return 400;
 }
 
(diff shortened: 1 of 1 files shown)


TurboGit:
Perfect! Thanks.

PR #1923, opened 2018-12-20
Update German usermanual for darktable 2.6

mepi0011:
All new chapters and changes for darktable 2.6 are translated. It would be nice if the German user manual will be official released as PDF and on the darktable homepage together with darktable 2.6!
Do you need additional support for the official release?


upegelow:
Thanks for the work. I think it's ready to be merged. Please note that the English version of the manual still lacks documentation on filmic and on the recent changes of unbreak input profile. The 2.6 manual will be released after these pending parts have been added and the translators will need to follow accordingly.

PR #1929, opened 2018-12-23
[WIP] add support for target clones compilation

aurelienpierre:
* add SIMD() OpenMP pragma in every loop for auto-vectorization when possible
* add target clones for SSE3, SSE4.1, AVX and AVX2 so generic builds for distro packaging embed optimized variants for each architecture, and not only older SSE2.

This needs to be tested and benchmarked. So far, I get ×2 to ×3 performance gain in OpenMP SIMD codepath, wich is on-par with manually optimized SSE2 codepath. Also, tell me if you think more clones should be added.

Ideally, it would be great to have a way to auto-tune the SIMD buffer size too, but that's too low-level for my understanding (@darix @cryptomilk help !).

### DIFF
--- a/src/common/gaussian.c
+++ b/src/common/gaussian.c
@@ -154,7 +154,7 @@ error:
   return NULL;
 }
 
-
+__attribute__((target_clones( "avx2", "avx", "sse4.2", "sse3", "popcnt", "default")))
 void dt_gaussian_blur(dt_gaussian_t *g, const float *const in, float *const out)
 {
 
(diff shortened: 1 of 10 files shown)


TurboGit:
Note that this is supported by GCC 6.0 and above and current dt still supports GCC 5.0.

darix:
the __attribute__() thing should be in a define and only be enabled on x86_64 ... it will break ppc64le and aarch64 builds otherwise.

aurelienpierre:
> the **attribute**() thing should be in a define and only be enabled on x86_64

How so ? (I'm definitely not a low-level guy)

aurelienpierre:
@darix could you show me a complete example ?

aurelienpierre:
I have acquired new knowledge on that so I will close that and do one PR/module.

PR #1930, opened 2018-12-23
Allow upscaling on export by setting one one size limit zero

wocket:
This is my first pull request for darktable. So, I hope I did it right.  It is just a minor change/fix regarding the export with up-scaling enabled when only one size limit is given and the other is set to zero. Currently this always results in a scale of 1.0 and the image is exported in its native size, but not scaled.
I did not find a bug report addressing this behavior, but I assume it is not intended. I think this commit fixes it.


TurboGit:
Manually merged. Thanks.

PR #1938, opened 2018-12-27
[WIP] atrous: optimize loops

aurelienpierre:
This is an attempt to optimize the atrous module (equalizer) SSE2 version, since I need the code for another module.

* decompose the 2D convolution (25 iterations / pixel) into 2 1D convolutions (5 × 2 iterations / pixels) 
* collapse the parallel loops on pixels
* unroll the inner convolution loop
* avoid mixed castings with `size_t` and `int` types (use only `size_t`)
* vectorize the convolution filter instead of the pixels to avoid SSE intrinsics shuffling.

… And yet I find that the overall performance increases of 0-15 %, not always reproductible, so maybe I'm missing something.

Also, I'm not entirely sure this algo is equivalent (e.g. fully separable) because of the `weight_sse` exponential function. I would like a feedback from @hanatos if possible.


aurelienpierre:
Closing. The 2D filter is not separable

PR #1940, opened 2018-12-27
piwigo: build only if the required CURL version is found.

TurboGit:


upegelow:
Yepp, works well.

TurboGit:
Ok, looking at the doc it seems that this was added in 7.56.0. So indeed quite restrictive change. Let's adjust this now.

PR #1941, opened 2018-12-27
filmic: uncomment presets + add graph tooltip

aurelienpierre:
that was added after string freeze


TurboGit:
Manually merged in master and darktable-2.6.x branch. Thanks.

PR #1942, opened 2018-12-28
color balance : minor improvements

aurelienpierre:
* display the contrast in signed % (more explicit)
* be more conservative on the bound constraints of the auto-neutralizer (less divergence in the optimization)
* use `native_power` in OpenCL (less accurate but 2× faster)


TurboGit:
Manually merged. Thanks.

TurboGit:
Maybe we want this in 2.6.1? Sounds quite safe.

aurelienpierre:
> Maybe we want this in 2.6.1? Sounds quite safe.

Yes, I think so.

PR #1943, opened 2018-12-28
Filmic : new parameter + speed-up

aurelienpierre:
Optimizations:

* optimizes the loops for faster execution (SSE2 and C)
* use `native_log2` in OpenCL (faster/less accurate)
* use only SSE2 intrinsics in the SSE2 version (no vectors unpacking/repacking)
* avoid false-sharing of the parameters in the parallel loops

New feature:

* add a global saturation parameter, applied linearly at the beginning of the module (it's exactly what color balance saturation would do, but centralizing all settings in one module). NB: setting the global saturation at 100 % is fully equivalent to the original version. 

Better labels:
* relabel the current saturation "extreme luminance saturation"
* relabel the destination parameter with correct names

### DIFF
--- a/src/iop/filmic.c
+++ b/src/iop/filmic.c
@@ -286,7 +286,6 @@ void init_presets(dt_iop_module_so_t *self)
 
   // Fine-tune settings, no use here
   p.interpolator = CUBIC_SPLINE;
-  p.balance = 0.0f;
 
   // Output - standard display, gamma 2.2
   p.output_power = 2.2f;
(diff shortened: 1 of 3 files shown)


src/iop/filmic.c
```diff
@@ -253,8 +301,9 @@ void init_presets(dt_iop_module_so_t *self)
   // Presets indoors
   p.grey_point_source = 18.0f;
   p.preserve_color = 0;
-  p.latitude_stops = 2.0f;
+  p.latitude_stops = 2.0f;git
```
TurboGit: I suppose "git" should have been put in the console :)

aurelienpierre: Obviously ^^

TurboGit:
Sound good to me now. Thanks. Manually merged.

aurelienpierre:
Would it be possible to have it in 2.6.1 ? Or the policy is to have only bug fixes in minor releases ?

TurboGit:
No, not only bug fixes for 2.6.1 but bug fixes and ultra-safe enhancements. What's your take?

aurelienpierre:
This looks safe to me. I'll shake it and try to break it in the next days, then I will confirm that.

PR #1944, opened 2018-12-28
basecurve: BUGFIX

aurelienpierre:
fix an old bug where the camera base curve was applied no matter the global option `plugins/darkroom/basecurve/auto_apply_percamera_presets` setting.

Basically, the boolean `auto_apply` was handled by a pointer, but what was tested was if the pointer was defined, not its actual value.

### DIFF
--- a/src/iop/basecurve.c
+++ b/src/iop/basecurve.c
@@ -329,7 +329,7 @@ static void set_presets(dt_iop_module_so_t *self, const basecurve_preset_t *pres
     dt_gui_presets_update_ldr(_(presets[k].name), self->op, self->version(), FOR_RAW);
     // make it auto-apply for matching images:
     dt_gui_presets_update_autoapply(_(presets[k].name), self->op, self->version(),
-                                    force_autoapply ? 1 : presets[k].autoapply);
+                                    (force_autoapply) 1 : (presets[k].autoapply && force_autoapply));
     // hide all non-matching presets in case the model string is set.
     // When force_autoapply was given always filter (as these are per-camera presets)
     dt_gui_presets_update_filter(_(presets[k].name), self->op, self->version(),


src/iop/basecurve.c
```diff
@@ -340,14 +340,14 @@ static void set_presets(dt_iop_module_so_t *self, const basecurve_preset_t *pres
 void init_presets(dt_iop_module_so_t *self)
 {
   // sql begin
-  DT_DEBUG_SQLITE3_EXEC(dt_database_get(darktable.db), "BEGIN", NULL, NULL, NULL);
+  //DT_DEBUG_SQLITE3_EXEC(dt_database_get(darktable.db), "BEGIN", NULL, NULL, NULL);
```
TurboGit: you've removed the transaction to ensure atomic commit.

src/iop/basecurve.c
```diff
@@ -340,14 +340,14 @@ static void set_presets(dt_iop_module_so_t *self, const basecurve_preset_t *pres
...
 
-  set_presets(self, basecurve_presets, basecurve_presets_cnt, NULL);
   int force_autoapply = dt_conf_get_bool("plugins/darkroom/basecurve/auto_apply_percamera_presets");
-  set_presets(self, basecurve_camera_presets, basecurve_camera_presets_cnt, &force_autoapply);
+  set_presets(self, basecurve_presets, basecurve_presets_cnt, force_autoapply);
+  set_presets(self, basecurve_camera_presets, basecurve_camera_presets_cnt, force_autoapply);
 
   // sql commit
-  DT_DEBUG_SQLITE3_EXEC(dt_database_get(darktable.db), "COMMIT", NULL, NULL, NULL);
+  //DT_DEBUG_SQLITE3_EXEC(dt_database_get(darktable.db), "COMMIT", NULL, NULL, NULL);
```
TurboGit: end of transaction was here.

src/iop/basecurve.c
```diff
@@ -329,7 +329,7 @@ static void set_presets(dt_iop_module_so_t *self, const basecurve_preset_t *pres
     dt_gui_presets_update_ldr(_(presets[k].name), self->op, self->version(), FOR_RAW);
     // make it auto-apply for matching images:
     dt_gui_presets_update_autoapply(_(presets[k].name), self->op, self->version(),
-                                    force_autoapply ? *force_autoapply : presets[k].autoapply);
+                                    force_autoapply && presets[k].autoapply);
```
TurboGit: you've changed the semantic. to me the pointer was correct and the line above we to be read:

- if force_autoapply is not null
   => for apply if force_autoapply is 1 (*force_autoapply)
   => otherwise apply if preset.autoapply is set.

Of course we can do the same with a standard int or even better a gboolean and this is probably better than having this pointer.

src/iop/basecurve.c
```diff
@@ -186,30 +186,30 @@ static const basecurve_preset_t basecurve_camera_presets[] = {
   // clang-format off
 
   // nikon d750 by Edouard Gomez
-  {"Nikon D750", "NIKON CORPORATION", "NIKON D750", 0, FLT_MAX, {{{{0.000000, 0.000000}, {0.018124, 0.026126}, {0.143357, 0.370145}, {0.330116, 0.730507}, {0.457952, 0.853462}, {0.734950, 0.965061}, {0.904758, 0.985699}, {1.000000, 1.000000}}}, {8}, {m}}, 0, 1},
```
TurboGit: all those changes are not correct I think. The autoaplly should be 0 (deactivated) for the  built-in basecurve. The way to autoapply them is to create let the user create a preset based on it and then set the autoapply with possibly conditions from within dt.

src/iop/basecurve.c
aurelienpierre: I don't fully understand what this does, but other modules don't use that or have it commented, so I thought it was merely a debug thing. It doesn't seem to cause issues.

src/iop/basecurve.c
aurelienpierre: Well the output was always TRUE because the preset was applied no matter what, so the global preference was actually useless. That's what I know. What happens in the piping is still a bit unclear to me, but the pointer seemed unnecessary so this was the faster way to fix the bug I found.

TurboGit: Sure the pointer is not needed. We can use a gboolean (better than int here) as I said. Maybe we can start with that (code clean-up) and try to understand the other issue after?

src/iop/basecurve.c
TurboGit: The transaction ensure that every changes in the DB  between the BEGIN / COMMIT (or ROLLBACK) is either fully committed or nothing is changed.

src/iop/basecurve.c
aurelienpierre: The logic I used was this:
1. We have generic brand curves
1. We have specific cameras curves.

Both can be auto-applied on user request, the specific curves superceding the generic brand ones. So, if the user requested it and the preset has an auto-apply flag, we auto-apply it, wether it's a specific on a generic preset. If the user requested to not do it, we discard the auto-apply no matter the preset. The thing is the specific presets did not have the auto-apply flag except for the D750. So, in my mind, that just makes the generic and specific presets consistent with each other.

aurelienpierre:
‎If you have time, yes. Thanks !                                                                                                                                                                                                                 De: Pascal ObryEnvoyé: samedi 29 décembre 2018 12:24À: darktable-org/darktableRépondre à: darktable-org/darktableCc: Aurélien PIERRE; AuthorObjet: Re: [darktable-org/darktable] basecurve: BUGFIX (#1944)@TurboGit commented on this pull request.



In src/iop/basecurve.c:
> @@ -329,7 +329,7 @@ static void set_presets(dt_iop_module_so_t *self, const basecurve_preset_t *pres
     dt_gui_presets_update_ldr(_(presets[k].name), self->op, self->version(), FOR_RAW);
     // make it auto-apply for matching images:
     dt_gui_presets_update_autoapply(_(presets[k].name), self->op, self->version(),
-                                    force_autoapply ? *force_autoapply : presets[k].autoapply);
+                                    force_autoapply && presets[k].autoapply);

Do you want me to do the clean-up?

{"api_version":"1.0","publisher":{"api_key":"05dde50f1d1a384dd78767c55493e4bb","name":"GitHub"},"entity":{"external_key":"github/darktable-org/darktable","title":"darktable-org/darktable","subtitle":"GitHub repository","main_image_url":"https://github.githubassets.com/images/email/message_cards/header.png","avatar_image_url":"https://github.githubassets.com/images/email/message_cards/avatar.png","action":{"name":"Open in GitHub","url":"https://github.com/darktable-org/darktable"}},"updates":{"snippets":[{"icon":"PERSON","message":"@TurboGit commented on #1944"}],"action":{"name":"View Pull Request","url":"https://github.com/darktable-org/darktable/pull/1944#discussion_r244504176"}}}
[
{
"@type": "EmailMessage",
"potentialAction": {
"@type": "ViewAction",
"target": "https://github.com/darktable-org/darktable/pull/1944#discussion_r244504176",
"url": "https://github.com/darktable-org/darktable/pull/1944#discussion_r244504176",
"name": "View Pull Request"
},
"publisher": {
"@type": "Organization",
"name": "GitHub",
"url": "https://github.com"
}
}
]

TurboGit:
@aurelienpierre : see my fix just pushed. I have removed the autoapply from D7200 and setting the pref auto_apply_percamera_presets to true I have been able to have the D7200 base curve applied automatically. So to me all is fine now. Can you double check? If ok I'll merge this in darktable-2.6.x. Thanks.

aurelienpierre:
It seems that your change is not fixing the problem… The base curve is still applied no matter the setting.

TurboGit:
Confirmed, I've just reset the autoapply in data.db for some cameras and now the setting is honored.

aurelienpierre:
Yeah but do we expect users to reset their database ? Could it be possible to force the db regeneration for these presets ?

TurboGit:
Hum... My bad, it is already reset but you need to exit and restart darktable and init_presets() is called once when starting darktable and just changing the pref won't help.

TurboGit:
So to me current committed version is ok.

src/iop/basecurve.c
```diff
@@ -329,7 +329,7 @@ static void set_presets(dt_iop_module_so_t *self, const basecurve_preset_t *pres
     dt_gui_presets_update_ldr(_(presets[k].name), self->op, self->version(), FOR_RAW);
     // make it auto-apply for matching images:
     dt_gui_presets_update_autoapply(_(presets[k].name), self->op, self->version(),
-                                    force_autoapply ? 1 : presets[k].autoapply);
+                                    (force_autoapply) 1 : (presets[k].autoapply && force_autoapply));
```
TurboGit: this can't be correct, your second force_autoapply is always 0 so this is equivalent to:

   force_autoapply ? 1 : 0

aurelienpierre: You are right and I need to wake-up. The last commit in master fixes it. Well done ! :+1:

PR #1946, opened 2018-12-30
fix hierarchical tags with different upper/lower cases

oexler:
Consider you have the following tags:

- a|a
- A|b
- a|c

In the collect module instead of

- A -> b 
- a -> a, c

you get

- a -> a
- A -> b
- a -> c

as the hierarchy. This happens because for the sorting the case independent name is used but afterwards the grouping is case dependent. So there are 3 parents a, A and a, because the case change triggers a new group. 
This PR removes the case independent sorting for tags.
Perhaps a similar problem exists for folders, but this was not tested.


TurboGit:
Issue reproduced and this fix works. Thanks.

PR #1947, opened 2018-12-30
blend: increase feathering radius

schenlap:
Current radius seems too small and increasing does not harm.

### DIFF
--- a/src/develop/blend_gui.c
+++ b/src/develop/blend_gui.c
@@ -2107,7 +2107,7 @@ void dt_iop_gui_init_blending(GtkWidget *iopw, dt_iop_module_t *module)
                      G_CALLBACK(_blendop_masks_feathering_guide_callback), bd);
 
 
-    bd->feathering_radius_slider = dt_bauhaus_slider_new_with_range(module, 0.0, 250.0, 0.1, 0.0, 1);
+    bd->feathering_radius_slider = dt_bauhaus_slider_new_with_range(module, 0.0, 1000.0, 0.1, 0.0, 1);
     dt_bauhaus_widget_set_label(bd->feathering_radius_slider, _("blend"), _("feathering radius"));
     dt_bauhaus_slider_set_format(bd->feathering_radius_slider, "%.1f");
     gtk_widget_set_tooltip_text(bd->feathering_radius_slider, _("spatial radius of feathering"));


src/develop/blend_gui.c
```diff
@@ -2107,7 +2107,7 @@ void dt_iop_gui_init_blending(GtkWidget *iopw, dt_iop_module_t *module)
                      G_CALLBACK(_blendop_masks_feathering_guide_callback), bd);
 
 
-    bd->feathering_radius_slider = dt_bauhaus_slider_new_with_range(module, 0.0, 250.0, 0.1, 0.0, 1);
+    bd->feathering_radius_slider = dt_bauhaus_slider_new_with_range(module, 0.0, 1000.0, 0.1, 0.0, 1);
```
TurboGit: we are talking about pixels here. 250 seems quite large already, maybe add a soft bounds? See for example exposure, the slider goes to 3 but can be expanded to 18 EV.

aurelienpierre: I wonder if the value shouldn't be expressed in % of the diagonal size for example. It seems to me that 250 can be quite small with modern crazy high-res. I increased the slider range for shadows-highlights and lowpass to 500 for the same reason.

schenlap: I thought about which value makes sense. 250 is too small, it takes too much time to draw the mask. The feathering works really great so there is no need to draw accurate masks. For my camera (24MPx)500px is ok, i decided to go up to 1000 for hi res cameras.

with soft bound:
-> longest side = 6048px -> 500 / 6048 * 100% = 8.3% -> a soft bound of 10% of the longest side seems pretty good for me. I don't thin calculating the diagonal is needed.

rabauke: A soft bound is a reasonable option for the feathering radius. Note that all arguments for a feathering radius soft bound also apply to the gaussian blur radius. For consistency both parameters should be treated in the same way.

In order not to break old edits, I would suggest the following bound:

max(250, round_to_multiples_of_50 (0.1 * image diagonal))

and similarly for the Gaussian blur radius:

max(100, round_to_multiples_of_50 (0.1 * image diagonal))

rabauke: Hitch with the soft bound approach: A particular value for a freathering/blur radius may become out-of-bound by resizing (shrinking) the image in an editing step that comes later in the history stack. Possibly the image diagonal of the imported image (the size after demosaic) should be taken as a reference.

rabauke: @TurboGit : Thanks for the clarification about the soft bound sliders. Please ignore my last comment as it is based on a misunderstanding.

TurboGit:
Any news on this? Do you agree with the soft bound approach?

schenlap:
I am not happy with the soft bound as it is not visible in the gui and the user does not know on which parameter a soft bound is implemented. It is ok for me to stay on 250px, when you think it is enough. For the few time I use it, I can draw a finer mask.

TurboGit:
Too bad, it could have been interesting for others!

PR #1948, opened 2018-12-30
channel mixer: add N&B film presets

aurelienpierre:
Add RGB coefficients for B&W film emulation for Ilford and Kodak most used emulsions.


TurboGit:
Sounds good. Thanks.

PR #1949, opened 2018-12-30
module list: add comprehensive toolboxes

aurelienpierre:
add more module lists presets ("toolboxes") for display in darkroom:
    * all-purpose
    * portrait & beauty
    * landscape & HDR
    * lowlight & high ISO
    * architecture & streets.

These toolboxes are intended to provide minimal comprehensive lists of modules to achieve different sort of editings, based on my experience. They discard duplicate modules (for example sharpen ~= highpass, shadow-highlights ~= lowpass) and modules having funky side-effects (halos, messed-up  saturation while you change the luminance, etc.). 

Also, fix a possible memory leak because the pointer `char *param_default` was never freed.

### DIFF
--- a/src/libs/modulelist.c
+++ b/src/libs/modulelist.c
@@ -362,11 +362,11 @@ void init_presets(dt_lib_module_t *self)
                                     "|demosaic|exposure|flip|graduatednd|grain|lens|levels|monochrome"
                                     "|shadhi|sharpen|temperature|tonecurve|vignette|");
 
-  dt_lib_presets_add(_("show none"), self->plugin_name, self->version(), params_none, len);
-  dt_lib_presets_add(_("show all"), self->plugin_name, self->version(), params_all, len);
+  dt_lib_presets_add(_("subset: no module"), self->plugin_name, self->version(), params_none, len);
+  dt_lib_presets_add(_("subset: all modules"), self->plugin_name, self->version(), params_all, len);
 
   /* the modules that are activated by default in the initial configuration */
-  dt_lib_presets_add(_("default"), self->plugin_name, self->version(), params, len);
+  dt_lib_presets_add(_("subset: default modules"), self->plugin_name, self->version(), params, len);
 
   params = gen_params(1, &len,
                         "|demosaic|exposure|colorin|temperature|clipping"           // raw handling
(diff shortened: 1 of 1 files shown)


src/libs/modulelist.c
```diff
@@ -366,10 +366,79 @@ void init_presets(dt_lib_module_t *self)
...
+  dt_lib_presets_add(_("default"), self->plugin_name, self->version(), params, len);
+
+  params = gen_params(1, &len,
+                        "|demosaic|exposure|colorin|temperature|clipping"           // raw handling
+                        "|filmic|tonecurve|levels"                                  // tones
+                        "|channelmixer|colorbalance|colorzones|colorchecker|vibrance" // colors
+                        "|sharpen|bilat|atrous"                                     // sharpness
+                        "|cacorrect|defringe|highlights|lens|colorreconstruction"   // image reconstruction
+                        );
+  dt_lib_presets_add(_("0. all-purpose toolbox"), self->plugin_name, self->version(), params, len);
```
TurboGit: What are the numbers meaning here? "0." above then "1."... "I.", "II." ?

Is that needed? Sounds like noise to me.

src/libs/modulelist.c
```diff
@@ -366,10 +366,79 @@ void init_presets(dt_lib_module_t *self)
...
+
+  params = gen_params(1, &len,
+                        "|demosaic|exposure|colorin|temperature|clipping"           // raw handling
+                        "|filmic|tonecurve|levels"                                  // tones
+                        "|channelmixer|colorbalance|colorzones|colorchecker|vibrance" // colors
+                        "|sharpen|bilat|atrous"                                     // sharpness
+                        "|cacorrect|defringe|highlights|lens|colorreconstruction"   // image reconstruction
+                        "|graduatednd|zonesystem|tonemap|shadhi"                    // HDR reconstruction - tones
+                        "|hazeremoval|channelmixer"                                 // HDR reconstruction - colors
+                        );
```
TurboGit: All ");" should be on the previous line ending the code.

src/libs/modulelist.c
aurelienpierre: The arabic numbers are for toolboxes, which are aimed at providing a complete workspace. The roman numbers are for modules subsets, which are aimed at exposing quickely some features (purely technical and purely creative modules), but are not full workspaces.

TurboGit: I see, let me try this then.

TurboGit: Sorry don't like the number because they do not fit well with others preset like all or none activated.

What about:

toolbox : all purpose
toolbox : portrait & beauty
...

subset: creative only
subset: technical only

and the current presets:

subset: all
subset: none
subset: default

We have now two groups, sorted and named.

aurelienpierre: that's another option indeed

TurboGit: So if ok with you, let's go this way. Thanks.

TurboGit:
Manually merged.

PR #1951, opened 2018-12-30
Fix title des sort

TurboGit:


TurboGit:
Will merge this one in master which is quite delicate. Should go in 2.6.1 but we need more field testing.

PR #1954, opened 2018-12-31
sharpen: add preference to auto-apply or not the sharpen module.

TurboGit:
The idea is that most recent cameras do not have a low pass filter
and it is better to not add any sharpen for those.

### DIFF
--- a/data/darktable.css.in
+++ b/data/darktable.css.in
@@ -702,3 +702,14 @@ cell:selected
 {
     color: #FFFF00;
 }
+
+#pref_section
+{
+    border-bottom: 2px solid @fg_color;
+}
+#pref_section *
+{
+    font-size: 1.3em;
+    margin-left: 10em;
+    margin-right: 10em;
+}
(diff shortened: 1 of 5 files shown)


upegelow:
I understand the reason behind. However, shouldn't it be better controlled by a preset where it can be made dependent on camera model? That would better especially for users with different camera models.

TurboGit:
You mean to disable it by default for all camera without low-pass filter? I don't see how this could be done, this list is probably long and won't fit our filtering circuitry. And we do not have the list of future camera without a low-pass filter, so we will need to adjust this all the time.

Or do you have something better in mind?

upegelow:
I mean regardless of the question if the module is on or off by default users should be able to define presets that can activate the module dependent on camera model. Let's take my situation as an example. I shoot with Canon and Fuji. The my Canon the current sharpen default is normally a good starting point. For my Fuji with xtrans sensor that level of sharpening is way too much. However, I still want apply a low level of sharpen, so I defined a preset in the sharpen module which gets automatically applied for Fuji raws.

TurboGit:
@upegelow : I see, this case is still covered. the pref is for the default but you can always activate the sharpen module based on preset with a corresponding filter. So, what is changed here is a global pref, then user can activate or disable based on camera, model, iso and the like.

TurboGit:
Exactly! With the slight advantage that if disabled this way sharpen won't even appear in the history

PR #1957, opened 2019-01-01
equalizer: new deblurring presets

aurelienpierre:
These presets use the equalizer module to define a band-pass filter using a gaussian window in high-frequencies and increase local-contrast while applying a subtle denoising to avoid noise amplification.

This is intended to undo the lens blur and the demoisaicing/interpolation blur that occurs as a side-effect, to replace the basic unsharp mask for better results.

9 presets are provided : 3 blur sizes × 3 intensities of deblurring. The low intensities are better used with full-frame DSLR + prime lenses, the high ones are better used with cropped sensors and entry-level zooms. The low blur sizes are better used for close subjects and portraits, the high blur sizes are better used with far subjects and atmospheric diffusion.

### DIFF
--- a/src/iop/atrous.c
+++ b/src/iop/atrous.c
@@ -1146,7 +1146,7 @@ void init_presets(dt_iop_module_so_t *self)
     p.x[atrous_Lt][k] = p.x[atrous_ct][k] = x;
     p.y[atrous_Lt][k] = p.y[atrous_ct][k] = noise;
   }
-  dt_gui_presets_add_generic(_("deblur: large ++++"), self->op, self->version(), &p, sizeof(p), 1);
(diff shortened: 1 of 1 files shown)


TurboGit:
I must say that I'm not fond of the "+", "+++"... not easy to find a better alternative.

Let's try... we have from 1 to 4 plus, what about : delicate, light, strong, extreme?

What do you think?

TurboGit:
Ok, in that case, to avoid the alpha sorting, juste "force 1", "force 2"...

Or maybe that's just me who don't like the "+++" :) Would be nice to have more feedback.

aurelienpierre:
So I went for "deblur: fine|medium|large blur, strength 1|2|3|4"

TurboGit:
Perfect! And strength is far better in English than force :)

PR #1960, opened 2019-01-01
Update darktableconfig.xml.in - typos for "shapen" and similar

milankni:
While translating current master I noticed few typos in the config options for sharpening.


TurboGit:
Good catch! Thanks.

PR #1962, opened 2019-01-02
filmic: BUGFIX

aurelienpierre:
The SSE2 function getting the horizontal RGB max over the pixel failed and the output was inconsistent with the OpenCL version.

This is probably because the 4th channel is alpha and its value is random, so it could be identified as the max RGB.

We fall back to a pure C querying of the maximum, which is more reliable for now and almost as fast.

Also, take this opportunity to improve the tooltips text and make it more specific.

### DIFF
--- a/src/iop/filmic.c
+++ b/src/iop/filmic.c
@@ -528,16 +528,6 @@ void process(dt_iop_module_t *self, dt_dev_pixelpipe_iop_t *piece, const void *c
 
 
 #if defined(__SSE__)
-
-/* Find the max in a SSE vector */
-static inline __m128 _mm_hmax_ps(__m128 x) {
-    const __m128 max1 = _mm_shuffle_ps(x, x, _MM_SHUFFLE(0,0,3,2));
-    const __m128 max2 = _mm_max_ps(x, max1);
-    const __m128 max3 = _mm_shuffle_ps(max2, max2, _MM_SHUFFLE(0,0,0,1));
-    const __m128 max4 = _mm_max_ps(max2, max3);
-    return max4;
-}
-
 void process_sse2(struct dt_iop_module_t *self, dt_dev_pixelpipe_iop_t *piece, const void *const ivoid,
              void *const ovoid, const dt_iop_roi_t *const roi_in, const dt_iop_roi_t *const roi_out)
 {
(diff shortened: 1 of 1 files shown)


src/iop/filmic.c
```diff
@@ -1777,7 +1775,8 @@ void gui_init(dt_iop_module_t *self)
   dt_bauhaus_slider_enable_soft_boundaries(g->global_saturation, 0.0, 1000.0);
   dt_bauhaus_slider_set_format(g->global_saturation, "%.2f %%");
   gtk_box_pack_start(GTK_BOX(self->widget), g->global_saturation, TRUE, TRUE, 0);
-  gtk_widget_set_tooltip_text(g->global_saturation, _("desaturates the output"));
+  gtk_widget_set_tooltip_text(g->global_saturation, _("desaturates the input of the module globally.\n"
+                                                      "you need to set this value below 100 \% \nif the chrominance preservation is enabled."));
```
TurboGit: should be %%

src/iop/filmic.c
```diff
@@ -1693,7 +1683,7 @@ void gui_init(dt_iop_module_t *self)
   gtk_box_pack_start(GTK_BOX(self->widget), g->grey_point_source, TRUE, TRUE, 0);
   dt_bauhaus_slider_set_format(g->grey_point_source, "%.2f %%");
   gtk_widget_set_tooltip_text(g->grey_point_source, _("adjust to match the average luminance of the subject.\n"
-                                                      "except in back-lighting situations, this should be around 18%."));
+                                                      "except in back-lighting situations, this should be around 18 \%."));
```
TurboGit: should be 18%%

src/iop/filmic.c
aurelienpierre: done

src/iop/filmic.c
aurelienpierre: done

src/iop/filmic.c
```diff
@@ -1693,7 +1683,7 @@ void gui_init(dt_iop_module_t *self)
   gtk_box_pack_start(GTK_BOX(self->widget), g->grey_point_source, TRUE, TRUE, 0);
   dt_bauhaus_slider_set_format(g->grey_point_source, "%.2f %%");
   gtk_widget_set_tooltip_text(g->grey_point_source, _("adjust to match the average luminance of the subject.\n"
-                                                      "except in back-lighting situations, this should be around 18%."));
+                                                      "except in back-lighting situations, this should be around 18 %%."));
```
TurboGit: Sorry my bad, in tooltip there is no need to have double or quoted %. I'm about to commit a fix with this patchset. I have also removed spaces around \n.

TurboGit:
Manually merged.

Issue #1956, opened 2018-12-31
color balance: slope, offset, power: optimize luma finds right values but initally uses wrong ones


trougnouf:
When using color balance: "slope, offset, power", optimize luma will produce very light results that appear completely off.

Scrolling back and forth on "factor" (which restores the values shown initially) produces the desired results.

Using Arch Linux, version 2.7.0+95~gf85f182e6 and files from an X-T1 (X-Trans)

aurelienpierre:
Hmm I should write the doc about that. The thing is the luma optimization tries to remap the max RGB to 100 % luminance, the min RGB to 0 % and the average RGB to 50 %, over the selected area. So, if you select a dark area to do the auto-optimization, what happens is normal.

But, I have found that there is a ×100 error in the UI when the auto-optimizer is run (the value displayed is a ratio between [0; 1] but it should be a %). So that explain why, when you scroll after the optmization, the values are actually remapped closer to 0. I must correct that.

trougnouf:
That does fix the scrolling behavior, thank you :)
The image still ends up very light, and I don't seem to be able to select a particular area when I click on optimize luma, is that normal behavior and I should wait for the documentation and close this or it's a potential bug?
![screenshot from 2019-01-02 13-34-41](https://user-images.githubusercontent.com/6899116/50592096-ef0afe80-0e92-11e9-82c4-82502d8b3254.png)
![screenshot from 2019-01-02 13-34-51](https://user-images.githubusercontent.com/6899116/50592064-cb47b880-0e92-11e9-82f9-49a8da826c98.png)

aurelienpierre:
> is that normal behavior and I should wait for the documentation and close this or it's a potential bug?

It really depends which area you sample for the auto-tuning. Also, I see you are using it with filmic, so what you see is a double up : both filmic and color balance are raising the mid-tones. Just stick with filmic to fix the lightness. Your first image is quite good, add a bit of contrast in colorbalance and you are good to go.

If you want to ensure the auto-tuner is working properly, disable filmic, spot an area with the color-picker, trigger the optimization, and ensure, with the global color picker that the average luminance over the zone is 45 - 50 % (it should be 50 %, but the measure is done in the display RGB, so it measures what you see after the output RGB conversion, and not the actual data).

trougnouf:
[...]

PR #1963, opened 2019-01-02
colorbalance: BUGFIX

aurelienpierre:
Fix values inconsistencies in factors, as described in #1956 . Now the values are displayed in % at the output of the auto-tuners and optimizer, as they should.


TurboGit:
Manually merged.

PR #1965, opened 2019-01-02
Updated Czech translation

milankni:
To reflect recent changes in English strings.

BTW, it would be usefull if there is a low-traffic communication channel for the translators - I missed the recent release of darktable, unfortunatelly.


TurboGit:
Hi, the announce for the translation was made in darktable-dev, I could send a mail to all translators if needed.

PR #1973, opened 2019-01-05
lighttable: make lighttable faster on High DPI screen.

TurboGit:
Maybe mouse_moved events are now filtered out (some measurement shows that
only 10% of the mouse_move are actually handled).

This is WIP, to be tested on HDPI screen (4k & 5k). This breaks the zoomable lighttable rating for
now.

### DIFF
--- a/src/libs/tools/filmstrip.c
+++ b/src/libs/tools/filmstrip.c
@@ -69,6 +69,13 @@ typedef struct dt_lib_filmstrip_t
   dt_lib_filmstrip_select_t select;
   int32_t select_id;
 
+  float thumb_size;
+  int offset_x;
+  int last_mouse_over_thumb;
+  int32_t last_exposed_id;
+  gboolean force_expose_all;
+  cairo_surface_t *surface;
+
   dt_gui_hist_dialog_t dg;
 } dt_lib_filmstrip_t;
 
@@ -288,6 +295,12 @@ void gui_init(dt_lib_module_t *self)
   d->mouse_over_id = -1;
   d->pointerx = -1;
   d->pointery = -1;
+  d->thumb_size = -1;
+  d->last_mouse_over_thumb = -1;
+  d->last_exposed_id = -1;
+  d->force_expose_all = FALSE;
+  d->offset_x = 0;
+  d->surface = NULL;
   dt_gui_hist_dialog_init(&d->dg);
 
   /* creating drawing area */
@@ -386,7 +399,7 @@ static gboolean _lib_filmstrip_mouse_leave_callback(GtkWidget *w, GdkEventCrossi
 
   /* suppress mouse over highlight upon leave */
   strip->pointery = -1;
-  gtk_widget_queue_draw(self->widget);
+  gtk_widget_queue_draw(strip->filmstrip);
 
   return TRUE;
 }
@@ -423,6 +436,8 @@ static gboolean _lib_filmstrip_size_handle_button_callback(GtkWidget *w, GdkEven
 
       gtk_widget_get_size_request(d->filmstrip, NULL, &d->size_handle_height);
       d->size_handle_is_dragging = TRUE;
+      cairo_surface_destroy(d->surface);
+      d->surface = NULL;
     }
     else if(e->type == GDK_BUTTON_RELEASE)
       d->size_handle_is_dragging = FALSE;
@@ -457,6 +472,8 @@ static gboolean _lib_filmstrip_size_handle_motion_notify_callback(GtkWidget *w, 
 
     dt_conf_set_int("plugins/lighttable/filmstrip/height", sy);
 
+    cairo_surface_destroy(d->surface);
+    d->surface = NULL;
     gtk_widget_set_size_request(d->filmstrip, -1, sy);
 
     return TRUE;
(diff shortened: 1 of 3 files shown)


cryptomilk:
It is not perfect yet, but definitely an improvement!

TurboGit:
A bit vague. Was it on a 4k screen? Was it quite unusable before and do you feel that with this patch it makes the lightable bearable or even usable? Somehow as fast as on a full HD screen?

TurboGit:
Here report that there is no change:
https://redmine.darktable.org/issues/10764#note-18

That's my second attempt, looks like it fails again :)

Maybe there is some other events that take CPU cycle... I don't own a 4k display, so this will probably have to wait.

TurboGit:
@cryptomilk : can you test the new version I just pushed? Thanks.

cryptomilk:
* Yes, I have a 4K display
* Yes, this is a small improvement but not as snappy as it could be.
* No, it is not perfect yet.

cryptomilk:
2b24086bdcae7b9a988f081ea7825862ee22ea75 seems to work!

TurboGit:
Ah ah... Is that fast now? How usable?

cryptomilk:
Moving the mouse is fast, the selection follows instantly However scrolling is broken with 2b24086bdcae7b9a988f081ea7825862ee22ea75. Only two thumbnails are changing. If I scroll more it is refreshed after some time ...

TurboGit:
Right, this is fixed in the new version, I'll push now.

TurboGit:
I've just pushed a new version that should fix most issues and handle also the zoomable lighttable.

cryptomilk:
Now scrolling works again but the borders for the thumbnails are gone :-)

cryptomilk:
Scrolling works, it is super fast but the borders of the thumbnails are gone, we are close :-)

TurboGit:
What do you mean by borders of the thumbs? Can you send a screenshot? Thanks.

cryptomilk:
https://xor.cryptomilk.org/darktable/darktable.png

aurelienpierre:
Sounds like a CSS rule gone.

aurelienpierre:
Wow ! Just tested, indeed the borders are gone in the lighttable (I won't miss them), but that's some blazing speed-up !

TurboGit:
Are the borders there during first display and disappear after? Or are they never there? I cannot reproduce that. But this will get fixed and should not degrade the perfs.

cryptomilk:
They are gone on startup.

cryptomilk:
29caefec7d8d83ce424cfaf3fc35104f90e8a6b9 doesn't fix it ... :-)

TurboGit:
Ok, don't really know what is happening :( I'll sleep on that.

aurelienpierre:
Also, when we delete pictures, there is a sort of lag before the thumbnails are updated and reordered. Rating them with stars is still a bit random (needs to be done twice), and laggy too (there is some sort of delay between the click and the actual stars update).

nilsholle:
screenshot_20190106_104427
This happens on zooming in/out using Ctrl+/Ctrl- (should be reproducible on standard 1080p displays).

TurboGit:
@nilsholle : this should be fixed.

TurboGit:
@aurelienpierre : yes the lag is there because when reordering we need to redraw everything. So I do expect some lag for some actions and have this at least "usable" in 4k or 5k display.

TurboGit:
@cryptomilk : about the borders, do you see the yellow ones when you have a group selected?

cryptomilk:
@TurboGit Yes, if I hover a group I get the yellow group selection border.

cryptomilk:
As soon as I use a custom darktable.css it stops working. It looks like the darkroom_bg_color gets applied to the lighttable.

cryptomilk:
Ah ok, I see I need to change the lightable background color to be different form my darkroom backround color.

cryptomilk:
We are good to go. Just looking at the patch it looks like a few things should be size_t like max_cols and max_rows and then some rows and cols variables dealing with it. But such cleanup could be done later ...

TurboGit:
As I write in Redmine:

> For 2.6.1, maybe will see but then this needs a quite extensive testing phase. Even if the patch is not that big, it is quite delicate and in a part that we cannot break (think about all filtering settings, the stars, local copies...). So please do as much testing as possible (and any dev around). It will be easier when merged to master, and report issues preferably in GitHub. Thanks.

PaoloAst:
I'd like to test this branch, but I don't know how to do it via git commands.

$> git fetch
$> git checkout <branch name>

and 

./build.sh

Am I missing something?

TurboGit:
$ git fetch origin
$ git checkout faster-lighttable-2

and yes,

./build.sh

plus the command to install the compiled version as displayed after the compilation succeeds.

PaoloAst:
Thanks, probably I forgot to call "git config". Now it works and it seems very faster than before. I'll make some test

nilsholle:
Two things go wrong when showing a single image in lighttable:
1. I think the image position is wrong. On the left, there is a small space (slightly lighter gray) that shouldn't be there.
2. When switching to darkroom and back, the top of the lighttable should be redrawn.
screenshot_20190106_190550

aurelienpierre:
Now, navigating with keyboard arrows is broken, only the active thumbnail in the lighttable is updated, not the full rows.

aurelienpierre:
>  yes the lag is there because when reordering we need to redraw everything. So I do expect some lag for some actions and have this at least "usable" in 4k or 5k display.

The lag was on HD 1080 screen. I did not have it before (even on 4K).

TurboGit:
@aurelienpierre : thanks I can reproduce the redraw issue with keyboard. will fix.

TurboGit:
I consider this last version as usable and should be working fine. I have made many tests. Please try it and report issues. Again, this is a very delicate part with many many possible interactions.

cryptomilk:
@TurboGit I will test this asap, probably tomorrow morning. I have to edit some pictures.

TurboGit:
@aurelienpierre  : I can reproduce the first two points. Not the others.

nilsholle:
Are the changes already implemented for the thumbnail view at the bottom of the darkroom? Feel like that is still a little slow.

TurboGit:
No, for now the filmstrip is still slow. This is next step.

TurboGit:
@nilsholle : the filmstrip should now be fast too.

cryptomilk:
This looks good, I didn't find any issue when testing.

TurboGit:
@cryptomilk : just checking... you've tested the last version with the filmstrip?

cryptomilk:
Yes, I did

homer3018:
Hi guys,

I'd be happy to assist and do some further testing on my newly bought iMac 5k, but as I'm new to mac I have some (possibly dumb) questions like is it ok to have the regular 2.6 DT installed along with the dev version ? If someone would be kind enough to guide me through that that'd be great. Once up and running I hope I can help further with some coding but we'll see to that at a later stage :)
Thanks and keep up the good work !

TurboGit:
@homer3018: this is now merged. I cannot help you with building on MacOS. You'll probably get helps on the mailing-list.

homer3018:
All good Pascal, thanks for assisting.

PR #1976, opened 2019-01-05
Soft boundaries for denoising sliders

rawfiner:
We don't know how noisy will be the user's picture.
Thus, enable soft boundaries to let the possibility of increasing strength and some denoising parameters much more than currently.
I set the boundaries such that the noisiest raws that I have can be denoised, as jpegs, and without camera profile.


TurboGit:
Sounds good to me. Thanks.

PR #1982, opened 2019-01-08
colorbalance: add film-emulation presets

aurelienpierre:
add presets to reproduce color film look:
- generic film
- Kodachrome
- Kodak Portra NC 
- Kodak Ektar

These presets need a neutral white balance and filmic to work as planned. They have been adjusted visually from the look and RGB response curves of the emulations and are not completely accurate. They are in beta state and need more testing from people used to work with film.

The most accurate should be the Kodachrome, then the Portra. The generic film is an intended to get very neutral whites.

About film emulations
--

The most accurate film emulations would use the color checker module and LUTs. However, good LUTs are incredibly difficult to produce, because the film masters should be neutralized and cleaned from any flaws of the scanner, and should be applied on properly linearized and calibrated digital RAWs. Moreover, completely reproducing the film look is often not desirable since we want to preserve the linearity of the RAW RGB signal for the most part of the processing, while keeping the film color feeling. Since no color science is able to predict accurately color shifts at given luminance (Lab, CIECAM02, etc.), it is not trivial to linearize the film tone-curve and extract only the color shifts to keep working on linear luminance, until the final filmic tone-curve application. Finally, LUT can produce local contrast losses, depending on their resolution, and could give different results on different cameras.

Also, there is not one film look for each emulsion, but a lot of looks depending on the processing and the storage of the chemicals. So much for reproductibility… The gamut mapping is a real concern too, since some emulations have a very large gamut (Kodachrome is very weird on that part), and clever decisions would have to be made to address that part.

To sum up, a true film emulation would be independant from the color-response of the digital camera, its dynamic range, and leave the tone-curve interpretation to the retoucher, who would choose it depending of the input dynamic range. And that set of specifications is nearly impossible, so it all sums up to experienced people matching colors on screens.

So, using general color transfer functions, while being inaccurate, is relatively universal and makes no assumption on the camera color response. These presets add just a bit of contrast, using the fulcrum algorithm, to reproduce the toe from the filmic tone-curve (the filmic modules adds a toe too, which works great for B&W films emulations, but lacks a bit of density in blacks for color films).

Examples
--
shoot minh-ly-0009-_dsc0009
shoot minh-ly-0009-_dsc0009_02
shoot minh-ly-0009-_dsc0009_03
shoot minh-ly-0009-_dsc0009_05

Picture by Nicolas Tissot:
nft_7225_01
nft_7225_04
nft_7225_05
nft_7225_06

### DIFF
--- a/src/iop/colorbalance.c
+++ b/src/iop/colorbalance.c
@@ -274,6 +274,26 @@ void init_presets(dt_iop_module_so_t *self)
   add_preset(self, _("split-toning teal-orange (1st instance)"),
              "gz02eJxjZACBBvugHXX2E3fU219f3GAP4n/TqLFvfd1oL8HZaH/2jI/9prn1cLHUtDSwGgaGCY7//tfbAwBRixpm", 3,
              "gz04eJxjZWBgYGUAgRNODFDApgwiq5wZIVyHD4E7bBnwggZ7CIYBRiBbBA8fXT1l/P5DX21i+pnA/Pfv8uw6OzzIMq9I5rgtSH//4wii1AMASbIlcw==", 8);
+
+  add_preset(self, _("generic film"),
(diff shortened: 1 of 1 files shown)


TurboGit:
Thanks! Looks good.

PR #1996, opened 2019-01-12
imageio_tiff: support for 8/16bit tiff with CIELAB/ICCLAB photometric

upegelow:
TIFF files with CIELAB or ICCLAB format get converted to sRGB on input.

### DIFF
--- a/src/common/imageio_tiff.c
+++ b/src/common/imageio_tiff.c
@@ -30,6 +30,8 @@
 #include <strings.h>
 #include <tiffio.h>
 
+#define LAB_CONVERSION_PROFILE DT_COLORSPACE_LIN_REC2020
+
 typedef struct tiff_t
 {
   TIFF *tiff;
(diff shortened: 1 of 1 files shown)


TurboGit:
Sounds good! Do you think this can/should go in 2.6.1?

parafin:
Shouldn't it be kept in LAB and just set LAB colorspace in colorin iop? Is conversion LAB->sRGB->LAB lossless?

aurelienpierre:
> > Sounds good! Do you think this can/should go in 2.6.1?
> 
> Let's wait a bit. I am interested to see if this really works well. Also the decision to convert to sRGB might be debatable with e.g. AdobeRGB being an alternative.

As a working space, linear RGB spaces like REC2020 are most secure due to the proximity with pure spectral primaries. If it's an achival space, ACES P0 is the best, Prophoto RGB works too.

TurboGit:
Ulrich, so what about 2.6?

upegelow:
> Ulrich, so what about 2.6?

It's fine to cherry-pick into 2.6 (see discussion in darktable-users).

calabi-yaui:
Hello Ulrich. This feature seems to be broken in v4, the images remain black. Any chance to get it back working? It is sorely missed.

kmilos:
> This feature seems to be broken in v4, the images remain black.

Please open an issue for this, and provide as many details as possible, including ideally a sample image.

PR #2000, opened 2019-01-13
Bugfix for opencl of nlmeans (denoiseprofile)

rawfiner:
Fixes #12400

Updated code is very similar to the one of https://github.com/darktable-org/darktable/pull/1999

This was tested with intel neo driver which is now blacklisted, it should be tested with another driver but unfortunately I do not have any other GPU than the intel one

I made a separated pull request as I experienced some strange behavior (the first 0-15 columns were sometimes oversmoothed, the number of oversmoothed columns depended on the image width), and I do not know if it was due to the driver (or something else), but I do not see it anymore. I did not see this issue with the "normal" non local means, hence the separated pull request.

Thus, this needs to be tested with a driver we are confident with, and with several images of different width (or using cropping)

### DIFF
--- a/data/kernels/denoiseprofile.cl
+++ b/data/kernels/denoiseprofile.cl
@@ -88,16 +88,23 @@ denoiseprofile_dist(read_only image2d_t in, global float* U4, const int width, c
 
   if(x >= width || y >= height) return;
 
+  int xpq = x + q.x;
+  int ypq = y + q.y;
+  // Convert out of bounds indexes to 0
   // Reminder: q.x and q.y can be negative
-  float dist = 0.0f;
-  if(x+q.x < width && y+q.y < height
-    && x+q.x >= 0 && y+q.y >= 0)
-  {
-    float4 p1 = read_imagef(in, sampleri, (int2)(x, y));
-    float4 p2 = read_imagef(in, sampleri, (int2)(x, y) + q);
-    float4 tmp = (p1 - p2)*(p1 - p2);
-    dist = tmp.x + tmp.y + tmp.z;
-  }
+  // the boolean conditions will be equal to
+  // 0 in case of out of bounds, and will be
+  // equal to 1 in the other case
+  xpq *= (x+q.x < width && x+q.x >= 0);
+  ypq *= (y+q.y < height && y+q.y >= 0);
+
+  float4 p1 = read_imagef(in, sampleri, (int2)(x, y));
+  float4 p2 = read_imagef(in, sampleri, (int2)(xpq, ypq));
+  float4 tmp = (p1 - p2)*(p1 - p2);
+  float dist = tmp.x + tmp.y + tmp.z;
+
+  // make dist equal to 0 in case xpq or ypq is out of bounds
+  dist *= (x+q.x < width && x+q.x >= 0 && y+q.y < height && y+q.y >= 0);
 
   U4[gidx] = dist;
 }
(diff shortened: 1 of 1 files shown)


rawfiner:
The effect of this change is subtle, and only visible on borders of the image. Before the PR, a band of some pixels is left undenoised on borders, while after this PR, the denoising result is similar to the CPU version even on borders

upegelow:
See comments in PR1999

rawfiner:
@upegelow I would be happy to read and take your comments into account, but I am afraid I can't see any comment of you in PR1999, maybe something went wrong?

upegelow:
Is it visible now?

rawfiner:
Yes thanks :-)

rawfiner:
Pushed the same changes as in #1999

data/kernels/denoiseprofile.cl
```diff
@@ -86,16 +86,26 @@ denoiseprofile_dist(read_only image2d_t in, global float* U4, const int width, c
...
 
   float4 p1 = read_imagef(in, sampleri, (int2)(x, y));
-  float4 p2 = read_imagef(in, sampleri, (int2)(x, y) + q);
+  float4 p2 = read_imagef(in, sampleri, (int2)(xpq, ypq));
   float4 tmp = (p1 - p2)*(p1 - p2);
   float dist = tmp.x + tmp.y + tmp.z;
-  
+
+  // make dist equal to 0 in case xpq or ypq is out of bounds
+  dist *= (x+q.x < width && x+q.x >= 0 && y+q.y < height && y+q.y >= 0);
```
upegelow: Oops, pay attention: in OpenCL the logical TRUE value is -1, not +1 as expected from C programmers.

You will need to do something like:

dist *= (logical term) ? 1.0f : 0.0f;

data/kernels/denoiseprofile.cl
```diff
@@ -214,33 +224,40 @@ denoiseprofile_accu(read_only image2d_t in, global float4* U2, global float* U4,
...
+  // indexes (int2)(x, y) + q and (int2)(x, y) - q)
+  // respectively
+  // we want wpq and wmq equal to 1 only if
+  // their associated index is in bounds
+  int wpq = 1;
+  int wmq = 1;
+
+  // handle bounds for x
+  // Reminder: q.x can be negative
+  wpq *= (x+q.x < width);
```
upegelow: see above

data/kernels/denoiseprofile.cl
rawfiner: Oh, ok, thank you, I will change that.
Thanks for the review :-)

data/kernels/denoiseprofile.cl
```diff
@@ -86,16 +86,23 @@ denoiseprofile_dist(read_only image2d_t in, global float* U4, const int width, c
...
-  // Reminder: q.x and q.y can be negative
   if(x >= width || y >= height) return;
-  if(x+q.x >= width || y+q.y >= height) return;
-  if(x+q.x < 0 || y+q.y < 0) return;
+
+  int xpq = x + q.x;
+  int ypq = y + q.y;
+  // Convert out of bounds indexes to 0
+  // Reminder: q.x and q.y can be negative
+  xpq *= (x+q.x < width && x+q.x >= 0) ? 1.0f : 0.0f;
```
upegelow: I suggest to use int constants here.

rawfiner: Oops, indeed.
I perfectly agree, thanks!

rawfiner:
Thanks for testing and review :-)

PR #2005, opened 2019-01-15
Update preferences.c

PaoloAst:
On line 202, I'm proposing to change GTK_WIN_POS_CENTER with GTK_WIN_POS_CENTER_ON_PARENT to avoid a misplacement of the preferences windows if the program window is reduced or moved.


PaoloAst:
On my system (Ubuntu 18.10) if I start with dt smaller than my screen, the preferences window is displayed in the middle of the screen, not in the middle of the darktable windows, where I suppose it should be.
However my change is just to solve a cosmetic issue, so it is not so important.

PR #2013, opened 2019-01-18
D850 propose change of WB presets to newer firmware

bilddateien:
I've added WB presets for Nikon D850 with newer firmware 1.03 instead of 1.01 - please review and decide.

### DIFF
--- a/src/external/wb_presets.c
+++ b/src/external/wb_presets.c
@@ -3078,13 +3078,13 @@ const wb_data wb_preset[] =
   { "Nikon", "D850", HighTempMercuryVaporFluorescent , 0, { 2.379882812, 1, 1.3671875,   0 } },
 
   /*NIKON D850 Software Ver.1.03*/
-  { "NIKON CORPORATION", "NIKON D850", DirectSunlight      , 0, { 1.900390625, 1, 1.375488281, 0 } },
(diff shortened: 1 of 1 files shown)


src/external/wb_presets.c
```diff
@@ -3068,6 +3077,15 @@ const wb_data wb_preset[] =
   { "Nikon", "D850", DaylightFluorescent             , 0, { 2.018066406, 1, 1.112304688, 0 } },
   { "Nikon", "D850", HighTempMercuryVaporFluorescent , 0, { 2.379882812, 1, 1.3671875,   0 } },
 
+  /*NIKON D850 Software Ver.1.03*/
```
TurboGit: You need to:
1. remove the old D850 entries
2. do not add the D500 in the same commit
3. rename the entries from ""NIKON CORPORATION" to "Nikon" and "NIKON D850" to "D850".

I have done 3 for the D500.

Thanks.

bilddateien: 1. Ok, I wasn't sure about that since somehow the number of entries differ. Will do that later.
2. D500 was in #2012 
3. I just used the output of extract_wb and already wondered about the difference. Will do that later today.

TurboGit:
I'm closing. The values are exactly the same as previous one up to last digits. You did not removed the old entries anyway, so this cannot be applied. And your WB set is lot smaller than the current one. All in all, nothing new in there.

PR #2016, opened 2019-01-19
Bugfix: filtering of mouse move events flawed

vacaboja:
**Summary**
The following screenshots show two situations in which a star has remained active even though the mouse pointer has moved away from it. This is not a merely cosmetic issue, because it means that the click event will be interpreted incorrectly. This PR fixes it, and also gives a performance improvement of about a factor of 2 to the recently introduced logic that filters mouse move events.

dt-pointer-bug

**Details**
In fact, although similar in appearance, the two screenshots lead to two independent bugs. The second is trivial: the background was not always redrawn, and it also has no functional implications.

The first bug, which happens sporadically, is more complex, and it is due to bugs in the recently introduced logic that filters mouse move events. The existing behaviour is to filter all mouse move events, that happen in the central area of a thumbnail, because UI elements (such as stars and the reject button) are usually near the border. There are multiple inaccuracies in the implementation of this, which lead to some necessary redraws being filtered. By contrast, the algorithm implemented in this pull request, is to compute which element, if any, the pointer is hovering over, and to redraw only on change. This behaviour always redraws when needed, and it is also more efficient, because it redraws only when needed, while the existing implementation fails to filter about half the mouse move events (the area of the rim region over the total area).

### DIFF
--- a/src/views/lighttable.c
+++ b/src/views/lighttable.c
@@ -136,8 +136,10 @@ typedef struct dt_library_t
   int max_rows;
   int32_t single_img_id;
 
+  float pointed_img_x, pointed_img_y, pointed_img_wd, pointed_img_ht;
+  dt_view_image_over_t pointed_img_over;
+
   float thumb_size;
-  int last_mouse_over_thumb;
   int32_t last_exposed_id;
   float offset_x, offset_y;
   gboolean force_expose_all;
@@ -484,7 +486,7 @@ void init(dt_view_t *self)
   lib->single_img_id = -1;
 
   lib->thumb_size = -1;
-  lib->last_mouse_over_thumb = -1;
+  lib->pointed_img_over = DT_VIEW_ERR;
   lib->last_exposed_id = -1;
   lib->force_expose_all = FALSE;
   lib->offset_x = 0;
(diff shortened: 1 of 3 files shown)


TurboGit:
Looks promising indeed! I found a little issue, it is almost impossible to trigger the list of module used in the history which happen when over the little icon top-right of the thumbs. I get it some times, but very rarely.

vacaboja:
The ying-yang thing: yes I totally forgot about that. I'll look into it, thanks.

vacaboja:
Thanks, I fixed the bug.

Also, I took the opportunity to do a bit of refactoring of the `dt_view_image_expose()` function in view.c. This function is almost a thousand lines long and does everything needed to draw a thumbnail and all the associated paraphernalia. I separated the logic that deals with the active elements, such as stars etc., into a new function called `dt_view_process_image_over()`. The new function also checks if an element is activated by the mouse pointer, and a parameter can be used to tell it to skip the drawing and only do this check. So the code related to the positioning of control elements is not duplicated (it was in my initial PR). Finally, `dt_view_process_image_over()` is called also by `dt_view_guess_image_over()` to check whether the some element might activate and, ultimately, if a redraw is required.

Besides, I noticed that the tooltip associated with the ying-yang icon is there only in the file manager view, and not in the zoomable lighttable. Is it so by design?

TurboGit:
For the code duplicate that's nice! I was about to raise this issue and to at least have a comment saying that if the layout change your new function to detect sensitive elements! It is even better to see this done in the same code now.

For the ying-yang and zoomable lighttable, don't know. Maybe an old issue, I would say better to have it handled here if possible. This can be done in a separate PR if you like.

TurboGit:
Really nice clean-up! Thanks. All working fine and the maintenance will be easier.

PR #2017, opened 2019-01-19
Tonecurve - LCh & RGB independent tabs

phweyland:
As tonecurve changes are sometime critical for saturation I think it is interesting to be able to fix it in the same module. 

That's why I've made a try to add an LCh mode to tonecurve.
There are 3 tabs: L (same as for Lab), C(L) and C(h). The last two let the user apply a correction on C based on L and h.
I've experimented that C and h tabs are not very helpful, but C(L) and C(h) tabs seem to me more convenient.

I've found some difficulties due to the fact tonecurve is truly Lab oriented. Therefore there are lot of specificities which could be avoided adapting the design to a more general case. 

Let me know if that can be interesting.
Your comments are welcome.
If the function make sense for you I'll work the open cl piece.

### DIFF
--- a/data/kernels/basic.cl
+++ b/data/kernels/basic.cl
@@ -617,9 +617,11 @@ colorin_clipping (read_only image2d_t in, write_only image2d_t out, const int wi
 /* kernel for the tonecurve plugin. */
 kernel void
 tonecurve (read_only image2d_t in, write_only image2d_t out, const int width, const int height,
-           read_only image2d_t table_L, read_only image2d_t table_a, read_only image2d_t table_b,
-           const int autoscale_ab, const int unbound_ab, global float *coeffs_L, global float *coeffs_ab,
-           const float low_approximation)
+           const int tc_mode, const int unbound_ab, const float low_approximation,
+           read_only image2d_t table_0, read_only image2d_t table_1,
+           read_only image2d_t table_2, read_only image2d_t table_3,
+           global float *coeffs_0, global float *coeffs_1,
+           global float *coeffs_2, global float *coeffs_3)
 {
   const int x = get_global_id(0);
   const int y = get_global_id(1);
@@ -629,27 +631,27 @@ tonecurve (read_only image2d_t in, write_only image2d_t out, const int width, co
   float4 pixel = read_imagef(in, sampleri, (int2)(x, y));
   const float L_in = pixel.x/100.0f;
   // use lut or extrapolation:
-  const float L = lookup_unbounded(table_L, L_in, coeffs_L);
+  const float L = lookup_unbounded(table_0, L_in, coeffs_0);
 
-  if (autoscale_ab == 0)
+  if (tc_mode == 0)
   {
     const float a_in = (pixel.y + 128.0f) / 256.0f;
     const float b_in = (pixel.z + 128.0f) / 256.0f;
 
     if (unbound_ab == 0)
     {
-      pixel.y = lookup(table_a, a_in);
-      pixel.z = lookup(table_b, b_in);
+      pixel.y = lookup(table_1, a_in);
+      pixel.z = lookup(table_2, b_in);
     }
     else
     {
       // use lut or two-sided extrapolation
-      pixel.y = lookup_unbounded_twosided(table_a, a_in, coeffs_ab);
-      pixel.z = lookup_unbounded_twosided(table_b, b_in, coeffs_ab + 6);
+      pixel.y = lookup_unbounded_twosided(table_1, a_in, coeffs_1);
+      pixel.z = lookup_unbounded_twosided(table_2, b_in, coeffs_2);
     }
     pixel.x = L;
   }
-  else if(autoscale_ab == 1)
+  else if(tc_mode == 1)
   {
     if(L_in > 0.01f)
     {
@@ -663,21 +665,32 @@ tonecurve (read_only image2d_t in, write_only image2d_t out, const int width, co
     }
     pixel.x = L;
   }
-  else if(autoscale_ab == 2)
+  else if(tc_mode == 2)
   {
     float4 xyz = Lab_to_XYZ(pixel);
-    xyz.x = lookup_unbounded(table_L, xyz.x, coeffs_L);
-    xyz.y = lookup_unbounded(table_L, xyz.y, coeffs_L);
-    xyz.z = lookup_unbounded(table_L, xyz.z, coeffs_L);
+    xyz.x = lookup_unbounded(table_0, xyz.x, coeffs_0);
+    xyz.y = lookup_unbounded(table_0, xyz.y, coeffs_0);
+    xyz.z = lookup_unbounded(table_0, xyz.z, coeffs_0);
     pixel.xyz = XYZ_to_Lab(xyz).xyz;
   }
-  else if(autoscale_ab == 3)
+  else if(tc_mode == 3)
   {
     float4 rgb = Lab_to_prophotorgb(pixel);
-    rgb.x = lookup_unbounded(table_L, rgb.x, coeffs_L);
-    rgb.y = lookup_unbounded(table_L, rgb.y, coeffs_L);
-    rgb.z = lookup_unbounded(table_L, rgb.z, coeffs_L);
+    rgb.x = lookup_unbounded(table_0, rgb.x, coeffs_0);
+    rgb.y = lookup_unbounded(table_0, rgb.y, coeffs_0);
+    rgb.z = lookup_unbounded(table_0, rgb.z, coeffs_0);
+    rgb.x = lookup_unbounded(table_1, rgb.x, coeffs_1);
+    rgb.y = lookup_unbounded(table_2, rgb.y, coeffs_2);
+    rgb.z = lookup_unbounded(table_3, rgb.z, coeffs_3);
     pixel.xyz = prophotorgb_to_Lab(rgb).xyz;
+  }
+  else if(tc_mode == 4)
+  {
+    float4 lch = Lab_2_LCH(pixel);
+    lch.x = L;
+    lch.y = lch.y * lookup_unbounded(table_1, L_in, coeffs_1) * 2.0f;
+    lch.y = lch.y * lookup_unbounded(table_2, lch.z, coeffs_2) * 2.0f;
+    pixel.xyz = LCH_2_Lab(lch).xyz;
   }
 
   write_imagef (out, (int2)(x, y), pixel);
--- a/data/kernels/basic.cl
+++ b/data/kernels/basic.cl
@@ -618,6 +618,7 @@ colorin_clipping (read_only image2d_t in, write_only image2d_t out, const int wi
 kernel void
 tonecurve (read_only image2d_t in, write_only image2d_t out, const int width, const int height,
            const int tc_mode, const int unbound_ab, const float low_approximation,
+           /*const int histogram_needed, global int *local_histogram,*/
            read_only image2d_t table_0, read_only image2d_t table_1,
            read_only image2d_t table_2, read_only image2d_t table_3,
            global float *coeffs_0, global float *coeffs_1,
@@ -679,14 +680,42 @@ tonecurve (read_only image2d_t in, write_only image2d_t out, const int width, co
     rgb.x = lookup_unbounded(table_0, rgb.x, coeffs_0);
     rgb.y = lookup_unbounded(table_0, rgb.y, coeffs_0);
     rgb.z = lookup_unbounded(table_0, rgb.z, coeffs_0);
+    pixel.xyz = prophotorgb_to_Lab(rgb).xyz;
+  }
+  else if(tc_mode == 4)
+  {
+    float4 rgb = Lab_to_prophotorgb(pixel);
+/*    if (histogram_needed)
+    { // process local histogram
+      local_histogram[4 * clamp((int)(pixel.x*2.55f), 0, 255)]++;
+      const float rgba[3] = {rgb.x, rgb.y, rgb.z};
+      for(int c=0; c<3; c++)
+      {
+        float4 fake_rgb;
+        fake_rgb.x = fake_rgb.y = fake_rgb.z = rgba[c];
+        float4 lab = prophotorgb_to_Lab(fake_rgb);
+        // histogram_RGB can be put to 255 bins
+        local_histogram[4 * clamp((int)(lab.x*2.55f), 0, 255) + c+1]++;
+      }
+    } */
+    rgb.x = lookup_unbounded(table_0, rgb.x, coeffs_0);
+    rgb.y = lookup_unbounded(table_0, rgb.y, coeffs_0);
+    rgb.z = lookup_unbounded(table_0, rgb.z, coeffs_0);
     rgb.x = lookup_unbounded(table_1, rgb.x, coeffs_1);
     rgb.y = lookup_unbounded(table_2, rgb.y, coeffs_2);
     rgb.z = lookup_unbounded(table_3, rgb.z, coeffs_3);
     pixel.xyz = prophotorgb_to_Lab(rgb).xyz;
   }
-  else if(tc_mode == 4)
+  else if(tc_mode == 5)
   {
     float4 lch = Lab_2_LCH(pixel);
+/*    if (histogram_needed)
+    { // process local histogram
+      // histogram_LCh can be put to 255 bins
+      local_histogram[4 * clamp((int)(lch.x*2.55f), 0, 255)]++;
+      local_histogram[4 * clamp((int)(lch.y*1.40095f), 0, 255) + 1]++;  // 255 / 182.019 = 1.40095
+      local_histogram[4 * clamp((int)(lch.z*255.0f), 0, 255) + 2]++;
+    } */
     lch.x = L;
     lch.y = lch.y * lookup_unbounded(table_1, L_in, coeffs_1) * 2.0f;
     lch.y = lch.y * lookup_unbounded(table_2, lch.z, coeffs_2) * 2.0f;
--- a/data/kernels/basic.cl
+++ b/data/kernels/basic.cl
@@ -617,26 +617,22 @@ colorin_clipping (read_only image2d_t in, write_only image2d_t out, const int wi
 float
 rgb_norm_vect (float4 rgb, int rgb_norm, float norm_exp)
 {
+  if(rgb.x < 0.0f || rgb.y < 0.0f || rgb.z < 0.0f) return -1;
+
   switch(rgb_norm)
   {
   case 1:  // norm L infinite = max
     {
       return max(rgb.x, max(rgb.y, rgb.z));
     }
-  case 2:  // norm L1 - bypass the powf for performance
+  case 2:  // general Lp norm (pseudo-norm if p < 1) - slow variant
     {
-      return (rgb.x + rgb.y + rgb.z) / 3;
-    }
-  case 3:  // norm L2 = euclidian norm - bypass the powf for performance
-    {
-      return sqrt((rgb.x * rgb.x + rgb.y * rgb.y + rgb.z * rgb.z) / 3);
-    }
-  case 4:  // general Lp norm (pseudo-norm if p < 1) - slow variant
-    {
-      return pow((pow(rgb.x, norm_exp) + pow(rgb.y, norm_exp) + pow(rgb.z, norm_exp)) / 3, 1.0f/norm_exp);
+      return native_powr((native_powr(rgb.x, norm_exp) +
+                native_powr(rgb.y, norm_exp) +
+                native_powr(rgb.z, norm_exp)), 1.0f/norm_exp);
       break;
     }
-  case 5:  // basic power norm
+  case 3:  // basic power norm
     {
       float R, G, B;
       R = rgb.x * rgb.x;
@@ -644,14 +640,13 @@ rgb_norm_vect (float4 rgb, int rgb_norm, float norm_exp)
       B = rgb.z * rgb.z;
       return (rgb.x * R + rgb.y * G + rgb.z * B) / (R + G + B);
     }
-  case 6: // weighted yellow power norm
+  case 4: // weighted yellow power norm
     {
-      float R, G, B;
-      R = 1.22f * rgb.x * 1.22f * rgb.x;
-      G = 1.20f * rgb.y * 1.20f * rgb.y;
-      B = 0.58f * rgb.z * 0.58f * rgb.z;
-      R *= R; G *= G; B *= B;
-      return 0.83743219f * (1.22f * rgb.x * R + 1.20f * rgb.y * G + 0.58 * rgb.z * B) / (R + G + B);
+      const float4 coeff = {1.22f, 1.20f, 0.58f, 1.0f};
+      const float4 coeff4 = {2.21533456f, 2.0736f, 0.11316496f, 1.0f};  // 1.22^4, 1.20^4, 0.58^4
+      const float4 rgb4 = rgb * rgb * rgb * rgb * coeff4;
+      const float4 rgb5 = rgb4 * rgb * coeff;
+      return 0.83743219f * (rgb5.x + rgb5.y + rgb5.z) / (rgb4.x + rgb4.y + rgb4.z);
     }
   default: {return -1;}
   }
@@ -661,7 +656,7 @@ rgb_norm_vect (float4 rgb, int rgb_norm, float norm_exp)
 kernel void
 tonecurve (read_only image2d_t in, write_only image2d_t out, const int width, const int height,
            const int tc_mode, const int unbound_ab, const float low_approximation,
-           const int rgb_norm,
+           const int rgb_norm, const float rgb_norm_exp,
            read_only image2d_t table_0, read_only image2d_t table_1,
            read_only image2d_t table_2, read_only image2d_t table_3,
            global float *coeffs_0, global float *coeffs_1,
@@ -723,12 +718,15 @@ tonecurve (read_only image2d_t in, write_only image2d_t out, const int width, co
     float norm = 0;
     if (rgb_norm != 0)
     {
-      norm = rgb_norm_vect( rgb, rgb_norm, 0.333333f);
-      norm = lookup_unbounded(table_0, norm, coeffs_0) / norm;
-      rgb.xyz *= norm;
+      float norm = rgb_norm_vect( rgb, rgb_norm, rgb_norm_exp); // compute the norm == luminance estimator
+      if (norm < 0) goto no_norm;
+      const float4 rgb_ratios = rgb / norm; // these ratios are the actual colors, independent from the luminance
+      norm = lookup_unbounded(table_0, norm, coeffs_0); // compute the curve on the luminance
+      rgb.xyz = (norm * rgb_ratios).xyz; // restore the colors from the original ratios and the new luminance
     }
     else
     {
+no_norm:
       rgb.x = lookup_unbounded(table_0, rgb.x, coeffs_0);
       rgb.y = lookup_unbounded(table_0, rgb.y, coeffs_0);
       rgb.z = lookup_unbounded(table_0, rgb.z, coeffs_0);
--- a/data/kernels/basic.cl
+++ b/data/kernels/basic.cl
@@ -614,11 +614,54 @@ colorin_clipping (read_only image2d_t in, write_only image2d_t out, const int wi
   write_imagef (out, (int2)(x, y), pixel);
 }
 
+float
+rgb_norm_vect (float4 rgb, int rgb_norm, float norm_exp)
+{
+  switch(rgb_norm)
+  {
+  case 1:  // norm L infinite = max
+    {
+      return max(rgb.x, max(rgb.y, rgb.z));
+    }
+  case 2:  // norm L1 - bypass the powf for performance
+    {
+      return (rgb.x + rgb.y + rgb.z) / 3;
+    }
+  case 3:  // norm L2 = euclidian norm - bypass the powf for performance
+    {
+      return sqrt((rgb.x * rgb.x + rgb.y * rgb.y + rgb.z * rgb.z) / 3);
+    }
+  case 4:  // general Lp norm (pseudo-norm if p < 1) - slow variant
+    {
+      return pow((pow(rgb.x, norm_exp) + pow(rgb.y, norm_exp) + pow(rgb.z, norm_exp)) / 3, 1.0f/norm_exp);
+      break;
+    }
+  case 5:  // basic power norm
+    {
+      float R, G, B;
+      R = rgb.x * rgb.x;
+      G = rgb.y * rgb.y;
+      B = rgb.z * rgb.z;
+      return (rgb.x * R + rgb.y * G + rgb.z * B) / (R + G + B);
+    }
+  case 6: // weighted yellow power norm
+    {
+      float R, G, B;
+      R = 1.22f * rgb.x * 1.22f * rgb.x;
+      G = 1.20f * rgb.y * 1.20f * rgb.y;
+      B = 0.58f * rgb.z * 0.58f * rgb.z;
+      R *= R; G *= G; B *= B;
+      return 0.83743219f * (1.22f * rgb.x * R + 1.20f * rgb.y * G + 0.58 * rgb.z * B) / (R + G + B);
+    }
+  default: {return -1;}
+  }
+}
+
 /* kernel for the tonecurve plugin. */
 kernel void
 tonecurve (read_only image2d_t in, write_only image2d_t out, const int width, const int height,
            const int tc_mode, const int unbound_ab, const float low_approximation,
-           /*const int histogram_needed, global int *local_histogram,*/
+           const int rgb_norm,
            read_only image2d_t table_0, read_only image2d_t table_1,
            read_only image2d_t table_2, read_only image2d_t table_3,
            global float *coeffs_0, global float *coeffs_1,
@@ -677,27 +720,24 @@ tonecurve (read_only image2d_t in, write_only image2d_t out, const int width, co
   else if(tc_mode == 3)
   {
     float4 rgb = Lab_to_prophotorgb(pixel);
-    rgb.x = lookup_unbounded(table_0, rgb.x, coeffs_0);
-    rgb.y = lookup_unbounded(table_0, rgb.y, coeffs_0);
-    rgb.z = lookup_unbounded(table_0, rgb.z, coeffs_0);
-    pixel.xyz = prophotorgb_to_Lab(rgb).xyz;
+    float norm = 0;
+    if (rgb_norm != 0)
+    {
+      norm = rgb_norm_vect( rgb, rgb_norm, 0.333333f);
+      norm = lookup_unbounded(table_0, norm, coeffs_0) / norm;
+      rgb.xyz *= norm;
+    }
+    else
+    {
+      rgb.x = lookup_unbounded(table_0, rgb.x, coeffs_0);
+      rgb.y = lookup_unbounded(table_0, rgb.y, coeffs_0);
+      rgb.z = lookup_unbounded(table_0, rgb.z, coeffs_0);
+    }
+  pixel.xyz = prophotorgb_to_Lab(rgb).xyz;
   }
   else if(tc_mode == 4)
   {
     float4 rgb = Lab_to_prophotorgb(pixel);
-/*    if (histogram_needed)
-    { // process local histogram
-      local_histogram[4 * clamp((int)(pixel.x*2.55f), 0, 255)]++;
-      const float rgba[3] = {rgb.x, rgb.y, rgb.z};
-      for(int c=0; c<3; c++)
-      {
-        float4 fake_rgb;
-        fake_rgb.x = fake_rgb.y = fake_rgb.z = rgba[c];
-        float4 lab = prophotorgb_to_Lab(fake_rgb);
-        // histogram_RGB can be put to 255 bins
-        local_histogram[4 * clamp((int)(lab.x*2.55f), 0, 255) + c+1]++;
-      }
-    } */
     rgb.x = lookup_unbounded(table_0, rgb.x, coeffs_0);
     rgb.y = lookup_unbounded(table_0, rgb.y, coeffs_0);
     rgb.z = lookup_unbounded(table_0, rgb.z, coeffs_0);
@@ -709,13 +749,6 @@ tonecurve (read_only image2d_t in, write_only image2d_t out, const int width, co
   else if(tc_mode == 5)
   {
     float4 lch = Lab_2_LCH(pixel);
-/*    if (histogram_needed)
-    { // process local histogram
-      // histogram_LCh can be put to 255 bins
-      local_histogram[4 * clamp((int)(lch.x*2.55f), 0, 255)]++;
-      local_histogram[4 * clamp((int)(lch.y*1.40095f), 0, 255) + 1]++;  // 255 / 182.019 = 1.40095
-      local_histogram[4 * clamp((int)(lch.z*255.0f), 0, 255) + 2]++;
-    } */
     lch.x = L;
     lch.y = lch.y * lookup_unbounded(table_1, L_in, coeffs_1) * 2.0f;
     lch.y = lch.y * lookup_unbounded(table_2, lch.z, coeffs_2) * 2.0f;
--- a/data/kernels/basic.cl
+++ b/data/kernels/basic.cl
@@ -625,14 +625,18 @@ rgb_norm_vect (float4 rgb, int rgb_norm, float norm_exp)
     {
       return max(rgb.x, max(rgb.y, rgb.z));
     }
-  case 2:  // general Lp norm (pseudo-norm if p < 1) - slow variant
+  case 2:  // norm L average(rgb)
+    {
+      return (rgb.x + rgb.y + rgb.z) / 3.0f;
+    }
+  case 3:  // general Lp norm (pseudo-norm if p < 1) - slow variant
     {
       return native_powr((native_powr(rgb.x, norm_exp) +
                 native_powr(rgb.y, norm_exp) +
                 native_powr(rgb.z, norm_exp)), 1.0f/norm_exp);
       break;
     }
-  case 3:  // basic power norm
+  case 4:  // basic power norm
     {
       float R, G, B;
       R = rgb.x * rgb.x;
@@ -640,7 +644,7 @@ rgb_norm_vect (float4 rgb, int rgb_norm, float norm_exp)
       B = rgb.z * rgb.z;
       return (rgb.x * R + rgb.y * G + rgb.z * B) / (R + G + B);
     }
-  case 4: // weighted yellow power norm
+  case 5: // weighted yellow power norm
     {
       const float4 coeff = {1.22f, 1.20f, 0.58f, 1.0f};
       const float4 coeff4 = {2.21533456f, 2.0736f, 0.11316496f, 1.0f};  // 1.22^4, 1.20^4, 0.58^4
--- a/src/iop/tonecurve.c
+++ b/src/iop/tonecurve.c
@@ -153,6 +153,10 @@ typedef struct dt_iop_tonecurve_gui_data_t
   float loglogscale;
   int scale_mode;
   GtkWidget *logbase;
+  // local histogram
+  uint32_t local_histogram[256 * DT_IOP_TONECURVE_MAX_CH];
+  // maximum levels in histogram, one per channel
+  uint32_t local_histogram_max[DT_IOP_TONECURVE_MAX_CH];
 } dt_iop_tonecurve_gui_data_t;
 
 typedef struct dt_iop_tonecurve_data_t
@@ -193,15 +197,15 @@ static const struct
   // 4 - C(h) (LCH_2_Lab)
   {2, 1, 0.0f, 1.0f, 0.0f, 360.0f, 1.0f, 0, 0, 1, 2, { { 0.0f, 0.5f }, { 1.0f, 0.5f } }},
   // 5 - L(RGB) - L(XYZ)
-  {0, 0, 0.0f, 1.0f, 0.0f, 100.0f, 100.0f, 1, 1, 1, 2, { { 0.0f, 0.0f }, { 1.0f, 1.0f } }},
+  {0, 0, 0.0f, 100.0f, 0.0f, 100.0f, 100.0f, 1, 1, 1, 2, { { 0.0f, 0.0f }, { 1.0f, 1.0f } }},
   // 6 - L(LRGB)
-  {0, 0, 0.0f, 1.0f, 0.0f, 100.0f, 1.0f, 1, 1, 1, 2, { { 0.0f, 0.0f }, { 1.0f, 1.0f } }},
+  {0, 0, 0.0f, 100.0f, 0.0f, 100.0f, 1.0f, 1, 1, 1, 2, { { 0.0f, 0.0f }, { 1.0f, 1.0f } }},
   // 7 - R (LRGB)
-  {1, 1, 0.0f, 1.0f, 0.0f, 100.0f, 1.0f, 0, 1, 1, 2, { { 0.0f, 0.0f}, { 1.0f, 1.0f } }},
+  {1, 1, 0.0f, 100.0f, 0.0f, 100.0f, 1.0f, 1, 1, 1, 2, { { 0.0f, 0.0f}, { 1.0f, 1.0f } }},
   // 8 - G (LRGB)
-  {2, 2, 0.0f, 1.0f, 0.0f, 100.0f, 1.0f, 0, 1, 1, 2, { { 0.0f, 0.0f }, { 1.0f, 1.0f } }},
+  {2, 2, 0.0f, 100.0f, 0.0f, 100.0f, 1.0f, 1, 1, 1, 2, { { 0.0f, 0.0f }, { 1.0f, 1.0f } }},
   // 9 - B (LRGB)
-  {3, 3, 0.0f, 1.0f, 0.0f, 100.0f, 1.0f, 0, 1, 1, 2, { { 0.0f, 0.0f }, { 1.0f, 1.0f } }}
+  {3, 3, 0.0f, 100.0f, 0.0f, 100.0f, 1.0f, 1, 1, 1, 2, { { 0.0f, 0.0f }, { 1.0f, 1.0f } }}
 };
 
 typedef enum dt_iop_tonecurve_UI_colorspace_t // could use the standard dt colorspace enum
@@ -334,7 +338,6 @@ int process_cl(struct dt_iop_module_t *self, dt_dev_pixelpipe_iop_t *piece, cl_m
   const int width = roi_in->width;
   const int height = roi_in->height;
   const int tc_mode = d->tc_mode;
-//  const int nb_ch = mode_curves[tc_mode].nb_ch;
   const int unbound_ab = d->unbound_ab;
   const float low_approximation = d->table[0][(int)(0.01f * 0x10000ul)];
 
@@ -387,6 +390,7 @@ void process(struct dt_iop_module_t *self, dt_dev_pixelpipe_iop_t *piece, const 
 {
   const int ch = piece->colors;
   dt_iop_tonecurve_data_t *d = (dt_iop_tonecurve_data_t *)(piece->data);
+  dt_iop_tonecurve_gui_data_t *g = (dt_iop_tonecurve_gui_data_t *)self->gui_data;
   const float xm[DT_IOP_TONECURVE_MAX_CH][2] = {
     {1.0f / d->unbounded_coeffs[0][0], 1.0f - 1.0f / d->unbounded_coeffs[0][3]},
     {1.0f / d->unbounded_coeffs[1][0], 1.0f - 1.0f / d->unbounded_coeffs[1][3]},
@@ -398,9 +402,12 @@ void process(struct dt_iop_module_t *self, dt_dev_pixelpipe_iop_t *piece, const 
   const int height = roi_out->height;
   const int tc_mode = d->tc_mode;
   const int unbound_ab = d->unbound_ab;
+  if (g)
+   for (int c = 0; c < DT_IOP_TONECURVE_MAX_CH * 256; c++ )
+      g->local_histogram[c] = 0;
 
 #ifdef _OPENMP
-#pragma omp parallel for default(none) shared(d) schedule(static)
+#pragma omp parallel for default(none) shared(d,g) schedule(static)
 #endif
   for(int k = 0; k < height; k++)
   {
@@ -494,6 +501,19 @@ void process(struct dt_iop_module_t *self, dt_dev_pixelpipe_iop_t *piece, const 
       {
         float rgb[3] = {0, 0, 0};
         dt_Lab_to_prophotorgb(in, rgb);
+        if (g)
+        { // process local histogram
+          for(int c=0; c<3; c++)
+          {
+            const float fake_rgb[3] = {rgb[c], rgb[c], rgb[c]};
+            float lab[3];
+            // convert colors to pseudo-Lab for UI:
+            // only the [0] element of the vector is relevant (pseudo-L)
+            dt_prophotorgb_to_Lab(fake_rgb, lab);
+            // histogram_RGB can be put to 255 bins
+            g->local_histogram[4 * (uint8_t)CLAMP(lab[0]*2.55f, 0.0f, 255.0f) + c+1]++;
+          }
+        }
         for(int c=0; c<3; c++)
         {
           rgb[c] = (rgb[c] < xm[ch_L][0]) ? d->table[ch_L][CLAMP((int)(rgb[c] * 0x10000ul), 0, 0xffff)]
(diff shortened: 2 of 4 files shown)


src/iop/tonecurve.c
```diff
@@ -344,6 +347,19 @@ void process(struct dt_iop_module_t *self, dt_dev_pixelpipe_iop_t *piece, const
                                          : d->table[ch_b][CLAMP((int)(b_in * 0x10000ul), 0, 0xffff)]);
         }
       }
+      else if(autoscale_ab == DT_S_SCALE_MANUAL_LCH)
+      {
+        const float pi = 3.14159f;
+        const float C_in = sqrtf(in[1]*in[1] + in[2]*in[2]) / 181.019f;
+        const float h_in = ((in[1]!=0.0f ? atanf(in[2]/in[1]) : (in[2]>0.0f ? 0.5f : -0.5f) * pi) + ((in[1] < 0.0f) ? pi : ((in[2] < 0.0f) ? 2.0f * pi : 0.0f))) / (2.0f * pi);
```
aurelienpierre: there is an `atan2(y, x)` function doing just that (computing atan in the right quadrant)

src/iop/tonecurve.c
```diff
@@ -344,6 +347,19 @@ void process(struct dt_iop_module_t *self, dt_dev_pixelpipe_iop_t *piece, const
...
+        float C_out = L_in == 0.0f ? C_in : C_in * d->table[ch_CL][CLAMP((int)(L_in * 0x10000ul), 0, 0xffff)] * 0.02f;
```
aurelienpierre: Please write a color conversion function in the lib  `src/common/colorspaces_inline_conversion.h`

src/iop/tonecurve.c
```diff
@@ -344,6 +347,19 @@ void process(struct dt_iop_module_t *self, dt_dev_pixelpipe_iop_t *piece, const
...
+        C_out = h_in == 0.0f ? C_out : C_out * d->table[ch_Ch][CLAMP((int)(h_in * 0x10000ul), 0, 0xffff)] * 0.02f;
```
aurelienpierre: so you are shifting input hue by the same amount you shift the chroma ? Don't you want to preserve the hue untouched ? I don't understand what's going on here.

src/iop/tonecurve.c
```diff
@@ -344,6 +347,19 @@ void process(struct dt_iop_module_t *self, dt_dev_pixelpipe_iop_t *piece, const
...
+        C_out = C_out * 181.019f;
+        const float h_out = h_in * (2.0f * pi);
```
aurelienpierre: I think you should handle hue shifts directly in radians to spare a few divisions and multiplications, if you really want to rotate the hue.

src/iop/tonecurve.c
```diff
@@ -344,6 +347,19 @@ void process(struct dt_iop_module_t *self, dt_dev_pixelpipe_iop_t *piece, const
...
+
+        out[1] = C_out * cosf(h_out);
```
aurelienpierre: you might want to use Taylor-Young approximations instead of pure cos/sin to speed-up the computation

src/iop/tonecurve.c
```diff
@@ -1314,19 +1411,23 @@ static void picker_scale(const float *in, float *out)
   out[2] = CLAMP((in[2] + 128.0f) / 256.0f, 0.0f, 1.0f);
 }
 
+static void picker_scale_LCh(const float *in, float *out)
+{
+  const float pi = 3.14159f;
+  float C_m = sqrtf(in[1]*in[1] + in[2]*in[2]) / 181.019f;
+  float h_m = (atanf(in[2]/in[1]) + ((in[1] < 0.0f) ? pi : ((in[2] < 0.0f) ? 2.0f * pi : 0.0f))) / (2.0f * pi);
```
aurelienpierre: same, you might want to use `atan2`

src/iop/tonecurve.c
```diff
@@ -344,6 +347,19 @@ void process(struct dt_iop_module_t *self, dt_dev_pixelpipe_iop_t *piece, const
                                          : d->table[ch_b][CLAMP((int)(b_in * 0x10000ul), 0, 0xffff)]);
         }
       }
+      else if(autoscale_ab == DT_S_SCALE_MANUAL_LCH)
+      {
+        const float pi = 3.14159f;
```
aurelienpierre: pi is already set as a global constant `M_PI` I think

src/iop/tonecurve.c
aurelienpierre: I have a SSE one in progress:
```
static inline __m128 dt_Lab_to_Lch_sse2(const __m128 Lab)
{
  __m128 Lch = Lab;
  __m128 Lab2 = Lab * Lab;
  Lch[1] = powf((Lab2[1] + Lab2[2]), 0.5f);
  Lch[2] = atan2f(Lab[2], Lab[1]);
  return Lch;
}

static inline __m128 dt_Lch_to_Lab_sse2(const __m128 Lch)
{
  __m128 Lab = Lch;
  Lab[1] = cosf(Lch[2]) * Lch[1];
  Lab[2] = sinf(Lch[2]) * Lch[1];
  return Lab;
}
```

src/iop/tonecurve.c
```diff
@@ -344,6 +347,19 @@ void process(struct dt_iop_module_t *self, dt_dev_pixelpipe_iop_t *piece, const
...
+        const float C_in = sqrtf(in[1]*in[1] + in[2]*in[2]) / 181.019f;
+        const float h_in = ((in[1]!=0.0f ? atanf(in[2]/in[1]) : (in[2]>0.0f ? 0.5f : -0.5f) * pi) + ((in[1] < 0.0f) ? pi : ((in[2] < 0.0f) ? 2.0f * pi : 0.0f))) / (2.0f * pi);
+        float C_out = L_in == 0.0f ? C_in : C_in * d->table[ch_CL][CLAMP((int)(L_in * 0x10000ul), 0, 0xffff)] * 0.02f;
+        C_out = h_in == 0.0f ? C_out : C_out * d->table[ch_Ch][CLAMP((int)(h_in * 0x10000ul), 0, 0xffff)] * 0.02f;
+        C_out = C_out * 181.019f;
```
aurelienpierre: same here, I think building your curve LUT between 0 and 181.019 would spare one mul and one div by pixel

aurelienpierre:
I don't understand why the hue gets involved in the chroma correction. The chroma is the colorfulness, more or less equivalent to the saturation, you usually want to keep it separated from the hue (that's the whole purpose of the Lch space). Hues shifts are nearly impossible to handle with curves, hence the color zones module, so I think you should keep only the L and C channels, keep C away from any interaction with hue, and let the hue handling for the color zones module.

src/iop/tonecurve.c
phweyland: ok. done

src/iop/tonecurve.c
phweyland: Great ! Done.

src/iop/tonecurve.c
phweyland: Hue doesn't change. C is modified for the selected hue.

src/iop/tonecurve.c
phweyland: No hue shift. Hue is just used as an index in the C table (needs 0->1).
Div and mul saving done.

src/iop/tonecurve.c
phweyland: LUT kept as is but saving done.

src/iop/tonecurve.c
phweyland: Need more time to understand this.

phweyland:
One question. 
Adding LCh doesn't change any parameter. From this standpoint there is no difference between version 4 and 5. Therefore is the version 5 necessary, desirable ?

src/iop/tonecurve.c
phweyland: I've added open cl support as done in the module. It uses basic.cl and colorspaces.cl. 
It doesn't use SSE. Could be interesting to improve that.

TurboGit:
@phweyland : no if there is no new parameter and if the current edit are 100% compatible with this new version you do not have to change the version.

What I understand from your comment (I didn't look at the code yet) is that this PR is only about a new GUI interface which are actually triggering the current parameters in a different way. Right?

phweyland:
>this PR is only about a new GUI interface which are actually triggering the current parameters in a different way. Right?

Exact.

> if there is no new parameter and if the current edit are 100% compatible with this new version you do not have to change the version.

OK, I'll remove the corresponding code.

phweyland:
> This will introduce conflicts with PR 1841, maybe it will be better to wait until it is merged.

Which kind of conflict ?

edgardoh:
PR 1841 adds rgb curves, so more or less the same lines has been modified. But more important, it adds the ability for the modules to ask for a specific colorspace, so the transform must be done outside the iop. This allows the module to have a LCh histogram, but for that the curve will have to be modified, so strange things will have to be done to keep backwards compatibility.

phweyland:
@edgardoh, thanks for explanation. Do you have any idea when this merge is expected ?

phweyland:
I've noticed an issue on data shifts. I'm unsure how to handle that.
The goal of LCh is to make changes on L and C letting the other channels untouched.
This never happens ...
Some screen captures here:
[tonecurve_calculation_issues.docx](https://github.com/darktable-org/darktable/files/2778293/tonecurve_calculation_issues.docx)
- Changing Luminance shifts a & b on color picker (same for LAB independant channels).
- Changing Chroma shifts Luminance and Hue.

That may be a normal behaviour but that's disturbing to have a measure (color picker) which contradicts what we try to do ... :-)

Where and when color picker does make the measure (before or after output color profile) ?
What do you think ?
Is it possible to improve this ?

edgardoh:
@phweyland , is not really up to me when it will be merged, but don't expect it for next week.
I'm not sure I understand you question, right now we have Lab, XYZ and RGB curves, independent or linked, each one has a different behaviour and that's hard-coded.

phweyland:
>I'm not sure I understand you question

No problem, the question is maybe stupid. I'm looking forward to seeing the new code.

aurelienpierre:
> I've noticed an issue on data shifts. I'm unsure how to handle that.
> The goal of LCh is to make changes on L and C letting the other channels untouched.
> This never happens ...
> Some screen captures here:
> [tonecurve_calculation_issues.docx](https://github.com/darktable-org/darktable/files/2778293/tonecurve_calculation_issues.docx)
> 
>     * Changing Luminance shifts a & b on color picker (same for LAB independant channels).
> 
>     * Changing Chroma shifts Luminance and Hue.
> 
> 
> That may be a normal behaviour but that's disturbing to have a measure (color picker) which contradicts what we try to do ... :-)
> 
> Where and when color picker does make the measure (before or after output color profile) ?
> What do you think ?
> Is it possible to improve this ?

That's a flaw of Lch/Lab (1976) color space itself… it's not *really* hue-invariant nor chroma-invariant.

A solution would be to use another color space like IPT-HDR (2011), and I'm working on just that these days.

aurelienpierre:
On another note, if you don't affect the hue, you don't need `sinf()` or `cosf()` and you can speed things up:

1. compute the chroma (aka euclidian distance, aka radius): C = (a^2 + b^2)^0.5
1. compute and store the cosinus = a / C
1. compute and store the sinus = b / C
1. compute the hue (aka the angle):
    - either atan2(b, a)
    - or sign(sinus) * sign(cosinus) * atan(sinus / cosinus) (maybe faster)
1. change C = funct(hue)
1. restore a = cosinus * C
1. restore b = sinus * C

phweyland:
>  if you don't affect the hue, you don't need sinf() or cosf() 

Excellent ! I should have seen that. I'll do it.
> it's not really hue-invariant nor chroma-invariant.

But maths on Lab<->LCh do not say that ... what does happen ?

aurelienpierre:
> But maths on Lab<->LCh do not say that ... what does happen ?
Hold on. Do you mean Lab -> Lch -> Lab is not an identity (ie Lab input ≠ Lab output), or Lab -> Lch -> c = funct(hue) -> Lab (ie chroma adjustement) does not retain the a and b values (or ratios) ?

aurelienpierre:
what you see seems normal to me. First of all, the global color picker works after the output color profile and the possible gamut compression it does, at the far end of the pipe, by converting back output RGB to Lab. So, don't bother to compare the values with the local color picker in tone curve, which works between the input and output of the module : too many things happen in-between.

Then, Lab is a crappy non-linear old attempt to separate color and luminance in a perceptual way. Keeping the numeric hue (as the angle betwen (a, b)) unchanged doesn't mean the color you actually see won't change. Convolve that with an RGB color profile conversion, and you get unpredictible results. So, overall I think you are just hitting the limits of the Lab space.

edgardoh:
@phweyland , if you are planing on continue with this then you can disregard my comment, my current approach don't work like I would like to so I'll have to change it and I don't know how long it will take, and I don't want to hold you back.
In fact, I want to do something as generic as possible, so having also LCh will help.

phweyland:
@edgardoh, are you talking about tc-rgb or PR1841 ?
I was thinking about rgb too, kind of LRGB (4 tabs) independent mode. In that case the module should be a bit refactored (1 or 3 tabs are hard coded everywhere). Was it your idea too ? It could be interesting to share our views. But this is not the subject of that thread, we may talk about this in another place.

edgardoh:
I added the rgb curve mainly to test some of the new functionality introduced on that PR, and what would take to add it to any module, so I just added it on top of the existent code, I did not have any redesign in mind.
Things seems to work fine, but there's the issue I describe on that PR, right now rgb/xyz curve display a lab histogram, that doesn't seem right. With my changes, xyz/linear rgb display a xyz/linear rgb histogram, I don't like that either.
The solution is probably independent of PR 1841, but maybe is related, and I think that any redesign of the tc should have it in mind. Your input is welcome on that, on any thread you choose.

About the 3 or 4 tabs, I just added a new option "rgb independent channels" to keep it consistent with the lab mode, also I didn't want to add another LUT, but is really the same to me.

phweyland:
@edgardoh. I've had a look at your code and effectively our code overlap everywhere.

I suggest to introduce RGB independent mode along with LCh (taking your code where necessary).
I think it would be better to transform RGB linked channels to RGB independent channels with 4 tabs (L, R, G, B). That's more work but I think it's better for the user to have the 4 tabs in the same module.
In that version I would exclude obviously the call to color space you are implementing and also histogram, but with your help, we can try to turn the next step (PR 1841) as easy as possible.

Would that make sense ?

edgardoh:
Sounds ok to me. Also, depending on what's decided about the histogram, my PR may not be needed at all. For instance, you can add the option for the modules to ask for a histogram on a specific colorspace and ready to display, in this case the module can still have the input image in Lab and display the histogram on all the colorspaces.

edgardoh:
> That sounds great but how can I ask for an histogram on a specific
> colorspace ?
>
By add I meant it has to be added (coded)

Second point. Your mode RGB linked channel doesn't behave as the normal
> one, as the later shows a different histogram (same as Lab in fact). This
> may be a problem for backward compatibility (nodes at different positions).
> Any thought ?
>
That's the issue I have described with my version, maybe its not clear, so
let's forget about my PR for a moment.
Right now a Lab histogram is displayed for all the curves, including XYZ
and RGB linked. I don't like it, but is an option. If you decide to keep
the current behavior then you can display a L(ab) histogram on L(Ch) and
nothing on (L)Ch. For the RGB independent channels you can display L(ab)
for each channel (doesn't seem to be reasonable) or display nothing.
If you want to display a histogram for each channel on every colorspace the
only option I see is to generate a histogram for each one, but when the
colorspace don't have a 50% grey an S curve will cause a shift on the
middle grey (that's what my version on my PR is doing), I don't think that
this is correct.
Another option is to generate a histogram for each colorspace but somehow
with 50% grey, That's what I mean by "histogram ready for display",
obviously this has to be better defined.
Depending on what you decide to do are the next steps to take.

I hope this makes sense.

phweyland:
Added a mode for LRGB, without histogram.
Open cl not yet updated.
Is there somewhere in dt some common definition for :
- LCH colorspace
- RGB luminance coefficients 
?

aurelienpierre:
What is RGB luminance coefficients ?

phweyland:
@TurboGit, There is one line conflict. What should I do ?

@edgardoh, I'd a look at histogram code. I've not found the way to delude the system to make calculate the histogram in the needed color space. And coding all this seems to be a lot of work which will be useless as soon your PR is on. Is not better to wait for your PR to complete the histogram piece ?

@aurelienpierre. RGB luminance from my standpoint is something like:
` 0.2126f * inl[0] + 0.7152f * inl[1] + 0.0722f * inl[2];  // luminance RGB`
I use that to provide the sample value for the L channel of LRGB, but it should be also necessary for the corresponding histogram.

aurelienpierre:
> @aurelienpierre. RGB luminance from my standpoint is something like:
> ` 0.2126f * inl[0] + 0.7152f * inl[1] + 0.0722f * inl[2]; // luminance RGB`
> I use that to provide the sample value for the L channel of LRGB, but it should be also necessary for the corresponding histogram.

Ok, that's exactly how we compute the Y channel of XYZ space.

TurboGit:
To resolve conflict:

```
$ git checkout land:tc-LCh2
$ git fetch
$ git rebase origin/master
```

Then, look at the code and resolve the conflict line by editing the corresponding files.

```
$ git add <fixed files>
$ git rebase --continue
```

And if all is fine, force push here:

`$ git push -f origin land:tc-LCh2`

edgardoh:
@phweyland , I don't know what you're planning to do, so I can't tell.

phweyland:
@TurboGit 
Thanks for the instructions. I did it (at least I hope). In that process I've seen conflicts with changes made by other people but not the in file I had changed.  And after final push I got several new conflicts (still other people changes). However when I do "git rebase origin/master" everything it says  "Current branch already up to date". I'm missing something ... :-(

TurboGit:
This is strange, what if you force push the rollabck commit:

`$ git push -f origin tc-LCh2`

If this does not work, something is quite messed up on your side. In that case you may want to start over master:

```
$ git branch -m tc-LCh2 tc-LCh2-old
$ git checkout tc-LCh2
$ git reset --hard origin/master
```

And get current sources from you current tc-LCh2 branch:

```
$ git show tc-LCh2-old:src/iop/tonecurve.c  > src/iop/tonecurve.c
$ git show tc-LCh2-old:data/kernel/basic.cl > data/kernel/basic.cl
```

And then commit this single commit for the current implementation:

`$ git commit -a -m "Add LCh curve mode to tonecurve"`

Double check that all is ok and then, force push to this branch:

`$ git push -f origin tc-LCh2`

Hope this will solve the mess :)

phweyland:
> This is strange, what if you force push the rollabck commit:
> `$ git push -f origin tc-LCh2`

Same answer with or without `-f`

> If this does not work, something is quite messed up on your side. In that case you may want to start over master:

I did the rest and except which didn't do anything (but I had the kept the files somewhere else):

> $ git show tc-LCh2-old:src/iop/tonecurve.c  > src/iop/tonecurve.c
> $ git show tc-LCh2-old:data/kernel/basic.cl > data/kernel/basic.cl

> Hope this will solve the mess :)

Yes ! Thank you spo much :-)

Just to check my normal workflow:
To update my local dt master:
`$ git fetch upstream`
`$ git merge upstream/master`
To create a new branch and update it on remote:
`$ git checkout -b <branch>`
`$ git commit -a`
`$ git push -u origin <branch>` first time then `$ git push origin <branch>`
Sometimes to update the local branch against local master:
`$ git merge master` but I've understood here I could use also `$ git rebase master`.
As an editor I use atom. Could the git command from atom introduce issues ?
Is there anything I should avoid ?

phweyland:
> 
> 
> > @aurelienpierre. RGB luminance from my standpoint is something like:
> > ` 0.2126f * inl[0] + 0.7152f * inl[1] + 0.0722f * inl[2]; // luminance RGB`
> > I use that to provide the sample value for the L channel of LRGB, but it should be also necessary for the corresponding histogram.
> 
> Ok, that's exactly how we compute the Y channel of XYZ space, so it's redundant.

I see some difference between XYZ and RGB on the image (dt 2.6 and 2.7) and on the color picker value (dt 2.7 with the same correction and this PR). I've not checked the matrix RGB<->XYZ transformation against the the above formula (I should, shouldn't I ? :-) ) but I'm afraid that's not quite the same.
Then if XYZ and RGB luminances were the same we should not keep both modules, correct ?

TurboGit:
You should avoid merging with upstream/master.

For a long lived topic branch like yours you should favor the rebase workflow.

`$ git rebase upstream/master`

This will avoid the merge conflicts or at least restrict them to the ones from your changes.

phweyland:
@edgardoh 
> I don't know what you're planning to do, so I can't tell.

Not sure understand this. I'm just proposing these two modes, LCh and (L)RGB. I've also tried to make easier the implementation of other future modes. Are still missing two items:
- histogram for new modes
- log histogram

As you have said displaying an histogram which is not related to the colorspace being used is weird. That's why I've removed them so far.
But the histograms are useful and we should provide them. I see two ways.
The first one is to wait for the selectable colorspace (your PR). In that case the histogram will be calculated automatically in the right colorspace by histogram.c (if I've" understood properly) and I'll just have to enable them where needed in tonecurve.
The second way would be to write some code to simulate in "tonecurve" the histogram calculation as you suggested. Is that correct ? But from my standpoint that would be useless work if your PR is coming.

phweyland:
> @phweyland , in my last message I give some options on how to display the histogram and how the curves can behave, what I don't know is which one (if any) you will choose. Will it have the same behaviour as my PR (shifting the 50% gray with an S curve)? Something different?

Ok, I've missed this one : `when the colorspace don't have a 50% grey an S curve will cause a shift on the middle grey , I don't think that this is correct.`.
I would not agree on this, I think this is correct. I'm happy with those you have produced. Whatever the colorspace (Lab or RGB), applying a manual S-curve (or whatever change) may change the grey point, ie, move it out of 18% or 50%.

The histogram which is shown is the module input histogram. There is no module output histogram. The general histogram shows the output of the output module if I'm right. It can be very different of the tonecurve module output one, especially the linear ones.
So maybe it could be interesting to show both the input and output histograms of tonecurve (overlapped drawing)...
Now the linear colorspaces show most of the points on the left handside of graph. That's where the log histogram helps the user to better control his changes on this part.

edgardoh:
> Ok, I've missed this one : when the colorspace don't have a 50% grey an S
> curve will cause a shift on the middle grey , I don't think that this is
> correct..
> I would not agree on this, I think this is correct. I'm happy with those
> you have produced. Whatever the colorspace (Lab or RGB), applying a manual
> S-curve (or whatever change) may change the grey point, ie, move it out of
> 18% or 50%.
>
Well, I don't find it intuitive and I don't think that's how other programs
behave, but not my call. But since this will change how the tone curve
behaves now, you should check with the dev team to see if they are OK with
this change.

> The histogram which is shown is the module input histogram. There is no
> module output histogram. The general histogram shows the output of the
> output module if I'm right. It can be very different of the tonecurve
> module output one, especially the linear ones.
> So maybe it could be interesting to show both the input and output
> histograms of tonecurve (overlapped drawing)...
> Now the linear colorspaces show most of the points on the left handside of
> graph. That's where the log histogram helps the user to better control his
> changes on this part.
>
My PR won't do any changes on the histogram and will handle only Lab and
RGB, anything else should be handled by the modules, so XYZ, prophotoRGB,
LCh histograms won't be automatically generated, in order to display it
some work is required.

phweyland:
>Well, I don't find it intuitive and I don't think that's how other programs behave, but not my call. But since this will change how the tone curve behaves now, you should check with the dev team to see if they are OK with this change.

Right.

aurelienpierre:
> > > @aurelienpierre. RGB luminance from my standpoint is something like:
> > > ` 0.2126f * inl[0] + 0.7152f * inl[1] + 0.0722f * inl[2]; // luminance RGB`
> > > I use that to provide the sample value for the L channel of LRGB, but it should be also necessary for the corresponding histogram.
> > 
> > 
> > Ok, that's exactly how we compute the Y channel of XYZ space, so it's redundant.
> 
> I see some difference between XYZ and RGB on the image (dt 2.6 and 2.7) and on the color picker value (dt 2.7 with the same correction and this PR). I've not checked the matrix RGB<->XYZ transformation against the the above formula (I should, shouldn't I ? :-) ) but I'm afraid that's not quite the same.
> Then if XYZ and RGB luminances were the same we should not keep both modules, correct ?

The coefficients you use for the [luma computation assume a REC709 RGB color space](https://en.wikipedia.org/wiki/Luma_(video)) (D65 illuminant) but you apply them in Prophoto RGB space (D50 illuminant), so the luma you compute is wrong in the first place. The RGB <-> XYZ matrices in darktable are right, and adjusted for a D50 illuminant, which is the standard inside dt's pipe.

Then, the XYZ mode of the tone curve is bugged, because the curve is applied on the 3 XYZ channels (which makes no sense from a color perspective), and shift the color toward green/yellow. The correct way to do it would be to use xyY color space, and affect only the Y channel, then convert back to XYZ. In xyY, x and y are the chromaticy coordinates at given luminance Y. In XYZ, Y reprents roughly the green response of the eye (which is equivalent to the relative luminance, since we are more sensible to green), X the ~ red and Z the ~ blue. But XYZ makes sense only to link radiometric coordinates (wavelengths and intensities) to human eye cone-response. It is a connexion space, not a physically-meaningful color space, so nobody should mess-up with it, except to adjust illuminants (Bradford and von Kries transforms) and switch between RGB spaces.

TL;DR don't mess up with RGB luminance and leave the xyY fix to me, I have it in another PR :slightly_smiling_face:

edgardoh:
I did some testing and RawTherapee and PhotoFlow don't shift the middle grey, RawTherapee works on prophotoRGB (as far as I know) and PhotoFlow allows to select the working colorspace. No idea internally how they handle the curve and the histogram, but the code and authors are there.

My main concern is that the meaning of the curve should remain the same, no matter the colorspace. And a standard S curve is not supposed to shift the middle grey, as shown by the current presets "contrast * (linear)"

I'm not saying that we should do exactly the same as other programs, if we find something better then great, but by changing colorspaces there should be a way of having the same curve with the same meaning.

Take for instance the contrast, changing the colorspace will shift the middle grey, so in my "basic adjustments" I have added a fulcrum so it can be corrected when changing from a linear to a non-linear RGB (is a WIP, so is far from perfect), maybe something similar can be done with the curves and histogram.

Maybe @TurboGit  has an opinion on this.

phweyland:
Looking at the way to add the waveform histogram I'm wondering where to put that code.
A lot drawing routines are put in draw.h, which surprises me a bit because the code distributed in every file referencing it, using it or not. Shouldn't we setup a draw.c file ?

phweyland:
>  And a standard S curve is not supposed to shift the middle grey, as shown by the current presets "contrast * (linear)"

Yes because this particular preset sets a node on (x,y) = (50,50) given in Lab color space, which is weird because in the same time the preset sets the tonecurve mode to RGB ... 
The other presets, which do not set a node on middle grey (50,50) (or (18,18) depending on the color space we use), do not preserve the middle grey.

But I agree we need a consistent method to display the histogram. 
Today, the pipe works on Lab and the main histogram is displayed on monitor RGB, the tonecurve one in Lab ...

On one hand we could say that the tone curve histogram should follow the main one (monitor RGB).
But on the other hand the tabs a & b for Lab , C & h for LCh, have nothing to do with RGB.

Currently the tonecurve mode selects the working color space (on which the curve is applied).
We could let the user choose also the tonecurve histogram color space, whatever the working color space.

I see now that only one preset works with my change, the Lab one (Nikon 750). The others pretend to use RGB but provide Lab nodes...

Yes, others standpoint would be very welcome.

TurboGit:
No, sorry no opinion on this. This is far from my expertise at this point!

aurelienpierre:
Ok you need to separate the [view and the data](https://en.wikipedia.org/wiki/View_model).

No matter the color space, the curve sets the middle grey at 50 % of the graph view. In `commit_params()`, the coordinates of the curve's LUT are then properly remapped in 1D to the target color space, meaning 50 % -> 18.45 % in RGB and XYZ modes:

https://github.com/darktable-org/darktable/blob/5b7b2f1b6abc7a52c98cc7f35d075fab24909cf8/src/iop/tonecurve.c#L583-L595

(But @phweyland, you have removed that part.)

Then, in `process()` the LUT mapping happens in the right color space.

So I think we are spinning around a non-issue. It's not a "Lab" or "RGB" curve, it's a curve drawn in a space where the middle grey is 50 % so the graph is centered (which happens to be Lab, but we don't care in that particular case. Could have been a simple ^2.35 too).

edgardoh:
This is what I'm saying, but we still don't know what to do with the
histograms. Do we transform to some gamma corrected colorspace? Apply some
gamma function?

phweyland:
> @aurelienpierre 
(But @phweyland, you have removed that part.)

OK, I understand the purpose of this now. The current full tonecurve module UI (histogram, color picker, presets and saved values) is Lab based while some LUT operations are made on other color spaces (XYZ and RGB). Displaying the Lab luminance for RGB or XYZ is not really an issue (while counter intuitive from my stand point). I'll revert my changes on this (at least to keep the downward compatibility).

We could do the same (translate presets to Lab) for LCh and LRGB but some channels (C(L), C(h), R, G B) have nothing to do with Lab luminance. For these channels the Lab luminance is useless. We need to present to the user a meaningful histogram with the associated color picker and presets values.
Remains, as @edgardoh pinpoints, to choose the correct RGB color space to be displayed (histogram, color picker, presets). Linear RGB (grey at 18%), the one that @edgardoh doesn't like too much ? the sRGB one as on the main histogram (not sure that grey is at 50%) ?
sRGB would have the advantage to be homogeneus with the main histogram (linear, log and waveform when available).

aurelienpierre:
Although having a Lab histogram is not super intuitive, it is consistent with the rest of the UI. For chroma-based channels, you can leave that linear, but you will need a special histogram. If you want an RGB histogram, you can convert each channel to Lab as if they were grey to build the histogram like that:

```
float fake_R[3] = { RGB[0], RGB[0], RGB[0] }; // red × 3
float fake_G[3] = { RGB[1], RGB[1], RGB[1] }; // green × 3
float fake_B[3] = { RGB[2], RGB[2], RGB[2] }; // blue × 3
float ui_R[3], ui_G[3], ui_B[3];
// convert colors to pseudo-Lab for UI
// only the [0] element of the vector is relevant (pseudo-L)
rgb_to_lab(fake_R, ui_R);
rgb_to_lab(fake_G, ui_G);
rgb_to_lab(fake_B, ui_B);
float ui_RGB[3] = { ui_R[0], ui_G[0], ui_B[0] }; // RGB corrected in UI space
```

But of course, it's not super efficient from a performance point of view.

aurelienpierre:
Also, sRGB is crap, because it has 2 different gamma depending the range of the input, so you have to branch stuff and the code is not vectorizable. If this is the path you want to take, just use a ^(1/2.45)

phweyland:
> Although having a Lab histogram is not super intuitive, it is consistent with the rest of the UI.

I'll make a try with this philosophy. Thanks for the piece of code. That enlightens the way the histogram is built.

phweyland:
I've made some progress with LRGB with Lab mode histogram.
But I've 2 issues.
1. The only place I've found in the module to get the input image is in process(). It's probably not the right place. Is there any other optional callback I could use instead ? When the callback doesn't provide the input buffer, is there a way to get from dt_iop_module_t ?
2. The histogram is not stable. Moving around a node of the curve make the histogram tremble. Probably because dt_iop_tonecurve_draw() runs in parallel (without exclusion).
My preference would be to build the histogram in dt_iop_tonecurve_draw() instead of process() but I haven't found the way to access the input image from there. Any idea ?

phweyland:
I've updated LRGB with Lab UI (presets, color picker, histogram), but open_cl.
That doesn't simplify the code but as @aurelienpierre said the UI is consistent with the rest of the module.
However, locally built histograms are still dancing...

phweyland:
@edgardoh 
As @aurelienpierre suggested I find the `Lab like` histogram quite effective for R, G & B channels.
The problem is that the module is not the right place to make these calculations. I cannot get access to input buffer when executing `dt_iop_tonecurve_draw()`. On the other hand I cannot control the exclusion between `process()` and `dt_iop_tonecurve_draw()`.
If I understand properly there are already histogram calculations for each module happening inside the pixelpipe execution. That would be the right place to do it, wouldn't be ? But we need to make the module able to choose the needed histogram. In some way that's what you are doing for the module's input/output colorspaces, but for the UI colorspace. What do you think ?

phweyland:
@turbogit, I think the current state is ok to be reviewed.
Both LCh and LRGB independent modes are added, showing histograms in LAB to be coherent with others modes. Open cl is also implemented.

aurelienpierre:
So what you call LRGB is the same curve applied on all RGB channels ? 

Nice job for integrating the separate (manual) RGB channels ! I was planning to do it at some point, that's definitely helpful. I will have a look at the color spaces conversions in the next few days.

phweyland:
> So what you call LRGB is the same curve applied on all RGB channels ?

In this mode, `RGB independent channels`, L is equivalent to `RGB linked channels`. R, G, B independent curves are applied just after.
NB: I'm making some tests adding to `RGB linked channels` the `weighted yellow power norm` described in the link you have provided on darktable-dev message `Color management in HDR setups`. I find the results interesting (not yet in that PR).
In fact we could propose several of these norms which control in some the behavior of the S curve.

@TurboGit. The build fails for some reason I don't understand. Any advice ?

aurelienpierre:
>  In fact we could propose several of these norms which control in some the behavior of the S curve.

These fall into chrominance preservation modes. In filmic, the only mode used for now is the `max(RGB)`, hence the `L infinite` norm. I have thought about making it more general and using other norms as well, so that could end up in a general API:

```
inline float norm_vec(float RGB[4], float norm)
{ 
  // ensure (norm < 0.0f) in GUI controls for better perf

  // norm L infinite = max
  if(norm == 0.0f) return fmaxf(RGB[0], fmaxf(RGB[1], RGB[2]); 

  // norm L1
  else if(norm == 1.0f) return RGB[0] + RGB[1] + RGB[2];

  // norm L2 = euclidian norm
  else if(norm == 2.0f) return sqrtf( RGB[0] * RGB[0] + RGB[1] * RGB[1] + RGB[2] * RGB[2]);

  // general Lp norm (pseudo-norm if p < 1)
  else 
  {
    if((RGB[0] < 0.0f || RGB[1] < 0.0f || RGB[2] < 0.0f) && norm < 1.0f) return -1;
    return powf( powf(RGB[0], norm) + powf(RGB[1], norm) + powf(RGB[2], norm), 1.0f/norm); 
  }
}
```

phweyland:
> so that could end up in a general API:

Interesting. In the article you've indicated the 2 norms are:
Basic power norm: (R^3+G^3+B^3)/(R^2+G^2+B^2)
Weighted yellow power norm: 0.83743219(1.22R^5+1.20G^5+0.58B^5)/(1.22R^4+1.20G^4+058B^4)

Is the routine you've given above already merged ? Of course we could put together all valuable ones.
Comment: as a lut is applied on the output, it should be scaled to [0,1], right ?

TurboGit:
@phweyland :
The issue is with clang only in the external rawspeed submodule.

I have noticed that you have changed the rawspeed submodule here. It should not be done, please make sure the submodule is not touched. Note that this is not the issue anyway, as current master fails with clang.

phweyland:
> It should not be done, please make sure the submodule is not touched.

OK. If I understand I should not execute  `git submodule update`, correct ?
Except maybe at each dt new release ? Otherwise how to know when I should do it ?

TurboGit:
No, before commit you should always run (to be sure) "git submodule update". This will ensure that the module is set to the proper history before the commit. In this way there is "no change" for this module in your history and then the commit won't contains it.

The issue you have is probably that you have moved to master, did a submodule update (or was done as part of the git pull maybe?) and then back to your branch. Then the module was not back to the point it was on your branch, and so some modifications (change from master) have been committed.

phweyland:
@aurelienpierre 
I've add the 2 norms I had plus yours to RGB linked channels. Please review the corresponding formulas. Of course we could remove the less relevant ones, but globally that seems promising.

aurelienpierre:
> > so that could end up in a general API:
> 
> Interesting. In the article you've indicated the 2 norms are:
> Basic power norm: (R^3+G^3+B^3)/(R^2+G^2+B^2)
> Weighted yellow power norm: 0.83743219(1.22R^5+1.20G^5+0.58B^5)/(1.22R^4+1.20G^4+058B^4)
> 
> Is the routine you've given above already merged ? Of course we could put together all valuable ones.
> Comment: as a lut is applied on the output, it should be scaled to [0,1], right ?

The routine above is just a project for now, nothing done yet. The L1, L2, Lp norms are algebraic vector norms ("real" vector maths), but the power and yellow weighted norms are only empiric ad-hoc formulas, so I'm very suspicious with that.

As for the LUT, theoritically yes, but the curve processing has an exponential interpolation branching for values above 1, and below 0 in some cases (which is idiotic if you ask me, but here we are…).

data/kernels/basic.cl
```diff
@@ -614,11 +614,54 @@ colorin_clipping (read_only image2d_t in, write_only image2d_t out, const int wi
   write_imagef (out, (int2)(x, y), pixel);
 }
 
+float
+rgb_norm_vect (float4 rgb, int rgb_norm, float norm_exp)
+{
+  switch(rgb_norm)
```
aurelienpierre: not sure about `case 2` and `case 3` since they are particular cases of `case 4` (with values of `norm_exp` nice enough to avoid slow calls to `pow`). I would just merge them into `case 4` and check the `norm_exp` to branch, or even completely ignore them in OpenCL since branches can make it super slow and it doesn't mind the load (so you just use the general case 4).

data/kernels/basic.cl
```diff
@@ -677,27 +720,24 @@ tonecurve (read_only image2d_t in, write_only image2d_t out, const int width, co
...
-    rgb.x = lookup_unbounded(table_0, rgb.x, coeffs_0);
-    rgb.y = lookup_unbounded(table_0, rgb.y, coeffs_0);
-    rgb.z = lookup_unbounded(table_0, rgb.z, coeffs_0);
-    pixel.xyz = prophotorgb_to_Lab(rgb).xyz;
+    float norm = 0;
+    if (rgb_norm != 0)
+    {
+      norm = rgb_norm_vect( rgb, rgb_norm, 0.333333f);
+      norm = lookup_unbounded(table_0, norm, coeffs_0) / norm;
+      rgb.xyz *= norm;
```
aurelienpierre: you got that wrong :-)

```C
norm = rgb_norm_vect( rgb, rgb_norm, 0.333333f); // compute the norm == luminance estimator
const float4 rgb_ratios = rgb /norm; // these ratios are the actual colors, independant from the luminance
norm = lookup_unbounded(table_0, norm, coeffs_0); // compute the curve on the luminance
rgb.xyz = (norm * rgb_ratios).xyz; // restore the colors from the original ratios and the new luminance
```

data/kernels/basic.cl
```diff
@@ -614,11 +614,54 @@ colorin_clipping (read_only image2d_t in, write_only image2d_t out, const int wi
...
+{
+  switch(rgb_norm)
+  {
+  case 1:  // norm L infinite = max
+    {
+      return max(rgb.x, max(rgb.y, rgb.z));
+    }
+  case 2:  // norm L1 - bypass the powf for performance
+    {
+      return (rgb.x + rgb.y + rgb.z) / 3;
```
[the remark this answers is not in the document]
phweyland: > this is not a norm, but a mean. You should not divide by 3

I've done but doesn't work.
As an equivalent to luminance it cannot be equal to 3 ...
Here 3 is not for average, but the max of the 3 values: rgb.x max + rgb.t max + rgb.z max.
How else to keep this inside [0,1] ? 
Same issue for the other cases.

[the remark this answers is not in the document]
aurelienpierre: So if you really need to divide, at least divide by `3^norm` to stay consistent.

data/kernels/basic.cl
phweyland: I understand this (keep the true color) with norm calculation. 
But in case of norms from your referred article, the goal is to fix some perceptual artefacts at the toe and shoulder of S curve. So the calculation I’ve proposed may be more appropriate in that case. In any case I’ll make some experiment for those.

phweyland: I understand this (keep the true color) with norm calculation. 
But in case of norms from your referred article, the goal is to fix some perceptual artefacts at the toe and shoulder of S curve. So the calculation I’ve proposed may be more appropriate in that case. In any case I’ll make some experiment for those.

phweyland:
In fact I’m afraid our calculations are exactly the same :-)
Rgb = rgb * lut(norm) / norm

aurelienpierre:
> In fact I’m afraid our calculations are exactly the same :-)
> Rgb = rgb * lut(norm) / norm

ah. Indeed. So it is mostly a problem of readability, when you overwrite the variable `norm` again and again, it's easy to lose if we are referring to the new or the old. So I just split steps in the most stupid simple way, and the code becomes easier to read.

data/kernels/basic.cl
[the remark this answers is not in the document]
phweyland: It works.
But while, for a given curve, L infinite or weighted yellow power do not change the image luminance, Lp has a strong effect on the contrast of the image for values below 2 (I've set a slider for norm_exp).
On the other hand above 3 I barely see any effect.
Is that what you expect ?

For norm_exp = 2, we have the length of the vector, I understand that. 
But what is the meaning of norm_exp = 1 ? 

What should be the range of norm_exp ?

data/kernels/basic.cl
```diff
@@ -614,11 +614,54 @@ colorin_clipping (read_only image2d_t in, write_only image2d_t out, const int wi
...
+    }
+  case 3:  // norm L2 = euclidian norm - bypass the powf for performance
+    {
+      return sqrt((rgb.x * rgb.x + rgb.y * rgb.y + rgb.z * rgb.z) / 3);
```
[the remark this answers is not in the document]
phweyland: Done

data/kernels/basic.cl
```diff
@@ -614,11 +614,54 @@ colorin_clipping (read_only image2d_t in, write_only image2d_t out, const int wi
...
+      B = rgb.z * rgb.z;
+      return (rgb.x * R + rgb.y * G + rgb.z * B) / (R + G + B);
+    }
+  case 6: // weighted yellow power norm
+    {
+      float R, G, B;
+      R = 1.22f * rgb.x * 1.22f * rgb.x;
+      G = 1.20f * rgb.y * 1.20f * rgb.y;
+      B = 0.58f * rgb.z * 0.58f * rgb.z;
+      R *= R; G *= G; B *= B;
```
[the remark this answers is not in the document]
phweyland: Yes the formula is the one from the article. Done.

data/kernels/basic.cl
```diff
@@ -614,11 +614,54 @@ colorin_clipping (read_only image2d_t in, write_only image2d_t out, const int wi
...
+    {
+      return (rgb.x + rgb.y + rgb.z) / 3;
+    }
+  case 3:  // norm L2 = euclidian norm - bypass the powf for performance
+    {
+      return sqrt((rgb.x * rgb.x + rgb.y * rgb.y + rgb.z * rgb.z) / 3);
+    }
+  case 4:  // general Lp norm (pseudo-norm if p < 1) - slow variant
+    {
+      return pow((pow(rgb.x, norm_exp) + pow(rgb.y, norm_exp) + pow(rgb.z, norm_exp)) / 3, 1.0f/norm_exp);
```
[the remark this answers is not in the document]
phweyland: test on negative values back and fall back on calculation without norm.

data/kernels/basic.cl
phweyland: removed case 2 and 3. In tonecurve (openmp) detection of exp =1 or 2 and use of simplified formulas.

data/kernels/basic.cl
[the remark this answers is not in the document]
aurelienpierre: Ok. Also, be carefull of the RGB space the formula expects, otherwise the coefficients will be wrong (I suspect they use ACES P0).

data/kernels/basic.cl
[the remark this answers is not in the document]
aurelienpierre: Lp norms can be different things. I have seen them used in conjunction with gradient vectors, to isolate noise for example. Maybe we could stick to p = [2 ; 3] for RGB vectors then. I would have to see what they produce to figure a possible use.

phweyland:
@aurelienpierre 
I've applied the suggested changes.
If you have the opportunity to build it, I've set open cl routine without normalization while I've inserted the per 3 division on openmp routine. So you can see the difference by enabling / disabling open cl.

>the power and yellow weighted norms are only empiric ad-hoc formulas, so I'm very suspicious with that.

I thought you were advertising this ... :-).

phweyland:
@edgardoh
So far I haven't made nothing specific to converge towards your PR. Is there anything I should look at or do to facilitate the integration with your PR ?

edgardoh:
Backwards compatibility will be guarantee, so there's nothing for you to
worry about.

TurboGit:
@phweyland : you still need to remove rawspeed from the commit here. this will certainly makes testing easier. thanks.

TurboGit:
Also on your branch remove all the merge commits. You can do that by rebasing your branch:

> git rebase origin/master

aurelienpierre:
> I thought you were advertising this ... :-).

No, I shared the video because it enlightens the core problem of pixel processing in a color-managed way, an issue that few FOSS developers are aware of (proof being nothing in dt tries to address it). But to me, every equation that does not derive from a vector field integral looks like fake science.

phweyland:
@TurboGit I'm sorry. I think I don't as I should. 

> $ git rebase origin/master
> Current branch tc-LCh2 is up to date.

Then I did try this:

> Philippe@DESKTOP-JQ7D8F9 MINGW64 ~/dt-phi/build
> $ git fetch upstream
> remote: Enumerating objects: 899, done.
> remote: Counting objects: 100% (899/899), done.
> remote: Compressing objects: 100% (11/11), done.
> remote: Total 1362 (delta 888), reused 899 (delta 888), pack-reused 463
> Receiving objects: 100% (1362/1362), 1.48 MiB | 1.33 MiB/s, done.
> Resolving deltas: 100% (1128/1128), completed with 225 local objects.
> From https://github.com/darktable-org/darktable
>    a9ce955a7..a3f7a6d24  darktable-2.6.x -> upstream/darktable-2.6.x
>  * [new branch]          gphoto          -> upstream/gphoto
>  * [new branch]          lighttable-undo -> upstream/lighttable-undo
>    3fedabc25..6b51ef1b8  master          -> upstream/master
>  * [new branch]          po-ui-refactor  -> upstream/po-ui-refactor
> 
> Philippe@DESKTOP-JQ7D8F9 MINGW64 ~/dt-phi/build
> $ git submodule update
> 
> Philippe@DESKTOP-JQ7D8F9 MINGW64 ~/dt-phi/build
> $ git rebase upstream/master
> First, rewinding head to replay your work on top of it...
> Fast-forwarded master to upstream/master.
> 
> Philippe@DESKTOP-JQ7D8F9 MINGW64 ~/dt-phi/build
> $ git checkout tc-LCh2
> M       src/external/rawspeed
> Switched to branch 'tc-LCh2'
> Your branch is up to date with 'origin/tc-LCh2'.
> 
> Philippe@DESKTOP-JQ7D8F9 MINGW64 ~/dt-phi/build
> $ git rebase master
> First, rewinding head to replay your work on top of it...
> Applying: restrict metadata to lightable call
> Using index info to reconstruct a base tree...
> M       src/develop/lightroom.c
> Falling back to patching base and 3-way merge...
> No changes -- Patch already applied.
> Applying: Add LCh and RGB independent curve modes to tonecurve
> Applying: Fixed log histogram
> Applying: Restore automatic XYZ and RGB Lab LUT conversion
> Applying: LRGB UI changed to lab; Added RGB independent histogram
> Applying: Fixed the instability of R, G & B histograms
> Applying: Added Open cl
> Applying: Added rgb_norm combo to RGB linked channels
> Applying: improvements from review
> 
> Philippe@DESKTOP-JQ7D8F9 MINGW64 ~/dt-phi/build
> $ git submodule update
> Submodule path '../src/external/rawspeed': checked out '0b91a7447ab33994dbb7dc9c249793218a2a4e00'
> 
> Philippe@DESKTOP-JQ7D8F9 MINGW64 ~/dt-phi/build
> $ git commit -a -m "rebase and fix rawspeed update issue"
> On branch tc-LCh2
> Your branch and 'origin/tc-LCh2' have diverged,
> and have 97 and 24 different commits each, respectively.
>   (use "git pull" to merge the remote branch into yours)
> 
> nothing to commit, working tree clean
> 
> Philippe@DESKTOP-JQ7D8F9 MINGW64 ~/dt-phi/build
> $ git push origin tc-LCh2
> Username for 'https://github.com':
> Password for 'https://phweyland@github.com':
> To https://github.com/phweyland/darktable.git
>  ! [rejected]            tc-LCh2 -> tc-LCh2 (non-fast-forward)
> error: failed to push some refs to 'https://github.com/phweyland/darktable.git'
> hint: Updates were rejected because the tip of your current branch is behind
> hint: its remote counterpart. Integrate the remote changes (e.g.
> hint: 'git pull ...') before pushing again.
> hint: See the 'Note about fast-forwards' in 'git push --help' for details.

I tried to rebase my local branch master on upstream, then to rebase the branch tc-LCh2 on local master.
Doesn't work neither.
Do I use correctly upsteam/master to my local branch master ?

phweyland:
I've made:
$ git submodule update (and got a new commit #)
$ git commit -a -m "fix rawspeed update issue" (tells me there is nothing to commit)
$ git push origin tc-LCh2 --force (which worked this time)

and rawspeed is still there as a changed file... (but commit history seems to be clean now)

TurboGit:
Ok, I did the clean-up. Please do a git fetch origin to get all branches from the official repository. And in your branch do:

`$ git reset --hard po/tc-LCh2
`
Then if all is well just force push your branch here.

phweyland:
`git fetch origin ` doesn't do anything.
so I've done `git fetch upstream` (which brought the branch po/tc-LCh2)
and then done `git reset --hard upstream/po/tc-LCh2` on my local tc-LCh2 branch.
I still see this rawspeed modified thing ....
At this stage should I git push origin tc-LCh2 ? (should not work with unstaged rawspeed change)

> Philippe@DESKTOP-JQ7D8F9 MINGW64 ~/dt-phi/build
> $ git checkout master
> M       src/external/rawspeed
> Switched to branch 'master'
> Your branch and 'origin/master' have diverged,
> and have 89 and 16 different commits each, respectively.
>   (use "git pull" to merge the remote branch into yours)
> 
> Philippe@DESKTOP-JQ7D8F9 MINGW64 ~/dt-phi/build
> $ git fetch origin
> 
> Philippe@DESKTOP-JQ7D8F9 MINGW64 ~/dt-phi/build
> $ git fetch upstream
> remote: Enumerating objects: 16, done.
> remote: Counting objects: 100% (16/16), done.
> remote: Total 24 (delta 16), reused 16 (delta 16), pack-reused 8
> Unpacking objects: 100% (24/24), done.
> From https://github.com/darktable-org/darktable
>  * [new branch]          po/tc-LCh2 -> upstream/po/tc-LCh2
> 
> Philippe@DESKTOP-JQ7D8F9 MINGW64 ~/dt-phi/build
> $ git checkout tc-LCh2
> M       src/external/rawspeed
> Switched to branch 'tc-LCh2'
> Your branch is up to date with 'origin/tc-LCh2'.
> 
> Philippe@DESKTOP-JQ7D8F9 MINGW64 ~/dt-phi/build
> $ git reset --hard upstream/po/tc-LCh2
> HEAD is now at 4f7c23756 improvements from review
> 
> Philippe@DESKTOP-JQ7D8F9 MINGW64 ~/dt-phi/build
> $ git status
> On branch tc-LCh2
> Your branch and 'origin/tc-LCh2' have diverged,
> and have 8 and 8 different commits each, respectively.
>   (use "git pull" to merge the remote branch into yours)
> 
> Changes not staged for commit:
>   (use "git add <file>..." to update what will be committed)
>   (use "git checkout -- <file>..." to discard changes in working directory)
> 
>         modified:   ../src/external/rawspeed (new commits)
> 
> no changes added to commit (use "git add" and/or "git commit -a")

TurboGit:
Great no more rawspeed in sight :)

TurboGit:
@phweyland : what is the status? do you fell this to be done? should it be considered for review/merging at this stage? thanks.

phweyland:
@TurboGit, I'm not an expert but I think that's all valuable. I'm done with what I wanted to do.
For me it's time for review and test by others. I would like to have their view. Useful or not ? Improvements ?
Maybe two questions in particular:
- RGB norms: we have the technical ones as @aurelienpierre suggested. What about some other experimental ones ? 
- Histogram implementation inside the module instead of the core histogram modules. Is that correct ?
Thanks

phweyland:
Having seen the tutorial 10 on https://3dlutcreator.com/3d-lut-creator---tutorials.html I thought it could be more useful to have these tabs on LCh mode: L, C(C), C(L) & L(C). 
And if this make sense, it would be also interesting to be able to switch of colorspace with the same tabs.

There is now some conflicts with tonecurve.c ... I understand we don't merge something which is not 100% finished. But on the other hand there is no way to get user feedback if we don't merge somewhere in a kind of sandbox or quality space for testing. How will things go forward ?

edgardoh:
>
> There is now some conflicts with tonecurve.c
>
There will be more, I'm making some more changes to it. They all are
related to the color picker so it should be easy to resolve.

> ... I understand we don't merge something which is not 100% finished. But
> on the other hand there is no way to get user feedback if we don't merge
> somewhere in a kind of sandbox or quality space for testing. How will
> things go forward ?
>
This is an issue for large developments. Ideally I would like to have an
intermediate branch 'dev' where things are merged when they are stable
enough, and have regular builds for all systems so users that can't build
are able to test it. When OK it will be merged to master.
No idea how to implement it, maybe github has some feature that make it
easy?

TurboGit:
To make progress on this at this stage we need first to rebase on top of master to resolve conflicts. Please a rebase, not merging master into this branch. Then I can start testing and do some review on the code. I've seen that Aurélien and Edgardoh have already looked at the math in this code which is reassuring.

phweyland:
I will do.
I'll remove also the changes I've made of RGB linked channels and RGB independent channels as @edgardoh has introduced rgbcurve.The main differences I can see:
- on RGB linked channels some norms (could be moved to rgbcurve "preserve colors")
- L channel along with RGB independent channels. Of course using 2 instances of rgbcurve we can do the same (and that may be better).

Will remain the LCh mode. I think we should get informed people critic / advice before pushing this in prod.

TurboGit:
I lost track of the status of this PR, @phweyland can you summarize? Thanks.

phweyland:
> can you summarize?

The module rgb_curve does the same as  this PR mode "RGB independent channels". However with some differences:
- I've 4 channels (Luminance, R, G, B) instead of 3 for rgb curve (R, G, B). 
- I'm given different norms to recover the color: Max(L), Average(L), power, ... while rgb curve gives the RGB luminance.

The other new mode is "LCh independent channels" which gives the channels: Luminance(Luminance), Chrominance(Chrominance), Chrominance(Luminance) and Luminance(Chrominance).
The background data are the same as Lab but with some easier ways to tweak Chrominance than a & b of Lab. These modes are available in RT.

My feeling.
RGB independent channels is not any more an added value since rgb curve module. I just would like to open a PR to add different norms to rbg curve and basic adjustments (preserve colors parameter) in the future.
Implement LCh represents some changes in tonecurve module structure (moving from 3 to 4 channels) without saying I have to adapt the code to the new pixelpipe (color spaces). Unless some people want this function I would not do it.
Without other input I would close that PR.

Nilvus:
@phweyland : I think that having modules doing same things is not the best. Ok, we can hide modules but if the list is too long, users will have to difficulties to find which one to use. So, as the new rgb curve is here, I'm agree for rgb chanel to improve rgb curve (adding luminance setting for example or your other suggestions) instead of tone curve.

For tone curve, just improving what's existing or correct issues would be sufficient I think. So I'm agree on your last conclusion. I don't use tone curve now, as I find rgb curve more suitable (and that's also important, easier to use as is more readable). See this issue for what I'm saying on it : #2402. Curve and histogram on rgb curve are really more readable than on tone curve (and it's not editable with css).

3 issues actually are related to tone curve. The two others are : #2506 and #2317

phweyland:
> For tone curve, just improving what's existing or correct issues would be sufficient I think

I think we should remove the "rgb linked channels" from tonecurve and, if possible, have an automatic fallback on rgb curve. @edgardoh, do you think that would be possible ?
The Lab part of tonercurve still makes sense for Lab work. I agree this part could be cleaned-up to make profit of last general dt improvements.

edgardoh:
> For tone curve, just improving what's existing or correct issues would be
> sufficient I think
>
> I think we should remove the "rgb linked channels" from tonecurve and, if
> possible, have an automatic fallback on rgb curve. @edgardoh
> <https://github.com/edgardoh>, do you think that would be possible ?
>
I don't think that's possible.

The Lab part of tonercurve still makes sense for Lab work. I agree this

TurboGit:
@phweyland : given your last comment "Without other input I would close that PR." and others reactions I'll close this PR indeed. Thanks.

PR #2019, opened 2019-01-20
data/noiseprofiles.schema: reduce minimum ISO check

hknaack:
31 seems to be a pretty arbitrary number, CHDK allowed me to take
pictures at ISO 10. Since there are no restrictions in the code, let's
just go with the lowest sane value: 1.

Signed-off-by: Hartmut Knaack <knaack.h@gmx.de>

### DIFF
--- a/data/noiseprofiles.schema
+++ b/data/noiseprofiles.schema
@@ -66,7 +66,7 @@
                       },
                       "iso": {
                         "type": "integer",
-                        "minimum": 31,
+                        "minimum": 1,
                         "maximum": 2048000
                       },
                       "a": {


TurboGit:
IIRC, the parsing does warn if a ISO value is outside of the range. So maybe 1 just remove all checks? Maybe go with 10 if there is a camera with such ISO?

hknaack:
The article on wikipedia [1] indicates, that the lowest ISO value for traditional film has been ISO 0.8, and the conversion formula indicates 0 to be the lowest value.
So what are the pros and cons for:

10 as lowest value: works with my camera (+), doesn't cause problems with existing code (probably +), check for sane values (+), additional work if someone has got a camera with lower ISO (-)

1 as lowest value: works with my camera (+), doesn't cause problems with existing code (probably +), check for sane values (+), most likely nobody will come up with a camera with lower ISO (+)

dropping this check: works with my camera (+), doesn't cause problems with existing code (probably +), no check for sane values (-), no trouble with new cameras with whatever lowest ISO (+)

So, from my perspective, using 1 would be the best choice in the long term.

[1] https://en.wikipedia.org/wiki/Film_speed

TurboGit:
I'm not convinced (loosing check for bad values), but well that a quite minor issue. Thanks.

Issue #2011, opened 2019-01-18
[UI usability] [WIP] masks selection from icons in 1-click


flowernert:
**Is your feature request related to a problem? Please describe.**
I make use of masks more than often, others probably do so, and it always bugs me that to access one of the few masking options I have to : 
- click on the combobox
- hover to my selected mask
- click on it

to be able to select my mask. 
To me, masks are a feature important enough for it to be a 1-click action to select the mask type and unfold its options.

**Describe the solution you'd like**
I would like to see an horizontal box with icons to select the mask type and allowing to access it in 1 click replacing the cumbersome combobox.

Screenshot of what I did after hacking a bit in the code : 

**basic view**
![image](https://user-images.githubusercontent.com/2923696/51400819-a1a4b780-1b49-11e9-9650-2d83e4c335f6.png)

**tooltips** on hover are preserved
![image](https://user-images.githubusercontent.com/2923696/51401310-b766ac80-1b4a-11e9-80f9-f58b867b3b27.png)

**toggling:** when one mask option is clicked, the mask becomes active and the box is toggled
![image](https://user-images.githubusercontent.com/2923696/51400982-04964e80-1b4a-11e9-933e-d7b70961ac70.png)

**disabling:** mask can be disabled by clicking the cross on the left
**1** hovering the cross 
![image](https://user-images.githubusercontent.com/2923696/51401258-98681a80-1b4a-11e9-8793-3c851eab989c.png)

**2** after clicking the cross
![image](https://user-images.githubusercontent.com/2923696/51401193-78385b80-1b4a-11e9-948a-8cda827325b4.png)

**Additional context**
I've taken caution of preserving the following functionalities : 
- in the darkroom, when the displayed image is updated (by film roll or key press), the toggled mask icon is updated appropriately to the updated image modules settings.
- only one mask type can be enabled at any time
and add this spec:
- the cross doesn't stay toggled once clicked as it is rather an un-toggling button

Currently I've implemented the proposed feature in an additive way (apart from not showing the combobox, every line of code that is related to it is still present). I'm looking forward to get rid of it without breaking anything yet, and to reuse at most the existing code and structures in develop/blend.h develop/blend.c and develop/blend_gui.c

I'll submit a PR once I'm completed.

Let me know what you think of the feature I propose :)

TurboGit:
Sounds like a nice improvement to me!

flowernert:
Thanks @turbogit

[...]

PR #2020, opened 2019-01-20
[UI usability] [WIP] masks selection from icons in 1-click 

flowernert:
Following [feature request #2011](https://github.com/darktable-org/darktable/issues/2011)

I tried to implement on a minimal change basis in order to avoid breaking something than works (especially the callback for the removed combobox) and to respect the logic of already existing code.

The new glist introduced in develop/blend.h is freed in dt_iop_gui_cleanup_blending (like all the already existing ones).

My manual testing consisted of :
- changing masks from one to another, everything gui element displaying fine 
- disable current mask by clicking on cross or untoggling the current mask
- having previous masks state properly restored when selecting another image in the film roll

### DIFF
--- a/src/develop/blend.h
+++ b/src/develop/blend.h
@@ -387,7 +387,6 @@ typedef struct dt_iop_gui_blend_data_t
   GList *blend_modes;
   GList *masks_modes;
   GList *masks_modes_toggles;
-	GtkWidget *selected_mask_mode;//FLO
   GList *masks_combine;
   GList *masks_invert;
   GList *masks_feathering_guide;
@@ -395,7 +394,7 @@ typedef struct dt_iop_gui_blend_data_t
   GtkWidget *iopw;
   GtkBox *top_box;
   GtkBox *bottom_box;
-  GtkBox *masks_modes_box;//FLO
+  GtkBox *masks_modes_box;
   GtkBox *blendif_box;
   GtkBox *masks_box;
   GtkBox *raster_box;
@@ -405,6 +404,7 @@ typedef struct dt_iop_gui_blend_data_t
   GtkLabel *lower_label[8];
   GtkLabel *upper_picker_label;
   GtkLabel *lower_picker_label;
+	GtkWidget *selected_mask_mode;
   GtkWidget *upper_polarity;
   GtkWidget *lower_polarity;
   GtkWidget *colorpicker;
--- a/src/develop/blend.h
+++ b/src/develop/blend.h
@@ -362,6 +362,7 @@ typedef struct dt_iop_blend_mode_t
   unsigned int mode;
 } dt_iop_blend_mode_t;
 
+
 /** blend gui data */
 typedef struct dt_iop_gui_blend_data_t
 {
--- a/src/develop/blend.h
+++ b/src/develop/blend.h
@@ -393,7 +393,7 @@ typedef struct dt_iop_gui_blend_data_t
   GtkLabel *lower_label[8];
   GtkLabel *upper_picker_label;
   GtkLabel *lower_picker_label;
-	GtkWidget *selected_mask_mode;
+  GtkWidget *selected_mask_mode;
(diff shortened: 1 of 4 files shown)


flowernert:
I've set up a desktop launcher for my custom darktable so I didn't even noticed the warnings! Good thing you told me, I'll inspect to fix them. 

For the icons I'm not sure the circle is representative or feels intuitive as a uniform mask, maybe a partial circle (like 300°) with progressive gradient would be better, kind of figuring an analog blending rotating button.

For the pencil and the raster mask it seems to figure pretty good the functionality, for drawn+parametric... Well I think it could be better, I'll try some other things if I feel inspired :-)

flowernert:
I found the culprits for the GTK warnings, it should be ok yet.

I'll do the buttons icons change next.

TurboGit:
Good news!

flowernert:
Ideas prototyping for uniform mask icons

partial circle gradient (simpler, fits better in "bauhaus" style)
image

partial circle gradient with outwards marker (more clear about functionnality)
image

partial circle gradient with inwards marker (same)
image

TurboGit:
I think for uniform, the first is better. The third for me looks more like a parametric mask, the arrow being the "setting" parameter for the mask.

TurboGit:
But yet, the gradient does not convey well the fact that it is uniform to me.

TurboGit:
Based on your prototypes, why not:

- a complete/close circle full grey for uniform
- a partial circle with gradient (and possibly arrow) for parametric
- and from there add the little pencil for drawn+parametic

?

flowernert:
>     * a partial circle with gradient (and possibly arrow) for parametric
> 
>     * and from there add the little pencil for drawn+parametic

Seems good to me

I'm still not satisfied for uniform but a circle will probably be ok.

The gradient for uniform is here to figure the notion of opacity that is adjusted through the slider, however I agree that the B/W gradient can be misleading.

TurboGit:
Which button icons?

For the module, the reset and preset icons are in paint.c, and they are draw with:

`btn = dtgtk_button_new(dtgtk_cairo_paint_reset, CPF_STYLE_FLAT | CPF_DO_NOT_USE_BORDER, NULL);`

Where dtgtk_cairo_paint_reset is the draw routine in paint.c.

flowernert:
Thanks for the reply, I meant the icons of each modules (circled in red).
image

I expected to see them drawn in the code through cairo in a `dtgtk_cairo_paint_*` function, but I realized they are separate png/svg icons loaded from the `/$DARKTABLE_DIR/share/darktable/pixmaps/plugins/darkroom/` directory in imageop.c code.

TurboGit:
Ah, ok those are indeed svg.

flowernert:
It took me a bit of time to get to the result I wanted using Cairo but that's it, I'm finally done with the icons :) 

image

I did the raster round-shaped as we discussed but I now realize that maybe it makes more sense to have it square.

TurboGit:
That's really nice! Like it!

TurboGit:
Really nice indeed! Icons are readable and this save some clicks! Nice usability fix to me.

aurelienpierre:
I don't see it in the UI while compiling master. Is there an option to enable ?

flowernert:
I thought the same will doing the toggle buttons instead of the combobox @aurelienpierre, however I also though that the section and the verbose tooltip that comes with it may be helpful to novice users.

I saw a commit today that removed this big unused space at the bottom.

aurelienpierre:
> I saw a commit today that removed this big unused space at the bottom.

Strange, I don't seem to have it. The commit is in git log, but the UI is the same.

> however I also though that the section and the verbose tooltip that comes with it may be helpful to novice users.

Since it's a systematic/recurring option available in all the IOPs, I think it's not a problem. People will need to open the manual and look at the tooltip, but it would declutter the UI a bit.

PR #2022, opened 2019-01-20
Nlmeans: enable bilateral filtering (P=0)

rawfiner:
This is already possible in denoise profile, but was not available here, while it does not require any code change to the core of the algorithm.
Note that I did not test the opencl code yet, I currently do not have access to my main computer.

To test this, in denoise non local, set patch size to 0 (the slider bound is a soft boundary. It is already possible to set it to 0 without this PR because I made a mistake on it earlier as I copy pasted the range from denoiseprofile without thinking there would be some memcpy in the case of P=0, and I tested only bigger values and did not paid attention to the value 0. Anyway, this PR gives a denoising behavior for P=0)


TurboGit:
Sounds good and working. Thanks!

rawfiner:
Thank you for the review and testing :-)

PR #2024, opened 2019-01-21
atrous (equalizer): optimize loops and vectorization

aurelienpierre:
Improve the performance by up to 30 % on AVX2 CPU (SSE2 codepath only).

* rewrite the pointers increments to move them to inner loops 
* consequently, ask OpenMP to collapse outer loops (width and height loops)
* avoid `size_t` and `int` castings and use only `size_t` indexers (avoid 2 (< 0) checks per pixel on indices)
* factorize some code and move `fast_expf_sse` to the `common/sse.h` lib
* use the same increment for `float` and `__m128` pointers (saves 2 multiplications per pixel)
* move constant definitions outsive of loops when possible
* rewrite SSE2 intrinsics into pure algebraic operations when possible (better readability, more compiler optimization in some cases)
* use `simd` pragma in OpenMP loops so compilers can do magic taking 128-bits SSE2 vectors and turning them into 512-bits AVX2 vectors on compatible architectures (it seems).

Now, zoomed in full-screen mode, atrous runs on-par or faster than local-contrast laplacian (which works on 1 channel instead of 3), and twice as slow at 100 % zoom (much better).

### DIFF
--- a/src/iop/atrous.c
+++ b/src/iop/atrous.c
@@ -463,8 +463,7 @@ static void eaw_decompose_sse2(float *const out, const float *const in, float *c
      * to avoid unneeded branching in the inner loops */
     for(size_t i = 2 * mult; i < width - 2 * mult; i++)
     {
-      __m128 sum_jj[5] = { _mm_setzero_ps() };
-      __m128 wgt_jj[5] = { _mm_setzero_ps() };
+      __m128 sum_jj[5], wgt_jj[5];
 
       const size_t inc = 4 * (j * width + i);
 
--- a/src/iop/atrous.c
+++ b/src/iop/atrous.c
@@ -345,7 +345,11 @@ static void eaw_decompose(float *const out, const float *const in, float *const 
       {
         for(size_t ii = 0; ii < 5; ii++)
         {
-          px2 = ((float *)in) + 4 * (i - 2 * mult + (j - 2 * mult) * width + jj * (width - 5) * mult + ii * mult);
+          const size_t iii = (ii)-2;
+          const size_t jjj = (jj)-2;
+          const size_t x = i + mult * iii;
+          const size_t y = j + mult * jjj;
+          px2 = (float *)in +  4 * (x + y * width);
           SUM_PIXEL_CONTRIBUTION_COMMON(ii, jj);
         }
       }
@@ -458,7 +462,11 @@ static void eaw_decompose_sse2(float *const out, const float *const in, float *c
       {
         for(int ii = 0; ii < 5; ii++)
         {
-          px2 = _mm_load_ps(((float *)in) + 4 * (i - 2 * mult + (j - 2 * mult) * width + jj * (width - 5) * mult + ii * mult));
+          const size_t iii = (ii)-2;
+          const size_t jjj = (jj)-2;
+          const size_t x = i + mult * iii;
+          const size_t y = j + mult * jjj;
+          px2 = _mm_load_ps((float *)in +  4 * (x + y * width));
           SUM_PIXEL_CONTRIBUTION_COMMON_SSE2(ii, jj);
         }
       }
--- a/src/iop/atrous.c
+++ b/src/iop/atrous.c
@@ -584,7 +584,7 @@ static void eaw_synthesize(float *const out, const float *const in, const float 
   const float boost[4] = { boostf[0], boostf[1], boostf[2], boostf[3] };
 
 #ifdef _OPENMP
-#pragma omp parallel for simd default(none) schedule(static) collapse(2)
+#pragma omp parallel for default(none) schedule(static) collapse(2)
 #endif
   for(size_t k = 0; k < (size_t)4 * width * height; k += 4)
   {
@@ -609,7 +609,7 @@ static void eaw_synthesize_sse2(float *const out, const float *const in, const f
   __m128 *const mask = (__m128 *)&maski;
 
 #ifdef _OPENMP
-#pragma omp parallel for simd schedule(static) default(none)
+#pragma omp parallel for schedule(static) default(none)
 #endif
   for(size_t k = 0; k < width * height * 4; k += 4)
   {
--- a/src/iop/atrous.c
+++ b/src/iop/atrous.c
@@ -34,6 +34,7 @@
 #include <memory.h>
 #include <stdlib.h>
 #if defined(__SSE__)
+#include "common/sse.h"
 #include <xmmintrin.h>
 #endif
 
@@ -146,27 +147,17 @@ void connect_key_accels(dt_iop_module_t *self)
     (a), (a), (a), (a)                                                                                       \
   }
 
-static const __m128 fone ALIGNED(16) = VEC4(0x3f800000u);
-static const __m128 femo ALIGNED(16) = VEC4(0x00adf880u);
 static const __m128 ooo1 ALIGNED(16) = { 0.f, 0.f, 0.f, 1.f };
-
-/* SSE intrinsics version of dt_fast_expf defined in darktable.h */
-static inline __m128 dt_fast_expf_sse2(const __m128 x)
-{
-  __m128 f = _mm_add_ps(fone, _mm_mul_ps(x, femo)); // f(n) = i1 + x(n)*(i2-i1)
-  __m128i i = _mm_cvtps_epi32(f);                   // i(n) = int(f(n))
-  __m128i mask = _mm_srai_epi32(i, 31);             // mask(n) = 0xffffffff if i(n) < 0
-  i = _mm_andnot_si128(mask, i);                    // i(n) = 0 if i(n) < 0
-  return _mm_castsi128_ps(i);                       // return *(float*)&i
-}
-
 #endif
 
 static inline void weight(const float *c1, const float *c2, const float sharpen, float *weight)
 {
   float square[3];
-  for(int c = 0; c < 3; c++) square[c] = c1[c] - c2[c];
-  for(int c = 0; c < 3; c++) square[c] = square[c] * square[c];
+  for(int c = 0; c < 3; c++)
+  {
+    square[c] = c1[c] - c2[c];
+    square[c] *= square[c];
+  }
 
   const float wl = dt_fast_expf(-sharpen * square[0]);
   const float wc = dt_fast_expf(-sharpen * (square[1] + square[2]));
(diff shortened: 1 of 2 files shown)


aurelienpierre:
Hold on, I just found an error on pointers shifting in the wavelet decomposition

aurelienpierre:
I need help to understand what makes CLANG fail

src/iop/atrous.c
```diff
@@ -220,29 +209,27 @@ static inline __m128 weight_sse2(const __m128 *c1, const __m128 *c2, const float
 #if defined(__SSE2__)
 #define SUM_PIXEL_CONTRIBUTION_COMMON_SSE2(ii, jj)                                                           \
   {                                                                                                          \
-    const __m128 f = _mm_set1_ps(filter[(ii)] * filter[(jj)]);                                               \
-    const __m128 wp = weight_sse2(px, px2, sharpen);                                                         \
-    const __m128 w = _mm_mul_ps(f, wp);                                                                      \
-    const __m128 pd = _mm_mul_ps(w, *px2);                                                                   \
-    sum = _mm_add_ps(sum, pd);                                                                               \
-    wgt = _mm_add_ps(wgt, w);                                                                                \
+    const float f = filter[(ii)] * filter[(jj)];                                                             \
```
TurboGit: const __m128 f = _mm_set1_ps(filter[(ii)] * filter[(jj)]);

src/iop/atrous.c
```diff
@@ -429,85 +399,134 @@ static void eaw_decompose(float *const out, const float *const in, float *const
...
+
+        // Loop on memory : load vectors
+        pix[0] = _mm_load_ps((float *)in +  4 * (shift_filter(mult, i, 0) + y));
+        pix[1] = _mm_load_ps((float *)in +  4 * (shift_filter(mult, i, 1) + y));
+        pix[2] = _mm_load_ps((float *)in +  4 * (shift_filter(mult, i, 2) + y));
+        pix[3] = _mm_load_ps((float *)in +  4 * (shift_filter(mult, i, 3) + y));
+        pix[4] = _mm_load_ps((float *)in +  4 * (shift_filter(mult, i, 4) + y));
+
+        // Loop on values
+        w[0] = filter_jj[0] * weight_sse2(px, pix[0], sharpen);
```
TurboGit: w[0] = _mm_set1_ps(filter_jj[0]) * weight_sse2(px, pix[0], sharpen);
        w[1] = _mm_set1_ps(filter_jj[1]) * weight_sse2(px, pix[1], sharpen);
        w[2] = _mm_set1_ps(filter_jj[2]) * weight_sse2(px, pix[2], sharpen);
        w[3] = _mm_set1_ps(filter_jj[3]) * weight_sse2(px, pix[3], sharpen);
        w[4] = _mm_set1_ps(filter_jj[4]) * weight_sse2(px, pix[4], sharpen);

TurboGit:
BTW, see my related PR here:
https://github.com/darktable-org/darktable/pull/2025

src/iop/atrous.c
aurelienpierre: thanks ! That's odd though, because I use the same `float` scalar × `__m128` vector in colorbalance and filmic, and it never complained

TurboGit: Hum... Where? I can't see one... Although I could have missed one...

src/iop/atrous.c
aurelienpierre: done

src/iop/atrous.c
aurelienpierre: seems like you found them :-) done in last commit

aurelienpierre:
Still something that doesn't agree with CLANG :roll_eyes:

src/iop/atrous.c
```diff
@@ -251,90 +238,75 @@ static inline __m128 weight_sse2(const __m128 *c1, const __m128 *c2, const float
...
   for(int c = 0; c < 4; c++) pdetail[c] = (px[c] - sum[c]);                                                  \
-  for(int c = 0; c < 4; c++) pcoarse[c] = sum[c];                                                            \
-  px += 4;                                                                                                   \
-  pdetail += 4;                                                                                              \
-  pcoarse += 4;
+  for(int c = 0; c < 4; c++) pcoarse[c] = sum[c];
 
 #if defined(__SSE2__)
 #define SUM_PIXEL_EPILOGUE_SSE                                                                               \
-  sum = _mm_mul_ps(sum, _mm_rcp_ps(wgt));                                                                    \
```
TurboGit: since sum is sse _m128 I think the we need to keep this line.

src/iop/atrous.c
```diff
@@ -251,90 +238,75 @@ static inline __m128 weight_sse2(const __m128 *c1, const __m128 *c2, const float
...
-                                                                                                             \
-  _mm_stream_ps(pdetail, _mm_sub_ps(*px, sum));                                                              \
```
TurboGit: And this one too.

src/iop/atrous.c
```diff
@@ -572,33 +599,28 @@ static void eaw_synthesize(float *const out, const float *const in, const float
...
-      pdetail++;
-      pin++;
-      pout += 4;
-    }
+    const __m128 pin = _mm_load_ps(in + k);
+    const __m128 pdetail = _mm_load_ps(detail + k);
+    float *pout = out + k;
+
+    const __m128 absamt
+        = _mm_max_ps(zeros, (_mm_andnot_ps(*mask, pdetail) - threshold));
```
TurboGit: Again, I'm not sure about using - instead of _mm_sub_ps

src/iop/atrous.c
```diff
@@ -572,33 +599,28 @@ static void eaw_synthesize(float *const out, const float *const in, const float
...
 #if defined(__SSE2__)
 static void eaw_synthesize_sse2(float *const out, const float *const in, const float *const detail,
-                                const float *thrsf, const float *boostf, const int32_t width,
-                                const int32_t height)
+                                const float *thrsf, const float *boostf, const size_t width,
+                                const size_t height)
 {
-  const __m128 threshold = _mm_set_ps(thrsf[3], thrsf[2], thrsf[1], thrsf[0]);
-  const __m128 boost = _mm_set_ps(boostf[3], boostf[2], boostf[1], boostf[0]);
+  const __m128 threshold = _mm_loadu_ps(thrsf);
```
TurboGit: This is loading threshold not in the same order as original. Likewise for boost.

src/iop/atrous.c
aurelienpierre: this is fully equivalent for the compiler but only more readable

src/iop/atrous.c
aurelienpierre: same

src/iop/atrous.c
aurelienpierre: you can ask @heckflosse about that

src/iop/atrous.c
aurelienpierre: yes it does :-) the `_mm_set_ps` takes its arguments in reverse order, as the `_mm_load_ps`, the natural order can be used with `_mm_setr_ps. See https://software.intel.com/sites/landingpage/IntrinsicsGuide/#text=_mm_set_ps&expand=4895

aurelienpierre:
> All those SSE circuitry are giving me headache :)

I know right ?

src/iop/atrous.c
TurboGit: My bad, sorry :)

aurelienpierre:
> I'll have to actually test this. Is that a final version?

I think so. Maybe we can let it sit for a while, to see if I have more ideas to speed-up the computations.

src/iop/atrous.c
heckflosse: Using '-', '+', '*', '/', '-=' .... works fine with gcc and clang, increases readability and also (at least in my experience) gives sometimes faster code (for example a + b * c can give an fma (if supported) while '_mm_add_ps(a, _mm_mul_ps(b,c)' or vice versa in my tests was translated following exactly the intrinsics.

Though I have to admit, that the last time I looked at the assembly is a while ago.

Anyway, in RT I always use the more readable notation and try to avoid intrinsics and never experienced problems, neither with clang, nor with gcc.

One exception is the following, which is supported by gcc, but not by clang:

```
__m128 a;
float b;

a = a * b;
```

TurboGit: @heckflosse : thanks for the input. indeed the float * _m128 is not supported by clang, we had just fixed this in filmic.

src/iop/atrous.c
```diff
@@ -187,21 +178,35 @@ static inline void weight(const float *c1, const float *c2, const float sharpen,
...
-  __m128 square2 = _mm_shuffle_ps(square, square, _MM_SHUFFLE(3, 1, 2, 0)); // (?, d2, d3, d1)
-  __m128 added = _mm_add_ps(square, square2);                               // (?, d2+d3, d2+d3, 2*d1)
-  added = _mm_sub_ss(added, square);                                        // (?, d2+d3, d2+d3, d1)
-  __m128 sharpened = _mm_mul_ps(added, vsharpen);                   // (?, -s*(d2+d3), -s*(d2+d3), -s*d1)
-  __m128 exp = dt_fast_expf_sse2(sharpened);                        // (?, wc, wc, wl)
+  const __m128 vsharpen = _mm_set1_ps(-sharpen);                    // (-s, -s, -s, -s)
+  const __m128 diff = (c1 - c2);
+  const __m128 square = diff * diff;                                // (?, d3, d2, d1)
+  const __m128 square2 = _mm_shuffle_ps(square, square, _MM_SHUFFLE(3, 1, 2, 0)); // (?, d2, d3, d1)
+  const __m128 added = square + square2 - square;                   // (?, d2+d3, d2+d3, d1)
```
heckflosse: @aurelienpierre May I ask what this is for?
Is this a trick for clipping small values of square2 to zero?

So for example 100000000 + 0.000001 - 100000000 = 0

aurelienpierre: @heckflosse I'm not the author of that code, so I think @hanatos will answer better than me. I just made the loops a tad faster here.

heckflosse: @aurelienpierre Ok, I just was curious because at least on first sight calculating `a = b + c - b` does not make much sense

aurelienpierre:
> Most timing were around 0.35s with current version and with this PR most timing are around 0.4s.

Nice :rofl: 

Did you test on portrait and landscape pictures ? I found that, at similar resolution, portrait-oriented pics process faster than landscapes.

heckflosse:
@TurboGit 
> To me on an i7-4710MQ @ 2.50GHz the new version is slower.

Does this CPU have one avx2 unit per core or one per hyperthread?
I ask, because on my AMD FX8350 (8 cores, 8 128bit units, but only 4 256bit units)
using AVX on all cores in parallel also is slower than using SSE2 on all cores in parallel because each 256bit unit has to be shared among 2 cores.

parafin:
One should also consider this when talking about AVX performance: https://blog.cloudflare.com/on-the-dangers-of-intels-frequency-scaling/

aurelienpierre:
Also please note that I compile with `-O3`.

src/iop/atrous.c
aurelienpierre: damit ! you are right ! so easy to confuse  `_mm_sub_ps` and `_mm_sub_ss`… Thanks !

heckflosse:
@TurboGit 
> Most timing were around 0.35s with current version and with this PR most timing are around 0.4s.

I didn't measure. But I measure very often in RT when I optimize the code.

In my experience a reliable measure is the median of 7 measures. Anything below that (median of 5 measures for examples) is not reliable in my experience.

Just my 2 cents...

aurelienpierre:
My reference for this was the local contrast module, because it does essentially the same things but on only 1 channel instead of 3. Zoomed at screen size, this PR makes atrous run just a bit faster than local contrast, which was clearly not the case before, and zoomed at 100 %, atrous runs twice as fast, which is good since it does 3 times as many operations.

cryptomilk:
Have you ever built darktable with "UndefinedSanitizer", this normally also reveals quite some issues.

aurelienpierre:
No, is that an LLVM option ?

cryptomilk:
See: https://gitlab.com/cmocka/cmocka/blob/master/cmake/Modules/DefineCompilerFlags.cmake#L38

TurboGit:

PR #2025, opened 2019-01-21
filmic: fix sse with clang.

TurboGit:


aurelienpierre:
clang complained here ? I missed that.

aurelienpierre:
something is broken in SSE version

Issue #2006, opened 2019-01-16
Broken "Find location" on Windows


techexo:
**Describe the bug**
I just want to draw attention on a Windows bug described on Redmine: [the "find location" feature is broken on Windows](https://redmine.darktable.org/issues/12523). It seems serious enough to have a totally broken feature on a new version build.

It is a Windows port issue, as the feature is working fine on Linux.

**To Reproduce**
1. Go to 'map'
2. Enter location
3. Press 'Enter'

**Expected behavior**
A list of possible locations should appear.

**Screenshots**

**Platform (please complete the following information):**
 - OS: Windows 10 64 bits.
 - Version 2.6.0

peterbud:
Confirmed. I'll look into this

peterbud:
What I see during debugging is that the curl `curl_easy_perform` call [here](https://github.com/darktable-org/darktable/blob/master/src/libs/location.c#L361) returns with `CURLE_SSL_CACERT_BADFILE`, so it seems this is a libcurl problem on the MSYS2/MINGW.

I'll keep researching.

techexo:
I don't know anything about that so don't take my word for it, but shouldn't there be somewhere where CA infos are stored to fetch OSM infos?

If I'm not mistaken, it was migrated from `http://` to `https://` (https://github.com/darktable-org/darktable/pull/1691) some time ago, maybe libcurl needs that the `CURLOPT_CAINFO` be explicitely defined on Windows to access CA root certificates? (https://curl.haxx.se/libcurl/c/CURLOPT_CAINFO.html)

Maybe [this](https://curl.haxx.se/mail/lib-2016-08/0118.html) could also be of interest.

peterbud:
Thanks.
Adding:
`curl_easy_setopt(curl, CURLOPT_SSL_VERIFYPEER, FALSE);`

solves the problem, and search is working,  but I'm not yet sure this is the right approach. I need to understand how curl is trying to locate the CA file.

parafin:
Obviously it's not. You need to include CA bundle file into package and point curl to it like that:
https://github.com/darktable-org/darktable/blob/4f9203a65a64b9bb3aa73eb18f9f048179e602f3/packaging/macosx/start#L44

techexo:
@peterbud Great, it means at least the the problem is identified, and is just a TLS configuration issue. I support @parafin in that we should not skip certificate verification, but link to  ca-bundle.

peterbud:
[...]

PR #2028, opened 2019-01-22
Fix TLS issues on darktable

techexo:
On the way to fix https://github.com/darktable-org/darktable/issues/2006.

The file `curl-ca-bundle.crt` should end up in `./share/curl/`.

The idea is to allow curl to use the bundle of root certificates to verify TLS connections to https websites, including openstreetmap (used in the map module, broken on Windows right now), facebook, picasa and
piwigo.

Please edit the PR, because I have absolutely no idea how to package anything :smile:.

### DIFF
--- a/cmake/windows-macros.cmake
+++ b/cmake/windows-macros.cmake
@@ -204,6 +204,13 @@ if (WIN32 AND NOT BUILD_MSYS2_INSTALL)
     endforeach()
   endif(ISO_CODES_FOUND)
 
+  # Add ca-cert for curl
+  install(FILES
+      "${MINGW_PATH}/../ssl/certs/ca-bundle.crt"
+      DESTINATION share/curl/
+      RENAME curl-ca-bundle.crt
+      COMPONENT DTApplication)
+
 endif(WIN32 AND NOT BUILD_MSYS2_INSTALL)
 
 endfunction()
--- a/src/imageio/storage/facebook.c
+++ b/src/imageio/storage/facebook.c
@@ -332,7 +332,7 @@ static JsonObject *fb_query_get(FBContext *ctx, const gchar *method, GHashTable 
   curl_easy_setopt(ctx->curl_ctx, CURLOPT_VERBOSE, 2);
 #endif
   curl_easy_setopt(ctx->curl_ctx, CURLOPT_WRITEFUNCTION, curl_write_data_cb);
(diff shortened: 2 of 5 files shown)


parafin:
I don't think that we should include curl-ca-bundle.crt in darktable repo. I think it should be instead copied into Windows package from some system location like it's done for macOS package. The reason being that someone has to regularly update curl-ca-bundle.crt in our repo if we choose to add it.

About the changes to the code - you specify relative path for CURLOPT_CAINFO, are you sure it works like that? Most likely it's relative to current directory, which can be anywhere. I think absolute path has to be constructed and probably tested for existence so that we don't break default system path for this file.

techexo:
Oh OK, sorry 🙏 Feel free to PR massively on techexo/darktable, I'll accept everything 😉

techexo:
For information, direct link towards last export from curl teams is https://curl.haxx.se/ca/cacert.pem.
@parafin, no I am not sure indeed for relative paths. I did them relative to `bin/` to go in `share/`, but it may not be that. Feel free to PR on techexo:tls-fix the right thing.

parafin:
About `curl_easy_setopt` calls I would say construct absolute path using `dt_loc_init_datadir` function and then check it for existence - I think this logic is platform-independent.

techexo:
@peterbud I rebased my branch on darktable/master and integrated your suggested change for the bundle in `cmake/windows-macros.cmake`.

@parafin & @peterbud You're obviously more competent than me to solve this issue. If you think it's easier, do not hesitate one second to make another PR and I'll close that one.

I'll try to rebase regularly, or I'll implement changes if you send me links to the patches. Whatever works to solve this issue :)

parafin:
My suggestions covers MacOS case and is at least harmless for Linux. I'm too busy right now to make a PR, just look for examples of `dt_loc_init_datadir` usage in dt sources.

TurboGit:
Do you think this can be fixed for 2.6.1? It remains 15 more days for integration.

peterbud:
I'll look into this latest at the weekend.

techexo:
Rebased my repo, if it can help.

techexo:
I think PR https://github.com/darktable-org/darktable/pull/2074 that was merged for 2.6.1 makes this PR useless. Thanks @peterbud for the work, this PR should be refused.

PR #2033, opened 2019-01-23
Add version string to opencl kernel cache dir to rebuild caches after driver update

kxv:
This adds the opencl driver version string to the kernel cache to cause a rebuild after a driver update.
According to https://github.com/intel/compute-runtime/issues/125#issuecomment-455504686 OpenCV is using a similar naming scheme.

This could solve https://redmine.darktable.org/issues/12541 .

### DIFF
--- a/src/common/opencl.c
+++ b/src/common/opencl.c
@@ -352,6 +352,7 @@ static int dt_opencl_device_init(dt_opencl_t *cl, const int dev, cl_device_id *d
   char dtcache[PATH_MAX] = { 0 };
   char cachedir[PATH_MAX] = { 0 };
   char devname[1024];
+  char drvversion[1024];
   double tstart, tend, tdiff;
   dt_loc_get_user_cache_dir(dtcache, sizeof(dtcache));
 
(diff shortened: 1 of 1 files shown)


TurboGit:
This has always worked for me with nvidia! Any explanation why this is need on your side?

kxv:
> This has always worked for me with nvidia! Any explanation why this is need on your side?

Please let me cite R from the linked darktable issue: 

> "I have been using the NEO drivers without issue on almost daily builds of master for quite sometime now. I ran into some issues like this before I discovered that you have to regenerate the opencl kernels sometimes after updating the neo drivers. Now, on archlinux, whenever the neo drivers get updated, I rm -rf ~/config/darktable/cached_kernels_for_IntelRGen9HDGraphicsNEO . Have not observed a problem since doing this."

This pull would automate the invalidation. The cached kernels are not supposed to work after a driver update of compute-runtime / NEO (see https://github.com/intel/compute-runtime/issues/125).

kxv:
Because of these problems, NEO / compute-runtime is on the driver blacklist per https://github.com/darktable-org/darktable/commit/4788b382c8278fab0840866db761c40ef6612dcf right now
This also means, one needs to circumvent blacklisting for testing.

TurboGit:
@kxv : I know, I did blacklist it.

But you have not answered my question. I'm just trying to understand why this is needed for NEO whereas it is not for AMD and NVIDIA. Maybe the issue is somewhere else? I just want to have at least some idea of what a PR is about before merging. That's all.

kxv:
I understand.
It surprises me, that NVDIA/AMD drivers never caused such an issue before. I can only assume this is caused by compute-runtime being a newer driver and thus being subject to bigger changes.
Maybe https://github.com/alalek can shed some more light on this.

aurelienpierre:
As for Neo, it breaks on Windows with dt, but @rawfiner is using it to debug OpenCL with success on Linux Mint or something. Maybe just blacklist it for Windows ?

rawfiner:
Well, indeed I use neo to debug opencl on linux mint usually with success, but I use it almost exclusively to debug only a few modules (and I had some strange bugs that disapeared magically and I still don't know why), so I am not very confident about it...

TurboGit:
On, reviewed and tested. This works fine and is generally useful even outside of the Intel neo driver. Thanks.

PR #2035, opened 2019-01-26
[WIP] Filmic v3 + lots of code refactoring

aurelienpierre:
This is a WIP, so I'm putting it here now in case people want to disagree early and avoid me to work for nothing.

This started as filmic v3, with 2 new features:
1. another mode of chroma preservation, in xyY space (less saturation issues with blue skies than the current RGB one, but other problems with reds)
1. a gamut compression in IPT-HDR color space

To save some space in the UI, the safety factor is removed (parameters non-affected). Also, the latitude is expressed in % of the dynamic range insteal of EV (parameters converted in `legacy_param`

Since I needed another colorspace, I started refactoring the `colorspaces_inline_conversion.h` lib to sync the SSE and C versions and write the color conversion fonctions as simple matrix-vector dot product.

This triggered another refactoring, since there are like 5 copies of exactly the same `mat3mulv` functions in the code, I created an `algebra.h` lib with basic linear algebra operations, so we can have a generic interface and start optimizing the code for SSE2-3-4, AVX1-2 at one place. Although, at this time, when I try to branch:

```
static inline void mat3mulv(float *dst, const float mat[3][3], const float *const v)
{
#ifdef __SSE2__
  mat3mulv_sse2((__m128 *)dst, mat, (const __m128 *cons)t v)
#else 
  for(int k = 0; k < 3; k++)
  {
    float x = 0.0f;
    for(int i = 0; i < 3; i++) x += mat[k][i] * v[i];
    dst[k] = x;
  }
}
#endif 
```
that segfault.

Chasing bugs with the new IPT-HDR space, I refactored some maths functions (fast_log, saturation) that are used in profilegamma or colorbalance too.

Bottom line, if you want to test:
1. the gamut mapping is implemented only in SS2 version, and bugged
2. the xyY chroma preservation mode is implemented in C and SSE2 versions only (no OpenCL)
3. the auto-tuner color-picker has been removed for now until I figure out how to put it at the right of the graph, to save space vertically.


TurboGit:
The segfault is probably in the SSE2 code path because you cast the float to _m128 but those are probably not properly aligned. You need to use _mm_set1_ps() I think.

TurboGit:
For the color picker next to the graph you need a hbox. I can give a hand on this if you like.

hanatos:
_mm_load_ps() is probably what you mean. but swizzling data from float to
sse registers may well eat up any performance benefit you might get here.

did you feed this code in a minimal main.c or similar to clang and look at
the disassembly? sometimes it uses vector code already. in a float[3][3]
case it may be hard to cast it to a __m128[4] block automatically though.

aurelienpierre:
> For the color picker next to the graph you need a hbox. I can give a hand on this if you like.

gladly :-)

> _mm_load_ps() is probably what you mean. but swizzling data from float to sse registers may well eat up any performance benefit you might get here. did you feed this code in a minimal main.c or similar to clang and look at the disassembly? sometimes it uses vector code already. in a float[3][3] case it may be hard to cast it to a __m128[4] block automatically though.

I have not yet checked that, I'm just benchmarking the running times for now. I have some background with ARM Cortex assembly, but not with X86, so the dissassembly won't be super useful info for me.

TurboGit:
@aurelienpierre : ok I will do this.

TurboGit:
> @aurelienpierre : ok I will do this.

Done. You can fetch the branch filmic-autotune

And pick the top commit to put it on your branch:

$ git cherry-pick filmic-autotune

aurelienpierre:
Thanks @TurboGit !

aurelienpierre:
closing that, I will start over.

PR #2039, opened 2019-01-27
Expose

TurboGit:
fix filmstrip display.
expose the stars/reject items


TurboGit:
@oexler : indeed, for the layout not sure now. maybe because you did set the filmstrip visible and then hide it. Not sure.

What's the issue with the hotkeys?

TurboGit:
@oexler : new version for the bug on filmstrip pushed. in enter() I have added a call to hide to ensure the filmstrip is initially hidden. Looks like this works fine.

TurboGit:
> Wait a moment, how is it that both 0 and 1 get replaced by DT_LAYOUT_ZOOMABLE? Shouldn't 0 be DT_LAYAOUT_FILEMANAGER?

Indeed! cut&paste error :( fixed. good catch, thanks.

oexler:
> What's the issue with the hotkeys?

Lighttable and filmstrip use the same hotkeys. I think even when filmstrip is hidden, those functions are called e.g. for ranking.

TurboGit:
@oexler : I think all issues (especially accels) are now fixed.

PaoloAst:
I've found some issue sorting images by: description, title, full path or color label. In these cases no images are displayed

PaoloAst:
@TurboGit did you test what I wrote? In the expose mode (very nice, indeed), if you sort by description, title, full path or color label, I'm not able to select any image

TurboGit:
Yes, and I can reproduce. Really strange behavior :) Will fix. Would have been good to open an issue.

PaoloAst:
Ok. I'll do it

Sent from my mobile

Il giorno dom 3 feb 2019, 12:54 Pascal Obry <notifications@github.com> ha
scritto:

TurboGit:
This is now fixed. Can you try again?

PaoloAst:
It works, thanks, but now shuffle sorting, starts running endless

PR #2040, opened 2019-01-28
filmic: ensure latitude_stops is initialized in auto-tune modes.

TurboGit:
Otherwise any value will be set and sanitisation can go crazy and set
the latitude to zero leading to a B&W picture.


aurelienpierre:
It looks like an ugly hack. In which case is the latitude not initialized ?

TurboGit:
I think I found the issue. A mistake of mine in the picker proxy!

PR #2042, opened 2019-01-30
fix a typo in UI string

rabauke:
Plural  possessive case must be used here, see https://www.grammarly.com/blog/possessive-case/


TurboGit:
Indeed! Thanks.

PR #2057, opened 2019-02-04
update for german translation

rabauke:
Many smaller fixes for the German translation. Most significant changes belong to the haze removal module and the new mask feathering (edge aware mask refinement) feature. In particular, the current German translation is imho misleading in the latter case.


TurboGit:
Thanks!

TurboGit:
You have updated the translation this morning, but I have another proposal:
https://github.com/darktable-org/darktable/pull/2060

Can you comment? Is that ok to be merged? Thanks.

PR #2058, opened 2019-02-04
bauhaus: fix format string

rabauke:
It's dangerous to use string specifier without width specification. Buffer overflow is possible.


TurboGit:
Sounds good!

PR #2062, opened 2019-02-05
Collect single-click

AlicVB:
This add an option in preferences to use single-click selection in the collect module (default remain to double-click)
As a side effect, this allow to directly select ranges with Shift-click (only works for ISO, aperture, exposure, focal-length, date and datetime)

### DIFF
--- a/data/darktableconfig.xml.in
+++ b/data/darktableconfig.xml.in
@@ -762,6 +762,13 @@
     <shortdescription>append or replace history stack if pasted</shortdescription>
     <longdescription>0 -- append on top of stack, 1 -- replace it.</longdescription>
   </dtconfig>
+  <dtconfig prefs="gui" section="lighttable">
+    <name>plugins/lighttable/collect/single-click</name>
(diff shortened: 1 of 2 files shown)


TurboGit:
Looks good and works fine. Thanks for reviving this, it is lot better to use the single click here. Do you think it would be possible to add aspect ratio ranges?

AlicVB:
done for aspect ratio :)
One little glitch I've found is that with single click, it's now impossible to expand a node without selecting it. I've not managed to determine if the click append on the arrow or in the text...

TurboGit:
A glitch or a feature. Clicking on the line open the children and select the parent. I find this ok.

Thanks a lot for handling the apsect filter.

And, yes almost 3 years after your first PR on this it will get merged right now!

TurboGit:
Just found an issue with this. You cannot open a folder in the folder view in the single click mode.

Can you have a look? Thanks.

AlicVB:
oops... should be fixed in #2084

PR #2063, opened 2019-02-05
main: fix memory leak & format string bug, minor reformating

rabauke:
Regarding the removed fscanf function: The number of conversion specifications in the format string must not exceed the number of pointer arguments. Giving a conversion specification and no pointer argument is an error.


TurboGit:
Looks good to me! Thanks.

PR #2066, opened 2019-02-06
fix color pick

edgardoh:
Color picker do not return the right values, as reported in [1]

This fix the rgb and lab values on the colorpicker and the final histogram and gamma module.

[1] https://www.mail-archive.com/darktable-user@lists.darktable.org/msg06952.html

### DIFF
--- a/src/develop/pixelpipe_hb.c
+++ b/src/develop/pixelpipe_hb.c
@@ -597,6 +597,273 @@ error:
 }
 #endif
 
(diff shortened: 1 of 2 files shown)


aurelienpierre:
So at which stage of the pipe do you pick the color now ?

edgardoh:
The same, but I use the input image instead of the output one.

aurelienpierre:
Ok. Ideally, we would like to pick before the output color profile is applied too.

TurboGit:
Looks good and a nice code refactoring! Thanks.

PR #2067, opened 2019-02-06
svd: fix check for non-convergence, minor refactoring

rabauke:


TurboGit:
Can you add a description as to what this is fixing (how to reproduce)? Otherwise it makes review/testing quite difficult. Thanks.

src/iop/svd.h
```diff
@@ -166,44 +174,46 @@ static inline int dsvd(
...
     }
     ++a[i*str+i];
   }
 
   /* diagonalize the bidiagonal form */
-  for (k = n - 1; k >= 0; k--) 
+  for (int k = n - 1; k >= 0; k--)
   {                             /* loop over singular values */
-    for (its = 0; its < 30; its++) 
+    for (int its = 0; its <= 30; its++)
```
rabauke: Here is the bug-fix.

src/iop/svd.h
```diff
@@ -246,13 +255,13 @@ static inline int dsvd(
         if (z < 0.0) 
         {              /* make singular value nonnegative */
           w[k] = -z;
-          for (j = 0; j < n; j++) 
+          for (int j = 0; j < n; j++)
             v[j*n+k] = (-v[j*n+k]);
         }
         break;
       }
       if (its >= 30) {
```
rabauke: Here we had a problem, was always false without fix above.

TurboGit:
@rabauke : ok, most of this patch is about refactoring. as already discussed please do separate commits. to review a code clean-up or refactoring I have nothing really to "check". for a fix I really need to review each lines. here we have in a long diff with only two definite changes. thanks for my time :)

src/iop/svd.h
```diff
@@ -28,12 +28,18 @@
...
-  double at = fabs(a), bt = fabs(b), ct, result;
-
-  if (at > bt)       { ct = bt / at; result = at * sqrt(1.0 + ct * ct); }
-  else if (bt > 0.0) { ct = at / bt; result = bt * sqrt(1.0 + ct * ct); }
-  else result = 0.0;
-  return result;
+  double at = fabs(a), bt = fabs(b);
+  if (at > bt)
+  {
+    double ct = bt / at;
```
TurboGit: as we are at it: this can be a const

src/iop/svd.h
```diff
@@ -28,12 +28,18 @@
...
+    return at * sqrt(1.0 + ct * ct);
+  }
+  if (bt > 0.0)
+  {
+    double ct = at / bt;
```
TurboGit: likewise, can be a const

src/iop/svd.h
```diff
@@ -269,9 +278,9 @@ static inline int dsvd(
 
       /* next QR transformation */
       c = s = 1.0;
-      for (j = l; j <= nm; j++) 
+      for (int j = l; j <= nm; j++)
       {
-        i = j + 1;
+        int i = j + 1;
```
TurboGit: if not modified maybe can be made a const?

rabauke:
I have reduced changes to a minimum to fix the bug in the convergence check and to exclude refactoring work. I will open a further pull request for refactoring as soon as this has been merged.

TurboGit:
Thanks, far easier to review.

PR #2068, opened 2019-02-07
masks: fix detection of mouse pointer inside a path

### LINKED ISSUE #2065: Control points disappear when mouse is inside a mask
**Describe the bug**
Some path masks cannot be edited because it is not correctly detected whether the mouse is located inside or outside the shape.
For example I created [this image](https://github.com/darktable-org/darktable/files/2834377/masks.zip): when the mouse is inside the red area the masks are not displayed as expected, with control points and outer border.

**To Reproduce**
Create a path mask where the outer border self-intersects at the first control point, but there might be other ways.

**Expected behavior**
Control points and outer border should not disappear when mouse is inside.

 - OS: Arch Linux
 - Version 2.6.0

**Additional context**
Problem should be in this function, if needed I can submit a PR tomorrow. https://github.com/darktable-org/darktable/blob/e00ec0349ec8f825dbf1441c31eb90bf09c080a9/src/develop/masks/masks.c#L2505

crazyhg:
Ray casting algorithm for mouse pointer now works even when there is a
skipping zone (self intersection) between the last points and the first
ones of a mask border.
I think this should be enough to fix #2065.

### DIFF
--- a/src/develop/masks/masks.c
+++ b/src/develop/masks/masks.c
@@ -2510,21 +2510,27 @@ int dt_masks_point_in_form_exact(float x, float y, float *points, int points_sta
 
   if(points_count > 2 + points_start)
   {
-    float last = isnan(points[points_count * 2 - 1]) ? -INFINITY : points[points_count * 2 - 1];
+    int start = isnan(points[points_start * 2]) && !isnan(points[points_start * 2 + 1])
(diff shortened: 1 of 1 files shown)


TurboGit:
Working fine! Nice fix. Thanks.

PR #2069, opened 2019-02-07
add colorspace to color picker

edgardoh:
This adds to the over/under exposed button the option to select a profile, that is also used for the histogram and color picker.

There's a current design limitation, there are modules after colorout, if any is enabled results may be inaccurate. This is fixed in PR #1841.

### DIFF
--- a/src/common/colorspaces.c
+++ b/src/common/colorspaces.c
@@ -945,12 +945,15 @@ static const char *_profile_names[] =
   N_("standard color matrix"),
   N_("enhanced color matrix"),
   N_("vendor color matrix"),
-  N_("alternate color matrix")
+  N_("alternate color matrix"),
+  N_("BRG (experimental)"),
+  N_("export profile"),
+  N_("softproof profile")
 };
 
 static dt_colorspaces_color_profile_t *_create_profile(dt_colorspaces_color_profile_type_t type,
                                                        cmsHPROFILE profile, const char *name,
-                                                       int in_pos, int out_pos, int display_pos)
+                                                       int in_pos, int out_pos, int display_pos, int category_pos)
 {
   dt_colorspaces_color_profile_t *prof;
   prof = (dt_colorspaces_color_profile_t *)calloc(1, sizeof(dt_colorspaces_color_profile_t));
@@ -960,6 +963,7 @@ static dt_colorspaces_color_profile_t *_create_profile(dt_colorspaces_color_prof
   prof->in_pos = in_pos;
   prof->out_pos = out_pos;
   prof->display_pos = display_pos;
+  prof->category_pos = category_pos;
   return prof;
 }
 
@@ -1109,6 +1113,7 @@ static GList *load_profile_from_dir(const char *subdir)
           prof->in_pos = -1;
           prof->out_pos = -1;
           prof->display_pos = -1;
+          prof->category_pos = -1;
           temp_profiles = g_list_append(temp_profiles, prof);
         }
 
@@ -1135,60 +1140,68 @@ dt_colorspaces_t *dt_colorspaces_init()
 
   int in_pos = -1,
       out_pos = -1,
-      display_pos = -1;
+      display_pos = -1,
+      category_pos = -1;
+
+  // init the category profile with NULL profile, the actual profile must be retrieved dinamically by the caller
+  res->profiles = g_list_append(res->profiles, _create_profile(DT_COLORSPACE_EXPORT, NULL,
+                                                               _("export profile"), -1, -1, -1, ++category_pos));
+
+  res->profiles = g_list_append(res->profiles, _create_profile(DT_COLORSPACE_SOFTPROOF, NULL,
+                                                               _("softproof profile"), -1, -1, -1, ++category_pos));
 
   // init the display profile with srgb so some stupid code that runs before the real profile could be fetched has something to work with
   res->profiles = g_list_append(res->profiles, _create_profile(DT_COLORSPACE_DISPLAY, dt_colorspaces_create_srgb_profile(),
-                                                               _("system display profile"), -1, -1, ++display_pos));
+                                                               _("system display profile"), -1, -1, ++display_pos, ++category_pos));
   // we want a v4 with parametric curve for input and a v2 with point trc for output
   // see http://ninedegreesbelow.com/photography/lcms-make-icc-profiles.html#profile-variants-and-versions
   // TODO: what about display?
   res->profiles = g_list_append(res->profiles, _create_profile(DT_COLORSPACE_SRGB,
                                                                dt_colorspaces_create_srgb_profile_v4(),
                                                                _("sRGB (e.g. JPG)"),
-                                                               ++in_pos, -1, -1));
+                                                               ++in_pos, -1, -1, -1));
 
   res->profiles = g_list_append(res->profiles, _create_profile(DT_COLORSPACE_SRGB,
                                                                dt_colorspaces_create_srgb_profile(),
                                                                _("sRGB (web-safe)"),
-                                                               -1, ++out_pos, ++display_pos));
+                                                               -1, ++out_pos, ++display_pos, ++category_pos));
 
   res->profiles = g_list_append(res->profiles, _create_profile(DT_COLORSPACE_ADOBERGB,
                                                                dt_colorspaces_create_adobergb_profile(),
                                                                _("Adobe RGB (compatible)"),
-                                                               ++in_pos, ++out_pos, ++display_pos));
+                                                               ++in_pos, ++out_pos, ++display_pos, ++category_pos));
 
   res->profiles = g_list_append(res->profiles, _create_profile(DT_COLORSPACE_LIN_REC709,
                                                                dt_colorspaces_create_linear_rec709_rgb_profile(),
                                                                _("linear Rec709 RGB"),
-                                                               ++in_pos, ++out_pos, ++display_pos));
+                                                               ++in_pos, ++out_pos, ++display_pos, ++category_pos));
 
   res->profiles = g_list_append(res->profiles, _create_profile(DT_COLORSPACE_LIN_REC2020,
                                                                dt_colorspaces_create_linear_rec2020_rgb_profile(),
                                                                _("linear Rec2020 RGB"),
-                                                               ++in_pos, ++out_pos, ++display_pos));
+                                                               ++in_pos, ++out_pos, ++display_pos, ++category_pos));
 
   res->profiles = g_list_append(res->profiles, _create_profile(DT_COLORSPACE_XYZ,
                                                                dt_colorspaces_create_xyz_profile(),
                                                                _("linear XYZ"),
                                                                ++in_pos,
-                                                               dt_conf_get_bool("allow_lab_output") ?  ++out_pos : -1, -1));
+                                                               dt_conf_get_bool("allow_lab_output") ?  ++out_pos : -1, -1, -1));
 
   res->profiles = g_list_append(res->profiles, _create_profile(DT_COLORSPACE_LAB,
                                                                dt_colorspaces_create_lab_profile(),
                                                                _("Lab"),
                                                                ++in_pos,
-                                                               dt_conf_get_bool("allow_lab_output") ?  ++out_pos : -1, -1));
+                                                               dt_conf_get_bool("allow_lab_output") ?  ++out_pos : -1, -1, -1));
 
   res->profiles = g_list_append(res->profiles, _create_profile(DT_COLORSPACE_INFRARED,
                                                                dt_colorspaces_create_linear_infrared_profile(),
                                                                _("linear infrared BGR"),
-                                                               ++in_pos, -1, -1));
+                                                               ++in_pos, -1, -1, -1));
 
   res->profiles = g_list_append(res->profiles, _create_profile(DT_COLORSPACE_BRG,
                                                                dt_colorspaces_create_brg_profile(),
                                                                _("BRG (for testing)"),
-                                                               ++in_pos, ++out_pos, ++display_pos));
+                                                               ++in_pos, ++out_pos, ++display_pos, -1));
 
   // temporary list of profiles to be added, we keep this separate to be able to sort it before adding
   GList *temp_profiles;
@@ -1209,17 +1222,22 @@ dt_colorspaces_t *dt_colorspaces_init()
     dt_colorspaces_color_profile_t *prof = (dt_colorspaces_color_profile_t *)iter->data;
     prof->out_pos = ++out_pos;
     prof->display_pos = ++display_pos;
+    prof->category_pos = ++category_pos;
   }
   res->profiles = g_list_concat(res->profiles, temp_profiles);
 
   // init display profile and softproof/gama checking from conf
   res->display_type = dt_conf_get_int("ui_last/color/display_type");
   res->softproof_type = dt_conf_get_int("ui_last/color/softproof_type");
+  res->histogram_type = dt_conf_get_int("ui_last/color/histogram_type");
   char *tmp = dt_conf_get_string("ui_last/color/display_filename");
   g_strlcpy(res->display_filename, tmp, sizeof(res->display_filename));
   g_free(tmp);
   tmp = dt_conf_get_string("ui_last/color/softproof_filename");
   g_strlcpy(res->softproof_filename, tmp, sizeof(res->softproof_filename));
+  g_free(tmp);
+  tmp = dt_conf_get_string("ui_last/color/histogram_filename");
+  g_strlcpy(res->histogram_filename, tmp, sizeof(res->histogram_filename));
   g_free(tmp);
   res->display_intent = dt_conf_get_int("ui_last/color/display_intent");
   res->softproof_intent = dt_conf_get_int("ui_last/color/softproof_intent");
@@ -1237,6 +1255,11 @@ dt_colorspaces_t *dt_colorspaces_init()
         && (!res->softproof_filename[0] || !g_file_test(res->softproof_filename, G_FILE_TEST_IS_REGULAR))))
     res->softproof_type = DT_COLORSPACE_SRGB;
 
+  if((unsigned int)res->histogram_type >= DT_COLORSPACE_LAST
+    || (res->histogram_type == DT_COLORSPACE_FILE
+        && (!res->histogram_filename[0] || !g_file_test(res->histogram_filename, G_FILE_TEST_IS_REGULAR))))
+    res->histogram_type = DT_COLORSPACE_SRGB;
+
   if((unsigned int)res->mode > DT_PROFILE_GAMUTCHECK) res->mode = DT_PROFILE_NORMAL;
 
   _update_display_transforms(res);
@@ -1249,8 +1272,10 @@ void dt_colorspaces_cleanup(dt_colorspaces_t *self)
   // remember display profile and softproof/gama checking from conf
   dt_conf_set_int("ui_last/color/display_type", self->display_type);
   dt_conf_set_int("ui_last/color/softproof_type", self->softproof_type);
+  dt_conf_set_int("ui_last/color/histogram_type", self->histogram_type);
   dt_conf_set_string("ui_last/color/display_filename", self->display_filename);
   dt_conf_set_string("ui_last/color/softproof_filename", self->softproof_filename);
+  dt_conf_set_string("ui_last/color/histogram_filename", self->histogram_filename);
   dt_conf_set_int("ui_last/color/display_intent", self->display_intent);
   dt_conf_set_int("ui_last/color/softproof_intent", self->softproof_intent);
   dt_conf_set_int("ui_last/color/mode", self->mode);
--- a/src/common/colorspaces.h
+++ b/src/common/colorspaces.h
@@ -57,7 +57,9 @@ typedef enum dt_colorspaces_color_profile_type_t
   DT_COLORSPACE_VENDOR_MATRIX = 13,
   DT_COLORSPACE_ALTERNATE_MATRIX = 14,
   DT_COLORSPACE_BRG = 15,
-  DT_COLORSPACE_LAST = 16
+  DT_COLORSPACE_EXPORT = 16, // export and softproof are categories and will return NULL with dt_colorspaces_get_profile()
+  DT_COLORSPACE_SOFTPROOF = 17,
+  DT_COLORSPACE_LAST = 18
 } dt_colorspaces_color_profile_type_t;
 
 typedef enum dt_colorspaces_color_mode_t
@@ -72,7 +74,8 @@ typedef enum dt_colorspaces_profile_direction_t
   DT_PROFILE_DIRECTION_IN = 1 << 0,
   DT_PROFILE_DIRECTION_OUT = 1 << 1,
   DT_PROFILE_DIRECTION_DISPLAY = 1 << 2,
-  DT_PROFILE_DIRECTION_ANY = DT_PROFILE_DIRECTION_IN | DT_PROFILE_DIRECTION_OUT | DT_PROFILE_DIRECTION_DISPLAY
+  DT_PROFILE_DIRECTION_CATEGORY = 1 << 3,  // categories will return NULL with dt_colorspaces_get_profile()
+  DT_PROFILE_DIRECTION_ANY = DT_PROFILE_DIRECTION_IN | DT_PROFILE_DIRECTION_OUT | DT_PROFILE_DIRECTION_DISPLAY | DT_PROFILE_DIRECTION_CATEGORY
 } dt_colorspaces_profile_direction_t;
 
 typedef struct dt_colorspaces_t
@@ -88,8 +91,10 @@ typedef struct dt_colorspaces_t
   // the current set of selected profiles
   dt_colorspaces_color_profile_type_t display_type;
   dt_colorspaces_color_profile_type_t softproof_type;
+  dt_colorspaces_color_profile_type_t histogram_type;
   char display_filename[512];
   char softproof_filename[512];
+  char histogram_filename[512];
   dt_iop_color_intent_t display_intent;
   dt_iop_color_intent_t softproof_intent;
 
@@ -108,6 +113,7 @@ typedef struct dt_colorspaces_color_profile_t
   int in_pos;                               // position in input combo box, -1 if not applicable
   int out_pos;                              // position in output combo box, -1 if not applicable
   int display_pos;                          // position in display combo box, -1 if not applicable
+  int category_pos;                         // position in category combo box, -1 if not applicable
 } dt_colorspaces_color_profile_t;
 
 int mat3inv_float(float *const dst, const float *const src);
--- a/src/develop/pixelpipe_hb.c
+++ b/src/develop/pixelpipe_hb.c
@@ -597,7 +597,8 @@ error:
 }
 #endif
 
-static void _pixelpipe_pick_from_image(const float *const pixel, const dt_iop_roi_t *roi_in, cmsHTRANSFORM xform_rgb2lab,
+static void _pixelpipe_pick_from_image(const float *const pixel, const dt_iop_roi_t *roi_in, 
+    cmsHTRANSFORM xform_rgb2lab, cmsHTRANSFORM xform_rgb2rgb,
     const float *const  pick_box, const float *const  pick_point, const int pick_size,
     uint8_t *pick_color_rgb_min, uint8_t *pick_color_rgb_max, uint8_t *pick_color_rgb_mean,
     float *pick_color_lab_min, float *pick_color_lab_max, float *pick_color_lab_mean)
@@ -647,11 +648,36 @@ static void _pixelpipe_pick_from_image(const float *const pixel, const dt_iop_ro
           = picked_color_rgb_max[i] = pixel[4 * (roi_in->width * point[1] + point[0]) + i];
   }
 
-  for(int i = 0; i < 3; i++)
+  // Converting the display RGB values to histogram RGB
+  if(xform_rgb2rgb)
   {
-    pick_color_rgb_mean[i] = round(picked_color_rgb_mean[i] * 255.f);
-    pick_color_rgb_min[i] = round(picked_color_rgb_min[i] * 255.f);
-    pick_color_rgb_max[i] = round(picked_color_rgb_max[i] * 255.f);
+    // Preparing the data for transformation
+    float rgb_ddata[9];
+    for(int i = 0; i < 3; i++)
+    {
+      rgb_ddata[i] = picked_color_rgb_mean[i];
+      rgb_ddata[i + 3] = picked_color_rgb_min[i];
+      rgb_ddata[i + 6] = picked_color_rgb_max[i];
+    }
+
+    float rgb_odata[9];
+    cmsDoTransform(xform_rgb2rgb, rgb_ddata, rgb_odata, 3);
+
+    for(int i = 0; i < 3; i++)
+    {
+      pick_color_rgb_mean[i] = round(rgb_odata[i] * 255.f);
+      pick_color_rgb_min[i] = round(rgb_odata[i + 3] * 255.f);
+      pick_color_rgb_max[i] = round(rgb_odata[i + 6] * 255.f);
+    }
+  }
+  else
+  {
+    for(int i = 0; i < 3; i++)
+    {
+      pick_color_rgb_mean[i] = round(picked_color_rgb_mean[i] * 255.f);
+      pick_color_rgb_min[i] = round(picked_color_rgb_min[i] * 255.f);
+      pick_color_rgb_max[i] = round(picked_color_rgb_max[i] * 255.f);
+    }
   }
 
   // Converting the RGB values to Lab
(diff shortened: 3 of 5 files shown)


TurboGit:
Won't this be better under the softproof & gammut  buttons? We already have here the screen profile. It seems awkward to me to have this under the over/under expose. Can you comment on this choice? Thanks.

edgardoh:
Are you talking about the location of the combo box for the overexposed
profile or the fact that there's a different setting for it and the
softproof?

TurboGit:
I'm speaking about the location of the new setting you have added. The new selection is done in the over/under expose dialog. I think it would be better if placed into the softproof & gamut dialog.

I've just built and tested and I see that the name for the new profile is "overexposed profile". So it looks like it is in the right place. But... then I fail to see a relation between the new "overexposed profile" and the picker or histogram. Can you clarify?

edgardoh:
Regarding the position of the combo, what's wrong for me is the display
profile under the gamut & softproof, it should have an independent button,
like the lighttable, but that's for another PR.

The overexposed/color picker/histogram is a matter of choice. I think they
are related and should work with the same profile, having a profile for
each one is overkill. But maybe this is not what you are asking...

TurboGit:
I'm just trying to understand. You did not answered one of my question.

Why the new profile is named "overexposed profile"? As I understand it works for picker and histogram, so I don't see the relationship.

edgardoh:
It works for the over/under exposed too, that's why is under that button.

It was requested many times to have a setting for the profile for the
histogram & color picker, I never heard about the overexposed, which is
strange to me because as I said, they are related, so I use the same
profile for the 3 features. Otherwise, you can check the histogram for
over/under exposed areas, use the colorpicker to see the exact value, but
when using the overexpose button you will get different results.

I agree that is not clear that the setting works for the 3 features, the
user that see this for the first time will think that it applies only for
the overexposed button. But this is just UI, and I thought that is better
to have this promoted as soon as possible so it can be tested and in time
make the UI more clear.

Right now the only options that I see are to add a combo on the colorpicker
section (in synch with the current one) and/or some text on the overexposed
profile. I don't really like any of those.

Does this makes sense?

TurboGit:
Ok I'm starting to understand now.

Another question if you don't mind (and possibly a stupid question as I'm no expert in this area), is this a kind of "working" profile? I mean, is that needed because most of the pixelpipe is Lab? I'm trying to understand why there is no such profile in other software?

edgardoh:
The working profile is where the edits are performed, in dt is mostly lab,
but also happens in camera rgb, prophotoRGB and the thing that colorout
returns.

This is independent from the working profile and the display profile, and
that's the point. One of the scenarios is that the user wants to check the
pixel values/overexposed areas/histogram on the image as it will be when
exported. The colorout returns an image in display profile, so that can't
be used, and the profile in colorout can be overwrite on export, so that
setting can't be used, so we add this new setting.

You can also check why people ask for it here:
https://www.mail-archive.com/darktable-user@lists.darktable.org/msg06952.html

No idea if other programs have something similar or if they don't have it
why.

parafin:
I don't get why we need another colorspace profile setting - there is already output profile (not display) in colorout. The fact that it can be overridden when exporting doesn't matter.

edgardoh:
> I don't get why we need another colorspace profile setting - there is
> already output profile (not display) in colorout. The fact that it can be
> overridden when exporting doesn't matter.
>
It does matter. Is the same concept as the gamut check, why it makes sense
to have a profile for the gamut but not for
overexposed/histogram/colorpicker?

parafin:
I guess it makes sense in case of CMYK printer profile, which I think can't be used as output profile. Otherwise I don't see why it should be different either.
You select output profile and develop with it in mind, so it makes sense to see histogram/overexposed in that colorspace, why would it need to be different? What workflow do you imagine where 3 separate settings for colorspace (4 if you count display profile) is needed?

edgardoh:
> I guess it makes sense in case of CMYK printer profile, which I think
> can't be used as output profile. Otherwise I don't see why it should be
> different either.
> You select output profile and develop with it in mind, so it makes sense
> to see histogram/overexposed in that colorspace, why would it need to be
> different? What workflow do you imagine where 3 separate settings for
> colorspace (4 if you count display profile) is needed?
>
I don't, this is based on some requests and I thought it was a good
solution. The problem with the output profile, as I said, is that can be
overwrite, to check with that other profile you'll need to change the
setting on the output profile.
I have also considered to use the softproof profile, but it doesn't have
the display profile option (I don't know why, so I won't change it without
knowing) and also I'm not sure if there's a scenario where it makes sense
to have separate settings for  softproof and overexposed, and I didn't want
to introduce a limitation.

My main goal here is to have a way to configure the profile for the
overexposed/histogram/colorpicker, If the problem with this feature is only
the extra profile that's easy to change, I can use any of the already
existing profiles (as long as the display option is available), but as long
as I'm playing with this I would like to make it useful for as many users
as possible.

aurelienpierre:
> You select output profile and develop with it in mind,

Nope, that's exactly what you don't want to do, and this is the typical display-referred ICC workflow I'm slowly reverting in dt.

You want to develop for the visible locus, in an output-agnostic fashion, the scene-referred data. Then, as a last step, deal with target profiles and gamut mapping. So whether you output in sRGB for the web, CMYK for print, or DCI P3 for your brand new Apple screen, you are good anyway, and you don't care.

In this context, you might need **scopes** or **views** on data, that are in a specific color space. But they should'nt affect the pipe, otherwise they are not scopes, and you are not in a space-agnostic pipe anymore.

So, I think what @edgardoh proposes goes in the right direction.

But, in the other hand, I wonder if just taking the overexposition in XYZ wouldn't be better. Because values above 255/255 in RGB might be overexposed or out of gamut, you don't know. In that case, that overlaps with the gamut check. Well, if the 3 RGB values are above, it's set then. So, the exposure alert should take the luminance instead (Y channel of XYZ space).

edgardoh:
I'm discussing this in two threads and that's never a good idea, but it seems that there are many different uses for this, and more generic means not-as-easy-to-use.

What I take so far is that some people want to check things with the printer profile to know that when printed everything will be OK.
Others with the export profile, because the same reason.
Others before the export profile is applied, because what @aurelienpierre  says.
I don't want to tell people what or how they should do things and I do want that dt provides the tools so each one can do their edits.
I think that this approach allows for all the scenarios, it adds a couple of clicks more than an automatic thing, but I don't see a way around it.

@aurelienpierre , I should think a bit more about it, but when checking for overexpose I want to know if an individual channel is overexposed, not only the luminance.

TurboGit:
What I don't like in this is that we have now:
- output profile
- profile set in the export module
- and the overexposed profile

And if I'm not mistaken those are all connected together and ideally the same profile should be used in the 3. This is not appealing :( Or am I wrong.

edgardoh:
Don't forget about the soft proof profile!

It does seems like an overkill and maybe we should rethink the entire
profile thing.

The output profile is obviously needed.
The overwrite profile (on the export module) is a nice thing to have.
The soft proof & gamut check are very common.

Maybe we can consider merging the soft proof & gamut profile with the
overexpose one, as I said before I created a new one so I don't introduce a
limitation, but if everybody is OK with it I can do the change.

aurelienpierre:
> I should think a bit more about it, but when checking for overexpose I want to know if an individual channel is overexposed, not only the luminance.

If an individual channel is **clipped** (not overexposed), it's actually out of gamut. The proof is: change your white balance or desaturate, and you will be able to get that channel back (that might get ugly, but it's not the point). You don't need to change the exposure. Think about it : exposure = log2(luminance). Right now, the overexposition preview overlaps with the out-of-gamut (gamut **clipping**) for that reason. Put you image in black and white, and your overexposure preview is not the same anymore. Exposure doesn't change when you switch to monochrome…

> What I don't like in this is that we have now:
> 
>     * output profile
> 
>     * profile set in the export module
> 
>     * and the overexposed profile
> 
> 
> And if I'm not mistaken those are all connected together and ideally the same profile should be used in the 3. This is not appealing :( Or am I wrong.

They are not necessarily connected. I agree that the option in the output profile module is in fact useless, because the output profile of the pipe is actually your display profile as long as you are in the software, and you can select it again at export time. So this one is even misleading.

aurelienpierre:
> Maybe we can consider merging the soft proof & gamut profile with the overexpose one, as I said before I created a new one so I don't introduce a limitation, but if everybody is OK with it I can do the change.

I don't think that's a good idea. Gamut has to do with chrominance/saturation. Exposure has to do with luminance.

edgardoh:
> I don't think that's a good idea. Gamut has to do with
> chrominance/saturation. Exposure has to do with luminance.
>
Sure, but when you check for those, do you use different profiles?
Let's say you check overexposed/histogram/colorpick in linear rec 2020, do
you check the gamut with other profile?

aurelienpierre:
>> Right now, the overexposition preview overlaps with the out-of-gamut
> What do you mean by overlaps?

The same feature is provided by 2 different options, because the overexposition is not limited to overexposed pixels, but shows anything clipped. 

> Sure, but when you check for those, do you use different profiles? Let's say you check overexposed/histogram/colorpick in linear rec 2020, do you check the gamut with other profile?

The gamut is related to the output color profile only. So we need to be able to get colors at the extreme end of the pipe, and check what is outside the [0, 255] range (gamut clipping). These output profiles/spaces will generally have a built-in gamma/OETF.

The histogram and colorpicks are related to the pipe, in which you **control** settings in Lab and various RGB spaces, and might need to pick colors in the same space you input the settings. These RGB spaces will be linear, and different from the output profile.

The output color profile defined in the module is purely informative and possibly misleading since it is defined once again in the export box and does not affect the preview, which is taken care by the display profile. So I think that control should go.

The overexposition is color-space agnostic and should only deal with luminance, so overexposed areas are overexposed no matter if you thange the saturation. XYZ seems fine to do that.

All in all, we need 4 profiles:

1. input to pixelpipe:
    * linear matrice if linear raw input
    * linearization transfer function if gamma-encoded input (so the pixelpipe is linear)
2. pixelpipe to output **file**: used to export, softproof, gamut check - useless inside the darkroom except for softproof control
    * ICC profile (possibly OCIO transform one day)
3. pixelpipe to display: used to send previews to screen
    * ICC profie (possibly OCIO transform one day)
4. pixelpipe to control view: used for histograms, colorpickers, overexposure
    * linear matrice

parafin:
@aurelienpierre, my comment is based on current system in DT. With what you're doing "output profile" should be just changed to some working space RGB color profile like ProPhoto or whatever is chosen. I don't see a need to see histogram in any other profile than that, and from the comments I see that it's like that in other software too. Having overexposed indicator work differently than histogram is IMHO misleading. I'm yet to see the scenario where it's needed, just because some people asked for it doesn't mean it's useful (unless they explain why they want it and it makes sense). Adding option for every possible thing you can change in the code is not the right way to design software. In this case it will also make things hard to use, since user will have to change the same setting in many places instead of one (and it's easy to forget this).

aurelienpierre:
@parafin forget about the output color profile. This thing has nothing to do in the darkroom, in fact changing the profile in the module does not change the preview at all, so that screams bad design. The output profile is relevant only during export, and everything before in the pipse should be fully output-agnostic.

The main colorspace in which you need to see the histogram is the one of your pipeline, e.g. of your raw data before they get messed up by gamma and silly ICC bits.

But… Once in a while, you need to keep an eye on the gamut repartition in a given color space. Nowadays, pictures are mainly used for print and display on screen on the web, so that makes at least 2 different color spaces to check. So you need a view on the gamut clipping.

About that clipping… The gamut check screams where clipping occurs. Fine. Then what ? 

Having the RGB histogram in the same space would allow to diagnose which channel is actually clipped in said space. So, instead of guessing and maybe desaturating everything, you could just have a look on the histogram in a different space and get a sense of what is clipping.

Then… As of now, an overexposure indicator that only shows clipped RGB channel is mostly redundant with the gamut clipping indicator. Do this test : open a picture with clipped highlights, enable the overexposure, turn the pic into monochrome (desaturation is fine), and suddenly your image has far less clipped areas. That's not ok. The exposure of a picture does not change when you turn it B&W, so that tells you the overexposure is not really showing you overexposition, but any channel clipping. So you need to show it in an XYZ space, and show clipping only on the Y channel (the luminance).

As of softproofing, the whole concept is foolish and researches show that even the more complicated color-adjustments algorithms (CIECAM02 and derivative by Fairchild and al.) fail to give a good preview, mainly because a screen is back-lit and works in additive lights, whereas prints work in subtracting pigments.

Even if some call that "painting by the numbers", and because softproofing is doomed by nature, the gamut clipping checks are super important, and I think the histogram is another scope on that issue.

That might be bad software design, but having a clean soft is something I care about when it's working as it should. Image quality first.

parafin:
So how many colorspace settings do you think are needed (aside from display/output which shouldn't affect the processing at all)? I see only 2:
1. working space RGB profile, probably non-configurable or at least not in bottom panel, currently it's display profile;
2. gamut check profile - as an idea: when gamut check is active - recompute histogram and overexposed in that profile instead of working space.
Again I don't see the need for a separate setting just for overexposed, it doesn't make any sense.

aurelienpierre:
you need the working RGB space (which should be configurable, because every RGB space has different drawbacks when it comes to pixel pushing), the gamut check profile, and the display profile (which can be useful in multi-display setups).

> Again I don't see the need for a separate setting just for overexposed, it doesn't make any sense.

Did you understand the difference between channel clipping and overexposure ? Do you agree that, from a physical point of view, desaturation should not change the overexposed areas because color is different from luminance ?

You don't need a separate setting for overexposure, you just need to compute it in the right space.

edgardoh:
Let's try another approach,

In any workflow we have the following profiles:

input
working
export
display
print

Some programs may have it hard-coded, others configurable, but they're always there.

dt don't have a working profile, that's added on PR 1841, until then the user will have to select it manually.

What I suggest is:

-rename the softproof (or gamut) profile with print profile
-rename my new overexposed profile with view profile (or some better name)
-create two new categories in addition to DT_COLORSPACE_DISPLAY: DT_COLORSPACE_PRINT and DT_COLORSPACE_EXPORT
-the profile view combo box will have:
display profile
export profile
print profile
(all the user selected profiles from out)

This way the design is more clean, and the user is a one click away from viewing the data in any point of the workflow he wants.

aurelienpierre:
Printing and exporting profiles are just output profiles for files. I don't think we need to split the feature, they work the same, but for different outputs. 

Then, again, using whatever RGB is foolish for overexposure.

The whole ICC system is built around the notion that 100 % luminance is 100 Cd/m². Over-exposure is achieved when one pixel is above 100 % luminance. Luminance is computed from the RGB values like Y = 0.22 * R + 0.76 * G + 0.02 * B (coeffs from memory). So, overexposure is achieved when Y > 1. 

Whatever you show in RGB is not (only) overexposure, but also gamut clipping.

edgardoh:
> Printing and exporting profiles are just output profiles for files. I
> don't think we need to split the feature, they work the same, but for
> different outputs.
>
Because they work for different outputs is why we need to split it.

aurelienpierre:
Sure, they work for different outputs, but it's just a label at that point. They are completely interchangeable, you could save a JPEG or a TIFF in printer space RGB as well as in sRGB or Adobe RGB, that wouldn't make any difference regarding the actual color correction.

Or maybe you mean splitting CYMK and RGB output profiles ? 

I'm not entirely convinced CYMK profiles are useful in darktable, since all consumer printers drivers happily take RGB as input and convert internally to CYMK, possibly better than Little CMS. Maybe @TurboGit has another opinion here, after having done the printing UI.

parafin:
1. @aurelienpierre, I didn't comment on how overexposed (both ordinary and raw) should work and whether it's broken now, because I haven't looked at the code. It should be defined what is considered overexposure, and looking at DT user manual I guess you are correct that current implementation is broken. Raw overexposure on the other hand shouldn't be affected by any module at all, since it's described as working on raw input.
2. @edgardoh, as for profile settings - I'm not convinced either that export and print profiles should be separated. And I didn't understand where do you want to put these preferences (view, export. print). Also I don't think it's useful to see overexposed areas or gamut check for display profile - display profile is only for displaying stuff, nothing else.

TurboGit:
> Printing and exporting profiles are just output profiles for files. I don't think we need to split the feature, they work the same, but for different outputs.

This sounds awkward to me. A print profile expect a file exported with a specific profile. That is, when creating the printer profile you need to specify the profile for the file that will be printed. That's why in the print module we have the export profile and the printer profile.

The TurboPrint module also expect that. So we can't say that this is dt specific.

edgardoh:
I think we are talking about different things, so let's first try to be on
the same page.

This PR is about allowing the user to see the data at any point of the
workflow (as described above). Right now this is hard-coded, and that is a
bad design, things should never be hard-coded if can be avoided. That's why
they are working on the UI refactor, dt is a raw developer, how we see the
data is far more important than the UI, so there should be no discussion
about this.

With my last proposal, if I want to see the how the data will look like
when printed I select print profile from the profile view combobox.
if I want to see the how the data will look like when exported I select
export profile from the profile view combobox.
if I want to see the how the data look like before any profile is applied
(that will be the concept of the working profile) right now I have to
select it manually from the list on the profile view combobox, Eventually
we will have a working profile entry.

The view profile will affect the histogram, color picker, overexposed,
gamut and soft proof. I propose this because is reasonable to assume that
if I'm checking the gamut in one profile I'll want to check the pixel
values for areas out-of-gamut on that same profile, and probably the
histogram too.

I'm sure that there are people out there that will prefer to have separate
setting for the gamut, and others may want individual setting for each one.
But this design is cleaner, easy to use, and probably work for most users.
And changing from one profile to another is very easy, just select it from
the view profile. We do the same when editing, we search for the module,
expand it, and adjust the setting, and that happens far more often than
changing the view profile.

edgardoh:
@TurboGit , not sure I understand, I don't know the print module and maybe I'm thinking at the naive version. If I want to generate the pixel values that are sent to the printer (for display on the colorpicker for instance), if not enough to apply to the image (whatever the colorspace of the image) the print profile? There's some intermediate steps required?

parafin:
@edgardoh, so I'll ask again - where do you propose to put UI controls for selecting each of those profiles? colorout IOP? gamut check menu? What would to be the steps for user to see gamut check for, say, "view" profile?

edgardoh:
> @edgardoh <https://github.com/edgardoh>, so I'll ask again - where do you
> propose to put UI controls for selecting each of those profiles? colorout
> IOP? gamut check menu? What would to be the steps for user to see gamut
> check for, say, "view" profile?
>
My first thought on the UI will be:
-add the view profile combobox on the darkroom, to the left of the
rawoverexpose button
-remove the right-click popup on the gamut and soft proof buttons
-add a new, dedicated button, like the one on the lighttable, with the
combobox for display profile and print profile

The view profile combobox will have the following entries:
-display profile: it will take the value from the display profile combobox.
This is just for backward compatibility, the default will be export anyway.
-export profile: if selected it will take the value from the colorout
module (the value selected on the output color profile combobox, not the
actual profile applied to the image)
-print profile: if selected it will take the value from the print profile
combobox.
-a list of profiles taken from out, like the gamut check is doing right now.

Gamut, soft proof, histogram, color picker, they all will work with the
profile select on the view profile combobox.

If I want to gamut check using the print profile, I select print profile
from the view profile combobox and click on the gamut check button.
Same if I want to soft proof.
While the view profile combobox has print profile selected the color picker
and histogram will display the data using that profile.

This makes the bottom of darkroom a bit more busy, but will make more
obvious the profile the user is working with and easier to change it if
needed.

parafin:
It's more clear now, thanx. Some notes:
1. you are moving display profile from colorout iop to bottom panel, might be a little hard to do currently;
2. print profile is just a helper for user so he doesn't have to look for it in the whole list of out profiles, I guess it might be useful;
3. on the other hand there's no such helper for "view" profile (working profile) - say user wants histogram in ProPhoto space, but after he selects print profile in view combobox in order to get back to ProPhoto he will have to browse through whole list of out profiles to find his choice of "view" profile;
4. also you are complicating soft-proofing/gamut-check - right now user selects gamut check profile once and then can toggle gamut check with just one click, with your suggestion he will have to select another profile in view profile combobox, then switch on gamut check (and reverse procedure for turning it off) - it's a lot more time-consuming.

3 might not need fixing if 4 is fixed. 4 can be fixed by leaving alone gamut check profile and switching view profile to that setting automatically when user switches on soft-proofing or gamut check. This will result in us having these settings for color out profiles:
1. display - needed
2. view/working - useful for fixing histogram until real fix gets merged
3. export - needed
4. print - just a helper for faster switching between export and print profile in gamut check
5. gamut check - just a setting for gamut-check/soft-proofing, will be set to one of the 2-4 most of the time

My question is the following - can't this PR wait until real fix for RGB working space is merged? It will change the meaning of view/working colorspace somewhat and make most of the code in this PR obsolete (aside from UI changes).

edgardoh:
> It's more clear now, thanx. Some notes:
>
>    1. you are moving display profile from colorout iop to bottom panel,
>    might be a little hard to do currently;
>
> Not sure I get this, colorout don't have an option "display profile", it
uses it internally while in darkroom, but I'm not changing that, the image
displayed will still be in display profile. What may be in a different
profile are the color picket, etc..

>
>    1. print profile is just a helper for user so he doesn't have to look
>    for it in the whole list of out profiles, I guess it might be useful;
>
> Correct

>
>    1. on the other hand there's no such helper for "view" profile
>    (working profile) - say user wants histogram in ProPhoto space, but after
>    he selects print profile in view combobox in order to get back to ProPhoto
>    he will have to browse through whole list of out profiles to find his
>    choice of "view" profile;
>
> Correct, its a limitation. PR 1841 introduces a working profile, until
then it will be hand picked.

>
>    1. also you are complicating soft-proofing/gamut-check - right now
>    user selects gamut check profile once and then can toggle gamut check with
>    just one click, with your suggestion he will have to select another profile
>    in view profile combobox, then switch on gamut check (and reverse procedure
>    for turning it off) - it's a lot more time-consuming.
>
> I don't see it as a lot more time consuming, is just a click more. If I
want to gamut check then I select print profile (or any other profile) and
click on gamut check, to disable gamut check I click again on gamut check
and depending on what I want I select another profile on the view profile
combobox.
Note that you don't need to be on a specific view profile to do your edits,
this setting only affects the color picker, histogram, etc, so you may do
your edits with view profile=printer and the only difference is that you'll
see the histogram and colorpicker values on the printer profile. And maybe
that's what you want.
I'm proposing it like this because people may want to gamut check things in
profiles other than print. Let's say I send an image for printing, I want
to gamut check the image I send (with export profile) and the printed image
(with print profile)

>
>
> 3 might not need fixing if 4 is fixed. 4 can be fixed by leaving alone
> gamut check profile and switching view profile to that setting
> automatically when user switches on soft-proofing or gamut check. This will
> result in us having these settings for color out profiles:
>
>    1. display - needed
>    2. view/working - useful for fixing histogram until real fix gets
>    merged
>    3. export - needed
>    4. print - just a helper for faster switching between export and print
>    profile in gamut check
>    5. gamut check - just a setting for gamut-check/soft-proofing, will be
>    set to one of the 2-4 most of the time
>
> Not sure I'm following this, gamut profile no longer exits, it has been
replaced by print profile. And by switching the view profile when gamut
checking is selected I will be limiting the gamut check to one profile.

>
>
> My question is the following - can't this PR wait until real fix for RGB
> working space is merged? It will change the meaning of view/working
> colorspace somewhat and make most of the code in this PR obsolete (aside
> from UI changes).
>
I don't know if you have a specific development in mind, but I'm doing this
taken into account PR 1841, it will work just fine with it.

parafin:
>> 1. you are moving display profile from colorout iop to bottom panel, might be a little hard to do currently;

> Not sure I get this, colorout don't have an option "display profile", it uses it internally while in darkroom, but I'm not changing that, the image displayed will still be in display profile. What may be in a different profile are the color picket, etc..

Right, sorry, it has already been moved some time ago, nevermind about that point.

>> 2. print profile is just a helper for user so he doesn't have to look for it in the whole list of out profiles, I guess it might be useful;

> Correct

>> 3. on the other hand there's no such helper for "view" profile (working profile) - say user wants histogram in ProPhoto space, but after he selects print profile in view combobox in order to get back to ProPhoto he will have to browse through whole list of out profiles to find his choice of "view" profile;

> Correct, its a limitation. PR 1841 introduces a working profile, until then it will be hand picked.

>> 4. also you are complicating soft-proofing/gamut-check - right now user selects gamut check profile once and then can toggle gamut check with just one click, with your suggestion he will have to select another profile in view profile combobox, then switch on gamut check (and reverse procedure for turning it off) - it's a lot more time-consuming.

> I don't see it as a lot more time consuming, is just a click more. If I want to gamut check then I select print profile (or any other profile) and click on gamut check, to disable gamut check I click again on gamut check and depending on what I want I select another profile on the view profile combobox. Note that you don't need to be on a specific view profile to do your edits, this setting only affects the color picker, histogram, etc, so you may do your edits with view profile=printer and the only difference is that you'll see the histogram and colorpicker values on the printer profile. And maybe that's what you want. I'm proposing it like this because people may want to gamut check things in profiles other than print. Let's say I send an image for printing, I want to gamut check the image I send (with export profile) and the printed image (with print profile).

I disagree here. First of all it's more than one more click - you also need to find where to click. Secondly I don't think always seeing histogram in print color profile is a good idea - I assume histogram is useful to look at during picture development, so it should be output-independent. Printers are interchangeable, user can have several choices of printers, so while it makes sense to check how it will look on specific output device once in a while, one doesn't develop pictures just for that device IMHO. Basically if user wants to develop the picture like that he will just keep soft-proofing on all the time, which will switch histogram to that color profile. As for last point about gamut checking - it's resolved by keeping separate gamut check profile setting.

>> 3 might not need fixing if 4 is fixed. 4 can be fixed by leaving alone gamut check profile and switching view profile to that setting automatically when user switches on soft-proofing or gamut check. This will result in us having these settings for color out profiles:
>> 1. display - needed
>> 2. view/working - useful for fixing histogram until real fix gets merged
>> 3. export - needed
>> 4. print - just a helper for faster switching between export and print profile in gamut check
>> 5. gamut check - just a setting for gamut-check/soft-proofing, will be set to one of the 2-4 most of the time

> Not sure I'm following this, gamut profile no longer exits, it has been replaced by print profile. And by switching the view profile when gamut checking is selected I will be limiting the gamut check to one profile.

It exists in my proposal to keep it;) So no limitations if it stays.

>> My question is the following - can't this PR wait until real fix for RGB working space is merged? It will change the meaning of view/working colorspace somewhat and make most of the code in this PR obsolete (aside from UI changes).

> I don't know if you have a specific development in mind, but I'm doing this taken into account PR 1841, it will work just fine with it.

I retracted that question.

edgardoh:
Let's try the following scenario: I have a picture that I want to post on
facebook, my personal web page, and also print, so

First I do a generic edit on some wide gamut profile, let's say prophoto

1-I set the output color profile to prophoto and do all may edits. When I'm
happy I check gamut, I do this in prophoto, right now I don't have any
other profile, so I don't need to change anything

2-I create a duplicate to work on my facebook version. I set the output
color profile to sRGB, fine tune it until I'm happy and again check gamut.
I want this in sRGB because that's what I'll use to output.

3-Same for my personal web page, but instead of sRGB I use a custom
profile. All the steps are the same.

4-I'll work on my print version now, I don't own a printer so I will
generate a tiff that I'll to print. I set the output color profile to
prophoto and the print profile to the one that will be used to print. I
fine tune the image and check the gamut. I didn't change anything so the
gamut will be checked on the output profile, and that's what I want.
When I'm happy I'll check how it will look like when printed, so I change
(for the first time) the view profile to print, check the gamut and fine
tune again if there is any out-of-gamut issue. This last fine tune I want
to do it in print profile, so I can check the data and histogram on the
same profile.

This seems very reasonable to me.

But also, my current implementation has the gamut and histogram separate,
and no one seem to like it, so I have to ask, why my current implementation
is not OK?

aurelienpierre:
1 - checking gamut in Prophoto makes no sense, this space has imaginary colors that are out of the visible locus.

2 - you really don't want to duplicate your edits depending you destination/output color space. That's the whole point of an output-agnostic pipeline. What you want is edit just once, and check that LittleCMS gamut mapping does its job properly in all the output spaces you need.

Please read https://github.com/darktable-org/darktable/wiki/Developer's-guide#understanding-color-and-color-management

edgardoh:
This is not my workflow but an example that I have manufactured because it
seems complex.

If you want, you can describe your typical workflow and see how my proposal
fits into it.

aurelienpierre:
The ideal workflow would not make use of gamut checking at all.

ICC profiles are numerical descriptions of the properties of color spaces (input and output). Using these descriptions, **c**olor **m**anagement **s**ystems (Little CMS, Argyll CMS, etc.) perform automatic color mapping (basically, RGB coordinates rescaling) giving precedence to luminance (relative colorimetric intent) or color (saturation intent) or homogeneity (perceptual intent), depending on what the user wants. Usually, these strategies work good enough, and photographers don't need to worry about gamut.

However, in some cases, they fail (for example, blue and magenta artificial spot lights, at events and concerts, are garanteed to fail big time). If the most part of the picture is out of gamut, they give unexpected results, and this means you need to take manual control. "Out of gamut" makes color geeks freak out, but that only means you might get a solid saturated color patch instead of a smooth color gradient (because we use a CMS). That is only a concern if the main subject of the picture is mostly out of gamut or if hard transitions (halos, rings, etc.) happen at the in/out-of-gamut boundaries. And that triggers the need for scopes to get a diagnostic.

So, a safe workflow would involve:

1. checking that the CMS gamut mapping works as expected
2. ensuring most of the areas of interest (main subject) are in gamut
3. diagnosing where you were reckless and caused damage in your retouch.

This affects only the fine-art people who are, generally, educated in these topics (sometimes mis-educated on the web, but that's another concern). Meaning this is a place where trying to make things simple will hurt more than it will help : settings should be transparent (e.g. no "general audience" simplified labels, use the accurate technical terms).

Using large gamut RGB spaces in the pipe is important only when said pipe is in 8-bits or 16-bits integers, meaning clipping happens out of the 0-255 or 0-65535 range. darktable works in 32 bits float unclipped, so any linear and well-behaved RGB space would do (REC709, REC2020, ACES P0, etc.). Besides, the true PhoPhoto RGB space has an OETF (gamma) of 1.8, whereas darktable's implementation is linear (so it's not a real ProPhoto RGB). But what matters is that the RGB space behaves consistently when you push pixels in it, meaning its primaries retain some spectral consistency (so REC2020 or ACES P0 seem to be the best choices here, for now). 

Back to the workflow… what you want to do on a picture is:

1. adjust the exposure, contrast and lightness in a linear space while retaining the scene-referred colors. You want to check that you don't get overexposure, so you do it looking at a scope that shows the Y channel clipping.
2. adjust the colors (hues, saturation) in a linear or log-encoded space, while staying in-gamut. At that point, the gamut is the visible locus, which is shorter than large RGB spaces like ProPhoto or ACES P0 or Lab. So, for now, checking for the gamut of the pipe space will do (with caution), but at some point, we will need some way to squeeze the human visible locus in that.
3. check that the CMS does a proper job remapping colors from the pipe space/visible locus to the output profiles you need, and possibly tweaking it (there are very few gamut mapping options and output display transforms in dt for now, that is a way for improvement for the years to come). For now, if the gamut mapping fails in one particular output space, you need to adjust the saturation lower in the general pipe, which is not really clean since it affects all the possible outputs when there is no need to (I have a "clever" gamut-mapping in IPT-HDR space incoming).
4. enjoy your 2 minutes of color-safe editing and have fun during the 8 additional minutes you would have wasted in darktable 2.6 recovering colors.

parafin:
@edgardoh, I'm confused about your scenario, because you say 4 is the first time you have to change view profile, but it's not, you have to change it for 2 and 3 to get gamut checking in needed color space (in your description you changed only output colorspace). Secondly what exactly are you tuning in steps 2 and 3? Changing output color profile won't change anything visible in darkroom unless you change "view" profile or enable soft-proofing. The problem with your described workflow is that it will become broken after RGB working color space gets implemented, then changing "view" profile will affect also the processing and output of the pipe. That's not what you want. And the idea is to not have yet another one color space setting. You get histogram/color picker in the working colorspace unless you enable soft-proofing/gamut check.

edgardoh:
>
> @edgardoh <https://github.com/edgardoh>, I'm confused about your
> scenario, because you say 4 is the first time you have to change view
> profile, but it's not, you have to change it for 2 and 3 to get gamut
> checking in needed color space (in your description you changed only output
> colorspace).
>
I don't need to change anything until 4 because the output profile is the
default profile on the view profile, so by changing the output profile I'm
also changing the view profile

> Secondly what exactly are you tuning in steps 2 and 3? Changing output
> color profile won't change anything visible in darkroom unless you change
> "view" profile or enable soft-proofing.
>
If view profile is set to output profile, changing output profile will
change histogram, color picker, gamut, etc.

> The problem with your described workflow is that it will become broken
> after RGB working color space gets implemented, then changing "view"
> profile will affect also the processing and output of the pipe. That's not
> what you want. And the idea is to not have yet another one color space
> setting. You get histogram/color picker in the working colorspace unless
> you enable soft-proofing/gamut check.
>
The working colorspace, no matter how is implemented, should affect things
only up to the output colorspace, once the image is transformed for output
(display, export or print) it shouldn't be affected by the working profile
anymore. Everything I'm doing here works on the final image (final
histogram, global color picker, gamut and soft proof) so the working
profile has nothing to do with it.

I give some more thoughts to your proposal, and for what follows you'll
have to check how the current implementation works, but it seems to me that
by adding the following to what's already done we get what you propose:

1-add to the overexposed profile the new categories output profile, gamut
profile
2-make gamut check take priority over overexposed profile, that is, when
gamut check is active color picker and histogram will work on the profile
selected on gamut
3-re-arrenge the UI to make it more user-friendly

If this is what you are suggesting I'm OK with it.

parafin:
I think I misunderstand the idea then. I thought that after working colorspace fix is merged we would want to have histogram in that working space, so taken before final transform to display/output profile. I think that's what other software like Lightroom is doing, isn't it? And your proposal is to have a separate setting for histogram/colorpicker/etc which by default is set to output profile?

About my proposal - yes. main idea is for gamut check/soft-proofing to override (when active) histogram/colorpicker colorspace with its setting.

edgardoh:
Let's take a couple of steps back. Right now the histogram and color picker collect the data from the image that is sent to the display, for short I call that the data is in display profile. We all agree that this isn't right. I also think that what's not right is that the profile is hard-coded, is not a good idea to hard-code things, so if the working profile is implemented and the histogram is always displayed in working profile that will be just wrong. I want to be able to check the data not only before the output profile is applied but also after, and for any export profile (file or print).
I'm sure that you want that too, you said that you edit with the output in mind, so you'll probably want to check things with the output profile, maybe not all the time, but I'm sure that you want the option (I know I do)

In order to do this we have (at least) this options
1-histogram, color picker, gamut and soft proof all work with the same profile
2-histogram and color picker work with one profile and gamut and soft proof work with another profile
3-there's an independent profile for each one of them

My current implementation is option 2 and that's also your proposal, so I say go with that. If at some point it is decided that histogram and color picker should have their own setting it will be some work, but for now this seems reasonable and is already an improvement on what we currently have.

When working profile is implemented it should be added to the list of options for the histogram and color picker because, again, hard-code it will be wrong.

If you agree with this we can continue fine tuning the design.

parafin:
I agree in general, at least for now. It can be revised later.

edgardoh:
@TurboGit , your call.

aurelienpierre:
2. looks fine to me too.

TurboGit:
Ok, I've reread most of the comments here. I can't say that I understand all the details but I agree that option 2 is a good compromise for now.

My feeling also that this will certainly not be easy to grasp for most users.

One side question. What will be the default for the picker/histogram profile? Will it be "same as gamut/softproof"? Something else? In other words, what would be a good default?

edgardoh:
In the case of this settings the default actually is what is used in case
something goes wrong, if the user selects a profile from file that
eventually dt can't read, it will fall back to the default, otherwise the
setting is "remembered".
Right now we don't have a working profile, so the output profile is
probably a good choice.

moy:
I only had a cursory look at the code and discussion, but if I understand correctly what the patch does is that it runs the whole pipeline, including conversion to display space, and then convert back to another space for picker and histogram. If so, this can be problematic in several ways:

* If modules appear after the output profile application, then the histogram and picker are still wrong (probably less "wrong" than using the display profile as is the case currently).

* If colors are out of the display profile's gamut, then they are clipped before being converted back to the picker&histogram profile, and therefore the end result can be wrong.

The "clean" approach would be to catch the image before the display profile, and then run it through the end of the pipeline using the picker&histogram profile. Obviously, the implementation of this would be much trickier.

My feeling is that this PR is a good step forward from the UI point of view (get the picker and histogram to be reasonably independent of the display is better than having them completely dependent of it), but I'm just wondering whether it's not a step backward from the internal architecture's point of view (i.e. if we get to the point where the clean approach can be implemented with reasonable effort, will the code from this PR make it harder?).

Anyway, again: I only had a cursory look, I may have missed the obvious and/or be completely wrong ;-).

edgardoh:
New version is ready.

I have moved the overexposed profile to the softproof popup and renamed it as histogram profile (tooltip says histogram and color picker).
I have also renamed the output profile on the output color profile iop as export profile, it does seems more accurate.

Here is how it works:
When in soft proof or gamut check, the histogram and color picker use the soft proof profile.
In any other case they use the option selected on histogram profile.

The histogram profile has this entries:
-export profile, if selected the profile selected on the output color profile iop is used
-softproof profile, if selected the softproof profile is used
-system display profile, idem
-a list of profiles from "out"

The overexposed always uses the histogram profile, is not mentioned anywhere in the GUI in case what @aurelienpierre suggest is ever implemented.

I don't print, so it will be helpful is someone that does can test it with an actual print profile and his own workflow.

edgardoh:
Color picker uses uint8 for the rgb values, so for clipped channels results are incorrect. I'll deal with that in a separate PR after this one.

TurboGit:
Looks good to me at this stage. I have tested it and all seems working as expected. It is indeed good to see the histogram adjusted when activating the softproof. Thanks!

aurelienpierre:
There is a problem with the circuitery for the over-exposed alert: the colors actually change in the image depending on the histogram profile. The image shown should be in display color space anytime. To compute the alert, the colorspace conversion should be done outside the pixelpipe (a dry-run, if you want), then, only the clipped pixels should be applied as an overlay on top of the image.

Also, changing the histogram profile while the gamut alert is on segfaults.

edgardoh:
I know, check the dev-list.

PR #2071, opened 2019-02-07
Updated Czech translation

milankni:
Thanks for heads up via email by Pascal O.


TurboGit:
As promised :) You're welcome and thanks for the translation.

PR #2074, opened 2019-02-08
Fix "find location" function

### LINKED ISSUE #2006: Broken "Find location" on Windows

peterbud:
Fixes #2006 

Similar changes are needed for using curl at facebook, piwigo and picasa upload, but I could not test those cases unfortunately. Facebook is not working for me in it's original state, and I don't have piwigo and picasa account, so I decided not to change them blindly.

In case somebody is maintaining those, it would be trivial to apply this solution before each `curl_easy_perform() `call

### DIFF
--- a/cmake/windows-macros.cmake
+++ b/cmake/windows-macros.cmake
@@ -206,6 +206,13 @@ if (WIN32 AND NOT BUILD_MSYS2_INSTALL)
     endforeach()
   endif(ISO_CODES_FOUND)
 
+  # Add ca-cert for curl
+  install(FILES
+      "${MINGW_PATH}/../ssl/certs/ca-bundle.crt"
+      DESTINATION share/curl/
+      RENAME curl-ca-bundle.crt
+      COMPONENT DTApplication)
+
 endif(WIN32 AND NOT BUILD_MSYS2_INSTALL)
 
 endfunction()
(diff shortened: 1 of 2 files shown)


TurboGit:
@peterbud : I understand for not changing blindly, but we are disable SSL check for the export module for now, so it works.

Now I'd like to fix this for better security. So I'll do the change, I'll check that all is well with picasa and piwigo (I have account). I'll change blindly for Facebook. And then I'll try to check that on Windows when 2.6.1 is released.

parafin:
> Why cant we use the system CA store?
> 
> Even more Why ... WHY is this on for all platforms? If windows needs it ... guard it accordingly.

It's also useful on macOS and shouldn't hurt on Linux either.

TurboGit:
@daris: the CA is only installed on Windows for now. So no problem. And if needed Linux and MacOS user could put a CA file at the right place. So all in all, currently it is a no-op expect on Windows.

techexo:
@TurboGit or @peterbud Shouldn't this PR be in the 2.6.1 milestone, as #2028 was, as it replaces it?

TurboGit:
Yes, it should be in 2.6.1. I'm waiting for more tests on Linux to ensure nothing is broken.

PR #2076, opened 2019-02-09
hazeremoval: fix median-of-three heuristic in quick_select

rabauke:
Fixes a bug in the implementation of the quick select algorithm and removes some unused code. Note that the bug affects the run time performance of the `quick_select` function only, not its correctness. Thus, this patch does not affect results of the hazeremoval module.


TurboGit:
Thanks.

PR #2079, opened 2019-02-09
fix some format string related issues

rabauke:
Among others, this pull request also reverts a recent change in `src/chart/main.c`, which I submitted. The format string in the `fscanf` function was actually ok. I was not fully aware of the meaning of an asterix in an `fscanf` format string. (An optional starting asterisk indicates that the data is to be read from the stream but ignored (i.e. it is not stored in the location pointed by an argument).) Reading to the end of line via `fscanf` is probably smarter than my `fgetc` approach.

### DIFF
--- a/src/libs/camera.c
+++ b/src/libs/camera.c
@@ -355,7 +355,7 @@ static void _expose_info_bar(dt_lib_module_t *self, cairo_t *cr, int32_t width, 
   pango_font_description_set_absolute_size(desc, fontsize * PANGO_SCALE);
   pango_layout_set_font_description(layout, desc);
   char model[4096] = { 0 };
-  sprintf(model + strlen(model), "%s", lib->data.camera_model);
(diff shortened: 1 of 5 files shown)


src/libs/camera.c
```diff
@@ -355,7 +355,7 @@ static void _expose_info_bar(dt_lib_module_t *self, cairo_t *cr, int32_t width,
   pango_font_description_set_absolute_size(desc, fontsize * PANGO_SCALE);
   pango_layout_set_font_description(layout, desc);
   char model[4096] = { 0 };
-  sprintf(model + strlen(model), "%s", lib->data.camera_model);
+  snprintf(model, strlen(model), "%s", lib->data.camera_model);
```
TurboGit: Nice one :) with model being empty the strlen() was a no-op!

TurboGit:
Thanks.

PR #2082, opened 2019-02-10
Add support for image renaming to the C and Lua APIs.

schwerdf:
This pull request modifies the C and Lua APIs to allow renaming of images during the move and copy processes, obviating the need to remove and re-import renamed images (thereby discarding group information, database IDs, etc.) or manually hack the database.

Specifics:

* A new C function `dt_image_rename()` has been added. This behaves similarly to `dt_image_move()`, but has an additional parameter, a string `newname`, containing a new filename that the image should be given when it is moved to the directory indicated by `filmid`. A test is performed to ensure that `newname` is a "basename" (no path separators, `.`, `..`, etc.).

* A new C function `dt_image_copy_r()` has been added, bearing an analogous relation to `dt_image_copy()`.

* The Lua API functions `darktable.database.move_image` and `darktable.database.copy_image` have been given an additional (optional) string parameter, which if provided will cause it to be passed to `dt_image_rename()` or `dt_image_copy_r()` respectively. If it is not given, the former behavior is retained.


TurboGit:
Thanks for starting this! I have two quick comments:

- I'd like dt_image_copy_r() to be renamed dt_image_copy_rename() more explicit. "r" here may means reverse (like in g_strrstr) or whatever.

- You've not handled the renaming of the XMP xmpMM:DerivedFrom tag which reference the RAW file. As it is now renamed you need to rename it in all XMP (main one and all duplicates).

TurboGit:
What about the XMP references? Do you plan to work on that or should I close this PR?

schwerdf:
The `xmpMM:derivedFrom` attributes seem to be getting properly updated in commit 850e5a4, after I added the line to `dt_image_rename()` to trigger sidecar updates. Is there something else that needs to be done for that?

TurboGit:
Just to be sure. This does not add renaming on dt but will provide this ability to Lua, right? And you plan to add such plug-in?

schwerdf:
Yes, this only modifies the C and Lua APIs, not the Darktable GUI.

As to implementing this functionality in the GUI, I would be willing to take on the task, but it would require many more design decisions than this change, since batch renaming is a deal more complex than batch copying or batch moving.

TurboGit:
@schwerdf : sorry for the late reply. so let's do this in two steps. I'll merge this when the conflicts will be resolved. And then please can you open an issue to discuss the design for the renaming? Thanks.

schwerdf:
I merged the most recent changes to eliminate the conflicts, and opened an issue (#2247) for discussion on the GUI question.

src/common/image.c
```diff
@@ -1560,10 +1625,23 @@ int32_t dt_image_copy(const int32_t imgid, const int32_t filmid)
     g_object_unref(src);
     g_clear_error(&gerror);
   }
+  if(oldFilename)
```
TurboGit: Not needed g_free(NULL) does nothing.

src/common/image.c
```diff
@@ -1560,10 +1625,23 @@ int32_t dt_image_copy(const int32_t imgid, const int32_t filmid)
...
+  {
+    g_free(oldFilename);
+  }
+  if(newFilename)
```
TurboGit: likewise

schwerdf:
I took out the conditionals around the `g_free` calls, but I could not reproduce any instance of an image being moved in the database when its file did not exist on disk.

My understanding was that it is practically a necessity for the move function to fail _silently_ in that case (as it currently does, only printing a line to the console) due to the way it handles moving multiple versions of an image.

TurboGit:
Indeed, there is no move if the image does not exist but there is no message either. To reproduce:

- create a test dir
- put one image on it
- launch dt and import this test dir
- quite darktable
- using a file explorer rename the test dir
- restart darktable
- go to test dir, select image and click [move]
- select the dest dir
- dt shows now the dest dir, the image is not moved but no message is displayed

I think in such case we need to display a message to the user and do nothing, that is stay on the test dir collection.

TurboGit:
And BTW, I know that this is the current situation. But I think it would be nice to fix that as part of this PR.

schwerdf:
All right. How do you propose to do it without it raising spurious error messages when moving multiple versions of the same file?

Also, while we are on the topic, there is a similar problem I found with the copy function: When copying an image, if a file already exists at the destination path, the copy function does not overwrite the file, but otherwise proceeds as if it had. This would most often result in the database information for one image being erroneously applied to a completely different image. But it is also relied on when copying multiple versions of the same file, so the simplest solution (not proceeding when a file already exists) would not be feasible.

TurboGit:
> All right. How do you propose to do it without it raising spurious error messages when moving multiple versions of the same file?

Just output a dt_control_log() message. Only the last one will be visible so this won't annoy the user. What is important is that in this case at least one message is sent and we don't go to the dest dir.

TurboGit:
For the copy don't we just have to check for the image only? And proceed if the image does not exist. For the .xmp no problem if some exists on the dest dir with the same name and there is no correcponding image I would say they are dead and can be overwritten. No?

schwerdf:
I changed the move function to output its error messages through the log instead of standard error.

The problem with the copy function is that two different cases demand different behavior and there is no way to distinguish between them within the function. Copying multiple versions of the same file is implemented as a sequence of calls to `dt_image_copy`. In all calls subsequent to the first, the destination file will exist, but the copy should still go forward. In all other cases, the copy should fail.

TurboGit:
Ok, I'll test this again. I think at this point we can merge this then and have the copy fixed in another ticket. Can you rebase this PR and resolve the merge conflicts? Thanks.

TurboGit:
Oh, never mind, I've been able to merge this manually... now testing...

src/common/image.c
```diff
@@ -1356,7 +1386,7 @@ int32_t dt_image_move(const int32_t imgid, const int32_t filmid)
         GFile *cnew = g_file_new_for_path(copydestpath);
 
         if(g_file_move(cold, cnew, 0, NULL, NULL, NULL, NULL) != TRUE)
-          fprintf(stderr, "[dt_image_move] error moving local copy `%s' -> `%s'\n", copysrcpath, copydestpath);
```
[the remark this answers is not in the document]
TurboGit: what I'd like is a check on the source file to check it is accessible and then display:

dt_control_log(_("cannot access local copy %s"), srcpath);
and only display the basename BTW.

and check the dest and display same message.

And more importantly, when a move fails we want to stay on the current folder in the lighttable. Some pictures where not moved, we want to stay where they are I think.

src/common/image.c
```diff
@@ -1366,7 +1396,7 @@ int32_t dt_image_move(const int32_t imgid, const int32_t filmid)
     }
     else
     {
-      fprintf(stderr, "[dt_image_move] error moving `%s' -> `%s'\n", oldimg, newimg);
+      dt_control_log(_("error moving `%s' -> `%s'\n"), oldimg, newimg);
```
TurboGit: and also no \n on control log message.

src/common/image.c
```diff
@@ -1366,7 +1396,7 @@ int32_t dt_image_move(const int32_t imgid, const int32_t filmid)
     }
     else
     {
-      fprintf(stderr, "[dt_image_move] error moving `%s' -> `%s'\n", oldimg, newimg);
```
TurboGit: same as above

schwerdf:
I think I have all these change requests addressed. I was a bit unclear about which of the `fprintf` statements should stay and which should go, but I elaborated error messages to distinguish between a missing source file, an existing destination file, and other move errors, and changed the local-copy errors to display the basename only.

The batch handler for GUI moves and copies now only changes the collection in the light table if every individual move/copy it runs returns a non-(-1) exit status.

junkyardsparkle:
Would this perhaps be a good point to rethink the way darktable names its duplicate XMP files for a single image file? Using `<basename>_<nn>.<ext>.xmp` seems like it might not be the best idea, since it's possible for a file named `<basename_nn>.<ext>` to exist for other reasons. Wouldn't it be better to use something like `<basename.ext>.<nn>.xmp`?

I'm bringing this up here because if darktable is going to get involved in things like batch renaming, it would be good to keep in mind that a common application of this involves naming a bunch of files to something like `<basename>_<sequence>.<ext>`...

...yes, backwards compatibility might be painful.

schwerdf:
The naming-pattern functionality for imports already supports the `<basename>_<sequence>.<ext>` style of renaming, so the sidecar nomenclature is already a problem even without batch renaming.

junkyardsparkle:
Isn't that only for *exports* (which don't need the sidecar files anyway)? I didn't think there were currently any methods of actually renaming the working files within darktable (not considering Lua scripts... is that what you mean?) Anyway, it seems like if the sidecar naming is ever going to get fixed, this would be a good time to do it. :)

schwerdf:
No, I was talking about the [session options](https://darktable.gitlab.io/doc/en/session_options.html) used when importing files from a camera.

I agree that the `<basename.ext>.<nn>.xmp` ordering would be much better, but I would also prefer not re-implementing half of Darktable in a single pull request, which means sticking to problems actually introduced by the pull request itself.

junkyardsparkle:
> No, I was talking about the [session options](https://darktable.gitlab.io/doc/en/session_options.html) used when importing files from a camera.

Ah... I forget that people use darktable to pull pictures onto disk from cameras.

> I agree that the `<basename.ext>.<nn>.xmp` ordering would be much better, but I would also prefer not re-implementing half of Darktable in a single pull request, which means sticking to problems actually introduced by the pull request itself.

Sorry, I didn't mean to suggest it was your problem to fix, only that if it did seem worth fixing, it might make sense to wait, and implement renaming functionality afterwards. Nevermind, disregard.

schwerdf:
I added the proposed change to the sidecar nomenclature as [Redmine ticket #12649](https://redmine.darktable.org/issues/12649).

I also added [a Redmine ticket](https://redmine.darktable.org/issues/12650) for fixing the erroneous behavior of the move and copy functions when operating over multiple versions of a file.

PR #2083, opened 2019-02-10
Splitting strings for a new translation

mihkel-t:
Hi,

As discussed on the dev mailing list[1], here comes the PR that makes most of the necessary preparations for adding the Estonian translation.
Apart from splitting strings by context, four strings (in two commits) got marked for translation (I hope there wasn't a good reason why these weren't translatable before :), and one string ("%.0f mired") gained ngettext support (hope I did it right).

[1] https://www.mail-archive.com/darktable-dev@lists.darktable.org/msg03926.html

Best regards,
Mihkel


src/control/jobs/control_jobs.c
```diff
@@ -1544,7 +1544,7 @@ void dt_control_copy_images()
 
   GtkWidget *filechooser = gtk_file_chooser_dialog_new(
       _("select directory"), GTK_WINDOW(win), GTK_FILE_CHOOSER_ACTION_SELECT_FOLDER, _("_cancel"),
-      GTK_RESPONSE_CANCEL, _("_select as destination"), GTK_RESPONSE_ACCEPT, (char *)NULL);
+      GTK_RESPONSE_CANCEL, C_("for copying", "_select as destination"), GTK_RESPONSE_ACCEPT, (char *)NULL);
```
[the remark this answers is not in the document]
mihkel-t: Changed both. Btw is there a reason behind "select as destination" instead of "move here" and "copy here"?

src/control/jobs/control_jobs.c
```diff
@@ -1484,7 +1484,7 @@ void dt_control_move_images()
 
   GtkWidget *filechooser = gtk_file_chooser_dialog_new(
       _("select directory"), GTK_WINDOW(win), GTK_FILE_CHOOSER_ACTION_SELECT_FOLDER, _("_cancel"),
-      GTK_RESPONSE_CANCEL, _("_select as destination"), GTK_RESPONSE_ACCEPT, (char *)NULL);
+      GTK_RESPONSE_CANCEL, C_("for moving", "_select as destination"), GTK_RESPONSE_ACCEPT, (char *)NULL);
```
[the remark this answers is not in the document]
mihkel-t: Changed.

TurboGit:
As discussed in mailing-list this will be integrated when 2.6.1 is out to not break all translations just before merging them in 2.6 branch.

src/iop/colorbalance.c
```diff
@@ -2333,7 +2333,7 @@ void gui_init(dt_iop_module_t *self)
   dt_bauhaus_widget_set_label(g->controls, NULL, _("color control sliders"));
   dt_bauhaus_combobox_add(g->controls, _("HSL"));
   dt_bauhaus_combobox_add(g->controls, _("RGBL"));
-  dt_bauhaus_combobox_add(g->controls, _("both"));
+  dt_bauhaus_combobox_add(g->controls, C_("color control sliders", "both"));
```
TurboGit: Do we really need this context so low-level? Having a context name for each module seems wrong to me.

Here we are talking about "options" or a "choice" and we can select "both" of those. In colorreconstruction.c just below we have "none" (same context to me, options or choice).

src/iop/colorzones.c
```diff
@@ -1112,9 +1112,9 @@ void gui_init(struct dt_iop_module_t *self)
   c->select_by = dt_bauhaus_combobox_new(self);
   dt_bauhaus_widget_set_label(c->select_by, NULL, _("select by"));
   gtk_widget_set_tooltip_text(c->select_by, _("choose selection criterion, will be the abscissa in the graph"));
-  dt_bauhaus_combobox_add(c->select_by, _("hue"));
-  dt_bauhaus_combobox_add(c->select_by, _("saturation"));
-  dt_bauhaus_combobox_add(c->select_by, _("lightness"));
+  dt_bauhaus_combobox_add(c->select_by, C_("select by", "hue"));
```
TurboGit: Here the context is more "color", we are talking about hue, saturation or lightness in the "color" context.

src/iop/globaltonemap.c
```diff
@@ -695,9 +695,9 @@ void gui_init(struct dt_iop_module_t *self)
   g->operator= dt_bauhaus_combobox_new(self);
   dt_bauhaus_widget_set_label(g->operator, NULL, _("operator"));
 
-  dt_bauhaus_combobox_add(g->operator, "reinhard");
-  dt_bauhaus_combobox_add(g->operator, "filmic");
-  dt_bauhaus_combobox_add(g->operator, "drago");
+  dt_bauhaus_combobox_add(g->operator, C_("global tonemap operator", "reinhard"));
```
TurboGit: Here the context is "options" or "choice".

po/LINGUAS
```diff
@@ -5,6 +5,7 @@ da
 de
 el
 es
+et
```
TurboGit: Changing LINGUAS here is not needed. Better to keep this in a separate commit with the et.po translation.

src/iop/colorbalance.c
mihkel-t: Come to think of it, I have probably chosen to add context to the wrong instance of "both" here - this one is quite fine as it is without any context at all, as the meaning is rather basic here, as opposed to the other instances of "both" where it actually means 'in both directions'. What do you think, would context like "flip/rotation direction" be OK for the other instances?

src/iop/globaltonemap.c
mihkel-t: I only added context to these so as to avoid having the same string for the "filmic" module name and the tonemap operator of the same name - but again, come to think of it, might be better to add "modulename" context to the other "filmic" instead :)

po/LINGUAS
mihkel-t: OK, done.

src/iop/colorzones.c
mihkel-t: In a way you're right, of course, however the same "color" context would apply elsewhere as well, while "select by" wouldn't. The preceding label ("select by") was the reason for me wanting these here split, so I could translate "hue", etc. here to go nicely with "select by". (Other inflectional languages might appreciate having this option, too.)

src/gui/preferences.c
```diff
@@ -1454,7 +1454,7 @@ static void edit_preset(GtkTreeView *tree, const gint rowid, const gchar *name,
   gtk_widget_set_hexpand(GTK_WIDGET(g->focal_length_max), TRUE);
 
   // raw/hdr/ldr
-  label = gtk_label_new(_("format"));
+  label = gtk_label_new(C_("raw/hdr/ldr", "format"));
```
TurboGit: C_("file format", "format") ?

src/gui/preferences.c
```diff
@@ -622,7 +622,7 @@ static void init_tab_accels(GtkWidget *book)
   hbox = gtk_box_new(GTK_ORIENTATION_HORIZONTAL, 5);
 
   // Adding the restore defaults button
-  button = gtk_button_new_with_label(C_("preferences", "default"));
+  button = gtk_button_new_with_label(C_("preferences, button", "default"));
```
TurboGit: > C_("preferences, button", "default")

Maybe just button?

C_("button", "default")

We really want to find the simplest context and mre generic one. I don't thins a "preference button" would be handled differently than a "button" anywhere in the interface.

src/gui/presets.c
```diff
@@ -508,7 +508,7 @@ static void edit_preset(const char *name_in, dt_iop_module_t *module)
   gtk_grid_attach_next_to(GTK_GRID(g->details), g->focal_length_max, g->focal_length_min, GTK_POS_RIGHT, 1, 1);
 
   // raw/hdr/ldr
-  label = gtk_label_new(_("format"));
+  label = gtk_label_new(C_("raw/hdr/ldr", "format"));
```
TurboGit: likewise

src/imageio/format/copy.c
```diff
@@ -121,7 +121,7 @@ const char *extension(dt_imageio_module_data_t *data)
 
 const char *name()
 {
-  return _("copy");
+  return C_("export file format", "copy");
```
TurboGit: "file format"

src/imageio/format/pdf.c
```diff
@@ -675,7 +675,7 @@ void gui_init(dt_imageio_module_format_t *self)
 
   d->pages = dt_bauhaus_combobox_new(NULL);
   dt_bauhaus_widget_set_label(d->pages, NULL, _("TODO: pages"));
-  dt_bauhaus_combobox_add(d->pages, _("all"));
+  dt_bauhaus_combobox_add(d->pages, C_("pdf pages", "all"));
```
TurboGit: "pdf pages" -> "count" ? we are counting something?

src/imageio/format/pdf.c
```diff
@@ -583,7 +583,7 @@ void gui_init(dt_imageio_module_format_t *self)
   gtk_grid_attach(grid, widget, 0, ++line, 1, 1);
 
   d->title = GTK_ENTRY(gtk_entry_new());
-  gtk_entry_set_placeholder_text(d->title, "untitled");
+  gtk_entry_set_placeholder_text(d->title, C_("pdf placeholder", "untitled"));
```
TurboGit: "pdf placeholder" -> is that a name? a title? maybe just "name" or "title"?

src/imageio/storage/gallery.c
```diff
@@ -175,7 +175,7 @@ void gui_init(dt_imageio_module_storage_t *self)
   g_free(tooltip_text);
 
   widget = dtgtk_button_new(dtgtk_cairo_paint_directory, CPF_DO_NOT_USE_BORDER, NULL);
-  gtk_widget_set_tooltip_text(widget, _("select directory"));
+  gtk_widget_set_tooltip_text(widget, C_("button tooltip", "select directory"));
```
TurboGit: "button" ?

src/imageio/storage/gallery.c
```diff
@@ -104,7 +104,7 @@ static void button_clicked(GtkWidget *widget, dt_imageio_module_storage_t *self)
   gallery_t *d = (gallery_t *)self->gui_data;
   GtkWidget *win = dt_ui_main_window(darktable.gui->ui);
   GtkWidget *filechooser = gtk_file_chooser_dialog_new(
-      _("select directory"), GTK_WINDOW(win), GTK_FILE_CHOOSER_ACTION_SELECT_FOLDER, _("_cancel"),
+      C_("dialog title", "select directory"), GTK_WINDOW(win), GTK_FILE_CHOOSER_ACTION_SELECT_FOLDER, _("_cancel"),
```
TurboGit: "title" ?

src/iop/clipping.c
```diff
@@ -1963,7 +1963,7 @@ void gui_init(struct dt_iop_module_t *self)
   dt_bauhaus_combobox_add(g->hvflip, _("none"));
   dt_bauhaus_combobox_add(g->hvflip, _("horizontal"));
   dt_bauhaus_combobox_add(g->hvflip, _("vertical"));
-  dt_bauhaus_combobox_add(g->hvflip, _("both"));
+  dt_bauhaus_combobox_add(g->hvflip, C_("flip direction", "both"));
```
TurboGit: "direction" ? "orientation"?

src/imageio/storage/gallery.c
mihkel-t: Won't that be misleading?

src/gui/preferences.c
mihkel-t: Neither RAW, LDR nor HDR is really a file format, as opposed to e.g. JPG or PDF. This is one example where English easily gets by using a rather broad term ("format"), but other languages might need to specify whether it is a "file format" (like JPG) or something to in the vein of "image type" (like HDR).

src/imageio/storage/gallery.c
TurboGit: Why? Would that work for you? Maybe the context is not even needed here? Do you need it?

src/gui/preferences.c
TurboGit: Other languages, which one? Do you need it? I really want to have a minimal change at this stage. The minimum for you to get a clean translation and as I said with the more generic context as possible otherwise we'll just end up with lot of more translated string to do and this in all languages.

src/iop/colorzones.c
TurboGit: Maybe I can't tell for your language, but I'm concerned. Will we have:

   _("select by", "hue")
   _("choose", "hue")
   _("button", "hue")
   ...

Here it seems to me that "hue" is used in the context/theme of color. The context in C_("")  macro is really the context/theme of the term used not an action "select by". Again I don't know much about languages structure, but I want the translation to stay at a minimum level for maintenance.

src/iop/colorbalance.c
TurboGit: just "direction" maybe or "orientation"?

src/gui/preferences.c
mihkel-t: Agreed; now changed :)

src/gui/preferences.c
mihkel-t: The specific "other" language I had in mind here was of course Estonian, where I would very much like to be able to translate this occurrence differently from the generic "format".
I checked Wikipedia articles about RAW and HDR in e.g. Danish, Norwegian, Swedish, Finnish and German - RAW was said to be either a "common denominator" for various file formats or just file format (so based on that RAW actually could be argued to fall under the generic "format" as well), however HDR was mostly called a "technique". So translators of at least those languages might also like the opportunity to translate this hdr/ldr/raw "format" differently from the generic format.

src/imageio/format/copy.c
mihkel-t: Changed.

src/imageio/format/pdf.c
mihkel-t: Frankly I'm not sure anymore why I wanted to add context here, as this is the only occurrence of "untitled"... I'll undo this one. Sorry.

src/imageio/format/pdf.c
mihkel-t: From the code it didn't seem like it's counting something - rather it seems to select what kinds of pages should be added to the resulting PDF file: single images, contact sheet, or all of those. Of course I might've misunderstood, and since this feature is in "TODO" state, it doesn't actually even matter right now, as it won't be visible in the application interface.

src/imageio/storage/gallery.c
mihkel-t: The rationale behind adding context to this and other occurrences of "select directory" is to be able to immediately distinguish (as a translator) whether it's a dialog title or a button or a tooltip. For Estonian translation, the distinction between button and dialog title is important (as the translation should be different), while button and tooltip would share the same translation. So as far as the upcoming Estonian translation is concerned, the contexts "dialog title" (or even just "title") vs. "button/tooltip" (or even just "button") would be fine. I'll change them accordingly.

TurboGit: But the issue is that for each different context we will have to translate the text. So that's many many more translation and I really want to avoid that. Translating is already a quite challenging job!

mihkel-t: From my ca. 15 years experience of translating software, I'd say it's more frustrating to be unable to translate a word/phrase differently in different contexts, than to be forced to translate the same word/phrase two or more times. The latter would just mean picking the previously translated entry from translation memory :)

mihkel-t: Now changed these tooltips to share the "button" context.

src/imageio/storage/gallery.c
mihkel-t: I'd prefer "dialog title" as this makes it unmistakably clear for translators what and where this is. While there probably won't be any confusion with this specific string, I think it's better to use precise terminology, especially as there's no practical downside that I can think of...

src/iop/colorbalance.c
mihkel-t: For my purposes right now, it doesn't matter which we use, as long as we do use one. From a general perspective, I'd say "flip direction" is clearer, though (NB: not "flip/rotation direction" as I mistakenly wrote above).

P.S. If another occurrence of "both" gets added somewhere in darktable, chances are the Estonian translation would need it to have its own context anyway (unless it would again be about flipping something), as our word for "both" is inflected in several possible ways depending on, ahem, context :)

src/iop/colorzones.c
mihkel-t: "Hue" is basically always used in the context of color, so this won't really help. What about "colorzones", the module where these occur?

Longer explanation:
As you probably know, there are currently two specified contexts for "hue" in dt:
1) "blendmode"
2) "accel"
... plus the unspecified context which is used everywhere else.
I'd like (or need) to separate three more contexts, in order for the Estonian translation to look pretty:
3) "channelmixer" (the string "gray" in that module has this context already)
4) "color reconstruction precedence" (just "color reconstruction" works just as well)
5) "colorzones" (a.k.a. "select by" or however we agree this to be named :) )
... as the translation would differ slightly in those places.
Of course there would still be the undefined context as well, with its remaining 8 occurrences of "hue".

To lessen the impact of this on existing translations, we could agree that I either do another PR afterwards or amend this one to update existing .po files so they'd retain their current translation of "hue" and other similarly affected strings in these added contexts. (For about 2/3 of the languages I'd have to mark these strings as fuzzy, though, as I don't read all the languages darktable has been translated to, and am therefore unable to decide whether the translation could benefit from the new contexts or not.)

mihkel-t: Renamed this context to "colorzones" for now - let me know if that's not an improvement :)

mihkel-t:
I went through the changes in this PR and tried to make the contexts less "low-level" - I think it should be a bit better now. I didn't (yet) change the ones where I had countered your suggestions in the comments above.

mihkel-t:
Could you point out where you feel the added context is still too "low-level"?

TurboGit:

PR #2088, opened 2019-02-11
remove warnings and special treatment for 32-bit architectures

rabauke:
This PR is to determine the bitwidth of the target architecture (32-bit vs. 64-bit architectures) at compile time rather than at run time.

### DIFF
--- a/data/darktableconfig.xml.in
+++ b/data/darktableconfig.xml.in
@@ -1824,13 +1824,6 @@
     <longdescription>number of images to load in advance in the background when showing full-size preview</longdescription>
   </dtconfig>
   <dtconfig>
-    <name>please_let_me_suffer_by_using_32bit_darktable</name>
-    <type>bool</type>
-    <default>false</default>
-    <shortdescription>if using a 32-bit build, hide the message recommending a 64-bit build</shortdescription>
-    <longdescription>the 32-bit build of darktable is known to exhibit sporadic issues and crashes. we show a warning message hinting at this issue. with this option, you can suppress the warning.</longdescription>
-  </dtconfig>
-  <dtconfig>
     <name>codepaths/sse2</name>
     <type>bool</type>
     <default>true</default>
--- a/src/common/darktable.c
+++ b/src/common/darktable.c
@@ -77,6 +77,7 @@
 #include <sys/types.h>
 #include <unistd.h>
 #include <locale.h>
+#include <limits.h>
 
 #if defined(__SSE__)
 #include <xmmintrin.h>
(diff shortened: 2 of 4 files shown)


TurboGit:
I'm not sure to understand the goal here. Only 64 bits systems are supported at this point.

cryptomilk:
Normally you do that checking CMAKE_SIZEOF_VOID_P and then put it into config.h

rabauke:
I agree CMAKE_SIZEOF_VOID_P might be a better solution if cmake is employed. Will update the patch.

rabauke:
I revised my pull request to employ `CMAKE_SIZEOF_VOID_P` to set up `config.h`.

TurboGit:
What is the status here? Also you never answered my question above. Thanks.

rabauke:
@TurboGit Which question? I am willing to improve the patch, but currently I have no clue where there might be a problem.

TurboGit:
@rabauke : the qestion above:

https://github.com/darktable-org/darktable/pull/2088#issuecomment-462453892

I'm not clear about what is this solving.

rabauke:
Currently, the code in darktable.c checks at run-time if the program runs on a 32-bit architecture or on a 64-bit architecture. Depending on this check different code paths are taken. This mean that there is some dead code path that will never be taken depending on the architecture. My patch excludes these dead code paths by a compile-time check.

A good optimizer might remove unreachable code paths. Nevertheless, I think one should avoid unreachable code.

If only 64-bit architectures are currently supported, I would prefer to remove the code 32-bit specific code completely.

TurboGit:
That's my point indeed. We claim that we do not support 32bit architecture since some year now. We have a message saying "run at your own risk" IIRC. I'm think that at this stage we can remove the 32bit code path. Please let's do that in a single commit. If we need to revert it will be easier. Thanks.

rabauke:
I removed the 32bit code path.

moy:
If we officially drop support for 32 bits (which is probably a good thing, anyway it dt seems heavily broken there, and clearly untested by developers), then the runtime check and warning should probably be dropped too:

https://github.com/darktable-org/darktable/blob/master/src/views/view.c#L257-L290

Ideally, there would be a compile-time check to prevent building for 32 bits platform.

rabauke:
I also removed the code mentioned by @moy . I have not yet implemented a compile-time check to prevent building for 32-bit platforms, although I have considered this before. I am aware of at least one Linux distro (Mageia), which ships DT 2.6.0 for i586. Nevertheless, I would not hesitate to do so if there I a consensus to push users and distros harder into the direction of 64 bit.

moy:
I haven't tried a 32bits build myself since before the support was semi-dropped, but most feedback about 32bits I got was along the lines of "there was a stupid warning which I ignored, and then nothing worked, I don't understand why" ;-). And since no dev is testing on 32 bits, the situation can only get worse and worse.

Perhaps the best trade-off would be to prevent the build, but to provide an override like

    cmake -DI_ACKNOWLEDGE_32_BITS_DT_IS_BROKEN=true

or so for people who like to suffer.

TurboGit:
Let's this in, we don't test 32bit systems. We don't want to support them. So better to disable it. Thanks.

PR #2089, opened 2019-02-11
Google Photo

TurboGit:
Version using the new Google Photo API.

Issue is that the CLIENT_ID and CLIENT_SECRET in the code is limited to 10.000 call per day.

What could/should we do? Have a preference to set this and let every user mess with that?

Any other solution?

### DIFF
--- a/data/darktableconfig.xml.in
+++ b/data/darktableconfig.xml.in
@@ -1314,6 +1314,20 @@
     <longdescription/>
   </dtconfig>
   <dtconfig>
+    <name>plugins/imageio/storage/gphoto/id</name>
+    <type>string</type>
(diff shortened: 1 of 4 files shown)


TurboGit:
Forgot to add:
pro:
- it is now possible to create albums
cons:
- the photo api can only add pictures in album create by itself

whitwhittle:
I realize this is late, but regarding the CLIENT_ID & CLIENT_SECRET, I would definitely have each user create those for themselves and add into configuration file.

TurboGit:
That's possible, but how many people know how to do that? So a default is provided, even if in this case the upload limit can be a real issue.

PR #2090, opened 2019-02-11
fix memory leaks

rabauke:
The call to `dt_conf_get_string` returns the result of the `g_strdup` function, which allocates memory for the duplicated string. The returned string must be freed with `g_free` when no longer needed. See https://developer.gnome.org/glib/stable/glib-String-Utility-Functions.html#g-strdup

### DIFF
--- a/src/gui/gtk.c
+++ b/src/gui/gtk.c
@@ -952,9 +952,12 @@ int dt_gui_gtk_init(dt_gui_gtk_t *gui)
   dt_loc_get_datadir(datadir, sizeof(datadir));
   dt_loc_get_user_config_dir(configdir, sizeof(configdir));
 
-  const gchar *css_theme = dt_conf_get_string("ui_last/theme");
+  gchar *css_theme = dt_conf_get_string("ui_last/theme");
   if(css_theme)
+  {
     g_snprintf(gui->gtkrc, sizeof(gui->gtkrc), "%s", css_theme);
+    g_free(css_theme);
+  }
   else
     g_snprintf(gui->gtkrc, sizeof(gui->gtkrc), "darktable");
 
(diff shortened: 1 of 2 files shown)


src/gui/gtk.c
```diff
@@ -954,7 +954,10 @@ int dt_gui_gtk_init(dt_gui_gtk_t *gui)
 
   const gchar *css_theme = dt_conf_get_string("ui_last/theme");
```
TurboGit: I would prefer the const to be removed here...

src/gui/gtk.c
```diff
@@ -954,7 +954,10 @@ int dt_gui_gtk_init(dt_gui_gtk_t *gui)
...
   if(css_theme)
+  {
     g_snprintf(gui->gtkrc, sizeof(gui->gtkrc), "%s", css_theme);
+    g_free((gpointer)css_theme);
```
TurboGit: and so (gpoiner) cast won't be needed.

rabauke:
I removed the `const` specifier.

src/gui/preferences.c
```diff
@@ -247,7 +247,6 @@ static void hardcoded_gui(GtkWidget *grid, int *line)
   const char *theme_name = dt_conf_get_string("ui_last/theme");
```
TurboGit: same here :) no const

src/gui/preferences.c
```diff
@@ -258,6 +257,7 @@ static void hardcoded_gui(GtkWidget *grid, int *line)
     if(!g_strcmp0(name, theme_name)) selected = k;
     k++;
   }
+  g_free((gpointer)theme_name);
```
TurboGit: and no gpointer. thanks.

rabauke:
Oh, I missed this. Has been fixed too now.

PR #2098, opened 2019-02-13
tags: fix memory leak

rabauke:
We have to take care if the `tag` string is allocated on heap or is just a substring from the input `dest`.


TurboGit:
Looks good. Thanks.

PR #2107, opened 2019-02-15
fix color picker

edgardoh:
Changed the variables for the rgb values from uint8 to float so it doesn't overflow on clipped values.

### DIFF
--- a/src/libs/colorpicker.c
+++ b/src/libs/colorpicker.c
@@ -149,7 +149,7 @@ static void _update_picker_output(dt_lib_module_t *self)
 
     // always adjust picked color:
     int m = dt_conf_get_int("ui_last/colorpicker_mode");
-    uint8_t *rgb;
+    float *rgb;
     float *lab;
     switch(m)
     {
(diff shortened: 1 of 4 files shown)


src/libs/colorpicker.c
```diff
@@ -268,17 +268,17 @@ static void _update_samples_output(dt_lib_module_t *self)
     }
 
     // Setting the output button
-    sample->rgb.red = rgb[0] / 255.0;
-    sample->rgb.green = rgb[1] / 255.0;
-    sample->rgb.blue = rgb[2] / 255.0;
+    sample->rgb.red = rgb[0];
```
TurboGit: Why don't you CLAMP the values here? As done in line 181. Seems like it is needed here too, no?

edgardoh: I'm clamping only when I assign it to the color widget, I want to display the actual pixel values, even when out of range.

TurboGit: But that's what happen here no? You queue the update for the color_patch just below. Those are for the live samples which happen also to have a widget to display the color for the given area.

edgardoh: Ups, you're right, its fixed now. Thanks!

PR #2115, opened 2019-02-19
lighttable : add timeline lib

AlicVB:
this allow to select image by date/time using a "timeline" like this : 
capture d ecran_2019-02-19_19-57-20
Features : 
- ctrl-scroll to zoom in/out
- [ and ] shortcuts to set the start/end of the selection (or with the mouse)
- Ctrl-F to hide/show the timeline (like the filmstrip)

Questions : 
- how about date localization. I use dd/mm/yy hh:mm:ss is it different in other language ? (and how to deal with that in that case ?)
- I've not found an obvious way to show on the timeline that there's a selection defined (if not visible)
- When you put the mouse on module ends, it scroll the timeline, but I'm not sure I've done that right. What do you think ?

Future : 
- add the possibility to add multiple date selections (with Ctrl key ?)


TurboGit:
>     * how about date localization. I use dd/mm/yy hh:mm:ss is it different in other language ? (and how to deal with that in that case ?)

I think you can look at the image information module which does that.

TurboGit:
Some comments;
- The zoom in/out has a too big step I found. I have 15 years in the timeline. A single zoom-in with the scroll and I get one year a 2 months displayed.
- When changing collection it would be nice to have an indicator of the covered period in the timeline (not like the selection which is sync with date/time on collect module.
- I agree that the scroll on the side of the timeline is not very nice. First we scroll once and then we need to move the cursor. And the move is a bit too slow I think.

TurboGit:
I think the timeline should be deactivated in "exposé" mode. In the current version the timeline is displayed on the left of the filmstrip. I really don't think the timeline is of any use while in "exposé" mode anyway.

TurboGit:
I forget that I really like this new lib. Very handy. Thanks.

AlicVB:
>     * The zoom in/out has a too big step I found. I have 15 years in the timeline. A single zoom-in with the scroll and I get one year a 2 months displayed.
For instance, the zoom level are defined so a "block" (~120px) represent one year/month/day/... we can split them in 2 (3 ?) something like 3 years/years/6months/month/...
>     * When changing collection it would be nice to have an indicator of the covered period in the timeline (not like the selection which is sync with date/time on collect module.
Hmm not sure how to get that dates without redoing all the sql manually, but I will have a look
>     * I agree that the scroll on the side of the timeline is not very nice. First we scroll once and then we need to move the cursor. And the move is a bit too slow I think.
I find this particularly useful when you want to scroll the timeline when doing a selection with the mouse.

>     * I think the timeline should be deactivated in "exposé" mode. In the current version the timeline is displayed on the left of the filmstrip. I really don't think the timeline is of any use while in "exposé" mode anyway.
So there's a bug somewhere, it should be disabled (and it works for me...)

TurboGit:
> Hmm not sure how to get that dates without redoing all the sql manually, but I will have a look

I was think about using a callback on COLLECTION_CHANGE. Then you have all the images in the collected_images in-memory table. From there you can get the starting and ending dates and display some indicator on the timeline.

TurboGit:
> So there's a bug somewhere, it should be disabled (and it works for me...)

Even if you do a ctrl-f to enable it while in exposé mode?

TurboGit:
@cryptomilk : if I agree with the comment I'm not sure it applies here. The dates are only displayed at the bottom of the timeline only. These are somehow already sorted for you anyway :)

AlicVB:
@cryptomilk : No need of sortable dates. Ideally the dates should be displayed in whatever format used usually in UI language. Example : for French peoples, it's more readable to display the date like 29/06/2014, but I know it's not the case for other languages. I'll see if I manage to take care of that point.

TurboGit:
Nice, for me the zoom factor and the scroll are ok now.

AlicVB:
now the timeline is keep in sync with the actual collection. Bar graphs have 2 "colors" to show the number of existing/collected images.
When you have some collect rules defined,
- selection in the timeline either add a new datetime rule or modify the last rule if it's of datetime type
- selection with shift press remove all existing rules and replace them with a datetime one

There's still some bugs I have to fix, and the datetime translation too.

AlicVB:
My TODO list for this lib is now empty :)
...except for the datetime translation : 
The pb is for date range, example : (21-31)/05/2008 (We don't have enough space to show full dates like 21/05/2008-31/05/2008, so we can't use strftime with %x like in metadata viewer)
Maybe we can stay with the current untranslated datetimes, as we just use numeric formats.

TurboGit:
Let me fill the todo list then :)

When you close the timeline the lighttable is not properly refreshed. I suppose a call to dt_control_queue_redraw_center() is missing.

TurboGit:
Looks good now. I have merged (manually and resolved the conflicts with my lighttable undo support) it on master to get more testing. Thanks again for the nice contribution.

PR #2118, opened 2019-02-20
Malloc overhaul

markfeit:
I've been using this for real processing and am satisfied that it works.
The next phase of this will be to switch calls to the g_* functions.  Those are safe(r) as-is since they exit the program if an allocation fails, but consistency would be good.


markfeit:
The one remaining problem is `src/chart/thinplate.c`:
```
/darktable/src/chart/thinplate.c
/build/darktable/src/chart/thinplate.c: In function ‘thinplate_match’:
/build/darktable/src/chart/thinplate.c:196:3: warning: ISO C90 forbids variable length array ‘r’ [-Wvla]
   double(*r)[wd] = dt_malloc(dim * wd * sizeof(double));
   ^~~~~~
```

The code in question:

```
 // XXX do we need these explicitly?                                                                                               
  // residual = target vector                                                                                                       
  double(*r)[wd] = dt_malloc(dim * wd * sizeof(double));
  const double **b = dt_malloc(dim * sizeof(double *));
  for(int k = 0; k < dim; k++) b[k] = target[k];
  for(int k = 0; k < dim; k++) memcpy(r[k], b[k], wd * sizeof(double));
```

How was this passing muster before this change?

rabauke:
Looks good. Some minor issues:

* The format string of size_t is "%zu", afaik.
* I would like to suggest to squeeze all commits into a single one, see git rebase.
* I would also remove all the casts of the return values of the memory allocation functinons. These cast are K&R C style and not required in ANSI C (but in C++, albeit for a different reason).

rabauke:
I see no reason to declare the inline functions as static.

moy:
> I see no reason to declare the inline functions as static.

`inline` is only a compiler hint, but there's no guarantee that the inlining will occur. Defining a function in a `.h` file without `static` is invalid (link may fail with duplicate symbol error) as soon as the `.h` file is included from several compilation units. Actually, the cases where you want `inline` without `static` are [very rare](https://stackoverflow.com/questions/6312597/is-inline-without-static-or-extern-ever-useful-in-c99).

markfeit:
Format string:  Old habits die hard.  Fixed.  :-)

Commit squeezing:  The squash and merge option when accepting the PR will take care of that with no rebase freebasing required.

Casts:  I considered removing casts while I was doing this but elected to keep the number of changes to the code to the minimum required.

Static inline:  I'd advise leaving this as-is.  If the compiler elects to emit the function instead of inlining it (which it can, because `inline` is advisory-only), the symbol will be externally-visible and the linker will choke on the duplicates.  This is also the most-portable way to do it without a lot of preprocessor gymnastics; the only other alternative is `extern inline`, but its behavior is compiler-dependent.  I surveyed the symbol tables of a half-dozen random object files in the dt build and found the compiler was doing exactly what I expected.  There are a number of other inline functions in the source; addressing how they're dealt with should be a project-wide effort.

rabauke:
@markfeit Regarding the inline functions: I thought the `static` specifier might be omitted because I am more conversant with C++, where `inline` has a slightly different meaning. After reading the corresponding sections in the C99 standard, I agree, however. Declaring the inline functions as `static inline` is a good idea. (Although, I think, the rationale is a bit different from your explanation.)

markfeit:
@LebedevRI  Those updates appear to have originated in darktable-org/rawspeed.  I didn't touch them.

LebedevRI:
> @LebedevRI Those updates appear to have originated in darktable-org/rawspeed. I didn't touch them.

I would advise you to read about git submodules.
Don't `git add` `src/extermal/rawspeed` into commits,
unstage it from the commits you have already added it into.
(or in other words, don't use `git add -a`)

aurelienpierre:
> > I see no reason to declare the inline functions as static.
> 
> `inline` is only a compiler hint, but there's no guarantee that the inlining will occur. Defining a function in a `.h` file without `static` is invalid (link may fail with duplicate symbol error) as soon as the `.h` file is included from several compilation units. Actually, the cases where you want `inline` without `static` are [very rare](https://stackoverflow.com/questions/6312597/is-inline-without-static-or-extern-ever-useful-in-c99).

Notice that inlining functions prevents them from being auto-vectorized when possible.

rabauke:
A lot of conflicts emerged due to recent progress in the master branch. Is there still an interest / a chance to get this into master? Help needed or wanted?

markfeit:
I got (and still am) kind of tied up.  I'll go through and resolve them when I get some time.

TurboGit:

PR #2119, opened 2019-02-20
fix overexposed

edgardoh:
This displays the image on display profile while in overexpose check, keeping the check in histogram profile.

I didn't change the behaviour, still shows the clipped areas and that's probably what we want. Maybe we can change the name to clipped check, at least is what RT does.
The only thing I don't like is that it doesn't give any clue about which channel is clipped.


TurboGit:
Looks good. Thanks.

PR #2120, opened 2019-02-20
bugfix: halos in local contrast and other weird things

aurelienpierre:
Remove the harmfull `-ffast-math` compiler flag for release builds. That fixes https://redmine.darktable.org/issues/12349 and possibly many weird demoisaicing bugs in Windows and Mac (not tested).

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -248,8 +248,8 @@ if(USE_OPENMP)
 endif(USE_OPENMP)
 
 if(USE_DARKTABLE_PROFILING)
-	add_definitions(-DUSE_DARKTABLE_PROFILING)
-	set(SOURCES ${SOURCES} "common/profiling.c")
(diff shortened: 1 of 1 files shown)


TurboGit:
I cannot reproduce this issue. Any simple way to reproduce?

aurelienpierre:
For me, is appears after building with `sh build.sh --build-type Release`, running without OpenCL and using local contrast in laplacian mode.

But I also suspect the packagers for Win and Mac (@parafin and @peterbud) use the `Release` build, and the recent bunch of bugs affecting colors in darkroom vs. export (unreproductible on Linux) could be related to that. Basically, that flag triggers non-safe maths bypasses for the sake of performance, but you have to handle things properly in the code.

parafin:
I’m using the default build type which I think is RelWithDebInfo that uses just -O2 flag.

TurboGit:
The risk is high indeed and since the quality of the final result is definitely the main goal I'll approve this PR. Thanks.

PR #2122, opened 2019-02-21
OpenCL: remove -cl-fast-relaxed-math compiler flag

aurelienpierre:
Same as for the CPU build, this option removes the NaN checks and enables unsafe optimizations possibly harmful given the absence of code quality insurance in dt.

On the other hand, we enable `-cl-mad-enable` which is only an approximation of `a * b + c` (a bit less accurate, but doesn't skip checks). See https://www.khronos.org/registry/OpenCL/sdk/1.2/docs/man/xhtml/clBuildProgram.html


TurboGit:
Sounds like a good move to me. Let's try this for master.

PR #2123, opened 2019-02-22
ashift: enable SIMD auto-vectorization

aurelienpierre:
* collapse parallel loops and rewrite pointers indexing
* add OpenMP simd in parallel loops
* align all arrays on 16 bits blocks
* avoid false-sharing in parallel loops using `const` or `*const`
* `restrict` variables to avoid pointer aliasing

### DIFF
--- a/src/iop/ashift.c
+++ b/src/iop/ashift.c
@@ -529,7 +529,10 @@ void connect_key_accels(dt_iop_module_t *self)
 
 // multiply 3x3 matrix with 3x1 vector
 // dst needs to be different from v
-static inline void mat3mulv(float *dst, const float *const mat, const float *const v)
+#ifdef _OPENMP
+#pragma omp declare simd
+#endif
+static void mat3mulv(float *dst, const float *const mat, const float *const v)
 {
   for(int k = 0; k < 3; k++)
   {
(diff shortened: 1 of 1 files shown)


aurelienpierre:
CLang keeps crashing on `simd` instructions he is not able to vectorize, while GCC can. I think we need to modify the compilation flags to just ignore failed loop vectorizations, because it does not prevent the functionality to work as expected.

TurboGit:
Do you have timing to compare the perfs?

aurelienpierre:
About 25 to 27 % faster with the last version. 

Basically, I don't understand why all RGB vector algebra in darktable are handled with raw `(float *)` pointers while we know beforehand they are `float[3]` or sometimes `float[3][3]`. Just using fixed-size arrays whenever possible gives the compiler a hint on auto-vectorization and memory management, it speed-ups things at no cost. 

I have only rewritten the trivial cases here, more care should be given to the `mat3mul` functions because some funky stuff happens to the matrices regarding indexing, and I don't understand it. (For God's sake just use `float[3][3]` arrays and don't be clever with pointers arithmetic…).

aurelienpierre:
This will need to be tested for output consistency. Stay tuned.

johnny-bit:
With unit testing from #4296 it should be easier to guarantee output consistency in works such as this :)

PR #2124, opened 2019-02-22
Updated Catalan translation

josepvm:
I do not know if we have already reached the "string freeze" for Darktable 2.6.1, so I continue updating the translation in master branch if I see some changes ...


TurboGit:
No problem. There is no hard string freeze of minor release. You can update the translation. All strings in 2.6 should be fixed at this stage.

PR #2127, opened 2019-02-23
Denoise profile updates

rawfiner:
Current state:
1- Fix anscombe transform so that the white balance coefficients are correctly taken into account
2- Adds an option in case user doesn't want a white balance adaptive transform
3- Fix non local means patch norm computation (as a result, increasing patch size does not require to increase the force anymore -> easier to use)
4- Add a slider to control the weight of central pixel in non local means patch comparison metric

Basically I currently use non local means like this :
- I put the patch size to 4
- I increase the coarse grain noise slider until no very coarse chroma noise remains (some fine grain chrominance noise can sometimes remain)
- if image is too much blurred, I reduce the force *slightly*
- then I increase the details slider to my taste
- I fix the remaining chroma noise with the equalizer
Note that with these changes, I am mostly using only one instance of denoiseprofile, without any particular blending mode.

It works very well for medium and high iso, but is not perfect for very high iso. In such case, it can be combined with denoise bilateral to get nice results.

This is work in progress, I did not update the presets yet, and I have ongoing work to make the anscombe transform perform better.

### DIFF
--- a/src/iop/denoiseprofile.c
+++ b/src/iop/denoiseprofile.c
@@ -120,6 +120,7 @@ typedef struct dt_iop_denoiseprofile_params_t
   float y[DT_DENOISE_PROFILE_NONE][DT_IOP_DENOISE_PROFILE_BANDS]; // values to change wavelet force by frequency
   // backward compatibility options
   gboolean fix_anscombe;
+  gboolean wb_adaptive_anscombe; // whether to adapt anscombe transform to wb coeffs
 } dt_iop_denoiseprofile_params_t;
 
 typedef struct dt_iop_denoiseprofile_gui_data_t
@@ -147,6 +148,7 @@ typedef struct dt_iop_denoiseprofile_gui_data_t
   float draw_xs[DT_IOP_DENOISE_PROFILE_RES], draw_ys[DT_IOP_DENOISE_PROFILE_RES];
   float draw_min_xs[DT_IOP_DENOISE_PROFILE_RES], draw_min_ys[DT_IOP_DENOISE_PROFILE_RES];
   float draw_max_xs[DT_IOP_DENOISE_PROFILE_RES], draw_max_ys[DT_IOP_DENOISE_PROFILE_RES];
+  GtkWidget *wb_adaptive_anscombe;
   GtkWidget *extra_expander;
   GtkWidget *extra_toggle;
   // backward compatibility options
@@ -166,6 +168,7 @@ typedef struct dt_iop_denoiseprofile_data_t
   float force[DT_DENOISE_PROFILE_NONE][DT_IOP_DENOISE_PROFILE_BANDS];
   // backward compatibility options
   gboolean fix_anscombe;
+  gboolean wb_adaptive_anscombe; // whether to adapt anscombe transform to wb coeffs
 } dt_iop_denoiseprofile_data_t;
 
 typedef struct dt_iop_denoiseprofile_global_data_t
@@ -329,6 +332,7 @@ int legacy_params(dt_iop_module_t *self, const void *const old_params, const int
     }
     v7->scattering = v6.scattering;
     v7->fix_anscombe = FALSE; // don't fix anscombe for backward compatibility
+    v7->wb_adaptive_anscombe = TRUE;
     return 0;
   }
   return 1;
@@ -1033,27 +1037,27 @@ static void process_wavelets(struct dt_iop_module_t *self, dt_dev_pixelpipe_iop_
   float wb[3] = { wb_mean, wb_mean, wb_mean };
   if(d->fix_anscombe)
   {
-    if(wb_mean != 0.0f)
+    if(wb_mean != 0.0f && d->wb_adaptive_anscombe)
     {
-      wb[0] = piece->pipe->dsc.temperature.coeffs[0] * d->strength * (in_scale * in_scale);
-      wb[1] = piece->pipe->dsc.temperature.coeffs[1] * d->strength * (in_scale * in_scale);
-      wb[2] = piece->pipe->dsc.temperature.coeffs[2] * d->strength * (in_scale * in_scale);
+      for(int i = 0; i < 3; i++) wb[i] = piece->pipe->dsc.temperature.coeffs[i];
     }
-    else
+    else if(wb_mean == 0.0f)
     {
       // temperature coeffs are equal to 0 if we open a JPG image.
-      // in this case, consider them equal to 1.
-      wb[0] = d->strength * (in_scale * in_scale);
-      wb[1] = d->strength * (in_scale * in_scale);
-      wb[2] = d->strength * (in_scale * in_scale);
+      // in this case consider them equal to 1.
+      for(int i = 0; i < 3; i++) wb[i] = 1.0f;
     }
+    // else, wb_adaptive_anscombe is false and our wb array is
+    // filled with the wb_mean
   }
   else
   {
-    wb[0] = 2.0f * piece->pipe->dsc.processed_maximum[0] * d->strength * (in_scale * in_scale);
-    wb[1] = piece->pipe->dsc.processed_maximum[1] * d->strength * (in_scale * in_scale);
-    wb[2] = 2.0f * piece->pipe->dsc.processed_maximum[2] * d->strength * (in_scale * in_scale);
+    wb[0] = 2.0f * piece->pipe->dsc.processed_maximum[0];
+    wb[1] = piece->pipe->dsc.processed_maximum[1];
+    wb[2] = 2.0f * piece->pipe->dsc.processed_maximum[2];
   }
+  // update the coeffs with strength and scale
+  for(int i = 0; i < 3; i++) wb[i] *= d->strength * (in_scale * in_scale);
   // only use green channel + wb for now:
   const float aa[3] = { d->a[1] * wb[0], d->a[1] * wb[1], d->a[1] * wb[2] };
   const float bb[3] = { d->b[1] * wb[0], d->b[1] * wb[1], d->b[1] * wb[2] };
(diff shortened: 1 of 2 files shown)


rawfiner:
I updated the presets, rebased on master and force-pushed.
I will make another PR later for the anscombe changes, they take more time and may be bigger that I first thought.

TurboGit:
So you mean that this is ready for review/integration?

rawfiner:
Yes, this is ready for review

TurboGit:
@rawfiner : I don't like the backward compatibility GUI section. I understand that you need to keep that with a values FALSE for both boolean for previous edits. What I would do:

* solution 1:
- for old edits display the backward compatibility section with both values not selected
- this make it possible for users to tick both options (fix ascomb, fix patch)
- when both options are selected do not display this section at all, I don't think someone would want to uncheck the fix (at least we certainly don't want to encourage this).

* solution 2:
The other solution is to never show those options. Both are FALSE for old edit and we set them to TRUE automatically for new edit or when the module is reset.

If we go with solution 1, I would even have a single tick:
 [ ] fix ascom transform and patch normalisation
For old edit andf again do not display that if it has already been selected by users.

src/iop/denoiseprofile.c
```diff
@@ -278,6 +308,41 @@ int legacy_params(dt_iop_module_t *self, const void *const old_params, const int
...
+    {
+      for(int c = 0; c < DT_DENOISE_PROFILE_NONE; c++)
+      {
+        v7->x[c][b] = v6.x[c][b];
+        v7->y[c][b] = v6.y[c][b];
+      }
+    }
+    v7->scattering = v6.scattering;
+    v7->central_pixel_weight = 0.0;
+    v7->fix_anscombe = FALSE;     // don't fix anscombe to ensure backward compatibility
```
TurboGit: A single boolean here:
`  v7->fixed_algo`

?

src/iop/denoiseprofile.c
```diff
@@ -1508,12 +1685,47 @@ static int process_nlmeans_cl(struct dt_iop_module_t *self, dt_dev_pixelpipe_iop
...
   const int P = ceilf(d->radius * scale); // pixel filter size
   const int K = ceilf(d->nbhood * scale); // nbhood
-  const float norm = 0.015f / (2 * P + 1);
 
+  // Each patch has a width of 2P+1 and a height of 2P+1
+  // thus, divide by (2P+1)^2.
+  // The 0.045 was derived from the old formula, to keep the
+  // norm identical when P=1, as the norm for P=1 seemed
+  // to work quite well: 0.045 = 0.015 * (2 * P + 1) with P=1.
+  float norm = .045f / ((2 * P + 1) * (2 * P + 1));
```
TurboGit:
```
float norm;
if(d->fixed_algo)
   norm = .045f / ((2 * P + 1) * (2 * P + 1));
else
   norm = .015f / (2 * P + 1);
```

src/iop/denoiseprofile.c
```diff
@@ -1508,12 +1685,47 @@ static int process_nlmeans_cl(struct dt_iop_module_t *self, dt_dev_pixelpipe_iop
...
+    norm = .015f / (2 * P + 1);
+  }
+
+  const float wb_mean = (piece->pipe->dsc.temperature.coeffs[0] + piece->pipe->dsc.temperature.coeffs[1]
+                         + piece->pipe->dsc.temperature.coeffs[2])
+                        / 3.0f;
+  // we init wb by the mean of the coeffs, which corresponds to the mean
+  // amplification that is done in addition to the "ISO" related amplification
+  float wb[4] = { wb_mean, wb_mean, wb_mean, 0.0f };
+  if(d->fix_anscombe)
```
TurboGit:
```
if(d->fixed_algo)
{
}
else
{
}
```

src/iop/denoiseprofile.c
```diff
@@ -1792,9 +2007,36 @@ static int process_wavelets_cl(struct dt_iop_module_t *self, dt_dev_pixelpipe_io
...
-  const float wb[4] = { 2.0f * piece->pipe->dsc.processed_maximum[0] * d->strength * (scale * scale),
-                        piece->pipe->dsc.processed_maximum[1] * d->strength * (scale * scale),
-                        2.0f * piece->pipe->dsc.processed_maximum[2] * d->strength * (scale * scale), 0.0f };
+  const float wb_mean = (piece->pipe->dsc.temperature.coeffs[0] + piece->pipe->dsc.temperature.coeffs[1]
+                         + piece->pipe->dsc.temperature.coeffs[2])
+                        / 3.0f;
+  // we init wb by the mean of the coeffs, which corresponds to the mean
+  // amplification that is done in addition to the "ISO" related amplification
+  float wb[4] = { wb_mean, wb_mean, wb_mean, 0.0f };
+  if(d->fix_anscombe)
```
TurboGit:
```
if(d->fixed_algo)
...
```

src/iop/denoiseprofile.c
```diff
@@ -2774,24 +3086,70 @@ void gui_init(dt_iop_module_t *self)
...
+
   g->stack = gtk_stack_new();
   gtk_stack_set_homogeneous(GTK_STACK(g->stack), FALSE);
+  gtk_box_pack_start(GTK_BOX(self->widget), g->profile, TRUE, TRUE, 0);
+  gtk_box_pack_start(GTK_BOX(self->widget), g->wb_adaptive_anscombe, TRUE, TRUE, 0);
+  gtk_box_pack_start(GTK_BOX(self->widget), g->mode, TRUE, TRUE, 0);
   gtk_box_pack_start(GTK_BOX(self->widget), g->stack, TRUE, TRUE, 0);
+  gtk_box_pack_start(GTK_BOX(self->widget), g->strength, TRUE, TRUE, 0);
+
+  // add collapsable section for those extra options that are generally not to be used
```
TurboGit: Replace all this by a single boolean option and only displayed if (d->fixed_alfo is FALSE)

rawfiner:
Thank you for reviewing and testing.
I don't mind removing the backward compatibility section, it will simplify the GUI.
I do not know which solution is the best for the user:
- if we hide completely the options the user may find it confusing that from one photo (which had already a denoise profile instance) to another (a new one) he gets very different results with similar settings.
- if we show the option only in case of an old edit, the GUI code is a little bit more complicated, but the user can see something changes. Yet, it may be confusing too to see an option on an image and don't see it on another one?
As such, I do not know which solution to choose. Do you have a preference?

TurboGit:
I will go for 2 and if the option is named:

`[  ] migrate to new algorithm`

And in the tooltip message you can have a longer explanation about the ascomb and patch fix and the fact that when selected it won't be possible to go back later.

It will not be confusing to not see this in new edits I would say.

rawfiner:
Ok, thanks for the idea, I will do that :-)

rawfiner:
@TurboGit , should be better now.
Yet, I have some GUI issues, I don't manage yet to make the option not appear on new instances

src/iop/denoiseprofile.c
```diff
@@ -2116,11 +2352,14 @@ void reload_defaults(dt_iop_module_t *module)
...
     }
 
+    ((dt_iop_denoiseprofile_params_t *)module->default_params)->wb_adaptive_anscombe = TRUE;
     ((dt_iop_denoiseprofile_params_t *)module->default_params)->radius = 1.0f;
     ((dt_iop_denoiseprofile_params_t *)module->default_params)->nbhood = 7.0f;
     ((dt_iop_denoiseprofile_params_t *)module->default_params)->scattering = 0.0f;
+    ((dt_iop_denoiseprofile_params_t *)module->default_params)->central_pixel_weight = 0.0f;
     ((dt_iop_denoiseprofile_params_t *)module->default_params)->strength = 1.0f;
     ((dt_iop_denoiseprofile_params_t *)module->default_params)->mode = MODE_NLMEANS;
+    ((dt_iop_denoiseprofile_params_t *)module->default_params)->fix_anscombe_and_nlmeans_norm = TRUE;
```
TurboGit: By default it should be FALSE, no?

src/iop/denoiseprofile.c
```diff
@@ -2148,6 +2387,8 @@ void init(dt_iop_module_t *module)
       tmp.y[ch][k] = 0.5f;
     }
   }
+  tmp.fix_anscombe_and_nlmeans_norm = TRUE;
```
TurboGit: FALSE also here?

rawfiner:
> In mode_callback you also need to hide the (fix anscomb...) after:
> gtk_stack_set_visible_child_name(GTK_STACK(g->stack), "wavelets");

I added this in the fix_anscombe_and_nlmeans_norm_callback, I think it makes more sense than in the mode callback.
Basically, it makes the button disapear when the user click on it.

And your commit 1589f0d has solved my issue with new instances, many thanks! :-) \o/

TurboGit:
> I added this in the fix_anscombe_and_nlmeans_norm_callback, I think it makes more sense than in the mode callback.

I don't think this is correct. A user may want to activate and deactivate it multiple time at different zoom level to asses the changes and if he want to keep the previous algorithm or not.

rawfiner:
> > I added this in the fix_anscombe_and_nlmeans_norm_callback, I think it makes more sense than in the mode callback.
> 
> I don't think this is correct. A user may want to activate and deactivate it multiple time at different zoom level to asses the changes and if he want to keep the previous algorithm or not.

Ok, I have removed the commit. Do you think we should do it when the user change mode however?

TurboGit:
> Do you think we should do it when the user change mode however?

When changing mode we should probably keep the current status of the fix boolean.

rawfiner:
> > Do you think we should do it when the user change mode however?
> 
> When changing mode we should probably keep the current status of the fix boolean.

Sorry, I am not sure I understand correctly.
If I understand correctly, you mean that changing mode we should not do anything, and keep the checkbox shown if it is shown?

TurboGit:
Exactly. And not shown if not shown.

rawfiner:
> Exactly. And not shown if not shown.

Ok :-)
I was confused with the remark about the mode_callback.
Thanks for clarifying
Last pushed version should work as expected.

TurboGit:
Ok, UI fixed now, thanks. But the presets are not working on my side now.

rawfiner:
Good remark. I will update them again

rawfiner:
The presets should be fixed
I also updated one of the tooltip text, which was inaccurate

edgardoh:
This is amazing! I have never been able to get good results with this denoise until now, thanks soooo much!

TurboGit:
Really nice indeed! Great work rawfiner. Thanks.

rawfiner:
@edgardoh @TurboGit Thanks for the feedback :-)
I am glad you like it !

And thanks for the code reviews !

rawfiner:
@TurboGit just to check, the modifications seem to appear twice in master history (all my commits + a "merge" commit that contains all the changes) is that normal?

TurboGit:
I don't see that on my side.

edgardoh:
I'm having an issue, when exporting downscaled is like the module is disabled, there's no effect at all, if I export in HQ then things go back to normal. Is this expected?

rawfiner:
@edgardoh This issue was already here in 2.6 (and before).
The module still has some effect, but the effects are different than the ones we could expect: for instance there is undersmoothing.
The problem is that it is hard to adapt the algorithms to downsampling.
If we note I the original image, D the downsampling, f the denoising function, p its parameters, it is hard to find parameters p' such as:
D(f(I,p)) = f(D(I),p')
For now, the parameters p' which are estimated are not good enough.
If you want the result to be accurate with what you see at 100% zoom, you have to select the option "do high quality resampling during export", so that downsampling is down at the end of the pipe and not at the beginning.

edgardoh:
OK, thanks for the explanation.

edgardoh:
@rawfiner , I don't know if you are working on making this scale-independent, but since right now results are not usable on a downscaled image, it is possible to just skip the process? This is a heavy module, so this will make the pipe faster. Maybe choose a scale and anything below just copy input -> output.

rawfiner:
@edgardoh Yes, at least I will try to make it more scale-independent. I am not in favor of skipping completely the process, I would rather have a better preview. We will see if this is feasible (I have some ideas, but not tested them yet). If if does not work correctly, maybe should we tell the user to zoom at 100%, and do copies at other zoom levels, as you suggest. But I hope the preview will improve enough so that we don't need to do that.

edgardoh:
That sounds great, thanks!

Issue #1981, opened 2019-01-08
Make color checker module compatible with RGB Cube LUTs


aurelienpierre:
There are plenty of good film-emulation LUTs out there (https://www.colorgradingcentral.com/free-luts/) used for video editors. These are RGB LUTs taking usually log-encoded input from cameras, and outputing linear RGB (ACES or REC 2020).

Making darktable colorchecker module able to read these LUTs would enable users to take advantage of the existing film-emulations databases. These are really difficult to reproduce with the color-checker tool because they need a film scanner properly neutralized/calibrated, to provide a scanner-agnostic film reference, then a properly neutralized raw file, to provide a camera-agnostic base for the LUT grading, and only then it becomes possible to transfer the film look. This is not possible without a full lab and adequate ressources, and darktable options only allow local work-arounds. Also, we don't want to fully apply the film characteristics (especially the tone curve), since its dynamic range can be more limited than digital for color emulsions, but we might want to retain the advantages of digital (shadows recovery) and the feeling of the film (color, smooth transitions).

Instead, making the colorchecker capable of reading cube LUTs would require a log transformation of the input, then a modification of the actual algo from Lab to RGB. The output would need to be converted back to Lab, although if #1841 achieves to provide a full RGB pipe, that would not be necessary anymore.

Technical details on the 3D Cube LUTs can be found here : 

- http://download.autodesk.com/us/systemdocs/pdf/lustre_color_management_user_guide.pdf
- http://juanmelara.com.au/blog/print-film-emulation-luts-for-download

aurelienpierre:
Note that @houz has something like that working, but not it dt

phweyland:
Looking a bit at this, colorchecker is designed for 49 patches max and is a kind of 1D LUT file (while not RGB but Lab).
3D LUT files can have 256x256x256 patches max. I've opened some with LUT_3D_SIZE 64 (64x64x64 patches). The corresponding file size is above 7Mb. That will impact seriously the preset size.

>>Note that @houz has something like that working, but not it dt
Where we can find this ?

aurelienpierre:
> The corresponding file size is above 7Mb.

Are you sure ? The LUTs I have tested weight 970 kB max. Given that they are only text files, 7 MB seems quite heavy. Also, we don't have to store them in the database, they can be retrieved from the files like ICC profiles when needed.

phweyland:
[...]

PR #2128, opened 2019-02-23
3D LUT module

phweyland:
I've reused the work of https://github.com/PkmX/darktable which apply haldclut (png files) and adapted it to apply 3D LUT cube files.
There is still some work to be done and some choices as well. I open a PR to get feedback.
It is not just a color grading tool, 3D Lut can be applied to mimic a lot a treatments: camera profile, log transform, black & white transform ... so the position in pixelpipe may be critical (and features of the PR 1841 useful).
See https://github.com/darktable-org/darktable/issues/1981 opened by @aurelienpierre.

### DIFF
--- a/data/kernels/lut3d.cl
+++ b/data/kernels/lut3d.cl
@@ -217,3 +217,16 @@ lut3d_pyramid(read_only image2d_t in, write_only image2d_t out, const int width,
   write_imagef(out, (int2)(x, y), output);
 }
 
+kernel void
+lut3d_none(read_only image2d_t in, write_only image2d_t out, const int width, const int height)
+{
+  const int x = get_global_id(0);
+  const int y = get_global_id(1);
+
+  if(x >= width || y >= height) return;
+
+  float4 input = read_imagef(in, sampleri, (int2)(x, y));
+  
+  write_imagef(out, (int2)(x, y), input);
+}
+
--- a/data/kernels/lut3d.cl
+++ b/data/kernels/lut3d.cl
@@ -1,0 +1,225 @@
+/*
+    This file is part of darktable,
+    copyright (c) 2019 philippe weyland.
+
+    darktable is free software: you can redistribute it and/or modify
+    it under the terms of the GNU General Public License as published by
+    the Free Software Foundation, either version 3 of the License, or
+    (at your option) any later version.
+
+    darktable is distributed in the hope that it will be useful,
+    but WITHOUT ANY WARRANTY; without even the implied warranty of
+    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+    GNU General Public License for more details.
+
+    You should have received a copy of the GNU General Public License
+    along with darktable.  If not, see <http://www.gnu.org/licenses/>.
+*/
+
+#include "common.h"
+
+kernel void
+lut3d_tetrahedral(read_only image2d_t in, write_only image2d_t out, const int width, const int height,
+           global float *clut, const int level)
+{
+  int4 rgbi = (int4)(0);
+  float4 rgbd = (float4)(0.0f);
+  const int level2 = level * level;
+  const int x = get_global_id(0);
+  const int y = get_global_id(1);
+
+  if(x >= width || y >= height) return;
+
+  float4 input = read_imagef(in, sampleri, (int2)(x, y));
+  float4 output = (float4)(0.0f);
+
+  input.x = (input.x < 0.0f) ? 0.0f : (input.x > 1.0f) ? 1.0f : input.x;
+  input.y = (input.y < 0.0f) ? 0.0f : (input.y > 1.0f) ? 1.0f : input.y;
+  input.z = (input.z < 0.0f) ? 0.0f : (input.z > 1.0f) ? 1.0f : input.z;
+
+  rgbd = input * (float)(level - 1); 
+  rgbi = min( max( convert_int4(rgbd), 0), (int)(level - 2));
+
+// delta r, g, b
+  rgbd = rgbd - convert_float4(rgbi); 
+
+// indexes of P000 to P111 in clut
+  const int color = rgbi.x + rgbi.y * level + rgbi.z * level2;
+  const int i000 = color * 3 ;  // P000
+  const int i100 = i000 + 3;  // P100
+  const int i010 = (int)(color + level) * 3;  // P010
+  const int i110 = i010 + 3;  //P110
+  const int i001 = (int)(color + level2) * 3;  // P001
+  const int i101 = i001 + 3;  // P101
+  const int i011 = (int)(color + level + level2) * 3;  // P011
+  const int i111 = i011 + 3;  // P111
+
+// float3 are 16 bytes aligned. So no way to set clut as an array of float3.
+  const float4 clut000 = (float4)(clut[i000], clut[i000+1], clut[i000+2], 0.0f);
+  const float4 clut100 = (float4)(clut[i100], clut[i100+1], clut[i100+2], 0.0f);
+  const float4 clut010 = (float4)(clut[i010], clut[i010+1], clut[i010+2], 0.0f);
+  const float4 clut110 = (float4)(clut[i110], clut[i110+1], clut[i110+2], 0.0f);
+  const float4 clut001 = (float4)(clut[i001], clut[i001+1], clut[i001+2], 0.0f);
+  const float4 clut101 = (float4)(clut[i101], clut[i101+1], clut[i101+2], 0.0f);
+  const float4 clut011 = (float4)(clut[i011], clut[i011+1], clut[i011+2], 0.0f);
+  const float4 clut111 = (float4)(clut[i111], clut[i111+1], clut[i111+2], 0.0f);
+
+  if (rgbd.x > rgbd.y)
+  {
+    if (rgbd.y > rgbd.z)
+    {
+      output = (1.0f-rgbd.x)*clut000 + (rgbd.x-rgbd.y)*clut100 + (rgbd.y-rgbd.z)*clut110 + rgbd.z*clut111;
+    }
+    else if (rgbd.x > rgbd.z)
+    {
+      output = (1.0f-rgbd.x)*clut000 + (rgbd.x-rgbd.z)*clut100 + (rgbd.z-rgbd.y)*clut101 + rgbd.y*clut111;
+    }
+    else
+    {
+      output = (1.0f-rgbd.z)*clut000 + (rgbd.z-rgbd.x)*clut001 + (rgbd.x-rgbd.y)*clut101 + rgbd.y*clut111;
+    }
+  }
+  else
+  {
+    if (rgbd.z > rgbd.y)
+    {
+      output = (1.0f-rgbd.z)*clut000 + (rgbd.z-rgbd.y)*clut001 + (rgbd.y-rgbd.x)*clut011 + rgbd.x*clut111;
+    }
+    else if (rgbd.z > rgbd.x)
+    {
+      output = (1.0f-rgbd.y)*clut000 + (rgbd.y-rgbd.z)*clut010 + (rgbd.z-rgbd.x)*clut011 + rgbd.x*clut111;
+    }
+    else
+    {
+      output = (1.0f-rgbd.y)*clut000 + (rgbd.y-rgbd.x)*clut010 + (rgbd.x-rgbd.z)*clut110 + rgbd.z*clut111;
+    }
+  }
+  output.w = input.w;
+  write_imagef(out, (int2)(x, y), output);
+}
+
+kernel void
+lut3d_trilinear(read_only image2d_t in, write_only image2d_t out, const int width, const int height,
+           global float *clut, const uint level)
+{
+  int4 rgbi = (int4)(0);
+  float4 rgbd = (float4)(0.0f);
+  const int level2 = level * level;
+  const int x = get_global_id(0);
+  const int y = get_global_id(1);
+  float4 tmp1;
+  float4 tmp2;
+  
+  if(x >= width || y >= height) return;
+
+  float4 input = read_imagef(in, sampleri, (int2)(x, y));
+  float4 output = (float4)(0.0f);
+
+  input.x = (input.x < 0.0f) ? 0.0f : (input.x > 1.0f) ? 1.0f : input.x;
+  input.y = (input.y < 0.0f) ? 0.0f : (input.y > 1.0f) ? 1.0f : input.y;
+  input.z = (input.z < 0.0f) ? 0.0f : (input.z > 1.0f) ? 1.0f : input.z;
+
+  rgbd = input * (float)(level - 1); 
+  rgbi = min( max( convert_int4(rgbd), 0), (int)(level - 2));
+
+  // delta r, g, b
+  rgbd = rgbd - convert_float4(rgbi); 
+
+  // indexes of P000 to P111 in clut
+  const int color = rgbi.x + rgbi.y * level + rgbi.z * level2;
+  const int i000 = color * 3 ;  // P000
+  const int i100 = i000 + 3;  // P100
+  const int i010 = (int)(color + level) * 3;  // P010
+  const int i110 = i010 + 3;  //P110
+  const int i001 = (int)(color + level2) * 3;  // P001
+  const int i101 = i001 + 3;  // P101
+  const int i011 = (int)(color + level + level2) * 3;  // P011
+  const int i111 = i011 + 3;  // P111
+
+  // float3 are 16 bytes aligned. So no way to set clut as an array of float3.
+  const float4 clut000 = (float4)(clut[i000], clut[i000+1], clut[i000+2], 0.0f);
+  const float4 clut100 = (float4)(clut[i100], clut[i100+1], clut[i100+2], 0.0f);
+  const float4 clut010 = (float4)(clut[i010], clut[i010+1], clut[i010+2], 0.0f);
+  const float4 clut110 = (float4)(clut[i110], clut[i110+1], clut[i110+2], 0.0f);
+  const float4 clut001 = (float4)(clut[i001], clut[i001+1], clut[i001+2], 0.0f);
+  const float4 clut101 = (float4)(clut[i101], clut[i101+1], clut[i101+2], 0.0f);
+  const float4 clut011 = (float4)(clut[i011], clut[i011+1], clut[i011+2], 0.0f);
+  const float4 clut111 = (float4)(clut[i111], clut[i111+1], clut[i111+2], 0.0f);
+  
+  tmp1 = clut000*(1.0f-rgbd.x) + clut100*rgbd.x;
+  tmp2 = clut010*(1.0f-rgbd.x) + clut110*rgbd.x;
+  output = tmp1*(1.0f-rgbd.y) + tmp2*rgbd.y;
+  tmp1 = clut001*(1.0f-rgbd.x) + clut101*rgbd.x;
+  tmp2 = clut011*(1.0f-rgbd.x) + clut111*rgbd.x;
+  tmp1 = tmp1*(1.0f-rgbd.y) + tmp2*rgbd.y;
+  output = output*(1.0f-rgbd.z) + tmp1*rgbd.z;
+
+  output.w = input.w;
+  write_imagef(out, (int2)(x, y), output);
+}
+
+kernel void
+lut3d_pyramid(read_only image2d_t in, write_only image2d_t out, const int width, const int height,
+           global float *clut, const uint level)
+{
+  int4 rgbi = (int4)(0);
+  float4 rgbd = (float4)(0.0f);
+  const int level2 = level * level;
+  const int x = get_global_id(0);
+  const int y = get_global_id(1);
+
+  if(x >= width || y >= height) return;
+
+  float4 input = read_imagef(in, sampleri, (int2)(x, y));
+  float4 output = (float4)(0.0f);
+
+  input.x = (input.x < 0.0f) ? 0.0f : (input.x > 1.0f) ? 1.0f : input.x;
+  input.y = (input.y < 0.0f) ? 0.0f : (input.y > 1.0f) ? 1.0f : input.y;
+  input.z = (input.z < 0.0f) ? 0.0f : (input.z > 1.0f) ? 1.0f : input.z;
+
+  rgbd = input * (float)(level - 1); 
+  rgbi = min( max( convert_int4(rgbd), 0), (int)(level - 2));
+
+  // delta r, g, b
+  rgbd = rgbd - convert_float4(rgbi); 
+
+  // indexes of P000 to P111 in clut
+  const int color = rgbi.x + rgbi.y * level + rgbi.z * level2;
+  const int i000 = color * 3 ;  // P000
+  const int i100 = i000 + 3;  // P100
+  const int i010 = (int)(color + level) * 3;  // P010
+  const int i110 = i010 + 3;  //P110
+  const int i001 = (int)(color + level2) * 3;  // P001
+  const int i101 = i001 + 3;  // P101
+  const int i011 = (int)(color + level + level2) * 3;  // P011
+  const int i111 = i011 + 3;  // P111
+
+  // float3 are 16 bytes aligned. So no way to set clut as an array of float3.
+  const float4 clut000 = (float4)(clut[i000], clut[i000+1], clut[i000+2], 0.0f);
+  const float4 clut100 = (float4)(clut[i100], clut[i100+1], clut[i100+2], 0.0f);
+  const float4 clut010 = (float4)(clut[i010], clut[i010+1], clut[i010+2], 0.0f);
+  const float4 clut110 = (float4)(clut[i110], clut[i110+1], clut[i110+2], 0.0f);
+  const float4 clut001 = (float4)(clut[i001], clut[i001+1], clut[i001+2], 0.0f);
+  const float4 clut101 = (float4)(clut[i101], clut[i101+1], clut[i101+2], 0.0f);
+  const float4 clut011 = (float4)(clut[i011], clut[i011+1], clut[i011+2], 0.0f);
+  const float4 clut111 = (float4)(clut[i111], clut[i111+1], clut[i111+2], 0.0f);
+
+  if (rgbd.y > rgbd.x && rgbd.z > rgbd.x)
+  {
+    output = clut000 + (clut111-clut011)*rgbd.x + (clut010-clut000)*rgbd.y + (clut001-clut000)*rgbd.z
+      + (clut011-clut001-clut010+clut000)*rgbd.y*rgbd.z;
+  }
+  else if (rgbd.x > rgbd.y && rgbd.z > rgbd.y)
+  {
+    output = clut000 + (clut100-clut000)*rgbd.x + (clut111-clut101)*rgbd.y + (clut001-clut000)*rgbd.z
+      + (clut101-clut001-clut100+clut000)*rgbd.x*rgbd.z;
+  }
+  else
+  {
+    output = clut000 + (clut100-clut000)*rgbd.x + (clut010-clut000)*rgbd.y + (clut111-clut110)*rgbd.z
+      + (clut110-clut100-clut010+clut000)*rgbd.x*rgbd.y;
+  }
+  output.w = input.w;
+  write_imagef(out, (int2)(x, y), output);
+}
+
--- a/data/kernels/lut3d.cl
+++ b/data/kernels/lut3d.cl
@@ -33,10 +33,8 @@ lut3d_tetrahedral(read_only image2d_t in, write_only image2d_t out, const int wi
   float4 input = read_imagef(in, sampleri, (int2)(x, y));
   float4 output = (float4)(0.0f);
 
-  input.x = (input.x < 0.0f) ? 0.0f : (input.x > 1.0f) ? 1.0f : input.x;
-  input.y = (input.y < 0.0f) ? 0.0f : (input.y > 1.0f) ? 1.0f : input.y;
-  input.z = (input.z < 0.0f) ? 0.0f : (input.z > 1.0f) ? 1.0f : input.z;
-
+  input = clamp(input, (float4)0.0f, (float4)1.0f);
+  
   rgbd = input * (float)(level - 1); 
   rgbi = min( max( convert_int4(rgbd), 0), (int)(level - 2));
 
@@ -115,9 +113,7 @@ lut3d_trilinear(read_only image2d_t in, write_only image2d_t out, const int widt
   float4 input = read_imagef(in, sampleri, (int2)(x, y));
   float4 output = (float4)(0.0f);
 
-  input.x = (input.x < 0.0f) ? 0.0f : (input.x > 1.0f) ? 1.0f : input.x;
-  input.y = (input.y < 0.0f) ? 0.0f : (input.y > 1.0f) ? 1.0f : input.y;
-  input.z = (input.z < 0.0f) ? 0.0f : (input.z > 1.0f) ? 1.0f : input.z;
+  input = clamp(input, (float4)0.0f, (float4)1.0f);
 
   rgbd = input * (float)(level - 1); 
   rgbi = min( max( convert_int4(rgbd), 0), (int)(level - 2));
@@ -173,9 +169,7 @@ lut3d_pyramid(read_only image2d_t in, write_only image2d_t out, const int width,
   float4 input = read_imagef(in, sampleri, (int2)(x, y));
   float4 output = (float4)(0.0f);
 
-  input.x = (input.x < 0.0f) ? 0.0f : (input.x > 1.0f) ? 1.0f : input.x;
-  input.y = (input.y < 0.0f) ? 0.0f : (input.y > 1.0f) ? 1.0f : input.y;
-  input.z = (input.z < 0.0f) ? 0.0f : (input.z > 1.0f) ? 1.0f : input.z;
+  input = clamp(input, (float4)0.0f, (float4)1.0f);
 
   rgbd = input * (float)(level - 1); 
   rgbi = min( max( convert_int4(rgbd), 0), (int)(level - 2));
--- a/src/common/colorspaces.c
+++ b/src/common/colorspaces.c
@@ -1259,8 +1259,8 @@ dt_colorspaces_t *dt_colorspaces_init()
                                      ++work_pos, ++display2_pos));
 
   res->profiles = g_list_append(res->profiles, _create_profile(DT_COLORSPACE_REC709, dt_colorspaces_create_gamma_rec709_rgb_profile(),
-                                     _("gamma Rec709 RGB"), ++in_pos, ++out_pos, ++display_pos, ++category_pos,
-                                     ++work_pos, ++display2_pos));
+                                     _("gamma Rec709 RGB"), ++in_pos, ++out_pos, -1, -1,
+                                     ++work_pos, -1));
                                      
   res->profiles = g_list_append(
       res->profiles, _create_profile(DT_COLORSPACE_LIN_REC2020, dt_colorspaces_create_linear_rec2020_rgb_profile(),
--- a/src/common/colorspaces.c
+++ b/src/common/colorspaces.c
@@ -325,6 +325,19 @@ static cmsHPROFILE dt_colorspaces_create_brg_profile()
   return profile;
 }
 
+static cmsHPROFILE dt_colorspaces_create_gamma_rec709_rgb_profile(void)
+{
+  cmsFloat64Number srgb_parameters[5] = { 2.2, 1.0 / 1.099,  0.099 / 1.099, 1.0 / 4.5, 0.081 };
+  cmsToneCurve *transferFunction = cmsBuildParametricToneCurve(NULL, 4, srgb_parameters);
+
+  cmsHPROFILE profile = _create_lcms_profile("Gamma Rec709 RGB", "Gamma Rec709 RGB",
+                                             &rec709_primaries_pre_quantized, transferFunction, TRUE);
+
+  cmsFreeToneCurve(transferFunction);
+
+  return profile;
+}
+
 // Create the ICC virtual profile for adobe rgb space
 static cmsHPROFILE dt_colorspaces_create_adobergb_profile(void)
 {
@@ -1245,6 +1258,10 @@ dt_colorspaces_t *dt_colorspaces_init()
                                      _("linear Rec709 RGB"), ++in_pos, ++out_pos, ++display_pos, ++category_pos,
                                      ++work_pos, ++display2_pos));
 
+  res->profiles = g_list_append(res->profiles, _create_profile(DT_COLORSPACE_REC709, dt_colorspaces_create_gamma_rec709_rgb_profile(),
+                                     _("gamma Rec709 RGB"), ++in_pos, ++out_pos, ++display_pos, ++category_pos,
+                                     ++work_pos, ++display2_pos));
+                                     
   res->profiles = g_list_append(
       res->profiles, _create_profile(DT_COLORSPACE_LIN_REC2020, dt_colorspaces_create_linear_rec2020_rgb_profile(),
                                      _("linear Rec2020 RGB"), ++in_pos, ++out_pos, ++display_pos, ++category_pos,
--- a/src/common/colorspaces.h
+++ b/src/common/colorspaces.h
@@ -72,7 +72,8 @@ typedef enum dt_colorspaces_color_profile_type_t
   DT_COLORSPACE_SOFTPROOF = 17,
   DT_COLORSPACE_WORK = 18,
   DT_COLORSPACE_DISPLAY2 = 19,
-  DT_COLORSPACE_LAST = 20
+  DT_COLORSPACE_REC709 = 20,
+  DT_COLORSPACE_LAST = 21
 } dt_colorspaces_color_profile_type_t;
 
 typedef enum dt_colorspaces_color_mode_t
--- a/src/common/colorspaces_inline_conversions.h
+++ b/src/common/colorspaces_inline_conversions.h
@@ -121,7 +121,7 @@ static inline __m128 dt_sRGB_to_XYZ_sse2(__m128 rgb)
   rgb = _mm_or_ps(_mm_and_ps(mask, rgb0), _mm_andnot_ps(mask, rgb1));
 
   __m128 XYZ
-      = srgb_to_xyz_0 * _mm_shuffle_ps(rgb, rgb, _MM_SHUFFLE(0, 0, 0, 0)) + 
+      = srgb_to_xyz_0 * _mm_shuffle_ps(rgb, rgb, _MM_SHUFFLE(0, 0, 0, 0)) +
         srgb_to_xyz_1 * _mm_shuffle_ps(rgb, rgb, _MM_SHUFFLE(1, 1, 1, 1)) +
         srgb_to_xyz_2 * _mm_shuffle_ps(rgb, rgb, _MM_SHUFFLE(2, 2, 2, 2));
   return XYZ;
--- a/src/common/colorspaces_inline_conversions.h
+++ b/src/common/colorspaces_inline_conversions.h
@@ -236,18 +236,6 @@ static inline void dt_XYZ_to_sRGB(const float *const XYZ, float *sRGB)
     sRGB[c] = rgb[c] <= 0.0031308 ? 12.92 * rgb[c] : (1.0 + 0.055) * powf(rgb[c], 1.0 / 2.4) - 0.055;
 }
 
-/** uses D50 white point. */
-static inline void dt_XYZ_to_linear_sRGB(const float *const XYZ, float *sRGB)
-{
-  const float xyz_to_srgb_matrix[3][3] = { { 3.1338561, -1.6168667, -0.4906146 },
-                                           { -0.9787684, 1.9161415, 0.0334540 },
-                                           { 0.0719453, -0.2289914, 1.4052427 } };
-
-  // XYZ -> sRGB
-  for(int r = 0; r < 3; r++)
-    for(int c = 0; c < 3; c++) sRGB[r] += xyz_to_srgb_matrix[r][c] * XYZ[c];
-}
-
 /** uses D50 white point and clips the output to [0..1]. */
 static inline void dt_XYZ_to_sRGB_clipped(const float *const XYZ, float *sRGB)
 {
@@ -274,18 +262,6 @@ static inline void dt_sRGB_to_XYZ(const float *const sRGB, float *XYZ)
     rgb[c] = sRGB[c] <= 0.04045 ? sRGB[c] / 12.92 : powf((sRGB[c] + 0.055) / (1 + 0.055), 2.4);
   for(int r = 0; r < 3; r++)
     for(int c = 0; c < 3; c++) XYZ[r] += srgb_to_xyz[r][c] * rgb[c];
-}
-
-static inline void dt_linear_sRGB_to_XYZ(const float *const sRGB, float *XYZ)
-{
-  const float srgb_to_xyz[3][3] = { { 0.4360747, 0.3850649, 0.1430804 },
-                                    { 0.2225045, 0.7168786, 0.0606169 },
-                                    { 0.0139322, 0.0971045, 0.7141733 } };
-
-  // sRGB -> XYZ
-  XYZ[0] = XYZ[1] = XYZ[2] = 0.0;
-  for(int r = 0; r < 3; r++)
-    for(int c = 0; c < 3; c++) XYZ[r] += srgb_to_xyz[r][c] * sRGB[c];
 }
 
 static inline void dt_XYZ_to_prophotorgb(const float *const XYZ, float *rgb)
--- a/src/common/colorspaces_inline_conversions.h
+++ b/src/common/colorspaces_inline_conversions.h
@@ -236,6 +236,18 @@ static inline void dt_XYZ_to_sRGB(const float *const XYZ, float *sRGB)
     sRGB[c] = rgb[c] <= 0.0031308 ? 12.92 * rgb[c] : (1.0 + 0.055) * powf(rgb[c], 1.0 / 2.4) - 0.055;
 }
 
+/** uses D50 white point. */
+static inline void dt_XYZ_to_linear_sRGB(const float *const XYZ, float *sRGB)
+{
+  const float xyz_to_srgb_matrix[3][3] = { { 3.1338561, -1.6168667, -0.4906146 },
+                                           { -0.9787684, 1.9161415, 0.0334540 },
+                                           { 0.0719453, -0.2289914, 1.4052427 } };
+
+  // XYZ -> sRGB
+  for(int r = 0; r < 3; r++)
+    for(int c = 0; c < 3; c++) sRGB[r] += xyz_to_srgb_matrix[r][c] * XYZ[c];
+}
+
 /** uses D50 white point and clips the output to [0..1]. */
 static inline void dt_XYZ_to_sRGB_clipped(const float *const XYZ, float *sRGB)
 {
@@ -262,6 +274,18 @@ static inline void dt_sRGB_to_XYZ(const float *const sRGB, float *XYZ)
     rgb[c] = sRGB[c] <= 0.04045 ? sRGB[c] / 12.92 : powf((sRGB[c] + 0.055) / (1 + 0.055), 2.4);
   for(int r = 0; r < 3; r++)
     for(int c = 0; c < 3; c++) XYZ[r] += srgb_to_xyz[r][c] * rgb[c];
+}
+
+static inline void dt_linear_sRGB_to_XYZ(const float *const sRGB, float *XYZ)
+{
+  const float srgb_to_xyz[3][3] = { { 0.4360747, 0.3850649, 0.1430804 },
+                                    { 0.2225045, 0.7168786, 0.0606169 },
+                                    { 0.0139322, 0.0971045, 0.7141733 } };
+
+  // sRGB -> XYZ
+  XYZ[0] = XYZ[1] = XYZ[2] = 0.0;
+  for(int r = 0; r < 3; r++)
+    for(int c = 0; c < 3; c++) XYZ[r] += srgb_to_xyz[r][c] * sRGB[c];
 }
 
 static inline void dt_XYZ_to_prophotorgb(const float *const XYZ, float *rgb)
--- a/src/common/iop_order.c
+++ b/src/common/iop_order.c
@@ -61,7 +61,8 @@ static int _ioppr_legacy_iop_order_step(GList **_iop_order_list, GList *history_
     _ioppr_move_iop_before(_iop_order_list, "colorout", "clahe", dont_move);
     _ioppr_insert_iop_after(_iop_order_list, history_list, "basicadj", "colorin", dont_move);
     _ioppr_insert_iop_after(_iop_order_list, history_list, "rgbcurve", "levels", dont_move);
-
+    _ioppr_insert_iop_after(_iop_order_list, history_list, "lut3d", "velvia", dont_move);
+    
     new_version = 2;
   }
 
@@ -173,7 +174,6 @@ static GList *_ioppr_get_iop_order_v1()
                                                   { 0.0, "vignette" },
                                                   { 0.0, "splittoning" },
                                                   { 0.0, "velvia" },
-                                                  { 0.0, "lut3d" },
                                                   { 0.0, "clahe" },
                                                   { 0.0, "finalscale" },
                                                   { 0.0, "overexposed" },
--- a/src/common/iop_order.c
+++ b/src/common/iop_order.c
@@ -48,7 +48,7 @@ static void _ioppr_move_iop_before(GList **_iop_order_list, const char *op_curre
 static int _ioppr_legacy_iop_order_step(GList **_iop_order_list, GList *history_list, const int old_version, const int dont_move)
 {
   int new_version = -1;
-
+  
   if(0) // I left this for now so I don't get the unused error, we'll add some modules soon and we'll remove it
   {
     _ioppr_insert_iop_before(_iop_order_list, history_list, "dummy1", "colorin", dont_move);
@@ -67,7 +67,7 @@ static int _ioppr_legacy_iop_order_step(GList **_iop_order_list, GList *history_
 
   if(new_version <= 0)
     fprintf(stderr, "[_ioppr_legacy_iop_order_step] missing step migrating from version %i\n", old_version);
-
+  
   return new_version;
 }
 
@@ -173,6 +173,7 @@ static GList *_ioppr_get_iop_order_v1()
                                                   { 0.0, "vignette" },
                                                   { 0.0, "splittoning" },
                                                   { 0.0, "velvia" },
+                                                  { 0.0, "lut3d" },
                                                   { 0.0, "clahe" },
                                                   { 0.0, "finalscale" },
                                                   { 0.0, "overexposed" },
@@ -203,7 +204,7 @@ static GList *_ioppr_get_iop_order_v1()
 dt_iop_order_entry_t *dt_ioppr_get_iop_order_entry(GList *iop_order_list, const char *op_name)
 {
   dt_iop_order_entry_t *iop_order_entry = NULL;
-
+  
   GList *iops_order = g_list_first(iop_order_list);
   while(iops_order)
   {
@@ -226,7 +227,7 @@ double dt_ioppr_get_iop_order(GList *iop_order_list, const char *op_name)
 {
   double iop_order = DBL_MAX;
   dt_iop_order_entry_t *order_entry = dt_ioppr_get_iop_order_entry(iop_order_list, op_name);
-
+  
   if(order_entry)
     iop_order = order_entry->iop_order;
 
@@ -239,14 +240,14 @@ double dt_ioppr_get_iop_order(GList *iop_order_list, const char *op_name)
 static void _ioppr_insert_iop_before(GList **_iop_order_list, GList *history_list, const char *op_new, const char *op_next, const int check_history)
 {
   GList *iop_order_list = *_iop_order_list;
-
+  
   // check that the new operation don't exists on the list
   if(dt_ioppr_get_iop_order_entry(iop_order_list, op_new) == NULL)
   {
     // create a new iop order entry
     dt_iop_order_entry_t *iop_order_new = (dt_iop_order_entry_t*)calloc(1, sizeof(dt_iop_order_entry_t));
     snprintf(iop_order_new->operation, sizeof(iop_order_new->operation), "%s", op_new);
-
+    
     // search for the previous one
     int position = 0;
     int found = 0;
@@ -264,10 +265,10 @@ static void _ioppr_insert_iop_before(GList **_iop_order_list, GList *history_lis
       }
       iop_order_prev = order_entry->iop_order;
       position++;
-
+      
       iops_order = g_list_next(iops_order);
     }
-
+    
     // now we have to check if there's a module with iop_order between iop_order_prev and iop_order_next
     if(found)
     {
@@ -277,22 +278,22 @@ static void _ioppr_insert_iop_before(GList **_iop_order_list, GList *history_lis
         while(history)
         {
           dt_dev_history_item_t *hist = (dt_dev_history_item_t *)(history->data);
-
+          
           if(hist->iop_order >= iop_order_prev && hist->iop_order <= iop_order_next)
             iop_order_prev = hist->iop_order;
-
+          
           history = g_list_next(history);
         }
       }
     }
     else
       fprintf(stderr, "[_ioppr_insert_iop_before] module %s don't exists on iop order list\n", op_next);
-
+    
     if(found)
     {
       // set the iop_order
       iop_order_new->iop_order = iop_order_prev + (iop_order_next - iop_order_prev) / 2.0;
-
+      
       // insert it on the proper order
       iop_order_list = g_list_insert(iop_order_list, iop_order_new, position);
     }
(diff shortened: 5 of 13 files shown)


phweyland:
I've placed the module on the lab part of the pixelpipe (before tonecure).
I've moved the clut encoding from process() to commit() (in fact should have been the comment of last git commit).
To apply a log 3D curve one has just to use unbreak module to provide a compatible input.

phweyland:
I've played a little with some free 3D lut files I've found on the web. That's sometime amazing.
Some of them seem to work with gamma sRGB D50 (as converted from Lab in dt).
But a lot of them are specified for these colorspaces:
- Generic log. Not sure that corresponds to unbreak log + prophotoRGB ...
- REC709 
- BMD Film Color Space
In any case white balance should be done before applying the lut.

phweyland:
@TurboGit 
I haven't escaped the rawspeed issue here:
`$ git diff lut3d master`
gives:

```
diff --git a/src/external/rawspeed b/src/external/rawspeed
index 0b91a7447..d761f50b1 160000
--- a/src/external/rawspeed
+++ b/src/external/rawspeed
@@ -1 +1 @@
-Subproject commit 0b91a7447ab33994dbb7dc9c249793218a2a4e00
+Subproject commit d761f50b13a32448f9d8968794c75a2ef4860525
```

`$ git submodule update` doesn't say anything (I guess it's ok).

I haven't yet got what I'm doing wrong...

TurboGit:
You have again committed rawspeed in commit :
e1f7fb686bc75f49171012e74196381d1ff7c10d

Of course :
> $ git submodule update

Says nothing since you have committed rawspeed, so now nothing to update from your branch.

You want to always run "git submodule update" before committing (or at least *everytime* you swicth branches).

Now, well I suppose I'll have to do the clean-up.

phweyland:
I'm sorry. I can try to do it.

TurboGit:
You need to play with rebase --interactive and ask for editing the commit with rawspeed.

phweyland:
I'm there but I don't figure out what to do ... 
should I just remove the 2 lines mentioning rawspeed ?
image

TurboGit:
I think you should have done:

> $ git commit --amend
(to commit the staged rawspeed revert)

And then:
> $ git reset --hard
(to discard the wrong rawspeed)

And then:

> $ git rebase --continue

src/iop/lut3d.c
```diff
@@ -583,8 +593,8 @@ setvbuf(stdout, NULL, _IONBF, 0);
   g->colorspace = dt_bauhaus_combobox_new(self);
   dt_bauhaus_widget_set_label(g->colorspace, NULL, _("application color space"));
   dt_bauhaus_combobox_add(g->colorspace, _("linear RGB"));
-  dt_bauhaus_combobox_add(g->colorspace, _("gamma sRGB"));
-  dt_bauhaus_combobox_add(g->colorspace, _("LAB"));
+  dt_bauhaus_combobox_add(g->colorspace, _("sRGB"));
```
aurelienpierre: I don't think there should be any LUT produced on gamma-encoded data. If this is the case, usually the LUT is used to linearize the input.

phweyland: I don't understand here.
Gamma sRGB anf REC.709 are very close from each other. Same matrix, just the gamma curve is a bit different. A lot of LUT are given to be applied on REC.709 which is a gamma-encoded transformation by definition.

aurelienpierre: REC709 defines the RGB primaries used in sRGB but has no gamma itself, it's a linear, scene-referred space. Gamma sRGB is just sRGB. Gamma REC709 can be sRGB. Now, HDTVs use REC709 primaries upon which they add a gamma, but that's none of our business in image processing.

You can't do linear alegebra on non-linear data, so applying an sRGB/sRGB LUT makes no sense, you need to revert the gamma, apply the LUT, and apply the gamma back. But I don't know any corporate LUT maker dumb enough to use sRGB. Usually, you want very large RGB spaces to apply LUTs.

Some LUTs can be applied on log-encoded video footage, and will therefore revert the log and remap the output to scene-referred. That's a particular way to bring back footage to scene-referred space.

But this does not affect us here, and more importantly, you don't want to correct colors on anything but linear. The LUT defines color samples for the remapping. For colors between the samples, an interpolation is used, if it is a linear interpolation and you are in log/gamma encoded in and out, the interpolation error will increase because it assumes the patches are evenly spaced, when gamma compresses highlights and raise midtones.

Really, we should remove any non-linear conversion from the pipeline and keep the gamma and stuff as the very last step before sending pixels to display. Gamma will mess-up every linear algebra you want to do: blur, rotation, color matching, etc.

phweyland: Our previous comments are related to display referenced LUT.
For those which are scene referenced, things are less clear for me. In particular for the log LUT. We have the option unbreak(log) + lut3d(prophotoregb). Is that the right option ? Any other way to rebuild the LOG colorspace ?

phweyland: > REC709 defines the RGB primaries used in sRGB but has no gamma itself, it's a linear, scene-referred space.

All the references I've found associate REC709 to TV display and give the definition with gamma.  Could you please provide some links where I could find better reference ?

aurelienpierre:
Sorry but that code is no good.

The interpolation of RGB triplets (3D vector) should use a tetrahedral interpolation, to retain the relationship between the RGB code values. See https://blogs.mathworks.com/steve/2006/11/24/tetrahedral-interpolation-for-colorspace-conversion/

Here, the interpolation is a simple linear interpolation on the de-correlated RGB channels. So the color are not preserved. 

But… actually, you are right. I did some checking, and the Cube LUTs and such actually use gamma encoding, for legacy reasons. That's not a problem when you use a tetrahedral interpolation, but this is what will happen with linear interpolation:
lut

Basically, the linear interpolation will put the intermediate points proportionaly to their distance to the closest samples, which is wrong. It's not that bad if the linear interpolation is applied to linear data, but the fact that the channels are not correlated is bad anyway, and defeats the purpose.

You can try hacking the interpolation here with that: http://www.hpl.hp.com/techreports/98/HPL-98-95.pdf

Also, I think it should supports 16 bits LUTs as well.

phweyland:
I can believe that the tetrahedral interpolation respect better the color than trilinear one due the interpolation along the grey axis. Some papers are not so affirmative : http://ijetch.org/papers/318-T860.pdf.
I understand also your schemas making the difference of interpolation on linear data and log/gamma data.
But I don't see the relationship you make between linear or log/gamma and the interpolation method, tetrahedral or trilinear. Both are linear and will deal better with linear data, correct ?

I haven't completely decoded yet the algo from eskil steenberg (quelsolaar) but I'll make a try for tetrahedral interpolation.

> Also, I think it should supports 16 bits LUTs as well, and maybe floating point too.

The current implementation uses floating point values ([0.0,1.0]).

aurelienpierre:
> But I don't see the relationship you make between linear or log/gamma and the interpolation method, tetrahedral or trilinear. Both are linear and will deal better with linear data, correct ?

Both are linear indeed, and near-perfect interpolations will only occur in linear spaces, so we will have to content ourselves with rough approximations anyway, for log or gamma encoded data. But the approach used in the tetrahedral interpolation is better fitted for what we do, and more widely used.

The paper you pointed out seems suspicious, since they don't measure the ability of the interpolation to fall back on previously known numeric values (build a test LUT, remove some lines, compute them from the interpolation, then see how close you get), but measure the screen display output (that's quite a lot of steps added in-between) and use the delta E as a metric. That's not really how you evaluate the performance of an interpolation, since they evaluate the performance of a whole chain, and that might explain why they find different results from everyone else.

Anyway, I have found code to do that already : https://github.com/imageworks/OpenColorIO/commit/062d6a38a323190a2aebefb061d940be1729358d

phweyland:
> imageworks/OpenColorIO

Great ! I’ll have a look at it ASAP.

phweyland:
I've added tetrahedral mode. Interesting to dive into these lut interpolation methods. :-)
But side by side I've to admit I don't see any difference. I've probably to change my glasses ! :-)

aurelienpierre:
Nice ! It looks like it could easily be vectorizable. On a side note, when you add/multiply/subtract/etc. constants in float, you should write `+ 1.0f` instead of ` + 1`. I think GCC treats them as double otherwise, so they are computed as double and then casted to float, which is slower. I'm not sure about that, but it can't hurt anyway.

For the visual difference, it's always the same… You need either need a protocol to test with charts (for exemple the HaldCLut), and compare the input/output, or ask veteran retouchers where to push to make it break. I will ask around.

phweyland:
I've added the pyramid interpolation. It brings more visible difference on the image and some more noise on histogram than the others. On histogram we can see some difference between the 3 methods.
I've also moved the module toward the end of lab pixelpipe where I find the results more convincing (on REC709 or sRGB). But I'm not sure about the method. Where the priority of the module should be determined ?
I've not found any LUT to be applied on linear RGB so far. The same for Lab and XYZ. Any link would be appreciated.
Remains the case of LUT on LOG space. I can get some result tweaking exposure + unbreak + tonecurve and applying the LOG LUT on linear RGB. But that's far less straight forward.

aurelienpierre:
> I've added the pyramid interpolation. It brings more  visible difference on the image and some more noise on histogram than  the others.

It's definitely good to have several options here.

> But I'm not sure about the method. Where the priority of the module should be determined ?

I think that should go straight before the output color profile.

> I've not found any LUT to be applied on linear RGB so far. The same for Lab and XYZ. Any link would be appreciated.

DaVinci Resolve is free and works on Linux (RH/Fedora/CentOS, and with some tweaking on Debian/Ubuntu), they ship several LUTs for Kodak emulsions, maybe you could see if you can strip them and input them in your module. If not, I can try for you (my TODO list is already packed though).

> Remains the case of LUT on LOG space. I can get some result tweaking exposure + unbreak + tonecurve and applying the LOG LUT on linear RGB. But that's far less straight forward.

That's the kind of specific workflow where the ability to reorder modules in the pipe would really help.  Otherwise, you could reuse the log in your module. I have added that function in `src/develop/imageop_math.f` in #2035 (not merged yet):

```
/** Logarithmic shaper / tone-mapping
 * from https://github.com/ampas/aces-dev/blob/master/transforms/ctl/utilities/ACESutil.Lin_to_Log2_param.ctl
 * */
static inline void lin_to_log2f(float *lin, const float middle_grey,
                                    const float min_exposure, const float dynamic_range)
{
  const float logNorm = (dt_fast_log2f(*lin / middle_grey) - min_exposure) / dynamic_range;

  if( logNorm < 0.0f || *lin <= 0.0f)
  {
    *lin = 0.0f;
  }
  else
  {
    *lin = logNorm;
  }
}

#if defined(__SSE__)
#define zeros_sse _mm_setzero_ps()
#define ones_sse _mm_set1_ps(1.0f)
static inline void lin_to_log2f_sse(__m128 *lin, const __m128 middle_grey,
                                    const __m128 min_exposure, const __m128 dynamic_range)
{
  __m128 positive = _mm_cmpgt_ps(*lin, zeros_sse); // strictly positive
  const __m128 logNorm = (_mm_log2_ps(*lin / middle_grey) - min_exposure)/dynamic_range;

  positive = _mm_and_ps(positive, _mm_cmpge_ps(logNorm, zeros_sse));

  *lin = _mm_or_ps(_mm_and_ps(positive, logNorm), _mm_andnot_ps(positive, zeros_sse));
}
#undef zeros_sse
#undef ones_sse
#endif
```

phweyland:
> DaVinci Resolve is free and works on Linux

On windows too, but unfortunately it doesn't come with any lut and, worse, it doesn't uninstall properly. :-(
I had already those which are on free download (only Log and rec709).


> you could reuse the log

I guess I haven't any other option than to follow LAB -> linear prophotorgb -> log. It's already a bit far from raw but I'll have a look at that.

phweyland:
> you could reuse the log in your module

I did it, it works but the result is not consistent.
Original image with only exposure.
image
With M31 REC709 applied on REC709
image
With M31 Log apply on log2 (`log2 to lin` alllows to disable the inverse transformation and tweak the log parameters with an identity LUT)
image

I've found a Fuji spec which gives their log formula (http://www.fujifilm.com/support/digital_cameras/software/lut/). I'll make a try to see if I can get something better.

phweyland:
With the Fuji log. 

image

TurboGit:
I think the LUT files should be placed into a dedicated places in dt (like icc, watermakrs). I would choose a "lut" directory for this. And in the interface you'll want to only display the base name.

phweyland:
> I think the LUT files should be placed into a dedicated places in dt

Good point. Dedicated, yes that helps. In dt file structure, I'm not sure as LUT files,  and ICC and watermark files as well, can be common to other tools. So I find useful to be able to set the folder somewhere, in dt default folders or elsewhere if needed.
We can already do this in darktablerc where there is a specific line for that: `plugins/darkroom/lut3d/def_path=C:\Documents\Data`. I've set it manually.
It would be excellent to have a global setting to define this root folder. I'll look at that. If any tip to do it please tell me.

> And in the interface you'll want to only display the base name.

base name ? we have 3 parts:
- LUT folder root (global setting)
- sub-folders
- LUT file name

So if the chosen LUT file is under the root folder we should show only sub-folders/filename.ext. In other cases, root not defined, or LUT file not under LUT root, we display the full path. Correct ?

And the same for saved parameters. Still ok ?

TurboGit:
> It would be excellent to have a global setting to define this root folder. I'll look at that. If any tip to 
> do it please tell me.

Why not! That's one more config option where as a dedicated dir as for watermark would not need that. But I'm not opposed to a new pref.

For the 3 parts I would certainly simplify. The root (dedicated or selected in pref) as discussed above and the base name. Why sub-folders? Do we really expect to have many LUT files? Bear with me I have never use one :)

phweyland:
I've found that preferences have an automated process to generate the code but I haven't found any option to select a folder... while I have seen that lua plugins are able to do that ... still some search to do on that. Meanwhile I've that: 
image

TurboGit:
There is no folder selection in the preferences. But you can create a references to store a string (a folder or whatever). But as you are doing all in the module itself you don't need a preference all will be stored into the module params.

phweyland:
So far the folder is saved as a preference and the lut file is saved in params.
This way, if the root folder changes for some reason, the user has just to update it once to remap automatically all the images. If not he should update each image individually. This works as is but I would find more consistent to have the root folder managed in preferences.

TurboGit:
This is not good. If the folder is saved in preferences it should not be seen here. So of you want to handle multiple possible folders (as already discussed) I think both the folder and the lut file should be in params. Acutally I think only the full filename should be in params. And only the base name should be displayed here.

phweyland:
> If the folder is saved in preferences it should not be seen here

Agreed. But I've to find the way to put this part in preferences.



> So of you want to handle multiple possible folders (as already discussed)

There is a misunderstanding here. The root folder is unique. But you can have several sub-folders of lut files below (at least png and cube ones).

TurboGit:
> Agreed. But I've to find the way to put this part in preferences.

I see, and then the folder won't be displayed in the iop then. For the prefs just use a simple string, I don't think there is circuitry to have a file selector there.

rabauke:
I downloaded sources from https://github.com/phweyland/darktable today and compiled the lut3d branch. The new module does not work for me. Essentially I get pictures with more or less random colors (mainly black), no matter which options (color space, etc.) are chosen. I employed some *.cube files, which work fine in Gmic. See https://www.rocketstock.com/free-after-effects-templates/35-free-luts-for-color-grading-videos/

rabauke:
Also random colors with PNG-luts from http://gmic.eu/color_presets/index.shtml .

rabauke:
Is the LUT included into to xmp file? What happens if LUT file is removed? Becomes the xmp file unusable?

TurboGit:
@rabauke : good point, yes the edit becomes broken. this also applies to watermark using external svg.

phweyland:
@rabauke, thanks for testing this. 
I'm trying to solve the conflicts with master. I need a bit of time to do this hoping I won't break everything ...
My build on windows works. Are you on linux ? But I don't see where there can be a difference as soon as files are read properly. 
So far open cl is not coded.
I'm not sure if I do things properly to place the module in pixelpipe. On my side it works, but if the module is not on the lab part it cannot work.
Last point I can think about is exposure which must be more or less correct before applying the lut but I'm sure you have done that.
If you start dt from console do you get any error message ?

Here is a snapshot with one of the lut you talk about (The github version has not yet the split root folder and file)
image

rabauke:
@phweyland I was able to get the module working, at least with hald cluts given by a PNG file. The problem was that your implementation of `calculate_clut_haldclut` assumes that the PNG file has 8 bit bit-depth. A PNG file, however, might have a bit-depth of 1, 2, 4, 8 or 16 bits. Here a quick fix: [fix.txt](https://github.com/darktable-org/darktable/files/2948779/fix.txt)

I assume there is also an issue with reading cube files, which still does not work on my Linux system.

phweyland:
@rabauke, great !
Thanks for the png fix. I'll try to reproduce the issue on windows and, in any case, make an update considering png depth.
About cube,... it assumes float values between [0, 1] (default domain). 
I'm using the function strtod to convert ascii to float: `lclut[i+j] = strtod(token[j], NULL)`
Is there any possibility to have different compiler behavior between linux and windows ?

rabauke:
`strtod` might be a problem. It assumes/requires that the float is given as specified by the current locale. The representation in the lut file is, however, fix, eg. uses a decimal “.” rather than a “,” as required by my German locale. `sscanf` might be more appropriate here.

phweyland:
> sscanf might be more appropriate here.

I've replaced the first line by the second one. It still works for me. If it works for you too I'll push the change.


//        for (int j=0; j < 3; j++) lclut[i+j] = strtod(token[j], NULL);
        for (int j=0; j < 3; j++) sscanf(token[j], "%f", &lclut[i+j] );

parafin:
sscanf is no better than strtod, it's also affected by locale setting. I know a fix in C++, but not for pure C.

rabauke:
I can confirm that cube luts work as expected as soon as I switch to the "C" locale, which uses decimal points rather than decimal commas as the German locale which is the default on my system. As parafin mentioned, we  have to come up with our own implementation of an atof-like functions because _all_ string conversion functions of the C library are locale sensitive.

parafin:
Here's a solution in C++, which we can for example put in separate .cpp file and use through a header with extern "C":
```
#include <locale>
#include <sstream>
static const std::locale& c_locale = std::locale("C");
int scan_str_to_float(char* string, float* output_value)
{
        float val = 0;
        std::istringstream istring(string);
        istring.imbue(c_locale);
        istring >> val;
        if(istring.fail())
                return 0;
        *output_value = val;
        return 1;
}
```

rabauke:
Here another locale-agnostic implementation of `atof` for parsing floating point numbers. Unlike the C standard implementation it does not recognize numbers in hexadecimal format. (Just an ad hoc implementation, with some basic testing.)
[atof.txt](https://github.com/darktable-org/darktable/files/2949454/atof.txt)

phweyland:
Thanks @rabauke & @parafin for your help.
@rabauke, the atof routine works perfectly (so far :-) ).

phweyland:
> 
> 
> `src/common/iop_order.c` has a lot of changes which just remove white space. Possibly, these changes should by reverted for this pull request. Then then change would reduce to an addition of a single line.

I can do that, for sure. For my understanding what is the reason for that ? Is it not good to clean up as much as we can ?

phweyland:
@rabauke, I've tried several times, copying the iop_order.c from master. Every time the spaces disappear.
I see them in the master file, but I loose them somewhere... :-(

rabauke:
@phweyland What kind of editor or IDE do you use? Some editors can be configured to remove spaces at the end of each line when the file is saved. Look into the configuration of your editor and turn this behavior off or switch temporarily to another editor, preferably a dumb one.

phweyland:
> What kind of editor or IDE do you use? 

I use atom. And yes, there is a setting which removes the trailing spaces and can be disabled.
Thanks !!

TurboGit:
But it should be enabled :) We don't want to have spaces at end of line.

phweyland:
I've moved the selection of root folder from the module to preferences. 
For this I've tried to mimic as far as I could lua/preference.c for choose folder widget. If that works I'm still a bit unsatisfied:
- the folder chooser widget doesn't look as the lua one (I'm using the same functions ... as far as I can see).
- the reset behavior is not as clean as the lua one (reset to (none)).

Now on, I'll try to see what we can improve thanks to @edgardoh's flexible pixelpipe position feature, in particular for log luts. I'm concerned by the fact I do not get comparable results with log and rec709 pair of luts, for example M31:
image
The linear to log transform (to prepare the log lut application) should happen as close as possible from raw data ... but after exposure (?) and white balance (?) ...

Ideas and suggestions are welcome.

phweyland:
> But it should be enabled :) We don't want to have spaces at end of line.

Yes ... I've just disabled it not to pollute to much the iop_order.c file changes as @rabauke requested.

But ... at every addition of module ... this will be necessary. It may have been better to clean it up now.

src/common/colorspaces.c
```diff
@@ -1176,6 +1189,11 @@ dt_colorspaces_t *dt_colorspaces_init()
                                                                dt_colorspaces_create_adobergb_profile(),
                                                                _("Adobe RGB (compatible)"),
                                                                ++in_pos, ++out_pos, ++display_pos, ++category_pos, ++work_pos));
+                                                               
+  res->profiles = g_list_append(res->profiles, _create_profile(DT_COLORSPACE_REC709,
+                                                               dt_colorspaces_create_gamma_rec709_rgb_profile(),
+                                                               _("gamma Rec709 RGB"),
+                                                               -1, ++out_pos, ++display_pos, -1, -1));
```
edgardoh: Are you sure you want this on the display and not in the work category?

src/common/colorspaces.h
```diff
@@ -70,7 +70,8 @@ typedef enum dt_colorspaces_color_profile_type_t
   DT_COLORSPACE_EXPORT = 16, // export and softproof are categories and will return NULL with dt_colorspaces_get_profile()
   DT_COLORSPACE_SOFTPROOF = 17,
   DT_COLORSPACE_WORK = 18,
-  DT_COLORSPACE_LAST = 19
+  DT_COLORSPACE_LAST = 19,
```
edgardoh: DT_COLORSPACE_LAST must remain last, it should be:

DT_COLORSPACE_REC709 = 19,
DT_COLORSPACE_LAST = 20

src/iop/lut3d.c
```diff
@@ -0,0 +1,1200 @@
...
+  GtkWidget *colorspace;
+  GtkWidget *interpolation;
+  GtkWidget *middle_grey;
+  GtkWidget *min_exposure;
+  GtkWidget *dynamic_range;
+  GtkWidget *log2tolin;
+  GtkWidget *log_options;
+} dt_iop_lut3d_gui_data_t;
+
+typedef struct dt_iop_lut3d_global_data_t
```
edgardoh: This should be dt_iop_lut3d_data_t, with the params + this fields, something like this:

typedef struct dt_iop_lut3d_data_t
{
  dt_iop_lut3d_data_t params;
  float *clut;  // cube lut pointer
  uint8_t level; // cube_size
} dt_iop_lut3d_data_t;

src/iop/lut3d.c
```diff
@@ -0,0 +1,1200 @@
...
+    lout[ch] = Norm < 0.0f ? 0.0f : Norm > 1.0f ? 1.0f : Norm;
+  }
+}
+
+void process(struct dt_iop_module_t *self, dt_dev_pixelpipe_iop_t *piece, const void *const ibuf, void *const obuf,
+             const dt_iop_roi_t *const roi_in, const dt_iop_roi_t *const roi_out)
+{
+//printf("process\n");
+  dt_iop_lut3d_params_t *p = (dt_iop_lut3d_params_t *)self->params;
+  dt_iop_lut3d_global_data_t *gp = (dt_iop_lut3d_global_data_t *)self->data;
```
edgardoh: Instead of self->params should be:

dt_iop_lut3d_data_t *d = (dt_iop_lut3d_data_t *)(piece->data);

and remove the global data.

src/iop/lut3d.c
```diff
@@ -0,0 +1,1200 @@
...
+  const dt_interpolation_worker interpolation_worker = (p->interpolation == DT_IOP_TETRAHEDRAL) ? correct_pixel_tetrahedral
+                                                          : (p->interpolation == DT_IOP_TRILINEAR) ? correct_pixel_trilinear
+                                                          : correct_pixel_pyramid;
+
+  const dt_iop_order_iccprofile_info_t *const dest_profile
+      = colorspace == DT_IOP_REC709 ? dt_ioppr_add_profile_info_to_list(self->dev, DT_COLORSPACE_REC709, "", INTENT_PERCEPTUAL)
+      : colorspace == DT_IOP_SRGB ? dt_ioppr_add_profile_info_to_list(self->dev, DT_COLORSPACE_SRGB, "", INTENT_PERCEPTUAL)
+      : colorspace == DT_IOP_LIN_REC709 ? dt_ioppr_add_profile_info_to_list(self->dev, DT_COLORSPACE_LIN_REC709, "", INTENT_PERCEPTUAL)
+      : NULL;
+  const dt_iop_order_iccprofile_info_t *const rgb_profile = dt_ioppr_get_iop_work_profile_info(self->dev);
```
edgardoh: The work profile can be NULL, I know this is a work in progress, just so you know it.

src/iop/lut3d.c
```diff
@@ -0,0 +1,1200 @@
...
+    -6.0f,  // shadows range
+    10.0f,   // dynamic range
+    1  // log2 to lin
+  };
+
+  memcpy(self->params, &tmp, sizeof(dt_iop_lut3d_params_t));
+  memcpy(self->default_params, &tmp, sizeof(dt_iop_lut3d_params_t));
+}
+
+void init_global(dt_iop_module_so_t *self)
```
edgardoh: This is not needed right now.

src/iop/lut3d.c
```diff
@@ -0,0 +1,1200 @@
...
+}
+
+void cleanup(dt_iop_module_t *self)
+{
+printf("cleanup\n");
+  free(self->params);
+  self->params = NULL;
+}
+
+void cleanup_global(dt_iop_module_so_t *self)
```
edgardoh: Same as init_global()

src/iop/lut3d.c
```diff
@@ -0,0 +1,1200 @@
...
+  free(self->data);
+  self->data = NULL;
+}
+
+void commit_params(struct dt_iop_module_t *self, dt_iop_params_t *p1, dt_dev_pixelpipe_t *pipe,
+                   dt_dev_pixelpipe_iop_t *piece)
+{
+printf("commit\n");
+    dt_iop_lut3d_params_t *p = (dt_iop_lut3d_params_t *)self->params;
+    dt_iop_lut3d_global_data_t *gd = (dt_iop_lut3d_global_data_t *)self->data;
```
edgardoh: This should be dt_iop_lut3d_data_t, then you copy params:

memcpy(&d->params, p, sizeof(dt_iop_lut3d_params_t));

and do the calculations on d instead of gd.

src/iop/lut3d.c
```diff
@@ -0,0 +1,1200 @@
...
+      }
+      else if (g_str_has_suffix (p->filepath, ".cube") || g_str_has_suffix (p->filepath, ".CUBE"))
+      {
+        gd->level = calculate_clut_cube(fullpath, &gd->clut);
+      }
+      g_free(fullpath);
+    }
+  }
+
+void init_pipe(struct dt_iop_module_t *self, dt_dev_pixelpipe_t *pipe, dt_dev_pixelpipe_iop_t *piece)
```
edgardoh: void init_pipe(struct dt_iop_module_t *self, dt_dev_pixelpipe_t *pipe, dt_dev_pixelpipe_iop_t *piece)
{
  piece->data = malloc(sizeof(dt_iop_3dlut_data_t));
  self->commit_params(self, self->default_params, pipe, piece);
}

void cleanup_pipe(struct dt_iop_module_t *self, dt_dev_pixelpipe_t *pipe, dt_dev_pixelpipe_iop_t *piece)
{
  free(piece->data);
  piece->data = NULL;
}

src/iop/lut3d.c
phweyland: You attach the data to piece instead of module. Just to understand, why that ?
Is that to simplify the code (and to defer calculation from xx_callback() to commit()) ?
Or does that solve an issue ?

src/common/colorspaces.c
phweyland: I'm not sure... :-)
Some 3D luts are built to be applied on this profile. There may be some other transformations to be  done on that profile.

src/iop/lut3d.c
phweyland: If work profile is NULL, does that mean the module is not between colorin and colorout ? If not what does that mean ?

src/iop/lut3d.c
edgardoh: Not just me, this is how is done. This is to make the pipe thread safe, so the parameters can be updated while the pipe is running.
You're having problems with the pipe right now, let's first make this right and then we can focus on the transforms is still there are issues.

src/common/colorspaces.c
edgardoh: OK, but that has nothing to do with the display group, and since the profile is there maybe someone can use it as a work profile.

src/iop/lut3d.c
edgardoh: Yes, that's what it means.

phweyland:
I've done the suggested changes. The colorspace transform behavior has not changed however.

src/iop/lut3d.c
phweyland: > This is to make the pipe thread safe, so the parameters can be updated while the pipe is running.

Great ! Thanks.

src/iop/lut3d.c
edgardoh: Should be dt_ioppr_get_pipe_work_profile_info(piece->pipe) here instead of dt_ioppr_get_iop_work_profile_info(self->dev)

src/common/iop_order.c
```diff
@@ -60,7 +60,8 @@ static int _ioppr_legacy_iop_order_step(GList **_iop_order_list, GList *history_
     _ioppr_move_iop_before(_iop_order_list, "colorout", "clahe", dont_move);
     _ioppr_insert_iop_after(_iop_order_list, history_list, "basicadj", "colorin", dont_move);
     _ioppr_insert_iop_after(_iop_order_list, history_list, "rgbcurve", "levels", dont_move);
-
+    _ioppr_insert_iop_after(_iop_order_list, history_list, "lut3d", "velvia", dont_move);
```
edgardoh: This may not work as expected and return a NULL work profile. Until is fixed you can change velvia with splittoning.

edgardoh:
There's a bug on my side, I already added PR #2263

phweyland:
Great, that works now. Thanks so much.
About `module, pipe, piece`, ... is there any place which describes how they interact (`piece` in particular)?

edgardoh:
Not that I'm aware of.

src/iop/lut3d.c
phweyland: If work profile is NULL, how can I know if the module is before colorin or after colorout ?

edgardoh:
Check self->iop_order against colorin/colorout->iop_order. There's a
function in iop_order.h that returns colorin module (I don't have the code
right now)

phweyland:
Thanks, this should be this one:
```
/** returns the iop_order from iop_order_list list with operation = op_name */
double dt_ioppr_get_iop_order(GList *iop_order_list, const char *op_name);
```
When I move the module, the first event I see is `commit()`. 
Is it the best place to set properly the UI (I haven't seen specific event for module move) ?

edgardoh:
> Thanks, this should be this one:
>
> /** returns the iop_order from iop_order_list list with operation = op_name */
> double dt_ioppr_get_iop_order(GList *iop_order_list, const char *op_name);
>
> This one returns the default iop_order, not what you want. The one I was
thinking of is not there anymore, sorry. If useful I can add it back.

>  When I move the module, the first event I see is commit().
> Is it the best place to set properly the UI (I haven't seen specific event
> for module move) ?
>
There's no event when a module is moved.

phweyland:
In fact I don't know really what to do for Log LUTs. With the new pixelpipe we can technically apply a log transform on iop_cs_raw data then apply the Log LUT. But this seems useless as we are before colorin.

Log colorspace are usually an encoding of RAW data (Canon, Sony, Cineon, ...) and each manufacturer has its one. 
Then Log LUTs are applied on this log colorspace which is kept until the final transform to display rendering to save the data accuracy.
So it would make sense to use Log LUTs if we could have a log colorspace as a work colorspace in dt (between colorin and colorout).
@aurelienpierre What do you think ?
@edgardoh would the current mechanics support log work colorspace (and corresponding transforms) ? Usually it's something like that:
- x < t1 => a.x + b
- x >= t1 => c.(log(x+d)+e) + f

If that could work we would have a (or some) new log colorspace(s) as an output of colorin.

edgardoh:
Actually, the output of colorin is still lab, the image is transformed to
the rgb work colorspace on demand before the process() of each module is
called.

What is currently needed so an icc profile can be used as a work colorspace
is:
1-lcms2 must be able to handle it
2-dt internal transform routines must be able to transform lab <--> rgb and
rgb <--> rgb

So, and just in theory, if a loc icc profile can be handled by lcms2 and
you are able to code the mentioned transforms this should work as a work
profile.

aurelienpierre:
Otherwise, since the pipe is reorderable, one could just move the profilegamma IOP just before the LUT and fake a manufacturer log curve.

phweyland:
> 1-lcms2 must be able to handle it

I haven't seen any lcms2 mode able to handle it.

> one could just move the profilegamma IOP just before the LUT and fake a manufacturer log curve

I've already included unbreak log2 in the module. But The log2 is different from the manufacturers' ones by the fact they have a threshold (linear part + log part) which clips in some way the darkest points.
Then the transform raw rgb -> prophotorgb introduces some non expected differences (we cannot even control).

From my standpoint we haven't an enough solid background for log LUTs and I suggest to remove the log options.
We could add back them in the future when we clarify the prerequisites.
What do you think ?

aurelienpierre:
> From my standpoint we haven't an enough solid background for log LUTs and I suggest to remove the log options.
> We could add back them in the future when we clarify the prerequisites.
> What do you think ?

agreed. Log LUTs are only a way to reuse LUTs intended for video cameras with log encodings, but since every manufacturer has its own encoding, there is no universal way to handle them. Affected users can tweak something with profilegamma, and I don't think it is useful to overdo it here.

phweyland:
I've not made any open cl or other optimization.
What is the standard now ? What do you recommend ?
There is also the travis failure I would like to solve. Any advice ?

aurelienpierre:
The SIMD/SSE optimization will probably take me an afternoon, so you can leave to me if you want (although I won't have time for a couple of weeks).

OpenCL will be good for the interpolation part, but I suspect the LUT reading/indexing will take a penalty on OpenCL (GPUs are bad at memory accesses).

phweyland:
Working on opencl part. Tetrahedral seems to work. Trilinear and Pyramid have still some issue.
Is float3 as efficient as float4 ?

phweyland:
@aurelienpierre 
process() and process_cl only apply the transformation (not sure however that open_cl improves a lot the response time).
Internal Lut is built when the user changes of lut file and/or when he opens the image. SIMD/SSE optimization can be done later on if necessary (I'm keen to learn this part too).
@TurboGit 
The module is functional and could be merged, if reviewers are ok.

tools/generate_prefs.xsl
```diff
@@ -403,6 +420,22 @@
 </xsl:text>
   </xsl:template>
 
+  <xsl:template match="dtconfig[type='dir']" mode="tab">
+		<xsl:text>    widget = gtk_file_chooser_button_new(_("select directory"), GTK_FILE_CHOOSER_ACTION_SELECT_FOLDER);
+		gtk_file_chooser_button_set_width_chars(GTK_FILE_CHOOSER_BUTTON(widget), 20);
+//		g_object_ref_sink(G_OBJECT(widget));
```
TurboGit: to be removed?

src/iop/lut3d.c
```diff
@@ -0,0 +1,1023 @@
...
+typedef struct dt_iop_lut3d_params_t
+{
+  char filepath[512];
+  int colorspace;
+  int interpolation;
+} dt_iop_lut3d_params_t;
+
+typedef struct dt_iop_lut3d_gui_data_t
+{
+//  GtkWidget *lutfolder;
```
TurboGit: to be removed.

src/iop/lut3d.c
```diff
@@ -0,0 +1,1023 @@
...
+{
+  return _("lut 3D");
+}
+
+int flags()
+{
+  return IOP_FLAGS_INCLUDE_IN_STYLES | IOP_FLAGS_SUPPORTS_BLENDING;
+}
+
+int groups()
```
TurboGit: Should probably be:

int default_group()
{
  return IOP_GROUP_COLOR;
}

?

src/iop/lut3d.c
```diff
@@ -0,0 +1,1023 @@
...
+  const float r = input[0] * (float)(level - 1) - red; // delta red
+  const float g = input[1] * (float)(level - 1) - green; // delta green
+  const float b = input[2] * (float)(level - 1) - blue; // delta blue
+
+// indexes of P000 to P111 in clut
+  const int color = red + green * level + blue * level * level;
+  i = color * 3;  // P000
+  j = (color + 1) * 3;  // P100
+
+  tmp[0] = clut[i++] * (1 - r) + clut[j++] * r;
```
aurelienpierre: Please use `clut[i]`, then `clut[i+1]` and finally `clut[i+2]` instead of `clut[i++]` so, again, the compiler will understand with no ambiguity that we work on contiguous data (again, for auto SIMD). Also, it makes the code more readable.

src/iop/lut3d.c
```diff
@@ -0,0 +1,1023 @@
...
+  const dt_iop_order_iccprofile_info_t *const work_profile
+    = dt_ioppr_get_pipe_work_profile_info(piece->pipe);
+  const gboolean transform = (work_profile != NULL && lut_profile != NULL) ? TRUE : FALSE;
+  if (clut)
+  {
+    if (transform)
+    {
+      dt_ioppr_transform_image_colorspace_rgb(ibuf, obuf, roi_in->width, roi_in->height, work_profile, lut_profile, "linrec2020 to LUT Space");
+#ifdef _OPENMP
+#pragma omp parallel for default(none) schedule(static)
```
aurelienpierre: `pragma omp parallel for simd default(none) schedule(static)`

src/iop/lut3d.c
```diff
@@ -0,0 +1,1023 @@
...
+#endif
+      for(int j = 0; j < roi_out->height; j++)
```
aurelienpierre: You can easily collapse the 2 loops to help the autovectorization and parallel loops splitting:
```C
for(size_t k; k < roi_out->width * roi_out->height * ch; k += ch)
    float *const out = ((float *const)obuf) + k;
    interpolation_worker(out, out, clut, level);
```

src/iop/lut3d.c
```diff
@@ -0,0 +1,1023 @@
...
+        }
+      }
+      dt_ioppr_transform_image_colorspace_rgb(obuf, obuf, roi_in->width, roi_in->height, lut_profile, work_profile, "LUT space to linrec2020");
+    }
+    else
+    {
+#ifdef _OPENMP
+#pragma omp parallel for default(none) schedule(static)
+#endif
+      for(int j = 0; j < roi_out->height; j++)
```
aurelienpierre: same here

src/iop/lut3d.c
```diff
@@ -0,0 +1,1023 @@
...
+  tmp[1] = tmp[1] * (1 - g) + tmp[4] * g;
+  tmp[2] = tmp[2] * (1 - g) + tmp[5] * g;
+
+  output[0] = output[0] * (1 - b) + tmp[0] * b;
+  output[1] = output[1] * (1 - b) + tmp[1] * b;
+  output[2] = output[2] * (1 - b) + tmp[2] * b;
+}
+// from OpenColorIO
+// https://github.com/imageworks/OpenColorIO/blob/master/src/OpenColorIO/ops/Lut3D/Lut3DOp.cpp
+void correct_pixel_tetrahedral(float *input, float *output, const float *const clut, const uint8_t level)
```
aurelienpierre: ` void correct_pixel_tetrahedral(float *const input, float *const output, const float *const restrict clut, const uint8_t level) `

`restrict` avoid fake aliasing

src/iop/lut3d.c
```diff
@@ -0,0 +1,1023 @@
...
+{
+  int red, green, blue;
+  const int level2 = level * level;
+
+  for(int c = 0; c < 3; ++c) input[c] = input[c] < 0.0f ? 0.0f : (input[c] > 1.0f ? 1.0f : input[c]);
```
aurelienpierre: you might want to use `input[c] = fminf(fmaxf(input[c], 0.0f), 1.0f);` instead, because the compiler knows how to autovectorize it.

src/iop/lut3d.c
aurelienpierre: On another note, it might be good to use 2 different buffers, such that `interpolation(in, out, clut, level)`, to make better use of the cache and speed it up a bit (to be tested). This way, you can declare `void correct_pixel_tetrahedral(float *const restrict input, float *const restrict output, const float *const restrict clut, const uint8_t level)`

data/kernels/lut3d.cl
```diff
@@ -0,0 +1,225 @@
...
+  const int i000 = color * 3 ;  // P000
+  const int i100 = i000 + 3;  // P100
+  const int i010 = (int)(color + level) * 3;  // P010
+  const int i110 = i010 + 3;  //P110
+  const int i001 = (int)(color + level2) * 3;  // P001
+  const int i101 = i001 + 3;  // P101
+  const int i011 = (int)(color + level + level2) * 3;  // P011
+  const int i111 = i011 + 3;  // P111
+
+// float3 are 16 bytes aligned. So no way to set clut as an array of float3.
```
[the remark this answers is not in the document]
phweyland: `float3` does work. The problem is that a `float3` array has the same footprint than the equivalent float4 array. As the clut is an array of `float[3]` it cannot be read as an array of `float3`.
We could build the clut as an array of `float[4]` <=> `float4` at the expense of 33% increase of clut size if necessary.

tools/generate_prefs.xsl
phweyland: removed.

src/iop/lut3d.c
phweyland: removed

src/iop/lut3d.c
phweyland: yes

src/common/colorspaces.c
```diff
@@ -325,6 +325,19 @@ static cmsHPROFILE dt_colorspaces_create_brg_profile()
   return profile;
 }
 
+static cmsHPROFILE dt_colorspaces_create_gamma_rec709_rgb_profile(void)
+{
+  cmsFloat64Number srgb_parameters[5] = { 2.2, 1.0 / 1.099,  0.099 / 1.099, 1.0 / 4.5, 0.081 };
+  cmsToneCurve *transferFunction = cmsBuildParametricToneCurve(NULL, 4, srgb_parameters);
+
+  cmsHPROFILE profile = _create_lcms_profile("Gamma Rec709 RGB", "Gamma Rec709 RGB",
```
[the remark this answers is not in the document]
phweyland: Same chromaticity as sRGB but with a different gamma and linear bottom.
https://en.wikipedia.org/wiki/Rec._709 (Transfer characteristics) is a reference I've used but all definitions I've found were equivalent.

data/kernels/lut3d.cl
```diff
@@ -0,0 +1,225 @@
...
+  const int y = get_global_id(1);
+  float4 tmp1;
+  float4 tmp2;
+  
+  if(x >= width || y >= height) return;
+
+  float4 input = read_imagef(in, sampleri, (int2)(x, y));
+  float4 output = (float4)(0.0f);
+
+  input.x = (input.x < 0.0f) ? 0.0f : (input.x > 1.0f) ? 1.0f : input.x;
```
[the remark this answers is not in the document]
phweyland: ok

data/kernels/lut3d.cl
```diff
@@ -0,0 +1,225 @@
...
+  const int level2 = level * level;
+  const int x = get_global_id(0);
+  const int y = get_global_id(1);
+
+  if(x >= width || y >= height) return;
+
+  float4 input = read_imagef(in, sampleri, (int2)(x, y));
+  float4 output = (float4)(0.0f);
+
+  input.x = (input.x < 0.0f) ? 0.0f : (input.x > 1.0f) ? 1.0f : input.x;
```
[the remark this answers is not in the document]
phweyland: ok

data/kernels/lut3d.cl
```diff
@@ -0,0 +1,225 @@
...
```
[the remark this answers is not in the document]
phweyland: ok

src/common/colorspaces_inline_conversions.h
```diff
@@ -236,6 +236,18 @@ static inline void dt_XYZ_to_sRGB(const float *const XYZ, float *sRGB)
     sRGB[c] = rgb[c] <= 0.0031308 ? 12.92 * rgb[c] : (1.0 + 0.055) * powf(rgb[c], 1.0 / 2.4) - 0.055;
 }
 
+/** uses D50 white point. */
+static inline void dt_XYZ_to_linear_sRGB(const float *const XYZ, float *sRGB)
```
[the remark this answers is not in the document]
phweyland: You are correct. I've forgotten to remove this (not used).

src/iop/lut3d.c
```diff
@@ -0,0 +1,1023 @@
...
+  if(red > level - 2) red = level - 2;
+  if(red < 0) red = 0;
+  green = input[1] * (float)(level - 1);
+  if(green > level - 2) green = level - 2;
+  if(green < 0) green = 0;
+  blue = input[2] * (float)(level - 1);
+  if(blue > level - 2) blue = level - 2;
+  if(blue < 0) blue = 0;
+
+  const float r = input[0] * (float)(level - 1) - red; // delta red
```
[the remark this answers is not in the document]
phweyland: done

src/iop/lut3d.c
phweyland: done

src/iop/lut3d.c
phweyland: done

src/iop/lut3d.c
phweyland: Loop done. 
To test the efficiency of the second one I would need to measure the execution time. What I've made so far (pick up system time at the beginning and end of process varies a lot) is not accurate. Is there an existing protocol I could follow ?

src/iop/lut3d.c
phweyland: done

src/iop/lut3d.c
phweyland: good to know. done

src/iop/lut3d.c
aurelienpierre: One basic test would be to launch dt with `-d perf` option and compare the timing. For the loop, I can already tell you it will be better with the collapsed version because OpenMP is able to evenly divide the array between the threads, whereas with 2 nested loops, the last chunk is likely to be smaller than the others.

phweyland:
> I need to test this and to do that I'd like to have a small introduction on how to use this module. Where to get lut files? How to set up? ... Anything for a starter. Thanks.

The introduction point: Issue #1981 opened by @aurelienpierre.
A lut transform is a pre-calculated transform in opposition to a parametric one.
Most of the time the lut transform is used to give a film or style look. But it could be used to apply more technical transforms.

Film industry uses `cube` files which are text files while photography apps like g'mic, gimp, rt or photoflow use `haldclut (png)` files. Both are characterized by the size (or level) of the RGB cube transformation they represent.
Greater the size is, the more accurate the transform is. As the format is more compact the `haldclut` files offer usually higher level.
The transform works properly in a given color space, but there is no color space information in a `cube` file. a color space could be present in a `png` file, but the module considers only the user selected one.
Usually `cube` file works on non linear REC709. Very often you will find Log color space but this is not really covered by the module today (Log profile comes out of cameras and dt pixelpipe is not ready for that).
As far as I know, haldclut file are designed to work on sRGB.

The module. By default it is placed closed before colorout. But it can moved outside colorin-colorout (in that case pixelpipe color profile is used). 
image
Before being able to select a lut file the user has to define the (root) folder where the lut files are stored in preferences (core options / miscellaneous):
image
Depending on the lut file chosen the color space must set (usually sRGB or REC709). Linear spaces are there only for experiment. Other spaces could be added if there is some need.
By default the interpolation is tetrahedral but could be set to trilinear and pyramid. The effect can be more visible with small size luts.

Where to find lut ?
If allowed I can upload some here ...
Otherwize, you will find a lot of referenced sites for cube files here : 
https://jonnyelwyn.co.uk/film-and-video-editing/free-film-luts-for-editors-dits-and-colorists/
16 bits haldclut files can be found here : http://gmic.eu/color_presets/index.shtml
more here : http://rawpedia.rawtherapee.com/Film_Simulation/fr#Collection_de_Simulation_de_Film_de_RawTherapee
some basic 8 bits haldclut here : http://www.quelsolaar.com/files/index.html

src/iop/lut3d.c
phweyland: I thought it was just a matter of copying the pixel data in another place before calling interpolation. But your suggestion is to copy the full image instead, is that correct ?

src/iop/lut3d.c
phweyland: done

src/common/colorspaces.c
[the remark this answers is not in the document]
aurelienpierre: Ok then. But still, the definition on Wikipedia is not clear on the standard it follows and I don't understand this "scene-referred gamma"

data/kernels/lut3d.cl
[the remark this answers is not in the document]
aurelienpierre: Ok, that's fine. You really don't want to increase the size of the buffers you send to the GPU if you don't have to, so that works. I was just curious.

src/iop/lut3d.c
```diff
@@ -0,0 +1,1010 @@
...
+    = (d->params.interpolation == DT_IOP_TETRAHEDRAL) ? correct_pixel_tetrahedral
+    : (d->params.interpolation == DT_IOP_TRILINEAR) ? correct_pixel_trilinear
+    : correct_pixel_pyramid;
+  const int colorspace
+    = (d->params.colorspace == DT_IOP_SRGB) ? DT_COLORSPACE_SRGB
+    : (d->params.colorspace == DT_IOP_REC709) ? DT_COLORSPACE_REC709
+    : (d->params.colorspace == DT_IOP_LIN_REC709) ? DT_COLORSPACE_LIN_REC709
+    : DT_COLORSPACE_LIN_REC2020;
+  const dt_iop_order_iccprofile_info_t *const lut_profile
+    = (colorspace != DT_COLORSPACE_LIN_REC2020) ? dt_ioppr_add_profile_info_to_list(self->dev, colorspace, "", INTENT_PERCEPTUAL) : NULL;
```
edgardoh: Why lut_profile is NULL if colorspace != DT_COLORSPACE_LIN_REC2020?

TurboGit:
I have many issues reported by the compiler:

`In file included from /home/obry/dev/builds/darktable/build/build/src/iop/introspection_lut3d.c:21:
/home/obry/dev/builds/darktable/build/src/src/iop/lut3d.c: In function ‘correct_pixel_trilinear’:
/home/obry/dev/builds/darktable/build/src/src/iop/lut3d.c:123:13: error: implicit declaration of function ‘min’; did you mean ‘fmin’? [-Werror=implicit-function-declaration]
   rgbi[0] = min( max( (int)rgbd[0], 0), level - 2);
             ^~~
             fmin
/home/obry/dev/builds/darktable/build/src/src/iop/lut3d.c:123:18: error: implicit declaration of function ‘max’; did you mean ‘fmax’? [-Werror=implicit-function-declaration]
   rgbi[0] = min( max( (int)rgbd[0], 0), level - 2);
                  ^~~
                  fmax
In file included from /home/obry/dev/builds/darktable/build/build/src/iop/introspection_lut3d.c:21:
/home/obry/dev/builds/darktable/build/src/src/iop/lut3d.c: In function ‘calculate_clut_cube’:
/home/obry/dev/builds/darktable/build/src/src/iop/lut3d.c:565:54: error: format ‘%zu’ expects argument of type ‘size_t’, but argument 3 has type ‘uint32_t’ {aka ‘unsigned int’} [-Werror=format=]
         dt_print(DT_DEBUG_DEV, "[lut3d] allocating %zu bytes for cube lut - level %d\n", buf_size, level);
                                                    ~~^                                   ~~~~~~~~
                                                    %u
/home/obry/dev/builds/darktable/build/src/src/iop/lut3d.c: In function ‘button_clicked’:
/home/obry/dev/builds/darktable/build/src/src/iop/lut3d.c:915:7: error: attempt to use poisoned "strcpy"
       strcpy(filepath, (gchar *)c);
       ^
cc1: all warnings being treated as errors
make[2]: *** [src/iop/CMakeFiles/lut3d.dir/build.make:68: src/iop/CMakeFiles/lut3d.dir/introspection_lut3d.c.o] Error 1
make[1]: *** [CMakeFiles/Makefile2:6236: src/iop/CMakeFiles/lut3d.dir/all] Error 2
make[1]: *** Waiting for unfinished jobs....
[ 86%] Built target colorize
[ 86%] Built target cacorrect
[ 86%] Linking C shared module libhotpixels.so
[ 86%] Built target hotpixels
2019/04/20 16:28:49 : (FAILED) build failed with error 2
`

src/iop/lut3d.c
phweyland: Because DT_COLORSPACE_LIN_REC2020 is the default color space for the module, so we we skip the transform.

edgardoh: The input image may not be in lin rec 2020, that's up to the user.

phweyland:
> I have many issues reported by the compiler:

I'm sorry for that. I don't understand  `error: attempt to use poisoned "strcpy"` ... I've just coded the equivalent.

src/common/colorspaces.c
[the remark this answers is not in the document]
phweyland: > I don't understand this "scene-referred gamma"
Agreed, should be "display-referred" ...
Here the recommendation (conversion opto-électronique):
https://www.itu.int/dms_pubrec/itu-r/rec/bt/R-REC-BT.709-6-201506-I!!PDF-F.pdf

matze:
> I'm sorry for that. I don't understand error: attempt to use poisoned "strcpy" 
Overlapping strings must not be used with `strcpy`. By the way, why not just using `g_path_get_basename`?

phweyland:
> By the way, why not just using `g_path_get_basename`?

As shown on the above screenshot, I want to keep in `filename` the path between root and the file name itself. For example if the root is `c:\data\lut` and the full path is `c:\data\lut\cube\kodak\K_TONE Vintage.cube` filename should be `cube\kodak\K_TONE Vintage.cube`

src/iop/lut3d.c
phweyland: I don't understand the point.
The input image can be in whatever color space, sure.
As the module requests the lin rec 2020 as default color space (which works between colorin and colorout) it get the image in in lin rec 2020. So if the lut color space is also lin rec 2020 there is no need to transform it. If the lut color space is different then the transformation is done.
Otherwise, before colorin, the module receives the image in raw rgb and after colorout in sRGB, without any option to change it.
Where am I wrong ?

parafin:
> > I have many issues reported by the compiler:
> 
> I'm sorry for that. I don't understand `error: attempt to use poisoned "strcpy"` ... I've just coded the equivalent.

strcpy isn't safe, you have to use strncpy instead, which checks the output buffer size

phweyland:
> you have to use strncpy instead,

With `strncpy` the compiler issues an error saying there is a potential overlap.
Then I'm still with the current code which works for me:

```
    if (strcmp(lutfolder, filepath) < 0)
    { // remove root lut folder from file path
      const int j = strlen(lutfolder) + 1;
      int i;
      for(i = 0; filepath[i+j] != '\0'; i++)
        filepath[i] = filepath[i+j];
      filepath[i] = '\0';
    }
```

src/iop/lut3d.c
aurelienpierre: Your interpolation works inplace, meaning it takes the `out` buffer as an input and overwrite it (so *input = *output). When you want to (auto)vectorize the code, it's usually better to use read-only and write-only buffers, so maybe copy the `obuf` somewhere else, then *input ≠ *output and the compiler should be able to perform further optimizations.

phweyland: ... inside optimization routine `input` is not used a lot compared to other calculations. Not sure it could improve really. Should be measured but I haven't time for this currently. However here is the new code if I understand properly (I've also put the `restrict` directive on the routines' header):
```
      for(int i = 0; i < width * height * ch; i+=ch)
      {
        float input[4];
        float *const out = ((float *const)obuf) + i;
        for (int j = 0; j < ch; j++) input[j] = out[j];
        interpolation_worker((float *const)&input, out, clut, level);
      }
```
Is that in line with your suggestion ? If ok, I'll push the new code.

src/iop/lut3d.c
edgardoh: You can't ask for a particular rgb color space, the input image will always be in the work color space. If you want it to be in any particular rgb color space you have to transform it.

phweyland: I see. I've missed that input color profile could change the the work profile.
So I have to check if the work profile is the required one for LUT or not before making the transformation. Correct ?

edgardoh: The dt_ioppr_transform_image_colorspace_rgb() already does that, so you can call it if both profiles are not NULL, or you can do the check before the call if it gives you better performance.

phweyland: Great. Thanks. I let dt_ioppr_transform_image_colorspace_rgb() do the job then.

src/iop/lut3d.c
phweyland: Above code is pushed.

matze:
First of all: great job, I really like the outcome of some of the LUTs I found. Here's a bit of feedback:

- I had to patch two print statements which would otherwise break the build ([print.patch.txt](https://github.com/darktable-org/darktable/files/3108634/print.patch.txt))
- The initial experience is a bit confusing and it took me a while to figure out that I had to set the LUT path in the preferences before the select button would do anything.
- I see no difference in the interpolation methods, maybe just select one to reduce the UI and the amount of code?
- If there is no LUT selected, darktable reports `error -30 enqueue kernel` which looks like an OpenCL error but is not one, you just set the error variable.

phweyland:
@matze Thanks for your feedback. 
>     I had to patch two print statements which would otherwise break the build ([print.patch.txt] 
(https://github.com/darktable-org/darktable/files/3108634/print.patch.txt))

You reverted some of the changes I've made because @TurboGit's build failed. `%zu` was in the code I've re-used. I'm not sure about the right way here. 

>     The initial experience is a bit confusing and it took me a while to figure out that I had to set the LUT path in the preferences before the select button would do anything.

I'll include the information into the field description. That will be more efficient than the initial message which disappears too quickly. 

>     I see no difference in the interpolation methods, maybe just select one to reduce the UI and the amount of code?

With large lut we hardly see a difference. With small ones (from cube world) I can see some changes on the histogram. If everyone agrees, we can keep the tetrahedral interpolation and remove the others for the sake of simplicity.

>     If there is no LUT selected, darktable reports `error -30 enqueue kernel` which looks like an OpenCL error but is not one, you just set the error variable.

I'll clear the Open CL error and set a specific log error.

Please expect some delay for the changes themselves as I've to rebuild my windows dev environment.

phweyland:
Ready to make changes above.
Everyone is OK to remove the choice of interpolation and use the tetrahedral one ?

aurelienpierre:
I think it doesn't hurt to keep both, since each has its advantages. If users don't understand what it is, they can just use the default.

Also I have asked around for the REC 709 "scene-referred" gamma you linked from Wikipedia, it seems that the Wikipedia article is utterly wrong. I will have more details soon.

phweyland:
> Also I have asked around for the REC 709 "scene-referred" gamma you linked from Wikipedia, it seems that the Wikipedia article is utterly wrong. I will have more details soon.

I don't if you have seen this doc I've posted above some days ago. I understand that BT709 is display device referred.
Here the recommendation (conversion opto-électronique):
https://www.itu.int/dms_pubrec/itu-r/rec/bt/R-REC-BT.709-6-201506-I!!PDF-F.pdf

phweyland:
I've let the interpolation options but tried to fix the issues pinpointed by @matze.

TurboGit:
On error reported with clang-7:

```
In file included from src/iop/introspection_lut3d.c:21:
/build/darktable/src/iop/lut3d.c:739:9: error: loop not vectorized: failed explicitly specified loop vectorization [-Werror,-Wpass-failed=loop-vectorize]
#pragma omp parallel for simd default(none) schedule(static)
        ^
1 error generated.
```

TurboGit:
Thanks! Merged.

phweyland:
I've some cleaning on going...

TurboGit:
Our message have crossed. Please in another PR now.

src/iop/lut3d.c
```diff
@@ -0,0 +1,1071 @@
...
+  gtk_box_pack_start(GTK_BOX(g->hbox), button, FALSE, FALSE, 0);
+  g_signal_connect(G_OBJECT(button), "clicked", G_CALLBACK(button_clicked), self);
+
+  gtk_box_pack_start(GTK_BOX(self->widget), GTK_WIDGET(g->hbox), TRUE, TRUE, 0);
+
+  g->colorspace = dt_bauhaus_combobox_new(self);
+  dt_bauhaus_widget_set_label(g->colorspace, NULL, _("application color space"));
+  dt_bauhaus_combobox_add(g->colorspace, _("sRGB"));
+  dt_bauhaus_combobox_add(g->colorspace, _("REC.709"));
+  dt_bauhaus_combobox_add(g->colorspace, _("lin sRGB/REC.709"));
```
TurboGit: "lin" here stand for linear?

Can't we rename this to "sRGB/REC.709 (linear)" ?

Likewise for the next line. What do you think?

phweyland: Yes, it would better with full words: 
```
linear sRGB/REC.709
linear prophoto RGB
```
Another change I'd like to make. At the last minute I've replaced `copyright (c) 2016 Chih-Mao Chen` with my name. I should have kept this line and added another one for myself.

Do I issue a new PR ? Do we wait for more comments ?

PS: When I've rebased last time I had copied the full lut3d.c file into another editor and pasted it again to get back my last changes. That's probably why you have found unexpected crlf into the file.

aurelienpierre: Please be careful with names. sRGB is REC709 primaries + a piece-wise "gamma". There is no linear sRGB. Linear REC709 is REC709 primaries + nothing else. These are standards with precise meanings, it's bad to mess around with their names.

phweyland: If I understand properly we end up with:
_sRGB_ // Linear REC709 + gamma
_REC.709_ // Linear REC709 + gamma (OETF)
_linear REC.709_
_linear prophoto RGB_

It could be useful to provide the source of technical vocabulary we refer to (I've to admit I've read everything and its opposite ...)

aurelienpierre: Ok, to be super strict on the terms:
1. only displays (screens) have a gamma, so every transfer function built into a color space is an OETF (or EOTF, depending if it's an input or output space). What we call gamma for RGB spaces is an abuse and should be called OETF.
2. a color space is defined by RGB primaries (expressed as subvectors of the XYZ space) AND an OETF/EOTF.

Because video cameras don't record raw signals but use an encoding for the sake of storage space, they output a non-linear signal with a built-in OETF. So, to apply a LUT intended for video camera to photographs coming out of a DSLR, we need to convert from camera RGB to linear REC709, then apply a the same OETF (only for the sake of compatibility of the LUT) as the camera one.

Such OETFs could be the Rec.UIT-R BT.709-6 OETF, but could also be any log curve (S-log, N-log, etc.) on top of the Rec709 primaries (as far as I understand).

But sRGB is an output space, for display (not used in recording), so it's "gamma" is an EOTF. I believe it's useful for us only for compatibility with HaldCLUTs.

So, the options we have are :

* sRGB
* Rec.709 + HDTV OETF
* Linear Rec.709
* Linear ProphotoRGB (why this one ?)

phweyland: I follow you on OETF and EOTF (sRGB or Rec709) ends. If Rec709 primaries are used in the middle for the TV world (today), they are not the only ones, Rec2020 can be used as well. On the other hand, the log raw has been introduced to improve the color accuracy in the shadows (out of camera) but I haven't found anything saying it refers specifically to Rec709.

Practically I haven't found any effective color grading lut working in linear Rec709 (in linear prophoto RGB neither). I've let these color space for testing purpose but until we find some luts working with linear spaces we can remove them.

Another point could come in the future. There is no reason to have always input = output color space for the lut transform. We could have a lut working on linear Rec709 and producing an image in Rec709 EOTF for example. But again I haven't not found any reference about this. I've made some tests during the development but without convincing results. 

To keep it simple for the time being, we could just keep:
- sRGB
- Rec.709
both display referred.
In the future we could add the log ones (if we get camera log raw) and the linear ones (if useful). We may also offer to select a different ouput space if the need shows up.

phweyland:
> 
> 
> Is the LUT included into to xmp file? What happens if LUT file is removed? Becomes the xmp file unusable?

Hi @rabauke, with the G'MIC compressed LUTs, it could be worth to include the lut data in the image parameters (and xmp) (if compressed obviously). 
What do you think ?
Any advice to do it ?

Compressed LUTs:
image

PR #2133, opened 2019-02-26
align all memory buffers on 64 bytes blocks

aurelienpierre:
This is a first step to allow casting to AVX2 vectors on supported platforms and enable further SIMD optimizations


TurboGit:
All good to me now. Thanks.

PR #2143, opened 2019-02-28
replace free by g_free

rabauke:
Replace `free` by `g_free` when memory has been allocated by GTK functions. Some minor reformating.

### DIFF
--- a/src/lua/preferences.c
+++ b/src/lua/preferences.c
@@ -254,7 +254,6 @@ static int write_pref(lua_State *L)
 }
 
 
-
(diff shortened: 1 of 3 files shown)


src/lua/preferences.c
```diff
@@ -266,6 +265,8 @@ static void response_callback_enum(GtkDialog *dialog, gint response_id, pref_ele
     g_free(text);
   }
 }
+
```
TurboGit: No strong opinion here but I'll keep a single empty line.

PR #2144, opened 2019-03-01
Update darktable-chart documentation

junkyardsparkle:
Update "base curve" to "tone curve", and add a note about vignetting correction. In-camera correction is seldom as "absolute" as the Lensfun corrections (and recently some Lensfun profiles have even been found to over-correct slightly).


TurboGit:
Looks good to me!

PR #2146, opened 2019-03-02
fix color picker

### LINKED ISSUE #2145: colorpick don't show range on blenif
**Describe the bug**
On some modules when area is selected on the color picker the blend module color picker is de-selected, it seems that happens when the module has a color picker of his own, like the exposure or filmic.

In other modules the color picker is not de-selected but the input range is not displayed, like in the retouch. It seems that this is when the init_single_picker() is used.

**To Reproduce**
Enable parametric mask on any of the modules described above
Select the blend color picker
On the color picker module select "area"
Depending on the module the blend color picker is de-selected, the input range is not displayed or it works as expected.

**Expected behavior**
Input and output range should be displayed on all modules.

edgardoh:
This fixes  #2145

Also:

When the colorpicker modify the values of a module, like the monochrome, the blend colorpicker also was modifying it.

When a module display the colorpicker values, like the tonecurve, the blend colorpicker also was displaying it.

I have ported the exposure and temperature to the new proxy schema.

The colorchecker still needs to be ported, but I have no idea how it works, so someone else will have to do it.

There's still some minor issues, mostly iop related, I'll work on some of them on a separate PR.

### DIFF
--- a/src/develop/imageop.c
+++ b/src/develop/imageop.c
@@ -42,6 +42,7 @@
 #include "gui/accelerators.h"
 #include "gui/gtk.h"
 #include "gui/presets.h"
+#include "gui/color_picker_proxy.h"
 #include "libs/modulegroups.h"
 #ifdef GDK_WINDOWING_QUARTZ
 #include "osx/osx.h"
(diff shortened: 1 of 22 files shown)


edgardoh:
I get the following travis errors generated:

/lib_darktable.dir/common/undo.c.o   -c /build/darktable/src/common/undo.c
/build/darktable/src/common/undo.c:61:1: error: mutex 'self->mutex' is not held on every path through here [-Werror,-Wthread-safety-analysis]
}
^
/build/darktable/src/common/undo.c:57:5: note: mutex acquired here
    dt_pthread_mutex_lock(&self->mutex);
    ^
/build/darktable/src/common/undo.c:67:5: error: releasing mutex 'self->mutex' that was not held [-Werror,-Wthread-safety-analysis]
    dt_pthread_mutex_unlock(&self->mutex);
    ^
2 errors generated.

TurboGit:
@edgardoh : I made a tentative fix for this.

edgardoh:
@TurboGit , I still get the error.

I'm not saying that the fix does not work, just that travis is complaining. And I'm bringing this up just so when you see the error here you'll have an idea of what's going on, unlike travis I'm not complaining.

TurboGit:
Should now be fixed, this is a false positive from clang. Not very clever analysis!

TurboGit:
Nice fix. Thanks.

PR #2149, opened 2019-03-03
Fix strict aliasing issues

cryptomilk:
This is a patchset to fix strict aliasing issues. Those warnings should not be ignored because they are relevant for the optimizer!

### DIFF
--- a/src/common/interpolation.c
+++ b/src/common/interpolation.c
@@ -1480,9 +1480,9 @@ static void dt_interpolation_resample_plain(const struct dt_interpolation *itor,
 #endif
     for(int y = 0; y < roi_out->height; y++)
     {
-      float *i = (float *)((char *)in + (size_t)in_stride * (y + roi_out->y) + x0);
-      float *o = (float *)((char *)out + (size_t)out_stride * y);
-      memcpy(o, i, out_stride);
+      memcpy((char *)out + (size_t)out_stride * y,
+             (char *)in + (size_t)in_stride * (y + roi_out->y) + x0,
+             out_stride);
     }
 #if DEBUG_RESAMPLING_TIMING
     ts_resampling = getts() - ts_resampling;
--- a/tools/CMakeLists.txt
+++ b/tools/CMakeLists.txt
@@ -1,5 +1,3 @@
-CHECK_COMPILER_FLAG_AND_ENABLE_IT(-fno-strict-aliasing)
-
 if (BUILD_CURVE_TOOLS)
     add_subdirectory(basecurve)
 endif()
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 build*/
+obj*/
 *~
 *.pot
 *.mo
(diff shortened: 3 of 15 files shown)


tools/CMakeLists.txt
```diff
@@ -1,5 +1,3 @@
-CHECK_COMPILER_FLAG_AND_ENABLE_IT(-fno-strict-aliasing)
-
```
moy: From the commit message, it seems you are confusing `-Wno-strict-aliasing` and `-fno-strict-aliasing`. The first silents warnings about strict aliasing, but the second allows the compiler to use strict aliasing rules for optimizations (https://gcc.gnu.org/onlinedocs/gcc-4.9.2/gcc/Optimize-Options.html).

Removing this option does not just activate warnings, it also enables potentially dangerous optimizations. Eliminating the warnings does not guarantee that other optimizations are safe. To me, the potential danger of activating strict aliasing is too high compared to the potential performance benefit (which is probably negligible actually, but without numbers this is pure speculation on my part).

cryptomilk: As the strict aliasing issues have been fixed, we can turn that on. I've fixed the commit message.

cryptomilk:
The errors in `src/common/undo.c` seem to be in new code introduced by @TurboGit. @TurboGit did you run this with helgrind?

tools/CMakeLists.txt
moy: The *warnings* have been fixed, but how can you be sure that all the *issues* have been fixed? Warnings are not meant to be exhaustive (the doc actually says explicitly "The warning does not catch all cases", https://gcc.gnu.org/onlinedocs/gcc-4.9.2/gcc/Warning-Options.html#Warning-Options). If the compiler could warn about all dangerous cases, it would mean that it would be able to do exact alias analysis, and there would be no point using the strict aliasing rule.

cryptomilk: But how should we find and fix the source if we hide/disable it?

cryptomilk:
This got a few more patches ...

cryptomilk:
I'm only rebasing this on top of master. No changes made since March 5th.

src/common/interpolation.c
```diff
@@ -1480,9 +1480,9 @@ static void dt_interpolation_resample_plain(const struct dt_interpolation *itor,
 #endif
     for(int y = 0; y < roi_out->height; y++)
     {
-      float *i = (float *)((char *)in + (size_t)in_stride * (y + roi_out->y) + x0);
-      float *o = (float *)((char *)out + (size_t)out_stride * y);
-      memcpy(o, i, out_stride);
+      memcpy((char *)in + (size_t)in_stride * (y + roi_out->y) + x0,
```
TurboGit: The memcpy() parameters seems to have been inverted. Bug?

cryptomilk: Ouch. Yes, you're right. Fix pushed.

cryptomilk:
I've fixed the issue.

TurboGit:
Ok, looks good now. Time to let this in. Thanks.

PR #2150, opened 2019-03-03
Update Polish translation

tomaszg7:
BTW. There is a problem affecting possibly all translations: word `latitude` is used in metadata viewer and filmic module with completely different meaning. Right now it is not possible to translate both occurrences separately.


TurboGit:
Thanks.

PR #2151, opened 2019-03-04
upgrade color picker

edgardoh:
This PR adds:

-ported blend_gui to the new proxy schema
-add display range to the lch and hsl tabs on blend module
-cntrl + click on the color picker display an area on the blend module, tone curve and fill light
-a new color picker on the blend module that sets the range on the slider based on the selected area

### DIFF
--- a/src/common/color_picker.c
+++ b/src/common/color_picker.c
@@ -18,13 +18,15 @@
 
 #include "common/color_picker.h"
 #include "common/darktable.h"
+#include "common/colorspaces_inline_conversions.h"
 #include "develop/format.h"
 #include "develop/imageop.h"
 #include "develop/imageop_math.h"
 
 static void color_picker_helper_4ch_seq(const dt_iop_buffer_dsc_t *dsc, const float *const pixel,
                                         const dt_iop_roi_t *roi, const int *const box, float *const picked_color,
-                                        float *const picked_color_min, float *const picked_color_max)
+                                        float *const picked_color_min, float *const picked_color_max,
+                                        const int cst_to)
 {
   const int width = roi->width;
 
@@ -38,18 +40,20 @@ static void color_picker_helper_4ch_seq(const dt_iop_buffer_dsc_t *dsc, const fl
     for(size_t i = box[0]; i < box[2]; i++)
     {
       const size_t k = 4 * (width * j + i);
-      const float L = pixel[k];
-      const float a = pixel[k + 1];
-      const float b = pixel[k + 2];
-      picked_color[0] += w * L;
-      picked_color[1] += w * a;
-      picked_color[2] += w * b;
-      picked_color_min[0] = fminf(picked_color_min[0], L);
-      picked_color_min[1] = fminf(picked_color_min[1], a);
-      picked_color_min[2] = fminf(picked_color_min[2], b);
-      picked_color_max[0] = fmaxf(picked_color_max[0], L);
-      picked_color_max[1] = fmaxf(picked_color_max[1], a);
-      picked_color_max[2] = fmaxf(picked_color_max[2], b);
+      float Lab[3] = { pixel[k], pixel[k + 1], pixel[k + 2] };
+      if(cst_to == iop_cs_LCh)
+        dt_Lab_2_LCH(pixel + k, Lab);
+      if(cst_to == iop_cs_HSL)
+        dt_RGB_2_HSL(pixel + k, Lab);
+      picked_color[0] += w * Lab[0];
+      picked_color[1] += w * Lab[1];
+      picked_color[2] += w * Lab[2];
+      picked_color_min[0] = fminf(picked_color_min[0], Lab[0]);
+      picked_color_min[1] = fminf(picked_color_min[1], Lab[1]);
+      picked_color_min[2] = fminf(picked_color_min[2], Lab[2]);
+      picked_color_max[0] = fmaxf(picked_color_max[0], Lab[0]);
+      picked_color_max[1] = fmaxf(picked_color_max[1], Lab[1]);
+      picked_color_max[2] = fmaxf(picked_color_max[2], Lab[2]);
     }
   }
 }
@@ -57,7 +61,7 @@ static void color_picker_helper_4ch_seq(const dt_iop_buffer_dsc_t *dsc, const fl
 static void color_picker_helper_4ch_parallel(const dt_iop_buffer_dsc_t *dsc, const float *const pixel,
                                              const dt_iop_roi_t *roi, const int *const box,
                                              float *const picked_color, float *const picked_color_min,
-                                             float *const picked_color_max)
+                                             float *const picked_color_max, const int cst_to)
 {
   const int width = roi->width;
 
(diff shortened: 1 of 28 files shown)


src/common/color_picker.c
```diff
@@ -395,10 +401,14 @@ static void color_picker_helper_xtrans(const dt_iop_buffer_dsc_t *dsc, const flo
 
 void dt_color_picker_helper(const dt_iop_buffer_dsc_t *dsc, const float *const pixel, const dt_iop_roi_t *roi,
                             const int *const box, float *const picked_color, float *const picked_color_min,
-                            float *const picked_color_max)
+                            float *const picked_color_max, const int image_cst, const int picker_cst)
```
TurboGit: const dt_iop_colorspace_type_t image_cst, const dt_iop_colorspace_type_t cst_to

src/develop/imageop.h
```diff
@@ -280,6 +281,12 @@ typedef struct dt_iop_module_t
   float picked_color[4], picked_color_min[4], picked_color_max[4];
   /** place to store the picked color of module output (before blending). */
   float picked_output_color[4], picked_output_color_min[4], picked_output_color_max[4];
+  /** requested colorspace for the color picker, valid options are:
+   * -1: module colorspace
+   * iop_cs_LCh: for Lab modules
+   * iop_cs_HSL: for RGB modules
+   */
+  int picker_cst;
```
TurboGit: dt_iop_colorspace_type_t picker_cst

src/common/color_picker.h
```diff
@@ -23,7 +23,8 @@ struct dt_iop_roi_t;
 
 void dt_color_picker_helper(const struct dt_iop_buffer_dsc_t *dsc, const float *const pixel,
                             const struct dt_iop_roi_t *roi, const int *const box, float *const picked_color,
-                            float *const picked_color_min, float *const picked_color_max);
+                            float *const picked_color_min, float *const picked_color_max,
+                            const int image_cst, const int picker_cst);
```
TurboGit: const dt_iop_colorspace_type_t image_cst, const dt_iop_colorspace_type_t cst_to

src/develop/blend_gui.c
```diff
@@ -547,41 +527,101 @@ static void _blendop_blendif_polarity_callback(GtkToggleButton *togglebutton, dt
   dt_dev_add_history_item(darktable.develop, data->module, TRUE);
 }
 
-
-static void _blendop_blendif_tab_switch(GtkNotebook *notebook, GtkWidget *page, guint page_num,
-                                        dt_iop_gui_blend_data_t *data)
+static int _blendop_blendif_get_picker_colorspace(dt_iop_gui_blend_data_t *bd)
```
TurboGit: static dt_iop_colorspace_type_t _blendop_blendif_get_picker_colorspace(...)

src/develop/blend_gui.c
```diff
@@ -547,41 +527,101 @@ static void _blendop_blendif_polarity_callback(GtkToggleButton *togglebutton, dt
...
 {
-  data->tab = page_num;
-  dt_iop_gui_update_blendif(data->module);
-}
+  int picker_cst = -1;
```
TurboGit: dt_iop_colorspace_type_t picker_cst;

edgardoh:
This is by design, the regular color picker is not deactivated when
changing tabs so I don't deactivate this one. If the user changes tabs
while active probably he wants to select the range for that tab.
The other option is to deactivate the picker, but since we don't "remember"
the area I think this is better.
Remembering the are is in my to-do list, but right now I have no idea how
to do it.

All other fixes are ready.

TurboGit:
Ok, I agree that the behavior makes sense. But in that case why only the selection apply to the first change of tab?

edgardoh:
That was a bug, should be fixed now.

TurboGit:
Ok, looks better now. A little issue still. When the range is selected in any blend channel, the 4 values (just above the slider) are not updated to correspond to the actual selected area.

TurboGit:
Also I find that the feather between the cursor (2 steps actually) is a bit small and creates by default some scale. Of course having a wider feather will select more than the selected area... But maybe 2 steps inside the computer range and 2 steps outside (for a total of 4) will alleviate a bit this issue.

What do you think?

edgardoh:
I don't understand this one...

edgardoh:
The feather is hard-coded to 0.01f (both left and right), the cursors in
the middle are exactly the range values. I don't like hard-coded things but
I can't find a formula that gives good results on all scenarios.
I haven't used it enough to say what the best values may be, but for now I
think is best to leave the cursors in the middle to have the exact range
value (that's the difficult part to achieve by hand). I can increase the
feather if you find it too small, 0.02f will be OK?
This is an auto feature anyway, so we can change it at any point if we find
something better without having to worry about backward compatibility.

TurboGit:
For each slider we have 4 handles to select the start and end of the selection. Two for the start and the distance between them is a "feathered" area where the mask goes for %0 opacity to 100% opacity. Same of the right side for the end.

My point is that the distance (or feathered area range) is of 2 values and this creates hard/abrupt transitions. I'm wondering if a distance of 4 wont be better. Hope this is clearer :)

TurboGit:
Our messages crossed :)

Yes, I suppose 0.02f will be better in more cases.

TurboGit:
> I don't understand this one... 

The 4 values representing the selected range in the sliders are not updated.

edgardoh:
>
> I don't understand this one...
>
> The 4 values representing the selected range in the sliders are not
> updated.
>
I have updated the 0.02f, but either I can't duplicate this or I still
don't get it. When I draw an area the slider is updated, on any of the
tabs.

TurboGit:
Let's look at the screenshot, I have marked in red the values not updated.

capture d ecran de 2019-03-04 19-30-17

edgardoh:
Oh! The labels!

edgardoh:
Its ready, I have also:

-take your suggestion and made the feather around the selected values,
seems nicer this way
-reset the color picker if slider values are changed, it makes strange
things otherwise. This is only for the set values, the regular color picker
is not reset.

TurboGit:
Perfect! Looks good now to me. Thanks.

edgardoh:
@TurboGit , we should do something about the documentation. I won't update the user manual for obvious reasons, but if you ask me by the end of the year what I have done I probably won't remember half of it.
Maybe something as simple as a text file somewhere that can be updated with each feature, so when someone wants to update the manual we can be sure that everything is there.
You can even force to have some description with each new feature.
What do you think?

TurboGit:
Good idea! I was thinking alike just this morning. I suppose the new features can be documented into the RELEASE_NOTES.md  file.

Just list that in [New Features And Changes] section. Works for you?

TurboGit:
And BTW, not sure the obvious reasons are so obvious :) Is that because you're not a native English? I'm not but we have nice people on the project that will review the text to ensure the wording is correct.

edgardoh:
Seems OK to me, let's give it a try. I'll start with the color picker and
we can take from there.

edgardoh:
> And BTW, not sure the obvious reasons are so obvious :) Is that because
> you're not a native English? I'm not but we have nice people on the project
> that will review the text to ensure the wording is correct.
>
Not just I'm not native but my english skills s*ks, I'm not comfortable
writing in english and will be probably more work for everyone involved to
review and re-write than just do it right from the beginning.

PR #2153, opened 2019-03-05
Cmake: use C11 std if available

aurelienpierre:
C11 is mandatory to be able to unlock Fused Multiply Add : https://accurate-algorithms.readthedocs.io/en/latest/ch03fma.html

Right now, Cmake checks for C11 but sets C99. Similarly, it checks for C++14 but sets C++11. This change allows to use the highest version available.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -173,14 +173,22 @@ include_directories(SYSTEM ${CMAKE_CURRENT_SOURCE_DIR}/external)
 # initial compiler flags
 add_definitions("-DHAVE_CONFIG_H")
 
+message(STATUS "Checking for -std=c99 support")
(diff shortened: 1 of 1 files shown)


cryptomilk:
@aurelienpierre I'm already working to clean up this mess, see:

https://github.com/cryptomilk/darktable/tree/master-cmake

However this needs a lot more work.

aurelienpierre:
Ok, that was a quick fix because I'm working on FMA-able code

cryptomilk:
If we require C11 we should really require it and not fall back to C99, this doesn't really work ...

aurelienpierre:
My version here requires only C99, then use C11 if available or falls back to C99.

cryptomilk:
We should choose one, if one uses features of C11 they wont work in a C99 fallback!

TurboGit:

PR #2154, opened 2019-03-05
Cmake: add -fopt-info-vec-missed for Release build

aurelienpierre:
Release build uses `-O3` flag which enable autovectorization. We add `-fopt-info-vec-missed` for this build to get feedback on the loops the compiler was unable to autovectorize. This is done in the spirit of making devs aware of SIMD and push them to write vectorizable code.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -516,7 +516,7 @@ if(NOT CUSTOM_CFLAGS)
   # also, in IOPs, in-loops branches could be forced to be compiled as different loops variants
   set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} ${MARCH} ${DT_REQ_INSTRUCTIONS} -g")
   set(CMAKE_C_FLAGS_RELWITHDEBINFO "${CMAKE_C_FLAGS_RELWITHDEBINFO} -O2")
-  set(CMAKE_C_FLAGS_RELEASE "${CMAKE_C_FLAGS_RELEASE} -O3")
+  set(CMAKE_C_FLAGS_RELEASE "${CMAKE_C_FLAGS_RELEASE} -O3 -fopt-info-vec-missed")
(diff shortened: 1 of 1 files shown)


TurboGit:
But this breaks the release mode on my side. It is not good since most of us are using the debug mode and when a release will be prepared will have lot of errors. So I would be more in favor of adding this switch for the dev build and make it non fatal (warning and not error).

aurelienpierre:
It's not blocking on my side. It's supposed to only give info on the parts the compiler failed to vectorize, and hints. Strange…

TurboGit:
Yes, strange. I'll try to understand. But still don't you think this should be done for the dev build and not the release one? It is too late to display that when building a release.

aurelienpierre:
> But still don't you think this should be done for the dev build and not the release one? It is too late to display that when building a release.

There is no auto-vectorization in debug building, so I think it's pointless. The release build (with `-O3` flag) needs to be used during development too.

rawfiner:
@aurelienpierre If we should always use Release build and not RelWithDebInfo, maybe also change build.sh so that the default build type is Release?
This way, there will be a greater chance that devs will see these autovectorization info

TurboGit:
I'm not sure we want that. With RelWithDebInfo we get a proper backtrace, without it would be hard to debug crashes from releases.

rawfiner:
Yes, I agree.
Maybe I misunderstood, is there auto-vectorization when doing a RelWithDebugInfo, or only when doing a Release?

aurelienpierre:
> Maybe I misunderstood, is there auto-vectorization when doing a RelWithDebugInfo, or only when doing a Release?

Only with `Release`, which uses the `-O3` flag. `RelWithDebugInfo` uses `-O2` and triggers no auto-vectorization.

But auto-vectorization this way is performed by the compiler and is different from OpenMP SIMD pragmas, so I realize now it might not be what we are after nor useful feedback. So, let's drop that.

PR #2155, opened 2019-03-05
gitignore: add rawspeed

aurelienpierre:
Once for all, remove rawspeed from contributor's commits

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,6 +5,7 @@ build*/
 doc/usermanual/darktable-usermanual*.pdf
 doc/api/
 tools/iop_deps.pdf
+src/external/rawspeed/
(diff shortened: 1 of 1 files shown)


TurboGit:
That would be too easy :) This does not work. Go to rawspeed, checkout master and go back to dt root. Do a git status and you'll see that rawspeed IS considered for the commit. Do the commit and you'll see that it has been added.

aurelienpierre:
−_____−

damit.

TurboGit:
I think I have the fix for this:
https://github.com/darktable-org/darktable/commit/c336ce136e3ad6d7439fd3214b1aacaff8a27818

PR #2157, opened 2019-03-05
upgrade colorzones

edgardoh:
This adds a color picker range and histogram to the color zones.

It also changes how color space is handled by the color picker, it didn't work when a module request a different color space than the default.

The histogram is calculated on histogram.c, I added it there because there's a LCh tone curve in progress and there's no point on having the code duplicated on both modules.

The color zones is a work in progress, I would like the curve to have a dynamic number of nodes, but I'll work on that on a separate PR, I think is better to have the color picker fix and the histogram first.

### DIFF
--- a/src/iop/colorzones.c
+++ b/src/iop/colorzones.c
@@ -20,6 +20,7 @@
 #endif
 #include "bauhaus/bauhaus.h"
 #include "common/colorspaces.h"
+#include "common/colorspaces_inline_conversions.h"
 #include "common/darktable.h"
 #include "common/debug.h"
 #include "common/opencl.h"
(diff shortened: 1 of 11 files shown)


src/iop/colorzones.c
```diff
@@ -568,12 +569,19 @@ static void dt_iop_colorzones_get_params(dt_iop_colorzones_params_t *p, const in
   }
 }
 
+static void picker_scale(const float *const in, float *out)
```
TurboGit: We have a picker_scale in tonecurve.c, the implementation is quite different even if somehow similar. Is that expected? Is that because we have two different CS?

edgardoh:
The picker here returns LCh (now), on the tone curve Lab. I have taken this
from this iop, but the blend_gui does something different, no idea which
one is correct, but because of backward compatibility better not to change
it.

TurboGit:
And BTW, not sure the obvious reasons are so obvious :) Is that because you're not a native English? I'm not but we have nice people on the project that will review the text to ensure the wording is correct.

TurboGit:
Noted! Thanks.

PR #2160, opened 2019-03-06
Fix #2138 by making handling of tag "changed" more effective

### LINKED ISSUE #2138: Collecting images with "history altered" lists images with only default modules applied
**Describe the bug**
When using "Collect Images" to find images which images have been edited, by adding the "History" selector and choosing "Altered", images are also shown that have been opened in the darkroom once, without being edited.
Images that have been edited have Flags `l...r..ap...r` and show the "edited" logo. Images that have only the standard modules applied to them because they were opened once have Flags `l......ap....` and do not show the "edited" logo. They are still counted as "Altered"

**To Reproduce**
1. Import some fresh images
2. Open a image in darkroom
3. Without touching any module, go back to the lighttable
4. Use "Collect Images" to filter out images with altered History
5. Find that the image which was previously opened is now displayed 

**Expected behavior**
Just opening an image once in the darkroom should not count as "altering the history". I only want to get the images that I have really edited.

**Screenshots**
edited image
unedited image

I encountered this on:
- Windows using v2.4.4 and v2.6.0
- Arch linux using v2.6.0

Ablinne:
This pull request changes the behaviour of the "changed" tag for images, in order to fix #2138.

Previously, the tag "changed" was set more or less unconditionally on `darktable.c leave()` and `develop.c dt_dev_write_history_ext()`. Now it is set on `dt_dev_add_history_item()`, `dt_history_copy_and_paste_on_image` and removed on `dt_history_delete_on_image()`.

I am not sure if these are all the correct places or if I am missing sommething, but I wanted to put it out there to get some feedback.

### DIFF
--- a/src/common/history.c
+++ b/src/common/history.c
@@ -1098,6 +1098,11 @@ int dt_history_copy_and_paste_on_image(int32_t imgid, int32_t dest_imgid, gboole
   else
     ret_val = _history_copy_and_paste_on_image_overwrite(imgid, dest_imgid, ops);
 
+  /* attach changed tag reflecting actual change */
+  guint tagid = 0;
+  dt_tag_new("darktable|changed", &tagid);
+  dt_tag_attach(tagid, dest_imgid);
+
   /* if current image in develop reload history */
   if(dt_dev_is_current_image(darktable.develop, dest_imgid))
   {
--- a/src/common/history.c
+++ b/src/common/history.c
@@ -1098,11 +1098,6 @@ int dt_history_copy_and_paste_on_image(int32_t imgid, int32_t dest_imgid, gboole
   else
     ret_val = _history_copy_and_paste_on_image_overwrite(imgid, dest_imgid, ops);
 
-  /* attach changed tag reflecting actual change */
-  guint tagid = 0;
-  dt_tag_new("darktable|changed", &tagid);
-  dt_tag_attach(tagid, dest_imgid);
-
   /* if current image in develop reload history */
   if(dt_dev_is_current_image(darktable.develop, dest_imgid))
   {
--- a/src/common/history.c
+++ b/src/common/history.c
@@ -361,6 +361,7 @@ void dt_history_delete_on_image(int32_t imgid)
 
   /* remove darktable|style|* tags */
   dt_tag_detach_by_string("darktable|style%", imgid);
+  dt_tag_detach_by_string("darktable|changed", imgid);
 }
 
 void dt_history_delete_on_selection()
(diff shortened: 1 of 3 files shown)


TurboGit:
This is not correct and does not corresponds to the intent. A darktable|changed tag is only applied when an image has been selected into the darkroom. Copying some history does not count as changed (as in this file has been handled, edited so developed).

The only part to be fixed is as commented into the TODO to only add the tag if the image has been actually changed into the darkroom and not only viewed.

Ablinne:
> Copying some history does not count as changed (as in this file has been handled, edited so developed).
But doesn't this depend on the use case? I often have a series of pictures, that I want to develop using the exact same settings. I edit one of them in the darkroom and copy the history to the remaining pictures. To me, all of the pictures are  then developed. I want to use the "changed" tag to quickly select all pictures that I need to export and in this case I also want those images selected that I have copied the history to. Setting the tag manually after copying the history to the remaining images is something that could easily be forgotten.

Is there maybe another tag that should be used for this purpose? Should I instead introduce a completely new tag?

TurboGit:
Hum, I'm not sure this cover a use case for many users. It is certainly quite rare to not have to crop a bit a picture for example.

We have currently:

```
darktable|changed
darktable|exported
darktable|format|%s
darktable|local-copy
darktable|printed|%s
darktable|style|%s
```

The "changed" read more like what you are proposing I agree. If we had a "darktable|edited" it would have been easier, but we cannot change that. At this point I'm not sure what tag name could be used.

Ablinne:
In my case it's most of the times pictures that are going to be stitched to a panorama. The panorama will be cropped after stitching, so no need to crop/rotate the pictures before.
However, if copied history does not account for `darktable|changed`, I will remove the call in `dt_history_copy_and_paste_on_image()`.
I will just have to remember to set the tag manually in this case or export the pictures immediately.
Is the rest of the changes allright?

TurboGit:
I wen't back to the documentation and there is nothing said about "darktable|changed" so your first proposal may be as correct as my view on this. So after all let's keep your first proposal (I'm sorry) and ensure that darktable|changed is set whenever there is a change in the history.

I think I'm going to add then darktable|edited when a user has actually done some changes in the darkroom.

What do you think?

Ablinne:
Ok, that seems reasonable. I have reverted my latest commit to reinstate my first proposal. Thanks for looking into this.

TurboGit:
Ok, thanks. I'll squash the 3 commits to avoid the commit+revert.

PR #2165, opened 2019-03-07
French update of 2.6 manual

Nilvus:
Here it is the final traduction of lasts update of the 2.6 manual in french.


TurboGit:
Thanks!

PR #2167, opened 2019-03-07
upgrade color zones

edgardoh:
Rework color zones:

-changed the process, now results are what's expected
-make the GUI like the tone curve (but horizontal)

This is still a work in progress, but I'll appreciate some feedback.

To preserve backwards compatibility I added a mode variable, but is hidden. I can add a combo so it can be edited, but the GUI will remain the same.
Or I can just create a new module 'LCh equalizer' and we keep the color zones.

I like this GUI better, but I don't know how it will look like on other displays, to be tested.

The process() don't CLAMP, but maybe the C should be circular, not sure if the Lab <--> LCh tranform take care of that.

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -105,6 +105,16 @@ when updating from the currently stable 2.4.x series, please bear in mind that y
 
 - Add soft boundaries for denoise profile controls.
 
+- The color zones module curve acts now like the tone curve (but horizontal),
+  it has two edit modes: edit by area is the former one, if not checked nodes
+  can be edited like in the tone curve.
+  A new process mode has been added: smooth is the former one, strong is new.
+  An interpolation method has been added that allows to select different types
+  of curves.
+  The color picker allows to select by area with cntrl+click.
+  A new color picker has benn added that creates a curve based on the area
+  selected from the image.
+
 ## Bug fixes
 
 - The color picker support has been fixed by a complete rewrite. It
--- a/data/kernels/basic.cl
+++ b/data/kernels/basic.cl
@@ -1865,7 +1865,7 @@ enum
 
 
 kernel void
-colorzones (read_only image2d_t in, write_only image2d_t out, const int width, const int height, const int channel,
+colorzones_v3 (read_only image2d_t in, write_only image2d_t out, const int width, const int height, const int channel,
             read_only image2d_t table_L, read_only image2d_t table_a, read_only image2d_t table_b)
 {
   const int x = get_global_id(0);
@@ -1908,6 +1908,47 @@ colorzones (read_only image2d_t in, write_only image2d_t out, const int width, c
   pixel.x = L;
   pixel.y = cos(2.0f*M_PI_F*(h + hm)) * Cm * C;
   pixel.z = sin(2.0f*M_PI_F*(h + hm)) * Cm * C;
+
+  write_imagef (out, (int2)(x, y), pixel);
+}
+
+kernel void
+colorzones (read_only image2d_t in, write_only image2d_t out, const int width, const int height, const int channel,
+            read_only image2d_t table_L, read_only image2d_t table_C, read_only image2d_t table_h)
+{
+  const int x = get_global_id(0);
+  const int y = get_global_id(1);
+
+  if(x >= width || y >= height) return;
+
+  float4 pixel = read_imagef(in, sampleri, (int2)(x, y));
+
+  float4 LCh;
+  const float normalize_C = 1.f / (128.0f * sqrt(2.f));
+
+  LCh = Lab_2_LCH(pixel);
+
+  float select = 0.0f;
+  switch(channel)
+  {
+    case DT_IOP_COLORZONES_L:
+      select = LCh.x * 0.01f;
+      break;
+    case DT_IOP_COLORZONES_C:
+      select = LCh.y * normalize_C;
+      break;
+    case DT_IOP_COLORZONES_h:
+    default:
+      select = LCh.z;
+      break;
+  }
+  select = clamp(select, 0.f, 1.f);
+
+  LCh.x *= native_powr(2.0f, 4.0f * (lookup(table_L, select) - .5f));
+  LCh.y *= 2.f * lookup(table_C, select);
+  LCh.z += lookup(table_h, select) - .5f;
+
+  pixel.xyz = LCH_2_Lab(LCh).xyz;
 
   write_imagef (out, (int2)(x, y), pixel);
 }
--- a/src/develop/blend_gui.c
+++ b/src/develop/blend_gui.c
@@ -120,47 +120,6 @@ static const dt_iop_gui_blendif_colorstop_t _gradient_HUE[]
         { 1.0f, { NEUTRAL_GRAY, 0, 0, 1.0 } } };
 
 
-static void _cairo_paint_colorpicker_set_values(cairo_t *cr, gint x, gint y, gint w, gint h, gint flags, void *data)
-{
-  gint s = (w < h ? w : h);
-  cairo_translate(cr, x + (w / 2.0) - (s / 2.0), y + (h / 2.0) - (s / 2.0));
-  cairo_scale(cr, s, s);
-
-  /* draw pipette */
-
-  cairo_set_line_cap(cr, CAIRO_LINE_CAP_ROUND);
-
-  // drop
-  cairo_set_line_width(cr, 0.15);
-  cairo_move_to(cr, 0.08, 1. - 0.01 + 0.05);
-  cairo_line_to(cr, 0.08, 1. - 0.09 + 0.05);
-  cairo_stroke(cr);
-
-  cairo_set_line_width(cr, 0.2);
-  // cross line
-  cairo_move_to(cr, 0.48, 1. - 0.831 + 0.05);
-  cairo_line_to(cr, 0.739, 1. - 0.482 + 0.05);
-  // shaft
-  cairo_move_to(cr, 0.124, 1. - 0.297 + 0.05);
-  cairo_line_to(cr, 0.823, 1. - 0.814 + 0.05);
-  cairo_stroke(cr);
-
-  // end
-  cairo_set_line_width(cr, 0.35);
-  cairo_move_to(cr, 0.823, 1. - 0.814 + 0.05);
-  cairo_line_to(cr, 0.648, 1. - 0.685 + 0.05);
-  cairo_stroke(cr);
-
-  // plus sign
-  cairo_set_line_width(cr, 0.2);
-  cairo_move_to(cr, 0.20, 0.01);
-  cairo_line_to(cr, 0.20, 0.41);
-  cairo_stroke(cr);
-  cairo_move_to(cr, 0.01, 0.20);
-  cairo_line_to(cr, 0.41, 0.20);
-  cairo_stroke(cr);
-}
-
 static void _blendif_scale(dt_iop_colorspace_type_t cst, const float *in, float *out)
 {
   out[0] = out[1] = out[2] = out[3] = out[4] = out[5] = out[6] = out[7] = -1.0f;
@@ -1548,8 +1507,8 @@ void dt_iop_gui_init_blendif(GtkBox *blendw, dt_iop_module_t *module)
     gtk_widget_set_tooltip_text(bd->colorpicker, _("pick GUI color from image\nctrl+click to select an area"));
     gtk_widget_set_size_request(GTK_WIDGET(bd->colorpicker), bs, bs);
 
-    bd->colorpicker_set_values
-        = dtgtk_togglebutton_new(_cairo_paint_colorpicker_set_values, CPF_STYLE_FLAT | CPF_DO_NOT_USE_BORDER, NULL);
+    bd->colorpicker_set_values = dtgtk_togglebutton_new(dtgtk_cairo_paint_colorpicker_set_values,
+                                                        CPF_STYLE_FLAT | CPF_DO_NOT_USE_BORDER, NULL);
     gtk_widget_set_tooltip_text(bd->colorpicker_set_values, _("set the range based on an area from the image\n"
         "click+drag to use the input image\nctrl+click + drag to use the output image"));
     gtk_widget_set_size_request(GTK_WIDGET(bd->colorpicker_set_values), bs, bs);
--- a/src/dtgtk/paint.c
+++ b/src/dtgtk/paint.c
@@ -1195,6 +1195,46 @@ void dtgtk_cairo_paint_colorpicker(cairo_t *cr, gint x, gint y, gint w, gint h, 
   cairo_stroke(cr);
 }
 
+void dtgtk_cairo_paint_colorpicker_set_values(cairo_t *cr, gint x, gint y, gint w, gint h, gint flags, void *data)
+{
+  gint s = (w < h ? w : h);
+  cairo_translate(cr, x + (w / 2.0) - (s / 2.0), y + (h / 2.0) - (s / 2.0));
+  cairo_scale(cr, s, s);
+
+  /* draw pipette */
+
+  cairo_set_line_cap(cr, CAIRO_LINE_CAP_ROUND);
+
+  // drop
+  cairo_set_line_width(cr, 0.15);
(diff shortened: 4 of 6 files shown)


edgardoh:
Added opencl.

rawfiner:
Hi edgardoh
Here is some feedback (even if I guess you already have some of these points on your todo list ;-) ) :
- when dragging the curve, if the pointer goes out of the area, the point goes back to its original place, which makes it hard to drag the curve down to the bottom for instance
- it would be nice, in "hue" selection mode, to have these hues represented in the background of the curve, like it was in the old colorzone module
Apart of that, I like the fact of having a bigger area, and I am looking forward to test this again when these points will be fixed.
Thanks for all the work!

edgardoh:
>
>    - it would be nice, in "hue" selection mode, to have these hues
>    represented in the background of the curve, like it was in the old
>    colorzone module
>
> Not sure if I get this. Right now the histogram, bottom bar and color
picker use the "select by" channel. The background use the modified channel
(the tabs). If the color picker is active, the selected color is used to
draw things, if not, a hard-coded blue color is used.

rawfiner:
> - it would be nice, in "hue" selection mode, to have these hues represented in the background of the curve, like it was in the old colorzone module Not sure if I get this. Right now the histogram, bottom bar and color
> picker use the "select by" channel. The background use the modified channel (the tabs). If the color picker is active, the selected color is used to draw things, if not, a hard-coded blue color is used.

Oh, I tested too quickly I did not try the picker, and did not see this feature!
This is really nice when the picker is used.

Yet, wouldn't it make sense to have a shade of hues instead of using the "blue" value when the color picker is not used? My issue is that I find it harder to spot where is each color on the curve than with previous version of colorzone, when the color picker is not activated. On the contrary, when the color picker is used I like your idea of having the background customized for this particular color, I find this nice to use.

edgardoh:
>
> Yet, wouldn't it make sense to have a shade of hues instead of using the
> "blue" value when the color picker is not used?
>
Sorry but still I don't get it. The background display only the modified
channel (the tabs), if I don't have a base color (i.e. the color picker is
not active) I have to use something for the lightness and saturation, then
the hard-coded color. For the hue I display all the combinations, so for
each x you know the destination color by looking at the y.
Maybe if you describe exactly the active tab/select by/color picker status
I'll get a better understanding of the issue.

edgardoh:
@rawfiner , the dragging issue should be fixed now, thanks for reporting it!

rawfiner:
> Sorry but still I don't get it. The background display only the modified channel (the tabs), if I don't have a base color (i.e. the color picker is not active) I have to use something for the lightness and saturation, then the hard-coded color. For the hue I display all the combinations, so for each x you know the destination color by looking at the y. Maybe if you describe exactly the active tab/select by/color picker status I'll get a better understanding of the issue.

When the color picker is active, I like the backgrounds of your implementation.

When the color picker is not active, I would prefer to have backgrounds similar to the ones of the 2.6 version.
For instance, on the lightness tab, if I select by hue, I see this:
colorzone
What I like with this representation is that I can see easily which part of the curve impacts the greens, which one impacts the red, etc.
With the uniform blue gradient of your curve, I am forced to look at the hue gradient which is shown bellow, and I find it much less convenient.

I would like backgrounds to be similar to this when I select by hue without color picker.
For instance, this would give backgrounds like that (these are examples made quickly, not perfect examples, just to show the idea)
lightness tab:
colorzone_light
saturation tab:
colorzone_saturation

rawfiner:
Thanks for fixing the dragging issue, much better now :-)

Another little GUI bug:
only the curve of the "active" channel is reset when we click on the "reset parameters" button, until we click on other channel:
before clicking on reset parameters:
colorzone1
after:
colorzone2
the curve of active channel is correctly flattened, but not the ones of the inactive channels

edgardoh:
OK, I get it now, I'm displaying only the modified channel and you like
better to have both the modified & select by channels.

I made it this way because on some combinations I find the mix confusing.
For instance, on lightness by saturation I don't really know what I'm
looking at, so I always display the modified channel and I added the bottom
bar for the select by channel. I do it for all the combinations so it is
consistent, otherwise it will be even more confusing.

Let's give it a try to this, if in the end we decide that the previous was
better it can always be changed back, this is only GUI.

edgardoh:
Its fixed now, thanks.

TurboGit:
In original implementation it was easier to select a dedicated color (with the sliders) and the area-of-control based on the size of the circle for moving points. Do you plan to add the same level of support in this implementation?

edgardoh:
> In original implementation it was easier to select a dedicated color (with
> the sliders) and the area-of-control based on the size of the circle for
> moving points. Do you plan to add the same level of support in this
> implementation?
>
Not really, but I did this because I don't find the original implementation
convenient at all, this is much easy to me, if you want two nodes around a
hue you just have to cntrl + click to add them, and for moving you have the
mouse and keyboard + cntrl and shift to select the step. Maybe is just me?

edgardoh:
I just added the old edit mode back, there's a check box at the bottom.

There's also a combo to choose between the old process and the new one: smooth=old process, strong=new process.

@rawfiner , I've been playing with the background and the gradients hue-lightness  / hue-saturation don't look good on this resolution.
With the knots close to the bottom bar is now easier to see the select by hue. You can also use the color picker by range and the second color picker to create a curve, so there are a number of options to play with.
If you still like better the old background I don't mind adding it back, but it will have to be the big squares with lines and I don't know how to handle the GUI, the only option I can think of is to have another check box to switch between backgrounds.

rawfiner:
@edgardoh I get again the same issue with dragging out of the zone when I get out of the zone by dragging toward the left or toward the right (top and bottom are ok)

Concerning the background:
- when color picker is activated, I would prefer not to have the histogram shown, so that I can easily see what "output" I want and directly drag the curve at the point I want. The histogram makes this harder to do, I think. Maybe an option to display it/hide it would be nice?
- I still prefer the "idea" of the old backgrounds when color picker is not activated for lightness and saturation tabs, but if possible without the big squares with lines: the "hue" tab background you made is perfect for use with select by hue for instance. Note that I do not know at all how these backgrounds are implemented in the code and if that is something possible to do without too much effort, don't hesitate to tell me if I ask for impossible things.

Also, this is only my humble opinion, I do not know what other users would think :-)

TurboGit:
I also prefer the old backgrounds.

edgardoh:
> @edgardoh <https://github.com/edgardoh> I get again the same issue with
> dragging out of the zone when I get out of the zone by dragging toward the
> left or toward the right (top and bottom are ok)
>
I'll check this, if you are talking about the left and right nodes, they
are tricky, any other on the middle should be removed if dragged to the
left.

> Concerning the background:
>
>    - when color picker is activated, I would prefer not to have the
>    histogram shown, so that I can easily see what "output" I want and directly
>    drag the curve at the point I want. The histogram makes this harder to do,
>    I think. Maybe an option to display it/hide it would be nice?
>
> I can hide it when the color picker is active, you're right, is not very
useful then.

>
>    - I still prefer the "idea" of the old backgrounds when color picker
>    is not activated for lightness and saturation tabs, but if possible without
>    the big squares with lines: the "hue" tab background you made is perfect
>    for use with select by hue for instance. Note that I do not know at all how
>    these backgrounds are implemented in the code and if that is something
>    possible to do without too much effort, don't hesitate to tell me if I ask
>    for impossible things.
>
> Is not about the effort, I don't mind doing it, but it does not look nice
with the current resolution (the size of the squares) I'll work on a
prototype so you can see it for yourself.

edgardoh:
> I also prefer the old backgrounds.
>
Well, that settle it, I'll work on a prototype and we'll take it from
there.

edgardoh:
There's a new option now to switch between backgrounds, let me know what you think.

TurboGit:
I like this last version. Yet, I would move the GUI options "edit by areas" and "draw background mixed" as first options just below the graph.

rawfiner:
I like the last version too. :-)
Thanks! :-)
Just a remark: shouldn't the top right part of the lightness vs lightness background be completely white?

edgardoh:
Just pushed another version

@TurboGit , something like this?

@rawfiner , not sure how to do that, the issue is here:
https://github.com/darktable-org/darktable/pull/2167/files#diff-43d0368000c0c0077befff968dec18c5R937
but it get clipper no matter what I do...

TurboGit:
Yes, better. Thanks.
When "draw by area" is selected, leaving the graph should clear the area display on the graph.

edgardoh:
@TurboGit , right, is fixed now.

@rawfiner , this last version is darker but it doesn't clip, check if you like it better.

rawfiner:
Thank you @edgardoh
I won't be able to check today, but I will check asap :-)

TurboGit:
Looks all good to me. One more question, it seems that the "mix" slider and the "process mode" (strong smooth) are quite related. Maybe have them next to each other? ("mix" just below "process mode"). What do you think?

edgardoh:
> Looks all good to me. One more question, it seems that the "mix" slider
> and the "process mode" (strong smooth) are quite related. Maybe have them
> next to each other? ("mix" just below "process mode"). What do you think?
>
I agree.
I'm not sure if I like the 'select by' at the bottom, maybe it should be
the first after the graph? On the other hand, probably is not something
that will be changed a lot.

TurboGit:
Maybe or at least just above the "process mode" ?

edgardoh:
Done.

I'm thinking about removing the other background, seems that I'm the only one that like it and we'll save one control on the UI and some code, partially duplicated, that will make more difficult to maintain it, what do you think?

TurboGit:
Yes I would remove the other backgrounds.

edgardoh:
@TurboGit , should be ready now.

TurboGit:
No more comments. Looks ok. Thanks again for the energy you put on all those PR!

PaoloAst:
Awesome! 

Just one thing. When I move the eyedropper around the picture, I don't see the vertical line updated on the iop.

edgardoh:
You need to release the mouse button, is not updating while dragging.

rawfiner:
@edgardoh the lightness vs lightness background is very nice now, thanks :)

rawfiner:
Also, could you give an exemple of the way you use the "curve create" color picker and for what purpose please?

I like a lot this new interface, thanks for your work!

edgardoh:
Very often I work with a particular color, to change the hue and/or
saturation, and I find difficult to create by hand a small range. With this
new picker I select the color I want to work with and it creates a curve
with 5 point, 3 for the max, min and average and 2 for the feather. With
only this I can see the effect on the image and if not what I expect I keep
playing with the picker until I get something close enough, then I tweak it
by hand, usually I delete some of the middle nodes, but that depends on the
image.

Still, this is not perfect, for small areas is not easy to get the right
curve. I would like to have a zoom feature, but I'm still deciding how to
handle the GUI.

rawfiner:
> Very often I work with a particular color, to change the hue and/or saturation, and I find difficult to create by hand a small range. With this new picker I select the color I want to work with and it creates a curve with 5 point, 3 for the max, min and average and 2 for the feather. With only this I can see the effect on the image and if not what I expect I keep playing with the picker until I get something close enough, then I tweak it by hand, usually I delete some of the middle nodes, but that depends on the image. Still, this is not perfect, for small areas is not easy to get the right curve. I would like to have a zoom feature, but I'm still deciding how to handle the GUI. 

Ok, thanks for your answer. I tested it a bit more, and liked it.
Would it be possible to have a mode were the nodes are added on the x-axis but were the curve remains flat? It is not always easy to edit the curve fastly when the nodes are already at different y-positions, while when they are on a flat curve, we can easily and fastly move them were we want using the "edit by area" feature.

Apart of that, a little GUI issue:
when we drag a node towards the right while there is another node at the right, the node we drag is deleted when we go too much to the right. I would expect the node to stop going to the right, but not to be deleted. (there is this same behavior when we drag the node toward the left, I just choose the right for explaining)
Example in the image bellow, pushing the node indicated by the arrow towards the right will make it disapear
colorzone_node

edgardoh:
The behavior is the same as in the tone curve (minus bugs, of course), so
you can add nodes by cntrl+click, is that what you expect?

The nodes are deleted when node1.x > node2.x, so no matter if you move it
to the right or left, if nodes overlap one is deleted. You can try the same
on the tone curve.

I could make this to work different than the tone curve, but if I stop the
node from being moved I'm not sure how to delete it.

rawfiner:
Ok, I did not notice that it was the case with tone curve, and indeed it is the case! 
Let's keep the behaviors consistent with the tone curve then.
Thanks for the quick answer :-)

Regarding the first part of my comment, to clarify, I was referring to the "curve create" tool. My question was if it would be possible that this picker, instead of creating a curve, only adds nodes on a flat curve at the positions of max, min, average, and feather points?

edgardoh:
I see, plain click create a positive curve and cntrl+click a negative one.
I like that. Will it work for you if shift+click creates flat one? Or maybe
plain click creates a flat and with modifiers either positive or negative?

rawfiner:
> I see, plain click create a positive curve and cntrl+click a negative one. I like that. Will it work for you if shift+click creates flat one? Or maybe plain click creates a flat and with modifiers either positive or negative?

Yes, that would be awesome :-)

junkyardsparkle:
My two cents: This interface is different enough from the old one that it might be upsetting to some users... making it a new module instead would free you from constraints and allow for improving it in whatever other ways might occur to you. :)

That said, the only problem I'm having with it so far is that right-clicking to add a spline immediatly adds it at the y=0 position, regardless of where I clicked. This could be an issue with my somewhat old GTK3 version (3.18.9) if nobody else is seeing it.

edgardoh:
> My two cents: This interface is different enough from the old one that it
> might be upsetting to some users... making it a new module instead would
> free you from constraints and allow for improving it in whatever other ways
> might occur to you. :)
>
> That said, the only problem I'm having with it so far is that
> right-clicking to add a spline immediatly adds it at the y=0 position,
> regardless of where I clicked. This could be an issue with my somewhat old
> GTK3 version (3.18.9) if nobody else is seeing it.
>
Cntrl+click add a node to the existing y, not to y=0.

junkyardsparkle:
> Cntrl+click add a node to the existing y, not to y=0.
So it's the intended behavior? It seems "unexpected" compared to any other curve tool I've ever used.

edgardoh:
Have you ever user the tone curve?

junkyardsparkle:
> Have you ever user the tone curve?

The one in darktable, yes, and many other ones over the last 20 years. None of them exhibit this behavior.

edgardoh:
dt tone curve work like this. Actually, this works like the tone curve.

junkyardsparkle:
Ok, I have more time now, maybe I can explain better now what's happening. If darktable's tone curve module, if you left-click ~on the existing curve, then move the mouse even slightly, a node is created at the position of the mouse. This is the expected behavior. In this module, if I do the same thing, I get this:
screenshot-1552610640
The node is created at the top of the grid, instead of at the mouse pointer in the center. Maybe y=0 wasn't the best way to describe this, but hopefully the screenshot makes it clear.

edgardoh:
It still works the same as the tone curve for me, the node is created where
the mouse is, if you place it over the curve you'll get a node at y=0, if
you place it on the top, the node is created on the top.

junkyardsparkle:
Ok, if nobody else reports it, then it's my problem, :) It's odd that it doesn't happen with any of the other modules with similar interfaces, though, and doesn't happen with ctrl+click in this one. I'll look into it further when I get a chance.

junkyardsparkle:
Now that I've used this for some time, there's one thing I miss from the old version: the ability to click-and-drag the picker over a small area of the image while watching the result continuously update in the module. This was a convenient way to see the range within an area, which now requires much more clicking. I don't know if this is very important to most users, but some other people might use it.

edgardoh:
This was done for performance reasons, but mostly for the area mode, in the point mode maybe is not that bad and the user can always choose to click instead of drag if he/she don't want the re-process, so I think I can add the feature back in point mode.

@TurboGit , what do you think?

TurboGit:
@edgardoh : I would say yes, in point mode it could be nice to have the display update.

edgardoh:
OK, I'll work on it.

Benja1972:
Have some issue with color zone new module. When I click on 'edit by area' It produce only two point on bottom scale, origin (left) and  end point (right)

PR #2168, opened 2019-03-07
new module basic adjustments

edgardoh:
This new module allows to set the exposure, highlight recovery, contrast, brightness and saturation.
It has an auto exposure option that works on the entire image or a user defined area.

Is an alternative to the base curve, usually I get nicer results for a edit starting point.

The auto option and highlight recovery are based on RawTherapee auto levels & highlight recovery.


TurboGit:
Sounds good! I'll test this one.

At some point I feel that we should have a "beginner" module group and this IOP should be part of this group.

TurboGit:
Maybe use a % for sliders contrast, brightness and saturation.

For exposure 4 digits are too much, 2 will be certainly enough (as in exposure module). This will make this simple module even more beginner friendly.

If you agree and don't have time I can handle this?

edgardoh:
I'm ok with 2 digits for exposure, but I'm not sure about the %, the
current contrast, brightness saturation don't use a %, shouldn't this be
the same?

edgardoh:
I just moved all to 2 digits, seems nice this way.

aurelienpierre:
Sorry, I don't like that.

Exposure adjustments should come before the input color profile, but the color adjustments should come after. So you can't bundle them together in the same IOP.

Then, the lightness adjustment is a simple gamma on decorrelated RGB channels, which is bad especially this early in the pipe.

Also, you are duplicating the fulcrum contrast and saturation from the color balance. I planned on adding a vibrance and chroma adjustments, maybe we could keep it the color hub of dt, as Capture One or Da Vinci Resolve do ?

On another note, when you write:

```C 
  // exposure
  pixel.x = (pixel.x - black_point) * scale;
  pixel.y = (pixel.y - black_point) * scale;
  pixel.z = (pixel.z - black_point) * scale;
```

you should write full vectors operations, like:

```C
pixel = (pixel - black_point) * scale;
```

because OpenCL is designed to be vectorized (and I don't know if the compiler is clever enough to detect you are working on vector coordinates and pack them).

edgardoh:
@TurboGit , this has the wip label but I'm not sure what's expected to change here.

TurboGit:
Just that I did not visit this PR since some time. Not wip, so I'll remove the flag. BTW, this needs a conflict resolution.

TurboGit:
@edgardoh : also Aurélien's comment about pixel vectorization does not seem to have been taken into account. Any issue with that?

edgardoh:
Thanks, I did not mean to rush you, just an FYI.

I'll upgrade this.

edgardoh:
@TurboGit , all fixed.

src/iop/basicadj.c
```diff
@@ -0,0 +1,1744 @@
...
+    // gavg += histogram[ii] * _get_LUTf(gamma2curve, gamma2curve_size, val);
+    gavg += histogram[ii] * gamma2(val);
+    val += increment;
+  }
+
+  gavg /= sum;
+
+  if(black < gavg)
+  {
+    int maxwhiteclip = (gavg - black) * 4 / 3
```
TurboGit: const

src/iop/basicadj.c
```diff
@@ -0,0 +1,1744 @@
...
+  // compute contrast that spreads the average spacing of octiles
+  contr = 50.0f * (1.1f - ospread);
+  contr = MAX(0.f, MIN(100.f, contr));
+  // take gamma into account
+  double whiteclipg = gamma2(whiteclip * corr);
+
+  float gavg = 0.f;
+
+  float val = 0.f;
+  float increment = corr * (1 << histcompr);
```
TurboGit: const

src/iop/basicadj.c
```diff
@@ -0,0 +1,1744 @@
...
+
+#include "bauhaus/bauhaus.h"
+#include "develop/imageop.h"
+#include "gui/color_picker_proxy.h"
+
+DT_MODULE_INTROSPECTION(1, dt_iop_basicadj_params_t)
+
+#define exposure2white(x) exp2f(-(x))
+
+typedef enum dt_iop_basicadj_preservecolors_t {
```
TurboGit: { should be on new line

src/iop/basicadj.c
```diff
@@ -0,0 +1,1744 @@
...
+    double f;
+    int64_t i;
+  } tmp;
+  tmp.i = i;
+  return tmp.f;
+}
+
+static inline int ilogbp1(double d)
+{
+  int m = d < 4.9090934652977266E-91;
```
TurboGit: const

src/iop/basicadj.c
```diff
@@ -0,0 +1,1744 @@
...
+  u = longBitsToDouble(((int64_t)(q + 0x3ff)) << 52);
+  return x * u;
+}
+
+static inline double xlog(double d)
+{
+  double x, x2, t, m;
+  int e;
+
+  e = ilogbp1(d * 0.7071);
```
TurboGit: const int e = ...

src/iop/basicadj.c
```diff
@@ -0,0 +1,1744 @@
...
+  m = ldexpk(d, -e);
```
TurboGit: const double m = ...

src/iop/basicadj.c
```diff
@@ -0,0 +1,1744 @@
...
+  float expcomp = 0.f;
+  float black = 0.f;
+  float bright = 0.f;
+  float contr = 0.f;
+  float hlcompr = 0.f;
+  float hlcomprthresh = 0.f;
+
+  float scale = 65536.0f;
+
+  int imax = 65536 >> histcompr;
```
TurboGit: const

src/iop/basicadj.c
```diff
@@ -0,0 +1,1744 @@
...
+  rawmax <<= histcompr;
+  whiteclip <<= histcompr;
+  ave = ave * (1 << histcompr);
+  median <<= histcompr;
+  shc <<= histcompr;
+
+  // compute exposure compensation as geometric mean of the amount that
+  // sets the mean or median at middle gray, and the amount that sets the estimated top
+  // of the histogram at or near clipping.
+  float expcomp1 = (log(midgray * scale / (ave - shc + midgray * shc))) / log(2.f);
```
TurboGit: const ?

src/iop/basicadj.c
edgardoh: We still have this clang issue, even if I change it, clang will change it back.

edgardoh:
Review changes ready.

TurboGit:
Thanks! Looks good merged.

PaoloAst:
Great job! I don't see any effect of "middle grey" slider...

edgardoh:
Thanks.

The middle grey works with the contrast slider, it defines the lightness
that is not modified by the contrast. The default is the middle grey for
the work rgb colorspace. If the work profile changes, the default will
change too, but you'll need to reset it by hand.

PaoloAst:
Thanks @edgardoh !

GLLM:
> This new module allows to set the exposure, highlight recovery, contrast, brightness and saturation.
> It has an auto exposure option that works on the entire image or a user defined area.
> 
> Is an alternative to the base curve, usually I get nicer results for a edit starting point.
> 
> The auto option and highlight recovery are based on RawTherapee auto levels & highlight recovery.

Hello @edgardoh 
Can I ask (couldnt find it in the doc) what is the difference between the Exposure and the Brightness effects ?
Moreover, the Middle grey slider does not do anything ? What is it meant for ? What can I possibly do wrong ?
Thanks a lot.
GLLM

PR #1904, opened 2018-12-16
tone equalizer

aurelienpierre:
introduce a new "tone equalizer " module, intended to allow selective exposure adjustements in linear RGB (supposed to be color-safe). This is very similar to Lightroom, with blacks/shadows/midtones/highlights/whites cursors.

The luminance channels are blended with gaussian masks and the modifications are applied on a gaussian pyramid to preserve the local contrast (luminance modified in low frequency layers).

For now, nothing is optimized (naive C implementation) and the module is expected to be slow. This is just a proof of concept. I will make it faster once the core algorithm is validated.

PR #2169, opened 2019-03-08
new module rgb curve

edgardoh:
A new RGB curve module with linked and independent channels.


TurboGit:
We really want to have a clear idea of what to do with this and #2017. Those seems duplicate, right?

edgardoh:
I can't speak about #2017, but I've been working for a while on this, I
like it, so I decided to share it.

Having a new module or merge it into the current one is part of a larger
discussion. One of my goals while doing what I'm doing is to have an RGB
workflow, for this some Lab modules need to be ported to RGB and we have
the same question for all of them, add the RGB option or create a new one.

I think is better to create a new one, it make the code simpler an easier
to maintain, and a native RGB module use RGB for the blend, while a Lab use
Lab.

TurboGit:
I see! I think @aurelienpierre had said something along these lines. I'd like to have his point of view on this.

Aurélien?

aurelienpierre:
As much as I like a RGB workflow, I think this duplicates the efforts of @phweyland on #2017. But the RGB blending is strong argument, actually blending in non-linear spaces is doomed from the basic theory of it. Is there now way to get an RGB blending for those Lab modules ?

edgardoh:
No, there's not.

aurelienpierre:
So we will need to be clever on the UI side to avoid confusion then.

phweyland:
If I can put my cent on this...

I've understood that the new mode brought the possibility to choose the working colorspace inside each module. But to be complete that should including the blending mode. If not, the move is not finished yet, is it ? Otherwise we will have to duplicate all the modules ...

The multiplication of modules doing very similar things is a bit messy (some people want to turn dt yet more professional, right ?). It would be really better to make the existing modules work the right way (including tonecurve).

Now, really, not having to convert lab to rgb  and having blending in the same colorspace sounds really good.

TurboGit:
I understand arguments on both side I think. Yet, we need to remember that in dt it is not because we have a module that we need to use it even see it. We can just disable it.

So the point I see that we need a clear idea about is:
- do we want to add more code into each module to have an RGB mode
- do we prefer a simple approach (and simpler UI)) and have new modules dedicated to RGB mode
?

Personally I have no strong opinion. I think I could live with both.

MStraeten:
OSX:
there seems to be a segfault on rgb curve being activated with a former processed raw.
scenario: history stack contains: rotate, crop&rotate, perspective correction, demosaic (amaze)
activating rgb --> segfault: 
[console.txt](https://github.com/darktable-org/darktable/files/2948356/console.txt)
trace: [trace2.txt](https://github.com/darktable-org/darktable/files/2948357/trace2.txt)

doing the same starting from scratch, first applying perspective correction, crop&rotate, demosaic, and then activing rgb works fine

edgardoh:
Can you share the xmp?

MStraeten:
with history stack: [20190223-IMG_5936.CR2.xmp.txt](https://github.com/darktable-org/darktable/files/2948429/20190223-IMG_5936.CR2.xmp.txt)
from scratch: [20190223-IMG_5937.CR2.xmp.txt](https://github.com/darktable-org/darktable/files/2948430/20190223-IMG_5937.CR2.xmp.txt)

edgardoh:
@MStraeten , should be fixed now, thanks for reportig it. 

The module must be between input color profile and output color profile, you'll have to move it by hand, until you do so you'll get a random image and an error on the console.

phweyland:
> Personally I have no strong opinion. I think I could live with both.

I think that may be a strategic choice... 
But whatever it is, I'm happy to contribute where possible.

edgardoh:
I already say where I stand, but to make my case consider the levels. We all agree that it should work on RGB, if we add it to the current module we'll have a Lab mode that we'll strongly suggest not to use, a blend mode in Lab and the code will be more difficult to maintain.

With the tone curve we'll have a RGB mode hard-coded in prophotoRGB, a "work profile RGB" mode, in addition of all existent modes, and the blend will also work in Lab (not to mention the code)

Regarding the LCh I have no strong opinion, but LCh is close to Lab, so at least it should be considered to add it to the tone curve (is already there anyway)

TurboGit:
New modules are migrating toward an RGB working mode. Fine. But I'm not certain to understand the relationship with the UI. We can have a curve in Lab and RGB, both actually tweaking the module in RGB mode with proper color conversion, right? Or are the color conversion the problem? Not conservative enough? For example the colorbalance can have RGB or HSL sliders and both are actually tweaking the same circuitry in the module with proper conversions.

edgardoh:
> New modules are migrating toward an RGB working mode. Fine. But I'm not
> certain to understand the relationship with the UI. We can have a curve in
> Lab and RGB, both actually tweaking the module in RGB mode with proper
> color conversion, right? Or are the color conversion the problem? Not
> conservative enough? For example the colorbalance can have RGB or HSL
> sliders and both are actually tweaking the same circuitry in the module
> with proper conversions.
>
Not sure if I get this, I don't have a problem with the UI, and this
implementation display the curve in RGB or Lab (with the 'compensate middle
grey' checkbox). My problem is with the functionality (offering more
options than needed) and the code (that gets more complex and difficult to
maintain)

phweyland:
> that gets more complex and difficult to maintain

I share the concern of complexity, but I don't think it comes with the different options we could offer.
The actual complexity is to use a Lab design (module + environment) to work in rgb. In that design working on rgb doesn't come for free, we have to build everything:
- color space, double conversion
- color picker
- histogram capture
- without missing blending modes.
If a module can really choose the color space (with all items above aligned) the complexity has gone.
If we say that the new design enables the choice of color space it should include the above items. That would avoid unnecessary complexity. Otherwise, the possibility of choosing the color space is misleading. That's just a new color space segment (rgb) introduced into the pixelpipe. 

That's where we come to the strategic choices.
Either a new family of modules, working on rgb, are going to live beside (and could progressively substitute) the lab modules (In that case we cannot advertise the module can choose its color space).
Or,  we let the module choose the best color space to make its operation. But the full environment must be available. Then, with few extra work it becomes possible to enhance the existing modules which would work better on rgb.

From my standpoint, it is strategic to define clearly the targeted environment, whatever it is, where color picker, histogram, modules order are meaningful and reliable for the different usages which are offered. 
The Lab design may have some drawbacks, but a lack of design is a disaster.

phweyland:
@edgardoh 
to learn how to use the new pixelpipe I've installed your rgbcurve branch.
There is the option called `compensate middle grey`.
Have you already explained somewhere what are the principles of this option ? How does that work (it seems that it involves colorprofile, trc, lab) ? If not could you present it a little bit ?
Thanks

edgardoh:
It scale the curve and histogram so the middle grey is at 50%. In order to
do that it transform things to Lab, and for that, since the input can be
any RGB colorspace it uses the work profile.

If you switch it with an S curve and the work profile on something linear
you'll see how the curve changes, but the image remains the same.

phweyland:
> since the input can be any RGB colorspace it uses the work profile.

Just seen that we have to choose the work color profile in the module input color profile... :-)
I guess we could add some log profile(s) to that list ...

aurelienpierre:
> New modules are migrating toward an RGB working mode. Fine. But I'm not certain to understand the relationship with the UI. We can have a curve in Lab and RGB, both actually tweaking the module in RGB mode with proper color conversion, right? Or are the color conversion the problem? Not conservative enough? For example the colorbalance can have RGB or HSL sliders and both are actually tweaking the same circuitry in the module with proper conversions.

Having the module work, internally, in Lab or RGB doesn't change the fact that the blending options still depend on the in/out space of the module, hence on its position in the pipe. The problem of blending is it uses alpha blending and possibly blur, and blurring colours in a non-linear spaces (Lab or gamma-encoded RGB) results in uneven transitions (see : https://blog.johnnovak.net/2016/09/21/what-every-coder-should-know-about-gamma/#colour-blending). The whole concept of alpha blending is foolish in non-linear scenery (https://blog.johnnovak.net/2016/09/21/what-every-coder-should-know-about-gamma/#alpha-blending--compositing).

That's why I think even the Lab modules should get a linear RGB blending option. And if this becomes reality, there is no reason to duplicate the modules with Lab (with possible internal RGB conversions) and RGB flavours. But if it's too complicated, well fine with the duplication, although the argument of "you don't have to see every module" sounds like hiding dust under the carpet to me.

As for the code complexity, I can say there is a lot of code duplication in dt, first between modules (blur libs, graphs drawing, etc.), then inside the modules themselves (C + SSE paths). I have tried to factorize code and build standard libs (for example, vector algebra and maths functions), but there is a lot of work to do. 

For example, I think SSE codepaths are useless if proper C codepaths (with OpenMP SIMD) are written properly from the beginning (but they are not), so maybe we could start with progressively getting rid of the pure SSE pathes, refactor, clean and get auto-vectorizable code by default (see what I have done in #1904, I get ×6 speed-up with OpenMP SIMD compared to original naive C, with no SSE intrinsics whatsoever).

In design patterns, best practices involve writing small functions, doing only one thing at the time, with an average of 2.5 `if` branching by function. Such functions are easier to test and prepare with `#pragma openmp declare simd` for vectorization. They make the code more readable and re-usable. We are far from such practices in dt, with functions of hundreds of lines and duplicated stuff. 

TL;DR before concluding on the complexity of making old modules speak Lab and RGB at the same time, I would take this opportunity to split current modules into elementary functions, write stupid simple C pathes with OpenMP pragmas, get OpenMP SIMD to work on-par with the SSE2 pathes, then get rid of the SSE2 pathes, then share code between modules, then add the translating options LAB <-> RGB.

---

**Digression** : Also, the choice of limitating SSE features to SSE2 (2004) does not make sense anymore, because newer versions have other speed-ups like 256/512 bits vectors, hardware-accelerated fused multiply-add (FMA) etc., and limiting us to the SSE2 subset forbid us to use them for compatibility reasons, while using OpenMP and compiler auto-vectorization (with code written to be vectorizable) would allow the compiler to use all the fancier options available and adapt to whatever platform it's compiled on.

aurelienpierre:
Also, while we are here, the tone curve is using a pre-computed look-up table with various spline interpolations between the nodes. That LUT is built between [0;1] and exponential extrapolations happen out of that range. Hence branching.

That was optimal with single-threaded scalar code, but my tests show that, with SSE2 (and OpenCL is a kind of highly paralellized SSE2), all the memory accesses it creates (`LUT[CLAMP((uint16)x * 65536), 0, 65535)]`) are often slower than asking the CPU/GPU to compile the exact interpolation in floating point for each pixel. So maybe it's worth it to get rid of the LUT now.

edgardoh:
@TurboGit , I'll resume working on this now, let me know if there's anything you need (other than time) to make a desition about it.

TurboGit:
@aurelienpierre : I agree that limiting ourselves to SSE2 is no good. I'm not a hardware guy, what will be good is to know since when SSE3 is well supported in all CPU. Same for SSE4.

TurboGit:
Here is my new thinking. Nothing committed and to be discussed with all knowledgeable people on image algorithm.

I'm leaning toward having this module in and separated from the current tone curve. Have new module working in rgb mode seems to be the way to go. Then we will be able to have "module list" preset for a full rgb oriented pixel-pipe using those new modules and exclusively those that are better at avoiding hue deviations if I understand correctly.

I think you'll agree and this is also Aurélien point of view. Right?

Again this discussion is still opened of course, but that's my current decision on this.

edgardoh:
Great! I'm still working on it, I'll let you know when is ready for review.

edgardoh:
Ready for test & review.

TurboGit:
Just tested it and got:

> [rgbcurve process] rgb curve must be between input color profile and output color profile

And all the colors are psychedelic :)

edgardoh:
Yes, same as with the basic adjustments. I'll work on that after this is
promoted.

TurboGit:
> I'll work on that after this is promoted.

What do you mean by promoted here?

edgardoh:
Sorry, I should explain better.

By promoted I mean merged.

Both this modules work with the work profile, and before the colorin
there's no work profile available, so some features will not work. The idea
behind this is to have them after the colorin, but because of backward
compatibility that's not something we can enforce, so I'll have to do
something about it.
I want to have both modules available before working on this because they
have different scenarios, this way I can work on something that include all
of them.
Until then, if the module is before colorin, is not even processed and the
error message that you saw is displayed, so it should be moved by hand
after the colorin.

On new edits it should be after the colorin by default, so I assume that
you have tested it on an old edit.

TurboGit:
> By promoted I mean merged.

But I cannot merge something that I cannot test :) This is a chicken and eggs problem! Or can you give me a quick patch to make it working? Or a PR? What do you suggest?

edgardoh:
If you move the module after the colorin or start a new edit it should work
fine. If not it means that there's a bug, so let me know and I'll fix it.

edgardoh:
I have moved it after the levels, so even if the user enables it on an old edit things should work fine.

TurboGit:
Ok, thanks I missed in your previous message that it should be ok on new edits!

Just a minor issue. When I used the color picker to define an area I get the area on the curve. Good. I have points on the curve on each side of the area. Fine. But I have two points (very close to each others) defined in the middle. I would have expected a single one to be able to smoothly move the curve inside the area up or down. As it is it is not possible to edit the curve, when you move one of the inner point you create a cusp.

TurboGit:
Also it seems the it is not possible to create a positive or negative curve. It is always flat for me.

edgardoh:
This is up for discussion, I add one in the middle and one in the average
(the color picker show the average), so the user can choose one (and delete
the other).

edgardoh:
The option is not available, but if you find it useful I can add it. We
should define how it should work, in the color zones we have two more nodes
that act like feather, here we haven't, but I can add them.

TurboGit:
I see for the middle vs average point. I suppose that with a good documentation (and in the hint message) it is fine.

For the positive/negative curve, yes please add support for that to match the others implementation of this kind of color picker. I find this useful.

edgardoh:
How this should work? I usually define two types of curves, one for
contrast and another to brighten / darken the image. The contrast curve may
require more nodes (we can have the feather or not, not sure), the brighten
/ darken should only move the middle / average node.

TurboGit:
I'll go for the simple brighten curve for now.

edgardoh:
OK

edgardoh:
Ready the new curves.

TurboGit:
Ok, looks ok. Thanks again for the hard work!

TurboGit:
Forgot about a point I wanted to raise. This module is in the CORRECT group by default. Maybe better in the tonality one?

edgardoh:
I have this in the code:

int default_group()
{
  return IOP_GROUP_TONE;
}

TurboGit:
My bad! I have a non default layout and the iop-layout scripts were not updated. Done now for the new rgb curve and the basic adjustment iops.

PR #2174, opened 2019-03-08
lighttable undo

TurboGit:
This is the last undo support. After we'll have undo for almost all important items.


TurboGit:
Now support styles too.

Issue #2041, opened 2019-01-29
make the histogram read-only


aurelienpierre:
I'm working on a clean-up of the UI, to make it more minimalistic in order to keep the focus on the picture. That is, less distractions in the panels. See #2037 

Currently, the histogram is an active area where one can drag the shadows and the highlights to quickely edit the exposure and black level. The thing is, the setting increments are rather coarse, not configurable (except in darktablerc, when you know it), and I discovered this feature by accident.

I would like to make the histogram read-only, meaning removing the active parts. 

I also would like to remove the EXIF settings display, because it's redundant with the EXIF box on the left panel.

Then, I would like to make the histogram slightly smaller in height, to leave more space to modules.

Any objection ?

chrik5:
Hello @aurelienpierre,

Ordinary user here with no programming skills, which really appreciates the time you spend for dt.

> Currently, the histogram is an active area where one can drag the shadows and the highlights to quickely edit the exposure and black level. The thing is, the setting increments are rather coarse, not configurable (except in darktablerc, when you know it), and I discovered this feature by accident.

I use this quite frequently as it is much faster than searching for the exposure module. Would be nice to have more of this easy access features in dt.

> I also would like to remove the EXIF settings display, because it's redundant with the EXIF box on the left panel.

Left panel is most of the time collapsed in my case.

> Any objection ?

Asking here won't give you much feedback from users. 
I remember a few basic tutorials which mention this feature as nice addition. 

>(except in darktablerc, when you know it)

Well that rc's are linux style and I touch darktablerc very seldom. But since dt is avaliable for others OS an easier handling of settings in darktablerc might help user to find all this hidden features. Recently, more settings were added to the preferences dialog. Maybe it would make sense to expose most non critical darktablerc settings in some extended preferences/config dialog?

Thanks,
Christian

aurelienpierre:
Actually, my bad, the exposure steps are hard-coded and not configurable.

> Left panel is most of the time collapsed in my case.

The real question is: are the EXIF info useful in the histogram ? My feeling is I don't care about the settings at all during retouch.

> I use this quite frequently as it is much faster than searching for the exposure module. Would be nice to have more of this easy access features in dt.

And using it in a blind fashion does not bother you ? If so, did you consider setting a keyboard shortcut
? I'm asking because, everytime I'm "using" the histogram to set the exposure, it's a touchpad accident while reaching for the color channels or waveform. So, it annoys me more than anything.

[...]

PR #2181, opened 2019-03-09
new "quick" image_infos lib

AlicVB:
To display quickly important infos about the current image in darkroom.
Capture d’écran_2019-03-09_17-29-30
The goal is to have important information about the current image displayed without the need to expand the metadata view module and search in all the information.

This is a rework of an old try/idea resurrected by some comment here : https://github.com/darktable-org/darktable/issues/2041#issuecomment-459952932

The module can be positioned in 3 places (alterable in prefs) : 
- on top of the navigation
- on top of the histogram
- center on the top line (between filters combobox and the G etc icons
- center on the bottom line (between styles icons and clipping icons)

template (alterable in prefs) : 
- it use the same template system as used in the export lib
- I've extended the predefined variables used there, so this lib as well as export can now use : $(EXIF_EXPOSURE) $(EXIF_APERTURE) $(EXIF_FOCAL_LENGTH) $(EXIF_FOCUS_DISTANCE) $(LONGITUDE) $(LATITUDE) $(ELEVATION)
- Special chars which can be displayed (ex : ° ...) are only printed for the image info lib (not export jobs, to avoid problems)
- It support pango markup
- it support multilines by using the $(NL) variable

Also it would be good to have an option to set place to "none", but I've to find a nice way to not load libs (for now, you can set an empty template and place bottom or top center)


TurboGit:
Sounds good and so we can certainly remove the image information from the histogram as discussed in #2037.

TurboGit:
> it support multilines by using the $(NL) variable

This seems equivalent to entering directly \n, right?

src/libs/tools/image_infos.c
```diff
@@ -0,0 +1,152 @@
+/*
+    This file is part of darktable,
+    copyright (c) 2011 Henrik Andersson.
```
TurboGit: Have you changed name? :)

AlicVB:
> > it support multilines by using the $(NL) variable
> 
> This seems equivalent to entering directly \n, right?

For some reason, \n doesn't work if entered "manually". It's removed at some point in the variable replacement process, and as I fear to break something for the export, I've preferred to do it like that.

TurboGit:
Ok, fine with me anyway. That was more a question to understand.

aurelienpierre:
> Also it would be good to have an option to set place to "none", but I've to find a nice way to not load libs (for now, you can set an empty template and place bottom or top center)

Just set it wherever you want and use `gtk_widget_set_visible(GTK_WIDGET(widget), FALSE)`

TurboGit:
Yep, agreed with Aurélien, you just have to hide it.

Nilvus:
This PR with #2037 will make the new darktable interface really awesome ! Great work.

AlicVB:
ok, finally, I've just set views to none if position is set to "hidden"

TurboGit:
Thanks ! Looks good.

PR #2187, opened 2019-03-10
fix blend rgb grey

edgardoh:
The blend module is calculating the rgb grey with:

0.3f * input[0] + 0.59f * input[1] + 0.11f * input[2]

This changes it to use the work profile when available.

### DIFF
--- a/data/kernels/blendop.cl
+++ b/data/kernels/blendop.cl
@@ -16,8 +16,8 @@
     along with darktable.  If not, see <http://www.gnu.org/licenses/>.
 */
 
-#include "common.h"
 #include "colorspace.cl"
+#include "color_conversion.cl"
 
 #define BLEND_ONLY_LIGHTNESS     8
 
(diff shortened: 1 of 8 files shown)


edgardoh:
I will need help with OpenCL here, according to the documentation is OK to pass NULL arguments, but inside opencl there is not a NULL or null, so I'm comparing to zero:

https://github.com/darktable-org/darktable/pull/2187/files#diff-826ecc2c60510d2905cd5ed0e104d4d4R232

I haven't found any reference to good practices regarding this scenario.

TurboGit:
Not found good practice or on my side either. Maybe just add an int (used as a boolean since an OpenCL true is -1) to indicate that the profile_info is set or not? Not sure it is better, a NULL pointer is 0 anyway.

edgardoh:
> Not found good practice or on my side either. Maybe just add an int (used
> as a boolean since an OpenCL true is -1) to indicate that the profile_info
> is set or not? Not sure it is better, a NULL pointer is 0 anyway.
>
A flag is not pretty but is safer. Let's wait for a while, maybe someone
knows something. If not I'll change it, better not pretty than risk having
a buid error on some systems, or worst, some random issues.

RELEASE_NOTES.md
```diff
@@ -42,6 +42,8 @@ when updating from the currently stable 2.4.x series, please bear in mind that y
 - Add new raster mask supports. This is a copy of a parametric mask
   which is stable during the whole pixel-pipe.
 
+- Modules can now be re-ordered with shift+drag.
```
TurboGit: It is not good to add RELASE_NOTES unrelated to current PR because if there is troubles and we revert it will also revert an needed information. And if we need to merge in a branch a feature and not the other this will require editing the commits before merging.

So, please in the future keep this in mind. Thanks.

edgardoh: I didn't thought about that, I'll move this to another PR.

edgardoh:
I just added the flag, if we find a better way to do this we can change it.

TurboGit:
Agreed, let's go with this then. Thanks.

Issue #2183, opened 2019-03-10
Overexposed module slows down the pipeline a lot


trougnouf:
The overexposed module slows down the pipeline significantly, making darktable very painful to use (I have to switch to the 1080p display because it's unusable on the 4K) and often making the preview fall back to CPU.

Here is a -d perf opencl output without overexpose: https://pastebin.com/raw/ntedGNSA (.471s)

With overexpose: https://pastebin.com/raw/Rj7WNKjQ (3.648s)

Version: commit 6188d46 (today) on Arch Linux
GPU: NVidia GTX 1070 (8GB) with opencl-nvidia 418.43-2

TurboGit:
Probably because it now properly report information given the export profile whereas before it was given some "information" not fully correct. Not sure if the speed can be improved.

edgardoh:
There are some improvements that can be made, I'll be working on that.

TurboGit:
This should be fixed now. Thanks Edgardoh.

trougnouf:
It works great, thank you @edgardoh !

PR #2191, opened 2019-03-11
color transform performance

edgardoh:
This uses internal routines for the color transformation when available for the overexposed and final histogram.
On my system is 30 times faster.

This addresses #2183

### DIFF
--- a/data/kernels/colorspaces.cl
+++ b/data/kernels/colorspaces.cl
@@ -58,3 +58,25 @@ colorspaces_transform_rgb_matrix_to_lab(read_only image2d_t in, write_only image
 
   write_imagef(out, (int2)(x, y), pixel);
 }
+
+kernel void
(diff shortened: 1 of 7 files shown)


TurboGit:
This is not working on my side. Either I have a full blue output (under exposed) or almost full red (over exposed) central view. I have tested with and without OpenCL. And tried with a CR2 and NEF. Really strange.

TurboGit:
In fact this is not an issue in this PR but it is already in master. To reproduce:
- activate overexpose
- select in display profile AND histogram profile the same profile (for whatever profile)

And in this case you have either full blue or full red output.

TurboGit:
Ok, and since this PR is ok I'll merge it. I have opened an issue for the overexpose:
https://github.com/darktable-org/darktable/issues/2192

PR #2195, opened 2019-03-12
add a iop search box

edgardoh:
This add a search text to the module groups. 

-It has a config option to show only the modules groups, only the search text or both.
-Modules are searched by name (localized).
-If displaying only the search text, the active pipe modules are displayed when the search text is empty.
-A darkroom shortcut cntrl+r set the focus on the search box

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -136,6 +136,12 @@ when updating from the currently stable 2.4.x series, please bear in mind that y
   Both the contrast and the auto feature use the middle grey point, the
   default for this setting is based on the work profile.
 
+- A search text has been added to the module groups in the darkroom. It has a
+  config option to show only the modules groups, only the search text or both.
+  Modules are searched by name (localized).
+  If displaying only the search text, the active pipe modules are displayed when the search text is empty.
+  A darkroom shortcut ctrl+a set the focus on the search box.
+
 ## Bug fixes
 
 - The color picker support has been fixed by a complete rewrite. It
--- a/src/libs/modulegroups.c
+++ b/src/libs/modulegroups.c
@@ -37,7 +37,16 @@ typedef struct dt_lib_modulegroups_t
 {
   uint32_t current;
   GtkWidget *buttons[DT_MODULEGROUP_SIZE];
+  GtkWidget *text_entry;
+  GtkWidget *hbox_buttons;
+  GtkWidget *hbox_search_box;
 } dt_lib_modulegroups_t;
+
+typedef enum dt_lib_modulegroup_iop_visibility_type_t {
+  DT_MODULEGROUP_SEARCH_IOP_TEXT_VISIBLE,
+  DT_MODULEGROUP_SEARCH_IOP_GROUPS_VISIBLE,
+  DT_MODULEGROUP_SEARCH_IOP_TEXT_GROUPS_VISIBLE
+} dt_lib_modulegroup_iop_visibility_type_t;
 
 /* toggle button callback */
 static void _lib_modulegroups_toggle(GtkWidget *button, gpointer data);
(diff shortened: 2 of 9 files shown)


edgardoh:
Some issues:

-I'm not sure the tags are searched by the translated text and I have no idea how to test it.
-The GUI needs more work, but I can't make it look like I want to.
-It should search by group name, but we don't have a group name right now, and it should be the same as the one used in the description.
-When the user find a module and the text looses the focus it still display the search result list of iop. Not sure what to do about it because if I clear the search the module will move making it inconvenient.

edgardoh:
This is ready now for test/review.

TurboGit:
Working fine but it takes some space and I'm not sure I will have a use for it. So I'd prefer having this widget controlled by a preference. You already have one to select the search context, I would add "none" to disable this widget if not needed.

TurboGit:
BTW, it works fine with the translated text.

edgardoh:
@TurboGit , it is already controled by preference, see my first post, you can show it alone, with the groups or hide it.

edgardoh:
> Working fine but it takes some space and I'm not sure I will have a use
> for it.
>
This probably depends on your workflow and editing style, I move the
modules a lot, the best for this is to be on the active pipe group, and
changing groups to add new modules is annoying. This makes it easier.

> So I'd prefer having this widget controlled by a preference. You already
> have one to select the search context, I would add "none" to disable this
> widget if not needed.
>

RELEASE_NOTES.md
```diff
@@ -129,6 +129,12 @@ when updating from the currently stable 2.4.x series, please bear in mind that y
   When pressing the shortcut 'pan&zoom while editing masks' the draw area
   can now be zoomed. Double-click on the bottom bar resets the zoom.
 
+- A search text has been added to the module groups in the darkroom. It has a
+  config option to show only the modules groups, only the search text or both.
+  Modules are searched by name (localized).
+  If displaying only the search text, the active pipe modules are displayed when the search text is empty.
+  A darkroom shortcut cntrl+r set the focus on the search box.
```
TurboGit: cntrl -> ctrl (for consistency this is was we currently use everywhere).

TurboGit: and you talk about ctrl+r ...

src/views/darkroom.c
```diff
@@ -2833,6 +2841,9 @@ void init_key_accels(dt_view_t *self)
 
   // add an option to allow skip mouse events while editing masks
   dt_accel_register_view(self, NC_("accel", "allow to pan & zoom while editing masks"), GDK_KEY_a, 0);
+
+  // set focus to the search modules text box
+  dt_accel_register_view(self, NC_("accel", "search modules"), GDK_KEY_a, GDK_CONTROL_MASK);
```
TurboGit: ... but here we have ctrl+a

src/libs/modulegroups.c
```diff
@@ -37,8 +37,17 @@ typedef struct dt_lib_modulegroups_t
 {
   uint32_t current;
   GtkWidget *buttons[DT_MODULEGROUP_SIZE];
+  GtkWidget *text_entry;
+  GtkWidget *hbox_buttons;
+  GtkWidget *hbox_search_box;
 } dt_lib_modulegroups_t;
 
+typedef enum dt_lib_modulegroup_iop_visibility_type_t {
```
TurboGit: sorry again this :)

{ should be at start of line

edgardoh: I don't know what to do about this, I use clang and is doing it like this. clang is the recommended way to go, so it should be modified to format it the way you like it. Otherwise I'll have to disable for this and enable for the rest.

TurboGit: You mean that clang compiler is changing that? Don't know I'm using emacs and gcc? But I do compile with clang to check PR and don't have this issue.

TurboGit:
Looks good to me. Thanks.

edgardoh:
clang-format is doing this, as defined in contributing.md / coding style

TurboGit:
I see, sorry don't run it then as dt seems to not follow the full clang style then.

TurboGit:
This will have to be sorted out.

junkyardsparkle:
I'm not sure how much I'll use this, but it seems like a nice feature... a few thoughts: It would be more useful if it could be bound to a shortcut, since using it will involve the keyboard anyway. If it shipped with a default shortcut of '/' or similar, it could be hidden unless activated, saving space. If the search included an implicit leading '\b' (word boundary) it might return more relevant results with fewer keystrokes.

PR #2205, opened 2019-03-14
add a zoom to the color zones

edgardoh:
The draw area of the color zones can now be zoomed when pressing the shortcut for 'pan&zoom while editing masks'.

A plain click on the second color picker creates a flat curve, with modifiers it creates either a positive or negative one.

Nodes can be deleted only with right-click.

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -121,7 +121,7 @@ when updating from the currently stable 2.4.x series, please bear in mind that y
   A new color picker has been added that creates a curve based on the area
   selected from the image.
   When pressing the shortcut 'pan&zoom while editing masks' the draw area
-  can now be zoomed.
+  can now be zoomed. Double-click on the bottom bar resets the zoom.
 
 ## Bug fixes
 
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -57,10 +57,6 @@ when updating from the currently stable 2.4.x series, please bear in mind that y
   area from the image. Click to adjust the input image slider,
   ctrl+click to adjust the out image one.
 
-- The color zones module now display an histogram based on the ‘select
-  by’ channel and displays the selected range if the color picker is
-  in area mode.
-
 - The picasa module has been renamed to googlephoto and completely
   rewrited to support the new Google Photo API. It is now again
   possible to create albums directly from the export module.
@@ -111,15 +107,21 @@ when updating from the currently stable 2.4.x series, please bear in mind that y
 
 - Add soft boundaries for denoise profile controls.
 
-- The color zones module curve acts now like the tone curve (but horizontal),
+- The color zones module now display an histogram based on the ‘select
+  by’ channel and displays the selected range if the color picker is
+  in area mode.
+  It acts now like the tone curve (but horizontal),
   it has two edit modes: edit by area is the former one, if not checked nodes
-  can be edited like in the tone curve.
+  can be edited like in the tone curve, delete only works by right-click
+  when not in edit area mode.
   A new process mode has been added: smooth is the former one, strong is new.
   An interpolation method has been added that allows to select different types
   of curves.
   The color picker allows to select by area with cntrl+click.
-  A new color picker has benn added that creates a curve based on the area
+  A new color picker has been added that creates a curve based on the area
   selected from the image.
+  When pressing the shortcut 'pan&zoom while editing masks' the draw area
+  can now be zoomed.
 
 ## Bug fixes
 
--- a/src/gui/draw.h
+++ b/src/gui/draw.h
@@ -275,6 +275,19 @@ static inline void dt_draw_histogram_8_linear(cairo_t *cr, uint32_t *hist, int32
   cairo_fill(cr);
 }
 
+static inline void dt_draw_histogram_8_zoomed(cairo_t *cr, uint32_t *hist, int32_t channels, int32_t channel,
+                                              const float zoom_factor, const float zoom_offset_x,
+                                              const float zoom_offset_y)
+{
+  cairo_move_to(cr, -zoom_offset_x, -zoom_offset_y);
+  for(int k = 0; k < 256; k++)
+    cairo_line_to(cr, ((float)k - zoom_offset_x) * zoom_factor,
+                  ((float)hist[channels * k + channel] - zoom_offset_y) * zoom_factor);
+  cairo_line_to(cr, (255.f - zoom_offset_x), -zoom_offset_y * zoom_factor);
+  cairo_close_path(cr);
+  cairo_fill(cr);
+}
+
 static inline void dt_draw_histogram_8_log(cairo_t *cr, uint32_t *hist, int32_t channels, int32_t channel)
 {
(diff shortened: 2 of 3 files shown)


edgardoh:
@rawfiner , you were right, deleting nodes while moving them is extremely annoying, so I have removed the feature and now nodes can only be deleted by right click while not in edit area mode.

rawfiner:
Thank you for all this amazing work @edgardoh :-)

TurboGit:
Looks nice but the zoom is inverted. A scroll up zoom-in for the center area whereas it zoom-out in colorzones. 

And BTW, instead that could the zoom be triggerred while over the little bar just below the curve? And a double click on this are would reset the zoom (which is not possible over the curve as a double click reset the curve). And doing that on the little bar won't need to use the "pan&zoom while editing mask".

TurboGit:
Forget the zoom in the "little bar", we need to keep the reference point (mouse pos) for the zoom. Yet, we may still have:
- double click on little bar to reset zoom
- fix zoom up/down for consistency

edgardoh:
For the zoom consistency, I have noticed that, but the circle already works
this way in the color zones, so I kept the same. We 'zoom' works different
on the image, the masks, the sliders, etc. and I agree, we should something
about that.

edgardoh:
Ready the reset zoom.

edgardoh:
Just checked, seems that the masks are the problem, so I'll revert it here and we can decide what to do with the masks on a different PR.

edgardoh:
Seems that dt_gui_get_scroll_deltas() is returning the wrong value:

https://github.com/darktable-org/darktable/blob/master/src/gui/gtk.c#L299

so modules multiply it by -1:

https://github.com/darktable-org/darktable/blob/master/src/iop/basecurve.c#L1826

I'm tempted to leave things as they are here and fix that, what do you think?

TurboGit:
For me the way mask are working is fine: Augment size of circle with a scroll-down and decrease size with scroll-up. This is natural, scroll-up is like zoom on image, you concentrate on a zone so zoom-in on image or reduce size of circle.

So I'll fix (invert) the colorzone zoom in/out.

edgardoh:
Done.

TurboGit:
When you zoom-in you have a strange band on the right witch does not have proper display.

edgardoh:
Can't duplicate, any particular setting?

TurboGit:
Something fishy, I have hard time to reproduce. It did happen again to me... But cannot reproduce now. But I see something that may explain the issue.

The histogram seems to be broken, change mode from string to smooth things seems to go wrong at some point.

TurboGit:
I confirm that the issue seems to be with the histogram. Without this patch the background histogram looks correct. But with this PR it is different and seems really wrong.

edgardoh:
I'm comparing it against master and it looks the same to me. Can you take a
snapshot of both right and wrong? Or maybe is a setting I'm missing?

TurboGit:
Ok, it seems to happen on some images and not others. Take the following picture:

https://drive.google.com/file/d/1Mm0zP_AD8Gi_Patd9op5KE0bWRADibLo/view?usp=sharing

Open in darkroom. Activate the colorzone and you'll see no histogram but the whole curve area is greyed as part of the histogram seems to be cover all the area.

edgardoh:
OK, this is frustrating, I'm still get the correct result. With and without
opencl.

Maybe something with the work profile? Or some other active module?

TurboGit:
Indeed frustrating... I have put all my profiles (softproof, screen, histogram) to sRGB. Still the same issue:

Not activated:

Capture d’écran de 2019-03-14 19-26-08

Activated;

Capture d’écran de 2019-03-14 19-26-26

As you can see, the second one when activated the area is far less colorful and I don't have the histogram (well I suppose it is over the whole area).

Note that you need to fly over the curve for the histogram to be refreshed.

TurboGit:
And my history is clean. No base curve, but I have tested with a base curve and still have the same issue.

edgardoh:
Yes, you have to move the mouse over the graph to refresh it when enable
the module, no way around that. But when I do I get the histogram.

It seems that you have a light grey square around the module that I don't
have, maybe some CSS issue?

edgardoh:
This what I get by just going to darkroom and enable the module:

Screenshot_2019-03-14_15-40-46

TurboGit:
With my image?

This is the histogram on my side (without this PR, with previous version):

Capture d’écran de 2019-03-14 19-56-29

TurboGit:
And I have just recompiled with your patch and get the view as described in "Activated" second of my previous message... Something fishy here. Probably some non initialized variables.

TurboGit:
And I have tested with and without OpenCL. Very same issue on both.

TurboGit:
Would be nice to have someone else testing this!

rawfiner:
I will try to find some time tomorrow morning for testing :-)

TurboGit:
Found the issue by looking at the code. This works fine as long as the main histogram is using the linear view. If set to log or waveform it is not working.

edgardoh:
This doesn't seem right, or at least is not what I get with the current
master version. Others histograms work fine? Tone curve, levels, etc..

TurboGit:
Yes on the tonecurve and level the histogram is fine. When in waveform the log view is used but otherwise when in linear or log the histogram in the tonecurve and level are correct and matching the main histogram.

But it looks like even in linear view the histogram in colorzone does not match the main histogram:

I have this with colorzone:

Capture d’écran de 2019-03-14 20-30-24

And this with tonecurve:

Capture d’écran de 2019-03-14 20-28-13

edgardoh:
That seems correct, the histogram is based on the 'select by' channel, and
my french is not very good but it seems that is by hue.

TurboGit:
Yes it is hue. So the only issue is when main histogram is not linear. This does not work on colorzones for some reasons.

edgardoh:
The histogram works only in LCh, there's no option right now to transform
it according the work profile.

TurboGit:
Found and fixed the issue. You were requesting the histogram in log but displaying it in linear form.
See #2214 

If you agree I'll merge.

rawfiner:
I have just tested, I have been able to reproduce the issue.
With #2214 I do not see the issue anymore.

TurboGit:
Thanks for testing. I have just added a fix to avoid negative values in log mode which could makes wrong histogram.

rawfiner:
I did not find how to zoom however
I tried shift-A (which is the shortcut for pan & zoom), and the mouse wheel.
Maybe have I missed something?

TurboGit:
I have this on A key and it works for me.

rawfiner:
Oh, I was pressing the key once instead of pushing it continuously!
It works great :-)
However, when zoomed in, modifying the curve with "edit by area" does not work as expected on my side. To reproduce:
- add 3 nodes close to each other
- zoom in to this zone
- try to move up the central node using the "edit by area" option

edgardoh:
@TurboGit , look goo to me, thanks.

edgardoh:
I'll look into this on a new PR so 2214 can be merged.

PR #2206, opened 2019-03-14
fix format string issues

rabauke:
Platform dependent integer aliases require special format strings. See also https://en.cppreference.com/w/c/types/integer .


TurboGit:
Thanks!

PR #2211, opened 2019-03-14
color picker performance

edgardoh:
The module color picker is reprocessing the entire full and preview pipe when all is needed is to process the preview pipe from the caller module.


TurboGit:
Looks good thanks!

PR #2214, opened 2019-03-15
colorzones zoom - plus fix

TurboGit:
Fix colorzone histogram zoom when histogram is not in linear mode.

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -63,10 +63,6 @@ when updating from the currently stable 2.4.x series, please bear in mind that y
   area from the image. Click to adjust the input image slider,
   ctrl+click to adjust the out image one.
 
-- The color zones module now display an histogram based on the ‘select
-  by’ channel and displays the selected range if the color picker is
-  in area mode.
-
 - The picasa module has been renamed to googlephoto and completely
   rewrote to support the new Google Photo API. It is now again
   possible to create albums directly from the export module.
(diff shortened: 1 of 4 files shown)


junkyardsparkle:
Interesting... after this, the issue I described in #2167 is reversed: now a newly added node appears at the *bottom* of the graph instead of the top.

edgardoh:
@junkyardsparkle , it seems that we don't understand each other.

There are two ways to add a node:

-click and drag: the node is added at mouse position and then it follows the mouse
-cntrl+click: the node is added at x=mouse position, y=curve position

have you found a 3rd way?

junkyardsparkle:
I understand you, and I'm trying very hard to be understandable myself. The behavior I'm describing is with the first case you mention. When I click and drag, the node is *not* added at the mouse position, as you describe (and as I would expect). Peviously, it was being added at the top of the graph, and after this commit, it is now being added at the *bottom*. The x position is always correct. The ctrl+click behavior is also correct. 

By now it's clear to me that I'm seeing a different behavior than you and others. This could be due to some old cairo bug, I don't know. It has never been the case with the other modules which present a "curves" interface, though (including the old color zones). If more people test this and don't get "bitten", it may be too obscure to worry about... there probably aren't too many people still using Slackware 14.2 at this point. ;) I wasn't planning to mention it again, except that the behavior *changed* at this point, which may be a clue.

edgardoh:
If you're getting this by click+drag it won't be easy to fix, it is working fine for me. 
I have an open PR #2217 with some fixes for the color zone, if you can build it then try that, if still not working let me know and we can work together to try to debug it.

junkyardsparkle:
I'm looking around too, but I don't have a good grasp of gtk. Probably the best thing at this point is to get as many people as possible to test the module. If it's broken for anyone else, we can "triangulate".

PR #2217, opened 2019-03-15
fix edit by area zoom

edgardoh:
Fixed in color zones, when edit by area is selected, zoomed in and nodes are edited.

### DIFF
--- a/src/iop/colorzones.c
+++ b/src/iop/colorzones.c
@@ -100,9 +100,9 @@ typedef struct dt_iop_colorzones_gui_data_t
   GtkWidget *chk_edit_by_area;
   int picker_set_upper_lower; // creates the curve flat, positive or negative
   dt_iop_colorzones_channel_t channel;
-  float draw_xs[DT_IOP_COLORZONES_RES], draw_ys[DT_IOP_COLORZONES_RES];
-  float draw_min_xs[DT_IOP_COLORZONES_RES], draw_min_ys[DT_IOP_COLORZONES_RES];
-  float draw_max_xs[DT_IOP_COLORZONES_RES], draw_max_ys[DT_IOP_COLORZONES_RES];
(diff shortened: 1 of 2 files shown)


junkyardsparkle:
The y-offset issues I was having previously appear to be fixed in this branch. :)

src/iop/colorzones.c
```diff
@@ -1239,15 +1224,19 @@ static gboolean _bottom_area_button_press_callback(GtkWidget *widget, GdkEventBu
 
 static int _sanity_check(const float x, const int selected, const int nodes, const dt_iop_colorzones_node_t *curve)
 {
-  int point_valid = 1;
+  int point_valid = TRUE;
```
TurboGit: should be a gboolean

edgardoh:
> *@TurboGit* requested changes on this pull request.
>
> Just a minor comment. An int used as a boolean.
>
> A question, I found that a slight move of the luminence curve in strong
> mode gives very quickly artifacts. Do you reproduce that? As soon as I
> change for the smooth mode all is correct. It seems that the strong mode
> creates some cusp.
>
Can be, according to the comments on the code there's a mask used to avoid
artifacts, but is used only on hue, the lightness and chroma should work
the same. But the side effect of that mask is that it can produce a shift
on the destination hue, so sometimes you may not get the desired effect.
Having this two modes give the user the choice, if there's artifacts on the
image one can always change to the smooth method.
Actually I haven's seen any artifacts, but that may depend on the
definition of artifact, on very noisy images it brings the noise quite a
lot, that can be considered as artifacts?

------------------------------

edgardoh:
Ready.

TurboGit:
Thanks!

TurboGit:
I'll open an issue with an image with artifacts just to be sure there is nothing broken.

edgardoh:
If you're gonna be playing with it and you find the time check how nodes
are added when edit by area is selected. I don't use that mode so for me is
OK like it is, but you never know.

PR #2220, opened 2019-03-16
liquify distord

### LINKED ISSUE #2210: distorted mask not correct
**Describe the bug**
This is not a new issue, but we may want to revisit it. I've never been able to fix it personalty.

**To Reproduce**
1. Open the attached picture in darkroom
2. Activate the monochrome
3. Create a selection on the green square (or use attached .xmp)
4. The square is grey, all ok
5. Activate the perspective correction module and set rotation to 10
6. All OK, the mask is distorted and still cover all the square
7. Activate liquify, and add a single node pushing on one of the face of the square

Here is the issue, the mask is distorted but not correctly and so some green part of the square are not now covered.

8. If the perspective correction module is deactivate then all is OK.

So the issue is only when Liquify and perspective correction are used together.

**Expected behavior**
Mask properly distorted and the green should never be visible.

**Screenshots**

Initial  image with mask:

Capture d’écran de 2019-03-14 14-56-27

With perspective correction:

Capture d’écran de 2019-03-14 14-56-37

With Liquify:

Capture d’écran de 2019-03-14 14-56-44

With both:

Capture d’écran de 2019-03-14 14-56-50

 - OS: Linux
 - Version all up to now

edgardoh:
Fixes #2210

### DIFF
--- a/src/iop/liquify.c
+++ b/src/iop/liquify.c
@@ -447,14 +447,12 @@ int _dev_distort_transform_plus(dt_develop_t *dev, dt_dev_pixelpipe_t *pipe, con
                                   float *points, size_t points_count)
 {
   // this is called from the dt_dev_distort_transform_plus(), so the history is already locked
-//  dt_pthread_mutex_lock(&dev->history_mutex);
   GList *modules = g_list_first(pipe->iop);
   GList *pieces = g_list_first(pipe->nodes);
   while(modules)
   {
     if(!pieces)
     {
-//      dt_pthread_mutex_unlock(&dev->history_mutex);
       return 0;
     }
     dt_iop_module_t *module = (dt_iop_module_t *)(modules->data);
@@ -471,21 +469,18 @@ int _dev_distort_transform_plus(dt_develop_t *dev, dt_dev_pixelpipe_t *pipe, con
     modules = g_list_next(modules);
     pieces = g_list_next(pieces);
   }
-//  dt_pthread_mutex_unlock(&dev->history_mutex);
   return 1;
 }
 int _dev_distort_backtransform_plus(dt_develop_t *dev, dt_dev_pixelpipe_t *pipe, const double iop_order, const int transf_direction,
                                       float *points, size_t points_count)
 {
   // this is called from the dt_dev_distort_backtransform_plus(), so the history is already locked
-//  dt_pthread_mutex_lock(&dev->history_mutex);
   GList *modules = g_list_last(pipe->iop);
   GList *pieces = g_list_last(pipe->nodes);
   while(modules)
   {
     if(!pieces)
     {
-//      dt_pthread_mutex_unlock(&dev->history_mutex);
       return 0;
     }
     dt_iop_module_t *module = (dt_iop_module_t *)(modules->data);
@@ -502,7 +497,6 @@ int _dev_distort_backtransform_plus(dt_develop_t *dev, dt_dev_pixelpipe_t *pipe,
     modules = g_list_previous(modules);
     pieces = g_list_previous(pieces);
   }
-//  dt_pthread_mutex_unlock(&dev->history_mutex);
   return 1;
 }
 
@@ -575,7 +569,7 @@ typedef struct
   float from_scale;
   float to_scale;
   int transf_direction;
-  int from_distort_transform;
+  gboolean from_distort_transform;
 } distort_params_t;
 
 static void _distort_paths (const struct dt_iop_module_t *module,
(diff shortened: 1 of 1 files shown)


edgardoh:
@TurboGit , as I said before I don't really know this module, so a full test should be done on this.

src/iop/liquify.c
```diff
@@ -512,6 +575,7 @@ typedef struct
   float from_scale;
   float to_scale;
   int transf_direction;
+  int from_distort_transform;
```
TurboGit: shouldn't this be a gboolean?

src/iop/liquify.c
```diff
@@ -443,6 +443,69 @@ static void path_delete (dt_iop_liquify_params_t *p, dt_liquify_path_data_t *thi
   node_gc (p);
 }
 
+int _dev_distort_transform_plus(dt_develop_t *dev, dt_dev_pixelpipe_t *pipe, const double iop_order, const int transf_direction,
+                                  float *points, size_t points_count)
+{
+  // this is called from the dt_dev_distort_transform_plus(), so the history is already locked
+//  dt_pthread_mutex_lock(&dev->history_mutex);
```
TurboGit: can be removed

src/iop/liquify.c
```diff
@@ -443,6 +443,69 @@ static void path_delete (dt_iop_liquify_params_t *p, dt_liquify_path_data_t *thi
...
+  GList *modules = g_list_first(pipe->iop);
+  GList *pieces = g_list_first(pipe->nodes);
+  while(modules)
+  {
+    if(!pieces)
+    {
+//      dt_pthread_mutex_unlock(&dev->history_mutex);
```
TurboGit: likewise

src/iop/liquify.c
```diff
@@ -443,6 +443,69 @@ static void path_delete (dt_iop_liquify_params_t *p, dt_liquify_path_data_t *thi
...
+                          || (transf_direction == DT_DEV_TRANSFORM_DIR_BACK_INCL && module->iop_order <= iop_order)
+                          || (transf_direction == DT_DEV_TRANSFORM_DIR_BACK_EXCL && module->iop_order < iop_order)) && 
+      !(dev->gui_module && dev->gui_module->operation_tags_filter() & module->operation_tags()))
+    {
+      module->distort_transform(module, piece, points, points_count);
+    }
+    modules = g_list_next(modules);
+    pieces = g_list_next(pieces);
+  }
+//  dt_pthread_mutex_unlock(&dev->history_mutex);
```
TurboGit: likewise

src/iop/liquify.c
```diff
@@ -443,6 +443,69 @@ static void path_delete (dt_iop_liquify_params_t *p, dt_liquify_path_data_t *thi
...
+  return 1;
+}
+int _dev_distort_backtransform_plus(dt_develop_t *dev, dt_dev_pixelpipe_t *pipe, const double iop_order, const int transf_direction,
+                                      float *points, size_t points_count)
+{
+  // this is called from the dt_dev_distort_backtransform_plus(), so the history is already locked
+//  dt_pthread_mutex_lock(&dev->history_mutex);
```
TurboGit: can be removed

src/iop/liquify.c
```diff
@@ -619,9 +696,10 @@ static void _distort_paths (const struct dt_iop_module_t *module,
 static void distort_paths_raw_to_piece (const struct dt_iop_module_t *module,
                                         dt_dev_pixelpipe_t *pipe,
                                         const float roi_in_scale,
-                                        dt_iop_liquify_params_t *p)
+                                        dt_iop_liquify_params_t *p,
+                                        const int from_distort_transform)
```
TurboGit: const gboolean ?

edgardoh:
> *@TurboGit* requested changes on this pull request.
>
> I'm impressed! Congrats this works just fine. I have some minor comments.
>
> One question, would this have been possible before your work on the pixel
> pipe? I lost track of the implementation back when I worked on this iop.
>
Thanks. And yes, I haven't changed the way the pipe or the distord works,
so this should work even without the module order changes. And the pipe and
the distord works very well, seems that all the distord modules can have
any order, that's really nice.

A couple of remarks/questions:

-I think is better to leave the original code commented in this 2
functions, makes sense to me as they must be an exact copy of the original
(minus the lock)

-You also requested to use gboolean on my other PR. I always use int for
bool expressions, is standard C and is used this way all over dt. Is there
any advantage of using gboolean?


------------------------------

TurboGit:
Yes there is advantage at least in readability. When you see a gboolean you know that it is a 2 states variables. And this is also important for consistency as all dt code do use gboolean when possible. I'm sure there is case where this is not done, but they should be fixed when possible.

edgardoh:
I really don't agree but I will assume that this has already been discussed
and I missed, so too bad for me. I'll fix both PR.

edgardoh:
Ready

TurboGit:
Thanks!

PR #2248, opened 2019-03-19
allow rgb modules to be before colorin

### LINKED ISSUE #2245: segfault if using opencl in rgbcurve if inputprofile is set via style
**Describe the bug**
applying a mask in rgbcurve results in a segmentation fault if 
- opencl ist activated and 
- a mask within a different module is already defined 
- and input profile is set via applied style .
i found it not beeing reproducible if the colorin section of the attached xmp file is deleted 

**To Reproduce**
1. open file
2. add 'exposure' and apply a gradient mask
3. add 'rgb curve' and apply a gradient mask
4. add style [Canon M6 Neutral.dtstyle.txt](https://github.com/darktable-org/darktable/files/2980421/Canon.M6.Neutral.dtstyle.txt)

resulting xmp file: [20190223-IMG_5937.CR2.xmp.txt](https://github.com/darktable-org/darktable/files/2980445/20190223-IMG_5937.CR2.xmp.txt)

58,034366 [dev_pixelpipe] took 0,275 secs (0,752 CPU) processed `Entrastern' on GPU with tiling, blended on CPU [thumbnail]
[dev_pixelpipe] module `Entrastern' min: (0,000512; 0,000000; 0,000000) max: (0,620160; 0,617545; 0,618875) [thumbnail]
58,034740 [opencl memory] device 0: 290400 bytes (0,3 MB) in use
58,035256 [opencl memory] device 0: 580800 bytes (0,6 MB) in use
58,036358 [opencl memory] device 0: 581056 bytes (0,6 MB) in use
58,036382 [opencl memory] device 0: 653656 bytes (0,6 MB) in use
58,036457 [masks Verlauf #1] gradient draw took 0,0001 sec
58,036472 [masks Verlauf #1] gradient transform took 0,0000 sec
58,036564 [masks Verlauf #1] gradient fill took 0,0001 sec
58,036612 [masks 0] combine took 0,0002 sec
58,036618 [masks] render all masks took 0,0002 sec
58,036631 [opencl memory] device 0: 726256 bytes (0,7 MB) in use
Magick: abort due to signal 11 (SIGSEGV) "Segmentation Fault"...
Abort trap: 6
full console output: [opencl_mask_rgbcurve.txt](https://github.com/darktable-org/darktable/files/2980451/opencl_mask_rgbcurve.txt)

 - OS: OSX
 - Version
this is darktable 2.7.0+769~gb311cd37c-dirty

compile options:
  bit depth is 64 bit
  normal build
  OpenMP support enabled
  OpenCL support enabled
  Lua support enabled, API version 5.0.1
  Colord support disabled
  gPhoto2 support enabled
  OpenEXR support enabled

edgardoh:
Fixes #2245

-basic adjustments and rgb curve can now be before colorin
-fixed the opencl call in blend module that wasn't allowing NULL parameters
-fixed linked curve creation nodes when using the 2nd color picker

### DIFF
--- a/data/kernels/basicadj.cl
+++ b/data/kernels/basicadj.cl
@@ -95,7 +95,8 @@ basicadj(read_only image2d_t in, write_only image2d_t out, const int width, cons
            const int process_saturation, const float saturation,
            const int process_hlcompr, const float hlcomp, const float hlrange,
            const float middle_grey, const float inv_middle_grey,
-           global const dt_colorspaces_iccprofile_info_cl_t *profile_info, read_only image2d_t lut)
+           global const dt_colorspaces_iccprofile_info_cl_t *profile_info, read_only image2d_t lut,
+           const int use_work_profile)
 {
   const int x = get_global_id(0);
   const int y = get_global_id(1);
@@ -111,7 +112,7 @@ basicadj(read_only image2d_t in, write_only image2d_t out, const int width, cons
   // highlight compression
   if(process_hlcompr)
   {
-    const float lum = get_rgb_matrix_luminance(pixel, profile_info, lut);
+    const float lum = (use_work_profile == 0) ? dt_camera_rgb_luminance(pixel): get_rgb_matrix_luminance(pixel, profile_info, lut);
     if(lum > 0.f)
     {
       const float ratio = hlcurve(lum, hlcomp, hlrange);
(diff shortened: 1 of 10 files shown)


aurelienpierre:
The problem of doing that is every module using the XYZ Y channel (luminance), for example saturation adjustments or curve color preservation, will get a wrong Y value before the input profile, because it's only this module that maps input RGB to standard XYZ D50 (and then to Lab).

edgardoh:
And that's why those modules are after colorin, this is for backward
compatibility, I'm not forcing anyone to move the modules before colorin.

aurelienpierre:
Not forcing people to do stupid things is one thing, but users have no way to guess that it is stupid, because they likely don't know what XYZ is, and how it is used to compute the saturation. And, as much as I like offering options and letting users decide, they should at least have some info to realize it's a bad idea.

PR #2250, opened 2019-03-19
color picker updates

edgardoh:
-This adds the piece to the picker->apply() so the color space profile can be used if needed.

-There is some flickering on the rgb curve histogram when using the second color picker, this is because the picked values are set to -inf long before the new values are calculated and the draw() is picking this intermediate values, so I use a temp variable for the calculations and then update the actual fields.

-The global picker is not showing the right values, but for now there's no way to fix this.


TurboGit:
Looks good thanks!

PR #2253, opened 2019-03-20
update contextual help url with new doc url

rawfiner:
Also removes espanol from the languages currently supported for online documentation
The code checks if the old url is used to update it.


TurboGit:
Thanks!

PR #2254, opened 2019-03-20
lib: fix function signature

rabauke:
See definition of `expandable` in `dt_lib_module_t`.


TurboGit:
Indeed! Good catch. Thanks.

PR #2257, opened 2019-03-20
color picker updates

edgardoh:
-Add the global color picker to rgb curve and color zones

-Avoid recursion when calling apply()

-Make modules remember the color picker area

### DIFF
--- a/src/common/iop_order.c
+++ b/src/common/iop_order.c
@@ -2010,6 +2010,15 @@ dt_iop_order_iccprofile_info_t *dt_ioppr_set_pipe_work_profile_info(struct dt_de
   return profile_info;
 }
 
+dt_iop_order_iccprofile_info_t *dt_ioppr_get_histogram_profile_info(struct dt_develop_t *dev)
+{
+  dt_colorspaces_color_profile_type_t histogram_profile_type;
+  char *histogram_profile_filename;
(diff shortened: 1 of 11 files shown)


edgardoh:
When a parameter is changed inside the apply() the pipe is processed and the apply() is called again, this is not an infinite recursion because this second time params are the same, so the pipe is not executed again, but I don't want it to be executed even this time.
This should be added to all modules that modify parameters inside the apply(), but let's test it with this modules first to be sure is working.

There's one issue left with the color picker, related to my PR on performance. When the parameters or the area of the picker is not updated the pipe is not executed and the apply() is not called. That's OK in almost every scenario, but in the modules where the picker work with modifiers, even if the area is not modified, the apply() should be called, To duplicate:
-on rgb curve click on the second color picker, nodes are added depending on the area
-without modifying anything else ctrl+click on the second picker, nothing happens, but it should create a positive curve.

I don't want to go back to re-process the entire pipe just for this, and I don't want to modify how the cache or the hash works, and I didn't find any other way around it yet.

TurboGit:
The issue is minor as the solution is to deselect the picker and re-select it with ctrl+click. So I agree with you, we don't want to remove the perf enhancement for this.

TurboGit:
Apart from that it looks fine to me. Love the remember area! Thanks.

PR #2259, opened 2019-03-20
remove meaningless const declarations in return types

rabauke:
Const declarations for return types are meaningless and unusual. They may confuse programmers and trigger warnings by pedantic compilers. Also removes a const cast.


TurboGit:
Ok, thanks!

PR #2261, opened 2019-03-21
check for NULL on db start up

edgardoh:
check for NULL on db start up

### DIFF
--- a/src/common/iop_order.c
+++ b/src/common/iop_order.c
@@ -1022,12 +1022,41 @@ int dt_ioppr_check_db_integrity()
   int ret = 0;
   sqlite3_stmt *stmt;
 
-  DT_DEBUG_SQLITE3_PREPARE_V2(dt_database_get(darktable.db), "SELECT * FROM main.history WHERE iop_order <= 0 OR iop_order IS NULL",
+  DT_DEBUG_SQLITE3_PREPARE_V2(dt_database_get(darktable.db), "SELECT imgid, operation, module FROM main.history WHERE iop_order <= 0 OR iop_order IS NULL",
                               -1, &stmt, NULL);
   if(sqlite3_step(stmt) == SQLITE_ROW)
   {
     ret = 1;
     fprintf(stderr, "\nThere are unassigned iop_order in the history!!!\n\n");
+    
+    int count = 0;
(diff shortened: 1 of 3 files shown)


src/common/iop_order.c
```diff
@@ -1016,6 +1017,24 @@ int dt_ioppr_move_iop_after(GList **_iop_list, dt_iop_module_t *module, dt_iop_m
...
 // from here just for debug
 //--------------------------------------------------------------------
+int dt_ioppr_check_db_integrity()
+{
+  int ret = 0;
+  sqlite3_stmt *stmt;
+
+  DT_DEBUG_SQLITE3_PREPARE_V2(dt_database_get(darktable.db), "SELECT * FROM main.history WHERE iop_order <= 0 OR iop_order IS NULL",
+                              -1, &stmt, NULL);
+  if(sqlite3_step(stmt) == SQLITE_ROW)
```
TurboGit: I would go even farther. I would output the number of broken history entries and display the first 20 (columns: imgid, operation, module) or so. This could be useful to actually look at library.db and fix the issues.

What do you think?

edgardoh:
I also added the styles to the check, let's see how it goes.

src/common/iop_order.c
```diff
@@ -1016,6 +1017,53 @@ int dt_ioppr_move_iop_after(GList **_iop_list, dt_iop_module_t *module, dt_iop_m
...
+
+  DT_DEBUG_SQLITE3_PREPARE_V2(dt_database_get(darktable.db), "SELECT imgid, operation, module FROM main.history WHERE iop_order <= 0 OR iop_order IS NULL",
+                              -1, &stmt, NULL);
+  if(sqlite3_step(stmt) == SQLITE_ROW)
+  {
+    ret = 1;
+    fprintf(stderr, "\nThere are unassigned iop_order in the history!!!\n\n");
+    
+    int count = 0;
+    while(sqlite3_step(stmt) == SQLITE_ROW && count++ < 20)
```
TurboGit: Here you get the second stmt as the first one is already retrieved from previous sqilte3_step().

edgardoh: Right! Fixed.

TurboGit:
Ok, looks good! I'll squash the 3 commits into a single one to ease reverting (if needed) for the release. We'll see how it goes. Thanks.

PR #2263, opened 2019-03-22
fix rgb colorspace transform

edgardoh:
C version of RGB <--> RGB transform was not working when input and output image are different.


TurboGit:
Thanks!

PR #2264, opened 2019-03-22
read tiff colorspace

### LINKED ISSUE #2252: New non-CIELAB pipe breaks the darktable-chart utility
Previously, the `darktable-chart` utility made use of [a "hidden" darktablerc option to enable a no-op output color profile which dumped the CIELAB data directly to a file](https://www.darktable.org/usermanual/en/darktable_chart_making_input_images.html). The recent changes to the pipeline in the master branch have broken this "feature". Prior to any 2.8 release, this should probably be addressed somehow.

edgardoh:
Lab tiff files are supported but were not handled.

Fixes #2252

### DIFF
--- a/src/common/imageio.c
+++ b/src/common/imageio.c
@@ -489,7 +489,7 @@ dt_imageio_retval_t dt_imageio_open_ldr(dt_image_t *img, const char *filename, d
   ret = dt_imageio_open_jpeg(img, filename, buf);
   if(ret == DT_IMAGEIO_OK || ret == DT_IMAGEIO_CACHE_FULL)
   {
-    img->buf_dsc.cst = iop_cs_rgb;
+    img->buf_dsc.cst = iop_cs_rgb; // jpeg is always RGB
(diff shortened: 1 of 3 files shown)


TurboGit:
Looks good! Thanks again, sorry I keep repetting that but you are really doing a great job!

edgardoh:
Thanks!

This fix is correct as the tiff colorspace was not handled, but I'm not
sure that it will fix the reported issue as I don't really know how that
module works, maybe is something else going on.

junkyardsparkle:
I haven't had a chance to check yet, but note that #2252 didn't actually involve a Lab TIFF file; the problematic TIFF was in the native linear scanner color space, the PFM file required by `darktable-chart` needs to be *exported* in Lab. In any case,, I'll try to have something actually useful to report shortly.

junkyardsparkle:
TIFF file can now be successfully exported as Lab PFM file from a clean config, thanks for your effort! Now I'll try to figure out why it's not working from an updated 2.6 configuration...

PR #2265, opened 2019-03-22
hazeremoval: remove duplicate code

rabauke:
The haze removal module is based on a guided filter. A guided filter was also introduced in darktable 2.6 for general use in various modules (currently only used for blending with drawn and parametric masks). There is no longer a reason that the haze removal module comes with its own implementation.


TurboGit:
Good job! Thanks.

PR #2267, opened 2019-03-22
lighttabe : full preview & exposé zoom

AlicVB:
This add the ability to zoom and pan inside images in full preview mode and exposé layout (in exposé, zooms and positions are synchronized between images)

To handle this I've added some caching mechanisms to avoid exporting full image multiple times : 
- a new mipmap size to load full size images. This "special" mipmap level images are only store on disk if user tick a new preference entry (deactivated by default, because it can create huge cache size)
- full image are keep in memory to avoid reload when panning. This can eat quite a lot of memory, especially in expose layout. So there's an hardcoded limit (fixed at 6) above that, no zooming is allowed (and no caching occurred). Not sure if this limit should better grow darktable.rc

Here it "seems" to work, but please test as there's so many way to display images in lighttable...

### DIFF
--- a/data/darktableconfig.xml.in
+++ b/data/darktableconfig.xml.in
@@ -42,6 +42,13 @@
     <shortdescription>enable disk backend for thumbnail cache</shortdescription>
     <longdescription>if enabled, write thumbnails to disk (.cache/darktable/) when evicted from the memory cache. note that this can take a lot of memory (several gigabytes for 20k images) and will never delete cached thumbnails again. it's safe though to delete these manually, if you want. light table performance will be increased greatly when browsing a lot. to generate all thumbnails of your entire collection offline, run 'darktable-generate-cache'.</longdescription>
   </dtconfig>
+  <dtconfig prefs="core" section="cpugpu">
+    <name>cache_disk_backend_full</name>
+    <type>bool</type>
+    <default>false</default>
+    <shortdescription>enable disk backend for full preview cache</shortdescription>
+    <longdescription>if enabled, write full preview to disk (.cache/darktable/) when evicted from the memory cache. note that this can take a lot of memory (several gigabytes for 20k images) and will never delete cached thumbnails again. it's safe though to delete these manually, if you want. light table performance will be increased greatly when zooming image in full preview mode.</longdescription>
+  </dtconfig>
   <dtconfig prefs="core" section="quality">
     <name>cache_color_managed</name>
     <type>bool</type>
--- a/data/darktableconfig.xml.in
+++ b/data/darktableconfig.xml.in
@@ -1844,6 +1844,13 @@
     <longdescription>number of images to load in advance in the background when showing full-size preview</longdescription>
   </dtconfig>
   <dtconfig>
+    <name>plugins/lighttable/preview/max_in_memory_images</name>
+    <type>int</type>
+    <default>4</default>
+    <shortdescription>maximum number of full-res images to load in memory</shortdescription>
+    <longdescription>if more images are display in expose mode, zooming will be desactivated</longdescription>
+  </dtconfig>
+  <dtconfig>
     <name>codepaths/sse2</name>
     <type>bool</type>
     <default>true</default>
--- a/src/common/focus.h
+++ b/src/common/focus.h
@@ -216,7 +216,8 @@ static void dt_focus_create_clusters(dt_focus_cluster_t *focus, int frows, int f
 }
 
 static void dt_focus_draw_clusters(cairo_t *cr, int width, int height, int imgid, int buffer_width,
-                                   int buffer_height, dt_focus_cluster_t *focus, int frows, int fcols)
+                                   int buffer_height, dt_focus_cluster_t *focus, int frows, int fcols,
+                                   float full_zoom, float full_x, float full_y)
 {
   const int fs = frows * fcols;
   cairo_save(cr);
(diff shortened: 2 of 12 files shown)


Nilvus:
Oh yeah, long time awaiting this. That's seems to be exactly what I've described for new feature with this feature request : https://redmine.darktable.org/issues/12597
Thanks for that ! I will test this.

Nilvus:
I've made few tests. I've some times on file navigation mode (with Alt+z function) darktable crash with this message (in terminal) : "[mipmap cache] memory allocation failed!". I have 8gb of RAM. On expose mode, I have difficulties to have 100% zoom.
Apart these bugs, that's a really great feature. I make a lot of birds photos and it's very useful.

A feature to add that would be great would be a on/off option for synchronization between images on expose mode.

Again, thanks a lot for this long awaiting feature. I love it. I will make more tests.

AlicVB:
Yes, the memory part is quite difficult to handle especially in expose mode where you quickly have to keep in memory lot of full buffers. One thing you can test, is to allow to store full preview image on disk (there's a new entry in preference). It should avoid to recompute the whole pipe each time you need the image. But it'll not solve your crash, I fear...
Some questions : 
- What are your images sizes ?
- Alt-Z trigger the "sticky preview", right ?
- Do you enter in this sticky preview mode with more than 1 image selected ?

Nilvus:
I've tried with the new entry in pref and that's not solve the crash.
Image sizes used : about 24 Mo
Yes, Alt-Z trigger the "sticky preview" and not I don't enter in this mode with more than 1 image selected. I always enter in this mode with only one image selected.

TurboGit:
Working fine on my side. Very very nice. When I zoom with 6 images it takes quite some time to update the cache for the 6 pictures. I'm wondering if limiting to 4 won't be better.

AlicVB:
> Image sizes used : about 24 Mo
That's strange, because your mipmap message means that you don't have enough ram to handle the image buffer, which should in any case be < 1Go, if I count right...

> I'm wondering if limiting to 4 won't be better.
Let's add a "hidden" pref and set it to 4 by default.

Nilvus:
Even the new commit doesn't change crash I have. I can't have 100% zoom with ctrl+ wheel mouse on file navigation mode but it's ok on expose view.

darktable crash each time I'm on 100% zoom and make another ctrl+wheel mouse (like trying to go above 100% zoom). I have the same bug on expose and file navigation mode.

One time when darktable crash, I've had this message on terminal :
```
double free or corruption (fasttop)
Abandon (core dumped)
```
Other times, It's the message : ""[mipmap cache] memory allocation failed!"

This message remains even after increase memory options on darktable preferences.

AlicVB:
Last commit doesn't change anything for your problem.
What I don't really get is why your system can't allocate the memory. Can you watch how many ram unused you have before the crash ? maybe try with other "untouched" images too (to exclude the pipe) ?
If you can, maybe add a printf in src/common/mimap_cache.c before line 328 to show the value of entry->data_size ? (something like printf("size to alloc : %ld\n", entry->data_size); )

edgardoh:
Works great for me, no crash at all.

TurboGit:
A very nice addition to darktable! Thanks for this, it has been asked many times. Great job.

Nilvus:
It's quite strange. Size allocated changes a lot. See that when I've tried to use shortcut to go from fit screen to 100% zoom in expose mode (with darktable crash) :

`size to alloc : 79232
size to alloc : 79232
size to alloc : 79232
size to alloc : 79232
size to alloc : 79232
size to alloc : 79232
size to alloc : 79232
size to alloc : 79232
size to alloc : 79232
size to alloc : 79232
size to alloc : 79232
size to alloc : 79232
size to alloc : 79232
size to alloc : 79232
size to alloc : 79232
size to alloc : 16384032
size to alloc : 41943072
size to alloc : 65536032
size to alloc : 3233394724
[mipmap cache] memory allocation failed!
`
I have a lot of crash with these new functions, even with 2,8 Go of RAM remaining when darktable is open and more than 800 Mb of swap file (I only use a swap file of 1Go. I have to say that i have increase it from 512Mb to 1Gb and it doesn't change anything on darktable crash). I'm quite surprise to see that edgardoh and pascal didn't have any crash.

Nilvus:
I probably found why I have crash. It's my swap file. I don't know why but darktable use my swap file before RAM. Don't know why because I have a setting which ask Linux to use Swap file only if it remains 5% of RAM. I have seen a crash at the exact time my swap file was full with my RAM using only 3.9 Gb on 7,7Gb.

Nilvus:
I can confirm that it was swap file size the problem. I've increase its size to 6Gb and no crash anymore. These functions needs lot of memory, even with only one or 2 images.

parafin:
You should check how much memory DT tries to allocate in the malloc that fails. It's probably some very large number.

Nilvus:
> You should check how much memory DT tries to allocate in the malloc that fails. It's probably some very large number.

Of course, sometime, DT allocate a very large amount of memory, as you can use in a previous comment or in this :

`size to alloc : 79232
size to alloc : 3233394724
size to alloc : 1056
size to alloc : 79232
`
Ok, so what to do with it ? I've increased my swap file and some parameters and DT preferences. That's better but still remains sometime bugs (display glitches or more rarely crash). Each time, these bugs appear, each time I have the allocated size increase a lot, like the example on this comment. And when darktable don't crash and the glitch disappear, when dt go in normal usage, the allocated size go to a little number then to a more normal number. When all is ok and darktable run great, size to alloc is always on 79232 on my system.

AlicVB:
@Nilvus : I think I know where the problem is (fixed cache size, vs dynamic)
I'll try to come with a fix asap. I need to understand how all this cache memory things are working...

TurboGit:
I think it would be nice to have somewhere the indication of the current zoom level. Do you think it would be possible? Maybe adding this in the hint area ?

AlicVB:
@Nilvus : please test PR #2287, it should solve your problem.
@TurboGit : sure, I'll have a look, but messages in the hint area will not be visible in full preview...

Nilvus:
@AlicVB : I will try to test it this week but I'm not sure to have time. This week and next one will be particularly busy. Thanks for your work.

Nilvus:
PR #2287 is ok, no more crash now. Thanks for your work.

PR #2270, opened 2019-03-23
timeline : update state when changing view

### LINKED ISSUE #2268: Expose and timeline when returning from darkroom issue
**Describe the bug**
While in expose mode if I double click on an image I can go to darkroom (sometimes it takes a few try), when returning to lighttable the timeline is also visible to the left of the filmstrip.

If I double click on the filmstrip I can't go into darkroom.

**To Reproduce**
In expose mode double click on an image (re-try if doesn't work), it goes to darkroom
Double click again, it goes to lighttable, but the timeline is also visible.

Doble click on any image on the filmstrip, it does not go to darkroom.

**Expected behavior**
Timeline is not visible and I can go to darkroom by double click on the filmstrip.

AlicVB:
this should fix #2268 for the timeline part

### DIFF
--- a/src/views/lighttable.c
+++ b/src/views/lighttable.c
@@ -2218,10 +2218,19 @@ void enter(dt_view_t *self)
 
   // show/hide filmstrip when entering the view
   dt_lib_module_t *m = darktable.view_manager->proxy.filmstrip.module;
+  dt_lib_module_t *timeline = darktable.view_manager->proxy.timeline.module;
+  gboolean vs = dt_lib_is_visible(timeline);
+
   if(get_layout() == DT_LIGHTTABLE_LAYOUT_EXPOSE)
+  {
+    gtk_widget_hide(GTK_WIDGET(timeline->widget));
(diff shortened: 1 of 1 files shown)


edgardoh:
An off-topic question, is there a way to switch between timeline and filmstrip? I can't find any...

AlicVB:
Timeline and filmstrip shouldn't be display at the same time : filmstrip is only available for the expose layout where timeline should always be hidden...

edgardoh:
>
> Timeline and filmstrip shouldn't be display at the same time
>
Sure, that's reasonable.

> : filmstrip is only available for the expose layout where timeline should
> always be hidden...
>
Is there a reason for this? Sometimes I would like to have the filmstrip on
the file manager, depending on the task it works better for me.

TurboGit:
Thanks!

TurboGit:
@edgardoh ; I would find the filmstrip + lighttable quite redundant on my side. But I'm not opposed to having this possibility. Let's open an issue for this if really needed.

PR #2272, opened 2019-03-23
fix asan build

### LINKED ISSUE #2112: asan build error
When building with  --asan I get the following error:

src/common/opencl.c: In function ‘dt_opencl_device_init’:
src/common/opencl.c:131:12: error: stack usage might be 32864 bytes [-Werror=stack-usage=]
 static int dt_opencl_device_init(dt_opencl_t *cl, const int dev, cl_device_id *devices, const int k,

edgardoh:
fix buil error while using asan
also fix some memory leaks

fixes #2112

### DIFF
--- a/src/common/opencl.c
+++ b/src/common/opencl.c
@@ -183,6 +183,16 @@ static int dt_opencl_device_init(dt_opencl_t *cl, const int dev, cl_device_id *d
   cl_uint vendor_id = 0;
   cl_bool little_endian = 0;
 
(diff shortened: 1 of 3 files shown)


TurboGit:
Looks like the asan errors where false positive, no?

edgardoh:
You mean the build errors when compiling with --asan? There was a potential heap overflow there.

TurboGit:
Yes, potential, but I'm not sure it was the case. Anyway, this looks good.

PR #2273, opened 2019-03-23
Lua API: darkroom_display_image

wpferguson:
Adds a function to darktable.gui.views.darkroom to change the displayed image.  Currently the only way to change the image displayed in darkroom view from lua is by changing back to lighttable view, selecting a different image, then changing to darkroom view.

This would allow tethering type scripts (i.e. tools/watch_folder.sh) to open an imported image in darkroom view, then import a new image and simply display it.

Satisfies issue 12640.


TurboGit:
Sounds good! Thanks.

PR #2276, opened 2019-03-23
blend: replace macro by inline function

rabauke:
Employing macro functions is bad, in particular when complex expressions are passed as macro arguments. This can lead to multiple evaluations of these complex expressions. Not to mention the lack of type safety. Negative effects of macros can be avoided by inline functions without compromising performance.

Furthermore, I have removed many of the redundant parentheses. I tend to agree that redundant parentheses can be useful to clarify intent in complex expressions when operator precedence might be not easy to gasp. Here, however, the use of redundant parentheses was excessive (imho).

### DIFF
--- a/src/common/math.h
+++ b/src/common/math.h
@@ -1,6 +1,6 @@
 /*
  *    This file is part of darktable,
- *    copyright (c) 2018 Heiko Bauke.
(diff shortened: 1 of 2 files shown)


TurboGit:
Looks good to me. Thanks.

PR #2277, opened 2019-03-24
Lua API: hide, show and query visibility of gui panels

wpferguson:
Extended the lua API to allow hiding, showing and querying the visibility of the gui top, bottom and side panels.  

These functions are useful for making more space available for the central window.  

My specific use case is I've written a lua script to compare images.  So if for instance you've taken a series of headshots and are trying to select the best one to work on you would select the images, click the compare button in the selected images module and enter "compare" mode.  The lighttable gets set to filemanager mode and 2 images (issue 12614) with the panels collapsed to get the largest view available.  The first image is the "selected" one.  Using shortcuts you step through the images comparing them to the "selected" one.  If you find one you like better, you swap them.  When you exit compare mode you go back to lighttable with the "selected" image as the selection. The lighttable view goes back to the previous settings and the panel visibility is restored.

This satisfies issue 12641 and partially satisfies issue 9902

### DIFF
--- a/doc/usermanual/lua/lua_api.xml
+++ b/doc/usermanual/lua/lua_api.xml
@@ -742,6 +742,100 @@
 </variablelist>
 </section>
 
+<section status="final" id="darktable_gui_panel_visible">
+<title>darktable.gui.panel_visible</title>
+<indexterm>
+<primary>Lua API</primary>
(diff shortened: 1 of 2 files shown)


TurboGit:
Looks good to me! Thanks.

PR #2283, opened 2019-03-24
Variance computation mode in denoiseprofile

rawfiner:
### Add the ability to compute the variance after the anscombe transform is performed and show it in the GUI of denoiseprofile:
Capture du 2019-03-24 23-23-09

### Why checking variance in denoiseprofile?
denoiseprofile is doing a generalized anscombe transform based on a profile to transform the image so that the noise variance become independant of the overall brightness, and becomes equal to 1.

The noise profiling tool seems to be based on the hypothesis that noise is fine grain:
the noiseprofiling tool assumes all the noise lies in the finest scale of a wavelet decomposition, which is not true in practice.
The real problem is that the noiseprofiling tool is validating itself its own profiles, with the same methods, so the profiles appeared ok even if they were not.
That's why we need a way to compute variance in denoiseprofile: computing variance on a noisy picture of a uniform surface that has been transformed using the profile allows to check that the profile is correct.

### How to use?
Take a picture of a uniform surface (the variance computed is the global variance, thus is it influenced by all variations in the image). As an example, you should get a picture like this, perfectly uniform, with only variations coming from the noise: 
DSCF6591
In the darkroom, zoom at 100% and activate denoiseprofile in computevariance mode. The variance values should be close to 1 if the profile is accurate.
On my test pictures, profiles are strongly erroneous unfortunately, I have to push force to 12-15 to get proper variance values...

### Why is it implemented this way?
I wanted to keep this implementation simple, as it will be used for a very sensitive purpose.
That's why there is no opencl implementation, and that the cpu implementation is not parallelized and use recursive functions.
Note that it is already efficient: it takes about half a second to process 24Mpix. So, to process the darkroom preview, I usually had execution times of about 0.001 seconds.
Now why having a recursive implementation? The implementation uses a divide-and-conqueer principle so that the sums of 2 floats always sum 2 numbers of the same order of magnitude, in order to have a precise result.

### DIFF
--- a/src/iop/denoiseprofile.c
+++ b/src/iop/denoiseprofile.c
@@ -2269,6 +2269,133 @@ error:
   return FALSE;
 }
 
+static void sum_rec(const unsigned npixels, const float *in, float *out)
+{
+  if(npixels <= 3)
+  {
+    for(int c = 0; c < 3; c++)
+    {
+      out[c] = 0.0;
+    }
+    for(int i = 0; i < npixels; i++)
+    {
+      for(int c = 0; c < 3; c++)
+      {
+        out[c] += in[i * 4 + c];
+      }
+    }
+    return;
+  }
+
+  unsigned npixels_first_half = npixels >> 1;
+  unsigned npixels_second_half = npixels - npixels_first_half;
+  sum_rec(npixels_first_half, in, out);
+  sum_rec(npixels_second_half, in + 4 * npixels_first_half, out + 4 * npixels_first_half);
+  for(int c = 0; c < 3; c++)
+  {
+    out[c] += out[4 * npixels_first_half + c];
+  }
+}
+
+/* this gives (npixels-1)*V[X] */
+static void variance_rec(const unsigned npixels, const float *in, float *out, const float mean[3])
+{
+  if(npixels <= 3)
+  {
+    for(int c = 0; c < 3; c++)
+    {
+      out[c] = 0.0;
+    }
+    for(int i = 0; i < npixels; i++)
+    {
+      for(int c = 0; c < 3; c++)
+      {
+        float diff = in[i * 4 + c] - mean[c];
+        out[c] += diff * diff;
+      }
+    }
+    return;
+  }
+
+  unsigned npixels_first_half = npixels >> 1;
+  unsigned npixels_second_half = npixels - npixels_first_half;
+  variance_rec(npixels_first_half, in, out, mean);
+  variance_rec(npixels_second_half, in + 4 * npixels_first_half, out + 4 * npixels_first_half, mean);
+  for(int c = 0; c < 3; c++)
+  {
+    out[c] += out[4 * npixels_first_half + c];
+  }
+}
+
+static void process_variance(struct dt_iop_module_t *self, dt_dev_pixelpipe_iop_t *piece, const void *const ivoid,
+                             void *const ovoid, const dt_iop_roi_t *const roi_in,
+                             const dt_iop_roi_t *const roi_out)
+{
+  const dt_iop_denoiseprofile_data_t *const d = piece->data;
+
+  const int width = roi_in->width, height = roi_in->height;
+  size_t npixels = (size_t)width * height;
+
+  memcpy(ovoid, ivoid, npixels * 4 * sizeof(float));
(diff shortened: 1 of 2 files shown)


edgardoh:
Not sure I get what is this used for. Let's say I calculate the variance, then what? Should I enter that value somewhere else?

rawfiner:
@edgardoh the main usage will be to check that noise profile are correct. In my opinion, this is the only way to be confident about the noise profiling: if the values obtained on a flat image are really different than 1, there is an issue. For instance, I tested that with shots of a flat uniform surface at ISO 6400 to 25600 with my camera and I got variance values in the range [12-16], which shows the profile was very wrong (or there is an error in the variance computation, but I kept it as simple as possible to minimise this risk)

As the force slider modifies the values obtained with the profile (a force of 2 is almost equivalent to changing the profile to the next iso value), it can also be used to find how close we are from the correct value.

rabauke:
This new feature might be useful to check the quality of noise profiles. But I cannot see any benefit when it comes to routine image editing. As it is a function of limited use for developers (and possibly for advanced users) it should be present only when activated in the configurations. I think we have something similar for image export in LAB color space. Furthermore, afaik noise profiles are checked for uniformity of the transforned data when they are created. Thus, I am a little bit surprised when you report issues with noise profiles here.

rawfiner:
@rabauke I agree, I did not think about adding a configuration entry for this, and that's a good idea! Thanks. :-)

Noise profiles are checked under the same hypothesis that they are created. To create a noise profile, the following steps are taken:
- export a pfm (basically the float values) with all modules disabled except the mandatory ones (demosaic, input and output color profile). Input color profile is set to REC709, but that is not an identity function, so the value we get for profiling is already different that the one manipulated by denoiseprofile. This could explain why we get different profiles for red, green and blue channels.
- the noiseprofiling tool reads the pfm, does a wavelet decomposition of 1 scale, and assumes that all the noise is in the first scale, and that the residual is smooth. This is not true in practice, we have coarser noise. But still, it generates some profile.
- to check the profile, the noiseprofiling tool applies the anscombe transform on the data, then do again a wavelet decomposition of one scale to separate the noise from the real data. This is where we have a problem: we check the profile using exactly the same procedure as the one used to create it, which is based on erroneous assumptions. That's why I would like another way to check within the module's core the profile's correctness. This way, we can be confident about the data we manipulate.

edgardoh:
@rawfiner , I didn't try it yet, but let's say I find out that the profile for my camera is not good, what should I do? Generate a new profile? I don't know how profiles are generated (I can look into that), but the program hasn't changed, how do I get better results?

rawfiner:
@edgardoh I am working on an update for the noise profiling tool. Once it is ready, I guess we will have to generate a new profile for all cameras, unfortunately... If we are lucky we might find some relationship between the broken profiles and the real ones, to convert the profiles without having to regenerate them, but I currently don't know if such relationship exists and can be found.

edgardoh:
Sounds great, thanks for the update.

rawfiner:
I have just added a configuration entry for this
The compute variance mode is shown only when the configuration string "plugins/darkroom/denoiseprofile/show_compute_variance_mode" is set to TRUE in darktablerc, or on images where the mode is set to MODE_VARIANCE

rawfiner:
I previously left a scaling of the force parameter accordingly to the zoom level. As this variance check mode aims to be as reliable as possible, I removed this scaling: better have the real value at a particular zoom level than an estimate more or less (in)accurate.

src/iop/denoiseprofile.c
```diff
@@ -3103,6 +3344,11 @@ void gui_init(dt_iop_module_t *self)
   dt_bauhaus_widget_set_label(g->strength, NULL, _("strength"));
   dt_bauhaus_combobox_add(g->mode, _("non-local means"));
   dt_bauhaus_combobox_add(g->mode, _("wavelets"));
+  gboolean compute_variance = dt_conf_get_bool("plugins/darkroom/denoiseprofile/show_compute_variance_mode");
```
TurboGit: const ...

And you need an entry data/darktableconfig.xml.in as others hidden prefs, this at least serve as documentation.

rawfiner: Thank you for the review @TurboGit :-)
I have now made these changes

aurelienpierre:
Welcome quality control in darktable ! \o/

PR #2289, opened 2019-03-26
blend: just reformating

rabauke:
Code was reformated via the new clang-format support in Clion, not by using clang-format directly. I hope the IDE applied the right rules, i.e. the ones in `.clang-format`. But results look reasonable to me.


TurboGit:
Looks good to me. But let's do that for all files after the UI revamp from Aurélien is merged otherwise this will create lot of conflicts I suppose.

rabauke:
Ok, no problem. I will close this pull request and reapply clang-format later when Aurélien's work has been merged.

PR #2292, opened 2019-03-26
duplicate : keep thumb in sync with current preview

AlicVB:
Now the thumb of the current version in the lib is keep in sync with the main view.

### DIFF
--- a/src/libs/duplicate.c
+++ b/src/libs/duplicate.c
@@ -52,7 +52,6 @@ typedef struct dt_lib_duplicate_t
   uint8_t *rgbbuf;
   int buf_mip;
   int buf_timestamp;
-
 } dt_lib_duplicate_t;
 
 const char *name(dt_lib_module_t *self)
@@ -272,6 +271,7 @@ static gboolean _lib_duplicate_thumb_draw_callback (GtkWidget *widget, cairo_t *
   params.zoom = 5;
   params.full_preview = TRUE;
 
+  int lk = 0;
   // if this is the actual thumb, we want to use the preview pipe
   if(imgid == dev->image_storage.id)
   {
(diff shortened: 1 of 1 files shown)


src/libs/duplicate.c
```diff
@@ -259,6 +270,60 @@ static gboolean _lib_duplicate_thumb_draw_callback (GtkWidget *widget, cairo_t *
...
+                                                         d->buf_height, stride);
+      }
+    }
+    params.full_surface = &(d->surface);
+    params.full_rgbbuf = &(d->rgbbuf);
+    params.full_surface_mip = &(d->buf_mip);
+    params.full_surface_id = &imgid;
+    params.full_surface_wd = &d->buf_width;
+    params.full_surface_ht = &d->buf_height;
+    int lk = 0;
```
TurboGit: There is an issue here. You take the address of lk just below...

src/libs/duplicate.c
```diff
@@ -259,6 +270,60 @@ static gboolean _lib_duplicate_thumb_draw_callback (GtkWidget *widget, cairo_t *
...
+    params.full_surface_w_lock = &lk;
+  }
+
   dt_view_image_expose(&params);
```
TurboGit: And use it there but lk is not anymore on the stack as the block is existed.

src/libs/duplicate.c
```diff
@@ -45,6 +45,14 @@ typedef struct dt_lib_duplicate_t
...
   int imgid;
   dt_lib_duplicate_select_t select;
+
+  int32_t buf_width;
+  int32_t buf_height;
+  cairo_surface_t *surface;
+  uint8_t *rgbbuf;
+  int buf_mip;
+  int buf_timestamp;
+
```
TurboGit: you can also remove this empty line here.

TurboGit:
Looks good now! Thanks.

PR #2293, opened 2019-03-26
expose : scroll shouldn't change filemanager zoom

### LINKED ISSUE #2279: Mouse wheel changes image size while in expose mode
When in expose mode, if the mouse wheel is used over the image nothing happens, but when changing to file manager view the size of the thumbs has changed.

It will be great if the mouse wheel could offset the selecting in expose mode, let's say I have images 1 and 2 selected, by using the wheel it offset to 2, 3; 3, 4; etc.

If that's not possible at least the mouse wheel shouldn't be propagated to the file manager view.

AlicVB:
this fix #2279


TurboGit:
Right, just perfect thanks.

PR #2294, opened 2019-03-26
fix buffer overflow

rabauke:
The current implementation of the function `dt_show_times` has two issues. It does not ensure that the `vsnprintf` function writes beyond the end of the buffer `buf` if it was already exhausted by the previous call to `snprintf`. Note that the man page states

The functions snprintf() and vsnprintf() do not write  more  than  size bytes  (including the terminating null byte ('\0')).  If the output was truncated due to this limit, then the return value  is  **the  number  of characters  (excluding the terminating null byte) which would have been written to the final string if enough space had been available.**   Thus, a  return  value  of  size or more means that the output was truncated.

In this case `sizeof buf - i` becomes a large unsigned integer.

Furthermore, `dt_show_times` uses its last non-variadic parameter to modify the function's behavior for the case that it is a NULL pointer. I think it is better to employ two different functions rather than this signaling parameter. Considering the declaration

    void dt_show_times(const dt_times_t *start, const char *prefix,
        const char *suffix, ...) __attribute__((format(printf, 3, 4)));

it is surprising that gcc and clang accept the function's current usage (E.g., icc does not.) because

    printf(NULL);

is not a legal usage of the `printf` function.

### DIFF
--- a/src/common/darktable.c
+++ b/src/common/darktable.c
@@ -1178,24 +1178,35 @@ void dt_free_align(void *mem)
 }
 #endif
 
-void dt_show_times(const dt_times_t *start, const char *prefix, const char *suffix, ...)
+void dt_show_times(const dt_times_t *start, const char *prefix)
 {
-  dt_times_t end;
-  char buf[160]; /* Arbitrary size, should be lots big enough for everything used in DT */
-  int i;
-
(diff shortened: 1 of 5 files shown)


TurboGit:
Looks good to me. Thanks.

PR #2295, opened 2019-03-26
update picker in point mode while dragging

edgardoh:
This restores the "update picker in point mode while dragging" functionality.

Also fixes a random crash when using the full preview.


TurboGit:
Looks good, thanks.

PR #2298, opened 2019-03-27
add skip on conflict export option

edgardoh:
A new option 'skip' is added to the 'on conflic' setting on the export module that skips the export if the file already exists.


TurboGit:
Looks good to me. Thanks.

PR #2299, opened 2019-03-27
undo: fix meaningless casts

rabauke:
This pull request fixes some casting issues. The undo facility has to deal with data of different types for which `dt_undo_data_t` is introduced and the undo API utilizes arguments of type `dt_undo_data_t *`. `dt_undo_data_t` is defined as

    typedef void *dt_undo_data_t;

Consequently, `dt_undo_data_t *` is a pointer to a void-pointer. But what we actually need here is a generic pointer,i.e., a void-pointer. 

Furthermore, to get a pointer to a function the address-of operator can be omitted. It is odd that the current code sometimes utilizes the `&` operator and sometimes not. Thus, I have removed the `&` where not required.

### DIFF
--- a/src/common/colorlabels.c
+++ b/src/common/colorlabels.c
@@ -38,7 +38,7 @@ typedef struct dt_undo_colorlabels_t
   uint8_t after;
 } dt_undo_colorlabels_t;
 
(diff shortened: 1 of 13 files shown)


TurboGit:
Looks good! Nice clean-up.

PR #2302, opened 2019-03-27
dt_history_get_items: order by num as in previous version.

### LINKED ISSUE #2300: Copy history stack, lighttable
**Describe the bug**
- in dt. 2.6.x and also the former official versions, the copy of history stack in lighttable had the order of the history stack I see in darkroom
- in the current dt git-master (2.7.0+796~g8a5ce514f) the order is different in that pop-up window
- I know has nothing to do with the actual sequential working order of those modules, when it comes to pixel-pipe
- However, inside that copy-window in lighttable it is challenging to disable basecurve, sharpen and rotation as you have to dig them out of the list somewhere, where they previously where at the beginning (means bottom) of the list.
- My point: handling of that “which elements to copy in lighttable from one pic to another” is less convenient now

**To Reproduce**
1. Go to 'lighttable'
2. Click on 'copy history stack'
3. See the sequence is not anymore as in 2.6.2 and earlier officials (and maybe git-master)

**Expected behavior**
I would be very happy to see it in the previous behaviour, as it makes it more easier to find the elements one want to alter

 - OS: [e.g.gentoo linux]
 - Version [e.g. 2.7 git-master several versions up to 2.7.0+816~g5e8d065a7

TurboGit:
This order corresponds to the history order as seen by the user
and is more natural to find some specific items in the list.

Fixes #2300

### DIFF
--- a/src/common/history.c
+++ b/src/common/history.c
@@ -756,7 +756,8 @@ GList *dt_history_get_items(int32_t imgid, gboolean enabled)
   DT_DEBUG_SQLITE3_PREPARE_V2(dt_database_get(darktable.db),
                               "SELECT num, operation, enabled, multi_name FROM main.history WHERE imgid=?1 AND "
                               "num IN (SELECT MAX(num) FROM main.history hst2 WHERE hst2.imgid=?1 AND "
-                              "hst2.operation=main.history.operation GROUP BY multi_priority) ORDER BY iop_order DESC",
+                              "hst2.operation=main.history.operation GROUP BY multi_priority) "
+                              "ORDER BY num ASC",
                               -1, &stmt, NULL);
   DT_DEBUG_SQLITE3_BIND_INT(stmt, 1, imgid);
   while(sqlite3_step(stmt) == SQLITE_ROW)


TurboGit:
@edgardoh : can you review? Will this be compatible with the iop_order ? Thanks.

edgardoh:
This will impact the order in which the modules are paste onto the dest
image (the dialog fills a list with the history entries in the order they
are displayed). I can't remember right now if the order is important in the
paste operation, I have checked the code and doesn't seem to be, but there
are a lot of scenarios and with the multi-instance and module renaming
things are tricky.

I do remember that this was more convenient to me in order to check if
things have been done as they should, the displayed order is the pipe, and
is easier to check the pipe than the history.

If this is really more intuitive for the user I say let's give it a try, if
there's any issues we can always decouple the display from the process.

junkyardsparkle:
While this dialog is being worked on, could  the labels also be made sensitive for checking/unchecking, consistent with how the checkboxes in modules work? It would be nice to not have to aim precisely at the boxes when there are many to check/uncheck.

TurboGit:
@edgardoh ; thanks, merging then.

AxelG-DE:
Currently that list is upside down compared to previous versions and also compared to history stack. 

It is better now than unsorted. However, same orientation than history stack would be helpful, I guess

PR #2304, opened 2019-03-28
display a confirm message on export

edgardoh:
This PR display a confirm message when exporting to disk on overwrite mode.

This way user settings are not overwritten and it get consistent with other potentially dangerous operations.


TurboGit:
Looks good to me! Thanks.

PR #2309, opened 2019-03-29
lighttable : offsets & selection enhancements

AlicVB:
- reset offset when changing collection in order to see first images of the collection.
- don't reset offset when switching back to filemanager from expose layout

If you enter full preview mode with exactly 1 select image, and this image is the one you display (your mouse is over it) : 
- the selection will follow the image shown in full preview : if you move to the next image, this image become the new selection.
- when you switch back to filemanager, scroll the view in order to ensure that the new selected image is visible.

I hope that this will not break anything in other users workflow...

### DIFF
--- a/src/views/lighttable.c
+++ b/src/views/lighttable.c
@@ -241,13 +241,14 @@ static void check_layout(dt_view_t *self)
 {
   dt_library_t *lib = (dt_library_t *)self->data;
   const dt_lighttable_layout_t layout = get_layout();
+  const dt_lighttable_layout_t layout_old = lib->current_layout;
 
   if(lib->current_layout == layout) return;
   lib->current_layout = layout;
 
   if(layout == DT_LIGHTTABLE_LAYOUT_FILEMANAGER)
   {
-    if(lib->first_visible_zoomable >= 0)
+    if(lib->first_visible_zoomable >= 0 && layout_old == DT_LIGHTTABLE_LAYOUT_ZOOMABLE)
     {
       lib->first_visible_filemanager = lib->offset = lib->first_visible_zoomable;
     }
@@ -389,6 +390,15 @@ static void _view_lighttable_collection_listener_callback(gpointer instance, gpo
   lib->force_expose_all = TRUE;
   _unregister_custom_image_order_drag_n_drop(self);
   _register_custom_image_order_drag_n_drop(self);
+
+  // in filemanager, we want to reset the offset to the beggining
+  const dt_lighttable_layout_t layout = get_layout();
+  if(layout == lib->current_layout && layout == DT_LIGHTTABLE_LAYOUT_FILEMANAGER && lib->offset > 0
+     && lib->first_visible_filemanager > 0)
+  {
+    lib->offset = lib->first_visible_filemanager = 0;
+    lib->offset_changed = TRUE;
+  }
 
   _full_preview_destroy(self);
   lib->full_zoom = 1.0f;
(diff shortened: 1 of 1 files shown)


src/views/lighttable.c
```diff
@@ -2682,6 +2695,32 @@ static void _stop_audio(dt_library_t *lib)
   lib->audio_player_id = -1;
 }
 
+static void _ensure_image_visibility(dt_library_t *lib, uint32_t rowid)
+{
+  if(get_layout() != DT_LIGHTTABLE_LAYOUT_FILEMANAGER) return;
+
+  // if we are before the firts visible image, we move back
```
TurboGit: firts -> first

src/views/lighttable.c
```diff
@@ -2682,6 +2695,32 @@ static void _stop_audio(dt_library_t *lib)
...
+  int offset = lib->offset;
+  while(offset > rowid)
+  {
+    offset -= lib->images_in_row;
+  }
+
+  // Are we after the last fuly visible image ?
```
TurboGit: fuly -> fully

TurboGit:
Apart from the two very minor comments all good to me. Thanks.

AlicVB:
Thanks... seems like that I suffer from "keyboard dyslexia" ;)

TurboGit:
Thanks! Now merged.

src/views/lighttable.c
```diff
@@ -390,6 +393,15 @@ static void _view_lighttable_collection_listener_callback(gpointer instance, gpo
   _unregister_custom_image_order_drag_n_drop(self);
   _register_custom_image_order_drag_n_drop(self);
 
+  // in filemanager, we want to reset the offset to the beggining
```
edgardoh: This seems to be causing #2320, I also have it when editing an image and go back to lighttable.

AlicVB: I can't reproduce, neither with copy-pasting history or with switching back to lt from dr... Do you have something special in your collect queries ? or filters ?
Because this "reset" happens only when the collection is changed, and I don't see why this signal would happen in this cases...

edgardoh:
I did some more tests, it seems to happen only when the image being
modified was an un-edited one, if the image already has edits things are
normal. But I can't find why is this, maybe some event when the state
changes from un-edited to edited?

edgardoh:
OK, is the dt_tag_attach(tagid, dest_imgid) in
dt_history_copy_and_paste_on_image(), but I'm not sure what to do with that.

edgardoh:
The dt_collection_update_query() is called and triggers the event, it shouldn't unless the collection actually changes, but I don't know how to check for that.

About this new feature, I'm not sure about it, if I'm on the file manager and I set the stars to zero in an image I would like the offset to remain the same, now it jumps to the top and I have to scroll down each time to get where I was.
It will be nice if this only happens when the user changes to a new collection, not when the current collection is updated.

AlicVB:
Yes, and this answer your comment on your other PR too : We may need to have 2 events :+1: - one for 1- query update : same query run another time because the result can be different (somenthing in 1 image change)

edgardoh:
> Yes, and this answer your comment on your other PR too : We may need to
> have 2 events 👍 - one for 1- query update : same query run another time
> because the result can be different (somenthing in 1 image change)
>
Just to clarify, do you want me to do it or are you going to handle
yourself?

AlicVB:
I'll have some time to work on that tomorrow, if you want.

edgardoh:
Great, thanks!

PR #2316, opened 2019-03-31
change paste history order

edgardoh:
Changes the order of copied/pasted history items.


edgardoh:
Also, for some reason, when pasting history the file manager jumps to the top, no idea what's going on there.

TurboGit:
Looks good to me! I'll have a look at the file manager jumps I

TurboGit:
But I'll need to reproduce... I don't see a jump... What is the steps to reproduce? Thanks.

edgardoh:
Right now it happens when I paste the history on some image or when I
return to lighttable from the darkroom. It happens all the time, no special
scenario.

PR #2319, opened 2019-03-31
add a passthrough mode to colorin and colorout

edgardoh:
This add a passthrough mode to colorin and colorout and also changes the len of the icc profiles to match colorspaces.h definition.

The idea is to have both modules with this setting and no other module after colorin enabled, if not some error/warning messages can be displayed. This is for advanced users only, so I won't do anything about it.

The intended use is to be able to export the image in camera rgb for the denoise profiles generation.

### DIFF
--- a/data/darktableconfig.xml.in
+++ b/data/darktableconfig.xml.in
@@ -1349,6 +1349,13 @@
     <longdescription/>
   </dtconfig>
   <dtconfig>
+    <name>darkroom/ui/iccprofile_passthrough_visible</name>
+    <type>bool</type>
+    <default>false</default>
+    <shortdescription>whether to show the passthrough option in colorin and colorout</shortdescription>
+    <longdescription/>
+  </dtconfig>
+  <dtconfig>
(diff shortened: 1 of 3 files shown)


edgardoh:
@rawfiner , be aware that this changes the version on both colorin and colorout, it can mess with previous edits.

If we decide to keep it I'll show/hide the setting based on a hidden config variable, if not I'll remove it and keep only the change in profile len.

rawfiner:
Thanks a lot @edgardoh

edgardoh:
@TurboGit , I just noticed that the profiles are storing the full path now, this seems like a limitation, if I want to use some edit on another config dir dt won't find the icc profile, even if the file is in the propper in/out directory.
Sorry if this as already been discussed, I just couldn't find the thread.

edgardoh:
This is ready now.

Not sure what's best, merge this or port it to @rawfiner  PR. Is really the same to me.

rawfiner:
It is the same for me as well.
@TurboGit , do you have any preference?

PR #2321, opened 2019-03-31
new lighttable mode: culling

edgardoh:
This new mode is very similar to the 'expose' mode, but it display a fixed number of consecutive images, starting from the first selected image. It can be navigated with the mouse wheel and keyboard and the number of displayed images can be set with an entry at the bottom.

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -158,6 +158,11 @@ when updating from the currently stable 2.4.x series, please bear in mind that y
 - A new option 'skip' is added to the 'on conflic' setting on the export module
   that skips the export if the file already exists.
 
+- A new lighttable mode 'culling' is added. It displays a fixed number of consecutive images,
+  starting from the first selected image and allows to pan & zoom them.
+  It can be navigated with the mouse wheel and keyboard and the number of
+  displayed images can be set with an entry at the bottom.
+
 ## Bug fixes
 
 - The color picker support has been fixed by a complete rewrite. It
--- a/src/libs/tools/lighttable.c
+++ b/src/libs/tools/lighttable.c
@@ -18,6 +18,7 @@
 
 #include <gdk/gdkkeysyms.h>
 
+#include "common/collection.h"
 #include "common/selection.h"
 #include "control/conf.h"
 #include "control/control.h"
@@ -35,6 +36,9 @@ typedef struct dt_lib_tool_lighttable_t
   GtkWidget *layout_combo;
   dt_lighttable_layout_t layout, previous_layout;
   int current_zoom;
+  GtkWidget *display_num_images; // slider & entry for number of images to be displayed in culling mode
+  GtkWidget *display_num_images_entry;
+  int current_display_num_images; // number of images to be displayed in culling mode
 } dt_lib_tool_lighttable_t;
 
 /* set zoom proxy function */
@@ -45,6 +49,9 @@ static gint _lib_lighttable_get_zoom(dt_lib_module_t *self);
 static dt_lighttable_layout_t _lib_lighttable_get_layout(dt_lib_module_t *self);
 static void _lib_lighttable_set_layout(dt_lib_module_t *self, dt_lighttable_layout_t layout);
 
+static void _lib_lighttable_set_display_num_images(dt_lib_module_t *self, const int display_num_images);
+static int _lib_lighttable_get_display_num_images(dt_lib_module_t *self);
+
 /* lightable layout changed */
 static void _lib_lighttable_layout_changed(GtkComboBox *widget, gpointer user_data);
 /* zoom slider change callback */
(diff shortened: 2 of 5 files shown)


edgardoh:
@AlicVB , I'm having the same issue here with the offset, when the user changes to a new collection I would like to display the first image, but there's no way to do it right now.
So, how about adding a new event "new collection activated" where we can place specific code to handle that?
I can do it here, or if you don't want to wait you can add it and I'll update this when ready.

Nilvus:
I like the idea but if it's so similar to expose, why not just add this possibility to expose mode with, for example a switch to one side set fixed images (your actual idea) and one the other side the actual mode ? I don't know if it easier to add this way but we have to care about having too much views, especially if two of them are very close together.

AlicVB:
I agree with @Nilvus : Why not using the expose mode, detect if the image are consecutive or not. If yes, you allow to "navigate", if not you block it (and maybe display some message). And if you want to display more/less images you can always use the filmstrip to select/deselect images...
(That said, I've not the same need as you, so my proposal is maybe not very practical...)

TurboGit:
I have tested this, I like it. Really nice.

As @Nilvus and @AlicVB I think we should have a single mode and a switch on the bar to activate or not the culling mode. All this is just a GUI rework as the mode is working fine.

TurboGit:
So maybe the best proposal is AlicVB one? If you have selected a set of X consecutive images you can navigate forward/backward? Would that work for you? I'm not sure as I didn't used this a lot...

edgardoh:
@TurboGit , I really think this should be a different mode, it doesn't feel right to me to have a setting to switch between them because I think of it as different modes, and we already have a combo box for that.

Maybe is on me because I opened this by saying that this is very similar to the expose, but what I meant is that, as with the expose, images are displayed side-by-side and it allows to pan & zoom, but that's it. Take it like I say that both the file manager and zoomable lighttable display the thumbs one next to the other.

TurboGit:
Ok, let's keep that as a new mode for now.

I would suggest to add a key accel (as for exposé mode) in lib/tools/lighttable.c

Maybe using "ctrl-x" as the default for exposé is "x" ?

edgardoh:
I was thinking about that and I would like something more like the alt+tab
for the windows, but it seems that it cannot be done with the shortcuts (or
I don't know how to do it), but maybe we can have two shortcuts, one that
cycles between the views forward and another one backwards, we re-use the
existing one for expose and add the backwards.
This is just a thought, maybe is just better to have one for the culling
now.

TurboGit:
Well having an accel close to the exposé one will be easy to remember. Since both mode are very close and related to me ctrl-x seems good. We can always change the default later after using this new mode. And anyway users will always be able to change it.

edgardoh:
OK, I'll add a new one for the culling. I'll wait until the fix offset jump
is merged so I can rebase and use the new event too.

TurboGit:
@edgardoh : thanks, offset jump fix is now merged.

edgardoh:
Rebased & added shortcut.

TurboGit:
One issue, when you click on the filmstrip the images are not updated until the mouse cursor is moved in the center area. Looks like a missing redraw?

edgardoh:
Exactly that, fixed!

edgardoh:
Sorry, not yet, I found something else, I'll let you know.

TurboGit:
Another behavior that we could improve, the scenario:
- select two consecutive images
- press "x" to select exposé to compare them
- then press "ctrl-x" to select culling mode and advance with mouse scroll
- you found two other very similar images
- press "x" to select exposé again

=> only one image is selected

I think that in this case (moving from culling to exposé) we want to select two (or whatever the number of images was selected in culling mode) consecutive images.

edgardoh:
That may be tricky, in culling I select one image (the first) no matter how
many images are displayed. I'll see if I can find a way.

edgardoh:
It wasn't that tricky after all.
This works only if there's one image selected, the user can select multiple images from the filmstrip, in that case I respect that and do not change it.

TurboGit:
Works fine now! Really nice, keeping the selection really makes the experience more natural.Thanks.

junkyardsparkle:
Good timing, I have a lot of culling to do :)  ...a few comments:

It would be nice to be able to use the same shortcuts used in darkroom view for navigating the image list (IIRC "space" and "backspace" by default, I use Page Up/Down).

My usual culling method is to use the "r" shortcut while navigating through images in darkroom view, eventually reviewing using "rejected only" in filemanager mode, then select all > delete. I found that using "r" key in this mode had the expected effect of moving to the next image (as in darkroom view) but that the filmstrip showed the wrong thumbnail being removed. Hovering over the filmstrip causes the thumbs to be redrawn correctly.

When reviewing using "rejected only", it would be nice if hitting "r" toggled the rejected state off again, as it does in darkroom view. (I just checked filemanager mode, and this doesn't work there, either, although I thought it did before? May be a false memory.)

One additional behavior that I've always thought could be improved: if rejecting the last item in the filmstrip, move to previous item rather than the first. Currently culling mode is consistent with darkroom view, but both would be better with this change IMHO.

It will be interesting to see what other optimizations can be made to a dedicated culling mode... thanks for the work.

pixmin:
Whoever worked on this: thank you so much!!! After removing most unwanted shots with Geeqie I can now compare several similar (or not) photos with the this culling mode in Darktable, so so good! Thank you!

PR #2326, opened 2019-04-02
lighttable : fix drag-n-drop and zoom center

### LINKED ISSUE #2305: Drag zoomed image display thumb icon
When dragging a zoomed image in file manager it drags it a bit and then it displays the drag icon for the thumb.
It works fine on expose mode.

In both file manage & expose the zoom don't follow the mouse when in full preview or when comparing images.

AlicVB:
This PR should fix #2305 : 
- The bug that doesn't deactivate drag_n_drop in full preview mode
- The feature to center the zoom at pointer position in full preview (deactivated in expose mode as it's way more complex)

### DIFF
--- a/src/views/lighttable.c
+++ b/src/views/lighttable.c
@@ -110,6 +110,10 @@ typedef struct dt_preview_surface_t
   cairo_surface_t *surface;
   uint8_t *rgbbuf;
   int w_lock;
+
+  float w_fit;
+  float h_fit;
+  float zoom_100;
 } dt_preview_surface_t;
 
 /**
@@ -296,6 +300,10 @@ static void _full_preview_destroy(dt_view_t *self)
     lib->fp_surf[i].height = 0;
     lib->fp_surf[i].imgid = -1;
     lib->fp_surf[i].w_lock = 0;
+
+    lib->fp_surf[i].zoom_100 = 10.0f;
+    lib->fp_surf[i].w_fit = 0.0f;
+    lib->fp_surf[i].h_fit = 0.0f;
   }
 }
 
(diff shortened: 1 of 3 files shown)


TurboGit:
Thanks.

edgardoh:
It seems that I'm not able to zoom as much as before, is that intended?

AlicVB:
Normally, the max zoom is at 100% (the real image size), maybe check with another viewer to compare (I've just checked, and it's ok here). Note that there's a small jump the first time you reach the max zoom. It's due to the fact that we don't know what is the max output size of the image before computing the last mipmap...

edgardoh:
I did some debug and here I get a 2.5:

https://github.com/darktable-org/darktable/blob/master/src/views/lighttable.c#L2568

seems that the lib->fp_surf[i].zoom_100 is modified by the view manager

AlicVB:
Oh ok : indeed I've added this to ensure that zoom levels stay the same across all pictures in expose mode. If you have image with different sizes, that means that the max zoom is reach when the smallest image reach 100%... Is that your case ?

TurboGit:
I have seen that indeed. I don't think we want to go above 100% so the only solution would be to MAX the zoom at 100% for each individual pictures. That is, if one is lot smaller it will stays at 100% while the others are zoomed. Would that work?

edgardoh:
Not on my tests, all images are from the same camera, same size,
orientation, etc.

edgardoh:
That will be OK with me, but its not what is happening, I can barely zoom
when displaying same size images.

AlicVB:
@TurboGit : not straightforward to code, but I'll have a look
@edgardoh : so there's a bug somewhere. I'll do some more tests on my side to try to reproduce... Just to be sure : does it happen in full preview too ?

edgardoh:
Full preview is working fine on all view modes, the issue is in regular
preview on expose and culling.

PR #2331, opened 2019-04-04
Expose zoom fix

AlicVB:
This fix an error which prevent zooming in some cases because of rounding errors in mipmap computation.
This fix also a bug which can cause multiple full-res image computation because of the cache system max memory. This one should avoid grey/black thumbnails on quick zooming to the full size, especially in expose mode


TurboGit:
I see that you now block zooming when a 100% zoom is reached for one picture.

Looks ok, thanks.

AlicVB:
yes, the other idea (block zooming at 100% for each pictures) as discussed earlier is still planned. But I first wanted to fix that ;)

TurboGit:
That will be perfect! Thanks Aldric.

PR #2333, opened 2019-04-04
Lighttable panels + full preview filmstrip

AlicVB:
- Store specific panel states per layout, with a special mode for preview. This allow for example to hide lateral bars specifically for expose, etc...
- Filmstrip is enabled in full preview mode. It's hidden by default, but you can show the bottom panel to see it (and dt will remember it's state with the previous point)
- (All views) : clicking on the arrow of the bottom panel now change states like this : all visible => toolbar hidden => all hidden => all visible => ... This allow finer tuning of this 2 panels states

### DIFF
--- a/data/darktableconfig.xml.in
+++ b/data/darktableconfig.xml.in
@@ -1160,6 +1160,48 @@
     <shortdescription>scroll to lighttable modules when expanded/collapsed</shortdescription>
     <longdescription>when this option is enabled then darktable will try to scroll the module to the top of the visible list</longdescription>
   </dtconfig>
+  <dtconfig>
+    <name>lighttable/ui/preview/bottom_visible</name>
+    <type>bool</type>
+    <default>false</default>
+    <shortdescription>show bottom panel in preview mode</shortdescription>
+    <longdescription/>
+  </dtconfig>
+  <dtconfig>
+    <name>lighttable/ui/preview/left_visible</name>
+    <type>bool</type>
+    <default>false</default>
+    <shortdescription>show left panel in preview mode</shortdescription>
+    <longdescription/>
+  </dtconfig>
+  <dtconfig>
+    <name>lighttable/ui/preview/right_visible</name>
+    <type>bool</type>
+    <default>false</default>
+    <shortdescription>show right panel in preview mode</shortdescription>
+    <longdescription/>
+  </dtconfig>
+  <dtconfig>
+    <name>lighttable/ui/preview/header_visible</name>
+    <type>bool</type>
+    <default>false</default>
+    <shortdescription>show header panel in preview mode</shortdescription>
+    <longdescription/>
+  </dtconfig>
+  <dtconfig>
+    <name>lighttable/ui/preview/toolbar_bottom_visible</name>
+    <type>bool</type>
+    <default>false</default>
+    <shortdescription>show toolbar bottom panel in preview mode</shortdescription>
+    <longdescription/>
+  </dtconfig>
+  <dtconfig>
+    <name>lighttable/ui/preview/toolbar_top_visible</name>
+    <type>bool</type>
+    <default>false</default>
+    <shortdescription>show toolbar top panel in preview mode</shortdescription>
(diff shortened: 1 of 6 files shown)


TurboGit:
> (All views) : clicking on the arrow of the bottom panel now change states like this : all visible => toolbar hidden => all hidden => all visible => ... This allow finer tuning of this 2 panels states

Can you tell me more about this? I don't get what it is doing exactly... Thanks.

edgardoh:
When you talk about full preview mode, is: shortcuts->views->lighttable->preview? Or maybe there's other full preview I don't know about...

AlicVB:
currently, when you click on the arrow, you show/hide the toolbar panel (in lighttable, this is the one with the layout combo) and the bottom panel (filmstrip/timeline) all together. so when you click multiple times, it does "all visible => all hidden => all visible..."
With my change, it does "all visible => toolbar hidden => all hidden => all visible => ..."

there's just the "toolbar but not bottom panel" case which is not reachable. If needed, I can add this state in the "click sequence" (I hope it's more clear now)

AlicVB:
@edgardoh : yes, the one to show the image in full screen. Note that filmstrip is specifically useful with "sticky" preview

edgardoh:
OK, I just wanted to be sure we are talking about the same thing.

I don't use the preview, I have to keep pressing the key and I don't find that useful, not sure how other people use it (or what for)
With the sticky preview the "remember panels state" is not working for me, it displays the state for the last view mode used no matter the active view.

AlicVB:
oh ? that's strange... can you you share your darktablerc file (in fact just the lines with lighttable/ui/preview/) How do you access full preview exactly ? you hover an image in filemanager and you press the key mapped to sticky preview ?
(great that you are here, to always find bugs I can't reproduce ;) )

edgardoh:
Again is only me? This starts to be annoying...

I don't have any lines with lighttable/ui/preview in darktablerc

In culling, I press the shotcut for sticky preview
with the mouse I expand left panel
exit sticky preview with the same shortcut
change to expose
enter sticky preview
left panel is displayed and I hide it with the mouse
exit sticky preview
change to culling
enter sticky preview
panel is hidden

AlicVB:
Well, if I understand correctly, that's the intended behavior...
Panels states are saved for each layout independently, and we count here full preview as if it was a layout. So we save panels states for : filemanager, zooming lighttable, expose, culling, full preview
In your test, that's exactly the case : left panel state is saved and restore specifically for full preview...
In fact, in full preview (sticky), expending panels is quite useless imho, except to show and use the filmstrip.

(oh, and it's not annoying for us, your "it affect only me" syndrome is really useful to find bugs ! )

edgardoh:
OK, I didn't understand then, I thought it was saved by view mode
regardless of the full preview status, so is working fine then.

And sure, the left panel was just to test it, if anything I'll use the
filmstrip (bottom panel)

TurboGit:
Not sure if this is due to this commit but I think it would be nice to fix the following behavior:

- on lighttable
- click "x" to go to exposé
- click "x" again you are back to lighttable

Fine, this is really what is expected. Now:

- on lighttable
- click "x" to go to exposé
- click "ctrl-x" to go to culling
- click "ctrl-x" and you're back to exposé

You don't have a mean to go back to lighttable filemanager and this is quite annoying when you have hidden the combo to change the view.

What do you think?

TurboGit:
Sorry I miss to put my proposal :)

I would always return to filemanager when hitting a key in the mode we are already. That is "x" in exposé mode should always returns to filemanager, ctrl-x in culling mode should always returns in filemanager.

AlicVB:
No, I don't think it's due to this changes, but I tend to have the same workflow : filemanager as primary mode and expose/culling as a temporary mode to help me sort.
I'll try to implement what you propose (do you prefer I add that to this PR, or create a separate one ?)

TurboGit:
In another PR then. I'll merge this one since it is not a regression due to this PR. Thanks.

PR #2334, opened 2019-04-04
culling: make filmstrip follow full preview

edgardoh:
This makes the sticky preview to update the filmstrip and vice-versa. Also uses the up & down shortcuts.

It removes the exit sticky preview shortcut, there's no need for it as the actual shortcut toggles it.


TurboGit:
Nice! Thanks again for all your work!

PR #2337, opened 2019-04-05
guided filter: reduce consumption of temporary memory

rabauke:
The performance of the guided filter does not depend of the size of the filter kernel (as long as the image is much larger than the kernel). This beneficial characteristics of the implementation of the guided filter in darktable is realized by utilizing a rather large amount of temporary storage. The memory foot print, however, can be reduced by about 20% by a minor modification. Some temporary memory can be used twice for different purposes, see the diff.


TurboGit:
Looks good thanks!

PR #2338, opened 2019-04-05
lighttable : enhance full preview zoom memory

AlicVB:
- This should limit the flickering/black image problem when zooming in expose/culling layout.
- This should avoid the zoom "lock" when zooming quickly

Technically the problem is that when we have for example 4 images at 5000x3000, this is about 240M in memory, just for the MIPMAP_8 level, and the cache by default (at least here) is set to 256M so we get quickly out of memory and dt drop images from the cache, which in turn provoke recomputing of this thumbnails...
To avoid that, I drop the cache of this mipmap once we have computed the corresponding surface (only when zooming is > 1.0f).
This seems to work fine.

But we still can fill the cache to easily and I wonder if increasing it's default value would make sense ? (and how to do it, because I've not understand how it is set, darketableconfig.xml.in has very strange value...)

### DIFF
--- a/src/common/mipmap_cache.c
+++ b/src/common/mipmap_cache.c
@@ -1019,6 +1019,27 @@ void dt_mipmap_cache_remove(dt_mipmap_cache_t *cache, const uint32_t imgid)
     }
   }
 }
+void dt_mipmap_cache_remove_at_size(dt_mipmap_cache_t *cache, const uint32_t imgid, dt_mipmap_size_t mip)
+{
+  // get rid of the thumbnails:
+  const uint32_t key = get_key(imgid, mip);
+  dt_cache_entry_t *entry = dt_cache_testget(&_get_cache(cache, mip)->cache, key, 'w');
+  if(entry)
+  {
+    ASAN_UNPOISON_MEMORY_REGION(entry->data, dt_mipmap_buffer_dsc_size);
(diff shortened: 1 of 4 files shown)


src/common/mipmap_cache.c
```diff
@@ -1019,6 +1019,27 @@ void dt_mipmap_cache_remove(dt_mipmap_cache_t *cache, const uint32_t imgid)
...
   }
 }
+void dt_mipmap_cache_remove_at_size(dt_mipmap_cache_t *cache, const uint32_t imgid, dt_mipmap_size_t mip)
+{
+  // get rid of the thumbnails:
+  const uint32_t key = get_key(imgid, mip);
+  dt_cache_entry_t *entry = dt_cache_testget(&_get_cache(cache, mip)->cache, key, 'w');
+  if(entry)
+  {
+    ASAN_UNPOISON_MEMORY_REGION(entry->data, dt_mipmap_buffer_dsc_size);
```
TurboGit: Is this expected or a left over the debug session?

AlicVB: This new fct is just a copy/paste of the dt_mipmap_cache_remove fct for only 1 mip level. So I think it's needed...

TurboGit:
Ok, looks good! Thanks.

PR #2345, opened 2019-04-06
culling: keep zoom when selection changes

edgardoh:
This adds two changes:

-Keeps the zoom level when an image is rejected.
-Stays at the end of the collection if the last image is rejected.


TurboGit:
Thanks, looks good!

PR #2349, opened 2019-04-07
guided filter: OpenCL implementation

rabauke:
Currently, the OpenCL implementation of the guided filter just copies data to cpu memory and carries out the whole calculation by the CPU with an OpenMP code. This pull request is to contribute a genuine OpenCL implementation of the guided filter. In the case that the gpu memory might be not large enough to carry out the calculation on the gpu a cpu fall-back routine is still employed. The code was tested on a low-end Nvidia graphics card (GeForce GT 730M) only.

### DIFF
--- a/data/kernels/guided_filter.cl
+++ b/data/kernels/guided_filter.cl
@@ -1,0 +1,292 @@
+/*
+    This file is part of darktable,
+    copyright (c) 2019 Heiko Bauke
+
+    darktable is free software: you can redistribute it and/or modify
+    it under the terms of the GNU General Public License as published by
+    the Free Software Foundation, either version 3 of the License, or
+    (at your option) any later version.
+
+    darktable is distributed in the hope that it will be useful,
+    but WITHOUT ANY WARRANTY; without even the implied warranty of
+    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+    GNU General Public License for more details.
+
+    You should have received a copy of the GNU General Public License
+    along with darktable.  If not, see <http://www.gnu.org/licenses/>.
+*/
+
+#include "common.h"
+
+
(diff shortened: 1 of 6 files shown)


rabauke:
rawspeed changes in the pull request are not intended. Will be fixed asap.

rabauke:
Rawspeed sources have been removed from this pull request.

TurboGit:
I have tested this new version and I found some noticeable differences between the OpenCL and CPU version. The mask when displayed (in yellow) used to be very close or identical in both code path but now there is really some differences.

rabauke:
For the previous OpenCL implementation you have not found any difference between the OpenCL version and the cpu version because the OpenCL version just copied all data to main memory and called the cpu version. The output of the new (real) OpenCL implementation should of course give the same results. I was not able to find any differences in my tests.

@TurboGit How did you test the new implementation?

rabauke:
There was a minor error, which was accidentally not visible in the case of mask adjustment for modules that operate in Lab space and all my test had been carried out in Lab space. This issue has been fixed now.

TurboGit:
This does not work for me.

With the following png and xmp:

[masks.png.xmp.txt](https://github.com/darktable-org/darktable/files/3057583/masks.png.xmp.txt)
masks

Without OpenCL:

Capture d’écran de 2019-04-09 09-04-12

With OpenCL:

Capture d’écran de 2019-04-09 09-03-23

rabauke:
@TurboGit Before I submitted my revised version of my pull request I removed some debugging code. During this clean-up one line of code was removed that was actually part of my revision, which caused a run-time failure while calling an OpenCL kernel. Consequently, the guided filter was not working at all in the OpenCL code path. This has been fixed now. Thanks for testing.

TurboGit:
Perfect this now give the very same results on my test pictures with and without OpenCL.

Again great work! Thanks.

PR #2351, opened 2019-04-07
[WIP] implement detachable left/right panels

AlicVB:
Quick and basic detachable panel implementation. At first sight, it works but it certainly break lot of things...
About color management, left and right panels haven't a lot of critical colors. I've added an icon and a tooltip for the navigation and color picker to warn user that color can be false.
The filmstrip is not detachable, because here color management is critical and I don't know a way to handle that point across multiple windows/screens

### DIFF
--- a/data/darktableconfig.xml.in
+++ b/data/darktableconfig.xml.in
@@ -23,6 +23,20 @@
     <longdescription>(needs a restart)</longdescription>
   </dtconfig>
   <dtconfig prefs="gui">
+    <name>panels/detached_left</name>
+    <type>bool</type>
+    <default>false</default>
+    <shortdescription>detach the left panel (needs a restart)</shortdescription>
+    <longdescription>Be carefull : don't trust colors in this panel !</longdescription>
+  </dtconfig>
+  <dtconfig prefs="gui">
+    <name>panels/detached_right</name>
+    <type>bool</type>
+    <default>false</default>
+    <shortdescription>detach the right panel (needs a restart)</shortdescription>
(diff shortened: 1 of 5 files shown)


edgardoh:
I don't have the first clue about color management, and to prove it: the widgets need something special to be color managed or is only about the profile applied to the image?

TurboGit:
Not looked at this but I don't think we want detachable panels. I think for dt what we need is a second ful screen display on a second monitor. I don't think have panels on one side and display on the other is usable, each time you want to create/modify a mask for example you'll need to go back and forth between the primary and secondary screen. Not good to me.

AlicVB:
In my case, it's quite not a problem, as there's not space between my too screen, so it's just like if we have a "big" border between panel and main view. Anyway in this case, one can stay with my actual state : a window which expand on the second screen.

So feel free to close this PR, no pb ;) (at least we can remember that it's not difficult to implement)

About the idea you suggest, it would be awesome, but we'll have to handle the color management problem, and as @edgardoh I know quite nothing in this area !

edgardoh:
I finally managed to fix the "sharp" issue, so I'll create a PR with the second window.
As I see it, is better to have a first version not color managed than nothing, that can be added later, but of course it will have to be discussed.

TurboGit:
Given @edgardoh work on the second darkroom window I think that this can be closed now.

AxelG-DE:
@TurboGit @AlicVB

Pity
As much I like @edgardoh did 2ndpreview, it does not fit so much to my dreams :-) 

See my posts here:
https://github.com/darktable-org/darktable/pull/2365

PR #2358, opened 2019-04-08
culling: keep cache in memory

edgardoh:
This re-uses cache when in culling & expose mode.

### DIFF
--- a/src/views/lighttable.c
+++ b/src/views/lighttable.c
@@ -325,28 +325,33 @@ static void check_layout(dt_view_t *self)
   }
 }
 
+static inline void _destroy_preview_surface(dt_preview_surface_t *fp_surf)
+{
+  if(fp_surf->surface) cairo_surface_destroy(fp_surf->surface);
+  fp_surf->surface = NULL;
+  if(fp_surf->rgbbuf) free(fp_surf->rgbbuf);
+  fp_surf->rgbbuf = NULL;
+  fp_surf->mip = 0;
+  fp_surf->width = 0;
+  fp_surf->height = 0;
+  fp_surf->imgid = -1;
+  fp_surf->w_lock = 0;
+
+  fp_surf->zoom_100 = 40.0f;
+  fp_surf->w_fit = 0.0f;
+  fp_surf->h_fit = 0.0f;
+
+  fp_surf->max_dx = 0.0f;
+  fp_surf->max_dy = 0.0f;
+}
+
 static void _full_preview_destroy(dt_view_t *self)
 {
   dt_library_t *lib = (dt_library_t *)self->data;
 
   for(int i = 0; i < FULL_PREVIEW_IN_MEMORY_LIMIT; i++)
   {
-    if(lib->fp_surf[i].surface) cairo_surface_destroy(lib->fp_surf[i].surface);
-    lib->fp_surf[i].surface = NULL;
-    if(lib->fp_surf[i].rgbbuf) free(lib->fp_surf[i].rgbbuf);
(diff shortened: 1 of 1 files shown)


edgardoh:
@AlicVB , could you please review this?

-I'm doing it for both modes, maybe it should be done for culling only?
-I'm not sure how max_in_memory_images plays here, should I keep only the first max_in_memory_images from fp_surf?

AlicVB:
the code seems ok for me.
- you keep in memory unsed surfaces. that's intended ?
- max_in_memory is there to limit the number of image stored in buffer. If you go above, it should disable zooming (and disabled the used of fp_surf too) Your change in this area are great as they fix a leftover of my dev (mix of pref value and harcoded number)
- it's not useful in expose mode as change can only be made with filmstrip and that reinitialize the cache.

edgardoh:
> the code seems ok for me.
>
>    - you keep in memory unsed surfaces. that's intended ?
>
> Yes, when culling I display two images and is very common to go back and
forward one image to compare it, and the refresh is annoying, this way it
makes the experience more pleasant, but I don't know the downside of having
the cache full all the time (OK, it takes resources). I can keep only the
first max_in_memory_images and clear the rest after the sort, as long as
max_in_memory_images > 2 it will ok for me.

>
>    - max_in_memory is there to limit the number of image stored in
>    buffer. If you go above, it should disable zooming (and disabled the used
>    of fp_surf too) Your change in this area are great as they fix a leftover
>    of my dev (mix of pref value and harcoded number)
>    - it's not useful in expose mode as change can only be made with
>    filmstrip and that reinitialize the cache.
>
> In culling selecting with the filmstrip does not reset the cache, I don't
know why in the expose it does, but I can add an if() and sort only in
culling, your call here.

AlicVB:
> In culling selecting with the filmstrip does not reset the cache, I don't know why in the expose it does

because _view_lighttable_selection_listener_callback doesn't call same fct for culling and expose (and in the expose one, it destroy the cache)

> but I can add an if() and sort only in culling, your call here.

I would say add the if, it's easy to remove if we need it one day for expose too.

A side point, do you think we can put the call to _sort_preview_surface outside the expose fct ? just when we change the images ? it will speed up a little bit the expose fct which is called very often (at each mouse movement)

edgardoh:
> In culling selecting with the filmstrip does not reset the cache, I don't
> know why in the expose it does
>
> because _view_lighttable_selection_listener_callback doesn't call same fct
> for culling and expose (and in the expose one, it destroy the cache)
>
> but I can add an if() and sort only in culling, your call here.
>
> I would say add the if, it's easy to remove if we need it one day for
> expose too.
>
OK

> A side point, do you think we can put the call to _sort_preview_surface
> outside the expose fct ? just when we change the images ? it will speed up
> a little bit the expose fct which is called very often (at each mouse
> movement)
>
I can call it on the selection & collection callbacks, I need the list of
displayed images but that should be easy. Worst case scenario, we'll have
some unwanted refresh.

Not sure yet if you think that I should free the cache when is >
max_in_memory_images...

AlicVB:
> Not sure yet if you think that I should free the cache when is > max_in_memory_images...

forgot to answer this point, sorry... No I think we should never allocate fp_surf(x] if x > max_in_memory. If the user say "don't store more than ... in memory" we should respect it imho.

edgardoh:
Ready the if() == culling and keeping only max_in_memory_images entries.

I did not moved the call from expose_expose() because I'm afraid that it can produce some thread issues. This way is safer and it should be fast, there's only a few cache entries to process.

TurboGit:
Looks good to me! Thanks.

PR #2360, opened 2019-04-08
retouch : allow to change algorithm afterwards

AlicVB:
For stupids (like me) who realize that they have not choose the right algo after having drawn some complex shape...
Be careful that I'm not fully aware of all the subtleties of this iop (continuous mode, ...)

### DIFF
--- a/src/iop/retouch.c
+++ b/src/iop/retouch.c
@@ -2154,7 +2154,16 @@ static void rt_select_algorithm_callback(GtkToggleButton *togglebutton, dt_iop_m
 
   rt_show_hide_controls(self, g, p, g);
 
-  if(darktable.develop->form_gui->creation && (darktable.develop->form_gui->creation_module == self))
+  const int index = rt_get_selected_shape_index(p);
+  if(index >= 0)
+  {
+    if(p->algorithm != p->rt_forms[index].algorithm)
+    {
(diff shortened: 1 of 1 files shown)


edgardoh:
This has been requested a couple of times and I didn't want to do it, for me the problem is the UI, it will be very easy to change the algorithm by accident, and that's worst than having to delete and add a new shape.

That being said, have in mind that heal/clone use a clone mask, fill/blur uses regular mask, so either you have to change the mask type (not easy) or limit the change depending on the algorithm selected.

TurboGit:
I agree with @edgardoh, we want to change algorithm only for compatible mask, I'm not sure what happens otherwise.

edgardoh:
I just did a quick test, and with non-continuos add, if adding a circle, the circle is selected right after the add, so to add a different algorithm the user will have to un-select the circle, not very nice, or we'll have to remove the feature, but I like it, so I hope that won't happen.

In continuos add shapes seem not to be selected, I don't remember internally, but that's easy to test.

AlicVB:
@edgardoh : no, the idea is just to enhance usability, not to break anything !
- I've limited shape change to clone/heal and fill/blur
- The change require ctrl key, in order to avoid unwanted algorithm change

TurboGit:
Looks good to me. Thanks.

Issue #2230, opened 2019-03-16
Expose opencl_synch_cache on preferences


edgardoh:
darktablerc->opencl_synch_cache is set to false by default and in my system using the cache with opencl makes it at least 10 times faster.

Having it on the preferences make it easy to play with it and see what configuration work best for each system. I'm sure that most people are not even aware of this setting.

johans3:
Indeed - I've a relative powerful desktop Mac but the performance isn't butter smooth. I will try out your tips, thanks

dtorop:
To clarify: With `opencl_synch_cache` set to false, is it true that the full pixelpipe is executed even when only a later iop (say "tone curve") is changed? And hence time-consuming earlier iops such as "demosaic" or "lens correction" will always contribute to the pipeline processing time under OpenCL? But that the cost of copying intermediate output from GPU to CPU is such that it is generally best to keep the image data on the GPU rather than copying it back to the CPU (and the cache) after each iop?

If so, would it make sense to selectively cache only the input buffer of the currently focused iop? That way future operations on the currently focused iop would be able to start from the cache, but there would not need to be many useless round-trips between GPU and CPU.

I'm looking at https://github.com/darktable-org/darktable/blob/f3aa9835b43b2378926e1cb88851fa032737f479/src/develop/pixelpipe_hb.c#L1696-L1724 for the code which currently syncs from GPU to CPU, and https://github.com/darktable-org/darktable/blob/f3aa9835b43b2378926e1cb88851fa032737f479/src/develop/pixelpipe_hb.c#L2305-L2310 for the code which currently tries to give priority to caching the currently focused iop.

Incidentally, here toggling `opencl_synch_cache` is a wash: it appears that the extra time spent copying memory obviates the speed-up of caching. But either way, it's laggy to update an iop's parameters, so I'm curious if caching and OpenCL could work together.

dtorop:
See https://github.com/dtorop/darktable/commit/be46298e5c958b8c9617fd06e01979a40bdc44b0 for an attempt to deal with this by only copying from the GPU the input to the currently focused iop.

At quick testing this gives me noticeable speedup (over 2x). I'm curious how this seems from there.

edgardoh:
In my system it doesn't seem to be that expensive the copy from device to host, as I said, I have a 10x improvement. Of course it depends of the module being edited, the position in the pipe, the earlier modules, etc.

[...]

PR #2367, opened 2019-04-10
allow caching within OpenCL pixelpipe

dtorop:
Make `opencl_synch_cache` an enum, with new default option "active module". This will cache input to the current module in the OpenCL pipe. This offers a speed-up when the current module is repeatedly adjusted (a normal case) and the time to run the pixelpipe up to the current module is noticeable.

As suggested by #2230.

### DIFF
--- a/data/darktableconfig.xml.in
+++ b/data/darktableconfig.xml.in
@@ -112,13 +112,6 @@
     <longdescription>if set to TRUE OpenCL pixelpipe will not be synchronized on a per-module basis. this can improve pixelpipe latency. however, potential OpenCL errors would be detected late; in such a case the complete pixelpipe needs to be reprocessed instead of only a single module. export pixelpipe will always be run synchronously.</longdescription>
   </dtconfig>
   <dtconfig>
-    <name>opencl_synch_cache</name>
-    <type>bool</type>
-    <default>false</default>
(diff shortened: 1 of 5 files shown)


TurboGit:
Looks good, let's this be merged now for having more field testing. Thanks.

dtorop:
Great. Extant configs will have the prior default of `opencl_synch_cache`, which was FALSE, so users won't see results of this change unless they go in and edit their darktablerc, or start using darktable with an empty config. This may limit field testing...

dtorop:
For the record, to test this with an extant config, alter ~/.conf/darktablerc (assuming a standard install):

```diff
< opencl_synch_cache=false
> opencl_synch_cache=active module
```

AxelG-DE:
@dtorop would you mind to mention this on pixls.us and link to here?

dtorop:
> @dtorop would you mind to mention this on pixls.us and link to here?

@AxelG-DE slowly catching up! I'm happy to mention this, presumably on https://discuss.pixls.us/c/software/darktable, yes? I haven't been a pixls.us user, but just set up an account. Presumably we should wait until #2514 is solved before posting, though...

AxelG-DE:
I could also do it.

Maybe it would be also good to have wider user base for field tests?

PR #2370, opened 2019-04-10
lighttable : hide scrollbar in preview/expose/culling

AlicVB:
@edgardoh : can you confirm scrollbars are not needed/wanted in culling layout (to be sure I've not missed some functionality)

### DIFF
--- a/src/views/lighttable.c
+++ b/src/views/lighttable.c
@@ -280,6 +280,20 @@ static inline void filmstrip_set_active_image(dt_library_t *lib, const int imgid
   if(lib->full_preview_id > -1) lib->full_preview_id = imgid;
 }
 
+static void _scrollbars_restore()
+{
+  char *scrollbars_conf = dt_conf_get_string("scrollbars");
+
+  gboolean scrollbars_visible = FALSE;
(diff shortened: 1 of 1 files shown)


TurboGit:
I have the scrollbar activated only on the lighttable but do not see them on exposé or culling mode. Is that only me?

TurboGit:
Of course I agree that they should not be present on those mode.

edgardoh:
I have never used scrollbars on culling or expose, so this is ok with me.

AlicVB:
In fact there's 2 different scrollbars : one for showing the position at the extrem left and right : read only
and one just on the right, which is not read only. It's visible status is determined by the "scrollbars" pref entry.
We can show the 2 right scrollbars on the screenshot above (take in filemanager layout, but that was exactly the same in preview)
Capture d’écran_2019-04-10_19-08-38

TurboGit:
In full preview I see the scrollbars, yes but not on culling or exposé.

TurboGit:
And indeed this hides the scrollbars from full preview. So merging. Thanks.

PR #2371, opened 2019-04-10
mipmaps : increase default size

AlicVB:
too low cache size can be problematic when zooming with multiple images (expose or culling) or with big images (pano)
Also avoid to preload full res mipmaps as they are not stored in cache anyway

### DIFF
--- a/src/views/lighttable.c
+++ b/src/views/lighttable.c
@@ -2254,7 +2254,7 @@ static int expose_full_preview(dt_view_t *self, cairo_t *cr, int32_t width, int3
       * Simply swapping DESC and ASC in the SQL won't help because we rely on the LIMIT clause, and
       * that LIMIT has to work with the "correct" sort order. One could use a subquery, but I don't
       * think that would be terribly elegant, either. */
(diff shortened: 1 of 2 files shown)


TurboGit:
Looks good! Thanks.

src/views/lighttable.c
```diff
@@ -2254,7 +2254,7 @@ static int expose_full_preview(dt_view_t *self, cairo_t *cr, int32_t width, int3
       * Simply swapping DESC and ASC in the SQL won't help because we rely on the LIMIT clause, and
       * that LIMIT has to work with the "correct" sort order. One could use a subquery, but I don't
       * think that would be terribly elegant, either. */
-      while(--count >= 0 && preload_stack[count] != -1)
+      while(--count >= 0 && preload_stack[count] != -1 && mip != DT_MIPMAP_8)
```
edgardoh: I think this is messing with the zoom, at least on culling, sometimes I can't go over 100% on edited images, and when changing images I get different max zoom even if the images have similar edits and are from the same camera.

PR #2387, opened 2019-04-13
Make info line non-modifiable

matze:
Otherwise, a text cursor is visible and the user can drag and drop the text with mouse.

P.S.: I wasted most of the time on fixing the unpleasant situation with the rawspeed submodule …


TurboGit:
Sounds good. Thanks!

PR #2389, opened 2019-04-13
add error messages to colorin

edgardoh:
Add some fprintf() to colorin, we want those errors on the console.


TurboGit:
Looks good, thanks.

PR #2391, opened 2019-04-13
Added second order sorting functionality to image collection. 

rrd1:
Most helpful when sorting by rating; example: sort by GROUP, RATING. It simply takes previous sorting criteria as second order criteria.

See also on Redmine https://redmine.darktable.org/issues/12630

### DIFF
--- a/src/common/collection.c
+++ b/src/common/collection.c
@@ -40,7 +40,7 @@
 
 
 #define SELECT_QUERY "SELECT DISTINCT * FROM %s"
-#define ORDER_BY_QUERY "ORDER BY %s"
+#define ORDER_BY_QUERY "ORDER BY" /*changed format for second order sorting!*/
 #define LIMIT_QUERY "LIMIT ?1, ?2"
 
 static const char *comparators[] = {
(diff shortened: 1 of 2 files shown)


src/common/collection.c
```diff
@@ -40,7 +40,7 @@
 
 
 #define SELECT_QUERY "SELECT DISTINCT * FROM %s"
-#define ORDER_BY_QUERY "ORDER BY %s"
+#define ORDER_BY_QUERY "ORDER BY" /*changed format for second order sorting!*/
```
TurboGit: Well since ORDER_BY_QUERY is now a simple string...

src/common/collection.c
```diff
@@ -470,57 +555,115 @@ gboolean dt_collection_get_sort_descending(const dt_collection_t *collection)
...
+  }
+
 
   if(collection->params.descending)
   {
     switch(collection->params.sort)
     {
       case DT_COLLECTION_SORT_DATETIME:
-        sq = dt_util_dstrcat(sq, ORDER_BY_QUERY, "datetime_taken DESC, filename DESC, version DESC");
+        sq = dt_util_dstrcat(sq, "%s datetime_taken DESC, %s, filename DESC, version DESC", ORDER_BY_QUERY, second_order);
```
TurboGit: Better write "ORDER BY datatime_taken DESC, %s...", second_order.

Here and all other occurrences.

src/common/collection.c
```diff
@@ -470,57 +555,115 @@ gboolean dt_collection_get_sort_descending(const dt_collection_t *collection)
...
+          second_order = dt_util_dstrcat(NULL, "m.value %s, caption %s", (collection->params.descending ? "DESC" : ""), (collection->params.descending ? "DESC" : ""));
+          break;
+
+        case DT_COLLECTION_SORT_ASPECT_RATIO:
+          second_order = dt_util_dstrcat(NULL, "aspect_ratio %s", (collection->params.descending ? "DESC" : ""));
+          break;
+
+        case DT_COLLECTION_SORT_SHUFFLE:
+          /* do not remember shuffle for second order */
+          if( !second_order ) second_order = dt_util_dstrcat(NULL, "filename %s", (collection->params.descending ? "DESC" : ""));/*only set if not yet initialized*/
```
TurboGit: no spaces inside parenthesis:  if(!second_order) ...

src/common/collection.c
```diff
@@ -470,57 +555,115 @@ gboolean dt_collection_get_sort_descending(const dt_collection_t *collection)
 gchar *dt_collection_get_sort_query(const dt_collection_t *collection)
 {
   gchar *sq = NULL;
+  gchar *second_order = NULL;/*remember previous sorting criteria as second order sorting criteria*/
+
+  switch(collection->params.sortSecondOrder)/*build ORDER BY string for second order*/
+  {
+        case DT_COLLECTION_SORT_DATETIME:
```
TurboGit: I feel that the indentation is far too much here. Can you double check?

src/common/collection.c
```diff
@@ -470,57 +555,115 @@ gboolean dt_collection_get_sort_descending(const dt_collection_t *collection)
 gchar *dt_collection_get_sort_query(const dt_collection_t *collection)
 {
   gchar *sq = NULL;
+  gchar *second_order = NULL;/*remember previous sorting criteria as second order sorting criteria*/
```
TurboGit: I don't see this released, so I suspect a memory leak.

TurboGit:
It takes the previous sort order as second order.

What happen the first time? If second_order is NULL, won't the query be broken?

src/common/collection.c
rrd1: done.

src/common/collection.c
rrd1: done.

src/common/collection.c
rrd1: done.

src/common/collection.c
rrd1: True, done.

src/common/collection.c
rrd1: This is just local variable that keeps string for query. To avoid that it remains NULL, I just added default case which shouldn't occur anyway.

rrd1:
> It takes the previous sort order as second order.
> 
> What happen the first time? If second_order is NULL, won't the query be broken?

second_order is just the local variable to keep query string, see comment above. Previous sorting criteria is stored in collection->params.sortSecondOrder.

collection->params.sortSecondOrder is stored in preferences same as the already existing params.sort. In case of the very first program start it may be DT_COLLECTION_SORT_NONE. Which was covered by switch. Additionally, I just added default case just to be safe. Then it just takes filename as second order.

src/common/collection.c
TurboGit: But even so, you assign value to this variable:

`second_order` = dt_util_dstrcat(...);`

So you need to free this variable as dr_util_sstrcat() does allocate memory.

So when no more needed you need `g_free(second_order);`

src/common/collection.c
```diff
@@ -300,6 +379,7 @@ void dt_collection_reset(const dt_collection_t *collection)
   params->comparator = dt_conf_get_int("plugins/collection/rating_comparator");
   params->filter_flags = dt_conf_get_int("plugins/collection/filter_flags");
   params->sort = dt_conf_get_int("plugins/collection/sort");
+  params->sortSecondOrder = dt_conf_get_int("plugins/collection/sortSecondOrder");
```
TurboGit: style here, all variable are lower-case in dt, so I would use `params->sort_second_order`.

src/common/collection.h
```diff
@@ -128,6 +128,7 @@ typedef struct dt_collection_params_t
 
   /** sorting **/
   dt_collection_sort_t sort; // Has to be changed to a dt_collection_sort struct
+  dt_collection_sort_t sortSecondOrder;
```
TurboGit: sort_second_order

TurboGit:
Some more changes, I'll test in next iteration. Thanks.

src/common/collection.c
rrd1: Oups, you're right. I'll fix this!

TurboGit:
Looks good and seems to be working as expected. Thanks.

rrd1:
> Looks good and seems to be working as expected. Thanks.

Great, thanks!

PR #2397, opened 2019-04-13
mipmap : fix if embedded preview size is too small

### LINKED ISSUE #2377: lighttable: crash when zoom in on full preview of image without history
**Describe the bug**
The new zoom feature in lighttable (#2267) gets confused when zooming in on images without a history. This can result in a crash.

**To Reproduce**
1. In lighttable view, go to an image with no history (e.g. freshly imported).
2. Type "z" to show a fullscreen preview
4. ctrl-scroll to zoom in.
5. dt may crash within `dt_iop_flip_and_zoom_8`.

Alternately, dt will zoom in on the image, but not to the full resolution of the raw file but rather to the maximum resolution of the JPEG preview embeddded within the raw file.

This may be further complicated by the setting of `never_use_embedded_thumb` conf value.

I have trouble replicating this bug consistently. I think it's easiest when there are not yet entries in the cache from the image. (Try running with --cachedir set to an empty directory.)

**Expected behavior**
This is a tricky one. Should it zoom in on the JPEG within the raw? In this case we don't see the full resolution of the raw file. But if we run the pixelpipe when we start showing a mip8 to show all the pixels, this will not be consistent with the embedded JPEG data used to produce mip0-7.

**Backtrace**
Here's a backtrace with OpenMP turned off

```
Thread 5 "worker 0" received signal SIGSEGV, Segmentation fault.
[Switching to Thread 0x7fffdefa6700 (LWP 32085)]
dt_iop_flip_and_zoom_8 (in=<optimized out>, iw=0, ih=<optimized out>, out=0x7fff9af30062 "\261", ow=0, oh=0, orientation=ORIENTATION_NONE, 
    width=0x7fff9af30040, height=0x7fff9af30044) at ../src/develop/imageop_math.c:91
91	          out2[k] = // in3[k];
(gdb) bt
#0  0x00007ffff7dd5f6b in dt_iop_flip_and_zoom_8
    (in=<optimized out>, iw=0, ih=<optimized out>, out=0x7fff9af30062 "\261", ow=0, oh=0, orientation=ORIENTATION_NONE, width=0x7fff9af30040, height=0x7fff9af30044) at ../src/develop/imageop_math.c:91
#1  0x00007ffff7d773c6 in _init_8
    (buf=<optimized out>, width=<optimized out>, height=<optimized out>, iscale=<optimized out>, color_space=<optimized out>, imgid=<optimized out>, size=<optimized out>) at ../src/common/mipmap_cache.c:1272
#2  0x00007ffff7d76919 in dt_mipmap_cache_get_with_caller
    (cache=<optimized out>, buf=0x7fffdef97678, imgid=<optimized out>, mip=<optimized out>, flags=<optimized out>, mode=<optimized out>, file=<optimized out>, line=<optimized out>) at ../src/common/mipmap_cache.c:862
#3  0x00007ffff7da1208 in dt_image_load_job_run (job=<optimized out>) at ../src/control/jobs/image_jobs.c:35
#4  0x00007ffff7d9a367 in dt_control_job_execute (job=0x2015a00) at ../src/control/jobs.c:304
#5  0x00007ffff7d9aba3 in dt_control_run_job (control=<optimized out>) at ../src/control/jobs.c:323
#6  0x00007ffff7d9aba3 in dt_control_work (ptr=<optimized out>) at ../src/control/jobs.c:568
#7  0x00007ffff7c53fa3 in start_thread (arg=<optimized out>) at pthread_create.c:486
#8  0x00007ffff7b8482f in clone () at ../sysdeps/unix/sysv/linux/x86_64/clone.S:95
```

AlicVB:
this fix #2377 
There's 2 problem with some raw with small embedded preview images (https://raw.pixls.us/getfile.php/1942/nice/Fujifilm%20-%20X100F%20-%2014bit%20uncompressed%20(3:2).RAF) :
- crash if one zoom too quickly in full preview mode
- the embedded preview is small (1920x1280) so the zoom is quite limited (real image size is 6160x4032)

In fact the main problem is that exiv2 read the embedded preview size in place of the image size... So I've added a way to run minimal rawspeed pass, just to get the real dimension of images with embedded preview.

### DIFF
--- a/src/common/image.c
+++ b/src/common/image.c
@@ -1194,7 +1194,7 @@ uint32_t dt_image_import_lua(const int32_t film_id, const char *filename, gboole
 
 void dt_image_init(dt_image_t *img)
 {
-  img->width = img->height = 0;
+  img->width = img->height = img->verified_size = 0;
   img->crop_x = img->crop_y = img->crop_width = img->crop_height = 0;
   img->orientation = ORIENTATION_NULL;
   img->legacy_flip.legacy = 0;
--- a/src/common/image.c
+++ b/src/common/image.c
@@ -383,16 +383,54 @@ void dt_image_set_location_and_elevation(const int32_t imgid, dt_image_geoloc_t 
   dt_image_cache_write_release(darktable.image_cache, image, DT_IMAGE_CACHE_SAFE);
 }
 
+void dt_image_reset_final_size(const int32_t imgid)
+{
+  dt_image_t *imgtmp = dt_image_cache_get(darktable.image_cache, imgid, 'w');
+  imgtmp->final_width = imgtmp->final_height = 0;
+  dt_image_cache_write_release(darktable.image_cache, imgtmp, DT_IMAGE_CACHE_RELAXED);
+}
+
 gboolean dt_image_get_final_size(const int32_t imgid, int *width, int *height)
 {
-  dt_develop_t dev;
+  // get the img strcut
+  dt_image_t *imgtmp = dt_image_cache_get(darktable.image_cache, imgid, 'r');
+  dt_image_t img = *imgtmp;
+  dt_image_cache_read_release(darktable.image_cache, imgtmp);
+  // if we already have computed them
+  if(img.final_height > 0 && img.final_width > 0)
+  {
+    *width = img.final_width;
+    *height = img.final_height;
+    return 0;
+  }
 
+  // special case if we try to load embedded preview of raw file
+
+  // the orientation for this camera is not read correctly from exiv2, so we need
+  // to go the full path (as the thumbnail will be flipped the wrong way round)
+  const int incompatible = !strncmp(img.exif_maker, "Phase One", 9);
+  if(!img.verified_size && !dt_image_altered(imgid) && !dt_conf_get_bool("never_use_embedded_thumb")
+     && !incompatible)
+  {
+    // we want to be sure to have the real image size.
+    // some raw files need a pass via rawspeed to get it.
+    char filename[PATH_MAX] = { 0 };
+    gboolean from_cache = TRUE;
+    dt_image_full_path(imgid, filename, sizeof(filename), &from_cache);
+    imgtmp = dt_image_cache_get(darktable.image_cache, imgid, 'w');
+    dt_imageio_open(imgtmp, filename, NULL);
+    imgtmp->verified_size = 1;
+    img = *imgtmp;
+    dt_image_cache_write_release(darktable.image_cache, imgtmp, DT_IMAGE_CACHE_RELAXED);
+  }
+
+  // and now we can do the pipe stuff to get final image size
+  dt_develop_t dev;
   dt_dev_init(&dev, 0);
   dt_dev_load_image(&dev, imgid);
-  const dt_image_t *img = &dev.image_storage;
 
   dt_dev_pixelpipe_t pipe;
-  int wd = img->width, ht = img->height;
+  int wd = img.width, ht = img.height;
   int res = dt_dev_pixelpipe_init_dummy(&pipe, wd, ht);
   if(res)
   {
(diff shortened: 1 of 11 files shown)


dtorop:
This looks impressive. One quick thought is if `verified_size` should be a gboolean rather than an int. That seems in keeping with dt coding practices.

I want to look more carefully at the `dt_imageio_open_XXX()` changes, and haven't actually tested this yet. More in a bit.

AlicVB:
> This looks impressive. One quick thought is if `verified_size` should be a gboolean rather than an int. That seems in keeping with dt coding practices.

I've just mimic the other "boolean" present in the structure...


> I want to look more carefully at the `dt_imageio_open_XXX()` changes, and haven't actually tested this yet. More in a bit.

This is clearly the sensible area of change (that with all the cache get/release)

dtorop:
I tested this a bit and it looks good. Not only does It not crash, but it behaves just as I'd expect/hope, without having to think about any subtleties.

>> I want to look more carefully at the dt_imageio_open_XXX() changes
>> ...
> This is clearly the sensible area of change (that with all the cache get/release)

My one regret about the `dt_imageio_open()` work is that I wish that each sub-image type didn't have to know separately how to handle a NULL buf, and that there's implicit behavior where opening a raw goes all the way through calculating width/height before returning on NULL buf, but not for other formats. But I imagine the only way to unwind this would be to make a separate dt_imageio_get_dims() function which was used by dt_imageio_open, and that seems both absurd and wasteful, so I see why what you did makes sense.

dtorop:
One question: it looks like even for mipmaps < mip8, it looks like now the pixelpipe results appear rather than embedded JPEG. Is this expected/right?

AlicVB:
> One question: it looks like even for mipmaps < mip8, it looks like now the pixelpipe results appear rather than embedded JPEG. Is this expected/right?

What is expected, is that the computed mipmap appears for all mipmap levels where the embedded thumbnail is too small. In the case of the fuji example, the embedded thumbnail should be used for mipmap <= 4 and computed after (if I remember correctly the mipmap sizes). Is that what you see ?

Sadly, I can't test anything : I've broken my system with a tentative of upgrading gtk :), so it's time to a clean reinstall of a more recent OS version. Anyway, I'm leaving for holidays tomorrow, so I'll come back to dev in 10-15 days.

TurboGit:
@AlicVB : since you won't be around for some days I'll wait for your come back to merge this. Ping me when merging can be done. Thanks.

AlicVB:
Ok, I've just added some refactoring around the image "final size" computation. Now final size are stored in dt_image_t and compute if needed (first need / after image changes)
As a side effect, this allow to avoid some "jumpy" effect in lighttable preview zooming when you zoom to max quickly.

Note : 
- this sizes are not stored in db, so they need to be recompute at each session. Eventually this can be changed, but this need a db schema change and to be sure to update the db on each image change (2 sensible areas I know nothing about, so I've avoid it)

src/common/image.c
```diff
@@ -383,16 +383,54 @@ void dt_image_set_location_and_elevation(const int32_t imgid, dt_image_geoloc_t
...
+    *width = img.final_width;
+    *height = img.final_height;
+    return 0;
+  }
+
+  // special case if we try to load embedded preview of raw file
 
+  // the orientation for this camera is not read correctly from exiv2, so we need
+  // to go the full path (as the thumbnail will be flipped the wrong way round)
+  const int incompatible = !strncmp(img.exif_maker, "Phase One", 9);
```
TurboGit: This does not look like refactoring. The orientation is wrong with current version? Note that I have recently fixed lot of orientation issues. Have you tested with current version? Just to be sure this is still needed.

AlicVB: No idea, this is copy/paste from mipmap_cache.c line 1187. Git blame says it's a commit from @hanatos 7 years ago, and it's a problem for Phase One h25 models (raw pixls didn't seems to have this raw...)

src/views/lighttable.c
```diff
@@ -353,7 +353,7 @@ static inline void _destroy_preview_surface(dt_preview_surface_t *fp_surf)
   fp_surf->imgid = -1;
   fp_surf->w_lock = 0;
 
-  fp_surf->zoom_100 = 40.0f;
+  fp_surf->zoom_100 = 1001.0f; // dummy vaule to say it need recompute
```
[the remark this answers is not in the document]
AlicVB: fixed

AlicVB:
for the rawspeed files, I don't know where they come from, git show doesn't seems to have them here, certainly some subtilities of the submodules...
I'm not sure how to fix it, github doesn't want to remove them directly from the pull request... Any hint ?
Thank you... (and apologize if I've done something stupid...)

TurboGit:
Working fine on my side.
I have merged this manually to remove the rawspeed change.

Thanks.

PR #2399, opened 2019-04-13
hazeremoval: OpenCL implementation

rabauke:
Now where we have a proper OpenCL implementation of the guided filter, also the haze removal module can be implemented in OpenCL.


TurboGit:
Looks good, working fine. Thanks.

Just a comment about style, I would like to have all the new lines added into a separate commit. In fact I'm not even sure we want double empty line after each function.

PR #2411, opened 2019-04-15
UI: set font weight to normal in history stack

rabauke:
The bold font weight for items of the history stack gives (imho) too much emphasis to the history stack. Looks distracting to me. Fixes also some inconsistency in the font selection.


TurboGit:
Looks good to me. Thanks.

PR #2414, opened 2019-04-15
bauhaus: fix linker problem

rabauke:
Suddenly, some linker error popped up today when compiling darktable using clang-7 or gcc-8. As a consequence of the inline keyword's semantics, inline functions should be declared static in C header files.

### DIFF
--- a/src/bauhaus/bauhaus.h
+++ b/src/bauhaus/bauhaus.h
@@ -317,7 +317,7 @@ void dt_bauhaus_vimkey_exec(const char *input);
 // give autocomplete suggestions
 GList *dt_bauhaus_vimkey_complete(const char *input);
 
-inline void set_color(cairo_t *cr, GdkRGBA color)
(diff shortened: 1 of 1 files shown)


TurboGit:
Just curious, which version of GCC 8 exactly ? It builds fine on my side with GCC 8.3.0.

BTW, it also works with clang-8.

rabauke:
Versions are

```
clang version 7.0.0-3~ubuntu0.18.04.1 (tags/RELEASE_700/final)
Target: x86_64-pc-linux-gnu
Thread model: posix
InstalledDir: /usr/bin
```

and

```
gcc-8 (Ubuntu 8.2.0-1ubuntu2~18.04) 8.2.0
```

Considering that the problem emerged at the same time with both compilers, the issue might actually not be caused by the compiler but by some other program in the tool chain. But I did not investigate this in more detail.

TurboGit:
Ok, anyway looks fine. Thanks.

PR #2417, opened 2019-04-16
bauhaus: fix zero-point of soft-bound sliders

rabauke:
The zero position of sliders with soft bounds is currently indicated incorrectly unless hard and soft bounds feature some symmetries. This issue can be observed for the contrast slider of the basic adjustments module. This slider has soft bounds [-1, 1] and hard bounds [-1, 5]. This pull request provides a fix for the described bug. It also fixes a typo, which might be relevant for periodic sliders, see comment in code.

To demonstrate the bug here two screen shorts before and after the fix. Picture show basic adjustment module with default settings. Compare the contrast slider.

Broken soft-bound slider:
broken

Fixed soft-bound slider:
fixed


TurboGit:
Looks correct. Thanks.

PR #2419, opened 2019-04-16
dt_control_queue_redraw() is causing recursion

edgardoh:
dt_control_queue_redraw() is causing recursion, to test run dt with -d perf:

6,193952 [dev] took 0,211 secs (0,181 CPU) to load the image.
7,293687 [dev] took 0,000 secs (0,000 CPU) to load the image.
7,407167 [dev_pixelpipe] took 0,000 secs (0,000 CPU) initing base buffer [full]
7,423194 [dev_pixelpipe] took 0,016 secs (0,041 CPU) processed raw black/white point' on CPU, blended on CPU [full] 7,440804 [dev_pixelpipe] took 0,018 secs (0,068 CPU) processed white balance' on CPU, blended on CPU [full]
7,459428 [dev_pixelpipe] took 0,019 secs (0,067 CPU) processed highlight reconstruction' on CPU, blended on CPU [full] 7,512458 [dev_pixelpipe] took 0,053 secs (0,082 CPU) processed hot pixels' on CPU, blended on CPU [full]
8,310029 [dev_pixelpipe] took 0,798 secs (5,688 CPU) processed raw denoise' on CPU, blended on CPU [full] 8,821389 [dev_pixelpipe] took 0,511 secs (3,169 CPU) processed demosaic' on CPU, blended on CPU [full]
8,940998 [dev_pixelpipe] took 0,120 secs (0,149 CPU) processed exposure' on CPU, blended on CPU [full] 8,973271 [dev_pixelpipe] took 0,032 secs (0,058 CPU) processed exposure 1' on CPU, blended on CPU [full]
9,000515 [dev_pixelpipe] took 0,027 secs (0,052 CPU) processed exposure 2' on CPU, blended on CPU [full] 9,215437 [dev_pixelpipe] took 0,215 secs (0,509 CPU) processed retouch' on CPU, blended on CPU [full]
9,218615 [dev_pixelpipe] took 0,003 secs (0,014 CPU) processed base curve' on CPU, blended on CPU [full] 9,221695 [dev_pixelpipe] took 0,003 secs (0,016 CPU) processed input color profile' on CPU, blended on CPU [full]
9,228279 [dev_pixelpipe] took 0,007 secs (0,036 CPU) processed sharpen' on CPU, blended on CPU [full] 9,249403 [dev_pixelpipe] took 0,021 secs (0,126 CPU) processed lowpass' on CPU, blended on CPU [full]
9,257166 [dev_pixelpipe] took 0,008 secs (0,037 CPU) processed output color profile' on CPU, blended on CPU [full] 9,260789 [dev_pixelpipe] took 0,004 secs (0,021 CPU) processed gamma' on CPU, blended on CPU [full]
9,260857 [dev_process_image] pixel pipeline processing took 1,856 secs (10,141 CPU)
9,260896 [dev] took 0,000 secs (0,000 CPU) to load the image.
9,260977 [dev_process_image] pixel pipeline processing took 0,000 secs (0,000 CPU)
9,302050 [dev] took 0,000 secs (0,000 CPU) to load the image.
9,302173 [dev_process_image] pixel pipeline processing took 0,000 secs (0,000 CPU)
9,353275 [dev] took 0,000 secs (0,000 CPU) to load the image.
9,353378 [dev_process_image] pixel pipeline processing took 0,000 secs (0,000 CPU)
9,440795 [dev] took 0,000 secs (0,000 CPU) to load the image.
9,440906 [dev_process_image] pixel pipeline processing took 0,000 secs (0,003 CPU)
9,494164 [dev] took 0,000 secs (0,000 CPU) to load the image.
9,494259 [dev_process_image] pixel pipeline processing took 0,000 secs (0,000 CPU)
9,506879 [dev] took 0,000 secs (0,000 CPU) to load the image.
9,506954 [dev_process_image] pixel pipeline processing took 0,000 secs (0,000 CPU)
9,548917 [dev] took 0,000 secs (0,000 CPU) to load the image.
9,549008 [dev_process_image] pixel pipeline processing took 0,000 secs (0,000 CPU)
9,602631 [dev] took 0,000 secs (0,000 CPU) to load the image.
9,602739 [dev_process_image] pixel pipeline processing took 0,000 secs (0,000 CPU)
9,661684 [dev] took 0,000 secs (0,000 CPU) to load the image.
9,661825 [dev_process_image] pixel pipeline processing took 0,000 secs (0,000 CPU)
9,672816 [dev] took 0,000 secs (0,000 CPU) to load the image.
9,673004 [dev_process_image] pixel pipeline processing took 0,000 secs (0,000 CPU)
9,716285 [dev] took 0,000 secs (0,000 CPU) to load the image.
9,716423 [dev_process_image] pixel pipeline processing took 0,000 secs (0,000 CPU)
9,765645 [dev] took 0,000 secs (0,000 CPU) to load the image.
9,765741 [dev_process_image] pixel pipeline processing took 0,000 secs (0,000 CPU)
9,972753 [dev_pixelpipe] took 0,000 secs (0,000 CPU) initing base buffer [preview]
9,974363 [dev_pixelpipe] took 0,002 secs (0,000 CPU) processed raw black/white point' on CPU, blended on CPU [preview] 9,975432 [dev_pixelpipe] took 0,001 secs (0,004 CPU) processed white balance' on CPU, blended on CPU [preview]
9,976565 [dev_pixelpipe] took 0,001 secs (0,004 CPU) processed highlight reconstruction' on CPU, blended on CPU [preview] 9,979126 [dev_pixelpipe] took 0,003 secs (0,000 CPU) processed hot pixels' on CPU, blended on CPU [preview]
10,001437 [dev_pixelpipe] took 0,022 secs (0,143 CPU) processed raw denoise' on CPU, blended on CPU [preview] 10,017906 [dev_pixelpipe] took 0,016 secs (0,061 CPU) processed demosaic' on CPU, blended on CPU [preview]
10,125192 [dev_pixelpipe] took 0,107 secs (0,153 CPU) processed exposure' on CPU, blended on CPU [preview] 10,147949 [dev_pixelpipe] took 0,022 secs (0,061 CPU) processed exposure 1' on CPU, blended on CPU [preview]
10,166531 [dev_pixelpipe] took 0,018 secs (0,062 CPU) processed exposure 2' on CPU, blended on CPU [preview] 10,463780 [dev_pixelpipe] took 0,297 secs (0,632 CPU) processed retouch' on CPU, blended on CPU [preview]
10,467547 [dev_pixelpipe] took 0,004 secs (0,024 CPU) processed base curve' on CPU, blended on CPU [preview] 10,470977 [dev_pixelpipe] took 0,003 secs (0,026 CPU) processed input color profile' on CPU, blended on CPU [preview]
10,480352 [dev_pixelpipe] took 0,009 secs (0,054 CPU) processed sharpen' on CPU, blended on CPU [preview] 10,509463 [dev_pixelpipe] took 0,029 secs (0,147 CPU) processed lowpass' on CPU, blended on CPU [preview]
10,519134 [dev_pixelpipe] took 0,010 secs (0,052 CPU) processed output color profile' on CPU, blended on CPU [preview] 10,524584 [dev_pixelpipe] took 0,005 secs (0,029 CPU) processed gamma' on CPU, blended on CPU [preview]
image colorspace transform RGB-->RGB took 0,020 secs (0,129 CPU) [final histogram]
10,654463 [dev_process_preview] pixel pipeline processing took 0,716 secs (1,740 CPU)
10,654546 [dev_process_preview] pixel pipeline processing took 0,000 secs (0,000 CPU)

### DIFF
--- a/src/views/darkroom.c
+++ b/src/views/darkroom.c
@@ -221,20 +221,19 @@ void expose(
   if(dev->image_status == DT_DEV_PIXELPIPE_DIRTY || dev->image_status == DT_DEV_PIXELPIPE_INVALID
      || dev->pipe->input_timestamp < dev->preview_pipe->input_timestamp)
   {
-    dt_control_queue_redraw();
-    dev->image_timeout_handle = g_timeout_add(DT_DARKROOM_PROCESS_TIMEOUT, G_SOURCE_FUNC(dt_dev_process_image), dev);
+    dt_dev_process_image(dev);
   }
 
   if(dev->preview_status == DT_DEV_PIXELPIPE_DIRTY || dev->preview_status == DT_DEV_PIXELPIPE_INVALID
      || dev->pipe->input_timestamp > dev->preview_pipe->input_timestamp)
   {
-    dev->image_timeout_handle = g_timeout_add(DT_DARKROOM_PROCESS_TIMEOUT, G_SOURCE_FUNC(dt_dev_process_preview), dev);
+    dt_dev_process_preview(dev);
   }
 
   if(dev->preview2_status == DT_DEV_PIXELPIPE_DIRTY || dev->preview2_status == DT_DEV_PIXELPIPE_INVALID
      || dev->pipe->input_timestamp > dev->preview2_pipe->input_timestamp)
   {
-    dev->image_timeout_handle = g_timeout_add(DT_DARKROOM_PROCESS_TIMEOUT, G_SOURCE_FUNC(dt_dev_process_preview2), dev);
+    dt_dev_process_preview2(dev);
   }
 
   dt_pthread_mutex_t *mutex = NULL;
@@ -916,6 +915,7 @@ static gboolean zoom_key_accel(GtkAccelGroup *accel_group, GObject *acceleratabl
     default:
       break;
   }
+  dt_control_queue_redraw_center();
   return TRUE;
 }
 
@@ -2358,8 +2358,6 @@ void enter(dt_view_t *self)
   dev->form_gui->formid = 0;
   dev->gui_leaving = 0;
   dev->gui_module = NULL;
-  dev->image_timeout_handle = 0;
-  dev->preview_timeout_handle = 0;
 
   select_this_image(dev->image_storage.id);
 
--- a/src/develop/develop.h
+++ b/src/develop/develop.h
@@ -142,10 +142,6 @@ typedef struct dt_develop_t
   struct dt_iop_module_t *gui_module; // this module claims gui expose/event callbacks.
   float preview_downsampling;         // < 1.0: optionally downsample preview
 
-  // handle GUI events timeout
-  int preview_timeout_handle;
-  int image_timeout_handle;
-
   // width, height: dimensions of window
   int32_t width, height;
 


src/views/darkroom.c
```diff
@@ -221,20 +221,19 @@ void expose(
   if(dev->image_status == DT_DEV_PIXELPIPE_DIRTY || dev->image_status == DT_DEV_PIXELPIPE_INVALID
      || dev->pipe->input_timestamp < dev->preview_pipe->input_timestamp)
   {
-    dt_control_queue_redraw();
-    dev->image_timeout_handle = g_timeout_add(DT_DARKROOM_PROCESS_TIMEOUT, G_SOURCE_FUNC(dt_dev_process_image), dev);
```
TurboGit: This was done to avoid too many processing, the delay was making the UI more responsive IIRC (not tested this change). So I'm not sure this change is correct.

Note also that you have change rawspeed here, to avoid that make sure you do a:

`$ git submodule update`

Before committing changes on your side.

edgardoh:
> This was done to avoid too many processing, the delay was making the UI
> more responsive IIRC (not tested this change).
>
I know, but this is not the right way to do it, that's why I'm reverting it.

Note also that you have change rawspeed here, to avoid that make sure you
> do a:
>
> $ git submodule update
>
> Before committing changes on your side.
>
I always do, no idea what happen here (and in a coupe of previous PR), and
now I can't get rid of it.

TurboGit:
Any idea about the right way?

This issue occurs only with --d perf option, right? So it would seem that the code is not that buggy as all is fine in standard mode. What is specific to -d perf in this case? Have you debugged this a bit?

I'd like to have Aurélien feedback about this to see if we can devise a proper solution.

edgardoh:
I haven't checked this and I'm having some family issues now, so I won't be
able to spend much time in dt for a while, but the right solution is not
trivial at all.

-d perf only printf() some stats, so what you see on the log is what's
actually happening without the -d perf.

There's clearly a recursion going on here, that's because a draw() is
called from the expose, and this is a don't. As per aurelien's comment on
my previous PR that call is required, so something else should be done.

From the user's perspective everything may seem ok, but this is a very
sensitive area and the theory is there for something, I don't think this is
worth the risk just to avoid one reprocess when zooming.

src/views/darkroom.c
aurelienpierre: Yes, the `g_timeout_add` is mandatory to avoid recomputing the pipe twice (or more) when zooming with the mouse scroll (otherwise, one computation is triggered at the beginning of the scroll event, and another at the end, and sometimes more in-between).

aurelienpierre:
I guess the
```
9,260857 [dev_process_image] pixel pipeline processing took 1,856 secs (10,141 CPU)
9,260896 [dev] took 0,000 secs (0,000 CPU) to load the image.
9,260977 [dev_process_image] pixel pipeline processing took 0,000 secs (0,000 CPU)
9,302050 [dev] took 0,000 secs (0,000 CPU) to load the image.
9,302173 [dev_process_image] pixel pipeline processing took 0,000 secs (0,000 CPU)
9,353275 [dev] took 0,000 secs (0,000 CPU) to load the image.
9,353378 [dev_process_image] pixel pipeline processing took 0,000 secs (0,000 CPU)
9,440795 [dev] took 0,000 secs (0,000 CPU) to load the image.
9,440906 [dev_process_image] pixel pipeline processing took 0,000 secs (0,003 CPU)
9,494164 [dev] took 0,000 secs (0,000 CPU) to load the image.
9,494259 [dev_process_image] pixel pipeline processing took 0,000 secs (0,000 CPU)
9,506879 [dev] took 0,000 secs (0,000 CPU) to load the image.
9,506954 [dev_process_image] pixel pipeline processing took 0,000 secs (0,000 CPU)
9,548917 [dev] took 0,000 secs (0,000 CPU) to load the image.
9,549008 [dev_process_image] pixel pipeline processing took 0,000 secs (0,000 CPU)
9,602631 [dev] took 0,000 secs (0,000 CPU) to load the image.
9,602739 [dev_process_image] pixel pipeline processing took 0,000 secs (0,000 CPU)
9,661684 [dev] took 0,000 secs (0,000 CPU) to load the image.
9,661825 [dev_process_image] pixel pipeline processing took 0,000 secs (0,000 CPU)
9,672816 [dev] took 0,000 secs (0,000 CPU) to load the image.
9,673004 [dev_process_image] pixel pipeline processing took 0,000 secs (0,000 CPU)
9,716285 [dev] took 0,000 secs (0,000 CPU) to load the image.
9,716423 [dev_process_image] pixel pipeline processing took 0,000 secs (0,000 CPU)
9,765645 [dev] took 0,000 secs (0,000 CPU) to load the image.
9,765741 [dev_process_image] pixel pipeline processing took 0,000 secs (0,000 CPU)
```
recursion occurs because the timeout is implemented at a low level, and the general pipe instructions are still triggered at an higher level. So the main pixelpipe loop is triggered on UI events, then aborted internally if there is already an active timeout. The cleaner way to do it would be to disable the loop at an higher level, but I don't know the pipe circuitery well enough to mess in there.

edgardoh:
The recursion is because of the dt_control_queue_redraw(), if that can be removed then I have no objections with the timeout.

aurelienpierre:
> The recursion is because of the dt_control_queue_redraw(), if that can be removed then I have no objections with the timeout.

I'm not saying the `dt_control_queue_redraw()` can't be removed at all, but I added it because I noticed that, on some UI events, the pipe was recomputed but the display wasn't updated, so the computation was useless and resulted in pure UI lag for nothing. Maybe there is a cleaner way… I don't know yet.

edgardoh:
A call to dt_control_queue_redraw() when the pipe finish processing
shouldn't cause a recursion as the pipe status will be VALID. It can cause
some extra draw() (or expose()), I don't know the internals of GTK so I
can't say if it will be an issue, but on the scroll it is (or was) called
for each mouse wheel change and everything seems to work fine with that.

aurelienpierre:
I agree with you, but for the time being, I have no better solution.

TurboGit:
So what about removing dt_control_queue_redraw() for now and see how often it is happening and maybe better understand the context of this issue to devise a better fix?

aurelienpierre:
> So what about removing dt_control_queue_redraw() for now and see how often it is happening and maybe better understand the context of this issue to devise a better fix?

We could do that indeed.

TurboGit:
I have removed the call to dt_control_queue_redraw(). Such a call should never be added in expose() anyway. So I suppose that the real fix will need a call to dt_control_queue_redraw() in some other places. We'll see.

I'm closing this ticket at this stage.

PR #2423, opened 2019-04-19
Add different timeouts depending on the importance of the view.

TurboGit:
May help having a more responsive UI.

I'd like @aurelienpierre and @edgardoh point of view on this. Thanks.

### DIFF
--- a/src/views/darkroom.c
+++ b/src/views/darkroom.c
@@ -100,7 +100,10 @@ static void dt_dev_change_image(dt_develop_t *dev, const uint32_t imgid);
 
 static void _darkroom_display_second_window(dt_develop_t *dev);
 static void _darkroom_ui_second_window_write_config(GtkWidget *widget);
-#define DT_DARKROOM_PROCESS_TIMEOUT 500
+
+#define DT_DARKROOM_PROCESS_MAIN_TIMEOUT     500   // the main working window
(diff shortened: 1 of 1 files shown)


aurelienpierre:
Seems fine by me. I will need to test.

aurelienpierre:
Tested. The label "working…" is not displayed anymore after zooming, when the preview is recomputed.

edgardoh:
Some modules use the preview pipe to do some calculations, as they need the entire image, so preview must run before the full pipe, I'm not sure if this can create some conflict, it has to be tested.

TurboGit:
> Tested. The label "working…" is not displayed anymore after zooming, when the preview is recomputed.

Isn't this related to the removal of the dt_control_queue_redraw() call removed in previously?

TurboGit:
> Some modules use the preview pipe to do some calculations, as they need the entire image, so preview must run before the full pipe, I'm not sure if this can create some conflict, it has to be tested.

I see, but my idea was to give the priority to the full preview (or working space) as this one must be recomputed as fast as possible. Too bad, I'll put the preview first and we'll see.

TurboGit:
Ok, so I have set preview timeout to 400, main view to 500 and second preview to 600.

TurboGit:
Any objection to push that now?

edgardoh:
None on my side.

TurboGit:
Ok, so let's see how this works.

PR #2443, opened 2019-04-24
improve comments for CSS sections (help designers)

aurelienpierre:


TurboGit:
Good idea! Thanks.

Nilvus:
Thanks for that. Adding this comments also on darktable-elegant theme would be a good idea.

aurelienpierre:
darktable-elegant should ideally inherit the default CSS (with an `@import` statement) and only overwrite the needed fonts. But I'm not sure how to do that in a way compatible with Win/Mac/Linux.

Nilvus:
That of course would be better but that seems difficult to make it possible for all OS in such CSS file.

But I really think taht darktable-elegant.css theme should be now the default theme not darktable.css with OS font...

aurelienpierre:
> But I really think taht darktable-elegant.css theme should be now the default theme not darktable.css with OS font...

His Highness doesn't like it…

TurboGit:
Since both CSS are in the same directory, importing with relative to "./" may work? I'll try that at some point.

TurboGit:
And yes I agree that the elegant theme should be the default. And the other should have never existed, time was/is needed to get the CSS feel right for everybody. But well...

Nilvus:
> > But I really think taht darktable-elegant.css theme should be now the default theme not darktable.css with OS font...
> 
> His Highness doesn't like it…

Of course, I've a few seconds forgotten that His Highness hold the truth and we are ignorant ones...

theres:
> Since both CSS are in the same directory, importing with relative to "./"

It works like a charm on linux at least.

// offtopic

Please, If you have personal problems with some devs, solve it somewhere else. You can use IRC (still official channel to talk with most of the devs). This is at least not needed here (IMHO it's just rude), gives no input to the discussion and introduces more noise to the GH notifications stream. 

@houz is second contributor (https://github.com/darktable-org/darktable/graphs/contributors), this tool is here also thanks to him even if you disagree with some of his decisions.

That's all, I'm out, hope this discussion **will not** be continued here.

aurelienpierre:
There is no discussion to have. Only regrets and sorrow, sometimes damage control.

As of noise, IRC is only that.

PR #2444, opened 2019-04-24
lighttable : redraw when moving view with keys

### LINKED ISSUE #2376: Scrolling in lighttable with arrow-keys gives weird results
**Describe the bug**
Using the arrow-keys by scrolling thru the thumbnaisl of lighttable gives a very weird result, where not all the pics are scrolling. Some just remain at the position, they where

It is hard to explain. Just the rows do not stay connected as they where and as they actually doing normally when using the mouse wheel....

**To Reproduce**
1. Go to lighttable and set it to 5 pics in a row (me having a selection of 150 pics)
2. Click on second row middle pictures
3. Hit the down-arrow several times

**Expected behavior**
Same behaviour then using the mouse wheel, where all the pics scrolling logically row by row

**Screenshots**
I have a screen cast, but it is 7.2MB and also is my first time to use OBS-Studio, hence things are not ideal, yet. Do I need to upload or is it "clear" by trying to reproduce?

 - OS: [gentoo linux]
 - Version 2.7.0+927~ge1835549f-dirty; 2.7.0+930~g85ccaaa16-dirty

AlicVB:
this fix #2376

### DIFF
--- a/src/views/lighttable.c
+++ b/src/views/lighttable.c
@@ -1200,7 +1200,7 @@ end_query_cache:
 
             // if we have moved the view we need to expose again all pictures as the first row or last one need to
             // be redrawn properly. for this we just record the missing thumbs.
-            if(offset_changed && mouse_over_id != -1)
+            if(lib->offset_changed && mouse_over_id != -1)
             {
               missing += iir;
             }


TurboGit:
A recent regression since I'm sure I had tested this scenario when working on the lighttable speed-up. Thanks for fixing!

PR #2448, opened 2019-04-24
remove a bit of contrast on modules header (names)

aurelienpierre:
The extra contrast was not necessary for readability but a bit distracting from the picture.
On @cryptomilk 's request.


TurboGit:
Looks good. Thanks.

PR #2449, opened 2019-04-24
rgb curve: scale displayed values

### LINKED ISSUE #2344: rgb curve 0-100?
In difference to tone curve the range in rgb curve is from 0 to 1 instead of 0 to 100. So adjusting nodes by cursor keys doesnt change displayed values for several steps even the node moves far away from its original place.

change multiplicator of the range so finer adjustments can be identified

edgardoh:
Scale displayed values to 0..255 on rgb curve so is consistent with the global color picker.
Fixes #2344


TurboGit:
Looks good. Thanks.

PR #2450, opened 2019-04-25
Update windows build instruction

Mark-64:
Updated Windows build instructions as per discussion 
https://discuss.pixls.us/t/build-for-windows-help-please/12133/68

### DIFF
--- a/packaging/windows/BUILD.txt
+++ b/packaging/windows/BUILD.txt
@@ -6,25 +6,22 @@ A) Native compile using MSYS2:  How to make a darktable windows installer (64 bi
 
     Install MSYS2 (instructions and prerequisites can be found on official website: https://msys2.github.io/).
     Update the base MSYS2 system until no further updates are available using:
-		$ pacman -Syu
+	$ pacman -Syu
 
     From MSYS2 terminal, install x64 developer tools and x86_64 toolchain and git from the MSYS2 prompt:
-		$ pacman -S base-devel
-		$ pacman -S mingw-w64-x86_64-{toolchain,cmake,nsis}
-		$ pacman -S git
+	$ pacman -S base-devel
+	$ pacman -S mingw-w64-x86_64-{toolchain,cmake,nsis}
+	$ pacman -S git
 
     Install required libraries and dependencies:
-		$ pacman -S mingw-w64-x86_64-{exiv2,lcms2,lensfun,dbus-glib,openexr,sqlite3,libxslt,libsoup,libwebp,libsecret,lua,graphicsmagick,openjpeg2,gtk3,pugixml,libexif,osm-gps-map,libgphoto2,flickcurl,drmingw,gettext,python3,iso-codes,python3-jsonschema,python2-setuptools,python3-setuptools}
+	$ pacman -S mingw-w64-x86_64-{exiv2,lcms2,lensfun,dbus-glib,openexr,sqlite3,libxslt,libsoup,libwebp,libsecret,lua,graphicsmagick,openjpeg2,gtk3,pugixml,libexif,osm-gps-map,libgphoto2,flickcurl,drmingw,gettext,python3,iso-codes,python3-jsonschema,python3-setuptools}
 
     For gphoto2:
     You need to restart the MINGW64 or MSYS terminal to have CAMLIBS and IOLIBS environment variables  properly set for LIBGPHOTO are available. You can check them with:
-		$ echo $CAMLIBS
-		$ echo $IOLIBS
+	$ echo $CAMLIBS
+	$ echo $IOLIBS
 
-    From MINGW64 terminal, update your lensfun database:
-		$ lensfun-update-data
-	
-	NOTE: at the moment, the lensfun version provided by MSYS2 (0.3.95-1) is broken. If you get errors in the build process, manually downgrade lensfun to version 0.3.2-4:
+   	NOTE: at the moment, the lensfun version provided by MSYS2 (0.3.95-1) is broken. If you get errors in the build process, manually downgrade lensfun to version 0.3.2-4:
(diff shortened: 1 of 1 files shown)


peterbud:
Hi @Mark-64 - thanks for this revision!

I see many lines with only whitespace changes - I don't see any reason to change the whitespace. Could you pls revert to the original back on those lines pls and push again the changes?

packaging/windows/BUILD.txt
```diff
@@ -6,44 +6,61 @@ A) Native compile using MSYS2:  How to make a darktable windows installer (64 bi
...
-    $ pacman -S mingw-w64-x86_64-{toolchain,cmake,nsis}
-    $ pacman -S git
+    From MSYS2 terminal, install x64 developer tools and x86_64 toolchain and git from the MSYS2 prompt:
+		$ pacman -S base-devel
+		$ pacman -S mingw-w64-x86_64-{toolchain,cmake,nsis}
+		$ pacman -S git
 
     Install required libraries and dependencies:
-    $ pacman -S mingw-w64-x86_64-{exiv2,lcms2,lensfun,dbus-glib,openexr,sqlite3,libxslt,libsoup,libwebp,libsecret,lua,graphicsmagick,openjpeg2,gtk3,pugixml,libexif,osm-gps-map,libgphoto2,flickcurl,drmingw,gettext,python3,iso-codes}
+		$ pacman -S mingw-w64-x86_64-{exiv2,lcms2,lensfun,dbus-glib,openexr,sqlite3,libxslt,libsoup,libwebp,libsecret,lua,graphicsmagick,openjpeg2,gtk3,pugixml,libexif,osm-gps-map,libgphoto2,flickcurl,drmingw,gettext,python3,iso-codes,python3-jsonschema,python2-setuptools,python3-setuptools}
```
peterbud: No need for python2-setuptools: we don't install python 2
IMHO no need for python3-setuptools - can you elaborate why it is needed?

packaging/windows/BUILD.txt
```diff
@@ -6,44 +6,61 @@ A) Native compile using MSYS2:  How to make a darktable windows installer (64 bi
...
     You need to restart the MINGW64 or MSYS terminal to have CAMLIBS and IOLIBS environment variables  properly set for LIBGPHOTO are available. You can check them with:
-    $ echo $CAMLIBS
-    $ echo $IOLIBS
-
-    Update your lensfun database:
-    $ lensfun-update-data
+		$ echo $CAMLIBS
+		$ echo $IOLIBS
+
+    From MINGW64 terminal, update your lensfun database:
```
peterbud: This is not needed here. It's enough to update the lensfun database **after** you download lensfun manually and fix the folder as you describe a few lines below.

packaging/windows/BUILD.txt
Mark-64: Without python3-jsonschema there is a warning in the configuration
`Missing jsonschema, problems in noiseprofiles.json might go unnoticed`

With python3-jsonschema only there is an error in the build

```
[ 92%] Checking validity of noiseprofiles.json
Traceback (most recent call last):
  File "C:\msys64\mingw64\bin\jsonschema-script.py", line 6, in <module>
    from pkg_resources import load_entry_point
ModuleNotFoundError: No module named 'pkg_resources'
make[2]: *** [data/CMakeFiles/validate_noiseprofiles_json.dir/build.make:59: data/CMakeFiles/validate_noiseprofiles_json] Error 1
make[1]: *** [CMakeFiles/Makefile2:8021: data/CMakeFiles/validate_noiseprofiles_json.dir/all] Error 2
```

The missing module pkg_resources is provided by the package python3-setuptools.
The package python2-setuptools is not needed you are right.

packaging/windows/BUILD.txt
Mark-64: Correct, but sooner or later the lensfun issue will be fixed and there will be no need to downgrade.
By then we will remove the downgrade part and everything will be fine

Mark-64:
> 
> 
> Hi @Mark-64 - thanks for this revision!
> 
> I see many lines with only whitespace changes - I don't see any reason to change the whitespace. Could you pls revert to the original back on those lines pls and push again the changes?

That was only cosmetics, I notices that the lines 
$ <command>
were sometimes indented and sometimes not. I proposed to indent them all, but if you don't like it I will revert

Mark-64:
Hi peterbud, 
I implemented the changes you suggested, plus added a section on how to build a generic package

packaging/windows/BUILD.txt
peterbud: I see. Actually seems to be an issue with the packaging of python3-jsonschema, and needs to be fixed there. Anyhow leave then the python3-setuptools here for the time being.

peterbud:
@Mark-64 thanks for the update, I'm fine to merge

TurboGit:
So I'm merging. Thanks.

zyrill:
Hey guys - thanks for the awesome work on darktable. I just played around a little with compilation on Win10 64bit and had to change the following line in ./build/src/cmake_install.cmake to make it work:
file(INSTALL DESTINATION "${CMAKE_INSTALL_PREFIX}/share/lensfun/" TYPE DIRECTORY FILES "C:/msys64/mingw64/bin/../share/lensfun/version_1") -> ../share/lensfun/version_2

parafin:
version_2 I think means that you are building with lensfun 0.3.95 (or git), which shouldn't be used, see #2813

zyrill:
you are absolutely right, I did use 0.3.95 - didn't know about the broken dependency. Thanks for the hint. Nothing I can do about it sadly, my coding skills are insufficient to work on a new lensfun release. :(

PR #2465, opened 2019-04-27
expose/culling : enhance layout

AlicVB:
2 changes : 
- fix a problem with aspect ratio if you have an image distorted which don't have aspect ratio set (ex : import a portrait raw ; import an image with some history stack in the xmp, ...)
- center images vertically in each row. As a side effect, when you zoom in you have the full row height as zooming area


TurboGit:
Looks good! Thanks.

PR #2471, opened 2019-04-28
expose : refactor to limit recompute

AlicVB:
currently, the expose/culling layout (images sizes and places) are recompute on each expose call. That means each time one move the mouse. That's unnecessary, hence the refactoring.
@edgardoh : can you double check if I've done this right with culling ? I hope I didn't forgot to reset layout at some places...
Note that I didn't change (and/or understand) the layout logic...

### DIFF
--- a/src/views/lighttable.c
+++ b/src/views/lighttable.c
@@ -2197,7 +2197,8 @@ static int expose_expose(dt_view_t *self, cairo_t *cr, int32_t width, int32_t he
       params.full_surface_wd = &lib->fp_surf[i].width;
       params.full_surface_ht = &lib->fp_surf[i].height;
       params.full_surface_w_lock = &lib->fp_surf[i].w_lock;
-      if(lib->fp_surf[i].zoom_100 >= 1000.0f || lib->fp_surf[i].imgid != lib->slots[i].imgid)
+      if(lib->full_zoom > 1.0f
+         && (lib->fp_surf[i].zoom_100 >= 1000.0f || lib->fp_surf[i].imgid != lib->slots[i].imgid))
         lib->fp_surf[i].zoom_100
             = _preview_get_zoom100(lib->slots[i].width, lib->slots[i].height, lib->slots[i].imgid);
       params.full_zoom100 = lib->fp_surf[i].zoom_100;
@@ -2359,6 +2360,17 @@ static int expose_full_preview(dt_view_t *self, cairo_t *cr, int32_t width, int3
     }
   }
 
+  if(!lib->slots || lib->slots_count != 1 || lib->slots[0].imgid != lib->full_preview_id
+     || lib->slots[0].width != width || lib->slots[0].height != height)
+  {
+    _expose_destroy_slots(self);
+    lib->slots_count = 1;
+    lib->slots = calloc(lib->slots_count, sizeof(dt_layout_image_t));
+    lib->slots[0].imgid = lib->full_preview_id;
+    lib->slots[0].width = width;
+    lib->slots[0].height = height;
+  }
+
   dt_view_image_expose_t params = { 0 };
   params.image_over = &(lib->image_over);
   params.imgid = lib->full_preview_id;
@@ -2370,7 +2382,8 @@ static int expose_full_preview(dt_view_t *self, cairo_t *cr, int32_t width, int3
   params.zoom = 1;
   params.full_preview = TRUE;
   params.full_zoom = lib->full_zoom;
-  if(lib->fp_surf[0].zoom_100 >= 1000.0f || lib->fp_surf[0].imgid != lib->full_preview_id)
+  if(lib->full_zoom > 1.0f
+     && (lib->fp_surf[0].zoom_100 >= 1000.0f || lib->fp_surf[0].imgid != lib->full_preview_id))
     lib->fp_surf[0].zoom_100 = _preview_get_zoom100(width, height, lib->full_preview_id);
   params.full_zoom100 = lib->fp_surf[0].zoom_100;
   params.full_maxdx = &lib->fp_surf[0].max_dx;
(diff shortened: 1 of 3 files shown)


TurboGit:
Looks good to me. I'll let @edgardoh review and approve this before merging.

src/views/lighttable.c
```diff
@@ -315,6 +328,9 @@ static void check_layout(dt_view_t *self)
   // layout has changed, let restore panels
   dt_ui_restore_panels(darktable.gui->ui);
 
+  // make sure we reset expose layout
+  _expose_destroy_slots(self);
```
edgardoh: check_layout() is called from expose(), so slots are destroyed on each expose(), not sure if this is necessary or can be avoided.

src/views/lighttable.c
```diff
@@ -2523,6 +2563,8 @@ static gboolean go_pgup_key_accel_callback(GtkAccelGroup *accel_group, GObject *
     move_view(lib, DIRECTION_PGUP);
   else
   {
+    // reset culling layout
+    _expose_destroy_slots(self);
```
edgardoh: This is not used by culling.

src/views/lighttable.c
```diff
@@ -2545,6 +2587,8 @@ static gboolean go_pgdown_key_accel_callback(GtkAccelGroup *accel_group, GObject
   }
   else
   {
+    // reset culling layout
+    _expose_destroy_slots(self);
```
edgardoh: This is not used by culling.

src/views/lighttable.c
```diff
@@ -2723,6 +2767,9 @@ static void drag_and_drop_received(GtkWidget *widget, GdkDragContext *context, g
 // shitf the first select image by 1 with up direction
 static void shift_first_selected_image(dt_library_t *lib, const int up)
 {
+  // reset the culling layout
+  _expose_destroy_slots(darktable.view_manager->proxy.lighttable.view);
```
edgardoh: This is not used by culling.

edgardoh:
Just tested and works fine, the only issue I've found on culling is when I hide the side panels (with tab) the images are not re-drawn, if I move to the next image size/position is OK, if then I show the panels, again images are not re-drawn.

src/views/lighttable.c
AlicVB: just before, check_layout return if there's no change, so this is called only on layout change.

edgardoh: Didn't see that, all OK then.

src/views/lighttable.c
AlicVB: yes it is ;) afaik, it's called when you press arrow keys (see key_pressed)

src/views/lighttable.c
AlicVB: ok

src/views/lighttable.c
AlicVB: ok

AlicVB:
oh, forgot the panels change (in fact, the size change of the center area). Will look at this.
Thanks for the review...

edgardoh:
As a side note, and not related to this PR, I've noticed that scrolling on culling seems more slow now. Sometimes when scrolling to the next image it takes some time or don't scroll at all and I have to scroll again.
It seems that sometimes the image is recomputed before the scroll, but is not always the case, as other times I can see how the next image is recomputed, so I'm just checking here if you noticed the same (or similar).

src/views/lighttable.c
edgardoh: Ups! My bad!

AlicVB:
done.
@edgardoh : I'll check tomorrow

AlicVB:
@edgardoh : I don't see anything obvious here, even with heavy retouched image and fresh cache...
As it's not related to this PR, can you open a new issue so we can keep track of this ? and if you have time to find a step by step to reproduce or even better bissect the problem... Thanks

edgardoh:
@AlicVB , is not a matter of edits, I've run it with -d perf and the delay is there even if nothing is processed. I still can't get if the issue is that the cache takes too long to load, a lock conflict or something else entirely. I'll keep digging on this and open a new issue or PR if I find something more concrete.

AlicVB:
fix the mipmap "recompute" (in fact not really recomputed, just reloaded in the cache) discussed in #2486

edgardoh:
@AlicVB , just tested, the performance (or "lag") is fixed, but the recompute (or reload) is still present, both on edited or un-edited images.

To duplicate, go thorough a series of images in culling, close darktable, start it again and go thorough the same images, the image seems to be recomputed.

TurboGit:
In fact no, my patch is still needed. All this looks fine to me. I've tested it and all seems working as expected. Let's merge it to get more field testing. Thanks for all the work on this.

AxelG-DE:
@AlicVB
That lag @edgardoh described above when scrollimg, I still have that today (+1113) and -d perf does not really show it. 

Seems it depends on the size of fimroll (the one in my mind has 12k photos and by filters down to 850... Reverse scrolling is even worse.

AlicVB:
@AxelG-DE : I *think* I know where it comes from : `dt_collection_get_all` is called, and it go throught all the images, so time cost depends on the size of the collection.
I'll try to find a fix, maybe using the "in-memory" sqlite table...

AxelG-DE:
@AlicVB 
thanks, than I keep calm to avoid "panic" and we get flooded with issues.

a bit off-topic. Sometimes I do not get the announcement email that my name got mentioned. (like in that other PR, where edgardoh asked me to report back my findings). Anything I can do??? It is not in spam.

AlicVB:
@AxelG-DE : Can you test PR #2521 ? This *should* be better...

AxelG-DE:
@AlicVB 
First of all, it builds very slow. It uses merely one out of 8x2 cores. After `git checkout master` all 16 cores are used again.

I can confirm, it is faster, after the first scan or rebuild
It still has one strange behaviour. I have culling set to 4pics. When I press "W" or "Strg+W" it will display full screen, without side panels, as expected. Releasing the keys should go back to  4pics but it drops me to one pic (with side panels) first. After touching arrow-keys or scrolling with the mouse it displays 4pics again.

AxelG-DE:
now after `git checkout master` and build again, I am at +1124 and it also seems to be improved?! Or do I need to withdraw my judgement, that it is faster :smile:

AxelG-DE:
I am back on master
Yes PR #2521 is faster :+1:

Issue #2484, opened 2019-04-29
[Performance] darkroom much less performant,  dt-2.7 vs. dt-2.6.2


AxelG-DE:
**Describe the bug**
According to my observation, we do have a more than noticeable performance issue for darkroom in git-master vs. official 2.6.2, which makes it feel very _sticky_ on my side.

Please read this entirely, before you draw your conclusion. I think it is a bit complex and even I tried to be narrow, the report got a bit longer (also there is always a risk, I do make mistakes in English)

This bug report, from my point of view (and I might be wrong), tries to comprehend:
- [High CPU load even opencl is working #2404](https://github.com/darktable-org/darktable/issues/2404)
and
- [double preview triggered when zooming in #2064](https://github.com/darktable-org/darktable/issues/2064)

**To Reproduce**
1. Load this picture: https://pixls-discuss.s3.dualstack.us-east-1.amazonaws.com/original/3X/3/4/34f49707aff221f5795c8a18cf8ffc14a186cb2a.DNG
(play raw, red tulips by pphoto CC-BY-NC 4.0)
2. (pixelpipe merely has sharpen, rotation and basecurve)
3. Change WB to 2000, close dt
4. Start dt with the -d perf option
5. **Change WB from 2000 to 25000** (copy the relevant lines from log)
6. repeat 1-5 with tonecurve _Gamma 0.5_  and Zoom _from 200% to fit2screen_

- All those I have done and concluded in attached .xlsx file
- In coloumn "A" I put a formular like =D11-$D$11 in order to eliminate the offset and have a relative time
- In coloumn "B" I placed a formular like =B12-B11 in order to have the used time per step
- in the WB-Chart I have marked some obvious findings, where in the other sheets, for "efficiency" reasons, merely marked the most obvious one

From there you can see, that dt-2.7 (at least here) responds way slower in total, than 2.6.2, while the single actions, they seem to be faster.

The lag I "feel" (until screen gets updated), seems actually even harder than the values I found here... (see section _Expected behavior_)

**Expected behavior**
Being not slower, "laggier" than 2.6.2, ideally even faster, more responsive than 2.6.2

I don't want to misapply (correct word acc. to dict. ???), that @aurelienpierre just now stated in [#2064](https://github.com/darktable-org/darktable/issues/2064), for him, the 2.7 is way more responsive than 2.6.2
It might be, we find either something totally different (anything else might influence the UI) or a hardware related issue.

[...]

PR #2486, opened 2019-04-30
[WIP] scroll images is slower

edgardoh:
When scrolling through images in culling there's some delay, it seems to be introduced by _preview_get_zoom100()

I have partially removed that change and also added a prefetch on culling to avoid the refresh of the images.

### DIFF
--- a/src/views/darkroom.c
+++ b/src/views/darkroom.c
@@ -224,19 +224,19 @@ void expose(
   if(dev->image_status == DT_DEV_PIXELPIPE_DIRTY || dev->image_status == DT_DEV_PIXELPIPE_INVALID
      || dev->pipe->input_timestamp < dev->preview_pipe->input_timestamp)
   {
-    dev->image_timeout_handle = g_timeout_add(DT_DARKROOM_PROCESS_MAIN_TIMEOUT, G_SOURCE_FUNC(dt_dev_process_image), dev);
+    dt_dev_process_image(dev);
   }
 
   if(dev->preview_status == DT_DEV_PIXELPIPE_DIRTY || dev->preview_status == DT_DEV_PIXELPIPE_INVALID
      || dev->pipe->input_timestamp > dev->preview_pipe->input_timestamp)
   {
-    dev->image_timeout_handle = g_timeout_add(DT_DARKROOM_PROCESS_PREVIEW_TIMEOUT, G_SOURCE_FUNC(dt_dev_process_preview), dev);
+    dt_dev_process_preview(dev);
   }
 
   if(dev->preview2_status == DT_DEV_PIXELPIPE_DIRTY || dev->preview2_status == DT_DEV_PIXELPIPE_INVALID
      || dev->pipe->input_timestamp > dev->preview2_pipe->input_timestamp)
   {
-    dev->image_timeout_handle = g_timeout_add(DT_DARKROOM_PROCESS_PREVIEW2_TIMEOUT, G_SOURCE_FUNC(dt_dev_process_preview2), dev);
+    dt_dev_process_preview2(dev);
   }
 
   dt_pthread_mutex_t *mutex = NULL;
--- a/src/views/lighttable.c
+++ b/src/views/lighttable.c
@@ -173,6 +173,7 @@ typedef struct dt_library_t
   int32_t collection_count;
 
   int last_first_selected; // offset of the last selected image inside the current selection
+  gboolean culling_prefetched; // culling has already prefetch the current displayed images?
 
   // stuff for the audio player
   GPid audio_player_pid;   // the pid of the child process
@@ -277,6 +278,7 @@ static inline void filmstrip_set_active_image(dt_library_t *lib, const int imgid
   }
 
   lib->last_first_selected = offset;
+  lib->culling_prefetched = FALSE;
   dt_selection_select_single(darktable.selection, imgid);
   dt_view_filmstrip_set_active_image(darktable.view_manager, imgid);
   if(lib->full_preview_id > -1) lib->full_preview_id = imgid;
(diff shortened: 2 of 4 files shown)


edgardoh:
@AlicVB , I don't know what _preview_get_zoom100() does so take this just as a try. Also if you could review the prefetch to be sure it doesn't mess with other things.

I also made some changes to pixelpipe to continue with #2484.

AlicVB:
_preview_get_zoom100() is used to know the value of full resolution zoom value in advance. Inside this function, it's dt_image_get_final_size which can take time (although not really significant here). What I propose is :
1- to only call this on the first zoom change. That will delay a little bit the first zoom step, but shouldn't block scrolling anymore.
2- to track down what is taking time in dt_image_get_final_size.

For 2 : what images are you using ? are the images unaltered ? does -d perf show something useful ?
Thanks

edgardoh:
> _preview_get_zoom100() is used to know the value of full resolution zoom
> value in advance. Inside this function, it's dt_image_get_final_size which
> can take time (although not really significant here). What I propose is :
> 1- to only call this on the first zoom change. That will delay a little
> bit the first zoom step, but shouldn't block scrolling anymore.
>
Right now (on master)  _preview_get_zoom100() is called every time, even on
full preview, and in culling for each displayed image, even if not zoomed.
If I understand correctly it should be called only when zoomed in (and in
the first zoom?), but I have no idea how to do that.
This should be done in a separate PR anyway, I added it here just to
confirm what is causing the delay.

> 2- to track down what is taking time in dt_image_get_final_size.
>
> For 2 : what images are you using ? are the images unaltered ? does -d
> perf show something useful ?
>
Both edited and unaltered. I'll try to explain better what I'm doing here:
When scrolling, the images that are not in memory seem to be reprocessed,
first a kind of blurred image is displayed and shortly a sharper one is
displayed. This both on edited and un-edited images. What happens is that
the expose can't find a cache with the right size, so a smaller one is
used, then on a second pass, the right cache is retrieved.
What I'm doing is to prefetch the next and previous images, so when the
cache is requested the right size can be retrieved. This is how the file
manager works (it prefetch an entire row)

Thanks

AxelG-DE:
It seems to work fine for my case :+1:

AlicVB:
> Right now (on master) _preview_get_zoom100() is called every time

Yes, but compute only occurs at the first call, after that, final_width/height are store in the image structure and just returned without computing anything.


> but I have no idea how to do that.

I'll take care of that soon in a separate PR

> When scrolling, the images that are not in memory seem to be reprocessed,

Just noticed that too. It shouldn't.  (and hopefully, it's the root cause of our problem here)... I'll investigate here too

edgardoh:
I'll keep this on hold then, there's another active PR anyway so better
wait.

AlicVB:
@edgardoh : I've move the final size computation to only zoom cases in #2471 : can you test it ? (I've added it to this already pending PR, as I needed the changes done there)

I delay the first zoom step, but after some thought, I'm not really for reverting it completely, as it solves a lot of other bug (in mipmap sizes, with some special raws with small embedded preview, etc...)

I'll investigate of the mipmap recompute problem now (in a completely new PR I hope)

AlicVB:
ok, it was not a full recompe, just a reloading in the cache.
I've fixed it in #2471 too
Now I think, I'll stop here, before #2471 becomes a nightmare to review, and @TurboGit want to kill me ;)

If you agree with my changes there, I think this PR should be closed (maybe just open a new PR for the other changes you made)

TurboGit:
Ok, I'll close this PR and will review #2471 then. Thanks, and I'll try to refrain from killing someone, promised :)

edgardoh:
Why? We still don't know if this is needed, that's for @AlicVB  to say.

TurboGit:
@edgardoh : AlicVB says just above that we could close this PR if #2471 is merged :)

edgardoh:
@TurboGit  : @AlicVB says if I agree, and I do not. But even if he didn't say it, this is my PR and you didn't even ask me.

And in my last comment in #2471 I'm reporting an issue that still needs to be discussed.

TurboGit:
Ok, ok I reopen then. I did not understood.

AlicVB:
@edgardoh : I've focused on "my" part of the changes you've done (=fix the regressions due to my work), and completely forgot the prefetch part... my bad. That should enhance image scrolling indeed.

So, to resume, we should keep the prefetch part, and drop the other changes, right ? (correct me, if needed)
I'll take care of that in another PR, and let you close this one once you review it if you agree

edgardoh:
@AlicVB , OK, I'll create another another PR with the prefetch stuff, but let's keep this one open, @AxelG-DE  is running some test and we may need it if he find something.

@AxelG-DE , please leave a note here when you're done with this so we can close it.

AlicVB:
@edgardoh : in fact, I've already done it here https://github.com/AlicVB/darktable/commit/898d567980a1490c70d5230cf8545558b5709f32
(I've adapt it a little bit to the recent changes)
I don't know how to push something with adding another author name (but I know it's possible) so if the code is fine for you, just copy/paste it in your PR. if it can save some of your time...

edgardoh:
I didn't understand that you'll do it, but great, just push this one, never
mind the authors.

Thanks!

AlicVB:
@edgardoh : done in #2508 
and to be sure you keep working, I've this one for you : #2509 ;)

edgardoh:
> @edgardoh <https://github.com/edgardoh> : done in #2508
> <https://github.com/darktable-org/darktable/pull/2508>
>
Thanks!

>
> and to be sure you keep working, I've this one for you : #2509
> <https://github.com/darktable-org/darktable/issues/2509> ;)
>
Thanks? ;)

edgardoh:
Testing is done, so I'm closing this.

AxelG-DE:
@edgardoh @AlicVB 
Sorry, I didn't get the notification (as often) from above, where you asked me to report... 

Actually I did report on #2484. I guess you saw it. 

In general I think 
- that big lag in darkroom is solved
- culling is even today 4th May still slower than I wish, at least on bigger film rolls
- something suspicious on [opencl_synch_cashe](https://github.com/darktable-org/darktable/issues/2514)

PR #2487, opened 2019-04-30
Grey theme for new UI

Nilvus:
Here it is, the grey css theme for the new UI. As seen with @aurelienpierre, I've worked on this theme and tested it with few users who makes some feedback.

The idea is to have an interface who reduce optic illusions that could add a dark or bright one about exposure. Having a grey theme as an alternative as the dark default one let the user choose.

For optic illusions about grey and shadows (so proximity of dark zones and the perception of colors), one of the best known experience is the checker shadows illusion from Adelson. See :
https://en.wikipedia.org/wiki/Checker_shadow_illusion
https://www.deviantart.com/butisit/art/Checker-shadow-illusion-263331875

I've tried to propose the better compromise between reducing optic illusions, having a good contrast between elements of the interface, and having a good design. That was not so easy and had need some few hours to find that.

Capture-20190430212952-1920x1052

Capture-20190430213043-1920x1052

### DIFF
--- a/data/darktable-elegant-grey.css
+++ b/data/darktable-elegant-grey.css
@@ -1,6 +1,6 @@
 /*
     This file is part of darktable,
-    copyright (c) 2019 Aurélien Pierre.
+    copyright (c) 2019 Nicolas Auffray
 
     darktable is free software: you can redistribute it and/or modify
     it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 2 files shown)


jpg54:
Great job, thanks for sharing and work, Nico.

Nilvus:
Just add a darktable icons grey css

TurboGit:
But those are not installed by default.

I'll add the necessary cmake code.

TurboGit:
Thanks.

Nilvus:
> Thanks.

And thanks to you for updating cmake code. I don't anything about that !

TurboGit:
No problem, and now all new CSS added into data/themes will be installed automatically.

PR #2493, opened 2019-05-01
darkroom : fix gtk errors with some iop

AlicVB:
Avoid gtk_criticals with scalepixels and rotatepixels iops when entering darkroom.


TurboGit:
Looks good. Thanks.

Issue #2503, opened 2019-05-02
Import dialog: CSS styling missing from titlebar


garrett:
**Describe the bug**
The new CSS theme is missing support for dialogs with actions (other than close).

**To Reproduce**
1. Go to Lighttable
2. Click on Import (either image or folder)
3. See error

**Expected behavior**
The dialog should have a titlebar and frame.

**Screenshots**
![darktable-import](https://user-images.githubusercontent.com/10246/57080520-fb914d80-6cf2-11e9-8ef6-9be362ff0b3f.png)

As the style is missing from the header bar, it's seen as transparent.

**Platform (please complete the following information):**
 - OS: Fedora 29
 - Darktable version: master (b9d32f687f189d0415a73c33a4a5503fe8fd6797)

Nilvus:
It seems that you're using grey css. I forgot to verify this dialog but what is strange is don't have same dialog. See what I have with the new grey css I made (I think anyway there one or two things to improve) :

![Capture-20190502173212-978x593](https://user-images.githubusercontent.com/45535283/57087564-9c3a3a00-6d00-11e9-84b7-869eef7a668d.png)

For your problem, which desktop environment are you using (name and version) and which theme for this desktop environment are you using (theme that display title bar of course) ?

Nilvus:
Euh, something strange. On my computer, title bar is black not transparent but here it is... It seems that it is transparent by default but why on my computer title bar is black, I don't know. I'm note sure it is a css issue, maybe a gtk one as the default black css render same title bar.
@aurelienpierre or @TurboGit if you have some tips about that ?

Anyway, I will see if I find something on css.

Nilvus:
@garrett : please this PR #2504 to see if it correct your issue.

garrett:
@Nilvus: Thanks for the PR! I made a suggestion on how the padding could be improved for the sides of the titlebar.

(FWIW: I happen to be a longtime FOSS designer & developer who knows a bit about CSS and such, in addition to a photographer and Darktable user. But it's just a very simple change, so that doesn't really matter much. :wink:)

Nilvus:
@garrett : so could you please close this issue and continue on the PR. I will finish to correct that today.

garrett:
@Nilvus: Great! Thanks!

PR #2504, opened 2019-05-02
CSS improvments for dialog headerbar and make separator transparent in grey css theme

Nilvus:
This correct issue #2503 for both themes. I've also make separator transparent for more consistent UI on grey theme, especially on dialog windows. Now, dialog will look like this :

dark theme :
Capture-20190502185323-978x593

grey theme :

Capture-20190502185235-978x593


Nilvus:
@TurboGit : just wait to merge. I just see how to have a more better titlebar and will put a new commit.

Nilvus:
Commit merge. I've update screens captures. @aurelienpierre as I've change default css, please say if it's ok for you.

TurboGit:
Sure I'm marking it wip.

data/themes/darktable.css
```diff
@@ -908,6 +908,13 @@ dialog *
   padding:0;
 }
 
+dialog headerbar
+{
+  background-color: @border_color;
+  margin: 0;
+  padding: 1pt;
```
garrett: Adding 1pt for padding works for the top and bottom of the header. However, there isn't enough padding for the sides. Looking elsewhere in the CSS file, I see the value of 0.5em used for element padding, so I'd suggest the following:

```suggestion
  padding: 1pt 0.5em;
```

There is still the issue with the buttons in the header having no padding for them, but that's (probably) a different bug / different PR, unless you think it's suitable here too.

Nilvus: I've replace your suggestion with :
`padding: 0.2em 0.6em`
That seems better for alignement and spacing on my config.

Nilvus: I will commit it if it's ok for you :

Capture-20190504155748-994x86

garrett: Yeah, that looks good from what I see in the screenshot. :+1:

I was wondering why `pt` was used, as it doesn't scale* with the font, unlike `em`, which is tied to the font size. Good to see it in `em` here. Perhaps it makes sense to review & generalize the units used across the CSS, possibly with variables — but that's a different conversation outside the scope of this PR. :wink:

(* At least this is with web CSS — and I'm guessing it's probably the case with GTK CSS too?)

Nilvus: I don't know why pt was used as it was the original work of @aurelienpierre.

I also improve buttons just now with this add :

`
dialog headerbar button
{
  padding: 1px 2px;
}
`
If it concerns improvement css theme, I think this PR could be also for other concerns. If needed I could change title.

I've think about change all scales and delete pt but it could also break something. I don't know, I will check more. W3C recommands use of em and px for screen viewing.

With that said, I'm not for changing units in px by now as it works. Using em could break something as users could use another font for example.

TurboGit:
This looks ready to me now, no?

Nilvus:
> This looks ready to me now, no?

For me too, I just don't have feedback from garret or Aurélien. If they want improvment, they could answer anyway. I'm ok if you want to merge this.

TurboGit:
Thanks, merged!

garrett:
Thanks again for the fix! I was AFK yesterday, but the headerbar is there and looks good to me! :+1:

PR #2508, opened 2019-05-02
culling : prefetch next/previous images

AlicVB:
original code from @edgardoh in #2486


TurboGit:
Looks ok to me. Thanks.

PR #2511, opened 2019-05-03
expose : fix image order

### LINKED ISSUE #2509: culling order of images
not sure if it's a bug or a feature, but when you scroll images in culling layout, it doesn't follow the filemanager order.
if in you filemanager, you see images 1,2,3,4,5,... you select image 2, enter culling mode (4 images), you see : 2,3,4,5 (expected) but when you scroll up, you get 6,3,4,5 and then 6,4,5,7... If it's intended, then I don't understand...
I just hope I've not broken something...

AlicVB:
this fix #2509
technically, the problem comes from `ORDER BY INSTR('%s', id)` where %s is the sorted list of the ids, which can be something like `105, 106, 1, 2` and here, INSTR, will obviously not do the right sort...
I've replace that part with something like `ORDER BY id=105 DESC, id=106 DESC, id=1 DESC, id=2 DESC`

### DIFF
--- a/src/views/lighttable.c
+++ b/src/views/lighttable.c
@@ -1860,6 +1860,7 @@ static gboolean _expose_compute_slots(dt_view_t *self, int32_t width, int32_t he
   if(img_count == 0) return FALSE;
 
   gchar *imgids = NULL;
+  gchar *sortids = NULL;
 
   // build the image ids for the SQL 'in' where clause.
   if(layout == DT_LIGHTTABLE_LAYOUT_EXPOSE)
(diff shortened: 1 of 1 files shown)


TurboGit:
What about the performances here? Have you tested with a large collection?

AlicVB:
on the code level, I don't think it can be noticeable, the full query should look like this : `SELECT id, aspect_ratio FROM images WHERE id IN (105, 106, 1, 2) ORDER BY id=105 DESC, id=106 DESC, id=1 DESC, id=2 DESC` where the number of ids shouldn't be too big, as it's the number of image exposed...
Anyway, I'll try with larger collection.

AlicVB:
@TurboGit : this doesn't regress performance in my tests.
With a "large" collection (37K) there's some benchmark (<0.1 sec) changes only when you expose images with "big" ids. After some more test, without any ordering, it's the same, so I imagine it's just the where clause which take longer... and we can't do anything here, I fear (at least, I've not touch the where clause)

TurboGit:
Perfect then! Thanks.

PR #2521, opened 2019-05-05
expose/culling : limit call to database

AlicVB:
avoid going through all entries to speedup the slots creation. In particular `dt_collection_get_all` with -1 parameter, can take time with large collection. I have switch to using the "in-memory" sqlite table with the help of `LIMIT` and so on.
This *should* enhance loading times for expose/culling slots. In my benchmarks there's a gain of about 1 sec with 27K images (but I have a quite good config)
@edgardoh : as usual, can you have a look/test. This as become a quite big rewrite, so I hope I've not break anything (or forgot some feature)

### DIFF
--- a/src/views/lighttable.c
+++ b/src/views/lighttable.c
@@ -103,6 +103,8 @@ static gboolean _expose_again_full(gpointer user_data);
 
 static void _force_expose_all(dt_view_t *self);
 
+static gboolean _expose_recreate_slots(dt_view_t *self, const dt_lighttable_layout_t layout);
+
 typedef struct dt_preview_surface_t
 {
   int mip;
@@ -178,8 +180,6 @@ typedef struct dt_library_t
 
   int32_t collection_count;
 
-  int last_first_selected; // offset of the last selected image inside the current selection
-
   // stuff for the audio player
   GPid audio_player_pid;   // the pid of the child process
   int32_t audio_player_id; // the imgid of the image the audio is played for
@@ -193,6 +193,7 @@ typedef struct dt_library_t
   dt_preview_surface_t fp_surf[FULL_PREVIEW_IN_MEMORY_LIMIT];
   dt_layout_image_t *slots;
   int slots_count;
+  gboolean slots_changed;
   int last_num_images, last_width, last_height;
 
   /* prepared and reusable statements */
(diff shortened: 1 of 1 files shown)


AxelG-DE:
Tested. 
Works for me.
THank you very much!

edgardoh:
@AlicVB , sorry but I don't have the time right now to do a test. I don't think I get what you are doing here and maybe you already considered this, but what if we store the collected images list in memory when the collection changes? The only downside will be that on large collection it can be resources consuming.

AlicVB:
@edgardoh : no problem, we aren't in a hurry !
What you propose is already in place in a different form : on each collection change, a new sqlite table (memory.collected_images) is stored. So we have just to reuse it. Then, we can play with `JOIN` and `LIMIT` to only get what we want. That's basically what I've done...

edgardoh:
Well, now I'm sure I don't understand what you did!

My idea is to modify only dt_collection_get_all(), if the in-memory list is empty it will perform the select as it does now, if not empty it returns a new list but using the in-memory list as a source. This way we don't need to modify the existing code. Can this be done with the memory.collected_images? Or maybe you are introducing other optimizations?

AlicVB:
The main idea is that if we use in-memory database, we can do everything in one pass. for example, here is the steps for culling : 
1- get the first image (as before, we only reuse slots values, instead of last_first_selected)
2- `SELECT m.imgid, b.aspect_ratio FROM memory.collected_images AS m, images AS b WHERE m.imgid = b.id AND m.rowid >= (SELECT rowid FROM memory.collected_images WHERE imgid = %d) ORDER BY m.rowid LIMIT %d`
3- we get all what we want (and not more) directly !
Apart from that, I've done some other optimizations, mainly resulting as the above changes.

edgardoh:
OK, I just took a quick look at your code, so there's a good chance that what I say makes no sense. If you think this is faster than a list I'm OK with it.

AlicVB:
I think so, because, we don't have to construct/verify/... the list, and as said because getting the ids can be done in 1 pass (and sqlite is quite performant for query stuff, especially with in-memory tables)
...and it simplify the code at the same time !

AxelG-DE:
@AlicVB

I don't know, if it is related, but would you mind looking into my comment of PR #2471

Thanks

AlicVB:
@AxelG-DE : done. thanks for the report (for the record, the problem comes from one of my previous PR, not this one)

TurboGit:
Better speed is ok, the changes looks a bit complex to me. Anyway, this works fine on my side so I'll merge now for more field testing as it is difficult to assess if this is 100% correct just be review. Thanks.

TurboGit:
Can you just fix the French word "avec" in the first commit log message?

AlicVB:
@TurboGit : done :)

TurboGit:
Thanks.

AxelG-DE:
@AlicVB
Thank you!

Nilvus:
Great works @AlicVB. I've never see images display so fast. Really faster. Thanks a lot ! I've passed about 900 images without any problem  By now, all is 100% perfect.

AxelG-DE:
Need to correct my self, my collection is 12,000 pics in one folder, filtered to 850 :-)

PR #2529, opened 2019-05-06
culling: fix prefetch

edgardoh:
In culling images must be prefetched each time the displayed images change.

### DIFF
--- a/src/views/lighttable.c
+++ b/src/views/lighttable.c
@@ -180,6 +180,8 @@ typedef struct dt_library_t
 
   int32_t collection_count;
 
+  gboolean culling_prefetched; // culling has already prefetch the current displayed images?
+
   // stuff for the audio player
   GPid audio_player_pid;   // the pid of the child process
   int32_t audio_player_id; // the imgid of the image the audio is played for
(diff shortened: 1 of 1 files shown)


src/views/lighttable.c
```diff
@@ -180,6 +180,8 @@ typedef struct dt_library_t
 
   int32_t collection_count;
 
+  gboolean culling_prefetched; // culling has already prefetch the current displayed images?
+
```
AlicVB: I don't think this is needed, I've added `slots_changed` params, which indicate when the images shown have changed. It serve the same thing.
Note that I may have forgot places where to set it to false...

edgardoh:
@AlicVB  , it should be like this?

  gboolean prefetch = FALSE;
  if(lib->last_width != width || lib->last_height != height || !lib->slots
     || lib->last_num_images != get_display_num_images())
  {
    if(!_expose_recreate_slots(self, layout)) return 0;
  }
  if(lib->slots_changed)
  {
    if(!_expose_compute_slots(self, width, height, layout)) return 0;
    lib->slots_changed = FALSE;
    if(layout == DT_LIGHTTABLE_LAYOUT_CULLING) prefetch = TRUE;
  }

edgardoh:
This will be fixed on a different PR, I'm closing this one.

PR #2530, opened 2019-05-07
Culling prefetch & selection fixes

AlicVB:
- fix prefetch which doesn't happen
- reuse previous/next image infos collected in prefetch for moving (less database call)
- ensure we always show the requested nb of image in collection end
- ensure the selection follow culling images when returning to filemanager and in filmstrip
- ensure the last culling images are visible when returning to filemanager

### DIFF
--- a/src/views/lighttable.c
+++ b/src/views/lighttable.c
@@ -2205,6 +2205,14 @@ static int expose_expose(dt_view_t *self, cairo_t *cr, int32_t width, int32_t he
 
   for(int i = 0; i < lib->slots_count; i++)
   {
+    // set mouse over id
+    if(pointerx > lib->slots[i].x && pointerx < lib->slots[i].x + lib->slots[i].width && pointery > lib->slots[i].y
+       && pointery < lib->slots[i].y + lib->slots[i].height)
+    {
+      mouse_over_id = lib->slots[i].imgid;
+      dt_control_set_mouse_over_id(mouse_over_id);
+    }
+
     cairo_save(cr);
     // if(zoom == 1) dt_image_prefetch(image, DT_IMAGE_MIPF);
     cairo_translate(cr, lib->slots[i].x, lib->slots[i].y);
@@ -2250,14 +2258,6 @@ static int expose_expose(dt_view_t *self, cairo_t *cr, int32_t width, int32_t he
 
     missing += dt_view_image_expose(&params);
     cairo_restore(cr);
-
-    // set mouse over id
-    if(pointerx > lib->slots[i].x && pointerx < lib->slots[i].x + lib->slots[i].width && pointery > lib->slots[i].y
-       && pointery < lib->slots[i].y + lib->slots[i].height)
-    {
-      mouse_over_id = lib->slots[i].imgid;
-      dt_control_set_mouse_over_id(mouse_over_id);
-    }
   }
 
   // if needed, we prefetch the next and previous images
--- a/src/views/lighttable.c
+++ b/src/views/lighttable.c
@@ -127,6 +127,7 @@ typedef struct dt_layout_image_t
 {
   gint imgid;
   gint width, height, x, y;
+  double aspect_ratio;
 } dt_layout_image_t;
 
 /**
@@ -195,6 +196,7 @@ typedef struct dt_library_t
   int slots_count, slots_count_old;
   gboolean slots_changed;
   dt_layout_image_t culling_previous, culling_next;
+  gboolean select_desactivate;
   int last_num_images, last_width, last_height;
 
   /* prepared and reusable statements */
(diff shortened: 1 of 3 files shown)


edgardoh:
Is working great for me. The only thing about having all the images selected on culling is that if we apply some rating there's a chance that it will be applied to all images, but that's only if the mouse if not over an image, so I'm not sure that having just the first selected is better.

src/views/lighttable.c
```diff
@@ -1879,12 +1907,21 @@ static gboolean _expose_compute_slots(dt_view_t *self, int32_t width, int32_t he
 
   GList *slots = NULL;
 
+  // reinit size and positions
+  int i = 0;
```
TurboGit: to be remove

src/views/lighttable.c
```diff
@@ -1879,12 +1907,21 @@ static gboolean _expose_compute_slots(dt_view_t *self, int32_t width, int32_t he
...
+  for(i = 0; i < lib->slots_count; i++)
```
TurboGit: for(int i = 0; ...

src/views/lighttable.c
```diff
@@ -1879,12 +1907,21 @@ static gboolean _expose_compute_slots(dt_view_t *self, int32_t width, int32_t he
...
+    lib->slots[i].x = lib->slots[i].y = 0;
+  }
+
   unsigned int total_width = 0, total_height = 0;
   int distance = 1;
   float avg_ratio = 0;
 
   // Get total window width and max window width/height
-  int i = 0;
   for(i = 0; i < lib->slots_count; i++)
```
TurboGit: for(int i = 0; ...

src/views/lighttable.c
```diff
@@ -2062,62 +2118,90 @@ static void _culling_prefetch(dt_view_t *self)
...
-    dt_mipmap_size_t mip = dt_mipmap_cache_get_matching_size(darktable.mipmap_cache,
-                                                             imgwd * lib->slots[lib->slots_count - 1].width * fz,
-                                                             imgwd * lib->slots[lib->slots_count - 1].height * fz);
-
-    if(mip < DT_MIPMAP_8)
-      dt_mipmap_cache_get(darktable.mipmap_cache, NULL, imgid_next, mip, DT_MIPMAP_PREFETCH, 'r');
-  }
+    gchar *query
+        = dt_util_dstrcat(NULL,
+                          "SELECT m.imgid, b.aspect_ratio FROM memory.collected_images AS m, images AS b WHERE "
```
TurboGit: I understand that this was the case in this code, but I think we should try to have better readability for the SQL commands. I'm doing:

```
   "SELECT ..."
   " FROM ..."
   " WHERE ..."
   " ORDER_BY .."
```

Of course if the FROM is small it can be on the same line as SELECT, but having the others starting on a new line makes things lot more readable.

src/views/lighttable.c
```diff
@@ -2062,62 +2118,90 @@ static void _culling_prefetch(dt_view_t *self)
...
+                          "m.rowid > (SELECT rowid FROM "
+                          "memory.collected_images WHERE imgid = %d) AND m.imgid = b.id ORDER BY m.rowid LIMIT 1",
+                          lib->slots[lib->slots_count - 1].imgid);
+    sqlite3_stmt *stmt;
+    DT_DEBUG_SQLITE3_PREPARE_V2(dt_database_get(darktable.db), query, -1, &stmt, NULL);
+    if(stmt != NULL)
+    {
+      if(sqlite3_step(stmt) == SQLITE_ROW)
```
TurboGit: we have a very similar block above but using culling_prev instead of culling_next. Maybe it is possible to factorize this?

src/views/lighttable.c
AlicVB: yes, I didn't like that too, but it was here before... time to remove !

src/views/lighttable.c
AlicVB: done (in other queries too)

src/views/lighttable.c
AlicVB: done

AlicVB:
I've fix your comments. Thanks
I've added another minor change. If you prefer, I can drop it and do another PR once this one is merged (I don't want to put mess in your review process)

TurboGit:
In this last version I cannot change images in the culling mode. The mouse scroll does not work anymore.

TurboGit:
That's fine for the new feature. Let's keep it there even though it would have been better in a new PR :)

TurboGit:
Just checked the scroll issue is a regression on this PR.

AlicVB:
should be fixed, obvious mistakes from my one on the last factorizing...

TurboGit:
Thanks all good now.

PR #2532, opened 2019-05-08
Spanish user manual translation update

EdgarLux:
Updating and correcting some typos


TurboGit:
Thanks.

EdgarLux:
Bonjour @TurboGit! I have noticed that since this update, there is no pdf online or at https://www.darktable.org/resources/
Could it be added before continuing to update it for version 3.0?

TurboGit:
@EdgarLux : The manuals are built only when the release is done. This is because it is not a trivial work and so we do not update the doc regularly. If at some point I'm able to have a script doing all the work for me it may change :)

EdgarLux:
Ok, I understand. So, for the next version 3.0, when should the manual be ready? after or before the final release?

PR #2534, opened 2019-05-08
3D LUT module - Added GMIC compressed LUT

phweyland:
This PR includes the following topics:
- Fixed an issue where the clut was calculated as many times as commit() was called. This is now done in filename_callback().
- Added read GMIC compressed LUT using the GMIC library
- Added cache for compressed LUT ((user cache)/gmic) to speed up the compressed LUT. The cache is shared with G'MIC app).

https://discuss.pixls.us/t/3d-lut-module-in-darktable-2-7-dev/12341
https://discuss.pixls.us/t/clut-compression/11752


TurboGit:
There is conflicts and the rebase on master first commit is suspicious as a rebase do not add commits, that's the whole point of a rebase :)

phweyland:
> the rebase on master first commit is suspicious as a rebase do not add commits

You mean that commit should have told me : "nothing to commit" after rebase, correct ?

What about doing the following ?
- rename my local branch to lut3d.old
- git fetch upstream
- git checkout master
- git rebase upstream/master
- git commit -a -m "rebase master" => should refuse "nothing to commit"
- checkout -b lut3d
- copy the last state files from lut3d.old over lut3d
- git commit -a -m "Added compressed LUTs"
- git push origin lut3d --force

Would that work ?

TurboGit:
Should work, yes.

TurboGit:
What is the status on this? Is this ready for review?

phweyland:
> 
> 
> What is the status on this? Is this ready for review?

Yes, please.

src/iop/lut3d.c
```diff
@@ -1048,7 +1144,7 @@ void gui_init(dt_iop_module_t *self)
   dt_bauhaus_combobox_add(g->interpolation, _("trilinear"));
   dt_bauhaus_combobox_add(g->interpolation, _("pyramid"));
   gtk_box_pack_start(GTK_BOX(self->widget), GTK_WIDGET(g->interpolation) , TRUE, TRUE, 0);
-  gtk_widget_set_tooltip_text(g->interpolation, _("select the interpolation method"));
+  gtk_widget_set_tooltip_text(g->interpolation, _("select the interpolation method\n"));
```
TurboGit: please no \n on tooltip. if space is needed this should be done in CSS for all tooltip anyway.

src/iop/lut3d.c
```diff
@@ -1036,10 +1133,9 @@ void gui_init(dt_iop_module_t *self)
   dt_bauhaus_widget_set_label(g->colorspace, NULL, _("application color space"));
   dt_bauhaus_combobox_add(g->colorspace, _("sRGB"));
   dt_bauhaus_combobox_add(g->colorspace, _("REC.709"));
-  dt_bauhaus_combobox_add(g->colorspace, _("lin sRGB/REC.709"));
-  dt_bauhaus_combobox_add(g->colorspace, _("lin prophoto RGB"));
+  dt_bauhaus_combobox_add(g->colorspace, _("linear REC.709"));
   gtk_box_pack_start(GTK_BOX(self->widget), GTK_WIDGET(g->colorspace) , TRUE, TRUE, 0);
-  gtk_widget_set_tooltip_text(g->colorspace, _("select the color space in which the LUT has to be applied"));
+  gtk_widget_set_tooltip_text(g->colorspace, _("select the color space in which the LUT has to be applied\n"));
```
TurboGit: please no \n on tooltip. if space is needed this should be done in CSS for all tooltip anyway.

TurboGit:
I'm not sure to understand, I thought the idea was to put the compressed lut into the params itself as when compressed the size was "small". This would avoid the need for a cache of compressed lut. It will also mean that we can have style and preset with the lut embedded and so exchange them easily.

What is the size of a compressed lut?

phweyland:
> What is the size of a compressed lut?

image
instead of:
image

The max size of a compressed lut is 2046x2x3 = 12 288.
The first advantage of this is clearly to facilitate the exchange and the storage of these luts. 
This sparked my interest because it may enable there inclusion into params of the image. This not yet done, but if there is an agreement on this we can do (+13k in params).
I have added the cache mechanism because the decompression introduce a visible delay (1/2 or 1 second). The cache avoids this, even if we'd include the compressed lut in params.

TurboGit:
> This sparked my interest because it may enable there inclusion into params of the image. This not yet done, but if there is an agreement on this we can do (+13k in params).

I would love to see that. 13k is somewhat "big" for params but not something that would not work I think. So yes please, let's try this. Thanks.

phweyland:
> So yes please, let's try this.

OK, I can work on it.
The not compressed luts will not be put in params. We could call GMIC to compress it but the compression process can last some minutes.
So it would be great to have a mechanism to keep the current params size for not compressed lut and have the 13k params size for compressed luts. Do you think that's possible ?
If possible, we could also adjust the params size to the actual compressed lut size (sometimes below 1k).

phweyland:
I'm not too far to have this done. But I had to revert the change below which produce which jeopardizes the history and styles.

>     * Fixed an issue where the clut was calculated as many times as commit() was called. This is now done in filename_callback().

Then I'm back with 2 clut calculations  (in fact on each commit) every time I change the clut. Is there a way to over come this ?

phweyland:
Compressed lut added to params.
I've checked that the original compressed clut file can be wiped. The image keeps its transformation.
Style seems to work too.

I don't find solution to avoid double calculation of the clut as long as this calculation is done in commit().
It should be done in filepath_callback(). Where should I save the calculated clut pointer to make it available in commit() ? 

Can I use dt_iop_lut3d_global_data_t to pass it from file_path_callback() to commit() ? If yes, how can I clean it up properly ?

'm also wondering why sometime a change on the module add an history step and sometime not. I ask the question because that impacts the performance.
Once the history compressed the steps are not added any more.

image

I've let on purpose the printf() for on-going tests.

src/iop/lut3d.c
phweyland: done

src/iop/lut3d.c
phweyland: done

TurboGit:
> I'm also wondering why sometime a change on the module adds an history step and sometime not. I ask the question because that impacts the performance.
> Once the history is compressed the steps are not added any more.

This is a bug in dt since a long time. I never found the root cause but for sure this is not an issue with this PR.

TurboGit:
> Can I use dt_iop_lut3d_global_data_t or dt_iop_lut3d_gui_data_t to pass it from file_path_callback() to commit() ?

Why not in dt_iop_lut3d_data_t as done in watermark for example?

phweyland:
> Why not in dt_iop_lut3d_data_t as done in watermark for example?

As far as I understand dt_iop_watermark_data_t or dt_iop_lut3d_data_t are accessed from piece. The callback routines get only the widget and dt_iop_module_t...

Could you pinpoint the example you think of ?
I'm trying to understand the pixelpipe (pixelpipe_hb.c) ... any description or explanation would be welcome. :-)

TurboGit:
Sorry I may be missing something but there is no file_path_callback in your PR. But I would say that you should pass the compressed lut the very same way you used to pass the filename. Won't that work? Where was this? In the GUI data? So let's go with that.

phweyland:
> But I would say that you should pass the compressed lut the very same way you used to pass the filename

No, because the filename is part of params but not the calculated lut. That would work if I included the full calculated lut into params.
Then there is another commit() case (beside the filepath_callback()). It is at the initialisation, just after init_pipe(). So I need to trig the lut calculation without filepath_callback().

edgardoh:
CImg.h should go into scr/external. Maybe a submodule?

BTW, if we ship CImg.h, building also GMIC is just one step further.

phweyland:
> I don't find solution to avoid double calculation of the clut as long as this calculation is done in commit().

The calculated clut, volatile data, from compressed clut or from whatever file, is done currently in commit(). But commit() is called several times for the same filepath (per type of pipe, I think). That was my concern. Could be done in process() with the same behaviour.

In fact, to solve the issue I think it would be necessary to have access to history sets of data. In the following case :
image
dt needs to remember the data set for each step (here #2 and #4), in fact a copy of params. But this done behind the scene, and we cannot associate a calculated clut to a specific step.

If I'm right, I think we cannot do better and we have to keep the current code.

phweyland:
> CImg.h should go into scr/external.

OK, I'll do it.

> Maybe a submodule?

What does that mean ? Would you have some example I could try to follow ?

> BTW, if we ship CImg.h, building also GMIC is just one step further.

OK, but what would be the advantage ?

On my side, the question is about the size of this lib. 5MB is a bit high. Is it worth to try to eliminate what is not used ? Is there any tool to help on this ?

edgardoh:
> CImg.h should go into scr/external.
>
> OK, I'll do it.
>
> Maybe a submodule?
>
> What does that mean ? Would you have some example I could try to follow ?
>
A Git submodule, like is done with rawspeed. But this is literally a
question, I'm not sure what's best.

BTW, if we ship CImg.h, building also GMIC is just one step further.
>
> OK, but what would be the advantage ?
>
> On my side, the question is about the size of this lib. 5MB is a bit high.
> Is it worth to try to eliminate what is not used ? Is there any tool to
> help on this ?
>
This is related to #2557, where I'm having troubles with the integration.
When I saw GMIC here I thought you were doing the same, Don't pay attention
to it, I'll continue the discussion on my PR so I don't add noise here.

Regarding the size of CImg.h, probably this can be done using libgmic, so
CImg.h won't be needed. But I don't kow if there's any advantage on that.

TurboGit:
Wait ! I'm not ok with that for the same reasons given to @edgardoh. We don't want to integrate GMic sources here I would say. The CImg.h is part of cimg-dev package. We want this external library to be installed or then we disable the module. As for the CUPS for the print module or the lua lib for the dt lua support/

Si I would add a check for this library to be available and if not then lut3d won't be built.

Integrating CImg.h means that we will need to upgrade it from time to time or loose possible bug fixes or enhancement...

phweyland:
> Integrating CImg.h means that we will need to upgrade it from time to time

Rawspeed, even as a library, needs to be updated time to time. But I agree that this kind of library is better, especially if there are several modules using the same resources.
I can GMIC as a submodule, build the library manually, but I don't know yet how to integrate it in the global build, neither how to deal with the new cimg.h to link lut3d to the library ...

matze:
> Si I would add a check for this library to be available and if not then lut3d won't be built.

Why not make availability of cimg a compile time define and just do not support compressed LUTs if not available but keep compiling the "old" code loading un-compressed LUTs?

phweyland:
>  just do not support compressed LUTs if not available but keep compiling the "old" code loading un-compressed LUTs?

That's also a possibility, better that remove the module. Then the module would have to manage the possible regression of some images already treated with compressed lut in a new version of dt without cimg.
As far as I understand I would prefer to have cimg as a submodule as we have already rawspeed. Any guidance on how to build that is welcome...

TurboGit:
But for rawspeed this is because without rawspeed there is no dt, this module is mandatory so we have a submodule.

For the CImg this is not the same. Nothing mandatory and if it turn out to be difficult to build on Windows for example then we just can drop Lut-3D module. It is a nice module but nothing mandatory. We could also at some point need a light weight version of dt without all modules, having a way to build without such or such dependencies will be handy then.

phweyland:
I've followed the example given by the @rabauke gmic branch.
If it seems I'm able to generate libgmic and to use it in the dt build under msys2.
But <gmic.h> is not enough, and as in @rabauke example, the compilation fails if I remove the reference to CImg.h library. So no progress so far.

phweyland:
> The CImg.h is part of cimg-dev package.

If I understand properly, unless @david_tschumperle offers another gmic.h file including CImg interface (or equivalent) the libgmic library is useless for that PR. And CImg.h file is necessary, whatever the OS.

see example `foo3.cpp` in https://gmic.eu/libgmic.shtml

edgardoh:
There's probably a gmic command to load the lut, then the you can read it as a CImage. But if you are using gmic you can also apply the lut to the image with it.

phweyland:
> There's probably a gmic command

I hadn't thought to gmic commands. 🙄
Good idea. I'll make a try. Thanks

phweyland:
> There's probably a gmic command

I've changed my code to use only gmic commands and it compiles (without CImg.h).
But the link of libgmic.a gives some errors on fftw dependency. I have no idea how to add this dependency into the dt build.
Note: the gmic build seems to be happy with fftw dependency. I've tried to build gmic with fft off (without fft) but I haven't succeeded.

dt build error:
```
[ 76%] Linking CXX shared module liblut3d.dll
C:/msys64/mingw64/lib/libgmic.a(libgmic.o):gmic.cpp:(.text$_ZN12cimg_library8CImgListIfE3FFTEb[_ZN12cimg_library8CImgListIfE3FFTEb]+0xf6): undefined reference to `fftw_destroy_plan'
C:/msys64/mingw64/lib/libgmic.a(libgmic.o):gmic.cpp:(.text$_ZN12cimg_library8CImgListIfE3FFTEb[_ZN12cimg_library8CImgListIfE3FFTEb]+0xfe): undefined reference to `fftw_free'
[...]
```


GMIC Makefile:

```
# Enable Fast Fourier Transforms, using the FFTW3 library.
# (http://www.fftw.org/)
FFTW_CFLAGS = -Dcimg_use_fftw3 $(shell pkg-config --cflags fftw3)
FFTW_LIBS = $(shell pkg-config --libs fftw3 || echo -lfftw3)
ifneq ($(OS),Windows)
FFTW_LIBS += -lfftw3_threads
endif
```

phweyland:
@rabauke  
> On a typical Debian-based Linux distribution, CImg.h is provided by installing the package cimg-dev, which copies the header file to /usr/include. On a Windows system with msys2 packman can be used to install (common) packages. For the CImg library, however, a manual installation is required. Can not describe the procedure in detail as I am currently not on Window

I still don't understand, I'm sorry :-( . A normal header file doesn't contain code. It just contains references to a code which is elsewhere, normally in a library installed on the OS. However CImg.h contains all the source code of the cimg-dev library. Then any usage of CImg.h includes all this code automatically (this is what you have done with your own gmic dt branch if I'm not mistaken). On the other hand there is no way to install cimg-dev as a standard library (*.a, *.so, *.dll, whatever...) and there is no header file to provide the references to this hypothetical library. At least I have not found the way to do it...

edgardoh:
@phweyland , I'm no expert on the subject but I think that CImg.h is meant to be included, not linked as a library. But you should confirm that with someone with better knowledge than me.

Also, could you try building my PR? It works fine on a clean linux installation, so it should work on windows too.

phweyland:
> CImg.h is meant to be included, not linked as a library

I agree. This is what I've understood so far.

rabauke:
@phweyland The only thing you have to do is to copy `CImg.h` so some place outside the dt source tree and to configure your compiler (via. command line switches, environment variables or what ever the compiler requires) such that it is able to find the header file.

parafin:
> @rabauke
> 
> > On a typical Debian-based Linux distribution, CImg.h is provided by installing the package cimg-dev, which copies the header file to /usr/include. On a Windows system with msys2 packman can be used to install (common) packages. For the CImg library, however, a manual installation is required. Can not describe the procedure in detail as I am currently not on Window
> 
> I still don't understand, I'm sorry :-( . A normal header file doesn't contain code. It just contains references to a code which is elsewhere, normally in a library installed on the OS. However CImg.h contains all the source code of the cimg-dev library. Then any usage of CImg.h includes all this code automatically (this is what you have done with your own gmic dt branch if I'm not mistaken). On the other hand there is no way to install cimg-dev as a standard library (*.a, *.so, *.dll, whatever...) and there is no header file to provide the references to this hypothetical library. At least I have not found the way to do it...

It's called a header-only library and the only difference from user point of view that you don't have to link to anything, only include the header (like you do with any other library). It doesn't really matter if there are references or actual code in the header.

phweyland:
> The only thing you have to do is to copy `CImg.h` so some place outside the dt source tree and to configure your compiler (via. command line switches, environment variables or what ever the compiler requires) such that it is able to find the header file.

Even if you put CImg.h outside the dt source tree, the compiler includes it in the dt code ... @TurboGit finds we are distributing the CImg code with dt and that is not the correct way. However, using the CImg.h library, is an easy way to be sure it will work on all OS.

I still don't know if the CImg code is really embedded in libgmic but at least, without the appropriate header file, we haven't access to it. 
Then I'm trying now to use only gmic commands instead of CImg resources as @edgardoh has suggested. The gmic commands seem ok. I have still to find the way to install properly libgmic for windows (at least for msys2). I have some difficulties:
- libgmic doesn't build dll file (idll switch is in the MakeFile but seems not to be effective). However dt build is happy with libgmic.a file (so far).
- MakeFile is not aligned with CMakeLists.txt (as far as I understand), so I'm not able to use it to select the components (fftw, ...).
- Even if I remove the fftw references from MakeFile, gmic.cpp still needs them.
- fftw is visible on my msys2 through pkg-config but pacman doesn't see it ...
my competences stop here for the time being ...

phweyland:
I have added in FindGMIC.cmake this:

if(WIN32)
  find_library(FFTW_LIBRARY NAMES fftw3)
endif(WIN32)

It may not be perfect but I'm able now to build dt with libgmic and make lut3d work as before (without CImg.h).
Remains to make the changes to manage the presence or not of the library in the module itself.

phweyland:
I've made some tests between with and without libgmic dt builds. 
In the version without libgmic I've save the params space to store the compressed lut. It seems to be a bad idea, because dt detects a mismatch when it opens the image with the other dt build than the one used to treat the image, and remove the corresponding module step. Therefore we always loose the former information when we move from a dt build to the other. This is not smart. I was expecting that the opening the image to fail but not to loose the module params.
If anyone knows a way to avoid this behaviour please tell me. 

Meanwhile I'll fix that keeping the compressed lut space in params for the without gmic dt build.

phweyland:
> Why not make availability of cimg a compile time define and just do not support compressed LUTs if not available but keep compiling the "old" code loading un-compressed LUTs?

Updated this way on the current PR (using gmic commands instead of CImg resources).

src/iop/lut3d.c
```diff
@@ -424,72 +510,61 @@ double dt_atof(const char *str)
...
     }
     fractional_result /= fmultiplier;
-  }
+    }
   double result = sign * (integral_result + fractional_result);
   if (*str == 'e' || *str == 'E')
   {
     str++;
     double power_sign = 1;
-    if (*str == '+')
```
TurboGit: All those changes should be reverted.

src/iop/lut3d.c
```diff
@@ -831,65 +905,180 @@ void cleanup_global(dt_iop_module_so_t *module)
...
+  memcpy(&d->params, self->default_params, sizeof(dt_iop_lut3d_params_t));
   d->clut = NULL;
   d->level = 0;
   self->commit_params(self, self->default_params, pipe, piece);
 }
 
 void cleanup_pipe(struct dt_iop_module_t *self, dt_dev_pixelpipe_t *pipe, dt_dev_pixelpipe_iop_t *piece)
 {
-  dt_iop_lut3d_data_t *d = (dt_iop_lut3d_data_t *)piece->data;
+  dt_iop_lut3d_data_t *d = (dt_iop_lut3d_data_t *)piece->data;;
```
TurboGit: probably one ; should be removed :)

phweyland: Done

src/iop/lut3d.c
phweyland: Applied back the rules

TurboGit:
> It seems to be a bad idea, because dt detects a mismatch when it opens the image with the other dt build than the one used to treat the image

You mean with a version of dt which does not have been built with Gmic support, right? That's not a problem to me. Or is this about another scenario?

TurboGit:
> using gmic commands instead of CImg resources

What do you mean by this? gmic API? I suppose you don't spawn a gmic process, right?

phweyland:
> You mean with a version of dt which does not have been built with Gmic support, right? That's not a problem to me. Or is this about another scenario?

The problem can happen when a user has used a dt with gmic and then (version change for example) uses a version without gmic. In that case opening an image treated with lut3d, trigs the mismatch detection and the user looses his work (in database and in xmp file). No, I don't find that smart.

Not to have this issue, I've let the the same parameter size for both versions. Obviously dt without gmic does not try to use the compressed data, just informs that the input (compressed) file is not compatible.

> gmic API? I suppose you don't spawn a gmic process, right?

Yes, gmic API instead of CImg API. No, I don't launch any gmic process.

There is still something I don't understand with the "at build" detection of the presence or not of gmic.
For those who build dt themselves, that's already questionable for not Linux OS, but what about the standard distribution of dt ? How does that work ?

phweyland:
> How does that work ?

Trying to make progress ...
Generating gmic library we can get libgmic.a, libgmic.so, libgmic.dll & libgmic.dll.a.

1. libgmic.a is a static library. The build is happy with it but the link includes it in plugins/liblut3d.dll. Same effect as with CImg.h... => +5Mb and everything works fine.
2. If I add libgmic.dll.a along with libgmic.a, the link produces a plugins/liblut3d.dll at normal size (143KB). 
2.1. if dt doesn't find the library at execution time (ie. libgmic.dll) it says that it cannot load plugins/liblut3d.dll and the module lut3d doesn't appear in the module list. Not too bad, but not correct.
2.2. Adding libgmic.dll in dt/bin folder (where I know dt is able to find some libraries), dt loads plugins/liblut3d.dll. But when I selected a compressed lut, gmic fails and displays messages showing that parameters are messed up in the transfer to gmic. And this makes dt crash. Something is still wrong will libgmic.dll.

So far I understand 2 things (if I'm not mistaken):
1. I'm not yet able to generate the lib files for gmic properly. 
2. The current build process to check if the library is on the system or not,  is not complete. It tests the presence of an import library but doesn't test if the run time library is available and visible by applications.

Any advice or piece of information is welcome.

Here is the gmic Makefile part for generating the library. I have added the Windows part to get the dlls.

```
lib:
   $(MAKE) "CFLAGS+=$(GMIC_LIB_CFLAGS) $(OPT_CFLAGS)" "LIBS+=$(GMIC_LIB_LIBS)" _lib
_lib: libgmic.o use_libgmic.cpp
   ar rcs libgmic.a libgmic.o
ifeq ($(OS),Darwin)
   $(CXX) -shared -o libgmic.so libgmic.o $(LIBS)
else
ifeq ($(OS),Windows)
    $(CXX) -shared -o libgmic.dll \
    -Wl,--out-implib=libgmic.dll.a \
    -Wl,--export-all-symbols \
    -Wl,--enable-auto-import \
    -Wl,--whole-archive libgmic.o \
    -Wl,--no-whole-archive $(LIBS)
else
    $(CXX) -shared -Wl,-soname,libgmic.so.$(VERSION1) -o libgmic.so libgmic.o $(LIBS)
    $(CXX) -o use_libgmic use_libgmic.cpp -L. -lgmic $(LIBS)
endif
endif
```

This produces these files:
image

phweyland:
>     1. I'm not yet able to generate the lib files for gmic properly.

The test application **use_libgmic** I've built using the same options as dt build **works perfectly** with the windows dll.

Here are the build command:

`g++ -Wall -Wno-format -Wshadow -Wtype-limits -Wvla -Wno-error=varargs -Wno-format-truncation -Wframe-larger-than=32768 -Wstack-usage=32768 -Wlarger-than=524288 -std=c++14 -std=c++11 -fopenmp -march=native -msse2 -g -mfpmath=sse -O3 -DNDEBUG -O3 -fexpensive-optimizations -Wl,--as-needed -Wl,--enable-runtime-pseudo-reloc -o use_libgmic -Wl,--major-image-version,0,--minor-image-version,0 -Wl,--whole-archive use_libgmic.cpp -Wl,--no-whole-archive /C/msys64/mingw64/lib/libgmic.dll -lfftw3
`

phweyland:
Here are the gmic message when dt is linked with libgmic.dll:

```
[gmic]-0./ Start G'MIC interpreter.
[gmic]-0./ Input file 'C:\Us▒▒U' at position 0
[gmic]-0./ *** Error *** Unknown filename 'C:\Us▒▒U'.
[gmic]-1./ Start G'MIC interpreter.
[gmic]-1./ Input file '-decompr▒▒U' at position 1
[gmic]-1./ *** Error *** Unknown command or filename '-decompr▒▒U'.
```

for the first command the parameter is:
`-i C:\Users\philippe\AppData\Local\Microsoft\Windows\INetCache\gmic\black_and_white.cimgz,uchar`

and for the second one:
`-decompress_clut 48,48,48`

phweyland:
Some other experiment.
If I allocate memory to pass the gmic string arguments, gmic gets them properly. But the gmic treatment still fails, probably because of the other passed parameters.
It seems that static memory buffers are not passed correctly to the dll (while again they are passed correctly to the static library).

1. Is there somebody to run the test under Linux to see what happens out there ? 
1.1. libgmic.a alone for static library build
1.2. then add libgmic.so for dynamic library build.
Here are the libgmic resources: https://gmic.eu/libgmic.shtml
2. @peterbud. Would you have any advices, tests to run or directions to try to solve this issue on windows side ?

phweyland:
Other than the build in Windows and Mac this is ready for review.

phweyland:
Some regression and some progress.
Progress. I'm now able to work to libgmic.dll.
Regression. Compressed luts have changed a bit with GMIC 2.6.4. I've fixed my code to make it work, but ...

As compressed lut is a GMIC specific process I think we should use the specific GMIC files for this (*.gmz for example) instead of two lines *.png files as of today. Using *.png brings a bit of space saving but using *.gmz files would simplify the users' life. Moreover *.gmz format can handle multiple luts. So, as a future improvement, we could read directly the GMIC lut library gmic_luts.gmz as well.

I'll make some rework on this with the aim to manage uncompressed *.png & *.cube LUTs as already merged, adding the compressed *.gmz  LUTs.
PS I will have to make some changes on the dt blog too.

phweyland:
I still have to increment the version as parameters have changed.

As gmz files can contain several luts each, I would like to make them scrollable.

phweyland:
@TurboGit I think this is working.
I cannot say the management of multi luts file, gmz, handles properly the huge gmic_cluts.gmz file from G'MIC, as the combobox (at least the way I use it) is not made to handle 750+ items, but it works well for any gmz files containing few to 50 luts. This should be good for those who manage their gmz files per type of luts (film kodak, fuji,..., b&w,...). 
To manage a higher number of luts, the combobox popup should have a scroll bar. But I haven't found the way to do it so far.
Another potential improvement would be to present thumbnails of the current image, instead of a list of lut name, as suggested on pixls. Another project I think...
Anyway those improvements are essentially on the UI and shouldn't change the basics.
On the other hand I've issued a PR [#5518](https://github.com/msys2/MINGW-packages/pull/5518) on msys2/mingw to add gmic to the distribution. Still WIP but I think we can make it work.

Could you please review. Comment and suggestion for improvement are welcome.

phweyland:
> > I'm also wondering why sometime a change on the module adds an history step and sometime not. I ask the question because that impacts the performance.
> > Once the history is compressed the steps are not added any more.
> 
> This is a bug in dt since a long time. I never found the root cause but for sure this is not an issue with this PR.

I get the repetition of the module step every time I go away and come back to the module (or the image), even if I haven't modified any other module.
This comes from focus_hash in develop.c here:
```

    if(!history // if no history yet, push new item for sure.
       || module != hist->module
       || module->instance != hist->module->instance             // add new item for different op
       || module->multi_priority != hist->module->multi_priority // or instance
       || ((dev->focus_hash != hist->focus_hash)                 // or if focused out and in
       && (// but only add item if there is a difference at all for the same module
         (module->params_size != hist->module->params_size) ||
         include_masks ||
         (module->params_size == hist->module->params_size && memcmp(hist->params, module->params, module->params_size)))))
```
If the last module in the stack is the same I think we should not add a step in the history, having lost the focus or not.
However this seems to have been designed this way (even if not completely consistent because successive changes without lost of focus are lost). An alternative would be a preference setting like "auto compressed history". This would avoid any repetition. Any thought ?

TurboGit:
> I get the repetition of the module step every time I go away and come back to the module (or the image), even if I haven't modified any other module.

Please open an issue for this to be discussed. I'm not sure we want to change this behavior.

TurboGit:
I cannot compile this. Looks like you may have a missing entry in DefineOptions.cmake (root of repository)? On my side USE_GMIC is never set I think and so no support in dt.

How do you compile this on your side?

phweyland:
> Please open an issue for this to be discussed. I'm not sure we want to change this behavior.

I agree, I think this behaviour has been wanted and is as designed. This is not the infinite repetition of steps which still happens sometimes...

phweyland:
> How do you compile this on your side?

I've just tried to build again after having emptied the build folder. And I haven't any more the flag GMIC found in the first make.
This remembers me that a part of CMakeFile.txt with these statements :
```
option(BUILD_NOISE_TOOLS "Build tools for generating noise proifles" OFF)
option(BUILD_CURVE_TOOLS "Build tools for generating base and tone curves" OFF)
option(USE_LENSFUN "Enable LensFun support" ON)
option(USE_GMIC "Use G'MIC image processing framework." ON)

```
has been removed with one of the last merges/rebases I've made. This part is not present present any more. 

So where this setting should be done now ?

So far my build has worked probably keeping the last found modules in a previous cmake.
(to be sure a new library is taken in account I've seen I've to delete the content of the build folder. if not the change are not detected).

phweyland:
> So where this setting should be done now ?

Ok, I've made the link with DefineOptions.cmake now. I look at this and come back to you.

phweyland:
I've fixed the DefineOptions.cmake option issue.
So,
1. If libgmic library is not found, lut3d is compiled without gmic and has the features of the already merged module. The only difference is that params keep room for compressed data. That way, if the user has already made some compressed lut work, it won't be lost if it happens to compile dt on a system without gmic.
2. If libgmic is found, lut3d is compiled with the compressed lut features.
2.1. if dt doesn't find the libgmic library at startup, the lut3d module is missing.
2.2. if dt finds the libgmic at startup, well... it should work. For the time being I'm copying libgmic.dll into darktable/bin on windows. I hope the msys2 gmic will work soon. And with the help of @peterbud the windows build will be able to detect if libgmic.dll is avaible at run time.

phweyland:
PS: in FindGMIC.cmake I'm looking for fftw3. In the @edgardoh GMIC PR it seems this is not needed on Linux... To be confirmed.

if(WIN32)
  if(GMIC_LIBRARY)
    find_library(FFTW_LIBRARY NAMES fftw3)
  ENDIF(GMIC_LIBRARY)
endif(WIN32)

phweyland:
Here are some gmic commands to create/manipulate gmz (compressed) files I've prepared in case this PR gets merged ...

```
###Compression
Compression is a process which takes time (can be several minutes).
* *haldclut* file:
`gmic -i <filename1>.png -compress_clut 8,2,2048 -o <filename2>.gmz,uchar`
To add an *haldclut* to an existing gmz library:
`gmic <filename>.png -compress_clut 8,2,2048 -i <library>.gmz -o <library>.gmz,uchar`
* *cube* file:
`gmic -input_cube <filename1>.cube -compress_clut 8,2,2048 -o <filename2>.gmz,uchar,uchar`
To add a *cube* file to an existing gmz library:
`gmic -input_cube <filename>.cube -compress_clut 8,2,2048 -i <library>.gmz -o <library>.gmz,uchar`
###Extraction
To extract all LUTs contained in the library as single compressed LUT gmz files,
the filename is given by the lut name:
`gmic -i library.gmz -repeat $!,k -local[$k] -o d/{0,n}.gmz,uchar -endlocal`
To extract the LUT `lutname` to a single LUT compressed gmz file:
`gmic -i gmic_cluts.gmz -o[lutname] lutname.gmz,uchar`
```

And here some gmz files to play with.

[singlelutgmz.zip](https://github.com/darktable-org/darktable/files/3317102/singlelutgmz.zip)
[multiplelutsgmz.zip](https://github.com/darktable-org/darktable/files/3317105/multiplelutsgmz.zip)

phweyland:
Update.
PR [#5518](https://github.com/msys2/MINGW-packages/pull/5518) to add gmic to msys2 has been merged but the 32 bits compilation doesn't work (g++ memory allocation error ...).
There is a long time the GMIC team hasn't provided 32 bits version and has no advice to solve the issue.
@Alexpux (msys2) says that as long as they support 32 bits only packages build for both architectures are published.
So, despite we can build the 64 bits package for gmic, it is not published on msys2 site.
Any idea or advice ?

TurboGit:
> There is a long time the GMIC team hasn't provided 32 bits version and has no advice to solve the issue.

I have one. Drop the 32bit compilation, dt does not support 32 bits systems anymore anyway.

phweyland:
> Drop the 32bit compilation, dt does not support 32 bits systems anymore anyway.

It was my first attempt. But Msys2 distribution (https://msys2.github.io/) accepts only packages which build for both, 32 and 64 bits, architectures...

phweyland:
@TurboGit 
Replacing the combox by a treeview list allows to manage easily files with hundreds luts like gmic_cluts.gmz.
This is now fully functional.

TurboGit:
I'm not sure what will happen to this but there is at least one feature that I'd like to recover and have back into dt. Today IIRC when you use a cube or png (non compressed) you have a reference to the name of the cube/png into the XMP and database. Right? That is, the development depends on some file on the disk and cannot be shared easily and one can lost the original lut file.

The feature I'd like to have into dt is to be able to compress all the Lut in dt when needed and always record this into the XMP and database. No more reference to external files is the goal here.

What do you think?

phweyland:
Technically there two points:
- decompression process (compression is very time consuming and can be done externally). The compressed LUT is stored in params and XMP.
- cache file to avoid decompression time and speed up opening time.

These 2 processes are GMIC dependent so far.
I have worked on GMIC distribution on MSYS2 but only the 64bits distribution works (the 32 bits doesn't - compiler issue - GMIC team doesn't support 32 bits and didn't help). Therefore the MSYS2 team didn't want to accept the PR.

The different ways to get there:
- convince MSYS2 team that GMIC distribution is worth to have (64bits only). I think the Linux distribution is OK but I don't know for Mac.
- include the GMIC stuff as source code or static library (and manage the updates of GMIC source code manually when needed (no need for MSYS2 distribution any more)).
- copy/paste/adapt just the needed code into dt. A lot more work... and C++.
- use another compression/decompression algorithm (to be defined). Would not be compatible with GMIC compressed luts distribution any more.

I've made work both dynamic and static libraries locally but I don't master the consequences on the different OS distributions. Still ready to support however.

TurboGit:
> compression is very time consuming and can be done externally

But that's what I'd like to avoid in fact as the incantation to compress such Lut is not easy and requires some external tool. As far as I understand compressed Lut are already supported in current dt?

> I have worked on GMIC distribution on MSYS2 but only the 64bits distribution works (the 32 bits doesn't - compiler issue - GMIC team doesn't support 32 bits and didn't help).

I'm not sure to understand this, dt does not work in 32bits. So we don't need support for 32bits anyway. Or are you saying that you're using a 32bit toolset somehow?

Why not have the G'MIC library has an optional dependency and if present build compressed lut support in dt? We already have optional dependencies like lensfun, lcms2... possibly others. How does that sounds?

TurboGit:
Anyway forget about my proposal, I tried to compress a Lut and see that what we call take a long time is not some seconds but **many minutes**. So this cannot be done from dt anyway.

phweyland:
> > compression is very time consuming and can be done externally
> 
> But that's what I'd like to avoid in fact as the incantation to compress such Lut is not easy and requires some external tool. As far as I understand compressed Lut are already supported in current dt?

Anyone can compress lut (cube or png) using GMIC. I've given the command lines on dt blog. Takes few to some minutes per lut (to compress).
The other way to get compressed lut is to use those distributed by GMIC.
This PR doesn't implement the compression. But this is possible. We could use background jobs to do it....

With this PR dt is able to read these compressed luts, to store them in params. Additionally, to avoid the decompression time (half a second to few seconds) 3 dimensions cache files (GMIC provided) are implemented. So if the cache file is found, it is used instead of decompressing the params lut.


> Anyway forget about my proposal, I tried to compress a Lut and see that what we call take a long time is not some seconds but **many minutes**. So this cannot be done from dt anyway.

Even with a background job ?

phweyland:
> > I have worked on GMIC distribution on MSYS2 but only the 64bits distribution works (the 32 bits doesn't - compiler issue - GMIC team doesn't support 32 bits and didn't help).
> 
> I'm not sure to understand this, dt does not work in 32bits. So we don't need support for 32bits anyway. Or are you saying that you're using a 32bit toolset somehow?

Yes but the MSYS2 team doesn't want to have a distribution only in 64 bits. They want to have both ...
as long as they support 32 bits.

TurboGit:
> Even with a background job ?

Well one need to wait to be able to embedded the Lut into the XMP or db when leaving a picture to edit a new one or leave dt. This is really impractical to me.

phweyland:
> > > I have worked on GMIC distribution on MSYS2 but only the 64bits distribution works (the 32 bits doesn't - compiler issue - GMIC team doesn't support 32 bits and didn't help).
> > 
> > 
> > I'm not sure to understand this, dt does not work in 32bits. So we don't need support for 32bits anyway. Or are you saying that you're using a 32bit toolset somehow?
> 
> Yes but the MSYS2 team doesn't want to have a distribution only in 64 bits. They want to have both ...
> as long as they support 32 bits.

But maybe we could use an other distribution way than MSYS2. But here I'm not competent.

TurboGit:
> With this PR dt is able to read these compressed luts, to store them in params. Additionally, to avoid the decompression time (half a second to few seconds) 3 dimensions cache files (GMIC provided) are implemented. So if the cache file is found, it is used instead of decompressing the params lut.

Ok, this is very good already! I thought this was already implemented as described into the blog article :)

TurboGit:
> But maybe we could use an other distribution way than MSYS2. But here I'm not competent.

Ok, clear now thanks !

phweyland:
> > Even with a background job ?
> 
> Well one need to wait to be able to embedded the Lut into the XMP or db when leaving a picture to edit a new one or leave dt. This is really impractical to me.

If a user modifies some images with LUT module, the lut could be registered for a background job (asynchronous). If the user leaves dt (or modifies again the same image) before the end of compression process, this process would be aborted. If the compression process ends the image's params would be updated with the new compression lut. Some lock process would be probably necessary.
If the user quits dt before the end of the process the user could be asked if he really want to quit.

phweyland:
> > > Even with a background job ?
> > 
> > 
> > Well one need to wait to be able to embedded the Lut into the XMP or db when leaving a picture to edit a new one or leave dt. This is really impractical to me.
> 
> If a user modifies some images with LUT module, the lut could be registered for a background job (asynchronous). If the user leaves dt (or modifies again the same image) before the end of compression process, this process would be aborted. If the compression process ends the image's params would be updated with the new compression lut. Some lock mechanism would be probably necessary.
> If the user quits dt before the end of the process the user could be asked if he really want to quit.

We could also cache the compressed luts themselves to avoid to compress again the same ones ....

TurboGit:
Well I have an ongoing compression since more than 1h now... The PC is slow because it uses lot of CPU. Really impractical to me.

phweyland:
> Well I have an ongoing compression since more than 1h now... The PC is slow because it uses lot of CPU. Really impractical to me.

Wondering if, inside dt, it would be possible to limit the compression process to only one CPU core. 

However, beside compression topic, this PR is still of interest to me if we can afford in some way the dependency on GMIC. Let me know if I can do anything to help.

Alexpux:
@phwsyland MSYS2 provide GMIC package for about half a year maybe

phweyland:
> MSYS2 provide GMIC package for about half a year maybe

Sounds great. I'll have a look at it. Thanks a lot.

phweyland:
I've rebased to master and made a try.

pacman -S mingw-w64-x86_64-gmic adds the libgmic stuff to lib and include. The build detects gmic and finishes without error. But at dt start, the module lut3d fails to launch at:

`module->module = g_module_open(libname, G_MODULE_BIND_LAZY | G_MODULE_BIND_LOCAL);`

I get this message:
`[iop_load_module] failed to open operation lut3d: C:\msys64\opt\darktable\lib\darktable/plugins\liblut3d.dll: The specified module could not be found.`

I've checked that liblut3d.dll is at the right place and that libgmic.dll is in bin folder.

So far I don't understand why this fails.

phweyland:
Hi @Alexpux , I cannot find the gmic package on "https://github.com/Alexpux/MSYS2-packages". Is there a way to fetch it from somewhere ? Thanks

Alexpux:
@phweyland this is mingw package, not msys. So you need look at https://github.com/msys2/MINGW-packages

phweyland:
> 
> 
> @phweyland this is mingw package, not msys. So you need look at https://github.com/msys2/MINGW-packages

Great ! Thanks!

phweyland:
dt works again with gmic 2.6.4.
Still to understand what doesn't work with the last mingw gmic version.

phweyland:
@Alexpux 
I have found two points. The first one.
gmic 2.8 works with a small modification in 0001-gmic-2.6.4-windows-build.patch:
line 57
```
+ifeq ($(OS),Windows)
+	$(CXX) -shared -std=c++11 -pedantic -o libgmic$(SO) \
```
becomes

```
+ifeq ($(OS),Windows)
+	$(CXX) -shared -o libgmic$(SO) \
```
I don't know what is the impact of -std and -pedantic but at least one of them is problematic here.
Do I issue a PR ?

The second point is about pacman.
`$  pacman -S mingw-w64-x86_64-gmic`
doesn't propose the version 2.8 (you have issued few days ago) but only 2.6.7:
`Packages (1) mingw-w64-x86_64-gmic-2.6.7-1`
What could be the reason ? Any idea ?

Thanks for your help.

Alexpux:
phweyland do you update packages database to get latest packages?
pacman -Syu

phweyland:
> do you update packages database to get latest packages?
> pacman -Syu

Time to time. Not to often, this can give some additional work :)
I make a try. Thanks

phweyland:
As expected some work to get the chain working again... :)
But now I get the last gmic version from MINGW-packages.

Remains just to fix the patch file.

phweyland:
The gmic package from msys2 works (with the last version). Thanks @Alexpux.
The last commit to make the scrolling in the gmz lut list work as the combo box works for files (when the user scrolls the list the lut selection moves accordingly).
Added also gmic as an optional package in windows/build.txt

Ready to be reviewed. Thanks.

phweyland:
With compressed lut the UI is this one:

image

phweyland:
I've setup a debian machine and started building dt on it. No issue with dt master so far.
No more issue with this PR without GMIC library.
But I haven't found the way to build GMIC 2.8. An opencv component is missing (calib3) which seems to be in more advanced version of opencv.
Any idea ?

phweyland:
with gmic installed from https://www.fosshub.com/GMIC.html plus a small change in FindGMIC.cmake I've succeeded in building dt with compressed lut.

I've seen some comments on compressed lut in 3.00 announcements ... maybe it's time to try to merge this ?

phweyland:
I've checked it works too on windows 10.

phweyland:
@turbogit, is there anything left which prevent to consider this pr ?

TurboGit:
@phweyland : nothing except time :) I still need to work on my #3908 as I think it should be merged sooner than later.

TurboGit:
BTW, should this be for 3.0.1 or later ? How do you fill about this ? IIUC it needs support for a new external lib, so maybe wait for 3.2 ? I don't fill that adding new dependencies on the 3.0.1 branch is safe.

phweyland:
> I don't fill that adding new dependencies on the 3.0.1 branch is safe.

I can understand that. I have spent a significant amount of time trying to solve the dependencies and some surprises are always possible.
But this PR is ready now and waiting for more 12 months doesn't sound great at all.

> IIUC it needs support for a new external lib

The up-to-date external libs are available on https://www.fosshub.com/GMIC.html for linux. I've tested it on Debian 10.
For windows it is available on msys2 thanks to @Alexpux help. Windows build.txt file is updated accordingly.
The only point I haven't checked is mac os.

This feature has been discussed on github since last April or May and seemed to get good support (https://discuss.pixls.us/t/3d-lut-module-in-darktable-2-7-dev/12341). Even yourself has been supportive at some point of time... 
Maybe I can ask people there to check the integration point. What do you think ?

TurboGit:
>  Even yourself has been supportive at some point of time...

And I'm still are supporter of this feature ! Don't take me wrong, I do my best. Sorry to have this for so long lingering here, just missing time and side tracked by very important issues (remember the iop-order earthquake we had just before the release !).

Let me test this now, code looks ok. AFAIU this require the external library to work, if not found then it is not activated, so all in all not a required dependencies which makes it ok for 3.0.1 probably.

I'm don't remember do we have a .gmz somewhere to test?

TurboGit:
Ok, I found the gmic_film_cluts.gmz file. I open it and have a huge list of Lut, fine.

But for anyone I select I have a fully black image. Is there something more to do ?

phweyland:
> But for anyone I select I have a fully black image. Is there something more to do ?

Have installed a recent GMIC version ? If I remember well Debian comes wit 2.4 which is not enough.
The gmic 2.8.1 from the link above works for me. I had to resolve a dependency.

Any message on the console ?

phweyland:
Here the gmz I use.
[gmic_cluts.gmz.txt](https://github.com/darktable-org/darktable/files/4023542/gmic_cluts.gmz.txt)

TurboGit:
> Any message on the console ?

Yes, many:

```
[lut3d gmic] error: incompatible compressed LUT [0] 60's
[lut3d gmic] error: incompatible compressed LUT [0] 60's
[lut3d gmic] error: "*** Error in ./decompress_clut/*repeat/*local/ *** Command '*': Invalid selection [1] (contains indice '1', not in range -1...0)."
```

TurboGit:
So probably an issue with my version:

```
$ apt-cache policy libgmic-dev
libgmic-dev:
  Installed: 2.4.5-1+b1
  Candidate: 2.4.5-1+b1
  Version table:
 *** 2.4.5-1+b1 750
        750 http://ftp.fr.debian.org/debian unstable/main amd64 Packages
        100 /var/lib/dpkg/status
     2.4.5-1 650
        650 http://ftp.fr.debian.org/debian stable/main amd64 Packages

```

phweyland:
Not yet an expert on this ... what I have. 2.8.1 installed in local.

```
philippe@philippe-pc:~/git/dt/build$ apt list | grep gmic
gimp-gmic/stable 2.4.5-1 amd64
gimp-gmic/stable 2.4.5-1 i386
gmic-zart/stable 2.4.5-1 amd64
gmic-zart/stable 2.4.5-1 i386
gmic/now 2.8.1 amd64 [installed,local]
gmic/stable 2.4.5-1 i386
krita-gmic/stable 2.4.5-1 amd64
krita-gmic/stable 2.4.5-1 i386
libgmic-dev/stable 2.4.5-1 amd64
libgmic-dev/stable 2.4.5-1 i386
libgmic1/stable 2.4.5-1 amd64
libgmic1/stable 2.4.5-1 i386
```

phweyland:
With your command I get this :

$ apt-cache policy libgmic-dev
libgmic-dev:
  Installed: (none)
  Candidate: 2.4.5-1
  Version table:
     2.4.5-1 500
        500 http://deb.debian.org/debian buster/main amd64 Packages

TurboGit:
@phweyland : The first thing to do is to check for the minimal libgmic version as done for liblensfun or libosmgpsmap for example.

TurboGit:
I'm not sure we can get a library version from cmake but I'm no expert. Maybe you'll have to resort on checking the version in the code at runtime using gmic_version define and so disable handling .gmz.

phweyland:
> I'm not sure we can get a library version from cmake

Indeed I'm struggling there. So far I did not get it.

phweyland:
Hi @Nilvus, I've seen you have fixed a number of css issues. Maybe you could tell me how I could solve the following one.

In that image, there are 2 highlighted rows. The darker one is the selected one. The lighter one is the prelit one.
The selected one follows the scrolling wheel and that's the one I want to keep to show to the user which lut is selected and active. The second one is disturbing here. I haven't found the way to get rid of it.
Would there be a way to prevent prelit thanks to css ? If yes, any direction is welcome.
Thanks.

image

phweyland:
I haven't found the way to get the version from cmake.
So I've implemented the run time version detection. Which is probably more flexible if other usages of gmic are added.

To test this I've looped on this sequence:
1. uninstall gmic-2.8.1
2. install libgmic-dev from debian repository (4.5)
3. delete build/CMakeCache.txt
4. cmake
5. make install
6. check that UI is the same as without gmic + warning on images with gmic 2.8.1.
7. uninstall libgmic-dev
8. install gmic-2.8.1
9. delete build/CMakeCache.txt
10. cmake
11. make install
12. check that normal gmic operations are OK. If you haven't changed the previous images with gmic 2.8.1 they should be restored correctly.

On windows there is no way to come back to an older version of gmic.

Nilvus:
> Hi @Nilvus, I've seen you have fixed a number of css issues. Maybe you could tell me how I could solve the following one.
> 
> In that image, there are 2 highlighted rows. The darker one is the selected one. The lighter one is the prelit one.
> The selected one follows the scrolling wheel and that's the one I want to keep to show to the user which lut is selected and active. The second one is disturbing here. I haven't found the way to get rid of it.
> Would there be a way to prevent prelit thanks to css ? If yes, any direction is welcome.
> Thanks.
> 
> image

Hi @phweyland: I have never heard this prelit word (deepl and Google translate doesn't give me a translation) so not sure to really understand here as it's the main word. English is not my native language (even my english dictionnary doesn't have this word on it ; it's an old one).

I didn't use this module, but on what I have understand, the darker one line seems to be an hover one so it's an hover tag to adjust. Anyway, removing hover tag could impact other parts of the UI and removing it could make something inconsistent. I will try to test (with some lut) to check and be sure on what you want.

phweyland:
> So what's prelit mean ?

It's a word I've picked up from gtktreeview documentation (renderer), a kind of pre-lightening ?
In short when one hovers an object it is highlighted (lighter) as shown above.

Thank you a lot for responding. Meanwhile I've made my way and found this solution:

I've given to this treeview the name "lutname".
In css it's enough to change the line 1234 in darktable.css to : `treeview:not(#lutname) *:hover,` to prevent the hover effect for this treeview, and only for this one. :-)

Nilvus:
> In css it's enough to change the line 1234 in darktable.css to : `treeview:not(#lutname) *:hover,` to prevent the hover effect for this treeview, and only for this one. :-)

Yes, that's a way good to avoid breaking other parts. Good work . But you need to rebase this on master. This part is not anymore on line 1234 in darktable.css master so you create a conflict. It's now on line 1228 as I recently clean css and fix some issues (some redundant lines removed)

phweyland:
>  It's now on line 1228 as I recently clean css and fix some issues (some redundant lines removed).

Indeed. Done

Nilvus:
> > It's now on line 1228 as I recently clean css and fix some issues (some redundant lines removed).
> 
> Indeed. Done

Euh, you remove a comment I have recently added to help understand the CSS. Line 1227 :+1:
`/* Treeviews and filechoosers states */`

Could you please put in back ? Thanks

phweyland:
> Euh, you remove a comment I have recently added to help understand the CSS. Line 1227 +1
/* Treeviews and filechoosers states */

Ooooops. Sorry. Fixed hopefully ...

TurboGit:
Thanks! This is now merged on master and 3.0.1!

phweyland:
@TurboGit , thank you for your time and your support !

chhil:
Trying to build on windows.

Got following warning

CMake Warning at cmake/modules/LibFindMacros.cmake:263 (message):
  WARNING: MISSING PACKAGE

  We could not find development headers for GMIC.  Do you have the necessary
  dev package installed? This package is NOT REQUIRED and you may ignore this
  warning but by doing so you may miss some functionality of darktable.

  Relevant CMake configuration variables:

    GMIC_INCLUDE_DIR-NOTFOUND=
    GMIC_INCLUDE_DIR=<not found>
    GMIC_LIBRARY-NOTFOUND=
    GMIC_LIBRARY=<not found>


-----
Realized build.txt was updated and  pacman -S mingw-w64-x86_64-{gmic} was added.

Tried the command and got an error
$  pacman -S mingw-w64-x86_64-{gmic}
**error: target not found: mingw-w64-x86_64-{gmic}**

phweyland:
> Tried the command and got an error
$ pacman -S mingw-w64-x86_64-{gmic}
error: target not found: mingw-w64-x86_64-{gmic}

You are right. Works with several modules but fails with only one. :(
I'll fix that. Thanks

phweyland:
PR #4058 to fix this.

chhil:
Thank you @phweyland

PR #2545, opened 2019-05-10
2479: fix file action on highlighted, not selected

EvgeniiFrolov:
The problem takes place when
some images are selected (lighttable view) and mouse hovers over
non-selected one.
If, say, delete key is pressed on keyboard, then hovered (not selected) image is processed (deleted).

In this case
`mouse_over_id > 0` and
`sqlite3_step(darktable.view_manager->statements.is_selected) != SQLITE_ROW`.
Hence, mouse_over_id is returned, and hovered image is processed
when `-1` should be returned for selected images to be processed.

New statement
`darktable.view_manager->statements.all_selected`
is prepared to check if any images are selected.
In new version if no images are selected, mouse_over_id is returned.
If at least one image is selected,
`sqlite3_step(darktable.view_manager->statements.all_selected) == SQLITE_ROW`
and therefore, `-1` is returned.

### DIFF
--- a/src/views/view.c
+++ b/src/views/view.c
@@ -59,6 +59,8 @@ void dt_view_manager_init(dt_view_manager_t *vm)
   /* prepare statements */
   DT_DEBUG_SQLITE3_PREPARE_V2(dt_database_get(darktable.db), "SELECT imgid FROM main.selected_images "
                               "WHERE imgid = ?1", -1, &vm->statements.is_selected, NULL);
+  DT_DEBUG_SQLITE3_PREPARE_V2(dt_database_get(darktable.db), "SELECT imgid FROM main.selected_images ",
+                              -1, &vm->statements.all_selected, NULL);
   DT_DEBUG_SQLITE3_PREPARE_V2(dt_database_get(darktable.db), "DELETE FROM main.selected_images WHERE imgid = ?1",
                               -1, &vm->statements.delete_from_selected, NULL);
   DT_DEBUG_SQLITE3_PREPARE_V2(dt_database_get(darktable.db),
@@ -765,13 +767,9 @@ int32_t dt_view_get_image_to_act_on()
   else
   {
     /* clear and reset statement */
-    DT_DEBUG_SQLITE3_CLEAR_BINDINGS(darktable.view_manager->statements.is_selected);
-    DT_DEBUG_SQLITE3_RESET(darktable.view_manager->statements.is_selected);
+    DT_DEBUG_SQLITE3_RESET(darktable.view_manager->statements.all_selected);
 
-    /* setup statement and iterate over rows */
-    DT_DEBUG_SQLITE3_BIND_INT(darktable.view_manager->statements.is_selected, 1, mouse_over_id);
-
-    if(mouse_over_id <= 0 || sqlite3_step(darktable.view_manager->statements.is_selected) == SQLITE_ROW)
+    if(mouse_over_id <= 0 || sqlite3_step(darktable.view_manager->statements.all_selected) == SQLITE_ROW)
       return -1;
     else
       return mouse_over_id;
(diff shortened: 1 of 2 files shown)


TurboGit:
As far as I understand (please check the documentation) this is expected. You can fly over the lighttable to assign stars, color labels for example on image under the mouse. That's why the mouve under the cursor is highlighted with a lighter color than the selected ones.

EvgeniiFrolov:
> As far as I understand (please check the documentation) this is expected. You can fly over the lighttable to assign stars, color labels for example on image under the mouse. That's why the mouve under the cursor is highlighted with a lighter color than the selected ones.

Indeed, this is what the docs say. I was just agree with the issue opener, that in case you've selected several images, you want them to be deleted, not the one outside the selection and under the mouse.

Talking about stars and color labels, may be the same logic should apply. Say, you want to set 3 stars for a group of images:
1) You select a number of images
2) You push "3".

Why should 3 stars be assigned to some other image that is not selected, but is under the cursor?

Probably, I don't know all the uses, but for me selection (if anything is selected) is the scope of the action. And if you want to make an action on another image that is not the part of selection, either the selection should be dropped, or gui buttons on this another image should be used.

Nilvus:
> Why should 3 stars be assigned to some other image that is not selected, but is under the cursor?
> Probably, I don't know all the uses, but for me **selection** (if anything is selected) **is the scope** of the action. And if you want to make an action on another image that is **not the part** of selection, either the selection should be dropped, or gui buttons on this another image should be used.

Of course, by default that what we made most of the times. But quite regularly, I use the way of changing a star or lable without having to select the image. For example, I select an image, I put 4 stars, I continue and then select an image which I find need 4 stars and realize the previous should have 3 stars instead of 4. Actually, I just have to move the cursor on the previous image, use keyboard shortcut and I'm on. Quite faster than adding selecting it. That doesn't change a lot for one image but sometimes (and for other actions), I make that for few images (not on same time of course as selected is better for few or more images on same time). But the better action for using hover and not selection on image is for copying quickly a development on image to an other.

> If such an analogy could be presented for the sake of argument:
> Imagine, one would like to delete several files. The files are being selected. Then one file is intentionally excluded from selection via ctrl+click. Then delete is pushed on keyboard without any mouse movement.
> Expected result: selected files are deleted, the file excluded from selection - not.
> Actual result: excluded file is unexpectedly deleted, selected files are (unexpectedly) not deleted.
> 
> Personally, I find such behavior confusing and prone to errors.

I don't find that confusing, it's just a feature, so we have two possibility to act on images : classic selection you want and hover "selection" for quick action.

But of course, I agree with initial action on deleting. I'm on to having delete actions (only them) applied only on file selection like the issue you want to solve point.

Apart of that, I think the better way to delete images is 1st to reject them, then use images filter to only see rejected one, then select all and delete them. With that usage, no risk at all to delete an image who don't want to delete.

AlicVB:
> But of course, I agree with initial action on deleting. I'm on to having delete actions (only them) applied only on file selection like the issue you want to solve point.

hmm, but that would break the consistency of our interface... not to speak that on a such delicate topic, changing historic behaviors is dangerous as it can dramatically break things for long time users...

iirc, there has been some hot discussions on that specific point in the past (long time ago) and what has been decided, is to add a dialog confirmation with the nb of images involved in the deletions, so one can check and decide. Of course this doesn't solve everything as users may click "confirm" without reading (me first) but well...
Maybe, we can add some even more visible warning if we detect that there's a selection and pointer is hovering another image ?


> Apart of that, I think the better way to delete images is 1st to reject them, then use images filter to only see rejected one, then select all and delete them. With that usage, no risk at all to delete an image who don't want to delete.

:+1:

EvgeniiFrolov:
> that would break the consistency of our interface...

Agreed. It seems better to have single consistent behavior. When behavior differs from one situation to another I need to keep in mind more conditions and tend to be confused.

> add a dialog confirmation with the nb of images involved in the deletions

The dialog is there and serves it well. Personally for me this behavior is strange for starring too.

> there has been some hot discussions on that specific point in the past

Should the discussion be started again?
Or should "feature not a bug" be written in the issue thread?

TurboGit:
> Or should "feature not a bug" be written in the issue thread?

I would say open an issue for this. But don't hold your breath :)

I'll close this PR at this stage.

PR #2546, opened 2019-05-10
fix histogram not updated

### LINKED ISSUE #2538: dt2.7: Histogram in darkroom sometimes not updated when toggling modules active/inactive
**Describe the bug**
In darkroom, the histogram (upper right corner) is sometimes not updated when toggling modules active/inactive.

**To Reproduce**
1. Go to darkroom
2. Apply image correction using modules. Lens correction as a simple example, but also reproducible with other modules.
3. de-active then re-active corrections modules several times and observe the histogram after each clic
4. Sometimes, after several activation/deactivation cycles, the histogram is not updated anymore.

**Expected behavior**
Histogram update each time the picture correction changes

 - OS: Ubuntu 18.10
 - Version : 2.7.0+1146~gf34ea1c6b

edgardoh:
Histogram is calculated using gamma input image, but sometimes it may not be available on the pixelpipe, so the histogram is not updated.

To fix it, when the input image is not available I'm using the output image. Since output is uint8_t, there may be some rounding errors.

There's still update issues with the waveform and module histograms, but to address that a redesign will be needed. Until then this seems to be a good enough solution.

Fixes #2538

### DIFF
--- a/src/develop/pixelpipe_hb.c
+++ b/src/develop/pixelpipe_hb.c
@@ -275,8 +275,10 @@ void dt_dev_pixelpipe_create_nodes(dt_dev_pixelpipe_t *pipe, dt_develop_t *dev)
     piece->histogram_params.bins_count = 256;
     piece->histogram_stats.bins_count = 0;
     piece->histogram_stats.pixels = 0;
-    piece->colors = ((module->default_colorspace(module, pipe, NULL) == iop_cs_RAW) && 
-          (pipe->image.flags & DT_IMAGE_RAW)) ? 1 : 4;
+    piece->colors
+        = ((module->default_colorspace(module, pipe, NULL) == iop_cs_RAW) && (pipe->image.flags & DT_IMAGE_RAW))
+              ? 1
+              : 4;
     piece->iscale = pipe->iscale;
     piece->iwidth = pipe->iwidth;
     piece->iheight = pipe->iheight;
@@ -663,11 +665,12 @@ error:
 }
 #endif
 
-static void _pixelpipe_pick_from_image(const float *const pixel, const dt_iop_roi_t *roi_in, 
-    cmsHTRANSFORM xform_rgb2lab, cmsHTRANSFORM xform_rgb2rgb,
-    const float *const  pick_box, const float *const  pick_point, const int pick_size,
-    float *pick_color_rgb_min, float *pick_color_rgb_max, float *pick_color_rgb_mean,
-    float *pick_color_lab_min, float *pick_color_lab_max, float *pick_color_lab_mean)
+static void _pixelpipe_pick_from_image(const float *const pixel, const dt_iop_roi_t *roi_in,
+                                       cmsHTRANSFORM xform_rgb2lab, cmsHTRANSFORM xform_rgb2rgb,
+                                       const float *const pick_box, const float *const pick_point,
+                                       const int pick_size, float *pick_color_rgb_min, float *pick_color_rgb_max,
+                                       float *pick_color_rgb_mean, float *pick_color_lab_min,
+                                       float *pick_color_lab_max, float *pick_color_lab_mean)
 {
   float picked_color_rgb_min[3];
   float picked_color_rgb_max[3];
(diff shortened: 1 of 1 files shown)


src/develop/pixelpipe_hb.c
```diff
@@ -275,8 +275,10 @@ void dt_dev_pixelpipe_create_nodes(dt_dev_pixelpipe_t *pipe, dt_develop_t *dev)
     piece->histogram_params.bins_count = 256;
     piece->histogram_stats.bins_count = 0;
     piece->histogram_stats.pixels = 0;
-    piece->colors = ((module->default_colorspace(module, pipe, NULL) == iop_cs_RAW) && 
-          (pipe->image.flags & DT_IMAGE_RAW)) ? 1 : 4;
+    piece->colors
```
TurboGit: Just a comment, this is reformatting it is always better to have this in a separate commit. Why?

Because when I see a "minor reformatting" or "style fix" commit I can unplug my brain and just merge the changes. But for every line in a standard commit I read it and try to make sense of it. Thanks.

src/develop/pixelpipe_hb.c
```diff
@@ -663,11 +665,12 @@ static void pixelpipe_picker_cl(int devid, dt_iop_module_t *module, dt_iop_buffe
 }
 #endif
 
-static void _pixelpipe_pick_from_image(const float *const pixel, const dt_iop_roi_t *roi_in, 
-    cmsHTRANSFORM xform_rgb2lab, cmsHTRANSFORM xform_rgb2rgb,
-    const float *const  pick_box, const float *const  pick_point, const int pick_size,
-    float *pick_color_rgb_min, float *pick_color_rgb_max, float *pick_color_rgb_mean,
-    float *pick_color_lab_min, float *pick_color_lab_max, float *pick_color_lab_mean)
+static void _pixelpipe_pick_from_image(const float *const pixel, const dt_iop_roi_t *roi_in,
```
TurboGit: likewise !

src/develop/pixelpipe_hb.c
```diff
@@ -792,16 +795,16 @@ static void _pixelpipe_pick_live_samples(const float *const input, const dt_iop_
                                                        DT_PROFILE_DIRECTION_OUT | DT_PROFILE_DIRECTION_DISPLAY);
   if(d_profile) display_profile = d_profile->profile;
 
-  if((histogram_type != darktable.color_profiles->display_type) || 
-      (histogram_type == DT_COLORSPACE_FILE &&
-      strcmp(histogram_filename, darktable.color_profiles->display_filename)))
+  if((histogram_type != darktable.color_profiles->display_type)
```
TurboGit: likewise...

src/develop/pixelpipe_hb.c
```diff
@@ -862,16 +865,16 @@ static void _pixelpipe_pick_primary_colorpicker(dt_develop_t *dev, const float *
...
```
TurboGit: likewise...

src/develop/pixelpipe_hb.c
edgardoh: Sorry about that, again clang-format. I think I'll stop using it. Do you want me to reverse-it?

TurboGit: No, this is now merged.

leclercj:
I'm surprised by the difference between the two histogram versions for the exact same image.

In my observations, histograms are very similar in Luminance, but Red & Blue components are significantly modified. Is there any possibility that the Red and Blue components are inverted ?

See 2 examples here:

Exemple 1 / Original histogram (from gamma input)
Capture d’écran du 2019-05-10 22-36-04

Exemple 1 / New histogram (from output)
Capture d’écran du 2019-05-10 22-36-26

Exemple 2 / Original histogram (from gamma input)
Capture d’écran du 2019-05-10 22-43-56

Exemple 2 / New histogram (from output)
Capture d’écran du 2019-05-10 22-45-56

edgardoh:
Indeed! I didn't noticed that! I'll fix it.

edgardoh:
@leclercj , I have entered #2554, could you please confirm?

leclercj:
Much better!
PR https://github.com/darktable-org/darktable/pull/2554 is very fine with me
Still a small difference visible between the two histogram versions, but this is now negligible
Thanks @edgardoh

PR #2381, opened 2019-04-12
Ui rework

TurboGit:
@aurelienpierre : see last commit on this PR, it fixes the full preview mode.

PR #2547, opened 2019-05-10
user interface tweaks

pmjdebruijn:


Nilvus:
Don't forget that darktable.css is not the real theme of the new UI. It's just here to keep (as the old UI) using font system. So your changes on font weight reduce too much on darktable-elegant themes (real themes of the new UI ; I work on grey one).

Using middle grey on background image (grey_50) was discussed and the reason is for reducing illusion optics. The PR that propose the new UI and this choice for background was here for more than 3 months and this choice was discussed. So, I'm totally against revert back this. I've reacted also on your commit for buttons (why not in this PR ?), adding your padding was a good idea but just too much (see my comment on this commit).

If you want your changes for your using, don't forget that you could make your own theme and put it in your config/themes folder.

pmjdebruijn:
Thanks for your review. And thanks for the many improvements you've made to our UI.

[0] I still have to look into the new themes. To be honest I have mixed feelings about not using system fonts, there's reason why almost all applications use system fonts, namely because people might have their own preferences in this regard. And overriding that preference, may or may not be wise. But from what I've seen from screenshots, text legibility has been decreased, mostly due to the condensed fonts.  There's a very strong relation between font width and legibility. But this is generally speaking, I still have to give the new themes a try. So I might have a change of heart regarding this :)

[1] As for the weight changes, I don't see any valid use for proper bold fonts in regular user interface text. So if this change in darktable.css creates in issue in the themes, isn't the problem really in the themes as opposed to in darktable.css? Presumably it's simply a matter of explicitly defining any custom weights in the themes? 
Is this something you can look into?

[2] As for the middle grey on the background image, I get the theory, but preventing the optical illusion problem probably only works well when most of the screen is middle grey, not just one small area. So I do get the middle grey theme variant. But, I've never really run into a practical issue in this regard, so I'm under the impression it's more or a theoretical problem than a practical one. Another concern with a middle grey theme is the text contrast, which is heavily reduced, again reducing text legibility. Not everybody has equal eyesight. But as long as it's an opt-in, that's fine.
Presuming you could look into [1], I would suggest I eventually commit this change, so that the dark theme remains mostly dark (and consistent), and that the people can switch to the grey theme if they suffer from optical illusions.

[3] As for the commit on the button size, I committed that directly, because I thought it was a no-brainer. I merely went for 3pt, because of the bold fonts, which skewed my decision, otherwise I'd probably have opted for 2pt straight away too. I've already committed the change. Thanks for pointing this out.

[4] As for local changes, I care a lot about the default experience, not customization.

Nilvus:
> Thanks for your review. And thanks for the many improvements you've made to our UI.
I just made the grey css from @aurelienpierre work.

> [0] To be honest I have mixed feelings about not using system fonts, there's reason why almost all applications use system fonts, namely because people might have their own preferences in this regard. I understand that and I have mixed feelings also for same reasons than yours. But the problem of OS font is the user as he could choose some fonts that will make a strange UI. A default font, if installed, is the better way to avoid that and guarantee keeping the Ui as it was design. But, good point is that css is editable and customisable.

> [1] As for the weight changes, I don't see any valid use for proper bold fonts in regular user interface text. So if this change in darktable.css creates in issue in the themes, isn't the problem really in the themes as opposed to in darktable.css? Presumably it's simply a matter of explicitly defining any custom weights in the themes?
> Is this something you can look into?
That's not this way to see that. By default, with the new UI, default darktable.css was darktable-elegant.css one. That was renamed to make this last theme darktable.css with OS font. After that, and to avoid repeating thing, we link elegant css to darktable.css.

So, this point is related to your point 0. So if we keep actual darktable.css by default with OS font and improve it as you want and as needed in that case, of course, other themes have to be adapted to keep them as they actually are. And yes, I can look into it. 

> [2] As for the middle grey on the background image, I get the theory, but preventing the optical illusion problem probably only works well when most of the screen is middle grey, not just one small area. So I do get the middle grey theme variant. But, I've never really run into a practical issue in this regard, so I'm under the impression it's more or a theoretical problem than a practical one. Another concern with a middle grey theme is the text contrast, which is heavily reduced, again reducing text legibility. Not everybody has equal eyesight. But as long as it's an opt-in, that's fine.
Of course, it's even better on a total grey css theme, as the one I made, accordingly with @aurelienpierre. 
We will probably never had or see an issue about middle grey image background, and just because is an optic illusion. And the purpose of optic illusion is, by default, we do not see them but that would not say we don't have wrong behavior. And having middle grey on background image should have effect on darker theme, surely less than on grey theme.

> Presuming you could look into [1], I would suggest I eventually commit this change, so that the dark theme remains mostly dark (and consistent), and that the people can switch to the grey theme if they suffer from optical illusions.
People won't see that they suffer from optical illusions, even if they have them.

> [4] As for local changes, I care a lot about the default experience, not customization.
Great, as @aurelienpierre and me !
If you want to understand why choices of the new UI was made and all discussions about that, I recommand you to read that PR (original PR of @aurelienpierre work) : https://github.com/darktable-org/darktable/pull/2037 ; PR merged after that by this PR : https://github.com/darktable-org/darktable/pull/2381

In reading you and your arguments, I'm quite sure we could make a better choice on actual css and what you propose.

@aurelienpierre, as it is your original work, of course you point of view about that will be appreciate.

pmjdebruijn:
[1] regardless of whether darktable.css is default or not, it should look acceptable, and bold fonts in regular ui text are not. Thanks for willing to take a look.

[2] with regard to the 50% grey in the dark theme, I'll do some more experimenting on my end as well. I'll probably report back on this in a week or so.

[4] I've slowly began to work my way through these PRs, but to be honest, in the first few screenshots I immediately see a problem. Most of the design work is being done on a high-dpi display, which is a best case scenario with regard to font rendering and legibility. Doing design work on a high-dpi display gives you a very positively biased view of what a lot of people are getting. Which may explain why a condensed font was even considered at all. Keep in mind that's is very hard to find any condensed font usage in any application at all, and for good reason.
While roboto may be more elegant than many other fonts, it's legibility (even non-condensed) is inferior to many other fonts (Noto, Droid and particularly DejaVu). And given that Darktable in many places uses smaller fonts than average, legibility really really should be paramount.

pmjdebruijn:
BTW, from a internationalization viewpoint, Noto may be particularly beneficial, given it's language coverage.

Nilvus:
> [1] regardless of whether darktable.css is default or not, it should look acceptable, and bold fonts in regular ui text are not. Thanks for being willing to take a look.

Of course, you're right here. The problem is just that this default css should not be here as I as said in a previous answer. But as it here, we have to improve it.

> [2] with regard to the 50% grey in the dark theme, I'll do some more experimenting on my end as well. I'll probably report back on this in a week or so.

If you want but I really think that would not be a point to change as it is based on serious scientific studies about color perception. And as you said, you want to care about user experience. And changing this background color,as @aurelienpierre said, "it's not sexy, but it's good for color perception". So it's all about user experience, even it the user doesn't realize this. And here, I'm totally agree, since I made studies on psychology. Studying color perception was part of them.

> [4] Most of the design work is being done on a high-dpi display, which is a best case scenario with regard to font rendering and legibility. Doing design work on a high-dpi display gives you a very positively biased view of what a lot of people are getting.

Yes, @aurelienpierre work on a really high-display (4k or even 8k if I remember) but I only use an external full HD display (so not considered high-display) and it looks really great on my display.

So, the main problem of all this discussion (except middle-grey background) is all about font choice, as it created a not achieved default theme that was not wanted on start when @aurelienpierre does this new UI. As all that font choice and middle grey themes was already discussed (PR was here more than 3 months), darktable-elegant themes (dark and grey) have to keep like they actually look like.

So **I would suggest to finish adapting the darktable.css theme to be better on viewing with OS font** (ie : use reduced font-weight you suggest), but keeping middle-grey background. You could just add a customize css with grey_25 for you usage if you want.

**If it's finally ok for you, it would be better I think to have 1 PR for all this, as it will need to changes darktable and darktable-elegant css to improve the first one. I can do this.**

Nilvus:
> BTW, from a internationalization viewpoint, Noto may be particularly beneficial, given it's language coverage.

This was all reviewed on #2037 PR. You can even see screens capture on most languages.

pmjdebruijn:
[1] It would probably make sense to get the bold font stuff fixed separately, as opposed to making these changes depend on each other. I'd like to get this resolved quite quickly to be honest. If you want to handle that in a seperate PR, I'm totally fine with that. 

[2] the thing is there is a tradeoff here, having 50% in the centre, creates a UI contrast that is an order of magnitude to big. You argue asif it's just an improvement without any downside. But in reality you are trading stuff around. So regardless something has to change somehow. Though I'm not quite sure exactly what yet. And this is not quite as pressing as [1].

With regard to the studies, the studies indeed prove color vision is affected by surrounding area. What the studies however do not address, is how it affects particular use-cases, like ours. So it's very easy to accidentally misrepresent a generic study into a specific application, again like ours.

[4] The big question here, is whether the elegant is going to be the new default.
I'm aware that there are people who really want to default to remain whether the operating system font is, and to be honest I'm probably in that category too. And there's a lot of good arguments for doing so. As it's really uncommon for application to define their own fonts, an deviating from the norm shouldn't be taken lightly, and it's not very nice to pull in font packages on an application install either.
As long as the new themes are optional, than any particular choice in them isn't that relevant, as it's a user opt-in.
But that said, if the elegant theme is going to the new default, then Roboto is essentially a no-go. Roboto is factually significantly less legible than many other open source fonts, so I don't see any merit in using it.

Nilvus:
I will here only answer on that thing at first : again that's not my choice (even if I agree with that choice). I only made middle grey css theme form @aurelienpierre work (so as Roboto choice). I've previously say that we could find a better choice, reading you now let me conclude that I was totally wrong.

Yes, for middle-grey, that what is surrounding image and that why the middle grey css is of course the better one and I'm quite sure it also have positive effects with just background image, of course less but see images carefully on start #2037 (same image before and after on dark theme)

And the elegant theme are not optional, I will not explain that again. And you also could read after that thread about the New UI on pixl.us and see what people think about it. The two only downsides were fonts and modules icons, that's also why you see darktable icons theme and the darktable.css with OS font (this theme was just the elegant theme with OS font, again).

I will let all this by now as it is. It seems that another choice of what you want is not possible (coming after release of the UI that was discussed as a PR a long time). 
Here it's not who is right or not, I will just repeat it a last time, font and middle-grey background have been discussed more than 3 months by many developers and users...

And about font choice, I agree there's a lot of good arguments for OS font as for using a font like Roboto (or Noto...)... That's just the not same arguments.

By now, I'm on for just [1] point.

pmjdebruijn:
The grey theme isn't an option as default, while it may be fine for some people with exceptionally good eyesight, there's reason why almost no application (nor OS default themes) have a 50% grey UI. The problem again is text legibility. 

Text legibility is all about local contrast, and having a 50% grey UI, means you only have a maximum contrast of 50% of a display's brightness, which is not a lot. This is exactly why most themes are near black text on a light grey background, or near white text on a dark grey background, which in both cases usually ends up being a contrast of 70-80% of a display's brightness.

Nilvus:
> there's reason why almost no applications (nor OS default themes) have a 50% grey UI. 
The problem again is text legibility.

Oh, I have used since more than 10 years a Mac and default UI of MacOS is nearer to the middle grey than a dark one. Apple has only quite recently make a move to add a dark theme... Just an example !

> Text legibility is all about local contrast, and having a 50% grey UI, means you only have a maximum contrast of 50% of a display's brightness, which is not a lot. This is exactly why most themes are near black text on a light grey background, or near white text on a dark grey background, which in both cases usually ends up being a contrast of 70-80% of a display's brightness.

You're right on the first point, it's of course more difficult to have good text legibility on an a grey UI. I know that quite good as I spend some time to find a good contrast for that reason. And that's why my grey theme is not totally based for the main color of a middle grey. But about that and choice of color, dark themes are more than marketing and pleasant view (like for example colored sound on most popular hifi systems). Two examples : Aperture from Apple always have a grey interface and many users used that. Even Lightroom was based on a grey one on start. It's on 2008 that they change to a dark one. And even at this time, choice of dark theme was full of critics, like today for the opposite choice...
See this article (in french) for example of this Adobe choice : http://photofloue.net/2008/09/13/modifier-interface-lightroom/

aurelienpierre:
> But from what I've seen from screenshots, text legibility has been decreased, mostly due to the condensed fonts. There's a very strong relation between font width and legibility. But this is generally speaking, I still have to give the new themes a try. So I might have a change of heart regarding this :)

The trade-off here is squeezing more text in a minimum amount of space (to keep the image space maximal) vs. making it easier to read. Then, we have to consider the difference between high DPI screens (> 200 DPI, meaning 2.5 K and above at 15 inches), which deal nicely with thin and condensed fonts, vs. low DPI screens (< 160 DPI, meaning HD 1080 and below at 15 inches) which does not. The reason seems to be they perform font anti-aliasing by blurring, so the spatial radius of the blur is more disturbing as the letter size decreases at lower DPIs.

I agree that the app should honor the user settings for text-based software (text editors and publishing softs), but I think the case of image editors is different because we deal with smaller blocks of text and need to free as much space as possible for the main content… e.g. the picture.

The purpose of using Roboto is this font is designed for high DPI display and has every requested variant + language support, while being open source. While it's proabably not everyone's taste, it's purely a design choice that fits the requirements of having a consistent and predictible darktable experience, no matter the OS and language.

> As for the middle grey on the background image, I get the theory, but preventing the optical illusion problem probably only works well when most of the screen is middle grey, not just one small area

Just test it by comparing the thumbnail brightness (displayed on a 15 % background) to the main preview window (on a 50 % one), in the darkroom. The thumbnail looks "better" on the dark background, which is what we want to avoid (you don't edit an audio master on bass-boosted speakers, you use neutral monitoring speakers – it's the same idea here). The amount of visible background is user-defined with the border size, and since all the panels can now be hidden with a shortcut, you can have pretty much the whole frame at the right shade while keeping maximum contrast for the text in sidebars.



> Text legibility is all about local contrast, and having a 50% grey UI, means you only have a maximum contrast of 50% of a display's brightness, which is not a lot. This is exactly why most themes are near black text on a light grey background, or near white text on a dark grey background, which in both cases usually ends up being a contrast of 70-80% of a display's brightness.

I would agree with you for a text editor, but, again, our primary content here is pictures and we should aim at making the image proofing as reliable as possible, meaning reduce every chance of possible optical illusions. There is a reason why early versions of Photoshop were grey, and the newer darker versions are only marketing choices with a bit of trend-following. Every graphic designer I know who uses PS for a living works with the grey UI.

PR #2550, opened 2019-05-10
Fix OpenMP 4.0 issues with GCC9

cryptomilk:
darktable doesn't build with GCC9 on Feodra 30.

More details at https://redmine.darktable.org/issues/12568#note-14

### DIFF
--- a/src/chart/dtcairo.c
+++ b/src/chart/dtcairo.c
@@ -187,7 +187,10 @@ cairo_surface_t *cairo_surface_create_from_xyz_data(const float *const image, co
   unsigned char *rgbbuf = (unsigned char *)malloc(sizeof(unsigned char) * height * width * 4);
 
 #ifdef _OPENMP
-#pragma omp parallel for schedule(static) default(none) shared(rgbbuf)
+#pragma omp parallel for default(none) \
+  dt_omp_firstprivate(height, image, width) \
+  shared(rgbbuf) \
+  schedule(static)
 #endif
   for(int y = 0; y < height; y++)
   {
--- a/src/chart/main.c
+++ b/src/chart/main.c
@@ -1426,10 +1426,13 @@ static void get_xyz_sample_from_image(const image_t *const image, float shrink, 
   double sample_x = 0.0, sample_y = 0.0, sample_z = 0.0;
   size_t n_samples = 0;
 #ifdef _OPENMP
-#pragma omp parallel for schedule(static) default(none)                                                           \
-    shared(corners, x_start, y_start, x_end, y_end, delta_x_top, delta_y_top, delta_x_bottom, delta_y_bottom,     \
-           delta_x_left, delta_y_left, delta_x_right,                                                             \
-           delta_y_right) reduction(+ : n_samples, sample_x, sample_y, sample_z)
+#pragma omp parallel for default(none) \
+  dt_omp_firstprivate(image) \
+  shared(corners, x_start, y_start, x_end, y_end, delta_x_top, delta_y_top, \
+         delta_x_bottom, delta_y_bottom, delta_x_left, delta_y_left, \
+         delta_x_right, delta_y_right) \
+  reduction(+ : n_samples, sample_x, sample_y, sample_z) \
+  schedule(static) 
 #endif
   for(int y = y_start; y < y_end; y++)
     for(int x = x_start; x < x_end; x++)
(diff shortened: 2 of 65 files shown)


cryptomilk:
Looks like this should be addressed like:

https://github.com/darktable-org/rawspeed/blob/develop/src/config.h.in#L58-L80

rawfiner:
Wouldn't it be better to use firstprivate for the consts variables instead of shared?

cryptomilk:
Yes, but this is broken in older OpenMP versions, see the link to rawspeed.

We need a macro which uses firstprivate() if we can compile it with const variables. So we need a configure check for that.

rawfiner:
Oh ok, I did not know that! Thanks for the information. :-)
So indeed, using a macro makes sense

cryptomilk:
I will do some cmake cleanup to implement that. Once that is in place help is needed to fix this everywhere. If you grep there are more than 400 of those lines which need fixing.

cryptomilk:
I will look into it later today.

cryptomilk:
See https://redmine.darktable.org/issues/12568#note-14 for more details

darix:
we should probably backport this to 2.6.

cryptomilk:
> If a SHARED variable in a parallel region is read by the threads executing the region, but not written to by any of the threads, then specify that variable to be FIRSTPRIVATE instead of SHARED. This avoids accessing the variable by dereferencing a pointer, and avoids cache conflicts.

http://docs.oracle.com/cd/E19059-01/stud.10/819-0501/7_tuning.html

cryptomilk:
Could someone please do a review and maybe already merge it. Then we can fix the rest. I just want to make sure that the approach is fine ...

cryptomilk:
As rawspeed requires OpenMP 4.0 I've also check for that version, this means one build fails becaue the OpenMP version is only 3.1. The GTK failures I don't understand, could someone please help who knows that stuff?

cryptomilk:
@TurboGit Could you please look into the GTK issue?

TurboGit:
The GTK issues are amazing !!!! I'll have a look.

TurboGit:
Looks like a bad file naming which conflict with standard Gtk one (cairo).

Please rename `chart/cairo.{hc}` to `chart/dtcairo.{hc},` you'll need to also fix the CMakeFile.txt accordingly.

cryptomilk:
@TurboGit What about OpenMP 4.0, one build fails because it doesn't provide it. However if you look into rawspeed it also requires at least OpenMP 4.0.

cryptomilk:
@TurboGit Ok, Roman helped me to address the last issue. It would be great if you could look through that PR and say if the approach is fine. There are hundreds code points which need to be updated for OpenMP 4.0. I've only touched 4 files till now. I just don't to fix all and then I need to change the same or something like that ...

TurboGit:
This sounds good to me.

TurboGit:
This sounds good to me.

If I understand correctly dt_omp_firstprivate() is void on OpenMP < 4.0 and set firstprivate() otherwise.

Now, can you summarize why this is needed? Is that because OpenMP has some more checks and require those firstprivate() variable to be tagged?

cryptomilk:
Quoting Roman:

> There was a change in the OpenMP spec after 3.1 (i.e. in openmp 4). gcc9 has decided to finally adopt that change. Roughly, if you use `const` variable, and use `default(none)`, then that was ok with openmp3.1, but is not ok with openmp4, and you need to explicitly specify that variable as either `shared()` (the previous implicit default), or `firstprivate()` (the better choice, actually).

TurboGit:
If I understand correctly we are speaking of LLVM < 7.0 which are rejecting firstprivate. Right?

Version 6.0 was out in July 2018, so I suppose we still need to support it. If this is correct, please add a comment in the code about dt_omg_firstprivate() being needed for LLVM < 7.0 and let's go with that.

Thanks.

cryptomilk:
I've added the comment you requested.

cryptomilk:
@pobry This is ready. It was a bit of work ... :-)

rawfiner:
Thank you for doing this work @cryptomilk !

TurboGit:
I'm a bit late in the game :) Thanks a lot for the work and good to see this integrated.

cryptomilk:
You're welcome :-)

Issue #2552, opened 2019-05-10
Darkroom: 'crop and rotate' should not allow to drop a corner outside the original picture


leclercj:
**Describe the bug**
In darkroom, the 'crop and rotate' module should not allow to drop a corner outside the original picture

**To Reproduce**
1. Go to 'Darkroom'
2. Open the 'crop and rotate' module
3. Select a corner of the image and drag it to a new position.
4. See possibility to drop the corner outside the original picture
5. See also possibility to define a picture with a null dimension (0 pixel)

**Expected behavior**
1. The corner drag & drop is boxed within picture boundaries.
2. A minimum size should be kept (1 pixel ?)

**Screenshots**
![Capture d’écran du 2019-05-10 23-52-44](https://user-images.githubusercontent.com/22872204/57558633-bddd9600-737e-11e9-803c-242e64b88fe5.png)

**Platform (please complete the following information):**
 - OS: Linux (Ubuntu 18.10)
 - Same behaviour with dt 2.6.2 or dt 2.7 master

**Additional context**
Not a big deal as this is not triggering any crash, rather a suggestion for minor UI improvement

leclercj:
I'll try to propose a fix for that. Seems a good exercise for me to learn more about dt dev.

PR #2555, opened 2019-05-11
Fix 'crop and rotate' to not exceed picture boudaries

leclercj:
This is a proposal to fix https://github.com/darktable-org/darktable/issues/2552


TurboGit:
Looks good and working fine. Maybe we should allow for a bit smaller area in the crop? Not sure, I'll merge and we'll see how this is perceived. Thanks.

PR #2557, opened 2019-05-11
GMIC integration

edgardoh:
This is another attempt to integrate GMIC with darktable.

In this approach there's a new module 'gmic' that reads darktable-GMIC commands from any .dtgmic and custom commands from any .gmic file located in {config directory}/GMIC

There's a darktable_gmic_commands.dtgmic and gmic_custom_commands.gmic located in examples/gmic on this PR that describes the syntax and has some commands as example.

The darktable-gmic commands syntax is a subset of GMIC plus some additions specific to darktable.

Unlike previous attempts (and not mutually exclusive), this is oriented so the end user can build their own filters, is not the idea to ship any filter.

The darktable-gmic and gmic custom commands are loaded on start up. The custom commands are readed from .gmic files and concatenated to each other, is user responsibility to keep it as small as possible so it doesn't take too much resources.


TurboGit:
> -If darktable is build with GMIC and installed on a system without GMIC, dt crashes. This is related to the previous point, should we ship GMIC with darktable?

I don't think we should distribute it as we do not distribute Lua, libcups or libssl for example.

This is a new dependency and so we should be careful what this is dragging. For sure this must be controlled by a compile time flag to activate or not this support (as done for the print module which requires CUPS for example).

TurboGit:
> -gmic.h is required, right now is in src/external/gmic.h, maybe is better a GMIC sub-directory? Or a submodule?

Why? Can't we use the one installed on the system when installing libgmic-dev?

TurboGit:
> This is another attempt to integrate GMIC with darktable.

Just curious I don't remember having seen a previous attempt. Did I missed something?

edgardoh:
> -If darktable is build with GMIC and installed on a system without GMIC,
> dt crashes. This is related to the previous point, should we ship GMIC with
> darktable?
>
> I don't think we should distribute it as we do not distribute Lua, libcups
> or libssl for example.
>
I agree, this is by far the most delicate point of this PR, and I don't
know the right way here. If we don't ship it and the user install a version
compiled with GMIC on a system without libgmic, dt will crash, I don't know
a way around it.
If we ship it, we can mess with any current GMIC installation the user may
have.
Maybe someone here has more knowledge of this and can bring some light into
the matter.

This is a new dependency and so we should be careful what this is dragging.
> For sure this must be controlled by a compile time flag to activate or not
> this support (as done for the print module which requires CUPS for example).
>
It is controlled at compile time, so it can be build without GMIC.

edgardoh:
> -gmic.h is required, right now is in src/external/gmic.h, maybe is better
> a GMIC sub-directory? Or a submodule?
>
> Why? Can't we use the one installed on the system when installing
> libgmic-dev?
>
I didn't look into that. If is there I'll change it.

edgardoh:
> This is another attempt to integrate GMIC with darktable.
>
> Just curious I don't remember having seen a previous attempt. Did I missed
> something?
>
I think you were part of the discussion, but it was a while ago (before the
iop custom order) so I may be wrong. I don't even remember who author the
PR. I do remember that the GMIC commands were hard-coded into dt, this one
read it from file.

rawfiner:
@TurboGit previous attempt was by @rabauke. He talked about it on mailing list, but without PR I think. You can see the email here: https://www.mail-archive.com/darktable-dev@lists.darktable.org/msg03238.html

edgardoh:
Yes, that was it. And now that modules can be moved I don't have any
objections.

TurboGit:
@rawfiner : ok, my memory is very bad :) thanks for the link.

TurboGit:
> This is a new dependency and so we should be careful what this is dragging.

Right, if we have a new dependency then we will document this properly into the RELEASE_NOTES for the packager to be aware of it.

TurboGit:
> I agree, this is by far the most delicate point of this PR, and I don't know the right way here. If we don't ship it and the user install a version compiled with GMIC on a system without libgmic, dt will crash, I don't know a way around it.

We do not ship dt in binary format so not issue.

This will be taken care by packagers on Linux hence the important point that we should document the new dependency and so if the package is correctly done then installing dt will install also the needed libraries.

For Windows and Mac I suppose that the maintainers will build the needed library and distribute it with the binary package. All in all there is no way a dt built with GMic will be run on a machine without the needed library. Again, as for CUPS or Lua BTW, no magic.

rabauke:
I played around with my GMIC module quite some time, see https://github.com/rabauke/darktable/tree/gmic I think it is a bad idea to copy GMIC source files into darktable's source tree. See my changes in CMakeLists.txt, cmake/modules/FindGMIC.cmake and src/CMakeLists.txt at
https://github.com/rabauke/darktable/commit/c5adee4b37bb7ed1ac41654e6755b4f33f2398dc how to search for a local system's installation of LibGMIC at built time.

I liked the idea to incorporate this powerful tool into darktable. I will, however, not open a pull request to include this into main nor I am (no longer) convinced that we should have something like this in darktable for several reasons. 

- For many (of the more useful) GMIC filters modules with similar functionality already exist in darktable (even color looks via 3d-lut files now).
- The darktable modules often give better results (just my subjective valuation).
- Many GMIC modules are of little use for photo editing but are of highly artistic character.
- GMIC filters always(?) clamp RGB values to [0, 1], which is different from darktable's approach to allow unnormalized RGB values along the pixel pipe and to normalize at the end.
- For some GMIC modules it is notoriously difficult to calculate a relalistic preview image that is on a different scale than the finally exported image, as some filters are rather sensitive to fine details.
- My darktable integration of the GMIC library does not take into account the zoom level of the center preview image, which has consequences e.g. for the sharpening filters. A preview of a GMIC sharpening filter at, let's say, 25% looks much sharper than the final (exported at 100%) down scaled to 25%.

edgardoh:
Most (if not all) of GMIC filters are developed for GIMP, so they take as input an sRGB [0..255] image, and they don't care much about image scale as that is not needed in GIMP. 
But the idea behind this is not to re-use the GIMP-GMIC filter but develop new ones or adapt the existent ones to darktable.
The sharpen [texture] in here is an example of how it can be adapted to dt.

phweyland:
> I played around with my GMIC module quite some time, see https://github.com/rabauke/darktable/tree/gmic

I've GMIC cli running on windows.
I've built your branch locally (msys2).
The build doesn't detect GMIC. This sounds normal as GMIC is not installed on msys2.
I have 2 questions.
1. All the cmake/modules/FindXXXX.cmake file are refering to msys2 resources (packages installed on msys2). How this can detect if GMIC is available on Windows ?
2. Any tip to have GMIC on msys2 ? Could I use resources from https://gmic.eu/download.shtml ?

rabauke:
@phweyland I did not jet try to compile darktable on windows. I speculate, you have to compile and install gmic from its sources. This should also provide the development files for gmic not just the binaries.

phweyland:
>  I did not jet try to compile darktable on windows. I speculate, you have to compile and install gmic from its sources. This should also provide the development files for gmic not just the binaries.

It's what I did under msys2. During gmic build, libgmic.so and gmic.h are copied where expected (mingw64/lib amd mingw64/include) . I had to copy also libgmic.a to make the dt build recognize the presence of gmic library. So this part seems ok so far.
Where I'm stuck now is that gmic.h doesn't cover ICmg interfaces... But this is not related to this PR, only the PR #2534.

rabauke:
@phweyland For CImg.h you have to install the right package or get the source from https://framagit.org/dtschump/CImg. On Ubuntu I just installed the cimg-dev package.

edgardoh:
Time to ask for help:

The point() parameter has a GUI, so the user can set the position on the image. The coordinates are in %, so point(50,50) means the center of the image.
Unlike other modules, this coordinate refers to the image in that point of the pipe, so I don't need to take care of distortions for previous modules, but I do need to distord-transform for the following modules when displaying the point or reading the mouse position.
So far I wasn't able to do it, all I can do is transform for all modules, including the previous ones, but in this case that's not correct.

To try it, the Blur [Angular] filter has a point.

phweyland:
>  For CImg.h you have to install the right package

Not sure to understand properly your suggestion. But I already use this CImg.h file without problem, excepted that it includes the CImg code into dt and @turbogit wants to avoid this. He wants the usage of standard library and to be able to check if the library is available or not on the target OS.
However this is not an issue for that gmic PR.

rabauke:
@phweyland The header file `CImg.h` should **not** be included into the darktabe source. As for every 3rd party source, the system where darktabe is compiled has to be configured such that all required external libraries (with their corresponding header files) can be found by the compiler/build system. 

On a typical Debian-based Linux distribution, `CImg.h` is provided by installing the package `cimg-dev`, which copies the header file to `/usr/include`. On a Windows system with msys2 packman can be used to install (common) packages. For the CImg library, however, a manual installation is required. Can not describe the procedure in detail as I am currently not on Windows.

edgardoh:
Could you please move this discussion to the right thread?

phweyland:
> Also, could you try building my PR? It works fine on a clean linux installation, so it should work on windows too.

I get the errors below. I have the libgmic library installed on the msys2 environment but not its own dependencies apparently.
NB: that's half the problem, because having these libraries on msys2 doesn't mean that they are installed on the target windows system. 
We may need the @peterbud's help ...


```
[ 37%] Linking C shared library libdarktable.dll
C:/msys64/mingw64/lib/libgmic.a(libgmic.o):gmic.cpp:(.text$_ZN12cimg_library8CImgListIfE3FFTEb[_ZN12cimg_library8CImgListIfE3FFTEb]+0xf6): undefined reference to `fftw_destroy_plan'
C:/msys64/mingw64/lib/libgmic.a(libgmic.o):gmic.cpp:(.text$_ZN12cimg_library8CImgListIfE3FFTEb[_ZN12cimg_library8CImgListIfE3FFTEb]+0xfe): undefined reference to `fftw_free'
C:/msys64/mingw64/lib/libgmic.a(libgmic.o):gmic.cpp:(.text$_ZN12cimg_library8CImgListIfE3FFTEb[_ZN12cimg_library8CImgListIfE3FFTEb]+0x103): undefined reference to `fftw_cleanup_threads'
[...]
```

edgardoh:
As far as I understand it (and correct me if I'm wrong) is not dt responsibility to have the dependencies installed, so as the development goes you can install fft on your system or have a libgmic version compiled without fft and things should work.

phweyland:
>  is not dt responsibility to have the dependencies installed

You are probably right, but I've understood that at build time the dependencies should be checked. If not how to ensure the dt will work after install ?
On Linux, when you build dt, the build checks the presence of the library on your system and includes or not the dependent modules. On windows (I don't know about mac) the OS is not windows itself. So the checking is fake somehow. This is not clear for me.

> you can install fft on your system or have a libgmic version compiled without fft and things should work
This is right. I've just to learn how to do it (both ways).

phweyland:
> could you try building my PR?

Some progress. I have added in FindGMIC.cmake this:

```
if(WIN32)
  find_library(FFTW_LIBRARY NAMES fftw3)
endif(WIN32)
```

So the build is going a bit further> I get now these errors:

```
Scanning dependencies of target gmic_dt
[ 85%] Building C object src/iop/CMakeFiles/gmic_dt.dir/introspection_gmic_dt.c.obj
In file included from C:/msys64/home/philippe/dt-gmic/build/src/iop/introspection_gmic_dt.c:66:
In function 'dt_gmic_get_command',
    inlined from 'process' at C:/msys64/home/philippe/dt-gmic/src/iop/gmic_dt.c:1347:19:
C:/msys64/home/philippe/dt-gmic/src/iop/gmic_dt.c:344:3: error: 'strncpy' output truncated before terminating nul copying as many bytes from a string as its length [-Werror=stringop-truncation]
   strncpy(command, gmic_command->command, command_len);
   ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
C:/msys64/home/philippe/dt-gmic/src/iop/gmic_dt.c: In function 'process':
C:/msys64/home/philippe/dt-gmic/src/iop/gmic_dt.c:342:24: note: length computed here
   size_t command_len = strlen(gmic_command->command);
                        ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~
In function 'dt_gmic_get_command',
    inlined from 'process' at C:/msys64/home/philippe/dt-gmic/src/iop/gmic_dt.c:1347:19:
C:/msys64/home/philippe/dt-gmic/src/iop/gmic_dt.c:462:9: error: 'strncpy' output truncated before terminating nul copying as many bytes from a string as its length [-Werror=stringop-truncation]
         strncpy(tmp + index, str1, len);
         ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
C:/msys64/home/philippe/dt-gmic/src/iop/gmic_dt.c: In function 'process':
C:/msys64/home/philippe/dt-gmic/src/iop/gmic_dt.c:456:25: note: length computed here
         const int len = strlen(str1);
                         ^~~~~~~~~~~~
In function 'dt_gmic_get_command',
    inlined from 'process' at C:/msys64/home/philippe/dt-gmic/src/iop/gmic_dt.c:1347:19:
C:/msys64/home/philippe/dt-gmic/src/iop/gmic_dt.c:434:11: error: 'strncpy' output truncated before terminating nul copying as many bytes from a string as its length [-Werror=stringop-truncation]
           strncpy(tmp + index, str1, len);
           ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
C:/msys64/home/philippe/dt-gmic/src/iop/gmic_dt.c: In function 'process':
C:/msys64/home/philippe/dt-gmic/src/iop/gmic_dt.c:428:27: note: length computed here
           const int len = strlen(str1);
                           ^~~~~~~~~~~~
cc1.exe: all warnings being treated as errors
make[2]: *** [src/iop/CMakeFiles/gmic_dt.dir/build.make:68: src/iop/CMakeFiles/gmic_dt.dir/introspection_gmic_dt.c.obj] Error 1
make[1]: *** [CMakeFiles/Makefile2:7013: src/iop/CMakeFiles/gmic_dt.dir/all] Error 2

```

edgardoh:
Other than the build in Windows and Mac this is ready for review.

phweyland:
I've tried again to build on windows.
I get the same errors as above (strncpy).
Is it something which should be accepted by the build ?

edgardoh:
Never seen that error before, but it seems to have a point.

Should be fixed now, do you have the entire log for this errors?

phweyland:
> Should be fixed now, do you have the entire log for this errors?

That was the only one. I launch the build again.

phweyland:
Unfortunately it gives only 2 errors at a time ...

```
[ 85%] Building C object src/iop/CMakeFiles/gmic_dt.dir/introspection_gmic_dt.c.obj
In file included from C:/msys64/home/philippe/dt-gmic/build/src/iop/introspection_gmic_dt.c:66:
In function 'dt_gmic_get_command.isra.16',
    inlined from 'process' at C:/msys64/home/philippe/dt-gmic/src/iop/gmic_dt.c:1379:19:
C:/msys64/home/philippe/dt-gmic/src/iop/gmic_dt.c:468:9: error: 'strncpy' output truncated before terminating nul copying as many bytes from a string as its length [-Werror=stringop-truncation]
         strncpy(tmp + index, str1, len);
         ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
C:/msys64/home/philippe/dt-gmic/src/iop/gmic_dt.c: In function 'process':
C:/msys64/home/philippe/dt-gmic/src/iop/gmic_dt.c:462:25: note: length computed here
         const int len = strlen(str1);
                         ^~~~~~~~~~~~
In function 'dt_gmic_get_command.isra.16',
    inlined from 'process' at C:/msys64/home/philippe/dt-gmic/src/iop/gmic_dt.c:1379:19:
C:/msys64/home/philippe/dt-gmic/src/iop/gmic_dt.c:440:11: error: 'strncpy' output truncated before terminating nul copying as many bytes from a string as its length [-Werror=stringop-truncation]
           strncpy(tmp + index, str1, len);
           ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
C:/msys64/home/philippe/dt-gmic/src/iop/gmic_dt.c: In function 'process':
C:/msys64/home/philippe/dt-gmic/src/iop/gmic_dt.c:434:27: note: length computed here
           const int len = strlen(str1);
                           ^~~~~~~~~~~~
cc1.exe: all warnings being treated as errors
make[2]: *** [src/iop/CMakeFiles/gmic_dt.dir/build.make:68: src/iop/CMakeFiles/gmic_dt.dir/introspection_gmic_dt.c.obj] Error 1
make[1]: *** [CMakeFiles/Makefile2:7013: src/iop/CMakeFiles/gmic_dt.dir/all] Error 2

```

peterbud:
So my POV:
We don't have gmic library in MSYS2, and that's a prerequisite for any subsequent work. @phweyland if you really want to work on this integration your first step is to make sure you can build and package gmic lib properly for MSYS2, just like nay other library which we are using for dt, like exiv2, pugixml, graphicsmagick etc.
Your starting point for that is here: https://github.com/msys2/msys2/wiki

There is a functioning [GIMP package](https://github.com/msys2/MINGW-packages/blob/master/mingw-w64-gimp/PKGBUILD), maybe you just need to enable this library for the build? I don't know anything about GIMP unfortunately.

Without that gmic library the Windows version of dt won't be able to support this scenario, and the Windows version of dt should be built without this feature.

Once you have this library available we can discuss how to package it with the installer, and how to detect whether GIMP is available on the given machine etc.

phweyland:
> Your starting point for that is here: https://github.com/msys2/msys2/wiki

Thanks for the answer. May be a long way but I'll try, sure.

For the PR #2534 I've already built gmic lib (on mwing64) but the resulting dll fails. Could you have a look on this ? I don't understand  that the static library works well while the dll fails.

edgardoh:
I have changed all strncpy with g_strlcpy so it should work now.

phweyland:
> I have changed all strncpy with g_strlcpy so it should work now.

Yes, the build is ok now. I'll check also how the gmic calls works.

phweyland:
> I'll check also how the gmic calls works.

dt starts. I see the gmic module but I haven't gmic command neither *.dtgmic files.
There is no such file in the dt_gmic branch. What the user should do at this point ?
Do you have such files ready to use ?

edgardoh:
In this PR you have /examples/GMIC, copy the GMIC dir to your config dir.

phweyland:
> In this PR you have /examples/GMIC, copy the GMIC dir to your config dir.

Sorry, I should seen that. Thanks.
The gmic module works on windows (with libgmic.dll). So the solution for my PR issue should be somewhere in your code! :-)

edgardoh:
Great!

When you call gmic, instead of using a local variable for the command try to malloc() it.

edgardoh:
In this PR you have /examples/GMIC, copy the GMIC dir to your config dir.

moy:
> Unfortunately it gives only 2 errors at a time ...

`make -k` is your friend to get more errors. By default, `make` stops at the first error. You get more than one with `make -j N` because when `make` decides to stop, it still terminates the N-1 ongoing jobs. `make -k` will continue until the only jobs to do depend on the one who fail (in practice: it compiles all files, but doesn't attempt a link since some `.o` files are missing).

phweyland:
> Your starting point for that is here: https://github.com/msys2/msys2/wiki

Started to work on this and opened the PR [#5518](https://github.com/msys2/MINGW-packages/pull/5518).
Still WIP but no reason not to make it work.

aurelienpierre:
Guys, I really think Gmic is bad news in darktable. Aside from it being a random collection of toy filters, it has no GPU codepath and the worst possible UI. Way to shoot yourself in the foot.

What is the problem you try to solve with Gmic ?

edgardoh:
@TurboGit , can you remove the wip label? It seems that I don't have rights for that.

TurboGit:
@edgardoh : wip removed. there is a conflict to resolve on iop_order.c.

edgardoh:
@TurboGit , conflict resolved.

I'm not too sure about this right now, if GMIC was already integrated to dt I'll probably push for this feature, I find it to be a nice addition to darktable.
But since it seems that there's no interest in this, dealing with a new dependency for something that may not be used doesn't seem productive.

Let's keep this open to see if someone jumps in, if not I'll close it eventually.

Issue #2556, opened 2019-05-11
Clear Search module field when module(s) found, or...


ClaesGitHub:
**Is your feature request related to a problem? Please describe.**

I appreciate the _Search module_ function very much
(and yes, I know that there is a _Clear search field button_
to erase the text entered, and reset the modules list to its original state).

**Describe the solution you'd like**

I believe that the work flow would be faster if the search field could be cleared (and the modules list enters its normal state again) in more ways, for instance if the user clicks on a tab (like the tone group, or the colour group, or ...).

edgardoh:
I did try to clear the search text when the search if finished, but I couldn't find the right definition of 'search is finished'
But when the modules group change I think that we can clear the text, if that's what you mean.

ClaesGitHub:
(Why is it so difficult to describe something in words???:-)
Please let me try. Here is a typical Use Case:
a) I open a new RAW file i darktable.
b) I enter _high_ in the _search module_ field.
c) Three modules show (highpass, shadows and highlights, and highlight reconstruction).
d) I do something in highpass.
e) Now I would like to do something in the _tone group_, and click the _tone group_ tab.
f) Nothing happens, because the _search module_ field still contains the word _high_.
g) I have to explicitly click _clear text_ to make the modules in the _tone group_ show,
and I feel that is an unnecessary step for the user to take.
-- IF a click on a group tab would clear the text in the search module, I feel we would have a much more smooth work flow.

edgardoh:
Yes, that's what I meant. I'll see what I can do.

AxelG-DE:
Than we are all on the same page, I am with what @edgardoh said and I have the impression, that is the same, what @ClaesGitHub also would like to have.

#metoo :smile: 
(plesase nobody challenge me on political incorrectness ;-) )

edgardoh:
I just entered #2560, could you please confirm?

AxelG-DE:
@edgardoh 
It works on English language, but not on German. I hope this gives you some useful feedback

edgardoh:
Are you sure? Language should not be relevant here.

AxelG-DE:
I particularly tried.
First I wanted to say "oh does not work" than I changed to English and it worked...

AxelG-DE:
sorry, to cause you more headaches

AxelG-DE:
@edgardoh 

as you asked me, I checked out to #2560 again and I have to admit, now it works also in German language, as you supposed.

[...]

PR #2560, opened 2019-05-11
clear search text on module group change

edgardoh:
This clear the darkroom search text when module group change.

Addresses #2556


TurboGit:
Looks good! Thanks.

PR #2561, opened 2019-05-12
cmake: Fail if we don't have llvm for testbuild opencl programs

cryptomilk:
This is just a small cleanup to directly fail we want to testbuild but don't have the right tools available.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -147,7 +147,6 @@ if(WIN32)
   set(USE_KWALLET OFF)
   set(BUILD_CMSTEST OFF)
   set(BUILD_PRINT OFF)
-  set(TESTBUILD_OPENCL_PROGRAMS OFF)
   if(BUILD_MSYS2_INSTALL)
     add_definitions(-DMSYS2_INSTALL)
   endif()
@@ -296,8 +295,9 @@ if(VALIDATE_APPDATA_FILE)
 endif(VALIDATE_APPDATA_FILE)
 
 if(TESTBUILD_OPENCL_PROGRAMS)
-  set(TESTBUILD_OPENCL_PROGRAMS OFF)
-
+  # find_package(LLVM 3.9 CONFIG) only looks for the EXACT version, because
+  # major LLVM versions are API incompatible. So we need a macro to start
+  # looking for the latest version.
   macro(find_llvm versions)
     foreach(version ${versions})
       find_package(LLVM ${version} CONFIG)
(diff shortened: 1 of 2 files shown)


MStraeten:
This patch makes it impossible to build darktable on osx using the instructions on https://github.com/darktable-org/darktable/blob/master/packaging/macosx/BUILD.txt

cryptomilk:
Well, you should provide the error message you're getting. I don't have OSX

parafin:
I think default for TESTBUILD_OPENCL_PROGRAMS has to be changed to OFF if this change to be accepted. Otherwise I think it’s a good idea for configure system not to switch options on and off automatically, it just adds pain for packagers.

cryptomilk:
I'm fine switching it off, but I still want to see what actually fails and why :-)

parafin:
Probably it fails because that option is ON and llvm is not found.

MStraeten:
cmake .. \
 -DCMAKE_OSX_DEPLOYMENT_TARGET=10.7 \
 -DCMAKE_C_FLAGS=-I/opt/local/include/libomp \
 -DCMAKE_CXX_FLAGS="-stdlib=libc++ -I/opt/local/include/libomp" \
 -DCMAKE_LIBRARY_PATH=/opt/local/lib/libomp \
 -DBINARY_PACKAGE_BUILD=ON \
 -DRAWSPEED_ENABLE_LTO=ON

returns:
....
-- Could NOT find LLVM (missing: LLVM_DIR)
CMake Warning at CMakeLists.txt:377 (message):
  Could not find LLVM 3.9+
...

MStraeten:
but it seems to exist: 
clang --version
Apple LLVM version 10.0.1 (clang-1001.0.46.4)
Target: x86_64-apple-darwin18.5.0
Thread model: posix
InstalledDir: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin

parafin:
Yeah, I've never investigated what exactly is needed for OpenCL kernel compilation, but I doubt Xcode LLVM is enough. Probably custom-build vanilla clang is needed.

cryptomilk:
I'm sorry, but clang is a C compiler and LLVM is a compiler infrastructure. It is looking for the development tools of LLVM (the compiler infastructure) not the C compiler!

In Linux terms this means the `llvm-devel` package is missing!

homer3018:
Hello everyone.
Noob here. I'm having issues with the compilation of what's needed for OpenCL (issue #2568 ) and I was wondering if what you're discussing and the issue I face could be related.
Apologies if this is not related.

cryptomilk:
@homer3018 totally unrelated! :-)

@MStraeten Does it work if you install llvm devel files?

MStraeten:
I installed llvm-8 and clang-8 from macports but llvm-config isnt in the path. Apples compilers are detected further. Unfortunately i dont know, what‘s needed to replace the default osx build chain. The Build instructions seems to be tailored for the apple toolchain. Setting LLVM_DIR or replacing CC environment variables as i was told via irc resulted in a completely messup of the build environment.

parafin:
I wouldn't advice trying to replace Xcode compiler - it will end in tears.
I suggest to remove this functionality for Mac platform from cmake (`if(NOT APPLE)`), since I don't think anybody is interested in it.

cryptomilk:
I've change the default to OFF on Apple and Win32.

cryptomilk:
Rebased to master ...

cryptomilk:
I've rebased the patch and adopted it for master.

cryptomilk:
@TurboGit What are we doing with this?

cryptomilk:
There doesn't seem to be interest in this ...

Issue #2564, opened 2019-05-12
Piwigo export doesn't work with special characters in the password


pvalsecc:
**To Reproduce**
1. Change your password to `E36#q$tvFR&I` in piwigo's web interface
2. try to login on piwigo in darktable with this password

**Expected behavior**
Export working.

**Platform (please complete the following information):**
 - OS: linux
 - Version 2.6.2

**Additional context**
I think this is because you don't URL encode the password when doing the POST.

leclercj:
This is fixed in dev / master branch now
Thanks for this report

pvalsecc:
Thanks for the fix!

PR #2572, opened 2019-05-14
Piwigo: escape username and password strings to support usage of special characters

leclercj:
Fixes https://github.com/darktable-org/darktable/issues/2564


TurboGit:
Looks good! Thanks.

PR #2574, opened 2019-05-15
Panasonic DC-LX100M2: add noise profiles support

d2r2:
Panasonic GX9 noise profiles were copy pasted to Panasonic LX100M2, since it has same sensor. This could be considered as a temporary decision, until someone provide real data.

From my personal experience copied profiles work acceptable.


TurboGit:
@rawfiner : how does it feel to you to copy a noise profile? Since it is the same sensor it should be ok I would think.

rawfiner:
I am not confident about this. It would be better to create real profiles I think, just to be sure. The sensor may be identical, but if the firmware encodes the sensor data differently (or preprocess it differently), the noise profiles may be different...

d2r2:
@TurboGit, @rawfiner I can provide LX100M2 raw files with some noticeable noise if needed for the test.

d2r2:
@rawfiner, do we have instruction like one exists for "white balance" profiles extract, for noice profiles generation?

rawfiner:
@d2r2 you can follow this guide to create a profile: https://discuss.pixls.us/t/how-to-create-camera-noise-profiles-for-darktable/7428

d2r2:
@rawfiner, thanks! I will try to do this, and get back in few days with any result.

TurboGit:
So closing this for now.

PR #2575, opened 2019-05-15
smooth scrolling for touchpads

emeikei:
I added an option for this as some systems can apparently have issues with this.

### DIFF
--- a/data/darktableconfig.xml.in
+++ b/data/darktableconfig.xml.in
@@ -1229,6 +1229,13 @@
     <shortdescription>mouse wheel scrolls modules side panel by default</shortdescription>
     <longdescription>when enabled, use mouse wheel to scroll modules side panel.  use ctrl+alt to use mouse wheel for data entry.  when disabled, this behavior is reversed</longdescription>
   </dtconfig>
+  <dtconfig prefs="gui">
+    <name>ui_last/enable_smooth_scrolling</name>
+    <type>bool</type>
+    <default>true</default>
+    <shortdescription>enable smooth scrolling for touchpads</shortdescription>
+    <longdescription>may cause issues on some systems</longdescription>
+  </dtconfig>
   <dtconfig prefs="gui" section="darkroom">
     <name>plugins/darkroom/ui/border_size</name>
     <type>int</type>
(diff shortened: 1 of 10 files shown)


TurboGit:
Looks good. One question you said it can causes issues but it is activated by default. What is the percentage of issues? If probable maybe it is better to have it deactivated by default. I even think it is safer to have it deactivated and let user test and report about this before activating it by default. What do you think?

emeikei:
I don't know what the 'issues' are or if they're still relevant.  There's just a brief comment in the existing code. The issues question was raised in the forum but nobody responded.
My thinking is that by having it enabled by default, we stand more chance of finding out if these issues are still relevant.  And if nobody reports any issues then maybe we can eventually remove the option and just leave it enabled.

homer3018:
If needed I could give it a try using my wacom tablet. If it works nicely it'b great !

TurboGit:
@homer3018 : your wacom tablet has a touchpad or something similar? If so, yes please test and report. Thanks.

homer3018:
@TurboGit yes, it is basically a giant touch pad which supports multi touch gestures also (zoom with 2 fingers, etc.) 
I'll see if I have the chance tonight, but yes I'll test anyway and report.

TurboGit:
Any news on the test?

homer3018:
Sorry, I have had forgot about this one.
Just tested it, and I do not see any difference in the lighttable with the option set to ON (default) or OFF. I'm unsure whether or not this is coming from my setup (say maybe the Wacom tablet is not recognised as a smooth scrolling capable device ?) or something else.

In the darktable however, both the side panels can be scrolled quite smoothly, and I'm seeing this regardless of the option state...

Is there a location where I am supposed to see a difference with the option ON and OFF ? If so could you please let me know where ?

homer3018:
Interesting. I've switched back to 2.7.0+1247~g277c7aa7d which was the one I was using recently, and the scrolling (both side panels in the darkroom) do not scroll as smoothly as they did with yours. I do like the smooth scrolling :)

m93a:
How exactly is this supposed work? I installed the latest commit (`2.7.0~git1468.f762741f8`) and smooth scrolling still doesn't work for me in Lightroom. And I don't see any new options under GUI. I'm on Acer A315 (with quad-touch touchpad) runnning Ubuntu 18.04.

TurboGit:

PR #2576, opened 2019-05-15
side panel scrollin update

emeikei:
updated to allow scrolling through graphical elements in accordance with prefs and accelerator.


TurboGit:
Looks good! Thanks.

PR #2580, opened 2019-05-16
German translation: avoid random capitalization of initials

rabauke:
For consistency with other UI elements avoid random capitalization of initials.


TurboGit:
Thanks.

PR #2581, opened 2019-05-16
UI css minor tweaks

Nilvus:
This PR is related to #2547 one as it make darktable theme more suitable on text fonts too bold with OS font, as proposed by @pmjdebruijn. darktable-elegant.css has been adjusted to avoid changing his design on dark and grey themes as it's related to darktable.css.

As suggested on pixl.us, I also modify icons module alignment to center them vertically, what seems more suitable. This is for darktable icons theme (both dark and grey).

I suggest to close #2547 as this one correct font weight as wanted by @pmjdebruijn and adjust as necessary css related to darktable.css and grey background had been discussed on his PR and was decided on UI rework PR.

See this capture screen for new icons alignment  :

Capture-20190516221840-327x447

### DIFF
--- a/data/themes/darktable-elegant.css
+++ b/data/themes/darktable-elegant.css
@@ -89,6 +89,14 @@ button
 
 }
 
+#recent-collection-button,
+#history-button
+{
+  /* Exception because history and recent collections modules make use of a
+   * list of GTK buttons but we don't want them to look like action buttons */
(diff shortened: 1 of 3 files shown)


TurboGit:
Looks good to me! Thanks for the continued effort to make the new UI looks great for everyone.

PR #2583, opened 2019-05-17
cmake: Fix building darktable-chart

cryptomilk:
Don't name a header cairo.h, this creates name clashes with upstream
cario! ;-)


TurboGit:
Thanks!

Issue #2524, opened 2019-05-06
UI: color picker live samples areas displayed as small strips


MStraeten:
**Describe the bug**
live sample picker areas are displayed as small strips instead of boxes:
![image](https://user-images.githubusercontent.com/39386816/57220960-8a58df80-6ffd-11e9-806a-c679ed2ece14.png)

**Platform (please complete the following information):**
 - OS: osx, windows
 - Version  darktable 2.7.0+1123~g8555f5feb

Nilvus:
Confirmed also on Linux. I didn't find any possibility to change that on CSS. So, it's probably on GTK code that this could be changed (or with a very hidden css tag ;-) !).
@aurelienpierre or @TurboGit  : any idea ?

live samples should be like this : 
![picker](https://user-images.githubusercontent.com/45535283/57319285-89f33e00-70fc-11e9-9d76-b4dbd12490cd.png)

Nilvus:
@TurboGit : so to be closed.

TurboGit:
Closing now.

PR #2585, opened 2019-05-18
[WIP] Correct live sample size on picker module

Nilvus:
This resolve issue #2524. I find what line break this by testing some commits. This issue was cause by this commit : https://github.com/darktable-org/darktable/commit/db27f068a4dc145c2edb0109e42456b34874aa5f

@TurboGit : as I don't know anything on GTK and as it revert one line on the last commit from you, could you please see if it's ok or if there a better way as you want remove these type of lines on the commit.

I've also test to put this line instead of revert the original line, to use CSS as other lines you made : `gtk_widget_set_name(GTK_WIDGET(sample->color_patch), "live-sample-picker");`
But it seems not possible to use css for this widget. My line seems correct as when launching darktable after with GTK inspector, I find this css tag. I've use margin, padding and min-width on CSS, the live sample remains small strips.

Maybe It's possible on another way but as this line correct this issue, maybe we could just revert it back to have correct live sample.

### DIFF
--- a/src/libs/colorpicker.c
+++ b/src/libs/colorpicker.c
@@ -363,6 +363,7 @@ static void _add_sample(GtkButton *widget, gpointer self)
   gtk_box_pack_start(GTK_BOX(data->samples_container), sample->container, TRUE, TRUE, 0);
 
   sample->color_patch = gtk_drawing_area_new();
+  gtk_widget_set_size_request(sample->color_patch, DT_PIXEL_APPLY_DPI(40), -1);
   gtk_widget_set_events(sample->color_patch, GDK_ENTER_NOTIFY_MASK | GDK_LEAVE_NOTIFY_MASK | GDK_BUTTON_PRESS_MASK);
   gtk_widget_set_tooltip_text(sample->color_patch, _("hover to highlight sample on canvas, "
                                                        "click to lock sample"));


src/libs/colorpicker.c
```diff
@@ -363,6 +363,7 @@ static void _add_sample(GtkButton *widget, gpointer self)
   gtk_box_pack_start(GTK_BOX(data->samples_container), sample->container, TRUE, TRUE, 0);
 
   sample->color_patch = gtk_drawing_area_new();
+  gtk_widget_set_size_request(sample->color_patch, DT_PIXEL_APPLY_DPI(40), -1);
```
TurboGit: Indeed this is not correct as we have made sure all hard coded size are removed. I'm not sure about the proper solution but IIRC it is not possible to set size of a  gtk_drawing_area_new() from CSS, you need to place it into a container (vbox for example) and you may need to name this widget to  add min-width/min-height in CSS.

Do you feel like trying that?

Nilvus: No, I don't know how to make that. I have just tried to help as I've find where the issue was but I really don't know anything on coding except for CSS. Here I Just find that was this line.

TurboGit: Ok, let me have a look then.

TurboGit:
Fixed.

Nilvus:
> Fixed.

Is that means you find how correct that and you will commit that soon ?

Issue #2549, opened 2019-05-10
Compress history vertically 


kanyck:
New 2.7 look is great... except the history too sparse and therefore takes more vertical space, which is always shortage of.

Nilvus:
See this PR #2586

TurboGit:
Done now. Closing.

Issue #2428, opened 2019-04-20
Controls in global preferences are almost twice as big as the ones in modules


edgardoh:
Controls in global preferences are almost twice as big as the ones in modules, see in print screen the text entry compared with the one in the export module.

It should be the same or have some way to configure it.

edgardoh:
![Screenshot_2019-04-19_16-25-10](https://user-images.githubusercontent.com/16430501/56457856-fd542c00-6356-11e9-80b7-76c3e5c34700.png)

Nilvus:
dialog prefs is made of a grid, so entry, label, etc. adjust their height on the line.

The only way I see is to remove margin on this tag css :
dialog *
{
  margin: 0pt;
  padding:0;
}

On actual css from @aurelienpierre, the margin is 1pt;

This margin is here to aerate the options. See how prefs looks when margin is put on 0 :

![Capture-20190517205940-806x830](https://user-images.githubusercontent.com/45535283/57950284-1e248880-78e7-11e9-85ff-9765c04bcfca.png

edgardoh:
Looks better to me.

BTW, your message is not on github, did you send it by mail or directly on
the issue? I ask because I have lost some of my and I'm not sure why.

Nilvus:
> Looks better to me. BTW, your message is not on github, did you send it by mail or directly on the issue? I ask because I have lost some of my and I'm not sure why. El vie., 17 may. 2019 a las 16:03, Nilvus (<notifications@github.com>) 

No, I delete just after posting it as I find a better way. I will, this week-end, post a new PR with other improvements on the UI. Stay on !

Nilvus:
@edgardoh : as the PR is merged, could you please close your issue if the new prefs window is ok for you ? Thanks.

edgardoh:
Sure, and thanks for your work!

Issue #2481, opened 2019-04-29
GTK undershoot and overshoot cover content in scrollable windows (e.g. preferences)


Tracerneo:
#2380 is also present on Linux, is not fixed in #2445 

Please see https://github.com/darktable-org/darktable/pull/2445#issuecomment-487612378 for details.

**Platform:**
 - OS: Gentoo
 - darktable 2.6.2
 - GTK+ 3.24.7

aurelienpierre:
Ok then.

Nilvus:
As answer on #2445 I don't think this workaround will work on 2.6.x as it's based on new UI where all changed. It certainly need other tweaks on 2.6.x. But as this is resolve on 2.7, you certainly have to wait next major release to see this resolve (or test 2.7 development version).
@TurboGit : what do you think of that ? To be closed ?

Nilvus:
@TurboGit : this one to close also. Thanks.

TurboGit:
Closing.

PR #2586, opened 2019-05-18
Correct some CSS issues remaining on UI

Nilvus:
A new set of CSS improvements. This PR should resolve some remaining issues :

- #2428 : @edgardoh, see if it's ok for you now ? As items in preferences (entries, checkbox, labels) are all on a grid, size is only adjusted by margins between items. So here it's not possible to adjust each one separately. This seems the better compromise on my opinion.
- #2549 : @kanyck, see if it's now ok for you ? It's possible to reduce a little more but that would be too condensed. Have an aerate UI is also great, so I try to find also the better compromise here between aerate text and not too coarse history.

I also reduce size on some dropdown button arrows, especially on filter collection (I also reduce height size on entry item near this dropdown) and display dropdown on filmic that was way too big.

@aurelienpierre : could you see if it's ok for you as it change some of your choices ?

Before merging that, please test as much as you can. I made lot of reviews tests and compare with actual themes what change and if this new css breaks other part. I don't find anything on lot of modules I test, even on map and print parts (I don't test capture part as I don't have a camera compatible).

As I change lot of things, I want to be sure that nothing is broken on the UI.


data/themes/darktable.css
```diff
@@ -648,7 +666,7 @@ combobox button
   border: 1px solid transparent;
   padding: 1px;
   background-color: transparent;
-  border-radius: 2pt;
+  border-radius: 2px;
```
TurboGit: I don't think we want PX here. As far as I know PT is better as it translate properly on HiDPI screen.

edgardoh:
@Nilvus , looks the same to me, something else than building it that I should do?

data/themes/darktable.css
Nilvus: pt is normally not recommanded on screens. As css rules are maintain by w3c, see here : https://www.w3.org/Style/Examples/007/units.en.html
But as I don't use HiDPI screen (only full HD one), maybe GTK doesn't respect w3c rules. Could you confirm that ?

Nilvus: I just read some things on gtk manual. The CSS properties page related to w3c pages. And I also read that HiDPI was added to GTK with using scale factor for HiDPI so I suppose that would be good.

Nilvus:
> @Nilvus , looks the same to me, something else than building it that I should do?

That's not really looks the same to me, but that's not the same as right panel. I've made lot of tests without success. What I propose to you at first yesterday break other dialog box. But I've just had another idea that should much better. See new commit.

Nilvus:
Sorry for two last commit. Someone I forget and I didn't understand on resolving that I was just merging master css... I will stop for tonight.

Nilvus:
Issue #2481 now corrected.

Nilvus:
@edgardoh : could you test last updates of css. What do you think about these changes, especially prefs entries (and filter collection one) as it's related to your issue.

TurboGit:
@Nilvus : sounds like you are correct and I'm wrong:

https://www.w3.org/Style/Examples/007/units.en.html

pt should not be used for screen and px should be preferred.

Nilvus:
By now, I hope this last commit will end this PR. I have just improve contrast with revert back grey_55 to default css color. That make a better contrast especially on filmstrip (difficult to make it better without breaking things).

I also clean grey css as it have redundant lines with darktable.css related theme. A comment needs to be more precise so I updated it.

Nilvus:
@TurboGit : anyway I also found that 1pt=1.33px, so I can't change just the unit. For example, I have replaced font-size of blending tabs from 8pt to 8 px and see that line becoming too small. But changing it to 0.8em (or 10px or 11px, as 8pt = 10.64px) make it better.
By now, I let existing pt as it is (rendering si the most important part of that).

Nilvus:
As I hope UI is better now, see screenshots of new main updates to help review :

**Filter collection entry and arrow dropdown**
filtercollection

**Preferences dialog window**
preferenceswindow

**More condensed darkroom history**
history

**New live sample _(main work by @TurboGit on another commit, just improve text lisibility)_**
livesamples

**More reasonable size of display dropdown arrow on filmic**
filmicarrow

edgardoh:
@Nilvus , much better. Still the controls (text entry, check box) could be smaller (in height), also the space between controls. But if can't be done I can live with it.

Nilvus:
> @Nilvus , much better. Still the controls (text entry, check box) could be smaller (in height), also the space between controls. But if can't be done I can live with it.

You will have to forgot having a smaller height as the window is a grid, I could only reduce height with margin and padding already to 0. And negative values are not recognized. So for that, it's not possible without a Gtk rework...

For space, it's possible by reduce margin from this part :
#preferences_notebook scrolledwindow label,
#preferences_notebook scrolledwindow entry,
#preferences_notebook scrolledwindow button
{
  padding: 0px;
  margin: 5px 0;
}

If I reduce from 5px to 3px, see new spacing (keep quite aerate would be better so I'm not on to reduce spacing more) :

Capture-20190519181316-800x830

Nilvus:
Finally, prefs window is probably better with reducing space as last screen capture. So committed with a darktable-elegant code cleanup (many lines redundant from darktable default css and some lines forgotten and not needed, ie live-sample-data better with darktable default settings).

I also changed font-weight from 400 to normal to have the same setting (400=normal on css) on css.

PR #2587, opened 2019-05-18
lighttable : define preview frame in css

### LINKED ISSUE #2577: On culling a frame is draw around the images
On culling a frame is draw around the images, this didn't happen when I first worked on culling, no idea when it was introduced or why, but it goes against the 50% grey background rule, so at least and option should be available for it, and disabled by default.

Same goes for the 'focus' feature, but I don't have an strong opinion on it. I find it distracting when comparing two images, but is useful when rating, so I can be sure on which image the action will be applied.

AlicVB:
This fix #2577 
We now have 2 new colors in css : `preview_border_color`  and `preview_hover_border_color` They are use for full preview and culling.
- They are set to actual values in darktable.css (that means frames are drawn like before)
- In darktable-elegant.css, I've hidden the "normal color" (by setting it to transparent)
- In darktable-elegant-grey.cs, I've also hidden the hover border

Note that I've changed a little how hover border is drawn : it's now just a really thin line around the image.
screenshot with darktable-elegant : 
Capture d’écran_2019-05-18_19-09-40


Nilvus:
> This fix #2577
>     * In darktable-elegant-grey.cs, I've also hidden the hover border
> screenshot with darktable-elegant :
> Capture d’écran_2019-05-18_19-09-40

That's really great and make the preview even border and more readable with this thin line. What I don't understand is why you have hidden the hover border on grey elegant theme. As I'm using it (and made it), that seems when I'm on preview window, I don't have any differences (except exif infos displayed or not) to see which image is selected on this theme. And that doesn't make it consistent between themes.
As background image is the same on all themes (middle grey), I think it will be better to display hover the same as dark elegant theme. I propose to change that on my actual WIP PR about css by just removing the line preview_hover... on darktable-elegant-grey. Is that ok for you ?

AlicVB:
@Nilvus : no pb... I've just done that because in my mind, grey theme = absolutely no distraction. But as you say, the background is the same in each theme, so better be consistent. (and you, with @aurelienpierre,  are far more competent than me to decide that !)

Nilvus:
> @Nilvus : no pb... I've just done that because in my mind, grey theme = absolutely no distraction. But as you say, the background is the same in each theme, so better be consistent. (and you, with @aurelienpierre, are far more competent than me to decide that !)

Ok, so I will made the line back on grey theme on my next commit.

Issue #2558, opened 2019-05-11
[culling] star-rating when number-key pressed twice or come back from full screen with "W"


AxelG-DE:
**Describe the bug**
when I press e.g. "4" (to give 4 stars) twice, the star rating 4 is applied to all pictures in current culling view (rating 4 is applied to 4 pictures)

**To Reproduce**
1. Go to 'culling, 4 (or 6) pics to be displayd, star-rating in my case >=3'
2. press 4 twice on the same image
or
2b. view full screen with "W" key and after come back change the rating
3. See "error"

**Expected behavior**
do not auto-apply a rating to all pics in view
- as it might happen, one press the key twice because:
- not clear, first one got taken
- cannot see the stars well, due to background
- afraid of lag

@AlicVB if that is a feature, I'd prefer to be able to switch it off :-)

**Platform (please complete the following information):**
 - OS: [gentoo linux]
 - Version [current master +1154]

Nilvus:
I don't know if it is a feature or not but I don't understand why press twice on same image ? Even for undo, I prefer using other choice I want (for example, 3 shortcut if I want to change for 3 stars...). But as this mode is in discussion to merge with expose mode, maybe this could be related to the actual PR for this merge : #2539

AlicVB:
As @Nilvus said, there's a big rework in progress in this area.
If I understand well, you want actions to be applied just to the image under the mouse, right ? If so it involve how selection is kept (or not) in synch with culling view, and it's "under construction"

AxelG-DE:
Guys, forgive me, if my explanation wasn't very well.
What I found was, suddenly my choice of selecting 4* was applied to all images in view. So I tried to figure out, what made it happen.

I do not press the same key twice on purpose :), so I can understand the wonders of @Nilvus 

@AlicVB you are right, I want my "command" to be applied only to the pic under mouse.

As I said, when I leave "W" mode (full screen), also weird things  happen. If all this is under rework, I will also shut up for some time :smile: :smile:

Nilvus:
@AxelG-DE : indeed, that's not the same thing :smile: :smile:

What you want seems to be quite the opposite than this PR : #2545

AxelG-DE:
@Nilvus 
>What you want seems to be quite the opposite than this PR : #2545 

not really, as I was not referring to multiple selections and intended behaviour outside the selection. My issue is without particular selection and hover on one pic, all over sudden the change is applied to all in view without intending this, sorry for repeating :) 

[...]

PR #2539, opened 2019-05-08
merging expose layout into culling

AlicVB:
We currently have 2 very close layouts: expose and culling. I'd like to drop expose in favor of culling which already include almost all expose features. This would be good in terms of coding (less paths to tests/maintain/...) and for users (2 tools for the same thing)

To resume the 2 layouts (correct me if I'm false) :
Expose : show all and only the selected images. You can't move
Culling : show a determined number of images, starting from the first selected. You move inside the whole collection

If we drop completely the expose mode, we miss 2 features :
1- show non contiguous images
2- show "selection length" number of images

What I would propose is to do the same things we currently do with full-preview :
**case 1 :** number of selected images <= 1 (maybe if current selection is contiguous and it's size = nb of images of culling layout too)
- use the current culling behavior

**case 2 :** number of images > 1
- use the current culling behavior, but only navigate inside selected images (instead of all the images in the collection)

What will change ?
- we don't access to the current culling behaviors if we have more than 1 selected images => anyway, actually culling change the selection, so the selection will be lost when re-entering filemanager. imho, it's bad as user doesn't expect to loose his selection... and it's not consistent with how full preview works...
- we can't show lot of images in 1 screen like in expose mode (say you select 10 images, you will have to change the slider to show them all) => anyway, showing lot of images to compare in not really useful imho (and it's still possible with the slider for special cases)
+ we can now use culling in non-contiguous selections

**To summarize, I think we'll not loose crucial features and even extend culling features in a nice way !**

Here's a very roughly implementation of this, just as a proof of concept, and a base for discussion.
There's lot of small glitch to fix, and I've even not removed expose mode, just changed culling to what it can became with my proposal.

Don't hesitate to share your comments and concerns, (especially @edgardoh @oexler @TurboGit)

PR #2588, opened 2019-05-18
don't reset mouse_over_id

AlicVB:
currently, to ensure correct drawing in `dt_view_image_expose`, filmstrip (and culling in rare cases) reset the global `mouse_over_id` value, and set it again during the draw loop if the mouse is over a thumb. This is bad, as it will reset the value if the mouse is not here (example if the mouse is inside culling area and filmstrip is redrawn).
This is a better fix for issue #2558 , it solves some mouse_over jump and false highlight in culling too.

### DIFF
--- a/src/libs/tools/filmstrip.c
+++ b/src/libs/tools/filmstrip.c
@@ -727,7 +727,6 @@ static gboolean _lib_filmstrip_draw_callback(GtkWidget *widget, cairo_t *wcr, gp
 
   if(darktable.gui->center_tooltip == 1) darktable.gui->center_tooltip++;
 
-  int mouse_over_id = -1;
   strip->image_over = DT_VIEW_DESERT;
 
   /* fill background */
@@ -798,6 +797,7 @@ static gboolean _lib_filmstrip_draw_callback(GtkWidget *widget, cairo_t *wcr, gp
   cairo_translate(cr, empty_edge, 0.0f);
   const int before_last_exposed_id = strip->last_exposed_id;
   const int initial_mouse_over_id = strip->mouse_over_id;
+  int mouse_over_id = -1;
   int missing = 0;
 
   for(int col = 0; col < max_cols; col++)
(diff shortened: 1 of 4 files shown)


edgardoh:
Works very well, thanks.

TurboGit:
Found an issue, go to darkroom, fly over the filmstrip and lot of images are highlight.

TurboGit:
I had also a crash with this PR.

AlicVB:
@TurboGit : I've fixed the filmstrip highlight.
For the crash, I can't reproduce (and in the code, I don't see where it can crash...)

AxelG-DE:
@TurboGit @AlicVB 
I confirm, I also had this "several thumps on filmstrip are highlighted" actually darkroom and ligthtable/culling.

Now with 40d07e214 it is gone, but now I need to click again into dt-window prior keystroke
example
**situation pre PR 2588**
- I was somewhere else on second screen
- hover dt main window
- scroll works
- W or star rating only works after click into dt-main-window

**initial PR 2588** from @AlicVB **
that last step's click into the main window was not necessary

** situation with  40d07e214**
- W or star rating (after those steps above) only works after click into dt-main-window

I am fine, when you guys say, "that is the nature of the beast". Just want to report, that I found :-)

AlicVB:
@AxelG-DE : To be sure, for features, everything works like current master, but without the "jump" of the highlight in culling, right ? If so, that's what is intended :)
What you report (need to click in main windows before keystroke) is explicitly disabled in the code : when mouse leave the view, mouse_over_id is reset. Maybe we can change that, but I fear that we'll break something else ! (in any case, it's not the purpose of this PR)
Anyway, thanks a lot for all your tests !

AxelG-DE:
@AlicVB 
>@AxelG-DE : To be sure, for features, everything works like current master, but without the "jump" of the highlight in culling, right ? If so, that's what is intended :)

...and I once asked for it in #2539 :smile:

> What you report (need to click in main windows before keystroke) is explicitly disabled in the code : when mouse leave the view, mouse_over_id is reset. Maybe we can change that, but I fear that we'll break something else ! (in any case, it's not the purpose of this PR)

I am afraid we can easily have a misunderstanding here. You say it is disabled, but 43d200e worked without click and 40d07e2 requires click again (while master is with click as well)

You are very welcome!

AlicVB:
> (while master is with click as well)

As said, this is the "normal" behavior currently. This PR is only valid with the 2 commits, as the first one alone is buggy.

TurboGit:
Ok, looks good now. The crash is not related to this PR. I have reproduced without it.

AlicVB:
great (not for the crash ;) )
Thanks for all your reviews

PR #2593, opened 2019-05-19
culling : Allow to zoom/pan specific image

### LINKED ISSUE #2559: [Feature] Adding possibility to dissociate position displayed on an image when zooming on culling/expose modes
@AlicVB 
One great feature to add would be the possibility to compare images with position of viewing dissociated on zooming. To be clear, sometimes, when we zoom to compare images, zoom could make an offset on view displayed between image.

I think that a good behavior would be to use a shortcut with mouse hover the image wanted to adjust and moving the position displayed on the image without moving others. And of course, this only with shortcut pressed (like actual W to display preview on lighttable only when it pressed).

AlicVB:
This fix #2559 
If you add SHIFT key to usual behavior, only the image under the cursor is zoomed/panned :
- CTRL+SHIFT+scroll = change the zoom of current image only
- SHIFT+pan = change position of current image only

Note that once you have changed a specific image, you can reuse classic zoom/pan. All images will zoom/pan in sync, except that the specifically modified image will keep the shift you've made

### DIFF
--- a/src/views/lighttable.c
+++ b/src/views/lighttable.c
@@ -121,6 +121,8 @@ typedef struct dt_preview_surface_t
   float h_fit;
   float zoom_100;
 
+  float zoom_delta;
+
   float max_dx;
   float max_dy;
 } dt_preview_surface_t;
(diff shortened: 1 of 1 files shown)


Nilvus:
Glad to see that, really thanks for all your work on lighttable part and especially these compare options. I will test it as soon as possible.

TurboGit:
That's great enhancement! Thanks, working fine on my side.

Nilvus:
Working fine here too, thanks a lot ! This new culling mode with all recent improvements are really a great feature and I'm sure many users will love it !!!

PR #2594, opened 2019-05-19
culling : fix if grouping active

AlicVB:
only count visible images in the selection
Otherwise, starting with only 1 images (grouped), culling think there's 2 images selected (all the grouped images) and activate the "selection" mode

### DIFF
--- a/src/views/lighttable.c
+++ b/src/views/lighttable.c
@@ -3155,8 +3155,12 @@ static void _culling_scroll(dt_library_t *lib, const int up)
                 || !lib->culling_use_selection))
     {
       if(lib->culling_use_selection)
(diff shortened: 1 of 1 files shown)


src/views/lighttable.c
```diff
@@ -3137,7 +3156,7 @@ static void _culling_scroll(dt_library_t *lib, const int up)
     {
       if(lib->culling_use_selection)
         dt_control_log(_("you have reached the start of your selection (%d images)"),
```
TurboGit: I suppose here (and below) we want a string for handling single image or multiple (singular / plural).
see ngettext() usage.

AlicVB: done

Issue #2405, opened 2019-04-14
2nd preview: can it be borderless?


AxelG-DE:
**Is your feature request related to a problem? Please describe.**
Not a real Problem, just a wish for design.

I'd like to see the 2nd preview window (can I call it 2pW) is borderless, just like dt can do

**Describe the solution you'd like**
same like main window, when I activate that window and press F11 it can be borderless.

rabauke:
I am missing this feature as well. I started to figure out what has to be done to implement a full-screen mode for the 2nd preview and would like to implement this. However, I am not very familiar with GTK gui programming and might therefore need some time (which is lacking). Hints are welcome. Possibly, this is even a trivial task for somebody else.

TurboGit:
I put the second window in the desired screen and press F11 (full screen mode), it is almost borderless this way. Won't this work on your side?

AxelG-DE:
At least for me nothing special happened once I pressed F11

TurboGit:
Well which DE are you using? On GNOME Shell F11 is the "maximize" action.

AxelG-DE:
DE?

As you sufferingly know :smile: I'm using the rolling distro Gentoo with quite up to date KDE. Now on mobile phone, hence cannot check the PC now......

TurboGit:
Sorry, DE -> Desktop Environment

AxelG-DE:
Hihihi
Now you got it.... 
Update kde

I'll not hide, currently I do have issues with keystroke for screenshots (not working) . So I cannot neglect a potential risk something lies within my system...

rabauke:
@TurboGit There might be some desktop environments which provide short keys for maximizing application windows to full screen. But this is not generally the case. For example, F11 does not work for the 2nd preview on my system (Ubuntu 18.04, Gnome). 

Full screen mode for the main window is triggered via `fullscreen_key_accel_callback` in `gtk.c` if I understand the darktable code correctly, not by the desktop environment. A similar mechanism has to be implemented for the 2nd preview window as well.

Nilvus:
I test that and F11 doesn't work for 2nd preview window, as it works great for main window (or other software I use). I'm on Gnome 3.32. So as @rabauke said, it's not implemented. This would be great to implement that. When using two screens, it could be great to have the 2nd preview window on the other screen full size.

[...]

PR #2595, opened 2019-05-19
2nd window: allow fullscreen

edgardoh:
Allows fullscreen on 2nd window.

Addresses #2405

### DIFF
--- a/src/views/darkroom.c
+++ b/src/views/darkroom.c
@@ -3307,10 +3307,7 @@ static void second_window_expose(GtkWidget *widget, dt_develop_t *dev, cairo_t *
     cairo_rectangle(cr, 0, 0, wd, ht);
     cairo_set_source_surface(cr, surface, 0, 0);
     cairo_pattern_set_filter(cairo_get_source(cr), CAIRO_FILTER_FAST);
-    cairo_fill_preserve(cr);
-    cairo_set_line_width(cr, 1.0);
-    cairo_set_source_rgb(cr, .3, .3, .3);
-    cairo_stroke(cr);
+    cairo_fill(cr);
     cairo_surface_destroy(surface);
     dt_pthread_mutex_unlock(mutex);
     image_surface_imgid = dev->image_storage.id;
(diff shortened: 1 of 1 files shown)


edgardoh:
@Nilvus , this now goes to fullscreen with F11/Escape hard-coded, no idea yet how to make it configurable. To be discussed.

2nd window has a name now, second_window.

As for the background color, I use DT_GUI_COLOR_DARKROOM_BG.

Could you please test and let me know what's missing?

Nilvus:
Great works ! I will follow when this will be merged as it seems I will have to update css to remove border. It seems that you set second_window name tag if I read correctly your code.

Nilvus:
I write my comment same time as yours. Not sure it will be possible to change border with your code. It seems that you set 'second_window' tag name on widget and not on window. It's window background which is top parent who display border. Widget a subparent of window theme and display middle grey part and image inside.

edgardoh:
Not sure I follow, I set the name just like in the main window, and yes, is
using the widget. Does this works on the main window?

Nilvus:
> Not sure I follow, I set the name just like in the main window, and yes, is using the widget. Does this works on the main window? El dom., 19 may. 2019 a las 14:54, Nilvus (<notifications@github.com>)

Yes, main_window tag is on top parent window widget. But I can't detail here, I'm only able to code on css and use Gtk inspector to see that. I don't know how to code except CSS/HTML.

Nilvus:
Forget what I said, 'second_window' is correctly tagged on window top parent. And I didn't understand that you also removed the border so I don't see anything to be customized or added, but that could be for those who wants to with this tag, so separately from main window.
Works really great and it really resolve the issue totally.

The only thing is that I find the time to display image a little longer than before (1 or 2 seconds more) and I've this on starting and when I go fullscreen and revert back to classic window.

edgardoh:
> Forget what I said, 'second_window' is correctly tagged on window top
> parent. And I didn't understand that you also removed the border so I don't
> see anything to be customized or added, but that could be for those who
> wants to with this tag, so separately from main window.
> Works really great and it really resolve the issue totally.
>
> The only thing is that I find the time to display image a little longer
> than before (1 or 2 seconds more) and I've this on starting and when I go
> fullscreen and revert back to classic window.
>
Depending on the change in size the image is reprocessed, so maybe that's
what you see. Try it with -d perf, if you see that is processing the pipe
more than it should we have a problem. For me is working the same as
before.

edgardoh:
I reuse now the global accel "toggle fullscreen" and "leave fullscreen", it works only if the 2nd window has the focus.

The image is draw on the center of a grid, so it should be possible to define some border around it with CSS, not sure if some work is needed for that.

TurboGit:
I don't think we want to make the F11/escape configurable as those keys are common to almost (if not all) desktop environment.

edgardoh:
> I don't think we want to make the F11/escape configurable as those keys
> are common to almost (if not all) desktop environment.
>
They already are for the main window, seems reasonable to use it for the
2nd one.

edgardoh:
@Nilvus , I think I found it, it was not using the preview pipe if the preview2 was not available, so it seems to take longer, could you confirm?

Nilvus:
> I don't think we want to make the F11/escape configurable as those keys are common to almost (if not all) desktop environment.

What about other OS (Mac and Windows) ? I'm not sure if they use this shortcut for full screen (even on MacOS I use for about 10 years, I don't remember that ;-) !). So maybe more secure to have configurable shortcut, and it's more consistent to have same than main window. So great for me.

@edgardoh : I have test with preview pipe adding and it's now really quick (instantaneous most of the time on my computer) so perf preview is ok now.
About line around image, I think it's not needed on this window so I will not check if it could be made or not. A pure and simple preview window as it is is great like that for me.

edgardoh:
Great!

@TurboGit , if you agree with the shortcut then this is ready for review.

TurboGit:
I kind of remember that F11 works on Windows too but my memory is very bad. And configuration is not bad anyway.

TurboGit:
Working fine on my side! Thanks.

PR #2596, opened 2019-05-19
Iop icon themes

emeikei:
Allow for theme folders where a folder in the themes folder has a .css file of the same name. eg:
`.../themes/name/name.css`
Also an option to use css for iop icons.  When disabled, old style icons will be used and will be looked for in theme folders before falling back on defaults.  Images are found in:
`.../themes/name/img/`
I've also included a sample theme 'dark'.


Nilvus:
I'm not sure to understand what you want to do with that. Could you more clear and precise about this PR ? And what the goal of this sample theme ? We already have icons themes (grey and dark ones).

emeikei:
We don't have to be stuck with one set of icons and if you want custom icons for your theme then it's much easier to just do the icons without having to do the css aswell.  It also gets a bit cluttered when each base theme has an icons version.
Your grey theme could do with custom icons to be more visible so simply putting everything in a theme folder would reduce the clutter in the themes combo box.  Those that want the icons would have them and those that don't wouldn't.
Plus, this is early days yet.  Whose to say what other files may end up in the theme folder...

TurboGit:
Well if not need custom icons just duplicate the darktable-icon.css and change the icons name and make them relative to the darktable theme directory. All this is possible already without a single line of code change.

PR #2605, opened 2019-05-20
Panasonic DC-LX100M2: add noise profiles support

d2r2:
Achieved using instruction: https://pixls.us/articles/how-to-create-camera-noise-profiles-for-darktable/.

Raw files are generated with Panasonic LX100M2 equipped with Firmware Ver.1.1. Pictures are taken with Extended ISO is on (range 100-25600 with ISO increment 1/3 EV).

Details on noise profiles generation can be found here: https://github.com/d2r2/lumix-lx100m2-noise-profiles.


TurboGit:
Looks ok to me.

PR #2606, opened 2019-05-21
[WIP] compatibility with lensfun git version

parafin:
Warnings about obsolete functions are not fixed, so it will fail to compile with -Werror without -Wno-obsolete (or whatever the actual option name).
Also don't try to use it with lensfun-0.3.95 - it won't compile and even if you make it compile it will just crash. Version from git is required.

Someone should port this to master and re-write the code to use non-obsolete API. I just decided to publish the progress I made, I'm not planning to continue this work myself anytime soon.

### DIFF
--- a/src/iop/lens.c
+++ b/src/iop/lens.c
@@ -1074,7 +1074,12 @@ void init_global(dt_iop_module_so_t *module)
   gd->kernel_lens_distort_lanczos3 = dt_opencl_create_kernel(program, "lens_distort_lanczos3");
   gd->kernel_lens_vignette = dt_opencl_create_kernel(program, "lens_vignette");
 
-  lfDatabase *dt_iop_lensfun_db = lf_db_new();
+  lfDatabase *dt_iop_lensfun_db =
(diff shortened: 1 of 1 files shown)


leclercj:
Thanks @parafin for sharing this work.

Could you elabore a bit why lensfun-0.3.95 can't work ? Is it because the C API is broken in lensfun-0.3.95 but fixed in git version ? Or a more general concern about lensfun-0.3.95?

I've the impression looking at lenfun github that lenfun is mainly developping and testing its C++ API, and that the C API is poorly maintained & tested. So one idea could be to use the lensfun C++ API in darkatble instead. Your thoughts on that ?

parafin:
1. C API is broken
2. It segfaults

First point can indeed be fixed by using C++ API, which I think dt tries to avoid if possible.
Second point I haven't investigated, maybe if we switch to non-obsolete functionality it will actually work.

parafin:
ported to master by @jakubfi in #2805, closing

PR #2612, opened 2019-05-22
preview : zoom fix

### LINKED ISSUE #2608: Zoom on lighttable preview doesn't work anymore
**Describe the bug**
When using one image preview from lighttable (W or Alt+W shortcuts), it's now impossible to zoom. When using ctrl+mouse wheel or even shortcut to go to 100% zoom (I've reset shortcut but that doesn't resolve anything), nothing happen. The image still remains fit to screen.

Zooms works great on culling mode and it seems that preview zoom disappear since some last commit about culling or maybe 2nd preview window ? It was working some days before with an older commit, but don't which one.

**To Reproduce**
1. Go to 'lighttable ; navigation file'
2. Click on 'Alt+W or W'
3. Try to zoom with ctrl+mouse wheel or shortcut to 100% zoom
4. See that nothing happen. No zoom

**Expected behavior**
A working zoom on preview view with ctrl+mouse wheel and shortcut for that function.

 - OS: Ubuntu Linux 19.04
 - darktable 2.7 last revision

AlicVB:
this fix #2608
there was an obvious bug where zoom value was not applied except for culling.
I've fixed some check too, because we can have culling + full preview

### DIFF
--- a/src/views/lighttable.c
+++ b/src/views/lighttable.c
@@ -3560,7 +3560,7 @@ static gboolean _lighttable_preview_zoom_add(dt_view_t *self, float val, double 
       nz = fmaxf(nz, 1.0f);
 
       // if full preview, we center the zoom at mouse position
-      if(lib->full_zoom != nz && get_layout() != DT_LIGHTTABLE_LAYOUT_CULLING && posx >= 0.0f && posy >= 0.0f)
+      if(lib->full_zoom != nz && lib->full_preview_id > -1 && posx >= 0.0f && posy >= 0.0f)
       {
         // we want to zoom "around" the pointer
         float dx = nz / lib->full_zoom
@@ -3574,7 +3574,7 @@ static gboolean _lighttable_preview_zoom_add(dt_view_t *self, float val, double 
       }
 
       // culling
-      if(get_layout() == DT_LIGHTTABLE_LAYOUT_CULLING)
+      if(lib->full_preview_id < 0)
       {
         // if shift+ctrl, we only change the current image
         if((state & GDK_SHIFT_MASK) == GDK_SHIFT_MASK)
@@ -3609,6 +3609,10 @@ static gboolean _lighttable_preview_zoom_add(dt_view_t *self, float val, double 
           if(lib->full_zoom + lib->fp_surf[i].zoom_delta > lib->fp_surf[i].zoom_100)
             lib->fp_surf[i].zoom_delta = lib->fp_surf[i].zoom_100 - lib->full_zoom;
         }
+      }
+      else // full preview
+      {
+        lib->full_zoom = nz;
       }
 
       // redraw


Nilvus:
Zoom works, thanks. But preview window is also really slow to render image especially when I zoom.
Culling mode with two images (for example) is quite fast and smooth to render, zoom in/out and pan images. Preview window with only one image could make few seconds to render image. And I can't pan on images even after waiting one minute. I sometimes succeed to see images pan a little (few millimetres or 1-2cm on screen). In state, it's difficult to use that and better on culling mode.

AlicVB:
hmm, it's normal that zooming in culling is faster than in full preview, as the images are way more small. But it shouldn't take too much time, except maybe on heavy development history (time to "load") with big screen resolution (panning).
I don't have that here (except on some test images with excessively big res or with extreme developments)
Can you test some things for me ?
- in culling if you zoom to maximum, do you experience same problems than with full preview at 100% ?
- with images with no dev history
- can you verify your thumbnail settings ? (memory and the 2 "save to disk" ones)
- Last thing, what is your screen resolution ? Can you test with smaller window size (always at 100%) ?

Nilvus:
For rendering, after some new tests, It's quite the same if zooming to maxium on culling and preview mode.
It's of course quite faster if there is no dev history. I have added the save to disk for 100% image (the other was already checked) and it's better (some time to render but quite faster and ok for my computer. I was avoiding 100% cache for disk space but the difference make it a good thing to use.

My screen resolution is full HD (1920x1080) on 27" screen.

So with 100% cache on prefs checked it's better. I don't have a really powerful computer, a macbook pro from 2013 with intel i5 2,4Ghz (4 cores) and an integrade Intel HD 500 graphics (so no OpenCL), but for that, if it's possible, maybe an optimization would be great.

AlicVB:
Sadly, for the loading time, I can't do anything, as it's done via dt exporting logic. In fact the loading time is aprox. the same as when you export your image in jpeg...

What I don't get in why you can't pan in the image once loaded...

Are you sure it's a recent problem ? you don't experience that before ?

Nilvus:
> What I don't get in why you can't pan in the image once loaded...
> 
> Are you sure it's a recent problem ? you don't experience that before ?

Sorry, I was not clear about pan on my last reply. With 100% image cache check, panning is ok now. So it seems that the time rendering was too long and stuck my computer (maybe) as it's not a really powerful one for rendering.

I didn't know that was the exporting logic used but that make sense. Maybe using that exporting logic is not the best way for preview window but that's not the purpose of this PR.

So, for me this PR is ok and could be merged. Panning problem with your replies and new tests is clearer here only when rendering is too long and make my computer on its limits.

AlicVB:
yes, dt is optimized to used all the power of the proc. The downside is that it doesn't let anything for doing other things...
Anyway thanks for your tests

TurboGit:
So merging, thanks to both of you for the fix and the testing!

PR #2613, opened 2019-05-22
Fix or Accept/Ignore compilation warnings

leclercj:
This is proposal to fix one and accept/ignore several compilation warnings (3 files), present in the code for several years
I'd like to have feedback from @LebedevRI on that to make sure I'm not breaking any rules here.

### DIFF
--- a/src/chart/thinplate.c
+++ b/src/chart/thinplate.c
@@ -138,7 +138,7 @@ static inline int solve(double *As, double *w, double *v, const double *b, doubl
 }
 
 #pragma GCC diagnostic push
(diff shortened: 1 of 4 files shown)


LebedevRI:
(no comment)

TurboGit:
Looks good to me! Thanks.

todd-prior:
I think it is a good idea...they are a bit disconcerting and if they indeed have no impact on the software removing the warnings is a good idea...given that the errors appear to relate to color mapping and color transfer...suspecting errors there would make you less than confident that things were working as intended...just my opinion...thanks guys

PR #2620, opened 2019-05-23
darkroom : add dynamic accels

AlicVB:
For all iop sliders which can have accels, you can now set a "dynamic" accel.

More than complex explanation, here is an example how this works : 
- you set "E" key to image operation/exposure/exposure/dynamic
- enter darkroom with an image
- you can now set the exposure value, with mouse scroll AND key "E" pressed

Current dynamic accel if any are written in the hinter message box

### DIFF
--- a/src/control/control.c
+++ b/src/control/control.c
@@ -29,6 +29,7 @@
 #include "control/conf.h"
 #include "control/control.h"
 #include "develop/develop.h"
+#include "gui/accelerators.h"
 #include "gui/draw.h"
 #include "gui/gtk.h"
 #include "views/view.h"
@@ -669,6 +670,22 @@ int dt_control_key_pressed_override(guint key, guint state)
   /* check if key accelerators are enabled*/
   if(darktable.control->key_accelerators_on != 1) return 0;
 
+  // dynamic accels
+  darktable.view_manager->current_view->dynamic_accel_current = dt_dynamic_accel_find_by_key(key, state);
+  if(darktable.view_manager->current_view->dynamic_accel_current)
+  {
+    gchar **vals = g_strsplit_set(darktable.view_manager->current_view->dynamic_accel_current->path, "/", -1);
+    gchar *txt = "";
+    if(vals[0] && vals[1] && vals[2] && vals[3])
+    {
+      txt = dt_util_dstrcat(NULL, "scroll to change <b>%s</b> of %s module", vals[3], vals[2]);
+    }
+    dt_control_hinter_message(darktable.control, txt);
+    g_free(txt);
+    g_strfreev(vals);
+    return 1;
+  }
+  
   if(key == accels->global_sideborders.accel_key && state == accels->global_sideborders.accel_mods)
   {
     /* toggle panel viewstate */
@@ -733,6 +750,10 @@ int dt_control_key_released(guint key, guint state)
   // this line is here to find the right key code on different platforms (mac).
   // printf("key code pressed: %d\n", which);
 
+  // be sure to reset dynamic accel
+  if(darktable.view_manager->current_view->dynamic_accel_current) dt_control_hinter_message(darktable.control, "");
+  darktable.view_manager->current_view->dynamic_accel_current = NULL;
+  
   int handled = 0;
   switch(key)
   {
--- a/src/control/control.c
+++ b/src/control/control.c
@@ -674,7 +674,7 @@ int dt_control_key_pressed_override(guint key, guint state)
   darktable.view_manager->current_view->dynamic_accel_current = dt_dynamic_accel_find_by_key(key, state);
   if(darktable.view_manager->current_view->dynamic_accel_current)
   {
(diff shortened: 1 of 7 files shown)


TurboGit:
Wonderful! I'll test this ASAP.

TurboGit:
This does not work on my side.

I have indeed set E to dynamic on exposure module.

When I press E and the scroll the mouse I have the exposure updated but also the zoom.

Also, at this point when I just click E (no mouse scroll) the exposure is changing.

AlicVB:
@TurboGit : 
It was not working if that was not the center image which has the focus (ex : if it was an iop) I've just fixed that.
But I can't reproduce the last point (changing exposure without mouse scroll) and I don't get how it's even possible, as I only touch slider values inside the `scrolled` function in `view/darkroom.c`... Is there any step to reproduce ?

TurboGit:
Sorry my bad, E was on lower exposure and not on dynamic. I probably clicked on the wrong line :(

TurboGit:
I'll continue my testing tomorrow.

src/control/control.c
```diff
@@ -664,6 +670,22 @@ int dt_control_key_pressed_override(guint key, guint state)
   /* check if key accelerators are enabled*/
   if(darktable.control->key_accelerators_on != 1) return 0;
 
+  // dynamic accels
+  darktable.view_manager->current_view->dynamic_accel_current = dt_dynamic_accel_find_by_key(key, state);
+  if(darktable.view_manager->current_view->dynamic_accel_current)
+  {
+    gchar **vals = g_strsplit_set(darktable.view_manager->current_view->dynamic_accel_current->path, "/", -1);
+    gchar *txt = "";
```
TurboGit: This is static string...

src/control/control.c
```diff
@@ -664,6 +670,22 @@ int dt_control_key_pressed_override(guint key, guint state)
...
+    if(vals[0] && vals[1] && vals[2] && vals[3])
+    {
+      txt = dt_util_dstrcat(NULL, "scroll to change <b>%s</b> of %s module", vals[3], vals[2]);
+    }
+    dt_control_hinter_message(darktable.control, txt);
+    g_free(txt);
```
TurboGit: and you free it there.

It may explain the crash I have with this PR each time I exit the preference dialog.

src/gui/accelerators.c
```diff
@@ -967,6 +1016,51 @@ void dt_accel_rename_lua(const gchar *path, const gchar *new_path)
...
 
+static gint _dynamic_accel_find(gconstpointer a, gconstpointer b)
+{
+  dt_accel_dynamic_t *da = (dt_accel_dynamic_t *)a;
+  GtkAccelKey *ak = (GtkAccelKey *)b;
+  if(da->accel_key.accel_key == ak->accel_key && da->accel_key.accel_mods == ak->accel_mods) return 0;
+  // not the right one
+  return 1;
+}
+dt_accel_dynamic_t *dt_dynamic_accel_find_by_key(guint accel_key, GdkModifierType mods)
```
TurboGit: empty line before

src/gui/accelerators.c
```diff
@@ -967,6 +1016,51 @@ void dt_accel_rename_lua(const gchar *path, const gchar *new_path)
...
+{
+  GtkAccelKey ak = { 0 };
+  ak.accel_key = accel_key;
+  ak.accel_mods = mods;
+  GSList *da = g_slist_find_custom(darktable.control->dynamic_accelerator_valid, &ak, _dynamic_accel_find);
+  if(da && da->data) return (dt_accel_dynamic_t *)da->data;
+  return NULL;
+}
+void dt_dynamic_accel_get_valid_list()
```
TurboGit: likewise

src/gui/accelerators.c
TurboGit: empty line before

src/gui/accelerators.c
TurboGit: empty line before

AlicVB:
@TurboGit : I've fixed the crash (it was a reinitialization of glist problem) and all your other comments

TurboGit:
Ok, looks good now! Thanks!

aurelienpierre:
That's awesome !!! Thanks @AlicVB

dterrahe:
@AlicVB I'm looking at the accelerator code at the moment and was wondering why it was necessary to special case the shortcuts for dynamic accels. I.e. what functionality was impossible to implement using normal accels with closures to enable the dynamic accel scroll state and then in dt_control_key_released disable it again?

AlicVB:
@dterrahe : I've to admit that I don't remember... If my memory doesn't fail, I remember to have test that at first, but that didn't work (scroll event not taken in account ???)

dterrahe:
@AlicVB thanks for your reply. I've now simplified to code to switch into dynamic mode using normal means and then the scrolling is done as before. See #6584. Works for me, so hopefully there isn't anything I overlooked.

PR #2626, opened 2019-05-25
themes : change bauhaus themes on the fly too

AlicVB:
no need to restart dt anymore to see the effect of a theme change on bauhaus controls, especially colors and fonts.
For widgets sizes (margins, ...) this don't works (as before, and as for the other "classic" widgets)

### DIFF
--- a/src/bauhaus/bauhaus.c
+++ b/src/bauhaus/bauhaus.c
@@ -472,14 +472,8 @@ static void dt_bh_class_init(DtBauhausWidgetClass *class)
   // widget_class->draw = dt_bauhaus_draw;
 }
 
-void dt_bauhaus_init()
+void dt_bauhaus_load_theme()
 {
-  darktable.bauhaus = (dt_bauhaus_t *)calloc(1, sizeof(dt_bauhaus_t));
-  darktable.bauhaus->keys_cnt = 0;
-  darktable.bauhaus->current = NULL;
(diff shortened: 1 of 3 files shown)


Nilvus:
That was fast and not a so big rework. Of course, that doesn't work for all but colors and fonts are the main part.
What about histogram ?

AlicVB:
histogram colors are updated, like the other colors (except if I miss something).
In fact, even if I know you're not a "coder", you can see all the updated colors, starting from this line (click on the link) : https://github.com/darktable-org/darktable/blob/c3c38fb8b0234067d724baf8d9bfe5504a3c1fdf/src/bauhaus/bauhaus.c#L492

Nilvus:
> histogram colors are updated, like the other colors (except if I miss something).
> In fact, even if I know you're not a "coder", you can see all the updated colors, starting from this line (click on the link) :
> 
> https://github.com/darktable-org/darktable/blob/c3c38fb8b0234067d724baf8d9bfe5504a3c1fdf/src/bauhaus/bauhaus.c#L492

Ok, I will test your PR, that will be better ;-) !

Nilvus:
Great work, that rocks ! Good to see these last things change on fly and without restart. About margins, that's not so important as colors and fonts as, actually, all margins are the same between all themes.
For me, it's ready to merge ! Thanks again for your work.

TurboGit:
Nice, thanks!

PR #2627, opened 2019-05-26
rgb curve - added preserve colors norms

phweyland:
Added the following norms (preserve colors parameter) to rgb curve:
- max(r,g,b)
- average(r,g,b)
- norm(r,g,b) = (r^2+g^2+b^2)^(1/2)
- power norm =  (r^3+g^3+b^3)/ (r^2+g^2+b^2)

### DIFF
--- a/data/kernels/rgbcurve.cl
+++ b/data/kernels/rgbcurve.cl
@@ -63,11 +63,15 @@ rgbcurve(read_only image2d_t in, write_only image2d_t out, const int width, cons
       {
         lum = (pixel.x + pixel.y + pixel.z) / 3.0f;
       }
-      else if (preserve_colors == 4) // DT_RGBCURVE_PRESERVE_LNORM
+      else if (preserve_colors == 4) // DT_RGBCURVE_PRESERVE_LSUM
+      {
+        lum = pixel.x + pixel.y + pixel.z;
+      }
+      else if (preserve_colors == 5) // DT_RGBCURVE_PRESERVE_LNORM
       {
         lum = native_powr(pixel.x * pixel.x + pixel.y * pixel.y + pixel.z * pixel.z, 0.5f);
       }
-      else if (preserve_colors == 5) // DT_RGBCURVE_PRESERVE_LBP
+      else if (preserve_colors == 6) // DT_RGBCURVE_PRESERVE_LBP
       {
         float R, G, B;
         R = pixel.x * pixel.x;
--- a/data/kernels/rgbcurve.cl
+++ b/data/kernels/rgbcurve.cl
@@ -41,19 +41,7 @@ rgbcurve(read_only image2d_t in, write_only image2d_t out, const int width, cons
   }
   else if(autoscale == 0) // DT_S_SCALE_AUTOMATIC_RGB
   {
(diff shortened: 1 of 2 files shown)


TurboGit:
Why not! Just a question, would you be able to add the corresponding documentation for those modes? The changes are subtle but visible, now we need to find English word to describe them, hence my question.

phweyland:
> would you be able to add the corresponding documentation for those modes? The changes are subtle but visible, now we need to find English word to describe them, hence my question.

The norm is a way to mitigate the distortion of rgb values induced by the curve. 

For me that's (necessary) cooking. I have no scientific explanation for any of them. I have not seen any method working perfectly in every case neither.
But as you say every one produces subtle differences, which gives the user as many options to get the more appropriate color/chrominance.

@aurelienpierre has given most of them here: https://github.com/darktable-org/darktable/pull/2017#issuecomment-465473804
His norm L1 changes significantly the image luminance so I didn't include it at first. 
Beside those I've put the average rgb and the basic power which provides also intermediary effects compared to the original luminance option.

Naming can be changed, L1 can be added,... that's open.

Where would you like to add the description ? In dt itself  (combo box) ?
In darktable manual (tone modules) ? If so, I haven't seen any description for rgb curve. And I haven't found the corresponding git repository (darktable.gitlab.io ? not part of dtorg if I'm not mistaken).

TurboGit:
> In darktable manual (tone modules) ? If so, I haven't seen any description for rgb curve.

In the manual indeed. This will be needed for the next release. I'll send messages to the devs for them to describe their work to be included into the manual.

All in all, I'll merge this, as I said this works. Thanks.

TurboGit:
@aurelienpierre : we had only none or luminance before all other options have been added after the current mode. so to me this is not breaking history.

aurelienpierre:
Ok, my bad. I saw the commit change in my mail box and it wasn't clear if was a change over the core or over a PR.

aurelienpierre:
By the way, @phweyland, did you notice better results than average with the power norm ? I have it in #1904 and it seems to be the best to preserve local contrast. That mindblows me since there is no physical meaning in this norm. Just wondering…

phweyland:
> By the way, @phweyland, did you notice better results than average with the power norm ? I have it in #1904 and it seems to be the best to preserve local contrast. That mindblows me since there is no physical meaning in this norm. Just wondering…

So far I was just liking the result, without knowing why. Now you say it I would agree that local contrast is not too bad ...
BTW, if you are able to explain the effect of each of the norms you have suggested I would appreciate to know ... :-)

aurelienpierre:
> BTW, if you are able to explain the effect of each of the norms you have suggested I would appreciate to know ... :-)

max(RGB) is equivalent to HSV value and is very good at preventing individual channel clipping (hence hue shift in highlights).

average(RGB) is an efficient way to get the closest grey from the RGB vector. It's roughly similar to HSL lightness (L = average(min(RGB) , max(RGB))

the euclidian norm is the mathematically accurate way to get the closest grey, but not in a very meaningfull way, perceptually speaking.

the power norm is black magic. I don't know why it works, but it does.

phweyland:
Thanks. Makes sense.
I'll keep that to document the module when it's time.

PR #2629, opened 2019-05-26
colorzones: do not move nodes on x-axis

edgardoh:
On colorzones, do not allow to move the first or last node on the x-axis, and when resetting the values reset both nodes if on hue mode.


TurboGit:
Looks good, thanks.

PR #2635, opened 2019-05-28
exposure: display a warning in GUI when there are negative RGB values

aurelienpierre:
Check for negative values after exposure module in the preview pipe and display a warning in control log.

Since checks are expensive in OpenCL, we introduce an additional kernel with the negative checking built-in that will be used only for display pipelines.

Negative values are a burden when using log or square root based algorithm since darktable doesn't clip them. Many users have taken the wrong habit to adjust the contrast using the black level compensation, which is too soon in the pipe and will harm other modules output.

### DIFF
--- a/data/kernels/basic.cl
+++ b/data/kernels/basic.cl
@@ -161,6 +161,22 @@ exposure (read_only image2d_t in, write_only image2d_t out, const int width, con
   write_imagef (out, (int2)(x, y), pixel);
 }
 
+kernel void
+exposure_check_negatives (read_only image2d_t in, write_only image2d_t out, const int width, const int height, const float black, const float scale,
+          global int *negatives)
+{
+  const int x = get_global_id(0);
+  const int y = get_global_id(1);
+
+  if(x >= width || y >= height) return;
+  float4 pixel = read_imagef(in, sampleri, (int2)(x, y));
+  pixel.xyz = (pixel.xyz - black)*scale;
+  write_imagef (out, (int2)(x, y), pixel);
+
+  if(any(isless(pixel.xyz, (float3)0.0f)))
+    *negatives = 1;
+}
+
 /* kernel for the highlights plugin. */
 kernel void
 highlights_4f_clip (read_only image2d_t in, write_only image2d_t out, const int width, const int height,
--- a/src/iop/exposure.c
+++ b/src/iop/exposure.c
@@ -99,6 +99,7 @@ typedef struct dt_iop_exposure_data_t
 typedef struct dt_iop_exposure_global_data_t
 {
   int kernel_exposure;
+  int kernel_exposure_check_negatives;
 } dt_iop_exposure_global_data_t;
 
 
@@ -363,6 +364,31 @@ static void process_common_setup(dt_iop_module_t *self, dt_dev_pixelpipe_iop_t *
   d->scale = 1.0 / (white - d->black);
 }
 
+static void check_negatives(struct dt_iop_module_t *const self, dt_dev_pixelpipe_iop_t *const piece,
+                            const dt_iop_roi_t *const roi_out, void *const out)
+{
+  // Check for negative values in pixels RGB channels since they are not clipped and super
+
+  if(self->dev->gui_attached && piece->pipe->type == DT_DEV_PIXELPIPE_PREVIEW)
+  {
+    gint negatives = FALSE;
+
+  // Note : don't enable SIMD/SSE for this, it's twice as slow
+#ifdef _OPENMP
+#pragma omp parallel for default(none) schedule(static) shared(negatives)
+#endif
+    for(size_t k = 0; k < (size_t)piece->colors * roi_out->width * roi_out->height; k++)
+    {
+      if(!negatives)
+        if(((float *)out)[k] < 0.0f)
+          negatives = TRUE;
+    }
+
+    if(negatives)
+      dt_control_log(_("exposure: negative RGB values, check black level"));
+  }
+}
+
 #ifdef HAVE_OPENCL
 int process_cl(struct dt_iop_module_t *self, dt_dev_pixelpipe_iop_t *piece, cl_mem dev_in, cl_mem dev_out,
                const dt_iop_roi_t *const roi_in, const dt_iop_roi_t *const roi_out)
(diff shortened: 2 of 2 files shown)


edgardoh:
I don't think this is a good idea, the information is not relevant, it affects performance and having a distracting message each time the exposure is processed is not cool.

TurboGit:
> I don't think this is a good idea, the information is not relevant, it affects performance and having a distracting message each time the exposure is processed is not cool.

I understand, that was my first impression, but fact is that black level should (almost) never be changed.

TurboGit:
For the perf issue maybe we should do the **check only when the black level is changed**?

edgardoh:
Maybe, but this is not the right way to address that. There are a lot of
advanced setting that should be used only "if you know what you're doing".

aurelienpierre:
I have timed both, the perf penalty is 10% to 15% only on the thumbnail computation, which is an absolute difference of 2 to 5 ms.

aurelienpierre:
> **check only when the black level is changed**

For now, the computation is done only if the pixel pipe is attached to a GUI and in a preview pipe (thumbnail), meaning the main display and the export pipe are not affected.

edgardoh:
At the very least there should be a setting, disabled by default, to
activate this.

But still it doesn't feel right, if some modules have a problem with
negative values is up to that modules to handle it. If the problem is that
some users may use a setting the wrong way a warning on the user manual
and/or the tooltip is more appropriate.

aurelienpierre:
It's not an issue with how modules handle negatives values, but rather a challenge to find out which module is responsible for black blobs. Some rawspeed black levels are wrong so it's best to diagnose it soon enough and correct them early in the pipe, I think. But that could be an option.

edgardoh:
But this way you are not diagnosing that, you just say that there are
negative values, and that really says nothing because it can be only one
pixel, and in the preview pipe.

If the user has some black blobs (or any other edit problem) a tutorial
seems the way to go.

parafin:
Is the issue here negative RGB values or the fact that user touches black level slider? Most DT modules don't clip, that's unbounded pipe for you. Even if we would want to ban negative values, why single out exposure iop? It will make more sense to add a mode like -d nan we have for NaNs.

If you just want to display a warning when black level is touched at all, then it makes more sense to do so near the slider IMHO. But if we go that way, we would need a lot of warnings added to DT. I don't really think it's a good idea.

What is the actual problem we're trying to solve? Some iop can't handle negative values? Why not clip them on input there then?

aurelienpierre:
Filmic, color balance and input profile correction have to clip below zero because they do operations like log or square roots, which are undefined for negative inputs. Also, negative values would encode negative energy which is an inconsistency.

Although modules handle negative values and avoid NaN with checks, they do so by clipping the affected channel, without taking care of the actual pixel color, meaning they don't correct the other channels accordingly, which can lead to strange artifacts like black pixels in dark saturated reds gradients.

So the problem is not NaN or divisions by zeros, but just letting users know that they made something stupid, or rawspeed black levels are a little off, because the negatives handling is far from perfect in the following modules, so you would better take care of that sooner than later.

Otherwise, we could have an option to simply clip negatives in the exposure module.

parafin:
Clipping negative values in exposure iop makes sense, that's what rawspeed black level does, does it not?
Adding a warning like this PR does IMHO does not make sense.

aurelienpierre:
Adding a warning does make sense to me, since the automatic negative values handling (in modules) does not care about color preservation, there is a need for a manual "clever" correction (just slide the black level), thus there is a need to know if you have crossed the red line or not.

The thing is, even with clipping in exposure, it will still not be color-safe. Or we (I) have to develop some complicated way to correct the 2 unclipped channels of every pixel having one clipped, when all you have to do is lower the black level a bit.

edgardoh:
As I already say negative values are not fixed, so if you don't have any on
the preview pipe it does not mean that won't be any on export, and by
saying that there's negative values you will raise false positives most of
the time, so the message has really no meaning, or worst, will make think
that there's something wrong where there isn't.

If there are some process that is returning the wrong values (negative or
positive) then that process should be fixed.
If some modules don't like negative values, is up to those modules to
handle it.
If the user is not getting the expected result with the edit, is not up to
darktable to fix/diagnose/prevent it.

parafin:
I would say getting weird colors if you went too far with black level is an expected result, no need to fix it. So I agree with @edgardoh.
If you want to give a guide to the user on how far he can go, maybe you can use histogram to render some line on the slider.

aurelienpierre:
> As I already say negative values are not fixed, so if you don't have any on the preview pipe it does not mean that won't be any on export, and by saying that there's negative values you will raise false positives most of the time, so the message has really no meaning, or worst, will make think that there's something wrong where there isn't.

That's a valid point. But there won't be false positive (since negative are statistical outliers, interpolation will make them disappear amoung their positive neighbours), the worst that could happen is actually false negative.

> If there are some process that is returning the wrong values (negative or positive) then that process should be fixed. If some modules don't like negative values, is up to those modules to handle it.

There are no wrong values, that's the problem. The camera sensor just throws code values, it's up to the software **and** to the user to decode these values and give them meaning. The software cannot know whether negative values are noise or wrong black levels in the lib or something else, so you can't just program something and expect it will work all the time. The best and simple correction is to warn the user that something is off, hoping he will find the problem that the soft couldn't handle. Every automatic correction will be more computationnaly expensive than a user action and will fail at least 20 % of the time. 

Negative values are something that shouldn't happen. If they do, something bad happened in the pipe and the user, I think, needs to be warned. It's not a normal situation, and no matter how the software should theoritically handle it, it didn't so… fallback to human correction.

parafin:
1. Can negative values happen in the pipe before exposure iop and how if so?
2. Can negative values happen in the pipe after exposure iop and how if so?
3. Shouldn't we check for negative values after each iop then?
4. Can negative values have meaning and if so why check for them?

aurelienpierre:
> Can negative values happen in the pipe before exposure iop and how if so?

Yes, if:

- rawspeed coefficients are off for that sensor, which happens sometimes
- the pictures is noisy and not properly denoised

> Can negative values happen in the pipe after exposure iop and how if so?

Yes, if the black level setting in exposure was ill-set.

> Shouldn't we check for negative values after each iop then?

No, because only the exposure and color balance modules can actually create negative values that weren't there at their input already. (To create negatives, you need to apply a transfer function with a negative offset in it, very few modules do that).

> Can negative values have meaning and if so why check for them?

Inside a black hole, yes. Otherwise, no. There is no negative light on Earth.

parafin:
So negative values can happen in 3 places, right?
1.  When rawspeed subtracts black level - shouldn't output be clamped to zero? As you say there are no negative light. Some denoising algorithms may benefit from these negative values, but they should be placed before black level subtraction then, I guess.
2.  In exposure module - this is user error and it may be intentional, so I can only suggest to give a guide (see my previous comments) and probably clamp negative numbers again.
3.  In color balance module - probably the same case as with exposure.

aurelienpierre:
>     1. When rawspeed subtracts black level - shouldn't output be clamped to zero? As you say there are no negative light. Some denoising algorithms may benefit from these negative values, but they should be placed before black level subtraction then, I guess.

I also said RGB vectors encode light, they are the map here, not the territory. So if the map shows a physically inconsistent data, the map is more likely to be wrong that the actual territory. And clamping individual channels is the worst way to do it.

>     2\. I can only suggest to give a guide (see my previous comments) and probably clamp negative numbers again.

The guide will not be enough to know whether or not you are sitting on the right side of the red line.

parafin:
I think these are 2 separate issues that need independent approach. One is how to detect and report when we have wrong black level in rawspeed for this camera, another is advice user when he mis-use exposure iop. I'm not sure about the first (rawspeed black level) - what user is supposed to do with this information? Tune black level manually? Report it as a bug?
For the second one (exposure black level) - what is the use-case here to touch black level slider at all?  Wouldn't it screw up the colors even if no negative values are produced?

AlicVB:
My 2 cents : if I understand well, it's too difficult to automatically correct/clip/... the negative values without screwing the colors, right ? In this case, as @aurelienpierre says, users should have a way to know when they cross the "red line". 
I really dislike the message implemented in this PR as it's too intrusive imho, and as @TurboGit says not up to date...
I would prefer some meaningful message inside the iop, why not under the black level slider, maybe with some danger icon if you think it's needed.
We can also go the route to add another tool like over-exposed and co (not sure if it's possible or relevant, as @aurelienpierre speak about some statistical detection...) or at least some meaningful icon in this area (with a nice explanation in the tooltip)

aurelienpierre:
> if I understand well, it's too difficult to automatically correct/clip/... the negative values without screwing the colors, right ?

Yes. While the solution to fix most problem would be to simply lower the black level compensation in exposure, to fix everything at once.



> I really dislike the message implemented in this PR as it's too intrusive imho, and as @TurboGit says not up to date...

I can insert the value of the faulty setting in it, or I could put it in the module widget, but I think it is important enough to bother the user until the setting is fixed.

To explain better, here is an example of an RGB vector {-5, 20, 50} % displaying the result of the clipping strategy and the offsetting one (the black level compensation is an offset) :
clipping-offsetting
So you see it's not the same color anymore, the brightness has shifted (obviously), but the hue goes from a medium blue to marine blue. Imagine that pixel trapped in the middle of a clean medium blue gradient… Clipping it would create noise.

> what user is supposed to do with this information? Tune black level manually? Report it as a bug?

Tune the black level manually is the fastest way to get work done. Not everyone can wait for the next release to deliver a proper image.

parafin:
Black level can be tuned in raw black white point iop, so I'll repeat my question - what is the use-case for black point slider in exposure iop?

aurelienpierre:
The raw black point is actually 3 integers black points, so 3 values to tweak instead of one. Then come other modules like temperature which could create negatives values (not sure, but since it shifts the white point, it seems theoritically possible). What is sure is there is no raw black point module if you are working on raster images, for example scanned films inverted in darktable (another chance to get negatives), so the exposure modules comes at the end of the danger zone and fixes everything before going in the input color profile.

parafin:
OK, so use-case for black level in exposure iop is to correct wrong black level. Your idea is to report to the user when black level is set too high (black level here being both rawspeed black level and exposure black level applied together). I agree with @AlicVB that icon or a message near the slider is much better than currently implemented GUI warning. Aside from visuals I see 3 problems:
1.  Exposure iop is off by default and not always used. So user won't be able to notice wrong rawspeed black level in general case.
2.  Check for negative values can only detect black level set too high, so too low black level is left undetected. I guess we are ignoring this issue in this discussion for now, since it causes less trouble.
3.  Are we sure that camera can't produce values lower than black level? That's not my understanding, I think it's expected that values below black level can happen and should be clipped to zero.

aurelienpierre:
> 1\. Exposure iop is off by default and not always used. So user won't be able to notice wrong rawspeed black level in general case.

Yes but exposure is one of the most used modules in general, so there is a high probability it will be used.

>  2\. Check for negative values can only detect black level set too high, so too low black level is left undetected. I guess we are ignoring this issue in this discussion for now, since it causes less trouble.

There is no "too low" black level, from a signal processing point of view. The worst that can happen is blacks get faded, which is easy to see (and to over-correct…), but not every picture needs true blacks, and there is no mathematical issue there with log or gamma.

> 3\. Are we sure that camera can't produce values lower than black level? That's not my understanding, I think it's expected that values below black level can happen and should be clipped to zero.

Cameras produce 12 or 14 bits unsigned integers. The black level is a pure convention based on the SNR of the sensor and other electrical artifacts, supposed to be the threshold between noise (or base voltage) and actual data. First of all, each sensor has its own caracteristics, even in the same batch, so every generic black level based on average SNR is just a guideline. Especially since the exposure comes after 2 denoising and the demosaicing modules, if denoised pixels still get negative values, it means their neighbours are negative too, which is not an isolated accident anymore. Then, the problem is the black level does not affect just dark pixels since it works on individual channels. Clipping recklessly one channel in mid-tones can make a big difference, especially around blue and purple (see the example above). Anyway, the point is there is no general answer and that's why the user needs to check himself and decide what to do (clip or lift).

parafin:
You are assuming user will use denoising (and exposure). I don't think everyone always use denoising (I certainly do not for low ISO shots). And without denoising you are saying having values below black level is quite normal due to differences between sensors and noise. So wouldn't it trigger this warning basically always?

aurelienpierre:
I'm assuming users work properly and know what they are doing, meaning they denoise when necessary.

So, assuming that, what comes at the input of the exposure module is assumed to be clean of noise (either low ISO or denoised high ISO), meaning if negatives occur, the culprit is likely to be the processing before the exposure module and not the signal itself.

Negative values are not normal anyway, they are not supposed to happen anywhere in the pipe, and if the warning is triggered all the time, so be it, because it means bad news for our color management and we need to know it sooner than later.

TurboGit:
I also think that having a "warning" on the module itself is better. A red icon with a tooltip explaining the problem?

AxelG-DE:
Hey guys,
Is it helpful for you, if I explain, what I do in real live with this blackpoint slider, as I am actually using it? 

I understood already, why it is disbeneficial and would find an alternative in the future (e.g. tone curve). So no need for replies like "_in that case you shouldn', or you should..._"

Many tuts tell you, exp. slider to be used in a way, to avoid clipping on the right (clear so far) and also you could use black point to reach out for the limit on the left, if not there already. 

Now I often draw a mask around the eyes (sometimes for humans but more often on animals) and lift the exposure a bit and give it a little lift on black. Try to do this almost unnoticable but if provides a nice gloss.
With
2018-05-15_131056_AG7_9521_01
Without
2018-05-15_131056_AG7_9521_02

If I am misleading the topic, just ignore my post :smile:

TurboGit:
I'm closing this as it lost traction. Feel free to reopen if needed.

Issue #2506, opened 2019-05-02
no histogram in tone curve until reset


s5k6:
**Describe the bug**
When editing the tone curve, the histogram that was usi=ually visible as a backdrop is missing.

![screenshot TQdpozyZxZ](https://user-images.githubusercontent.com/16781735/57092244-3d2df280-6d0b-11e9-8637-48a26ff18b27.png)

The histogram will show up when resetting the tone curve by clicking on it's ⏼, but not when resetting it with a double click on the curve.  Changing the color space, or choosing a preset has no such effect.

After going to lighttable and back to darkroom, the histogram is missing again.

**To Reproduce**
1. open image in darkroom
2. open "tone curve" module

**Expected behavior**
Always show histogram in tone curve module, or at least while it is switched on.

**Platform (please complete the following information):**

  * current ArchLinux
  * no desktop environment (plain X11 and FVWM)
  * darktable 2.7.0+1087~g7fe35d67d

Nilvus:
It also show up if you select one option on drop-down just below the graph.

Nilvus:
Have you tested new rgb curve tone module ? Maybe it will be better and doesn't have this behavior.

s5k6:
Nilvus (2019-May-19, excerpt):
> Have you tested new rgb curve tone module ? Maybe it will be better
> and doesn't have this behavior.

Yes, "rgb curve" does not show this problem.  What's the difference
between the two?  I could not find docs for "rgb curve", I did not
even know it existed.


-- 
http://stefan-klinger.de                                        o/X
I prefer receiving plain text messages, not exceeding 32kB.     /\/
                                                                  \

Nilvus:
> Nilvus (2019-May-19, excerpt):
> Have you tested new rgb curve tone module ? Maybe it will be better and doesn't have this behavior.
> Yes, "rgb curve" does not show this problem. What's the difference between the two? I could not find docs for "rgb curve", I did not even know it existed.

rgb curve was merged on the second part of march and it's a tone curve in rgb mode. So maybe use it instead of tone curve. On the PR was discussed the fact to have a rgb curve with actual tone curve or having an improving tone curve (on Lab mode) with a rgb mode added. Merging new rgb curve was done but I don't know if tone curve will be update to also have a rgb mode (it doesn't seems necessary now). The PR to add this is always on but not updated since a time (see on bottom page of pull requests).

So I think you could :
[...]

Issue #2402, opened 2019-04-13
UI 2.7: low contrast in histogram of 'tone curve' module


pphotography:
**Describe the bug**
In the top-right corner of the histogram in the new 'tone curve' module from version 2.7 the contrast between the histogram data and the background is getting very low. It becomes hard to distinguish from the background.
Foreground is around #969696 and background around #929292 which makes a contrast ratio of 1.05:1
See area in red square below

**Expected behavior**
Contrast should be ok if the background gradient's brightness goes up to the grey that is in the green square below

**Screenshots**
![Contrast_UI](https://user-images.githubusercontent.com/49001045/56086201-e478e180-5e51-11e9-82f0-62855047bb7b.jpg)

**Platform (please complete the following information):**
 - Version2.7.xx

Nilvus:
After some tests and using Gtk inspector, that could not be improved by CSS. Nodes and item around widget tone curve could be edited only, not the graph and histogram on this module.

But now, you could use new rgb tone curve that doesn't have this behavior.

Nilvus:
To be closed as solved by PR #2638 (@TurboGit)

TurboGit:
Closing.

PR #2636, opened 2019-05-28
tonecurve - fixed error init of scale in gui

phweyland:
tonecurve histogram did not appear at init time because scale was not set.


Nilvus:
So it resolve this issue #2506 

@phweyland : what do you think of this issue (not possible to change scale background with css) : #2402 ? Is it possible to also improve that with that PR (or another) ?

TurboGit:
Looks good, thanks.

phweyland:
> what do you think of this issue (not possible to change scale background with css) : #2402 ? Is it possible to also improve that with that PR (or another) ?

It is easy to improve the readibility. That's also a change in PR #2638.
But to make it compatible css that's another story... Is rgb curve compatible css ?

Nilvus:
no, rgb curve is not css compatible. But here just a better readability would be ok. About css, if something made, that would be the idea to have same thing like the histogram, so here especially the background.

PR #2638, opened 2019-05-28
tonecurve - cleanup histogram with log scale

phweyland:
Work in progress

**Problem**: The histogram is not displayed as soon as scale is set as not linear.
The scale can be set for both x & y, x or y.
image

There is a basic conflict with the pipeline histogram, which has also an option linear/log (for y).

I suggest to remove the scale options y and xy (log-log(xy) and semi-log(y)) to keep only linear and log x in tonecurve.
If the user wants to have also log for y, he selects log on the global histogram. In that case log y is not scalable but I don't think that's an issue.

What do you think ?
I ask the question before going further because there is a lot of cleaning (including with tonecurve color picker)

For the time being some screenshots for the current changes:

tc linear + histo linear
image

tc linear + histo log(y)
image

tc log(x) + histo linear
image

tc log(x) + histo log(y)
image

### DIFF
--- a/src/gui/draw.h
+++ b/src/gui/draw.h
@@ -370,17 +370,6 @@ static inline void dt_draw_histogram_8_log_base(cairo_t *cr, const uint32_t *his
     dt_draw_histogram_8_logxliny(cr, hist, channels, channel, base_log);
   else  // log y
     dt_draw_histogram_8_logxlogy(cr, hist, channels, channel, base_log);
-/*
-  cairo_move_to(cr, 0, 0);
-  for(int k = 0; k < 256; k++)
-  {
-    const float x = (float)k / 255.0f;
-    cairo_line_to(cr, logf(x * (base_log - 1.0f) + 1.0f) / logf(base_log), logf(1.0 + hist[channels * k + channel]));
-  }
-
-  cairo_line_to(cr, 255, 0);
-  cairo_close_path(cr);
-  cairo_fill(cr);*/
 }
 
 // linear x
--- a/src/gui/draw.h
+++ b/src/gui/draw.h
@@ -296,7 +296,8 @@ static inline int dt_draw_curve_add_point(dt_draw_curve_t *c, const float x, con
   return 0;
 }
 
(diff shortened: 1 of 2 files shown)


Nilvus:
Just here to associate the issue #2402 on what have been discussed in PR #2636

phweyland:
Solve:
- Histogram with log
- Visibility of histogram issue #2402 

Please review.

Note about the following comments I've found in tonecurve.c.
I'm not sure we have to let them in the code.
Who are the targeted people ? Who should be hurt ?
Is there any intention to turn this module ugly ?

```
  // since Cairo paints with sRGB at gamma 2.4, linear gradients are not linear but garbage and, at 50%,
  // we dont see the neutral grey we would expect in the middle of a linear gradient between
  // 2 complimentary colors. So we add it artificially, but that will break the smoothness
  // of the transition. Maybe this will help people understand how broken are Lab and non-linear
  // spaces for editing, so let it be ugly to teach them a lesson.

```

phweyland:
Option. There is some more cleaning to do to remove all reference to `semilog` parameter which is now useless. Let me know if you want I do it.

TurboGit:
No, no intention to have thi ugly. Please remove comment which is not appropriate here. I would just write:

```
  // since Cairo paints with sRGB at gamma 2.4, linear gradients are not linear and, at 50%,
  // we dont see the neutral grey we would expect in the middle of a linear gradient between
  // 2 complimentary colors. So we add it artificially, but that will break the smoothness
  // of the transition.
```

phweyland:
Just to know. Is there already a module ready for css which could taken as an example ?

Nilvus:
> Just to know. Is there already a module ready for css which could be taken as an example ?

If it's about trying to make tone curve more css editable, so you talk about histogram colour and background tone curve I guess. I don't see any module to compare with but the histogram part on top right of darkroom is probably the best part to see. On what I see in GTK inspector, the square where curve and histogram are is a one widget, on tone curve or rgb curve, even on base curve for example.

But on what I see, I'm not sure it's so important to bring that on css by now as the contrast is great now.

phweyland:
I would let this as is, except if there is some request to go further.

TurboGit:
Thanks! Looks good.

PR #2643, opened 2019-05-30
Tonecurve - added norms to tc RGB

phweyland:
Added norms to tonecurve RGB - linked channels (as for rgb curve module).

These norms are applicable on diverse modules (curves, basic adjustments, maybe filmic,...).
The corresponding routine could be shared, both for process() and open_cl. If agreed, where should we put them ?

@edgardoh. The luminance norm has simple part (dt_camera_rgb_luminance()) but another one requiring work_profile parameters (dt_ioppr_get_rgb_matrix_luminance()). In tonecurve I've bypassed this last one for the time being. Here are my questions on this:
- How important is the impact of the rgb profile on the luminance calculation for norm ?
- If important, what are your advices if we want to have a general norm routine ?


edgardoh:
The luminance depends on the rgb profile, so the rgb profile is a must. In case of native rgb modules that is the work profile, for the tone curve is prophotoRGB hard-coded.
You can add the work profile as a parameter in your new routine and create an in-memory prophotoRGB for the tone curve.

phweyland:
I've created a new color profile prophoto rgb in colorspaces (to map the hardcoded one in tonecurve).
Created a routine dt_rgb_norm (process & process cl).
Updated tonecurve accordingly.

If it works for you, I can update rgbcurve in the same way.

edgardoh:
Sure, and if is not too much to ask maybe also on the basic adjustments, so
we have the same norms everywhere.

phweyland:
I've a travis issue on compilation of rgb_norms.cl. But this file is only to be included in basic.cl, rgbcurve.cl, ... and compiled as a standalone cl file. Should I rename it as rgb_norms.h ?


> also on the basic adjustments

I'll do.

phweyland:
> basic adjustments

Here you have 2 luminance usages, one for contrast one for saturation, the second being not dependent on "preserve colors" flag. I let the second one as it is, except you tell me the opposite.

edgardoh:
I'm no expert on opencl but I think that including a .cl is ok, based on
the error it seems that you are missing some include:

/build/darktable/data/kernels/rgb_norms.cl:32:16: error: unknown type name
'dt_colorspaces_iccprofile_info_cl_t'
  global const dt_colorspaces_iccprofile_info_cl_t *profile_info, read_only
image2d_t lut)
               ^
/build/darktable/data/kernels/rgb_norms.cl:36:34: warning: implicit
declaration of function 'dt_camera_rgb_luminance' is invalid in OpenCL
    return (work_profile == 0) ? dt_camera_rgb_luminance(in):
get_rgb_matrix_luminance(in, profile_info, lut);
                                 ^
/build/darktable/data/kernels/rgb_norms.cl:36:63: warning: implicit
declaration of function 'get_rgb_matrix_luminance' is invalid in OpenCL
    return (work_profile == 0) ? dt_camera_rgb_luminance(in):
get_rgb_matrix_luminance(in, profile_info, lut);

edgardoh:
That's ok, the norm is used only for the contrast.

phweyland:
> based on the error it seems that you are missing some include:

This is true if rgb_norms.cl (or rgb_norms.h) are considered as standalone.
This is not true as included file:
```
#include "colorspace.cl"
#include "color_conversion.cl"
#include "rgb_norms.h"
```
because these definitions are given in the previous includes.

So, maybe the question is, how to make sure that the file is copied in the installation kernel folder, while it is not a program in itself (number in programs.conf).

phweyland:
I think that's ok like that. Travis is happy. And I've checked that removing the cached compiled cl file (basics.cl which includes rgb_norms.h for example) dt was able to recompile at start up.

Issue #2641, opened 2019-05-29
monotonic splines are not monotonic


rabauke:
**Describe the bug**

For monotonic cubic spline interpolation, local extrema of the interpolating curve must coincide with guiding nodes, see https://en.wikipedia.org/wiki/Monotone_cubic_interpolation

In in the color zones module, however, local extrema may occur between two nodes. Possibly also other modules suffer from this issue. 

**To Reproduce**
1. Go to color zones module.
2. Choose nodes as indicated by the figure below.

**Expected behavior**
An interpolating spline that is monotonous between successive nodes.

**Screenshots**
Nodes are highlighted by red circles.
![Bildschirmfoto vom 2019-05-29 19-27-02](https://user-images.githubusercontent.com/1159508/58579227-df79c100-824a-11e9-8dad-105b48138902.png)

**Platform (please complete the following information):**
 - OS: Ubuntu 18.04
 - Version: Git master

**Additional context**
I re-implemented spline-interpolation routines from scratch extending them to incorporate optionally periodic boundary conditions to address issue https://github.com/darktable-org/darktable/issues/2533 . I became aware of the described issue when I compared the behavior of the existing spline-interpolation code with my new implementation (which behaves as expected, i.e., as described in the wikipedia article).

rabauke:
This issue has been addressed in my pull request #3242.

Issue #2533, opened 2019-05-08
Nodes in color zones module, periodic boundary conditions do not work properly


rabauke:
**Describe the bug**
If lightness, saturation or hue are selected in the color zones module by hue periodic boundary condition have to be applied to the interpolating curve. These boundary conditions are not implemented correctly, which leads to various unintuitive effects.

By default, two nodes are present at the extreme left and right end of the hue spectrum, which refer to the same hue due to periodicity of the hue scale. If one of the node is pushed slightly into the direction of the center the other node moves into the opposite direction. Why? This is not the correct way to implement periodic boundary conditions for an interpolation curve. Pushing the the nodes further, nodes may even loose contact with the interpolating curve.

**To Reproduce**
Push left and right most nodes.

**Expected behavior**
Proper periodic boundary conditions for the interpolating curve.

**Screenshots**
Nodes are indicate by arrows.
![color_zones](https://user-images.githubusercontent.com/1159508/57357743-ee051900-7173-11e9-909f-1398799b32e1.png)

Another example illustrating the non-periodicity of the interpolating curve. Problem applies to all three interpolation methods.
![color_zones](https://user-images.githubusercontent.com/1159508/57632001-d88f5500-75a0-11e9-92fa-8d902a31c65a.png)

**Platform (please complete the following information):**
 - OS: Ubuntu 18.04
 - Version: git master

Nilvus:
@rabauke: if I'm not wrong, this is fixed since some weeks or months now. I close it, we need to clean issues. I sometimes use this module since few months and never had this issue. Feel free to reopen if I'm wrong.

rabauke:
I think, I was involved in fixing. @Nilvus Thanks for closing this issue.

PR #2644, opened 2019-05-30
color zones: fix periodic boundary conditions when color is selected by hue

rabauke:
Periodic boundary conditions must be applied when color is selected by hue in the color zones module. These periodic boundary conditions **cannot** be realized by tweaking the interpolation nodes, the interpolation routines themselves must obey periodic boundary conditions. Not doing so leads to various bugs, see issue https://github.com/darktable-org/darktable/issues/2533 and issue https://github.com/darktable-org/darktable/issues/2641.

Interpolation routines have been reimplemented from scratch and the color zones module has been revised. This bugfix will break some edits, i.e., old edits might give slightly different results with the revised code.

Code may affect _every_ module that utilizes spline interpolation. Testing needed. Currently, strong artifacts occur in the basecurve module in extreme highlights.


TurboGit:
Can you please rebase on top of current master and solve the conflict? Thanks.

TurboGit:
Looks good to me! Will merge hoping the diff with old edits won't be too big. Anyway, merging sooner will make a better field testing possible.

@rabauke : A side note, do not hesitate to add a comment when you push a fix as otherwise I do not notice the new version. There is no notification for a push, but I do get notification for comments.

edgardoh:
@TurboGit , do you really want to risk breaking compatibility on all the curves, that is all our edits, just because of the periodic boundary on the hue curve on the color zones?

And why the default nodes are now added at the center of the curves? How that makes our work easier? Not to mention the inconsistencies that this introduces.

Can you please consider reverting this?

TurboGit:
> And why the default nodes are now added at the center of the curves? How that makes our work easier? Not to mention the inconsistencies that this introduces.

Can you tell me more about this? I didn't see that on my side.

TurboGit:
>  do you really want to risk breaking compatibility on all the curves, that is all our edits, just because of the periodic boundary on the hue curve on the color zones?

The changes should not be that big, at least what I have tested on my edits. Do you have found issues with that? If it breaks badly some old edits, yes I'll revert.

edgardoh:
The default for the initial two nodes is not 0, 1 anymore, they are in the
middle. This does not makes much sense for most of the edits.

edgardoh:
Is not about the intensity (is that the word?) of the changes, Heiko admits
that there's differences, and we guaranty backward compatibility, so for me
that alone is a don't.

If this makes the periodic boundary more accurate for the hue curve, then
great, let's go for it, but let's modify *only* the hue curve, not all
curves. The way to do this without breaking compatibility is to add new
interpolation methods, initially only on the color zones, and eventually
deprecate the old ones, like has been done for some blend modes. This may
give us more code, but is the price to pay to ensure backward compatibility.

And if the way nodes behaves on the color zones will change, at least give
some notice so we have the chance to discuss it.

TurboGit:
Ok, reverted for now.

aurelienpierre:
> The default for the initial two nodes is not 0, 1 anymore, they are in the middle. This does not makes much sense for most of the edits.

yeah but that's only the settings starting point, is it ? How does that break compatibility for existing nodes ?

What makes me wonder here is mostly the C++ lib in a project where nobody likes that, and no sign of vectorization/parallelization of the loops (as far as I understand C++).

TurboGit:
@rabauke : is it possible to make the initial nodes on 0 1 instead of creating them in the middle of the curve? If not, why was this changed? I'd really like having this merged even if old edits are slightly changed. We are speaking about correctness of the routine. If there is real issues with that, do you think we can keep the old routine and have the new one activated only for new edits?

Last, if possible moving to plain C instead of C++ would be nice. But this is not blocking to me.

Thanks.

TurboGit:
BTW, it would be nice to create a new PR for discussion as I cannot reopen a PR a "marked" as merged.

aurelienpierre:
@Turbogit I guess if the periodic condition is respected, the 0 node == 1 node (that's the whole point of the periodicity) so having both makes no sense anymore.

Issue #2618, opened 2019-05-23
dt2.7 "crop and rotate" active/inactive switch should come back to a 2-state logic


leclercj:
**Describe the bug**
In darkatble 2.7, the "crop and rotate" active/inactive switch ⏻ in darkroom toggles between 3 different states, whatever the module is expended or collapsed:
  1. active / original picture displayed + crop zone visible
  2. active / cropped picture displayed
  3. inactive / original picture displayed

**To Reproduce**
1. Go to 'darkroom'
2. Use the 'crop and rotate' module to resize the picture
3. Click on the activate/deactivate switch ⏻ several times
4. See the different possible states

**Expected behavior**
Come back to dt2.6, with 2-state toggling: active/inactive

If module is expended:
  1. active / original picture displayed + crop zone visible
  2. inactive / original picture displayed, crop zone hidden

If module is collapsed:
  1. active / cropped picture displayed
  2. inactive / original picture displayed, crop zone hidden

**Platform (please complete the following information):**
 - OS: ubuntu 18.10 or Windows 7
 - darktable version: 2.7 master (2.6 is not affected)

**Additional context**
- I'm not using the "expends the module when it is activated, and collapse it when disable" option
- Same issue also with "perspective correction" module

leclercj:
After some investigations, I believe this regression has been introduced by https://github.com/darktable-org/darktable/commit/309f20b2665f84015aa9bc7ea295a5cfb60b1de5 and https://github.com/darktable-org/darktable/commit/23c81b9c81e15d9eeef7d137ac711ae2dc70a0ed commits.
More specifically, I can fix this issue by reverting back the code of dt_iop_gui_off_callback() in develop/imageop.c, but this is breaking the "automatically expand modules when switching on, and collapse when swttching off" code.
Here I'd appreciate help from @aurelienpierre

leclercj:
https://github.com/darktable-org/darktable/issues/2535 has same origin

TurboGit:
@leclercj : can you please try #2642 ?

Need to check very carefully as this is a tricky area. Thanks.

Issue #2535, opened 2019-05-08
Darkroom: Correction shapes should not be visible when a module is not expanded


leclercj:
**Describe the bug**
In the darkroom, when retouch or spot removal modules are used, correction shapes should not be visible when the modules are not expanded

**To Reproduce**
1. Go to 'Darkroom'
2. Use retouch or spot removal modules to correct the picture (might be applicable to other modules as well, I don't know)
3. Collapse the module
4. Switch the module OFF (deactivate)
5. Switch the module ON (activate) again and see the correction shapes displayed on the picture

**Expected behavior**
Correction shapes should not be visible when the module is not expanded
![Capture d’écran du 2019-05-08 19-51-43](https://user-images.githubusercontent.com/22872204/57396481-c8572e80-71ca-11e9-957e-2370b02330fe.png)

At least this is what looks logical to me and what I'm observing with dt 2.6

**Platform (please complete the following information):**
 - OS: Ubuntu
 - Darkatble version 2.7/master

**Additional context**

leclercj:
After some investigations, I believe this regression has been introduced by https://github.com/darktable-org/darktable/commit/309f20b2665f84015aa9bc7ea295a5cfb60b1de5 and https://github.com/darktable-org/darktable/commit/23c81b9c81e15d9eeef7d137ac711ae2dc70a0ed commits.
More specifically, I can fix this issue by reverting back the code of dt_iop_gui_off_callback() in develop/imageop.c, but this is breaking the "automatically expand modules when switching on, and collapse when swttching off" code. Same as https://github.com/darktable-org/darktable/issues/2618
Here I'd appreciate help from @aurelienpierre

PR #2645, opened 2019-05-30
Fix for #2618 and #2535

leclercj:
This is fixing both https://github.com/darktable-org/darktable/issues/2618 and https://github.com/darktable-org/darktable/issues/2535 when auto-expand option is not used
Still some glitch when the auto-expand option is used

### DIFF
--- a/src/develop/imageop.c
+++ b/src/develop/imageop.c
@@ -1071,14 +1071,14 @@ static void dt_iop_gui_off_callback(GtkToggleButton *togglebutton, gpointer user
       if(dt_conf_get_bool("darkroom/ui/scroll_to_module"))
         darktable.gui->scroll_to[1] = module->expander;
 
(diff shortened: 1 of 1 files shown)


edgardoh:
We probably want the module to have the focus when enabled and loose it when disabled, but:

-if the module is not expanded it should not have the focus
-if the module does not have the focus the get_focus(NULL) should not be called

Can you check if with this logic the issues are fixed?

leclercj:
Thanks @edgardoh for your comment. I've added 1 commit to this PR to also fix the "expand when activate" pref case.
I see this PR now fully solving #2618 and #2535. However, will continue to test further, as there are many possible combinations mixing on/off, expand/collapse, and preference flags.

leclercj:
I did more tests and now see this ready for your review to pull in master

PR #2648, opened 2019-06-01
Some more css tweaks

Nilvus:
Two minor tweaks for css themes :
- Most important one (1st commit): entries on clone manager was way too big in height and too short in width. This make entries better sized. This commit also remove a redundant line.
- Text on some modules (bauhaus sliders) has now better contrast, so more readable.

### DIFF
--- a/data/themes/darktable.css
+++ b/data/themes/darktable.css
@@ -1011,13 +1011,13 @@ entry
   margin: 2px 0;
 }
 
-#left #lib-plugin-ui entry /* This will avoid have too much height on entry item in filter collections modules */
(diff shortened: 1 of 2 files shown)


data/themes/darktable.css
```diff
@@ -1017,10 +1017,16 @@ entry
   margin: 8px 0px;
 }
 
+#left #lib-plugin-ui scrolledwindow entry /* This will avoid have too much height on entry item in clone manager without touching other parts of the UI */
```
TurboGit: avoid have -> avoid having
entry item -> entry items ?

Nilvus: Oups. Commit submitted for better English comment ! Thanks !

PR #2650, opened 2019-06-01
 colorzones: add display selection

edgardoh:
This:
- add an option to display the selection on the color zones module, in the same way the blend module does. The selection itself is usually too weak to be visible, so I multiply it by 4, I don't like that very much so suggestions are welcome.
I have also moved the bypass_blendif from the module to the pipe, so different pipes can display the final image or the selection/mask/etc.

- the retouch module now display the scale on the full pipe only, so on darkroom we can have the scale displayed and on the second window the final image.

- fixed a bug on levels, that sometimes it doesn't grab the focus and the cache is not used.

- allow to reset to Y = 0 a single node on color zones by ctrl+right click on it.

### DIFF
--- a/src/develop/blend.c
+++ b/src/develop/blend.c
@@ -2771,7 +2771,7 @@ void dt_develop_blend_process(struct dt_iop_module_t *self, struct dt_dev_pixelp
                               const void *const ivoid, void *const ovoid, const struct dt_iop_roi_t *const roi_in,
                               const struct dt_iop_roi_t *const roi_out)
 {
-  if(self->bypass_blendif && self->dev->gui_attached && (self == self->dev->gui_module)) return;
+  if(piece->pipe->bypass_blendif && self->dev->gui_attached && (self == self->dev->gui_module)) return;
 
   const dt_develop_blend_params_t *const d = (const dt_develop_blend_params_t *const)piece->blendop_data;
   if(!d) return;
@@ -3088,7 +3088,7 @@ int dt_develop_blend_process_cl(struct dt_iop_module_t *self, struct dt_dev_pixe
                                 cl_mem dev_in, cl_mem dev_out, const struct dt_iop_roi_t *roi_in,
                                 const struct dt_iop_roi_t *roi_out)
 {
-  if(self->bypass_blendif && self->dev->gui_attached && (self == self->dev->gui_module)) return TRUE;
+  if(piece->pipe->bypass_blendif && self->dev->gui_attached && (self == self->dev->gui_module)) return TRUE;
 
   dt_develop_blend_params_t *const d = (dt_develop_blend_params_t *const)piece->blendop_data;
   if(!d) return TRUE;
--- a/src/develop/blend_gui.c
+++ b/src/develop/blend_gui.c
@@ -609,19 +609,6 @@ static void _blendop_blendif_showmask_clicked(GtkWidget *button, GdkEventButton 
 {
   if(darktable.gui->reset) return;
 
-  // if blendif is bypassed don't allow to set this button on
-  if(module->bypass_blendif)
-  {
-    dt_control_log(_("display mask is currently disabled by another module"));
-
-    if(darktable.gui->reset) return;
-    const int reset = darktable.gui->reset;
-    darktable.gui->reset = 1;
-    dtgtk_button_set_active(DTGTK_BUTTON(button), FALSE);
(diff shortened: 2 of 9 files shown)


Benja1972:
Nice proposal. Thank you. 

1. Regarding color zone i would also add possibility to add 'region points' (5 points) by plus-color-picker to existing ones. It can help to alter distant  colors at same curve. (same behavior as  shift-selection add to selection in many editors)

2. And one more proposal. If we alter color too much it will produce artifacts on image even if altered color spear over region in hue scale. The reason is that in color zones we don't have fuzziness in color selection. I don't know if it is possible to add it to existing color zone module.

3. And last proposal. Plus-color-picker produce 5 points in all three curves (lightness, saturation and hue). Now it works like this: by plus-color-picker I select rang of colors. It produce 5 points on one active curve. Than I change tab and click one more time on plus-picker, it produce 5 point on second curve at same positions as range of selection rest same in buffer. And than I finish by clicking on plus-picker for third tab resulting on same 5 points on last curve. There is bug probably, because this scenario doesn't work until I alter some points on first curve.

Benja1972:
Here an example of artifacts. I altered lightness of greens much and white robe got grey patches as it has green cast, very slight but enough to be affected.

edgardoh:
> Nice proposal. Thank you.
>
>    1.
>
>    Regarding color zone i would also add possibility to add 'region
>    points' (5 points) by plus-color-picker to existing ones. It can help to
>    alter distant colors at same curve. (same behavior as shift-selection add
>    to selection in many editors)
>
> Not sure what you mean by region points.

>
>    1.
>    2.
>
>    And one more proposal. If we alter color too much it will produce
>    artifacts on image even if altered color spear over region in hue scale.
>    The reason is that in color zones we don't have fuzziness in color
>    selection. I don't know if it is possible to add it to existing color zone
>    module.
>
> fuzziness is done here by the curve itself, we can't blur it as in the
blend module because is not a mask.

>
>    1.
>    2.
>
>    And last proposal. Plus-color-picker produce 5 points in all three
>    curves (lightness, saturation and hue). Now it works like this: by
>    plus-color-picker I select rang of colors. It produce 5 points on one
>    active curve. Than I change tab and click one more time on plus-picker, it
>    produce 5 point on second curve at same positions as range of selection
>    rest same in buffer. And than I finish by clicking on plus-picker for third
>    tab resulting on same 5 points on last curve. There is bug probably,
>    because this scenario doesn't work until I alter some points on first curve.
>
> This is a known issue introduced when doing a performance enhancement, you
need to do some modification on the module data (in this case a node or any
other setting).

edgardoh:
Yes, not much I can do other than replacing the entire algorithm. That's
why 5 nodes are created, you can use the 2 in the middle to define the
'fuzziness'

Benja1972:
5 nodes is OK. 
1. Here I can explain. Suppose I would like to alter reds. I take plus-color picker, select range. Edit curves. Done. Now on same curve I would like alter greens and want change a range of colors by using plus-color-picker one more time. Imagine what happens to my nodes around reds?

3. Here one more clarification. I would like to have an option: plus-color-picker produce 5 nodes on ALL three curves in one instance at absolutely same positions.

edgardoh:
I get it now, the main issue with this is the GUI, we are running out of
modifiers (the shift and ctrl are already used).

For me that won't solve a bigger issue: when working with colors I find
difficult to get the right selection and the right adjustment. What I
usually end up doing is to define the selection with the blend module and
then I use a flat hue curve to get the color I want. Then with different
modules and using a raster mask I adjust the saturation and brightness
(usually with the rgb curve).
That get the job done, but is more work that it should and it doesn't seem
I'm using the module the right way.

Still not sure what to do about it.

Benja1972:
I understand you. I found a way to use a color zone with parametric mask selecting range by hue. Than I apply flat curves hue, saturation, lightness and alter color. Mask can be highlighted and blurred (etc.). unfortunately  it gives possibility to alter only one color range per color zone module. 

PS as a remark, probably in future we can make a new module for selective color adjustments (color editor). I understand it should not be interfered with existing color zone module.

edgardoh:
@TurboGit , ready the button change.

edgardoh:
@Benja1972 , maybe a new module with just a slider for hue, saturation and brightness is the way to go, but I don't want to create yet another module without having more feedback first.

Issue #2657, opened 2019-06-02
enter/leave notify are not executed on gradient slider


edgardoh:
The enter_notify_event and leave_notify_event are not executed on gradient slider, as a result the active handle is not highlighted unless is being dragged.

https://github.com/darktable-org/darktable/blob/master/src/dtgtk/gradientslider.c#L383

Not sure if this a gtk issue, I tested it before the CSS and iop order were merged and the issue was already there.

dterrahe:
This seems to be resolved on master. Please test and close if appropriate.

johnny-bit:
According to my tests it indeed works on master.

PR #2658, opened 2019-06-03
gradient slider: rework markers highlight

edgardoh:
This changes the way markers (the triangles) are highlighted on the gradient slider:

- mouse Y position is taken into account, if the mouse is over the top half of the slider only top markers are highlighted, if it is on the bottom half then only bottom markers are highlighted.

- selected marker takes priority, markers selection has not changed, they can be selected by click on it and right click toggles the selection.

- if no marker is selected then the closest one to the mouse is highlighted, depending on the Y position relative to the slider.

- changes are applied to the highlighted marker, i.e. click / drag / keyboard / scroll.

Because of #2657 the markers remain highlighted even if the mouse has leave the slider, but this is better than not knowing which marker will be updated with the change.

### DIFF
--- a/src/dtgtk/gradientslider.c
+++ b/src/dtgtk/gradientslider.c
@@ -71,6 +71,21 @@ static gboolean _gradient_slider_postponed_value_change(gpointer data)
   return DTGTK_GRADIENT_SLIDER(data)->is_dragging; // This is called by the gtk mainloop and is threadsafe
 }
 
+static inline gboolean _test_if_marker_is_upper_or_down(const gint marker, const gboolean up)
+{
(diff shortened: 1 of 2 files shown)


edgardoh:
Right, to select a new marker you have to right click on it, but that's not intuitive at all.

I want to keep the behaviour as close as possible to the regular sliders, so let's try this one:

Now click will update the closest marker, no matter if there's a selected one, so you can drag any marker even if there's a selected one.
The downside is, if there's a selected marker, the closest one won't be highlighted, so it may update a marker other than the highlighted one.
With practice is easy to anticipate which one will be updated, but at first it can be confusing.

TurboGit:
@edgardoh : I agree but on my side with the code I have tested I was not able to grad another slider even on clicking on it.

Issue #2617, opened 2019-05-23
perspective correction: reconsider parameter limits


afr-e:
E.g., lens shift (vertical). See: https://discuss.pixls.us/t/perspective-correction-does-not-respect-proportion/12668/10

>Is there a limit in the vertical lens shift? I cannot seem to be able to correct this image to orthogonality.

>Although PTLens seems to have the same limitation as Darktable, I was able to achieve the following result with Hugin.

>Indeed, it seems the limit is too conservative.

leclercj:
With this PR https://github.com/darktable-org/darktable/pull/2662 we may produce expected result also with darktable:

![65282ac4641242164604211804906cbb73192c63_2_1035x775](https://user-images.githubusercontent.com/22872204/58899958-6f61b400-86fe-11e9-844b-76cf717957bf.jpg)

PR #2662, opened 2019-06-04
ashift: Extend maximum lensshift range

leclercj:
Some pictures need it, and both GUI and automatic calculation algorithm allow this
This fixes https://github.com/darktable-org/darktable/issues/2617


TurboGit:
Looks good to me. Thanks.

PR #2663, opened 2019-06-04
BUGFIX: colorzones parallel loop for OpenMP 4.0

aurelienpierre:
explicitely set `firstprivate` and `shared` variables to avoid compilations errors on GCC 9 / Fedora 30

Ping @edgardoh

### DIFF
--- a/src/iop/colorzones.c
+++ b/src/iop/colorzones.c
@@ -390,7 +390,9 @@ void process(struct dt_iop_module_t *self, dt_dev_pixelpipe_iop_t *piece, const 
     memcpy(ovoid, ivoid, roi_out->width * roi_out->height * ch * sizeof(float));
 
 #ifdef _OPENMP
-#pragma omp parallel for default(none) schedule(static) shared(d)
+#pragma omp parallel for default(none) schedule(static) \
+  dt_omp_firstprivate(ch, ivoid, ovoid, roi_out, display_channel) \
+  shared(d)
 #endif
     for(size_t k = 0; k < (size_t)roi_out->width * roi_out->height; k++)
     {


edgardoh:
This will be quite annoying, is there a way to make Travis warn us about this?

BTW, Travis is complaining here.

aurelienpierre:
It should be possible to add a GCC 9 recipe for Travis. Not sure what is breaking here with GCC 8                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                        De: edgardohEnvoyé: mardi 4 juin 2019 16:25À: darktable-org/darktableRépondre à: darktable-org/darktableCc: Aurélien PIERRE; AuthorObjet: Re: [darktable-org/darktable] BUGFIX: colorzones parallel loop for OpenMP 4.0 (#2663)This will be quite annoying, is there a way to make Travis warn us about this?
BTW, Travis is complaining here.

[
{
"@type": "EmailMessage",
"potentialAction": {
"@type": "ViewAction",
"target": "https://github.com/darktable-org/darktable/pull/2663?email_source=notifications\u0026email_token=AAVGQFMGBXRBGL6AAKM6YP3PYZ3LRA5CNFSM4HS37T22YY3PNVWWK3TUL52HS4DFVREXG43VMVBW63LNMVXHJKTDN5WW2ZLOORPWSZGODW4XTWQ#issuecomment-498694618",
"url": "https://github.com/darktable-org/darktable/pull/2663?email_source=notifications\u0026email_token=AAVGQFMGBXRBGL6AAKM6YP3PYZ3LRA5CNFSM4HS37T22YY3PNVWWK3TUL52HS4DFVREXG43VMVBW63LNMVXHJKTDN5WW2ZLOORPWSZGODW4XTWQ#issuecomment-498694618",
"name": "View Pull Request"
},
"publisher": {
"@type": "Organization",
"name": "GitHub",
"url": "https://github.com"
}
}
]

src/iop/colorzones.c
```diff
@@ -372,8 +372,8 @@ void process_v3(struct dt_iop_module_t *self, dt_dev_pixelpipe_iop_t *piece, con
   }
 }
 
-void process(struct dt_iop_module_t *self, dt_dev_pixelpipe_iop_t *piece, const void *const ivoid,
-             void *const ovoid, const dt_iop_roi_t *const roi_in, const dt_iop_roi_t *const roi_out)
+void process(struct dt_iop_module_t *self, dt_dev_pixelpipe_iop_t *piece, const void *const restrict ivoid,
+             void *const restrict ovoid, const dt_iop_roi_t *const roi_in, const dt_iop_roi_t *const roi_out)
```
edgardoh: I don't think restrict belongs here.

src/iop/colorzones.c
```diff
@@ -390,7 +390,7 @@ void process(struct dt_iop_module_t *self, dt_dev_pixelpipe_iop_t *piece, const
     memcpy(ovoid, ivoid, roi_out->width * roi_out->height * ch * sizeof(float));
 
 #ifdef _OPENMP
-#pragma omp parallel for default(none) schedule(static) shared(d)
+#pragma omp parallel for default(none) schedule(static) dt_omp_firstprivate(ivoid, roi_in, roi_out, ch, d, display_channel) shared(ovoid)
```
edgardoh: This one works for me:

#ifdef _OPENMP
  dt_omp_firstprivate(ch, ivoid, ovoid, roi_out) \
  shared(d) \
  schedule(static)
#endif

src/iop/colorzones.c
aurelienpierre: restrict is just a way to prevent the compiler from detecting false aliasing and may enable further optimizations. It should work this way (already tested in other IOPs).

src/iop/colorzones.c
aurelienpierre: done

src/iop/colorzones.c
```diff
@@ -390,7 +390,9 @@ void process(struct dt_iop_module_t *self, dt_dev_pixelpipe_iop_t *piece, const
     memcpy(ovoid, ivoid, roi_out->width * roi_out->height * ch * sizeof(float));
 
 #ifdef _OPENMP
-#pragma omp parallel for default(none) schedule(static) shared(d)
+#pragma omp parallel for default(none) schedule(static) \
+  dt_omp_firstprivate(ivoid, roi_in, roi_out, ch, d, display_channel) \
+  shared(ovoid)
```
edgardoh: Still getting an error because of this, that's different from my suggestion.

src/iop/colorzones.c
edgardoh: Not sure if process() should be different from its definition, but in any case it should go in a separate PR, or at least on a separate commit.

src/iop/colorzones.c
aurelienpierre: Ahhh sorry about that, I got distracted.

src/iop/colorzones.c
hanatos: interesting indeed that the "restrict" keyword doesn't change the type of the function pointer. since it's a good thing to have here (some vectorisation will insist to have it or else fail) i agree we should change the function definitions around here https://github.com/darktable-org/darktable/blob/master/src/develop/imageop.h#L256

aurelienpierre: ok I removed the restrict keyword, let's do that elswhere

cryptomilk:
I use gitlab in the meantime because the CI is so much better. You just add a docker container and build what you want. So I think there is no Debian with GCC9 yet. Fedora 30 and openSUSE Tumbleweed already provide it. I use both to test the software I care about.

cryptomilk:
The patch looks fine for me.

TurboGit:
Looks ok. Thanks.

PR #2665, opened 2019-06-04
accels : ask before replacing accels

AlicVB:
currently, when you record an already mapped accel, the old one is silently deleted to avoid conflict.
This add a confirmation box in this case to let the user choose.

### DIFF
--- a/src/gui/accelerators.c
+++ b/src/gui/accelerators.c
@@ -243,7 +243,7 @@ void dt_accel_register_lua(const gchar *path, guint accel_key, GdkModifierType m
 }
 
 
-static dt_accel_t *_lookup_accel(gchar *path)
+static dt_accel_t *_lookup_accel(const gchar *path)
 {
   GSList *l = darktable.control->accelerator_list;
   while(l)
@@ -1067,6 +1067,10 @@ void dt_dynamic_accel_get_valid_list()
   }
 }
 
+dt_accel_t *dt_accel_find_by_path(const gchar *path)
+{
+  return _lookup_accel(path);
+}
 // modelines: These editor modelines have been set for all relevant files by tools/update_modelines.sh
 // vim: shiftwidth=2 expandtab tabstop=2 cindent
 // kate: tab-indents: off; indent-width 2; replace-tabs on; indent-mode cstyle; remove-trailing-spaces modified;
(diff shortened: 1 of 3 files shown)


AlicVB:
should be fixed now

TurboGit:
Something strange:
- open the pref
- select global options
- in slideshow I have "S" currently
- I try to use S for "change view" (no accel at the moment)

dt says to me that S accel is already registered for "views/darkroom/brush layer".

But when I go into this section I see that there is no accel for "brush larger" !

AlicVB:
oh... that's strange indeed. I can't reproduce on my side...
just to be sure : 
- it says "views/darkroom/brush la**rg**er" (suppose a typo in your previous msg)
- did it say that in english or in translated path (like "vue/chambre noire/augmenter la taille du pinceau" for you) ?
- can you reproduce, or does it appears just 1 time ?
- did you have some other accel registered for "views/darkroom/brush larger"

TurboGit:
> oh... that's strange indeed. I can't reproduce on my side...
> just to be sure :
> 
>     * it says "views/darkroom/brush la**rg**er" (suppose a typo in your previous msg)

Right, a typo.

>     * did it say that in english or in translated path (like "vue/chambre noire/augmenter la taille du pinceau" for you) ?

In English, but all my dt is in french. So that's strange to.
 
>     * can you reproduce, or does it appears just 1 time ?

It reproduces all the time.

>     * did you have some other accel registered for "views/darkroom/brush larger"

```
$ grep "\"s\"" ~/.config/darktable/keyboardrc
; (gtk_accel_path "<Darktable>/global/slideshow view" "s")
(gtk_accel_path "<Darktable>/views/darkroom/brush larger" "s")
```

So it looks like S (default for slideshow) is the one displayed in the accel pref but actually the one set is "brush larger" which is not displayed as being set in the accel pref!

Something fishy...

AlicVB:
> $ grep "\"s\"" ~/.config/darktable/keyboardrc
> ; (gtk_accel_path "<Darktable>/global/slideshow view" "s")
> (gtk_accel_path "<Darktable>/views/darkroom/brush larger" "s")

Those lines means that you have manually set 's' for brush larger, while slideshow has not been touch...
- This should not have happen at all with the current master behavior, as it should have silently reset slideshow accel ! (same behavior with this PR, except the confirmation box)
- The path should have been translated
- The accel for brush larger should have been displayed in the prefs

So yes something is fishy...

There's some other strange things in this areas, thought, like the fact that you can have several same default shortcuts in different views, but if you change one, you reset the other, ...
If you agree, I propose to close this PR for the moment and to prepare a larger one.

PR #2677, opened 2019-06-05
Travis : add Clang 8 and GCC 9 recipies

aurelienpierre:
Update compilers versions

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -62,6 +62,16 @@ matrix:
     sudo: required
     services:
       - docker
+    env: CC=clang-8 CXX=clang++-8 TARGET=build
+  - os: linux
+    sudo: required
+    services:
+      - docker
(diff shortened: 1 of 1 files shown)


aurelienpierre:
It seems something more needs to be done on the virtual host doing the building. Maybe @LebedevRI could check ?

LebedevRI:
As you may be able to notice from the file you are editing, those ci entries are based on docker image.
That docker image is based on debian testing, which is presently frozen in preparation for next stable
release. And the debian testing, presently, does not have gcc-9/clang-8, otherwise i would have updated them already a while ago.

LebedevRI:
gcc 9 migrated to testing, ci updated; no clang 8 yet..

TurboGit:
Thanks Roman! Will help detecting issues.

PR #2678, opened 2019-06-05
add target clones

aurelienpierre:
For compilers that support it, add target clone support. By tagging the functions containing vectorized loops and such, it will compile the functions several times with special optimizations for different SSE generations, in order to get optimized versions in generic release builds.

### DIFF
--- a/src/common/darktable.h
+++ b/src/common/darktable.h
@@ -93,6 +93,15 @@ typedef unsigned int u_int;
 
 #endif /* _OPENMP */
 
+/* Create cloned functions for various CPU SSE generations */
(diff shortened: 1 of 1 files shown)


TurboGit:
This does nothing at the moment, right? It is a preparation for other work or we re missing some commits?

aurelienpierre:
> This does nothing at the moment, right? It is a preparation for other work or we re missing some commits?

This does nothing at the moment indeed.

TurboGit:
Ok that was understanding, just wanted to be sure a commit was not missing.
Let's go with that for now.

PR #2679, opened 2019-06-05
Optimize exposure module

aurelienpierre:
Make the simple `process` function run on-par with the SSE one (± 2 ms) by setting OpenMP properly. Collapse the loops and make use of fused multiply-add for performance.

Notice that I get CPU runtimes about half the OpenCL one (8 ms with SSE on Intel Xeon 1505M @ 3 Ghz, 16 ms on Nividia Quadro M2000). I think we should deprecate the OpenCL kernel since the algorithmic complexity is not worth the buffer copies overhead.

### DIFF
--- a/src/iop/exposure.c
+++ b/src/iop/exposure.c
@@ -396,6 +396,7 @@ error:
 }
 #endif
 
+__DT_CLONE_TARGETS__
 void process(struct dt_iop_module_t *self, dt_dev_pixelpipe_iop_t *piece, const void *const restrict i, void *const restrict o,
              const dt_iop_roi_t *const roi_in, const dt_iop_roi_t *const roi_out)
 {
@@ -427,6 +428,7 @@ void process(struct dt_iop_module_t *self, dt_dev_pixelpipe_iop_t *piece, const 
 }
 
 #if defined(__SSE__)
+__DT_CLONE_TARGETS__
 void process_sse2(struct dt_iop_module_t *self, dt_dev_pixelpipe_iop_t *piece, const void *const restrict i,
                   void *const restrict o, const dt_iop_roi_t *const roi_in, const dt_iop_roi_t *const roi_out)
 {
--- a/src/iop/exposure.c
+++ b/src/iop/exposure.c
@@ -396,7 +396,7 @@ error:
 }
 #endif
 
-void process(struct dt_iop_module_t *self, dt_dev_pixelpipe_iop_t *piece, const void *const i, void *const o,
+void process(struct dt_iop_module_t *self, dt_dev_pixelpipe_iop_t *piece, const void *const restrict i, void *const restrict o,
              const dt_iop_roi_t *const roi_in, const dt_iop_roi_t *const roi_out)
 {
   const dt_iop_exposure_data_t *const d = (const dt_iop_exposure_data_t *const)piece->data;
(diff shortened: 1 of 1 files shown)


parafin:
Exactly because of buffer copies overhead OpenCL kernel should be present.

aurelienpierre:
‎Are you sure the image buffers stay on GPU vRAM between two OpenCL kernels ? Because I got the impression that OpenCL kernels are triggered by the CPU and the image buffers go back and forth between each kernel and RAM to vRAM.                                                                                                                                                                                                                                                                                                                                                                                                                   De: parafinEnvoyé: mercredi 5 juin 2019 22:30À: darktable-org/darktableRépondre à: darktable-org/darktableCc: Aurélien PIERRE; AuthorObjet: Re: [darktable-org/darktable] Optimize exposure module (#2679)Exactly because of buffer copies overhead OpenCL kernel should be present.

[
{
"@type": "EmailMessage",
"potentialAction": {
"@type": "ViewAction",
"target": "https://github.com/darktable-org/darktable/pull/2679?email_source=notifications\u0026email_token=AAVGQFJ76VO2XB63DO644O3PZAO3BA5CNFSM4HUGG5YKYY3PNVWWK3TUL52HS4DFVREXG43VMVBW63LNMVXHJKTDN5WW2ZLOORPWSZGODXA5NOI#issuecomment-499242681",
"url": "https://github.com/darktable-org/darktable/pull/2679?email_source=notifications\u0026email_token=AAVGQFJ76VO2XB63DO644O3PZAO3BA5CNFSM4HUGG5YKYY3PNVWWK3TUL52HS4DFVREXG43VMVBW63LNMVXHJKTDN5WW2ZLOORPWSZGODXA5NOI#issuecomment-499242681",
"name": "View Pull Request"
},
"publisher": {
"@type": "Organization",
"name": "GitHub",
"url": "https://github.com"
}
}
]

parafin:
Yes, I'm sure.

aurelienpierre:
Ok then. But how do you explain that the massively parallelized device takes twice as much time to perform a  simple FMA than a regular CPU ?

parafin:
Probably the measurement method is incorrect - for example it may take transfer times into account too. E.g. if exposure is the only iop in the pipe, OpenCL will be slower than CPU even if GPU executes the calculations much faster, because of input and output data transfers to/from GPU memory.

aurelienpierre:
Does someone know what makes Clang 7 crash ?

phweyland:
> 
> 
> Does someone know what makes Clang 7 crash ?


```

In file included from src/iop/introspection_exposure.c:29:
/build/darktable/src/iop/exposure.c:416:9: error: loop not vectorized: failed explicitly specified loop vectorization [-Werror,-Wpass-failed=loop-vectorize]
        ^
1 error generated.
```

FreddieWitherden:
Can someone confirm the rationale for the explicit `fmaf` call?  In my experience compilers generally do a very good job at spotting opportunities to issue FMA's.  Further, unless `FP_FAST_FMAF` is defined you are likely to suffer severe performance penalties on systems which do not have hardware support for FMA's for it will be emulated in software.  Moreover, given the nature of the kernel I highly doubt there is any performance edge to issuing an FMA over a chained MUL + ADD.  Both loops will achieve the same throughput.

TurboGit:
I'm closing this as it lost traction, feel free to reopen if you plan to work on it.

Issue #2681, opened 2019-06-05
3D Lut from Windows to Linux


gadolf66:
**Describe the bug**
In linux, 3D Lut module shows a warning when it tries to load a xmp with 3D lut created on windows

**To Reproduce**
Add a 3D Lut in Windows. Load the xmp in Linux.
**Expected behavior**
lut loaded
**Screenshots**
![image](https://user-images.githubusercontent.com/46730249/58995642-64c12080-87cb-11e9-83e4-535d22fafbbf.png)

**Platform (please complete the following information):**
 - OS: Ubuntu 18.04
 - Version [e.g. 2.7.0]

**Additional context**
Issue found while doing a Play Raw session in discuss.pixls.us: (https://discuss.pixls.us/t/play-raw-playing-with-b-w/12868/9?u=gadolf)

matze:
It's practically impossible to share such edits at the moment because the LUTs are not embedded into the xmp but load from a fixed external path (and this has of course nothing to do with the operating system). #2534 might help in the future but even that won't help with existing edits.

edgardoh:
I'm not familiar with this module, but if it saves the relative path you can save your luts on one that can be accessed in both linux and windows (i.e. the config directory)

phweyland:
I haven't paid attention too much to potential filepath issues between different OS. 
I'll try to find if dt has some resources to overcome this.

phweyland:
I've tried to fix that PR #2690 but I cannot check the effect on Linux...

johnny-bit:
AFAIK LUT3D files are external dependencies for edits, meaning that if one was to share edits done with lut included, the lut file would have to be added too (and then lut module path adjusted).

There's also another problem: some lut files are commercial + copyrighted, meaning you're not supposed to share them in things like PlayRAWs etc ;) So imo this is not really a bug.

(Also I tried sharing xmp file with my wife who has same luts as I do in same structure, but on windows and it seemed to work so... Fixed?)

PR #2690, opened 2019-06-07
fix relative filepath issue between OS - issue #2681

phweyland:
Try to fix the issue #2681 
Before using the filepath parameter it should convert '\\' characters to '/' on Linux. The other way on Windows.

### DIFF
--- a/src/iop/lut3d.c
+++ b/src/iop/lut3d.c
@@ -793,13 +793,8 @@ void process(struct dt_iop_module_t *self, dt_dev_pixelpipe_iop_t *piece, const 
 void filepath_set_OS_separator(char *filepath)
 {
   const int len = strlen(filepath);
-#if defined(_WIN32)
-  for(int i=0; i<len; ++i)
-    if (filepath[i]=='/') filepath[i] = '\\';
-#else
   for(int i=0; i<len; ++i)
     if (filepath[i]=='\\') filepath[i] = '/';
-#endif
 }
 
 void init(dt_iop_module_t *self)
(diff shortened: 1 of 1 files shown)


src/iop/lut3d.c
```diff
@@ -790,6 +790,18 @@ void process(struct dt_iop_module_t *self, dt_dev_pixelpipe_iop_t *piece, const
   }
 }
 
+void filepath_set_OS_separator(char *filepath)
```
TurboGit: I would just replace all \ to / on Windows as / is also supported on Windows. This will does nothing on Linux as there is no \ in PATH. It will be simpler and won't need OS dependent code.

phweyland: Excellent ! Done

phweyland:
Done as suggested

src/iop/lut3d.c
```diff
@@ -790,6 +790,13 @@ void process(struct dt_iop_module_t *self, dt_dev_pixelpipe_iop_t *piece, const
...
```
TurboGit: Maybe change the name now :) : filepath_set_unix_separator

phweyland: It is not only for unix ... the goal is to be ok on the 3 main OS.
Would be filepath_set_all_OS_separator() convenient ?

TurboGit: No, not ok. The point is that the routine must be named to convey the actual implementation. The implementation is to change dir separators to UNIX style. So filepath_set_unix_separator() looks perfect to me.

phweyland: Question of standpoint. Not a big deal. I do it.

TurboGit:
Closing as part of #2697 too.

PR #2692, opened 2019-06-07
Make disk export more configurable

d2r2:
Make image disk export even more configurable with implementation of 2 extra variables: `$(MAX_WIDTH)` and `$(MAX_HEIGHT)`. New variables allow to create path template like this `$(FILE_FOLDER)/darktable_exported_$(MAX_WIDTH)x$(MAX_HEIGHT)/$(FILE_NAME)`, which in turn make it possible to create multiple `darktable_exported` folders. In my case, with this template I can generate full resolution images and images limited to 2048x2048 which automatically sorted by 2 folders (`darktable_exported_0x0` and `darktable_exported_2048x2048`).

Implementation steps taken:
1) Add extra parameters (int max_width, int max_height) to dt_imageio_module_storage_t (*store) method. All dt_imageio_module_storage_t implementations were updated.
2) Add public `$(MAX_WIDTH)`, `$(MAX_HEIGHT)` variables to expand, once found in export path template.
3) Lua API doesn't change (store function), since I don't want to break existing scripts. Though it is possible to change in the future.
4) Add russian localization for `$(MAX_WIDTH)`, `$(MAX_HEIGHT)` variables description.

### DIFF
--- a/src/common/import_session.c
+++ b/src/common/import_session.c
@@ -217,14 +217,6 @@ void dt_import_session_set_filename(struct dt_import_session_t *self, const char
 }
 
 
-/*
-void dt_import_session_set_max_width_height(struct dt_import_session_t *self, int max_width, max_height)
-{
-  dt_variables_set_max_width_height(self->vp, max_width, max_height);
-}
-*/
-
-
 int32_t dt_import_session_film_id(struct dt_import_session_t *self)
 {
   if(self->film) return self->film->id;
--- a/src/common/import_session.c
+++ b/src/common/import_session.c
@@ -217,6 +217,14 @@ void dt_import_session_set_filename(struct dt_import_session_t *self, const char
 }
 
 
+/*
+void dt_import_session_set_max_width_height(struct dt_import_session_t *self, int max_width, max_height)
+{
+  dt_variables_set_max_width_height(self->vp, max_width, max_height);
+}
+*/
+
+
 int32_t dt_import_session_film_id(struct dt_import_session_t *self)
 {
   if(self->film) return self->film->id;
--- a/doc/usermanual/lighttable/panels/export.xml
+++ b/doc/usermanual/lighttable/panels/export.xml
@@ -133,6 +133,22 @@
             </row>
             <row>
               <entry>
+                <code>$(MAX_WIDTH)</code>
+              </entry>
+              <entry>
+                maximum image width to limit within export session
+              </entry>
+            </row>
+            <row>
+              <entry>
+                <code>$(MAX_HEIGHT)</code>
+              </entry>
+              <entry>
+                maximum image height to limit within export session
+              </entry>
+            </row>
+            <row>
+              <entry>
                 <code>$(YEAR)</code>
               </entry>
               <entry>
--- a/doc/usermanual/preferences/session_options.xml
+++ b/doc/usermanual/preferences/session_options.xml
@@ -97,6 +97,22 @@
         </row>
         <row>
           <entry>
+            $(MAX_WIDTH)
+          </entry>
+          <entry>
+            maximum image width to limit within export session
+          </entry>
+        </row>
+        <row>
+          <entry>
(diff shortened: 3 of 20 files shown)


src/common/import_session.c
```diff
@@ -217,6 +217,14 @@ void dt_import_session_set_filename(struct dt_import_session_t *self, const char
 }
 
 
+/*
```
TurboGit: Either remove or put a comments explaining why is this commented out code.

TurboGit: BTW, I'm not sure to understand why this is needed on session?

d2r2: @TurboGit, I agree. I mistakenly left this code commented out, but certainly it should be removed.

d2r2: @TurboGit, resolved.

d2r2:
@TurboGit, taking this opportunity, want to ask: since this change will go to the new major release, do I have to create same change for branch `darktable-2.6.x`? Again, some time ago I propose Lumix LX-100M2 changes for `darktable-2.6.x`. Will they automatically get into the `master` branch, or I should prepare similar change for `master`?

TurboGit:
Thanks. Merged now.

TurboGit:
> same change for branch darktable-2.6.x

If you want this to be in 2.6, why not but i'm not sure a new 2.6 release will be made as there is some many changes in master that merging will be difficult. You can see if this change is easy enough to merge.

TurboGit:
> Again, some time ago I propose Lumix LX-100M2 changes for `darktable-2.6.x`.

AFAIK the white-balance has been added into master and will be added into 2.6.x later if needed.

d2r2:
@TurboGit, thanks for the clarification!

houz:
could you please tell me why you needed these two variables? the width/height settings in the export module were already max values, that describe the bounding box of the exported image.

d2r2:
@houz, as I describe above, using this new variables, you can generate multiple `darktable_exported` folders, where $(MAX_WIDTH) and $(MAX_HEIGHT) will be constant values over export session, thus let you create, for instance, output folders, kind of `darktable_export_0x0` and `darktable_export_2048x2048`. Or if explained differently, $(MAX_WIDTH) and $(MAX_HEIGHT) is a copy of GUI `max size` field values.

houz:
My point is, the max width/height is already in fdata. Why did you need all those api changes?

d2r2:
@houz, it seems I have missed this opportunity (although I remember that fdata have fields named max_width, max height).

To be exactly sure, I just replace line in my code:
```
dt_variables_set_max_width_height(d->vp, max_width, max_height);
```
with:
```
dt_variables_set_max_width_height(d->vp, fdata->max_width, fdata->max_height);
```
... and as I see everything work as it should.

So, should I create new change with reverting (*store) API change?

houz:
If it's not too much of a problem I would welcome that. We are already piling up too many arguments in these internal functions, and having even more that duplicate other data fields sounds like a sure way to confuse people.

d2r2:
@houz, I agree that it is no sense to change API just for nothing. I will create a correction update later today. Thanks for pointing to this!

d2r2:
@houz, last question: why `gboolean high_quality, gboolean upscale` variables in `(*store)` method not tied to fdata structure? Or this is, as you stated, question of not optimal variable allocation?

TurboGit:
Oh! I missed the fact that the value were already available. So houz is right.

d2r2:
@TurboGit, then I have following question: why gboolean high_quality, gboolean upscale variables in (*store) method not tied to fdata structure as well? Or this is, as houz stated, question of not optimal variable allocation?

TurboGit:
Not sure about this. AFAIK this was done long time ago, maybe we can improve this part. But I would suggest to do that in another PR anyway.

d2r2:
@TurboGit, thanks for reply. OK, for now I will return API to the original view in next PR.

PR #2697, opened 2019-06-08
Lut3d - make lut filename scrollable 

phweyland:
I've lost the reference to this suggestion but I've found it excellent.

With this change, once the user has selected a lut file, all files sharing the same extension in the current folder are added in the filepath combobox.
Then the mouse wheel let the user scroll the combobox and apply the corresponding luts (wihout opening the file selection dialog box).

### DIFF
--- a/src/iop/lut3d.c
+++ b/src/iop/lut3d.c
@@ -794,7 +794,6 @@ void process(struct dt_iop_module_t *self, dt_dev_pixelpipe_iop_t *piece, const 
 void init(dt_iop_module_t *self)
 {
   setvbuf(stdout, NULL, _IONBF, 0);
-  printf("init\n");
   self->data = NULL;
   self->params = calloc(1, sizeof(dt_iop_lut3d_params_t));
   self->default_params = calloc(1, sizeof(dt_iop_lut3d_params_t));
@@ -843,7 +842,6 @@ void cleanup_global(dt_iop_module_so_t *module)
 void commit_params(struct dt_iop_module_t *self, dt_iop_params_t *p1, dt_dev_pixelpipe_t *pipe,
                    dt_dev_pixelpipe_iop_t *piece)
 {
-  printf("commit_params\n");
   dt_iop_lut3d_params_t *p = (dt_iop_lut3d_params_t *)p1;
   dt_iop_lut3d_data_t *d = (dt_iop_lut3d_data_t *)piece->data;
   if (strcmp(p->filepath, d->params.filepath) != 0)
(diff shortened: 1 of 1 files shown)


TurboGit:
Crash at startup:

```
Using host libthread_db library "/lib/x86_64-linux-gnu/libthread_db.so.1".
0x00007ff81ce2546a in __GI___waitpid (pid=pid@entry=20604, stat_loc=stat_loc@entry=0x0, options=options@entry=0) at ../sysdeps/unix/sysv/linux/waitpid.c:30
30	../sysdeps/unix/sysv/linux/waitpid.c: Aucun fichier ou dossier de ce type.
warning: Currently logging to /tmp/darktable_bt_OED52Z.txt.  Turn the logging off and on to make the new setting effective.
#0  0x00007ff81ce2546a in __GI___waitpid (pid=pid@entry=20604, stat_loc=stat_loc@entry=0x0, options=options@entry=0) at ../sysdeps/unix/sysv/linux/waitpid.c:30
#1  0x00007ff81d050228 in _dt_sigsegv_handler (param=11) at /home/obry/dev/builds/darktable/build/src/src/common/system_signal_handling.c:118
#2  0x00007ff81cd96840 in <signal handler called> () at /lib/x86_64-linux-gnu/libc.so.6
#3  0x00007ff81cebb181 in __strlen_avx2 () at ../sysdeps/x86_64/multiarch/strlen-avx2.S:65
#4  0x00007ff7afdc702e in check_same_extension (filename=0x5556f2dc2443 ".", ext=0x0) at /home/obry/dev/builds/darktable/build/src/src/iop/lut3d.c:933
#5  0x00007ff7afdc715e in update_filepath_combobox (filepath=0x5556eded8000 "", lutfolder=<optimized out>, g=<optimized out>) at /home/obry/dev/builds/darktable/build/src/src/iop/lut3d.c:958
#6  0x00007ff7afdc792f in update_filepath_combobox (g=0x5556f1c10a60, lutfolder=0x5556f1dadaa0 "/home/obry/Images/Photos/3dlut", filepath=0x5556eded8000 "") at /home/obry/dev/builds/darktable/build/src/src/iop/lut3d.c:1059
#7  0x00007ff7afdc792f in gui_update (self=<optimized out>) at /home/obry/dev/builds/darktable/build/src/src/iop/lut3d.c:1059
#8  0x00007ff81d0860df in dt_iop_gui_update (module=0x5556f1848530) at /home/obry/dev/builds/darktable/build/src/src/develop/imageop.c:1630
#9  0x00007ff81d0860df in dt_iop_gui_update (module=0x5556f1848530) at /home/obry/dev/builds/darktable/build/src/src/develop/imageop.c:1622
```

phweyland:
I've also introduced the filepath_set_unix_separator() because separator is also visible in the filepath combobox.

phweyland:
> the scroll direction is reversed. Pulling the scroll toward me I should do down in the list (as for combo) and that's not the case.

? For me it works like you describe and as the other comboboxes ...
Do you have any idea where the issue you see could come from ?

TurboGit:
No, don't know. When I scroll on the color space combo just below the 3dlut file I go from sRGB to REC.709, so down the list. But when I select the first 3dlut in a folder, when I scroll in the same direction I do not go to the next 3dlut file. It is blocked and I need to scroll the other way to get the second 3dlut file. Really strange! That's the only combo working like this.

TurboGit:
I think I found why, it seems the combo is listing the file in the reverse order. That is, the first item on the folder I have selected is at the end of the combo where I would have expected it to be at the top. And this is in turn because I had the sort of my dialog set to reverse. So all in all a "problem" on my side.

PR #2698, opened 2019-06-08
do not use cache for gamma on preview pipe

edgardoh:
When gamma iop is retrieved from cache the final histogram is computed with the gamma output image, causing some rounding errors. By not using the cache on the preview pipe the gamma is always processed and the histogram is correct.
The difference in performance is noticeable if you look for it, but this only happens when a module is turned on/off and the only module processed is gamma & histogram, and that's better than having an incorrect histogram.

Also this do not redraw the main window when the pipe finishes, that was causing a redraw of every widget for every pipe.


TurboGit:
Looks good, thanks.

PR #2705, opened 2019-06-09
accels "remind-me" window

AlicVB:
Accels are really great, but remembering all of those can become a pain at certain point... (or maybe at certain age...)
This allow to quickly show the list of currently active accels keys and mouse actions. By "currently active", I mean the ones you can really use at the moment of the popup.
It's WIP and doesn't currently looks very well but it looks like that actually : 
Capture d’écran_2019-06-09_18-58-05
How it works : 
- set an accel in the pref for global shortcuts -> show accel window (There's currently no default accel. We'll have to decide one in the future, as the feature can be especially useful for beginners)
- When you press this accel, you see the popup on top of everything, and when you release it, the popup disappear.

What is shown : 
1- every accel with valid key which is "accessible" at the current moment.
2- every possible mouse actions at the current moment.

For point 2, I've added a new function in view api to return them. Currently, I've only added some actions for lighttable and darkroom, and even those should be refined and rechecked.

Of course, it's WIP, there's still some work to be done : 
- add all mouse actions for all views, masks, etc...
- enhance the design
- check my english ;)
- check for problem if one wants to interact with the popup
- fix bugs :)

### DIFF
--- a/src/control/control.c
+++ b/src/control/control.c
@@ -757,7 +757,7 @@ int dt_control_key_released(guint key, guint state)
   // this line is here to find the right key code on different platforms (mac).
   // printf("key code pressed: %d\n", which);
 
-  dt_control_accels_t *accels = &darktable.control->accels;
+  const dt_control_accels_t *accels = &darktable.control->accels;
 
   // be sure to reset dynamic accel
   if(darktable.view_manager->current_view->dynamic_accel_current) dt_control_hinter_message(darktable.control, "");
--- a/src/control/control.c
+++ b/src/control/control.c
@@ -736,6 +736,12 @@ int dt_control_key_pressed_override(guint key, guint state)
     dt_dev_modulegroups_search_text_focus(darktable.develop);
     return 1;
   }
+  // show/hide the accels window
+  else if(key == accels->global_accels_window.accel_key && state == accels->global_accels_window.accel_mods)
+  {
+    dt_view_accels_show(darktable.view_manager);
+    return 1;
+  }
   return 0;
 }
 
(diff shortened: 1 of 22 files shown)


AlicVB:
for css gurus, gui refinement should now be doable via css class `accels_window` `accels_window_scroll` `accels_window_box` `accels_window_cat_title` `accels_window_list`
ask me if you need more

TurboGit:
Nice idea indeed!

AlicVB:
current status : 
- mouse actions for drawn masks have been added
- mouse actions for iops have been added (for those which use mouse actions) Like for views, this is done with a new optional function in iop api : `mouse_actions`

what is still to be done (help needed) : 
1. double check if I've not forget something (or done some errors) especially in mouse actions
2. double check and correct spelling
3. add mouse actions for liquify iop (@TurboGit ?)
4. add mouse actions for slideshow, map and print view (I don't really use them)
5. (not sure) is there some special mouse actions on the main view for some libs ?

I think 1 and 2 should be done before merging.
For points 3 and 4, it seems simpler to add them in separate PR after this one is merged.

Thanks for your help

TurboGit:
@AlicVB , what is the question for point 3 above? It seems that you are listing all accels registered, so why liquify should be different? Does it needs to be special cased for some reasons?

AlicVB:
@TurboGit : yes, for accels, everything is ok and straightforward.
But I have also extend this "help" window to mouse actions depending on the context. For example, for gratuated density, I've added ctrl-scroll and shift-scroll which change density and compression. same for some other iop, for drawn masks...
I've not touch liquify as I never use it  and as (If I remember well) your are the author, maybe it'll be easier for you. If you don't want or don't have time, I can do it by reading code code and manual...
Anyway, it can be done later

AlicVB:
should now be ready for review. Thanks

TurboGit:
Working on my side.

One question, the last column for modules was full on my side and I had no scrollbar. I'm wondering if I'm seeing all the accels or if some are not visible because of that? Do we have the possibility to have two columns for modules?

AlicVB:
As I have a quite big screen, everything fit inside the screen here, but I've tested with a smaller window and the scrollbars appears when needed.

For the second point, I'd like a better layout too.
There's the categories which can become to long (esp. modules and iop) but it would also be great if we can display 2 small categories on the same columns...
I've searched and I fear it's not really possible in gtk. Note, that it's very possible that I miss something here, don't hesitate to point me to the right direction in that case :)

Nilvus:
About the layout, I don't know if you know that old shortcut "card". Of course, right column is not useful here but the layout of this card is simple and just great viewing. I don't know if something like this could be use to improve how the "remind-me" window look but having something near that refcard would be great I think. Hope that helps you !

TurboGit:
@AlicVB : if there is a scrollbar no problem. This first version is nice and we can improve it later. Thanks.

AlicVB:
@Nilvus : yes, it's exactly what I have in mind...
The sad news, is that the layout of this refcards is done in latex. Here, we want to done that with gtk, and that's the problem !

PR #2710, opened 2019-06-11
lut3d: sort contents of the filename combobox

jakubfi:
Contents of the LUT filenames combobox are not displayed in any particular order (due to how readdir() works). I think sorting them by name would be a commonly expected behavior.
@phweyland

### DIFF
--- a/src/iop/lut3d.c
+++ b/src/iop/lut3d.c
@@ -945,6 +945,12 @@ gboolean check_same_extension(char *filename, char *ext)
   g_free(fext);
   return res;
 }
+
+static gint list_str_cmp(gconstpointer a, gconstpointer b)
+{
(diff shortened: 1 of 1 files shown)


phweyland:
On windows the list is well ordered.
@TurboGit has made this remark before noting the order was depending on the sort of his dialog set.
So, that will prevent those who know (I don't) and would like to use the dialog set properties to do it.
But for this usage, I agree, that makes sense to have them sorted by name.

phweyland:
And thanks for making me learn about `g_list_sort()` !

parafin:
I have no idea what is “dialog set properties”, but it has no relation to the fact that readdir function returns unsorted results. The fact that on some systems list appears to be sorted in some way should be considered as a coincidence. So this is a bugfix, not a feature.

TurboGit:
Ok, looks good. Thanks.

PR #2711, opened 2019-06-11
Some cleanup for cmake

cryptomilk:
Some cleanup for cmake.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -78,12 +78,6 @@ endif()
 
 MESSAGE(STATUS "Building SSE2-optimized codepaths: ${BUILD_SSE2_CODEPATHS}")
 
-# When cross compiling, CMAKE_INSTALL_PREFIX will point to something like "/opt/darktable", but that's not useful when using the path to load
-# modules on runtime. Then we need something like "C:\Program Files\darktable". Doesn't need to be set when doing regular compiles.
-if(NOT DEFINED RUNTIME_INSTALL_PREFIX)
-  set(RUNTIME_INSTALL_PREFIX ${CMAKE_INSTALL_PREFIX})
-endif(NOT DEFINED RUNTIME_INSTALL_PREFIX)
-
 # Generate multi arch triplet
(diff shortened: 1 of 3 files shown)


packaging/windows/BUILD.txt
```diff
@@ -223,7 +223,6 @@ B) Cross compile on Linux
 
     cmake -DCMAKE_TOOLCHAIN_FILE=../cmake/toolchain_mingw64.cmake \
         -DCMAKE_INSTALL_PREFIX=${INSTALL_PREFIX} \
-        -DRUNTIME_INSTALL_PREFIX=${RUNTIME_PREFIX} \
```
TurboGit: Isn't this breaking cross-compilation?

cryptomilk: It isn't used anywhere. It is also not a CMAKE variable.

CMakeLists.txt
```diff
@@ -79,45 +78,6 @@ endif()
...
-  MESSAGE(FATAL_ERROR "The target platform is not supported!")
-endif(NOT IS_SUPPORTED_PLATFORM)
-MESSAGE(STATUS "Is the target platform supported: ${IS_SUPPORTED_PLATFORM}")
-
-set(CMAKE_REQUIRED_INCLUDES "${CMAKE_REQUIRED_INCLUDES_OLD}")
-unset(CMAKE_REQUIRED_INCLUDES_OLD)
-
-# When cross compiling, CMAKE_INSTALL_PREFIX will point to something like "/opt/darktable", but that's not useful when using the path to load
-# modules on runtime. Then we need something like "C:\Program Files\darktable". Doesn't need to be set when doing regular compiles.
-if(NOT DEFINED RUNTIME_INSTALL_PREFIX)
```
TurboGit: It is used here, in part of the code you've removed.

cryptomilk: This will just set a default value if it is not set, but where is it used. I don't find anything.

TurboGit:
Ok, let's see how this fly then. Merging. Thanks.

PR #2716, opened 2019-06-14
denoiseprofile: new scattering formula (better local contrast)

rawfiner:
This change has subtle effects, but still visible.
The new formula allows to push the parameter value more, as it smooths less the local contrast than the old formula. It basically scatters the search neighbourhood in a way which keeps more points near the centre.
The effect is visible for high values of the parameter (for instance values > 1)

Some examples:
before:
DSCF7071_10
after (the details on the apple are a bit more visible):
DSCF7071_09

before:
P2170287_08
after (the local contrast on the orange part of the bird is better):
P2170287_07

### DIFF
--- a/src/iop/denoiseprofile.c
+++ b/src/iop/denoiseprofile.c
@@ -1286,12 +1286,12 @@ static void process_nlmeans(struct dt_iop_module_t *self, dt_dev_pixelpipe_iop_t
     {
       // This formula is made for:
       // - ensuring that kj = kj_index and ki = ki_index when d->scattering is 0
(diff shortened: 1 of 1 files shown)


Nilvus:
Great works ! That's seems promising !

rawfiner:
I've just realized that I forgot to update the tiling part. I will do this asap.

rawfiner:
Tiling part updated :-)

TurboGit:
Result is maybe subtle but quite visible! Very nice work!

TurboGit:
Love it! Thanks.

rawfiner:
Happy that you like it ;-)
Thanks for the code review, testing and merging :-)

PR #2717, opened 2019-06-14
basic adjustment: add middle grey to contrast computation

edgardoh:
This adds the middle grey setting to the contrast computation on the auto option of the basic adjustments module. It gives a not-so-contrasty result when middle grey is not 50% that I find better for an initial edit, which is the point of this feature.


TurboGit:
Looks good, thanks.

PR #2719, opened 2019-06-14
base curve: add norms

edgardoh:
Added the preserve colors option to base curve and make it the default. This gives a better result as colors goes, avoiding the over-saturated result we usually get.

For now it works only for the base curve, not for fusion.

I have added only the luminance, when #2643 is merged the generic version can be ported here.


TurboGit:
The result is really nice! Thanks.

Issue #2707, opened 2019-06-10
IOP preset menus do not react to button-release


s5k6:
**Describe the bug**
Releasing the mouse button over an entry in a module preset menu closes the menu, but does not apply the preset.  Instead, it is required to release the mouse button over the ≡-symbol before moving it, and to then perform another click on the desired item.

**To Reproduce**
1. Open an image in darkroom, press (and hold) left mouse button on an IOP preset menu (≡-symbol).  E.g., use the "equalizer" module.
2. Drag the mouse down to the preset you want to apply, e.g., "bloom".
3. Release the left mouse button.

The menu will be closed, but the preset will not be applied.

**Expected behavior**
The preset should be applied when releasing the mouse button.  

**Platform (please complete the following information):**
  * current ArchLinux
  * no desktop environment (plain X11 and FVWM)
  * darktable 2.7.0+1375~g6f0010465

Nilvus:
Fixed by #2720 so could be closed (@TurboGit

TurboGit:
Closing!

PR #2720, opened 2019-06-15
Apply selected IOP preset on mouse button release

jakubfi:
Fixes issue #2707.

Expected behavior described in the bug report is common for desktop GUI applications and most likely assumed by majority of users. Additionaly, I'd say there are three cases of selecting an entry from a drop-down menu:
1. Click (press and release) the menu icon - menu opens - move to desired position - click to select.
2. Press on the menu icon - menu opens - move to desided position - release to select.
3. Click the menu icon - menu opens - press on a position - realise that this is not what you wanted - move to another position while holding the mouse button - release to select the correct entry.

When selecting a preset from the presets menu, Darktable allows only for case 1. But for "favourite" and "store new preset.." menu entries it (inconsistently) allows also for 2. and 3.

This patch moves the preset activation from "mouse press" to "mouse release", thus fixing the original issue as well as making the menu behavior consistent.

### DIFF
--- a/src/gui/presets.c
+++ b/src/gui/presets.c
@@ -698,19 +698,17 @@ static void menuitem_pick_preset(GtkMenuItem *menuitem, dt_iop_module_t *module)
   gtk_widget_queue_draw(module->widget);
 }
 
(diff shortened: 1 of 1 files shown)


TurboGit:
Looks good, and working on my side. Thanks.

Issue #2725, opened 2019-06-16
PR #2643 brakes opencl


AxelG-DE:
**Describe the bug**
opencl is not working anymore after darktable 2.7.0+1463~g77b3f7762

cltest gives me:
<kernel>:23:10: fatal error: 'rgb_norms.h' file not found
#include "rgb_norms.h"

bisect gives me:
[77b3f77629f739f3cc00c4a0d83f4b1bb1ccfb72] Tonecurve - added norms to tc RGB (#2643)

@phweyland could you please look into it?
TNX #

MStraeten:
sudo cp <...>/darktable/data/kernels/rgb_norms.h /usr/local/share/darktable/kernels/ (or wherever the kernels are located on your system)helps
the makefiles needs to be updated

AxelG-DE:
I'm lost.
8d19bd6f7 compiles well and opencl works
77b3f7762 compiles too but opencl does not work

Why should I need manual copy action?

MStraeten:
the updated rgbcurve.cl requires the rgb_norms.h the former version didn't
On my system (osx) make install copies the required stuff from my build directory to /usr/local ...
but not the rgb_norms.h file ;)

AxelG-DE:
Got it, will check soon

Anyhow, manual actions will fail on lots of users in the first place, I guess :)

AxelG-DE:
That did the trick...

Hmmm where should be the proper copy of rgb_norms.h being invoked? 
How did you find that fast?

MStraeten:
on osx darktable dev has some strange behaviour if starting from an installed binary (needs several attempts until all kernels are properly compiled to enable opencl) so i usually start it via darktable -d opencl from the command line. In general there aren't problems during stratup but this time compiling of kernels run into an error: missing rgb_norms.h resulting in an disabled opencl
Comparing the source and target kernel directory gave the result.

AxelG-DE:
are you on IRC, if yes, which NICK?

phweyland:
I'm building again from scratch to check what happens.

phweyland:
I've just build this PR again.
rgb-norms.h is not copied in `C:\msys64\opt\dt-tc-norm\share\darktable\kernels`, I've cleared the kernel caches and open_cl starts.

![image](https://user-images.githubusercontent.com/23012047/59564383-8ec0e100-901c-11e9-9840-80cac6b6ad68.png)

[...]

PR #2727, opened 2019-06-16
fix for #2725 - rgb_norms.h missing

phweyland:
attempt to fix the issue rgb_norms.h missing file at open_cl start-up.
the rgb_norms.h file is copied to installed kernels folder.
remain to check that travis is happy.


phweyland:
Travis is happy. Should fix #2725

TurboGit:
Confirmed that this fix OpenCL issue. Too bad I missed that, dt does not issue a console error in such case :(

PR #2728, opened 2019-06-16
lighttable : fix group borders drawing

AlicVB:
currently group borders are not always drawn when you enter a group, but sometimes when you leave it or when you move to another image...

In fact it's just that we need to get group_id _after_ we have "computed" the mouse_over_id


TurboGit:
Perfect! Thanks.

PR #2735, opened 2019-06-18
use CAIRO_FILTER_GOOD on navigation instead of CAIRO_FILTER_BEST

edgardoh:
Use CAIRO_FILTER_GOOD on navigation instead of CAIRO_FILTER_BEST. The difference in quality is visible depending on the image, but the navigation don't need to be high quality, and this makes the sliders and curves a lot more responsive, and saves some time on the overall process.
The CAIRO_FILTER_FAST is (obviously) better for performance, but the difference in quality is too noticeable, so let's try this one first.

Also, don't redraw the navigation when full pipe finish, it only need the preview pipe.


TurboGit:
Nice! Thanks.

PR #2737, opened 2019-06-18
filmstrip : reuse in-memory sqlite table

### LINKED ISSUE #2734: Overall preset speed decreases when "collect images" is in tag mode
**Describe the bug**
Some actions on darkroom mode, most noticeably applying a preset, become much slower when "collect images" is using tag.

**To Reproduce**
1. Open darktable with a larger database (mine has around 340MB)
2. In lighttable click on 'collect images' and select a set of images per tag
3. Select a image and go darkroom mode
4. Click any preset

**Expected behavior**
It should be significantly slower than normal, in my case was around 7 to 10x slower

 - OS: Ubuntu 19.04
 - Version 2.6.2

**Additional context**
Suddenly I noticed a huge hit on performance for no apparent reason, some actions were taking much longer than usual. Initially I thought it could be related with OpenCL and started running darktable in debug mode and checking the logs. I noticed no issues in the pipeline with the OpenCL side, but several queries with LIKE where clauses got me suspicious, sometimes  an action would trigger dozens and dozens of queries like the one below. I switched the Collect Images from Tag to another and problem was no longer there. Changing the collect  and were able to replicate the behavior.

167,811531 [sql] /build/darktable-xuXDPt/darktable-2.6.2/src/libs/tools/filmstrip.c:693, function _lib_filmstrip_draw_callback(): prepare "SELECT DISTINCT id FROM images AS a WHERE   (flags & 256) != 256 and (flags & 7) >= 1 AND (flags & 7) != 6 AND (1=1 AND (id IN (SELECT imgid FROM main.tagged_images AS a JOIN data.tags AS b ON a.tagid = b.id WHERE **name LIKE 'Places|Europe|Netherlands'**))) ORDER BY datetime_taken DESC, filename DESC, version DESC LIMIT ?1, ?2"

AlicVB:
this will speedup redraw and keep shuffle order on if needed
this fix #2734

### DIFF
--- a/src/libs/tools/filmstrip.c
+++ b/src/libs/tools/filmstrip.c
@@ -760,22 +760,11 @@ static gboolean _lib_filmstrip_draw_callback(GtkWidget *widget, cairo_t *wcr, gp
   const int img_pointerx = (int)fmodf(pointerx, wd);
   const int img_pointery = (int)pointery;
 
-  const dt_collection_sort_t current_sort = dt_collection_get_sort_field(darktable.collection);
-  const gboolean reverse = dt_collection_get_sort_descending(darktable.collection);
-
-  // we disable the shuffle sort on the filmstrip as this cannot be work with the current implementation. On each redraw
-  // we get a new order for the collection.
-  if(current_sort == DT_COLLECTION_SORT_SHUFFLE)
-  {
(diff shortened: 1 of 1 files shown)


TurboGit:
Looks good, thanks!

PR #2739, opened 2019-06-19
mask gradient enhancements

### LINKED ISSUE #2730: [FEATURE] Improve mask drawing
I would like to have some easier mask drawing. Currently you have to do a lot of clicks to get a result. If we use the gradient as an example.

You have to select it, click somewhere in the picture, then change the direction then change the transition zone.

I would like to click in the picture and keep the mouse button pressed. The direction I move the mouse to now defines the direction of the gradient. The further I move the mouse away from the first point the bigger the transition zone gets. This would make it much faster to draw a gradient mask.

The same can be applied to circles. The first clicks selects the middle, if I move it away from the mouse I can define the size. However it would need to drawn already while I move the mouse.

AlicVB:
this change : 
- show a preview of the gradient at creation time
- allow to change gradient compression at creation time with mouse scroll
- allow to set gradient rotation by click+drag the gradient line at creation time

Gradient now works like circles, ellipses, ...
This fix #2730

### DIFF
--- a/src/develop/masks/gradient.c
+++ b/src/develop/masks/gradient.c
@@ -350,6 +350,12 @@ static int dt_gradient_events_mouse_moved(struct dt_iop_module_t *module, float 
     if(gui->edit_mode != DT_MASKS_EDIT_FULL) return 0;
     return 1;
   }
+  // add a preview when creating a gradient
+  else if(gui->creation)
+  {
+    dt_control_queue_redraw_center();
+    return 1;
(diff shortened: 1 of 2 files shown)


TurboGit:
Working perfectly! Thanks.

TurboGit:
Working perfectly! Thanks.

TurboGit:
Closing as merged. Github bug?

PR #2740, opened 2019-06-20
conf.h minor rework

jakubfi:
This introduces the following codebase enhancements:
1. **Split conf.h into interface (conf.h) and implementation (conf.c).**
conf.h is included in over 100 source files so this implementation is compiled over and over again. The only explanation I could think of for doing it this way (together with "static inline" definitions) was an attempt to get better performance. I did some quick tests and image edits, and this implementation has no impact on performance on my side. On the other hand, full compilation is now few seconds quicker. And (obviously) changing dt_conf_*() implementation doesn't require recompiling >100 files now.
NOTE: This also required updating header includes for several other files which earlier, by including conf.h, "automagically" got other required includes (which became unnecessary for conf.h in the meantime).

1. **Move config mutex handling into one common place for "set", and one for "get" operations**
Mutex locks/unlocks were scattered around several set/get functions. I moved them down to two low-level functions.

1. **Make critical sections thinner**
Critical sections guarded by mutex locks/unlocks were sometimes unnecessarily thick and included things like string manipulation, calculations and memory management. Now they are a bit thinner.


TurboGit:
Ok, works for me. Thanks.

PR #2741, opened 2019-06-21
accels window enhancements

### LINKED ISSUE #2738: Improve accels window
Actually, the new accels window is a good shortcuts reminder. But I think this window could be improved. The window is a a sort of "pop-out" window inside the main window and that recover the whole window. When this window is activated, there's no possibility to show the main window without quit this reminder. Even when this reminder is activated, it's impossible to show another application. The other application remains on background, the reminder always remains on first foreground. I find that behaviour trying to work on CSS for this reminder, but quite difficult without moving GTK inspector and text editor on another screen. Not the best way to work.

@AlicVB : Making this reminder window as an **independant AND resizable window** would, I think, be better. To be more precise, So being the same type of window as the 2nd preview one on darkroom. 

And *why not add a button to display this reminder, for example with group, help, prefs... buttons?**

AlicVB:
This contain : 
- set default accel to 'H'
- add mouse actions for slideshow and map views
- implement "sticky" accels window : by clicking on the little icon on top-right of the accels window, the window become a normal one, with title, decoration, resizable, ...
This window can be keep around if you want. Window content is updated when changing view, active iop or drawn mask. This fix #2738
- certainly fix compile bug in osx

### DIFF
--- a/data/themes/darktable.css
+++ b/data/themes/darktable.css
@@ -1151,3 +1151,9 @@ cell:selected
    font-family: courier;
    font-weight: bold;
 }
+
+.accels_window_stick
+{
+  min-width: 2em;
+  min-height : 2em;
+}
--- a/src/develop/imageop.c
+++ b/src/develop/imageop.c
@@ -1741,6 +1741,10 @@ void dt_iop_request_focus(dt_iop_module_t *module)
     if(module->gui_focus) module->gui_focus(module, TRUE);
   }
 
+  /* update sticky accels window */
+  if(darktable.view_manager->accels_window.window && darktable.view_manager->accels_window.sticky)
+    dt_view_accels_refresh(darktable.view_manager);
+
   dt_control_change_cursor(GDK_LEFT_PTR);
 }
 
--- a/src/develop/masks/masks.c
+++ b/src/develop/masks/masks.c
@@ -1719,6 +1719,10 @@ void dt_masks_change_form_gui(dt_masks_form_t *newform)
 {
   dt_masks_clear_form_gui(darktable.develop);
   darktable.develop->form_visible = newform;
+
+  /* update sticky accels window */
+  if(darktable.view_manager->accels_window.window && darktable.view_manager->accels_window.sticky)
+    dt_view_accels_refresh(darktable.view_manager);
 }
 
 void dt_masks_reset_form_gui(void)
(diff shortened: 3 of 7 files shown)


Nilvus:
Great works. Just what's needed.

TurboGit:
> implement "sticky" accels window : by clicking on the little icon on top-right

When I click on this icon I get a sticky windows, but on my side full screen no title, not resizable and at this stage I cannot make it disappear except by quitting dt and restarting. Any idea to what could be going wrong?

Nilvus:
@TurboGit : quite strange, I don't have this behaviour. I've tried again just now and when I click on this icon I have the window with title, resizable, etc. I've tried on full screen and not.

To quit this window, when not independent, just press same shortcut that display this accels window.

TurboGit:
Maybe because I'm on a 2 screens setup? I'll try again on a single screen.

Nilvus:
> Maybe because I'm on a 2 screens setup? I'll try again on a single screen.

I don't think so as I also use a 2 screens setup (my laptop screen and a 27" external display).

AlicVB:
@TurboGit : double screen here too... and no problem. It's really strange, here's what I'm doing when clicking on the sticky icon : 
- create a new "top_level" window
- explicitly allow to resize it
- set default size to 0.7*main_window
- copy the scroll_window widget from the "popup" to this new window

A question : once you have clicked on the sticky icon, do you see the new window in the taskbar ? with which name ? Is the sticky icon hidden ?

parafin:
Maybe you need explicitly unfullscreen the new window. It is BTW needed on macOS, so please add the following code after gtk_window_new line:
```
#ifdef GDK_WINDOWING_QUARTZ
      dt_osx_disallow_fullscreen(win);
#endif
```
Same has to be done for accels_window

AlicVB:
@parafin : oops forgot that... I've just added the line for sticky window
btw, I've fixed previously in this PR a stupid copy-paste mistake for osx in this specific line for accels_window popup. I'm even quite surprised it doesn't prevent master to compile on osx

parafin:
Are you sure anyone tried compiling it on osx?:) I did not;)

AlicVB:
So I've added a new line to PR description ;)
and if merging take time for any reasons, I'll push the fix directly in master

TurboGit:
Ok, better now. I think the issue is due to GNOME Shell. When my dt is full screen on the external monitor I can't quite the full screen mode. This issue also plague the accel window. So at the end this is not an issue on this PR, if I start dt not in full screen all is fine. So... Merging! Thanks.

PR #2742, opened 2019-06-22
SSE2 instructions for fdc demosaic.

ILiebhardt:
SSE2 instructions for convolution filters in fdc demosaic. Yields around 20% faster speeds.


ILiebhardt:
Cool, thx @TurboGit. I thought that I once read that SSE2 was a mandatory requirement, but that was a long time ago, so it might be outdated by now. 

One little question: in sharpen.c and colorbalance.c, we're dealing with `process` functions that are rather self-contained, where it makes plenty of sense to have full `process_sse2` alternatives. In demosaic.c, the `process` function calls other functions, where the actual functionality is implemented. 

Furthermore, I only aim at optimising a small part of the code, because that's where it spends most of the time.

Would it be an alternative to wrap the SSE2 parts of the code with `#if defined(__SSE__)`, as it's done for OPENMP throughout the code? (of course with an alternative code-path for non-SSE2)

Sorry for asking newbie questions - this is only my third PR. I'm thinking about trying to limit code duplication.

TurboGit:
> Would it be an alternative to wrap the SSE2 parts of the code with `#if defined(__SSE__)`

I suppose so. In that case you need to identify the part of the code that is SSE and non SSE. Create two routines one for SSE one for standard instruction set and then call the proper one at the proper place depending of the  `#if defined(__SSE__)`. The SSE based version should have the same name than the non SSE one with a _sse suffix (that's the coding standard). 

You can see what is done in src/common/gaussian.c or src/common/dwt.c.

Please let me know if this is clear enough. Thanks.

parafin:
It is possible to disable SSE codepath in runtime (see codepaths/sse2 config option). So I would say adding process_sse2 function is the only correct way.

edgardoh:
@ILiebhardt , you can add a parameter use_sse to high level function and then call it from process() and process_sse(). The the src/iop/retouch.c is done that way.

ILiebhardt:
Thanks for all the hints and feedback, especially to @edgardoh. If we want to be able to disable SSE code-path at runtime, that changes a lot of course. I used retouch.c as a starting point and made a `process_sse2` along the same lines. I decided for two separate functions `xtrans_fdc_interpolate` and `xtrans_fdc_interpolate_sse2`, but for the rest minimised code duplication. 

As a side-note: I even had a version of `xtrans_fdc_interpolate` combining both non-SSE2 and SSE2 code paths in one single function, but this became rather unreadable because of too many `#if defined(__SSE__)` and too many branching points. So for the sake of maintainability, I went for the present compromise, although it has a bit of code duplication.

FreddieWitherden:
I think it is seriously worth asking if this change is 'worth it'.  A 20% performance improvement is not that substantial given the additional maintenance/testing overhead.  As a point of reference for single precision work SSE2 has the potential to enable up to a 4x improvement compared to the 1.2x we are seeing here.  Moreover, looking at the code I can't help but get the feeling that if our goal is just a 1.2x improvement that there are easier ways of achieving this without having to resort to explicit SSE2 intrinsics.

ILiebhardt:
Ultimately, I even (partially) agree to @FreddieWitherden. The reasons why I still think it might be worth it are the following:

1) The 20% is based on the optimisation of the most time-consuming loop alone. I have some further ideas, and think that further optimisation is feasible once an SSE code-path exists.
2) The changes look more dramatic in the DIFF viewer than they actually are. The `process` and `process_sse2` functions are straightforward, the `process_internal` is pretty much identical to the previous `process`. So the only real change in substance is the introduction of the separate SSE code path, by means of `xtrans_fdc_interpolate_ss2`. This is, insofar I can see it, not the only place in darktable where a second function is introduced for the SSE functionality. Although it might appear otherwise in the DIFF viewer, I did not touch anything else.

But OK, I'd like to leave the final decision to the maintainers.

ILiebhardt:
Hi all,
I'll close this PR in the coming days, and I'll put it on the back burner until I find a more elegant way to enable both code paths. I agree to the comments here that this approach isn't really elegant... ;-)

PR #2743, opened 2019-06-22
[needs testing on various CPUs] denoiseprofile: nlmeans speedup

rawfiner:
Changing openmp scheduling from static to dynamic results on a 2x speedup for denoiseprofile's non local means on my main computer (with an intel i7).
However, on some other computers (like my laptop with an intel atom), it does not give any speedup, and even gives a slight slowdown (<10%).
Thus, I don't know if we should integrate this or not, but it is worth testing on various CPU to see if I am not alone to experience huge gain on a CPU, and if so if there is more to win than to loose in general.

### DIFF
--- a/src/iop/denoiseprofile.c
+++ b/src/iop/denoiseprofile.c
@@ -1305,7 +1305,7 @@ static void process_nlmeans(struct dt_iop_module_t *self, dt_dev_pixelpipe_iop_t
       dt_omp_firstprivate(d, ovoid, P, roi_in, roi_out) \
       firstprivate(inited_slide) \
       shared(kj, ki, in, Sa) \
-      schedule(static)
+      schedule(dynamic)
 #endif
       for(int j = 0; j < roi_out->height; j++)
       {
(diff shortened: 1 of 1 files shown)


jakubfi:
Did a quick test on my Intel(R) Core(TM) i5-4570S CPU @ 2.90GHz:
* static: 73,453467 [dev_pixelpipe] took 0,581 secs (2,169 CPU) processed `denoise (profiled)' on CPU, blended on CPU [full]
* dynamic: 79,642152 [dev_pixelpipe] took 0,549 secs (2,146 CPU) processed `denoise (profiled)' on CPU, blended on CPU [full]

So the performance boost is minor on my side. But what is more concerning is this (1:1 crop, same minimal module stack - wb/highlights/denoise/orientation):

static:

dynamic:

rawfiner:
Thanks for testing @jakubfi !
Indeed, if this change can lead to a difference in the algorithm behavior, then we don't want this.

Issue #2603, opened 2019-05-20
GUI - partial copy check box not visible


TurboGit:
**Describe the bug**
The check boxes on the partial copy dialog are not visible (tested on standard & elegant thème).

**To Reproduce**
1. Select an image on lighttable
2. Select copy on development module or ctrl-shift+c

**Expected behavior**
The check boxes should be visible.

**Platform (please complete the following information):**
 - OS: Linux
 - Version current master 2.7

Nilvus:
Oh yes, I will see that.

Nilvus:
I see it again and something strange here. The whole gui is a one box. Anyway, I can make checkbox appear by adding this (after dialog zone) :

```
.dialog-vbox treeview
{
  background-color: @grey_25;
}
```
And changing this line to a more brighter color (like grey_55 here) :
`@define-color field_hover_bg @grey_55;`

But that only work when hovering the line (and only for lines not selected, not for the line selected). I've tried all field lines, all "check" ones by using some flashy colors and can't find how to make checkbox appear with CSS for all items, even when not hovering.

So I'm now quite sure that something missed on GTK code to correct that with CSS. If someone who have skills on GTK could check that. @TurboGit or @aurelienpierre or @AlicVB

TurboGit:
Found a solution, this is now fixed for 3.0 release.

TurboGit:
@Nilvus : thanks for the hint it makes me on the right track, we just missed "check":

`.dialog-vbox treeview check`

Nilvus:
> @Nilvus : thanks for the hint it makes me on the right track, we just missed "check":
> 
> `.dialog-vbox treeview check`

Wow, I was really sure that's could be solved before by GTK tweak. Never thought a second of just the check add... Anyway, good to see that fixed and quite surprise to see it was really simple.

PR #2744, opened 2019-06-22
Css tweaks to improve fields and new accel window

Nilvus:
Two things improved here on the UI :

- First thing is how fields looks. That's also fix partially this issue : #2603 (for more details on why it fixes partially, see the issue ; needs probably to change GTK code to better see checkbox).
- As @AlicVB add a quite useful accels window, this is a proposal to a better UI of this window. Hope you like it and be free to suggest any improvement you see. And as picture is better than words, here it's this window in grey theme for example (quite same except of course global colors in dark themes) :

Capture-20190622183248-1919x1046

### DIFF
--- a/data/themes/darktable-elegant-grey.css
+++ b/data/themes/darktable-elegant-grey.css
@@ -71,11 +71,10 @@
 @define-color button_hover_bg @grey_65;
 
 /* text fields */
-@define-color field_bg @grey_25;
+@define-color field_bg @grey_30;
 @define-color field_fg @grey_85;
-@define-color field_active_bg @grey_40;
+@define-color field_active_bg @grey_50;
 @define-color field_active_fg @grey_90;
-@define-color field_hover_bg @grey_40;
 
 /* Tooltips and contextual helpers */
 @define-color tooltip_bg_color @grey_15;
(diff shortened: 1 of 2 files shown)


AlicVB:
Perfect for my part. Thanks a lot !

TurboGit:
Looks ok, I'm merging.

Can you tell me more about "to change GTK code to better see checkbox" ? I'm ready to help of you have hints about a possible solution.

Nilvus:
> Looks ok, I'm merging.
> 
> Can you tell me more about "to change GTK code to better see checkbox" ? I'm ready to help of you have hints about a possible solution.

I'm not sure how to explain that but on CSS, checkbox seems to be considered as fields and when I try changing fields color (I always test with flashy colors, like red or blue or green..., when I want to be sure to find the good tag). I test all fields tags and the only one where I have flashy color choosed was the hover one (see field_hover tag I change) and the real change is only the "field" hover background by contrast of the main treeview background. All lines of the copy/paste development option are a single treeview box. See this screens captures beside to understand what GTK debugger see (the header and two buttons tags inside the treeview one are from the header line with "include" and "item" text.

I don't how treeview boxes works with GTK but in state, it seems not possible to change checkbox with CSS except with hover tag.

Hope it will help you here.

capture
Capture-20190624230609-824x538

PR #2747, opened 2019-06-24
tonecurve: properly release the profile in OpenCL implementation.

TurboGit:


TurboGit:
@edgardoh : can you validate this? It seems an obvious missing call but we never know. I have been tracking an OpenCL issue since some time... Maybe this is it :)

TurboGit:
After long time editing no more issue on my side with OpenCL being deactivated, so will merge.

PR #2751, opened 2019-06-26
accels window : reduce number of refresh

### LINKED ISSUE #2749: Sticky accels window slows down image switch in darkroom
**Describe the bug**
Having invoked the new accels window and made it sticky switching between images in darkroom using the film strip is much slower than when the accels window has not been invoked.

**To Reproduce**
see above

**Expected behavior**
Similar switching speed with and without accels window

**Screenshots**

 - OS:  both ubuntu 18.04 and windows 7
 - Version latest master

**Additional context**

AlicVB:
This avoid repetitive refresh of the window when in sticky mode and in darkroom view.
this fix #2749

### DIFF
--- a/src/develop/masks/masks.c
+++ b/src/develop/masks/masks.c
@@ -1724,11 +1724,13 @@ void dt_masks_clear_form_gui(dt_develop_t *dev)
 
 void dt_masks_change_form_gui(dt_masks_form_t *newform)
 {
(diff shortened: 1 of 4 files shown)


TurboGit:
Merged and manually resolving conflicts. Can you double check all is correct? Thanks.

AlicVB:
@TurboGit : all seems correct. Thanks for the review

Issue #2565, opened 2019-05-12
[Culling] add row-by-row srcolling, eg. shift+scroll


AxelG-DE:
**Is your feature request related to a problem? Please describe.**
This Feature Request is acc. to the discussions in the (very nice, btw) PR #2539 from @AlicVB 

I would like to use culling in two scenarios:
- before editing I select 1* and 2* prior edit, only 2* will get edited
- after editing they are 3* and now I select 4* and 5*. That is the 2nd attempt, using culling

Current git master's culling mode, once it is used in >=4 pics view, the scrolling or better to say, what you see/experience during scrolling, can be very confusing, when you have different aspect ratio in your selection (3:2, 2:3, 2:1, 1:1) after editing.

Depending on the aspect ratio, sometimes the 
- **first row** shows picture n°1,2 and 
- **second row** n°3;4 or it is

next moment/scroll
- **first row** shows maybe n°3,4,5, and
- **second row** n°6

next moment/scroll
- yet again different

So it is really hard (maybe only for me) to have a clear impression, what you are looking at (it is a bit better, when you also have the filmstrip in mind)

**Describe the solution you'd like**
I would like to scroll row by row, when pressing _Shift_ during scrolling

So, Shift-Scroll would than do:
- n°1;2 out of sight
- **first row** n°3,4
- **second row** n°5,6 _and so forth_

Here it is getting tricky, because;
- either it forces to omit those one-pic in a line or
- even other sortings, e.g. in 6-pic view (4+2) will be difficult or 
- further tricky with odd numbers of view, like 5-pic, 7-pic

**Describe alternatives you've considered**
As I have described [here on #2539](https://github.com/darktable-org/darktable/pull/2539#issuecomment-493652698), I would be already satisfied, if the number of pics per row would not change while scrolling. It should always stay on 2 by 2 (if zoom=4). Currently it sometimes jump to 3pics in first row and 1 pic in 2nd row. That finally makes the left most pic of first row jump to middle pic in first row. Very confusing in my opinion

**Additional context**
I know, my dreams may jeopardize or make things extremely complex, when now the exposure and culling might be on the way to be merged/fusioned

AxelG-DE:
@AlicVB 
So for the sort order I have ammended her on "alternatives". Is that o.k., or you would need that really in a separate issue?

AlicVB:
no pb.
But I'm not convinced we should change how images sizes and positions are computed... Let's wait for other opinions...
[...]

PR #2752, opened 2019-06-26
Aspect ratio fix

AlicVB:
this fix some case where aspect ratio is not computed : 
- when switching to another image in darkroom view
- when deleting history stack in lighttable

this also improve a little ratio computation by reusing preview pipe if available in darkroom
this as been discussed in #2565 (but doesn't fix the main issue)


TurboGit:
Sound perfect to me! Thanks.

PR #2755, opened 2019-06-28
Fixing color matrix for non-D65 CalibrationIlluminant dng files.

jenshannoschwalm:
The obvious problem:
Some DNG files show strangely colored and saturated content; also the color temperature is much higher than expected, often above 25k

Whats wrong?
When including a file into dt  the CalibrationIlluminant and ColorMatrix (1and2) are tested and read in.
If one of both possible Illuminants is 21=D65 everything is fine and the matrix is used. If no Illuminant=21 has been found, the Illuminant/Matrix-1 pair is used. For some cameras (Leica Qs and SL for example) this is StandardLightA --> this leads to the strange effects.

How to fix?
1) If no D65 matrix is found it is better to use the Illuminant/Matrix2, expected to be closer to D65 so less wrong-coloring.

2) We multiply a chromatic Adaption matrix to the found ColorMatrix2, they are taken from http://www.brucelindbloom.com. This gives expected colors.

3) If we don't find an adaptable matrix we should *not* set the img->d65_color_matrix, the colorin module will later show an error message and use a fallback matrix.

4) Possibly there is tested&better matrix available in dt_dcraw_adobe_coeff

Please discuss, comment or merge :-)
Hanno Schwalm


jenshannoschwalm:
BTW i checked this with a large nr of files from raw.pixls ...

src/external/adobe_coeff.c
```diff
@@ -273,8 +273,8 @@ static void dt_dcraw_adobe_coeff(const char *name, float cam_xyz[1][12])
     { "Leaf Credo 60", { 8035,435,-962,-6001,13872,2320,-1159,3065,5434 } },
     { "Leaf Credo 80", { 8035,435,-962,-6001,13872,2320,-1159,3065,5434 } },
     { "Leaf", { 8236,1746,-1314,-8251,15953,2428,-3673,5786,5771 } },
-    { "Leica M9 monochrom", { 6260,-1019,-470,-3730,11450,1930,-1409,2950,6210 } }, /* dummy */
-    { "Leica M9", { 6260,-1019,-470,-3730,11450,1930,-1409,2950,6210 } },
+//    { "Leica M9 monochrom", { 6260,-1019,-470,-3730,11450,1930,-1409,2950,6210 } }, /* dummy */
```
TurboGit: either you add comments saying why this is commented out or you just remove the code.

TurboGit:
Any RAW to share to test this?

src/external/adobe_coeff.c
jenshannoschwalm: ok, remove it. M9 has a correct Matrix and the monochrome files don't need it.

jenshannoschwalm:
There are files for Leica Q,Q2 and SL at raw.pixls. If you want more (Q2) let me know.

TurboGit:
@jenshannoschwalm : ok, but the comment says "some DNG", so I'm not going to download thousand of DNG to find one wrong.  Please help the reviewers by adding proper steps to reproduce a problem with all necessary material otherwise I'll just close this. Think about the time we need to review and check everything already on our free time :)

jenshannoschwalm:
Sorry about not expressing myself in a more precise way.

"Some DNG" files should be understood as:

Some cameras **never** generate DNG files including a D65 (Illuminant=21) matrix, **all** files by these cameras apply a wrong matrix.


I looked through the available exif data of all DNG files at rawpixls, the affected cameras currently are the Leicas Q,Q2 and SL, they have Illuminant1=17 and Illuminant2=23. (The current upstream code selects Illuminant1=17 which is far off D65)

Available example files are
[https://raw.pixls.us/getfile.php/3204/nice/Leica%20-%20Q2%20-%2014bit%20uncompressed%20(3:2).DNG](url)
[https://raw.pixls.us/getfile.php/1797/nice/Leica%20-%20Q%20(Typ%20116)%20-%2014bit%20(3:2).DNG](url)
[https://raw.pixls.us/getfile.php/1951/nice/Leica%20-%20SL%20(Typ%20601)%20-%2014bit%20(3:2).DNG](url)

Hope this is what you rightfully expected, Hanno

src/common/exif.cc
```diff
@@ -502,6 +504,19 @@ static void _find_datetime_taken(Exiv2::ExifData &exifData, Exiv2::ExifData::con
   }
 }
 
+static void mat3mul(float *dst, const float *const m1, const float *const m2)
```
TurboGit: at some point we will want to have a way to share this, we have mat3mul in many places now!

PR #2758, opened 2019-06-28
recompute aspect ratio on styles and copy/paste

AlicVB:
second round. This recompute aspect ratio after applying a style in lighttable and after copy/pasting history stack

### DIFF
--- a/src/common/history.c
+++ b/src/common/history.c
@@ -629,9 +629,10 @@ static int _history_copy_and_paste_on_image_overwrite(int32_t imgid, int32_t des
   sqlite3_step(stmt);
   sqlite3_finalize(stmt);
 
-  DT_DEBUG_SQLITE3_PREPARE_V2(dt_database_get(darktable.db),
-                              "UPDATE main.images SET history_end = 0, iop_order_version = 0 WHERE id = ?1",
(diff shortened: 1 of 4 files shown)


src/common/history.c
```diff
@@ -745,10 +746,8 @@ int dt_history_copy_and_paste_on_image(int32_t imgid, int32_t dest_imgid, gboole
   dt_mipmap_cache_remove(darktable.mipmap_cache, dest_imgid);
   dt_image_reset_final_size(imgid);
 
-  /* update the aspect ratio if the current sorting is based on aspect ratio, otherwise the aspect ratio will be
```
TurboGit: IIRC I did that because the speed of copy/paste a in a large set of images was really slow. Computing the aspect ration need some CPU cycle. So the idea was to do that here only if the current sorting was based on aspect ratio (for obvious reasons) and delay that at some other point otherwise.

So, have you tested the perf?

AlicVB:
@TurboGit : you're of course right for the perfs. I've changed the PR to reset aspect_ratio, instead of recomputing it (except in case of sort order by aspect_ratio, like before)

TurboGit:
Looks good now, thanks!

PR #2759, opened 2019-06-29
Fix potentially invalid structure typecast

d2r2:
Structures `dt_print_format_t` and `dt_slideshow_format_t` explicitly converted to the `dt_imageio_module_data_t` (via pointer typecast) in the code.

At the moment everything is fine, but `dt_print_format_t` and `dt_slideshow_format_t` contains copy of fields of `dt_imageio_module_data_t`, but not nested `dt_imageio_module_data_t` inclusion, thus once the structure of `dt_imageio_module_data_t` will change one day this typecast will become invalid without any compiler warning or error.

### DIFF
--- a/src/libs/print_settings.c
+++ b/src/libs/print_settings.c
@@ -122,10 +122,7 @@ position ()
 
 typedef struct dt_print_format_t
 {
-  int max_width, max_height;
-  int width, height;
-  char style[128];
-  gboolean style_append;
+  dt_imageio_module_data_t head;
   int bpp;
   dt_lib_print_job_t *params;
 } dt_print_format_t;
@@ -153,15 +150,15 @@ static int write_image(dt_imageio_module_data_t *data, const char *filename, con
 {
   dt_print_format_t *d = (dt_print_format_t *)data;
 
-  d->params->buf = (uint16_t *)malloc(d->width * d->height * 3 * (d->bpp == 8?1:2));
+  d->params->buf = (uint16_t *)malloc(d->head.width * d->head.height * 3 * (d->bpp == 8?1:2));
 
   if (d->bpp == 8)
   {
     const uint8_t *in_ptr = (const uint8_t *)in;
     uint8_t *out_ptr = (uint8_t *)d->params->buf;
-    for(int y = 0; y < d->height; y++)
+    for(int y = 0; y < d->head.height; y++)
     {
-      for(int x = 0; x < d->width; x++, in_ptr += 4, out_ptr += 3)
+      for(int x = 0; x < d->head.width; x++, in_ptr += 4, out_ptr += 3)
         memcpy(out_ptr, in_ptr, 3);
     }
   }
@@ -169,9 +166,9 @@ static int write_image(dt_imageio_module_data_t *data, const char *filename, con
   {
     const uint16_t *in_ptr = (const uint16_t *)in;
     uint16_t *out_ptr = (uint16_t *)d->params->buf;
-    for(int y = 0; y < d->height; y++)
+    for(int y = 0; y < d->head.height; y++)
     {
-      for(int x = 0; x < d->width; x++, in_ptr += 4, out_ptr += 3)
+      for(int x = 0; x < d->head.width; x++, in_ptr += 4, out_ptr += 3)
         memcpy(out_ptr, in_ptr, 6);
     }
   }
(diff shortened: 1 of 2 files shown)


TurboGit:
I'm not sure about this one. There is plenty of such "inheritance" in dt code. This is the way it is since its inception.

d2r2:
@TurboGit, I am not sure that in this specific case we talk about storage functionality, and as I explain before both structure explicitly converted to `dt_print_format_t` in the code:
1) `dt_slideshow_format_t` in the line: https://github.com/darktable-org/darktable/blob/f0c9c47ae2c7e4fe44026f0cd2ad2ebc1885a37f/src/views/slideshow.c#L212
2) `dt_print_format_t` in the line: https://github.com/darktable-org/darktable/blob/295bd95a16de35004b145756c0fcaf2a2e6059ea/src/libs/print_settings.c#L247

Honestly I can't understand what is correct in fields copying (obviously it's not safe). Probably I missed some thoughts why it's made so.

TurboGit:
> Honestly I can't understand what is correct in fields copying especially in case of explicit conversion (obviously it's not safe). Probably I missed some thoughts why it's made so.

Are we sure that the nested struct will have the same layout (bit to bit counting the possible alignment adjustements) than the expanded fields. This is important because as you noted we pass the struct to dt_imageio_export_with_flags(). So maybe at least instead of passing the whole struct to dt_imageio_export_with_flags() we want to just pass &dat.head ?

d2r2:
@TurboGit, I agree that we want to pass only `&dat.head` to `dt_imageio_export_with_flags()`.

My main concern, that if somebody someday alter  `dt_imageio_module_data_t` struct (add new field in the middle, for instance), but forget to update accordingly structs `dt_print_format_t`, `dt_slideshow_format_t`, compiler gives no single warning, but corresponding functionality with `dt_imageio_export_with_flags()` will be broken (in runtime). That is why I propose to replace struct fields copy with nested struct, which will automatically update `dt_print_format_t`, `dt_slideshow_format_t` structs fields layout in case of any change in original `dt_imageio_module_data_t`.

TurboGit:
> I agree that we want to pass only `&dat.head` to `dt_imageio_export_with_flags()`

Do you plan to do that? I think that this in print_settings.c and slideshow.c should be easy and safer.

d2r2:
@TurboGit, I'm not sure that we understand each other.
Let me go to the other side and provide more details.

So, I have an intention to modify `dt_imageio_module_data_t` struct from original:
```
typedef struct dt_imageio_module_data_t
{
  int max_width, max_height;
  int width, height;
  char style[128];
  gboolean style_append;
} dt_imageio_module_data_t;
```
to this view (embedding fields `high_quality` and `upscale`):
```
typedef struct dt_imageio_module_data_t
{
  int max_width, max_height;
  gboolean high_quality;
  gboolean upscale;
  int width, height;
  char style[128];
  gboolean style_append;
} dt_imageio_module_data_t;
```

I want to do this, to simplify bunch of functions declaration and corresponding call stack, where `dt_imageio_module_data_t` is involved.
For instance, original function declaration:
```
int store(dt_imageio_module_storage_t *self, dt_imageio_module_data_t *sdata, const int imgid,
           dt_imageio_module_format_t *format, dt_imageio_module_data_t *fdata, const int num, const int total,
          const gboolean high_quality, const gboolean upscale, dt_colorspaces_color_profile_type_t icc_type,
          const gchar *icc_filename, dt_iop_color_intent_t icc_intent)
```
will become more lightweight (moving parameters high_quality and upscale from open stack into the struct parameter):
```
int store(dt_imageio_module_storage_t *self, dt_imageio_module_data_t *sdata, const int imgid,
           dt_imageio_module_format_t *format, dt_imageio_module_data_t *fdata, const int num, const int total,
          dt_colorspaces_color_profile_type_t icc_type, const gchar *icc_filename, dt_iop_color_intent_t icc_intent)
```
So, if I do all these modifications with `dt_imageio_module_data_t` struct corresponding functions declaration and call stack, everything will work fine, except functionality in `darktable/src/views/slideshow.c` and `darktable/src/libs/print_settings.c`. Functionality will be broken there in run-time without any compiler warning, because implementation of `dt_print_format_t` and `dt_slideshow_format_t` are not based on inheritance from `dt_imageio_module_data_t` structure but just copying current fields of `dt_imageio_module_data_t`, and at the same time `dt_print_format_t` and `dt_slideshow_format_t` variables converted explicitly there to `dt_imageio_module_data_t` in the call to `dt_imageio_export_with_flags`.

So, I can cancel this specific proposal, and provide the one I describe here, which cover not only fix for potential issue with `dt_print_format_t` and `dt_slideshow_format_t` structs, but also with `dt_imageio_module_data_t` modification.

If from your point of view all these changes do not make much sense, either you have concerns that this modification is not so straightforward as I imagine, then let me know and I will not waste your and my time on this. Thanks.

TurboGit:
I think I perfectly understand the issue. I agree that the current situation is not very good as all modifications in dt_imageio_module_data_t() should be also done in other part of the code.

But you've never answered my question : what about alignment ? I'm no expert on this area but are we sure that given:

```
typedef struct dt_imageio_module_data_t
{
  int max_width, max_height;
  int width, height;
  char style[128];
  gboolean style_append;
} dt_imageio_module_data_t;

typedef struct xyz_t
{
  dt_imageio_module_data_t data;
  int some_other_var;
} xyz_t;

dt_imageio_module_data_t A; 
xyz_t B;

void call (dt_imageio_module_data_t *v);

call (&A);
call (&B);
```

Then in routine call() v->max_width will be valid in both cases? The C standard says that any padding can be added to ensure that all fields are properly aligned...

That's the only concern I have, if we are sure that it is equivalent then I'm all for merging that.

What do you think?

parafin:
In your example there is definitely no problem, because the struct in question is in the beginning of parent struct. There can't be any padding in the beginning of the struct.

TurboGit:
So we are all good. Thanks @parafin !

PR #2765, opened 2019-07-01
exif: Calculate focus distance for Sony cameras

cryptomilk:
This is required to calculate vignetting correction. This will be supported with exiv2 0.27.2. Tested with 0.27.2 RC2.


TurboGit:
Looks good to me. Thanks.

PR #2767, opened 2019-07-02
Tags - set tag entry with selected tag in the list

phweyland:
It happens the tag name is very long, example:
`Eukaryota|Plantae|Angiosperms|Magnoliids|Nymphaeales|Nymphaeaceae|Nénuphar rose`

When one needs to create a new entry, for example `Nénuphar blanc`, the full tag name must be written.
With this PR, I enter `Nénu` to get the existing list, select one of them, replace `rose` by `blanc` and I'm done with the new tag.

### DIFF
--- a/src/libs/tagging.c
+++ b/src/libs/tagging.c
@@ -296,6 +296,24 @@ static void tag_name_changed(GtkEntry *entry, gpointer user_data)
   set_keyword(self, d);
 }
 
+static void set_selected_entry(GtkTreeView *view, gpointer user_data)
+{
+  dt_lib_module_t *self = (dt_lib_module_t *)user_data;
+  dt_lib_tagging_t *d = (dt_lib_tagging_t *)self->data;
+
+  GtkTreeIter iter;
+  GtkTreeModel *model = NULL;
+  GtkTreeSelection *selection = gtk_tree_view_get_selection(GTK_TREE_VIEW(view));
+  if(!gtk_tree_selection_get_selected(selection, &model, &iter)) return;
+
+  char *tag;
+  gtk_tree_model_get(model, &iter, DT_LIB_TAGGING_COL_TAG, &tag, -1);
+  if (!tag) return;
+  gtk_entry_set_text(d->entry, tag);
+  g_free(tag);
+  set_keyword(self, d);
+}
+
 static void delete_button_clicked(GtkButton *button, gpointer user_data)
 {
   dt_lib_module_t *self = (dt_lib_module_t *)user_data;
@@ -563,6 +581,7 @@ void gui_init(dt_lib_module_t *self)
   gtk_widget_set_tooltip_text(GTK_WIDGET(d->related), _("related tags,\ndoubleclick to attach"));
   dt_gui_add_help_link(GTK_WIDGET(d->related), "tagging.html#tagging_usage");
   g_signal_connect(G_OBJECT(d->related), "row-activated", G_CALLBACK(attach_activated), (gpointer)self);
+  g_signal_connect(G_OBJECT(d->related), "cursor-changed", G_CALLBACK(set_selected_entry), (gpointer)self);
   gtk_container_add(GTK_CONTAINER(w), GTK_WIDGET(d->related));
 
   // attach and delete buttons
--- a/src/libs/tagging.c
+++ b/src/libs/tagging.c
@@ -48,7 +48,7 @@ typedef struct dt_lib_tagging_t
   GtkTreeView *current, *related;
   int imgsel;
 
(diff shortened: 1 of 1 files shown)


phweyland:
Double click interferes with simple selection. Then it is safer to have a dedicated button to copy the selected tag into the entry field.

phweyland:
There are 2 others tags improvements I'm thinking about (other PRs):
- rename tag (keeping same tag id)
- tree view of tags (as in collection)

Would you have any advice or direction for those ?

TurboGit:
@phweyland : I'm not ok with the [copy] button. Why not just use the simple click? This does nothing to day. And click is to select, which seems more appropriate.

phweyland:
> Why not just use the simple click?

That was my first solution. However if there is an event double-click ("row-activated") I haven't found the event simple-click. I've used the event “cursor-changed” which is also triggered with double-click.
Then double-click copies the tag also, which is not desired.
I've also tried “select-cursor-row” event, but it did not trig.

That why I've found safer to use a specific button. If there is a way a get a single click event well discriminated from double-click I can do it, sure.

TurboGit:
The difference between click and 2-click is by checking the event->type. For a double-click:

    if(event->type == GDK_2BUTTON_PRESS)

phweyland:
Single click cannot be ignored when a double click occurs (at least I haven't found the way to do it). Events occur this when double click:
```
GDK_BUTTON_PRESS
GDK_BUTTON_RELEASE
GDK_BUTTON_PRESS
GDK_2BUTTON_PRESS
GDK_BUTTON_RELEASE
```
The other point is that connecting the event "button-press-event" masks the original event "row-activated". Then I've also treated double-click in the same routine.

At the end, I copy the tag into entry on right click (which behaves the same way as the copy button).

TurboGit:
Ok, but what about this:

- GDK_BUTTON_PRESS -> copy tag in selected raw into the entry
- GDK_2BUTTON_PRESS  -> attach the tag to image and clear the entry

So this was a single click on the tag will fill the entry only. When a double click is done with add the tag to the entry, attach to the image and clear the entry. The first copy is not needed but after all this should work and have only the left click used which seems more the expected behavior.

What do you think? Am I missing something?

phweyland:
> When a double click is done with add the tag to the entry, attach to the image and clear the entry.

When I replace right-click by left-click (keeping the same code otherwise) the double-click simply doesn't happen.
The fact of selecting one item in the list refresh the list position. Then when the second click physically happen, the item under the mouse has already gone. The only case this doesn't happen is when the first click doesn't move the list (if I double-click on the first item for example).

TurboGit:
Let's go back one step, there is something I don't understand:

> When one needs to create a new entry, for example Nénuphar blanc, the full tag name must be written.

Agreed!

So your change to select a tag is fine.

Now, the current behavior on my side is that when I enter some letters the tags starting with those letters are displayed in the drop-down list. And when I click on one entry it gets added into the entry and then as you said:

 > to get the existing list, select one of them, replace rose by blanc and I'm done with the new tag.

So to me it seems that the only change needed is the way to select given sub-string.

Am I missing something? Are we looking at two different things?

phweyland:
> when I click on one entry it gets added into the entry

This is not what I have. A single click does not do anything (with or without filtering). So I don't understand the rest of your thought.

phweyland:
Without action:
image

With click:
image

With filter + click:
image

With right click (new behaviour):
image

Then, I can start to modify the entry to create a new tag.

TurboGit:
But don't you have a window opening with all the selected entries (matching the current entry)? It seems that you are selecting on the list and not on the drop-down window.

phweyland:
> It seems that you are selecting on the list and not on the drop-down window.

Correct. But I don't see any drop-down window. I miss something ! :-(
I haven't seen the code for that.
How to make this drop-down window appear ?

TurboGit:
On my side, with the current version I enter a single letter for which lot of tags have a name starting with this letter. I then have a drop-down (or floating, not sure about the correct term) window opening. And BTW, if the list is too long it is opened above the entry and not below (you can probably force this by opening some module above to have the tagging module low in the screen).

phweyland:
The pop-up I have comes with Ctrl-T and I can see the code (_lib_tagging_tag_show()) (dt 2.7).
On window I haven't the pop-up you talk about and I don't see the corresponding code.
Could you help me to find it ?

phweyland:
I see. Probably a standard behaviour of the liststore.
But ... you have to enter the beginning of the tag to get it. And, yes, clicking on an item, copy it to the entry.
I had seen that. 
So yes, it's enough to fulfil the need for the time being.

TurboGit:
Ok, let's go with that. I'm not sure the right-click will be discovered by lot of people though. For sure we need to update the documentation. Can you do that?

Issue #2715, opened 2019-06-14
Temporarily disable all modules after active one in Darkroom


wilecoyote2015:
**Is your feature request related to a problem? Please describe.**
When performing complex image edits, a large amount of modules may be used.
Hence, the preview in Darkroom may become very slow and unresponsive.
Consider the following case:
When retouching a portrait, multiple instances of Equalizers, Liquify and Retouch may be used. 
Those modules exhibit high demands on processing power, so that the whole pipeline may need around 10 seconds to process on an average computer.
If changes to the last of those modules in the processing pipeline are made, editing is still responsive because the module's input from previous modules is cached, so that only the active and  the following modules are reprocessed.
If changes to a module earlier in the stack are made, however, the preview becomes very unresponsive.

Often, I encounter the situation where I need to tune a module that is positioned relatively early in the pipeline. In those situations, there may be no need to see the effect of the modules following in the pipeline, so that I disable all following modules by hand in order to obtain better performance on the preview.

**Describe the solution you'd like**
I believe it would be useful to have a switch in the UI that, only for the darkroom, disables all non-mandatory modules (so except e.g. output profile) in the pipeline after the active module.
At the cost of not seeing the final result, it provides a significant improvement in preview performance for tuning the active module's parameters. 

**Describe alternatives you've considered**
The only currently available option is to disable all modules after the active module by hand and then re-enable them.

pphotography:
I share your thoughts and ideas. Removing 'all' other modules than the current one would lead to some problems:
For example 'White Balance'. You modify the camera's white balance settings. After that you continue with a color module and disable 'all other' modules. What about the value of your color corrections without the final white balance?

Think that the problem is the definition of 'non-mandatory' modules which I did not found so far.

Currently for most of my images I manually disable some modules temporarily. For example, I denoise and sharpen the image and turn off the modules used for that. For final export I reactivate them.

I put images with more complex edits (portraits for example) in a separate folder. I do basic edits like white balance, brightness, colors, sharpen and denoising and export the result as a TIFF in the same folder.
After that I continue work on the TIFF (like retouching) and copy the history stack to the original image in 'append' mode. This keeps the rendering times low and I don't lose my history stack.

Think that there should be a button that lets darktable render the current history stack and use this temporary image as a basis for upcoming edits (similar to the 'snapshot' function).

wilecoyote2015:
Thanks for your input!

[...]

PR #2769, opened 2019-07-02
add option to disable next modules in pipe

edgardoh:
This add an option in the multy-instance menu to skip processing modules after the current one.
Only modules that has a GUI and are not default enabled are skipped.

This is useful only for a couple of scenarios, but when needed is nice to have it.

I could use some help with the GUI:
-I'm using a check box menu, but the check is not displayed. The separator is barely visible, so maybe is a CSS thing.
-There should be a visual clue when the option is active, maybe the module name with a different color, configurable via CSS, but I have no idea how to do that.

### DIFF
--- a/src/develop/imageop.c
+++ b/src/develop/imageop.c
@@ -399,6 +399,7 @@ int dt_iop_load_module_by_so(dt_iop_module_t *module, dt_iop_module_so_t *so, dt
   module->histogram_middle_grey = FALSE;
   module->request_mask_display = DT_DEV_PIXELPIPE_DISPLAY_NONE;
   module->suppress_mask = 0;
+  module->skip_next_modules = FALSE;
   module->enabled = module->default_enabled = 0; // all modules disabled by default.
   g_strlcpy(module->op, so->op, 20);
   module->raster_mask.source.users = g_hash_table_new(NULL, NULL);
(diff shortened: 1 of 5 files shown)


wilecoyote2015:
I believe this is related to #2715

wilecoyote2015:
Are the next modules skipped for display only, or during rendering, too?

Given the former case, I'd propose to place the switch within the bar under the preview, next to the color space and soft-proofing symbols, as this is more a setting referring to the preview than to a specific module. Toggling the switch on would then always skip all modules after the one whose parameters are being changed currently.

edgardoh:
I like the setting on the module better, but let's if this is accepted first.

TurboGit:
I've been trying this. I'm not convinced by the UI and would certainly lean more toward what @wilecoyote2015 is proposing. Now I'm must say that I'm not convinced either by the idea :) What could be a good scenario to work on a picture where the final dev is not viewed? Depending on what is **next** to the current module the display could be really different. So what is the point? At this stage I see this as extra complexity for maybe a very very marginal gain (supposing there is one as you've proposed that). So first, what is a good scenario for this?

edgardoh:
In general, any scenario when you want to work on one of the first modules
on the pipe and the rest don't get in the way. Recently I'm re-visiting
some edits to try the new denoise profiled, so I use this option on the
base curve (or exposure, basic adjustments, etc) to better view the image
and I can play with the denoise, as I don't need all the other modules to
compare the effects. Also, #2715 give some arguments.

As I say on my first post, there's just a couple of scenarios, for me is
the denoise or the retouch when I already have some heavy edits.

I agree that this adds complexity (and worst, on the pipe), even if the
change is not that complex, so if you decide that this is not worth it I'm
OK with that.

Nilvus:
Hi erdgadoh,
I also agree with @TurboGit and so @wilecoyote2015 here. Arguments of @aurelienpierre on #2715 are also to be considered here.

About checkbox not displayed correctly, I'm quite sure it's not about CSS but Gtk way of actually displayed them. See this issue #2603 and comments of this PR #2744  to understand what I've found (and don't know how to correct that as I have no skills on GTK and C coding).

edgardoh:
I don't get it, are you in favor or against this feature?

Nilvus:
> I don't get it, are you in favor or against this feature? 

Not against but not totally in favor of how you propose that. I'm more on what @wilecoyote2015 proposal.

edgardoh:
Well, at this stage we are only discussing the functionality. Once it's accepted we can go deeper into the details. But that's up to @TurboGit .

PR #2771, opened 2019-07-03
make denoise profiled scale-invariant

edgardoh:
This makes denoise profiled (closer to) scale-invariant, still not 100% but much closer than it is now.


edgardoh:
@rawfiner , could you please review this to make sure I didn't forget anything?

rawfiner:
Thanks for this PR
You did not forget anything, this is a good first step toward scale invariance: I completely agree that we should multiply by scale instead of scale\*scale, as variance appear to be reduced quite linearly when downscaling. :-)

TurboGit:
Thanks for the review @rawfiner and thanks @edgardoh for this work. Looks ok to me.

Issue #2724, opened 2019-06-15
[performance] first attempt using a module is sluggish


AxelG-DE:
**Describe the bug**
During [#2514](https://github.com/darktable-org/darktable/issues/2514#issuecomment-497859835) I found this issue...

- My pixle pipe is rather emtpy.
- Open a NEF or ORF file
- I grab & drag the WB slider (WB always on), the response is a bit(!) sluggish, means the reaction is a bit scattering (also the "working..." indicator is less "hectical")
- also the histogram shows that scattering
- relieve the slider and grab&drag it again, now the response is way more fluent including the "working..." indicator is flickering more hectical...
- change opencl_synch_cache from _active module_ to _false_, the issue still persists
- start dt with --disable-opencl --> the issue persists **= not an opencl-issue**
- not only WB also exposure behaves the same
- I tried darktable -d perf, but cannot see a useful ouput
- darktable -d dev shows a little little evidence in a very little higher "cache hit rate so far" after re-touching the slider

**To Reproduce**
see above

**Expected behavior**
Same responsiveness from the beginning, as what I see, after I relieved the slider for once

**Platform (please complete the following information):**
 - OS: [gentoo linux]
 - Version [**2.6.2 and current master**]

**Additional context**
Which further analysis could I do?
@dtorop I still hope, you have a good idea out of our discussions in #2514 (what one I will close now)

@ClaesGitHub I know you don't copy my case. As we have comparable HW, may I ask you, go thru this one here again, and have a brief look at your system?

TNX y'all...

AxelG-DE:
Folks,

After I saw, I can visualize it with CPU-usage (higher on 2nd touch), I spent time analysing "-d perf" again
See attached .xlsx file, where I dragged the WB slider (keep in mind, it works also with exposure)...

- Filter column "E" after [dev_process_preview] and you can see in row 6, 2nd attempt benefits ~20 - 30%
- Filter column "E" after [dev_process_preview] is ~6% benefit (insignificant)
- Filter column "E" after  [dev_pixelpipe] and column L to "base" it is ~34% benefit (hmmmm one by one.....)
- Filter column "E" after  [dev_pixelpipe] and column L to "highlight" it is ~30% benefit (hmmmm one by one.....)
- Filter column "E" after  [dev_pixelpipe] and column L to "input" it is ~20% benefit (hmmmm one by one.....)

So, I hope this helps a bit....

[...]

PR #2772, opened 2019-07-04
optimize render image

edgardoh:
Right now the darkroom center window (the image) is refreshed only when the full pipe is valid, so when playing with some settings we have to wait until the process finish. To test it, grab any curve and start dragging it, the image is refreshed only when the drag stops.

I have added an image buffer for the full and preview2 pipes and moved the preview buffer from the navigation, this buffers are updated when the pipe finish the process, so now the image is updated with the last edit result, making the display more fluid.

I also replaced some draw on the entire window with the center and navigation and added a queue res for the preview2 pipe. This probably won't be noticed as performance goes, but everything counts.

### DIFF
--- a/src/control/control.c
+++ b/src/control/control.c
@@ -517,6 +517,11 @@ void dt_control_queue_redraw_center()
   dt_control_signal_raise(darktable.signals, DT_SIGNAL_CONTROL_REDRAW_CENTER);
 }
 
+void dt_control_navigation_redraw()
+{
+  dt_control_signal_raise(darktable.signals, DT_SIGNAL_CONTROL_NAVIGATION_REDRAW);
+}
+
 static gboolean _gtk_widget_queue_draw(gpointer user_data)
 {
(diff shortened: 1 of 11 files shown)


edgardoh:
The duplicate has also a buffer for the preview pipe, but if I use this new buffer I get a random crash, so I kept it and just updated it on the draw. My guess is the dt_view_image_expose() sometimes free() the buffer, but I don't know code enough to mess with it. Maybe someone more familiar with it can take a look so we can avoid having an extra buffer.

TurboGit:
I'm impressed! This makes a huge difference indeed. Thank a lot.

AxelG-DE:
Dear @edgardoh,

thank you very much for being in for performance!!!

When I saw this now after my vacation, I was in big hope, it would also solve #2724 hence I tested right away. Unfortunately for that occasion it did not do anything. Now, as you are "warm up" for this topic :-) would you do me a favour and look into #2724 please. I would be so happy to see a solution for that.

Thanks in advance!
Cheers
 Axel

Issue #2729, opened 2019-06-17
New in-paint module


leclercj:
I’m considering to develop a new module to provide in-paint capability inside darktable
In order to not re-invent the wheel, the idea would be to link darktable with the G’MIC library, which is proposing several in-paint algorithms out of the box
G'MIC in-paint algorithms are already available in GIMP as a plug-in. See possibilities here: https://gimpchat.com/viewtopic.php?f=28&t=13354
I see this as a good complement of the "spot removal" and "retouch" modules.
Your thoughts ? Any suggestion or objection before I invest more time on this ?

Nilvus:
There's actually G'Mic PR waiting to be merged (Work in progress) so maybe waiting to them to be finished before that ?
Anyway, I'm not on to remove objects but for those you want, why not.

edgardoh:
I did try some of GMIC filters when working on the healing and they all had the same problem to me, for large areas they were just too slow for a raw developer. Same happens with the current heal, so that's not the reason I choose it.
For a proof of concept it should be easy to take my PR and replace some of the retouch algorithms with a GMIC inpaint filter. It will be nice to have an in-paint in dt, if something can be done I think it will be welcome.

TurboGit:
I was about to ask the a question in the sense of edgardoh.'s message.. What is the speed of such algorithm? If not well under the second it won't fit the pixelpipe I fear.

TurboGit:
And I also agree that this would be a nice addition to dt.

pphotography:
To remove objects I use multiple masks in the retouch module and remove piece after piece of that object until it's gone.

A dedicated object removal tool would make work much easier. Draw a mask and dt fills the area with content from around the mask area.

But how much processing time does it take? I know some GMIC filters that take several seconds to apply. On a pixel-based program like Gimp this is not an issue because the calculation is done once. 
But on a RAW developer the calculation is executed again with every change you make. Don't know if it's just that specific GMIC filter or if content aware filtering takes much time in other algorithms too.

Possibilities can be extended to content-aware image resizing, but what about the processing times?
https://github.com/chaitanya100100/Content-Aware-Image-Resizing

wilecoyote2015:
I guess the algorithm might also be useful for filling the borders resulting from image rotation or perspective correction!

leclercj:
Thanks for your feedback

I agree that performances will be critical. Need to select a fast algorithm, even if not the best in term of result.
[...]

PR #2777, opened 2019-07-07
[WIP] First basic inpaint module with G'MIC for trial

leclercj:
This PR is a first attempt to answer https://github.com/darktable-org/darktable/issues/2729

It proposes 5 different in paint algorithms, all G'MIC based, very close to what's proposed in G'MIC filters for GIMP. Same parameters, same ranges.
1) G'MIC Holes
2) G'MIC Morphological
3) G'MIC Multi-Scale
4) G'MIC Patch-Based
5) G'MIC Transport Diffusion

Algorithms 2) to 5) are requesting a mask to specify the in-painting zone
In this basic version, the in-paint module is using a pre-defined colour as mask. ('red' is used by default, as in GIMP, but few other colours are proposed if needed)
For this purpose, this PR is also providing a new "mask" module, used to fill part of the picture with the pre-defined colour, using the usual darktable 'mask' capabilities.

G'MIC installation is a pre-requisite for compilation & execution.
You can refer to https://gmic.eu/download.shtml

Here is an example of use

Intent is to remove the string and the shadow on the following picture: 
Capture d’écran du 2019-07-07 22-42-16

First step is to define the inpainting mask thanks to the "mask" module
Capture d’écran du 2019-07-07 22-45-41

Then usage of the in-painting module:
Capture d’écran du 2019-07-07 22-43-48

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -412,6 +412,8 @@ if(USE_GRAPHICSMAGICK)
   endif(GraphicsMagick_FOUND)
 endif(USE_GRAPHICSMAGICK)
 
+list(APPEND LIBS cgmic)
+
 #
 # Add HAVE_xxx defines used by darktable
 #
--- a/src/common/iop_order.c
+++ b/src/common/iop_order.c
@@ -58,6 +58,8 @@ static int _ioppr_legacy_iop_order_step(GList **_iop_order_list, GList *history_
     _ioppr_insert_iop_after(_iop_order_list, history_list, "rgbcurve", "levels", dont_move);
     _ioppr_insert_iop_after(_iop_order_list, history_list, "lut3d", "grain", dont_move);
     _ioppr_insert_iop_before(_iop_order_list, history_list, "rgblevels", "rgbcurve", dont_move);
+    _ioppr_insert_iop_after(_iop_order_list, history_list, "mask", "colorin", dont_move);
+    _ioppr_insert_iop_after(_iop_order_list, history_list, "inpaint", "mask", dont_move);
 
     new_version = 2;
   }
--- a/src/iop/CMakeLists.txt
+++ b/src/iop/CMakeLists.txt
@@ -140,6 +140,8 @@ add_iop(hazeremoval "hazeremoval.c")
 add_iop(filmic "filmic.c" DEFAULT_VISIBLE)
 add_iop(mask_manager "mask_manager.c")
 add_iop(lut3d "lut3d.c")
+add_iop(mask "mask.c")
+add_iop(inpaint "inpaint.c")
 
 if(RSVG2_FOUND)
   add_iop(watermark "watermark.c")
--- a/src/iop/inpaint.c
+++ b/src/iop/inpaint.c
@@ -1,0 +1,1101 @@
+/*
+    This file is part of darktable,
+    copyright (c) 2019 Jacques Le Clerc
+
+    darktable is free software: you can redistribute it and/or modify
+    it under the terms of the GNU General Public License as published by
+    the Free Software Foundation, either version 3 of the License, or
+    (at your option) any later version.
+
+    darktable is distributed in the hope that it will be useful,
+    but WITHOUT ANY WARRANTY; without even the implied warranty of
+    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+    GNU General Public License for more details.
+
+    You should have received a copy of the GNU General Public License
+    along with darktable.  If not, see <http://www.gnu.org/licenses/>.
+*/
+
+
+#ifdef HAVE_CONFIG_H
+#include "config.h"
+#endif
+#include "bauhaus/bauhaus.h"
+#include "develop/imageop.h"
+#include "dtgtk/button.h"
+#include "dtgtk/resetlabel.h"
+#include "gui/gtk.h"
+#include "iop/iop_api.h"
+
+/* We use G'MIC library in this module */
+#include "gmic_libc.h"
+
+#include <gtk/gtk.h>
+#include <stdlib.h>
+
+#define min(a, b) ((a) < (b)) ? (a) : (b)
+#define max(a, b) ((a) > (b)) ? (a) : (b)
+
+// this is the version of the modules parameters,
+// and includes version information about compile-time dt
+DT_MODULE_INTROSPECTION(1, dt_iop_inpaint_params_t)
+
+
+typedef enum dt_iop_inpaint_algo_t
+{
+  DT_IOP_INPAINT_GMIC_HOLES = 0,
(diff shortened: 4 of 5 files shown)


houz:
I don't think we should add G'MIC as a dependency to darktable.

moy:
Could this be made a tool within the retouch module?

I understand that the code behind it is different, but from the user's point of view, the use-cases for in-painting are very close to the ones of the "heal" tool in retouch. Essentially, it's a heal tool magic enough not to need a source patch.

One advantage of making it a tool within retouch is that one would get the split-frequency editing of retouch for free (or almost).

SilvioGrosso:
Hello everyone,

Sorry to intrude but I suppose it is perhaps worthwhile to add this information in case  you are not aware of it. In short, these past days, it looks like Øyvind Kolås has been adding this same function on Gegl [1]

Maybe there is some code which can be shared between these two different implamentations?
As for me, I am not a developer. 
I am a Gimp user and I have used quite a lot the G'MIC filters regarding the inpainting stuff in the past. 
I am glad you are trying to add them into Darktable :-)

BTW, I even own a copy of Affinity Photo where this feature is also available. Compared to G'MIC is much more faster.

[1] https://gitlab.gnome.org/GNOME/gegl/commits/master

TurboGit:
@houz : what's your proposal? should we copy the gmic sources in dt?

leclercj:
As next steps I see two different dimensions to develop :

1) Add source code of one or several **inpainting algorithm directly inside darktable**.
This to avoid external dependency as much as possible, to use native dt image format and avoid needs for format conversions and color space changes. This also to be in position to develop a GPU version if possible.
I won’t recommend to copy the full G’MIC source code in dt. G’MIC is using its own scripting language, has a slightly different image format, etc.. Also, from maintenance perspective, this does not make sense.
Getting inspired by G’MIC inpainting source code could be an option. 
We also have several different proposals in open-source that I’d like also to explore:
- The Fast Image Inpainting Based on Coherence Transport I proposed in PR https://github.com/darktable-org/darktable/issues/2729 : https://github.com/maerztom/inpaintBCT/tree/master/inpaint_code
- The PatchMatch algorith proposed by @geneing : https://github.com/KDE/krita/blob/master/plugins/tools/tool_smart_patch/kis_inpaint.cpp
Others may surface later

2) Better **inpainting mask management, in a darktable way**
The proposed “mask” module in this PR to fill inpainting area in Red is a quick and dirty shortcut, inspired from GIMP. Good for a trial, but not aligned with usual way of working in darktable.
Inpainting mask needs some cares, as to be efficient, inpainting is requiring a mask fitting as close as possible the targeted zone. The smallest the mask is, the shortest the processing time and the best the result will be.
Ideally, I’d be happy to propose something close to the existing combined drawn and parametric mask capability from blending. Any possibility to use the newly introduced “raster mask” collection inside an iop module ?
Adding the inpainting algorithm in the existing “retouch” module as proposed by @moy is also an interesting idea.
My concern here is that “retouch” is already a quite complex module. I’m afraid that adding further options will make it overcomplicated
Also, I do not see the current shapes as proposed by the retouch module fully adapted to inpainting. For inpaint, we need only destination mask/shape, not source+destination, and I’d like to have parametric shapes, not just drawn shapes.

TurboGit:
>     1. Add source code of one or several **inpainting algorithm directly inside darktable**.

I'm not fond of that as it brings extra maintenance. If the GMIC guys have provided a library to link against it seems that the better path is to use it as you've done in this PR. I'd like to see @houz arguments for the "no".

> For inpaint, we need only destination mask/shape, not source+destination, and I’d like to have parametric shapes, not just drawn shapes.

That's the case for the blur tool. So at the end it may just be a matter of adding a new icon for inpaint in the current retouch module. Not bad from a user point of view :)

TurboGit:
BTW, I find this work quite exiting ! Thanks a lot for working on that.

edgardoh:
Some thoughts on this, in no particular order:

Adding this to the retouch (or spot removal) is tricky, both use an array for the parameters, so any new parameter added will be duplicated by the size of the array. Right now this new module has a lot of parameters.

To call GMIC there's probably no need to colorspace transform, this algorithms should work the same on any rgb colorspace.

I know this is a WIP, but for a more accurate performance testing the modify_roi_in should be added, if it takes 10 minutes to process when zooming in then probably the algo won't be a good fit for darktable.

Another algo to try is the resynthesizer, is very easy to port, and when I try it, it seem faster than some others.
https://github.com/bootchk/resynthesizer

aurelienpierre:
> Copying the sources are probably not a good idea for maintenance.

But it is, however, for performance. I'm not familiar with Gmic sourcecode, but since all buffers in dt are now 64-bytes aligned, we can enable autovectorization free boost by using the right OpenMP pragmas in loops.

Libs suck. See how LittleCMS2 drains power to perform 3 dots products (aka RGB <-> Lab conversions)  in 250 ms or so… Filmic does 2 RGB <-> Lab conversions, plus log and tonecurve in about 15 ms on CPU, and 7 ms on GPU. Optimizing the code yourself is really worth it.

TurboGit:
@aurelienpierre : for performance, maybe but that's not a good solution at long term. The code could be optimized in GMIC and you'll have to redo again and again the tweak in dt code each time you import the code. So copying, why not but all optimization should be proposed upstream for inclusion into GMIC code to ensure the next manual sync in dt will come for free (or at least not cost lot of our energy). Not doing that means that at some point we'll have to maintain all the code ourself :(

aurelienpierre:
@TurboGit it really depends on the algorithmic complexity of the code. I find myself wasting less time recoding my own algos from scratch with proper vector logic, than trying to squeeze existing libs and optimize them to work with SIMD. Libs are a burden one way or the other. And anyway, we will have to do our own OpenCL kernels anyway.

TurboGit:
I understand, for OpenCL we anyway need to control the algorithm to ensure the very same code is done by all path (standard, SSE, and OpenCL). That's a strong argument for copying the code.

spaceChRiS:
I just want to highlight that inpainting would be extremely useful for dealing with scans of negatives and slides that carry information from dust and scratch detection (infrared scan) in their alpha channel. Possibility of such an approach directly in darktable was discussed earlier but most of the necessary ingredients have not been available and therefore this was not going anywhere. However, with the availability of inpainting, given the pipe's alpha channel could transport the images alpha channel, it may not be too difficult to implement scratch removal in darktable.

With the help of the community, I implemented a rough prototype of the method in G'MIC, you can find my (and the communitie's) findings here: https://discuss.pixls.us/t/scanned-image-scratch-removal-with-ice/2350.

In my workflow, scratch removal is done early, since I have to bring the corrected picture as an extra file into darktable, but I do not see any drawbacks in doing this step early. Therefore, I could imagine this as an addition to the invert module.

This is just to show a possible application for your effort besides hand drawn masks.

leclercj:
Closing, as I've no time to work on it anymore

PR #2778, opened 2019-07-08
remove auto tags

edgardoh:
remove auto tags:

Some operation on the filemanager are slow down by some tags access. On my system, discarding the history to 200 images take a few seconds, if I bypass the tag access it takes a second.
I have added a config option under core options->miscellaneous->add informational tags to the image, enabled by default, that will allow to skip the access to the darktable|* tags (darktable|changed, darktable|exported, ..). This tags are not used internally so its ok to be user's choice.

lock rawspeed read file:

Right now, when thumbs are generated, sometimes darktable stops processing (or it seems so) for a while, then it start again. 
To test it:
- start darktable with -d perf
- set a collection that take several pages
- copy the history from one image and paste it onto the entire collection
- scroll down until all visible thumbs are unprocessed

eventually on the console no process is displayed, and after a while it start again, sometimes it takes longer than other.

The delay happens when rawspeed is reading the raw and there's several reads at the same time, so I added a lock, this way only one read happens.
This is kind of a patch, in the sense that I'm not sure what's going on, in theory we should be able to read different files from multiple threads without any noticeable delay. Maybe the file should be readed in chunks instead of all at once, or maybe there's some lock that I'm not aware of. But for now this makes the thumbs generation faster, and if this is ever fixed removing the lock is easy.

### DIFF
--- a/src/common/image.c
+++ b/src/common/image.c
@@ -1711,6 +1711,8 @@ int32_t dt_image_copy_rename(const int32_t imgid, const int32_t filmid, const gc
 
         // write xmp file
         dt_image_write_sidecar_file(newid);
+
+        dt_collection_update_query(darktable.collection);
       }
 
       g_free(filename);
--- a/src/common/undo.c
+++ b/src/common/undo.c
@@ -17,6 +17,7 @@
 */
 
 #include "common/darktable.h"
+#include "common/collection.h"
 #include "common/undo.h"
 #include <glib.h>    // for GList, gpointer, g_list_first, g_list_prepend
 #include <stdlib.h>  // for NULL, malloc, free
@@ -225,6 +226,8 @@ void dt_undo_do_redo(dt_undo_t *self, uint32_t filter)
     l = g_list_next(l);
   }
   UNLOCK;
+
+  dt_collection_update_query(darktable.collection);
 }
 
 void dt_undo_do_undo(dt_undo_t *self, uint32_t filter)
@@ -305,6 +308,8 @@ void dt_undo_do_undo(dt_undo_t *self, uint32_t filter)
   }
 
   UNLOCK;
+
+  dt_collection_update_query(darktable.collection);
 }
 
 static void _undo_clear_list(GList **list, uint32_t filter)
--- a/src/common/darktable.c
+++ b/src/common/darktable.c
@@ -435,6 +435,7 @@ int dt_init(int argc, char *argv[], const gboolean init_gui, const gboolean load
   dt_pthread_mutex_init(&(darktable.plugin_threadsafe), NULL);
   dt_pthread_mutex_init(&(darktable.capabilities_threadsafe), NULL);
   dt_pthread_mutex_init(&(darktable.exiv2_threadsafe), NULL);
+  dt_pthread_mutex_init(&(darktable.readFile_mutex), NULL);
   darktable.control = (dt_control_t *)calloc(1, sizeof(dt_control_t));
 
   // database
@@ -1130,6 +1131,7 @@ void dt_cleanup()
   dt_pthread_mutex_destroy(&(darktable.plugin_threadsafe));
   dt_pthread_mutex_destroy(&(darktable.capabilities_threadsafe));
   dt_pthread_mutex_destroy(&(darktable.exiv2_threadsafe));
+  dt_pthread_mutex_destroy(&(darktable.readFile_mutex));
 
   dt_exif_cleanup();
 }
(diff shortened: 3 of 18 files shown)


src/common/image.c
```diff
@@ -744,8 +744,11 @@ int32_t dt_image_duplicate_with_version(const int32_t imgid, const int32_t newve
     sqlite3_finalize(stmt);
 
     // make sure that the duplicate doesn't have some magic darktable| tags
-    dt_tag_detach_by_string("darktable|changed", newid);
```
TurboGit: As above, it seems to me that we want to always detach the tags here as it was done.

src/common/history.c
```diff
@@ -97,8 +97,11 @@ void dt_history_delete_on_image_ext(int32_t imgid, gboolean undo)
   dt_image_reset_final_size(imgid);
 
   /* remove darktable|style|* tags */
-  dt_tag_detach_by_string("darktable|style%", imgid);
```
TurboGit: It seems to me that we want to always detach the tags here as it was done.

TurboGit:
Looking at the code it looks like used_tags is not used anyway. The db is created in database.db and updated in tags.c but never queried. To be double checked...

edgardoh:
used_tags is used in _synchronize_tags in database.c, I didn't check internally where the remove tag spend most of the time.

I'm not sure what you don't agree here with, so I can't make my case, but I take that you can duplicate the performance issue. Do you think that we shoudn't have a setting to bypass this feature?

src/common/history.c
edgardoh: The point here is to do not call the detach, that's why I'm not calling the attach. If no tag is attached there's no need to detach it.

TurboGit:
> used_tags is used in _synchronize_tags in database.c, I didn't check internally where the remove tag spend most of the time.
> 
> I'm not sure what you don't agree here with, so I can't make my case, but I take that you can duplicate the performance issue. Do you think that we shoudn't have a setting to bypass this feature?

My point is that maybe we don't need this setting if the speed is coming from a routine that is actually not needed.

In _synchronize_tags() we may probably compute the used tags as this is done once at startup. So if the speed is indeed in mostly due to dt_tag_update_used_tags() I would prefer not having this new setting as not needed after all. So it would be nice to have figures with the call to dt_tag_update_used_tags().

TurboGit:
The used tags is given by (if I'm not mistaken):

> $ select distinct tagid from tagged_images;

And this on my side with a 50k images db is immediate.

edgardoh:
dt_collection_update_query() takes the longest, but is still faster if we bypass the entire thing.

If you don't want this setting maybe dt_collection_update_query() can be called only if tags are used on the querry, but I don't know how to check that as tags can be combined with any other selection criteria.

There's also the DT_SIGNAL_TAG_CHANGED signal, that is not always raised when tags change, maybe dt_collection_update_query() can be replaced by that signal and let the collection handle it.

src/common/history.c
TurboGit: But this is a routine to delete history, what about old edits?

edgardoh: As I see it is user responsability, if someone don't want the old tags when disabling this functionality is easy to delete them on the lighttable.

edgardoh:
I've done some more testing and the only option I see is to have a version of the add/delete tags that don't call dt_collection_update_query(), then, after the main function, dt_collection_update_query() can be triggered.
This seems to be more 'correct', the collection should not be updated for every single image when the entire selection is processed, and this way we have a performance improvement regardless of any setting.

TurboGit:
@edgardoh : your last proposal looks promising indeed!

edgardoh:
@TurboGit , done.

src/common/undo.c
```diff
@@ -225,6 +226,8 @@ void dt_undo_do_redo(dt_undo_t *self, uint32_t filter)
     l = g_list_next(l);
   }
   UNLOCK;
+
+  dt_collection_update_query(darktable.collection);
```
TurboGit: I suppose this could be done only for lighttable undo types (that is no doing that when history or mask is changed), something like that:

if(filter & DT_UNDO_LIGHTTABLE)  dt_collection_update_query(darktable.collection);

src/common/undo.c
```diff
@@ -305,6 +308,8 @@ void dt_undo_do_undo(dt_undo_t *self, uint32_t filter)
   }
 
   UNLOCK;
+
+  dt_collection_update_query(darktable.collection);
```
TurboGit: likewise.

src/common/undo.c
edgardoh: We need this to work on the filmstrip, on discard history, when an image is first edited, etc. Actually any operation that may change the collection.

TurboGit: I see, and indeed I forgot that a collection could be changed by history stack if the aspect ratio is changed. So looks good indeed.

PR #2783, opened 2019-07-12
module order: dither before framing and watermark

jakubfi:
Currently dt applies dithering after framing and watermark, which may lead to things like noisy border, watermark rendered in low bit depth, monochrome, etc. This can be changed by reordering modules ofcourse, but I think the default behavior which user would expect from the software is to apply dithering only to actual image content, and then overlay it with "artificial" content like border and watermark.


edgardoh:
This function must not be modified, if you want to change the iop order you have to do it here:

https://github.com/darktable-org/darktable/blob/master/src/common/iop_order.c#L60

jakubfi:
Oh, I missed that function completely. Thank you @edgardoh, fixed.

TurboGit:
Looks good to me. Thanks.

PR #2785, opened 2019-07-14
basecurve: use new norms api

edgardoh:
This moves the base curve to the new norm api.


TurboGit:
Looks good to me. Thanks.

Issue #2732, opened 2019-06-17
Better support for raws from negative scanners


kubrickfr:
Hello,

It would be great if you could support this scanner's raw files. It already kind of works but the invert module is really not working great with it so I suspect it's due to the lack of colour matrix or something...

I have uploaded a sample DNG.

vonj:
Inverting color film negatives is actually a bit more involved than one can believe at first glance. Have a look at the negfix8. https://github.com/chrishunt/negfix8/blob/master/negfix8

It fails on color negatives if there is too much unexposed area in the film frame. (Too much of the orange film base.)

My workaround is to create a color profile from a crop without the unexposed area, then use that for the complete image.

There are commercial alternatives of course, a new one is:

https://www.negativelabpro.com


A side question - is it against the license of darktable to offer closed source plugins for it? (I assume yes, because of GPLv3)

kubrickfr:
You are right in that I assume that inverting negatives, color or b&w, is relatively easy.
Specifically though, my problem is with b&w negatives, and considering how pink the pictures look in Darktable, I assumed that the lack of colour matric in DT was the issue:
![image](https://user-images.githubusercontent.com/2634188/59965785-1625bd00-9513-11e9-8f1e-293e3234fab3.png)
Then in Invert, if I select the darkest area possible, the result is thus:
![image](https://user-images.githubusercontent.com/2634188/59965807-3a819980-9513-11e9-9162-289c0377df4a.png)
Intrestingly enough, the colour of the film material appears gray:
![image](https://user-images.githubusercontent.com/2634188/59965814-4c633c80-9513-11e9-9c9e-24c1acd8b0ce.png)

kubrickfr:
Maybe it's just the default WB (set in the DNG?) that is F-ed up?
![image](https://user-images.githubusercontent.com/2634188/59966043-b03b3480-9516-11e9-8743-afe08f534c3c.png)
If I set it to 1/1/1 it looks grey...

kubrickfr:
RawTherapee doesn't have the same issues. No funky tint, inverting is as easy as inverting the tone curve (which I'm now doing in dt too, gives better results) and it opens files that dt just refuses to read, like this one:
https://drive.google.com/open?id=11n68m0ZLY23pkKBGof2_l-AT6f95sTJ4

kubrickfr:
Do you want me to open a separate bug for the DNG that dt cannot open? I feel like I'm talking to myself here...

[...]

PR #1621, opened 2018-01-09
Improve monochrome raw handling #11912

kelvie:
This does the following automatically for monochrome raws (namely, the ones
coming out of a Leica Monochrom camera):

1. Disable TCA for lens correction by default, as the corrections sometimes
   introduce colour onto a monochrome image (!)
2. Re-enable normal raw processing for monochrome raws; previously it processed
   them as sraws, which disabled auto-exposure
3. Use the passthrough demosaic method, as demosaicing isn't necessary
4. Turn off the white balance module, which in turn disables an error message in
   prepare_matrices, which is a noop for monochrome raws

PR #2787, opened 2019-07-15
Improving monchrome raw images from cameras and film scanners.

jenshannoschwalm:
**Describing the problems**
1) Monochrome raw files are only detected when they are from "leica" and have "monochrom"
   in the model name.
2) Monochrome raw files show color artefacts in darktable mode.

**How to reproduce?**
When a monochrome file is opened in darktable mode we see a colorising (yellow or purple)
artefact. This is inconsistently visible - open a monochrome file and try to enlarge with
the mouse wheel.
Try one of these examples for out-of-camera files:
[ttps://raw.pixls.us/getfile.php/974/nice/Leica%20-%20M%20Monochrom%20-%2016bit%20(3:2).DNG](url)
[https://raw.pixls.us/getfile.php/1087/nice/Leica%20-%20LEICA%20M%20MONOCHROM%20(Typ%20246)%20-%2012bit%20(3:2).DNG](url)

Not all monochrome files are detected as such as the test so far only uses exif maker and model.
There is the Issue darktable-org#2732 describing the same underlying problem, it can be tested using
[https://raw.pixls.us/getfile.php/3215/nice/Plustek%20-%20OpticFilm%208100%20-%2016bit%20(4:3).dng](url)

**What is the underlying problem?**
We don't test in a proper way and later don't have the information available.

**What has been changed in this commit and why?**

a) We want to know about a detected grayscale image and have that information available.
   As dt_image_flags_t has some free bits i used one, this info is a) kept in the database
   and is b) always available via dt_image_t struct  (src/common/image.h)

b) We need testing for a monochrome file with
   int dt_image_is_monochrome(const dt_image_t *img)
   This is now a simple mask &-ing (src/common/image.c)

c) How can we detect a monochrome raw file?
   This is handled & tested in
   imageio_rawspeed.cc  --> line 360 by dt_imageio_open_rawspeed_sraw
   We check
   cpp = r->getCpp();  to be exactly 1
   in this context this can **only be a monochrome** file.
   So we keep that information in the flags.

d) How do we make sure this information is available and used the next time we load
   the image? (After import)
   When reading a raw file into the cache this is done by
   dt_imageio_open_rawspeed
   The test in line 230 should also branch for monochrome files to
   dt_imageio_open_rawspeed_sraw

Hanno


TurboGit:
Right, the Plustek DNG is now properly rendered. But the invert  module is then disabled for the Plustek image. So probably something missing.

jenshannoschwalm:
Please note that the invert modules had been intentionally disabled for monochrome files before this commit. Try the Leica monochrome examples mentioned above.

Disabling the invert module in it's current state makes sense imo as the value the data are inverted against are from the colorpicker and this leads to very strange effects and is not easy to predict for a user.

I see these possibilities atm
a) Let the invert module do a simple 1-val calculation and leave the module running for monochromes but disabling the picker
b) give a hint at the invert module button to use one of the curve modules instead.

Entropy512:
As to fixes to the invert module UI, that's been on my list of things I've been wanting to take a look at.  The current colorpicker approach is highly inefficient (large number of clicks needed to fine-tune the thresholds)

Moving from a colorpicker to sliders, and only showing a "value" slider when a monochrome image is present are good ideas IMO.  It wasn't entirely clear in the referenced issue that the DNG itself was monochrome, as opposed to a color scan of a B&W negative.

jenshannoschwalm:
Enabling the invert module for monochrome has been pushed. (if this way is accepted there is stuff to be simplified)

it works ok this way, although i can't really tell whether this is a good solution for workflows developing such files because i have not been using such stuff before.

I just started crawling through the dt sources and came across this issue.

TurboGit:
> Please note that the invert modules had been intentionally disabled for monochrome files before this commit. Try the Leica monochrome examples mentioned above.

Right, and I suppose this is correct. I was not clear in my message, the invert module should be enabled for the Plustek files not because it is monochrome but because it is an invert file (scan of a negative). But I'm not sure how to handle that!?!

We need to find out why invert was intentionally disabled for monochrome images.

This has been done here: 15d1bf52998b90444af1a92739349de2b306566a

Without explanation of why. Is it ok for you to contact the author and discuss this with him?

Entropy512:
> Without explanation of why. Is it ok for you to contact the author and discuss this with him?
As a side note, from someone who is trying to come up to speed with the state AND HISTORY of certain darktable modules - going forward, could the development team please enforce that commit messages are more informative?

"without explanation of why" applies to a very large number of commits in the codebase.  For example, any commit that is resolving a bug ticket/issue should reference that issue/ticket at a minimum.  (I've seen numerous commits that can be forward-referenced if you find the issue in the issue tracker, but cannot be back-referenced from the commit by someone who is looking at the code and trying to understand WHY something was done.)

jenshannoschwalm:
Yes i'll discuss it and see what we can/should do.
I also don't have a clue to decide when a monochrome file is a negative film and when not.
There are no dng specific exif tags available.
Let the user set a flag 'this image is a negative image'? Would just pollute the gui probably.

parafin:
FYI github shows a link to original PR if you open the mentioned commit: https://github.com/darktable-org/darktable/pull/1621

Entropy512:
"And same for the other modules that are between rawprepare and demosaic AND require CFA: invert, cacorrect."

It appears that might be the case here - turning it back on after fixing the monochrome detection issue might be causing breakage?

I suspect that currently, there's probably a fundamental issue where if invert is turned on with monochrome input, it will still (if you're not careful) apply different weights to varying pixels dependent on where they are (improperly) assumed to lie within the CFA.  This would cause very strange artifacts later on in the pipeline since demosaicing is turned off.  Which means that the invert module probably needs to be fixed such that there's a single slider when monochrome input is detected.

jenshannoschwalm:
> The current "invert" function appears to be causing a monochrome image to be converted into an RGB image with equal values in the channels

This is not correct. The single-channel to RGB conversion conversion already takes place in imageio_rawspeed.cc lines 389++

> What matters is whether the input is actual RGB color as captured by a camera, or if it's monochrome that was channel-duplicated to RGB.ssumption here would be:

Absolutely true.
And you are totally right that the invert module should test for a monochrome and use a different code path.
Testing for a monochrome is now simple; we have the flag which is correctly set in imageio_rawspeed.

TurboGit:
@parafin : yes, but there is no justification for disabling invert. I undertsand that cacorrect and temperature should be disabled. But invert seems harmless and I don't see why one cannot invert a monochrome picture.

parafin:
See https://github.com/darktable-org/darktable/pull/1621#pullrequestreview-87592567 and source code of invert module - there's a lot of mentions of bayer pattern and colors, and no monochrome special case. It's not spelled-out in the commit message, but after reading all the comments in pull request it's more or less obvious.

Entropy512:
Ideally the color selector of invert would get locked to only shades of white when monochrome input is detected, in this case even if the application of inversion were assuming "bayer", as long as all of the thresholds were identical you'd be OK.  Obviously it's even better to just have a single threshold in that case, that is slightly more parallelizable.

The current colorpicker-based UI is really not ideal in many ways (there's no good way to finetune your thresholds), and making it difficult to hide anything associated with color vs. if the red/green/blue thresholds were sliders - then you could just hide them and display a grey level slider when monochrome inptu is detected.

Of course there are plenty of other challenges with invert for some use cases, some of which are not easy to solve.  Some of this results from the fact that you're effectively processing two captures of an image - the first one being the digitization of the negative, the second one being the actual image recorded by the negative itself, and thus some of the things that are usually only considered to be done once in the processing flow need to be done twice or in very atypical orders.

As an example, ideally for workflows involving digitization with a camera and macro lens, lens correction should be applied before invert.  But right now, lens correction has to happen after demosaic, which is after white balance - but white balance should be after the invert phase.  No easy way of solving this without duplicating a version of white balance that operates later in the pipeline.  (Yes, you can sort-of do it with the levels tools, but it's easier to work with the WB tool to color-correct the post-inversion negative.)

Some of the recent changes to iop ordering rules might make such things easier to solve...  I'm still wrapping my head around those.

kelvie:
As seen in the other PR: https://github.com/darktable-org/darktable/pull/1621 I was asked to also disable the invert module for monochrome raws; I hadn't really questioned why.

The history of the patches I made are in that PR.

src/common/imageio_rawspeed.cc
```diff
@@ -378,6 +378,9 @@ dt_imageio_retval_t dt_imageio_open_rawspeed_sraw(dt_image_t *img, RawImage r, d
   // if buf is NULL, we quit the fct here
   if(!mbuf) return DT_IMAGEIO_OK;
 
+  // We test for monochrome before allocating the mipmap cache 
+  if(cpp == 1) img->flags |= DT_IMAGE_MONOCHROME;
```
kelvie: Have you tested raw loading from a leica monochrom raw? There are samples at: https://raw.pixls.us/ if you search for "monochrom"

jenshannoschwalm: Yes, i tested  both  leica monochrom DNGs i found there.

Aah i see @LebedevRI commenting

> And same for the other modules that are between rawprepare and demosaic AND require CFA: invert, cacorrect.

TurboGit:
> This is not correct. The single-channel to RGB conversion conversion already takes place in imageio_rawspeed.cc lines 389++

Right, so at the end we have an RGB image. That's one more reasons to have the invert module activated for those images. At least I see no possible issues. I would go with this and this will let time to have more field testing.

jenshannoschwalm:
I have one more commit for the invert module to "keep monochrome a monochrome".
Will push tonight, please wait for this.

TurboGit:
@jenshannoschwalm : sure I've marked this as wip and will wait for your go. there is nothing urgent anyway. thanks.

jenshannoschwalm:
@TurboGit , the latest small commit just ensures the values taken from the colorpicker have luminosity on all 3 channels to "keep monochrome a monochrome". I would like you to pull.

I concider Issue #2732 to be closed for now.

Still missing a hopefully much better user interface for the invert module

TurboGit:
This is all good to me! I have been able to get decent results with the diapo scan on the Plustek. The invert module does the job here. I'll merge that, if there is issues (color artifacts) it is probably in other parts of the code and it is not a good idea to hide it by disabling the invert module.

Thanks for all the work on this.

PR #2788, opened 2019-07-16
For hidden slider move show toast with value

dterrahe:
When moving hidden slider via shortcut or mouse wheel, pop up toast to show new value.

### DIFF
--- a/src/bauhaus/bauhaus.c
+++ b/src/bauhaus/bauhaus.c
@@ -2219,6 +2219,19 @@ static void dt_bauhaus_slider_set_normalized(dt_bauhaus_widget_t *w, float pos)
   {
     g_signal_emit_by_name(G_OBJECT(w), "value-changed");
     d->is_changed = 0;
+
+    if(!gtk_widget_is_visible(GTK_WIDGET(w)) && *w->label)
+    {
+      char text[256];
+      const float f = d->min + d->pos * (d->max - d->min);
+      const float fc = d->callback(GTK_WIDGET(w), f, DT_BAUHAUS_GET);
+      snprintf(text, sizeof(text), d->format, fc);
+
+      if(w->module && !strstr(w->module->name(), w->label))
+        dt_control_log(_("%s/%s: %s"), w->module->name(), w->label, text);
+      else
+        dt_control_log(_("%s: %s"), w->label, text);
+    }
   }
 }
 
--- a/src/iop/exposure.c
+++ b/src/iop/exposure.c
@@ -542,7 +542,10 @@ void gui_focus(struct dt_iop_module_t *self, gboolean in)
   // switch off auto exposure when we lose focus (switching images etc)
   dt_iop_exposure_gui_data_t *g = (dt_iop_exposure_gui_data_t *)self->gui_data;
 
+  const int reset = darktable.gui->reset;
+  darktable.gui->reset = 1;
   dt_bauhaus_slider_set(g->autoexpp, 0.01);
+  darktable.gui->reset = reset;
 }
 
 void init(dt_iop_module_t *module)


dterrahe:
When switching to darkroom, this will pop up one spurious toast for the last slider initialised.

dterrahe:
Thanks @TurboGit for having a look. Is your initial feeling that this is something that others might want included? My use case is that I've started to work with a midi controller to adjust parameters and that therefore I'd like to have my image full screen without the right panel. But then I wouldn't be able to see the changes I'm making. Even when the panel is showing, if the turn of a knob activates a previously inactive module, it won't be expanded and I won't be able to see the slider.

I used dt_control_log as a quick&dirty proof of concept but was actually surprised how well it worked and how few changes were necessary (and all in one location). But there is the one toast popping up incorrectly when all sliders are initialised when the image is opened in darktable. This might need adjustments in a few places to get rid of, but I'm completely new to the code, so have no idea where to look, really. Rather than the simple toast, it might be nicer to have an actual slider pop up, but that could be a lot of work/not even possible? And timeout should be much shorter.

Also, I guess, this would need an off switch in preferences (defaulting to off). If you think this looks useful I could try my hand at that, but I expect that to be much more involved (translations and everything) so I'd need to get a lot more guidance on how to not break all kinds of rules).

TurboGit:
The last message is probably also reproducible with a reset of the iop while collapse? This is indeed annoying. We'll have to find a fix for that.

dterrahe:
When I press the reset button on a collapsed iop module in darktable view I do not get a pop-up with the value of the last slider that was re-initialised. Is that what you meant? I'm guessing the resets are enclosed by a darktable.gui->reset = 1 / darktable.gui->reset = 0 pair.

Would this be a possible solution for the spurious pop-up at the switch to darktable view? BTW the pop-up doesn't appear when switching between images in darktable.

While testing this I noticed that pressing reset on a disabled module also enables it. Is that intentional?

TurboGit:
> Would this be a possible solution for the spurious pop-up at the switch to darktable view?

Maybe, worth a try. I was about to suggest to test this...

> While testing this I noticed that pressing reset on a disabled module also enables it. Is that intentional?

Yes, that's always been the behavior.

dterrahe:
In darkroom:enter there is the following (pseudo) code:
darktable.gui->reset = 1;
while(modules) module->gui_init(module);
darktable.gui->reset = 0;

If a similar pair of reset = 1/0 exists in one of the early module:gui_inits, then gui->reset will be set to 0 when it is done and the next module will be called with it still at 0.

As a quick and dirty fix, I moved the reset = 1 code into the loop and that seems to work; I no longer get a toast when switching to darkroom.

But the correct fix should be to use the following code everywhere:
    darktable.gui->reset = 1; // avoid history item
    ...
    darktable.gui->reset = reset;

This is already used in 38 places (and 5 more where the variable is called old_reset).

There are 68 places in the code where reset is simply set to 1 and then 0. I would assume that these are all safe to change to the save&restore variant above and would be happy to make these changes and submit as a separate patch. Would you like me to go ahead with that or do you expect that this approach could run into problems?

TurboGit:
> Would you like me to go ahead with that or do you expect that this approach could run into problems?

No, go ahead I think this is a good solution. Thanks.

dterrahe:
The patch in gui-reset_cleanup that I just sent a pull request for seems to have resolved the extra toast at switch to darkroom issue.

However there is a toast for exposure/clipping threshold popping up whenever the exposure module is collapsed. So maybe there is a gui->reset = 1 missing there somewhere. I haven't extensively tested all other modules, but none of my other standard modules exhibit this behaviour.

dterrahe:
The one spurious toast I got from the exposure module could be traced back to the following code in exposure.c, which hard resets the clipping threshold to 0.01% every time the module gets or loses focus.

I wrapped that in gui->reset = 1/reset brackets which solved that problem and hopefully doesn't introduce new ones.

Now I don't see any unintended toasts, so Works For Me.

dterrahe:
I'm a newby to the whole git/hub thing. I noticed that I did a couple of local merges that added a few commits to this pull request. I've new reset back out of those and force pushed the result. Hopefully that made things better, but if not, I would be happy to nuke my fork, generate a new set of commits and submit a fresh pull request.

TurboGit:
Looks ok to me now. Thanks.

PR #2792, opened 2019-07-18
Added button to refresh image data

### LINKED ISSUE #2764: [FEATURE] Add lighttable button to read image data (exif)
Currently to reread for example the exif data from an image, you have to delete it and then import it again. We should have a button in the lighttable to select all images and read the data again to update the metadata in the library.

I've added support for calculating the focus distance for Sony A7 and A6xxxx cameras with exiv2 0.27.2. So I need to reread all images to get the focus distance that lensfun can calculate the vignetting correction.

sams96:
Fixes #2764

I added a button to the bottom of the image information module as shown below. It refreshes the info for all the selected images, not just the one shown.

I'm not sure on the placement of the button, I think either where I've put it (but I'd have to horizontally fill the module), just as an icon next to the "refresh parameters" button, or in the "selected image[s]" module.

refresh

### DIFF
--- a/src/control/jobs/control_jobs.c
+++ b/src/control/jobs/control_jobs.c
@@ -1275,6 +1275,38 @@ static int32_t dt_control_local_copy_images_job_run(dt_job_t *job)
   return 0;
 }
 
+static int32_t dt_control_refresh_exif_run(dt_job_t *job)
+{
+  int imgid = -1;
+  dt_control_image_enumerator_t *params = (dt_control_image_enumerator_t *)dt_control_job_get_params(job);
+  GList *t = params->index;
+  guint total = g_list_length(t);
+  double fraction = 0;
+  char message[512] = { 0 };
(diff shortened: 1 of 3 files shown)


Nilvus:
If this button is applied on all selected images (and I think it's a good idea), moving this button to "selected images" module on the right panel would be more appropriate. That will just add a line but. In that case, renaming the button to something like "refresh exif infos" would be clear.

sams96:
Ok so I moved the button to the selected images module and made it function more like the other buttons there (so it has the fancy progress bar). The only downside is that the actual text in the image information box doesn't update until you mouse-over the images.

TurboGit:
For the image information to update you probably want to raise the signal DT_SIGNAL_DEVELOP_IMAGE_CHANGED when refresh of exif is done.

sams96:
Yeah that works great thanks

TurboGit:
Looks good to me. Thanks.

cryptomilk:
Thanks @sams96 

Waiting for exiv2 0.27.2 now :-)

Issue #2332, opened 2019-04-04
Update to lensfun 0.3.95


leclercj:
**Is your feature request related to a problem? Please describe.**
lensfun has issued a 0.3.95 version in 2018, which is more and more used in many distributions
To my experience, trying to build, install, and execute darktable with this latest lensfun version will fail. The lastest lensfun version supported by darktable seems 0.3.2

As far as I observed, lensfun 0.3.95 introduces several disruptive changes :

1) Static lfXX::Create() and lfXX:Destroy() methods of various Lensfun classe are now marked as deprecated. C++ new/delete syntax should be used instead.
This is creating several warnings for darktable at build time.
_C:/msys64/home/jlecl001/darktable/src/iop/lens.c:354:3: error: 'lf_modifier_new' is deprecated [-Werror=deprecated-declarations]_

2) the database format is now at version 2
It has impact on darktable Makefile, as lensfun database version number is used in installation paths
_CMake Error at src/cmake_install.cmake:157 (file):
  file INSTALL cannot find
  "C:/msys64/mingw64/bin/../share/lensfun/version_1"._

3) C interface: all lf_lens_interpolate_...() functions now require an additional crop parameter.
Probably the reason why I got an an Access Violation exception trying to use it anyhow.

See http://lensfun.sourceforge.net/changelog/2018/06/29/Release-0.3.95-Changelog/ for more details

**Describe the solution you'd like**

1) Document the need for dt 2.6 to use lensefun <= 0.3.2, with explanation where to find the old lensfun package for major distributions and how to install it

For instance, for Windows / MSYS2:
Check lensfun version with:
$ pacman -Q mingw-w64-x86_64-lensfun
If lensfun version > 0.3.2, then download mingw-w64-x86_64-lensfun-0.3.2-4-any.pkg.tar.x from http://repo.msys2.org/mingw/x86_64/
and rever lensfun to 0.3.2 using :
$ pacman -R mingw-w64-x86_64-lensfun
$ pacman -U mingw-w64-x86_64-lensfun-0.3.2-4-any.pkg.tar.xz

2) Prepare darktable to support lensfun 0.3.95 for next major darktable release
It  would also be nice to keep a backward compatibility with lensfun < 0.3.95

leclercj:
Was already discussed in may 2018 in https://redmine.darktable.org/issues/12200
The need is more stringent now that lensfun 0.3.95 is released and is not a dev-only version anymore

parafin:
[...]

PR #2793, opened 2019-07-18
lens correction: port to C++, add support for lensfun 0.3.95

jakubfi:
This is work in progress. Putting it out for feedback and testing. See also #2332 and #2606.

It has been mentioned in #2606 that moving to C++ API is one way of making dt work with the newest lensfun version. I wanted to explore this option as I don't see any obvious downsides:
1. This is the native lensfun API
1. Porting seems (so far) rather straightforward
1. Code readability and maintainability is not compromised, this is basicaly 1:1 port
1. Performance is not compromised (see below)
1. Parts of dt are in C++ already

It has been done in three steps:
1. Make the old lens.c compile with a C++ compiler (and rename it to lens.cc)
1. Switch lensfun stuff from C to C++ API
1. Add support for lensfun 0.3.95, while keeping support for 0.3.2

This should compile against and work with both lensfun versions (tested on Gentoo Linux with gcc 8.3). So far I've done only minimal functional testing. As for the performance, switching to C++ has no impact, but 0.3.95 is >2x slower:

**C, 0.3.2:**
```
1,434711 [dev_pixelpipe] took 0,003 secs (0,011 CPU) processed `lens correction' on CPU, blended on CPU [thumbnail]
3,746754 [dev_pixelpipe] took 0,099 secs (0,332 CPU) processed `lens correction' on CPU, blended on CPU [full]
3,967192 [dev_pixelpipe] took 0,097 secs (0,350 CPU) processed `lens correction' on CPU, blended on CPU [preview]
```
**C++, 0.3.2:**
```
1,402391 [dev_pixelpipe] took 0,003 secs (0,009 CPU) processed `lens correction' on CPU, blended on CPU [thumbnail]
3,693311 [dev_pixelpipe] took 0,088 secs (0,325 CPU) processed `lens correction' on CPU, blended on CPU [full]
3,924291 [dev_pixelpipe] took 0,105 secs (0,353 CPU) processed `lens correction' on CPU, blended on CPU [preview]
```
**C++, 0.3.95:**
```
1,438606 [dev_pixelpipe] took 0,007 secs (0,024 CPU) processed `lens correction' on CPU, blended on CPU [thumbnail]
5,156225 [dev_pixelpipe] took 0,210 secs (0,805 CPU) processed `lens correction' on CPU, blended on CPU [full]
5,521191 [dev_pixelpipe] took 0,234 secs (0,868 CPU) processed `lens correction' on CPU, blended on CPU [preview]
```


TurboGit:
Nice work! I'll try to allocate some cycle to test this.

jakubfi:
Small update on the 0.3.95 performance: last test was done on a debug build of lensfun, silly me. Release build is a bit faster, but still ~2x slower than 0.3.2:
```
1,314050 [dev_pixelpipe] took 0,006 secs (0,011 CPU) processed `lens correction' on CPU, blended on CPU [thumbnail]
4,005145 [dev_pixelpipe] took 0,177 secs (0,664 CPU) processed `lens correction' on CPU, blended on CPU [full]
4,355618 [dev_pixelpipe] took 0,190 secs (0,720 CPU) processed `lens correction' on CPU, blended on CPU [preview]
```

jakubfi:
At this point all lens correction functionality has been ported to lensfun 0.3.95. While this is a step forward, few things are worth mentioning:
- 0.3.95 is just a development snapshot and I wouldn't even call it an alpha release...
- ...thus 0.3.95 support in dt should be considered an experimental feature.
- lensfun API has already changed since 0.3.95, so this won't work with git version.
- I don't think package maintainers will ever provide 0.3.95 in any distribution, so dt users are on their own if they want to try this.

jakubfi:
I've tried to retest the performance with lf 0.3.95, but after recompiling everything from scratch I cannot reproduce my previous results. Performance seems to be on par with lf 0.3.2:
```
1,323827 [dev_pixelpipe] took 0,003 secs (0,008 CPU) processed `lens correction' on CPU, blended on CPU [thumbnail]
3,977505 [dev_pixelpipe] took 0,091 secs (0,341 CPU) processed `lens correction' on CPU, blended on CPU [full]
4,208551 [dev_pixelpipe] took 0,099 secs (0,362 CPU) processed `lens correction' on CPU, blended on CPU [preview]
```
Looks like either my build was borked or my testing procedure was wrong.

TurboGit:
Looks nice! I'll do some testing and will merge this soon. Thanks.

TurboGit:
Looks ok to me. I've found that the focal is not properly displayed into the combo. This was already the case, maybe you want to have a look?

parafin:
This PR doesn't incorporate the work I've done in #2606 to make new lensfun work on macOS. So macOS app bundle probably won't work now if built with new lensfun. Probably same situation with Windows port.

TurboGit:
@parafin : too bad indeed, this PR was marked WIP and without activity since some time. I didn't know it was ready! Hope you'll be able to rebase the changes (not extensive) into current master. Thanks.

jakubfi:
@TurboGit There are several small things in lens correction module that work unexpectedly in some cases and focal length is indeed one of them. I will take a look at these.
I am also planning to separate manual corrections from automatic corrections and expanding them by adding manual distortion and vignetting correction.

parafin:
Status of my branch is explained in my PR. This PR just removes a portion of existing code (that I ported in #2606) using preprocessor when LF_0395 is defined. This is a breaking change.

jakubfi:
@parafin I have no way of testing it on Windows/macOS, unfortunately. 0.3.95 changelog states that now we have "Well-defined database search locations on OS X and Windows platforms" and I assumed that Load() should now take care of properly loading database from all available locations. Also, HomeDataDir is now deprecated and all magic with HomeDataDir done in 0.3.2 code path has to go away anyway. As I stated before - this work is experimental from the beginning (mainly due to 0.3.95 nature) and needs testing (and, above all, stable lensfun release). In this context I wouldn't call it a breaking change. I will gladly update it to work on all major architectures if bugs arise.

As you are not planning to work on this, could you test it on macOS as-is and report with results? I'll take it from there, if that's OK. I would like to understand the problem first.

TurboGit:
@parafin : you have a knowledge that nobody has, I have never used a Mac. In the code (look at the changes here) there is no comment about something done for MacOS and really needed for whatever reasons. So yes a breakage is possible, I don't have a magic brain! Also this PR is opened since some time now, it is too bad that you've seen that when committed.

I cannot do better, I'm almost the only one to review and comment PR, I know nothing to Windows not MacOS. So unless there is clear comment this can happen.

I hope you understand my position and I hope you'll merge the necessary fix for MacOS based on current master. Thanks.

parafin:
@TurboGit, that code might as well be needed for Linux if we supported re-location on Linux. This PR removes code and have an explicit comment with question, that was left unanswered. I have no idea what you are going to merge, maybe if you give longer notice before you merge, then I would have time to add a comment before you hit the button. But nothing bad happened here, especially since we can always revert, so don't worry so much;)

@jakubfi, I haven't seen that notice about search locations on macOS and Windows, do you know more what they mean by that? I don't have time right now to test anything. The code in question looks for lensfun DB in the same location where darktable is installed and not the default install path of lensfun library. It is needed for code re-location, for example as it's done in case of macOS app bundles.

jakubfi:
First of all, let me explain my thought process with this PR. I think I wasn't clear enough with it from the beginning, which I apologize for.
1. We all want new lensfun for obvious reasons.
1. As a part of community using lensfun, dt wants to help make it better, also by incorporating it in some way in early stages.
1. At the same time, we want a stable dt release, so dt should _not_ be shipped with lensfun 0.3.95 (nor git) to a regular user ever, period.
1. dt wants to be "ready" when next stable lensfun drops.

Now on how to approach this:

1. My assumption is that no sane package maintainer will ship 0.3.95 (marked "alpha") as stable, in any distro.
1. Thus, even if we ship dt code with 0.3.95 support, no "mainline" darktable build will use it. There is no breakage.
1. dt should specifically mark 0.3.95 support as experimental.
1. If a user is brave enough and installs 0.3.95 - great! dt can be built against it and tested - this is what dt wants, this is what lensfun wants.
1. dt fixes bugs. lensfun fixes bugs.

If you think I am too optimistic with my first assumption above, we can #undef LF_0395, I'm OK with that. This means one more step for a curious user, but the option is there.

Also, I just noticed that the condition when testing for lensfun 0.3.95 in my code should be "=", not ">=". :-) Bummer. Will fix it.

jakubfi:
@parafin That search locations note is in the changelog (http://lensfun.sourceforge.net/changelog/2018/06/29/Release-0.3.95-Changelog/) and I don't have anything more than that.

As for the code in question: Aha! I didn't know that a "production" build of dt exists, where it is bundled together with lensfun and installed in a common location, different from the usual system directory structure (if I understand correctly). If that's the case, that would somehow explain all the weird "HomeDataDir" code, thank you. :-) It looks wrong anyway, but now I know where to start. I guess lensfun could (should) also be built to a known directory structure and everything would work fine.

parafin:
@jakubfi, please look at @2606, my PR already fixed usage of HomeDataDir for newer lensfun. The main idea is not that location is different, but that it's not pre-defined, user can move the application to any directory. So just rebuilding lensfun won't solve anything.

MStraeten:
Macports currently installs lensfun @0.3.2_4. I just compiled darktable for osx (darktable-2.7.0+1603~gb0252c89a) and lens correction keeps working

jakubfi:
@parafin I did read the patch in #2606 - it still uses the 0.3.95-deprecated dt_iop_lensfun_db->HomeDataDir. It needs to be replaced with appropriate lfDatabaseLoad(const char *pathname) call. I will take a look into it. At worst I will prepare a non-tested 1:1 port to the new API that compiles cleanly. At best I will find a way to test it on my side and provide a working solution.

jakubfi:
@MStraeten that's good to know, that's what I would expect, thank you! Is there lensfun 0.3.95 available in macports at all? Would you be able and willing to test dt with this version?

MStraeten:
in the port files repository no 0.3.95 is available

parafin:
@jakubfi, you read it wrong. HomeDataDir is used only in old code, that I left as-is, as per comments. Pay attention to #if directives.

jakubfi:
@parafin Yep, I [insert your favourite curse word] did read it wrong. Sorry for the unnecessary noise.

jakubfi:
@todd-prior The current (stable) version of lensfun is 0.3.2, not 0.3.95. I would be against using 0.3.95 (which is an unstable development snapshot) in any mainline build of darktable.

todd-prior:
Thanks sorry I don’t really understand the lensfun portion of DT other than it provides camera specific data for corrections…what are the collective changes made by the code that you have added??  Just briefly no big explanation required if you have the time to comment
-

jakubfi:
@todd-prior These changes will make it easier to use next stable lensfun release when it finally arrives and allows for testing the development snapshot of the library.

todd-prior:
Thanks for the clarification..

PR #2796, opened 2019-07-19
Correctly restore previous state of darktable.gui->reset

dterrahe:
Consistantly switch on darktable.gui->reset by enclosing code that requires it with pairs of :
darktable.gui->reset = 1;
/
darktable.gui->reset = reset;

Remove one instance in darkroom.c where darktable.gui->reset got switched off twice. The current code corresponds to what happened previously though perhaps not to what was intended.

### DIFF
--- a/src/develop/blend_gui.c
+++ b/src/develop/blend_gui.c
@@ -898,6 +898,7 @@ static int _blendop_masks_show_and_edit(GtkWidget *widget, GdkEventButton *event
 
   if(event->button == 1)
   {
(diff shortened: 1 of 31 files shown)


dterrahe:
This patch may cause gui->reset to stay switched on longer than before, in the case of nested calls. Even though this is intentional and should be correct, it may uncover previously hidden bugs. I have not immediately noticed any though.

TurboGit:
Looks good, thanks!

PR #2797, opened 2019-07-20
common/opencl_drivers_blacklist: Only blacklist NEO on Windows

Entropy512:
The issue that caused this driver to be blacklisted across the board was fixed in
pull request #2033.  There have been no reported issues on Linux machines
since then, and on an i5-7200U, the blacklist causes a 4x reduction in
performance for compute intensive pipelines (12 seconds vs. 3 for a flow
including exposure fusion).

TODO:  There is very little information about the Windows failures hinted at in pull 2033 comments,
it may turn out it is the same issue fixed by #2033 and a blacklist on Windows is also not necessary.

### DIFF
--- a/src/common/opencl_drivers_blacklist.h
+++ b/src/common/opencl_drivers_blacklist.h
@@ -27,7 +27,16 @@ static const gchar *bad_opencl_drivers[] =
 
   "beignet",
   "pocl",
+/*
+  Neo was originally blacklisted due to improper cache invalidation, but this has been fixed.
+  During the discussion of that issue in pull request 2033, it was hinted that Neo may be still be
+  problematic on Windows, so keep it blacklisted there for now
+
+  TODO:  Determine if Windows failures were due to the same cache invalidation issue.
+*/
+#if defined _WIN32
   "neo",
+#endif
   NULL
 
   // clang-format on


aurelienpierre:
@rawfiner has tested extensively OpenCL on Neo on Linux. It breaks subtly.

Entropy512:
Examples?  That's not what he said in his comment on the pull request I referenced.

saknopper:
Not to mention that the NEO driver is still under heavy development unlike the abandoned beignet driver. So for an unreleased version (like the master branch basically is) it might be beneficial to more easily test the NEO driver.

parafin:
That's actually more of an argument to not allow it - most users won't have latest build from master branch.

saknopper:
Absolutely, so in the released versions you can add NEO to the blacklist even right before the release is made if it is warranted. While you can remove it from the blacklist while developing (in master branch).

Entropy512:
As to NEO being under heavy development, that may have been the case back when it was first released in early 2018, but it's matured enough to be packaged with various distributions (which implies at least some level of testing and QC) - my experience is that what Ubuntu choose for 19.04 has not provided any issues for me yet.

Ideally there would be finer grained checking of driver versions + hardware platforms, although the need for that is likely to go away over time.  Broadwell support was marked as beta until mid-May, but is now positioned as production.  Did anyone ever see an issue on a Neo driver/hardware combination that was designated as "production" by Intel?

Entropy512:
As an FYI, I'm perfectly fine with blacklisting Atoms.  They seem to be more problematic - lower OCL version supported, and the recent downgrade to "beta" status by Intel only on Atoms.  Beignet got blacklisted due to an Atom after all...

As to how to blacklist Atoms - TBD.  Will require some investigation.

rawfiner:
@Entropy512 the bugs I experienced fall into the category of strange bugs that disappeared magically. The first one was after the driver install, and disappeared after a reboot if I remember correctly. The second one was more complicated and I spent hours debugging it, and finally it disappeared magically after a git rebase (I think it was in early january 2019). Both were for non local means of denoiseprofile, as I was working on improving it.
It was on linux mint sarah, with an intel i7 6700. I don't know the driver version however (I reinstalled a different linux distribution since that time).

Entropy512:
Hmm.  The first issue sounds like a situation where the X session needed a restart (while it seems like, in theory, you can change OCL drivers without restarting the X session, there might be cases in practice where this fails...)

The second issue reeks of a caching issue - most likely as you were changing nlmeans, only related components were getting rebuilt.  git rebase likely forced everything to get rebuilt, which would also have caused the cache to get invalidated.

What sort of odd behaviors were you seeing in nlmeans?  I'll beat on that a bit tonight.

rawfiner:
I don't remember exactly. If I recall correctly, with the first one I was having different results on various parts of the image, and the second one the left border of the image was left unprocessed.

Entropy512:
OK, I didn't get as much done on Monday as I'd hoped, I'll try to beat on it some tonight...  But my focus is going to be finishing up my exposure fusion patch.

Unfortunately my current biggest use/testcase for nlmeans happens to have its own issues stemming from the fact that the official methodology for profiling a camera recommends having images that are out of focus - which is REALLY hard with a 180+ degree fisheye lens on a small sensor camera that does not have a focus adjustment.  I haven't figured out an "ideal" way to profile my Mi Sphere.  (My A7III has a profile, however you've got to go to nosebleed ISOs in order to have noise high enough to be able to test the denoise module properly.  :) )

TurboGit:
@rawfiner : I remember the same thing. Maybe at some point the NEO will work.

I've reviewed my log comment and I must say that it was quite bad :(

   opencl: blacklist Intel NEO.

I should have mentioned why it was blacklisted. Everything that seems obvious at some point become quite obscure the next day!

Entropy512:
NEO is, so far, working quite well for me - I'm unable to reproduce any of the failures described.

However I have found a potential issue with the cache invalidation improvements on darktable's side that need to be fixed:

See the "example" OpenCV naming scheme in https://github.com/intel/compute-runtime/issues/125#issuecomment-455504686 - this has driver versions that are consistent with Intel's release page.

On my machine, instead, the cache directory is:
cached_kernels_for_IntelRGen9HDGraphicsNEO_100

100?  That doesn't look right to me...

At least in the future, probably more questions should be asked before blacklisting an entire driver category:
What driver version?  Is it one of Intel's releases, a distribution-packaged release (which distribution?) or self-compiled (if so, what compilers?  This is relevant since sometimes compiler upgrades can uncover new regressions, transient issues when moving from LLVM 3.9->4.0 appears to be a potential contributing factor to the original beignet blacklist.)
What hardware platform?
Is that driver version/platform combination marked as "Production" on Intel's NEO release page?
Does the Khronos CTS pass on the users' machine?  (This may be easier said than done, I wish the Khronos CTS were easier to run - but if the CTS passes but darktable fails that should be a big red flag of an issue with darktable)
Does removing .cache/darktable fix the issue?  (This would be a big red flag that DT's caching is still broken)

Blacklisting only certain hardware platforms (such as Atoms and maybe Broadwell - see the history of Intel's NEO releases where it seems like these platforms get hit with regressions/CTS failures when no other platform is.) should be considered instead of blacklisting the driver entirely on all hardware, and blacklisting the driver because of only 1-2 bug reports without the information above getting collected should be avoided.  Otherwise, it's potentially an open-source project blacklisting another open-source project because that other project is open-source and a user could break it in ways that are not possible with binary releases (such as aggressive compiler optimization settings).

I'll take a look at OpenCV's caching code vs. DT's, it looks like that probably needs improvement anyway.

TurboGit:
Ok, at this stage let's enable this on Linux indeed. This will give us more field testing which is needed here to have a better idea about the possible issue. We have some months to blacklist it again for next release if needed.

ariznaf:
Is there a fix in windows?

I have a i7 8700 with integrated HD 630. 
There were some problems with opencl being dropped in some driver update.
But with new drivers opencl 2.1 NEO seems to be working. Capture One works OK wit opencl activated.
But darktable says it is blacklisted and does not activate opencl.

Is there anyway to regenerate the cache as was sais in Linux environments?

coffeeandspaceships:
I have done my best to look through old threads on this and figure out what the original problems were and how to test them. Most of the information seems to predate Kaby Lake and quite a lot of it refers to Beignet rather than NEO.

I am happy to help with (re)testing for Windows on an Intel HD 530 if someone here can summarise the current status and what kind of symptoms I should look out for - or indeed if there are some SPECIFIC tests from previous defects. (That seems to be what we're short of here - reading the history of all this for an hour it really looks like a couple of Windows users had a bad time with chipsets that were already vintage when they tested them, and everything has been blacklisted ever since.)

Also, is there any feedback yet from the 10 months or so since commit https://github.com/darktable-org/darktable/commit/f1450d1cccd7e3e993acbe750cfa5e4abdc8823b which re-enabled Neo support for Linux?

ariznaf:
It seems that windows user are minority using darktable.
I can do some testing too if I am pointed to instructions on how to do them and what to look for.
As I said I have integrated HD 630 with 16 GB DR3 system memory.

Intel had issues with opencl in the past (they have removed opencl during some driver upgrades) but seem to have fixed them.

Entropy512:
I no longer use darktable, but your observation that it seems like blacklisting occurs with very little investigation or root cause analysis was my experience.

As to Windows - my concern back then (and it still seems valid) is that Intel only is formally supporting Linux with NEO.  While NEO might be buildable on Windows, formal releases are only for Linux - https://github.com/intel/compute-runtime/releases

As such, on Windows there are not testing artifacts/qualification test results that can be referenced, unlike Linux where the above link provides a "Quality Expectations" list for all supported platforms.

ariznaf:
In Linux it seems that neo has been taken out from blacklist.

But this thread is about windows installations.

NEO is not only buildable on windows: it is being distribiuted officially by Intel in its graphics driver.

I have not made any additional install, I have just installed last version of intel ghraphics driver and opencl 2.1 NEO  driver was installed with it.

Opencl is active and working in Capture One, for example.

It seems there has been few testing of the problems caused by intel opencl driver under windows and no testing at all with the new drivers.

That is why we are offering to do testing to see if it is working correctly now and can be activated on windows too.

coffeeandspaceships:
According to Intel's notes at https://github.com/intel/compute-runtime, Skylake, Kaby Lake, Coffee Lake are fully supported at OpenCL 2.1 AND the alternate API Level Zero is supported. (NEO is the name of the initiative which supports both these APIs.)

Useful summary [on StackOverflow](https://stackoverflow.com/questions/57302414/what-is-difference-between-intel-compute-runtime-intel-opencl-runtime-and-inte) of legacy versions which might help with putting together a decision tree around what to blacklist/whitelist.

coffeeandspaceships:
It was @aurelienpierre (in [this comment](https://github.com/darktable-org/darktable/pull/2033#issuecomment-456941813_) on PR2033) who suggested the defensive approach of re-enabling on Linux first, which seems sensible. 

@aurelienpierre can you comment on any particular tests the project could use to determine whether the OpenCL problems with Windows seen a few years ago have really been eliminated on Coffee Lake and recent Intel Compute Runtime (aka Neo) which formally supports OpenCL 2.1?

Regarding "formal support," the current Windows 10 drivers are built from the same source as the Linux drivers AND they participate in the Microsoft support scheme for drivers and are signed and available in Windows Update, so I read that as fully supported.

Entropy512:
@coffeeandspaceships he briefly mentioned it there, but subsequently he fought hard to just not support OpenCL on Intel at all.  See, for example, his comment that is the second one on this pull request.  You'll get no support from him in this regard.

aurelienpierre:
1. We don't care for support for OpenCL 2.1 since the code forces OpenCL 1.2. We just need reliable stuff that supports 1.2.
2. Intel Neo is currently working on my laptop with Linux + darktable, but I use it only for debug.
3. On Windows 10, last time I checked, Intel Neo made dt crash right at starting.
4. I'm not opposed to Intel OpenCL support on Windows, but then I will redirect complaining users to you, and let you deal with them.
5. Speed-up between C/CPU and OpenCL/Intel GPU is only marginal.

coffeeandspaceships:
Cool thanks. Answering your points in order:
1. Good to know. I'll see if I can find any policy statements from Intel on how long 1.2 will be maintained. But see point 4 anyway. :-)
2. Good, that's more evidence it's relatively stable on Linux now.
3. Good, obvious test case there! (What Intel generation / IGA is your laptop please, assuming that is where you last tried Windows 10?)
4. I'm happy to triage, and contribute to prob resolution, and hell I might even fix things and create PRs if they're within my limited dev talent.
5. I saw quite a lot of discussion around that. I think that was true on versions before Kaby Lake for sure (and beignet confusion did not help.) In order to address this properly, two things:
a. I noticed there is now an explicit "hello world" performance test in DT OpenCL startup, so that's a defensive catch for this case.
b. Testers can provide benchmarks and I can collate them into summary evidence of how far back in Intel history we should whitelist.

aurelienpierre:
Ok so if you are willing to commit to that, then we can re-open the Neo/Windows topic. 

My laptop has Kabylake Xeon E3-1505M v6. I have no idea what Neo version Windows uses, it came pre-installed with Windows 10 workstation edition, I don't even have it listed in the list of software.

coffeeandspaceships:
Brilliant, thank you. I'll re-open the Neo/Windows topic by creating a proper issue for it, which will also help manage the priority against everything else. Will do this evening.

ariznaf:
If you provide me with an executable (or installer) for windows 64 that does not blacklist intel driver, I can do the test you want to see if it works correctly now.

And make some speed testing (just need to be pointed to a document explaining what test are needed).

Darktable complains about opencl 2.1 NEO not being an appropiate driver and being blacklisted.
Does not provide any more info or detail about the exact version of driver.

My processor is an intel i7 8700 wiht HD 630 GPU Coffe Lake (GT2)
OpenCL 2.1 - Intel(R) UHD Graphics 630 - CUs:24@1200MHz
it reports OpenCL 2.1 NEO driver version 27.20.100.8187

There has been problems with openCL. Intel disabled openCL for some time in their driver upgrades.

But now seem to be working OK (at least it works in Capture One).

coffeeandspaceships:
I have raised https://github.com/darktable-org/darktable/issues/5054 to continue discussion and track resolution.

PR #2799, opened 2019-07-20
iop/basecurve: Reconcile code path divergence

Entropy512:
At some point in the past, the fusion basecurve LUT and non-fusion LUT code were
separated.  Comments in past commits hint that the non-fusion path was a
"fast path" to eliminate the multiply operation.

Eventually this "fast path" and the fusion "lookup with multiply" were
separated into their own functions.  The former "fast path" now has
a set of fixes to avoid chromaticity shifts.  As it was a separate path,
the fusion flow did not get these fixes.

Reorganize code so that both the non-fusion and fusion paths use the same
LUT code, with the non-fusion path simply requesting a 1.0 multiplier on
the input.  Also, for CPU paths, reorganize so that the two paths
(preserve colors and non-preserve-colors) each have their own loops,
keeping a conditional outside of the OpenMP processing loops.

Rename the paths without "preserve colors" as "legacy" to imply that they
are primarily still here for backwards compatibility.  (There are artistic
reasons one might choose not to preserve colors, such as "punchy sunset"
effects, but there are likely better ways to achieve this.  If you disable
"preserve colors", HERE BE DRAGONS.)

Also, now permit "preserve colors" to be used with fusion since the
merged code paths now support the requirements of both fusion and
non-fusion workflows for both "preserve colors" and "don't preserve colors"
paths.

Some additional comments were added to describe what's going on for
legacy vs. current

### DIFF
--- a/src/iop/basecurve.c
+++ b/src/iop/basecurve.c
@@ -304,7 +304,7 @@ typedef struct dt_iop_basecurve_global_data_t
 {
   int kernel_basecurve_lut;
   int kernel_basecurve_zero;
-  int kernel_basecurve_ev_lut;
+  int kernel_basecurve_legacy_lut;
   int kernel_basecurve_compute_features;
   int kernel_basecurve_blur_h;
   int kernel_basecurve_blur_v;
@@ -499,6 +499,7 @@ int process_cl_fusion(struct dt_iop_module_t *self, dt_dev_pixelpipe_iop_t *piec
 {
   dt_iop_basecurve_data_t *d = (dt_iop_basecurve_data_t *)piece->data;
   dt_iop_basecurve_global_data_t *gd = (dt_iop_basecurve_global_data_t *)self->global_data;
+  const dt_iop_order_iccprofile_info_t *const work_profile = dt_ioppr_get_pipe_work_profile_info(piece->pipe);
 
   cl_int err = -999;
 
@@ -508,6 +509,11 @@ int process_cl_fusion(struct dt_iop_module_t *self, dt_dev_pixelpipe_iop_t *piec
   const int height = roi_in->height;
   const int rad = MIN(width, (int)ceilf(256 * roi_in->scale / piece->iscale));
 
+  cl_mem dev_profile_info = NULL;
+  cl_mem dev_profile_lut = NULL;
+  dt_colorspaces_iccprofile_info_cl_t *profile_info_cl;
+  cl_float *profile_lut_cl = NULL;
+
   cl_mem *dev_col = calloc(num_levels_max, sizeof(cl_mem));
   cl_mem *dev_comb = calloc(num_levels_max, sizeof(cl_mem));
 
@@ -515,6 +521,13 @@ int process_cl_fusion(struct dt_iop_module_t *self, dt_dev_pixelpipe_iop_t *piec
   cl_mem dev_tmp2 = NULL;
   cl_mem dev_m = NULL;
   cl_mem dev_coeffs = NULL;
+
+  const int use_work_profile = (work_profile == NULL) ? 0 : 1;
+  const int preserve_colors = d->preserve_colors;
+
+  err = dt_ioppr_build_iccprofile_params_cl(work_profile, devid, &profile_info_cl, &profile_lut_cl,
+                                            &dev_profile_info, &dev_profile_lut);
+  if(err != CL_SUCCESS) goto error;
 
   int num_levels = num_levels_max;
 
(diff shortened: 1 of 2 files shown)


src/iop/basecurve.c
```diff
@@ -1213,54 +1313,26 @@ void process_lut(struct dt_iop_module_t *self, dt_dev_pixelpipe_iop_t *piece, co
 {
   const float *const in = (const float *)ivoid;
   float *const out = (float *)ovoid;
-  const int ch = piece->colors;
+  //const int ch = piece->colors; <-- it appears someone was trying to make this handle monochrome data, however the for loops only handled RGBA
```
TurboGit: This looks wrong to me. The number of colors per channel should not be encoded anywhere. IIRC this value is 4 most of the time but we don't want to assume this in the code anywhere.

src/iop/basecurve.c
```diff
@@ -516,6 +522,13 @@ int process_cl_fusion(struct dt_iop_module_t *self, dt_dev_pixelpipe_iop_t *piec
   cl_mem dev_m = NULL;
   cl_mem dev_coeffs = NULL;
 
+  const int use_work_profile = (work_profile == NULL) ? 0 : 1;
+  const int preserve_colors = d->preserve_colors;
+
+  err = dt_ioppr_build_iccprofile_params_cl(work_profile, devid, &profile_info_cl, &profile_lut_cl,
```
edgardoh: You need to call dt_ioppr_free_iccprofile_params_cl() to match this.

edgardoh:
@TurboGit , I can't test this right now, but other than my comment the code looks ok to me.

src/iop/basecurve.c
Entropy512: I think we found a flaw in the existing code.  If you read lower down in the function, you'll see, in the legacy curve function:
`for(int i = 0; i < 3; i++)` - if you read what's in that loop, followed by the `outp[3] = inp[3]`, it's performing per-channel lookup on R, G, and B, then passing through what I assume is intended to be A.

Again, in the new apply_curve, we have:
```
    for(size_t c = 0; c < 3; c++)
    {
      outp[c] = (ratio * inp[c]);
    }
    outp[3] = inp[3];
```

Again, - pretty clearly written for RGBA.

I think you've found an existing bug regarding failure to handle the monochrome case.  After all, "preserve colors" really has no meaning with monochrome data - so fixing those issues probably needs a bunch of additional work (either by adding an additional commit to this pull request, or I promise to work on that next in a subsequent pull request before pushing up my exposure fusion changes).

Although in the discussion of https://github.com/darktable-org/darktable/pull/2787 - it sounds like at some point in the pipeline, monochrome images are converted into RGB (with equal values for each of R, G, and B) - under what conditions would we see a channel count other than 4 at this point in the pipe?  What would the pixel format be in this case?

src/iop/basecurve.c
Entropy512: Thanks!  Good catch, I can easily fix that tonight.

One question:  Since github has the ability to make a rebase/squash/force-repush of a pull request less painful for all reviewers, is that acceptable for this fix?  Doing a rebase/squash has the benefit of keeping the commit history cleaner/more concise for someone trying to follow history down the line.

I've noticed people have tended towards lots of "fix" commits without a squash prior to merge, that sometimes makes the history of the code quite difficult to read - but is that an artifact of when darktable lived elsewhere other than github?

edgardoh: Not really my area, I always try to have one commit per functionality, but is not always the case.

Entropy512: That's generally my goal, but I'm not sure if there's a policy against rebasing/squashing an amended pull request.  Some people really dislike it (even if it does in the long run result in cleaner commits)

Linus hates it, but the email-based kernel development flow pretty much amounts to an alternative to rebases/squashes within git, by resending a patch set as emails instead.

Entropy512:
Just as an FYI, I probably won't be able to get around to addressing the changes until Wednesday evening at this point.  I'll try to push the amended patch up then.

TurboGit:
@Entropy512 : no problem, I'll mark this as wip. ping me when ready. thanks.

src/iop/basecurve.c
Entropy512: I pushed an amended commit that frees those data structures

Entropy512:
@TurboGit OK, I made the changes @edgardoh requested.  To keep the commit history clean, I amended and repushed the commit - hopefully this doesn't break github workflow too much (Most of my past code review experience is with gerrit, rebasing/squashing/amending commits prior to them being accepted and merged was easy in that work flow.)

TurboGit:
Ok, looks good now. Thanks.

PR #2801, opened 2019-07-21
Add exif data to webp files (fixes #2768)

### LINKED ISSUE #2768: EXIF metadata missing from WEBP files
I exported my photos in webp format but all metadata is missing from the exported files. I expect EXIF data to be reserved.

sams96:
This seems to work fine for me but I kind of think it must've not been done for a reason.


TurboGit:
Sounds like an omission. Thanks.

PR #2804, opened 2019-07-22
Fix bug in middle button zoom for second preview window

dterrahe:
Fix a bug where zooming in the second preview window, with a middle mouse button click, would use the size of the image in the main darktable window to calculate the location of the click. Especially when using the 2nd preview on a larger second monitor and with the panels showing in the main window, these different sizes would lead to the zoomed window being centered far away from the point that was middle clicked on.

Turned out to be a simple fix; working correctly now.


TurboGit:
Looks good indeed! Thanks.

PR #2805, opened 2019-07-22
lensfun: fix library version check, fix db loading

jakubfi:
#2793 introduced lensfun library version check used to determine whether LF_0395 should be #defined, but wrong comparison operator was used. LF_0395 should be defined _only_ for 0.3.95, as future releases break the API again.

Additionaly, #2793 missed changes introduced in #2606. Those were added and ported to 0.3.95 C++ API.


TurboGit:
Thanks! Looks good.

PR #2809, opened 2019-07-23
Refresh navigation preview after zoom change

### LINKED ISSUE #2806: Bounding box overlay in preview not removed after cycling back to Fit to Screen using middle button
In darkroom, when cycling through zoom modes 100%/200%/fit to screen, using the middle mouse button, the preview window shows a navigation bounding box with the correct zoomed size. However, this box should disappear when we get back to fit-to-screen. But it doesn't until something else triggers a refresh of the preview window.

1. open image in darkroom
2. middle-click on main image
3. zoom switches to 100%; preview window top left still shows full image, but with bounding box for zoomed area
4. middle-click again
5. zoom now 200%, bounding box shrinks further
6. middle-click again
7. zoom now fit-to-screen; full image visible in center view
8. preview still shows smallest bounding box corresponding to 200% zoom
9. click-drag in center view
10. preview now gets updated and bounding box disappears.

Expected behavior:
Preview window updates as soon as middle mouse button clicked and bounding box immediately disappears when zoom is switched to fit-to-screen.

dterrahe:
Trigger navigation control redraw every time the zoom setting changes, via middle click or shortcut key, to update the bounding box.

Fixes #2806

### DIFF
--- a/src/views/darkroom.c
+++ b/src/views/darkroom.c
@@ -916,7 +916,6 @@ static gboolean zoom_key_accel(GtkAccelGroup *accel_group, GObject *acceleratabl
       dt_control_set_dev_zoom_x(zoom_x);
       dt_control_set_dev_zoom_y(zoom_y);
       dt_control_set_dev_closeup(closeup);
-      dt_dev_invalidate(dev);
       break;
     case 2:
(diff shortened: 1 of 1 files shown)


TurboGit:
Looks ok, thanks!

PR #2811, opened 2019-07-23
Add a tree view for tagging module + tag rename function

phweyland:
This PR brings the following features to tagging module:
- choice between the list view and the tree view (saved in conf)
- choice to get the suggested tags (those based on selected images) or not (saved in conf)
- indicator telling when the tags in the list are used by the selected images
- number of images attached to each tag

list view as before (with suggested tags):
image

list view without suggested tags and added infos:
image

tree view with added infos:
image

rename tag function added:
image

image

For the time being I've let the buttons (delete & rename). I would suggest to remove them but some people may prefer the buttons to right-click. Please give advices.

There is an unsolved issue for which I need direction.
The button sugg / no sugg(estion) always appears at start. However it should be disable for the tree view (gtk_widget_set_visible() correctly set at the end of gui_init).  It seems there is somewhere a gtk_widget_show_all() which overrides this initial setting. Do I miss something ? Is that a known issue ? What could be the solution ?


phweyland:
Nodes which have no corresponding tag are shown in italic
image

If a new tagname is found as existing, renaming is aborted:
image

Nilvus:
Wow, I love that. Thanks Philippe to work on that. That's a really great improvement. I also make nature photos and having a tree is very useful.

I would like to see in the future the possibility to have on collection filter (tag one of course) the possibility to see all images below a tag more easily. Actually if for example I have this tree :

Faune|Oiseaux|Turdidés|Merle noir

I add the tag 'Merle noir' to all photos of of course a 'Merle noir'. If I want to show all Turdidés photos, I have to select it on collection filters tag then add '%'. But that also hide my tag tree on the filter collections. So If after that I want to show another category of birds (Oiseaux for non french people and this example), I have to reset filter or remove '%' then hit enter then select other category and add '%'.
Other possibility is to add the 'Faune|Oiseaux|Turdidés|' to all 'Turdidés' images. But If I want all 'Faune|Oiseaux|', that same problem. And if add all parents to all images, that makes lot of tags and could make the tag list quite big.

I'm not sure it could be related to this PR but adding on filter collections a simple quick option to display all images below on a parent tag without hiding the selected tree (like actually when using '%' character) would be really great.

TurboGit:
I agree with Nilvus, really nice! I'll test that in the coming days.

phweyland:
> I would like to see in the future the possibility to have on collection filter (tag one of course) the possibility to see all images below a tag more easily.

I think this is already done:
- click on a node: images with that tag
- ctrl-click on a parent node displays all images under this hierarchy. 
- shift-clicking on a parent node displays all images in and under this hierarchy.

>  I would also like to see on all parents and children, total images on the level (so by counting all images on children if a parent is selected.

In tagging module the count is the count of images with that tag (and I think that's the right way). 
In collection, tag doesn't work as folders. I agree it would be interesting to have the cumulated number (as folders does). If I can make something work I'll do it.

Nilvus:
> > I would like to see in the future the possibility to have on collection filter (tag one of course) the possibility to see all images below a tag more easily.
> 
> I think this is already done:
> 
>     * click on a node: images with that tag
> 
>     * ctrl-click on a parent node displays all images under this hierarchy.
> 
>     * shift-click on a parent node displays all images in and under this hierarchy.

Thanks, so nothing to add on what I was saying. I have read the manual two times and view some lot of videos but never heard of that shortcuts. So maybe not so well documented. I just tested shortcuts and that just what I was waiting (already here). Thanks for sharing them !

> In tagging module the count is the count of images with that tag (and I think that's the right way).
> In collection, tag doesn't work as folders. I agree it would be interesting to have the cumulated number (as folders does). If I can make something work I'll do it.

Here I was talking about filter collections module. About tagging module, I agree with you, how it works is the right way. That will be great if you could improve filter collection on that part and of course, maybe displaying number related on click+shortcut (or not) used.

phweyland:
> it would be interesting to have the cumulated number (as folders does)

@Nilvus  I had a look about this but that cannot work as folders. In folders every image is counted once. In tags, every image can have several tags. Then when the counts are cumulated one ends up with far more images than reality.
However, even without cumulating, the current version doesn't display the tag node count if the tag is a not a leave. I'll make a PR on this.

src/libs/tagging.c
```diff
@@ -384,12 +536,230 @@ static void delete_button_clicked(GtkButton *button, gpointer user_data)
...
+  char *subtag = g_strrstr(tagname, "|");
+
+  gint tag_count;
+  gint img_count;
+  dt_tag_count_tags_images(tagname, &tag_count, &img_count);
+  if (tag_count == 0) return;
+
+  GtkWidget *win = dt_ui_main_window(darktable.gui->ui);
+  GtkWidget *dialog = gtk_dialog_new_with_buttons("Rename tag?", GTK_WINDOW(win), GTK_DIALOG_DESTROY_WITH_PARENT,
+                                       _("_Rename"), GTK_RESPONSE_YES, _("_Cancel"), GTK_RESPONSE_NONE, NULL);
```
TurboGit: All strings (Rename , Cancel) should be in lower-case.

src/libs/tagging.c
```diff
@@ -384,12 +536,230 @@ static void delete_button_clicked(GtkButton *button, gpointer user_data)
...
+    gtk_tree_model_get(model, &iter, DT_LIB_TAGGING_TREE_COL_PATH, &tagname, -1);
+  char *subtag = g_strrstr(tagname, "|");
+
+  gint tag_count;
+  gint img_count;
+  dt_tag_count_tags_images(tagname, &tag_count, &img_count);
+  if (tag_count == 0) return;
+
+  GtkWidget *win = dt_ui_main_window(darktable.gui->ui);
+  GtkWidget *dialog = gtk_dialog_new_with_buttons("Rename tag?", GTK_WINDOW(win), GTK_DIALOG_DESTROY_WITH_PARENT,
```
TurboGit: The title should be in lower-case and translatable.

src/libs/tagging.c
```diff
@@ -624,6 +1097,20 @@ void gui_init(dt_lib_module_t *self)
...
   g_signal_connect(G_OBJECT(button), "clicked", G_CALLBACK(export_button_clicked), (gpointer)self);
 
+  button = gtk_button_new_with_label(_("tree"));
+  d->toggle_tree_button = button;
+  gtk_widget_set_tooltip_text(button, _("toggle list / tree view"));
+  dt_gui_add_help_link(button, "tagging.html#tagging_usage");
+  gtk_box_pack_end(hbox, button, FALSE, TRUE, 0);
+  g_signal_connect(G_OBJECT(button), "clicked", G_CALLBACK(toggle_tree_button_callback), (gpointer)self);
+
+  button = gtk_button_new_with_label(_("sugg"));
```
TurboGit: The is too many buttons on the same line. This goes out of screen with French translation at least. We have:

new / delete / import / export / rename / tree / sugg

I would propose, 2 lines:

import / export / suggestion
new / delete / rename

And I would put the toggle for the tree/iist with a small icon on the right of suggestion.

src/libs/tagging.c
```diff
@@ -106,21 +118,44 @@ void connect_key_accels(dt_lib_module_t *self)
...
     int imgsel = dt_control_get_mouse_over_id();
     d->imgsel = imgsel;
     count = dt_tag_get_attached(imgsel, &tags, FALSE);
+    view_type = 2;
   }
   else // related tags of typed text
-    count = dt_tag_get_suggestions(d->keyword, &tags);
-
+  {
+    gboolean suggestion = dt_conf_get_bool("plugins/darkroom/tagging/suggestions");
```

src/libs/tagging.c
```diff
@@ -106,21 +118,44 @@ void connect_key_accels(dt_lib_module_t *self)
...
+    view_type = dt_conf_get_bool("plugins/darkroom/tagging/treeview") ? 1 : 0;
```

src/libs/tagging.c
```diff
@@ -130,24 +165,182 @@ static void update(dt_lib_module_t *self, int which)
...
+        guint tagc = ((dt_tag_t *)taglist->data)->count;
+        guint sel = ((dt_tag_t *)taglist->data)->select;
+        if(tag == NULL) continue;
+        char **tokens;
+        tokens = g_strsplit(tag, "\1", -1);
+        if(tokens != NULL)
+        {
+          // find the number of common parts at the beginning of tokens and last_tokens
+          GtkTreeIter parent = last_parent;
+          int tokens_length = string_array_length(tokens);
```

src/libs/tagging.c
```diff
@@ -130,24 +165,182 @@ static void update(dt_lib_module_t *self, int which)
...
+      for(GList *taglist = tags; taglist; taglist = g_list_next(taglist))
+      { // order such that sub tags are coming directly behind their parent
+        gchar *tag = ((dt_tag_t *)taglist->data)->tag;
+        for(char *letter = tag; *letter; letter++)
+          if(*letter == '|') *letter = '\1';
+      }
+      tags = g_list_sort(tags, sort_tag);
+      for(GList *taglist = tags; taglist; taglist = g_list_next(taglist))
+      {
+        gchar *tag = ((dt_tag_t *)taglist->data)->tag;
```
TurboGit: All const from tag to sel.

src/libs/tagging.c
phweyland: ok

src/libs/tagging.c
phweyland: ok

src/libs/tagging.c
phweyland: ok

src/libs/tagging.c
phweyland: not this one.

src/libs/tagging.c
phweyland: ok

src/libs/tagging.c
phweyland: ok

src/libs/tagging.c
phweyland: I've split in two lines as suggested. The other option would have been to remove the buttons delete & rename to keep only the click-right sub-menu.
I'm ok to use icons but I need to understand how I've to do that. I that case I think both suggestion/no suggestion and tree/list should be icons. For suggestion/no suggestion + / - could be good, couldn't it ?

TurboGit:
When starting dt after having selected the "tree" view for tags the button suggestion/no-suggestion is displayed even if this mode is not available in tree mode.

TurboGit:
And I agree, we can remove the delete / rename and let only the actions in the pop-up menu. If we use icons for suggestion and tree we can probably put back all buttons on the same line. This would be really nice.

phweyland:
> 
> 
> When starting dt after having selected the "tree" view for tags the button suggestion/no-suggestion is displayed even if this mode is not available in tree mode.

Yes, that was the issue I've described in the first post. I've found the way to make it work (  gtk_widget_set_no_show_all(GTK_WIDGET(button), TRUE);)

phweyland:
> If we use icons for suggestion and tree

I've found the paint.c which provides icons. The plus/minus one can be applied for suggestion. I will have to build a new one for list/tree.

Nilvus:
> @Nilvus I had a look about this but that cannot work as folders. In folders every image is counted once. In tags, every image can have several tags. Then when the counts are cumulated one ends up with far more images than reality.
> However, even without cumulating, the current version doesn't display the tag node count if the tag is a not a leave. I'll make a PR on this.

Ok, make as you can. Of course, if total images of a parent is the real total, not a simple cumulated one that could count an image two or more times.

phweyland:
Using cairo paint, is it the right way ? ... not sure it compatible with css icons ...

phweyland:
image


image

image


Here is where I am. The plus/minus is the one used in blending. I've made the other.
It seems the active state (toggle button) is highlighted. This is not desirable.

What do you think ?

I've not found yet an example with icon. Could pinpoint some code ? We could use icons like below:

format-indent-less-rtl  format-justify-fill

zoom-in  zoom-out

phweyland:
I'm happy with the following. Toggle button highlighted means the button is activated. That's consistent with the rest of dt.

image

(default)
image

image

src/libs/tagging.c

phweyland:
> When starting dt after having selected the "tree" view for tags the button suggestion/no-suggestion is displayed even if this mode is not available in tree mode.

Fixed

TurboGit:
Looks really nice. Yet we still want to remove the rename / delete buttons as with French translation (and possibly some other verbose language like German) the line becomes too long. I think having just the pop-up menu (+ documentation) will be good enough.

src/common/tags.c
```diff
@@ -727,6 +749,166 @@ uint32_t dt_tag_get_suggestions(const gchar *keyword, GList **result)
...
+                              "SELECT COUNT(DISTINCT ti.imgid) FROM main.tagged_images AS ti "
+                              "JOIN memory.similar_tags AS st ON st.tagid = ti.tagid",
+                              -1, &stmt, NULL);
+
+  sqlite3_step(stmt);
+  *img_count = sqlite3_column_int(stmt, 0);
+  sqlite3_finalize(stmt);
+
+  DT_DEBUG_SQLITE3_EXEC(dt_database_get(darktable.db), "DELETE FROM memory.similar_tags", NULL, NULL, NULL);
+
```
TurboGit: superfluous empty line here.

src/common/tags.c
```diff
@@ -727,6 +749,166 @@ uint32_t dt_tag_get_suggestions(const gchar *keyword, GList **result)
...
+  sqlite3_step(stmt);
+  sqlite3_finalize(stmt);
+
+  g_free(keyword_expr);
+
+  DT_DEBUG_SQLITE3_PREPARE_V2(dt_database_get(darktable.db),
+                              "SELECT ST.tagid, T.name FROM memory.similar_tags ST "
+                              "JOIN data.tags T ON T.id = ST.tagid ",
+                              -1, &stmt, NULL);
+  dt_tag_t *t;
```
TurboGit: remove and ...

src/common/tags.c
```diff
@@ -727,6 +749,166 @@ uint32_t dt_tag_get_suggestions(const gchar *keyword, GList **result)
...
+  while(sqlite3_step(stmt) == SQLITE_ROW)
+  {
+    t = g_malloc0(sizeof(dt_tag_t));
```
TurboGit: dt_tag_t *t = g_malloc0(...);

src/common/tags.c
```diff
@@ -727,6 +749,166 @@ uint32_t dt_tag_get_suggestions(const gchar *keyword, GList **result)
...
+                              "SELECT count(*) FROM main.selected_images",
+                              -1, &stmt, NULL);
+
+  sqlite3_step(stmt);
+  uint32_t nb_selected = sqlite3_column_int(stmt, 0);
+  sqlite3_finalize(stmt);
+
+  /* Now put all the bits together */
+  DT_DEBUG_SQLITE3_PREPARE_V2(dt_database_get(darktable.db),
+                              "SELECT T.name, ST.tagid, MT.count, CT.tagnb FROM memory.similar_tags ST "
```
TurboGit: A nice SQL statement :)

src/common/tags.c
```diff
@@ -727,6 +749,166 @@ uint32_t dt_tag_get_suggestions(const gchar *keyword, GList **result)
...
+                              "LEFT JOIN (SELECT tagid, COUNT(*) AS tagnb FROM tagged_images AS i "
+                                "WHERE imgid IN (SELECT imgid FROM main.selected_images) GROUP BY tagid ) AS CT "
+                                "ON CT.tagid = ST.tagid "
+                              "WHERE T.name NOT LIKE 'darktable|%%' "
+                              "ORDER BY T.name ",
+                              -1, &stmt, NULL);
+
+  /* ... and create the result list to send upwards */
+  uint32_t count = 0;
+  dt_tag_t *t;
```
TurboGit: remove...

src/common/tags.c
```diff
@@ -727,6 +749,166 @@ uint32_t dt_tag_get_suggestions(const gchar *keyword, GList **result)
...
+  while(sqlite3_step(stmt) == SQLITE_ROW)
+  {
+    t = g_malloc0(sizeof(dt_tag_t));
```
TurboGit: dt_tags_t *t = ...

src/libs/tagging.c
```diff
@@ -106,21 +119,44 @@ void connect_key_accels(dt_lib_module_t *self)
   dt_accel_connect_lib(self, "tag", g_cclosure_new(G_CALLBACK(_lib_tagging_tag_show), self, NULL));
 }
 
+static int string_array_length(char **list)
```
TurboGit: Isn't this what g_strv_length() does?

src/libs/tagging.c
```diff
@@ -130,24 +166,182 @@ static void update(dt_lib_module_t *self, int which)
...
+      for(GList *taglist = tags; taglist; taglist = g_list_next(taglist))
+      {
+        const gchar *tag = ((dt_tag_t *)taglist->data)->tag;
+        const guint id = ((dt_tag_t *)taglist->data)->id;
+        const guint tagc = ((dt_tag_t *)taglist->data)->count;
+        const guint sel = ((dt_tag_t *)taglist->data)->select;
+        if(tag == NULL) continue;
+        char **tokens;
+        tokens = g_strsplit(tag, "\1", -1);
+        if(tokens != NULL)
```
TurboGit: if(tokens)

src/libs/tagging.c
```diff
@@ -130,24 +166,182 @@ static void update(dt_lib_module_t *self, int which)
...
+              {
+                gtk_tree_model_get(model, &parent2, DT_LIB_TAGGING_TREE_COL_SEL, &sel1, -1);
+                if (!sel1)
+                  gtk_tree_store_set(GTK_TREE_STORE(model), &parent2, DT_LIB_TAGGING_TREE_COL_SEL, 1, -1);
+                child = parent2;
+              }
+            }
+            common_length++;
+            parent = iter;
+            g_free(pth2);
```
TurboGit: remove

src/libs/tagging.c
```diff
@@ -130,24 +166,182 @@ static void update(dt_lib_module_t *self, int which)
...
+          }
+          g_free(pth);
```
TurboGit: and replace this with:
g_strfreev (tokens);

Less error prone and when reading code it makes it clear that the vector is freed.

src/libs/tagging.c
```diff
@@ -130,24 +166,182 @@ static void update(dt_lib_module_t *self, int which)
...
+
+          // remember things for the next round
+          if(last_tokens) g_strfreev(last_tokens);
+          last_tokens = tokens;
+          last_parent = parent;
+          last_tokens_length = tokens_length;
+        }
+      }
+      g_strfreev(last_tokens);
```
TurboGit: don't we want a:
g_strfreev(tokens);

here?

src/common/tags.c
phweyland: done

src/common/tags.c
phweyland: done

src/common/tags.c
phweyland: done

src/common/tags.c
phweyland: :)

src/common/tags.c
phweyland: done

src/common/tags.c
phweyland: done

src/libs/tagging.c

src/libs/tagging.c
phweyland: That's the same and consistent with few lines above.

src/libs/tagging.c
phweyland: yes :)

phweyland:
> Yet we still want to remove the rename / delete buttons

Done.

phweyland:
> (+ documentation)

If I'm not mistaken the dt manual is not in dt-org. What is the repository I should use for that ?

Nilvus:
> > (+ documentation)
> 
> If I'm not mistaken the dt manual is not in dt-org. What is the repository I should use for that ?

https://gitlab.com/darktable/doc

TurboGit:
All good to me now ! Thanks.

TurboGit:
@Nilvus : this is the live doc, but the documentation source is in the main repository.

TurboGit:
For the tagging module, It is in doc/usermanual/lighttable/panels/tagging.xml

phweyland:
> It is in doc/usermanual/lighttable/panels/tagging.xml

Thanks Pascal!

junkyardsparkle:
This is wonderful, thank you! Some comments after abusing it for a short time:
- It would be nice if the list didn't scroll back to the top each time you delete a tag, making cleanup less tedious
- I can't seem to rename a tag so as to move it out of its current heirarchy. Example, I have accidentally tagged images with `airports|seatac` which should have been `airport|seatac` (no 's')... renaming to `airport|seatac` creates `airports|airport|seatac`. Might there be some way to imply "root" in the renaming feature?
- There's a feature in [geeqie](http://www.geeqie.org/) called "helper keywords" (contrasted with "active keywords") which is just a parent node used for organizing tags in the tree, but which isn't actually applied as part of the heirarchy when one of the children is applied. It can be handy for organizing loose "label" tags... for example, I use one of these for "activities", since I don't really need the tag "activities" to be applied to the images. I'm not sure how hard this would be to implement, but I just thought I'd mention it for consideration.

Already as it is, this addition has helped me with doing some cleanup, thanks again for the work.

phweyland:
>     * t would be nice if the list didn't scroll back to the top each time you delete a tag, making cleanup less tedious

Agreed. So far I've just modified the existing module which reinitializes the tree each time there is a change. Not only for delete, but attach, detach, ... nearly everything. Needs some additional work ... working at this. 

>     * I can't seem to rename a tag so as to move it out of its current hierarchy.

Correct. The next useful function should be a "move" of a subtree to another place. Future work ?

>     * There's a feature in [geeqie](http://www.geeqie.org/) called "helper keywords"

Thanks for the link, I'll look at it. And yes, there are a lot of potential improvements around tags (category, synonyms, export flags, ...) ... Some of them imply changes in the db...

junkyardsparkle:
> Not only for delete, but attach, detach, ... nearly everything

Yes, this was the next thing I noticed after cleaning up the tags and starting to actually tag some pictures... quite a problem for some workflows (as when moving through images of group activities in darkroom mode to better see which people feature prominently enough in the shot to warrant being added as a tag). To be fair, the old version wasn't great for this either, partly due to the small view into the working list; the larger one in the tree view seems like an improvement by itself... maybe it would enhance "list" mode as well, and keep the size of the module consistent when switching modes?

> Correct. The next useful function should be a "move" of a subtree to another place. Future work ?

My thought would be to simply present the entire hierarchy in the rename field, allowing the user to change whatever parts they want... unless this is a Bad Idea for some reason which I'm overlooking...

phweyland:
> quite a problem for some workflows (as when moving through images
> keep the size of the module consistent when switching modes

Good catches !

> allowing the user to change whatever parts they want

This one is trickier ...

PR #2815, opened 2019-07-24
fix images count in collect.c for tags

phweyland:
the current state is:
image

it should be this one:
image


TurboGit:
Looks good! Thanks.

Issue #2413, opened 2019-04-15
Filmic crashes


PaoloAst:
[_IGP1545.DNG.zip](https://github.com/darktable-org/darktable/files/3081145/_IGP1545.DNG.zip)

**Describe the bug**
With the attached picture the activation of filmic module causes a crash
**To Reproduce**
1. Open this image
2. click on filmic module
3. enjoy the crash

**Platform (please complete the following information):**
 - OS: Ubuntu 
 - Version 18.10

**Additional context**
This happens with or without OpenCL. I'm posting a specific image, but this happens with all images (at least with DNG from Pentax)

Nilvus:
Yes revision d5053b13 from Tobias make filmic crash. Revert back to revision before this one to have filmic ok with :

git checkout 1b2804a579555b0d483de8e62a28f7965d5e9edd

then

git submodule update

and compile again

It seems that this guy like to send revisions without consultation with other developers that could hassle users (like revert back awful and annoying 1st april game on its own way or change css on its own way ; why since darktable is themable ?) or modify other revisions that  other developers have discussed before sending them... I know that he is an old developer but darktable is a free and community software. And all choices have sides effects for all users, not one person or few ones, but lot of users. Ok, it's made on free time but all choices have consequences to have in mind...

I hope he will read this because I have enough to see that as a user (I'm not a developer, I only contribute on translating manual).

TurboGit:
I have added a guard to avoid the crash. My fix is probably not the best one but at least filmic should be usable again. I agree @Nilvus it is quite painful to have to deal with those regressions.

Nilvus:
Yes, it's fixed. Thanks @TurboGit for that.

PaoloAst:
Thank you!

AxelG-DE:
@Nilvus if you start studying commits in that detail, you are a dev in my eyes. Otherwise, how to categorize when I consider my self being a power user :smile:

If you want Tobias to read your comment above and giving him the chance to prove acting considerate, reasonable and objectively, you could have added @houz :wink:
May either one convince the other one. Me as a power user I appreciate all your guys efforts ans commitment.

Nilvus:
> @Nilvus if you start studying commits in that detail, you are a dev in my eyes. Otherwise, how to categorize when I consider my self being a power user smile

[...]

PR #2818, opened 2019-07-26
fix combobox behavior to mimic pre-d5053b1 state

jakubfi:
While working on #2793 I noticed that comboboxes for setting focal length, f-stop and distance in lens correction module are not being set to any value automaticaly. This worked fine in 2.6.2. @TurboGit pointed it out too in https://github.com/darktable-org/darktable/pull/2793#issuecomment-513809052. Initialiy I planned to fix this in place, but since this part of lens.c code didn't really change, it looked like an issue potentialy affecting all comboboxes. So I went hunting.

This got broken in 08e9290cd7c7e8a3eb8bb7f783834fc50eaf4884, which together with dbdbbcd97445bffd1bc7e9ad6037c91b53a4af66 and 964a3c35056f2c38ad013bf9c4c7eadd8b7f02c5 were fixes for #2413 introduced in d5053b13865e5a7d2ee08061c3f1a617f7e730f4. Before this series of changes all comboboxes had position 0 active by default, as I understand. Lens correction UI relied on that, perhaps other modules do as well. After this change, comboboxes have now active=-1 until position is explicitly set.

This PR makes combobox mimic the original behavior:
- when (and only when) the first entry is inserted or added to a combobox, it becames the active one,
- when the last entry is removed from a combobox, it is deactivated.

Additionaly, when removing entries from a combobox with dt_bauhaus_combobox_remove_at(), it will now move the active position up when necessary.

@TurboGit @houz - let me know if the proposed solution is acceptable or requires more work. This can alternatively be solved by adding explicit dt_bauhaus_combobox_set() in all places where a combobox is populated but not set to any position (if the intention was to disallow having any comboboxes without an explicit default state).

### DIFF
--- a/src/bauhaus/bauhaus.c
+++ b/src/bauhaus/bauhaus.c
@@ -1043,6 +1043,7 @@ void dt_bauhaus_combobox_add_full(GtkWidget *widget, const char *text, dt_bauhau
   d->num_labels++;
   dt_bauhaus_combobox_entry_t *entry = new_combobox_entry(text, align, TRUE, data, free_func);
   d->entries = g_list_append(d->entries, entry);
+  if(d->active < 0) d->active = 0;
 }
 
 void dt_bauhaus_combobox_set_editable(GtkWidget *widget, int editable)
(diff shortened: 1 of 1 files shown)


TurboGit:
A tricky area but looks like it is working fine. I'll merge now to ensure more field testing. Thanks.

Issue #2817, opened 2019-07-26
New tag tree : expand/collapse tree not possible by clicking on the left arrow


Nilvus:
**Describe the bug**
@phweyland : great new tag tree but an annoying bug on my system. I can't keep the tree expanded, if I click on a left arrow, the tree collapse the second I stop the click on it. It's only when clicking on the left arrow that I have the behavior. If I click on the tag just after or on the column on the left of the arrow, it's ok

**To Reproduce**
1. Go to 'tags module on right'
2. Click on 'the new toggle tree/list view' to be on tree view
3. Click on an arrow
4. release the click and see the tree closing himself

**Expected behavior**
Having the possibility to expand normally the tree on normal behavior.

**Platform (please complete the following information):**
 - OS: Ubuntu 19.04
 - Version last master

Nilvus:
@phweyland : if you have read the mail before this, forget my first "bug" with the quite strange tag. It's maybe a tag from an image I have previously download from french darktable forum and delete the image since without seeing this tag. Just think about that after posting this issue.

Anyway, a bug remains.

phweyland:
> I can't keep the tree expanded, if I click on a left arrow, the tree collapse
The expansion/collapsing works well for me (my OS is Windows).
Wherever I left-click on the line (before the arrow, on it or after it) the line toggles its expansion (as it is expected).

What do you mean by "I click on a left arrow" ?
The only arrows I have are down and right.

Double-click attaches the tag (if exists) and, yes, the tree collapses (this may be a wrong behaviour, I'll check this). Have configured somewhere double-click = simple-click ?

Nilvus:
> > I can't keep the tree expanded, if I click on a left arrow, the tree collapse
> > The expansion/collapsing works well for me (my OS is Windows).
> > Wherever I left-click on the line (before the arrow, on it or after it) the line toggles its expansion (as it is expected).
> 
> What do you mean by "I click on a left arrow" ?
> The only arrows I have are down and right.
> 
> Double-click attaches the tag (if exists) and, yes, the tree collapses (this may be a wrong behaviour, I'll check this). Have configured somewhere double-click = simple-click ?

Nothing related to double click here

I talk about that (just on the left of a parent tag (red circle below)). The behavior is not the same as clicking (on same line) on the column on the left and on the tag text.

![Capture-20190726205253-297x258](https://user-images.githubusercontent.com/45535283/61974899-f852e880-afe7-11e9-9f1d-887dd4321264.png)

[...]

PR #2820, opened 2019-07-28
Tagging.c - avoid to reinit the tag tree/list on common events

phweyland:
The goal of this PR is to make easier the tagging workflow keeping the tag window as stable as possible.
Compared to the previous version the window is now maintained for the following events:
- hover on images
- new selection of images
- lighttable <-> darkroom move
- attach & detach tag
- rename tag
- delete tag

Thanks to @Nilvus and @junkyardsparkle for their comments.

### DIFF
--- a/src/common/tags.c
+++ b/src/common/tags.c
@@ -834,6 +834,34 @@ void dt_tag_get_tags_images(const gchar *keyword, GList **tag_list, GList **img_
   DT_DEBUG_SQLITE3_EXEC(dt_database_get(darktable.db), "DELETE FROM memory.similar_tags", NULL, NULL, NULL);
 }
 
+uint32_t dt_selected_images_count()
+{
+  sqlite3_stmt *stmt;
+
+  DT_DEBUG_SQLITE3_PREPARE_V2(dt_database_get(darktable.db),
+                              "SELECT count(*) FROM main.selected_images",
+                              -1, &stmt, NULL);
+  sqlite3_step(stmt);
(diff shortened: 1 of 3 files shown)


Nilvus:
@phweyland : Thanks for your work on that.
I don't think it's related to issue I post #2817 but this issue still exist with that PR.

TurboGit:
All this looks good to me... And a nice clean-up in the code. Thanks!

junkyardsparkle:
Thanks, this is working great! The one issue I can see now is that without a reinit on collection change, the checkboxes in the left margin (which I think are supposed to indicate that the tags are present within the current collection?) may now be stale and wrong when the collection is changed, until "flushed" by toggling between tree/list view.

phweyland:
> The one issue I can see now is that without a reinit on collection change, the checkboxes in the left margin (which I think are supposed to indicate that the tags are present within the current collection?) may now be stale and wrong when the collection is changed, until "flushed" by toggling between tree/list view.

You are right. I've been a bit lazy on this. It requires to evaluate the full branch from root to leave to update them properly. Possible, but a bit of work to do it. I thought it wasn't mandatory at this stage, but I may be wrong. 

>  the meaning of the checkmarks (and line) in the left column... how do those work, exactly?

* nothing => the tag is not present on selected images
* dash => On tag (leave) at least one selected image, but not all, has this tag attached. On path nodes, means that at least one selected image has at least one child tag attached
* check mark => all selected images have this tag attached. Doesn't happen on path nodes.

PR #2821, opened 2019-07-28
Adding support for DNGs DefaultUserCrop

jenshannoschwalm:
DNG files might have the tag _DefaultUserCrop_ (V. 1.4), this is used by cameras like LeicaQ2 for example using cropping to have pseudo-35,50,75mm lenses. (You see a frame inside the viewfinder or back display.)
When you shoot raw+jpeg the jpeg is cropped accordingly, the DNG contains the whole image but has the DefaultUserCrop tag telling where the image is suggested to be cropped by the appication. From what i have read, this is lightrooms behaviour.
Currently this is **not** supported by dt.

How should this be handled? We don't want a 'forced' clipping (because the scene layout could be better than i thought when shooting), instead it would be nice to have the clipping module behaving in this way:
**Whenever i switch the clipping module ON or click the modules reset button the visible clipping rectangle should not cover the whole image but take the DefaultUserCrop rectangle.**

How is this implemented here?
1. The  `dt_image_t` struct has ` float usercrop[4]` holding the clipping parameters. We make sure they contain [0,0,1,1] by **default for all images** except for those with a defined DefaultUserCrops
2. In the clipping modules `void reload_defaults(dt_iop_module_t *self)` we can now take `usercrop[4]` as default borders.

What about performance?
1. We don't want to bloat the database with data used only by a smaller number of images.
2. We don't want a performance penalty. When importing or re-reading exif data we test for valid and usable DefaultUserCrop data. If such data is found `DT_IMAGE_HAS_USERCROP` in `dt_image_t flags` is set. This is kept in the database, nothing else.

How are the clipping parameters made available?
1. Whenever a raw file is read into the cache this is done by `dt_imageio_open_rawspeed`.
2. There we test for the flag `DT_IMAGE_HAS_USERCROP`, if true we read the DefaultUserCrop data.

How to test?
rawpixls does not want an example as it does not differ for them. Here you can get one: [https://www.dropbox.com/s/xtgnjexbnaizx47/L1030415.DNG?dl=0](url)

Anything more?
1. The collections should also be able to search for images with cropping ... this is wip.


jenshannoschwalm:
As such clipped images might want a different workflow we want them to be searched.
So i added:
When a file has a valid DefaultUserCrop the tag darktable|mode|usercrop is added.

TurboGit:
> darktable|mode|usercrop

I would use:

darktable|mode|default-user-crop

The default is important here as we really want to convey that this has nothing to do with a crop done by the user with the crop&rotate module for example.

jenshannoschwalm:
>darktable|mode|exif-crop ?
 
good point. This seems to be better.

BTW, i think that automatically added tags could be helpful if a condition means: It is very likely that a special workflow might be appropriate.

1) monochrome is one example. Do you want me to add darktable|mode|monochrome for that?
2) other conditions coming to your mind?

TurboGit:
Good idea, I'm for darktable|mode|monochrome tag too.

No other conditions come to mind right now. Thanks.

Entropy512:
As far as confusion with a crop done with the crop and rotate module...  Since some modules already have "camera-derived" defaults such as whitebalance - what about using these tags to turn on the crop/rotate module by default with a "camera" preset?

TurboGit:
@Entropy512 : not sure about this, but for sure this should be part of another ticket.

TurboGit:
@jenshannoschwalm : please do rebase on master and not merge master into your topic branch. this will clean-up the history. thanks.

jenshannoschwalm:
@TurboGit ; i have changed tagname to darktable|mode|exif-crop and added darktable|mode|monochrome.
Also: Sorry about my git 'noise', rebase does not help right now. Do you want me to close this and re-open with another branch?

@Entropy512 I think it would be nice to have the module presets looking for tags too. -- Another ticket.

jenshannoschwalm:
Closing this as it is replaced by #2833

PR #2824, opened 2019-07-30
Tagging.c : rename path and selection indicator update

phweyland:
This PR brings the 2 items:
* add a command `rename path` which allows to move a part of the tree, insert or remove a level by adding or removing the separator '|'. This command is active only on path element and not on leave.
* update the selection indicator along the tree in case of images selection change and attach / detach tag. Was an issue pinpointed by @junkyardsparkle .

### DIFF
--- a/src/libs/tagging.c
+++ b/src/libs/tagging.c
@@ -153,6 +153,19 @@ static int get_treeview_type(dt_lib_module_t *self)
   else return DT_LIB_TAGGING_VIEW_LIST; // list view - TAGGING_TREE model
 }
 
+static void propagate_sel_to_parents(GtkTreeModel *model, GtkTreeIter *iter)
+{
+  guint sel;
+  GtkTreeIter parent, child = *iter;
+  while(gtk_tree_model_iter_parent(model, &parent, &child))
+  {
+    gtk_tree_model_get(model, &parent, DT_LIB_TAGGING_TREE_COL_SEL, &sel, -1);
+    if (!sel)
+      gtk_tree_store_set(GTK_TREE_STORE(model), &parent, DT_LIB_TAGGING_TREE_COL_SEL, 1, -1);
+    child = parent;
+  }
+}
+
 static void init_treeview(dt_lib_module_t *self, int which)
 {
   dt_lib_tagging_t *d = (dt_lib_tagging_t *)self->data;
@@ -249,17 +262,7 @@ static void init_treeview(dt_lib_module_t *self, int which)
                               DT_LIB_TAGGING_TREE_COL_SEL, sel,
                               -1);
             if (sel)
-            { // propagate selection to parents
-              guint sel1;
-              GtkTreeIter parent2, child = iter;
-              while(gtk_tree_model_iter_parent(model, &parent2, &child))
-              {
-                gtk_tree_model_get(model, &parent2, DT_LIB_TAGGING_TREE_COL_SEL, &sel1, -1);
-                if (!sel1)
-                  gtk_tree_store_set(GTK_TREE_STORE(model), &parent2, DT_LIB_TAGGING_TREE_COL_SEL, 1, -1);
-                child = parent2;
-              }
-            }
+              propagate_sel_to_parents(model, &iter);
             common_length++;
             parent = iter;
             g_free(pth2);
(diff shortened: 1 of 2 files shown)


phweyland:
This will remove the `test` level from the path.
image

src/libs/tagging.c
```diff
@@ -383,11 +386,141 @@ static void raise_signal_tag_changed(dt_lib_module_t *self)
...
+    reset_sel_on_path(model, &parent, FALSE);
+    calculate_sel_on_path(model, &parent, FALSE);
+  }
+}
+
+// get the new selected images and update the tree selection
+static void update_sel_on_tree(GtkTreeModel *model)
+{
+  GList *tags = NULL;
+  guint count = dt_tag_get_attached(-1, &tags, TRUE);
```

src/libs/tagging.c
```diff
@@ -533,15 +645,24 @@ static void detach_selected_tag(dt_lib_module_t *self, dt_lib_tagging_t *d)
   const int view_type = get_treeview_type(self);
   if (view_type != DT_LIB_TAGGING_VIEW_SIMPLE)
   {
+    guint count = dt_tag_images_count(tagid);
```

src/libs/tagging.c
```diff
@@ -751,6 +873,21 @@ static void rename_button_clicked(GtkButton *button, dt_lib_module_t *self)
   if (gtk_dialog_run(GTK_DIALOG(dialog)) == GTK_RESPONSE_YES)
   {
     const char *newtag = gtk_entry_get_text(GTK_ENTRY(entry));
+    char *message = NULL;
+    if (!newtag[0])
+      message = _("null string is not allowed. Aborting");
```
TurboGit: null -> empty

I would write a more user friendly message:

"empty tag is not allowed, aborting."

src/libs/tagging.c
```diff
@@ -751,6 +873,21 @@ static void rename_button_clicked(GtkButton *button, dt_lib_module_t *self)
...
+    if(strchr(newtag, '|') != 0)
+      message = _("'|' character is not allowed for renaming tag.\nTo modify the hierachy use rename path instead. Aborting.");
```
TurboGit: To -> to (no capital letter). Likewise for Aborting -> aborting.

TurboGit:
Forgot the most important. This needs to be released on current master to resolve conflicts. Thanks.

src/libs/tagging.c

src/libs/tagging.c

src/libs/tagging.c

src/libs/tagging.c

phweyland:
> This needs to be released on current master to resolve conflicts.

Which conflicts ? Not sure to understand. I've already merged with master and resolved the conflicts. Is there some remaining ?

Another question. the old tag mode, as is, turns the code a bit complicated. Its own added value is about the suggestions, but it is not very clean because it mixes suggestions with tag list. I propose to have a specific box/window to offer suggestions (and only suggestions). This box could be collapsed if not needed. And to keep only the 2 tag lists (instead of 3), list or tree. This is how it's done in other DAMs. 
To resume there would be 3 boxes in tagging: Attached tags (as today), Suggestions (collapsible) and Tag list or tree.
What do you think ?

TurboGit:
You need to rebase, not merge master on your topic branch. Anyway, I see that now conflicts are resolved. Thanks.

TurboGit:
Ok, looks good. Thanks again for all the work on tagging support!

junkyardsparkle:
> Another question. the old tag mode

It might be good to take this question to the user list for a more comprehensive survey of how people are currently using this, and their feelings about changing it. Myself, I never developed a workflow that I liked using that mode. It would tempt me into trying to use primarily the keyboard, but that usually ended up with losing my image selection after tabbing through the filtered list and hitting enter... it seems to require a lot of switching between mouse and keyboard for me.

For the case of trying to keep a small collection of keywords visible for clicking (for instance, the names of a group of friends), typing "friends|" into the input field works, but the list widget is so small that there ends up being much scrolling involved. In some cases, typing a placename into the field collects a list of associated people, but this is haphazard and depends on already having enough existing tagging for the "smart" association to work with... and again, the list widget is really too small.

I'll probably be using the tree mode most of the time now, since it seems to involve less keyboard/mouse switching for most of my use cases, so I won't be upset about changes to the old mode, but other people may feel differently.

> add a command rename path which allows to move a part of the tree, insert or remove a level by adding or removing the separator '|'. This command is active only on path element and not on leave.

This is working great for me so far. I'm wondering if it would be good for discoverability to also include this option in the leaf menu, since people may be clicking there looking for a "move" operation (it would then of course work the same way as if selected on the parent)? Thanks again for the work!

PR #2828, opened 2019-07-31
[WIP] iop/basecurve: Rework exposure fusion

Entropy512:
The enfuse algorithm defaults to a target brightness of 0.5 for well exposed pixels

While it isn't made explicit in the paper of Mertens et all, it's strongly implied
by this value they are operating in an sRGB color space with the sRGB transfer curve.  Most cameras
target 18% reflectance (0.18) as "middle grey" for exposure - which lands at around 0.5 after applying
a 2.4 gamma curve

Convert all images to a 2.4 gamma curve before applying exposure fusion, and then convert to linear at the end

Only calculating weights in gamma space but not performing merging results in strange haloing artifacts

With this patch, visual results are very close to enfuse operating on sRGB JPEG images.  In fact, for highlights
this approach appears to do better by tracking values that would have been clipped in the JPEG, de-weighting
highlights that are severely overexposed more so than if any of their channels were clipped to 1.0

Also - disable saturation and contrast weighting

Saturation is constant when our images are all exposure shifts without any clipping

Similarly, contrast is just a linear or square function of exposure shift the way it is currently calculated,
and again - we don't clip which is the primary reason to weight contrast

Exposure optimum and width are now exposed as sliders instead of harcoded.

Tooltip for exposure bias updated to indicate you really shouldn't be using it - use the highlight reconstruction
and exposure compensation modules instead.  (-EV compensation is nonintuitive, but is potentially required to
offset whitebalance multipliers.)

Remaining TODO items:
1. Even though it's hard to believe anyone ever got visually acceptable results with the old implementation
(I was never able to...), figure out how to preserve compatibility with old edits.  This requires preserving
a lot of strange weighting calculations that don't match the original Mertens paper or enfuse implementation.
2. If local contrast weighting isn't nuked, it needs to be renamed to something more informative than
"adjust features" to explain what it is actually doing for the next developer to look at this
3. Consider using existing darktable colorspace management code instead of apply_pow()
4. Consider not hardcoding the gamma curve value of 2.4 - but we've got a LOT of sliders for fusion now...

### DIFF
--- a/data/kernels/basecurve.cl
+++ b/data/kernels/basecurve.cl
@@ -113,7 +113,7 @@ basecurve_legacy_lut(read_only image2d_t in, write_only image2d_t out, const int
 }
 
 kernel void
-basecurve_compute_features(read_only image2d_t in, write_only image2d_t out, const int width, const int height)
+basecurve_apply_pow(read_only image2d_t in, write_only image2d_t out, const int width, const int height, const float adjust_pow)
 {
   const int x = get_global_id(0);
   const int y = get_global_id(1);
@@ -121,23 +121,46 @@ basecurve_compute_features(read_only image2d_t in, write_only image2d_t out, con
   if(x >= width || y >= height) return;
 
   float4 value = read_imagef(in, sampleri, (int2)(x, y));
-  
+  value.x = native_powr(clamp(value.x, 0.0f, 999.0f),adjust_pow);
+  value.y = native_powr(clamp(value.y, 0.0f, 999.0f),adjust_pow);
+  value.z = native_powr(clamp(value.z, 0.0f, 999.0f),adjust_pow);
+
+  write_imagef (out, (int2)(x, y), value);
+}
+
+kernel void
+basecurve_compute_features(read_only image2d_t in, write_only image2d_t out, const int width, const int height, const float opt, const float w)
+{
+  const int x = get_global_id(0);
+  const int y = get_global_id(1);
+
+  if(x >= width || y >= height) return;
+
+  float4 value = read_imagef(in, sampleri, (int2)(x, y));
+/*
+  Disable saturation weighting for now, it will only be useful when "preserve colors" is turned off since we don't clip.
+
+  Enfuse also defaults to turning this off now.
+
+  TODO:  Figure out how to preserve legacy compatibility, even though it's questionable whether anyone got visually acceptable results
+  from the old configuration.  I never could...  Unfortunately the method of combining weights here is completely different
+  from the Mertens paper AND enfuse, so it would be difficult to preserve compatibility without keeping a pile of hardcoded constants we
+  don't want to exist in the future around.
+*/
+/*
   const float ma = max(value.x, max(value.y, value.z));
   const float mi = min(value.x, min(value.y, value.z));
   
-  const float sat = 0.1f + 0.1f * (ma - mi) / max(1.0e-4f, ma);
+  const float sat = 0.2f * (ma - mi) / max(1.0e-4f, ma);
   value.w = sat;
-
-  const float c = 0.54f;
+*/
   
-  float v = fabs(value.x - c);
-  v = max(fabs(value.y - c), v);
-  v = max(fabs(value.z - c), v);
+  float v = fabs((value.x + value.y + value.z)/3.0f - opt);
 
-  const float var = 0.5f;
-  const float e = 0.2f + fast_expf(-v * v / (var * var));
+  const float var = w * 1.4142135f;
+  const float e = fast_expf(-v * v / (var * var));
 
-  value.w *= e;
+  value.w = e;
   
   write_imagef (out, (int2)(x, y), value);
 }
--- a/src/iop/basecurve.c
+++ b/src/iop/basecurve.c
@@ -47,7 +47,7 @@
 #define MAXNODES 20
 
 
-DT_MODULE_INTROSPECTION(6, dt_iop_basecurve_params_t)
+DT_MODULE_INTROSPECTION(7, dt_iop_basecurve_params_t)
 
 typedef struct dt_iop_basecurve_node_t
 {
@@ -66,7 +66,22 @@ typedef struct dt_iop_basecurve_params_t
   float exposure_stops;   // number of stops between fusion images
   float exposure_bias;    // whether to do exposure-fusion with over or under-exposure
   int preserve_colors;
+  float exposure_optimum; // Optimum brightness for exposure fusion
+  float exposure_width; // Exposure weight function variance
 } dt_iop_basecurve_params_t;
+
+typedef struct dt_iop_basecurve_params6_t
+{
+  // three curves (c, ., .) with max number of nodes
+  // the other two are reserved, maybe we'll have cam rgb at some point.
+  dt_iop_basecurve_node_t basecurve[3][MAXNODES];
+  int basecurve_nodes[3];
+  int basecurve_type[3];
+  int exposure_fusion;    // number of exposure fusion steps
+  float exposure_stops;   // number of stops between fusion images
+  float exposure_width; // Exposure weight function variance
+  int preserve_colors;
+} dt_iop_basecurve_params6_t;
 
 typedef struct dt_iop_basecurve_params5_t
 {
@@ -112,7 +127,7 @@ typedef struct dt_iop_basecurve_params1_t
 int legacy_params(dt_iop_module_t *self, const void *const old_params, const int old_version,
                   void *new_params, const int new_version)
 {
-  if(old_version == 1 && new_version == 6)
+  if(old_version == 1 && new_version == 7)
   {
     dt_iop_basecurve_params1_t *o = (dt_iop_basecurve_params1_t *)old_params;
     dt_iop_basecurve_params_t *n = (dt_iop_basecurve_params_t *)new_params;
@@ -131,10 +146,12 @@ int legacy_params(dt_iop_module_t *self, const void *const old_params, const int
     n->exposure_fusion = 0;
     n->exposure_stops = 1;
     n->exposure_bias = 1.0;
+    n->exposure_optimum = 0.5;
+    n->exposure_width = 0.2;
     n->preserve_colors = DT_RGB_NORM_NONE;
     return 0;
   }
-  if(old_version == 2 && new_version == 6)
+  if(old_version == 2 && new_version == 7)
   {
     dt_iop_basecurve_params2_t *o = (dt_iop_basecurve_params2_t *)old_params;
     dt_iop_basecurve_params_t *n = (dt_iop_basecurve_params_t *)new_params;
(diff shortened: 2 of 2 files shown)


Entropy512:
As to an analysis of weighting (old vs. new):

[https://pixls-discuss.s3.dualstack.us-east-1.amazonaws.com/optimized/3X/6/a/6a18c66387bab7cfbe533f606f4e6512f327b35a_2_690x550.png](url)

Entropy512:
@edgardoh - Sorry this took so long to publish for you to look at

Entropy512:
Hmm, looks like translations failed due to me changing a tooltip?

TurboGit:
@Entropy512 : indeed, point 1 for backward compatibility is important. thanks.

edgardoh:
> @edgardoh <https://github.com/edgardoh> - Sorry this took so long to
> publish for you to look at
>
Don't be, we're all working on our free time here.

edgardoh:
I played a bit with this and I like it, but I think that there's still room for improvement. A couple of things that bother me are the pow() on the original image, the halos on linear data and the way the exposure increments for every exposure fusion.

With a flat curve I still get terrible halos, so the pow() alone is not enough. The exposure increments too much for my taste, when fusing different shots, usually each shot has +1/+2, not pow(EV*number of shot). The highlights are much better now but still too bright. 
It seems that the weight function is not picking the right pixels to blend, or is giving too much weight to extreme highlights and shadows. Maybe the exp() is not the right function for us, and instead of messing with the input image (with the pow()) we should find a better weight function.

I did a couple of tests using a contrast function (like the one on the basic adjustments) instead of the pow() and I can get an image free of halos with a flat curve. The image is also flat, but it was just a first test.
Do you know if the enfuse is using this exp() as a weight function or something different?

edgardoh:
BTW, travis is reporting an omp error, there's some variables that must be added to the shared()/first private().

Entropy512:
One thing as far as highlights (This was actually the reason I changed the tooltip for exposure bias) - you want to tune your original image (before turning fusion on) so the highlights are below +1.0 - you'll often need -EV exposure compensation in the exposure module to compensate for the white balance multipliers.

Not sure what you mean about the per-exposure shift...  The number used in the slider is EV.  This is converted into a straight multiplier (FYI, this is why I renamed the "EV" variable to "mul" in a previous patch - because by the time that variable was passed, it was a plain old multiplier, not an EV shift).

I've found that if you go beyond +1.5EV shift per exposure or so, the risk of haloing even when operating in gamma-compressed space increases.  I need to go back and run some tests on the original enfuse implementation to see if we might have found another possible bug in DT's implementation.

One thing I need to dig into as far as such bugs - in the original enfuse implementation, weights are normalized to sum to 1.0 prior to building the gaussian pyramid.  It appears (if I'm reading things correctly, I need to reread it tonight), DT is normalizing the weights within each pyramid level - I'm thinking this MIGHT be a contributing factor to the haloing that seems to be far more common in DT's implementation.

I'll look through the travis logs again - what I saw was a translations error and one of the other GCC 9 tests appeared to pass, but I could easily have forgotten one of the OMP pragmas.  GCC 9 is obviously much better at catching silly typos and omissions!  Sometime in the coming weeks I'll try and get that installed on my Ubuntu machine so I can test more thoroughly.

Enfuse defaults to a Gaussian ( exp(-x^2) function for weighting, but does offer alternatives.  The default has always worked well for me.

Entropy512:
I attempted to rework the order of operations so that weights are normalized before the gaussian pyramid is constructed, it appears that this is not the source of the haloing issue, or I didn't "fix" it properly.  Need to do more digging.

I compared the results of this patch with +2EV exposure shift and default weights to enfuse fed with three sRGB JPEGs at 0, +2EV, and +4EV - the enfuse results are vastly superior still (no haloing).

I exported them at too large of a size, otherwise I'd add them as examples.  Time to do some more digging into the blending code...

edgardoh:
I have exposed a lot of settings, and by tweaking them I'm able to get good results, comparable to enfuse, even with a flat curve. The difference is that enfuse is doing it with default settings.

Still there's some halos, probably because of the pyramid.

Entropy512:
Ah, what settings did you expose?

I've been fiddling around a lot, so far it seems like the comments regarding boundary conditions in gauss_reduce() are NOT the issue.  Attempting to pad up the input before running everything has no change in the haloing, same for trying to adapt the code in common/locallaplacian.c

The current enfuse 4.2 release doesn't default to saturation weight of 0.0 - turns out that must be in the version control history but not released yet?  However defaulting it to 0.0 does improve local contrast in the highlights without triggering haloing.

Darktable with this patch is providing MUCH more local contrast in the highlights than enfuse default, at the cost of haloing.  Cranking up the exposure width to a very high value (close to 1 vs. enfuse default of 0.2) greatly reduces haloing but seems to also cause highlight local contrast to become worse than enfuse.

As a side note, grabbing the colorspace transform code from lut3d provides very similar results to apply_pow() without having a bunch of "extra" code.  I'm probably going to push that as an additional patch later this weekend before trying to figure out what's going on down the line.  It will make preserving the "old" behavior easier.

edgardoh:
I added a lot of settings, here is the code:
https://github.com/edgardoh/darktable/tree/rlucyde

Is not the idea to keep them all, but some seem to be a good option.

An example from some play_raw, not remember which one now:
This one is from enfuse:
20190224_DU_TTurtle_0833_Mesh_Bracket_01_DT_enfuse

This one tweaking the settings:
20190224_DU_TTurtle_0833_Mesh_Bracket_01_DT_fusion

There are some halos on the left side not present on the enfuse version, but the enfuse version also has some halos.

I did notice the exposure width, any attempt to reduce the highlights start to produce halos. This is a big issue, because one of the points of this is to recover highlights, or at least not increase them.

In my test version I'm clamping the intermediate images, this has removed some of the ugliest halos that I had so far. I don't think is a problem to clamp it, as long the original image is not clamped.

I'm not sure about the colorspace transform, the lut is doing it because some luts are designed to work with a specific colorspace, but this is not the case here, it seems that what is needed is a gamma curve, and this will be probably faster than a colorspace transform.
Also it has to take into account the work profile, that can be anything, so we should have some sort of setting.

Entropy512:
Ooooh...  sneaky sneak - crank up the verbosity of enfuse, and:
`enfuse: info: using 9 blending levels
enfuse: info: CIELUV color conversion from/to "sRGB" profile
enfuse: info: generating Gaussian pyramid:  g0 g1 g2 g3 g4 g5 g6 g7 g8
enfuse: info: generating Laplacian pyramid: l0 l1 l2 l3 l4 l5 l6 l7 l8
enfuse: info: generating Gaussian pyramid:  g0 g1 g2 g3 g4 g5 g6 g7 g8
enfuse: info: CIELUV color conversion from/to "sRGB" profile
enfuse: info: generating Gaussian pyramid:  g0 g1 g2 g3 g4 g5 g6 g7 g8
enfuse: info: generating Laplacian pyramid: l0 l1 l2 l3 l4 l5 l6 l7 l8
enfuse: info: generating Gaussian pyramid:  g0 g1 g2 g3 g4 g5 g6 g7 g8
enfuse: info: CIELUV color conversion from/to "sRGB" profile
enfuse: info: generating Gaussian pyramid:  g0 g1 g2 g3 g4 g5 g6 g7 g8
enfuse: info: generating Laplacian pyramid: l0 l1 l2 l3 l4 l5 l6 l7 l8
enfuse: info: generating Gaussian pyramid:  g0 g1 g2 g3 g4 g5 g6 g7 g8
enfuse: info: collapsing Laplacian pyramid: l8 l7 l6 l5 l4 l3 l2 l1 l0
enfuse: info: CIELUV color conversion
`
It's not entirely clear originally, but if you dig deeper by searching for thise info prints:
Enfuse performs weighting based on the input colorspace (typically sRGB)
When it constructs the pyramid, it moves everything into CIELUV (at least by default):
https://github.com/jackmitch/enblend-enfuse/blob/master/src/fixmath.h#L2084 - note that I don't think that's an official enfuse version control repo (I think it's a git conversion of their SVN repo), but it's helpful for browsing.

edgardoh:
According to the user manual for floating point images it uses rgb as a
default.

Entropy512:
For floating point, yes - it converts everything to a pseudo-logarithmic RGB curve using log1p() - I say "pseudo" because of the (1+x) inherent to log1p - unless there is a prescaling of x I'm missing to avoid having odd behavior where x<1.0 (which is typical)

I've gotten some strange (but at least no haloing) results by feeding FP to enfuse.  I need to investigate this more.

I've been trying, in general, to make DT behave like enfuse by adding more and more of its behaviors, but I think I need to see if I can figure out how to trigger the haloing I see with darktable when seeing enfuse.  I think there's still something not quite right somewhere in the pyramidal blending stuff.

Widening the exposure width/variance setting in DT appears to mitigate the issue (albeit at the cost of needing a follow-on exposure module to bring the highlights back down below 1.0)
Narrowing the exposure width setting in enfuse does not trigger haloing, even at absurdly low values, indicating it isn't just an issue with scaling of the exposure weighting function.

LAB vs. LUV in enfuse doesn't seem to make a perceptible difference.  I need to try identity on sRGB images tonight.
LAB vs. sRGB vs. gamma-compressed with apply_pow() doesn't seem to make much of a difference in darktable as far as mitigating haloing

Dropping num_levels to 8 (matching DT) in enfuse causes some very slight haloing to appear, but it is much lower in magnitude than in DT - trying to reduce the exposure width setting does not affect haloing significantly.  Dropping num_levels to 6 seems to give the closest match to DT's behavior - but again, it's still significantly lower in magnitude
Adjusing num_levels downwards in DT causes the haloing band to become narrower - it appears to be approximately either 2**num_levels or 2**(num_levels-1) wide

Fiddling with boundary conditions for the Gaussian expansion doesn't seem to make much visible difference - including an attempt to pad all images up by 2**nlevels (similar to what is done in locallaplacian.c)
Enfuse appears to use boundary conditions of either "wrap around" (intended for 360 degree panoramas) or "zero" - switching between these doesn't seem to affect the haloing.  However, there is still some experimentation to be done here.

edgardoh:
I didn't want to go there, but the issue seems to be in the algorithm, dt is using the local laplacian one, but the paper and enfuse is different.

Entropy512:
DT's appears to be somewhat closer to the implementation in enfuse and the paper - but there appears to be a subtle difference still missing.  All of the approaches (including the local laplacian filter) involve creating a Laplacian pyramid of the image(s) in question before operating on them.  The implementation described by the Mertens paper, enfuse, and DT's fusion implementation is pretty similar to the blending approach described in https://www.cs.toronto.edu/~mangas/teaching/320/slides/CSC320L10.pdf - e.g. fairly common.

However, there are some subtle differences:
In the Toronto class slides, the image is padded up to 1+2^n size
In the DT local laplacian implementation, the image is padded up by 2^nlevels (but not to a 1+2^n square) - attempting to adapt this (including padding up) to the fusion code here didn't seem to fix the haloing issues.
In the current DT fusion implementation, at the boundaries, mirrored samples past the edge are chosen
Choosing to perform a "sample and hold" for off-edge pixels in the gaussian blur function doesn't seem to make much difference
Neither does just assuming anything off-edge is zero - this is what enfuse appears to do by default (It optionally wraps around the image, intended for 360 degree panoramas), unless there's some additional boundary condition handling that isn't obvious which is buried in all of the performance optimizations they did.

One thing that is obvious:  In the current dt implementation, output pixels are in some cases darker than the darkest input image.  Which shouldn't be possible (and is not possible in the enfuse implementation - even in the worst case of selecting hardmasks, which provides a similar result to dropping the exposure width/variance parameter to an absurdly low value, the output isn't ever darker than the darkest input.)

I'm going to take another tough look at enfuse's pyramid.h - and if I don't see anything obvious, probably take a break for a few days so I can come back at this with fresh eyes.

edgardoh:
It was the algorithm, among other things. I'll push a PR soon with a working version.

Entropy512:
Nice!  I'm pretty sure when I see it I'll slap myself for missing something obvious.  I've found a few deltas that give minor improvements (such as implementing the boundary extrapolation method described in section 2.4 of http://persci.mit.edu/pub_pdfs/spline83.pdf which is the original paper on multiresolution blending, enfuse/enblend's blending phase is HEAVILY based/derived from that paper) - but it's been clear that there's some error remaining in the implementation that I'm missing, and I've been too busy/fried this week to dig in deeper to figure out what that is.

Taking a break and letting fresh eyes look at the problem is always a good thing.  :)

TurboGit:
This has lost traction, closing. Feel free to reopen if you plan to continue working on this.

PR #2830, opened 2019-08-01
Added placeholder text in styles dialog, fixes #2825

### LINKED ISSUE #2825: Better dialog to insert presets
Working with darktable, it's quite annoying that the dialog to insert the preset doesn't have any label specifing what the textfield has to contain.

They should have a simple label with "Name" and "Description" in order to let the user know what is inserting. I'm a newbie with darktable and I find that should be a minor change that can improve the user experience.

Thanks in advance

sams96:
It's not perfect, since the placeholder text for the name field isn't visible initially due to it being selected, but I thought this was better than taking up space putting titles for each box next to them.


TurboGit:
Right, not perfect but better. Merging. Thanks.

PR #2831, opened 2019-08-02
Tagging.c - bug fixing + allow change path on leaves

phweyland:


phweyland:
Should also solve the issue #2817 .

TurboGit:
Looks ok to me. Thanks.

PR #2838, opened 2019-08-04
Fixing Exif.GPSInfo.GPSLongitude and Exif.GPSInfo.GPSLatitude

### LINKED ISSUE #2835: dartable 2.6.2 always crash upon importation of folders containing pictures edited with Snapseed
I'm on Linux Fedora 30, with darktable-2.6.2-1.fc30.

While trying to import 2 picture folders, darktable starts the importation process and then crash at some point. I can manage to complete the importation process by starting it multiple times, with the same folder, and going through multiple crashes. I've tried with and without the options related to recursive folder import and metadata upon import, without any difference. 

The 2 folders in cause include RAW (nef & dng) + JPG and are relatively large: one of 312 files, for 2.1GB and the other of 36 GB for 5262 files.

Backtrace attached.

[core_backtrace.txt](https://github.com/darktable-org/darktable/files/3463775/core_backtrace.txt)

jenshannoschwalm:
See the Issue #2835, the file mentioned simply has wrong GPS metadata, both Longitude and Latitude should have 3 rationals, in that jpeg file there is only one.

So in exif.cc reading Exif.GPSInfo.GPSLongitude and Exif.GPSInfo.GPSLatitudeboth **both test for 3 rationals** now.

To test: simply import or re-read exif data for that file.

Fixes #2835,

### DIFF
--- a/src/common/exif.cc
+++ b/src/common/exif.cc
@@ -770,7 +770,7 @@ static bool dt_exif_read_exif_data(dt_image_t *img, Exiv2::ExifData &exifData)
     if(FIND_EXIF_TAG("Exif.GPSInfo.GPSLatitude"))
     {
       Exiv2::ExifData::const_iterator ref = exifData.findKey(Exiv2::ExifKey("Exif.GPSInfo.GPSLatitudeRef"));
-      if(ref != exifData.end() && ref->size())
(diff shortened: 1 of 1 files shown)


TurboGit:
Good to see this fixed! Thanks a lot for the continued support.

PR #2846, opened 2019-08-09
new module: exposure fusion

edgardoh:
This module fuses different exposures of an image. It is based on Tom Mertens, Jan Kautz and Frank van Reeth “Exposure Fusion”.


edgardoh:
@TurboGit , I'm missing how to add the translations and a couple of things more needed for new modules, can you help me with this?

src/iop/expfusion.c
```diff
@@ -0,0 +1,1268 @@
...
+
+#ifdef _OPENMP
+#pragma omp parallel for default(none) \
+  dt_omp_firstprivate(size, exp, img_src, img_dest) \
+  schedule(static)
+#endif
+  for(int i = 0; i < size; i++)
+  {
+    img_dest[i] = img_src[i] * exp;
+    img_dest[i] = CLAMP(img_dest[i], 0.f, 1.f);
```
Entropy512: How does the algorithm behave if you don't do this?  This can potentially cause clipped highlights to get weighted in more than they should be.  On the other hand, perhaps this is the mitigation for haloing???

TurboGit:
You need to add this new file into po/POTFILES.in for it to be part of the translation.

One question how this relate to the new exposure fusion code proposed in basecurve?

Thanks.

src/iop/expfusion.c
```diff
@@ -0,0 +1,1268 @@
...
+        img_tmp[(i * wd + j) * ch + k]
+            = img_src[((i)*wd + (j - 2)) * ch + k] * fx[0] + img_src[((i)*wd + (j - 1)) * ch + k] * fx[1]
+              + img_src[((i)*wd + (j)) * ch + k] * fx[2] + img_src[((i)*wd + (j + 1)) * ch + k] * fx[3]
+              + img_src[((i)*wd + (j + 2)) * ch + k] * fx[4];
+
+    // left edge
+    int j = 0; // 1 0 [0 1 2 ... ]
+    for(int k = 0; k < ch; k++)
+      img_tmp[(i * wd + j) * ch + k]
+          = img_src[((i)*wd + (j + 1)) * ch + k] * fx[0] + img_src[((i)*wd + (j)) * ch + k] * fx[1]
```
Entropy512: It looks like you're mirroring here - have you confirmed that this is indeed what enfuse is doing in their expand() function as opposed to the extrapolation described in Burt & Adelson 1983 (which Mertens derives their method from)?

src/iop/expfusion.c
edgardoh: If I don't do this I get very ugly halos. I can't compare it to enfuse, as I can't generate a tiff or jpg with values > 1, but probably the same happens.

src/iop/expfusion.c
```diff
@@ -0,0 +1,1268 @@
...
+}
+
+static void _build_gaussian_pyramid(const float *const img_src, const size_t wd, const size_t ht, const int ch,
+                                    dt_image_pyramid_t *const img_tmp1, dt_image_pyramid_t *const img_tmp2,
+                                    dt_pyramid_t *const pyramid_dest)
+{
+  // copy image to the finest level
+  _image_copy(img_src, wd, ht, ch, pyramid_dest->images[0].img);
+
+  const float a = 0.4;
```
Entropy512: Is there a reason you're using a different value for a than enfuse here?  Even more specifically, why does this differ from the value of a used in reconstruct_laplacian?

edgardoh:
@TurboGit , translations added, thanks.

The idea is for this to be a replacement of the fusion in the base curve.

src/iop/expfusion.c
edgardoh: I didn't look into enfuse that much, my goal is not to have the exact same result as enfuse but a good/consistent result. If enfuse give better results then we can check it.

src/iop/expfusion.c
edgardoh: This is the value used in LocalLaplacian, it gives good results. The one in the basecurve doesn't.
Again, I didn't check enfuse, but that's easy to try, provided that you know where the values are.

src/iop/expfusion.c
Entropy512: I'm fine with that.  Enfuse may not use B&A's approach either - but I think their extrapolation approach may potentially provide better results.

Instead of mirroring across a boundary, or sample-and-hold, they do (hopefully my notation isn't too unclear) for the left/top boundary:
g(-l) = 2*g(0) - g(l)
As opposed to using g(0) for sample-and-old or g(l) for mirroring

I guess you could describe it as inverted mirroring?

edgardoh: If I'm not mistaken this is the author's approach, but that doesn't mean that it cannot be improven. But I don't see why inverted mirror can be better.

src/iop/expfusion.c
Entropy512: That hints that there might still be something missing in this approach not being done by enfuse.

I'll have to run some tests with your approach when I get home tonight to see how it handles some particular cases I can't figure out why enfuse is handling well.

Specifically, hardmasks in enfuse should have a similar result to allowing unclipped values to go into the weight function (all highlights get a weight of 1 in the base exposure with hardmasks) - which should cause similar haloing, but doesn't.

This has been the #1 thing I'm trying to figure out what we're missing - why hardmasks don't cause haloing in enfuse, but adjusting the weighting function to approximate a hardmask does cause haloing in current DT implementations.

Entropy512:
Doing this as a new module solves aurelienpierre's biggest objection to fusion (wrong location in pipeline) - and I'm fine with it for 95%+ of my needs.

I do happen to have one use case where getting it working in basecurve would be preferable.  For all the ways it's mathematically wrong, turning off "preserve colors" has some interesting results for fiery sunsets.

I may continue with my pull request once Edgardo finishes this work to handle that particular use case, but at least for the short term I'm going to take a long hard look at this and try to make sure it's as good as it can be.  Solving aurelienpierre's objection of "wrong pipeline position" was on my long-term TODO list, but @edgardoh has done the heavy lifting here so prioritizing it is a good thing.

Looks like OpenCL is still on the TODO list here?

src/iop/expfusion.c
edgardoh: Not really, enfuse never get an image with values > 1, so clipping it makes the input image the same as enfuse is getting.

edgardoh:
I don't share the objection with the pipe position, but modules can be moved, so it is not a problem.

I don't get what use-case you are referring to, or what work should I do to address it.

Yes, opencl and sse are not implemented yet. I want to have some feedback & tests before jumping into that.

About the base curve, I'm making this a new module mostly because on my tests the curve don't make much of a difference, and in fact it gives extreme highlights.
Adding this to the base curve will add a lot of extra complexity, and having the curve on each exposure make it slower, so it just make sense to create a new module.
Not to speak of the backward compatibility.

src/iop/expfusion.c
Entropy512: At least the theory on inverted mirror is that the first derivative in the spatial domain remains constant across the boundary, and the second derivative is zero.

I need to dig into mertens' Matlab code to see how they handled boundary conditions.

It appears that Enfuse is performing some sort of scaling of the pixel values when expanding, but it's not entirely clear what their intent is.  They do strongly hint that they are doing extrapolation (which is probably a better term to use than "inverse mirror").

Moving from mirroring to extrapolation seemed to offer a slight improvement when I tried it, I'll take a look tonight and see if it's beneficial in your implementation.  Since you've unrolled all of the for() loops, it will be easier to tweak the boundary condition handling.  :)

src/iop/expfusion.c
Entropy512: Hmm, maybe it's a corner case where having unclipped values is doing something undesirable even when the associated weight is zero or near-zero...

It would be interesting to see what happens if you weight unclipped inputs but blend clipped inputs - this would approximate enfuse's hardmask function.

I was never particularly happy with the results when attempting to use enfuse to handle floats, but I'm wondering if their log-based approach might give better results with some tuning.  I need to reread their code, but I think if you fed enfuse linear inputs, it weighted in linear but blended in log-compressed space.

src/iop/expfusion.c
Entropy512: Interesting.  When I played with it, it didn't have significant changes.  One observation still is that this doesn't match the values used when you collapse the pyramid - that's using 0.375

Enfuse implementation is at https://github.com/jackmitch/enblend-enfuse/blob/master/src/pyramid.h - if you look at the comments, they're using a pattern prior to normalization if [1, 4, 6, 4, 1] - which corresponds to a value for a of 0.375.  The [1,4,6,4,1] pattern is what the exposure fusion module used too.

src/iop/expfusion.c
```diff
@@ -0,0 +1,1268 @@
...
+  for(size_t i = 0; i < size; i += ch)
+    img_dest[i] = img_dest[i + 1] = img_dest[i + 2] = _grey_projector(img_src + i, grey_projector, work_profile);
+}
+
+static void _exposure_fusion(const float *const img_src, const size_t wd, const size_t ht, const int ch,
+                             const int num_exposures, float *const img_dest, struct dt_iop_module_t *self,
+                             const dt_iop_order_iccprofile_info_t *const work_profile,
+                             dt_iop_fusion_data_t *const d)
+{
+  const int num_levels = floor(log2(MIN(wd, ht)));
```
Entropy512: Increasing num_levels definitely has significant benefits as far as haloing, although ideally there should be a testing goal to be approximately on par with or better than enfuse when num_levels is 8 - otherwise that's a strong indication that num_levels is masking an underlying flaw in the algorithm.

Also, I looked briefly at the tiling callbacks in the old module, ideal values of num_levels could present a very nasty challenge on memory-constrained systems by requiring excessive amounts of tile overlap.  I'm wondering if that is why num_levels was capped at 8 in the original implementation - which would limit tile overlap to 256 pixels.

src/iop/expfusion.c
Entropy512: Ooooh...  I got some pretty impressive results by weighting in gamma-corrected space and blending in log-compressed space.  Although instead of enfuse's log1p() based implementation (which is basically ln(1+x)), I used log2f(x + 1e-6).  However it appears that either boundary extrapolation doesn't play nice now, or I made mistakes in my boundary extrapolation attempt.

I'm going to push what I have as a reference later tonight to an alternative branch in my own repo so people can look at it, but then start working with your code instead.

Entropy512: https://github.com/Entropy512/darktable/tree/fusion_hax - I'll be taking a break for the rest of tonight and start taking a look at your code tomorrow.

edgardoh:
I just added the log to the colorspace combo, I use log1p() as I get better results with it.
It seems to work better on some images, specially on the highlights, but I like better the Lab on other images. More test is required.

edgardoh:
You're right about the filter, I forget to change it, and it doesn't make any difference.

I'm still not happy about the highlights, it gives too much weight to the brightest images, but I can't find why.

Entropy512:
My original theory was that the reason my rework seemed to be doing so much better in the highlights was that pixels way above the clip point were deweighted more aggressively than if you had a "clipped" pixel which would cause it to enter the weighting function closer to the optimum value.  I don't think that was the real reason though.

I suspect that if you dig into things, it may be fundamentally impossible to achieve the highlight performance you desire without inducing haloing.  I think we can probably do a bit better than enfuse, but I don't think we'll get the gorgeous highlights I've seen sometimes with the old approach I was working on without those damn halos joining the party.

You can get the highlights to look a lot better if you put another -EV exposure module after the basecurve module, but obviously you now darken your shadows down again (although not all the way back to the starting point at least).

It would have been nice to find the "smoking gun" for why the haloing was behaving so badly in the old code...  But at least now we've got what is a good solid starting point from which an attempt to improve further might be possible, other than "ugh this is still not performing nearly as well as enfuse".

I'm probably going to be hiking tomorrow so I won't be able to fiddle with some things I'd like to dig into as options for at least a day or two.

edgardoh:
That's not what I meant, but never mind, is already fixed.

src/iop/expfusion.c
```diff
@@ -1292,21 +1086,8 @@ static void _buil_weight_map(const float *const img_src, const size_t wd, const
...
-      }
-      else
-      {
-        E = _well_exposedness(rgb[0], weight_mode, exposure_optimum, exposure_width, exposure_left_cutoff, exposure_right_cutoff) *
-              _well_exposedness(rgb[1], weight_mode, exposure_optimum, exposure_width, exposure_left_cutoff, exposure_right_cutoff) *
-              _well_exposedness(rgb[2], weight_mode, exposure_optimum, exposure_width, exposure_left_cutoff, exposure_right_cutoff);
-      }
-
-      img_map[y * wd + x] = E;
+      const float lum = powf(_grey_projector(rgb, grey_projector, work_profile), exposure_optimum);
```
Entropy512: Are you sure it's a good idea to have the gamma function be the same as exposure optimum here?

If optimum is around 0.4-0.5 or so there wont' be much difference, but outside of that range things might get strange.

src/iop/expfusion.c
```diff
@@ -0,0 +1,1268 @@
...
+
+      img_map[y * wd + x] = E;
+    }
+  }
+}
+
+static float exposure_increment(const float stops, const int e)
+{
+  const float white = exp2f(-stops * (float)e);
+  const float scale = 1.0f / white;
```
Entropy512: given that exp2f(-x) = 1/exp2f(x) - why do this as opposed to removing the - from the exp2f argument and returning that?

aurelienpierre:
As I have repeated numerous times, Mertens exposure fusion algo is display-referred shit that does not belong in an unbounded raw editor. It will produce halos and fringes no matter how hard you try to hide them under the carpet.

edgardoh:
@TurboGit , this is ready for review.

Entropy512:
Some thoughts on the UI aspect of the "optimum exposure" value:

Do we want to consider different terminology for this?  What's here is consistent with the original Mertens paper but has definitely led to some confusion in past discussions of fusion.

It's also still not clear to a user that the "optimum" of 0.5 corresponds to roughly 0.18 in linear space.  I'm wondering if this slider should maybe be in units of EV below 1.0, which might make it more perceptually linear if you get away from the 0.4-0.5 range?

Nilvus:
Not sure to understand the purpose of this module regarding of modules we already have. And there's already a PR from @Entropy512. Probably his PR or this one will be one to merge and the other to close. Or both to close ?
I'm not asking to explain me what the goal of this module, it's clear but I really don't understand why having this one regarding of modules we already have ? And if he could produce halos or fringes, as @aurelienpierre said, I'm not sure it's a good thing to add. Many users are lost in the whole bunch of modules we have, so I really think we have to be careful about which new modules to add.

TurboGit:
@Nilvus : If I understand correctly some responses:
- the fusion in the base-curve is so broken that it does not work
- the PR from @Entropy512 will probably be closed
- this PR try to provide a better fusion algorithm (close to enfuse IIUC)
- at the end we may deprecate the fusion in base-cure
- later we may add more algorithm on this dedicated module

So maybe not prefect, but far better that what we have today.

Note: I have not tested that until now. I don't have lot of time (I'm in vacations :) and I plan to first look at the tag enhancements PR.

aurelienpierre:
Exposure fusion is relevant in case we have several images to merge. Here, the several images are actually the same one, with corrected exposures. As such, it falls back to a simple 1D tone-curve with some local contrast preservation, due to the multi-scale pyramid of laplacian filtering. We already have something doing that in darktable : the local contrast module, in local laplacian mode (even if it does so in Lab, and needs to be ported to XYZ at some point or anything preserving scene-linear chrominance). 

Mertens algorithm will fail around windows frames and other sharp, contrasted features, by producing halos. The reason is grounded in its algorithm, and non-linear encoding. Bad results mean bad colour models. 

We really don't need another half-baked module that fails randomly.

Entropy512:
@TurboGit As far as the general policy of not breaking edits, does that only apply to releases and not to work being done in git master?

I have some potential refinements I want to investigate, but I'm fine with this being merged as is (perhaps with some commit squashing just to make the history a bit easier to read?) if we can make some tweaks in the near future that might break edits.

Most notably - I see @edgardoh dropped support for log blending, but I think there are some use cases where it is necessary.  I don't like the idea of the module relying on clipping the input data to work, and at least my preliminary investigation into log blending was that it was the only solution that allowed input data with values above 1.0 to be fed into the algorithm without causing issues.  I won't be able to look in detail, but if @edgardoh implemented log blending without removing the input clamping, he might not have realized all of the potential advantages to operating in a space that was transformed with log2f() or log1p()

As far as the suitability of Mertens' algorithm - it, and not another approach, is the one chosen by Google for their HDR+ implementation due to its robustness.  Their algorithm (after the alignment and stacking operation) effectively does what we're doing here - compressing the dynamic range of a single RAW dataset.  (They are not bracketing, they are stacking a set of same-exposure bursts to average out noise.)

Interestingly, Tim Brooks' implementation used an iterative approach (always processing pairs of images, and then using the output of a fused pair as the input to the next iteration) - this is definitely better as far as memory consumption, TBD what impact it would have on overall processing time.

Google indicates that their implementation in https://hdrplusdata.org/hdrplus.pdf only operates on a single pair (e.g. base exposure and one pushed exposure), or at least did at the time of publishing.  Current implementations in the Pixel family may be (and in fact almost surely are) using more iterations or more exposures, since they are no longer subject to the constraints of operating on the CPU of a Snap810.

Nilvus:
> @Nilvus : If I understand correctly some responses:
> 
>     * the fusion in the base-curve is so broken that it does not work
> 
>     * the PR from @Entropy512 will probably be closed
> 
>     * this PR try to provide a better fusion algorithm (close to enfuse IIUC)
> 
>     * at the end we may deprecate the fusion in base-cure
> 
>     * later we may add more algorithm on this dedicated module
> 
> 
> So maybe not prefect, but far better that what we have today.
> 
> Note: I have not tested that until now. I don't have lot of time (I'm in vacations :) and I plan to first look at the tag enhancements PR.
> 
> @edgardoh : correct me if I'm wrong.

Thanks for your comment. And as I consider base curve as an obsolete module and filmic as a great replacement one (and much more than just a "base curve" of course), I so still doesn't see the need to this module, especially if it brings some not wanted side effects.

So, I don't say that we don't have to merge this module. I just say that we have to be sure the modules really improve a lot what's already here and be sure to have a module that doesn't provide bad side-effects or the less ones is possible. So if this module if not on these requirements (that same for the @Entropy512 PR), I think we have to wait before merging it.

src/iop/expfusion.c
```diff
@@ -0,0 +1,1268 @@
...
+  dt_bauhaus_slider_set(g->sl_num_exposures, p->num_exposures);
+  dt_bauhaus_slider_set(g->sl_exposure_stops, p->exposure_stops);
+  dt_bauhaus_slider_set(g->sl_exposure_optimum, p->exposure_optimum);
+  dt_bauhaus_slider_set(g->sl_exposure_width, p->exposure_width);
+  dt_bauhaus_combobox_set(g->cmb_weight_mode, p->weight_mode);
+  dt_bauhaus_combobox_set(g->cmb_grey_projector, p->grey_projector);
+  dt_bauhaus_combobox_set(g->cmb_fusion_colorspace, p->fusion_colorspace);
+  dt_bauhaus_combobox_set(g->cmb_fusion_grey_projector, p->fusion_grey_projector - 1);
+  dt_bauhaus_slider_set(g->sl_exposure_left_cutoff, p->exposure_left_cutoff * 100.f);
+  dt_bauhaus_slider_set(g->sl_exposure_right_cutoff, p->exposure_right_cutoff * 100.f);
```
Entropy512: There should be some method of disabling the exposure cutoffs.  (Right now, 100% effectively disables it, but this will not continue to be the case if the algorithm is fixed to handle data >1.0f)

src/iop/expfusion.c
```diff
@@ -0,0 +1,1268 @@
...
+  self->gui_data = malloc(sizeof(dt_iop_fusion_gui_data_t));
+  dt_iop_fusion_gui_data_t *g = (dt_iop_fusion_gui_data_t *)self->gui_data;
+  dt_iop_fusion_params_t *p = (dt_iop_fusion_params_t *)self->params;
+
+  self->widget = gtk_box_new(GTK_ORIENTATION_VERTICAL, DT_BAUHAUS_SPACE);
+  dt_gui_add_help_link(self->widget, dt_get_help_url(self->op));
+
+  g->sl_num_exposures = dt_bauhaus_slider_new_with_range(self, 2, 5, 1, p->num_exposures, 0);
+  gtk_widget_set_tooltip_text(g->sl_num_exposures, _("number of different exposures to fuse"));
+  dt_bauhaus_slider_set_format(g->sl_num_exposures, "%.2fEV");
```
Entropy512: The number of different exposures is an integer - it shouldn't be EV and should not be %.2f since it can only be integers

src/iop/expfusion.c
Entropy512: Also - you're only clamping the input for the pushed exposures.  If the original input has values over 1.0 - strange things may happen.  Haven't tested, as I've said elsewhere it's my opinion we shouldn't break when values are over 1.0f

Entropy512:
As far as not merging a module because there's the possibility that a user just MIGHT choose a combination of slider settings that has an undesirable result - well then nearly every module in darktable should be removed.

(That said, I'm not sure if allowing more than +2EV shift per step is a good idea.  Especially now that the module supports up to 5 exposures.)

aurelienpierre:
@sobotka please help me, I have exhausted my patience here.

sobotka:
I am struggling to fathom how a trivial merge is being complexities here?

1.  Take a given sensor range, cut the upper and lower arbitrary percentage where the sensor skews from Luther-Ives
2. Multiply the values to fix it in the appropriate exposure range
3. Merge into the base image.

This isn’t nearly as complex as folks are making it out to be. You can do it by hand in a compositor, ghosting aside.

https://github.com/colour-science/colour-hdri

TurboGit:
Given the current discussion I don't think we are ready or we should merge this. A new module which does not a great job is not that appealing. We already have a broken exposure fusion in the basecurve.

What would be nice is to have the exposure fusion works on multiple image with different exposure. Probably not a module in the darkroom but an enhanced and better algorithm for the HDR merging already supported in the lighttable. This would be great as the HDR code in lighttable does not work very well to my experience.

Do you think something like this could be done reusing the code in this PR?

edgardoh:
@TurboGit , I can't believe this, did you even try it?

Entropy512:
@TurboGit Did you actually test this out and find an issue?  (if so, please arrange for us to have a copy of your RAW test case)

Or did you just accept the word of @aurelienpierre that:
1)  A peer reviewed acadamic paper published in the ACM journal is, in his words, "shit" and "garbage"
2)  The 394 academic papers which cited Mertens were all written by morons - https://scholar.google.com/scholar?uact=5&um=1&ie=UTF-8&lr&cites=14707502593458229605
3)  Anyone who positively reviewed the camera functionality of the Google Pixel family of devices (which depends on the Mertens algorithm for tonemapping in its default operating mode, per https://hdrplusdata.org/hdrplus.pdf ) must have been on hallucinogenic drugs?

If it is his position that any filter which could possibly, under any circumstance, cause haloing is fundamentally broken, and you agree with this position - why is the local contrast module present?

localcontrast_halo

Yes, using the Mertens fusion technique to implement a tonemapping operator by creating synthetic exposures is not what it was originally designed to do, but it does handle that use case well as evidenced by the positive results Google has achieved by making it their default operational mode.

(One observation - Google's implementation solely fuses a greyscale luminance representation of the image, and then scales the original RGB values using the results - this is likely desirable for the implementation here for a variety of reasons, not just computational complexity issues.)

As to fusion of multiple independent images - generating a bunch of synthetic images using the history clone and duplication of the exposure module, followed by using the enfuse lua script does achieve this, but its UX is clunky, slow, does not allow for finetuning the parameters with a realtime preview, and clutters the lighttable with multiple exposure-shifted copies of each image being worked on.

sobotka:
@Entropy512 I'm pretty sure you aren't reading between the lines here. The DT folks are trying to _fix_ DT's broken pipeline. That means not adding more problematic modules that deviate from a scene linear approach.

Scene linear isn't nearly as complex as some folks would make it out to be, and merging exposures under such a model is vastly less complex than what many would lead one to believe. Dare I say one can merge four exposures by hand and understand the principles vastly faster than any army of paper citations.

Entropy512:
> @Entropy512 I'm pretty sure you aren't reading between the lines here. The DT folks are trying to _fix_ DT's broken pipeline. That means not adding more problematic modules that deviate from a scene linear approach.
> 
> Scene linear isn't nearly as complex as some folks would make it out to be, and merging exposures under such a model is vastly less complex than what many would lead one to believe. Dare I say one can merge four exposures by hand and understand the principles vastly faster than any army of paper citations.
Really - you can perform a local tonemapping operator by hand?  Show me.

> 
> Hopefully more of the legacy modules will be removed, and a cleaner pipeline established.
> 
> > @sobotka - did you perhaps miss that the goal here is to perform local tonemapping, which is not covered or discussed at all in the link you provided (only global tonemapping is)? The method you describe does not address tonemapping in any way, shape, or form.
> 
> You missed the entire issue of performing manipulations on non-scene linear data; _It doesn't work_. You don't apply a transfer function until the very last step of image manipulation, after all manipulations are complete. This is precisely why OpenColorIO exists.

So, local contrast enhancement is absolutely horrible - except in every instance where @aurelienpierre uses it?  You do realize that nearly every example I've seen him post is using the local contrast module in the pipeline?  Despite it being, by his standards, "garbage" and "shit" since it can potentially cause haloing if abused?  Hypocritical much?

sobotka:
>Really - you can perform a local tonemapping operator by hand? Show me.

If you read what I typed, I explained how to do a merge by hand. Yes it's trivial. Yes there is even an example at Pixls done by someone. It's not rocket science.

A transfer function has no place mid-pipeline. Full stop.

>So, local contrast enhancement is absolutely horrible - except in every instance where @aurelienpierre uses it?

I don't know what "local contrast" is, but I do know that a transfer function belongs at the end of the entire chain of manipulations. This is where the "module by module" approach falls apart, but that's not my issue nor concern.

>As to "You don't apply a transfer function until the very last step of image manipulation, after all manipulations are complete. " - you mean exactly what I'm doing?

If you think that's what you are doing in the context of the entire pipeline, we are living on different planets.

Entropy512:
> > As to "You don't apply a transfer function until the very last step of image manipulation, after all manipulations are complete. " - you mean exactly what I'm doing?
> 
> If you think that's what you are doing in the context of the entire pipeline, we are living on different planets.

So, you keep on saying how operations should all be operating on linear data - but now you're insisting that this transfer curve happens after the final conversion to sRGB?  Or did you not bother to look at the screenshot and notice that the ONLY operation after this tonemapping operator was the transform to sRGB?

sobotka:
>So, you keep on saying how operations should all be operating on linear data - but now you're insisting that this transfer curve happens after the final conversion to sRGB? Or did you not bother to look at the screenshot and notice that the ONLY operation after this tonemapping operator was the transform to sRGB?

If you look at how a library like OpenColorIO functions, or examine something like Nuke, Blender, Fusion, etc. and how the entirety of the camera rendering transform[1] operates, this is exactly what happens. The individual doing the manipulation is performing all of the manipulations on the scene linear data, which is some extremely small value to some extremely large value, and viewing the output _under_ the camera rendering transform, and even possibly a customized look per shot etc.

This way, the data remains entirely scene linear through the entire pipeline, and in fact is never actually "baked" into the data until output.

[1] _Camera rendering transform_ here is used to loosely wrap up a transfer function and any sweeteners that may be applied. This may include gamut mapping along both chroma and intensity, which is tightly coupled to transfer function application. Given there are other facets located at this point, the looser term of _camera rendering transform_ is applied. If for example, you are using a complex film emulation transform, it too would fall under such a banner, and may be vastly more complex than a simple transfer function.

Entropy512:
So, forgive me for being confused, since what you're saying is in direct conflict with everything Pierre has been saying.  (e.g. my example where I dared to touch linear data in any way prior to the linear->sRGB conversion was too early in the pipeline, even though it was the last step prior to doing so)...

Why is this OK?
toneeq

sobotka:
Take two photos that someone is seeking to merge. The merge _must_ result in scene referred data, otherwise a convolution such as a lens blur, would fail. Rotation? It would fail as well. Cross expose with another image? Yes, that fails too. Spatial upsampling or downsampling? You guessed it. Applying a lens glare such as via a rear net? Same same.

_Any_ manipulation on anything other than scene referred data will fail, hence the whole point of a merge is that it must result in scene linear output in the reference working space of the image maker’s choosing. By applying _any_ nonlinear adjustment to the camera values, you end up discarding the critical information in the camera sensor encode.

Entropy512:
So, I'm fairly certain at this point that you've come here to shit on this pull request without the slightest goddamn understanding of what the intended design goals or usecases are.

You still didn't answer my question:  Why is it OK that @aurelienpierre tonemapped an HDR representation of an image to an LDR output by using a global tonemapping operator followed by a local contrast enhancement (which has been proven to have the exact same potential risks for abuse that he is using to declare this module to be unfit), but not OK to tonemap an HDR internal representation down to an LDR output using a local tonemapping operator that preserves the local contrast in the first place?

TurboGit:
@edgardoh : No I didn't test it. My decision was made after reading the comments here. It seems that this module raises lot of questions and seems to not integrate well with the new pixel-pipe.

That being said, what about my proposal to have a proper exposure fusion from multiple images in the lighttable? The current HDR code is quite broken to me as I have never had proper result out of it.

Entropy512:
@TurboGit Given the use case example provided in my screenshot, how is this less well integrated in the pixel pipe than any other module which applies a nonlinear transform to the data, such as filmic or the local contrast module?

It seems that assumptions are being made about the design intent and use cases for this module (e.g. a tonemapping operator) that are not the actual design intent and use cases of the module.

What you are proposing sounds exactly to me like the sort of use cases that @sobotka is crusading against.  (use cases which are mostly met by the enfuse lua script, and given the large number of manual operations involved in what I think you're proposing, there's likely not much performance benefit of doing it internally.)

When you reference the current HDR code, do you mean the exposure fusion code in basecurve?  (yeah, that's definitely broken, which is why I started my work there...  In 90% of my use cases, tonemapping is all I need to be happy, so despite basecurve being "too early" in the pipeline, it was still at the end of the pipe in most of my tests.  But moving tonemapping via fusion to a separate module makes a lot of sense, and that permits the module to default to being much later in the pipeline with other tonemapping modules.  Which actually is a potential code review item - given the intended use cases of this module, shouldn't it be in IOP_GROUP_TONE and not IOP_GROUP_BASIC?  Have people been screaming about this module merely because of that mistake, instead of leaving a review comment on that line?)

I'm still kind of scratching my head as to what happened with the original exposure fusion implementation - the blog makes it pretty clear it is intended to be a variation on the Mertens/enfuse approach, but it seems like no one asked any questions as to why the results were so vastly different from the implementations they were intended to be based on?  Also it seems from commit history that a lot of twiddling with hardcoded values was done, that should have been a gigantic red flag that the value should have been a user-exposed setting?

Or do you mean the lighttable HDR merging code I've seen references to which appears to be attempting to achieve the use case @sobotka describes and appears to be intended to handle algorithm category 1 above?  I haven't played with this much yet other than seeing references to the function existing and creating a merged DNG, although merge and align algorithm experimentation is next on my TODO list after tonemapping.

TurboGit:
@edgardoh : I have seen that you've close all your PR, I hope that this is not due to our disagreement about this PR. You've made quite a lot of good feature and fixes in dt and probably one of the most productive for next major release. It would be a shame to see you leaving just for this. Disagreement is part of the life, we cannot always be 100% ok on everything, that's life but I want to assure you that there is really nothing personal and that I really do appreciate your investment. Hope this clarify the situation.

edgardoh:
@TurboGit , we didn't have a disagreement, you didn't discussed this with me and you didn't even try it. Hope this clarify the situation.

TurboGit:
@edgardoh : I'm nobody, the disagreement is between the dt team and coders (you in this matter). I'm the current integrator, ok, I have almost zero expertise on image processing, I'm an IT guy. I'm trying my best to make dt stay living. I'm reading messages from all parties and try to take decisions (hard, I just don't like this and all the time I'm making someone sad). But I do hope that we can live with that. Until now I hope you find my handling of dt not that bad?

I didn't try it because I'm in vacation, I try to keep in eye on the tag PR for the moment and have lot lot of more time for other things.

Entropy512:
I agree with @edgardoh - judgement was passed upon this module based on false assumptions made about its design goals, its use case intent, and false claims that it fundamentally would result in problematic behavior in all cases (as opposed to extreme corner cases) by someone who routinely uses a module that can be proven to also fail in corner cases in the exact same manner that they claim indicates a broken module which has no business in darktable.  No attempt was made to verify these claims with actual testing and evaluation.

While I think @edgardoh did jump the gun on kicking off final code review (see the various comments I've made regarding a few small details), the module was scrapped with no discussion whatsoever because of one person's personal bias against Mertens and comments from an outside party who came in with obviously false assumptions about the intended uses of this module.

We were then suggested to implement the use case said third party was crusading against.

aurelienpierre:
I don't have biases. I have dealt for 7 years with the shit stack that is display-referred encoding and Lab space, and tried every combination of plugins until I finally understood that the problem is not the algo but the colour space itself. 

I'm fed up with this empiric non-sense. Try it, then it works, then what, it will work all the time ? There are no extreme corner cases, they are proofs that the algo does not work conceptually. I don't care if it works 90% of the time, as long as people don't push it too far. You don't build tools that work 90% of the time, when it's possible to have them work all the time. But for this, you have go check the theory, and stop with this shitty habit of hacking this and that, and wiring bits of award-winning research teams without understanding that the problem they try to solve does not apply here.

I would like we fix what's broken in dt instead of keeping piling up shit. Lab and display-referred algos are what's wrong. And this belongs to the same team.

1. blending an image on top of itself is no blending, it's a simple transfer function. Using a fake multi-exposure setup is overkill and slow for no good reason.
2. that transfer function need, ideally, to preserve chrominance in some way, aka the relationship between RGB code values because **RGB code values are not numbers, they are radiometric ratios and these ratios need to be kept in sync otherwise you destroy colour** (how do you ensure this when you gamma up the 3 channels independantly ??? **YOU DON'T**). Merten's algo doesn't preserve radiometric ratios, and, worse, blends in gamma encoded around magic values believed to be in the [0 ; 1] range (what if values go above ? remember, energy is unbounded, and sensor clipping is a technicality, not a concept).
3. that transfer function need to go as late as possible in the pipe, for **well-behaved** scene-linear algos to be performed before
4. if you still need to perform non-linear tone mapping with local contrast preservation and detection, the local laplacian module does just that, and might need to be tweaked.

Now, let answer @Entropy512 attempts to make shit look acceptable.

> Also, how is attempting to preserve local contrast using a local tonemapping operator (which always has a risk of haloing if abused) worse than damaging local contrast and then attempting to synthetically bring it back after it has been lost - using an algorithm that as can be shown above, can also halo if abused?

Use the local contrast in local laplacian mode, not the bilateral model (you carefully forgot to show the parameters you used, which are not the one I use in my demos). This one is proven to be halo-free, but at the expense of being super slow.

https://www.di.ens.fr/willow/pdfscurrent/Aubry14tog.pdf

> So, forgive me for being confused, since what you're saying is in direct conflict with everything Pierre has been saying. (e.g. my example where I dared to touch linear data in any way prior to the final linear->sRGB conversion was too early in the pipeline, even though it was the last step prior to doing so)...
> 
> Why is this OK?

This is ok because we do a linear operation (exposure compensation) on top of linear data. The exposure factors just change depending the exposure of the pixel. So, radiometric ratios preserved. It's also ok because it comes before the input profile, which is optimized by least squares methods to minimize the RGB -> XYZ matching error around mid-tones and low-saturation colours, so we gently push shadows to that and reduce the conversion error.

> You still didn't answer my question: Why is it OK that @aurelienpierre tonemapped an HDR representation of an image to an LDR output by using a global tonemapping operator followed by a local contrast enhancement (which has been proven to have the exact same potential risks for abuse that he is using to declare this module to be unfit), but not OK to tonemap an HDR internal representation down to an LDR output using a local tonemapping operator that preserves the local contrast in the first place?

The local laplacian algo in the local contrast module is tweaked for display-referred and applied in Lab, so, surprisingly, it behaves better after the filmic emulation. Tweak it to work in XYZ and it will go before.

> When you reference the current HDR code, do you mean the exposure fusion code in basecurve? (yeah, that's definitely broken, which is why I started my work there... In 90% of my use cases, tonemapping is all I need to be happy, so despite basecurve being "too early" in the pipeline, it was still at the end of the pipe in most of my tests. But moving tonemapping via fusion to a separate module makes a lot of sense, and that permits the module to default to being much later in the pipeline with other tonemapping modules. Which actually is a potential code review item - given the intended use cases of this module, shouldn't it be in IOP_GROUP_TONE and not IOP_GROUP_BASIC? Have people been screaming about this module merely because of that mistake, instead of leaving a review comment on that line?)

There is **no need for HDR code**. The property of linear editing is HDR or LDR use the same technics and algorithms (the joys of scaling). The very fact you need a special code path to deal with HDR means your colour model doesn't scale, and is inherently flawed.

**HDR photography is not a thing** (unless it's to merge several shots). **It's the consequence of poor colour models**.

PR #2848, opened 2019-08-11
tags: prevent completion windows from apearing when entry is cleared.

TurboGit:


phweyland:
Hi Pascal,
I'm closed to open a PR with a lot of changes on this.
In particular I've removed the dependency of entry on db, so now it's fast and completion is more a burden (on top of the window) than an help.
I'll integrate all your minor changes before raising the PR. Can you please wait a moment before merging this one ?

TurboGit:
Sure, I'll wait.

phweyland:
I've checked I have the problem too.
This a question of focus. In fact when we start to edit entry, the focus is given to entry and all dt shortcuts are deactivated. As is we need to put the focus elsewhere after starting edition to get back Ctrl-T active. Not sure there is a real solution to this.

TurboGit:
This seems to be fixed now with last tagging rework.

PR #2850, opened 2019-08-12
Tagging improvements

phweyland:
Some improvements for tagging:
- toggle button to mask darktable tags in attached tags window to save visual room
- toggle button to present the short tag name instead of the full tag name (for both attached and dictionary tags windows)
- toggle button to sort by name or by number (for both attached and dictionary tags windows)
- Ctrl-scroll to resize both the attached and dictionary tags windows
- added synonyms => exported if "plugins/lighttable/export/export_tag_synonyms" set
- added category flag. => the tag is not to be exported but still can have synonyms
- added private flag => the tag is exported if "plugins/lighttable/export/export_private_tags" set
- menu entry `rename tag` changed to menu `edit tag` to integrate the 2 flags and the synonyms with the tag name
- overlay tooltip to show the tag's associated data if any to avoid to have to edit it to see them
- add menu entry `delete branch` (deleting 1000 tags one by one can be tedious)
- suggestions are now related to already attached tags on selected images
- filtering doesn't involve any more the db, speeding up a lot the search of tags
- filtering takes in account the synonyms
- display of tags inside image information module
- import and export tags updated to take in account the synonyms
- export updated to apply the private and synonyms choices
- some display speed optimization

image

image

What I've not done:

- integrate "plugins/lighttable/export/export_tag_synonyms" and "plugins/lighttable/export/export_private_tags" in export module. I think this should be done but the module is already big...
- apply the effect of the previous flags on piwigo & flicker
- optimize `rename path` to keep the tree opened

### DIFF
--- a/src/common/database.c
+++ b/src/common/database.c
@@ -1410,7 +1410,8 @@ static void _create_memory_schema(dt_database_t *db)
                            "(tmpid INTEGER PRIMARY KEY, id INTEGER UNIQUE ON CONFLICT IGNORE, count INTEGER)",
                NULL, NULL, NULL);
   sqlite3_exec(db->handle, "CREATE TABLE memory.similar_tags (tagid INTEGER)", NULL, NULL, NULL);
-  sqlite3_exec(
+  sqlite3_exec(db->handle, "CREATE TABLE memory.darktable_tags (tagid INTEGER)", NULL, NULL, NULL);
+    sqlite3_exec(
       db->handle,
       "CREATE TABLE memory.history (imgid INTEGER, num INTEGER, module INTEGER, "
       "operation VARCHAR(256) UNIQUE ON CONFLICT REPLACE, op_params BLOB, enabled INTEGER, "
--- a/src/common/database.c
+++ b/src/common/database.c
@@ -1228,8 +1228,24 @@ static int _upgrade_data_schema_step(dt_database_t *db, int version)
   }
   else if(version == 2)
   {
-    TRY_EXEC("ALTER TABLE data.tags RENAME COLUMN description TO synonyms;",
-             "[init] can't change tags column name from description to synonyms\n");
+    //    With sqlite above or equal to 3.25.0 RENAME COLUMN can be used instead of the following code
+    //    TRY_EXEC("ALTER TABLE data.tags RENAME COLUMN description TO synonyms;",
+    //             "[init] can't change tags column name from description to synonyms\n");
+
+    TRY_EXEC("ALTER TABLE data.tags RENAME TO tmp_tags",  "[init] can't rename table tags\n");
+
+    TRY_EXEC("CREATE TABLE data.tags (id INTEGER PRIMARY KEY, name VARCHAR, "
+             "synonyms VARCHAR, flags INTEGER)",
+             "[init] can't create new tags table\n");
+
+    TRY_EXEC("INSERT INTO data.tags (id, name, synonyms, flags) SELECT id, name, description, flags "
+             "FROM tmp_tags",
+             "[init] can't populate tags table from tmp_tags\n");
+
+    TRY_EXEC("DROP TABLE tmp_tags", "[init] can't delete table tmp_tags\n");
+
+    TRY_EXEC("CREATE UNIQUE INDEX data.tags_name_idx ON tags (name)",
+             "[init] can't create tags_name_idx on tags table\n");
 
     sqlite3_exec(db->handle, "COMMIT", NULL, NULL, NULL);
 
@@ -1376,7 +1392,7 @@ static void _create_data_schema(dt_database_t *db)
   sqlite3_step(stmt);
   sqlite3_finalize(stmt);
   ////////////////////////////// tags
-  sqlite3_exec(db->handle, "CREATE TABLE data.tags (id INTEGER PRIMARY KEY, name VARCHAR, icon BLOB, "
+  sqlite3_exec(db->handle, "CREATE TABLE data.tags (id INTEGER PRIMARY KEY, name VARCHAR, "
                            "synonyms VARCHAR, flags INTEGER)", NULL, NULL, NULL);
   sqlite3_exec(db->handle, "CREATE UNIQUE INDEX data.tags_name_idx ON tags (name)", NULL, NULL, NULL);
   ////////////////////////////// styles
--- a/src/common/database.c
+++ b/src/common/database.c
@@ -40,7 +40,7 @@
 // whenever _create_*_schema() gets changed you HAVE to bump this version and add an update path to
 // _upgrade_*_schema_step()!
 #define CURRENT_DATABASE_VERSION_LIBRARY 19
-#define CURRENT_DATABASE_VERSION_DATA 2
+#define CURRENT_DATABASE_VERSION_DATA 3
 
 typedef struct dt_database_t
 {
@@ -1226,6 +1226,15 @@ static int _upgrade_data_schema_step(dt_database_t *db, int version)
 
     new_version = 2;
   }
+  else if(version == 2)
+  {
+    TRY_EXEC("ALTER TABLE data.tags RENAME COLUMN description TO synonyms;",
+             "[init] can't change tags column name from description to synonyms\n");
+
+    sqlite3_exec(db->handle, "COMMIT", NULL, NULL, NULL);
+
+    new_version = 3;
+  }
   else
     new_version = version; // should be the fallback so that calling code sees that we are in an infinite loop
 
@@ -1368,7 +1377,7 @@ static void _create_data_schema(dt_database_t *db)
   sqlite3_finalize(stmt);
   ////////////////////////////// tags
   sqlite3_exec(db->handle, "CREATE TABLE data.tags (id INTEGER PRIMARY KEY, name VARCHAR, icon BLOB, "
-                           "description VARCHAR, flags INTEGER)", NULL, NULL, NULL);
+                           "synonyms VARCHAR, flags INTEGER)", NULL, NULL, NULL);
   sqlite3_exec(db->handle, "CREATE UNIQUE INDEX data.tags_name_idx ON tags (name)", NULL, NULL, NULL);
   ////////////////////////////// styles
   sqlite3_exec(db->handle, "CREATE TABLE data.styles (id INTEGER, name VARCHAR, description VARCHAR)",
@@ -1411,7 +1420,7 @@ static void _create_memory_schema(dt_database_t *db)
                NULL, NULL, NULL);
   sqlite3_exec(db->handle, "CREATE TABLE memory.similar_tags (tagid INTEGER)", NULL, NULL, NULL);
   sqlite3_exec(db->handle, "CREATE TABLE memory.darktable_tags (tagid INTEGER)", NULL, NULL, NULL);
-    sqlite3_exec(
+  sqlite3_exec(
       db->handle,
       "CREATE TABLE memory.history (imgid INTEGER, num INTEGER, module INTEGER, "
       "operation VARCHAR(256) UNIQUE ON CONFLICT REPLACE, op_params BLOB, enabled INTEGER, "
--- a/src/common/tags.c
+++ b/src/common/tags.c
@@ -552,7 +552,7 @@ uint32_t dt_tag_get_attached(gint imgid, GList **result, gboolean ignore_dt_tags
   if(imgid > 0)
   {
     char query[1024] = { 0 };
-    snprintf(query, sizeof(query), "SELECT DISTINCT T.id, T.name, T.flags, 1 AS inb "
+    snprintf(query, sizeof(query), "SELECT DISTINCT T.id, T.name, T.flags, 1 AS inb, T.description "
                                    "FROM main.tagged_images AS I "
                                    "JOIN data.tags T on T.id = I.tagid "
                                    "WHERE I.imgid = %d %s ORDER BY T.name",
@@ -563,7 +563,7 @@ uint32_t dt_tag_get_attached(gint imgid, GList **result, gboolean ignore_dt_tags
   {
     if(ignore_dt_tags)
       DT_DEBUG_SQLITE3_PREPARE_V2(dt_database_get(darktable.db),
-              "SELECT DISTINCT I.tagid, T.name, T.flags, COUNT(DISTINCT S.imgid) AS inb "
+              "SELECT DISTINCT I.tagid, T.name, T.flags, COUNT(DISTINCT S.imgid) AS inb, T.description "
               "FROM main.selected_images AS S "
               "LEFT JOIN main.tagged_images AS I ON I.imgid = S.imgid "
               "LEFT JOIN data.tags AS T ON T.id = I.tagid "
@@ -573,7 +573,7 @@ uint32_t dt_tag_get_attached(gint imgid, GList **result, gboolean ignore_dt_tags
               -1, &stmt, NULL);
     else
       DT_DEBUG_SQLITE3_PREPARE_V2(dt_database_get(darktable.db),
-              "SELECT DISTINCT I.tagid, T.name, T.flags, COUNT(DISTINCT S.imgid) AS inb "
+              "SELECT DISTINCT I.tagid, T.name, T.flags, COUNT(DISTINCT S.imgid) AS inb, T.description "
               "FROM main.selected_images AS S "
               "LEFT JOIN main.tagged_images AS I ON I.imgid = S.imgid "
               "LEFT JOIN data.tags AS T ON T.id = I.tagid "
@@ -600,6 +600,7 @@ uint32_t dt_tag_get_attached(gint imgid, GList **result, gboolean ignore_dt_tags
     // 1: tag attached on some selected images
     // 2: tag attached on all selected images
     t->select = (nb_selected == 0) ? 0 : (imgnb == nb_selected) ? 2 : (imgnb == 0) ? 0 : 1;
+    t->synonym = g_strdup((char *)sqlite3_column_text(stmt, 4));
     *result = g_list_append(*result, t);
     count++;
   }
@@ -614,7 +615,7 @@ uint32_t dt_tag_get_attached_export(gint imgid, GList **result)
   if(imgid > 0)
   {
     DT_DEBUG_SQLITE3_PREPARE_V2(dt_database_get(darktable.db),
-            "SELECT DISTINCT T.id, T.name, T.flags, S.selected FROM data.tags AS T "
+            "SELECT DISTINCT T.id, T.name, T.flags, T.description, S.selected FROM data.tags AS T "
             "JOIN (SELECT DISTINCT I.tagid, T.name "
               "FROM main.tagged_images AS I  "
               "LEFT JOIN data.tags AS T ON T.id = I.tagid "
@@ -629,7 +630,7 @@ uint32_t dt_tag_get_attached_export(gint imgid, GList **result)
   else
   {
     DT_DEBUG_SQLITE3_PREPARE_V2(dt_database_get(darktable.db),
-            "SELECT DISTINCT T.id, T.name, T.flags, S.selected FROM data.tags AS T "
+            "SELECT DISTINCT T.id, T.name, T.flags, T.description, S.selected FROM data.tags AS T "
             "JOIN (SELECT DISTINCT I.tagid, T.name "
               "FROM main.selected_images AS S "
               "LEFT JOIN main.tagged_images AS I ON I.imgid = S.imgid "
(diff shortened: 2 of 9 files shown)


phweyland:
The completion change I've copied from your PR causes an access violation when pushing new button.
I'll have to remove that. Otherwise I've also seen some regression on the export. I'll do that tomorrow with the travis issue.

TurboGit:
> The completion change I've copied from your PR causes an access violation when pushing new button.

Really strange given what my PR does. But no problem, I'll rework this after this is merged.

TurboGit:
Just forgot to say that all the enhancements sound really good!

phweyland:
I've removed the completion change because I got this when I tired to create a new tag (button `new`,  I have not tried to understand why):

```
darktable.exe caused an Access Violation at location 000000000107896D in module libgtk-3-0.dll Reading from location 0000000100000000.

AddrPC           Params
000000000107896D 0000000013824080 0000000000000000 00000000126233E0  libgtk-3-0.dll!gtk_entry_set_completion
0000000014BD5B6C 00000A59CF03CD4B 0000000017D45210 0000000000000000  libtagging.dll! ??   [C:/msys64/home/philippe/dt-phi/src/libs/tagging.c @ 890]
   888: 
   889:   /** clear input box */
>  890:   gtk_entry_set_completion(GTK_ENTRY(d->entry), NULL);
   891:   gtk_entry_set_text(d->entry, "");
   892:   gtk_entry_set_completion(GTK_ENTRY(d->entry), d->completion);
```

After this PR, the next work would be something like that:

- attached window popup menu to attach the selected tag to all selected images
- exportation flags config (today works with manual setting in darktablerc)
- mass tag delete (today awfully slow one by one process on 3 tables)
- enable (if possible) tagging preset

NB: Export & Import should cooperate smoothly. In particular, it should be easy to export one's current tags, add the synonyms in the exported text file and import back this file to get the new synonyms without losing the previous tags and their attachments.

phweyland:
from one minute to delete 1000 tags to one second to delete 8000 tags (both without attachment)

phweyland:
I think I'm done now. Please review when you can. Thanks

phweyland:
Normally the choice of tags which should be exported with exported image, should de defined in export module (or other variations like flicker, ...).
Looking at the code, it seems very difficult to implement such choices. Metadata in general will face the same issue. We have this kind of code on various levels and in tens of modules:

```
  /* export image to file */
  if(dt_imageio_export(imgid, filename, format, fdata, high_quality, upscale, TRUE, icc_type, icc_filename,
                       icc_intent, self, sdata, num, total) != 0)
```

On the other side using configurations turns the code very simple. But the user has to change them every time he exports for a different usage. 

What do you think ? What will be applied to metadata ?

src/common/tags.c
```diff
@@ -922,27 +1121,154 @@ uint32_t dt_tag_get_with_usage(const gchar *keyword, GList **result)
...
+  {
+    synonyms = dt_util_dstrcat(synonyms, ", %s", synonym);
+  }
+  else
+  {
+    synonyms = g_strdup(synonym);
+  }
+  sqlite3_stmt *stmt;
+  DT_DEBUG_SQLITE3_PREPARE_V2(dt_database_get(darktable.db),
+                              "UPDATE data.tags SET description = ?2 WHERE id = ?1 ",
```
TurboGit: description = ?2
and ?2 is synonyms
and synonyms can be "first, second".

That cannot work or I'm missing something.

src/libs/tagging.c
```diff
@@ -102,7 +99,7 @@ const char **views(dt_lib_module_t *self)
   static const char *v1[] = {"lighttable", "darkroom", "map", "tethering", NULL};
   static const char *v2[] = {"lighttable", "map", "tethering", NULL};
 
-  if(dt_conf_get_bool("plugins/darkroom/tagging/visible"))
+  if(dt_conf_get_bool("plugins/lighttable/tagging/visible"))
```
TurboGit: Looks a wrong change. Can you explain? lt only view is v2.

src/libs/tagging.c
```diff
@@ -506,27 +593,113 @@ static void update_sel_on_tree(GtkTreeModel *model)
...
+  else
+  {
+    gtk_list_store_remove(GTK_LIST_STORE(model), iter);
+  }
+}
+
+// delete a branch of the tag tree
+static void delete_tree_path(GtkTreeModel *model, GtkTreeIter *iter, gboolean root, gboolean tree)
+{
+  if (tree)
```
TurboGit: some comments on the different paths would help here I think.

src/libs/tagging.c
```diff
@@ -728,44 +1012,75 @@ static void entry_activated(GtkButton *button, dt_lib_module_t *self)
...
   if(!gtk_tree_selection_get_selected(selection, &model, &iter)) return;
-  gtk_tree_model_get(model, &iter, DT_LIB_TAGGING_COL_ID, &tagid, -1);
 
-  // First check how many images are affected by the remove
-  const int count = dt_tag_remove(tagid, FALSE);
-  if(count > 0 && dt_conf_get_bool("plugins/lighttable/tagging/ask_before_delete_tag"))
+  gtk_tree_model_get(model, &iter, DT_LIB_TAGGING_COL_PATH, &tagname,
+          DT_LIB_TAGGING_COL_ID, &tagid, -1);
+  if (!tagid) return;
+  guint img_count = dt_tag_remove(tagid, FALSE);
```

src/libs/tagging.c
```diff
@@ -1078,7 +1542,7 @@ static void view_popup_menu_rename_path(GtkWidget *menuitem, dt_lib_module_t *se
         dt_tag_rename(((dt_tag_t *)taglist->data)->id, new_tagname);
         g_free(new_tagname);
       }
-
+// TODO see if we could update without reinit
```
TurboGit: indentation

src/dtgtk/paint.c
```diff
@@ -326,6 +326,28 @@ void dtgtk_cairo_paint_plusminus(cairo_t *cr, gint x, gint y, gint w, gint h, gi
   cairo_restore(cr);
 }
 
+void dtgtk_cairo_paint_sorting(cairo_t *cr, gint x, gint y, gint w, gint h, gint flags, void *data)
+{
+  cairo_save(cr);
+  gint s = w < h ? w : h;
```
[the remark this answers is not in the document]
phweyland: I've copied an existing one. Do I have to review all of them ?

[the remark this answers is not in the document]
TurboGit: You do not need to review all of them, but when you spot one, fix it yes. That's really helpful to understand/read the code.

src/common/database.c
```diff
@@ -1410,7 +1410,8 @@ static void _create_memory_schema(dt_database_t *db)
                            "(tmpid INTEGER PRIMARY KEY, id INTEGER UNIQUE ON CONFLICT IGNORE, count INTEGER)",
                NULL, NULL, NULL);
   sqlite3_exec(db->handle, "CREATE TABLE memory.similar_tags (tagid INTEGER)", NULL, NULL, NULL);
-  sqlite3_exec(
+  sqlite3_exec(db->handle, "CREATE TABLE memory.darktable_tags (tagid INTEGER)", NULL, NULL, NULL);
+    sqlite3_exec(
```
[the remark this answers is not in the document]
phweyland: ok

src/common/tags.c
```diff
@@ -228,6 +228,66 @@ guint dt_tag_remove(const guint tagid, gboolean final)
...
+      flatlist = NULL;
+      tcount = tcount + count;
+      count = 0;
+    }
+  }
+  if(flatlist && strlen(flatlist) > 1)
+  {
+    flatlist[strlen(flatlist)-1] = '\0';
+    dt_tag_delete_tag_batch(flatlist);
+    g_free(flatlist);
```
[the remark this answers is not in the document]
phweyland: ok

src/common/tags.c
```diff
@@ -517,35 +581,59 @@ void dt_tag_detach_by_string(const char *name, gint imgid)
...
 
+static void dt_set_darktable_tags()
+{
+  sqlite3_stmt *stmt;
+
+  DT_DEBUG_SQLITE3_PREPARE_V2(dt_database_get(darktable.db),
+                              "SELECT COUNT(*) FROM memory.darktable_tags",
+                              -1, &stmt, NULL);
+  sqlite3_step(stmt);
+  guint count = sqlite3_column_int(stmt, 0);
```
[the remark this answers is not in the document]
phweyland: ok

src/common/tags.c
```diff
@@ -575,46 +668,159 @@ uint32_t dt_tag_get_attached(gint imgid, GList **result, gboolean ignore_dt_tags
   return count;
 }
 
+uint32_t dt_tag_get_attached_export(gint imgid, GList **result)
```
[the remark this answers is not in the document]
phweyland: ok

src/common/tags.c
```diff
@@ -627,19 +833,20 @@ GList *dt_tag_get_hierarchical(gint imgid)
   int count = dt_tag_get_attached(imgid, &taglist, TRUE);
 
   if(count < 1) return NULL;
+  gboolean export_private_tags = dt_conf_get_bool("plugins/lighttable/export/export_private_tags");
```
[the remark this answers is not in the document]
phweyland: ok

src/common/tags.c
```diff
@@ -922,27 +1121,154 @@ uint32_t dt_tag_get_with_usage(const gchar *keyword, GList **result)
...
+  return synonyms;
+}
+
+gchar *dt_tag_get_synonyms(gint tagid)
+{
+  sqlite3_stmt *stmt;
+  gchar *synonyms = NULL;
+
+  DT_DEBUG_SQLITE3_PREPARE_V2(dt_database_get(darktable.db),
+                              "SELECT description FROM data.tags WHERE id = ?1 ",
```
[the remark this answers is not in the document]
phweyland: Correct, I reuse "description" as I haven't any usage and nobody has protested...
I have no experience of this kind of change on dt database schema. I suppose I'll have to code a migration path. I will look at it.
If you could pinpoint a similar example that would help for sure.

src/common/tags.c
phweyland: Yes, synonyms = description field = "first, second, third", or more.
I don't understand your doubt.

image

src/libs/tagging.c
phweyland: correct I've made confusion with the new flags. Ok.

src/libs/tagging.c
phweyland: ok. done

src/libs/tagging.c
phweyland: ok

src/libs/tagging.c
phweyland: ok

src/dtgtk/paint.c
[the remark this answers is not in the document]
phweyland: done on all (80)

src/common/tags.c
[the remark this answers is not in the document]
phweyland: As the field "description" is not used anywhere I do not need to have a migration path. 
I can just change at the table creation time, correct ?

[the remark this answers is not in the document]
phweyland: That's OK, I'm able to rename the field `description` to `synonyms` (migration path of data db)

src/common/database.c
```diff
@@ -1226,6 +1226,15 @@ static int _upgrade_data_schema_step(dt_database_t *db, int version)
 
     new_version = 2;
   }
+  else if(version == 2)
+  {
+    TRY_EXEC("ALTER TABLE data.tags RENAME COLUMN description TO synonyms;",
```
TurboGit: I'm surprised, is that really working now? Last time I checked the RENAME COLUMN was not implemented in SQLite. Is that fixed now? In which SQLite version?

phweyland:
mingw-w64-x86_64-sqlite3 3.28.0-1

my version sqlitestudio doesn't accept this command but it is in the doc and worked with dt.

phweyland:
We could remove the button "detach" as we have also the double-click and the sub-menu to do the same action. 
The button "attach" remains necessary because it works with "entry".
What do you think ?

TurboGit:
Given the commit and the documentation of release I think that RENAME COLUMN was first introduced in version 3.25.0. We now need to asses if this is ok for most platforms or if we need to rely on the old trick:

- creating a temp table
- copy, data
- delete current table
- create new table with proper field name
- copy data from temp

This has already been done many times in database.c.

TurboGit:
I would keep the detach button a good match for attach and does not take much space.

phweyland:
> We now need to asses if this is ok for most platforms or if we need to rely on the old trick

Let me know ...
As the column is not used, something delete column and add new column should do it. 
The icon column could be deleted too...

phweyland:
I've found a simple way to implement this [request](https://discuss.pixls.us/t/tagging-improvements-in-darktable-2-7/13571/13?u=phweyland) (see the images associated to a tag and back).
If that works can I add it without bothering you ?

TurboGit:
@phweyland : and the change sounds good to me.

Back to "RENAME COLUMN", the latest Ubuntu LTS is 18.04 and it has version 3.22 of SQLite which **does not** support column renaming. So this need to be addressed. And note that there is no "DELETE COLUMN" you can only add a new column. So I fear that you'll need to follow the procedure I have quoted above. You'll see example in database.c. Let me know if you need help.

phweyland:
OK. Do I keep he "icon" column ?

TurboGit:
No let's remove icon column. It is not used and it is anyway easy to add it back if needed.

TurboGit:
In "list mode", I don't have the completion anymore. That is, when I type the start of a tag, a possible completion is not proposed. This was very handy. Why is that not working anymore? Any idea? Did I missed something?

TurboGit:
The "go to collection" and "back to work" work fine. Like it. Just a small issue:

- got to collection.
- manually go back to the previous collection and you sill have the "back to work" entry.
- many each time the collection is changed the status should be reset?

Not 100% sure about this... What do you think?

src/libs/tagging.c
```diff
@@ -1387,6 +2223,8 @@ int position()
 
 void gui_init(dt_lib_module_t *self)
 {
+  setvbuf(stdout, NULL, _IONBF, 0);
```
TurboGit: This line and the following looks like debug ones. Should be removed;

phweyland:
> In "list mode", I don't have the completion any more.

I've tried to respect the old behaviour everywhere but I've removed this one because it always offended me. As a hierarchical tag user, to get the real tag you have to digit 20 or 30 letters ... The dictionary list reacts thousand times faster. Moreover the completion list covers the bottom window and prevents you to see the dictionary list. Just painful for me.

> This was very handy. 

Do you use it really ? What is the good case ?

No problem to add it back. I'd just put a configuration point to let the possibility to get rid of it. 
NB: the Ctrl-T completion works differently. Instead of searching from the beginning, it searches in plain text, providing a quicker and better fit. This could be done also for the entry completion list.
Let me know ...

>     * manually go back to the previous collection and you sill have the "back to work" entry.

Good catch. Once back to original collection, remove "back to work".

>     * many each time the collection is changed the status should be reset?

Makes sense. I could use signal_connect on collection change and reset "back to work". Is it what you mean ?

We could also let the user jump from tag to tag collection ("go to tag collection" sub menu) keeping the original collection "back to work" menu entry until back to original collection. 

Thanks

phweyland:
> I would keep the detach button a good match for attach and does not take much space.

Starting to write the doc, I've seen my mistake.  Both "detach" and attach button could be removed. The button related to entry is "new" and enter, not "attach". The button "new" could be moved at the current place of "attach" and "detach".

phweyland:
Applied the above comments for "go to collection" and "go back to work".

TurboGit:
> Do you use it really ? What is the good case ?

Yes, I'm using this all the time this way I do not leave the hand from the keyboards. I have some long hierarchical tags and having completion here speed up the process a lot.

TurboGit:
> NB: the Ctrl-T completion works differently. Instead of searching from the beginning, it searches in plain text, providing a quicker and better fit. This could be done also for the entry completion list.
> Let me know ...

Sounds promising if we can have this working with completion too.

Nilvus:
@phweyland : if you read #2859 issue, it's a really minor one but by the way, I find a little tweak to improve (a little bit) how buttons are displayed more consistently on tagging module. So, if it's ok to you and, of course, if rendering remains good with buttons you add with this PR, could you please add the two following lines in line 385 of darktable.css file with blank lines above and below to let space on file) :

`
#lib-plugin-ui-main button
{
  padding-left: 4px;
  padding-right: 4px;
}

`

phweyland:
> dd the two following lines in line 385 of darktable.css file

I can do that. But what about the other css file ? Do they inherit darktable.css ?

Nilvus:
> > dd the two following lines in line 385 of darktable.css file
> 
> I can do that. But what about the other css file ? Do they inherit darktable.css ?

Yes, that's why I only talk about this css file.

phweyland:
Using button expansion I get this :

image

Is that ok for everyone ?
Of course if the number of graphic buttons on the right varies, the width of other buttons varies too.

image

TurboGit:
I would say better then to disable the + button (instead of hiding) to avoid the size of the others to changes.

Nilvus:
> Using button expansion I get this :
> 
> image
> 
> Is that ok for everyone ?
> Of course if the number of graphic buttons on the right varies, the width of other buttons varies too.
> 
> image

I didn't know that was possible here regarding of how GTK widgets works. Great, it's of course better and more consistent for darktable UI. Good to see that. Thanks.

I agree with @TurboGit : better to disable + button.

timur-davletshin:
> > Using button expansion I get this :
> > image
> > Is that ok for everyone ?
> > Of course if the number of graphic buttons on the right varies, the width of other buttons varies too.
> > image
> 
> I didn't know that was possible here regarding of how GTK widgets works. Great, it's of course better and more consistent for darktable UI. Good to see that. Thanks.
> 
> I agree with @TurboGit : better to disable + button.

Those buttons are much better!

Nilvus:
> Using button expansion I get this :
> 
> image
> 
> Is that ok for everyone ?
> Of course if the number of graphic buttons on the right varies, the width of other buttons varies too.

Just a thing, have you test buttons with width expansion, without my css tweak ? As my css tweak was only increase padding left and right on buttons, I'm quite sure now my tweak is no more necessary with width expansion.

phweyland:
with expansion, without css trick and disabled suggestion button:

image

phweyland:
I haven't succeeded in making completion work with the tree treeview, only with the list.
Then I've disabled it with tree mode. I've also added a configuration to disable it if embarrassing.

junkyardsparkle:
This is great stuff! Just a couple of questions so far:

When typing into the field in list mode, I can type in, say, a person's given name, and get a filtered list of the long hierarchical tags which match, and I can then move to that field via tab key, and select among the choices with arrows, but... is there any way of actually making the selection with the keyboard? I can't seem to do it. This may (or not) be similar to the workflow @TurboGit was describing above.

Being able to adjust the sizes for the two list views is *very* nice (along with hiding the darktable tags). Now that this is in place, is there any reason for not making the range fairly large, just so that if someone wants to devote screen space to that, they can?

Thanks again, now I don't feel bad about waiting so long to tag large parts of the collection. :beer:

phweyland:
> When typing into the field in list mode

I don't know if that's the @turbogit way, but for me it works like this: typing my name for example, the completion list gives all my relatives, I use the `down` arrow to reach the right record, `enter` to get it in entry, `enter` again to get it attached without using the mouse. 


> any reason for not making the range fairly large

no, I can enlarge this without problem.

junkyardsparkle:
>  I use the down arrow to reach the right record, enter to get it in entry, enter again to get it attached without using the mouse.

Hmmm, this wasn't working for me, but I tried a fresh config, and it does work there, so it looks like I need to investigate my configured shortcuts. I did notice, however, that there is now no tagging module for me in darkroom mode... am I missing something?

phweyland:
> there is now no tagging module for me in darkroom mode

There is a secret configuration in daktablerc: `plugins/darktable/tagging/visible`. 

could be added in configuration. BTW configuration would deserve some more structuring.

junkyardsparkle:
> There is a secret configuration in daktablerc: plugins/darktable/tagging/visible.

Ah, thanks. I've now noticed one other small detail: in lighttable mode, hovering over the thumbnails updates the contents of the image information module, but moving the pointer from an image that has tags applied to one that doesn't fails to flush the previously shown tags; they remain in place until hovering over another image with tags applied.

phweyland:
> in lighttable mode, hovering over the thumbnails updates the contents of the image information module, but moving the pointer from an image that has tags applied to one that doesn't fails to flush the previously shown tags; they remain in place until hovering over another image with tags applied.

Thanks for reporting this. Fixed in PR #2866.

phweyland:
👍 thanks !

PR #2855, opened 2019-08-13
Add release metadata

matze:
[Release metadata](https://www.freedesktop.org/software/appstream/docs/chap-Metadata.html) is used by GNOME Software, Flathub etc. and required during the Flathub build process. I don't think it makes much sense to list the entire release history though, so this just adds the 2.6.x series.


TurboGit:
Looks ok to me. Thanks.

PR #2866, opened 2019-08-17
Tagging - bug fixing + presets enabled

phweyland:
Some fixed bugs:

- some records don't appear when filtering the tree view
- when entry completion enabled, gtk error message when entry is changed on tree view
- any attach or detach while jumping from tag collection to the other resets the "go to back work"
- any attach or detach while jumping from tag collection to the other is painfully slow. This is due to the tag collection refresh => The fix consists in not issuing "tag changed" signal as long as "go to back work" is active. This signal is always issued otherwise.
- dt_tag_detach calls collection refresh. Costly and not symmetric with dt_tag_attach. => created a new dt_tag_detach_from_gui (as for attach) not to change the behaviour of control_jobs.c. From my point of view it could be worth to check if this collection refresh is really useful there.

new:

- the user can now define tag presets
- added a button to clear the entry (for those who are not full keyboard ;-) )

### DIFF
--- a/src/common/exif.cc
+++ b/src/common/exif.cc
@@ -3148,10 +3148,10 @@ static void dt_exif_xmp_read_data_export(Exiv2::XmpData &xmpData, const int imgi
     longitude = fabs(longitude);
     latitude = fabs(latitude);
 
-    int long_deg = (int)floor(longitude);
-    int lat_deg = (int)floor(latitude);
-    double long_min = (longitude - (double)long_deg) * 60.0;
-    double lat_min = (latitude - (double)lat_deg) * 60.0;
+    const int long_deg = (int)floor(longitude);
+    const int lat_deg = (int)floor(latitude);
+    const double long_min = (longitude - (double)long_deg) * 60.0;
+    const double lat_min = (latitude - (double)lat_deg) * 60.0;
 
     char *str = (char *)g_malloc(G_ASCII_DTOSTR_BUF_SIZE);
 
@@ -3272,14 +3272,14 @@ static void dt_exif_xmp_read_data_export(Exiv2::XmpData &xmpData, const int imgi
   DT_DEBUG_SQLITE3_BIND_INT(stmt, 1, imgid);
   while(sqlite3_step(stmt) == SQLITE_ROW)
   {
-    int32_t mask_num = sqlite3_column_int(stmt, 8);
-    int32_t mask_id = sqlite3_column_int(stmt, 1);
-    int32_t mask_type = sqlite3_column_int(stmt, 2);
+    const int32_t mask_num = sqlite3_column_int(stmt, 8);
+    const int32_t mask_id = sqlite3_column_int(stmt, 1);
+    const int32_t mask_type = sqlite3_column_int(stmt, 2);
     const char *mask_name = (const char *)sqlite3_column_text(stmt, 3);
-    int32_t mask_version = sqlite3_column_int(stmt, 4);
+    const int32_t mask_version = sqlite3_column_int(stmt, 4);
     int32_t len = sqlite3_column_bytes(stmt, 5);
     char *mask_d = dt_exif_xmp_encode((const unsigned char *)sqlite3_column_blob(stmt, 5), len, NULL);
-    int32_t mask_nb = sqlite3_column_int(stmt, 6);
+    const int32_t mask_nb = sqlite3_column_int(stmt, 6);
     len = sqlite3_column_bytes(stmt, 7);
     char *mask_src = dt_exif_xmp_encode((const unsigned char *)sqlite3_column_blob(stmt, 7), len, NULL);
 
@@ -3512,8 +3512,6 @@ int dt_exif_xmp_attach_export(const int imgid, const char *filename)
 
     // last but not least attach what we have in DB to the XMP. in theory that should be
     // the same as what we just copied over from the sidecar file, but you never know ...
-    // the previous comment is not ok. First for exportation tags (and metadata) can be different
-    // second we could save time not reading the xmp file, just using the database data
     dt_exif_xmp_read_data_export(xmpData, imgid);
 
     img->writeMetadata();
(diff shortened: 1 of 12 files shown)


phweyland:
Found another bug.  dt_tag_get_list is called at different levels and I did miss one.
As a result the xmp companion file is also affected by the choice of exportation (private, synonyms) and should not.
So I duplicated dt_exif_xmp_read_data() (which was called for exportation but also to write xmp files) with a new dt_exif_xmp_read_data_export() separating the two streams. All routines on the exportation path are now suffixed "_export".

This routine reads the existing xmp file if any and overwrites it with data from database. I don't understand the benefit of this. On my side I don't use the XMP files and my exportations work pretty well without. Do I miss something ? If not we could remove this part and probably speed up exportation for those who use XMP files.

I have not updated watermark.c because a I don't know the usage of tags in that context.

phweyland:
just to document the last change here is the sequence of calls for exportation:
```
dt_imageio_export()
dt_imageio_export_with_flags()
dt_exif_xmp_attach_export()
dt_exif_xmp_read_data_export()
dt_tag_get_list_export() & dt_tag_get_hierarchical_export()
dt_tag_get_attached_export()
```
piwigo.c and flicker.c call also dt_tag_get_list_export()

That may be a first step in clarifying the export process from metadata standpoint.

src/common/exif.cc
```diff
@@ -3097,6 +3097,293 @@ static void dt_exif_xmp_read_data(Exiv2::XmpData &xmpData, const int imgid)
...
+  if(!std::isnan(longitude) && !std::isnan(latitude))
+  {
+    char long_dir = 'E', lat_dir = 'N';
+    if(longitude < 0) long_dir = 'W';
+    if(latitude < 0) lat_dir = 'S';
+
+    longitude = fabs(longitude);
+    latitude = fabs(latitude);
+
+    int long_deg = (int)floor(longitude);
```
TurboGit: const
(and the 3 next lines)

src/common/exif.cc
```diff
@@ -3097,6 +3097,293 @@ static void dt_exif_xmp_read_data(Exiv2::XmpData &xmpData, const int imgid)
...
+  xmpData.add(Exiv2::XmpKey("Xmp.darktable.masks_history"), &tvm);
+
+  DT_DEBUG_SQLITE3_PREPARE_V2(
+      dt_database_get(darktable.db),
+      "SELECT imgid, formid, form, name, version, points, points_count, source, num FROM main.masks_history WHERE imgid = ?1 ORDER BY num",
+      -1, &stmt, NULL);
+  DT_DEBUG_SQLITE3_BIND_INT(stmt, 1, imgid);
+  while(sqlite3_step(stmt) == SQLITE_ROW)
+  {
+    int32_t mask_num = sqlite3_column_int(stmt, 8);
```
TurboGit: All in int32_t should be const I suppose.

src/common/exif.cc
```diff
@@ -3097,6 +3097,293 @@ static void dt_exif_xmp_read_data(Exiv2::XmpData &xmpData, const int imgid)
...
+    int32_t mask_id = sqlite3_column_int(stmt, 1);
+    int32_t mask_type = sqlite3_column_int(stmt, 2);
+    const char *mask_name = (const char *)sqlite3_column_text(stmt, 3);
+    int32_t mask_version = sqlite3_column_int(stmt, 4);
+    int32_t len = sqlite3_column_bytes(stmt, 5);
+    char *mask_d = dt_exif_xmp_encode((const unsigned char *)sqlite3_column_blob(stmt, 5), len, NULL);
+    int32_t mask_nb = sqlite3_column_int(stmt, 6);
+    len = sqlite3_column_bytes(stmt, 7);
```
TurboGit: const

src/common/exif.cc
```diff
@@ -3225,7 +3512,9 @@ int dt_exif_xmp_attach(const int imgid, const char *filename)
 
     // last but not least attach what we have in DB to the XMP. in theory that should be
     // the same as what we just copied over from the sidecar file, but you never know ...
-    dt_exif_xmp_read_data(xmpData, imgid);
```
TurboGit: Not visible in the diff here, can you tell me what is happening for dt_exif_xmp_read_data(). The header is removed, is the body still needed?

phweyland: > The header is removed, is the body still needed?
Sorry I don't understand this ...

> what is happening for dt_exif_xmp_read_data()

That's the routine which gathers all metadata from dt database, plus history of masks and plugins into xmpData.
It was used both to write xmp companion file and to include metadata in exported files.
But the metadata can, but not have to be the same in both cases.
XMP companion has to mirror the database (not to loose anything)
Export file has to transmit the information needed by destination.

dt_exif_xmp_read_data_export() is a copy of dt_exif_xmp_read_data() except it calls the tag routines for exportation. And logically other metadata should be filterable in the same manner.
But history is useless for exportation, so I suggest to remove it from dt_exif_xmp_read_data_export().
So both routines are needed.

src/common/exif.cc
phweyland: I'll do it.

src/common/exif.cc
phweyland: I'll do it.

src/common/exif.cc
phweyland: I'll do it.

src/common/exif.cc
TurboGit: Ok, I see, thanks for the explanations.

parafin: > But history is useless for exportation, so I suggest to remove it from dt_exif_xmp_read_data_export().

It's not useless, so such change has to be discussed separately.

TurboGit: Maybe a confusion, are you saying that you are removing the xmp which is embedded in JPEG (at exportation time)?

phweyland: > It's not useless,

HI @paraffin, thanks for jumping in. What the utility of history in exported images?

parafin: It's a documented feature, see "load sidecar file" button description in the manual: https://darktable.gitlab.io/doc/en/history_stack.html

phweyland: OK, clear. Thanks. So we have to let history there. To keep or not the history can be an option amongst the others to control the metadata to be exported.

TurboGit:
Ok, looks good now. Thanks.

Issue #2774, opened 2019-07-07
Compress history in lighttable breaks data structure


s5k6:
**Describe the bug**

After compressing history in lighttable, recording further changes in darkroom to the history is messed up:  Every individual edit change of a value may be recorded to history, e.g., using the scroll wheel on the exposure module's slider may easily add 20 items to the history

**To Reproduce**

Open image in darkroom and create a history that will change under compression, e.g., change the exposure, then the cropping, and then the exposure again.

Back in lighttable, select the same image, choose "history stack/compress hist…".

Open the image in darkroom again.  Go to, say, the "exposure" module, and change the exposure slider by dragging or scrolling.  Every single change will be recorded to the history.  The same effect appears with other modules.

I guess "lighttable/history stack/compress hist…" does not correctly enumerate the image operations in the history stack.

**Expected behavior**

All consecutive changes in the exposure moudule should create only one entry in the history.

**Platform (please complete the following information):**

  * current ArchLinux
  * no desktop environment (plain X11 and FVWM)
  * release-2.7.0-1514-g122cca227

s5k6:
Related to #2420.

s5k6:
One can use the database to look at the history stack.  Use the image ID to find the image.  Say it's 9.  After using "history stack/compress hist…" in lighttable, I observe this:

    $ sqlite3 tmp/dt/config/library.db
    sqlite> .headers on
    sqlite> .mode columns
    sqlite> select num, module, operation, enabled from history where imgid == 9;
    num         module      operation   enabled
    ----------  ----------  ----------  ----------
    0           2           flip        1
    1           6           basecurve   1
    4           5           clipping    1
    5           5           exposure    1

Note that `num` does not contain 2 and 3.  My guess is that
lighttable's compression does not reenumerate the itms properly.
Darkroom;s compression leaves this column densely populated.

jenshannoschwalm:
Fixed by #2949

jenshannoschwalm:
@Turbogit Please close as fixed.

TurboGit:
Indeed, closing.

PR #2870, opened 2019-08-19
Correcting history compression in lighttable mode.

### LINKED ISSUE #2780: Compress history in lighttable does not retain disabled modules
**Describe the bug**

When a module is disabled, compressing its history stack from lighttable will remove that module from the stack.

**To Reproduce**

Open image in darkroom.  Change exposure.  Then disable the "exposure" module by clicking on its ⏻ symbol.  The history stack will contain "exposure (off)" as topmost entry.  Go to lightroom.  Make sure the same image is selected.  Choose "history stack/compress hist...".  Open the same image in darkroom again.  The exposure module has vanished from the history stack.

If you cannot reproduce this, it may be an effect of #2774.  Make sure you have a consistent history stack to start with.

**Expected behavior**

According to [this](https://github.com/darktable-org/darktable/issues/2775#issuecomment-509542189), compressing history should retain disabled modules.

  * current ArchLinux
  * no desktop environment (plain X11 and FVWM)
  * $ git describe: release-2.7.0-1514-g122cca227

jenshannoschwalm:
Fixes #2780 

The underlying reason for this bug is in src/common/history.c

In `void dt_history_compress_on_image(int32_t imgid)` we should **not** look for enabled modules but keep the summary status for **all** modules that have be enabled or disabled while developing as in `static void _lib_history_compress_clicked_callback` found in src/libs/history.c

Also the masks should be preserved as in darktable mode.

### DIFF
--- a/src/common/history.c
+++ b/src/common/history.c
@@ -870,76 +870,76 @@ int dt_history_copy_and_paste_on_selection(int32_t imgid, gboolean merge, GList 
   return res;
 }
 
-void dt_history_compress_on_image(int32_t imgid)
+void basic_dt_history_compress_on_image(int32_t imgid)
 {
   // make sure the right history is in there:
   dt_dev_write_history(darktable.develop);
 
   // compress history, keep disabled modules as in darkroom mode - adapted from libs/history.c
-  sqlite3_stmt *stmt_local;
+  sqlite3_stmt *stmt;
   DT_DEBUG_SQLITE3_PREPARE_V2(dt_database_get(darktable.db),
                               "DELETE FROM main.history WHERE imgid = ?1 AND num "
                               "NOT IN (SELECT MAX(num) FROM main.history WHERE "
                               "imgid = ?1 AND num < ?2 GROUP BY operation, "
                               "multi_priority)",
-                              -1, &stmt_local, NULL);
-  DT_DEBUG_SQLITE3_BIND_INT(stmt_local, 1, imgid);
-  DT_DEBUG_SQLITE3_BIND_INT(stmt_local, 2, darktable.develop->history_end);
-  sqlite3_step(stmt_local);
-  sqlite3_finalize(stmt_local);
+                              -1, &stmt, NULL);
+  DT_DEBUG_SQLITE3_BIND_INT(stmt, 1, imgid);
+  DT_DEBUG_SQLITE3_BIND_INT(stmt, 2, darktable.develop->history_end);
+  sqlite3_step(stmt);
+  sqlite3_finalize(stmt);
 
-  // delete all mask_manager entries - copied from libs/history.c
+  // delete all mask_manager entries
   int masks_count = 0;
   char op_mask_manager[20] = { 0 };
   g_strlcpy(op_mask_manager, "mask_manager", sizeof(op_mask_manager));
 
   DT_DEBUG_SQLITE3_PREPARE_V2(dt_database_get(darktable.db),
-                              "DELETE FROM main.history WHERE imgid = ?1 AND operation = ?2", -1, &stmt_local,
+                              "DELETE FROM main.history WHERE imgid = ?1 AND operation = ?2", -1, &stmt,
                               NULL);
-  DT_DEBUG_SQLITE3_BIND_INT(stmt_local, 1, imgid);
-  DT_DEBUG_SQLITE3_BIND_TEXT(stmt_local, 2, op_mask_manager, -1, SQLITE_TRANSIENT);
-  sqlite3_step(stmt_local);
-  sqlite3_finalize(stmt_local);
+  DT_DEBUG_SQLITE3_BIND_INT(stmt, 1, imgid);
+  DT_DEBUG_SQLITE3_BIND_TEXT(stmt, 2, op_mask_manager, -1, SQLITE_TRANSIENT);
+  sqlite3_step(stmt);
+  sqlite3_finalize(stmt);
 
-  // compress masks history --> took this from libs/history.c
+  // compress masks history
   DT_DEBUG_SQLITE3_PREPARE_V2(dt_database_get(darktable.db), "DELETE FROM main.masks_history WHERE imgid = ?1 AND num "
                                                              "NOT IN (SELECT MAX(num) FROM main.masks_history WHERE "
-                                                             "imgid = ?1 AND num < ?2)", -1, &stmt_local, NULL);
-  DT_DEBUG_SQLITE3_BIND_INT(stmt_local, 1, imgid);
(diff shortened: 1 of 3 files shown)


TurboGit:
I've not looked at the code, but I suppose we should have a single compress routine in common/history and call it also in libs/history with possibly some more calls here to satisfy the lib requirement (update gui or whatever....).

Do you think this is possible? At least sharing code for this seems a nice goal.

jenshannoschwalm:
@TurboGit I will give you a hint to merge after refactoring

TurboGit:
Perfect, thanks!

aurelienpierre:
> Do you think this is possible? At least sharing code for this seems a nice goal.

The lighttable code needs to iterate over all the selected pictures. Thinking about it, the darkroom path could be a special case of the more general lighttable case with only one selected image.

aurelienpierre:
> In void dt_history_compress_on_image(int32_t imgid) we should not look for enabled modules but keep the summary status for all modules that have be enabled or disabled while developing as in static void _lib_history_compress_clicked_callback found in src/libs/history.c

The problem, then, is every module that has been activated then disactivated will be kept in history. The real, underlying, reason, is some module are applied in dt by default (base curve, white balance, sharpening), and are force-applied ON if no evidence is found that they have been force-disabled by user. That needs to be addressed too (and I don't know how).

jenshannoschwalm:
@aurelienpierre 

> The problem, then, is every module that has been activated then disactivated will be kept in history.
Indeed.

> The real, underlying, reason, is some module are applied in dt by default (base curve, white balance, sharpening), and are force-applied ON if no evidence is found that they have been force-disabled by user.

Right again. I agree that this is a design decision. **But** we should have an identical behaviour for both use cases. 
You seem to think it's wrong to do it this way. I think it's just right. If you **actively** disable a module it should be saved in history as _disabled_, it does not matter whether you enabled it before or it was enanbled by default settings.

@TurboGit From your comments on #2780, do we want disabled modules marked as disabled in history or not?

TurboGit:
To me we want disabled module to be marked as disabled in the history. It is not an option to remove them as they may have settings ready to be re-enabled at any time. So we want to keep them in the history even when disabled.

aurelienpierre:
When things go bad, for example while applying styles or copy-pasting history, sometimes you end up with bugged history stacks of more than 50 entries, mostly duplicates instances. In some cases, I ever had images I could not even open, with hundreds of entries in history (and that was back when I used stable dt from repository).

I'm not opposed to both compression buttons having the same behaviour. I'm saying the default state of darktable is buggy, and users should have a way to manually flatten the history without opening the image, in cases it has gone south. Otherwise, the only other option is to discard the history entirely.

TurboGit:
> When things go bad, for example while applying styles or copy-pasting history, sometimes you end up with bugged history stacks of more than 50 entries

In that case we need to fix this and not hide it by adapting the history compression here I would say.

jenshannoschwalm:
> In that case we need to fix this and not hide it by adapting the history compression here I would say.

I sort of disagree here.

Both ways offer a compression of the history, we must define the way we want it to behave, either include **only enabled** modules or all modules that have been used.

The point of a buggy history is a different point in my opinion so it should be addressed in another PR, i would certainly also do that.
What is wrong with the history?

1. Scrolling a slider for example _pollutes_ the history, is this really necessary? Difficult to fix?
2. Can we overflow the history? Is this the problem aurelianpierre ran into?
3. Can we clearly define situations when we can **always** assume it's safe to compress history automatically?
4. Whatever way for history compression we agree on can be done via the lighttable button!, thats what aurelianpierre wanted i guess.

TurboGit:
> What is wrong with the history?

Nothing. My thinking is that the compression algorithm on the darkroom has always been working properly. We need to have the same code used in lighttable.

> Can we overflow the history? Is this the problem aurelianpierre ran into?

No we can't.

> Can we clearly define situations when we can always assume it's safe to compress history automatically?

It is always safe as long as you keep disabled modules (last history of it only). There is always a loss when compression of course. That's why it is not automatic but done when user ask for it.

jenshannoschwalm:
@TurboGit done the refactoring. I think this also fixes #2774

aurelienpierre:
> It is always safe as long as you keep disabled modules (last history of it only). There is always a loss when compression of course. That's why it is not automatic but done when user ask for it.

The only modules that **need** to be kept in history when disabled are the ones that will be re-applied automatically if we ditch them (AFAIK, only white balance is affected, and base curve/sharpen if they are to be applied by default). Most modules can be safely removed if disabled. I think someone on the mailing list had a PR coming to handle white balance and others specifically.

TurboGit:
>  Most modules can be safely removed if disabled.

I don't agree. All disabled modules have been used and have a set of settings that may be reused  later by just enabling back the module. That's the behavior since dt inception I don't think we want to change that by default.

If we want to change this behavior we need a preference if people want to really use a dangerous semantic.

src/common/history.c
```diff
@@ -870,66 +870,76 @@ int dt_history_copy_and_paste_on_selection(int32_t imgid, gboolean merge, GList
   return res;
 }
 
-void dt_history_compress_on_image(int32_t imgid)
+void basic_dt_history_compress_on_image(int32_t imgid)
 {
   // make sure the right history is in there:
   dt_dev_write_history(darktable.develop);
```
TurboGit: this should probably be in dt_history_compress_on_image() only as this can be called from lighttable nothing to do.

src/common/history.h
```diff
@@ -55,6 +55,7 @@ void dt_history_delete_on_selection();
 /** compress history stack */
 void dt_history_compress_on_selection();
 void dt_history_compress_on_image(int32_t imgid);
```
TurboGit: I would rename this one:
dt_history_compress_on_image_and_reload() // intended to be used in the darkroom

src/common/history.h
```diff
@@ -55,6 +55,7 @@ void dt_history_delete_on_selection();
...
+void basic_dt_history_compress_on_image(int32_t imgid);
```
TurboGit: And this one renamed to just:
dt_history_compress_on_image()

src/common/history.c
TurboGit: Well, maybe safe to keep it there after all. If a style is applied and the history not properly written after... Probably safer to ensure that history is properly written before dealing with it. So I withdraw my comment. Do you agree with the analysis?

jenshannoschwalm: Yes i do. Would keep it

src/common/history.h
jenshannoschwalm: Sure

src/common/history.h
jenshannoschwalm: Sure

TurboGit:
Thanks !

PR #2872, opened 2019-08-20
culling : fix pan if some image are not zoomed

### LINKED ISSUE #2851: culling mode cannot pan on 2nd image
I cannot pan when on second image.

**To Reproduce**
1. Start dt
2. Select one picture
3. Select "culling" mode with 2 pictures displayed
4. Zoom on 2nd one (on the right, ctrl-shift-scroll)
5. Pan on this image, nothing happens

I have also some strange & different behavior. So I'm not sure this will reproduce all the time but it is quite frequent.
 
**Expected behavior**

We should be able to pan. Sometimes you can pan the second image going over the first one.

 - OS: GNU/Debian
 - Version git current master

AlicVB:
this fix #2851


TurboGit:
Thanks a lot! Works fine on my side.

PR #2874, opened 2019-08-21
Revert direction of shadows/highlights balance slider. Fixes #2610

### LINKED ISSUE #2610: Shadow/Higlights slider in Split Tone Module.
Not sure if you'd call it a "bug", or just an oddity.
The shadow/highlights slider in the Split Toning module is orientated in a non-intuitive manner.
Shadows are on the right end of the slider, and highlights are on the left end of the slider.
Given that the histogram (and every other tool that maps luminosity in some fashion) has shadows at the left, and highlights at the right, wouldn't it make more sense for this slider to be orientated the same way?
Cheers,
Bruce.

hellbuster:
I reverted the direction of the slider, so shadows are on the left and highlights on the right.

As an unrelated question, I find the names for the gui data structs to be too generic, is there any convention for them? For example, balance_scale or highlight_hue_gslider instead of scale1 and gslider1 would make a lot of comments unnecessary and would avoid a lot of typos.

### DIFF
--- a/src/iop/splittoning.c
+++ b/src/iop/splittoning.c
@@ -286,7 +286,7 @@ static void balance_callback(GtkWidget *slider, gpointer user_data)
   dt_iop_module_t *self = (dt_iop_module_t *)user_data;
   if(self->dt->gui->reset) return;
   dt_iop_splittoning_params_t *p = (dt_iop_splittoning_params_t *)self->params;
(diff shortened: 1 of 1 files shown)


TurboGit:
Looks ok to me. Merged.

For the naming convention, there is none, but surely those generic names should be avoided. I'll be glad to merge any improvement on this.

Issue #2544, opened 2019-05-09
Consistent UI : having metadata import options on all 3 import options (not only image and folder ones)


Nilvus:
**Describe the bug**
I've just found that my metada settings on preferences are not applied anymore since march. I've seen that my photos imported on 9th of March have these metadata (I never applied them with metadata editor) and since 23th of March, no metadata applied on importing (no photos between these two dates).

**To Reproduce**
1. Have creator, rights or one of the four metadata that could be applied on importing,
2. Click on import from camera (with camera or card plugged of course)
3. Select images you want to import and import them
4. See image information or if datas are applied on metadata editor (these infos are empty)

**Expected behavior**
Have default metadata settings on prefs applied on importing

**Platform (please complete the following information):**
 - OS: Ubuntu 19.04
 - darktable master

**Additional context**
It's probably a revision between the 1st march and 23th March that cause this bug as I regularly update darktable master.

TurboGit:
Maybe to my work on metadata (and more generally lt) undo. I'll check.

Nilvus:
@TurboGit : just a reminder but not the most urgent one. I will just add (valmy on french forum of darktable) precise that metadata editor keeps datas on fields when clicking on applying, making thinking that nothing applying. Of course, when using metadata editor, that works but having a way to know that options have been applied would also be better (a message for example just between fileds and buttons in grey saying that fiels have been applied would be sufficient I think).

jenshannoschwalm:
I had a look into this; in fact i think it's not really a bug, it's just working in a different way than the manual says.

- The metadata import works only when enabling **apply metadata on import** in the **import options**.
- Maybe we should set the default to **On** to keep the old behaviour style?

Nilvus:
>     * The metadata import works only when enabling **apply metadata on import** in the **import options**.
> 
>     * Maybe we should set the default to **On** to keep the old behaviour style?

@jenshannoschwalm : Thanks for helping here.

I've searched 3 times on preferences window and didn't find this option. I've always had import working by default on darktable 2.6 and master one until march.

Anyway, I look on darktablerc file and indeed I find this line:

`ui_last/import_apply_metadata=FALSE`

Change FALSE to TRUE and restarting darktable making import options working automatically as it was before.

So, you're right that's not a bug but a changed and not wanted behaviour. Probably by commit published on march.

I agree with you, **this should be activated by default to keep the old behaviour and most of the time, when we set import options (author, title...) on preferences window, we want this to be activated.**

[...]

PR #2875, opened 2019-08-21
Add metadata support in the camera import dialog.

TurboGit:
This makes the import from camera dialog consistent with the import
files or folders.

For #2544.


Nilvus:
That's rocks ! I've tested import with and with option check and metadata are imported or not depending on choice. I also tested to see what happening when checking/unchecking the same option on image/folder/camera import and see if it's updated on other windows. On what I have tested, all work as expected. I close my issue as this solved it.

Thanks for your help and work on improving that. We now have consistent behaviour on all import options. Good to see that !

PR #2878, opened 2019-08-23
fix EXR read with channels other than RGB

claudeha:
Instead of requiring exactly R,G,B and optionally A channels, require at least R,G,B channels and optionally A.  Any extra channels are ignored.

Instead of checking just the first character of the channel name string, check the whole name.

Excerpt from `exrheader example.exr` that could not be loaded without this PR and can be loaded with this PR:

```
channels (type chlist):
    B, 16-bit floating-point, sampling 1 1
    G, 16-bit floating-point, sampling 1 1
    N, 32-bit unsigned integer, sampling 1 1
    NF, 32-bit floating-point, sampling 1 1
    R, 16-bit floating-point, sampling 1 1
```

Previously to this PR, a workaround is using ImageMagick's `convert in.exr out.exr`.


TurboGit:
Sounds good! I don't have a way to test that though. Thanks.

PR #2889, opened 2019-08-26
Metadata export

phweyland:
The goal of this PR is to allow the user to define the metadata he wants to export with his images.
A first part is a set of global flags (geotags, private, ...).
A second part let the user choose the metadata tags he wants to export and give the formula to calculate them with existing dt data. Here we want to avoid to enter several times the same information when once is enough.

This uses the preset mechanism, but instead of presenting a new module, this is integrated with `export selected` thanks to a preferences button on the right of combo-box which allows to choose the export_metadata template.

For the time being this looks like that:
image

This is a work in progress. Only the UI is done.
Globally this should not interfere with the PR related to metadata UI.
Remains to do:
- interpretation of formulas (as for file + extension for tags)
- integration in the export stream when metadata are generated.

The code repartition is not the final one (between library, export.c and other export_metadata stuff) but this is faster than developing at library level. I'll do that at the end.

EDIT: As usual, comments and suggestions are welcome.


Nilvus:
Oh yes, that's sound just awesome. Thanks a lot to improve all that things !

phweyland:
Current UI:

image

phweyland:
If formula is let empty, the corresponding tag is removed from exported image.
Formulas with $(TAG<n><category>) works with <n> from 0 to 3 and <category> = first level of tag. 

image

phweyland:
Next steps :
- extend this (disk device) to other export devices (piwigo, flickr, ...), if you agree, ... I'll remove the metadata hard coding in these threads.
- do I add the metadata preset to "export selected" params ? => version increment.
- some code clean up

phweyland:
>     * extend this (disk device) to other export devices (piwigo, flickr, ...), if you agree, ... I'll remove the metadata hard coding in these threads.

Looking at the code, it's possible but can bring some trouble to user.
As is, few parameters (title, description and tags) are sent to remote storage through the link (and not embedded within the image). However exif data (and geotags) are sent inside the image.
The mechanics of this PR controls the embedded data. To make it work properly it is necessary to understand how and which metadata each external storage expects to interpret. For example piwigo read the embedded exif and iptc data (nothing about xmp). The metadata settings for piwigo are these  (piwigo understands these embedded metadata):

image

Of course, each remote storage (flickr, ...) may read different metadata.

To resume, the current remote storage code simplifies the user life, as he hasn't to understand which metadata are needed by the storage. On the other hand the new mechanisms would offer far more possibilities and flexibility, at the cost of understanding on the user side.

So, for the time being, I would let the remote storages as they are. As a piwigo user, I will use the disk storage to export my images keeping full control of metadata and upload them to piwigo outside dt.

What do you think ?

phweyland:
Currently metadata presets are independent of export presets. This way you can use a same metadata preset for different export presets.
On the other hand a metadata preset can be deleted while it was referenced by an export preset...

Do you think it would be worth to try to integrate export metadata settings to export presets ?

phweyland:
just a button to call metadata settings (enabled only if disk storage)

image

edit in list (BTW substitution works):
image

new list of variables:
image

Ready for review when you can. Thanks !

TurboGit:
> (enabled only if disk storage)

Why? Does that means that we can't control the metadata for other exports?

phweyland:
> Why? Does that means that we can't control the metadata for other exports?

Exif and geotag are already transferred through embedded metadata. But title, description and tags (each interface has one to three of them), are hardcoded and sent through link and not embedded in images.

So, as done, it's quite easy to apply the settings to remote storages as well and remove the hard coding, but for sure that may/will change the interface behaviour from the user standpoint. Furthermore , as shown above for piwigo, the user will have to know/understand which metadata the specific remote storage expects or is able to pick up from embedded metadata.

phweyland:
Remote storages. Other suggestion.
- Keep the hard coded title, description and creator where they are (as main expected data).
- Remove tag_export flag on flickr and piwigo as redundant with new metadata presets. metadata presets will have to be manually updated as i don't know how to migrate legacy sub-module (storage modules) params to export params.
- Enable all metadata presets for all storages otherwise. Here too, it may be possible that metadata presets had to be tweaked a bit to match previous behaviour if wished.

I think that would be worth... Please let me know.

TurboGit:
Cab you tell me more about point 2 above? Not sure I follow the "manually updated" and the migrate part?

phweyland:
> Cab you tell me more about point 2 above? Not sure I follow the "manually updated" and the migrate part?

for flickr and piwigo there is a parameter flags_export in UI and params. This flag enables / disables the keywords metadata. This is params' element of `flickr `or `piwigo` which are storage sub modules of `export`. If I can set up a `legacy` routine for `flickr` and `piwigo` to remove the `flags_export` flags (which in itself is useless) I still cannot update accordingly the params of `export` to save the value. So we will loose the `flags_export = no` parameter / preset. In that case the user will have to disable the `tags` option (yes per default) in `export` parameter / preset, manually.

Does that help ?

TurboGit:
If I understand correctly. The new metadata export we talk here is to select which metadata we want to export (all, part of nothing). And this for all exports from all storage. Right? And the export setting on this new editor is global and not saved into the parameters, Right?

If it is the case, why not keep the flickr and piwigo parameter 'export_tags' as a main selector to enable or disable the tags export settings from the new editor?

phweyland:
> And the export setting on this new editor is global and not saved into the parameters

The metadata export settings are saved with the parameters of `export selected` module. So each `export selected` preset has its own metadata export settings. metadata for a customer are probably not the same as those for your family site.

>  why not keep the  flickr and piwigo parameter 'export_tags' as a main selector

Why to code X times the same functionality ? 
flickr, piwigo and the other storages are specific modules treated as sub modules of `export selected`. This makes sense to support specific features but should not duplicate common ones, right ?

TurboGit:
> Why to code X times the same functionality ?

There is no point indeed. It was not clear to me that the new export tags settings was saved into the presets. I didn't read the code yet :)

> flickr, piwigo and the other storages are specific modules treated as sub modules of `export selected`. This makes sense to support specific features but should not duplicate common ones, right ?

So, I don't see a good solution or at least nothing that we can do easily to bring the old "export tags" setting into the new code. Maybe just drop this setting and remove the corresponding gui widget in piwigo and flickr.

phweyland:
I've made the changes as described above except removing of the global configuration "omit hierarchy in simple tag list". The export flow has its own flag but the global flag is used other features: xmp companion file, watermark, and probably others I haven't identified. So I've let it to be more secure. 

I've tested the piwigo case.

src/libs/export.c
```diff
@@ -1097,11 +1148,12 @@ void *get_params(dt_lib_module_t *self, int *size)
     iccfilename = NULL;
   }
   if(!iccfilename) iccfilename = g_strdup("");
+  if(!metadata_export) metadata_export = g_strdup("");
 
   char *fname = mformat->plugin_name, *sname = mstorage->plugin_name;
   int32_t fname_len = strlen(fname), sname_len = strlen(sname);
   *size = fname_len + sname_len + 2 + 4 * sizeof(int32_t) + fsize + ssize + 6 * sizeof(int32_t)
-          + strlen(iccfilename) + 1;
+          + strlen(iccfilename) + 1 + strlen(metadata_export) + 1;
```
TurboGit: I think last +1 should be removed.

src/libs/export_metadata.c
```diff
@@ -0,0 +1,434 @@
...
+static void formula_edited(GtkCellRenderer *renderer, gchar *path, gchar *new_text, dt_lib_export_metadata_t *d)
+{
+  GtkTreeIter iter;
+  if (gtk_tree_model_get_iter_from_string(GTK_TREE_MODEL(d->liststore), &iter, path))
+    gtk_list_store_set(d->liststore, &iter, DT_LIB_EXPORT_METADATA_COL_FORMULA, new_text, -1);
+}
+
+uint32_t dt_lib_export_metadata_default_flags()
+{
+  const uint32_t flags = DT_META_EXIF | DT_META_METADATA | DT_META_GEOTAG | DT_META_TAG;
```
TurboGit: DT_META_DT_HISTORY should be added there. It is the current default and this is really a safe setting.

src/gui/accelerators.c
```diff
@@ -192,7 +192,7 @@ void dt_accel_register_lib(dt_lib_module_t *self, const gchar *path, guint accel
     else if(strcmp(views[i], "*") == 0)
       accel->views |= DT_VIEW_DARKROOM | DT_VIEW_LIGHTTABLE | DT_VIEW_TETHERING | DT_VIEW_MAP | DT_VIEW_PRINT
                       | DT_VIEW_SLIDESHOW;
-    i++;  
+    i++;
```
TurboGit: Not important, but could have been omitted from this PR. All formatting are adding more things to review.

src/libs/export_metadata.c
phweyland: > this is really a safe setting

Do you think that really ?

This may be a code beauty but for me this is even not a nice to have ... people should be instructed to backup properly their databases instead of relying on this. The same for xmp files used as safety. At least I have the setting to avoid this last one but I'll have to think to remove this one every time . :-(
I'll do as you request. :-)

src/gui/accelerators.c
phweyland: Sorry, I'll add this space back.

TurboGit: No, no need :) It was a comment for next PR.

src/libs/export_metadata.c
TurboGit: Fact is I do use (and know other people doing that) this as a simple way to show to others the was it was developed. Giving the RAW + Jpeg with the result people can replay the development if needed. Otherwise you also have to hand the .xmp.

I had also one time looked at an exported jpg to recover an history stack I had lost.

src/gui/accelerators.c
phweyland: Sorry ... too late ...

phweyland: ok, understood !

src/libs/export_metadata.c
phweyland: ok, I understand. And in any case I totally agree we shouldn't change the basic behavior without agreement of the dev team ... :-)

phweyland:
> Manually merged.

I guess I've made something wrong ... What should I not do next time ? The merge branch master ?

TurboGit:
Nothing really wrong, I just want to avoid the a master merge into a topic branch. So I have just rebased your branch before merging.

aurelienpierre:
Just tested that, it's quite nice ! Thanks @phweyland !

But is it normal that the settings don't get saved in parameters for the next reboot ?

phweyland:
Metadata parameters are saved along with the export selected module parameters. So as long as you haven't done "update preset xxxx" or "store new presets" they are not saved.

aurelienpierre:
Sure, but it would be great to have them stored in `darktablerc` and reloaded by default at startup, as other parameters in the export module are, so you don't need to reapply the preset everytime you reboot the software, especially if you changed a couple of settings from the base preset. If you got some spare time, that would be a nice detail to run the extra mile and provide a smooth UX.

phweyland:
Understood. Good idea. 
At first glance the only issue will be to store in darktablerc multiple parameters in a single line (as is in params) with some non ascii separators. Otherwise we will have a new line for each formula with some kind of incremental keyword, like formula01, formula02, ...
What would be the best track ?

TurboGit:
I would go with multiple lines with incremental keyword.

PR #2892, opened 2019-08-27
lens: fix transform routines.

TurboGit:
This fixes the position "jump" when for example adding a circle
mask in exposure module while the lens correction one is activated. Yet,
the circle and the actual area of the effect are not fully overlapping.

### DIFF
--- a/src/iop/lens.cc
+++ b/src/iop/lens.cc
@@ -862,10 +862,15 @@ void tiling_callback(struct dt_iop_module_t *self, struct dt_dev_pixelpipe_iop_t
   return;
 }
 
+// lensfun does not provide a back-transform routine. So we do it iteratively by assuming that
+// a back-transform at one point is just moving the same distance in the opposite direction. This
+// is of course not fully correct so we do adjust iteratively the transformation by checking that
+// the back transformed points are when transformed very close to the original point.
+//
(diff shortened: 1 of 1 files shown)


TurboGit:
Note that to me, the effect (on exposure module) is correct as it does keeps the same position on the image with and without the lens correction applied.

But the mask is wrong as it keeps the screen position but not the image position and then goes out of sync with the actual position of the effect.

TurboGit:
Ok, inverting transform and backtransform seems to fix the issue.

jenshannoschwalm:
Yeah! Almost done :-)
There is one minor problem remaining, the displayed circle and borders are not always circles but sometimes are like this
maskshot

I guess the reason for this is that you approximate for closeness on the horizontal dist1 only. Maybe
do another iteration on dist2 and take the middle?
(The direction of the shift depends on the position in the picture btw and might jump.) 

Otherwise, perfect!

TurboGit:
@jenshannoschwalm : Indeed, I have been able to reproduce. should be fixed now. can you test the new version? Thanks.

jenshannoschwalm:
Done! Congratulations. (Surprised this works btw)

TurboGit:
Thanks for testing.

Issue #2775, opened 2019-07-07
Darkroom's compress history does not minimize stack


s5k6:
**Describe the bug**

Using "historu/compress history stack" in darkroom does not “create a minimal history stack which produces the same image” as the tooltip promises.  Instead, "off" modules remain in the stack.  

**To Reproduce**

Open an image in darkroom.  Change its exposure, then switch off the exposure module.  Change its tone curve, then switch off the tone curve module.  Select "history/compress history stack".  The modules "exposure (off)" and "tone curve (off)" remain on the stack.

**Expected behavior**

A “minimal history stack which produces the same image” must not contain modules which do not contribute to the image.

**Additional context**

Using "lighttable/history stack/compress hist…" will remove disabled modules, but is otherwise [buggy](https://github.com/darktable-org/darktable/issues/2774).

I have [suggested a solution here](https://www.mail-archive.com/search?l=darktable-dev@lists.darktable.org&q=subject:%22%5C%5Bdarktable%5C-dev%5C%5D+implementation+question%5C%3A+remove+all+unused+modules%22&o=newest).

TurboGit:
That's the expected behavior. The "off" module still retains the last edit for the module and can then be activated back.

Issue #2514, opened 2019-05-04
[opencl] opencl_synch_cache might be buggy?


AxelG-DE:
**Describe the bug**
What I describe here, might be related to [ this #2230](https://github.com/darktable-org/darktable/issues/2230) but can also be finally triggered by any other function or several issues are related to each other.

Why I say so? Several weired things happen and to say it in advance, I cannot track them with `-d perf`

When I set `opencl_synch_cache=active module` instead of `= false`

- Moving WB slider causes a slight stuttering in screen update (setting back to false and it is smooth again)
- sometimes adjusting the sliders in one module several times for fine tuning, each touch makes a new entry in the history stack 
(somewhere I saw an issue about such, here on github, once I found it, I would add reference to this)
- When I use, e.g. levels at the end of a heavily edited picutre, like [this one, on playraw (incl. xmp)](https://discuss.pixls.us/t/play-raw-nightscape-from-orange-free-state/10103/24?u=axelg), in the first place, it is really sticky and update are lagged by ~1s-2s on my quite fast machine. Leaving the module and come back or switching it of and on, all over sudden, the cache seems to work

**To Reproduce**
1. activate `opencl_synch_cache=active module`
2. pick a heavily edited pic like [this from above](https://discuss.pixls.us/t/play-raw-nightscape-from-orange-free-state/10103/24?u=axelg)
3. try to slide the black levels of the module levels
4. Switch levels off an on again and slide the black levels again

**Expected behavior**

- `opencl_synch_cache=active module` should work reliably, not only after re-approach the module or switching it off and on
- Multiple touches of modules (indeed without touching others) should not create a new history stack entry
- `opencl_synch_cache=active module` maybe logically cost some performance, acc. to this [usermanual/en/darktable_and_opencl_optimization](https://www.darktable.org/usermanual/en/darktable_and_opencl_optimization.html) ideally it does not create lag

**Platform (please complete the following information):**
 - OS: [gentoo linux] up to date
 - Version [dt-master +1113]

**Additional context**
I hope @edgardoh and @dtorop could look into this? :-)

edgardoh:
To check if this setting is working, on an image with several modules and with -d perf, update one of the last modules (on the pipe):

With opencl_synch_cache=active module or true the modules before the one being modified should not be reprocessed, with opencl_synch_cache=false modules should be reprocessed.

If that's working, then, on the same image, update one of the first modules (WB should do it)

[...]

Issue #2420, opened 2019-04-17
Tone curve creates countless new entries/lines in history stack


AxelG-DE:
**Describe the bug**
when ever you click or track anything in the controls of tone curve, it creates a new entry in the histroy stack. Rapidly you reach 300 lines and lose all performance of dt.

**To Reproduce**
1. Go to 'darktroom'
2. Click on 'tone curve'
3. drag the black point to somewhere or activate a mask and change chormacity'
4. enjoy the error

**Expected behavior**
like all the other modules and the behavior before, there should be merely one line in history stack per each attempt to change or edit

**Screenshots**
![Screenshot_20190417_230102](https://user-images.githubusercontent.com/49004103/56320622-dc2be980-6164-11e9-970f-31b44ef41440.png)

**Platform (please complete the following information):**
 - OS: [ gentoo linux]
 - Version [darktable 2.7.0+1025~g0384a4bb1-dirty]

**Additional context**

AxelG-DE:
maybe it is not just, or not at all a tone curve issue, just now I figured strange things in the history stack... (unfortunately cannot reproduce that color zones created hell of line too)

TurboGit:
I also get this from time to time (rare though) and generally compressing the history fix the issue.

AxelG-DE:
Yes compressing history stack does help unless you touch tone curve again :-)

Actually 2.7-1040 does not show this behavior anymore. Probably this issue can be closed?

TurboGit:
Ok, closing.

AxelG-DE:
@TurboGit 
acc. to #2538 does it make sense, to reopen this one?

AxelG-DE:
@TurboGit
Please see this video at around 9 minutes and please reopen this bug report. 

@brucewilliamsphotography is using official release, and I still see this on git-master from time to time. 

Also I have big suspiciousy, once we solved this, we might also solve other performance issues like #2514

Thanks in advance!

dtorop:
I've also seen this over the years, but not recently. Hard to remember, but my suspicion may have been that it happened with older XMPs being opened with newer versions of darktable, but I never verified this. Perhaps the workaround was to switch to lighttable then back to darkroom.

AxelG-DE:
@dtorop
That pic from Bruce was a very new one, so was my case. 

In that #2514 case I have a similar issue, that every time I change levels, I get a new entry in history stack. Indeed workarounds are possible but it feels buggy...

[...]

PR #2898, opened 2019-08-30
Fixing lighttables history compression.

jenshannoschwalm:
As the common function for history compression uses the global
darktable.develop we must load the image for every compression.
Before this commit we had these problems

1. compressing history on a selection of images could lead to complete loss of history for the selected images
2. Problems with the masks history have been existing before #2870 

@TurboGit , could you review please?

This is one of the problems with history i am currently aware of, there are probably more things left

1. There might be a history "pollution" like described in #2774. I observed this too, although i could not find a definite way to trigger but ended in a history stack of several hundred entries for a single module easily. (Still hunting this bug.) BTW, I did **not** observe the described num-leak 
2. @aurelienpierre reported situations where compression of history was not possible and he even was not able to load an image with problematic history. See discussion of #2870. I observed it can take quite long to compress a very large history. Also i was not able to load the image as he described when such a large history was existing.
3. As the history compression of many pictures might take a while we could have some progress information?

Anything more to tackle?

### DIFF
--- a/src/common/history.c
+++ b/src/common/history.c
@@ -872,9 +872,15 @@ int dt_history_copy_and_paste_on_selection(int32_t imgid, gboolean merge, GList 
 
 void dt_history_compress_on_image(int32_t imgid)
 {
-  // make sure the right history is in there:
+  // ok, just make sure the current history is written!
   dt_dev_write_history(darktable.develop);
 
+  // We use the global darktable.develop so when we want to use a specific image
+  // for history and masks in the database we have to load it in case it's not there
+  if(!dt_dev_is_current_image(darktable.develop, imgid))
+  {
+    dt_dev_load_image(darktable.develop, imgid); 
+  }
   // compress history, keep disabled modules as documented
   sqlite3_stmt *stmt;
   DT_DEBUG_SQLITE3_PREPARE_V2(dt_database_get(darktable.db),
@@ -975,7 +981,8 @@ void dt_history_compress_on_image_and_reload(int32_t imgid)
 
 void dt_history_compress_on_selection()
 {
-  // Get the list of selected images
+
+// Get the list of selected images
   sqlite3_stmt *stmt;
   DT_DEBUG_SQLITE3_PREPARE_V2(dt_database_get(darktable.db), "SELECT imgid FROM main.selected_images", -1, &stmt,
                               NULL);
@@ -985,6 +992,7 @@ void dt_history_compress_on_selection()
     dt_history_compress_on_image_and_reload(sqlite3_column_int(stmt, 0));
   }
   sqlite3_finalize(stmt);
+
 }
 
 // modelines: These editor modelines have been set for all relevant files by tools/update_modelines.sh
--- a/src/common/history.c
+++ b/src/common/history.c
@@ -874,17 +874,18 @@ void dt_history_compress_on_image(int32_t imgid)
 {
   if (imgid < 0) return;
 
-  // make sure the current history is written
-  dt_dev_write_history(darktable.develop);
-
-  // We use the global darktable.develop so when we want to use a specific image
-  // for history and masks in the database we have to load it in case it's not there
-  if(!dt_dev_is_current_image(darktable.develop, imgid))
-    dt_dev_load_image(darktable.develop, imgid); 
+  // We use the global darktable.develop so if we want to use a specific images
+  // history and masks in the database we have to load it.
   if(!dt_dev_is_current_image(darktable.develop, imgid))
   {
-     fprintf(stderr,"\nCould not load image# %i while trying to compress history",imgid);
-     return;
+    darktable.develop->iop = dt_iop_load_modules(darktable.develop);
+    dt_dev_read_history_ext(darktable.develop, imgid, FALSE);
+    // This is a necessary trick to enable reloading the history!
+    darktable.develop->image_storage.id = imgid;
+  }
+  else
+  {
+    dt_dev_write_history(darktable.develop);
   }
 
   // compress history, keep disabled modules as documented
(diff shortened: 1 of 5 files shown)


src/common/history.c
```diff
@@ -985,6 +992,7 @@ void dt_history_compress_on_selection()
     dt_history_compress_on_image_and_reload(sqlite3_column_int(stmt, 0));
   }
   sqlite3_finalize(stmt);
+
```
TurboGit: not intended I suppose?

src/common/history.c
```diff
@@ -872,9 +872,15 @@ int dt_history_copy_and_paste_on_selection(int32_t imgid, gboolean merge, GList
 
 void dt_history_compress_on_image(int32_t imgid)
 {
-  // make sure the right history is in there:
+  // ok, just make sure the current history is written!
   dt_dev_write_history(darktable.develop);
 
+  // We use the global darktable.develop so when we want to use a specific image
```
TurboGit: But then we have:

```
void dt_history_compress_on_image_and_reload(int32_t imgid)
{
  dt_history_compress_on_image(imgid);

  /* if current image in develop reload history */
  if(dt_dev_is_current_image(darktable.develop, imgid))
```
The condition above is always TRUE now. Worth checking that this is not broken now? Or remove the conditional?

src/common/history.c
jenshannoschwalm: No :-)

src/common/history.c
jenshannoschwalm: Right. Probably we can remove `dt_history_compress_on_image_and_reload` and directly use  `dt_history_compress_on_image`.

TurboGit: @jenshannoschwalm  : Any news on that?

jenshannoschwalm: I am still hunting for a history leak bug not yet understood.

jenshannoschwalm: The second commit
1. Moves more stuff from lib/history.c to common/history.c for refactoring
2. the compress_and_reload also keeps the selected position in the history stack
3. The lighttable compress history button has a check box -- also as a feedback.
4. I couldn't produce history leaks any more.
5. For me the compressed history always is ok.

**BUT** there is one remaining problem with the lighttables compression.
Try this:
Select some pictures in the lighttable and do the compression. Now leave darktable and start it again. The compressed histories are ok as expected.

If you do the compression and open **any** picture right after that you will see the history is **wrong**, some entries have found their way into the history.

Could you give me a hint? Do we need to 'invalidate' the darktable.develop or its history somehow?

jenshannoschwalm:
The 3rd commit to this PR has 

- Simpler code because of further refactoring
- Does proper reordering the history
- fixes the leak observed before.

It took me a while to accept & understand that there probably is no way to compress&reload the history while maintaining masks whithout _loading_ the complete image data with all information about how to process as it is done in the darkroom mode. (I worked on this quite hard to get a better performance...)
At least it works fine now.

@TurboGit Could you review again?
Fixes #2774

jenshannoschwalm:
@TurboGit Please hold on with review, i am still testing this, also looking how this might interfere with your latest #2918

TurboGit:
Sure, and indeed #2918 is a very important fix so we need to be sure this is working fine together with it. Thanks.

jenshannoschwalm:
There is one more commit!
It takes care of wrongful `_dev_add_history_item_ext`.
Finally this should fix a number of reported issues ( #2420 , #2514 , #2775 ) , where we had large numbers of history entries added especially after compression in the lightroom. It took me a long time to find :-)
@TurboGit This also seems to be a good one :-)

TurboGit:
Agreed, looks like a nice fix. I'll have a look.

src/libs/history.c
```diff
@@ -701,15 +701,11 @@ static void _lib_history_change_callback(gpointer instance, gpointer user_data)
...
+  if(imgid<0) return;
 
   dt_history_compress_on_image(imgid);
 
-  sqlite3_stmt *stmt;
-
-  // load new history and write it back to ensure that all history are properly numbered without a gap
-  dt_dev_reload_history_items(darktable.develop);
-  dt_dev_write_history(darktable.develop);
+/*  sqlite3_stmt *stmt;
```
TurboGit: Either remove this code or add a clear comment about why it is kept otherwise this will be noise for others and anyway at some point it will get removed and nobody will remember why it is there :)

src/common/history.c
```diff
@@ -956,21 +968,31 @@ void dt_history_compress_on_image(int32_t imgid)
...
+  DT_DEBUG_SQLITE3_PREPARE_V2(dt_database_get(darktable.db), "SELECT IFNULL(MAX(num)+1, 0) FROM main.history "
+                                                             "WHERE imgid=?1", -1, &stmt, NULL);
+  DT_DEBUG_SQLITE3_BIND_INT(stmt, 1, imgid);
 
-  /* if current image in develop reload history */
-  if(dt_dev_is_current_image(darktable.develop, imgid))
-  {
-    dt_dev_reload_history_items(darktable.develop);
-    dt_dev_write_history(darktable.develop);
-    dt_dev_modulegroups_set(darktable.develop, dt_dev_modulegroups_get(darktable.develop));
```
TurboGit: Difficult to understand from the diff, but it seems that the call dt_dev_modulegroups_set() is completely removed. Is that really intended? Was this call superfluous as already taken care in another routine? Just want to be sure this is not an oversight.

TurboGit:
I found one issue. Take a picture with "mask manager" into the history. Do compress history from darkroom or lighttable and the history_end is not correct and so top item not highlighted. The history_end in the images table seems to be +1 from the correct value.

This works fine with current master.

src/libs/history.c
jenshannoschwalm: Sure, i kept this as a reminder ... will be removed.

src/common/history.c
jenshannoschwalm: Thanks for observing this, but do we need this for every picture after compression or only if it is in darktable mode?

jenshannoschwalm:
> I found one issue. Take a picture with "mask manager" into the history. Do compress history from darkroom or lighttable and the history_end is not correct and so top item not highlighted. The history_end in the images table seems to be +1 from the correct value.

Aah. I missed that. Another hint for a leak. I'll find the reason...

src/common/history.c
jenshannoschwalm: Activated always. No problem.

jenshannoschwalm:
@TurboGit indeed there were more leaks to be tested.
1. Fixed that in develop.c
2. activated dt_dev_modulegroups_set after every history compression
3. Signalling history change after leaks

jenshannoschwalm:
Closing this as there is #2939 as a replacement.

PR #2901, opened 2019-09-02
lens: Adapted to Lensfun API

bronger:
It is now in sync with Lensfun commit 8e44481b (2019-07-18).

### DIFF
--- a/src/iop/lens.cc
+++ b/src/iop/lens.cc
@@ -356,20 +356,20 @@ static lfModifier * get_modifier(int *mods_done, int w, int h, const dt_iop_lens
   int mods_done_tmp = 0;
 
 #ifdef LF_0395
-  mod = new lfModifier(d->crop, w, h, LF_PF_F32, d->inverse);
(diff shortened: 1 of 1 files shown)


jakubfi:
I would be against adding compatibility with the current git lensfun version, as it is still versioned as 0.3.95 in the tree, while being API-incompatibile with the released 0.3.95 tgz. There is currently no clean way (if any) to distinguish between 0.3.95 and current. I know this is experimental anyway, but let's stick to a released version at least.
There is a nice solution though, while I have your attention: 0.3.96 pretty please? :-)

bronger:
Fair enough. I agree that both options are sub-optimal.

I don’t make releases as they are made by Sebastian Kraft.

TurboGit:
Any update on this ? What is the status on lensfun side ? Last stable version ? Should this be pursued or closed ? TIA.

TurboGit:
Since 0.3.95 is buggy and now even reverted from MSYS repository let's just close this.

PR #2903, opened 2019-09-02
Run `optipng -o7` on all PNG files

simon04:


TurboGit:
Can you tell me more about this ? What it is for ? What the goal ? Thanks.

aurelienpierre:
It looks like a lossless optimization of PNG assets of the program.

TurboGit:
I'm pretty sure it is an optimization step, but at the end what do we gain? That's my question, the PNG is smaller ? Load faster ? I'd like more input from @simon04.

agusmba:
After a quick look at the documentation, it seems that the optimization is size-wise, so they wouldn't necessarily load faster, unless tests show otherwise.

TurboGit:
Ok, now what I'd like to have is the overall gain in size... Can you clarify @simon04 ? Thanks.

simon04:
Sorry for the late reply; I was on vacation. From the [man page of optipng](https://linux.die.net/man/1/optipng):

> The OptiPNG program shall attempt to optimize PNG files, i.e. reduce their size to a minimum, without losing semantic information.

In particular, the screenshots from doc/usermanual/ have been reduced by up to 200 KB (such as doc/usermanual/images/darkroom_view.png). As a result, the user manual loads faster on slower internet connections.

For the data/pixmaps/ only a few bytes have been saved. Thus, no big difference will be noticeable…

PR #2909, opened 2019-09-04
culling : fix toggle mode shortcut

AlicVB:
this should only work in culling layout (currently, in some case, it change slider value even in filemanager layout...)


TurboGit:
Looks good! Thanks.

PR #2916, opened 2019-09-06
Change open preference button style from toggle button to button

mbaumgae:
In current 2.7 master the open preference button has toggle button style, which means, after closing the preference dialog, the open preference button is shown as active:
PreferenceButton
This PR changes the button to a standard button, as in version 2.6.2


TurboGit:
Looks good! Thanks.

PR #2918, opened 2019-09-06
Add all default modules that can't be disabled but are hidden into hi…

TurboGit:
…story.

This ensures that the corresponding module have an entry into the database
and into the XMP. This is needed to have proper handling of the reorder
of the modules.

Also, the corresponding entries are hidden from the history GUI.

Part of #2905.

### DIFF
--- a/src/common/history.c
+++ b/src/common/history.c
@@ -987,6 +987,22 @@ void dt_history_compress_on_selection()
   sqlite3_finalize(stmt);
 }
 
+gboolean dt_history_check_module_exists(int32_t imgid, const char *operation)
+{
+  gboolean result = FALSE;
+  sqlite3_stmt *stmt;
+
+  DT_DEBUG_SQLITE3_PREPARE_V2(
+    dt_database_get(darktable.db),
+    "SELECT imgid FROM main.history WHERE imgid= ?1 AND operation = ?2", -1, &stmt, NULL);
+  DT_DEBUG_SQLITE3_BIND_INT(stmt, 1, imgid);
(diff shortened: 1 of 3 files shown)


jenshannoschwalm:
I checked it with current master (2019.09.12) and  #2939 included. Works fine.

MStraeten:
in the history stack in darkroom the changes on the input color profile arent longer displayed. if using styles containing input color profile settings these aren't shown in the history stack, so it's hard to realize where changes were applied.

parafin:
That's why I originally suggested not to hide anything - hiding logic will be prone to bugs like that.

TurboGit:
> That's why I originally suggested not to hide anything - hiding logic will be prone to bugs like that.

I missed that then... Here you wrote the opposite :+1: 

https://github.com/darktable-org/darktable/pull/2905#issuecomment-528485286

But certainly I agree with you and I'm also more for displaying everything.

parafin:
Your question was "Do we want them to appear also into the history GUI?", my answer was "I guess yes" ;) By "transparently mapped from DB/XMP to GUI" I meant not to hide anything.

TurboGit:
My bad :) I didn't get it right!

TurboGit:
So, nothing hidden in this new version.

@MStraeten : can you double check that your issue is fixed now?

MStraeten:
confirmed, now it's fine

TurboGit:
@MStraeten : thanks for testing.

Let me merge this now. This is potentially delicate and better to have early field testing.

PR #2919, opened 2019-09-06
Piwigo export: explicit gui init + small clean-up

leclercj:


src/imageio/storage/piwigo.c
```diff
@@ -773,7 +774,7 @@ void gui_init(dt_imageio_module_storage_t *self)
   g_signal_connect(G_OBJECT(ui->server_entry), "changed", G_CALLBACK(_piwigo_server_entry_changed), (gpointer)ui);
   gtk_entry_set_width_chars(GTK_ENTRY(ui->server_entry), 0);
   gtk_box_pack_start(GTK_BOX(hbox), GTK_WIDGET(label), FALSE, FALSE, 0);
-  gtk_box_pack_start(GTK_BOX(hbox), GTK_WIDGET(ui->server_entry), TRUE, TRUE, 0);
+  gtk_box_pack_start(GTK_BOX(hbox), GTK_WIDGET(ui->server_entry), TRUE, TRUE, 10);
```
TurboGit: We have worked hard to remove all hard coded spacing in dt, so this should be done in the CSS. Not sure how though as trying to add a margin-left do not work!

leclercj:
Let me try another quick way. If still too dirty, I'll just remove this spacing change from this PR as I don't know how to do that with CSS

TurboGit:
Yes, still too dirty to my taste. In fact this is no better than previous version as it still force a given space without a way to theme it.

leclercj:
Ok, then let's remove this part

TurboGit:
Ok, look good now.

TurboGit:
Thanks.

PR #2925, opened 2019-09-09
mark keyboard shortcuts in RELEASE_NOTES.md

thomas-mc-work:
… by the semi-official mardkown/html tag `<kbd>`.


TurboGit:
We'll see how this works, but copy/paste as plain text won't be that easy/clean.

PR #2926, opened 2019-09-09
Fix instance menu initialization when no module groups are selected

### LINKED ISSUE #2726: Darktable instance options all greyed out except rename
**Describe the bug**
If none of the groups heading is highlighted, then the instance options are disabled (except rename)

**To Reproduce**

image

If one of the groups is selected, then the options enable:

image

Also, when the last active group heading is disabled again, the instance options remain enabled:

image

Not sure if I got the idea, but now the workaround seems to just click on one of the heading groups.

**Expected behavior**
Regardless one of the heading groups is selected or not, all instance options should be enabled

**Screenshots**
see above

 - OS: ubuntu 18.04
 - Version:
image

**Additional context**
It happens regardless the size of the history stack.

amousset:
When no module groups selection was restored from previous start, nothing updates the instance menu, and all entries are disabled until a group is selected (see #2726). This commit forces module instance status refresh after entering the darkroom.

I was not sure where to put this, it may be better somewhere else.

Fixes #2726

### DIFF
--- a/src/views/darkroom.c
+++ b/src/views/darkroom.c
@@ -2464,6 +2464,9 @@ void enter(dt_view_t *self)
     g_free(active_plugin);
   }
 
+  // update module multishow state now modules are loaded
+  dt_dev_modules_update_multishow(dev);
+
   // image should be there now.
   float zoom_x, zoom_y;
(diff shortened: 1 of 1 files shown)


TurboGit:
Looks ok to me. Thanks.

PR #2927, opened 2019-09-09
Filter exiv2 output in get_image_iso()

junkyardsparkle:
In at least some cases (including my current case of exiv2 0.26 and Olympus *.orf files) there is extra output ("IMAGE ORF") from exiv2 that seems impossible to suppress using any combination of options. Since this is expected to return a string of digits or nothing, use grep to explicitly filter for this.


TurboGit:
Ok, looks good thanks.

PR #2931, opened 2019-09-10
Do not try to display db error window when not in gui

amousset:
This could happen when trying to run darktable-cli when
darktable is started.

### DIFF
--- a/src/common/darktable.c
+++ b/src/common/darktable.c
@@ -836,7 +836,7 @@ int dt_init(int argc, char *argv[], const gboolean init_gui, const gboolean load
     if(connection) g_object_unref(connection);
 #endif
 
-    if(!image_loaded_elsewhere) dt_database_show_error(darktable.db);
+    if(init_gui && !image_loaded_elsewhere) dt_database_show_error(darktable.db);
 
     return 1;
   }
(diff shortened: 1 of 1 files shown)


src/common/darktable.c
```diff
@@ -836,7 +836,7 @@ int dt_init(int argc, char *argv[], const gboolean init_gui, const gboolean load
     if(connection) g_object_unref(connection);
 #endif
 
-    if(!image_loaded_elsewhere) dt_database_show_error(darktable.db);
+    if(init_gui && !image_loaded_elsewhere) dt_database_show_error(darktable.db);
```
TurboGit: I suppose this should be !has_gui ? See other instance in this same file.

amousset: I don't see a `has_gui` variable, and the `init_gui` seems to be used elsewhere for the same kind of condition. But it seems the dbus connection is only initialized, so the whole "trying to open the images in the running instance" block should probably be skipped when not in gui, I'll update the fix.

TurboGit: Right ! My bad, I was looking at database.c which is using has_gui but is a parameter set using init_gui. Sorry for the noise.

PR #2932, opened 2019-09-10
Fix shortcut to hide navigation thumbnail

### LINKED ISSUE #2483: New "Shift+Ctrl+N" shortcut to toggle navigation thumbnail visibility doesn't work
**Describe the bug**
As the title, this shortcut doesn't work. For example, other new ones like Shift+Ctrl+H for histogram works great but not the one for navigation thumbnail.
I've tried few times to change or reactivate the shortcut in darktable preferences tab but same behavior : nothing happen.

@aurelienpierre 

**Expected behavior**
A working shortcut ;-)

 - OS: Ubuntu 19.04
 - Master compile version 7fe35d67d

amousset:
Fixes #2483

Keeping "hide navigation thumbnail" would maybe be more consistent (with histogram hiding shortcut) but it would break already customized shortcuts.

### DIFF
--- a/src/libs/navigation.c
+++ b/src/libs/navigation.c
@@ -566,12 +566,12 @@ static gboolean _lib_navigation_leave_notify_callback(GtkWidget *widget, GdkEven
 
 void init_key_accels(dt_lib_module_t *self)
 {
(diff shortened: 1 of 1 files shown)


TurboGit:
Thanks!

PR #2935, opened 2019-09-11
Fix tag creation when tag_id is not defined

### LINKED ISSUE #2933: Creating Tag in tree view crashes Darktable
**Describe the bug**
In the the tagging tree view selecting the parent and trying to create a new tag under it crashes DT. However, creating  a sub tag on a child node  works.

**To Reproduce**
1. Have a hierarchical tag tree.
2. Select main parent and right click and create a new tag.
3. In the window that pops up add some text for name.
4. Click save.
5. DT crashes.
See image.
**Expected behavior**
Allow creation of tags under parent.

**Screenshots**

image

 - OS: Win 10
 - Version 2.7.0.1679

**Additional context**
Log file created.
[darktable_bt_Y6YY7Z.txt](https://github.com/darktable-org/darktable/files/3599361/darktable_bt_Y6YY7Z.txt)

`000000000F864727 000000000000001F 0000000064A156E0 000000000125B020  libtagging.dll!pop_menu_dictionary_create_tag  [F:/msys64/home/chill/darktable/src/libs/tagging.c @ 1409]
  1407:       message = _("'|' character is not allowed to create a tag. aborting.");
  1408:     char *new_tagname = g_strdup(newtag);
> 1409:     const gboolean root = !gtk_toggle_button_get_active(GTK_TOGGLE_BUTTON(parent));
  1410:     if (!root)
  1411:     {`

amousset:
The `parent` is only created when tag_id is true.
Apply the same condition before using it.

May fix #2933 (I got `Gtk-CRITICAL **: 00:10:44.143: gtk_toggle_button_get_active: assertion 'GTK_IS_TOGGLE_BUTTON (toggle_button)' failed` errors on the same line, but only in different conditions, like on deleted tags).

### DIFF
--- a/src/libs/tagging.c
+++ b/src/libs/tagging.c
@@ -1406,13 +1406,17 @@ static void pop_menu_dictionary_create_tag(GtkWidget *menuitem, dt_lib_module_t 
     if(strchr(newtag, '|') != 0)
       message = _("'|' character is not allowed to create a tag. aborting.");
     char *new_tagname = g_strdup(newtag);
-    const gboolean root = !gtk_toggle_button_get_active(GTK_TOGGLE_BUTTON(parent));
+    gboolean root = TRUE;
+    if (tagid)
+    {
+      root = !gtk_toggle_button_get_active(GTK_TOGGLE_BUTTON(parent));
+    }
     if (!root)
     {
       new_tagname = g_strdup(path);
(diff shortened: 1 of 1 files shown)


amousset:
Indeed, thanks for fixing it @phweyland

PR #2938, opened 2019-09-13
Filmic v3

aurelienpierre:
## Summary

Move filmic to the new module API and use an RGB default space : the module will work in pipeline RGB space and will have RGB blending params.

(Note : It's not an upgrade of previous filmic, but a new, rewritten module)

OpenCL version not provided here.

## Algo changes

- Ditch the multiple interpolations/intents options and use a custom-made spline with a pure-linear latitude and first + second order continuity imposed between parts.
See background here : https://eng.aurelienpierre.com/2018/11/30/filmic-darktable-and-the-quest-of-the-hdr-tone-mapping/#filmic_tone_curve
- Ditch the LUT and use explicit curve expressions vectorized with masks (runtimes on-par with LUT but accuracy improved)
- Ditch the gaussian mask for selective desaturation of extreme luminance but use the second derivative of the explicit curve expression
- Ditch most of the curve nodes sanitization checks since the grey node is almost garanteed to be on the linear part, improving UX while setting controls (more intuitive behavior)
- Desaturate extreme luminances before filmic curve
- Add more norms for chroma preservation
- Fix the over-saturation issues in chroma-preservation mode by preserving ratios through the display power function too -> remove global saturation parameter

## UI changes

Screenshot_20190913_193437

Alert when curve over/under-shoots:
Screenshot_20190913_193701

- Reorganize the user parameters in 3 tabs enforcing users to follow a rational setting path
- Set the latitude in % of the dynamic range so changes in the scene exposures setting preserve the general shape of the S curve
- Paint the grey node in orange on the graph
- Paint the over/under-shoots of the interpolated curve in orange to facilitate diagnostics
- Make the latitude symmetrical around the grey node by default

## Benefits

- More intuitive behaviours : each parameter produces only one change at a time and don't need compensation on other parameters anymore
- No more saturation issues and red shifts
- Better default params for quicker settings


MStraeten:
me again with an issue on OSX:
```
[ 66%] Building C object src/iop/CMakeFiles/filmicrgb.dir/introspection_filmicrgb.c.o
In file included from /Users/martinstraeten/src/darktable/build/src/iop/introspection_filmicrgb.c:51:
/Users/martinstraeten/src/darktable/src/iop/filmicrgb.c:524:20: error: 
      unexpected OpenMP clause 'aligned' in directive '#pragma omp parallel for'
  schedule(static) aligned(in, out:64)
                   ^
1 error generated.

```
compiled with current master + just #2938

dcendal:
I've playing with it for 20 minutes and 0 error. I was able to replicate the same quality as the previous version but in a quarter of time!, thanks a lot for this.

aurelienpierre:
Should be fixed @MStraeten

MStraeten:
yep bug fixed - very impressing results

MStraeten:
UI suggestion: since the graph isnt interactive it wastes a lot of vertical space so maybe a reduced height would be better

jakubfi:
After playing with it for a few minutes: filmic
Resetting the module, switching it off and on, discarding history - nothing fixes it. When I move sliders - curve moves, but is still borked.
Going back to lighttable and reopening image fixes the problem.

jakubfi:
In contrast to what @dcendal said, with v3 I cannot get the results I got with v2. I suspect I'm doing something wrong, so before anything else: @aurelienpierre are the prerequisites for using v3 the same as for v2 (proper demosaic + denoise + optimize dynamic range with exposure)? Is there anything new I should be doing?

aurelienpierre:
What do you mean @jakubfi ? Inputting the same parameters produces very different results or the setting path is different ?

You don't **need** to optimize the dynamic range with exposure (either with v2 or v3), it's only a commodity that let you then assume the white exposure is at +2.45 EV once optimized (but you could just tweak that value instead).

As a general rule, try to keep the white exposure ~= - black exposure so the curve is centered around  50% grey and the interpolation behaves better (same as v2). For this reason, the new defaults in v3 use a centered dynamic range, and the grey value setting (auto-adjusting white and black exposures) should do 60 % of the job itself.

For HDR images, you might need to decrease the black exposure to give more space to shadows, therefore de-balance the dynamic range. Instead of that, you can increase the exposure (in exposure module) until midtones are properly lit, not caring about highlights clipping, then, in filmic, move the white exposure slider until you recover the blown highlights.

jakubfi:
@aurelienpierre setting path is different by design, of course. UI has changed, some parameters value meaning changed too - latitude, contrast too, I guess. What I meant was that after ~15 minutes of playing with v3 I was unable to get the pleasing results of v2. I stopped trying to verify if I'm not making any noob mistake. I probably am anyway. I will give it another try and get back to you with more organized and "measurable" results.
The only meaningful things I can point out at the moment are:
- Default settings for v3 are much worse than those in v2: enabling the module makes the image blown/burned much more (+desaturation). V2 starting point was ok-ish.
- Auto tune in v2 did a much better job of setting reasonable initial values.

NOTE: I'm not doing anything "HDR" related, I'm just replacing basecurve with filmic.

aurelienpierre:
@jakubfi btw v2 uses the RGB decorrelated version by default (no chroma preservation), which resaturates lowlights and desaturates highlights (same as basecurve), making further saturation adjustements difficult (shadows always oversaturated). v3 uses the chroma preservation by default which does not affect saturation and looks probably less pleasing by default but makes further edits more consistent (homogenous saturation along the dynamic range). 

For the auto-tuning helpers, v2 works better *in some cases* but is really unreliable with X-Trans sensors and high ISO pictures. Don't hesitate to rescale the dynamic range once you triggered an auto-tuning.

I will see what I can do for the default params.

TurboGit:
I suppose you need to rebase the commits and do some clean-up now that toneequalizer is merged.

aurelienpierre:
@TurboGit the pure C algorithm is ready unless someone has another bug to report. I plan to resolve conflicts tomorrow and have the OpenCL variant ready by the end of the week. But maybe it's better to merge the C version asap, validate the colour handling by the algo, then write the OpenCL.

TurboGit:
I'm ok for a merge before the OpenCL version.

aurelienpierre:
@TurboGit conflicts solved and commit squashed. Let's see if it passes CI.

TurboGit:
All good to me! Code is clean and this is working fine on my side. Another one for next major release. Very impressed by all the energy and work you've put on this. Thanks.

bastnic:
A few notes after a 5 minutes test:
- I do have the same bug as https://github.com/darktable-org/darktable/pull/2938#issuecomment-531486524: with sqare-ish curve at first open that disappear whith a short trip to lighttable.
- As signaled https://github.com/darktable-org/darktable/pull/2938#issuecomment-531789618: I do have a way more pleasing result with auto tune level than filmic v2. Just moving a little the grey point and the lattitude and that's a good starting point.

chhil:
I had filmic v2 in my history from a previous 2.7.0 build. Now with the current build the old filmic is gone and the new filmic rgb is present.

So should the history still show filmic  even though its no longer there?
I was expecting the fimic rgb to be activated and have some values migrated from the filmic v2 but its deactivate in the pipeline.

Is some clean up necessary?
Using darktable-2.7.0+1834~gae1ec2687-win64 currently.

TurboGit:
> So should the history still show filmic even though its no longer there?

It does, if an old edit is opened then filmic will be present if it was used. But it cannot be used on new edits. For this one need to use the new filmic module.

bastnic:
Just confirm that it does on edit from last wwek with filmic v2.

chhil:
Does the filmic v2 effect remain or is it like a no operation in the current build?
Once filmic v2 in an old edit, the new one needs to be added and tuned again?
Does the new one need to be moved to an appropriate position in the pipeline ?

bastnic:
on an old edit with filmic v2 : filmic v2 is enabled and editable, works fully as before. You see both filmic (v2 : `filmic` and v3 : `filmicrgb`) modules and only the v2 is enabled.

on new edit : you can't configure filmic v2 as you can't see it. V3 is the only option.

on new edit with a preset with filmic v2 enabled : filmic v2 is applied and fully functional. You can disable it and go with v3 but you start from scratch and this is two modules that only share part of the name.

I see no regression whatsoever (for now).

chhil:
I am not seeing what you are saying. I will do some fresh tests tomorrow across versions with clean a clean db..
Image has filmic v2 used, but in the modules I only see filmic v3 .  Hence I raised my query.
Thank you @bastnic for the explanation.

image

bastnic:
Hmm @chhil I see. Same comportement here as your screen. If I search on an old edit / new with preset with v2 enabled, I see filmic and i can act on it BUT I cannot find it by searching "film" in the module search bar. It's a bug on the search component not on filmic. 

My guess is that search only purpose not deprecated module, without following the list of modules enabled on the current edit.

chhil:
@bastnic , you are correct, its there in the "active modules" tab. So its not searchable but there. As long as the behavior is clear and maybe documented , I am fine with it. Thanks again.

bastnic:
@chhil nevertheless, I opened an issue.

PR #2940, opened 2019-09-14
lighttable : always change offset from other views

### LINKED ISSUE #2921: switching from darkroom to lighttable does not display current photo
When I switch from darkroom to lighttable, the current photo is not displayed.  The photos displayed are where I started working, not where I am currently.

I would expect to see the current/selected item visible.

openSUSE Tumbleweed 20190905
darktable-2.7.0~git1675.94549e1a4

AlicVB:
Otherwise, only the first offset change is taken in account.
this should fix #2921


AlicVB:
As issue #2921 also reference another bug : 

> fwiw: having a not rejected photo selected in lightroom and switching view parameters from "all" to "all except rejected" moves the "selected" image off lighttables display. Also the reverse does the same, "all except rejected" to "all".

I've added a corresponding fix here too in a second commit.

TurboGit:
Looks good! Thanks.

PR #2941, opened 2019-09-14
darkroom : small refactor of image change code

AlicVB:
so image change with filmstrip or keys launch same actions.
changes : 
- filmstrip : avoid double ratio compute
- keys : disable picker + clear undo list


TurboGit:
Looks good! Thanks.

PR #2364, opened 2019-04-09
[WIP] update noise profiling tool

rawfiner:
This is a work in progress to have a new noise profiling tool without the flaws of the previous one.
What is done:
- update the noise profiling tool
What is not done:
- handle the fact that we can have several profiles for the same camera in denoiseprofile's code and json file. For this, I would like advices: how should we modify the noiseprofile.json file? There is a "version" number in the json file but I am not sure it can be used for this

Note that I cherry-picked a commit from edgardoh's pull request #2319 for this, it allows to export an image corresponding to the data we get in denoiseprofile, i.e. data without any colorin and colorout profiles.

How to use:
- you need to take REALLY blurry pictures. While previous noise profiling tool tolerated well not so blurry pictures, this one is much more sensitive to it. As a hint, open a picture, and enable rawdenoise with noise threshold set to the maximum (you can enter the value 1 with right clic), and with "difference" blending mode: you should see only noise.
- the pictures should be underexposed, and overexposed if possible (but underexposure is more important that overexposure)
- once you have the pictures, you can launch the noise profiling tool as usual (assuming you are in the directory of your pictures): darktable-gen-noiseprofile -d .
- you should probably indicate to the tool where darktable-cli is. You can for instance set absolute paths in darktable-gen-noiseprofile

PR #2946, opened 2019-09-16
Change variance stabilization transform of denoiseprofile and add an auto mode to denoiseprofile

rawfiner:
This PR implements in denoiseprofile a different variance stabilization transform (VST): the one from #2364 
This VST is an extension of the old VST which was a generalized anscombe transform.
The new VST writes variance as V(X) = a(E[X]+b)^p
The old VST corresponds to the case where p = 1

This gives a new parameter, p, which allows to make denoising stronger on shadows than on highlights for instance.
I think this parameters helps a lot for high ISOs.

In addition to this, the default params of the modules are estimated automatically from the profile, so that the user has a good setting of the parameters to start with.
Basically, with these, the user only has to set the module on to get a good denoising result.

Compared to #2364 , this PR guesses the p parameter, whereas in #2364 the p parameter was found through profiling.

This is work in progress, the work remaining is:
- some code cleaning and comments
- port the opencl part

So you can start testing on CPU implementations if you want :-)

### DIFF
--- a/src/iop/denoiseprofile.c
+++ b/src/iop/denoiseprofile.c
@@ -2956,6 +2956,7 @@ void init(dt_iop_module_t *module)
   }
   tmp.fix_anscombe_and_nlmeans_norm = TRUE;
   tmp.wb_adaptive_anscombe = TRUE;
+  tmp.upgrade_vst = TRUE;
   memcpy(module->params, &tmp, sizeof(dt_iop_denoiseprofile_params_t));
   memcpy(module->default_params, &tmp, sizeof(dt_iop_denoiseprofile_params_t));
 }
@@ -3145,6 +3146,9 @@ static void mode_callback(GtkWidget *w, dt_iop_module_t *self)
       gtk_widget_hide(g->box_wavelets);
       gtk_widget_hide(g->box_variance);
       gtk_widget_show_all(g->box_nlm);
+      gtk_widget_set_visible(g->radius, FALSE);
+      gtk_widget_set_visible(g->nbhood, FALSE);
+      gtk_widget_set_visible(g->scattering, FALSE);
       break;
     case 2:
       p->mode = MODE_WAVELETS;
@@ -3165,7 +3169,10 @@ static void mode_callback(GtkWidget *w, dt_iop_module_t *self)
       gtk_widget_show_all(g->box_variance);
       break;
   }
-  gtk_widget_set_visible(g->overshooting, (p->mode == MODE_NLMEANS_AUTO) || (p->mode == MODE_WAVELETS_AUTO));
+  gboolean auto_mode = (p->mode == MODE_NLMEANS_AUTO) || (p->mode == MODE_WAVELETS_AUTO);
+  gtk_widget_set_visible(g->shadows, !auto_mode);
+  gtk_widget_set_visible(g->bias, !auto_mode);
+  gtk_widget_set_visible(g->overshooting, auto_mode);
   dt_dev_add_history_item(darktable.develop, self, TRUE);
 }
 
@@ -3253,6 +3260,9 @@ void gui_update(dt_iop_module_t *self)
       gtk_widget_hide(g->box_wavelets);
       gtk_widget_hide(g->box_variance);
       gtk_widget_show_all(g->box_nlm);
+      gtk_widget_set_visible(g->radius, FALSE);
+      gtk_widget_set_visible(g->nbhood, FALSE);
+      gtk_widget_set_visible(g->scattering, FALSE);
       break;
     case MODE_WAVELETS:
       combobox_index = 2;
@@ -3278,7 +3288,10 @@ void gui_update(dt_iop_module_t *self)
       break;
   }
   dt_bauhaus_combobox_set(g->mode, combobox_index);
-  gtk_widget_set_visible(g->overshooting, (p->mode == MODE_NLMEANS_AUTO) || (p->mode == MODE_WAVELETS_AUTO));
+  gboolean auto_mode = (p->mode == MODE_NLMEANS_AUTO) || (p->mode == MODE_WAVELETS_AUTO);
+  gtk_widget_set_visible(g->shadows, !auto_mode);
+  gtk_widget_set_visible(g->bias, !auto_mode);
+  gtk_widget_set_visible(g->overshooting, auto_mode);
   if(p->a[0] == -1.0)
   {
     dt_bauhaus_combobox_set(g->profile, 0);
--- a/src/iop/denoiseprofile.c
+++ b/src/iop/denoiseprofile.c
@@ -47,6 +47,15 @@
 #define DT_IOP_DENOISE_PROFILE_RES 64
 #define DT_IOP_DENOISE_PROFILE_BANDS 5
 
+// the following fulcrum is used to help user to set shadows and strength
+// parameters.
+// applying precondition on this value will give the same value even
+// if shadows slider is changed, as strength will be adjusted to
+// guarantee that.
+// from a user point of view, it separates "shadows" area from the rest
+// of the image.
+#define DT_IOP_DENOISE_PROFILE_P_FULCRUM 0.05f
+
 typedef enum dt_iop_denoiseprofile_mode_t {
   MODE_NLMEANS = 0,
   MODE_WAVELETS = 1,
@@ -1296,7 +1305,7 @@ static void process_wavelets(struct dt_iop_module_t *self, dt_dev_pixelpipe_iop_
   const float aa[3] = { d->a[1] * wb[0], d->a[1] * wb[1], d->a[1] * wb[2] };
   const float bb[3] = { d->b[1] * wb[0], d->b[1] * wb[1], d->b[1] * wb[2] };
 
-  const float compensate_p = 0.05f / powf(0.05f, d->shadows);
+  const float compensate_p = DT_IOP_DENOISE_PROFILE_P_FULCRUM / powf(DT_IOP_DENOISE_PROFILE_P_FULCRUM, d->shadows);
   if(!d->upgrade_vst)
   {
     precondition((float *)ivoid, (float *)ovoid, width, height, aa, bb);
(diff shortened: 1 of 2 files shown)


rawfiner:
Forgot to say, denoiseprofile should be moved before colorin to get better results (which will be the case once #2905 is merged)

rawfiner:
opencl now supported.
What remains now is code cleanup and comments.

junkyardsparkle:
Seems to be working great! The new single "preserve shadows" slider seems more intuitively usable than the previous implementation, and the default was good on the images I tested. I did notice that the slider returns to the value of 1.0 when double-clicked, rather than the default which results from clicking the module's reset button. I'm not sure what the best behavior for this would be, though.

rawfiner:
@junkyardsparkle thanks for testing and for the feedback. For the sliders reset, I don't know if it is better to reset the value to the "guessed" setting, which varies depending on the profile, or to the "standard" setting.
@TurboGit maybe do you have an advice on what would be the best approach for this question?

junkyardsparkle:
One other aspect of the "guessing" behavior is that presets currently hardcode this value, which makes it harder to create a single preset that can be (optimally) applied to images of different ISO ratings.

I'm not sure how this would be best handled either, since not being able to override the guessing with a preset when you *want* to would also be an annoyance, I suppose? Unless that slider was clearly related to the automatic ISO profile selection in the GUI, and understood to not be "sticky" in presets?

If the "guessing" is expected to always be as good as what I observed for my camera, I wouldn't personally disagree with this approach... but if it's expected to be something that *needs* adjusting in some cases, then maybe not such a great idea.

rawfiner:
@junkyardsparkle For presets, I think it is best to keep the behaviour with the values hardcoded, and not guessed. Guessed values probably needs some adjustment in some cases, for example to decrease the scattering slider to get back some local contrast if its value is too high. Another example where guessed values "fail" is for very underexposed images, which is expected.

junkyardsparkle:
That makes sense (to be honest, I actually didn't yet have a chance to notice that the scattering was also guessed). I tend to use a workflow which applies the profiled denoising only as a dtstyle in the export module, once I have one that I like. Up to now, this worked because the only ISO-dependent parameter was auto-selected from the profile. If more parameters become ISO-dependent but hard-coded, it would slightly complicate this. I don't know how common this workflow is, but anyway, I hope that makes it a little more clear where I'm looking at it from.

It's possible that it's time for me to consider changing this workflow anyway. :-)

rawfiner:
Ok, I better understand your need now. I have to think about it. Maybe if I add an "nlmeans auto" mode in the list of modes, where everything is set automatically, would that be a solution?
If it's ok for you I will wait for this PR to be integrated before investigating this issue, this will give some time to think about the best solution and reduce risks.

junkyardsparkle:
I just wanted to point out one possibly overlooked use case, since I think much of the discussion on forums etc. tends to be biased towards people spending a lot of time applying elaborate processing to a single image, but many of darktable's strengths are related to efficient batch workflow. If there's a way for a single dtstyle to be adaptive with respect to ISO, I would consider that a good thing... but the best thing would probably be to get some more opinions from a larger test group.

junkyardsparkle:
...but yes, I certainly don't mean to stall this PR. :-)

parafin:
This automatic stuff is a general problem in dt without clean idea on how to fix it in general yet. I think in this case "auto" mode can be a solution. E.g. it can still display all sliders, but as soon as user touches any of them, mode switches to manual. Logic will be quite transparent then IMHO. And you would strongly suggest to get it done before dt release.

rawfiner:
Thanks for your comments.
As a first trial, I have just added profile adaptive presets. This allows to handle underexposure as well.
But maybe indeed just having an auto mode with just an underexposure, a bias, and a force slider would be better?

junkyardsparkle:
> One question: should this mode be in the same combobox as nlmeans and wavelets (in that case it would be called "nlmeans auto" and "wavelets auto"), or in a different combobox?

The auto mode might be most "discoverable" with its own widget, but I don't know if it's worth the extra space or not.

Also wanted to say that after getting some sleep, I realized that I could get good results from a single dtstyle with a "compromise" value for the "preserve shadows" slider (and I subjectively prefer higher values for scattering and central pixel weight anyway), so the issues I mentioned above are probably less important than I thought (to me, but I don't tend to use the extreme part of the ISO range).

rawfiner:
I added 2 auto modes, one for non local means and one for wavelets.
So you will be able to do some style or presets of "auto" settings ;-)

I have a small issue though, when adjusting the parameter that controls autosetting the other ones, I get a message on screen, like the ones we get if we change a parameter from keyboard shortcut.
Any ideas if it is possible to avoid that?
Capture

rawfiner:
I think I am done with the last changes (auto modes and sliders defaults values are computed from profile).
I have rebased on master.
You can test and review if you want :-)

junkyardsparkle:
Hmm, did you update the rawspeed module when you rebased? Building halts with errors there for me...

junkyardsparkle:
Seems to be working well for me, except that the other recent pixelpipe order commits to master seem to not be playing nicely with already-imported images (new imports are fine, I used those for testing). I only get the pop-up alert when adjusting that control for wavelet automode, but not nlmeans, if that's any help...

rawfiner:
@junkyardsparkle thanks for spotting the pop-up in wavelets mode, I forgot to test that slider in wavelets mode. It should be better now.
What do you mean that already imported images have problems? On my side, edits from 2.6.2 imports fine.

junkyardsparkle:
Yes, the issue only affects a specific case, and it doesn't have anything to do with this PR (AFAIK), it just happened to appear along with it in the last build. I wrote up a description in issue #2954 if you want the gory details. :-)

rawfiner:
Ok, I wasn't sure if you meant something related to this PR on not, thanks for clarifying :-)

TurboGit:
> @TurboGit maybe do you have an advice on what would be the best approach for this question?

Frankly I have no preference for the reset of the slider. Both solutions sounds good, maybe resetting to the guessed value makes more sense as anyway we can reset to 0 with the "reset" module. So this gives both solutions.

TurboGit:
You need to update the two presets in init_presets(). The new base64 encoded values are displayed on the console to ease the update.

src/iop/denoiseprofile.c
```diff
@@ -3227,6 +3764,18 @@ static void fix_anscombe_and_nlmeans_norm_callback(GtkWidget *widget, dt_iop_mod
   dt_dev_add_history_item(darktable.develop, self, TRUE);
 }
 
+static void upgrade_vst_callback(GtkWidget *widget, dt_iop_module_t *self)
+{
+  if(darktable.gui->reset) return;
+  dt_iop_denoiseprofile_params_t *p = (dt_iop_denoiseprofile_params_t *)self->params;
+  dt_iop_denoiseprofile_gui_data_t *g = (dt_iop_denoiseprofile_gui_data_t *)self->gui_data;
+  p->upgrade_vst = gtk_toggle_button_get_active(GTK_TOGGLE_BUTTON(widget));
+  gboolean auto_mode = (p->mode == MODE_NLMEANS_AUTO) || (p->mode == MODE_WAVELETS_AUTO);
```
TurboGit: const

src/iop/denoiseprofile.c
```diff
@@ -107,20 +118,40 @@ typedef struct dt_iop_denoiseprofile_params_v6_t
...
   float central_pixel_weight; // increase central pixel's weight in patch comparison
+  float overshooting; // adjusts the way parameters are autoset
   float a[3], b[3]; // fit for poissonian-gaussian noise per color channel.
   dt_iop_denoiseprofile_mode_t mode; // switch between nlmeans and wavelets
   float x[DT_DENOISE_PROFILE_NONE][DT_IOP_DENOISE_PROFILE_BANDS];
   float y[DT_DENOISE_PROFILE_NONE][DT_IOP_DENOISE_PROFILE_BANDS]; // values to change wavelet force by frequency
   gboolean wb_adaptive_anscombe; // whether to adapt anscombe transform to wb coeffs
   // backward compatibility options
   gboolean fix_anscombe_and_nlmeans_norm;
+  gboolean upgrade_vst;
```
TurboGit: As this is a state and not an action, I suppose that "upgraded_vst" would be a better name or maybe "use_new_vst"?

Very minor point I must say.

src/iop/denoiseprofile.c
```diff
@@ -176,18 +214,22 @@ typedef struct dt_iop_denoiseprofile_data_t
...
```
TurboGit: Same as above.

TurboGit:
Quick review and did some tests. The new version is a bit better even if very subtle. At 200% we can see more details so this is really nice!

rawfiner:
Thanks a lot for the review, I will do the changes tonight.
The effect is subtle if you denoise mid-range ISO images, but much more visible on very high ISO images I think ;-)
This is because for medium ISO, the shadows parameter is close to 1, so we get the same transform as before, whereas for high ISO the shadows value is < 1.

Also, I don't know if you compared using the presets, but these presets are not the best way anymore to get good results: using only one instance of non local means gives much better results than the 2 wavelets instances of the presets.
Maybe should we mark the presets as deprecated?

TurboGit:
> Also, I don't know if you compared using the presets,

No I have compared with "new auto" mode and tuned version to get the best I could. My picture was 1600 ISO on a So,y A550.

I think we should keep the presets, maybe not the best... but people are used to have those presets and after all a preset is just that "a preset" and it is often needed to fine tune after anyway as for most modules BTW.

rawfiner:
>
> No I have compared with "new auto" mode and tuned version to get the best I could. My picture was 1600 ISO on a So,y A550.

So that explains the very small difference, as this camera at 1600 ISO has a similar profile to my fuji at 3200 ISO, and I start to see significant improvements ISO values strictly higher than 12800 ISO. So probably on your camera you will see improvements only at values strictly higher than 6400 ISO.
> 
> I think we should keep the presets, maybe not the best... but people are used to have those presets and after all a preset is just that "a preset" and it is often needed to fine tune after anyway as for most modules BTW.

Ok, let's keep them.

rawfiner:
Presets updated and the small code changes are made :-)
I also rebased on master.

TurboGit:
Ok, all good to me now. Tested again. Code is ok, I have just did a minor edit of a label.

rawfiner:
Thanks for the code reviews and testing! :-)

PR #2947, opened 2019-09-16
iop: Change the haze removal default values

cryptomilk:
The haze removal default values are in 99% of the cases too strong. This
changes to some better defaults picked by experience.


TurboGit:
I'm all for this !

cryptomilk:
@rabauke ???

Nilvus:
> I'm all for this !

I agree. Most of the time, I use 0.2 or 0.35 instead of 0.5 force, so I totally agree with that choice. Default ones are actually too strong.

TurboGit:
Ok, vote is in favor of this. Let's go for it. Thanks.

PR #2939, opened 2019-09-14
Fixing history, easier to read than numerous commits in PR #2898

jenshannoschwalm:
This pr is a replacement for #2898 to make testing and review easier.

Current history management has two significant bugs most obvious after compression in lighttable mode. There are several issues, see #2774, #2420, #2514, #2775

1. At several places there might be 'leaks' in the history stack, there can be invalid items above dev->history_end or NIL items at the top of stack.

2. The compression code is simply wrong for use within lighttable mode because it uses the global history stack with darktable.development->history_end without making sure the history for a specific image is used.

This pr 

1. takes care for the leaks in development.c
2. makes sure, the correct history stack for an image is used
3. uses code for compression in common/history.c for lighttable & darkroom history compression.
4. Adds a checkbox when compressing history for selections 

@TurboGit -- hope this helps for reviewing / testing.

PR #2949, opened 2019-09-16
Fixing history compression and closely related stuff in development.

jenshannoschwalm:
What does this include?
----------------------------------------------------------------------------------

** Fixing two leaks when writing history or adding an item to the history stack **

1) dt_dev_write_history_ext(dt_develop_t *dev, const int imgid)
   Should write all available data on the history stack to the database.
   Also it should write the correct history_end to main.images for the specified image.
   This end was dev->history_end, it must be the number of written items instead.

2) _dev_add_history_item_ext puts an item on top of the history stack.
   It only adds a *new* item if necessary, if appropriate it *replaces* the top item.
   We check for push-or-replace but before that check we must remove
     a) items above history_end
     b) NIL items.

** Managing the history database and stack **
1) dt_history_compress_on_image
   MUST ONLY use the database itself.
   No touch on the history stack like reloading or writing!
   There is only one exception, we *load* the history stack for the image at start

2) dt_history_compress_on_selection
   Does the database compression
   After that it re-reads the history stack and writes it back to ensure a proper size.

3) dt_image_synch_xmp(imgid);
   Is *not* done in the common compression code but seperately in the darkroom callback
   and for every image in the selection.

4) _lib_history_compress_clicked_callback
   Is allowed so and *must* reload the history stack as this happens in darkroom
   and we want the changes to be valid at once.
   It also rises a DT_SIGNAL_DEVELOP_HISTORY_CHANGE and sets the module groups.

5) The extra compress_and_reload funtion did not help at all.

6) The lightroom button for only one image needs extra care.

### DIFF
--- a/src/develop/develop.c
+++ b/src/develop/develop.c
@@ -765,6 +765,8 @@ int dt_dev_write_history_item(const int imgid, dt_dev_history_item_t *h, int32_t
 static void _dev_add_history_item_ext(dt_develop_t *dev, dt_iop_module_t *module, gboolean enable, gboolean no_image, gboolean include_masks)
 {
     GList *history = g_list_nth(dev->history, dev->history_end);
+    // look for leaks on top of history in two steps
+    // first remove obsolete items above history_end
     while(history)
     {
       GList *next = g_list_next(history);
@@ -774,6 +776,10 @@ static void _dev_add_history_item_ext(dt_develop_t *dev, dt_iop_module_t *module
       dev->history = g_list_delete_link(dev->history, history);
       history = next;
     }
+    // then remove NIL items there
+    while ((dev->history_end>0) && (! g_list_nth(dev->history, dev->history_end - 1)))
+      dev->history_end--;
+
     history = g_list_nth(dev->history, dev->history_end - 1);
     dt_dev_history_item_t *hist = history ? (dt_dev_history_item_t *)(history->data) : 0;
     if(!history // if no history yet, push new item for sure.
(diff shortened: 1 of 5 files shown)


jenshannoschwalm:
After two days of thinking and relaxing i think i know what i did wrong.

See discussion of #2939 and #2898 

@TurboGit and @MStraeten -- try again please.

TurboGit:
I'll test... If others want to test too and report it would be nice as this is a delicate part ! Thanks.

jenshannoschwalm:
It's indeed at least delicate.

BTW, there is defenitely one remaining bug i am aware of in history compression - has been in master since historic times.
The masks ownership is lost when compressing history. (Probably related to src/common/history.c lines 923-927)

That is another delicate part, i will work on that later with another pr, ok.

MStraeten:
quick check: no crash and multiple applied styles are compressed fine ...

chhil:
Is this related ?
https://github.com/darktable-org/darktable/issues/2952#issuecomment-532255915

TurboGit:
@chhil : No, doesn't look related to me.

TurboGit:
For me this works fine too. So I'll merge, better sooner. Thanks for all the work here, very appreciated on this delicate area.

TurboGit:
Sadly just found a regression:
- open a developed image on darkroom
- select a line in the history module
- select another image on the filmstrip
- go back to original image

=> the history end is set at the top of the history and not at the item selected

jenshannoschwalm:
Ok, i will do my work and fix it.  





On Thu, Sep 19, 2019 at 3:41 PM +0200, "Pascal Obry" <notifications@github.com> wrote:












Sadly just found a regression:

open a developed image on darkroom
select a line in the history module
select another image on the filmstrip
go back to original image



=> the history end is set at the top of the history and not at the item selected

src/develop/develop.c
```diff
@@ -1206,17 +1212,20 @@ void dt_dev_write_history_ext(dt_develop_t *dev, const int imgid)
   sqlite3_step(stmt);
   sqlite3_finalize(stmt);
   GList *history = dev->history;
+  // the history size is NOT defined by current history_end but by the number actually written
```
TurboGit: Ah, did miss that and this is the issue. The history_end must be always kept as-is. The user can set an old history and keep it as long as he wants until the history is compressed. So this part is probably to be reverted.

PR #2958, opened 2019-09-18
full preview : add keys for sticky preview + update doc

AlicVB:
default key are now same as for "classic" preview + alt key : 
sticky preview : alt+w
sticky preview with focus detection : ctrl+alt+w

I've also updated the doc accordingly (and correct preview default keys which were not 'z' anymore)

### DIFF
--- a/doc/usermanual/lighttable/lighttable.xml
+++ b/doc/usermanual/lighttable/lighttable.xml
@@ -316,7 +316,7 @@
       <title>Full preview</title>
 
       <para>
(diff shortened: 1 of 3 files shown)


bastnic:
thanks.

bastnic:
maybe also update the RELEASE_NOTE file which talk about w ant ctrl+w ?

TurboGit:
Looks good, working fine. Thanks.

PR #2961, opened 2019-09-19
Fixing the regression introduced by pr #2949

jenshannoschwalm:
I commented in that pr
   dt_dev_write_history_ext(dt_develop_t *dev, const int imgid)
   Should write all available data on the history stack to the database.
   Also it should write the correct history_end to main.images for the
   specified image.
   This end was dev->history_end, it must be the number of written items
   instead.

As @TurboGit detected this was wrong.

### DIFF
--- a/src/develop/develop.c
+++ b/src/develop/develop.c
@@ -1212,20 +1212,18 @@ void dt_dev_write_history_ext(dt_develop_t *dev, const int imgid)
   sqlite3_step(stmt);
   sqlite3_finalize(stmt);
   GList *history = dev->history;
-  // the history size is NOT defined by current history_end but by the number actually written
-  int hist_size = 0;
+
   for(int i = 0; history; i++)
   {
     dt_dev_history_item_t *hist = (dt_dev_history_item_t *)(history->data);
     (void)dt_dev_write_history_item(imgid, hist, i);
     history = g_list_next(history);
-    hist_size++;
   }
 
   DT_DEBUG_SQLITE3_PREPARE_V2(dt_database_get(darktable.db),
                               "UPDATE main.images SET history_end = ?1, iop_order_version = ?3 WHERE id = ?2", -1,
                               &stmt, NULL);
-  DT_DEBUG_SQLITE3_BIND_INT(stmt, 1, hist_size);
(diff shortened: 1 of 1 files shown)


jenshannoschwalm:
Working now as it should.

There is one thing worrying me; if we leave the history_end somewhere in the middle of the history stack while developing, forget about it and later do a lighttable history compression we might accidently lose stuff not intended.

1. Shall we add a checkbox as we have in discard? Another pr ?
2. Should we default to top_of_history_stack when doing lighttable compression? Another pr?

TurboGit:
Quite good question !

I don't think 2 is safe, it means that from the lighttable we do change the development of a picture by changing the module used.

Point 1, I think you mean a dialog? Seems the way to go. I would:
- detect if in the selected image or collection there is a history_end not at the top
- in this case (and only in this case) as for confirmation about the compress

How does that sounds?

In another PR anyway. Thanks.

TurboGit:
Looks good, merging. Thanks.

jenshannoschwalm:
> Point 1, I think you mean a dialog? Seems the way to go. I would:
>  detect if in the selected image or collection there is a history_end not at the top
    in this case (and only in this case) as for confirmation about the compress
> How does that sounds?

Mmh, i would certainly delay or break the workflow. And what would be the message for the confirmation box? Showing a thumbnail? But do you really remember what picture that would be in such a situation?

What about:

- When compressing a selection we test for every image : is history_end at the top of history stack?
- if not-> set a tag darktable|problems|history
- If so -> compress at once and clear that flag
- After compression the selection you get a hint: There are remaining uncompressed histories, look for the above tag. So you could look for those pictures in darkroom and decide what to do.
- Every image would clear that tag when developed in darkroom.

TurboGit:
That sounds like a plan to me!

Indeed, the tag is very handy in this case and let the user a chance to review all images.

I would just use a more descriptive tag name, maybe:

   darktable|cannot compress history
or
   darktable|compress history locked

Or any better proposal you could come with. That's a nitpick anyway.

jenshannoschwalm:
I would favor darktable|problem|history-compression

The darktable|problem|xxxx tree could be a way to mark any sort of problem detected by darktable.

TurboGit:
Ok, fine with me.

PR #2966, opened 2019-09-20
usermanual: document double-click on a copy history dialog.

TurboGit:


doc/usermanual/lighttable/panels/history.xml
```diff
@@ -47,6 +47,8 @@
         Copy the history stack of the selected image. You will be prompted for which items are
         to be include. If more than one image is selected, the history stack is taken from the
         image that has been selected first.
+
+        A double-click on an item copies this one only and close the dialog.
```
Balise42: I'd go for "copies that item" (but then I'm somewhat fuzzy on this vs that, so....)

doc/usermanual/lighttable/panels/history.xml
```diff
@@ -47,6 +47,8 @@
         Copy the history stack of the selected image. You will be prompted for which items are
         to be include. If more than one image is selected, the history stack is taken from the
```
Balise42: to be included

brucewilliamsphotography: Copy the history stack of the selected image. You will be prompted for which items are to be included.

doc/usermanual/lighttable/panels/history.xml
brucewilliamsphotography: No, Balise42, you're spot on there. It should be "copies that item".

TurboGit:
Thanks!! That is merged now.

PR #2970, opened 2019-09-21
Again work on history.

jenshannoschwalm:
1) The lighttable history compression makes sure that **only** the images with
   history_end at the top are really compressed for safety reasons.

2) What about images that had no history compression?
   - You will get a message box telling you there were problems.
   - Also these images have a tag attached "darktable|problem|history-compress"
     so you can take care about those in darkroom
   - Whenever you open a picture in darkroom this tag will be automatically removed

3) there is the public dt_set_history_compress_problem(int32_t imgid, gboolean set)

4) There has been more work on dt_history_compress_on_image(int32_t imgid)
   It does **not** touch the history_stack at all!
   When we have a mask manager already there at position 0 we don't have to
   change num in main.history and history_end in main.images any more possibly
   avoiding leaks.

5) Also the dt_history_compress_on_selection() has major rework.
   - it returns the number of images that were not compressed as a helper for the
     lighttable button message box.
   - it does **not** load any history stack any more!
   - it does proper reordering of the history items doing a double
        _history_copy_and_paste_on_image_overwrite
     as i assume this has been tested well.
     So we have a properly ordered and leak-free at the end.

@TurboGit and @MStraeten  
Again delicate stuff i would want you to test carefully.
Of course there is the security check for lighttable mode.

But more important are the internal changes done for history compression. They **never** touch the
current history stack any more and the lighttable side generates properly ordered data in history without leaks.

The darkroom history compression does this reordering on it's own. 

Also, if you accept and like the code for reordering done in lighttable mode, could it make darkrooms mode safer? Probably not but i'm not sure.

### DIFF
--- a/src/common/history.c
+++ b/src/common/history.c
@@ -1074,7 +1074,6 @@ static void _history_reorder(int32_t imgid)
 int dt_history_compress_on_selection()
 {
   int uncompressed=0;
-  int32_t imgid = -1;
 
   // Get the list of selected images
   sqlite3_stmt *stmt;
@@ -1082,7 +1081,7 @@ int dt_history_compress_on_selection()
 
   while(sqlite3_step(stmt) == SQLITE_ROW)
   {
-    imgid = sqlite3_column_int(stmt, 0);
+    int imgid = sqlite3_column_int(stmt, 0);
     const int test = dt_history_end_attop(imgid);
     if (test == 1) // we do a compression and we know for sure history_end is at the top!
     {
--- a/src/common/history.c
+++ b/src/common/history.c
@@ -870,7 +870,7 @@ int dt_history_copy_and_paste_on_selection(int32_t imgid, gboolean merge, GList 
   return res;
 }
 
-void dt_set_history_compress_problem(int32_t imgid, gboolean set)
+void dt_history_set_compress_problem(int32_t imgid, gboolean set)
 {
   guint tagid = 0;
   char tagname[64];
@@ -955,7 +955,7 @@ void dt_history_compress_on_image(int32_t imgid)
   int masks_count = 0;
   char op_mask_manager[20] = { 0 };
 
-  bool manager_position = FALSE;
+  gboolean manager_position = FALSE;
   // do we already have a mask manager at the correct position? We don't want to increase history nums later
   g_strlcpy(op_mask_manager, "mask_manager", sizeof(op_mask_manager));
 
@@ -1047,11 +1047,11 @@ static void _history_reorder(int32_t imgid)
 
   if(no_forced_reordering)
   {
-    if (give_reorder_information) fprintf(stderr,", no action");
+    if (give_reorder_information) fprintf(stderr,", no action\n");
   }
   else
   {
-    if (give_reorder_information) fprintf(stderr,", reorder");
+    if (give_reorder_information) fprintf(stderr,", reorder\n");
 
     _history_copy_and_paste_on_image_overwrite(imgid, dummy, 0);
     _history_copy_and_paste_on_image_overwrite(dummy, imgid, 0);
@@ -1073,7 +1073,6 @@ static void _history_reorder(int32_t imgid)
 
 int dt_history_compress_on_selection()
 {
-  int test;
   int uncompressed=0;
   int32_t imgid = -1;
 
@@ -1084,10 +1083,10 @@ int dt_history_compress_on_selection()
   while(sqlite3_step(stmt) == SQLITE_ROW)
   {
     imgid = sqlite3_column_int(stmt, 0);
-    test = dt_history_end_attop(imgid);
+    const int test = dt_history_end_attop(imgid);
     if (test == 1) // we do a compression and we know for sure history_end is at the top!
     {
-      dt_set_history_compress_problem(imgid, FALSE);
+      dt_history_set_compress_problem(imgid, FALSE);
       dt_history_compress_on_image(imgid);
       _history_reorder(imgid);
  
@@ -1153,10 +1152,10 @@ int dt_history_compress_on_selection()
     if (test == 0) // no compression as history_end is right in the middle of history
     {
       uncompressed++;
-      dt_set_history_compress_problem(imgid, TRUE);
+      dt_history_set_compress_problem(imgid, TRUE);
     }
     if (test == -1)
-      dt_set_history_compress_problem(imgid, FALSE);
+      dt_history_set_compress_problem(imgid, FALSE);
   }
 
   sqlite3_finalize(stmt);
--- a/src/common/history.c
+++ b/src/common/history.c
@@ -870,34 +870,77 @@ int dt_history_copy_and_paste_on_selection(int32_t imgid, gboolean merge, GList 
   return res;
 }
 
-
-static void _clear_history_stack()
+void dt_set_history_compress_problem(int32_t imgid, gboolean set)
 {
-  while(darktable.develop->history)
-  {
-    dt_dev_free_history_item(((dt_dev_history_item_t *)darktable.develop->history->data));
-    darktable.develop->history = g_list_delete_link(darktable.develop->history, darktable.develop->history);
-  }
+  guint tagid = 0;
+  char tagname[64];
+  snprintf(tagname, sizeof(tagname), "darktable|problem|history-compress");
+  dt_tag_new(tagname, &tagid);
+  if (set)
+    dt_tag_attach(tagid, imgid);
+  else
+    dt_tag_detach(tagid, imgid);
 }
+
+static int dt_history_end_attop(int32_t imgid)
+{
+  int size=0;
+  int end=0;
+  sqlite3_stmt *stmt;
+
+  // get highest num in history
+  DT_DEBUG_SQLITE3_PREPARE_V2(dt_database_get(darktable.db),
+    "SELECT MAX(num) FROM main.history WHERE imgid=?1", -1, &stmt, NULL);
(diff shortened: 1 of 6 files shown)


TurboGit:
I'll try to find some time to review that soon.

src/common/history.c
```diff
@@ -870,48 +870,106 @@ int dt_history_copy_and_paste_on_selection(int32_t imgid, gboolean merge, GList
...
-  DT_DEBUG_SQLITE3_BIND_INT(stmt, 2, darktable.develop->history_end);
+  DT_DEBUG_SQLITE3_BIND_INT(stmt, 2, my_history_end);
   sqlite3_step(stmt);
   sqlite3_finalize(stmt);
 
-  // delete all mask_manager entries
   int masks_count = 0;
   char op_mask_manager[20] = { 0 };
+
+  bool manager_position = FALSE;
```
TurboGit: bool -> gboolean

TurboGit: bool -> gboolean

I don't remember all details (bool is C++?) but this is a style we enforce.

src/common/history.c
```diff
@@ -978,28 +1037,130 @@ void dt_history_compress_on_image(int32_t imgid)
...
+#define give_reorder_information FALSE
+static void _history_reorder(int32_t imgid)
+{
+  int32_t dummy = 0x7fffffff;
+  sqlite3_stmt *stmt;
+  if(give_reorder_information) fprintf(stderr,"\n\n_history reorder for image: %i",imgid);
+
+  if(no_forced_reordering)
+  {
+    if (give_reorder_information) fprintf(stderr,", no action");
```
TurboGit: \n at end of string

src/common/history.c
```diff
@@ -978,28 +1037,130 @@ void dt_history_compress_on_image(int32_t imgid)
...
+  }
+  else
+  {
+    if (give_reorder_information) fprintf(stderr,", reorder");
```
TurboGit: \n at end of string

src/common/history.c
```diff
@@ -978,28 +1037,130 @@ void dt_history_compress_on_image(int32_t imgid)
...
   {
     imgid = sqlite3_column_int(stmt, 0);
-    dt_history_compress_on_image(imgid);
-    // Now read the history again into the stack
-    _clear_history_stack();
-    dt_dev_read_history_ext(darktable.develop, imgid, FALSE);
-    // writing ensures we have no gaps
-    dt_dev_write_history_ext(darktable.develop, imgid);
-    dt_image_synch_xmp(imgid);
+    test = dt_history_end_attop(imgid);
```
TurboGit: const int test =

src/common/history.c
```diff
@@ -978,28 +1037,130 @@ void dt_history_compress_on_image(int32_t imgid)
...
+    sqlite3_step(stmt);
+    sqlite3_finalize(stmt);
+  }
+}  
+#undef give_reorder_information
+#undef no_forced_reordering
+
+int dt_history_compress_on_selection()
 {
+  int test;
```
TurboGit: remove from here, see below.

src/common/history.c
```diff
@@ -978,28 +1037,130 @@ void dt_history_compress_on_image(int32_t imgid)
...
+
   // Get the list of selected images
   sqlite3_stmt *stmt;
-  DT_DEBUG_SQLITE3_PREPARE_V2(dt_database_get(darktable.db), "SELECT imgid FROM main.selected_images", -1, &stmt,
-                              NULL);
+  DT_DEBUG_SQLITE3_PREPARE_V2(dt_database_get(darktable.db), "SELECT imgid FROM main.selected_images", -1, &stmt, NULL);
 
   while(sqlite3_step(stmt) == SQLITE_ROW)
   {
     imgid = sqlite3_column_int(stmt, 0);
```
TurboGit: const int32_t imgid =

src/common/history.c
```diff
@@ -978,28 +1037,130 @@ void dt_history_compress_on_image(int32_t imgid)
...
+  }
+}  
+#undef give_reorder_information
+#undef no_forced_reordering
+
+int dt_history_compress_on_selection()
 {
+  int test;
+  int uncompressed=0;
   int32_t imgid = -1;
```
TurboGit: As we are at it, I would remove imgid decl from here, see below.

src/develop/develop.c
```diff
@@ -689,6 +689,9 @@ void dt_dev_load_image(dt_develop_t *dev, const uint32_t imgid)
   dt_dev_read_history(dev);
 
   dev->first_load = 0;
+
+  // Loading an image means we do some developing and so remove the darktable|problem|history-compress tag
+  dt_set_history_compress_problem(imgid, FALSE);
```
TurboGit: Select corresponding tag in collection to see all images with problem. Go to darkroom, then back to lighttable the count of image is not updated and the image is still displayed eventhough the tag has been removed. I think we should update the collection here.

src/libs/copy_history.c
```diff
@@ -152,30 +152,27 @@ static void copy_button_clicked(GtkWidget *widget, gpointer user_data)
...
 static void compress_button_clicked(GtkWidget *widget, gpointer user_data)
 {
-  const int img = dt_view_get_image_to_act_on();
+  const GtkWidget *win = dt_ui_main_window(darktable.gui->ui);
+  if (dt_collection_get_selected_count(darktable.collection) < 1 ) return;
 
-  if(img < 0)
-    dt_history_compress_on_selection();
-  else
+  int missing = dt_history_compress_on_selection();
```
TurboGit: const int missing =

src/libs/copy_history.c
```diff
@@ -152,30 +152,27 @@ static void copy_button_clicked(GtkWidget *widget, gpointer user_data)
...
-    // writing ensures we have no gaps
-    dt_dev_write_history_ext(darktable.develop, img);
-    dt_image_synch_xmp(img);
-    while(darktable.develop->history)
-    {
-      dt_dev_free_history_item(((dt_dev_history_item_t *)darktable.develop->history->data));
-      darktable.develop->history = g_list_delete_link(darktable.develop->history, darktable.develop->history);
+    GtkWidget *dialog = gtk_message_dialog_new(
+    GTK_WINDOW(win), GTK_DIALOG_DESTROY_WITH_PARENT, GTK_MESSAGE_QUESTION, GTK_BUTTONS_CLOSE,
+    ngettext("No history compression of an image.\nSee tag: darktable|problem|history-compress.",
```
TurboGit: I would use %d here too or 1 as this is what is done on other messages. Just for consistency.

src/libs/copy_history.c
```diff
@@ -152,30 +152,27 @@ static void copy_button_clicked(GtkWidget *widget, gpointer user_data)
...
+             "No history compression of %d images.\nSee tag: darktable|problem|history-compress.", missing ), missing);
+#ifdef GDK_WINDOWING_QUARTZ
+    dt_osx_disallow_fullscreen(dialog);
+#endif
+
+    gtk_window_set_title(GTK_WINDOW(dialog), _("History compression warning"));
```
TurboGit: No capital H for history. This is current style even if not something all people like. Again for consistency.

src/common/history.c
```diff
@@ -870,48 +870,106 @@ int dt_history_copy_and_paste_on_selection(int32_t imgid, gboolean merge, GList
   return res;
 }
 
+void dt_set_history_compress_problem(int32_t imgid, gboolean set)
```
TurboGit: To follow the naming convention it should be:
dt_history_set_compress_problem()

src/develop/develop.c
TurboGit: But this is not easy, updating the collection means that the tag will be collapsed as I'm not sure we can update count on collect modules without reseting/redrawing the whole module. So maybe it is good as-is.

jenshannoschwalm: I was not sure about this. Probably better to update the collection when closing darkroom?

src/libs/copy_history.c
jenshannoschwalm: Ok. Ugly but consistant :-)

src/develop/develop.c
TurboGit: But then you're back into lighttable and the collection is fully collapsed. So I'm not sure.

jenshannoschwalm:
All done. Thanks for your precise comments about coding style and using 'const' when possible. It won't be my last work on dt so probably not wasting time :-)

src/libs/copy_history.c
TurboGit: BTW, we don't want capital letters there too.

src/develop/develop.c
jenshannoschwalm: Ok, let me think about it

jenshannoschwalm: What about this. If it hurts too much i could test and possibly remove tag and do the recalculation just in case.

But i think this is good anyway. We might alter more tags while developing...

jenshannoschwalm:
For me this looks good now.

src/develop/develop.c
TurboGit: No this is not good to me. Select a tag in the collection module for example darktable|format|jpg go to darkroom back to lighttable and the "collect images" module tree is collapse. You need to reopen the tree... All context is lost!

Same if you select an entry in time, date.

TurboGit: EDIT: no I was not testing the wrong code :( Sorry, indeed all is good now. Thanks.

Issue #2480, opened 2019-04-29
darktable-curve-tool weird ILCE-7M3


pauledd:
Hi

While testing my new Sony ILCE-7M3 I stumbled upon the baseline tools directory and I wanted to create my own baseline based on that tool.
So for a simple test I shoot one RAW+JPG with some over- and underexposed part in it.
Test files: 

http://p-bg.de/pics/PBG00152.ARW
http://p-bg.de/pics/PBG00152.JPG

Then I did a:

`for raw in $(ls *.ARW); do /darktable/tools/darktable-curve-tool-helper "$raw";done`

which gave this output:

```
if you accumulated enough data, you can now output the curves with
darktable-curve-tool  -z -e PBG00132.ARW -n 16
```
then I did a:

`/darktable/tools/darktable-curve-tool -z -e PBG00132.ARW | tee mycameracurves.sh`

and executed that script. Now I have a new basecurve in darktable named "ILCE-7M3" but when I select it it looks not correct:
![Bildschirmfoto_2019-04-29_13-58-51](https://user-images.githubusercontent.com/12088907/56894736-0647aa80-6a87-11e9-8665-3eca1c03580b.png)

What I am doing wrong? 
In the camera settings I choose JPG to be maximum resolution (24MP) same as RAW, disabled any effects leaving it at "standart"... I looked in "darktable/src/external/adobe_coeff.c" and dcraw.c::adobe_coeff::table and they are the same...
Anyone any ideas?

pauledd:
I just noticed that I took pictures in APS-C mode so the files are wrong size. I shoot another pair in 24MP format and corrected my images above. 
To be sure not to make any mistakes I followed the manual steps:
```
dcraw -6 -W -g 1 1 -w PBG00152.ARW && convert PBG00152.JPG PBG00152-jpg.ppm
darktable-curve-tool PBG00152.ppm PBG00152-jpg.ppm && darktable-curve-tool -z
gnuplot plot.basecurve
```
That gives this plots for the baseline:

![Bildschirmfoto_2019-04-30_11-10-40](https://user-images.githubusercontent.com/12088907/56951684-c80ac380-6b38-11e9-934b-f5791b926745.png)
![Bildschirmfoto_2019-04-30_11-11-00](https://user-images.githubusercontent.com/12088907/56951685-c80ac380-6b38-11e9-8293-9436028a1f6c.png)

[...]

PR #2971, opened 2019-09-22
Fix the curve tool fit_curve initialization. (#2480)

mschulkind:
It seems that a bit over a year ago, a superfluous check for `pos < 0` was blindly fixed, which then ended up breaking this tool. The original code was a bit weird, since `pos` can't even be negative anyway, and then this "fix" made the code even stranger and entirely broken.

This PR fixes the initialization code so that it works again and removes all bounds checking in order to make the code more straightforward.

### DIFF
--- a/tools/basecurve/darktable-curve-tool.c
+++ b/tools/basecurve/darktable-curve-tool.c
@@ -470,18 +470,16 @@ fit_curve(CurveData* best, int* nopt, float* minsqerr, CurveSample* csample, int
   {
     if(i == 0 || drand48() < p_large)
     { // large step
-      for(int k=0;k<tent.m_numAnchors;k++)
+      for(int k=0;k<tent.m_numAnchors-1;k++)
       {
         float x = k/(tent.m_numAnchors-1.0f);
         x *= x*x; // move closer to 0
-        uint32_t pos = 0;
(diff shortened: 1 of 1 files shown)


mschulkind:
Pretty sure the CI failure is not due to this PR as the file I changed isn't even included in the CI build.

cryptomilk:
Are you sure that `x * CURVE_RESOLUTION` does not overflow and uint32_t?

mschulkind:
Yes, for sure. `x` varies between 0 and 1, so as long as `CURVE_RESOLUTION`, which is a constant, does not already overflow `uint32_t`, then the calculation certainly won't. The more important bound is the length of `curve`, which it is also guaranteed never to overflow since `x` is always less than 1 in that loop now. The old code was some weird attempt at trying to get `pos` as close to the highest index of `curve` without overflowing that array, but doing the calculation in the floating point domain, which is why there was a risk of overflow due to floating point precision. It was a buggy implementation of an already strange strategy.

cryptomilk:
LGTM.

mschulkind:
How do we get this merged?

aurelienpierre:
Yeah, tone equalizer is failing again with Clang, for some vectorization failure warning. Annoying.

aurelienpierre:
Works for me. Thanks !

PR #2974, opened 2019-09-22
Back out iop-order changes in v2.

TurboGit:
This version restore the original v2 version of the iop-order. Changes
on this order must also upgrade the iop-order version.


wpferguson:
Tested and it works, however....

If you have edited a file with the current git, it preserves the corrupt order so I selected all images, then did a remove, got rid of the xmp files, then reimported.  Bloom and color lookup table both work correctly.  I also opened files edited with my older git version from before Sept 16, and those displayed correctly with this PR.

TurboGit:
Yes, nothing magic here. I suppose that changing the order manually could work also.

So let me merge this now to avoid more mess.

Thanks a lot for testing.

PR #2976, opened 2019-09-23
Fix loading the tone equalizer module

cryptomilk:
Error:

```
[iop_load_module] failed to open operation `toneequal': 'process': /usr/lib64/darktable/plugins/libtoneequal.so: undefined symbol: process
```

The documentation for target_clones states:

> It also creates a resolver function (see the ifunc attribute above) that dynamically selects a clone suitable for current architecture. The resolver is created only if there is a usage of a function with target_clones attribute. 

Looking at the elf the lib only provides a process.resolver symbol. So the dynamic loader doesn't find a process symbol.

### DIFF
--- a/src/iop/toneequal.c
+++ b/src/iop/toneequal.c
@@ -828,7 +828,8 @@ static inline void apply_toneequalizer(const float *const restrict in,
 
 
 __DT_CLONE_TARGETS__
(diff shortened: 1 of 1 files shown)


TurboGit:
Looks good! Thanks.

PR #2977, opened 2019-09-23
Prevent warnings from being treated as errors

aurelienpierre:
Prevent warnings from being treated as errors for Release and RelWithDebInfo builds. Devs will need to build Debug to catch warnings. I'm not sure it's a good idea but since Clang complains way more than GCC, it might be a way to make them behave the same.

Re-enable OpenMP for Clang users in tone equalizer.

I'm abusing Travis CI to test that because issue does not reproduce on Clang 7 or 8 on my system.


hellbuster:
Only the vectorization warning was giving problems, right? In that case only that warning should be "disabled". It is true that Clang is pickier, but that tenda to be a good things, because it catches more fishy stuff in the code.

aurelienpierre:
Warnings are not disabled, they are still there, but they are not treated as errors anymore and, as such, are not blocking the building for Release builds. They are still blocking for Debug builds.

aurelienpierre:
by the way, Apple plateform already had a special handling where warnings were not treated as errors:
```
if(NOT SOURCE_PACKAGE AND NOT APPLE)
  # apple build fails on deprecated warnings..
  add_definitions(-Werror)
endif()
```
Only Clang on Linux has problems.

parafin:
macOS example isn't a good one to follow, if anything we should disable just deprecated warnings being treated as warnings there, not all of them. Anyway, what's the point of fixing just Release builds? Debug ones still have to work.

aurelienpierre:
Debug builds are used by people who have means to understand and fix errors, and might want building to stop at errors instead of having to browse thousands of lines of log to find the first error. Release builds are used by people who probably can't debug and don't care if some loop vectorization failed.

parafin:
I don't understand what problem this PR is trying to fix. If you fix Debug build, then Release will work without this PR too.

aurelienpierre:
The problem is Clang issues warnings that don't bother GCC, so the behaviour is not consistent between the 2, and these warnings are later treated as errors that makes the building fail. Among these warnings, 2 are particularly annoying because they are minor: loop vectorization misses (from `openmp simd` pragma) and unknown pragmas (because Clang tries to catch GCC-specific pragmas).

So the present PR keeps the current behaviour for Debug (aka let Clang faint like a princess for no reason and let devs deal with it), but makes warnings non-blocking (aka remove `-Werror` and ignore unknown pragmas) for Release and RelWithDebInfo, since they should not prevent the soft from working.

TurboGit:
This looks fine to me. Indeed, the fact that clang on MacOS and clang on Linux was not consistent is a bad thing and this explains why the error was not reproduced on a Mac. So this is just going in the right direction. Thanks for the investigation and fix Aurélien.

hellbuster:
You should only disable the warnings that are problematic.

Also, as @parafin is saying this is braking the Debug release for everybody. A developer may want to use that build for debugging something completely unrelated to vectorization but would be unable to.

It may a little more work, but I think that it is fine to remove the specific warning that is causing trouble for all builds, but other warnings should still be blocking. Removing all of them may make easier to introduce buggy code, unless the CI uses the Debug build.

aurelienpierre:
> Also, as @parafin is saying this is braking the Debug release for everybody. A developer may want to use that build for debugging something completely unrelated to vectorization but would be unable to.

The Debug is not broken, it is not affected by any change here (so it's just not fixed).

And again… Warnings **are not disabled**. They are just not treated as errors anymore. That's all.

hellbuster:
@aurelienpierre I understand that they are not actually disabled (sorry for the not acurrate language here), but the question here is: why not only change behaviour for only the problematic warning?

Erroring on warnings is there to try to prevent people to submit fishy code, so unless the CI uses the Debug build, this may enable people to submit "worse quality" code that before wouldn't pass.

aurelienpierre:
> why not only change behaviour for only the problematic warning?

Because it seems Clang and GCC use different warning flags and we haven't been able to catch the exact one responsible for the current false-positives.

parafin:
We either support clang or we don't. If we do, then warnings should be fixed (or exact warnings disabled), if we don't, then remove -Werror just for clang and forget about it. I assume we don't want the second choice.
It looks to me, that what actually happened here is that CI test was rightfully failing, and you disabled that test together with a lot of others (other compiler warnings) to make it pass again. That is not "going in the right direction".

aurelienpierre:
> warnings should be fixed

So what's the difference between errors and warnings then ?

Blocking compilation for a non-vectorized loop is not rightfully failing, it's being a sassy bitch. What is the flag to bypass it ? `-Wno-error=loop-vectorize` does not work.

aurelienpierre:
Also https://embeddedartistry.com/blog/2017/5/3/-werror-is-not-your-friend

parafin:
Do you want to disable -Werror for Debug build too?

I don't know how to fix the warning, but you have to do it someway in any case, do you not? Explain to me the point of this PR after you do it.

aurelienpierre:
> Do you want to disable -Werror for Debug build too?

Not really. Debug can still be the crash test that fails if Jupiter is ill-aligned with Venus.

> I don't know how to fix the warning, but you have to do it someway in any case, do you not? Explain to me the point of this PR after you do it.

Preventing the next case of Clang overdoing it. It's bound to happen, just wait for Clang 9 or something. It already happened with Fedora 30 bumping GCC to v9, 6 months ago : OpenMP parallel loops pragma that were previously assumed shared if not specified otherwise begun to raise warnings treated as errors. The good folks at Fedora didn't care to investigate and just disabled OpenMP for their package. They didn't care to mention it to us, and users started to complain that dt was suddently super slow and using only 20 % CPU. Until @cryptomilk found out.

parafin:
How that example helps your point? Without -Werror Fedora folks wouldn't even notice that there is a problem. With -Werror they had a chance to understand the issue and fix it correctly (which they didn't take, but that's their fault, not ours).

I know the arguments for not having -Werror for Release builds, but that's not how and why this PR was introduced here. darktable project has CI tests configured, which alerts developers about errors (and warnings) on different compilers/systems. If you think that specific configuration is no longer relevant (I assume that warning actually indicates at least a minor problem, e.g. optimization requested being disabled), then it should be removed from CI config. If you think that warnings shouldn't be treated as errors in Release builds, then CI builds have to use Debug builds. Instead you disabled that useful check altogether, and now developers will find out about the problems only after users begin to complain. If this is the "right direction", maybe we should have turned left instead...

parafin:
Actually GCC9 issue doesn't help either side of the argument, since those were errors, and so -Werror has nothing to do with it.

TurboGit:
> If this is the "right direction", maybe we should have turned left instead...

Maybe :) So let's turn left now, what you'll you propose to:
- fix CI builder
- keep Apple user's happy
- make clang compilation on MacOS fails when a warning is issued
- make clang compilation on Linux fails when a warning is issued

?

That's a system with too many constraints to me, and at least at the moment I don't have a good solution to propose.

parafin:
> > If this is the "right direction", maybe we should have turned left instead...
> 
> Maybe :) So let's turn left now, what you'll you propose to:
> 
> * fix CI builder

obviously yes, otherwise why have it at all

> * keep Apple user's happy

is there a problem on macOS?

> * make clang compilation on MacOS fails when a warning is issued

Ideally yes, but it will require some work. I propose to leave it as it is now (and was before this PR) for the time being.

> * make clang compilation on Linux fails when a warning is issued

Yes, at least for the version of clang we want to support (and include in CI) and at least for the build type CI is using.

> 
> ?
> 
> That's a system with too many constraints to me, and at least at the moment I don't have a good solution to propose.

darktable was clean of the warnings on Linux before tone equalizer was merged. I propose to keep it that way. This PR is a big step in the opposite direction. Making macOS build warning-free would be a step to the better on the other hand. Allowing warnings will without a doubt degrade code quality in darktable. And there has to be at least some system that enforces warning-free build to keep that code out of the repository. Keeping CI happy is IMHO the easiest solution for that problem.

TurboGit:
> darktable was clean of the warnings on Linux before tone equalizer was merged. I propose to keep it that way. This PR is a big step in the opposite direction. Making macOS build warning-free would be a step to the better on the other hand. Allowing warnings will without a doubt degrade code quality in darktable. And there has to be at least some system that enforces warning-free build to keep that code out of the repository. Keeping CI happy is IMHO the easiest solution for that problem.

I agree with all that. But I asked for a *solution* to have all this coming a reality. I don't have one.

parafin:
I assume we are talking about that warning issued by clang 7? If I understand correctly it means that compiler failed to vectorize the loop. Do we care about that or not? If we care, then I guess code has to be rewritten, so that clang 7 can vectorize it. If we don't, then I see 2 options: 1). wrap that pragma in #ifdef that checks for clang 7; 2). pass -Wno-pass-failed option to clang 7. Maybe option 2 can be modified to have effect only local to toneequalizer using pragmas. Note that both these options should be implemented as specific to clang 7, otherwise nobody will notice when vectorization will break in future clang updates.

aurelienpierre:
On one hand, we have `-Wall` enabled. That means all the current and future warning that stupid compilers might find useful to throw will be thrown, and nobody here can anticipate what will be added.

On the other hand, we have `-Werror` enabled too. That means any warning will be blocking, even minor ones.

So:
* either we disable `-Wall` and choose which warnings are important to us, manually add them, then keep `Werrors`,
* or we keep `-Wall` and disable `-Werror`, then manually check build logs for warnings and sorting the major from the minor.

But right now:
* Clang and GCC are not consistent with each other, regarding flags and behaviour compiling same code, 
* Clang doc sucks and all the info you get is mailing-lists archives from 2-3 versions ago or pre-versions drafts never updated,
* Clang tries to support GCC pragmas that doesn't concern him, yet screams when he doesn't recognize them (but still sets `__GNUC__` so forget about compiler checks in macros),
* one issue warnings while the other doesn't care, 
* overall, giving them a wildcard, we take the chance to be fucked every time an overzealous compiler will add some new smart-ass warnings in a new release.

Yes, darktable codebase was warning-free before tone equalizer, due to SSE2 code path redundancy and general under-optimization (probably thanks to this impossibility to make both compilers happy at the same time). I have tried to avoid this by using simple C code and OpenMP pragmas as auto-vectorization hints, and made optimizations so various SIMD targets could be cloned (but only for GCC). But it seems the same code can be vectorized by GCC and not by Clang, and both don't behave that way.

The only way to avoid that is just to drop OpenMP SIMD support and fall back to stupid C, or spend another 6 months duplicating SSE code I already have to rewrite for OpenCL. All that for a goddam loop that missed auto-vectorization. 

I get that stack use warnings or unused variables or even hints on restricted pointers aliasing are important, but the thing is they don't even get triggered consistently and having developers lazily wait for the compiler to always spot them is just as bad making them disappear. 

As a side note, I couldn't build the Debug version until 3 weeks ago or so, because it triggered a stack use warning on the colour mapping module introspection file. Release and RelWithDebInfo built fine.

parafin:
If you don't care about auto-vectorization not working on Clang I have already suggested what flag you need to pass to compiler for it to shut up about that. No need to do such global changes just because you are frustrated about one specific issue. I disagree with you about usefulness of compiler warnings and with both your suggestions. Manually checking build logs - until yesterday you and @TurboGit didn't even know CI existed in darktable, so I wouldn't hold my breath about that option. Enabling only warnings we want - I think blacklisting is much better strategy then whitelisting here - new versions of compilers will introduce new warnings types, that might expose issues in our code, that we should fix. Generally false-positives are less frequent than real warnings IMHO. And anyway - wouldn't you want to enable warning about loop not being vectorized? Otherwise how would you know your code actually does what you intended?

aurelienpierre:
> If I understand correctly it means that compiler failed to vectorize the loop. Do we care about that or not? 

We don't care. In GCC, I use a `pragma GCC optimize` that performs several loops reorderings using graphs of variables dependencies and cost models heuristics to determine whether vectorization is worth it or not, depending on the data locality and cache use. So, the idea is to flag the array as vectorizable (because it is 64-bits aligned) and let the compiler decide if the overhead is worth the savings on current architecture and available vector sizes. Such pragma has no equivalent in Clang 7, and I don't know for the newer ones. The bottom line is if compiler decides not to vectorize, it might be for the best.

For the record, the faulty loop was a stupid copy of a grey image into an RGB image (where R = G = B = grey for each pixel). The epitoma of stupid simple…

> I disagree with you about usefulness of compiler warnings and with both your suggestions.

So you rely on compilers for a language that is not even capable of checking its own bugs before actually running the program ? Everytime a new compiler version is out, you will update your (working) code to comply with the new disease it diagnoses you ?

This is too random for me. Choose between `-Wall` or `-Werrors`, but both is the path to nightmare.

parafin:
Yes, you may need to update your code for a new compiler. Code gets broken sometimes by a new compiler, because it's rarely written entirely correctly. New optimisations introduced in newer compiler versions might expose such issues and warnings help deal with this too. Also warnings might tell you about existing bugs you just weren't aware of. If you don't want to deal with this - just don't upgrade the compiler.

Also I think you are misrepresenting the issue of warnings by making it seem much bigger than it is. This whole discussion is because of one warning, which actually gave you useful information, and for which I already proposed a possible workaround.

Anyway, I rest my case. If -Werror gets disabled in darktable, at least it should be spelled out what it means, instead of hiding such big change as a simple fix for one warning.

PR #2979, opened 2019-09-23
Re-introduce the new iop order as v3.

TurboGit:
This is properly using a migration path from v2 to v3. Note that only
new edits will use this new ordering. Old ones are kept on the original
iop_order to not break edits.

### DIFF
--- a/src/common/iop_order.c
+++ b/src/common/iop_order.c
@@ -31,7 +31,7 @@
 #include <stdlib.h>
 #include <string.h>
 
-#define DT_IOP_ORDER_VERSION 2
+#define DT_IOP_ORDER_VERSION 3
 
 static void _ioppr_insert_iop_after(GList **_iop_order_list, GList *history_list, const char *op_new, const char *op_previous, const int dont_move);
 static void _ioppr_insert_iop_before(GList **_iop_order_list, GList *history_list, const char *op_new, const char *op_next, const int dont_move);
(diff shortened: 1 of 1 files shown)


jenshannoschwalm:
I just saw this pr and gave it at test for an hour or so (todays master + #2979 + #2970).
Overall this is working absolutely fine right now.
1. having an older iop_order is respected and kept in xmp after more developing.
2. discard history is a proper method to start work from base.
3. No visible problems (as have been described in several issues) after 2. Also the history gives values for iop_order that seem to be ok.

Two more thoughts
- as the new iop_order should give better results as described earlier, and there were problems with the first approach -- what would be a situation where the iop_order could be **re-assigned** to make use of the improved processing? After a complete history compression?
- iop_order is a double. So if you divide by two for every module inserted in the same group, how long does it take to have no different medium? (56?)

TurboGit:
@jenshannoschwalm : thanks for testing.

For the first point, you mean that you would want a way to migrate a developed image to the new order without loosing history? There is no provision for that at the moment, but we could probably add such feature later.

For the second point, there is little chance that we would add 56 modules between two that are only distant of 1.0. Initial modules are ranging from 1.0 to 71.0. Lot of room.

TurboGit:
I have now merged this with some little adjustments as some modules where missing in the description. Better having this more field testing.

leclercj:
I've 2 questions:
1) This PR is fixing the iop_order at the start of any new edit. This is very good, but would it be better to also add more rules to forbid the user later during the edit to apply a non-supported or non-meaningfull order ? For instance, what about adding a rule to enforce orientation(flip) before crop&rotate(clipping), ashift before orientation(flip), ... ?
2) About iop_order as double: I see iop_order stored with 6 digits only in the .xml files. this is giving much less room I'm afraid. May be less than 20 module addition or move ? What about re-numering the iop order at some point to get rid of this risk?

TurboGit:
For 1, indeed the order is correctly set but can be changed by users in a destructive way. At the moment this is no provision to enforce this.

For 2, you may be correct. But we are not talking about adding modules between 0 .. 1. The initials values are ranging from 1 .. 70. So that gives a bit more room than 20. We have a max of 60 modules in dt and adding 20 modules between two having an iop_order diff close to 1 is low probability. But yet I agree with you, this is a problem we will certainly run into at some point.

jenshannoschwalm:
For 2. You may be right but i still think you underestimate the problem.
For some reason i don't understand some iop_orders want to be very close to 38.9xx Here is a look on the database just after some 'random' adding modules.

65387037-19fbf200-dd43-11e9-8360-7bb143263a18

**Look fur num 14 and 18**, they already have the same first 6 digits.

There is enought room to divide a a sum of two vals by 2 in double float format so iop_order will still be valid in the database. But the sidecars will be wrong pretty soon.

PR #2980, opened 2019-09-23
toneequal.c : remove the calls to _init_drawing() in gui_init() and gui_update()

aurelienpierre:
_init_drawing() uses the GtkAllocation struct to get the width and height of the drawing area widget.
Triggering it in gui_init() or gui_update() may give  a NULL GtkAllocation pointer, aka widget size might not be allocated yet in some cases,
and this makes g->graph_width segfault in init_nodes_x() when advanced tab is clicked early.


TurboGit:
Looks good ! Thanks.

PR #2991, opened 2019-09-24
make ovoid and ivoid firstprivate in parallel loop

aurelienpierre:
having them shared with *const restrict pointer makes building fail for some users


TurboGit:
Thanks.

PR #2994, opened 2019-09-24
correct place of filmicrgb in iop layout

bastnic:
Now that #2938 is merged, configure it in `tools/iop-layout.sh` script to be at the same place at original filmic.


TurboGit:
Just perfect! I had forgotten about that. Thanks.

PR #2999, opened 2019-09-25
toneequal: attempt to fix crashes.

### LINKED ISSUE #2978: toneequal exception
[darktable_bt_332L8Z.txt](https://github.com/darktable-org/darktable/files/3642709/darktable_bt_332L8Z.txt)

I don't know if that's related but it was just after ctrl-Z when I tried to move from simple to advanced tab.
After restart it fails again on the same move. 
The issue has gone when I reset the module.

Another point. toneequal is not retained in favorites from one session to the other.

 - OS: windows
 - Version 2.7

TurboGit:
May fix #2978.

### DIFF
--- a/src/iop/toneequal.c
+++ b/src/iop/toneequal.c
@@ -260,8 +260,8 @@ typedef struct dt_iop_toneequalizer_gui_data_t
   GtkStyleContext *context;
 
   // Event for equalizer drawing
-  float nodes_x[CHANNELS] DT_ALIGNED_ARRAY;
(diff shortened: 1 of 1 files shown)


TurboGit:
Manually merged with some more comments on the changelog.

aurelienpierre:
it's weird that the alignment could cause it.

TurboGit:
My thinking is that it is tagged as aligned but may be not. The allocation could start at whatever position and I don't think the nodes_x/y array are moved inside the struct to ensure alignment. Tested a lot, no more crashes.

PR #3001, opened 2019-09-25
ensure filmic rgb is inserted right after filmic v2

aurelienpierre:
I discovered that in iop_order.c, you need to declare the constraints on modules relative position in the right order, otherwise they get mixed in a random way. So new modules should be inserted at the end of the program to get expected insertion.


TurboGit:
Looks good! Thanks.

PR #3002, opened 2019-09-25
allow searching in deprecated modules that are enabled

### LINKED ISSUE #2995: Module search does not show deprecated but used module
**Describe the bug**
Filmic v3 just landed in master and so filmic v2 is deprecated. But in old edit we still have filmic enabled. We can see it "active modules" tab and in the correct group depending on your iop layout config. 
BUT it never shows up if you search for it. Only filmicrgb will be there, but never filmic.

Not sure it's a bug or a feature but it seems like a bug to me.

see https://github.com/darktable-org/darktable/pull/2938#issuecomment-534748307

**To Reproduce**
1. old edit with filmic v2 enabled
2. see that it's present in active module
3. search "filmic"
4. only "filmicrb" shows up

**Expected behavior**
Search should not show deprecated module if disabled. But if enabled, it should show it.

 - OS: not relevant
 - Version: master

aurelienpierre:
fixes #2995

### DIFF
--- a/src/libs/modulegroups.c
+++ b/src/libs/modulegroups.c
@@ -363,8 +363,8 @@ static void _lib_modulegroups_update_iop_visibility(dt_lib_module_t *self)
       // if there's some search text show matching modules only
       if(text_entered && text_entered[0] != '\0')
       {
-        /* don't show deprecated ones */
-        if(module->flags() & IOP_FLAGS_DEPRECATED)
+        /* don't show deprecated ones unless they are enabled */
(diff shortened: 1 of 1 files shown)


bastnic:
yeah @aurelienpierre, thanks. Confirming that it fixes my scenario, aka that filmic shows up on an old edit where it's enabled but never shows up on a new one.

ping @chhil

TurboGit:
Every details count ! Thanks.

PR #3007, opened 2019-09-25
Lighttable

### LINKED ISSUE #2998: Stars / rating should not be resized (Negatives)
**Stars growing too big in the negatives display**
I think, the stars and the rating - sitting underneath each photo - are way too big. Especially when I change the size of the negatives to bigger/less photos, the stars are enlarging too. 
I am finding this not very helpful, as these huge stars are distracting from the photo itself.
I am trying to explain with the following screenshots I made:

_The size of the stars is okay when looking at a lot of photos:_

Bildschirmfoto 2019-09-25 um 10 38 03

_Even by zooming in just a bit, the stars are growing:_

Bildschirmfoto 2019-09-25 um 10 38 28

_With enlarging the photos even more, the stars are enlarged way too much:_

Bildschirmfoto 2019-09-25 um 10 38 47

**Make the stars and colour rating part of the UI**
I would love to see a consistent size, no matter what zoom factor in the negatives view.
I don`t know, how to achieve this – I was messing with the css-files (in the themes folder) without finding a solution.  

**A look at RawTherapee**
Just to explain a bit more the idea, see following screenshots taken from RawTherapee.
(Don't get me wrong, I like the Darktable view better. I just think, RawTherapee handles the zoom better..)

Bildschirmfoto 2019-09-25 um 11 02 56

_When resizing the images, stars are not being resized and still looking neat:_

Bildschirmfoto 2019-09-25 um 11 03 13

**This is no bug**
Of course – but I think, this would be an overall improvement of the look and UI of Darktable. Unfortunately, I am no developer and I have no idea where to change this.  
I tried to change the css-files in the themes. But that does not do the trick...

aurelienpierre:
1. Refactor the lighttable to the general theme by removing rounded corners, dropped shadows and unnecessary borders.
1. Fix #2998 by making 20px × DPI the max size for text and icons in lighttable thumbnails, no matter the zoom size (for small zoom sizes, icons and text are still proportional : we just ensure they don't get oversized at large zoom levels).
1. Add a dark overlay background on thumbnail to help legibility of overlays stars and stuff.

All this is to focus on pictures and avoid distractions around them.

this PR:
Screenshot_20190926_011326

darktable master:
Screenshot_20190926_012314

this PR:
Screenshot_20190926_012500

darktable master:
Screenshot_20190926_012012

@AlicVB could you take over for the culling view ? I'm not sure what happens here with the `zoom == 1`, exif info display does not behave in culling like file manager.

### DIFF
--- a/src/dtgtk/paint.c
+++ b/src/dtgtk/paint.c
@@ -1204,26 +1204,26 @@ void dtgtk_cairo_paint_label(cairo_t *cr, gint x, gint y, gint w, gint h, gint f
   switch((flags & 7))
   {
     case 0:
-      cairo_set_source_rgba(cr, 1, 0.0, 0.0, alpha);
+      cairo_set_source_rgba(cr, 0.9, 0.0, 0.0, alpha);
       break; // red
     case 1:
-      cairo_set_source_rgba(cr, 1, 1.0, 0.0, alpha);
+      cairo_set_source_rgba(cr, 0.9, 0.9, 0.0, alpha);
       break; // yellow
     case 2:
-      cairo_set_source_rgba(cr, 0.0, 1, 0.0, alpha);
+      cairo_set_source_rgba(cr, 0.0, 0.9, 0.0, alpha);
       break; // green
     case 3:
-      cairo_set_source_rgba(cr, 0.0, 0.0, 1, alpha);
+      cairo_set_source_rgba(cr, 0.0, 0.0, 0.9, alpha);
       break; // blue
     case 4:
-      cairo_set_source_rgba(cr, 1, 0.0, 1.0, alpha);
-      break; // purple
+      cairo_set_source_rgba(cr, 0.9, 0.0, 0.9, alpha);
+      break; // magenta
     case 7:
       // don't fill
       cairo_set_source_rgba(cr, 0, 0, 0, 0);
       break;
     default:
-      cairo_set_source_rgba(cr, 1, 1, 1, alpha);
+      cairo_set_source_rgba(cr, 0.75, 0.75, 0.75, alpha);
       def = TRUE;
       break; // gray
   }
--- a/src/views/lighttable.c
+++ b/src/views/lighttable.c
@@ -1418,6 +1418,8 @@ escape_image_loop:
     if(sqlite3_step(lib->statements.is_grouped) != SQLITE_ROW) mouse_over_group = -1;
   }
 
+  cairo_set_line_cap(cr, CAIRO_LINE_CAP_ROUND);
+
   for(int row = 0; row < max_rows; row++)
   {
     for(int col = 0; col < max_cols; col++)
--- a/src/views/view.c
+++ b/src/views/view.c
@@ -798,17 +798,13 @@ int dt_view_process_image_over(dt_view_image_over_t what, int active, cairo_t *c
   int ret = 0; // return value
 
   float fscale = DT_PIXEL_APPLY_DPI(fminf(width, height));
-  float r1, r2;
-  if(zoom != 1)
-  {
-    r1 = 0.05 * width;
-    r2 = 0.022 * width;
-  }
-  else
-  {
-    r1 = 0.015 * fscale;
-    r2 = 0.007 * fscale;
-  }
+
+  // we need to squeeze 5 stars + 2 symbols on a thumbnail width
+  // each of them having a width of 2 * r1 and spaced by r1
+  // that's 14 * r1 of content + 6 * r1 of spacing
+  // inner margins are 0.045 * width
+  float r1 = fminf(DT_PIXEL_APPLY_DPI(20.0f) / 2.0f, 0.91 * width / 20.0f);
+  float r2 = r1 / 2.5f;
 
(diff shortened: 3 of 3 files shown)


TurboGit:
Looks better. I'll just fix the gradient that is ALWAYS on when overlay is activated. We need this only when over the image. Thanks a lot for this, just in time for dt 3.0.

aurelienpierre:
> We need this only when over the image.

This is debatable. The problem now is overlays are mostly unreadable on top of bright pictures with the grey theme, and on top of dark pictures with the default dark theme. So the idea of this darkening background is to give more predictibility/reliability to that.

aurelienpierre:
Side note for later, the colour in default themes for thumbnails overlays should probably be brightened and synced with the base text colour.

TurboGit:
> This is debatable. The problem now is overlays are mostly unreadable on top of bright pictures with the grey theme, and on top of dark pictures with the default dark theme. So the idea of this darkening background is to give more predictibility/reliability to that.

Well having the gradient on all images on the lt was just awful ::) On you screenshot above you don't have that BTW because you don't have the overlay always on. We may need to refine this, we'll see.

aurelienpierre:
> Well having the gradient on all images on the lt was just awful ::) On you screenshot above you don't have that BTW because you don't have the overlay always on. We may need to refine this, we'll see.

Oh I agree it's ugly, that's why I don't enable overlays ^^

AlicVB:
@aurelienpierre : zoom == 1 is mainly used for full preview and culling. I can take it over, but I'll need some guidance to know what to do (I'm very bad in UX design)...
Do we want to reduce font size ? same as for thumbnails ?
currently the font is in light grey, surrounded with dark frame : Capture d’écran_2019-09-26_18-13-29
Do we want to keep that ? (this is especially useful to reduce visibility pb on dark/bright images)

AlicVB:
... and my 2 cents about this : 
- the plain grey rectangle when you enable "extended overlay" is ugly and prevent to see the image... is it on purpose ? (I don't use them, but...)
- To solve the visibility on dark/bright images, what about using a "border" in different grey value like in the capture of my previous msg ?

AxelG-DE:
@AlicVB 
in culling mode I observed, I get a white (or light grey) border on mouse hover, except the top-left one, that stays without this border and even the info like stars etc. comes with a delay, means, after hover, I need to wait a sec and move the mouse a bit, then I can see (top right, I was playing around with filmicrgb :smile: )....
Screenshot_20190926_182938

AlicVB:
@AxelG-DE confirmed for the border, and this has been introduce by this PR. I'll have a look.
For the other point, I don't reproduce. It'll need more investigations. Can you open a new issue, in order not to pollute this thread more that what I've done ? Thanks

AxelG-DE:
Ooops, donnon what happened, when I started dt+1874 everything was like @aurelienpierre showed above, but all over sudden the darkening overlay is gone and thumbnail text only on hover (in between I merely switched on and off extended thumb overlay to try...) Do I need to open an issue for this?

Screenshot_20190926_184531

aurelienpierre:
>  zoom == 1 is mainly used for full preview and culling. I can take it over, but I'll need some guidance to know what to do (I'm very bad in UX design)...
Do we want to reduce font size ? same as for thumbnails ?

for zoom == 1, I think 20 px × DPI is good as font size. The thing is the overlay at zoom == 1 doesn't look the same as in culling view (icon spacing is changed), even if it seems culling view is setting it to 1 too. 

> currently the font is in light grey, surrounded with dark frame : 
Do we want to keep that ? (this is especially useful to reduce visibility pb on dark/bright images)

it doesn't help that much readability. Better use a plain background. 

> the plain grey rectangle when you enable "extended overlay" is ugly and prevent to see the image... is it on purpose ? (I don't use them, but...)

that hasn't change here.

> @AxelG-DE confirmed for the border, and this has been introduce by this PR. I'll have a look.
For the other point, I don't reproduce. It'll need more investigations. Can you open a new issue, in order not to pollute this thread more that what I've done ? Thanks

The selected image (top left) is supposed to get a 4 px × DPI black frame, and an additionnal 2 px × DPI white frame when hovering. Indeed, the white background doesn't work anymore.

AlicVB:
First, I quite like all your changes (need to be said)
> The thing is the overlay at zoom == 1 doesn't look the same as in culling view (icon spacing is changed), even if it seems culling view is setting it to 1 too.

Ok, now I see what you mean. I'll take care


> Better use a plain background.

I don't think it's a good idea, it will hide part of the image each time you move the mouse on it. I will try to come with something else, or keep the current behavior (text difficult to read is better than hidding part of the image, imho)


> that hasn't change here.

You're right, ugly, but as before, let's keep this apart, then.


> The selected image (top left) is supposed to get a 4 px × DPI black frame, and an additionnal 2 px × DPI white frame when hovering

Are you sure for the black frame ? I don't see it, only the white one (and it's great like that)



> Indeed, the white background doesn't work anymore.

I've a fix for that.

Anyway, I'm doing some work on all that points in the WIP PR #3023 don't hesitate to test it

PR #3013, opened 2019-09-26
deprecate some modules

aurelienpierre:
1. rename "equalizer" into "contrast equalizer" to make the difference between tone equalizer clear
1. deprecate the `relight` IOP:
  its use case falls back to the "fill-in" preset in tone equalizer, and uses the same logic to blend the exposure change. Unfortunately, it works in Lab display-referred space, where the concept of exposure is void, and the exposure blending ends-up flattening the affected areas a lot. Results are really random.
1. deprecate the `zone system` IOP:
  its use case falls back to the tone equalizer module, but tone equal allows to affect every channel independantly, while zone system redistributes the zones (you can only push a zone by pulling another), so the tuning can be more challenging and, again, local contrast flattening is to be expected.
1. deprecate the `global tone mapping` IOP:
  the colour science behind this module is dead wrong from the fundamentals. First of all, the tone mapping operators used there were designed for RGB spaces, yet they are applied in Lab. Second, the Lab space is designed around the idea that 100 % L means 100 Cd/m² (nits) display-referred, so it is not suitable for any HDR work (Lab-HDR has been developed a couple of years ago by Mark Fairchild and al.). Third, the Drago operator works fairly (even if it heavily desaturates shadows), the filmic operator falls back to the specific filmic module use case, but the Reinhard operator is completely broken and its output depends on the image size (different results in lightness have been reported depending on the file export resolution).

Modules left available that should be deprecated at some point:

1. tone mapping (the local one): its algo, using a gaussian blur to tonemap locally, will always produce halos around hard edges and there is no solution to that (using mild settings to hide dust under carpets is not considered a solution). This algo will never be robust, its flaw is on its theoritical background.
1. shadows and highlights: again, it's very prone to halos and the practically usable setting range is quite small. This module also adds local contrast as a side effect, for some settings combinations (that people might find pleasing), yet it's not its purpose and overlaps on local contrast module use cases. Finally, this module, working in Lab, is not really suitable for HDR work.

On a side note, here, HDR means any image which dynamic range is larger than the one of the output medium. Modern screens still have no more than 8 to 10 EV of theoritical dynamic range, and prints range between 4-7 EV. Yet, even entry-level DSLR shooting at base ISO hit 12 EV easily, and high-end ones almost hit 14 EV. So, pretty much every picture shot at base ISO now should be considered HDR, and a good image processing pipeline should be ready for that (aka allow dramatic edits without screwing colour).

### DIFF
--- a/src/common/fast_guided_filter.h
+++ b/src/common/fast_guided_filter.h
@@ -398,7 +398,7 @@ dt_omp_firstprivate(image, out, num_elem, sampling, clip_min, clip_max) \
 schedule(static) aligned(image, out:64)
 #endif
     for(size_t k = 0; k < num_elem; k++)
-      out[k] = fast_clamp(image[k], clip_min, clip_max);
+      out[k] = image[k];
   }
   else if(sampling == 1.0f)
   {
@@ -477,12 +477,6 @@ static inline void fast_surface_blur(float *const restrict image,
     {
       // Process the intermediate filtered image
       apply_linear_blending(ds_image, ds_ab, num_elem_ds);
-    }
-    else
-    {
-      // Increase the radius for the next iteration
-      ds_radius *= 2.0f;
-      feathering /= 2.0f;
     }
   }
 
--- a/src/iop/toneequal.c
+++ b/src/iop/toneequal.c
@@ -419,38 +419,38 @@ void init_presets(dt_iop_module_so_t *self)
   p.details = DT_TONEEQ_GUIDED;
   p.method = DT_TONEEQ_NORM_2;
 
-  p.blending = 12.5f;
-  p.feathering = 5.0f;
-  p.iterations = 3;
-  p.quantization = 1.0f;
-  p.exposure_boost = -1.0f;
-  p.contrast_boost = 2.0f;
+  p.blending = 33.0f;
+  p.feathering = 10.0f;
+  p.iterations = 1;
+  p.quantization = 0.0f;
+  p.exposure_boost = 0.0f;
+  p.contrast_boost = 0.0f;
   dt_gui_presets_add_generic(_("mask blending : landscapes"), self->op, self->version(), &p, sizeof(p), 1);
 
   p.blending = 25.0f;
-  p.feathering = 5.0f;
+  p.feathering = 25.0f;
   p.iterations = 2;
-  p.quantization = 1.0f;
-  p.exposure_boost = -1.5f;
-  p.contrast_boost = 3.0f;
+  p.quantization = 0.0f;
+  p.exposure_boost = 0.0f;
+  p.contrast_boost = 0.0f;
   dt_gui_presets_add_generic(_("mask blending : all purposes"), self->op, self->version(), &p, sizeof(p), 1);
 
   p.blending = 25.0f;
   p.feathering = 25.0f;
   p.iterations = 4;
-  p.quantization = 1.0f;
-  p.exposure_boost = -1.5f;
-  p.contrast_boost = 3.0f;
+  p.quantization = 0.0f;
+  p.exposure_boost = -0.5f;
+  p.contrast_boost = 1.0f;
   dt_gui_presets_add_generic(_("mask blending : isolated subjects"), self->op, self->version(), &p, sizeof(p), 1);
 
   // Shadows/highlights presets
 
   p.blending = 25.0f;
-  p.feathering = 10.0f;
+  p.feathering = 25.0f;
   p.iterations = 2;
-  p.quantization = 1.0f;
-  p.exposure_boost = -1.5f;
-  p.contrast_boost = 3.0f;
+  p.quantization = 0.0f;
+  p.exposure_boost = -0.5f;
+  p.contrast_boost = 1.0f;
 
   p.noise = 0.05f;
   p.ultra_deep_blacks = 0.15f;
@@ -464,12 +464,12 @@ void init_presets(dt_iop_module_so_t *self)
 
   dt_gui_presets_add_generic(_("compress shadows/highlights : soft"), self->op, self->version(), &p, sizeof(p), 1);
 
-  p.blending = 12.5f;
-  p.feathering = 20.0f;
-  p.iterations = 3;
-  p.quantization = 1.0f;
-  p.exposure_boost = -1.0f;
-  p.contrast_boost = 2.0f;
+  p.blending = 25.0f;
+  p.feathering = 10.0f;
+  p.iterations = 2;
+  p.quantization = 0.0f;
+  p.exposure_boost = -0.5f;
+  p.contrast_boost = 1.0f;
 
   p.noise = 0.5f;
   p.ultra_deep_blacks = 0.9f;
@@ -484,11 +484,11 @@ void init_presets(dt_iop_module_so_t *self)
   dt_gui_presets_add_generic(_("compress shadows/highlights : strong"), self->op, self->version(), &p, sizeof(p), 1);
 
   p.blending = 25.0f;
-  p.feathering = 10.0f;
+  p.feathering = 25.0f;
   p.iterations = 2;
-  p.quantization = 1.0f;
-  p.exposure_boost = -1.5f;
-  p.contrast_boost = 3.0f;
+  p.quantization = 0.0f;
+  p.exposure_boost = 0.0f;
+  p.contrast_boost = 0.0f;
 
   p.noise = 0.0f;
   p.ultra_deep_blacks = 0.15f;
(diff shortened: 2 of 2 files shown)


rawfiner:
Maybe "denoise non local means" could be deprecated too. The core algorithm is the same as denoise profiled in non local means mode, but all recent fixes and improvements have been made only in denoiseprofiled module and not in "denoise non local means". Now that modules can be reordered, I think that "denoise non local means" becomes a bit useless.

aurelienpierre:
> Maybe "denoise non local means" could be deprecated too. The core algorithm is the same as denoise profiled in non local means mode, but all recent fixes and improvements have been made only in denoiseprofiled module and not in "denoise non local means". Now that modules can be reordered, I think that "denoise non local means" becomes a bit useless.

About that, it seems the denoise non-local means (working in Lab space) is specifically designed for astrophotography, which has specific poissonian noise-patterns. But I learned it by discussing face to face with @hanatos, so maybe it's time to put it in doc.

TurboGit:
I have an objection for "zone system" because I have used quite a lot and like it. Maybe this can be replaced by tone equalizer but probably too early. People will need time to move to a new workflow. So let's maybe re discus deprecating it for next release.

aurelienpierre:
> I have an objection for "zone system" because I have used quite a lot and like it. Maybe this can be replaced by tone equalizer but probably too early. People will need time to move to a new workflow. So let's maybe re discus deprecating it for next release.

Well, this PR doesn't need to be merged right now. I'm just trying to decrease the confusion amongst users facing several modules that do more or less the same thing, some with side effects they will discover at their expense.

TurboGit:
I think the renaming of **equalizer** to **contrast equalizer** should be done for the release. I'll cherry-pick this commit if ok for you.

BTW, Aurélien you may want to revisit the module presets to add new filmic and the tone equalizer.

PaoloAst:
From my humble point of view, I'd like to deprecated the exposure module in
favor of the basic adjustment one.
That module is quite flexible and contains all the adjustments of the
exposure iop.

Sent from my mobile

Il gio 26 set 2019, 13:42 Aurélien PIERRE <notifications@github.com> ha
scritto:

aurelienpierre:
> I think the renaming of equalizer to contrast equalizer should be done for the release. I'll cherry-pick this commit if ok for you.

No problem.

> BTW, Aurélien you may want to revisit the module presets to add new filmic and the tone equalizer.

Yep, that too is on my to-do list. I will have to take a break in October though ^^

aurelienpierre:
> From my humble point of view, I'd like to deprecated the exposure module in favor of the basic adjustment one. That module is quite flexible and contains all the adjustments of the exposure iop.

Yeah the problem with the basic adjustments module is it squeezes several image operations that should happen at different places in the pipe into a single chunk, and duplicates other features that happen at the right place. So it's a temporary patch to make basic users happy, but it shouldn't be promoted as a clean way of editing. The right thing to would have been to merge basic modules into the same UI, leaving the actual filters separated in the pipe. The basic adjustments module might be enough for photographers not doing demanding edits, but for those pushing the limits of the soft, it will backfire, and can't replace what it duplicates.

MStraeten:
> From my humble point of view, I'd like to deprecated the exposure module in favor of the basic adjustment one. That module is quite flexible and contains all the adjustments of the exposure iop.
> […](#)

the exposure is a very basic and flexible module - in combination with masks it can be used to do isolated exposure corrections. In my opinion it doesnt make sense to use modules with several functions if only one basic operation is needed.

jenshannoschwalm:
If modules are marked as deprecated, would we want a tool that searches the database and marks affected images?

TurboGit:
> If modules are marked as deprecated, would we want a tool that searches the database and marks affected images?

Everything is possible, but to me this is not really useful. A deprecated module will stay there forever to ensure old edits are not broken in any way. What would be more useful is a way to load deprecated module only when needed to avoid consuming memory. But that certainly a delicate work.

Or maybe you have another scenario for the mark?

jenshannoschwalm:
If the is a strong consensus about a module 'to be deprecated' this could be interpreted as: "You are using a module with inferior quality, maybe erconcider your work"

I thought of adding an automatic tag to the image like darktable|problem|filmic-v2 or alike. Don't know whether its really useful and worth the work.

parafin:
I don't think it makes sense. Why would anyone re-develop old pictures?

TurboGit:
To hint people to migrate to new modules? Not sure... At this stage I may have thousand of such pictures :) Also, this would have do be done at every start of dt, could take long time!

parafin:
They will get the hint when developing new pictures (or even cloning old ones).

TurboGit:
Ok, if it is an indication in the darkroom that the edited image is using a deprecated module then I agree that this could be nice. If we open an image in darkrroom it is probably that we want to revisit the edit, and then having this indication makes sense.

Maybe using a different color in the history for deprecated module?

And sending a log message to notify clearly that deprecated modules are being used?

And maybe a pref if people don't want to get bothered with the log message?

Just some rhoughts about this.

parafin:
In the darkroom of course, some indication should be present, but there's no need for database updates or special tags to achieve that.

jenshannoschwalm:
Maybe in the history display we could add some info like we do when the module is switched off?

jenshannoschwalm:
Something non-instrusive like this?
Bildschirmfoto von 2019-09-29 10-36-32

aurelienpierre:
I like that @jenshannoschwalm

TurboGit:
As we are at it, since this is some characters and that Gtk is supporting utf-8, why not use a special character **before** the IOP name? Something like:

♻ or ⚠

?

chhil:
I am trying to look at someones edit who provided a raw + xmp.

Look at filmic, it's got an unprintable indicator.

image

Using 2.7.0.1921 on win 10.

TurboGit:
Arg!!!! This is a Win10 issues, the character is expected to convey the fact that filmic is a deprecated module. Looks like Windows does not support properly some UTF-8 symbols.

TurboGit:
@jenshannoschwalm : maybe we want to fallback to /!\ 

This is three characters that looks like a warning flag; What do you think?

jenshannoschwalm:
before returning to good old ascii art
@chhil could you do me a favour and test this - (i have no access to a win system)

In src/libs/history.c you find
`g_snprintf(marker, sizeof(marker), "\u267B");`

This is unicode and might be wrong, could you replace this by
`    g_snprintf(marker, sizeof(marker), "\xe2\x99\xbb");`
and report?

parafin:
More likely it's about installed fonts and whether matching ones have this symbol, so this can happen on Linux too.

chhil:
I am building the windows version and will test shortly. I did do a quick test in java and the unicode char is shown. I think its about the charset that selected and that value may not be available in that codepage. 


image

chhil:
@jenshannoschwalm 
Rebuilt it with the one liner code change. Did not change anything. I did not expect it to. 
Replaced this
>g_snprintf(marker, sizeof(marker), "\u267B");
with
>g_snprintf(marker, sizeof(marker), "\xe2\x99\xbb");

Even ran a small test C code, it prints it correctly. Something in the DT build is causing it to not display.

image

jenshannoschwalm:
@parafin Probably the win gtk/cairo/pango needs some hint?

chhil:
@jenshannoschwalm 
Not sure if this helps, but could you could you compare the values you see for "marker" and label?
image

image

jenshannoschwalm:
As it seems to be really tricky to add fonts to the windows gtk3 system without changing the installer details shoud we revert to suggestion "/!\" as a string freeze exception :-)

parafin:
There's no string freeze yet.

junkyardsparkle:
Personally, I like the idea of using a simple ASCII indicator, but tbh I find the "slash triangle" slightly odd and visually confusing... and when parsed correctly maybe just a little bit too alarming? What about just putting the exclamation in square brackets, as a somewhat more familiar convention? Just a thought.

TurboGit:
@junkyardsparkle : maybe a bit more visible... hard to find something that will please everyone anyway. as we are not in string freeze yet we can still play with some alternatives.

junkyardsparkle:
> as we are not in string freeze yet we can still play with some alternatives.

It seems worth thinking about only because in many cases it's likely to be persistent in the history stack (realistically, most people won't be re-doing all their previous work on older photos). Given that, it would make sense to have a big, one-time alert explaining the "deprecated module" concept, what it does and doesn't mean to the user, and then make the history stack indicator a relatively unobtrusive reminder... unless I'm thinking about it wrong. :-)

Issue #2964, opened 2019-09-20
New pictures in chapter 2.3.12 Tagging are pale-coloured


mepi0011:
The images inserted in chapter 2.3.12 are pale compared to the other images in the manual.
For example see picture [panel-tagging-default.png](https://github.com/darktable-org/darktable/blob/master/doc/usermanual/lighttable/images/panel_tagging-default.png) or attached picture below.

**The screenshots should be adapted to the images in the manual.**

![2019-09-20_darktable-usermanual-de](https://user-images.githubusercontent.com/9118135/65313498-bc29a780-db94-11e9-8e40-a4a60e02859e.png)

TurboGit:
Ah indeed! Now that we have multiple theme we must be sure the dark one is used for all screenshot.

phweyland:
Then the theme to be used for doc is "darktable", correct ?

TurboGit:
No, I would go with darktable-elegant which is cleaner. And this will certainly be what Aurélien will use for its documentation part.

phweyland:
OK, I'll do it.

mepi0011:
I would like to return to the subject.
By default, darktable uses the design darktable, which corresponds to the brighter images.
Therefore, the images in the user manual should also correspond to the brighter design. 
I would like to replace a few pictures in the user manual with newer ones.
So that I don't do the work twice, the design to be used should be clarified.
Currently the pictures in the user manual are still mixed!

Nilvus:
@TurboGit: I think this could be closed as if I'm not wrong, #3015 PR by @phweyland solved this one.

TurboGit:
I think so indeed. Thanks.

PR #3015, opened 2019-09-26
moved doc tagging pictures from darktable-elegant-grey to darktable.css

phweyland:
should answers the issue #2964 .
However I'm not happy because the font seems to be a lot smaller.
Since the grey snapshots I've moved to another computer. The unique difference is that my display is now 1900x1200 instead of 1680x1080 or 1920x1080, I don't remember. But that should not influence the pixels, right ?
I've checked that roboto fonts are installed:

image

Any advice ?


TurboGit:
Hum, indeed the text seems quite small. Don't know what the issue but this does not look correct to me. What about using the default darktable thème? Is it better?

phweyland:
Using darktable theme is better. I've tried to understand why darktable-elegant gives such small font.
Without success so far.

With a simple theme like:

```
@import url("darktable.css");

treeview,
treeview*
{
    font-family: "Segoe UI Light", sans-serif;
}
```

I still get the small font in the treeview. "Segoe UI" (regular) gives a clearer font but with same size. 
Interesting point is that the line height does not change. Only the font size ...

Removing Segoe UI Light, I get the same size as darktable theme. 
```
@import url("darktable.css");

treeview,
treeview*
{
    font-family: sans-serif;
}
```

phweyland:
Same size but different weight (than elegant).

phweyland:
Still not found why I have so much size difference between sans-serif and Segoe UI (first commit).
I'm pretty sure I had not this issue on the former computer.

aurelienpierre:
> Still not found why I have so much size difference between sans-serif and Segoe UI (first commit).
I'm pretty sure I had not this issue on the former computer.

This is all about font properties. Some fonts are wider, darker, etc. than other, but also they might not have all the subsets. The elegant theme use a light and narrow version of the font (more suitable for high DPI displays), but if it's not installed, it falls-back to the closest `font-weight` and `font-variant` available, which might look bigger at the same `font-size`, depending how the font is actually drawn.

phweyland:
> This is all about font properties.

That's OK but how to explain such a size difference between the module name and normal texts ?

Darktable.css (10-15%)
Annotation 2019-10-02 095654 PNG

Dartable-elegant-darker (20-25%)
Annotation 2019-10-02 095655 PNG

phweyland:
GTKInspector states 14,4px for module name and 12px for other texts, or 17% smaller (CSS decoding).

I have not found any place with GTKInspector, to see the actual font face, actual font-size, ... applied. Any tip ?

chhil:
I believe I am seeing something similar.

https://github.com/darktable-org/darktable/issues/3057#issue-501359168

Other than darktable.css rest are unusable.

TurboGit:
What I would propose to make progress is to use the default theme (darktable.css) for the new pictures as the other themes have different rendering depending on the OS.

How does that sound?

phweyland:
> What I would propose to make progress is to use the default theme (darktable.css) for the new pictures as the other themes have different rendering depending on the OS.


In fact, as elegant is designed, it works well for high resolution screens but may diverge for lower resolutions (not completely sure the OS makes a lot difference). And probably screenshots made with high resolution will be better whatever the used font face.
So yes, your proposal is fine for me. The last commit has done with darktable.css, if the quality is good enough.

Question: to get an acceptable image rendering I have to tweak the width percentage:
`<graphic fileref="lighttable/images/panel_tagging-default.png" scalefit="1" width="90%" align="center" />`
Is this percentage related to available space in page ? Is there a way to request to display the image at 100% of it size ?

Nilvus:
@phweyland and @TurboGit : this PR is stuck since 2 months and milestone is stil for 3.0. What about this ?

phweyland:
On my side I've made again the snapshots using normal darktable.css instead of the grey one.
Now I'm not completely happy with this due to the current machine screen I work on. If someone else want to provide better quality ones that's OK for me.

mepi0011:
I realize that the appearance of the window bar depends on the settings used by the operating system. But in the user manual window bars are always shown in a dark color. The white title bars are very prominent. Could you please change them, as an example see the attached image.

Screenshot_20200111_082811

phweyland:
@TurboGit . What do we do here ?
In term of quality I haven't better than this.
Some snapshot deserve an update because css have evolved. 
I can do it but my tags are in French.
What do you think ?

TurboGit:
@phweyland : I would say go for it. The GUI must be in English, but I understand that the tags are in the dev language. For me its ok, if someone object he can propose screenshots with English tags. So, let's go thanks in advance.

phweyland:
I've translated the visible tags :)
Normally I haven't changed any piece of text.

TurboGit:
Looks ok! Thanks for the hard work.

Issue #3010, opened 2019-09-26
Masks done before crop show up outside image after crop


chhil:
**Describe the bug**
Used brush masks in retouch module.
Then cropped the image,
Revisited retouch to for more brush masking.
Some of the brush masks were outside the crop as seen in the image.

**Screenshots**
![image](https://user-images.githubusercontent.com/1119439/65662065-f1307100-e050-11e9-994d-3b0bd9cbbf12.png)

**Platform (please complete the following information):**
 - OS: Win10
 - Version darktable-2.7.0+1862~g879f1e5f9-win64

TurboGit:
Cannot reproduce.

jenshannoschwalm:
Could you send the xmp file for that image please?

chhil:
@jenshannoschwalm XMP zip attached.
[P1137439.RW2.zip](https://github.com/darktable-org/darktable/files/3660679/P1137439.RW2.zip)

jenshannoschwalm:
Thanks for sharing.

Did you ever remove that image from darktables collection and reimport it before you tried to compress the history?

1. If so - the problem is pretty obvious. if you look at the iop_orders of used modules you will see same values "39" for various different modules. So when compressing modules will be removed in a way not easy to predict. Also you might see strange visual things because the order in the processing pipeline is not correct.

1. If not -  any way to share the data from main.history for that image?

I think that's another example for the new iop_order problems. 

#3016 is intended to fix this (you have to remove that file from the database and must **also** remove the xmp file)

chhil:
> Did you ever remove that image from darktables collection and reimport it before you tried to compress the history?

I honestly dont recall. I have had so many issues that I have been seeing and one of them is related to removing images from a collection  that were imported twice and I dont know if I did that on this folder.

> main.history

Not sure what you mean.
But I have attached the config folder and maybe you are looking for 
`select * from history where imgid = 1369`

I got the image id from the image info  from 
![image](https://user-images.githubusercontent.com/1119439/65745444-564f9980-e119-11e9-8268-0fd2f7aca5aa.png)


[darktable_config.zip](https://github.com/darktable-org/darktable/files/3661048/darktable_config.zip)


I will try #3106 but that would mean re editing it, if I understand what is being requested.

jenshannoschwalm:
Could you try again with current master after removing file from database and removing xmp file?

chhil:
[...]

Issue #3009, opened 2019-09-26
Compress History in darkroom view does weird things


chhil:
**Describe the bug**
Compress history in darkroom gives unpredictable results.

**To Reproduce**
See images.

**Expected behavior**
Compress should compress history to the point selected in history.

**Screenshots**
Original history
![image](https://user-images.githubusercontent.com/1119439/65660607-d5c36700-e04c-11e9-8838-d4e3822555cd.png)

Selecting Original
![image](https://user-images.githubusercontent.com/1119439/65660627-eb389100-e04c-11e9-9af5-358031dd8459.png)

Hitting compress
![image](https://user-images.githubusercontent.com/1119439/65660659-fd1a3400-e04c-11e9-9ac3-e93f49ed4c9b.png)

Selecting step 14 and then hitting compress
![image](https://user-images.githubusercontent.com/1119439/65660732-38b4fe00-e04d-11e9-8bd9-2dc5cc25ea15.png)

Now selecting 5 in history and hitting compress.
![image](https://user-images.githubusercontent.com/1119439/65660775-626e2500-e04d-11e9-85f9-4aee7c9bf0f8.png)

**Platform (please complete the following information):**
 - OS:Win 10
 - Version darktable-2.7.0+1862~g879f1e5f9-win64

Only way to reset is to discard history in lightable view.

TurboGit:
Really strange! I cannot reproduce.

jenshannoschwalm:
Might be related to #3016

jenshannoschwalm:
Could you try again with current master after removing file from database and removing xmp file?

chhil:
Couldnt reproduce on a cleandarktable-2.7.0+1898~ge37281ae4-win64.

Though an ideal test would be to upgrade from 2.6.2 to current.

jenshannoschwalm:
maybe you can close this issue for now. testing the 2.62 -> current master is ok, the problem was the iop_ordering and now that seems to be fixed.

chhil:
@jenshannoschwalm 

When you import the pic and the default history is present, you can't compress it to the original step 1 in history.
Compressing to any other point in the existing default history also shows.

Default history.
![image](https://user-images.githubusercontent.com/1119439/65843858-76699d80-e351-11e9-821b-d83e651b8ba7.png)

Selecting original
![image](https://user-images.githubusercontent.com/1119439/65843880-8bdec780-e351-11e9-93c7-99fbb980c9c0.png)

Clicking compress history
[...]

PR #3016, opened 2019-09-26
Fixing new iop_order for sidecar files.

jenshannoschwalm:
Discussing and debugging the new iop_order sheme showed there might be a problem
with the insertion of modules into the iop_order list.

Further tests here showed the new v3 algorithm is ok and gives correct iop_order
defined as a double.

But these tests also demonstrated, the exiv2 writing a float/double key to the sidecar uses low precision
1. resulting in iop_order of 38.9991 to be the same as 38.9992 **and** 
2. numbers close to a 'full' as 38.9999 to be in fact written as 39.0.

This means the priority of modules in sidecars is definitely wrong.
You **can** observe this by deleting the image from history and reimport it or just compare existing sidecar files with the data in main.history

This commit
- helps to debug and understand the problem, there is now
  #define DT_IOP_ORDER_INFO FALSE in iop_order.c
  If set to TRUE you will see what is inserted and what iop_order is given.
- Writes the iop_order **not** by toFloat but instead as a converted string with a precision of %.13f

Is this really enough? I don't really know. Possible problems:
- Already iop_order_v3 developed images deleted from the database will give wrong iop_orders
  when reimported.
- Do we also have to make sure exiv2 reads the high-precision iop_order as a double? Reading exiv2 docs about this did not give a definite answer for me and by experiment i would say it's ok as now.

### DIFF
--- a/src/common/exif.cc
+++ b/src/common/exif.cc
@@ -2115,7 +2115,7 @@ static GList *read_history_v2(Exiv2::XmpData &xmpData, const char *filename)
       }
       else if(g_str_has_prefix(key_iter, "darktable:iop_order"))
       {
-        current_entry->iop_order = history->value().toFloat();
+        current_entry->iop_order = history->value().toFloat();  // This is a problem ??
       }
       else if(g_str_has_prefix(key_iter, "darktable:blendop_version"))
       {
@@ -2441,7 +2441,7 @@ static int history_v1_to_v3(const int imgid)
       sqlite3_finalize(stmt);
       if(multi_priority_max >= 0)
       {
-        const float iop_order = ((float)(multi_priority_max + 1 - multi_priority) / 1000.0) + prior_v1->iop_order;
+        const double iop_order = ((float)(multi_priority_max + 1 - multi_priority) / 1000.0) + prior_v1->iop_order;
 
         DT_DEBUG_SQLITE3_PREPARE_V2(dt_database_get(darktable.db),
                                     "UPDATE main.history SET iop_order = ?2 WHERE imgid = ?1 AND num = ?3", -1,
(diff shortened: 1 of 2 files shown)


TurboGit:
I understand the issue but I'm not sure about the solution proposed here.

My first thinking on that was to re-compute the iop-order dynamically each time a new module is moved or added. In iop_order.c we do find the "right" place for a module in line 388. What about redistributing all iop-order in the current list ? This may also need to recomputing in the current dev history list? Not sure how easy this is or even if possible.

What do you think?

jenshannoschwalm:
It's definitely **not** a **final** solution. it just is a temporary fix for problems we have atm as sidecar files have _vastly_ wrong information. When importing such data many things may happen.

About line 388, that's exactly the point so i added the DT_IOP_ORDER_INFO section right there :-)

So what is the problem of the algorithm?

When we use the new iop_order distribution algorithm we do lots of `_ioppr_insert_iop_before` and as the rules define this all takes place __very closely__. As we always divide by 2 one could say we populate a binary very densely at a small area. If we use a double float to represent the repeated divide we probably can safely do 40 insertions all together but as we don't keep track **where** we insert a number in that tree we do **not** know how many more we could do.

Anyway -- representing the iop_order as a float as done before when writing sidecars is definitely wrong. It just didn't become obvious so far as we didn't do so many `_ioppr_insert_iop_before`

Problems and how to solve
1. My proposal leaves room for some insertions even at the densest parts so it's a ok for a while.
2. Your idea about recalculating. Very tricky as we also have to respect data from xmp files and data available in main.history
3. The many iop_order checks do help as long as we write high precision data to xmp
4. We could introduce a counter for current code increasing with every iop_order insertion to make sure we don't overflow the binary divide.

src/common/exif.cc
```diff
@@ -2918,7 +2918,14 @@ static void dt_set_xmp_dt_history(Exiv2::XmpData &xmpData, const int imgid, int
     snprintf(key, sizeof(key), "Xmp.darktable.history[%d]/darktable:multi_priority", num);
     xmpData[key] = multi_priority;
     snprintf(key, sizeof(key), "Xmp.darktable.history[%d]/darktable:iop_order", num);
-    xmpData[key] = iop_order;
+    // This code only writes iop_order as the default which is float, so it's wrong.
```
TurboGit: Ok, let's go with that but I would remove this comment which will be not clear when the code is removed. I would say:

// we ensure that we write the iop_order as high precision...

jenshannoschwalm:
@TurboGit , i won't be at my pc until tomorrow so feel free to modify that yourself. otherwise i'll do it tomorrow.

There are several issues around that might be related. Unfortunately this pr does **not** fix the problem if you took the history data from an xmp file that has been written since introducing the new iop_order.
#3010 and #3009 are probably related.
When we have problems that could be related to history/iop_order we should always advice to discard history and try again after that to ensure the information in main.history is same as in xmp file.

I'll keep an eye open. There might be more coming. In all such issues we have to ask people for the data in main.history **and** xmp to understand a possible problem. We have the sidecar writing tool, could we extend that to also write the data from the history?

src/common/exif.cc
jenshannoschwalm: Fixed.

jenshannoschwalm:
good now.

TurboGit:
Thanks ! Manually merged, the "..." at the end of my comment proposal was meant to be completed :)

jenshannoschwalm:
@TurboGit and @aurelienpierre 

I suggest a surprisingly simple way to distribute the iop_orders more evenly without reordering.

I'm not doing a pr yet, just would like to hear what you both think because you seem to be the ones involved in this stuff.

See my branch relaxediop; the numbers shown for iop_order are **much** better.

Do i oversee any side effects? Would we need a new version?

PR #3017, opened 2019-09-26
Add safeguard rules in iop_order to protect clipping iop

leclercj:
"clipping" iop GUI simply does not work when orientation ("flip") or "ashift" are moved after in pipe order.
This PR is enforcing the rules to forbid moving flip or ashift after clipping during edits.


TurboGit:
Nice ! I didn't know it was possible to enforce this. Thanks.

PR #3021, opened 2019-09-26
Change the char type to use CHAR_MIN/MAX for it's limits

RobbieAB:
char on x86(_64) defaults to a signed 8-bit int type. On ppc64le it defaults to an unsigned 8-bit int type.

G_MININT8 and G_MAXINT8 are the limit values for gint8, which is a signed type on all platforms.

### DIFF
--- a/tools/introspection/ast.pm
+++ b/tools/introspection/ast.pm
@@ -373,7 +373,7 @@ sub new
   bless($reference, $self);
 
   $reference->{unsigned} = 0;
-  @{$reference->{limits}} = ("G_MININT8", "G_MAXINT8", "0", "0", "G_MAXUINT8", "0");
+  @{$reference->{limits}} = ("CHAR_MIN", "CHAR_MAX", "0", "0", "UCHAR_MAX", "0");
   $reference->{code_type} = "char";
 
   return $reference;


houz:
We can not use a define which differs on some platforms. These are params, they need to be interpreted the same everywhere! It seems only the C++ compiler has a problem with that. So let's find out if/how we can beat it into submission.

hellbuster:
@houz But char itself is already being interpreted differently on different platforms (at least when you view them as an integer type), so unless you are using them to store integers, it shouldn't cause much problem or am I mistaken?

houz:
Are they also interpreted differently on supported platforms? Or just those hacked-on architectures like PPC? In any case, instead of a quick fix that has the potential to make things worse we need to investigate this.

shawnanastasio:
The C/C++ standards do not define a signedness for the `char` type, so assuming one way or the other is incorrect. If darktable relies on signed 8-bit values, then it should use `int8_t`, or even `signed char` instead.

houz:
I am well aware of the standard, I was asking about reality.

shawnanastasio:
Well, reality tends to follow the standard and the only way to write correct and portable code is to adhere to it as much as possible.

As far as other architectures that use unsigned char, I seem to recall that some ARM ABIs also do as well. Either way, just because something happens to work on x86 doesn't mean that it's correct or not worth fixing :smile:

houz:
So far SSE was a requirement for a platform to be supported, so only those matter. Now that dt starts to allow a wider set of systems we need to

1. evaluate if the current set of supported platforms was coherent
2. see how to change dt to keep the current set of platforms working the same way as before while fixing the situation for the newcomers.

So I totally agree that this needs to be fixed in dt, but before we change anything we should see what the status quo is. And for that we need to know what GCC and Clang do on the currently supported platforms. PPC point 2 in that list.

shawnanastasio:
Sure. On x86(_64) with GCC, Clang, and any ABI-compliant compiler, `char` will be a signed 8-bit value. For more information, see page 13 of the ABI document [here](https://refspecs.linuxfoundation.org/elf/x86_64-abi-0.99.pdf).

My suggestion of using `int8_t` or `signed char` in place of `char` where signedness is needed wouldn't change the compiler behavior on x86 at all, but would bring the code into compliance with the standard and fix ppc64le, arm, etc. as a result.

RobbieAB:
SSE was not a requirement when I started on this (or it would have killed my efforts straight away!), and my understanding was the ARMv8 port didn't use SSE. The change proposed here should not change anything on a platform with a signed char type, so if it DOES break anything, I am probably the only person who will be impacted!

If the decision is that maintaining ppc64le support would require too extensive changes to justify, I am happy to accept that.

houz:
If 8873a71847b18b8fc9929a7a497ead2d1532be7b fixed the issue we can close this PR.

houz:
@RobbieAB confirmed the issue to be solved. Closing here.

PR #3024, opened 2019-09-26
Update denoiseprofiled wavelets

rawfiner:
Just added 2 more scales to denoiseprofile's wavelets.
This allows to use wavelets for high ISO images, whereas before there were not enough scales to remove the coarse grain noise.

I also updated the presets, and chose to change blend modes from color and lightness to HSV color and HSV lightness, as PR 2829 was not updated for a long time.


TurboGit:
Looks good! Thanks.

PR #3033, opened 2019-09-28
French translation: Small typo

EdgarLux:
Using 'HEIGHT' instead of 'WIDTH'


TurboGit:
Thanks for sporting that!

PR #3035, opened 2019-09-28
Read xmp iop_order

jenshannoschwalm:
After making sure iop_order is written as a high precision double this is also done for reading.

### DIFF
--- a/src/common/exif.cc
+++ b/src/common/exif.cc
@@ -2115,7 +2115,12 @@ static GList *read_history_v2(Exiv2::XmpData &xmpData, const char *filename)
       }
       else if(g_str_has_prefix(key_iter, "darktable:iop_order"))
       {
-        current_entry->iop_order = history->value().toFloat();  // This is a problem ??
+        // we ensure reading the iop_order as a high precision float
+        string str = g_strdup(history->value().toString().c_str());
+
+        // don't want to modify the locale so we do a simple test and maybe replace a dot by a comma
+        if ( 0.5f > stod("0.6")) replace(str.begin(), str.end(), '.', ',');
+        current_entry->iop_order = std::strtod(str.c_str(), NULL);
       }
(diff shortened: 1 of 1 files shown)


src/common/exif.cc
```diff
@@ -2115,7 +2115,12 @@ static GList *read_history_v2(Exiv2::XmpData &xmpData, const char *filename)
       }
       else if(g_str_has_prefix(key_iter, "darktable:iop_order"))
       {
-        current_entry->iop_order = history->value().toFloat();  // This is a problem ??
+        // we ensure reading the iop_order as a high precision float
+        string str = g_strdup(history->value().toString().c_str());
+
+        // don't want to modify the locale so we do a simple test and maybe replace a dot by a comma
+        if ( 0.5f > stod("0.6")) replace(str.begin(), str.end(), '.', ',');
```
TurboGit: I'm not fond of that... But I have nothing better to propose :( So let's go.

parafin: Actually in C++ there is a way (and this is C++ module):
```
	static const std::locale& c_locale = std::locale("C");
	std::istringstream istring(str);
	istring.imbue(c_locale);
	istring >> current_entry->iop_order;
```

jenshannoschwalm: @parafin thanks for hinting. I confess i never touched C++ until i started reading dt code :-)

I roughly understand what going on in your code snippet. But how/where is interpretation ensured to be a double - i guess it converts as it can and writes to the target knowing it's type.

Worth another pr?

parafin: It knows the type, yes. @TurboGit should decide is it worth a PR.

TurboGit: You can just push a new commit on top of this one here. Thanks.

jenshannoschwalm: @TurboGit just committed parafins suggestion.

PR #3037, opened 2019-09-28
Relaxing stress on iop_order

jenshannoschwalm:
When inserting a new iop_order we always insert to the middle of two already existing numbers.
This populates the iop_numbers densely (in a very special way) and stresses the representation in double floats.

What could we do avoiding any sort of runtime reordering? Two ideas, both based on
the idea "**we** know better how  and where to insert than a simple divide-by-2"

1) before inserting we could always give a hint like jump to the middle (as we did before), jump to the far right, jump to slightly left ...

2) Or we could look at the code and find out what we do while populating:
We jump mostly to the right so we could not jump to middle-right but
only 'slightly' right. In maths that would mean, divide by a larger number
than two - i picked 8. If the analysis is right when inserting 20 iop_orders
we would not lose 6 decimal digits but a bit more than 2.

I have this working here proving the algorithm works and reduces stress in the order that was expected.

@TurboGit and @aurelienpierre as you seem to have worked on the new order most.

1. What did i miss?
2. Do we need a new version?

### DIFF
--- a/src/common/iop_order.c
+++ b/src/common/iop_order.c
@@ -110,12 +110,16 @@ static int _ioppr_legacy_iop_order_step(GList **_iop_order_list, GList *history_
     // move frequency filters right after input profile - convolutions need L2 spaces
     // to respect Parseval's theorem and avoid halos at edges
     // NB: again, frequency filter in Lab make no sense
+    // sharpen is marking the end of this section
+    _ioppr_move_iop_after(_iop_order_list, "sharpen", "defringe", dont_move);
     _ioppr_move_iop_after(_iop_order_list, "atrous", "defringe", dont_move);
     _ioppr_move_iop_after(_iop_order_list, "lowpass", "atrous", dont_move);
     _ioppr_move_iop_after(_iop_order_list, "highpass", "lowpass", dont_move);
-    _ioppr_move_iop_after(_iop_order_list, "sharpen", "highpass", dont_move);
+//    _ioppr_move_iop_after(_iop_order_list, "sharpen", "highpass", dont_move);
 
     // color adjustments in scene-linear space : move right after colorin
+    // bloom is marking the end of this section
+    _ioppr_move_iop_after(_iop_order_list, "bloom", "sharpen", dont_move);
     _ioppr_move_iop_after(_iop_order_list, "channelmixer", "sharpen", dont_move);
     _ioppr_move_iop_after(_iop_order_list, "colorchecker", "channelmixer", dont_move);
     _ioppr_move_iop_after(_iop_order_list, "colormapping", "colorchecker", dont_move);
@@ -126,28 +130,33 @@ static int _ioppr_legacy_iop_order_step(GList **_iop_order_list, GList *history_
     _ioppr_move_iop_after(_iop_order_list, "basicadj", "colorbalance", dont_move);
     _ioppr_move_iop_after(_iop_order_list, "rgbcurve", "basicadj", dont_move);
     _ioppr_move_iop_after(_iop_order_list, "rgblevels", "rgbcurve", dont_move);
-    _ioppr_move_iop_after(_iop_order_list, "bloom", "rgblevels", dont_move);
+//    _ioppr_move_iop_after(_iop_order_list, "bloom", "rgblevels", dont_move);
 
     // scene-linear to display-referred encoding
     // !!! WALL OF THE NON-LINEARITY !!! There is no coming back for colour ratios
+    // shadi ending this section
+    _ioppr_move_iop_after(_iop_order_list, "shadhi", "bloom", dont_move);
+
     _ioppr_move_iop_after(_iop_order_list, "basecurve", "bloom", dont_move);
     _ioppr_move_iop_after(_iop_order_list, "filmic", "basecurve", dont_move);
     _ioppr_move_iop_after(_iop_order_list, "colisa", "filmic", dont_move);
     _ioppr_move_iop_after(_iop_order_list, "tonecurve", "colisa", dont_move);
     _ioppr_move_iop_after(_iop_order_list, "levels", "tonecurve", dont_move);
 
-    _ioppr_move_iop_after(_iop_order_list, "shadhi", "levels", dont_move);
+//    _ioppr_move_iop_after(_iop_order_list, "shadhi", "levels", dont_move);
 
     // recover local contrast after non-linear tone edits
     _ioppr_move_iop_after(_iop_order_list, "bilat", "shadhi", dont_move);
 
     // display-referred colour edits
+    // colorcontrast ending this section
+    _ioppr_move_iop_after(_iop_order_list, "colorcontrast", "bilat", dont_move);
     _ioppr_move_iop_after(_iop_order_list, "colorcorrection", "bilat", dont_move);
     _ioppr_move_iop_after(_iop_order_list, "colorzones", "colorcorrection", dont_move);
     _ioppr_move_iop_after(_iop_order_list, "vibrance", "colorzones", dont_move);
     _ioppr_move_iop_after(_iop_order_list, "velvia", "vibrance", dont_move);
     _ioppr_move_iop_after(_iop_order_list, "colorize", "velvia", dont_move);
-    _ioppr_move_iop_after(_iop_order_list, "colorcontrast", "colorize", dont_move);
+//    _ioppr_move_iop_after(_iop_order_list, "colorcontrast", "colorize", dont_move);
 
     // fix clipping before going in colourout
     _ioppr_move_iop_before(_iop_order_list, "colorreconstruct", "colorout", dont_move);
@@ -391,7 +400,7 @@ static void _ioppr_insert_iop_before(GList **_iop_order_list, GList *history_lis
     if(found)
     {
       // set the iop_order
-      iop_order_new->iop_order = iop_order_prev + (iop_order_next - iop_order_prev) / 8.0;
+      iop_order_new->iop_order = iop_order_prev + (iop_order_next - iop_order_prev) / 2.0;
       if (DT_IOP_ORDER_INFO) fprintf(stderr,"\n  _ioppr_insert_iop_before %16s: %14.11f [xmp:%8.4f], prev %14.11f, next %14.11f",op_new,iop_order_new->iop_order,iop_order_new->iop_order,iop_order_prev,iop_order_next);
 
       // insert it on the proper order
@@ -498,7 +507,7 @@ static void _ioppr_move_iop_before(GList **_iop_order_list, const char *op_curre
   if(found)
   {
     // set the iop_order
-    iop_order_current->iop_order = iop_order_prev->iop_order + (iop_order_next->iop_order - iop_order_prev->iop_order) / 8.0;
+    iop_order_current->iop_order = iop_order_prev->iop_order + (iop_order_next->iop_order - iop_order_prev->iop_order) / 2.0;
 
     // insert it on the proper order
     iop_order_list = g_list_insert(iop_order_list, iop_order_current, position);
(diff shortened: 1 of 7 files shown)


spaceChRiS:
TLDR: Just a short question out of curiosity, why is the IOP order/pipe position implemented as float and not as linked list structure, as it seems this causes a lot of trouble?

I am not a programmer, but in the basic programming class at university we were thought all these fancy data structures and concepts, and it happens often that these concepts are not applied in real software. I wonder what the reason is, I'm sure I overlook something. This happened e.g. as well in image reordering in lighttable, where an integer is used to identify the position instead of a probably more robust list concept. Most probably the mistake is on my side, but I want to understand what I overlook when I ponder about these things.

parafin:
In case of iop order it's mostly historical - when iop order was fixed priorities were statically assigned to modules and it was the most straightforward solution. With flexible order linked list indeed makes more sense, but it was easier to just modify original concept instead of re-implementing it from scratch, I guess. But you would still need id (numerical is easiest) for each iop, since you need to serialize a link somehow in database and xmp. Same goes for lighttable - serializing a linked list isn't so obvious.

TurboGit:
I don't get it! How can we loose less digits? Maybe because when inserting we are lucky to insert in a way that does not stress much the list? If you insert always into the small interval created by / 8 it should be worst? And not only we insert, but we move also from time to time. I'm not sure how this can solve the issue...

jenshannoschwalm:
I'll try to explain 'visually'. I looked at the iop_orders generated at runtime while inserting and was surprised that in almost all cases it looked like this (add always 'basic' iop_order like 37.00) because insertions were **mostly** done 'after'
37. 0  -- 38.0
37. 0 -- **37 1/2** -- 38.0
37. 0 -- 37 1/2 -- **37 3/4** -- 38.0
37. 0 -- 37 1/2 -- 37 3/4 -- **37 7/8** -- 38.0 
... so after some inserts you end up at decimal 36.999999xx (you may well check that in history database) and when keeping data in xmp that was rounded to 37.0

So with my proposal you will have
37. 0 -- 38.0
37. 0 -- **37 1/8** -- 38.0
37. 0 -- 37 1/8 -- **37+1/8+7/64** -- 38.0
37.0 -- 37 1/8 -- 37+1/8+7/64 -- **37+1/8+7/6 + xxx** -- 38.0
...
After many insertions you will end up for the latest insertions much more distant from 37.0 (36.99x) and so are "loosing less digits". Of course the space between let's say 1/8->1/8+7/64 is much smaller but it's used **much much** less for insertions. 

You have the `DT_IOP_ORDER_INFO` in iop_order.c try and see the numbers before and after this pr.

parafin:
That only works if you always add before 38.0. If you actually always add after 37.0 this change would make things worse. It's not obvious why one case should be more often than another...

parafin:
For your example it works: original algorithm gives (1/2)^n distance after n insertions, yours gives (7/8)^(n-1)-(7/8)^n == (7/8)^(n-1)/8, which becomes bigger starting with n == 4.

src/common/iop_order.c
```diff
@@ -391,7 +391,7 @@ static void _ioppr_insert_iop_before(GList **_iop_order_list, GList *history_lis
     if(found)
     {
       // set the iop_order
-      iop_order_new->iop_order = iop_order_prev + (iop_order_next - iop_order_prev) / 2.0;
+      iop_order_new->iop_order = iop_order_prev + (iop_order_next - iop_order_prev) / 8.0;
```
aurelienpierre: not sure I get the reason here. `/ 2.0` was a bissection between previous and last module. What does the `/8.0` to the logic of the code ?

jenshannoschwalm:
Yes i was wondering about that too. Look at the way the modules get their default iop_order
[using_2.txt](https://github.com/darktable-org/darktable/files/3665819/using_2.txt)
[using_8.txt](https://github.com/darktable-org/darktable/files/3665821/using_8.txt)

**When you look carefully at the master version divide by 2**
  db: 38,96875000000   xmp  38,9688            sharpen
  db: 38,98437500000   xmp  38,9844       channelmixer
  db: 38,99218750000   xmp  38,9922       colorchecker
  db: 38,99414062500   xmp  38,9941      colortransfer
  db: 38,99609375000   xmp  38,9961       colormapping
  db: 38,99804687500   xmp  38,9980       colorbalance
  db: 38,99902343750   xmp  38,9990           basicadj
  db: 38,99951171875   xmp  38,9995           rgbcurve
  db: 38,99975585938   xmp  38,9998          rgblevels
  db: 38,99987792969   xmp  38,9999              bloom
  db: 38,99993896484   xmp  38,9999          basecurve
  db: 38,99996948242   xmp  39,0000             filmic
  db: 38,99997711182   xmp  39,0000          filmicrgb
  db: 38,99998474121   xmp  39,0000             colisa
  db: 38,99999237061   xmp  39,0000          tonecurve
  db: 38,99999618530   xmp  39,0000             levels
  db: 38,99999809265   xmp  39,0000             shadhi
  db: 38,99999904633   xmp  39,0000              bilat
  db: 38,99999952316   xmp  39,0000    colorcorrection
  db: 38,99999976158   xmp  39,0000         colorzones
  db: 38,99999988079   xmp  39,0000           vibrance
  db: 38,99999994040   xmp  39,0000             velvia
  db: 38,99999997020   xmp  39,0000           colorize
  db: 38,99999998510   xmp  39,0000      colorcontrast
  db: 39,00000000000   xmp  39,0000      globaltonemap
**All** modules above colisa when using the recently introduced iop_order sheme had the same iop_orders in the xmp files and so we had theses massive problems so hard to understand. History forgetting, visual artefacts... This is ok as long as we don't add any more modles between colorcontrast and globaltonemap.

**For divide by 8**
  db: 38,48709106445   xmp  38,4871            sharpen
  db: 38,55120468140   xmp  38,5512       channelmixer
  db: 38,60730409622   xmp  38,6073       colorchecker
  db: 38,61343996972   xmp  38,6134      colortransfer
  db: 38,65639108419   xmp  38,6564       colormapping
  db: 38,69934219867   xmp  38,6993       colorbalance
  db: 38,73692442384   xmp  38,7369           basicadj
  db: 38,76980887086   xmp  38,7698           rgbcurve
  db: 38,79858276200   xmp  38,7986          rgblevels
  db: 38,82375991675   xmp  38,8238              bloom
  db: 38,84578992716   xmp  38,8458          basecurve
  db: 38,86506618626   xmp  38,8651             filmic
  db: 38,86717452710   xmp  38,8672          filmicrgb
  db: 38,88193291298   xmp  38,8819             colisa
  db: 38,89669129886   xmp  38,8967          tonecurve
  db: 38,90960488650   xmp  38,9096             levels
  db: 38,92090427569   xmp  38,9209             shadhi
  db: 38,93079124123   xmp  38,9308              bilat
  db: 38,93944233607   xmp  38,9394    colorcorrection
  db: 38,94701204406   xmp  38,9470         colorzones
  db: 38,95363553856   xmp  38,9536           vibrance
  db: 38,95943109624   xmp  38,9594             velvia
  db: 38,96450220921   xmp  38,9645           colorize
  db: 38,96893943306   xmp  38,9689      colorcontrast
  db: 39,00000000000   xmp  39,0000      globaltonemap
the situation is quite different, of course they are close together but safely keep distinct.

Whatever you do later -- you can safely do a division by 2 several times without getting undistintable -- depends vastly on how the iop_orders are distributed when getting the default iop_orders.

Before introducing the new iop_ordering this just didn't become obvious because the number of iterations was **much** smaller so everything kept 'in precision' of a single float for xmp.
BTW this was the reason it was sometimes so difficult to reproduce a problem, malfunctioning depended on a) data read from xmp and b) the order of the module. As we later move orders a bit this was evident for some modules, for some it was not.

parafin:
Why all the numbers are so close to 39? I think that's the real problem here.

jenshannoschwalm:
> Why all the numbers are so close to 39? I think that's the real problem here.

Agreed. The set of rules _defines_ this. With other rules the bisection by 2 could be much better.
L114 in iop_order.c and following are the interesting ones and correspond to always bisecting to the upper half for many times. Changing these rules for a more even distribution?

jenshannoschwalm:
What about `_ioppr_move_iop_before` to bisect at 7/8 and ` _ioppr_move_iop_after` at 1/8?

aurelienpierre:
The problem of iop_order is rules are not flattened before resolution, so the actual order depends on the positional constrains you set relatively to modules, but also in which order you script them. Meaning same constrains written in a different order lead to different results.

The state of the art way to do it is to build a graph of distances (relative distance between module) using the local constraints (module to module), then [solve it](https://www-sop.inria.fr/members/Nicolas.Nisse/lectures/0Intro.pdf).

jenshannoschwalm:
> The state of the art way to do it is to build a graph of distances (relative distance between module) using the local constraints (module to module), then solve it.

Yes, true and full of art. I just don't feel like doing such a thing in this part of code as i am not sure how to be safe later while developing.

> Meaning same constrains written in a different order lead to different results.

Or you could say the current code is not artfull. Look at L114 to 150!

How to solve?
1. more arts while defining the list?
2. `_ioppr_move_iop_after` leaving more room for following `_ioppr_move_iop_after` (section at 1/8) **and** `_ioppr_move_iop_before` also leaving more room (section at 7/8)

aurelienpierre:
> How to solve?

    more arts while defining the list?
    _ioppr_move_iop_after leaving more room for following _ioppr_move_iop_after (section at 1/8) and _ioppr_move_iop_before also leaving more room (section at 7/8)

But these are temporary and local patches to hide dust under the carpet and pretend it's clean. It's doom to fail again in the future. There was a python file that solved the graph of nodes then wrote their priorities in the C code in `src/tools`, I don't find it anymore.

jenshannoschwalm:
Your dust and carpet analogy is nice. Of course a thorough cleaning is a good solution but when your mother-in-law is arriving in 10min it may well be solution. #3041 is just defining the order in a different way -- for some time (for this release) this will hide the underlying problem and will remind us for later more profound work.

aurelienpierre:
If we release brittle code to handle the biggest change of darktable 3.0 we are so proud of, aka pipe reordering, then need to change it in a later version because it fails, and users lose their history on previous edits because priority computations are not the same, we will look like the merry little band of fools we are. 

I will write it myself if needed, but this is not good enough as-is. Delay the release if it comes to that, but I'm not ok to release something half-done and hastily patched.

jenshannoschwalm:
@aurelienpierre i agree with you on many parts, none of us wants to look like or even be a fool. I understand why it is so important and such a big step in dt history to have pipe reordering. But pipe reordering is not the problem here -- at least i don't think so.

The choosen algorithm to insert a iop_order "sectioning by two" if fully ok and probably the best guess if you insert randomly. But in lately introduced new default iop orders this is definitely not the case. (The reasons to introduce this btw are absolutely right and fully accepted). But: When defining the default rules i counted 26 insertions to the high (leaving 1/2^26). **This** has been the major problem with modules working out of order or history confusion i guess, squeezing all numbers closer and closer to 39 stretching the precision for xmp files too far.

maybe we need some distance checking while inserting an iop_order that could
1. atm at least give a warning to make the situation understandable
2. be used to trigger a not yet existing re-calculation.

Keep going, none of us want to write bad code, neither in style, performance nor correctness. I promise to try hard! It's not easy with such a large code base to understand all implications of changing parts of the code but i'll improve on this ...

aurelienpierre:
> The choosen algorithm to insert a iop_order "sectioning by two" if fully ok and probably the best guess if you insert randomly. But in lately introduced new default iop orders this is definitely not the case. (The reasons to introduce this btw are absolutely right and fully accepted). But: When defining the default rules i counted 26 insertions to the high (leaving 1/2^26). This has been the major problem with modules working out of order or history confusion i guess, squeezing all numbers closer and closer to 39 stretching the precision for xmp files too far.

What concerns me here is the order in which you define constraints (in the code) for module order affects the computed order too. The new IOP order only shows the limit of that (of course, you can tweak the definitions to hide it), but it means that if users move close modules too many times, they will hit the same dead-end. 

While being at it, maybe make the new default order a static ordered list and skip the dynamic insertions. I confess I didn't investigate that.

TurboGit:
@aurelienpierre @jenshannoschwalm can you have a look at #3046 

As I said not perfect but leave room for insertion and multi-instance.

PR #3040, opened 2019-09-29
Improve readability of preference window

amousset:
The idea is to make the preference window more consistent with the rest of the interface, and a bit more dense.

* Add a "General" section for language and theme for consistency
* Make the preferences a bit more dense (especially as their number continues growing) (this is achieved by replacing the homogeneity of row height by a min-height, which is not perfect, but prevents the language field which is bigger than others to be used as a minimum)
* Make the section title style more consistent with rest of the interface (left align and no underlining)
* Add some space between widgets and scrollbar

comp2

### DIFF
--- a/data/themes/darktable.css
+++ b/data/themes/darktable.css
@@ -1135,9 +1135,18 @@ cell:selected
 
 #preferences_notebook scrolledwindow #pref_section > label
 {
-    font-size: 1.5em;
+    font-size: 1.2em;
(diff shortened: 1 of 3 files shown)


Nilvus:
I like these improvements except titles. I find them less readable now even if it's more consistent with the rest of the UI.

TurboGit:
Better consistency and quite better overall. Thanks.

PR #3041, opened 2019-09-29
Relaxing stress on iop_orders by regrouping default iop_orders

jenshannoschwalm:
Unfortunately i completely messed up the branch for #3037

So here is a new suggestion leaving the algorithm as is but re-grouping the
modules iop_order defaults to have less stress on double float precision.

It's not perfect but leaves the rules as much as possible for better understanding.

### DIFF
--- a/src/common/iop_order.c
+++ b/src/common/iop_order.c
@@ -33,7 +33,7 @@
 
 #define DT_IOP_ORDER_VERSION 3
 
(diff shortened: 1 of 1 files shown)


TurboGit:
We can't change the order of v3. This will break current edits.

See #3046 for a proposal.

jenshannoschwalm:
ok. Closing this for now.

PR #3045, opened 2019-09-29
Fix/comparison result is always the same

codingdave:
works on issues reported by https://lgtm.com/projects/g/darktable-org/darktable/?mode=tree&ruleFocus=2154840804

### DIFF
--- a/src/common/points.h
+++ b/src/common/points.h
@@ -1056,27 +1056,9 @@ void init_by_array(sfmt_state_t *s, uint32_t *init_key, int key_length)
 {
   int i, j, count;
   uint32_t r;
-  int lag;
-  int mid;
+  int lag = 11;
   int size = N * 4;
-
-  if(size >= 623)
-  {
-    lag = 11;
-  }
-  else if(size >= 68)
-  {
-    lag = 7;
-  }
-  else if(size >= 39)
-  {
-    lag = 5;
-  }
-  else
-  {
-    lag = 3;
-  }
-  mid = (size - lag) / 2;
+  int mid = (size - lag) / 2;
 
   memset(s->sfmt, 0x8b, sizeof(s->sfmt));
   if(key_length + 1 > N32)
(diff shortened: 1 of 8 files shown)


src/common/points.h
```diff
@@ -1056,27 +1056,9 @@ void init_by_array(sfmt_state_t *s, uint32_t *init_key, int key_length)
 {
   int i, j, count;
   uint32_t r;
-  int lag;
-  int mid;
+  int lag = 11;
   int size = N * 4;
```
TurboGit: I don't think this is correct. N is defied as:

#define N (MEXP / 128 + 1)

And MEXP comes from an include definition which may depends on the platform.

size = N * 4 is 155 if MEXP is set to default (see top of this file), and so path "size  >= 60" will be taken.

src/iop/demosaic.c
```diff
@@ -2073,11 +2073,6 @@ static void xtrans_fdc_interpolate(struct dt_iop_module_t *self, float *out, con
             maxval = (maxval < hm[d] ? hm[d] : maxval);
           }
           maxval -= maxval >> 3;
-          for(int d = 0; d < ndir - 4; d++)
```
TurboGit: This looks very suspicious to me!

src/iop/liquify.c
```diff
@@ -2804,7 +2805,7 @@ int mouse_moved (struct dt_iop_module_t *module,
                  int which)
 {
   dt_iop_liquify_gui_data_t *g = (dt_iop_liquify_gui_data_t *) module->gui_data;
-  int handled = g->last_hit.elem ? 1 : 0;
+  bool handled = g->last_hit.elem ? true : false;
```
TurboGit: no bool (C++ things) we use boolean.

codingdave:
Is it ok to correct the commits (rewriting history) and force push? Or will the discussion above get damaged?

TurboGit:
Yes it is correct to rewrite and force push. It is what I would expect from any topic branch.

PR #3046, opened 2019-09-30
iop_order: new version that ensures all orders have room for insertion.

TurboGit:
What about this to relax the iop-order list ?

### DIFF
--- a/src/common/iop_order.c
+++ b/src/common/iop_order.c
@@ -348,7 +348,7 @@ dt_iop_order_entry_t *dt_ioppr_get_iop_order_entry(GList *iop_order_list, const 
 double dt_ioppr_get_iop_order(GList *iop_order_list, const char *op_name)
 {
   double iop_order = DBL_MAX;
-  dt_iop_order_entry_t *order_entry = dt_ioppr_get_iop_order_entry(iop_order_list, op_name);
+  const dt_iop_order_entry_t *order_entry = dt_ioppr_get_iop_order_entry(iop_order_list, op_name);
 
   if(order_entry)
     iop_order = order_entry->iop_order;
--- a/src/common/iop_order.c
+++ b/src/common/iop_order.c
@@ -31,7 +31,7 @@
 #include <stdlib.h>
 #include <string.h>
 
-#define DT_IOP_ORDER_VERSION 3
+#define DT_IOP_ORDER_VERSION 4
 
 #define DT_IOP_ORDER_INFO FALSE	// used while debugging
 
@@ -40,6 +40,24 @@ static void _ioppr_insert_iop_before(GList **_iop_order_list, GList *history_lis
 static void _ioppr_move_iop_after(GList **_iop_order_list, const char *op_current, const char *op_prev, const int dont_move);
 static void _ioppr_move_iop_before(GList **_iop_order_list, const char *op_current, const char *op_next, const int dont_move);
 
+// this routine rewrite the iop-order to have all them evenly spaced
+// into the list ensuring that we can insert safly at least 40 iop
+// between any two of them. Probably not the best fix, this is mostly
+// a workaround the current limitation and will avoid issues for next
+// release.
+// ?? TODO: redo the whole pipe ordering
+static void _rewrite_order(GList *iop_order_list)
+{
+  GList *l = iop_order_list;
+  double order = 1.0;
+  while(l)
+  {
+    dt_iop_order_entry_t *order_entry = (dt_iop_order_entry_t *)l->data;
+    order_entry->iop_order = order;
+    order += 1.0;
+    l = g_list_next(l);
+  }
+}
 
 /* migrates *_iop_order_list from old_version to the next version (version + 1)
  * limitations:
@@ -159,6 +177,11 @@ static int _ioppr_legacy_iop_order_step(GList **_iop_order_list, GList *history_
     _ioppr_insert_iop_after(_iop_order_list, history_list, "filmicrgb", "filmic", dont_move);
 
     new_version = 3;
+  }
+  else if(old_version == 3)
+  {
+    if(!dont_move) _rewrite_order(*_iop_order_list);
+    new_version = 4;
   }
 
   if(new_version <= 0)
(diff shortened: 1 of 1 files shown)


parafin:
What does iop_order look like for fresh image? Because here I see only conversion from previous version.

TurboGit:
I'll close, it looks like this "simple" idea just does not fly.... Sorry for the noise.

TurboGit:
In fact I did not test correctly. This new version helps a lot:

With previous version (v3):
```
0|rawprepare|1.0
1|temperature|3.0
2|highlights|4.0
3|demosaic|7.50003051757812
4|colorin|37.5
5|colorout|63.5
6|gamma|71.0
7|flip|8.0078125
8|basecurve|38.9999389648437
9|exposure|8.25
10|toneequal|21.5
11|clipping|8.015625
12|clipping|8.015625
13|invert|2.0
14|ashift|8.00390625
15|lens|8.0009765625
16|sharpen|38.96875
17|hazeremoval|8.001953125
18|cacorrect|5.0
```

With this new version (v4 new edit):
```
iop order for 332:
0|rawprepare|1.0
1|temperature|3.0
2|highlights|4.0
3|demosaic|8.0
4|colorin|29.0
5|colorout|69.0
6|gamma|77.0
7|flip|16.0
8|basecurve|45.0
9|exposure|21.0
10|toneequal|24.0
11|shadhi|51.0
12|colorreconstruct|68.0
13|clipping|17.0
14|clipping|17.0
15|basecurve|45.0
16|invert|2.0
17|spots|19.0
18|defringe|31.0
19|sharpen|35.0
20|lens|13.0
21|hazeremoval|14.0
22|ashift|15.0
```
So as you see, when only activating modules you get all iop_order evenly spaced with a distance of 1. This leaves room for lot of insertions and multi-instance between any modules. 

This is not perfect but will let me more confident that we won't break edits on v3.

TurboGit:
Fact is that we won't re-implement the pipe order algorithm for v3.0. This should help avoiding most if not all issues.

aurelienpierre:
Why not just rewrite statically the pipe order as a flat list, like the legacy one (as of 2.6) was defined ? I think we are trying to avoid the drawbacks of the priority computations by over-engineering it.

TurboGit:
Because every time you'll add a new module you'll have to list the 70 or more modules again and again. Here you'll just have to do:

```
  else if(old_version == <N>)
  {
    _ioppr_move_iop_after(_iop_order_list, "new_module", "some_module", dont_move);
    _rewrite_order(*_iop_order_list);
    new_version = <N+1>;
  }
```
A trade-off !

TurboGit:
Let's reopen.

aurelienpierre:
> Because every time you'll add a new module you'll have to list the 70 or more modules again and again. Here you'll just have to do:

Sure, but it's not what I mean. I mean let's flatten the v3, since it's such a big rewrite, to make sure at least it starts clean, then start building again on top of it with dynamic insertions when new modules are added.

jenshannoschwalm:
> Fact is that we won't re-implement the pipe order algorithm for v3.0. This should help avoiding most if not all issues.

Agreed. 

Maybe also checks for delta iop_order beeing less than used precision resulting in a warning box ?

TurboGit:
> Sure, but it's not what I mean. I mean let's flatten the v3

What do you mean by flatten? I fear that we can't change v3 now as it will break the current edits using it.

aurelienpierre:
> What do you mean by flatten? I fear that we can't change v3 now as it will break the current edits using it.

Write it as a simple static list of iop names with no dynamic insertions, as in `static GList *_ioppr_get_iop_order_v1()`

TurboGit:
I see, let's do that indeed. Then v3 will be readable as a first clean step.

jenshannoschwalm:
Any problems with current edits this way?

TurboGit:
Current edits will have the issue but this will only affect current users of the dev version which is far less that the user code base on 2.6.2 who will be migrating.

jenshannoschwalm:
I thought -- we need a warning about this isuue then or open an Issue describing the modifications and what to do to be in a clean state for an Image. Otherwise we will see many issues popping up :-)

TurboGit:
> I thought -- we need a warning about this isuue then or open an Issue describing the modifications and what to do to be in a clean state for an Image.

Well, I'm not sure about that. People using the dev versions are prepared for this and this will be very rare issue (need many reordering and/or some multi-instances). So the probability is low!

And as I said people migrating from 2.6.2 will be immuned to this issue with my patch on this PR.

jenshannoschwalm:
There is one thing i am suspicious about.

The data in `const dt_iop_order_entry_t iop_v3[]` -- i guess you took that from some log.

I don't understand the full code path while upgrading from v3->v4, do you have to compare those values with data from the pre-existing database?

If so, there might be subtle differences / rounding errors as data from the old database/xmp files. Is this properly checked?

TurboGit:
> The data in `const dt_iop_order_entry_t iop_v3[]` -- i guess you took that from some log.

Yes.

> I don't understand the full code path while upgrading from v3->v4, do you have to compare those values with data from the pre-existing database?

No. The new values are creating a new iop_version. The current edits on v3 will use the v3 ordering and only the new edits will be using v4.

TurboGit:
Ok, wrong, I'm wrong! The iop-order MUST BE CONSTANT from one version to another. The comments seems misleading or there is really something I don't understand.

@edgardoh : can you shed some light please? this is a quite critical issue for the next release. thanks.

jenshannoschwalm:
WHY do they have to be constant? I thought in earlier versions we had a defined pipeline order.

Has this order really changed?
1. If not, can't we read the used iop_orders in history and sort them in to current default ranges on_the_fly?
2. If yes, can't we define a simple matrix telling us in which order the modules must be included?

TurboGit:
Good question! I don't have the clue at the moment but iop-order must be constant from one version to another it seems and we can only insert new module in successive version. Just try the first commit on this patch set and you'll see that it breaks current edit by displaying errors about gamma not being the last in the pipe for example. Strange as the first commit only a version 4 that I thought would not be used, but it is.... Which somehow makes sense since if there is a new module in v4 it must be known and positioned in the pipe.

And so, my v4 which is just re-aranging the iop-order values to have a distance of one is breaking current edit (gamma was 71 on v3) and finalscale module is set to 71 after the call to _rewrite_order().

So my conclusion is that v3 edit on the db must be kept AS-IS and not changed in any way. Maybe there is a missing call somewhere when reading an edit? Don't know...

I've asked @edgardoh to help.

TurboGit:
I may have a new version working see #3053.

Closing this one.

PR #3047, opened 2019-09-30
Marking obsolete modules

jenshannoschwalm:
When reopening an already developed image using modules which concidered
obsolete the is an indicator in the history indicating such.

See discussion in #3013


TurboGit:
Great! Thanks.

PR #3048, opened 2019-09-30
tone equalizer: more general default params

aurelienpierre:
This should help users who don't want/know how to mess with masking parameters to get a proper histogram spreading over the nodes out-of-the-box.


TurboGit:
Ok, thanks.

PR #3051, opened 2019-10-01
module name sizing moved from code to darktable.css

phweyland:
on windows sans-serif has no condensed counter part. 
If this is the same on other OS, removing the  `font-stretch: condensed;` option from darktable.css would avoid the console messages :

```
(darktable.exe:68480): Pango-WARNING **: 12:13:30.447: couldn't load font "sans-serif Condensed Not-Rotated 10.349609375", falling back to "Sans Condensed Not-Rotated 10.349609375", expect ugly output.
(darktable.exe:68480): Pango-WARNING **: 12:13:30.447: couldn't load font "Sans Condensed Not-Rotated 10.349609375", falling back to "Sans Not-Rotated 10.349609375", expect ugly output.
```

### DIFF
--- a/data/themes/darktable.css
+++ b/data/themes/darktable.css
@@ -198,7 +198,7 @@ entry *,
 dialog,
 dialog *,
 colorswatch,
-colorswatch *
+colorswatch *,
 stack,
 stack *,
 scrollbar,
@@ -464,7 +464,7 @@ entry selection
   font-weight: normal;
   font-stretch: condensed;
   font-family: sans-serif;
-  font-size: 0.9em;
+  font-size: larger;
 }
 
 /* Labels of controls sections in modules */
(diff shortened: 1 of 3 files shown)


phweyland:
Sorry, I see a previous commit from a previous branch. Will clean up.

data/themes/darktable.css
```diff
@@ -198,7 +198,7 @@ entry *,
 dialog,
 dialog *,
 colorswatch,
-colorswatch *
+colorswatch *,
```
TurboGit: Good catch ! A nice bug.

TurboGit:
Thanks!

data/themes/darktable.css
aurelienpierre: Indeed.

PR #3052, opened 2019-10-01
culling : fine tune selection frame

AlicVB:
split colors for culling selected frame in order to : 
- hide frame in culling for darktable-elegant-* (same as before)
- always show frame in culling filmstrip

both colors (culling main part and culling filmstrip part) are themable


TurboGit:
We are speaking of the black frame on first image in culling mode? This seems removed in grey and dark theme, but not in darker one.

AlicVB:
oops, forgot that new theme. This should be fixed now

TurboGit:
Looks good now. Thanks.

PR #3053, opened 2019-10-01
po/iop order v4

TurboGit:
Please test, I suppose this is working.

### DIFF
--- a/src/common/iop_order.c
+++ b/src/common/iop_order.c
@@ -31,7 +31,7 @@
 #include <stdlib.h>
 #include <string.h>
 
-#define DT_IOP_ORDER_VERSION 3
+#define DT_IOP_ORDER_VERSION 4
 
 #define DT_IOP_ORDER_INFO FALSE	// used while debugging
 
@@ -40,6 +40,24 @@ static void _ioppr_insert_iop_before(GList **_iop_order_list, GList *history_lis
 static void _ioppr_move_iop_after(GList **_iop_order_list, const char *op_current, const char *op_prev, const int dont_move);
 static void _ioppr_move_iop_before(GList **_iop_order_list, const char *op_current, const char *op_next, const int dont_move);
 
+// this routine rewrite the iop-order to have all them evenly spaced
+// into the list ensuring that we can insert safely at least 40 iop
+// between any two of them. Probably not the best fix, this is mostly
+// a workaround for current limitation and will avoid issues for the next
+// release.
+// ?? TODO: redo the whole pipe ordering
+static void _rewrite_order(GList *iop_order_list)
+{
+  GList *l = iop_order_list;
+  double order = 1.0;
+  while(l)
+  {
+    dt_iop_order_entry_t *order_entry = (dt_iop_order_entry_t *)l->data;
+    order_entry->iop_order = order;
+    order += 1.0;
+    l = g_list_next(l);
+  }
+}
 
 /* migrates *_iop_order_list from old_version to the next version (version + 1)
  * limitations:
@@ -160,6 +178,43 @@ static int _ioppr_legacy_iop_order_step(GList **_iop_order_list, GList *history_
 
     new_version = 3;
   }
+  else if(old_version == 3)
+  {
+    // version 4 is a rewrite of the iop-order of previous list. As it
+    // can be seen above some modules have very close iop-order values
+    // and does not have lot of room for insertion (multi-instance
+    // module and/or reordering).
+
+    if(!dont_move) _rewrite_order(*_iop_order_list);
+    new_version = 4;
+  }
+  // each new version MUST be written as the following (_rewrite_order IS VERY important)
+
+  // If a new module is to be added, it must be added in the current
+  // version (just above) and then a new version must be created with
+  // the following code to ensure modules are evenly spaced (leaving
+  // room for multi-instances and user's re-ordering):
+  /*
+  else if(old_version == <N>)
+  {
+    // reorder modules to ensure they are all evenly spaced
+    if(!dont_move) _rewrite_order(*_iop_order_list);
+    new_version = <N+1>;
+  }
+  */
+
+  // If a module is to be reodered in the pipe, the following code
+  // must be used. A new version must be created to ensure modules are
+  // evenly spaced (leaving room for multi-instances and user's
+  // re-ordering):
+  /*
+  else if(old_version == <N>)
+  {
+    _ioppr_move_iop_[before|after](_iop_order_list, "new_module", "some_module", dont_move);
+    if(!dont_move) _rewrite_order(*_iop_order_list);
+    new_version = <N+1>;
+  }
+  */
 
   if(new_version <= 0)
     fprintf(stderr, "[_ioppr_legacy_iop_order_step] missing step migrating from version %i\n", old_version);
--- a/src/common/iop_order.c
+++ b/src/common/iop_order.c
@@ -380,7 +380,7 @@ dt_iop_order_entry_t *dt_ioppr_get_iop_order_entry(GList *iop_order_list, const 
 double dt_ioppr_get_iop_order(GList *iop_order_list, const char *op_name)
 {
   double iop_order = DBL_MAX;
-  dt_iop_order_entry_t *order_entry = dt_ioppr_get_iop_order_entry(iop_order_list, op_name);
+  const dt_iop_order_entry_t *order_entry = dt_ioppr_get_iop_order_entry(iop_order_list, op_name);
 
   if(order_entry)
     iop_order = order_entry->iop_order;


TurboGit:
Hopefully this will fix iop-order issues for dt 3.0.

jenshannoschwalm:
Unfortunately i can't test properly until friday but reading the code looks **really** good.
1. It looks like v3 edits will translate to v4 on the fly, right?
2. Maybe edits from the time before proper xmp writing/reading might have a problem?

src/common/iop_order.c
```diff
@@ -40,6 +40,24 @@ static void _ioppr_insert_iop_before(GList **_iop_order_list, GList *history_lis
 static void _ioppr_move_iop_after(GList **_iop_order_list, const char *op_current, const char *op_prev, const int dont_move);
 static void _ioppr_move_iop_before(GList **_iop_order_list, const char *op_current, const char *op_next, const int dont_move);
 
+// this routine rewrite the iop-order to have all them evenly spaced
+// into the list ensuring that we can insert safly at least 40 iop
```
parafin: safely, iops

src/common/iop_order.c
```diff
@@ -40,6 +40,24 @@ static void _ioppr_insert_iop_before(GList **_iop_order_list, GList *history_lis
...
+// between any two of them. Probably not the best fix, this is mostly
+// a workaround the current limitation and will avoid issues for next
```
parafin: workaround for, for the next

src/common/iop_order.c
```diff
@@ -74,92 +92,126 @@ static int _ioppr_legacy_iop_order_step(GList **_iop_order_list, GList *history_
...
-    // fix clipping before going in colourout
-    _ioppr_move_iop_before(_iop_order_list, "colorreconstruct", "colorout", dont_move);
-    _ioppr_move_iop_before(_iop_order_list, "vignette", "colorreconstruct", dont_move);
-
-    _ioppr_move_iop_before(_iop_order_list, "dither", "borders", dont_move);
-
-    // new modules here
-    _ioppr_insert_iop_after(_iop_order_list, history_list, "filmicrgb", "filmic", dont_move);
+    // The following is a flattened list from original code. The goal is to have a clean starting point for
+    // futur modifications.
```
parafin: future

TurboGit:
>     1. It looks like v3 edits will translate to v4 on the fly, right?

No, the v3 will stay on v3. This new version just ensure that v3 won't have order of the *module* set to v4 (!dont_move before rewite_order call). That was the issue on my previous version, the history order was on v3 but the modules on v4, so the iop-order were not synchronized at all.

>     2. Maybe edits from the time before proper xmp writing/reading might have a problem?

I don't think so, but I'm not sure to see the scenario. I'll wait for more testing anyway before merging, but on my side all the conflicts I had with the previous version are gone.

src/common/iop_order.c
TurboGit: Fixed.

src/common/iop_order.c
TurboGit: Fixed.

src/common/iop_order.c
TurboGit: Fixed.

jenshannoschwalm:
Hi @TurboGit 

I found time to test this pr and thanks about information about **not** transforming v3->v4. This is important!

1. Of course with fresh history all developments now are fine with good enough distance for sure.
2. I looked at quite a number of images developed with iop_order v2 and they all look ok, order is kept and when re-developing everything stayed intact -- although i might have missed some cases.

For images developed with V3 the situation is a bit different. Let me explain.

We don't have a **definite** V3 because unfortunately some PR's like introducing V3 and reading/writing xmp data with another precision lead to inconsistencies.

I see these problems:

The list of iop_orders you got from the log has a number of significant decimal digits that might not be the same as the value read from the double in history#s iop_order.

- The decimal value might have been rounded up so comparing both is slightly wrong. If you have **not** deleted and reimported you are probably pretty safe (until you do more
critical stuff)
- if you have deleted and reimported **after** high precision reading/wrinting you might have more rounding errors that could again result in wrong order.

Suggestion for both above cases: wouldn't it be better to slightly decrease the given number in the V3-table (the last digit) ?

**A really different story is for V3 developed images if the image has been deleted and re-imported with low precision**

I really think these developments can't be saved at all! (You already see some issues where dt is crashing or at least showing big problems.) Why? Here you have quite **a number of modules having the same iop_order of 39.0** I don't know what module is exactly is making dt crash -- but i observed it too when working on xmp precision.

What can be done here -- (it's only evident for people working on a daily basis with fresh master)?

Or - can we define an absolutely unsafe situation telling before opening the image?
I think so;
- If we find **any** module with an iop_order in history where the decimal fraction is larger than lets say 0.999999999 this **must** be wrong
- if the iop_order is 39 for any of the modules filmic to colorcontrast it **must** be wrong.
Both situations will either crash darktable right now or lead to problems very soon.

I would suggest a testing for both cases before open-for-develope and in case open a box suggesting to delete history and deny opening.

jenshannoschwalm:
>I would suggest a testing for both cases before open-for-develope and in case open a box suggesting to delete history and deny opening.

 Or maybe _be rude-but-safe_ and clear history for such images?

parafin:
If we find iops with the same priority, we could fix it by changing the priority so that order of those iops is the same as in unmodified original iop order. No need to destroy user's history.

TurboGit:
> Suggestion for both above cases: wouldn't it be better to slightly decrease the given number in the V3-table (the last digit) ?

No, I think the best is to forget about the "clean-state" and so removing the last commit in this patch set to keep the very v3 iop-order as before.

> What can be done here -- (it's only evident for people working on a daily basis with fresh master)?

Yes, basically v3 (whatever the version is) is fully broken in different ways. This indeed will affect only users of the master version. So people compiling dt and certainly prepared to understand the situation.

TurboGit:
> If we find iops with the same priority, we could fix it by changing the priority so that order of those iops is the same as in unmodified original iop order. No need to destroy user's history.

This is ok for main modules, but this issue can arise (and is more likely to happen) for multi-instance. And in this case there is not easy new iop-order to set.

TurboGit:
New version pushed with:
- no more flattening as it introduces issues.
- more comments about the way to handle new iop and to move iop in the pipe

parafin:
> > If we find iops with the same priority, we could fix it by changing the priority so that order of those iops is the same as in unmodified original iop order. No need to destroy user's history.
> 
> This is ok for main modules, but this issue can arise (and is more likely to happen) for multi-instance. And in this case there is not easy new iop-order to set.

My point is that even random order is better than dropping all history or crashing. For multi-instance you can look at number of the instance for example.

TurboGit:
@parafin : sure, I'm not for dropping history.

The best would be to have a clean migration from v3 (broken in many ways) to v4. Maybe a script could be provided to help with that. The algorithm would be:

for a given picture:
- for each "operation"
- get iop_order for this operation in v4 set -> NEW_IOP_ORDER
- if multi_instance is > 0 : Add multi_instance / 100 to NEW_IOP_ORDER
- set iop_order_version = 4 for the given picture
- update iop_order in history with the new calculated one

How does that sounds?

Any volunteer to write such script?

jenshannoschwalm:
What's the difference between a script and on-the-fly conversion?

TurboGit:
The on-the-fly is certainly more work and a code that will be useful only for user's of master since some months only. Some code to maintain and this cost. The idea of the script is to be hepfull but with not maintenance and something we can drop at some point.

That was my thinking.

parafin:
darktable still need to handle the case of matching iop orders, script won't solve it.

jenshannoschwalm:
Script for 3 OS's? 

Probably i don't understand the iop re-ordering thing correctly? What that affect iop orders not being correctly handled by script / on-the-fly?

TurboGit:
> darktable still need to handle the case of matching iop orders, script won't solve it.

iop-order are static, I can dump them for every iop for v4. Then matching operation is easy. Just follow the algo above and we should be fine.

TurboGit:
> Script for 3 OS's?

Yes :( That mean probably Python.

jenshannoschwalm:
No chance :-) Suggest FORTH and i'm you man :-) But probably that would mean many questions and a tutorial howto do that. So another no.

On-the-fly? When reading history?

TurboGit:
@jenshannoschwalm : ok, if you want to go ahead on-the-fly in the code. In fact this could be reused later to migrate a picture from one iop-order version to another. At the moment, please ONLY handle v3 to v4 to minimize the change. Is the algo I described clear enough?

TurboGit:
In the mean time, since this PR is ok I will merge to avoid more people using broken iop-order v3.

jenshannoschwalm:
It's ok, will take a few days though, too much work in hospital atm

parafin:
> > darktable still need to handle the case of matching iop orders, script won't solve it.
> 
> iop-order are static, I can dump them for every iop for v4. Then matching operation is easy. Just follow the algo above and we should be fine.

Maybe we are talking about different things. I'm saying that when image history has 2 modules with the same order/priority float number darktable needs to handle this situation gracefully somehow (I've suggested how to do it). What should exactly the script you are talking about do?

PR #3054, opened 2019-10-01
save export metadata parameters to darktablerc

phweyland:
as @aurelienpierre suggested on [PR #2889](https://github.com/darktable-org/darktable/pull/2889#issuecomment-536600489)


phweyland:
when formulas are removed the last darktablerc formulas entries remain with no value.

PR #3058, opened 2019-10-02
split tags and categories in image information

phweyland:
Tags which are set as categories may not make sense without the parent category.
This PR to display separately tags and categories in image information.

Example:
Annotation 2019-10-02 095657 PNG

### DIFF
--- a/src/libs/metadata_view.c
+++ b/src/libs/metadata_view.c
@@ -84,6 +84,7 @@ enum
 
   /* tags */
   md_tag_names,
+  md_categories,
 
   /* entries, do not touch! */
   md_size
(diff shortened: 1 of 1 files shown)


phweyland:
3.2 ? This is very far. The change is only alignment of visualization of the new tagging features...
I think this is coherent with what is delivered in 2.8.

phweyland:
I would say this is a bug. Imagine three "location" categories : creator location, agency location and customer location. Without this change, the three "locations" will appear in the tag list and you don't know what is each one.

TurboGit:
@phweyland : let me reconsider this but at this point we want to be very safe if we want the 3.0 release to be stable in time :)

TurboGit:
Ok, let's go with that. Thanks.

phweyland:
> we want the 3.0 release to be stable in time :)

Got it !

PR #3061, opened 2019-10-02
Ascii-art for "trash" or "warning" character

jenshannoschwalm:
As UTF-8 on current windows gtk packed with darktable doesn't allow those
characters to be displayed i took pascals obrys suggestion.


TurboGit:
Too bad, but not much we can do at the moment. Thanks.

PR #3067, opened 2019-10-03
History v3->v4 on-the-fly-conversion

jenshannoschwalm:
After all the mess with history v3 #3053 introduced history v4

That pr didn't fix images already developed with a current history version 3. See discussion there.

This pr prepares for an _on-the-fly_ history conversion.

Basically there is just one new function in iop_order.c  `void dt_ioppr_convert_onthefly(const int imgid)`. It checks for proper versioning (currently **only** v3->v4) itself.

It is called when we have to read the images history, done in `void dt_dev_read_history_ext` to do that conversion soon enough to leave the lighttable safe.

Please review and look for correct results.

After comments and testing there will be 2 more - hopefully short - review phases 
1. Prepare the database changing code without making it ready-to-shoot
2. Make it ready-to-go

@TurboGit @parafin go ahead

### DIFF
--- a/src/common/iop_order.c
+++ b/src/common/iop_order.c
@@ -35,6 +35,7 @@
 
 #define DT_IOP_ORDER_INFO FALSE	// used while debugging
 #define DT_ONTHEFLY_INFO FALSE   // while debugging on-the-fly conversion
+#define DT_ONTHEFLY_WRITING FALSE // If TRUE will do history update
 
 static void _ioppr_insert_iop_after(GList **_iop_order_list, GList *history_list, const char *op_new, const char *op_previous, const int dont_move);
 static void _ioppr_insert_iop_before(GList **_iop_order_list, GList *history_list, const char *op_new, const char *op_next, const int dont_move);
@@ -1507,7 +1508,6 @@ void dt_ioppr_convert_onthefly(const int imgid)
   }
 
   // process history
-  // read in the history
   for (int i=0;i<history_size;i++)
   {
     struct dt_onthefly_history_t *this = &myhistory[i];
@@ -1522,11 +1522,33 @@ void dt_ioppr_convert_onthefly(const int imgid)
     fprintf(stderr,"\n %3i %20s multi%3i :: iop %14.11f -> %14.11f",this->num,this->operation,this->multi_priority,this->old_iop_order,this->new_iop_order);
   }
 
-  // really write history, not yet
+  // really write history?, not yet
+  if (DT_ONTHEFLY_WRITING)
+  {
+    for (int i=0;i<history_size;i++)
+    {
+      struct dt_onthefly_history_t *this = &myhistory[i];
+      if (this->old_iop_order != this->new_iop_order)
+      {
+        DT_DEBUG_SQLITE3_PREPARE_V2(dt_database_get(darktable.db),
+          "UPDATE main.history SET iop_order = ?1 WHERE imgid = ?2 AND num = ?3", -1, &stmt, NULL);
+        DT_DEBUG_SQLITE3_BIND_DOUBLE(stmt, 1, this->new_iop_order);
+        DT_DEBUG_SQLITE3_BIND_INT(stmt, 2, imgid);
+        DT_DEBUG_SQLITE3_BIND_INT(stmt, 3, this->num);
+        sqlite3_step(stmt);
+        sqlite3_finalize(stmt);
+      }
+    }
 
+    DT_DEBUG_SQLITE3_PREPARE_V2(dt_database_get(darktable.db),
+      "UPDATE main.images SET iop_order_version = ?2 WHERE id = ?1", -1, &stmt, NULL);
+    DT_DEBUG_SQLITE3_BIND_INT(stmt, 1, imgid);
+    DT_DEBUG_SQLITE3_BIND_INT(stmt, 2, DT_IOP_ORDER_VERSION);
+    sqlite3_step(stmt);
+    sqlite3_finalize(stmt);
+  }
 
   free(myhistory);
-
 }
 
 int dt_ioppr_check_iop_order(dt_develop_t *dev, const int imgid, const char *msg)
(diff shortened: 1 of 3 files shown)


TurboGit:
To me this is going in the right direction. It was what I had in mind. I hope it will work as expected at the end. Please go ahead. Thanks.

src/develop/develop.c
```diff
@@ -1470,6 +1470,8 @@ void dt_dev_read_history_ext(dt_develop_t *dev, const int imgid, gboolean no_ima
 
   dev->iop_order_version = 0;
 
+  dt_ioppr_convert_onthefly(imgid);
```
TurboGit: Maybe this routine could return the iop_order_version for imgid and then we can remove the code below that is reading back iop_order_version from db?

jenshannoschwalm: Absolutely, i had it in mind but didn't do so until i know we are safe in the on-the-fly stuff and don't have hard ends.

jenshannoschwalm:
Ok, the database updating code is already present.

src/common/iop_order.c
```diff
@@ -1411,6 +1413,144 @@ static void _ioppr_check_rules(GList *iop_list, const int imgid, const char *msg
...
+  for (int i=0;i<history_size;i++)
+  {
+    struct dt_onthefly_history_t *this = &myhistory[i];
+    fprintf(stderr,"\n %3i %20s multi%3i :: iop %14.11f -> %14.11f",this->num,this->operation,this->multi_priority,this->old_iop_order,this->new_iop_order);
+  }
+
+  // really write history?, not yet
+  if (DT_ONTHEFLY_WRITING)
+  {
+    for (int i=0;i<history_size;i++)
```
TurboGit: I would wrap this loop inside:

```
sqlite3_exec(db->handle, "BEGIN TRANSACTION", NULL, NULL, NULL);
if (...)
{
}
sqlite3_exec(db->handle, "COMMIT", NULL, NULL, NULL);
```

To ensure that the whole image is updated or nothing is done in case of crash.

src/common/iop_order.c
```diff
@@ -1411,6 +1413,144 @@ static void _ioppr_check_rules(GList *iop_list, const int imgid, const char *msg
...
+
+  // get the number of known iops
+  int valid_iops = 0;
+
+  GList *iops_order = g_list_last(current_iop_list);
+  while(iops_order)
+  {
+    valid_iops++; 
+    iops_order = g_list_previous(iops_order);
+  }
```
TurboGit: Equivalent to:
`const int valid_iops = g_list_length (current_iop_list);`

src/common/iop_order.c
jenshannoschwalm: How about the main.images /iop_order_version change? Can/should that also be included before sqlite3_exec?

TurboGit: main.images update should be part of the same transaction.

src/common/iop_order.c
jenshannoschwalm: Thanks :-)

src/common/iop_order.c
jenshannoschwalm: done

src/develop/develop.c
jenshannoschwalm: dt_ioppr_convert_onthefly now returns the iop_order_version from db ...

jenshannoschwalm:
ok, added/changed all suggestions; anyone else? @parafin ?

src/develop/develop.c
TurboGit: So let's use the return value here now. And remove code below.

src/common/iop_order.c
```diff
@@ -34,6 +34,8 @@
 #define DT_IOP_ORDER_VERSION 4
 
 #define DT_IOP_ORDER_INFO FALSE	// used while debugging
+#define DT_ONTHEFLY_INFO FALSE   // while debugging on-the-fly conversion
+#define DT_ONTHEFLY_WRITING FALSE // If TRUE will do history update
```
TurboGit: And of course we need now to activate it. No?

TurboGit:
I just found out that the code is also called when in the lighttable and a refresh of the thumb is in action. Not surprising. Now the question, is that ok? I would say so as we are the stage of avoiding crash of buggy v3 iop-order.

Apart from that it worked in 5 images I did test.

jenshannoschwalm:
> I just found out that the code is also called when in the lighttable and a refresh of the thumb is in action. Not surprising. Now the question, is that ok? I would say so as we are the stage of avoiding crash of buggy v3 iop-order.

That was intended :-) There were some issues about that.

src/common/iop_order.c
jenshannoschwalm: done

src/common/iop_order.c
```diff
@@ -1411,6 +1413,152 @@ static void _ioppr_check_rules(GList *iop_list, const int imgid, const char *msg
...
+  }
+
+  // process history
+  for (int i=0;i<history_size;i++)
+  {
+    struct dt_onthefly_history_t *this = &myhistory[i];
+    this->new_iop_order = dt_ioppr_get_iop_order(current_iop_list, this->operation) + (double)this->multi_priority / 100.0;
+  }
+
+  // process some more checks possibly; any sort data that can't be correct?
```
jenshannoschwalm: Anthing that might be wanted here?

jenshannoschwalm:
Ok, i also tested on what i had of v3 images. All problems gone here.
I activated writing history now.

jenshannoschwalm:
@TurboGit maybe there is one thing we missed.

If the history has been changed on-the-fly shouldn't we raise a signal?
`dt_control_signal_raise(darktable.signals, DT_SIGNAL_DEVELOP_HISTORY_CHANGE);
`

TurboGit:
@jenshannoschwalm : I don't think so. We are changing the database BEFORE even reading the history. So in fact all this is happening being the scene and a signal does not seems needed to me. But I may be wrong.

TurboGit:
Just want to add that a signal is used to refresh part of the GUI when some information are changed. Again here the data changed on the DB is not yet on the GUI.

Issue #3034, opened 2019-09-28
Garbled text color picker live sample


alcomposer:
**DT Version:** Git Master

**Issue:**

Load an image- enter darkroom, add a color picker live sample, values display as garbled text. Am I missing a font? Also the column extends past its area- and looks cut off.

![issue_garbled_text_color_picker](https://user-images.githubusercontent.com/12004932/65811387-def03780-e1f6-11e9-92a4-03f81b28f499.png)

Even worse in Lab view:
![lab_view](https://user-images.githubusercontent.com/12004932/65811404-0941f500-e1f7-11e9-89d5-6e6f04307f1d.png)

Nilvus:
This text use a specific font : courier. It's a quite very standard font but it's possible you don't have this one on your OS. I have the font and the text displayed correctly.

@TurboGit : what do you think about that ? Does that need to add this font needed on release notes ?

alcomposer:
Roboto comes in monospace as well. 

If if user has Roboto, then they also have monospace.

aurelienpierre:
monospace makes a lot more sense indeed.

Nilvus:
Indeed, that's better.

alcomposer:
This may be a bit bike-shedding, *but*...

Is there a particular reason why the live-sample-data is bold? Would not regular monospace be fine? 

Also- why are these values monospace (as in courier) yet other number values in DT are not?
![live_sample](https://user-images.githubusercontent.com/12004932/66094025-9effca80-e5d1-11e9-9059-a6520e9fc990.png)

alcomposer:
I uninstalled the bad courier font that was in my Arch Linux system (apparently an Adobe one that for some reason installs - but is broken).

It works now- but not using Roboto-Mono. Either way it was my system fault.

But may still be good to specify a font family that is related to the main font in DT.

PR #3068, opened 2019-10-03
fix GUI glitches

### LINKED ISSUE #3057: Darktable themes other than the "darktable" one mess up the fonts
**Describe the bug**
image

The fonts in history and the parameters inside module look horrible.

aurelienpierre:
Use both fonts subsets names and general font names with corresponding stretch and weight properties to help OS to find the correct font family.

### DIFF
--- a/data/themes/darktable-elegant-dark.css
+++ b/data/themes/darktable-elegant-dark.css
@@ -72,10 +72,6 @@
 @define-color field_hover_bg @grey_60;
 @define-color field_hover_fg @grey_20;
 
-/* Tooltips and contextual helpers */
-@define-color tooltip_bg_color @grey_20;
-@define-color tooltip_fg_color @grey_85;
-@define-color really_dark_bg_color @grey_20;
 
 /* Views */
 @define-color darkroom_bg_color @grey_60;
--- a/data/themes/darktable-elegant-darker.css
+++ b/data/themes/darktable-elegant-darker.css
@@ -23,14 +23,7 @@
 /* hide selection frame in culling */
 @define-color culling_selected_border_color transparent;
 
-*
-{
-    font-weight: 200;
-    font-family: "Roboto Light", "Segoe UI Light",
-                 "SF Pro Display Light", "Ubuntu Light",
-                 "Cantarell", sans-serif;
-}
-
+*,
 #lib-plugin-ui,
 #iop-plugin-ui,
 #lib-modulelist,
@@ -79,40 +72,44 @@ eventbox *,
 box,
 box *
 {
-    font-family: "Roboto Light", "Segoe UI Light",
-                 "SF Pro Display Light", "Ubuntu Light",
-                 "Cantarell", sans-serif;
+    font-family: "Roboto Light", "Roboto",                  /* best case scenario */
+                 "Segoe UI Light", "Segoe UI",              /* Windows 10 default */
+                 "SF Pro Display Light", "SF Pro Display",  /* Mac OS X default */
+                 "Ubuntu Light", "Ubuntu",                  /* Ubuntu default */
+                 "Cantarell",                               /* Gnome default */ 
+                 sans-serif;                                /* default default */
 }
 
-button
+button,
+#section_label
 {
-    font-weight: 500;
-    font-family: "Roboto Medium", "Segoe UI Semibold",
-                 "SF Pro Display Medium",
-                 "Ubuntu Medium", "Cantarell", sans-serif;
-
+    font-family: "Roboto Medium", "Roboto", 
+                 "Segoe UI Semibold", "Segoe UI",
+                 "SF Pro Display Medium", "SF Pro Display",
+                 "Ubuntu Medium", "Ubuntu",
+                 "Cantarell", 
+                  sans-serif;
 }
 
 #iop-panel-label,
 #lib-panel-label
 {
-    font-family: "Roboto Condensed","Segoe UI Condensed",
-                 "SF Pro Display", "Ubuntu Condensed",
-                 "Cantarell", sans-serif;
-}
-
-#section_label
-{
-    font-family: "Roboto Medium", "Segoe UI Semibold",
-                 "SF Pro Display Medium",
-                 "Ubuntu Medium", "Cantarell", sans-serif;
-    font-weight: 600;
+    font-family: "Roboto Condensed", "Roboto",
+                 "Segoe UI Condensed", "Segoe UI",
+                 "SF Pro Display", "SF Pro Display",
+                 "Ubuntu Condensed", "Ubuntu",
+                 "Cantarell", 
+                 sans-serif;
 }
 
 notebook tabs,
 #modules-tabs,
 #blending-tabs
 {
-    font-family: "Roboto", "Segoe UI", "SF Pro Text", "Ubuntu",
-                 "Cantarell", sans-serif;
+    font-family: "Roboto", 
+                 "Segoe UI", 
+                 "SF Pro Text", 
+                 "Ubuntu",
+                 "Cantarell", 
+                  sans-serif;
 }
--- a/data/themes/darktable-elegant-grey.css
+++ b/data/themes/darktable-elegant-grey.css
@@ -74,11 +74,6 @@
 @define-color field_hover_bg @grey_75;
 @define-color field_hover_fg @grey_35;
 
-/* Tooltips and contextual helpers */
-@define-color tooltip_bg_color @grey_35;
-@define-color tooltip_fg_color @grey_95;
-@define-color really_dark_bg_color @grey_35;
-
 /* Views */
 @define-color darkroom_bg_color @grey_75;
 @define-color darkroom_preview_bg_color @grey_65;
--- a/data/themes/darktable.css
+++ b/data/themes/darktable.css
@@ -465,7 +465,6 @@ entry selection
   border:0;
   padding: 0 2pt 2pt 0.5em;
   font-weight: normal;
-  font-stretch: condensed;
   font-family: sans-serif;
   font-size: larger;
 }
--- a/data/themes/darktable.css
+++ b/data/themes/darktable.css
@@ -767,11 +767,13 @@ notebook tab
 {
   border-bottom: 2px solid transparent;
   padding: 3px;
+  font-weight: normal;
 }
 
 notebook tab label /* disabled tabs */
 {
   color: shade(@button_checked_fg, 0.75);
+  font-weight: normal;
 }
 
 /* This will avoid have too much height on entry items in dialog windows; label and button are needed as they are related */
@@ -1189,7 +1191,7 @@ notebook tab:hover
   background-color: @button_hover_bg;
 }
 
-notebook tab:hover label
+notebook tab:hover labelnotebook
 {
   color: @button_hover_fg;
 }
--- a/src/bauhaus/bauhaus.c
+++ b/src/bauhaus/bauhaus.c
@@ -529,7 +529,7 @@ void dt_bauhaus_load_theme()
   cairo_surface_t *cst = cairo_image_surface_create(CAIRO_FORMAT_ARGB32, 128, 128);
   cairo_t *cr = cairo_create(cst);
   layout = pango_cairo_create_layout(cr);
-  pango_layout_set_text(layout, "X", -1);
+  pango_layout_set_text(layout, "m", -1);
   pango_layout_set_font_description(layout, darktable.bauhaus->pango_font_desc);
   pango_cairo_context_set_resolution(pango_layout_get_context(layout), darktable.gui->dpi);
   int pango_width, pango_height;
@@ -538,7 +538,7 @@ void dt_bauhaus_load_theme()
   cairo_destroy(cr);
   cairo_surface_destroy(cst);
 
-  darktable.bauhaus->scale = 1.3f;
+  darktable.bauhaus->scale = 1.33f;
   darktable.bauhaus->line_height = pango_height / PANGO_SCALE;
   darktable.bauhaus->widget_space = INNER_PADDING / 2.0f; // used as a top/bottom margin for widgets
   darktable.bauhaus->quad_width = darktable.bauhaus->line_height;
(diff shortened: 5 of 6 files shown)


chhil:
I used your darktable-elegant-darker.css and updated mine with it in C:\Program Files\darktable\share\darktable\themes\darktable-elegant-darker.css

Did not see any change.

phweyland:
With the normal darktable-elegant-darker.css (seems to fall back on Segoe UI):

image

With yours (which picks up Roboto):

image

The fact is that one month ago, on another windows computer I was happy with the elegant one:
https://github.com/darktable-org/darktable/pull/2889#issuecomment-529115274

Since then it has completely changed. Can be some settings on my computer or something in darktable.

aurelienpierre:
hmm in both cases @phweyland you have issues with anti-aliasing in RGB (colour fringes around letters) and kerning adaptation (the distance between e-x-p is not supposed to look like that) which looks to me like bad Clear Type setting. Fonts have not been change for several months AFAIK.

Nilvus:
@aurelienpierre : why not also improve text on #live-sample-data tag and fix #3034 issue (even if it's closed, I'm quite agree with him). Having another font than courier would be more consistent. Maybe a monospace font on darktable.css and Roboto monospace one for elegant themes. And seeing if bold is really needed would be better. I rarely use color picker module but the goal is to make darktable 3.0 the best release ever. Details are one thing to show that !

phweyland:
With Clear Type deactivated.
The font is still too small (this explains at least partially the weird effect of anti-aliasing). 
The letter a is 8 pixels high on module name (good), but 6 pixels high on the other texts (too small). On the good sample one month earlier the height was 7 pixels (better).


image

phweyland:
Here is a 100% screenshot of sans-serif, Segoe UI and Roboto under windows without clear type nor font edge smoothing. 

Size 9 seems to be the one used for the normal text font (the one I see on dt).

sans-serif - roboto PNG

Follow some snapshot of GTKInspector saying that module name size should be 14.999 and normal size 12px. Where is truth ?

GTKInspector button label

GTKInspector module name

Finally a dt snapshot

image

phweyland:
I've just built dt with the commit on 25th August. Same issue.
So its most likely an issue with my windows settings or some library in build chain ...
Do all windows users experience this small font size issue ?

aurelienpierre:
It's a problem I have had on both Windows 10 and Gnome/KDE Linux. Each time, the culprit was the font hinting and anti-aliasing performed by the desktop/OS, and you have to tweak it until you are happy.

This is what I get on Fedora KDE with hinting set to full and RGB anti-aliasing on:
Screenshot_20191004_014034

And with both disabled:
Screenshot_20191004_014316


> Follow some snapshot of GTKInspector saying that module name size should be 14.999 and normal size 12px. Where is truth ?

> Size 9 seems to be the one used for the normal text font (the one I see on dt).

Font size are traditionally measured in points, not pixels, and 9 pt is equivalent to 12 px.

phweyland:
> 9 pt is equivalent to 12 px.

OK, so GTKInspector and my screen are aligned.
Remains that the normal text is too small, whatever the anti-aliasing.

Your snapshots look very good.
For module name / normal text, my screen shows 8 / 6 pixels height. Yours, 17 / 17.

Why such a difference ? 
My screen size is 1600 x 1200. I haven't changed the dt dpi setting. 

Other strange point on your snapshot. The ratio is 1. Should be something around 17% smaller for the normal text (if we have the same css).

aurelienpierre:
> Your snapshots look very good.
For module name / normal text, my screen shows 8 / 6 pixels height. Yours, 17 / 17 (edit : lower case height)

The module names are supposed to use a condensed font, so the size is a bit different. Also, my system use a DPI scaling of 200%.

> Other strange point on your snapshot. The ratio is 1. Should be something around 17% smaller for the normal text (if we have the same css).

Not sure I understand.

phweyland:
> > Other strange point on your snapshot. The ratio is 1. Should be something around 17% smaller for the normal text (if we have the same css).
> 
> Not sure I understand.

```
/* Labels in modules */
#iop-panel-label,
#lib-panel-label
{
  background-color: @bg_color;
  color: @plugin_label_color;
  border:0;
  padding: 0 2pt 2pt 0.5em;
  font-weight: normal;
  font-stretch: condensed;
  font-family: sans-serif;
  font-size: larger;
}
```

 `font-size: larger; ` should make the module name 15-20 bigger than normal text. On your screenshot I don't see any size difference between the module name and the normal text lower case size.

> my system use a DPI scaling of 200%

OK, so I understand that a font-size which is perfect on your screen can be too small on smaller screen.
Is it a dt setting or your OS/desktop which does that ?

phweyland:
If I change font-stretch in your css (this PR) to normal it finds the Roboto Condensed.
 
```
#iop-panel-label,
#lib-panel-label
{
    font-family: "Roboto Condensed", "Roboto",
                 "Segoe UI Condensed", "Segoe UI",
                 "SF Pro Display", "SF Pro Display",
                 "Ubuntu Condensed", "Ubuntu",
                 "Cantarell", 
                 sans-serif;
    font-weight: 300;
    font-stretch: normal;
}

If I set screen DPI to 150 (instead of 94 which is default one if I'm not mistaken) I get this:

![image](https://user-images.githubusercontent.com/23012047/66199383-c53f6c00-e69e-11e9-829f-8febb2968eb6.png)

Which has very good definition but uses too much space.



```
image

phweyland:
The css settings to get Roboto and Roboto Condensed on windows are a good point for me.

I have not found any window setting improving really the appearance (and of course would impact the other apps).

But setting dt screen dpi to 100 improves a bit the appearance. lower case height jumps to 7 pixels.

image

@chhil , could check if tweaking `screen_dpi_overwrite` in darktablerc you can improve your display ?

phweyland:
Removing Roboto and checking Segoe UI, Segoe UI is more readable than Segoe UI Light (still with 100 DPI).

Segoe UI Light:
image

Segoe UI:
image

chhil:
> @chhil , could check if tweaking screen_dpi_overwrite in darktablerc you can improve your display ?

@phweyland , I will get to this some time tomorrow and report back.

chhil:
@phweyland 
The screen dpi to 100 didn't do anything for me or at least to my naked eye.

I tried reducing the choices in the font family to Roboto only and got better results than when a number are provided and picking happens based on priority, thus forcing it to use Roboto as I do have that on my system. The results were a tad bit better from a viewing perspective when elegant was chosen compared to the others.

[darktable-elegant.css.txt](https://github.com/darktable-org/darktable/files/3692854/darktable-elegant.css.txt)

phweyland:
>> The screen dpi to 100 didn't do anything for me or at least to my naked eye

100 was the value enough for my screen 1920x1200 to enlarge a lower case height from 6 to 7 pixels.
your screen has a higher definition than mine. So you could start with 200 dpi as Aurelien and decrease the Dpi to find the minimum one which gives enough definition for your visual comfort.

chhil:
My screen res is 2560x1080.

A dpi of 120 gave me the comfort I wanted with fonts looking decent (theme elegant).
Beyond that the preference dialog was extending to a point where I could not see the close button.

phweyland:
> A dpi of 120 gave me the comfort I wanted with fonts looking decent (theme elegant).

So what I understand so far.
Adding Roboto after Roboto Light make Roboto work on Windows OS.
But font-weight = "Condensed" prevents the OS find "Roboto Condensed". (font-weight = "Normal" seems to solve that).
Remains the too small font size for some screen resolutions. I haven't understood everything but it seems dt forces the line-height and then adjusts the font-size inside the remaining space, turning lower case of Roboto or Segoe UI font types smaller than desired (due to smaller relative height of lower case of these font faces). A workaround for the time being is to tweak the DPI used by dt pre-calculations. But that's probably not the solution.

It is worth to note that DPI setting seems not to affect the graphical part of lighttable or darkroom.

aurelienpierre:
>  Adding Roboto after Roboto Light make Roboto work on Windows OS.

nice.

> But font-weight = "Condensed" prevents the OS find "Roboto Condensed". (font-weight = "Normal" seems to solve that).

`condensed` is an option of `font-stretch` property, `font-weight` sets the grease (bold/semi-bold/roman/display/light). In strict CSS 3 standard, Roboto Condensed should be registered as a sub-family of Roboto and calling Roboto with `font-stretch: condensed` or Roboto Condensed should be equivalent. I'm guessing either Windows or GTK are messing CSS 3…

> Remains the too small font size for some screen resolutions. I haven't understood everything but it seems dt forces the line-height and then adjusts the font-size inside the remaining space, 

It's actually the opposite. In `bauhaus.c`, I have forced the font size (namely, the height of an X) to be the reference size for the line height, but it's possible that both sizes have a sweet spot depending of the DPI scaling, and since all labels are lower-case, it might make more sense to use an x as the size reference.

phweyland:
>> it might make more sense to use an x as the size reference

Sounds good for me.

phweyland:
>> Roboto Condensed should be registered as a sub-family of Roboto

On Windows Roboto and Roboto Condensed appear as 2.font families....

aurelienpierre:
I have removed the explicit `font-weight` and `font-stretch` declarations, I believe it will improve things a bit.

phweyland:
> I have removed the explicit `font-weight` and `font-stretch` declarations, I believe it will improve things a bit.

I've made a try on this. The control "font-stretch: condensed;" in darktable.css remains valid if you don't override it with "font-stretch: normal;".  So I get:

```
(darktable.exe:11312): Pango-WARNING **: 21:56:53.670: couldn't load font "Roboto Condensed,Roboto,Segoe UI Condensed,Segoe UI,SF Pro Display,SF Pro Display,Ubuntu Condensed,Ubuntu,Cantarell,sans-serif Condensed Not-Rotated 10.7998046875", falling back to "Sans Condensed Not-Rotated 10.7998046875", expect ugly output.

(darktable.exe:11312): Pango-WARNING **: 21:56:53.670: couldn't load font "Sans Condensed Not-Rotated 10.7998046875", falling back to "Sans Not-Rotated 10.7998046875", expect ugly output.
```

To get Roboto Condensed on module name I have either, to comment "font-stretch: condensed;" in darktable.css, or to add "font-stretch: normal;" in your file.

aurelienpierre:
@phweyland I believe everything should be in order now.

phweyland:
> I believe everything should be in order now.

To test it I should fetch your branch, right ?

aurelienpierre:
yes !

phweyland:
with roboto (no console message):
image

Still the weird bold when in focus:
image

Without roboto (should be Segoe UI)

image
with these messages:
```
Operation not supported
_cairo_win32_scaled_font_set_metrics:GetTextMetrics: The system could not find the environment option that was entered.
(darktable.exe:3548): Pango-WARNING **: 20:46:18.191: failed to create cairo scaled font, expect ugly output. the offending font is 'Roboto Condensed, Light 10.798828125'
(darktable.exe:3548): Pango-WARNING **: 20:46:18.191: font_face status is: error occurred in the Windows Graphics Device Interface
(darktable.exe:3548): Pango-WARNING **: 20:46:18.191: scaled_font status is: error occurred in the Windows Graphics Device Interface
```

With the master 2.7.0+2055~g36f87d176 I have this:
image
with these messages:
```
(darktable.exe:15732): Pango-WARNING **: 20:50:32.598: couldn't load font "Roboto Condensed,Segoe UI Condensed,SF Pro Display,Ubuntu Condensed,Cantarell,sans-serif Condensed Not-Rotated 10.7998046875", falling back to "Sans Condensed Not-Rotated 10.7998046875", expect ugly output.
(darktable.exe:15732): Pango-WARNING **: 20:50:32.598: couldn't load font "Sans Condensed Not-Rotated 10.7998046875", falling back to "Sans Not-Rotated 10.7998046875", expect ugly output
```

TurboGit:
Isn't  this the Gtk bug which was not supporting multiple fonts? I kind of remember such discussion. Maybe the Windows build does not use the proper Gtk library with the fix?

phweyland:
> Isn't this the Gtk bug which was not supporting multiple fonts?

I don't think so, because without roboto, it uses Segoe UI (correctly with the current master), with an issue for condensed with the Aurelien changes.

With the current master I prefer Segoe UI, but with Aurelien changes Roboto seems to behave better.

aurelienpierre:
> Isn't this the Gtk bug which was not supporting multiple fonts? I kind of remember such discussion. Maybe the Windows build does not use the proper Gtk library with the fix?

@TurboGit not exactly. It seems the font variants (thin, display, condensed, etc.) don't get associated with matching CSS rules (`font-weight`and `font-stretch` values). So if you are using for example the font-family `Roboto Condensed Display` with rules `font-weight: 300` and `font-stretch: condensed` (which mean theoriticaly the same thing), GTK will take the font family drawing but try to re-scale it automatically to meet the specified weight and stretch, while it is already properly scaled in the OTF file. All in all, it seems that OTF font sub-families are treated as independent families, not subsets of a larger family (at least on Windows). The mistake I made was to over-constraint the CSS instructions trying to be more general.

parafin:
I'm getting the following warning with 3.0.0rc0 on macOS:
(darktable:1553): Pango-WARNING **: 10:30:50.357: couldn't load font "sans-serif Condensed Not-Rotated 14.400390625", modified variant/weight/stretch as fallback, expect ugly output.

aurelienpierre:
> I'm getting the following warning with 3.0.0rc0 on macOS:
> (darktable:1553): Pango-WARNING **: 10:30:50.357: couldn't load font "sans-serif Condensed Not-Rotated 14.400390625", modified variant/weight/stretch as fallback, expect ugly output.

Yeah, that should be fixed with this PR.

@TurboGit could you consider merging that ?

TurboGit:
Ok, merging! Thanks.

data/themes/darktable-elegant-dark.css
```diff
@@ -72,10 +72,6 @@
 @define-color field_hover_bg @grey_60;
 @define-color field_hover_fg @grey_20;
 
-/* Tooltips and contextual helpers */
```
TurboGit: I'm not sure it was a good idea to remove this after all. The tooltip black backgroud for a grey or dark theme is a bit disturbing I found. What was the rational for this change?

Nilvus: That's why I put it back on grey theme. I could also see that on my #3362 PR if you want. I will note that also for dark theme.

aurelienpierre: right click on a slider with this change and you will see that the cursor border (triangle) has a different colour than the popup. This is because I haven't been able to dynamically change the sliders border color. I agree it's not consistent but it's less ugly until I find a way to change cursor border color with CSS rules.

PR #3083, opened 2019-10-06
Improve tools menu consistency

amousset:
Use the same height for buttons.

Before:
dt-gauche
dt-droite

After:
dt5g
dt5d


TurboGit:
Nice, this has been bothering me for some time. Thanks!

Issue #3077, opened 2019-10-05
[LUT 3D] Dropdown in module doesn't show 3d luts even after setting preferences and restarting.


chhil:
**Describe the bug**
New Install, first use, luts don't get displayed in the lut dropdown after preference has been updated to point to the lut folder and app restarted.

**To Reproduce**

1. Clean install first use clean db.
2. Preferences select LUT folder.
3. Import images [no xmp file].
4. Select image.
5. Enable lut module.
6. Drop down for luts does not show anything in the module.
7. Select folder icon on the left of the drop down.
8. In the selection folder dialog that opens, change filter from haldcut png to cube. I don't have any haldclut pngs, only cube files
9. Now the drop down for luts is populated.

**Expected behavior**
Png/cube  or any supported extensions should be shown in the lut modules drop down that shows the luts once the preference is pointed to a lut folder. The folder selection seems redundant to populate the dropdown.

OS Win 10
Version 2.7.0.1945

phweyland:
> Png/cube or any supported extensions should be shown in the lut modules drop down that shows the luts once the preference is pointed to a lut folder

Agreed. I'll try to simplify this/

phweyland:
Hi @chhil, the pr #3084 should have solved that. Do you still have the issue with the current master ?

gadolf66:
I concur with @chhil, once you select the right folder, the dropdown menu shouldn't show the full path, because it gets truncated and the lut files them selves don't show up, specially when there's a bunch of files in the folder, as in

![image](https://user-images.githubusercontent.com/46730249/97779815-6f6a1b80-1b5f-11eb-9c26-c2e1ef2e0db7.png)

phweyland:
Agreed this is not good.
With the current master is what I have:
![image](https://user-images.githubusercontent.com/23012047/97783527-232ad580-1b77-11eb-83a1-60d7d1cd980b.png)

... Should be solved since #5925.

PR #3084, opened 2019-10-06
lut3d - show all valid lut files in selector

phweyland:
the selector shows now png and cube files in a same folder.
However the user cannot see only png files or only cube files at a time. Is that acceptable ?

to try to answer issue #3077


TurboGit:
For 3.1 now.

TurboGit:
Merged.

PR #3092, opened 2019-10-08
fix exception when no tag on image

phweyland:
sorry, last change has revealed an old bug.

### DIFF
--- a/src/libs/metadata_view.c
+++ b/src/libs/metadata_view.c
@@ -622,7 +622,6 @@ static void _metadata_view_update_values(dt_lib_module_t *self)
         }
       }
       if(tagstring) tagstring[strlen(tagstring)-2] = '\0';
-      _metadata_update_value(d->metadata[md_tag_names], tagstring);
     }
     _metadata_update_value(d->metadata[md_tag_names], tagstring ? tagstring : NODATA_STRING);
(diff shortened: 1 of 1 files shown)


src/libs/metadata_view.c
```diff
@@ -622,7 +622,6 @@ static void _metadata_view_update_values(dt_lib_module_t *self)
         }
       }
       if(tagstring) tagstring[strlen(tagstring)-2] = '\0';
-      _metadata_update_value(d->metadata[md_tag_names], tagstring);
```
TurboGit: I'm not sure to see the fix. The line removed above is exactly the one that will be called below.

`_metadata_update_value(d->metadata[md_tag_names], tagstring ? tagstring : NODATA_STRING);`

As tagstring is set (not NULL), then the code will be:

`_metadata_update_value(d->metadata[md_tag_names], tagstring );`

Which is the line removed above. So we remove a duplicate call, but I don't understand what is the fix?

Am I missing something?

hellbuster: @TurboGit I think that is the problem: the removed line doesn't handle the NULL case

TurboGit: Indeed, so obvious... I should sleep more :)

Thanks.

Issue #3098, opened 2019-10-08
How to remove deprecated modules from favourites


leclercj:
**Describe the bug**
Now that some iop modules are deprecated, I'd like to remove them from my favourites and replace them with new ones. I can't do so because the deprecated modules are not listed anymore in the 'more modules' section.

**To Reproduce**
1. In darktable 2.6, elect 'filmic' (as an exemple) as favorite module in darkroom
2. Upgrade to darktable 2.7
3. 'filmic' still present as favorite module, but not proposed anymore in the 'more module' section as replaced by 'filmic rgb'. No way to remove filmic from favourites.

**Proposed behavior**
'more modules' should continue to list the deprecated modules (at least as an option).
This to solve this use case, but also to give time to users to change their habits

**Screenshot**
![Capture d’écran du 2019-10-08 19-22-28](https://user-images.githubusercontent.com/22872204/66417781-06b67b00-ea01-11e9-9e6a-82abe88314e3.png)

**Platform (please complete the following information):**
 - Version 2.7.0+1956~g552a

jenshannoschwalm:
Shouldn't deprecated modules be always invisible except in history?

TurboGit:
Can't you remove by clicking on the preset menu and select "favourite" ?

TurboGit:
@jenshannoschwalm : yes it should, looks like favourite flag gets precedence over deprecated one.

jenshannoschwalm:
Right, will take care of this.

TurboGit:
Thanks !

leclercj:
> Can't you remove by clicking on the preset menu and select "favourite" ?

Yes it works. I didn't noticed this possibility. This was not intuitive to me. Thanks.
However, I still believe users need time to change their habits. What about having an option to keep the deprecated modules in 'more modules' section ? Of course, with a clear indication that they are deprecated

TurboGit:
This is now fixed.

I'm not fond of the idea to have an option to keep the deprecated modules. This means that people will keep using them and if they are deprecated it is because there is easier/faster/safer way to achieve the same.

PR #3099, opened 2019-10-08
Hide deprecated modules in favourite modules menu.

jenshannoschwalm:
#3098 re-opened discussion about visibility of deprecated modules except in history.
As these modules are hidden in all other tabs they should be hidden here too for consistency.


TurboGit:
Thanks!

Issue #3081, opened 2019-10-06
Denoise profiled gets very slow in darkroom when using toneequalizer


rawfiner:
**Describe the bug**
At 100% zoom level, toneequalizer parameters have a huge impact on denoise profiled processing time: the larger the smoothing diameter in toneequalizer, the slower denoise profiled gets.
This is quite easy to explain: toneeequalizer requests a larger roi as smoothing diameter increases, which requires denoise profiled to denoise a larger image for the preview.
This gives a big issue in terms of execution time: with smoothing diameter at its maximal shown value (45%), denoise profiled takes on my computer about 17 seconds to denoise the image (viewed at 100% zoom level), instead of about 1.2 seconds without toneequalizer.

**To Reproduce**
0. Launch dt with -d perf option to see the execution times
1. Go to darkroom
2. Zoom at 100%
3. Activate denoiseprofiled
4. Activate toneequalizer with a big smoothing diameter (the default value is already big enough to see big effects)

**Expected behavior**
The preview should be much faster, and time of preview should be kept bellow 3-4 seconds.

**Platform (please complete the following information):**
 - OS: fedora 30
 - Version master

**Questions related to this problem**
@aurelienpierre , if we force modify_roi_in in toneequalizer to keep the roi smaller than the one that is really needed (if the roi is very big) would that be very problematic for the preview? For instance, we could ensure the number of requested pixels is multiplied by 2 maximum. What do you think?

aurelienpierre:
Ok, documentation time… :grin: 

The smoothing diameter in toneeq is the width of the window for the box blur/box average (which is the core of the guided filter), in % of the largest image dimension. If using multiple iterations of guided filter, the window width is multiplied by 2 between each iterations to avoid singularities.

As you know, a box blur is a convolution, and every convolution is sensitive to boundary conditions. In this implementation, we assume a constant boundary condition, meaning the last pixel on each border of the image is counted as many times as needed to fill the convolution window around borders.

That triggers a problem if your ROI is not the full image, meaning you get "artificial" borders inside the picture when zooming-in, because the guided filter will not look the same zoomed-in or at full-size around the borders of the ROI. It's especially bad since the GUI mask is computed on the current ROI (zoomed and panned) but the buffered mask for interaction is computed on the full-size thumbnail, so they don't even match.

[...]

PR #3101, opened 2019-10-09
toneeq: add tooltip to warn user about preview slowdown

rawfiner:
See issue #3081


TurboGit:
Works for me. Thanks.

PR #3110, opened 2019-10-10
denoiseprofile: ensure pixel values are >= 0 before backtransform

rawfiner:
In some cases, in wavelets mode, some pixels may appear white after
backtransform instead of black if they are < 0:
before:
Capture d’écran_2019-10-10_07-59-34
after:
Capture d’écran_2019-10-10_08-02-17


TurboGit:
Looks good to me. Thanks.

PR #3113, opened 2019-10-10
BUG: toneeq/fast guided filter

aurelienpierre:
Fix a bug in the guided filter where the guided and guiding images were inverted in function arguments, so the quantization parameter had to be compensated with exposure, which was not intuitive. This will slightly break previous edits using non-zero quantization, but will be easily corrected by adjusting the mask exposure compensation.


TurboGit:
Thanks!

PR #3125, opened 2019-10-12
threadsafe rowids in history.

jenshannoschwalm:
I sometimes observed non-plausible history for images in situations where multiple threads were involved (importing lots of images or updating thumbnails) and i started development,  added styles or compressed history while these threads were still running.

The history database - as it is used - requires nums to be sorted by rowid
in certain cases. (See comment in develop.c L.~1365)

Although we use sqlite3 in serialized mode this does not absolutely ensure
database lines are written in a definite rowid order.

At lease these need &darktable.db_insert protection
`static void _dev_merge_history(dt_develop_t *dev, const int imgid)`
`static void _history_reorder(int32_t imgid)`

### DIFF
--- a/src/common/history.c
+++ b/src/common/history.c
@@ -1052,6 +1052,9 @@ static void _history_reorder(int32_t imgid)
   else
   {
     if (give_reorder_information) fprintf(stderr,", reorder\n");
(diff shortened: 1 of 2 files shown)


TurboGit:
Looks indeed a good thing to protect those. Will merge. Thanks.

Issue #3119, opened 2019-10-10
No EXIF in files exported from darkroom mode


tomaszg7:
**Describe the bug**
I export my photos to 16-bit TIFF from darkroom mode. In current GIT HEAD (fa2d1c204ab73a4c0d2d4600d66531c29cb8ed38), exported files are missing EXIF data. Surprisingly, it happens only if I export from darkroom view (via ^E). If I do it from lighttable view, exif shows up. As far as I could tell all file formats are affected.

Exif metadata doesn't show either with `exiftool` or `exiv2`.

**To Reproduce**
1. Enter darkroom mode
1. Export photo (e.g. to 16-bit TIFF) via ^E.
1. Check for EXIF.

**Platform (please complete the following information):**
 - OS: Gentoo Linux
 - exiv2-0.27.2

Attaching log from `-d all`:
[dt.log.gz](https://github.com/darktable-org/darktable/files/3714979/dt.log.gz)

phrrk:
I have observed similar issue when exporting from lightable mode, when using a storage type I have written as a lua script.  This script uses phototeleport (a windows application) as an uploader to Flickr.  exiftool reports that none of the exif data stored in the .xmp file are incorporated into the exported file.  However image description and GPS tagging do seem to make it through to the exported file, but title and darktable processing history do not (I have checked export settings).  All exif data are included if I simply export to file on disk.

phweyland:
ooops, I've not tested the export (crtl-e) from darkroom indeed. I'll look at this asap.

> when using a storage type I have written as a lua script
@phrrk, your case in not clear to me. I don't know well what can be done with lua for export. Could you explain a bit more how that works ? Maybe share the script ?

TurboGit:
@phweyland : thanks!

phweyland:
> @phweyland : thanks!

I may not have too much time the next 2 or 3 days. Do you have any hard deadline for that ?

TurboGit:
No, I would say that before end of month is ok.

phrrk:
@phweyland, I attach my lua script which is based on the gimp one.  Essentially all it does is export an image, or images, into a temporary storage location and then passes this image, or images, to the external application (phototeleporter) in this case.  Once the uploads are complete the exported files are deleted from the temporary storage.  I haven't checked but I would assume that the same issue applies with the gimp.lua script that comes with darktable as my script is heavily based on that one.

[...]

Issue #3114, opened 2019-10-10
darktable-cli Segmentation fault


ClaesGitHub:
**Describe the bug**

Segmentation fault (core dumped)
when using darktable-cli from version dt 2.7.0+1972~g066980eb1

**Steps to reproduce**

> /opt/darktable/bin/darktable-cli Something.RAF Name.jpg

Backtrace written into 
[darktable_bt_T3RD9Z.txt](https://github.com/darktable-org/darktable/files/3712908/darktable_bt_T3RD9Z.txt)

**Platform:**
OS: Manjaro 18.1.0 KDE

**Additional context**
CPU 3rd generation Ryzen (3900X)
Motherboard Aorus Pro

Help, please...

/Claes in Lund, Sweden

TurboGit:
Problems seems obvious with the trace. Will fix.

ClaesGitHub:
Merci! But sorry: the bug is still there.
Here is a fresh dump, based on latest dt version:
[newDump.txt](https://github.com/darktable-org/darktable/files/3713924/newDump.txt)

TurboGit:
That's another instance of the same bug. Will fix !

TurboGit:
Second fix installed. Can you try again ? Thanks.

TurboGit:
Good to hear! Thanks.

PR #3127, opened 2019-10-12
fix metadata export issues #3119

phweyland:
Should be fixed:
- ctrl-e in darkroom
- lua export
- dt cli - set default export metadata params instead of null

I have not tested lua export, nor dt cli.
Few code changes but some moves to common library.


phweyland:
Should have solved also #3114. I had removed the null test in exif.cc after adding metadata everywhere (so I thought :-) ). Thanks.

TurboGit:
Works good on my side. Code looks ok. Thanks.

PR #3129, opened 2019-10-12
iop_order: fix allocation success check in OpenCL colorspace transform

flannelhead:
Noticed this suspicious-looking `NULL` check in `iop_order`.

There's another issue where at least some `iop`s give a black or grey image when enabled together with OpenCL. I'll post a proper ticket probably tomorrow, didn't have time to scrutinize the issue further yet.


TurboGit:
Obviously correct ! Thanks.

Issue #3026, opened 2019-09-27
What impact will existing styles have with v3.0 


chhil:
**Describe the bug/RFI**
I have imported styles and created a couple, with the change in iop order fixes what steps should be taken to get styles in line if any?
Do they get impacted ?

**Platform (please complete the following information):**
 - OS: Any
 - Version v 2.7.xxxx

jenshannoschwalm:
Hi, unfortunately it took a while to understand possible problems related to style & iop_order.
There is indeed a problem here that should be fixed for 3.0

@TurboGit I observed severe dt crashes after applying old styles requiring several restart/crash cycles. When using the -d ioporder option it became obvious that adding that style resulted in **different modules having the same iop_order** and that lead to the crash (in my specific case that was iop=23.0 but i think that doesn't matter).

Unfortunately the darktable_style version="1.0" doesn't give a clue about the iop order version (or is this wrong?). The existing on-the-fly conversion can't be used as-is here, always to default is also **not** ok.

Probably we have to go the hard way and for each entry in style we must ensure a proper iop depending on already existing modules in the history stack.

jenshannoschwalm:
I think this is fixed. Please concide to close issue :-)

chhil:
@jenshannoschwalm 
There is a long trail of comments and separate tickets for this. 
Could you give me a one liner on what one should  keep in mind when I import a style that has modules that are spread across IOP order?

jenshannoschwalm:
Sure, allow me a few more lines :-)

There have been really a number of issues about history, iop_order and styles. Some very technical.

A style basically adds a module with style-defined parameters to the pipeline. This has been the case and still is.

The iop_order of the inserted module is **not** defined by the style but is calculated from default iop_order for the target image respecting multiple instances of modules.

So you can **not** use a style to apply a certain modified iop_order.

Atm we can't do it otherwise.

But be patient, you can use copy & paste (preferably using the selection mode) history to do so.
If the source & target have the same history version the iop_order is respected.

chhil:
@jenshannoschwalm Thank you for the few lines ;)

Really appreciate you taking the time to fix and explain things.

PR #3140, opened 2019-10-13
Fixing iop_order when applying a style

jenshannoschwalm:
There is one thing fundamentally wrong (at least i think so) with styles and iop_order.
The style database simply does **not know** about the iop_order version!

When a style is applied to an image the iop_orders for each module are granted to be
correct and are copied to the history as is.

Of course this might be exactly what you want and i think it would be great to have styles
with specifically re-ordered modules but at the moment this is absolutely unsafe.

Iop_orders in old styles can be completely out-of-order and i observed multiple crashes after applying such styles.
These crashes left dt in deep trouble resulting in multiple crashes as long there were images with
thumbnails to be updated.

What can be done?

We need a new style database including iop_order version so we can test: Are the iop_orders
valid in this style or do we have to make safe assumtions? This means a code overhaul
at many places, certainly **not** for dt3.0

This pr is meant as a security fix. What is done:

1. When you apply a style the iop_order is **not** taken from the style database but is read from
the images iop_order list (depending on it's iop_order version). For multiple instances the iop_order is increased by 1/100 per multi_instance.

2. A minor fix, if we apply a style we also want the modules tab to be redisplayed.

### DIFF
--- a/src/common/styles.c
+++ b/src/common/styles.c
@@ -39,6 +39,8 @@
 #include <stdio.h>
 #include <string.h>
 
+#define DT_IOP_ORDER_INFO (darktable.unmuted & DT_DEBUG_IOPORDER)
+
 typedef struct
 {
   GString *name;
@@ -686,7 +688,23 @@ void dt_styles_apply_to_image(const char *name, gboolean duplicate, int32_t imgi
 
     dt_dev_pop_history_items_ext(dev_dest, dev_dest->history_end);
 
+    int my_iop_order_version = 0;
+    // check images iop order version
+    DT_DEBUG_SQLITE3_PREPARE_V2(dt_database_get(darktable.db), "SELECT iop_order_version FROM main.images WHERE id = ?1",
+                              -1, &stmt, NULL);
+    DT_DEBUG_SQLITE3_BIND_INT(stmt, 1, imgid);
+    if(sqlite3_step(stmt) == SQLITE_ROW)
+    {
+      my_iop_order_version = sqlite3_column_int(stmt, 0);
+    }
+    sqlite3_finalize(stmt);
+
+    GList *current_iop_list = dt_ioppr_get_iop_order_list(&my_iop_order_version);
+
     dt_ioppr_check_iop_order(dev_dest, newimgid, "dt_styles_apply_to_image 1");
+
+    if (DT_IOP_ORDER_INFO)
+      fprintf(stderr,"\n^^^^^ Apply style on image %i, iop_order version %i, history size %i",imgid,my_iop_order_version,dev_dest->history_end);
 
     // go through all entries in style
     DT_DEBUG_SQLITE3_PREPARE_V2(dt_database_get(darktable.db),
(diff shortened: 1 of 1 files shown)


TurboGit:
> The style database simply does **not know** about the iop_order version!

And I would say that this is fine as a style can be applied to image using any iop order version.

TurboGit:
> We need a new style database including iop_order version so we can test: Are the iop_orders
> valid in this style or do we have to make safe assumtions? This means a code overhaul
> at many places, certainly **not** for dt3.0

And probably not a good solution anyway. See previous comment.

TurboGit:
>     1. When you apply a style the iop_order is **not** taken from the style database but is read from
>        the images iop_order list (depending on it's iop_order version). For multiple instances the iop_order is increased by 1/100 per multi_instance.

To me your solution implemented in this PR is the correct one and the only one that seems a sensible solution. Of course I may be missing something...

src/common/styles.c
```diff
@@ -713,12 +731,23 @@ void dt_styles_apply_to_image(const char *name, gboolean duplicate, int32_t imgi
       style_item.blendop_params = (void *)sqlite3_column_blob(stmt, 5);
       style_item.params_size = sqlite3_column_bytes(stmt, 3);
       style_item.blendop_params_size = sqlite3_column_bytes(stmt, 5);
-      style_item.iop_order = sqlite3_column_double(stmt, 9);
 
+      double old_iop_order = sqlite3_column_double(stmt, 9);
```
TurboGit: const

jenshannoschwalm: right :-(

jenshannoschwalm:
> And I would say that this is fine as a style can be applied to image using any iop order version.

Fine. But this leaves the question is the specific modules iop_order in a style correct for the image it is applied on?

jenshannoschwalm:
There could be a situation where you want to have some module, let's say lut3d in a certain place in the pipeline. This would require it to be positioned with more rules like
1. images iop_order version
2. fences
3. The style would also have to give more hints.

I think this is nothing for now for sure!

TurboGit:
> But this leaves the question is the specific modules iop_order in a style correct for the image it is applied on?

I see, a fully correct fix would be to handle the style as an history and do a migration of it from any iop-order version to any other as done in iop_order.c (_ioppr_legacy_iop_order_step). But for this you neeed iop-order-version in the style indeed. This will be tricky but probably something to do indeed. Maybe for 3.0.1.

jenshannoschwalm:
Anyway this pr gives a safe solution. And if you add a style you will most likely want to develop further so you can reorder if you like.

jenshannoschwalm:
@TurboGit ready from my side.

jenshannoschwalm:
Fixes #3026

junkyardsparkle:
Ouch. I've been using the ability of a style to manipulate iop position rather heavily since the the re-ordering was implemented. I understand the issue with styles created with earlier versions of the ordering and the "safety" aspect, but... would it actually be unfeasible for 3.0 to add some version flag checking to styles to allow newly created ones to have this functionality?

I think it will be confusing for people when saving a style, then applying it doesn't reproduce the effect it was intended to because an iop position was changed but the change wasn't reproduced on "playback"... if it can't be done for 3.0, this situation should probably be made clear in the release notes, at least. :-)

jenshannoschwalm:
@junkyardsparkle yes, it should be in the manual "Please note that currently styles do not reflect the order in the processing pipeline"

About the flag. The version has to be included in the database **and** the style file ex- or imported. That needs more thinking and a code overhaul at more places. So not for feature freeze i guess.

@TurboGit BTW there is one more problem that needs attention. Not sure about: are ^C ^V on history safe right now.  I will possibly report with another pr.

junkyardsparkle:
> About the flag. The version has to be included in the database and the style file ex- or imported. That needs more thinking and a code overhaul at more places. So not for feature freeze i guess.

Fair enough (and I really appreciate the effort going into making the iop ordering a not-broken feature despite being a never-ending "can of worms")... would it be reasonable to implement, for now, a "hidden" darktablerc preference to retain the previous order-from-styles behavior for those who understand the issues and would really like to use the functionality (without having to patch every single build)?  :-)

jenshannoschwalm:
I'd say it't the housekeeper to decide. Technically it's not a problem, although when applying a style in an unchecked way may lead to much deeper trouble. @TurboGit Do you favour such a flag?

junkyardsparkle:
> Technically it's not a problem, although when applying a style in an unchecked way may lead to much deeper trouble.

Understood. This would just be a way to use the ordering for styles created after the 3.0 release by people who really want to be able to do that, while still preventing the issues with other users applying "legacy" styles. Otherwise, there's really no way to make efficient use of the full potential of reordering, which is a major feature of the upcoming release IMHO. It seems like a reasonable compromise until a more comprehensive version checking can be implemented.

TurboGit:
I have moved the discussion here: #3149 

We really need to go back to design step as this is clearly a "can of worms" indeed :)

PR #3143, opened 2019-10-13
Making the hue sliders of the colorbalance iop more consistent

msdm:
This PL makes the "step" values for the hue sliders of color balance iop equal to each other. Also, it fine-tunes them to have a better experience in changing the values using the mouse wheel.

### DIFF
--- a/src/iop/colorbalance.c
+++ b/src/iop/colorbalance.c
@@ -2500,7 +2500,7 @@ void gui_init(dt_iop_module_t *self)
   (void)lift_messages;
   ADD_FACTOR(lift)
 
-  g->hue_lift = dt_bauhaus_slider_new_with_range_and_feedback(self, 0.0f, 360.0f, 0.5f, 0.0f, 2, 0);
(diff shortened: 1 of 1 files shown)


TurboGit:
This looks more like an enhancement than a bug fix. I have so filed this doe 3.0.1 as we currently are in feature freeze.

msdm:
No problem. Actually, I am not sure if this is a bug fix or an enhancement. The step value of the hue slider for highlights has been mistakenly set to 0.05 while the other two hue sliders (shadows and midtones) have a 0.5 step value. I think this causes inconsistent behavior when someone uses the mouse wheel to change the value of those three sliders.

TurboGit:
I'll wait for @aurelienpierre comment anyway here as he is the author of this module.

aurelienpierre:
I'm fine with that.

TurboGit:
Given the amount of changes recently made in master I see that as ok now.

TurboGit:
Thanks.

Issue #3089, opened 2019-10-07
Portrait pictures imported with LightRoom edits are displayed head-down


leclercj:
**Describe the bug**
When I'm importing portrait pictures in darktable previously edited with LightRoom, I got the images head-down in darktable

**To Reproduce**
1. Edit a portrait picture in LightRoom so that a LightRoom xmp file is created and associated to the picture in the disk.
2. Switch to darktable and import this same picture
3. See the picture displayed head down.

**Expected behavior**
Portrait picture displayed head up
(and LightRoom edits translated into darktable edits)

**Platform (please complete the following information):**
 - OS: linux
 - Version: 2.7.0+1956~g552a1d6f7

**Additional context**
- Same operation with darktable 2.4 and 2.6 was working fine.
- Same portrait picture imported in darktable 2.7 without the LightRoom xmp is correctly displayed (but LightRoom edits lost)

leclercj:
Sample of wrongly imported picture available here (too big for GitHub, so using We Transfer, link available 1 week only)
https://we.tl/t-noHrdVuZGV

TurboGit:
Thanks for reporting. I suppose this is due to the fix on orientation put in dartable which has not been propagated to the lightroom import module. I'll have a look.

leclercj:
Good progress with this https://github.com/darktable-org/darktable/pull/3100
It fixes my sample case, but not all cases.
I'll propose a complement PR after additional tests

TurboGit:
I was sure that some other cases where not properly fixed. I have no mean to test, so yes please open new PR for the other issues with what is needed to reproduce. Thanks a lot!

PR #3146, opened 2019-10-13
Fix orientation in LR import

leclercj:
This is to fix https://github.com/darktable-org/darktable/issues/3089
Covers all orientation/flip cases and tested with Lightroom 6
Difficult to share the test patterns, as Raw files are too large to be shared in GitHub

### DIFF
--- a/src/common/image.h
+++ b/src/common/image.h
@@ -89,6 +89,18 @@ typedef struct dt_image_raw_parameters_t
   unsigned user_flip : 8; // +8 = 32 bits.
 } dt_image_raw_parameters_t;
 
+typedef enum dt_exif_image_orientation_t
+{
+  EXIF_ORIENTATION_NONE              = 1,
+  EXIF_ORIENTATION_FLIP_HORIZONTALLY = 2,
+  EXIF_ORIENTATION_FLIP_VERTICALLY   = 4,
+  EXIF_ORIENTATION_ROTATE_180_DEG    = 3,
+  EXIF_ORIENTATION_TRANSPOSE         = 5,
+  EXIF_ORIENTATION_ROTATE_CCW_90_DEG = 8,
+  EXIF_ORIENTATION_ROTATE_CW_90_DEG  = 6,
+  EXIF_ORIENTATION_TRANSVERSE        = 7
+} dt_exif_image_orientation_t;
+
 typedef enum dt_image_orientation_t
 {
   ORIENTATION_NULL    = -1,     //-1, or autodetect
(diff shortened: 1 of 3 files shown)


leclercj:
Does NOT fix the crop&rotate/clipping import, also broken. I'll prepare another PR for that part.

src/common/image.h
```diff
@@ -89,6 +89,18 @@ typedef struct dt_image_raw_parameters_t
   unsigned user_flip : 8; // +8 = 32 bits.
 } dt_image_raw_parameters_t;
 
+typedef enum dt_exif_image_orientation_t
+{
+  EXIF_ORIENTATION__NONE              = 1,
```
TurboGit: For consistency, in the current code we do not use double underscore. So I'll do the same here and use a single underscore after ORIENTATION.

leclercj: You're right. Let me fix that

leclercj:
Change done

src/common/image.h
```diff
@@ -89,6 +89,18 @@ typedef struct dt_image_raw_parameters_t
...
 } dt_image_raw_parameters_t;
 
+typedef enum dt_exif_image_orientation_t
+{
+  EXIF_ORIENTATION_NONE              = 1,
+  EXIF_ORIENTATION_FLIP_HORIZONTALLY = 2,
+  EXIF_ORIENTATION_FLIP_VERTICALLY   = 4,
+  EXIF_ORIENTATION_ROTATE_180_DEG    = 3,
+  EXIF_ORIENTATION_TRANSPOSE         = 5,
+  EXIF_ORIENTATION_ROTATE_CCW_90_DEG = 8,
```
TurboGit: I think CCW and CW are inverted.

TurboGit: To be precise, I mean CW should be 8 and CCW 6.

leclercj: I didn't changed that and beleive this is correct and not inverted
See curent definition in image.h:
```
/** return the raw orientation, from jpg orientation. */
static inline dt_image_orientation_t dt_image_orientation_to_flip_bits(const int orient)
{
  switch(orient)
  {
    case 1:
      return ORIENTATION_NONE;
    case 2:
      return ORIENTATION_FLIP_HORIZONTALLY;
    case 3:
      return ORIENTATION_ROTATE_180_DEG;
    case 4:
      return ORIENTATION_FLIP_VERTICALLY;
    case 5:
      return ORIENTATION_TRANSPOSE;
    case 6:
      return ORIENTATION_ROTATE_CW_90_DEG;
    case 7:
      return ORIENTATION_TRANSVERSE;
    case 8:
      return ORIENTATION_ROTATE_CCW_90_DEG;
    default:
      return ORIENTATION_NONE;
  }
```

Also, see exif values described here : https://www.daveperrett.com/articles/2012/07/28/exif-orientation-handling-is-a-ghetto/
- when a picture is in case 6, we have to apply a CW rotation to correct it
- when a picture is in case 8, we have to apply a CCW rotation to correct it

TurboGit: Ok, I see you're correct. The number is the rotation to be applied. Thanks.

PR #3155, opened 2019-10-14
Fix crop in LR import

leclercj:
Fix crop data migration from Lightroom to darktable
crop migration moved after orientation migration, and done with respect of user orientation
merge_two_orientations() moved from iop/flip.c to common/image.c for code factoring

### DIFF
--- a/src/develop/lightroom.c
+++ b/src/develop/lightroom.c
@@ -1230,26 +1230,30 @@ void dt_lightroom_import(int imgid, dt_develop_t *dev, gboolean iauto)
 
     if(data.has_crop)
     {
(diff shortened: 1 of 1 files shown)


leclercj:
This morning I'm not sure anymore this is the shortest & simpliest way to fix the problem. Let me think about it one day before merging.

TurboGit:
Sure! Thanks for working on this @leclercj !

leclercj:
Less changes for the same result.
Tested with Lightroom 6 with combinations of exif (raw format) rotation + user rotations in LR
Should be ok to merge now.

TurboGit:
Perfect! Thanks.

PR #3156, opened 2019-10-14
don't compare pointer value to numeric constant

rabauke:
Some tools indicated this issue. I am not completely certain if my fix is correct. Short review would be appreciated.


TurboGit:
Nice finding! A real bug in fact. Thanks.

PR #3157, opened 2019-10-14
remove superfluous argument from function pointer call

rabauke:
Note version was defined as

    int (*version)();

which denotes a function pointer to a function that takes an _unspecified number_ of arguments. See also _Notes_ in https://en.cppreference.com/w/c/language/function_declaration


TurboGit:
Looks good to me! Nice clean-up/fixes. Thanks.

PR #3166, opened 2019-10-15
fix format strings

rabauke:
Fixes various format string related (portability) issues.


TurboGit:
Looks ok. Thanks.

chhil:
In mipmap is the hardcoded "PRIu32" intentional ?

rabauke:
This is the portable way to print a uint32_t via printf, see documentation on inttypes.h.

PR #3168, opened 2019-10-15
enhance code readability

rabauke:
My static code analysis tool indicated (falsely) some errors here and I also had some difficulties to understand the code. The revised version is technically equivalent to the former but more explicit and therefore more easy to understand.


cryptomilk:
In those cases I normally use brackets.

`int *ip = &(lookup[row][col][1]);`

Which makes sure that the array element is referenced and makes it clearer to the reader.

rabauke:
> Can you take @cryptomilk comment into account? Thanks.

@TurboGit This has been incorporated into my recent revision of this pull request.

rabauke:
I think, it is safe to merge this now.

TurboGit:
Ok, merging now. Thanks.

PR #3170, opened 2019-10-15
enhance portability

rabauke:
Code assumes that `int` has 32 bits. Shifting a _signed_ integer bit by 31 is undefined behavior. Replaces some functions by standard library calls.

### DIFF
--- a/src/iop/basicadj.c
+++ b/src/iop/basicadj.c
@@ -792,15 +792,6 @@ void gui_cleanup(struct dt_iop_module_t *self)
   self->gui_data = NULL;
 }
 
-static inline double mla(double x, double y, double z)
-{
-  return x * y + z;
-}
-static inline int xisinf(double x)
-{
-  return x == INFINITY || x == -INFINITY;
-}
-
 static inline int64_t doubleToRawLongBits(double d)
 {
   union {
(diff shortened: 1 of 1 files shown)


cryptomilk:
Did you start running darktable with libubsan?

rabauke:
@cryptomilk From time to time I throw some static code analysis tools (cppcheck, Clion, PVS Studio) on darktable. I did not employ santinizers so far.

cryptomilk:
I'm using csbuild for that in CI:

https://gitlab.com/cmocka/cmocka/-/jobs/322207429

https://github.com/kdudka/csmock

TurboGit:
Thanks.

src/iop/basicadj.c
```diff
@@ -830,14 +821,13 @@ static inline int ilogbp1(double d)
...
 }
 
-static inline double ldexpk(double x, int q)
+// calculate  x * 2^q
+static inline double ldexpk(double x, int32_t q)
 {
-  double u;
-  int m;
-  m = q >> 31;
+  int32_t m = q < 0 ? 1 : 0;
```
LebedevRI: 1. there is nothing wrong with the original code
2. the new code is blatantly wrong https://godbolt.org/z/Pvf7ao

parafin: 1. I'm not sure how it's in C, but in C++ until C++-20 results of right-shift of negative values is implementation-defined, so with >>31 you may get non-negative number...
2. Indeed

rabauke: Quoting ISO/IEC 9899:1999 (E)

> The result of E1 >> E2 is E1 right-shifted E2 bit positions. If E1 has an unsigned type
or if E1 has a signed type and a nonnegative value, the value of the result is the integral
part of the quotient of E1 / 2^E2 . If E1 has a signed type and a negative value, the
resulting value is implementation-defined.

I am not aware of that this would have changed in more recent versions of the C programming language.

parafin: Still, it has to be `int32_t m = q < 0 ? -1 : 0;`

Issue #3164, opened 2019-10-15
Culling mode history copying doesn't work


chhil:
**Describe the bug**
I have images that I have edited and put 3 starts on them. In Culling mode I select the filter of 3 stars. Then select an image and copy history, I make sure its copy and not copy all so that I know the history popped up is being selected.
Then I select 2 starred images , the ones I have not edited. Then select an image and paste history to see whats being pasted.
Then I open this image in darkroom and the history is not updated,  its the default history of a fresh image import.

OS win 10
DT : 2.7.0.2015

jenshannoschwalm:
Has been fixed. Concider to close issue please.

chhil:
Successfully tested fix on darktable-2.7.0+2054~gdd12b7089-win64

Issue #3123, opened 2019-10-11
Copying history stack between photos from Fuji X-E2 results in double base curve module


tomaszg7:
**Describe the bug**
I had to process several RAW files from Fuji X-E2 and noticed a strange behaviour that was not present when working with RAWs from any of my own Canon cameras. Namely, while copying history stack between photos, the `base curve` module was duplicated. Both copies had the same default curve (I didn't touch it at any point in my workflow) resulting in too bright image. I had to manually delete one instance of this module and things returned to normal.

**To Reproduce**
1. Add two photos from Fuji X-E2 to Darktable
2. Do some changes to one of them.
3. Copy the changes to another photo (in darkroom mode with ^c and ^v)
4. Notice double base curve module.

**Expected behavior**
Only one copy of base module should be enabled.

**Platform (please complete the following information):**
 - OS: Gentoo Linux
 - Version fa2d1c204ab73a4c0d2d4600d66531c29cb8ed38

jenshannoschwalm:
Did you auto-apply any style?

tomaszg7:
I have auto-apply on demosaic (selecting Amaze), which should not work on Fuji X-trans cameras. I don't think I auto-apply any other style but I don't know how to check it globally. For sure nothing on base curve module.

jenshannoschwalm:
There is only one X-E2 file on pixls. Could you share another?

tomaszg7:
I would guess making a copy of that one photo should do the trick for the purpose of this bug, but here are two more: http://tomaszg.pl/fuji I'll keep them online for a few days.

TurboGit:
Yes, I suppose it is also related to #3153 as I have also seen strange things and indeed some times double base curve!

jenshannoschwalm:
Hi @tomaszg7 , i tested with latest master and for me it seems to be fixed. If you auto apply basecurve you will see two entries after copy&merge but (at least here) no strange effects. Can you confirm?

TurboGit:
Fixed, closing.

PR #3172, opened 2019-10-16
Fixing copy&paste of history

### LINKED ISSUE #3153: copy all / paste all
**Describe the bug**
copy all / paste all from lighttable completely destroy the edits

**To Reproduce**
1. Select one image
2. Got to darkroom
3. Activate module monochrome (easy effect to see)
4. Got back to lighttable
5. Select modified image, do copy all (or ctrl + c)
6. Select all images, do paste all (or ctrl + v)
7. Many images are fully black, open it in darkroom, the history stack is fully broken

**Expected behavior**
Only monochrome is added on all image history

 - OS Debian
 - Version 2.7.0+2009~g896cbe3c4-dirty

jenshannoschwalm:
See discussion issue #3153

This is what i made of the discussion, it should fix #3153 #3123 #3164 .

- tests in history.c
- writes to console what is done when using -d ioporder
- commented a **very** noisy and now irrelevant iop_order information
- modified the hist_dialog disabling gamma. Any better solution?

Please review & test @TurboGit and @parafin

### DIFF
--- a/src/common/history.c
+++ b/src/common/history.c
@@ -612,10 +612,9 @@ static int _history_copy_and_paste_on_image_merge(int32_t imgid, int32_t dest_im
       dt_iop_module_t *mod_src = (dt_iop_module_t *)(modules_src->data);
 
       // copy from history only if
-      if((_search_history_by_module(dev_src, mod_src) != NULL) &&  // module is in history of source image
-         !(mod_src->default_enabled && !mod_src->enabled) &&       // it's not a disabled module normally enabled by default
-         !dt_iop_is_hidden(mod_src) &&                             // it's not a hidden module
-         !memcmp(mod_src->params, mod_src->default_params, mod_src->params_size)
+      if((_search_history_by_module(dev_src, mod_src) != NULL) && // module is in history of source image
+         !(mod_src->default_enabled && mod_src->enabled && !memcmp(mod_src->params, mod_src->default_params, mod_src->params_size) && // it's not a enabled by default module with unmodified settings
+         !dt_iop_is_hidden(mod_src)) 
         )
       {
         double old_iop_order = mod_src->iop_order;
--- a/src/common/history.c
+++ b/src/common/history.c
@@ -611,9 +611,12 @@ static int _history_copy_and_paste_on_image_merge(int32_t imgid, int32_t dest_im
     {
       dt_iop_module_t *mod_src = (dt_iop_module_t *)(modules_src->data);
 
-      // only if module is in history of source image, not auto-enabled and not-hidden
-      if((_search_history_by_module(dev_src, mod_src) != NULL) &&
-         !mod_src->default_enabled && !dt_iop_is_hidden(mod_src))
+      // copy from history only if
+      if((_search_history_by_module(dev_src, mod_src) != NULL) &&  // module is in history of source image
+         !(mod_src->default_enabled && !mod_src->enabled) &&       // it's not a disabled module normally enabled by default
+         !dt_iop_is_hidden(mod_src) &&                             // it's not a hidden module
+         !memcmp(mod_src->params, mod_src->default_params, mod_src->params_size)
+        )
       {
         double old_iop_order = mod_src->iop_order;
         if (iop_order_version_src != iop_order_version_dest)
--- a/src/common/history.c
+++ b/src/common/history.c
@@ -34,6 +34,8 @@
 #include "develop/blend.h"
 #include "develop/masks.h"
 
+#define DT_IOP_ORDER_INFO (darktable.unmuted & DT_DEBUG_IOPORDER)
+
 void dt_history_item_free(gpointer data)
 {
   dt_history_item_t *item = (dt_history_item_t *)data;
@@ -546,6 +548,8 @@ static int _history_copy_and_paste_on_image_merge(int32_t imgid, int32_t dest_im
   dev_dest->iop = dt_iop_load_modules_ext(dev_dest, TRUE);
 
   dt_dev_read_history_ext(dev_src, imgid, TRUE);
+
+  // This prepends the default modules and converts just in case it's an empty history
   dt_dev_read_history_ext(dev_dest, dest_imgid, TRUE);
 
   dt_ioppr_check_iop_order(dev_src, imgid, "_history_copy_and_paste_on_image_merge ");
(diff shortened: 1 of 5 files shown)


src/common/history.c
```diff
@@ -570,31 +584,53 @@ static int _history_copy_and_paste_on_image_merge(int32_t imgid, int32_t dest_im
...
     GList *modules_src = g_list_first(dev_src->iop);
     while(modules_src)
     {
       dt_iop_module_t *mod_src = (dt_iop_module_t *)(modules_src->data);
 
-      // but only if module is in history in source image
-      if(_search_history_by_module(dev_src, mod_src) != NULL)
+      // only if module is in history of source image, not auto-enabled and not-hidden
+      if((_search_history_by_module(dev_src, mod_src) != NULL) &&
+         !mod_src->default_enabled && !dt_iop_is_hidden(mod_src))
```
parafin: My suggestion was a little different, by checking also that params != default_params. Also, I just realized, we need to check enabled - some default_enabled modules can be disabled, and that should probably be copy-pastable.

src/gui/hist_dialog.c
```diff
@@ -217,12 +217,14 @@ int dt_gui_hist_dialog_new(dt_gui_hist_dialog_t *d, int imgid, gboolean iscopy)
     do
     {
       dt_history_item_t *item = (dt_history_item_t *)items->data;
-
-      gtk_list_store_append(GTK_LIST_STORE(liststore), &iter);
-      gtk_list_store_set(GTK_LIST_STORE(liststore), &iter, DT_HIST_ITEMS_COL_ENABLED,
+      // This is sort of a hack, any better solution?
+      if (strcmp(item->op,"gamma") != 0)
```
parafin: I was mostly concerned with history widget in darkroom, not copy-paste dialog;)

jenshannoschwalm: Just in case nobody misses it :-)

src/gui/hist_dialog.c
```diff
@@ -217,12 +217,14 @@ int dt_gui_hist_dialog_new(dt_gui_hist_dialog_t *d, int imgid, gboolean iscopy)
     do
     {
       dt_history_item_t *item = (dt_history_item_t *)items->data;
-
-      gtk_list_store_append(GTK_LIST_STORE(liststore), &iter);
-      gtk_list_store_set(GTK_LIST_STORE(liststore), &iter, DT_HIST_ITEMS_COL_ENABLED,
+      // This is sort of a hack, any better solution?
```
TurboGit: Maybe we can test if commit_params() is present or NULL. The module gamma has no parameter and so no commit_params() routine, in such a case we know that it cannot be modified. At least we know that there is nothing to copy. What do you think?

parafin: maybe just remove all IOP_FLAGS_HIDDEN iops?

src/gui/hist_dialog.c
TurboGit: Maybe checking this in dt_history_get_items() ?

As this routine is used for both style dialog and history dialog we will avoid also gamma in the style dialog?

src/gui/hist_dialog.c
TurboGit: My proposal is not simple as at this point we only have an history item with no module information at all.

src/gui/hist_dialog.c
TurboGit: BTW, a check for mask_manager is already done in  dt_history_get_items() so it seems a good place to put a second hack at the same place until a clean solution is found.

src/gui/hist_dialog.c
TurboGit: @parafin : that would remove gamma, finalscale, mask_manager, overexposed & rawoverexposed. Sounds like it will work yes. But again, at this point we don't have the module information, just the module name. So we need first to get the module given the name and check the flags. Would be the best solution.

TurboGit: Let me add a new routine get_module_by_name() to make this possible.

src/common/history.c
jenshannoschwalm: Ok, committed as i understood

src/gui/hist_dialog.c
jenshannoschwalm: Yes. I chose the flag testing version

src/gui/hist_dialog.c
```diff
@@ -209,6 +214,12 @@ int dt_gui_hist_dialog_new(dt_gui_hist_dialog_t *d, int imgid, gboolean iscopy)
   gtk_tree_selection_set_mode(gtk_tree_view_get_selection(GTK_TREE_VIEW(d->items)), GTK_SELECTION_SINGLE);
   gtk_tree_view_set_model(GTK_TREE_VIEW(d->items), GTK_TREE_MODEL(liststore));
 
+  // we need access to the modules flags so ...
```
TurboGit: I have a simpler solution where you could change the previous code with:

-      // This is sort of a hack, any better solution?
-      if (strcmp(item->op,"gamma") != 0)
+      if(!(get_module_flags(item->op) & IOP_FLAGS_HIDDEN))
       {

I'm about to push the commit to have the get_module_flags() routine.

src/common/history.c
```diff
@@ -570,31 +584,56 @@ static int _history_copy_and_paste_on_image_merge(int32_t imgid, int32_t dest_im
...
     {
       dt_iop_module_t *mod_src = (dt_iop_module_t *)(modules_src->data);
 
-      // but only if module is in history in source image
-      if(_search_history_by_module(dev_src, mod_src) != NULL)
+      // copy from history only if
+      if((_search_history_by_module(dev_src, mod_src) != NULL) &&  // module is in history of source image
+         !(mod_src->default_enabled && !mod_src->enabled) &&       // it's not a disabled module normally enabled by default
+         !dt_iop_is_hidden(mod_src) &&                             // it's not a hidden module
+         !memcmp(mod_src->params, mod_src->default_params, mod_src->params_size)
```
TurboGit: This means that a module with default param will be excluded and so will not override dest module with non default params?

TurboGit: Confirmed by testing:
- select two images reset history
- go to image A, set exposure to +2EV
- go to image B set exposure to default (click reset settings)
- go to lighttable, select history on B (ctrl+c)
- paste on A, exposure is not changed and set to +2EV

To me this is not the expected behavior.

TurboGit: Note that copying default params is the only way to reset a module on a set of pictures otherwise the only solution is to go to each image individually and disable the module.

parafin: Yeah, that's wrong. memcmp check is only relevant for default_enabled. Correct if should look like that:
```
if((_search_history_by_module(dev_src, mod_src) != NULL) && // module is in history of source image
!(mod_src->default_enabled && mod_src->enabled && !memcmp(mod_src->params, mod_src->default_params, mod_src->params_size) && // it's not a enabled by default module with unmodified settings
!dt_iop_is_hidden(mod_src)) // it's not a hidden module
```

jenshannoschwalm: Of course you are right :-(

src/gui/hist_dialog.c
jenshannoschwalm: Thanks again.

jenshannoschwalm:
Unfortunately i messed up the branch by rebasing. Everything you suggested/fixed is in the "Fixing copy&paste of history (v2) branch now. Is it ok this way ?

PR #3176, opened 2019-10-17
common: Improve debug output if image support is missing

cryptomilk:
It should make more clear what is actually wrong and point them to clinfo.

See https://discuss.pixls.us/t/no-opencl-darktable-2-7-0-build-and-installed-from-darktable-git/14508


TurboGit:
Thanks.

PR #3181, opened 2019-10-18
lock: lock db at high level when copy/paste or deleting history.

TurboGit:


src/common/history.c
```diff
@@ -67,6 +67,8 @@ void dt_history_delete_on_image_ext(int32_t imgid, gboolean undo)
     dt_history_snapshot_undo_create(hist->imgid, &hist->before, &hist->before_history_end);
   }
 
+  dt_pthread_mutex_lock(&darktable.db_insert);
```
jenshannoschwalm: I think this is really good here. Makes much more sense than the lock in develop.c _dev_merge_history

src/develop/develop.c
```diff
@@ -1255,9 +1255,6 @@ static gboolean _dev_auto_apply_presets(dt_develop_t *dev)
 
   if(imgid <= 0) return FALSE;
 
-  // be extra sure that we don't mess up history in separate threads:
```
jenshannoschwalm: see above

jenshannoschwalm:
Oh man, thanks for this idea. just one more thought,look here

https://github.com/darktable-org/darktable/blob/f0d2245525f9cbc94af45bf9ad352c8b33e99c92/src/develop/develop.c#L1493-L1509

Does this section need extra care?

TurboGit:
> Does this section need extra care?

Hard to say, we need to find all call point for dt_dev_read_history_ext() and maybe lock at higher level.

TurboGit:
```
$ git grep dt_dev_read_history_ext
src/common/history.c:  dt_dev_read_history_ext(dev_src, imgid, TRUE);
src/common/history.c:  dt_dev_read_history_ext(dev_dest, dest_imgid, TRUE);
src/common/styles.c:      // TODO: this is copied from dt_dev_read_history_ext(), maybe do a helper with this?
src/common/styles.c:    dt_dev_read_history_ext(dev_dest, newimgid, TRUE);

```

TurboGit:
I think the part in history is covered.

TurboGit:
The styles part is not covered, we may want a lock in dt_styles_apply_to_image().

jenshannoschwalm:
Yes but don't we also need `dt_dev_read_history`?

TurboGit:
No :)

```
void dt_dev_read_history(dt_develop_t *dev)
{
  dt_dev_read_history_ext(dev, dev->image_storage.id, FALSE);
}
```

jenshannoschwalm:
Surprise -- surprise i see the code. But `dt_dev_read_history_ext` has no lock :-( What do i oversee?

TurboGit:
I'm stupid :) Indeed we need to handle dt_dev_read_history() has the locking is NOT in dt_dev_read_history_ext().

jenshannoschwalm:
Maybe just missing or old glasses :-)

I really think that `void dt_dev_load_image(dt_develop_t *dev, const uint32_t imgid)` might be the "really bad one".

TurboGit:
Actually maybe not that bad:

```
$ git grep dt_dev_read_history | grep -v _ext
src/develop/develop.c:  dt_dev_read_history(dev);
src/develop/develop.c:  dt_dev_read_history(dev);
src/views/darkroom.c:  dt_dev_read_history(dev);
```
after removing matches in string and the declaration/implementation.

TurboGit:
In darkroom.c it is when switching image. Not sure we need to lock here.

And the two in develop are for dt_dev_reload_history_items() and dt_dev_load_image().

jenshannoschwalm:
I found those too.

`static void dt_dev_change_image(dt_develop_t *dev, const uint32_t imgid)` also seems to be pretty complicated. But i strongly think we also have to handle this, as it also interferes with mipmap cache and filmstrip ...

TurboGit:
I just pushed a new version where I have tested:
- on the lt delete history on a bunch of images and enter a picture while refreshing
- on the lt copy and paste some history on a bunch of images and enter a picture while refreshing
- on the lt apply a style on a bunch of images and enter a picture while refreshing
- on darkroom's filmstrip copy paste history on a bunch of images and switch to another picture

No lock and history seems clean. Can you do some more testing?

Do you see something wrong that I missed?

src/develop/develop.c
```diff
@@ -685,6 +685,8 @@ void dt_dev_load_image(dt_develop_t *dev, const uint32_t imgid)
   dev->first_load = 1;
   dev->image_status = dev->preview_status = dev->preview2_status = DT_DEV_PIXELPIPE_DIRTY;
 
+  dt_pthread_mutex_unlock(&darktable.db_insert);
```
jenshannoschwalm: ??? Double unlocking in this function? Will certainly leave the underlying mutex in a problematic state.

jenshannoschwalm: If you intended a lock, wouldn't that have to include _dt_dev_load_raw some lines above too?

src/develop/develop.c
```diff
@@ -1004,6 +1008,9 @@ void dt_dev_free_history_item(gpointer data)
 void dt_dev_reload_history_items(dt_develop_t *dev)
 {
   dev->focus_hash = 0;
+
+  dt_pthread_mutex_unlock(&darktable.db_insert);
```
jenshannoschwalm: unlock???

jenshannoschwalm:
I tried after changing the 2 mentioned unlocks to lock.
First test lead to freeze when importing a folder while updating mipmaps.


Why? I didn't think about it yet but i think the underlying problem is the mutex initialising code.
The db.insert mutex needs to be recursive!
`dt_pthread_mutex_init(&(darktable.db_insert), NULL);`

src/develop/develop.c
TurboGit: I'm not sure as the load_raw do not touch the database.

src/develop/develop.c
TurboGit: cut&paste :(

TurboGit:
New version does not freeze. I have removed the lock in dt_dev_load_image() because there is nothing to lock here except dt_dev_read_history(dev) call which is already locked. That was the freeze you were experienced.

TurboGit:
I redid all the tests with the new version and again all seems fine now.

jenshannoschwalm:
> That was the freeze you were experienced.

Ok :-) Will do a thorough test tonight. Hospital work now ...

TurboGit:
@jenshannoschwalm : Thank! Just a little note that I have just introduced a recursive policy for the db_insert lock. That is, the **same thread** can lock an already owned lock and then I was able to reintroduce lock in dt_dev_load_image include the raw reading. This will help adding more lock for the db insert/read without headache :)

jenshannoschwalm:
- The `static inline int dt_pthread_mutex_init_with_caller` test for BSD mutexes is ok this way?
- Any changes required here for other OS platforms?

Thanks for making it recursive :-) locking larger sections now possible

parafin:
macOS doesn't support recursive pthread mutexes. Please revert that change. IMHO the need for recursive locking indicates bad design.

TurboGit:
@parafin : well each time we refactor dt there is risk that some lock is used at different place in the stack.  if I agree that it means something probably badly designed it is almost impossible to fix that at this time.

Now, a recursive lock for the same thread is quite common to me, is there any way this can be achieved on MacOS?

parafin:
Yes, there is a recursive mutex in macOS using different (ObjC) API. But in order to use it in darktable we need to introduce some kind of HAL layer, meaning that all dt_pthread_mutex_* calls will have to rewritten. Is this something anyone wants to do?

TurboGit:
@parafin : I suppose that if you are not up doing that nobody will do :)

I won't commit this last change then. But it will make it really harder to proper lock safely the db in a Gtk event oriented application like dt.

TurboGit:
@parafiin: looking at some forum it seems that recursive locking is supported by MacOS. I have just tested the following program:

```
#include <stdio.h>
#include <pthread.h>

void main(void)
{
  pthread_mutexattr_t attr;
  pthread_mutex_t lock;

  int res;
  res = pthread_mutexattr_init(&attr);
  printf("1 %d\n", res);

  res = pthread_mutexattr_settype(&attr, PTHREAD_MUTEX_RECURSIVE);
  printf("2 %d\n", res);

  res = pthread_mutex_init(&lock, &attr);
  printf("3 %d\n", res);

  res = pthread_mutex_lock(&lock);
  printf("4 %d\n", res);

  res = pthread_mutex_lock(&lock);
  printf("  4.1 %d\n", res);

  res = pthread_mutex_unlock(&lock);
  printf("  5.1 %d\n", res);

  res = pthread_mutex_unlock(&lock);
  printf("5 %d\n", res);
}
```

And it works as expected. If I replace the PTHREAD_MUTEX_RECURSIVE by PTHREAD_MUTEX_NORMAL it locks, so it seems that we have proper support for this on MacOS.

parafin:
Here's an example why recursive mutexes are bad. Let's say you have 2 mutexes: M1 and M2. Then there are 2 functions:
```
F1() {
lock M1;
lock M2;
do_some_work(); //requires M1 and M2 locked
unlock M2;
unlock M1;
}
F2() {
lock M1;
lock M2;
do_some_more_work(); //requires M1 and M2 locked
unlock M2;
unlock M1;
}
```
Everything works fine, then you add one more function:
```
F3() {
lock M2;
do_some_other_work(); // requires only M2 locked
F1();
unlock M2;
}
```
If M2 is recursive F3 will work most of the time. But a deadlock can happen because of the wrong order in which mutexes are now locked when F3 is called - consider what happens when F2 has already locked M1 in one thread, then F3 locks M2 in another thread, and here you go - it's a deadlock. So making mutexes recursive won't save you from having to understand the full call stack of each used function. You will just introduce other hard-to-catch bugs. Having mutexes non-recursive on the other hand highlights any problems right away when you double-lock some mutex.

If you are completely sure that F3 is valid this way (say there's no M1 mutex), then I can propose 2 solutions (aside from honest redesign of the code):

1. introduce F1_unlocked function (with just do_some_work()) and call it from F1 and F3 after locking the mutex;
2. add an argument locked (with default value false) to F1 function and lock/unlock the mutex only when it's false.

Both of these options aren't too pretty, but at least they are explicit on what is happening.

So, even if I had time to add recursive macOS mutex implementation (which I don't really), I wouldn't do it, because I think it's a bad idea in general.

jenshannoschwalm:
@parafin does the maxOS port **not** use pthreads? pthreads certainly supports recursive mutexes. Non-standard is unlocking a mutex that is owned by another thread.

TurboGit:
@jenshannoschwalm : it does support and as I just posted it seems to work as expected.

parafin:
Apple might have introduced recursive pthread mutexes in some newer OS release, but it wasn't supported at least in 10.7. So if you want to go that way, we will need to bump minimal required macOS version.

parafin:
Ah, I think I'm indeed mistaken. It's not recursion that isn't supported, it's pthread_mutex_timedlock (which isn't used by darktable). But I still think recursive mutexes are bad.

TurboGit:
> If M2 is recursive F3 will work most of the time. But a deadlock can happen because of the wrong order in which mutexes are now locked when F3 is called - consider what happens when F2 has already locked M1 in one thread, then F3 locks M2 in another thread, and here you go

Even with non recursive locking this can happen IINM. This is a classic locking when using multiple resources.

jenshannoschwalm:
@parafin about your example code F1-F3 sure you are right there but the code you present is about using several mutexes -- of course you have the problem you have. But here i think the code using a recursive mutex is just perfect as we use **one** mutex for locking the database.

Ok -- if we add another mutex also used by database access that would be prone to locks. So i think the only restriction we introduce by a recursive mutex is: "Stay with this single mutex"

TurboGit:
> So making mutexes recursive won't save you from having to understand the full call stack of each used function.

Yes, in some cases this will have to be understood properly. That's the standard story with locking anyway. But I really think that we will avoid more problems and allowing more locking when dealing with the DB (history, style...).

As seen on this thread we have some issues remaining since long time. You apply a style on a bunch on image then while refreshing you select one picture to enter darkroom. In many case you have garbage in history. I have learn to wait for the refresh to be over before enterring darkroom. But well, better to have a proper fix for that.

Let's see how my proposal works after @jenshannoschwalm  testing. On my side I have not yet seen the common issues I used to see.

parafin:
At least document this "Stay with this single mutex" rule somewhere. For anyone doing changes to the code in the future it won't be obvious that you have to look for possible locations of mutex locks up the call stack too, not just down the stack. My point is that recursive locks makes code much harder to understand. Sure it's easier to write, but it's hard to comprehend if it's valid.

jenshannoschwalm:
> As seen on this thread we have some issues remaining since long time. You apply a style on a bunch on image then while refreshing you select one picture to enter darkroom. In many case you have garbage in history. I have learn to wait for the refresh to be over before enterring darkroom. But well, better to have a proper fix for that.

Indeed a **very** old bug.
As promised i will do intensive testing.

The only downside from this bugfix i am already aware of is responsiveness -- you already mentioned this. We can handle this later, maybe 3.01; it won't be that difficult. We will need a per-image locking, probably just a state flag in image struct wrapped by a global mutex for setting, clearing and waiting for that state flag.

parafin:
> The only downside from this bugfix i am already aware of is responsiveness -- you already mentioned this. We can handle this later, maybe 3.01; it won't be that difficult. We will need a per-image locking, probably just a state flag in image struct wrapped by a global mutex for setting, clearing and waiting for that state flag.

Here goes the "single mutex" rule out of the window already... Locking is locking, no matter how you call it, "mutex" or "state flag"... OK, whatever...

jenshannoschwalm:
Let's stay to the "single mutex rule" :-)

There are two ideas that helped with other real time applications for me.

I we can't start developing an image right now because other jobs are also "fighting" for the mutex lock we could a) decrease the priority of the background threads b) add some nanosleep after an image update in the background thread.

Of course this depends on the OS sheduler but both result in a higher chance for the main thread to lock the mutex first.

jenshannoschwalm:
@TurboGit i tested a lot and i am quite happy to report i could **not** trigger any hang, crash or history pollution as we observed before.
Tested on current master plus mods from this pr, 8 threads enabled, with and without cl code on nvidia gpu.

TurboGit:
@jenshannoschwalm : that's great news! I'll merge then.

Thanks for all the testing !

PR #3184, opened 2019-10-19
Fix rgbcurve when switching first time to independant channels.

### LINKED ISSUE #3183: rgb curbe setting non consistent
**Describe the bug**
rgb curbe setting non consistent

**To Reproduce**
1. Select rgb curve
2. Change curve
3. Select the independent channels 
4. Only the R curve is set

**Expected behavior**
All curves (R, ,G, B) should be set with the same value

TurboGit:
In this case the G and B channels are set as the R channel to ensure
switching the first time to independant chanel will not change the
image rendering.

Fixes #3183.


TurboGit:
Manually merged.

PR #3189, opened 2019-10-20
update OpenCL headers and solve RGB OpenCL conversions issues

### LINKED ISSUE #3128: Black picture with flimic rgb
**Describe the bug**

I've edited a few photos with 'filmic rgb' already. With the latest master 0ae2c5cc174df610c79bcfa1c5781f1f4f286192 those pictures are rendered to completely black. Disabling filmic rgb doesn't fix it. If I go in the history stack back to the item before filmic was enabled the first time it works again ...

I'm using OpenCL (AMD ROCm)

@aurelienpierre

aurelienpierre:
Attempt to fix #3128.

I found that OpenCL colour conversions fail with Intel Neo and AMD, which use OpenCL 2.1, while Nvidia is fine, using OpenCL 1.2. This doesn't fix it, but it's maybe a first step : update the OpenCL headers from https://github.com/KhronosGroup/OpenCL-Headers (maybe it would be worth it to have it as a git submodule), and force the 1.2 version.

I think we will need to update the OpenCL version in the Clang test-compilation of kernels to get the errors newer hardware can have, but I don't remember where they are coded.

### DIFF
--- a/src/common/opencl.c
+++ b/src/common/opencl.c
@@ -400,7 +400,7 @@ static int dt_opencl_device_init(dt_opencl_t *cl, const int dev, cl_device_id *d
   escapedkerneldir = dt_util_str_replace(kerneldir, " ", "\\ ");
 #endif
 
-  options = g_strdup_printf("-cl-mad-enable %s -D%s=1 -I%s",
+  options = g_strdup_printf("-w -cl-fast-relaxed-math %s -D%s=1 -I%s",
                             (cl->dev[dev].nvidia_sm_20 ? " -DNVIDIA_SM_20=1" : ""),
                             dt_opencl_get_vendor_by_id(vendor_id), escapedkerneldir);
   cl->dev[dev].options = strdup(options);
@@ -412,7 +412,7 @@ static int dt_opencl_device_init(dt_opencl_t *cl, const int dev, cl_device_id *d
   g_free(escapedkerneldir);
   escapedkerneldir = NULL;
 
-  const char *clincludes[DT_OPENCL_MAX_INCLUDES] = { "colorspace.cl", "common.h", NULL };
+  const char *clincludes[DT_OPENCL_MAX_INCLUDES] = { "color_conversion.cl", "colorspaces.cl", "colorspace.cl", "common.h", NULL };
   char *includemd5[DT_OPENCL_MAX_INCLUDES] = { NULL };
   dt_opencl_md5sum(clincludes, includemd5);
 
--- a/data/kernels/basecurve.cl
+++ b/data/kernels/basecurve.cl
@@ -52,8 +52,8 @@ fast_expf(const float x)
 */
 kernel void
 basecurve_lut(read_only image2d_t in, write_only image2d_t out, const int width, const int height,
-              const float mul, read_only image2d_t table, global float *a, const int preserve_colors,
-              global const dt_colorspaces_iccprofile_info_cl_t *profile_info, read_only image2d_t lut,
+              const float mul, read_only image2d_t table, constant float *a, const int preserve_colors,
+              constant dt_colorspaces_iccprofile_info_cl_t *profile_info, read_only image2d_t lut,
               const int use_work_profile)
 {
   const int x = get_global_id(0);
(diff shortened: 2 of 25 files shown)


TurboGit:
> Neo and AMD, which uses OpenCL 2.1

Don't you mean 1.1 ?

aurelienpierre:
> Don't you mean 1.1 ?

No, 2.1. The latest OpenCL version is 2.2 as of today.

I finally fixed the bug by adding `-cl-opt-disable` in kernel building options (please @cryptomilk test and validate it), which disables all optimizations and trigger a ×3 performance penalty. Not sure it's a long-term solution, but at least we begin to understand where it comes from.

src/common/opencl.c
```diff
@@ -400,7 +400,7 @@ static int dt_opencl_device_init(dt_opencl_t *cl, const int dev, cl_device_id *d
   escapedkerneldir = dt_util_str_replace(kerneldir, " ", "\\ ");
 #endif
 
-  options = g_strdup_printf("-cl-mad-enable %s -D%s=1 -I%s",
+  options = g_strdup_printf("-w -Werror -cl-std=CL1.2 -cl-opt-disable %s -D%s=1 -I%s",
```
TurboGit: We should do that for Intel AMD but not Nvidia since it works there. I don't this we want Nvidia users to be imposed a huge slowdown if the bug is not there.

TurboGit: `if(!strncasecmp(vendor, "NVIDIA", 6))`

aurelienpierre: Even for Intel & AMD users, that's something I would prefer to avoid since the perf penalty is huge and everything else in dt seems to be working correctly. I will keep looking for the exact origin of the problem, and if I have no success in a couple of weeks, then we can fallback to this ugly hack.

TurboGit: Yes, I understand...

TurboGit:
> No, 2.1. The latest OpenCL version is 2.2 as of today.

Ok, so AMD and Intel are using more recent version of OpenCL. 

So this may be just an incompatibility with new OpenCL version and not something that Nvidia do properly. I mean, maybe Nvidia with OpenCL version 2.1 will also fails. Right? Or I'm still confused or missing something?

aurelienpierre:
> So this may be just an incompatibility with new OpenCL version and not something that Nvidia do properly. I mean, maybe Nvidia with OpenCL version 2.1 will also fails.

We can't know if Nvidia + OpenCL 2.1 works since it appears than all their drivers don't support OpenCL past 1.2.

But, at least for Intel, I know that forcing OpenCL 1.2 (putting `-cl-std=CL1.2` in building options) doesn't work either. There is some difference in the optimizations.

aurelienpierre:
Let it be said that Nvidia are fucking assholes for only supporting OpenCL 1.2, which is 8 years old now, since competitors (Intel, ARM and AMD), possibly with fewer resources (AMD), have supported OpenCL 2.1 for 2 years now.

aurelienpierre:
Ok, found it : it's the OpenCL dot product that fails for Intel. I have solved it roughly, I will finish tomorrow.

ptilopteri:
I am on openSUSE Tumbleweed with
libOpenCL1-2.2.12-1.1.x86_64
and NVIDIA GF106 [GeForce GTS 450] vendor: PNY driver: nvidia v: 390.129

and have no problem with
   darktable-2.7.0~git2049.91130e9c1-1709.1.x86_64

TurboGit:
@aurelienpierre : good finding !

cryptomilk:
I use AMD ROCm, which reports:

```
  Device Version                                  OpenCL 1.2 
  Driver Version                                  2982.0 (HSA1.1,LC)
  Device OpenCL C Version                         OpenCL C 2.0
```

src/common/opencl.c
cryptomilk: These options don't work for AMD ROCm. It doesn't understand `-cl-opt-disable`.

aurelienpierre: I would love it if vendors started to support **the full set of standard instructions**. :man_facepalming:

cryptomilk:
This works for me but the patch does a lot of things. I think splitting it would be beneficial :-)

TurboGit:
Merged manually. Closing this PR.

PR #3194, opened 2019-10-22
tone equalizer: improve tooltips and remove labels

aurelienpierre:
1. Improve the existing tooltips
1. Add some more
1. Remove the "highlights, mitones, etc." labels in the sliders labels: they were there as indications but users took them too seriously since they actually depend on the dynamic range of the picture and the mask post-processing settings.


TurboGit:
Thanks!

PR #3195, opened 2019-10-22
filmicrgb: remove presets hints about use

aurelienpierre:
again, the presets apply to dynamic range settings, and the use-case hints (landscape, studio, etc.) are merely indicative, but users take them too seriously and don't pay enough attention to the important info here : the EV number. So we remove the hints and possibly force users to RTFM.


TurboGit:
All good. Thanks!

PR #3196, opened 2019-10-22
just reformating

rabauke:
via clang-format


TurboGit:
We are in feature freeze, I'd like to avoid merging that now. I'll do it for 3.0.1.

TurboGit:
Thanks !

PR #3197, opened 2019-10-22
Hope to fix writing hdr dng files.

### LINKED ISSUE #3186: HDR merging should filter out ACTIVEAREA exif tag
darktable "create HDR" (from selected images)

**To Reproduce**
1. Go to lighttable and select an some Images 
2. Click on "create HDR"
3. HDR image is created but cant be vewed
4. on the command line 
```
[rawspeed] (IMG_20191018_200915.1-hdr.dng) void rawspeed::DngDecoder::handleMetadata(const rawspeed::TiffIFD*), line 441: Rectangle (0, 0, 2592, 1944) not inside image (0, 0, 2576, 1928).
allocation failed???
```

**Expected behavior**
be able to view the created HDR image

 - OS: Debian testing
 - Version 2.6.2-3

**Additional context**
https://github.com/darktable-org/rawspeed/issues/203

jenshannoschwalm:
Should fix #3186

Writing hdr-merged dng files is in some ways broken in current master, it might
work in many situations, in some it simply generates unreadable data.

There are some reasons for this, basically they are all based on `dt_exif_read_blob`
just copying found exif data instead of doing at least basic corrections.

1. If using dng files as input we **must** remove all tag information about cropping
 as ActiveArea, UserCrops and so on. If we don't do so image data is expected outside
 existing ranges and thus this image will simply not load.

2. If using dng files as input we also must remove Opcode lists and tags defining the
 embedded color matrix.

Both points have been fixed in exif.cc

One more problem has been around, a dng file must have at least one pair of
CalibrationIlluminant / ColorMatrix.

That was **not** added at all, instead we relied on Exif.Photo.ColorSpace which
will be ok for many dng readers and in fact dt worked fine but it's for sure not
as in the specification. This has been included in imgeio_dng.h

### DIFF
--- a/src/common/exif.cc
+++ b/src/common/exif.cc
@@ -1300,11 +1300,9 @@ int dt_exif_read_blob(uint8_t **buf, const char *path, const int imgid, const in
 
     // get rid of thumbnails
     Exiv2::ExifThumb(exifData).erase();
+    Exiv2::ExifData::const_iterator pos;
 
-    // ufraw-style exif stripping:
-    Exiv2::ExifData::iterator pos;
     {
-    /* Delete original TIFF data, which is irrelevant*/
       static const char *keys[] = {
         "Exif.Image.ImageWidth",
         "Exif.Image.ImageLength",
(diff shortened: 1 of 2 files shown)


TurboGit:
This broke the final DNG orientation. Take a portrait picture. Click "Create HDR" and you'll see the image created with landscape orientation.

jenshannoschwalm:
@TurboGit fixed now.

TurboGit:
Looks good now! Thanks.

PR #3201, opened 2019-10-23
lock db at high level **per image**

jenshannoschwalm:
This pr **fixes the responsiveness issue** related to #3181, see discussion there.

The first attempt fixed a number of issues related to threading and database
access including history changes.

Basically it was the problem, that multiple threads access & modify the database in parallel,
so as an example one thread was updating the mipmap cache in the background (and thus possibly
modify the database) and the user accessed an image for development at the same time
--> database & history for an image were undefined or even broken.

The SQLite works in serialized mode, so itself is threading safe.

The bad responsiveness is the result of a global lock or the "single mutex rule".

This design can be improved, what we need is **not a global database lock** but a
**lock for a specific image** in the database. How can we achieve this?

We can't have a recursive mutex for every image, but we can map all images on a number
of mutexes, the number of mutexes should be high enough to cover multi-core cpus
and a large display with a large number of thumbnails displayed in lighttable mode.

I chose 64 recursive mutexes and the mapping goes as (imgid & 63)
These mutexes are initialised and destroyed in darktable.c

The database lock&unlock for an image are defined as new functions and replace the direct
use of the darktable.db_insert mutex.
`void dt_database_lock_image(u_int32_t imgid);`
`void dt_database_unlock_image(u_int32_t imgid);`

darktable.db_insert mutex is now used only for film locking.

@TurboGit @parafin you were both deeply involved in the discussion ... I have this running here since yesterday and all tests were fine without a crash but with **much** better user experience, almost without a delay. I know it's certainly not a bug fix ...

### DIFF
--- a/src/common/darktable.h
+++ b/src/common/darktable.h
@@ -264,9 +264,6 @@ typedef struct darktable_t
   struct dt_colorspaces_t *color_profiles;
   struct dt_l10n_t *l10n;
   dt_pthread_mutex_t db_image[DT_IMAGE_DBLOCKS];
-  pthread_mutexattr_t db_image_attr[DT_IMAGE_DBLOCKS];  
-  dt_pthread_mutex_t db_film;
-  dt_pthread_mutex_t db_multi_image;
   dt_pthread_mutex_t plugin_threadsafe;
   dt_pthread_mutex_t capabilities_threadsafe;
   dt_pthread_mutex_t exiv2_threadsafe;
--- a/src/common/darktable.h
+++ b/src/common/darktable.h
@@ -39,6 +39,8 @@
 #define O_BINARY 0
 #endif
 
+#include "ThreadSafetyAnalysis.h"
+
 #ifdef HAVE_CONFIG_H
 #include "config.h"
 #endif
@@ -305,6 +307,36 @@ void dt_free_align(void *mem);
 #define dt_free_align(A) free(A)
 #define dt_free_align_ptr free
 #endif
+
+static inline void dt_lock_image(uint32_t imgid) ACQUIRE(darktable.db_image[imgid & (DT_IMAGE_DBLOCKS-1)])
+{
+  dt_pthread_mutex_lock(&(darktable.db_image[imgid & (DT_IMAGE_DBLOCKS-1)]));
+}
+
+static inline void dt_unlock_image(uint32_t imgid) RELEASE(darktable.db_image[imgid & (DT_IMAGE_DBLOCKS-1)])
+{
+  dt_pthread_mutex_unlock(&(darktable.db_image[imgid & (DT_IMAGE_DBLOCKS-1)]));
+}
+
+static inline void dt_lock_image_pair(uint32_t imgid1, uint32_t imgid2) ACQUIRE(darktable.db_image[imgid1 & (DT_IMAGE_DBLOCKS-1)], darktable.db_image[imgid2 & (DT_IMAGE_DBLOCKS-1)])
+{
+  if(imgid1 < imgid2)
+  {
+    dt_pthread_mutex_lock(&(darktable.db_image[imgid1 & (DT_IMAGE_DBLOCKS-1)]));
+    dt_pthread_mutex_lock(&(darktable.db_image[imgid2 & (DT_IMAGE_DBLOCKS-1)]));
+  }
+  else
+  {
+    dt_pthread_mutex_lock(&(darktable.db_image[imgid2 & (DT_IMAGE_DBLOCKS-1)]));
+    dt_pthread_mutex_lock(&(darktable.db_image[imgid1 & (DT_IMAGE_DBLOCKS-1)]));
+  }
+}
+
+static inline void dt_unlock_image_pair(uint32_t imgid1, uint32_t imgid2) RELEASE(darktable.db_image[imgid1 & (DT_IMAGE_DBLOCKS-1)], darktable.db_image[imgid2 & (DT_IMAGE_DBLOCKS-1)])
+{
+  dt_pthread_mutex_unlock(&(darktable.db_image[imgid1 & (DT_IMAGE_DBLOCKS-1)]));
+  dt_pthread_mutex_unlock(&(darktable.db_image[imgid2 & (DT_IMAGE_DBLOCKS-1)]));
+}
 
 static inline gboolean dt_is_aligned(const void *pointer, size_t byte_count)
 {
--- a/src/common/darktable.h
+++ b/src/common/darktable.h
@@ -266,6 +266,7 @@ typedef struct darktable_t
   dt_pthread_mutex_t db_image[DT_IMAGE_DBLOCKS];
   pthread_mutexattr_t db_image_attr[DT_IMAGE_DBLOCKS];  
   dt_pthread_mutex_t db_film;
+  dt_pthread_mutex_t db_multi_image;
   dt_pthread_mutex_t plugin_threadsafe;
   dt_pthread_mutex_t capabilities_threadsafe;
   dt_pthread_mutex_t exiv2_threadsafe;
--- a/src/common/darktable.h
+++ b/src/common/darktable.h
@@ -185,6 +185,8 @@ static inline int dt_version()
 #undef STR
 #define STR(x) STR_HELPER(x)
 
+#define DT_IMAGE_DBLOCKS 64
+
 struct dt_gui_gtk_t;
 struct dt_control_t;
 struct dt_develop_t;
@@ -261,6 +263,8 @@ typedef struct darktable_t
   struct dt_undo_t *undo;
   struct dt_colorspaces_t *color_profiles;
   struct dt_l10n_t *l10n;
+  dt_pthread_mutex_t db_image[DT_IMAGE_DBLOCKS];
+  pthread_mutexattr_t db_image_attr[DT_IMAGE_DBLOCKS];  
   dt_pthread_mutex_t db_insert;
   dt_pthread_mutex_t plugin_threadsafe;
   dt_pthread_mutex_t capabilities_threadsafe;
--- a/src/common/darktable.h
+++ b/src/common/darktable.h
@@ -265,7 +265,7 @@ typedef struct darktable_t
   struct dt_l10n_t *l10n;
   dt_pthread_mutex_t db_image[DT_IMAGE_DBLOCKS];
   pthread_mutexattr_t db_image_attr[DT_IMAGE_DBLOCKS];  
-  dt_pthread_mutex_t db_insert;
+  dt_pthread_mutex_t db_film;
   dt_pthread_mutex_t plugin_threadsafe;
   dt_pthread_mutex_t capabilities_threadsafe;
   dt_pthread_mutex_t exiv2_threadsafe;
--- a/src/common/film.c
+++ b/src/common/film.c
@@ -200,7 +200,6 @@ int dt_film_new(dt_film_t *film, const char *directory)
                                 -1, &stmt, NULL);
     DT_DEBUG_SQLITE3_BIND_TEXT(stmt, 1, datetime, -1, SQLITE_STATIC);
     DT_DEBUG_SQLITE3_BIND_TEXT(stmt, 2, directory, -1, SQLITE_STATIC);
-    dt_pthread_mutex_lock(&darktable.db_film);
     rc = sqlite3_step(stmt);
     if(rc != SQLITE_DONE)
       fprintf(stderr, "[film_new] failed to insert film roll! %s\n",
@@ -211,7 +210,6 @@ int dt_film_new(dt_film_t *film, const char *directory)
     DT_DEBUG_SQLITE3_BIND_TEXT(stmt, 1, directory, -1, SQLITE_STATIC);
     if(sqlite3_step(stmt) == SQLITE_ROW) film->id = sqlite3_column_int(stmt, 0);
     sqlite3_finalize(stmt);
-    dt_pthread_mutex_unlock(&darktable.db_film);
   }
 
   if(film->id <= 0) return 0;
--- a/src/common/film.c
+++ b/src/common/film.c
@@ -200,7 +200,7 @@ int dt_film_new(dt_film_t *film, const char *directory)
                                 -1, &stmt, NULL);
     DT_DEBUG_SQLITE3_BIND_TEXT(stmt, 1, datetime, -1, SQLITE_STATIC);
     DT_DEBUG_SQLITE3_BIND_TEXT(stmt, 2, directory, -1, SQLITE_STATIC);
-    dt_pthread_mutex_lock(&darktable.db_insert);
+    dt_pthread_mutex_lock(&darktable.db_film);
     rc = sqlite3_step(stmt);
     if(rc != SQLITE_DONE)
       fprintf(stderr, "[film_new] failed to insert film roll! %s\n",
@@ -211,7 +211,7 @@ int dt_film_new(dt_film_t *film, const char *directory)
     DT_DEBUG_SQLITE3_BIND_TEXT(stmt, 1, directory, -1, SQLITE_STATIC);
     if(sqlite3_step(stmt) == SQLITE_ROW) film->id = sqlite3_column_int(stmt, 0);
     sqlite3_finalize(stmt);
-    dt_pthread_mutex_unlock(&darktable.db_insert);
+    dt_pthread_mutex_unlock(&darktable.db_film);
   }
 
   if(film->id <= 0) return 0;
--- a/src/common/history.c
+++ b/src/common/history.c
@@ -67,7 +67,7 @@ void dt_history_delete_on_image_ext(int32_t imgid, gboolean undo)
     dt_history_snapshot_undo_create(hist->imgid, &hist->before, &hist->before_history_end);
   }
 
-  dt_database_lock_image(imgid);
+  dt_pthread_mutex_lock(&(darktable.db_image[imgid & (DT_IMAGE_DBLOCKS-1)]));
 
   sqlite3_stmt *stmt;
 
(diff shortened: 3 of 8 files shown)


jenshannoschwalm:
Don't understand why travis doesn't build the man pages ...

parafin:
Idea is quite nice, but it creates problems where you have to lock 2 images. There you changed one lock to 2 locks, and it can deadlock, e.g. if one thread is doing copy-paste from image 1 to 2, and the second from 2 to 1.

parafin:
Not sure how to work around that, but, as you can see, breaking "one mutex rule" has consequences. In this case recursion has nothing to do with it, since locks are in the same function and it's classic multiple resource locking problem which can be at least easily seen during code review. But because of recursive mutexes there might other possible deadlocks too, now it's quite hard to tell.

parafin:
OK, I have a workaround for the particular deadlock I described, but, as I said, since you now allow recursive locking, it's basically impossible to tell that there are no other deadlocks present in the code.

Workaround goes like that - create dt_database_lock_two_images function that lock the mutexes in the ascending order. So both dt_database_lock_two_images(1,2) and dt_database_lock_two_images(2,1) should call dt_database_lock_image(1) then dt_database_lock_image(2).

To re-iterate once again - I think enabling recursive mutexes was a bad design choice and this PR is now probably unsafe in any form.

jenshannoschwalm:
@parafin thanks for pinpointing to the deadlock problem :-)

As far as "recursive mutexes impose more deadlocking problems" i really disagree.
It just means: take care about n-locks pair n-unlocks instead of 1-locks pair 1-unlocks. 

About the remaining db_insert mutex, it's just used for films so it probably should be renamed db_film and it also doesn't need to be recursive.

jenshannoschwalm:
When you talk about "bad design" do you have any idea how to do it otherwise?
How else can we make sure thread(a) can read a complete set of data for an image while thread(b) is happily modifying it?

TurboGit:
Looks promising to me. I have indeed seen some performance penalty. So maybe this is more a bug fix than a feature. I'll tag accordingly. I have no time right now to have a look.

TurboGit:
> Because to me introduction of recursive mutex for DB meant "I can't keep track of the call stack, so I give up and now can lock at any place", which implies you can't basically guarantee anything.

Right, with an application like dt with signals from GUI and events triggerred here and there I don't see how to predict the full locking stack. So yes, call it "bad design" if you like, for me it is perfectly fine. There is no way the same thread should be locked because it tries to get the same lock at different layers in the application. The layers are in other part of the code called by different routines in different context. So again to me this is a sensible choice to allow recursive locking.

TurboGit:
Oh and don't get me wrong. I'm not saying it is perfect, the perfect solution would be to have a team step in and propose to redo good part of the code handling the database and especially the part handling the history. But until this happens we'll have to live with the current situation. Personally I can't do better today.

TurboGit:
> About the remaining db_insert mutex, it's just used for films so it probably should be renamed db_film and it also doesn't need to be recursive.

Probably right, yes.

parafin:
Call me idealist or perfectionist then;) I'm just warning of potential consequences, increasing technological debt (and line count) makes it harder to deal with later. If benefits outweighs the risks right now, that's OK then, but I think we should be aware of the both sides of the argument, that's all.

jenshannoschwalm:
> Are you sure, that in the whole call stack after first dt_database_lock_image you will lock only the same one (or two for suggested dt_database_lock_two_images) imgids?

Looking into that was indeed the time-spending job. In fact there is one problematic function.
`dt_history_copy_and_paste_on_image` At least from how i understand it's used we can't deadlock. Anyway i will work on that.

parafin:
Also you broke CI:
```
/build/darktable/src/common/database.c:2225:1: error: mutex 'darktable.db_image[imgid & (64 - 1)]' is still held at the end of function [-Werror,-Wthread-safety-analysis]
}
^
/build/darktable/src/common/database.c:2224:3: note: mutex acquired here
  dt_pthread_mutex_lock(&darktable.db_image[imgid & (DT_IMAGE_DBLOCKS-1)]);
  ^
/build/darktable/src/common/database.c:2229:3: error: releasing mutex 'darktable.db_image[imgid & (64 - 1)]' that was not held [-Werror,-Wthread-safety-analysis]
  dt_pthread_mutex_unlock(&darktable.db_image[imgid & (DT_IMAGE_DBLOCKS-1)]);
  ^
2 errors generated.
```

parafin:
You should look at how dt_pthread_mutex_* functions are declared in dtpthread.h and do the same for dt_database_(un)lock_image. So fix CI.

jenshannoschwalm:
Any better suggestion than using the BAD variants? maybe a #define?

parafin:
Yeah, that’s not what I meant. Try adding ACQUIRE() and RELEASE() to the declaration of your functions (in the header) and maybe NO_THREAD_SAFETY_ANALYSIS to the definition (in the cpp file). See more here: https://clang.llvm.org/docs/ThreadSafetyAnalysis.html

src/common/darktable.h
```diff
@@ -261,7 +263,10 @@ typedef struct darktable_t
   struct dt_undo_t *undo;
   struct dt_colorspaces_t *color_profiles;
   struct dt_l10n_t *l10n;
-  dt_pthread_mutex_t db_insert;
+  dt_pthread_mutex_t db_image[DT_IMAGE_DBLOCKS];
+  pthread_mutexattr_t db_image_attr[DT_IMAGE_DBLOCKS];
```
TurboGit: I think we should avoid this declaration. Have a single pthread_mutexattr_t var and reuse it for all mutex. After all we are using the same attribute for all mutex.

src/common/history.c
```diff
@@ -760,8 +765,9 @@ int dt_history_copy_and_paste_on_image(int32_t imgid, int32_t dest_imgid, gboole
     return 1;
   }
 
-  // Just in case lock the database
-  dt_pthread_mutex_lock(&darktable.db_insert);
+  dt_pthread_mutex_lock(&darktable.db_multi_image);
+  dt_database_lock_image(imgid);
```
TurboGit: This one IS dangerous. Can't we have a scenario where we have the opposite lock scenario?

```
   dt_database_lock_image(des_imgid);
   dt_database_lock_image(imgid);
```

src/common/history.c
```diff
@@ -1114,7 +1125,9 @@ static void _history_reorder(int32_t imgid)
     if (give_reorder_information) fprintf(stderr,", reorder\n");
     // make sure running jobs can't interfere here as the followiing code uses a fixed dummy id
     // and also intends to have a "properly" orderered database
-    dt_pthread_mutex_lock(&darktable.db_insert);
+    dt_pthread_mutex_lock(&darktable.db_multi_image);
+    dt_database_lock_image(imgid);
+    dt_database_lock_image(dummy);
```
TurboGit: Likewise, dangerous.

src/common/history.c
parafin: I've already covered this issue. @jenshannoschwalm tries to solves this problem by adding one more mutex - db_multi_image. It seems to solve the deadlock between imgid and dest_imgid, but it may introduce other deadlocks down the call stack, I'm not sure here. My proposed workaround didn't add new mutexes. I really don't think it's a good idea to solve the problems of multiple mutex deadlock by adding more mutexes;)

parafin: Also my proposed solution doesn't introduce another global image lock after we just removed one;)

src/common/history.c
```diff
@@ -67,7 +67,7 @@ void dt_history_delete_on_image_ext(int32_t imgid, gboolean undo)
     dt_history_snapshot_undo_create(hist->imgid, &hist->before, &hist->before_history_end);
   }
 
-  dt_pthread_mutex_lock(&darktable.db_insert);
+  dt_database_lock_image(imgid);
```
parafin: Why this lock is needed at all?

src/common/film.c
```diff
@@ -200,7 +200,7 @@ int dt_film_new(dt_film_t *film, const char *directory)
                                 -1, &stmt, NULL);
     DT_DEBUG_SQLITE3_BIND_TEXT(stmt, 1, datetime, -1, SQLITE_STATIC);
     DT_DEBUG_SQLITE3_BIND_TEXT(stmt, 2, directory, -1, SQLITE_STATIC);
-    dt_pthread_mutex_lock(&darktable.db_insert);
+    dt_pthread_mutex_lock(&darktable.db_film);
```
parafin: Why this lock is needed at all?

parafin:
It would be a good idea to create some kind of report describing all functions accessing DB, their dependencies (who calls who) and where locks are placed. I think it's the only way to understand how it's supposed to work without every reviewer doing that investigation by himself. I assume it was done by @jenshannoschwalm at least at some level, since some assumptions about possible and impossible call stacks are being made in these PRs. This is a suggestion to improve code quality in darktable, if anyone is interested in that.

src/common/history.c
jenshannoschwalm: I think your suggestion does not work. Binding two mutex locks/unlocks in a function one after another does not help as jobA(1,3) and jobB(3,1) would possible deadlock. Either check visually by code analysis :-) or prepend a global mutex in **all** functions that will lock two images. That's my idea.

src/common/film.c
jenshannoschwalm: Right. Probably someone feeling that database safety work is pending? :-)

jenshannoschwalm:
> Yeah, that’s not what I meant. Try adding ACQUIRE() and RELEASE() to the declaration of your functions (in the header) and maybe NO_THREAD_SAFETY_ANALYSIS to the definition (in the cpp file). See more here: https://clang.llvm.org/docs/ThreadSafetyAnalysis.html

thanks for that link. i confess i never heard of this before (you have guessed so :-)

Will take some time reading, atm i think it has been a bad idea to introduce those functions as they will block compiler security checking. Thanks for introduction!

src/common/history.c
jenshannoschwalm: Rethinking here, i still think your proposal is not ok but introducing a global lock is just bad. Need to find another way. What about this?

We have a pool of known image id's that are not used by dt itself. (lets say last 64imgs in the uint32 range.) By trylocking we can find a not used id.

If we have a multi-image function using img1 and img2 we lock img1, get a still unlocked and right now locked imgx from pool, copy database fom img1->imgx, unlock img1, lock img2, continue by using imgx instead of img1, and at the end of the function unlock img2 and imgx.

jenshannoschwalm:
> It would be a good idea to create some kind of report describing all functions accessing DB, their dependencies (who calls who) and where locks are placed.

I fully agree here and "yes" i made myself somewhat "at home" with history and database code and will **not** pass work to someone else. Will take a bit of time though, don't know yet what should be included, lets say Mid November.

I will fix the commented stuff tonight.

After that i will be not available for 10d, (visiting my mother-in-law). As the code is working stable for me, (no crashes, halts or database corruptions) i would kindly ask you to test code as is to learn about maybe overseen problems.
Fore sure there are debatable issues pending and the quality of code might not be as good as required

I you want to continue work on this stuff just go ahead and take everything or wait ...

TurboGit:
> It would be a good idea to create some kind of report describing all functions accessing DB, their dependencies (who calls who) and where locks are placed.

If I agree in general on such analysis, if it is not automatic then it is a waste of time. After code refactoring and some merge this will be completely wrong again and will require another manual analysis. And I don't know if we can create such data automatically ?

TurboGit:
> After that i will be not available for 10d, (visiting my mother-in-law). As the code is working stable for me, (no crashes, halts or database corruptions) i would kindly ask you to test code as is to learn about maybe overseen problems.

@jenshannoschwalm : you mean the code on this PR right? Just to be sure.

The mid-november date worries me, at this time I'd like to avoid any large fixes if possible. So I would be in favor of merging this after your tonight fix and we will see if we can do better later.

Anyway, ping me and I'll review. Thanks again for all the work on this tricky area.

parafin:
> > It would be a good idea to create some kind of report describing all functions accessing DB, their dependencies (who calls who) and where locks are placed.
> 
> If I agree in general on such analysis, if it is not automatic then it is a waste of time. After code refactoring and some merge this will be completely wrong again and will require another manual analysis. And I don't know if we can create such data automatically ?

Automatic would be nice indeed. But I disagree that manual analysis is useless. First of all it will prove that this particular PR doesn't break anything (assuming that analysis was done correctly). Secondly later it can be referred to, even if outdated, or even better - update this documentation every time changes to DB handling are made. Any description of internal darktable's workings are welcome IMHO.

src/common/history.c
parafin: dummy doesn't have its own lock, it shares it with 1/64 of all images.

src/common/history.c
parafin: Please stop complicating the logic. You won't solve the issue by making it harder to understand.
My proposal works. Please re-read it and think of counter-example if you still think it doesn't work.

src/common/film.c
parafin: A lot of other more dangerous stuff is done unlocked (e.g. removing a image), but this on the contrary looks safe to me...

jenshannoschwalm:
@TurboGit @parafin this is as far as i could get.

src/common/history.c
```diff
@@ -1095,45 +1112,35 @@ void dt_history_compress_on_image(int32_t imgid)
...
-  else
-  {
-    if (give_reorder_information) fprintf(stderr,", reorder\n");
-    // make sure running jobs can't interfere here as the followiing code uses a fixed dummy id
-    // and also intends to have a "properly" orderered database
-    dt_pthread_mutex_lock(&darktable.db_insert);
+  // make sure running jobs can't interfere here as the followiing code uses a fixed dummy id
+  // and also intends to have a "properly" orderered database
+  dt_pthread_mutex_lock(&(darktable.db_image[imgid & (DT_IMAGE_DBLOCKS-1)]));
+  dt_pthread_mutex_lock(&(darktable.db_image[dummy & (DT_IMAGE_DBLOCKS-1)]));
```
parafin: same ascending ordering must be applied here too

parafin:
Yeah, wrapper functions looked much better... We need to figure out how to please clang here.

src/common/history.c
parafin: or maybe not actually. since dummy & (DT_IMAGE_DBLOCKS-1) is a constant, it is already ordered lock. OK then.

jenshannoschwalm:
> Yeah, wrapper functions looked much better... We need to figure out how to please clang here

First part, yes, maybe #defines? Second part, i don't know, will see what you teach me :-)

parafin:
These wrapper functions work:
```diff
diff --git a/src/common/darktable.h b/src/common/darktable.h
index 3c451e934..7f8e48cac 100644
--- a/src/common/darktable.h
+++ b/src/common/darktable.h
@@ -39,6 +39,7 @@
 #define O_BINARY 0
 #endif
 
+#include "ThreadSafetyAnalysis.h"
 #ifdef HAVE_CONFIG_H
 #include "config.h"
 #endif
@@ -306,6 +307,36 @@ void dt_free_align(void *mem);
 #define dt_free_align_ptr free
 #endif
 
+static inline void dt_lock_image(uint32_t imgid) ACQUIRE(darktable.db_image[imgid & (DT_IMAGE_DBLOCKS-1)])
+{
+  dt_pthread_mutex_lock(&(darktable.db_image[imgid & (DT_IMAGE_DBLOCKS-1)]));
+}
+
+static inline void dt_unlock_image(uint32_t imgid) RELEASE(darktable.db_image[imgid & (DT_IMAGE_DBLOCKS-1)])
+{
+  dt_pthread_mutex_unlock(&(darktable.db_image[imgid & (DT_IMAGE_DBLOCKS-1)]));
+}
+
+static inline void dt_lock_image_pair(uint32_t imgid1, uint32_t imgid2) ACQUIRE(darktable.db_image[imgid1 & (DT_IMAGE_DBLOCKS-1)], darktable.db_image[imgid2 & (DT_IMAGE_DBLOCKS-1)])
+{
+  if(imgid1 < imgid2)
+  {
+    dt_pthread_mutex_lock(&(darktable.db_image[imgid1 & (DT_IMAGE_DBLOCKS-1)]));
+    dt_pthread_mutex_lock(&(darktable.db_image[imgid2 & (DT_IMAGE_DBLOCKS-1)]));
+  }
+  else
+  {
+    dt_pthread_mutex_lock(&(darktable.db_image[imgid2 & (DT_IMAGE_DBLOCKS-1)]));
+    dt_pthread_mutex_lock(&(darktable.db_image[imgid1 & (DT_IMAGE_DBLOCKS-1)]));
+  }
+}
+
+static inline void dt_unlock_image_pair(uint32_t imgid1, uint32_t imgid2) RELEASE(darktable.db_image[imgid1 & (DT_IMAGE_DBLOCKS-1)], darktable.db_image[imgid2 & (DT_IMAGE_DBLOCKS-1)])
+{
+  dt_pthread_mutex_unlock(&(darktable.db_image[imgid1 & (DT_IMAGE_DBLOCKS-1)]));
+  dt_pthread_mutex_unlock(&(darktable.db_image[imgid2 & (DT_IMAGE_DBLOCKS-1)]));
+}
+
 static inline gboolean dt_is_aligned(const void *pointer, size_t byte_count)
 {
     return (uintptr_t)pointer % byte_count == 0;
```
I suggest to use dt_(un)lock_image_pair for dummy imgid too, just to avoid inconsistency in the code.

jenshannoschwalm:
ok, last commit on this pr for me.

TurboGit:
@jenshannoschwalm : I'll test tomorrow and will come back to you if any issue, otherwise I'll merge this for 3.0.

@parafin : thanks also to you for the detailed review and proposals.

TurboGit:
Very nicer user's experience with this PR! Lot less lag, quite responsive in massive changes in large set of images (test on 100). In lt or filmstrip. Also, I did not get crashes or history corruption. Nice job! Thanks.

PR #3226, opened 2019-10-27
LR import: Code simplification + Fix for rotation angle

leclercj:


TurboGit:
Looks quite simpler indeed ! Will land in 3.0.1 now. Thanks.

leclercj:
New commit added to fix https://github.com/darktable-org/darktable/issues/3237

src/develop/lightroom.c
```diff
@@ -1020,6 +1021,48 @@ static void _handle_xpath(dt_develop_t *dev, xmlDoc *doc, int imgid, xmlXPathCon
     }
 }
 
+static inline float dabs(double a)
```
TurboGit: This is standard fabs().

leclercj: You're right. I'll modify that

TurboGit:
Thanks for fixing that for next release! I'm not sure how it gets broken over the time but this is really welcome! Will merge now.

Issue #3206, opened 2019-10-24
Tag export: "omit hierarchy" behavior can't actually be disabled


junkyardsparkle:
EDIT: the actual issue has been [summarized below](https://github.com/darktable-org/darktable/issues/3206#issuecomment-547623079), since I was initially confused about the new intended behavior.

**Describe the bug**

When exporting an image which has hierarchical tags applied, and the "omit hierarchy" checkbox is NOT selected in the export settings, all tags within the branch are still not included as individual tags.

**To Reproduce**

Start darktable with new profile, import an image, add a tag such as 'foo|blah|baz|bar', export image using default JPEG settings. Observe XMP Subject field in exported file contains only 'bar'. No combination of options selected in export settings would change this for me.

**Expected behavior**

Include all elements of hierarchical tags when "omit hierarchy" is not selected.

**Platform (please complete the following information):**

Linux, darktable 2.7.0+2075, exiv2 0.26 (XMP Core 4.4.0-Exiv2)

**Additional context**

In fact, I was having more complicated, harder-to-reproduce issues under real-world conditions when trying to export images with multiple hierarchical tags. In some cases, only the first and last element of the "path" were included. In a few cases, all were. None of these tags has been set as "private" or "category" or any other special class, as far as I can tell (and selecting those boxes didn't help). This is my attempt to provide a simple test that can be reproduced, in the hopes that it will be enough to start an investigation. ;-)

If anyone can NOT reproduce this, could you tell what exiv2 version you build against? Thanks!

junkyardsparkle:
I finally updated my system to exiv2 0.27.2, and it didn't fix the issue.

Here's a more methodical procedure I just followed on a clean, untouched profile:
Import image (with existing XML sidecar having tags 'something' and 'foo|blah|baz|bar'); tags are displayed correctly in module view. Without changing default settings in export module, export image. The JPEG contains neither 'Subject' nor 'Hierarchical Subject' XML tags, even though 'tags' checkbox is selected by default. 

Select the 'Hierarchical tags' checkbox and export again. JPEG now has both tags, but 'Subject' tag contains only 'bar', 'exported' and 'something' keywords, even though 'Omit hierarchical' is unselected by default. The 'Hierachical Subject' tag contains all full keywords as expected.

Select the 'omit hierarchy' checkbox, export, same results as above.

Se;ect the 'synonyms' checkbox, export, same results as above.

Select the 'private tags' checkbox, export, same results as above.

[...]

PR #3238, opened 2019-10-30
fix #3203 (scroll) + #3206 (dt tags + default for hierarchical tags)

### LINKED ISSUE #3203: Size of tagging attached windows and dictionary window
With darktable 2.7, it is possible to resize the two windows of the "tagging" module with a "CTRL+wheel". It works very well on my laptop, but not on my desktop computer. On the latter device, the CTRL+wheel only enlarges the windows. Impossible to reduce them. The only way to restore the windows to their original size is to reset these lines:
`plugins/lighttable/tagging/heightattachedwindow=100`
and
`plugins/lighttable/tagging/heightdictionarywindow=100`
in darktablerc.
I continued my investigations and found out. It's related to my screen. I'm lucky (?) to have a screen that, if it's not 4K, still has a fairly high pitch, higher than the darktable reference of 96dpi. My 27" screen displays 2560x1440 pxl (WQHD) or a pitch of about 106dpi.

In src/libs/tagging.c two functions manage the resizing of these windows:
`static gboolean mouse_scroll_attached_mouse()`
and
`static gboolean mouse_scroll_dictionary()`

In each function, the code that calculates and applies the new window size is:
```
height = height + (10.0 * event->delta_y);
height = (height < 100.0) ? 100.0 : (height > 500.0) ? 500.0 : height;
gtk_widget_set_size_request(GTK_WIDGET(d->attached_window), -1, DT_PIXEL_APPLY_DPI((gint)height));
```
It is the same code for both functions, except the name of the window and the value of the constants that different.

The problem comes from the application of the macro DT_PIXEL_APPLY_DPI() on the calculated value. In my case, since the multiplier of the macro is higher than the default value, the calculated value is always increased and never decreased.

In fact the macro should be applied to the constants, not to the calculated value, which corresponds strictly to the logic of this macro.

Here is the code I wrote to modify these functions:
```
static gboolean mouse_scroll_attached(GtkWidget *treeview, GdkEventScroll *event, dt_lib_module_t *self)
{
 dt_lib_tagging_t *d = (dt_lib_tagging_t *)self->data;
 if (event->state & GDK_CONTROL_MASK)
 {
   gint width, height;
   gint increment = DT_PIXEL_APPLY_DPI(10.0);
   gint min_height = DT_PIXEL_APPLY_DPI(100.0);
   gint max_height = DT_PIXEL_APPLY_APPLY_DPI(500.0);
   gtk_widget_get_size_request(GTK_WIDGET(d->attached_window), &width, &height);
   height = height + increment * event->delta_y;
   height = (height < min_height) ? min_height : (height > max_height) ? max_height : height;
   gtk_widget_set_size_request(GTK_WIDGET(d->attached_window), -1, height);
   dt_conf_set_set_int("plugins/lighttable/tagging/heightattachedwindow", (gint)height);
   return TRUE;
 }
 return FALSE;
}
```
The second function is strictly identical except the max constant, the window name and the key for darktablerc

With this code, everything works! I tested on my WQHD screen of course, but also on an HD screen for which there was no problem anyway.

phweyland:
should fix 2 bugs:
- ctrl-scroll for 4k screen (as suggested in #3203)
- just created dt tags are exported (seen in #3206 3206)

comes back to previous logic about path elements in hierarchical tags: considered as keywords by default instead of category (#3206). If the leave tag is set as category, then all the path elements are also seen as category.
The flag "omit hierarchy" is also more logical with that change. 
This should avoid to break former habits about exported tags (Subject).

### DIFF
--- a/src/common/tags.c
+++ b/src/common/tags.c
@@ -162,6 +162,12 @@ gboolean dt_tag_new(const char *name, guint *tagid)
   }
   sqlite3_finalize(stmt);
 
+  if(g_strstr_len(name, -1, "darktable|") == name)
+  {
+    // clear darktable tags list
+    DT_DEBUG_SQLITE3_EXEC(dt_database_get(darktable.db), "DELETE FROM memory.darktable_tags", NULL, NULL, NULL);
+  }
+
   DT_DEBUG_SQLITE3_PREPARE_V2(dt_database_get(darktable.db), "INSERT INTO data.tags (id, name) VALUES (NULL, ?1)",
                               -1, &stmt, NULL);
   DT_DEBUG_SQLITE3_BIND_TEXT(stmt, 1, name, -1, SQLITE_TRANSIENT);
@@ -852,6 +858,13 @@ GList *dt_tag_get_hierarchical(gint imgid)
   return tags;
 }
 
+static gint is_tag_a_category(gconstpointer a, gconstpointer b)
+{
+  dt_tag_t *ta = (dt_tag_t *)a;
+  dt_tag_t *tb = (dt_tag_t *)b;
+  return ((g_strcmp0(ta->tag, tb->tag) == 0) && ((ta->flags | tb->flags) & DT_TF_CATEGORY)) ? 0 : -1;
+}
+
 GList *dt_tag_get_list_export(gint imgid, int32_t flags)
 {
   GList *taglist = NULL;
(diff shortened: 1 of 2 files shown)


phweyland:
The path elements topic was to make the demo and can be removed from the PR if not desirable...

src/common/tags.c
```diff
@@ -162,6 +162,12 @@ gboolean dt_tag_new(const char *name, guint *tagid)
   }
   sqlite3_finalize(stmt);
 
+  if(g_strstr_len(name, -1, "darktable|") == name)
```
TurboGit: I fail to understand this part. in dt_tag_new() you clear the memory.dartable_tags table if the tag starts with "darktable|". Right?

Can you add some comments to explain the logic ? No clear to me today and will certainly raises questions later when reviewing/maintaining this part of the code. Thanks.

TurboGit:
Thanks!

jpverrue:
@phweyland 
In lines 2050 and 2069 you left DT_PIXEL_APPLY_DPI(). 
I was able to test it today, it doesn't work with 4K displays.

I had proposed:
`gtk_widget_set_size_request(GTK_WIDGET(d->attached_window), -1, (gint)height);`
It works well.

phweyland:
Sorry. I'll look at it asap.

phweyland:
You are correct, I'm applying twice the macro. I'll issue a new PR for that. Thanks.

PR #3193, opened 2019-10-22
avoid mixed precision floating point number arithmetic

rabauke:
... and avoid unnecessary explicit casts

PR #3242, opened 2019-10-30
fix spline interpolation in color zones module

rabauke:
This is my 2nd attempt to fix interpolation routines in the color zones module, see  #2641 and #2533. Periodic boundary conditions must be applied when color is selected by hue in the color zones module. These periodic boundary conditions cannot be realized by tweaking the interpolation nodes, the interpolation routines themselves must obey periodic boundary conditions. Not doing so leads to various bugs, see issue #2533 and issue #2641.

Interpolation routines have been reimplemented from scratch and the color zones module has been revised. This bug fix will **not** break old edits, i.e., old edits will continue to use the old interpolation routines. The behavior of the revised color zones module depends on if the user works on an old edit or a new one. Main differences between the new and the old mode

- The old mode required two nodes at the extreme left/right ends of the hue (or saturation or brightness) scale. Nodes can be moved freely without any constraints in the new mode.
- The old mode employed two invisible interpolation nodes that can not be modified by the user directly. These two ghost nodes are no longer used/needed in the new mode.

This revised version of the color zones module introduces a new module parameters version. Do not test this with your production data base.

In order to ease review, I can split this pull request into several pull requests, which include
- pure reformating
- minor enhancements not directly related to the revision of the interpolation routines
- new interpolation routines

### DIFF
--- a/src/gui/draw.h
+++ b/src/gui/draw.h
@@ -24,8 +24,9 @@
 #include "config.h"
 #endif
 
-#include "common/darktable.h"
 #include "common/curve_tools.h"
+#include "common/darktable.h"
+#include "common/splines.h"
 #include <cairo.h>
 #include <glib.h>
 #include <math.h>
@@ -44,7 +45,7 @@ typedef struct dt_draw_curve_t
 } dt_draw_curve_t;
 
 /** draws a rating star
-*/
+ */
 static inline void dt_draw_star(cairo_t *cr, float x, float y, float r1, float r2)
 {
   const float d = 2.0 * M_PI * 0.1f;
@@ -110,8 +111,8 @@ static inline void dt_draw_grid_zoomed(cairo_t *cr, const int num, const float l
   }
 }
 
-static inline void dt_draw_loglog_grid(cairo_t *cr, const int num, const int left, const int top,
-                                       const int right, const int bottom, const float base)
+static inline void dt_draw_loglog_grid(cairo_t *cr, const int num, const int left, const int top, const int right,
+                                       const int bottom, const float base)
 {
   float width = right - left;
   float height = bottom - top;
@@ -127,7 +128,7 @@ static inline void dt_draw_loglog_grid(cairo_t *cr, const int num, const int lef
 }
 
 static inline void dt_draw_semilog_x_grid(cairo_t *cr, const int num, const int left, const int top,
-                                       const int right, const int bottom, const float base)
+                                          const int right, const int bottom, const float base)
 {
   float width = right - left;
   float height = bottom - top;
@@ -143,7 +144,7 @@ static inline void dt_draw_semilog_x_grid(cairo_t *cr, const int num, const int 
 }
 
 static inline void dt_draw_semilog_y_grid(cairo_t *cr, const int num, const int left, const int top,
-                                       const int right, const int bottom, const float base)
+                                          const int right, const int bottom, const float base)
 {
   float width = right - left;
   float height = bottom - top;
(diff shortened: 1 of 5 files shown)


TurboGit:
@rabauke : not sure you get my comments, in #3193 I asked if possible to have the fix in separate commit for example. you never replied, its hard to see how to make progress.

aurelienpierre:
Just for the record, I think the color zones is a lost cause. There is something in the colour science in it that just doesn't work, no matter how hard you try to fix the interpolation (which is nice). 

At that point, I think the best option for this kind of GUI is to work in HSV, until I find a way to make it work linearly in spectral domain.

rabauke:
> @rabauke : not sure you get my comments, in #3193 I asked if possible to have the fix in separate commit for example. you never replied, its hard to see how to make progress.

@TurboGit For this purpose I opened #3196, which just contains the reformating part.

TurboGit:
@rabauke : yes but I would have preferred to have the fix in a separate ticket for merging in 3.0. anyway this is too late for 3.0 now I suppose. we'll integrate this for 3.0.1.

aurelienpierre:
Do we absolutely have to use a cpp file to do that ?

rabauke:
@aurelienpierre No, the spline routines 

    float interpolate_val_V2(int n, CurveAnchorPoint Points[], float x, unsigned int type);
    float interpolate_val_V2_periodic(int n, CurveAnchorPoint Points[], float x, unsigned int type, float period);

can be implemented in C as well. This, would require just considerable more lines of code.

I think at the moment we should first decide if we want to fix the broken spline interpolation routines in darktable and how. This will affect in the long run all modules the use splines in 
common/curve_tools.h, not only color zones, but also rgb and LAB courve tools and others. See my bug reports on broken spline interpolation routines.

rabauke:
I have resolved all conflicts that have emerged since my first commit.

src/gui/draw.h
```diff
@@ -259,34 +260,61 @@ static inline void dt_draw_curve_set_point(dt_draw_curve_t *c, const int num, co
...
+                                                         const int res, float *x, float *y)
+{
+  c->csample.m_samplingRes = res;
+  c->csample.m_outputRes = 0x10000;
+  CurveDataSampleV2Periodic(&c->c, &c->csample);
+  dt_draw_curve_smaple_values(c, min, max, res, x, y);
+}
+
+static inline void dt_draw_curve_calc_values_V2(dt_draw_curve_t *c, const float min, const float max,
+                                                const int res, float *x, float *y, const _Bool periodic)
```
TurboGit: _Bool should be gboolean. This makes clang failing.

rabauke: Why? `_Bool` is a standard data type of the C programming language just like `int`. See https://en.wikipedia.org/wiki/C_data_types It is used in darktable sources at other places without causing any problems, see `blend.c` for example.

aurelienpierre: Maybe you should include `<stdbool.h>` then

TurboGit: No, please use gboolean as this is what is used in other routine in draw.h.

TurboGit: @rabauke : Why ?

Because it breaks the automatic test on Travis ! And because for consistency we want to use gboolean. _Bool is only used in some part (10 occurrences) of blend.c and some instance on Lua interface, we don't want to spread its use.

rabauke: I investigated this issue in some detail. It is an interoperability problem between C and C++. Compilation error is caused when `src/iop/introspection_lens.cc` includes `src/gui/draw.h`. I will fix this as soon as possible.

rabauke: I replaced `_Bool` by `gboolean` as requested. Code compiles also with Clang now.

rabauke:
@TurboGit Thanks for reviewing. I will investigate the fusion slider issue.

rabauke:
@TurboGit I fixed the mix (fusion) slider issue. When creating the lut in `commit_params` I did not account correctly for the `strength` parameter.

PR #3251, opened 2019-11-01
Revise the altivec wrapper to make the #undefs unneeded.

RobbieAB:
This is a revised, cleaner, wrapper with better behaviour.

### DIFF
--- a/src/ppc64le/altivec.h
+++ b/src/ppc64le/altivec.h
@@ -1,10 +1,11 @@
 
-#include_next <altivec.h>
-
-#undef pixel
-
-#ifdef __cplusplus
-#undef bool // type/macro name collisions are bad.
+#ifndef __APPLE_ALTIVEC__
+/* Prevent gcc from defining the keywords as macros. Do not manually
+ * undef for c99 stdbool.h compat.
+ */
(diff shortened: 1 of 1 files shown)


TurboGit:
Should we plan to merge this (and the other altivec PR) in 3.0.1 ? Should I wait for next major release? Please, let me know, I don't even know what this build is about, so I'm a bit in the dark here. Thanks.

RobbieAB:
I possess a Raptor Talos system, which uses IBM Power 9 CPUs, so it's an altivec system, but it's definitely not an Apple system.

This PR is about improving the wrapper used to work around gccs altivec implementation which is causing failures for building on ppc64le systems due to altivec.h containing a pixel macro. I don't see any benefit to delaying this change for a major release.

The other altivec PR is a bit more complicated as it's a "first pass" on a function using altivec SIMD which would be used on a Power 9 system instead of SSE2 SIMD from x86_64. It is probably worth getting broader feedback on how the altivec (or other, NEON?) SIMD should be integrated before merging.

aurelienpierre:
Since @RobbieAB is the only one to use Altivec here and CI doesn't complain, I'm merging.

PR #3256, opened 2019-11-02
Revert "lock db at high level **per image** (#3201)"

TurboGit:
This reverts commit 979787652e0106d15023c879271f498a20b2fa87.

### DIFF
--- a/src/common/film.c
+++ b/src/common/film.c
@@ -200,6 +200,7 @@ int dt_film_new(dt_film_t *film, const char *directory)
                                 -1, &stmt, NULL);
     DT_DEBUG_SQLITE3_BIND_TEXT(stmt, 1, datetime, -1, SQLITE_STATIC);
     DT_DEBUG_SQLITE3_BIND_TEXT(stmt, 2, directory, -1, SQLITE_STATIC);
+    dt_pthread_mutex_lock(&darktable.db_insert);
     rc = sqlite3_step(stmt);
     if(rc != SQLITE_DONE)
       fprintf(stderr, "[film_new] failed to insert film roll! %s\n",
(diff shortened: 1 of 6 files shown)


TurboGit:
Please DO NOT MERGE. This is for testing.

src/common/film.c
```diff
@@ -200,6 +200,7 @@ int dt_film_new(dt_film_t *film, const char *directory)
                                 -1, &stmt, NULL);
     DT_DEBUG_SQLITE3_BIND_TEXT(stmt, 1, datetime, -1, SQLITE_STATIC);
     DT_DEBUG_SQLITE3_BIND_TEXT(stmt, 2, directory, -1, SQLITE_STATIC);
+    dt_pthread_mutex_lock(&darktable.db_insert);
```
TurboGit: I'm wondering if the issue is now there. A lock has been removed.

parafin: I'm not sure how this code can race, but sure, it's worth a try to bring back this lock. Though there are a lot of other places where DB is modified, but no lock is present.

TurboGit:
Closing, the fix was merged in commit 344ce8c34d3.

src/common/film.c
TurboGit: I didn't re-introdouce this lock as after analysis and testing it is not necessary to me.

jenshannoschwalm:
There is _dev_merge_history that looks very suspicious to me.

You get a list of row id's that are **not** kept inside a transcaction. So another thread might change database lines between collection the id's and processing data that might now be incorrectly referenced.

TurboGit:
@jenshannoschwalm : I have merged the proper fix (I think), this was only for testing purpose.

PR #3260, opened 2019-11-02
culling : fix selections changes

### LINKED ISSUE #3133: culling moves to first line of collection if space is pressed on the first image of the culling view
**Describe the bug**
Running 2.7.0+1971~g4315ff6e6 culling loses track of the culled images when I press <space> on the first culled image. lighttable then jumps to the beginning of the collection and the culling view loses all previously culled images.

**To Reproduce**
1. Go to lighttable
2. switch to the culling mode
3. Select the first image in the culling view
4. Press <space>
5. You will see that your culling view loses all culled images and jumps to the beginning of the collection.

**Expected behavior**
Behave like pressing <SPACE> on any other image inside the culling view.

**Platform:**
 - OS: ubuntu linux
 - Version 2019.04

AlicVB:
this fix #3133

Due to the actual freeze, I've tried to as less intrusive as possible.

### DIFF
--- a/src/views/lighttable.c
+++ b/src/views/lighttable.c
@@ -625,7 +625,7 @@ static void _view_lighttable_selection_listener_callback(gpointer instance, gpoi
   if(lib->current_layout == DT_LIGHTTABLE_LAYOUT_CULLING)
   {
     int idover = dt_control_get_mouse_over_id();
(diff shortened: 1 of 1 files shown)


TurboGit:
Thanks fixed on my side too!

PR #3261, opened 2019-11-02
update release notes

AlicVB:
Add some of my work to release notes.
Feel free to drop them if you think they are not important enough to be mentioned, I don't care ;)


TurboGit:
Of course not, I won't drop them as those are important. I just forgot about mentioning them. Thanks.

PR #3262, opened 2019-11-02
accel window : default shortcut

AlicVB:
Map accels window to "H" by default.
There was currently no default shortcut for this feature (I've completely forgot that part in my first implementation) . This was too bad, as it's especially useful for newcomers.
Note that I've already mentioned H as default shortcut for this in the usermanual...


TurboGit:
As it is in the documentation this is really a bug, so I'll merge in 3.0. Thanks.

PR #3263, opened 2019-11-02
tone equalizer: better default settings

aurelienpierre:
This is another attempt to provide more general masking settings so users may not need to bother at all with the masking tab in tone equalizer for at least 80% of their pictures.


TurboGit:

TurboGit:
Let's also merge this, working on my side and yes ultra-safe.

PR #3264, opened 2019-11-02
filmicrgb: improve slightly the default parameters

aurelienpierre:
Just some fine-tuning of default parameters based on further experimentation.

### DIFF
--- a/src/iop/filmicrgb.c
+++ b/src/iop/filmicrgb.c
@@ -1240,17 +1240,17 @@ void init(dt_iop_module_t *module)
   dt_iop_filmicrgb_params_t tmp
     = (dt_iop_filmicrgb_params_t){
                                  .grey_point_source   = 9.225, // source grey
-                                 .black_point_source  = -4.37,  // source black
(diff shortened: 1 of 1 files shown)


TurboGit:

TurboGit:
Of course there is no issue with stability. But I cannot merge code without at least testing it and see if there is no shortcoming. And I short on time at the moment. If you tested it and there is more positive feedback I can indeed merge it.

MStraeten:
i tested these new settings - these are more convincing and in the appearance more similar (less flat) to the already defined presets. Since these are only parameters and does not exceed the range  sed in already defined presets in the module, the functionality itself is not touched, so there's no risk for the stability of 3.0.0 in my opinion

Nilvus:
Great adjustments. Tested and it's better. Quite complementary to bilat (contrast module) PR. I really think they need to be merge on 3.0 to avoid people making habits with less better default params (see my comment on bilat PR).

TurboGit:
Ok, let's this in 3.0 then. Looks ok to me too.

PR #3265, opened 2019-11-02
bilat (local contrast): improve slightly the default params

aurelienpierre:
Taking into account that the local contrast module is placed after filmic (global tonemapping), modify the default params to restore the local contrast that might have been lost in the tone mapping.

### DIFF
--- a/src/iop/bilat.c
+++ b/src/iop/bilat.c
@@ -394,7 +394,7 @@ void init(dt_iop_module_t *module)
   module->params_size = sizeof(dt_iop_bilat_params_t);
   module->gui_data = NULL;
   // init defaults:
-  dt_iop_bilat_params_t tmp = (dt_iop_bilat_params_t){ s_mode_local_laplacian, 1.0, 1.0, 0.2, 0.2 };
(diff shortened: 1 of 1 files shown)


TurboGit:

Nilvus:
Just tested. Much better, especially with new filmic RGB module. It's quite complementary I think to merge both PR filmic RGB adjustments and this PR on same milestone, v3.0 I think as it's more a needed adjustments to new modules. And it's better to have new modules with best default params found when released than waiting a new release as it will make more people working with previous params between.

TurboGit:
Ok, let's this in 3.0 then. Looks ok to me too.

TurboGit:
Found a regression :( That's why I really hesitate to integrate changes now :( Fact is when you change the mode it still used the old values!

Nilvus:
Quite strange as only default params are changed, nothing else.

alcomposer:
Nice save @TurboGit! Thanks for working this out.

junkyardsparkle:
> Quite strange as only default params are changed, nothing else.

Yes, it was pointed out on the mailing list that the defaults when double-clicking sliders are no longer the same as the init defaults... is this intentional?

aurelienpierre:
> same as the init defaults... is this intentional?

Nope. Some modules use the default params as the default slider values, some other don't and define different sliders values as default. Maybe this should be put straight.

PR #3272, opened 2019-11-03
slideshow: fix minimum delay to 2s and 4s when using high quality.

### LINKED ISSUE #3253: Slideshow Extremely slow and freezes
**Describe the bug**
Have collection of 448 pics.
In lightable view.
Press S to start slideshow.
Machine starts spinning out of control with memory usage spiking, mouse disappears.
Slideshow preference has a 1 second delay between images.
Press the space bar to start slideshow.
Images dont change, clicking on the screen or using arrow keys or tab does nothing or it takes a long time to do anything
When it does start to change  the pic the CPU usage is 100% memory usage crosses 1Gb, after  slide change it comes back to 0.
I have seen the usage climb as high as 
image

What I have figured out with trial and error, left click takes it forward, right click takes it backwards, havent figured out how to get out of the slideshow yet. Is this documented somewhere ?

Have to kill darktable to get out of slideshow.
Opencl is permanently disabled on my 2.7 version.
Have 32 GB Ram.
Images are on external SSD.

 - OS: Win 10
 - Version 2.7.0+2113~gefde14b3d

TurboGit:
Fixes #3253.

### DIFF
--- a/src/views/slideshow.c
+++ b/src/views/slideshow.c
@@ -545,6 +545,8 @@ int key_pressed(dt_view_t *self, guint key, guint state)
 {
   dt_slideshow_t *d = (dt_slideshow_t *)self->data;
   dt_control_accels_t *accels = &darktable.control->accels;
+  const gboolean high_quality = dt_conf_get_bool("plugins/slideshow/high_quality");
+  const int delay_min = high_quality ? 4 : 2;
 
   if(key == accels->slideshow_start.accel_key && state == accels->slideshow_start.accel_mods)
   {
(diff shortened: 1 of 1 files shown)


parafin:
Doesn't delay heavily depend on PC processing power and history stack applied to the particular image?

TurboGit:
> Doesn't delay heavily depend on PC processing power and history stack applied to the particular image?

Yes it does. But there is almost no way to compute this properly (disk speed SSD or not, swap, current jobs...). So what I propose here is just a minimum that will help avoiding some/most problems.

parafin:
Assumptions about computer performance tend not to hold for very long. "640 KB is enough for everybody."
I mean I'm not sure what problem this PR might solve. In a lot of cases the limits you are proposing won't be high enough anyway. And in others (maybe in the future) these limits instead will put unnecessary restrictions on the user. IMHO the only thing relevant here is how fast image change makes sense in a slideshow. 1 second is a good guess.

TurboGit:
I have a better solution... Closing this PR.

PR #3282, opened 2019-11-04
fix accel window glitch

Nilvus:
Fix a glitch on accel window (see below). I've made multiple tests/checks to be sure removing this line doesn't affect any part. Other lists on lighttable and darkroom seen (many of them) remain the same with or without this line, except the glitch that this PR solve.

Without this PR :
Capture-20191104204454-1119x70

With this PR :
Capture-20191104210022-1075x74


TurboGit:
Looks good to me! Thanks.

Issue #3284, opened 2019-11-04
Issue appears (right side of picture) with "lens correction" module, but only when processed with CPU


d2r2:
Issue with right side of processed picture, when "lens correction" module is activated, and only when processed with CPU.

Examples of buggy pictures (processed with CPU):
![image](https://raw.github.com/d2r2/darktable_bug_lens_correction_2019.11.05/master/darktable_exported_512x512_with_CPU_(bug)/P1013194_0.jpg)
![image](https://raw.github.com/d2r2/darktable_bug_lens_correction_2019.11.05/master/darktable_exported_512x512_with_CPU_(bug)/P1014506_0.jpg)
![image](https://raw.github.com/d2r2/darktable_bug_lens_correction_2019.11.05/master/darktable_exported_512x512_with_CPU_(bug)/P1015221_0.jpg)

Examples of same pictures (processed with OpenCL):
![image](https://raw.github.com/d2r2/darktable_bug_lens_correction_2019.11.05/master/darktable_exported_512x512_with_OpenCL/P1013194_0.jpg)
![image](https://raw.github.com/d2r2/darktable_bug_lens_correction_2019.11.05/master/darktable_exported_512x512_with_OpenCL/P1014506_0.jpg)
![image](https://raw.github.com/d2r2/darktable_bug_lens_correction_2019.11.05/master/darktable_exported_512x512_with_OpenCL/P1015221_0.jpg)

You may find full set of these samples (raws + processed jpegs) here:
https://github.com/d2r2/darktable_bug_lens_correction_2019.11.05

I have a filling, that issue appears in 2 weeks or so, because I'm using development Darktable branch for long time, and found this issue in last week only.

I will try to localize commit, which causes this problem in next day.

Nilvus:
Related to #3249 and #3136 issues. @rawfiner and @TurboGit : maybe this could help to resolve this margin problem. From what I see that is common on these issues is the use of CPU (without OpenCL) and a module related to image transformation size/geometry (lens correction or rotate and crop modules) and, apparently denoise profile (@d2r2 : I don't download your images but could you confirm if denoise profile is activate or not on these images ?).

TurboGit:
I have downloaded the first one and I have no lens recognized. In fact the lens module does nothing and in the image information I don't have any lens detected ?

d2r2:
@Nilvus, good point! You right, when I switch off denoise profile (color+ligthness), then it produce good images with CPU also.

[...]

PR #3294, opened 2019-11-05
denoiseprofile: update wavelets' tiling

### LINKED ISSUE #3249: dt 2.7/denoise profiled produces strange margin
sorry for creating another bug report but I am much closer now

**Describe the bug**
dt 2.7 produces strange margin on the right edge when editing raws from Olympus e-m10 and denoise profiled/preset chroma is active

**To Reproduce**
1. import any raw file produced by an Olympus E-M10 (mark I)
2. open the file in darkroom
3. activate module denoise profiled with preset chroma
4. export the file to original size

**Expected behavior**
there is a margin at the right edge of the exported photo

 - OS: Debian 11, MX 19
 - 2.7

Screenshot at 2019-11-01 11-46-34
[OP7240005ORF.ORF.zip](https://github.com/darktable-org/darktable/files/3797578/OP7240005ORF.ORF.zip)

rawfiner:
Update wavelets tiling that was still using the legacy number of scales, and not the new one.
May fix #3249 and #3284

### DIFF
--- a/src/iop/denoiseprofile.c
+++ b/src/iop/denoiseprofile.c
@@ -596,7 +596,7 @@ void tiling_callback(struct dt_iop_module_t *self, struct dt_dev_pixelpipe_iop_t
   }
   else
   {
-    const int max_max_scale = 5; // hard limit
+    const int max_max_scale = DT_IOP_DENOISE_PROFILE_BANDS; // hard limit
     int max_scale = 0;
     const float scale = roi_in->scale / piece->iscale;
     // largest desired filter on input buffer (20% of input dim)


rawfiner:
Thanks!
And I'd like to thank you for having guessed that it was a tiling issue, it made it way easier for me to find the problem :-)

I don't understand either why it reproduces only on some machines. I guess it is a matter of how many tiles will be generated, and this depends on memory settings, but I don't see why with same memory settings it is not fully reproductible...

TurboGit:
Merging now. Thanks.

PR #3301, opened 2019-11-05
slideshow: rework to lock less to enable multiple thread to render im…

TurboGit:
…age.

Also skip auto-advance if some rendering are to be done. This makes
the slideshow more responsive to key action while rendering images.

### DIFF
--- a/src/views/slideshow.c
+++ b/src/views/slideshow.c
@@ -71,6 +71,7 @@ typedef struct dt_slideshow_t
   dt_pthread_mutex_t lock;
 
   uint32_t auto_advance;
+  int exporting;
   int delay;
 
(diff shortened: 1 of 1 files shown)


jenshannoschwalm:
Have tested this pr, for me it's just perfect, **much** better responsiveness. In fact this is the first time i felt the slideshow tool is working as it should. Why not merge now?

TurboGit:
I was waiting for more feedback, if you find it more usable (as myself) I'll reconsider this for RC1 then. Thanks.

TurboGit:
Ok, let's go for 3.0 then, after all it is not dt core and slideshow has never been a good experience to me anyway (and I have enhanced it since years) but this time it looks indeed in far better shape.

PR #3302, opened 2019-11-06
Properly install tools and support files

darix:
Please also port that to 3.0

### DIFF
--- a/tools/CMakeLists.txt
+++ b/tools/CMakeLists.txt
@@ -4,3 +4,12 @@ endif()
 if (BUILD_NOISE_TOOLS)
     add_subdirectory(noise)
 endif()
+
+install(
+  FILES
+      common.sh  purge_from_cache.sh  purge_non_existing_images.sh  purge_unused_tags.sh
+  DESTINATION
+      ${CMAKE_INSTALL_DATAROOTDIR}/darktable/tools
+  PERMISSIONS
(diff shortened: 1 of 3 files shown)


darix:
@houz better?

tools/CMakeLists.txt
```diff
@@ -4,3 +4,12 @@ endif()
 if (BUILD_NOISE_TOOLS)
     add_subdirectory(noise)
 endif()
+
+install(
+  FILES
+      common.sh  purge_from_cache.sh  purge_non_existing_images.sh  purge_unused_tags.sh
```
TurboGit: should we also consider dtstyle_to_xmp.py ?

darix: which then begs the question if we should move all the "non end user tools" into a subdirectory (or vice versa) to make it easier to know what we should install with "make install"

From a packager point of view ... the python script adds python-elementtree as dependency. also before we start shipping it ... we should make it is python3 ready. many of the big distros are moving to python3 only as python2 is EOL end of the year.

darix: so I would not put that into this PR but into a 2nd and check if we are py3 ready.

TurboGit: Ok, makes sense!

darix:
anything left to be fixed to have it merged? I have the update for the master package ready which relies on this PR

TurboGit:
@darix on my side it looks ok. so if you say it's ready I'll merge.

darix:
it is all done. I also build tested it with opensuse tumbleweed and debian 10 already

TurboGit:
Thanks!

PR #3309, opened 2019-11-06
po/opencl khronos

TurboGit:
Let's try this after 3.0 is released.


darix:
please test building without the submodule as distros prefer to use their copy of the headers. your cmake change seems to imply that you want to block that.

TurboGit:
@darix: indeed, but is it wise to let distro choose their own random headers that have not been tested to be ok for darktable?

darix:
By that argument you should also ship every other dependency that DT needs so it only used your tested versions? maybe compiler too?

As I pointed out to @parafin  on irc  ... we prefer to use system copies because then we fix it once and every tool using those shared libraries is fixed. (in case for header only libraries of course we need to ship the rebuild of the dependent packages)

TurboGit:
> By that argument you should also ship every other dependency that DT needs so it only used your tested versions?

Don't try to describe things differently than they are. Here qe are talking about header files only. We do not link against anything so that's quite different than a library. And even for libraries dt check some minimal specific versions, we do not link against any random library that is found on the system.

And for sure I'm against copy. The mess that has triggered this PR is because we had a copy. If there is something to fix we should report it upstream not mess with copies. As said by @parafin the OpenCL copies should be kept pristine and pure copy of upstream.

parafin:
@darix's point is that it's a general policy of some distributions to remove such things (and indeed it makes sense for headers too) and it's easier not to make exceptions. OpenCL headers can be argued to be an exception, because they are a part of an official standard, so these files cannot change, no patches are possible here. The reason they are included in the repo is because it's not always obvious where to get those in the system, at least historically not every distribution had an easy way to install them. I would vote not to break either use-case and to remove that check in cmake. Users should be already familiar with submodules, because rawspeed is one. If they don't update submodules, then their build is broken already.

parafin:
Or just leave it as it is and close this PR without merging.

TurboGit:
I have made the message non fatal to let the possibility to packagers to use the OS installed headers.

src/CMakeLists.txt
```diff
@@ -120,7 +120,7 @@ FILE(GLOB SOURCE_FILES
   "libs/lib.c"
   "views/view.c"
   )
-FILE(GLOB HEADER_FILES "*.h" "common/*.h" "external/CL/*.h" "control/*.h" "iop/*.h" "libs/*.h" "views/*.h")
+FILE(GLOB HEADER_FILES "*.h" "common/*.h" "external/OpenCL/CL/*.h" "control/*.h" "iop/*.h" "libs/*.h" "views/*.h")
```
darix: shouldnt this only be done if the headers are actually found? maybe in the If body in src/external/CMakeLists.txt.

TurboGit: This is harmless if the submodule is not there as the globing won't do anything, no?

TurboGit:
I won't put that in 3.0.1, this will go in master soon for 3.2 release and all dev builds for more field testing.

PR #3315, opened 2019-11-07
Disable GTK client side decoration for Windows builds

### LINKED ISSUE #3307: Darktable window cannot be resized (DT 3.0, Windows 10)
**Describe the bug**

After installing the RC of Darktable 3.0 on Windows 10, i cannot resize the main application window. Fullscreen Toggle works, and on the top border of the winow at some locations the height can be changed. On the left, right and lower border resize does not work, i.e. the mouse cursor does not change to indicate a resizable window.

**To Reproduce**
1. Install the Darktable 3.0 RC on Windows 10
2. Start
3. Try to resize Window

**Expected behavior**
The Main window should be resizable

 - OS: Windows 10 Pro 
 - Version 1909 (Build 18363.418)

**Additional context**
I have two monitors, one 4K at 125% Scaling and one WQHD at 100%. I changed to only one Display at 100% but nothing changed.
Already have done a fresh install with all config files of old versions deleted.

mbaumgae:
Disable GTK client side decoration for Windows builds to prevent main window resizing issues on Windows OS.

Fixes #3307

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -84,9 +84,10 @@ int main(int argc, char *argv[])
     printf("start: %s\n", datetime);
     printf("\n");
 
-    // make sure GTK client side decoration is disabled, otherwise windows resizing issues can be oberved
-    g_setenv("GTK_CSD", "0", TRUE);
   }
+
+  // make sure GTK client side decoration is disabled, otherwise windows resizing issues can be observed
+  g_setenv("GTK_CSD", "0", TRUE);
 #endif
 
   if(dt_init(argc, argv, TRUE, TRUE, NULL)) exit(1);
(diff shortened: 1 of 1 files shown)


parafin:
As a suggestion - maybe use glib wrapper g_setenv instead.

mbaumgae:
Thanks @parafin for the code suggestion

peterbud:
Should it be only for the Windows platform? Or the other platforms are OK with disabling CDS?

mbaumgae:
I think the other platforms should be ok with the current GTK default CSD behavior, which is enabled by default as far as I understand. So this PR only affects the Windows build. This should provide more independence from the decision of the MSYS-team, whether to incorporate the CSD patch or not.

parafin:
It's inside #ifdef _WIN32, so this PR is OK.

peterbud:
OK, it's inside a `#ifdef _WIN32`, but _within_ a conditional block which is executed **only** when redirect is needed. I think this environment variable should be set irrespective of whether output redirection is needed or not.

mbaumgae:
Thanks @peterbud for this catch!

PR #3320, opened 2019-11-07
fix prophoto profile and rework profile creation (simpler)

TurboGit:
Using code from Elles profiles I get a proper output for Prophoto RGB.

This is WIP as the profile tags are not properly added, but probably the solution at hand.

Comment from profile experts welcome.

### DIFF
--- a/src/common/colorspaces.c
+++ b/src/common/colorspaces.c
@@ -361,16 +361,6 @@ static cmsHPROFILE dt_colorspaces_create_adobergb_profile(void)
   return profile;
 }
 
-static cmsToneCurve *build_linear_gamma(void)
-{
-  double Parameters[2];
-
-  Parameters[0] = 1.0;
-  Parameters[1] = 0;
-
-  return cmsBuildParametricToneCurve(0, 1, Parameters);
-}
-
 int dt_colorspaces_get_darktable_matrix(const char *makermodel, float *matrix)
 {
   dt_profiled_colormatrix_t *preset = NULL;
@@ -465,7 +455,7 @@ cmsHPROFILE dt_colorspaces_create_alternate_profile(const char *makermodel)
   cmsToneCurve *Gamma[3];
   cmsHPROFILE hp;
 
-  Gamma[0] = Gamma[1] = Gamma[2] = build_linear_gamma();
+  Gamma[0] = Gamma[1] = Gamma[2] = cmsBuildGamma(NULL, 1.0);
 
   hp = cmsCreateRGBProfile(&WP, &XYZPrimaries, Gamma);
   cmsFreeToneCurve(Gamma[0]);
@@ -515,7 +505,7 @@ cmsHPROFILE dt_colorspaces_create_vendor_profile(const char *makermodel)
   cmsToneCurve *Gamma[3];
   cmsHPROFILE hp;
 
-  Gamma[0] = Gamma[1] = Gamma[2] = build_linear_gamma();
+  Gamma[0] = Gamma[1] = Gamma[2] = cmsBuildGamma(NULL, 1.0);
 
   hp = cmsCreateRGBProfile(&WP, &XYZPrimaries, Gamma);
   cmsFreeToneCurve(Gamma[0]);
@@ -565,7 +555,7 @@ cmsHPROFILE dt_colorspaces_create_darktable_profile(const char *makermodel)
   cmsToneCurve *Gamma[3];
   cmsHPROFILE hp;
 
-  Gamma[0] = Gamma[1] = Gamma[2] = build_linear_gamma();
+  Gamma[0] = Gamma[1] = Gamma[2] = cmsBuildGamma(NULL, 1.0);
 
   hp = cmsCreateRGBProfile(&WP, &XYZPrimaries, Gamma);
   cmsFreeToneCurve(Gamma[0]);
@@ -619,7 +609,7 @@ static cmsHPROFILE dt_colorspaces_create_xyz_profile(void)
 
 static cmsHPROFILE dt_colorspaces_create_linear_rec709_rgb_profile(void)
 {
-  cmsToneCurve *transferFunction = build_linear_gamma();
+  cmsToneCurve *transferFunction = cmsBuildGamma(NULL, 1.0);
 
   cmsHPROFILE profile = _create_lcms_profile("Linear Rec709 RGB", "Linear Rec709 RGB",
                                              &rec709_primaries_pre_quantized, transferFunction, TRUE);
@@ -631,7 +621,7 @@ static cmsHPROFILE dt_colorspaces_create_linear_rec709_rgb_profile(void)
 
 static cmsHPROFILE dt_colorspaces_create_linear_rec2020_rgb_profile(void)
 {
-  cmsToneCurve *transferFunction = build_linear_gamma();
+  cmsToneCurve *transferFunction = cmsBuildGamma(NULL, 1.0);
 
   cmsHPROFILE profile = _create_lcms_profile("Linear Rec2020 RGB", "Linear Rec2020 RGB",
                                              &rec2020_primaries_prequantized, transferFunction, TRUE);
@@ -643,7 +633,7 @@ static cmsHPROFILE dt_colorspaces_create_linear_rec2020_rgb_profile(void)
 
 static cmsHPROFILE dt_colorspaces_create_linear_prophoto_rgb_profile(void)
 {
-  cmsToneCurve *transferFunction = build_linear_gamma();
+  cmsToneCurve *transferFunction = cmsBuildGamma(NULL, 1.0);
 
   cmsHPROFILE profile = _create_lcms_profile("Linear prophoto RGB", "Linear prophoto RGB",
                                              &prophoto_primaries_prequantized, transferFunction, TRUE);
@@ -661,7 +651,7 @@ static cmsHPROFILE dt_colorspaces_create_linear_infrared_profile(void)
     rec709_primaries_pre_quantized.Green,
     rec709_primaries_pre_quantized.Red,
   };
-  cmsToneCurve *transferFunction = build_linear_gamma();
+  cmsToneCurve *transferFunction = cmsBuildGamma(NULL, 1.0);
 
   cmsHPROFILE profile = _create_lcms_profile("linear infrared bgr", "Darktable Linear Infrared BGR",
                                              &primaries_pre_quantized, transferFunction, FALSE);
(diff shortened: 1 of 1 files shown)


src/common/colorspaces.c
```diff
@@ -641,17 +648,98 @@ static cmsHPROFILE dt_colorspaces_create_linear_rec2020_rgb_profile(void)
...
+    { 0.1596, 0.8404, 0.711874 }, /* green */
+    { 0.0366, 0.0001, 0.000086 }, /* blue  */
+  };
+
+  mlu1 = cmsMLUalloc(NULL, 1);
+  mlu2 = cmsMLUalloc(NULL, 1);
+
+  /* icc profile (ProPhoto RGB) */
+  icc_context = cmsCreateContext(NULL, NULL);
+  gamma18 = cmsBuildGamma(icc_context, 1.8);
```
aurelienpierre: this is not a linear RGB space though

aurelienpierre:
I remember some work I did to include ProPhoto RGB in dt's default spaces, I had the same issue with pink highlights as the ones that appear now in 3.0.

One thing to remember is ProPhoto as an imaginary range, meaning it can encode with valid RGB values some (invalid) colours that don't exist in our visible light spectrum.

src/common/colorspaces.c
TurboGit: Right, this was WIP. The good news is that I have all profile re-implemented and the pink cast is gone. I'll publish a PR later today.

TurboGit:
@aurelienpierre : can you review? thanks.

src/common/colorspaces.c
```diff
@@ -318,16 +323,24 @@ static cmsHPROFILE dt_colorspaces_create_srgb_profile_v4()
...
+  cmsXYZ2xyY(&primaries.Red,   &brg_primaries.Red);
+  cmsXYZ2xyY(&primaries.Green, &brg_primaries.Green);
+  cmsXYZ2xyY(&primaries.Blue,  &brg_primaries.Blue);
+
+  cmsCIExyY d65xyY;
+  cmsXYZ2xyY(&d65xyY, &d65);
 
   cmsHPROFILE profile = _create_lcms_profile("BRG", "BRG",
-                                             &primaries_pre_quantized, transferFunction, TRUE);
+                                             &d65xyY, &primaries, transferFunction, TRUE);
```
aurelienpierre: same about the D50 to D65 change

src/common/colorspaces.c
```diff
@@ -338,9 +351,17 @@ static cmsHPROFILE dt_colorspaces_create_gamma_rec709_rgb_profile(void)
...
+  cmsXYZ2xyY(&primaries.Red,   &rec709_primaries_prequantized.Red);
+  cmsXYZ2xyY(&primaries.Green, &rec709_primaries_prequantized.Green);
+  cmsXYZ2xyY(&primaries.Blue,  &rec709_primaries_prequantized.Blue);
+
+  cmsCIExyY d65xyY;
+  cmsXYZ2xyY(&d65xyY, &d65);
 
   cmsHPROFILE profile = _create_lcms_profile("Gamma Rec709 RGB", "Gamma Rec709 RGB",
-                                             &rec709_primaries_pre_quantized, transferFunction, TRUE);
+                                             &d65xyY, &primaries, transferFunction, TRUE);
```
aurelienpierre: same about the D50 to D65 change

src/common/colorspaces.c
```diff
@@ -352,25 +373,23 @@ static cmsHPROFILE dt_colorspaces_create_adobergb_profile(void)
...
+  cmsXYZ2xyY(&primaries.Red,   &adobe_primaries_prequantized.Red);
+  cmsXYZ2xyY(&primaries.Green, &adobe_primaries_prequantized.Green);
+  cmsXYZ2xyY(&primaries.Blue,  &adobe_primaries_prequantized.Blue);
+
+  cmsCIExyY d65xyY;
+  cmsXYZ2xyY(&d65xyY, &d65);
 
   cmsHPROFILE profile = _create_lcms_profile("Adobe RGB (compatible)", "Adobe RGB",
-                                             &adobe_primaries_prequantized, transferFunction, TRUE);
+                                             &d65xyY, &primaries, transferFunction, TRUE);
```
aurelienpierre: same about the D50 to D65 change

src/common/colorspaces.c
```diff
@@ -619,10 +638,18 @@ static cmsHPROFILE dt_colorspaces_create_xyz_profile(void)
...
+  cmsXYZ2xyY(&primaries.Red,   &rec709_primaries_prequantized.Red);
+  cmsXYZ2xyY(&primaries.Green, &rec709_primaries_prequantized.Green);
+  cmsXYZ2xyY(&primaries.Blue,  &rec709_primaries_prequantized.Blue);
+
+  cmsCIExyY d65xyY;
+  cmsXYZ2xyY(&d65xyY, &d65);
 
   cmsHPROFILE profile = _create_lcms_profile("Linear Rec709 RGB", "Linear Rec709 RGB",
-                                             &rec709_primaries_pre_quantized, transferFunction, TRUE);
+                                             &d65xyY, &primaries, transferFunction, TRUE);
```
aurelienpierre: same about the D50 to D65 change

src/common/colorspaces.c
```diff
@@ -631,10 +658,18 @@ static cmsHPROFILE dt_colorspaces_create_linear_rec709_rgb_profile(void)
...
+  cmsXYZ2xyY(&primaries.Red,   &rec2020_primaries_prequantized.Red);
+  cmsXYZ2xyY(&primaries.Green, &rec2020_primaries_prequantized.Green);
+  cmsXYZ2xyY(&primaries.Blue,  &rec2020_primaries_prequantized.Blue);
+
+  cmsCIExyY d65xyY;
+  cmsXYZ2xyY(&d65xyY, &d65);
 
   cmsHPROFILE profile = _create_lcms_profile("Linear Rec2020 RGB", "Linear Rec2020 RGB",
-                                             &rec2020_primaries_prequantized, transferFunction, TRUE);
+                                             &d65xyY, &primaries, transferFunction, TRUE);
```
aurelienpierre: same about the D50 to D65 change

src/common/colorspaces.c
```diff
@@ -656,15 +691,23 @@ static cmsHPROFILE dt_colorspaces_create_linear_prophoto_rgb_profile(void)
...
+  cmsXYZ2xyY(&primaries.Green, &bgr_primaries.Green);
+  cmsXYZ2xyY(&primaries.Blue,  &bgr_primaries.Blue);
+
+  cmsCIExyY d65xyY;
+  cmsXYZ2xyY(&d65xyY, &d65);
 
-  cmsHPROFILE profile = _create_lcms_profile("linear infrared bgr", "Darktable Linear Infrared BGR",
-                                             &primaries_pre_quantized, transferFunction, FALSE);
+  cmsHPROFILE profile = _create_lcms_profile("Linear Infrared BGR", "Darktable Linear Infrared BGR",
+                                             &d65xyY, &primaries, transferFunction, FALSE);
```
aurelienpierre: same about the D50 to D65 change

src/common/colorspaces.c
TurboGit: All previous primaries where D65 (at least that's what previous implementation was using when creating the profile). So I kept D65, but there is zero documentation about that.

For sure the new ProPhoto I have implemented has primaries for D50.

src/common/colorspaces.c
```diff
@@ -246,49 +248,44 @@ static cmsHPROFILE dt_colorspaces_create_lab_profile()
...
 
-  cmsMLU *mlu0 = cmsMLUalloc(NULL, 1);
-  cmsMLUsetASCII(mlu0, "en", "US", "Public Domain");
-  cmsMLU *mlu1 = cmsMLUalloc(NULL, 1);
-  cmsMLUsetASCII(mlu1, "en", "US", desc);
-  cmsMLU *mlu2 = cmsMLUalloc(NULL, 1);
-  cmsMLUsetASCII(mlu2, "en", "US", "Darktable");
-  cmsMLU *mlu3 = cmsMLUalloc(NULL, 1);
+  cmsSetHeaderFlags(profile, cmsEmbeddedProfileTrue | cmsUseAnywhere);
+  cmsSetHeaderRenderingIntent(profile, INTENT_RELATIVE_COLORIMETRIC);
```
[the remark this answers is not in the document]
TurboGit: What you would suggest? perceptual?

src/common/colorspaces.c
```diff
@@ -41,26 +41,28 @@
...
-};
-
-static const cmsCIEXYZTRIPLE adobe_primaries_prequantized = {
-  {0.60974121, 0.31111145, 0.01947021},
-  {0.20527649, 0.62567139, 0.06086731},
-  {0.14918518, 0.06321716, 0.74456787}
+// http://www.brucelindbloom.com/index.html?WorkingSpaceInfo.html
+static cmsCIExyYTRIPLE prophoto_primaries = {
+  /*     x,      y,        Y */
+  { 0.7347, 0.2653, 0.288040 }, /* red   */
```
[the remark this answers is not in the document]
TurboGit: Maybe... The values are copied from link referenced just above.

src/common/colorspaces.c
TurboGit: The previous code was doing:

```
  cmsWriteTag(profile, cmsSigMediaWhitePointTag, &d65);
  cmsWriteTag(profile, cmsSigMediaBlackPointTag, &black);
```

TurboGit: Unconditionally for all created profiles.

src/common/colorspaces.c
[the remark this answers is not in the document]
aurelienpierre: See https://en.wikipedia.org/wiki/ProPhoto_RGB_color_space

src/common/colorspaces.c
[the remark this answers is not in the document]
aurelienpierre: Ideally, nothing. I don't know if LCMS2 can just skip the intent and keep off-gamut values as-is, but it would be the sensible thing to do here.

src/common/colorspaces.c
aurelienpierre: Ok if the previous code used D65. You just need to check that the primaries are valid for the white point that is tagged.

src/common/colorspaces.c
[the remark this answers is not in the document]
TurboGit: Values updated.

src/common/colorspaces.c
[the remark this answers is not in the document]
TurboGit: It can, there was no intent in previous code in fact. Now my thinking is that does nothing on the creation of the colorspace, or does it? The intent would only be used while mapping an bunch of pixels in the colorspace. But anyway, I'll remove this.

PR #3322, opened 2019-11-07
Lightroom import enhancements

jgus:
- Refactored the Lightroom import code to be modular, to make it easier to extend.
- Fixed several bugs with orientation/rotation/crop which were then _also_ fixed in parallel in the mainline - I think my fixes are cleaner ;-)
- Added support for importing things processed by Adobe Camera Raw - same processes as Lightoom just a different XMP tag
- Import GPS elevation as well as lat/long if available
- Import color temperature
- Import contrast & saturation
- Import shadows & highlights

The latter three bullets are based on empirical measurements from sending a batch of test images through LR and DT with various settings and solving for the correspondence, so they're approximations (especially shadows & highlights) but they're much better than not importing them at all, IMHO.


darix:
Just curious ... the whole rest of the code base is C (ignoring rawspeed for a moment that should be a shared library)

Why did you go with C++?

TurboGit:
As @darix, I really think we should keep plain C here. Note that I'm against C++, just that for consistency we need to keep C or at least this must be discussed more broadly before.

jgus:
> Just curious ... the whole rest of the code base is C (ignoring rawspeed for a moment that should be a shared library)

Well, FWIW, not *all* of it:

```
% find . -iname \*.cc
./src/common/imageio_rawspeed.cc
./src/common/imageio_exr.cc
./src/common/exif.cc
./src/iop/lens.cc
./src/iop/bilateral.cc
./src/iop/amaze_demosaic_RT.cc
./src/iop/tonemap.cc
./src/imageio/format/exr.cc
```

And the `CMakeLists.txt` has setup logic for C++, setting the standard level, etc. Looks to me like a project that's mostly C, with some C++ too, not a C-only one. And if there's already C++, why _not_ improve other places with it too? ;-)

The comment about the shared library for rawspeed is apt, though - TBH LR import, and probably _all_ import, should be a shared library too, loaded when needed then unloaded afterward, or possibly even a plug-in - though, I looked at trying to do this as a plug-in, and it looked like it was going to be _way_ more work than just doing this, so :man_shrugging: 

> Why did you go with C++?

OOD around operations, mostly. In the current code, the logic for any given operation is kinda scattered around a very large file - a struct here, some import logic there, transform here, commit there. Frankly, it's a little hard to read through and figure out what needs to be done to add a new operation. So, it made sense to me to reorganize things so that everything specific to an operation was consolidated into one place, and to separate those from the core logic, which then just has to implement the general import framework without knowing or caring about any of the details of any particular operation. That's a pretty textbook case for OOD, and while you _can_ make that work in vanilla C, it's never made sense to me to reinvent the wheel when there's already a great tool for OOP in C called C++ :grin: 

Then too, at the point I started these changes, this code hadn't really been touched in two years. (Looks like there were a series of fixes made to orientation/rotation/crop while I made the same fixes and others in my fork, but I didn't see that until updating my changes for this PR.) TBH I figured if nobody else cared about it and I did, then I might as well give it some love, clean it up, and make it easier for myself to maintain - and who knows? If I found the current code too hard to contribute to, maybe others were in the same boat and by reducing that barrier maybe make it easier for others to contribute here too.

Nilvus:
@TurboGit and @jgus : I would also add that if this PR propose good fixes for orientation/rotate/crop, it would be better to separate them to another PR. As fixes could be added to 3.0 and enhancements on next release. So better to separate fixes to enhancements.

parafin:
All examples of C++ code in darktable is when 3rd party library (counting rawspeed) provides only C++ API or some C++ code was taken as-is from another project. So for newly written code it is expected that it will be written in C. Having a mix of languages like that doesn't make sense. It is possible to write good C code. The fact that you know C++ better, doesn't mean that other developers know it too, while knowing C is still a requirement.

jgus:
> @TurboGit and @jgus : I would also add that if this PR propose good fixes for orientation/rotate/crop, it would be better to separate them to another PR. As fixes could be added to 3.0 and enhancements on next release. So better to separate fixes to enhancements.

Not sure if there's a lot of value to that at this point - I _believe_ the existing fixes to orientation/rotate/crop are effectively the same ones I made, except that mine are in the context of my refactor. Well, my fix also uses a functional approach rather than the existing case-wise one, so it's _possible_ there's still a bad case or two in the existing code, but the ones I _know_ about have been fixed. So I guess it comes down to whether it's worth switching to the functional logic.

jgus:
> All examples of C++ code in darktable is when 3rd party library (counting rawspeed) provides only C++ API or some C++ code was taken as-is from another project. So for newly written code it is expected that it will be written in C. Having a mix of languages like that doesn't make sense. It is possible to write good C code. The fact that you know C++ better, doesn't mean that other developers know it too, while knowing C is still a requirement.

I don't think it's a question of language familiarity but of suitability of the language and using the right tool for the right job, but that's my point of view. In any case, you (or anyone else) are certainly welcome to refactor my changes to vanilla C, but I don't have the time (or frankly the interest) to invest in that. What I have here works for me, and I'm just trying to share that time & effort with anyone else who would find it valuable.

parafin:
I wouldn't have argued against re-writing the whole darktable in C++;))) But I don't think C++ is the right tool to write an addition to basically pure-C project.
I don't know what will happen with this PR, but if you're gonna write another PR for darktable with expectation that it will be merged - use C.

TurboGit:
> * mport color temperature
> 
> * Import contrast & saturation
> 
> * Import shadows & highlights

I really think it would be a shame to not have this in dt at some point. Please can you consider moving this to C that I could merge it?

I'm not for a battle of C vs C++, this is just consistency. If it was my own project it won't have been done in C or C++ :) But we must enure the project us keeping straight with the original desing.

@jgus : I'm really looking forward a merge. thanks.

jgus:
> Please can you consider moving this to C that I could merge it?

I'm happy to leave it here for anyone else that wants to pull & run it for an import, or maybe someone else will come along who wants to see this merged badly enough to refactor it, but I can't see me putting that much more work into this at this point, sorry.

TurboGit:
Not a nice decision for the community, but that's your call.

PR #3337, opened 2019-11-08
filmstrip : keep mouseover in sync with mouse position

### LINKED ISSUE #3312: Darkroom: mouse hover filmstrip strange/misaligned
**Describe the bug**
When I want to select another picture in darkrooms filmstrip, the highlighting flips at a strange moment, when you reached ~50% of the next picture.

**Expected behavior**
When the mouse cursor slides over the filmstrip, the highlighting should flip, once the mouse is about a certain (small) amount of pixels of the next photo

**Screenshots**
I think it is very hard to demonstrate via screenshot, if someone has a good idea, how I could show better than with words, I will add...

 - OS: [gentoo linux]
 - Version 3.0.0rc0+47~gd859b89a2

AlicVB:
this fix #3312
some margin count error...
@aurelienpierre can you double check, please ? git blame says it's your code... and to be honest, it works, but I would have say that the correct fix would be to **add** the margin also when we draw image (line 857) instead of removing it when we don't draw thumb... (but this don't work...)

### DIFF
--- a/src/libs/tools/filmstrip.c
+++ b/src/libs/tools/filmstrip.c
@@ -792,7 +792,7 @@ static gboolean _lib_filmstrip_draw_callback(GtkWidget *widget, cairo_t *wcr, gp
   {
     if(col < col_start)
     {
-      cairo_translate(cr, wd + line_width, 0);
+      cairo_translate(cr, wd, 0);
       continue;
     }
 


aurelienpierre:
so you remove the margin in the drawing ? That doesn't look right. I would rather add an offset in the code that maps cursor coordinates to hovered picture instead.

TurboGit:
I agree with Aurélien here.

TurboGit:
Ah no I see, this is when col < col_start where we don't want to take into account any margin as there is no image in this area ! This looks ok to me now. Thanks!

aurelienpierre:
Ok then.

PR #3338, opened 2019-11-08
Toneequalizer: fix GUI glitches

### LINKED ISSUE #3202: tone equalizer cursor issue
Color picker activation deactivates the over image tone equalizer control (and vertical line of under cursor luminance). Deactivating color picker doesn't bring back the over image tone equalizer control.

I have noticed that sometimes the same problem appear without obvious reason.
Deactivating, reactivating tone equalizer usually doesn't bring back the over image tone equalizer control.
Applying a manual correction on one of the tone equalizer module slider (curve smoothing for example) brings back the over image tone equalizer control every time.

Normal:
image

Cursor missing (the mouse is at the same place as previously):
image

 - OS: windows 10
 - Version 2.7.0+2055~g36f87d176

aurelienpierre:
* force the nodes to be recomputed at every redrawing of the equalizer view in case the module got resized since previous drawing (useful when the sliding bar is enabled on-the-fly, otherwise nodes keep their original coordinates and overflow the graph)
* fix #3202 by forcing tabs and graph interactions to give the focus to the module, and consequently enable the interactive cursor.

### DIFF
--- a/src/iop/toneequal.c
+++ b/src/iop/toneequal.c
@@ -2594,6 +2594,13 @@ static gboolean area_draw(GtkWidget *widget, cairo_t *cr, gpointer user_data)
   //if(!g->graph_valid)
   if(!_init_drawing(self->widget, g)) return FALSE; // this can be cached and drawn just once, but too lazy to debug a cache invalidation for Cairo objects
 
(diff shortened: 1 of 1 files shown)


TurboGit:
Looks ok, thanks!

PR #3350, opened 2019-11-09
update default modules presets and workspaces with new modules

### LINKED ISSUE #3030: Update modules presets
The modules presets are outdated in current master, no tone-equalizer, curves rgb, lut3d, filmic 3.., an the rest of new good stuff.

Related  with PR #3013 and in someway with #3029

aurelienpierre:
* fix #3030
* fix a bug in modules names handling that didn't use the needle in haystack pattern, so "rgblevels" would match "levels" too.
* rework the default_visible modules in the initial configuration to provide a minimal toolkit to quickely edit 90% of pictures from start to finish with no headache.


TurboGit:
Works for me. Thanks!

PR #3356, opened 2019-11-09
Increased visibility when the mouse goes over the buttons

### LINKED ISSUE #3343: Unreadable text on some controls where hovering over with mouse
Some controls (buttons, drop down menus) become unreadable
(white text on light grey background) when hovering with mouse over them:

Hovering 1

Hovering 2a

Hovering 3a

Hovering 4a

Hovering 5a

Text should change color to something more dark, like this:

Hovering 6a

or as implemented in  lighttable | darkroom | other   tabs.

Version: 3.0.rc0
OS: Windows 10

roussieau:
Hi, 
When the mouse is above the buttons, the label takes the value specified in the themes. It becomes darker to increase visibility (see image attached).

PS : In the photo, my mouse is on the button, it is only my screenshot software that removed it.

Have a good day,

Julian
Screenshot from 2019-11-10 00-33-20
 fixes #3343

### DIFF
--- a/data/themes/darktable.css
+++ b/data/themes/darktable.css
@@ -160,7 +160,6 @@
   background-image: none;
   font-family: sans-serif;
   font-size: 1em;
-  color: @fg_color;
   text-shadow:none;
   box-shadow:none;
   padding: 0;


Nilvus:
Thanks for helping but this part also affects many other items of the UI. So, yes it fixes #3343 but bring new issues. Not wanted.

aurelienpierre:
> If you see the css carefully, you will find that hover parts already exists, but for some parts of the UI, they have no effects (menu items and buttons especially)

Yes, I know, I have put the rules in there myself ^^, but they seemed to be completely ignored. Great if you have a fix.

PR #3361, opened 2019-11-10
copy metadata feature

phweyland:
Starting copy/paste feature for color & rate, metadata, geotags and tags.

image

image

There are two key points in there:
- undo/redo
- image groups

If I create a top level undo group for the 4 above features (dt_undo_start_group() ) and then call individual feature routines with undo already implemented (like dt_ratings_apply_to_image_or_group() ), should that work (i.e. a unique undo would undo the 4 features at once) ?

Advice, guidelines and comments are welcome.

### DIFF
--- a/src/common/colorlabels.c
+++ b/src/common/colorlabels.c
@@ -1,6 +1,7 @@
 /*
     This file is part of darktable,
     copyright (c) 2009--2011 johannes hanika.
+    copyright (c) 2019 philippe weyland
 
     darktable is free software: you can redistribute it and/or modify
     it under the terms of the GNU General Public License as published by
@@ -21,6 +22,7 @@
 #include "common/debug.h"
 #include "common/image_cache.h"
 #include "common/undo.h"
+#include "common/grouping.h"
 #include "control/conf.h"
 #include "control/control.h"
 #include "gui/gtk.h"
@@ -38,6 +40,33 @@ typedef struct dt_undo_colorlabels_t
   uint8_t after;
 } dt_undo_colorlabels_t;
 
+int dt_colorlabels_get_labels(const int imgid)
+{
+  sqlite3_stmt *stmt;
+  DT_DEBUG_SQLITE3_PREPARE_V2(dt_database_get(darktable.db),
+                              "SELECT color FROM main.color_labels WHERE imgid = ?1", -1, &stmt, NULL);
+  DT_DEBUG_SQLITE3_BIND_INT(stmt, 1, imgid);
+  int colors = 0;
+  while(sqlite3_step(stmt) == SQLITE_ROW)
+    colors |= (1<<sqlite3_column_int(stmt, 0));
+  sqlite3_finalize(stmt);
+  return colors;
+}
+
+static void _pop_undo_execute(const int imgid, const uint8_t before, const uint8_t after)
+{
+  for(int color=0; color<5; color++)
+  {
+    if(after & (1<<color))
+    {
+      if (!(before & (1<<color)))
+        dt_colorlabels_set_label(imgid, color);
+    }
+    else if (before & (1<<color))
+      dt_colorlabels_remove_label(imgid, color);
+  }
+}
+
 static void _pop_undo(gpointer user_data, dt_undo_type_t type, dt_undo_data_t data, dt_undo_action_t action, GList **imgs)
 {
   if(type == DT_UNDO_COLORLABELS)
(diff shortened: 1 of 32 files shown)


TurboGit:
Sounds promising ! Thanks for starting that.

phweyland:
Working on colorlabels.
1. the basic implementation using one record per color in color_labels table turns things complex while I don't see the benefit (could be one record with the mask of the selected colors...). 
2. undo is not implemented completely and nothing about image group is ...

Consumers are libs/image.c, lua/image.c, exif.cc, lightroom.c, filmstrip.c and lighttable.c.
Lighthroom.c and exif.cc do not need (and cannot use) neither undo nor image group, right ?
All the others should be compliant with both undo and image group, right (including lua/image.c ?) ?

Better to clean up this first, isn't it ? What's your view ?

phweyland:
working on tags.
It seems the table main.used_tags is updated but never used. Any information about the role of the table ?

phweyland:
@TurboGit, triggering the full dt refresh (especially collection) on any single undo can be very heavy ...
In case of embedded undo levels is there a way to have a trigger or a flag telling that the highest level is ending for a given filter ? (assuming the highest level ends after all embedded ones).
This way it would be very simple to trig the needed refreshes only once.

phweyland:
Got better understanding of undo usage. Should work with the current resources.

phweyland:
I'm ready to rewrite my changes consistently across the different metadata.
Before doing it, I would like to have your advice Pascal about few points.

- All operations do not need undo/redo and images group application, like exif.cc or lightroom.c. So we may need a convention to name routines. For example today we have dt_tag_attach(). We need a version with undo/redo and images group application and another version without. The same for ratings, colorlabels, ... Any guideline to name properly these routines ?

- colorlabels works toggling per color values. For images group, I'm toggling the value of each image of the group, even if the the color values are different from one image to the other. But the user could expect to get the same color values across all images of the group.  In that case we should toggle the color value of the main image and copy the result to the other images of the group. What do you think ?

phweyland:
>     * We need a version with undo/redo and images group application and another version without

Another option could be to add parameters 'undo' and 'group'.

phweyland:
on tagging, when the user enter new tags in entry box and press enter, Crtl-Z is not active, as long as the user does not click somewhere in dt window.
That's because the entry box grabs the focus. So far I have not found the way to make enter box release the focus (to enable Ctrl-Z and all other shortcuts as well).
Any idea or suggestion ?

phweyland:
I have a general issue with ctrl-z and dt terminating with "Stopped" console message. 
If I'm not mistaken this happens only when the mouse has got out of dt window. If dt is in full screen it seems I have not the issue.
Any idea to trap this ?
Does "Stopped" message comes from assert() instruction ? There are more than 600 occurrences of this instruction (without lua related ones).

phweyland:
Main.selected_images table includes grouped images.
In the context of some operation like toggle colorlabel this can be a drawback.
In this case, the operation is executed image per image (main and grouped). If the image input data are different for some reason, a grouped image may end up with different results from those of the main image, which is not intuitive.
Here comes my question. Would that make sense for this type of operation to calculate the result on the main image and copy it on the grouped ones ?

parafin:
> I have a general issue with ctrl-z and dt terminating with "Stopped" console message.
> If I'm not mistaken this happens only when the mouse has got out of dt window. If dt is in full screen it seems I have not the issue.
> Any idea to trap this ?
> Does "Stopped" message comes from assert() instruction ? There are more than 600 occurrences of this instruction (without lua related ones).

Maybe you pressed Ctrl-Z in the console and backgrounded DT?

phweyland:
> Maybe you pressed Ctrl-Z in the console and backgrounded DT?

As long as the mouse stays over dt window ctrl-z/crtl-y work smoothly. 
If I click on dt console window (dt has lost the foreground) and press ctrl-z, dt seems to end consistently (with Stopped message).

parafin:
Why do you consider this a bug? This is expected behaviour. Probably you resume darktable by executing an fg command after that.

phweyland:
Why do you consider this a bug?

Well, I've never seen ctrl-z / ctrl-y ending whatever application.
Now considering ctrl-z applies to dt console window that's the right behavior, but that's surprising as long as I don't realize this. :-)
Thanks for the explanation !

phweyland:
> Main.selected_images table includes grouped images.

Per default.
Would not it be more correct to have in main.selected_images only selected and visible ones and have an option or general routine to add the grouped ones when necessary ?

For routines which accept imgid =-1 or a specific `image id`, this not consistent. With -1 main.selected_images return the grouped images, while with `image id` the grouped images are ignored.

The current behavior prevents also for example to copy metadata (rating, metadata, ...) from main images to their grouped images, because I cannot select (easily) main images of group without their grouped images.

phweyland:
> > Main.selected_images table includes grouped images.
> 
> Per default.

dt_collection_get_selected() should return what I'm expecting.

phweyland:
After Ctrl-Z or Ctrl-Y,  dt_image_synch_xmp() should be called to update xmp files, right ?
Today it is not if I'm not mistaken.
Undo routines do not know the images list on which action are undone (or redone).

Any idea or suggestion on how to treat this ?

TurboGit:
You're talking about the undo/redo in the lighttable. Right?

If so, this is using a table in memory. So I suppose the undo/redo routine know what are the images changed. I don't remember all the details about the code though.

phweyland:
> You're talking about the undo/redo in the lighttable. Right?
> 
> If so, this is using a table in memory. So I suppose the undo/redo routine know what are the images changed. I don't remember all the details about the code though.

I'm talking about the routine dt_undo_do_undo() or dt_undo_do_redo().
They are called by lighttable, map and darkroom, but yes my current concern is about lighttable.

An option is to call  dt_image_synch_xmp() inside individual _pop_undo() routines, but in case of copy and paste, dt_image_synch_xmp() can be called 5 times per image. This doesn't sound right.

I could propose the following.
Before calling undo() routine for each undo item, dt_undo_do_undo() and dt_undo_do_redo() could initialize an empty list of images and pass its pointer to undo() routine.
If undo() routine affect one or several image it adds their id to the list.
When all the undo items are treated, either dt_undo_do_undo() and dt_undo_do_redo() call themselves dt_image_synch_xmp() for each image of the list or return the list to the caller.
What do you think ?

TurboGit:
@phweyland : sounds like this could be working indeed and will solve the sync between the db and the xmp. I suppose this should be part of another PR to avoid having too many features in the same PR.

phweyland:
> sounds like this could be working indeed and will solve the sync between the db and the xmp. I suppose this should be part of another PR to avoid having too many features in the same PR.

Ok. I'm close to the end then. I'll make some clean-up and some more testing and I'll be ready for review. Thanks.

phweyland:
The UI looks like this:

image

If needed I can set up a specific module for metadata copy.
There is a missing function: update grouped images metadata with main images metadata (for the selected main images). For another PR.

Ready for review. Thanks

TurboGit:
The current default behavior is not what I would have expected for multivalued metadata like color labels or keywords. For me a past should merge the copied metadata into the target image and not replace it. The [add] button is for that, I understand but the default is copy / paste to me.

I would either remove the [add] button and have a check box : overwrite to select the desired behavior.

Alternatively I would rename the [paste] button to [merge] and the [add]  to [replace].

What do you think?

TurboGit:
Also please can you rebase code on master? Thanks.

phweyland:
> What do you think?

Your last proposal is the best one for me.

> multivalued metadata like color labels

For metadata and tags the current "add" button adds the values to existing ones. But not color labels. I'll do this as well.

> Also please can you rebase code on master?

I'll do.

phweyland:
I see you have mentioned "rebase". I've made a "merge" instead (what I do usually). Had you a specific need for '"rebase" ?

TurboGit:
@phweyland : Yes I need rebase to get a clean topic branch (your branch) before merging. Merging a topic branch with many merges from master makes a mess in the Git history. A topic branch should be linear, not necessary from current to master but one should never merge master into them.

phweyland:
@TurboGit : I understand.
I've run a rebase. My code is completely messy now and doesn't compile. I don't recognize it.
Yet I did accept all my own changes...
I'll try to restart from 6bd8790 above.

phweyland:
Rebase is difficult to be followed when development has some back on forth because it's difficult to replay everything in the good order.
Second rebase doesn't work on this particular code. I don't why, but it proposes choices which do not make sense. Both sides are crap (like it wanted to mix different versions).

Is there a way to flatten the commits of the full PR first and then rebase ?

phweyland:
It seems that git merge --squash makes the trick.
I'll redo my this morning work at top of it.

TurboGit:
The mess is probably because there was merging from master multiple times.

One solution:
- merge again master into your topic branch to be clean
- copy/backup the modified files (all of them)
- reset your branch to be current master
`  $ git reset --hard origin/master`
- copy back the files into your working tree
- do a single commit with your changes
- force push into your upstream branch

This should flatten the history and keep your current changes.

phweyland:
> The mess is probably because there was merging from master multiple times.

You are right. So ... I should not use merge, right ?
And use rebase just at the latest ?

Thanks for the explanation. Meanwhile I've done this, which seems to work too:

- git branch -m meta-copy meta-copy-crap (not to lose everything)
- git fetch origin meta-copy
- git checkout meta-copy
- git checkout 6bd8790
- git checkout -b meta-copy-restart
- git checkout master
- git branch -D meta-copy
- git checkout meta-copy
- git merge --squash meta-copy-restart
- git commit

I may have skip the checkout on a previous commit...

TurboGit:
> You are right. So ... I should not use merge, right ?

Right.

> And use rebase just at the latest ?

Not, that's not mandatory. Only rebase if you need to get access to some changes (API) in master that you need to use in your topic branch. Otherwise it is best to not rebase (and not merge). Just work on the topic branch. You can still test it on a separate branch (or your master) to see if all goes well if it is a long lived topic branch. But that's anyway what I do on my side. I'm merging the topic on my master and test them before integrating.

Hope this clarifies the situation :)

phweyland:
> it is best to not rebase (and not merge).

 That's a simple and clear instruction! Good. :)


> Hope this clarifies the situation :)

Yes indeed. Thanks.
I hope the last push is ok for you.

TurboGit:
With all my recent merging this has conflicts now, can you fix them? Thanks!

phweyland:
> With all my recent merging this has conflicts now, can you fix them? Thanks!

Should be done now.

TurboGit:
@phweyland : What about moving the metadata copy widget into a separate tab ? I find the current widget a bit crowded now and the copy metadata is not something you use all the time. Thoughts?

phweyland:
> I find the current widget a bit crowded now and the copy metadata is not something you use all the time. 

Agreed


> What about moving the metadata copy widget into a separate tab ?

My first idea would be to make the metadata operations block expandable (folded by default). 
Or do you think about a new module in lighttable ?

Another point where I'm not completely comfortable is the wording which is different from "history stack". A "mode" selector "paste/overwrite" would be more consistent perhaps.

TurboGit:
> My first idea would be to make the metadata operations block expandable (folded by default).
> Or do you think about a new module in lighttable ?

We had expandable at some point for some modules and now in dt we do prefer using tabs. This is clean, as expandable but does not change the widget height. When a widget is expanded it becomes too high anyway.

 

> Another point where I'm not completely comfortable is the wording which is different from "history stack". A "mode" selector "paste/overwrite" would be more consistent perhaps.

You mean to control the replace or fusion mode? Why not !

phweyland:
The names of the tabs will be: `images` for the current content and `metadata` for the new addition. I will move also `refresh exif` to metadata tab.
That's OK ?

TurboGit:
Sounds perfect to me! Thanks.

PR #3362, opened 2019-11-10
Fix UI glitches remaining

### LINKED ISSUE #3281: grey theme has a bit low contrast for waveform and contrast equalizer UI
as the subject says ... maybe we could keep the darker background for those?

### LINKED ISSUE #3343: Unreadable text on some controls where hovering over with mouse

### LINKED ISSUE #3359: hbox'd buttons expand on hover
**Describe the bug**
Buttons contained within a GTK hbox expand on hover, these include buttons in the following modules, styles, tagging, vignetting, rgb levels, levels, color mapping, basic adjustments. Thanks to Timur in the mailing list for reporting this.

**To Reproduce**
Hover over buttons in one of the modules listed above

**Expected behavior**
Buttons should stay the same size

**Screenshot**
test

 - OS: I'm on Arch Linux, but I don't think this is platform specific.
 - Version 3.0RC0

**Additional context**
I fixed this issue in the metadata modules in 53deee7 by changing from a hbox to a grid, but I thought maybe there's a more elegant way to fix this. If not I'll change the rest to grids soon.

Nilvus:
Fixes #3343. @mlales please test to confirm I don't forgot anything
To fix that, it was necessary to separate background color and color on all items listed as some needs both, other one or the other only to work correctly. That makes a bigger CSS but a better UI !
Thanks again to @roussieau as his try makes me think after that to make some more tests and find that.

Also improve UI consistency on menus and tooltips background on grey theme.

And also :

- some submenus not readable on darkroom improved
- other menu on top right visibility fix
- improve tooltip background on other themes (as it made for grey theme)
- improve notebooks tabs readability
- fix #3281
- and seems to fix #3359 also


data/themes/darktable-elegant-grey.css
```diff
@@ -74,6 +74,9 @@
 @define-color field_hover_bg @grey_75;
 @define-color field_hover_fg @grey_35;
 
+/* Tooltips and contextual helpers */
+@define-color tooltip_bg_color @grey_35;
```
aurelienpierre: this should inherit the default CSS colour, that is better suited with bauhaus widget (sliders). Better leave it untouched in child-themes I think.

data/themes/darktable-elegant-grey.css
```diff
@@ -100,3 +103,8 @@
 @define-color graph_fg_active @grey_95;
 @define-color graph_grid @grey_35;
 @define-color inset_histogram alpha(@grey_95, 0.50);
+
+menu
+{
+  background-color: @grey_35;
```
aurelienpierre: please avoid directly setting colours in themes, but use defined properties instead (`@define-color background_something @grey_whatever` then `menu { color: @background_something ;}`). It will make maintenance and derived themes creation easier.

data/themes/darktable-elegant-grey.css
Nilvus: having dark background on tooltips in a clearer theme like grey one is not consistent to the UI. So I don't see how to change that background another way. And I really don't see how a dark tooltip background is better suited with bauhaus widget.

data/themes/darktable-elegant-grey.css
Nilvus: You're right, I will improve that.

data/themes/darktable-elegant-grey.css
aurelienpierre: right click on a slider with this change and you will see that the cursor border (triangle) has a different colour. This is because I haven't been able to dynamically change the sliders border color. I agree it's not consistent but it's less ugly until I find a way to change cursor border color with CSS rules.

Nilvus: This line doesn't affect the cursor border color, even in popup mode. It's only the really_dark line that you also removed that impact this. In fact, the border color of the slider cursor remains dark when popup activated instead of using bauhaus_indicator_border setting. So this tooltip line doesn't need to inherit from darktable.css theme. To be sure, I change this line to set a red color. See above, that doesn't affect popup right clider cursor slider :
Capture-20191110145742-306x456

Nilvus:
Need help to improve submenus and hovering popup list (styles one in darkroom and presets one for examples).

Adding this :

```
menuitem:hover label,
{
  color: @button_hover_fg;
}
```
improve readability of text when hovering items but break display of submenus items on styles menu list in darkroom.

While removing the menuitem:hover label make text hovering not so visible and remains back submenu text when not hovering.

So, by now and after a lot of tests, I could only have one of this two items ok. Any idea how to solve both here ? I will continue to think about that but any help/idea would be appreciate

Nilvus:
@aurelienpierre : I see a possibility here but I would like your view on that. We could darken the hover background. You set it to grey_55 for darktable.css for example, we could downgrade it to grey_45 (and same gap on other themes). That makes all the menus works great, even the submenu. But this will change all items hovered. Other possibility is to set a button_hover_bg2 just for menuitem but would break a little the consistency of the UI...

Of course the better would be a possibility to set separately menu hover text and submenu _not hover_ text but can't find how...

cryptomilk:
In the darktable-elegant-gray theme. If you open the contract equalizer and modify it in the luma tab, it is very hard to see what you're doing.

Nilvus:
> In the darktable-elegant-gray theme. If you open the contract equalizer and modify it in the luma tab, it is very hard to see what you're doing.

Another thing that was ok when I finish this theme at the end of June and now breaks...

Nilvus:
Finally, I just changed a bit contrast in buttons fg and bg hover. That makes menus good now and keep other parts ok too. It's the best I find by this time.

I also improve contrast in waveform contrast equalizer as @cryptomilk or @darix point (it should fix #3281 if I finally understand what @darix say...).

Last commit also improves readability of notebooks tabs.

TurboGit:
> Another thing that was ok when I finish this theme at the end of June and now breaks...

Thanks for the hard work @Nilvus ! I understand how frustrating this can be. No more WIP, I'll test and merge when I'll have time.

Nilvus:
> > Another thing that was ok when I finish this theme at the end of June and now breaks...
> 
> Thanks for the hard work @Nilvus ! I understand how frustrating this can be. No more WIP, I'll test and merge when I'll have time.

I propose to wait that people mentionned (at least one or two of them) test that to be sure nothing else need to be changed. Or waiting 2 or 3 days if no answer.

cryptomilk:
Looks fine for me now!

aurelienpierre:
I will test tomorrow. Thanks @Nilvus !

data/themes/darktable-elegant-grey.css
aurelienpierre: > This line doesn't affect the cursor border color, even in popup mode. It's only the really_dark line that you also removed that impact this.

Oops, my bad, you are right. Proceed then.

mlales:
It's fine for me. Thanks @Nilvus !

mlales:
BTW, I noticed that some icons have very low contrast. Two of them (workspace presets, reset parameters) are almost invisible. Would it be possible to increase contrast a bit (maybe optional) ?

Icons

TurboGit:
> I will test tomorrow. Thanks @Nilvus !

Nilvus:
@TurboGit : I will also look on what @mlales show about icons this afternoon. @mlales : I agree with you but only for preset ones. I'm not sure it will be possible though to improve that with css and without breaking something else. Other icons are good for me. Don't forget that the purpose of this new UI is to let the view on image first so contrast of some is also for that. Too much contrast would not be good for this purpose.

aurelienpierre:
I think the main issue here, @mlales, is your icons are too small. They should be the same height than the text. As icons get bigger, you need less contrast to see them clearly.

mlales:
I agree. Is it possible to configure the size of icons ?

Nilvus:
@mlales : indeed, I just made few tests and have better visibility by default on my system. And I confirm that my icons are a little bigger regardless of text. I also test that the only possibility (without breaking UI consistency) is to apply grey_70 to button_bg_color. Anyway, with normal size icons, the visibility difference is too small to propose that. So default colors are great as they actually are. If everything is ok for that PR, I will just wait to have your feedback @aurelienpierre. It's good for me.

aurelienpierre:
Icon size is supposed to get set from text size, which slightly concerns me here because it looks smaller than it should.

Nilvus:
Do you have some old settings on your darktablerc file that could change size (like screen_dpi ones for examples) ? Could you just try to see how the UI is rendering on your size by just moving your darktablerc file to another folder (with darktable closed) and restart darktable to see what you have with a new default darktablerc file ?

Nilvus:
@TurboGit : you can merge !

Nilvus:
@TurboGit : just wait, I see another glitch (a little one) on grey theme (middle grey color picker on rgb levels module not visible). I will try to fix that this afternoon.

mlales:
I tried with uninstall & fresh install. Symbols inside icons are very small again. But now I can see that icon size per se is probably OK, just "content" (symbols) is very small. Visibility is a bit better with grey theme, but I prefer darker.

Icons3

Nilvus:
Your text is not render as it should, so the whole UI is impacted. Have you installed Roboto fonts ? I suggest you to open a post on pixl.us to continue discussing it as what you see is not related to a darktable issue but to a config one. It will be difficult to discuss this PR and your issue on this thread.

Nilvus:
@TurboGit and @aurelienpierre : the last glitch I see is this one above. I've made some tests and I can't solve that with CSS. The only possibility would be to change background color but that totally breaks the actual good balance UI we have. So I will not do that. Anyway, I try to change color, border, background color on the color picker buttons only. I see on Gtk Inspector they're applied but not effect on the UI. Has they are specific color picker, I think they're constant and hardcoded. The middle grey is not visible on grey theme, and the dark one not really visible on darker and default themes.
Capture-20191111133128-301x26

If it's ok for you, I suggest to merge this PR as I'm quite sure this last glitch couldn't be fix (without breaking anything else) with CSS. Anyway, if you have any idea...

mlales:
@Nilvus OK, thank you once again !

TurboGit:
@Nilvus : if I add specific names for the three pickers do you think this could be fixed ?

I could add name like "picker_black" "picker_grey" and "picker_white" ?

aurelienpierre:
or maybe just draw borders on the pickers ? One of them at least is going to fail no matter the background colour…

Nilvus:
@TurboGit : that would be the best option.
@aurelienpierre : I already try to add borders and as I say, Gtk inspector confirm that borders are applied but they don't appear on the UI. For example, I test on colorpicker on the right of auto button and I can change all things but not for the color picker dark, grey and white ones.

TurboGit:
@Nilvus : I'm on it... stay tuned...

aurelienpierre:
> I already try to add borders and as I say, Gtk inspector confirm that borders are applied but they don't appear on the UI. For example, I test on colorpicker on the right of auto button and I can change all things but not for the color picker dark, grey and white ones.

No, I mean in the C code, use `cairo_stroke` to draw the borders in addition to `cairo_fill` that is currently used to draw the pickers.

TurboGit:
Ok names added: 
```
picker-black
picker-grey
picker-white
```

Nilvus:
@TurboGit : thanks, it helps. Anyway, it lacks something else. I'm now sure CSS options added are overwritten. I've test by adding a red background-color on middle grey picker. When I point my cursor on this picker, I see the red background appear a little second before having the hover options. When I move my cursor outside, I still have the grey background module and not the red one.
I can't understand why and quite sure it's not css here.
Do you have any other idea ? Or maybe, as @aurelienpierre propose, adding borders with Cairo ?

I really think the best would be the possibility to make pickers appears like the auto buttons below for example (maybe a little less contrasted here).

TurboGit:
@Nilvus : Found, we override the color in the code indeed :)
I'll remove that in a PR to let you test. Again... stay tuned!

TurboGit:
@Nilvus : can you test with this : #3377 ?

Nilvus:
> @Nilvus : can you test with this : #3377 ?

I'm on !

Nilvus:
@TurboGit and @aurelienpierre : that's good now. Thanks @TurboGit for your PR. I only put colors of pickers with css. Finally, only grey picker on grey theme need to be improved with that tweak. That needs to, of course, change the grey color of the picker. I test with grey_65 and grey_70. Contrast is better on grey_70 of course but near the middle grey one. Finally, I think grey_65 is a good compromise (picker visible and nearest to the middle grey ; grey_60 doesn't show good visibility). I will propose commit with @grey_65.

Capture-20191111153330-316x235

If that's ok for you two, this could be merge after #3377 PR merged and after seeing my next commit here.

aurelienpierre:
ok for me

TurboGit:
@Nilvus : Thanks, looks ok to me. Will merge all that in a moment.

Nilvus:
@TurboGit : commit ok. You could merge when you want ! Thanks for your help and all your hard work and review.

PR #3370, opened 2019-11-10
fixes parsing when comma in exif

codingdave:
fixes issues when exiv2 lens metadata contains comma:

The function `parse_maker_model` uses the comma in the input string to
separate the maker from the model. This does not work when the input
already contains a comma, as it is e.g. in "Sigma 24-70mm F2,8 DG OS HSM
Art".

This PR removes the special case in which the comma separates maker from model (and removes maker). (The function was supporting someting like parsing "Nikon, Sigma someLens F2.8" into maker="Nikon" model="Sigma someLens F2.8" and I must question its validity looking at the exiv2 database: I did not find any lens metadata that contains the maker and the model.)

* renames parse_maker_model to parse_model
* removes code to parse the maker

You can test that by adding the file ~/.exiv2 that contains something
like this:

cat ~/.exiv2
[nikon]
201=Sigma 24-70mm F2,8 DG OS HSM Art

replace "nikon" with your body and "201" with your lensID. Then
import images into darktable (it will not affect images already in your
database).

At the same time I have created a PR (https://github.com/Exiv2/exiv2/pull/1055) on exiv2 to fix the comma.
replace it by a dot.

### DIFF
--- a/src/iop/lens.cc
+++ b/src/iop/lens.cc
@@ -1678,34 +1678,13 @@ static void camera_menu_fill(dt_iop_module_t *self, const lfCamera *const *camli
   g_ptr_array_free(makers, TRUE);
 }
 
-static void parse_maker_model(const char *txt, char *make, size_t sz_make, char *model, size_t sz_model)
+static void parse_model(const char *txt, char *model, size_t sz_model)
 {
-  const gchar *sep;
-
   while(txt[0] && isspace(txt[0])) txt++;
-  sep = strchr(txt, ',');
-  if(sep)
-  {
-    size_t len = sep - txt;
-    if(len > sz_make - 1) len = sz_make - 1;
-    memcpy(make, txt, len);
-    make[len] = 0;
-
-    while(*++sep && isspace(sep[0]))
-      ;
-    len = strlen(sep);
-    if(len > sz_model - 1) len = sz_model - 1;
-    memcpy(model, sep, len);
-    model[len] = 0;
-  }
-  else
-  {
-    size_t len = strlen(txt);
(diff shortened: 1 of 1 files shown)


cryptomilk:
Note that if you change exiv2 to use a dot instead of a comma, you might need to change lensfun too and add an alias to the lens :-)

TurboGit:
No sure but isn't this an issue with locale ? If we call exiv2 API with a C locale defined, won't this be a better fix?

codingdave:
@cryptomilk I did not yet push my lensfun data, do you still think it is required, even if the exiv2 project might have added the lens?

@turbogit the split happens on the ',' character. It's locale independent. The input string comes as it is from the exiv2 project.

TurboGit:
I understand about the parsing. But the coma is there after having read the data using exiv2. And I'm sure the coma vs dot is a locale issue. My point is maybe **at the time we read exiv2 data**, if we ensure the r**eading is always done with en_US locale** we won't have this issue.

codingdave:
I understand your idea but the input on exiv2 side is a string. It's not assembled using a decimal separator that is locale dependent. Please see my PR for exiv2.

TurboGit:
@codingdave : Which PR?

codingdave:
https://github.com/Exiv2/exiv2/pull/1055 and the previous one https://github.com/Exiv2/exiv2/issues/743. You can easily see the line in the exiv2 database in the diff.

TurboGit:
Ok, thanks for the links.

codingdave:
Thank you!

PR #3379, opened 2019-11-11
filmicrgb: correct the output power in auto tuner too

aurelienpierre:
1. The output power was auto-adjusted in sliders callbacks (on manual changes) to fit the log grey value to the display grey, but not when using auto tuners. Fix that.
1. The black exposure was bounded by (- white exposure) in auto-tuners which lead to an exaggerated contrast most of the time. Delete that bound.
1. Bound the auto-set parameters to sane values to avoid black pictures when the color-pickers selected crap.


TurboGit:
Works better indeed. Thanks.

junkyardsparkle:
This keeps a badly placed gray picker from making the black and white go instantly insane, *but*... after you've clicked it outside the image the lower bound of the *slider* goes to zero, and just sliding all the way to the left will now induce madness...

PR #3390, opened 2019-11-11
Fix height histogram scale of 'rgb levels' and 'rgb curve' iops

leclercj:
In RGB-linked mode, histogram height was set at max value of RED channel only. Now adding GREEN and BLUE channels in the max value calculation.


TurboGit:
Looks good to me, thanks!

PR #3391, opened 2019-11-11
Add distinguishable text for always-on entries in history.

TurboGit:
This was hitching me since some tine now. @Nilvus, what do you think?

### DIFF
--- a/data/themes/darktable.css
+++ b/data/themes/darktable.css
@@ -328,7 +328,9 @@ button
 }
 
 #recent-collection-button,
-#history-button
+#history-button,
+#history-button-always-enabled,
+#history-button-default-enabled
 {
   /* Exception because history and recent collections modules make use of a
    * list of GTK buttons but we don't want them to look like action buttons */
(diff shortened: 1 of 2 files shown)


jenshannoschwalm:
This really looks great! Will lead to less confusions.

TurboGit:
@jenshannoschwalm : this with #3392 should avoid most issues about the history regarding default modules.

I just want to have feedback from @Nilvus on the CSS/GUI part before merging.

Nilvus:
@TurboGit :  I’m at work so I will see that tonight

aurelienpierre:
I like it.

Nilvus:
@TurboGit : a really good thing to distinguish these modules on all part of darkroom. A more consistent and efficient UI with your adds on these. Great job. I really like it.

Only thing is I find the color text lacks a little bit contrast on all themes and especially on grey themes. Just change shade color on css you use from 0.8 to 0.9 make a better contrast on all themes. So I would change line 346 to that :

`  color: shade(@plugin_label_color, 0.9);`

TurboGit:
Fine with me. Will do.

TurboGit:
I have also tagged the first "original" node as always one. Thanks for the review.

PR #3396, opened 2019-11-12
Hover and focus effect forgotten on always on modules in history

Nilvus:
We just forgot hover and focus effect on last #3391 PR by @TurboGit.


TurboGit:
Thanks, indeed this was missing.

PR #3399, opened 2019-11-12
blend gui: fix a few glitches

upegelow:
* background of polarity icons set to transparent
* state of polarity buttons is now only reflected by the icon shape (plus or minus) and not by different colors or opacity
* polarity button display is now updated on every state change rather than only  when leaving the widget


TurboGit:
Looks good to me. Thanks Ulrich!

PR #3400, opened 2019-11-12
css: fix history selection buttons.

TurboGit:
@Nilvus : if you can have a look and test?

### DIFF
--- a/data/themes/darktable-elegant-dark.css
+++ b/data/themes/darktable-elegant-dark.css
@@ -32,7 +32,6 @@
 /* Scroll bars (sliders) */
 @define-color scroll_bar_inactive @grey_55;
 @define-color scroll_bar_active @grey_65;
-@define-color scroll_bar_focus @grey_60;
 @define-color scroll_bar_bg @grey_25;
 
 /* Modules box (plugins) */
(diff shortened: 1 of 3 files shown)


Nilvus:
I never questioned that lines as they were here from start and @aurelienpierre works. That's indeed solve the issue but also remove all focus options. Anyway, finally I find it better without focus options. Sometimes, focus let have the focus on a button previously clicked even when usage is over. For example, when clicking compress history, focus was remaining on button after clicking on it. So I'm on to remove all.
@aurelienpierre : are you ok for that or did you see an action/item where focus option is better to keep ?

I find actions on the UI smoother now, not only in history.

**Two points here :**
1) Anyway, by removing these lines, no more used focus option. So that needs to also remove on all themes file the two define-color button_focus (fg and bg), no more used.
2) I also see that I forgot hover option on "other" dropdown menu. Not really related to this PR but to have a correct hover on that menu item, we just need changing line 1166 from :
`#view_dropdown:hover`
to : 
`#view_dropdown .combo:hover *`
Could you add it to your PR to just avoid a PR for that ?

TurboGit:
1 & 2 : done.

TurboGit:
I have removed all :focus rules.

Nilvus:
Great. Thanks for all that. I'm ok for this PR. Just see what @aurelienpierre think about that.

Nilvus:
And by the way, it's also make the CSS cleaner and more easier to maintain/update.

aurelienpierre:
works for me.

TurboGit:
Thanks! Merging.

PR #3409, opened 2019-11-13
iop-order: don't move colorin/colorout when converting v1 -> v2.

### LINKED ISSUE #3401: iop order isn't saved when image history from 2.6 is migrated
**Describe the bug**
Darktable 3.0.0rc0 is shorting in wrong order the input color profil and unbreak input profil iops when open an earlier edited image (2.6.2).

Update:
**Not only GUI problem, the processing pipeline have wrong order too!**

What I see in the color groups panel of dt 3.0.rc0:
Képernyőkép erről: 2019-11-12 20-09-57

```
.
.
.
.
unbreak input profil
color profil
```

But the right order is:
Képernyőkép erről: 2019-11-12 20-22-27

```
.
.
.
.
color profil
unbreak input profil
```

**Platform:**
 - OS: [Ubuntu 19.10]
 - Version [3.0.0rc vs. 2.6.2]

TurboGit:
The rules where quite wrong as colorin was moved after demosaic. This
is ok, but between demosaic and colorin there is many modules like
profile_gamma for example.

In fact we don't want to move any module while migrating to v2 as the
legacy order is static and properly ordered (it is the legacy order
implemented in darktable since its inception).

We just keep a single move (dither before borders) as those where only
separated by a single module and should not break the pipe order.

Fixes #3401.

### DIFF
--- a/src/common/iop_order.c
+++ b/src/common/iop_order.c
@@ -77,16 +77,18 @@ static void _rewrite_order(GList *iop_order_list)
 }
 
 /* migrates *_iop_order_list from old_version to the next version (version + 1)
- * limitations:
- * - to move an existing module that is always enabled a new version must be created, otherwise
- *   modules can be added/moved in the current version
- * - a module can't be more than once on the same version
+ *
+ * Basically here there is two orders:
+ *   v2 : legacy order used for edits done prior to 3.0
+ *   v5 : new order for all edits starting with 3.0
  */
 static int _ioppr_legacy_iop_order_step(GList **_iop_order_list, GList *history_list, const int old_version, const int dont_move)
 {
   int new_version = -1;
 
   // version 1 --> 2
+  // v2 is the version that will be used for all edits started before 3.0 was out.
+  // v2 corresponds to the legacy order (as proposed up to 2.6.3) with the new modules available.
   if(old_version == 1)
   {
     // EVERY NEW MODULE MUST BE ADDED HERE
@@ -103,6 +105,7 @@ static int _ioppr_legacy_iop_order_step(GList **_iop_order_list, GList *history_
     new_version = 2;
   }
   // version 2 --> 3
+  // v3 is a temporary version corresponding to first attempt to propose a new order during development of 3.0.
   else if(old_version == 2)
   {
     // GENERAL RULE FOR SIGNAL PROCESSING/RECONSTRUCTION
@@ -194,6 +197,8 @@ static int _ioppr_legacy_iop_order_step(GList **_iop_order_list, GList *history_
 
     new_version = 3;
   }
+  // version 3 --> 4
+  // v4 is a temporary version corresponding to second attempt to propose a new order during development of 3.0.
   else if(old_version == 3)
   {
     // version 4 is a rewrite of the iop-order of previous list. As it
@@ -204,6 +209,8 @@ static int _ioppr_legacy_iop_order_step(GList **_iop_order_list, GList *history_
     if(!dont_move) _rewrite_order(*_iop_order_list);
     new_version = 4;
   }
+  // version 4 --> 5
+  // v5 is the final order proposed for new edits in 3.0.
   else if(old_version == 4)
   {
     if(!dont_move)
--- a/src/common/iop_order.c
+++ b/src/common/iop_order.c
@@ -89,9 +89,6 @@ static int _ioppr_legacy_iop_order_step(GList **_iop_order_list, GList *history_
   // version 1 --> 2
   if(old_version == 1)
   {
-    _ioppr_move_iop_after(_iop_order_list, "colorin", "demosaic", dont_move);
-    _ioppr_move_iop_before(_iop_order_list, "colorout", "clahe", dont_move);
-
     // EVERY NEW MODULE MUST BE ADDED HERE
     // there should be no _ioppr_insert_iop[before|after] in any other places
     _ioppr_insert_iop_after(_iop_order_list, history_list, "basicadj", "colorin", dont_move);


junkyardsparkle:
Seems to work for me on 2.6.3 histories:
```
^^^^ Writing history image: 6, iop version: 2
          rawprepare, num 0, order   1.00000, v(1), multiprio 0, enabled
         temperature, num 1, order   3.00000, v(3), multiprio 0, enabled
          highlights, num 2, order   4.00000, v(2), multiprio 0, enabled
            demosaic, num 3, order   8.00000, v(3), multiprio 0, enabled
            colorout, num 4, order  58.00000, v(5), multiprio 0, enabled
               gamma, num 5, order  71.00000, v(1), multiprio 0, enabled
                flip, num 6, order  20.00000, v(2), multiprio 0, enabled
           basecurve, num 7, order  23.00000, v(6), multiprio 0, enabled
       profile_gamma, num 8, order  25.00000, v(2), multiprio 0, enabled
             colorin, num 9, order  27.00000, v(6), multiprio 0, enabled
       profile_gamma, num 10, order  25.00000, v(2), multiprio 0, enabled
vvvv
```

parafin:
I don't understand why you need to convert from v1 to v2 in `dt_ioppr_convert_onthefly`. Shouldn't `dt_ioppr_legacy_iop_order` take care of adding new iops to iop order?
Also your changes here probably invalidate code that defined v3 and v4 iop order.
I assume v2 iop order actually existed, so re-writing what it actually was isn't a correct fix IMHO. Instead you should revert commit 3562593 and re-think that fix.

TurboGit:
@parafin : after all explanations I gave I'm not sure why you keep saying that I should revert 3562593 ! I did this on purpose and still think it was a good choice. One choice against probably some others but that's my choice. I don't think your comment is helpful here, sorry. As I said having v1 and v2 in the code is clean; See issues #3401 for my points. I've been managing all iop-order issues until there, we are close to a full resolution, the fix in this PR is simple. The original code was wrong, two lines removed and the iop-order is now correct. At least this PR is not obscure.

TurboGit:
@junkyardsparkle : thanks for testing!

parafin:
I pretty sure that this PR changed definition of iop order v3 and v4. You have removed colorout moving before clahe, and colorout isn't touched in v3 definition. So sure, ignore my questions...

TurboGit:
@parafin : again v3 and v4 are not final iop-order just transient mess needed for devs having used some random dt 3.0. So all v2 will go through v3, v4, and v5 which is properly defined and reordered.

TurboGit:
And I'm not saying that there is no remaining issues, right. I'm saying that I don't think there is one, if you prove me wrong no problem I'll fix. Just send an XMP to reproduce.

parafin:
Bug-driven development model, I see...

OK, so if I understand correctly, you are saying that it doesn't matter that iop order v3 changed, because it's not used for new images anyway, and there's no conversion from v2 to v3 in darktable's code which will allow moving iops. Which basically means that all lines 111-196 in iop_order.c serve no purpose at all. And you're calling this PR not obscure?? This logic is so complicated, that any change is obscure in this code - it's very hard to predict consequences.

TurboGit:
> Bug-driven development model, I see...

Not at all, but if you can foresee any issue and do everything right, go ahead... I can't.

I never said v3 and v4 were not used. Please reread my messages.

parafin:
I read all your messages and also relevant code. What will happen if we remove lines 111-196 from iop_order.c?

TurboGit:
All devs having migrated to v3 will have broken edits.
All devs having migrated to v4 will have broken edits.

(maybe there is no much remaining v3 and v4 around, but who knows)

All this is transient mess done to avoid breaking edits of people building master versions between IIRC Sep to Oct. This is a general policy we always try to avoid breaking edits even for dev people. The stabilization of this iop-order circuitry has been long and painful but we are getting there.

All in all, look at the final v5, it is clean and quite simple : a flat list.

But to get to this point we had lot of issues, I think you've seen a good part of them on the mailing-list and GitHub. You think you could have done better, maybe. I think that what I have done here is not that bad considering that iop-order  was not implemented by me :) At this stage I think I do understand it ok to be able to fix it and see where something will go wrong or not.

I'm not ignoring your comments but I think they were not correct because there is probably some parts you're missing in this iop-order code.

TurboGit:
One more point to understand:
- v3 and v4 iop-order are known to be broken and incorrect.

So no matter what, this is broken that's why we have made a v4 to fix v3 and that's why we have made a v5 to fix v4. Every picture broken and using v3 or v4 should either be fixed in GUI (changing iop-order) or the history clearer and redone (so using v5).

TurboGit:
Now that I think about it, maybe you thought that the migration was always done as in the iop parmas legacy upgrade? That's not the case here. Hope this clarifies the situation.

parafin:
I know all the story. Please answer the particular questions about the code I'm asking. Or don't answer it if you don't want my review on this (or any iop-order-related) PR.

TurboGit:
I've already answered just above !

All devs having migrated to v3 will have broken edits. 
All devs having migrated to v4 will have broken edits.

parafin:
> Now that I think about it, maybe you thought that the migration was always done as in the iop parmas legacy upgrade? That's not the case here. Hope this clarifies the situation.

https://github.com/darktable-org/darktable/blob/2bca84cbdbb84d0a4aadfd95e9a76dfcb305f3f7/src/develop/develop.c#L1740 - what is this then? By design dt_ioppr_legacy_iop_order should handle missing from old iop order modules, that's why I'm saying that https://github.com/darktable-org/darktable/commit/356259332285b83faee87fecdfac62d7229c581e looks incorrect and should be reverted. dt_ioppr_legacy_iop_order doesn't move modules, so it's not enough to fix v3/v4 mess and that's why dt_ioppr_convert_onthefly was introduced - it will move modules when upgrading iop order. You abused it for v1->v2 conversion, that's why we had bug #3401.

parafin:
> I've already answered just above !
> 
> All devs having migrated to v3 will have broken edits.
> All devs having migrated to v4 will have broken edits.

I don't understand what this means. Let's say user have an v1 image, v3 image and new unedited image. Will removing lines 111-196 from iop_order.c affect him in any way when he opens these images?

- If not, then this PR is fine. Which also means that those lines can be just dropped (that's my current understanding)
- If it affects him, then this PR affects him also. Because you removed `_ioppr_move_iop_before(_iop_order_list, "colorout", "clahe", dont_move);`, meaning v1->v2 conversion changed, which in turn changed v1->v3 conversion, which means v3 order is different.

TurboGit:
Sorry @parafin, your message is all in the form... I don't understand so revert...

Not helpful. Please can you say why this is wrong?

`dt_ioppr_legacy_iop_order(&dev->iop, &dev->iop_order_list, dev->history, dev->iop_order_version);
`
In the line above dev->iop_order_version is the iop-order version of the loaded image (so either v2 or v5 at this stage). And so yes we had new modules here, that's why they must be added into v2 and v5 (full flat list).

If you think something is wrong please describes a scenario where it is wrong not just some guessing that there is issues. somewhere to justify a revert.

TurboGit:
To you question :

You cannot remove as I said because if people have a v3 it won't work and will broke even more v3 edits. But v3 is already broken (possible wrong order) anyway so we don't really care.

If people have only v1 and unedited images we could indeed remove those lines.

TurboGit:
> v1->v3 conversion

Will never happen.

parafin:
> Sorry @parafin, your message is all in the form... I don't understand so revert...
> 
> Not helpful. Please can you say why this is wrong?

I am explaining why this is wrong.

> `dt_ioppr_legacy_iop_order(&dev->iop, &dev->iop_order_list, dev->history, dev->iop_order_version); `
> In the line above dev->iop_order_version is the iop-order version of the loaded image (so either v2 or v5 at this stage). And so yes we had new modules here, that's why they must be added into v2 and v5 (full flat list).

Before https://github.com/darktable-org/darktable/commit/356259332285b83faee87fecdfac62d7229c581e there were 3 possibilities - v1, v2 and v5. This function adds priorities of iops added in higher versions of iop order (see comment in develop.c), so I don't see why you have to forcibly convert v1 to v2 in dt_ioppr_convert_onthefly

parafin:
> > v1->v3 conversion
> 
> Will never happen.

This conversion is used in dt_ioppr_get_iop_order_list to get v3 iop order list.

parafin:
> To you question :
> 
> You cannot remove as I said because if people have a v3 it won't work and will broke even more v3 edits. But v3 is already broken (possible wrong order) anyway so we don't really care.

How exactly it will break? What will change?

TurboGit:
> How exactly it will break? What will change?

Probably badly as some part of the code will request an order for v3 which will not exists anymore.

parafin:
> > How exactly it will break? What will change?
> 
> Probably badly as some part of the code will request an order for v3 which will not exists anymore.

Then I'm going to repeat once again - this PR also affects v3 order, so by your logic it will also break v3 images in some way.

TurboGit:
> This conversion is used in dt_ioppr_get_iop_order_list

Not exactly. The routine dt_ioppr_get_iop_order_list get a parameter:

- if 0 it will create a v5 (going through v3 & v4 but redefined in v5 as a flat and controlled list).
All fine.
- if 5 : all fine too (same case as above)
- if 2 : all fine

- if 3 or 4 : broken as expected as v3 and v4 are broken.
BUT : routine dt_ioppr_convert_onthefly() will convert all edits in v3 and v4 to v5. So final order is once more defined as a flat and controlled list. So all fine again.

TurboGit:
> Then I'm going to repeat once again - this PR also affects v3 order, so by your logic it will also break v3 images in some way.

Then I'm going to repeat one again - this PR does not affect v3 order in any way that will broke edit as converted to v5 on-the-fly.

parafin:
Yes, this analysis looks correct to me and that's what I understood in the end too, but it wasn't obvious at all from the code. Also it means that lines 111-196 don't affect anything.

Issue #3323, opened 2019-11-08
After MSYS2 Update CMAKE errors in windows current build


chhil:
**Describe the bug**

-- Configuring done
CMake Error in src/external/rawspeed/CMakeLists.txt:
  Imported target "Pugixml::Pugixml" includes non-existent path

    "F:/msys64/mingw64/include/pugixml-1.9"

  in its INTERFACE_INCLUDE_DIRECTORIES.  Possible reasons include:

  * The path was deleted, renamed, or moved to another location.

  * An install or uninstall procedure did not complete successfully.

  * The installation package was faulty and references files it does not
  provide.

CMake Error in src/external/rawspeed/CMakeLists.txt:
  Imported target "Pugixml::Pugixml" includes non-existent path

    "F:/msys64/mingw64/include/pugixml-1.9"

  in its INTERFACE_INCLUDE_DIRECTORIES.  Possible reasons include:

  * The path was deleted, renamed, or moved to another location.

  * An install or uninstall procedure did not complete successfully.

  * The installation package was faulty and references files it does not
  provide.

chhil:
The build fails in the end with...

[ 93%] Checking validity of noiseprofiles.json
Traceback (most recent call last):
  File "F:\msys64\mingw64\bin\jsonschema-script.py", line 6, in <module>
    from pkg_resources import load_entry_point
  File "F:/msys64/mingw64/lib/python3.8/site-packages\pkg_resources\__init__.py", line 3251, in <module>
    def _initialize_master_working_set():
  File "F:/msys64/mingw64/lib/python3.8/site-packages\pkg_resources\__init__.py", line 3234, in _call_aside
    f(*args, **kwargs)
  File "F:/msys64/mingw64/lib/python3.8/site-packages\pkg_resources\__init__.py", line 3263, in _initialize_master_working_set
    working_set = WorkingSet._build_master()
  File "F:/msys64/mingw64/lib/python3.8/site-packages\pkg_resources\__init__.py", line 583, in _build_master
    ws.require(__requires__)
  File "F:/msys64/mingw64/lib/python3.8/site-packages\pkg_resources\__init__.py", line 900, in require
    needed = self.resolve(parse_requirements(requirements))
  File "F:/msys64/mingw64/lib/python3.8/site-packages\pkg_resources\__init__.py", line 786, in resolve
    raise DistributionNotFound(req, requirers)
[...]

PR #3410, opened 2019-11-13
Update windows build instructions for recent MSYS update

mbaumgae:
Added mingw-w64-x86_64-python3-importlib-metadata and mingw-w64-x86_64-python3-more-itertools package to required package list.
Both are required for the update to Python 3.8.

See issue #3323

### DIFF
--- a/packaging/windows/BUILD.txt
+++ b/packaging/windows/BUILD.txt
@@ -14,7 +14,7 @@ A) Native compile using MSYS2:  How to make a darktable windows installer (64 bi
 	$ pacman -S git
 
     Install required libraries and dependencies:
-	$ pacman -S mingw-w64-x86_64-{exiv2,lcms2,lensfun,dbus-glib,openexr,sqlite3,libxslt,libsoup,libwebp,libsecret,lua,graphicsmagick,openjpeg2,gtk3,pugixml,libexif,osm-gps-map,libgphoto2,flickcurl,drmingw,gettext,python3,iso-codes,python3-jsonschema,python3-setuptools}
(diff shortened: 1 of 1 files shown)


leclercj:
Line 32, I think there is a need to replace 'python3.7' by 'python3.8'

peterbud:
You don't need this anymore, the python package dependency has been fixed in MSYS2 at least for more-itertool:
https://github.com/msys2/MINGW-packages/pull/5950

The right approach is to fix dependencies.

leclercj:
Dependencies are fixed on MINGW-packages git, but not yet released and available on official website: https://msys2.github.io/
So this workaround seems still required today in these instructions

peterbud:
The fixed python-zipp is available already at the repo:
http://repo.msys2.org/mingw/x86_64/mingw-w64-x86_64-python3-zipp-0.6.0-2-any.pkg.tar.xz

I'll add the missing `pyhton-zipp` dependency to `importlib-metadata` and then it's enough the existing `python3-jsonschema`, as it will install automatically python-path->python-importlib-metadata->python-zipp->python-more-itertools

So if you follow the description and listed python packages will automatically install their dependencies.

mbaumgae:
Thanks @peterbud for the explanation and the fix to the package in MSYS github. Your explanations are very much appreciated since they help to understand a lot of the MSYS environment!
I'll close this PR, since the root cause will be fixed soon.

leclercj:
Ok, so that this PR is not needed anymore because issue will be completely fixed shortly on MSYS2.
Just remain the need to change 'python3.7' by 'python3.8', line 32 ?

peterbud:
Actually, the updated importlib-metadata with the python-zipp dependency is already in the repo since 2 weeks, so no change is needed, and the existing dependencies are correct:
http://repo.msys2.org/mingw/x86_64/mingw-w64-x86_64-python3-importlib-metadata-0.23-2-any.pkg.tar.xz

@leclercj  the python3.7->python3.8 are needed. The latest lensfun package is unfortunately not usable yet on MSYS2.

PR #3419, opened 2019-11-14
RELEASE_NOTES: propose a few minor language tweaks

junkyardsparkle:
@aurelienpierre Can you comment on whether these changes are OK for you? I'm just trying to streamline some longer parts, hoping to increase the chances of all of this actually being read. :-)

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -47,16 +47,16 @@ making a backup is strongly advised.
      - darktable-icons-dark     : lighter version. with module icons
      - darktable-icons-grey     : even lighter version, with module icons
 
-  New shortcuts have been introduced to quickely collapse borders, sidebars, histogram and 
+  New shortcuts have been introduced to quickly collapse borders, sidebars, histogram and 
   navigation modules, allowing a new borderless editing experience.
   
   Note that the new GUI requires Gtk+ 3.22 or higher to work properly.
   
-- It is now possible to associate dynamic key shortcuts to sliders, and then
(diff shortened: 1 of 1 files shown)


RELEASE_NOTES.md
```diff
@@ -105,22 +105,22 @@ making a backup is strongly advised.
...
 
 - A new module 'filmic RGB' which, like the previous 'filmic', is designed
-  to replace the 'base curve', 'shadows and highlights' modules and other tone-mapping options. 
+  to replace 'base curve', 'shadows and highlights' and other tone-mapping modules. 
   This new version replaces the one introduced in 2.6.2; it should be easier to use,
   and it will reduce color casts. The old 'filmic' module is now deprecated and
   is only available on images where it was already used for editing.
 
-- A new module 'tone equalizer', designed to merge the features of 'local tone mapping', 
+- A new module 'tone equalizer', designed to merge the features of 'tone mapping',
```
aurelienpierre: the module is called "tone mapping", but it's actually a local tone mapping, so it should be put at least : 
(local) 'tone mapping'

junkyardsparkle: Ok. About that distinction: should filmic rgb be said to replace other *global* tone mapping modules, or is that too specific in that case?

aurelienpierre: Yes, indeed, filmic is a global tone mapping.

junkyardsparkle: Done.

TurboGit:
Thanks !

PR #3420, opened 2019-11-14
Fix undeclared side effect in src/common/cpuid.c

wierton:
In inline assembly at line src/common/cpuid.c:80, the instruction setne will override register eax, this side effect may corrupt the normal data flow in certain context. For example, assume somebody incidently write a assignment expression which assigned from a function call, the return value of function call will use the register eax which will cause data flow corruption under some context:
```
  int val = printf("Hello world\n");
  __asm volatile(
      "pushf\n"
      "pop     %0\n"
      "mov     %0, %1\n"
      "xor     $0x00200000, %0\n"
      "push    %0\n"
      "popf\n"
      "pushf\n"
      "pop     %0\n"
      "cmp     %0, %1\n"
      "setne   %%al\n"
      "movzb   %%al, %0\n"
      : "=r"(ax), "=r"(tmp));
  printf("%d\n", val); // should be 12
```
The above code will produce output 1 when compiled with flag -O2 where
it should be 12.

### DIFF
--- a/src/common/cpuid.c
+++ b/src/common/cpuid.c
@@ -79,7 +79,7 @@ dt_cpu_flags_t dt_detect_cpu_features()
                    "cmp     %0, %1\n"
                    "setne   %%al\n"
                    "movzb   %%al, %0\n"
-                   : "=r"(ax), "=r"(tmp));
+                   : "=r"(ax), "=r"(tmp) : : "al");
 
     if(ax)
     {


TurboGit:
Sounds ok, but I'd like to have a second vote for this to be merged as I'm really not expert here.

Anyone with assembly expertise?

aurelienpierre:
While we are at it, 
```
if(cx & 0x00040000) cpuflags |= CPU_FLAG_SSE4_1;
 if(cx & 0x00080000) cpuflags |= CPU_FLAG_SSE4_2;
```

could be updated with AVX, AVX2 and AVX512 too.

LebedevRI:
Just use [`__get_cpuid()`](https://github.com/darktable-org/rawspeed/blob/f134594aa7979d821528657b7ffdd6c819492a20/src/librawspeed/common/Cpuid.cpp#L23-L41)?

jenshannoschwalm:
> Just use __get_cpuid()?

What shall be done with this pr? Create a new one with the above function? I looked at the docs and it seems to be supported on gcc and clang.

TurboGit:
I would just do the new implementation on this PR and force push this here.

jenshannoschwalm:
I don't know how to force-push safely from my account. Instead i have uploaded the source so you could manage ...


[cpuid.c.txt](https://github.com/darktable-org/darktable/files/3870750/cpuid.c.txt)

cryptomilk:
`git push -f myaccount mybranch`

for exmaple

`git push -f wierton master`

jenshannoschwalm:
This means writing into someone else's master branch? really?

TurboGit:
@jenshannoschwalm : I'll have a look, thanks.

TurboGit:
@jenshannoschwalm : closing this, I made your work available here #3564.

PR #3441, opened 2019-11-15
Fix OpenCL compile errors on Windows

### LINKED ISSUE #3384: Windows super slow startup (55 secs) - v3.0.0rc1 always compiles opencl .cl files on app run
CacheDarktable

**Describe the bug**
The problem I see in the debug log is that Darktable compiles the .cl files every time I open it. It can't find and load the cached compiled files because they are saved with a different name as the one Darktable expects.

This happens with all the .cl files located in: 'C:\Users\USER_NAME\AppData\Local\Microsoft\Windows\INetCache\darktable\cached_kernels_for_IntelRHDGraphics400_2019155058\'

For example:
'atrous.cl.bin' is compiled with the following name: 'atrous.cl.bin.ea42012ca03bfaa433b24a71ddf4ae34'.
The next time you launch darktable, it can't find 'atrous.cl.bin', so it compiles it again.

I deleted the cached folder, launched Darktable several times and watched what happened to the .bin files.

. The fist time you launch Darktable, it saves the .bin files with a long string appended at the end of the file, example: 'atrous.cl.bin' is saved as 'atrous.cl.bin.ea42012ca03bfaa433b24a71ddf4ae34'.

. The second time, it can't find the .bin and compiles again but this time it saves the .bin file only with the long string as name, example: 'atrous.cl.bin' is saved as 'ea42012ca03bfaa433b24a71ddf4ae34'.

. The third time it does the same as the second time, it can't find the .bin, so it compiles and again, saves the .bin files with a long number as name, example: 'atrous.cl.bin' is saved ea42012ca03bfaa433b24a71ddf4ae34'. And it replaces the previous .bin file that had this name.

**UPDATE**
..........................................
If I **rename the .bin files manually** with the correct name (without the appended string), DT doesn't find the .bin files and compiles the .cl files again.

**So .cl files are compiled, always, no matter what. And the startup time is always, 55 to 60 seconds**.
.........................................

I append a sample of the log output below.

Also, I see that even if you have opencl disabled, Darktable compiles and loads all the .cl modules, why? what for?...

Thank you and keep up the good work! :-)
...
Sample log output:

4.798949 [opencl_build_program] saving binary4.832309 [opencl_init] compiling program `atrous.cl'
4.832309 [opencl_load_program] could not load cached binary program from file `C:\Users\USER_NAME\AppData\Local\Microsoft\Windows\INetCache\darktable\cached_kernels_for_IntelRHDGraphics400_2019155058\atrous.cl.bin'! (-42)
4.838943 [opencl_load_program] could not load cached binary program, trying to compile source
4.845680 [opencl_load_program] successfully loaded program from `C:\Program Files\darktable3rc0\share\darktable\kernels\atrous.cl'
5.579130 [opencl_build_program] successfully built program
5.579130 [opencl_build_program] BUILD STATUS: 0
5.579130 BUILD LOG:
5.579130 <built-in>:1:9: warning: '__FINITE_MATH_ONLY__' macro redefined<built-in>:113:9: note: previous definition is herefcl build 1 succeeded.fcl build 2 succeeded.bcl build succeeded.
....

**To Reproduce**
Just launch Darktable.

**Expected behavior**
Darktable should load the cached .bin files.

. Darktable 3.0.0rc1 executed as normal user (not as administrator)
. Windows 10 updated with the latest patches
. opencl 1.2 reported in the debug log (opencl dlls in my system show 2.0.2 as version).

peterbud:
Fixes #3384 

This fix handles the following issues:
* Using binary flag for g_fopen() when producing OpenCL compiled/cached binaries
* Properly deleting the OpenCL compiled/cached binaries
* Also adding a bit more diagnostic

### DIFF
--- a/src/common/opencl.c
+++ b/src/common/opencl.c
@@ -1838,7 +1838,7 @@ int dt_opencl_build_program(const int dev, const int prog, const char *binname, 
           //store has using a simple filerename
           char finalfilename[PATH_MAX] = { 0 };
           snprintf(finalfilename, sizeof(finalfilename), "%s" G_DIR_SEPARATOR_S "%s.%s", cachedir, bname, md5sum);
-          if(g_rename(link_dest, finalfilename) != 0) goto ret;
+          rename(link_dest, finalfilename);
 #else
           if(symlink(md5sum, bname) != 0) goto ret;
 #endif //!defined(_WIN32)
--- a/src/common/opencl.c
+++ b/src/common/opencl.c
@@ -1655,7 +1655,7 @@ int dt_opencl_load_program(const int dev, const int prog, const char *filename, 
         rd = fread(cached_content, sizeof(char), cached_filesize, cached);
         if(rd != cached_filesize)
         {
-          dt_print(DT_DEBUG_OPENCL, "[opencl_load_program] could not read all of file `%s'!\n", binname);
+          dt_print(DT_DEBUG_OPENCL, "[opencl_load_program] could not read all of file '%s' MD5: %s!\n", binname, md5sum);
         }
         else
         {
@@ -1665,8 +1665,8 @@ int dt_opencl_load_program(const int dev, const int prog, const char *filename, 
           if(err != CL_SUCCESS)
           {
             dt_print(DT_DEBUG_OPENCL,
-                     "[opencl_load_program] could not load cached binary program from file `%s'! (%d)\n",
-                     binname, err);
+                     "[opencl_load_program] could not load cached binary program from file '%s' MD5: '%s'! (%d)\n",
+                     binname, md5sum, err);
           }
           else
           {
@@ -1685,6 +1685,7 @@ int dt_opencl_load_program(const int dev, const int prog, const char *filename, 
   {
     // if loading cached was unsuccessful for whatever reason,
     // try to remove cached binary & link
+#if !defined(_WIN32)
     if(linkedfile_len > 0)
     {
       char link_dest[PATH_MAX] = { 0 };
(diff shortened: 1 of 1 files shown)


src/common/opencl.c
```diff
@@ -1833,7 +1838,7 @@ int dt_opencl_build_program(const int dev, const int prog, const char *binname,
           //store has using a simple filerename
           char finalfilename[PATH_MAX] = { 0 };
           snprintf(finalfilename, sizeof(finalfilename), "%s" G_DIR_SEPARATOR_S "%s.%s", cachedir, bname, md5sum);
-          rename(link_dest, finalfilename);
+          if(g_rename(link_dest, finalfilename) != 0) goto ret;
```
parafin: Yeah, this doesn't help, and actually hurts because of chdir. chdir actually should be removed from this code, it doesn't make sense.

peterbud: I'm not sure I can follow you. What is the issue with the chdir? That was in the original code, which I have not touched. Could you pls elaborate a bit?

parafin: Original code is bad - it shouldn't do chdir. By jumping out with "goto ret" second chdir (restoring original working directory) is skipped, so this code can end up changing the working directory. That's why I'm saying you made it worse. Doing nothing actually was better in this case. If you want to handle rename error, you need to think how to do it, not just copy code from neighbourhood - there's no guarantee that original code makes sense. E.g. you can at least print some error message. Maybe you could try to remove the finalfilename if it exists, though it makes sense to do before calling rename, not after it fails.
To summarize - think about what you are trying to achieve here. Checking return code if you don't know what to do with it doesn't help anyone, it looks like this code handles a error condition, when in reality it doesn't.

peterbud: Fair comment. Let's then just focus on the issue at hand and leave fixing the potential rename error together with the fixing of the original code later.

parafin: Yes, just leave it as it is for now.

PR #3464, opened 2019-11-17
lighttable : don't reset offset if it has just been changed

### LINKED ISSUE #3461: Recently used collections' saved positions are ignored.
**Describe the bug**
When a collection in the "recently used collections" module is clicked, the light table is always positioned at the first image in the collection, ignoring the previous position as stored in the `plugins/lighttable/recentcollect/pos...` line.

**To Reproduce**
1. Open a collection.
1. Scroll to a position in which the first image in the collection is not visible.
1. Switch to another collection.
1. Click on the first collection's entry in the "recently used collections" module.
1. Note that the light table is positioned at the first image in the collection.

**Expected behavior**
The light table should have returned to the same position as it was in at the end of step 2.

 - OS: macOS 10.14
 - Version: 3.0.0rc1

AlicVB:
fix #3461

I just hope I've not broken another use case...


TurboGit:
Looks ok, and I haven't seen regressions :)

Issue #3463, opened 2019-11-17
open and close a picture with Darkroom kills the stack


Haribo-66:
**Describe the bug**
open and close a picture with Darkroom kills the whole stack
OR open a picture in darkroom and compress history within Darkroom mode kills whole history stack!

**To Reproduce**
1. Go to 'Lighttable' and press D, return from Darkroom to Lighttable and the history stack is gone forever.

OR got to Darkroom and press compress history stack and it is completely gone.

**Expected behavior**
Do nothing. Keep or compress the stack.

**Screenshots**
before pressing compress stack button
![stack before compress](https://user-images.githubusercontent.com/53818783/69006698-fdd79400-0932-11ea-8d47-5b3bea58b541.jpg)
after pressing compress stack button
![stack after compress](https://user-images.githubusercontent.com/53818783/69006699-fdd79400-0932-11ea-8c4e-634c787b7241.jpg)

**Platform (please complete the following information):**
darktable 3.0.0 rc1
Windows 10 prof. / OS Build 18362.476

**Additional context**
open a picture in darkroom and compress history within Darkroom mode kills whole history stack!

I did an intense run using "darktable-generate-cache" (~130.000 pictures) seen lots of on the fly conversions v1 to v2 and V4 to V5 and more error messages, remembered image IDs and opened some.

Haribo-66:
I found 2 messages (warning + critical) during start up

![capture_11172019_151458](https://user-images.githubusercontent.com/53818783/69008664-721f3100-094d-11ea-9bc0-0b17b661851c.jpg)

leclercj:
On Windows 10 I've these 2 error messages at darktable startup (1 warning + 1 critical), but I do not reproduce this issue. Hower, I not using a database migrated from 2.6 but a fresh one
Do you have same problem with fresh image in dt3.0 ?

leclercj:
Adding @peterbud to comment on this gdbus error on Windows porting, if possible.

Haribo-66:
I tried a separate installation, deleted the existing database (version 19) and a new one was created during start up, then I reimported a folder -> dt is working as expected now -> problem theoretically solved.

Is there any idea available to not reimport everything into an empty database?
 
Seems that there was / is (?) a bug in the dt 2.6 to dt 3 database conversion.

peterbud:
> Adding @peterbud to comment on this gdbus error on Windows porting, if possible.

Sorry, how the gbus message is related to the issue presented here?

[...]

PR #3468, opened 2019-11-17
dt_image_cache_write_release should be locked per image.

jenshannoschwalm:
Might be related to #3463

### DIFF
--- a/src/common/image_cache.c
+++ b/src/common/image_cache.c
@@ -219,6 +219,7 @@ void dt_image_cache_write_release(dt_image_cache_t *cache, dt_image_t *img, dt_i
       uint32_t u;
   } flip;
   if(img->id <= 0) return;
+  dt_lock_image(img->id);
   sqlite3_stmt *stmt;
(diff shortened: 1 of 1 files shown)


jenshannoschwalm:
Could be triggered by darktable-generate-cache

TurboGit:
Looks ok and cannot harm. Thanks!

jenshannoschwalm:
@Turbogit Unfortunately this does harm in a way i don't understand. When i import a folder it locks, i thought maybe img->id is changed magically while writing the sidecar file or cache_release but this isn't the case. I can unlock_image right after sqlite3_finalize(stmt); and it keeps locking.

Absolutely no idea right now by looking at the code. What does this mean?

TurboGit:
@jenshannoschwalm : Not sure to see what the lock occurs, but the seems that dt_cache* are already ensuring that the calls are thread safe (dt_cache_get / dt_cache_[read/write]_release).

I'll revert that for now anyway as it prevent importing folders :(

PR #3470, opened 2019-11-17
Change order of image info line

fvollmer:
This is a very minor change. The image info line had a different order than the info line you get if you hover your mouse pointer over the image.

Before:
image

After:
image


TurboGit:
Every details count! Thanks!

PR #3471, opened 2019-11-17
Capitalize iso at dt_image_print_exif

fvollmer:
Another minor change. I noticed that the capitalization of ISO is different for the mouse over:

image


TurboGit:
Thanks! Because string freeze this will have to wait for 3.0.1 release.

parafin:
Looks like these lines are non-translatable, so string-freeze is irrelevant.

TurboGit:
@parafin : Indeed! So let's merge. Thanks.

PR #3474, opened 2019-11-18
bilateral: fix implemetation to correspond to OpenCL version.

### LINKED ISSUE #3421: Verify CPU and GPU results
Do we make the assumption that an image processed by the CPU should yield the same result as the image processed by the GPU?

Validate the outputs of CPU and GPU operations.  Create an automated testing program that will run each iop on the CPU and GPU and compare the results for "significant" differences.  I think this can be accomplished relatively easily using the darktable-cli interface.

I'm willing to implement a proof a concept if the maintainers of this project see value in this sort of testing.

TurboGit:
The CPU version was brolen by recent changes. This patch revert part
of it to have bilateral CPU identical to OpenCL version.

Fixes #3421.

### DIFF
--- a/src/common/bilateral.c
+++ b/src/common/bilateral.c
@@ -128,7 +128,6 @@ void dt_bilateral_splat(dt_bilateral_t *b, const float *const in)
   const int ox = 1;
   const int oy = b->size_x;
   const int oz = b->size_y * b->size_x;
-
   const float sigma_s = b->sigma_s * b->sigma_s;
   float *const buf = b->buf;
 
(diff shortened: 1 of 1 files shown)


aurelienpierre:
I think you just need to revert:
* 6578d8b
* 0162acf
* a55adaf

for `bilateral.c`.

TurboGit:
@aurelienpierre : yes I know, it is working when I do that but I try to recover part of the OpenMP speedup :) I just need to please clang :)

aurelienpierre:
It was foolish on me to try optimizing anything before having a regression-test tool to validate the output is the same, so I suggest we just ditch all that for now, get an Image Magick script or something to compute RMSE on the exports, then give it another try.

TurboGit:
@aurelienpierre : right, a regression tests framework would be nice.

Anyway, all clean now. I'll merge for the release.

Issue #3455, opened 2019-11-16
make darktable 3.0 use a different database


aurelienpierre:
Many bugs have affected the 2.6 and 2.7 database migration to 3.0, between styles, pipeline order, etc. I fear we will not have time to defuse all the bombs before release, or at least we should consider what to do if we can't.

Once the db migration 2.x -> 3.x  is done, there is no coming back and darktable 2.x will not run on a 3.x DB.

As a safety measure, I propose darktable 3.0 series copies the 2.x `library.sql` database to `library-3.sql` and update the DB structure of this DB, while leaving the original library as is. This way, users will be able to still use dt 2.x in parallel as a backup measure if 3.0 fails to import their old edits properly (or breaks them), and will still have the option to reimport the 2.x edits to the 3.x series once all bugs are fixed (maybe adding a button in UI to redo the import from 2.x to 3.x upon user request).

Pinging @jenshannoschwalm @TurboGit @parafin as they took part in this.

parafin:
Idea isn’t bad, but using 2.6 and 3.0 in parallel also presents a problem with XML sidecar files. They are also incompatible and they will be overwritten by each darktable version in this case. I propose not to give an impression that this is workable usage of darktable and instead backup old database to library-2.6.db while continuing to use library.db.

aurelienpierre:
Aaah… forgot about those bloody XMP. Isn't there some way to tag the XMP fields with `darktable:2` or `darktable:3` classes ?

TurboGit:
This just looks like a giant can of worms :)

TurboGit:
And anyway, too late now. We cannot start such change for 3.0.

c-h-r-i-s-t-i-a-n:
I fear this will introduce more new problems as it solves.

jenshannoschwalm:
Nothing to add. Also i am confident we have either fixed everything yet or will be able to do so.

aurelienpierre:
> Nothing to add. Also i am confident we have either fixed everything yet or will be able to do so.

None of us saw the pile of bugs we got these last few weeks coming. Trust, faith or confidence are out of the plan right now: we need a fallback in case things go south.

> And anyway, far too late now. We cannot start such change for 3.0.

[...]

PR #3480, opened 2019-11-18
database: backup db for each new release.

TurboGit:
We backup library.db and data.db with to library.db-pre-<version> and
data.db-pre-<version>. The version string here is the version of the
current darktable release (using MAJOR.MINOR.PATCH, e.g. 3.0.0).

For #3455.

### DIFF
--- a/src/common/database.c
+++ b/src/common/database.c
@@ -17,6 +17,10 @@
     along with darktable.  If not, see <http://www.gnu.org/licenses/>.
 */
 
+#ifdef HAVE_CONFIG_H
+#include "config.h"
+#endif
+
 #include "common/database.h"
 #include "common/darktable.h"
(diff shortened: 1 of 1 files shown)


TurboGit:
Discussed in #3455, but I'm really not convinced that this will help. The XMP modifed by the new release won't be read by the old one... And may even crash dt I suppose.

parafin:
There are limitations to usefulness of this, yes, but it won't hurt. I think if re-scan option isn't enabled in preferences, then DB take precedence and XMPs will be just overwritten on edit. Also one can just delete XMPs of the images present in DB.

aurelienpierre:
Great !

TurboGit:
Ok, let's this in then.

Issue #3473, opened 2019-11-18
Framing module crash (RC1)


c-h-r-i-s-t-i-a-n:
This seems to work in any image:

- switch on the frame tool in Darktable 3.0.0rc1-win64
- adjust frameline size > 0%
- adjust frameline offset = 100%
- get a crash

**Screenshots**

![Bild1](https://user-images.githubusercontent.com/57685590/69041153-4b193b80-09ef-11ea-8781-94ade5025aef.png)

**Platform (please complete the following information):**
 - OS: Windows-10 / 64bit
 - Version 3.0RC1

**Additional context**
This was originally reported on the dev mailing list.

leclercj:
This I can reproduce easily
backtrack attached here:
[darktable_bt_FWLFB0.txt](https://github.com/darktable-org/darktable/files/3858447/darktable_bt_FWLFB0.txt)

c-h-r-i-s-t-i-a-n:
How can I add the red label "BUG" to this issue?

aurelienpierre:
Reproduced on Linux without OpenCL, doesn't reproduce with OpenCL

aurelienpierre:
Preliminary report:
```
double free or corruption (out)
#0  0x00007ffff7ab3e35 in raise () at /lib64/libc.so.6
#1  0x00007ffff7a9e895 in abort () at /lib64/libc.so.6
#2  0x00007ffff7af723f in __libc_message () at /lib64/libc.so.6
#3  0x00007ffff7afe7bc in  () at /lib64/libc.so.6
#4  0x00007ffff7b001a8 in _int_free () at /lib64/libc.so.6
#5  0x00007ffff7e26e68 in dt_dev_pixelpipe_cache_get_weighted (cache=0x34fe0b0, hash=0, size=21788832, data=0x7fffc57eb3d0, dsc=0x7fffc57eb450, weight=weight@entry=0)
    at /home/aurelienpierre/Documents/Programmes/darktable/src/develop/pixelpipe_cache.c:170
        max_used = <optimized out>
        max = <optimized out>
        sz = <optimized out>
```

Faulty line is 
```dt_free_align(cache->data[max]);```, 
yet adding
```if(cache->data[max] != NULL) dt_free_align(cache->data[max]);```
doesn't solve it.

parafin:
Freeing NULL pointer doesn't crash. There's no need for that check.

aurelienpierre:
How do you check for double freeing then ?

[...]

PR #3484, opened 2019-11-18
Fix frame module crash (#3473)

leclercj:
Fix for https://github.com/darktable-org/darktable/issues/3473
Tested on Linux and Windows 10.


TurboGit:
Looks ok! Thanks.

PR #3485, opened 2019-11-19
Added check for scroll option to export spin boxes

### LINKED ISSUE #3348: "mousewheel scrolls modules side panel by default" ignored in "export selected"
**Describe the bug**
The export selected module ignores the "mousewheel scrolls modules side panel by default" setting.
Similar to https://github.com/darktable-org/darktable/issues/3240

**To Reproduce**
1. Enable "mousewheel scrolls modules side panel by default" in main settings -> GUI options -> misc
2. Go to lighttable mode, open the export selected module and try to scroll away from it while hovering one of the "max size" boxes

**Expected behavior**
Scrolling of the modules, not changing module properties.

**Platform**
 - OS:  linux 5.3.7 (arch linux with i3 WM)
 - Darktable version 2.7.0+2073

sams96:
Fixes #3348

So it does disable scrolling in the width and height boxes but doesn't scroll the modules if you're hovering over them. I figure it's better than it was before but I couldn't figure out how to get that to work too.

### DIFF
--- a/src/libs/export.c
+++ b/src/libs/export.c
@@ -531,6 +531,14 @@ static void _lib_export_styles_changed_callback(gpointer instance, gpointer user
   g_list_free_full(styles, dt_style_free);
 }
 
+static gboolean export_scrolled(GtkWidget *widget, GdkEventScroll *event, gpointer user_data)
+{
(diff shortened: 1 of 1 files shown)


TurboGit:
I'm not sure it is better. The real fix would be to scroll the module indeed. We have the same issue in the tagging module where we still scroll the tags.

sams96:
Alright, I'll make a new PR if I can figure it out.

PR #3486, opened 2019-11-19
iop-order: invert order for multiple instances when importing.

### LINKED ISSUE #3477: Multiple module instances reversed on import from 2.6.x
**Describe the bug**
Import an edit from 2.6.x that uses multiple instances of a module.  Open the image in darkroom and the history stack is correct.  Open the active module list and the instances are in reverse order.  I've shared an image and the edit at https://drive.google.com/drive/folders/1Iq_J_a4xmNEHc9CWs3z6Ze77LcS5oDfm?usp=sharing.  It uses 5 tone curves, the last of which is a contrast curve. 

**To Reproduce**
1. Load the shared image and xmp file in a folder
2. Import the folder
3. Open the image in darkroom
4. Take a snapshot
5. Open the active module list and reverse the order of the tone curves
6. Compare the image to the snapshot and notice the difference 

**Expected behavior**
Instances of a module should be applied in the original order.  The history stack reflects the correct order.  The active module list is reversed.

 - OS: Ubuntu 18.04
 - Version 3.0.0rc1+29~g006f5384a

TurboGit:
When importing from version previous to 3.0 we do invert the
iop-order to ensure the same order as before.

Fixes #3477.

### DIFF
--- a/src/common/database.c
+++ b/src/common/database.c
@@ -39,8 +39,8 @@
 
 // whenever _create_*_schema() gets changed you HAVE to bump this version and add an update path to
 // _upgrade_*_schema_step()!
-#define CURRENT_DATABASE_VERSION_LIBRARY 19
-#define CURRENT_DATABASE_VERSION_DATA 3
+#define CURRENT_DATABASE_VERSION_LIBRARY 20
+#define CURRENT_DATABASE_VERSION_DATA 4
 
 typedef struct dt_database_t
 {
@@ -1142,6 +1142,28 @@ static int _upgrade_library_schema_step(dt_database_t *db, int version)
     sqlite3_exec(db->handle, "COMMIT", NULL, NULL, NULL);
     new_version = 19;
   }
+  else if(version == 19)
+  {
+    sqlite3_exec(db->handle, "BEGIN TRANSACTION", NULL, NULL, NULL);
+
+    // create a temp table to invert all multi_priority
+    TRY_EXEC("CREATE TEMPORARY TABLE m_prio (id INTEGER, operation VARCHAR(256), prio INTEGER)",
+             "[init] can't create temporary table for updating `history and style_items'\n");
+
+    TRY_EXEC("INSERT INTO m_prio SELECT imgid, operation, MAX(multi_priority)"
+             " FROM main.history GROUP BY imgid, operation",
+             "[init] can't populate m_prio\n");
+
+    TRY_EXEC("UPDATE main.history SET multi_priority = "
+             "(SELECT prio FROM m_prio "
+             " WHERE main.history.operation = operation AND main.history.imgid = id) - main.history.multi_priority",
+             "[init] can't update multi_priority for history\n");
+
+    TRY_EXEC("DROP TABLE m_prio", "[init] can't drop table `m_prio' from database\n");
+
+    sqlite3_exec(db->handle, "COMMIT", NULL, NULL, NULL);
+    new_version = 20;
+  }
   else
     new_version = version; // should be the fallback so that calling code sees that we are in an infinite loop
 
@@ -1250,6 +1272,30 @@ static int _upgrade_data_schema_step(dt_database_t *db, int version)
     sqlite3_exec(db->handle, "COMMIT", NULL, NULL, NULL);
 
     new_version = 3;
+  }
+  else if(version == 3)
+  {
+    sqlite3_exec(db->handle, "BEGIN TRANSACTION", NULL, NULL, NULL);
+
+    // create a temp table to invert all multi_priority
+    TRY_EXEC("CREATE TEMPORARY TABLE m_prio (id INTEGER, operation VARCHAR(256), prio INTEGER)",
+             "[init] can't create temporary table for updating `history and style_items'\n");
+
+    TRY_EXEC("INSERT INTO m_prio SELECT styleid, operation, MAX(multi_priority)"
+             " FROM data.style_items GROUP BY styleid, operation",
+             "[init] can't populate m_prio\n");
+
+    // update multi_priority for style items and history
+    TRY_EXEC("UPDATE data.style_items SET multi_priority = "
+             "(SELECT prio FROM m_prio "
+             " WHERE data.style_items.operation = operation AND data.style_items.styleid = id)"
+             " - data.style_items.multi_priority",
+             "[init] can't update multi_priority for style_items\n");
+
+    TRY_EXEC("DROP TABLE m_prio", "[init] can't drop table `m_prio' from database\n");
+    sqlite3_exec(db->handle, "COMMIT", NULL, NULL, NULL);
+
+    new_version = 4;
   }
   else
     new_version = version; // should be the fallback so that calling code sees that we are in an infinite loop
--- a/src/common/database.c
+++ b/src/common/database.c
@@ -1007,6 +1007,8 @@ static int _upgrade_library_schema_step(dt_database_t *db, int version)
   }
   else if(version == 17)
   {
+    sqlite3_exec(db->handle, "BEGIN TRANSACTION", NULL, NULL, NULL);
+
     ////////////////////////////// masks history
     TRY_EXEC("CREATE TABLE main.masks_history (imgid INTEGER, num INTEGER, formid INTEGER, form INTEGER, name VARCHAR(256), "
              "version INTEGER, points BLOB, points_count INTEGER, source BLOB)",
@@ -1121,6 +1123,8 @@ static int _upgrade_library_schema_step(dt_database_t *db, int version)
   //   }
   else if(version == 18)
   {
+    sqlite3_exec(db->handle, "BEGIN TRANSACTION", NULL, NULL, NULL);
+
     TRY_EXEC("UPDATE images SET orientation=-2 WHERE orientation=1;",
              "[init] can't update images orientation 1 from database\n");
 
@@ -1250,6 +1254,8 @@ static int _upgrade_data_schema_step(dt_database_t *db, int version)
   }
   else if(version == 2)
   {
+    sqlite3_exec(db->handle, "BEGIN TRANSACTION", NULL, NULL, NULL);
+
     //    With sqlite above or equal to 3.25.0 RENAME COLUMN can be used instead of the following code
     //    TRY_EXEC("ALTER TABLE data.tags RENAME COLUMN description TO synonyms;",
     //             "[init] can't change tags column name from description to synonyms\n");
--- a/src/common/exif.cc
+++ b/src/common/exif.cc
@@ -2493,6 +2493,22 @@ end:
   return all_ok;
 }
 
+// get MAX multi_priority
+int _get_max_multi_priority(GList *history, const char *operation)
+{
+  int max_prio = 0;
+
+  for(GList *iter = history; iter; iter = g_list_next(iter))
+  {
+    history_entry_t *entry = (history_entry_t *)iter->data;
+
+    if(!strcmp(entry->operation, operation))
+      max_prio = MAX(max_prio, entry->multi_priority);
+  }
+
+  return max_prio;
+}
+
 // need a write lock on *img (non-const) to write stars (and soon color labels).
 int dt_exif_xmp_read(dt_image_t *img, const char *filename, const int history_only)
 {
@@ -2676,7 +2692,10 @@ int dt_exif_xmp_read(dt_image_t *img, const char *filename, const int history_on
         sqlite3_bind_null(stmt, 7);
       }
       DT_DEBUG_SQLITE3_BIND_INT(stmt, 8, entry->blendop_version);
-      DT_DEBUG_SQLITE3_BIND_INT(stmt, 9, entry->multi_priority);
+      // check for max and do
+      int priority = entry->multi_priority;
+      if(entry->iop_order == -1) priority = _get_max_multi_priority(history_entries, entry->operation) - entry->multi_priority;
+      DT_DEBUG_SQLITE3_BIND_INT(stmt, 9, priority);
       if(entry->multi_name)
       {
         DT_DEBUG_SQLITE3_BIND_TEXT(stmt, 10, entry->multi_name, -1, SQLITE_TRANSIENT);
--- a/src/common/history.c
+++ b/src/common/history.c
@@ -598,7 +598,8 @@ static int _history_copy_and_paste_on_image_merge(int32_t imgid, int32_t dest_im
         if (iop_order_version_src != iop_order_version_dest)
         {
           hist->module->iop_order =
-            dt_ioppr_get_iop_order(dest_iop_list, hist->module->op) + (double)hist->module->multi_priority / 100.0f;
+            dt_ioppr_get_iop_order(dest_iop_list, hist->module->op)
+            + (double)hist->module->iop_order / 100.0f;
         }
 
         if (!dt_iop_is_hidden(hist->module))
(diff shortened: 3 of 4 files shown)


jenshannoschwalm:
Yes, i tried the same and it works. Although i was surprised about the name <-> priority "mismatch".

Also there should be a related style import bug if i remember correctly.

parafin:
So higher multi_priority means instance is lower in the stack? Makes sense. Is this field still used now that we have iop_order? Is it the only place where multi_priority is misinterpreted?

TurboGit:
@jenshannoschwalm : styles were already properly migrated. see database.c line 1211.

parafin:
Something is broken here. I just tried multi-instance on 3.0.0rc1. Clicking on "create new instance" creates an instance above the base one (as documented in the manual), gives it name "1" and priority "1" (which is logical so that we don't change these settings for the base module). iop_order is set accordingly. That's how I've expected it to work, and that's why I assumed higher multi_priority means later in the pipe. But this PR says, that multi_priority meaning actually is inverted, so iop_order and multi_instance are now in conflict with each other.
@jenshannoschwalm, could you do the same test on 2.6? Also try to re-open the image in darkroom after you add the new instance to see if order changes. Did you re-order instances when creating that XMP? I wonder why the name doesn't match the priority.

TurboGit:
@parafin : yes in 3.0 that's how it works.

parafin:
So you're saying it was changed since 2.6? This is a mess... Now multi_priority basically doesn't have a meaning (because there are 2 inverted meanings), but I bet it's used somewhere in the code.

jenshannoschwalm:
@parafin I also wonder about the mismatch. Isn't this something that should be handled in database conversion somehow? At least i - yet - don't understand the reason so i didn't do a pr as this.

@TurboGit i see the database code and am curious about the correctness.

See style.c L. 726

Unfortunately i am at work and should do intensive care on ventilators instead of thinking about this :-)

TurboGit:
@parafin : yes it has changed because the iop-order is the value that now support the ordering in the pipe and multi-priority is only a number of multi-instance not corelated from the pipe order.

TurboGit:
@jenshannoschwalm : indeed, we also need to fix styles.c ! we suspected something wrong regarding multi-instance and iop-order, we are at it :)

parafin:
Well, good luck fixing it, since, as I said, multi_priority has no meaning now. Does style save somewhere its iop_order version?

TurboGit:
And this is because iop-order on a style means nothing. We can't use it as anyway for the target image where it is applied as the order of module may have been reordered. That's why in style we have the computation of iop-order at line 726. This was added recently but was not taken into account the fact that the iop-order is reverted compared to multi-priority.

parafin:
"iop-order is reverted compared to multi-priority" was a fact in 2.6, now it's not a fact.

TurboGit:
There was no iop-order in 2.6, only multi-priority and the circuitry to compute the pipe order was not correct in 3.0 as assuming a wrong order of multi-priority.

TurboGit:
I've just proposed a fix for the style. I'll see if I can reproduce the issue without this patch and if this actually is a proper fix.

parafin:
You can't look at multi_priority for newly created styles, since it has no correlation to iop_order, doesn't it?

parafin:
There was iop order in 2.6, it was just fixed (and called differently) and therefore not recorded in the history.

parafin:
There are 3 scenarios with styles:

1. styles that were present in database when upgrading from 2.6 to 3.0;
2. styles created in 3.0
3. styles created and exported in 2.6, then imported in 3.0

with the following situation:

1. multi_priority is inversed compared to iop_order, iop_order correctly set
2. multi_priority has no correlation to iop_order, but iop_order is correctly set
3. multi_priority has inversed meaning compared to iop_order, but iop_order incorrectly calculated here: https://github.com/darktable-org/darktable/blob/master/src/common/styles.c#L1268 (I hope I tracked it correctly).

I see 2 ways to fix this mess (after fixing the already mentioned bug):

1. forget about multi_priority when applying the style and look only at iop_order;
2. re-calculate multi_priority from iop_order at some point and continue to use multi_priority in code that applies the style.

First option is better, because it would be strange to have multi_priority mean one thing in one place and another thing in all the other places.

There is of course a third option - leave multi instance broken in styles as iop order is broken. But it would be a regression compared to 2.6.

TurboGit:
>     1. forget about multi_priority when applying the style and look only at iop_order;

Not possible as I explained.

parafin:
Where? Sorry, it's not clear to me.

TurboGit:
Let's just clear something. multi_priority has not been inverted. multi_priority is not used anymore for pipe ordering. pipe ordering is using iop_order and iop_order was not properly computed.

TurboGit:
So at the end the order in the pipe was wrong but this is not because multi_priority has been changed.

parafin:
I'm not saying it was inverted, I'm just saying it had inverse order compared to iop order all along.
And I'm not proposing to apply iop_order from the style, I'm saying you have to look at iop_order's of multi-instance iops in styles to decide their relative position to each other.

TurboGit:
For the specific case of styles.

The iop_order is not used as it cannot be applied safely t the target image. If the target image ahs reordered the pipe manually and say exposition was put at iop-order 10 and in the style you have a style created from another image with it was at iop-order of 100 you cannot use it.

When applying a style we need to use the ordering of the target image to properly place the different modules.

parafin:
Yes, that' correct. See above for what I meant.

TurboGit:
> I'm saying you have to look at iop_order's of multi-instance iops in styles to decide their relative position to each other.

I see, that would work but more complex than the solution proposed here.

parafin:
You don't have a solution yet. See case 2.

TurboGit:
For what its worth. I have tested this PR with the following scenario:
- migrate from 2.6.2 a style with multi instance
- import a style created with 2.6.2 having multi instance

And I got a proper order of the instances in current master.

parafin:
Yes, that's cases 1 and 3, try now case 2;)

TurboGit:
You mean this:
> 2. styles created in 3.0

parafin:
Also it's strange that imported style worked, https://github.com/darktable-org/darktable/blob/master/src/common/styles.c#L1268 looks wrong to me, isn't it used for conversion?

TurboGit:
@parafin: iop-order is not used in style at all. In fact it should not have been added into the database as we don't need this information.

parafin:
Ah, right, it's not used now, so it doesn't matter yet.

TurboGit:
@parafin : of case 2 is still broken... I'm on it :)

TurboGit:
@parafin : I was wrong ! multi_priority has changed order. this was because it is easier to update when creating a new instance. before we needed to slide all current instance to insert a multi_priority of 0. but this of course broke style creation. fix will be painful :(

parafin:
Strange. How did you determine that? When I did a simple test by creating a second instance and then swapping it with original instance, I haven't noticed multi_priority changing value in XMP. So I concluded you were right saying that multi_priority is now independent from iop order. It also makes sense, since there's no need for priority as such (iop order defines that), but you still want to distinguish the instances somehow.

parafin:
And I'll repeat my suggestion - forget about multi_priority in styles (it is already screwed) and look at iop order to determine which instance comes first. Coding that isn't too simple, yes, but at least it's obvious what needs to be done and that it will work (after fixing the bug I mentioned with imported 2.6 styles - there multi_priority is ignored and history stack order is used instead for some reason)

TurboGit:
@parafin: 
> Strange. How did you determine that?

I'm speaking about the migration from 2.6 to 3.0. It is inverted and I was wrong.

```
In 2.6: first instance prio 1 / second instance prio 0
In 3.0 : first instance priot 0 / second instance prio 1
```

So I have migrated the multi_priority in database. This way  styles created in 3.0 are using the proper order.

The missing point is to detect style before 3.0 (no iop_order nodes) and swap the multi_priority to have a proper iop_order. I think we are getting there.

In fact it is easier (at least for me to understand the code). All the code will be based on 3.0 and we do only proper migration from old versions:
- database is done with this PR
- remain old styles

parafin:
I think you are wrong. Try creating a second instance in 3.0 and moving it down before the original instance. Then look at multi_priority and multi_name - they haven't changed.

parafin:
Sorry, I was wrong - multi_priority is changed when you create a style, it doesn't change in image's XMP. That's confusing.

parafin:
Shit, no, ignore my last message, I don't understand how it works.

TurboGit:
In the new version I have also tested that styles imported from 2.6.x and styles created with 3.0 the iop-order is correct.

@parafin : this is complex (much more than it should) stuff and I'm on it since more than 8 months!

parafin:
I did the following 2 tests on 3.0.0rc1:
Test 1:
1. Discarded history on a image, activated framing module with default white frame.
2. Created a second instance of framing, which was placed above original instance, changed color to black.
3. Created a style from this 2 framing iops.
4. Applied this style to a duplicate of the image (after discarding its history), result looks correct.
Result:
1. multi_name="", multi_priority="0", iop_order="75"
2. multi_name="1", multi_priority="1", iop_order="75.5" (original image and style) or iop_order="75.01" (duplicate with applied style)

Test 2 (continues from test 1):
1. Swapped positions of duplicate instances in the original images.
2. Created a new style from update 2 framing iops.
3. Applied this new style to another duplicate of the image, result looks incorrect (the same as in test 1).
Result:
Original image and style:
1. multi_name="1", multi_priority="1", iop_order="74.5"
2. multi_name="", multi_priority="0", iop_order="75"
Duplicate with applied style:
1. multi_name="", multi_priority="0", iop_order="75"
2. multi_name="1", multi_priority="1", iop_order="75.01"

I suggest you repeat these 2 tests with this PR applied on your end.

parafin:
So after all I think you are wrong (but was right before), and meaning of multi_priority didn't change to inverted order, but instead lost any relation to module ordering. So the new fixes you are doing now don't affect the 3.0 styles with multi-instance being broken, instead they are just another way to fix old 2.6 styles.

parafin:
Another point - multiple instances of one iop can be moved to the different positions in the pipe, so there will be other iops between them. Trying to keep track of which instance comes first using multi_priority is harder because of that, and doesn't bring any value (since there is iop_order) aside from this mess with styles, which is already unfixable with just multi_priority - you have to look at iop_order in 3.0 styles anyway. So again - you were right, that multi_priority has no relation to iop_order in 3.0.

TurboGit:
@parafin: yes multi_priority has no relation with iop_order AFTER migration from 2.6.x.

We only need to take care of proper migration of multi-instance from 2.6 to 3.0 to ensure then a proper computation of iop-order at migration point.

My PR here now take into account the database, the xmp and style files.

Note that if I'm not mistaken multi-priority is still used to order the module in the GUI. Maybe at some point we will be able to remove multi-priority altogether.

parafin:
No, it's not used for GUI either.

TurboGit:
So we will probably be able to remove it sooner.

jenshannoschwalm:
About style at export - imageio.c L. 685 - does this need attention?

TurboGit:
@jenshannoschwalm : no :) be sure to look at the new patch it takes another approach and in fact I have removed all the (90 - multi_priority). This was working or for multi_priority from 2.6 but not from 3.0 where the order is reverted. A change that made an earthquake but necessary to  make adding/handling multi-instance easier in 3.0.

TurboGit:
**### ALL IF YOU TEST THIS PR PLEASE BACKUP YOUR DB.**

parafin:
@TurboGit, have you tried to reproduce the two tests I described? Since you've touched only 2.6->3.0 conversion, I'm pretty sure 3.0 multi-instance styles are still broken. Not all of them, but when user used move up/down functionality.

TurboGit:
@parafin : no not yet, I'll do that... not sure if I'll have time tonight. I'll report anyway.

junkyardsparkle:
Does multi-priority actually need a complicated fix for styles? I seem to remember styles being described as a "bag of presets" in earlier discussion about why they can't convey iop ordering. They're already somewhat broken for use in 3.0 in the sense that if you create one from a non-default ordering pipe, it won't recreate the same effect when applied to another image. But somehow multi-instance ordering needs to be preserved? Sorry if I'm being stupid, reading through all this made my head spin.

TurboGit:
> in earlier discussion about why they can't convey iop ordering.

No they can't. But at least we must respect the order of module's multi instances .

junkyardsparkle:
> No they can't. But at least we must respect the order of module's multi instances .

I'm not sure why, since styles created in 3.0 will lose this anyway... right?

junkyardsparkle:
is multi-insstance ordering in styles somehow going to be exceptional, even after being converted to iop ordering?

parafin:
The difference here is that iop re-ordering is new feature, so the fact that styles don't fully support it yet can be considered as a missing feature, not a big. But multi-instance and its order was already present in 2.6, so removing or crippling support for it is a regression and therefore a bug.

junkyardsparkle:
I understand it philosophically, I just don't understand how it can work... if in 3.0 there is only iop priority and no multi priority. How is the distinction then maintained?

parafin:
Applying style with multi_priority is well-defined and works in 2.6. Applying style with iop_order is not defined, that's why it's not implemented. But iop_order can be converted to old-style multi_priority quite easily - multi_priority was an order of the instances of the same iop, and this order can be determined in 3.0 from iop_order of the instances.

junkyardsparkle:
So will a multi-instance style *created in 3.0* work the same way as an ingested legacy style? Or will it be missing the multi priority information? That's what I'm not understanding.

parafin:
It should work the same, that's what is still unfixed in this PR. It may use iop_order instead of multi_priority field in style itself to achieve that, but that's a technical detail.

junkyardsparkle:
Ok, thanks for bearing with me. :-) So, a style created from a pipe with multiple instances, possibly not grouped together, when applied will group those instances together, at destination default location, but the ordering of those instances relative to each other will be maintained. Correct?

parafin:
yes

TurboGit:
@parafin: your scenario was broken on this PR indeed :) Next version is pushed and is based on iop-order as indeed multi-priority is not used after migration.

TurboGit:
@parafin: I have tested again the new version of this PR with your scenario and all is fine on my side. I'm confident that we are seeing the end of this issue :)

TurboGit:
Given my testing and review I think this is ready now. I'll merge that tomorrow, if your testing shows some issues just report before I merge. Thanks.

jenshannoschwalm:
I first tested on a small number of images, everything was fine; after that i gave my complete system a try and again **no problem detected** and for me everything looks fine. I tried @parafin suggested tests and - if understood correctly - again no problem!

Pascal, i think you did a great and hard job here!

parafin:
I didn't test the code, but only looked it over. I think last commit should be reverted - it is based on the assumption that all instances have matching integral part of the iop_order, which can be false when iops were re-ordered. Simple test is to move one of the instances to some other location in the pipe - iop_order will be changed drastically for it then and relative order of the instances will get screwed up when calculated with current version of this PR.

There's also another general issue with styles and iop re-order, which isn't directly related to this issue, so feel free to ignore it for now. If we are applying a style to an image with re-ordered iops, iop_order of these iops can be basically anything (you can always come up with a sequence of re-order operations so that iop_order of the module is changed to the target number). So using default iop_order of the module when applying the style may not be the best strategy. Maybe it's better to get the iop_order of base instance (multi_priority == 0) of given module in the specific image we are applying a style to.

TurboGit:
> I didn't test the code, but only looked it over. I think last commit should be reverted - it is based on the assumption that all instances have matching integral part of the iop_order, which can be false when iops were re-ordered.

Oh! Indeed, I missed that we can reorder multi-instance far away from each others! Will revert.

junkyardsparkle:
> Maybe it's better to get the iop_order of base instance (multi_priority == 0) of given module in the specific image we are applying a style to.

Styles are becoming a strange beast, half "smart", half "dumb". ;-)
That's an interesting point, though... one problem I can think of with that approach is that if a user did lots of reordering on the destination history, and didn't have the module(s) from the style active, the current location for those might be a bad one, because the user wouldn't know/care up to that point. Still, this approach might not be any worse than default in that case, I guess (probably depends on the specific module)?

TurboGit:
Now merged! Thanks for all the feedback and comments.

junkyardsparkle:
(Best branch name typo ever, by the way.)

Issue #3437, opened 2019-11-14
iop order diferent when migrate from 2.6 - again


vsz1lu:
Similar problem what is closed in #3401 

When You open an image in dt 3.0.0rc1 what have sidecar from dt 2.6 the iop order not match the default.

Why problem?
a, The output is not same in most of case in real work.
b, Working with pictures is different, when You reopen an old work and than pick up a new image.

------------------------------------------------------------
What I do?

1, delete config dir of darktable
2, start darktable 2.6.2+13
3, change language to english - restart
4, open an virgin raw image, sidecar xmp not present
5, close darktable
6, delete config dir of darktable
7, rename the newborn xmp
8, start darktable 3.0.0rc1
9, change language to english - restart
10, open the raw
11, set 'subset: all modules'
12, turn on sharpen, what is not default in 3.0.0
13, make duplicate by pressing CTRL+D
14, load the sidecar from dt 2.6.2+13 for the clone in lighttable
15, compare the iop order side by side

the result:
![01 basic](https://user-images.githubusercontent.com/57588193/68902032-88ec3a80-0737-11ea-994f-f879f88a6a08.png)
![02 tone](https://user-images.githubusercontent.com/57588193/68902040-90abdf00-0737-11ea-9400-7fb63df63e83.png)
![03 color](https://user-images.githubusercontent.com/57588193/68902053-973a5680-0737-11ea-8c60-5633d0911759.png)
![04 correction](https://user-images.githubusercontent.com/57588193/68902058-9b667400-0737-11ea-96f5-86642abe0292.png)
![05 effects](https://user-images.githubusercontent.com/57588193/68902073-a4574580-0737-11ea-9863-1f78b15439f9.png)

When load a sidecar from dt 2.6.2+13 where all iops on (xmp attached):

***** On-the-fly history V[1]->V[2], imageid: 1 ****************
   0                 flip multi  0 :: iop 20,00100000000 -> 20,00000000000
   1              sharpen multi  0 :: iop 53,00100000000 -> 53,00000000000
   2            basecurve multi  0 :: iop 23,00100000000 -> 23,00000000000
   3               colisa multi  0 :: iop 47,00100000000 -> 47,00000000000
   4               shadhi multi  0 :: iop 40,00100000000 -> 40,00000000000
   5     colorreconstruct multi  0 :: iop 28,00100000000 -> 28,00000000000
   6             clipping multi  0 :: iop 21,00100000000 -> 21,00000000000
[...]

PR #3487, opened 2019-11-19
Fix colorzones strength upgrade from V3 to V4

leclercj:
Need for fix from https://github.com/darktable-org/darktable/issues/3437


TurboGit:
Yes, correct ! Thanks.

PR #3496, opened 2019-11-20
blend gui: addition of alternative display modes for parametric mask sliders

upegelow:
* logarithmic view for L, R, G, B, C and g sliders (=> better control for dark pixels)
* zoom view for a and b sliders (=> better control for low saturated pixels)

views are toggled by typing "a" when mouse is over the slider. note that view
is not part of the parameter set, so it is not saved when you leave darkroom.

this is an enhancement for post 3.0.


TurboGit:
Will look at this. I have changed from 3.2 to 3.0.1, if ready why not in the next minor release.

upegelow:
I'm still considering if value changes (slider positions, colorpicker, gradient, reset values) should be moved completely into the gradientslider regime so that from the outside only the corresponding callback function needs to be registered and all is set.
Currently re-calculations are done outside of the widget logic.

rawfiner:
Not tested it yet, but I think this is a great idea, thanks @upegelow ! :-)

TurboGit:
Just tested as I was getting really curious! A nice addition, will be quite handy in some situations I had encountered!

TurboGit:
Merged, thanks!

PR #3506, opened 2019-11-21
[dt30] Second part of translation into German[

mepi0011:
Translation for next darktable release


mepi0011:
Hi @spaceChRiS,

Thank you for pointing that out.
The translation into "Anhängsel" does not work at all! :-)

Translation from tag into Tag is possible. However, other programs like digikam and KDE also translate tag into "Stichwort"! Therefore I do not find the translation absurd and would leave it that way.

spaceChRiS:
Yes, I know it doesn't work :) This was not a serious suggestion. But at least it would better describe what tags actually are.

Tags can do so much more than what is covered by "Stichwort". "Tree", "house", ... are the kind of tags covered by "Stichwort", but "darktable|exported" is a workflow tag and not a "Stichwort".

However, I do not have a strong opinion on this, I just wanted to point out that this translation may be misleading for new users. But I have as well no better solution than keeping the anglicism "tag" or ignoring my concern.

mepi0011:
Hello @houz and @cryptomilk,
what is your opinion about the topic of translation from "Tag" to "Stichwort"?

cryptomilk:
I think "Schlüsselwort" would fit better.

mepi0011:
Now I have translated the English word tag again with tag instead of keyword.
The translation can now be used in darktable 3!

PR #3509, opened 2019-11-21
module list : Ensure we update iop list in all cases

### LINKED ISSUE #3330: old filmic iop is still displayed
When updating from 2.6 installation to 3.0 if filmic was enabled in more modules, then it will be still displayed in the iop list for new images, so it's possible to use it (and get that warning sign in the history for newly developed images). And it's not listed in more modules, so it's not possible to remove it either.

AlicVB:
This fix #3330
in the case of no group selected (show all iops), nothing was done to update iop list, contrary to all other cases...

### DIFF
--- a/src/libs/modulegroups.c
+++ b/src/libs/modulegroups.c
@@ -249,6 +249,7 @@ void gui_init(dt_lib_module_t *self)
   gtk_box_pack_start(GTK_BOX(self->widget), d->hbox_search_box, TRUE, TRUE, 0);
 
   gtk_toggle_button_set_active(GTK_TOGGLE_BUTTON(d->buttons[d->current]), TRUE);
+  if(d->current == DT_MODULEGROUP_NONE) _lib_modulegroups_update_iop_visibility(self);
   gtk_widget_show_all(self->widget);
   gtk_widget_show_all(d->hbox_buttons);
   gtk_widget_set_no_show_all(d->hbox_buttons, TRUE);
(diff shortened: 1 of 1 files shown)


src/libs/modulegroups.c
```diff
@@ -249,6 +249,8 @@ void gui_init(dt_lib_module_t *self)
   gtk_box_pack_start(GTK_BOX(self->widget), d->hbox_search_box, TRUE, TRUE, 0);
 
   gtk_toggle_button_set_active(GTK_TOGGLE_BUTTON(d->buttons[d->current]), TRUE);
+  printf("%d\n", d->current);
```
TurboGit: debug code to be removed.

AlicVB: oops... fixed

aurelienpierre:
I didn't tested yet, but I don't see how that solves the bug mentioned above, which affects the widget in `modulelist.c`.

AlicVB:
Oh ok, that was too easy ;)
What I've understand (and reproduce) in the bug is that deprecated iop are still listed in the list of iop if you open an image with all iop shown (no group selected) but I've certainly misread the thread and found another bug !
You say the problem was in modulelist.c which is the "more module" panel, right ? Let's try to reread #3330 with that in mind...

aurelienpierre:
> You say the problem was in modulelist.c which is the "more module" panel, right ? Let's try to reread #3330 with that in mind...

Yes, exactly.

AlicVB:
hmm... after rereading the bug, it seems that I've fixed exactly what @parafin report in his first comment :
> When updating from 2.6 installation to 3.0 if filmic was enabled in more modules, then it will be still displayed in the iop list for new images, so it's possible to use it

Do you see deprecated iop in the "more module" panel too ?

Now, reading you replies, I wonder if your goal is to let some kind of "semi-hidden" way to use those modules... if that's the case, I'm not sure that showing them in "more module" is a good idea...

Anyway, I think that this fix is needed as it ensure that the list of iops is synchronized in any cases

aurelienpierre:
No, the issue is deprecated modules don't appear in the "more modules" panel, ever when enabled in current image. So if users hide filmic in 2.6 but have it enabled one some pictures, they will never be able to show it again in UI.

parafin:
I think all enabled modules are shown in active modules list, even if disabled in more modules. They won’t be listed in groups though. If I’m right, then I don’t think that it’s actually an issue.

aurelienpierre:
> I think all enabled modules are shown in active modules list, even if disabled in more modules. They won’t be listed in groups though. If I’m right, then I don’t think that it’s actually an issue.

Yes.

TurboGit:
Thanks AlicVB ! Nice and tricky fix.

TurboGit:
I'll merge that and the other part of the fix (if needed) will be part of another PR.

PR #3511, opened 2019-11-22
Fix obvious defects and warnings in zh translation

boltomli:
Some are legacy issues and some are introduced or fuzzy.


TurboGit:
Thanks !

PR #3523, opened 2019-11-22
Button order fixes

### LINKED ISSUE #3517: UI → Wrong button order
Since dt uses reversed button order in UI (approval buttons are on the right side). This can be considered a bug. Similar problem was fixed before in export settings.

Steps to reproduce:

1. Open lighttable mode → tagging module.
2. Create any tag if list of available tags is missing.
3. Right click on created tag:
        a) "delete tag" action opens confirmation window with reversed
button order. Same for "delete branch".
        b) "create tag..." opens dialog window with reversed button
order. Same for "edit tag...".

sams96:
Fixes #3517, awaiting response with regards to some other button orders.


sams96:
I think that's everything

TurboGit:
This makes the UI more consistent! Thanks a lot.

PR #3538, opened 2019-11-23
history on-the-fly msg : ensure ending lines

AlicVB:
very small glitch only visible in the command line : we must ensure that we always have '\n' at the end of the message, otherwise next message are shown on the same line...


TurboGit:
Thanks, there is others I suppose but this is going in the right direction.

PR #3539, opened 2019-11-23
Plenty of grammar+syntax+capitalizing on French language (fr.po)

GLLM:
Grammar & syntax corrections
Light rewording
Rationalize syntax : e.g.e  right-click & right click
Capilized : IL, EV, JP*G, ISO, DCP, ICC, EXIF, XMP
Did not capitalize "raw" because its not an abbreviation but a word.

### DIFF
--- a/po/fr.po
+++ b/po/fr.po
@@ -222,7 +222,7 @@ msgstr "active la superposition étendue des miniatures"
 msgid "if set to true, thumb overlay shows filename and some exif data."
 msgstr ""
 "si activé, la superposition des miniatures affiche le nom du fichier et "
-"quelques données EXIF."
+"quelques données Exif."
 
 #: ../build/src/preferences_gen.h:2402
 msgid "use single-click in the collect panel"
@@ -2391,7 +2391,7 @@ msgstr "copie locale d'images"
 
 #: ../src/control/jobs/control_jobs.c:1751 ../src/libs/image.c:276
 msgid "refresh exif"
-msgstr "rafraîchir EXIF"
+msgstr "rafraîchir Exif"
 
 #: ../src/control/jobs/control_jobs.c:1814
 #, c-format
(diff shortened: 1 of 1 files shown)


timur-davletshin:
EXIF should be spelled as Exif according to wikipedia.

po/fr.po
```diff
@@ -3657,17 +3657,17 @@ msgstr "fichier de stockage"
 msgid "general"
 msgstr "général"
 
-#. ignoring of jpegs. hack while we don't handle raw+jpeg in the same directories.
+#. ignoring of JPEGs. hack while we don't handle raw+JPEG in the same directories.
```
TurboGit: Maybe RAW+JPEG then, no?

po/fr.po
```diff
@@ -8546,8 +8546,8 @@ msgstr ""
 "ne jamais toucher sauf si vous savez ce que vous faites."
 
 #: ../src/iop/filmicrgb.c:192
-msgid "filmic rgb"
-msgstr "filmique rvb"
+msgid "filmic RGB"
```
TurboGit: You can't change this string as it is the one in the code. This is the "key" that dt will use at runtime to find the translation. Changing that breaks the translation.

po/fr.po
```diff
@@ -6386,7 +6386,7 @@ msgstr "position verticale"
 
 #: ../src/iop/borders.c:1043
 msgid ""
-"select the vertical position ratio relative to left or right click and type "
+"select the vertical position ratio relative to left or right-click and type "
```
TurboGit: You cannot change the main English string here. This should be done in the code and cannot be done at this stage as we are in string freeze.

po/fr.po
```diff
@@ -9465,7 +9465,7 @@ msgstr "outil point: éditer, ajouter et supprimer des points"
 
 #: ../src/iop/liquify.c:3607
 msgid ""
-"ctrl-click: add node - right click: remove path\n"
+"ctrl-click: add node - right-click: remove path\n"
```
TurboGit: no, see above.

po/fr.po
```diff
@@ -9474,7 +9474,7 @@ msgstr ""
 #: ../src/iop/liquify.c:3609
 msgid ""
 "click and drag to move - click: show/hide feathering controls\n"
-"ctrl-click: autosmooth, cusp, smooth, symmetrical - right click to remove"
+"ctrl-click: autosmooth, cusp, smooth, symmetrical - right-click to remove"
```
TurboGit: no, see above.

GLLM:
> You need to revert all changes done in main English strings.

Oh. I'm very sorry about this, I'll remove my PR.
Sadly, I checked the de.po file and noticed that the english strings were capitalized differently, so I thought it was fine...

For the raw+JPG ... in my first commit comment, i thought about it but did not do it as JPEG are initials whereas RAW is not, just a word (afaik).

I'm sorry @TurboGit if my PR triggered some unnecessary work on your side with the english strings & the few miss I later saw you correct.

I will remove my PR, so that I don't fill the commits list with more crap whereas there are plenty of useful code addition pending.

Merci
GLLM

PR #3544, opened 2019-11-23
UI -> rgb_curve 

### LINKED ISSUE #3528: UI → rgb curve
Steps to reproduce:
1. Import two files.
2. Apply rgb curve module for both of them with independent channels option enabled.
3. Open "G" or "B" channel inside rgb curve.
4. Press space to move to the next file (do not close rgb curve module).

Problem:
rgb curve module does not refresh  UI ("R", "G", "B" tabs) for the next file. "R" curve is loaded correctly but UI remain on "G" or "B". See that on attached screenshot.

Untitled

wpferguson:
Added check to change_image() to see if a channel was already set.  If a channel was set, we use that instead of defaulting to the red channel.

Fixes #3528.


TurboGit:
Looks ok to me! Another option would have been to always initialize the tab to R but I have no strong opinion. Thanks!

Issue #2843, opened 2019-08-06
Drag and drop does not work from light table, only from darkroom bottom panel filmstrip


lsolbach:
**Describe the bug**
As I am often work with bracketed shots for HDR processing in photomatix, I need to drag a bunch of RAWs on the photomatix window. While this works from the filmstrip in the bottom of the darkroom, it does not work from the file manager (or zoomable lighttable) in the light table.

This is an inconvenience, because in the light table you have a much better overview of the images.

**Expected behavior**
Enable Drag and drop from the light table, at least in file manager mode.

**Platform (please complete the following information):**
 - OS: Ubuntu 19.04
 - Version 2.6.0

johnny-bit:
the filmstrip & filemanager have been overhauled with completely different feature set. While I won't hide that I don't understand your workflow and simply never use d&d functionality anywhere, I must close this issue as it doesn't match at all current dt featureset. Please check recent release and if you still have a problem, file a bug against current release :)

lsolbach:
Just for the reference, it works in 3.4.

PR #3563, opened 2019-11-25
Accelerators for filmstrip don't conflict with lighttable

### LINKED ISSUE #3562: Accelerators for lighttable and filmstrip conflict
The Preferences/Shortcuts editor will not allow the same keys to be defined for an action in lighttable and filmstrip (which is used in darkroom). This is inconvenient, because in the standard keyboard layout (and logically) rating and color assignments use the same keys in both (1..5, F1..F5). If one accidentally reassigns one of these keys, _both_ original assignements will be removed, but it will only be possible to set one of them back.

dterrahe:
Change #2604 made sure that any accelerators defined for filmstrip are ignored in the context of lighttable.

This change lets preferences/shortcuts allow the same keys in both.

Closes #3562

### DIFF
--- a/src/gui/accelerators.c
+++ b/src/gui/accelerators.c
@@ -183,7 +183,7 @@ void dt_accel_register_lib(dt_lib_module_t *self, const gchar *path, guint accel
   const gchar **views = self->views(self);
   while (views[i])
   {
-    if (strcmp(views[i], "lighttable") == 0) accel->views |= DT_VIEW_LIGHTTABLE;
+    if (strcmp(views[i], "lighttable") == 0 && strcmp(accel->module, "filmstrip") != 0) accel->views |= DT_VIEW_LIGHTTABLE;
     else if (strcmp(views[i], "darkroom") == 0) accel->views |= DT_VIEW_DARKROOM;
     else if (strcmp(views[i], "print") == 0) accel->views |= DT_VIEW_PRINT;
     else if (strcmp(views[i], "slideshow") == 0) accel->views |= DT_VIEW_SLIDESHOW;
--- a/src/gui/accelerators.c
+++ b/src/gui/accelerators.c
@@ -183,7 +183,7 @@ void dt_accel_register_lib(dt_lib_module_t *self, const gchar *path, guint accel
   const gchar **views = self->views(self);
   while (views[i])
   {
-    if (strcmp(views[i], "lighttable") == 0 && strcmp(accel->module, "filmstrip") != 0) accel->views |= DT_VIEW_LIGHTTABLE;
+    if (strcmp(views[i], "lighttable") == 0) accel->views |= DT_VIEW_LIGHTTABLE;
     else if (strcmp(views[i], "darkroom") == 0) accel->views |= DT_VIEW_DARKROOM;
     else if (strcmp(views[i], "print") == 0) accel->views |= DT_VIEW_PRINT;
     else if (strcmp(views[i], "slideshow") == 0) accel->views |= DT_VIEW_SLIDESHOW;
--- a/src/libs/tools/filmstrip.c
+++ b/src/libs/tools/filmstrip.c
@@ -154,8 +154,12 @@ const char *name(dt_lib_module_t *self)
 
 const char **views(dt_lib_module_t *self)
 {
-  static const char *v[] = {"lighttable", "darkroom", "tethering", "map", "print", NULL};
-  return v;
+  static const char *v1[] = {"lighttable", "darkroom", "tethering", "map", "print", NULL};
+  static const char *v2[] = {"darkroom", "tethering", "map", "print", NULL};
+  if(self->data)
+    return v1;
+  else
+    return v2; // We are registering accelerators; don't conflict with overlapping lighttable versions
 }
 
 uint32_t container(dt_lib_module_t *self)
--- a/src/libs/tools/filmstrip.c
+++ b/src/libs/tools/filmstrip.c
@@ -212,6 +212,10 @@ void init_key_accels(dt_lib_module_t *self)
 
 void connect_key_accels(dt_lib_module_t *self)
 {
+  // on lighttable, does nothing and report that it has not been handled
+  const dt_view_t *cv = dt_view_manager_get_current_view(darktable.view_manager);
+  if (cv->view((dt_view_t *)cv) == DT_VIEW_LIGHTTABLE) return;
+
   // Rating accels
   dt_accel_connect_lib(self, "rate 0", g_cclosure_new(G_CALLBACK(_lib_filmstrip_ratings_key_accel_callback),
                                                       GINT_TO_POINTER(DT_VIEW_DESERT), NULL));
@@ -409,13 +413,6 @@ static gboolean _lib_filmstrip_mouse_leave_callback(GtkWidget *w, GdkEventCrossi
   gtk_widget_queue_draw(strip->filmstrip);
 
   return TRUE;
-}
-
-static inline gboolean _is_on_lighttable()
-{
-  // on lighttable, does nothing and report that it has not been handled
-  const dt_view_t *cv = dt_view_manager_get_current_view(darktable.view_manager);
-  return cv->view((dt_view_t *)cv) == DT_VIEW_LIGHTTABLE;
 }
 
 static gboolean _lib_filmstrip_size_handle_cursor_callback(GtkWidget *w, GdkEventCrossing *e,
@@ -945,8 +942,6 @@ static gboolean _lib_filmstrip_copy_history_key_accel_callback(GtkAccelGroup *ac
                                                                GObject *aceeleratable, guint keyval,
                                                                GdkModifierType modifier, gpointer data)
 {
-  if(_is_on_lighttable()) return FALSE;
-
   dt_lib_filmstrip_t *strip = (dt_lib_filmstrip_t *)data;
   const int32_t mouse_over_id = dt_control_get_mouse_over_id();
   if(mouse_over_id <= 0) return FALSE;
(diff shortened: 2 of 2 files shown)


TurboGit:
That's clever! Need good testing I suppose.

dterrahe:
Well, i tried a nicer version first, where i set a flag in init_key_accels to let the call to views know that we were setting up accelerators. But then i found that self->data had not been initialised yet at that point, so i didn't know where to store (or initialise) the flag.

I've tested this quick and dirty hack quite a bit and haven't found any side effects yet; gui_unit gets called immediately after accels so from then on self->data is allocated and everything is as before.

Of course this isn't as robust to changes to core assumptions. Any suggestions you'd have I would gladly implement. An alternative would be to put a special exception into the accelerator code that ignores "lighttable" for the "filmstrip" module. This would be a one line change, if it works which i haven't tried, but i somehow felt this hack was more "localised". Let me try the other approach and you can see which you like better.

dterrahe:
What do you think; even more minimal change, tested working without apparent side effects, arguably more self-documenting, but module specific code in general accelerator routine...

AlicVB:
filmstrip code already take care of not applying accels routines if we are inside lighttable (to avoid double effect)
So now I think the less intrusive way to handle the problem of conflict on shortcuts reassignment is to add an exception directly in the pref code. near https://github.com/darktable-org/darktable/blob/3d6a29f8bf800dd6087eae2e834a0417159105d6/src/gui/preferences.c#L848 (but iirc, there's at least one other place in the file where we do that check...)

dterrahe:
Also, in a separate commit, only enabling the accelerators for filmstrip when we are not in lighttable, so that each accel_callback does not have to separately check and disable itself.

This whole approach is now all or nothing for all filmstrip accelerators in lighttable. If for same reason it were desirable to enable some, but not others, then my pseudocode would be:
```
static gboolean registering_accels = FALSE;

const char **views(dt_lib_module_t *self)
{
  static const char *v1[] = {"lighttable", "darkroom", "tethering", "map", "print", NULL};
  static const char *v2[] = {"darkroom", "tethering", "map", "print", NULL};
  if(registering_accels)
    return v2;
  else
    return v1; // We are registering accelerators; don't conflict with overlapping lighttable versions
}

void init_key_accels(dt_lib_module_t *self)
{
  // register accels that should work everywhere
  registering_accels = TRUE;
  // register accels that should not work in lighttable
  registering_accels = FALSE;
}

void connect_key_accels(dt_lib_module_t *self)
{
  // connect accels that should always work

  const dt_view_t *cv = dt_view_manager_get_current_view(darktable.view_manager);
  if (cv->view((dt_view_t *)cv) == DT_VIEW_LIGHTTABLE) return;

  // connect accels that should not work in lighttable
}
```

dterrahe:
> So now I think the less intrusive way to handle the problem of conflict on shortcuts reassignment is to add an exception directly in the pref code. 

Wouldn't it be clearer to correctly reflect in the definitions of the accelerator that they are not active in lighttable, by not including the DT_VIEW_LIGHTTABLE flag? Then any code in prefs automatically takes that into account. 

> (but iirc, there's at least one other place in the file where we do that check...)
So this would not be an issue and additional problems would not accidentally be introduced in the future.

I kind of like the solution I propose in the pseudo code above; it looks a bit messier in the filmstrip module, but to be honest that's the one that creates the issue ;-) , but it allows for more flexibility in case some accelerators are not overlapping and therefore should work also in lighttable. Plus no changes are required in other files (accelerator or prefs) so the whole complexity is localised and can be made more apparent in the filmstrip module itself...

AlicVB:
sorry, I missed your last commit and comments.
I prefer that approach (the last one with pseudo-code)
There's still some problems, thought : 
I'm not sure how preference will be able to determine which shortcuts overlap with lighttable or not... and there's the shortcut preview window problem too...
I fear we'll end up with problem sooner or later with that sort of "exceptions" patch...
I'll try some ideas I have and report them if they work

dterrahe:
Actually, I just checked that it fixed the shortcut preview window (h) too!

I'll turn my pseudocode into something working and report back after testing.

AlicVB:
ok, so what I've tried (and it seems to work quite well) is to move actual filmstrip shortcuts to darkroom.c and transfer them to filmstrip with a proxy.
With that solution, shortcuts belongs to the right view, we can add filmstrip specific shortcuts, and as it doesn't change how shortcuts are handled by the rest of dt, there's less chance it can break something...
here's the code : https://github.com/AlicVB/darktable/pull/new/filmstrip_shortcuts
note that it only handle ratings atm and the code needs cleanups.

dterrahe:
I've just added a new commit which limits all the changes to filmstrip, allows both accelators that _are_ active in lighttable and accelerators that are disabled there (because of overlap) and that makes the preferences dialog correctly deal with conflicts and the help window show all the correct shortcuts. And it uses the same accelerators, so users will not have to redefine their existing keyboardrcs.

Moving the accelerators to darktable means that we lose them in the other views (maps, tether and print) where they currently work.

I was actually thinking in the opposite direction of your change; if the shortcuts in lighttable are removed, we could just use filmstrip everywhere. This would have the advantage that the same shortcut doesn't have to be defined in two places (and customized by the user if they so wish). It would have the drawback of not allowing different keys in darkroom and lighttable (but is that something to be encouraged?)

AlicVB:
> Moving the accelerators to darktable means that we lose them in the other views (maps, tether and print) where they currently work.

good point

> I was actually thinking in the opposite direction of your change; if the shortcuts in lighttable are removed, we could just use filmstrip everywhere. This would have the advantage that the same shortcut doesn't have to be defined in two places (and customized by the user if they so wish). It would have the drawback of not allowing different keys in darkroom and lighttable (but is that something to be encouraged?)

another drawback (same as my proposal) : in prefs windows shortcuts will be difficult to find...

about your last proposal : We may get some sort of race condition if some part of dt call views() during accels initialization... (I agree the probability is low, but not null...)

Another idea (more general) : adding a `dt_accel_register_lib_for_views` routine in accelerators.c to let libs init accels for a specif views. I think it'll avoid all the drawback we have with previous proposal, and even allow other "cross-views" libs to do the same (image infos, ...)
see here : https://github.com/AlicVB/darktable/tree/filmstrip_shortcuts

dterrahe:
> another drawback (same as my proposal) : in prefs windows shortcuts will be difficult to find...
What I remember from my initiation into dt, was that it wasn't that hard to find the shortcut for "rating" in the preferences (you walk through the whole thing and eventually stumble upon it and now with the help screen it is even easier), but that it was confusing that there were two places and that it was unclear if there is any difference. Just having this under filmstrip will be fine, I think.

> about your last proposal : We may get some sort of race condition if some part of dt call views() during accels initialization... (I agree the probability is low, but not null...)
Remember that this is set up during dt start only. Not even gui_init is called yet. Is there any multi-threading going on at this point? Is there any guarantee that other code is reentrancy safe?

> Another idea (more general) : adding a `dt_accel_register_lib_for_views` routine in accelerators.c to let libs init accels for a specif views. I think it'll avoid all the drawback we have with previous proposal, and even allow other "cross-views" libs to do the same (image infos, ...)
> see here : https://github.com/AlicVB/darktable/tree/filmstrip_shortcuts
Looks good. A bit more involved than my initial hack, but certainly better design and more robust!

Would you still agree with my other patch which, instead of ignoring accelerators when we are in lighttable, simply doesn't connect them in the first place? (82275c7)

AlicVB:
> Would you still agree with my other patch which, instead of ignoring accelerators when we are in lighttable, simply doesn't connect them in the first place?

good point ! In order to stay more global (so it can be easily reused for other libs), I've just added the check for views directly inside accelerator.c : before connecting lib accel, we ensure that the current view is inside registered accels views.
I've updated my branch mentioned earlier is you want to have a look and test...

I've also updated filmstrip.c to remove the check for lighttable (not entirely, as there are still needed for dnd)

dterrahe:
> In order to stay more global (so it can be easily reused for other libs), I've just added the check for views directly inside accelerator.c : before connecting lib accel, we ensure that the current view is inside registered accels views.
> I've updated my branch mentioned earlier is you want to have a look and test...

I had a look at the code and it looks like that will work (I'm in the middle of something, so haven't tested).

It does feel a bit like overkill, to me, to add a possibly expensive test for each and every accelerator to connect, when we can avoid trying to connect them at all with one single test in filmstrip that is stable and maintainable/self-documenting... Do you expect many/any exceptions like this beside filmstrip, where a module is visible but its accelerators need to be disabled?

Also, in accelerator.c we can only _prevent_ accels from being connected if they are not flagged for that view; we can not actively enforce all accelerators for a view to be connected if the right calls aren't made from the module. So accelerator.c is not in a position to take full ownership of enforcing the relationship. I would see it more as documenting it...

But your call of course!

> as there are still needed for dnd

Does drag and drop need to be disabled in lighttable? Without that check dragging into a filemanager from the filmstrip seems to work fine. With the check I still get the mouse pointer changing shape, but the drop doesn't work.

dterrahe:
Enabling drag and drop from the filmstrip in lighttable (while in culling mode) could maybe partially eleviate #2843

AlicVB:
> Also, in accelerator.c we can only _prevent_ accels from being connected if they are not flagged for that view; we can not actively enforce all accelerators for a view to be connected if the right calls aren't made from the module. So accelerator.c is not in a position to take full ownership of enforcing the relationship. I would see it more as documenting it...

not sure to understand (sorry for my English...)
About the cost of all the check, I agree that this can be overkill, as it'll only concern few libs. We can also let to libs the responsibility of connecting accels or not (as simple check like `_is_on_lighttable` should be sufficient)
About dnd, I've absolutely no idea of what is involved behind that, and why it has been deactivated in lighttable (there's certainly a reason for that) @TurboGit ?
That's said, the proper way to resolve #2843 is to implement the feature directly in lighttable.

dterrahe:
Better implementation by AlicVB in #3594 
Thanks!

Issue #3556, opened 2019-11-25
DNG → fix demosaic UI


timur-davletshin:
Darktable shows demosaic module turned on when opening DNGs with regular pixel order (already demosaiced). It looks like internally it processes them correctly since no visible changes happen when passthrough (or other options) are changed.

Expected behavior: disable demosaic module for such files just like for JPGs or regular TIFF files.

Picture author notice: Serge Ramelli, publicly available on his site.

[Episode 178 San francisco.zip](https://github.com/darktable-org/darktable/files/3885152/Episode.178.San.francisco.zip)

vsz1lu:
Confirmed. Tested with grayscale raw dng from transparent film scan by VueScan.

![Képernyőkép erről: 2019-11-25 17-01-19](https://user-images.githubusercontent.com/57588193/69557190-76c79300-0fa6-11ea-8bfe-04645e76971d.png)


Demosaic modul turned on but automatically set to passtrough mono. So I think it's not problem (for me) by technically - however hiding this module maybe more elegant solution in similar case.

timur-davletshin:
It doesn't set to passthrough for color images.

vsz1lu:
You're right!

Here is 2 1600dpi negative dng in grayscale and AdobeRGB color space (20MB)  - free use :
[https://drive.google.com/open?id=18gdWB-zoJZdEp5p-tuNjPeoxk7qVDH2u](https://drive.google.com/open?id=18gdWB-zoJZdEp5p-tuNjPeoxk7qVDH2u)

The demosaic module goes to PPG by automatically after open.

timur-davletshin:
@vsz1lu exactly! That was my point. Passthrough is fine, but PPG is wrong.

TurboGit:
Should be fixed.

PR #3565, opened 2019-11-25
fix #3555 and #3556. DNG files should keep the color matrix if supplied, non-RAW dng disable some iop

### LINKED ISSUE #3555: DNG file has strange default wb values, wrong colors and error message
**Describe the bug**
DNG files from a CHDK point-and-shoot Canon show strange default values in white balance module and very pale colors.
Besides, everytime the file is opened or a module is activated, it shows a warning/error message stating that

`Canon Canon Powershot A3100 IS color matrix not found
`
Rawtherapee and Photoflow show "normal" values. By "normal" I mean a color temperature around 6, 7K, not 25K as show by darktable.

**To Reproduce**
1. Open attached dng file
2. See default white balance values
3. Open the same file in Rawtherapee and/or Photoflow and compare.

**Expected behavior**
1) Color temperature is way too high (25K)
2) Colors are very faint

**Screenshots**
After opening file in dt:
Screenshot from 2019-11-24 23-59-45

After opening file in rt:
Screenshot from 2019-11-24 23-27-12

After opening file in PhF:
Screenshot from 2019-11-24 23-28-00

 - OS: ubuntu 18.04
 - Version 3.0.0rc1-git82.cde1d

**Additional context**
raw file:

[IMG_5861.zip](https://github.com/darktable-org/darktable/files/3884577/IMG_5861.zip)

jenshannoschwalm:
One more correction matrix and two more cases handled. Fixes #3555 wrong colour temperature.

### DIFF
--- a/src/common/exif.cc
+++ b/src/common/exif.cc
@@ -976,6 +976,14 @@ static bool dt_exif_read_exif_data(dt_image_t *img, Exiv2::ExifData &exifData)
         for(int i = 0; i < 9; i++) colmatrix[i] = cm1_pos->toFloat(i);
         illu = illu1;
       }
+      // In a few cases we only have one color matrix; it should not be corrected
+      if(illu == -1 && cm1_pos != exifData.end() && cm1_pos->count() == 9 && cm1_pos->size())
+      {
+        for(int i = 0; i < 9; i++) colmatrix[i] = cm1_pos->toFloat(i);
+        illu = 0;
+      }
+
+
       // Take the found CalibrationIlluminant / ColorMatrix pair.
       // D65 or default: just copy. Otherwise multiply by the specific correction matrix.
       if(illu != -1)
--- a/src/common/exif.cc
+++ b/src/common/exif.cc
@@ -936,7 +936,7 @@ static bool dt_exif_read_exif_data(dt_image_t *img, Exiv2::ExifData &exifData)
       // The correction matrices are taken from
       // http://www.brucelindbloom.com - chromatic Adaption.
       // using Bradford method: found Illuminant -> D65
-      const float correctmat[6][9] = {
+      const float correctmat[7][9] = {
         { 0.9555766, -0.0230393, 0.0631636, -0.0282895, 1.0099416, 0.0210077, 0.0122982, -0.0204830,
           1.3299098 }, // 23 = D50
         { 0.9726856, -0.0135482, 0.0361731, -0.0167463, 1.0049102, 0.0120598, 0.0070026, -0.0116372,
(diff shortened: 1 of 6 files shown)


jenshannoschwalm:
The first commit didn't take care of some special dng files having a ColorMatrix but no Illuminant. Such files can be found via #3556, i was not aware of this, but dng specs seem to allow this. This lead to an undefined color matrix resulting in a greenish image.
Fixed in second commit.
 
Some dng files have a regular pixel order and are **not**raw. The global test in image.c as suggested in #3569 is not a good idea, instead the strictly **raw-only** iops test directly via the flag to make sure. 
Rawprepare has not been taken care of as this breaks monochrome Leicas.

@TurboGit concider this pr as relevant, can be tested in a simpler way. #3569 will be closed.

TurboGit:
The non RAW dng now seems to have back the demosaic module enabled. We need a three state check:

- plan RAW
- DNG RAW
- DNG non RAW

Changing dt_image_is_raw() to use a direct flags mask does not seems a right move to me. Better to have a proper check for special case the DNG?

TurboGit:
@jenshannoschwalm : I'm wondering if dt_image_is_raw() is not bogus in fact! Why do we return TRUE if rawspeed did not put the DT_IMAGE_RAW flag? Why do we expect all non listed extensions to be RAW file?

Maybe separate both routines?

dt_image_is_raw() just checking the flags.

dt_image_supported() just to check for all supported extensions?

Or as an alternative:

`dy_image_is_raw((const dt_image_t *img, gboolean guess);`

if guess is TRUE then we check the extensions? That's just proposal, but I think it is bad to not use is raw and directly check the flag in the code.

jenshannoschwalm:
> Why do we expect all non listed extensions to be RAW file?

You are right in the above, was just going to mention, the assumption is simply wrong.

In fact, i think the question itself is wrong. And it's not only a tri-state test, we have always non-raw, raw_monochrome ...).

dt_image_is_raw() is used at many places but we should ask precisely

1. ca_correct_possible(const dt_image_t *img)
2. matrix_coreection(const dt_image_t *img)

For the commits the question would have been image_really_raw by checking flags.

jenshannoschwalm:
#3573 could also be related ...

jenshannoschwalm:
BTW i am sure the code in this pr works right as it is now; maybe you have a test first and we decide later before merging how we define the **correct** testing?

Also - if
1. using an already developed image
2. we apply a style or copy &pase history
3. Incorrect setting of auto-apply of modules

there could be a situation where a module is active in history/pipeline but it should **not** be. How should this be handled? Do some testing and auto-remove such modules?

TurboGit:
@jenshannoschwalm : Yes my testing shows that it is working. I'm really not fond of the change to read directly the flags. But if this ts temporary solution (really temporary!) then I can merge this in 3.0.

Your proposal to have different routines with high level semantics like:

```
dt_image_is_ca_correct_supported(imgid)
dt_image_is_matrix_correction_supported(imgid)
```

is probably the way to go and we may do that later in 3.0.1. 

Does that plan works for you?

jenshannoschwalm:
Your plan is fine for me and i understand that i will have to make a **proper** pr for 3.01 :-)
Please close issues #3555 and #3556 :-) #3573 has not been checked yet

TurboGit:
Great! Merging this then.

PR #3566, opened 2019-11-25
fix #3516 - make lut size 256 work and add messages to limit input lut size to 256

### LINKED ISSUE #3516: 3D LUT → segmentation fault
System: Debian 10 stable + dt master builds. Segmentation fault on loading big HaldLUT.

File: 
https://drive.google.com/open?id=1WoPFoouBbJlbQa5VsWXiT2kPJMH5NREI

File comes from:
http://rawpedia.rawtherapee.com/Film_Simulation#RawTherapee_Film_Simulation_Collection

Additional: where was report on mailing list that dt windows builds corrupt image but do not fail.

phweyland:


phweyland:
Just seen another issue:

image

doesn't match with:

image

REC.2020 is actually used instead of prophoto. Fixing.

src/iop/lut3d.c
```diff
@@ -378,6 +378,14 @@ uint8_t calculate_clut_haldclut(char *filepath, float **clut)
     return 0;
   }
   level *= level;  // to be equivalent to cube level
+  if(level > 256)
+  {
+    fprintf(stderr, "[lut3d] error - LUT 3D size %d > 256\n", level);
+    dt_control_log(_("error - LUT 3D size %d > 256"), level);
```
TurboGit: This breaks translation... Maybe we can wait for 3.0.1. Also as this is a user's message I would just wrote:

   "Lut 3D size exceed the maximum supported"

The fprintf will tell what is the issue to developers.

src/iop/lut3d.c
```diff
@@ -614,6 +622,14 @@ uint8_t calculate_clut_cube(char *filepath, float **clut)
       else if (strcmp("LUT_3D_SIZE", token[0]) == 0)
       {
         level = atoll(token[1]);
+        if(level > 256)
+        {
+          fprintf(stderr, "[lut3d] error - LUT 3D size %d > 256\n", level);
+          dt_control_log(_("error - LUT 3D size %d > 256"), level);
```
TurboGit: likewise

src/iop/lut3d.c
phweyland: If that breaks translation only for that module this is maybe affordable ...
I'll make the changes.

phweyland: We could just protect dt from fault (for 3.0) and add proper messages later on (for 3.0.1). Let me know.

src/iop/lut3d.c
phweyland: ok

TurboGit:
You still have ProPhoto RGB referenced at the top of the file:

```
typedef enum dt_iop_lut3d_colorspace_t
{
  DT_IOP_SRGB = 0,
  DT_IOP_REC709 = 1,
  DT_IOP_LIN_REC709 = 2,
  DT_IOP_LIN_PROPHOTORGB = 3,
} dt_iop_lut3d_colorspace_t;
```

src/iop/lut3d.c
parafin: Does adding new strings break the translation? Modifying - sure, but adding a new string will probably just mean that it will be left untranslated. If there is some technical problem with adding a new string, then we can just leave it non-translatable in 3.0 (remove _()) and revert it for 3.0.1, how about that?

phweyland:
> You still have ProPhoto RGB referenced at the top of the file:

Well done. I'll fix that too.

TurboGit:
Given that this fixes a crash I will certainly consider this PR for 3.0 even though it breaks translation. Better to have an English message on the interface than crashing dt!

TurboGit:
@phweyland : I tested with the photo from Timur and I still get a crash. Do you confirm?

src/iop/lut3d.c
```diff
@@ -378,33 +378,45 @@ uint8_t calculate_clut_haldclut(char *filepath, float **clut)
...
+    png_destroy_read_struct(&png.png_ptr, &png.info_ptr, NULL);
     return 0;
   }
   if (read_image(&png, buf))
   {
-    dt_free_align(buf);
     fprintf(stderr, "[lut3d] error - could not read png image `%s'\n", filepath);
     dt_control_log(_("error - could not read png image %s"), filepath);
+    dt_free_align(buf);
+    fclose(png.f);
```
TurboGit: After read image this is not needed and indeed crash dt.

src/iop/lut3d.c
```diff
@@ -378,33 +378,45 @@ uint8_t calculate_clut_haldclut(char *filepath, float **clut)
...
   const size_t buf_size_lut = (size_t)png.height * png.height * 3;
   dt_print(DT_DEBUG_DEV, "[lut3d] allocating %zu floats for png lut - level %d\n", buf_size_lut, level);
   float *lclut = dt_alloc_align(16, buf_size_lut * sizeof(float));
   if(!lclut)
   {
-    dt_free_align(buf);
     fprintf(stderr, "[lut3d] error - allocating buffer for png lut\n");
     dt_control_log(_("error - allocating buffer for png lut"));
+    dt_free_align(buf);
+    fclose(png.f);
```
TurboGit: likewise

src/iop/lut3d.c
```diff
@@ -419,6 +431,8 @@ uint8_t calculate_clut_haldclut(char *filepath, float **clut)
       lclut[i] = (256.0f * (float)buf[2*i] + (float)buf[2*i+1]) * norm;
   }
   dt_free_align(buf);
+  fclose(png.f);
```
TurboGit: likewise

src/iop/lut3d.c
phweyland: Sorry, I remove that. Doesn't crash on my system ...

phweyland:
Removed the faulty statements. 
The Timur's lut works for me.

src/iop/lut3d.c
TurboGit: @parafin: adding a new string or changing it has the same result, it means that the string is not translated anymore. That is what I call "breaking translation", we promise a string freeze and we do not  fully do it. But this is better than a crash in this case, so will merge as soon it is stable.

TurboGit:
Looks ok now. Thanks.

PR #3569, opened 2019-11-25
Fix #3556 Disable demosaic for non-raw dng files.

### LINKED ISSUE #3556: DNG → fix demosaic UI

jenshannoschwalm:
Fixes issue #3556.

DNG files can have both types of data
1. raw      demosaic module enabled
2. non-raw  demosaic should be disabled

As dng files are tested in rawspeed for a bayer matrix they have the DT_IMAGE_RAW set and
will return true in dt_image_is_raw(const dt_image_t *img).
Including a test for the .dng file extension is therefore possible.

### DIFF
--- a/src/common/image.c
+++ b/src/common/image.c
@@ -113,7 +113,7 @@ int dt_image_is_raw(const dt_image_t *img)
   const char *dt_non_raw_extensions[]
       = { ".jpeg", ".jpg",  ".pfm", ".hdr", ".exr", ".pxn", ".tif", ".tiff", ".png",
           ".j2c",  ".j2k",  ".jp2", ".jpc", ".gif", ".jpc", ".jp2", ".bmp",  ".dcm",
-          ".jng",  ".miff", ".mng", ".pbm", ".pnm", ".ppm", ".pgm", NULL };
+          ".jng",  ".miff", ".mng", ".pbm", ".pnm", ".ppm", ".pgm", ".dng", NULL };
 
   if(img->flags & DT_IMAGE_RAW) return TRUE;
 
--- a/src/common/image.c
+++ b/src/common/image.c
@@ -110,6 +110,8 @@ int dt_image_is_hdr(const dt_image_t *img)
 int dt_image_is_raw(const dt_image_t *img)
 {
   // NULL terminated list of supported non-RAW extensions
+  // rawspeed takes care of RAW dng files by setting the DT_IMAGE_RAW flag
+  // so NON-RAW dng files are tested correctly.  
   const char *dt_non_raw_extensions[]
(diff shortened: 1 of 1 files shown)


src/common/image.c
```diff
@@ -113,7 +113,7 @@ int dt_image_is_raw(const dt_image_t *img)
   const char *dt_non_raw_extensions[]
       = { ".jpeg", ".jpg",  ".pfm", ".hdr", ".exr", ".pxn", ".tif", ".tiff", ".png",
           ".j2c",  ".j2k",  ".jp2", ".jpc", ".gif", ".jpc", ".jp2", ".bmp",  ".dcm",
-          ".jng",  ".miff", ".mng", ".pbm", ".pnm", ".ppm", ".pgm", NULL };
+          ".jng",  ".miff", ".mng", ".pbm", ".pnm", ".ppm", ".pgm", ".dng", NULL };
 
   if(img->flags & DT_IMAGE_RAW) return TRUE;
```
TurboGit: I would add comment to make this clear:

// rawspeed set the DT_IMAGE_RAW to DNG file which are RAW. The non RAW DNG file will be taken into
// account below.

Or something like that.

jenshannoschwalm: done with second commit.

TurboGit:
When I import the test image from Timur with this PR I get a string green cast on the whole image. I have tested with #3565 too. Anything missing?

jenshannoschwalm:
I confess i have not cross-tested.

chhil:
@jenshannoschwalm should upgrades from 2.6.x to 3.0take this into account ?

Entropy512:
@TurboGit Linear DNGs should (in most cases) be treated as demosaiced but still in camera colorspace.

e.g. skip demosaic but don't skip any of the colormatrix stuff.  Sounds like this is skipping the color matrix and white balance?

jenshannoschwalm:
@Entropy512 i think you are absolutely right and i missed that. Will close this pr and include a correct version in #3555 for better testing soon.

jenshannoschwalm:
@Entropy512 you were correct but it was not the whole story :-)

> Linear DNGs should (in most cases) be treated as demosaiced but still in camera colorspace.

Can you provide an example file?

jenshannoschwalm:
@TurboGit See #3565 ...

PR #3590, opened 2019-11-27
lighttable : ensure to update dnd state after layout change

### LINKED ISSUE #3582: Custom order drag-and-drop not re-enabled when switching from culling to filemanager mode
In lighttable view, file manager mode, when the sort order is switched to custom sort, drag and drop is enabled to allow reordering the images. When the mode is switched to culling, drag and drop is disabled. When it is switched back to file manager, it is not re-enabled again, so now it is not possible to change the order of the images.

The same issue arises when the sort order is switched to custom sort while in culling mode; drag and drop is not enabled and neither is it enabled when afterwards the mode is switched to file manager.

When the sort order is changed in darkroom view, then, upon switching to lighttable/filemanager, drag and drop is correctly enabled.

AlicVB:
this fix #3582

### DIFF
--- a/src/views/lighttable.c
+++ b/src/views/lighttable.c
@@ -400,6 +400,10 @@ static void check_layout(dt_view_t *self)
     g_timeout_add(200, _expose_again_full, self);
     _scrollbars_restore();
   }
+
+  // put back or desactivate the reorder dnd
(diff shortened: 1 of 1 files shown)


TurboGit:
Looks good, thanks!

dterrahe:
Thank you!

PR #3594, opened 2019-11-27
avoid accels conflict with libs (filmstrip)

### LINKED ISSUE #3562: Accelerators for lighttable and filmstrip conflict

AlicVB:
this fix #3562 , close #3563 
lib can now init their accels for specific views. But it's up to them to not connect them (to avoid a general check which would take little time when changing views)
this is applied to filmstrip, to avoid conflict with ratings/etc which are handle direcly by the view code in lighttable and by filmstrip in other views

thanks to @dterrahe for report, ideas and tests

### DIFF
--- a/src/gui/accelerators.c
+++ b/src/gui/accelerators.c
@@ -164,7 +164,8 @@ void dt_accel_register_iop(dt_iop_module_so_t *so, gboolean local, const gchar *
   darktable.control->accelerator_list = g_slist_prepend(darktable.control->accelerator_list, accel);
 }
 
-void dt_accel_register_lib(dt_lib_module_t *self, const gchar *path, guint accel_key, GdkModifierType mods)
+void dt_accel_register_lib_for_views(dt_lib_module_t *self, dt_view_type_flags_t views, const gchar *path,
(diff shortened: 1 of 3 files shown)


TurboGit:
@AlicVB : Given the patch I would not merge it in 3.0 but will wait for 3.0.1. We have lived this for some time and we certainly still can wait a bit even if this is annoying. But at this stage I really want to play the safety and finally stabilize dt for the finale release.

AlicVB:
no problem. I agree with you, this is far from critical !

dterrahe:
Absolutely; this is just polishing and not a stability issue. No need to risk causing problems in the release.

TurboGit:
Works for me. Thanks.

PR #3596, opened 2019-11-27
navigation : verify imgid before drawing

### LINKED ISSUE #3575: Miniature → Caching
Steps to reproduce:

1. Start DT in lightroom mode
2. Open any file in darkroom mode
3. Leave darkroom by pressing 'l'
4. Open any other file in darkroom mode

Notice that miniature (picture in top left corner) always shows previous picture before current picture is loaded and modules settings applied. Probably it is cached somewhere. Just showing nothing will do fine, showing previous picture is confusing.

AlicVB:
this fix #3575
Same check is already applied for darkroom main view.

In a second time, we may want to draw some "waiting screen" when entering darkroom, to avoid keeping a "freezed" image of the previous view before the image is loaded.
This is especially visible if image has heavy settings (I usually use denoise profiled with search radius at maximum for testing)

### DIFF
--- a/src/libs/navigation.c
+++ b/src/libs/navigation.c
@@ -174,7 +174,7 @@ static gboolean _lib_navigation_draw_callback(GtkWidget *widget, cairo_t *crf, g
   gtk_render_background(context, cr, 0, 0, allocation.width, allocation.height);
 
   /* draw navigation image if available */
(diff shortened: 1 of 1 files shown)


TurboGit:
Nice! Thanks.

PR #3601, opened 2019-11-28
Add sliders in cropping module to adjust margins

dterrahe:
This allows (dynamic) keyboard shortcuts to be defined to adjust margins without going into the crop/rotate/keystone module.

The max boundary of each slider is adjusted as its opposite slider is moved, to show that the margins cannot overlap.

### DIFF
--- a/src/iop/clipping.c
+++ b/src/iop/clipping.c
@@ -204,6 +204,8 @@ int legacy_params(dt_iop_module_t *self, const void *const old_params, const int
 
 typedef struct dt_iop_clipping_gui_data_t
 {
+  GtkNotebook *notebook;
+
   GtkWidget *angle;
   GtkWidget *cx,*cy,*cw,*ch;
   GtkWidget *hvflip;
@@ -2024,8 +2026,24 @@ void gui_init(struct dt_iop_module_t *self)
   g->k_selected = -1;
   g->old_width = g->old_height = -1;
 
-  self->widget = gtk_box_new(GTK_ORIENTATION_VERTICAL, DT_BAUHAUS_SPACE);
+  // Init GTK notebook
+  self->widget = gtk_box_new(GTK_ORIENTATION_VERTICAL, 0);
+
+  g->notebook = GTK_NOTEBOOK(gtk_notebook_new());
+  GtkWidget *page1 = GTK_WIDGET(gtk_box_new(GTK_ORIENTATION_VERTICAL, 0));
+  GtkWidget *page2 = GTK_WIDGET(gtk_box_new(GTK_ORIENTATION_VERTICAL, 0));
+  GtkWidget *page3 = GTK_WIDGET(gtk_box_new(GTK_ORIENTATION_VERTICAL, 0));
+
+  gtk_notebook_append_page(GTK_NOTEBOOK(g->notebook), page1, gtk_label_new(_("main")));
+  gtk_notebook_append_page(GTK_NOTEBOOK(g->notebook), page2, gtk_label_new(_("margins")));
+  gtk_notebook_append_page(GTK_NOTEBOOK(g->notebook), page3, gtk_label_new(_("guides")));
+  gtk_widget_show_all(GTK_WIDGET(gtk_notebook_get_nth_page(g->notebook, 0)));
+  gtk_box_pack_start(GTK_BOX(self->widget), GTK_WIDGET(g->notebook), FALSE, FALSE, 0);
+
+  dtgtk_justify_notebook_tabs(g->notebook);
+
   dt_gui_add_help_link(self->widget, dt_get_help_url(self->op));
+
   g->hvflip = dt_bauhaus_combobox_new(self);
   dt_bauhaus_widget_set_label(g->hvflip, NULL, _("flip"));
   dt_bauhaus_combobox_add(g->hvflip, _("none"));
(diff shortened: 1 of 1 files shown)


MStraeten:
i'd suggest to split the module options on to tabs like in tone equalizer. The boundary sliders could be located on an Expert mode tab, so the UI is not that overcrowded

TurboGit:
I agree, this will avoid a too crowded module.

dterrahe:
Since I added a notebook anyway, I moved the "guides" settings to a separate tab as well, since they can add quite a few sliders as well, especially when "grid" is selected. However, the widgets on the guides tab get stretched vertically to fill the whole page, so if a simple guide (or none) is selected that doesn't require many additional widgets, the dropdowns become multi-line. I think there might need to be an adjustment in css to avoid this behaviour or maybe the default could be changed somewhere. I haven't seen any other modules suffering from this.

dterrahe:
I was also considering changing the "keystone" dropdown into a set of buttons:
horizontal, vertical, both, clear. Either the first three or the last one would be enabled, depending on whether a keystone adjustment has currently been applied. The icons from perspective correction/automatic fit could be reused for this. Makes sense? Maybe I'm missing a rationale for the current solution.

TurboGit:
Ok, let's go with a tab for the guides and one for the crop sliders.I won't change others things now, just let's see how we stand after those changes.

TurboGit:
@dterrahe : After testing I find the current PR ok as proposed. The only thing is fixing the multi-line guide. A really issue...

TurboGit:
Found the issue and fixed it. Will merge in master and 3.0.1, thanks!

jakubfi:
@TurboGit @dterrahe Guys, I worked a bit with the tabbed version of "crop and rotate" and I think separating the guides controls from the main tab was not a good idea. Margins - sure, those are perfect for a separate tab. But main and guides in my oppinion should be merged back together.

How I see it:
- Selecting guides is one of the most important features of the cropping module (this is what helps us with the composition) and should be always readily available.
- There are some specific workflows which rely on frequent switching between guides:
    - image straightening (by eye): rotation is on the main tab, but switching to helpful  'metering' or 'grid' guides require changing tabs,
    - recomposition when changing image aspect: again - aspect is on the main tab.
- The amount of vertical space saved in modules area is negligible in case of most guides (with the exception of 'grid' guides, which are rather rarely used, I would assume).

In ~90% of my usage cases I personally find it irritating to have guides on a separate tab.
Thoughts? Comments?

dterrahe:
Thanks for the feedback! It was just an exeriment really, so I'm perfectly happy to change that back. Shall i just reduce the number of tabs to two, or, since there is a notebook tab line now anyway, is there a less useful selection of controls i could move to an "expert" tab?
 
 
  On Mon, Dec 30, 2019 at 18:14, jakubfi<notifications@github.com> wrote:   
@TurboGit @dterrahe Guys, I worked a bit with the tabbed version of "crop and rotate" and I think separating the guides controls from the main tab was not a good idea. Margins - sure, those are perfect for a separate tab. But main and guides in my oppinion should be merged back together.

How I see it:
   
   - Selecting guides is one of the most important features of the cropping module (this is what helps us with the composition) and should be always readily available.
   - There are some specific workflows which rely on frequent switching between guides:      
      - image straightening (by eye): rotation is on the main tab, but switching to helpful 'metering' or 'grid' guides require changing tabs,
      - recomposition when changing image aspect: again - aspect is on the main tab.

   - The amount of vertical space saved in modules area is negligible in case of most guides (with the exception of 'grid' guides, which are rather rarely used, I would assume).

In ~90% of my usage cases I personally find it irritating to have guides on a separate tab.
Thoughts? Comments?

TurboGit:
Well, I'm not using a lot the guides and it expands and change the size of the widget. Now that it is on a separate tab this effect is fixed. To me this is perfect as-is. But let's see what others think about this.

jakubfi:
@TurboGit If you're not using guides a lot I guess you don't see the effect of module size changing too often. ;-)
One idea would be to put those guides with additional parameters to the end of the list, to reduce the effect. Another idea would be to compact the 'horizontal' and 'vertical' sliders of the 'grid' guides into one line (they don't need much slider resolution), and maybe even triple their max value and get rid of 'subdivisions' alltogether?

jakubfi:
P.S. Or even better: just reduce those three sliders for 'grid' to one: number of lines (for both horizontal and vertical) ranging form 1 to, say, 40.

jakubfi:
@TurboGit can you reopen this, so people can notice the discussion and voice their opinion, or should I open a separate issue?

TurboGit:
I cannot reopen a merged PR, only issues can be reopened. So yes please, can you create a new issue?

TurboGit:
@dterrahe : see #3968, looks like merging back guides to main tab is the way to go. Can you do that? Thanks.

dterrahe:
Will do when I get back to the pc tomorrow.

Should I change anything in the order of the guides dropdown? @jakubfi suggested putting "grid" at the end of the list (because it adds most parameters).

A few of the guides are symmetrical, so they don't need a "flip" dropdown. Any objections against me trying to hide it for those cases? I believe this even applies to all settings of "grid" so it would reduce the maximum number of widgets by one.

jakubfi:
@dterrahe I would suggest merging 'main' and 'guides' tabs first and then, in a separate PR, dealing with cleaning up other things like removing unnecessary "flips" and possibly simplifying grid guides parametrization. Solve the problem at hand and go from there. :-) Small steps - easier to take, manage, and test, easier to introduce and discuss.

TurboGit:
Agreed. One step at a time. I would certainly add disabling the ineffective flip controls in the first iteration thought as there is no point keeping ineffective controls.

PR #3614, opened 2019-11-28
fix some casts

rabauke:
Casting to `size_t` after (rather than before) multiplication might lead to integer overflow.


TurboGit:
Looks good to me. Thanks.

PR #3615, opened 2019-11-28
Fix selected text issue and improve dialog bar

### LINKED ISSUE #3610: UI → Selected text in default theme
Select any text, lets say, in image information. This selection is almost the same in text color and background as regular text. It should stand out.

Screenshot (that line is actually selected):
selected

Nilvus:
This fix #3610 and improve dialog windows (especially to add missing contrast in import options fields).

@timur-davletshin, @blitzgneisserin and @GLLM : please test if that doesn't break any other label or textview I don't see. It's normally ok. If you find any other part missing, please precise it here.

### DIFF
--- a/data/themes/darktable.css
+++ b/data/themes/darktable.css
@@ -675,8 +675,8 @@ menuitem check
 dialog,
 filechooserdialog
 {
-  background-color: @bg_color;
-  color: @fg_color;
+  background-color: @plugin_bg_color;
+  color: @plugin_fg_color;
 }
 
 dialog *,
@@ -1319,7 +1319,9 @@ entry:active
   background-color: @field_active_bg;
 }
 
-entry:selected
+entry:selected,
+label *,
+textview text *
 {
   color: @field_selected_fg;
   background-color: @field_selected_bg;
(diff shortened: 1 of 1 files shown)


timur-davletshin:
Looks fine for me.

blitzgneisserin:
will test asap.

blitzgneisserin:
looks ok

TurboGit:
Probably worth adding that in 3.0 ? Is that still WIP ?

Nilvus:
> Probably worth adding that in 3.0 ? Is that still WIP ?

Of course for 3.0 here. I'm not 100% satisfied with the all tag (the * after label and textview) because that select all things after these tags (so possibly not wanted one, even if it seems not). I was expected the :selected add to work as it is possible with Gtk on what I find but I don't understand why it doesn't work. The only option working I find, after many tries, is that proposal. By this time, I didn't find anything changed other than what is expecting. I will just wait a little to use darktable more and be sure I didn't miss something else. And maybe I will think/find a better way (more selected one). Or maybe it's just the good/better way.

So as we have time before 3.0 release, I prefer to be careful.

Nilvus:
@TurboGit : just find safer fix here and finally simple one. Now good to be merged.

TurboGit:
Looks good to me, thanks!

PR #3616, opened 2019-11-29
Use blendop params when style applied on export

### LINKED ISSUE #3571: blendop_rgb and blendop_mask_rgb are not applied in style
I set Profiled Denoise to use parametric mask (to apply heavier blending for darker parts), then save it as a style to apply only on export (to not slow down my editing). This is working correctly in 2.6.

However, on latest master branch, the style applied on export doesn't have the blending applied. If I apply the style to the image in lighttable, then export without the style, it still works fine.

I ran darktable -d all, and compared the console outputs of the two cases. The only difference is the missing of "blendop_mask_rgb" and "blendop_rgb" (between "denoiseprofile_backtransform" and "denoiseprofile_init").

**To Reproduce**
1. Apply a module with parametric mask
2. Save the module to a style, and apply the style on export
3. The parametric mask doesn't apply on export.

**Expected behavior**
The parametric mask should be applied when applying the module from a style on export.

**Platform:**
 - OS: Ubuntu
 - Version 18.04

**Additional context**
Original bug report: https://redmine.darktable.org/issues/12690

wpferguson:
Changed module->default_blendop_params to style_item->blendop_params so that masks are applied to styles on export.  This makes option 1 and option 3 of the if statement the same, but option 3 is the one taken when the style is applied on export, so I left it that way in case some other processing needs to happen when a style is applied on export versus a style applied in darktable.

Fixes #3571.

### DIFF
--- a/src/common/styles.c
+++ b/src/common/styles.c
@@ -593,11 +593,13 @@ void dt_styles_apply_style_item(dt_develop_t *dev, dt_style_item_t *style_item, 
               && dt_develop_blend_legacy_params(module, style_item->blendop_params, style_item->blendop_version,
                   module->blend_params, dt_develop_blend_version(), style_item->blendop_params_size) == 0)
       {
+        // legacy style
         // do nothing
       }
       else
       {
-        memcpy(module->blend_params, module->default_blendop_params, sizeof(dt_develop_blend_params_t));
+        // apply style on export
+        memcpy(module->blend_params, style_item->blendop_params, sizeof(dt_develop_blend_params_t));
       }
 
       if(module->version() != style_item->module_version || module->params_size != style_item->params_size


src/common/styles.c
```diff
@@ -593,11 +593,13 @@ void dt_styles_apply_style_item(dt_develop_t *dev, dt_style_item_t *style_item,
...
                   module->blend_params, dt_develop_blend_version(), style_item->blendop_params_size) == 0)
       {
+        // legacy style
         // do nothing
       }
       else
       {
-        memcpy(module->blend_params, module->default_blendop_params, sizeof(dt_develop_blend_params_t));
+        // apply style on export
+        memcpy(module->blend_params, style_item->blendop_params, sizeof(dt_develop_blend_params_t));
```
TurboGit: I don't think this is the right fix. This else part is when the blend_param from the style does not correspond to the current blend version. So we cannot use style_item blendop_params and we just set to default blend.

The case where the version matches is in the first part of the if statement above.

TurboGit:
Closing see : c0fd32cd.

PR #3617, opened 2019-11-29
Suppress [dt_ioppr_transform_image_colorspace] warning for monochromes

jenshannoschwalm:
Monochrome images from scanners or a few Leicas don't want a
dt_ioppr_transform_image_colorspace but just set converted_cst = cst_from.

The warning: [dt_ioppr_transform_image_colorspace] module %s must be between input color profile and output color profile\n"
is simply misleading so it is suppressed now by checking by
dt_image_is_monochrome(&self->dev->image_storage)


TurboGit:
Thanks !

Issue #2808, opened 2019-07-23
Darktable Crash while using  ColorZone with exception log


chhil:
**Describe the bug**
Tweaking the color zone module ,crashed Darktable. This happened twice, the second time I was doing the same image with the colorzone module. Stacktrace generated has been attached.
[darktable_bt_DQIF5Z.txt](https://github.com/darktable-org/darktable/files/3421635/darktable_bt_DQIF5Z.txt)

Always happen during parametric mask while moving the right bound inwards.
![image](https://user-images.githubusercontent.com/1119439/61706528-eb26c700-ad65-11e9-9eaf-4da48c31ff95.png)

**Platform (please complete the following information):**
 - OS:Win 10.
 - Version 2.7.0+1558~gefe494082

chhil:
Basically parametric mask is broken.

TurboGit:
Sure I'm using parametric masks very often and I have yet to get a crash. Thanks for helping us understanding the issue!

chhil:
I am no longer seeing that behavior with 2.7.0+1594.

Something definitely changed between 1558 and 1594.

edgardoh:
I don't have a crash, and with a normal use everything is fine, but if I stress it I get some strange things, like pixman/gtk/cairo errors/warnings.
It take a while to duplicate, and it happens on the color zones and tone curve (maybe some other modules too). 
I activate the parametric mask, and play with the curve and gradient slider, always with the module's or blend module's color picker active.
A couple of times, when I return to lighttable, the side panels are not visible (very scary), more often I get this:

(darktable:1079): Pango-CRITICAL **: 12:34:10.985: pango_layout_get_line_count: assertion 'layout != NULL' failed

(darktable:1079): Gtk-CRITICAL **: 12:34:10.985: gtk_render_layout: assertion 'PANGO_IS_LAYOUT (layout)' failed

And the image on darkroom stops refreshing.

I coudln't find what's going on, maybe some gtk/cairo resource not being free?

chhil:
I still see it. This exception log has more info.
[darktable_bt_RPDS5Z.txt](https://github.com/darktable-org/darktable/files/3464561/darktable_bt_RPDS5Z.txt)

chhil:
Still happens with parametric masks in 3.0.rc1+22. This time it was in contrast equalizer while moving the sliders around to mask the right area.
Exception Log is not helpful.
[...]

PR #3618, opened 2019-11-29
Fix thread safety of picker handling.

TurboGit:
The apply routine is called from the pixelpipe and from there we call
some Gtk routines. This is not fine as we are then calling Gtk routines
outside of the main Gtk loop.

We new record a call from the main loop for the apply picker hook.

Possible fix for #2808.

### DIFF
--- a/src/gui/color_picker_proxy.c
+++ b/src/gui/color_picker_proxy.c
@@ -96,6 +96,22 @@ static void _internal_iop_color_picker_update(dt_iop_color_picker_t *picker)
   darktable.gui->reset = reset;
 }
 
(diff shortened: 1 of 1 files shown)


TurboGit:
Tested to death... Seems really ok on my side!

jenshannoschwalm:
testing to death is something i wouldn't suggest professionally :-)

If i understand your code correctly you define a callback from gtk code into the pipeline? That would mean you are sure about the pipes threading safety. Maybe add a mutex encapsulating pipe?

TurboGit:
Closing, better solution here : #3621

PR #3626, opened 2019-11-29
crop dimensions string visibility fixes

### LINKED ISSUE #3624: clipping: crop dimensions label is sometimes rendered off-screen
When cropping while zoomed in, the default placement of the crop dimensions string (e.g. "1920x1080") at the center of the crop window can result in the dimensions string being rendered partially or completely off-screen.

Example:

screenshot-clip-label-offscreen

### LINKED ISSUE #3625: clipping: crop dimensions label can be hard to see over some images
The crop dimensions string (e.g. "1920x1080") can be difficult to see when rendered over an image area with pixels similar to the text color.

Example:

screenshot-clip-label-bad-vis

kamalmostafa:
Fixes #3624; Fixes #3625.

### DIFF
--- a/src/iop/clipping.c
+++ b/src/iop/clipping.c
@@ -2312,7 +2312,7 @@ void gui_post_expose(struct dt_iop_module_t *self, cairo_t *cr, int32_t width, i
     char dimensions[16];
     dimensions[0] = '\0';
     PangoLayout *layout;
-    PangoRectangle ink;
+    PangoRectangle ext;
     PangoFontDescription *desc = pango_font_description_copy_static(darktable.bauhaus->pango_font_desc);
     pango_font_description_set_weight(desc, PANGO_WEIGHT_BOLD);
     pango_font_description_set_absolute_size(desc, DT_PIXEL_APPLY_DPI(16) * PANGO_SCALE / zoom_scale);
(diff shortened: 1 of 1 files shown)


src/iop/clipping.c
```diff
@@ -2324,9 +2324,24 @@ void gui_post_expose(struct dt_iop_module_t *self, cairo_t *cr, int32_t width, i
     snprintf(dimensions, sizeof(dimensions), "%.0fx%.0f", (float)procw * g->clip_w, (float)proch * g->clip_h);
```
TurboGit: As we are at it, can you add spaces around x :  "%.0f x %.0f"

This will be even more readable I think. Thanks.

TurboGit:
If you can do that soon I'll even consider this for 3.0 instead of 3.0.1. Given that 3.0 is a full rework of the GUI it will be nice to have this enhancement for the final release.

src/iop/clipping.c
kamalmostafa: Sure.  And while we're at that, what about using a Unicode "Multiplication X" character instead of the regular lower-case x?  So:  **1920 ✕ 1080**

https://www.fileformat.info/info/unicode/char/2715/index.htm

I think that would look even nicer, and the center of the X would mark the exact center of the crop window if we're careful to use e.g. "%5.0f ✕ %-5.0f".

Any objection to using that Unicode "Multiplication X"?

TurboGit: Yes, sadly last time I have introduced an Unicode character in dt it was an issue for Windows. So I would prefer a plan x for now until this is fixed (maybe need another font) on Windows.

kamalmostafa: Yep, I had that concern too.  Fwiw though, iop/clipping.c does already use a unicode character...
`pango_layout_set_text(layout, "ꝏ", -1); `
... though I do not know whether it renders correctly in Windows.  ;-)

The commit I'm adding to this pull req just sticks with the plain old lower-case "x".

TurboGit:
Ok, I'll merge now. Note that there is many variables that can be made consts. Will do that, but please remember for your next PR. Thanks.

src/iop/clipping.c
parafin: FYI, these symbols aren't rendered for me on Linux, at least in a web browser. So please don't use such characters, it's not just about Windows.

Issue #3629, opened 2019-11-30
Exposure iop with masks export different results CPU/GPU


groutr:
**Describe the bug**
Applying the xmp file below to any raw file will give different results if opencl is enabled/disabled.

**To Reproduce**
1. Save xmp below to a file
2. Execute `darktable-cli <input_img> <xmpfile> <output_img> --core --conf opencl=true`
and `darktable-cli <input_img> <xmpfile> <output_img> --core --conf opencl=false`
3. Observe that the exposures of the outputs differ noticeably.

**Expected behavior**
I expect both outputs to be the same.

**Screenshots**
With CPU:
[![TEST-CPU-cpu.jpg](https://i.postimg.cc/V6s7cPQw/TEST-CPU-cpu.jpg)](https://postimg.cc/v1j7WjZ2)

With GPU:
[![TEST-GPU.jpg](https://i.postimg.cc/SQJBXC3h/TEST-GPU.jpg)](https://postimg.cc/YjBnZGtn)

**Platform (please complete the following information):**
 - OS: archlinux
 - Version 3.0.0rc1+102~ge9963abec
 - intel 520 w/ NEO opencl libraries

**Additional context**
```xml
<?xml version="1.0" encoding="UTF-8"?>
<x:xmpmeta xmlns:x="adobe:ns:meta/" x:xmptk="XMP Core 4.4.0-Exiv2">
 <rdf:RDF xmlns:rdf="http://www.w3.org/1999/02/22-rdf-syntax-ns#">
  <rdf:Description rdf:about=""
    xmlns:exif="http://ns.adobe.com/exif/1.0/"
    xmlns:xmp="http://ns.adobe.com/xap/1.0/"
    xmlns:xmpMM="http://ns.adobe.com/xap/1.0/mm/"
    xmlns:dc="http://purl.org/dc/elements/1.1/"
    xmlns:darktable="http://darktable.sf.net/"
   xmp:Rating="4"
   darktable:xmp_version="3"
   darktable:raw_params="0"
   darktable:auto_presets_applied="1"
   darktable:history_end="3"
   darktable:iop_order_version="5">
   <darktable:masks_history>
    <rdf:Seq>
     <rdf:li
      darktable:mask_num="0"
      darktable:mask_id="1572019312"
      darktable:mask_type="16"
      darktable:mask_name="gradient #3"
      darktable:mask_version="4"
      darktable:mask_points="1779f63ea7dcfd3e02e486430000003f00000000"
[...]

PR #3635, opened 2019-11-30
mask refinement: fix opencl codepath for bug in opacity/contrast handling

upegelow:
fix for #3629

### DIFF
--- a/data/kernels/blendop.cl
+++ b/data/kernels/blendop.cl
@@ -1017,10 +1017,10 @@ blendop_mask_tone_curve(__read_only image2d_t mask_in, __write_only image2d_t ma
   const int x = get_global_id(0);
   const int y = get_global_id(1);
 
-  if ( x>= width || y >= height || gopacity <= 1e-4f) return;
+  if (x >= width || y >= height) return;
 
   float opacity = read_imagef(mask_in, sampleri, (int2)(x, y)).x;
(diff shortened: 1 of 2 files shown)


data/kernels/blendop.cl
```diff
@@ -1017,10 +1017,10 @@ blendop_mask_tone_curve(__read_only image2d_t mask_in, __write_only image2d_t ma
   const int x = get_global_id(0);
   const int y = get_global_id(1);
 
-  if ( x>= width || y >= height || gopacity <= 1e-4f) return;
+  if (x >= width || y >= height) return;
 
   float opacity = read_imagef(mask_in, sampleri, (int2)(x, y)).x;
-  float scaled_opacity = (2.f * opacity - 1.f) / gopacity;
+  float scaled_opacity = (2.f * opacity / gopacity - 1.f);
```
TurboGit: So we can have a negative scaled_opacity ?

TurboGit: Looking add the line below, it seems that yes. Strange but just want to be sure.

upegelow: > So we can have a negative scaled_opacity ?

Don't really understand that code. This PR is just about making CPU == GPU.

groutr:
I verified that this does indeed fix #3629.

PR #3639, opened 2019-11-30
histogram: prevent flickering when using colorpickers

upegelow:
mutex protection and double-buffering of histogram data

that flickering always looked a bit unprofessional.


TurboGit:
Thanks, looks ok.

PR #3647, opened 2019-12-01
Hide blending widgets if modules can't be enabled.

### LINKED ISSUE #3573: Hot pixels/Raw denoise → Blending operations
Blending operations are not disabled even if modules (hot pixels and raw denoise) cannot be activated (e.g. non raw file).

Should look like other non-applicable modules. Compare it with scale pixels.

jenshannoschwalm:
Modules that can't be enabled are marked by module->hide_enable_button.
In this case the blending stuff should not be shown in the modules tab.
Fixes #3573

### DIFF
--- a/src/develop/imageop.c
+++ b/src/develop/imageop.c
@@ -2014,7 +2014,7 @@ GtkWidget *dt_iop_gui_get_expander(dt_iop_module_t *module)
 
   /* add the blending ui if supported */
   gtk_box_pack_start(GTK_BOX(iopw), module->widget, TRUE, TRUE, 0);
-  dt_iop_gui_init_blending(iopw, module);
+  if(!module->hide_enable_button)  dt_iop_gui_init_blending(iopw, module);
   gtk_widget_set_name(module->widget, "iop-plugin-ui-main");
   gtk_widget_hide(iopw);
 


TurboGit:
This does not work. As an example use "raw denoise". Oopen a jpeg from lighttable, no blending OK, switch to a RAW using the filmstrip and the blending is not activated. Same with opposite, open a RAW blending activated, switch to a Jpeg and blending is still activated.

TurboGit:
Maybe the change should be done in dt_iop_request_focus() in imageop. Not tested, but I would try there.

jenshannoschwalm:
1. So far i can't see how this can be done in dt_iop_request_focus().
2. I missed the problem because it works fine in lighttable mode. Yet not understood why this is a problem in filmstrip.

jenshannoschwalm:
dt_dev_change_image seems to be the point of interest?

TurboGit:
Yes, looks like a good candidate for this.

jenshannoschwalm:
That code is somewhat obfuscicating for me. Overall selecting an image via filmstrip should enable/disable modules for the existing history, should other modules set to default and should not take any parameters/data from the image developed before. Is that correct? Excuse me for asking this, i normally don't use it and might oversee special points to take care of...

upegelow:
The "canonical" way is the following: when the user switches between images in the darkroom (e.g. via filmstrip) for each module the `gui_update()` function is called. It's main purpose is to reset the gui to a default state or the state that is required by the parameter set of the new image.
For blending there is a central function `dt_iop_gui_update_blending()` in blend_gui.c which is called for every module. Here is also the place where all the hide/show magic is happening. This is your point of entry.
You should never call any gui function from within the development pixelpipe `(dt_dev_...).` They run in different threads and in general the respective functions have side effects and are not thread safe.

TurboGit:
@upegelow : Thanks, I was looking at the best place and yes you know pretty well this part.

jenshannoschwalm:
@upegelow thanks for this help. Immediately there are two questions arising.
1. I already looked into `dt_iop_gui_update_blending()` and the dt_iop_gui_blend_data_t struct. I just don't get how i can hide the whole stuff. For each module with IOP_FLAGS_SUPPORTS_BLENDING this seems to be activated and but for always-forced-off modules like rawdenoise in jpegs the whole blending should just not be seen at all.
2. In darkroom.c `dt_dev_change_image(dt_develop_t *dev, const uint32_t imgid)` you find `gtk_widget_destroy` and `dt_ui_container_add_widget` beside others. Is that not correct to do so?

upegelow:
>   ... but for always-forced-off modules like rawdenoise in jpegs the whole blending should just not be seen at all.

Here is my idea: you hide `bd->masks_modes_box`. As a consequence the user has no option to activate blending for that module in the first place (blend = off) and therefore no further blend controls are visible. Don't forget to show the widget in all other cases. Question which comes to my mind: what happens if a user imports a history stack of a raw image onto a jpeg? Is raw denoise ignored?

> Is that not correct to do so?

Hopefully it is. I have never made a deep dive in the view logic. Probably this code is running in the gui thread, but not sure.
Anyhow, it would certainly be preferable to have blend gui related stuff handled in blend_gui.c (if possible).

jenshannoschwalm:
> Here is my idea: you hide bd->masks_modes_box

You mean like this:
```
if(!module->hide_enable_button)
{
  bd->masks_modes_box = GTK_BOX(gtk_box_new(GTK_ORIENTATION_HORIZONTAL, 0));
  //mask selection buttons packing in mask_box
  for (int i = 0; i < g_list_length(bd->masks_modes_toggles); i++)
  gtk_box_pack_start(GTK_BOX(bd->masks_modes_box), GTK_WIDGET(g_list_nth_data(bd->masks_modes_toggles, i)), TRUE, TRUE, 0);
  gtk_box_pack_start(GTK_BOX(box), GTK_WIDGET(bd->masks_modes_box), FALSE, FALSE, 0);
  dt_gui_add_help_link(GTK_WIDGET(bd->masks_modes_box), "blending.html");
  gtk_widget_set_name(GTK_WIDGET(bd->masks_modes_box), "blending-tabs");
}
```
This works fine (as my first attempt) if i switch from lighttable to darkroom mode.

Being in darkroom and using filmstrip unfortunately is not working this way.
There we seem to keep the widgets when switching to another image so the "internals" don't change.

Also: The hide_enable_button for non-raw images on raw modules also does not correctly work in filmstrip.

I certainly lack competence on how to solve these problems. Anyone take over?

upegelow:
I might not be 100% sure what effect you really want to reach but I would add

```
  if(module->hide_enable_button)
  {
    gtk_widget_hide(GTK_WIDGET(bd->masks_modes_box));
  }
  else
  {
    gtk_widget_show(GTK_WIDGET(bd->masks_modes_box));
  }
```
Close to line 1950 in `blend_gui.c:dt_iop_gui_update_blending()`

TurboGit:
I had a look (not at the code in this PR) but at the current master code and it seems that there is an issue with hide_enable_button. Indeed, just open a jpeg file and you'll see that the "raw denoise" module can still be activated even though it set hide_enable_button to TRUE.

This is to be checked with 2.6.x but I'm pretty sure the on/off button was disabled in this case.

TurboGit:
To me, just this patch in blend_gui.c should be working:

```
@@ -2067,6 +2068,9 @@ void dt_iop_gui_update_blending(dt_iop_module_t *module)
     gtk_widget_hide(GTK_WIDGET(bd->blendif_box));
   }
 
+  if(module->hide_enable_button) gtk_widget_hide(GTK_WIDGET(bd->masks_modes_box));
+
   darktable.gui->reset = reset;
 }

```

upegelow:
> This is to be checked with 2.6.x but I'm pretty sure the on/off button was disabled in this case.

It looks like the hide_enable_button status is inherited from the previous image. If this has been a jpeg and you move to a raw you can no longer enable raw denoise. There needs to be an update of that status when the image is changed.

upegelow:
>  if(module->hide_enable_button) gtk_widget_hide(GTK_WIDGET(bd->masks_modes_box));

But then what happens when we move to an image that allows raw denoise? There is no automatic mechanism that would show the widget again. At some point we need to do it.

There is another issue with using `bd->masks_modes_box`. If a user copies a rawdenoise instance of a raw image with blending activated onto a jpeg image then the raw denoise module is disabled, the masks_modes_box is hidden but all the other blending features (e.g. parametric, drawn, etc.) will be visible as they were in the raw image.

The better solution would be to show/hide the blending parent box which hosts all the subwidgets rather than `bd->masks_modes_box`. I think we do not store the widget of that box within the gui params but we should change that. It's the box that we create in line 2533 in `dt_iop_gui_init_blending()`. Maybe also the box iopw which we are handed in `dt_iop_gui_init_blending()` is a candidate but I'm not sure if we should take control of it as the blending code does not own it.

TurboGit:
In fact the on/off button was a red hearing, the following code works for me:

```
@@ -2067,6 +2070,11 @@ void dt_iop_gui_update_blending(dt_iop_module_t *module)
     gtk_widget_hide(GTK_WIDGET(bd->blendif_box));
   }
 
+  if(module->hide_enable_button)
+    gtk_widget_hide(GTK_WIDGET(bd->masks_modes_box));
+  else
+    gtk_widget_show(GTK_WIDGET(bd->masks_modes_box));
+
   darktable.gui->reset = reset;
 }
```

We can still activate the module but it is another bug.

@jenshannoschwalm : I let you test the above code and I'll see how to fix the on/off issue in a separate PR.

TurboGit:
> There is another issue with using `bd->masks_modes_box`. If a user copies a rawdenoise instance of a raw image with blending activated onto a jpeg image then the raw denoise module is disabled, the masks_modes_box is hidden but all the other blending features (e.g. parametric, drawn, etc.) will be visible as they were in the raw image.

Indeed, probably a bug we have since long time. The proper fix is to ensure that a disabled module for a certain kind of picture must be reset when loading the history. I'll try to reproduce and if so I'll open an issue.

jenshannoschwalm:
> @TurboGit The proper fix is to ensure that a disabled module for a certain kind of picture must be reset when loading the history

> @upegelow  If a user copies a rawdenoise instance of a raw image with blending activated onto a jpeg image ...

Indeed. I guess we need to add a flag to modules whether they operate raw_only. This also relates to a proper testing of image being raw, non-raw, "sraw". There is already the pending issue.

TurboGit:
And BTW, even if the on/off module is disabled it is still possible to reset the module in case of copy/paste of module from RAW to Jpeg where it doesn't make sense. So at least we have a solution.

jenshannoschwalm:
@TurboGit i see #3662, looks much better this way. I am buried at work for two days and can't do testing .. please include your leftover suggestions discussed above and close thie pr?

PR #3648, opened 2019-12-01
basic module replacement

AlicVB:
Hello,
### basic module
imho, currently the basic module is not really satisfactory : 
- The computing doesn't respect how effects should be applied to an image, so results can be weird and even not recoverable. @aurelienpierre can explain that far better than me ;)
- This module essentially consist of code duplication from other existing modules, which is weird in term of maintainability.
- if, after having used the basic module, one decide that he want to tweak is image more precisely, ex by adding masks, one need to remember the value set in basic module, reset it and set them back in the "main" module, which is far from convenient (and here, I speak only about settings that have a direct and exact copy in a "main" module)

But having a quick access to basic settings can be really good : 
- for beginners
- to dev a roughly version and then tweak it more precisely

### proposal
In order to solve all that points, my proposal, is to add a way for iops to declare some widgets that can be shown and set outside of the module, a little bit like **a shortcut with gui**. All those widgets will be regrouped to form a new basic module. This way, effects are applied at the right place in the pipe, and there's no code duplication.
plus : 
- with each "basic" widget, I've added a button to switch to the "real" module
- each time masks or multiinstances are used in the real module, the basic widget is disabled (with a special icon and a msg)
- the real module itself can decide to disable the widget (for example if some other linked params are set)
- it's up to the real module to decide how widgets should work. They can be a duplicate of an existing widget or even reflect a combination of params.

Here is a image of how it look atm (remember, it's just a POC) Capture d’écran_2019-12-01_17-47-23

### To be decided
0. do we want something like this ?
1. how to show all this : 

- in a special group (like in this PR, but with a better icon)
- in a separate lib
- in a "fake" module

2. which widgets do we want here (the one present in this PR are just here as a POC)
3. There's maybe some new iop to add (next "vibrance" ?)

Please share your thought, esp. @aurelienpierre @TurboGit @edgardoh 
(and take the actual code to what it is : just a POC)

### DIFF
--- a/data/themes/darktable.css
+++ b/data/themes/darktable.css
@@ -512,9 +512,17 @@ overshoot.right
   margin-left: 0.5em;
 }
 
+#modulegroups-basics-pop *
+{
+ padding-top: 0.2em;
+ margin-top : 0;
+}
+
 #modulgroups-basics-pop-title *
 {
   padding-left: 2em;
+  font-weight: bold;
+  color: @plugin_label_color;
 }
 
 /* icon buttons in modules main body */
--- a/data/themes/darktable.css
+++ b/data/themes/darktable.css
@@ -1437,6 +1437,7 @@ margin: 0 2px;
 #basics-box-labels #basics-widget,
 #basics-box-labels #section_label
 {
+  padding-top: 2px;
   margin: 0 0.75em 0.25em 0.75em;
 }
 
--- a/data/themes/darktable.css
+++ b/data/themes/darktable.css
@@ -1419,10 +1419,12 @@ margin: 0 2px;
   padding-bottom: 0.75em;
 }
 
-#basics-box #module-enable-button
+#basics-box #module-enable-button,
+#basics-link
 {
   min-height: 1em;
   min-width: 1em;
+  border: none;
 }
 
 #basics-widget
--- a/data/themes/darktable.css
+++ b/data/themes/darktable.css
@@ -471,7 +471,8 @@ overshoot.right
 /*** NOTE: bauhaus controls inherit their font properties from there ***/
 #lib-plugin-ui,
 #iop-plugin-ui,
-#basics-box
+#basics-box,
+#basics-box-labels
 {
   color: @plugin_fg_color;
   background-color: @plugin_bg_color;
@@ -1413,13 +1414,19 @@ margin: 0 2px;
   padding: 0.45em;
 }
 
-#basics-box
+#basics-box,
+#basics-box-labels
 {
-  padding-top: 0.5em;
   padding-bottom: 0.75em;
 }
 
+#basics-box
+{
+  padding-top: 0.5em;
+}
+
 #basics-box #module-enable-button,
+#basics-box-labels #module-enable-button,
 #basics-link
 {
   min-height: 1em;
@@ -1427,15 +1434,26 @@ margin: 0 2px;
   border: none;
 }
 
-#basics-widget
+#basics-box-labels #basics-widget,
+#basics-box-labels #section_label
+{
+  margin: 0 0.75em 0.25em 0.75em;
+}
+
+#basics-box #basics-widget
 {
   margin: 0 0.75em 0.5em 0.75em;
+}
+
+#basics-box-labels #section_label
+{
+  margin-top: 1em;
 }
 
 #basics-widget.basics-widget_group_start
 {
   padding-top: 0.5em;
-  border-top: 0.5px solid  @section_label;
+  border-top: 0.5px solid @section_label;
 }
 
 #basics-iop_name
--- a/data/themes/darktable.css
+++ b/data/themes/darktable.css
@@ -484,47 +484,6 @@ overshoot.right
   padding: 0.75em;
 }
 
-#basics-box
-{
-  padding-top: 0.5em;
-  padding-bottom: 0.75em;
-}
-
-#basics-widget
-{
-  margin: 0 0.75em 0.5em 0.75em;
-}
-
-#basics-widget.basics-widget_group_start
-{
-  padding-top: 0.5em;
-  border-top: 0.5px solid  @section_label;
-}
-
-#basics-iop_name
-{
-  padding-left: 0.25em;
-}
-
-#basics-widget #bauhaus-combobox + #dt-button,
-#basics-widget #bauhaus-slider + #dt-button
-{
-  margin-left: 0.5em;
-}
-
-#modulegroups-basics-pop *
-{
- padding-top: 0.2em;
- margin-top : 0;
-}
-
-#modulgroups-basics-pop-title *
-{
-  padding-left: 2em;
-  font-weight: bold;
-  color: @plugin_label_color;
-}
-
 /* icon buttons in modules main body */
 #iop-plugin-ui-main #dt-button,
 #iop-plugin-ui-main #dt-toggle-button,
@@ -939,8 +898,7 @@ popover,
 combobox window,
 dialog combobox window,
 #bauhaus-combobox *,
-#bauhaus-slider *,
-#modulegroups-iop-popup
+#bauhaus-slider *
 {
   opacity: 1;  /* this is needed for tooltip on/off feature with Shift+t shortcut */
   background-color: @tooltip_bg_color;
@@ -950,11 +908,6 @@ dialog combobox window,
   min-height: 0.4em;
   min-width: 0.4em;
   padding: 0.2em 0;
-}
-
-#modulegroups-iop-popup viewport
-{
-  background-color: @tooltip_bg_color;
 }
 
 popover
@@ -1371,22 +1324,9 @@ margin: 0 2px;
 
 #modulegroups-header #dt-button,
 #modulegroups-preset #dt-button,
-#modulegroups-preset-activated #dt-button,
-#modulegroups-iop-popup-name,
-#modulegroups-iop-popup-cb
+#modulegroups-preset-activated #dt-button
 {
   border: 0;
-}
-
-#modulegroups_iop_title,
-#modulegroups-iop-popup-name
-{
-   padding: 0.2em 0.5em;
-}
-
-#modulegroups-iop-popup-cb
-{
-  margin-left: 2em;
 }
 
 #modulegroups-groups-title
@@ -1472,6 +1412,48 @@ margin: 0 2px;
 {
   background-color: #852A2A;
   padding: 0.45em;
+}
+
+#basics-box
+{
+  padding-top: 0.5em;
+  padding-bottom: 0.75em;
+}
+
+#basics-widget
+{
+  margin: 0 0.75em 0.5em 0.75em;
+}
+
+#basics-widget.basics-widget_group_start
+{
+  padding-top: 0.5em;
+  border-top: 0.5px solid  @section_label;
+}
+
+#basics-iop_name
+{
+  padding-left: 0.25em;
+}
+
+#basics-widget #bauhaus-combobox + #dt-button,
+#basics-widget #bauhaus-slider + #dt-button
+{
+  margin-left: 0.5em;
+}
+
+#modulegroups-popup *
+{
+ padding-top: 0.1em;
+ padding-bottom: 0.1em;
+}
+
+#modulegroups-popup-title *
+{
+  padding-left: 2em;
+  padding-bottom: 0.2em;
+  font-weight: bold;
+  color: @plugin_label_color;
 }
 
 /*---------------------------------------------------------------
--- a/data/themes/darktable.css
+++ b/data/themes/darktable.css
@@ -479,8 +479,7 @@ overshoot.right
 
 #iop-plugin-ui-main,
 #lib-plugin-ui-main,
-#blending-box,
-#basics-widget
+#blending-box
 {
   padding: 0.75em;
 }
@@ -493,13 +492,24 @@ overshoot.right
 
 #basics-widget
 {
-  padding-top: 0.5em;
-  padding-bottom: 0;
+  padding: 0.5em 0 0 0;
+  margin: 0 0.75em 0.5em 0.75em;
+}
+
+.basics-widget_group_start
+{
+  border-top: 0.5px solid  @section_label;
 }
 
 #basics-iop_name
 {
-  padding-right: 0.25em;
+  padding-left: 0.25em;
+}
+
+#basics-widget #bauhaus-combobox + #dt-button,
+#basics-widget #bauhaus-slider + #dt-button
+{
+  margin-left: 0.5em;
 }
 
 #modulgroups-basics-pop-title *
--- a/data/themes/darktable.css
+++ b/data/themes/darktable.css
@@ -580,7 +580,6 @@ overshoot.right
 #module-reset-button,
 #module-enable-button,
 #module-instance-button,
-#module-enable-button,
 #module-collapse-button,
 #module-always-enabled-button,
 #module-always-disabled-button
@@ -1418,6 +1417,12 @@ margin: 0 2px;
 {
   padding-top: 0.5em;
   padding-bottom: 0.75em;
+}
+
+#basics-box #module-enable-button
+{
+  min-height: 1em;
+  min-width: 1em;
 }
 
 #basics-widget
--- a/data/themes/darktable.css
+++ b/data/themes/darktable.css
@@ -470,7 +470,8 @@ overshoot.right
 /* Frame around modules boxes */
 /*** NOTE: bauhaus controls inherit their font properties from there ***/
 #lib-plugin-ui,
-#iop-plugin-ui
+#iop-plugin-ui,
+#basics-box
 {
   color: @plugin_fg_color;
   background-color: @plugin_bg_color;
@@ -478,9 +479,32 @@ overshoot.right
 
 #iop-plugin-ui-main,
 #lib-plugin-ui-main,
-#blending-box
+#blending-box,
+#basics-widget
 {
   padding: 0.75em;
+}
+
+#basics-box
+{
+  padding-top: 0.5em;
+  padding-bottom: 0.75em;
+}
+
+#basics-widget
+{
+  padding-top: 0.5em;
+  padding-bottom: 0;
+}
+
+#basics-iop_name
+{
+  padding-right: 0.25em;
+}
+
+#modulgroups-basics-pop-title *
+{
+  padding-left: 2em;
 }
 
 /* icon buttons in modules main body */
@@ -492,7 +516,8 @@ overshoot.right
 #blending-box #dt-button,
 #blending-box #dt-toggle-button,
 #blending-box #show_mask_button,
-#blending-box #keep-active
+#blending-box #keep-active,
+#basics-widget #dt-button
 {
   min-height: 1.3em;
   min-width: 1.3em;
--- a/data/themes/darktable.css
+++ b/data/themes/darktable.css
@@ -492,12 +492,12 @@ overshoot.right
 
 #basics-widget
 {
-  padding: 0.5em 0 0 0;
   margin: 0 0.75em 0.5em 0.75em;
 }
 
-.basics-widget_group_start
+#basics-widget.basics-widget_group_start
 {
+  padding-top: 0.5em;
   border-top: 0.5px solid  @section_label;
 }
 
--- a/src/libs/modulegroups.c
+++ b/src/libs/modulegroups.c
@@ -1920,7 +1920,7 @@ static void _manage_direct_save(dt_lib_module_t *self)
     dt_lib_presets_apply((gchar *)C_("modulegroup", FALLBACK_PRESET_NAME), self->plugin_name, self->version());
 }
 
-static void _manage_direct_module_toggle(GtkWidget *widget, GdkEventButton *event, dt_lib_module_t *self)
+static void _manage_direct_module_toggle(GtkWidget *widget, dt_lib_module_t *self)
 {
   gchar *module = (gchar *)g_object_get_data(G_OBJECT(widget), "module_op");
   dt_lib_modulegroups_group_t *gr = (dt_lib_modulegroups_group_t *)g_object_get_data(G_OBJECT(widget), "group");
@@ -1936,8 +1936,6 @@ static void _manage_direct_module_toggle(GtkWidget *widget, GdkEventButton *even
     gr->modules = g_list_delete_link(gr->modules, found_item);
   }
 
-  GtkWidget *pop = (GtkWidget *)g_object_get_data(G_OBJECT(widget), "popup");
-  gtk_widget_destroy(pop);
   _manage_direct_save(self);
 }
 
@@ -2003,7 +2001,7 @@ static void _manage_editor_basics_add(GtkWidget *widget, dt_lib_module_t *self)
   }
 }
 
-static void _manage_editor_module_add(GtkWidget *widget, GdkEventButton *event, dt_lib_module_t *self)
+static void _manage_editor_module_add(GtkWidget *widget, dt_lib_module_t *self)
 {
   gchar *module = (gchar *)g_object_get_data(G_OBJECT(widget), "module_op");
   dt_lib_modulegroups_group_t *gr = (dt_lib_modulegroups_group_t *)g_object_get_data(G_OBJECT(widget), "group");
@@ -2013,8 +2011,6 @@ static void _manage_editor_module_add(GtkWidget *widget, GdkEventButton *event, 
   {
     gr->modules = g_list_append(gr->modules, g_strdup(module));
     _manage_editor_module_update_list(self, gr);
-    GtkWidget *pop = (GtkWidget *)g_object_get_data(G_OBJECT(widget), "popup");
-    gtk_widget_destroy(pop);
   }
 }
 
@@ -2053,19 +2049,12 @@ static int _manage_editor_module_so_add_sort(gconstpointer a, gconstpointer b)
 static void _manage_module_add_popup(GtkWidget *widget, dt_lib_modulegroups_group_t *gr, GCallback callback,
                                      gpointer data, gboolean toggle)
 {
-  GtkWidget *pop = gtk_popover_new(widget);
-  gtk_widget_set_name(pop, "modulegroups-iop-popup");
-  GtkWidget *sw = gtk_scrolled_window_new(NULL, NULL);
-  gtk_scrolled_window_set_policy(GTK_SCROLLED_WINDOW(sw), GTK_POLICY_NEVER, GTK_POLICY_AUTOMATIC);
-  GtkWidget *vb = gtk_box_new(GTK_ORIENTATION_VERTICAL, 0);
-  GtkWidget *vb0 = gtk_box_new(GTK_ORIENTATION_VERTICAL, 0);
-  GtkWidget *vb1 = gtk_box_new(GTK_ORIENTATION_VERTICAL, 0);
-  GtkWidget *vb2 = gtk_box_new(GTK_ORIENTATION_VERTICAL, 0);
-  GtkWidget *lb = NULL;
+  GtkWidget *pop = gtk_menu_new();
+  gtk_widget_set_name(pop, "modulegroups-popup");
 
-  int rec_nb = 0;
-  int other_nb = 0;
-  int in_nb = 0;
+  int nbr = 0; // nb of recommended items
+  int nba = 0; // nb of already present items
+  int nbo = 0; // nb of other items
 
   GList *m2 = g_list_copy(g_list_first(darktable.iop));
   GList *modules = g_list_sort(m2, _manage_editor_module_so_add_sort);
@@ -2077,18 +2066,13 @@ static void _manage_module_add_popup(GtkWidget *widget, dt_lib_modulegroups_grou
     {
       if(!g_list_find_custom(gr->modules, module->op, _iop_compare))
       {
-        GtkWidget *bt;
-        bt = gtk_button_new_with_label(module->name());
-        gtk_widget_set_name(bt, "modulegroups-iop-popup-name");
-        gtk_widget_set_tooltip_text(bt, _("click to add this module"));
-        g_object_set_data(G_OBJECT(bt), "module_op", module->op);
-        g_object_set_data(G_OBJECT(bt), "group", gr);
-        g_object_set_data(G_OBJECT(bt), "popup", pop);
-        gtk_label_set_xalign(GTK_LABEL(gtk_bin_get_child(GTK_BIN(bt))), 1.0);
-        g_signal_connect(G_OBJECT(bt), "button-press-event", callback, data);
+        GtkMenuItem *smi = (GtkMenuItem *)gtk_menu_item_new_with_label(module->name());
+        gtk_widget_set_tooltip_text(GTK_WIDGET(smi), _("click to add this module"));
+        g_object_set_data(G_OBJECT(smi), "module_op", module->op);
+        g_object_set_data(G_OBJECT(smi), "group", gr);
+        g_signal_connect(G_OBJECT(smi), "activate", callback, data);
 
         // does it belong to recommended modules ?
-        GtkWidget *vbc = vb2;
         if(((module->default_group() & IOP_GROUP_BASIC) && g_strcmp0(gr->name, _("base")) == 0)
            || ((module->default_group() & IOP_GROUP_COLOR) && g_strcmp0(gr->name, _("color")) == 0)
            || ((module->default_group() & IOP_GROUP_CORRECT) && g_strcmp0(gr->name, _("correct")) == 0)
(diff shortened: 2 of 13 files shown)


aurelienpierre:
0. YES !
1. in a special group (like in this PR, but with a better icon)
2. white balance, exposure, contrast, saturation, vibrance, "lightness" (grey point of filmic)
3. vibrance is something I plan for colour balance, but I need to rewrite it for the new RGB API.

TurboGit:
So we are at it!

I'm not tempted to remove the basic module that will come for the first time in dt 3.0 !

This is really a nice step forward AlicVB!

To answer:

0 : Yes ! Yes ! Yes !
1 : Special group or fake module.
2 : ultimatly, the widget in there should be selectable in preference for each users to be able to select the widgets to list.

TurboGit:
Some questions to add too:
- Can we have presets for values in this new basic module?
- Can we have a global reset for the basic module?
- Can we have a on/off support for the basic module?

If yes, a fake module is probably the way to go.

AlicVB:
> 2 : ultimatly, the widget in there should be selectable in preference for each users to be able to select the widgets to list.

yes, I've thought about a menu with checkboxes like Libreoffice...

> Some questions to add too[..]

to answer the above questions, I would say no to all : in my mind a basic module should stay really basic, anything a little more complex should be handled by the "main" module.
And anyway, this would be tricky to implement, as widgets in basic module are only a subset of the main module gui...

TurboGit:
@AlicVB : if the answer is not to all, which I can understand then it is better to put those controls into a separate group.

TurboGit:
BTW, why is it disabled when there is mask in the main module?

TurboGit:
The exposure  key or moving the histogram works even if exposure module has masks.

TurboGit:
For multi-instance I agree, it should be disabled.

AlicVB:
because at first, I wanted that any complexity added to the main module disable the basic widget (I've even started with an implementation where it was disabled as soon as another param in the main module is changed)
But it would have artificially limited the use of this to beginner, although it can be convenient for other too.
So for masks, why not... I don't see any blocking to let them enabled...

AlicVB:
@aurelienpierre : 
- for the "lightness", I'm not sure about adding the filmic one, as enabling filmic change lot of other things in the image and returning to default value isn't "neutral" (not sure to be clear...)
- for the contrast, I usually set it via filmic, but I would prefer to use another module for the reasons mentioned above. Do you think the colorbalance is good for this ?

Also, I'm not sure about that ones : 
- exposure black (do we want to expose here a often misused widget ?)
- colorbalance tint

What I plan to add : 
- rotation
- black&white conversion combobox (on/off) (with which module ? monochrome ?)

Of course everything is debatable :)

GLLM:
Hi all 
Just thinking about what I'd use if I had this toolbox menu ... and I thought about _Local contrast_ !
Would that make any sense ? and w.r.t. the pixel pipeline ?
Thanks

TurboGit:
Some responses:

I agree that Filmic should not be used there. Only module with neutral settings should be considered otherwise it will certainly confuse a lot the beginners.

I won't add black level. This is really not a beginner setting.

Rotation seems reasonable.

I'm not for B&W, there is no one button to get a nice black and white. Most of the time without fine tuning it the B&W is just flat.

For contrast, using the colorbalance one will certainly be frightening for beginners too If one need to go to fine tune something. The module "contrat, luminosity, saturation" is better for beginners but at the same time I'm not sure it is wise to advocate it. So maybe colorbalance is the only possible choice. I'd like to hear Aurélien on this one.

kbarni:
Good idea!

First, I agree that (the current) Basic Adjustments module might be simplified: only one instance, no masks. To become really a *basic* module. Anyway, gathering the most important sliders to a new tab would really be a better solution!

By the way, what about adding the **Styles** module from the *lighttable* to this tab?
The result of applying a style on a thumbnail (in lighttable) is difficult to see and going to darkroom/lighttable to set a new style isn't very intuitive...

chhil:
> By the way, what about adding the Styles module from the lighttable to this tab?
Arent styles already available in darkroom bottom left above the filmstrip?
image

GLLM:
> Good idea!
> 
> First, I agree that (the current) Basic Adjustments module might be simplified: only one instance, no masks. To become really a _basic_ module. Anyway, gathering the most important sliders to a new tab would really be a better solution!
> 
> By the way, what about adding the **Styles** module from the _lighttable_ to this tab?
> The result of applying a style on a thumbnail (in lighttable) is difficult to see and going to darkroom/lighttable to set a new style isn't very intuitive...

I use the _Basic adjustments_ module with masks most of the times ... 
I beg to differ on this  : I like it with multiple instances & masks !

kbarni:
What about making a brainstorming on this question on the pixls.us forum? This is a big reorganization of the interface and we could have some feedback by the darktable users.

TurboGit:
I'm not against some kind of brainstorming but don't call that a "big reorganization of the interface". It is just another tag to have quick access to simple sliders for beginners. We don't have even decided if the current basic adjustment module should be deprecated or not.

AlicVB:
> I use the _Basic adjustments_ module with masks most of the times ...
> I beg to differ on this : I like it with multiple instances & masks !

And your are not alone, I fear. One of the main problem, as @aurelienpierre has already said many times in the recent past (cf here for example : https://github.com/darktable-org/darktable/pull/3121#issuecomment-552181283), is that the basic module doesn't respect signal processing theory by putting at the same place in the pipe different operation that shouldn't be together...
Not to speak about code maintenance...

That said, I agree with @TurboGit : it's not a big ui change ! (and I certainly don't want to deal with a "big" dt's ui change !)

AlicVB:
Now, to speak about which controls should or shouldn't be in my proposal : 
- I'll implement a way for the user to hide/show widgets to their needs. So we can afford to have a "not to small" list of controls available here. We'll "just" have to decide which ones are shown by default (very few of them)
- So I'm not against any "simple" and "basic" widgets to be there, as soon as their "main" module is not a problematic one (too complex, too delicate to set for beginners, with bad side effects, ...). We should avoid duplicate controls too.
- Even if I see the usefulness of it, I certainly don't want to add masks, multi-instances or reordering here. It's out of the scope of this. The right places to do such finetuning are in the "main" modules (and anyway, I don't think it's possible in the code)

GLLM:
> And your are not alone, I fear. One of the main problem, as @aurelienpierre has already said many times in the recent past (cf here for example : [#3121 (comment)](https://github.com/darktable-org/darktable/pull/3121#issuecomment-552181283)), is that the basic module doesn't respect signal processing theory by putting at the same place in the pipe different operation that shouldn't be together...
> Not to speak about code maintenance...

* The code maintenance side, believe me, I completely get !
* Not being alone, indeed ... alas
* Not using the proper module to retouch : probably, and knowingly ... 

I've watched Filmic/DT videos from Aurelien, and I get the pipeline thing, the linear vs. non-linear argument, and I actually try to use FilmicRGB as much as possible. 
But often, it's just too long to get to the result *I* expect (that I more or less manage to get, faster & more easily with basic adj)...
I often have 2 masks : one for highlights, that I may pull down, and one for shadows, that I may push up ... 
Apart from setting a correct black & white level in Basic adj (ensure no clipping), I manage to quickly do it with the rest of the sliders in Basic adj, without getting headaches ... which unfortunately still cant do with FilmicRGB.
I get the signal theory, the math & pipeline order ... yet I do have to use 7-8 slides in Filmic when I sometimes only need 4-5 in Basic Adj :'-(

AlicVB:
OK, I've added the menu to control widget visibility and rework some parts, so it can be tested.
(don't look at the git log mess, I'll rebase everything at the end)

aurelienpierre:
> I'm not for B&W, there is no one button to get a nice black and white. Most of the time without fine tuning it the B&W is just flat.

WB can also be achieved simply by setting saturation = 0, using the luminance.

> I agree that Filmic should not be used there. Only module with neutral settings should be considered otherwise it will certainly confuse a lot the beginners.

Fair enough. But then, we still might need some lightness adjustment. How do we do that ? Colour balance power ?

> I get the signal theory, the math & pipeline order ... yet I do have to use 7-8 slides in Filmic when I sometimes only need 4-5 in Basic Adj :'-(

Basic adjustments was never about tone mapping. You are comparing eggs to apples.

> What about making a brainstorming on this question on the pixls.us forum? This is a big reorganization of the interface and we could have some feedback by the darktable users.

Brainstorming about topics like this one may result in a lot of input from people who don't really understand the issues. Users are not designers. Let users state their problems and let designers craft a solution. Mixing both can backfire real fast.

> For contrast, using the colorbalance one will certainly be frightening for beginners too If one need to go to fine tune something. The module "contrat, luminosity, saturation" is better for beginners but at the same time I'm not sure it is wise to advocate it. So maybe colorbalance is the only possible choice. I'd like to hear Aurélien on this one.

Color blance fulcrum contrast is the one in basic adjustments already, and the algo recommended by the ACES. It works in RGB and is compatible with HDR data. "contrat, luminosity, saturation"  works in Lab and… you know the issues of Lab.

aurelienpierre:
One more things here @AlicVB : I plan to deprecate color balance soon to rewrite it for the RGB API (and let users chose the RGB working space).

AlicVB:
> One more things here @AlicVB : I plan to deprecate color balance soon to rewrite it for the RGB API (and let users chose the RGB working space).

Great !
Question is : do we want to advertise for "soon-deprecated" iops ? Or maybe we can postpone this until your new iop is ready (I don't really care about timings)

aurelienpierre:
> do we want to advertise for "soon-deprecated" iops ?

Sounds like more work. You can still draft the feature with the old colour balance, but I would like to have the new one merged before this PR, if possible, so we won't need to deal with deprecation and stuff.

GLLM:
> > I get the signal theory, the math & pipeline order ... yet I do have to use 7-8 slides in Filmic when I sometimes only need 4-5 in Basic Adj :'-(
> 
> Basic adjustments was never about tone mapping. You are comparing eggs to apples.
@aurelienpierre  : I fear being completely off topic.

But isn't all that matter : getting a picture without any black or white clipping, without color clipping as well, a suiting WB, and nice looking contrast/colors ... so that in the end the result suits the person doing the PP ?
Candid/true question : is using the basic adj with masks (for HL, shadows, and midtones if necessary) a kind of tone mapping ?

AlicVB:
> but I would like to have the new one merged before this PR, if possible, so we won't need to deal with deprecation and stuff.

This feature will not have to deal with deprecation :). Once a module is deprecated, its controls will not be shown anymore in the basic module (even if there's still some settings for the "main" module).
What I want to avoid is that too many people, especially beginners, use color balance whereas we already know it will deprecated soon.
So I've no objection to wait for your new module before merging this.

aurelienpierre:
> But isn't all that matter : getting a picture without any black or white clipping, without color clipping as well, a suiting WB, and nice looking contrast/colors ... so that in the end the result suits the person doing the PP ?

And then, what if the results don't suit the person ? I fear we are working under the assumption that every user has the same taste and the same needs, so we can pre-bake him an median editing path. But that's not true ! Commercial and fine-art photographers are not your average daddy taking pics of the dog and the kids. So the matter is to give them a reliable toolset they can adapt. 

> Candid/true question : is using the basic adj with masks (for HL, shadows, and midtones if necessary) a kind of tone mapping ?

Tone mapping (same as gamut mapping) implies you have 2 media (input and output) which have a different dynamic ranges (or gamut), and you know their properties beforehand, so you can rescale the input/output ranges doing educated choices (for example: the intents in gamut mapping). It's a technical operation that will need to be adjusted depending on the output medium. 

Tones corrections are artistic choices that don't depend on the output. 

Think of your picture as if it was a movie. The movie raw footages are called the masters, and then are edited in an output-agnosic way. After editing, the movies are exported and encoded for DVD and movie theaters from the master (2 different media that need different adjustments). Say a new TV comes out with HDR capabilities and Rec2020 colour space, it's just a matter of exporting again. But if you put your gamut and tone mapping in the master, you can't do that.

Unfortunately, this kind of distinction master/output does not exist in photography, since we are constantly assuming output is paper, so even screens are calibrated to match paper, and we can mix every step in the pipeline with no visual difference on the result. But the day they invent some HDR printer, you will have to throw you edits and start again from scratch if you didn't separate master and output.

GLLM:
Thank you very much @aurelienpierre for your thorough & constructive reply !
I'll read it later a few more times to make sure I really get the full/true meaning of it & its implications.
I'll stop polluting this PR, starting now :)

todd-prior:
Along these lines and if there is this much work going into a future version would it be a consideration to allow those iops with auto setting potential to have those be a default. for example with exposure or filmic or a new version of the basic module that is changing contrast etc....that way when used in styles they would adapt to each image. Spot white balance when first applied uses the whole image and often does a nice job correcting so this is also like an auto setting......It would be something like adding a check box in the iop dialog ...the iop would use whatever defaults or entered values if unchecked and if checked would activate the auto setting when applied

AlicVB:
Still WIP until @aurelienpierre new IOP is ready (no urgency here...)

AlicVB:
This is my next working area.
@aurelienpierre : any news about your iop ? just to know. again, no urgency !

aurelienpierre:
I'm swamped with #4800 and #4940 :-)

TurboGit:
@AlicVB : this was scheduled for 3.4, I suppose it won't make it so I'm postponing it for some later minor release.

AlicVB:
yes, not for 3.4, I'm little bit short in times this days...
and that will let us some more time with @aurelienpierre to decide the way to go with new iops (or not)

dterrahe:
Is the intention for each module that would want to support this to have to create duplicate widgets itself and add a lot of boiler plate? For me the more logical approach would be to slightly extend the framework in #6648 and have a "basic" module that creates a separate widgets by copying all the parameters (labels, soft/hard limits) from basically any bauhaus slider/combobox that has been registered using dt_bauhaus_widget_set_label. Any move to such a new widget would then be passed through to the existing widget, which would then respond as before, without the need for separate implementations. The copied widget would keep in sync by registering to the value-changed signal of the original.

I'm sorry that I didn't notice that this PR had progressed in the background. I'm a bit sad that the approach here seems to be to add back a lot of the boilerplate that I just spend a lot of time removing so that we could have an open, flexible framework that doesn't need work in each and every module.

AlicVB:
@dterrahe : I'm not fully sure of what you mean exactly by boilerplate here, but just to precise how I see this atm : 
- The list of available widgets to display should stay really small, in order to answer the goal of this PR. Maybe 10-12 widget at maximum, with 6-8 shown by default. This implies that not all widget should be available here (contrary to shortcuts) and this implies too that each module should be responsible to provide some widget to show (or not).
- Module can eventually propose a widget which is not present in its main gui : for example, channelmixer can present a simple combobox "color-B&W" which is relied to some well defined settings.
- apart for this special case, widgets for "basic editing" is a duplication of an existing one. And I admit that in current implementation I've done it weird (I've duplicated all the creation/events steps. What could be great is to have a way to get a completely synchronized widget from an existing one, but I've not found an easy way to do it (especially to reflect each changes of the original to the duplicate, even if the change doesn't comes from user input)

I hope this answer a little bit your concerns... Thanks for your feedback

dterrahe:
What tends to happen is that features get implemented with a very limited initial scope and that the same approach is then continuously expanded until it is applied all across the code base. At that point someone might come in to refactor and implement a more general solution, but they spend most of their time cleaning up the existing implementations, which ended up being special cased all over the place because, hey, flexibility. But when the correct solution is completed, you start getting features for free in all new uses. For example all modules are supported by history diff because they use introspection for their fields. If I had wanted a quick solution I would just have built something in exposure (and maybe some other "important" modules) so that you could easily see if the exposure was adjusted. And the general approach would never have seen the light.

I understand your goal is to have a limited set of widgets for new users. But that is just a special case of a random collection of widgets that for power users should be configurable the way they want. Your approach will not give that possibility because for each widget extra work has to be done to support it. Just like people were debating whether certain sliders needed shortcuts. Now they all have them, without effort (apart from more scrolling in the settings dialog, OK, I admit that).

I'm working on a proof of concept implementation to demonstrate my approach, even though not all the underlying infrastructure is there yet.

dterrahe:
To answer some of your points more specifically:
> * This implies that not all widget should be available here.

Why "should"? Are you not talking about sensible defaults and maybe _favorates_ or _popular_ widgets in the configuration screen? Why can't I have any of them if I want?

> * Module can eventually propose a widget which is not present in its main gui : for example, channelmixer can present a simple combobox "color-B&W" which is relied to some well defined settings.

Sure; any hidden widgets would be available too.

> * What could be great is to have a way to get a completely synchronized widget from an existing one.

This should not be too hard to achieve with either the standard value-changed signal (which might not work at gui_update) or a new synchronise signal, only to be implemented within bauhaus, so transparent to any modules that use the bauhaus widgets. Gradients etc could be supported the same way, if needed.

AlicVB:
You're right of course.
Now, if we go this line (propose all available widgets) we have to find a way for : 
- having a nice interface to choose which widget to show (I can do that as an extension of the modulegroups manager)
- the content of this basic module can be store inside modulegroups presets, I've anticipated that.
- select carefully the set of widget to show in order to stay "basic" <= this is the main goal here

That way, I think we can satisfied everybody : basic users and power users.

Can I let you have a look at the "duplication" part ? (you know that part way better than me) I can do it too if you can't
Ideally, I would need a way to : 
- identify by an unique identifier widgets (something like "module_name|widget_label" we will have to deal with eventual duplicate names too)
- get a duplicated gtkwidget from the identifier

Thanks !

dterrahe:
POC implementation here https://github.com/darktable-org/darktable/commit/05df7f62caf172cb4cd355c12c43d120bb3eb449

This bastardises one of the iop modules without gui, _scale pixels_, to show copies of a hardcoded list of widgets from other modules. Just to show how it could work technically; this is of course stupid, but otherwise I'd have the add a whole new file to makefiles etc just to show the concept. You'll probably not want to put it in a module anyway, but in a separate "group". Although... a fake module at the top could also work...

Anyway, to see in action; switch to darkroom, open _scale pixels_. It will be empty, because of initialisation order issues. You need to switch to another image first; the module then gets populated in reload_defaults. (don't keep switching images; I'm not doing any cleanup or anything here; the list will just grow).

The hardcoded list are just "dynamic" identifiers from the keyboardrc file. Only sliders for the demo, but supporting combos would be just as easy; simply copy over the list contents. Point is; I have so far cleaned up/restructured setting up widgets so that they all cleanly call dt_bauhaus_widget_set_label and get registered there, but the only _backend_ to use it, for now, is keyboardaccelerators (and vimkeys). Next step is to decide how/with what goals to maintain the list of widgets/accelerators/commands so that all anticipated use cases can be served. This is the discussion in #5063. For now I'm hacking the accelerator list closures to get at the widgets.

Another approach, if you don't put this in a lists with other modules, so you never see the "same" widget in the original and in the "basic" list at the same time, is to temporarily reparent the original widget. Assuming we only need to deal with boxes and possibly grids, this would just require saving the child properties in order to put them back correctly after. Expect problems to solve with widgets that get dynamically (un)hidden. This approach could in theory also be used for graph widgets

AlicVB:
@dterrahe : great ! I've done a quick test and indeed, this works well (just a typo : you need to decrease `darktable.gui->reset` at the end of `sync_change`)
If we go this way, we may also want to have a switch for module to exclude some of its widget from this (ex : widget which are not working alone, in which the value doesn't means anything without other ones, etc...)

I'll give a try to your "reparent" proposal as it would avoid even more boilerplate (and potentially allow to use non bauhaus widget too)

AlicVB:
@dterrahe : 
So here is my POC attempt for your reparent idea : https://github.com/AlicVB/darktable/tree/basic2
This seems to work... but there's some things to sort out : 
1. reparenting can be challenging, so we'll have to deal with different containers (nothing difficult, just lot of special cases)
2. I'm not sure how the closures things works regarding initialization, but if I understand well, the list of closure is only populated once each module has connected its accels. So we'll need either a signal once every connection is done either another way to retrieve widgets... One solution could be to keep a glist globally with each widget declared in dt_bauhaus_widget_set_label that would have the advantage to be disconnected from accels and thus allow more flexibility (graph widgets, etc...)
3. this is just a POC. list of basic widgets is hardcoded, image switch can break, lot of stuff to do...

If you find time to give your opinion, that would be great. Thanks in advance !

dterrahe:
> 1. reparenting can be challenging, so we'll have to deal with different containers (nothing difficult, just lot of special cases)

Boxes and Grids are what is used almost everywhere. Maybe here and there a widget is directly put into a Notebook (and Notebook are a hassle to deal with; they have some inconsistent behavior), but I would suggest dealing with that by wrapping the widget in a box first, rather than trying to make this code work with notebooks.

> 2. I'm not sure how the closures things works regarding initialization, but if I understand well, the list of closure is only populated once each module has connected its accels. So we'll need either a signal once every connection is done either another way to retrieve widgets... One solution could be to keep a glist globally with each widget declared in dt_bauhaus_widget_set_label that would have the advantage to be disconnected from accels and thus allow more flexibility (graph widgets, etc...)

There is some trickiness involving multi-instance, so that actual closures only get linked when the desired ordering between instances of the same module is known (based on the preferences). This is why we need to settle on a structure first that will work across the different use cases. @houz was right to reject my midi patch which just piggy backed on an existing structure; we shouldn't end up here with a special case for just basic modules again. It would limit what we can do in future. #5063.

Currently the accel list is based around the individual keys that can be pressed and the same slider widget gets listed 5 times (for decr/incr/dyn...). The better approach would be to maintain a list of manipulatable widgets and their types and for each type have a list of actions. So then for example the tone curve would be in the list and actions, I don't know, could be moving up/down up to 8 points on the graph, whatever. But now _you_ for this purpose could pick up the widget directly from that list and reparent it. Someone else could exposure it via lua.

And again, multi-instance needs to be integrated in the approach, even if _you_, _here_ maybe want to ignore it (and always use the _base_ widget? Is that what you'd want? Could be less powerful.)

> 3. this is just a POC. list of basic widgets is hardcoded, image switch can break, lot of stuff to do...

Looking at the code only; yes, this is the kind of thing I was thinking of.

To be honest though, I'm not sure if going with a separate tab for "basic" widgets is the right way. It means you can't have _that_ *and* another widget visible at the same time if you want to work on them together. I'm thinking more of a "pinning" approach, where a collection of widgets is always visible at the top of the module list and the other modules are below that, as now. But that would mean both the original and the pinned version could be visible at the same time, so stealing by reparenting won't work.

I guess we could have both (since your approach allows graphs and the simple phantom widget approach never will).

AlicVB:
thanks for your answer.
I'll try to go further in this direction (reparent and global list) in the next weeks.

dterrahe:
Great!

We should end up with one global list of widgets/actions that is suitable for all purposes (#5063). One of the next steps would be integrating vimkeys as well. And making it accessible via lua/dbus.

AlicVB:
OK, so I've rewritten all from scratch and force pushed.
This was needed due to the internal changes and to the amount of work of an eventual rebasing...

Visually no real change, but lot of internal changes : 
- a list of individual widget, corresponding to module widgets + a link to orginal module (same as before)
- the list is in a special group with a new icon (if someone comes with a better one, no problem)
- the visibility of the group and its content can be setup using module layout manager, in the same way as the other groups
- actually all module widgets that are sliders, combobox, on/off button, can be added to the list (big change from previous version)
- there's a list of "recommended widget" with appears in first when you want to add a new widget, and which are the ones added to default presets
- no widget can be shown if its "real" module is not present in one of the other groups

Internally, after @dterrahe comments, we have a list of all available widgets : dt_develop_t -> iop_widgets which stay in sync with current pipeline. This list is designed to be evolutive in order to be used for other purposes...

The main points that need review discuss : 
1. currently, if there's multiple module instances, the widget is disabled. This is to stay "basics", but codewise, it would be easy (simpler in fact) to show each instance widget. It will be useful for power users and "beginners" who plays with multiple instances are not really beginners...
2. the default/recommended list of widgets. Currently : exposure/exposure ; temperature/temperature ; temperature/tint ; colorbalance/contrast ; colorbalance/output saturation ; clipping/angle ; denoiseprofile/on-off ; lens/on-off ; bilat/on-off
3. I'm not fully satisfied of the visual effect of this list of widgets (seems messy imo) I've tried other things, but nothing good... Maybe @Nilvus or someone else will have better idea

Thanks for your feedbacks

dterrahe:
> Internally, after @dterrahe comments, we have a list of all available widgets : dt_develop_t -> iop_widgets which stay in sync with current pipeline. This list is designed to be evolutive in order to be used for other purposes...

These other purposes already exist. Another layer is added here with no path towards integration. This additional use will need to be supported in a future cleanup/redesign, making it more complex/unlikely.

https://github.com/darktable-org/darktable/pull/3722#issuecomment-564629002

AlicVB:
Well, I was afraid of your answer...
Sadly, I fear that the underlining code you are speaking of may not land before looong time, considering the amount of work and its complexity (I'm now fully certain that I don't have understood all of it)
I've already face too many times that situation which stop projects, waiting for some "better-do-that-before" refactoring/core change

So I see 2 options now : 
1. I ditch the newly created list and go back with shortcut system "hijacking". This will have the real advantage that once the new underlining system is ready, portage will be straightforward (same as for shortcuts)
2. We keep the actual list and I port it to your new system once it's ready. After all, I've added very few complexity to the core code, and this shouldn't be hard to port.

Of course if you say "the new system will be ready next month" no problem, I'll wait...
But I certainly don't want to keep up-to-date and rebase all the work here for years... especially since imo this feature is really needed in dt !

dterrahe:
Look, I'm in the same boat you're in. I spent quite a bit of time implementing midi support, _without_ needing any new data structures (except, obviously midi specific ones) and was told basically, in the post I quoted, that I'd be responsible for fixing up the infrastructure or otherwise it would never get merged.

We still haven't heard much concrete from the core developers of what that new structure should look like. There have been several attempts to get the discussion started (see here #5063 for example) with no to very little response. So I'm just here pointing out that it seems we're only getting further away from the goal. A switchover to a new structure will only be possible if all the connecting points have been cleaned up and so far it feels like I've been the only one doing the work. Putting more work in front of me isn't exactly the best motivation is all I'm saying.

Now obviously it isn't up to me. At all. And it would not surprise me if different rules apply, like with dynamic accelerators at the time. But if this gets merged in this form I would really like it if someone could explain to me what is so different about midi.

I'm not at all trying to stop you from merging this; I think it is very useful functionality. What I would like is for there to be a bit more strategic thought, and help implementing it, so that all the different pieces we are building work better together and new features leverage and enhance all existing functionality rather than complicating/confusing or competing with it.

AlicVB:
So here is a "new" version. I'm using the first suggestion of @dterrahe here : reusing accels.

That way nothing is touch to the core (no addition, ...) So this PR is now completely distinct for the previous discussion and safe to be merged (on that point). That shouldn't stop the great work started in #5063 but at least this won't depend of this huge work.
Once the new "pluggable system" in place, it will be exactly the same amount of work to port this than to port accels (and I'll take care of course)

Back on the PR itself, I think it's now ready to review. So I remove the WIP. Of course, no urgency here as it's not for 3.4 !

Nilvus:
@AlicVB: just a thing. You ask me to check CSS and I see more than just CSS adjustment to do. I will try to propose a mockup ASAP. As the feature freeze for darktable 3.4 is today, I don't know if @TurboGit would like to integrate that on 3.4 or for 3.4.1.

TurboGit:
This is for after 3.4 clearly as noted by @AlicVB.

dterrahe:
Just did first test. Looking pretty promising!

Seems that when widgets get reparented back to their original module, they sometimes (but not always!) end up one position lower than where they started, eventually making it to the end of their box. I've added two reds in colorbalance; one ends up below the blue, the other stays where it belongs.

Nilvus:
@AlicVB: here is a proposal of a design for that part. My first idea was also to add module titles for those with at least 2 lines related. Here color balance and white balance. But by doing that in my mockup, I found that it overload it. On following mockup, I've put lines between each related module but without names, line between saturation output and exposure could be removed.

Line used is the line you could see in subtitles in modules like in color balance or white balance. So CSS setting already exist. I think that on/off toggle is better on start and as you can see with that mockup, we see clearly that it's not the same as a module, even on start. Simple design but efficient I think (and hope!).

basic

dterrahe:
Maybe a better icon for "go to full version" is an arrow pointing right?

The layout will look more cramped if there is a color picker.

Nilvus:
> Maybe a better icon for "go to full version" is an arrow pointing right?

I'm not convinced by an arrow pointing for that. But maybe a better icon is possible. I have no idea and so by now, the actual gear icon is the best I see. We have time to maybe find a better idea.

> The layout will look more cramped if there is a color picker.

At this stage, there's no color picker and as it's a basic "module", I think it's better to not have one on it.

dterrahe:
You can (currently) select to add a slider that has a colour picker in the original and it will appear here.

Nilvus:
> You can (currently) select to add a slider that has a colour picker in the original and it will appear here.

Oh, I didn't test a lot so didn't see that. Will check later and probably so propose another mockup. Thanks for pointing that.

AlicVB:
@Nilvus : great ! I like your mockup as usual... Thanks you again for all your work.
I'll let master code settled a bit on the modulegroups side and then try to rebase. Hard work, I fear... I don't know who has done so much changes here lately but... ...oops, that's me :D

Nilvus:
> The layout will look more cramped if there is a color picker.

@AlicVB: I've checked that statement of @dterrahe. He is right, that's cramped. I've made some tests and adding that CSS code will make UI more aerate when having color picker just before the gear icon to go to a module (that code avoid to apply the margin to another dt-button):

```
#basics-widget #bauhaus-combobox + #dt-button
{
margin-left: 0.5em;
}
```

Of course, feel free to add that to your code PR.

johnny-bit:
Oh Silly bot, we still want this basic module replacement!

AlicVB:
I'll rebase and continue the work here once #6975 is merged, no need to fix conflicts 2 times ;)

AlicVB:
Result of one full day of fighting : rebase 1-0 me... I've finally restarted from scratch and copy-paste manually everything...

So this is now in quite good shape imho... ready to be test...
- everything seems to works
- direct add-remove has been added like for other groups
- @Nilvus suggestions are done
- @dterrahe bug with ordering should be fixed too

johnny-bit:
Well, first look tells me this is FUN! Default presets need some work and considerations (eg scene refered + modern color woukflow would render temp/tint of white balance module useless ;) )

However I managed to break it when trying to re-edit module presets:

```
#0  0x00007f0233cda617 in wait4 () at /lib64/libc.so.6
#1  0x00007f0237756be0 in _dt_sigsegv_handler (param=11) at /home/johnny/darktable_repos/darktable/src/common/system_signal_handling.c:114
#2  0x00007f0233c4c9b0 in <signal handler called> () at /lib64/libc.so.6
#3  0x00007f01ce1c4407 in _basics_cleanup_list (self=0x5593d9ac8a90, edition=1) at /home/johnny/darktable_repos/darktable/src/libs/modulegroups.c:1747
#4  0x00007f01ce1c4787 in _manage_editor_load (preset=preset@entry=0x5593dc0a1460 "workflow: scene-referred_1", self=self@entry=0x5593d9ac8a90) at /home/johnny/darktable_repos/darktable/src/libs/modulegroups.c:2979
#5  0x00007f01ce1c897b in _manage_show_window (self=0x5593d9ac8a90) at /home/johnny/darktable_repos/darktable/src/libs/modulegroups.c:3365
#6  manage_presets (self=0x5593d9ac8a90) at /home/johnny/darktable_repos/darktable/src/libs/modulegroups.c:3396
#7  0x00007f0236673f6d in g_closure_invoke () at /usr/lib64/libgobject-2.0.so.0
```

[darktable_bt_I270V0.txt](https://github.com/darktable-org/darktable/files/5726666/darktable_bt_I270V0.txt)

AlicVB:
@johnny-bit : thanks for your tests. Both points you mentioned should now be fixed.

Nilvus:
Just a quick test and it's indeed much better. Just a thing in the UI: I finally find the on/off button too big. I think it's better balanced with that CSS tweak:

```
#basics-box  #module-enable-button
{
  padding: 0.2em;
}
```

AlicVB:
@Nilvus : done (not as proposed, as it doesn't change anything here, but directly with min-width/height)
I've also uniformized and refined the context-menu (the one when you right click on group icons and the ones inside the manager to add new modules/widgets)

TurboGit:
Tested for the fist time, it looks really nice ! I'll test more but one thing that I found missing is that you cannot add a preset here.

We have the possibility for add the on/off for a module but I think it would be nice to also add entries for presets. This will make possible to not only on/off "local contrast" but use a specific setting of this module.

TurboGit:
The separator between module is a thin line. I would also add the name of the module as right now it is not always easy to see which control is doing what. For example I have added the three factor sliders from "color balance" and it is hard to see which is which. Ok, there is a tooltip but yet...

image

TurboGit:
The wheel on the right is a bit too big also to me. It should have the same size of the on/off button.

johnny-bit:
I was just about to wrrite what Pascal suggested about presets

In addition to that - currently there's button per line of control to go to the main module. when i have configured off/on + 2+ controls for given module that gives 3 buttons to go to same module... so having a single button in that situation would be less visually straining imo. I don't know whether that would be the case without off/on line...

As for newest suggestion with factors - that's a tricky beast. 3 identically named controlls make 100% sense in original module but not one atop each other :/ would make sense with either name like "power / factor" "slope /factor" etc... or... "custom" name for control in adjustment? (I'm just chucking ideas, don't listen to me ;) )

dterrahe:
> This will make possible to not only on/off "local contrast" but use a specific setting of this module.

This starts to feel like a lot of duplication from normal custom module groups. Personally, I would not even have added the possibility to add module headers and on/off buttons here, because to me the point is to just have direct access to the controls. But I can see how others could have different opinions. If there is too much duplication, it starts to get confusing (because you would expect _everything_ to work here the same; can you also define _new_ presets? without even seeing all the controls that will be included in them? reset? everything or just the visible sliders?)

> As for newest suggestion with factors - that's a tricky beast. 3 identically named controls make 100% sense in original module but not one atop each other :/ would make sense with either name like "power / factor" "slope /factor" etc... or... "custom" name for control in adjustment? (I'm just chucking ideas, don't listen to me ;) )

I think color balance is a very specific case, where there are indeed many sliders with identical names. I don't think it makes sense to add only some or all of these to the basic module. For the same reason that @aurelienpierre thought putting them in different tabs is nonsensical; you manipulate all of them at the same time. If an advanced user _does_ want to have access to a limited number of these, they probably know what they are doing and know from the order which is which. I wouldn't optimise for this particular use case at the cost of more likely ones.

AlicVB:
great ! thanks all for you ideas and tests. I'll try to answer what I can : 
- I've added on-off buttons with "auto" configured modules in mind : lens, denoise profiled, ... This way we can quickly (de)activate lens correction for example.
- for the presets, it would be great, but there's a problem : it's different from other widgets as you can't desactivate them. in term of UI, I'm not sure how to present that, as it should be clear that it's not a setting like the other (even on-off can be seen as a setting, see above). If someone has ideas... (this already exists in the "quick preset menu" in the bottom panel)
- for the gear : no problem to reduce it's size. I'm not so sure about putting only one per module : for the alignment, that won't save any space for widgets and imho we should insist on the fact that important changes should be done in the main module...
- for the widgets labels : with @Nilvus we both have experimented to add the module name on top of each "module group" but it result in an overloaded aspect (see it's msg above) Maybe we can add an option for that ?
- Changing the widget label itself will be challenging to not get weird side effects, but I'll have a try. The risk -again- is to have a too long label for some widgets...

AlicVB:
last changes :
- smaller gear icon
- new prefs plugins/darkroom/modulegroups_basics_sections_labels to show/hide module names. I've added margin between modules "groups" to avoid a little bit the "overload" effect. After all the main intention is not to add here too much widget, so we should have enough spaces in majority of cases. @Nilvus (my UI guru) : when you have time, I'd like to have your opinion here :)
- widgets labels now integrate "section" name too. This is esp. useful for iop like colorbalance. Adding also module name is definitely not an option as it's too long in almost all cases
- the widget order inside a same module is now in the exact order that the module as registered them (and not in the reverse order as before)
- if a widget is hidden in the module itself (ex : green widgets in colorbalance with basic config) it shown deactivated with a specific tooltip msg. It's better than not showing anything !

data/themes/darktable.css
```diff
@@ -1431,6 +1414,73 @@ margin: 0 2px;
...
+#basics-box-labels #section_label
+{
+  margin: 0 0.75em 0.25em 0.75em;
+}
+
+#basics-box #basics-widget
+{
+  margin: 0 0.75em 0.5em 0.75em;
+}
+
```
Nilvus, line 1447: I suggets to add those lines here to aerate a little more:

```
#basics-widget
{
  padding-top: 2px;
}
```

Nilvus (changes requested):

Nilvus:
>     * smaller gear icon
> 
>     * new prefs plugins/darkroom/modulegroups_basics_sections_labels to show/hide module names. I've added margin between modules "groups" to avoid a little bit the "overload" effect. After all the main intention is not to add here too much widget, so we should have enough spaces in majority of cases. @Nilvus (my UI guru) : when you have time, I'd like to have your opinion here :)
> 
>     * widgets labels now integrate "section" name too. This is esp. useful for iop like colorbalance. Adding also module name is definitely not an option as it's too long in almost all cases
> 
>     * the widget order inside a same module is now in the exact order that the module as registered them (and not in the reverse order as before)
> 
>     * if a widget is hidden in the module itself (ex : green widgets in colorbalance with basic config) it shown deactivated with a specific tooltip msg. It's better than not showing anything !

I find gear icon small but it is ok as it is. I suggest  a little thing to improve, all other things are just good for me.

TurboGit:
I'm back on this one as it is time for merging this work. I found a little issue though. Go to basic group where you have the lens correction and denoise profile. Go back to module where denoise is and you get a module without the on/off button:

image

TurboGit (changes requested):
fix missing on/off button.

TurboGit:
BTW, I have planned this for 3.6 but we can also try for 3.4.x series.

AlicVB:
> fix missing on/off button.

fixed...
For the milestone, I don't mind ;)

src/libs/modulegroups.c
```diff
@@ -1149,18 +1640,184 @@ static void _manage_editor_groups_cleanup(GList **groups)
...
+  if(ma->multi_priority >= mb->multi_priority) return 0;
+  return 1;
+}
+
+static void _manage_editor_basics_update_list(dt_lib_module_t *self)
+{
+  dt_lib_modulegroups_t *d = (dt_lib_modulegroups_t *)self->data;
+
+  // first, we remove all existing modules
+  GList *lw = gtk_container_get_children(GTK_CONTAINER(d->edit_basics_box));
```
TurboGit, line 1715: given the doc I think lw should be freed.

src/libs/modulegroups.c
```diff
@@ -1149,18 +1640,184 @@ static void _manage_editor_groups_cleanup(GList **groups)
...
+
+  gtk_widget_show_all(d->edit_basics_box);
+}
+
+static void _basics_cleanup_list(dt_lib_module_t *self, gboolean edition)
+{
   dt_lib_modulegroups_t *d = (dt_lib_modulegroups_t *)self->data;
 
+  // ensure here that there's no basics widget of a module not present in one other group
+  GList *l;
```
TurboGit, line 1773: GList *l = edition ? d->edit_basics : d->basics;

src/libs/modulegroups.c
```diff
@@ -1149,18 +1640,184 @@ static void _manage_editor_groups_cleanup(GList **groups)
...
+  if(edition)
+    l = d->edit_basics;
+  else
+    l = d->basics;
+  while(l)
+  {
+    dt_lib_modulegroups_basic_item_t *item = (dt_lib_modulegroups_basic_item_t *)l->data;
+    gboolean exists = FALSE;
+    GList *ll;
```
TurboGit, line 1782: as above

src/libs/modulegroups.c
```diff
@@ -1338,15 +1992,75 @@ static void _manage_direct_module_toggle(GtkWidget *widget, GdkEventButton *even
...
+  return g_strcmp0(ia->id, (char *)b);
+}
+
+static void _manage_direct_basics_module_toggle(GtkWidget *widget, dt_lib_module_t *self)
+{
+  dt_lib_modulegroups_t *d = (dt_lib_modulegroups_t *)self->data;
+  gchar *wid = (gchar *)g_object_get_data(G_OBJECT(widget), "widget_id");
+  if(g_strcmp0(wid, "") == 0) return;
+
+  GList *found_item = g_list_find_custom(d->basics, wid, _basics_item_find);
```
TurboGit, line 2013: really not clear if found_item should be freed or not, the Gtk doc is not clear.

TurboGit (changes requested):
some comments, apart from that it is ready for merging to me.

src/libs/modulegroups.c
johnny-bit: found item shouldn't bee freed. `g_list_find_custom` returns pointer to element inside list, check for example https://github.com/GNOME/glib/blob/ce005e83c64d3d4fbf482f4bfd27a8aefc50b719/glib/glist.c#L919-L933 :)

AlicVB:
thanks for the review. All comment (apart last one) are fixed (+1 one-liner)

This will probably need some refinement on the default widget list, but that will be easier if it has landed in master first... So we can open a new PR and discussion to do that cleanly...

TurboGit:
All good, this is now merged into master. I'll keep that for 3.6 as I feel that some refinements will be done after first field tests.

Thanks a lot for this work which may help dt beginners to start learning dt in an easier way.

johnny-bit:
And now for PR to https://github.com/darktable-org/dtdocs to describe this goodie :D

PR #3649, opened 2019-12-01
color picker sRGB values between 0 and 255

### LINKED ISSUE #3479: Color picker -> sRBG range <> 0..255
**Describe the bug**
Using darktable  3.0.0.rc1+28 and the color picker, shown RGB values are outside the 0..255 range. Histogram profile = sRGB. See images.

**Screenshots**
image
rgb
Lab

Present version Manjaro/KDE

wpferguson:
CLAMPed rgb values in output strings the same way they are CLAMPed in the color patch display, which limits the values between 0 and 255.  Fixes #3479.

### DIFF
--- a/src/libs/colorpicker.c
+++ b/src/libs/colorpicker.c
@@ -171,7 +171,9 @@ static void _update_picker_output(dt_lib_module_t *self)
     {
       case 0: // rgb
         snprintf(colstring, sizeof(colstring), "%d, %d, %d",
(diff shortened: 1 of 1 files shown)


TurboGit:
This is not the right fix as we are lying to the user :) The output will be presented between 1 .. 255 but the actual value is not. We want to clip the actual value at the time it is taken/computed.

TurboGit:
In fact I was wrong, the RGB values are also clamped.

TurboGit:
I have merged my version with some other fixes.

Issue #3646, opened 2019-12-01
Artifacts on right corner of thumbnails


upegelow:
**Describe the bug**
Thumbnails of Images from my Canon 5DMkIV show artifacts on the right border. These artifacts are quite obvious if the lowpass module with a high radius and high saturation is applied. For me the artifacts manifest as a greenish halo (see screenshot). The tint of the green artifact is similar or the same than the color of the non-whitebalanced image.

The artifact is only visible in thumbsnails at certain magnification steps. It's not visible in the the other darktable views nor in exported images. It goes away if I crop away about 15 lines from the right hand side of the input image.

**To Reproduce**
1. Open 5DMkIV image into darktable
2. Reset history stack
3. In darkroom: activate lowpass module with a radius of 150 and saturation of 2
4. switch to another image or leave darkroom to have the thumbnail be reprocessed

**Expected behavior**
No artifacts.

**Screenshots**
![Screenshot_20191201_121239](https://user-images.githubusercontent.com/1794020/69913168-7d398d00-1434-11ea-8cb9-737ad66dfb24.png)

**Platform (please complete the following information):**
 - OS: OpenSuSE 15.1
 - Version: 3.0.rc2

**Additional context**
I personally don't assume that the lowpass module is the cause of the issue. Most likely the rightmost part of the scaled down thumbnail image (maybe only one line) has an artifact. The lowpass module just smears it over and makes it more visible.

jenshannoschwalm:
I took two raws from pixls ("Canon - EOS 5D Mark IV - RAW (3_2).CR2" and "Canon - EOS 5D Mark IV - RAW (3_2)(1).CR2" and both did not show any artefacts. (Fedora 31, git master) Neither cl activated or not. ???

upegelow:
> Neither cl activated or not. ???

Both. The artifacts are a bit more pronounced with OpenCL active.

Here is an example image which produces the issue for me:

[https://www.dropbox.com/s/93z6vgc6i9rpjer/IMG_4347.cr2?dl=0](url)

upegelow:
Just checked. Issue is also present with darktable 2.6.3.

timur-davletshin:
No problem on Debian 10. See no artifacts.

jenshannoschwalm:
@parafin your link is not valid (any more?).

I tested thumbnails generated from raw or not. Still can't reproduce.

upegelow:
I took one image from here:
[...]

PR #3666, opened 2019-12-04
thumbs/export: prevent reading outside of the input image's dimensions

upegelow:
this fix corrects for possible problems with the roi calculations. under
certain conditions (mainly thumbnail processing) the pixelpipe would request
input image dimensions higher than available thus reading outside of the
input image buffer and leading to artifacts - typically (a) column(s) of pixels
on the rightmost image border.

this is a fix for #3646


TurboGit:
A very simple fix and quite localized but I can imagine how much time it has taken to find the cause of this issue. Congrats Ulrich a nice one!

PR #3668, opened 2019-12-04
Fix header bar & background import options on folder/image import

Nilvus:
As we now have a lighter UI, these last times I frequently find the header bar quite too big (and also the group, favs, help and prefs icons) regardless of the rest of the UI. The goal of this new UI is also to let the most possible place to images.
I finally improve that and reduce particularly the header bar height and cited icons. I also reduce also some padding label text on the right (shortcuts to views).

Before :

Capture-20191204211932-1919x110

And after this PR :

Capture-20191204211846-1919x88

This also fix a not wanted background on check import options of import image/folder dialog box.

As this is related to UI and no string change, it's of course for 3.0 milestone.

### DIFF
--- a/data/themes/darktable.css
+++ b/data/themes/darktable.css
@@ -1248,7 +1248,7 @@ filechooser *:checked
     color: @field_selected_fg;
 }
 
-filechooser *:checked
+filechooser checkbutton:checked
 {
     background-color: transparent;
 }
--- a/data/themes/darktable.css
+++ b/data/themes/darktable.css
@@ -1248,6 +1248,11 @@ filechooser *:checked
     color: @field_selected_fg;
 }
 
+filechooser *:checked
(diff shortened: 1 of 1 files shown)


Nilvus:
Just a precision : changes impacted also on footer buttons are good also on that part but feel free to check if ok for you also.

Nilvus:
Other suggestion possible to reduce header bar is to change the line 642 from padding: 0.4 em; to padding: 0.2em 0.4em;

So reducing a little more margin top and bottom to have that header :
Capture-20191204221355-1919x83

TurboGit:
Looks good! Thanks for paying attention to all those details.

Nilvus:
> Looks good! Thanks for paying attention to all those details.

My pleasure !

Issue #3667, opened 2019-12-04
better OpenCL Scheduling Profile strategy


hqhoang:
**Is your feature request related to a problem? Please describe.**
Both nVidia card and the Intel 620 IGP support OpenCL. I set the "OpenCL Scheduling Profile" to default and set the device priority manually in darktablerc file (so that the faster nVidia handles main pipelines while the Intel handles thumbnail and preview). However, the nvidia-uvm module doesn't get unloaded on suspend/sleep if an application is using it (in this case, darktable). Thus, I usually lose OpenCL support for the nVidia device upon wakeup/resume (this is on Ubuntu). 

After closing darktable and modprobe nvidia-uvm, darktable would start up and see the nVidia again, then set the Scheduling Profile to "Multiple GPUs", which utilizes the Intel as the primary device (as it's device #0).

**Describe the solution you'd like**
Since there's already device benchmarking on startup, might as well use that to determine the primary vs secondary device and set priorities accordingly in case of "Multiple GPUs".

**Describe alternatives you've considered**
Or, if user manually specified device priorities in darktablerc, don't change the Scheduling Profile away from "default". Or, both.

upegelow:
Valid point. I would opt for your second proposal. I assume that only users will notice and are able to manage who would also be able to manually adjust darktablerc. In that case we could let them write a "magic number" into `opencl_checksum` which disables the automatic check for configuration updates - instead the settings once chosen by the user remain active until they are changed again.

hqhoang:
I also prefer second option, since the device_priorities is in darktablerc only.

However, I still want to improve the auto-priorities for Multiple GPUs a bit. Currently I think it sets to `*/*/*/*`, but if one device is at least 2x faster than the other, than the priorities can be better, e.g. `+1,*/!1,*/+1,*/0,*` (this is my current setting, with 1 being the faster), or just `1,*/!1,*/1,*/0,*`

I set the last one to `0,*` because I noticed in darkroom, when I hit the spacebar to go to the next image, the faster device sometimes is used to generate thumbnail for the current image, thus, not available to render the "full" view of the next image, thus, causing a slight delay.

upegelow:
I just saw that the config variable `opencl_checksum` is anyhow stored as a string. I will introduce the option to set this to OFF meaning no automatic re-checks in case of configuration changes.

jenshannoschwalm:
Does this also respect the case when there is a new graphics driver found requiring updating the compiled cl code?

[...]

PR #3670, opened 2019-12-05
opencl: allow to bypass automatic profile setting when system config changes

upegelow:
some (multi-GPU) systems change their configuration spontaneously and thereby
trigger each time an automatic setting of the opencl profile. users can
now set opencl_checksum=OFF in darktablerc which inhibits this behavior
and leaves settings completely under user control.

this addresses #3667

### DIFF
--- a/data/darktableconfig.xml.in
+++ b/data/darktableconfig.xml.in
@@ -158,7 +158,7 @@
     <type>string</type>
     <default></default>
     <shortdescription>checksum representing the setup of opencl devices on this computer</shortdescription>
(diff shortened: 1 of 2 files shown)


TurboGit:
@upegelow : What's you take on this? Should it be in 3.0 ? Looks safe enough to me but I prefer asking.

upegelow:
Let's wait for 3.0.1. It's an enhancement and not a bug fix plus it changes strings (though these seem not to be translatable). Anyhow we are in feature and string freeze and this one does not justify to break the rules.

hqhoang:
Applied your commit to my local copy and recompiled, set opencl_checksum=OFF, repeated my test case where the number of detected OpenCL devices goes from 2 => 1 => 2, the Scheduling Profile stayed at "default" now. Works great. Thanks.

TurboGit:
Merged.

PR #3678, opened 2019-12-07
basecurve: revert back to no color preservation as in 2.6

### LINKED ISSUE #3677: Basecurve → Preserve colors as new default
dt 3.0 is using preserve color = luminance as new default. This gives results different from in camera JPEGs. Which is a good start for new edits especially from new users — they expect results close to in camera conversion.

I propose to revert old defaults preserve color = none.

In camera JPEG:
In camera JPEG

Old default (no preserve colors):
Basecurve (old default)

New default (preserve color = luminance):
Basecurve (new default)

Raw image:
https://drive.google.com/open?id=1vY9rmpcNYz5qTSHzwcwETgAKeOut24t1

Difference is even stronger on sunset or bright sky scenes when color shift is very visible.

TurboGit:
Fixes #3677.


TurboGit:
The rational is that there is no real point of changing the default for basecure. Better keep compatibility with 2.6 here. And anyhow advanced users will probably not be using the basecurve but will move to a filmic based workflow.

TurboGit:
I'm closing. Preserving the color is expected to be better even if not the visual effect we are used with 2.6 as default.

PR #3683, opened 2019-12-07
darkroom loading screen

AlicVB:
Display a empty screen during image loading. This avoid the impression of freezed screen when loading image with heavy settings.
When I was at it, I've also removed the hardcoded font color when drawing the "no image" screen in lighttable.

So we'll have to add and set 2 colors in our themes : 
- darkroom_bg_font_color
- lighttable_bg_font_color

^^^ @Nilvus ???


Nilvus:
@AlicVB : I will test it today and report you a suggestion for that.

Nilvus:
After some tests, why not simply adding the "loading image xxxx" with the "working..." when we load an image from lighttable to darkroom mode ? It will be better seeing. I find it strange to have 2 different lines with two different seeing. It seems that the "working.." text remains hardcoded as I don't find how to change how it renders. By the way, it would be good to simply add a point on "working.." text as it only have 2 points. See line 307 on src/control/control.c file.

AlicVB:
> why not simply adding the "loading image xxxx" with the "working..." when we load an image from lighttable to darkroom mode ?

because I'm lazy :) The "working" message is sort of common to all operation that can take time. I'm not sure where it's located in the code... but I'll have a look (and don't hardcode it if I can)

Nilvus:
> > why not simply adding the "loading image xxxx" with the "working..." when we load an image from lighttable to darkroom mode ?
> 
> because I'm lazy :) The "working" message is sort of common to all operation that can take time. I'm not sure where it's located in the code... but I'll have a look (and don't hardcode it if I can)

Line 307 on src/control/control.c,

AlicVB:
yes, I've seen that in your first msg. Thanks for avoiding me a search ;)
The problem is that the "working" msg is tied to `darktable.control->log_busy` which is used in many delicate places in the code, so I'm quite reluctant to touch it. Of course we can use `dt_control_log`, but : 
- msg will be overwrite by "working" on each redraw
- msg stay for a predefined nb of second, not like the "working" one which is shown only and exactly during the pipe work...

The simplest solution might be to draw the loading msg with the same appearance and just above the "working" one. I'll do a try.

Oh and when I was at it, if I read the code correctly, the "working" msg use `bg_color, selected_bg_color, fg_color` from the center widget. Can you theme this ?
I fear that changing color will be the only point we can change, as it's a cairo drawing (I'm not a specialist, maybe I'm wrong...)

AlicVB:
here's how it looks now : 
Capture d’écran_2019-12-08_19-38-37

Nilvus:
> yes, I've seen that in your first msg. Thanks for avoiding me a search ;)
> The problem is that the "working" msg is tied to `darktable.control->log_busy` which is used in many delicate places in the code, so I'm quite reluctant to touch it. Of course we can use `dt_control_log`, but :
> 
>     * msg will be overwrite by "working" on each redraw
> 
>     * msg stay for a predefined nb of second, not like the "working" one which is shown only and exactly during the pipe work...
> 
> 
> The simplest solution might be to draw the loading msg with the same appearance and just above the "working" one. I'll do a try.
> 
> Oh and when I was at it, if I read the code correctly, the "working" msg use `bg_color, selected_bg_color, fg_color` from the center widget. Can you theme this ?
> I fear that changing color will be the only point we can change, as it's a cairo drawing (I'm not a specialist, maybe I'm wrong...)

I can't theme that because bg_color, selected one and fg_color are general tags applied on other parts and there's no specific tag on that part I could point. Said that, I find it as it is good like what you done. Having the loading image just below the working one is good for me and working colors/font is good also for me, so as the new line.

Nilvus:
@AlicVB : on line 307 on src/control/control.c, there still missing a point on suspension one for working. It could be good to correct that with your PR.

Nilvus:
@AlicVB : not sure is necessary here but as you want. I really find now these lines good as they are. But if you add something, 2 colors will be ok.

AlicVB:
@Nilvus : so I've added 2 "new" colors : `log_bg_color` and `log_fg_color` and removed `darkroom_bg_font_color`. I've done the css files too, as you have suggested (if you prefer, I can remove the css part from this PR and let you do it after)

When I was at it, I've slightly change and simplify the log msg drawing by removing the border and center the text in both directions.

data/themes/darktable-elegant-dark.css
```diff
@@ -72,12 +72,15 @@
 /* Tooltips and contextual helpers */
 @define-color tooltip_bg_color @grey_20;
 @define-color menu_bg_color @grey_20;
+@define-color log_bg_color @grey_50;
```
Nilvus: @grey_40 would be better, make text more readable

data/themes/darktable.css
```diff
@@ -116,12 +116,15 @@
 @define-color tooltip_bg_color @grey_05;
 @define-color tooltip_fg_color @grey_70;
 @define-color really_dark_bg_color @grey_05;
+@define-color log_bg_color @grey_35;
```
Nilvus: @grey_25 would be better, make text more readable

AlicVB:
@Nilvus : I've just pushed your changes. Thanks ! (I'm a bit embarrassed to commit your changes in my name, thought...)

Nilvus:
> @Nilvus : I've just pushed your changes. Thanks ! (I'm a bit embarrassed to commit your changes in my name, thought...)

No worries about that. The goal is just to improve darktable. And you add credits on commit title. And making a PR just for that, not needed. You make 99% of the work, I just suggest you little improvements.

TurboGit:
@AlicVB : You can always commit for another user using:

`$ git commit --author="Firstname Lastname <somename@here.com>"`

I do that for some translations sent to me directly by e-mail for example.

PR #3690, opened 2019-12-07
spots : don't allow user to add more than 64 shapes

### LINKED ISSUE #3689: Unlimited number of strokes in spot removal tool
Opening a photo in Darktable, I noticed a lot of dust was on the lense and I wanted to remove it from the photo with the spot removal tool. After adding 64 shapes, the stains were visibly not removed anymore. I tried to export the photo but there as well, no removal.

I want to be able to add any number of shapes for spot removal.

**Describe alternatives you've considered**
I don't really see any reason why to limit the number of shapes added, but if there was, it should be settable in the app settings.

**Additional context**
It seems like the algorithm only iterating over the first 64 strokes (hardcoded number) and then silently quits the process loop as the condition doesn't match anymore:

https://github.com/darktable-org/darktable/blob/ad1110f98ffccb826d7340ea9f0c94b223cea1b7/src/iop/spots.c#L393

AlicVB:
this fix #3689 

This may be for 3.01, as it's not critical and I've added a new translatable string

### DIFF
--- a/src/iop/spots.c
+++ b/src/iop/spots.c
@@ -164,24 +164,38 @@ static void _resynch_params(struct dt_iop_module_t *self)
 }
 
 
-static void _reset_form_creation(GtkWidget *widget, dt_iop_module_t *self)
+static gboolean _reset_form_creation(GtkWidget *widget, dt_iop_module_t *self)
 {
   dt_iop_spots_gui_data_t *g = (dt_iop_spots_gui_data_t *)self->gui_data;
-  if(gtk_toggle_button_get_active(GTK_TOGGLE_BUTTON(g->bt_path)) ||
-      gtk_toggle_button_get_active(GTK_TOGGLE_BUTTON(g->bt_circle)) ||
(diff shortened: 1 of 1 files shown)


src/iop/spots.c
```diff
@@ -164,24 +164,38 @@ static void _resynch_params(struct dt_iop_module_t *self)
...
-      gtk_toggle_button_get_active(GTK_TOGGLE_BUTTON(g->bt_ellipse)))
+
+  // we check the nb of shapes limit
+  dt_masks_form_t *grp = dt_masks_get_from_id(self->dev, self->blend_params->mask_id);
+  guint nb = 0;
+  if(grp && (grp->type & DT_MASKS_GROUP)) nb = g_list_length(grp->points);
+
+  if(nb >= 64)
+  {
+    dt_control_log(_("Spot module is limited to 64 shapes. Please add a new instance !"));
```
TurboGit: Just spotted this... "Spot module..." Spot should have a lowercase S at start. Likewise for "Please".

AlicVB: oups

TurboGit:
Merged.

PR #3694, opened 2019-12-08
Writing for split-toning module standardized

mepi0011:
The writing of Split Toning adapted to the usual writing.
See also pull request # 3695


src/iop/splittoning.c
```diff
@@ -87,7 +87,7 @@ typedef enum dt_iop_splittoning_picker_t
 
 const char *name()
 {
-  return _("split toning");
+  return _("Split Toning");
```
TurboGit: All word in dt are lower case. So we cannot change that.

src/iop/colorbalance.c
```diff
@@ -272,10 +272,10 @@ static void add_preset(dt_iop_module_so_t *self, const char *name,
 void init_presets(dt_iop_module_so_t *self)
 {
   // these blobs were exported as dtstyle and copied from there:
-  add_preset(self, _("split-toning teal-orange (2nd instance)"),
+  add_preset(self, _("Split Toning teal-orange (2nd instance)"),
```
TurboGit: I would probably standardize with "split-toning".

Changing the case here is not possible or we need to change all the case of all dt strings.

src/iop/splittoning.c
mepi0011: I didn't know this rule.

Changing to lower case will no problem, but then I would prefer to write it split-toning as you mentioned below.

I would use the same spelling for the user manual and adjust it accordingly.

src/iop/colorbalance.c
mepi0011: See comment above.

src/iop/splittoning.c
TurboGit: Fine with me.

mepi0011: Done!

TurboGit:
Thanks !

TurboGit:
Merged on master and 3.0.1.

PR #3696, opened 2019-12-08
opencl: avoid collecting debug information if not needed

upegelow:
this avoids unnecessary and expensive calls to clGetDeviceInfo


TurboGit:
Thanks!

PR #3698, opened 2019-12-08
fix no update xmp after undo/redo

phweyland:
In dt_undo_do_undo() and dt_undo_do_redo() I've added the an image list filled in by undo routine.
At the end, if there are some images in the list, xmp is refreshed for them.

Exception: libs/history.c where there is no image id visible. And anyway the xmp file for dev is updated only when the user moves to another image (or on dt restart).

Doubt: common/history_snapshot.c. The undo routine adds the image to the list, but Ctrl-Z/Ctrl-Y don't update the xmp file. I don't know why.

### DIFF
--- a/src/common/colorlabels.c
+++ b/src/common/colorlabels.c
@@ -38,7 +38,7 @@ typedef struct dt_undo_colorlabels_t
   uint8_t after;
 } dt_undo_colorlabels_t;
 
-static void _pop_undo(gpointer user_data, dt_undo_type_t type, dt_undo_data_t data, dt_undo_action_t action)
+static void _pop_undo(gpointer user_data, dt_undo_type_t type, dt_undo_data_t data, dt_undo_action_t action, GList **imgs)
 {
   if(type == DT_UNDO_COLORLABELS)
   {
@@ -59,6 +59,7 @@ static void _pop_undo(gpointer user_data, dt_undo_type_t type, dt_undo_data_t da
         if(labels & (1<<color))
           dt_colorlabels_set_label(clabels->imgid, color);
 
+      *imgs = g_list_prepend(*imgs, GINT_TO_POINTER(clabels->imgid));
       list = g_list_next(list);
     }
   }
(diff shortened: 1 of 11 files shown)


phweyland:
> Note: during tests I've seen that metadata UI is partly broken. If you select several images with different metadata, modify one of the metadata item and press enter, all visible metadata are written on all images. Before, when a metadata element was not the same on every image some <I don't remember the text> did appear. And this element was not updated. Not the case anymore.

In fact if a metadata item has different defined values it appears as <leave unchanged>.
When a metadata item is defined for one image and not for the others, the same message should appear, right ?

TurboGit:
@phweyland : you said 

> Doubt: common/history_snapshot.c. The undo routine adds the image to the list, but Ctrl-Z/Ctrl-Y don't update the xmp file. I don't know why.

What do you mean? I did a delete history and did an undo and the history was properly restored into the XMP?

phweyland:
> @phweyland : you said
> 
> > Doubt: common/history_snapshot.c. The undo routine adds the image to the list, but Ctrl-Z/Ctrl-Y don't update the xmp file. I don't know why.
> 
> What do you mean? I did a delete history and did an undo and the history was properly restored into the XMP?

With this PR ? When I made the test (remove some steps in the history) the xmp file was not updated after Ctrl-Z but I may have made some mistake somewhere.

phweyland:
My doubt was also on the necessity to include common/history_snapshot.c in that change. The update of xmp file could be already done somewhere else in the dev pipe.

TurboGit:
> The update of xmp file could be already done somewhere else in the dev pipe.

Indeed it is already done as we invalidate the dev to recompute the thumbnail, and so the XMP got recreated properly already. So maybe we should remove this part with some comment to be sure we remember why we do not add the list of imgid.

phweyland:
As I don't know really how the xmp update is triggered in the dev pipe I'm not very sure of the comment I've put in history_snapshot undo.

PR #3699, opened 2019-12-08
Fixing testing for a mask manager at position 0 in history 

jenshannoschwalm:
We have to test whether a mask_manager module is in position 0 of the history when we do a compression of history. The test had been wrong, this could possibly lead to
1. wrong num in main.history
2. wrong history_end

### DIFF
--- a/src/common/history.c
+++ b/src/common/history.c
@@ -963,7 +963,6 @@ static int dt_history_end_attop(int32_t imgid)
   DT_DEBUG_SQLITE3_PREPARE_V2(dt_database_get(darktable.db),
     "SELECT history_end FROM main.images WHERE id=?1", -1, &stmt, NULL);
   DT_DEBUG_SQLITE3_BIND_INT(stmt, 1, imgid);
-
   if (sqlite3_step(stmt) == SQLITE_ROW)
     end = sqlite3_column_int(stmt, 0);
   sqlite3_finalize(stmt);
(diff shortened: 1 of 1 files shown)


src/common/history.c
```diff
@@ -1007,35 +1007,38 @@ void dt_history_compress_on_image(int32_t imgid)
...
-  DT_DEBUG_SQLITE3_BIND_INT(stmt, 2, my_history_end);
-  sqlite3_step(stmt);
-  sqlite3_finalize(stmt);
 
   int masks_count = 0;
   char op_mask_manager[20] = { 0 };
 
   gboolean manager_position = FALSE;
+
   // do we already have a mask manager at the correct position? We don't want to increase history nums later
```
TurboGit: Maybe update this comment?

// do we already have a mask manager in the history

src/common/history.c
```diff
@@ -1007,35 +1007,38 @@ void dt_history_compress_on_image(int32_t imgid)
...
+                              "multi_priority)",
+                              -1, &stmt, NULL);
+  DT_DEBUG_SQLITE3_BIND_INT(stmt, 1, imgid);
+  DT_DEBUG_SQLITE3_BIND_INT(stmt, 2, my_history_end);
+  sqlite3_step(stmt);
+  sqlite3_finalize(stmt);
+
+
   // delete all mask_manager entries
   g_strlcpy(op_mask_manager, "mask_manager", sizeof(op_mask_manager));
```
TurboGit: Already done above?

src/common/history.c
```diff
@@ -1007,35 +1007,38 @@ void dt_history_compress_on_image(int32_t imgid)
...
-  sqlite3_step(stmt);
-  sqlite3_finalize(stmt);
 
   int masks_count = 0;
   char op_mask_manager[20] = { 0 };
 
   gboolean manager_position = FALSE;
+
   // do we already have a mask manager at the correct position? We don't want to increase history nums later
   g_strlcpy(op_mask_manager, "mask_manager", sizeof(op_mask_manager));
```
TurboGit: Maybe time to clean this up? I suppose we could just have and remove this line.

const char *op_mask_manager = "mask_manager";

no?

See also below.

src/common/history.c
jenshannoschwalm: right

src/common/history.c
jenshannoschwalm: Nope. The topmost mask manager in history will still be there ...

jenshannoschwalm:
Have worked your requested changes.

TurboGit:
Looks good to me! Thanks!

PR #3700, opened 2019-12-08
fix metadata display issue when item is present for one image but not for others

phweyland:
When several images are selected and a metadata item is defined for some images but not all the message \<leave unchanged\> is not displayed. Instead one of the value (1) for that item is displayed.
As a consequence if the user clicks on apply or enter another metadata item, the displayed value (1) is attached to all selected images. This is not correct.
This PR hopefully fixes this.


TurboGit:
Oh a nice fix indeed! Thanks for tracking this down.

PR #3716, opened 2019-12-10
add a focus peaking mode on cairo surfaces

aurelienpierre:
Add a focus peaking display mode in lighttable thumbnails, full preview and darkroom that displays sharp edges to help assessing in-focus areas during culling.

The feature is triggered with Ctrl+Shift+F by default. The algorithm prefilters the noise, then computes the gradients in picture, and finally the gradients statistics (average and predicator), assuming gradients follow an hyper-laplacian distribution in natural images (similar to a gaussian distribution but with heavier tails) : 

* edges gradients above esperance + 10 × predicator are shown in yellow (very sharp), 
* edges gradients above esperance + 7.5 × predicator are shown in green (medium sharp),
* edges gradients above esperance + 5 × predicator are shown in blue (little sharp).

Following the transitions of colours along edges, one can follow the path of the depth of field. This doesn't replace the focus preview in lighttable (Ctrl+W). Both can be showed together.

Notice that this performs a contrast detection, assuming high contrast means sharpness. It will make no difference between optical sharpness and high contrast transitions, for example between bright backgrounds and dark foregrounds.

Demo lighttable:
Capture d’écran du 2019-12-10 01-11-24

Demo darkroom:
Capture d’écran du 2019-12-20 01-21-40

The darkroom focus-peaking will be useful to assert the quality of the various debluring algorithms.

Demo at shallow depth of field (f/2 135 mm):
Capture d’écran du 2019-12-20 01-23-35

One has just to pick a picture where the eyes are sharp.

Details:

The code fetches images from the cairo surface after the picture is drawn in UI. It doesn't care if the image is an embedded thumbnail or a processed raw, or if the pipeline has been recomputed. The benefit is it works also on the mipmap cache and doesn't need access to the original raw files (the third demo shows pictures stored on a hard drive that was not connected to my computer).

### DIFF
--- a/src/common/focus_peaking.h
+++ b/src/common/focus_peaking.h
@@ -65,13 +65,13 @@ static inline float laplacian(const float *const image, const size_t index[8])
 #ifdef _OPENMP
 #pragma omp declare simd
 #endif
-static inline void get_indices(const size_t i, const size_t j, const size_t width, const size_t height, size_t index[8])
+static inline void get_indices(const size_t i, const size_t j, const size_t width, const size_t height, const size_t delta, size_t index[8])
 {
-  const size_t upper_line = (i - 1) * width;
-  const size_t center_line = upper_line + width;
-  const size_t lower_line = center_line + width;
-  const size_t left_row = j - 1;
-  const size_t right_row = j + 1;
+  const size_t upper_line = (i - delta) * width;
+  const size_t center_line = i * width;
+  const size_t lower_line = (i + delta) * width;
+  const size_t left_row = j - delta;
+  const size_t right_row = j + delta;
 
   index[0] = upper_line + left_row;       // north west
   index[1] = upper_line + j;              // north
@@ -111,7 +111,7 @@ schedule(static) collapse(2) aligned(image, luma:64)
     }
 
   // Prefilter noise
-  fast_surface_blur(luma, buf_width, buf_height, 8, 0.01f, 4, DT_GF_BLENDING_LINEAR, 1, 0.0f, exp2f(-8.0f), 0.0f);
+  fast_surface_blur(luma, buf_width, buf_height, 12, 0.00001f, 4, DT_GF_BLENDING_LINEAR, 1, 0.0f, exp2f(-8.0f), 1.0f);
 
   // Compute the gradients magnitudes
   float *const restrict luma_ds =  dt_alloc_sse_ps(buf_width * buf_height);
@@ -120,19 +120,26 @@ schedule(static) collapse(2) aligned(image, luma:64)
 dt_omp_firstprivate(luma, luma_ds, buf_height, buf_width) \
 schedule(static) collapse(2) aligned(luma_ds, luma:64)
 #endif
-  for(size_t i = 1; i < buf_height - 1; ++i)
-    for(size_t j = 1; j < buf_width - 1; ++j)
+  for(size_t i = 2; i < buf_height - 2; ++i)
+    for(size_t j = 2; j < buf_width - 2; ++j)
     {
-      size_t index[8];
-      get_indices(i, j, buf_width, buf_height, index);
-      luma_ds[i * buf_width + j] = laplacian(luma, index);
+      size_t index_close[8];
+      get_indices(i, j, buf_width, buf_height, 1, index_close);
+
+      size_t index_far[8];
+      get_indices(i, j, buf_width, buf_height, 2, index_far);
+
+      // Computing the gradient on the closest neighbours gives us the rate of variation, but doesn't say if we are
+      // looking at local contrast or optical sharpness.
+      // so we compute again the gradient on neighbours a bit further.
+      // if both gradients have the same magnitude, it means we have no sharpness but just a big step in intensity,
+      // aka local contrast. If the closest is higher than the farthest, is means we have indeed a sharp something,
+      // either noise or edge. To mitigate that, we just subtract half the farthest gradient but add a noise threshold
+      luma_ds[i * buf_width + j] = laplacian(luma, index_close) - 0.67f * (laplacian(luma, index_far) - 0.00390625f);
     }
 
-  // Postfilter to join isolated dots and draw lines
-  fast_surface_blur(luma_ds, buf_width, buf_height, 12, 0.0001f, 4, DT_GF_BLENDING_LINEAR, 1, 0.0f, exp2f(-8.0f), 0.0f);
-
-  // Anti-aliasing filter
-  box_average(luma_ds, buf_width, buf_height, 1, 1);
+  // Anti-aliasing
+  box_average(luma_ds, buf_width, buf_height, 1, 2);
 
   // Compute the gradient mean over the picture
   float TV_sum = 0.0f;
@@ -142,11 +149,11 @@ schedule(static) collapse(2) aligned(luma_ds, luma:64)
 dt_omp_firstprivate(luma_ds, buf_height, buf_width) \
 schedule(static) collapse(2) aligned(luma_ds:64) reduction(+:TV_sum)
 #endif
-  for(size_t i = 1; i < buf_height - 1; ++i)
-    for(size_t j = 1; j < buf_width - 1; ++j)
+  for(size_t i = 2; i < buf_height - 2; ++i)
+    for(size_t j = 2; j < buf_width - 2; ++j)
       TV_sum += luma_ds[i * buf_width + j];
 
-  TV_sum /= (float)(buf_height - 2) * (float)(buf_width - 2);
+  TV_sum /= (float)(buf_height - 4) * (float)(buf_width - 4);
 
   // Compute the predicator of the hyper-laplacian distribution
   // (similar to the standard deviation if we had a gaussian distribution)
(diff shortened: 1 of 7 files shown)


jenshannoschwalm:
@aurelienpierre congrats - something like this has been on my personal wishlist at a rather top position.

I tried it and think - you seem to have recognised this also - the algorithm is still too much dominated by luminosity changes, at least your commit headline lets me think so.

Have you tried to use the minimum or maximum of the r/g/b channels as input source for the contrast detection? Another idea to enhance concentration more on the edge information would be desat, darken or grayscale the image itself. What's your point on this?

aurelienpierre:
> Have you tried to use the minimum or maximum of the r/g/b channels as input source for the contrast detection? Another idea to enhance concentration more on the edge information would be desat, darken or grayscale the image itself. What's your point on this?

There is no link to the global maximum or minimum once you go in frequential space, because we simply compute the first order discrete derivative between one pixel and its neighbours. To say it otherwise, we measure how much each pixel stands out from the 8 pixels nearby it by getting its rate of variation, aka its derivative. Then, taking the hypotenuse of the derivatives along x and y, we get the magnitude of the gradient, which is what you see on the overlay.

The derivatives we measure are du/dx and du/dy, with u the pixel intensity (colour or grey channel, doesn't matter). True sharpness (optical modulation transfer function) happens when dx or dy are close to zero, so the derivative are very high. It means we have a steep slope, but not necessarily a large step in intensity, between 2 pixels. Fake sharpness (local contrast) happens when du is very high, so the derivative are very high too. And you can visually fake sharpness by simply increasing du, that's the point of the unsharp masking (sharpness module in dt), contrast equalizer and local contrast modules. 

Since we are in a discretized space, dx = dy = 1 px, then all the info we can get is u and du = u(i, j) - u(i, j - 1), thus we can't make the difference between optical sharpness and simple local contrast, because we can't say if du/dy is big because dy is small or du is big. 

The more resolution you have in your picture, the closer your contrast detection becomes to a sharpness detection. But that would be crazy slow to run in lighttable.

Speaking of which, sit tight, I just got an idea.

aurelienpierre:
yep, I just found an elegant way to remove sensitivity to local contrast.

jenshannoschwalm:
I think you misunderstood slightly.

> The derivatives we measure are du/dx and du/dy, with u the pixel intensity (colour or grey channel, doesn't matter
 
I think it surely does matter. du/dx is pretty abstract - don't you use a one-dimensional signal for further calculations? So this signal could be luminance. **Or** it might be just the value from the red channel if we were interested only in this for some reason. **Or** if we hypothetically would have a camera with an extra IR sensor (ir/r/g/b) we could use that value to detect warm/cold edges. It's just a matter of "what kind of signal changes are we interested in?"
```
      luma[index] = sqrtf( powf(uint8_to_float(image[index_RGB]), exponent) +
                           powf(uint8_to_float(image[index_RGB + 1]), exponent) +
                           powf(uint8_to_float(image[index_RGB + 2]), exponent) );
```
That's what you chose as your input signal - some sort of luma by equally weighing all rgb channels. A (0.1,0.1,0.4) --> (0.4,0.1,0.1) signal change would not result in any diff although our eyes/brain clearly detect this? So, is there anything better than the above luma calculation?
(You already do the sqrtf to detect signal changes in dark regions)

An example, photos from olympic NP with excessive greens in the rain forest. Just using the red channel makes faces and dried leaves standing out.

aurelienpierre:
> I think it surely does matter. du/dx is pretty abstract - don't you use a one-dimensional signal for further calculations? So this signal could be luminance. **Or** it might be just the value from the red channel if we were interested only in this for some reason. **Or** if we hypothetically would have a camera with an extra IR sensor (ir/r/g/b) we could use that value to detect warm/cold edges. It's just a matter of "what kind of signal changes are we interested in?"

That doesn't change anything. An image is a collection of 2D gradients. du/dy is very concrete : it's the rate of variation. Take the R, G, B channels or any combination of these, the gradients map of any channel will perfectly overlap the gradient map of any other channel, except where you have noise or chromatic aberrations. There is no perception involved here, it's optics. The only reason I take the euclidean norm of the RGB as a "luma" estimator is because it dilutes the RGB noise to some extent, but if your signal was clean, it would give exactly the same results as taking any RGB channel, or any linear combination of them.

> it's just a matter of "what kind of signal changes are we interested in?"

> An example, photos from olympic NP with excessive greens in the rain forest. Just using the red channel makes faces and dried leaves standing out.

It's way more complicated than that. You need to think in terms of Fourier spectral analyse here. We are looking for high frequencies, we don't care about the average values. See https://en.wikipedia.org/wiki/Edge_detection, in particular, this algorithm here is a variation of the Canny filter https://en.wikipedia.org/wiki/Canny_edge_detector (adapted since I'm looking for a focus heatmap, not an edge detection).

jenshannoschwalm:
> That doesn't change anything. An image is a collection of 2D gradients. du/dy is very concrete : it's the rate of variation of the pixel intensity.

Not really -- what is **intensity** here? If you only use the red channel for calculation, changes in b/g would not have any difference at all.

Bildschirmfoto von 2019-12-11 06-52-22
Is what you suggested (before your latest commit)

Bildschirmfoto von 2019-12-11 06-50-42
This what you get while using
`luma[index] = sqrtf(sqrtf(sqrtf(powf(3.0 * (uint8_to_float(image[index_RGB + 2])), exponent) )));`

And 
Bildschirmfoto von 2019-12-11 06-59-43
`luma[index] = sqrtf(powf(3.0 * (uint8_to_float(image[index_RGB + 2])), exponent) );`

> There is no perception involved here

Right. But we don't want a pure signal analysis but want an image with points of signal-changing standing out. There is the "perception" - not for color but for sort of "content".

aurelienpierre:
> what is **intensity** here?

Intensity is the value at pixel coordinates for whatever channel or metric you chose.

> But we don't want a pure signal analysis but want an image with points of signal-changing standing out. There is the "perception" - not for color but for sort of "content".

No, we want to see where the autofocus has done the focus. It's purely physics. Applying gamma-like curves makes no sense and will only make the noise explode. We don't want to draw nice contours, we are not doing machine vision, we just want a heatmap of sharpness so one can pick, between 2 images, the one that has the eyes in focus. I think you are confusing the goal.

dterrahe:
This looks great; will be very useful in whichever exact form it eventually takes.

Feel free to ignore this suggestion, but I would immediately remap this to CTRL-F (to be consistent with CTRL-G gamut and CTRL-Soft Proof). I have CTRL-SHIFT-F mapped to Filmstrip toggle, to match CTRL-L/T/R/B for toggling side panels.

jenshannoschwalm:
> No, we want to see where the autofocus has done the focus. It's purely physics.

You can't do the first, and pure physics is simply wrong.

I don't know whether i am confusing the goal, i think we both don't want contour detection. If you mean, we want to detect unblurred regions and make them "visibly outstanding" by whatever method, i fully agree. 

I read your algorithm. There are many papers around about exactly this problem because it's not trivial and not physics. You might already know
1. https://www.cs.cmu.edu/~htong/pdf/ICME04_tong.pdf
2. https://www.sciencedirect.com/science/article/abs/pii/S0923596518302832 (which i can't download)
3. https://www.cs.usask.ca/faculty/eramian/defocusseg/

And cited from the last paper (i marked as bold in the comment myself to point out what i observe in the implemented algorithm) 

> Measures  such  as  higher  order  statistics  [21],  variance  of wavelet coefficients [52], local variance image field [53] have been used directly in segmentation of objects of interest in low-depth-of-field  images. These local sharpness metrics are based on local image energy which means that **the measures will not only decrease if the energy of the PSF (point spreadfunction) decreases (becomes more blurry), but also decreases if the energy of the image content drops.** Thus, a blurry, high-contrast edge region could have a higher sharpness score than an in-focus, low-contrast one.  These metrics are suitable for relative sharpness measures, e.g. in focal stacking, but do not behave  very well for local sharpness measure across various image contents. This deficiency has already been pointed out in [55].

> Applying gamma-like curves makes no sense and will only make the noise explode and produce false positive.

Noise increasing? Probably yes but i am not sure. Basically it emphasises signal changes in dark regions.

@aurelienpierre just discussing whether there are better algos around for this purpose.

aurelienpierre:
The paper you are citing aims at segmenting images depending on focus to extract something like an alpha channel. It's more constraints than what we do here, because they need continuity in their zones and need to infer a PSF, which is not what we do here (and, again, it's pure physics, since it derivates from the thin lens optical model -> PSF).

The approach I'm taking to limit the sensibility to image content drops (global contrast), is computing 2 gradients for each pixel:
1. one for the closest neighbours (1 pixel distance)
2. one for farther neighbours (2 pixels distance)

If both gradients have the same magnitude, we can fairly assume we are at the edge of background/foreground and don't have real sharpness. If the first gradient is higher in magnitude than the second, we can assume we have either a sharp edge or noise (but noise will be filtered twice in the pre- and post-filtering). To limit, the sensitivity to local contrast, I use the difference `gradient_1 - 0.67 gradient_2`, 67% being rougly the cumulative probability of a gaussian distribution (area under the density of probability curve) between ± std.

The quality of the detection is relatively easy to assess visually, given that the focus is done on a focal plane ± (depth of field) / 2: all the sharpest details should be located on a plane at the same distance from the lens (so you see that 2 of the examples you gave above don't make sense physically). 

>  Basically it emphasises signal changes in dark regions.

Hence it emphasises noise, which is false local contrast mostly annoying in dark regions.

jenshannoschwalm:
@aurelienpierre Thanks a lot!  BTW - have a good party!

aurelienpierre:
Thanks ;-)

TurboGit:
Can you fix the conflict ? Thanks!

TurboGit:
BTW, when I tested this some days ago I found that on the lighttable when activating this mode only the selected picture had the focus-peaking lines displayed. I have to fly over all pictures to have the focus peaking appearing. Probably something to fix.

aurelienpierre:
Conflict fixed. For the light table issue, the feature calls `_force_expose_all(self);` to re-expose it, so I guess that method doesn't do what it says.

TurboGit:
Ok, merged, the expose all was working properly but you had a mismatch for the accels names.

PR #3722, opened 2019-12-10
[WIP][RFC] Add support for MIDI controllers mapping to sliders

dterrahe:
2020/03/27:
I've made significant changes and uploaded a pre-compiled DLL that can be used on Windows with version 3.0.1. This module can now also interpret key presses to send commands, so there is no need for external programs anymore. It comes pre-configured for X-Touch mini and BeatStep. Please find a new introduction, also explaining the changed way to link controllers to sliders, here:
https://github.com/darktable-org/darktable/pull/3722#issuecomment-605129593

Original Comment:
Based on the midi-controller module in gimp (also GPL3); obviously I'd like to give credit for this in the most appropriate way.

Supports oss and alsa. No windows support yet. There's some support for multiple simultaneous controllers, each with their own configuration file. The midirc file name includes the device model name, when using alsa. Under oss, it tries to find a controller id by sending a Universal Device Inquiry and parsing a response. Only 50% of my devices respond to this query. Tested with Arturia Beatstep and Behringen X-Touch Mini. Relative controllers with all kinds of encodings are supported and absolute ones too. No high resolution NRPN, since I don't have anything to test with. Updates are sent back to the controller on image change and mouse slider moves, so indicator lights and absolute limitless rotators or motorised sliders stay in sync. Non-motorised ones need to be moved to the current position of the linked slider before they lock on. An indicator will show whether that is up or down. If one doesn't have a physical controller, http://vmpk.sourceforge.net/ can be used to simulate an absolute one. It is easiest to create a new Spin Box controller using Edit/Extra Controllers so that single steps can be sent.

There is no configuration UI at all yet. With current hardcoded midi_devices string, it will try to grab /dev/midi[1-4] _and_ open an alsa port. Alsa controllers can be connected using aconnect(gui); both source and destination if updates need to be sent back to the device. But if those same devices are also corresponding to /dev/midix, the alsa connection will fail. If multiple devices are connected via alsa, then the name of the last one will determine the config file name. Also, if they are sending on the same channel, then there will obviously be interference (in both directions).

To connect a controller to a slider:
1. Press Ctrl-M
2. Move a slider in one of the IOP modules that has a dynamic accelerator linked to it (unfortunately that is not all of them).
3. Move the controller knob up one notch
4. Move the knob down one notch.
Step 3&4 allow to determine whether it is an absolute slider (the values received will differ by exactly one) or which kind of relative encoding we are dealing with; one notch down could be 127, 63, 31, 15, 65, 33 or 17. Fast moves will send -2, -3 etc (or one of the other encodings and depending on the controllers acceleration settings). If multiple events for the same knob are received, they are aggregated first before updating the slider. If two knobs are turned simultaneously, this will still lead to flooding of the history stack...
Using the note keys it is possible to switch between different pages with their own configuration of sliders; the selected page will be shown in the hinter area at the top of the screen while the note is pressed. The first note pressed will be linked to page 0 (the default) and afterwards higher notes will select subsequent pages. The same slider can be linked to multiple controllers (both physical and on pages). On the beatstep, it is easy to switch channels so that could be used as an alternative to pages. On the X-touch mini, there are two banks, A and B, that can be switched with buttons, but the "invisible" bank will not receive updates and therefore get out of sync.

The code will need quite a bit of review. For one thing, there are probably better solutions then I used to capture feedback from the bauhaus sliders. For all sliders that have an accelerator, the value-changed signal is connected. This is done when DT_SIGNAL_VIEWMANAGER_VIEW_CHANGED is received and the new view is darkroom. There may also be memory leaks all over the place.

Todo:
Configuration of ports via preferences
Alsa connection by hw address
Dynamic (dis)connection of multiple controllers via Alsa
Pop-up controller board layout display
Windows support
Collect default layouts for different controller models and include in distribution

Linking notes to actions is not implemented at all. Technically easier than controllers, but configuration should be integrated fully in the existing accelerator framework.
In the mean time, notes can be translated to key presses with external programs that pass through everything else to darktable:
https://gitlab.com/enetheru/midi2input
Windows:
http://75r.de/midikey2key/
https://www.bome.com/products/mtclassic
http://charlie-roberts.com/midiStroke/


parafin:
@houz

TurboGit:
Really nice! I think @aurelienpierre had looked at this at some point, or was discussed with him... That was long time ago.

dterrahe:
Thanks @TurboGit !
I did see some discussions indeed a few years ago, but it seemed to have gone quiet so I thought I'd give it a go...

The initial approach I have chosen was to be as unobtrusive as possible; everything is in a self-contained separate module. With the addition of default templates in /share for the most common affordable devices (the x-touch mini currently is £35) this could be really working out of the box. However, at that point it would be useful to have note keys working as well (for ratings and previous/next especially), because fiddling with separate programs for that would form a barrier to entry. That would require a lot more configurability within the accelerator framework (especially with multi-controller support). So I'd like to agree on a path forward first before messing with that.

What I have now already WorksForMe, except that I mainly use the windows port and I have only started looking at how to get it to work there. Maybe portmidi will be helpful, even though it doesn't look very active. Other suggestions for a simple cross platform gtk3/midi library would be very welcome.

houz:
Please don't merge anything like that, adding another layer of stuff. Instead we need to unify bauhaus vimkeys, Lua callbacks, key accels, and dbus binding into one coherent set of actions in the background and make those existing systems tap into that as frontends. Then MIDI can be just another one. It's a really big task, but the only sane way to go forward.

dterrahe:
> Please don't merge anything like that

Absolutely! This is nowhere near ready for merging. But as a POC it is working remarkably well, so I hope it will provide motivation for all the work that is needed to provide this functionality in a sensible way.

> adding another layer of stuff

The current implementation just piggybacks on the existing dynamic accelerators for sliders in iop modules. It does add separate configuration files and a configuration method based on selecting the actual slider you want to map, rather than going into a tab in the settings dialog. It is all a dirty hack at the moment, but I find that approach working very conveniently. If it can be supported in a clean generic way for defining keyboard shortcuts as well, that could be nice.

Building a completely generic way to configure all kinds of simultaneous input methods is indeed a big job and that's why I wanted to come here for directions (or to help deciding on those together) before putting something together that would be thrown away anyway.

> Instead we need to unify bauhaus vimkeys, Lua callbacks, key accels, and dbus binding into one coherent set of actions in the background and make those existing systems tap into that as frontends. 

I hadn't heard about vimkeys (and from a quick look at the source still have no idea what functionality it provides), did not know dt has dbus bindings and thought lua was only intended for simple scripting, but that is clearly wrong. So I have a lot more to learn...

> Then MIDI can be just another one. 

This definitely applies to programmable note key presses; if we ignore the velocity, they should be dealt with in the same way as ordinary keyboard presses. There's some added complexity because multiple devices can be hotplugged and have different layouts. So if someone decides to use an 88-key piano at home, they'd want to configure that separately from a little x-touch mini they take with them on the road.

What I mean to say is we should think about and define supported use cases.

Sliders and knobs I think should just map to bauhaus sliders, not any other programmable action. They are somewhat different because the absolute ones do need feedback/initialisation as well to stay in sync with the slider on screen. I don't know if going through a generic flexible layer will be performant and easily configurable enough for that. 

Is it possible to link lua into the gtk3 event loop? Or how could midi events otherwise be delivered?

> It's a really big task

I'd like to help with that. Do you have any concrete plans that I could take pieces of? Or a high level design/approach?

dterrahe:
Added Windows support via portmidi (required).

This pre-compiled module for 3.0 can be dropped in C:\Program Files\darktable\lib\darktable\plugins\lighttable
[libmidi.zip](https://github.com/darktable-org/darktable/files/4127707/libmidi.zip)

Also tweaks, especially for X-Touch mini.

houz:
I started writing down some design ideas once, but never got around to finishing it. :-(

dterrahe:
Realistically, this doesn't sound like it is going to happen, since it is too big and complex a task for me to attempt without buy-in and knowing what the ideas of those most involved in this aspect of the project are.

I think it is useful to have this available as a use-case if ever somebody does find the time to redesign the input layer, so I'll continue to push my changes to my "midi" branch. I don't expect too many issues in keeping this compatible with future versions, since it does compile as a stand-alone module.

Should I close this pull request?

dterrahe:
Attached a compiled module dll for Windows, updated for 3.0.1
[darktable midi.zip](https://github.com/darktable-org/darktable/files/4314720/darktable.midi.zip)

To get this to work, place libmidi.dll in C:\Program Files\darktable\lib\darktable\plugins\lighttable and libportmidi.dll in C:\Program Files\darktable\bin

If you have a Behringer X-touch mini, you can use the other files in the zip for a useful base.

midirc-X-TOUCH MINI goes in your darktable configuration directory (C:\Users\...\AppData\Local\darktable)

To get some useful key mappings, you can route the midi signal through Midi Translator Classic first. Install from here https://www.bome.com/products/mtclassic and also install LoopBe1 from https://www.nerds.de/en/loopbe1.html to feed the output from Midi Translator into darktable. Run the .reg file in the zip to set the correct default input/outputs. In Midi Translator load X-touch mini.bmtp

If you have compiled the branch under linux, you can use m2i from https://gitlab.com/enetheru/midi2input to map keys. Run from a directory that contains config.lua and x-touch.lua
In aconnect, or aconnectgui, feed the output from x-touch into m2i and the output from m2i into darktable. The output from darktable goes back to the x-touch, so that it can control lights and set the controller positions.

johnny-bit:
@houz - can you start a thread about this on discuss.pixls.us so we could discuss and get input on it? It seems like a better place for draft ideas ;)

dterrahe:
Just made a quick video of the current functionality for those without a MIDI controller lying around:
https://youtu.be/I5gB4VtyNlg

johnny-bit:
OK, this IS awesome. We do need that new input system in dt.

Makotophoto:
I did some few test on a BCF2000 controller (on windows) : it's working quite well. I don't see why loopbe1 is needed, since mtclassic did not intercept all the midi data for me. 
I miss a feedback moving motorised fader and changing knob value.
I could note a strange behavior when playing with temperature and exposition (and with others parameter) : changing one make the other "un-linked" with his controller. 

Many thanks for all your work.

dterrahe:
> I did some few test on a BCF2000 controller (on windows) : it's working quite well. 

Thanks for testing and providing feedback. Especially for motorized faders, because I don't have access to one of those.

> I don't see why loopbe1 is needed, since mtclassic did not intercept all the midi data for me.

Did you find another way to feed the midi data into both mtclassic and darktable? For me, if one grabs the channel, the other one is denied access. Hence why I chained them together in series via loopbe1

> I miss a feedback moving motorised fader and changing knob value.

Do both input and output midi devices get opened in darktable? You should see in your log file (mine is at C:\Users\Diederik\AppData\Local\Microsoft\Windows\INetCache\darktable\darktable-log.txt) something similar to:
Portmidi device name: LoopBe Internal MIDI
Portmidi output device name: X-TOUCH MINI

> I could note a strange behavior when playing with temperature and exposition (and with others parameter) : changing one make the other "un-linked" with his controller.

This isn't a problem on the x-touch midi as long as there is bi-directional communication since it immediately reconnects. If it doesn't work as well on the BCF2000 then more advanced housekeeping may need to be done to decide when to drop the link. One of the complications is that more than one physical slider can be connected to the same screen slider.

Makotophoto:
Ok, I made some more test this morning and finded some config error.
First, mtclassic previously detected 3 midi in and 3 midi out for my BCF : it's probably why the mesages did pass through and messed some things. 

But to regain control of fader config through a app named BC Manager ([https://mountainutilities.eu/bcmanager](url)), I should have to change the usb port for my controller, and now i only have 1 midi in and 1 midi out so, now i need loopBe1.
The "strange behaviour" i mentioned before is rarely happening now. 

> Do both input and output midi devices get opened in darktable? You should see in your log file (mine is at C:\Users\xxxx\AppData\Local\Microsoft\Windows\INetCache\darktable\darktable-log.txt) 

Mine is empty. Is there some option to give to darktable ? As far as I could see with BC Manager (if he really can monitor Loop), LoopBe1 don't receive any data from darktable when mooving a slider withe the mouse.

dterrahe:
The _output_ from darktable should directly go to the BCF2000, _not_ via loopbe1, because that would cause a short circuit. Or fail to open the port. You have to set the default input and output for darktable using PmDefaults (download here https://sourceforge.net/projects/portmedia/files/portmidi/217/ )

On my system, the correct settings look like this:
image

Makotophoto:
Thanks, with this, the feed back is working. Some few note: 
  - every change of a parameter seem to update all the controllers, it could be a good idea to limit to the corresponding midi chanel (i'm earing my faders motors being activated each time i'm changing a parameter)
  - the tint parameter, assigned to a rotative knob don't like extreme value of 0 or 127: some jump of value happend after that. I may have a hint related to this, with the temperature parameter. For this last one, when the rotative knob had reach 0, continuing to turn it permit to decrease further the temperature value. Workaround : mapping on fader instead of rotary knob.

I'm looking forward to your future releases, and if you need some information on the bcf or a default config used with lightroom, just ask me.

dterrahe:
> * every change of a parameter seem to update all the controllers, it could be a good idea to limit to the corresponding midi chanel (i'm earing my faders motors being activated each time i'm changing a parameter)

I had hoped that the motors would not engage if the controller was already in the right position. I'll see if with a bit of extra bookkeeping I can avoid sending some updates. Then it becomes more important to implement something to detect if synchronisation was lost; this can happen for example on the x-touch mini when the user presses the A button twice (which resets all controllers to 0) or switches between A and B rows. So that configuration should really not be used.

> * the tint parameter, assigned to a rotative knob don't like extreme value of 0 or 127: some jump of value happend after that. I may have a hint related to this, with the temperature parameter. For this last one, when the rotative knob had reach 0, continuing to turn it permit to decrease further the temperature value.

Not sure I understand exactly what you mean here. The code tries to calculate relative changes from the rotating knobs, even if they send absolute values, by sending back the actual value of the screen slider after a 1 step change (rather than calculating the step as 1/127th of the full range). This works surprisingly well on the x-touch mini and I'd hope that the BCF2000 acts the same way, but maybe it doesn't.

Makotophoto:
I did made some more tests :
  - first with a BCF2000 you get more than one midi port if the controller is configured with something other than USB-1 (other mode is "U-2", "U-3", and "U-4", who link physical port with the USB port)
  - then, I investigating what is send back from darktble to the controler, using MIDI-OX software. But, for that, i needed to reverse the midi chain: BCF -> PortMIDI -> Loopbe1 -> Xtouch -> BCF.
The result is in the attached image.
DT_midi problem

dterrahe:
Thanks @Makotophoto. Would you please send me the midirc file that was generated for your BCF2000? (in the C:\Users\xxxx\AppData\Local\darktable directory)

Makotophoto:
Here the generated file. The reverse version is the same, just the name changing in "midirc-LoopBe Internal MIDI".
[midirc-BCF2000.zip](https://github.com/darktable-org/darktable/files/4389812/midirc-BCF2000.zip)

dterrahe:
Please find a new release for 3.0.1 here:
[darktable_midi.zip](https://github.com/darktable-org/darktable/files/4394965/darktable_midi.zip)

Extract the zip file, preserving paths, into C:\Program Files\darktable and restart darktable.
It contains settings templates for X-Touch mini, Arturia Beatstep and (untested for now) BCF2000. If anybody is interested I would like (to help) them to contribute templates for additional devices.

This latest version contains significant changes and also requires a changed file format for the settings, so please delete any existing midirc file in C:\Users\xxx\AppData\Local\darktable so a new one will be generated from defaults.

It is no longer necessary to use another program to interpret key presses since this is now implemented in this module. It simply translates a midi Note message to a gtk key press code. This needs to be edited directly in the midirc file, since there is no configuration ui.

If you have previously set up routing via Bome Midi Translater/loopbe1, you should use PmDefaults ( https://sourceforge.net/projects/portmedia/files/portmidi/217/ ) to set both input and output to the same midi controller:
image
For x-touch mini,  beatstep or BCF2000 you can use one of these .reg files to set this up without needing PmDefaults:
[portmidi config.zip](https://github.com/darktable-org/darktable/files/4394957/portmidi.config.zip)

On the x-touch mini only layer A should be used; the code can’t detect switches between A and B and double pressing one of these buttons resets all dials to 0 and loses the connection with the on-screen slider. Use pages (below) to assign different layouts to the dials.

The following functionality is implemented and configured by default. In the darkroom:

The lower row of buttons select different pages (1-8). The button for the selected page lights up and while the button is pressed the top area of the screen shows the configured sliders.

The upper row of buttons can be configured in midirc and send different keys depending on the page.
On page 1: buttons 1-5 change the rating (and the corresponding number of buttons light up). Button 6 rejects, button 7 goes to previous image, 8 to next image.
On page 2: buttons 1-5 switch on/of color labels (and corresponding button lights up), 6/7/8 as on page 1.

To link a slider to a dial:
1. press Ctrl-M
2. move the slider on screen with the mouse 
3. carefully, slowly, turn the dial _left_ (_down_).
This is a change from previously (when it needed to be moved up and then down). The code looks for two 1 step decrements and determines the encoding from that (either absolute or one of a number of different relative encodings). If the dial is turned too fast or in the wrong direction, the mapping either fails or the wrong encoding is used (and the slider might not move or jump). Just repeat the mapping steps.

On the x-touch mini, the lights around the dial will use a different pattern depending on whether the slider goes from 0-100, -x to +x or some other range. It may be possible to achieve this on the BCF2000 as well, but it doesn’t respond to the same midi note messages to achieve this and I have not yet found any documentation that describes the correct ones. If it is just a matter of a different channel and different codes then this can be configured in the midirc file.

On the x-touch mini the controller knobs can be pressed to reset the value of the slider they are linked to. Again, this may work on the BCF2000 if the configuration is correct, but so far I have mostly guessed those values.

Once linked, knobs can be slowed down or sped up. Press Ctrl-M and turn knob left (to half acceleration) or right (to double it).

Look at the video to see this working https://youtu.be/I5gB4VtyNlg?t=45 (but ignore the configuration steps at the start because those are not relevant anymore)

Makotophoto:
@dterrahe i'm working on you're files to make it work. But what is the meaning of the parameter "num_rotators=" ?

dterrahe:
@Makotophoto num_rotators sets the number of knobs that the code will try to initialise. According to the documentation the BCF2000 has 4 virtual groups of 8 knobs, so I've set num_rotators to 32. I've also set LED_ring_behavior_channel to -1, which disables any attempt to set the pattern of LED lighting around the knobs. This is what I think previously caused the problem you had with using the knobs, since the channels for this pattern setting and for the value setting were the same. On the X-Touch mini the value gets set on a different channel (10) so there was no conflict.

Are your results any better now than before?

Makotophoto:
Here a fonctional config for BCF2000. You will need to config your controler with BC Manager ( [https://mountainutilities.eu/bcmanager](url) ). I've added a little tutorial to do it.
[midirc-BCF2000.zip](https://github.com/darktable-org/darktable/files/4397557/midirc-BCF2000.zip)
[BCF_DT.zip](https://github.com/darktable-org/darktable/files/4397559/BCF_DT.zip)

@dterrahe Ok, the ring behavior solved the problem. I changed the "num_rotator" to 8, since the group change you implemented could make it confusing. And so, i replaced the rotator group with your group. For imformation, the config look like this, now:
Config BCF DT
And a standard Lightroom config with midi2LR look like this
Config BCF Lightroom

dterrahe:
Many thanks for this @Makotophoto I really appreciate your help. From the manual it already looked like the BCF2000 is endlessly configurable. Thanks for documenting how to get it set up so that it works with the midirc file. I had been thinking that it might be useful to extend the code and model specific configuration file so that it sends a sysex message that would configure the machine correctly. But the risk is that someone accidentally loses the configuration they painstakingly had set up for something else.

With your midirc, does pressing a rotator knob correctly reset the associated slider? If I read your specs correctly, I think reset_knob_key should be set to 33.
And just to confirm; do the rating/color/page lights light up correctly as well?

The only thing to figure out would be how to set the pattern of the lights around the knobs depending on whether they are 0-100 or +/-. I know that it can be done via BC manager, but maybe it is possible by sending midi messages too?

Makotophoto:
It seems that I made a mistake in the midirc config file: all the "toneequal" should be had a value of 1.0000 (0.0000 or 0.5000 make darktable freeze for few seconds)
Pressing a knob reset it if "reset_knob_key=33" and "first_knob_key=1"
I don't know if it's possible, but we may do the same with the faders and the button above (note 73 to 80) ?

Rating/color/page are lighting up corectly. There is just one minor thing with the pages buttons: if one of page 2 to 8 is selected when launching darktable, it will stay light up until selecting the coresponding page.

I don't know if it's possible to change the light pattern through midi message. I don't know if you played with BC Manager, but here the possibilities : 
knob config

One last thing: in the BCF preset i implemented something when changing preset: it send "note 103" for requesting a parameter update from the software. Just in case... :)

fermit:
Great work. Thanks for it.
The knob mapping is a joy and very simple to set up.

Here is some feedback for the buttons, personal perspective.
> 
> The lower row of buttons select different pages (1-8). The button for the selected page lights up and while the button is pressed the top area of the screen shows the configured sliders.
>
I find a little overwhelming having 8 pages. I can't / want remember so many button configurations. 
I don't think I will ever need more then 2,3 pages, max 4. 
Would it be possible to limit the number of pages? 
I would rather use the remaining buttons of the lower row for generic stuff, like navigation, etc, what you already did in the preset on the upper row.

>  
> The upper row of buttons can be configured in midirc and send different keys depending on the page.
>
As I understand the keyboard shortcuts are used, but I had no success on mapping anything new.
For example I mapped "/" on to keyboard for CW rotation, add it to the midirc (automatically created in  C:\Users\xxx\AppData\Local\darktable), but it doesn't work using the xTouch Mini.

Could you please help clarifying how to do this mapping? Are there any constraints?

dterrahe:
Thanks again @Makotophoto 

> (0.0000 or 0.5000 make darktable freeze for few seconds)

I haven't tested 0 but am surprised a little because this is just used to multiply the size of the slider move with. Definitely 0.5 should work.

> Pressing a knob reset it if "reset_knob_key=33" and "first_knob_key=1"

Will set that in the version I will include.

> I don't know if it's possible, but we may do the same with the faders and the button above (note 73 to 80) ?

Currently the code doesn't have the flexibility to support fader/rotator blocks with different offset reset button codes.

> Rating/color/page are lighting up corectly. There is just one minor thing with the pages buttons: if one of page 2 to 8 is selected when launching darktable, it will stay light up until selecting the coresponding page.

Yeah, I'm aware of this. Basically I'm just not initialising the page button lights :-) Purely cosmetic but I might fix it if it annoys me enough ;-)

> I don't know if it's possible to change the light pattern through midi message. I don't know if you played with BC Manager, but here the possibilities :
> knob config

I saw those settings, but since I don't actually have a BCF2000 connected, BC Manager does not activate the button to send the settings to the device. Can you intercept what data it sends and check what commands you get for the different LED patterns?

> One last thing: in the BCF preset i implemented something when changing preset: it send "note 103" for requesting a parameter update from the software. Just in case... :)

Not sure what you mean with this?

Makotophoto:
> One last thing: in the BCF preset i implemented something when changing preset: it send "note 103" for requesting a parameter update from the software. Just in case... :)

I've read that you can't uses the A/B preset on the x-touch mini due to a lack of updated data. This could be a workaround (at least on the BCF) but i'm not sure it's really usefull since we already have 8 pages of config.

For the led paterns, i can't intercept the data. But i'm not sure it will be usefull, as we should rewrite the memory with bc manager (and i could only see what should be some acknoledgement sysex message). If you know the messages to send to the xtouch, i could try them on the bcf.

> Currently the code doesn't have the flexibility to support fader/rotator blocks with different offset reset button codes.

If i map the fader next to the rotator, and do the same with the button, could it be a good idea ? and then wich parameter should be modified ? "num_columns" ? "num_rotators" ?

dterrahe:
> I've read that you can't uses the A/B preset on the x-touch mini due to a lack of updated data. This could be a workaround (at least on the BCF) but i'm not sure it's really usefull since we already have 8 pages of config.

I see what you are saying. The code would need to support a key to request (re)sending updated data. As you say, it is probably not urgently needed for now.

> For the led paterns, i can't intercept the data. But i'm not sure it will be usefull, as we should rewrite the memory with bc manager (and i could only see what should be some acknoledgement sysex message). If you know the messages to send to the xtouch, i could try them on the bcf.

On the x-touch, sending a Controller message on channel 0, with key 1 to 8 and velocity one of 1, 2 or 4 sets the patterns I'm interested in. Normal Controller messages that send the value (either from the x-touch to darktable or back) get sent on channel 10, so there is no conflict. However, the BCF2000 sends normal messages on channel 0. Maybe, as you say, it only uses sysex messages for configuration (this would be the more "correct" midi thing to do). And maybe it can only send a complete sysex configuration, that dumps all the settings (from BC Manager for example) in one go and it might be impossible to only change one setting for one controller at a time. This would need experimenting...

> If i map the fader next to the rotator, and do the same with the button, could it be a good idea ? and then wich parameter should be modified ? "num_columns" ? "num_rotators" ?

Yes, that should work; if you map each of the buttons that match the rotator/fader to the same number as the rotator/fader (so rotators go 1-8, faders 9-16, pressing rotator is 1-8 and button above fader is 9-16, you could have first_knob_key=reset_knob_key=1 and num_columns=16. num_columns is impacting different things so this is not completely clean.

dterrahe:
> Would it be possible to limit the number of pages?

You could set num_columns=5. This also impacts some other things, so if you set it to 4, not all 5 lights would switch on if rating=5.

> I would rather use the remaining buttons of the lower row for generic stuff, like navigation, etc, what you already did in the preset on the upper row.
> 
> Edited: Key shortcuts work just fine.

Great. I guess you figured out you have to use key names (for example "slash","minus") rather than symbols ("/", "-"). See https://github.com/linuxmint/gtk/blob/master/gdk/keynames.txt

I actually couldn't get shortcuts for "orientation" to work, even on the normal keyboard...

fermit:
> I actually couldn't get shortcuts for "orientation" to work, even on the normal keyboard...
Try angle rotation with 360 speed/acceleration. Works like a charm.

Makotophoto:
[BCF_DT_v2.zip](https://github.com/darktable-org/darktable/files/4405112/BCF_DT_v2.zip)
Here an updated config files for the BCF2000. The mains change are :  
  - the buttons over the faders are reseting them
  - pages limited to the 4 buttons under the display

This file also contain the corresponding midirc base config.
Since it's not using a workaround for the fader reset, there is some little bug: the help text on the top of the screen is only on one ligne, and it may rezise the screen (with french and page 2 selected, the right part of the windows disapear), like the following image:
eat


Note: in the BCF config file for BC manager, **preset 2** is only here for print purpose, and **is not a working config**.

dterrahe:
Thanks @Makotophoto I merged your changed config file. At some point the other configuration steps should be added to a wiki or something.

The help screen was just meant as a temporary stopgap solution. If this ever gets included in dt, I'd hope to implement a nicer help screen (like "h") that shows the complete layout (maybe overlaying a picture of the actual controller) that also displays note keys.

Another thing that I would like to implement then is a separation between the direct control of the sliders (including feedback to the physical controller) and the note key/command and configuration parts like pages. The latter could live in lua and would then be configurable by the user. It would then be responsible for sending configuration updates (the currently selected set of sliders) to the C module. The advantage of this would be that we could support very different kinds of layouts. Like the Loupedeck+ which has some sliders that have "pages" (hue/sat/lum) and others that do not switch pages. But for this to work, more of the functionality of dt needs to be exposed to lua, which is where the whole "fundamental review of the input layer" comes into play.

jpg54:
Hi, I just resumed testing but I have a runtime problem on Linux Mint 19.3.
image

dterrahe:
In latest version you don't need midi2dt anymore, since the module now provides a similar simple mapping from midi note to key press itself.

What controller are you trying to get to work?

jpg54:
I don't know which controller I use and how do I know or which one I have to install? Thank you for your answer.

dterrahe:
I mean the physical device you are connecting to your computer. There's some support for x-touch mini, beatstep and bcf2000. Others would require you to set up your configuration yourself.

jpg54:
I have the X-touch mini Behringer.

dterrahe:
In that case, if you compile my branch, there is a good chance you could have a working configuration.

If you compile with the portmidi library installed, it'll try to connect to a midi device (hopefully your x-touch mini) at startup. If not, you manually need to attach alsa devices using aconnect or aconnectgui. Both input and output must be connected.

jpg54:
Thanks, okay, I'll try to compile your branch.

jpg54:
The compilatThe compilation is ok:
image
I'll try to work with it. Thank you.

jpg54:
Oups, j'ai oublié de compiler avec portmidi.

dterrahe:
You don't need m2i anymore.

In your screenshot, when you ran aconnect -l, was darktable running?

You should then see something like this:
aconnect -l
client 0: 'System' [type=kernel]
    0 'Timer           '
    1 'Announce        '
client 20: 'X-TOUCH MINI' [type=kernel,card=1]
    0 'X-TOUCH MINI MIDI 1'
        Connecting To: 128:1[real:0]
        Connected From: 128:0
client 128: 'Client-128' [type=user,pid=94313]
    0 'port-0          '
        Connecting To: 20:0
    1 'port-1          '
        Connected From: 20:0[real:0]
client 129: 'darktable' [type=user,pid=94313]
    0 'darktable MIDI  '

In this case, portmidi has set up an alsa client Client-128 and automatically connected to the x-touch.
If you don't have portmidi, then you have to manually connect the x-touch client to the darktable client.

jpg54:
I just recompiled. I feel like there's a problem:
image

jpg54:
Yes, darktable master is active.

dterrahe:
You need my "midi" branch of darktable.

Forget about midi2input. You don't need it anymore.

jpg54:
Ha, okay, I'll compile it.

jpg54:
I can't make git clone :+1 :+1 :
image/(image)

jpg54:
so I made a zip that I decompressed in the directory without being able to continue with git submodule init

dterrahe:
use
git clone https://github.com/dterrahe/darktable.git
and then
git checkout midi

jpg54:
It's a long shot:
image

dterrahe:
Don't know why your compiler is using stricter settings than mine, but you could replace the offending lines with:
midi.c:389
    if (!write(g_io_channel_unix_get_fd (midi->io), buf, 3))
      return;
and
midi.c:1223
     if (!write(g_io_channel_unix_get_fd (midi->io), inquiry, 6))
        return TRUE;

jpg54:
Hello, thank you for your answer, I have just made the changes and I have new errors. I put you the start of the build to see the compiler using:
image
and error codes:
image

jpg54:
I have just corrected the errors caused by the copy/paste. Build ok. Yes:
image
Thank you.

jpg54:
If I don't exaggerate too much, how do I know the assignment of the buttons and potentiometers on the X TOUCH mini?
You intervened on daktable.fr : https://forums.darktable.fr/showthread.php?tid=3803&pid=39678#pid39678

dterrahe:
The buttons on the bottom row (MC etc) switch between pages. While you hold a page button down, the top of the screen shows the mapping of the potentiometers.

The second row of buttons can be mapped in midirc-X-TOUCH\ MINI. There is no UI.

The defaults in /opt/darktable/share/darktable/midi/midirc-X-TOUCH\ MINI are:
Page 1: 1-5 assign rating, 6=reject, 7=previous image, 8=next image
Page 2: 1-5 assign color (F1-F5), 6,7,8 as on page 1.

jpg54:
Here's the only thing I get from moving the potentiometers:
image
Button presses have no action. Here is the midirc-X-TOUCH MINI file:
[midirc-X-TOUCH MINI.txt](https://github.com/darktable-org/darktable/files/4469101/midirc-X-TOUCH.MINI.txt)

dterrahe:
Please show aconnect -l

jpg54:
https://github.com/darktable-org/darktable/pull/3722#issuecomment-612817484

dterrahe:
You mean you haven't reconnected the x-touch directly to darktable? You need to connect both in and out, otherwise the midi module won't know what config file to load. It now probably tried to load "midirc-Midi Through".

After setting up correct links using aconnect, reload module using Ctrl-M Ctrl-M and go out and back into darkroom.

jpg54:
Don't I get it? To be a problem of French translation.

dterrahe:
https://sourceforge.net/p/portmedia/wiki/PmDefaults/

jpg54:
I don't see how to connect X TOUCH mini as input and darktable as output:
image

jpg54:
Hello, looking at the screenshots of the terminal, I understand that XTOUCH Mini is connected in and out as well as darktable.
image
With aconnectgui

dterrahe:
Screenshot_20200414_094517

jpg54:
Problem connecting the console to darktable:
image

dterrahe:
Try adding this line to ~/.config/darktable/darktablerc
plugins/lighttable/midi/devices=alsa

jpg54:
Win, device wasn't informed:
image

jpg54:
Yes:
20200414_134715r

jpg54:
Do you think it will be possible to have it with the new version 3.0.x?
Thank you for your friendly and patient help.

dterrahe:
Very unlikely:
https://github.com/darktable-org/darktable/pull/3722#issuecomment-564629002
https://github.com/darktable-org/darktable/pull/4581#issuecomment-606864120

dterrahe:
Because of a change in master I'll have to make some incompatible changes to make sure toasts keep popping up for hidden sliders. Therefore I have set up a separate fork of the 3.0.x branch with the midi patches applied.

https://github.com/dterrahe/darktable/tree/midi-3.0.x

Let me know if there are any problems compiling on linux. The pre-compiled Windows dll I provided earlier for 3.0.1 still seems to work with the released binaries for 3.0.2, so I'm not updating that unless I hear of problems.

dterrahe:
finally rebased my midi branch against current master, after getting around problems with submodules, and switched to using the new toasts.

TheSirC:
@dterrahe  Thank you a lot for this port ! 
_Is there some documentation available for the format of the files (like [data/midi/midirc-Arturia BeatStep](https://github.com/dterrahe/darktable/blob/b510de99e52ae0d3b59818f8022d27fb0444291b/data/midi/midirc-Arturia%20BeatStep)) that enables support for controllers ?_ 

I have a controller ([the eagletone tiny control](http://www.eagletone.com/home-studio/m.a.o/controleurs/p122691-tiny-control) for reference) that is not supported but that is recognized (the output in the command line gives back values for all buttons and slider except the bank changer and the channel mixer; both sending `MIDI: sysex start` and `MIDI: sysex end`. The UI giving `knob x on channel 0 not mapped  in group 1`, x being the number associated with the relevant slider number.

So I was wondering what all the options for the other controllers where linked to and how to adapt it to mine.

dterrahe:
> @dterrahe Thank you a lot for this port !

Thanks for your interest! There is no documentation except for the information in this thread, the source code and the settings files. The module is not supported in any way and even if eventually it makes its way into dt proper, the configuration options will almost certainly be very different.

> So I was wondering what all the options for the other controllers where linked to and how to adapt it to mine.

Most information is here:
https://github.com/darktable-org/darktable/pull/3722#issuecomment-605129593

To define slider links, press Ctrl-M, then move a slider on-screen and then turn the knob or move the slider that you want to associate with it. A new configuration file will be saved in your darktable .config directory, so you could then analyse that.

To support multiple pages, if you could tell me what key codes and channels you receive when pressing any of the buttons, I could suggest the settings to edit for those.

Please be aware that it looks like your rotators are absolute rather than relative (and obviously the sliders are) so you'll always have to move them to match the current value of a dt slider when you move to a new picture before it gets picked up and then afterwards further moves will translate to adjustments in dt.

TheSirC:
> To define slider links, press Ctrl-M, then move a slider on-screen and then turn the knob or move the slider that you want to associate with it.

That's exactly what I did : the knobs are mappable but not the sliders for some reason.

> Please be aware that it looks like your rotators are absolute rather than relative (and obviously the sliders are)

Indeed you are right : both knobs and sliders are absolute but I can not understand why one would work and not the other :sweat_smile: 

> To support multiple pages, if you could tell me what key codes and channels you receive when pressing any of the buttons, I could suggest the settings to edit for those.

I thought the change of bank would not only affect the buttons but also the sliders and knobs :confused: 
Anyway, the key codes and channel appearing in the command-line do not seem to change when I press on the bank button but here are the codes  : 
```
MIDI (ch 00): controller 1 (value 127)
MIDI (ch 00): controller 1 (value 0)
```

and the value 1 can take values in those ranges : 1-2, 23-31, 44-49 or 64 and 67 depending on which button I press.

Thanks a lot for helping !

dterrahe:
The product page you linked to suggests that it is possible to individually enable or disable knobs or sliders and to switch buttons between sending a note signal or control code. However, you'll need the software to do it, and for me the download link doesn't work.

If you _do_ have the software, you should set the buttons to send _note_ signals, as you can then use them to either change configuration pages or translate to a keyboard code that will be send to darktable. I guess the module could be made more flexible to also deal with control codes being sent, but at the moment I'm not putting further development effort into it until a path towards integration into dt is clear.

dterrahe:
Reopened as #6830 since this one has been force-pushed several times since closing, so it can't be reused.

dterrahe:
@Makotophoto in case you are still interested in using dt with the BCF2000, #12375 is for you.

PR #3725, opened 2019-12-11
Set image aspect ratio on import or update

### LINKED ISSUE #3724: sort by aspect_ratio PAINFULLY slow
Bugs report just saying it crashes without filling most of the items
below and making no effort to help developpers to reproduce will be
closed without notice.

**Describe the bug**
Aspect ratio is not calculated on import.  Therefore sorting more than a few images by aspect ratio is painfully slow because the aspect_ratio has to be calculated for each image and the database updated.

**To Reproduce**
1. Import 2500 images with mixed orientation
2. Click on the sort the sort by drop down and select aspect ratio
3. Go fix dinner, then eat it.  Come back and check to see if the sort is done...  :-)

**Expected behavior**
Most sorts take less than a second even for large image sets.  Aspect ratio sorting should be no different.  The second time you sort by aspect ratio, it responds as quickly as any other sort.

 - OS: Ubuntu 18.04
 - Version current git
 - OpenCL activated
 - Nvidia with nividia-390 drivers

**I have a fix**
I modified dt_image_cache_write_release() to calculate the aspect_ratio if it doesn't exist and add it to the database when it writes the rest of the image information.  An added benefit to this is that any time you open an image in darkroom that doesn't have an aspect_ratio, one is calculated and added.  This will help some with the images that already exist in the database without an aspect_ratio.  PR on the way.

wpferguson:
Added check aspect_ratio to dt_image_t.  Set aspect_ratio = 0.0 in image init.  Added check to see if aspect_raio exists and if not calculate it so that it's saved in dt_image_cache_write_release().

Fixes #3724

### DIFF
--- a/src/common/image_cache.c
+++ b/src/common/image_cache.c
@@ -219,43 +219,51 @@ void dt_image_cache_write_release(dt_image_cache_t *cache, dt_image_t *img, dt_i
       uint32_t u;
   } flip;
   if(img->id <= 0) return;
+  if(!img->aspect_ratio || img->aspect_ratio < .0001)
+  {
+    if(img->orientation < ORIENTATION_SWAP_XY)
+      img->aspect_ratio = (float )img->width / (float )img->height;
+    else
+      img->aspect_ratio = (float )img->height / (float )img->width;
+  }
   sqlite3_stmt *stmt;
   DT_DEBUG_SQLITE3_PREPARE_V2(
       dt_database_get(darktable.db),
-      "UPDATE main.images SET width = ?1, height = ?2, filename = ?3, maker = ?4, model = ?5, "
-      "lens = ?6, exposure = ?7, aperture = ?8, iso = ?9, focal_length = ?10, "
-      "focus_distance = ?11, film_id = ?12, datetime_taken = ?13, flags = ?14, "
-      "crop = ?15, orientation = ?16, raw_parameters = ?17, group_id = ?18, longitude = ?19, "
-      "latitude = ?20, altitude = ?21, color_matrix = ?22, colorspace = ?23, raw_black = ?24, "
-      "raw_maximum = ?25 WHERE id = ?26",
+      "UPDATE main.images SET width = ?1, height = ?2, aspect_ratio = ROUND(?3,1), filename = ?4, maker = ?5, model = ?6, "
+      "lens = ?7, exposure = ?8, aperture = ?9, iso = ?10, focal_length = ?11, "
+      "focus_distance = ?12, film_id = ?13, datetime_taken = ?14, flags = ?15, "
+      "crop = ?16, orientation = ?17, raw_parameters = ?18, group_id = ?19, longitude = ?20, "
+      "latitude = ?21, altitude = ?22, color_matrix = ?23, colorspace = ?24, raw_black = ?25, "
+      "raw_maximum = ?26 WHERE id = ?27",
       -1, &stmt, NULL);
   DT_DEBUG_SQLITE3_BIND_INT(stmt, 1, img->width);
   DT_DEBUG_SQLITE3_BIND_INT(stmt, 2, img->height);
-  DT_DEBUG_SQLITE3_BIND_TEXT(stmt, 3, img->filename, -1, SQLITE_STATIC);
-  DT_DEBUG_SQLITE3_BIND_TEXT(stmt, 4, img->exif_maker, -1, SQLITE_STATIC);
-  DT_DEBUG_SQLITE3_BIND_TEXT(stmt, 5, img->exif_model, -1, SQLITE_STATIC);
-  DT_DEBUG_SQLITE3_BIND_TEXT(stmt, 6, img->exif_lens, -1, SQLITE_STATIC);
-  DT_DEBUG_SQLITE3_BIND_DOUBLE(stmt, 7, img->exif_exposure);
-  DT_DEBUG_SQLITE3_BIND_DOUBLE(stmt, 8, img->exif_aperture);
-  DT_DEBUG_SQLITE3_BIND_DOUBLE(stmt, 9, img->exif_iso);
-  DT_DEBUG_SQLITE3_BIND_DOUBLE(stmt, 10, img->exif_focal_length);
-  DT_DEBUG_SQLITE3_BIND_DOUBLE(stmt, 11, img->exif_focus_distance);
-  DT_DEBUG_SQLITE3_BIND_INT(stmt, 12, img->film_id);
-  DT_DEBUG_SQLITE3_BIND_TEXT(stmt, 13, img->exif_datetime_taken, -1, SQLITE_STATIC);
-  DT_DEBUG_SQLITE3_BIND_INT(stmt, 14, img->flags);
-  DT_DEBUG_SQLITE3_BIND_DOUBLE(stmt, 15, img->exif_crop);
-  DT_DEBUG_SQLITE3_BIND_INT(stmt, 16, img->orientation);
+  DT_DEBUG_SQLITE3_BIND_DOUBLE(stmt, 3, img->aspect_ratio);
+  DT_DEBUG_SQLITE3_BIND_TEXT(stmt, 4, img->filename, -1, SQLITE_STATIC);
+  DT_DEBUG_SQLITE3_BIND_TEXT(stmt, 5, img->exif_maker, -1, SQLITE_STATIC);
+  DT_DEBUG_SQLITE3_BIND_TEXT(stmt, 6, img->exif_model, -1, SQLITE_STATIC);
+  DT_DEBUG_SQLITE3_BIND_TEXT(stmt, 7, img->exif_lens, -1, SQLITE_STATIC);
+  DT_DEBUG_SQLITE3_BIND_DOUBLE(stmt, 8, img->exif_exposure);
+  DT_DEBUG_SQLITE3_BIND_DOUBLE(stmt, 9, img->exif_aperture);
+  DT_DEBUG_SQLITE3_BIND_DOUBLE(stmt, 10, img->exif_iso);
+  DT_DEBUG_SQLITE3_BIND_DOUBLE(stmt, 11, img->exif_focal_length);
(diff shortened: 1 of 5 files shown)


AlicVB:
hmm, I'm not sure, but if you add aspect_ratio to dt_image_t struct, it may be wise to rework all part in the code where this is retrieve, recomputed, ... from database or cache...
for example here : https://github.com/darktable-org/darktable/blob/8319719bab971f1d7da5b5156928bd449592f841/src/common/image.c#L624
and also ensure that the image struct is updated each time you recompute the ratio...

wpferguson:
I went back and looked.  The dt_image_set_aspect_ratio() is a database update to add an aspect_ratio value to the database so that sort by aspect ratio works.  I searched the code for every place aspect_ratio was used and almost every case was a check to see if aspect_ratio was null or < .0001.  There was a comment in collection.c about eventually all images in the database would have aspect_ratio set (or at least that was the goal) by doing repeated database queries for aspect_ratio equal to NULL or 0.0, then calculating and populating the field based on requests to sort by aspect ratio.

This update sets the aspect_ratio field on import, so that it doesn't have to get fixed later by dt_image_set_aspect_ratio().  As a bonus, aspect_ratio will also get updated if you preview the image, are culling, open the image in darkroom, or run darktable-generate-cache (and probably a few other ways, but those are the ones I tried).

wpferguson:
Another data point.  This also works if you re-import an existing film roll.  I imported 3089 images and the import took ~50 secs.  The aspect_ratio was updated for all images (it was null before).

TurboGit:
@wpferguson : this is not correct as the aspect ratio is not the one of the RAW file but the one of the image and possible cropped with different aspect ratio.

To test, with ypur path:
- select sort by aspect ratio in lighttable
- open an image in darkroom, crop square
- go back to lighttable, the image is not seen as square

Select by "aspect" in collection module and you'll see that it is not listed in square filter.

This scenario works without your patch.

Also, note that it is slow ok, but only the first time one select the aspect ratio filter. After that the values are cached into the database.

AlicVB:
I'm not really sure that what is proposed here will not work at all, but it need some more work for sure. Here's how I see what is needed : 
- on import with no history, set ratio as in this PR
- on import with history (xmp), set aspect-ratio to 0.0, to force recomputing.
- remove actual aspect_ratio_xx functions
- add a set_aspect_ratio fct which will update database and image struct (eventually the recomputing can be done here too)
- add a get_aspect_ratio fct which will return the value in struct or recompute if set to 0.0

What do you think ?

ptilopteri:
you will probably need to keep two aspect ratio, one for the original raw and one for the expected output image.  and sorts would need to specify which, raw or output image

TurboGit:
Looks quite complex to me for something that is not really an issue and will certainly makes the maintenance of this code delicate.

Sorting on aspect the first time on a collection of about 250 pictures is almost unnoticeable. And there is a time out in the computation, that is you won't be locked for more than some seconds (don't remember the timeout). I've been using this feature regularly without issue on my side.

That being said, you you think it can be made to be clear on the code I'll be ok to have a look of course.

wpferguson:
@TurboGit how many images in your database?  The behavior you describe for sort by aspect ratio is nothing like I experience.  darktable goes non-responsive and the operating system offers to kill it.  I've learned to just tell it to wait and eventually it gets done.  Almost every time I do a sort by aspect ratio, it's minutes before I get a response.  But, my image database currently has 474,361 images.

TurboGit:
@wpferguson : how strange, looks like another issue to me.

First, do you see the aspect ratio in the database properly updated? The aspect ratio should be computed ONLY is set to 0 which is the value set at import time.

I have 50.000 pictures and it is instantaneous now that all aspect ratio have been computed.

TurboGit:
Hum... Just tried again and indeed it takes lot of time... I'll look at this, a "recent" regression to me.

wpferguson:
I tried a fresh database and sorted my play raw directory of ~230 images and it only took a second or two.  Then I imported 672 images with the sort still at aspect ratio and it took minutes, because it imported the images and then went back and updated them one by one.

I tried importing the same filmroll on a fresh database with sort by filename, then sorting by aspect ratio.  I got the timeout message at 3+ minutes and control returned to the gui.

TurboGit:
@wpferguson : and when filtering on aspect ratio after, does it takes lot of time?

TurboGit:
@wpferguson : The first time it is expected to be long as one need to render the image to get the aspect ratio.

TurboGit:
But 3+ minutes delay is too long and furthermore there is no message.

wpferguson:
@AlicVB it sounds like I have some learning to do :).  It also sounds like a good approach.

@TurboGit I agree this is a lot more complex than it seemed at first look.  I'm willing to keep working at it, but it'll take awhile (but I'll be more knowledgeable when I'm done :)).  If I can't come up with a clean way to do it, then I'll scrap it.

Another thought, is there a way we can buffer up the SQL update statements and run them as a transaction?  To me it seems there is a threshold for number of images to update.  If you stay below that, then sort by aspect ratio is quick, but once you exceed it, then it slows down dramatically.

wpferguson:
@TurboGit, the first time is the only time it's slow.  After that it's as quick as any other sort by field.

TurboGit:
@wpferguson : ok so we are seeing the same thing (fast after first update).

Now, let's see if we can improve the initial speed. And yes, you'll learn the code working on this. Thanks!

wpferguson:
**What I did**

I left the import alone, so on import the raw aspect_ratio is calculated and stored.  Next I made the assumption that dt_image_set_aspect_ratio_to() and dt_image_reset_aspect_ratio() work correctly, so I just changed them to get the timage from dt_image_cache, set image->aspect_ratio, then do a dt_image_cache_write_release to write it to the database instead of just writing it directly.

**What that did**

This fixed the changed aspect ratio in the crop and rotate module.  It also sets the aspect ratio as you scroll through a film roll generating thumbnails, which solves the problem of importing from xmp.    In other words even though I may not set the aspect_ratio correctly on import depending on processing, the aspect_ratio will get corrected during use.  And, if you are scrolling through the film roll, with the sort set to aspect_ratio, it will be resorted on the fly.

**What I figured out**

Between runs I would go in the database and set the aspect_ratio to NULL.  Sometimes I got quick sorts, but most of the time it was slow.  I tried wrapping the sort in a transaction which made the quick sorts instantaneous, but didn't help with the slow sorts.  I finally figured out that the culprit was the mipmap cache.  So, if all the images were in the mipmap cache then the sort was quick.  But, if the images were not in the mipmap cache, the each image was read from disk, then processed, then the aspect_ratio was calculated and stored.  Therefore, if you import a new film roll the display updates and the first few images are loaded into the mipmap cache.  If you do a sort by aspect_ratio, then the rest of the images need to be loaded and processed before the sort can complete.  I also noticed that the processing got slower every 100 images or so.

**My solution**

I created dt_image_set_raw_aspect_ratio() which sets the aspect_ratio from the dt_image_cache which is populated.  I replaced the call to dt_image_set_aspect_ratio() in collection.c with dt_image_cache_set_raw_aspect_ratio().  I sorted 2875 images in ~6 seconds.  I got the timeout error, but I checked the database and all the aspect ratios were set.  The aspect ratio is set from the raw data, so it might not be correct for all images, but they will get corrected as the thumbnails get built.

One thing I did discover during testing.  If you leave the sort set to aspect ratio, then close darktable, set the aspect_ratio to NULL in the database, then start darktable you get a really big crash.  That was because I was trying to read from the dt_image_cache, but it wasn't populated yet.  I don't think this will be a problem during normal usage because the only way I could created it was by manipulating the database and setting the aspect_ratio to NULL while darktable wasn't running.

**Thoughts?**

I'll clean up the code and push it, so you can see it.  Any other things I should look at/fix?

wpferguson:
It appears that my comment about the aspect_ratio updating as the thumbnail was updated is incorrect.  So It seems I still have some work to do

wpferguson:
Fixed dt_image_load_job_run so that it can calculate the processed image aspect ratio.  Added dt_image_set_aspect_ratio_if_different to update the database's idea of the aspect ratio if it was different from the one calculated in dt_image_load_job_run.  So now it works as I described earlier.

I tested a large film roll with a few random images cropped to square.  Loaded all and set the raw aspect ratio.  As I scrolled through the collection the square image aspect ratios were updated and the images were placed in the proper place in the collection.

I tested taking a raw image and cropping it to square, with the sort by field set to aspect_ratio.  When I returned to lighttable, the image was put in the correct order in the sort.

I imported multiple film rolls, stopped darktable, set the aspect_ratio to null, started darktable and sorted each by aspect ratio.  It took a second or two on average to assign the raw aspect ratio and do the sort.

I imported film rolls with sort by set to aspect ratio.  It takes a little longer than importing with a different sort order because every time an image is imported _collection_update_aspect_ratio is called and it sets the raw aspect ratio.  I thought about checking which sort by method was selected then setting it to filename if it was set to aspect ratio, doing the import, then setting it back to aspect ratio.  I managed to trace down where film rolls were imported, but I wasn't able to figure a way to wrap the import process reliably.  If anyone has ideas about how to wrap it, I'll play with it some more.  In the meantime, I'll think about it some more...

Any more tests anyone would like performed?

wpferguson:
Fixed the longer import with sort by aspect ratio.  dt_image_set_raw_aspect_ratio was signalling that the collection had changed after updating each image.  I removed the signal and now import speed is on par with all the other sort by filters.

@TurboGit I think this is complete.

src/common/image_cache.c
```diff
@@ -218,6 +222,13 @@ void dt_image_cache_write_release(dt_image_cache_t *cache, dt_image_t *img, dt_i
       struct dt_image_raw_parameters_t s;
       uint32_t u;
   } flip;
+  if(!img->aspect_ratio || img->aspect_ratio < .0001)
```
TurboGit: This is equivalent to:
if(img->aspect_ratio==0 || img->aspect_ratio < .0001)
so to:
if(img->aspect_ratio < .0001)

Right?

TurboGit:
Ok, tested and looks good. I'll fix the test mentioned above after merging. Thanks.

PR #3736, opened 2019-12-13
add a colour assessment mode in darkroom (white frame on middle grey …

### LINKED ISSUE #3686: Add support for ICC recommended (ISO 12646:2008) colour proofing conditions
**Référence:**  http://www.color.org/resources/r8-13/3.20170809_Consistent_Colour_Appearance_assessment.pdf

Screenshot_2019-12-07 Standard images - 3 20170809_Consistent_Colour_Appearance_assessment pdf

**Proposal:**
In lighttable and darkroom, add a white frame around pictures on top of a grey background (L = 50) for the zoomed-out/full preview image (about 10-15% width of the picture):
Capture d’écran du 2019-12-07 19-13-11
That could be a preference or even a toggle on/off in darkroom (ISO 12646-compliant mode).

**Benefit:**
The white frame gives a visual reference to assess the global contrast in image. The overall grey backgound gives a reference for the average exposure.

**Note**
As usual, I'm basing this on indirect info I have found on the ISO standard, not directly the standard since you have to pay for it to get the PDF.

aurelienpierre:
…background)

fix #3686 

Add a toggle button in darkroom (the light bulb):
toolkit

Before, the image might look ok because of the dark background:
Capture d’écran du 2019-12-13 03-09-12

Toogle the ISO 12646 colour assessment conditions, and you immediately see that your whites are actually grey:
Capture d’écran du 2019-12-13 03-09-46

The feature is only adding a white frame and forcing the background colour of the darkroom to middle grey. The frame persists while zooming. This overcomes the Hunt and Bartleson–Breneman effects that affect colour and contrast perception in dim lit surrounds.

Example on grey theme:
Capture d’écran du 2019-12-13 03-22-50

Zoomed-in:
Capture d’écran du 2019-12-13 03-25-07

### DIFF
--- a/src/common/focus.h
+++ b/src/common/focus.h
@@ -278,7 +278,7 @@ static void dt_focus_draw_clusters(cairo_t *cr, int width, int height, int imgid
     dt_dev_cleanup(&dev);
   }
 
-  const int32_t tb = DT_PIXEL_APPLY_DPI(dt_conf_get_int("plugins/darkroom/ui/border_size"));
+  const int32_t tb = darktable.develop->border_size;
   const float scale = fminf((width - 2 * tb) / (float)wd, (height - 2 * tb) / (float)ht) * full_zoom;
   cairo_scale(cr, scale, scale);
   float fx = 0.0f;
--- a/src/develop/develop.c
+++ b/src/develop/develop.c
@@ -127,6 +127,8 @@ void dt_dev_init(dt_develop_t *dev, int32_t gui_attached)
 
   dev->proxy.exposure = NULL;
 
+  dev->border_size = DT_PIXEL_APPLY_DPI(dt_conf_get_int("plugins/darkroom/ui/border_size"));
+
   dev->rawoverexposed.enabled = FALSE;
   dev->rawoverexposed.mode = dt_conf_get_int("darkroom/ui/rawoverexposed/mode");
   dev->rawoverexposed.colorscheme = dt_conf_get_int("darkroom/ui/rawoverexposed/colorscheme");
@@ -136,6 +138,8 @@ void dt_dev_init(dt_develop_t *dev, int32_t gui_attached)
   dev->overexposed.colorscheme = dt_conf_get_int("darkroom/ui/overexposed/colorscheme");
   dev->overexposed.lower = dt_conf_get_float("darkroom/ui/overexposed/lower");
   dev->overexposed.upper = dt_conf_get_float("darkroom/ui/overexposed/upper");
+
+  dev->iso_12646.enabled = FALSE;
 
   dev->second_window.zoom = DT_ZOOM_FIT;
   dev->second_window.closeup = 0;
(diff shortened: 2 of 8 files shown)


TurboGit:
Nice! Thanks.

Nilvus:
Nice work @aurelienpierre. I know why using white frame but anyway I just already see some users asking to choose the color... showing they don't understand the purpose !

TurboGit:
@aurelienpierre : I'm wondering if in this mode all the tabs should be closed by default, just letting the image alone? Or is it expected to work on the image with this mode activated? If so maybe just masking the top, left, bottom tab? Or even better, as for the lt culling mode, remember the last layout when leaving this mode? Thoughts?

aurelienpierre:
@TurboGit We have to keep at least the right sidebar to allow working on the picture while in colour assessement mode (there is not much interest to it if it's only a preview mode). Since there are shortcuts now to hide every panel and bar, I would leave it to users to decide what they want to collapse.

TurboGit:
Ok, I understand for letting user to collapse whatever but my point was to keep record of the layout in color assessment mode independently of the normal edit mode to avoid having to manually collapse/expand whatever user find necessary between the two modes. This is was we do for the culling vs standard lightttable mode today, very handy.

MStraeten:
on osx this pr causes a segfault in darktable-cltest:

lldb output:
```
1.217686 [opencl_create_kernel] successfully loaded kernel `guided_filter_generate_result' (41) for device 1
1.217708 [opencl_priorities] these are your device priorities:
1.217710 [opencl_priorities] 		image	preview	export	thumbnail	preview2
1.217712 [opencl_priorities]		1	0	1	1	1
1.217714 [opencl_priorities]		0	-1	0	0	0
1.217715 [opencl_priorities] show if opencl use is mandatory for a given pixelpipe:
1.217717 [opencl_priorities] 		image	preview	export	thumbnail	preview2
1.217718 [opencl_priorities]		0	0	0	0	0
1.217722 [opencl_synchronization_timeout] synchronization timeout set to 200
libdarktable.dylib was compiled with optimization - stepping may behave oddly; variables may not be available.
Process 35979 stopped
* thread #1, queue = 'com.apple.main-thread', stop reason = EXC_BAD_ACCESS (code=1, address=0x370)
    frame #0: 0x00000001001f00cb libdarktable.dylib`dt_dev_init(dev=0x0000000105899c00, gui_attached=<unavailable>) at develop.c:130:22 [opt]
   127 	
   128 	  dev->proxy.exposure = NULL;
   129 	
-> 130 	  dev->border_size = DT_PIXEL_APPLY_DPI(dt_conf_get_int("plugins/darkroom/ui/border_size"));
   131 	
   132 	  dev->rawoverexposed.enabled = FALSE;
   133 	  dev->rawoverexposed.mode = dt_conf_get_int("darkroom/ui/rawoverexposed/mode");
Target 0: (darktable-cltest) stopped.
(lldb) 

```

this is darktable 2.7.0+2578~g474d1476e-dirty

compile options:
  bit depth is 64 bit
  normal build
  OpenMP support enabled
  OpenCL support enabled
  Lua support enabled, API version 5.0.2
  Colord support disabled
  gPhoto2 support enabled
  OpenEXR support enabled

TurboGit:
I'm preparing a "next" branch for after the release. I have integrated this PR into it, fixed the crash and added support for keeping the panels' state for darkroom's standard editing mode and color assessment one.

aurelienpierre:
Great, thanks @TurboGit.

Egocentrix:
Thank you for adding this view aurelienpierre! Makes it a lot easier to judge exposure. 

Sorry for bumping the PR. If it is preferred to open a new issue for this I will do that, but I have a request: currently it is not possible to set a key shortcut for the new mode, as it is for the overexposure and softproofing views. Is it possible to add this?

i have tried looking into this myself, but the dt codebase and the accels framework is intimidating.

PR #3737, opened 2019-12-13
toneequalizer: default params to make people shut up

aurelienpierre:
Disable exposure and contrast compensations in default params, and use only one iteration of guided filter. 

This will give so-so masking results for everybody, instead of the current defaults that give good results in maybe 60% of use cases and are completely unpredictable in the remaining 40%. I'm getting married in 10 days, at that point I'm tired of hearing people whining about defaults params.


GLLM:
Priceless PR title !!
Oh, and congrats .... :-)

TurboGit:
I'm not ok with the log comment :) I'll update before merging as I don't want this to be in dt history.

TurboGit:
Merged manually. Thanks.

PR #3741, opened 2019-12-13
cholesky.h: fix a possible pointer bug

aurelienpierre:
2 non-heap buffers where declared with `DT_ALIGN_ARRAY` which should be reserved for heap buffers or may cause issues.

### DIFF
--- a/src/iop/choleski.h
+++ b/src/iop/choleski.h
@@ -393,8 +393,8 @@ static inline int pseudo_solve(float *const restrict A,
     return valid;
   }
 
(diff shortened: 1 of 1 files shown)


parafin:
One should use DT_ALIGNED_ARRAY or dt_alloc_sse_ps. Declaring a pointer with DT_ALIGNED_ARRAY makes pointer being aligned, not data. Using both DT_ALIGNED_ARRAY and dt_alloc_sse_ps at the same time doesn't make sense. So this change is correct (though I'm not sure which issues it may cause).

TurboGit:
Thanks parafin. Merging then.

Issue #3721, opened 2019-12-10
Failed Intel OpenCL compilation prevents darktable to use NVIDIA OpenCL on Windows 10


infjb:
The CPU of my system is an Intel i5-3570k with a HD Graphics 4000 GPU. The graphics card is a NVIDIA RTX 2060 Super.

The operating system is Windows 10. Drivers are the latest I could get from Intel and NVIDIA.

Darktable finds the NVIDIA GPU, compiles the OpenCL kernels and puts them into the caching directory.

Then it tries to compile the kernels for the Intel GPU and fails. 

At that moment darktable decides that the system has no working OpenCL device and declares that OpenCL is not available.

darktable 3.0.0rc2 and 2.6.2 (which I removed from the system when I installed 3.0.0rc1) are both behaving in the same way.

I did not find out the reason why that fails, yet. I would not mind the fact that it fails to use the Itel GPU. The NVIDIA is far more powerful. 

In my opinion darktable should use the NVIDA GPU even if it fails to use the Intel GPU.

Any hints why compiling the the kernels for the Intel GPU are failing are also welcome. I'm suspecting the BIOS settings at the moment.

The output of darktable-cltest is as folows:

`
PS C:\Program Files\darktable\bin> .\darktable-cltest.exe
0.153950 [opencl_init] opencl related configuration options:
0.155949 [opencl_init]
0.156949 [opencl_init] opencl: 1
0.157948 [opencl_init] opencl_library: ''
0.158948 [opencl_init] opencl_memory_requirement: 768
0.160947 [opencl_init] opencl_memory_headroom: 300
0.161947 [opencl_init] opencl_device_priority: '*/!0,*/*/*'
0.163947 [opencl_init] opencl_mandatory_timeout: 200
0.165946 [opencl_init] opencl_size_roundup: 16
0.166945 [opencl_init] opencl_async_pixelpipe: 0
0.167945 [opencl_init] opencl_synch_cache: false
0.169945 [opencl_init] opencl_number_event_handles: 25
0.170944 [opencl_init] opencl_micro_nap: 1000
0.172944 [opencl_init] opencl_use_pinned_memory: 0
0.174943 [opencl_init] opencl_use_cpu_devices: 0
0.175942 [opencl_init] opencl_avoid_atomics: 0
0.176942 [opencl_init]
0.184940 [opencl_init] found opencl runtime library 'OpenCL.dll'
[...]

PR #3742, opened 2019-12-13
opencl: do not fail completely if one of several devices cannot be initialized

upegelow:
this mostly affects multi-GPU systems:
the old logic has been to disable opencl completely if any device could
not be initialized. there have been different error types, some fatal in
the above sense others not. this has been changed: now if a device cannot be
initiatlized opencl can still be used if at least one device initializes
successfully.

this addresses #3721


TurboGit:
Looks good! Thanks.

Issue #3674, opened 2019-12-06
Contrast equalizer: lags when dragging 


ClaesGitHub:
In dt 3.0.0rc2+66~gee083f665
Expand contrast equalizer.
Grab one of the nodes (indicated by a circle)
using left-mouse-click and drag it up then down
then up again. The motion and update is not fluid (as expected),
but very jumpy. openCL on or off makes no difference.

**Platform:**
Latest Manjaro/KDE. Ryzen 3900X. GTX-1660. Nvidia.

AxelG-DE:
With OpenCL working, I cannot confirm. Very smooth. CPU (9900K) loads 14-20% and nvidia-smi shows me 85% on device0 and ~35% on device1.

Then I started dt with --disable-opencl. CPU is loaded ~85% and the behaviour is like you explained above.

Are you sure, your opencl works really propperly? (watch nvidia-smi)
I have had performance issues, when you had an driver update but not yet rebooted, for instance

ClaesGitHub:
> Does my openCL work properly?
I believe that it does work correctly. Using the dt benchmark my present clockings are 2.879 seconds (with openCL) versus 8.404 seconds (without openCL).

AxelG-DE:
As you have a specific issue now, I would recommend not to use that benchmark to check, as it uses darktable-clip, which potentially could react differently.

So I suggest, you open a terminal window, start _watch nvidia-smi_ and observe it, while you drag the equaliser node.

ClaesGitHub:
Nice idea :-) Thank you -- but the output looks normal to me.

AxelG-DE:
Darned...
May I kindly ask, what you consider "normal" in that respect? 

Why I feel some similarities to #2724? Just now it is you instead of me? :-/  :)

aurelienpierre:
The contrast equalizer, like the curves, redraw the graph after the pipe is recomputed, and the pipe is recomputed every time you touch the settings. It's just a matter of inverting the sequence. In tone equalizer, the graph is drawn before the pipe is recomputed, maybe you can test and see if it feels more responsive.

upegelow:
> The contrast equalizer, like the curves, redraw the graph after the pipe is recomputed, and the pipe is recomputed every time you touch the settings. It's just a matter of inverting the sequence. In tone equalizer, the graph is drawn before the pipe is recomputed, maybe you can test and see if it feels more responsive.

[...]

Issue #3574, opened 2019-11-26
Image information → Geolocation


ghost:
Display GPS direction/heading in image information both in lighttable and darkroom modes. Do not show lots of empty lines if location data is missing, 'geolocation is not available' or something similar will be enough.

PR #3750, opened 2019-12-14
proof of concept: delayed reprocessing on parameter change

upegelow:
do not trigger reprocessing more frequently than the pixelpipe can handle.
this is an idea how to tackle issues like #3674

t.b.d.

### DIFF
--- a/src/develop/develop.c
+++ b/src/develop/develop.c
@@ -60,6 +60,7 @@ void dt_dev_init(dt_develop_t *dev, int32_t gui_attached)
   dev->average_delay = DT_DEV_AVERAGE_DELAY_START;
   dev->preview_average_delay = DT_DEV_PREVIEW_AVERAGE_DELAY_START;
   dev->preview2_average_delay = DT_DEV_PREVIEW_AVERAGE_DELAY_START;
+  dev->timeout_handle = 0;
   dev->gui_leaving = 0;
   dev->gui_synch = 0;
   dt_pthread_mutex_init(&dev->history_mutex, NULL);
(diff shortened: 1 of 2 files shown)


TurboGit:
I'll test that, seems a nice improvement. IIRC Aurélien (or was it Edgardoh?) did try something like that with a different implementation and it was creating issues. Anyway, having a delay and sending a signal only when there is no redraw already in the queue is something we want to see.

upegelow:
Frankly, I am not sure if this is the way to go. Actions on curves are a bit less bumpy now but also far from being smooth. We could delay more and even delay starting the pixelpipe when its idle but there is the nasty side effect of a less snappy pixelpipe. We do a lot all over the place to make image processing as responsive as possible but then voluntarily delay it here. Does not sound clever.

As the issue seems to be linked to curve manipulation (while sliders seem to act smoothly) we better try to fix the issue at the source, not in this central location.

upegelow:
I have just checked in the tonecurve module. It's sufficient if we call `gtk_widget_queue_draw()` before `dt_dev_add_history_item()` to make everything smooth. I do not see any disadvantage so this should be done all over the place where modules have complex control elements.

We can close this one.

aurelienpierre:
I beg to differ. The mouse wheel actions, especially when using a touchpad emulation of the scroll, trigger a pipe recompute at the beginning and the end of the scroll event (to reproduce, just try to zoom in with scroll). I tried to insert a time out there, it didn't work. But I think there should be some high level regulation of pipe recomputations to avoid dt freezing on mouse event settings input.

upegelow:
A combination might be a good concept. Suggestion for the contrast equalizer in #3574 and it seems to work well.

PR #3754, opened 2019-12-14
gui responsiveness: improve for contrast equalizer

upegelow:
Delay dt_dev_add_history_item() when nodes are moved.

This works really well with a fast GPU and is a clear improvement on CPU.

Other candidates would be:
basecurve
colorzones
lowlight
rawdenoise
rgbcurve
tonecurve

### DIFF
--- a/src/iop/atrous.c
+++ b/src/iop/atrous.c
@@ -96,6 +96,7 @@ typedef struct dt_iop_atrous_gui_data_t
   float band_max;
   float sample[MAX_NUM_SCALES];
   int num_samples;
+  int timeout_handle;
 } dt_iop_atrous_gui_data_t;
 
 typedef struct dt_iop_atrous_global_data_t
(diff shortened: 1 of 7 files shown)


upegelow:
This certainly relates to a bug as this is negatively affecting user experience.

TurboGit:
This introduce a common pattern in many places. Do you think we could come up with some kind of code sharing or proxy a bit like what was done for color pickers ?

upegelow:
> This introduce a common pattern in many places. Do you think we could come up with some kind of code sharing or proxy a bit like what was done for color pickers ?

Hmm, the most important element is the timeout_handle variable which is private to each individual module and sits in a structure (gui data) which is different from module to module. So we can not write a generic timeout handler function. Then there are the currently hard coded scaling factors which derive the delay from the average delay time of the currently running pixelpipe. They are all the same right now but there could be some need for a per-module adaptation as well.

TurboGit:
Very smooth now indeed! I merge that for 3.0. Thanks a lot.

PR #3760, opened 2019-12-15
[user manual] translation of newly added functions and chapters for dt 3.0

mepi0011:
The German user manual is 100% translated. Please add for dt 3.0.


TurboGit:
Thanks! Sqashing and merging.

Issue #3591, opened 2019-11-27
cleaner dng support


TurboGit:
A DNG file can contain either a RAW or non RAW data.

In #3565 this has been fixed but a better solution has been devised. This ticket is to track work on this for 3.0.1.

The idea is to introduce more high level routines like:

```
dt_image_is_ca_correct_supported(imgid);
dt_image_is_matrix_correction_supported(imgid);
```

And use that in IOP where needed.

After doing that the dt_image_is_raw() routine will just check the flags and this routine will be used in place where we check for the flags directly. Partly reverting #3565 

We do expect this will ease maintenance.

aurelienpierre:
See #7129

TurboGit:
Probably fixed now.

PR #3764, opened 2019-12-15
proper test for raw & dng files.

jenshannoschwalm:
Work on issue #3591

To make sure a proper raw/dng testing we have to distinguish between raw/sraw.
So introduced DT_IMAGE_S_RAW set&reset by all loaders.

Also direct testing for bits in img->flags is prone to errors so there are a
few tests available.

int dt_image_is_raw(const dt_image_t *img) now only tests the raw flag
and these two are new
int dt_image_is_matrix_correction_supported(const dt_image_t *img);
int dt_image_is_rawprepare_supported(const dt_image_t *img);

Existing modules use these functions now.

### DIFF
--- a/src/common/image.c
+++ b/src/common/image.c
@@ -107,35 +107,29 @@ int dt_image_is_hdr(const dt_image_t *img)
     return 0;
 }
 
+// NULL terminated list of supported non-RAW extensions
+//  const char *dt_non_raw_extensions[]
+//    = { ".jpeg", ".jpg",  ".pfm", ".hdr", ".exr", ".pxn", ".tif", ".tiff", ".png",
+//        ".j2c",  ".j2k",  ".jp2", ".jpc", ".gif", ".jpc", ".jp2", ".bmp",  ".dcm",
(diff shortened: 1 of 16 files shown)


TurboGit:
I have set this to 3.0.1, ok? Too much risky for 3.0 to me.

jenshannoschwalm:
Sure.

jenshannoschwalm:
Dear @TurboGit while merging and testing the massive pile of pr's there is propbably one thing you oversaw in this pr. Commit bc06a08 disables the highlight reconstruction for non-raw...

I forgot to mention explicitely :-( Please let me know how this module should behave for raw & non-raw regarding masks and enabling in your opinion. I will then do another pr fixing whatever you suggest.

TurboGit:
Indeed I missed that. I'll revert this commit. Nothing more to do at the moment as I'm not clear as to what should be done here. To me the partial work on jpeg seems ok and was there since very long time so I'm reluctant to change this without very strong arguments.

PR #3771, opened 2019-12-15
historgram : don't draw it if image is not the right one

### LINKED ISSUE #3763: Histogram → Caching
Same as #3575 but for histogram.

Steps to reproduce:

1. Start DT in lightroom mode
2. Open any file in darkroom mode
3. Leave darkroom by pressing 'l'
4. Open any other file in darkroom mode

Notice that histogram of previous image is shown (in top right corner) before current picture is loaded and modules settings applied. Probably it is cached somewhere. Just showing nothing will do fine, showing previous histogram is confusing.

AlicVB:
This fix #3763 
This the same fix as for navigation lib. but if someone with better histogram understanding can have a look...

### DIFF
--- a/src/libs/histogram.c
+++ b/src/libs/histogram.c
@@ -237,7 +237,7 @@ static gboolean _lib_histogram_draw_callback(GtkWidget *widget, cairo_t *crf, gp
     dt_draw_grid(cr, 4, 0, 0, width, height);
 
   // draw histogram
-  if(hist_max > 0.0f)
+  if(hist_max > 0.0f && dev->image_storage.id == dev->preview_pipe->output_imgid)
   {
     cairo_save(cr);
     if(dev->histogram_type == DT_DEV_HISTOGRAM_WAVEFORM)


timur-davletshin:
Come on! It's one line fix and 3.0.1

TurboGit:
One line fix, but only for a GUI glitch and can you assert that the "simple" added test does not have side effects and is always true? I can't and we are at less than 10 days from next release.

AlicVB:
thanks ;)

... and yes a "one line fix" can break things. If a more skilled dev can ensure that's it'll not trigger any weird bug, then why not. But I'm reluctant to push a fix for non-blocking bugs so close to a release !

Nilvus:
Just tested. Work on my side. This complete your PR #3683 about darkroom loading screen.
Made some tests and didn't by now any side effects. Anyway, on few days to 3.0 release, I agree with you that it's better to merge that on 3.0.1. To be careful and as it also complete your other PR, programmed for 3.0.1.

TurboGit:
Merged.

PR #3776, opened 2019-12-15
full preview : fix if selected image from same group

### LINKED ISSUE #3775: Sticky preview: fix navigation in a selection of group images
Dear developer, 
I encounter the following minor bug.
Kind regards

**Description**
The persistent preview ignores the selection when it only contains images from the same group.

**Reproducible:** always

**To Reproduce**

1. In Lighttable,  select a few images belonging to the same group
2. open the persistent preview
3. Navigate through the images

**Expected behavior**
Only the selected images should be displayed.

**Workaround:** Add an image taken outside the group to lock the selection. 

**Platform:**
 - Ubuntu 18.04 and Debian 
 - darktable 3,0rc2 (compiled src and OBS bin)
 - darktable 2.6.x

AlicVB:
this fix #3775
In order to get the screen selection, we need to combine collected_images with selected_images, as selected_images contains all images from the same group if groups are collapsed on screen.

### DIFF
--- a/src/views/lighttable.c
+++ b/src/views/lighttable.c
@@ -2581,43 +2581,33 @@ static int expose_full_preview(dt_view_t *self, cairo_t *cr, int32_t width, int3
     /* If more than one image is selected, iterate over these. */
     /* If only one image is selected, scroll through all known images. */
     sqlite3_stmt *stmt;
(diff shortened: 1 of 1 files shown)


TurboGit:
Merged.

Issue #3756, opened 2019-12-14
OpenCL / Local Contrast / Local Laplacian - issue with AMD/ROCM: 'amplified effect'


arigit:
This is an issue reported in redmine 1 year ago, didn't get any traction, bringing it here since bug reporting seems to have moved and it may be found by other people affected.

This issue affects Darktable when used with AMD-based GPUs, OpenCL and the ROCM opensource driver which is the currently supported OpenCL driver for all new AMD discrete and integrated GPUs. It's been proven to affect both Polaris and Vega GPUs.

This issue does not happen with the legacy closed-source amdgpu-pro OpenCL driver (that many still use). 

Original issue: https://redmine.darktable.org/issues/12423

**Describe the bug**

The result of applying Local Contrast > Local Laplacian is totally different with OpenCL "on" and OpenCL "off", for the exact same settings of the module. The more % of Detail is set, the more striking is the difference.

This issue does NOT happen for Local Contrast > Bilateral Grid, it is specific to Local Laplacian. When I select Bilateral Grid, the result is the same regardless of whether OpenCL is enabled or not.

It looks like the Local Contrast > Local Laplacian OpenCL implementation has a problem. See the attached JPG snapshots, taken with the exact same settings of the model, one with OpenCL "on" and one with OpenCL "off". Basically LocalContrast+LocalLaplacian is unusable (and destroys the image) when OpenCL is on.

It looks like with OpenCL enabled, the "Detail" effect of the Local Contrast Filter is like grossly amplified.

Example: 
**Local Contrast + Local Laplacian, OpenCL = "off"**
![image](https://user-images.githubusercontent.com/652673/70853064-eb3f7480-1e76-11ea-9331-0aeade2e3131.png)

**Local Contrast + Local Laplacian, OpenCL = "on"**
![image](https://user-images.githubusercontent.com/652673/70853067-027e6200-1e77-11ea-931f-934d5ddecc6e.png)

No other other OpenCL kernel from darktable has problems with the official ROCM driver from AMD in my setup (I have been using ROCM OpenCL with darktable for 1.5 years now, meaning thousands of pictures developed, using profiled denoise, filmic, retouch, basecurve, etc etc). 

My workaround in order to keep OpenCL "on" has been to remove the LocalLaplacian kernel.

```
sudo mv /usr/share/darktable/kernels/locallaplacian.cl /usr/share/darktable/kernels/locallaplacian.cl.temporarilyRemovedDueToROCM
```

[...]

PR #3782, opened 2019-12-16
Fix locallaplacian padding

### LINKED ISSUE #3753: dt3.0: lut 3D: Gtk-WARNING drawing failure for widget 'GtkWindow' & 'GtkDrawingArea'
**Describe the bug**
The combobox at first line of lut 3D module is producing two Gtk warnings when the "lut file" selection icon is not used before.
100% reproducible in this context

**To Reproduce**
1. Select an image and Go to 'darkroom'
2. Open the 'lut 3D' module
3. At the first parameter line, do not use the "lut file" icon on the left, but directly Click on the combobox on the right.
4. No choice proposed (no file listed), but See the following 2 Gtk warnings in the console:
```
(darktable:26403): Gtk-WARNING **: 17:03:45.033: drawing failure for widget 'GtkDrawingArea': invalid value (typically too big) for the size of the input (surface, pattern, etc.)

(darktable:26403): Gtk-WARNING **: 17:03:45.033: drawing failure for widget 'GtkWindow': invalid value (typically too big) for the size of the input (surface, pattern, etc.)
```
**Expected behavior**
No Gtk warning

 - OS: Ubuntu 19.10 / Gtk 3.24.12-1ubuntu1
 - darktable 3.0.0rc2+97~gb26d5128f

**Additional context**
When the "lut file" icon on the left is used before, and a lut file selected, then the list of choices on the combobox is not empty, and no Gtk warning reported

dt compiled by myself from latest git master using standard build instruction. No specific option added

**Side comment**
I do not fully understand the logic to have two different widgets (the "lut file" icon, and then the combobox) on the same line to select the lut file

aurelienpierre:
Attempt to reduce possible errors sources from #3756 for OpenCL and fix https://redmine.darktable.org/issues/12349 for SSE2 and for good.

### DIFF
--- a/src/common/locallaplaciancl.c
+++ b/src/common/locallaplaciancl.c
@@ -95,23 +95,20 @@ dt_local_laplacian_cl_t *dt_local_laplacian_init_cl(
     g->dev_processed[k] = (cl_mem *)calloc(max_levels, sizeof(cl_mem *));
 
   g->num_levels = MIN(max_levels, 31-__builtin_clz(MIN(width,height)));
-  const int max_supp = 1<<(g->num_levels-1);
-  const int paddwd = width  + 2*max_supp;
-  const int paddht = height + 2*max_supp;
-
-  const size_t bwidth = ROUNDUPWD(paddwd);
-  const size_t bheight = ROUNDUPWD(paddht);
+  g->max_supp = 1<<(g->num_levels-1);
+  g->bwidth = ROUNDUPWD(width  + 2*g->max_supp);
+  g->bheight = ROUNDUPHT(height + 2*g->max_supp);
 
   // get intermediate vector buffers with read-write access
   for(int l=0;l<g->num_levels;l++)
   {
-    g->dev_padded[l] = dt_opencl_alloc_device(devid, ROUNDUPWD(dl(bwidth, l)), ROUNDUPHT(dl(bheight, l)), sizeof(float));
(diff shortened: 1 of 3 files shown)


aurelienpierre:
For the record, the last commit deletes the special handling of the full pipeline preview in SSE2 path. What it did was waiting for the thumbnail preview to setup the laplacian transformation, then fetch the buffers and reuse them for speed-up. It lead to inconsistencies when the preview pipe finished after the full pipe (the infamous "local laplacian: inconsistent output") and possible racing conditions with the mutex.

So I have put the SSE2 path on the same page than the OpenCL and plain C pathes, and the full preview pipeline recomputes independently from the thumbnail. It doesn't seem to penalize performance.

aurelienpierre:
@TurboGit could you consider merging that for 3.0 or do you think it's too risky ?

1ceae79 is not the bugfix it should have been but stores sizes in struct instead of recomputing them everytime. It doesn't change the actual processing.

055f8c7 disables the clever handling of full preview with the info already computed and cached for the thumbnail, because there is no guaranty that the thumbnail will be computed first (when it's not, you get the "inconsistent output" message). Note that it only affects SSE2 path, and essentially makes SSE2 behaves as the plain C and OpenCL ones, for better consistency (even if some perf penalty should be expected, but my tests don't show a noticeable difference, mainly because the former code made the full pipeline wait for thumbnail mutex locks to be released).

TurboGit:
@aurelienpierre : yes too risky so close to the tag now. this is not a regression, let's wait a bit more we will have time to do proper field testing for 3.0.1.
Thanks for tracking this down!

TurboGit:
Works for me! Thanks.

PR #3784, opened 2019-12-16
filmic OpenCL: try to fix grey screen with chroma preservation

aurelienpierre:
Reimplement an internal `dot` function since it seems some drivers have issues with the built-in method.


TurboGit:
Looks ok and cannot hurt. Thanks.

PR #3785, opened 2019-12-16
drawn masks: let changes also take effect on the preview2 pipe

upegelow:
Obviously the update of the preview2 pipe was missing.

It would be interesting to further investigate if we really need DT_DEV_PIPE_SYNCH rather than DT_DEV_PIPE_TOP_CHANGED. A complete pipe synch is quite expensive as it needs a full parameter commit of all pixelpipes and an update of the complete history stack. Can take easily 40ms or even 200ms on images with a long history stack. In order to experience the slowness just change the radius of an existing circle shape.


TurboGit:
Indeed the 2nd preview was not updated when mask opacity was changed for example. Better have this in 3.0. Just in time, thanks!

PR #3789, opened 2019-12-17
filmic: fix OpenCL subscripted vector access

aurelienpierre:
fix `error: subscripted access is not allowed for OpenCL vectors` by taking .xyz members of the float4 vector, and don't use the 4th channel (alpha) which might not be set to 0 all the time.


TurboGit:
Ok, working fine as previous version. Thanks.

PR #3797, opened 2019-12-17
OpenCL: remove native functions calls

### LINKED ISSUE #3756: OpenCL / Local Contrast / Local Laplacian - issue with AMD/ROCM: 'amplified effect'

aurelienpierre:
replace native_exp, native_powr, native_log2, which are implementation-defined (depending on the driver and possibly faster) by standard OpenCL exp, powr, log2, to alleviate the inconsistencies between the 2 AMD, the 2 Intel and the Nvidia drivers.

Might be the second part to fix #3756

### DIFF
--- a/data/kernels/atrous.cl
+++ b/data/kernels/atrous.cl
@@ -24,8 +24,8 @@ float4
 weight(const float4 c1, const float4 c2, const float sharpen)
 {
   // native_exp is faster than the cpu floating point aliasing hack:
-  const float wc = native_exp(-((c1.y - c2.y)*(c1.y - c2.y) + (c1.z - c2.z)*(c1.z - c2.z)) * sharpen);
-  const float wl = native_exp(- (c1.x - c2.x)*(c1.x - c2.x) * sharpen);
+  const float wc = exp(-((c1.y - c2.y)*(c1.y - c2.y) + (c1.z - c2.z)*(c1.z - c2.z)) * sharpen);
+  const float wl = exp(- (c1.x - c2.x)*(c1.x - c2.x) * sharpen);
   return (float4)(wl, wc, wc, 1.0f);
 }
 
--- a/data/kernels/basic.cl
+++ b/data/kernels/basic.cl
@@ -447,7 +447,7 @@ lookup_unbounded_twosided(read_only image2d_t lut, const float x, constant float
       // two-sided extrapolation (with inverted x-axis for left side)
       const float xx = (x >= ar) ? x : 1.0f - x;
       constant float *aa = (x >= ar) ? a : a + 3;
-      return aa[1] * native_powr(xx*aa[0], aa[2]);
+      return aa[1] * powr(xx*aa[0], aa[2]);
     }
   }
   else return x;
@@ -471,7 +471,7 @@ lerp_lookup_unbounded0(read_only image2d_t lut, const float x, global const floa
       const float l2 = read_imagef(lut, sampleri, p2).x;
       return l1 * (1.0f - f) + l2 * f;
     }
-    else return a[1] * native_powr(x*a[0], a[2]);
+    else return a[1] * powr(x*a[0], a[2]);
   }
   else return x;
 }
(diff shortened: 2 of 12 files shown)


upegelow:
Please evaluate potential performance losses vs. the desired gains in accuracy especially if these function calls are used in inner loops.

AFAIK there is a substantial performance difference.

aurelienpierre:
For now I'm mostly worried about **losses** in accuracy, resulting in https://github.com/darktable-org/darktable/issues/3756#issuecomment-566282942. I'm not sure this is the cause for the bug, but if it is, it's no good to have fast code that runs on only 2/3 of available platforms.

upegelow:
Yepp, fine if this is found to be the cause of the bug. I do not ask you to stop that direction but take it only when justified by fixing the bug. Else we are probably better off with the native functions.

aurelienpierre:
Doesn't seem to solve the bug.

PR #3799, opened 2019-12-17
Update hu.po

BathoryPeter:
Translated the most important strings, but I'm not ready yet.

How much time do I have before the release? I try to complete as much as I can...


TurboGit:
You have 3 days left before I cut the release.

PR #3803, opened 2019-12-18
opencl: increase opencl_memory_headroom to 400MB

upegelow:
The original setting of 300MB comes from a time when typical
graphic cards had 1-2GB of memory. Modern GPUs are equipped with 4GB
or more. Over the years drivers seem to need more and more memory
for their own tasks. I have experienced situations during export
when some modules (e.g. contrast equalizer) needed to fall back
to CPU as available memory on the GPU was lower than expected
(output is fine then but takes much longer). With a memory headroom
of 400MB we make things a bit less tight.
Note: This affects the default setting when installing darktable anew.
On existing systems users will need to manually adjust darktablerc.

### DIFF
--- a/data/darktableconfig.xml.in
+++ b/data/darktableconfig.xml.in
@@ -86,7 +86,7 @@
   <dtconfig>
     <name>opencl_memory_headroom</name>
     <type>int</type>
-    <default>300</default>
+    <default>400</default>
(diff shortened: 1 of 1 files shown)


TurboGit:
I vote to have this in 3.0, do you agree ?

upegelow:
Yes, this is something we should bring into 3.0 as we might expect quite some new users who will take advantage.

TurboGit:
Merged. Thanks.

PR #3804, opened 2019-12-18
lighttable : visible_rows value can't be 0

### LINKED ISSUE #3786: Exiting darkroom changes the selected image (when you navigated in darkroom)
**Describe the bug**
Going back from the darkroom to the lightable loses the image which had focus on the darkroom. Instead, it goes back to the image I entered the darkroom from. 

**To Reproduce**
1. In the lightable, use a collection with more than 1 image. 
2. Go to the darkroom from image A (with double-click or menu) 
3. In the darkroom, change to image B 
4. Exit to the lightable (with double-click or menu)
5. You are back to image A. 

**Expected behavior**
Previously (2.6), going back from darkroom to lightable used to keep the same photo selected. On 5. above, it should stay on image B. 

**Platform**
 - OS: Linux Mint 19.1 Cinnamon
 - Version: 3.0.0rc2~git115.eff9c65e1

AlicVB:
as there's always something visible
this fix #3786

### DIFF
--- a/src/views/lighttable.c
+++ b/src/views/lighttable.c
@@ -1110,7 +1110,7 @@ static int expose_filemanager(dt_view_t *self, cairo_t *cr, int32_t width, int32
 
   const int max_rows = 1 + (int)((height) / ht + .5);
   lib->max_rows = max_rows;
-  lib->visible_rows = height / ht;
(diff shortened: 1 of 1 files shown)


TurboGit:
Merged.

Issue #3757, opened 2019-12-14
Incorrect file size on export


juhavlintula:
**Describe the bug**
A vertical image that has been cropped to 16:9 and that I want to export as 1440×2560 is exported as 1439×2560 (or as 1439×2559)

**To Reproduce**
1. Crop image to 9x16
2. Define max dimensions as 1440×2560
3. Export JPG
4. The exported image is not 1440×2560

**Platform (please complete the following information):**
 - OS: Win10 & Ubuntu 19.10
 - Version: 3.0rc2 and newer

**Additional context**
 - works correctly on 2.7.0
 - tested on Olympus RAWs

jade-nl:
Confirmed and not Olympus specific.

Tested and confirmed with: Mamiya, Nikon, Pentax and Huawei RAW's

Linux - Debian 10.2
Version latest 3.0.0rc2-102

upegelow:
This is probably a side effect of #3666.

Question out of curiosity: why are people so obsessed with exact image dimensions like +/- 1 on a 4 Megapixel image?

aurelienpierre:
I'm sorry, what is the bug ? Integer rounding ?

upegelow:
#3666 always rounds down in order to never read past the original image dimensions. As a side effect even a minor float rounding error e.g. leading to 1439.999878 results in 1439. Previously we would round to the next integer with issues (see comments in #3666)

We could consider something like:

`int i_width = floorf(f_width + eps)`

where `eps` is some small delta to account for float errors (e.g. 1e-6).

aurelienpierre:
Why not simply `int i_width = MAX(roundf(f_width), raw_width)` ?

upegelow:
It's about this part of the code:

https://github.com/darktable-org/darktable/blob/c09d36df9b01cd940bb90aea180a76618953c552/src/common/imageio.c#L756-L757

The task here is: adjust the **output** dimensions so that with all roi calculations we end up with the correct input dimensions. 

Originally that code would add 0.5f and then convert to int, effectively rounding to the closest integer value.

The problem in #3666 has been triggered by very small output dimensions, namely thumbnails. Typically in the range of 150x100 and a scale factor of 1/40. If we select the output width/height even by +1 too high we may read dozens of pixels past the input image dimensions. That's the reason for moving to the floor function.

This code needs to deal with situations of different background:

- thumbnails: process something reasonably close to the desired dimensions with a aspect ratio that roughly fits

[...]

PR #3808, opened 2019-12-18
thumb/export: better finetuning of output dimensions

upegelow:
this tries to compromise between #3757 (make output dimensions exactly as user
requests) and #3646 (do not read outside of input/raw image buffer).


TurboGit:
Thanks for working on that!

PR #3809, opened 2019-12-18
colorpicker: Fix Lab color picker samples alignment.

TurboGit:


Nilvus:
So you already make it for @Mark-64. It's ok for me and will fix the last picker glitch UI.

Mark-64:
I was about to do it, but you have been quicker :-)

TurboGit:
Sorry @Mark-64  I did not understood that you would have proposed a PR. Anyway, after your investigation is was trivial. Thanks!

PR #3810, opened 2019-12-18
Not showing hightlights reconstruction for non-raw images

jenshannoschwalm:
The highlights reconstruction module and it's gtk subs should be disabled
& not shown for non-raw images as it's done for the other only-raws.

### DIFF
--- a/src/iop/highlights.c
+++ b/src/iop/highlights.c
@@ -60,6 +60,8 @@ typedef struct dt_iop_highlights_gui_data_t
 {
   GtkWidget *clip;
   GtkWidget *mode;
+  GtkWidget *box_raw;
+  GtkWidget *label_non_raw;
 } dt_iop_highlights_gui_data_t;
 
 typedef dt_iop_highlights_params_t dt_iop_highlights_data_t;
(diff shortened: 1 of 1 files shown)


TurboGit:
But this module works together with "shadows & highlights" which can be used with Jpeg.

jenshannoschwalm:
Sorry, i was not aware that this could interfere and also can't find a hint in shadhi.c.
Probably i'm just blind right now...

jenshannoschwalm:
For me it's strange you can't enable the module via the button but still can change the threshold slider. I never thought of highlights reconstruction for jpegs or alike.
Shouldn't `hide_enable_button` be left as 0 if we want it enabled for all sort of images?

jenshannoschwalm:
@Turbogit , i think of closing this pr as i understand we want this module also for jepgs. Still; for non-raw images like jpgs we **can't** use masks right now because `hide_enable_button` is non-zero? Is this really intentional?

TurboGit:
@jenshannoschwalm : I'm not yet really clear on this, but on my side I have no support for masks on highlight reconstruction module for jpeg nor for RAW files.

TurboGit:
If you agree you can close this PR. Thanks.

jenshannoschwalm:
Closing this although if we have masks for raws we should have them for jpegs too. And the enable-module button can't be enabled for jpegs but they can use the module. For me it's not consistant.

TurboGit:
@jenshannoschwalm : I agree, I may be confused but the mask issue was not part of this PR ? Did not test it....

jenshannoschwalm:
@TurboGit this pr was about **disabling** highlight reconstruction for non-raw.

If you agree that we would rather want masks for raw and non-raw for the sake of consistency - whatever the use case is - give me a hint.

PR #3817, opened 2019-12-19
image: extends the list of iop for non altered flag.

### LINKED ISSUE #3815: Imported Images opened in darkroom mode and closed show as "Edited" even if no changes are done (DT 3.0RC3)
In the past (DT 3.0RC1 and prior stable releases), importing an image (RAW or even JPG for culling), entering the darkroom to inspect it and returning to the lighttable would NOT cause the "edited" flag (ying-yang symbol) to turn on if no changes were made in the darkroom.

Also, in the darkroom it was possible to "compress" the history stack back to the Original Image and discard everything else - this was a way to revert any editing (in addition to using "Discard History Stack" in the lighttable)

WIth RC3, any attempt to "compress" the history stack reverts to including nine (!) items in the stack, and there is no way to get the "Edited" flag to turn off (unless using "Discard History Stack" in the lighttable)

image

This seems to be a bug. In prior releases, compressing the stack down to "Original Image" and exiting the darkroom would resent the Edited flag, and upon entering the darkroom again, the "autoapplied" modules would pop up again (just baseline and orientation) but this would NOT cause the Edited flag to turn on. 

Even JPGs that are visualized in the darkroom (e.g. just to check focus / zoom) are flagged as edited upon exiting the darkroom even if not changes are done whatsoever.

Any chance to revert to the prior behavior for the Edited flag that was showing if any **real** editing was done or not to the picture?

TurboGit:
Fixes #3815.

### DIFF
--- a/src/common/image.c
+++ b/src/common/image.c
@@ -854,6 +854,12 @@ int dt_image_altered(const uint32_t imgid)
     if(!strcmp(op, "sharpen") && dt_conf_get_bool("plugins/darkroom/sharpen/auto_apply")) continue;
     if(!strcmp(op, "dither")) continue;
     if(!strcmp(op, "highlights")) continue;
+    if(!strcmp(op, "rawprepare")) continue;
+    if(!strcmp(op, "colorin")) continue;
+    if(!strcmp(op, "colorout")) continue;
+    if(!strcmp(op, "gamma")) continue;
+    if(!strcmp(op, "demosaic")) continue;
+    if(!strcmp(op, "temperature")) continue;
     altered = 1;
     break;
   }


arigit:
Built, installed and tested: this patch works fine. With this patch, entering the darkroom and exiting the darkroom without modifying the image no longer triggers the Edited flag. Also, doing any editing in the image turns on the flag as expected.
Lastly, entering the darkroom with an edited image, and rolling back the history stack by compressing it back to the "Original Image", and then exiting the image, clears the flag.

Bottom line the patch restored the correct and expected behavior. Please merge

TurboGit:
Thanks for testing. Merging.

PR #3820, opened 2019-12-20
 Release notes: Add warning about styles+ordering

junkyardsparkle:
Users should be warned about this, there are already a couple of issues posted about it. Also fixed markup and copy edited the section about mandatory modules.


TurboGit:
Good point! Thanks.

PR #3826, opened 2019-12-21
fix forgotten 2.4 -> 2.6 in release notes

sjjh:
forgot to create pull request


spaceChRiS:
It also has to be 3.0 instead of 2.8.

sjjh:
I closed this pull request, as it was unnecessary. I apparently just forgot this commit half a year ago and by now, as you said, it's 3.0 and the change is already reflected in the release notes...

PR #3838, opened 2019-12-23
views: show hide bottom panel when changing filmstrip status.

TurboGit:
This is needed as when the bottom panel is hidden, trying to show/hide
the filmstrip will do nothing. We also want to handle the visibility of
the bottom panel.


TurboGit:
Merged.

PR #3841, opened 2019-12-23
po/fix masks rotation

TurboGit:
Long standing bug in rotation for ellipse and gradient when flip is activated.


TurboGit:
Merged.

PR #3844, opened 2019-12-23
Mask Manager / label - typo correction

arigit:
Use "inverted" instead of "inversed"
The meaning of "Inversed" does not apply in this context
Note: I am not clear if this small change has any impact on Translations - a search in the code showed that 4 translations files contains the wrong string ("use inversed shape") as well.


TurboGit:
Merged.

PR #3850, opened 2019-12-24
clipping: fix straightening when flip is applied.

### LINKED ISSUE #3833: Horizontal flip followed by staright using right click in crop rotates it in the wrong direction
1. Flip an image horizontally.
2. Go to crop and use the right click to level horizon.
3. Rotates in opposite direction. The angle should be negative of what was determined as the horizon levelling still uses the actual image.

Alternate is to level before flipping.

Version : darktable-3.1.0+8~g7871a4f25-win64

TurboGit:
This has been tested with vertical, horizontal and both flip activated.

Fixes #3833.

### DIFF
--- a/src/iop/clipping.c
+++ b/src/iop/clipping.c
@@ -3039,12 +3039,18 @@ int button_released(struct dt_iop_module_t *self, double x, double y, int which,
 
   if(g->straightening)
   {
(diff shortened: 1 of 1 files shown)


TurboGit:
Merged.

PR #3851, opened 2019-12-24
watermark: update watermark text to 512 characters max.

### LINKED ISSUE #3840: Watermark text length restricted to 60 character when composited on the image
Was trying to put a text longer than 60 as watermark.
Text field of the module allows the length but when composited on the image it only displays 60. 

This limit needs to be bigger or at least configurable.

Version : darktable-3.1.0+8~g7871a4f25-win64

TurboGit:
This is important as we can now have variables in the simple text entry
and so a limit of 64 was quickly reached.

Fixes #3840.

### DIFF
--- a/src/iop/watermark.c
+++ b/src/iop/watermark.c
@@ -46,7 +46,7 @@
 #include "common/utility.h"
 
 #define CLIP(x) ((x < 0) ? 0.0 : (x > 1.0) ? 1.0 : x)
-DT_MODULE_INTROSPECTION(4, dt_iop_watermark_params_t)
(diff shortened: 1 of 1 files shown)


chhil:
Any way to determine what can  fit on the fly using font/dpi/orientation of text and image size?

TurboGit:
That won't work. This setting can be copied from image to image and moreover we are talking about IOP parameters which must have a static size.

Nilvus:
512 is already a quite long text for watermark. And much appropriate now than previous 64.

TurboGit:
Ok, I'll go with that. Thanks.

PR #3861, opened 2019-12-25
UI: make modules labels a bit smaller

aurelienpierre:
The purpose of the condensed font was to reduce the width footprint of the modules labels, which is defeated by the increased font size. Since these labels are isolated (not lost in a full block of text), I find they stay legible even at relatively small sizes. To be discussed…

### DIFF
--- a/data/themes/darktable.css
+++ b/data/themes/darktable.css
@@ -477,7 +477,7 @@ entry selection
   padding: 0 2pt 2pt 0.5em;
   font-weight: normal;
   font-family: sans-serif;
-  font-size: larger;
+  font-size: 1.1em;
 }
 
 /* Labels of controls sections in modules */


Nilvus:
I'm torn here. The current size is perfectly legible and ok for me. This new size seems a bit small (maybe a question of habit). But at the same time, this new size makes the interface more UI consistent and pleasant. Quite more beautiful !

aurelienpierre:
> But at the same time, this new size makes the interface more UI consistent and pleasant. Quite more beautiful !

I know right ? :-)

TurboGit:
I think it is ok for 3.0.1 (more consistent size) and since we will be working on providing a way to change font size from the prefs it will be a good base.

TurboGit:
Merged.

PR #3864, opened 2019-12-25
panels enhancements

AlicVB:
This PR is intended to be merged to 'next' branch. It include following change/additions : 
### rework/enhance filmstrip hide/show : 
- save filmstrip/timeline state per view/layouts
- for all view except lighttable, show/hide filmstrip by toggling bottom panel visibility (bottom panel contains only filmstrip there). This avoid strange behavior when clicking on border button if filmstrip has been hidden (I've removed the routine in lib.c/h as this is as 2 lines code, and it's not universal)
- for all lighttable layouts except culling and preview : same as other views, but with timeline
- for lighttable culling and preview, it's a bit more tricky, as we can have both timeline and filmstrip. So consecutive CTRL-F show : filmstrip+timeline => timeline => bottom panel hidden => filmstrip => filmstrip+timeline
- due to this last change, rename lighttable CTRL-F shortcut to show that it act on filmstrip and timeline.
- fix some part of of code where timeline or filmstrip has been forgotten.

### rework filmstrip/timeline resizing : 
- refactor to use same routine as for left/right panels resizing
- this allow to set heights per view/layouts
- ensure that timeline and filmstrip keep same sizes when used together

### partially revert commit 048971a06f506639d2a0b2f4d7f11eb78dfbe0f4
- when restoring panels in "all collapsed" state, we want to keep old panel state values to recover them after exiting this state (by another TAB)
- with the commit, if you have all panels shown, you press TAB, exiting dt and relaunch, if you click another time on TAB, you don't recover your panels state...

### DIFF
--- a/data/darktableconfig.xml.in
+++ b/data/darktableconfig.xml.in
@@ -29,6 +29,20 @@
     <shortdescription>maximum width of the side panels in pixels</shortdescription>
     <longdescription>(needs a restart)</longdescription>
   </dtconfig>
+  <dtconfig>
+    <name>min_panel_height</name>
+    <type>int</type>
+    <default>64</default>
+    <shortdescription>minimum width of the side panels in pixels</shortdescription>
+    <longdescription>(needs a restart)</longdescription>
+  </dtconfig>
+  <dtconfig>
+    <name>max_panel_height</name>
+    <type>int</type>
+    <default>400</default>
+    <shortdescription>maximum width of the side panels in pixels</shortdescription>
+    <longdescription>(needs a restart)</longdescription>
+  </dtconfig>
   <dtconfig prefs="gui">
     <name>slideshow_delay</name>
     <type>int</type>
@@ -577,26 +591,12 @@
     <shortdescription>enable filmstrip</shortdescription>
     <longdescription>enable the filmstrip in darkroom, tethering and map modes.</longdescription>
   </dtconfig>
-  <dtconfig>
-    <name>plugins/lighttable/filmstrip/height</name>
-    <type>int</type>
-    <default>120</default>
-    <shortdescription>height of filmstrip</shortdescription>
-    <longdescription>height of the filmstrip in pixels</longdescription>
-  </dtconfig>
   <dtconfig prefs="gui">
     <name>plugins/lighttable/timeline/visible</name>
     <type>bool</type>
     <default>true</default>
     <shortdescription>enable timeline</shortdescription>
     <longdescription>enable the timeline in lighttable mode.</longdescription>
-  </dtconfig>
-  <dtconfig>
-    <name>plugins/lighttable/timeline/height</name>
-    <type>int</type>
-    <default>72</default>
-    <shortdescription>height of timeline</shortdescription>
-    <longdescription>height of the timeline in pixels</longdescription>
   </dtconfig>
   <dtconfig>
     <name>plugins/lighttable/timeline/last_zoom</name>
--- a/src/gui/gtk.c
+++ b/src/gui/gtk.c
@@ -1643,22 +1643,32 @@ static void _ui_init_panel_size(GtkWidget *widget)
 {
   gchar *key = NULL;
 
+  int s = 128;
   if(strcmp(gtk_widget_get_name(widget), "right") == 0)
   {
     key = _panels_get_panel_path(DT_UI_PANEL_RIGHT, "_size");
+    s = 350; // default panel size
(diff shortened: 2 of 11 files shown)


TurboGit:
I find the timeline+filmstrip not really good. First it looks strange and it makes the rotation between each mode too long as there is many different state. I suppose a simple:

nothing displayed -> timeline -> filmstrip -> [ nothing -> ... ]

Sufficient. No strong opinion on this, would be nice to have more feedback.

TurboGit:
Apart from that all this looks ok to me.

Nilvus:
> I find the timeline+filmstrip not really good. First it looks strange and it makes the rotation between each mode too long as there is many different state. I suppose a simple:
> 
> nothing displayed -> timeline -> filmstrip -> [ nothing -> ... ]
> 
> Sufficient. No strong opinion on this, would be nice to have more feedback.

I agree. Too much press. It's the same with the top panel and header. Sometimes, I only want to see top panel then hide it (and still having same state for hea

To be more precise, der). For example, I always hide header and top panel on darkroom, to let more space to image. But sometimes I need to use top panel, so with 3.0, I only press one time Maj+Ctrl+Shift to show this panel then I'm finish with this panel, I just press one time more same shortcut.
Now, with your last changes, I need to press Maj+Ctrl+Shift to show panel (right) then if I want to hide it, I need to press this shortcut 3 times more (to activate header then hide top panel then hide header). Quite a regression.

So I think, we still have (like actual 3.0) to keep shortcuts top and bottom panels as they are on 3.0.

AlicVB:
> nothing displayed -> timeline -> filmstrip -> [ nothing -> ... ]

So you can't have filmstrip and timeline at the same time ? personally, I never use timeline outside filemanager, and I agree that both libs at the same time are quite ugly...

> Too much press. It's the same with the top panel and header.

What I have in mind when I've done that changes : 
- add some consistency for the border click : the arrow is on top of every panels, but clicking on it only impact toolbar, not header, although it's the closer widget.
- avoid accidental ctrl-h which hide the header bar "forever" (see the nb of beginners issues about that)
- add some consistency in the ui : top bars works the same as bottoms ones.

Now I perfectly understand your use case. What I can propose : 
1. adding a new accel for top toolbar only (empty at beginning) so you can map it to your actual 'top panel' one and find back the old behavior. (we can do the same for bottom panel) => better for beginners, but advanced user will need to tweak accels to come back to the old behavior
2. reverting the accel to the old 'toolbar only' behavior and add a new empty one for toggling both panels => as the arrow click will let beginners get back all panels, we shouldn't have too much 'header disappearing' issues...

TurboGit:
> So you can't have filmstrip and timeline at the same time ? personally, I never use timeline outside filemanager, and I agree that both libs at the same time are quite ugly...

Probably a better idea indeed. I have never used the timeline outside of filemanager either so just remove it from the culling view.

I'm not sure about your both proposals. To me the only issue with this PR is in culling view where we can have the timeline. I would just special case this to only have filmstrip in this case.

AlicVB:
@TurboGit : so I've 'removed' timeline from culling and preview :)
At the same time, I've fix a bug when using both TAB and borders arrows click (in certain cases, it was not possible to get panels back)

TurboGit:
Perfect for me now! Thanks a lot.

Nilvus:
> > Too much press. It's the same with the top panel and header.
> 
> What I have in mind when I've done that changes :
> 
>     * add some consistency for the border click : the arrow is on top of every panels, but clicking on it only impact toolbar, not header, although it's the closer widget.
> 
>     * avoid accidental ctrl-h which hide the header bar "forever" (see the nb of beginners issues about that)
> 
>     * add some consistency in the ui : top bars works the same as bottoms ones.
> 
> 
> Now I perfectly understand your use case. What I can propose :
> 
>     1. adding a new accel for top toolbar only (empty at beginning) so you can map it to your actual 'top panel' one and find back the old behavior. (we can do the same for bottom panel) => better for beginners, but advanced user will need to tweak accels to come back to the old behavior
> 
>     2. reverting the accel to the old 'toolbar only' behavior and add a new empty one for toggling both panels => as the arrow click will let beginners get back all panels, we shouldn't have too much 'header disappearing' issues...

Quite good arguments to change the way you do so. With your last proposals and explanation, point 1 seems to best choice. Advanced users like me could configure the way they want more easily than beginners. If I see quite good, these choices are not implemented by now. Hope you could add this soon to have back my use case ;-) ! Thanks for all these (again) really great improvements.

PR #3868, opened 2019-12-25
Fix sentinel warnings

Ikke:
This fixes build warning: "error: missing sentinel in function call"
where it expects the sentinel to be a char pointer, not just '0'.


Ikke:
This caused build failures on Alpine Linux.

TurboGit:
Thanks !

Issue #3731, opened 2019-12-12
Lens correction deforms gradient


pphotography:
**Describe the bug**
When I activate lens correction the gradients I use in masks get deformed, too.

**To Reproduce**
- Open a photo 
- Add exposure module 
- Add a gradient via drawn mask
- Gradient is a straight line, see:
![mask01](https://user-images.githubusercontent.com/49001045/70700722-74b24380-1ccb-11ea-9d3c-89793f85e6d2.png)
- Open lens correction module
- Use profile for Panasonic, DMC-FZ1000
- Distortion gets corrected, see:
![mask02](https://user-images.githubusercontent.com/49001045/70700877-b7741b80-1ccb-11ea-9e42-97e129a5a9c6.png)
- Open exposure module again
- Activate the display of mask elements
- Gradient has got a curved line, see:
![mask03](https://user-images.githubusercontent.com/49001045/70701026-f5713f80-1ccb-11ea-91ff-b046b49a7ecc.png)

**Expected behavior**
Gradient should stay straight. I don't think this is intended because all gradients I add afterwards are deformed, too.

**Platform (please complete the following information):**
- darktable 3.0.0rc2~git83.5981e2aec
- Ubuntu 18.04

**Additional context**
Happens with and without OpenCL

agusmba:
This could be related (and expected) to iop order.
Did you try gradients on tools that come after lens correction in the pipeline?

pphotography:
I moved lens correction behind exposure without any change.

junkyardsparkle:
In general, you *want* drawn masks to work this way: they're usually closely related to image features and need to maintain that relationship no matter what the lens correction module is doing. It does get kind of weird in the case of gradient masks, though... I've sometimes thought that it would be better if they were their own category of masks, rather than a type of drawn mask. It would be nice to have radial gradients...

TurboGit:
That's a known issue. Yes I do feel it to be an issue but there is no easy fix. One would want to be able to set a forms to not be distorted in any ways (non default behavior). But this is not trivial.

upegelow:
> One would want to be able to set a forms to not be distorted in any ways (non default behavior). But this is not trivial.

I would not open that can of worms. You will end up with a situation where some modules do distort, others not, some shapes yes, few excluded and so on. This would be a massive system of complexity further complicated by the option to move modules around in pixelpipe.

[...]

PR #3879, opened 2019-12-27
Curved gradient

upegelow:
This adds an the option of  a curved gradient shape in blending. Mathematically the line characterizing the gradient takes the form of a parabola:

`y = a * x**2`

The default is a = 0 which falls back to the original linear shape.

This addresses #3731. Typical use case is masking the horizon in an image taken with strong lens distortion. The user will usually activate the lens correction module as one of the first steps. As the coordinate system of all mask shapes is the one of the original input image a linear gradient appears curved in the output image, making it difficult to mask the horizon. Now the user can change the curvature of the gradient for a better match.

Curvature is controlled by mouse scroll without a modifier. Previously this action did change the gradient compression which is now done by shift+scroll. IMHO this makes the behavior of the gradient shape more consistent with other shapes: scrolling without modifier changes the form of the shape, shift+scroll changes compression/feathering.

The width of the transition area (feathering/compression) is not equidistant relative to the parabola. I had considered that but the underlying calculations would have been extremely expensive computationally.

Masks version number is pushed from 4 to 5 by this commit. So the usual warnings apply.

### DIFF
--- a/src/develop/masks.h
+++ b/src/develop/masks.h
@@ -25,7 +25,7 @@
 
 #include <assert.h>
 
-#define DEVELOP_MASKS_VERSION (4)
+#define DEVELOP_MASKS_VERSION (5)
 
 /**forms types */
 typedef enum dt_masks_type_t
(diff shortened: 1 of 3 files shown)


TurboGit:
@upegelow that's a very nice addition for a long standing missing feature. Thanks a lot!

Working fine on my side.

wpferguson:
@upegelow very nice!

Issue #3860, opened 2019-12-25
3.0.0 (color balance and graduated density) not working with DNG (Leica)- black output MacOS 10.14.6 (18G2022)


adulau:
**Describe the bug**``
3.0.0 (color balance and graduated density) not working - black output MacOS 10.14.6 (18G2022)

`**To Reproduce**

1. Go to 'color balance or graduated density' module
2. Enable the module
3. Output of the image is always black (whatever the parameters)

**Expected behavior**

In version 2.6 those modules were working.

**Screenshots**

**Platform (please complete the following information):**
 - MacOS 10.14.6 (18G2022)
 - 10.14.6 (18G2022)
 - OpenCL is not available

**Additional context**

 - This can be only reproduced with RAW images (DNG from Leica) - it works as expected on JPEG.

pphotography:
Cannot reproduce that with dt 3.1.0~git10.af8b11a17 on Ubuntu 18.04 with/without OpenCL.
I took Leica RAW files from
https://rawsamples.ch/index.php/en/leica
https://www.kenrockwell.com/leica/m9/sample-photos-3.htm

Can you tell us which Leica camera u use or provide example files?

aurelienpierre:
Seems related to RGB/Lab on-the-fly conversions, might be the same as #3865.

adulau:
Indeed, if I disable the "filmic RGB module", the other modules work as intended. It sounds like to be a related issue.

Nevertheless, the sample DNG file (from a Leica M10) is available there as requested by @pphotography :

https://www.foo.be/sample-leica/L1002530.DNG

The file is in CC-0 (public domain) feel free to share it and/or use it for test cases.

MStraeten:
seems to be fixed with https://github.com/darktable-org/darktable/pull/3890

adulau:
Thanks a lot. Is there a daily/nightly build available for testing?

MStraeten:
try https://discuss.pixls.us/t/current-osx-build/13213/45

adulau:
Thanks a lot but I run Mojave. I'll compile it and we will make some tests.

johnny-bit:
Duplicate of #3327

PR #3886, opened 2019-12-27
try to fix #3865 and #3860

### LINKED ISSUE #3865: filmic rgb + base curve = black
I installed the new 3.0 version.

I have the base curve set as a default module.

If I activate the new filmic rgb module, I get a dark image, regardless of the filmic rgb parameters values.

After that, if I deactivate the base curve module, I go back to the non-black image

aurelienpierre:
remove OpenCL pointers from the `constant` namespace and store them to the `global` with `const` tags (revert 5035177adb).


aurelienpierre:
This is not to be merged until proved really useful because it can penalize performance.

aurelienpierre:
The bug had other sources.

PR #3888, opened 2019-12-27
opencl: make preview2 pipe part of normal scheduling logic

upegelow:
The basics have been there already but the device priority for the preview2
pipe was never really defined. As a result the preview2 pipe is processed
on the CPU unless the user manually changes opencl_device_priority. We now
set the same values as for the preview pipe. Note that this only affects
systems run in 'multiple GPUs' or 'very fast GPU' profile or new installations
(existing opencl_device_priority is not overwritten).


TurboGit:
Thanks !

PR #3890, opened 2019-12-27
OpenCL filmic rgb: try to fix #3865

### LINKED ISSUE #3865: filmic rgb + base curve = black

aurelienpierre:


MStraeten:
is fmaxf in line 212 a typo? i changed it in my environment to fmax and then the kernel compiled. Also ih didn't get the black image ;)

TurboGit:
Not a typo, man pages says it exists and the def is:

`float fmaxf(float x, float y);`

MStraeten:
ok, i grepped around in the kernel directory: just filmic.cl kernel is using fmaxf - all other kernels use fmax an they compile fine on osx, filmic.c doesn't

MStraeten:
several test later: in darkroom everything is now fine, but in lighttable a set of 30 shots with similar conditions, shows a mixture of well processed and black images if using power norm. After switching to darkroom for a selected black image this is displayed well processed there and afterwards also in lighttable. 
But switching to culling or changing the number of images in a line brings back different black images. Cleaning the mipmaps cache directory doesn't change the behaviour.

aurelienpierre:
So `fmax` it is for OpenCL, and 

> several test later: in darkroom everything is now fine, but in lighttable a set of 30 shots with similar conditions, shows a mixture of well processed and black images if using power norm.

means I need to do the same for the plain C path.

MStraeten:
yep, it's fine now, thanks

aurelienpierre:
YEAAAAH !!! :smiley:

TurboGit:
Thanks !

PR #3898, opened 2019-12-28
copy metadata - move metadata operations to a different tab

phweyland:
I've tried to save space but that's maybe not the best. let me know.

### DIFF
--- a/src/libs/image.c
+++ b/src/libs/image.c
@@ -332,7 +332,7 @@ static void tags_flag_callback(GtkWidget *widget, dt_lib_module_t *self)
 
 static void pastemode_combobox_changed(GtkWidget *widget, gpointer user_data)
 {
-  int mode = dt_bauhaus_combobox_get(widget);
+  const int mode = dt_bauhaus_combobox_get(widget);
   dt_conf_set_int("plugins/lighttable/copy_metadata/pastemode", mode);
 }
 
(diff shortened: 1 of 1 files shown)


TurboGit:
The mode replace is not working for me it also do a merge action.

Also I would put the [paste] button on the same line as the mode (the two modes are short text and all should fit I suppose.

So, in ASCII art it will be:

`
   [paste]     mode  merge v`

src/libs/image.c
```diff
@@ -334,20 +330,38 @@ static void tags_flag_callback(GtkWidget *widget, dt_lib_module_t *self)
   dt_conf_set_bool("plugins/lighttable/copy_metadata/tags", flag);
 }
 
+static void pastemode_combobox_changed(GtkWidget *widget, gpointer user_data)
+{
+  int mode = dt_bauhaus_combobox_get(widget);
```
TurboGit: cont

TurboGit:
Sorry testing error :) The merge vs replace works just fine !

TurboGit:
So I would only update the layout if you agree.

phweyland:
> Also I would put the [paste] button on the same line as the mode (the two modes are short text and all should fit I suppose.

I've tried this too but it isn't smarter ... anyway I do the change. That's simple to revert it if necessary.

phweyland:
Here is how it looks like now:

image

phweyland:
Another option would be to have 3 columns for the check boxes, so reduce them to 2 lines. Then we could keep the mode on complete line (as in history stack).

phweyland:
image

or

image

src/libs/image.c
phweyland: done

phweyland:
This one looks the best for me (if room for text is enough):

image

or

image


what do you think ?

TurboGit:
Let's go with first one, the one you like best is also looking good to me.

phweyland:
Done.

TurboGit:
Thanks, merged. I have re-aranged a bit the widgets as in French and with smaller panels it was not fitting on my side. Let me know what you think.

phweyland:
On my screen it's a bit messy

image

I would set the offset of colors and geotags to 3 (instead of 2) to let the check boxes use the full width.

the mode value is also misaligned. But I admit this combo box is not pretty (with so short words). The original buttons merge / replace were more elegant. If you agree I can change back this.

TurboGit:
Agreed! And done.

PR #3912, opened 2019-12-28
Added RGB model-based color spaces

VasylTretiakov:
To complement the advent of new rgb modules like filmic rgb, etc.


TurboGit:
Thanks !

PR #3922, opened 2019-12-29
Enhanced module summary

VasylTretiakov:
1. rephrased module types
2. moved and rephrased paragraph on multiple module instances and control over their order

### DIFF
--- a/doc/usermanual/overview/overview.xml
+++ b/doc/usermanual/overview/overview.xml
@@ -98,7 +98,7 @@
 
     <listitem><para>
       darktable comes with more than 60 image processing modules which cover correction of 
-      digital image imperfectoins and common defects, basic adjustments, tonal value
+      digital image imperfections and common defects, basic adjustments, tonal value
       changes, color manipulation, as well as enhancements and artistic effects.
     </para></listitem>
 
--- a/doc/usermanual/overview/overview.xml
+++ b/doc/usermanual/overview/overview.xml
@@ -103,7 +103,7 @@
     </para></listitem>
 
     <listitem><para>
-      You can instantiate most modules multiple times and control order in which they get applied.
+      You can instantiate most modules multiple times and control the order in which they get applied.
     </para></listitem>
 
     <listitem><para>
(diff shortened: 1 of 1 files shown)


doc/usermanual/overview/overview.xml
```diff
@@ -97,9 +97,13 @@
     </para></listitem>
 
     <listitem><para>
-      darktable comes with more than 60 image operation modules which cover everything from
-      basic operations, tonal value changes, color manipulation, correction of common image
-      defects to artistic effects.
+      darktable comes with more than 60 image processing modules which cover correction of 
+      digital image imperfectoins and common defects, basic adjustments, tonal value
```
brucewilliamsphotography: imperfections

VasylTretiakov: indeed, fixed

doc/usermanual/overview/overview.xml
```diff
@@ -97,9 +97,13 @@
...
-      darktable comes with more than 60 image operation modules which cover everything from
-      basic operations, tonal value changes, color manipulation, correction of common image
-      defects to artistic effects.
+      darktable comes with more than 60 image processing modules which cover correction of 
+      digital image imperfections and common defects, basic adjustments, tonal value
+      changes, color manipulation, as well as enhancements and artistic effects.
+    </para></listitem>
+
+    <listitem><para>
+      You can instantiate most modules multiple times and control order in which they get applied.
```
Balise42: "control the order"?

Also, one of the takeaways I got from Aurélien's videos is "you shall be super careful if you change the order of the modules and actually know what you're doing". Would a warning/note be a good thing here, or is it too early to get into that kind of detail?

VasylTretiakov: added "the" article
true, but here in the high-level summary I am not sure it is a good idea to insert warnings and notes. besides, here I am mainly referring to moving instances of the same module which should be relatively safe (unless one uses the raster mask of the other, then it seems it could crash darktable). moving different modules across the pixelpipe may indeed cause undesired or sub-optimal effects on the final picture (but hopefully does not crash the darktable)

TurboGit:
Thanks !

PR #3923, opened 2019-12-29
manual: alternative raster masks section

VasylTretiakov:
The suggested version seems more clear succinct to me with additional insights provided by the blog post.
For example, if we aim to hide technical details, there is little motivation to introduce alpha map which then gets renamed into raster mask. Instead it is simpler to call it a raster mask from the start.
I also have a doubt whether "raster mask" UI element should be called a drop-down or a combo box (the latter a general concept introduced in section 3.2.2.2. Comboboxes).
Anyway, I let you judge if this brings any value or not.

### DIFF
--- a/doc/usermanual/darkroom/concepts/raster_masks.xml
+++ b/doc/usermanual/darkroom/concepts/raster_masks.xml
@@ -13,10 +13,8 @@
   </indexterm>
 
   <para>
-    When a <quote>drawn mask</quote> or a <quote>parametric mask</quote> is active
-    the final mask can be reused in other modules. This works because all the shapes from the drawn mask 
-    and all the blend functions from the parametric mask of a module assemble a final mask,
-    which internally is stored as a raster image and made accessible to other modules.
+    A raster mask produced by <quote>drawn mask</quote> and/or <quote>parametric mask</quote> in the
+    context of a specific module can be reused by modules that follow it in the pixelpipe.
   </para>
 
   <sect4 status="draft" id="raster_mask_overview">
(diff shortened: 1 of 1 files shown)


doc/usermanual/darkroom/concepts/raster_masks.xml
```diff
@@ -13,50 +13,54 @@
   </indexterm>
 
   <para>
```
brucewilliamsphotography: Personally, I think the original wording here is the better option. Sure, it's a little more verbose, but I think the subject matter warrants the deeper explanation. Just my 2 cents.

VasylTretiakov: Thanks for your feedback. My main doubts about the original lines 16-19 are as follows:
1. I am not sure what it means for a mask to be active.
My understanding is that a a specific mask option (uniform, drawn, parametric, or drawn and parametric) needs to be enabled in the context of a given module.
Does enabling a mask option make a mask active? I'm not sure.
- Enabling a drawn mask without drawing a single shape already makes it available as a raster mask down the pixelpipe. Can an empty drawn mask be considered active?
- The same question applies to a parametric mask in default state as it is also made available down the pixelpipe.
-  Disabling a module with a mask still leaves the raster mask available down the pixelpipe (at least in the drop-down). Can a mask in disabled module be considered active?
2. I do not understand a reference to blend functions in the context of a parametric mask.
Blending operations/modes (normal, lighten, darken, etc.) are available for any mask type. Parametric mask controls help in defining opacity of what the manual calls a blend mask (again possible clash with blend modes) via selection curves. Maybe if we start referring to selection curves of parametric masks, we can prevent this clash of the terms at least in the discussed paragraph. Or maybe my perception is wrong and there is no conflict here.
3. I am not sure if the notion of a final mask is used in other parts of the manual.
If not, it seems to be needed only for a specific use case which kind of pollutes the domain language.
4. I have a doubt about the phrasing "..shapes.. and blending curves.. assemble the final mask".
I would say they contribute to final mask, as assembly implies a union, while a combined drawn and parametric mask allow easily excluding areas which distinct masks would select.
5. I am concerned that we are mentioning only drawn or parametric masks.
Uniform masks on modules also appear in raster mask drop-down. Of course, their use case is rather primitive (just like the ones of default drawn or parametric mask). BTW, this also introduces an issue to the phrasing I am suggesting, so it would need to be fixed.
Drawn and parametric masks are implied by mentioning drawn masks and parametric ones, but I wonder if they should be mentioned explicitly.

Any constructive comments are welcome.

PR #3926, opened 2019-12-29
manual: enhance visibility of tagging panel views

VasylTretiakov:
tagging panel parts were extracted to an item list

### DIFF
--- a/doc/usermanual/lighttable/panels/tagging.xml
+++ b/doc/usermanual/lighttable/panels/tagging.xml
@@ -50,19 +50,20 @@
 
     <para>
       The panel is divided into two areas:
-        <itemizedlist>
+      <itemizedlist>
 
-          <listitem><para>
-            The upper one called attached tags area focuses on the tag(s) currently attached to the image
-            under the cursor or the selected image(s) if the mouse is outside the lighttable. It contains
-            the attached tags list view and the row of five buttons under it.
-          </para></listitem>
+        <listitem><para>
+          The upper one called attached tags area focuses on the tag(s) currently attached to the image
+          under the cursor or the selected image(s) if the mouse is outside the lighttable. It contains
+          the attached tags list view and the row of five buttons under it.
+        </para></listitem>
                   
-          <listitem><para>
-            The lower one called tag dictonary area focuses on all tags  available in the database.
-            It contains the text box, the tag dictionary list view and the row of five buttons under it.
-            </para></listitem>
-          </itemizedlist>
+        <listitem><para>
+          The lower one called tag dictonary area focuses on all tags  available in the database.
+          It contains the text box, the tag dictionary list view and the row of five buttons under it.
+          </para></listitem>
+        </itemizedlist>
+      </para>
     <sect4 id="tagging_tag_properties">
 
       <title>tag properties</title>
--- a/doc/usermanual/lighttable/panels/tagging.xml
+++ b/doc/usermanual/lighttable/panels/tagging.xml
@@ -37,7 +37,7 @@
               </para>
               <para>
                 Unlike child tags, parent tags do not add information to the image, but rather represent categories.
-                Consider the <quote>color|white</quote> tag where <quote>color</quote> surves as a category 
+                For example, in the <quote>color|white</quote> tag <quote>color</quote> is a category 
                 for the <quote>white</quote> tag.
               </para>
             </entry>
--- a/doc/usermanual/lighttable/panels/tagging.xml
+++ b/doc/usermanual/lighttable/panels/tagging.xml
@@ -29,39 +29,15 @@
             <entry>
               <para>
                 This panel is for managing tags on images. Tags are stored in both sidecar files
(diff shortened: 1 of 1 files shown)


doc/usermanual/lighttable/panels/tagging.xml
```diff
@@ -29,29 +29,15 @@
             <entry>
               <para>
                 This panel is for managing tags on images. Tags are stored in both sidecar files
-                (.xmp) and within the darktable database for a faster access. The panel is divided
-                into two parts: the upper part contains the tag(s) currently attached to the image
-                under the cursor or the selected image(s) if the mouse is outside the lighttable.
-                The lower part contains a list of all available tags in the database, the dictionary, which can be
-                filtered in the text box above.
+                (.xmp) and within the darktable database for a faster access.
```
brucewilliamsphotography: Actually, the "a" is redundant. This should read...
(.xmp) and within the darktable database for faster access.

doc/usermanual/lighttable/panels/tagging.xml
```diff
@@ -29,29 +29,15 @@
...
-                    A tree view to navigate within your hierarchical tags
-                  </para></listitem>
-
-                  <listitem><para>
-                    A list of suggestions based on already attached tags on selected images.
-                  </para></listitem>
-
-                </itemizedlist>
+                The goal of tags is to add information to the images within a given keyword set.
+                But when the number of tags becomes large it is interesting to organize them in a hierachical manner, as a tree.
```
brucewilliamsphotography: But when the number of tags becomes large, it is more helpful to organise them in a hierarchical manner, as a tree.

doc/usermanual/lighttable/panels/tagging.xml
```diff
@@ -29,29 +29,15 @@
...
+                The parent tags may not add information to the image, for example the tag <quote>color</quote>
+                in <quote>color|white</quote>. In this case <quote>color</quote> will automatically
```
brucewilliamsphotography: in <quote>color|white</quote>. In this case <quote>color</quote> will automatically be seen as a <quote>category</quote> if it is not in the dictionary itself. If it already exists as a tag, it can be set explicitly as a <quote>category</quote> (see create or edit command).

doc/usermanual/lighttable/panels/tagging.xml
Balise42: Only a matter of taste: I'd tend to write "You can organize the tags hierarchically as a tree, which is interesting when the number of tags becomes large".
In any case: "hierarchical" (missing an r).

doc/usermanual/lighttable/panels/tagging.xml
```diff
@@ -29,29 +29,15 @@
...
+                seen as <quote>category</quote> if it is not in the dictionary itself. If exist as a tag, it can be set
+                explicitly as a <quote>category</quote> (see create or edit command).
```
Balise42: can we add a crosslink here?

doc/usermanual/lighttable/panels/tagging.xml
VasylTretiakov: Ironically, I worked only on the markup using already present text. But, as I agree that phrasing is sup-optimal, I rephrased towards simplification and (hopefully) clarity.
Frankly speaking, I still do not get the point of using tags as categories (why not two distinct entities?), but that requires an in-depth technical analysis, hence needs to be tackled separately.
I also do not like the fact that lighttable is a view, and tag dictionary is a view, and tree/list view is also.. a view :)

doc/usermanual/lighttable/panels/tagging.xml
VasylTretiakov: I removed the reference in parentheses as usage instructions follow in a dedicated section and I see no point in speaking of implicit and explicit categories here.

doc/usermanual/lighttable/panels/tagging.xml
```diff
@@ -28,16 +28,17 @@
...
-                The parent tags may not add information to the image, for example the tag <quote>color</quote>
-                in <quote>color|white</quote>. In this case <quote>color</quote> will automatically
-                seen as <quote>category</quote> if it is not in the dictionary itself. If exist as a tag, it can be set
-                explicitly as a <quote>category</quote> (see create or edit command).
+                Tags provide the means of adding information to the images within a given keyword dictionary.
+                You can manage tags as a hierarchical tree, which is useful when their number becomes large.
+              </para>
+              <para>
+                Unlike child tags, parent tags do not add information to the image, but rather represent categories.
+                Consider the <quote>color|white</quote> tag where <quote>color</quote> surves as a category
```
Balise42: Suggestion: For instance, in the <quote>color|white</quote> tag, <quote>color</quote> is a category for the <quote>white</quote> tag.
In any case: I think you mean "serve" and not "surve"?

VasylTretiakov: merged

TurboGit:
Thanks!

Issue #3918, opened 2019-12-29
DT Crash : Drawn gradient mask on image, then click uniformly circle icon next to drawn mask.


chhil:
I have tried this on various modules so its not module related.
![image](https://user-images.githubusercontent.com/1119439/71551534-e995b500-2a0f-11ea-89a1-e53a79ff1159.png)

Select drawn mask and then I dropped a gradient mask on the image.
Then click the uniformly (circle icon) next to it, DT window just disappears.

Version : darktable-3.1.0+221~ga8695754f-win64 (local release build)

chhil:
It crashes even when you do a drawn masks and select path tool. As soon as you click the image to put a point for the path it crashes.

chhil:
I have reverted to 3.0 as the master build is currently not usable for me. 3.0 does not have the issue.

upegelow:
This might be the same as #3907

chhil:
@upegelow 
Built dt this morning that has the gradient fix.

It has got worse, 
On a clean dt with clean db, imported images with xmp.
Dropped gradient mask.
Tried to rotate it and it did not rotate, tried a couple of times and DT crashed.

Reopening DT did not work after that. Had to move the config dir out to treat is as clean.
Now by just dropping a gradient DT crashes.

upegelow:
Am 30.12.19 um 04:16 schrieb Murtuza Chhil:
> @upegelow <https://github.com/upegelow>
> Built dt this morning that has the gradient fix.
> 
> It has got worse,
Please supply raw file plus xmp.

chhil:
@upegelow , requested files are attached.

Darktable log file.
[darktable-log.txt](https://github.com/darktable-org/darktable/files/4009625/darktable-log.txt)

Zip file contains raw+xmp. Remove the txt extension.
[P1044714.zip.txt](https://github.com/darktable-org/darktable/files/4009629/P1044714.zip.txt)

No module processing was done by me , just an import folder for images (no xmps were present), picked the image, went to exposure module, selected drawn mask, dropped gradient on image and it crashed.

upegelow:
Thanks. Please check PR #3946.

chhil:
One thing I forgot to mention.
When I drop the gradient , its supposed to be horizontal, but on drop it becomes vertical and then app dies.

chhil:
@upegelow 
I currently am on the latest master, build 15 min ago.
Once your #3946 makes it to master I will build it and try again.

[...]

Issue #3907, opened 2019-12-28
"showing" gradient mask broken in dt 3.1.0+222


jenshannoschwalm:
While

- defining a new gradient mask or 
- using a developed image and use the "show and edit mask elements"

you will observe these problems (the mask visualizing and the desired effect seem to be ok)

1. the gradient is displayed completely wrong like in this
![Bildschirmfoto von 2019-12-28 21-40-36](https://user-images.githubusercontent.com/50982232/71549326-fd173080-29bb-11ea-8f97-c907ab186331.png)

2. At the same time dt is **very** unresponsive and cpu consumption is ~100% for one thread, this takes up to a few seconds (cairo seems to be drawing ...)
3. you will not be able to see the "mask list" as here, .
![Bildschirmfoto von 2019-12-28 21-43-02](https://user-images.githubusercontent.com/50982232/71549384-afe78e80-29bc-11ea-8bb7-145128743a0a.png)

Conditions to trigger
1. Use a fresh edit on any image, create a gradient and position it. Or "show" the gradient mask in an already developed image 
2. lens correction is a pre-requisite, problem is **only** detectable if lens correction is on
3. Cl or cpu code doesn't matter

 - Darktable Version: current git master (compiled after clearing build and removing all cl kernels)
 - OS: Fedora
 - nvidia 440.44

TurboGit:
Really strange... There must be something more as I cannot reproduce !

upegelow:
Am 29.12.19 um 08:50 schrieb Pascal Obry:
> Really strange... There must be something more as I cannot reproduce !

I think I've spotted the problem. A bug I have introduced with curved 
gradients. Will fix asap.

jenshannoschwalm:
@upegelow thanks, i just wanted to ping you.

Not being able to reproduce might be related to
1. You **must** have the lens correction activated **before** switching to the module where you view the mask.
2. The problem is significant only if the lens correction is pretty strong (try LeicaQ test images for example)

The wrong lines & "cairo doing wrong stuff" are visible in masks/gradient.c ~L.615. I tested with
```
    {
      _gradient_point_transform(xref, yref, gpt->points[i * 2] + dx, gpt->points[i * 2 + 1] + dy, sinv, cosv,
                                &x, &y);
      fprintf(stderr," (%.1f|%.1f)",x,y);
      cairo_line_to(cr, x, y);
    }
```
and many x/y are nan (cairo takes long in these cases...)

upegelow:
[...]

PR #3929, opened 2019-12-29
drawn masks: fix for curved gradient

upegelow:
hopefully fixing #3907 and #3918

### DIFF
--- a/src/develop/masks/gradient.c
+++ b/src/develop/masks/gradient.c
@@ -760,7 +760,7 @@ static int dt_gradient_get_points(dt_develop_t *dev, float x, float y, float rot
   const float cosv = cos(v);
   const float sinv = sin(v);
 
-  *points_count = 2 * sqrtf(wd * wd + ht * ht) + 3;
+  *points_count = sqrtf(wd * wd + ht * ht) / 2 + 3;
   *points = dt_alloc_align(64, 2 * (*points_count) * sizeof(float));
   if(*points == NULL) return 0;
   memset(*points, 0, 2 * (*points_count) * sizeof(float));
@@ -783,11 +783,12 @@ static int dt_gradient_get_points(dt_develop_t *dev, float x, float y, float rot
   (*points)[5] = y2;
 
   // we set the line point
-  const float xstart = MAX(-sqrtf(1.0f / fabsf(curvature)), -1.0f);
+  const float xstart = fabs(curvature) > 1.0f ? -sqrtf(1.0f / fabsf(curvature)) : -1.0f;
   const float xdelta = -2.0f * xstart / (*points_count - 3);
+
   for(int i = 3; i < *points_count; i++)
   {
-    const float xi = xstart + i * xdelta;
+    const float xi = xstart + (i - 3) * xdelta;
     const float yi = curvature * xi * xi;
     const float xii = (cosv * xi + sinv * yi) * scale;
     const float yii = (sinv * xi - cosv * yi) * scale;
--- a/src/develop/masks/gradient.c
+++ b/src/develop/masks/gradient.c
@@ -84,12 +84,15 @@ static int dt_gradient_events_mouse_scrolled(struct dt_iop_module_t *module, flo
 {
   if(gui->creation)
   {
-    float compression = MIN(1.0f, dt_conf_get_float("plugins/darkroom/masks/gradient/compression"));
-    if(up)
-      compression = fmaxf(compression, 0.001f) * 0.8f;
-    else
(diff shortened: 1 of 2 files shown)


TurboGit:
Will have a look ASAP.

Also I have spotted a discrepancies. In creation mode the mouse scroll change the gradient, in edit mode it changes the curvature.

upegelow:
Am 29.12.19 um 16:49 schrieb Pascal Obry:
> Also I have spotted a discrepancies. In creation mode the mouse scroll 
> change the gradient, in edit mode it changes the curvature.

Certainly this has to be consistent.

TurboGit:
And I think mouse scroll should change the form as for all other shapes. Changing the gradient should be mouse scroll. Ok for you?

upegelow:
Am 29.12.19 um 17:14 schrieb Pascal Obry:
> And I think mouse scroll should change the form as for all other shapes. 
> Changing the gradient should be mouse scroll. Ok for you?

I was considering to allow change of curvature only after the gradient 
is created.

The way how all these scroll actions work is by storing the last value 
in darktablerc. So next time the same form is created we start with the 
previous value (e.g. diameter of a circle/ellipse).

For the gradient I assume that 99% of the cases the user wants a 
straigth line. A curved gradient is a very special case and users might 
be irritated to suddenly find that all new gradients are curved.

TurboGit:
Makes sense. Let's go with that.

upegelow:
Am 29.12.19 um 17:29 schrieb Pascal Obry:
> Makes sense. Let's go with that.
> 
I will add a commit to the existing PR.

jenshannoschwalm:
I just tested this pr; the lagging feels much better (you can pick the anchor and rotate but drawing that takes up to 2 secs)
but there are still wrong lines visible like before.

upegelow:
Am 29.12.19 um 17:57 schrieb Hanno Schwalm:
> I just tested this pr; the lagging feels much better (you can pick the 
> anchor and rotate but drawing that takes up to 2 secs)
> but there are still wrong lines visible like before.
> 

Here everything is very responsive, even when selecting lens correction 
with a strong distorting lens (e.g. Leica lens of a Panasonic FZ1000).

Probably there are some issues if we feed coordinates which are too far 
outside of the image frame into transform/backtransform of lens 
correction. For calculating the gradient guides we do so in order to 
have some meaningful indicators when the user shifts the gradient. 
Admittedly the guides extend quite a lot. I'll try to make this behave 
more reasonable.

BTW: please suggest a camera/lens combination for further testing.

jenshannoschwalm:
Leica Q, has a fixed lens with pretty strong correction. Test image available in pixls - i am using that for testing too - the lensfun file is
here.
[leica-q.xml.txt](https://github.com/darktable-org/darktable/files/4008373/leica-q.xml.txt)

upegelow:
Am 29.12.19 um 18:46 schrieb Hanno Schwalm:
> Leica Q, has a fixed lens with pretty strong correction. Test image 
> available in pixls - i am using that for testing too - the lensfun file is
> here.
> leica-q.xml.txt 
> <https://github.com/darktable-org/darktable/files/4008373/leica-q.xml.txt>
> 

Interesting :) That profile seems to behave very badly if confronted 
with coordinates too far out of the image frame. I will need to dig a 
bit deeper into that issue.

upegelow:
It's not specific for the gradient shape. If you make a circle shape 
large enough and shift it far outside the frame (zoom out to 8%) the 
lens correction module goes all berserk.

TurboGit:
This last version looks good to me. If there is still an issue with the lens correction I suppose we should fix that in another PR. Will merge this in master and 3.0.1. Thanks.

TurboGit:
Actually the curve gradient is for 3.2, so only merged in master.

jenshannoschwalm:
@upegelow latest commit fixes here too :-)

jenshannoschwalm:
> Actually the curve gradient is for 3.2, so only merged in master.

Is that decision debatable? At least for some Leicas it would mean a severe restriction for any 3.0.x version. And there are more issues around that could be related.

TurboGit:
@jenshannoschwalm : Yes, that's an open project, everything is debatable as long as the solution is workable.

upegelow:
Am 29.12.19 um 21:08 schrieb Hanno Schwalm:
> @upegelow <https://github.com/upegelow> latest commit fixes here too :-)
> 
I think we have a general topic with the transform/backtransform logic 
here. What input value are permissible? How can we tell between valid 
and invalid output values? I think this is not defined but should be.

PR #3930, opened 2019-12-29
manual: update lighttable tag overlay desc

VasylTretiakov:
it seems no one updated tag overlay content after tagging module rewrite


doc/usermanual/lighttable/panels/tagging.xml
```diff
@@ -137,16 +137,15 @@
...
       </para>
 
       <para>
         <graphic fileref="lighttable/images/panel_tagging-submenu.png" scalefit="1" width="55%" align="center" />
       </para>
 
       <para>
-        An overlay appears hovering a tag which has some properties set.
-        This avoids to edit the tag to see its properties.
+        Hovering cursor over a tag raises an overlay with tag and container window usage instructions.
```
Balise42: I'm a bit confused by the "tag and container window" in here. I would have expected "tagging window usage instructions" (because the window is called tagging). If so, then I'd suggest:
"Hovering over a tag displays an overlay with the usage instructions for the tagging window".

VasylTretiakov: I rephrased

TurboGit:
Thanks !

Issue #3932, opened 2019-12-29
git master build failing


RobbieAB:
git master build is failing with the following error:

`[ 78%] Building C object src/iop/CMakeFiles/watermark.dir/introspection_watermark.c.o
In file included from /home/robert/Code/darktable/build/src/iop/introspection_watermark.c:48:
/home/robert/Code/darktable/src/iop/watermark.c: In function ‘_watermark_get_svgdoc’:
/home/robert/Code/darktable/src/iop/watermark.c:863:3: error: unknown type name ‘dt_variables_params_t’
  863 |   dt_variables_params_t *params;
      |   ^~~~~~~~~~~~~~~~~~~~~
/home/robert/Code/darktable/src/iop/watermark.c:864:3: error: implicit declaration of function ‘dt_variables_params_init’ [-Werror=implicit-function-declaration]
  864 |   dt_variables_params_init(&params);
      |   ^~~~~~~~~~~~~~~~~~~~~~~~
/home/robert/Code/darktable/src/iop/watermark.c:865:9: error: request for member ‘filename’ in something not a structure or union
  865 |   params->filename = image->filename;
      |         ^~
/home/robert/Code/darktable/src/iop/watermark.c:866:9: error: request for member ‘jobcode’ in something not a structure or union
  866 |   params->jobcode = "watermark";
      |         ^~
/home/robert/Code/darktable/src/iop/watermark.c:867:9: error: request for member ‘sequence’ in something not a structure or union
  867 |   params->sequence = 0;
      |         ^~
/home/robert/Code/darktable/src/iop/watermark.c:868:9: error: request for member ‘imgid’ in something not a structure or union
  868 |   params->imgid = image->id;
      |         ^~
/home/robert/Code/darktable/src/iop/watermark.c:869:3: error: implicit declaration of function ‘dt_variables_set_tags_flags’ [-Werror=implicit-function-declaration]
  869 |   dt_variables_set_tags_flags(params, flags);
      |   ^~~~~~~~~~~~~~~~~~~~~~~~~~~
/home/robert/Code/darktable/src/iop/watermark.c:870:12: error: implicit declaration of function ‘dt_variables_expand’ [-Werror=implicit-function-declaration]
  870 |   svgdoc = dt_variables_expand(params, svgdata, FALSE);
      |            ^~~~~~~~~~~~~~~~~~~
[...]

PR #3933, opened 2019-12-29
Include variables.h in iop/watermark.c

RobbieAB:
Without it, I am getting build failures for introspection_watermark.

fixes: #3932


TurboGit:
Really strange that not everyone has the issue, but this PR seems ok so I'll merge. Yet, if you have any explanation about what was happening it will nice to know. Thanks.

PR #3934, opened 2019-12-29
fix strip_semicolons_from_keymap

rabauke:
Fixes the following bug that has been introduced in commit 89be55654ac16dcd8215992034c9eb81f812364c: 

The function `strip_semicolons_from_keymap` removes not only the first three lines from the input file but also the 1st character of the 4th line. Consequently the file `keyboradrc_default` is corrupted and resetting keyboard accelerators to defaults does not work.

I am not sure if the whole treatment for possible line endings different from pure new-line (introduced in in commit 89be55654ac16dcd8215992034c9eb81f812364c) is meaningful. `gtk_accel_map_save` writes new-lines only, thus `gtk_accel_map_load` may not be able to understand new-line carriage-return line endings anyway.

### DIFF
--- a/src/common/darktable.c
+++ b/src/common/darktable.c
@@ -175,8 +175,6 @@ static void strip_semicolons_from_keymap(const char *path)
   char pathtmp[PATH_MAX] = { 0 };
   FILE *fin = g_fopen(path, "rb");
   FILE *fout;
-  int i;
-  int c = '\0';
 
   if(!fin) return;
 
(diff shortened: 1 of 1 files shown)


TurboGit:
Can you describe what this fixes? Hard to do testing/review without knowing what to look at. Thanks.

rabauke:
> Can you describe what this fixes? Hard to do testing/review without knowing what to look at. Thanks.

I already described the bug above. Maybe I have to go more into the details.

When you start dt for the very first time the file `.config/darktable/keyboardrc_default` is created via `gtk_accel_map_save`. Initially this file contains all key accelerators, but they are all commented out by a `; ` at the beginning of the line. There are also three lines of comments at the top of the file.

The function `strip_semicolons_from_keymap` shall remove these three lines and strip the semicolons from the beginning of the following lines. Due to a bug, however, to many characters are removed. `.config/darktable/keyboardrc_default` starts with 

    gtk_accel_path "<Darktable>/image operations/highpass/sharpness/decrease" "")

expected result is 

    (gtk_accel_path "<Darktable>/image operations/highpass/sharpness/decrease" "")

TurboGit:
Ok, thanks for the details information. Clearer to me now. Thanks.

PR #3935, opened 2019-12-29
Fix build on PPC and ARM (cpuid)

cryptomilk:
This fixes building on ARM and PPC and detects support for `__get_cpuid()` correctly.

### DIFF
--- a/ConfigureChecks.cmake
+++ b/ConfigureChecks.cmake
@@ -1,8 +1,13 @@
 include(CheckCSourceCompiles)
 include(TestBigEndian)
 include(CheckIncludeFile)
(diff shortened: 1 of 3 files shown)


TurboGit:
@cryptomilk : is that only for 3.2 ? or should it go to 3.0.1 ? I suppose the former given it is for PPC and ARM.

cryptomilk:
@TurboGit Depends on if you pushed the cpuid patchset to the 3.0 branch too. I recognized it failing in the OBS master build.

TurboGit:
No, the cpuid patchset is not on 3.0 branch. So only for 3.2 for this one too.

PR #3946, opened 2019-12-30
drawn masks: further restrictions on display of gradient guides

upegelow:


TurboGit:
All good on my side. Thanks.

PR #3947, opened 2019-12-30
cmake: Only look for avif cmake configs and not find modules

cryptomilk:
Should reduce output of cmake when not available.


TurboGit:
Thanks !

Issue #3944, opened 2019-12-30
metadata exportation details / exif


majobu:
With darktable 3.0.0 metadata exportation details are not working as expected. Variables of the manual do not work. 

For example, using $(EXIF_FOCAL_LENGTH) works, but $(EXIF_APERTURE) returns a value of 1.0.

phweyland:
Could you please describe (or screenshot) your settings and attach the resulting data ? Thanks

chhil:
The image information snapshot might be helpful to the dev to see what was expected.

phweyland:
I confirm the issue using these variables for exported images metadata.
Currently these variables (all) return a string (text), which is convenient for file name or watermark text, but sometime not to fulfill metadata. 
I'll try to see if we can include in the PR #3951 some more control on this.

junkyardsparkle:
Lua treats these as strings, too, I don't know if that might complicate things. I currently use a Lua script to fill several of these variables on import for images taken with old manual lenses. The new values work fine within darktable (matching in the lens correction module, for example), but they've never been exported by darktable. It would be great if they could be.

johnny-bit:
This seems to be fixed by #3951 which is present in 3.0.1+, 3.2.1 and 3.4.x so closing :)

If it's not fixed, please reply and I'll reopen

PR #3951, opened 2019-12-30
export metadata - replace fixed tag list by exiv2 list 

phweyland:
Replaced the hard coded metadata list by an exiv2 standard one.

### DIFF
--- a/src/common/exif.cc
+++ b/src/common/exif.cc
@@ -65,6 +65,7 @@ extern "C" {
 #include "common/tags.h"
 #include "common/iop_order.h"
 #include "common/variables.h"
+#include "common/utility.h"
 #include "control/conf.h"
 #include "develop/imageop.h"
 #include "develop/blend.h"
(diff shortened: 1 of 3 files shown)


phweyland:
Retrieving more metadata from exiv2 brings some data type which are not covered today like Rational or SRational types. Some more work must be done at least to prevent to fulfill a tag with not coherent values.

phweyland:
adding metadata exiv2 type information

phweyland:
Some variables like $(EXIF_APERTURE) are not suitable to set up the metadata tag Exif.Photo.ApertureValue which a RATIONAL (two longs) as commented in that thread https://discuss.pixls.us/t/darktable-3-0-exif-on-export/15486.

I've worked on the exiv2 metadata types (which are now visible in the metadata tag dictionary) but it seems difficult to convert properly the variables to such metadata. Except other argument I don't plan any work on this.

However in the same thread it appears useful to be able to remove all exif tags except some. Here comes the last commit which adds the possibility to enter '=' in the formula. This says that, despite exif tags are disabled, such a tag has to be exported.

image

The metadata tag dictionary looks now like that:
image

phweyland:
3.2... end of the year. :(

TurboGit:
@phweyland : Don't be sad, it's an open project we can discuss everything. I put 3.2 because I feel this could wait but we can reconsider. Maybe for 3.0.1 or 3.0.2 ?

phweyland:
There are 2 points in that PR.
I had left the first one on the to do list. That's the title of the PR.
The second one comes from an issue faced by a user for a use case I hadn't considered (do not export exif except some of them), issue #3944.
There is not urgency, but there is no more reason to wait. 
If not an issue for you, 3.0.1 would be fine.

TurboGit:
Manually merged on master and 3.0.1 branch.

PR #3957, opened 2019-12-30
darkroom : ensure we draw image at integer position

### LINKED ISSUE #3876: Drag-adjustable panels cause darkroom ui lag in certain width configurations
**Describe the bug:**

After trying to setup the darkroom view to my liking with the new drag-adjustable panels, I noticed that with certain panel widths the preview image and image adjustment sliders have serious lag (stutter/delay). On some panel widths everything is smooth as butter, seems to be quite arbitrary behaviour. And if I happen to find a config that works for one image, the same config causes lag when using it for an image with different dimensions/aspect ratio (from different camera). I also ran -d perf and it seems the lag is just ui lag and not affecting the actual processing speed.

**Steps to reproduce the behavior:**

Import image to darkroom, drag both left and right panels to different widths and then try adjusting exposure with different panel sizes.

**Expected behavior:**
No lag, smooth update of the preview image and no slider stutter.

**Screenshots:**
Configurations with and without lag:

No lag
no_lag

Lag
lag

Lag
lag2

**Platform:**
 - OS: Fedora Workstation 31
 - Version: darktable-3.1.0~git132
 - OpenCL on/off makes no difference
 - NVIDIA Quadro  w/ NVIDIA Driver 440.44

**Additional context**

 - Can you reproduce with another dt version?

No, 3.0.0. works smoothly with the hardcoded panel values (using panel width 322px) and similar configuration with filmstrip minimized and rating panel hidden.

AlicVB:
this fix #3876

Now that we use CAIRO_FILTER_BEST, to draw image buffer, we need to be sure that we stay at an integral value for the initial drawing point, otherwise, that would trigger some time expensive cairo filtering.
I've fallen back to previous CAIRO_FILTER_FAST filter in case of closeup, as we have to scale the image in this case, and this will automatically trigger filtering...

@aurelienpierre : as it's your code, can you have a look, please ?

### DIFF
--- a/src/views/darkroom.c
+++ b/src/views/darkroom.c
@@ -310,7 +310,7 @@ void expose(
     }
     cairo_paint(cr);
 
-    cairo_translate(cr, .5f * (width - wd), .5f * (height - ht));
+    cairo_translate(cr, ceilf(.5f * (width - wd)), ceilf(.5f * (height - ht)));
     if(closeup)
     {
       const double scale = 1<<closeup;
@@ -328,7 +328,11 @@ void expose(
 
     cairo_rectangle(cr, 0, 0, wd, ht);
     cairo_set_source_surface(cr, surface, 0, 0);
-    cairo_pattern_set_filter(cairo_get_source(cr), CAIRO_FILTER_BEST);
+    if(closeup)
+      cairo_pattern_set_filter(cairo_get_source(cr), CAIRO_FILTER_FAST);
+    else
+      cairo_pattern_set_filter(cairo_get_source(cr), CAIRO_FILTER_BEST);
+
     cairo_fill(cr);
 
     if(darktable.gui->show_focus_peaking)


aurelienpierre:
The change looks ok but I don't understand the issue. Cairo coordinates are `double`, how does that fit into the "keeping integer position" goal ?

jenshannoschwalm:
The ceilf are supposed to do the magic. Cairo performance is indeed very picky about this. Good point @AlicVB

AlicVB:
well, I don't know cairo internals... I found that by try and guess...
What I *may* have understand : in normal cases (no closeup) we don't change scale, so drawing image should be just a "copy" operation. But in the case where we call cairo_translate with non integral value, cairo seems to do some pixel filtering, instead of just copying them. This doesn't seems completely illogical...

AlicVB:
If you want to reproduce the issue, just ensure that : 
`cairo_translate(cr, .5f * (width - wd), .5f * (height - ht));` is called with non-integral values, for example : 
`cairo_translate(cr, .5f * (width - wd)+0.2f, .5f * (height - ht));`

You should end with a "laggy" drawing in darkroom

TurboGit:
Given the comment, I'm merging that for digging on this and fixing @AlicVB.

jenshannoschwalm:
@TurboGit and @AlicVB 

> Now that we use CAIRO_FILTER_BEST ...

This comment suggest there was a discussion about the filter used. I probably was not reading dt git at that time. I did a lot of cairo stuff while developing the graphics library for a forth system and it was obvious that CAIRO_FILTER_BEST is ~4 times slower on some machines compared to CAIRO_FILTER_BILINEAR or GOOD (this depends largely on the workload, cache size&speed and the cairo library).

I personally did not observe a relevant lagging problem but on slower systems this might be a real problem. Why not introduce a config parameter forcing a fallback to GOOD? Any objections?

PR #3964, opened 2019-12-30
Fix a simple mistake in translation which might have some grave consequences

tomaszg7:
Reported by @lkarcz.


TurboGit:
Thanks !

PR #3965, opened 2019-12-31
Fix compile error with gcc 7.4

matze:
Since in C a `static const` identifier is not considered a compile-time constant expression, gcc (at least 7.4 does) barks at using A when initializing RA and B:

    /home/matthias/dev/darktable/src/common/colorspaces.c: In function ‘_HLG_fct’:
    /home/matthias/dev/darktable/src/common/colorspaces.c:379:31: error: initializer element is not constant

This change replaces the expressions with explicit value assignments.


TurboGit:
Works for me. Thanks.

PR #3985, opened 2020-01-01
manual: enhanced denoise profiled summary in overview

VasylTretiakov:
mentioned a selection of denoise modules, of which denoise profiled is only one
included a reference to the denoise profiled module section
eliminated "denoise profile" in favor of "noise profile" from module's section
removed "single-click" phrase as "simple" seems enough to present the idea
removed "always" which is not very technical
removed "trade mark" sign which looks strange in doc for an open source product


EdgarLux:
> removed "trade mark" sign which looks strange in doc for an open source product
Hi @VasylTretiakov, this proposal has already been made and merged previously (4039).

VasylTretiakov:
In this pull request I am suggesting 6 changes. You are telling me that a later one (#4039 vs #3985) partially covers one (of the six).
If you want to ignore the suggested enhancements, please let me know and I'll close this pull request.

EdgarLux:
> In this pull request I am suggesting 6 changes. You are telling me that a later one (#4039 vs #3985) partially covers one (of the six).
> If you want to ignore the suggested enhancements, please let me know and I'll close this pull request.

What I mean is, maybe that's the cause of the merger conflict message. And it may be resolved if the other 5 changes are applied to the most recent file.

PR #3986, opened 2020-01-01
In crop & rotate, remove separate guides tab and don't stretch crop sliders vertically

### LINKED ISSUE #3968: Crop and rotate: merge 'main' and 'guides' tabs together
NOTE: Opening this issue as a result of a short discussion in https://github.com/darktable-org/darktable/pull/3601#issuecomment-569753233 so people can see it and voice their opinion.

In #3601 crop and rotate module got new sliders for adjusting the crop. Those were put into a separate tab, which is fine. But apart from that, guides-related controls were separated and put into 'guides' tab.

I worked a bit with the tabbed version of "crop and rotate" and I think separating the guides controls from the main tab was not a good idea. Margins - sure, those are perfect for a separate tab. But main and guides in my oppinion should be merged back together.

How I see it:
- Selecting guides is one of the most important features of the cropping module (this is what helps with the composition) and should be always readily available.
- There are some specific workflows which rely on frequent switching between guides:
    - image straightening (by eye): rotation is on the main tab, but switching to helpful  'metering' or 'grid' guides require changing tabs,
    - recomposition when changing image aspect: again - aspect is on the main tab.
- The amount of vertical space saved in modules area is negligible in case of most guides (with the exception of 'grid' guides, which are rather rarely used, I would assume).

In ~90% of my usage cases I personally find it irritating to have guides on a separate tab.

@TurboGit argues that current solution helps with the issue of the module changing its size, which is a valid point. But this could be remedied by reducing the number of sliders for the most problematic 'grid' guides.

dterrahe:
Based on discussions in #3601 and #3968.

Closes #3968.

### DIFF
--- a/src/iop/clipping.c
+++ b/src/iop/clipping.c
@@ -2032,11 +2032,9 @@ void gui_init(struct dt_iop_module_t *self)
   g->notebook = GTK_NOTEBOOK(gtk_notebook_new());
   GtkWidget *page1 = GTK_WIDGET(gtk_box_new(GTK_ORIENTATION_VERTICAL, 0));
   GtkWidget *page2 = GTK_WIDGET(gtk_box_new(GTK_ORIENTATION_VERTICAL, 0));
-  GtkWidget *page3 = GTK_WIDGET(gtk_box_new(GTK_ORIENTATION_VERTICAL, 0));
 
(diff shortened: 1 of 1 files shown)


TurboGit:
Thanks for the quick fix.

PR #3990, opened 2020-01-01
manual: rename views into perspectives

VasylTretiakov:
There are many places in the manual which refer to views in completely unrelated contexts, hence it is desirable to come up with another concept for darktable operation modes.
I feel like the concept of "perspective" is a good (possibly not the absolute best) option. It implies focus on specific workflow step and does not conflict with any other concept.
Hence, I suggest to evaluate this conceptual transition.
Other alternatives considered:
- module (used in Lightroom, but, obviously, already used in a different context in darktable)
- work mode
- operation mode
- work group


chrik5:
perspective is a term in photographie and does not seem to fit here.

In software context "viewport" is not unusual and using "view" seems reasonable for me as non English native.

TurboGit:
I'm not a native English speaker but to me a perspective is a different layout to convey the fact that the same view is better suited for such or such work. In some editor for software we have the debug perspective, that is the same as the editing one plus some panel with debug information.

Here the lighttable and darkroom are completely different views to me not just perspectives.

PR #4000, opened 2020-01-02
fix rating discard issue

### LINKED ISSUE #3971: switch back to 2.6.x order of pixelpipe
In  #3761 i have describe the problem with the new (3.0) order of the using vignette and splittoning.
Now I have the next problem:
I use the module monochrome to convert a picture into black and white. If I use a red or green filter the image gets darker. So I use the module levels to compensate the loss of light. In version 2.6.x it was executed **after** the monochrome module. Now with 3.0.0 it is run before, which leads to some problems. 
darktable_113
darktable_114
darktable_115
darktable_116

All other modules for tonecurve manipulation are also executed before monochrome. The only solution is moving the module in the pixelpipe and this has be done after each usage.

I have the impression that the arrangement of the pixelpipe in version 2.6.x was very well thought out. I never missed the order change function.
In 3.0.0 more theoretical aspects (monochrome is a creative module) were considered. This is not good.

Is it possible to use the Version 2.6.x order of the pixelpipe in version 3.0.0 by default? Perhaps via an option like 'auto-appy sharpen' .

### LINKED ISSUE #3972: Reject tag does not toggle anymore as it's only rejecting.
**Describe the bug**
Hitting the R keystroke or the dedicated icon on a hovered/selected picture will first mark it as discarded with the message "rejecting 1 image". However hitting the R keystroke once more (or using the dedicated icon) will not revert that state, and will also display the same message. 
Version 3.0.0 is fine with that (and does not display any message whatsoever).

**To Reproduce**
1. Go to the lighttable
2. Click on any picture you like (or dislike in this case)
3. Hit the R keystroke or use the dedicated icon to mark it as discarded
4. See that it's been marked correctly
5. Hit R once more or use the dedicated icon
6. See that the image is still marked as discarded

**Expected behavior**
The discarded state should be a toggle or a flag, not an irreversible state.

 - Darktable Version: 3.1.0+263 - 349fe88b419bbc7d87ee50382e8edac69a8a553f
 - OS: MacOS 10.15
 - OpenCL activated

**Additional context**
3.0.0 does not exhibit this behaviour and does not display any message upon mark an image as discarded.

phweyland:
Should fix #3972 .
Note: on 3.0.0 the message was only displayed for selection. Now it's displayed both when hovering or for selection, which is more consistent.
Note2: Discard symbol is absent from the bottom view:
image
Is that intentional ?

### DIFF
--- a/src/common/ratings.c
+++ b/src/common/ratings.c
@@ -23,6 +23,7 @@
 #include "common/ratings.h"
 #include "common/undo.h"
 #include "common/grouping.h"
+#include "views/view.h"
 #include "control/conf.h"
 #include "control/control.h"
 #include "gui/gtk.h"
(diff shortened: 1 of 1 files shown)


ptilopteri:
does this restore the toggle effect of the "r" key or is it a "work-a-round"

phweyland:
> does this restore the toggle effect of the "r" key

should restore it.

ptilopteri:
> > does this restore the toggle effect of the "r" key
> 
> should restore it.

tks, part of my remembered keys

TurboGit:
> Note: on 3.0.0 the message was only displayed for selection. Now it's displayed both when hovering or for selection, which is more consistent.

I don't feel this is right. For selection it makes sense, but when hovering on a single picture it is clear that it is applied or not the start or color label changes. In this case the message is more noise than necessary to me.

phweyland:
> > Note: on 3.0.0 the message was only displayed for selection. Now it's displayed both when hovering or for selection, which is more consistent.
> 
> I don't feel this is right. For selection it makes sense, but when hovering on a single picture it is clear that it is applied or not the start or color label changes. In this case the message is more noise than necessary to me.

Not sure ... Applying to grouped images it make sense to display the number of modified images.

TurboGit:
So maybe just kill the message if the number of image is 1 ? I really find it too disturbing to have this message pop-up all the time! Maybe just me... Let's see what others think.

phweyland:
> So maybe just kill the message if the number of image is 1 ?

Agreed. I do that.

phweyland:
BTW, only rating and geotags issue messages. Color labels, metadata and tagging don't.

TurboGit:
Thanks, looks good now.

PR #4002, opened 2020-01-02
remove refs to "panel_width" setting

AlicVB:
there's no more setting like that, as panels are resizable...
What I've done : 
1. fix popup dialog that depends from panel_width to 350 (last value)
2. fix histogram and navigation height to 175 (1/2 of last value) libs height are not dependent from lib width anymore, but I find that more pleasant, otherwise, setting panel size to low or height values, result in unusable sizes (not to say that we keep some uniformity for left and right panels)

Note that I've used "raw" values, for sizes, like it was before, but I wonder if we should apply some dpi/ppd/... transformation here...

### DIFF
--- a/src/gui/gtk.c
+++ b/src/gui/gtk.c
@@ -1427,15 +1427,6 @@ static void init_widgets(dt_gui_gtk_t *gui)
 
   configure_ppd_dpi(gui);
 
-  // set constant width from conf key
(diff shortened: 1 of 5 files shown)


TurboGit:
Manually merged in master and 3.0.1.

TurboGit:
Thanks!

PR #4003, opened 2020-01-02
doc for last panel changes

### LINKED ISSUE #3992: [user manual] Outdated description for Filmstrip in chapter 1.2.3
I think the description in chapter 1.2.3 is not up to date anymore, because with the new UI the filmstrip is only available in darkroom and not in lightroom like in the previous versions. This fact should be clearly emphasized.

AlicVB:
Note that I'm not sure which version is targeted by the usermanual... This is for the 3.01 one
This fix #3992

### DIFF
--- a/doc/usermanual/lighttable/panels/timeline.xml
+++ b/doc/usermanual/lighttable/panels/timeline.xml
@@ -25,7 +25,7 @@
   </para>
 
   <para>
-    In lighttable, you can display a timeline in the bottom panel with the shortcut <emphasis>Ctrl+T</emphasis>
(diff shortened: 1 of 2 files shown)


mepi0011:
Hi @AlicVB ,

IMHO and as a non-native speaker I find it a bit strange to call lighttable now lighttable filemanager.
If it the new spelling it should change in the hole document.

AlicVB:
This is because there's multiple layouts in lighttable : filemanager, culling, preview, ... and in this case, filmstrip is available for culling and preview, but refers to what is set in filemanager...
Now, I'm quite bad in English, so if someone find a better phrasing...

Issue #3998, opened 2020-01-02
No automatic RAW+JPEG grouping


fxthomas:
**Describe the bug**

Couples of RAW+JPEG are not grouped when importing them in Darktable 3.0.

The grouping behavior was also wonky in Darktable 2.6 (images were sometimes grouped, sometimes not).

I use a Canon 550D, if that's useful.

**To Reproduce**
1. Insert SD card containing RAW+JPEG images
2. Click on "Mass Storage Camera > Import from camera"
3. Select all images
4. Click "Import"

**Expected behavior**
Couples of images with the same name (e.g. IMG_1234.CR2 and IMG_1234.JPG) should be imported as a single group.

**Screenshots**

During import:
![Screenshot from 2020-01-02 13-38-12](https://user-images.githubusercontent.com/613594/71667528-ca778a00-2d65-11ea-93ad-54c03be95c5c.png)

Grouping enabled in lighttable, no group shown:
![Screenshot from 2020-01-02 13-43-37](https://user-images.githubusercontent.com/613594/71667553-e9761c00-2d65-11ea-83c3-c310050793bf.png)

**Platform (please complete the following information):**
 - Darktable Version: 3.0
 - OS: Arch Linux
 - OpenCL active
 - GPU: Intel HD Graphics 530 (Skylake GT2)

jenshannoschwalm:
For automatic grouping dt requires the filenames to be the same except the extension. I think the default sets this to a job number. You should have a look into the options ...

jenshannoschwalm:
@TurboGit and @aurelienpierre i'd like your opinion about this issue.

The situation is pretty clear: While importing the filename is **not always** kept as is was on the medium/camera, so when looking in the database for other files with the same filename-minus-extension for automatic grouping it fails.

I see two ways to tackle this.

1. To test for grouping we should look for all files with the same film_id to have the same exif original_filename, this would have a heavy performance penalty.
2. Also keep original filename in the database and check for that.

For both situations we would have to accept, not all images have that exif data available.

Any other idea? Or just close this as "can't be fixed"

jenshannoschwalm:
@fxthomas i guess this is fixed now, please close issue if you are happy with current master.

fxthomas:
Thank you so much, let me try to compile that and I'll give it a go!

fxthomas:
It's working nicely if I select the "keep original filename" option in the settings (I guess that's the intended behavior). Thank you!

PR #4012, opened 2020-01-03
Prevented premature iteration of filenames

sams96:
This stops the filename generator from iterating on it's first attempt, which was stopping the grouping of RAW+JPG groups, fixing #3998.

However it will assume that the next file imported with a different file type should be grouped with the current file. This also starts the filenames at 0 instead of .

Ideally we would compare the actual filenames before import but there doesn't seem to be a way to access that information from import_session at the moment.


TurboGit:
> However it will assume that the next file imported with a different file type should be grouped with the current file. This also starts the filenames at 0 instead of .

IIUC this fix seems worst than the current behavior (not grouping RAW+jpg).

PR #4013, opened 2020-01-03
fix #3978: importing all duplicates in Windows

### LINKED ISSUE #3978: Re-importing a folder ignores second xmp files created for duplicates
Using the Duplicate function in lighttable mode creates a second xmp file with _01 in the file name. If the photo or its folder is removed under Collect Images, then re-imported, darktable only recognizes the original xmp file and not the second xmp made from the duplicate.  

Darktable Version: 3.0
Windows 10
OpenCL  not activated

**Based on comments below and on pixls.us this seems to be a problem just on the Windows version**

Mark-64:
Fix for #3978 
Also fixed a related bug, leaving .xmp files behind when trashing images

### DIFF
--- a/src/common/image.c
+++ b/src/common/image.c
@@ -34,6 +34,7 @@
 #include "control/control.h"
 #include "control/jobs.h"
 #include "develop/lightroom.h"
+#include "win/filepath.h"
 #ifdef USE_LUA
 #include "lua/image.h"
 #endif
@@ -47,6 +48,7 @@
 #include <glob.h>
 #endif
 #include <glib/gstdio.h>
+
 
 static int64_t max_image_position()
 {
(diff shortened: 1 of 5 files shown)


src/common/image.c
```diff
@@ -999,8 +1005,39 @@ void dt_image_read_duplicates(const uint32_t id, const char *filename)
...
+			
+					while ((c3 <= c4) && valid_filename)
+					{
+						if (!( *c3 >= '0' && *c3 <= '9' )) valid_filename = false;
+						c3++;
+					}
+				
+				}
+				if (valid_filename) files = g_list_append(files, g_build_filename(imgpath, file, NULL));
+				g_free(file);
```
TurboGit: This code is duplicated below. I would like to see that put into a separate routine maybe src/win/filepath.c ?

Also not that we do not use bool (C++) but gboolean and there is no space after if, while.

src/win/filepath.c
```diff
@@ -0,0 +1,62 @@
+/*
+    This file is part of darktable,
+    copyright (c) 2009--2012 johannes hanika.
+    copyright (c) 2010--2012 tobias ellinghaus.
```
TurboGit: Please remove those lines. You can probably add your name here but I feel it should better to remove all copyright in the sources. Does not makes sense at this time almost no one update them.

src/win/filepath.h
```diff
@@ -0,0 +1,28 @@
+/*
+    This file is part of darktable,
+    copyright (c) 2009--2012 johannes hanika.
+    copyright (c) 2010--2012 tobias ellinghaus.
```
TurboGit: Likewise.

src/common/image.c
```diff
@@ -999,8 +1007,8 @@ void dt_image_read_duplicates(const uint32_t id, const char *filename)
       do
       {
         char *file = g_utf16_to_utf8(data.cFileName, -1, NULL, NULL, NULL);
-        files = g_list_append(files, g_build_filename(imgpath, file, NULL));
-        g_free(file);
+				if(win_valid_duplicate_filename(file)) files = g_list_append(files, g_build_filename(imgpath, file, NULL));
```
TurboGit: Looks like too much spaces here or tabulations are used?

Mark-64:
changes requested implemented

TurboGit:
Thanks  for the first contribution ! And you are fixing a long standing bug, congrats!

TurboGit:
Also merged in 3.0.1 branch.

PR #4027, opened 2020-01-04
fix #3780 live samples position

### LINKED ISSUE #3780: UI: color picker sample areas positions displayed incorrectly
In color picker, live samples areas are displayed to an offset. The more samples, the bigger the offset.

Steps to reproduce:
1) In color picker select "display sample areas on image"
2) click on the eyedropper in area mode
3) select an area, click add to live samples
4) select another area, click add
5) repeat

You will see the blue box shifted with respect to the areas selected, the shift increases with the number of areas.

For example see below. Each selection was done precisely on one patch of the greyscale

Capture2

DT 3.0 RC2 on Windows 10, default theme

Mark-64:
Fixing "UI: color picker sample areas positions displayed incorrectly" #3780 
There was a cairo translation (used to reinforce the line width, I guess) that was not reverted and was accumulating across live samples, producing a progressive shift


TurboGit:
Perfect! Thanks.

PR #4032, opened 2020-01-05
Fake support for dual pixel dng files; fixes #4006

### LINKED ISSUE #4006: Add support for Dual Pixel Raw dng files
**Describe the bug**
A DNG file converted  frrom a CR3 file via Adobe DNG Converteer, Dual Pixel Raw) is not being opened in darktable. Rawspeed issue 214 indicates that this is possibly an error from darktable. Thumbnail of the Dng file is not being generated. While opening the file darktable is running into a deadlock.

Error:
void rawspeed::AbstractDngDecompressor::decompress() const, line 208: Too many errors encountered. Giving up. First Error:
rawspeed::LJpegDecompressor::LJpegDecompressor(const rawspeed::ByteStream&, const rawspeed::RawImage&), line 42: Unexpected component count (2)
allocation failed???

Issue 214 rawspeed:
While it is easy to convince rawspeed to load it
diff --git a/src/librawspeed/decompressors/LJpegDecompressor.cpp b/src/librawspeed/decompressors/LJpegDecompressor.cpp
index f5d6d86c..64b2fb2b 100644
--- a/src/librawspeed/decompressors/LJpegDecompressor.cpp
+++ b/src/librawspeed/decompressors/LJpegDecompressor.cpp
@@ -39,6 +39,7 @@ LJpegDecompressor::LJpegDecompressor(const ByteStream& bs, const RawImage& img)
     ThrowRDE("Unexpected data type (%u)", mRaw->getDataType());
 
   if (!((mRaw->getCpp() == 1 && mRaw->getBpp() == 2) ||
+        (mRaw->getCpp() == 2 && mRaw->getBpp() == 4) ||
         (mRaw->getCpp() == 3 && mRaw->getBpp() == 6)))
     ThrowRDE("Unexpected component count (%u)", mRaw->getCpp());

**Expected behavior**
Darktable will show thumbnail in lighttable. Darktable will open DNG file

**Screenshots**

 - Darktable Version: 3.1.0
 - OS: Ubuntu 19.10
 - <!-- OpenCL activated

jenshannoschwalm:
This pr handles the dt side of dual pixel raw files, besides this there are
two wrong checks in rawspeed code. One has already been fixed by @LebedevRI
in the develop branch, the second fix has been suggested.

Fixing issue #4006

### DIFF
--- a/src/common/imageio_rawspeed.cc
+++ b/src/common/imageio_rawspeed.cc
@@ -240,14 +240,16 @@ dt_imageio_retval_t dt_imageio_open_rawspeed(dt_image_t *img, const char *filena
 
     const int cpp = r->getCpp();
 
-    // checking for corrupted info supports cpp to be either 1 or 2
+    // checking for corrupted info supports cpp to be either 1 or 2 for TYPE_USHORT16
+    // for TYPE_FLOAT32 only cpp==1 is accepted
     if((r->getDataType() != TYPE_USHORT16) && (r->getDataType() != TYPE_FLOAT32)) return DT_IMAGEIO_FILE_CORRUPTED;
 
-    if(((r->getBpp() / cpp) != sizeof(uint16_t)) && (r->getBpp() != sizeof(float))) return DT_IMAGEIO_FILE_CORRUPTED;
+    if((r->getBpp() != cpp * sizeof(uint16_t)) && (r->getBpp() != cpp * sizeof(float))) return DT_IMAGEIO_FILE_CORRUPTED;
 
-    if((r->getDataType() == TYPE_USHORT16) && ((r->getBpp() / cpp) != sizeof(uint16_t))) return DT_IMAGEIO_FILE_CORRUPTED;
+    if((r->getDataType() == TYPE_USHORT16) && (r->getBpp() != cpp * sizeof(uint16_t))) return DT_IMAGEIO_FILE_CORRUPTED;
 
-    if((r->getDataType() == TYPE_FLOAT32) && (r->getBpp() != sizeof(float))) return DT_IMAGEIO_FILE_CORRUPTED;
+    // this also assures cpp==1
+    if((r->getDataType() == TYPE_FLOAT32)  && (r->getBpp() != sizeof(float))) return DT_IMAGEIO_FILE_CORRUPTED;
 
     if((cpp < 1) || (cpp>2)) return DT_IMAGEIO_FILE_CORRUPTED;
 
@@ -369,6 +371,7 @@ dt_imageio_retval_t dt_imageio_open_rawspeed(dt_image_t *img, const char *filena
       }
       else
       { 
+        // This code should never be reached because of a test above
         fprintf(stderr,"\n[rawspeed] (%s) has unsupported dual pixel FLOAT",img->filename);
         return DT_IMAGEIO_FILE_CORRUPTED;
       }
--- a/src/common/imageio_rawspeed.cc
+++ b/src/common/imageio_rawspeed.cc
@@ -326,6 +326,10 @@ dt_imageio_retval_t dt_imageio_open_rawspeed(dt_image_t *img, const char *filena
     // if buf is NULL, we quit the fct here
     if(!mbuf) return DT_IMAGEIO_OK;
 
+    // we have to set the planes in the img struct now to allow a proper mipmapcache data size
+    if(img->buf_dsc.datatype == TYPE_UINT16) 
+      img->pixel_rawplanes = cpp;
+
     void *buf = dt_mipmap_cache_alloc(mbuf, img);
     if(!buf) return DT_IMAGEIO_CACHE_FULL;
 
@@ -340,7 +344,8 @@ dt_imageio_retval_t dt_imageio_open_rawspeed(dt_image_t *img, const char *filena
 
     /*
       For dual pixels we can't just copy or use the flip as data are interwoven.
-      Here we use the medium of both sub-pixels, not worth to use _OPENMP.
+      Instead we copy the data organized in a way like standard cfa patterns so we can
+      use existing rawprepare algos.
       As there are no dual-float-pixel images we have them disabled.
     */
     if (cpp == 1)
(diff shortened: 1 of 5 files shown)


src/common/imageio_rawspeed.cc
```diff
@@ -237,21 +237,22 @@ dt_imageio_retval_t dt_imageio_open_rawspeed(dt_image_t *img, const char *filena
       dt_imageio_retval_t ret = dt_imageio_open_rawspeed_sraw(img, r, mbuf);
       return ret;
     }
+    const int cpp = r->getCpp();
```
LebedevRI: newline

src/common/imageio_rawspeed.cc
```diff
@@ -333,16 +334,43 @@ dt_imageio_retval_t dt_imageio_open_rawspeed(dt_image_t *img, const char *filena
...
-      dt_imageio_flip_buffers((char *)buf, (char *)r->getDataUncropped(0, 0), r->getBpp(), dimUncropped.x,
-                              dimUncropped.y, dimUncropped.x, dimUncropped.y, r->pitch, ORIENTATION_NONE);
+      if(img->buf_dsc.datatype == TYPE_UINT16)
+      {
+        for(int j = 0; j < img->height; j++)
+        {
+          const uint16_t *in = (uint16_t *) r->getDataUncropped(0, j);
+          uint16_t *out = ((uint16_t *)buf) + (size_t) j * img->width;
+          for(int i = 0; i < img->width; i++)
+            out[i] = (in[i*2] + in[i*2+1]) / 2;
```
LebedevRI: 1. This doesn't really look like correct dual pixel handling - is averaging really the correct handling?
2. how do we know this handling is correct for every cpp=2 file out there?

i suspect it should be an iop

src/common/imageio_rawspeed.cc
```diff
@@ -237,21 +237,22 @@ dt_imageio_retval_t dt_imageio_open_rawspeed(dt_image_t *img, const char *filena
...
     if((r->getDataType() == TYPE_FLOAT32) && (r->getBpp() != sizeof(float))) return DT_IMAGEIO_FILE_CORRUPTED;
 
-    const float cpp = r->getCpp();
-    if(cpp != 1) return DT_IMAGEIO_FILE_CORRUPTED;
+    if((cpp < 1) || (cpp>2)) return DT_IMAGEIO_FILE_CORRUPTED;
 
     img->buf_dsc.channels = 1;
 
-    switch(r->getBpp())
+    switch(r->getBpp() / cpp)
```
LebedevRI: This also accepts `2*sizeof(float)`. (I.e. it assumes it would have been discarded above)
I think you want
```
    switch(r->getBpp() / cpp)
    {
      case sizeof(uint16_t):
        img->buf_dsc.datatype = TYPE_UINT16;
        break;
      case 2 * sizeof(uint16_t):
        img->buf_dsc.datatype = TYPE_UINT16;
        break;
      case sizeof(float):
        img->buf_dsc.datatype = TYPE_FLOAT;
        break;
      default:
        return DT_IMAGEIO_FILE_CORRUPTED;
    }
```

src/common/imageio_rawspeed.cc
jenshannoschwalm: I think you are right here in a general sense. For these canon dual pixels it's correct but might be completely wrong in other cases.

jenshannoschwalm: > i suspect it should be an iop

I don't understand ...

src/common/imageio_rawspeed.cc
jenshannoschwalm: `2 * sizeof(uint16_t)` is the same as `sizeof(float)` for most compilers / os so the compiler mocks about two same test cases.

In line 249 there is already a test for >1channel floats ...

src/common/imageio_rawspeed.cc
jenshannoschwalm: ok

src/common/imageio_rawspeed.cc
LebedevRI: Ah right.
Add a comment next to the switch explaining that than.

src/common/imageio_rawspeed.cc
LebedevRI: I don't know what darktable does internally nowadays, but it really doesn't look correct to do
such an operation *here*, not in the pixel pipeline, if requested by some module (iop).

src/common/imageio_rawspeed.cc
jenshannoschwalm: done

src/common/imageio_rawspeed.cc
jenshannoschwalm: Aah. This pr is **not** intended as something clever to do with dual-pixel images, for that we would of course need to pass all data of the images in an ordered manner and make access to that data valid in the pixel pipeline. That would require much more work and changes and i doubt it's really worth it. (Personally i absolutely do not care about those dual pixel files). And so far none of the devs ever seemed to need/implement this feature... missing something maybe...

This pr **only** presents the raw data of one of the images so the rest of the pipeline can access that data as if there was only one. Thus i thought it would be ok this way as we do the copy-data only once when reading into the cache.

At least people can use these dng files as if they were just-raw only missing the 'candy-features' of dual-pixel.

parafin: I wonder if user has to be informed somehow that this is not a real dual pixel support.

jenshannoschwalm: In release note or as a hint message when opening?

parafin: dunno... At least one of those;)

LebedevRI: That didn't really answer my second question.
Even if it results in not-totally-terrible combined images for these dual-pixel raws,

> 2. how do we know this handling is correct for every cpp=2 file out there?

jenshannoschwalm: 1. They are not _not-totally-terrible combined images_, what we see here is exactly one of two images, what we don't is some clever algorithm combining both.

2. An yes, i am not totally sure about **every** file, unfortunately there is no valid documentation i found so far. Do you know of any?

jenshannoschwalm:
I did some searching on available information about these dual pixel files. The SamplesPerPixel exif data should be understood as: we have two consecutive values (16bit) per component. There is no documented standard how these should be taken, there are comments telling they can be interpreted independently, probably the first value is avaraged when created by adobe converter. What the canon software does is not documented, no public algorithm.

So what can be done? We could either present the complete dataset of both pixels to the pipeline and adopt the matrix algorithms. Very tricky, likely lousy performance and certainly out of my competence. 

Second option would be to make both planes available to the pipeline independently so the data of the image can be processed as before. That's what i do here.

The image struct holds an additional value, the number of pixel planes. The mipmap cache must also hold both datasets, these are filled in by `dt_imageio_open_rawspeed`. As we want the data to be aligned the mipmap allocation had to be changed a bit, for dual pixel images it holds both datasets plus the one that is presented to the pipeline. Right now - and as default - this is identical with the the first data set.

As this is not a "complete" dual pixel support (i don't like the word fake, it too much reminds me of folks in DC) there is a information message when opening such files in darkroom.

What could be done further? In rawprepare there should be at a switch to select one of the pixel data sets (planes) or implement more clever stuff. This would be another pr anyway.

jenshannoschwalm:
As discussed in darktable-org/rawspeed#216 there are pending issues related to blacklevels - plus likely more - regarding dual/multiple pixel images. Closing this now and see what can be done on rawspeed side. If that can be handled in a proper way i will do another pr depending on available rawspeed bound into dt.

PR #4035, opened 2020-01-05
crop and rotate: use floating point numbers for keystone lines

flannelhead:
Using integers here resulted in misplaced lines especially when zoomed in (lines did not go through the control circles).


TurboGit:
Better indeed, works for me. Thanks.

PR #3726, opened 2019-12-11
left & right panels resizing

AlicVB:
This fix #3710 
- this allow user to change the width of left & right panels with the mouse.
- there's no change to the ui.
- sizes are store per panel, views & layout, like panels state.
- default size is set in the code to 350.

PR #4041, opened 2020-01-06
histogram: panel resize fix

dtorop:
The buffer holding the waveform histogram needs to be re-allocated now that the right panel can resize (since #3726). Also, the buttons on the histogram need to be repositioned depending on panel width.

Without this PR, darkroom view will crash if you perform the following steps:

1. set histogram mode to waveform
2. resize the right panel

I wonder if this PR could be written better. It's a quick fix. Is it necessary to recalculate the preview to generate a new histogram on every resize of the right panel in waveform mode? Would it be better to just resize the waveform histogram data itself, and lazily/later recalculate the preview?

Also, would it be better to make the buttons on the histogram not depend on the widget width, but simply be right justified?

### DIFF
--- a/src/develop/develop.c
+++ b/src/develop/develop.c
@@ -80,7 +80,6 @@ void dt_dev_init(dt_develop_t *dev, int32_t gui_attached)
   dt_pthread_mutex_init(&dev->pipe_mutex, NULL);
   dt_pthread_mutex_init(&dev->preview_pipe_mutex, NULL);
   dt_pthread_mutex_init(&dev->preview2_pipe_mutex, NULL);
-  //   dt_pthread_mutex_init(&dev->histogram_waveform_mutex, NULL);
   dev->histogram = NULL;
   dev->histogram_pre_tonecurve = NULL;
   dev->histogram_pre_levels = NULL;
@@ -153,7 +152,6 @@ void dt_dev_cleanup(dt_develop_t *dev)
   dt_pthread_mutex_destroy(&dev->pipe_mutex);
   dt_pthread_mutex_destroy(&dev->preview_pipe_mutex);
   dt_pthread_mutex_destroy(&dev->preview2_pipe_mutex);
-  //   dt_pthread_mutex_destroy(&dev->histogram_waveform_mutex);
   if(dev->pipe)
   {
     dt_dev_pixelpipe_cleanup(dev->pipe);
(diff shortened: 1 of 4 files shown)


dtorop:
Also, more questions:
- Should the histogram catch an event to know that the width has changed? This would be nicer than checking if `dev->histogram_waveform_width != width`. See https://github.com/darktable-org/darktable/blob/35ca1e0cb641f2db040942261b3ff202b7ddc62a/src/libs/histogram.c#L195-L197
- Do we need a mutex? See https://github.com/darktable-org/darktable/blob/35ca1e0cb641f2db040942261b3ff202b7ddc62a/src/develop/develop.h#L185-L187

TurboGit:
>     * Should the histogram catch an event to know that the width has changed? This would be nicer than checking if `dev->histogram_waveform_width != width`. See

I agree probably nicer and safer.

dtorop:
> > ```
> > * Should the histogram catch an event to know that the width has changed? This would be nicer than checking if `dev->histogram_waveform_width != width`. See
> > ```
> 
> I agree probably nicer and safer.

I'll work on a version that does this.

dtorop:
Some further changes pushed to this PR. This version should work better. Improvements in this PR:
- Fix crash on right panel resize when waveform histogram is selected.
- Resizing right panel is smoother when waveform histogram is selected, due to only running the preview pixelpipe post-processing to generate a histogram, rather than running the entire preview pipe then redrawing the entire workspace.
- A corresponding change also makes it faster to enter waveform histogram mode.
- Unify allocation code for regular/waveform histogram on histogram draw. Previously the regular histogram buffer was always allocated even when it wasn't visible.
- Properly (I hope) wrap code in mutexes, so that histogram redraw waits until after any current preview pipe has run. Also, don't reallocate the waveform histogram buffer while the preview pipe is running and possibly writing to that buffer.
- Remove vestigial `histogram_waveform_mutex` which would have offered finer grain locking, which doesn't seem necessary as the code here is already pretty smooth on panel resize. Also, using a more fine grained lock could cause the histogram to redraw once while the preview pipe is running, and again once the preview pipe post-processing has finished and has calculated a new histogram.

dtorop:
I need a moment more to check if I missed something on the locking, so this should still be WIP.

dtorop:
This PR should be ready now for testing/review. I added one more commit, to try to be extra-safe about locking.

TurboGit:
Thanks ! Manually merged into master and 3.0.1.

dtorop:
Thank you. It's good to see the panels resizing smoothly!

PR #4044, opened 2020-01-06
Only show "flip guides" dropdown for asymmetrical guides

dterrahe:
In the crop & rotate module, when selecting to show guidelines, there always appears a dropdown to allow to flip horizontally or vertically. This is not useful for symmetrical guides, like grid, but it does add another widget to an already crowded module.

This change allows guideline definitions to specify if a "flip" dropdown should be shown.

For now, this doesn't change the lua interface; a flip dropdown will still always be shown for those.


TurboGit:
Looks good! Thanks.

Issue #3936, opened 2019-12-29
Resize 'Collect Images' module


Egocentrix:
I often use the Collect Images module in folder mode to browse my library, and I have quite a lot of folders so sometimes it would be nice for the module to be taller.

The new tagging module (kudos for that, btw, huge improvement) can be resized by ctrl+scroll. Would love to be able to do the same for the collect module!

Egocentrix:
Solved by #4046

PR #4046, opened 2020-01-06
Resize 'collect images' module with ctrl+scroll

Egocentrix:
The 'collect images' module can now be resized by ctrl+scroll, in the same way that the tagging module works. This resolves issue #3936.

I plan to save the size with dt_conf, after I figure out how that works.


jpverrue:
very good news !

TurboGit:
Thanks! Works for me.

PR #4057, opened 2020-01-08
histogram draw cleanup

dtorop:
- high dpi fix, particularly so that histogram is the right height
- slightly cleaner button position math, should not produce any change to current behavior


TurboGit:
Manually merged in master and 3.0.1. Thanks!

PR #4063, opened 2020-01-08
Save/restore collect module height

Egocentrix:
Follow-up for #4046, the height of the collect window is now saved in darktablerc and restored on restart.


TurboGit:
Looks ok! Thanks.

Issue #4067, opened 2020-01-08
darktable crash with gmic library


jpverrue:
@phweyland
**Describe the bug**
I just installed the gmic library needed for the new 3D lut module, then I recompiled it without any problem. But since then, darktable crashes whenever I try to do anything. More precisely, I can launch darktable without any problem, the LT interface is displayed, I can open and close modules, scroll in my pictures, but as soon as I try to go to DR, zoom an image (W), launch an export and "à priori" any action (I haven't tested everything), darktable crashes with the following error message :
```
/opt/darktable/bin/darktable: symbol lookup error: /opt/darktable/lib64/darktable/plugins/liblut3d.so: undefined symbol: _ZN12cimg_library8CImgListIfE6assignEj
```
**Platform (please complete the following information):**
 - Darktable Version: 3.1.0+320~g218dd5d8c
 - OS: opensuse tumbleweed 20200105
 
**Additional context**
Packages loaded
```
i+ | libgmic-devel       | package | 2.8.1-1.1 | x86_64 | Dépôt principal (OSS) (20190718)
i  | libgmic1            | package | 2.8.1-1.1 | x86_64 | Dépôt principal (OSS) (20190718)
```
I removed the packages and recompiled. darktable is working again without any problem.

kikoucalou:
Probably somethink similar,  DT crash when lauding image with " ... took 0,236 secs (0,422 CPU) to load the image.
terminate called after throwing an instance of 'gmic_exception' ..."

Platform :
+ darktable version 3.1.0+323~gaae21bfd5
+ compiled on Linux Mint 19.3 Tricia

Libgmic1 1.7.9+zart-4build3
libgmic-devel 1.7.9+zart-4build3

kikoucalou:
When reverting to darktable version 3.1.0+308~g2f82a6e43, DT run nicely

phweyland:
There two issues there:
One your side. GMIC library must be >= 2.7.0.
One on dt side, because if the version is less than 2.7.0 it should not try to use GMIC.

phweyland:
@jpverrue, your gmic version seems to be the last one.
There are probably some missing dependencies, but the `undefined symbol: _ZN12cimg_library8CImgListIfE6assignEj` doesn't help me for the moment.
I'll try to find the list of gmic dependencies and I'll post it here.

phweyland:
@jpverrue, did you open the lut3d module and select a compressed lut ?

phweyland:
[...]

PR #4072, opened 2020-01-09
lut3d - improve live gmic version detection

phweyland:
Try to improve gmic version detection and possibly solve the issue #4067 .
Two main points
- consider versions with 4 digits instead of 3
- check both the build time version and the run time version.


TurboGit:
I'm not in a position to check that but it looks ok so I'll merge. Thanks!

phweyland:
Thanks
I've made another small improvement (avoid to call at each init the run time library to check the version), but I will wait if we don't have another fix to do on the current issue.

PR #4074, opened 2020-01-09
Folder, Directory, Filmroll. When to use what?

codingdave:
In darkroom clicking on "import"->"folder" opens a dialog with the title "import film". This is unexpected because the dialog enables the user to import a folder, not a "film". The user might expect that "film" is a container containing all images (e.g. an archive) or a reference file that lists all images being part of the "film" (e.g. a xml file). 

This fix only renames the title of the dialog to get rid of this confusion and updates the documentation accordingly. 

What is missing is a correct translation of all the .po files. But I leave this to the translators.

Suggestion: Is it possible to make the translators aware of the missing
translation by removing the existing translation for "film roll"?

### DIFF
--- a/doc/usermanual/lighttable/panels/import.xml
+++ b/doc/usermanual/lighttable/panels/import.xml
@@ -78,7 +78,7 @@
 
     <para>
       As the name implies, checking <quote>import folders recursively</quote> will import
-      images in the currently selected directory, and all subdirectories. It is not recommended,
+      images in the currently selected folder, and all subfolders. It is not recommended,
       and wastes resources, to do this on an exhaustive list of images. darktable would generate
       thumbnails for all of them, but in the end only keep the recent ones in its cache. It is
       better to import images in smaller chunks, making logical film rolls.
@@ -143,8 +143,8 @@
       </para>
 
       <para>
-        You define the base directory for storing imported images and the naming pattern of sub
-        directories and individual images in the preferences dialog (see
+        You define the base folder for storing imported images and the naming pattern of subfolders 
+        and individual images in the preferences dialog (see
         <xref linkend="session_options"/>).
       </para>
 
--- a/doc/usermanual/lighttable/panels/import.xml
+++ b/doc/usermanual/lighttable/panels/import.xml
@@ -77,7 +77,7 @@
     </para>
 
     <para>
-      As the name implies, checking <quote>import directories recursively</quote> will import
+      As the name implies, checking <quote>import folders recursively</quote> will import
       images in the currently selected directory, and all subdirectories. It is not recommended,
       and wastes resources, to do this on an exhaustive list of images. darktable would generate
       thumbnails for all of them, but in the end only keep the recent ones in its cache. It is
--- a/src/common/film.c
+++ b/src/common/film.c
@@ -279,7 +279,7 @@ int dt_film_import(const char *dirname)
   // when called without job system running the import will be done synchronously and destroy the film object
   int filmid = film->id;
 
-  /* at last put import film job on queue */
+  /* at last put import folder job on queue */
   film->last_loaded = 0;
   g_strlcpy(film->dirname, dirname, sizeof(film->dirname));
   char *last = &film->dirname[strlen(film->dirname) - 1];
--- a/src/common/film.c
+++ b/src/common/film.c
@@ -279,7 +279,7 @@ int dt_film_import(const char *dirname)
   // when called without job system running the import will be done synchronously and destroy the film object
   int filmid = film->id;
 
-  /* at last put import folder job on queue */
+  /* at last put import film job on queue */
   film->last_loaded = 0;
   g_strlcpy(film->dirname, dirname, sizeof(film->dirname));
   char *last = &film->dirname[strlen(film->dirname) - 1];
--- a/src/common/import_session.c
+++ b/src/common/import_session.c
@@ -152,7 +152,7 @@ void dt_import_session_destroy(struct dt_import_session_t *self)
 {
   if(--self->ref != 0) return;
 
-  /* cleanup of session import folder */
+  /* cleanup of session import film roll */
   _import_session_cleanup_filmroll(self);
 
   dt_variables_params_destroy(self->vp);
--- a/src/common/import_session.c
+++ b/src/common/import_session.c
@@ -152,7 +152,7 @@ void dt_import_session_destroy(struct dt_import_session_t *self)
 {
   if(--self->ref != 0) return;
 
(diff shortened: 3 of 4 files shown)


TurboGit:
Well not so sure, a film roll is a term used in dt to be equivalent to a folder. A film in dt is a folder in fact.

Nilvus:
Folder, why not but in reality that's not inconsistent. darktable use lot of terms related to old analogical cameras and paper film. So using film here is consistent with this logic.

codingdave:
- The Folder Metaphor [1] says: "There is a difference between a directory, which is a file system concept, and the graphical user interface metaphor that is used to represent it (a folder)"
- Looking at the variable $(FILE_**FOLDER**) and its explanation it is very visible that darktable wants to use the term **folder**.

Currently we have several words meaning the same:
- Folder
  - import.c: select a **folder** to import as **film roll**
  - gtk_entry: $(FILE_**FOLDER**) - **folder** containing the input image, 
  - lut_3d.c: Lut root **folder** not defined
  - lut_3d.c: Select file outside Lut root **folder** is not allowed
  - lut_3d.c: CAUTION: 3D lut **folder** must be..., 
  - cli/main.c: error: can't open **folder**
  - camera_control.c: failed to get file information of %s in **folder** %s on device
- Film, film roll, filmroll
  - import.c: select a **folder** to import as **film roll**
  - import.c: recursively import **subdirectories**. each **directory** goes into a new **film roll**
  - import_session.c: Failed to initialize **film roll**
  - film.c: failed to insert **film roll**
  - film.c: cannot remove **film roll** having local copies with non accessible originals
  - film.c: Couldn't create **film** for **directory**
- Directory, Subdirectory
  - import.c: recursively import **subdirectories**. each **directory** goes into a new **film roll**
  - control_jobs.c: failed to create **film roll** for destination **directory**, aborting move.."
  - film.c: Couldn't create **film** for **directory**
  - generate_cache/main.c: creating cache **directory**
  - generate_cache/main.c: could not create **directory**
  - lua/film.c: Couldn't create **film** for **directory**
  - gallery.c: could not create **directory**
  - latex.c: could not create **directory**
  - disk.c: could not create **directory**
  - could not write to **directory**
  - l10n.c: error: can't open **directory**
  - film.c: failed to open **directory**
  - opencl.c: failed to create **directory**

What we can see is that
- darktable makes use of several variations at different places. 
- directory is most often used for error messages. 
- Only the import dialog deals with "filmroll" and "subdirectories" (checkbox and toopltip of the checkbox). This is fixed with this MR.
- film roll (sometimes film, sometimes filmroll) is the dt view on imported folders. Thats fine as long as it is consistent. For me it looks consistent after this MR.

[1]: https://en.wikipedia.org/wiki/Directory_(computing)#Folder_metaphor

codingdave:
I have no idea how to deal with po files. Thanks for updating me.

TurboGit:
It's simple : you don't have to deal with them. nothing to do.

codingdave:
No, I ment I dont have any more questions. I already have rebased my commits.

How do I mark the request as done?

codingdave:
Is there anything else I need to do? I have reacted to the review and fixed the requested changes. Still it looks like there are "Changes requested". How can I inform github that the requested changes have been done?

houz:
For reference, the difference between a folder and a film roll is this:

> The difference is quite simple: A "folder" is something on your system level,
inside dt we call it a "filmroll". That's it. In general it's a 1:1 mapping,
however, you can copy images to a folder after importing it or remove imported
images [in] dt while keeping them on the disk. So there may be cases where the
content of folder and filmroll differ, thus we call them differently.

(copied from https://darktable-users.narkive.com/4sUxvL9D/film-roll-vs-folder#post10)

A directory and a folder are the same thing.

houz:
I was asked to give my opinion about the actual pr:

i don't like the changes to the comments in the code (they are useless and introduce a discrepancy between comment and function name) so the one thing that would leave is renaming the dialog. which is also a bad idea, since then the dialog is named differently than the button that brought it up

codingdave:
I agree with the changes in code. I will remove this. 

I dont understand "the one thing that would leave is renaming the dialog. which is also a bad idea, since then the dialog is named differently than the button that brought it up". Its the other way around. Currently the names do not match, the button "import folder" opens a dialog with the title "import film". The PR is for fixing this. 

folder/directory is quite similar and not a windows thing. Folder is a usability term (desktop methaphor) and directory is the file system (technical concept) name. But this does not really matter if we agree on this or not because we seem to agree on the idea that the dialog and the button shall have the same name.

codingdave:
I undid the changes in the comments of the code.
@TurboGit seems that due to force-pushing I am not able to tell github that I have done what you were requesting. Please tell me if I can do anything more.

houz:
I'd prefer if we used "directory" instead of "folder". The latter sounds wrong to me. But that is personal preference.

codingdave:
If we are going for dictionary (fine for me as well as long as it is consistent) I would change all "folder" terms to "dictionary" terms.

TurboGit:
For me directory is really computer oriented and more "technical" than a folder. I would rather use folder everywhere.

codingdave:
@turbogit: do you mean in the UI only or also in the core e.g. when logging? So would you change "disk.c: could not create directory" to "disk.c: could not create folder"?

TurboGit:
Only in the UI when dealing with "folder" as defined in the collect module. For logging error related to disk drive path we should use directory as this is the correct name in this context.

On the UI the creator of dt have chosen names like film-roll and folder which are (to me) related to pre-digital photography. This should be kept as-is I would say.

codingdave:
I completely agree with that. Then I think the PR is also doing exactly that. Or is there anything that needs to be changed?

johnny-bit:
https://en.wikipedia.org/wiki/Directory_(computing)#Folder_metaphor

So - _folder_ should be used for user facing concepts for container of documents. _Directory_ should be used for computer-related stuff, such as file-system errors etc. _film roll_ is therefore specific subtype of folder, which represents folder containing only source images, like a pre-digital photographers used filmrolls.

This should be reflected in ui - users import _folders_ which create _film rolls_ as a darktable concept. Export should export to _folder_ and if there's problem with file system, darktable should log problem with _directory_

So... @TurboGit summary is on point.
To sum up my point:

- `directory`: filesystem and stuff. Application/Programmer facing.
- `folder`: UI representation of `directory`. User facing.
- `film roll`: darktable UI specific context use of `folder`. User facing and a concept within darktable.

TurboGit:
Looks ok to me. Thanks.

Issue #4065, opened 2020-01-08
Banding artifacts in Waveform after resizing


groutr:
**Describe the bug**
After #4041 (thanks @dtorop) was merged, I tested it and found some vertical banding artifacts after resizing the waveform.  I don't know if this caused by this PR, or merely uncovered by this PR.  I haven't noticed these pattern before.

**To Reproduce**
1. Open an image in darkroom
2. Resize the right panel to a wide width
3. Observe the vertical banding artifacts in the screenshot below.

**Expected behavior**
I would expect to see a waveform with no banding.  The pattern of the artifacts appear to depend on the width of the waveform with some widths giving a very noticeable pattern.

**Screenshots**
[![waveform-aliasing.png](https://i.postimg.cc/DznZJDPG/waveform-aliasing.png)](https://postimg.cc/K1pFC0Vc)

**Platform (please complete the following information):**
 - Darktable Version: 3.1.0.r321.ge3db8d52f
 - OS: archinux
 - OpenCL: yes (intel NEO)

dtorop:
I'm seeing this in older versions of darktable as well. I've gone back as far as release-3.0.0rc2 and still see it. Resizing the panel might make this more obvious.

Here is an example from release-3.0.0rc2 with a 350 width panel on a hidpi display:

![image](https://user-images.githubusercontent.com/2311860/72035492-0af36e00-3266-11ea-8c17-a8bd48e16ea1.png)

I'll still look and see if there's an obvious reason for this.

@groutr If you're able to replicate this in older versions, I'd be curious.

dtorop:
I think the problem is how the source image data gets processed into the waveform. In `_pixelpipe_final_histogram_waveform()`, the preview image gets downsampled horizontally to the width of the waveform histogram. The binning is simple:

https://github.com/darktable-org/darktable/blob/aae21bfd58d15c5231c0271693c3655df6133c22/src/develop/pixelpipe_hb.c#L1019

The preview image is 1338 pixels wide. With a 350 pixel wide panel, generally five preview pixels map to one waveform column, but every sixth waveform column has only three pixels mapped to it, and hence is 3/4 the brightness. The histogram gamma correction code may further exaggerate the banding.

A quick fix might be to weight each bin by the # of pixels in it. I'll check that out. A proper downsampling may be too slow.

dtorop:
Also, it may be that

[...]

PR #4075, opened 2020-01-10
histogram: waveform banding fix

dtorop:
This fixes banding identified by @groutr in #4065. The banding happens because columns were integrally binned, but not weighted by columns/bin. As only a few columns were in each bin, this produced large differences in brightness, and hence vertical striations in the waveform.

As an improvement which also fixes this, the waveform histogram is now created in a buffer with dimensions independent of the GUI, removing some pixelpipe/GUI interaction. The buffer dimensions are currently set so that the quality/speed is comparable to before this PR.

@groutr has tested this and reports that the waveform histogram is now cropped at the bottom. I haven't been able to replicate this.

### DIFF
--- a/src/develop/develop.c
+++ b/src/develop/develop.c
@@ -114,11 +114,17 @@ void dt_dev_init(dt_develop_t *dev, int32_t gui_attached)
     dev->histogram_pre_tonecurve_max = -1;
     dev->histogram_pre_levels_max = -1;
 
-    // _init_f() of mosaiced image may be twice the dimensions of DT_MIPMAP_F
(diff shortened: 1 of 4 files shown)


TurboGit:
There is conflicts now, can you fix by **rebasing** onto master ? Thansk.

dtorop:
> There is conflicts now, can you fix by **rebasing** onto master ?

@TurboGit It's rebased...

I still have a few thoughts about how to make the waveform histogram better/faster and more memory efficient, but this should be a discrete set of commits that fixes a long-standing banding bug, and also has clearer roles for the UI and the pixelpipe code.

TurboGit:
Works for me ! Thanks.

PR #4083, opened 2020-01-10
Some more code sharing

Mark-64:
Some more code reuse in finding duplicates for import and deletion


TurboGit:
Looks good to me ! Thanks.

PR #4088, opened 2020-01-11
Fix image information datetime field on Windows with Russian locale

avostrik:
Issue is that datetime field shows '-' instead of date and time.

On Windows with Russian locale strftime does not produce valid UTF-8
string that is expected by _metadata_update_value.

Most likely same issue is present on other locales that do not use Latin characters.

Signed-off-by: Andrey Vostrikov <av.linux.dev@gmail.com>

### DIFF
--- a/src/libs/metadata_view.c
+++ b/src/libs/metadata_view.c
@@ -472,7 +472,7 @@ static void _metadata_view_update_values(dt_lib_module_t *self)
       tt_exif.tm_isdst = -1;
       mktime(&tt_exif);
       // just %c is too long and includes a time zone that we don't know from exif
-      strftime(datetime, sizeof(datetime), "%a %x %X", &tt_exif);
+      strftime(datetime, sizeof(datetime), "%x %X", &tt_exif);
       _metadata_update_value(d->metadata[md_exif_datetime], datetime);
     }
     else


src/libs/metadata_view.c
```diff
@@ -472,7 +472,7 @@ static void _metadata_view_update_values(dt_lib_module_t *self)
       tt_exif.tm_isdst = -1;
       mktime(&tt_exif);
       // just %c is too long and includes a time zone that we don't know from exif
-      strftime(datetime, sizeof(datetime), "%a %x %X", &tt_exif);
+      strftime(datetime, sizeof(datetime), "%x %X", &tt_exif);
```
TurboGit: This just remove the abbreviated name from the date time string ! I think that if there is an issue it is elsewhere!

Maybe setting the local to be UTF-8 ?

avostrik:
Windows console does not respect UTF-8 and produces 1251 encoded ASCII char.
Locale is set via C call setlocale(LC_ALL,"") at the initialization, and then via language selection in settings, but still falls back to 1251. Can be easily verified by code:

```
   char *old = setlocale(LC_ALL, ""); 
   printf("%s\n", old);
```

```
mingw64 $ ./testme
Russian_Russia.1251
```

In most places strftime used to print only numbers, so it is fine but in this place and watermark with month name.

Short weekday name does not add any useful information into 'datetime' and is low risk change to make it work.

As a full featured fix, I suggest to eliminate all strftime in favor of GDateTime for example, which always produces UTF-8, but it has issues on Windows as well, as glib-2.0 locale file is not bundled with darktable package and requires more work to translate properly C-time structures.

Also glib formatting is not one-to-one with strftime, even though it has same format specifiers.

TurboGit:
Hopefully a better fix put in place #6263. So closing.

PR #4093, opened 2020-01-12
avoid depricated cmake policy

rabauke:
According to the cmake documentation 

> A policy is a deprecation mechanism and not a reliable feature toggle. A policy should almost never be set to OLD, except to silence warnings in an otherwise frozen or stable codebase, or temporarily as part of a larger migration path. The OLD behavior of each policy is undesirable and will be replaced with an error condition in a future release.

Thus, I  propose this pull request to utilize the new cmake policy without breaking the `CMakeList.txt` functionality. Version may be overridden via `-DPROJECT_VERSION=...` option when calling `cmake`.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2,7 +2,7 @@ cmake_minimum_required(VERSION 3.10.0)
 
 cmake_policy(SET CMP0011 NEW)
 cmake_policy(SET CMP0025 NEW)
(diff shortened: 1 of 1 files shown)


cryptomilk:
Looks good to me!

TurboGit:
Merging then. Thanks to both of you!

PR #4096, opened 2020-01-13
lut3d- add 3dl file support

phweyland:
- fix the call to gmic routine to get runtime version at each image init.
- support now 3dl files along with cube ones.

### DIFF
--- a/src/control/signal.c
+++ b/src/control/signal.c
@@ -104,6 +104,8 @@ static dt_signal_description _signal_description[DT_SIGNAL_COUNT] = {
     NULL, NULL, FALSE }, // DT_SIGNAL_HISTORY_CHANGE
   { "dt-develop-module-remove", NULL, NULL, G_TYPE_NONE, g_cclosure_marshal_generic, 1,
     pointer_arg, NULL, TRUE }, // DT_SIGNAL_MODULE_REMOVE
+  { "dt-develop-module-moved", NULL, NULL, G_TYPE_NONE, g_cclosure_marshal_VOID__VOID, 0,
+    NULL, NULL, FALSE }, // DT_SIGNAL_DEVELOP_MODULE_MOVED
   { "dt-develop-image-changed", NULL, NULL, G_TYPE_NONE, g_cclosure_marshal_VOID__VOID, 0,
     NULL, NULL, FALSE }, // DT_SIGNAL_DEVELOP_IMAGE_CHANGE
   { "dt-control-profile-changed", NULL, NULL, G_TYPE_NONE, g_cclosure_marshal_VOID__VOID, 0,
--- a/src/control/signal.h
+++ b/src/control/signal.h
@@ -139,6 +139,9 @@ typedef enum dt_signal_t
     */
   DT_SIGNAL_DEVELOP_MODULE_REMOVE,
 
+  /** \brief This signal is raised when order of modules in pipeline is changed */
+  DT_SIGNAL_DEVELOP_MODULE_MOVED,
+
   /** \brief This signal is raised when image is changed in darkroom */
   DT_SIGNAL_DEVELOP_IMAGE_CHANGED,
 
--- a/src/develop/imageop.c
+++ b/src/develop/imageop.c
@@ -575,8 +575,8 @@ static void dt_iop_gui_delete_callback(GtkButton *button, dt_iop_module_t *modul
     gtk_box_reorder_child(dt_ui_get_container(darktable.gui->ui, DT_UI_CONTAINER_PANEL_RIGHT_CENTER),
                           module->expander, -1);
 
-    gtk_widget_destroy(module->widget);
     dt_iop_gui_cleanup_module(module);
+    gtk_widget_destroy(module->widget);
   }
 
   // we remove all references in the history stack and dev->iop
--- a/src/develop/imageop.c
+++ b/src/develop/imageop.c
@@ -737,6 +737,8 @@ static void dt_iop_gui_movedown_callback(GtkButton *button, dt_iop_module_t *mod
   prev->dev->preview_pipe->cache_obsolete = 1;
   prev->dev->preview2_pipe->cache_obsolete = 1;
 
+  dt_control_signal_raise(darktable.signals, DT_SIGNAL_DEVELOP_MODULE_MOVED);
+
   // invalidate buffers and force redraw of darkroom
   dt_dev_invalidate_all(prev->dev);
 }
@@ -776,6 +778,8 @@ static void dt_iop_gui_moveup_callback(GtkButton *button, dt_iop_module_t *modul
   next->dev->pipe->cache_obsolete = 1;
   next->dev->preview_pipe->cache_obsolete = 1;
   next->dev->preview2_pipe->cache_obsolete = 1;
+
+  dt_control_signal_raise(darktable.signals, DT_SIGNAL_DEVELOP_MODULE_MOVED);
 
   // invalidate buffers and force redraw of darkroom
   dt_dev_invalidate_all(next->dev);
--- a/src/iop/lut3d.c
+++ b/src/iop/lut3d.c
@@ -1631,6 +1631,25 @@ void gui_reset(dt_iop_module_t *self)
 {
   memcpy(self->params, self->default_params, sizeof(dt_iop_lut3d_params_t));
 }
+/*
+static void _show_hide_colorspace(dt_iop_module_t *self)
+{
+  dt_iop_lut3d_gui_data_t *g = (dt_iop_lut3d_gui_data_t *)self->gui_data;
+  const int imgid = darktable.develop->image_storage.id;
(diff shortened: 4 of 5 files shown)


phweyland:
@TurboGit the discussion about 3dl lut to be applied on raw image remembers me that lut3d can be moved outside colorin-colorout.
But in that case the combobox to chose the colorspace should be masked or at least disabled.
Is there a simple way for a module at init time to know if it is or not outside colorin-colorout ? (I had a look at iop_order.c but I' haven't found...).
Thanks

TurboGit:
@phweyland : As I'm working actively on the full re-implementation of the iop-order support we can work on that when it is merged. With the new implementation it will be possible and quite simple to check this specific case.

So let's this PR wait for #3908.

chrik5:
Can you please add a version check for the build process also? 
I doesn't make sense to build dt with a GMIC version which is not supported?

TurboGit:
> Can you please add a version check for the build process also?
> I doesn't make sense to build dt with a GMIC version which is not supported?

There seem to be no way to do that. Any advice ?

parafin:
If you can check it at runtime, then it's possible at configure time also - just compile a test program doing a runtime check.

phweyland:
> If you can check it at runtime, then it's possible at configure time also - just compile a test program doing a runtime check.

Trying to understand the idea. 
Before building dt, build a small program with libgmic linked making in some way its version available. Then, at dt build time, use the the result of this program. Correct ?

chrik5:
> then it's possible at configure time also
Yes, that seems to be what I had in mind. A GMIC version check on configure time. Is this somewhere here to add: 
https://github.com/darktable-org/darktable/blob/master/src/CMakeLists.txt#L432

phweyland:
> > then it's possible at configure time also
> 
> Yes, that seems to be what I had in mind. A GMIC version check on configure time. Is this somewhere here to add:
> https://github.com/darktable-org/darktable/blob/master/src/CMakeLists.txt#L432

Yes, should be. But libgmic doesn't return its version (at least I haven't succeeded in making it work).

parafin:
The following cmake code works:
```
set(CMAKE_REQUIRED_INCLUDES ${GMIC_INCLUDE_DIRS})
check_cxx_source_compiles("
#include <gmic.h>
#if gmic_version < 281
#error OLD_VERSION
#endif
int main() { return 0; }
" GMIC_VERSION_OK)
if(NOT GMIC_VERSION_OK)
  message(FATAL_ERROR "Too old GMIC version")
endif()
```
I don't suggest to use it as-is, just it's just to demonstrate the idea.

phweyland:
> I don't suggest to use it as-is, just it's just to demonstrate the idea.

Great, thank you. I'll make a try asap.

phweyland:
Yes, that works ! I'll remove the code corresponding to runtime version detection and issue a PR for that. Thanks a lot @parafin.

-- Performing Test GMIC_VERSION_OK
-- Performing Test GMIC_VERSION_OK - Failed
-- Found GMIC but version < 270

If libgmic is turned able to return the version to cmake, we'll revert this workaround.
see https://discuss.pixls.us/t/gmic-lookup-symbol-err-on-opensuse-darktable-master-build/15827/24

TurboGit:
@phweyland : Now that iop-order PR is merged you can use dt_ioppr_is_iop_before() in iop_order.c to check for specific iop ordering. Let me know if this fill your need.

phweyland:
> ow that iop-order PR is merged you can use dt_ioppr_is_iop_before() in iop_order.c to check for specific iop ordering. Let me know if this fill your need.

I was off. I'll do it asap. Thanks

phweyland:
@TurboGit 

I've given some try. To make it work properly one must be informed when a module is moved in the pixelpipe. I've not found such an event. Any suggestion ?

Also I cannot use dt_ioppr_is_iop_before() to check if the module is after colorin.
If can use the same routine when I have only one instance it doesn't work for multiple instance.
I get this message which is clear enough (after_colorout code is not correct):
`cannot get iop-order for colorout instance 1`

With the same code I get a segmentation fault when I restart dt but that may be normal.
[darktable_bt_USC1F0.txt](https://github.com/darktable-org/darktable/files/4177665/darktable_bt_USC1F0.txt)

The piece of code I've added is this one:
```
  const int imgid = darktable.develop->image_storage.id;
  GList *iop_order_list = dt_ioppr_get_iop_order_list(imgid, FALSE);
  const gboolean before_colorin = dt_ioppr_is_iop_before(iop_order_list, "colorin", self->op, self->multi_priority);
  const gboolean after_colorout = dt_ioppr_is_iop_before(iop_order_list, self->op, "colorout", self->multi_priority);
  if(before_colorin || after_colorout)
  {
    gtk_widget_hide(g->colorspace);
  }
  else
  {
    gtk_widget_show(g->colorspace);
  }
```

phweyland:
Using dt_ioppr_get_iop_order() (instead of dt_ioppr_is_iop_before()) and comparing it between modules gives the right result. It seems to work too with multi instances of the module.

So remain to find a way to be triggered when the module is moved.

TurboGit:
I think the signal DT_SIGNAL_DEVELOP_HISTORY_CHANGE is raised when module moves.

phweyland:
> I think the signal DT_SIGNAL_DEVELOP_HISTORY_CHANGE is raised when module moves.

That works but this may trig more often than required. Should I add a signal, like DT_SIGNAL_MODULE_MOVED ?

Sometimes I get a huge iop_order value (2147483647 here below).
Here are my traces (when I create the new instance, first occurrence of instance 1):

```
imgid 56915 module lut3d mprio 0 mod 21 in 29 out 69
moved
moved
cannot get iop-order for lut3d instance 1
imgid 56915 module lut3d mprio 1 mod 2147483647 in 29 out 69
moved
cannot get iop-order for lut3d instance 1
imgid 56915 module lut3d mprio 1 mod 2147483647 in 29 out 69
imgid 56915 module lut3d mprio 0 mod 21 in 29 out 69
```

When I compress the history stack the iop_value is back ok.

Maybe linked to that, the stack compression doesn't work properly:

image

I should have only 2 lut 3D lines.

My current code is this one:

```
  const int imgid = darktable.develop->image_storage.id;
  GList *iop_order_list = dt_ioppr_get_iop_order_list(imgid, FALSE);
  const int order_lut3d = dt_ioppr_get_iop_order(iop_order_list, self->op, self->multi_priority);
  const int order_colorin = dt_ioppr_get_iop_order(iop_order_list, "colorin", -1);
  const int order_colorout = dt_ioppr_get_iop_order(iop_order_list, "colorout", -1);
  printf("imgid %d module %s mprio %d mod %d in %d out %d\n", imgid, self->op, self->multi_priority, order_lut3d, order_colorin, order_colorout);
  if(order_lut3d < order_colorin || order_lut3d > order_colorout)
  {
    gtk_widget_hide(g->colorspace);
  }
  else
  {
    gtk_widget_show(g->colorspace);
  }
```

and for the trigger:

```
void module_moved_callback(gpointer instance, dt_iop_module_t *self)
{
  printf("moved\n");
  dt_control_signal_block_by_func(darktable.signals, G_CALLBACK(module_moved_callback), self);
  gui_update(self);
  dt_control_signal_unblock_by_func(darktable.signals, G_CALLBACK(module_moved_callback), self);
}
```

Let me know if you need more information.

TurboGit:
> Should I add a signal, like DT_SIGNAL_MODULE_MOVED ?

I suppose so. Will be handy and could be reused in ioporder lib.

TurboGit:
> Also I cannot use dt_ioppr_is_iop_before() to check if the module is after colorin.

Maybe a fix to be put in then. It should be ok with multiple instances I suppose.

phweyland:
Set proper sequence to use dt_ioppr_get_iop_order(). My error above was to use dt_ioppr_get_iop_order_list() which returns a static list (not updated by the moves).

Maybe dt_ioppr_is_iop_before() can be removed now.

Added a specific "module moved" signal.

Fixed the delete module instance sequence (widget was deleted before calling gui_cleanup()) in imageop.

Should be OK for review. Thanks.

TurboGit:
@phweyland : can you point me to a 3dl file to test? I just cannot find one :(

phweyland:
[C-8120-STD.3dl.txt](https://github.com/darktable-org/darktable/files/4189095/C-8120-STD.3dl.txt)
[C-9730-STD.3dl.txt](https://github.com/darktable-org/darktable/files/4189096/C-9730-STD.3dl.txt)
[Choweed - Standard.3dl.txt](https://github.com/darktable-org/darktable/files/4189097/Choweed.-.Standard.3dl.txt)
[Adoras - Standard.3dl.txt](https://github.com/darktable-org/darktable/files/4189098/Adoras.-.Standard.3dl.txt)

phweyland:
here https://blog.sowerby.me/fuji-film-simulation-profiles/

TurboGit:
Looks ok ! Merged. Thanks.

PR #4097, opened 2020-01-13
Convert lensfun implementation to pure C code

cryptomilk:
This converts our lensfun implementation to pure C code.


parafin:
Why? It was changed to C++ for a reason.

parafin:
Specifically this breaks compilation with lensfun 0.3.95 (see LF_0395 macros).

cryptomilk:
What's the reason for C++?

parafin:
C API is broken in lensfun 0.3.95. They probably fixed it in git version, but C++ API is native for lensfun, so it's better tested and supported.

parafin:
That decision was already made when that PR was merged, you are basically reverting it, so what's the reason for the revert?

jakubfi:
@cryptomilk lensfun API has been switched from C to C++ due to the fact that C++ is the native API of the library, while C API was found to be sometimes not up to date and buggy. Using the native API is always a good idea, especially that in this case it doesn't harm the code in any way.

Support for (pre-)alpha version of lensfun (0.3.95) is experimental and was added to make the transition to lensun-next easier and allow for early compatibility testing. As 0.3.95 is alpha, it is expected to never be a generally available version in any distro, thus no regular user will use this combination. But if user wants to test it, dt can be compiled against it and should work.

If there are any concerns whether dt should even compile against 0.3.95, support for it can be disabled (#undef LF_0395 or something similar), but I would be against completely removing the 0.3.95 related code.

TurboGit:
I agree, this decision was made to avoid the buggy C API. We certainly don't want to go back at this stage. @cryptomilk : I let you close?

PR #4101, opened 2020-01-13
dt_lua_image_t.is_raw manual update

### LINKED ISSUE #4086: Update Lua API doc for Image flags
**Describe the bug**
When importing a new folder, not all image flags (visible in image information) are set correctly.
In particular, the flag "RAW", is not set on import, but only when opening image in darkroom.
This causes the property image.is_raw in lua to malfunction (see issue 209 in lua scripts) and maybe other effects.
Other flags might be incorrect too.

**To Reproduce**
Import a new folder, never imported before, with a few images.
do no enter in darkroom.
see the RAW flag (fifth from the left) not set in the image information module in the left panel
now open one image in darkroom. The RAW flags is set
go back in lighttable, the flag RAW is now correct for the image opened in darkroom

**Expected behavior**
The image flags should be set correctly

**Screenshots**
Cattura20

Tested with DT 3.1 master in Windows 10 and Ubuntu 19.10

**Additional context**
As I finally managed to setup Darktable in Eclipse under Windows, I could run the debugger, and it seems to me that the only place where the RAW flag is set is in dt_imageio_open_rawspeed() in imageio_rawspeed.cc, but that function is not called when importing images in lighttable. 
It is only called when opening in darkroom and there the flags are set correctly.

wpferguson:
Added warning about dt_lua_image_t.is_raw not being set on import

fixes #4086

### DIFF
--- a/doc/usermanual/lua/lua_api.xml
+++ b/doc/usermanual/lua/lua_api.xml
@@ -4567,6 +4567,8 @@ This function is recursion-safe and can be used to dump _G if needed.</para>
 </indexterm>
 <synopsis>boolean</synopsis>
 <para>True if the image is a RAW file.</para>
+<para>WARNING: This is a flag that gets set the first time the image is opened, either by an edit or thumbnail generation.  If images are imported using
+darktable.database.import(), then dt_lua_image_t.is_raw is not guaranteed to be correct.</para>
 
 </section>
 


aurelienpierre:
Great ! Thanks !

TurboGit:
@aurelienpierre : for each merge we want to asses if the changes must also go into 3.0.1 or not. When (or before merging) please update the Milestone to be either 3.2 (no merging in 3.0.1, will only be part of next major release) or 3.0.1 otherwise.

TurboGit:
Also if one need to merge a PR with multiple commits we need to rebase. We must never ever merge master in darktable-3.0.x (or the opposite). The PR must probably be rebased. For a PR with a single commit you can cherry pick the needed commit into darktable-3.0.x. If something is not clear let's discuss that.

TurboGit:
First part then : Here obviously there is no need for a merge in 3.0.1 as the documentation is only worked from master branch.

PR #4102, opened 2020-01-13
check runtime errors in guided filter OpenCL implementation

rabauke:
The guided filter utilizes quite a lot memory. Therefore, the OpenCL implementation in `guided_filter_cl` estimates the amount of memory that will be used and falls back to a CPU code path if the GPU lacks a sufficient amount of memory. Under some circumstances, however, it can happen that the calculation of the guided filter fails although the estimate indicates that the memory requirement would be met. (The OpenCL run time may need some a priory unknown amount of GPU memory which is not taken into account.)

This patch changes the safety factor from 9/8 to a more conservative value of 5/4 and falls back to a GPU code path if GPU calculation fails at run time.

### DIFF
--- a/src/common/guided_filter.c
+++ b/src/common/guided_filter.c
@@ -556,16 +556,15 @@ static int cl_generate_result(const int devid, const int width, const int height
 }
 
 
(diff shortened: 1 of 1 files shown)


TurboGit:
Looks good to me. Thanks!

PR #4104, opened 2020-01-13
fix #4092 (increase haze on overexposed images)

### LINKED ISSUE #4092: Highlights reconstruction breaks haze removal
**Describe the bug**
Haze removal appears to increase haze on certain images.  The behavior differs depending on when hazeremoval is performed relative to other operations.  The pixel processing order has not been reordered.

**To Reproduce**
Unzip the attached file and load it with the associated processing XMP file.
Enable Haze Removal with default parameters.  

**Expected behavior**
Apparent haze should be reduced.

**Screenshots**

 - Darktable Version: 3.0.0 (also latest 3.1.0 trunk)
 - OS: Windows 10 (also Ubuntu 18.04)
 - Same behavior with and without OpenCL
 - Nvidia K2200 431.94 (Windows), 435.21 (Ubuntu)

**Additional context**
The behavior differs depending on what previous operations have been done to the image.  The attached XMP sidecar contains the compressed history stack.  If I export the image as a TIFF or JPG and edit the exported file, haze removal works as expected.  I have no previous experience with Darktable source, however I poked around and added logging in src/iop/hazeremoval.c.  I noticed the comment about computing distance_max from the PREVIEW pixelpipe vs FULL pixelpipe.  In the case where haze removal is not working as expected, distance_max is -0.187453.  When editing the exported TIFF, distance_max is  0.381375
[DSCN1262.NRW.zip](https://github.com/darktable-org/darktable/files/4050818/DSCN1262.NRW.zip)

rabauke:
The haze removal module still requires properly exposed images, even with this patch. The proposed patch, however, ensures that haze is not enhanced in highly overexposed images.

### DIFF
--- a/src/iop/hazeremoval.c
+++ b/src/iop/hazeremoval.c
@@ -757,7 +757,7 @@ void process(struct dt_iop_module_t *self, dt_dev_pixelpipe_iop_t *piece, const 
 
   // finally, calculate the haze-free image
   const float t_min
-      = fmaxf(expf(-distance * distance_max), 1.f / 1024); // minimum allowed value for transition map
+      = fminf(fmaxf(expf(-distance * distance_max), 1.f / 1024), 1.f); // minimum allowed value for transition map
   const float *const c_A0 = A0;
   const gray_image c_trans_map_filtered = trans_map_filtered;
 #ifdef _OPENMP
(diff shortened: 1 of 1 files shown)


TurboGit:
@rabauke : I'm not sure to understand.
The PR title is : "increase haze on overexposed images"
And the comment in the PR says : "ensures that haze is not enhanced in highly overexposed images"

So, I'm not really clear about the goal here. Can you clarify? Thanks.

rabauke:
In bug #4092 it was reported that the haze removal module may _increase_ haziness  rather than to reduce haze in some images. This only happens when the haze removal module gets a strongly overexposed input image, i.e. a large portion of rgb values is larger than 1. This means the haze removal module behaves erroneously when input data is unsuitable for the haze removal module. Users can easily avoid this issue by reducing the exposure, either already when taking the image or in post processing by placing the exposure module before haze removal.

The small change in the haze removal module that I propose here ensures that haze is never enhanced in such over exposed images. The revised module will become a no-op for cases where the current version increases haze. This change in the haze removal algorithm does not affect edits where the current version works.

TurboGit:
Thanks for the explanations!

PR #4108, opened 2020-01-14
Always set the white & black point when creating profiles.

### LINKED ISSUE #4106: Internal D65 colour profiles possibly wrong
**To reproduce:**
1. export a a raw picture to TIFF 32 bits float with Adobe RGB output profile
2. open the exported TIFF (input profile set to "embedded ICC matrix", no module except input/output profiles)
3. export it again with the same parameters.

Then, either overlap both exports using the darkroom snapshots or compute the RMSE using GraphicMagick `gm compare -metric RMSE export-1.tif export-2.tif -file diff.png`

**Issue:**

The blue channel has twice the RMSE as the other channels. On the picture, especially in dark greens, the second export (export from export) displays darker colours.

With LCMS2 on (CPU SSE2 path):
```
Image Difference (RootMeanSquaredError):
           Normalized    Absolute
          ============  ==========
     Red: 0.0000082394        0.5
   Green: 0.0000052063        0.3
    Blue: 0.0000157842        1.0
   Total: 0.0000107103        0.7
```

With LCMS2 off (OpenCL path):
```
Image Difference (RootMeanSquaredError):
           Normalized    Absolute
          ============  ==========
     Red: 0.0000067537        0.4
   Green: 0.0000059158        0.4
    Blue: 0.0000203831        1.3
   Total: 0.0000128592        0.8
```

sRGB exhibits the same issue:
```
Image Difference (RootMeanSquaredError):
           Normalized    Absolute
          ============  ==========
     Red: 0.0000119955        0.8
   Green: 0.0000108273        0.7
    Blue: 0.0000247451        1.6
   Total: 0.0000170630        1.1
```

Comparing with an external ProPhoto ICC profile:
```
Image Difference (RootMeanSquaredError):
           Normalized    Absolute
          ============  ==========
     Red: 0.0000011652        0.1
   Green: 0.0000012465        0.1
    Blue: 0.0000031726        0.2
   Total: 0.0000020798        0.1
```
Diff image (differences highlighted in red):
diff

I suspect an error in the internal Adobe RGB and sRGB profiles on the B channel coeff.

Based on @blitzgneisserin https://discuss.pixls.us/t/jpegs-in-adobergb-too-little-saturated/15785 (but not sure it's the same problem).

TurboGit:
Possibly fix #4106.

### DIFF
--- a/src/common/colorspaces.c
+++ b/src/common/colorspaces.c
@@ -308,6 +308,7 @@ static cmsHPROFILE _create_lcms_profile(const char *desc, const char *dmdd,
                                         const cmsCIExyY *whitepoint, const cmsCIExyYTRIPLE *primaries, cmsToneCurve *trc,
                                         gboolean v2)
 {
+  const cmsCIEXYZ black = { 0, 0, 0 };
   cmsMLU *mlu1 = cmsMLUalloc(NULL, 1);
   cmsMLU *mlu2 = cmsMLUalloc(NULL, 1);
   cmsMLU *mlu3 = cmsMLUalloc(NULL, 1);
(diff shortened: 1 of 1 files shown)


blitzgneisserin:
Should I (try to) test?

TurboGit:
Yes sure go ahead !

blitzgneisserin:
I opened an imported JPEG in AdobeRGB and the colors are still wrong, i.e. same as before (less saturation and little different hue). I will also try what Aurelien did though that seems to be more complicated.

src/common/colorspaces.c
```diff
@@ -316,15 +317,11 @@ static cmsHPROFILE _create_lcms_profile(const char *desc, const char *dmdd,
   cmsToneCurve *out_curves[3] = { trc, trc, trc };
   cmsHPROFILE profile = cmsCreateRGBProfile(whitepoint, primaries, out_curves);
 
+  if(v2) cmsSetProfileVersion(profile, 2.1);
```
aurelienpierre: Oh I missed that. Internal ICC should always be v2, because v4 removes the white point tag at all, so the RGB primaries need to be manually adapted to D50.

TurboGit: I suppose I should close this PR, right? It is wrong and the issue is not there?

aurelienpierre: Not sure at this point. Inputing the white point is definitely a good thing anyway, it can't hurt. Since external ICC profiles work, the problem is most probably in the internal profiles we create.

TurboGit: Ok, I'll close this for now.

aurelienpierre:
One mistake here: the white point is given as xyY, while the tag seems to expect XYZ (investigation WIP)

TurboGit:
Ok, I'll reopen then to not forget to double check this xyY vs XYZ.

Are you looking at this? Or should I ?

aurelienpierre:
I'm on it now. Converting to XYZ doesn't solve the problem. Stay tuned…

TurboGit:
Fixed by #4106.

Issue #4034, opened 2020-01-05
Tags DT 3.1 not committed to database ?


mimijojo:
With Darktable release 3.1.0~git299.40c7be068, editing tags seem to work, but edits seems not to be committed to the database before closing DT.

Easy to reproduce : When closing DT and reopening BT, all the tag edits seem to be lost.

phweyland:
I cannot reproduce (3.1.0+303~ga2a202f86).
Could you please make some snapshots to describe what you do and what you get ?

johnny-bit:
Fixed by  #4112 IMO - the PR was not properly connected to the issue

PR #4112, opened 2020-01-14
fix changed tags not saved in db

phweyland:
table `used_tags` was not used anymore but at start time dt was still synchronizing data.tags with used_tags, removing some changes made on tags.

table `used_tags` is now removed.

fix issue #4034


phweyland:
If that helps to reproduce the issue: it was visible when one changes the name / path of an existing tag.
After restarting dt the old tag is recreated.
The other changes (new tag, tag property) were kept.

TurboGit:
Works for me!

PR #4118, opened 2020-01-15
Add Y0U0V0 mode in denoiseprofiled's wavelets

rawfiner:
This aims at improving wavelets' result using a different color space, but one that is specially adapted to denoising.
I chose Y0U0V0 as it was created especially for denoising. See for instance in this book http://dev.ipol.im/~ibal/SecretsOfImageDenoisingCuisine.pdf page 40.
I am doing a small change to the conversion, as Y0 was defined R/3+G/3+B/3 in order to reduce the noise variance in this channel, supposing R, G, and B have equal variance. I am using a weighted average instead, that does basically the same thing, but with white balance removed.
I removed the wavelets chroma and luma presets, as this feature can be used instead.

### DIFF
--- a/data/kernels/denoiseprofile.cl
+++ b/data/kernels/denoiseprofile.cl
@@ -84,6 +84,37 @@ denoiseprofile_precondition_v2(read_only image2d_t in, write_only image2d_t out,
   write_imagef (out, (int2)(x, y), t);
 }
 
+kernel void
+denoiseprofile_precondition_Y0U0V0(read_only image2d_t in, write_only image2d_t out, const int width, const int height,
+                             const float4 a, const float4 p, const float4 b, global float *toY0U0V0)
+{
+  const int x = get_global_id(0);
+  const int y = get_global_id(1);
+
+  if(x >= width || y >= height) return;
+
+  float4 pixel = read_imagef(in, sampleri, (int2)(x, y));
+  const float alpha = pixel.w;
+
+
+  float4 t = 2.0f * native_powr(fmax((float4)0.0f, pixel + b), 1.0f - p / 2.0f) / ((-p + 2.0f) * sqrt(a));
+
+  float4 outpx = (float4)0.0f;
+  outpx.x += toY0U0V0[0] * t.x;
+  outpx.x += toY0U0V0[1] * t.y;
+  outpx.x += toY0U0V0[2] * t.z;
+  outpx.y += toY0U0V0[3] * t.x;
+  outpx.y += toY0U0V0[4] * t.y;
+  outpx.y += toY0U0V0[5] * t.z;
+  outpx.z += toY0U0V0[6] * t.x;
+  outpx.z += toY0U0V0[7] * t.y;
+  outpx.z += toY0U0V0[8] * t.z;
+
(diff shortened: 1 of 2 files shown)


rawfiner:
opencl code is working. Basically, after some code cleaning it will be ready for review.

About difficulty, some basic explanation of the changes:
The change here is basically just a matrix product ;-)
The matrix is derived from the conversion matrix to Y0U0V0.
The first line of my matrix (for Y0) is basically an extension of the one of the original matrix to noise that depends on white balance.
Apart of that, all the lines are normalized in a way that should produce an identical variance for Y0, U0, and for V0.

About the milestone, I think after this PR I will focus on a project of adding a blind denoising module that would not require profiling, so if everything goes well I hope to have this done for 3.2 (by "focus", I include only my work on features, I will of course continue to do bug fixes for denoise profiled if needed). Having that in mind, I think the change to denoise profiled made in this PR could come before the 3.2, in a minor release: it is in the same spirit of changes made for 3.0 of making the denoise profiled easier to use. Basically here, the user can denoise an image with wavelets in one instance, without using inappropriate blending modes, and using only the strength slider the user can control quite well the denoising. I think it would maybe come a bit "too late" if it is only in 3.2, as the new module may give hopefully better performance.
What do you think?

rawfiner:
The PR is now ready for testing and review :-)

TurboGit:
Just curious, I cannot test right now, do you have screenshots showing the improvements of the denoising?

TurboGit:
Sure, I put 3.2 as a conservative choice but we can reconsider. I have tested it and it works ok on my side, I must say that I did not found lot of differences just by looking at the results. I'll review the code in the coming days.

rawfiner:
The difference in result is mainly visible on images where color noise is visible. Also note that you may have to give higher values to the strength slider in Y0U0V0 mode than in RGB mode.

Some comparisons (with some of my noisy images from https://drive.google.com/open?id=11LxZWpZbS66m7vFdcoIHNTiG20JnwlJT ):
RGB mode:
P2170287_02

Y0U0V0 mode trying to get a close look (notice that the color noise is better smoothed out):
P2170287_03

On another image:
In RGB mode, with increasing strength to get rid of the color noise:
(1):
DSCF4037
(2):
DSCF4037_01
(3):
DSCF4037_02
When we manage to remove color noise with attempt (3), we removed too much the luminance details.

In Y0U0V0 mode, we can get a look similar to (1) without the color noise:
DSCF4037_03

We can also smooth only the color, or smooth less fine details on Y0 without getting back some color noise playing with the curves.

All in all, this PR basically can give results similar to what one can get using 2 instances like dt 2.6 presets, the advantages are just the fact that we have one instance only and we don't need to use blending modes. (also, the color space should, in theory, behave better as it has been designed for denoising).

TurboGit:
Impressive, the wavelet in this mode is better (less smoothing) than the non-local mean. That's a great job especially since the wavelet mode is faster! Congrats again, great job!

I'll merge in master and 3.0.1.

rawfiner:
Thank you for the testing and code review @TurboGit :-)

PR #4122, opened 2020-01-15
Fix hang in timeline due to infinite loop

mbaumgae:
This PR fixes a darktable hang using the timeline with zoom level jour:

Screenshot (2)

If you move your mouse hovering over the timeline from right to left at the hour zoom level (see picture above), there is a chance that darktable hangs due to an infinite loop.


AlicVB:
Good catch ! An obvious copy-paste error !

TurboGit:
Thanks ! Merging in master and 3.0.1.

PR #4124, opened 2020-01-15
lut3d - removed gmic version from params

phweyland:
as discussed with @parafin on #4121.


TurboGit:
Looks ok to me, given last discussion let's see if this is ok with @parafin to.

TurboGit:
Thanks to both of you ! Merging then.

PR #4125, opened 2020-01-16
Updating POTFILES.in

EdgarLux:
Removing the lines related to the deletion of the code from facebook, flickr, and google photo and to avoid the "xgettext:error..." when updating the po file for translations.


TurboGit:
Thanks!

PR #4132, opened 2020-01-17
Waveform histogram speed-up and resolution increase

dtorop:
Changes to calculation of waveform histogram:

- speed up waveform histogram calculation via OpenMP and caching the gamma calculation
- calculate a more detailed histogram (especially on hidpi displays) in equivalent time, thanks to this speedup

I tested this code on both a recent and an eight year old laptop. In both cases, the new code is either faster or just as fast as the prior code, while producing less blurry output. Note that if detailed waveform histograms are not considered a good in themselves, it would be easy to turn down the resolution in dt_dev_init() to make the histogram math happen faster.

This PR also contained changes to how the waveform histogram appears in the UI. I've moved these to #4140 and rebased.

Here is before/after for lodpi:

Before:
waveform lodpi before

After (includes new grid from #4140):
waveform lodpi after

And before/after for hidpi:

Before:
waveform hidpi before

After:
image

This is a follow-up to #4041 and #4075.

### DIFF
--- a/src/libs/histogram.c
+++ b/src/libs/histogram.c
@@ -425,7 +425,7 @@ static gboolean _lib_histogram_button_press_callback(GtkWidget *widget, GdkEvent
 
   if(!hooks_available) return TRUE;
 
-  if(event->type == GDK_2BUTTON_PRESS)
+  if(event->type == GDK_2BUTTON_PRESS && (d->highlight == 1 || d->highlight ==2))
   {
     dt_dev_exposure_reset_defaults(darktable.develop);
   }
--- a/src/develop/develop.c
+++ b/src/develop/develop.c
@@ -115,15 +115,14 @@ void dt_dev_init(dt_develop_t *dev, int32_t gui_attached)
     dev->histogram_pre_levels_max = -1;
 
     // FIXME: allow setting these via dt_conf system?
-    // useless once it exceeds width of DT_MIPMAP_F, (which is
-    // darktable.mipmap_cache->max_width[DT_MIPMAP_F]*2 for mosaiced
-    // imaages)
-    //dev->histogram_waveform_width = 768;
-    dev->histogram_waveform_width = darktable.mipmap_cache->max_width[DT_MIPMAP_F] * 2;
-    // this is sufficient visual information though the waveform
-    // widget will probably be either 175 or 350 pixels high depending
-    // on hidpi
-    dev->histogram_waveform_height = 512;
+    // Calculate more pixels for hidpi. Don't exceed width of
+    // DT_MIPMAP_F (darktable.mipmap_cache->max_width[DT_MIPMAP_F]*2
+    // for mosaiced images) or make it too slow to calculate
+    // (regardless of ppd)
+    dev->histogram_waveform_width = darktable.mipmap_cache->max_width[DT_MIPMAP_F] * MAX(2, darktable.gui->ppd);
+    // Hardcoded hack: histogram widget will probably be either 175 or
+    // 350 pixels high depending on hidpi
+    dev->histogram_waveform_height = 175 * MAX(2, darktable.gui->ppd);
     dev->histogram_waveform_stride = 4 * dev->histogram_waveform_width;
     dev->histogram_waveform = (uint8_t *)calloc(dev->histogram_waveform_height * dev->histogram_waveform_stride, sizeof(uint8_t));
   }
--- a/src/develop/develop.c
+++ b/src/develop/develop.c
@@ -118,11 +118,12 @@ void dt_dev_init(dt_develop_t *dev, int32_t gui_attached)
     // useless once it exceeds width of DT_MIPMAP_F, (which is
     // darktable.mipmap_cache->max_width[DT_MIPMAP_F]*2 for mosaiced
     // imaages)
-    dev->histogram_waveform_width = 512;
+    //dev->histogram_waveform_width = 768;
+    dev->histogram_waveform_width = darktable.mipmap_cache->max_width[DT_MIPMAP_F] * 2;
     // this is sufficient visual information though the waveform
     // widget will probably be either 175 or 350 pixels high depending
     // on hidpi
-    dev->histogram_waveform_height = 192;
+    dev->histogram_waveform_height = 512;
     dev->histogram_waveform_stride = 4 * dev->histogram_waveform_width;
     dev->histogram_waveform = (uint8_t *)calloc(dev->histogram_waveform_height * dev->histogram_waveform_stride, sizeof(uint8_t));
   }
--- a/src/develop/develop.c
+++ b/src/develop/develop.c
@@ -115,14 +115,15 @@ void dt_dev_init(dt_develop_t *dev, int32_t gui_attached)
     dev->histogram_pre_levels_max = -1;
 
     // FIXME: allow setting these via dt_conf system?
-    // Calculate more pixels for hidpi. Don't exceed width of
-    // DT_MIPMAP_F (darktable.mipmap_cache->max_width[DT_MIPMAP_F]*2
-    // for mosaiced images) or make it too slow to calculate
-    // (regardless of ppd)
-    dev->histogram_waveform_width = darktable.mipmap_cache->max_width[DT_MIPMAP_F] * MAX(2, darktable.gui->ppd);
+    // Don't use absurd amounts of memory, exceed width of DT_MIPMAP_F
+    // (which will be darktable.mipmap_cache->max_width[DT_MIPMAP_F]*2
+    // for mosaiced images), nor make it too slow to calculate
+    // (regardless of ppd). Try to get enough detail for a (default)
+    // 350px panel, possibly 2x that on hidpi.
+    dev->histogram_waveform_width = darktable.mipmap_cache->max_width[DT_MIPMAP_F]/2;
     // Hardcoded hack: histogram widget will probably be either 175 or
     // 350 pixels high depending on hidpi
-    dev->histogram_waveform_height = 175 * MAX(2, darktable.gui->ppd);
+    dev->histogram_waveform_height = 175;
     dev->histogram_waveform_stride = 4 * dev->histogram_waveform_width;
     dev->histogram_waveform = (uint8_t *)calloc(dev->histogram_waveform_height * dev->histogram_waveform_stride, sizeof(uint8_t));
   }
--- a/src/develop/develop.c
+++ b/src/develop/develop.c
@@ -114,13 +114,15 @@ void dt_dev_init(dt_develop_t *dev, int32_t gui_attached)
     dev->histogram_pre_tonecurve_max = -1;
     dev->histogram_pre_levels_max = -1;
 
-    // mosaiced image may be twice the dimensions of DT_MIPMAP_F but
-    // half width still gives sufficient data and is reasonably fast
-    dev->histogram_waveform_width = darktable.mipmap_cache->max_width[DT_MIPMAP_F] / 2;
+    // FIXME: allow setting these via dt_conf system?
+    // useless once it exceeds width of DT_MIPMAP_F, (which is
+    // darktable.mipmap_cache->max_width[DT_MIPMAP_F]*2 for mosaiced
+    // imaages)
+    dev->histogram_waveform_width = 512;
     // this is sufficient visual information though the waveform
     // widget will probably be either 175 or 350 pixels high depending
     // on hidpi
-    dev->histogram_waveform_height = 128;
+    dev->histogram_waveform_height = 192;
     dev->histogram_waveform_stride = 4 * dev->histogram_waveform_width;
     dev->histogram_waveform = (uint8_t *)calloc(dev->histogram_waveform_height * dev->histogram_waveform_stride, sizeof(uint8_t));
   }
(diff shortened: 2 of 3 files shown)


dtorop:
Questions I have regarding this PR:

- This PR could be split into at least two PRs: the performance improvements and resulting resolution increase (changes to `src/develop/{pixelpipe_hb,develop}.c`) would make for one, and the UI changes (found in `src/{libs/histogram.c,gui/draw.h,iop/levels.c}`) another. Would this aid review of this code?
- Does a higher resolution waveform histogram help a user when editing an image? Would it be better to keep the detail as is and use the saved CPU time elsewhere? Presumably so long as the user can see about a 100 vertical gradations of dark->light and a few hundred horizontal bins pixels, they should have adequate information. (Though on my fast/hidpi machine, `_pixelpipe_final_histogram_waveform()` now runs in about 0.006 secs., so it's not as if reducing that time via less detailed output will open up that much CPU power for other things.)
- Does the gamma caching add too much complexity/brittleness to the code? (The OpenMP changes seem pretty standard with similar code throughout dt, hence I'm less worried about that.)
- Is there a nicer way to draw the the grid lines, or a rationale for the existing code? (I've always missed there being a notation at 100%, hence this change.)

src/iop/levels.c
```diff
@@ -857,10 +857,11 @@ static gboolean dt_iop_levels_area_draw(GtkWidget *widget, cairo_t *crf, gpointe
   // draw grid
   cairo_set_line_width(cr, DT_PIXEL_APPLY_DPI(.4));
   cairo_set_source_rgb(cr, .1, .1, .1);
-  if(dev->histogram_type == DT_DEV_HISTOGRAM_WAVEFORM)
-    dt_draw_waveform_lines(cr, 0, 0, width, height);
-  else
-    dt_draw_vertical_lines(cr, 4, 0, 0, width, height);
+  // NOTE: if we actually draw a waveform in the background, we might want this:
```
TurboGit: I would just remove this commented out code:
- level is deprecated and one should prefer the rgblevel iop
- I don't think we will ever draw a waveform in any iop backgroud

src/libs/histogram.c
```diff
@@ -333,29 +339,36 @@ static gboolean _lib_histogram_motion_notify_callback(GtkWidget *widget, GdkEven
...
-  {
-    float black = d->black - (event->x - d->button_down_x) * .1f / (float)allocation.width;
-    dt_dev_exposure_set_black(darktable.develop, black);
+    const float diff = dev->histogram_type == DT_DEV_HISTOGRAM_WAVEFORM ? d->button_down_y - event->y
+                                                                        : event->x - d->button_down_x;
+    const int range = dev->histogram_type == DT_DEV_HISTOGRAM_WAVEFORM ? allocation.height
+                                                                       : allocation.width;
+    if (d->highlight == 2)
+    {
+      float exposure = d->exposure + diff * 4.0f / (float)range;
```
TurboGit: const

src/libs/histogram.c
```diff
@@ -333,29 +339,36 @@ static gboolean _lib_histogram_motion_notify_callback(GtkWidget *widget, GdkEven
...
+      dt_dev_exposure_set_exposure(darktable.develop, exposure);
+    }
+    else if(d->highlight == 1)
+    {
+      float black = d->black - diff * .1f / (float)range;
```
TurboGit: const

dtorop:
> A minor comment. Sure it would be nice to separate the UI from core/speed changes if it is not too much work. Thanks.

@TurboGit I spun out #4140 which has the UI changes (and has a commit to make the relevant fixes requested here). This PR now only has the core/speed-up changes.

src/iop/levels.c
dtorop: Sounds fair. Fixed in #4140, where this part of the code has migrated.

src/libs/histogram.c
dtorop: Fixed in #4140 (where UI changes now are).

src/libs/histogram.c
dtorop: Fixed in #4140 (where UI changes now are).

TurboGit:
> I'm still curious about whether the underlying waveform buffer should be this wide:

Indeed, good question. If I read correctly we had around 512k used for the histogram and now we have 16x more memory used. This is quite a lot, so (again if my figures are correct) we really need a good incentive to move to this direction. I don't have an HDPI screen, so I can't comment on the added quality/readability of the histogram. The screenshots you posted does show this clearly to me.

ptilopteri:
openSUSE Tumbleweed 20200116
NVIDIA GF106 [GeForce GTS 450], 390.129
darktable 3.1.0~git390.fa0e8ee01
OpenCL on 

there is a problem introduced here, if you cycle thru the different histogram displays for an image, returning to the default only shows the high end or low end depending on position of the cursor

dtorop:
> If I read correctly we had around 512k used for the histogram and now we have 16x more memory used.

@TurboGit It's a good point. Here are the relevant buffers:

- dev->histogram_waveform (always allocated in darkroom view)
- temporary buffer in `_pixelpipe_final_histogram_waveform()` to count colors (when preview pixelpipe runs)
- temporary buffer in `_lib_histogram_draw_callback()` (when histogram is drawn), always same size as histogram buffer

Here's the #'s I came up with:

- The "regular" histogram buffer is 4K (4 * 256 * sizeof(uint32_t)): pretty negligible, and a good baseline.
- The waveform histogram in dt 3.0.0 (pre fc0f6499f958a7facc35e1c72702b0f2287d52fb) used a buffer which varied with the panel dimensions. For the default 350px wide panel it was 239K (175 * 350 * 4). There's a 717K (175 * 350 * 3 * 4) temporary buffer to count colors. Hence the code uses 956K to calculate and 478K to draw.
- The waveform histogram as currently in master requires 180K (128 * (720/2) * 4). The 720 is pegged to `darktable.mipmap_cache->max_width[DT_MIPMAP_F]`, and 128 is hardcoded, hence neither varies with the UI. I set these to make sure that memory (and CPU use) was less than the prior code. The calculation temp buffer will be no larger than 540K (128 * (720/2) * 3 * 4). So the code uses 720K to calculate and 360K to draw.
- In this PR, I got carried away. The waveform buffer is 720x175 for lodpi, and 1440x350 for hidpi. It requires 492K (175 * 720 * 4) for lodpi and 1968K (350 * 1440 * 4) for hidpi to store the waveform histogram. Calculation uses a temp buffer no larger than 738K lodpi (720 * 175 * 3 * 2) or 2953K hidpi (1440 * 350 * 3 * 2) and a cache that should be smaller than 1K (hidpi) or 2K (lodpi). Hence peak usage to calculate will be 1232K for lodpi and 4922K for hidpi. Drawing will require 984K for lodpi and 3936K for hidpi.

I'm embarrassed to say I hadn't done this math before. I spent too much time looking at CPU speed and visuals. And even though hidpi screenshots look great on lodpi (as they're so big and detailed), a lot of the fine detail is nearly invisible at hidpi (the very rationale for hidpi, I guess...).

Furthermore, as the panel defaults to 350px wide, and is hardcoded at 175px tall, there's no need to have the waveform buffer to have a width of 720 (lodpi) or 1440 (hidpi). For lodpi a 360x175 buffer should be fine, or 720x350 for hidpi. (I'm pegging to MIP_F width in hopes that this makes sense with bin sizes.) Also, there's a good argument that the vertical resolution of the buffer should not be much larger than 100, if it's true that our eyes are sensitive to about that many tones in a dark to light (or dark to primary color) range.

If the waveform resolution is 360x175, the buffer will be 246K with calculation using ~373K (369K temp buffer, 3.5K cache), hence 619K to calculate and 492K to draw. This might be more reasonable -- lower peak memory usage than what is in master now (mainly due to using a uint16_t instead of uint32_t, though). I'll work on this more, as well as see what is happening with what @ptilopteri reported.

ptilopteri:
I also see great performance decline but cannot tie it to a particular commit.  on my systems dt is nearly impossible to use productively.

dtorop:
> darktable 3.1.0~git390.fa0e8ee01
> 
> there is a problem introduced here, if you cycle thru the different histogram displays for an image, returning to the default only shows the high end or low end depending on position of the cursor

> I also see great performance decline but cannot tie it to a particular commit. on my systems dt is nearly impossible to use productively.

@ptilopteri I'm not able to see this bug, or the slowdown you mentioned. I'm confused by the darktable version string. Are you running darktable head (fa0e8ee01)? If so, you're not running the code in this PR, though you are running the code in #4140, which did change histogram behavior.

In looking for this, I did find one long-standing bug: clicking rapidly on the histogram buttons could reset the exposure iop. That's fixed in ab2e00a in this PR.

Would you be able to make a screenshot of what you're seeing with the histogram? It would also be helpful to know if you're seeing this and the slowdown with dt v3.0.0 or a later version. It could be worth testing 9e0fb91c3b2020e0838159825e5fb8e239bab6b7 as it is the last commit before the PRs which changed the histogram post v3.0.0. Or `git bisect`.

Also, are you seeing the slowdown only when the waveform histogram is visible or at all times?

As of something like ebcd07f2450d1a226e6cc42a2cca5623d5ec6e3e, the waveform histogram is rescaled on the fly (by Cairo) before it is drawn. It's possible that with an older graphics card that this is a slow operation, but I'd be surprised (and interested) if this were true.

dtorop:
> we really need a good incentive to move to this direction

@TurboGit I lowered the resolution to 360 x 175. This is more than what is in master (360 x 128) and approx. what was in dt 3.0.0 (panel width x 175). Using this PR does give a speed increase, and has reasonable memory use compared to either dt 3.0.0 or master. But once could ask whether the added code is worth it.

FWIW, here is a screenshot of what current resolution looks like on hidpi:

waveform 360x175

I haven't rebased onto master, hence these are the old grid lines.

ptilopteri:
> > darktable 3.1.0~git390.fa0e8ee01
> > there is a problem introduced here, if you cycle thru the different histogram displays for an image, returning to the default only shows the high end or low end depending on position of the cursor
> 
> > I also see great performance decline but cannot tie it to a particular commit. on my systems dt is nearly impossible to use productively.
> 
> @ptilopteri I'm not able to see this bug, or the slowdown you mentioned. I'm confused by the darktable version string. Are you running darktable head ([fa0e8ee](https://github.com/darktable-org/darktable/commit/fa0e8ee017dd04994585e86e389b87ea3675de4f))? If so, you're not running the code in this PR, though you are running the code in #4140, which did change histogram behavior.
> 
> In looking for this, I did find one long-standing bug: clicking rapidly on the histogram buttons could reset the exposure iop. That's fixed in [ab2e00a](https://github.com/darktable-org/darktable/commit/ab2e00a0d660f47924d44f5ea501ac56a7cf54a5) in this PR.
> 
> Would you be able to make a screenshot of what you're seeing with the histogram? It would also be helpful to know if you're seeing this and the slowdown with dt v3.0.0 or a later version. It could be worth testing [9e0fb91](https://github.com/darktable-org/darktable/commit/9e0fb91c3b2020e0838159825e5fb8e239bab6b7) as it is the last commit before the PRs which changed the histogram post v3.0.0. Or `git bisect`.
> 
> Also, are you seeing the slowdown only when the waveform histogram is visible or at all times?
> 
> As of something like [ebcd07f](https://github.com/darktable-org/darktable/commit/ebcd07f2450d1a226e6cc42a2cca5623d5ec6e3e), the waveform histogram is rescaled on the fly (by Cairo) before it is drawn. It's possible that with an older graphics card that this is a slow operation, but I'd be surprised (and interested) if this were true.

openSUSE Tumbleweed 20200116
NVIDIA GF106 [GeForce GTS 450], 390.129
darktable 3.1.0~git390.fa0e8ee01
OpenCL on 

git390.fa0e8ee01 is the latest available for me, I do not build, package is from darix's OBS.

I replaced ~/.config/darktable/ with backup from prior to trying git390.fa0e8ee01and no longer see the display problem with the histogram but the left or right side is now highlighted (the other dark background) which was not the case before.

the performance decline is now intermittent but two things I noticed:
  [dt_ioppr_check_iop_order] history module not used but enabled!! exposure (179769313486231570814527423731704356798070567525844996598917476803157260780028538760589558632766878171540458953514382464234321326889464182768467546703537516986049910576551282076245490090389328944075868508455133942304583236903222948165808559332123348274797826204144723168738177180919299881250404026184124858368.000000) image 308163 (dt_dev_read_history_no_image end)

and:
  sqlite3 error: /home/abuild/rpmbuild/BUILD/darktable-3.1.0~git390.fa0e8ee01/src/iop/basecurve.c:384, function init_presets(), query "BEGIN": cannot start a transaction within a transaction
sqlite3 error: /home/abuild/rpmbuild/BUILD/darktable-3.1.0~git390.fa0e8ee01/src/iop/basecurve.c:390, function init_presets(), query "COMMIT": database is locked
sqlite3 error: /home/abuild/rpmbuild/BUILD/darktable-3.1.0~git390.fa0e8ee01/src/iop/lowlight.c:327, function init_presets(), query "BEGIN": cannot start a transaction within a transaction
sqlite3 error: /home/abuild/rpmbuild/BUILD/darktable-3.1.0~git390.fa0e8ee01/src/iop/lowlight.c:483, function init_presets(), query "COMMIT": database is locked
sqlite3 error: /home/abuild/rpmbuild/BUILD/darktable-3.1.0~git390.fa0e8ee01/src/iop/splittoning.c:138, function init_presets(), query "BEGIN": cannot start a transaction within a transaction
sqlite3 error: /home/abuild/rpmbuild/BUILD/darktable-3.1.0~git390.fa0e8ee01/src/iop/splittoning.c:176, function init_presets(), query "COMMIT": database is locked

TurboGit:
@ptilopteri : please this is another issue and nothing to do with the waveform do not post unrelated data here. you seems to have a broken edit on your db.

dtorop:
@TurboGit That's great, thank you!

PR #4133, opened 2020-01-17
XMP files are alway overwritten from the DB (bug #12247 fix)

esq4:
XMP files are alway overwritten from the DB regardless of the state of the “search for updated XMP files at startup” checkbox

### DIFF
--- a/src/control/crawler.c
+++ b/src/control/crawler.c
@@ -75,7 +75,7 @@ GList *dt_control_crawler_run()
     const int id = sqlite3_column_int(stmt, 0);
     const time_t timestamp = sqlite3_column_int(stmt, 1);
     const int version = sqlite3_column_int(stmt, 2);
-    gchar *image_path = (gchar *)sqlite3_column_text(stmt, 3);
+    gchar *image_path = g_locale_from_utf8((gchar *)sqlite3_column_text(stmt, 3),-1,NULL,NULL,NULL);
     int flags = sqlite3_column_int(stmt, 4);
 
     // no need to look for xmp files if none get written anyway.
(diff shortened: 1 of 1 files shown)


TurboGit:
Yes XMP are overwritten.... The fix seems to change filename to UTF-8, no clear to me given the change log what you are actually trying to solve. Can you clarify? Thanks.

esq4:
https://redmine.darktable.org/issues/12247  :

```
**Description**
If you set the "search for updated XMP files at startup" option, Darktable 
automatically updates the XMP files from the DB. This is in the Windows.
But in Linux Darktable allows me to select an action in the dialog box.

**History**
#1 Updated by Mike Romanov over 1 year ago
It looks like darktable ignores (does not detect) the changes in xmp files, 
gets history stacks from DB and then in darkroom just overwrites xmp files.

#2 Updated by m laverdiere about 1 year ago
Here's what I see in 2.6 version for Windows: when the option to look for 
modifications to xmp files on startup is selected, it does nothing on startup, 
i.e. it reloads the data from the database and eventually overwrite the xmp, 
without any dialog box to choose what to do. On my Linux box, I see that 
dialog box where the user can choose between loading the modified xmp 
or overwrite them.
```

In other words. I am editing the same shots on  Windows or Linux computers. On Linux, the programm allows me to accept the changes at startup, while on Windows it overwrites them with outdated data  from the database without asking.

The problem turned out to be that Windows 7 uses a regional code page for the file 
system (cp1251 in my case), while Darktable uses utf-8 to write file paths in db.

TurboGit:
Please DO NOT merge master branch into topic branches. If really needed you need to rebase.

TurboGit:
Merged in master. Thanks.

PR #4134, opened 2020-01-18
Panasonic DC-GH5 white balance presets

mjamado:
Added the Panasonic DC-GH5 white balance presets.


TurboGit:
Thanks! Merged in master and 3.0.1.

Issue #4040, opened 2020-01-06
Watermark preview blurs on rotation


simonkolotov:
When trying to rotate a watermark by more than 30 degrees it starts to blur very significantly.
My use case is that sometimes I will intentionally rotate the image by a significant angle and leave it uncropped.
In that case - if I choose to add a watermark - I like it rotated so it looks like a part of the image (see attached).

I noticed this behaviour in version 2.6 (maybe even earlier) and as I upgraded to 3.0 just now it's still there.

Note: As I was adding more information to this bug I noticed that it's a render issue - as the image is rendered smaller than 1:1, the SVG is blurred as if it were significantly downsampled.

Thanks,

Simon

ATTACHED:
screenshots at "full screen" lighttable and maximal zoom in darkroom, as well as a (downsampled) exported version of the image, where you see the watermark is blurred.
I also attach the SVG of the watermark

![IMG_9854_01](https://user-images.githubusercontent.com/4603429/71794561-f0678d80-2ff6-11ea-8f3e-ef7fc053fa13.jpg)
![Screenshot at 2020-01-05 19-58-54](https://user-images.githubusercontent.com/4603429/71794404-32dc9a80-2ff6-11ea-89f8-70bbd8a07b3d.png)
[SK.svg.zip](https://github.com/darktable-org/darktable/files/4024664/SK.svg.zip)

jenshannoschwalm:
@aurelienpierre you agree i go for this one? I know cairo stuff fairly well.

aurelienpierre:
@jenshannoschwalm No problem, thanks !

TurboGit:
@jenshannoschwalm : sure, the code is no one property :) go ahead ! I would say you just have to send a notice to avoid having many people working on the same issue.

jenshannoschwalm:
Has been fixed in master.

jenshannoschwalm:
@aurelienpierre please close this manually ...

jenshannoschwalm:
@TurboGit this had been fixed via #4135

TurboGit:
Right. Closing.

PR #4135, opened 2020-01-18
Watermark blurring and crashes fixed. Issue #4040

jenshannoschwalm:
Watermark blurring was reported but that was only one of the possible symptoms.
The main problem is the rsvg_handle_render_cairo function, it does not like rotated
complex svg data. This resulted in distortions of text, blurring, missing of text
parts while rotating or even crashes.

This fix does corrects this by adding another cairo surface that hold the scaled
but not rotated svg data. This is used as the source for the rotation afterwards.

There is a performance penalty if the svg image is large and we zoom in.

### DIFF
--- a/src/iop/watermark.c
+++ b/src/iop/watermark.c
@@ -895,22 +895,19 @@ void process(struct dt_iop_module_t *self, dt_dev_pixelpipe_iop_t *piece, const 
   }
 
   /* setup stride for performance */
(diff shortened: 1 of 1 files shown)


TurboGit:
Better a performance penalty than a crash or wrong rendering. This works for me. The penalty seems small on my side or bearable at least. Thanks.

PR #4139, opened 2020-01-18
Fixed metadata not allowing apostrophes

### LINKED ISSUE #4138: Cannot save title/description containg an apostrophe
Steps to reproduce:
* Select a photo in Lighttable
* Open Metadata Editor
* Enter `I'm here` to either title or description, press Enter or Apply
* The field becomes empty instead of saving the text

Tested on current master

sams96:
Instead of inserting the values into the SQL query via string formatting they now go in through SQL bind, which means they can contain anything and not cause problems (I think).

Fixes #4138.

I had to remove a couple of frees otherwise I'd get errors which I'm not totally sure about so if someone could have a quick look at that that'd be great.

Thanks

### DIFF
--- a/src/common/metadata.c
+++ b/src/common/metadata.c
@@ -57,20 +57,19 @@ static gchar *_get_tb_removed_metadata_string_values(GList *before, GList *after
     }
     if(!same || different)
     {
-      metadata_list = dt_util_dstrcat(metadata_list, "%d,", atoi(b->data));
+      metadata_list = dt_util_dstrcat(metadata_list, "%d", atoi(b->data));
     }
     b = g_list_next(b);
     b = g_list_next(b);
   }
-  if(metadata_list) metadata_list[strlen(metadata_list) - 1] = '\0';
   return metadata_list;
 }
 
-static gchar *_get_tb_added_metadata_string_values(const int img, GList *before, GList *after)
+static GList *_get_tb_added_metadata_string_values(int img, GList *before, GList *after)
 {
   GList *b = before;
   GList *a = after;
-  gchar *metadata_list = NULL;
+  GList *metadata_list = NULL;
 
   while(a)
   {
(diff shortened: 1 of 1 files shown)


src/common/metadata.c
```diff
@@ -84,12 +83,15 @@ static gchar *_get_tb_added_metadata_string_values(const int img, GList *before,
     }
     if(!same || different)
     {
-      metadata_list = dt_util_dstrcat(metadata_list, "(%d,%d,\'%s\'),", GPOINTER_TO_INT(img), atoi(a->data), (char *)a2->data);
+      char *img_str = NULL;
+      img_str = dt_util_dstrcat(img_str, "%d", GPOINTER_TO_INT(img));
+      metadata_list = g_list_append(metadata_list, img_str);
```
TurboGit: You need to g_strdup() all values added in the list.

src/common/metadata.c
```diff
@@ -107,30 +109,31 @@ static void _bulk_remove_metadata(const int img, const gchar *metadata_list)
...
 {
   gchar *tobe_removed_list = _get_tb_removed_metadata_string_values(before, after);
-  gchar *tobe_added_list = _get_tb_added_metadata_string_values(imgid, before, after);
+  GList *tobe_added_list = _get_tb_added_metadata_string_values(imgid, before, after);
 
   _bulk_remove_metadata(imgid, tobe_removed_list);
   _bulk_add_metadata(tobe_added_list);
 
   g_free(tobe_removed_list);
-  g_free(tobe_added_list);
```
TurboGit: You need to free the list here then:
g_list_free_full(to_be_added_list, g_free)

sams96: Thanks, I've added them in a new commit

src/common/metadata.c
```diff
@@ -84,12 +83,15 @@ static gchar *_get_tb_added_metadata_string_values(const int img, GList *before,
     }
     if(!same || different)
     {
-      metadata_list = dt_util_dstrcat(metadata_list, "(%d,%d,\'%s\'),", GPOINTER_TO_INT(img), atoi(a->data), (char *)a2->data);
+      char *img_str = NULL;
+      img_str = dt_util_dstrcat(img_str, "%d", GPOINTER_TO_INT(img));
+      metadata_list = g_list_append(metadata_list, g_strdup(img_str));
```
TurboGit: Not this one. dt_util_ddtrcat() already allocates strings.

TurboGit:
You also want to fix the undo the same way. Enter a title, do Ctrl+Z and you get:

`
sqlite3 error: ...src/src/common/metadata.c:105, function _bulk_remove_metadata(), query "DELETE FROM main.meta_data WHERE id = 36 AND key IN (2,)": near ")": syntax error`

sams96:
It doesn't need to be done the same way because when deleting it's not using the values. I'm not totally sure why it worked before but I just removed the comma from the delete statement and that fixed it.

sams96:
Never mind, apparently I just added that comma by accident, either way it's gone now.

TurboGit:
Thanks! Merged on master and 3.0.1 branch.

PR #4140, opened 2020-01-18
waveform histogram: ui improvements

dtorop:
- draw 50% grid line slightly thicker so it is legible, and draw 100% grid line thicker and dotted to call attention to it as a a reasonable upper limit
- when in waveform histogram, make the regions which adjust exposure via mouse drag/scroll be horizontal rather than vertical, to correspond to the waveform
- display waveform histogram even when image is totally black (previously it would disappear, which made sense for the regular histogram, but not for the waveform, for which we should see a horizontal line at the very bottom)
- remove vestigial code which sometimes displayed waveform grid lines if levels iop despite regular histogram always being show

These changes were originally in #4132 but are spun out into their own PR as they don't depend on the core and speed-up changes in the PR.

I made the changes which @TurboGit suggested on that PR in 305bae65c165b8f37fa76b37546d9e74856f3c65.

Here is how the waveform histogram looks after this PR on a hidpi display.

waveform with UI changes hidpi low detail

And this is the waveform histogram with the drag region for exposure highlighted via mouseover:

waveform with drag region

### DIFF
--- a/src/develop/develop.c
+++ b/src/develop/develop.c
@@ -110,6 +110,7 @@ void dt_dev_init(dt_develop_t *dev, int32_t gui_attached)
     dev->histogram_pre_tonecurve = (uint32_t *)calloc(4 * 256, sizeof(uint32_t));
     dev->histogram_pre_levels = (uint32_t *)calloc(4 * 256, sizeof(uint32_t));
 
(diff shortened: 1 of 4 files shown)


TurboGit:
All good to me! Thanks, will merge in master and 3.0.1.

PR #4144, opened 2020-01-19
Modify .gitignore for Ecplise users

Mark-64:
some more rows in .gitignore for people like me using Eclipse IDE


TurboGit:
Fine with me. Thanks!

PR #4154, opened 2020-01-20
Added simple unit test sample for demonstration

burrima:
- based on cmocka (https://cmocka.org) - thus, it needs to be installed before building. On Ubuntu 18.04: "apt install libcmocka-dev"
- build darktable with option -DBUILD_TESTING=On
- then run 'make test' in the build directory

Pull request contains 2 commits that were made earlier by @cryptomilk - thank you very much for your help!


cmake/modules/AddMockedTest.cmake
```diff
@@ -0,0 +1,57 @@
...
+## Add unit test with mocking support
+#  \param name unit test name (excluding extension and 'test_' prefix)
+#  \param SOURCES optional list of source files to include in test executable
+#  (beside test_${name}.c)
+#  \param MOCKS optional list of functions to be mocked in executable
+#  \param COMPILE_OPTIONS optional list of options for the compiler
+#  \param LINK_LIBRARIES optional list of libraries to link (used as
+#  -l${LINK_LIBRARIES})
+#  \param LINK_OPTIONS optional list of options to be passed to linker
+function(add_mocked_test name)
```
cryptomilk: Could you rename it to `add_cmocka_mock_test()`?

src/CMakeLists.txt
```diff
@@ -877,6 +877,9 @@ add_subdirectory(chart)
 
 # have some test programs for unit testing, ...
 if(BUILD_TESTS)
+  find_package(CMocka CONFIG REQUIRED)
```
cryptomilk: `find_package(cmocka 1.1.0 CONFIG REQUIRED)`

Please add a version requirement, it should be at least 1.1.0

cmake/modules/AddMockedTest.cmake
cryptomilk: Same for the file name ;-)

src/common/darktable.h
```diff
@@ -39,7 +39,7 @@
 #define O_BINARY 0
 #endif
 
-#include "ThreadSafetyAnalysis.h"
+#include "external/ThreadSafetyAnalysis.h"
```
cryptomilk: I think this should be split out and put into a separate patch?

src/CMakeLists.txt
cryptomilk: I would suggest to put the checking for cmocka and the addition of the cmake function for cmocka tests in a separate commit.

aurelienpierre:
Thanks a lot for this, that was so so so much needed ! I will let @cryptomilk do the review.

cryptomilk:
FYI: I will give a talk about cmocka at devconf.cz this weekend. I guess it is recorded and will be available next week.

cmake/modules/AddMockedTest.cmake
burrima: sure, no problem, makes sense

src/CMakeLists.txt
burrima: Added version requirement and made separate commit.
I have put the library check into src/tests/unittests/CMakeLists.txt - do you think this is the right place?

src/common/darktable.h
burrima: did so

TurboGit:
That's a wonderful step forward! Thanks a lot, for sure this will help dt to avoid regressions and will make big refactoring less scary! Looking forward to have this in action !

src/tests/unittests/CMakeLists.txt
```diff
@@ -0,0 +1,8 @@
+include_directories(${CMAKE_CURRENT_SOURCE_DIR}/../..)
+include_directories(${CMAKE_CURRENT_BINARY_DIR}/../..)
+
+find_package(cmocka 1.1.0 CONFIG REQUIRED)
+include(AddCMockaTest)
+include(AddCMockaMockTest)
+
+add_cmocka_mock_test(simplesample)
```
cryptomilk: This is a comment for the commit message.

If you want to run the tests, do:

    make test

or

    ctest --output-on-failure

In the build directory!

src/tests/unittests/test_filmicrgb.c
```diff
@@ -0,0 +1,145 @@
...
+ */
+void __wrap_dt_iop_color_picker_reset(dt_iop_module_t *module, gboolean update)
+{
+  check_expected_ptr(module);
+  check_expected(update);
+}
+
+dt_iop_order_iccprofile_info_t *__wrap_dt_ioppr_get_pipe_work_profile_info(struct dt_dev_pixelpipe_t *pipe)
+{
+  return NULL;
```
cryptomilk: Please use: `return mock_ptr(dt_iop_order_iccprofile_info_t *);`

src/tests/unittests/test_filmicrgb.c
```diff
@@ -0,0 +1,145 @@
...
+}
+
+/******************************************************************************
+ * TEST FUNCTIONS
+ */
+static void test_sample(void **state)
+{
+  expect_value(__wrap_dt_iop_color_picker_reset, module, NULL);
+  expect_value(__wrap_dt_iop_color_picker_reset, update, TRUE);
```
cryptomilk: `will_return(__wrap_dt_ioppr_get_pipe_work_profile_info, NULL);`

Instruct the mock to return NULL ...

src/tests/unittests/test_filmicrgb.c
```diff
@@ -0,0 +1,145 @@
...
+}
+
+
+/******************************************************************************
+ * MAIN FUNCTION
+ */
+int main()
+{
+  const struct CMUnitTest tests[]
+      = { cmocka_unit_test(test_sample), cmocka_unit_test(test_get_pixel_norm), cmocka_unit_test(test_process) };
```
cryptomilk: Please put one test per line, which is easier to read.

burrima: This was clang-format, I will fix it.

src/tests/unittests/CMakeLists.txt
burrima: Hmm, this doesn't work on my machine. I'll wait until you find time to look into it (as discussed on IRC).

cryptomilk:
I've checked what the issue is and it is rawspeed which uses `BUILD_TESTING`. So it is turned off for rawspeed but this turns it off globally.

burrima:
> I've checked what the issue is and it is rawspeed which uses `BUILD_TESTING`. So it is turned off for rawspeed but this turns it off globally.

Not sure if you are right. My code is listening to BUILD_TESTS and not BUILD_TESTING. But I just realized that it works when I include cmocka in the base CMakeLists.txt.

cryptomilk:
See https://cmake.org/cmake/help/latest/command/enable_testing.html

BUILD_TESTING is forced to OFF.

burrima:
> See https://cmake.org/cmake/help/latest/command/enable_testing.html
> 
> BUILD_TESTING is forced to OFF.

I see. To me, cmake is still very new. Do you have any suggestion on how to fix this issue? Current state of this PR is working fine but I think it only works by chance and it is not a proper solution (cmake should probably be included in ./src/CMakeLists.txt and not in ./CMakeLists.txt, right?).

cryptomilk:
We need to change the variable in rawspeed to BUILD_RS_TESTING first. It shouldn't use BUILD_TESTING.

burrima:
Ok, I think I understand now. I gave it a try and made a pull request for rawspeed: https://github.com/darktable-org/rawspeed/pull/223
In this pull request here, I still need to update the rawspeed submodule ref as soon as above PR has merged.

LebedevRI:
Please, please. When writing a commit message, don't just state "we need to do XYZ".
That may be FUD. State what the problem is, and why the particular solution may be the right fix.

burrima:
Thank you, @LebedevRI, your comment made me re-think again. As far as I understand now, the BUILD_TESTING is set to Off by rawspeed - or more precisely: in src/external/CMakeLists.txt. But it is done "hidden" - i.e. the original value of BUILD_TESTING is stored at the beginning and reverted at the end. Thus, I think the pull request in rawspeed is not needed.

Also, I found out that my first commit (fixing the include of ThreadSafetyAnalysis.h) is not needed anymore with the current implementation.

My current code is working fine on my machine (retried both cases with/without tests and deleted the buld dir each time before). I will update this PR. @cryptomilk could you please have a look again? Does it make sense to you as well?

LebedevRI:
> As far as I understand now, the BUILD_TESTING is set to Off by rawspeed - or more precisely: in src/external/CMakeLists.txt

Note that it is darktable's file, not rawspeed's.

> But it is done "hidden" - i.e. the original value of BUILD_TESTING is stored at the beginning and reverted at the end.

Yep, at least that is the intent.

burrima:
> Note that it is darktable's file, not rawspeed's.
Yep, updated my above comment

cryptomilk:
Ok, it looks like BUILD_TESTING is not a problem. I'm on the train home and finally tracked it down.

See here: https://github.com/cryptomilk/darktable/tree/master-fix

```
$ make test
Running tests...
Test project /home/asn/workspace/projects/darktable/obj
    Start 1: test_sample
1/2 Test #1: test_sample ......................   Passed    0.00 sec
    Start 2: test_test_filmicrgb
2/2 Test #2: test_test_filmicrgb ..............   Passed    0.05 sec

100% tests passed, 0 tests failed out of 2

Total Test time (real) =   0.05 sec
```

cryptomilk:
@burrima Do you want to push my branch here?

burrima:
@cryptomilk yes sure, thank you very much for your support. 

Honestly I cannot follow what issue you are addressing with the extra commits. On my PC I get (almost) exactly the same output:

`burrima@vulcano:~/src/darktable/build$ make test
Running tests...
Test project /home/burrima/src/darktable/build
    Start 1: test_simplesample
1/2 Test #1: test_simplesample ................   Passed    0.00 sec
    Start 2: test_filmicrgb
2/2 Test #2: test_filmicrgb ...................   Passed    0.02 sec

100% tests passed, 0 tests failed out of 2

Total Test time (real) =   0.03 sec`

Why do you get "test_test_filmicrgb" (note the double-"test_")?

cryptomilk:
I've already fixed that. I've split the code a bit more and include(CTest) needs to be included in the main CMakeLists.txt else it doesn't work.

burrima:
Did you get my latest code updates to this branch? - I already have the CTest included in the topmost CMakeLists.txt. The reason why I am asking is because I see that you still have an old state of test_fimlicrgb.c in your branch. I re-formatted it with clang and removed the ugly looking `/*******...` lines. Thus, just taking your branch is maybe not the best idea from my perspective.

I could include your 2 fixing commits before my commits if you wish so (though maybe a separate pull request for them would be more clever since I think the code is not related). But at the end, your voice has more power since you are in this project way longer than me :-)

BTW: how was your presentation? I looked through your slides on the devconf.cz site and found them very informative.

cryptomilk:
Can you merge my work into yours? :-)

burrima:
Yes, I'll do that. Sounds like the best strategy for me as well :-) Please give me some time since I'm going for dinner now.

burrima:
Ok, done.

BTW: your branch `master-fix` does not build since there is still a `test_` too much in src/cmake/modules/AddCMockaMockTest.cmake ;-)

cryptomilk:
@burrima Could you add https://github.com/cryptomilk/darktable/commit/c19aba4a33431dbfa4af924fa9ced38da496e413

burrima:
Cool, thank you @cryptomilk

cryptomilk:
Please rebase on master and repush tomorrow. New docker images with cmocka should be available then ...

burrima:
Ran docker on my ubuntu 18.04 pc. Then built dt inside the docker image. Tests then were running fine...
What I did in particular:
export SD=/home/burrima/src/darktable
sudo docker run -it --volume $SD:$SD:rw --workdir $SD darktable/darktable bash
cd build
export CC=/usr/bin/gcc-9
export CXX=/usr/bin/g++-9
cmake .. -DBUILD_TESTING=ON
cmake --build . -- -j6
ctest

cryptomilk:
@LebedevRI Roman, I'm not really familiar with TravisCI, do you have an idea why the tests fail with "Permission denied"?

@burrima Maybe also print:

```
tail -100 /var/log/messages
tail -100 /var/log/secure
tail -100 /var/log/audit/audit.log
```

cryptomilk:
@burrima Try with https://github.com/cryptomilk/darktable/commit/74477824a85327086dfe5fb7bffa1339bbf4c6a9

cryptomilk:
That's it. https://github.com/cryptomilk/darktable/commits/master-cmocka works :-)

cryptomilk:
@TurboGit This is good to go!

TurboGit:
@cryptomilk : Maybe I expected something else so I'm not sure to understand what is provided here. Is that only unit test on the code ?

What I thought was worked on was a way to give a source image, some actions like an XMP apply it and check that the output is consistent what what is expected (fuzzy comparison with the target image). Is that covered here ?

Whatever the answer to the question above I think we want a README in the test directory explaining how to add a test ? I did not find the high level description about this and I do feel this very important as adding tests is something we want to see encouraged as much as possible.

burrima:
@TurboGit This is code for function unit testing of the darktable code. I will add methods to create very simple test images in a next pull request. Thus, process() methods of modules can be tested. But unit-tested. So, no function testing as you described it.

A README is underway as well. Together with already implemented tests.

burrima:
@TurboGit feel free to look at my work in progress. But please be gentle, it's really still under progress :-)

https://github.com/burrima/darktable/tree/test_filmicrgb

cryptomilk:
@TurboGit Pascal, what you expected is called integration test, this is unit testing :-) Martin already found bugs in filmic rgb writing unit tests for it :-)

DefineOptions.cmake
```diff
@@ -25,7 +25,6 @@ option(BUILD_PRINT "Build the print module" ON)
 option(BUILD_RS_IDENTIFY "Build the darktable-rs-identify debug aid" ON)
 option(BUILD_SSE2_CODEPATHS "(EXPERIMENTAL OPTION, DO NOT DISABLE) Building SSE2-optimized codepaths" ON)
 option(VALIDATE_APPDATA_FILE "Use appstream-util (if found) to validate the .appdata file" OFF)
-option(BUILD_TESTS "Build tests in src/tests/, runnable from the build/ directory" OFF)
```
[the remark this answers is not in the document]
cryptomilk: There is already an option for this by cmake which is called BUILD_TESTING, see https://cmake.org/cmake/help/latest/command/enable_testing.html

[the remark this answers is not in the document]
cryptomilk: So we use BUILD_TESTING everywhere now.

burrima:
Hi @TurboGit, may I kindly ask what makes you hold back this pull request? Besides of documentation (which I promised to deliver in my next pull request) what else is missing or making you struggle? - Or do you just need more time to review? I'm not in a hurry at all though it would be a great satisfaction to me to see my first real contribution to darktable going in. Best regards, Martin

TurboGit:
@burrima : Just time and priorities. I've been working on integrating the new iop-order and working on big performance degradation. This PR is certainly a nice contribution, I just need more time to review and merge.

burrima:
Sure, no problem, take your time. I was just wondering if you are expecting more from my side.

DefineOptions.cmake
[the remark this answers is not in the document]
houz: Ok.

TurboGit:
@burrima : I had a look and the code as-is is ok to me. I'm not really sure how easy or feasible it will be to create serious test. BTW, is that to test only iop code or also the other parts of dt ? And in this case how all this will interact with the database (many routines are doing SQL) ? The pixel-pipe ? Are all those parts covered here ?

If it is only iop I think it is "simpler" and still just for this it will be nice to have in.

I have then nothing against merging this, still please give me some feedback regarding my questions above. Thanks.

burrima:
@TurboGit thank you very much for the review and your feedback. From my side, I will focus on the iop for a good while. But there is generally nothing against writing unit tests for other parts. The nice thing about cmocka is that you can "mock" (https://en.wikipedia.org/wiki/Mock_object) functions, meaning that you can "overwrite" functions that are called by the function-under-test to mimic different behavior. Thus, you can break dependencies (e.g. to the database etc.).

Unit tests are in my eyes just a start - there is definitely the need for more automated testing (e.g. when it comes to GUI interaction).

TurboGit:
Ok, so let's move on ! I'm eager to see unit test be added :) Thanks for the work!

burrima:
@TurboGit cool, thank you very much! Unit tests are in the pipe :-)

PR #4157, opened 2020-01-21
fix - GMIC not found message is now displayed

phweyland:
fix - GMIC not found message is now displayed when cmake does find the a GMIC library


TurboGit:
Obviously correct ! Thanks.

PR #4159, opened 2020-01-21
let users choose the cairo filter in pref

### LINKED ISSUE #4152: Slow and sluggish preview panning in dt 3.1 git
**Describe the bug**
Panning a zoomed image is slow and unresponsive. Clear stutter in preview image and "rubberbanding" between the mouse cursor and preview image. Seems to be worst when around 50% zoom. 100% zoom performs okay. Zooming in and out (with mouse wheel) also feels less responsive than with 3.0.

**To Reproduce**

Zoom preview to ~50% and try panning. Compare with 3.0 release -> big difference.

**Expected behavior**
Responsive and stutter-free panning and zooming

**Screenshots**
Screen recording comparison of both 3.0 and 3.1, not sure how well the difference comes across in this (OpenCL off in both cases):
https://streamable.com/uv48s

 - Darktable Version: darktable-3.1.0~git414
 - OS: Fedora Workstation 31 GNOME DE
 - OpenCL on could be slightly better, hard to say really. OpenCL on produces more single core spiking in 3.1 than 3.0 when panning.
 - NVIDIA Quadro w/ NVIDIA 440.44 driver

**Additional context**

No, 3.0 is smooth as butter. I'm referencing [this fix](https://github.com/darktable-org/darktable/pull/3957) for previously reported performance regression (maybe the panning regression is caused by something similar).

aurelienpierre:
Fix #4152
Set default cairo filter to fast (== nearest). Keep the nearest setting in parts of the code where exact buffer/display pixel sizes is ensured.
Default can be changed in darktablerc: "ui/cairo_filter = fast, good, best"

### DIFF
--- a/src/gui/gtk.c
+++ b/src/gui/gtk.c
@@ -1492,9 +1492,9 @@ static void init_widgets(dt_gui_gtk_t *gui)
 
   if(dt_conf_key_exists("ui/cairo_filter"))
   {
-    if(strcmp(dt_conf_get_string("ui/cairo_filter"), "best"))
+    if(strcmp(dt_conf_get_string("ui/cairo_filter"), "best") == 0)
       gui->filter_image = CAIRO_FILTER_BEST;
-    else if (strcmp(dt_conf_get_string("ui/cairo_filter"), "good"))
+    else if (strcmp(dt_conf_get_string("ui/cairo_filter"), "good") == 0)
       gui->filter_image = CAIRO_FILTER_GOOD;
   }
   else
--- a/src/gui/gtk.c
+++ b/src/gui/gtk.c
@@ -1486,6 +1486,19 @@ static void init_widgets(dt_gui_gtk_t *gui)
 
   gtk_widget_set_visible(gui->scrollbars.hscrollbar, FALSE);
   gtk_widget_set_visible(gui->scrollbars.vscrollbar, FALSE);
+
+  // Fetch the cairo filter to draw scaled surfaces where exact 1:1 buffer/viewport size is not guaranteed
+  gui->filter_image = CAIRO_FILTER_FAST;
+
+  if(dt_conf_key_exists("ui/cairo_filter"))
+  {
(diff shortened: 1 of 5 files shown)


src/gui/gtk.c
```diff
@@ -1486,6 +1486,19 @@ static void init_widgets(dt_gui_gtk_t *gui)
 
   gtk_widget_set_visible(gui->scrollbars.hscrollbar, FALSE);
   gtk_widget_set_visible(gui->scrollbars.vscrollbar, FALSE);
+
+  // Fetch the cairo filter to draw scaled surfaces where exact 1:1 buffer/viewport size is not guaranteed
+  gui->filter_image = CAIRO_FILTER_FAST;
+
+  if(dt_conf_key_exists("ui/cairo_filter"))
+  {
+    if(strcmp(dt_conf_get_string("ui/cairo_filter"), "best"))
```
TurboGit: This should be !strcmp

src/gui/gtk.c
```diff
@@ -1486,6 +1486,19 @@ static void init_widgets(dt_gui_gtk_t *gui)
...
+      gui->filter_image = CAIRO_FILTER_BEST;
+    else if (strcmp(dt_conf_get_string("ui/cairo_filter"), "good"))
```
TurboGit: likewise

aurelienpierre:
I have put `fast` as a default because it was the default mode for the longest time, and nobody seemed to complain. I changed that in the UI revamping last April out of precaution for the edge cases, but since it's now configurable, I think we can default to the fastest and then point out other options if users complain.

src/gui/gtk.c
aurelienpierre: strcmp can output 1, -1, or 0, so I have put == 0 instead (I have no idea what `!(-1)` is)

TurboGit:
Perfect ! That's merged in master and 3.0.1. Thanks.

PR #4160, opened 2020-01-21
drawn masks: sigmoidal transition in gradient shape

### LINKED ISSUE #4145: DT 3: Strange hard edge on gradient mask
**Describe the bug**
I’m having a great time with the new Darktable version 3, and especially the new Filmic RGB and other RGB modules which are simply fantastic.

But now I’ve come across an issue which I can’t really explain. In one photo I have a very bright sky and so I want to use the Exposure module with a gradient mask to select the sky and not the ground. I set the gradient to roughly match the horizon and fade into the sky over an area of perhaps 40-50 pixels.

I then dial down the exposure. And I do that quite a bit, I think down to the full range of -3 ev. Mostly just because I wanted to see how much I could recover.

Anyway, what I see is that at the exact point where the gradient part of the mask stops and the “solid mask” begins, there is a very hard edge. The top part of the sky is darkened down by -3 ev, and then it very abruptly goes up to perhaps -1 ev and fades out to 0 ev at the horizon. Right where the actual gradient begins there is a sudden jump in brightness that I can’t explain.

**To Reproduce**
I have an example image with sidecar to demonstrate the issue:

[dt3-gradient-issue.zip](https://github.com/darktable-org/darktable/files/4082758/dt3-gradient-issue.zip)

In the example file, there is an adjustment by the exposure module called "Exposure 1". As you can see the exposure compensation has been dialed down all the way to -3 ev, which demonstrates the problem.

I understand there are sliders to further adjust the gradient mask including contrast (s-shape of the gradient), feathering and other more exotic parameters. Feathering is not useful in this case because it affects other things than just the gradient (the area around the thin branches in the foreground get halos). The mask contrast slider does help somewhat but out past 0.67 the "hard edge" in the gradient just moves to a different place. Frankly I would not expect to have to make adjustments to any of these sliders to achieve the basic functionality of providing a smooth gradient.

I asked over at the dt forum on discuss.pixls.us ([link here](https://discuss.pixls.us/t/darktable-3-strange-hard-edge-on-gradient-mask/15910)) and noone there could explain it.

**Expected behavior**
I expect a gradient mask to provide a smooth gradient from full (100%) to nothing (0%) over a defined distance. In my example, this does not appear to be the case. The gradient falls off much too sharply at the start.

**Screenshots**
No screenshots but please see the attached example raw file (dng format) and sidecar file.

[dt3-gradient-issue.zip](https://github.com/darktable-org/darktable/files/4082758/dt3-gradient-issue.zip)

 - Darktable Version: 3.0.0
 - OS: Ubuntu 19.10
 - OpenCL not activated as far as I know
 - No proprietary graphics drivers in use
$ uname -a
Linux michael-kontor 5.3.0-26-generic #28-Ubuntu SMP Wed Dec 18 05:37:46 UTC 2019 x86_64 x86_64 x86_64 GNU/Linux

upegelow:
Make the transition in gradients sigmoidal rather than linear. Legacy
gradients still process as they did. If needed users can toggle between
the default sigmodial and linear transition by shift-click in the gradient.
This commit also gives up on the steepness parameter. That functionality
has always been broken. Fortunately that parameter has been set to zero
from day one and was never exposed to the user.

This fixes #4145

### DIFF
--- a/src/develop/masks/gradient.c
+++ b/src/develop/masks/gradient.c
@@ -1146,19 +1146,30 @@ static int dt_gradient_get_mask(dt_iop_module_t *module, dt_dev_pixelpipe_iop_t 
     return 0;
   }
 
+#ifdef _OPENMP
+#if !defined(__SUNOS__) && !defined(__NetBSD__)
+#pragma omp parallel for default(none) \
+  dt_omp_firstprivate(lutsize, lutmax, hwscale, state, normf, compression) \
+  shared(lut)
+#else
+#pragma omp parallel for shared(points)
+#endif
+#endif
   for(int n = 0; n < lutsize; n++)
   {
     const float distance = (n - lutmax) * hwscale;
     const float value = 0.5f + 0.5f * ((state == DT_MASKS_GRADIENT_STATE_LINEAR) ? normf * distance: erff(distance / compression));
     lut[n] = (value < 0.0f) ? 0.0f : ((value > 1.0f) ? 1.0f : value);
   }
+
+  // center lut around zero
   float *clut = lut + lutmax;
 
 
 #ifdef _OPENMP
 #if !defined(__SUNOS__) && !defined(__NetBSD__)
 #pragma omp parallel for default(none) \
-  dt_omp_firstprivate(gh, gw, sinv, cosv, xoffset, yoffset, hwscale, ihwscale, curvature) \
+  dt_omp_firstprivate(gh, gw, sinv, cosv, xoffset, yoffset, hwscale, ihwscale, curvature, compression) \
   shared(points, clut)
 #else
 #pragma omp parallel for shared(points)
@@ -1309,12 +1320,23 @@ static int dt_gradient_get_mask_roi(dt_iop_module_t *module, dt_dev_pixelpipe_io
     return 0;
   }
 
+#ifdef _OPENMP
+#if !defined(__SUNOS__) && !defined(__NetBSD__)
+#pragma omp parallel for default(none) \
+  dt_omp_firstprivate(lutsize, lutmax, hwscale, state, normf, compression) \
+  shared(lut)
+#else
+#pragma omp parallel for shared(points)
+#endif
+#endif
   for(int n = 0; n < lutsize; n++)
   {
     const float distance = (n - lutmax) * hwscale;
     const float value = 0.5f + 0.5f * ((state == DT_MASKS_GRADIENT_STATE_LINEAR) ? normf * distance: erff(distance / compression));
     lut[n] = (value < 0.0f) ? 0.0f : ((value > 1.0f) ? 1.0f : value);
   }
+
+  // center lut around zero
   float *clut = lut + lutmax;
 
 #ifdef _OPENMP
--- a/src/develop/masks/gradient.c
+++ b/src/develop/masks/gradient.c
@@ -152,7 +152,6 @@ static int dt_gradient_events_button_pressed(struct dt_iop_module_t *module, flo
     dt_masks_point_gradient_t *gradient = (dt_masks_point_gradient_t *)(g_list_first(form->points)->data);
 
     gradient->curvature = 0.0f;
-    gradient->state = DT_MASKS_GRADIENT_STATE_LINEAR;
     dt_dev_add_masks_history_item(darktable.develop, module, TRUE);
 
     dt_masks_gui_form_remove(form, gui, index);
--- a/src/develop/masks/gradient.c
+++ b/src/develop/masks/gradient.c
@@ -152,12 +152,22 @@ static int dt_gradient_events_button_pressed(struct dt_iop_module_t *module, flo
     dt_masks_point_gradient_t *gradient = (dt_masks_point_gradient_t *)(g_list_first(form->points)->data);
 
     gradient->curvature = 0.0f;
+    gradient->state = DT_MASKS_GRADIENT_STATE_LINEAR;
     dt_dev_add_masks_history_item(darktable.develop, module, TRUE);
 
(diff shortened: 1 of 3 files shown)


src/develop/masks/gradient.c
```diff
@@ -1084,22 +1126,40 @@ static int dt_gradient_get_mask(dt_iop_module_t *module, dt_dev_pixelpipe_iop_t
...
+  }
+  float *clut = lut + lutmax;
+
 
 #ifdef _OPENMP
 #if !defined(__SUNOS__) && !defined(__NetBSD__)
 #pragma omp parallel for default(none) \
-  dt_omp_firstprivate(gh, gw, sinv, cosv, xoffset, yoffset, hwscale, normf, steepness, curvature) \
-  shared(points)
+  dt_omp_firstprivate(gh, gw, sinv, cosv, xoffset, yoffset, hwscale, ihwscale, curvature) \
```
TurboGit: You need to add compression in the firstprivate list.

TurboGit:
Good now! Thanks a lot @upegelow !

TurboGit:
As we don't have curved gradient in 3.0.1 I won't merge this in this branch as there is conflicts.

aurelienpierre:
I still don't get the benefit of a sigmoidal transition.

mberg2007:
I wanted to give this a try since the source issue was reported by me.

So I built dt from the current master (a945c2c) and got some strange results. I am no longer able to change the extent of the mask gradient. Rolling the mouse wheel up and down (previously this would expand or contract the gradient) very strangely now changes the shape of the gradient horizon. It goes from a straight line to a smile (yes a curve) when rolling the wheel forward, or a downsmile when going the other way. It's the most bizarre thing I've ever seen.

I tried a number of modifying keys like alt and ctrl while rolling, but there is no way to change the size of the gradient anymore. Just the curve of the gradient.

I have not been keeping up on current developments in dt so this might actually be a feature in which case I apologize. But I wanted to mention it because it could also be a bug in the fix that was committed here.

If it is a feature then how do you change the gradient extent now?

upegelow:
Yepp, this is a new feature plus a change on how to set the parameters. In order to be somewhat consistent with the other shapes scroll without a modifier changes the form/size of a shape (in gradient: change the curvature). Shift+scroll changes the feathering/hardness of the shape (in gradient: the width of the transition area). There should be a tooltip at the top border which explains this.

I am optimistic that someone finds better ways on how to revise the handling so that things are even more easy and consistent.

upegelow:
BTW the option for a curved gradient has been introduced on user request. #3731 is an example. But there have been many more in the past.

mberg2007:
@upegelow Thanks a lot for taking time to fill me in. I did figure that this was probably a feature :-)

While I understand the sentiment to align mouse wheel functionality with other shape tools, from a user experience point of view this was rather confusing. I never did find the tooltip which could have helped me here. If you put the tooltip on the "add gradient" icon itself I think more users will find it.

A different way to think about this might be to consider what functions are used "most" and provide those functions with the least amount of difficulty for the user. With respect to a gradient mask, I would say that the mask transition area size is something you always (as in *always*) adjust when using this tool, while the curvature is a secondary and likely only rarely used type of capability. With that in mind it might be worth considering assigning the curvature to a modifier in stead. This would also provide a more consistent experience to existing users of Darktable, who I expect will be as surprised as I was to discover this change. I'll be happy to create a separate issue on that if you think this suggestion would have a chance to gain traction.

As for the gradient transition itself I'm happy to report that this looks great now. Thanks a lot for fixing this!

upegelow:
Thanks for your comment. I am happy to discuss any UI change on handling the shapes. Ideally you open up a new issue and then let's see what opinions others may add.

mberg2007:
> BTW the option for a curved gradient has been introduced on user request. #3731 is an example. But there have been many more in the past.

I just took a minute to review that issue and boy this is a can of worms isn't it :-) It never occurred to me that coordinates change in the pixel pipeline due to geometric transformations like lens correction and how this affects shapes.

I do see how distortions can impact a gradient mask and how an option to curve the gradient can help with that to some extent. But the implementation does not compensate for more exotic distortions like liquify for example. The curved gradient mask is a solution to a rather special case of distortion more than being a general solution unfortunately. This again suggests that the option to curve the gradient mask should be selectable through a modifier, rather than being the default behavior for the mouse wheel. I'll be happy to open a separate issue on that.

I wonder why this problem does not affect the crop tool and what considerations were done with respect to geometric distortions taking place before or after the crop. Seems like a similar solution might have been possible to solve the problem of gradient masks being distorted by lens corrections.

upegelow:
> The curved gradient mask is a solution to a rather special case of distortion more than being a general solution unfortunately.

Think of it like a digital version of a physical GND filter. The standard use case is fitting the gradient to the (straight) horizon. If you happen to have a lens with strong distortions the horizon on your sensor will not be straight but curved. As all shapes in darktable are relative to the input image coordinate system this requires a curved line to fit. Of course a simple parabolic shape will not compensate for the more complicated cases but it will be much better than a simple straight line. The rest should be achieved with a bit of blur. In the end also physical GND filters are no super sharp.

mberg2007:
> Think of it like a digital version of a physical GND filter. The standard use case is fitting the gradient to the (straight) horizon. If you happen to have a lens with strong distortions the horizon on your sensor will not be straight but curved.

Yes I understand that. However I think that with Darktable graduated masks are being used for many things, including lightening the left, bottom or right side of an image. Or a corner. Or applying gradually applying a clarity effect across specific parts of the image. Darkening skies and brightening foregrounds are just a few of the many useful cases for this tool. I think that is also the challenge here, because how do you reconcile these different use cases with a consistent UI experience.

Issue #4170, opened 2020-01-22
Import image stuck darktable


Nilvus:
**Describe the bug**
Clicking on image button in import options (or when dialog window open on a RAW file) makes darktable stuck. Need to force quit darktable. 
~~Reproducible each time on your last commit, @TurboGit, 092f7408a98ed009163e87484bfba9359283bdb2.~~

**To Reproduce**
Just click on import image button. Import folder is ok.

**Expected behavior**
Having dialog window and possibility to import correctly image.

**Platform (please complete the following information):**
darktable master
Ubuntu 19.10

Nilvus:
Forget to wait finishing recompiling again your last commit and now it's working. I made same steps on compiling (even removing build folder and darktable install one before). After having this issue on your commit then after compiling with previous commit where that was working, or to be more precise, I've had the dialog window but it sucks then when selecting image (before clicking on open button).

Nilvus:
Last test made : at first I have had the stuck problem with no dialog import window opening on this button always on last commit cited and not on other before (on tests made tonight) and now after recompiling last one and previous one, I had dialog window opening on two commits cited on start and now have darktable stuck on that window on both commits. So it seems that's not related to 092f7408a98ed009163e87484bfba9359283bdb2 commit. 

Quite strange behavior. Will try to make some more tests and previous commits to better understanding this strange issue...

aurelienpierre:
I have this behaviour too, it's been a couple of days (with e28c49986a03566c01e72e3f4ef15152cc32f217 and already a bit before). When importing a single image, as soon as I click on the picture, darktable hangs forever. It seems only some raw do that, not TIFF. I'm busy on another part of the code, so I didn't investigate more for now.

Nilvus:
@aurelienpierre : thanks for reporting that. More tests I've made since show that was older indeed. I use that option not really often. So a bug to fix (need some labels ;-) !)

TurboGit:
This is due to recent work on metadata.

@phweyland : Can you have a look ? I have seen that dt is locked into _get_tb_added_metadata_string_values().

phweyland:
> @phweyland : Can you have a look ? I have seen that dt is locked into _get_tb_added_metadata_string_values().

Yes, I'll do.

TurboGit:
Issue seems to be in _metadata_execute(imgs, metadata, &undo, undo_on, DT_MA_ADD); 
called from metadata.c:627

That's what I have learnt from initial investigation.

[...]

Issue #4151, opened 2020-01-20
Crash on export in DT 3.1 git master


Mark-64:
DT 3.1 git master, built today. Nvidia 1060.
See the following image (free of copyright):

[https://drive.google.com/file/d/1VpPb7Lr-vCfuNeJFOxH0IPk9J8wOa0Pr/view?usp=sharing](url)
[DSC04344.ARW.xmp.txt](https://github.com/darktable-org/darktable/files/4084853/DSC04344.ARW.xmp.txt)

Export result is DT freeze on Ubuntu 19.10 and crash on Windows 10, with or without OpenCL, export format jpeg or tiff.
Here is the backtrace on Ubuntu 
[darktable_bt_QEKVE0_ubuntu.txt](https://github.com/darktable-org/darktable/files/4084862/darktable_bt_QEKVE0_ubuntu.txt)

and on Windows
[darktable_bt_VSY2E0_windows.txt](https://github.com/darktable-org/darktable/files/4084864/darktable_bt_VSY2E0_windows.txt)

st-binder:
DT branch 3.0.x, built today crashes on export to disk too. Revert to Commit 4d0717dda60a9915a17dd0f5b06bdcbd2b4259ec works for me!
system: Arch Linux, AMD, AMD Polaris GPU

Mark-64:
Reverting to 3.1.0+308~g2f82a6e43 works for me too, so it must be one of the latest commits

junkyardsparkle:
I can't reproduce this so far myself on c2b133cfecaacea50f820660c5583d5f91d3e0d6, but possibly it's related to the stuff fixed in #4147?

st-binder:
latest version of DT 3.0.x branch works fine for me, export doesn't crash DT anymore.

Mark-64:
The crash is caused by corrupted metadata in library.db.
In a working version of the database I have this

![immagine](https://user-images.githubusercontent.com/43290988/72842043-e51f8d80-3c97-11ea-9dbc-c732e9def975.png)

while in the corrupted it's like this

![immagine](https://user-images.githubusercontent.com/43290988/72842176-231cb180-3c98-11ea-93f3-29ecfbbb7f94.png)

note the _NULL_ value instead of the string "(NULL)". This cause the function dt_metadata_get(imgid, "Xmp.dc.description", NULL) in exif.cc:1710 to return an invalid char* which provoke the crash.
[...]

PR #4176, opened 2020-01-23
Diverse fixes - metadata

phweyland:
- I haven't reproduced the issue #4170 but I've fixed a case related to import a single image. In that case imgid = 0 is used to get the orientation of the imported image. Surprisingly import folder doesn't use this.
- I haven't reproduced the issue #4151 either. But reading metadata from DB should be now protected against NULL metadata value pointers.
- Writing / updating metadata to DB removes empty metadata from DB and avoids to store empty metadata.
- Fix a bug related to change on multiple image selection. <leave unchanged>  metadata were wiped when one changes another item.
- Revert previous apostrophe fix and apply a simpler one.
- A bit of wording to make the code more readable

I think there is still a lot of cleaning to be done on metadata. In particular remove specific metadata code and generalize it. Adding one metadata item should be done in a single place with (almost) no code change otherwise. There are also some pieces of code which are not used. I'll open another PR for that.

### DIFF
--- a/src/common/metadata.c
+++ b/src/common/metadata.c
@@ -47,15 +47,16 @@ static gchar *_get_tb_removed_metadata_string_values(GList *before, GList *after
 
   while(b)
   {
-    GList *same = g_list_find_custom(a, b->data, _compare_metadata);
+    GList *same_key = g_list_find_custom(a, b->data, _compare_metadata);
     GList *b2 = g_list_next(b);
-    gboolean different = FALSE;
-    if(same)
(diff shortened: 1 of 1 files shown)


TurboGit:
Works for me and the issue seems to be fixed! I'll merge right now. Thanks !

TurboGit:
Merged in master and 3.0.1.

Nilvus:
I can confirm that this PR fix the #4170 issue. Thanks @phweyland.

PR #4179, opened 2020-01-24
Fix #4005 icons size for HiDPI displays

### LINKED ISSUE #4005: UI: super tiny on/off history icons on Windows 10
DT 3.1 git master on Windows 10, 4k monitor.
The on/off icons in the history stack are super small
Cattura14

while the same icons in modules are normal
Cattura15

Mark-64:
fix for #4005 and top right icons.
In 4k display, Windows 10:

before 
Cattura22

after
Cattura23

before
Cattura24

after
Cattura27

### DIFF
--- a/src/dtgtk/button.c
+++ b/src/dtgtk/button.c
@@ -49,9 +49,6 @@ static gboolean _button_draw(GtkWidget *widget, cairo_t *cr)
   /* update paint flags depending of states */
   int flags = DTGTK_BUTTON(widget)->icon_flags;
 
-  /* set inner border */
-  int border = DT_PIXEL_APPLY_DPI((flags & CPF_DO_NOT_USE_BORDER) ? 4 : 6);
-
   /* prelight */
   if(state & GTK_STATE_FLAG_PRELIGHT)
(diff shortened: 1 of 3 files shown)


TurboGit:
Looks ok to me!

@Nilvus : What do you think?

Nilvus:
> Looks ok to me!
> 
> @Nilvus : What do you think?

I will not have time this week to test but on what I see in screen captures, that's great. And it's ok if it doesn't change anything on lowdpi displays and on other OS like Linux or MacOs.

Mark-64:
The correction only depends on darktable.gui->dpi_factor and it's pretty mild. 
For lowdpi screens the factor is normally 1.0 and there is no change. 
The correction becomes visible only when dpi_factor is significantly higher than 1.0, like in my 4k display under Windows10 where it is 1.75. The same display in Ubuntu has a factor of 1.0, don't know why.
More tests are welcome of course.

TurboGit:
It changes nothing on lowdpi, so I'll merge on master and 3.0.1. Thanks!

PR #4193, opened 2020-01-27
test if data base exists before backup and make backup non-writable

rabauke:
This patch is to avoid warnings about failing database backups, when darktable has not been used before. It also makes the backup read-only to avoid accidental damage of the backup.


TurboGit:
Thanks !

TurboGit:
Merged in master and 3.0.1.

PR #4194, opened 2020-01-27
Metadata cleaning

phweyland:
The main goal of this PR is to generalize metadata to be able then to add new metadata at a single place.

For that I've removed the txt configuration file and concentrated necessary definition on common/metadata.c and h. Then I've cleaned up the usage of "Xmp.dc..." where possible, and some unused pieces of code as well.
However I have let iptc & exif conversions to xmp in exif.cc.
The generalization impacts mostly libs/metadata.c, libs/import.c, libs/metadata_view.c and exif.cc.

Beside this, to illustrate the new possibilities, I've added a small configuration dialog for metadata where the user can hide the metadata he doesn't need and set as private the metadata he doesn't want to export (as for tags). Currently I use the same hidden flag in image information with the same purpose. When the user imports new images, the hidden flag is used to filter imported metadata. This last one would be better sit in an actual import metadata configuration (I think there is a lot of work there, another PR).

I've also set <leave unchanged> in italic and added a tool-tip to see the full text when it doesn't fit in the entry box.

Last, I've added a "notes" field as suggested [here](https://discuss.pixls.us/t/metadata-with-darktable/13707/7?u=phweyland) which was a good idea for me.

But, I've failed to turn these fields multi-line (at least wrapped if too long). If somebody has a good idea I'm ready to spend a bit more time on this.

Some snapshots:

image

image

image

image

Comments and suggestions are welcome.


phweyland:
I think I've found a way ...
image

phweyland:
image


In case of selected images with different metadata click-right pops up a window to choose one of them.

image

Not yet done: completion in case of selected images with different metadata. Not very useful but I'll try to port the entry completion to text view completion.

Problem not yet solved: At gui_init() time gtk_widget_set_no_show_all() is usually used not to show a widget despite the gtk_widget_show_all() made by the plugin management (lib.c). Usually that works, but doesn't with textview nor with its scrolledwindow... (with no-show-all the textview is never displayed).
Currently metadata widget appears with all metadata (despite the selection the user has made). 
The user selections becomes effective after hoovering images.
Any idea ?

At start time (no-show-all works for labels):
image

phweyland:
Height of textviews is controlled with ctrl-scroll as for tagging views.

phweyland:
completion on textview will come with gtk4, using sourceview and sourcecompletion features.

phweyland:
I'm done with the scope I've given to myself for that PR.

There is an adjustment for which I haven't not made my mind:
Xmp.acdsee.notes could be replaced by Xmp.darktable.notes. The first one is exportable (with the current mechanisms) on exported images. The second one would allow to have a real multi line field (with lf inside), what I cannot do with standard Xmp tags.
What do you think ?

phweyland:
@TurboGit , I don't see what I could add here. Except if you see something on your side you can removed the wip label.

I see two related topics but I think they deserve specific PRs:
- import configuration to let user control what he actually imports (same spirit as export configuration but probably simpler)
- generalization of metadata in collection. And, I don't know if the current design allow this, but I think the possibility to get images without metadata, without tags, without... would be appreciated by users.

Please let me know what you think.

src/libs/metadata.c
```diff
@@ -296,18 +285,164 @@ int position()
...
+
+  GtkWidget *grid = gtk_grid_new();
+  gtk_container_add(GTK_CONTAINER(area), grid);
+  GtkWidget *label = gtk_label_new(_("metadata"));
+  gtk_grid_attach(GTK_GRID(grid), label, 0, 0, 1, 1);
+  label = gtk_label_new(_("hidden"));
+  gtk_grid_attach(GTK_GRID(grid), label, 1, 0, 1, 1);
+  label = gtk_label_new(_("private"));
+  gtk_grid_attach(GTK_GRID(grid), label, 2, 0, 1, 1);
+  label = gtk_label_new(_("height"));
```
TurboGit: What is this height ? I have it on my dialog but there is no check box associated ? It does not appears on your screenshot BTW.

TurboGit:
Capture d’écran de 2020-02-11 22-19-12

With the height column.

TurboGit:
The width too big:

Capture d’écran de 2020-02-11 22-19-12

src/libs/metadata.c
```diff
@@ -324,100 +459,243 @@ void connect_key_accels(dt_lib_module_t *self)
...
+  grid = (GtkGrid *)gtk_grid_new();
   gtk_grid_set_column_homogeneous(grid, TRUE);
 
-  button = gtk_button_new_with_label(_("clear"));
+  GtkWidget *button = gtk_button_new_with_label(_("clear"));
   d->clear_button = button;
   gtk_widget_set_tooltip_text(button, _("remove metadata from selected images"));
-  gtk_grid_attach(grid, button, 0, 0, 1, 1);
-  g_signal_connect(G_OBJECT(button), "clicked", G_CALLBACK(clear_button_clicked), (gpointer)self);
+  gtk_grid_attach(grid, button, 0, 0, 5, 1);
```
TurboGit: to allow for smaller panel: 5 -> 3

(see below)

src/libs/metadata.c
```diff
@@ -324,100 +459,243 @@ void connect_key_accels(dt_lib_module_t *self)
...
+  g_signal_connect(G_OBJECT(button), "clicked", G_CALLBACK(clear_button_clicked), self);
 
   button = gtk_button_new_with_label(_("apply"));
   d->apply_button = button;
   gtk_widget_set_tooltip_text(button, _("write metadata for selected images"));
-  gtk_grid_attach(grid, button, 1, 0, 1, 1);
-  g_signal_connect(G_OBJECT(button), "clicked", G_CALLBACK(apply_button_clicked), (gpointer)self);
+  gtk_grid_attach(grid, button, 5, 0, 5, 1);
```
TurboGit: 5 & 5 replaced by 3 & 3

src/libs/metadata.c
```diff
@@ -324,100 +459,243 @@ void connect_key_accels(dt_lib_module_t *self)
...
+  g_signal_connect(G_OBJECT(button), "clicked", G_CALLBACK(apply_button_clicked), self);
+
+  button = dtgtk_button_new(dtgtk_cairo_paint_preferences,
+      CPF_DO_NOT_USE_BORDER | CPF_STYLE_BOX | CPF_BG_TRANSPARENT, NULL);
+  d->config_button = button;
+  gtk_widget_set_tooltip_text(button, _("configure metadata"));
+  gtk_grid_attach(grid, button, 10, 0, 1, 1);
```
TurboGit: 10 replaced by 6.

And the we can have a small panel.

phweyland: I did this to reduce the space used by the configuration button. How can keep the configuration space as reduced as possible ?

TurboGit: Maybe by using the same kind of code we have in the export module ?

src/libs/metadata.c
phweyland: I have forgotten this.Must be removed. It's now done by ctrl-wheel.

src/libs/metadata.c
phweyland: in export module boxes are used instead of grid.
With you data I get this:
image

phweyland: got it:
image

phweyland: and no need 3 & 3 neither. 1 & 1 works.

phweyland: Thanks for challenging ! :)

phweyland: image

phweyland: image

TurboGit:
Found an issue, the presets CC-BY-* are not updated. Using them makes dt just delete them from the presets.

phweyland:
> Found an issue, the presets CC-BY-* are not updated. Using them makes dt just delete them from the presets.

Right ... I've forgotten to update legacy_params() and init_presets() when adding "notes".
Should work now (need to restart from V1).

TurboGit:
Thanks ! Merged.

GLLM:
A little late @phweyland for the work ! :)

One question, using LUA, I'm trying to fill the Notes (DT internal field, I know) with some Maker notes I retrieve from the file using exiftool 
 image.notes = film_simulation .. saturation

... but I get an error "(...)lua/FujifilmSimulation.lua:84: field "notes" can't be written for type dt_lua_image_t"

Am I doing something wrong ? Are we missing an accessor for this field ?

Thank you
GLLM

phweyland:
Hi @GLLM, from exiv2 standpoint the tag corresponding to dt "notes" is named "Xmp.acdsee.notes".
Maybe with exiftool the naming is probably slightly different ..., but "image.something" sounds for me more related to exif info than xmp ones...

GLLM:
Ohhh, I'll look in this direction. Thanks for the hint :-)

I'm browsing a selection of files (from a dt.gui.action_images), image per image, and try to add a string into this newly added metadata field named Notes .

```
 for _, image in ipairs(images) do
	    image.Xmp.notes = "a random string"
 end
```

Tried to set
* image.notes
* image.Xmp.acdsee.notes
* image.acdsee.notes
To no avail !

I understand from your message that rather than trying to browse/write the Xmp.
* Xmp.notes
* Xmp.acdsee.notes
I'll try to see how to get access to the Xmp, based on the ID i get from my dt.gui.action_images

Thanks
GLLM

phweyland:
I may be mistaken.
I've understood you wanted to import images with the metadata `notes ` fulfilled with data from another source. You use exiftool to set up properly this info.
If that's correct, you can set up the xmp.acdsee.notes tags with your info in the image before importing it (with exiftool or exiv2) and you should be done.

If you want, in a lua plugin fulfill the metadata notes, that's another story... and there is a long time I' haven't used lua.
If you are able to set `title` ("Xmp.dc.title") you should be able to do the same with `notes` ("Xmp.acdsee.notes").

GLLM:
No worries.
indeed, with exiftool, I get an info from different fields that aggregate and want to store in darktable Metadata > Notes field.
I'm indeed not sure how to access to this field.

I do manage to successfully read image.creator as well as image.description.
I've manually filled the Notes for this photo, and I cannot manage to even retrieve the info in LUA.

phweyland:
In lua code we have:
```
  lua_pushcfunction(L, creator_member);
  dt_lua_type_register(L, dt_lua_image_t, "creator");
  lua_pushcfunction(L, publisher_member);
  dt_lua_type_register(L, dt_lua_image_t, "publisher");
  lua_pushcfunction(L, title_member);
  dt_lua_type_register(L, dt_lua_image_t, "title");
  lua_pushcfunction(L, description_member);
  dt_lua_type_register(L, dt_lua_image_t, "description");
  lua_pushcfunction(L, rights_member);
  dt_lua_type_register(L, dt_lua_image_t, "rights");
```
the counter part for "notes" doesn't exist.
I think your issue is there. But that's a general lua issue, because when we add some new metadata in dt, somebody has to modify the lua code The lua code is not (yet) generic ...

GLLM:
Haha, I just remove the comment I was adding, I finally found this indeed :)

* The section dedicated to Read only is missing a Notes entry
image

PLUS there is not WRITE access to those fields, let apart the Notes node.

I think I'll stop digging here !
Thanks for your help anyway !

PR #4197, opened 2020-01-28
Another step to modern cmake

cryptomilk:
If you want to learn more:

https://cliutils.gitlab.io/modern-cmake/


parafin:
Last commit "cmake: Don't use $ORIGIN by default on Linux" should be removed.

cryptomilk:
@parafin: You have to explain why ...

parafin:
As far as I understand CMAKE_BUILD_RPATH_USE_ORIGIN doesn't affect rpath of installed binaries, only the ones in build directory. So you are breaking relocation ability. Do not remove things that were added for a reason.

cryptomilk:
Fixed it.

parafin:
Not really. CMAKE_INSTALL_RPATH is now incorrect on macOS. Also it is incorrect basically for darktable plugins. Also also - why make it optional? I ask again not to touch things you don't fully understand.

cryptomilk:
I'm looking at this from a distribution packager standpoint and you're creating something which should be done for packaging.

If you want to have a zip file which binaries you can move around, then $ORIGIN is fine, but not for package management!

Why is it wrong on MacOSX now? Looking at it, it was set to `@loader_path` as it was before on APPLE ...

parafin:
rpath was set to "${RPATH_DT}/../${CMAKE_INSTALL_LIBDIR}/darktable", which never evaluates to just "@loader_path". Please pay closer attention.

What's wrong with $ORIGIN from package management point of view? It evaluates to the same exact location as absolute path if you don't move the files.

cryptomilk:
Ah, I see, the older code overwrote RPATH_DT with $ORIGIN... mostly. Fixed ...

cryptomilk:
Apple should be back to as before, on Linux you need to turn it on if needed. It shouldn't by default.

parafin:
This is wrong again and I don't have time to explain again. I strongly against merging this PR unless it is explained why this is needed and someone tests it on macOS.

cryptomilk:
There should be no change on MacOSX but yes, someone should test it there ...

parafin:
You removed all mentions of @loader_path, how is this no change? And note that rpath was set to specific targets before, not globally. Relative path is not correct for all locations. I get that you care only about the case when relocation is off (you still haven't explained why it's a good thing), but you seem not to understand how cmake, linking, rpath and $ORIGIN come together.

Do not merge this PR with commit about $ORIGIN.

cryptomilk:
@parafin Could you please test on MacOSX. I'm happy to help fixing it should something not work.

parafin:
We are going in circles, you have reverted to a previous wrong solution, I have already explained how it is wrong and you still haven't answered why this is needed at all.

cryptomilk:
@parafin:

Is `@loader_path` the same as `$ORIGIN` on Linux and BSD? Is the directory structure the same on MacOSX? If yes, I think we want `@loader_path/../lib/`

Is there **ldd** on MacOSX? If yes, can you show `ldd /path/to/bin/darktable` which gives me some hints how this work on OSX.

cryptomilk:
@parafin could you please test with MacOSX?

parafin:
1. You haven't addressed the fact that now rpath is set for all targets, which is incorrect.
2. You haven't explained why this is needed at all.

I don't see why I would spend my time on testing, until these 2 questions are addressed, even if I had the time right now, which I don't.

cryptomilk:
I've changed it to set it on the targets.

On Linux distributions you should not create relocateable binaries with $ORIGIN but use the correct RPATH on installation. The latest commit addresses it. It keeps it as is with Apple.

parafin:
Why it shouldn't be used? I suppose this is a rule of some distributions, but what was the reason for introducing it? What harm can it do?

cryptomilk:
I didn't remove it. There is an option to turn it on e.g. for AppImage, Snap or whatever they are called.

`cmake -DBUILD_RELOCATABLE_PACKGAGE=ON ..`

parafin:
You've added extra functionality, so extra code to maintain. Why should we merge this? I'm really tired of repeating the same question.

cryptomilk:
What do you mean by extra code to maintain. If you look at the code, it keeps all stuff about RPATH in one place removing code deduplication. It fixes RPATH for Linux distributions which want absolute path if RPATH is used.

parafin:
Extra code to maintain means that you are adding one more build option, that can break, one more thing to test if one modifies something.
Why can't you just answer the question? What is the point of non-relocatable build? What's so secret about it?

cryptomilk:
https://docs.fedoraproject.org/en-US/packaging-guidelines/#_beware_of_rpath

Red Hat has tools checking that $ORIGIN is not used, probably due to security concerns. An absolute path is more secure than a relative one. However I don't find an up to date doc which states that. I will ask for an update.

parafin:
Yeah, no mention of $ORIGIN in that doc. I get that using absolute paths seems more secure, but I fail to imagine the scenario where it is actually so.

cryptomilk:
In the world of exploits a lot of people can't image that some code can be exploited and then they are proven wrong. So the rule is use absolute path to be sure, even for non security relevant applications like darktable.

parafin:
Still, how exactly using absolute path helps? $ORIGIN evaluates to the same path, unless binary is moved.

cryptomilk:
According to my colleagues, problems with $ORIGIN pop up every now and then. For example:

https://sourceware.org/bugzilla/show_bug.cgi?id=10253

It has limitations for AT_SECURE programs but should be OK for darktable.

parafin:
Glad we've sorted that out. Without BUILD_RELOCATABLE_PACKAGE option this PR now looks correct to me.

cryptomilk:
@TurboGit ready to go ;-)

TurboGit:
Thanks, will merge.

PR #4206, opened 2020-01-31
Makes the folder ordering based on film_id.

TurboGit:
This makes sure the recent folders are listed first.

The previous order was based on the folder name and case sensitive
which makes finding the right folder challending sometimes.


TurboGit:
This has always been itching me... What do you think? Am I the only one?

Nilvus:
I agree, good idea to fix that. Thanks @TurboGit

TurboGit:
Ok, merged on master and 3.0.1. Thanks.

Issue #4056, opened 2020-01-08
Darktable 3.0 crash (Access Violation) what importing edits made in Linux into Windows.


atomicwrites:
Hello, I usually use my Linux desktop at home to edit pictures, but once in a while I use a windows laptop. I only noticed this after the 3.0 update but it may be a coincidence since i don't use Darktable on Windows often. I got halfway through editing a film roll at home and copied the folder to my laptop, but on import I got many unhandled exception message boxes, and going to the path it says for the log there are a couple 1KB files that just say:

```
this is darktable 3.0.0 reporting an exception:
```

and  one actual log, for example:

<details>
<summary>
Click to expand rest.

```
this is darktable 3.0.0 reporting an exception:

-------------------

Error occurred on Tuesday, January 7, 2020 at 18:08:19.

darktable.exe caused an Access Violation at location 0000000064942EE3 in module libwinpthread-1.dll Writing to location 0000000066384EE2.

AddrPC           Params
0000000064942EE3 00000000055CE360 00000000638705A0 0000000000000000  libwinpthread-1.dll!pthread_mutex_lock
000000006370D22A 0000000000000003 0000000063870580 00000000638705A0  libdarktable.dll!dt_dev_pixelpipe_cleanup  [D:/build/darktable/src/common/dtpthread.h @ 307]
000000006362FCAF 000000000000001C 00000000638757FC 00000000055CF720  libdarktable.dll!dt_imageio_export_with_flags  [D:/build/darktable/src/common/imageio.c @ 947]
000000006366065C 0000000000000000 00000000055CFD10 00000000019A5888  libdarktable.dll! ??   [D:/build/darktable/src/common/mipmap_cache.c @ 1255]
000000006368F7B4 FFFFFFFFFFFFFFFF 000000000E172110 000000000E172138  libdarktable.dll!dt_image_load_job_run  [D:/build/darktable/src/control/jobs/image_jobs.c @ 35]
0000000063687BA7 0000000000000000 0000000000000000 0000000000000000  libdarktable.dll!dt_control_job_execute  [D:/build/darktable/src/control/jobs.c @ 304]
00000000636880E8 00000000047A6FF0 0000000000000000 0000000000000000  libdarktable.dll!dt_control_work  [D:/build/darktable/src/control/jobs.c @ 323]
0000000064945092 000000000479D3E0 0000000000000000 0000000000000000  libwinpthread-1.dll!pthread_create_wrapper
[...]

Issue #4043, opened 2020-01-06
Error message with error popup"failed to allocate memory for thumbnail export, please lower the threads used for export or buy more memory"


ydyote:
**Describe the bug**
I want to import a folder recursively in lighttable and if the number of images is too high it will crash with message "failed to allocate memory for thumbnail export, please lower the threads used for export or buy more memory".

I'm not even sure to which memory is this message referring to graphic card memory or standard RAM?

Also it spawns a small window, in which when I click OK it will close darktable.

If you don't click the button and just wait, the import will finish.

(maybe this is a gpu issue, because I think it happens when the processing load is too high)

**To Reproduce**
1. Have ~10K images in multiple folders under 1 folder named "images".
2. Go to import -> folder
3. Check import recursively.
4. Select the "images" folder and click open.
5. Wait for a second and it should spawn the error window.

**Expected behavior**
I would expect at least clear steps to fix this on my side, if it is specific to my installation and PC.

**Screenshots + Files**

![image](https://user-images.githubusercontent.com/54998537/71813154-44b35180-3079-11ea-9af7-56eb2534f74d.png)
![image](https://user-images.githubusercontent.com/54998537/71813164-4a109c00-3079-11ea-8b65-ebcbf4620e0c.png)
[darktable_bt_THRVD0.txt](https://github.com/darktable-org/darktable/files/4025737/darktable_bt_THRVD0.txt)

**Platform (please complete the following information):**
 - Darktable Version: 3.0.0
 - OS: Windows 10 Pro v1809 build 17763.914
 - OpenCL activated, very fast GPU profile

CPU:           AMD Phenom II X4 955 Black Edition (Deneb, RB-C3 (Ridgeback))
               3700 MHz (18.50x200.0) @ 3709 MHz (18.50x200.5)
Motherboard:   MSI 770-C45 (MS-7599)
Chipset:       AMD 770/870 (RX780) + SB750/SB710
Memory:        16384 MBytes @ 668 MHz, 9-9-9-24
               - 4096 MB PC10600 DDR3 SDRAM - Kingston 9905471-006.A02LF
               - 4096 MB PC10600 DDR3 SDRAM - Kingston 9905471-006.A02LF
               - 4096 MB PC10600 DDR3 SDRAM - Kingston 99U5474-028.A00LF
               - 4096 MB PC10600 DDR3 SDRAM - Kingston 99U5474-028.A00LF
Graphics:      MSI N750Ti OC (MS-V8A9)
[...]

Issue #3298, opened 2019-11-05
loading huge EXR image crashes


claudeha:
**Describe the bug**
`darktable 32000x18000.exr` → segmentation fault, and `darktable` can no longer be started from that directory without crashing

**To Reproduce**
1. Create an EXR file with very large image size (I tried 32000x18000 which crashes; 16000x9000 loads ok)
2. Try to load it in `darktable`
3. Crash

**Expected behavior**
Ideally, for the image to be loaded successfully.
Pragmatically, an error message saying that the dimensions are too big, without a crash.

**Platform (please complete the following information):**
 - OS: Debian Buster
 - Version: git master at `release-3.0.0rc0-22-gc03143cb9`

**Additional context**
My EXR files have R G B half channels.  They contain computer generated fractal imagery generated with https://mathr.co.uk/kf/kf.html whose output tiles were assembled into a large image using https://mathr.co.uk/exrtact

**Crash log**
[darktable_bt_72GVA0.txt](https://github.com/darktable-org/darktable/files/3811088/darktable_bt_72GVA0.txt)

claudeha:
master branch at release-3.1.0-426-g88bf11bd2 still crashes
[darktable_bt_YULOE0.txt](https://github.com/darktable-org/darktable/files/4106393/darktable_bt_YULOE0.txt)

johnny-bit:
Since #4210 is merged in master can you re-check?

claudeha:
Still crashes with master branch at release-3.5.0-139-gca5a21527.

TL;DR 32bit signed int overflow at `darktable/src/common/imageio_exr.cc:132` gives a very huge size in `memset(buf, 0, 4 * img->width * img->height * sizeof(float));`  ($8 is the size it should be)

```
Thread 1 "darktable" received signal SIGSEGV, Segmentation fault.
__memset_avx2_unaligned_erms () at ../sysdeps/x86_64/multiarch/memset-vec-unaligned-erms.S:202
202	../sysdeps/x86_64/multiarch/memset-vec-unaligned-erms.S: No such file or directory.
(gdb) bt
#0  __memset_avx2_unaligned_erms () at ../sysdeps/x86_64/multiarch/memset-vec-unaligned-erms.S:202
#1  0x00007ffff7e49cf4 in dt_imageio_open_exr(dt_image_t*, char const*, dt_mipmap_buffer_t*)
[...]

PR #4210, opened 2020-02-01
Fixing pixelpipe cache data buffers if allocation goes wrong.

jenshannoschwalm:
There are the issues #4056 and #4043. From my understanding both are related to
the error handling in `dt_dev_pixelpipe_cache_init` if allocation for one of the
cache->data buffers goes wrong.

In such an error case we should not call `dt_dev_pixelpipe_cache_cleanup` but
only reset each cache->size and cache->data to 0/NULL so the pipeline data are
still valid (so avoiding the reported crash) and the report about the memory-stress
is still reported.

Unfortunately i don't have a windows machine to test but i tried with

```
cache->data[k] = (void *)dt_alloc_align(64, size*100);
if(!cache->data[k]) goto alloc_memory_fail;
```
and very-very large images and this pr fixed the crashes i could also observe
under such memory-stress.

### DIFF
--- a/src/develop/pixelpipe_cache.c
+++ b/src/develop/pixelpipe_cache.c
@@ -63,7 +63,17 @@ int dt_dev_pixelpipe_cache_init(dt_dev_pixelpipe_cache_t *cache, int entries, si
   return 1;
 
 alloc_memory_fail:
-  dt_dev_pixelpipe_cache_cleanup(cache);
+  //  dt_dev_pixelpipe_cache_cleanup(cache);
+  // The above code seems to be not correct as failing to allocate the cache->data buffers
+  // should not cleanup the whole pixelpipe cache but only reset the buffers to null.
+  // A warning about low memory will appear but the pipeline still has valid data so dt won't crash
+  // but will only fail to generate thumbnails for example.
+  for(int k = 0; k < cache->entries; k++)
+  {
+    dt_free_align(cache->data[k]);
+    cache->size[k] = 0;
+    cache->data[k] = NULL;
+  }
   return 0;
 }
 


src/develop/pixelpipe_cache.c
```diff
@@ -63,7 +63,17 @@ int dt_dev_pixelpipe_cache_init(dt_dev_pixelpipe_cache_t *cache, int entries, si
...
+  // The above code seems to be not correct as failing to allocate the cache->data buffers
+  // should not cleanup the whole pixelpipe cache but only reset the buffers to null.
+  // A warning about low memory will appear but the pipeline still has valid data so dt won't crash
+  // but will only fail to generate thumbnails for example.
+  for(int k = 0; k < cache->entries; k++)
+  {
+    dt_free_align(cache->data[k]);
+    cache->size[k] = 0;
+    cache->data[k] = NULL;
+  }
```
TurboGit: But aren't we then here leaking memory for:

```
  cache->dsc
  cache->hash
  cache->used
```

jenshannoschwalm: That's what the old code implied. But that's exactly what i think is wrong.

The allocated memory for `cache->dsc` and friends is deallocated later via
`void dt_dev_pixelpipe_cleanup(dt_dev_pixelpipe_t *pipe)` found in pixelpipe_hb.c.

If we free everything here via `dt_dev_pixelpipe_cache_cleanup` we later have
1. access in the pipeline code to the already deallocated pointers resulting in access violations
2. double freeing in `dt_dev_pixelpipe_cleanup`

The paired cache_init and cache_cleanup has been checked by adding

` fprintf(stderr,"\ndt_dev_pixelpipe_cache_init %p",cache);` and `fprintf(stderr,"\ndt_dev_pixelpipe_cache_cleanup %p",cache);`

jenshannoschwalm:
BTW #3298 seems to be related too.

TurboGit:
I see, I'm no expert but this seems indeed a good fix. Thanks.

TurboGit:
Merged in master and 3.0.1.

PR #4212, opened 2020-02-01
fix:filmicrgb: fixed rgb power norm for dark areas

burrima:
The denominator in the function pixel_rgb_norm_power() was limited
to a minimum value of 1e-6 (which is about 20 EV stops below 1.0).

But the denominator is the square-sum of R, G and B and thus the norm
was computed wrongly for values at about 10 EV stops (and lower) below
1.0. The minimum denominator value should be (1e-6)^2 = 1e-12.

For the computation of the norm, such a small value should be fine
since the only purpose of this minimum limit is that there is no
division by 0. The norm is specified to be in ]0; +infinity[.

### DIFF
--- a/data/kernels/filmic.cl
+++ b/data/kernels/filmic.cl
@@ -209,7 +209,7 @@ inline float pixel_rgb_norm_power(const float4 pixel)
   const float4 RGB = fabs(pixel);
   const float4 RGB_square = RGB * RGB;
   const float4 RGB_cubic = RGB_square * RGB;
-  return (RGB_cubic.x + RGB_cubic.y + RGB_cubic.z) / fmax(RGB_square.x + RGB_square.y + RGB_square.z, 1e-6f);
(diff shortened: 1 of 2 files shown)


burrima:
Done, thank you for the hint! Just for curiosity: is the opencl kernel auto-generated or hand-written?

aurelienpierre:
> Just for curiosity: is the opencl kernel auto-generated or hand-written?

Fully hand-written ! That's why maintaining all that is cumbersome : each module has the same code written 2 to 3 times (plain C, SSE2, OpenCL), so it's very easy to have inconsistencies.

burrima:
I see. That makes it also hard to test. Unit-testing the C-code is a start but then there should be a mechanism that verifies that SSE2 and OpenCL behave exactly the same - ideally function block by function block.

TurboGit:
Merging then ! Thanks.

PR #4214, opened 2020-02-02
Fix for multiple lines fo the same lighttable module shortcut

Mark-64:
This is a fix for a bug causing multiple entries of the same shortcut in the shortcut list for lighttable modules.

Switching back and forth between lighttable and darkroom made the shortcut list grow continuously.
If Lua module was registered, the list was doubling.

Screenshot

### DIFF
--- a/src/gui/accelerators.c
+++ b/src/gui/accelerators.c
@@ -185,7 +185,8 @@ void dt_accel_register_lib_for_views(dt_lib_module_t *self, dt_view_type_flags_t
   accel->local = FALSE;
   // we get the views in which the lib will be displayed
   accel->views = views;
-  darktable.control->accelerator_list = g_slist_prepend(darktable.control->accelerator_list, accel);
+  if (!dt_accel_find_by_path(accel->path))
+    darktable.control->accelerator_list = g_slist_prepend(darktable.control->accelerator_list, accel);
 }
 void dt_accel_register_lib(dt_lib_module_t *self, const gchar *path, guint accel_key, GdkModifierType mods)
 {
--- a/src/gui/accelerators.c
+++ b/src/gui/accelerators.c
@@ -173,9 +173,11 @@ void dt_accel_register_lib_for_views(dt_lib_module_t *self, dt_view_type_flags_t
                                      guint accel_key, GdkModifierType mods)
 {
   gchar accel_path[256];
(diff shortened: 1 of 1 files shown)


src/gui/accelerators.c
```diff
@@ -185,7 +185,8 @@ void dt_accel_register_lib_for_views(dt_lib_module_t *self, dt_view_type_flags_t
   accel->local = FALSE;
   // we get the views in which the lib will be displayed
   accel->views = views;
-  darktable.control->accelerator_list = g_slist_prepend(darktable.control->accelerator_list, accel);
+  if (!dt_accel_find_by_path(accel->path))
+    darktable.control->accelerator_list = g_slist_prepend(darktable.control->accelerator_list, accel);
```
AlicVB: we need to free accel (or better make the test before the allocation)

Mark-64: Checking before allocation as you suggested.
the function is called multiple times, not only when initializing accels, but also when switching views (which make the list grow continuously) and when registering lua modules.
So perhaps there is a better way to fix at higher level, but I don't want to break other things

Mark-64:
In the last commit I put checking before allocation as you suggested.
The function is called multiple times, not only when initializing accels, but also when switching views (which make the list grow continuously) and when registering lua modules.
So perhaps there is a better way to fix at higher level and avoid that the function is even called, but I don't want to break other things

Mark-64:
@AlicVB 
I will continue to dig into the code to see if there are other potential adverse effects of the bug which my patch doesn't cure.  At the moment I don't see any.
As I optimized the patch as you requested, I suggest to accept this PR in the meantime, it's just one line and it can be easily rolled back if a better solution is found.

AlicVB:
Ok, I've found the source of the problem : it's presets !
here : https://github.com/darktable-org/darktable/blob/1fd728123f345588c2d34df757218f6e970a52f7/src/libs/lib.c#L1144
we init _and_ connect each accels, and this called each time you enter a view with the module...
Ideally we should split that in 2 : initialization of presets accels in the module init routine, and connection like for instance !

But well, I don't have the time atm to change and test that, so either someone (you ?) is brave enough to do it, either we go with your current solution (which work but add some useless check)

Mark-64:
Yeah, I found that, but my first attempt to split the function didn't work.
I will try again

TurboGit:
Thanks ! Merging in master and 3.0.1.

Issue #4172, opened 2020-01-22
severe performance degradation 


ptilopteri:
openSUSE Tumbleweed 20200118
NVIDIA GF106 [GeForce GTS 450], 390.129
darktable 3.1.0~git422.092f7408a
OpenCL on 
i7 12-core 36GB

I began with a new library and gradually imported images.  As the library grows, so does the delay between activating the cursor on the screen (click hold mouse button) and when the cursor screen position is displayed inside the crop box/area, ie: 3205,2125.  Attempting to drag the crop box/area before the cursor position appears results in no movement or delayed movement.

I see the same delay when attempting to increase or decrease sharpness amount

And again, as the library grows, the delay increases.  Is this perhaps related to recording the history as usually a single <ctrl><z> does not work and sometimes requires repeating 3 or 4 times.

And I do not build but use darix's master OBS builds.

And my present schedule provides ample time for testing.

ptilopteri:
My current library contains >42,000 images

Nilvus:
Could you precise which kernel you use on your Linux system ? I have had some severe performance degradation solved by just upgrade my 5.3.0 kernel (on Ubuntu 19.10) by upgrading to a 5.4.11 kernel (I now use a 5.4.13 with same good performances as the 5.4.11).
And which settings do you have in core options, CPU/GPU/Memory parameters on darktable preferences window ?

ptilopteri:
kernel-default-5.4.7-1.1.x86_64
memory in megabytes to use for thumbnail cache:  512
number of background threads:  8
host memory limit (in MB) for tilling:  18040
minimum amount of memory (in MB) for a single buffer in tilling:  16

ptilopteri:
fwiw:  I installed darktable-3.0.0 and a library from backup dated 2019-1130 an observe normal operation, ie: no delays on editing operations.

aurelienpierre:
Ah, so I'm not the only one. Same here, GUI events seem very slow.

ptilopteri:
I only see the slow-down or delays when editing and only editing actions, not changing images or changing from darkroom to lighttable or back or increasing/decreasing image display size.

But the delays make "master" un-usable.

ptilopteri:
I can run a debug session of dt while operating in an edit session where I experience the delays if that would help.

MStraeten:
i found having a lot of drawn masks (e.g. used for retouch) results in a massive slowdown. Log file is full f processing masks ...

[...]

PR #4222, opened 2020-02-03
tagging - set tag list view as default 

phweyland:
doesn't solve but helps to avoid situation as issue #4172 .
Set tag list view as default instead of suggestions view, which accesses db and then is slower.


TurboGit:
Looks ok. Together with #4224 we should get rid of all slow down.

phweyland:
> Together with #4224 we should get rid of all slow down.

The indexes on memory tables should improve performance, sure, but I haven't understood the root cause of the issue.
As far as PR #4112 is involved, the issue should be related to main.tagged_images...
It has already a double primary key (imgid, tagid) plus an index (tagid). Let's see :-).

TurboGit:
> The indexes on memory tables should improve performance, sure, but I haven't understood the root cause of the issue.

Not only see #4224, there was an in memory table set but not used and the SQL was quite slow on this one.

Issue #4021, opened 2020-01-04
[mask refinement] feathering radius weird artifacts


AxelG-DE:
**Describe the bug**
When mask opacity has been set to 1,00 in a drawn mask, feathering radius creates strange lines, and they are actually in the mask than

I know it makes no sense in a drawn mask only, but drawn and parametric, it does, typically L-Mask). Just the error also appears with drawn only, so we can concentrate on that.

**To Reproduce**
1. Go to any module, e.g. exposer
2. create a drawn mask
3. change opacity to 1,00
4. change feathering  radius to e.g. 2,4

**Expected behavior**
Mask should be handled propperly, as it is, when mask opacity is not changed

**Screenshots**
![Screenshot_20200104_142610](https://user-images.githubusercontent.com/49004103/71766347-5a047080-2eff-11ea-9e86-d2d811403e1c.png)

**Platform (please complete the following information):**
 - Darktable Version: **3.0.0** and **current master**
 - OS: **Gentoo Linux**
 - OpenCL activated or no? **on and off**
 - Which graphics card and driver version **2x nvidia GTX1060; nvidia-drivers 440.44-r1**

**Additional context**
 - Can you reproduce with another Darktable version? **Yes**
 - Can you reproduce with a RAW or Jpeg or both? **RAW**
 - Are the steps above reproduce with a fresh edit (removing history)? **Yes**
 - Attach an XMP if this is necessary 
[2019-12-14_124134__EM50100.ORF.xmp.tar.gz](https://github.com/darktable-org/darktable/files/4021953/2019-12-14_124134__EM50100.ORF.xmp.tar.gz)
 - Did you compile Darktable yourself? If so which compiler was used, with what options? **x86_64-pc-linux-gnu-9.2.0 CMAKE_BUILD_TYPE="Release" CFLAGS="-march=native -O3 -mtune=native -pipe"**
 - Is the issue still present using an empty/new config-dir **Yes**

jade-nl:
Confirmed.

The order  in which the opacity and feathering are applied doesn't matter.

Played around a bit with the values:
- Opacity needs to be 1 or larger.
- Feathering needs to be .2 or larger.

Using:
- Debian 10.2
- darktable 3.1.0+291
- opencl off
- nVidia GF104 [GeForce GTX 460]
- 2 random RAW's (different camera's)

PR #4226, opened 2020-02-04
Restricting masks brightness to < 1.00 (fixing #4021)

jenshannoschwalm:
As reported there are strong artefacts if the mask opacity (brightness)
in mask refinement has been set to 1.0

This pr does not touch history data or the database if this had been set to 1.0, it
just makes sure the brightness value is clipped to something less than 1.0 in
`dt_develop_blend_process` - also for cl path.

### DIFF
--- a/src/develop/blend.c
+++ b/src/develop/blend.c
@@ -3029,7 +3029,7 @@ void dt_develop_blend_process(struct dt_iop_module_t *self, struct dt_dev_pixelp
     if(mask_tone_curve && opacity > 1e-4f)
     {
       const float e = expf(3.f * d->contrast);
-      const float brightness = d->brightness;
+      const float brightness = fminf(d->brightness,0.999f);
 #ifdef _OPENMP
 #pragma omp parallel for default(none) \
       dt_omp_firstprivate(brightness, buffsize, e, mask, opacity)
@@ -3438,7 +3438,7 @@ int dt_develop_blend_process_cl(struct dt_iop_module_t *self, struct dt_dev_pixe
     if(mask_tone_curve && opacity > 1e-4f)
     {
       const float e = expf(3.f * d->contrast);
-      const float brightness = d->brightness;
+      const float brightness = fminf(d->brightness,0.999f);
       dt_opencl_set_kernel_arg(devid, kernel_mask_tone_curve, 0, sizeof(cl_mem), (void *)&dev_mask_2);
       dt_opencl_set_kernel_arg(devid, kernel_mask_tone_curve, 1, sizeof(cl_mem), (void *)&dev_mask_1);
       dt_opencl_set_kernel_arg(devid, kernel_mask_tone_curve, 2, sizeof(int), (void *)&owidth);


aurelienpierre:
I need to check what the masking algo does underneath.

rabauke:
Do we really need to enforce that users can not do unreasonable things, e.g. smoothening the mask via Gaussian or guided filter and counteracting this by pushing mask brightness such that the mask degenerates to a set of completely opaque or completely transparent values? 

I hardly understand why users complain about artifacts when pushing sliders to extremes rather than reducing them to more reasonable values. Thats a little bit like complaining about that an image becomes almost completely white when pushing exposure by several EV.

jenshannoschwalm:
> Do we really need to enforce .....

Mostly agreed. But restricting parameters to "being sensible" is not a bad thing, may it be only for less noise.

johnny-bit:
Why not limit only the sliders to "sane" value and let the input still be unbounded by sanity? I think that's what @aurelienpierre done in color balance module (and maybe others)

aurelienpierre:
The difference here is "sane" values depend on the combination of parameters, which is not really predictable. I don't remember who said that preventing users from making stupid things ends up making stupid software. I agree with Heiko here.

jenshannoschwalm:
> The difference here is "sane" values depend on the combination of parameters, which is not really predictable

I think this is not correct. Of course there are combinations that don't make sense at all. But a brightness of 1.0 is a corner case that is simply wrong. See https://github.com/darktable-org/darktable/blob/1f046ed717380443be36af1ba6d3f9cd80433102/src/develop/blend.c#L3029-L3059

rabauke:
@jenshannoschwalm : Brightness equal to +1 or -1 is an edge case, indeed. The mask transformation becomes a Heaviside step function. This is why, the code shown above takes special care of this case in the first two if-cases.

jenshannoschwalm:
@rabauke Exactly :-) If you set it to 1.0 and check the code sections as

`x = (x + brightness) / (1.f - brightness);`

you will see that exactly these corner cases 3/4 are executed. (try after disabling OPENMP)
Why don't they crash? I think this is due to "good luck" or better floating peculiarities.

rabauke:
@jenshannoschwalm : For `brightness == 1`, the code fragment that you cite will never be reached, because of the 

     if (1.f - brightness <= 0.f)
              x = mask[k] <= FLT_EPSILON ? -1.f : 1.f;

a few lines before. Thus there is no reason to enforce absolute brightness values less than one.

One might, however, consider to replace

        if (1.f - brightness <= 0.f)
          x = mask[k] <= FLT_EPSILON ? -1.f : 1.f;
        else if (1.f + brightness <= 0.f)
          x = mask[k] >= 1.f - FLT_EPSILON ? 1.f : -1.f;

by 

        if (1.f - brightness <= 0.f)
          x = mask[k] <= epsilon ? -1.f : 1.f;
        else if (1.f + brightness <= 0.f)
          x = mask[k] >= 1.f - epsilon ? 1.f : -1.f;

where `epsilon` is a suitable small positive numerical value (larger than the current `FLT_EPSILON`). This should avoid the artifacts that you might have seen for a non-zero _feathering radius_ and unity _mask opacity_.

jenshannoschwalm:
Hi @rabauke i agree with you that your pr fixes the mentioned issue in a more appropriate / better way :-)

PR #4229, opened 2020-02-04
iop_order: add const and restrict qualifiers on pointers

aurelienpierre:


parafin:
The usage of restrict looks very icky to me. Are you sure you're using it correctly? E.g. does it make sense to declare pointer both const and restrict? Does it really help with performance to risk potential strange bugs due to compiler optimizations?

aurelienpierre:
const declare immutable pointers, and restrict declare non-aliasing pointers (telling the compiler there is no need to reload the object from memory). Since GLists are chained lists read in sequence without concurrent access, I don't think pointers ever overlap, so all in all, you need only one copy and one store by Glist handling. That's my understanding, I may be wrong.

parafin:
Yeah, I guess `const * restrict` does make sense. Still, the question remains - isn't it a premature optimization? I'm all for const modifiers, because they can't hurt, but restrict is quite obscure language feature. How often are those functions called that such optimization was considered necessary? It's not very obvious that code is correct, since these restrict pointers are copied between variables and passed to functions. I would really think about removing them unless the performance increase is measurable and needed at all here.

src/common/iop_order.h
```diff
@@ -158,83 +158,83 @@ typedef struct dt_iop_order_rule_t
...
 
 // for debug only
-int dt_ioppr_check_iop_order(struct dt_develop_t *dev, const int imgid, const char *msg);
-void dt_ioppr_print_module_iop_order(GList *iop_list, const char *msg);
-void dt_ioppr_print_history_iop_order(GList *history_list, const char *msg);
-void dt_ioppr_print_iop_order(GList *iop_order_list, const char *msg);
+int dt_ioppr_check_iop_order(struct dt_develop_t *const dev, const int imgid, const char *msg);
+void dt_ioppr_print_module_iop_order(GList *const iop_list, const char *msg);
+void dt_ioppr_print_history_iop_order(GList *const history_list, const char *msg);
+void dt_ioppr_print_iop_order(GList *const iop_order_list, const char *msg);
```
rabauke: Const in function _declarations_ is redundant and can be omitted (const to declare the variable or pointer itself as const, not to indicate a pointer that points to something constant). This means, it is absolute valid to declare 

`int dt_ioppr_check_iop_order(struct dt_develop_t *dev, int imgid, const char *msg);`

and to implement

`int dt_ioppr_check_iop_order(struct dt_develop_t *const dev, int imgid, const char *msg) { ... };`

aurelienpierre: Ok, thanks for the tips. It's really stupid that `f(const *x)` needs to go in the header, but `f(*const x)` is as good as nothing. Same with `restrict`. Such an inconsistent language… What about `__attribute__((aligned(64)))` ?

aurelienpierre:
@rabauke not sure how hot is that spot, but there are lots of checkings so I would rather make it efficient.

src/common/iop_order.h
parafin: I think you didn't get the point. *const and restrict does nothing in the header, because it matters only for implementation, so you can use them in definition. In declaration they just don't mean anything and are just noise.
Not sure what __attribute__((aligned(64))) has to do with anything. It makes no sense in function prototype.

parafin:
I really doubt it makes a difference here performance-wise and `restrict` keywords decrease SNR of the code and may confuse people reading (or worse - modifying) the code.

src/common/iop_order.c
```diff
@@ -663,14 +663,14 @@ static void _ioppr_reset_iop_order(GList *iop_order_list)
...
-void dt_ioppr_resync_iop_list(dt_develop_t *dev)
+void dt_ioppr_resync_iop_list(dt_develop_t *const dev)
 {
   // make sure that the iop_order_list does not contains possibly removed modules
 
   GList *l = g_list_first(dev->iop_order_list);
   while(l)
   {
-    GList *next = g_list_next(l);
+    GList *const restrict next = g_list_next(l);
```
parafin: This makes no difference,  because nothing is accessed through that pointer, so you're just adding noise. Also if something was accessed through it, it would be wrong, because list is later modified through another pointer (`dev->iop_order_list`).
Plus, not part of this patch, but 6 lines below there's a memory leak - g_list_remove_link doesn't free the removed element of the list.

src/common/iop_order.c
```diff
@@ -682,17 +682,17 @@ void dt_ioppr_resync_iop_list(dt_develop_t *dev)
...
   // and reset all module iop_order
 
-  GList *modules = g_list_first(dev->iop);
+  GList *restrict modules = g_list_first(dev->iop);
   while(modules)
   {
-    dt_iop_module_t *mod = (dt_iop_module_t *)(modules->data);
-    GList *next = g_list_next(modules);
+    dt_iop_module_t *const restrict mod = (dt_iop_module_t *)(modules->data);
+    GList *const restrict next = g_list_next(modules);
```
parafin: again, just noise

parafin:
I've looked over a couple of usages of restrict, and they are either wrong or don't make a difference. I really think they all should be removed, unless proven that there is a difference in speed (or at least generated code).

TurboGit:
I agree that restrict is maybe too restrictive and we don't have real evidence about speed. As this seems controversial I'll change this PR to have const only (very useful) and discuss the restrict one. But this is certainly more noise than something really helping.

rabauke:
I also had a second look at to code. I completely agree with what @TurboGit wrote. I like the additional `const` declarations, I cannot find any case where the `restrict` declaration makes sense to me. Usually there is no risk of aliasing anyway.

aurelienpierre:
Ok then, I will do a second pass on that tomorrow and keep only the const and maybe the *const restrict

PR #4230, opened 2020-02-04
po/speed up tags

TurboGit:


TurboGit:
Another large speed-up. At least the darkroom is very responsive now.

TurboGit:
@phweyland : Ok for you ?

src/common/tags.c
```diff
@@ -214,7 +214,7 @@ gboolean dt_tag_new(const char *name, guint *tagid)
     // tagid already exists.
     if(tagid != NULL) *tagid = sqlite3_column_int64(stmt, 0);
     sqlite3_finalize(stmt);
-    return TRUE;
+    return FALSE;
```
phweyland: This is not correct. Callers expect TRUE is tag exists or is created.

src/common/tags.h
```diff
@@ -78,11 +78,14 @@ void dt_tag_rename(const guint tagid, const gchar *new_tagname);
 gboolean dt_tag_exists(const char *name, guint *tagid);
 
 /** attach a tag on selected images. tagid id of tag to attach. imgid the image
- * id to attach tag to, if < 0 selected images are used. */
-void dt_tag_attach(const guint tagid, const gint imgid, const gboolean undo_on, const gboolean group_on);
+ * id to attach tag to, if < 0 selected images are used. Returns FALSE if tag was already present */
```
phweyland: Same as before. The requirement for the caller is to have an existing tag before attaching it.

phweyland:
>  At least the darkroom is very responsive now.

Great !

src/common/tags.c
phweyland: in tag.c line 248 dt_tag_new answer is used to raise signal.
But development doesn't create new tag. It only attaches "changed" tag. So this shouldn't impact the response time.

TurboGit: But what the point of raising a signal if the tag was already present ? The comment says:

`  /* if everything went fine, raise signal of tags change to refresh keywords module in GUI */`

But what to refresh is nothing has changed ?

phweyland: The fact is that the tag already exists is not an issue. 
I agree it is useless to refresh if nothing has changed. But this should not occur that often.
Only lua calls dt_tag_new_from_gui() (and makes the signal). This is not related to the slowness issue. Development calls dt_tag_new().
I think your other change, on attachment ("changed") will make the difference.

TurboGit: I agree this is not the area where I gain the speed, but yet this seems quite superfluous. I'll remove this and will merge.

TurboGit:
Merged.

phweyland:
You have given me a lesson of issue catching and fixing.
Do you use any tool to analyze execution time and this kind of things ? If yes, which tools ?
Thanks !

TurboGit:
@phweyland : No tool really, just using dt -d sql log output. Analyzing it and trying to see what could be missing in SQL to gain speed up.  In the case at hand it was strange to me that dragging any sliders on the darkroom could make so many SQL request each time. And this boiled down to adding the darktable|changed tag. From there it was easy to fix.

ptilopteri:
@TurboGit 
Was that part of the solution to #4172 ?

TurboGit:
@ptilopteri : Yes, that was part of it.

PR #4251, opened 2020-02-07
rgbcurve: fix presets that were copied from tonecurve Lab mode.

### LINKED ISSUE #4247: [rgb curve] Shipped camera presets are wrongly implemented.
**Describe the bug**

Selecting either of the shipped  presets ends up showing the *RGB, independent channels* (the red channel) and the image has a red colour.

Manually switching from *RGB, independent channels* to *RGB, linked channels* solves the problem.

Tested this with images made with a D750 and D7000, both have the same unwanted result.

EDIT: Downloaded and tested and confirmed with D90, D5100 and E-M10II.

**To Reproduce**

1. Choose and open an image made with a Nikon: D5100, D7000, D7200, D7500, D750, D90 or Oympus OM-D E-M10II (one of the currently shipped presets)
2. Select the rgb curve preset that goes with the camera (either from within the module or from *quick access*).

**Expected behavior**

Preset(s) to be implemented correctly as shown by third screenshot.

**Screenshots**

Freshly imported image:
rgbcurve original

After the preset is applied:
rgbcurve preset

After manually setting it to *RGB, linked channels:
rgbcurve manually

 - Darktable Version: Stable 3.0.0 + Development 3.1.0+578
 - OS: Linux
 - Distro: Debian Buster (10.2)

**Additional info**
 - Is the issue still present using an empty/new config-dir: Yes.
 - Clean or filled history: Makes no difference.

TurboGit:
Fixes #4247.


TurboGit:
Note that the rendering is not 100% equivalent because the rgbcurve works in RGB and not in Lab. So maybe the best solution is to remove those presets ?

TurboGit:
I really think that profiles made for Lab cannot be just translated here. Will remove the presets.

PR #4265, opened 2020-02-09
Serif and higher-contrast themes

ralfbrown:
Since I dislike sans-serif fonts and gray40-on-gray50 low-contrast text, I've created themes that 1) switch to serif fonts, 2) boost the UI contrast somewhat, and 3) make both changes.


rabauke:
@ralfbrown : There is a reason, why most graphical user interfaces are based on sans serif fonts. On low resolution displays, the rendering quality of sans serif fonts is usually better than the rendering quality of serif fonts. There are not many serif fonts that render well on low-resolution displays, especially when it comes small font sizes.

Even if we consider adding a theme with serif fonts, "Liberation Serif" and "FreeSerif" are probably not best choices. "Noto Serif" and "Roboto Serif" might be better alternatives to the fonts used in this pull request.

Nilvus:
> @ralfbrown : There is a reason, why most graphical user interfaces are based on sans serif fonts. On low resolution displays, the rendering quality of sans serif fonts is usually better than the rendering quality of serif fonts. There are not many serif fonts that render well on low-resolution displays, especially when it comes small font sizes.
> 
> Even if we consider adding a theme with serif fonts, "Liberation Serif" and "FreeSerif" are probably not best choices. "Noto Serif" and "Roboto Slab" might be better alternatives to the fonts used in this pull request.

I totally agree here. And we also have to be careful to not add to many themes by default. About these themes, why not test that with changes proposed by @rabauke here. I would also, if we validate the add by default to these themes, to rename them High display themes or something like to (ie refer in the name that they're more about High Dpi displays and not low ones) ; even if new elegant themes especially are low and high displays compatible ; there's just some issues on some specific OS, especially for systems with a Gtk+ older than 3.24.7 or some specific anti-aliasing options in KDE or Windows from what I have see by now).

I don't have high DPI screen to really test these themes.

We also by the past discussed the idea to propose a themes library (to let users choose themes and add them on their config folder).

ralfbrown:
I'm aware that sans serif is better at low resolutions, but for serif/sans I will posit that 1920x1080 is sufficiently high resolution except for smaller-than-default text (my monitor is 2560x1600, so it's by no means a true HiDPI display).  I have my entire desktop and all applications that can be configured using serif fonts.  When I made the switch, it was so nice to finally be able to distinguish Il| :-).

I'm fine with renaming the themes, and was in fact expecting to be asked.

I've installed Noto and Roboto Slab fonts, and found that Noto is about 10% wider than Liberation, and Roboto even a bit wider than Noto.  Both are also slightly taller, so overall it's about like increasing the font size by a point or so with my desktop set for 11pt default size.

What's the correct procedure for resynchronizing with darktable-org:master?  Should I use the sequence
   git checkout master
   git pull upstream master
   git checkout serif-theme
   git rebase master

Nilvus:
That's make too many themes to add for what you want. We have to be careful to not have too much themes proposed by default. For that, if this is added by default, I would limit that to 2 themes. Otherwise, we have to think of a way to propose some themes resources on darktable.org, like dtstyles site for styles.

Names are actually better because more clear now !

ralfbrown:
I agree that the proper long-term solution is to have a themes site analogous to the dtstyles site.  In the short term, here are two possible approaches to making more themes available without cluttering the pulldown menu:

1) minimal effort: put most themes in an alt-themes directory, with just one representative among near variants in the main data/themes directory, and add a short paragraph to the documentation pointing at that directory and instructing users to copy files to their .config directory.  Can also move the multiple minor variations among the existing themes there if desired.  Since theme files are only a couple of KB each, the space used for their installation is minimal compared to the rest of darktable.

2) more effort: tree-structured menu built from the hierarchy of subdirectories under data/themes.  Probably not worth implementing if there is a definite plan in place for a dtthemes site.

TurboGit:
I will close this for now, nothing to be done on dt side.

PR #4266, opened 2020-02-09
fix #4021

### LINKED ISSUE #4021: [mask refinement] feathering radius weird artifacts

rabauke:
Addresses the same issue as pull request #4226. This change, however, is more appropriate in my view.

Context: Mask opacity equal to 1 means that all mask values become 1 (fully opaque) except those which where zero (fully transparent) before mask refinement via _mask opacity_ and _mask contrast_ sliders. In practice, however, we have to account for finite precision floating point arithmetics. The application of the guided filter to the image mask can change mask values which are equal to zero to some tiny non-zero values due to finite precision floating point arithmetics. We have to account for these possible rounding errors later in the mask refinement, in particular if _mask opacity_ is +/-1.

Furthermore, numerical accuracy of the guided filter is enhanced by using `double` for accumulation in the calculation of the sliding-box mean.

### DIFF
--- a/data/kernels/blendop.cl
+++ b/data/kernels/blendop.cl
@@ -1009,9 +1009,9 @@ blendop_mask_tone_curve(__read_only image2d_t mask_in, __write_only image2d_t ma
   float opacity = read_imagef(mask_in, sampleri, (int2)(x, y)).x;
   float scaled_opacity = (2.f * opacity / gopacity - 1.f);
   if (1.f - brightness <= 0.f)
-    scaled_opacity = opacity <= FLT_EPSILON ? -1.f : 1.f;
+    scaled_opacity = opacity <= 16 * FLT_EPSILON ? -1.f : 1.f;
   else if (1.f + brightness <= 0.f)
-    scaled_opacity = opacity >= 1.f - FLT_EPSILON ? 1.f : -1.f;
+    scaled_opacity = opacity >= 1.f - 16 * FLT_EPSILON ? 1.f : -1.f;
   else if (brightness > 0.f)
   {
     scaled_opacity = (scaled_opacity + brightness) / (1.f - brightness);
--- a/data/kernels/guided_filter.cl
+++ b/data/kernels/guided_filter.cl
@@ -18,7 +18,6 @@
 
 #include "common.h"
 
-
 kernel void guided_filter_split_rgb_image(const int width, const int height, read_only image2d_t in,
                                           write_only image2d_t out_r, write_only image2d_t out_g,
                                           write_only image2d_t out_b, const float guide_weight)
@@ -34,6 +33,16 @@ kernel void guided_filter_split_rgb_image(const int width, const int height, rea
 }
 
 
+// Kahan summation algorithm
+#define Kahan_sum(m, c, add)                                                                                      \
+  {                                                                                                               \
+    const float t1 = (add) - (c);                                                                                 \
+    const float t2 = (m) + t1;                                                                                    \
+    c = (t2 - m) - t1;                                                                                            \
+    m = t2;                                                                                                       \
+  }
+
+
 kernel void guided_filter_box_mean_x(const int width, const int height, read_only image2d_t in,
                                      write_only image2d_t out, const int w)
 {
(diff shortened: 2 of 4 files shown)


data/kernels/blendop.cl
```diff
@@ -1009,9 +1009,9 @@ blendop_mask_tone_curve(__read_only image2d_t mask_in, __write_only image2d_t ma
   float opacity = read_imagef(mask_in, sampleri, (int2)(x, y)).x;
   float scaled_opacity = (2.f * opacity / gopacity - 1.f);
   if (1.f - brightness <= 0.f)
-    scaled_opacity = opacity <= FLT_EPSILON ? -1.f : 1.f;
+    scaled_opacity = opacity <= 16 * FLT_EPSILON ? -1.f : 1.f;
```
TurboGit: Can you tell me more about the choice of 16 here ?

data/kernels/guided_filter.cl
```diff
@@ -17,7 +17,7 @@
 */
 
 #include "common.h"
-
+#pragma OPENCL EXTENSION cl_khr_fp64 : enable
```
TurboGit: Is that mandatory ? I understand that a pragma is just a request to the compiler that may or may not be honored, but what happens if there is no such OpenCL extension available ?

data/kernels/blendop.cl
rabauke: The factor 16 is basically an empirical safety factor. I have no better justification so far. :-(

data/kernels/guided_filter.cl
rabauke: This is mandatory to enable double precision floating point arithmetic. I revised my code to use the [Kahan summation algorithm](https://en.wikipedia.org/wiki/Kahan_summation_algorithm), which allows me to implement the sliding mean entirely in single precision without losing (too much of) numerical precision. In this way I was able to get rid of the pragma avoiding possible portability issues.

TurboGit:
Ok, works for me. Thanks.

TurboGit:
@rabauke : Just a side note, please use a meaningful first line in the comment log. Instead of:

> fix #4021  

I would have used here a first line (less than 80 chars) describing the why/where of the fix and the reference in the log after (as 3rd line). An empty line separate the subject from the body in Git world.

> Fix mask refinement for small values (opacity near 0).
> 
> Fixes #4021.

Thanks.

jenshannoschwalm:
@rabauke thanks for introducing Kahan's algorithm to me :-)

data/kernels/blendop.cl
cryptomilk: I would add that as a comment:

`// 16 is just an empirical safety factor`

or add a `#define` for it. Names are better than magic numbers :-)

PR #4268, opened 2020-02-10
Fix accelerator name for timeline

dterrahe:
The "/" in the accelerator name causes it to be split over two levels in the shortcuts dialog


TurboGit:
Indeed an accel name should not have a / otherwise it creates a hierarchy.

TurboGit:
Merged in master and 3.0.1.

PR #4271, opened 2020-02-10
po/copy module order

TurboGit:


TurboGit:
New version pushed which support styles.

Nilvus:
@TurboGit : Copy/paste from an image to another apply module order, so this part is ok for me.
But applying a style doesn't work (or maybe I missed something) : I have updated my style by selecting an image with 3.0 order and then include modules_order line then save style to update it. After that, when applying a style on an image, module_order is not updated.

TurboGit:
@Nilvus :: On my side it does ! Applying a style created from a v3.0 module order to an image having the 2.6 (legacy) module order does change it to 3.0 when I select the "module order" in the list.

Can you save your style and send it to me ? Thanks.

TurboGit:
BTW, I'm not clear of the dialog. Having the "module order" on first line is not really clean I think. I'd like to include a separator between the module order and all modules but I don't know how to do that in Gtk (or even if possible!).

Nilvus:
> @Nilvus :: On my side it does ! Applying a style created from a v3.0 module order to an image having the 2.6 (legacy) module order does change it to 3.0 when I select the "module order" in the list.
> 
> Can you save your style and send it to me ? Thanks.

I have updated a style. Maybe there's a difference between a new style and an updated one. I will test and send you ASAP my style.

PR #4272, opened 2020-02-10
Minor fix on collect.c

EdgarLux:
Adding line with _"no-c-format"_ to avoid error **"msgtr is not a valid C string format, unlike msgid"** when translating


TurboGit:
Perfect ! I had this error some time ago and did not know how to bypass it. Thanks!

EdgarLux:
> Perfect ! I had this error some time ago and did not know how to bypass it. Thanks!

I'm glad I found it and fix it. It's my pleasure!

PR #4282, opened 2020-02-12
tagging fixes

phweyland:
- gtk error message when delete tag with name containing '&'
- tree display issue when creating a tag which is part of the path


TurboGit:
Looks good ! Thanks.

PR #4284, opened 2020-02-13
doc/manual: update section about module order

burrima:
This patch updates the section in the manual about pixelpipe module
re-ordering. It is now explained that the module order can be changed
(with great care).

Signed-off-by: Martin Burri <info@burrima.ch>


TurboGit:
Looks good to me. Thanks.

PR #4289, opened 2020-02-14
imageio: Don't use the slowest speed which is not recommended

cryptomilk:
It isn't recommend to use the extremities.


TurboGit:
Thanks !

Issue #3424, opened 2019-11-14
Search case sensitive [Bug or feature?]


sotsoguk:
**Describe the bug**
I am using the german version of darktable 3.0.0rc0. When i am searching a module, not all expected results are found. At least in the german version, the module names are sometimes only lowercase and sometimes starting with a capitel letter. Looking for "Farb*" does not reveal some modules which in the german version are name "*f*arb".

**To Reproduce**
1. Go to module search
2. Search for a module

**Expected behavior**
For me it would be more useful if the search function would be case insensitive

**Platform (please complete the following information):**
 - Windows 10
 - 3.0.0rc0

**Additional context**
Is there any link / guide how to contribute to the translation of darktable? I would like to fix these problems of wrong upper / lowercase words in the german translation.

**Screenshots**
![darktable_vVBxLXRIt8](https://user-images.githubusercontent.com/462611/68845278-789e7600-06cb-11ea-8a8e-d22fdbe3e7aa.png)
![darktable_yVOk48P9AT](https://user-images.githubusercontent.com/462611/68845281-789e7600-06cb-11ea-90d7-e9d18c304f40.png)

c-h-r-i-s-t-i-a-n:
I agree that the search should be case insensitive.

junkyardsparkle:
Agreed, and as a side-note, I just noticed that changing the visibility of the searchbox in preferences seems to require a restart, but this isn't indicated in the tooltip.

arigit:
+1 and I would add: can we have "Escape" clear the search text box? currently there doesn't seem to be any keyboard shortcut to clear the searchbox

aurelienpierre:
Fixed by #4290

PR #4290, opened 2020-02-14
Made modules search case insensetive

salsergey:
Related to the issue #3424. It is really an issue for some languages, where the names of the modules contain capital letters.

### DIFF
--- a/src/libs/modulegroups.c
+++ b/src/libs/modulegroups.c
@@ -399,9 +399,9 @@ static void _lib_modulegroups_update_iop_visibility(dt_lib_module_t *self)
         }
         else
         {
-          int is_match = (g_strstr_len(g_utf8_casefold(dt_iop_get_localized_name(module->op), -1), -1,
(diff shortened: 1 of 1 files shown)


src/libs/modulegroups.c
```diff
@@ -399,7 +399,9 @@ static void _lib_modulegroups_update_iop_visibility(dt_lib_module_t *self)
         }
         else
         {
-          int is_match = (g_strstr_len(dt_iop_get_localized_name(module->op), -1, text_entered) != NULL);
+          int is_match = (g_strstr_len(g_utf8_casefold(dt_iop_get_localized_name(module->op), -1), -1,
```
TurboGit: As we are at it, please makes this a const. Thanks.

salsergey: Done. Thank you.

TurboGit:
Merged in master and 3.0.1.

PR #4291, opened 2020-02-14
Collection - Add two entries:

phweyland:
Collection - Add two entries:
- module
- order

if plugins/lighttable/collect/module=TRUE in darktablerc.

I've missed the way to get a more user friendly module name. Any idea ?

image  image  image

What does surprise me is to have so many custom images. Most of them come from dt 2.6 if I'm not mistaken.

### DIFF
--- a/src/common/collection.c
+++ b/src/common/collection.c
@@ -1502,6 +1502,23 @@ static gchar *get_query_string(const dt_collection_properties_t property, const 
       query = dt_util_dstrcat(query, "(id %s group_id)", (strcmp(escaped_text, _("group leaders")) == 0) ? "=" : "!=");
       break;
 
+    case DT_COLLECTION_PROP_MODULE: // dev module
+      query = dt_util_dstrcat(query, "(id IN (SELECT imgid AS id FROM main.history WHERE operation "
+                                     "LIKE '%%%s%%'))", escaped_text);
+      break;
+
+    case DT_COLLECTION_PROP_ORDER: // iop order
+      {
+        int i = 0;
+        for(i = 0; i < DT_IOP_ORDER_LAST; i++)
+        {
+          if(strcmp(escaped_text, _(dt_iop_order_string(i))) == 0) break;
+        }
+        query = dt_util_dstrcat(query, "(id IN (SELECT imgid AS id FROM main.module_order WHERE version "
+                                       "= %d))", i);
+      }
+      break;
+
     default:
       // we shouldn't be here
       break;
--- a/src/common/collection.c
+++ b/src/common/collection.c
@@ -1507,7 +1507,7 @@ static gchar *get_query_string(const dt_collection_properties_t property, const 
         dt_iop_set_module_name_table();
         query = dt_util_dstrcat(query, "(id IN (SELECT imgid AS id FROM main.history AS h "
                                        "JOIN memory.module_names AS m ON m.operation = h.operation "
-                                       "WHERE m.name LIKE '%%%s%%'))", escaped_text);
+                                       "WHERE m.name LIKE '%s'))", escaped_text);
       }
       break;
 
--- a/src/common/collection.c
+++ b/src/common/collection.c
@@ -1504,9 +1504,8 @@ static gchar *get_query_string(const dt_collection_properties_t property, const 
 
     case DT_COLLECTION_PROP_MODULE: // dev module
       {
-        dt_iop_set_module_name_table();
         query = dt_util_dstrcat(query, "(id IN (SELECT imgid AS id FROM main.history AS h "
-                                       "JOIN memory.module_names AS m ON m.operation = h.operation "
+                                       "JOIN memory.darktable_iop_names AS m ON m.operation = h.operation "
                                        "WHERE m.name LIKE '%s'))", escaped_text);
       }
       break;
--- a/src/common/collection.c
+++ b/src/common/collection.c
@@ -1503,8 +1503,12 @@ static gchar *get_query_string(const dt_collection_properties_t property, const 
       break;
 
     case DT_COLLECTION_PROP_MODULE: // dev module
-      query = dt_util_dstrcat(query, "(id IN (SELECT imgid AS id FROM main.history WHERE operation "
-                                     "LIKE '%%%s%%'))", escaped_text);
+      {
+        dt_iop_set_module_name_table();
+        query = dt_util_dstrcat(query, "(id IN (SELECT imgid AS id FROM main.history AS h "
+                                       "JOIN memory.module_names AS m ON m.operation = h.operation "
+                                       "WHERE m.name LIKE '%%%s%%'))", escaped_text);
+      }
       break;
 
     case DT_COLLECTION_PROP_ORDER: // iop order
--- a/src/common/collection.c
+++ b/src/common/collection.c
@@ -1517,8 +1517,11 @@ static gchar *get_query_string(const dt_collection_properties_t property, const 
         {
           if(strcmp(escaped_text, _(dt_iop_order_string(i))) == 0) break;
         }
-        query = dt_util_dstrcat(query, "(id IN (SELECT imgid AS id FROM main.module_order "
-                                       "WHERE version = %d))", i);
+        if(i < DT_IOP_ORDER_LAST)
+          query = dt_util_dstrcat(query, "(id IN (SELECT imgid FROM main.module_order "
+                                         "WHERE version = %d))", i);
+        else
+          query = dt_util_dstrcat(query, "(id NOT IN (SELECT imgid FROM main.module_order))");
       }
       break;
 
--- a/src/common/collection.c
+++ b/src/common/collection.c
@@ -1518,8 +1518,8 @@ static gchar *get_query_string(const dt_collection_properties_t property, const 
         {
           if(strcmp(escaped_text, _(dt_iop_order_string(i))) == 0) break;
         }
-        query = dt_util_dstrcat(query, "(id IN (SELECT imgid AS id FROM main.module_order WHERE version "
-                                       "= %d))", i);
+        query = dt_util_dstrcat(query, "(id IN (SELECT imgid AS id FROM main.module_order "
+                                       "WHERE version = %d))", i);
       }
       break;
 
--- a/src/libs/collect.c
+++ b/src/libs/collect.c
@@ -2204,7 +2204,7 @@ void gui_init(dt_lib_module_t *self)
   GtkBox *box;
   GtkWidget *w;
 
-  const gboolean show_module = dt_conf_get_bool("plugins/lighttable/collect/module");
+  const gboolean hide_module = dt_conf_get_bool("plugins/lighttable/collect/hidemodule");
 
   for(int i = 0; i < MAX_RULES; i++)
   {
@@ -2215,8 +2215,8 @@ void gui_init(dt_lib_module_t *self)
     gtk_box_pack_start(GTK_BOX(self->widget), GTK_WIDGET(box), TRUE, TRUE, 0);
     w = gtk_combo_box_text_new();
     d->rule[i].combo = GTK_COMBO_BOX(w);
-    // skip module and order if not configured
-    const int count = show_module ? dt_lib_collect_string_cnt : dt_lib_collect_string_cnt - 2;
+    // skip module and order if hidden
+    const int count = hide_module ? dt_lib_collect_string_cnt - 2 : dt_lib_collect_string_cnt;
     for(int k = 0; k < count; k++)
       gtk_combo_box_text_append_text(GTK_COMBO_BOX_TEXT(w), _(dt_lib_collect_string[k]));
     g_signal_connect(G_OBJECT(w), "changed", G_CALLBACK(combo_changed), d->rule + i);
@@ -2295,7 +2295,7 @@ void gui_init(dt_lib_module_t *self)
 
   _lib_collect_gui_update(self);
 
-  if(show_module)
+  if(!hide_module)
   {
     // force redraw collection images because of late update of the table memory.darktable_iop_names
     dt_collection_update_query(darktable.collection);
--- a/src/libs/collect.c
+++ b/src/libs/collect.c
@@ -1508,6 +1508,40 @@ static void list_view(dt_lib_collect_rule_t *dr)
                    _("group leaders"),  _("group followers"), where_ext);
         break;
 
+      case DT_COLLECTION_PROP_MODULE: // module
+        snprintf(query, sizeof(query),
+                 "SELECT operation, 1, COUNT(*) AS count"
+                 " FROM main.images AS mi"
+                 " JOIN "
+                 "   (SELECT imgid, operation FROM main.history) h"
+                 "  ON h.imgid = mi.id"
+                 " WHERE %s"
+                 " GROUP BY operation"
+                 " ORDER BY operation",
+                 where_ext);
+        break;
+
+      case DT_COLLECTION_PROP_ORDER: // modules order
+        {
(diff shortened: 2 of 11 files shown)


TurboGit:
Nice !

One question, do you even sleep or are you a robot :)

I'll review ASAP !

Nilvus:
@TurboGit: as @phweyland said, it would be good to see the user module name and not the file name. For example, having tone contrast instead of atrous.

TurboGit:
> I've missed the way to get a more user friendly module name. Any idea ?

Would this routine in imagiop.h do the trick:

`gchar *dt_iop_get_localized_name(const gchar *op);`

TurboGit:
@Nilvus : Yes, was on it and I think I found the solution.

TurboGit:
> What does surprise me is to have so many custom images. Most of them come from dt 2.6 if I'm not mistaken.

Or non edited pictures ?

src/libs/collect.c
```diff
@@ -1508,6 +1508,40 @@ static void list_view(dt_lib_collect_rule_t *dr)
...
 
+      case DT_COLLECTION_PROP_MODULE: // module
+        snprintf(query, sizeof(query),
+                 "SELECT operation, 1, COUNT(*) AS count"
+                 " FROM main.images AS mi"
+                 " JOIN "
+                 "   (SELECT imgid, operation FROM main.history) h"
+                 "  ON h.imgid = mi.id"
+                 " WHERE %s"
+                 " GROUP BY operation"
```
TurboGit: I'm wondering if group_by and order_by won't be slow without an index for history.operation. Have you seen speed issue ?

phweyland:
> Would this routine in imagiop.h do the trick:
> 
> `gchar *dt_iop_get_localized_name(const gchar *op);`

I'll look at this.

> Or non edited pictures ?

I'm not sure. I've migrated from lr early 2018 and a lot of raw files have been updated through ligthroom.c.
However I've imported also a lot of a lot of lr exports (jpg) in order not to loose too much work. :-)
 
The other doubt is that I try to keep my dt 2.7 database. I'll make try with a true 2.6 database.

src/libs/collect.c
phweyland: I haven't seen any issue there (so far).

TurboGit: So nothing to do, will see if needed later.

phweyland:
module collection lists now the localized names.
I still see some issues:
- filmic selection gives filmic + filmic rgb images
- very often the number of displayed images is lower than the counted ones in modules list. Should be the same. I don't know yet the cause (may be the state of my db).

phweyland:
image

phweyland:
Except the readiness of darktable.iop at start time that seems OK for me.
 
Doubt. In history there the module numeric field which seems to identify the module too. It could faster to use it instead of operation. However this number is not available on dt_iop_module_so_t.

src/common/collection.c
```diff
@@ -1502,6 +1502,27 @@ static gchar *get_query_string(const dt_collection_properties_t property, const
       query = dt_util_dstrcat(query, "(id %s group_id)", (strcmp(escaped_text, _("group leaders")) == 0) ? "=" : "!=");
       break;
 
+    case DT_COLLECTION_PROP_MODULE: // dev module
+      {
+        dt_iop_set_module_name_table();
```
TurboGit: We want to call this only once at startup I think. So better place it during initialization of the application in darktable.c.

TurboGit: I would propose at the end of dt_init() routine at least after the call to   dt_iop_load_modules_so();

phweyland: It's what I've done. But this doesn't solve that the list is not ready when collect.c refreshes the screen.

phweyland: sorry, not collect.c but dt_collection_update()

phweyland: dt_collection_update() is called by dt_collection_new() and dt_selection_new()
around line 900 in darktable.c

TurboGit: So let's try to move both calls around line 980 after dt_iop_load_modules_so(); It should work.

phweyland: seg fault ...
I've tried to call dt_collection_update(darktable.collection) after updating the list.
I get this:
image
Like a kind of redraw was missing ...

phweyland: but   dt_control_queue_redraw_center(); doesn't help

phweyland:
I think that's ok now.

phweyland:
Migrating from a 2.6 db, all the modules are custom for me.
I'd expected all legacy, wouldn't I ?

phweyland:
> Migrating from a 2.6 db, all the modules are custom for me.

If I'm not mistaken discrepancy comes between mask_manager (legacy order) and denoiseprofile (images) for all images...

TurboGit:
> If I'm not mistaken discrepancy comes between mask_manager (legacy order) and denoiseprofile (images) for all images...

Could be the case indeed, the mask_manager may move in the iop list IIRC. Not something that will break things but indeed annoying ! Probably something to fix in this area.

phweyland:
Hummm. 2.6 has not been released. So it's already an in between db.
But I've tried with a db I've marked a 2.4.4 (29/9/2018) and the oldest db I have (10/4/2018) with the same result.

phweyland:
Is the legacy_order list really correct ? If I remember correctly, lut3d was not born 

> . 2.6 has not been released.

Sorry, it has been actually.

phweyland:
> 2.6 has not been released. 

Sorry, my mistake,  it has been released actually.

src/libs/collect.c
```diff
@@ -2169,6 +2204,8 @@ void gui_init(dt_lib_module_t *self)
   GtkBox *box;
   GtkWidget *w;
 
+  const gboolean show_module = dt_conf_get_bool("plugins/lighttable/collect/module");
```
TurboGit: Probably time to remove this now ? This was for debug purpose, right ? Applying your PR I was surprised to not find the two new filters !

src/libs/collect.h
```diff
@@ -36,7 +36,9 @@ const char *dt_lib_collect_string[] = { N_("film roll"),    N_("folders"),     N
                                         N_("rights"),       N_("lens"),        N_("focal length"),
                                         N_("ISO"),          N_("aperture"),    N_("exposure"),
                                         N_("aspect ratio"), N_("filename"),    N_("geotagging"),
-                                        N_("grouping"),     N_("local copy") };
+                                        N_("grouping"),     N_("local copy"),
+                                        // last two ones - skipped if not configured in darktablerc
```
TurboGit: And the comment here.

src/libs/collect.c
phweyland: My idea was not to bother people who are not concerned with migration with this.
But of course we can show them all the time. not problem.

phweyland: > Activate those filter by default.

Ok, I do this.

phweyland:
> Maybe just ignoring the placement of the mask_manager in the list would be sufficient.

What about if I replace 0 in module_order by 1 everywhere ?

TurboGit:
> What about if I replace 0 in module_order by 1 everywhere ?

You'll break your edit, what you can try is to:
- replace 0 by 1 everywhere
- remove the iop_list by NULL for those entries

And backup your DB :)

phweyland:
> what you can try is to:
> 
>     * replace 0 by 1 everywhere
> 
>     * remove the iop_list by NULL for those entries

Seems to work fine.
I can set v3.0 (I get 2 in version, iop_list still NULL) or change the order (I get 0 in version and the custom list in iop_list).

TurboGit:
Well probably a bad advice. This will work for images without multi instance but it will break the ones that are really custom.

TurboGit:
So you should do this only for image without multi insurance. So only for images without a multi priority > 0 in the history.

phweyland:
Understood.
As my 2.7 backups are too messy I will have to consolidate on my last 2.6.
If you plan to mitigate this, I'll way. If not I may change the code as you suggested above to get the legacy where it should be.

phweyland:
Skipping mask_manager in dt_ioppr_get_iop_order_list_kind() for legacy gives me 100% of legacy.
So I think that at this time I was not using multi instances ... :)

phweyland:
With the change you suggested above, I've still 100% legacy from 2.6.
with these discrepancies:

legacy order - image
tonemap bilateral (most of them)
denoiseprofile bilateral
tonemap rotatepixels
tonemap bilateral
rawprepare collect

phweyland:
Checking the multi instances, ... would not there be an issue in darktable.c line 1293 ?

            if(next
               && strcmp(((dt_iop_order_entry_t *)(l->data))->operation,
                         ((dt_iop_order_entry_t *)(next->data))->operation))

I would replace by:

            if(next
               && !strcmp(((dt_iop_order_entry_t *)(l->data))->operation,
                         ((dt_iop_order_entry_t *)(next->data))->operation))

phweyland:
When I import my 2.6 db, I have some images without iop_order (not in the table module_order).
So I've added an order entry "none". 

When I open one of these images in darkroom the displayed order is v3.0. This image is then updated as v3.0 in module_order while I don't make any change on it.

> do you even sleep or are you a robot :)

Don't worry for the time, I'm on Brasilia time zone for the time being :)

TurboGit:
> Checking the multi instances, ... would not there be an issue in darktable.c line 1293 ?
> 
> ```
>         if(next
>            && strcmp(((dt_iop_order_entry_t *)(l->data))->operation,
>                      ((dt_iop_order_entry_t *)(next->data))->operation))
> ```
> 
> I would replace by:
> 
> ```
>         if(next
>            && !strcmp(((dt_iop_order_entry_t *)(l->data))->operation,
>                      ((dt_iop_order_entry_t *)(next->data))->operation))
> ```

Yes :( Again this stupid strcmp() mistake ! I hate C :(

phweyland:
My turn to worry about your sleep time... :)

TurboGit:
But this was not an issue, it was only a way to avoid writing the iop_list in module_order. With the wrong check we have many iop_list written even though the order is a canonical one (legacy or v3.0). So not big deal which is luck!

phweyland:
I've slightly modified the code to keep multiples instance as custom and all others set as legacy (as I suppose that order difference in 2.6 was only an error).

I guess that images without iop_order were missing in the previous history. I don't know if this case is a normal one, but the query in line database.c lines 1200-1204 could retrieve also the those records and the corresponding images set with a safe iop_order.

phweyland:
>all others set as legacy (as I suppose that order difference in 2.6 was only an error).

And BTW I'll save myself from the tons of warning I get with the wrong iop order messages :)

TurboGit:
> and all others set as legacy

My idea was to keep non edited picture to have them using v3.0 order first time entering in darkroom.

TurboGit:
Anyway, merged now. Thanks.

phweyland:
> y idea was to keep non edited picture to have them using v3.0 order first time entering in darkroom.

Are non edited pictures the ones which are not in the history table ?
If yes, that's the current behavior and that's ok.

Nilvus:
@phweyland: Thanks for this add, really useful. Anyway, I'm quite surprised to see that both activated and deactivated modules are proposed on module filter. Would it be possible to only have activated module on this filter ? For example, I have some old images where I see a module I don't use anymore but I use on the past because it just remains on history. One use I see actually to this filter is to see if I forgot some images with an old module I want to remove is always there. Not possible by now.

TurboGit:
@Nilvus : Indeed, makes more sense to have only activated modules I suppose.

phweyland:
Agreed. I'll do that.
I didn't bother with that because I usually clean up my history as much as I can. But this is boring, I'll try to see if I can create an option to make the history clean-up complete :)

phweyland:
BTW do you think it is worth to have the option to hide module and order entry ? 

Other question. I think we could remove the history entry as it is covered by order entry.
history - non altered = order - none ( I could change this one as 'non altered' to appear normal).

phweyland:
Is the entry 'order' well named. If not what is your preference ? history (replacing the existing one) ? module order ?

johnny-bit:
@phweyland - I tried small fix in #4301 - can you check if you have noticable difference in timings between query `WHERE h.enabled=1 and m.name LIKE '%s'`  and `WHERE m.name LIKE '%s' and h.enabled=1`?
My tests indicate that checking for `LIKE` first is a bit faster, but my library is tiny (~couple thousands photos and a bit more history items)

TurboGit:
> BTW do you think it is worth to have the option to hide module and order entry ?

No,, I think we can remove this.

Nilvus:
> Agreed. I'll do that.
> I didn't bother with that because I usually clean up my history as much as I can. But this is boring, I'll try to see if I can create an option to make the history clean-up complete :)

Why not an option to make history clean-up complete.

TurboGit:
My preference will go for "module order". This is the name of the lib to change module order, so seems good for consistency.

Nilvus:
> Other question. I think we could remove the history entry as it is covered by order entry.
> history - non altered = order - none ( I could change this one as 'non altered' to appear normal).

We could remove this but with an option button to do that. Some times, it's good to see which modules we disable.

phweyland:
> > Other question. I think we could remove the history entry as it is covered by order entry.
> > history - non altered = order - none ( I could change this one as 'non altered' to appear normal).
> 
> We could remove this but with an option button to do that. Some times, it's good to see which modules we disable.

If I understand your wish, it is not related to 'history' entry but to 'module' where we could have a configuration option to include disabled ones.
But we could also have two entries per module, one for enabled and disabled .

phweyland:
> can you check if you have noticable difference in timings between query `WHERE h.enabled=1 and m.name LIKE '%s'` and `WHERE m.name LIKE '%s' and h.enabled=1`?

I'll try to measure that.

Nilvus:
> > > Other question. I think we could remove the history entry as it is covered by order entry.
> > > history - non altered = order - none ( I could change this one as 'non altered' to appear normal).
> > 
> > 
> > We could remove this but with an option button to do that. Some times, it's good to see which modules we disable.
> 
> If I understand your wish, it is not related to 'history' entry but to 'module' where we could have a configuration option to include disabled ones.
> But we could also have two entries per module, one for enabled and disabled .

I have no wish there as I let history as it is. My only wish related to this PR is what what @johnny-bit do on his #4301 PR. Actual history is great as it is for me.

But for those wanted to clean their history, I think the best is just to add a button to delete unused entries on history (so disabled modules). This way, that let other users to keep history as it is, so also with entries about disabled modules.

phweyland:
> > can you check if you have noticable difference in timings between query `WHERE h.enabled=1 and m.name LIKE '%s'` and `WHERE m.name LIKE '%s' and h.enabled=1`?
> 
> I'll try to measure that.

I don't see any difference 0,07 average in both cases with 50.000 images

johnny-bit:
> But for those wanted to clean their history, I think the best is just to add a button to delete unused entries on history (so disabled modules)

I thought that's what "compress history" did... Turns out it's a bit more complicated...

> I don't see any difference 0,07 average in both cases with 50.000 images

Thanks! So I'll think I'll do as @TurboGit advised and do `enabled` first, then `name LIKE` part.

Nilvus:
> > But for those wanted to clean their history, I think the best is just to add a button to delete unused entries on history (so disabled modules)
> 
> I thought that's what "compress history" did... Turns out it's a bit more complicated...

Indeed not, compress history, just compressed modules changed more than once a time in the image (so it's simplify the history by merging similar entries.

Anyway, I have not ask to change history, that's just a suggestion from @phweyland and I have just answer to this. As nobody else ask for that, maybe the simple way is to not touch to history. I really like the way it is.

PR #4301, opened 2020-02-16
Collect only based on enabled module in dev history

johnny-bit:
This small change makes sure that module collection from PR #4291 is only based off on enabled modules as suggested by @Nilvus

### DIFF
--- a/src/common/collection.c
+++ b/src/common/collection.c
@@ -1506,7 +1506,7 @@ static gchar *get_query_string(const dt_collection_properties_t property, const 
       {
         query = dt_util_dstrcat(query, "(id IN (SELECT imgid AS id FROM main.history AS h "
                                        "JOIN memory.darktable_iop_names AS m ON m.operation = h.operation "
-                                       "WHERE m.name LIKE '%s'))", escaped_text);
+                                       "WHERE h.enabled = 1 AND m.name LIKE '%s'))", escaped_text);
       }
       break;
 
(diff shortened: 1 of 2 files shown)


src/common/collection.c
```diff
@@ -1506,7 +1506,7 @@ static gchar *get_query_string(const dt_collection_properties_t property, const
       {
         query = dt_util_dstrcat(query, "(id IN (SELECT imgid AS id FROM main.history AS h "
                                        "JOIN memory.darktable_iop_names AS m ON m.operation = h.operation "
-                                       "WHERE m.name LIKE '%s'))", escaped_text);
+                                       "WHERE m.name LIKE '%s' AND h.enabled = 1))", escaped_text);
```
TurboGit: I'm wondering if rewritting to : h.enabled = 1 AND m.name LINE '%s' won't be more efficient and enable = 1 is certainly faster to check and so we avoid a LIKE check for enabled = 0.

johnny-bit: I don't think i have big enough collection to reliably time it but I'll try to calculate difference it makes.

johnny-bit: I've done several test now and the results contradict intuition. It seems that `enabled` isn't index despite being simple integer and actually in my tests having LIKE check on `operation` which is primary key for in-memory table is marginally faster in my limited DB of couple thousands photos. Regardless if module's enabled or not and whether module is often used/unused.

johnny-bit:
I've missed the count part of collection :sweat_smile:

TurboGit:
Looks good to me ! Thanks.

PR #4302, opened 2020-02-16
collect module - remove configuration hidden module

phweyland:
some fixes discussed on #4291.
- only enabled modules
- 'order' renamed to 'module order'
- 'none' entry renamed to 'not altered'
- removed configuration 'hidden module'

I haven't removed collection history entry (redundant with module order). Do I do it ?

### DIFF
--- a/src/libs/collect.c
+++ b/src/libs/collect.c
@@ -1304,25 +1304,19 @@ static void list_view(dt_lib_collect_rule_t *dr)
         break;
 
       case DT_COLLECTION_PROP_HISTORY: // History, 2 hardcoded alternatives
-        {
-          char *orders = NULL;
-          for(int i = 0; i < DT_IOP_ORDER_LAST; i++)
-          {
-            orders = dt_util_dstrcat(orders, "WHEN mo.version = %d THEN '%s - %s' ",
-                                     i, _("altered"), _(dt_iop_order_string(i)));
-          }
-          orders = dt_util_dstrcat(orders, "ELSE '%s' ", _("not altered"));
-          snprintf(query, sizeof(query),
-                   "SELECT CASE %s END as ver, 1, COUNT(*) AS count"
+        g_snprintf(query, sizeof(query),
+                   "SELECT CASE altered"
+                   "         WHEN 1 THEN '%s'"
+                   "         ELSE '%s'"
+                   "       END as altered, 1, COUNT(*) AS count"
                    " FROM main.images AS mi"
-                   " LEFT JOIN (SELECT imgid, version FROM main.module_order) mo"
-                   "  ON mo.imgid = mi.id"
+                   " LEFT JOIN (SELECT DISTINCT imgid AS history_id, 1 AS altered"
+                   "            FROM main.history)"
+                   "   ON id = history_id"
                    " WHERE %s"
-                   " GROUP BY ver"
-                   " ORDER BY ver",
-                   orders, where_ext);
-          g_free(orders);
-        }
+                   " GROUP BY altered"
+                   " ORDER BY altered ASC",
+                   _("altered"),  _("not altered"), where_ext);
         break;
 
       case DT_COLLECTION_PROP_GEOTAGGING: // Geotagging, 2 hardcoded alternatives
@@ -1528,6 +1522,28 @@ static void list_view(dt_lib_collect_rule_t *dr)
                  where_ext);
         break;
 
+      case DT_COLLECTION_PROP_ORDER: // modules order
+        {
+          char *orders = NULL;
+          for(int i = 0; i < DT_IOP_ORDER_LAST; i++)
+          {
+            orders = dt_util_dstrcat(orders, "WHEN mo.version = %d THEN '%s' ",
+                                     i, _(dt_iop_order_string(i)));
+          }
+          orders = dt_util_dstrcat(orders, "ELSE '%s' ", _("none"));
+          snprintf(query, sizeof(query),
+                   "SELECT CASE %s END as ver, 1, COUNT(*) AS count"
+                   " FROM main.images AS mi"
+                   " LEFT JOIN (SELECT imgid, version FROM main.module_order) mo"
+                   "  ON mo.imgid = mi.id"
+                   " WHERE %s"
+                   " GROUP BY ver"
+                   " ORDER BY ver",
+                   orders, where_ext);
+          g_free(orders);
+        }
+        break;
+
       default: // filmroll
         g_snprintf(query, sizeof(query),
                    "SELECT folder, film_rolls_id, COUNT(*) AS count"
@@ -1609,7 +1625,7 @@ static void list_view(dt_lib_collect_rule_t *dr)
                     || property == DT_COLLECTION_PROP_PUBLISHER || property == DT_COLLECTION_PROP_RIGHTS
                     || property == DT_COLLECTION_PROP_TITLE || property == DT_COLLECTION_PROP_APERTURE
                     || property == DT_COLLECTION_PROP_FOCAL_LENGTH || property == DT_COLLECTION_PROP_ISO
-                    || property == DT_COLLECTION_PROP_MODULE))
+                    || property == DT_COLLECTION_PROP_MODULE || property == DT_COLLECTION_PROP_ORDER))
     gtk_tree_model_foreach(model, (GtkTreeModelForeachFunc)list_match_string, dr);
   // we update list selection
   gtk_tree_selection_unselect_all(gtk_tree_view_get_selection(d->view));
@@ -2478,6 +2494,7 @@ void init(struct dt_lib_module_t *self)
   luaA_enum_value(L,dt_collection_properties_t,DT_COLLECTION_PROP_LOCAL_COPY);
   luaA_enum_value(L,dt_collection_properties_t,DT_COLLECTION_PROP_GROUPING);
   luaA_enum_value(L,dt_collection_properties_t,DT_COLLECTION_PROP_MODULE);
+  luaA_enum_value(L,dt_collection_properties_t,DT_COLLECTION_PROP_ORDER);
 
 }
 #endif
--- a/src/libs/collect.c
+++ b/src/libs/collect.c
@@ -1304,19 +1304,25 @@ static void list_view(dt_lib_collect_rule_t *dr)
         break;
 
       case DT_COLLECTION_PROP_HISTORY: // History, 2 hardcoded alternatives
-        g_snprintf(query, sizeof(query),
-                   "SELECT CASE altered"
-                   "         WHEN 1 THEN '%s'"
-                   "         ELSE '%s'"
-                   "       END as altered, 1, COUNT(*) AS count"
+        {
+          char *orders = NULL;
+          for(int i = 0; i < DT_IOP_ORDER_LAST; i++)
+          {
+            orders = dt_util_dstrcat(orders, "WHEN mo.version = %d THEN '%s - %s' ",
+                                     i, _("altered"), _(dt_iop_order_string(i)));
+          }
+          orders = dt_util_dstrcat(orders, "ELSE '%s' ", _("not altered"));
+          snprintf(query, sizeof(query),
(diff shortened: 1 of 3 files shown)


Nilvus:
So first fix you add is already in #4301 
Other ones are better choices, thanks.

About collection history entry, it's not totally the same as it show developed or not images. And it's not a problem anyway to keep it.

phweyland:
> About collection history entry, it's not totally the same as it show developed or not images. And it's not a problem anyway to keep it.

History presents 2 options - developed or not
Module order, split developed in different order and keep the not developed ones.
So yes, history is redundant.

johnny-bit:
When I did #4301 I haven't thought you would do this part too as I was wondering if I could manage to at least improve on it ;) I'm happy whichever PR gets merged regarding showing only enabled mudules.

> Module order, splits developed in different orders and keeps the not developed ones.
So yes, history is redundant.

It's UI/UX problem. With `module order` users get
* custom
* legacy
* v3.0
* not altered

That's not the same as altered/not altered, because if I only had `module order` to work with I'd have to do some math in my head (and operator logic) to know how to check for altered files regardless of module order :)

Currently with `module order` name logic, gramatically speaking, "none" is correct name for collection filter (meaning "module order: none")

So IMO the choice in module order sholud be as was, so:
* custom
* legacy
* v3.0
* none

and keep `history` collection as is, because it's easy mode with good UX. Even if it might be redundant.

Nilvus:
> > About collection history entry, it's not totally the same as it show developed or not images. And it's not a problem anyway to keep it.
> 
> History presents 2 options - developed or not
> Module order, splits developed in different orders and keeps the not developed ones.
> So yes, history is redundant.

Oh, I didn't understand that you add this on new order (I have 0 image not developed, so I don't see option you say). Of course, this way, that's redundant.

phweyland:
> When I did #4301 I haven't thought you would do this part too as I was wondering if I could manage to at least improve on it

No problem, I can rebase after your PR is merged.

> Currently with `module order` name logic, gramatically speaking, "none" is correct name for collection filter (meaning "module order: none")

I agree that  "module order: none" is grammatically more correct, but when I see them the first time, I was very unsure about the meaning of absence module order. `Not altered` provides that explanation.  What about:
`module order: none (not altered)` ?

> and keep `history` collection as is, because it's easy mode with good UX. Even if it might be redundant.

This is non sense for me.

johnny-bit:
> No problem, I can rebase after your PR is merged.
<3

>What about:
module order: none (not altered) ?

looks good. I would go with `none (non-altered images)` for completness sake but that might be too verbose.

>This is non sense for me.

Historically `history` collection is one of very early ones (~10 years in code?) and provides simple boolean altered/not choice fo users. While it is redundant and less powerful than `module order` I still think it's easier to keep it and let users have both.

phweyland:
> Historically `history` collection is one of very early ones (~10 years in code?) and provides simple boolean altered/not choice fo users. While it is redundant and less powerful than `module order` I still think it's easier to keep it and let users have both.

no problem. We have just to keep in mind that house keeping should be done time to time :)

phweyland:
> > What about:
> > module order: none (not altered) ?
> 
> looks good. I would go with `none (non-altered images)` for completness sake but that might be too verbose.

Another possible way could be to reuse history entry (and remove module order). Instead of 
- altered
- not altered

it would become:
- custom
- legacy
- not altered
- v3.0

or any other wording you could find more appropriate

TurboGit:
> > > What about:
> > > module order: none (not altered) ?
> > 
> > 
> > looks good. I would go with `none (non-altered images)` for completness sake but that might be too verbose.
> 
> Another possible way could be to reuse history entry (and remove module order). Instead of
> 
>     * altered
> 
>     * not altered
> 
> 
> it would become: (EDIT)
> 
>     * altered - custom
> 
>     * altered - legacy
> 
>     * altered - v3.0
> 
>     * not altered
> 
> 
> or any other wording you could find more appropriate

Frankly I would prefer the current wording and just "none" for the non altered images.

src/libs/collect.c
```diff
@@ -1529,7 +1530,7 @@ static void list_view(dt_lib_collect_rule_t *dr)
           {
             orders = dt_util_dstrcat(orders, "WHEN mo.version = %d THEN '%s' ", i, _(dt_iop_order_string(i)));
           }
-          orders = dt_util_dstrcat(orders, "ELSE '%s' ", _("none"));
+          orders = dt_util_dstrcat(orders, "ELSE '%s' ", _("not altered"));
```
TurboGit: I prefer just "none" here. No strong opinion, but this seems clear enough and is less verbose.

johnny-bit:
@phweyland - Absolutelly agree!

@TurboGit - I think thet if @phweyland would actually replace `history` collection with `module order`, but call it `history` as to not confuse people accustomed to `history` if `history` was to be actually replaced :)

Have it like:

* altered - any
* altered - custom
* altered - legacy
* altered - v3.0
* not altered

And that would be absolutelly perfect (yeah, I know that `any` in that case would create internal redundancy). :)

so the question now: replace `history` with `module order` called `history` or leave `history` and add `module order` (in which case "none" instead of "not altered" wording is better)

phweyland:
> - altered - any

What is the case for this. If altered it's mandatorily either custom, legacy or v3 (any future version will be added automatically here).

TurboGit:
Ok, no strong opinion really. The solution to replace history with module order would work for me also.

phweyland:
image

johnny-bit:
> What is the case for this. If altered it's mandatorily either custom, legacy or v3 (any future version will be added automatically here).

If you allow multi-selection of options then there's no use-case for `any` :)

phweyland:
> If you allow multi-selection of options

This is another challenge. :)
I understand the case now. I'll see if I can do something.

phweyland:
> I understand the case now. I'll see if I can do something.

This is not impossible. There are different solutions but I won't do any for the time being.
- add an entry as you suggest. Union of tables is possible. But this is not coherent with the current UI... because there are some images which would sit in different entries. 
- add a root "altered" and some sub entries, custom, legacy, v3 ... This is coherent with the current UI. Elegant but we have to move from a liststore to treestore (as for folders, tags, ...). Surely more work than this simple fix.
- as you suggest, allow multiple selection.  This is really a challenge. This would be terrific for the whole collect module but I'm not sure that collection has been designed to support this...

johnny-bit:
>     * as you suggest, allow multiple selection.  This is really a challenge. This would be terrific for the whole collect module but I'm not sure that collection has been designed to support this...

Interesting though - you CAN select multiple in `date`, `time`, `focal length`, `ISO`, `aperture`, `exposure` & `aspect ratio` collections but that doesn't do anything. And adding support for that might pose a challenge (I wonder why it allows that though, but not on others?)

> add a root "altered" and some sub entries, custom, legacy, v3 ... This is coherent with the current UI. Elegant but we have to move from a liststore to treestore (as for folders, tags, ...). Surely more work than this simple fix.

That would be most elegant but most problematic... Otherwise I'd be all for it. And you're completelly right about UI coherence...

phweyland:
> Interesting though - you CAN select multiple in `date`, `time`, `focal length`, `ISO`, `aperture`, `exposure` & `aspect ratio` collections but that doesn't do anything. And adding support for that might pose a challenge (I wonder why it allows that though, but not on others?)

I didn't know that. I take note.

phweyland:
> Interesting though - you CAN select multiple in `date`, `time`, `focal length`, `ISO`, `aperture`, `exposure` & `aspect ratio` collections but that doesn't do anything.

plugins/lighttable/collect/single-click=TRUE plus shift-click makes the multi selection work...

phweyland:
> > Interesting though - you CAN select multiple in `date`, `time`, `focal length`, `ISO`, `aperture`, `exposure` & `aspect ratio` collections but that doesn't do anything.
> 
> plugins/lighttable/collect/single-click=TRUE plus shift-click makes the multi selection work...

But this works with numerical values. Not given for strings. :(
Good try !

TurboGit:
Note that current altered/non-altered in history filter does not give the same numbers of the order filter. Would be good to understand what the difference before replacing it.

phweyland:
> Note that current altered/non-altered in history filter does not give the same numbers of the order filter. Would be good to understand what the difference before replacing it.

On my freshly imported db from 2.6, both give the same data.
history is based on history table while the new query uses the module_order table.
If I've understood correctly both should have the images in them. Is that correct ?

phweyland:
But after importing some folders of images (600), the two tables diverge, actually.
77 which are in history but not in module_order,
112 which are in module_order but not in history.

phweyland:
I've reverted the merge of module order to history to make the differences visible here too.

phweyland:
@TurboGit please confirm if the number of images should be the same in both history and module_order.

phweyland:
Here are my findings:
- import a nef file without xmp adds an image is history but not in module_order
- remove an image removes it from history but not from module_order
- migration from 2.6 to 3.1 keeps not existing images in history
- altered symbol on thumbnails seems to follow history except for jpg

I would add that altered/not altered status from history is meaningless.
It is enough to open the image in darkroom to loose the not altered status. 
Importing raw image without xmp marks the image as altered
No change but altered ???

johnny-bit:
I think your findings require separate action as I think you've discovered inconsistency in history/module_order handling.

> * import a nef file without xmp adds an image is history but not in module_order
> * remove an image removes it from history but not from module_order

those should be imho 1:1 operations.

> It is enough to open the image in darkroom to loose the not altered status.
> Importing raw image without xmp marks the image as altered

IMHO: Importing raw image into lighttable shouldn't set altered state on it. Opening it in darktabe - yes, because even with no auto-applied presets/base-curve etc you still apply wb, orientation etc...

TurboGit:
To add to the complexity we have also the altered status in the lighttable thumbnails. This is using the dt_image_altered() routine which is not really in sync with what we have in the collect module.
 
Note that we used to have this comment in this routine:
```
    // FIXME: this is clearly a terrible way to determine which modules
    // are okay to still load the thumbnail and which aren't.
    // it is also used to display the altered symbol on the thumbnails.

```

The altered status from dt_image_altered() is a way to display if an image has been edited. That is, if any module in the history has been added either automatically (auto-preset) or manually by copy/pasting history or entering the darkroom. This is a good way to know what are the edited images and those that have not yet been worked.

So we have three area:
- the altered status icon in lighttable
- the altered/non-altered history filter in collect module
- the new order filter worked here

I think we want a single way (status) that actually keep the altered/non-altered status as it is today.

As you found out, entering the darkrroom and leaving it does change the order but not the altered status.

To answer some of your questions:

> Importing raw image without xmp marks the image as altered
> No change but altered ???

Maybe because you have auto-applied presets?

> I would add that altered/not altered status from history is meaningless.
> It is enough to open the image in darkroom to loose the not altered status.

That's not what I'm seeing. I tested with a RAW and Jpeg after suppressing the history in Lt I did entered darkroom and in both case leaving kept the non-altered status.

At this stage we really want to have a clear design about what we should do with all this. The ultimate goal would be to have a single status "somewhere" and use it for the altered icon in Lt and in the collect module.

phweyland:
> > Importing raw image without xmp marks the image as altered
> > No change but altered ???
> 
> Maybe because you have auto-applied presets?

I've unchecked all the auto-apply options in configuration ... 


> > It is enough to open the image in darkroom to loose the not altered status.
> 
> That's not what I'm seeing. I tested with a RAW and Jpeg after suppressing the history in Lt I did entered darkroom and in both case leaving kept the non-altered status.

I've just repeated this. select a jpg or a raw from history not-altered, go to darkroom and come back to lighttable, the image has gone and numbers altered/not altered are updated. 



> At this stage we really want to have a clear design about what we should do with all this. The ultimate goal would be to have a single status "somewhere" and use it for the altered icon in Lt and in the collect module.

Agreed. 

> separate action

I'll open an issue about this.

phweyland:
I've removed everything except the configuration hidden module removal to let you merge it without any other conflict.
I've also set back the "none" entry in "module order" as you suggested.

We will see later on what should be done with history entry.

TurboGit:
Perfect ! Tanks will merge this.

PR #4315, opened 2020-02-17
replace numerical constants by named constants

rabauke:
In response to comment by @cryptomilk, see https://github.com/darktable-org/darktable/pull/4266/files#r378697813


johnny-bit:
I would honestly also define that `16` as empirical safety factor :wink:

TurboGit:
Thanks !

PR #4317, opened 2020-02-17
fix macro usage in OpenCL kernel

rabauke:
OpenCL kernels must use `FLT_EPSILSON`, `CL_FLT_EPSILSON` is for host application code. Using the wrong macro may lead to compilation errors. See https://www.khronos.org/registry/OpenCL//sdk/2.2/docs/man/html/fpMacros.html


TurboGit:
Looks correct indeed! Good catch.

PR #4318, opened 2020-02-17
reduce scope of variables, minor fixes

rabauke:
Fixes for real bugs are indicated below.

### DIFF
--- a/src/common/opencl.c
+++ b/src/common/opencl.c
@@ -126,7 +126,7 @@ int dt_opencl_get_device_info(dt_opencl_t *cl, cl_device_id device, cl_device_in
 error:
   free(*param_value);
   *param_value = NULL;
-  param_value_size = 0;
+  *param_value_size = 0;
   return err;
 }
 
--- a/src/common/tags.c
+++ b/src/common/tags.c
@@ -554,7 +554,7 @@ void dt_tag_attach_string_list(const gchar *tags, gint imgid, const gboolean und
           e--;
         }
         e = *entry;
-        while(*e == ' ' && *e != '\0') e++;
+        while(*e == ' ') e++;
         if(*e)
         {
           guint tagid = 0;
@@ -836,7 +836,6 @@ GList *dt_tag_get_list(gint imgid)
     dt_tag_t *t = (dt_tag_t *)taglist->data;
     gchar *value = t->tag;
 
-    size_t j = 0;
     gchar **pch = g_strsplit(value, "|", -1);
 
     if(pch != NULL)
@@ -849,6 +848,7 @@ GList *dt_tag_get_list(gint imgid)
       }
       else
       {
+        size_t j = 0;
         while(pch[j] != NULL)
         {
           tags = g_list_prepend(tags, g_strdup(pch[j]));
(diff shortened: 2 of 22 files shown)


src/common/variables.c
```diff
@@ -513,7 +513,7 @@ static char *variable_get_value(dt_variables_params_t *params, char **variable)
           if(length >= 0)
             end = g_utf8_offset_to_pointer(start, MIN(length, start_utf8_length));
           else
-            end = g_utf8_offset_to_pointer(base_value + base_value_length, MAX(length, -1 * start_utf8_length));
```
rabauke: The result of the multiplication of -1 by the unsigned `start_utf8_length` and `base_value_utf8_length` yields an unsigned integer. Signed variables must be used here.

src/common/variables.c
```diff
@@ -454,7 +454,7 @@ static char *variable_get_value(dt_variables_params_t *params, char **variable)
       */
       {
         char *replacement = expand(params, variable, ')');
-        if(!base_value || !*base_value)
```
rabauke: This would lead to a null pointer dereference if `base_value` would be null.

src/common/tags.c
```diff
@@ -1323,7 +1323,7 @@ static gchar *dt_cleanup_synonyms(gchar *synonyms_entry)
         e--;
       }
       e = *entry;
-      while(*e == ' ' && *e != '\0') e++;
+      while(*e == ' ') e++;
       if(*e)
```
rabauke: Second test is redundant.

src/common/tags.c
```diff
@@ -554,7 +554,7 @@ void dt_tag_attach_string_list(const gchar *tags, gint imgid, const gboolean und
           e--;
         }
         e = *entry;
-        while(*e == ' ' && *e != '\0') e++;
```
rabauke: Second test is redundant.

src/common/opencl.c
```diff
@@ -126,7 +126,7 @@ int dt_opencl_get_device_info(dt_opencl_t *cl, cl_device_id device, cl_device_in
 error:
   free(*param_value);
   *param_value = NULL;
-  param_value_size = 0;
+  *param_value_size = 0;
   return err;
```
rabauke: Setting the _pointer_ `param_value_size` (a  function parameter, i.e., a local variable) to 0 makes no sense here.

TurboGit: Indeed!

src/common/variables.c
TurboGit: No as in C operator || does not evaluate the second argument if first one is true. Here NULL will make the first argument true and so if won't evaluate second one.

src/common/variables.c
TurboGit: Right!

src/common/variables.c
rabauke: You are right here. Nevertheless, the original code was confusing in my view. The null pointer test can be omitted.`!base_value ` _always_ evaluates to false, see comment in line 442. I suggest to omit the test here as done in line 473.

TurboGit: Agreed, the test should just be:

`if(!*base_value)`

TurboGit:
All good to me now. Thanks.

Issue #4309, opened 2020-02-17
Data race in compute_log_histogram in toneequal.c


mcv100:
**Describe the bug**
I was trying out and messing about with the tone equalizer and I found the behavior of the automatic mask exposure and contrast compensation very unpredictable. For instance, the histogram bar in the masking tab often spans to zero (-8 EV), even if the histogram clearly has no data around there.

**Debugging**
I built a custom version of Darktable, so that I could debug _compute_log_histogram_ and _histogram_deciles_ in _toneequal.c_. I found out that the total sum of the histogram is almost always lower than _num_elem_, which should definitely not happen (some pixels are not counted in the histogram). This leads to _last_pos_ being set to 0 if less than 90% of pixels are counted, giving rise to unpredictable behavior. 

The culprit can be found in the _compute_log_histogram_ function:

https://github.com/darktable-org/darktable/blob/8230fd80e13efa80027e58f667524286eadc707f/src/iop/toneequal.c#L1267-L1281

This for loop should not be parallelized, at least not in the way it is here. This will clearly lead to data race faults in incrementing _histogram_ in the threads (see e.g. https://stackoverflow.com/questions/26998183/how-do-i-deal-with-a-data-race-in-openmp). Commenting out the pragma fixes the problem in my testing.

**Possible fixes**
The easy fix is removing the omp pragma for this for loop of course. However, one may try to keep the code at least partially parallellized for performance. Perhaps have all threads work on their own version of _histogram_ with the _reduction_ clause?

TurboGit:
I'll fix that. Thanks.

mcv100:
Although this was tagged 3.0.1, as it stands, this fix (and less important, my improvement #4319) was not merged into 3.0.1. 

So still the exposure and contrast boost behave very funky. Not really a problem for me as I'm on master, but could be annoying for others.

PR #4319, opened 2020-02-18
Update OpenMP to 4.5 and improve performance in compute_log_histogram

mcv100:
Spawned by #4309, which got fixed by 7ef8429158f935303e87916ebc6e92051e31b584.

The fix in 7ef8429158f935303e87916ebc6e92051e31b584 can be improved a lot by avoiding atomic access with the reduction clause in OpenMP, which shaves off at least 10 ms on my machine. See also the discussion in 7ef8429158f935303e87916ebc6e92051e31b584.

Unfortunately, element wise array reduction is only supported from OpenMP 4.5 and up, so this does require a change in the required OpenMP version also. However, this version has been supported for a long time now, since GCC 6.1 and Clang 7.0 (?), so this should not pose serious issues. If it does, in principle the performance increase can be achieved with some other tricks as well, but the reduction clause is much easier to work with and simpler to follow.

Perhaps array reduction can be used in other cases in the existing code base as well, but I haven't looked much into other OpenMP usages yet.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -57,13 +57,13 @@ if (USE_OPENMP)
 
         # Clang has an option to specify the OpenMP standard to use. Specify it.
         # FIXME: Implement this in FindOpenMP.cmake
-        set(OPENMP_VERSION_SPECIFIER "-fopenmp-version=40")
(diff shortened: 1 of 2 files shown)


TurboGit:
Sure migrating to OpenMP 4.5 is ok to me as it is more than 4 years old.

Issue #4313, opened 2020-02-17
altered / not altered status and history / module_order inconcistencies


phweyland:
The altered / not altered status is available
- as an icon on thumbnails in lighttable
- as history entry in collect module

What I see:
- importing raw image without xmp and auto-apply options disabled sets the image as altered in history and on thumbnails.
- opening an image from history/not altered in darkroom and leaving darkroom without alteration sets the image as altered in history
- with history/altered selected, a lot of images haven't the icon altered
- EDIT: change white balance of a not altered image (thumbnail) and back to lighttable, the thumbnail is still not altered (but with different color :))

Related to that, images found in history and module_order are not kept consistent.

What I see (in database):

- ~~import a nef file without xmp, adds an image is history but not in module_order~~ done
- ~~remove an image, removes it from history but not from module_order~~ done

The altered / not altered status should be consistent across the UI and the user actions.
The tables history and module_order should stay consistent and, in some way, in agreement with the altered / not altered status.

phweyland:
From my standpoint this may be qualified as enhancement but that represents several bugs...

The routine dt_image_altered() is too simple to be accurate, but before trying to improve it, I think we can try to define properly what should be altered / not altered status.

Let's start. Please comment and complete as needed.

- just imported without auto-apply option nor xmp, the image should be set as not altered. The current list of modules and their state represent the not altered status.
- opening an image in darkroom without making any change should let it not altered
- any change made in darkroom, even change on an already existing module (white balance for example), should set the image as altered
- any move back to initial state of modules should set the image back to not altered (by ctrl-Z or by any other mean)
- at import time, if there is an xmp file, the altered / not altered status in xmp should be preserved.

open items
- if auto-apply options are set, do we want to have the status altered or not altered ? 
- for the internal mechanisms of dt, has the status altered / not altered the same meaning ?

TurboGit:
The current altered status is based on the fact that an image does not have a module that would change its rendering. As you find out for a RAW the white-balance is considered as a non-edit, this comes from dt_image_altered (let me put the code here):

```
           "SELECT 1"
           " FROM main.history, main.images"
           " WHERE id=?1 AND imgid=id AND num<history_end AND enabled=1"
           "       AND operation NOT IN ('flip', 'dither', 'highlights', 'rawprepare',"
           "                             'colorin', 'colorout', 'gamma', 'demosaic', 'temperature'%s%s)",
           basecurve_auto_apply ? ", 'basecurve'" : "",
[...]

Issue #4232, opened 2020-02-04
pixelpipe fired twice when opening darkroom


aurelienpierre:
**Describe the bug**
When opening a new image in darkroom (while darktable runs with `-d perf` in a terminal), it shows that the pixelpipe is computed twice for no reason (full preview only).

This is similar to some debugging I did while developing the tone equalizer: for some images, the IOP `commit_params()` method is fired up to 16 times when opening the darkroom.

**Platform :**
 - Darktable Version: 3.1
 - OS: Fedora 31
 - OpenCL Nvidia + Intel 
 - Nvidia 440.44, Intel Neo 20.04.15428

```
22,840833 [pixelpipe_process] [full] using device 0
22,857501 [dev_pixelpipe] took 0,016 secs (0,036 CPU) initing base buffer [full]
22,881662 [dev_pixelpipe] took 0,024 secs (0,032 CPU) processed `raw black/white point' on GPU, blended on GPU [full]
22,921965 [dev_pixelpipe] took 0,040 secs (0,060 CPU) processed `white balance' on GPU, blended on GPU [full]
22,953324 [dev_pixelpipe] took 0,031 secs (0,034 CPU) processed `highlight reconstruction' on GPU, blended on GPU [full]
23,061347 [dev_pixelpipe] took 0,108 secs (0,182 CPU) processed `demosaic' on GPU, blended on GPU [full]
23,187749 [dev_pixelpipe] took 0,126 secs (0,148 CPU) processed `exposure' on GPU, blended on GPU [full]
23,893583 [dev_pixelpipe] took 0,704 secs (0,702 CPU) processed `exposure 1' on GPU, blended on GPU [full]
24,136845 [dev_pixelpipe] took 0,242 secs (0,395 CPU) processed `lens correction' on GPU, blended on GPU [full]
24,257331 [dev_pixelpipe] took 0,119 secs (0,056 CPU) processed `orientation' on GPU, blended on GPU [full]
24,369267 [dev_pixelpipe] took 0,107 secs (0,047 CPU) processed `crop and rotate' on GPU, blended on GPU [full]
25,675875 [dev_pixelpipe] took 1,302 secs (8,598 CPU) processed `tone equalizer' on CPU, blended on CPU [full]
25,741823 [dev_pixelpipe] took 0,061 secs (0,056 CPU) processed `input color profile' on GPU, blended on GPU [full]
[...]

PR #4322, opened 2020-02-18
history MD5 hash calculation

phweyland:
Start for #4313 

History hash calculation
Display of hash on image information (just for debug)
Time displayed on console (hovering images)

```
hash time 0,000009
get history plus hash time 0,000957
hash time 0,000008
get history plus hash time 0,000557
hash time 0,000019
get history plus hash time 0,000614
hash time 0,000032
get history plus hash time 0,002047
hash time 0,000035
get history plus hash time 0,002301
hash time 0,000032
get history plus hash time 0,002467
```

I've added a routine to get the history copying some pieces of code without understanding every steps. Creating a dev context just to get history seems very heavy.
If hash calculation is made on dev->history when available this is not an issue. Otherwise I need to go deeper to determine the minimum necessary to get whatever image history.

### DIFF
--- a/src/common/database.c
+++ b/src/common/database.c
@@ -1418,7 +1418,8 @@ static int _upgrade_library_schema_step(dt_database_t *db, int version)
   else if(version == 23)
   {
     sqlite3_exec(db->handle, "BEGIN TRANSACTION", NULL, NULL, NULL);
-    TRY_EXEC("CREATE TABLE main.history_hash (imgid INTEGER PRIMARY KEY, initial BLOB, current BLOB)",
+    TRY_EXEC("CREATE TABLE main.history_hash (imgid INTEGER PRIMARY KEY, "
+             "initial_hash BLOB, default_hash BLOB, current_hash BLOB)",
              "[init] can't create table history_hash\n");
     sqlite3_exec(db->handle, "COMMIT", NULL, NULL, NULL);
 
@@ -1705,7 +1706,8 @@ static void _create_library_schema(dt_database_t *db)
 
   sqlite3_exec(db->handle, "CREATE TABLE main.module_order (imgid INTEGER PRIMARY KEY, version INTEGER, iop_list VARCHAR)",
                NULL, NULL, NULL);
-  sqlite3_exec(db->handle, "CREATE TABLE main.history_hash (imgid INTEGER PRIMARY KEY, initial BLOB, current BLOB)",
+  sqlite3_exec(db->handle, "CREATE TABLE main.history_hash (imgid INTEGER PRIMARY KEY, "
+               "initial_hash BLOB, default_hash BLOB, current BLOB)",
                NULL, NULL, NULL);
 }
 
--- a/src/common/database.c
+++ b/src/common/database.c
@@ -1421,7 +1421,24 @@ static int _upgrade_library_schema_step(dt_database_t *db, int version)
     TRY_EXEC("CREATE TABLE main.history_hash (imgid INTEGER PRIMARY KEY, "
              "initial_hash BLOB, default_hash BLOB, current_hash BLOB)",
              "[init] can't create table history_hash\n");
+
+    const gsize checksum_len = g_checksum_type_get_length(G_CHECKSUM_MD5);
+    guint8 *checksum = calloc(1, checksum_len);
+    sqlite3_stmt *ins_stmt = NULL;
+    char *query = dt_util_dstrcat(NULL, "INSERT INTO main.history_hash "
+                                        "SELECT DISTINCT id, "
+                                        "CASE WHEN imgid IS NULL THEN ?1 ELSE NULL END, "
+                                        "NULL, ?1"
+                                        "FROM main.images LEFT JOIN main.history ON imgid = id"
+                                        );
+    sqlite3_prepare_v2(db->handle, query,-1, &ins_stmt, NULL);
+    sqlite3_bind_blob(ins_stmt, 1, &checksum, checksum_len, NULL);
+    TRY_STEP(ins_stmt, SQLITE_DONE, "[init] can't initialize main.history_hash\n");
+    sqlite3_finalize(ins_stmt);
+
     sqlite3_exec(db->handle, "COMMIT", NULL, NULL, NULL);
+    g_free(query);
+    g_free(checksum);
 
     new_version = 24;
   }
@@ -1706,8 +1723,8 @@ static void _create_library_schema(dt_database_t *db)
 
   sqlite3_exec(db->handle, "CREATE TABLE main.module_order (imgid INTEGER PRIMARY KEY, version INTEGER, iop_list VARCHAR)",
                NULL, NULL, NULL);
-  sqlite3_exec(db->handle, "CREATE TABLE main.history_hash (imgid INTEGER PRIMARY KEY, "
-               "initial_hash BLOB, default_hash BLOB, current BLOB)",
+  sqlite3_exec(db->handle, "CREATE TABLE main.history_hash (imgid INTEGER PRIMARY KEY, status INTEGER, "
+               "initial_hash BLOB, default_hash BLOB, current_hash BLOB)",
                NULL, NULL, NULL);
 }
 
--- a/src/common/database.c
+++ b/src/common/database.c
@@ -27,6 +27,7 @@
 #include "common/file_location.h"
 #include "common/iop_order.h"
 #include "common/styles.h"
+#include "common/history.h"
 #include "control/conf.h"
 #include "control/control.h"
 #include "gui/legacy_presets.h"
@@ -1419,34 +1420,107 @@ static int _upgrade_library_schema_step(dt_database_t *db, int version)
   {
     sqlite3_exec(db->handle, "BEGIN TRANSACTION", NULL, NULL, NULL);
     TRY_EXEC("CREATE TABLE main.history_hash (imgid INTEGER PRIMARY KEY, "
-             "initial_hash BLOB, auto_hash BLOB, current_hash BLOB)",
+             "basic_hash BLOB, auto_hash BLOB, current_hash BLOB)",
              "[init] can't create table history_hash\n");
 
     // use the former dt_image_altered() to initialise the history_hash table
-    // insert an history_hash entry for all images (including non edited images)
+    // insert an history_hash entry for all images which have an history
+    // note that images without history don't get hash and are considered as basic
+    sqlite3_stmt *h_stmt;
     const gboolean basecurve_auto_apply = dt_conf_get_bool("plugins/darkroom/basecurve/auto_apply");
     const gboolean sharpen_auto_apply = dt_conf_get_bool("plugins/darkroom/sharpen/auto_apply");
-    const gsize checksum_len = g_checksum_type_get_length(G_CHECKSUM_MD5);
-    guint8 *checksum = calloc(1, checksum_len);
-    sqlite3_stmt *ins_stmt = NULL;
-    char *query = dt_util_dstrcat(NULL, "INSERT INTO main.history_hash "
-                                        "SELECT DISTINCT id, "
-                                        "CASE WHEN imgid IS NULL THEN ?1 ELSE NULL END, "
-                                        "NULL, ?1"
-                                        "FROM main.images LEFT JOIN main.history ON imgid = id"
-                                        " AND num < history_end AND enabled = 1"
-                                        " AND operation NOT IN ('flip', 'dither', 'highlights', 'rawprepare',"
-                                        "     'colorin', 'colorout', 'gamma', 'demosaic', 'temperature'%s%s)",
-                                        basecurve_auto_apply ? ", 'basecurve'" : "",
-                                        sharpen_auto_apply ? ", 'sharpen'" : "");
-    sqlite3_prepare_v2(db->handle, query,-1, &ins_stmt, NULL);
-    sqlite3_bind_blob(ins_stmt, 1, &checksum, checksum_len, NULL);
-    TRY_STEP(ins_stmt, SQLITE_DONE, "[init] can't initialize main.history_hash\n");
-    sqlite3_finalize(ins_stmt);
+    char *query = NULL;
+    query = dt_util_dstrcat(query,
+                            "SELECT id, CASE WHEN imgid IS NULL THEN 0 ELSE 1 END as altered "
+                            // first, images which are both in images and history (avoids history orphans)
+                            "FROM (SELECT DISTINCT id FROM main.images JOIN main.history ON imgid = id) "
+                            "LEFT JOIN (SELECT DISTINCT imgid FROM main.images JOIN main.history ON imgid = id "
+                            "           WHERE num < history_end AND enabled = 1"
+                            "             AND operation NOT IN ('flip', 'dither', 'highlights', 'rawprepare', "
+                            "             'colorin', 'colorout', 'gamma', 'demosaic', 'temperature'%s%s)) "
+                            "ON imgid = id",
+                             basecurve_auto_apply ? ", 'basecurve'" : "",
+                             sharpen_auto_apply ? ", 'sharpen'" : "");
+    TRY_PREPARE(h_stmt, query,
+                "[init] can't prepare selecting history for history_hash migration\n");
+    while(sqlite3_step(h_stmt) == SQLITE_ROW)
+    {
+      const int32_t imgid= sqlite3_column_int(h_stmt, 0);
+      const int32_t altered = sqlite3_column_int(h_stmt, 1);
+
+      guint8 *hash = NULL;
+      GChecksum *checksum = g_checksum_new(G_CHECKSUM_MD5);
+      gsize hash_len = 0;
+
+      // get history
+      sqlite3_stmt *h2_stmt;
+      sqlite3_prepare_v2(db->handle,
+                         "SELECT operation, op_params, blendop_params"
+                         " FROM main.history"
+                         " WHERE imgid = ?1 AND enabled = 1"
+                         " ORDER BY num",
+                         -1, &h2_stmt, NULL);
+      sqlite3_bind_int(h2_stmt, 1, imgid);
+      while(sqlite3_step(h2_stmt) == SQLITE_ROW)
+      {
+        // operation
+        char *buf = (char *)sqlite3_column_text(h2_stmt, 0);
+        if(buf) g_checksum_update(checksum, (const guchar *)buf, -1);
+        // op_params
+        buf = (char *)sqlite3_column_blob(h2_stmt, 1);
+        int params_len = sqlite3_column_bytes(h2_stmt, 1);
+        if(buf) g_checksum_update(checksum, (const guchar *)buf, params_len);
+        // blendop_params
+        buf = (char *)sqlite3_column_blob(h2_stmt, 2);
+        params_len = sqlite3_column_bytes(h2_stmt, 2);
+        if(buf) g_checksum_update(checksum, (const guchar *)buf, params_len);
+      }
+      sqlite3_finalize(h2_stmt);
+
+      // get module order
+      h2_stmt = NULL;
+      sqlite3_prepare_v2(db->handle,
+                         "SELECT version, iop_list"
+                         " FROM main.module_order"
+                         " WHERE imgid = ?1",
+                         -1, &h2_stmt, NULL);
+      sqlite3_bind_int(h2_stmt, 1, imgid);
+      if(sqlite3_step(h2_stmt) == SQLITE_ROW)
+      {
+        const int version_h = sqlite3_column_int(h2_stmt, 0);
+        g_checksum_update(checksum, (const guchar *)&version_h, sizeof(version_h));
+        if(version_h == DT_IOP_ORDER_CUSTOM)
+        {
+          // iop_list
+          const char *buf = (char *)sqlite3_column_text(h2_stmt, 1);
+          if(buf) g_checksum_update(checksum, (const guchar *)buf, -1);
+        }
+      }
+      sqlite3_finalize(h2_stmt);
+
+      const gsize checksum_len = g_checksum_type_get_length(G_CHECKSUM_MD5);
+      hash = g_malloc(checksum_len);
+      hash_len = checksum_len;
+      g_checksum_get_digest(checksum, hash, &hash_len);
+
+      g_checksum_free(checksum);
+      // insert the hash for that image
+      h2_stmt = NULL;
+      sqlite3_prepare_v2(db->handle,
+                         "INSERT INTO main.history_hash"
+                         " VALUES (?1, ?2, NULL, ?3)",
+                         -1, &h2_stmt, NULL);
+      sqlite3_bind_int(h2_stmt, 1, imgid);
+      sqlite3_bind_blob(h2_stmt, 2, altered ? NULL : hash, altered ? 0 : hash_len, SQLITE_TRANSIENT);
+      sqlite3_bind_blob(h2_stmt, 3, hash, hash_len, SQLITE_TRANSIENT);
+      TRY_STEP(h2_stmt, SQLITE_DONE, "[init] can't insert into history_hash\n");
+      sqlite3_finalize(h2_stmt);
+      g_free(hash);
+    }
+    sqlite3_finalize(h_stmt);
+    g_free(query);
 
     sqlite3_exec(db->handle, "COMMIT", NULL, NULL, NULL);
-    g_free(query);
-    g_free(checksum);
 
     new_version = 24;
   }
@@ -1732,7 +1806,7 @@ static void _create_library_schema(dt_database_t *db)
   sqlite3_exec(db->handle, "CREATE TABLE main.module_order (imgid INTEGER PRIMARY KEY, version INTEGER, iop_list VARCHAR)",
                NULL, NULL, NULL);
   sqlite3_exec(db->handle, "CREATE TABLE main.history_hash (imgid INTEGER PRIMARY KEY, "
-               "initial_hash BLOB, auto_hash BLOB, current_hash BLOB)",
+               "basic_hash BLOB, auto_hash BLOB, current_hash BLOB)",
                NULL, NULL, NULL);
 }
 
--- a/src/common/database.c
+++ b/src/common/database.c
@@ -1418,7 +1418,7 @@ static int _upgrade_library_schema_step(dt_database_t *db, int version)
   else if(version == 23)
   {
     sqlite3_exec(db->handle, "BEGIN TRANSACTION", NULL, NULL, NULL);
-    TRY_EXEC("CREATE TABLE main.history_hash (imgid INTEGER PRIMARY KEY, initial CHAR(40), current CHAR(40))",
+    TRY_EXEC("CREATE TABLE main.history_hash (imgid INTEGER PRIMARY KEY, initial BLOB, current BLOB)",
              "[init] can't create table history_hash\n");
     sqlite3_exec(db->handle, "COMMIT", NULL, NULL, NULL);
 
@@ -1705,7 +1705,7 @@ static void _create_library_schema(dt_database_t *db)
 
   sqlite3_exec(db->handle, "CREATE TABLE main.module_order (imgid INTEGER PRIMARY KEY, version INTEGER, iop_list VARCHAR)",
                NULL, NULL, NULL);
-  sqlite3_exec(db->handle, "CREATE TABLE main.history_hash (imgid INTEGER PRIMARY KEY, initial CHAR(40), current CHAR(40))",
+  sqlite3_exec(db->handle, "CREATE TABLE main.history_hash (imgid INTEGER PRIMARY KEY, initial BLOB, current BLOB)",
                NULL, NULL, NULL);
 }
 
--- a/src/common/database.c
+++ b/src/common/database.c
@@ -44,7 +44,7 @@
 
 // whenever _create_*_schema() gets changed you HAVE to bump this version and add an update path to
 // _upgrade_*_schema_step()!
-#define CURRENT_DATABASE_VERSION_LIBRARY 23
+#define CURRENT_DATABASE_VERSION_LIBRARY 24
 #define CURRENT_DATABASE_VERSION_DATA     5
 
 typedef struct dt_database_t
@@ -1415,6 +1415,15 @@ static int _upgrade_library_schema_step(dt_database_t *db, int version)
 
     new_version = 23;
   }
+  else if(version == 23)
+  {
+    sqlite3_exec(db->handle, "BEGIN TRANSACTION", NULL, NULL, NULL);
+    TRY_EXEC("CREATE TABLE main.history_hash (imgid INTEGER PRIMARY KEY, initial CHAR(40), current CHAR(40))",
+             "[init] can't create table history_hash\n");
+    sqlite3_exec(db->handle, "COMMIT", NULL, NULL, NULL);
+
+    new_version = 24;
+  }
   else
     new_version = version; // should be the fallback so that calling code sees that we are in an infinite loop
 
@@ -1695,6 +1704,8 @@ static void _create_library_schema(dt_database_t *db)
   sqlite3_exec(db->handle, "CREATE INDEX main.metadata_index ON meta_data (id, key)", NULL, NULL, NULL);
 
   sqlite3_exec(db->handle, "CREATE TABLE main.module_order (imgid INTEGER PRIMARY KEY, version INTEGER, iop_list VARCHAR)",
+               NULL, NULL, NULL);
+  sqlite3_exec(db->handle, "CREATE TABLE main.history_hash (imgid INTEGER PRIMARY KEY, initial CHAR(40), current CHAR(40))",
                NULL, NULL, NULL);
 }
 
--- a/src/common/database.c
+++ b/src/common/database.c
@@ -1419,9 +1419,13 @@ static int _upgrade_library_schema_step(dt_database_t *db, int version)
   {
     sqlite3_exec(db->handle, "BEGIN TRANSACTION", NULL, NULL, NULL);
     TRY_EXEC("CREATE TABLE main.history_hash (imgid INTEGER PRIMARY KEY, "
-             "initial_hash BLOB, default_hash BLOB, current_hash BLOB)",
+             "initial_hash BLOB, auto_hash BLOB, current_hash BLOB)",
              "[init] can't create table history_hash\n");
 
+    // use the former dt_image_altered() to initialise the history_hash table
+    // insert an history_hash entry for all images (including non edited images)
+    const gboolean basecurve_auto_apply = dt_conf_get_bool("plugins/darkroom/basecurve/auto_apply");
+    const gboolean sharpen_auto_apply = dt_conf_get_bool("plugins/darkroom/sharpen/auto_apply");
     const gsize checksum_len = g_checksum_type_get_length(G_CHECKSUM_MD5);
     guint8 *checksum = calloc(1, checksum_len);
     sqlite3_stmt *ins_stmt = NULL;
@@ -1430,7 +1434,11 @@ static int _upgrade_library_schema_step(dt_database_t *db, int version)
                                         "CASE WHEN imgid IS NULL THEN ?1 ELSE NULL END, "
                                         "NULL, ?1"
                                         "FROM main.images LEFT JOIN main.history ON imgid = id"
-                                        );
+                                        " AND num < history_end AND enabled = 1"
+                                        " AND operation NOT IN ('flip', 'dither', 'highlights', 'rawprepare',"
+                                        "     'colorin', 'colorout', 'gamma', 'demosaic', 'temperature'%s%s)",
+                                        basecurve_auto_apply ? ", 'basecurve'" : "",
+                                        sharpen_auto_apply ? ", 'sharpen'" : "");
     sqlite3_prepare_v2(db->handle, query,-1, &ins_stmt, NULL);
     sqlite3_bind_blob(ins_stmt, 1, &checksum, checksum_len, NULL);
     TRY_STEP(ins_stmt, SQLITE_DONE, "[init] can't initialize main.history_hash\n");
@@ -1723,8 +1731,8 @@ static void _create_library_schema(dt_database_t *db)
 
   sqlite3_exec(db->handle, "CREATE TABLE main.module_order (imgid INTEGER PRIMARY KEY, version INTEGER, iop_list VARCHAR)",
                NULL, NULL, NULL);
-  sqlite3_exec(db->handle, "CREATE TABLE main.history_hash (imgid INTEGER PRIMARY KEY, status INTEGER, "
-               "initial_hash BLOB, default_hash BLOB, current_hash BLOB)",
+  sqlite3_exec(db->handle, "CREATE TABLE main.history_hash (imgid INTEGER PRIMARY KEY, "
+               "initial_hash BLOB, auto_hash BLOB, current_hash BLOB)",
                NULL, NULL, NULL);
 }
 
--- a/src/common/exif.cc
+++ b/src/common/exif.cc
@@ -274,7 +274,7 @@ const char *dt_xmp_keys[]
         "Xmp.darktable.mask_id", "Xmp.darktable.mask_type", "Xmp.darktable.mask_name",
         "Xmp.darktable.masks_history", "Xmp.darktable.mask_num", "Xmp.darktable.mask_points",
         "Xmp.darktable.mask_version", "Xmp.darktable.mask", "Xmp.darktable.mask_nb", "Xmp.darktable.mask_src",
-        "Xmp.darktable.history_initial_hash", "Xmp.darktable.history_auto_hash", "Xmp.darktable.history_current_hash",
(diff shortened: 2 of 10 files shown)


phweyland:
> Otherwise I need to go deeper to determine the minimum necessary to get whatever image history

So far I understand that the history is available in db except the iop_order.
Then some additional calculation must be done to complete the iop_order for each active module. Is that correct ?

TurboGit:
> Creating a dev context just to get history seems very heavy.

Yes, I would certainly write a dedicated SQL query to get the enabled module only on the history with the needed information. This routine would return a GList*.

`SELECT operation, op_params, blendop_params FROM history WHERE enabled = 1 AND imgid = ?1 ORDER BY num;`

Note the `ORDER BY num` which ensure that the modules are always compared in the right order. I don't think we need the multi_priority. After all if you duplicate an instance, delete the original one you still have the same development, no?

TurboGit:
> So far I understand that the history is available in db except the iop_order.

Right, we need to include into the hash the version and iop_list from module_order I suppose. This with the history as proposed above should be enough to detect an altered status.

phweyland:
> After all if you duplicate an instance, delete the original one you still have the same development, no?

Hummm, multi multipriority = multi instance.
If so, each instance, added or removed changes the image, doesn't it ?

TurboGit:
> If so, each instance, added or removed changes the image, doesn't it ?

But without this value in the history you'll have one or two instance. The hash will be different anyway.

phweyland:
> Right, we need to include into the hash the version and iop_list from module_order I suppose. This with the history as proposed above should be enough to detect an altered status.

OK, that would be a way, instead of going through the iop_order calculation.

phweyland:
> > Right, we need to include into the hash the version and iop_list from module_order I suppose. This with the history as proposed above should be enough to detect an altered status.
> 
> OK, that would be a way, instead of going through the iop_order calculation.

I don't think that's a good idea ... most of the work will be done on dev phase (on dev->history).
So to make the comparison we need to calculate the checksum on the same structure.

If that's correct, whatever I do, I need to mimic exactly the dev_history list (at least with the needed parameters).

phweyland:
> > If so, each instance, added or removed changes the image, doesn't it ?
> 
> But without this value in the history you'll have one or two instance. The hash will be different anyway.

Right.

TurboGit:
> > 
> 
> I don't think that's a good idea ... most of the work will be done on dev phase (on dev->history).
> So to make the comparison we need to calculate the checksum on the same structure.

You don't need that (or not required). From lightable you don't have a dev struct and from the darkroom the calculation would be done after writing the history so all is in the DB.

TurboGit:
My idea is that the hash must be *content oriented* and not *topology oriented*.

phweyland:
> the calculation would be done after writing the history so all is in the DB

In that case that's ok.

src/common/history.c
```diff
@@ -1206,6 +1206,57 @@ GList *dt_history_duplicate(GList *hist)
...
+                              "SELECT operation, op_params, enabled, blendop_params"
+                              " FROM main.history"
+                              " WHERE imgid = ?1"
+                              " ORDER BY num",
+                              -1, &stmt, NULL);
+  DT_DEBUG_SQLITE3_BIND_INT(stmt, 1, imgid);
+
+  while(sqlite3_step(stmt) == SQLITE_ROW)
+  {
+    if(sqlite3_column_int(stmt, 2))
```
TurboGit: Why not put this in the SQL statement above? Looping around all non enabled modules are time consuming for nothing.

phweyland:
Here is the direct calculation against db. More efficient than through the dev context creation but far slower than on in memory history list.

```
get history plus hash img 56054 time 0,000111
get history plus hash img 56047 time 0,000212
get history plus hash img 56047 time 0,000202
get history plus hash img 56054 time 0,000248
get history plus hash img 56061 time 0,000220
get history plus hash img 56068 time 0,000221
get history plus hash img 56075 time 0,000218
get history plus hash img 56082 time 0,000148
get history plus hash img 56083 time 0,000052
get history plus hash img 56084 time 0,000059
```

TurboGit:
> Here is the direct calculation against db. More efficient than through the dev context creation but far slower than on in memory history list.

For sure, but we can use only the dev context for darkroom where the speed (only one image at a time) is not that an issue.

Maybe with the enabled on SQL we can gain some CPU cycles?

phweyland:
> Maybe with the enabled on SQL we can gain some CPU cycles?

Already done.

Ok, I'll provide the same calculation from dev_history_list.

src/common/history.c
phweyland: Right. I'll do.

TurboGit:
> Already done.

So you did not push last version ?

phweyland:
> > Already done.
> 
> So you did not push last version ?

Sorry, my mistake. I hadn't seen your comment on the code.

TurboGit:
Anyway I should give you more time :) I'm looking at this as soon as I see a change :) So eager to see all this stuff in action :)

phweyland:
I've tried to simulate the same calculation from dev (creating a mininum context as in the first commit) but I fail on params_size calculation. 

Last commit with enabled in where condition. Maybe a bit better.

phweyland:
> I fail on params_size calculation.

By the way the time difference may be shared between db accesses and params hash I may have missed with the first commit.

phweyland:
I think we have 3 hashes to store somewhere:
- creation time / default profile - calculated
- last darkroom leave (or last user change ?) - calculated
- last thumbnail refresh - copied

To store them we could add those fields in main.images ? main.module_order ? A specific table ?

What do you think ?

Just a note: running some tests I've seen that images on which I'd copied the dev from another image share the same hash.

phweyland:
> My idea is that the hash must be **content oriented** and not **topology oriented**.

I don't understand this. Could you elaborate ?

TurboGit:
@phweyland : By **content oriented** I mean that only the content of the data is meaningful and for example we don't care about the multi_priority. The fact that expose is on a multi_priiority of 1 or 2 or whatever should not be taken into account, the important point is that the actual content (op_params and blendop_params) is used for the hash. Another way to see that is, an exposure module with a given set of parameters will have the very same effect whether it is on a multi_priority of 1 or 2... I hope I'm clear :) And I hope this is correct regarding the issue we are trying to solve.

TurboGit:
Can you tell me more about:
> last thumbnail refresh - copied

Does that means that you also want to replace the current dt_image_altered() routine ? Or do you have something else in mind?

src/common/history.c
```diff
@@ -1206,6 +1206,54 @@ GList *dt_history_duplicate(GList *hist)
...
+                              " WHERE imgid = ?1 AND enabled = 1"
+                              " ORDER BY num",
+                              -1, &stmt, NULL);
+  DT_DEBUG_SQLITE3_BIND_INT(stmt, 1, imgid);
+
+  while(sqlite3_step(stmt) == SQLITE_ROW)
+  {
+    char *buf = (char *)sqlite3_column_text(stmt, 0);
+    if(buf) g_checksum_update(checksum, (const guchar *)buf, -1);
+    buf = (char *)sqlite3_column_text(stmt, 1);
```
TurboGit: fields 1 and 2 are not text. These are blob and the size if given by reading the first bytes:

```
    const int32_t params_len = sqlite3_column_bytes(stmt, 1);
    const void *params_blob = sqlite3_column_blob(stmt, 1);
```

Then you can use:

`
   if(params_blob) g_checksum_update(checksum, (const guchar *)params_blob, params_len);`

phweyland:
> Does that means that you also want to replace the current dt_image_altered() routine ?

What does this routine ?
It returns true if it exists another module than the basic one... (I'm a bit unsure about the condition ` num<history_end`). Nearly all the time ? Is that correct ?

I think we need to refresh the thumbnail iff the image has changed since the last refresh. 
I don't think we need to refresh the thumbnails all the time.

I may have misunderstood everything ... But if I'm correct we should replace this routine too.

> Another way to see that is, an exposure module with a given set of parameters will have the very same effect whether it is on a multi_priority of 1 or 2...

understood and agreed.

src/common/history.c
phweyland: I've missed that. I'll fix it.

TurboGit: Thanks ! And then redo the timing as with this code the checksum update was ending at the first \0 found in the blob.

src/common/history.c
```diff
@@ -1206,6 +1206,58 @@ GList *dt_history_duplicate(GList *hist)
...
+                              " FROM main.history"
+                              " WHERE imgid = ?1 AND enabled = 1"
+                              " ORDER BY num",
+                              -1, &stmt, NULL);
+  DT_DEBUG_SQLITE3_BIND_INT(stmt, 1, imgid);
+
+  while(sqlite3_step(stmt) == SQLITE_ROW)
+  {
+    char *buf = (char *)sqlite3_column_text(stmt, 0);
+    int32_t params_len = sqlite3_column_bytes(stmt, 0);
```
TurboGit: No need here as operation is TEXT.

src/common/history.c
```diff
@@ -1206,6 +1206,58 @@ GList *dt_history_duplicate(GList *hist)
...
+    if(buf) g_checksum_update(checksum, (const guchar *)buf, params_len);
+    buf = (char *)sqlite3_column_text(stmt, 1);
```
TurboGit: This one should be sqlite3_column_blob() and not sqlite3_column_text()

src/common/history.c
```diff
@@ -1206,6 +1206,58 @@ GList *dt_history_duplicate(GList *hist)
...
+    params_len = sqlite3_column_bytes(stmt, 1);
+    if(buf) g_checksum_update(checksum, (const guchar *)buf, params_len);
+    buf = (char *)sqlite3_column_text(stmt, 2);
```
TurboGit: Likewise _text -> _blob

src/common/history.c
```diff
@@ -1206,6 +1206,58 @@ GList *dt_history_duplicate(GList *hist)
...
+                              -1, &stmt, NULL);
+  DT_DEBUG_SQLITE3_BIND_INT(stmt, 1, imgid);
+  if(sqlite3_step(stmt) == SQLITE_ROW)
+  {
+    const int version = sqlite3_column_int(stmt, 0);
+    g_checksum_update(checksum, (const guchar *)&version, sizeof(version));
+    if(version == DT_IOP_ORDER_CUSTOM)
+    {
+      const char *buf = (char *)sqlite3_column_text(stmt, 1);
+      const int32_t params_len = sqlite3_column_bytes(stmt, 1);
```
TurboGit: Not needed, this is really TEXT !

src/common/history.c
phweyland: my bad ...

src/common/history.c
phweyland: all wrong ...

phweyland:
I hope I've fixed these stupid mistakes ... :(

Added a table history_hash. This can be changed if you think we should of course.
I've kept the hash as a string but it could set as an hexa array. Not sure we would save a lot.
Updating the table when the user leaves an image in darkroom.

phweyland:
Thinking loud.
I see 3 hashes:
- neutral (default module at default state)
- auto-applied
- current

Based on the comparison of hashes we should be able to split the images in these 3 categories. Remain to make some choices for images which have been created before (default rule using the current altered-not altered status).

The common action is to update the current hash on any action (darktable leave image, partial copy of history, undo, applying style, ...)

The neutral hash should be set (along with current) when
- import without xmp before auto-apply
- update of legacy_params if neutral = current

The auto-apply hash should be set
- when auto-apply if any

Remain import with xmp or full copy of history
- if xmp contains hashes just copy
- if not, default rule

From what I've seen so far these hashes should be updated in sync with module_order. So I think the best place for them should be the module_order table.

The dt_write_iop_order_list() could hold the current hash update.

Maybe, if we are able to add a context parameter to this routine we could cover the set/update of the 3 values.

Thoughts ?

phweyland:
blendop_params & blendop_version are not initialized for default modules.
Then if one opens an image in darkroom and exits the hash changes... (the params and blendop_params versions are not managed yet).

phweyland:
- in database fulfill history_hash table with altered / not altered status from history table
- collection / history and dt_image_altered() connected to history_hash (same source now)
- capture better initial / default / altered status
- capture legacy params (not tested)

TurboGit:
@phweyland : I would not save the hash in the XMP. Why is that needed? It can always be computed after importing and this is fast.

src/common/history.c
```diff
@@ -1205,6 +1205,175 @@ GList *dt_history_duplicate(GList *hist)
...
+      values = dt_util_dstrcat(values, "?2,");
+      conflict = dt_util_dstrcat(conflict, "default_hash=?2,");
+    }
+    if(type & DT_HH_CURRENT)
+    {
+      fields = dt_util_dstrcat(fields, "%s,", "current_hash");
+      values = dt_util_dstrcat(values, "?2,");
+      conflict = dt_util_dstrcat(conflict, "current_hash=?2,");
+    }
+    if(fields) fields[strlen(fields) - 1] = '\0';
```
TurboGit: To remove the comas? Worth a comment I would say.

src/common/history.h
```diff
@@ -25,6 +25,14 @@
 struct dt_develop_t;
 struct dt_iop_module_t;
 
+typedef enum dt_hash_history_t
+{
+  DT_HH_INITIAL = 1 << 0,
```
TurboGit: I think they are all exclusive, right ? So 0, 1, 2, 3 would have been ok too. Here it looks like as we would use bit-wise operators.

src/common/history.h
```diff
@@ -76,6 +84,15 @@ char *dt_history_get_items_as_string(int32_t imgid);
 /* check if a module exists in the history of corresponding image */
 gboolean dt_history_check_module_exists(int32_t imgid, const char *operation);
 
+/** calculate the hash for that history from db*/
+gsize dt_hash_history_compute_from_db(const int32_t imgid, guint8 **hash);
```
TurboGit: convention is that all routine in one module just start with dt_<module>_ so here it should all be:
`dt_history_hash...();`

For the three new routines.

src/common/history.h
```diff
@@ -25,6 +25,14 @@
 struct dt_develop_t;
 struct dt_iop_module_t;
 
+typedef enum dt_hash_history_t
+{
+  DT_HH_INITIAL = 1 << 0,
+  DT_HH_DEFAULT = 1 << 1,
+  DT_HH_CURRENT = 1 << 2,
+  DT_HH_UNKNOWN = 1 << 3
+} dt_hash_history_t;
```
TurboGit: type should be : dt_history_hash_t (see below)

src/views/darkroom.c
```diff
@@ -30,6 +30,7 @@
 #include "common/styles.h"
 #include "common/tags.h"
 #include "common/undo.h"
+#include "common/history.h"
```
TurboGit: looks superfluous ?

src/common/image.c
```diff
@@ -937,35 +939,19 @@ void dt_image_remove(const int32_t imgid)
...
-
-  DT_DEBUG_SQLITE3_PREPARE_V2(dt_database_get(darktable.db), query, -1, &stmt, NULL);
-  DT_DEBUG_SQLITE3_BIND_INT(stmt, 1, imgid);
-
-  const gboolean altered = (sqlite3_step(stmt) == SQLITE_ROW);
-  sqlite3_finalize(stmt);
-
-  return altered;
+  dt_hash_history_t status = dt_hash_history_get_status(imgid);
+  return status & DT_HH_CURRENT;
```
TurboGit: with the suggested changes in enum this would just be:
`
return status == DT_HH_CURRENT;`

src/common/history.h
TurboGit: And BTW, as this is in a spec maybe more meaningful name: DT_HISTORY_HASH_...

phweyland:
> @phweyland : I would not save the hash in the XMP. Why is that needed? It can always be computed after importing and this is fast.
> 
> EDIT: I was reacting about your previous message ([#4322 (comment)](https://github.com/darktable-org/darktable/pull/4322#issuecomment-589070414)) but it seems that this is not done. So all good.

It is still to be done. The current_hash is not worth in itself I agree. But the initial and default ones must be saved to know if the image is altered or not (the xmp is saved even if the image is not altered).

src/common/history.c
phweyland: Yes. I'll do.

src/common/history.h
phweyland: no, we can call the hash_history_write() asking for writing whatever hashes at once (usually default and current or initial and current), so the setting needs combination.
Currently I use the same enum for as setting and status. So I can set an other enum for the status (like you suggest). Is that you would like ?

src/common/history.h
phweyland: There is a dt_history_xxx, dt_dev_history_xxx, it's why I've added a dy_hash_history_xxx.

For the routine called only in one module I've seen that `dt_` prefix is replaced by `_`. So if that the subject of the remark, I should rename `dt_hash_history_compute_from_db()` to `_hash_history_compute_from_db()`. is that correct ? If not, I haven't understood the remark.

src/views/darkroom.c
phweyland: correct. I remove it.

src/common/history.h
phweyland: There is a dt_history_xxx, dt_dev_history_xxx, it's why I've added a dy_hash_history_xxx.

phweyland:
Just before committed I've added the status unknown in routine `dt_hash_history_get_status()` ... And now I've an avalanche of the status because `dt_image_altered()` is called far before the hash is created. Needs to understand why.

src/common/history.h
phweyland: It's look for me more consistent as is, but I can change. Let me know.

phweyland: done

src/common/history.h
phweyland: done, both.

phweyland:
to fulfill history_hash table I've used the collection altered/not altered rule.
But I'll change for the former `dt_image_altered()` rules. I'll use also the same to treat an xmp without hash data.
That way we should be able to recover also some default (auto) cases.

BTW I think I will rename `default` (default_hash, ...) by `auto`.

phweyland:
I've made the change for resolved comments, but I've let my current printf traces for my current tests.

phweyland:
> Just before committed I've added the status unknown in routine `dt_hash_history_get_status()` ... And now I've an avalanche of this status because `dt_image_altered()` is called far before the hash is created. Needs to understand why.

dt_view_image_expose() (refreshes mipmap) is called at the end of import process before dt_dev_read_history() (creates history). Nothing I can do about.

src/common/history.h
TurboGit: @phweyland ; Thanks, not that dt_dev_history exists indeed, but in develop.h which makes my point :)

src/common/database.c
```diff
@@ -1415,6 +1415,33 @@ static int _upgrade_library_schema_step(dt_database_t *db, int version)
...
+  {
+    sqlite3_exec(db->handle, "BEGIN TRANSACTION", NULL, NULL, NULL);
+    TRY_EXEC("CREATE TABLE main.history_hash (imgid INTEGER PRIMARY KEY, "
+             "initial_hash BLOB, default_hash BLOB, current_hash BLOB)",
+             "[init] can't create table history_hash\n");
+
+    const gsize checksum_len = g_checksum_type_get_length(G_CHECKSUM_MD5);
+    guint8 *checksum = calloc(1, checksum_len);
+    sqlite3_stmt *ins_stmt = NULL;
+    char *query = dt_util_dstrcat(NULL, "INSERT INTO main.history_hash "
```
TurboGit: The LEFT JOIN may look surprising, I would add a comment:

// Insert an history_hash entry for all images with some history already set (skip non edited images)

src/common/history.h
TurboGit: Fine with me, I missed the possible combination of multiple hash so ignore this part.

Still HH does not read directly, maybe to you but people coming to the code will certainly be more informed with HISTORY_HASH instead of HH (especially in a .h where some values will be used outside of the history file).

src/common/image.c
TurboGit: So this is good given previous comment.

phweyland:
I'm close from the scope I could think of.
With my migrated db this starts to work smoothly.
With a brand new one I have 2 issues:
- removing the auto-apply doesn't work on the first import(s ?).
- more concern about the history which is not always set up ... It's set up for opened image (several times see #4232), for another one sometimes and not for the others.

src/common/database.c
phweyland: The left join is to get also images without history.

// Insert an history_hash entry for all images with some history already set (including  non edited images)

src/common/history.h
phweyland: Understood. I'll do it.

phweyland: done

src/common/database.c
phweyland: done

src/common/exif.cc
```diff
@@ -2943,6 +2945,32 @@ int dt_exif_xmp_read(dt_image_t *img, const char *filename, const int history_on
...
+        hash.current = dt_exif_xmp_decode(pos->toString().c_str(), strlen(pos->toString().c_str()),
+                                          &hash.current_len);
+      }
+      if(hash.initial || hash.auto_apply || hash.current)
+      {
+        dt_history_hash_write(img->id, &hash);
+      }
+      else
+      {
+        dt_history_hash_write_from_history(img->id, DT_HH_CURRENT); // no choice, use the history itelf
```
phweyland: I've missed to apply the former dt_image_altered() rules here

phweyland: done

phweyland:
Applied your comments .

src/develop/develop.c
```diff
@@ -1310,6 +1311,18 @@ void dt_dev_write_history(dt_develop_t *dev)
   dt_dev_write_history_ext(dev, dev->image_storage.id);
 }
 
+static int _dev_get_module_nb_records()
+{
+  sqlite3_stmt *stmt;
+  DT_DEBUG_SQLITE3_PREPARE_V2(dt_database_get(darktable.db),
+                              "SELECT count (*) FROM  memory.history",
```
TurboGit: should be COUNT(*)

I would move this into history.h and use it in many places (at least in history.c, develop.c and lightrroom.c) where we are using this exact SQL statement. Would be a nice refactoring.

TurboGit: EDIT: sorry this is counting memory not main.history so forget about this.

src/libs/collect.c
```diff
@@ -1305,18 +1306,19 @@ static void list_view(dt_lib_collect_rule_t *dr)
...
-                   " LEFT JOIN (SELECT DISTINCT imgid AS history_id, 1 AS altered"
-                   "            FROM main.history)"
-                   "   ON id = history_id"
+                   " LEFT JOIN (SELECT DISTINCT imgid, initial_hash, auto_hash, current_hash"
+                   "            FROM main.history_hash) ON id = imgid"
                    " WHERE %s"
                    " GROUP BY altered"
                    " ORDER BY altered ASC",
-                   _("altered"),  _("not altered"), where_ext);
+                    ("initial"), _("auto applied"), _("altered"), _("none"), where_ext);
```
TurboGit: Probably a missing translation here _("initial") ?

phweyland: Right !

TurboGit:
I have tested this and see some strange things:
- first after migration I have many initial_hash set to NULL
- entering in the darkrroom with an image marked as non-altered mark it as altered (icon on lighttable)

This second point is maybe related to the first one as for new edits going to darkrroom does not mark the image as altered.

phweyland:
> I have tested this and see some strange things:
> 
>     * first after migration I have many initial_hash set to NULL

There are 3 cases:
- initial = current => initial status
- auto = current => auto applied status
- else => altered

So, after migration you can have
- initial = auto = NULL and current = hash => altered
- initial = current = hash (the same) and auto = NULL => initial
We don't have the case auto applied from migration


>     * entering in the darkrroom with an image marked as non-altered mark it as altered (icon on lighttable)
> 

That's an issue indeed. This should not happen if the image has an history (in that case it gets one and it changes to altered). But with the hash logic we should not have any image without history. 

Except that ... I've not solved the problem risen yesterday. With as fresh db I import images and they don't get the history. I understand that what I've coded (dt_imageio_export_with_flags) is wrong and must be normal import thread (dt_image_import_internal) instead. But I haven't yet found the configuration point which makes work my wrong code. I suspect your imported db has not this configuration flag and works as my fresh db.

In that case what you see is normal (no history at import time).
To make work the hash system we need an history at any time.

phweyland:
The fact to have several paths to import images is not safe ...

phweyland:
>     * entering in the darkrroom with an image marked as non-altered mark it as altered (icon on lighttable)

Understood I think. Not related to problem exposed above. 
At import time I set a fake checksum. This not an issue for altered images. But this is an issue for not altered ones. For sure, passing through darkroom, the image gets a different checksum. 
Without talking about the potential lack of history. 

To make this part work we could get 3 status from existing images:
- initial (without history)
- auto apply  (with history but not altered - former rule)
- altered the remaining images

second point, I need
- to get an history for initial images
- calculate the correct checksum for all images.

phweyland:
> But I haven't yet found the configuration point which makes work my wrong code.

Understood ! With my live db, I've the setting "don't use embedded preview jpeg but half-size raw".
This forces mipmap to reload the image. This what makes work the code as is.
Now I've just to complete dt_image_import_internal() and make import work too.

phweyland:
I've changed a bit my mind (I've understood better the current dt behavior :) )

With the current dt the confusion comes from the fact that fresh raw images import can lead to 2 different status depending on "don't use embedded preview jpeg but half-size raw" setting:
- if not set, no history (to speed up the import process I guess). the image gets its history when enters in darkroom.
- if set, thumbnails update provides them immediately a basic history.

That's where I've been confused. And of course `not altered` is not always equivalent to `no history` !

So, here I've kept the option not to create history at import time ( "don't use embedded preview jpeg but half-size raw" not set).

But we consider images without history or with history and basic_hash = current _hash as basic.

BTW I've changed the wording `initial` by `basic`. We have 3 states:
- basic
- auto-applied
- altered

I've fixed the migration issue, providing a real hash where needed. BTW, is there any mean to use the history hash routines here instead of duplicating code ?

There may be a improvement to be done on migration (and import of image with xmp as well).
We could improve the logic of the former dt_image_altered(), splitting images in 3 categories (basic, auto applied & altered) instead of 2 (basic & altered).
Thoughts ?

parafin:
Why would one need to distinguish basic from auto applied category?

phweyland:
> Why would one need to distinguish basic from auto applied category?

Is your question related to db migration process or is it general ?

parafin:
I'm talking about exposed in the GUI values.

phweyland:
> I'm talking about exposed in the GUI values.

This have been exposed [here](https://github.com/darktable-org/darktable/issues/4313#issuecomment-587647140) in #4313 .

I would add that I share this standpoint. basic images are surely in a different state than auto-applied ones and may deserve a different attention.

parafin:
As you noticed, just changing "don't use embedded preview jpeg but half-size raw" option will change freshly imported images from "basic" to "auto-applied". I don't think it gives any useful information to a user.

phweyland:
> As you noticed, just changing "don't use embedded preview jpeg but half-size raw" option will change freshly imported images from "basic" to "auto-applied". I don't think it gives any useful information to a user.

That's the behavior of the current master. It seems that's still the case with that PR and I'll try to fix it. That was clearly my intention to have them in basic ones.

Thanks for pinpointing this.

parafin:
If you manage to implement it, that would indeed be useful then.

phweyland:
> If you manage to implement it, that would indeed be useful then.

After checking more carefully it is like this PR work. Thanks.

phweyland:
I've run these test cases (seem ok to me). please add any missing case.:

new db - no auto apply - no write xmp - use embedded jpeg for thumbnail

- clean up dt files and start dt
- import 5 non xmp images, exit darkroom => history = basic(5), order none(4), V3.0(1)
- opening an image and exiting darkroom shouldn't change the figures
- write xmp for those images, remove them and import them again => basic(5), order V3.0(5)
- check altered symbol in lighttable
- remove all images

set auto-apply (all), exit & enter dt again
- import 5 non xmp images, exit darkroom => history = auto applied(1), basic(4), order none(4), V3.0(1)
- write xmp for those images, remove them and import them again => auto applied(1), basic(4), order V3.0(5)
- open one of the basic and modify the auto-applied ones. Exit darkroom => altered(2) basic(3), order V3.0(5)
- write xmp for the 2 altered images and remove them. import them again => altered(2) basic(3), order V3.0(5)
- opening a basic image and exiting darkroom should move this image to auto-applied


set don't use embedded preview jpeg and replay all
(exiting dt after every setting change & clean up xmp files)

- import 5 non xmp images, exit darkroom => history = basic(5), V3.0(5)
- write xmp for those images, remove them and import them again => basic(5), order V3.0(5)
- remove all images

set auto-apply (all), exit & enter dt again
- import 5 non xmp images, exit darkroom => history = auto applied(5), V3.0(5)
- write xmp for those images, remove them and import them again => auto applied(5), order V3.0(5)
- open two images and modify them. Exit darkroom => altered(2) auto applied(3), order V3.0(5)
- write xmp for the 2 altered images and remove them. import them again => altered(2) auto applied(3), order V3.0(5)

import of raw file with old xmp.
- xmp without history => basic or auto apply depending on setting (not tested)
- xmp with basic module only => basic
- xmp with altered images => altered
- open a folder with images with xmp (112 images) altered(108), basic(4), order custom(3), none(3), v3.0(106). 
none(3) is for image where duplicate 0 doesn't exist. basic(4) is for previuos 3 plus a png.

imported db

- clean dt files, copy dt files from backup and start dt and upgrade
- I get (from 2.6) altered(25426), basic(26973) = 52399 order custom(13), legacy(48451), none(3935) = 52399 
- check the consitency of the numbers
- open one none raw image and exit darkroom. it should move to v3.0. No change on history is no auto apply. Move to auto applied otherwise.
- open one legacy image and exit darkroom, history split should not change

TurboGit:
@phweyland : Not enough time to do a test and review of this at the moment. But this is not forgotten on my side.

TurboGit:
Really nice this new version ! I found a little glitch though.

From lightttable use the change orientation of one image and the status stays as non altered but as soon as you enter and leave darkrroom the status is updated. I suppose a missing update for the orientation change from ligthttable ?

phweyland:
> Really nice this new version ! I found a little glitch though.
> 
> From lightttable use the change orientation of one image and the status stays as non altered but as soon as you enter and leave darkrroom the status is updated. I suppose a missing update for the orientation change from ligthttable ?

Yes, indeed !

phweyland:
I've let some commented lines of code (printf) in the history_hash routines. Let me know if have to remove them.

src/libs/collect.c
```diff
@@ -1304,21 +1304,22 @@ static void list_view(dt_lib_collect_rule_t *dr)
...
                    "     END as altered, 1, COUNT(*) AS count"
                    " FROM main.images AS mi"
-                   " LEFT JOIN (SELECT DISTINCT imgid, initial_hash, auto_hash, current_hash"
+                   " LEFT JOIN (SELECT DISTINCT imgid, basic_hash, auto_hash, current_hash"
                    "            FROM main.history_hash) ON id = imgid"
                    " WHERE %s"
                    " GROUP BY altered"
                    " ORDER BY altered ASC",
-                    ("initial"), _("auto applied"), _("altered"), _("none"), where_ext);
+                   _("basic"), _("auto applied"), _("altered"), _("basic"), where_ext);
```
TurboGit: 2 times basic here ? Is that expected?

TurboGit:
And yes I would remove the debug printf. Thanks.

src/libs/collect.c
phweyland: yes, here we want to retrieve both the images without history + images which have current = basic hash into the same basic group.
Now, there is another way: to have a group "never opened" (= no history) separated from basic (current = basic hash). "Never opened" group should match "order=none". This is more factual but I'm not sure that will help the end user.
Let me know if you thinks this last way is better.

johnny-bit: I'm for "never opened" being present.

parafin: "never opened" stops being factual if you have disabled embedded thumnails

phweyland: > "never opened" stops being factual if you have disabled embedded thumbnails

This setting is not what one changes everyday. If not set, "never opened" tells the truth (and suggests the user to open the images) otherwise it doesn't appear.
Would be "not developed"  better (more accurate but less suggestive) ?

johnny-bit: "not developed" seems like better alternative, because as you've said it's more accurate.

parafin: Isn't "basic" (and actually "auto applied") also "not developed"? What use-case do you have in mind? And why would you want to add a feature that only works when embedded thumbnails are enabled?

phweyland: Not sure to understand your concern.

> Isn't "basic" (and actually "auto applied") also "not developed"? What use-case do you have in mind? 

As is we have:
- basic = `no history` (= no order = no hash = not developed) + `current_hash = basic_hash` (= only mandatory modules (then developed but not altered))
- auto applied = `current_hash = auto_hash ` (then develop but not altered)
- altered = current_hash != basic_hash and current_hash != auto_hash 

The discussion here is just to decide if it is worth to split the first case. We would have then:
- not developed (appears only if enabled embedded thumbnails) potentially basic or auto-applied
- basic (mandatory module only)
- auto applied (mandatory module + auto applied ones)
- altered

> And why would you want to add a feature that only works when embedded thumbnails are enabled?

The fact is the current master makes confusion about no history and not altered in particular when disabled embedded thumbnails. The goal of this PR is to solve the inconsistencies listed in #4313 , in particular those introduced by the import optimization which doesn't create the corresponding history, either masking (basic stays basic and doesn't move to altered if no change) that for the user (current state of the PR) or giving him the necessary information (never opened or not developed) to understand what does happen. 

I agree that neither "never opened" and "not developed" are perfect and can lead to inadequate interpretation. What would your proposal ?
For the time being I'd prefer slightly the first solution because the user may not be interested by the import history optimization. If it was the case he could use the "module order" collection, where he can see the "none" case (=no history...).

parafin: My question is how each of the mentioned cases are useful when using darktable. Otherwise it's just exposing internal states and complicating the UI.

phweyland: There is still another way: create the history at import time ! In that case the former discussion has no object any more.

johnny-bit: The usage might stem from module order collection being a bit wonky due to inconsistencies around history and then history being wonky due to fact that some developments were considered developed and some not.

I have couple of usecases for history (and module order) being:
1. i want list of images i haven't developed at all (just imported)
2. i want list of images that have auto-apply presets applied since I'm most sure I messd something with them.
3. i want list of developed (or altered) images (even if that's just white balance, which previously didn't affect history/altered icon), because those are most likely candidates for export
4. i want list of images with custom module order, because those are most likelly caused by my (wonky) mouse dragging stuff and in 99% of cases those might be messed edits and i need to fix those
5. i want list of images with older moduler order, so i could probably upgrade those to new order.

all those use cases and more are covered by changes by @phweyland :)

parafin: So we have:

1. just imported
2. with just auto-applied presets
3. altered

and

a. old module order
b. new default module order
c. modified module order

I don't see here "none"/"no history"/"not developed" or whatever.

phweyland: >     1. just imported

This is not correct.

parafin: What is not correct? From user point of view there are 3 actions one can do:
1. import
2. open in darkroom
3. edit (change settings)

What else is there?

phweyland: @TurboGit 
I don't see any @parafin points which would be misaligned with the current state of the PR.

I feel he is concerned with the split of basic group into "never opened" or "not developed" and "basic" one, but this is unclear for me.

Anyway, both ways are fine for me. Please let me know how to follow up on this.

parafin: I'm just advocating for having some high-level design notes (e.g. usage scenarios we want to cover) and working at implementing them. The fact that "embedded thumbnails" preference affects history creation isn't something we want to expose to users. Simple way of thinking about this is to imagine how you would describe a feature in user documentation. How do you explain the difference between "never opened" and "basic" without it becoming overly complicated?

TurboGit: @parafin: You missed one action from user's point of view, the fact that some auto-applied presets have been triggered or not. This is really a user's oriented action as auto-presets are added by users. So to me I still see 3 possibles states from user's point of view, but those:

1. imported (with possibly some required iop set like temperature for RAW) 
2. with auto-applied presets
3. manually edited

As discussed before (IIRC) the fact that we go into darkrroom is not an action and should not be counted if there is no modification done. Now the part that may be confusing is that we are also trying to mix information about iop-order (legacy or v3.0) and this order can be set by points 2 and point 3.

I think imported state above is what you said to be "basic", right?

parafin: I'm OK with this classification, but it's still only 3 states, not 4;)

TurboGit: @parafin ; If you're referring to "never opened" and "never developed" I agree with you. This won't make sense to end user's I suppose.

@phweyland : Most of the usages I see (and let's concentrate only on this for this PR, we'll see later if we want to split status) is to select quickly the non-edited images, the ones that need some work to be done. And of course the status altered from lighttable which is the "manually edited" (in darkroom or using copy/paste or style) state described above.

I think we are at this stage currently, right?

phweyland: > I think we are at this stage currently, right?

Yes, that's the current stage (3 states).

phweyland: > I'm just advocating for having some high-level design notes (e.g. usage scenarios we want to cover) and working at implementing them. The fact that "embedded thumbnails" preference affects history creation isn't something we want to expose to users. Simple way of thinking about this is to imagine how you would describe a feature in user documentation. How do you explain the difference between "never opened" and "basic" without it becoming overly complicated?

I agree with that. Saying that at the beginning would simplify the conversation, wouldn't it ?

TurboGit: So let me test this again as it is now and merge if all goes well.

phweyland:
@TurboGit 
Maybe another issue for this pr. 
Some users have error on the INSERT ... ON CONFLICT query.
https://discuss.pixls.us/t/sqlite3-error-when-moving-from-lighttable-to-darkroom/16781

I think this is because there sqlite version is too old (<3.24).

I haven't not found any constraint in the current dt build.

There are 2 possibilities. Either come back to a code compatible with older sqlite version (multiple queries, slower). Or put the version constraint into the dt build.

What do you think ?

TurboGit:
Well, SQLite3 v3.24 has been released on 2018-06-04. So a release before that is almost two years old. That's already an issue in my point of view.

Now I'm wondering if we could replace the "ON CONFLICT" statements with "INSERT OR REPLACE" ? I'm not sure about the semantic when not all fields are specified...

phweyland:
> Now I'm wondering if we could replace the "ON CONFLICT" statements with "INSERT OR REPLACE" ? I'm not sure about the semantic when not all fields are specified...

The doc says they are equivalent. So I guess we will have the same release issue.

TurboGit:
I don't think we will have the same issue as we have "INSERT OR REPLACE"  statements in current code since long time. So the  "ON CONFLICT" is probably just a new supported syntax. to be tested

phweyland:
I've made the confusion between INSERT OR REPLACE and INSERT OR UPDATE.
Anyway I'll make a try with INSERT OR REPLACE to see if if that accepts a subset of columns.

phweyland:
>  I'll make a try with INSERT OR REPLACE to see if if that accepts a subset of columns.

Works nicely. Why to do simple when we can do complicate ???
I'll issue a PR.

phweyland:
No doesn't work.
a not null field is lost when it is not provided by INSERT OR REPLACE.

phweyland:
UPSERT (UPDATE OR INSERT) has been introduced at the same time as ON CONFLICT.

johnny-bit:
I'd favour bumping (and naming verbatimly) sqlite requirement to 3.24.

It seems that all major distros have it in proper ver based on repology: https://repology.org/project/sqlite3/versions

TurboGit:
@johnny-bit : I agree and I'll do that.

phweyland:
> It seems that all major distros have it in proper ver based on repology: https://repology.org/project/sqlite3/versions

I don't understand what tells this site ...

Some user says that it's not available on its distribution.
A third way would be to provide conditional code before/after 3.24 to keep the compatibility.
Do you want me to do that ?

johnny-bit:
> I don't understand what tells this site ...

It's just list of package versions in various distro/version. From what I see, worst offenders and eadaches will be Ubuntu 18.04, since it uses 3.22

> A third way would be to provide conditional code before/after 3.24 to keep the compatibility.
> Do you want me to do that ?

While it'd be cool, I think that your history-related work will land in "stable" release 3.2 in december, and by that time sqlite 3.24 will be 2.5 years old, so requiring "sqlite ver >= 3.24" won't be seen as an outrageous requirement ;)

PR #4326, opened 2020-02-19
fix cleanup module_order on remove image

phweyland:


phweyland:
related to #4313

TurboGit:
Indeed! Good catch.

PR #4329, opened 2020-02-20
fix module order - synch with history at import time

phweyland:
Related to #4313

### DIFF
--- a/src/develop/develop.c
+++ b/src/develop/develop.c
@@ -1567,6 +1567,9 @@ void dt_dev_read_history_ext(dt_develop_t *dev, const int imgid, gboolean no_ima
 
     //  first time we are loading the image, try to import lightroom .xmp if any
     if(dev->image_loading && first_run) dt_lightroom_import(dev->image_storage.id, dev, TRUE);
+
+    // make sure module_dev is in sync with history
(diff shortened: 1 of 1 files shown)


phweyland:
Should be better after lightroom import.

This fix should help getting more synchronized info from collection history / module order. However not sure that is enough.

TurboGit:
One question though. Adding this here now means that all imported pictures have an iop-order list so we cannot distinguish the non yet edited pictures. Right? Won't that be a nuisance for your altered work? Or maybe I'm confused... (little time at the moment to review seriously).

phweyland:
> Adding this here now means that all imported pictures have an iop-order list so we cannot distinguish the non yet edited pictures. Right?

In fact that doesn't change anything on history side, just on module_order side.
The collection / history - altered / not altered relies only on history.

But this PR makes collection / module order match with collection / history one (so far :)). Wants to answer your concern merging collection / history & module order.

phweyland:
> Won't that be a nuisance for your altered work?

I don't think so. I'm still trying to find a firm ground to strengthen the altered / not altered status. This is part of it.

TurboGit:
Ok, let's go with that then. Thanks.

PR #4330, opened 2020-02-20
fix function parameter names in declaration

rabauke:
The function declaration does not match the corresponding function implementation (and usage), which is 

    dt_pdf_image_t *dt_pdf_add_image(dt_pdf_t *pdf, const unsigned char *image, int width, int height, int bpp, int icc_id, float border)
    { ... }

This mismatch may lead to erroneous usage.


TurboGit:
Nice catch ! Thanks.

PR #4331, opened 2020-02-20
fix format string errors

rabauke:
Just a format string fixes. Why does gcc/clang not complain? Are warning levels not sensitive enough?


TurboGit:
Looks ok, and no I don't know why the compilers are silent about this.

PR #4340, opened 2020-02-21
Fixing dt crashing while importing via libgphoto.

jenshannoschwalm:
There are mem management problems in _camctl_recursive_get_previews resulting
in these symptoms after repeatedly opening the import from a camera/sd-card
1) You might not see all thumbs in the list for jpegs.
2) When closing dt afterwards you will observe 'double free or corruption (out)'
   or 'corrupted size vs. prev_size while consolidating' messages on the console.

What's behind this?

gp_file_new allocates a CameraFile struct that is used when reading the
exif & preview sections from a file on the camera/sdcard.

Can we later safely gp_file_free the CameraFile after passing the data -
basically we take the filename and the preview-image-data to generate the thumb
displayed?

This works fine if we have set the data via gp_file_set_data_and_size before.
When gp_file_free is executed it automatically frees the data buffer. This is the case
for dt_exif_get_thumbnail generated previews.

It is **not** ok if we got the data via gp_camera_file_get and later find out about
the buffers address and size via gp_file_get_data_and_size. If we gp_file_free and
later ask for the data again we have a size of 0 (so we can't have a visible thumb)
and observe the corruptions later.

I am not sure whether this is a libgphoto2 bug or we don't use it's functions in a proper way.

What is done in this 'fix'?
1) We look for a preview image via libgphoto2.
   If found we do *not* later free the CameraFile struct.
   The disadvantage is of course a still valid struct plus data behind it.

2) If step 1 went wrong and the file can be accessed via a full path from a "disk:" the thumb is generated
   via dt_exif_get_thumbnail. This allows safe gp_file_free.

3) Fixing free -> g_free in import dialog

Could someone with libgphoto expertise review?


TurboGit:
Your first comment means we are leaking memory, right?

I'm no expert on libgphoto2, but I had experienced the same issue.

jenshannoschwalm:
> Your first comment means we are leaking memory, right?

Correct. It's not that bad as we only leak the preview data of jpegs. And it does not crash :-)

But as told i could not find a way to avoid this. We could use gp_file_new_from_fd but that is only possible in some cases as we need a valid fd.

jenshannoschwalm:
Closing this as i couldn't set the branch to a clean state.

PR #4341, opened 2020-02-21
restrict scope of local variables

rabauke:
Just a minor clean-up.


TurboGit:
Maybe minor clean-up but very much appreciated! This really helps reading/understanding the code. Thanks.

PR #4349, opened 2020-02-22
use exclusively bash in all shell scripts

rabauke:
Darktable shell script files use shebang `#!/bin/sh` as well as `#!/bin/bash`. Many of the sh-scripts use features not available in a POSIX shell. In my view, only a single shell flavor should be utilized.

In stead of replacing `#!/bin/sh` by `#!/bin/bash` one could alternatively aim for full POSIX compliance  and use `#!/bin/sh`. This, however, would require considerably more rework.

### DIFF
--- a/build-doc.sh
+++ b/build-doc.sh
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/bash
 #
 # This script is a quick hack to make it easier to build the darktable documetation
 # in PDF, html-multi and html-single formats, for the user manual and LUA API.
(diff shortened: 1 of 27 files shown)


build-doc.sh
```diff
@@ -127,7 +127,7 @@ num_cpu()
                         ncpu=$(grep -c "^processor" /proc/cpuinfo)
                 elif [ -x /sbin/sysctl ]; then
                         ncpu=$(/sbin/sysctl -n hw.ncpu 2>/dev/null)
-                        if [ $? -neq 0 ]; then
```
rabauke: There is no such thing as an operator `-neq`.

TurboGit: Indeed !

TurboGit:
Fine with me as bash is really a good shell to work with. A standard shell is far too limited and cryptic in many cases.

johnny-bit:
So we're free to do total bashisms now in scipts, not minding sh limits? YAY! (:wink:)

PR #4351, opened 2020-02-23
Enable highlights reconstruction as default for raw and sraw images

jenshannoschwalm:
The highlights reconstruction module should be enabled for all
'dt_image_is_rawprepare_supported' images per default.

As we use this module also for other images the enable button should not
be diabled via 'module->hide_enable_button'

This is not an important fix but the UI is somewhat confusion for non-raws
so far. Also at least for sraw images enabling is much better.

There has already been #3810 discussing this topic.


TurboGit:
Works for me, thanks !

TurboGit:
Merged in master and 3.0.1.

PR #4352, opened 2020-02-23
Minimal (faking) support for dual pixel dng files, second attempt

jenshannoschwalm:
A second attempt to give minimal support for dual pixel dng files.
Ref #4032 was first attempt, see discussion there
Ref #4006 as first issue

I am well aware that dual pixel images allow more functionality than what is handled
here. (And what can be done right now as the whole dt workflow is not aware of such images)

BUT: we could support basic reading and working on the image by just using the first of two pixels
so at least those images are somewhat usable in dt.

To make the user aware of this minimalistic approach there is a warning displayed every time
you open the image in darkroom.

There is one (minor) problem remaining atm, @LebedevRI is not convinced the 4 blacklevels are
always correct for dual pixel images so right now those have to be set by hand or camera
specific style. (They can be read via exiv2 of course)

### DIFF
--- a/src/common/image.c
+++ b/src/common/image.c
@@ -1434,6 +1434,7 @@ void dt_image_init(dt_image_t *img)
   img->wb_coeffs[3] = NAN;
   img->usercrop[0] = img->usercrop[1] = 0;
   img->usercrop[2] = img->usercrop[3] = 1;
+  img->planes = 1;
   img->cache_entry = 0;
 }
 
--- a/src/common/image.h
+++ b/src/common/image.h
@@ -206,6 +206,9 @@ typedef struct dt_image_t
 
   /* DefaultUserCrop */
   float usercrop[4];
+
+  /* Keep the number of cpp for dualpixel images */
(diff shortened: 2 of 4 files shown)


TurboGit:
I'm not convinced that this is a good solution. I understand the message displayed each time, but at the end of the day you have a half-backed solution not using the whole potential of the actual information in the DNG.

So I'm not for adding this as it introduces some more complexity in the code for a minimal added value. I'd like to see what others are thinking about this.

jenshannoschwalm:
You probably meant "half-baked" but i understand your point.
ATM i see no other way to support dual pixels _somehow_ so i gave it a try. I confess that my understanding of dual-pixel images potential is **very** vague, it seems very camera/vendor specific to me and i could not find find any paper describing algos about it. Also i don't have such a camera myself and i don't intend to get one.

I will will leave this open for a while and if there are no other strong pro- opinions i will close it in a while.

jenshannoschwalm:
Closing this for now as there seems to be no interest in that

PR #4353, opened 2020-02-23
Let exif.cc be more verbose in case of errors

jenshannoschwalm:
While testing / working on 'obscure' image files i often had [exiv2] messages
on the console and had t find out by hand what was wrong.

This does not change functionality but only gives more precise information.


TurboGit:
Thanks! Merged.

Issue #4355, opened 2020-02-23
Unreadable jpeg metadata fields when including develop history


elstoc:
When exporting an image to jpeg, the 'Title' metadata field cannot be read by exiftool (command-line), libxmp (python) or flickr. When de-selecting 'develop history' in the 'edit metadata exportation' window (in the 'export selected' module), the issue disappears.

The Title field appears to be present in the file (since Geeqie displays it) but something is preventing other tools from reading it properly.

I've only been able to reproduce this issue on a single image so far and with exiftool/libxmp (possibly due to the fact that the image in question had a large history stack) but another user has also experienced this with flickr (see https://discuss.pixls.us/t/darktable-3-1-does-not-export-title-of-file/16589/3). 

Darktable version: 3.1.0+449~gc5f14d958-dirty
OS: ArchLinux
OpenCL: Issue occurs both with and without OpenCL

XMP attached

[2017-07-23-18.56.15.0_01.cr2.xmp.txt](https://github.com/darktable-org/darktable/files/4242474/2017-07-23-18.56.15.0_01.cr2.xmp.txt)

johnny-bit:
Can you check and confirm whether #4357 fixed this for you?

elstoc:
I've retested on latest master and the issue is still present

phweyland:
> I've retested on latest master and the issue is still present

The xmp file seems to be ok. Could you give us access to the faulty jpeg file ?

elstoc:
Here are two exports

The broken one (where develop history was included)
![2017-07-23-18 56 15 0_broken](https://user-images.githubusercontent.com/9555491/75551595-77037e80-5a2c-11ea-8823-8ef9a550512c.jpg)

The working one (where develop history was not included)
![2017-07-23-18 56 15 0_notbroken](https://user-images.githubusercontent.com/9555491/75551597-779c1500-5a2c-11ea-8f70-ec21ab4ae863.jpg)

phweyland:
Strange. Not sure that will help.
the first has no xmp at all.
the second one has only the metadata with no development data.

```
[...]
Exif.Image.DateTimeOriginal                  Ascii      20  2017:07:23 18:56:15
Xmp.exif.DateTimeOriginal                    XmpText    19  2017:07:23 18:56:15
Xmp.xmp.Rating                               XmpText     1  2
Xmp.xmpMM.DerivedFrom                        XmpText    25  2017-07-23-18.56.15.0.cr2
[...]

PR #4357, opened 2020-02-24
fix history export diff

phweyland:
makes consistent dt_exif_xmp_read_data_export() with dt_exif_xmp_read_data() in relation with history.

Related to #4355, but I cannot say that would solve it.

### DIFF
--- a/src/common/exif.cc
+++ b/src/common/exif.cc
@@ -3300,7 +3300,7 @@ static void dt_exif_xmp_read_data_export(Exiv2::XmpData &xmpData, const int imgi
   double longitude = NAN, latitude = NAN, altitude = NAN;
   gchar *filename = NULL;
   gchar *datetime_taken = NULL;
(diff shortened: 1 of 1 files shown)


johnny-bit:
"use of undeclared identifier 'iop_order_version'"

It seems you've forgot to initialize it in `dt_exif_xmp_read_data_export`?

phweyland:
> "use of undeclared identifier 'iop_order_version'"
> 
> It seems you've forgot to initialize it in `dt_exif_xmp_read_data_export`?

Just forgotten to test ... :(

phweyland:
> Related to #4355, but I cannot say that would solve it.

It works for me with that fix (using exiv2).

johnny-bit:
For me it fixes #4355 (exiftool)

TurboGit:
Thanks! Merged.

PR #4361, opened 2020-02-25
resort and group blend modes

rabauke:
Darktable provides a large number of different blend modes. It might be useful to group them into groups of related blend modes to guide the user. 

There are probably several meaningful ways to group the blend modes. My choice was inspired to some degree by an [article about blend modes in photoshop](https://photoblogstop.com/photoshop/photoshop-blend-modes-explained). Some remarks

- _average_ is actually identical to _normal bounded_ with opacity reduced by a factor 1/2. One might consider to mark this mode as depreciated in the future.
- _difference_ does actually not fit into any group. I see its use mainly to visualize temporarily a module's effect while editing not to be used for blending. In order to have quick access to this mode I lumped it into the _normal & difference_ group.
- _subtract_ in darktable is equivalent to _linear burn_ in photo shop, not _subtract_.

Here is how the new blend mode combo box looks like:

Bildschirmfoto von 2020-02-25 01-26-28
Bildschirmfoto von 2020-02-25 01-25-27

### DIFF
--- a/src/bauhaus/bauhaus.c
+++ b/src/bauhaus/bauhaus.c
@@ -36,7 +36,7 @@ G_DEFINE_TYPE(DtBauhausWidget, dt_bh, GTK_TYPE_DRAWING_AREA)
 
 // INNER_PADDING is the horizontal space between slider and quad
 // and vertical space between labels and slider baseline
-#define INNER_PADDING 4.0
+static const double INNER_PADDING = 4.0;
 
 // fwd declare
 static gboolean dt_bauhaus_popup_draw(GtkWidget *widget, cairo_t *cr, gpointer user_data);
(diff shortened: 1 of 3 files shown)


johnny-bit:
This is great!

Do you think you can squeeze there info from https://pixls.us/articles/darktable-3-rgb-or-lab-which-modules-help/ about blending modes which should/shouldn't go into linear/non-linear part of pipeline?

TurboGit:
I'd like to see the left and right lines a bit more closer, that is reducing the overall width of the dialog.

src/bauhaus/bauhaus.c
```diff
@@ -1684,11 +1690,14 @@ static gboolean dt_bauhaus_popup_draw(GtkWidget *widget, cairo_t *crf, gpointer
           if(entry->alignment == DT_BAUHAUS_COMBOBOX_ALIGN_LEFT)
             label_width = show_pango_text(w, context, cr, entry->label, 0, ht * k, max_width, FALSE);
```
TurboGit: 0 should probably be INNER_PADDING to avoid the left label to be close to the border of left windows.

rabauke: If revised my pull request and fixed various padding issues for bauhaus combobox. Looks much better now.

rabauke:
> Do you think you can squeeze there info from https://pixls.us/articles/darktable-3-rgb-or-lab-which-modules-help/ about blending modes which should/shouldn't go into linear/non-linear part of pipeline?

@johnny-bit: Interesting write-up. Though, I do not see how to squeeze information about which  blending modes should/shouldn't go into the linear/non-linear part of pipeline into a simple combobox in a reasonable way. One might consider to mark the problematic modes as "dangerous". But this will lead to confusion because this matter requires some explanation which goes beyond what one can do in a combobox.

johnny-bit:
I think it could be indiated via an icon representing linear/non-linear blending mode. But honeslty it's best for later time. Your current sorting is great!

src/bauhaus/bauhaus.c
TurboGit: @rabauke : The width of the combo is still very big, can this be fixed somehow? Or maybe you forgot part of your patch ?

TurboGit: Well not sure now, this kind of combobox are used in many placed in dt, so you current proposal is what seems correct. I'll review again.

rabauke: Actually, I have not changed anything in particular to adjust the total width of the combobox. The width of the combobox automatically adjusts to the width of the (right) panel. It has the same width as a slider. This behavior is observed with and without my patch and applies to all comboboxes in darktable. I you feel the combobox is too broad, you may narrow the size of your right panel.

PR #4362, opened 2020-02-25
enhance script files

rabauke:
Fixes some minor issues mainly indicated by the [shellcheck](https://www.shellcheck.net/) plugin of my IDE. There are many issues more, but I tried to keep the change list small for easier review.


TurboGit:
Thanks !

PR #4372, opened 2020-02-26
"duplicate : avoid unneeded pipe compute" commit cd97cc58 from AlicVB repository

phweyland:
Related to #4232 
Avoids to double fire the dev pipe when opening an image into darkroom.

Just copied from @AlicVB


TurboGit:
Probably worth having that in coming 3.0.1.

PR #4377, opened 2020-02-26
Fixing dt crashing while importing via libgphoto (V2)

jenshannoschwalm:
As i crashed my branch for #4340 here is a cleaner approach avoiding leaks and also fixing another double-free error. @TurboGit 

There are mem management problems in `_camctl_recursive_get_previews` resulting
in these symptoms after repeatedly opening the import from a camera/sd-card

1) You might not see all thumbs in the list for jpegs.
2) When closing dt afterwards you will observe 'double free or corruption (out)'
   or 'corrupted size vs. prev_size while consolidating' messages on the console.

What's behind this?

`gp_file_new` allocates a CameraFile struct that is used when reading the
exif & preview sections from a file on the camera/sdcard.

Can we later safely `gp_file_free` the CameraFile after passing the data -
basically we take the filename and the preview-image-data to generate the thumb
displayed?

This works fine if we have set the data via `gp_file_set_data_and_size` before.
When gp_file_free is executed it automatically frees the data buffer. This is
so far the case for `dt_exif_get_thumbnail` generated previews.

It is **not** ok if we got the data via `gp_camera_file_get` and later find out about
the buffers address and size via gp_file_get_data_and_size. If we gp_file_free and
later ask for the data again we have a size of 0 (so we can't have a visible thumb)
and observe the corruptions later.

What do we do instead?
We always allocate a proper buffer, read the specific data via `gp_camera_file_read`
and set address&size by `gp_file_set_data_and_size`. If doing so we can always safely
use gp_file_free later and avoid leaks.

libgphoto2 uses internal caching of the image files, those caches might be growing
but that's not dt's business.

### DIFF
--- a/src/common/camera_control.c
+++ b/src/common/camera_control.c
@@ -999,14 +999,20 @@ static int _camctl_recursive_get_previews(const dt_camctl_t *c, dt_camera_previe
     for(int i = 0; i < gp_list_count(files); i++)
     {
       gp_list_get_name(files, i, &filename);
-      char *file = g_build_filename(path, filename, NULL);
 
       // Lets check the type of file...
       CameraFileInfo cfi;
-      if(gp_camera_file_get_info(c->active_camera->gpcam, path, filename, &cfi, c->gpcontext) == GP_OK)
+      if(!(gp_camera_file_get_info(c->active_camera->gpcam, path, filename, &cfi, c->gpcontext) == GP_OK))
+      {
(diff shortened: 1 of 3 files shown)


TurboGit:
Working fine on my side. Hopefully this bug will be definitely fixed, thanks a lot!

chhil:
Causes build failure
https://github.com/darktable-org/darktable/issues/4391

PR #4383, opened 2020-02-27
fix metadata - allows " in the text

phweyland:
This fixes one issue described on [pixls.us](https://discuss.pixls.us/t/darktable-metadata-editor/16662?u=phweyland), ie. if the text contains the character \" it cannot be saved.


TurboGit:
Thanks.

PR #4395, opened 2020-02-28
fix double mask issue

### LINKED ISSUE #4375: reopen image with mask doubles the masks
If I draw a new mask on a fresh image and afterwards reopen the image the list of available shapes has doubled. Doing the step of reopening again causes again a doubling, the original shape appears 4 times. And so on.

 - Darktable Version: 3.1.0+745~gc044acf54
 - OS: Debian Linux
 - no OpenCL

I did a `git reset --hard` and `git pull` on the master branch as suggested in Issue #4312.

phweyland:
Hopefully this should fix #4375.

### DIFF
--- a/src/develop/develop.c
+++ b/src/develop/develop.c
@@ -1252,11 +1252,9 @@ void dt_dev_pop_history_items(dt_develop_t *dev, int32_t cnt)
   dt_control_queue_redraw_center();
 }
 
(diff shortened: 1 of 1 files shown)


TurboGit:
Fixes the issue. Merging. Thanks.

phweyland:
Thank you both @johnny-bit and @turbogit for clearing out this and found the solution! ☺️

PR #4397, opened 2020-02-28
fix memory leak

rabauke:
Fixes a memory leak and other issues.

### DIFF
--- a/src/libs/metadata.c
+++ b/src/libs/metadata.c
@@ -736,23 +736,27 @@ void *legacy_params(dt_lib_module_t *self, const void *const old_params, const s
   if(old_version == 1)
   {
     size_t new_params_size = old_params_size + 1;
-    void *new_params = calloc(sizeof(char), new_params_size);
+    char *new_params = calloc(sizeof(char), new_params_size);
 
-    char *buf = (char *)old_params;
+    const char *buf = (const char *)old_params;
 
     // <title>\0<description>\0<rights>\0<creator>\0<publisher>
-    char *metadata[DT_METADATA_NUMBER];
-    int32_t metadata_len[DT_METADATA_NUMBER];
+    const char *metadata[DT_METADATA_NUMBER];
+    size_t metadata_len[DT_METADATA_NUMBER];
     for(unsigned int i = 0; i < DT_METADATA_NUMBER; i++)
     {
       metadata[i] = buf;
-      if(!metadata[i]) return NULL;
+      if(!metadata[i])
+      {
+        free(new_params);
+        return NULL;
+      }
       metadata_len[i] = strlen(metadata[i]) + 1;
       buf += metadata_len[i];
     }
 
     // <creator>\0<publisher>\0<title>\0<description>\0<rights>
-    int pos = 0;
+    size_t pos = 0;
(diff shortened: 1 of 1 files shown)


src/libs/metadata.c
```diff
@@ -736,23 +736,27 @@ void *legacy_params(dt_lib_module_t *self, const void *const old_params, const s
...
     size_t new_params_size = old_params_size + 1;
-    void *new_params = calloc(sizeof(char), new_params_size);
+    char *new_params = calloc(sizeof(char), new_params_size);
 
-    char *buf = (char *)old_params;
+    const char *buf = (const char *)old_params;
 
     // <title>\0<description>\0<rights>\0<creator>\0<publisher>
-    char *metadata[DT_METADATA_NUMBER];
-    int32_t metadata_len[DT_METADATA_NUMBER];
```
rabauke: Elements of `metadata_len` are used as `size_t` arguments to `memcpy`.

src/libs/metadata.c
```diff
@@ -736,23 +736,27 @@ void *legacy_params(dt_lib_module_t *self, const void *const old_params, const s
...
+    const char *metadata[DT_METADATA_NUMBER];
+    size_t metadata_len[DT_METADATA_NUMBER];
     for(unsigned int i = 0; i < DT_METADATA_NUMBER; i++)
     {
       metadata[i] = buf;
-      if(!metadata[i]) return NULL;
```
rabauke: Here the memory leak occurs.

src/libs/metadata.c
```diff
@@ -736,23 +736,27 @@ void *legacy_params(dt_lib_module_t *self, const void *const old_params, const s
...
+        return NULL;
+      }
       metadata_len[i] = strlen(metadata[i]) + 1;
       buf += metadata_len[i];
     }
 
     // <creator>\0<publisher>\0<title>\0<description>\0<rights>
-    int pos = 0;
+    size_t pos = 0;
     memcpy(new_params + pos, metadata[3], metadata_len[3]);
```
rabauke: Pointer arithmetic with pointer to void is not legal C. It's a language extension of the [GNU compiler](https://gcc.gnu.org/onlinedocs/gcc/Pointer-Arith.html).

src/libs/metadata.c
```diff
@@ -736,23 +736,27 @@ void *legacy_params(dt_lib_module_t *self, const void *const old_params, const s
   if(old_version == 1)
   {
     size_t new_params_size = old_params_size + 1;
-    void *new_params = calloc(sizeof(char), new_params_size);
+    char *new_params = calloc(sizeof(char), new_params_size);
 
-    char *buf = (char *)old_params;
```
rabauke: Do not cast away constness!

src/libs/metadata.c
```diff
@@ -736,23 +736,27 @@ void *legacy_params(dt_lib_module_t *self, const void *const old_params, const s
   if(old_version == 1)
   {
     size_t new_params_size = old_params_size + 1;
-    void *new_params = calloc(sizeof(char), new_params_size);
```
rabauke: Must be pointer to `char` as we are going to deal with chars.

TurboGit:
Thanks a lot for the fix and detailed analysis !

PR #4404, opened 2020-02-29
avoid variable length array syntax

rabauke:
There is really no benefit here in using variable length arrays. Expressions as 

    float(*const mean)[2] = malloc(2 * n * sizeof(float));

are somewhat intricate. Introducing 

    typedef float float2[2];

may enhance readability. Contains also some minor clean-up.

### DIFF
--- a/src/iop/colormapping.c
+++ b/src/iop/colormapping.c
@@ -59,6 +59,8 @@ DT_MODULE_INTROSPECTION(1, dt_iop_colormapping_params_t)
 #define HISTN (1 << 11)
 #define MAXN 5
 
+typedef float float2[2];
+
 typedef enum dt_iop_colormapping_flags_t
 {
   NEUTRAL = 0,
@@ -73,8 +75,8 @@ typedef struct dt_iop_colormapping_flowback_t
 {
   float hist[HISTN];
   // n-means (max 5?) with mean/variance
-  float mean[MAXN][2];
-  float var[MAXN][2];
+  float2 mean[MAXN];
+  float2 var[MAXN];
   float weight[MAXN];
   // number of gaussians used.
   int n;
@@ -95,15 +97,15 @@ typedef struct dt_iop_colormapping_params_t
   // hist matching table for source image
   float source_ihist[HISTN];
   // n-means (max 5) with mean/variance for source image
-  float source_mean[MAXN][2];
-  float source_var[MAXN][2];
+  float2 source_mean[MAXN];
+  float2 source_var[MAXN];
   float source_weight[MAXN];
 
   // hist matching table for destination image
   int target_hist[HISTN];
   // n-means (max 5) with mean/variance for source image
-  float target_mean[MAXN][2];
-  float target_var[MAXN][2];
+  float2 target_mean[MAXN];
+  float2 target_var[MAXN];
   float target_weight[MAXN];
 } dt_iop_colormapping_params_t;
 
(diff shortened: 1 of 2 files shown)


TurboGit:
I'm not convinced by this changes. Also as we have lot of array with two indices we would need to change this kind of code all over the place. Finally for the readability I don't think it bring something.

src/iop/colormapping.c
```diff
@@ -268,7 +267,7 @@ static void get_cluster_mapping(const int n, float mi[n][2], const float wi[n],
 
 
 // inverse distant weighting according to D. Shepard's method; with power parameter 2.0
-static void get_clusters(const float *col, const int n, float mean[n][2], float *weight)
```
rabauke: Arrays in function arguments decay to pointers. This rule applies to arrays with size known at compile time as well as to variable length arrays. Thus, the above is equivalent to

    static void get_clusters(const float *col, const int n, float *mean[2], float *weight) {...}

rabauke:
> I'm not convinced by this changes. Also as we have lot of array with two indices we would need to change this kind of code all over the place. Finally for the readability I don't think it bring something.

Contrary, I would ague, that (ab)using variable length array syntax does not bring anything here because:

- There is actually not a single variable length array created. All arrays have size known at compile time or are dynamically created via `malloc`.
- Array to pointer decay applies when arrays are used in function arguments. See also my other comment on the source.

johnny-bit:
I think @aurelienpierre mentioned somewhere that avoiding pointer magic helps compiller optimize when the array size is known at compile time, and it's true. Slowly moving from pointer magic into known array size is good practice.

TurboGit:
> I think @aurelienpierre mentioned somewhere that avoiding pointer magic helps compiller optimize when the array size is known at compile time, and it's true. Slowly moving from pointer magic into known array size is good practice.

Exactly and this patch is changing that AFAIK. That's why I'm reluctant to go this path. Maybe @aurelienpierre can comment on the patch?

TurboGit:
> Possibly it has been misunderstood what this patch is about. I would like to stress that the affected code does _not_ use currently variable length arrays at all. It just employs variable length array _syntax_ in function declarations.

Right, this is in get_cluster_mapping(). What I don't like is the float2 type, I have not argument as to why I don't like it though :) Just maybe because I see this kind of `typedef float[2] float2;` for the first time!

rabauke:
> > Possibly it has been misunderstood what this patch is about. I would like to stress that the affected code does _not_ use currently variable length arrays at all. It just employs variable length array _syntax_ in function declarations.
> 
> Right, this is in get_cluster_mapping(). What I don't like is the float2 type, I have not argument as to why I don't like it though :) Just maybe because I see this kind of `typedef float float2[2]` for the first time!

@TurboGit It's just a typedef for an array of fixed size which is not uncommon and imho can help to enhance readability

    typedef float my_array[16]; // an array of 16 floats
    my_arry *p; // a pointer to an array of 16 floats

vs.

    float *p[16]; // a pointer to an array of 16 floats (or was is an array of 16 pointers to float?)

I could revise my pull request not to use this typedef if this makes it easier for your to accept the pull request.

TurboGit:
> @TurboGit It's just a typedef for an array of fixed size which is not uncommon and imho can help to enhance readability

Sure I see that :) It still looks as a strange style. But well I won't make it a blocking point. Thanks.

aurelienpierre:
> Right, this is in get_cluster_mapping(). What I don't like is the float2 type, I have not argument as to why I don't like it though :) Just maybe because I see this kind of `typedef float float2[2]` for the first time!

OpenCL deals with RGBa vectors with something very similar (if not exactly the same).

PR #4406, opened 2020-03-01
Fixing build problems with camera_control.c (again)

jenshannoschwalm:
I did a stupid wrong attempt before.

As defined in gphoto2-camera.h

```
int gp_camera_file_read		(Camera *camera, const char *folder, const char *file,
		    		 CameraFileType type,
		    		 uint64_t offset, char *buf, uint64_t *size,
		    		 GPContext *context);
```
chunksize must be uint64_t
:-(((


TurboGit:
Looks good, thanks.

PR #4408, opened 2020-03-01
Use #!/usr/bin/env bash shebang in script rather than #!/bin/bash

mazhe:
Hi,

Since commit cdc51fc12f7125f39e1f594fa98ebc73624083ff, every shell script was forced to bash. Ok, but "#!/bin/bash" shebangs can be problematic on systems that do not install bash in /bin (FreeBSD for one) or have very outdated version there that if you would rather avoid if you really need bashisms (MacOS).

Would you consider using a "#!/usr/bin/env bash" shebang to work around this issue?

Thanks a lot,


rabauke:
I thought about using `#!/usr/bin/env bash` shebang but decided against as it is commonly considered as dangerous.

parafin:
Dangerous how? Insecure PATH setting is dangerous indeed, but that doesn't mean that using PATH is insecure;)

rabauke:
> Dangerous how? Insecure PATH setting is dangerous indeed, but that doesn't mean that using PATH is insecure;)

Do not get me wrong. I am not against the proposed change if it solves a real-world problem for a nonnegligible number of users. I just wanted to clarify that I decided not to include such kind of nontrivial change in cdc51fc. Note that `env` is not a magic thing that solves all portability issues and has some side effects (that might not be of practical relevance here). `#!/usr/bin/env` is kind of a hack as this usage is not what the tool was created for.

mazhe:
There's actually another way, a bit dirty: I could track the shell call and make their command line begin by "bash ..." (but a cleaner way would surely to find/set a BASH_EXECUTABLE variable in cmake).

That might also help the window port (I'm not sure if they are doing the build through cygwin or any other filesystem-fixing layer).

johnny-bit:
I wanted to find what all the fuss is about and it's another one of those "holy wars", where each side has arguments and problems. `env` nowadays doesn't solve problem nor it ever did guarantee portability, `/bin/bash` isn't gruaranteed to be there either. The only theoretically guarnateed way is one @mazhe mentioned - use cmake to find bash executable and force it on scripts installed, but that applies only to user scripts and not to dev scripts.

As it stands, even [Google shell style guide](https://google.github.io/styleguide/shellguide.html) suggests using `/bin/bash` and be done with it.

So I'd vote to either enhance build process with finding and then setting bash path during build process (which may pose problems if build is different than destination and doesn't guarantee bash location on destination) or to simply say "go `/bin/bash` or go home" :wink:

TurboGit:
I agree with @johnny-bit  @mazhe, I would also vote to do nothing here and if one has a distrib with bash installed in a different location it is always possible to run:

`$ bash /path/to/the/script.sh`

TurboGit:
@mazhe : I would close this PR if you agree now.

mazhe:
Actually, this is not always possible: some of those scripts are called during the build phase without giving this option. This is why I asked to a least call those scripts explicitly through bash (be it fullpath provided through detection or relying on PATH) in cmake scripts.

TurboGit:
Which scripts are causing trouble during build? Maybe we can keep them sh compatible as an immediate fix?

johnny-bit:
or just use `find_program()` in cmake scripts and run them thorugh program?

mazhe:
That would be nice, there's a few of them:
- tools/create_version_c.sh
- tools/get_git_version_string.sh
- tools/get_last_commit_year.sh
- tools/makeman.sh
- tools/parse_version_c.sh

TurboGit:
> That would be nice, there's a few of them:

Ok, 5 of them and possibly compatible with standard sh already. Have you looked at them? If they are sh compatible please create a PR to use sh and add a comment on top of each file with why we keep sh there. Thanks.

mazhe:
Ho, I forgot to add that yes, I tested them in FreeBSD /bin/sh version and bash sh-compat mode, they are very straightforward and have no issue.

rabauke:
> Ok, 5 of them and possibly compatible with standard sh already. Have you looked at them? If they are sh compatible please create a PR to use sh and add a comment on top of each file with why we keep sh there. Thanks.

Just a note: [shellcheck](https://www.shellcheck.net) is very handy when it comes to testing for POSIX compatibility and other shell programming related issues. There are also shellcheck plugins for some popular editors and IDEs, e.g., for vim, Clion and Visual Studio Code.

mazhe:
Sorry for the delay, the last weeks in france have been a bit busy, here are the recommended changes. Thank you for your time and taking this into account

TurboGit:
Thanks @mazhe , looks ok to me. Merging.

Issue #4407, opened 2020-03-01
Two keywords module issues


Nilvus:
**Describe the bug**
@phweyland 
Editing keywords needs to restart darktable to have them updated.
Also, some parents do not allow to add a new keyword directly on it.

**To Reproduce**

First issue:
1. Go to 'keyword modules'
2. Right-click on a keyword
3. Click on edit keyword
4. Edit and update the name
Nothing happens. Try to right click again and click to edit keyword again. No dialog appears.

Updating is made after restarting darktable.

Second issue:
1. Right-click on a parent
2. Select 'create a keyword'
3. Default behavior is to add directly the new keyword on the selected parent. The parent path appear on the dialog window. Sometimes, darktable doesn't propose the parent path, so it create the keyword on root tree. That's need to right-click then on this new keyword then update the path.

**Expected behavior**
1. Simply update directly the updated keyword/parent
2. Fix the path problem (second issue) but quite difficult to reproduce as this issue doesn't appear on all parents.

**Platform (please complete the following information):**
 - Darktable Version: master (last compile was 29th february)
 - OS: Ubuntu 19.10

phweyland:
Thank you for reporting this. I'll look at it asap.

phweyland:
I cannot reproduce case 1, but 1. it is very slow if a lot of images attached, 2. I get console gtk error messages. I'll look at it.

About the second issue. There are indeed 2 different behaviors.
1. you have the tags:
aaa
aaa|bbb

if the user user right-clicks on aaa and selects create tag, the dialog proposes aaa as a parent.

2. but if you have only
aaa|bbb

if the user right-clicks  on aaa and selects create tag, the tag is created at root level (because aaa doesn't exist).

I think the best way in that case, is not to have the option "create tag" in the popup menu. Would you agree ?

Nilvus:
> if the user right-clicks on aaa and selects create tag, the tag is created at root level (because aaa doesn't exist).
> 
> I think the best way in that case, is not to have the option "create tag" in the popup menu. Would you agree ?

I'm really surprised that you can't reproduce first issue. It's the one I reproduce each time. So maybe a problem in my database...

About issue 2, I disagree with your proposal. It would be good to have possibility to create a tag even if there's one child. I never applied the parent on images, only child.

I make nature photos and I class them by species.

For example (If I'm not wrong, you speak french so I will write a part of my tag tree):

Faune|Papillons|Nymphalidés|

[...]

PR #4415, opened 2020-03-03
tagging fixes

phweyland:
- edit (rename) & view refresh in diverse conditions
- allow tag creation on virtual node
- remove inactive context menu entries
- some formatting clean up
- EDIT: display virtual tags in treeview in italic (consistent with category). This is a change compared to old behavior and then can be controversial. I think this is more consistent with category (not exportable) and makes visible the tags which cannot be attached (virtual). If not acceptable I'll remove this. Let me know.

Related to #4407

### DIFF
--- a/src/libs/tagging.c
+++ b/src/libs/tagging.c
@@ -65,11 +65,9 @@ typedef struct dt_lib_tagging_t
 typedef struct dt_tag_op_t
 {
   gint tagid;
-  guint count;
   char *newtagname;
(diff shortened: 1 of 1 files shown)


Nilvus:
All done, great work. Thanks, my two issues related are fixed with this PR.
Display virtual tags in italic is good for me. Thanks again for your great work on darktable.

TurboGit:
Thanks !

Issue #4420, opened 2020-03-04
3d lut cube files fail to apply with error - cube lut lines number is not correct


mproffitt:
I have been seeing this error using lut files provided with (or generated by) Davinci Resolve whereby the LUT file will not apply.

The issue appears to be due to the use of integers or doubles within the LUT .cube file where the lut code within darktable is expecting floating point?

Converting the integers/doubles to floats in the .cube files allows darktable to recognise and apply the look correctly.

This can be validated with the LUTs available from https://smallhd.com/blogs/community/movie-looks-download which use the integer `1` instead of `1.000000`, or any of the luts provided with Davinci Resolve 16 (free version) which uses 10 decimal places per entry instead of 6.

Darktable version:
```
$ darktable --version
this is darktable 3.0.0
copyright (c) 2009-2019 johannes hanika
darktable-dev@lists.darktable.org

compile options:
  bit depth is 64 bit
  normal build
  SSE2 optimized codepath enabled
  OpenMP support enabled
  OpenCL support enabled
  Lua support enabled, API version 5.0.2
  Colord support enabled
  gPhoto2 support enabled
  GraphicsMagick support enabled
  OpenEXR support enabled
```

phweyland:
Could you please provide some luts with 10 decimals ?

phweyland:
I cannot reproduce the issue with 1 instead of 1.00000. 
The luts from the site you provide work well for me (some of them have some 1). The corresponding code has not changed since 3.0.0 I believe...

mproffitt:
So this is one of the luts which come with the free version of DR16
[Blackmagic 4.6K Film to Rec709 v3.cube.txt](https://github.com/darktable-org/darktable/files/4288467/Blackmagic.4.6K.Film.to.Rec709.v3.cube.txt)

So with regards to the 1 vs 1.000000: This might have been a red-herring as after re-downloading them myself and applying, those files work fine

phweyland:
> So this is one of the luts which come with the free version of DR16
> [Blackmagic 4.6K Film to Rec709 v3.cube.txt](https://github.com/darktable-org/darktable/files/4288467/Blackmagic.4.6K.Film.to.Rec709.v3.cube.txt)

OK, this one is problematic. I'll look at it.

phweyland:
LUT_3D_INPUT_RANGE 0.0000000000 1.0000000000

[...]

PR #4422, opened 2020-03-04
lut 3d cube file fix

phweyland:
- skip non standard text line

Related to #4420

### DIFF
--- a/src/iop/lut3d.c
+++ b/src/iop/lut3d.c
@@ -782,7 +782,7 @@ uint16_t calculate_clut_cube(const char *const filepath, float **clut)
           return 0;
         }
       }
-      else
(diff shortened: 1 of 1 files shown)


TurboGit:
Is that for plain 3d lut or 3dl ? I'm asking to know if this needs merging in 3.0.1 branch. Thanks.

TurboGit:
Looking at the issue it seems to be plain 3d lut cube file.

phweyland:
> Is that for plain 3d lut or 3dl ? I'm asking to know if this needs merging in 3.0.1 branch. Thanks.

The change is only for cube file, not for 3dl. Of course, depending on the source for these files we may find some issues of this type too.

PR #4423, opened 2020-03-04
negadoctor : inital commit

### LINKED ISSUE #4113: Improve the negative film inversion
## Problem

The current film inversion is very basic and applied before demosaicing and input profile. There are 2 cases of negative film processing in dt:

1. working on an high bit-depth TIFF from a film scanner,
2. working on a photograph of the negative taken with a DSLR in raw format.

For 1., the TIFF encoding RGB space needs to be decoded first (at least undoing the gamma, but possibly neutralizing the scanner color casts), for 2. the camera sensor needs to be profiled and neutralized in white balance and color casts first. Both cases need WB and input color profile before inversion, so the negative inversion should happen later in the pipe.

## Current work-around

Capture d’écran du 2020-01-14 17-09-20

## Proposed method

Derivated from Kodak Cineon system, it aims at inverting a log-encoded (density-scaled) scan, then linearize it (sort-of) for better handling in the scene-linear pipe.

Here is the algo, designed for 10 bits integer encoding (to be adapted for floats):

1. Density(c) = -log10(dmin(c)/pixelValue(c))
2. densityEncoding(c) = s(c) * Density(c) * 500 + 95
3. linearized(c) = 10^densityEncoding(c) (or similar, see ref. 3.)

where :
* c is the colour channel
* dmin is the RGB colour of the film substrate
* 500 is an encoding gain adjusted for films having a density of 2.046, computed from 1023/2.046 = 500 (number of code values for current bit depth / film density)
* 95 is a black offset to adjust for compensating the variabilitiy of dmin, depending of film processing

So, user inputs are:

- film density,
- substrate colour,
- black offset

The encoding gain should be a constant if performing operations in float.

## References

- Thanks to Sidharth Sadhujan from Rochester Institute of Technology.
- http://www.digital-intermediate.co.uk/film/pdf/Cineon.pdf
- https://www.kodak.com/uploadedfiles/motion/US_plugins_acrobat_en_motion_education_sensitometry_workbook.pdf
- https://lists.gnu.org/archive/html/openexr-devel/2005-03/msg00009.html

aurelienpierre:
- introduce a film inversion module based on Kodak Cineon densitometry model that simulates a print from a digitized film negative.
- allows colour and contrast corrections.
- OpenCL kernel provided and tested on Intel Neo and Nvidia.
- fix #4113

From this:
house_03
To that:
house_02
In one single module, assisted with 7 color pickers to avoid guessing the best parameters.

Tutorial :

https://www.youtube.com/watch?v=DiNlHBZE888&feature=youtu.be

### DIFF
--- a/data/kernels/negadoctor.cl
+++ b/data/kernels/negadoctor.cl
@@ -1,6 +1,6 @@
 /*
     This file is part of darktable,
-    copyright (c) 2020 - Aurélien Pierre.
+    Copyright (C) 2020 darktable developers.
 
     darktable is free software: you can redistribute it and/or modify
     it under the terms of the GNU General Public License as published by
--- a/data/kernels/negadoctor.cl
+++ b/data/kernels/negadoctor.cl
@@ -1,0 +1,53 @@
+/*
+    This file is part of darktable,
+    copyright (c) 2020 - Aurélien Pierre.
+
+    darktable is free software: you can redistribute it and/or modify
+    it under the terms of the GNU General Public License as published by
+    the Free Software Foundation, either version 3 of the License, or
+    (at your option) any later version.
+
+    darktable is distributed in the hope that it will be useful,
+    but WITHOUT ANY WARRANTY; without even the implied warranty of
+    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+    GNU General Public License for more details.
+
+    You should have received a copy of the GNU General Public License
+    along with darktable.  If not, see <http://www.gnu.org/licenses/>.
+*/
+
+#include "common.h"
+
+
+kernel void
+negadoctor (read_only image2d_t in, write_only image2d_t out, int width, int height,
+            const float4 Dmin, const float4 wb_high, const float4 offset,
+            const float exposure, const float black, const float gamma, const float soft_clip, const float soft_clip_comp)
+{
+  const unsigned int x = get_global_id(0);
+  const unsigned int y = get_global_id(1);
+
+  if(x >= width || y >= height) return;
+
+  float4 i = read_imagef(in, sampleri, (int2)(x, y));
+  float4 o;
+
+  // Convert transmission to density using Dmin as a fulcrum
+  o = -native_log10(Dmin / fmax(i, (float4)2.3283064365386963e-10f)); // threshold to -32 EV
+
+  // Correct density in log space
+  o = wb_high * o + offset;
+
+  // Print density on paper : ((1 - 10^corrected_de + black) * exposure)^gamma rewritten for FMA
+  o = -((float4)exposure * native_exp10(o) + (float4)black);
+  o = native_powr(fmax(o, (float4)0.0f), gamma); // note : this is always > 0
+
+  // Compress highlights and clip negatives. from https://lists.gnu.org/archive/html/openexr-devel/2005-03/msg00009.html
+  o = (o > (float4)soft_clip) ? soft_clip + ((float4)1.0f - native_exp(-(o - (float4)soft_clip) / (float4)soft_clip_comp)) * (float4)soft_clip_comp
+                              : o;
+
+  // Copy alpha
+  o.w = i.w;
+
+  write_imagef(out, (int2)(x, y), o);
+}
--- a/src/common/iop_order.c
+++ b/src/common/iop_order.c
@@ -629,7 +629,7 @@ GList *dt_ioppr_get_iop_order_list(int32_t imgid, gboolean sorted)
         {
           // @@_NEW_MOUDLE: For new module it is required to insert the new module name in the iop-order list here.
           //                The insertion can be done depending on the current iop-order list kind.
-          _insert_before(iop_order_list, "nlmeans", "negadocton");
+          _insert_before(iop_order_list, "nlmeans", "negadoctor");
         }
       }
       else if(version == DT_IOP_ORDER_LEGACY)
--- a/src/common/iop_order.c
+++ b/src/common/iop_order.c
@@ -104,6 +104,7 @@ const dt_iop_order_entry_t legacy_order[] = {
   { {25.0f }, "profile_gamma", 0},
   { {26.0f }, "hazeremoval", 0},
   { {27.0f }, "colorin", 0},
+  { {27.5f }, "negadoctor", 0},
   { {27.5f }, "basicadj", 0},
   { {28.0f }, "colorreconstruct", 0},
   { {29.0f }, "colorchecker", 0},
@@ -185,6 +186,7 @@ const dt_iop_order_entry_t v30_order[] = {
   { {26.0f }, "profile_gamma", 0},
   { {27.0f }, "equalizer", 0},
   { {28.0f }, "colorin", 0},
+  { {28.5f }, "negadoctor", 0},      // Cineon film encoding comes after scanner input color profile
   { {29.0f }, "nlmeans", 0},         // signal processing (denoising)
                                   //    -> needs a signal as scene-referred as possible (even if it works in Lab)
   { {30.0f }, "colorchecker", 0},    // calibration to "neutral" exchange colour space
@@ -255,7 +257,7 @@ const dt_iop_order_entry_t v30_order[] = {
 static void *_dup_iop_order_entry(const void *src, gpointer data);
 static int _count_entries_operation(GList *e_list, const char *operation);
 
-#if 0
+
 static GList *_insert_before(GList *iop_order_list, const char *module, const char *new_module)
 {
   gboolean exists = FALSE;
@@ -303,7 +305,7 @@ static GList *_insert_before(GList *iop_order_list, const char *module, const ch
 
   return iop_order_list;
 }
-#endif
+
 
 dt_iop_order_t dt_ioppr_get_iop_order_version(const int32_t imgid)
 {
@@ -627,9 +629,7 @@ GList *dt_ioppr_get_iop_order_list(int32_t imgid, gboolean sorted)
         {
           // @@_NEW_MOUDLE: For new module it is required to insert the new module name in the iop-order list here.
           //                The insertion can be done depending on the current iop-order list kind.
-#if 0
-          _insert_before(iop_order_list, "<CURRENT_MODULE>", "<NEW_MODULE>");
-#endif
+          _insert_before(iop_order_list, "nlmeans", "negadocton");
         }
       }
       else if(version == DT_IOP_ORDER_LEGACY)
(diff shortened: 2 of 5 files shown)


TurboGit:
Can you provide a link to one negative image to play with and test all this ? Thanks.

aurelienpierre:
Here, first message : https://discuss.pixls.us/t/inverting-photos-of-color-negatives-with-darktable/9140

src/common/iop_order.c
```diff
@@ -627,9 +629,7 @@ GList *dt_ioppr_get_iop_order_list(int32_t imgid, gboolean sorted)
         {
           // @@_NEW_MOUDLE: For new module it is required to insert the new module name in the iop-order list here.
           //                The insertion can be done depending on the current iop-order list kind.
-#if 0
-          _insert_before(iop_order_list, "<CURRENT_MODULE>", "<NEW_MODULE>");
-#endif
+          _insert_before(iop_order_list, "nlmeans", "negadocton");
```
johnny-bit: I think you've meant "negadoctor" here :wink:

gadolf66:
In case you want one more well-behaved negative, please go to [https://discuss.pixls.us/t/inverting-photos-of-color-negatives-with-darktable/9140/77?u=gadolf](https://discuss.pixls.us/t/inverting-photos-of-color-negatives-with-darktable/9140/77?u=gadolf)

src/common/iop_order.c
aurelienpierre: Good catch, thanks !

matze:
Nice work Aurelien, I really like the result. Just a tiny remark concerning the name of the module that might cause quite some discussion: In German the pronunciation of "nega" is very close (and in some regional dialects indentical) to [Neger](https://de.wikipedia.org/wiki/Neger) which is a *very* offensive and insulting term for a person of color. I don't know if it's necessary to rename the module but the translator should find a suitable replacement.

aurelienpierre:
Ugh, people will see offensive stuff whenever there is nothing on TV and Facebook is open.

matze:
Fair enough, just saying that to my German ear it sounds weird. But again, I don't think you need to take action but approve with whatever the German translator comes up with (not that I would have good idea).

gadolf66:
In Brazilian Portuguese, _nega_ is an existent word for black people and also considered offensive.
Aurèlien, your excelent work transcends political correctness issues, my sugestion is to abide to our times and focus on what interests us all.

aurelienpierre:
neger, nigga, neger, nega, nero, negro, negre… meant black in latin. Now, if they overlap with negative (film or social connotation) in some languages, translators will need to deal with that based on the cultural parameters of their target crowd. Words are merely convenience symbols to designate something, they hold the meaning you assign them. I'm not interested in the 2020 bullshit about genders, races, and other boxes in which people are puc to cast some social group away from the rest based on stupid distinctions. 

Negadoctor is negative film doctor.

gadolf66:
> 
> 
> if they overlap with negative (film or social connotation) in some languages, translators will need to deal with that based on the cultural parameters of their target crowd. 

I had completely forgotten the translations, so I agree with that.
And only that :)

src/iop/negadoctor.c
```diff
@@ -0,0 +1,1387 @@
+/*
+    This file is part of darktable,
```
TurboGit: Header should now be:

  This file is part of darktable,
  Copyright (C) 2020 darktable developers.

src/iop/negadoctor.c
```diff
@@ -0,0 +1,1387 @@
...
+
+typedef struct dt_iop_negadoctor_global_data_t
+{
+  int kernel_negadoctor;
+} dt_iop_negadoctor_global_data_t;
+
+
+const char *name()
+{
+  return _("negadoctor");
```
TurboGit: BTW, a proposal for a French translation ?

src/iop/negadoctor.c
```diff
@@ -0,0 +1,1387 @@
...
+}
+
+
+int flags()
+{
+  return IOP_FLAGS_INCLUDE_IN_STYLES | IOP_FLAGS_ALLOW_TILING;
```
TurboGit: Does it makes sense to have multiple instances ?

src/iop/negadoctor.c
johnny-bit: Proposal for Polish translation: "doktor negatywów" :P

data/kernels/negadoctor.cl
```diff
@@ -0,0 +1,53 @@
+/*
+    This file is part of darktable,
```
TurboGit: As for C, header should be:

This file is part of darktable,
Copyright (C) 2020 darktable developers.

src/iop/negadoctor.c
jakubfi: @johnny-bit just in case ;-)

src/iop/negadoctor.c
aurelienpierre: No, indeed, I didn't thought of it.

src/iop/negadoctor.c
aurelienpierre: negadocteur ?

data/kernels/negadoctor.cl
TurboGit: You missed this one.

aurelienpierre:
Ooops, fixed.

aurelienpierre:
> Also, nice to see that the iop-order circuitry for new module is working. You're the first to enable this part of the code.

On that note, it might make more sense to have an `insert_after` function, since a module cares about what happens before in the pipe, rather than what comes after.

AxelG-DE:
Guys I love all the work from Aurélien. May I recommend to rename this module...

Nega from "Negadoctor" sounds in German very similar like Neger, which is in Englich Nigro and nowadays, where political correctness reaches a level, which is difficult to discuss on, this word might be even considered to be to close to 'Nigger'...

matze:
There has been a discussion about that topic, see above. Ideally translators will find a suitable replacement, I guess "Negativdoktor" or "Negativentwicklung" would be good alternatives.

spaceChRiS:
"Negativentwicklung" sounds to me much better than anything "...doktor" or "...doctor" does for another reason: because it describes much better what the module does. Having the reference to the doctor (of medicine) in one module (image doctor) is funny (problem with your image? Go to the image doctor ...), but personally I would avoid stuffing too many doctors into darktable. These modules would not stand out any longer but would become arbitrary, and it wound be much harder to understand the purpose of a module by its name.

On the other hand, do we already have a wiz(z)ard in darktable? ;)

aurelienpierre:
Negadoctor or image doctor are the names I chose to avoid overly technical names for modules that perform different kinds of complex operations but for one single purpose (negadoctor: virtual print of film negative, image doctor : all kinds of low-level image reconstruction). 

In English, "to doctor" is also used as a verb in the sense of "to alter something" (it's not only medical), and here it is used as a word play around the idea that it performs **quite evolved** alterations but sort of knows what it is doing (so it can be almost entirely auto-tuned), therefore users don't need to dive too deep into the technicalities.

All in all, I think it's just a funny name easier to remember, probably more catchy than "negative development". I just ask translators to keep in the name the idea that it is a negative film inversion with a print simulation and colour correction.

gadolf66:
This is a minor issue, but I'm not able to use CTRL+mouse scroll in the _print exposure adjustment_ slider.
I haven't tested the other ones.

leclercj:
I had a try with this module to process some scans of old negatives from my archives (back in 80's to 90's). I got very nice results. All my previous attempts using invert iop or LUT iop miserably failed. This negadoctor module is doing a very great job. Thank you Aurelien ! I'm looking forward to have it available in stable dt release.

guillaumedavidphd:
A quick note on the name: I must say, negadoctor sounds awfully close to the N-word of the American language. Not ideal. Further, the shortened version of the word negative "nega" is commonly used in French but not in English as far as I know. If you're dead set on the doctor part, I suggest you consider something like "negative doctor" instead.

AxelG-DE:
@guillaumedavidphd 
see my commend from March 8th

aurelienpierre:
Fix you racial issues, not software.

guillaumedavidphd:
In what way is the name software?

aurelienpierre:
How many actual black people are in this thread again ? And how many in general around the darktable software project ?

White educated men getting offended in advance on behalf on some minority is the worse and more perverse kind of patriarchy ever. It simply implies that the concerned minority is unable to defend itself.

Black people have emails and access to Github accounts too, if they are offended, they are welcome to say it. Meanwhile, white guys playing white knights are only enforcing the very thing they pretend to fight.

Email your black friends with that thread, let's hear them. I'll email mine. We will see.

Grandsome:
You could call it niggadoctor, and it would probably get a chuckle from most black people. But, if means  peace from white bourgeois liberals just call it negative doctor at the user end, it's also more descriptive. (BTW, do I have to show my black person card, they didn't issue one at my birth)

aurelienpierre:
> (BTW, do I have to show my black person card, they didn't issue one at my birth)

Spot-on :grinning:

Grandsome:
You could call it niggadoctor, and it would probably get a chuckle from most black people. But, if means  peace from white bourgeois liberals just call it negative doctor at the user end, it's also more descriptive. (BTW, do I have to show my black person card, they didn't issue one at my birth)

data/kernels/negadoctor.cl
```diff
@@ -0,0 +1,53 @@
...
+
+  // Correct density in log space
+  o = wb_high * o + offset;
+
+  // Print density on paper : ((1 - 10^corrected_de + black) * exposure)^gamma rewritten for FMA
+  o = -((float4)exposure * native_exp10(o) + (float4)black);
+  o = native_powr(fmax(o, (float4)0.0f), gamma); // note : this is always > 0
+
+  // Compress highlights and clip negatives. from https://lists.gnu.org/archive/html/openexr-devel/2005-03/msg00009.html
+  o = (o > (float4)soft_clip) ? soft_clip + ((float4)1.0f - native_exp(-(o - (float4)soft_clip) / (float4)soft_clip_comp)) * (float4)soft_clip_comp
```
Weltgeist, line 46: https://lists.gnu.org/archive/html/openexr-devel/2005-03/msg00009.html

soft_clip or b and soft_clip_comp or s. Not Equivalent to ref functions. did you mix them up?
Add doxygen commentary of expected value for softclip 0 to 0.4 or 0.6 to 1.

aurelienpierre: In the original code, s is an user param and b is an internal variable. Essentially, provided peak luminance is 1.0 by definition, s means "distance to peak luminance" (used as an analog of standard deviation in the exponential) and b means "absolute luminance threshold". 

For GUI intuitivity, I judged that defining b directly makes more sense for users (as in : define the luminance threshold, not the standard deviation-ish). So you directly know highlights start rolling off from 80% or 90% up to 100%. 

But for s and b in [0 ; 1], b = 1 - s and s = 1 - b, so this code is entirely equivalent to the original.

PR #4424, opened 2020-03-04
keyboard shortcuts to better handle multiple instances

### LINKED ISSUE #2760: Issue with changing exposure value from histogram view
**Describe the bug**
After creating multiple instances of exposure iop, changing exposure value by mouse roll over histogram view creates some weird artifacts all over the photo.

**To Reproduce**
1. Open a photo in darkroom
2. Enable exposure iop
3. Create a second instance of exposure iop and enable it
4. Select first exposure entry from history stack
5. Try changing exposure value by mouse roll from the histogram view

**Expected behavior**
Changing exposure value in the first (default) instance of exposure iop

**Screenshots**
Before:
Screenshot from 2019-06-30 17-18-32

After:
Screenshot from 2019-06-30 17-18-42

 - OS: Ubuntu 18.04
 - Version: git master

### LINKED ISSUE #3769: rolling the mouse wheel inside the histogram used to increase/decrease exposure but now displays individually odd pixel colors
rolling the mouse wheel inside the histogram used to increase/decrease
exposure but now displays individually odd pixel colors

**To Reproduce**
Open an image and display histogram
place mouse pointer in histogram display
roll mouse wheel.

should increase or decrease exposure but instead display odd color pixels

see screen shot:  http://wahoo.no-ip.org/~paka/Screenshot_20191215_112740.png

**Expected behavior**

**Screenshots**

 - OS: [e.g. funtoo linux]
 - Version [e.g. 2.6.0]
 - OpenCL activated or no
 - which GFX-card and GFX-driver

**Additional context**
Add any other context about the problem here, for example:
 - Can you reproduce with another dt version?
not available

only tested raw

 - Is dt compiled by yourself? If so which compiler is used? Which options?
using package from darix, openSUSE Tumbleweed
darktable-3.0.0rc2~git102.c09d36df9-2039.1.x86_64

### LINKED ISSUE #4285: Keyboard shortcuts with multiple instances
darktable has a number of keyboard shortcuts that allow you to interact with darkroom image processing modules. However, where multiple instances of a module are present, it is not easy to predict which instance of the module the shortcuts will affect (especially where instances have been renamed). I would like to propose a set of consistent rules that can be used to decide which module instance the keyboard shortcuts interact with.

I’m using the exposure module as an example, because it has a multiple ways to interact with it and it highlights inconsistencies between those methods. Also it’s probably the module most likely to be used with multiple instances.

I have set up shortcuts to show the module (Shift+e), increase the exposure (=), decrease the exposure (-), and a dynamic shortcut that allows me to adjust the exposure with the scroll wheel (e).

The following observations are all using an image that has two instances of the exposure module (“exposure” and “exposure 1”).

If I use the histogram to change the exposure (by mouse-dragging), those adjustments always impact the latest module in the pixelpipe (in this case “exposure 1” - if I swap the modules in the pixelpipe, the change impacts “exposure”).

If I use “Shift+e”, this also activates “exposure 1”. However, if I change the pixelpipe order, that shortcut still activates “exposure 1”. The same is true with the dynamic shortcut (e), increase (=) and decrease (-). 

I assume from these tests that darktable is applying the keyboard shortcut to the last created instance of the module. Renaming and reordering the instances doesn’t change which one is updated, and this is fairly important. Many users utilise named instances to manage their workflow and once an instance has been renamed it’s hard to know which was the last one created, potentially making shortcuts less useful or leading them to have unexpected/undesired effects. 

I would like to propose the following set of rules to decide which module instance a shorcut interacts with – this would be the most logical for my workflow but I know it may not suit everyone so I leave the exact details for discussion. 

I suggest that darktable starts with a list of all instances of a given module and applies the shortcut whenever the following filter operations reduce the number of available instances to 1:

1. If any instances are displayed (module is open on the RHS of the screen), select only those instances, else select all instances
2. If any instances from the previous list are active, select only those instances, else leave the list unchanged
3. If any instances from the previous list are global in operation (do not have masks associated) select only those instances, else leave the list unchanged
4. If there are still multiple instances in the list, select the instance that is latest in the pixelpipe

Assuming that we are keeping the interaction between the histogram and exposure module (I know there have been discussions about removing it), the same order should be used for this interaction.

elstoc:
Resolves #4285. Resolves #3769. Resolves #2760.

In order to make keyboard shortcuts within the darkroom more consistent where multiple module instances are present, some additional parameters have been introduced to the Preferences dialog (under the Miscellaneous tab) in a new section 'keyboard shortcuts with multiple instances'. These preferences control rules that are applied (in order) to decide which module instance keyboard shortcuts will be applied to.

1. **prefer expanded instances** default FALSE (if instances of the module are expanded, ignore collapsed instances)
2. **prefer enabled instances** default FALSE (after applying the above rule, if instances of the module are active, ignore inactive instances)
3. **prefer unmasked instances** default FALSE (after applying the above rules, if instances of the module are unmasked, ignore masked instances)
4. **selection order** default "last instance" (after applying the above rules, apply the shortcut to the first or last instance remaining)

These rules also apply when mouse-dragging on the histogram to change exposure.

Darktable will reattach the keyboard shortcuts (using the above rules) to a different instance, if required, in the following scenarios:

- When opening a new image in the darkroom (when initially loading the image and when navigating to previous/next image)
- When selecting a history item in the 'history' module
- (if prefer expanded instances is set) when expanding/collapsing an instance (via the gui or a keyboard shortcut)
- (if prefer enabled instances is set) when enabling/disabling an instance (via the gui or a keyboard shortcut)
- (if prefer unmasked instances is set) when changing the masking behaviour of an instance
- (if prefer active/enabled/unmasked instances is set) when choosing a preset (via the module or darkroom menus)
- (if prefer active/enabled/unmasked instances is set) when resetting module parameters
- When choosing a style (via the module or darkroom menus)
- When duplicating, cloning, deleting or moving a module
- When hiding/unhiding a module in the 'more modules' module

### DIFF
--- a/src/control/control.c
+++ b/src/control/control.c
@@ -28,6 +28,7 @@
 #include "control/conf.h"
 #include "control/control.h"
 #include "develop/develop.h"
+#include "develop/imageop.h"
 #include "gui/accelerators.h"
 #include "gui/draw.h"
 #include "gui/gtk.h"
@@ -712,9 +713,11 @@ int dt_control_key_pressed_override(guint key, guint state)
   if(darktable.view_manager->current_view->dynamic_accel_current)
   {
     gchar **vals = g_strsplit_set(darktable.view_manager->current_view->dynamic_accel_current->translated_path, "/", -1);
+    dt_iop_module_so_t *mod_so = darktable.view_manager->current_view->dynamic_accel_current->mod_so;
+    dt_iop_module_t *mod = dt_iop_get_module_accel_curr(mod_so);
     if(vals[0] && vals[1] && vals[2] && vals[3])
     {
-      gchar *txt = dt_util_dstrcat(NULL, _("scroll to change <b>%s</b> of %s module"), vals[3], vals[2]);
+      gchar *txt = dt_util_dstrcat(NULL, _("scroll to change <b>%s</b> of %s %s module"), vals[3], vals[2], mod->multi_name);
       dt_control_hinter_message(darktable.control, txt);
       g_free(txt);
     }
--- a/data/darktableconfig.dtd
+++ b/data/darktableconfig.dtd
@@ -2,7 +2,7 @@
 <!ELEMENT dtconfig (name,type,default,capability?,shortdescription?,longdescription?)>
 <!ATTLIST dtconfig
 	prefs (import|lighttable|darkroom|otherviews|processing|security|cpugpu|storage|misc) #IMPLIED
-        section (import|session|interface|other|tags|geoloc|slideshow|cpugpu|database|xmp) #IMPLIED
+        section (import|session|interface|other|tags|geoloc|slideshow|cpugpu|database|xmp|accel) #IMPLIED
         capability CDATA #IMPLIED
 >
 <!ELEMENT name (#PCDATA)>
--- a/data/darktableconfig.xml.in
+++ b/data/darktableconfig.xml.in
@@ -1,6 +1,39 @@
 <?xml version="1.0"?>
 <!DOCTYPE dtconfiglist SYSTEM "darktableconfig.dtd">
 <dtconfiglist>
+  <dtconfig prefs="misc" section="accel">
+    <name>accel/prefer_expanded</name>
+    <type>bool</type>
+    <default>false</default>
+    <shortdescription>prefer expanded instances</shortdescription>
+    <longdescription>if instances of the module are expanded, ignore collapsed instances</longdescription>
+  </dtconfig>
+  <dtconfig prefs="misc" section="accel">
+    <name>accel/prefer_enabled</name>
+    <type>bool</type>
+    <default>false</default>
+    <shortdescription>prefer enabled instances</shortdescription>
+    <longdescription>after applying the above rule, if instances of the module are active, ignore inactive instances</longdescription>
+  </dtconfig>
+  <dtconfig prefs="misc" section="accel">
+    <name>accel/prefer_unmasked</name>
+    <type>bool</type>
+    <default>false</default>
+    <shortdescription>prefer unmasked instances</shortdescription>
+    <longdescription>after applying the above rules, if instances of the module are unmasked, ignore masked instances</longdescription>
+  </dtconfig>
+  <dtconfig prefs="misc" section="accel">
+    <name>accel/select_order</name>
+    <type>
+      <enum>
+        <option>first instance</option>
+        <option>last instance</option>
+      </enum>
+    </type>
+    <default>last instance</default>
+    <shortdescription>selection order</shortdescription>
+    <longdescription>after applying the above rules, apply the shortcut based on its position in the pixelpipe</longdescription>
+  </dtconfig>
   <dtconfig>
     <name>bauhaus/scale</name>
     <type>float</type>
--- a/src/bauhaus/bauhaus.c
+++ b/src/bauhaus/bauhaus.c
@@ -21,6 +21,7 @@
 #include "common/darktable.h"
 #include "control/conf.h"
 #include "develop/develop.h"
+#include "develop/imageop.h"
 #include "gui/gtk.h"
 #ifdef GDK_WINDOWING_QUARTZ
 #include "osx/osx.h"
--- a/src/develop/blend_gui.c
+++ b/src/develop/blend_gui.c
@@ -382,6 +382,12 @@ static void _blendop_masks_mode_callback(const unsigned int mask_mode, dt_iop_gu
   }
 
   dt_dev_add_history_item(darktable.develop, data->module, TRUE);
+
+  if(dt_conf_get_bool("accel/prefer_unmasked"))
+  {
+    // rebuild the accelerators
+    dt_iop_connect_accels_multi(data->module->so);
+  }
 }
 
 
--- a/src/develop/develop.c
+++ b/src/develop/develop.c
@@ -2057,8 +2057,7 @@ static dt_dev_proxy_exposure_t *find_last_exposure_instance(dt_develop_t *dev)
 {
   if(!dev->proxy.exposure) return NULL;
 
-  dev->proxy.exposure = g_list_sort(dev->proxy.exposure, dt_dev_exposure_hooks_sort);
-  dt_dev_proxy_exposure_t *instance = (dt_dev_proxy_exposure_t *)(g_list_last(dev->proxy.exposure)->data);
+  dt_dev_proxy_exposure_t *instance = (dt_dev_proxy_exposure_t *)(g_list_first(dev->proxy.exposure)->data);
 
   return instance;
 };
--- a/src/develop/imageop.c
+++ b/src/develop/imageop.c
@@ -626,8 +626,9 @@ static void dt_iop_gui_delete_callback(GtkButton *button, dt_iop_module_t *modul
     dt_control_signal_raise(darktable.signals, DT_SIGNAL_DEVELOP_HISTORY_CHANGE);
   }
 
-  // we cleanup the module
-  dt_accel_disconnect_list(module->accel_closures);
+  // rebuild the accelerators (to point to an extant module)
+  dt_iop_connect_accels_multi(module->so);
+
   dt_accel_cleanup_locals_iop(module);
   module->accel_closures = NULL;
   // don't delete the module, a pipe may still need it
@@ -740,6 +741,8 @@ static void dt_iop_gui_movedown_callback(GtkButton *button, dt_iop_module_t *mod
   prev->dev->preview_pipe->cache_obsolete = 1;
   prev->dev->preview2_pipe->cache_obsolete = 1;
 
+  // rebuild the accelerators
+  dt_iop_connect_accels_multi(module->so);
   dt_control_signal_raise(darktable.signals, DT_SIGNAL_DEVELOP_MODULE_MOVED);
 
   // invalidate buffers and force redraw of darkroom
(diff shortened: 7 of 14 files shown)


johnny-bit:
Hmm... Interesting idea!
As for commits - since it's early in PR you could just squash commits into single one and force-push new commit to your branch, rewritting history but geting rid of that "merge" commit :)

elstoc:
Thanks @johnny-bit. Would you be able to give me a pointer please?!

johnny-bit:
Sure! on your branch do this in command line:

```
git reset $(git merge-base master $(git rev-parse --abbrev-ref HEAD))
git add -A
git commit -m "v1 - fixed most of the bugs - accelerators now apply to latest module in pipe"
git push origin --force
```

elstoc:
Hmm looks like the 'merge' commit is still there

johnny-bit:
Oh... It's because your master is poisoned with it: https://github.com/darktable-org/darktable/compare/master...elstoc:master

 You need to fix your master branch on your master repo so that it's even with darktable:master and then you can fixup your commit.

elstoc:
That's what I get for blindly following instructions on a random website. Still not 100% sure how to go about this :(

elstoc:
Thanks @johnny-bit. Commits are now much tidier.

src/control/control.c
```diff
@@ -667,9 +668,19 @@ int dt_control_key_pressed_override(guint key, guint state)
   if(darktable.view_manager->current_view->dynamic_accel_current)
   {
     gchar **vals = g_strsplit_set(darktable.view_manager->current_view->dynamic_accel_current->translated_path, "/", -1);
+    dt_iop_module_so_t *mod_so = darktable.view_manager->current_view->dynamic_accel_current->mod_so;
+    dt_iop_module_t *mod = dt_iop_get_module_accel_curr(mod_so);
     if(vals[0] && vals[1] && vals[2] && vals[3])
     {
-      gchar *txt = dt_util_dstrcat(NULL, _("scroll to change <b>%s</b> of %s module"), vals[3], vals[2]);
+      gchar *txt;
+      if(mod->multi_name && mod->multi_name[0] != '\0')
```
johnny-bit: This generates following error:
```
/build/darktable/src/control/control.c:676:15: error: address of array 'mod->multi_name' will always evaluate to 'true' [-Werror,-Wpointer-bool-conversion]

      if(mod->multi_name && mod->multi_name[0] != '\0')

         ~~~~~^~~~~~~~~~ ~~
```

elstoc: There wasn't any real need for the if statement any more so I removed it in latest commit. You'll get an extra space in the hint if there's only one version of the module but I don't think that's a big deal.

TurboGit:
> At the moment all this does is ensure that keyboard accelerators are always associated with the latest instance of a module in the pipe.

Would it makes more sense to have the keyboard shortcuts use the current active instance or if no instance have the focus currently use the last instance in the pipe. My point is that if you have an instance expanded and visible it would be quite strange to control another one non visible.

What do you think ?

elstoc:
@TurboGit the issue is still that of the 'toggle' shortcut would toggle the active one off and then toggle a different one back on the second time but maybe we could live with that?. Though I think we're probably only going to have a 'least worst' solution to this problem.

elstoc:
@TurboGit there are similar discussions happening both on the PR on the related issue - I'm not sure which is the best place but it should probably be just one of them. Can you advise which is the best place?

TurboGit:
@elstoc : Well for sure the discussion about implementation should happen here. But now it is difficult to draw a line for all discussions! So at the end I would say that if there is a PR the discussion should probably be done in the PR.

elstoc:
I do agree with both @TurboGit and @johnny-bit that it doesn't make sense, if you have an active instance of a module, for the shortcuts not to operate on that instance.  I'm also reluctant to have the 'common' shortcuts (expand/collapse, on/off, presets etc.) operate on a different instance to the shortcuts that interact directly with the module, both because it's confusing for the user and also because it'll be a pain to code.

Perhaps we should just live with the fact that the toggle shortcuts will always behave strangely with multiple instances - I just fear there's a chance it will lead to issues being raised further down the line because it will seem wrong when a user toggles twice. I don't think there is a right answer so am happy to go with the consensus.

elstoc:
That's the last of the issues on my list fixed now (I think). Just need to decide/add the new rules for module selection.

elstoc:
Ok so the best I can come up with is to set up a bunch of new preferences that default to the current (intended) behaviour of applying only to the latest instance.

I'll create a new preferences section in GUI Options, named "keyboard shortcuts with multiple instances" as follows:

- Prefer expanded modules (checkbox, default unchecked, tooltip "when choosing which module to apply keyboard shortcuts to, give preference to modules that are expanded in the right-hand panel")
- Prefer enabled modules (checkbox, default unchecked, tooltip "when choosing which module to apply keyboard shortcuts to, give preference to modules that are currently active")
- Prefer unmasked modules (checkbox, default unchecked, tooltip "when choosing which module to apply keyboard shortcuts to, give preference to modules that apply to the entire image (i.e. unmasked)")
- Apply shortcut to (drop-down, options "first instance | last instance", default "last instance", tooltip "if, after applying the above rules, multiple instances still remain, choose based on the instance order in the pixelpipe")

Rules will be applied in the order given. For the first three options, I'll add additional calls to my new function (to reassign the accelerators) when those parameters are changed against a module.

johnny-bit:
The logic is sound, the only problem I see is clear definition of _active_ module. I think if you added a header in gui saying something along the lines of "shortcuts work on _active_ module instance. follwing options define which instance is considered _active_" would be really great addition for those who can read.

I also thought about toggle switches (which can also be subject to rules but separate from those 4 mentioned): toggle to disable/hide should be applied to _active_ module. If toggle to disable/hide was used remember that module, so that toggle to enable/show is applied to _toggled module_. if any action is done in different module instance, *forget* recently _toggled module_ and instead toggle switches to _active_ module. Toggle to enable module with no instained instances should work normally.

elstoc:
> clear definition of _active_ module

I'm trying to avoid words that might imply another meaning - people might read _active_ as meaning 'enabled' or 'expanded' - especially when the instance considered 'active' might be neither of those things. Happy to make this change, but need a better word to use!

With regard to toggles working differently I'm not sure (from a code point-of-view) how this could be made to work without a fairly large rewrite. What you've suggested assumes that the code decides which instance to apply to when you press the button but what actually happens is that it calls a callback function that's been previously assigned. Part of the reason I've chosen the default values in the preferences is that those offer predictable behaviour. If the user chooses to do something else, any unpredictable behaviour is entirely their choice!

elstoc:
Latest commit is a fully working version implementing the draft rules I've given above. I've basically only tested it using multiple instances of the exposure module so will need more extensive testing, but it seems ok to me thus far.

Give it a try if you have time and let me know what you think.

elstoc:
I've updated the OP with a summary of the functionality and current progress.

elstoc:
Testing is now complete.

johnny-bit:
What's the recommended testing scenario?

elstoc:
With all the combinations of parameters and ways of arranging instances it was difficult to test everything so it was a case of thoroughly testing the individual preferences one at a time and then a few more complicated scenarios. 

To test the main functionality I ended up creating an image with 7 named instances of an exposure module with various combinations of 'expanded/collapsed', 'active/inactive', and mask types, combined with various combinations of preferences, ensuring that when I changed the properties of those instances the accelerator instance was changed (where applicable). With the displaying of the module instance as a hint when you use dynamic accelerators, this was a useful way to prove that the correct module was being selected. Using shortcuts to enable/disable the module or expand/collapse it works but obviously you get the toggle issue we've discussed before.

I could only really test whether or not the 'shortcut rebuild' had fired (the second set of bullets in my description) by putting a debug message in the header. But that's purely an efficiency thing and probably won't affect day-to-day use. 

Not sure if that helped but I only really fully documented the simpler tests so I could get my head around them.

johnny-bit:
@TurboGit - I haven't managed to break it, and in multi-instance scenario it's night-and-day compared to current master...

Can you test & merge? It's cool bit to have!

elstoc:
Thanks for your help on this one @johnny-bit. Glad you like it.

dterrahe:
In the hidden slider popup code in bauhaus.c you removed the second part of the condition below:
if(w->module && !strstr(w->module->name(), w->label))

The point of this is to not show the name of module if that is simply a set of names of the sliders within the module. For example, it is much easier to read:
   contrast
than
   contrast brightness saturation/contrast

When there are multiple copies of a module it would make sense to always show the module name.

elstoc:
I was going for consistency (the message would change quite a bit once you add a new instance) and to avoid having to count the number of active instances each time it's called, though I do take your point. I don't mind changing it if you think it's worth it.

dterrahe:
Yes please. It makes sense that the message changes if there are multiple instances, but that doesn't have to impact _all_ modules.

Thanks!

elstoc:
I've also incorporated a suggestion from @TurboGit's review of https://github.com/darktable-org/darktable/pull/4468 (to add spaces around the slash in the message)

elstoc:
@TurboGit this is ready for review when you have time.

johnny-bit:
I'm using this quite consistently :) Haven't broken anything and makes working with multiple instances fun ;)

IMO it's ready :)

TurboGit:
@elstoc : I suppose you'll want to rework part of this after the your preference revamp.

elstoc:
Indeed. Probably just move that section to the bottom of the darkroom tab

elstoc:
On further reflection I think I prefer to see the new options as 'shortcut' options rather than as 'darkroom' options and therefore add them to the miscellaneous tab (also because the misc tab already has sub-sections and it might risk us running out of space on the darkroom tab already).

elstoc:
I've done some brief retesting after incorporating some recent changes to master and I think this is ready to go again now.

johnny-bit:
It's very nice and working as advertised!

PR #4426, opened 2020-03-05
history_hash fix

### LINKED ISSUE #4400: Windows fragility
version 3.1.0_777~gbd7c89a0b-dirty, Windows 10, seems very fragile.
Crashes happen in very unpredictable ways... sometimes going to the darkroom, sometimes to the lighttable, sometimes on changing images, very often (but not always) on attempting to apply a style.

Sometimes the lock files are left active, sometimes not. Problems seem equally likely for any image, original or duplicate.

If I recover my DB from back-up and return to v708, all is well again.

So what should I look for in the -d log? It seems just to record ordinary daily life, then stops as though I'd closed the program.

phweyland:
- malloc size !!
- avoid ON CONFLICT where possible
- ~~make sqlite3 2.40 mandatory~~

Should fix #4400 - Sorry for all this time lost :(

### DIFF
--- a/src/common/history.c
+++ b/src/common/history.c
@@ -1357,12 +1357,9 @@ void dt_history_hash_write(const int32_t imgid, dt_history_hash_values_t *hash)
   {
     sqlite3_stmt *stmt;
     DT_DEBUG_SQLITE3_PREPARE_V2(dt_database_get(darktable.db),
-                                "INSERT INTO main.history_hash"
+                                "INSERT OR REPLACE INTO main.history_hash"
                                 " (imgid, basic_hash, auto_hash, current_hash)"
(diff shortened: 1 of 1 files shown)


phweyland:
I've removed the constraint about sqlite version as I find this a bit harsh for those who cannot easily update sqlite.
I'll prepare another PR, making 2.15 mandatory, and providing a backup way for versions below 2.40.

GrahamByrnes:
A massive thanks for sorting this. Will download and leave it to build while I'm at work...

TurboGit:
Thanks !

GrahamByrnes:
Works like a charm :-)

PR #4428, opened 2020-03-06
Sqlite version required >=  3.15

phweyland:
Two changes:
- Sqlite version build constraint > 3.15
- History-hash code for sqlite version < or >= 3.24


TurboGit:
Thanks !

src/CMakeLists.txt
```diff
@@ -337,7 +337,18 @@ if(USE_LENSFUN)
   endif()
 endif()
 
-foreach(lib ${OUR_LIBS} GIO GThread GModule PangoCairo Rsvg2 LibXml2 Sqlite3 CURL PNG JPEG TIFF LCMS2 JsonGlib)
+find_package(Sqlite3 3.15 REQUIRED)
+if(NOT Sqlite3_VERSION VERSION_LESS 3.24)
+  add_definitions("-DHAVE_SQLITE_324_OR_NEWER")
+  message(STATUS "SQLITE version greater or newer than 3.24")
+else()
+  message(STATUS "SQLITE version less than 3.24, some features may be downgraded")
```
johnny-bit: Similarly here you should use `Sqlite3` instead of `SQLITE`. And maybe instead of "less than" I'd use "lower than" or "earlier than"?

phweyland: True for both. I'll fix that. Thanks

PR #4430, opened 2020-03-06
bugfix: tone equalizer with masking

### LINKED ISSUE #4020: [Tone Equalizer] influences drawn masks in other modules when zooming
**Describe the bug**
Weird artifacts showing, when zooming to finetune a drawn mask and tone-eq has been switched on.
It merely shows in the lower half of the picture, the upper part is just B/W

**To Reproduce**
1. Use filmicRGB and tone equalizer
2. e.g. creaete a drawn mask in denoise profiled
3. middle click on the lower half of the pic to zoom
4b. now switch of tone-eq and do step 3 again = no effect

**Expected behavior**
Drawn mask shall be always shown correctly

**Screenshots**
overview
Screenshot_20200101_230615

Zoomed in lower haf
Screenshot_20200101_230638

Zoomed in middle of pic
Screenshot_20200101_231024

Tone Eq off
Screenshot_20200101_231209

 - Darktable Version: **3.0.0** and **current master**
 - OS: **Gentoo Linux**
 - OpenCL activated or no? **on and off**
 - Which graphics card and driver version **2x nvidia GTX1060; nvidia-drivers 440.44-r1**

**Additional context**
 - Can you reproduce with another Darktable version? **Yes**
 - Can you reproduce with a RAW or Jpeg or both? **RAW**
 - Are the steps above reproduce with a fresh edit (removing history)? **Yes**
 - Attach an XMP if this is necessary 
[2019-12-14_124134__EM50058_02.ORF.xmp.tar.gz](https://github.com/darktable-org/darktable/files/4021931/2019-12-14_124134__EM50058_02.ORF.xmp.tar.gz) 
not 100% same version, as I had to redraw the mask
 - Did you compile Darktable yourself? If so which compiler was used, with what options? **x86_64-pc-linux-gnu-9.2.0 CMAKE_BUILD_TYPE="Release" CFLAGS="-march=native -O3 -mtune=native -pipe"**
 - Is the issue still present using an empty/new config-dir **Yes**

aurelienpierre:
Don't use the default darktable alpha copy function, but pass-through the alpha mask with the properly offset loops.

Recall that tone equalizer input region of interest has to be larger than its output ROI to keep the guided filter preview consistent no matter the zooming scale. So the in -> out mapping has to take into account the pixel shifting to correlate both ROI coordinates.

This is a temporary fix to avoid crashes on segfault. I still need to understand how to map this with standard dt function (mask_distort and such).

Fix #4020

### DIFF
--- a/src/iop/toneequal.c
+++ b/src/iop/toneequal.c
@@ -751,7 +751,8 @@ static inline void compute_luminance_mask(const float *const restrict in, float 
  **/
 
 __DT_CLONE_TARGETS__
-static inline void display_luminance_mask(const float *const restrict luminance,
+static inline void display_luminance_mask(const float *const restrict in,
+                                          const float *const restrict luminance,
                                           float *const restrict out,
                                           const dt_iop_roi_t *const roi_in, const dt_iop_roi_t *const roi_out,
                                           const size_t ch)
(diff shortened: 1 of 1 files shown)


TurboGit:
Works for me. Thanks.

Issue #3353, opened 2019-11-09
Color/Spot picker can be used outside picture area, with different unwanted results.


jade-nl:
darktable 3.0.0rc0+82~g4c6df92a0

**Describe the bug**
When using the color/spot picker you are able to use it outside the image area. This produces (severe) unwanted results.

Depending on the module you try this with it (three examples):

- freezes/crashes dt (exposure module)
- produces a black (or blue) image (filmic rgb module)
- produces a black screen (WB module with spot)

The last two can be recovered by resetting the module.

**To Reproduce**
Use a 'picker', any picker, and click/select outside the image area.

**Expected behavior**
I expect nothing to happen. Click/select outside the image area should be ignored. As it is in version 2.6

**Platform (please complete the following information):**
 - OS: Linux Debian 10.1 (and also confirmed on up-to-date Gentoo)
 - Version  darktable 3.0.0rc0+82~g4c6df92a0

EDIT: words

AxelG-DE:
I confirm this

blitzgneisserin:
very similar behavior here

junkyardsparkle:
I couldn't reproduce the crash/freeze with the exposure module on an image first imported by recent master, however, on a random image from further back in time the freeze happened as soon as I clicked on the picker button, without any further action (image view was showing the default "image-minus-edges" selection)... so there might be more going on with this one.

junkyardsparkle:
The situation with the gray value slider in filmic RGB is especially ugly; the resulting zero value is out of normal bounds, and the response of the black and white sliders is to go insane and hide their labels... the labels don't come back after gray is restored to a sane value, or even by reseting the module; the sliders must be double-clicked.

jade-nl:
Update after getting a *Stale issue message*  in my in-box:

This issue is still relevant for the current dt versions. Just re-checked using:

- 3.0.0 (stable)
- 3.1.0-362-g39a47f6aa

Only one thing has changed since initially posting this issue: I haven't seen any crashes/freezes when testing to see if this is still relevant.

Just in case the initial description wasn't clear enough:

The picker can be used in the grey "border" area around the image (crossing the border itself is _not_ possible).

jade-nl:
I asked Aurelien on IRC about the bot closing this issue even though activity has taken place. He said he disabled the bot and I was hoping that he would re-open this issue as well. Guess not.

Can somebody reopen it, this isn't solved and tags have been added.

parafin:
[...]

PR #4431, opened 2020-03-06
Fixing Colorpicker ranges Issue #3353

jenshannoschwalm:
The #3353 issue reported, that the colorpicker might access data outside the image area if there is any border.
This pr solves two glitches

1) The mouse positions taken for the color picker area are restricted to the real-image-data area.
2) Make sure we always use an area. This avoid "selecting black" if just a dot or a line has bee used.


TurboGit:
Works for me ! Thanks.

PR #4432, opened 2020-03-07
fix sqlite version wording

phweyland:


TurboGit:
Thanks ! Better indeed.

PR #4436, opened 2020-03-07
Prohibit pollution of darkroom borders

jenshannoschwalm:
Both the color picker marking rectangles and the mask display are drawn after the
image in darkroom.c expose.

As the drawing of these is not clipped properly the borders (the grey and the iso)
may be partially overwritten.

In this pr we cairo_clip to the 'real' image thus avoiding the 'pollution'.


TurboGit:
Every details count ! Thanks for this fix.

Issue #4156, opened 2020-01-20
Glitch issue on focus peaking in culling mode (image zoomed)


Nilvus:
Just see the images below. First image is an image fit in the borders preview. Second one is the same zoomed. darktable always show focus peaking of all image, even outside the preview zone of the image. Reproducible on few set of images tested. Quite ugly ! @aurelienpierre and @AlicVB ?

**To Reproduce**
Just select for example to images and go in culling mode with focus peaking on. Zoom on the images and see the issue.

**Expected behavior**
Limit to image preview zone.

**Screenshots**
![Capture-20200120231632-1922x1054](https://user-images.githubusercontent.com/45535283/72761335-0327b800-3bdc-11ea-85dc-c1bf68a398d5.png)

![Capture-20200120231545-1922x1054](https://user-images.githubusercontent.com/45535283/72761342-0622a880-3bdc-11ea-80a5-32156fe69e77.png)

**Platform (please complete the following information):**
 - Darktable Version: master (last commit of this day)
 - OS: Ubuntu 19.10
 - No OpenCL

aurelienpierre:
I don't reproduce.

Nilvus:
> I don't reproduce.

Quite strange, it's always reproducible, on all images I tested (something like a dozen) and I can also reproduce it each time in classic preview view on lighttable (default Alt+W shortcut). See below (on little grey borders, you could see focus peaking appear):

![Capture-20200121225712-1922x1054](https://user-images.githubusercontent.com/45535283/72846809-9b3ba500-3ca1-11ea-9640-c89a0746d8be.png)

AlicVB:
oups, forgot this one... I can reproduce, and it comes from the fact that focus drawing doesn't clip to the image area like the image surface.
That said, actually for zoomed images, focus peaking is recompute each time you move the image, which is not necessary. But I'm reluctant to fix that now, as it'll be simpler once culling/preview is moved to specific widgets, like what I'm doing for filemanager.

jenshannoschwalm:
@Nilvus what about closing this?

PR #4439, opened 2020-03-07
Fix issue on focus peaking in culling mode #4156

jenshannoschwalm:
The issue #4156 describes the problem in a perfect way, also the analysis of @AlicVB
seems to be fully correct.

As we know the position & size of the image immediately before adding the peakings
we can restrict that to a clipped cairo region.

@aurelienpierre & @AlicVB ; any objections?

### DIFF
--- a/src/views/view.c
+++ b/src/views/view.c
@@ -1413,10 +1413,15 @@ int dt_view_image_expose(dt_view_image_expose_t *vals)
       cairo_fill(cr);
 
       if(darktable.gui->show_focus_peaking)
+      {
+        cairo_save(cr);
+        cairo_rectangle(cr, rectx, recty, rectw, recth);
+        cairo_clip(cr);
         dt_focuspeaking(cr, width, height, cairo_image_surface_get_data(surface),
                                            cairo_image_surface_get_width(surface),
                                            cairo_image_surface_get_height(surface));
-
+        cairo_restore(cr);
+      }
       if(!vals->full_surface || !*(vals->full_surface)) cairo_surface_destroy(surface);
     }
 


aurelienpierre:
The code looks ok to me. I will test later.

Nilvus:
@jenshannoschwalm: thanks, just tested with same images shown on my issue, it's now ok. Focus peaking is now only shown inside the image. As @AlicVB, it will now just stay what he answer on my issue: "That said, actually for zoomed images, focus peaking is recompute each time you move the image, which is not necessary."
But it will be for his next rework of culling mode (after lighttable one).

AlicVB:
> As @AlicVB, it will now just stay what he answer on my issue: "That said, actually for zoomed images, focus peaking is recompute each time you move the image, which is not necessary."
> But it will be for his next rework of culling mode (after lighttable one).

Indeed, that makes panning images in culling or full preview almost unusable for big images. But I'll take care of that in my next rework ;)

aurelienpierre:
> Indeed, that makes panning images in culling or full preview almost unusable for big images. But I'll take care of that in my next rework ;)

Do you have a caching mechanism ?

AlicVB:
@aurelienpierre : in fact, to allow smooth panning of zoomed images, the whole image surface is loaded once for all and we just clip it to show the wanted part. As focus peaking arrive late in `dt_view_image_expose`, it's recomputed at each move, which can be quite slow with big images.

But as said don't worry, I'll remove all that `dt_view_image_expose` mess soon (already done for classic thumbnails in my lighttable rewrite), and compute focus peaking only once.

aurelienpierre:
I designed the API of the focus peaking to be as general as possible, so I hope it helps in this case.

AlicVB:
It's ok, don't worry ;) in fact it's just a matter of "when" applying focus peaking...
Anyway, I'll ping you if I have specials needs :)

jenshannoschwalm:
Ok -  the new lightroom seems to be in a much better shape, i also had it running here for a test @AlicVB keep going :-)
This pr is here as a fix for the situation right now.

TurboGit:
Merging then! Thanks all !

PR #4416, opened 2020-03-03
Simple padding fix on export module

Nilvus:
Fix padding issue reported on #4412.
For 3.0.1

PR #4444, opened 2020-03-09
not so simple padding fix around spinbuttons

johnny-bit:
In PR #4416 @Nilvus introduced a padding fix for spin inputs that fixed export window spin input padding issue.

*Before #4416*
bad export padding
*After #4416*
god export padding

That however introduced padding problem in preferences window.

*Before #4416*
good preferences padding

*After #4416*
bad preferences padding

this PR fixes padding for spin input in preferences window without touching all other spin inputs.

This PR: good padding
This PR: godd padding

### DIFF
--- a/data/themes/darktable.css
+++ b/data/themes/darktable.css
@@ -1047,6 +1047,10 @@ spinbutton {
     margin-top: 0;
 }
 
+#preferences_notebook scrolledwindow spinbutton {
+    padding: 0px;
+}
+
 /* Color picker module in darkroom left panel */
 
 #color-picker-area
--- a/data/themes/darktable.css
+++ b/data/themes/darktable.css
@@ -1017,7 +1017,7 @@ cell:selected
     color: #ffff00;
 }
 
-spinbutton {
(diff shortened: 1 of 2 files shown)


Nilvus:
Thanks for that. I have checked others parts but forgot preferences window.

Did you check if it's possible to apply my fix only on lighttable instead of adding lines on css like your fix ?

I will not have time to check that today.

johnny-bit:
Thanks for grep homework and gtk homework and gui messing-around!

Now this change limits padding for spinbuttons only to `max size` spinbuttons without messing around with ANY other spinbutton, even in `export` module itself as prooven by `pdf` export:

Nice padding everywhere

Nilvus:
@johnny-bit. Thanks. Better for me. I now remember why I haven't add export module : it was not possible and Gtk is not something I can work on.

And i like your title. Quite good joke about my previous fix 😆

Nilvus:
@TurboGit: I think this could be merged. It's ok for me.

TurboGit:
Ok, merging. Thanks.

PR #4461, opened 2020-03-12
iop: restrict pointers and advertise alignment

aurelienpierre:
first step of optimizations on the IOP API. It doesn't seem to increase perfs, but will enable further optimizations later and force devs to use a cleaner separation between GUI functions, `commit_params` and `process`.

1. change iop `process()` API to make every argument constant except output image buffer. *Note : levels, rgblevels and basicadj write user params in `process()`, it's bad practice. These functions need to treat users params read-only. User params belong to GUI.*
2. change iop `commit_params()` API to make `dt_pipeline` and `dt_module_params` constant
3. repercutes the changes on `blend.c` and `iop_profile.c`
4. advertise the 64-bits alignment of in/out image buffers in iop `process()` to give hints to compilers about vectorization
5. manually inline some functions

Many loops will have to be re-written to keep improving parallelization with OpenMP pragmas (iop and masks).


TurboGit:
I will wait for the new lighttible to land-in before merging this.

TurboGit:
Moved to 3.4 as we cannot merge such a delicate change without lot of field testing.

TurboGit:
Forget about 3.4, lot of conflicts and too many things to do at the moment.

PR #4463, opened 2020-03-12
wip: try to add Exif.Photo.ExposureBiasValue to variables + image info

### LINKED ISSUE #2680: Missing exposure bias [EV] exif tag
Hi,

In 2.7 the option to configure EXIF tags shown can be tweaked with this new setting:

Pattern for the image info line

It says "see manual to see all the tags you can use". I can't find any mention of those tags... I grep'ed the source code to find the EXIF_xxx tags, but can't find any "exposure bias" related one... This is super important to see when you intentionally under/over expose a photo using the EV setting on camera, or when do you do bracketing.

I can be read from the photo with exiv2:

```
[gaetan@xps 06. Piana]$ exiv2 IMG_9564.JPG 
File name       : IMG_9564.JPG
[...]
Exposure time   : 1/400 s
Aperture        : F5.6
Exposure bias   : 0 EV
Flash           : No, compulsory
[...]

```

How can it be used?

Thanks

aurelienpierre:
Try to fix #2680 and https://forums.darktable.fr/showthread.php?tid=4483

Unfortunately, it turned out to be more complicated than I thought and I need some help to finish. (maybe @phweyland or @jenshannoschwalm since you have worked on these parts ?).

1. The variable `$(EXIF_EXPOSURE_BIAS)` is not replaced in the image info by the value (even the default one),
2. The `dt_image_t` member `exif_exposure_bias` seems to not be read from the file EXIF, does it need to be fetched from database ?

Context : the camera exposure bias is useful when working in bracketing mode, while doing HDR or looking for the best exposure.

### DIFF
--- a/src/common/exif.cc
+++ b/src/common/exif.cc
@@ -505,11 +505,11 @@ static bool dt_exif_read_xmp_data(dt_image_t *img, Exiv2::XmpData &xmpData, int 
       char *c ;
       while ( ( c = strchr(datetime,'T') ) != NULL )
       {
-	*c = ' ';
+  *c = ' ';
       }
       // replace '-' by ':'
       while ( ( c = strchr(datetime,'-')) != NULL ) {
-	*c = ':';
+  *c = ':';
       }
 
       g_strlcpy(img->exif_datetime_taken, datetime, sizeof(img->exif_datetime_taken));
@@ -734,6 +734,13 @@ static bool dt_exif_read_exif_data(dt_image_t *img, Exiv2::ExifData &exifData)
       // uf_strlcpy_to_utf8(uf->conf->shutterText, max_name, pos, exifData);
       img->exif_exposure = 1.0 / pos->toFloat();
     }
+
+    // Read exposure bias
+    if(FIND_EXIF_TAG("Exif.Photo.ExposureBiasValue"))
+    {
+      img->exif_exposure_bias = pos->toFloat();
+    }
+
     /* Read aperture */
     if(FIND_EXIF_TAG("Exif.Photo.FNumber"))
     {
--- a/src/common/image.c
+++ b/src/common/image.c
@@ -1400,6 +1400,7 @@ void dt_image_init(dt_image_t *img)
   g_strlcpy(img->exif_datetime_taken, "0000:00:00 00:00:00", sizeof(img->exif_datetime_taken));
   img->exif_crop = 1.0;
   img->exif_exposure = 0;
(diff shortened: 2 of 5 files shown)


jenshannoschwalm:
@aurelienpierre You are probably very aware the EXIF data is only read in while importing or "refreshing". So the image struct only holds these as long as it is still in the cache.

I read your code and probably you would be able to see the data if you refresh or import just a few images to have them in the cache.

You definitely need a new entry in the database (as done for focus distance as an example). 

Execptions are those exif data that are checked & presented via rawprepare, at least for DNGs there are examples.

Another approach would be to test such EXIF data and leave a flag in the image struct (kept in the database) and reading that exif data if required by a module. (Making things pretty slow)

Just a few comments - i can't test and help moret atm as i am stuck with work being an intensive care doc thinking more about containing covid-19 and keep us all safe (&working) in the hospitals than about improving my photo workflow :-!

phweyland:
@aurelienpierre , as Hanno said duplicatingthe  code related to focus_distance should makes the exposure bias available in dt.


>     $(EXIF_EXPOSURE_BIAS)

This is another subject, these variables are used to build export file name or watermark. If useful for that, should be added in variables.c (as "EXIF_FOCUS_DISTANCE"). 
On the other hand this info is already available in exported metadata.

aurelienpierre:
> You are probably very aware the EXIF data is only read in while importing or "refreshing". So the image struct only holds these as long as it is still in the cache.

I wasn't, thanks for pointing that out.

> You definitely need a new entry in the database (as done for focus distance as an example).

That was my fear… 

> Just a few comments - i can't test and help moret atm as i am stuck with work being an intensive care doc thinking more about containing covid-19 and keep us all safe (&working) in the hospitals than about improving my photo workflow :-!

No problem, go save lifes doc !

> This is another subject, these variables are used to build export file name or watermark. If useful for that, should be added in variables.c (as "EXIF_FOCUS_DISTANCE").

My code adds them in variables already. 

> On the other hand this info is already available in exported metadata.

Not sure I understand ? `Exif.Photo.ExposureBiasValue` is saved in exported files, is what you mean ?

phweyland:
@aurelienpierre, If that helps I'll be happy to complete this.

aurelienpierre:
@phweyland gladly, thanks. I'm hammering maths problems right now, it's good to be away from any C for a bit.

TurboGit:
Closing this one as it is being rewritten here #4592

PR #4468, opened 2020-03-13
keyboard shortcuts to cycle through iop module combobox options

### LINKED ISSUE #4356: Shortcut to cycle through guides in the crop tool
It would be great to have a keyboard shortcut for cycling through the many different guides when using the crop tool. This is something that Lightroom has (shortcut is `o`) and which I miss in Darktable.

elstoc:
This PR allow shortcuts to be assigned to previous and next values of a combobox in an iop module.

Each time the combobox value is changed through a keyboard shortcut its new value is displayed on an onscreen popup (similar to how changed slider values are displayed).

The following is a list of iop comboboxes included:

 basic group

- color reconstruction: "precedence"
- shadows & highlights: "soften with"
- base curve: "scale", "preserve colors", "fusion"
- basic adjustments: "preserve colors"
- tone equalizer: "luminance estimator", "preserve details"
- exposure: "mode"
- crop & rotate: "flip", "keystone", "automatic cropping", "aspect", "guides"
- demosaic: "method", "color smoothing", "match greens"
- highlight reconstruction: "method"
- white balance: "preset"

tone group

- local contrast: "mode"
- global tonemap: "operator"
- levels: "mode"
- tone curve: "color space", "interpolation method", "preserve colors", "scale"
- filmic rgb: "preserve chrominance"
- rgb levels: "preserve colors"
- rgb curve: "mode, "interpolation method", "preserve colors"

color group

- output color profile: "export profile"
- color zones: "select by", "process mode", "interpolation method"
- color balance: "mode", "color control sliders"
- channel mixer: "destination"
- lut 3D: "application color space", "interpolation"
- color look up table: "target color"
- input color profile: "input profile", "working profile", "gamut clipping"
- unbreak input profile: "mode"

correction group

- dithering: method
- defringe: operation mode
- perspective correction: "guides", "automatic cropping", "lens model"
- lens correction: "corrections", "geometry", "mode"
- denoise (profiled): "mode"

effects group

- watermark: "marker", "scale on"
- framing: "aspect", "orientation", "horizontal position", "vertical position"

Resolves #4356.

### DIFF
--- a/src/bauhaus/bauhaus.c
+++ b/src/bauhaus/bauhaus.c
@@ -1204,9 +1204,9 @@ void dt_bauhaus_combobox_set(GtkWidget *widget, int pos)
     if(!gtk_widget_is_visible(GTK_WIDGET(w)) && *w->label)
     {
       if(w->module && w->module->multi_name[0] != '\0')
-        dt_control_log(_("%s %s/%s: %s"), w->module->name(), w->module->multi_name, w->label, dt_bauhaus_combobox_get_text(widget));
+        dt_control_log(_("%s %s / %s: %s"), w->module->name(), w->module->multi_name, w->label, dt_bauhaus_combobox_get_text(widget));
       else if(w->module)
-        dt_control_log(_("%s/%s: %s"), w->module->name(), w->label, dt_bauhaus_combobox_get_text(widget));
+        dt_control_log(_("%s / %s: %s"), w->module->name(), w->label, dt_bauhaus_combobox_get_text(widget));
       else
         dt_control_log(_("%s"), dt_bauhaus_combobox_get_text(widget));
     }
--- a/src/bauhaus/bauhaus.c
+++ b/src/bauhaus/bauhaus.c
@@ -1197,7 +1197,20 @@ void dt_bauhaus_combobox_set(GtkWidget *widget, int pos)
   dt_bauhaus_combobox_data_t *d = &w->data.combobox;
   d->active = CLAMP(pos, -1, d->num_labels - 1);
   gtk_widget_queue_draw(GTK_WIDGET(w));
-  if(!darktable.gui->reset) g_signal_emit_by_name(G_OBJECT(w), "value-changed");
+  if(!darktable.gui->reset) 
+  {
+    g_signal_emit_by_name(G_OBJECT(w), "value-changed");
+
+    if(!gtk_widget_is_visible(GTK_WIDGET(w)) && *w->label)
+    {
+      if(w->module && w->module->multi_name[0] != '\0')
+        dt_control_log(_("%s %s/%s: %s"), w->module->name(), w->module->multi_name, w->label, dt_bauhaus_combobox_get_text(widget));
+      else if(w->module)
+        dt_control_log(_("%s/%s: %s"), w->module->name(), w->label, dt_bauhaus_combobox_get_text(widget));
+      else
+        dt_control_log(_("%s"), dt_bauhaus_combobox_get_text(widget));
+    }
+  }
 }
 
 gboolean dt_bauhaus_combobox_set_from_text(GtkWidget *widget, const char *text)
--- a/src/gui/accelerators.c
+++ b/src/gui/accelerators.c
@@ -526,8 +526,6 @@ static gboolean bauhaus_combobox_next_callback(GtkAccelGroup *accel_group, GObje
 
   dt_bauhaus_widget_t *w = (dt_bauhaus_widget_t *)DT_BAUHAUS_WIDGET(combobox);
 
-  g_signal_emit_by_name(G_OBJECT(combobox), "value-changed");
-
   if(!gtk_widget_is_visible(GTK_WIDGET(w)) && *w->label)
   {
     if(w->module && w->module->multi_name[0] != '\0')
@@ -551,8 +549,6 @@ static gboolean bauhaus_combobox_prev_callback(GtkAccelGroup *accel_group, GObje
   dt_bauhaus_combobox_set(combobox, prevval);
 
   dt_bauhaus_widget_t *w = (dt_bauhaus_widget_t *)DT_BAUHAUS_WIDGET(combobox);
-
-  g_signal_emit_by_name(G_OBJECT(combobox), "value-changed");
 
   if(!gtk_widget_is_visible(GTK_WIDGET(w)) && *w->label)
   {
--- a/src/gui/accelerators.c
+++ b/src/gui/accelerators.c
@@ -510,8 +510,9 @@ static gboolean bauhaus_combobox_next_callback(GtkAccelGroup *accel_group, GObje
 {
   GtkWidget *combobox = GTK_WIDGET(data);
 
-  int currentval = dt_bauhaus_combobox_get(combobox);
-  dt_bauhaus_combobox_set(combobox, ++currentval);
+  const int currentval = dt_bauhaus_combobox_get(combobox);
+  const int nextval = currentval + 1 >= dt_bauhaus_combobox_length(combobox) ? 0 : currentval + 1;
+  dt_bauhaus_combobox_set(combobox, nextval);
 
   dt_bauhaus_widget_t *w = (dt_bauhaus_widget_t *)DT_BAUHAUS_WIDGET(combobox);
 
@@ -534,8 +535,10 @@ static gboolean bauhaus_combobox_prev_callback(GtkAccelGroup *accel_group, GObje
 {
   GtkWidget *combobox = GTK_WIDGET(data);
 
-  int currentval = dt_bauhaus_combobox_get(combobox);
-  dt_bauhaus_combobox_set(combobox, --currentval);
+  const int currentval = dt_bauhaus_combobox_get(combobox);
+  const int prevval = currentval - 1 < 0 ? dt_bauhaus_combobox_length(combobox) : currentval - 1;
+
+  dt_bauhaus_combobox_set(combobox, prevval);
 
   dt_bauhaus_widget_t *w = (dt_bauhaus_widget_t *)DT_BAUHAUS_WIDGET(combobox);
 
--- a/src/gui/accelerators.c
+++ b/src/gui/accelerators.c
@@ -512,7 +512,19 @@ static gboolean bauhaus_combobox_next_callback(GtkAccelGroup *accel_group, GObje
 
   int currentval = dt_bauhaus_combobox_get(combobox);
   dt_bauhaus_combobox_set(combobox, ++currentval);
-  dt_control_hinter_message(darktable.control, dt_bauhaus_combobox_get_text(combobox));
+
+  dt_bauhaus_widget_t *w = (dt_bauhaus_widget_t *)DT_BAUHAUS_WIDGET(combobox);
+
+  if(!gtk_widget_is_visible(GTK_WIDGET(w)) && *w->label)
+  {
+    if(w->module && w->module->multi_name[0] != '\0')
+      dt_control_log(_("%s %s/%s: %s"), w->module->name(), w->module->multi_name, w->label, dt_bauhaus_combobox_get_text(combobox));
+    else if(w->module)
+      dt_control_log(_("%s/%s: %s"), w->module->name(), w->label, dt_bauhaus_combobox_get_text(combobox));
+    else
+      dt_control_log(_("%s"), dt_bauhaus_combobox_get_text(combobox));
+  }
+
   g_signal_emit_by_name(G_OBJECT(combobox), "value-changed");
   return TRUE;
 }
@@ -524,7 +536,19 @@ static gboolean bauhaus_combobox_prev_callback(GtkAccelGroup *accel_group, GObje
 
   int currentval = dt_bauhaus_combobox_get(combobox);
   dt_bauhaus_combobox_set(combobox, --currentval);
-  dt_control_hinter_message(darktable.control, dt_bauhaus_combobox_get_text(combobox));
+
+  dt_bauhaus_widget_t *w = (dt_bauhaus_widget_t *)DT_BAUHAUS_WIDGET(combobox);
+
+  if(!gtk_widget_is_visible(GTK_WIDGET(w)) && *w->label)
+  {
+    if(w->module && w->module->multi_name[0] != '\0')
+      dt_control_log(_("%s %s/%s: %s"), w->module->name(), w->module->multi_name, w->label, dt_bauhaus_combobox_get_text(combobox));
+    else if(w->module)
+      dt_control_log(_("%s/%s: %s"), w->module->name(), w->label, dt_bauhaus_combobox_get_text(combobox));
+    else
+      dt_control_log(_("%s"), dt_bauhaus_combobox_get_text(combobox));
+  }
+
   g_signal_emit_by_name(G_OBJECT(combobox), "value-changed");
   return TRUE;
 }
--- a/src/gui/accelerators.c
+++ b/src/gui/accelerators.c
@@ -537,7 +537,6 @@ static gboolean bauhaus_combobox_prev_callback(GtkAccelGroup *accel_group, GObje
 
   const int currentval = dt_bauhaus_combobox_get(combobox);
   const int prevval = currentval - 1 < 0 ? dt_bauhaus_combobox_length(combobox) : currentval - 1;
-
   dt_bauhaus_combobox_set(combobox, prevval);
 
   dt_bauhaus_widget_t *w = (dt_bauhaus_widget_t *)DT_BAUHAUS_WIDGET(combobox);
--- a/src/gui/accelerators.c
+++ b/src/gui/accelerators.c
@@ -217,6 +217,39 @@ void dt_accel_register_lib(dt_lib_module_t *self, const gchar *path, guint accel
   dt_accel_register_lib_for_views(self, v, path, accel_key, mods);
 }
 
+void dt_accel_register_combobox_iop(dt_iop_module_so_t *so, gboolean local, const gchar *path)
+{
+  gchar accel_next_path[256];
+  gchar accel_prev_path[256];
+
+  dt_accel_t *accel_next = (dt_accel_t *)g_malloc(sizeof(dt_accel_t));
+  dt_accel_t *accel_prev = (dt_accel_t *)g_malloc(sizeof(dt_accel_t));
+
+  //accel to select next value
+  snprintf(accel_next_path, 256, "<Darktable>/%s/%s/%s/%s", NC_("accel", "image operations"), so->op, path,
+           NC_("accel", "next"));
+  gtk_accel_map_add_entry(accel_next_path, 0, 0);
+  g_strlcpy(accel_next->path, accel_next_path, sizeof(accel_next->path));
+  dt_accel_path_iop_translated(accel_next_path, sizeof(accel_next_path), so, path);
+  g_strlcpy(accel_next->translated_path, accel_next_path, sizeof(accel_next->translated_path));
+  g_strlcpy(accel_next->module, so->op, sizeof(accel_next->module));
+  accel_next->local = local;
+  accel_next->views = DT_VIEW_DARKROOM;
+  darktable.control->accelerator_list = g_slist_prepend(darktable.control->accelerator_list, accel_next);
+
+  //accel to select previous value
+  snprintf(accel_prev_path, 256, "<Darktable>/%s/%s/%s/%s", NC_("accel", "image operations"), so->op, path,
+           NC_("accel", "previous"));
+  gtk_accel_map_add_entry(accel_prev_path, 0, 0);
+  g_strlcpy(accel_prev->path, accel_prev_path, sizeof(accel_prev->path));
+  dt_accel_path_iop_translated(accel_prev_path, sizeof(accel_prev_path), so, path);
+  g_strlcpy(accel_next->translated_path, accel_prev_path, sizeof(accel_prev->translated_path));
+  g_strlcpy(accel_prev->module, so->op, sizeof(accel_prev->module));
+  accel_prev->local = local;
+  accel_prev->views = DT_VIEW_DARKROOM;
+  darktable.control->accelerator_list = g_slist_prepend(darktable.control->accelerator_list, accel_prev);
+}
+
 void dt_accel_register_slider_iop(dt_iop_module_so_t *so, gboolean local, const gchar *path)
 {
   gchar increase_path[256];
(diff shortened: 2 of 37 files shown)


johnny-bit:
This is good. Very Good!

My problem: can you implement a wrap around? And smarter handlig of `prev`?

here's the problem for `next`: in `accelerators.c` line 513-514 you have:
```c
  int currentval = dt_bauhaus_combobox_get(combobox);
  dt_bauhaus_combobox_set(combobox, ++currentval);
```
this in `dt_bauhaus_combobox_set` gets `CLAMP`ed to `combobox->data.combobox->num_labels - 1`, so pressing "next" again and again has no effect.

I changed it to:
```c
  const int currentval = dt_bauhaus_combobox_get(combobox);
  const int maxval = dt_bauhaus_combobox_length(combobox);
  const int nextval = currentval + 1 >= maxval ? 0 : currentval + 1;
  dt_bauhaus_combobox_set(combobox, nextval);
```

and it works like a charm (wraps around, however might be a bit verbose ;)

the problem with prev is - it's currently doing:
```c
  int currentval = dt_bauhaus_combobox_get(combobox);
  dt_bauhaus_combobox_set(combobox, --currentval);
```
which clamps to... `-1`. So if you have combobox already on 1st item and click on prev you'll get memory access violation ;)

So adapting my suggestion from next you'll have wraparound and no memory access violation :)
```c
  const int currentval = dt_bauhaus_combobox_get(combobox);
  const int maxval = dt_bauhaus_combobox_length(combobox);
  const int prevval = currentval - 1 < 0 ? maxval : currentval - 1;
  dt_bauhaus_combobox_set(combobox, prevval);
```

Hmm... I wonder if I can PR your repo with this...

elstoc:
> I wonder if I can PR your repo with this...

You could but then you'd have to help me fix my repo again when I merge it wrong!! Probably easier for me to do it myself. Won't take long...

johnny-bit:
OK :) plz do so :) i haven't pr'ed to fork that's been sumbited to be pr itself ;)

src/gui/accelerators.c
```diff
@@ -472,6 +505,85 @@ static gboolean bauhaus_slider_reset_callback(GtkAccelGroup *accel_group, GObjec
...
+}
+
+static gboolean bauhaus_combobox_prev_callback(GtkAccelGroup *accel_group, GObject *acceleratable,
+                                                 guint keyval, GdkModifierType modifier, gpointer data)
+{
+  GtkWidget *combobox = GTK_WIDGET(data);
+
+  const int currentval = dt_bauhaus_combobox_get(combobox);
+  const int prevval = currentval - 1 < 0 ? dt_bauhaus_combobox_length(combobox) : currentval - 1;
+
```
johnny-bit: Just a stylistic note - in `bauhaus_combobox_prev_callback` you have empty line before this call. I'm no expert on code style but should be consistent ;)

elstoc: New commit or just amend the last one?

johnny-bit: amend. or have it a part of another commit in the future. whitespace changes are absolutelly miniscule things.

elstoc: Ok will do. I do want to add the lens correction combobox because I use this one quite a lot, so I'll do it with that commit.

elstoc:
I've added shortcuts for all the comboboxes I need including the one mentioned in the original issue. I've suggested a couple more in the OP but I'm not sure they're worth doing.

Thoughts?

johnny-bit:
Now it works great!

I thought a bit about which comboboxes could have sortcut and... dunno.
I checked modules around, tried playing with them etc...

Maybe following (I marked which items I think I'd use with :+1:):

## lighttable

### export selected
- target storage
- file format
- style

## darkroom - basic group
### base curve
- preserve colors
- fusion (good for wraparound since it has 3 items ;) )

### basic adjustments
- preserve colors 

### tone equalizer
- luminance estimator :+1: 
- preserve details

### crop and rotate
- [ ] flip
- [ ] keystone
- [x] aspect :+1:
- [x] guides :+1:

### highlight reconstruction
- [ ] mode

### white balance
- [x] preset :+1:

## darkroom - tone group

### local contrast
- mode

### global tonemap
- operator

### tone curve
- color space
- preserve colors

### filmic rgb
- [ ] preserve chrominance

### rgb levels
- mode
- preserve colors

### rgb curve
- mode
- preserve colors

## darkroom - color group

### output color profile
- export profile (dunno)

### color zones
- select by :+1: 

### input color profile
- input profile (dunno)

## darkroom - correction group

### dithering
- method

### defringe 
- operation mode

### lens correction
- [x] corrections

### denoise (profiled)
- [ ] mode :+1:

## darkroom - effects group

### watermark
- marker :+1: 

### framing
- aspect

## map

### map settings
- map source :+1:

## print

### print settings
#### printer
- printer (dunno)
- media
- profile
- intent (dunno)
#### page
- paper size
- orientation
#### print settings
- profile (dunno)
- intent (dunno)
- style

## tethering

### Live view
- guides :+1:
- overlay mode

elstoc:
Wow that's quite a list. Is that all the comboboxes in darktable?!

I'm happy to add the ones you've requested but no point in adding the others unless we think they're likely to be used.  I suspect a lot of them probably only make sense when you're already interacting with the module (i.e where people are more likely to manually select).

Please note however, that this change only works for darkroom iop modules so I won't be able to add those in other views.

johnny-bit:
> Wow that's quite a list. Is that all the comboboxes in darktable?!

Nope... Roughly 1/2 ;)

> I'm happy to add the ones you've requested but no point in adding the others unless we think they're likely to be used. I suspect a lot of them probably only make sense when you're already interacting with the module (i.e where people are more likely to manually select).

I tried to remove those that make sense only when interacting with module from the list TBH...

> Please note however, that this change only works for darkroom iop modules so I won't be able to add those in other views.

Yep! But it could be base for others ;)

dterrahe:
> Wow that's quite a list. Is that all the comboboxes in darktable?!

It would be nice if all were automatically available, like they already are in vimkeys, which is achieved there by piggybacking on set_label.

I wonder if it wouldn't be more convenient to get rid of all the init/connect_key_accel boilerplate in each module and have the definition of each module's widgets in gui_init also set up an introspection map that could be used by all user input methods and automation interfaces.

This would be part of the ambitious redesign of the input layer @houz hints at in https://github.com/darktable-org/darktable/pull/3722#issuecomment-564629002

If anybody here has an interest in starting a brainstorm on the goals and ways to achieve them for such a project, I am still eager to participate.

> I'm happy to add the ones you've requested but no point in adding the others unless we think they're likely to be used. I suspect a lot of them probably only make sense when you're already interacting with the module (i.e where people are more likely to manually select).

It is hard to predict where people will want to use this and always frustrating to find your own case not to be supported. With mouse wheel shortcut (and midi controller...) support, people do no longer need to manually select anything in the module they are interacting with, or even have it visible.

elstoc:
> It is hard to predict where people will want to use this and always frustrating to find your own case not to be supported. 
Sounds like you're suggesting I add _all_ the iop comboboxes. I don't mind doing the dev (mostly a copy-and-paste job) but my testing may not be much beyond making sure it doesn't crash.

dterrahe:
> Sounds like you're suggesting I add _all_ the iop comboboxes. I don't mind doing the dev (mostly a copy-and-paste job) but my testing may not be much beyond making sure it doesn't crash.

That's why I suggested that it would be ideal if all this setup happened automatically in gui_init, whenever it defines any bauhaus widgets (i.e. move all of it to the bauhaus module). I went through all the modules to add accelerators for sliders, so at least all/the majority of the init/connect_accel methods are there. But whenever widgets or new modules are added the accelerators tend to be forgotten and it is easy to get inconsistent labels so they don't connect. _One_ authoritative definition would be preferable.

johnny-bit:
It might be a copy-paste but there's no harm in it isn't it? :P Or maybe leave it as "excercise for future" since adding it is as you say "mostly a copy-and-paste job" so anybody wanting additional combobox on it can request it in Issue and have PR ready for it in under 2 lines of code ;)

However I found one thing I really would like - in this PR you've already introduced prev/next (similar to increase/decrease for sliders). Can you please add _dynamic_ shortcut like for sliders too? so one could press key and use mouse wheel to scroll nex/prev value? Please?

@dterrahe - I've saw your huuuge MIDI work and I do think @houz and you are right regarding input system. A discussion about it should be done and prefferably on discuss.pixls.us, not necessarilly derailing another PR ;)

dterrahe:
I'd agree with that. But I'd suggest don't spend a lot of time now adding boiler plate everywhere which might just be stripped out again later if we can make all the accelerators show up for "free".

It is nice to have the requirements for the new input system documented in existing functionality.

elstoc:
> That's why I suggested that it would be ideal if all this setup happened automatically in gui_init

Agree but as you say that's probably part of a bigger rewrite and beyond the scope of what I'd intended with this PR (and potentially also impacts another PR I'm working on -- https://github.com/darktable-org/darktable/pull/4424 -- a rewrite like that is probably above my skill level right now)

> It might be a copy-paste but there's no harm in it isn't it?

I agree. I will do it.

> Can you please add dynamic shortcut like for sliders too?

I'm still not 100% certain how dynamic shortcuts work or whether it's worth it for comboboxes (it'll be really easy to overshoot values unless I get it perfect). Perhaps might be better in a separate PR, since once I've put the hooks into all the modules all the work will be in a couple of functions in accelerators.c.

johnny-bit:
I love #4424 (so far haven't managed to break it yet)

> Agree but as you say that's probably part of a bigger rewrite and beyond the scope of what I'd intended with this PR

No problem. IMO _Done is better than perfect._

> whether it's worth it for comboboxes

It _is_ for me! 

> Perhaps might be better in a separate PR, since once I've put the hooks into all the modules all the work will be in a couple of functions in accelerators.c.

That's the best way to do it IMHO.

elstoc:
As an aside, in the preferences view, I see a number of accelerators listed in the preferences view that don't appear to be connected to anything. For example, against the base curve there's a section for 'fusion' that appears to be a 'slider' accelerator, even though fusion is not a slider and there is nothing implementing that accelerator in `basecurve.c`.

Does anyone know where these might have come from?

ptilopteri:
used to be available in exposure.  I guess it would be termed as legacy.

elstoc:
Do you know if there is any way to clear out all the invalid shortcuts?

On 16 March 2020 22:35:20 GMT+00:00, ptilopteri <notifications@github.com> wrote:
>used to be available in exposure.  I guess it would be termed as
>legacy.
>
>-- 
>You are receiving this because you authored the thread.
>Reply to this email directly or view it on GitHub:
>https://github.com/darktable-org/darktable/pull/4468#issuecomment-599788635

-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

dterrahe:
The "fusion" accelerators for each module link to the opacity slider of the blending settings. It doesn't activate blending though, so you'll first have to switch it on manually to see the accelerator having any effect.

This is just one of the things that should be rethought if we are rebuilding input; it might make more sense to have a set of accelerators for the blending settings that then apply for the currently active module (for example).

elstoc:
Ah I see. Thanks.

On 16 March 2020 22:48:45 GMT+00:00, dterrahe <notifications@github.com> wrote:
>The "fusion" accelerators for each module link to the opacity slider of
>the blending settings. It doesn't activate blending though, so you'll
>first have to switch it on manually to see the accelerator having any
>effect.
>
>This is just one of the things that should be rethought if we are
>rebuilding input; it might make more sense to have a set of
>accelerators for the blending settings that then apply for the
>currently active module (for example).
>
>-- 
>You are receiving this because you authored the thread.
>Reply to this email directly or view it on GitHub:
>https://github.com/darktable-org/darktable/pull/4468#issuecomment-599792355

-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

elstoc:
The code changes for this PR are now complete. As previously discussed, the majority of the actual functionality is in accelerators.c and was fully tested in the initial five commits. The remainder of the changes were boilerplate code additions to connect the comboboxes of the remaining iop modules and have not been tested beyond checking that darktable compiles, images load into the darkroom without error (with all modules visible therefore accelerators connected) and that the shortcuts are present in the preferences screen.

johnny-bit:
I checked it and in image ops it doesn't cause issues...

However it does cause issues when re-binding keys.
Here's how to reproduce it:
1. assign keys to next + prev of one slided
2. assign same key to next of another slider as the next key from pt1 (should be fine)
3. assing same key to prev of another slider as the prev key from pt2 < here's bug

here's the output from console:
```

(darktable:33546): Pango-WARNING **: 21:34:19.712: Invalid UTF-8 string passed to pango_layout_set_text()

(darktable:33546): Pango-WARNING **: 21:34:19.729: Invalid UTF-8 string passed to pango_layout_set_text()

(darktable:33546): Pango-WARNING **: 21:34:19.734: Invalid UTF-8 string passed to pango_layout_set_text()

(darktable:33546): Pango-WARNING **: 21:34:19.758: Invalid UTF-8 string passed to pango_layout_set_text()

(darktable:33546): Pango-WARNING **: 21:34:19.758: Invalid UTF-8 string passed to pango_layout_set_text()

(darktable:33546): Pango-WARNING **: 21:34:19.759: Invalid UTF-8 string passed to pango_layout_set_text()
```

I also get a bunch of "translation missing" errors despite being on english locale.

here's the screenshot:

Zrzut ekranu z 2020-03-18 21-39-52

elstoc:
Good spot (nothing to do with language translations). Give it another try.

johnny-bit:
that fixed it!

@TurboGit - imo it's good, plz check how you like it :)

elstoc:
Thanks again for your help @johnny-bit

TurboGit:
Just testes and first question. I have added the shortcut `Ctrl+Shift+Alt+L` for the "local contrast" IOP mode.

If I open the "local contrast" and I type the key, I do see the mode switchinig from "local laplacien" and "bilateral grid". All fine.

But if I collapse the iop I get the following hint message:

"local contrast / shadow 50%"
and
"local contrast / granularity 50% "

Looks like the wrong control data is being displayed.

elstoc:
I don't know. It's getting label and value data directly from the same widget that it's adjusting. The widget is being adjusted but the data displayed is wrong. 

Thing is it works on some comboboxes and not on others.

johnny-bit:
Oh for frog's snacks... this'll be in order of "shortcuts" options window.

## B
### base curve
1. exposure fusion: ok
2. preserve colors : ok
3. scale: ok
### basic adj
1. preserve colors: ok
## C
### channel mixer
1. destination: **BAD** msg: shows "channel mixer/ blue: 1.0000" or "channel mixer/blue: 0.0000"
### color balance:
1. color control: ok
2. mode: ok
### color look up table
1. target color: **BAD** msg: shows: "color look up table/ saturation: 0,00"
### color reconstruction
1. precedence: ok
### color zones
1. interpolation method: ok
2. process mode: ok
3. select by: ok
### crop and rotate
1. aspect ratio: ok
2. automatic cropping: ok
3. flip: ok
4. guide lines: ok
5. keystone: ok
## D
### defringe
1. operation mode: ok
### demosaic
1. color smoothing: ok
2. match greens: ok
3. method (bayer): ok
4. method (xtrans): ok
### denoise (profiled)
1. mode: ok
### dithering
1. method: ok
## E
### exposure
1. mode: ok

======

I'll test more later. Maybe this'll narrow the search down?
watch this space for updates

dterrahe:
The code 

```
  if(!gtk_widget_is_visible(GTK_WIDGET(w)) && *w->label)
  {
    if(w->module && w->module->multi_name[0] != '\0')
      dt_control_log(_("%s %s/%s: %s"), w->module->name(), w->module->multi_name, w->label, dt_bauhaus_combobox_get_text(combobox));
    else if(w->module)
      dt_control_log(_("%s/%s: %s"), w->module->name(), w->label, dt_bauhaus_combobox_get_text(combobox));
    else
      dt_control_log(_("%s"), dt_bauhaus_combobox_get_text(combobox));
  }

```
could be moved to dt_bauhaus_combobox_set  in bauhaus.c so it does not need to be duplicated in both next and prev routines.

elstoc:
I've worked out what the issue is with the local contrast mode switch: Changing the combobox (and sending the 'value-changed' signal runs `mode_callback` which calls `dt_bauhaus_slider_set` to set shadows and highlights to 50%, which calls `dt_bauhaus_slider_set_normalized`. This function displays the slider change, overwriting the combobox change popup.

I have fixed it but because of the way it's been fixed I can't also incorporate the changes suggested by @dterrahe. I'll push the new commit and you'll see what I mean.

johnny-bit:
try on channel mixer and color look up table.

as for code deduplication, i think you could put some static inline func?

elstoc:
Not one for this PR but I think properly, the current slider change log display in `dt_bauhaus_slider_set_normalized` should be moved _out_ of bauhaus.c because it should only ever really appear when executed via an accelerator callback.

johnny-bit:
Do another PR then. It's a very sensible observation imho... I wonder it that'll mess dynamic shortcut for sliders (remember I also would like dynamic key for comboboxes ;))

dterrahe:
> I have fixed it but because of the way it's been fixed I can't also incorporate the changes suggested by @dterrahe. I'll push the new commit and you'll see what I mean.

You are now doing an unconditional g_signal_emit_by_name(G_OBJECT(combobox), "value-changed"); immediately after calling dt_bauhaus_combobox_set when in that function there is already a conditional if(!darktable.gui->reset) g_signal_emit_by_name(G_OBJECT(w), "value-changed");

Are you sure this cannot lead to problems? I do remember when I added the pop-up for sliders there we tricky corner cases that depended on careful observation of gui->reset.

dterrahe:
> Not one for this PR but I think properly, the current slider change log display in `dt_bauhaus_slider_set_normalized` should be moved _out_ of bauhaus.c because it should only ever really appear when executed via an accelerator callback. Here, we're relying on the signal being handled before the subsequent message display.

My MIDI code also triggers this functionality. Basically, this should all be taken into account in the fundamental redesign of the input layer.

elstoc:
> You are now doing an unconditional g_signal_emit_by_name(G_OBJECT(combobox), "value-changed")

All I did was move the call (it was previously after the log message) but you're right, it is superfluous. And if I remove it I can also include your suggestion to move the log message into bauhaus.c

elstoc:
Now I remember why I put that logic into accelerators.c rather than bauhaus.c. It's because it's called by things other than keyboard shortcuts including when setting up comboboxes in the lighttable. So now I get a popup when I open darktable saying "sRGB (web-safe)".

johnny-bit:
Pffttt...

Seriously tho - check it against those tested etc for now.

For future - we need proper sane pluggable input layer.

dterrahe:
> So now I get a popup when I open darktable saying "sRGB (web-safe)".

Even when you add `if(!darktable.gui->reset)` ? It is possible that not all modules set this flag correctly; I remember having to fix some of those issues when I implemented this for sliders.

elstoc:
@dterrahe: I removed the emit of value-changed from my functions entirely and moved the log output to bauhaus.c:

```
void dt_bauhaus_combobox_set(GtkWidget *widget, int pos)
{
  dt_bauhaus_widget_t *w = DT_BAUHAUS_WIDGET(widget);
  if(w->type != DT_BAUHAUS_COMBOBOX) return;
  dt_bauhaus_combobox_data_t *d = &w->data.combobox;
  d->active = CLAMP(pos, -1, d->num_labels - 1);
  gtk_widget_queue_draw(GTK_WIDGET(w));
  if(!darktable.gui->reset) g_signal_emit_by_name(G_OBJECT(w), "value-changed");

  if(!gtk_widget_is_visible(GTK_WIDGET(w)) && *w->label)
  {
    if(w->module && w->module->multi_name[0] != '\0')
      dt_control_log(_("%s %s/%s: %s"), w->module->name(), w->module->multi_name, w->label, dt_bauhaus_combobox_get_text(widget));
    else if(w->module)
      dt_control_log(_("%s/%s: %s"), w->module->name(), w->label, dt_bauhaus_combobox_get_text(widget));
    else
      dt_control_log(_("%s"), dt_bauhaus_combobox_get_text(widget));
  }
}
```

@johnny-bit: the latest commit resolves issues with the modules you reference as well as with local contrast mode.

johnny-bit:
@elstoc - you mean commit 86a1532 ? or haven't you pushed new with changes to `bauhaus.c`?

elstoc:
> you mean commit 86a1532

Yes. I didn't push the changes to `bauhaus.c` because they had the side-effects I described above. As far as I'm concerned 86a1532 is fine, though I could probably do with removing the second call to `g_signal_emit_by_name` in my new functions.

dterrahe:
> > So now I get a popup when I open darktable saying "sRGB (web-safe)".
> 
> Even when you add `if(!darktable.gui->reset)` ? It is possible that not all modules set this flag correctly; I remember having to fix some of those issues when I implemented this for sliders.

bb08cdac3a346a76aa94efbce4318742e187245c is where TurboGit fixed a similar issue with a toast popping up from a slider being initialised when darktable was starting up.

elstoc:
I can't pin down where this is being done from. I suspect it's somewhere in the 'export selected' module but that module has calls to `dt_bauhaus_combobox_set` all over the place.

elstoc:
Beyond the duplication issue, are there other reasons for this code to be in bauhaus.c?

dterrahe:
> Beyond the duplication issue, are there other reasons for this code to be in bauhaus.c?

As @johnny-bit pointed out, you would get the pop-up "for free" if this is also implemented as a dynamic shortcut (and I could easily support it for MIDI as well).

And for consistency (so we don't achieve the exact same thing for sliders and comboboxes in different locations). This helps maintainability and makes the code base less daunting/surprising.

@TurboGit I experimented with setting gui->reset to 1 until the whole gui is set up. In darktable.c:
```
    darktable.gui = (dt_gui_gtk_t *)calloc(1, sizeof(dt_gui_gtk_t));
    if(dt_gui_gtk_init(darktable.gui))
    {
      fprintf(stderr, "ERROR: can't init gui, aborting.\n");
      return 1;
    }
    darktable.gui->reset = 1; // <<<<<
```
and
```
    darktable.undo = dt_undo_init();

    darktable.gui->reset = 0;  // <<<<<
```
It is necessary to revert bb08cda or gui->reset will be set back to 0 early.
I did not notice any breakage, but did not deeply test all functionality and this could of course cause initialisation problems _anywhere_ . Would you expect any?

@elstoc then the code below in bauhaus.c (notice all of the pop-up code is also conditional on !gui->reset) causes no unexpected polution at start-up.
```
void dt_bauhaus_combobox_set(GtkWidget *widget, int pos)
{
  dt_bauhaus_widget_t *w = DT_BAUHAUS_WIDGET(widget);
  if(w->type != DT_BAUHAUS_COMBOBOX) return;
  dt_bauhaus_combobox_data_t *d = &w->data.combobox;
  d->active = CLAMP(pos, -1, d->num_labels - 1);
  gtk_widget_queue_draw(GTK_WIDGET(w));
  if(!darktable.gui->reset) 
  {
    g_signal_emit_by_name(G_OBJECT(w), "value-changed");

    if(!gtk_widget_is_visible(GTK_WIDGET(w)) && *w->label)
    {
      if(w->module && w->module->multi_name[0] != '\0')
        dt_control_log(_("%s %s/%s: %s"), w->module->name(), w->module->multi_name, w->label, dt_bauhaus_combobox_get_text(widget));
      else if(w->module)
        dt_control_log(_("%s/%s: %s"), w->module->name(), w->label, dt_bauhaus_combobox_get_text(widget));
      else
        dt_control_log(_("%s"), dt_bauhaus_combobox_get_text(widget));
    }
  }
}
```

elstoc:
Thanks @dterrahe that was how I'd already altered bauhaus.c for testing.  Removal of bb08cda was obviously what I was missing. I'll wait for confirmation from @TurboGit before I push any more changes.

TurboGit:
> I did not notice any breakage, but did not deeply test all functionality and this could of course cause initialisation problems _anywhere_ . Would you expect any?

I would at least test that the message about resync of local copies is properly displayed with this change.

elstoc:
I've tested and can confirm that the message about resync of local copies is still displayed.

src/bauhaus/bauhaus.c
```diff
@@ -1197,7 +1197,20 @@ void dt_bauhaus_combobox_set(GtkWidget *widget, int pos)
...
   gtk_widget_queue_draw(GTK_WIDGET(w));
-  if(!darktable.gui->reset) g_signal_emit_by_name(G_OBJECT(w), "value-changed");
+  if(!darktable.gui->reset) 
+  {
+    g_signal_emit_by_name(G_OBJECT(w), "value-changed");
+
+    if(!gtk_widget_is_visible(GTK_WIDGET(w)) && *w->label)
+    {
+      if(w->module && w->module->multi_name[0] != '\0')
+        dt_control_log(_("%s %s/%s: %s"), w->module->name(), w->module->multi_name, w->label, dt_bauhaus_combobox_get_text(widget));
```
TurboGit: I would add a space around /

likewise 2 lines below.

elstoc:
@TurboGit I've just made a change to a similar toast when adjusting sliders in PR #4424 (in which I've also incorporated your above spacing suggestion). I'd like to replicate that change here for the combobox version but it'll be hard to test without first implementing #4424 because the change relates to multiple instances.

Is the best thing to review and merge #4424 to master first then merge master back into my branch before making the change? (In retrospect it might have been easier to have included both changes in the same PR)

PR #4477, opened 2020-03-15
new metadata field "version name"

### LINKED ISSUE #4115: Duplicate management : Naming and Export
When exporting duplicates, Darktable does not consider the duplicate name. It just keeps appending numbers to the original file name. For instance if i have image XYZ.jpg, and i create a duplicate named "ColorGraded" and another duplicated named "BW" and export, the new images will be XYZ_01, XYZ_02.jpg. It's difficult to figure out from the name of the exported images, which of the duplicate was used. 

The proposal is to append duplicate name as well to the exported image, from previous e.g. the exported images must be XYZ.jpg, XYX_ColorGraded.jpg & XYZ_BW.jpg.

Also, there always have to be default names for the duplicates. Right now it comes up blank, simple numbering/Lettering might work. The user can always modify to suite it.

Note : We might have to limit the size of duplicate title, to avoid too long names.

elstoc:
After some discussion, the following is a summary of the final agreed content of this PR:

Introduce a new metadata field "version name" which replaces use of title field in duplicate manager module.

The new field can be accessed through the variable `$(VERSION_NAME)` in the export module.

Also introduced with this PR is a new variable `$(VERSION_IF_MULTI)` which returns the same as the `$(VERSION)` variable, but only if there is more than one version of the current image (i.e. with the same file name and collection). If there is only one version, this variable returns NULL.

Resolves #4115


TurboGit:
@phweyland : I'd like to have your review on this, thanks.

phweyland:
> I'd like to have your review on this, thanks.

I just discover the subject and my understanding may be incomplete.

First, using a specific metadata is far better than using the `title` one (I would have disagreed with that :)).

But metadata are data which are intended to be exported (or not) with the final image. Not really the topic here. And this seems to be fully redundant with version info.

If I understand properly the purpose here is to improve version number which can be not explicit enough. Why not to change the numeric version field into a string field (can be defaulted by version number as today) ? That would serve the purpose of exported image and xmp as well.
$(VERSION) already exist as a variable and can be used manually  or automatically to name files.

Would that not be more appropriate than using metadata ?

NB: I've tried to see where version were used in dt but the version word is over-used ...

elstoc:
Hmm changing the version column probably hits quite a few bits of code. Certainly, you'd have to change the import, export and synchronization of xmp files plus, as you've discovered, with the fact it's called 'version' it might be hard to track down all the code that's using it. 

If you're creating a duplicate image, where all of the current version 'numbers' are freeformat text, how do you automatically generate the version of the duplicate (you can't just increment the number)? If you change the name of a duplicate, do you have to delete the existing xmp sidecar and create another one (since version number is used to ensure unique sidecar files)? Currently version number is entirely internal so there's no need to ensure it's not duplicated. We'd have to implement something to check for uniqueness if we wanted it to be a user-editable field.

It just seems like a whole lot of pain for not a whole lot of gain.  All things being equal it might be the best way to go, but I'm not sure it's worth the effort.

I think the main purpose (for me at least) is to be able to uniquely name duplicate images on export with something more meaningful than an integer (and something other than the existing title field - which I already use to generate more verbose text on my photo website) and I'm totally comfortable with that being independent of the internal version number. Perhaps the two issues wouldn't have been conflated if I hadn't included the default but again, that's more about unique default names than anything else (to keep the exported files unique, if I'm not bothered about being descriptive).

I kind of like that it's exported with the image.

phweyland:
> changing the version column probably hits quite a few bits of code.

Agreed. But using a functionality (metadata) to fix the version issue (not clear file naming) without understanding the code, just because it's easy, doesn't sound right to me.


> you can't just increment the number

Just convert it.

> We'd have to implement something to check for uniqueness if we wanted it to be a user-editable field.

You will have to do it in any case because the name given by the user is not guaranteed to be unique.

I don't say that using the metadata is necessarily wrong but easiness only is not relevant. 
Pros and cons need to be carefully evaluated.
The current usage of version number and its relationship with file naming should be understood. I don't agree for skipping this.
Also, creating a metadata item specific for duplicates introduces a new type of metadata which is not applicable to all images (!?). This will have to be managed in a specific way. Again not necessarily wrong but needs to be integrated in a global logic and should not be a particular case.

elstoc:
> But using a functionality (metadata) to fix the version issue (not clear file naming) 

My intention was not to fix a version issue but to introduce new metadata to use in export file naming. Because its only use is for export, it's up to the user to care about uniqueness (internally darktable won't care).

> Just convert it.

I was referring to the case where none of the existing versions are numeric and you choose to create a new one (with duplicate or clone). You can't convert a string to numeric and then increment it.

> You will have to do it in any case because the name given by the user is not guaranteed to be unique.

See above comment. If the user wants unique export names, they should key unique values into the new metadata field. Not doing so would not cause an issue, unless we decide to also use it internally as a unique key. A user could decide to not use this field for file naming at all, in which case they would still not get unique filenames.

phweyland:
> My intention was not to fix a version issue but to introduce new metadata to use in export file naming. Because its only use is for export, it's up to the user to care about uniqueness (internally darktable won't care).

So I'm afraid I disagree with both statements.

Metadata is not the right place to deal with duplicate names.

As already said your new metadata item will appear inappropriately in metadata list (not applicable to all images).
At the same time, it will be (as title today) a duplicate of the name field in duplicate module. No reason to have this defined in 2 places.

elstoc:
I don't know - perhaps we're getting hung up on the word duplicate. I can see myself using this field for file naming _even if I only had one version_ since I'd want to put more verbose things in the title and description fields.  Maybe the duplicate module should only have the version number in it (read-only as currently) rather than duplicating a metadata field?

phweyland:
>  perhaps we're getting hung up on the word duplicate. I can see myself using this field for file naming _even if I only had one version_ since I'd want to put more verbose things in the title and description fields. 

Sounds generic if applicable to all images. You convinced me here ! :)

> Maybe the duplicate module should only have the version number in it (read-only as currently) rather than duplicating a metadata field?

I would let the field in duplicate module. The user keeps the possibility to mask this item in metadata module, depending if he uses that feature only for duplicates or not.

So just remain to find a name less related to duplicate, right ?

elstoc:
Thanks @phweyland. Are you saying here that all you want me to do is rename the field? Or should I remove all that defaulting to version number stuff as well (it's not strictly necessary and does confuse the issue slightly)? 

I suppose if you wanted to ensure unique filenames, a user could use the $(VERSION) number field **plus** the new field? And removing the default values does make it more consistent with other metadata fields.

Naming things is hard. 

I assume you're happy for me to leave it in Xmp.darktable? I can't see another one in the Xmp.dc set that makes sense. Maybe call it 'short description/name/title'?

phweyland:
> I assume you're happy for me to leave it in Xmp.darktable?

Happy ? Not completely :). I am still concerned by the usage of metadata for that purpose (I think the right way is to work on image version or equivalent) but at least as I understand it now it's generic and not a particular case. 

I've given my standpoint. The dev team member have certainly their view.

Using metadata has some implications (maybe not desirable in that case):
If not hidden in metadata module, the field will appear in:
- import module
- metadata_view module
- in export_metadata the user will have to put = ""  in case he doesn't want to export it.

In all cases, it has to be included in metadata preset (weird if related to duplicate module, right ?).
To make it work you have to add it in the corresponding legacy_params part.

> should I remove all that defaulting to version number stuff as well (it's not strictly necessary and does confuse the issue slightly)?

In metadata module (and in meta_data table as well) we should not have any version number, I agree, only user text if defined. I would remove the corresponding changes in libs/metadata.c

In variables.c I would remove it also the version substitution. The file naming has already the syntax to deal with the absence of data: `$(SHORT_DESCRIPTION-$(VERSION))` if `SHORT_DESCRIPTION` is our variable name. File naming will use version if short_description is not defined.

I guess the export file naming uses today the version in case of duplicate (as this is not a user setting). I don't know where. This part should be updated to integrate the `short description`.

We could remove the field from duplicate module. The issue I see here is that on darkroom there is no direct access to metadata. So to have it in duplicate module seems to be necessary at this stage (we could make metadata visible in darkroom like tags. Maybe not too much UX however).

elstoc:
> To make it work you have to add it in the corresponding legacy_params part.

I might need a hand with that - not sure how it's supposed to work

phweyland:
> > To make it work you have to add it in the corresponding legacy_params part.
> 
> I might need a hand with that - not sure how it's supposed to work

in legacy_params() (libs?metadata.c) you should just add a string null terminator for the new metadata (this updates the existing presets):
```
  else if(old_version == 2)
  {
    size_t new_params_size = old_params_size + 1;
    char *new_params = calloc(sizeof(char), new_params_size);

    memcpy(new_params, old_params, old_params_size);

    *new_size = new_params_size;
    *new_version = 3;
    return new_params;
  }
```

and set:
`DT_MODULE(3)`

todd-prior:
Guys @elstoc @phweyland Thanks for working on this....I think it will be useful. If I have an image and i do some color grading on it maybe call one basic, one warmer, one cooler, one monochrome...it would be great to incorporate that into the exported name and or the xmp. It seemed a bit strange to me that you were offered the ability to add names in the duplicate panel but then not to use them for any purpose to help identify and organize your images..you would have to currently identify and export each image one at a time and add a text string to identify them as such...exported images are often viewed in the operating system and or other software or devices not equipped with Darktable and I think this would be a nice way to manage exported filenames.....

phweyland:
@todd-prior I think we all agree about the goal :). I even believe that xmp files should be included as well.

The discussion above is only about the way to achieve it...

phweyland:
> all one basic, one warmer, one cooler, one monochrome..

Those sound like categories. Instead of metadata, tags could be used ... (just another way)

elstoc:
As discussed, I've now changed this so it's essentially a new metadata item only, and is used in place of "title" in the duplicate manager.  A new variable $(SHORT_DESCRIPTION) can also be used to access this metadata in the export module.

phweyland:
I've looked at the code for export file naming. In fact everything is in the hands of the user for duplicate naming. Currently, for duplicates, the user has to define himself in export target storage:
`$(FILE_NAME)_$(TITLE-$(VERSION))`
With your change it becomes:
`$(FILE_NAME)_$(SHORT_DESCRIPTION-$(VERSION))`

Instead of `$(SHORT_DESCRIPTION)` the user can still use $(TITLE) if he wants to, or a category (like `$(CATEGORY0(whatever_category)) `).

Except the implicit relationship between the name in duplicate and the metadata `short description` everything is generic. As far as I can see your new code is in line with our discussion.

phweyland:
Playing around with duplicate export I still find some difficulties.
If I prepare the file naming as above for duplicates but I don't have duplicate on some images, those images end up with `_0` as version in the filename. A bit weird.

I haven't found the way to remove this useless `_0`. Do I miss something ?

Thoughts.
The duplicate feature should assume the naming differentiation of exported images as it does for xmp (and not let it in the user's hands as today).
An elegant way (to be checked carefully though) could be in the expansion of the variable $(FILE_NAME).
If no duplicate (not necessarily version 0 ?) $(FILE_NAME) should operate as today.
If duplicate name defined, append _<name> at the end of $(FILENAME)
else append _<version> (as done for xmp).

The relationship with whatever metadata would not be necessary anymore.
We could make the relationship with `short description` configurable to let the user use it as metadata.

elstoc:
> those images end up with _0 as version in the filename

Yes. xmp files manage this by having the first version not have any suffix, then having each subsequent version suffixed by the version number (_01, _02...). The export module tooltip states that it supports 'variables support bash like string manipulation' but I'm not entirely sure what that does or whether it could help remove the _0 suffix.  We could change $(VERSION) to return empty if it's v0 but you might still be left with the underscore.

Regarding your second suggestion, there might be users who want to control file name formatting entirely (not use the $(FILE_NAME) variable at all or want it to work how it does today). Personally I'd prefer leaving as much control in the users hands as possible in this module.

todd-prior:
Ya I am sure there are ways to do it…but that was just an example off the top of my head and not maybe a clear strategy all the time. I can actually use the title I think if its available as $(Title) as I would not use it for anything else and my primary need would be for exporting not xmp names but that would be full control for sure

phweyland:
> Regarding your second suggestion, there might be users who want to control file name formatting entirely (not use the $(FILE_NAME) variable at all or want it to work how it does today). Personally I'd prefer leaving as much control in the users hands as possible in this module.

If the user wants not to use  `$(FILE_NAME)` he still has all the control as today plus the new `short description`.
But at least there would be a basic mechanism which does not let a user with a problem to solve !

elstoc:
Currently, if I export an image, say IMGA.jpg, and then do so again, with 'on conflict - create unique filename', darktable will automatically create IMGA_01.jpg. 

There's a danger that if I have a duplicate image, it will also export as IMGA_01.jpg meaning that I might not be able to tell whether it's the second export of the original or the first export of the duplicate. I would therefore personally prefer to export using $(FILE_NAME)_$(VERSION) even for version 0 so as not to get confused about the naming conventions. In this case, the first export of v0 would export as IMGA_0.jpg and the second export of the same file would be IMGA_0_01.jpg.

Not being able to remove the underscore separator if the associated field is empty is a general issue with the export module, which I don't think can be resolved without a code change, based on looking at the manual. It is possible to remove the '0' if v0 - using $(VERSION/0/) - but not the underscore.

Also what you suggest implies explicitly using the new metadata field ('<name>' in your example) as a duplicate version name, which I thought we were trying to avoid.

phweyland:
> Not being able to remove the underscore separator if the associated field is empty is a general issue with the export module,

I can agree with that. But with the same principle this is not the responsibility of export module neither of metadata module to solve the general issue of duplicate module about filename. We are here only because you find simpler and less work to use metadata and variables instead of resolving the duplicate module issue.

> Also what you suggest implies explicitly using the new metadata field ('' in your example) as a duplicate version name

That's a way to see it. But let's say that the duplicate name is used to manage the filename, without talking of metadata (could be saved as a parameter of duplicate module for that duplicate image).  That would work without metadata.
But if we (or the user) find that useful, we can reuse this entry to set up a metadata (short description) to save manual input.

elstoc:
> We are here only because you find simpler and less work to use metadata

You are right, to some extent, but mainly because I saw the main issue being that an extra metadata field was required, whereas you think there is a more general issue with having the version number be an integer. I don't see that as a problem and am happy that a simple metadata field solves the issue I had. 

It sounds like we're probably not going to agree on this so I'm happy to close the PR and leave it to someone else to come up with a more comprehensive solution (TBH for me the $(VERSION) variable on its own solves most of my issues)

phweyland:
> It sounds like we're probably not going to agree on this

Yes, we have still some disagreements but we have made progress ! :)

I would say that, as is, it's really better than the title solution.
This is still a workaround, which probably helps some users when they face the duplicate filename issue. But you haven't added any dependency, which is safe.
Last, it offers the new `short description` metadata which can be useful in itself.

So for these reasons, the PR deserves to be merged.

That does mean the duplicate module filename issue is solved. At the very least because the user will have to find the workaround, and once found, it is not general and doesn't apply nicely in every situation.

EC1000:
> Naming things is hard.

Just a personnal opinion about the new field name: I find that `SHORT_DESCRIPTION` is too generic and not explicit about its purpose. There are already `TITLE` and `DESCRIPTION` standard metadata fields (for which you already have to decide what kind of info you will put into each field). Without knowing the purpose of this new field, the wording `SHORT_DESCRIPTION` seems kind of in between these two other fields... 
I would suggest `VERSION_DESCRIPTION` or `VERSION_NAME` or `VERSION_TITLE` to be more explicit (except if the idea is precisly to be less specific so users can use that field for another purpose ??).

elstoc:
> except if the idea is precisly to be less specific

That was the idea, though I agree that in terms of the name it sounds more like it's halfway between title and description - perhaps `SHORT TITLE` would make it clearer? Certainly I intend to use it for something shorter than my current use of `TITLE`.

elstoc:
Just a thought, we could create a new variable $(FILE_NAME_VERSION) that matches the naming convention of the xmp files (i.e. is equal to `$(FILE_NAME)` for v0 and equal to `$(FILE_NAME)_$(VERSION)` for all others. That way people have the option to continue using `$(FILE_NAME)` with its previous functionality if they like.

johnny-bit:
That's IMO good idea...
but two exports of the same file is weirder imo that in "uniqe" mode it does simply `filename_xx.JPG` where `_xx` is empty for 1st export and `_[0-9][0-9]` for subsequent instead of having `_xx` configureable.

Hmmm... Should I file a FR for "unique filename" export option to have suffix configureable? I'd personally change my suffix for unique filename to `$(EXPORT_DATETIME)` :grin:

elstoc:
Yes too many numbers all over the place meaning different things, hence my suggestion of "_v1,2,3" rather than just "_1,2,3". Perhaps if the number was prefixed by something that made it obvious it was another export of the same file - something like "(export 2)" for the second export, maybe.

elstoc:
There are issues similar issues with versions and xmp file names (see https://github.com/darktable-org/darktable/issues/4241#issuecomment-598260285) that could be resolved with better file naming conventions (e.g. put the version number immediately before the .xmp extension). Also bear in mind that if I have only one version of a file that doesn't necessarily mean it's version 0 - I could have duplicated it and then removed v0 (see https://github.com/darktable-org/darktable/issues/4385). It's all a bit of a minefield.

johnny-bit:
> Perhaps if the number was prefixed by something that made it obvious it was another export of the same file

too many people would like too many things with export. Me? I'd configure it to be export date & time. You? `(export $(num))`, some other person? something like `_v$(num)`. Someone else - "bring back default `_$(num with leading 0)`!!!"... That's why: "make it configureable" :)

elstoc:
> That's why: "make it configureable"

Agreed

phweyland:
> we could create a new variable $(FILE_NAME_VERSION) that matches the naming convention of the xmp files (i.e. is equal to `$(FILE_NAME)` for v0 and equal to `$(FILE_NAME)_$(VERSION)` for all others.

That makes sense to me. 
1. set as a default for export file naming it works out of the box (no unexpected issue related to  naming problem with duplicates).
2. it is compatible with today settings. But the user is still free to change it.
3. could work without metadata.
4. metadata can be linked to this process like it's done with this PR. Could be configurable though.

elstoc:
I edited my original comment to include another idea...

> To avoid the confusion with the 'two exports of the same file' issue I mentioned above, we could have the suffix be "_v1, _v2" etc.

Also, you say it could be configurable. How would you envisage this working? What would the configuration option be?

phweyland:
> Also, you say it could be configurable. How would you envisage this working? What would the configuration option be?

The idea is to make the relationship "duplicate name" <-> "metadata "short description" (or another one) optional.
So by default duplicate name could have been saved as image duplicate parameter without involving metadata.
Then, a configuration point could have linked  "duplicate name" <-> "metadata "short description" to avoid user double entry in case he wants to have this info also as a metadata.

As things are already implemented, I'm not sure the rework is worth to be done.

>  is equal to `$(FILE_NAME)` for v0 and equal to `$(FILE_NAME)_$(VERSION)` for all others.

1. it happened to me to delete v0 and keep only v1. Do I get the suffix `_v1` ? Maybe if there is no duplicate for that image and no given duplicate name, the version number could be also skipped.

2. `$(FILE_NAME)_$(VERSION)` should be replaced by `$(FILE_NAME)_$(SHORT_TITLE-$(VERSION))`

elstoc:
If I populated $(SHORT_DESCRIPTION) I'd probably want to use it even if there was only one version of the image. Much of what you suggest could be made easier if the export module itself treated the underscore as a delimiter and removed it if it was followed by an empty variable. The only thing that wouldn't resolve is if you wanted to remove the version entirely in the case of only one version existing. For that you could possibly introduce a new variable (though what you'd call it...)? 

Then your suggestion could be accomplished with `$(FILE_NAME)_$(SHORT_DESCRIPTION-$(VERSION_IF_MULTI))`, which would return $(FILE_NAME) if there was no short description and only one version (because the trailing underscore would be automatically removed).

phweyland:
> If I populated $(SHORT_DESCRIPTION) I'd probably want to use it even if there was only one version of the image

Agreed. It's what I've tried to said.



> Much of what you suggest could be made easier if the export module itself treated the underscore as a delimiter and removed it if it was followed by an empty variable.

True. That's  another way to fix the issue of the workaround :).

What I liked about your $(FILE_NAME_VERSION) suggestion is that's actually a duplicate feature. You remove the duplicate module your can remove $(FILE_NAME_VERSION), whatever the artifices you put in it. You don't constrain or pollute other functions to make duplicate work properly.

elstoc:
> True. That's another way to fix the issue of the workaround :).

I get that you're saying that a bit 'tongue-in-cheek' but the issue of including variables that might be empty in an export name is a more general issue that this would also resolve. I could always include $(SHORT_DESCRIPTION) in my export name but if it's blank I'd still be left with a standalone delimiter. Same goes for $(TITLE), category and all other metadata items that could be empty.

phweyland:
>  the issue of including variables that might be empty in an export name is a more general issue 

Agreed.

todd-prior:
In a perfect world it would be set-up like the collections module so you could add items variables and text and string then together…..The analogy would be the filter criteria would be the name…..ON1 uses this type of setup for export ..in the export dialog they have a rename option button …if you check it then you can add from a series of drop downs to construct a name…gives you pretty decent flexibility….why not consider maybe leveraging the collections filters code if possible and make up something like that….its maybe bigger than what you are thinking of but I think it would be good and it would be clear how you were naming your files….

todd-prior:
In the scenario  above the duplicate panel text boxes would if empty supply 0,1,2 etc and if populated would supply a string that would be linked into a DB variable called $(duplicate) that variable could be called from the filter setup described above to construct the file name...would it really need to be in the metadata or could it just be internal to DT???  Clearly I am way outside my understanding so forgive me.....

phweyland:
> ON1 uses this type of setup for export ..

The filename rules of export module are probably less easy to use but provides the same functionalities. 

> You could add one addtional tick box to that to rename the xmp or leave as 01, 02 or v1 v2

Excellent. But probably not in the export module, right ? Better somewhere in general configuration (along with xmp settings ?).

> the duplicate panel text boxes would if empty supply 0,1,2 etc and if populated would supply a string that would be linked into a DB variable called $(duplicate) that variable could be called from the filter setup described above to construct the file name

That's exactly what $(FILE_NAME)_$(SHORT_DESCRIPTION-$VERSION)) syntax does. Except we would like not to add the useless `_v0` when the image has no duplicate and duplicate module doesn't give any name.

That's where the @elstoc `$(FILE_NAME_VERSION)` suggestion is very interesting because it encapsulates the treatment of the above specific case without polluting anything. And more important for me, it works out of the box without exposing the user to duplicate naming issues and complexity. But the interested user is still able to elaborate the file naming rule he needs...

> would it really need to be in the metadata ?

No, variables are enough as soon as information are stored in db. Even if the user wants to export the information as metadata he can use the same variables.

To complete this point it's clear that using metadata mechanisms avoids to add a specific column somewhere in the db, at the price to pollute somehow the metadata module. But as the user can now hide the metadata he is not interested in, this is less a problem (and this one could even be hidden by default).

elstoc:
> That's where the @elstoc $(FILE_NAME_VERSION) suggestion is very interesting because it encapsulates the treatment of the above specific case without polluting anything.

I know I suggested it in the first place and am more than happy to include it but, combined with the changes being discussed in https://github.com/darktable-org/darktable/issues/4485 (namely ability to define how export conflicts are named and removal of delimiters where variables are empty) I'm not sure it's really necessary. I now prefer my alternative of $(VERSION_IF_MULTI) which the user can then combine with other metadata if they wish.

How about, for this PR, we restrict to the following:

- New metadata field (should we rename it to "short title"? - I think I prefer that)
- `$(FILE_NAME_VERSION)` variable, which returns `$(FILE_NAME)` if there's only one version, and `$(FILE_NAME)_$(VERSION)` if multiple. (or are you proposing that this should work differently to my initial proposal?)
- `$(VERSION_IF_MULTI)` variable which returns `$(VERSION)` only if there is more than one version present

The remainder of the issues can then be handled by the changes to the export module being proposed in https://github.com/darktable-org/darktable/issues/4485.

EC1000:
> > except if the idea is precisly to be less specific
> 
> That was the idea, though I agree that in terms of the name it sounds more like it's halfway between title and description - perhaps `SHORT TITLE` would make it clearer? Certainly I intend to use it for something shorter than my current use of `TITLE`.

Sorry, but I don't find `SHORT TITLE` clearer! :D 
If the goal is: 
- to add a info concerning the version (whether there is only one version or more), I think it should be named accordingly i.e. something with "version" in it, 
- to use it for naming export files, I think it should indeed reflect that it is designed to be relatively short (some users will probably also use it for a longer description of what they did to a specific version, but the goal here seems to be file-naming oriented). 

I would go for `VERSION_TITLE`, but that is just my opinion. 

Or do you have another possible workflow/use in mind for this field that would justify something less specific?

> > would it really need to be in the metadata ?
> 
> No, variables are enough as soon as information are stored in db. Even if the user wants to export the information as metadata he can use the same variables.
> 
> To complete this point it's clear that using metadata mechanisms avoids to add a specific column somewhere in the db, at the price to pollute somehow the metadata module. But as the user can now hide the metadata he is not interested in, this is less a problem (and this one could even be hidden by default).

Just to (try to) understand a little bit more the development aspect: it is easier to add it in a new metadata field rather than in a specific field? And if it is in the metadata, by defaut it would be exported as so (I know it is possible to set a metatada field not to be exported)?

From a user perspective I think it would be more logical not to export this info by default in metadata, and be able to use it when wanted (to name export file, or maybe sometimes to populate the title or description metadata fields). 
But that is just a end-user perspective, I don't know the technical advantage/simplicity to add it as a metadata field (if someone has the time to enlighten me, I would be interested, because the drawback of adding a new entry in the database seems to be recurrent, and also I have not understand why adding it as a metadata field is easier).

todd-prior:
How about VERSION_LABEL  ???  or VERSION_ID
 
Sent from 
Mail for Windows 10
 

From: EC1000
Sent: March 17, 2020 8:20 PM
To: darktable-org/darktable
Cc: Prior,Todd; 
Mention
Subject: Re: [darktable-org/darktable] new metadata field "short description" (#4477)

 





except if the idea is precisly to be less specific

That was the idea, though I agree that in terms of the name it sounds more like it's halfway between title and description - perhaps
SHORT TITLE would make it clearer? Certainly I intend to use it for something shorter than my current use of
TITLE.

Sorry, but I don't find SHORT TITLE clearer! :D
If the goal is:

to add a info concerning the version (whether there is only one version or more), I think it should be named accordingly i.e. something with "version" in it,
to use it for naming export files, I think it should indeed reflect that it is designed to be relatively short (some users will probably also use it for a longer description of what they did to a specific version, but the goal here seems to be file-naming
 oriented). 
I would go for VERSION_TITLE, but that is just my opinion.
Or do you have another possible workflow/use in mind for this field that would justify something less specific?


would it really need to be in the metadata ?

No, variables are enough as soon as information are stored in db. Even if the user wants to export the information as metadata he can use the same variables.
To complete this point it's clear that using metadata mechanisms avoids to add a specific column somewhere in the db, at the price to pollute somehow the metadata module. But as the user can now hide the metadata he is not interested in, this is less a problem
 (and this one could even be hidden by default).

Just to (try to) understand a little bit more the development aspect: it is easier to add it in a new metadata field rather than in a specific field? And if it is in the metadata, by defaut it would be exported as so (I know it is possible to set a metatada
 field not to be exported)?
From a user perspective I think it would be more logical not to export this info by default in metadata, and be able to use it when wanted (to name export file, or maybe sometimes to populate the title or description metadata fields).
But that is just a end-user perspective, I don't know the technical advantage/simplicity to add it as a metadata field (if someone has the time to enlighten me, I would be interested, because the drawback of adding a new entry in the database seems to be recurrent,
 and also I have not understand why adding it as a metadata field is easier).
—
You are receiving this because you were mentioned.
Reply to this email directly, 
view it on GitHub, or 
unsubscribe.

elstoc:
> I would go for VERSION_TITLE

> How about VERSION_LABEL  ???  or VERSION_ID

The idea of changing it was to be less specific, so that it can also be used where there are no duplicates. I intend to use it on all my images, duplicate or not. The suggestions above will lead users to think that this metadata field should only be used for a single purpose, in which case I do kind of agree with @phweyland that it probably should not be in the metadata editor. It should also be included in sidecar files in case you want to reload your database, and metadata is a good way to do this.  We could possibly include a preference item for users to choose what field is displayed in the duplicate manager, making it clear that all metadata items are intended to be generic.

phweyland:
> Just to (try to) understand a little bit more the development aspect: it is easier to add it in a new metadata field rather than in a specific field? 

Metadata mechanisms include the following:
- read/saved from/into db (only when needed)
- read/saved from/into xmp
- UI available (metadata module, image information, export configuration)
- exportable along with images
- with a minimum of effort it's visible in collection

For a new column the developer has to:
- modify the db, code the db migration,
- code input/output db
- code the UI
- code the usage

However the benefits are limited to text info. We have had some discussion about images time stamps recently. It has been found better to add new columns the db images table.

> And if it is in the metadata, by defaut it would be exported as so (I know it is possible to set a metatada field not to be exported)?

Here I think that everyone can have good reasons to have it or not in metadata by default. That's why it should be configurable.

Now, but this is a general subject (another PR :) ), we could easily improve the metadata configuration to let the coders use metadata mechanisms without polluting user's metadata, at least per default. 
For this it would be enough to add a parameter to metadata hard coded configuration saying :
- this metadata is purely internal, no interest for user => hidden
- this metadata is internal, but could be useful for the user => hidden by default but the user can make it visible (typically the case of this PR)
- this metadata is user one => the user can hide it if he wants to (as today)

phweyland:
Let's try to resume the convergences.

we would like to:
 - replace duplicate usage of `title` metadata by a more specific one.  
 - avoid polluting user metadata (but metadata is useful to saved the duplicate name)
 - avoid the version in export file naming when there is no duplicates.
 - to be able to use the duplicate name for export file naming

For each point the answer could be:
 - already done. Considering the next point, the metadata should be named for what it is : `version name` or `duplicate name` instead of `short description`
 - I prepare another PR to allow the creation of per default not user metadata as explained at the end of the post [above](https://github.com/darktable-org/darktable/pull/4477#issuecomment-600574385). The user will have the ability to make it visible if he wants to.
 - the @elstoc proposal of `$(VERSION_IF_MULTI)` fulfills the need. Could be integrated in that PR.
- already done. With `$(SHORT_DESCRIPTION)` replaced by `$(VERSION_NAME)` or `$(DUPLICATE_NAME)`


What do you think ?

elstoc:
> the metadata should be named for what it is : version name or duplicate name instead of short description

I thought this was exactly the thing we were trying to avoid. If that's what it takes to get the PR merged (I just want a new metadata field) then I'm ok with it but on reflection I would much prefer something more generic since I will likely be using it more for single images than for duplicates. Naming it like that takes a generic metadata field and suggests to users that it's for a specific purpose ("you should only use for distinguishing between multiple versions"), which it needn't be. I would prefer `$(SHORT_NAME)`. From your suggestions, though, perhaps `$(VERSION_NAME)` is the better one as it can be used on all versions, not just the duplicates.

I know that I started this by calling it dup_name but you persuaded me that metadata should be generic!

phweyland:
>  thought this was exactly the thing we were trying to avoid.

That's what the change on metadata

phweyland:
> I thought this was exactly the thing we were trying to avoid.

That's what the PR to manage Metadata type is for.

> Perhaps $(VERSION_NAME) is the better one as it can be used on all versions,

Let's go for this one then.

elstoc:
Ok. Do you have any objection to me including `$(VERSION_IF_MULTI)` in this PR as well?

phweyland:
> Ok. Do you have any objection to me including $(VERSION_IF_MULTI) in this PR as well?

It's what I've suggested.
If we have the agreement on the 4 points above and their solution, I think we have a clear package.

elstoc:
Sorry I read it as being in your new PR. Anyway, I've updated the OP with a summary of the agreed content of this PR and will start work on the changes. Could you give it a once over please just to make sure it is what we agreed!

phweyland:
>  Could you give it a once over please just to make sure it is what we agreed!

I'm OK with it. Hoping that the PR to hide this metadata will be accepted :)

elstoc:
Ok I think that should do it.

src/common/variables.c
```diff
@@ -252,6 +255,40 @@ static char *get_base_value(dt_variables_params_t *params, char **variable)
...
+      result = g_strdup((char *)res->data);
+    }
+    g_list_free_full(res, &g_free);
+  }
+  else if(has_prefix(variable, "VERSION_IF_MULTI"))
+  {
+    sqlite3_stmt *stmt;
+
+    // count duplicates
+    DT_DEBUG_SQLITE3_PREPARE_V2(dt_database_get(darktable.db),
```
phweyland: is not simpler something like that:
SELECT COUNT(*) FROM main.images AS i1
JOIN main.images AS i2 ON i2.film_id = i1.film_id AND i2.filename = i1.filename 
WHERE i1.id = ?1

?

src/gui/gtkentry.c
```diff
@@ -186,6 +186,8 @@ const dt_gtkentry_completion_spec *dt_gtkentry_get_default_path_compl_list()
           { "FILE_NAME", N_("$(FILE_NAME) - basename of the input image") },
           { "FILE_EXTENSION", N_("$(FILE_EXTENSION) - extension of the input image") },
           { "VERSION", N_("$(VERSION) - duplicate version") },
+          { "VERSION_IF_MULTI", N_("$(VERSION_IF_MULTI) - same as $(VERSION) only if multiple versions exist") },
```
phweyland: "same as $(VERSION) but null string if unique version" ?

src/common/variables.c
elstoc: Gives the same result. Don't know enough about sqlite to know if it's more efficient (I'm an Oracle developer in my day job). Happy to change if you think it's better.

src/gui/gtkentry.c
elstoc: I'm not sure if 'unique version' is clear enough (since each of multiple versions could be considered unique). How about 'null string if single version exists'.

phweyland: Perfect !

src/common/variables.c
phweyland: I don't know neither. The syntax just more familiar to me.

src/common/variables.c
```diff
@@ -64,6 +65,8 @@ typedef struct dt_variables_data_t
 
   uint32_t tags_flags;
 
+  int count_versios;
```
phweyland: I thought I had already commented this ...
Is this count used ? (I've not seen it else where)
If yes, should not be count_versions ?

elstoc: d'oh you're right. I started out adding it to this struct but then realised the new variable would be so rarely used it would just be inefficient. I'll remove it.

elstoc:
@phweyland I've incorporated your comments. I've left the SELECT as it was for now but I'm happy to change it if you want.

phweyland:
> I've incorporated your comments. I've left the SELECT as it was for now but I'm happy to change it if you want.

That's OK for me. Thanks

elstoc:
@phweyland I'm just looking at incorporating the changes from your PR into this and I noticed that if I set the new metadata field as hidden by default it's not possible to alter it from the duplicate manager. This is because the function `dt_metadata_set` does nothing if the metadata column is hidden. Am I ok to remove this code or will it have side-effects on the metadata editor?

phweyland:
> This is because the function `dt_metadata_set` does nothing if the metadata column is hidden. Am I ok to remove this code or will it have side-effects on the metadata editor?

This will have side effects. This is a good catch....

Explanation. Hidden means that the user is not interested with that metadata, and as such it is not imported. I did that for sake of simplicity.
But here you demonstrate it is not enough. We cannot use the hidden flag for more than what it says (just hide) and we need another control for importation. 
And maybe another one for exportation... 
This requires some adjustments.

So as is, we have only 2 options for this PR. Either to make it internal (not used in `dt_metadata_set()`) or to make it always visible. I'm afraid that both of them will let some dissatisfaction.

Would you mind to be patient the time I make the necessary changes (at least for importation) ?

elstoc:
That's fine by me. Would it be better to just close this PR and incorporate its changes into yours? After all this one isn't all that complicated.

phweyland:
> Would it be better to just close this PR and incorporate its changes into yours?

I'd prefer not. 
Not to be dependent, I suggest you set this new item as "internal". This should be fully functional.
The user has just to use only duplicate module to set up `duplicate name / $(VERSION_NAME)`.
When `hidden` becomes more flexible you will still be able to move it there if you think that's better.

elstoc:
Ok thanks @phweyland. That's done.  @TurboGit this is ready for review now.

phweyland:
@elstoc hopefully with #4559 both internal and hidden by default metadata should work with this PR.
Could you check and let us know if that's really the case ?
Thanks

elstoc:
@phweyland if I set my new metadata to either internal (DT_METADATA_TYPE_INTERNAL) or hidden by default (DT_METADATA_TYPE_OPTIONAL) the new field doesn't appear in the metadata editor - basically same behaviour for both settings. Is that intentional?

elstoc:
@phweyland: I have four different versions of darktable on the go and was building the wrong one. I can confirm that both internal and hidden-by-default work as expected.  I assume you're happy for me to change this back to 'hidden by default'?

phweyland:
> I can confirm that both internal and hidden-by-default work as expected

Thanks a lot for that.

> I assume you're happy for me to change this back to 'hidden by default'?

I would let it internal, but this not a strong opinion.
'hidden by default' is done for that purpose, but in that specific case we create 2 entries for the same info, which has some inherent drawbacks.
This gives room for potential surprise for a not well informed user. For example clearing metadata clears all visible metadata for that image (and duplicate names). Copy/paste/clear metadata on selected images could also impact the duplicate module.
The advantage of internal is that there is only one place to change the duplicate name. 'hidden by default' has the same effect as long as ... it remains hidden.
I know your preference is this last option. It would be interesting to get the opinion of other people.

Pascal or others, any thought ?

elstoc:
TBH I would prefer it was left _visible_ in the metadata editor by default. I edit my metadata in bulk from the lighttable and having to go into the darkroom just to check I've populated the new field would be highly inconvenient. I suspect if it was left internal I'd end up ignoring the new field and repurposing some other metadata field instead.

Ideally I'd have left it with a generic name and added a preference to choose which metadata field to use in the duplicate module.

TurboGit:
@phweyland : No strong opinion, fact is I'm not a big user of metadata so my vote would certainly not be the best. I'd like other people to comment on this.

phweyland:
Not a lot of comments ... :(

> Ideally I'd have left it with a generic name and added a preference to choose which metadata field to use in the duplicate module (perhaps even via a combobox in the duplicate module itself).

That could be an idea. This said the feature doesn't work without metadata. So a default one remains mandatory.

I suggest to go with the `hidden by default` option. By default there is no unexpected behavior for the user. That's fine. Only if he sets it as shown, he needs to be informed about the relationship with duplicate module. This can be done through a tooltip in configuration popup. I can add this later one.

elstoc:
Ok I'll push an update.

elstoc:
I think this is finished now(?)

houz:
About the version field, it is an internal thing for dt and shouldn't be used for anything besides telling duplicates apart. There is also no guarantee that when removing all images and importing them again that every image will get the same version as before. So keep it a number and don't mess with it. :smiley:

elstoc:
I agree. Version number and a text description like I've added here should be entirely separate things.

elstoc:
@TurboGit please can you review/merge this

src/common/variables.c
```diff
@@ -260,6 +261,40 @@ static char *get_base_value(dt_variables_params_t *params, char **variable)
     result = g_strdup(params->data->camera_alias);
   else if(has_prefix(variable, "ID"))
     result = g_strdup_printf("%d", params->imgid);
+  else if(has_prefix(variable, "VERSION_NAME"))
+  {
+    GList *res = dt_metadata_get(params->imgid, "Xmp.darktable.version_name", NULL);
+    res = g_list_first(res);
```
TurboGit, line 267: Isn't this superfluous ?

elstoc:
> I read the changes correctly the duplicate module now will display the version_name instead of the title of the picture. Right

That's correct.

TurboGit:
Can you resolve conflicts? And I'll merge... Thanks.

elstoc:
There aren't any conflicts shown here.

TurboGit:
For **Rebase** I have conflicts, for **Squash and merge** no conflict, I'll use this later option anyway. Thanks.

PR #4483, opened 2020-03-16
watermark - GLib-ERROR fix

phweyland:
GLib-ERROR overflow allocating when no text is given.
When this error occurs dt terminates.

I've just avoided the calculation issue.
Alternatively we could escape the process and log an error. 
Let me know.

### DIFF
--- a/src/iop/watermark.c
+++ b/src/iop/watermark.c
@@ -930,6 +930,9 @@ void process(struct dt_iop_module_t *self, dt_dev_pixelpipe_iop_t *piece, const 
   /* get the dimension of svg */
   RsvgDimensionData dimension;
   rsvg_handle_get_dimensions(svg, &dimension);
+  // if no text is given dimensions are null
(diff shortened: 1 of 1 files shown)


phweyland:
I don't know why travis build has been cancelled. Do I have to do anything to fix this ?

TurboGit:
Code looks ok, I'm not sure how this can happen though. Is that when creating a new svg ? Anyway, a crash must be avoided so will merge.

TurboGit:
Thanks!

phweyland:
> I'm not sure how this can happen though.

Every time the watermark variable you have selected returns an empty string (in my case, a `rights` metadata not defined for an image).

 I was surprised too to see that. Should have occurred for a long time...

phweyland:
Maybe this one could go to 3.0.2 ?

TurboGit:
Agreed and merged.

Issue #4471, opened 2020-03-14
[FR] make color picker "invalid input area" switch to "default area"


johnny-bit:
**Is your feature request related to a problem? Please describe.**

This comes on legs of #3631 
Here's the scenario:

- User selects any (even miniscule) area inside image - Currently OK
- User selects area starting outside image, but ending inside image - consider starting point at image area border and end as indicated by user - currently for this OK
- User selects area starting inside image, but extends outside image area - consider starting point inside and end at image area border - currently OK
- User starts area outside image area and ends outside image area at no point crossing image area - currently this causes a very thin area on the border of image to be used

**Describe the solution you'd like**

User starts area outside image area and ends outside image area at no point crossing image area - consider this as invalid area and reset color picker area to default area

**Alternatives**

It doesn't even have to be full area drawn outside of image to be considered invalid input and reset to default area, A simple click outside image area could also be seen as invalid input and reset color picker area to default.

There are some other possiblilities - I don't know whether right-click is utilized for color pickers but if it isn't, a right click inside or outside image area could reset color picker area...

TBH - it could be both ways - invalid input resets area AND some kind of shortcut to reset color picker area.
**Additional context**

Checkout issue #3631 starting from [my comment](https://github.com/darktable-org/darktable/issues/3631#issuecomment-598881811)
This FR could also be good(er) alternative to solve #4466 
This should IMHO be considered an additional case for PR #4431 

Looking for more input on this FR

Pinging @jenshannoschwalm and @jade-nl just to notify them ;)

GrahamByrnes:
I think this is just an issue of helping the module to not encourage errors. Ideally, I'd like the cursor to be held prisoner on the image: ie once you click on the picker-arrow, further clicks are ignored unless they are on the image... or on the arrow again to cancel.

That might be a pita to implement.

I think your proposed solution is the best, otherwise. The first alternative has the disadvantage that if you're trying to make a rectangle that runs right to the edge of the image, you might require eye-surgeon skills to avoid an invalid click. Repeat it several times with a return to default and I could imagine my blood pressure rising...

jenshannoschwalm:
That's all pretty easy to implement except the "be held prisoner on the image", that would be hell to do in a portable way.

Right mouse button click as a request to fall back to default as a new feature?

The scenario you describe as a blood pressure rising condition is what i also want to avoid.

johnny-bit:
> "be held prisoner on the image", that would be hell to do in a portable way.

[...]

Issue #4466, opened 2020-03-13
[FR] Minimum area for color picker


johnny-bit:
**Is your feature request related to a problem? Please describe.**

In #3631 it's clear that issue was caused in some instances by problems fixed by #4431. However it is still possible for users to select area that makes little to no sense in terms of spot picking.

**Describe the solution you'd like**

In #4431 @jenshannoschwalm fixed the ability to pick area outside of the image. It would be great if we could introduce (or let color-picker clients decide) minumum area for color picker to work on. 

My proposition: introduce a configureable minimum sensible area dimmension to work on, eg "10x10" px makes sense because it contains a bit better info than "100x1" px area.

**Alternatives**

Alternatively a color picker consumers could just ignore spot/picked parts that are too small for them to work on or show error (eg "area too small to get meaningful data")
**Additional context**

The best example is a screenshot by @jade-nl:
![3 1 0 spot outside](https://user-images.githubusercontent.com/56407344/76611933-9665d580-651b-11ea-9eb7-cd5004e6c399.png)

as you can see this area makes very little sense to be used by white balance module.

GrahamByrnes:
Hmmm, if you can zoom in and clearly place the box on a bit of grey... at the theoretical level, that shouldn't be a problem.

Otoh, there may be a practical problem, which is the precision of the box boundary.

Once upen a time I launched an issue related to adding a frame line (ie a very thin black border), which implies setting the black border to the same width as the frame (the IT version of Henri C-B filing out his negative holders). On screen, you see that sometimes the black doesn't cover the white, but it varies with zoom. 

It turned out that when exported/printed, everything is fine, it's just a display issue.

So it may be a matter of warning the client that the box s/he sees on screen may not be the actual box when mapped back to the original image for sampling.

johnny-bit:
This isn't about zoom (well, it mostly isn't). But it could help as it would force users to zoom in order to place weird/unexpected color picker areas.

Could you look also into #4471 and say what do you think about it?

johnny-bit:
Fixed by #4495

PR #4495, opened 2020-03-18
Fixing several colorpicker issues V2 (Issues #4471 #3631)

jenshannoschwalm:
This pr replaces #4476 as a again messed up a branch.

To get an idea about this i suggest to read issues #4471 #3631
In short; it would be nice having a way to reset the colour picker area to default.
Also the pickers behaviour might be confusing if the mouse is outside the "real part" of the image.

This commit adds this functionality if the colour picker is active.

    A right-button mouse click switches the box to default size (almost all of the image)
    If you drag the mouse out of the real-image part those positions are not accepted.
    To avoid fighting with the border you may in fact use positions a few pixels outside.
    When starting to draw the box, it's initially set to a small rectangle, this
    is corrected if you drag the mouse. A simple click has more likely wanted results this way.

This sounds like what i did in #4476, the first commit for that pr further messed up the way
the color picker works. The problem was i wasn't aware of closeup must be taken care of when using
dev->pipe->output_backbuf_***

So this is not WIP any more and also a fix for darkroom issues with color picker.

### DIFF
--- a/src/views/darkroom.c
+++ b/src/views/darkroom.c
@@ -527,10 +527,13 @@ void expose(
   if(dev->gui_module && dev->gui_module->request_color_pick != DT_REQUEST_COLORPICK_OFF && dev->gui_module->enabled)
   {
     // The colorpicker bounding rectangle should only be displayed inside the visible image
-    const float hbar = (self->width - dev->pipe->output_backbuf_width) * .5f;
-    const float tbar = (self->height - dev->pipe->output_backbuf_height) * .5f;
+    const int pwidth = dev->pipe->output_backbuf_width<<closeup;
+    const int pheight = dev->pipe->output_backbuf_height<<closeup;
+
+    const float hbar = (self->width - pwidth) * .5f;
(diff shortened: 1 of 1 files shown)


johnny-bit:
I've checked thoroughly and this is absolutely awesome!

This fixes #4466 becasue
> When starting to draw the box, it's initially set to a small rectangle [...] A simple click has more likely wanted results this way.

I love the right-click to cancel current area and reset to default. I was missing that couple times and it's great addition imo (NEEDS MENTION IN MANUAL! and/or in _shortcuts_ window ;)

This also fixes #4471 because right-click for default and ignoring input outside image area is great way to deal with invalid input!

I played with it and haven't managed to break it or get invalid interaction with modules using color-pickers.

I think it's ready :) also pinging @jade-nl

TurboGit:
Works for me ! Thanks.

jade-nl:
Perfect!

@jenshannoschwalm: Thanks for the time/effort!

jenshannoschwalm:
@jade-nl welcome!

PR #4499, opened 2020-03-19
metadata update

phweyland:
- add internal, optional and user metadata types
- add tooltips and fix some incoherences

Internal metadata allows dev to use metadata mechanisms (read/write <-> db & xmp) for internal information without interfering with user's metadata (today metadata). They are not visible to the user.

Optional metadata are per default hidden, but can be set visible by the user if he finds them useful.

Internal and hidden metadata are not visible in metadata editor, image information and import modules. They are not exported.

An example of this is the PR #4477 `version name` which belongs to duplicate module.
As it represents the duplicate name field, it should be internal. 
One can argue he wants to use it outside duplicate module (for images without duplicate). In this case it should be optional. To be decided.

NB: even if metadata `version name` is set as internal, and as such not exported with images, the corresponding variable `$(VERSION_NAME)` still allows the user to export the value.

### DIFF
--- a/src/libs/metadata.c
+++ b/src/libs/metadata.c
@@ -66,6 +66,25 @@ uint32_t container(dt_lib_module_t *self)
   return DT_UI_CONTAINER_PANEL_RIGHT_CENTER;
 }
 
+void init(dt_lib_module_t *self)
+{
+  for(unsigned int i = 0; i < DT_METADATA_NUMBER; i++)
+  {
+    const int type = dt_metadata_get_type(i);
+    if(type == DT_METADATA_TYPE_OPTIONAL)
+    {
+      const char *name = (gchar *)dt_metadata_get_name(i);
+      char *setting = dt_util_dstrcat(NULL, "plugins/lighttable/metadata/%s_hidden", name);
+      if(!dt_conf_key_exists(setting))
+      {
+        // per default this one should be hidden
+        dt_conf_set_bool(setting, TRUE);
+      }
+      g_free(setting);
+    }
+  }
+}
+
 static gboolean _is_leave_unchanged(const char *text)
 {
   return g_strcmp0(text, _("<leave unchanged>")) == 0;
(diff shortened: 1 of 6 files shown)


src/libs/metadata.c
```diff
@@ -408,34 +428,44 @@ static void config_button_clicked(GtkButton *button, dt_lib_module_t *self)
   gtk_grid_attach(GTK_GRID(grid), label, 0, 0, 1, 1);
   label = gtk_label_new(_("hidden"));
   gtk_grid_attach(GTK_GRID(grid), label, 1, 0, 1, 1);
+  gtk_widget_set_tooltip_text(GTK_WIDGET(label),
+            _("tick if the corresponding metadata is not of interest for you"
```
TurboGit: I'm no native English either, but I propose:

tick if the corresponding metadata is of no interest for you

src/libs/metadata.c
```diff
@@ -408,34 +428,44 @@ static void config_button_clicked(GtkButton *button, dt_lib_module_t *self)
...
+              "\nit will be hidden from metadata editor, image information and import modules"
```
TurboGit: it will be hidden from metadata editor, image information and import module

src/libs/metadata.c
```diff
@@ -408,34 +428,44 @@ static void config_button_clicked(GtkButton *button, dt_lib_module_t *self)
...
+              "\nit will not be exported neither"));
```
TurboGit: neither will it be exported

src/libs/metadata.c
phweyland: OK

src/libs/metadata.c
phweyland: the s was for the 3 modules ...

src/libs/metadata.c
phweyland: OK

src/libs/metadata.c
phweyland: I'll do it

phweyland:
Changes made as requested. Thanks

Issue #4487, opened 2020-03-17
Glitches about lighttable rewrite: list them all here !


Nilvus:
Thanks to the awesome and big work of @AlicVB merged this day, we now have a far better and faster lighttable. This change also some thing from how users see them (a little). I will work this week on CSS part to improve it.

So, I suggest to all of you add posts here to precise UI glitches about this part here.

Please be the more precise as possible: which part (if screen capture, not a too big one but just the part to fix) you see. Precise also with which of the actual four themes you see that. I will begin (I don't check things a lot by now) with the first issue I will fix in the second post.

For what I've seen by now, I think issues will not be a lot !

@TurboGit or @aurelienpierre or @AlicVB. One of you could assign this issue to me and necessary labels? Thanks.

Nilvus:
First issue:

On darktable-grey theme: text over thumbnails is not really readable (not to adjust color).

AlicVB:
More than that : as said (or not, don't remember) I've absolutely not done anything for theme other that the base one.
The idea was to first have a clean base theme, which will be easier to tweak after.

Again, thanks for taking care of this part, and ping me here or directly by mail if you need anything !

Nilvus:
> More than that : as said (or not, don't remember) I've absolutely not done anything for theme other that the base one.
> The idea was to first have a clean base theme, which will be easier to tweak after.

I confirm that you precise that on your PR.

AxelG-DE:
I donno this is a CSS clitch, actually I don't think so, but this is suddenly the max view I can get in lighttable view and I must say I do not like it so much (it was covering the full area before, w/o that DIA-Frame)....

![Screenshot_20200319_205402](https://user-images.githubusercontent.com/49004103/77109208-ce8d7c80-6a23-11ea-9f12-fe66dc13408c.png)

AlicVB:
That's intended. The idea is to keep a coherent design between mode. You are still in filemanager mode, and it's how image are display here.
Better use (sticky)full preview if you want the image with the framing background. Note that now, you can show all panels in full preview mode if you need them.

(and yes, not configurable via css)

AxelG-DE:
I am medium happy as I used this view in the past regularly. But I don't wanna be against, as we gonna see, what new things might bring good in the future and I will get used to alt+W :-)

Nilvus:
[...]

PR #4505, opened 2020-03-20
Update CSS on @AlicVB lighttable rewrite

Nilvus:
Update of the themes since lighttable rewrite by @AlicVB.

I have fix text over thumbnail and make stars and icons (group, altered...) consistent between them and between themes. I have also just move some lines to just order them (lighttable ones near them, easier to update files). See #4487 

Before merging, please check all things I've change and especially minor updates this work made on bauhaus parts (I've checked many modules on darkroom and all is good but I could have forgot something).

Of course, review by @aurelienpierre and @AlicVB as they had work respectively on GUI and lighttable are especially requested. All others reviews are of course welcomed.

### DIFF
--- a/data/themes/darktable-elegant-dark.css
+++ b/data/themes/darktable-elegant-dark.css
@@ -41,13 +41,14 @@
 @define-color plugin_label_color @grey_75;
 
 /* Modules controls (sliders and comboboxes) */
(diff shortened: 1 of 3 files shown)


TurboGit:
Merging then! Thanks.

PR #4509, opened 2020-03-21
Fix invalid CMake syntax

mhaehnel:
The syntax of the foreach statement in data/kernels/CMakeLists.txt
was invalid. This lead to errors when running cmake.

Use correct syntax to make the build work.


TurboGit:
Thanks!

PR #4510, opened 2020-03-21
Respect scrollbar option while in filestrip mode

jenshannoschwalm:
In `src/dtgtk/thumbtable.c` `_thumbtable_restore_scrollbars` there seem to be a wrong
checking for the configure option "scrollbars" depending on table->mode.

While in `DT_THUMBTABLE_MODE_ZOOM` mode the scrollbars should be off, in `DT_THUMBTABLE_MODE_FILMSTRIP` mode we should check for the "also darkroom" option.

### DIFF
--- a/src/dtgtk/thumbtable.c
+++ b/src/dtgtk/thumbtable.c
@@ -846,15 +846,21 @@ static gboolean _event_button_release(GtkWidget *widget, GdkEventButton *event, 
 static void _thumbtable_restore_scrollbars(dt_thumbtable_t *table)
 {
   table->scrollbars = FALSE;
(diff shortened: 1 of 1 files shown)


jenshannoschwalm:
@AlicVB your opinion?

TurboGit:
Can you describe the way to reproduce the wrong behavior ? Thanks.

jenshannoschwalm:
Pretty easy. Of course enable the scrollbars including darkroom.
When opening the image you can see the bars, now switch on filmstrip and the bars are gone. (If you start with filmstrip mode=on you will never see the bars.)

TurboGit:
I see I never use scrollbars in darkroom, that's why I did not spot the issue.

PR #4511, opened 2020-03-21
bugfix: ensure correct handling of darkroom/ui/single_module

### LINKED ISSUE #3584: UI → Module settings auto collapse
Steps to reproduce:
1. Make sure that "expand a single darkroom module at a time" is enabled in dt preferences.
1. Open (just as example) 'shadows and highlights' module settings.
2. Click 'new instance' to create new 'exposure' instance.

Expected result: 'shadows and highlights' module collapsed
Actual result: both modules are open

elstoc:
option to ensure that only one darkroom module is expanded at a time is now
correctly observed in the following scenarios:

 - when creating a new module instance
 - when opening a new image from the lighttable

Resolves #3584


TurboGit:
Thanks.

PR #4513, opened 2020-03-21
Clear bg also in lighttables expose_empty note.

jenshannoschwalm:
If the is an empty darktable.collection in `lighttable.c expose` we should
fill the whole area with the proper bg color before painting the visual info.

This problem was mentioned in #4487
@AlicVB agreed?


Nilvus:
Hi @jenshannoschwalm. I suppose it's about the refreshing problem I mentioned and you confirm with you screen capture about text glitch issue also? So, thanks for working on that.

PR #4517, opened 2020-03-21
bugfix: set/lose focus when expanding/collapsing modules

### LINKED ISSUE #2776: crop and rotate not activated properly
**Describe the bug**

Crop and rotate does not start in editing mode when switching between
image modes in darkroom.  Clicking ⏻ may not activate it.

**To Reproduce**

In "⚙/shortcuts/views/darkroom", make sure `Page Up` is bound to
"image back", and `Page Down` is bound to "image forward".

Open an image in darktable, which is **not** the first one in the
collection.  Make sure its "crop and rotate" module is disabled.

Hit `Page Up` to go to the previous image.  Open the "crop and rotate"
module, and change the crop by dragging the borders to reduce the
field of view.

Hit `Page Down` to go to the next image.  Good: The "crop and rotate"
module is expanded, but inactive.  Click its ⏻ button to activate it.
Bad: There will be no crop frame on the image in the middle that could
be dragged around.  Expected: A crop frame on the image in the middle
that could be dragged around.

Click its ⏻ button two more times.  The expected crop frame will show
up and allow to change the crop.

Hit `Page Up` to go to the previous image again.  Good: The "crop and
rotate" module is expanded and active.  Bad: The image is shown in
cropped state.  Expected: The full image with a crop frame that could
be dragged around.

Click its ⏻ button two more times.  The expected crop frame will show
up and allow to change the crop.

**Expected behavior**

When switchin images in darkroom with "crop and rotate" active, the next image should present "crop and rotate" in editing mode it it's active, or should activate it by pressing ⏻ if it's not active.

**Additional context**

Maybe this is related to #2618.

  * current ArchLinux
  * no desktop environment (plain X11 and FVWM)

    $ git describe
    release-2.7.0-1516-g319ee172c

elstoc:
Ensure focus is set correctly when expanding/collapsing modules using keyboard shortcut.

Resolves issues around crop/rotate guides remaining on screen when module collapsed and not reappearing when expanded.

### DIFF
--- a/src/develop/imageop.c
+++ b/src/develop/imageop.c
@@ -2127,8 +2127,11 @@ static gboolean show_module_callback(GtkAccelGroup *accel_group, GObject *accele
     dt_dev_modulegroups_switch(darktable.develop, module);
   }
 
-  dt_iop_gui_set_expanded(module, TRUE, dt_conf_get_bool("darkroom/ui/single_module"));
-  dt_iop_request_focus(module);
+  dt_iop_gui_set_expanded(module, !module->expanded, dt_conf_get_bool("darkroom/ui/single_module"));
+  if(module->expanded)
+  {
+    dt_iop_request_focus(module);
+  }
   return TRUE;
 }
 
--- a/src/views/darkroom.c
+++ b/src/views/darkroom.c
@@ -759,6 +759,9 @@ static void dt_dev_change_image(dt_develop_t *dev, const uint32_t imgid)
     return;
   }
 
+  // get last active plugin, make sure focus out is called:
+  gchar *active_plugin = dt_conf_get_string("plugins/darkroom/active");
+  dt_iop_request_focus(NULL);
   // store last active group
   dt_conf_set_int("plugins/darkroom/groups", dt_dev_modulegroups_get(dev));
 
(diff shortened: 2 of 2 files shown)


elstoc:
This was intentional (though maybe not a good idea - the main issue I was trying to fix was the one I added to the comments). Crop & Rotate guides don't currently depend on the module being active. They have always (or as far back as I can remember) displayed when you _expanded_ the module. I interpreted the original bug as requesting this functionality but perhaps I misread.

Maybe it's better that the guides only appear when the module is both enabled and expanded? But that's not how it works now.

johnny-bit:
After shortcuts change I can operate on crop&rotate via shortcuts, so I don't expand it and having guides for it is bonus (that issue whit keys for comboboxes was a godsend!)

So... however you plan on doing that it's problematic :)

elstoc:
The problem is that when you go to another image with crop&rotate still expanded the only way to be able to use the module again is to click the header twice, since the functionality relies on the module having _focus_, and that's the only way to give it focus. It doesn't matter if the module is both active and expanded, because if it doesn't have focus it won't work.  My change gives it focus if it had focus in the last image.

elstoc:
@TurboGit I've removed the code that was causing the regression. I do still think that there is an issue around Crop&Rotate - the fact that after you've double-clicked to commit your change you have to expand/collapse to get the guides back again - but I can't think of a nice solution that won't be bad for other modules. The second fix should be uncontroversial so let's leave it at that for now.

TurboGit:
Ok, works for me. Thanks.

PR #4520, opened 2020-03-21
fix some glitchs after filemananger rewrite

AlicVB:
fix bunch of issues reported recently for the lighttable rewrite. (Most of them are trivial changes)
I'm expecting some more glitch like this in the future. Don't hesitate to report them in separate issues.
Thanks to the testers


TurboGit:
Works for me. Thanks !

PR #4523, opened 2020-03-21
fix regression with importing single image

### LINKED ISSUE #4516: Single image import without xmp and empty collection fails to open properly
**Describe the bug**

Upon opening fresh darktable (no configs, empty db, no images open previously) and trying to import single image (that should get then opened in darkroom) a message is shown: "error loading file `filename`". After returning to lighttable, selecting collection and then doubleclicking on image it opens properly.

**To Reproduce**

1. have empty everything and make sure the pic you're about to export doesn't have xmp (no xmp file seems important)
2. Go to lighttable -> import
3. Click on `image...`
4. select image to import and click `open`
5. See error "error loading file `filename`"
6. see that it isn't opened in darkroom

**Expected behavior**

Upon importing single file the file should be opened in darkroom. Upon returning to lighttable it should be present there.

**Screenshots**

empty lightroom:
Zrzut ekranu z 2020-03-21 17-48-41

error right after importing single image:
Zrzut ekranu z 2020-03-21 17-48-51

however imported image IS present:
Zrzut ekranu z 2020-03-21 17-49-47

error present even in non-empty collection when there's no xmp for single file:
Zrzut ekranu z 2020-03-21 17-58-35

 - Darktable Version: darktable 3.1.0+999~gc250dda10
 - OS: Linux
 - OpenCL: NO

**Additional context**

doesn't reproduce before lighttable rewrite merge

*BOTH*

 - Are the steps above reproduce with a fresh edit (removing history)?
The error happens _only_ when there's no XMP for the file

YES!
```sh
./build.sh --disable-opencl --disable-game --install --build-type Debug
```
 - Is the issue still present using an empty/new config-dir

YES!

** `darktable -d all` output **
This is present in logs just after clicking `open` in import dialog:
```
11,942244 [sql] /home/johnny/darktable_repos/upstream/darktable/src/libs/collect.c:1572, function list_view(): prepare "SELECT folder, film_rolls_id, COUNT(*) AS count FROM main.images AS mi JOIN (SELECT id AS film_rolls_id, folder       FROM main.film_rolls)   ON film_id = film_rolls_id  WHERE (1=1) GROUP BY folder ORDER BY film_rolls_id DESC"
11,942835 [sql] /home/johnny/darktable_repos/upstream/darktable/src/libs/tools/timeline.c:624, function _time_read_bounds_from_collection(): prepare "SELECT db.datetime_taken FROM main.images AS db, memory.collected_images AS col WHERE db.id=col.imgid AND LENGTH(db.datetime_taken) = 19 AND db.datetime_taken > '0001:01:01 00:00:00' ORDER BY db.datetime_taken ASC LIMIT 1"
11,942961 [sql] /home/johnny/darktable_repos/upstream/darktable/src/libs/tools/timeline.c:645, function _time_read_bounds_from_collection(): prepare "SELECT db.datetime_taken FROM main.images AS db, memory.collected_images AS col WHERE db.id=col.imgid AND LENGTH(db.datetime_taken) = 19 AND db.datetime_taken > '0001:01:01 00:00:00' ORDER BY db.datetime_taken DESC LIMIT 1"
11,943186 [sql] /home/johnny/darktable_repos/upstream/darktable/src/libs/duplicate.c:301, function _lib_duplicate_init_callback(): prepare "SELECT i.version, i.id, m.value FROM images AS i LEFT JOIN meta_data AS m ON m.id = i.id AND m.key = ?3 WHERE film_id = ?1 AND filename = ?2 ORDER BY i.version"
12,079664 [sql] /home/johnny/darktable_repos/upstream/darktable/src/libs/tools/timeline.c:742, function _block_get_at_zoom(): prepare "SELECT db.datetime_taken, col.imgid FROM main.images AS db LEFT JOIN memory.collected_images AS col ON db.id=col.imgid WHERE LENGTH(db.datetime_taken) = 19 AND db.datetime_taken > '2018:01:01 00:00:00' ORDER BY db.datetime_taken ASC"
12,086154 [lighttable] expose took 0,0000 sec
12,146942 [lighttable] expose took 0,0000 sec
15,136823 [lighttable] expose took 0,0000 sec
15,184980 [lighttable] expose took 0,0000 sec
16,944527 [lighttable] expose took 0,0000 sec
```

elstoc:
when importing a single image without an existing xmp file, darkroom was
not being opened and instead an error was being returned.

Resolves #4516

### DIFF
--- a/src/common/image.c
+++ b/src/common/image.c
@@ -1325,22 +1325,6 @@ static uint32_t dt_image_import_internal(const int32_t film_id, const char *file
     DT_DEBUG_SQLITE3_BIND_TEXT(stmt, 2, imgfname, -1, SQLITE_STATIC);
     sqlite3_step(stmt);
     sqlite3_finalize(stmt);
-
-    //get the new minimum version (may be non-zero)
-    // returned to calling function
-    DT_DEBUG_SQLITE3_PREPARE_V2
-      (dt_database_get(darktable.db),
-       "SELECT id FROM main.images i1 WHERE film_id = ?1 AND filename = ?2 AND "
-          "version = (SELECT MIN(version) FROM main.images i2 WHERE i2.film_id = i1.film_id "
-          "AND i2.filename = i1.filename)", -1, &stmt, NULL);
-    DT_DEBUG_SQLITE3_BIND_INT(stmt, 1, film_id);
-    DT_DEBUG_SQLITE3_BIND_TEXT(stmt, 2, imgfname, -1, SQLITE_STATIC);
-    if(sqlite3_step(stmt) == SQLITE_ROW)
-    {
-      id = sqlite3_column_int(stmt, 0);
-    }
-    sqlite3_finalize(stmt);
-  
   }
   else
   {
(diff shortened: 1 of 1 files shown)


elstoc:
Please note that while this fixes issue 4516, it appears that another fault has been introduced. I'm fairly sure this fault didn't exist in my original testing (though don't quote me on that!) and may be caused by interaction with @AlicVB's lighttable rewrite. I can't pin down the cause.

Essentially what the image import does (with my fix) is it creates a dummy version of the image (with version -1), using that as a basis for loading xmp files (to duplicate from). If it loaded xmp files, that dummy record is no longer required and is deleted (else its version is changed to 0).

If you import a folder, the last of these dummy images seems to remain in the lighttable (being loaded from some cache somewhere?):

Screenshot_2020-03-21_21-46-32

If you go into the darkroom and back to the lighttable it's gone. Similarly if you load a single image it's not an issue, because that loads immediately into the darkroom.

GrahamByrnes:
Works for me. There are still the weird selection issues, but they are present with already loaded images, so I don't think Elstoc should be blaming himself tor them :-)

TurboGit:
Thanks!

PR #4526, opened 2020-03-22
more top/bottom panels accels

### LINKED ISSUE #4484: Add an option to toggle bottom and top bar without the timeline (and so on)
Since version 3.0.1, when we toggle the bottom or top "bar", we have to circle thought 3 modes (off/small bar/big bar) to activate or deactivate it.
Personally this makes me very frustrated and less productive, as I have to click 3 times instead of two, as I don't use (very often) the bottom image list/timeline, and as it changes a lot the size of the picture and as a result my perception of the image. The same goes for the top bar, I only use the small one when I change the settings, the rest of the time I have to click twice to get back to the lightroom for instance, and that's annoying and frustrating.

(I'm sorry I don't know the name of those "bars"/"ribbons"/"panels", so it's hard to describe)

Add an option to change this behaviour, for those who would like to revert it to the "on/off" behaviour (I suppose the current solution would be the default).

Thank you

AlicVB:
this fix #4484 

- this refactor header/filmstrip/timeline accels, so they are all handled globally in the same place.
- this add accels (with no default keys) for top toolbar, bottom toolbar, all top panels, all bottom panels

### DIFF
--- a/src/control/control.c
+++ b/src/control/control.c
@@ -663,13 +663,6 @@ int dt_control_key_pressed_override(guint key, guint state)
     gtk_widget_queue_draw(dt_ui_center(darktable.gui->ui));
     return 1;
   }
-  else if(key == accels->global_header.accel_key && state == accels->global_header.accel_mods)
-  {
-    /* toggle panel viewstate */
-    dt_ui_toggle_header(darktable.gui->ui);
-    gtk_widget_queue_draw(dt_ui_center(darktable.gui->ui));
-    return 1;
-  }
   // add an option to allow skip mouse events while editing masks
   else if(key == accels->darkroom_skip_mouse_events.accel_key && state == accels->darkroom_skip_mouse_events.accel_mods)
   {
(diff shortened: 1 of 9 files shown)


Nilvus:
So if understanding your PR, we have to go to preferences and sets our shortcuts? Maybe it would be better to have some default ones like the Ctrl+h for header, etc.

AlicVB:
well... I've keep ctrl-h, ctrl-f, but for all the new ones, it's so dependent of the user needs and workflow, that I'm reluctant to add some and reduce the already small list of available keys... And we have to keep in mind that this sort of accels are mainly used by "advanced users" (means : those who have already enough experience with dt to know exactly their workflow needs) which are able to define them.

That said, if other absolutely want default shortcuts and can come with a consistent list of keys, why not...

TurboGit:
Works for me. Thanks.

TurboGit:
I have merges as-is, if we want to have default accel this can be done in another PR. Sincerely I have no strong opinion on this.

Issue #4286, opened 2020-02-13
darkroom view: shortcut Alt+2 can produce a constant flickering / jittering


jade-nl:
**Describe the bug**

In darkroom view: In certain cases using the Alt+2 shortcut to zoom fill an image causes the image area to constantly jitter/flicker (_seizure_ if you want). It can be stopped by using Alt+1 or Alt+3.

**To Reproduce**

0. start darktable development version (easier to recreate conditions that lead to this behaviour)
1. preferences -> GUI options -> *show scrollbars for center view* should be set to *lighttable + darkroom*
2. open a roughly 4:3 image in darkroom view that is larger than the available image area
3. enable the filmstrip
4. adjust the height of the filmstrip and/or the side panels to make the image fit nicely. As in: borders around the image are as small as possible and evenly thick on all four sides (see screenshot below)
5. press Alt+2 (Alt+1 or Alt+3 to stop the jitter)

**Expected behavior**

Alt+2 to zoom fill the image as advertised.

**Screenshots**
This shows the layout and borders needed as mentioned in reproducibility step 4:
![Screenshot_2020-02-13_15-40-18](https://user-images.githubusercontent.com/56407344/74445936-7806ae80-4e77-11ea-857e-cec6fcbffc88.png)

**Platform:**

* darktable Version:
&nbsp;&nbsp;&nbsp;Stable 3.0.0
&nbsp;&nbsp;&nbsp;Development 3.1.0+639
* OS     : Linux - kernel 4.19.0-8-amd64
* Distro : Debian Buster (rel. 10.3)
* Graphics card   : NVIDIA GF104 [GeForce GTX 460] (rev a1)
* Graphics driver : NVIDIA x86_64 Kernel Module 390.116
* OpenCL activated : No
* Xorg : 7.7+19
* GTK+ : 3.24.5-1

After confirming the issue I filed this report on behalf of someone else, that user's specs are:

* darktable version : development 3.1.0-584
* OS : Linux 4.15.0-76
* Distro : Ubuntu 18.04.3 LTS
* Graphics card : Nvidia GeForce RTX 2060
* Graphics driver : Nvidia Driver 430.50
* OpenCL activated : Yes
* Using X11
* GTK+ : 3.22.30-1ubuntu4

**Additional context**

 - I filed this on behalf of someone else who mentioned it on pixls.us: [This thread](https://discuss.pixls.us/t/bug-darkroom-shortcut-alt-2/16392)

[...]

PR #4530, opened 2020-03-22
Fixing darkroom seizures, fixing #4286

jenshannoschwalm:
The reporting issue was #4286 although the report is somewhat misleading.

To trigger the issue you must be in darkroom mode with sliders enabled there.
If you carefully resize the window in a way that **both** height and width sliders are close
to full you can observe a resizing loop.

This is triggered by expose in darkroom, it checks for possibly adding or removing sliders
which results in gtk resizing which leads to recalculation of scales and thus again expose
and resizing.

This fix adds a threshhold for adding a slider thus avoiding the above.


TurboGit:
Works for me. Thanks.

PR #4532, opened 2020-03-22
Color picker and masks must respect PPD. Fixes #4531

### LINKED ISSUE #4531: color picker not resizable
In dt with 64d5ee2d1b282aeee5adc4764adc20b847038029, the area color picker size cannot be changed/edited and stays as the default (whole image covered). Affects global picker, white balance and modules pickers (filmic, color balance, negadoctor). Reverting that commit solves the issue.

I'm on HiDPI screen with screen PPD = 2 and DPI set to default.

jenshannoschwalm:
The cairo clipping to prevent masks and color picker boxes to draw on the surroung borders **and** the mouse position checks must respect the gui->ppd.

### DIFF
--- a/src/views/darkroom.c
+++ b/src/views/darkroom.c
@@ -464,8 +464,8 @@ void expose(
 
     cairo_save(cri);
     // The colorpicker samples bounding rectangle should only be displayed inside the visible image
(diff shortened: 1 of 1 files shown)


TurboGit:
Works for me, but cannot test on HDPI screen.

PR #4534, opened 2020-03-22
Migrate clip parameter of basicadj

hlecleme:
The vibrance parameter has been added before clip in the v2 parameters struct of basicadj. If I understood correctly how the migration works, the legacy_params function should then take the vibrance value of v1 as clip parameter of v2. Is this correct?

Should it be better to rewrite the legacy_params function to define the v1 struct as dt_iop_basicadj_params_v1_t (as in other modules, e.g. src/iop/borders.c)?

### DIFF
--- a/src/iop/basicadj.c
+++ b/src/iop/basicadj.c
@@ -107,7 +107,8 @@ int legacy_params(dt_iop_module_t *self, const void *const old_params, const int
     new->middle_grey = old->middle_grey;
     new->brightness = old->brightness;
     new->saturation = old->saturation;
-    new->clip = old->clip;
+    // version 2 vibrance field added before clip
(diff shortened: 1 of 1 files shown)


TurboGit:
This looks very wrong to me.

hlecleme:
Example XMP file created with Darktable 3.0.0. The clip parameter of basicadj was set to 0.35.

[basicadj.xmp.gz](https://github.com/darktable-org/darktable/files/4365664/basicadj.xmp.gz)

TurboGit:
I have just pushed a proper fix for this. Can you double check it works well now?

See 48b7e80ca.

hlecleme:
Yes, it works for me.
Thanks.

PR #4536, opened 2020-03-22
Folder levels code redundancy

GrahamByrnes:
While looking for something else I came across these rather surprising bits of redundant code:
CLAMPS followed by an if;
No bound on data entry but value bounded by the CLAMPS

They seem so obvious I wonder if there is a reason, but dt seems to be happy with the changes.


TurboGit:
Looks good to me. Thanks.

GrahamByrnes:
Cool, thanks.

PR #4335, opened 2020-02-20
Update wb_presets.c (Nikon Coolpix P1000)

Macchiato17:
Adding the whitebalance values for Nikon Coolpix P1000, firmware 1.2.

I extracted the values with the help of your Ruby script extract_wb on Windows 10 professional x64 as described on https://www.bilddateien.de/blog/2019-01-20-kamera-weissabgleichseinstellungen-fuer-darktable-ermitteln.html. 

When running the script, there was an error about something like "array not being an array" in line 164 (found_presets = Hash[found_presets.sort.zip].keys), so I commented this line and manually did the sorting and removal of duplicates in Notepad++. After that, there were some occurrences of value-duplicates (in the curly braces) with different WB-tags assigned, which I also removed from the list. If you need the original (unsorted) list, please feel free to send me an e-mail :-)

I'm wondering, why the whitebalancing shifting-values (e.g. -3...0...+3) which I applied where possible, is not found in the resulting list anymore.

PR #4548, opened 2020-03-25
fix Nikon P1000 wb presets

johnny-bit:
resort wb presets with finetuning according to coefficients and add proper finetuning options for them.

this fixes wb preset bug introduced in #4335


TurboGit:
Sounds ok ! Thanks.

PR #4549, opened 2020-03-25
filmstrip : disable buttons by default

### LINKED ISSUE #4545: Introduce toggle to turn off stars rating in filmstrip preview (darkroom)
I'm always frustrated when i click on a image in filmstrip and suddenly the image is gone.The reason is that i accidentally hit the star rating. for example 2 stars which is beyond my stars filter of 4 stars and the image disappears. Mode is Darkroom. 

add a toggle in configuration where you can disable the rating by mouseclick. One could still rate the image by keyboard!

**Alternatives**
a toggle in cofiguration where you can switch off star preview in darkroom completely. 

**Additional context**
it is really really annoying when you recognize later that an image which you can still remember, is gone and then when you look it up in lighttable it has suddenly 0 stars. one can accidently delete great images with this behaviour... 

greetings from Austria, 
Dave

AlicVB:
this fix #4545 
this can still be changed by hidden "plugins/filmstrip/readonly_buttons" pref

### DIFF
--- a/data/darktableconfig.xml.in
+++ b/data/darktableconfig.xml.in
@@ -648,6 +648,13 @@
     <shortdescription>enable extended thumb overlay</shortdescription>
     <longdescription>if set to true, thumb overlay shows filename and some exif data.</longdescription>
   </dtconfig>
+  <dtconfig>
+    <name>plugins/filmstrip/readonly_buttons</name>
+    <type>bool</type>
+    <default>true</default>
+    <shortdescription>disable buttons (stars, ...) in filmstrip thumbnail</shortdescription>
+    <longdescription>if set to true, thumbnail buttons will not be clickable</longdescription>
+  </dtconfig>
   <dtconfig prefs="gui" section="lighttable">
     <name>plugins/lighttable/thumbnail_tooltip_pattern</name>
     <type>string</type>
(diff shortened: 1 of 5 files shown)


TurboGit:
I would use the opposite default as it has always been. I'm using this often as after adding stars to some pictures to edit when in darkroom I want to remove stars to some of the first selected which actually are not that good. In the darkroom I also often add more stars to edits I love and makes the picture really great. Having to go back to lighttable for that would be really a lost of time.

AlicVB:
Fair enough, no problem :)
What I can propose to please everyone : make the pref visible in the dialog and set its default to false (to get the previous behavior by default)
It will clutter a little more the pref dialog, but well... otherwise I'm not sure it's worth the pain to add complexity in the code for a feature not really discoverable !

Note that for my personal use, I don't care : I'm used to shortcuts, and anyway I rarely show filmstrip :P

TurboGit:
Personally I would vote to do nothing. Rating from filmstrip has always be supported since very early dt inception... I see nothing new to change that.

johnny-bit:
This might help "butterfingers" (dunno how to call those) whose fingers slip etc ;) So this is imo can be usability switch: have filmstrip fully functional via mouse (default true) or readonly.

I think it would be 90-10% spread between those with fully functional vs readonly.

johnny-bit:
I think more users are looking for it ;) At least have it as an option :)

bschwehn:
Sorry for hijacking that other thread, I didn't see this PR. Personally, I think this may be worth a visible option even. Either way, I would definitely use this flag.

PR #4550, opened 2020-03-25
zoomable : ensure we always see some image on screen

### LINKED ISSUE #4529: Zoomable lighttable misses the plot...
Selection in clone & file manager seems to be much better now!

However, there is maybe a new issue:
If i select the last image in a short-ish film role then swith to zoom, it appears partly off the side of the viewing area. Zooming out makes it disappear completely, as though the zoom centre was placed one screen to right of the one I'm looking at.

If I select the first image, it seems to work ok (I guess, I haven't much used zoom view in the past).

dt3.1.0 1014, Windows 10

AlicVB:
may fix #4529 

This fix 2 bugs for zoomable lighttable : 
- when you zoom in, ensure you always keep some part of images on screen, otherwise you may have no way to find them back, apart from exiting/reentering
- when panning, fix the jump if mouse go out of the center area

### DIFF
--- a/src/dtgtk/thumbtable.c
+++ b/src/dtgtk/thumbtable.c
@@ -769,8 +769,6 @@ static gboolean _event_leave_notify(GtkWidget *widget, GdkEventCrossing *event, 
   if(event->detail == GDK_NOTIFY_INFERIOR) return FALSE;
 
   table->mouse_inside = FALSE;
-  table->last_x = -1;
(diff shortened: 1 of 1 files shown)


GrahamByrnes:
Yes, I think this and @TurboGit's earlier changes may have fixed it. It now seems very usable at least :-)

TurboGit:
Works for me, thanks !

PR #4553, opened 2020-03-25
Recent to distant sort

GrahamByrnes:


GrahamByrnes:
If you have photos folders containing the date, you have possibly been annoyed by the fact that the "collect" widget puts your most recent work way down the bottom of the list. The same is true for filmrolls, times and dates.

This PR attempts to fix that. It uses a single additional entry in Settings to reverse the sort order on these time-related file trees. The serious changes are in collect.c

It also removes the case sensitivity for tags: at present the sort order is zebra < Aardvark < Cat and this is most noticeable in tags (if like me you are a bit sloppy about typing tags). So tags are now sorted in a case-insensitive order: Aardvark < Cat < zebra.

This was always the case for folders, btw... I think various things were added into collect.c without understanding how the tree ordering worked.

I'll note that I changed the initial sort so that time and date are extracted in either ascending or descending order... this seems actually to make no difference, but maybe extracting them DESC will add some trivial amount of efficiency vs extracting in ASC and then re-sorting.

GrahamByrnes:
Strange, I've just updated everything, downloaded, built and run without problem... but Travis seems to be stuck.

TurboGit:
Please please, not merging of master in topic branch.

Can you rebase on origin master and force push your changes here? Thanks.

GrahamByrnes:
Not sure what happened... there was a branch, but then there were conflicts with @phweyland 's change, and somehow a bunch of code disappeared.
Am trying to fix.

phweyland:
> phweyland 's change, and somehow a bunch of code disappeared

Ooops, I hope I haven't made too many mistakes ... the pieces of code I've replaced are related to metadata.

GrahamByrnes:
No, none at all that I can see. There was just a line where I used a
gtk_combo which needed to be changed to a _combo
I made the change but suspect I screwed up saving... there was a whole
chunk of your code that went missing a few hundred lines down. Hence Travis
was unhappy.

Time for some more git-learning (without covid this time!)

GrahamByrnes:
Replaced by clean copy in #4561. @TurboGit I'm sorry, I'll get better at this :-(

Issue #4555, opened 2020-03-26
lighttable and zoom support


TurboGit:
A question for @AlicVB.

There is something strange in the way the zoom level of the lighttable is supported, at least it looks like we want to clean-up this.
- The slider for the zoom (libs/tools/lighttable.c) is set in range 1 .. 21
- A piece of code in the same file check for 1 .. 30
`d->current_zoom = MAX(1, MIN(30, dt_collection_get_selected_count(darktable.collection)));`
- the zoom cannot be changed with the mouse past 13 as there is a hard limit (const in view/lighttable.c and tools/lighttable.c):
`#define DT_LIBRARY_MAX_ZOOM 13`

We probably want to uniformize this, maybe 21 or 25 or 31....

@AlicVB : I'd like your input on this and I'm fine doing the actual work.

AlicVB:
oops, you're right, it's a missing part of my recent rework.
I think we have to define a "max_zoom" constant somewhere in the code (in an unique place) and use that. For the number, I think 20...25 is good. More is really unreadable, at least on my screen !

If you have some time to do this, that would be great. Otherwise I'll have a look when I find some time...

TurboGit:
See #4557, there was a third definition in thumbtable.c, should be easier to maintain now.

I have settled for 25 as max zoom.

PR #4557, opened 2020-03-26
Ensure a single definition is controling the max zoom level.

TurboGit:
This fixes issues where not the same value was used everywhere. It
will also ease the maintenance.

For #4555.


TurboGit:
> Great ! all good for me. Thanks for taking care of that !

A real pleasure, I'm reviewing so many PR that I don't have lot of time to actually code !

PR #4559, opened 2020-03-26
metadata - import configuration

phweyland:
EDIT - currently hidden metadata are not imported (in dt_metadata_set()) if hidden (in metadata editor). The idea is to control the metadata importation at import time and to make it independent of the hidden flag).

Opening this PR at early stage because that will have some impact on general metadata update.

Before modifying exif.cc I've tried to summarize the different exif readings ...

Here is the graph I get (I've started to rename some of exif routines to make them more readable):
image

For metadata import we have to control better the red nodes (I hope the graph is readable).

At this point some thoughts. Please correct me if I'm wrong.
1. dt_image_import() calls 1. dt_exif_read(image file) then 2. dt_exif_xmp_read(xmp file). ~~If the xmp already exists we should not read the image file anymore~~ EDIT my mistake, xmp file doesn't contain exif data, all good.
2. dt_exif_apply_global_overwrites() is the current place to apply the import metadata presets. ~~It should be called only if dt_exif_read() has been called by dt_image_import()~~. EDIT. That's ok as dt_exif_read() is called only is exif data are not initialized.
3. ~~in the same way, dt_metadata_set should be called by dt_exif_read_exif_data() and dt_exif_read_iptc_data() only if dt_exif_read() has been called by dt_image_import() ~~ (here we override some metadata with exif or iptc data)~~. EDIT. OK too for same reason as above.
4. ~~dt_imageio_open_exr() -> dt_exif_read_from_blob_to_image_struct() seems to be inconsistent. Metadata should have been read at import time~~. EDIT. OK. See Roman's explanation below.

About the metadata import configuration I'm here:

image

In that image creator and publisher are set hidden in the metadata editor configuration.
We may add a check column to say if the preset overrides the imported value or not. 
Please comment.

### DIFF
--- a/src/common/image.c
+++ b/src/common/image.c
@@ -999,10 +999,10 @@ GList* dt_image_find_duplicates(const char* filename)
 #endif
 }
 
-static int _image_read_duplicates(const uint32_t id, const char *filename)
+// Search for duplicate's sidecar files and import them if found and not in DB yet
+static void _image_read_duplicates(const uint32_t id, const char *filename)
 {
   int count_xmps_processed = 0;
-  // Search for duplicate's sidecar files and import them if found and not in DB yet
   gchar pattern[PATH_MAX] = { 0 };
 
   GList *files = dt_image_find_duplicates(filename);
@@ -1071,11 +1071,7 @@ static int _image_read_duplicates(const uint32_t id, const char *filename)
   }
 
   g_list_free_full(files, g_free);
-
-  return count_xmps_processed;
-
 }
-
 
 static uint32_t dt_image_import_internal(const int32_t film_id, const char *filename,
                                          gboolean override_ignore_jpegs, gboolean lua_locking)
@@ -1165,21 +1161,6 @@ static uint32_t dt_image_import_internal(const int32_t film_id, const char *file
     flags |= DT_IMAGE_HAS_TXT;
     g_free(extra_file);
   }
-
-  // insert dummy image entry in database
-
-  /* Image Position Calculation
-   *
-   * The upper int32_t of the last image position is increased by one
-   * while the lower 32 bits are masked out.
-   *
-   * Example:
-   * last image position: (Hex)
-   * 0000 0002 0000 0001
-   *
-   * next image position
-   * 0000 0003 0000 0000
-   */
 
   //insert a v0 record (which may be updated later if no v0 xmp exists)
   DT_DEBUG_SQLITE3_PREPARE_V2
--- a/data/themes/darktable.css
+++ b/data/themes/darktable.css
@@ -709,6 +709,24 @@ filechooserdialog scrolledwindow
   margin: 0;
 }
 
+frame#import_metadata label
+{
+  margin: 4px 3px 4px 2px;
+  padding: 0px;
+}
+
+frame#import_metadata entry
+{
+  margin: 2px 3px 2px 2px;
+  padding: 0px;
+}
+
+frame#import_metadata checkbutton
+{
+  margin: 0px;
+  padding: 0px;
+}
+
 .dialog-vbox treeview check
 {
   border-radius: 2px;
@@ -908,7 +926,7 @@ entry
 /* checkbutton check state */
 #lib-plugin-ui checkbutton check,
 #iop-plugin-ui checkbutton check,
-checkbutton check
+checkbutton:not(#import_metadata) check
 {
   margin: 5px 8px 5px 0;
 }
--- a/src/common/collection.c
+++ b/src/common/collection.c
@@ -621,8 +621,8 @@ const char *dt_collection_name(dt_collection_properties_t prop)
         if(type != DT_METADATA_TYPE_INTERNAL)
         {
           const char *name = (gchar *)dt_metadata_get_name_by_display_order(i);
-          char *setting = dt_util_dstrcat(NULL, "plugins/lighttable/metadata/%s_hidden", name);
-          const gboolean hidden = dt_conf_get_bool(setting);
+          char *setting = dt_util_dstrcat(NULL, "plugins/lighttable/metadata/%s_flag", name);
+          const gboolean hidden = dt_conf_get_int(setting) & DT_METADATA_FLAG_HIDDEN;
           if(!hidden) col_name = _(name);
         }
       }
--- a/src/common/exif.cc
+++ b/src/common/exif.cc
@@ -153,7 +153,7 @@ static void _get_xmp_tags(const char *prefix, GList **taglist)
   }
 }
 
-GList *dt_get_exiv2_taglist()
+GList *dt_exif_get_exiv2_taglist()
 {
   Exiv2::XmpParser::initialize();
   ::atexit(Exiv2::XmpParser::terminate);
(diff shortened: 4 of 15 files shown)


cryptomilk:
I've implemented the support for reading AVIF. It looks like we read an image twice. Once we open the file and read the pixels and then we read it again for the exif data.

We should only read the file once and be able to have a pointer to pass down the exif data as a raw blob from the image we just read. Importing would be much faster then.

phweyland:
> We should only read the file once and be able to have a pointer to pass down the exif data as a raw blob from the image we just read.

I agree with the principle.

The mipmap node is the one for reading pixels. If everything works well it takes care not to read again pixels if they are in cache.
But this part (on the right of mipmap) may not be run at import time (depending on setting `don't use embedded jpeg`). However I remember I have seen that even to display embedded jpeg thumbnails, exif data are read to get orientation...

Anyway, we may agree that the import flow (upper part of the graph) should always include the capture of all metadata (exif, iptc, xmp and presets if any) and stored them in the blob (database).
The button "refresh exif" (maybe incorrectly named (xmp as well, right ?) - dt_refresh_exif_run()) should allow to repeat that sequence.

Otherwise (mipmap flow) should use the metadata in cache or in the db blob, without needing to read back the files. Of course in that flow occurs the update of xmp files.

Then dt_exif_read() should be called only by dt_image_import(). Correct ?

phweyland:
@LebedevRI @hean01 
Roman, Henrik, one of you could explain the specific call (dt_exif_read_from_blob) for EXR file ? 
Is there a special need for this type of file ? Thanks

LebedevRI:
> @LebedevRI @hean01
> Roman, Henrik, one of you could explain the specific call (dt_exif_read_from_blob) for EXR file ?
> Is there a special need for this type of file ? Thanks

IIRC EXR does not have native support for any such metadata (nor for color space info),
so since it has to be preserved somehow, it is stored in an opaque blob,
albeit that ends up being a non-standard dt-specific thingy
other software likely won't understand. I think.

jenshannoschwalm:
@phweyland off topic, but how did you generate the graphics?

phweyland:
> it is stored in an opaque blob

I see. Thanks for explanation.

So I guess that, as for the other routines dt_imageio_open_xxx(), exif data should be read only once (img->exif_inited false).

phweyland:
> off topic, but how did you generate the graphics?

Manually with yEd graph editor. I have seen some automatic tools googling around but I haven't tested anything... Manual work is a bit tedious but helps to print into my brain... And also I can limit and arrange the graph as needed.

phweyland:
In red where there are some changes.

image

The current import metadata config.

image

phweyland:
@TurboGit Pascal, normally dt_exif_read should not be called in dev flow (i.e. this must be done at import time).
Should I remove this to clean up the graph ?
And maybe log an error if img->exif_inited is FALSE ?

What do you think ?

TurboGit:
> What do you think ?

I suppose your analysis an proposal is correct. But I'm not sure to see all the effects on such a change. Theoretically it should not have any impact (apart better perf as we remove calls). But do you have any idea to why the have the current situation? Just bad code introduced over the time?

Logging an error on the console is certainly a good way to ensure we do not break the code again later.

phweyland:
> Just bad code introduced over the time?

About this... Is there a way, using git, to find the commit(s) which have added/modified a line of code ?



> (apart better perf as we remove calls)

I don't expect any performance improvement here. Just to turn the code a bit more readable, nothing more. Not a big deal in any case.

TurboGit:
> About this... Is there a way, using git, to find the commit(s) which have added/modified a line of code ?

Yes :  

`$ git log -S"some text on this line" `

You'll get the log adding or removing "some text on this line" .

You can see the patch together with the log of course:

`$ git log --patch -S"some text on this line" 
`

phweyland:
> Yes :
> 
> `$ git log -S"some text on this line" `

Excellent. Thank you

phweyland:
@TurboGit Pascal I think I'm done for import metadata. I'm OK for review when you can.
I haven't removed the call to dt_exif_read() from dev because there is no clear benefit.

@Nilvus Nicolas, could check my changes on darktable.css and tell me if there is a better way to control margin on this dialog. In particular the line 929 is a bit weird but if I remove it margin 0 on checkbutton is not applied. I may have also missed some general settings for pop windows.

Thanks

Nilvus:
@phweyland: css is applied in the file order. So, better than adding what you add on line 929, I suggest to change lines added on 700 lines and move them after the 929 line. Just test and check that doesn't change anything else on what you develop here.

phweyland:
> css is applied in the file order. So, better than adding what you add on line 929, I suggest to change lines added on 700 lines and move them after the 929 line. Just test and check that doesn't change anything else on what you develop here.

Thanks for the answer. Did as you said but I get this (lines' height driven by checkboxes), even if put the lines on 700 at the end of the file: 

image

If I put back the weird line, the display is correct again...

Nilvus:
> If I put back the weird line, the display is correct again...

So if it works better, let's this as you done at first. It's also good.

phweyland:
> > If I put back the weird line, the display is correct again...
> 
> So if it works better, let's this as you done at first. It's also good.

OK. Thanks for looking at it !

src/common/image.c
```diff
@@ -999,10 +999,10 @@ GList* dt_image_find_duplicates(const char* filename)
 #endif
 }
 
-static int _image_read_duplicates(const uint32_t id, const char *filename)
+// Search for duplicate's sidecar files and import them if found and not in DB yet
+static void _image_read_duplicates(const uint32_t id, const char *filename)
 {
   int count_xmps_processed = 0;
-  // Search for duplicate's sidecar files and import them if found and not in DB yet
```
TurboGit, line 1005: Doesn't seems related to your PR ?

src/common/image.c
```diff
@@ -1166,21 +1162,6 @@ static uint32_t dt_image_import_internal(const int32_t film_id, const char *file
     g_free(extra_file);
   }
 
-  // insert dummy image entry in database
```
TurboGit, line 1169: Likewise...

src/common/image.c
phweyland: I've tried to fix some re-import behavior because that's not completely consistent but I gave up.
So nothing related, just a misplaced comment.

src/common/image.c
phweyland: Same context as above.
But this comment is misleading (the image is not dummy).

phweyland: So in image.c that's just some cleanup without behavior change.
Not directly related to metadata import.
Let me know if want me to revert this.

src/common/image.c
TurboGit: I don't understand, now there is just comment removed, but I can swear that when I commented there was other lines removed that were part of e26bb3ae430 !!!! A GitHub bug ????

src/common/image.c
```diff
@@ -1071,12 +1071,8 @@ static int _image_read_duplicates(const uint32_t id, const char *filename)
   }
 
   g_list_free_full(files, g_free);
-
-  return count_xmps_processed;
```
TurboGit, line 1075: This is still suspicious I read on this line:
`-  return count_xmps_processed;`

That is, removing the return for this routine having an int as returned. Is that also a GitHub bug ? I'm confused now....

TurboGit: I see, good then I'll remove the return int from spec.

TurboGit: Ok, void set back in you commit. I was confused by the clean-up mixed with the new feature, better to have two commits in this case.

phweyland: > better to have two commits in this case.

Sorry. I'll try to do that way next time. It is still not completely clear for me to get clean commits.
Before I had a lot of commits with a lot of cleaning ones. Was a bit a of mess. Now I join everything in the same commit, meaningless corrections disappear but all changes are mixed. Still to find the right term to get one commit per meaningful change...
Thanks

houz:
I didn't read all the detail in this PR, but it might be related to #4776

PR #4565, opened 2020-03-29
Allow more than 500 images for timelapse

OrthoCube:
The current 500 limit is quite restrictive especially for lengthy timelapse sessions. It would be very helpful if the limit is greatly increased (or removed altogether)

### DIFF
--- a/src/libs/camera.c
+++ b/src/libs/camera.c
@@ -464,7 +464,7 @@ void gui_init(dt_lib_module_t *self)
   gtk_grid_attach_next_to(GTK_GRID(self->widget), GTK_WIDGET(hbox), GTK_WIDGET(modes_label), GTK_POS_RIGHT, 1, 1);
 
   lib->gui.timer = gtk_spin_button_new_with_range(1, 60, 1);
-  lib->gui.count = gtk_spin_button_new_with_range(1, 5000, 1);
+  lib->gui.count = gtk_spin_button_new_with_range(1, 9999, 1);
   lib->gui.brackets = gtk_spin_button_new_with_range(1, 5, 1);
   lib->gui.steps = gtk_spin_button_new_with_range(1, 9, 1);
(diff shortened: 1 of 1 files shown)


TurboGit:
Why not set it to 9999 for now. I don't see a drawback. We need some limit for the spin button, but we may even push it to more if needed.

OrthoCube:
@TurboGit 

> Why not set it to 9999 for now. I don't see a drawback.

I thought about setting it to 9999 :D But thought about being conservative and chose 5000 because maybe the devs had a reason for the original 500 limit.

However, currently, there is no mechanism for cancelling a currently ongoing time lapse so if a user chose 9999 images, and it ended up being too much, the user has no choice but to kill the program to cancel. I was wondering if that might be a problem if the limit is set to too large.

TurboGit:
> However, currently, there is no mechanism for cancelling a currently ongoing time lapse so if a user chose 9999 images, and it ended up being too much, the user has no choice but to kill the program to cancel. I was wondering if that might be a problem if the limit is set to too large.

Well 500 with one image per second is already too long. So let's raise this limit, the proper fix is to propose a cancel of the task.

OrthoCube:
I decided to up the limit up to 9,999 for special cases. Please let me know if that was a bad thing. (This is my second pull request in my whole life)

TurboGit:
Looks good, thanks !

AxelG-DE:
> 
> Well 500 with one image per second is already too long. So let's raise this limit, the proper fix is to propose a cancel of the task.

Aaaeeeehmmmm :-))
Timelapses need at least 25 pics per second, than this sequence is just 20sec, so for TL that is just the basic! One night of astro-TLs you come home with 3.000 pics easily
SCNR :-)

@OrthoCube
if you are on to TLs, may I link you to my [pixls-thread](https://discuss.pixls.us/t/timelapse-photography-and-darktable/14579/59) ? :)

PR #4567, opened 2020-03-29
Changes permitting preview_downsampling.

GrahamByrnes:
This builds on work started by @bastibe in PR #4233. He noticed that a significant amount of computation was being done on the preview image, which could be reduced by downsampling. In fact the variable dev->preview_downsampling already existed, but was fixed to 1.f

Unfortunately Bastian got caught up in some other bugs, and his code had some self-cancelling features. I didn't like the fact that it was necessary to make use of the preview_downsampling variable directly in circle.c and brush.c, but not in ellipse.c or other masks.

So I re-wrote parts of one function in each of circle.c and brush.c borrowing from ellipse.c so that these now call the transform/backtransform functions in develop.c (as they do already for a bunch of other things). The only other changes are a a few lines in each of those functions, taken from Bastian's code, plus a few lines to extract the variable from darktablerc; and finally making the variable visible in darktablerc.

It's available as a choice of 1.0 (no change), 0.5 or 0.25. With 0.25 you have 1/16 the pixels, so there is in principle 44% less processing being done when you make a change in darkraoom.

On the other hand, this might hinder you if you use the preview for precise masking... In that case I'd suggest you leave the value at 1.0

### DIFF
--- a/data/darktableconfig.xml.in
+++ b/data/darktableconfig.xml.in
@@ -1462,6 +1462,19 @@
     <longdescription>interpolation when not viewing 1:1 in darkroom mode: bilinear is fastest, but not as sharp. middle ground is using PPG + interpolation modes specified below, full will use exactly the settings for full-size export. X-Trans sensors use VNG rather than PPG as middle ground.</longdescription>
   </dtconfig>
   <dtconfig prefs="core" section="quality">
+    <name>preview_downsampling</name>
+    <type>
+      <enum>
+        <option>1.0 (none)</option>
+        <option>0.5</option>
+        <option>0.25</option>
+      </enum>
+    </type>
+    <default>1.0 (none)</default>
+    <shortdescription>reduce resolution of preview image</shortdescription>
+    <longdescription>decrease to speed up preview rendering, may hinder accurate masking</longdescription>
+  </dtconfig>
+  <dtconfig prefs="core" section="quality">                                 
     <name>plugins/lighttable/export/pixel_interpolator</name>
     <type>
       <enum>
--- a/src/develop/develop.c
+++ b/src/develop/develop.c
@@ -53,7 +53,6 @@ void dt_dev_init(dt_develop_t *dev, int32_t gui_attached)
 {
   memset(dev, 0, sizeof(dt_develop_t));
   dev->full_preview = FALSE;
-  dev->preview_downsampling = 1.0f;
   dev->gui_module = NULL;
   dev->timestamp = 0;
   dev->average_delay = DT_DEV_AVERAGE_DELAY_START;
@@ -90,7 +89,9 @@ void dt_dev_init(dt_develop_t *dev, int32_t gui_attached)
   else if(g_strcmp0(mode, "waveform") == 0)
     dev->histogram_type = DT_DEV_HISTOGRAM_WAVEFORM;
   g_free(mode);
-
+  char *preview_downsample = dt_conf_get_string("preview_downsampling");
+  dev->preview_downsampling = 
+      (g_strcmp0(preview_downsample, "1.0 (none)") ? 1.0f : atoi(preview_downsample));                                                                      
   dev->forms = NULL;
   dev->form_visible = NULL;
   dev->form_gui = NULL;
@@ -680,6 +681,7 @@ float dt_dev_get_zoom_scale(dt_develop_t *dev, dt_dev_zoom_t zoom, int closeup_f
       if(preview) zoom_scale *= ps;
       break;
   }
+  
   return zoom_scale;
 }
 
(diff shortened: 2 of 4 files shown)


bastibe:
Exciting! I will check out your changes as soon as possible (i.e. hopefully before the end of the week).

This seems a rather better solution overall!

Nilvus:
If I understand this PR, it's possible to improve preview performance just by editing a line in darktablerc. But, if we use a mask, it's better to let the default 1.0. Why not but if we use a mask, we have to remember which setting is set, and edit darktablerc file then restart darktable to have the default 1.0 come back, and after go back again to set a faster set. Or maybe I don't understand correctly that because if I understand correctly, this is not for me a so good solution. And what about denoise work with that? It's often better to see that at 100% zoom, this PR will have impact or not here?

If I understand correctly that, is it not possible to make it more dynamically on darkroom with a slider and dynamic change when using a mask for example (and denoise if needed)?

GrahamByrnes:
Hi @Nilvus, I think I should explain more.
The masking is applied to both the main image in the centre of the screen, which may be at the same time zoomed or distorted, and to the small preview image at the top left in the darkroom.

It's important to note that the masks are specific to the image, what ever value is used for downsampling. There is no need to examine the image at different scales and in any case, the most precise view is in the main image. The preview mainly provides a context view with no zooming or scaling. If you place a mask it will always cover the same pixels in the final exported image. A circle placed in one view may be distorted when viewed in crop & rotate, because the original image pixels themselves are distorted.

The exception is for the color-pickers: these sample relative to the frame, rather than the image: you sample an area of the main frame and obtain an average RGB (or Lab) value, which you can use to set a global change to eg color balance. The selection itself leaves no visible trace and is active only while the picker is active. This is the same in the original code.

In fact, it should be almost always impossible for you to tell what value has been set, given the small size of the preview screen.

The exceptional case is if you are interested in extreme values selected in a mask for calculation of edge-specific masking. While an average of an average is the average, the maximum of an average is smaller than that of the original pixels. This is the origin of the comment on the use of leaving the value at 1.0 for precise masking, pointed out by @aurelienpierre

Nilvus:
@GrahamByrnes: ok, so it was a misunderstanding as I've at first not understand why masking and preview thumbnail could be related here. Clear now. Anyway, it seems that most of the times and for many users, the default 1.0 value is not needed. So why not use another default value to have the speed effect for many users?
Another question: this couldn't be a setting possible to change directly on preferences dialog window and better, without needing to restart darktable? Or not?

GrahamByrnes:
@Nilvus 

So that was interesting. I was taking the speed difference on trust... but thought I'd use it to check for a difference between changing downsampling with or without re-starting.

And... damn, there's no difference.

For a couple of styles applied then unapplied using ctrl-z, typical image processing is about 0.4s, preview about 0.15s, whatever I set downsampling to.

Embarassing.

Maybe I screwed something up... @bastibe had some impressive speed-ups.
I'll need to go away and look at this to see if I screwed up, or if there is nothing to be had here.

Sorry everyone.

GrahamByrnes:
Nope, I screwed up passing the parameter. Sigh.
This will take some fixing. Sorry again.

bastibe:
> @bastibe had some impressive speed-ups.

I have since learned a few things about the speed-ups:

- They seem to be highly dependent on the screen resolution. On a High-DPI screen the difference is much bigger than normal screens. As far as I can tell. This implies that the preview is calculated at resolution connected to the screen resolution somehow? If that's true, maybe a better solution would be to modify the preview resolution proper, instead of mucking with the downsampling factor? What is the default preview resolution, anyway?
- By default, the GPU is supposed to take care of the main view, while the CPU works on the preview. When this works, preview-downsampling has no meaningful effect. But it only works sometimes. Even restarting Darktable sometimes changes the proportions of stuff that is relegated to the GPU. Is there some profiling going on? This part of the data was very inconclusive to me. Nvidia/Windows seems to use more CPU time than AMD/Linux, in my measurements.

So perhaps it would be worthwhile to re-measure the speedup potential on a Low-DPI screen with working GPU acceleration, to get an idea of "typical" speedups.

Nilvus:
> > @bastibe had some impressive speed-ups.
>     * By default, the GPU is supposed to take care of the main view, while the CPU works on the preview. When this works, preview-downsampling has no meaningful effect. But it only works sometimes. Even restarting Darktable sometimes changes the proportions of stuff that is relegated to the GPU. Is there some profiling going on? This part of the data was very inconclusive to me. Nvidia/Windows seems to use more CPU time than AMD/Linux, in my measurements.
> 
> So perhaps it would be worthwhile to re-measure the speedup potential on a Low-DPI screen with working GPU acceleration, to get an idea of "typical" speedups.

And not forgot those who only have CPU and no OpenCL compatible card. So it need to thought on both CPU+GPU and CPU only, but also Low and High DPI screens. And of course all mix computers possible with that !

GrahamByrnes:
Well, I'm qualified... I have a 2k screen and no OpenCL :-)

A quick recap on the errors: I used atoi to read in the value of the parameter, so it remained at the default. Doh.

Otoh, it seem that some of my blind modifications were actually correct: circle.c, ellipse.c and brush.c all mis-behave in the same way, as do the color-pickers: the placed shape is exactly where I point the cursor. It's just that the drawing of the circle/ellipse/path/sampling area is contracted towards the centre of the image, ie at 25% of actual size and only 25% of the way from the centre to the cursor. So basically I just need to find what I broke.

Otoh, in response to @Nilvus whose question broke my day-dream software: yes, in fact you need to re-start to change the parameter. It would be nice to know if that has to be so.

bastibe:
An additional difficulty is making sure that mask positions are always saved at consistent pixel positions regardless of preview downsampling. They sometimes showed correctly in my versions, but would move if I changed the preview downsampling and restarted.

Nilvus:
> Otoh, in response to @Nilvus whose question broke my day-dream software: yes, in fact you need to re-start to change the parameter. It would be nice to know if that has to be so.

Sorry for that ;-) ;-)
We are all here for the same thing: improve darktable and find the better way we can to do that! And thanks for what you try on that with @bastibe.

PR #4571, opened 2020-03-30
bugfix: prevent extra duplicate image on import

### LINKED ISSUE #4385: Import creates two image instances when xmp filename ends "_01.cr2.xmp"
Import a folder that contains a previously edited image and an xmp file. If the xmp file has a filename "<img_filename>_01.cr2.xmp" darktable creates two duplicate instances of the image and generates a second xmp file (named "<img_filename>.cr2.xmp").  Same thing happens if I rename it to "<img_filename>_02.cr2.xmp".

If the xmp is named "<img_filename>.cr2.xmp" only a single instance of the image is imported.

Many of my xmp files are like this because when I transferred my edits to the new linear RGB workflow, I had to create duplicates in order to reset the module order to the new default, and later deleted the originals.  Either way, one xmp file should only ever create one image instance when imported.

Possibly related to https://github.com/darktable-org/darktable/issues/3978?

darktable version: 3.1.0+754~g4397473da
OS: ArchLinux
OpenCL: not active

### LINKED ISSUE #4562: Probably missing refresh when importing with pre-existing sidecars
Importing files with a leading underscore in name might lead to sort of "double import" with one of those images **unaccesible**.

**To Reproduce**
1. You need to import files from a directory where **no** image is in the database already (has been imported before). You can do so by dt-remove all files from that dir and re-import.

2. There must be one or more files with filename having a **leading** underscore letter.

3. One of the images seems to be **double** imported, one of those in not accessible (you see the skull). This will not be shown later when switching to lt mode again.

**Platform**
 - Darktable master Version: 3.1.0+1045
 - OS: Fedora 31
 - OpenCL doesn't matter

**Additional context**
This seems to be introduced within the last 2 weeks, unfortunately i currently don't have time to test via bisect myself ..

### LINKED ISSUE #4568: image id - group id issue on import of images when xmp
When I import images without xmp file image id = group id: correct

image

**Describe the bug**

When I import the same images with an xmp file image id = group id + 1 (and image id is pulling the value of group id): incorrect

image

EDIT: same behavior for both single images and folder

**Expected behavior**
group id must be equal to image id at this stage

**Context**
 - Darktable Version: 3.1.0+1049~gd1f0f0e14
 - OS:  Debian 10

elstoc:
second attempt to resolve issue with extra duplicate appearing on import
of files/folders where the minimum xmp version number is greater than
zero.

this replaces the previous solution delivered in #4393 while also
resolving some issues with group id not being set correctly and with
refresh of the lighttable.

Resolves #4385. Resolves #4562. Resolves #4568.

### DIFF
--- a/src/common/image.c
+++ b/src/common/image.c
@@ -998,8 +998,7 @@ GList* dt_image_find_duplicates(const char* filename)
 #endif 
 }
 
-
-int dt_image_read_duplicates(const uint32_t id, const char *filename)
+static int _image_read_duplicates(const uint32_t id, const char *filename)
 {
   int count_xmps_processed = 0;
   // Search for duplicate's sidecar files and import them if found and not in DB yet
@@ -1040,9 +1039,28 @@ int dt_image_read_duplicates(const uint32_t id, const char *filename)
       g_free(idfield);
     }
 
-    const int newid = dt_image_duplicate_with_version(id, version);
+    int newid = id;
+
+    if(!count_xmps_processed)
+    {
+      //this is the first xmp processed, just update the passed-in id
+      sqlite3_stmt *stmt;
+      DT_DEBUG_SQLITE3_PREPARE_V2
+        (dt_database_get(darktable.db),
+         "UPDATE main.images SET version=?1, max_version = ?1 WHERE id = ?2", -1, &stmt, NULL);
+      DT_DEBUG_SQLITE3_BIND_INT(stmt, 1, version);
+      DT_DEBUG_SQLITE3_BIND_INT(stmt, 2, id);
+      sqlite3_step(stmt);
+      sqlite3_finalize(stmt);
+    }
+    else
+    {
+      //create a new duplicate based on the passed-in id
+      newid = dt_image_duplicate_with_version(id, version);
+    }
     dt_image_t *img = dt_image_cache_get(darktable.image_cache, newid, 'w');
     (void)dt_exif_xmp_read(img, xmpfilename, 0);
+    img->version = version;
     dt_image_cache_write_release(darktable.image_cache, img, DT_IMAGE_CACHE_RELAXED);
 
     count_xmps_processed++;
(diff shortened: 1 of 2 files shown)


GrahamByrnes:
Just a couple of fiddly comments:
- I think the naming convention proposed by @phweyland was _image_read-duplicates... rather than _dt_image_read_duplicates
- Do you really need to set version=-1 then go to if(!test)... else following from line 1017 ?
You could just set it zero then go straight to the else:

version=0;
if(strncmp(xmpfilename, pattern, sizeof(pattern)) 
{
  ...
  version = atoi(idfield);
  g_free(idfield);
}

Otherwise, I like this version much more :-)
Cheers, G

elstoc:
> the naming convention

I agree. I'll change it.

> Do you really need to set version=-1 then...

That was preexisting code that I haven't changed in this PR.

AlicVB:
> Please note there is still a minor refresh issue in the lighttable, whereby the version number is not listed correctly where no v0 xmp file exists.

you mean in the image information table in the left panel, right ? If so, you'll have to update the image cache. Not sure how to do that... maybe you can load the cache for writing, update the img structure and realease the cahe... or maybe just use `dt_image_cache_remove`

elstoc:
> you mean in the image information table in the left panel, right

That's right. I think someone might have to give me some guidance before I resolve that one, since I don't know how the image cache works.

AlicVB:
First point : do you know at which point in the code we need to do the refreshing ? That mean at which point we are sure that all the version work is finished.

elstoc:
The only part which might change the version number is the UPDATE in _image_read_duplicates at line 1048 of image.c but we can be sure that **all** version stuff is complete after the call to _image_read_duplicates at line 1118 or 1332.

AlicVB:
then, I would add something like : 
```
dt_image_t *img = dt_image_cache_get(darktable.image_cache, id, 'w');
img->version = ?????????;
dt_image_cache_write_release(darktable.image_cache, img, DT_IMAGE_CACHE_RELAXED);
```
at this places (if you can call it just 1 time it would be better...)

elstoc:
Thanks @AlicVB that (img->version=version) appears to have sorted it.

phweyland:
Looks good to me ! Thanks.

TurboGit:
Thanks !

PR #4575, opened 2020-03-30
Powerfull extract_wb for superior wb presets extraction and finetuning

johnny-bit:
While working on `temperature.c` changes I noticed the need for easier tooling for WB extraction.

This is effect of over week's worth work although only partial.

First - this adds `extract_wb_from_images.sh` script that runs same `exiftool` command as `extract_wb` in order for users to avoid needing to install ruby, nekogiri & all that. This removes couple pain points in wb presets generation.

Next are changes in `extract_wb`:
- Detects finetuning used, so user won't need to manually edit generated data
- detects whether shift over Green-Magenta axis happened and complains about it
- support over 25 steps of WB finetuning (nikon)
- doesn't scream at user about ignored presets (:wink:)
- detects and mention if user didn't do finetuning over whole Amber-Blue axis in case finetuning was done

TODO:
- [x] write improved wiki entry for camera support, especially wb extraction - https://github.com/darktable-org/darktable/wiki/White-balance-presets
- [x] install `extract_wb_from_images.sh` in same step as other tools, so it's available for all users
- [x] test on more reallife data (ongoing)
- [x] integrate with `temperature.c` code (ongoing)

Currently tested with various Canon cameras and data provided selflessly by good members of pixls.us on https://discuss.pixls.us/t/call-for-white-balance-finetuning-samples/17137 thread.

thanks to them it's confirmed that the scripts work flawlessly on all cameras by:
- Canon 
- Nikon
- Sony
- Pentax
- Panasonic
- Olympus

Fuji probably doesn't have finetuning in exif

other cameras - need more samples but I'm sure it'll be fine

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -363,6 +363,15 @@ if(USE_XMLLINT)
   endif(${Xmllint_BIN} STREQUAL "Xmllint_BIN-NOTFOUND")
 endif(USE_XMLLINT)
 
+find_program(exiftool_BIN exiftool)
+if(${exiftool_BIN} STREQUAL "exiftool_BIN-NOTFOUND")
+  message(STATUS "Missing exiftool")
+  set(HAVE_EXIFTOOL 0)
+else()
+  message(STATUS "Found exiftool")
+  set(HAVE_EXIFTOOL 1)
+endif()
+
 # done with looking for programs
 if(NOT EXTERNAL_PROGRAMS_FOUND)
   message(FATAL_ERROR "Some external programs couldn't be found")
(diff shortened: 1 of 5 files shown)


johnny-bit:
Hey @TurboGit - this is no longer WIP. Confirmed to work with majority of camera manufacturers. It seems that Fuji doesn't expose finetuning data in their exif, but otherwise every other brand works.

I also created 2 new wiki pages:
https://github.com/darktable-org/darktable/wiki/Camera-support
https://github.com/darktable-org/darktable/wiki/White-balance-presets

Please check if everything's ok with them :)

TurboGit:
Looks good to me. Thanks.

PR #4579, opened 2020-03-31
Improve UI for hdr and monochrome images

jenshannoschwalm:
Both image types often want a different workflow than usual, for hdr's there has been a short discussion on pixls.us about this.

It might be helpful to visualize the information about such an image in the lighttable window, i have opted for a simpe textual addition instead of another icon there.
pic1

Also it is good to have a darktable tag also for hdr images, there is such already for monochromes.
pic2

The detection of "image being a hdr or b&w" is normally done inside `dt_imageio_open`, this is
ok and should be kept but we can also do an early detection when reading exif data in
`dt_exif_read_exif_data`. At least for files tested via `dt_image_is_hdr` or for dng's this
is pretty easy.

Why is this good? We have the monochrome/hdr flags already correctly set while importing.

Any performance penalty? No

Is it complete? i guess so. The **collect images** list already seem pretty large and as this is a somewhat special workflow searching via tags seems to be ok/better.

### DIFF
--- a/src/common/exif.cc
+++ b/src/common/exif.cc
@@ -1154,6 +1154,40 @@ static bool dt_exif_read_exif_data(dt_image_t *img, Exiv2::ExifData &exifData)
       }
     }
 
+    int is_monochrome = FALSE;
+    int is_hdr = dt_image_is_hdr(img);
+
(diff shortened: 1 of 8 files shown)


TurboGit:
Ok, looks good to me. Thanks.

PR #4580, opened 2020-03-31
thumb : fix missing variables.h include

### LINKED ISSUE #4544: Problem opening image by command line
**Describe the bug**

When I try to open a raw file by `darktable filename` I get the error `error loading: [filename]`. However when I do it again, the image loads correctly. It doesn't seem to matter if an instance of darktable was already running or not. Moreover sometimes I get duplicate "film rolls" for the same folder in lighttable view with some pictures in one and some in the other. I'm attaching parts of `darktable -d all` output for the described actions. I'm running current HEAD. It seems that the bug was introduced fairly recently as I recompile Darktable quite often and I noticed it only today. I can try bisect if it would help.

I tried the same after removing `.config/darktable`. First image seemed to open fine, but second or so produced the error. I also noticed once that newly opened image had a wrong thumbnail in darktable "film strip" area - thumbnail from completely different picture which I processed a month ago.

**To Reproduce**
1. Run `darktable filename` once or a few times for different files
2. See error

 - Darktable Version: HEAD 8e4d22171fafc9546c6e775b03eaa5c3e0b099b1
 - OS: Gentoo Linux
 - OpenCL: yes or no, no change
 - AMD Radeon Fury

AlicVB:
fix #4544

### DIFF
--- a/src/dtgtk/thumbnail.c
+++ b/src/dtgtk/thumbnail.c
@@ -26,6 +26,7 @@
 #include "common/image_cache.h"
 #include "common/ratings.h"
 #include "common/selection.h"
+#include "common/variables.h"
 #include "control/control.h"
 #include "dtgtk/button.h"
 #include "dtgtk/icon.h"


TurboGit:
Thanks !

Issue #4528, opened 2020-03-22
"Plain paper" message shown at startup


AlicVB:
**Describe the bug**
At startup (not always but very often) I have the log message "Plain paper" shown on screen.
I think that it's related to my recent changes in log drawing.

But to debug and fix, I need to find where that log message comes from, and I didn't manage... I though of `print.c` but I didn't find anything useful...
If someone know which `dt_control_log` call can display this msg, I would be grateful...

**Platform (please complete the following information):**
 - Darktable Version: master
 - OS: xubuntu 19.10
 - OpenCL both

Nilvus:
Will not help probably but yes I've just seen that on a darktable startup (in light view). In french, it's translated ('Papier ordinaire'. I've just search 2 times (on github file and with poedit) and can't find 'Papier ordinaire' or even the only word 'ordinaire' on fr.po file.

So, could it be from an external library? Gtk one?

AlicVB:
Same as you, unsuccessful search...
In fact I think the message may come from be cups or whatever. But even in that case it has to be shown by `dt_control_log`, and I've not seen a place in `print.c` where we show a message without some prefix or suffix...

elstoc:
If it's an option in a combobox it's probably caused by #4468, which adds a toast whenever a combobox setting is changed via `dt_bauhaus_combobox_set`. We're gonna be chasing these down for a while I suspect. The solution will be similar to that in #4507.

elstoc:
I can't reproduce myself (I don't print from darktable) but I'd say a good candidate would be line 518 of `libs/print_settings.c`.

jenshannoschwalm:
here it prints (null)

jenshannoschwalm:
Indeed it's the printer telling about the media found in dt printer options.
For cups-pdf this is null, otherwise what#s been selected.

AlicVB:
oh... you means its not my fault ? quite unusual this times ;)
Anyway, I'll have a look, I didn't think of automatic combobox msg...

elstoc:
@AlicVB: There's another one that appears when you change the display profiles in the lighttable but I can't pin down where it's coming from - none of the calls to `dt_bauhaus_combobox_set` in `views/lighttable.c` seem to be causing it (or at least the fix from #4507 isn't disabling the toast). If you have any suggestions I'd be grateful.

elstoc:
[...]

PR #4581, opened 2020-03-31
add dynamic accelerators to comboboxes

elstoc:
adds ability to assign a keyboard shortcut to dynamically change
comboboxes. Hold the key and scroll up/down to change to the
next/previous combobox value.

### DIFF
--- a/src/gui/accelerators.c
+++ b/src/gui/accelerators.c
@@ -233,9 +233,12 @@ void dt_accel_register_combobox_iop(dt_iop_module_so_t *so, gboolean local, cons
   gchar accel_next_path_trans[256];
   gchar accel_prev_path[256];
   gchar accel_prev_path_trans[256];
+  gchar accel_dynamic_path[256];
+  gchar accel_dynamic_path_trans[256];
 
   dt_accel_t *accel_next = (dt_accel_t *)g_malloc(sizeof(dt_accel_t));
   dt_accel_t *accel_prev = (dt_accel_t *)g_malloc(sizeof(dt_accel_t));
+  dt_accel_t *accel_dynamic = (dt_accel_t *)g_malloc(sizeof(dt_accel_t));
 
   gchar *module_name_fixed = dt_util_str_replace(so->name(), "/", "-");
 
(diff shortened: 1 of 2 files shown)


johnny-bit:
OMG YAAAASSS!!!

dterrahe:
With this in place it should be straightforward to implement this in my MIDI module as well (with a similar if(w->type == DT_BAUHAUS_SLIDER) conditional).

On the x-touch mini, I'll try to map each option in a combobox to a different LED on the ring around the rotator. So I should map 1-> 127/13, 2-> 2*127/13 etc, or something like that. The BCF2000 has 15 lights in the LED ring, so Yeah, another configuration option! ;-)

That would be _another_ place where I'd need to use the dt_control_log code to generate a toast. Maybe we should get @houz opinion again on adding a dt_bauhaus_combobox_set_notify call in bauhaus as you suggested in https://github.com/darktable-org/darktable/issues/4528#issuecomment-602270639

houz:
Absolutely not. If MIDI support gets added then not ontop the accelerators, wrapping layers over layers. It needs a clean rewrite of the whole control system in dt. All that code is basically unmaintainable already, I will not agree to monkey patching more stuff in.

johnny-bit:
So... This PR is OK to integrate? It very well works for my usecase and UI stuff, speeds up some ops :smiley: It's not yet so complicated, but goes into "right" direction usability-wise.

I understand that MIDI system is too big to monkey-patch in, but this small change looks nice...

As for whole control system rewrite: I'm very interested in it and I see that @elstoc and @dterrahe too, so where can we discuss design of it?

elstoc:
Leaving aside the MIDI support though, it might still be better to have a function dt_bauhaus_combobox_set_notify to replace the three places I now have the same combobox toast message (just to remove some of the duplication).

johnny-bit:
(just a note - @houz is OK with merging this PR, just expressed that the changes ned to be taken into account when doing new input system for darktable, that could handle all the things and be very much resilient to random breakages. that work though is long and hard. I'll post later on in the wiki wat was the discussion about)

TurboGit:
Works for me. Thanks!

Issue #4584, opened 2020-03-31
dead lock when importing image with xmp


phweyland:
**Describe the bug**

In lighttable select an image for which there is an xmp file and some metadata inside.
Import this image, dt freezes

**Platform (please complete the following information):**
 - Darktable Version: 3.1.0+1053~ge26bb3ae4-dirty
 - OS: debian 10

**Additional context**

_image_read_duplicates() gets the image cache before calling dt_exif_xmp_read().
Reading the metadata dt_metadata_set() raises the signal DT_SIGNAL_MOUSE_OVER_IMAGE_CHANGE which triggers metadata_view to refresh the UI.
_metadata_view_update_values() requests for the image cache. 
If the image is selected in lighttable the lock occurs.

What surprises me is that DT_SIGNAL_MOUSE_OVER_IMAGE_CHANGE is configured to be asynchronous. Then dt_control_signal_raise() should exit before _metadata_view_update_values() ~~ends~~ gets the cache. But this is not the case.

Did I miss something ?
Any direction to solve this in the best way ?

PR #4585, opened 2020-03-31
bugfix - reimport image with xmp

phweyland:
Quick fix for  #4584
However it would be probably more general to have dt_control_signal_raise() asynchronous.


TurboGit:
Works for me. Thanks.

PR #4592, opened 2020-04-01
add Exif.Photo.ExposureBiasValue to variables and image info

phweyland:
Relay on #4463 started by @aurelienpierre 

Overlay using $(EXIF_EXPOSURE_BIAS) with existing bias
image

Overlay when bias is missing in cache
image

with this formula:
image

On image information
image

Of course to get the value it is enough to select image(s) and `refresh exif`

### DIFF
--- a/src/common/database.c
+++ b/src/common/database.c
@@ -44,7 +44,7 @@
 
 // whenever _create_*_schema() gets changed you HAVE to bump this version and add an update path to
 // _upgrade_*_schema_step()!
-#define CURRENT_DATABASE_VERSION_LIBRARY 25
+#define CURRENT_DATABASE_VERSION_LIBRARY 26
 #define CURRENT_DATABASE_VERSION_DATA     6
 
 typedef struct dt_database_t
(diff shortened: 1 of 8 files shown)


src/common/database.c
```diff
@@ -155,7 +155,8 @@ static gboolean _migrate_schema(dt_database_t *db, int version)
       "raw_auto_bright_threshold REAL, raw_black INTEGER, raw_maximum INTEGER, "
       "caption VARCHAR, description VARCHAR, license VARCHAR, sha1sum CHAR(40), "
       "orientation INTEGER, histogram BLOB, lightmap BLOB, longitude REAL, "
-      "latitude REAL, color_matrix BLOB, colorspace INTEGER, version INTEGER, max_version INTEGER)",
+      "latitude REAL, color_matrix BLOB, colorspace INTEGER, version INTEGER, max_version INTEGER,"
```
TurboGit, line 158: I don't think this one is correct, and you missed the update in _create_library_schema().

phweyland: You are right, fixing that.

phweyland:
Should work better now. Thanks

TurboGit:
Looks ok to me now. Thanks.

Issue #4586, opened 2020-03-31
Base Curve gets removed after export


Grizzly1701:
**Describe the bug**
Darktable removes the base curve (that is automatically added by default when importing a folder) after exporting an unedited RAW file.

**To Reproduce**
1. Open Folder with RAW files (darktable automatically adds 8 steps, including base curve)
3. Export it as jpg (maybe works with other file formats too)
4. Reload the folder by clicking folder in the import section and choosing the folder in the popup
5. Double click on the RAW file in lighttable to get into darkroom
    --> Base curve is not in the history anymore, image seems flat, base curve module shows no base curve anymore (only 7 steps in history)

**Expected behavior**
Base curve is still applied after reopening the folder in darktable and going into darkroom

**Screenshots**
Before exporting (it uses the base curve that is automatically applied if you import a folder with RAW files):
https://ibb.co/H2hkYLL
Right after exporting the jpg and reopening the folder with darktable, going into darkroom: base curve is not there anymore:
https://ibb.co/WcxTd7M

**Platform (please complete the following information):**

    Darktable Version: 3.0.1
    OS: Windows 10
    GTX 1070

Additional context
It does not happen in following cases:
If I add the base curve manually (for example if I change something from the default setting in the basecurve module).
If I go into darkroom again (doubleclicking on the .NEF file) before reloading the folder in darktable.

**Attachements**

[DSC_0510_Before_Export.zip](https://github.com/darktable-org/darktable/files/4414221/DSC_0510_Before_Export.zip)
I had to delete the .NEF file in the second zip because it was too big for github (you need to copy the .NEF file from the zip above into the folder of the second zip to be able to open it in darktable)
[DSC_0510_After_Export.zip](https://github.com/darktable-org/darktable/files/4414266/DSC_0510_After_Export.zip)

Nilvus:
As said, it's not a bug but a normal behavior. Base curve use as changed. I copy here my answer on your first duplicate issue about this here:

> But the base curve is something that darktable applies on its own and should be still there after an export as it is also there before export after deleting the history. 

It's not anymore. You could even disable permanently base curve (filmic RVB replaced it for most users now) and even sharpen on preferences dialog. So that's why on XMP, we can't now consider base curve as a default module, even if on default start, base curve is automatically applied. It's no more considered as a default module like, for example, black and white point or white balance.

You could check article release to describe changes on darktable 3.0 here: https://www.darktable.org/2019/12/darktable-30/

[...]

PR #4595, opened 2020-04-02
Only set DT_IMAGE_AUTO_PRESETS_APPLIED while in gui mode. Fixing #4586

jenshannoschwalm:
The reporting issue is misleading, when watching the video the sequence is:
1. develop an image
2. discard history
3. export the image
4. reimport the image

The point is, after discarding the flag is cleared, export sets the flag but leaves no history and reimporting does **not** apply presets ... because the flag is already set.

How to fix? The auto apply should only be done while in gui.

### DIFF
--- a/src/develop/develop.c
+++ b/src/develop/develop.c
@@ -1342,7 +1342,8 @@ static gboolean _dev_auto_apply_presets(dt_develop_t *dev)
 
   // flag was already set? only apply presets once in the lifetime of a history stack.
   // (the flag will be cleared when removing it)
-  if(!run || image->id <= 0)
+  // Also make sure this is only done while in gui
+  if(!run || (image->id <= 0) || (!dev->gui_attached))
   {
     dt_image_cache_write_release(darktable.image_cache, image, DT_IMAGE_CACHE_RELAXED);
     return FALSE;


TurboGit:
Frankly to me this is not a bug. As I said to OP in the first issue I see nothing wrong.

You clear the history and export. Then the XMP is "frozen" as-is. If you have exported the picture without any history while would re-importing it should auto apply the presets?

Grizzly1701:
@TurboGit in my opinion if there are presets that get auto applied by discarding history it should save the steps made by the presets in the xmp file. When reimporting it should just reopen the xmp file (and not change it/deleting the steps of the presets) which would mean that the steps automatically made by the presets would remain applied.

parafin:
Discarding history should reset image to the state of freshly imported image, that means applying auto-applied presets, base curve included if it is enabled. This is an actual bug, no doubt about it.

jenshannoschwalm:
@Grizzly1701 that's not the point. Discarding history **removes** the history and does not set it to something like default.

@Turbogit This is what i concider to be a bug.

If i have no history and export i would either expect
1. apply presets while exporting **and** keep them in history 
2. do not apply presets and wait for them to be applied when developing in darkroom.

Both ways seem to be ok for me. But here we apply presets while exporting, don't keep the history **but** have the flag switched on.

jenshannoschwalm:
@parafin that's not exactly how it works. If i discard history dt removes history data, cleans up the xmp  and marks the image as not-having-applied-presets yet.

parafin:
I'm talking about from the point-of-view of a user, not how it works internally. If user discards history, he should get the same state as if he just imported the image, whether he exports it of opens in darkroom.

jenshannoschwalm:
@parafin yeah - sure - i misunderstood. That means you opt for 1. ?

parafin:
I guess so, yes, it makes more sense.

jenshannoschwalm:
ok and understood. That would require a different pr (this one chose 2. because that was what i expected)

let's see what @Turbogit and maybe other people think and i might close this and come back later ...

TurboGit:
You can't apply preset while exporting as this means that you cannot export a plain RAW without any module activated. This is needed for creating noise profile IIRC and also to create a preset for the color lut module.

TurboGit:
Oh, and I also export RAW without any presets to show the original picture and the one after development.

parafin:
You can go into darkroom and create any history you like, even without any iops, but discard history button is not for that.

TurboGit:
> You can go into darkroom and create any history you like, even without any iops, but discard history button is not for that.

I see and I agree for discard history. Yet if you have many images to export without any modules activated you'll have to go into darkroom one by one. If you want to create a noise profile out of 25 images it will be very painful. I suppose one can copy/paste deactivated modules, not impossible but still painful compared to what we have today.

I won't block this as I don't have string opinion so yes probably 1 is the way to go.

parafin:
You can create an xmp with desired (empty) history and use with dt-cli or import in GUI

jenshannoschwalm:
Ok. I see agreement on 1.

Will close this for now and look for a proper fix a.s.a.p.

Grizzly1701:
> @Grizzly1701 that's not the point. Discarding history removes the history and does not set it to something like default.

Ok I understand your statement: Discard history takes away everything created by darktable (including presets). 
Why does the exported jpg file then look like it has the base curve? Why does the history stack in darkroom say "base curve" after using "discard history" and going into darkroom? Both give the impression of darktable applying the presets when using "discard history".

Either it shouldn't apply base curve when using "discard history" at all (means it shouldn't show base curve in darkroom before exporting and not putting the base curve onto the jpg) or it should apply it and write it into the xmp file (and then of course remain applied after reimporting the folder because it would just read the xmp file (as it does now too doesn't it?))

To sum up things I would of course agree more with solution 1 but I would not apply the presets while exporting. Wouldn't this mean if I go to darktable after "discard history" in lighttable it wouldn't show my presets in the history stack but then apply the presets on the jpgs/exported images (exported images would look different than the image that showed up in darkroom)? Correct me if I misunderstood your idea.

jenshannoschwalm:
> Why does the exported jpg file then look like it has the base curve?

Because whenever an image is opened (whether in darkroom mode or lua, export or cli) it will apply all defaults if there no history. So your exported image is ok :-)

Internally there is wrong housekeeping about a flag in dt database and the history and xmp files. THIS is the bug that needs to be fixed.

It took me a while to understand that you clicked on discard history. This is something pretty unusual in the workflow so i didn't expect that. Normally it's only done if your image development is a complete mess and you want to start from the beginning.

BTW get yourself used to the history stack on the left side! See what happens while selecting earlier steps or compress history.

Grizzly1701:
> Because whenever an image is opened (whether in darkroom mode or lua, export or cli) it will apply all defaults if there no history. So your exported image is ok :-)

It does not apply presets when opening the RAW file (double clicking in lighttable --> opening in darkroom) the second time though (where the history is discarded): https://youtu.be/GFaFrRnF9lQ?t=33 0:33 in my video.

jenshannoschwalm:
Just wait for the next & correct pr and you will see everything is right :-)

Grizzly1701:
Sure. I think the same :) 

Just wanted to make sure you understand me right.

Thank you for your time and efforts!

jenshannoschwalm:
Sometimes it's difficult to understand what **exactly** is the technical bug but I think I got it 🤔

PR #4600, opened 2020-04-03
keyboard shortcuts to show/hide lib modules

### LINKED ISSUE #4126: Please include shortcut options for left side modules in darkroom
In my workflow in the darkroom I'm using keyboard shortcuts quite a lot to switch between modules or switch them on or off. After getting used to it, it speeds up the process significantly and now it feels really uncomfortable not to have shortcuts for modules which are often used.

Unfortunately there is no option to assign shortcuts for those modules on the left side, e.g. the stack, image information, mask manager and so on. 

The three I mentioned may take up a lot of space as they grow during the editing process. So that switching from one to another may require bringing the mousepointer to the scrollbar, scroll the mousewheel and a ckick on the module you are interested in.

Keyboard shortcuts could speed this up significantly. 

So it would be nice, if you could assign the same options for setting up keyboard shortcuts for the modules in the left column as for those in the right column. For my workflow options to unfold / fold the modules or to focus (scroll) them would be sufficiant but I think it would be generally be more helpfull if all modules would offer the same options for keyboard assignment, no matter if they seem to make sense for the individual module or not. In my experience others have come up with clever applications of options, which I couldn't even imagine.

elstoc:
adds a new keyboard shortcut "show module" to all lib modules, which
has the same action as clicking on the module header (i.e.
expands/collapses the module)

Resolves #4126

### DIFF
--- a/src/libs/lib.c
+++ b/src/libs/lib.c
@@ -616,6 +616,10 @@ static int dt_lib_load_module(void *m, const char *libname, const char *plugin_n
   {
     dt_accel_register_lib(module, NC_("accel", "show preset menu"), 0, 0);
   }
+  if(module->expandable(module))
+  {
+    dt_accel_register_lib(module, NC_("accel", "show module"), 0, 0);
+  }
 #ifdef USE_LUA
   dt_lua_lib_register(darktable.lua_state.state, module);
 #endif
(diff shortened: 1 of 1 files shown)


TurboGit:
Works for me. Thanks!

StefanDieke:
Excellent news! Thanks a lot! For me this is a huge improvement!

If I may express a related wish: now please add an option to 'show only a single left-hand-side module' to the settings. That would make this new feature even more convenient to use and would speed up the workflow further.

elstoc:
Please could you raise a separate feature request.

Issue #4576, opened 2020-03-30
Tags are applied to wrong images in darkroom


rabauke:
**Describe the bug**

When tags are added to an image in darkroom view, the tags may be applied to the wrong image.

**To Reproduce**

- Open a collection with several images in lighttable view.
- Select an image (image A) in lighttable view.
- Switch to darkroom view.
- Select another image (image B) from the film strip.
- Press Ctrl+T and enter a tag.

**Expected behavior**
The tag should be applied to image B, it is, however, applied to image A.

**Platform (please complete the following information):**
 - Darktable Version: git master
 - OS: Ubuntu 19.10, Intel x64

**Additional context**
The following issue might be related to this bug: Above the image in darkroom view, a text is shown: "1 image of  X (#Y) in current collection is selected". The number Y does _not_ change when switching from image A to image B.

phweyland:
Hi Heiko,
This is related to new behavior filmstrip. Pinging @AlicVB .
Ctrl-T works with selected images if any, otherwise it uses dt_control_get_mouse_over_id().
1. Selected image doesn't change when you click on filmstrip or moves with space/backspace.
2. if you have no selected image, and your mouse is "somewhere" on filmstrip the tag goes there ...

Aldric, what do you think ? Should I change the way to identify the current dev image when in darkroom ? But in that case we cannot apply the tag on selection in darkroom (which may not be stupid)....

rabauke:
@phweyland Thanks for your explanation. I think the current behavior is very unintuitive and inconvenient. In particular, when one switches from lighttable view to darkroom view via double click on an image, which also selects the image one clicks on. In my view it would be very reasonable to apply tags in darkroom view always and only to the image which is currently edited.

phweyland:
>  In my view it would be very reasonable to apply tags in darkroom view always and only to the image which is currently edited.

Agreed.

phweyland:
@Turbogit @AlicVB 
If I'm not mistaken with the previous darkroom behavior selected image = edited image. 
The tag module (not only ctrl-T feature) works on this assumption (and probably some other modules (export ?) which apply on different views).
I don't think this is good idea to check everywhere is we are in darkroom before using selected image.
And I agree with Heiko the current behavior of filmstrip under darkroom is not convenient.

Is there any discussion about this I can refer to understand the basics ?
What is the plan ?  Any help needed ?

AlicVB:
@phweyland : sorry, I'm missed that issue...
[...]

PR #4602, opened 2020-04-03
tagging update

phweyland:
- wire tagging on dt_view_get_images_to_act_on()
- rename static routines in tagging.c

solve #4576

notes
- hovered image, if any, is still the first target even in darkroom
- dt_image_synch_xmp() is also wired on dt_view_get_images_to_act_on() now
- tagging is still managing grouped images. When integrated in dt_view_get_images_to_act_on() some more cleaning can be done.

### DIFF
--- a/src/common/exif.cc
+++ b/src/common/exif.cc
@@ -1287,7 +1287,13 @@ void dt_exif_apply_default_metadata(dt_image_t *img)
     }
 
     str = dt_conf_get_string("ui_last/import_last_tags");
-    if(str != NULL && str[0] != '\0') dt_tag_attach_string_list(str, img->id, FALSE, FALSE);
+    if(img->id > 0 && str != NULL && str[0] != '\0')
+    {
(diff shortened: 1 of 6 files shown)


TurboGit:
Works for me. Thanks!

rabauke:
Question regarding renameing static functions: Is there any general agreement that names of static functions should start with an underscore in darktable?

phweyland:
> Question regarding renameing static functions: Is there any general agreement that names of static functions should start with an underscore in darktable?

I may be mistaken, but that the convention I've understood from some comments from Pascal. I've looked for some text or guideline but I haven't not found (wiki, ...). I think it is helpful but if it is not the right convention for dt, please tell me.

TurboGit:
> Question regarding renameing static functions: Is there any general agreement that names of static functions should start with an underscore in darktable?

Yes, that's the convention.

Issue #4597, opened 2020-04-02
Allow n>3 in $(CATEGORYn(category)) variable


EC1000:
**Is your feature request related to a problem? Please describe.**
In the export module, it is possible to use the (very nice and nifty) `$(CATEGORYn(category))` variable, which returns the tag name of level n [0,3] of selected category (or tag). 

It is then possible to use up to 4 levels of tags under a category (0 = one level under "category", to 3 = four levels under "category"), but not more. 

**Describe the solution you'd like**
It would be nice to be able to define n>3. 

To take on example with the classic "Places" category, one could use: 
_Places_|Continent|Country|State|City|Zipcode|Street Name|House Number

and then export in "Xmp.iptc.Location=6, 5, 4 3, 1" to have "House Number, Street Name, Zipcode City, Country"
(replace numbers above by the full `$(CATEGORYn(Places))` variable)

**Alternatives**
You can split your tags in two categories, like: 
_Places_|Continent|Country|State|City
_Address_|City|Zipcode|Street Name|House Number

I also did try $(CATEGORYn($(CATEGORYn(category)))) but it does not work! :)

EC1000:
As a side note, while redacting this message, I notice a small error on the manual. There is a additional `$` after `n` that sould not be there: 

> `$(CATEGORYn$(category))` : tag name of level n [0,3] of selected category (or tag)'
https://darktable.gitlab.io/doc/en/export_selected.html

phweyland:
> It would be nice to be able to define n>3.

Should be relatively easy to extend this to 9. :)

> I notice a small error on the manual. There is a additional `$` after `n` that should not be there:

Thanks for mentioning it.

I'll look at this.

EC1000:
That was quick! Thanks :)

PR #4605, opened 2020-04-04
$(CATEGORYn(category)) - accepts 10 levels instead of 4 [0-9]

phweyland:
$(CATEGORYn(category)) with n being 0 to 9 instead of 0 to 3.
Fixes also doc error as related in #4597

### DIFF
--- a/src/common/variables.c
+++ b/src/common/variables.c
@@ -392,10 +392,10 @@ static char *get_base_value(dt_variables_params_t *params, char **variable)
     result = g_strdup_printf("%d", params->data->max_height);
   else if (has_prefix(variable, "CATEGORY"))
   {
-    // TAG should be followed by n [0,3] and "(category)". category can contain 0 or more '|'
-    if (*variable[0] == '0' || *variable[0] == '1' || *variable[0] == '2' || *variable[0] == '3')
+    // CATEGORY should be followed by n [0,9] and "(category)". category can contain 0 or more '|'
+    if (g_ascii_isdigit(*variable[0]))
     {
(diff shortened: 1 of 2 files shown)


src/common/variables.c
```diff
@@ -392,10 +392,10 @@ static char *get_base_value(dt_variables_params_t *params, char **variable)
     result = g_strdup_printf("%d", params->data->max_height);
   else if (has_prefix(variable, "CATEGORY"))
   {
-    // TAG should be followed by n [0,3] and "(category)". category can contain 0 or more '|'
-    if (*variable[0] == '0' || *variable[0] == '1' || *variable[0] == '2' || *variable[0] == '3')
+    // CATEGORY should be followed by n [0,9] and "(category)". category can contain 0 or more '|'
+    if (g_ascii_isdigit(*variable[0]))
     {
-      const uint8_t level = (uint8_t)*variable[0] & 0b11;
+      const uint8_t level = (uint8_t)*variable[0] & 0b1111;
```
TurboGit, line 398: Probably better to use CLAMP here:
`const uint8_t level =  CLAMP((uint8_t)*variable[0], 0, 9);`

Otherwise we allow up to 15.

phweyland: > Otherwise we allow up to 15.

The previous test  ` if (g_ascii_isdigit(*variable[0]))`  makes sure we have an ascii digit
(30 to 39 hexa).

Moreover:
const uint8_t level = CLAMP((uint8_t)*variable[0], 0, 9);
gives always 9.

TurboGit: Ah right, good as-is then. Thanks.

PR #4608, opened 2020-04-04
shortcuts give crop and rotate focus

### LINKED ISSUE #3481: [Feature] Crop & rotate module : display grid when using key+mouse accel
Hi all !

In the Crop & rotate section of my shortcuts,, I have set a dynamic shortcut, so that I can correct my horizon by just pressing the key r and scrolling with my mouse !
It great, and I love those dual key+mouse accels (which configurability is missing in some module ;) )

I would really love that the grid defined in my Crop & rotate module is displayed when using this key+mouse accel ... it is not as of v3.0rc1 !

This way, I can correctmy horizon with a keypress + a mouse, without even having to care about the module itself !

Thank you

elstoc:
when using keyboard shortcuts to modify sliders and comboboxes on the
crop and rotate module, give gui focus to that module so that the
interactive tools and guide lines are shown.

Resolves #3481

### DIFF
--- a/src/iop/clipping.c
+++ b/src/iop/clipping.c
@@ -1645,6 +1645,7 @@ static void aspect_presets_changed(GtkWidget *combo, dt_iop_module_t *self)
     dt_conf_set_int("plugins/darkroom/clipping/ratio_n", p->ratio_n);
     if(self->dt->gui->reset) return;
     apply_box_aspect(self, GRAB_HORIZONTAL);
+    dt_iop_request_focus(self);
     dt_control_queue_redraw_center();
   }
 }
@@ -1655,6 +1656,7 @@ static void angle_callback(GtkWidget *slider, dt_iop_module_t *self)
   dt_iop_clipping_gui_data_t *g = (dt_iop_clipping_gui_data_t *)self->gui_data;
   dt_iop_clipping_params_t *p = (dt_iop_clipping_params_t *)self->params;
   p->angle = -dt_bauhaus_slider_get(slider);
+  dt_iop_request_focus(self);
   commit_box(self, g, p);
 }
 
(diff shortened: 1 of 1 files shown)


dterrahe:
I know, I know, this must sound annoying, but i actually like the fact that i can make small adjustments to rotation and margins _without_ having to jump into the module.

Because jumping into the module means I'll get zoomed all the way out to see the whole original and it makes it slightly less obvious whether an object on the last line at the border is visible or not. 

When I _do_ want to see guidelines and tools, I press c which is bound to _show module_.

Would you consider removing this change just for angle_/cxywh_callback?

elstoc:
@dterrahe _angle_ was the change specifically requested in the original issue. I'd like to see what others think. Ideally you'd be able to view guidelines without enabling everything else. But the crop&rotate is a funny module (because it includes the image-overlay elements and needs gui focus) and it's really hard to be consistent with it.

dterrahe:
I _know_! With this one it is impossible to make everyone happy. But I kindof think that it already _is_ possible to define a shortcut to _show_ the gui, but once these changes go through in my scenario I'd basically end up jumping in and out the gui continuously for each small fine-tuning change, because my option to _not_ show the gui has been taken away.

If you feel you have to, make the change for angle, but leave xywh, because the small fine tuning changes will in my opinion be much harder when the module has focus; that is why I added the sliders in the first place...

elstoc:
I think the argument for showing the guidelines when adjusting angle is a good one, because sometimes you need the gridlines to properly judge the angle.

I can't think of any good arguments why it's **required** for the xywh changes though. Again let's see if anyone else has any opinions. No hurry as this is a fairly low priority PR.

elstoc:
Closing in favour of #5041.

PR #4609, opened 2020-04-04
Updated Nikon Coolpix P1000 WB presets

Macchiato17:
Thanks to @johnny-bit and his current work on #johnny-bit:powerfull_extract_wb, this should now reflect all WB presets that are available with Nikon Coolpix P1000.


TurboGit:
Thanks !

Issue #4089, opened 2020-01-12
Default modules don't get always applied


progo:
**Describe the bug**

Even though I have selected to apply base curve and sharpening by default on new files, they can be missing from new images, see below.

**To Reproduce**

Session one -- Import a filmroll of virgin raw files. At this time all files will have got the default treatment: base curve, sharpening and so on. I can hit `W` to preview pics on the lighttable and get punchy previews (means that base curve has been applied as it should be).

Now, do subsequent sessions with other filmrolls. Go look at other filmrolls, close and open DT in between a few times. Now get back to the originally imported filmroll. Notice how base curve and sharpening are missing.

**Expected behavior**

The automatic defaults that were present during session #1 should remain applied.

**Platform (please complete the following information):**

 - Darktable Version: 3.0.0 built from gentoo portage
 - OS: gentoo linux
 - opencl 
 - geforce 1060 gtx + nvidia blob

**Additional context**

Two sidecars from the same picture. First one at 992 bytes is what darktable does when I import the directory the first time. The second one is when I opened the picture on darkroom. It is missing my defaults. 

The third one is a proper 5.7-kilobyte xmp that is the norm with my selected defaults. This xmp was created using the Duplicate manager / "create virgin duplicate" feature. 

[imported_once_not_opened.xmp.gz](https://github.com/darktable-org/darktable/files/4050502/imported_once_not_opened.xmp.gz)

[imported_once_opened_without_processing.xmp.gz](https://github.com/darktable-org/darktable/files/4050503/imported_once_opened_without_processing.xmp.gz)

[virgin_duplicate.xmp.gz](https://github.com/darktable-org/darktable/files/4050506/virgin_duplicate.xmp.gz)

progo:
As a good workaround,

* have experimented developing pics without the base curve. Usually brings better control to the process, especially highlight control. But this is more work per image for sure. 
* on lighttable/history module, discarded histories in mass. This brings the defaults back.

aurelienpierre:
Do you have a preset that disables the base curve automatically ?

progo:
No such things. No presets or styles that touch the base curve, automatically or otherwise.

aurelienpierre:
Ok, it seems you are not the only one. Probably something to do with pipe order management.

jenshannoschwalm:
Hi @progo i think this should be fixed in master right now. Could you test your problem again and either close this issue or come back with fresh details...

progo:
[...]

Issue #3891, opened 2019-12-27
base curve is applied when turned off after a fresh install


jade-nl:
**Describe the bug**
basecurve is auto applied after a fresh install and turning the auto-apply setting off.

Only happens after the first start. Does not happen again after the second start. The files that are opened in darkroom view during the first start still have the basecurve applied to them though.

**To Reproduce**
Steps to reproduce the behaviour:
1. remove xmp files from dir to be imported later
2. remove all files from ~/.config/darktable
3. start darktable
4. goto preferences -> core options: unset auto-apply basecurve and  close window
5. import a folder with images (or a single image)
6. open an image (any image) in darkroom view -> basecurve is applied
7. check preferences -> core options: auto-apply basecurve is not set
8. quit darktable and restart darktable -> problem is gone (except for those images you looked at in darkroom view the first time around, those have basecurve enabled).

**Expected behavior**
basecurve is not auto applied after disabling it.

**Platform (please complete the following information):**
 - OS: linux
- Distro: Debian 10.2 (buster)
 - Version: 3.1.0 (+192)
 - OpenCL: no

**Additional context**
Add any other context about the problem here, for example:
 - Can you reproduce with another dt version?
  Also happened yesterday with an earlier 3.1.0 version (+/- +80)
  This does not happen with the stable 3.0 (EDIT -> not 100% sure about this....)

jade-nl:
Some time has passed and because I wasn't 100% the last time about this happening in 3.0 stable I decided to re-do the checks.

**Stable, 3.0.0**
This issue _does_ happen in the current stable 3.0 version. After disabling the base-curve in the preferences and then importing images it is already clear to see in lighttable view. Opening an image in darkroom view confirms this: base curve is used.

**Development 3.1.0+390~gfa0e8ee01**
Nothing has changed in the latest development version compared to the previous one (+192), issue is still present.

Mark-64:
Confirmed in 3.1 git master as of today

Mark-64:
Can't reproduce.
Please be aware that changing auto-apply basecurve in DT core options requires a restart (it is explained in the tooltip)

jade-nl:
> (it is explained in the tooltip)

Sometimes you overlook the most obvious things....

Closing this issue.

phweyland:
I've exactly the symmetric problem.
When enabling auto-apply, I don't get any on next import

PR #4613, opened 2020-04-05
Write correct sidecar including history after applying presets

jenshannoschwalm:
Whenever we apply any presets on a 'fresh' image we want a **correct** sidecar file to be written.

As there is no history in  `_dev_auto_apply_presets` we must write the sidecar  later in
`dt_dev_read_history_ext`

This is related to #4586 (fixing it) #4595 (first attempt but not fully understanding the problem) and maybe #4089 

In the discussion #4595 i still got wrong the underlying problem. How to reproduce is pretty simple:
Use a fresh image and open it in darkroom, you will see presets applied in history stack. But look at the sidecar right now, it has
1. darktable:auto_presets_applied="1"
2. but is has **no** history data

This is simply wrong. We normally don't observe problems because reimporting such an image is unusual but in that case the sidecar is wrong and will prohibit applying the presets.

Anyway -- this might be a significant problem if we use img plus xmp as backup and reimport any time later.

### DIFF
--- a/src/common/exif.cc
+++ b/src/common/exif.cc
@@ -2701,19 +2701,17 @@ int dt_exif_xmp_read(dt_image_t *img, const char *filename, const int history_on
       img->legacy_flip.legacy = 0;
     }
 
+    int32_t preset_applied = 0;
+    
     if((pos = xmpData.findKey(Exiv2::XmpKey("Xmp.darktable.auto_presets_applied"))) != xmpData.end())
     {
-      const int32_t i = pos->toLong();
-      // set or clear bit in image struct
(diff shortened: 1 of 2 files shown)


jenshannoschwalm:
Commit https://github.com/darktable-org/darktable/pull/4613/commits/71b81cc1ada8aba61b95665df51d99116b035096 adds a check while reading exif data so xmp files having the problem fixed by this pr are detected and thus get presets applied.

@parafin @TurboGit You agree on this?

jenshannoschwalm:
Other issues that might be related #4137 #3891 (@phweyland about your symmetrical problem)

TurboGit:
I think this is definitely going into the right direction. I will merge.

Issue #4610, opened 2020-04-04
Drawn mask outline not visible outside of image area


kofa73:
1. Load any raw (at least, this happens with any NEF from my D7000)
2. Set view size to small (in my case: 12%)
3. Open any module that supports a drawn mask. Draw a path such that some points of the curve are outside the image area.

=> Problem: you cannot see the points you draw outside, neither the curve outside the image area.
[screenshot after the image is drawn](https://photos.smugmug.com/photos/i-RtKxNMB/1/8d007b8c/O/i-RtKxNMB.png)

4. Complete the path.
5. Zoom into the image (fit to screen or tighter).
6. Zoom back to 'small'

=> While darktable calculates the image to be displayed, the curve outside the image is visible, as with previous versions of the software.
[screenshot while darktable is calculating the image](https://photos.smugmug.com/photos/i-xM8FHPP/1/37e7dbcb/O/i-xM8FHPP.png)
=> Once the calculation is done, the curve outside the image disappears (looks the same as after it was drawn, see 'screenshot after the image is drawn'

I've tried with all the built-in themes, the behaviour is the same.
Happens both with OpenCL on and off.

Self-compiled from de8987c8f1; Ubuntu 19.10, 64-bit
```
ii  gcc                                                         4:9.2.1-3.1ubuntu1                          amd64        GNU C compiler
ii  gcc-8                                                       8.4.0-1ubuntu1~19.10                        amd64        GNU C compiler
ii  gcc-8-base:amd64                                            8.4.0-1ubuntu1~19.10                        amd64        GCC, the GNU Compiler Collection (base package)
ii  gcc-9                                                       9.2.1-9ubuntu2                              amd64        GNU C compiler
ii  gcc-9-base:amd64                                            9.2.1-9ubuntu2                              amd64        GCC, the GNU Compiler Collection (base package)
ii  gcc-9-base:i386                                             9.2.1-9ubuntu2                              i386         GCC, the GNU Compiler Collection (base package)
ii  clang                                                       1:9.0-49~exp1                               amd64        C, C++ and Objective-C compiler (LLVM based)
ii  clang-8                                                     1:8.0.1-3build1                             amd64        C, C++ and Objective-C compiler
ii  clang-9                                                     1:9-2                                       amd64        C, C++ and Objective-C compiler
```

elstoc:
Reproduced here on 07f870a, compiled on ArchLinux.

jenshannoschwalm:
[...]

PR #4616, opened 2020-04-05
Allow masks to be displayed in the border area (fixing #4610)

jenshannoschwalm:
A while ago the masks display was restricted to the true image area, this leads to
restrictions while modifying the masks so we should keep showing them :-(


TurboGit:
Makes sense indeed! Thanks.

PR #4622, opened 2020-04-05
Expand the "color look up table" tool

Lorenzoeffe:
Expand the "color look up table" tool
the change should allow you to view the colors on a wheel, after selecting them and adjusting the width of the area affected by the change via slide.
Slides for changing hue, intensity, saturation, brightness, contrast should then be introduced.


LebedevRI:
This needs to be rebased,  most of the changes shouldn't be here.

LebedevRI:
Also, this is wrong branch.

Lorenzoeffe:
> Also, this is wrong branch.
In what sense should it be reformulated?
What is the right place for these requests?
Thanks

PR #4625, opened 2020-04-06
overlay colors: multi-colored icon

elstoc:
change the bottom-panel icon for overlay line colors so it's more colorful.

### DIFF
--- a/src/dtgtk/paint.c
+++ b/src/dtgtk/paint.c
@@ -1020,16 +1020,24 @@ void dtgtk_cairo_paint_grid(cairo_t *cr, gint x, gint y, gint w, gint h, gint fl
   cairo_set_line_width(cr, 0.1);
   cairo_set_line_cap(cr, CAIRO_LINE_CAP_ROUND);
   
+  cairo_set_source_rgb(cr, 1.0, 0.0, 0.0);
   cairo_move_to(cr, 0.0, 0.3);
   cairo_line_to(cr, 1.0, 0.3);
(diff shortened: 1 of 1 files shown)


elstoc:
@TurboGit as requested. I've tried a few color combinations and I'm not 100% happy with the aesthetics of any of them. This is the least worst!

TurboGit:
Agreed a bit flashy, maybe with less opacity to have less aggressive colors like the RAW overexposed icon?

elstoc:
It's also difficult finding four different colors that work well together without seeming lop-sided. Perhaps just two colors: one for the vertical lines and one for the horizontal. Red and green maybe?

elstoc:
Another try. Why is this so hard?!

TurboGit:
Seems like you failed pushing something :) Anyway, I tried it and put a 0.5 opacity, the colors are visible but less aggressive on the UI. Let's see what others think about this.

elstoc:
I did. I forgot to add before I amended! I had just red and green:
Screenshot_2020-04-06_17-32-27

TurboGit:
> I did. I forgot to add before I amended!

That's what I thought :) What do you think about the current version ?

elstoc:
I edited my comment: it looks lop-sided because of the relative brightness of the colors. At first glance I only see the yellow and green lines.

TurboGit:
Ok, can you do a PR with 2 colors ? If it does not look right we can just revert and go back to grey grid. Probably the best option...

PR #4626, opened 2020-04-06
paint: use red/green only for the grid icon

elstoc:


TurboGit:
Ok, better. I have merged that and fixed the symmetry of the lines.

PR #4627, opened 2020-04-06
guides: make grid use selected overlay color

elstoc:
in the crop & rotate module, the 'grid' guides were gray, regardless of the selected color.


TurboGit:
Indeed, missed this one! Thanks for fixing.

Issue #4607, opened 2020-04-04
DT3.0.1: collection selection does not work after a while


andreilabato:
Hello, after several searches using collection filters (using tags...) result does not fit anymore:does not correspond to selected search criteria=>I get get all pictures, no pictures.... Difficult to reproduce but happens quite often when I use entensively different filters after several search. 
When problem happens:
-Exit from DT and open it back again: same problem.
-Reinstall of DT and open it back again: same problem.
-To solve I empty collection filter field and then reset the module. Then it works back again.
I am not the only one to have seen it in Windows (info from french forum of DT). What could I do to help solve the issue?
Thanks!
I use Windows 10
I think problem existed already in DT 2.6 and 3.0 but I am not sure.

phweyland:
I can reproduce some related issues.
I'll look at this.

andreilabato:
https://forums.darktable.fr/showthread.php?tid=4712

PR #4631, opened 2020-04-06
collect fix

phweyland:
- set/reset typing when necessary
- remove 'intert-text' signal

The first item improves the filtering based on entry. Should help on #4607 
The second item is just a clean up and should not impact anything. "insert-text" is to be used to control what is inserted. Useless here and always followed by "changed" event, which is enough.


phweyland:
One remark: moving the mouse over thumbnails makes the collect entry box loose the focus, producing unexpected effects when typing ...
1. click on entry (or somehwere inside the text in entry)
2. release the mouse somewhere above thumbnails
3. type... not cool

TurboGit:
Ok, looks good to me and less code = less bug :) Thanks!

PR #4636, opened 2020-04-07
Update Sony ILCE-7 WB presets

jakubfi:
Encouraged by @johnny-bit I've updated the Sony ILCE-7 WB presets for the latest camera firmware version. Also WB finetuning is now available, as WB has been calibrated over the full amber-blue range for each factory preset.


TurboGit:
Thanks !

PR #4639, opened 2020-04-07
WB Presets for Samsung NX1 firmware ver. 1.41

johnny-bit:
Sample set contributed by torebu (Tore Busch) via pixls.us


TurboGit:
Thanks !

PR #4644, opened 2020-04-07
lighttable: configurable border size

### LINKED ISSUE #4643: Feature Request: Preference for lighttable border size
It's currently possible to control the size of the border around the images in the darktable via a preference setting. Please could we have something similar for the lighttable.

elstoc:
make the border in the lighttable view (file manager and zoomable lighttable)
configurable - small/medium/large - via a preference, similar to
the darkroom view

Resolves #4643

I've restricted it to just the file manager and zoomable lighttable for this commit. Not sure whether it should affect culling mode/preview as well. IMO they might be better left alone.

### DIFF
--- a/data/darktableconfig.xml.in
+++ b/data/darktableconfig.xml.in
@@ -1392,6 +1392,19 @@
     <shortdescription>rating an image one star twice will not zero out the rating</shortdescription>
     <longdescription>do not have the rating of one star behave as documented in the manual--an image rated one star twice will result in a zero star rating.</longdescription>
   </dtconfig>
+  <dtconfig prefs="gui" section="lighttable">
+    <name>plugins/lighttable/ui/border_size</name>
+    <type>
+      <enum>
+        <option>small</option>
+        <option>normal</option>
+        <option>large</option>
+      </enum>
+    </type>
+    <default>normal</default>
+    <shortdescription>border around an image in lighttable mode</shortdescription>
+    <longdescription>defines the size of the gray border around the image in lighttable mode (filemanager and zoomable lighttable). Requires restart.</longdescription>
+  </dtconfig>
   <dtconfig>
     <name>darkroom/ui/rawoverexposed/mode</name>
     <type>int</type>
--- a/src/dtgtk/thumbnail.c
+++ b/src/dtgtk/thumbnail.c
@@ -115,6 +115,14 @@ static gboolean _event_image_draw(GtkWidget *widget, cairo_t *cr, gpointer user_
 {
   if(!user_data) return TRUE;
   dt_thumbnail_t *thumb = (dt_thumbnail_t *)user_data;
+
+  float imgsize = 0.0;
+  const gchar *bordersize = dt_conf_get_string("plugins/lighttable/ui/border_size");
+
+  if (strcmp(bordersize, "small") == 0) imgsize = 0.95;
+  else if (strcmp(bordersize, "large") == 0) imgsize = 0.87;
+  else imgsize = 0.91;
+
   if(thumb->imgid <= 0)
   {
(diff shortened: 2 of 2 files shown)


AlicVB:
As said in #4643 I agree with @Nilvus that it would better fit in css if possible (to be tested)



> I've restricted it to just the file manager and zoomable lighttable for this commit. Not sure whether it should affect culling mode/preview as well. IMO they might be better left alone.

Not sure : if we expose this setting, there's no reason to not expose the same thing for culling and preview. And I'm not sure that adding 2 quite similar params in the already cluttered pref dialog is the good thing to do...

elstoc:
I take your point. Happy for this to be moved to CSS if you can give me some pointers, or I'm ok to close the PR if you want to work on it.

ptilopteri:
Most but still not all users understand changing options using the "configuration" wheel, but not editing CSS.  I have not a preference for the method of configuration, but editing CSS is not in everyone's bailiwick.  The CSS code should be heavily commented to *what* and *where* particular options may be changed and what changes are available.

Nilvus:
> Most but still not all users understand changing options using the "configuration" wheel, but not editing CSS. I have not a preference for the method of configuration, but editing CSS is not in everyone's bailiwick. The CSS code should be heavily commented to _what_ and _where_ particular options may be changed and what changes are available.

The goal is first to remain consistent, if adding this on preferences, why not adding other things actually on CSS. And which ones? How avoid having a too overload preferences options. I regularly follow pixl.us and darktable.fr forums and issues here, and I often see people asking for adding preferences that already exist on preferences. In the other side, I rarely, very rarely, see people asking how to tweak such parts of the UI. So that's not so simple as just adding a new preference.

Yes, tweaking CSS should probably be made easier but just consider that probably most of the users don't care of that. The only thing I will see on preferences (quite often ask) would be font choosed and font size.

And CSS code is already quite heavily commented, just check it. Of course, probably it could be improved like all things!

phweyland:
I support 100% the CSS solution.
A huge effort has been made recently to remove layout dependencies from the code and there is probably still a lot of work to be done.
CSS is the place to tweak the dt appearance. 
If code changes for appearance have to be made that should be only to allow better CSS adjustments.
Then interested people would be able to propose improved or more elaborated themes for dt.

TurboGit:
Yes, everything that is to control GUI rendering must be done in CSS when possible.

AlicVB:
Ok, I've done a few tests and it seems that we face 2 issues : 
- percentages doesn't seems to be admit by gtk for margin values
- margin/padding doesn't seems to work with gtkoverlay (which is understandable as gtkoverlay has it's own positioning mechanisms)

Now I'm far from a gtk expert, so if someone manage to solve the second issue at least, it would be great !
Otherwise, the best I can think of is : 
- adding margins in pixels in css
- retrieving this value at the thumbnail creation and use it instead of the hardcoded ones. (like in `dtgtk/thumbnail_btn.c:49`)

This will of course require some comment and warning in the css file. But margin will not be tweakable in css !

What do you think ?

Nilvus:
@AlicVB: so margins in pixels in CSS seems the better way for me. Just needs to be test (solution and tweaking CSS to see).

elstoc:
Will that mean that the margins are the same regardless of the number of images per row in the lighttable?

AlicVB:
> Will that mean that the margins are the same regardless of the number of images per row in the lighttable?

No, it has to be dependent of the thumb size, as it was for instance. Otherwise, we'll be forced to set very very low margin to enable the view of very small thumbs... and that would be strange for bigger images sizes...
I'll try to come with something asap...

AlicVB:
to all : I've just proposed a new PR to do that here : #4660 
@elstoc : I hope you'll excuse me for having done the work you intended to do...

elstoc:
@AlicVB not at all. It's the result that counts and you're diving into areas I don't fully understand anyway. I'll close this one.

PR #4647, opened 2020-04-08
show lib module shortcut: respect lighttable expand pref

elstoc:
when showing/hiding lib modules using a keyboard shortcut, respect the
lighttable 'expand single module' preference when set


TurboGit:
Looks ok to me. Thanks.

Issue #4624, opened 2020-04-06
lt edited status icon not refreshed when discarding history


TurboGit:
**Describe the bug**

The edited icon is not properly refreshed when discarding history.

**To Reproduce**

- on one image, discard history
- got to darkrroom
- activate a module, say graduated filter
- move back to lighttable

All ok, the thumbs has the modified/edited icon activated.

- discard history again, the thumbs is still displaying the edited icon. 

**Expected behavior**

The edited icon should be removed after discarding history.

**Platform (please complete the following information):**
 - Darktable Version: 
 - OS: Linux

**Additional context**

This is only a refresh issue, if you scroll the lighttable to move the thumbs out of the current view and scroll back to display it, the edited icon is not displayed anymore.

phweyland:
I'll look at this.

johnny-bit:
Well - this seems to work flawlessly on 3.4.0 and current master :) Let's close this, ok?

PR #4648, opened 2020-04-08
fix history hash on discard, copy and compress history actions

phweyland:
There is still a questionable case however: in case of history copy of a virgin image over another virgin image with `append` option, the altered status is set.
I haven't tried to fix this. Let me know.

Plus some formatting alignments.


phweyland:
Related to #4624

TurboGit:
Looks ok. Thanks for looking at this.

PR #4649, opened 2020-04-08
fix $(EXIF_EXPOSURE) issue on watermark

### LINKED ISSUE #4646: $(EXIF_EXPOSURE) wrong value in Watermark Module
Same pattern show diferent values for exposure as shown in the screenshot below:

image

Darktable 3.0.1 (latest release)
Windows 10

ps: I had to put random characters before the pattern or else darktable would crash, I've already reported this issue.

phweyland:
Resolves #4646 $(EXIF_EXPOSURE) part.
I cannot reproduce the second related issue (random characters...).

### DIFF
--- a/src/iop/watermark.c
+++ b/src/iop/watermark.c
@@ -868,7 +868,7 @@ static gchar *_watermark_get_svgdoc(dt_iop_module_t *self, dt_iop_watermark_data
   dt_variables_params_t *params;
   dt_variables_params_init(&params);
   params->filename = image->filename;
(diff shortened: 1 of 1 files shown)


TurboGit:
Ok, makes sense. It still looks like a bit fragile that we rely on a special string for jobcode for this to work.

phweyland:
In fact I'd myself set "watermark" as jobcode here when I added variables expansion in watermark. At this time I hadn't seen the impact. At least I didn't take any risk for that fix. :)

PR #4652, opened 2020-04-08
Finetuned WB presets for various Panasonic bodies

johnny-bit:
samples provided by thokster over at pixls.us


johnny-bit:
This should work without a hitch and errors aren't related to wb prests at all...

johnny-bit:
NOW it should work :)

TurboGit:
Thanks !

Issue #4578, opened 2020-03-31
Add manual Vignetting correction


s7habo:
In darktable vignetting correction is automatically applied when lens is supported with Lensfun.
For lenses that are not supported by Lensfun, there is no reasonable way to manually correct vignetting in darktable. 

Also with some supported lenses, the vignetting correction is not (yet) integrated or not always optimal and it would be good to have the possibility to manually improve the correction afterwards.

Darktable has a vignetting module, but to make vignettes. The module is not suitable for correction.

Vignetting itself is not always annoying, but for some scenes it is too noticeable. Especially when you increase the contrast of the image, the darker border becomes stronger, which distracts too much from the subject, and unnecessarily increases the dynamic range of the image, which leads to additional processing steps.

chrik5:
> For lenses that are not supported by Lensfun, there is no reasonable way to manually correct vignetting in darktable.

Using the exposure module with a drawn mask might be a manual way to correct vignetting.

s7habo:
> Using the exposure module with a drawn mask might be a manual way to correct vignetting.

No. The feathering on drawn mask does not work correctly, so the vignette can be corrected. There is always a ring (haloing) visible.

elstoc:
Perhaps it could be done in the lens correction module as suggested in #1990

jenshannoschwalm:
Hi @s7habo the vignetting module has been modified slightly. It's not perfect but what do think?

s7habo:
Thanks @jenshannoschwalm!  This is definitely a big improvement.

One suggestion: in case the scale value is very low,  fall-off strength should also be in the range 0-200%. 

Here is an example:

Original:

![Vignette1](https://user-images.githubusercontent.com/38165484/79038368-03cb5b80-7bd9-11ea-807e-4b6f5d7af963.png)


With low scale:

![Vignette2](https://user-images.githubusercontent.com/38165484/79038413-4db44180-7bd9-11ea-91fc-310dd9536aa9.png)


The second problem with the module is that the lightening is not done correctly (you can clearly see this by the branches at the bottom left. They lose contrast and become grayish). I suppose this is intentional if you want to have a white vignette. 
Maybe the module should have an additional "correction mode" where the exposure is physically correct. 

Here is the same example with exposure module. Here I used the circle as a drawn mask: 

![Vignette3](https://user-images.githubusercontent.com/38165484/79038656-4aba5080-7bdb-11ea-9bdc-9fd795f7279f.png)

[...]

PR #4654, opened 2020-04-09
Allowing larger vignette scales

jenshannoschwalm:
The vignette module is mostly used for artistic reasons, the area to be used is
restricted to the image.

Some people have images shot without proper vignette correction via the lens module,
one example #4578. The vignetting module is not perfect for manual correction, but it's
not so bad at all. The main UI restriction seems to be a too small area that can be selected.

This pr simply allows the scale to be selected in the 0-200% range instead of 0-100. I tested
on many images and it's not perfect but much more suitable than before.


TurboGit:
Works for me. Thanks!

PR #4656, opened 2020-04-09
Finetuned WB presets for various olympus bodies

johnny-bit:
Samples contributed by user wiegemalt over at pixls.us


TurboGit:
Thanks!

PR #4657, opened 2020-04-09
fix lut 3d cube files

phweyland:
- consider out of range [0,1] values (warning) and accept negative values
- detect nan
- display number of lines when incorrect

EDIT Issue commented [here](https://discuss.pixls.us/t/cube-lut-lines-number-not-correct-error/17413/10) on pixls.us


TurboGit:
Works for me. Merging on 3.0.2 too. Thanks.

PR #4660, opened 2020-04-09
set thumbnails margins via css

### LINKED ISSUE #4643: Feature Request: Preference for lighttable border size

AlicVB:
fix #4643 

This allow images margins for filemanager, zoomable lighttable and filmstrip to be set via css.
As Gtkoverflow doesn't support natively margins and gtk doesn't support percentage values for css margins we do that by hand...

### DIFF
--- a/data/themes/darktable.css
+++ b/data/themes/darktable.css
@@ -1530,6 +1530,7 @@ Details :
 #thumb_image
 {
   border: none;
+  margin : 5px; /* not real pixels, but percentage of the thumb size */
 }
 #thumb_main:active #thumb_image
 {
--- a/data/themes/darktable.css
+++ b/data/themes/darktable.css
@@ -1530,7 +1530,7 @@ Details :
 #thumb_image
 {
   border: none;
-  margin : 5px; /* not real pixels, but percentage of the thumb size */
+  margin : 4px; /* not real pixels, but percentage of the thumb size */
 }
 #thumb_main:active #thumb_image
 {
--- a/src/dtgtk/thumbnail.c
+++ b/src/dtgtk/thumbnail.c
@@ -141,6 +141,17 @@ static gboolean _event_image_draw(GtkWidget *widget, cairo_t *cr, gpointer user_
   // if we don't have it in memory, we want the image surface
   if(!thumb->img_surf || thumb->img_surf_dirty)
   {
+    if(!thumb->img_margin)
+    {
+      // we retrieve image margins from css
+      GtkStateFlags state = gtk_widget_get_state_flags(thumb->w_image);
+      thumb->img_margin = gtk_border_new();
+      GtkStyleContext *context = gtk_widget_get_style_context(thumb->w_image);
+      gtk_style_context_get_margin(context, state, thumb->img_margin);
+    }
+    const float ratio_h = (float)(100 - thumb->img_margin->top - thumb->img_margin->bottom) / 100.0;
+    const float ratio_w = (float)(100 - thumb->img_margin->left - thumb->img_margin->right) / 100.0;
+
     if(v->view(v) == DT_VIEW_DARKROOM && dev->preview_pipe->output_imgid == thumb->imgid
        && dev->preview_pipe->output_backbuf)
     {
@@ -168,7 +179,7 @@ static gboolean _event_image_draw(GtkWidget *widget, cairo_t *cr, gpointer user_
       if(tmp_surface)
       {
         const float scale
-            = fminf(thumb->width * 0.91 / (float)buf_width, thumb->height * 0.91 / (float)buf_height);
+            = fminf(thumb->width * ratio_w / (float)buf_width, thumb->height * ratio_h / (float)buf_height);
         const int img_width = buf_width * scale;
         const int img_height = buf_height * scale;
         thumb->img_surf = cairo_image_surface_create(CAIRO_FORMAT_RGB24, img_width, img_height);
@@ -199,8 +210,8 @@ static gboolean _event_image_draw(GtkWidget *widget, cairo_t *cr, gpointer user_
     }
     else
     {
-      const gboolean res
-          = dt_view_image_get_surface(thumb->imgid, thumb->width * 0.91, thumb->height * 0.91, &thumb->img_surf);
+      const gboolean res = dt_view_image_get_surface(thumb->imgid, thumb->width * ratio_w, thumb->height * ratio_h,
+                                                     &thumb->img_surf);
       if(res)
       {
         // if the image is missing, we reload it again
@@ -778,6 +789,7 @@ void dt_thumbnail_destroy(dt_thumbnail_t *thumb)
   thumb->img_surf = NULL;
   if(thumb->w_main) gtk_widget_destroy(thumb->w_main);
   if(thumb->filename) g_free(thumb->filename);
+  if(thumb->img_margin) gtk_border_free(thumb->img_margin);
   free(thumb);
 }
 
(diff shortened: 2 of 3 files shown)


elstoc:
This works well for me. However, if I create a darktable.css in my config/themes directory with just that section of the css in it, the borders don't change. Am I doing it wrong?

AlicVB:
maybe there's a simpler way, but what I do is copying the full darktable.css from my share dir, rename it and tweak things inside. Then you'll have to set your theme to the new name you have choose.
@Nilvus : I'm right ?

elstoc:
If that's the case I think I feel another feature request coming on. Ideally what you want is to be able to use a core theme and then just tweak the things you want to tweak locally - i.e. I want to use the main darktable theme but then I have a single css file locally (with the same name) that just alters the border value.

That way if a new version of the main css file comes along, I get the advantage of those changes, without losing my local tweaks.

AlicVB:
wait, you should be able to do that by adding `@import url("full/path/of/the/base/theme.css");` at the beginning of the file and then add just what you tweak.
Not tested but that should work...

elstoc:
That does work, thanks. 

IMO it would be nicer to be able to do as I've suggested above but that might not be so easy, since gtk_css_provider_load_from_path doesn't appear to support loading multiple CSS files (which would be the simple way to achieve my suggestion). According to the docs  it clears any previously loaded information when called.

So you'd have to find a way to combine them in the code.

Anyway that's a separate topic - I'm happy with this PR - seems to do what I asked for.

Nilvus:
> wait, you should be able to do that by adding `@import url("full/path/of/the/base/theme.css");` at the beginning of the file and then add just what you tweak.
> Not tested but that should work...

@AlicVB and @elstoc: that's was I was describing on the issue. The bad way is to copy the whole CSS. The good way is to use this @import url line (like you can see in all other default css files except darktable.css ones. All depends of this darktable.css main themes by importing this file (or for some by importing a related theme that will then import recursively darktable.css.

That way, we just have to add all new things on darktable.css file only and just adjust those needed regardless of other themes (like adjusting colors for example to become consistent with the grey theme or dark one, etc.).

So, yes, in personal config folder, you just have to add that line with the full path, then only add lines you want to tweak. No need to have all other ones. Hope it's clear!

@elstoc: said that I don't understand what you said about a core theme. The core theme already exist as explained here: it's darktable.css file. And if you want to tweak for example the grey theme, you only have to change the import line, then again copy lines you want to tweak, lines from grey css theme or any other theme, like darktable.css one that grey theme import.

elstoc:
> I don't understand what you said about a core theme. 

Perhaps base theme might have been the better term. What I meant was that, rather than having to put the ``@import`` line at the top of the user-defined css, you could change ``dt_gui_load_theme`` so that if I put a file named, say "darktable-elegant-grey.css" in my config theme directory, and my theme preference is also set to "darktable-elegant-grey", darktable would load the file from the normal share location and then _also_ load the version in the config directory (essentially load both files and concatenate them together).

It's not a massive deal since, as I say, the ``@import`` line works ok. It's just one less thing you need to know in order to tweak themes, and it seems a fairly intuitive way to work.

Either way all this definitely does all need some documentation in the manual if the intention is to move all of the 'look-and-feel' stuff to CSS.

parafin:
Putting in full path is a PITA, since it's varies depending on the system. Customizing CSS theme should be simplified as @elstoc describes and documented somewhere with some simple examples like changing some color.

elstoc:
I agree. Personally I'm currently running three different versions of darktable each with a different config and share paths, so it's even worse.

Nilvus:
> Putting in full path is a PITA, since it's varies depending on the system. Customizing CSS theme should be simplified as @elstoc describes and documented somewhere with some simple examples like changing some color.

I don't say putting full path in default themes. That would be a non-sense. I talk about .config folder, so user theme. And users themes are what they are: personal ones so full path is adjusted by the user.

Nilvus:
@elstoc: for you last detailed post here, I agree to your proposal but that would need some changes on code. Let see if someone could do that (I only have skills on working on CSS here).

About manual, no time by now (or to be more precise, I have personal priorities on using my time by now). Anyway, it would be better to do when (if) more work is done here about your proposals.

elstoc:
I might be able to do the coding, perhaps with a bit of help. It doesn't look too hard (I think).

On 9 April 2020 22:19:51 BST, Nicolas Auffray <notifications@github.com> wrote:
>@elstoc: for you last detailed post here, I agree to your proposal but
>that would need some changes on code. Let see if someone could do that
>(I only have skills on working on CSS here).
>
>About manual, no time by now (or to be more precise, I have personal
>priorities on using my time by now). Anyway, it would be better to do
>when (if) more work is done here about your proposals.
>
>-- 
>You are receiving this because you were mentioned.
>Reply to this email directly or view it on GitHub:
>https://github.com/darktable-org/darktable/pull/4660#issuecomment-611757895

-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

Nilvus:
> I might be able to do the coding, perhaps with a bit of help. It doesn't look too hard (I think).
> […](#)
> On 9 April 2020 22:19:51 BST, Nicolas Auffray ***@***.***> wrote: @elstoc: for you last detailed post here, I agree to your proposal but that would need some changes on code. Let see if someone could do that (I only have skills on working on CSS here). About manual, no time by now (or to be more precise, I have personal priorities on using my time by now). Anyway, it would be better to do when (if) more work is done here about your proposals. -- You are receiving this because you were mentioned. Reply to this email directly or view it on GitHub: [#4660 (comment)](https://github.com/darktable-org/darktable/pull/4660#issuecomment-611757895)
> -- Sent from my Android device with K-9 Mail. Please excuse my brevity.

Maybe this also need to be discussed on @AlicVB idea (proposed on your issue) to add a CSS edit tab on preferences?

elstoc:
I'm not sure how those can be reconciled. I'll raise a new FR and we can move discussion there.

TurboGit:
I tried this and cannot see a difference. I have set the border to 20px, the thumbs are still looking the same... Am I missing something ?

AlicVB:
@TurboGit : strange... some questions to be sure we speak about the same thing :
- Do you set margin to 20px or border ?
- Did you have restart dt ?
- where have you changed the margin value ? in main theme, in another one, in a theme in your .config folder ?

TurboGit:
I have changed "margin: 4px" in darktable.css. And yes sure I restarted dt each time.

AlicVB:
@TurboGit : 
arghh... I've just re-tested many times and it seems to work correctly here, even if you change the value in darktable.css and use a child theme...

Just to be sure : 
- we speak about the margin around the images in filemanager thumbnail, right ?
- Your css change is something like that : `
#thumb_image
{
  border: none;
  margin : 20px; /* not real pixels, but percentage of the thumb size */
}
`
- you don't have a theme in your .config dir that overwrite the original theme ?

Another question, to be sure the theme is loaded : What happen if you change darktable.css with : 
```
#thumb_image
{
  border: 10px solid red;
  margin : 4px; /* not real pixels, but percentage of the thumb size */
}
```
Do you see big ugly borders around images

And last debugging : 
can you add some printf to show ratio_h and ratio_w values after : https://github.com/darktable-org/darktable/blob/0b56ebdfd4ada0ed909dc1bf5c3b0510aa30a903/src/dtgtk/thumbnail.c#L153

TurboGit:
Ok, procedural error on my side :( Sorry this works fine. Thanks.

elstoc:
Thanks @AlicVB. Much appreciated

AlicVB:
@TurboGit : thanks... I was afraid of another gtk version mess ;)

PR #4664, opened 2020-04-10
Keep white balance as spot when set via the picker (Fix #3521)

### LINKED ISSUE #3521: White balance switches from 'spot' to 'camera'
**Describe the bug**
After choosing the white balance preset 'spot' and applying a new white balance with the area picker correctly says 'preset: spot'.
When using the color picker on the left side or switching between 'point' and 'area' the preset on the right side in the white balance module says 'preset: camera' with the white balance chosen with spot

**Expected behavior**
'preset' in the white balance module should stay at 'spot' unless chosen otherwise.

**Screenshots**
1. Start with original white balance from camera
Bildschirmfoto von 2019-11-22 09-03-50
2. Manually chosen white balance 
Bildschirmfoto von 2019-11-22 09-04-03
3. White balance says 'preset: camera' after changing the color picker from 'area' to 'spot' (clicking on the picker next to it does the same)
Bildschirmfoto von 2019-11-22 09-04-12

DT 3.0.0rc1~git11.dc641cdc9 on Ubuntu 18.04

**Additional context**
Similar to https://github.com/darktable-org/darktable/issues/3228

jenshannoschwalm:
This has been outstanding for a while. What is the problem?

in `temperature.c` we use `gui_update`. We check the **current** coeffs against
all found presets but forgot about a 'not found' situation? That must be a spot :-)

### DIFF
--- a/src/iop/temperature.c
+++ b/src/iop/temperature.c
@@ -901,6 +901,10 @@ void gui_update(struct dt_iop_module_t *self)
         }
       }
     }
+
+    // Still not found after testing all sorts of presets? Then it must be a spot.
+    if(!found)
+      dt_bauhaus_combobox_set(g->presets, 2);
   }
 }
 


parafin:
No, it mustn't. You can also just move sliders around.

jenshannoschwalm:
> No, it mustn't
 
Oh man, myself being so blind... needs more and better thinking here ...

johnny-bit:
There's way more nuggets there that can get lost ;)

Issue #4611, opened 2020-04-04
Make color picker color values copyable as Hex values.


borissmidt:
**Is your feature request related to a problem? Please describe.**
You can easily select a color using the color picker. But then it gives you RGB values as 0-255 values. While most tools like css prefer the  hex format like `#RRGGBB`.  

Currently i have to manually calculate the code to a hex code and then type it over.
The value should be copy able for easy usage. Then you could also add the value to a different image so you can compare the values. 

Maybe other color spaces might be useful like HSL. 

**Describe the solution you'd like**
Add a right mouse action to copy the color as HEX code. 
Add an option to show RGB in the hex format '#FFBE82' like.
Add an option to paste an RGB color into the color picker to compare.

(sorry for the imperative tense i'm too used to the git commit message structure).

By the way i really like darktable 3.0.1 it seems a bit faster and more stable.

dterrahe:
@borissmidt if you are satisfied with the partial solution implemented in #4665 please close this issue. Thanks!

borissmidt:
Hey yes it is good, Thank you for the effort! hex codes are so much easier.

PR #4665, opened 2020-04-10
Add hex encoded rgb to colorpickers tooltips

jenshannoschwalm:
Inspired by #4611 this does not implement the whole "wish list" but it adds
the rgb hex value to the tooltip.

### DIFF
--- a/src/libs/colorpicker.c
+++ b/src/libs/colorpicker.c
@@ -175,13 +175,13 @@ static void _update_picker_output(dt_lib_module_t *self)
     switch(input_color)
     {
       case 0: // rgb
-        snprintf(tooltip, sizeof(tooltip), "%3d   %3d   %3d   %02X%02X%02X",
+        snprintf(tooltip, sizeof(tooltip), "%3d   %3d   %3d   (0x%02X%02X%02X)",
                  (int)round(rgb[0] * 255.f),
                  (int)round(rgb[1] * 255.f),
                  (int)round(rgb[2] * 255.f),
-                 (int)round(rgb[0] * 255.f),
-                 (int)round(rgb[1] * 255.f),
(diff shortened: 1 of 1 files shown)


src/libs/colorpicker.c
```diff
@@ -175,7 +175,10 @@ static void _update_picker_output(dt_lib_module_t *self)
     switch(input_color)
     {
       case 0: // rgb
-        snprintf(tooltip, sizeof(tooltip), "%3d   %3d   %3d",
+        snprintf(tooltip, sizeof(tooltip), "%3d   %3d   %3d   %02X%02X%02X",
```
TurboGit, line 178: Can we add some kind of separator here ? Maybe "%3d   %3d   %3d  - %02X%02X%02X" or using | ?

Or maybe "hex: " to clearly indicate that the values are hex encoded ?

src/libs/colorpicker.c
```diff
@@ -290,7 +293,10 @@ static void _update_samples_output(dt_lib_module_t *self)
     {
       case 0:
         // RGB
-        snprintf(tooltip, sizeof(tooltip), "%3d   %3d   %3d",
+        snprintf(tooltip, sizeof(tooltip), "%3d   %3d   %3d   %02X%02X%02X",
```
TurboGit, line 296: Likewise.

src/libs/colorpicker.c
jenshannoschwalm: Yeah that is better. Changed to `0   128   255  (0x0080FF)`

src/libs/colorpicker.c
jenshannoschwalm: Sure

TurboGit:
Works for me now. Thanks.

PR #4670, opened 2020-04-10
easier management of user CSS tweaks

### LINKED ISSUE #4662: Feature: Make CSS Tweaking Simpler
This feature request stems from discussions in #4660 and #4643.

Essentially the idea is that a user can use a standard darktable theme (for example, darktable-elegant-grey) but add their own tweaks to it, by creating a file with the same name as the theme (e.g. themes/darktable-elegant-grey) in their own config directory. That file would only contain the altered css.

When darktable loads the user's chosen theme, in this scenario, it would load the base theme first and then incorporate the user's changes. It would be equivalent to using ``@import`` to load the base theme at the start of the user's own css.

elstoc:
if a user wishes to make a minor tweak to their chosen darktable theme,
   this change allows them to place a file named user.css in their
   config directory containing those tweaks, which are applied if they
   select the appropriate option in the preferences.

Resolves #4662

### DIFF
--- a/src/gui/gtk.c
+++ b/src/gui/gtk.c
@@ -2634,7 +2634,7 @@ void dt_gui_add_help_link(GtkWidget *widget, const char *link)
 // load a CSS theme
 void dt_gui_load_theme(const char *theme)
 {
-  char path[PATH_MAX] = { 0 }, datadir[PATH_MAX] = { 0 }, configdir[PATH_MAX] = { 0 };
+  char path[PATH_MAX] = { 0 }, datadir[PATH_MAX] = { 0 }, configdir[PATH_MAX] = { 0 }, usercsspath[PATH_MAX] = { 0 };
   dt_loc_get_datadir(datadir, sizeof(datadir));
   dt_loc_get_user_config_dir(configdir, sizeof(configdir));
 
@@ -2663,8 +2663,34 @@ void dt_gui_load_theme(const char *theme)
 
   if(!gtk_css_provider_load_from_path(GTK_CSS_PROVIDER(themes_style_provider), path, &error))
   {
-    printf("%s: error parsing %s: %s\n", G_STRFUNC, path, error->message);
+    fprintf(stderr, "%s: error parsing %s: %s\n", G_STRFUNC, path, error->message);
     g_clear_error(&error);
+  }
+
+  if(dt_conf_get_bool("themes/usercss"))
+  {
+    g_snprintf(usercsspath, sizeof(usercsspath), "%s/user.css", configdir);
+
+    if(g_file_test(usercsspath, G_FILE_TEST_EXISTS))
+    {
+      //need to append currently loaded theme with user tweaks
+      const gchar *themecsscontent = gtk_css_provider_to_string(GTK_CSS_PROVIDER(themes_style_provider));
+
+      if(!gtk_css_provider_load_from_path(GTK_CSS_PROVIDER(themes_style_provider), usercsspath, &error))
+      {
+        fprintf(stderr, "%s: error parsing %s: %s\n", G_STRFUNC, usercsspath, error->message);
+        g_clear_error(&error);
+      }
+
+      const gchar *usercsscontent = gtk_css_provider_to_string(GTK_CSS_PROVIDER(themes_style_provider));
+      const gchar *combinedcsscontent = g_strjoin(" ", themecsscontent, usercsscontent, NULL);
+
+      if(!gtk_css_provider_load_from_data(GTK_CSS_PROVIDER(themes_style_provider), combinedcsscontent, -1, &error))
+      {
+        fprintf(stderr, "%s: error parsing combined CSS: %s\n", G_STRFUNC, error->message);
+        g_clear_error(&error);
+      }
+    }
   }
 
   g_object_unref(themes_style_provider);
--- a/src/gui/preferences.c
+++ b/src/gui/preferences.c
@@ -183,6 +183,13 @@ static void theme_callback(GtkWidget *widget, gpointer user_data)
   dt_bauhaus_load_theme();
 }
 
+static void usercss_callback(GtkWidget *widget, gpointer user_data)
+{
+  dt_conf_set_bool("themes/usercss", gtk_toggle_button_get_active(GTK_TOGGLE_BUTTON(widget)));
+  dt_gui_load_theme(dt_conf_get_string("ui_last/theme"));
+  dt_bauhaus_load_theme();
+}
+
 ///////////// gui language and theme selection
 
 static void language_callback(GtkWidget *widget, gpointer user_data)
@@ -277,6 +284,16 @@ static void hardcoded_gui(GtkWidget *grid, int *line)
   gtk_widget_set_tooltip_text(widget, _("set the theme for the user interface"));
   gtk_grid_attach(GTK_GRID(grid), label, 0, (*line)++, 1, 1);
   gtk_grid_attach_next_to(GTK_GRID(grid), widget, label, GTK_POS_RIGHT, 1, 1);
+
+  label = gtk_label_new(_("modify selected theme with user.css"));
+  gtk_widget_set_halign(label, GTK_ALIGN_START);
+  GtkToggleButton *cssbutton = GTK_TOGGLE_BUTTON(gtk_check_button_new());
+  gtk_widget_set_tooltip_text(GTK_WIDGET(cssbutton), _("load user.css from the user config directory to modify selected theme"));
+  gtk_toggle_button_set_active(cssbutton, dt_conf_get_bool("themes/usercss"));
+  g_signal_connect(G_OBJECT(cssbutton), "toggled", G_CALLBACK(usercss_callback), 0);
+  gtk_grid_attach(GTK_GRID(grid), label, 0, (*line)++, 1, 1);
+  gtk_grid_attach_next_to(GTK_GRID(grid), GTK_WIDGET(cssbutton), label, GTK_POS_RIGHT, 1, 1);
+
 }
 
 ///////////// end of gui and theme language selection


parafin:
Just a thought - wouldn't it be easier to prepend @import command to the contents of user CSS file instead of including the original theme ourselves?

AlicVB:
yes, agree with @parafin 
One problem with your approach is that you somehow "force" the user to overwrite an existent theme. Although we may want to keep original themes untouched and save tweaked themes with another name so user can easily revert back to it if there's a problem. (and it would ease issues debugging : `change your theme to darkatable.css. Does it happen again ?`)

So I can propose 2 others approach (they can be combined with this one) : 
1. At first dt startup, write a new empty theme file in `mytheme.css` user` .config `dir already filled with the correct` @import` line on top (with a quick tutorial as comments inside)
2. Before theme loading, check the `@import` line, and if correct it with full path if it as been written like `@import("darktable.css")` or even add it if it's not present. The correction should be done on the fly and not saved to the theme file.

Point 2 will make theme sharing easier as you don't have to fight with full path anymore.

What do you think ?

parafin:
What if user wants to base his modification on non-default theme? Say elegant of gray?

elstoc:
I'll just start this off by saying that I'm sure my approach could almost certainly do with some improvement - it's just the way I managed to get it working. I hoped it would provoke discussion.

> Just a thought - wouldn't it be easier to prepend @import command to the contents of user CSS

You mean read the user CSS file into a string, then prepend the import command before loading the style? That could work.

> One problem with your approach is that you somehow "force" the user to overwrite an existent theme.

No. A user can still create their own using @import. Presumably a user doing this will be more of a 'power' user anyway.

> At first dt startup, write a new empty theme file in mytheme.css user.configdir already filled with the correct @import line on top (with a quick tutorial as comments inside)

Which @import line? Pointing to which CSS file?

> check the @import line, and if correct it with full path if it as been written like import("darktable.css")

Again, to which CSS file should it point? What if it's been written as a full path or the css file is nonexistent?

> What if user wants to base his modification on non-default theme? Say elegant of gray?

Again, this doesn't replace the existing way of doing things. It's just a quick way of allowing minor style changes. I tested this by creating "darktable-elegant-grey.css" in my config directory and just adding changed grey values for the histogram.

elstoc:
Just to clarify the reason I had to do it this way (load each file into a theme then convert it to text and combine them) is because the ``@import`` commands in the 'base' styles are relative paths. Theoretically a user could do the same in their local style as well. So any assumptions about paths in the ``@import`` command would have to be spelled out to the user.

AlicVB:
What I have in mind : 
read the user theme file and find the content of the @import  (noted as f_css below)
- if f_css exist as it : no change
- if .config/themes/f_css exist : no change
- if install_dir/darktable/themes/f_css exist : replace f_css by install_dir/darktable/themes/f_css
- otherwise (no @import or non-finded file) replace f_css by install_dir/darktable/themes/darktable.css (the main original theme)

@parafin : if a user want a modified darktable-elegant-grey.css version, he would have to create a file in theme dir with @import("darktable-elegant-grey.css") at the beginning, without dealing with full paths

elstoc:
I was trying to avoid a user having to use ``@import`` for simple changes at all.

AlicVB:
@elstoc : again, I'm not sure that we should encourage non "power-user" to surcharge original themes without renaming them.
As @jenshannoschwalm has said somewhere, we may face to issue with malformed css files. And it will be easier if we can just say "try to revert to darktable theme in the prefs". That can even be something we ask as mandatory in the issue description.


> I was trying to avoid a user having to use `@import` for simple changes at all.

That's covered by the last point...

elstoc:
Then apologies - I got completely the wrong end of the stick from our previous discussions. @AlicVB when you say 'replace', are you talking about _changing_ the ``@import`` command directly in the user's CSS file or just internally re-interpreting it? I'm assuming the latter, since the former could break if the data directory subsequently changed location.

You'd have to be able to handle multiple ``@import`` commands being present. For example, I might create a single CSS file that just has my tweaks to the lighttable border and then want to reference it from multiple user styles, each of which would import both a 'base' style and my standard tweaks.

AlicVB:
> @AlicVB when you say 'replace', are you talking about _changing_ the `@import` command directly in the user's CSS file or just internally re-interpreting it? I'm assuming the latter, since the former could break if the data directory subsequently changed location.

Yes, internally. There should be a way to load css files from a text buffer (or we can use a temp file)

> You'd have to be able to handle multiple `@import` commands being present. For example, I might create a single CSS file that just has my tweaks to the lighttable border and then want to reference it from multiple user styles, each of which would import both a 'base' style and my standard tweaks.

Well, didn't think of that... So probably my idea is not so great :(

elstoc:
> There should be a way to load css files from a text buffer

Yes that's what ``gtk_css_provider_load_from_data`` does in my current solution, but afaik all of the ``@import`` directives would have to use absolute paths within that css string.

AlicVB:
> Yes that's what `gtk_css_provider_load_from_data` does in my current solution, but afaik all of the `@import` directives would have to use absolute paths within that css string.

yes but at this point we can guess it...

Anyway, multiple import may be a nightmare to handle...

My main concern about your approach is the fact that we user can't easily revert back to original theme without deleting/renaming/... is theme file.

But maybe this point can be worked around : 
Currently with your solution and admitting that If write a `darktable.css` file in my `.config` dir with `#thumb_image { border : 10px solid red; }` inside : I end up with 2 `darktable` lines in the pref theme combobox. And both ends to the same result...
If we can differentiate the 2 entries (`darktable` and `darktable (modified)` for example) and ensure that `darktable` is really the original theme without user additions, That would be ok...

Note sure to understandable. so ask if you need explanations...

elstoc:
> I end up with 2 darktable lines in the pref theme combobox

I hadn't considered that it might add another entry to the preferences dialog. In my proposal I would expect there to be only one entry, ``darktable``, in the preferences, which includes user additions if there are any. To revert to the base style the user would have to (as you say) delete/rename their own css. If the user wants their own named theme they should do so using the existing methods (using ``@import``).

Another way to achieve this PR would to include an additional multi-line preferences text box that allows you to put CSS tweaks only, which could then be concatenated to whatever theme was selected.

Nilvus:
> Another way to achieve this PR would to include an additional multi-line preferences text box that allows you to put CSS tweaks only, which could then be concatenated to whatever theme was selected.

That's probably a much better way, yes. And if doing such thing, I would see a new tab on preferences for that, with the possibility to select the theme first that the user want to tweak (couid be the used theme or possibility to select another theme without changing the used one. This would allow to add the good local path to the default theme tweaked on CSS file saved in user config folder (sot that way the user don't have to know/care of the default path).

Then below the additional multi-line preferences text box.

I'm not sure of the following idea but we could also think of having a first editable box to put CSS tweaks and just below a not editable one (but possibility to select text on it) which display the whole choosed theme. So with that, displaying all CSS with comments, and allowing user to just select which parts if want to edit and then copying this part on the editable box above.

This will clearly need to have a tutorial (an add on the manual) and probably improving comments. I will try to see soon if I could improve comments (should not be too long) but as I said already, I will not spend time by now on manual. Probably later this year when manual will be in markdown. This could wait as all this will be for next major release.

AlicVB:
@Nilvus : I quite like your proposal. The only tricky part is adding the good `@import` absolute path. This will -de facto- prevent theme sharing between users. Not that I absolutely want that users share themes, but we have to keep that in mind.
If we prohibit reusing "original" themes names, we can easily replace on the fly every occurrences of `@import("darktable*.css")` by `@import("full/path/to/darktable*.css")` That way, users themes will not have absolute paths inside and can be shared (and are more easy to write for users)

@elstoc : do you think this is doable ?

Nilvus:
> @Nilvus : I quite like your proposal. The only tricky part is adding the good `@import` absolute path. This will -de facto- prevent theme sharing between users. Not that I absolutely want that users share themes, but we have to keep that in mind.

I have just complete your first idea of a notebook part for that and @elstoc ideas. So my proposal is no more than a mix of your ideas to both of you and mine on top of that!

About the CSS path, of course is quite the tricky part. The best would have the possibility to link correctly to te default used theme and allow a simple way to share the custom theme. And to make that more tricky, think of path for all OS...

elstoc:
Ok I've mulled over everyone's comments. My intention here is for a simple change to allow very minor tweaks to css. For myself, I'll probably only use it right now to apply my own thumbnail margins using the code delivered in #4660 and ideally I'd like it to persist across styles - so I could change my colour scheme and still retain my adjusted margins.  There's nothing wrong with making user themes easier to create (such as the auto-path-completion of the ``@IMPORT`` directive suggested above) but It's not something I want to address in this PR.

In order to address some of the comments against this and the associated issue I'd like to propose an alternative solution which borrows from Firefox's handling of its userChrome.css file.  Essentially we would add a single new yes/no preference item immediately below the theme selection: "Apply usertweaks.css if present". If the user selects this, then, after applying the selected style, darktable would append the content of the usertweaks.css file if it is present in the root of the user's config directory. This has the following advantages:

- For those options (like overlay margins) which would previously have been in the preferences section it allows us to share small tweaks with users who want to make minor changes (like those given for firefox here https://github.com/Timvde/UserChrome-Tweaks) without them having to get in to creating their own themes (with all the ``@IMPORT`` fun that ensues)
- Users can change their overall theme without having to create multiple user styles - their tweaks will persist across themes
- Tweaks can be easily switched off without deleting the associated file
- We can place a warning in the tooltip for the option stating that we will not provide support for user tweaks to try and avoid a flood of issues and give us a first-response option on github of "did you try with this option turned off?" and then quick closure if they say no

I've thought again about the multi-line-preference option but IMO this would require a lot of space in the preferences dialog and I don't think it's necessary - a named file is the better approach.

TurboGit:
This looks promising, the main need that was discussed very early was that user's could change the `font` and overall `size` the font. Would that be feasible with this solution ?

elstoc:
I'm not sure that it would right now but perhaps with some changes to darktable.css. Font size currently seems to be hard coded in a few places in darktable.css (in 'ems').

parafin:
Idea sounds fine to me. As for changing font - I think it deserves a separate option in preferences, this feature is too common to hide it in user CSS file.

elstoc:
I think the difficulty with a global font size using this solution is that the usertweaks.css would be placed at the end of the theme css and probably what you want is to set a global font size at the beginning. I think the way this works with gtk is that you set the global font size at the start and then use 'rems' to define subsequent font variations. Though I may well be missing something.

AlicVB:
yes, sounds fine too.
Some remark (for enhancements, not criticism) : 
- for the place taken in pref dialog, What about a specific theme tab ? This would contains the theme dropdown, the "add user tweaks" checkbox and the rest of the space for a basic textarea.
- Maybe we can have an "startup" file with some of the main features inside, but commented with `/* */` That may work as a quick tutorial. Note that we will have to copy this file in user .config dir at startup if it doesn't exist.

elstoc:
As I said, IMO a file is a better solution - there's no need for a separate textarea in an already crowded preferences popup. The yes/no description can provide the instructions about where to put it.

Documentation for the rest (once the transfer to markdown is complete). We could set up a separate subdirectory in the git project with specific tweaks as for the firefox one I linked above.

Given that your enhancements would work on top of my suggestion, perhaps I could implement the basic file-based solution first and then consider enhancements later. Maybe as part of a general rework of the preferences layout.

AlicVB:
@elstoc : yes, no problem. Do as you want, this doesn't change the base idea. So we could always enhance it later if needed !

TurboGit:
>     * for the place taken in pref dialog, What about a specific theme tab ? This would contains the theme dropdown, the "add user tweaks" checkbox and the rest of the space for a basic textarea.

I love this idea as editing a CSS file in a specific area is not easy for anyone. With this option one can just copy/paste a CSS given by others to play with. Easy for everyone. It is probably more work and may come after the initial work but I vote for it !

elstoc:
Yes it is a good idea. I think generally though, the preferences dialog has grown rather organically over time and could probably do with an overhaul with some more tabs. Currently has lots crammed into the first two tabs and then a third tab that seems kind of wasted space. It would make sense to do this at the same time.

Nilvus:
> Yes it is a good idea, perhaps the file thing is just a personal preference. I think generally though, the preferences dialog has grown rather organically over time and could probably do with an overhaul with some more tabs. Currently has lots crammed into the first two tabs and then a third tab that seems kind of wasted space. It would make sense to do this at the same time.

I agree that the content of session options could be moved. I suggest to put these on top of core options tab. Then session options tab could become the CSS edit one.

elstoc:
New version pushed with the proposed changes. @AlicVB on a related note, I tried testing this with the CSS change from #4660 but the margin css no longer seems to change anything.

AlicVB:
@elstoc : strange, just tested with fresh master and this PR and everything seems to work. I've added `#thumb_image{ margin : 20px; }` and get big margins... (overlay mode "hover normal")

About your new proposal, that seems to work great. I'll add some minor comments directly in the code (if I manage to do so...)

src/gui/gtk.c
```diff
@@ -2667,6 +2667,32 @@ void dt_gui_load_theme(const char *theme)
     g_clear_error(&error);
   }
 
+  if(dt_conf_get_bool("themes/usertweaks"))
+  {
+    g_snprintf(usercsspath, sizeof(usercsspath), "%s/usertweaks.css", configdir);
```
AlicVB, line 2672: I suppose that the file isn't in the `theme` subdir to avoid mixing with "real" themes. Right ?
So why no using "just" `user.css` ? it may be simpler to remember and less prone to typing errors...

elstoc: Can do. Naming things is always where I run out of ideas.

And yes I've kept it out of themes on purpose because it needs a theme to be loaded as well.

elstoc: Also the description/tooltip in the preferences might need some editing. Again, I wasn't entirely sure how much to put.

src/gui/gtk.c
```diff
@@ -2667,6 +2667,32 @@ void dt_gui_load_theme(const char *theme)
...
+
+    if(g_file_test(path, G_FILE_TEST_EXISTS))
```
AlicVB, line 2674: should be `usercsspath` I imagine ;)

elstoc: Yes, though the code seems to work anyway!

elstoc:
Regarding the margin css I'm using my PR on top of 291c348 and I can't make it work with any of the overlay settings.

elstoc:
@AlicVB latest push should incorporate your suggestions and I've put slightly better label/tooltip on the new preference

AlicVB:
@elstoc : can you try on master with original `darktable.css` theme and *just* the line 1537 changed with `margin: 20px;` Does it change the margins ?

elstoc:
No it doesn't. But if I change the line above it to ``border: 4px solid red;`` that works as expected

AlicVB:
@elstoc : can you open an issue for that ? to avoid polluting too much your PR...

elstoc:
will do

elstoc:
#4748

elstoc:
Unless anyone has any further comments I think this is done

src/gui/gtk.c
```diff
@@ -2667,6 +2667,32 @@ void dt_gui_load_theme(const char *theme)
...
+    g_snprintf(usercsspath, sizeof(usercsspath), "%s/user.css", configdir);
+
+    if(g_file_test(usercsspath, G_FILE_TEST_EXISTS))
+    {
+      //need to append currently loaded theme with user tweaks
+      const gchar *themecsscontent = gtk_css_provider_to_string(GTK_CSS_PROVIDER(themes_style_provider));
+
+      if(!gtk_css_provider_load_from_path(GTK_CSS_PROVIDER(themes_style_provider), usercsspath, &error))
+      {
+        printf("%s: error parsing %s: %s\n", G_STRFUNC, usercsspath, error->message);
```
TurboGit, line 2681: Errors should go into stderr, in dt we use:

`fprintf(stderr, ...`

src/gui/gtk.c
```diff
@@ -2667,6 +2667,32 @@ void dt_gui_load_theme(const char *theme)
...
+        g_clear_error(&error);
+      }
+
+      const gchar *usercsscontent = gtk_css_provider_to_string(GTK_CSS_PROVIDER(themes_style_provider));
+      const gchar *combinedcsscontent = g_strjoin(" ", themecsscontent, usercsscontent, NULL);
+
+      if(!gtk_css_provider_load_from_data(GTK_CSS_PROVIDER(themes_style_provider), combinedcsscontent, -1, &error))
+      {
+        printf("%s: error parsing combined CSS: %s\n", G_STRFUNC, error->message);
```
TurboGit, line 2690: likewise

elstoc:
Thanks @TurboGit. As with the metadata PR I was just reusing what was there before, but I've updated with the latest commit, since there were only three occurrances.

TurboGit:
Looks good to me. Thanks!

PR #4671, opened 2020-04-10
Powerfull extract wb v2 + python version

johnny-bit:
This set of changes brings `extract_wb` further towards issue-free generation of finetuned presets for all camera bodies. It has been field tested on multiple bodies and data now, allowing for less and less chanes.

Newest changes include (in order of importance)
1. Allowing processing of data from multiple different camera bodies, so one can do virtually unlimited number of bodies.
2. Detection of actual fluorescent preset used on Sony body
3. Removal and translation of `ISO Studio Tungsten` to `Tungsten`
4. Adding `Natural Auto` to list of ignored presets.

//EDIT:
I'm also doing ruby->python translation since many complained about wanting to run `extract_wb` and directly contribute code, however being stopped short by ruby+nokogiri being stubborn.


johnny-bit:
There's whole new file - `extract_wb.py` which is direct translation of `extract_wb`, does the same, but is guaranteed to work on most if not all python3 installs :)

This should greatly simplify getting wb presets :)

This is no longer WIP and is ready for review.

TurboGit:
Why not remove the Ruby version now? We certainly don't plan to maintain both version.

johnny-bit:
That is actually good question. I thought about having it linger for a while to weed out any bugs but it doesn't really make sense...

TurboGit:
> I thought about having it linger for a while to weed out any bugs

It will be in Git history for eternity, isn't this long enough :)

johnny-bit:
done :)

TurboGit:
Thanks !

PR #4678, opened 2020-04-11
Only show masks in darkrooms expose for valid modules. Fix #4668

### LINKED ISSUE #4668: Gradient re-use issue still present.
So after a week away with msys2 problems, I was surprised to find the same problem survives.

Here is a minimal reproduction set:

Import an image with no xmp
Take it into darkroom
Exposure -> drawn mask -> gradient
Crop & rotate: crop in both directions, offset a little
Mask manager, click on the saved gradient, see segmentation error.
Same point in the code as before, even though the following lines are now gone...

dt3.1.0+1133 & Win 10.
Note the code is announced as dirty because I had to turn off maps to get it to build.
[darktable_bt_87DZI0.txt](https://github.com/darktable-org/darktable/files/4460829/darktable_bt_87DZI0.txt)

jenshannoschwalm:
There were dr crashes as reported & discussed in #4668 and #4614

We were somewhat on the wrong track. The crash could be reproduced but the important
detail was that the modules had to be closed on the right panel to get the crash.

There was simply a test missing, we only can display the mask if `dev->gui_module` is valid too :-)

### DIFF
--- a/src/views/darkroom.c
+++ b/src/views/darkroom.c
@@ -606,7 +606,7 @@ void expose(
   else
   {
     // masks
-    if(dev->form_visible && dev->gui_module->enabled)
+    if(dev->gui_module && dev->form_visible && dev->gui_module->enabled)
       dt_masks_events_post_expose(dev->gui_module, cri, width, height, pointerx, pointery);
     // module
     if(dev->gui_module && dev->gui_module->gui_post_expose)


GrahamByrnes:
Almost :-)
No more crashes, but now I can't use ctrl-scroll to change the opacity... which is stuck at 0%
BTW, what gets broken when the module->enabled is removed? Other than seeing crop & rotate helper grid a lot more than I should...

GrahamByrnes:
(I see one problem of deleting the `module->enabled` test: ctrl-Z  deletes straight back to the original image)

jenshannoschwalm:
Ok, still on the wrong track i guess!!!

The extra checking prohibits crashing but after reading your comments i tested more and found something that is definitely wrong in the mask manager and **might** explain the problem because the extra test prohibited critical cases.

Unfortunately this is deep in gtk stuff where i don't feel safe (yet :-1: )

The problem is very easy to trigger, open dt in console ...
1. Discard history from an image, open in dr
2. open expose module and create a mask (circle, gradient ...)
3. Leave the expose module open for now!
4. BTW you can add more masks now
5. Open mask manager while keeping the expose module open

Bildschirmfoto von 2020-04-11 18-39-20

6. Left Click on one of the **later** added masks (here just on the gradient text) allows to edit the name, a right click is also possible and won't hurt.
7. Doing the same, a left click repeated times will not allow to edit the name but will add an extra editable field which is simply wrong (cant screenshot this) and on the console you see such
```
(darktable:89398): GLib-GObject-WARNING **: 18:39:03.435: invalid unclassed pointer in cast to 'GtkTreeStore'

(darktable:89398): Gtk-CRITICAL **: 18:39:03.435: gtk_tree_store_set_valist: assertion 'GTK_IS_TREE_STORE (tree_store)' failed

(darktable:89398): Gtk-CRITICAL **: 18:43:22.751: _gtk_accel_group_detach: assertion 'g_slist_find (accel_group->priv->acceleratables, object) != NULL' failed
```
I think **this** is the problem
8. Now close the expose module and left click on gradient (or any other later module) and nothing else happens.
9. Click on circle and voila: the crash you described.

10. I guess there is a missing initialise of the list but couldn't detect it so far

@TurboGit @AlicVB Could you help out here?

GrahamByrnes:
Ah, an old friend:
`(darktable:89398): Gtk-CRITICAL **: 18:43:22.751: _gtk_accel_group_detach: assertion 'g_slist_find (accel_group->priv->acceleratables, object) != NULL' failed`
This is, I think, back to the problem of searching the list of tool-tips. I ran into it when I was playing with freeing variables in masks.c... tried a few things with the comparator function, got nowhere, left it alone.

jenshannoschwalm:
I did **a lot of more** testing.

There are some things interwoven here.
1. This tiny pr is absolutely correct, it does nothing more than checking in darkrooms expose: **Only** expose the masks if the related module in gui is expanded. This test is absolutely correct.
2. What you describe like loosing control of opacity and others are still correctly working if the gui section of the module is expanded and you activated masks with the arrow icon there.

What is broken but not topic of this pr.
1. As described above the mask manager does **not** keeps track of the gtk list in a correct way.
2. Selecting a mask via the manager is not reflected in the modules gui if it's expanded.
3. The opacity slider in the modules gui is not affected by mouse scroll action.

If we want to be able to visualize and/or control the masks with the mask manager only (without expanded modules gui) this pr would be wrong.

GrahamByrnes:
The irony is that I rarely use masks, but they are a crucial point in
debugging the preview_downsampling mod. Consequetly I'm not sure how some
of them *should* behave. It seems strange to me that one control curvature,
width/feathering, but not opacity via the control points (as invited by the
tool-tip.

So you're saying that the PR is correct, but fails because masks is broken;
while if we make line 609 incorrect it somehow patches the error in masks
(ie ctr-scroll works) at the cost of other errors (ctr-z, I think). Could
be...

I even noticed that the failure of images to refresh in lighttable after
applying a style appeared to be fixed when using... call it the 609--
version. If that is true, it's very deep in gtk-land and not limited to
masks.

There must be a way of tracing ctr-scroll to see what is failing to connect
that to the action in version 609++, why it works in 609--, and fails a
!NULL test in 609_master. That would have the advantage of not continually
recovering from crashes. My task for today then.

Do you happen to understand the warning in the trace-back, re "traceback
stopped because inside another frame, stack corrupted?"

Le dim. 12 avr. 2020 08:28, Hanno Schwalm <notifications@github.com> a
écrit :

GrahamByrnes:
Ok, sloppy testing by me: the ability to change the opacity of the gradient doesn't depend on mods to L609, it depends on having another mask. Add a circle, one circle, one can change the gradient opacity (or cicrle opacity) by ctr-scroll. Delete the circle, it's no longer possible.

In fact: 
circle, path & brush all respond to ctr-scroll, alone or together;
gradient and ellipse respond only if a circle is also present (presumably a path or brush also).

GrahamByrnes:
Maybe I have this completely ballsed-up... and I'm starting to understand @jenshannoschwalm !

If I suppose that a shape already placed on the image, and used to modify eg exposure, should only allow changes when a) the exposure module is open and b) the "drawn mask" submodule is open and c) the "show and modify masks" button is pressed.... then the opacity is the only parameter that behaves correctly.

If otoh I believe that clicking on a shape in mask manager should reconnect me with the shape I placed in an earlier session and allow me to edit it... then I have that freedom for the other parameters (placement, shape, feathering) are all correct, while opacity refuses to shift.

is that it?

Then there are things than can be done via the rhs sliders to shapes in the "group"... and I still don't really get how "blur" fits in among scroll, shift-scroll and ctrl-scroll (ie shape, feather-radius and opacity).

Maybe I should leave this alone... it seems user-unfriendly, but I have a strange vision of UI, so...

GrahamByrnes:
So I vote to merge this. It fixes the crashes, even if it then leaves the user in a confusing UX... but that should be fixed elsewhere.

TurboGit:
This cannot harm indeed and seems ok to me. Thanks!

Issue #2694, opened 2019-06-08
Snapshots: naming, identify and zooming


PaoloAst:
**Is your feature request related to a problem? Please describe.**
I have some difficulties identifying snapshots. Actually, I forgot often which is the snapshot and which is the picture (especially when I rotate upside-down the middle line).
I'd like the possibility to rename the snapshot as I can rename the iop modules on the right.

**Solution you'd like**

I'd like to have:

- The possibility to assign a name to the snapshot
- The snapshot's name superimposed to the image and something like "original" on the other part.
- The possibility to zoom in and out both part of the image (depending on where the mouse pointer is)

**Alternatives I've considered**
It seems there is not a real alternative so far. The only way seems to switch on/off the snapshot button.

GwenaelQ:
Hi everybody, 

I was about to create a new bug report when I saw that this one was already here. I think that what follows is very near (perhaps exact duplicate) of what is said above. 

**Describe the bug**

Zooming or moving in the darkroom view with the image splitted "before/after" (snapshot feature) will  be ineffective in the snapshot "half" of the image
 
**Expected behavior**

When zooming or moving in the darkroom view with the image splitted  (snapshot feature), the user expects that the two "halves" stay "connected".

Furthermore, +1 for the suggestion of naming snapshots and identifiing the parts of the image.


**Platform (please complete the following information):**

 - Darktable Version: 3.0.1 
 - OS: Windows 8.1



Thank you for listening

elstoc:
I sometimes click to rotate the dividing line on a snapshot and forget which is which

GwenaelQ:
> Perhaps just a visual indicator on the dividing line itself might be enough?

It seems to be a good idea.

> Zooming would be harder because the snapshot is basically just an image taken at the current preview scale. You could maybe zoom but it would be much lower res than the other side.

I didn't imagine that snapshot was to be taken in such a literal sense ! It is really not conceivable to imagine a part of the image calculated in one way and other part calculated in another way when zooming ?
Anyway, that's a minor issue so no need to spend a lot of time on it. Your solution is better than nothing :-)

Thank you for the  reply !

chakflying:
I really want to see at least zooming implemented, since the Before/After in Lightroom is so much better and comfortable to use. But as someone who hasn't written C in years I don't imagine I can tackle this easily. Has there been discussion about how to implement this? Would snapshots need to have the current history stack attached to them? Can the view even support zooming while in split view?

[...]

Issue #4650, opened 2020-04-08
Blurry previews in culling mode


Artefact2:
Using the culling view sometimes uses blurry pictures for images that have been edited in darkroom mode. If I zoom to 100%, the issue disappears after waiting for both images to render correctly, but the issue reappears when I zoom back.

Steps to reproduce:
* Make a duplicate of any picture
* Edit both copies in darkroom
* Go back to lighttable, select the two copies, open the culling view with dynamic zoom
* What you see is not what you saw in darkroom (much lower res?)

Screenshots:
The darkroom view, with a sharp preview:
![](https://lutim.l4p1n.ch/f0jcqdi6/2n0zGYGw.jpg)
Now the culling view, with blurry previews:
![](https://lutim.l4p1n.ch/YVfXspJ8/iGLR9vL3.jpg)

~~~
% uname -sr
Linux 5.5.13-arch2-1
% darktable --version
this is darktable 3.1.0+1083~gecd1346e3
copyright (c) 2009-2020 johannes hanika
darktable-dev@lists.darktable.org

compile options:
  bit depth is 64 bit
  normal build
  SSE2 optimized codepath enabled
  OpenMP support enabled
  OpenCL support enabled
  Lua support enabled, API version 5.0.2
  Colord support enabled
  gPhoto2 support enabled
  GraphicsMagick support enabled
  OpenEXR support enabled
~~~

elstoc:
Probably a similar thing to the issues discussed in #4638 (speed optimisations when creating preview images)

aurelienpierre:
This is caused by Cairo downsampling filter.

elstoc:
Is it different to the issue with the duplicate module?

Nilvus:
@Artefact2: culling has been completely rewrited by @AlicVB since your issue. You don't precise version you use but if you can, could you test last master to see if it's ok? If yes, think of closing this issue.

Artefact2:
Thanks. Fixed as of today's master.

Issue #4638, opened 2020-04-07
duplicate manager - inconsistent processing


elstoc:
Within the darkroom the duplicate manager allows you to click on any of the duplicate images to compare with the current edit. However, the image you get when you click on a duplicate is often not consistent with the image you'd get if you double-clicked to change image and view directly in the darkroom.

For example, I import an image and perform some edits on it. I then duplicate that image in the lighttable and do not perform any further edits (so both duplicate images - in this case v0 and v2 - are identical).

I expect that if I view version 0 in the darkroom and click on version 1 in the duplicate manager then, after some processing, the image should remain unchanged (they are identical edits).

The actual result, when the entire image is viewed, is that version 2 seems to be more 'fuzzy' than version 0:

v0 (the image being processed currently)
![full_v0](https://user-images.githubusercontent.com/9555491/78693081-a8f0e600-78f2-11ea-8410-d8f30f4eaf14.png)

v2 (viewed by single-clicking in the duplicate manager)
![full_v2](https://user-images.githubusercontent.com/9555491/78693146-c0c86a00-78f2-11ea-85a2-95d16d057469.png)

If I zoom to 100% the results are better, but the image position seems to be off (the image moves up and to the left by a few pixels when I click on version 2) and sometimes the brightness can be different (though I didn't reproduce that here).

v0 (the current image)
![zoom_v0](https://user-images.githubusercontent.com/9555491/78693260-ebb2be00-78f2-11ea-8879-ee7452ab771b.png)

v2 (from duplicate manager)
![zoom_v2](https://user-images.githubusercontent.com/9555491/78693284-f4a38f80-78f2-11ea-9361-8e518dce1f65.png)

I would like to be able to perform different edits on the same picture and use this module to perform an accurate comparison but this does not currently seem to be possible as I can't rely on the processing to be the same on both. The only reliable and consistent way to do this at the moment is to load v2 in the darkroom, take a snapshot, and then go back and enable the snapshot against v0

These issues seem fairly consistent (I've reproduced on a number of versions including current master and on a number of images) so I haven't uploaded the raw or xmp but I can if needed.

I'm running on ArchLinux and the issue occurs both with and without OpenCL.

Presumably because it's getting the images from the cache, if I switch to version 2 in the darkroom and view v0 from the duplicate manager, both images look the same as they did in my original example above (i.e. v2 is fuzzy).

[...]

Issue #2784, opened 2019-07-13
[Feature request & Issue] Make snapshots adjustable


Nilvus:
**Is your feature request related to a problem? Please describe.**
Snapshots are a great feature to compare image but actually this feature have an important lack : snapshot are not adjustable. For example, with the actual master, you have possibility to hide panels around the image. If you select and make snapshot (with left panel visible) and then hide this panel, the actual image is adjusted in the image frame but not the snapshot. See image below to what darktable show when you do that :

![capture](https://user-images.githubusercontent.com/45535283/61172889-af347a80-a58b-11e9-9b89-73954691f234.jpg)

So actually, we can't always compare differents steps of history as it should.

**Describe the solution you'd like**
Having the snapshot following how the "normal" current state of the image is shown in darkroom and especially by adjusting is rendering the same way (size, place, etc.).

Nilvus:
Still wanted

johnny-bit:
Users want it: https://github.com/darktable-org/darktable/issues/9260 :)

elstoc:
How practical is this to do though? 

The snapshot module is currently the only guaranteed way to get a 1:1 comparison between darkroom edits, because it's a direct copy of the darkroom center view. In order to achieve this with resizing / zoom we would need to either generate and save a full resolution copy of the image or find some way to run two darkroom pixelpipes in parallel (I don't think this is possible). Note that the duplicate module tries to do something similar, but it has limitations because it uses a different method to process the image than the darkroom pixelpipe - I still get odd artifacts when using the duplicate module to compare and I wouldn't want to replicate those issues here.

johnny-bit:
I have had several ideas that could benefit both snapshots & duplicates but would require running parallel pipes or delegated pipes or whatever you call it.

eg for snapshots: do "screenshot" of current pipe and then start low-prio backgroud pipe run for snapshot to cache. once done if user does any movement/adjustment follow the adjustment in main img in snapshot one.

similarly for duplicates.

One could then use cached runs (if of good enough quality) as a way of quick export if eg snapshot done IS the current image history.

Obviously all that is super hard.

yaitskov:
Do refactoring and remove global variables ;)

PR #4680, opened 2020-04-11
snapshot: surround with rectangle using overlay color

elstoc:
replaces the dividing line in the snapshot module (when showing a
snapshot) with a rectangle using the selected overlay line color. This
makes it easier to see which side of the image is the snapshot and which
is the currently-edited image.

Idea originated from #2694

### DIFF
--- a/src/libs/snapshots.c
+++ b/src/libs/snapshots.c
@@ -24,6 +24,7 @@
 #include "develop/develop.h"
 #include "gui/accelerators.h"
 #include "gui/gtk.h"
+#include "gui/draw.h"
 #include "libs/lib.h"
 #include "libs/lib_api.h"
 
@@ -139,20 +140,11 @@ void gui_post_expose(dt_lib_module_t *self, cairo_t *cri, int32_t width, int32_t
     cairo_rectangle(cri, x, y, w, h);
     cairo_fill(cri);
 
-    /* draw the split line */
-    cairo_set_source_rgb(cri, .7, .7, .7);
+    //draw a rectangle surrounding the snapshot
+    //using the selected overlay color
+    dt_draw_set_color_overlay(cri, 0.8, 1.0);
     cairo_set_line_width(cri, 1.);
-
-    if(d->vertical)
-    {
-      cairo_move_to(cri, width * d->vp_xpointer, 0.0f);
-      cairo_line_to(cri, width * d->vp_xpointer, height);
-    }
-    else
-    {
-      cairo_move_to(cri, 0.0f, height * d->vp_ypointer);
-      cairo_line_to(cri, width, height * d->vp_ypointer);
-    }
+    cairo_rectangle(cri, x, y, w, h);
     cairo_stroke(cri);
 
     /* if mouse over control lets draw center rotate control, hide if split is dragged */


Nilvus:
> replaces the dividing line in the snapshot module (when showing a
> snapshot) with a rectangle using the selected overlay line color. This
> makes it easier to see which side of the image is the snapshot and which
> is the currently-edited image.
> 
> Idea originated from #2694

Possible to add a screenshot to see how it's rendered without compiling?

Anyway, snapshot feature needs more things to improve and the most important one I think is the issue I have posted a while ago: #2784

elstoc:
@Nilvus here's an example screenshot with the overlay line colors set to 'red'. The rectangle is less noticeable with the default gray setting.

Screenshot_2020-04-11_11-04-42

With regard to #2784, I think that is also sort of covered within #2694. It would be a very useful feature but because the snapshot is just that - a PNG image of the preview at the point the snapshot was taken - this is a more fundamental and complex change. The alternative way of actually rendering another image has some quality limitations - see #4638 and #4650 - and being able to render the current image at a different point in the pipe would be even harder. Also, it's currently possible to view the snapshot on a different image or version than that which it was taken from (which itself may have been taken at a particular point in the history), increasing the complexity again!

Nilvus:
Thanks for screen capture. Why not the rectangle but the red color is really awful and not consistent to the theme.

I really think we have to keep the initial color (or a little darker) but probably the rectangle would be less visible. Instead of doing a rectangle, maybe it would better to add an indicator on the line (for near the top of the line a triangle pointing to the snapshot side.

Something like that:

snapshot

One more thing: it would also be good to adjust the line color with CSS instead having that hard-coding.

elstoc:
The color is not hard-coded - it's selectable with either ctrl-o or by the icon at the right of the bottom tab. It's also used for other overlays and defaults to the current gray.

The reason it's adjustable in the application and not in CSS is that it's an overlay over the photograph and different colors are more/less visible on different images. For example on a brightly colored image I would probably use gray but on a black&white image I would change it to, perhaps, cyan.

The indicator arrow is another way to go that wouldn't require the rectangle, yes, though we'd need to be sure it was over the image. I'm not 100% sure how to achieve that.

TurboGit:
But still why it is better with the rectangle, nothing tell me that the rectangle IS the snapshot. I'm not 100% satisfied by this solution. Building on top of @Nilvus  idea, maybe use a S as snapshot instead of the triangle ? Or a S (cut & transparent) inside the triangle ? And maybe not an S but using the first letter of the snapshot module to cop with translation ? All these are just ideas to be discussed.

elstoc:
Yes I do agree that a visual indicator like an arrow is probably better than the rectangle, especially when using a gray theme and gray overlay lines. The original issue did suggest overlaying the name of the snapshot but that might become problematic when the area of the rectangle gets smaller. I do like the arrow idea - it's nice and simple. The name of the snapshot is highlighted in the snapshot module so that might be enough to know which snapshot it is.

Nilvus:
> The color is not hard-coded - it's selectable with either ctrl-o or by the icon at the right of the bottom tab. It's also used for other overlays and defaults to the current gray.

Oups, missed that. Of course, better that way than CSS for what you precise after. Anyway, quite strange but 

> The indicator arrow is another way to go that wouldn't require the rectangle, yes, though we'd need to be sure it was over the image. I'm not 100% sure how to achieve that.

If it's too difficult, maybe just put the arrow a little bit below the rotate symbol.

> Or a S (cut & transparent) inside the triangle ? And maybe not an S but using the first letter of the snapshot module to cop with translation ?

Both idea are great but probably 'S' is simple one and clear to see that is the selected snapshot. Selected snapshot could be easily seen on snapshot module, even if multiple snapshots are here.

So @elstoc: for me, the idea of the arrow with the S (cut & transparent) inside the arrow suggested by @TurboGit would be the best thing (or the S just below the arrow) I think. Maybe adding a tooltip when pointing the cursor on the arrow/S that precise what that means would be also great.

elstoc:
> arrow with the S (cut & transparent) inside the arrow

> adding a tooltip when pointing the cursor on the arrow/S

Both good ideas that I unfortunately have no idea how to implement. I'm very much learning as I go here.  If someone could give me some pointers...

Nilvus:
> > arrow with the S (cut & transparent) inside the arrow
> 
> > adding a tooltip when pointing the cursor on the arrow/S
> 
> Both good ideas that I unfortunately have no idea how to implement. I'm very much learning as I go here. If someone could give me some pointers...

I don't know but maybe @aurelienpierre or @AlicVB?

@elstoc: I've just tested to change the color with Ctrl+O and icon on bottom right, but doesn't work on my side, even after restarting darktable. The color remains grey, even if I select red, yellow... I've just test that with a compiled master yesterday and a quite fresh compiled one just few minutes ago (so last actual commit). Seems broken. Did you use on this PR a recent master? Did I have to post an issue here?

elstoc:
Personally I think the S and tooltip are probably superfluous (or maybe I'm just thinking that because it's hard). As with the graduated density module and gradient masks, once you learn which way the arrow points you're probably not going to forget.

This PR is built on top of recent master (and yes the new icon is new). The change that introduced the color selection was 281f617 and some further minor changes have been made in subsequent commits.

Nilvus:
> Personally I think the S and tooltip are probably superfluous (or maybe I'm just thinking that because it's hard). As with the graduated density module and gradient masks, once you learn which way the arrow points you're probably not going to forget.
> 
> This PR is built on top of recent master (and yes the new icon is new). The change that introduced the color selection was (Edit:) [e3d9dd1](https://github.com/darktable-org/darktable/commit/e3d9dd1b08794be227c1b89dbbabe674fd399609) and some further minor changes have been made in subsequent commits.

If we improve visibility snapshot, I think S and tooltip are better to add. Maybe someone could help. Snapshot exist since a long time and this will probably be for 3.0.3 at least or 3.2. So we have time. If improving that, be sure to make that really great.


About commit you point, thanks for pointing it. Anyway, my snapshot line always remains grey....

elstoc:
> Anyway, my snapshot line always remains grey

I don't know the answer for that one. It works fine for me (the branch on which I developed it is taken from the last commit of current master 4cf35ef). Are you 100% certain you're looking at my PR? With a grey line it's virtually indistinguishable from the current functionality. You can check the color changing itself is working as expected by enabling crop/rotate guide lines or drawing a mask, then changing the color.

Nilvus:
> I don't know the answer for that one. It works fine for me (the branch on which I developed it is taken from the last commit of current master [4cf35ef](https://github.com/darktable-org/darktable/commit/4cf35ef40daf2c26d039d6534fdc0f1da575aa33)). Are you 100% certain you're looking at my PR? With a grey line it's virtually indistinguishable from the current functionality. You can check the color changing itself is working as expected by enabling crop/rotate guide lines or drawing a mask, then changing the color.

I don't use your PR. It's on the master (current one, so same commit you precise here). And as I've said previously, colors works great for other parts like guid lines of crop/rotate module or masks. It's just eh snapshot line color that remains grey. That's not the purpose here. I will make some tests and if necessary post an issue.

elstoc:
But coloring the snapshot line is _part of this PR_. In current master it remains gray.

TurboGit:
> Both good ideas that I unfortunately have no idea how to implement. I'm very much learning as I go here. If someone could give me some pointers...

Let me know if you want me to have a look. I can take over this PR if you are ok.

elstoc:
@TurboGit: If you want to take over I'm fine with that.

Nilvus:
> But coloring the snapshot line is _part of this PR_. In current master it remains gray.

Ok, I didn't think this was also a part of this PR. So no issue on my side as I haven't test your PR. Sorry for the noise.

TurboGit:
See https://github.com/darktable-org/darktable/pull/4684

elstoc:
Closing - replaced by #4684

PR #4684, opened 2020-04-11
po/snapsep

TurboGit:


elstoc:
Looks good. I'm still not entirely certain the "S" is necessary but it's not currently centred inside the triangle so looks a little off. Also, the triangle should be positioned so that it's always over the image, otherwise, with the default grey colour, it can blend into the background:

Screenshot_2020-04-11_19-12-50

I do like how the rotation handle changes transparency when the mouse is over the line, but might be better if it only changed transparency when it was over the part of the line that causes rotation when clicked. One of the annoyances currently is it's not obvious where to click just to move the line and where to rotate it.

TurboGit:
@elstoc : New version proposed which should take into account all yours comments. Can you test?

elstoc:
Looking good @TurboGit. The 'S' still isn't centred in the triangle (it's a bit below centre) 

Screenshot_2020-04-12_10-13-16

Apart from that I'm happy with it - it's a significant improvement on my PR.

TurboGit:
> The 'S' still isn't centred in the triangle (it's a bit below centre)

Indeed, this is hard, there is no clean computational way, just adding some padding... I'll improve.

elstoc:
I know. It's the one bit I knew I'd struggle with (and be too perfectionist to leave alone!)

TurboGit:
Should be better now, also the triangle is not displayed when dragging the split line to avoid distracting when analyzing the changes on both sides.

elstoc:
Yep looks much better now. One last thing (I think it's unrelated to your changes) but I find that the rotation is not 100% consistent. If I move the snapshot line higher up the image and click on the rotation symbol repeatedly (i.e. keep the mouse cursor where it is and just keep clicking) it keeps rotating even though my mouse is nowhere near the rotation symbol (which is probably good) but sometimes the snapshot flips 180 degrees.

elstoc:
Oh perhaps that's double clicking and intentional behaviour?

TurboGit:
> Oh perhaps that's double clicking and intentional behaviour? If it is it doesn't happen if I double-click quickly - only if I do it slowly!

That's the issue yes. Not sure we can do something.

elstoc:
I think you might be right. Anyway given that this is a separate issue that's also present in current master, let's call this PR good to merge and deal with it separately (assuming we can come up with a resolution).

TurboGit:
I put some code to avoid this (registering a rotation when one is not yet done).

elstoc:
Excellent thanks @TurboGit. Looks good to merge.

TurboGit:
Merged, thanks for testing!

elstoc:
No problem. Happy to help.

Nilvus:
Good work, really better now. Thanks.

Issue #2986, opened 2019-09-24
Channel Mixer "red" slider values below 0 in hue/saturation/lightness don't affect image


johnny-bit:
**Describe the bug**
Channel mixer module doesn't change image when destination is set to 'hue'/'saturation'/'lightness' and red slider's value is below 0. Image does change in those modes when blue or green sliders are below 0. all sliders affect image in positive range.

**To Reproduce**
1. Go to 'Colour Group'
2. Click on 'Channel Mixer'
3. Set 'destination' to either hue, saturation or lightness
4. Set 'red' slider to value below 0, you'll notice that image doesn't change
5. Set any other slider to value below 0 and notice changes.

**Expected behavior**
Regardless of destination, changing 'red' below 0 should affect image in similar way as changing 'green' or 'blue' does.

**Screenshots**
***Destination: hue***
green below 0 - image changed
![hue: green below 0 - image changed](https://user-images.githubusercontent.com/10687765/65500901-4ee18380-dec0-11e9-8284-70503bf3b361.png)

blue below 0 - image changed
![hue: blue below 0 - image changed](https://user-images.githubusercontent.com/10687765/65500903-4f7a1a00-dec0-11e9-8de3-043001adda22.png)

red below 0 - image unchanged
![hue: red below 0 - image unchanged](https://user-images.githubusercontent.com/10687765/65500904-4f7a1a00-dec0-11e9-9743-24e82617297f.png)

***Destination: saturation***
green below 0 - image changed
![saturation: green below 0 - image changed](https://user-images.githubusercontent.com/10687765/65500905-4f7a1a00-dec0-11e9-84e5-d5b5f541c0ce.png)

blue below 0 - image changed
![saturation: blue below 0 - image changed](https://user-images.githubusercontent.com/10687765/65500906-4f7a1a00-dec0-11e9-865f-5e7c915dea17.png)

red below 0 - image unchanged
![saturation: red below 0 - image unchanged](https://user-images.githubusercontent.com/10687765/65500907-4f7a1a00-dec0-11e9-9271-000fca2994bd.png)

***Destination: lightness***
green below 0 - image changed
![lightness: green below 0 - image changed](https://user-images.githubusercontent.com/10687765/65500910-5012b080-dec0-11e9-9fdf-b1725370fa8f.png)

blue below 0 - image changed
[...]

PR #4687, opened 2020-04-11
Update channelmixer to preserve highlights

hlecleme:
This change to allow using the channel mixer (e.g. to convert image to black and white) without clipping when:

- using the module between the exposure module and the filmic module,
- the exposure value is positive, causing some parts of the image to have RGB values greater than 1.

~There is an additional checkbox to keep the behavior of darktable 3.0.~ The behavior of darktable 3.0 will be kept on edits with the channel mixer activated until the parameters are reset. On new edits, the updated (unclipped) algorithm will be used.

~The change is not 100% equivalent with darktable 3.0 in the case some HSL parameters have been set, to ensure that the RGB values given to the RGB to HSL conversion are between 0 and 1. Note that when using HSL "mixing", the RGB values will always be clipped.~ This change should now be equivalent with darktable 3.0 until the parameters are reset and the updated algorithm is used.

Thanks.

### DIFF
--- a/src/iop/channelmixer.c
+++ b/src/iop/channelmixer.c
@@ -21,6 +21,7 @@
 #include "bauhaus/bauhaus.h"
 #include "common/colorspaces.h"
 #include "common/debug.h"
+#include "common/math.h"
 #include "common/opencl.h"
 #include "control/control.h"
 #include "develop/develop.h"
@@ -51,9 +52,7 @@
   Ilford HP5    23    37    40    Generic B/W   24  68  8
 */
 
-#define CLIP(x) ((x < 0) ? 0.0 : (x > 1.0) ? 1.0 : x)
-
-DT_MODULE_INTROSPECTION(1, dt_iop_channelmixer_params_t)
+DT_MODULE_INTROSPECTION(2, dt_iop_channelmixer_params_t)
 
 typedef enum _channelmixer_output_t
 {
@@ -75,15 +74,22 @@ typedef enum _channelmixer_output_t
   CHANNEL_SIZE
 } _channelmixer_output_t;
 
+typedef enum _channelmixer_algorithm_t
+{
+   CHANNEL_MIXER_VERSION_1 = 0,
+   CHANNEL_MIXER_VERSION_2 = 1,
+} _channelmixer_algorithm_t;
+
 typedef struct dt_iop_channelmixer_params_t
 {
-  //_channelmixer_output_t output_channel;
   /** amount of red to mix value */
   float red[CHANNEL_SIZE];   // $MIN: -1.0 $MAX: 1.0 
   /** amount of green to mix value */
   float green[CHANNEL_SIZE]; // $MIN: -1.0 $MAX: 1.0 
   /** amount of blue to mix value */
   float blue[CHANNEL_SIZE];  // $MIN: -1.0 $MAX: 1.0 
+  /** algorithm version */
+  _channelmixer_algorithm_t algorithm_version;
 } dt_iop_channelmixer_params_t;
 
 typedef struct dt_iop_channelmixer_gui_data_t
@@ -93,12 +99,19 @@ typedef struct dt_iop_channelmixer_gui_data_t
   GtkWidget *scale_red, *scale_green, *scale_blue;    // red, green, blue
 } dt_iop_channelmixer_gui_data_t;
 
+typedef enum _channelmixer_operation_mode_t
+{
+  OPERATION_MODE_RGB = 0,
+  OPERATION_MODE_GRAY = 1,
+  OPERATION_MODE_HSL_V1 = 2,
+  OPERATION_MODE_HSL_V2 = 3,
+} _channelmixer_operation_mode_t;
+
 typedef struct dt_iop_channelmixer_data_t
 {
-  //_channelmixer_output_t output_channel;
-  float red[CHANNEL_SIZE];
-  float green[CHANNEL_SIZE];
-  float blue[CHANNEL_SIZE];
+  float hsl_matrix[9];
+  float rgb_matrix[9];
+  _channelmixer_operation_mode_t operation_mode;
 } dt_iop_channelmixer_data_t;
 
 typedef struct dt_iop_channelmixer_global_data_t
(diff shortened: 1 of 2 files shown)


TurboGit:
I'll wait for @aurelienpierre  feedback.

src/iop/channelmixer.c
```diff
@@ -146,92 +146,144 @@ void connect_key_accels(dt_iop_module_t *self)
...
   shared(data) \
   schedule(static)
 #endif
-  for(int j = 0; j < roi_out->height; j++)
+  for(size_t k = 0; k < (size_t)ch * roi_out->width * roi_out->height; k += ch)
   {
-    const float *in = ((float *)ivoid) + (size_t)ch * j * roi_out->width;
-    float *out = ((float *)ovoid) + (size_t)ch * j * roi_out->width;
-    for(int i = 0; i < roi_out->width; i++)
+    const float *in = ((float *)ivoid) + k;
```
aurelienpierre, line 220: please use `const float *const restrict`

src/iop/channelmixer.c
```diff
@@ -146,92 +146,144 @@ void connect_key_accels(dt_iop_module_t *self)
...
+    float *out = ((float *)ovoid) + k;
```
aurelienpierre, line 221: please use `float *const restrict`

src/iop/channelmixer.c
```diff
@@ -146,92 +146,144 @@ void connect_key_accels(dt_iop_module_t *self)
...
+
+    rgb[0] = rmix;
+    rgb[1] = gmix;
+    rgb[2] = bmix;
+
+    if(gray_mix_mode)
+    {
+      // Calculate graymix
+      // clip blacks as negative values could be amplified by channel factors
+      graymix = (clamp_range_f(rgb[0], 0.0f, INFINITY) * data->red[CHANNEL_GRAY])
```
aurelienpierre, line 277: why clamp between `0.f` and `INFINITY` ? why not a simple `fmax(rgb, 0.f)` ? or did you mean to clamp between `0.f` and  `clip_value` ?

hlecleme:
Thanks for the reviews.

@aurelienpierre If you don't mind, I would be interested to give it a try.

To check if I understood correctly:

- To get the branching out of the pixel loops, did you mean to write one loop for the "HSL mixer" (that will be run if enabled) and one loop for the "RGB mixer" (they would be called sequentially) or one loop for both HSL and RGB mixer and a second one for RGB mixer only? In the first case, should I allocate memory to store the temporary pixel values or reuse the "out" memory (meaning, I think, that I cannot reuse restrict anymore for the second loop)? Do you want a separate loop for the "gray mixing" (where the same value is stored for red, green and blue channel) when RGB mixing is not needed?

- Use OpenMP simd directives for the matrix multiplication loop: is that what you meant by enabling OpenMP optimizations?

To check if the optimizations are effective, should I check the timings with -d perf or is there another way?

aurelienpierre:
@hlecleme my bad, I did a mistake. I thought that the HSL/RGB/grey variants were run separately, but looking at the current source code, they are partly run in sequence. Which is super bad since the HSL conversion happens first and there is no way to unclip it. So, indeed, you can't break entirely break the loops.

But then, the RGB mix is computed even if the grey mix is selected, and only the grey mix get saved to output in this case. 

That's a clusterfuck. I wonder if it wouldn't be better to just start a new module fresh. HSL pixel transform is bad and really not desired in a clean pipe. The RGB controls are inactive for negative values in HSL setup anyway. 

What about starting a new module with only RGB/monochrome variants and no clipping from scratch ?

PS: sorry for late answer, I'm supposed to get notified of replies by emails and never got this one.

aurelienpierre:
@hlecleme just to let you know, @rawfiner and I discussed the channel mixer today, and I got some ideas so here is another prototype : https://github.com/aurelienpierre/darktable/tree/reboot-channelmixer

hlecleme:
The main "if"s have been extracted from the pixel loop, but there is still one in the HSL variant.

When compiled without OpenMP or OpenCL, there is some gain when not using the HSL mixer. Otherwise the performances are similar before and after this change.

For the moment I couldn't find a way to improve the performance, but it probably means that I'm not doing it in the right way.

@aurelienpierre A new module could indeed make things a lot cleaner as we don't need to think about backward compatibility :)

hlecleme:
As @aurelienpierre created his pull request #4940 for the channel mixer replacement (without clipping :)), I think it is best to close this one.

If needed, it can be reopened and I'll be happy to fix the remaining issues.

johnny-bit:
I'm a bit interested in it since I really want to test some ideas with "a bit better" channel mixer rather than "channel mixer reboot" :)

If your changes are mostly optimization and improvement and not whole sale better colour science then there's hardly any harm in continuing, no? (I still continue with my white balance module changes, regardless that new channel mixer is better and more advanced at even temperature corrections :))

hlecleme:
When using the "HSL mixer" part, this change could produce a different result in some cases. I think that would be the only harm possible.

Regarding RGB and Gray, my understanding was that it would be possible to do it in the "channel mixer reboot" (with the same parameter values). At this time it seems to be not exactly the case as it always transforms the pixels to XYZ. I don't know if @aurelienpierre plans to support mixing channels in the work profile color space. If yes, then the modules would be equivalent for that part.

@TurboGit @aurelienpierre What is your opinion on this?

src/iop/channelmixer.c
hlecleme: Replaced this by fmax(rgb, 0.f), thanks

hlecleme:
Updated the implementation. There is no change anymore on the GUI side: the "old algorithm" with its issues (clipping and not correct clipping for HSL) will be used when upgrading from old edits, after a reset or the selection of a preset the new algorithm will be used.

This may fix #5578.

@TurboGit feel free to move this to 3.4

TurboGit:
You've implemented exactly what I described here : https://github.com/darktable-org/darktable/issues/5578#issuecomment-647025369

Ok, I'm not opposed to that, but again this module could well be make obsolete in 3.4 and completely hidden to let only the new channel mixer. So at then end all this work won't be used :)

GrahamByrnes:
I went charging ahead and only just discovered that you went charging ahead first, or faster and better :-) 
While you were in there, did you think of modifying the gui to make the slider ranges adjustable? -2 to 2 seems way too much to me.
Then I ran aground trying to have a way of preserving R+G+B=1.0 in grey mode... and work and stuff.

hlecleme:
For this change (mainly to address some limitations of the channel mixer module in scene referred editing) I would prefer to avoid implementing any additional features or user interface changes. That could be done in separate pull requests.

@aurelienpierre could you check if there are no big issues in the updated algorithms? thanks!

@TurboGit, there seems to be multiple pull requests related to this module (the other two are #5106 and #6069) to propose a fix to #5578 and #2986, would you mind having a look to decide which ones are worth to include while waiting for Aurélien's channel mixer (#4940)? thanks.

GrahamByrnes:
So long as we can agree that the current form, with differences between hsl with/without open_cl, and the clipping in the linear part has issues and it gets replaced... I'm just as happy for it to be this one.

jenshannoschwalm:
@TurboGit what is your position about this pr? Go for it or if not what would be required? (I have something minimal available just ensuring old edits kept and using the fixed algo for new edits but hesitate to pr as this one is already there ...)

TurboGit:
My current plan is to merge this as soon as I'll have tested it against all our integration tests. If it passes it will get merged.

TurboGit:
Let me do that now... running the integration tests...

TurboGit:
Ok, looks fine, let's merge this.

PR #4688, opened 2020-04-11
[RFC] Improve blending in scene linear space

hlecleme:
This is a proposal to improve the blending operations for modules between the exposure module and the filmic module, when some parts of the image have RGB values greater than 1.

It adds the possibility to select the color space that will be used for blending and adds a variant of the RGB color space blending operations that:

- have a set of unbounded blending operators
- uses JzCzhz instead of HSL for Chroma and hue masking

It also adds a boost factor for the parametric mask selection.

Thanks in advance :)

## User interface changes

It is possible to change the color space used for the blending (and masking) operations with the context menu on the side of the mask type selection bar. This is available for all modules where blending is enabled.

context-menu

When the *RGB (scene)* color space is enabled an additional slider is present to adapt the blending operation performed on the image. The slider is enabled on a subset of the blending operations.

blend-fulcrum

The blending operations in *RGB (scene)* are the following:

- normal
- *multiply*
- *multiply (reversed)*
- *addition*
- *subtraction*
- *subtraction (reversed)*
- difference
- *division*
- *division (reversed)*
- average
- geometric mean
- harmonic mean
- RGB red channel
- RGB green channel
- RGB blue channel
- lightness
- chroma

In place of the HSL parametric mask selection, the *RGB (scene)* blending space uses JzCzhz, which should better behave when there are specular highlights.

The parametric mask selection gets a new slider (in all blending spaces) to allow scaling the selection ranges to select those specular highlights. E.g. when selecting the clouds of an underexposed image (that has been exposed for the mid gray with the exposure module) the whole selection is outside of the range of the mask selection:

parametric-without-boost

When changing the boost factor, it is possible to set the selection of the clouds in the mask selection range:

parametric-with-boost

## Performance notes

There is a performance penalty when choosing a color space for the blending that is different than the color space of the module as the processing will need three (or two) color space conversions instead of zero (or one) when the next module has the same color space (respectively another color space).


TurboGit:
I'll wait for @aurelienpierre  feedback.

data/kernels/blendop.cl
```diff
@@ -580,49 +583,47 @@ blendop_Lab (__read_only image2d_t in_a, __read_only image2d_t in_b, __read_only
...
 
     /* fallback to normal blend */
+
+    case DEVELOP_BLEND_BOUNDED:
+    case DEVELOP_BLEND_NORMAL:
     case DEVELOP_BLEND_UNBOUNDED:
     case DEVELOP_BLEND_NORMAL2:
     default:
-      o =  (a * (1.0f - opacity)) + (b * opacity);
+      o =  clamp((a * (1.0f - opacity)) + (b * opacity), min, max);
```
aurelienpierre, line 626: normal blend shouldn't clip

data/kernels/blendop.cl
```diff
@@ -580,49 +583,47 @@ blendop_Lab (__read_only image2d_t in_a, __read_only image2d_t in_b, __read_only
...
+      o.z = clamp((a.z * (1.0f - opacity)) + (b.z * opacity), min.z, max.z);
       break;
       
     case DEVELOP_BLEND_HSV_LIGHTNESS:
     case DEVELOP_BLEND_HSV_COLOR:
     case DEVELOP_BLEND_RGB_R:
     case DEVELOP_BLEND_RGB_G:
     case DEVELOP_BLEND_RGB_B:
-      o = a;                            // Noop for Lab (without clamping)
+      o = clamp(a, min, max);
```
aurelienpierre, line 616: RGB blends shouldn't clip

data/kernels/blendop.cl
```diff
@@ -580,49 +583,47 @@ blendop_Lab (__read_only image2d_t in_a, __read_only image2d_t in_b, __read_only
...
-    case DEVELOP_BLEND_NORMAL:
-      o =  clamp((a * (1.0f - opacity)) + (b * opacity), min, max);
-      break;
-
     case DEVELOP_BLEND_LAB_LIGHTNESS:
     case DEVELOP_BLEND_LAB_L:
-      o.x = (a.x * (1.0f - opacity)) + (b.x * opacity);
-      o.y = a.y;
-      o.z = a.z;
+      o.x = clamp((a.x * (1.0f - opacity)) + (b.x * opacity), min.x, max.x);
```
aurelienpierre, line 588: simple alpha blending like that should not clip

data/kernels/blendop.cl
```diff
@@ -580,49 +583,47 @@ blendop_Lab (__read_only image2d_t in_a, __read_only image2d_t in_b, __read_only
...
+      o.y = clamp(a.y, min.y, max.y);
+      o.z = clamp(a.z, min.z, max.z);
       break;
 
     case DEVELOP_BLEND_LAB_A:
-      o.x = a.x;
-      o.y = (a.y * (1.0f - opacity)) + (b.y * opacity);
-      o.z = a.z;
+      o.x = clamp(a.x, min.x, max.x);
```
aurelienpierre, line 594: simple alpha blending like that should not clip

data/kernels/blendop.cl
```diff
@@ -580,49 +583,47 @@ blendop_Lab (__read_only image2d_t in_a, __read_only image2d_t in_b, __read_only
...
+      o.y = clamp((a.y * (1.0f - opacity)) + (b.y * opacity), min.y, max.y);
+      o.z = clamp(a.z, min.z, max.z);
       break;
 
     case DEVELOP_BLEND_LAB_B:
-      o.x = a.x;
-      o.y = a.y;
-      o.z = (a.z * (1.0f - opacity)) + (b.z * opacity);
+      o.x = clamp(a.x, min.x, max.x);
```
aurelienpierre, line 600: simple alpha blending like that should not clip

data/kernels/blendop.cl
```diff
@@ -580,49 +583,47 @@ blendop_Lab (__read_only image2d_t in_a, __read_only image2d_t in_b, __read_only
...
+      o.y = clamp(a.y, min.y, max.y);
+      o.z = clamp((a.z * (1.0f - opacity)) + (b.z * opacity), min.z, max.z);
       break;
 
     case DEVELOP_BLEND_LAB_COLOR:
-      o.x = a.x;
-      o.y = (a.y * (1.0f - opacity)) + (b.y * opacity);
-      o.z = (a.z * (1.0f - opacity)) + (b.z * opacity);
+      o.x = clamp(a.x, min.x, max.x);
```
aurelienpierre, line 606: simple alpha blending like that should not clip

data/kernels/blendop.cl
```diff
@@ -373,31 +373,34 @@ blendop_Lab (__read_only image2d_t in_a, __read_only image2d_t in_b, __read_only
...
+
   const float4 lmin = (float4)(0.0f, 0.0f, 0.0f, 1.0f);
   const float4 lmax = (float4)(1.0f, 2.0f, 2.0f, 1.0f);       /* max + fabs(min) */
   const float4 halfmax = (float4)(0.5f, 1.0f, 1.0f, 0.5f);    /* lmax / 2.0f */
   const float4 doublemax = (float4)(2.0f, 4.0f, 4.0f, 2.0f);  /* lmax * 2.0f */
   const float opacity2 = opacity*opacity;
 
-  float4 la = clamp(a + fabs(min), lmin, lmax);
-  float4 lb = clamp(b + fabs(min), lmin, lmax);
+  float4 la = clamp((a + fabs(min_clip)) / exposure_scale, lmin, lmax);
```
aurelienpierre, line 387: you can't scale Lab vectors like that, a and b represent the chromaticity, they are not linearly connected to L.

hlecleme:
Thanks for the review.

1. Yes, you're right. The scaling of the Lab (or RGB) pixel is only needed for the following blending modes: screen, overlay, softlight, hardlight, vividlight, linearlight, pinlight (and difference for Lab). As the operations are similar in Lab and (linear?) RGB and some of them seem to expect a "middle gray" at 0.5, I suppose they are to be applied on pixels after a "tone curve" has been applied. In this case I would expect the maximum value to be 1.0 and the exposure_scale should be 1 to not change anything. But perhaps this reasoning is completely wrong? To improve things for Lab should I transform pixels to XYZ, scale them, and transform them the other way for blending (but I guess this would have a too big impact on performance) or would it be better to always clip pixels for those blending modes?

2. This was to merge both the NORMAL_BOUNDED and NORMAL_UNBOUNDED blending modes, but it could have a performance penalty.

3. The clipping is done to keep the compatibility with previous versions of darktable. For those blending modes the pixels are not scaled, are them? Could you explain the last phrase?

4. This was done to have the same effect of the "clip button" on all blending modes, but I can revert that.

About the bounded and not bounded variants w.r.t. the performance, would it be preferable to duplicate the blending modes in the dt_develop_blend_mode_t enum (e.g. MULTIPLY and MULTIPLY_UNBOUNDED) and have a separate implementation for them?

Should I also try to optimize the OpenCL kernels by splitting all blending modes in separate kernels to avoid the switch inside the kernel or this would change nothing?

aurelienpierre:
> As the operations are similar in Lab and (linear?)

There 2 things at play here : the operations, and their meaning. screen, overlay, softlight, hardlight, vividlight, linearlight, pinlight are legacy blending modes inherited from the display-referred era, and need a gamma 2.2 to 2.4 so middle grey = 0.46 to 0.5. You get that these operations don't mean much in linear RGB, even if we changed the threshold to 0.18, because then the RGB range wouldn't be perceptually scaled. However, I see no reason why the output should be < 1, since the rest of the pipeline can handle it.

>  I suppose they are to be applied on pixels after a "tone curve" has been applied

Indeed, but we cannot ensure that, they can be used for every module no matter its place in the pipeline. That's why the whole display-referred approach is bad: it makes hard assumptions you can't check.

> To improve things for Lab should I transform pixels to XYZ, scale them, and transform them the other way for blending (but I guess this would have a too big impact on performance) or would it be better to always clip pixels for those blending modes?

The problem with this approach is it will change the blending, hence the output of previously edited images. Also, Lab messes up alpha blending entirely, because it tempers with the definition of optical occlusion, so I think we should let the Lab path die. Lab > 100 % is handled and allowed in modules such as tone curve by extrapolating the curve past the 100% node. We could simply keep Lab as-is and unclip the output (except for negatives). I can see no reason why Lab should be clipped (except for GUI control of the key-frame, but stuff affecting GUI should be handled at GUI level, not at pixel level). And, again, clipping is handled in colourout module, it's not the job of the masking to be clever about that.

> The clipping is done to keep the compatibility with previous versions of darktable. For those blending modes the pixels are not scaled, are them? Could you explain the last phrase?

The more I think about that, the more I'm convinced that all you need is a GUI scaling, so your 0-100% sliders get a boost factor before going to masking API, and get multiplied by 2, 4, 8…

This way, you avoid flip-flopping pixels twice. Also, there is no point in making screen, overlay, softlight, hardlight, vividlight, linearlight, pinlight ready for scene-referred processing, they are grounded in the display-referred framework from start.

The channels that need scaling are L, R, G, B, g. Then we need a way to make HSL accept RGB > 1. H and L will be fine, it's S that will cause problems. I need to think more about that.

> About the bounded and not bounded variants w.r.t. the performance, would it be preferable to duplicate the blending modes in the dt_develop_blend_mode_t enum (e.g. MULTIPLY and MULTIPLY_UNBOUNDED) and have a separate implementation for them?

If you absolutely need to clip, it's indeed better to separate both variants.

> Should I also try to optimize the OpenCL kernels by splitting all blending modes in separate kernels to avoid the switch inside the kernel or this would change nothing?

I have asked @hanatos about that. It seems that the minor performance gain from avoiding inner-kernel branching might be compensated by the multiplication of different kernels, resulting in memory footprint increase, and not bringing the expected results overall. So, switching cases in kernel is fine-ish.

Anyway, just to be clear, that work is tedious, necessary and your commitment is much appreciated, but maybe you should have reached out so we could lay down a plan before starting. Since I have a feeling you speak French, why don't you join the Matrix channel https://riot.im/app/#/room/#dev_darktablefr:matrix.org ?

TurboGit:
@hlecleme : I'm removing this for 3.4 for now. No pressure and looks like there is many things to update / discussed. Let me know if this is ok for you.

hlecleme:
@TurboGit The implementation has been updated to better match what @aurelienpierre requested. He is aware of it and I'm waiting on his feedback. So I guess it's up to him ;)

TurboGit:
@hlecleme : Understood, but I fear that @aurelienpierre will have too many things to do for 3.4. On his plate there is already FilmicV4 OpenCL, new channel mixel, documentation of FilmicRGB => v3... So yes it's up to Aurélien but I feel that it won't be ready in time, we are ~ 1 month before code freeze and I'd like to avoid such big work to be merged at the last minute.

TurboGit:
BTW, this PR has conflicts and need to be rebased.

hlecleme:
Rebased and retested.

data/kernels/blendop.cl
hlecleme: Code is kept as is for Lab/raw/display RGB blending modes. Scene referred blending mode does not clip.

data/kernels/blendop.cl
hlecleme: Code is kept as is for Lab/raw/display RGB blending modes. Scene referred blending mode does not clip.

data/kernels/blendop.cl
hlecleme: Code is kept as is for Lab/raw/display RGB blending modes. Scene referred blending mode does not clip.

data/kernels/blendop.cl
hlecleme: Code is kept as is for Lab/raw/display RGB blending modes. Scene referred blending mode does not clip.

data/kernels/blendop.cl
hlecleme: Code is kept as is for Lab/raw/display RGB blending modes. Scene referred blending mode does not clip.

data/kernels/blendop.cl
hlecleme: Code is kept as is for Lab/raw/display RGB blending modes. Scene referred blending mode does not clip.

data/kernels/blendop.cl
hlecleme: You're right. Lab vectors are not scaled anymore (the algorithms stay the same as in master). A separate set of not clamping blending modes have been added for scene referred blending.

aurelienpierre:
On my to-do list.

src/common/colorspaces_inline_conversions.h
```diff
@@ -421,6 +422,125 @@ static inline float dt_camera_rgb_luminance(const float *const rgb)
   return (rgb[0] * 0.2225045f + rgb[1] * 0.7168786f + rgb[2] * 0.0606169f);
 }
```
aurelienpierre, line 424: Minor : it's good practice to include source in comment here : https://www.osapublishing.org/oe/fulltext.cfm?uri=oe-25-13-15131&id=368272

src/common/colorspaces_inline_conversions.h
```diff
@@ -421,6 +422,125 @@ static inline float dt_camera_rgb_luminance(const float *const rgb)
...
 
+#ifdef _OPENMP
+#pragma omp declare simd
+#endif
+static inline void dt_XYZ_2_JzAzBz(const float *const XYZ_D65, float *const JzAzBz)
```
aurelienpierre, line 428: could add `const float *const restrict` and `float *const restrict` to inform compiler there is no aliasing

src/common/colorspaces_inline_conversions.h
```diff
@@ -421,6 +422,125 @@ static inline float dt_camera_rgb_luminance(const float *const rgb)
...
+{
+  static const float b = 1.15f;
```
aurelienpierre, line 430: `static` is unnecessary on variables here, it only tells the program to keep in memory to allow other parts of the program to change it while running. It could even harm performance in multithreading context.

src/common/colorspaces_inline_conversions.h
```diff
@@ -421,6 +422,125 @@ static inline float dt_camera_rgb_luminance(const float *const rgb)
   return (rgb[0] * 0.2225045f + rgb[1] * 0.7168786f + rgb[2] * 0.0606169f);
 }
 
+#ifdef _OPENMP
+#pragma omp declare simd
```
aurelienpierre, line 426: `#pragma omp declare simd aligned(XYZ_D65, JzAzBz:16)`

src/common/colorspaces_inline_conversions.h
```diff
@@ -421,6 +422,125 @@ static inline float dt_camera_rgb_luminance(const float *const rgb)
...
+  float XYZ[4] DT_ALIGNED_PIXEL = { 0.0f, 0.0f, 0.0f, 0.0f };
+  float LMS[4] DT_ALIGNED_PIXEL = { 0.0f, 0.0f, 0.0f, 0.0f };
+
+  // XYZ -> X'Y'Z
+  XYZ[0] = b * XYZ_D65[0] - (b - 1.0f) * XYZ_D65[2];
+  XYZ[1] = g * XYZ_D65[1] - (g - 1.0f) * XYZ_D65[0];
+  XYZ[2] = XYZ_D65[2];
+
+  // X'Y'Z -> L'M'S'
+  for(int i = 0; i < 3; i++)
```
aurelienpierre, line 459: maybe add `#pragma unroll` before the loop

src/common/colorspaces_inline_conversions.h
```diff
@@ -421,6 +422,125 @@ static inline float dt_camera_rgb_luminance(const float *const rgb)
...
+    LMS[i] = powf(fmax(LMS[i] / 10000.f, 0.0f), n);
+    LMS[i] = powf((c1 + c2 * LMS[i]) / (1.0f + c3 * LMS[i]), p);
+  }
+
+  // L'M'S' -> Izazbz
+  for(int i = 0; i < 3; i++) JzAzBz[i] = A[i][0] * LMS[0] + A[i][1] * LMS[1] + A[i][2] * LMS[2];
+  // Iz -> Jz
+  JzAzBz[0] = ((1.0f + d) * JzAzBz[0]) / (1.0f + d * JzAzBz[0]) - d0;
+}
+
```
aurelienpierre, line 471: I have checked the XYZ -> JzAzBz conversion (constants values and computations steps) against the source paper so it should be right :-)

src/common/colorspaces_inline_conversions.h
```diff
@@ -421,6 +422,125 @@ static inline float dt_camera_rgb_luminance(const float *const rgb)
...
+#ifdef _OPENMP
+#pragma omp declare simd
```
aurelienpierre, line 473: `#pragma omp declare simd aligned(JzCzhz, JzAzBz:16)`

src/common/colorspaces_inline_conversions.h
```diff
@@ -421,6 +422,125 @@ static inline float dt_camera_rgb_luminance(const float *const rgb)
...
+#endif
+static inline void dt_JzAzBz_2_JzCzhz(const float *const JzAzBz, float *const JzCzhz)
+{
+  float var_H = atan2f(JzAzBz[2], JzAzBz[1]) / (2.0f * DT_M_PI_F);
+  JzCzhz[0] = JzAzBz[0];
+  JzCzhz[1] = sqrtf(JzAzBz[1] * JzAzBz[1] + JzAzBz[2] * JzAzBz[2]);
```
aurelienpierre, line 479: there is a native `hypotf(x, y)` function that computes directly the `sqrtf(x *x + y * y)`, chances are it is optimized at a low level in the C maths lib.

src/common/colorspaces_inline_conversions.h
```diff
@@ -421,6 +422,125 @@ static inline float dt_camera_rgb_luminance(const float *const rgb)
...
+  JzCzhz[2] = var_H >= 0.0f ? var_H : 1.0f + var_H;
+}
+
+#ifdef _OPENMP
+#pragma omp declare simd
+#endif
+static inline void dt_JzCzhz_2_JzAzBz(const float *const JzCzhz, float *const JzAzBz)
+{
+  JzAzBz[0] = JzCzhz[0];
+  JzAzBz[1] = cosf(2.0f * DT_M_PI_F * JzCzhz[2]) * JzCzhz[1];
```
aurelienpierre, line 489: maybe store `2.0f * DT_M_PI_F` into a `#define DT_2_PI_F 2.0f * DT_M_PI_F` because I'm not sure the compiler is clever enough to premultiply the constant ahead instead of doing it at each call.

src/common/colorspaces_inline_conversions.h
```diff
@@ -421,6 +422,125 @@ static inline float dt_camera_rgb_luminance(const float *const rgb)
...
+  float LMS[4] DT_ALIGNED_PIXEL = { 0.0f, 0.0f, 0.0f, 0.0f };
+  float IzAzBz[4] DT_ALIGNED_PIXEL = { 0.0f, 0.0f, 0.0f, 0.0f };
+
+  IzAzBz[0] = JzAzBz[0] + d0;
+  IzAzBz[0] = IzAzBz[0] / (1.0f + d - d * IzAzBz[0]);
+  IzAzBz[1] = JzAzBz[1];
+  IzAzBz[2] = JzAzBz[2];
+
+  // IzAzBz -> LMS
+  for(int i = 0; i < 3; i++)
```
aurelienpierre, line 528: `#pragma unroll` before the loop, just to be safe

src/common/colorspaces_inline_conversions.h
```diff
@@ -421,6 +422,125 @@ static inline float dt_camera_rgb_luminance(const float *const rgb)
...
+  float var_H = atan2f(JzAzBz[2], JzAzBz[1]) / (2.0f * DT_M_PI_F);
+  JzCzhz[0] = JzAzBz[0];
+  JzCzhz[1] = sqrtf(JzAzBz[1] * JzAzBz[1] + JzAzBz[2] * JzAzBz[2]);
+  JzCzhz[2] = var_H >= 0.0f ? var_H : 1.0f + var_H;
+}
+
+#ifdef _OPENMP
+#pragma omp declare simd
+#endif
+static inline void dt_JzCzhz_2_JzAzBz(const float *const JzCzhz, float *const JzAzBz)
```
aurelienpierre, line 486: `float *const restrict`

src/common/colorspaces_inline_conversions.h
```diff
@@ -421,6 +422,125 @@ static inline float dt_camera_rgb_luminance(const float *const rgb)
...
+  // L'M'S' -> Izazbz
+  for(int i = 0; i < 3; i++) JzAzBz[i] = A[i][0] * LMS[0] + A[i][1] * LMS[1] + A[i][2] * LMS[2];
+  // Iz -> Jz
+  JzAzBz[0] = ((1.0f + d) * JzAzBz[0]) / (1.0f + d * JzAzBz[0]) - d0;
+}
+
+#ifdef _OPENMP
+#pragma omp declare simd
+#endif
+static inline void dt_JzAzBz_2_JzCzhz(const float *const JzAzBz, float *const JzCzhz)
```
aurelienpierre, line 475: `float *const restrict`

src/common/colorspaces_inline_conversions.h
```diff
@@ -421,6 +422,125 @@ static inline float dt_camera_rgb_luminance(const float *const rgb)
...
+  IzAzBz[0] = JzAzBz[0] + d0;
+  IzAzBz[0] = IzAzBz[0] / (1.0f + d - d * IzAzBz[0]);
+  IzAzBz[1] = JzAzBz[1];
+  IzAzBz[2] = JzAzBz[2];
+
+  // IzAzBz -> LMS
+  for(int i = 0; i < 3; i++)
+  {
+    LMS[i] = AI[i][0] * IzAzBz[0] + AI[i][1] * IzAzBz[1] + AI[i][2] * IzAzBz[2];
+    LMS[i] = powf(fmax(LMS[i], 0.0f), 1.0f/p);
```
aurelienpierre, line 531: maybe compute `1.f / p` before the loop.

src/common/colorspaces_inline_conversions.h
```diff
@@ -421,6 +422,125 @@ static inline float dt_camera_rgb_luminance(const float *const rgb)
...
+    LMS[i] = 10000.f * powf(fmaxf((c1 - LMS[i]) / (c3 * LMS[i] - c2), 0.0f), 1.0f/n);
```
aurelienpierre, line 532: same, `1.f / n` can be computed ahead

src/common/colorspaces_inline_conversions.h
```diff
@@ -421,6 +422,125 @@ static inline float dt_camera_rgb_luminance(const float *const rgb)
...
+  }
+
+  // LMS -> X'Y'Z
+  for(int i = 0; i < 3; i++) XYZ[i] = MI[i][0] * LMS[0] + MI[i][1] * LMS[1] + MI[i][2] * LMS[2];
```
aurelienpierre, line 536: `#pragma unroll`

src/common/colorspaces_inline_conversions.h
```diff
@@ -421,6 +422,125 @@ static inline float dt_camera_rgb_luminance(const float *const rgb)
...
+
+  // X'Y'Z -> XYZ_D65
+  XYZ_D65[0] = (XYZ[0] + (b - 1.0f) * XYZ[2]) / b;
+  XYZ_D65[1] = (XYZ[1] + (g - 1.0f) * XYZ[0]) / g;
+  XYZ_D65[2] = XYZ[2];
+}
+
```
aurelienpierre, line 543: I have checked the maths but not the values in the inverse matrices. I trust that came out of an algebra software ?

src/common/colorspaces_inline_conversions.h
parafin: #define won't change anything for compiler, because it's an instruction for preprocessor, which runs before the compiler. If DT_M_PI_F is a define too, then I'm very sure compiler will pre-calculate 2.0f * DT_M_PI_F. If it's a variable, then I guess what you need is another const variable.

data/kernels/colorspace.cl
```diff
@@ -325,3 +325,37 @@ inline float4 sRGB_to_XYZ(float4 sRGB)
   return XYZ;
 }
 
+inline float4 XYZ_to_JzAzBz(float4 XYZ_D65)
+{
+  float4 temp1, temp2;
```
aurelienpierre, line 330: you can store the constants into `float4` variables and unroll the dot products like `temp2.x = cnst.x * XYZ.x + cnst.y * XYZ.y + cnst.z * XYZ.z`. Since some GPU have an hardware acceleration for dot products, it could give a hint to the compiler that it can be used here.

data/kernels/colorspace.cl
```diff
@@ -325,3 +325,37 @@ inline float4 sRGB_to_XYZ(float4 sRGB)
...
+  temp1.w = XYZ_D65.w;
+  return temp1;
+}
+
+inline float4 JzAzBz_to_JzCzhz(float4 JzAzBz)
+{
+  float h = atan2(JzAzBz.z, JzAzBz.y) / (2.0f * M_PI_F);
+  float4 JzCzhz;
+  JzCzhz.x = JzAzBz.x;
+  JzCzhz.y = sqrt(JzAzBz.y * JzAzBz.y + JzAzBz.z * JzAzBz.z);
```
aurelienpierre, line 357: there is a `native_sqrt` too which may be a bit faster and a bit less accurate.

data/kernels/color_conversion.cl
```diff
@@ -138,6 +139,16 @@ inline float get_rgb_matrix_luminance(const float4 rgb, constant dt_colorspaces_
...
 }
 
+inline float4 rgb_matrix_to_xyz(const float4 rgb, constant dt_colorspaces_iccprofile_info_cl_t *profile_info, constant float *matrix, read_only image2d_t lut)
+{
+  if(profile_info->nonlinearlut)
+  {
+    float4 linear_rgb = apply_trc_in(rgb, profile_info, lut);
+    return matrix_product(linear_rgb, matrix);
+  }
+  return matrix_product(rgb, matrix);
```
aurelienpierre, line 149: adding `return` in an `if` statement might prevent the compiler to optimize the branches (unswitch and reorder). In SIMD environments, the returned value should not depend on the branches or change the control-flow. Problem is here you return in branch so the control-flow may not predictable ahead.
so you can rewrite it:
```C
inline float4 rgb_matrix_to_xyz(const float4 rgb, constant dt_colorspaces_iccprofile_info_cl_t *profile_info, constant float *matrix, read_only image2d_t lut)
{
  float out = 0.f;
  if(profile_info->nonlinearlut)
  {
    float4 linear_rgb = apply_trc_in(rgb, profile_info, lut);
    out = matrix_product(linear_rgb, matrix);
  }
  else
  {
    out = matrix_product(rgb, matrix);
  }
  return out;
}
```

src/common/color_picker.c
```diff
@@ -28,11 +28,32 @@ static inline size_t _box_size(const int *const box)
   return (size_t)((box[3] - box[1]) * (box[2] - box[0]));
 }
 
+#ifdef _OPENMP
+#pragma omp declare simd aligned(rgb, JzCzhz: 16)
```
aurelienpierre, line 32: `#pragma omp declare simd aligned(rgb, JzCzhz: 16) uniform(work_profile)`

data/kernels/colorspace.cl
hlecleme: Done for the M and A matrices

src/common/colorspaces_inline_conversions.h
hlecleme: Sorry, added a comment with the information

src/common/colorspaces_inline_conversions.h
hlecleme: As the file is included by some C++ code there was some troubles to use restrict.

src/common/colorspaces_inline_conversions.h
hlecleme: The matrices have been inverted by octave, e.g. for the M matrix:

```
format long
inv([0.41478972 0.579999 0.014648;-0.20151 1.120649 0.0531008; -0.0166008 0.2648 0.6684799])
```

src/common/colorspaces_inline_conversions.h
aurelienpierre: right, I forgot.

src/common/colorspaces_inline_conversions.h
aurelienpierre: cool, thanks

src/common/colorspaces_inline_conversions.h
aurelienpierre: ok then.

src/common/colorspaces_inline_conversions.h
aurelienpierre: ok, thanks

aurelienpierre:
I'm testing the code and I find it very impressive. Good job !

A question remains, though : do we need the output masking ? I have personnaly never used it, it's brittle since the masking will depend on the inner module settings, and I have never seen another imaging app masking the output. So, maybe this rewrite is the occasion to declutter this ?

Nilvus:
@TurboGit and @aurelienpierre: there's no milestone on that PR and I hope it should be for 3.4? Maybe set it would be good.

aurelienpierre:
@Nilvus it was implicitely 3.4 :-)

data/kernels/blendop.cl
```diff
@@ -146,100 +159,115 @@ typedef enum dt_dev_pixelpipe_display_mask_t
...
-  float result = 1.0f;
   float scaled[DEVELOP_BLENDIF_SIZE];
 
   if(!(mask_mode & DEVELOP_MASK_CONDITIONAL)) return (mask_combine & DEVELOP_COMBINE_INCL) ? 0.0f : 1.0f;
 
-  scaled[DEVELOP_BLENDIF_L_in] = clamp(input.x / 100.0f, 0.0f, 1.0f);			// L scaled to 0..1
-  scaled[DEVELOP_BLENDIF_A_in] = clamp((input.y + 128.0f)/256.0f, 0.0f, 1.0f);		// a scaled to 0..1
-  scaled[DEVELOP_BLENDIF_B_in] = clamp((input.z + 128.0f)/256.0f, 0.0f, 1.0f);		// b scaled to 0..1
+  scaled[DEVELOP_BLENDIF_L_in] = input.x / 100.0f,			// L scaled to 0..1
+  scaled[DEVELOP_BLENDIF_A_in] = input.y / 128.0f;		// a scaled to -0.5..0.5
```
aurelienpierre, line 179: doesn't that break compatibility with older edits ?

hlecleme: I hope not...

This change has to be reviewed together with the code in function `dt_develop_blend_process_cl` in blend.c.

1. Removed clamping to allow setting mask limits on values > 1.0f (although this is probably not valid from the color space point of view). The (open) limits (0.0f and 1.0f) are converted to -INFINITY and INFINITY. A value clamped to one of the limits before is now still in one of the ranges when the parametric mask parameters are open-ended.

2. Removed the offset, because it is now also handled by the pre-processing in blend.c (instead of changing the pixel values, the limits are changed).

rawfiner:
@aurelienpierre output masking can be sometimes useful, even if I do use input a lot more than output. Actually, I don't mind if output masking disappears.
Example where output can be useful: if you have a noisy image and you want to denoise a specific part (output will be less noisy than input so mask can be more precise).

MStraeten:
with LCh it was possible to mask skin tones from out of a slightly tinted grey background using the c slider (LCh Cvalue 24 vs 6) 
setting Cz =1 excludes both the greyish and skin tone. works as designed?

wpferguson:
@rawfiner thanks for the idea.  I hadn't thought of that. :)

There's more conversation about this subject in this [pixls.us thread](https://discuss.pixls.us/t/user-feedback-requested-on-masking-and-blending-parameters/21085)

MStraeten:
value range and step size of Jz and Cz might be improved. Stepsize 1 in the scale is too coarse to use for paramtric masks. Just switching to log improves the range, but the difference remains very coarse
e.g.
README-eg1-4
before input profile:
top left (pure red) - Cx=6.73
bottom left (white) - Cx=0.3
after input profile
top left (pure red) - Cx=2.07
bottom left (white) - Cx=0.53

BTW: shortkey "m" switches over to map display most times - since the mouse must be on the slider to toggle mask ...

aurelienpierre:
JzAzBz derivates from XYZ space, and can't be right before input profile BTW. The numbers JzAzBz produce are merely indicative, they will be inconsistent.

MStraeten:
error building for osx:
```
/Users/martinstraeten/src/darktable_debug/src/common/color_picker.c:57:67: error: no member named 'work_profile_info' in
      'struct dt_iop_buffer_dsc_t'
  const dt_iop_order_iccprofile_info_t *const work_profile = dsc->work_profile_info;
                                                             ~~~  ^
/Users/martinstraeten/src/darktable_debug/src/common/color_picker.c:95:67: error: no member named 'work_profile_info' in
      'struct dt_iop_buffer_dsc_t'
  const dt_iop_order_iccprofile_info_t *const work_profile = dsc->work_profile_info;
                                                             ~~~  ^
```

TurboGit:
@hlecleme @aurelienpierre : this PR does not compile with recent changes. can you fix ? TIA.

johnny-bit:
Weird... CI should caught any compilation error...

TurboGit:
> Weird... CI should caught any compilation error...

But this PR has been broken by a recent commit on master. I suppose you don't re-run every checks for all opened PR ?

aurelienpierre:
Mandatory note : this will slightly break backward compatibility.

If you try the integration test, the test `0000-noop` will produce : 
```
Test 0000-nop
      Image mire1.cr2
      CPU & GPU version differ by 24145 pixels
      Max  dE         2.0689
      Mean dE         0.0002
      Count above max 2
  FAILS: image visually changed
         (1798 pixels changed)
```

For the blending test, with darktable 3.2.1 : 
```

Test 0033-blending-modes-uniform
      Image mire1.cr2
      CPU & GPU version differ by 1.45631e+06 pixels
      Max  dE         0.0000
  OK


Total test 1
Errors     0

```

Because the GPU code was cleaner and more straightforward, the choice has been to sync CPU with GPU logic. However, the result is slightly different from previous version:

```
Test 0033-blending-modes-uniform
      Image mire1.cr2
      CPU & GPU version differ by 42083 pixels
      Max  dE         24.6516
      Mean dE         1.1656
      Count above max 784619
  FAILS: image visually changed
         (1.44977e+06 pixels changed)


Total test 1
```

So the consistency between GPU and CPU has been improved, but the result has changed.

Here are the diffs:

CPU curent to GPU current:
diff-cl

CPU current to CPU previous:
diff

TurboGit:
Sorry, again a conflict in colorout.c to be resolved.

hlecleme:
@MStraeten The default scaling of Jz and Cz has been adapted. Does this work better for you?

TurboGit:
@aurelienpierre : Not sure why 0000_nop is changing on your side (or maybe a diff from 3.2.1 to current master).

Anyway, on my side:

With current master I have a diff since some time, no big deal as we are speaking of 3364 pixels, but for reference here it is. All other tests are OK.

```
Test 0034-blending-modes-parametric
      Image mire1.cr2
      CPU & GPU version differ by 560145 pixels
      Max  dE         8.5868
      Mean dE         0.0168
      Count above max 3364
  FAILS: image visually changed
         (86043 pixels changed)
```

When I apply this PR I have only two failing tests on my side, not surprisingly those are using a blending. The results are:

```
Test 0033-blending-modes-uniform
      Image mire1.cr2
      CPU & GPU version differ by 42217 pixels
      Max  dE         6.8790
      Mean dE         1.1654
      Count above max 784509
  FAILS: image visually changed
         (1.44961e+06 pixels changed)

Test 0034-blending-modes-parametric
      Image mire1.cr2
      CPU & GPU version differ by 2.78662e+06 pixels
      Max  dE         54.9481
      Mean dE         13.1735
      Count above max 2507561
  FAILS: image visually changed
         (2.78626e+06 pixels changed)
```

The diff output for 0033 (corresponds to what you describe above, and probably OK to update the expected output):

diff-small
 
The diff output for 0034 is certainly more worrisome and needs more investigation as the output is very different:

diff-small

The diff between CPU & GPU:

diff-cl-small

And finally a screenshot of the image showing that the black are really stronger and the white mostly over burned:

image

So to me there is something really broken for this test.

TurboGit:
If it can help, the center image seems ok, but the preview in the duplicate manager and in the navigation widget (top left) are broken:

image

aurelienpierre:
I have designed the test so we have clipped pixels in the image, to assess the consistency of clipping at the masking level (which was a wrong thing to do in the first place) between both version. @hlecleme have you removed the clipping to 100% in the legacy code ?

hlecleme:
Sorry, the differences were due to some regressions after the optimization of the masking in the (old) RGB and Lab.
The legacy RGB and Lab blending operators are still clipped like before.

Now, I have the following results for the tests:
```
Test 0033-blending-modes-uniform
      Image mire1.cr2
      CPU & GPU version differ by 42207 pixels
      Max  dE         6.8790
      Mean dE         1.1654
      Count above max 784510
  FAILS: image visually changed
         (1.44961e+06 pixels changed)

Test 0034-blending-modes-parametric
      Image mire1.cr2
      CPU & GPU version differ by 30853 pixels
      Max  dE         2.6604
      Mean dE         0.0488
      Count above max 116
  FAILS: image visually changed
         (543629 pixels changed)
```

Diff output for 0034 (resized):
diff-small

Difference between CPU and GPU (resized):
diff-cl-small

src/develop/blend_gui.c
```diff
@@ -1217,6 +1340,124 @@ gboolean blend_color_picker_apply(dt_iop_module_t *module, GtkWidget *picker, dt
...
+
+  // add a section to switch blending color spaces
+  const dt_develop_blend_colorspace_t module_cst = dt_develop_blend_default_module_blend_colorspace(module);
+  const dt_develop_blend_colorspace_t module_blend_cst = module->blend_params->blend_cst;
+  if(module_cst == DEVELOP_BLEND_CS_LAB || module_cst == DEVELOP_BLEND_CS_RGB_DISPLAY
+      || module_cst == DEVELOP_BLEND_CS_RGB_SCENE)
+  {
+    char *markup;
+
+    mi = gtk_menu_item_new_with_label(_("default blend colorspace"));
```
TurboGit, line 1406: IIUC, this reset the blending to the default for the module, right?

So maybe use : "reset to default blend colorspace" or "select default blend colorspace" ?

TurboGit:
I have just a (very) minor comment especially in contrast with the amount of work done here! The work here is just huge!

I must be honest I have checked that there is nothing obviously wrong (upgrading blending is one). I trust Aurélien for the maths in there and I may well have overlooked something.

Anyway, let me know what you think about my proposal for the string.

I also see a CSS tweak needed to avoid the new blend menu to be stuck on the right, we need a little margin, but this will be done after this is merged.

TurboGit:
Something else to look at, on the console I have:

```
*** BUG ***
In pixman_region32_init_rect: Invalid rectangle passed
Set a breakpoint on '_pixman_log_error' to debug
```

and

```
[dt_ioppr_transform_image_colorspace] invalid conversion from 2 to -1
[_transform_matrix] invalid conversion from 1 to -1
[dt_ioppr_transform_image_colorspace] invalid conversion from 1 to -1
[_transform_matrix] invalid conversion from 2 to -1
[dt_ioppr_transform_image_colorspace] invalid conversion from 2 to -1
```

TurboGit:
I cannot always reproduce but from time to time I have discrepancies between the center area and the previews.

Likewise I cannot reproduce the error in the console as reported above. So let's forget about this, we will certainly find a scenario to reproduce at some point and be able to fix. A larger field testing will certainly helps.

data/themes/darktable.css
```diff
@@ -960,31 +960,35 @@ notebook tabs,
...
   min-width: 1.55em;
   padding: 0.45em;
-  margin: 0;
-  border: 0;
+}
+
+#blending-tabs #dt-button
+{
+	font-size: 0.8em;
+	padding: 0.1em;
```
TurboGit, line 982: Adding:

`margin: 0 0.5em;`

Avoid the button to be stuck on the right.

aurelienpierre (approved):
All good for me, aside from 2 minor changes !

Well done, congrats, and thank for that huge piece of work ! darktable is now 99.99% scene-referred \o/

TurboGit:
> darktable is now 99.99% scene-referred \o/

Only ? We are missing 0.01 % :+1: 

@aurelienpierre : A question, should we select the scene blending (RGB scene) mode when using the scene referred workflow ?

aurelienpierre:
I'm just keeping a margin of error, in case I forgot something.

> A question, should we select the scene blending (RGB scene) mode when using the scene referred workflow ?

Yes, because :
1. the HSL masking options from the RGB display blending do not work if RGB > 1, they would produce negative saturation,
2. most blending modes, either from RGB display or Lab, clip pixels RGB > 1 to 1 after the blending (for no reason at all, pixel values are no concern for the alpha blending, only opacity needs to be > 1).

TurboGit:
Ok, but my question was more do we want to do that automatically ? Currently the default selected modes are Lab or "RGB display".

TurboGit:
@hlecleme : we are almost ready for merging. hopefully you'll be able to fix the remaining issues soon. TIA.

Summary:
- 2 minor comments from Aurélien
- A comment about the label for selecting default blending from me
- A proposal fix for the CSS

data/themes/darktable.css
hlecleme: Done thanks. I would have preferred the button to take the whole width (i.e. without spaces on its side), but it seems that increasing the width changes the size of the icon too...

hlecleme:
@aurelienpierre @TurboGit A big thank you for the time taken to review and test this PR! :smiley:

src/develop/blend_gui.c
```diff
@@ -1616,6 +1908,18 @@ void dt_iop_gui_init_blendif(GtkBox *blendw, dt_iop_module_t *module)
       gtk_box_pack_start(GTK_BOX(bd->blendif_box), GTK_WIDGET(slider_box), TRUE, FALSE, 0);
     }
 
+    bd->channel_boost_factor_slider = dt_bauhaus_slider_new_with_range(module, 0.0f, 3.0f, .02f, 0.0f, 3);
+    dt_bauhaus_slider_set_format(bd->channel_boost_factor_slider, _("%.2f EV"));
+    dt_bauhaus_widget_set_label(bd->channel_boost_factor_slider, NULL, _("boost factor"));
```
dterrahe, line 1913: dt_bauhaus_widget_set_label now needs to receive an english label, so use N_("boost factor")

dterrahe: also, section is N_("blend")

TurboGit: I have fixed that. And one other instance. Thanks for catching !

TurboGit:
@hlecleme : can you propose some text for the RELEASE_NOTES.md file ? for this and if appropriate for other parts of the work you did on this. TIA.

hlecleme:
@TurboGit Created PR #6911 for this. Is that the correct way to update RELEASE_NOTES.md?

TurboGit:
Yes, that's the way. Thanks a minor comment before merging.

PR #4697, opened 2020-04-12
snapshot: fix rotation click action

elstoc:
fix the snapshot module to only rotate when clicking on centre icon
(respect line orientation)

### DIFF
--- a/src/libs/snapshots.c
+++ b/src/libs/snapshots.c
@@ -276,7 +276,7 @@ int button_pressed(struct dt_lib_module_t *self, double x, double y, double pres
     if(which == 1
        && (((d->vertical && xp > d->vp_xpointer - hhs && xp < d->vp_xpointer + hhs) && yp > 0.5 - hhs
             && yp < 0.5 + hhs)
-           || ((yp > d->vp_ypointer - hhs && yp < d->vp_ypointer + hhs) && xp > 0.5 - hhs && xp < 0.5 + hhs)))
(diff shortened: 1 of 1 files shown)


elstoc:
don't merge this yet - there's a better way...

TurboGit:
> fix the snapshot module to only rotate when clicking on centre icon

Isn't that already what is done ? Can you clarify ?

elstoc:
Hard to easily put into words what I'm doing but I'll try!

Current functionality is that if you move the line up (so that it's not central) and then click on the rotate icon and repeatedly click without moving the mouse, it keeps rotating as long as you don't move the mouse. When you've moved 180 degrees the line is back in its starting position (above centre).

If you do the same thing but first moving the line left of centre then after 180 degree rotation the line is central again and you have to move it again.

My fix above changes so that both versions revert the line back to being central after 180 degrees. But IMO the better functionality is that in the first paragraph. It seems this was accidental before (because of inconsistency in the if statement). I want to change so it works that way all the time.

elstoc:
See my last commit which should work as I've described.

TurboGit:
I see now! Indeed, never done that. Fix is good. Thanks.

elstoc:
Apologies @TurboGit I force-pushed a new commit without amending again! This doesn't actually contain the fix I intended (or do what the commit says). I'll push a new one that does.

elstoc:
See #4701

PR #4701, opened 2020-04-12
snapshot: rotation fix

elstoc:
when clicking on the rotate button in the snapshot module, if the user
does not move their mouse by more than the size of the rotate button
before clicking again, the line will rotate again,
even if the user is no longer on the rotate button.

prevents the dividing line re-centering after rotation

### DIFF
--- a/src/libs/snapshots.c
+++ b/src/libs/snapshots.c
@@ -66,7 +66,7 @@ typedef struct dt_lib_snapshots_t
 
   /* change snapshot overlay controls */
   gboolean dragging, vertical, inverted;
-  double vp_width, vp_height, vp_xpointer, vp_ypointer;
+  double vp_width, vp_height, vp_xpointer, vp_ypointer, vp_xrotate, vp_yrotate;
   gboolean on_going;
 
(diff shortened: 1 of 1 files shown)


elstoc:
Amended to give some margin for small mouse movements.

src/libs/snapshots.c
```diff
@@ -293,6 +296,8 @@ int button_pressed(struct dt_lib_module_t *self, double x, double y, double pres
       d->dragging = TRUE;
       d->vp_ypointer = yp;
       d->vp_xpointer = xp;
+      d->vp_xrotate = 0;
```
TurboGit, line 299: 0.0

Likewise next line.

elstoc:
Thanks @TurboGit I've made the changes you requested. The issue with the previous PR is that I amended my new commit to describe the functionality in this PR but I neglected to add the file before amending the commit (I really must stop doing that!) so the change log didn't accurately describe the PR, which only really fixed the first line of that change log "fix the snapshot module to only rotate when clicking on centre icon (respect line orientation)"

PR #4703, opened 2020-04-13
Speed & clean up purging script

l0b0:
Done:

1. Run fewer & faster queries (cuts down the runtime by about 60x on my machine with about 50k images).
1. Run fewer slow `read` commands.
1. Fix some shellcheck lints, inconsistencies etc..


l0b0:
Pushed as wrong user.

PR #4704, opened 2020-04-13
Speed & clean up purging script

l0b0:
Done:

1. Run fewer & faster queries (cuts down the runtime by about 60x on my machine with about 50k images).
1. Run fewer slow `read` commands.
1. Fix some `shellcheck` lints, inconsistencies etc.

### DIFF
--- a/tools/purge_non_existing_images.sh
+++ b/tools/purge_non_existing_images.sh
@@ -100,7 +100,7 @@ done
 
 if [ "$dryrun" -eq 0 ]
 then
-    # delete now-empty filmrolls
+    # delete now-empty film rolls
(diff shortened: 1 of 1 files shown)


tools/purge_non_existing_images.sh
```diff
@@ -5,107 +5,106 @@
...
 #
 
-if ! which sqlite3 > /dev/null; then
+if ! command -v sqlite3 >/dev/null
+then
     echo "error: please install sqlite3 binary".
     exit 1
 fi
 
-if pgrep -x "darktable" > /dev/null ; then
```
TurboGit, line 13: frankly I do prefer the then (and do below) at end of line.

TurboGit:
Looks ok to me apart from the indent comment. But I'll merge as-is anyway. Thanks.

PR #4705, opened 2020-04-13
Increase the falloff range in vignetting module

jenshannoschwalm:
As commented the falloff range might also need a 0-200 range.

From reading the algorithm it is wise not to extend the saturation range.


TurboGit:
Works for me! Thanks.

PR #4707, opened 2020-04-13
darkroom: keyboard shortcut to show/hide drawn masks

### LINKED ISSUE #2354: Hotkey to show/hide shapes and drawn masks
**Is your feature request related to a problem?**
In many modules, especially in 'retouch' and 'spot removal' I draw shapes into the photo that I edit.

To evaluate the results of the drawn masks and shapes often I turn the display of drawn shapes off and then on again.

To do this I use the button in the currently used module:

current process to achieve this:

Usually this starts when I just have drawn a shape on a picture, the mouse pointer is somewhere near the center of the screen.

- move the mouse pointer to the right side of the screen
- click the button to hide the shapes
- move the mouse back to the image
- maybe drag the image around and change zoom
- move the mouse pointer to the right side of the screen
- click the button to show the shapes again
- move the mouse back to the image
- continue work

To decrease the amount of needed mouse movements and clicks I'd like to have an assignable hotkey that toggles the button described above on and off.

The new process would be (from the same starting point)

- click hotkey to hide shapes
- maybe drag the image around and change zoom
- click hotkey to show shapes
- continue work

This is much easier.

The 'spot removal' module already has got an assignable key for that. In my opinion this should be implemented for all modules with drawn masks or shapes.
Moreover I think that for easier maintenance this hotkey should be one for all modules, not module-specific.

elstoc:
adds a new keyboard shortcut to show/hide blendop drawn masks

Partially resolves #2354, with the exception of the spot removal and
retouch modules, which handle drawn masks differently and need separate
consideration.

### DIFF
--- a/src/views/darkroom.c
+++ b/src/views/darkroom.c
@@ -1933,6 +1933,49 @@ static gboolean _overlay_cycle_callback(GtkAccelGroup *accel_group, GObject *acc
   return TRUE;
 }
 
+static gboolean _toggle_mask_visibility_callback(GtkAccelGroup *accel_group, GObject *acceleratable, guint keyval,
+                                             GdkModifierType modifier, gpointer data)
+{
+  if(darktable.gui->reset) return FALSE;
+
+  dt_develop_t *dev = (dt_develop_t *)data;
+  dt_iop_module_t *mod = dev->gui_module; 
+
+  //retouch and spot removal module use masks differently and have different buttons associated
+  //keep the shortcuts independent
+  if(mod && strcmp(mod->so->op, "spots") != 0 && strcmp(mod->so->op, "retouch") != 0)
+  {
+    dt_iop_gui_blend_data_t *bd = (dt_iop_gui_blend_data_t *)mod->blend_data;
+
+    const int reset = darktable.gui->reset;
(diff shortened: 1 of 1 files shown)


elstoc:
It would be nice to get the spot removal and retouch modules using this same shortcut but I can't work out how to do the module-specific functionality that's associated with the show/hide buttons in those modules.

TurboGit:
Works for me! Thanks. Note that for the liquify module you can right click on the canvas to show/hide the masks. For retouch I don't think there is a way.

elstoc:
There is a shortcut for spot removal but not for retouch. I'm sure one could be added but thought it better to keep that separate.

GrahamByrnes:
@elstoc you might be interested in the couple of videos I posted down the bottom of #4702 
It wasn't what I was trying to show, but you'll see that in one I deleted all the shapes from the mask manager, then in the second all of them are still there. Any wisdom you could bring to mask management would be welcome.

elstoc:
@GrahamByrnes my wisdom remains small I'm afraid. I didn't need to know much about masks for this one since it mostly involved replicating some functionality already attached to a button. So mostly copy-and-paste really.

PR #4708, opened 2020-04-13
thumbs : fix calculation of image margin if specified separatly

AlicVB:
otherwise setting top or left values lower that bottom or right ones doesn't works


TurboGit:
Works for me. Thanks!

PR #4710, opened 2020-04-13
styles: add append/overwrite mode

### LINKED ISSUE #2396: Lighttable: Option to append/overwrite style
I've got several artistic styles from dtstyle.net. After importing some images I want to try out how they look like.
For this I use the lighttable view. In the 'styles' section I doubleclick on a style and then it is applied to the selected image(s), which means it is appended to the image's history.

When I want to try another style I have to open the 'history stack' and click on 'discard' to remove that style.
After confirming I can doubleclick on the next style.

In the 'styles' section I want to have the option to 'append' or 'overwrite' the image history (like in the copy&paste options) when I apply a style.
With the option set to 'overwrite' I can doubleclick on a style to apply, when I want to try the next style I just have to doubleclick on the next one.

This would speed up the selection process for a certain style.

### LINKED ISSUE #4556: Application of a Style is not clearcut
I would expect applying a style would be well-defined in terms of what it does to the image but it seems to depend on what modules you have just been using.

The manual says a style is about "storing a history stack as a style, and applying it to other images".

**To Reproduce**
Have a style and check what it contains.
Edit an image and use a module not in the style.
In Lightable, apply the style to the image.
Go back into Darktable and note the module you just used is now applied to the image.

**Expected behaviour**
When you apply a style, nothing outside of the style contents should be applied.

**Screenshots**
Screenshot from 2020-03-26 12-42-56
I found Perspective was applied to my image even though it is not part of my style as the above image shows.

 - Darktable Version: 3.0.1+dirty
 - OS: Ubuntu 18.04
 -  none

elstoc:
adds an append/overwrite mode selection to the styles module which
functions the same as the similar mode button on the history stack
module (defaults to 'append')

Resolves #2396. Resolves #4556

### DIFF
--- a/src/common/styles.c
+++ b/src/common/styles.c
@@ -572,6 +572,8 @@ void dt_styles_apply_to_selection(const char *name, gboolean duplicate)
   const dt_view_t *cv = dt_view_manager_get_current_view(darktable.view_manager);
   if(cv->view((dt_view_t *)cv) == DT_VIEW_DARKROOM) dt_dev_write_history(darktable.develop);
 
+  const int mode = dt_conf_get_int("plugins/lighttable/style/applymode");
+
   /* for each selected image apply style */
   sqlite3_stmt *stmt;
   dt_undo_start_group(darktable.undo, DT_UNDO_LT_HISTORY);
@@ -580,6 +582,8 @@ void dt_styles_apply_to_selection(const char *name, gboolean duplicate)
   while(sqlite3_step(stmt) == SQLITE_ROW)
   {
     const int imgid = sqlite3_column_int(stmt, 0);
+    if(mode == DT_STYLE_HISTORY_OVERWRITE)
+      dt_history_delete_on_image_ext(imgid, FALSE);
     dt_styles_apply_to_image(name, duplicate, imgid);
     selected = TRUE;
   }
--- a/src/libs/styles.c
+++ b/src/libs/styles.c
@@ -15,6 +15,7 @@
     You should have received a copy of the GNU General Public License
     along with darktable.  If not, see <http://www.gnu.org/licenses/>.
 */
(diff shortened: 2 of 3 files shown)


src/common/styles.c
```diff
@@ -580,6 +582,8 @@ void dt_styles_apply_to_selection(const char *name, gboolean duplicate)
   while(sqlite3_step(stmt) == SQLITE_ROW)
   {
     const int imgid = sqlite3_column_int(stmt, 0);
+    if(mode == 1)
```
TurboGit, line 585: Can you please add an enum in styles.h with DT_STYLE_HISTORY_APPEND and DT_STYLE_HISTORY_OVERWRITE  and use DT_STYLE_HISTORY_OVERWRITE instead of 1 here? It is always better to use named number as it makes reading the code clear.

src/libs/styles.c
```diff
@@ -384,6 +385,12 @@ static void _styles_changed_callback(gpointer instance, gpointer user_data)
   _gui_styles_update_view(d);
 }
 
+static void applymode_combobox_changed(GtkWidget *widget, gpointer user_data)
+{
+  int mode = dt_bauhaus_combobox_get(widget);
```
TurboGit, line 390: const

elstoc:
Done

TurboGit:
> Done

Lighting speed :) I'll test now.

TurboGit:
Ok works for me. Thanks.

Not related to your change, but in append mode (or in current dt) applying a style does not refresh the thumbs. Let me know if you plan to have a look otherwise @phweyland is probably a good candidate as it may be related to the hash work.

elstoc:
Might be better for @phweyland to look at it - or at least I might need some help to do it myself.

PR #4712, opened 2020-04-13
Unify timestamp

jpverrue:
Change film_rolls.datetime_accessed to be consistent with other timestamps.

Column name changed : datetime_accessed --> access_timestamp
Data type changed : char(20) --> integer
Data in library.db updated


TurboGit:
Looks good to me and working fine. Thanks!

Issue #3859, opened 2019-12-25
Darkroom active window focus  issue.


chhil:
Have some text in clipboard, just ctrl+c some text from anywhere.
Put cursor in search module.
Mouse select a picture in the filmstrip at bottom, then ctrl+c to copy its history  , select another pic where you want paste, ctrl+v.
Nothing happens to the picture, but you will notice that the search module text entry will have the copied text.

A similar thing happens using short cut ctrl+e when in the search module, no export happens. One needs to physically click in the main image window in darkroom to export the image.

I had "abc" on the clipboard, then did one ctrl+c, and multiple  ctrl+v's resulting in multiple abc's in the search text window.
![image](https://user-images.githubusercontent.com/1119439/71432187-a8df2880-26fd-11ea-839d-d12e01521e1c.png)

Since we are on clipboard issue, I have seen windows clipboard stops functioning at times when dt is running, I have to close DT to get back control of ctrl+c ctrl+v on my windows system for other apps.

I believe this has happened after the change that redirects key strokes per window.

darktable-3.1.0+9~g962bc9ae3-win64

chhil:
Still an issue in the current dev version.

dterrahe:
This may not be optimal, but it is intended behavior; whenever a cursor is in a text entry field, any darktable shortcuts are disabled. Otherwise you wouldn't be able to search for "lens correction" because typing L would take you to lightroom. Copying history has been assigned the same shortcuts that gtk uses for text entry, CTRL-C/CTRL-V, assuming this helps people remember, but the copy doesn't actually pass via the normal clipboard.

Presumably what you want is that, while in text entry mode, only single keys and those used in standard gtk text entry are disabled as dt shortcuts. I could consider this as a feature request for #8078 (with fairly low priority if I'm honest) if you could specify an uncontroversial list of combinations that should be exempt from suppression.

Alternatively it may be possible in #8078 to have the Entry widget take a first shot at interpreting key presses (using gtk_widget_event?) and only process them in the shortcut dispatcher if the Entry doesn't absorb them. Even if it works, there may be side effects to that approach that make it less predictable than the current blanked disabling of dt shortcuts, but it may be worth experimenting.

johnny-bit:
@dterrahe - i think with #8078 and it's searchentry the problem is "gone" due to how search entry behaves... so IMO in my testing I'd consider #8087 "fixes" this issue. Would you agree?

dterrahe:
[...]

Issue #3543, opened 2019-11-23
not possible to zoom in and out with alt+f1/f3 if cursor is in the modules search field


blitzgneisserin:
**Describe the bug**
I am not sure if this is really a bug but I find it rather annoying: If the cursor is in the modules search field it is not possible to zoom in and out of the picture with alt+f1/f3. Mostly I need to go to a different tab so zooming with the keyboard is re-enabled. Apparently the cursor stays active even if I work in a module.

**To Reproduce**
1. open a photo in darkroom
2. click into the modules search field and type something
3. try to zoom in and out with alt+f1/f3 (zooming impossible)
4. click on a different tab, e.g. color
5. try to zoom with alt+f1/f3 (zooming possible)

**Expected behavior**
zooming with the keyboard should not be blocked by the cursor in the modules search filed

**Platform (please complete the following information):**
 - OS: MX Linux 19
 - Version 3.0rc1

PR #4721, opened 2020-04-13
module search: lose focus when clicking module header

### LINKED ISSUE #4717: module search retains keyboard focus after selecting module
When using the iop module search, if you select one of the modules that appears in the search results, the module search retains keyboard focus, meaning it's not possible to use keyboard shortcuts without first pressing tab to take keyboard focus away from the search box.

This should not be necessary: the search box should lose keyboard focus as soon as a module is selected.

elstoc:
when a module is given gui focus, temporarily give it the keyboard focus
as well in order that the module search input box loses focus.

Resolves #4717


elstoc:
Might also partially address #3543 and #3859

TurboGit:
Works for me, thanks!

PR #4723, opened 2020-04-14
add finetuned whitebalance presets for Nikon Z6

st-binder:
add finetuned whitebalance presets for Nikon Z6, add firmware version


johnny-bit:
Cool!

It looks like you haven't done "Flash" preset in finetune (does Z6 allow to finetune flash?). Also it looks like you've done it painstakingly by hand (like old wiki suggested). Curently if you still have the files, you can run `extract_wb` from current master (or prefferably from PR #4671) in the directory with images like `extract_wb *` and it'll process all files and neatly sort, assign finetuning, and format entries for you :)

st-binder:
flash finetuned wb presets added.
sorry for doing it by hand and the old script but the ruby script doesn't work for me.

johnny-bit:
by "ruby script doesn't work" you mean that you can't get ruby to run, or there's nokogiri error or something else at play?

st-binder:
I'm using archlinux on my pc. After installing ruby and nokogiri packages the script crashes with multiple error messages and the simple 'extract_wb_from_images.sh' script doesn't allow direct integration of the results in the source code, so I used the old perl script.
And sorry, I deleted the raw files of the test due to there size after getting the results.

johnny-bit:
yeah, the idea for `extract_wb_from_images.sh` is to generate tiny package that can be then parsed by someone that has working `extract_wb` setup.

can you jump on #darktable on freenode or over at discuss.pixls.us? I'd like to know what fails where and how. But honestly - ruby as a lang for processing that kind of data wouldn't be in my top 10 choices :P I just fixed what was already present, but I may be able to translate that in some more popular or wide-spread language like sh script (dunno if possible due to some replacements), perl or python (or even php :P)

If you have package created by `extract_wb_from_images.sh` i can parse it for you.

st-binder:
It was a lot of work yesterday and I would'nt do it again. I'm using a custom build of darktable so I needed the direct integration of the presets. So it's possible to verify the build and functionality for me before pushing it to the public repo. The source has been formatted the same way as other wb pull requests, like yours for the D7000. Now it's up to you to use it for darktable :-)

TurboGit:
Thanks!

PR #4724, opened 2020-04-14
update redefined color matrix for Nikon Z6

st-binder:
I found that the color matrix for the Nikon Z6 used in darktable differs from the one used by Adobe DNG Converter. Tested with Adobe DNG converter 12.1 and 12.2. Nikon Z6 firmware was 3.0.


st-binder:
here the exiv2 output:
exiv2 -pt nikon-z6.dng 2>/dev/null | grep ColorMatrix2
Exif.Image.ColorMatrix2                      SRational   9  9943/10000 -3269/10000 -839/10000 -5323/10000 13269/10000 2259/10000 -1198/10000 2083/10000 7557/10000

LebedevRI:
Adobe does sometimes change color matrixes after the initial support was added, so this is possible.

TurboGit:
So this is good to go. Thanks.

LebedevRI:
> So this is good to go. Thanks.

I didn't review this.

parafin:
Merging this was probably a mistake since it retroactively changes already developed images.

TurboGit:
> > 
> 
> I didn't review this.

Sorry Roman, I read your previous message as an approval. Let me know if I need to revert.

TurboGit:
> Merging this was probably a mistake since it retroactively changes already developed images.

Probably so, but I suppose very little. I'll let Roman cast the final vote for this to be in or out.

TurboGit:
I have reverted this at this point. @LebedevRI let me know if this needs to be added again or not. Thanks.

st-binder:
In my humble opinion, it should be better not to include a flawed color matrix in the program. There is a reason that the former color matrix was updated by Adobe. But it's understandable that changes in already developed images are a problem.

parafin:
Can you demonstrate how it is flawed? Is the reason Adobe changed the matrix stated anywhere? Maybe it will be easier to make decision with this information.

st-binder:
For my understanding, the color matrix is a simple way to translate the recorded values of the sensor to a linear rgb space (measured icc would be better). So there should be an optimization process involved to reduce the propagated error, there is a reason that the color matrix has changed, i.e. less propagation error. But as I have no direct contact to adobe engineers I can't know the actual reason. Recent DNG Converters use the new color matrix and if darktable respects the embedded matrix in DNG files there would be a difference in color presentation depending whether the NEF or the DNG format is used for the workflow. All I can tell is that the new color matrix changes the color only slightly. On my color calibrated monitor I have the subjective visual impression that the new one removes a slight yellowish cast.

trougnouf:
@LebedevRI can you review ? I think users (such as myself and @st-binder ) would like to have the most accurate color matrix and that it's what Adobe's update would aim to provide.

LebedevRI:
I think i don't care nowadays. I don't have that camera, and i don't really use darktable.
This is quite obviously a bug in darktable, someone on their side maybe should fix it.
Or maybe not.

TLDR: i'm fine with this either way.

trougnouf:
Thank you for the quick and effective response @LebedevRI 
@TurboGit Can you re-add the updated color matrix?

TurboGit:
@trougnouf : Fine with me, I have reverted the revert :)

trougnouf:
Thank you @TurboGit !

PR #4726, opened 2020-04-14
map fix

### LINKED ISSUE #4716: Map - add selected images from filmstrip to map works only for the first image
Select several images in filmstrip, and drag them to map.
Only the first images is set on map, instead of all selected.

 - Darktable Version: master
 - OS: debian 10

phweyland:
- multiple images drag & drop add / remove (resolves #4716)
- geotag undo (common to all views now)

Thanks to @AlicVB for the excellent starting point.


phweyland:
Some remarks:
- on the map, I've let the single click select one image, and added shift-click to select the group of images. Is that ok ? could be the opposite as well.
- I haven't found any usage of DND_TARGET_URI or "text/uri-list" drag & drop. then I haven't touched this part. Any indication about the goal or intention of this ?

AlicVB:
>     * I haven't found any usage of DND_TARGET_URI or "text/uri-list" drag & drop. then I haven't touched this part. Any indication about the goal or intention of this ?

iirc it allow to drop images in another soft

phweyland:
-

> > ```
> > * I haven't found any usage of DND_TARGET_URI or "text/uri-list" drag & drop. then I haven't touched this part. Any indication about the goal or intention of this ?
> > ```
> 
> 
> iirc it allow to drop images in another soft

I see (I've tried the other way around and nothing was triggered), thanks. 
It works with files program, but GIMP returns errors opening the file. I see other side effects.
Not sure this actually used (and useful)...

AlicVB:
iirc there's someone who use it with a soft for dealing with more precise exif and another about a gps soft too...

TurboGit:
For me the selection does not work on the map view. The click to select the first image on the filmstrip does nothing, so the selection can only start from the image selected from lighttable or darkroom and span from it. Can you reproduce ?

phweyland:
Not sure to understand the problem, but that may be related to the new behavior of filmstrip.

There are two ways.
1. filmstrip to map. If images selected click anywhere on the selection and drag. Otherwise select an image (alt-click) and complete the selection with ctrl or shift click. Then drag and drop.
2. map to filmstrip (to remove images from map). Simple click on an image stack select the first one and then you can drag it. If you drop on map it is moved, on filmstrip it is removed. Shift-click to selected the stack of images instead, drag, release shift and drop on map or on filmstrip (shift release must occur before the mouse release. I haven't not found the way to change that).

phweyland:
Please hold on.
Found an issue. Everything goes well for me when I drag and drop from map or from filmstrip. But if I remove some images and click on filmstrip the images come back to map...

phweyland:
> Found an issue. Everything goes well for me when I drag and drop from map or from filmstrip. But if I remove some images and click on filmstrip or move the map the images come back to map...

Since I've rebuilt, I haven't succeeded in reproducing this issue. Everything is doing well.


>  Can you reproduce ?

This one neither...

phweyland:
> iirc there's someone who use it with a soft for dealing with more precise exif and another about a gps soft too...

OK. This should work as before. So let's these potential users request change if necessary.

TurboGit:
Ok, sounds good. Thanks.

PR #4727, opened 2020-04-14
Finetuned WB presets for Olympus E-M5 Mark II

johnny-bit:
Samples provided by AxelG over at pixls.us


TurboGit:
Looks ok. Thanks!

PR #4729, opened 2020-04-14
retouch: various fixes and enhancements

elstoc:
added a new (local) keyboard shortcut to show / hide drawn shapes

resolved the following issues:

- clicking on the shape creation buttons while the module was disabled
  previously only toggled the shape button (user had to then enable
  the module to actually create a shape)

- show/edit shapes button was being activated inappropriately (when shapes
  were not being shown, when selecting earlier history items, when resetting
  module parameters and sometimes when adding a new shape)

- clicking on the show/edit shape button while in shape creation mode wasn't
  working - it now toggles shape creation modes off and switches to edit mode

### DIFF
--- a/src/iop/retouch.c
+++ b/src/iop/retouch.c
@@ -1026,6 +1026,13 @@ static int rt_shape_is_beign_added(dt_iop_module_t *self, const int shape_type)
 
 static gboolean rt_add_shape(GtkWidget *widget, const int creation_continuous, dt_iop_module_t *self)
 {
+  //turn module on (else shape creation won't work)
+  gtk_toggle_button_set_active(GTK_TOGGLE_BUTTON(self->off), TRUE);
+
+  //switch mask edit mode off
+  dt_iop_gui_blend_data_t *bd = (dt_iop_gui_blend_data_t *)self->blend_data;
+  if(bd) bd->masks_shown = DT_MASKS_EDIT_OFF;
+
   const int allow = rt_allow_create_form(self);
   if(allow)
   {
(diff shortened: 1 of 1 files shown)


elstoc:
N.B. The one issue I could not resolve was that if you have the retouch shapes in edit mode, click on an earlier history item, and then go back to the top of the stack, the 'show and edit shapes' button is active but the shapes are not shown. You have to click twice to get it back to being consistent.

The only resolution I could find for this problem would be to reset the current gui module to NULL whenever selecting items on the history stack but I wasn't sure whether that might cause problems for other modules (perhaps crop & rotate?). This solution would ensure that the 'show and edit shapes' button is not active.

TurboGit:
>     * clicking on the shape creation buttons while the module was disabled
>       previously only toggled the shape button (user had to then enable
>       the module to actually create a shape)

It was activated as soon a shape was actually drawn.

TurboGit:
Ok, looks good to me.

src/iop/retouch.c
```diff
@@ -3499,6 +3569,10 @@ void connect_key_accels(dt_iop_module_t *module)
 
   closure = g_cclosure_new(G_CALLBACK(_continuous_add_path_key_accel), (gpointer)module, NULL);
   dt_accel_connect_iop(module, "continuous add path", closure);
+
+  //show or hide
+  closure = g_cclosure_new(G_CALLBACK(_show_or_hide_accel), (gpointer)module, NULL);
+  dt_accel_connect_iop(module, "retouch show or hide", closure);
```
TurboGit, line 3575: I would rename this to "retouch show or hide shapes"

elstoc:
Agreed - I just blindly copied the one from spot removal without noticing it was a bit non-obvious what it did.

What do you think about removing gui focus when clicking the history stack? Too much just to fix a minor issue with retouch? I'm not sure there's another way to do it but it is a bit of an edge case.

TurboGit:
I would let the code as-is for now. I'm wondering why the registered callback when the togglebutton is set is not called. I would have supposed that calling gtk_toggle_set_active() would have the same effect than clicking on the button?

src/iop/retouch.c
TurboGit: In fact better to have "show or hide shapes"

TurboGit:
> I just blindly copied the one from spot removal without noticing it was a bit non-obvious

Right, "spot show or hide" is clear, we are speaking about spots. I have proposed another better choice.

elstoc:
There's a lot of odd stuff in retouch but it felt like picking it apart was probably going to do more harm than good so I tried to be surgical. Happy to leave the remaining fault for now since it'll probably never happen in active use.

TurboGit:
Looks better now. Thanks a lot for all your contributions.

TurboGit:
And I have finally unified the spot removal accel to use the same wording.

elstoc:
Might be worth a mention in the release notes - since that will have broken the key mapping for anyone using it already.

elstoc:
Thanks @TurboGit I see you added a section to the release notes but you referenced crop&rotate instead of spot removal.

TurboGit:
@elstoc : Indeed, fixed, thanks for catching that.

PR #4736, opened 2020-04-16
spot removal: various enhancements

### LINKED ISSUE #2342: Hide shapes in Spot removal module
After adding a spot to remove I want to take a look at the result. For the assessment  I need to hide the shape.
Other modules (like the 'retouch' module) have got a button to hide the used shapes. 

In the 'spot removal' module I have to deactivate the module. To hide the shapes I can collapse the module, to see the shapes again I can expand the module.
This behaviour is inconsistent to other modules with a dedicated button to show/hide the shapes

Please add a button that shows and hides all used shapes to the 'spot removal' module, see below

**Additional context**
grafik

elstoc:
a number of enhancements to the spot removal module, adapted from
similar functionality in the retouch module:

- new button added to show/hide shapes

- continuous shape creation mode added, which is now the default when
  clicking on the shape buttons. Can be switched to single shape
  creation with ctrl-click.

- new keyboard shortcuts added for continuous shape creation

- clicking on the shape creation buttons while the module is disabled
  now enables the module allowing immediate shape placement

Resolves #2342

### DIFF
--- a/src/develop/develop.c
+++ b/src/develop/develop.c
@@ -71,7 +71,7 @@ void dt_dev_init(dt_develop_t *dev, int32_t gui_attached)
 
   dt_image_init(&dev->image_storage);
   dev->image_status = dev->preview_status = dev->preview2_status = DT_DEV_PIXELPIPE_DIRTY;
-  dev->image_loading = dev->preview_loading = dev->preview2_loading = 0;
+  dev->image_loading = dev->preview_loading = dev->preview2_loading = dev->history_updating = 0;
   dev->image_force_reload = 0;
   dev->preview_input_changed = dev->preview2_input_changed = 0;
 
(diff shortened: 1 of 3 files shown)


elstoc:
Please note that this also implements some code that will allow resolution of the final retouch issue mentioned in #4729 (using the new ``dev->history_updating`` variable). Once this PR is merged I'll push another change to fix the retouch issue, to keep the two separate.

TurboGit:
Ok, works for me. Note that you do not have to spend too much time on this IOP has it is considered as somewhat deprecated and the preferred IOP is now module which support not only copy but heal mode which does better job.

One can also say that the spot removal is simpler for beginners but in that case we probably want to bring the heal tool there. That was my idea before I switched to Retouch module only.

That only to recall history about this module. But enhancement are always welcomed, so I'll merge this.

elstoc:
I understand. I did ask around on IRC before I worked on it and found a few who still prefer it for its simplicity. I certainly don't plan to do any more to it.

Thanks.

PR #4737, opened 2020-04-16
thumb overlays enhancements

### LINKED ISSUE #4545: Introduce toggle to turn off stars rating in filmstrip preview (darkroom)

### LINKED ISSUE #4689: Thumbnails should not use GtkOverlay
The new thumbnail code uses GtkOverlay to assemble its GUI. After consulting the GTK people in IRC they said that overlays are NOT meant to be used when the overlays overlap. They are only supposed to be put ontop the base widget. As a result it can happen that the image, which is quite big and NOT the base widget gets drawn over other overlays. Also, using regular widgets like GtkLabel is supposed to be a bad idea as these can have backgrounds that can get drawn and block some parts of the underlying stuff. I know that dt tries to remedy that in its CSS, but overall I think the design of the thumbnails needs to be revisited and everything has to be drawn in a single pass with cairo instead of using a GtkOverlay.

Once that is done we can also bring back the custom metadata overlay. I tried to add it in the current system but it doesn't work, the image always blocks the label I added.

AlicVB:
this fix #4545 
this fix #4689 
this close #4549 

Rework the thumbnails overlays. What change : 
- user can choose between 5 type of overlays via a popup associated to the star icon of the global toolbox
- overlays can be : none ; on mouse hover ; extended on mouse hover ; permanent ; permanent extended
- with the 2 "permanent" overlays, the space for overlays is "reserved" so the overlays doesn't overlap the image
- overlays settings are saved per thumbtable mode (filemananger, zoomable, filmstrip) and per thumbnail size (divided in categories)
- categories of sizes can be finetune in pref dialog by writing size values separated by | ex : "120|300" => category 0: sizes [0-120] ; category 1 : sizes [120-300] ; category 2 : sizes [300-...]
 - accels with no default key are available for switch overlays modes
- overlays modes and size categories can be used in css via specific class
- default overlays : none for size 0 ; mouse hover for size 1 (without background) ; permanent extended for size 2
- last (but not least) the extended overlay is now defined by a pattern in prefs like for the tooltip. Default value produce the same lines of text as currently.

I'm not sure how to add more flexibility...
Please test, especially to see if default overlays are ok for you.

### DIFF
--- a/src/libs/tools/global_toolbox.c
+++ b/src/libs/tools/global_toolbox.c
@@ -75,6 +75,13 @@ int position()
   return 1001;
 }
 
+static void _overlays_accels_callback(GtkAccelGroup *accel_group, GObject *acceleratable, guint keyval,
+                                      GdkModifierType modifier, gpointer data)
+{
+  dt_thumbnail_overlay_t over = (dt_thumbnail_overlay_t)GPOINTER_TO_INT(data);
+  dt_thumbtable_set_overlays_mode(dt_ui_thumbtable(darktable.gui->ui), over);
+}
+
 static void _overlays_toggle_button(GtkWidget *w, gpointer user_data)
 {
   if(!gtk_toggle_button_get_active(GTK_TOGGLE_BUTTON(w))) return;
@@ -393,7 +400,12 @@ void init_key_accels(dt_lib_module_t *self)
 {
   dt_accel_register_lib(self, NC_("accel", "grouping"), 0, 0);
   dt_accel_register_lib(self, NC_("accel", "preferences"), 0, 0);
-  dt_accel_register_lib(self, NC_("accel", "show overlays"), 0, 0);
+
+  dt_accel_register_lib(self, NC_("accel", "no overlays"), 0, 0);
+  dt_accel_register_lib(self, NC_("accel", "overlays on mouse hover"), 0, 0);
+  dt_accel_register_lib(self, NC_("accel", "extended overlays on mouse hover"), 0, 0);
+  dt_accel_register_lib(self, NC_("accel", "permanent overlays"), 0, 0);
+  dt_accel_register_lib(self, NC_("accel", "permanent extended overlays"), 0, 0);
 }
 
 void connect_key_accels(dt_lib_module_t *self)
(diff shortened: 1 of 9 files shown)


src/libs/tools/global_toolbox.c
```diff
@@ -37,14 +37,15 @@ DT_MODULE(1)
...
+  GtkWidget *over_popup, *over_label, *over_r0, *over_r1, *over_r2, *over_r3, *over_r4;
 } dt_lib_tool_preferences_t;
 
 /* callback for grouping button */
 static void _lib_filter_grouping_button_clicked(GtkWidget *widget, gpointer user_data);
 /* callback for preference button */
 static void _lib_preferences_button_clicked(GtkWidget *widget, gpointer user_data);
 /* callback for overlays button */
-static void _lib_overlays_button_clicked(GtkWidget *widget, gpointer user_data);
+// static void _lib_overlays_button_clicked(GtkWidget *widget, gpointer user_data);
```
TurboGit, line 48: To be removed, look like some missing clean-up.

src/libs/tools/global_toolbox.c
```diff
@@ -154,22 +236,6 @@ static void _lib_filter_grouping_button_clicked(GtkWidget *widget, gpointer user
...
-{
-  gboolean show = gtk_toggle_button_get_active(GTK_TOGGLE_BUTTON(widget));
-  if(show)
-    gtk_widget_set_tooltip_text(widget, _("hide image overlays"));
-  else
-    gtk_widget_set_tooltip_text(widget, _("show image overlays"));
-  dt_conf_set_bool("lighttable/ui/expose_statuses", show);
-  dt_thumbtable_set_overlays(dt_ui_thumbtable(darktable.gui->ui), show);
-
-#ifdef USE_LUA
```
TurboGit, line 167: I suppose we want this in  _overlays_toggle_button() and passing the boolean FALSE for DT_THUMBNAIL_OVERLAYS_NONE and TRUE otherwise.

It would be a pity to loose this. Later we can add mode information by passing the overlay mode in another lua call or updating the API but at least let's try to keep this.

src/libs/tools/global_toolbox.c
```diff
@@ -74,6 +75,61 @@ int position()
...
+
+static void _overlays_show_popup(dt_lib_module_t *self)
+{
+  dt_lib_tool_preferences_t *d = (dt_lib_tool_preferences_t *)self->data;
+
+  // we write the label with the size categorie
+  gchar *txt = dt_util_dstrcat(NULL, "%s %d", _("overlay mode for size"),
+                               dt_ui_thumbtable(darktable.gui->ui)->prefs_size);
+  gtk_label_set_text(GTK_LABEL(d->over_label), txt);
+
```
TurboGit, line 115: You need a g_free(txt) here.

src/libs/tools/global_toolbox.c
AlicVB: well in fact I've done TRUE if DT_THUMBNAIL_OVERLAYS_ALWAYS_NORMAL or DT_THUMBNAIL_OVERLAYS_ALWAYS_EXTENDED. and FALSE otherwise. So it'll mimic the old behavior little better.

AlicVB:
Thanks for the review.
I've fixed what you mentionned plus remove an unused fct in thumbtable code...
And yes , I prefer letting our specialist (@Nilvus) dealing with themes colors :)

TurboGit:
Perfect, merging now. We will certainly need more field testing, I have tested many combinations, but this is so feature rich that I could well have forgotten something. Really nice work again @AlicVB ! Thanks.

johnny-bit:
Will test it because I literally asked for it and I believe it's wonderfull ;)

elstoc:
Love the new options but there's always room for one more! I generally use the basic overlays (stars and colour ratings) only but I like as much of the screen as possible devoted to the images, and portrait images are so much smaller than landscape with the reserved space (especially since I've reduced the borders in CSS). I would prefer for just the star/colour ratings to be able to permanently overlay on top of the image as they do for the 'on hover' option. Maybe this could be a global preference rather than another (pair of) option(s)?

ptilopteri:
Overlapping the image on my display
see attachment

openSUSE Tumbleweed 20200414
darktable 3.1.0~git1222.bfc12a28e-2814.1.x86_64
OpenCL on 
i7 12-core 36GB

IMG_20200416_185404

TurboGit:
@ptilopteri : that's one option, see next one by clicking on the "stars" on top-right of the table. You'll see 5 options which are kept for different zoom level. The current zoom level is displayed on the memu proposing the 5 options.

AlicVB:
@elstoc : see #4742 (sorry to have forgotten your idea...)

Nilvus:
Good job @AlicVB: I will work on CSS this week-end and test/improve all default themes. @TurboGit: It will help me if possible if you can merge #4742 ASAP.

ptilopteri:
Should not the "star" be again highlighted to indicate one of the options was selected rather than no option "the no option" was chosen.  

Maybe the "no option" line could then be removed as superfluous using the "star" as a toggle for option to be active or not and reduce screen clutter.

TurboGit:
@Nilvus : #4742 is merged, you can now work on the CSS part.

PR #4738, opened 2020-04-16
retouch: fix button toggle

elstoc:
resolves issue with the 'show/hide shapes' button being incorrectly
toggled on. Issue occurs when selecting earlier history items while
retouch shapes are in edit mode. When you go back to the top of the
history stack the show/hide button was toggled on but shapes were not
shown.


TurboGit:
Works for me. It is indeed better to switch off the "edit masks" button when moving in history.Thanks.

PR #4740, opened 2020-04-16
introduce universal toast message for sliders and comboboxes

johnny-bit:
this should univerally cover toasts for sliders and comboboxes
when called from either normal or dynamic accel keys in a way
that won't trigger any reverts. also avoids code duplication
and covers multi-instance cases :)

This merges functionality provided by @dterrahe and @elstoc in single point, away from `bauhaus.c`

### DIFF
--- a/src/gui/accelerators.c
+++ b/src/gui/accelerators.c
@@ -544,7 +544,7 @@ void dt_accel_widget_toast(GtkWidget *widget)
 {
   dt_bauhaus_widget_t *w = (dt_bauhaus_widget_t *)DT_BAUHAUS_WIDGET(widget);
 
-  if(!gtk_widget_is_visible(GTK_WIDGET(w)) && *w->label && !darktable.gui->reset)
+  if(!gtk_widget_is_visible(GTK_WIDGET(w)) && !darktable.gui->reset)
   {
     char *text = NULL;
     int show = 1;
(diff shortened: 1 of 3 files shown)


dterrahe:
https://github.com/darktable-org/darktable/pull/4424#issuecomment-601846551

elstoc:
@dterrahe oh yes I forgot about that. We should incorporate that functionality too.

johnny-bit:
@dterrahe - done in bca3558 hope it's fine :)

dterrahe:
The slider code used to have:
```
else
        dt_control_log(_("%s: %s"), w->label, text);
```
You'd have to check if that works for comboboxes.

elstoc:
When I did the work for comboboxes I went through them all - I couldn't find any comboboxes where the name of the combobox also exists in the name of the module, and there are only a handful of sliders that do. 

The complete statement should probably add another ``else if (w->label) dt_control_log(_("%s: %s"), w->label, text);`` and then finish with the existing ``else``

dterrahe:
The rest of the code assumes that w->label != NULL, so if you feel the need to test it, you might want to wrap everything that's currently within if(show) with it. The original test was `if (*w->label)`. Can't remember if that was a mistake or if there's always a pointer, but sometimes to an empty string. And again, that might be different between sliders and combos. Something else to test...

elstoc:
Yes I wasn't sure how that last else could ever be reached. I think we can treat it like the default case - unlikely to ever be triggered but nice to catch things we haven't thought of.

I haven't been able to test it (apart from the early preview I looked at last night) but looks good to me now.

dterrahe:
Actually, the `if (*w->label)` test is still there, at the top. I overlooked, sorry.

elstoc:
Yes it should be removed from that top if statement. As previously stated it's probably vanishingly unlikely to ever occur.

johnny-bit:
OK, I'll fix build and test it in a bit.

at the start test is for 2 things:
1. we can't see widget
2. gui isn't resettting

once we're clear with that there's text preparation.

show is by default true and for sliders it CAN be false only if dragging is true, and it's impossible to set dragging via key accell, right? so this is just superflous test, just to make sure ;)

so now choosing the right format...
is it safe to assume all combos/sliders have label? most formats assume that.

elstoc:
I couldn't work out what the 'dragging' bit was used for. Perhaps it is superfluous for accels. 

I think it is safe to assume that all comboboxes *with keyboard shortcuts* have a label, because I added them myself, and they do.

johnny-bit:
well `label` is a static char array so if `w` is valid pointer to widget structure, it's guaranteeed to be ok... so... test if it's empty?

elstoc:
I suppose.

johnny-bit:
OK, this logic should cover all the corner cases ;)

elstoc:
If the label is empty ``strstr(w->module->name(), w->label)`` doesn't make sense. Just have ``dt_control_log(_("%s"), text);``. No need to make it more complicated than that, since it'll never happen anyway.

johnny-bit:
OK, so now this should cover all possible codepaths, even impossible/impossilby stupid ones ;) (I think)

elstoc:
Fine for me now.

TurboGit:
Works for me, thanks !

houz:
> When I did the work for comboboxes I went through them all - I couldn't find any comboboxes where the name of the combobox also exists in the name of the module, and there are only a handful of sliders that do.

When writing code for the dt core you shouldn't assume that the modules using it are only doing some subset of possible things. Modules might change and people might have modules that are not in our code.

elstoc:
Fair enough

Issue #4692, opened 2020-04-12
Can't build anymore on Windows


Mark-64:
Hi,
I have always been able to compile DT git master on Windows 10. Last version I built is 3.1.0+801~gad36572f5.
Now after some time, and having first updated my MSYS2 installation, I suddenly get this CMake error

```
CMake Error: The following variables are used in this project, but they are set to NOTFOUND.
Please set them or make sure they are set and tested correctly in the CMake files:
_osmgpsmap_LIBRARY
    linked by target "lib_darktable" in directory C:/msys64/home/marco/darktable-3.0/darktable/src
    linked by target "lib_darktable" in directory C:/msys64/home/marco/darktable-3.0/darktable/src
    linked by target "lib_darktable" in directory C:/msys64/home/marco/darktable-3.0/darktable/src
    linked by target "lib_darktable" in directory C:/msys64/home/marco/darktable-3.0/darktable/src
    linked by target "lib_darktable" in directory C:/msys64/home/marco/darktable-3.0/darktable/src
    linked by target "lib_darktable" in directory C:/msys64/home/marco/darktable-3.0/darktable/src
    linked by target "lib_darktable" in directory C:/msys64/home/marco/darktable-3.0/darktable/src
    linked by target "lib_darktable" in directory C:/msys64/home/marco/darktable-3.0/darktable/src
```

Also, I get a series of warnings that I am sure weren't there before, for example:
```
CMake Warning (dev) at C:/msys64/mingw64/share/cmake-3.17/Modules/FindPackageHandleStandardArgs.cmake:272 (message):
  The package name passed to `find_package_handle_standard_args` (PANGOCAIRO)
  does not match the name of the calling package (PangoCairo).  This can lead
  to problems in calling code that expects `find_package` result variables
  (e.g., `_FOUND`) to follow a certain pattern.
Call Stack (most recent call first):
  cmake/modules/FindPangoCairo.cmake:27 (find_package_handle_standard_args)
  src/CMakeLists.txt:355 (find_package)
This warning is for project developers.  Use -Wno-dev to suppress it.
```

GrahamByrnes:
I had a different failure about a week ago after updating msys2. That was specific to the gps features... it couldn't find osm-gps-map. Are you using msys2?
You could try turning off the map option in the file darktable/DefineOptions.cmake
Or it could be something else, and I agree that it's not ideal...

Mark-64:
[...]

PR #4741, opened 2020-04-17
Enable windows build with latest MSYS2

phrrk:
Latest MSYS2 uses cmake 3.17 which no longer searches for .dll files when locating libraries for linking.  This results in the windows build failing as various libraries required for mapping are not found (issue #4692) .  This PR explicitly points cmake to the location where relevant libraries are located.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -129,6 +129,7 @@ include(compiler-versions)
 
 if(WIN32)
   message("-- Win32 build detected, setting default features")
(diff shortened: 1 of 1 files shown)


TurboGit:
Thanks, I'll let @peterbud approve this as I'm not a Windows user.

phrrk:
That doesn't work after all as mingw cmake doesn't understand Unix style path in this context.

PR #4756, opened 2020-04-18
Fix indexing out of bounds in Chromatic Aberration module.

sh1llo:
This is a port of the upstream patch
https://github.com/Beep6581/RawTherapee/commit/2f32afa8, fixing issue
https://github.com/Beep6581/RawTherapee/issues/4116

At certain resolutions, border filling of the last tile will overwrite
the previous, creating a garbled line across the image.

Note that RawTherapee commit only fixes one set of copy/pasted loops,
while this commit fixes both. In fact the other fixed set seems
necessary to get rid of the artifacts I saw, and looking at the code,
it indeed seems to fix indexing out-of-bounds.


TurboGit:
Thanks !

Issue #4149, opened 2020-01-19
add RGB parade and vectorscope as histogram alternatives


dtorop:
**Is your feature request related to a problem? Please describe.**
Video software tends to have an "RGB Parade" and "Vectorscope". Would these be useful for darktable? They would provide alternate ways to visualize color/lightness data besides the regular and waveform histogram.

**Describe the solution you'd like**
An "RGB Parade" would be trivial to implement as a generalization to the current waveform histogram code. There would be no changes needed to the pixelpipe math, only an additional type of drawing in `libs/histogram.c`. It should be fast to draw.

A Vectorscope would require more code in `develop/pixelpipe_hb.c` and would use a bit more memory for a buffer. It would involve adding a bit of code to `libs/histogram.c`. The polar coordinate math wouldn't be too slow so long as the scope was somewhat lo-res. It should be fast to draw once the color counts are calculated out to a buffer.

With that many histogram/scope alternatives, it would make sense to add keyboard accelerators to choose a specific scope, or to cycle through scopes.

The current histogram widget is a bit squashed for waveforms (or rgb parade), much less a circular vectorscope. It might make sense to allow the widget to resize vertically via the GUI, or to make it automatically resize depending on its contents.

I'd be happy to work on an RGB parade or vectorscope if it were of interest and there was some clarity about a good design.

**Alternatives**
The waveform histogram already has essentially the same data as an RGB parade, though it may take selecting/deselecting color channels to understand how a particular channel works.

A vectorscope came up in #2140 as a way to visualize how to balance color. It's possible that a home for a vectorscope would be in some sort of interactive color balancing iop, rather than in the histogram widget. Though I'd imagine the readout would be useful as a reference when using a variety of iops.

I'm also curious about whether a "power editor" would want to see multiple scopes at once, e.g. both waveform and vectorscope.

**Additional context**
Some articles about using scopes:

- https://www.cinema5d.com/primary-colour-correction/
- https://waondering.com/2015/05/18/continuity-of-colors-matching-exposure-and-color-balance/
- https://www.premiumbeat.com/blog/rgb-parade/
- https://www.bhphotovideo.com/explora/video/tips-and-solutions/introduction-waveforms-scopes-and-exposure
- https://thevideoproguys.com/read-histogram-vectorscope-color-correcting-tools/

There appears to be vectorscope code in qctools (https://github.com/bavc/qctools/wiki/Vectorscope). This could be helpful as a model.

[...]

PR #4768, opened 2020-04-19
RGB parade as histogram alternative

dtorop:
This adds RGB parade as a histogram alternative (see #4149).

To not require cycling through too many histogram types, the single button which cycled between linear/logarithmic/waveform is replaced by two buttons. One toggles between "regular" histogram and waveform-style. The other, when in histogram mode, toggles between linear and logarithmic histogram. When in waveform mode, it toggles between waveform and RGB parade.

Much of the code changes (including a new conf entry) are to allow this mode/submode switching. This will also clear the way for other modes (e.g. vectorscope) without bothering users who already don't want to see a waveform when simply want to switch between linear and logarithmic histograms. 

There are also some optimizations, particularly to histogram calculation. This PR should be equivalent speed to the code in master, for both calculating and drawing histograms. It requires 3/4 the memory to store the calculated waveform histogram, though it uses more temporary memory when drawing the waveform histogram.

### DIFF
--- a/data/darktableconfig.xml.in
+++ b/data/darktableconfig.xml.in
@@ -1843,12 +1843,35 @@
     <name>plugins/darkroom/histogram/mode</name>
     <type>
       <enum>
-        <option>linear</option>
-        <option>logarithmic</option>
+	<option>histogram</option>
         <option>waveform</option>
       </enum>
     </type>
+    <default>histogram</default>
+    <shortdescription/>
+    <longdescription/>
+  </dtconfig>
+  <dtconfig>
+    <name>plugins/darkroom/histogram/histogram</name>
+    <type>
+      <enum>
+	<option>logarithmic</option>
+        <option>linear</option>
+      </enum>
+    </type>
     <default>logarithmic</default>
+    <shortdescription/>
+    <longdescription/>
+  </dtconfig>
+  <dtconfig>
+    <name>plugins/darkroom/histogram/waveform</name>
+    <type>
+      <enum>
+	<option>overlaid</option>
+        <option>parade</option>
+      </enum>
+    </type>
+    <default>overlaid</default>
     <shortdescription/>
     <longdescription/>
   </dtconfig>
--- a/src/develop/develop.c
+++ b/src/develop/develop.c
@@ -47,7 +47,9 @@
 #define DT_DEV_AVERAGE_DELAY_COUNT 5
 #define DT_IOP_ORDER_INFO (darktable.unmuted & DT_DEBUG_IOPORDER)
 
-const gchar *dt_dev_histogram_type_names[DT_DEV_HISTOGRAM_N] = { "logarithmic", "linear", "waveform" };
+const gchar *dt_dev_scope_type_names[DT_DEV_SCOPE_N] = { "histogram", "waveform" };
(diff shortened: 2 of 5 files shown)


prokoudine:
Your more recent 'more-scopes' branch fails to build though

```
[ 96%] Test-compiling OpenCL program blendop.cl
/home/alex/soft/photography/darktable/darktable-more-scopes/data/kernels/blendop.cl:1004:35: error: use of undeclared identifier 'CL_FLT_EPSILON'
  const float mask_epsilon = 16 * CL_FLT_EPSILON;  // empirical mask threshold for fully transparent masks
                                  ^
1 error generated.
make[2]: *** [data/kernels/CMakeFiles/opencl-program-blendop.cl.dir/build.make:62: data/kernels/blendop.cl.touch] Error 1
make[1]: *** [CMakeFiles/Makefile2:8763: data/kernels/CMakeFiles/opencl-program-blendop.cl.dir/all] Error 2
```

dtorop:
@prokoudine:

> Your more recent 'more-scopes' branch fails to build though
...
> /home/alex/soft/photography/darktable/darktable-more-scopes/data/kernels/blendop.cl:1004:35: error: use of undeclared identifier 'CL_FLT_EPSILON'
>   const float mask_epsilon = 16 * CL_FLT_EPSILON;  // empirical mask threshold for fully transparent masks

Try the rgb-parade branch (as in this PR). The more-scopes branch is very much work-in-progress, and even though it may have commits dated newer, is actually older code (I rebased before making rgb-parade branch). The more-scopes branch is from just before the fix in 7781cb1c1f27c4c697ce93e135b74096b76d6140, and won't compile without that fix.

I'll rebase & update more-scopes eventually, but it's very much a working branch.

Thank you for trying this out.

TurboGit:
This looks ok. I tested it and it works fine, I'm not quite clear about the way to use such histogram as it is new to me. I'm merging that. Thanks.

dtorop:
@TurboGit That's fabulous. Thank you for looking it over and merging it! I've come to like the RGB parade after a few weeks of using it. You get to see the character of each color channel and how their overall levels compare with a bit more clarity than waveform. But I also think waveform is great and probably the more useful overall visualization...

phweyland:
Nice !
Should not be the waveform / histogram icons inverted ?

image

dtorop:
@phweyland  writes:

> Should not be the waveform / histogram icons inverted ?

Good questions! I tried not to change the existing behavior (though I sometimes wondered if it is a bit backward).

Here's the behavior before this PR (737735db94617468392fd76d109741a1a530e034):

image

The rule seems to be that the icon represents the current state, and the tooltip gives guidance as to what a click will do. The rationale may be that the converse would be a bit strange -- to see an icon which is the opposite of the current state.

These buttons are a bit of a hack -- custom rolled outside of dt's themed gtk & bauhaus. Maybe the real problem is that these should be proper widgets, e.g. dtgtk_togglebutton, with some more thought to behavior/icon.

dtorop:
To follow up -- in fact, the histogram UI could use a general update. Currently how it positions/sizes the buttons and detects mouseover/click/drag is entirely hand-rolled (a monolithic widget). It redraws the entire histogram widget on motion-notify-event, which is more work than is needed.

PR #4771, opened 2020-04-19
Fix icons for HiDPI display in Windows 10

Mark-64:
This reverts some of the changes which I have done with #4179, which are no longer needed with the new Lighttable and are counter-productive.
Will fix size of the "G" and "?" icons for HiDPI display in Windows 10, while doesn't anything in low DPI


TurboGit:
Ok, looks ok on my non HIDPI screen.

PR #4772, opened 2020-04-19
lut3d update

phweyland:
- move the lut 3d folder definition from configuration to lut3d module
- static routines reformatting

related to discussion in #4747 about preferences

### DIFF
--- a/data/darktableconfig.xml.in
+++ b/data/darktableconfig.xml.in
@@ -1956,13 +1956,6 @@
     <shortdescription>executable for playing audio files</shortdescription>
     <longdescription>this external program is used to play audio files some cameras record to keep notes for images</longdescription>
   </dtconfig>
-  <dtconfig prefs="core">
-    <name>plugins/darkroom/lut3d/def_path</name>
-    <type>dir</type>
-    <default></default>
-    <shortdescription>3D lut root folder</shortdescription>
-    <longdescription>this folder (and sub-folders) contains Lut files used by lut3d modules. need to restart darktable.</longdescription>
-  </dtconfig>
   <dtconfig prefs="core" section="quality">
     <name>plugins/darkroom/basecurve/auto_apply</name>
     <type>bool</type>
--- a/src/dtgtk/paint.c
+++ b/src/dtgtk/paint.c
@@ -1103,6 +1103,28 @@ void dtgtk_cairo_paint_directory(cairo_t *cr, gint x, gint y, gint w, gint h, gi
   cairo_restore(cr);
 }
 
+void dtgtk_cairo_paint_cube(cairo_t *cr, gint x, gint y, gint w, gint h, gint flags, void *data)
+{
+  cairo_save(cr);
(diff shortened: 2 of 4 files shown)


phweyland:
image

TurboGit:
Why not keep the global prefs as the default and still allow the redefine it in the Lut3D module as done here? At least it gives a default and don't require to redefine it all the time.

parafin:
What is exactly the point of this setting? Is it a base path, while only relative path is saved in iop params? Then changing it might break existing history stacks. I think decision where to put this setting should come from description of how it works and should be used (use-case).

phweyland:
>  Is it a base path, while only relative path is saved in iop params? Then changing it might break existing history stacks.

Yes, that's exactly that.  In that sense it should be in configuration and, once set, forgotten. If the user changes it that breaks all existing iop params except if the user has moved this folder and just updates the new location.

But I've seen some complaints of people saying the module did not work while they had not set the base path on configuration.

Last, the default setting of folder in configuration is not very strong. It is enough to open configuration (without touching the lut 3d root folder) to make the flag configuration set (on my machine I get plugins/darkroom/lut3d/def_path=/home/philippe in darktablerc just opening the configuration). This breaks the initialization control (as it hasn't been intentionally set). At least I haven't found the way to make it work better. On the other hand, inside the module (and removed from global configuration) I can control properly the initial setting.

I'm happy with this setting in the module, but without this last issue I would not have moved it out from configuration.

Thoughts ?

elstoc:
On reflection I'm not entirely convinced it makes sense to have a global parameter set inside a module.  Without knowing the details I'd assume that setting was specific to the current image, and that the 'reset parameters' option would reset it back to its previous value.

parafin:
If it shouldn't be modified once set except when user changes something in his system, it's obviously should be in preferences and not in iop GUI. Having it in IOP GUI invites user to change it every time.

phweyland:
OK, I'll close this PR.

About the following issue any idea to solve it ?

> Last, the default setting of folder in configuration is not very strong. It is enough to open configuration (without touching the lut 3d root folder) to make the flag configuration set (on my machine I get plugins/darkroom/lut3d/def_path=/home/philippe in darktablerc just opening the configuration). This breaks the initialization control (as it hasn't been intentionally set). At least I haven't found the way to make it work better. On the other hand, inside the module (and removed from global configuration) I can control properly the initial setting.

elstoc:
The default for this parameter is empty so I assume (without looking at the code) it must be something specific about how the config manages directory entries. Perhaps I could look at it with my parameter overhaul.

phweyland:
> The default for this parameter is empty so I assume (without looking at the code) it must be something specific about how the config manages directory entries. Perhaps I could look at it with my parameter overhaul.

What I need to be able is to be confident on the dt_conf_key_exists().
So far, as said above, it's enough to open configuration / core options to make it existing.

elstoc:
You couldn't do something based on the stored value rather than using dt_conf_key_exists()?

PR #4775, opened 2020-04-20
Optimization of Denoise (bilateral filter) iop

### LINKED ISSUE #4764: [Denoise (bilateral)] performance issue in CPU code
**Describe the bug**

When exporting, Denoise (Bilateral filter) becomes very slow and uses only one core for most of the time it spends if "radius" >= 2.

**To Reproduce**

1. Run "darktable -d perf"
2. Add an instance of Denoise(bilateral) to a reasonably large image (e.g. 24MP).
3. Set "radius" to 1.9, red/green to 0.001, and blue to 0.002, then export a full-resolution TIFF.
4. Set radius to 2.0 and export again.
5. Note that the first second export took much longer, and left CPU usage at 100% for a long period near the beginning.
6. Patch iop/bilateral.cc so it uses the "naive" algorithm for larger radii (I changed from 6 to 16)
7. Rerun and note that exports are faster and CPU usage is much higher.

**Expected behavior**

Both exports should take roughly the same amount of time, and keep most or all cores busy.

**Screenshots**

relevant portion of "-d perf" output:

{{rad=1.9}}  	[dev_pixelpipe] took 1.200 secs (60.621 CPU) processed `denoise (bilateral filter)' on CPU, blended on CPU [export] 
{{rad=2.0}} 	[dev_pixelpipe] took 23.669 secs (378.667 CPU) processed `denoise (bilateral filter)' on CPU, blended on CPU [export] 
{{5.0 w/patch}}	[dev_pixelpipe] took 6.261 secs (371.227 CPU) processed `denoise (bilateral filter)' on CPU, blended on CPU [export]

Note that even with a substantially larger radius, the naive codepath is still faster than the PermutohedralLattice path.

 - Darktable Version: 3.1.0+1231 (but was also present on an earlier git-master version)
 - OS: Mageia Linux 7.1
no OpenCL

**Additional context**

Compiler: GCC 8.4

In looking over the code, I saw one part that wasn't parallelized, and tried commenting it out.  Naturally, this creates a completely messed-up image and probably causes additional reductions further on, but eliminating the call to lattice.merge_splat_threads() does restore the speed:
{{rad=2.0}} [dev_pixelpipe] took 1.698 secs (62.862 CPU) processed `denoise (bilateral filter)' on CPU, blended on CPU [export]

ralfbrown:
After finding that Bilateral spends a long time running on only one core in my new 32-core setup when using low(ish) values for blur amount and radius, I dived into the code and optimized it.  The long single-threaded processing was merging multiple hash tables into one, which triggered repeated expensive resize operations.  Optimizations include caching hash values for keys, eliminating the multiple resizes by adaptively resizing to slightly more than the eventual size before starting the merge, and eliminating redundant data (saving 20 bytes per pixel).

When checking into the required final size of the merged hash table, I found that relatively few keys were common between the tables participating in the merge, so simply resizing the destination table to have capacity to store the sum of the sizes will usually result in a table of exactly the size it would have reached through multiple on-demand resizes (since size is always a power of two).

Fixes #4764

### DIFF
--- a/src/iop/Permutohedral.h
+++ b/src/iop/Permutohedral.h
@@ -357,7 +357,7 @@ public:
     elevated[0] = elevated[1] + 2 * position[0] * scaleFactor[0];
 
     // prepare to find the closest lattice points
-    float scale = 1.0f / (D + 1);
+    constexpr float scale = 1.0f / (D + 1);
 
     // greedily search for the closest zero-colored lattice point
     int sum = 0;
@@ -483,7 +483,7 @@ public:
    */
   void slice(float *col, size_t replay_index)
   {
-    float *base = hashTables[0].getValues();
+    const float *base = hashTables[0].getValues();
     for(int j = 0; j < VD; j++) col[j] = 0;
     for(int i = 0; i <= D; i++)
     {
@@ -501,7 +501,7 @@ public:
     // Prepare arrays
     float *newValue = new float[VD * hashTables[0].size()];
     float *oldValue = hashTables[0].getValues();
-    float *hashTableBase = oldValue;
+    const float *hashTableBase = oldValue;
 
     float zero[VD];
     for(int k = 0; k < VD; k++) zero[k] = 0;
--- a/src/iop/Permutohedral.h
+++ b/src/iop/Permutohedral.h
@@ -501,7 +501,7 @@ public:
     // Prepare arrays
     float *newValue = new float[VD * hashTables[0].size()];
     float *oldValue = hashTables[0].getValues();
-    const float *hashTableBase = oldValue;
+    float *hashTableBase = oldValue;
 
     float zero[VD];
     for(int k = 0; k < VD; k++) zero[k] = 0;
--- a/src/iop/Permutohedral.h
+++ b/src/iop/Permutohedral.h
@@ -62,11 +62,21 @@ public:
     {
        for (int i = 0; i < KD; i++) key[i] = origin.key[i] + direction;
        key[dim] = origin.key[dim] - direction * KD;
+       setHash();
     }
     Key(const Key&) = default; // let the compiler write the copy constructor
     Key& operator= (const Key&) = default;
     void setKey(int idx, short val) { key[idx] = val; }
-
+    void setHash()
+    {
+      size_t k = 0;
+      for(int i = 0; i < KD; i++)
+      {
+        k += key[i];
+	k *= 2531011;
+      }
+      hash = (unsigned)k;
+    }
     bool operator== (const Key& other) const
     {
       for (int i = 0; i < KD; i++) { if (key[i] != other.key[i]) return false; }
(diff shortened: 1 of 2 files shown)


TurboGit:
What is the expected speed up ? Any figures would be nice. Thanks.

ralfbrown:
I have some numbers in the referenced issue (how does one link it to this PR?).  For extreme cases (small radius & blur) on a machine with lots of threads, I got the wall-clock time cut in half.  CPU time doesn't drop nearly as much, since the speedup is largely in the bottlenecking single-threaded function.

TurboGit:
@rawfiner : maybe you want to review this one ? It looks ok to me, but that's not a part of dt I know.

aurelienpierre:
I don't understand what's going on in that PR, but I don't really speak C++ and the permutohedral key stuff is completely unknown to me. Who is the original author of that IOP ?

ralfbrown:
The point of the PR is to get identical results, but faster (and as it happens, while using substantially less memory).  The key commits are 89cfec0, edfa454, and 3f4bb1b.  Much of the rest is code cleanup -- that file was rather a mess....

rawfiner:
I am afraid I don't know this part of dt either... (nor the principle of permutohedral lattice)

Though, I can reproduce ralfbrown's measurements with the parameters given in the issue: with rad 1.9, export takes 6.6 seconds, while with rad 2, export takes 90.9 seconds.

@aurelienpierre the author was hanatos, but the code comes from ImageStack https://github.com/abadams/ImageStack/blob/master/src/Permutohedral.h There were some changes to it since 2010 though, and I guess they were not put back in dt's code.

I did measurements with over several parameters with ralfbrown's code and I always got either an equal time, either a speedup so this is nice.

I also exported several images with and without the changes with several parameters and the images obtained are visually identical.

aurelienpierre:
> The point of the PR is to get identical results, but faster 

That much I understood ^^ The problem is, I can't say, from the code, if you messed something up. Reviewing that will need a lot of coffee.

parafin:
or write a unit test

aurelienpierre:
Unit tests work if you have a clear idea of the issue you are looking for, thus know where to push to make it fail. They don't solve the problem of having to understand code.

parafin:
but you can test that results don't change with this update with some degree of confidence

TurboGit:
In this new version I have moved the tests.

Also, I do check that the CPU & GPU version are the same or not and display a message + compute the diff image. Could help later, at the moment I do not except to have the identical image anyway.

I have also experienced a diff when the image is exported from cli or from darktable. This looks like some tiling issue but I have seen error around 10 or 12 dE !!! At this stage again this PR is not about tracking those issues, it may come later. We I want to have in place is a reference to check that new code do not break existing output. So the `expected.jpg` must now be generated with the cli. README.txt has been update accordingly.

TurboGit:
From my integration testsuite:

```
Test 0003-denoise-bilateral
      Image mire1.cr2
      CPU & GPU version differs
      Max  dE         0
      Mean dE         0
      Count above max 0
  OK

Total test 1
Errors     0
```

TurboGit:
So, I'll merge this. Thanks.

Issue #4760, opened 2020-04-18
A quite annoying bug: cancel a rejected image is no more possible


Nilvus:
**Describe the bug**
On last master, cancel a rejected image is no more possible.

**To Reproduce**
Provide detailed steps that can reproduce the behavior, such as:
1. On lighttable, just select an image (with shortcut or clicking on the cross)
2. Select the same rejected image and try to cancel last action by clicking again on the red cross or with 'r' shortcut.
=> image still rejected...

**Expected behavior**
Have possibility to cancel a rejected image.

phweyland:
However ctrl-z does work, right ?

Nilvus:
> However ctrl-z does work, right ?

Yes.

Nilvus:
I've just found another related behavior. If i select only rejected images to display (so with few rejected ones) on lighttable and then go to preview mode with Alt+w and try to cancel the rejected image, the image disappear (like before when cancelling a rejected image was possible). But If quit the preview mode, the image is still rejected and always on rejected filter mode.

progo:
I once made a remark on the IRC channel about the combined behavior of rejection and the setting "rating an image twice will not reset the rating to zero".

I want that behavior where I can reject an image multiple times and it'll stay rejected. I want to be looking at the images, not the stars or rejectedness when I'm culling through the shots. 

One could tie the behavior to that settings value. Someone who doesn't care about zero-star ratings can enable the  setting and experience idempotent ratings/rejections, whereas someone who desires it can set it to FALSE.

Nilvus:
> I once made a remark on the IRC channel about the combined behavior of rejection and the setting "rating an image twice will not reset the rating to zero".
> 
> I want that behavior where I can reject an image multiple times and it'll stay rejected. I want to be looking at the images, not the stars or rejectedness when I'm culling through the shots.
> 
> One could tie the behavior to that settings value. Someone who doesn't care about zero-star ratings can enable the setting and experience idempotent ratings/rejections, whereas someone who desires it can set it to FALSE.

I personally never read IRC. Then, it's a bug report here, not a feature request. So I find your answer not related here. Of course, you talk about reject one. I want to keep the possibility to cancel rejected images. I use that to select images I want to keep and a second time, I come back to rejected images to show if I don't reject an image that finally I want to keep.

If you want another behavior (ok, if we have the possibility to keep the current known behavior) but please post a separate feature request to help following what to do. By now, the goal is to correct actual bug not to change it.

[...]

PR #4779, opened 2020-04-20
po/reject toggle

TurboGit:


TurboGit:
@AlicVB : please double check that I have not made something bad with your recent work.

AlicVB:
@TurboGit : quickly reading the code,` rating.c:168` seems strange... I thing we should apply `new_rating` here ? Does the "1 star" toggle even works ? (sorry I can't test just now)

Nilvus:
> After a git blame, I think it's a typo from [3ee5436](https://github.com/darktable-org/darktable/commit/3ee54369557fcefe8cec1ceb6bbe8b27b5ef9c22) That would explain why @Nilvus says it's a very recent regression...

@AlicVB: thanks for pointing that. Just test and previous commit was ok. And yes, I confirm that is this commit from @phweyland that introduce my issue #4760. @TurboGit: something to maybe check?

TurboGit:
I'd like to propose that when the reject button is clicked to remove it the number of stars that are used at import time is selected instead of removing all stars.

TurboGit:
@AlicVB @Nilvus @phweyland ... what do you think ? This seems more natural to me when you recover images to go back as it was when importing.

ptilopteri:
> I'd like to propose that when the reject button is clicked to remove it the number of stars that are used at import time is selected instead of removing all stars.

not what I would expect.  If I opted to reject and image and then changed my mind, I would expect to regain the image as was when originally rejected.  Least supprise.

Please reconsider

phweyland:
> If I opted to reject and image and then changed my mind, I would expect to regain the image as was when originally rejected. Least supprise.

Maybe a good idea. That's the purpose of ctrl-z. As reject is the last action, we may trig undo when one click on the red cross. Should work.

phweyland:
> > If I opted to reject and image and then changed my mind, I would expect to regain the image as was when originally rejected. Least supprise.
> 
> Maybe a good idea.

Made some tests. I'm able to retrieve the rating before rejection from undo queue.
Then we can apply this rating when the user clears the rejection.

> I'd like to propose that when the reject button is clicked to remove it the number of stars that are used at import time is selected instead of removing all stars.

And if there is nothing in the queue (because it has been cleared) I'll apply the @TurboGit suggestion.

TurboGit:
@phweyland : sounds good, I'll wait for more comments about this before merging and you can implement the retrieve from undo list. Is that ok with you ?

houz:
> Made some tests. I'm able to retrieve the rating before rejection from undo queue.
Then we can apply this rating when the user clears the rejection.

That means that clicking the reject button of a rejected image does different things depending on some internal state that the user has no idea about.

phweyland:
> > Made some tests. I'm able to retrieve the rating before rejection from undo queue.
> > Then we can apply this rating when the user clears the rejection.
> 
> That means that clicking the reject button of a rejected image does different things depending on some internal state that the user has no idea about.

That does what Ctrl-z does with the difference that Ctrl-z acts on the last changed images while un-reject acts on 'that' image.

> I'll wait for more comments about this before merging and you can implement the retrieve from undo list. Is that ok with you ?

That's perfect.

TurboGit:
I'm wondering if we don't want to get this right using another way. I'm proposing to change the DT_VIEW_REJECT to be a flag using it's own bit. When set the image is rejected but we do keep the stars information. With this, no need to look at the undo, and we can at any time (even after restarting dt) recover an image by giving it back the last rating.

How does that sound?

phweyland:
> When set the image is rejected but we do keep the stars information.

That will be predictable for the user if these stars are still shown (half hidden for example). This sounds better than the queue search (even if I had fun coding it).

This said, the current behavior (without bug :) ) is correct for me. If the user has made a mistake ctrl-z makes it safe. And it sounds a bit unusual to move from 4 stars to rejected ...

jenshannoschwalm:
I also like @TurboGit suggestion introducing a reject flag

TurboGit:
@phweyland : sure ctrl+z does work but only in current session.

ptilopteri:
> > > Made some tests. I'm able to retrieve the rating before rejection from undo queue.
> > > Then we can apply this rating when the user clears the rejection.
> > 
> > 
> > That means that clicking the reject button of a rejected image does different things depending on some internal state that the user has no idea about.
> 
> That does what Ctrl-z does with the difference that Ctrl-z acts on the last changed images while un-reject acts on 'that' image.

To be clear, using the "X" to toggle rejection state should to ONLY that, toggle the state.  The subject image should be as it was then the previous before the toggled state was activated.
Rejection and UnRejection, NOTHING ELSE.  Least suprise.

TurboGit:
@ptilopteri : no need yo yell, thanks !

lhietal:
Preserving the raiting when rejecting an image sounds like a nice improvement. But I think xmp:rating=-1 sould be still used for importin and exporting since that is widely used in other software. (e.g. https://www.iptc.org/std/photometadata/specification/IPTC-PhotoMetadata#image-rating)

AlicVB:
@TurboGit : that should work quite well. We just need to be carefull with interoperability with other soft. Is rating registered in xmp in an specific dt tag or does this use a well defined one ?

For the implementation, maybe the simpler is to just toggle the sign of the rating : >0 rating <0 rejected

houz:
From the XMP specs:

> A user-assigned rating for this file. The value shall be -1 or in the range
[0..5], where -1 indicates “rejected” and 0 indicates “unrated”. If xmp:Rating
is not present, a value of 0 should be assumed.
NOTE Anticipated usage is for a typical “star rating” UI, with the addition of
a notion of rejection.

So unfortunately some other negative values than -1 are not allowed.

ptilopteri:
> @TurboGit : that should work quite well. We just need to be carefull with interoperability with other soft. Is rating registered in xmp in an specific dt tag or does this use a well defined one ?
> 
> For the implementation, maybe the simpler is to just toggle the sign of the rating : >0 rating <0 rejected

The "star" rating should not have anything to do with the "status" rejected.  The history of star rating is 0 -> 5 stars and nothing to do with the "status" "rejected".  They are exclusive.

Nilvus:
> This said, the current behavior (without bug :) ) is correct for me. If the user has made a mistake ctrl-z makes it safe. And it sounds a bit unusual to move from 4 stars to rejected ...

I agree with that, no need to make that more complicated. It's great as it is (without bug, as said!), no more needed I think and no need to spend more time on this, a lot of other issues/feature requests waiting are probably more and more important.

TurboGit:
Implementation done, code is not that complex. To be tested.

TurboGit:
@AlicVB @Nilvus :This is ready for comments / testing. Thanks.

phweyland:
If I understand this keep the information rejected independent of stars. This works internally but stars are lost in xmp file if rejected.
This creates a discrepancy between db and xmp files, doesn't it ?

TurboGit:
> This works internally but stars are lost in xmp file if rejected.

Right, but we cannot do otherwise as the XMP RFC do not permit that. And at this point the only case not covered is setting rejected, removing the picture and rei-mporting it and we get the rejected without a way to recover the stars. No big deal if you ask me.

Nilvus:
@TurboGit: as you want. As it is here, it's good for me and I can reject and cancel reject as I want, even with shortcut. So good for me.

TurboGit:
Merged !

PR #4782, opened 2020-04-20
Fix overlays icons size for HiDPI Windows 10

Mark-64:
Hi, this tries to fix the wrong size of some thumbnail overlay icons in the new Lightable for HiDPI displays in Windows 10.

Cattura24

It does not affect non-HiDPI displays.
However it's a workaround. The problem of icons scaling incorrectly cannot be solved if we keep a mix of Cairo drawing icons and simple text icons like "G" or "?". They will always scale differently with DPI and proportions cannot be kept accurately.
For a proper solution we should replace completely text icons.

### DIFF
--- a/src/dtgtk/button.c
+++ b/src/dtgtk/button.c
@@ -101,7 +101,7 @@ static gboolean _button_draw(GtkWidget *widget, cairo_t *cr)
   if(DTGTK_BUTTON(widget)->icon)
   {
     /* set inner border and icon size */
-    float f_border = (1 + 0.66 * (darktable.gui->dpi_factor-1)) * ((flags & CPF_DO_NOT_USE_BORDER) ? 4.0 : 6.0);
+    float f_border = ((flags & CPF_DO_NOT_USE_BORDER) ? 4.0 : 6.0);
     int border = round(f_border);
     int icon_width = round(text ? height - (f_border * 2) : width - (f_border * 2));
     int icon_height = round(height - (f_border * 2));
--- a/src/dtgtk/paint.c
+++ b/src/dtgtk/paint.c
@@ -1574,12 +1574,7 @@ void dtgtk_cairo_paint_preferences(cairo_t *cr, gint x, gint y, gint w, gint h, 
 {
   cairo_translate(cr, x + (w / 2.0), y + (h / 2.0));
 
-  //cairo_set_line_width(cr, .25);
-  //cairo_rectangle(cr, -w/2.0, -h/2.0, w, h);
-  //cairo_stroke(cr);
-
   float s = (w < h ? w / 2.0 : h / 2.0);
-  //s = 11.0;
   cairo_scale(cr, s, s);
 
   cairo_set_line_width(cr, .25);
--- a/src/dtgtk/paint.c
+++ b/src/dtgtk/paint.c
@@ -1383,22 +1383,23 @@ void dtgtk_cairo_paint_local_copy(cairo_t *cr, gint x, gint y, gint w, gint h, g
 
 void dtgtk_cairo_paint_altered(cairo_t *cr, gint x, gint y, gint w, gint h, gint flags, void *data)
 {
-  const gint s = (w < h ? w : h);
-  const float r = 0.5;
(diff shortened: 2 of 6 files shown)


parafin:
Why are they scaled differently?..

TurboGit:
The patch which is very specific looks wrong to me.

Mark-64:
Here are the samples on three systems, please look only at size of icons, not to colors.
This is without the patch:

Windows 10, LoDPI screen
Cattura_2w_lodpi__wo

Windows 10, HiDPI screen
Cattura_2w_hidpi__wo
(this looks plain wrong as you can see, text and graphics icons scale differently, proportions are not kept)

Linux Ubuntu, HiDPI screen
Cattura_2l_hidpi__wo
(Linux and Windows scaling mechanisms are different)

and this is with the patch:

Windows 10, LoDPI screen
Cattura_2w_lodpi__w

Windows 10, HiDPI screen
Cattura_2w_hidpi__w

Linux Ubuntu, HiDPI screen
Cattura_2l_hidpi__w

The patch has no effect on low dpi screens and in Linux, only affects Windows HiDPI screens.
Yes it is very specific correction, althoug the code is similar to function _button_draw() in button.c, but as I said it is a workaround. Proper fix would be replace text icons like "G" and "?" with Cairo drawings

TurboGit:
> For a proper solution we should replace completely text icons.

Then I vote for this.

Nilvus:
@Mark-64: I agree with @TurboGit. Having a proper solution would be better. And HiDPI icons with you patch on Linux is not really great: altered icon is a little bigger (and cropped) in regards to the group icon.

TurboGit:
@Mark-64: are you ok to work on creating Cairo based 'G' and '?' ?

Mark-64:
new Cairo icons for Group and Help. I still have to do some tests on different OS / screen.

TurboGit:
@Mark-64 : Perfect ! Ping me when testing done. I'll do some testing on my side too.

Nilvus:
Will test that also (Linux Ubuntu 20.04 and lodpi screen). Thanks for that work.

TurboGit:
I like the new grouped icon and the ? looks ok to me.

image

Mark-64:
The icons look ok on the three systems I have access to (Win10 lowdpi/hidpi and Ubuntu 19.10 hidpi).
Scaling is pixel accurate now, for example
Cattura26
Cattura27
Had to slightly retouch the "altered" icon too. So this is ready for review.

BUT...
I consider this a first step. I still don't like how DT looks in Win10 HiDPI. Some icons look too small with respect to Ubuntu, same screen. 
Different OSs have different scaling mechanisms and we are still far from having a seamless GUI experience across all.
Digging into paint.c and the button draw routines in various parts of the code, I realized there are still some gui elements where size is hardcoded in pixels. The Cairo icons do not always have the same size natively and are adjusted ad-hoc.
To have a truly proportional scaling there is quite a lot of work to do. Before starting a time consuming tasks, I would like to know if you support the idea.

parafin:
Maybe you should first outline the problem - how scaling is different on different OSes and what DT does wrong right now. It's not obvious for me at least.

Mark-64:
I added one last commit, removing an ad-hoc HDPI fix which I did some time ago, no longer needed

Mark-64:
@parafin I will, but not now. I have to switch to lua, as one of my scripts is broken after the LT rewrite

TurboGit:
Looks good to me! @Nilvus, ok to you ?

Nilvus:
> Looks good to me! @Nilvus, ok to you ?

Not test by now, lack of time today. Will try to check tonight.

Nilvus:
This looks good but I've test with some tricks as your branch is not rebased on recent master (you don't have my recent CSS work on overlays mode). I've test with my work but and all seems good.

Thanks for that work, like the new icons and especially the new groups icons (at first, just strange to not see the ugly old G icon!).

Nilvus:
Just an update: tested rebased on master. All good for me. @TurboGit: I think this could be merged.

TurboGit:
Merging then, Thanks.

Issue #4786, opened 2020-04-21
PR #4768 (RGB parade) crashes DT when entering darkroom under Windows 10


Mark-64:
Commit 737735db94617468392fd76d109741a1a530e034 from PR #4768 makes DT crash when opening any image in darkroom under Windows 10.
attached the backtrace log
[gdb.txt](https://github.com/darktable-org/darktable/files/4510010/gdb.txt)

GrahamByrnes:
Ah, not just me... I was about to blame it on inkscape...

GrahamByrnes:
Works :-)

Mark-64:
fixed

PR #4787, opened 2020-04-21
histogram fix - dt_free_align for windows

### LINKED ISSUE #4773: Imported folder and double clicking new image just crashes DT
darktable-3.1.0+1266~g737735db9-win64

Imported pictures from an already imported folder as I added new one.

After import, double clicking on a new image just makes DT disappear.

Previously seen this when memory allocated is not freed corectly.

phweyland:
should fix #4773 & PR #4786

### DIFF
--- a/src/libs/histogram.c
+++ b/src/libs/histogram.c
@@ -372,7 +372,7 @@ static gboolean _lib_histogram_draw_callback(GtkWidget *widget, cairo_t *crf, gp
         cairo_set_operator(cr, CAIRO_OPERATOR_ADD);
         cairo_paint(cr);
         cairo_surface_destroy(source);
-        free(wf);
+        dt_free_align(wf);
       }
       else
       { // RGB parade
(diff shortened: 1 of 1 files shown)


TurboGit:
Looks ok, thanks !

phweyland:
On windows the machine time is wrong and may explain the check doesn't start here.
That why I'm forcing again from linux. Seems to work better.

dtorop:
@phweyland Belatedly, thank you from fixing my mistake. Apologies not to have been present for this.

I was working to make the waveform and RGB parade draw with comparable speed to the regular histogram. There was a speed increase by aligning the temporary buffers -- presumably thanks to Cairo preferring this. (This is in contrast to waveform calculation, for which aligning the buffers made no noticeable improvement.)

Alas I didn't figure out that Windows (and hence all darktable code) needs special work to free aligned memory.

Looking at this, there's a bit more to do -- the histogram draw code doesn't need to allocate two buffers in the case of waveform. I'll work on that.

PR #4789, opened 2020-04-21
Local laplacian speedup (fixes #4788)

### LINKED ISSUE #4788: Local laplacian performance issue with high thread counts
**Describe the bug**

CPU code path for Local contrast using local laplacians gets increasingly slower beyond 24 threads.

**To Reproduce**

On a machine with at least 32 hardware threads, with an image that uses local constrast/laplacians:
1. run darktable -t 16 -d perf --disable-opencl
2. export the image
3. run darktable -t 32 -d perf --disable-opencl
4. export the image
5...n: run darktable with increasing number of threads, exporting the image each time. Note how the wall-clock time for local contrast actually increases dramatically with more threads.

**Expected behavior**

Wall time should decrease as the number of threads increases.

**Screenshots**

 - Darktable Version: git master
 - OS: Mageia Linux 7.1
 - no OpenCL

**Additional context**

Performance on my machine:
(64 threads) [dev_pixelpipe] took 11.644 secs (95.839 CPU) processed `local contrast' on CPU, blended on CPU [export]
(48 threads) [dev_pixelpipe] took 8.040 secs (68.568 CPU) processed `local contrast' on CPU, blended on CPU [export]
(32 threads) [dev_pixelpipe] took 6.005 secs (53.605 CPU) processed `local contrast' on CPU, blended on CPU [export]
(28 threads) [dev_pixelpipe] took 2.246 secs (55.617 CPU) processed `local contrast' on CPU, blended on CPU [export]
(24 threads) [dev_pixelpipe] took 2.140 secs (44.794 CPU) processed `local contrast' on CPU, blended on CPU [export]
(16 threads) [dev_pixelpipe] took 2.273 secs (32.487 CPU) processed `local contrast' on CPU, blended on CPU [export]
(8 threads)  [dev_pixelpipe] took 3.174 secs (23.555 CPU) processed `local contrast' on CPU, blended on CPU [export]
(4 threads)  [dev_pixelpipe] took 3.357 secs (12.458 CPU) processed `local contrast' on CPU, blended on CPU [export]
(2 threads)  [dev_pixelpipe] took 3.764 secs (6.905 CPU) processed `local contrast' on CPU, blended on CPU [export]

For comparison,
with OpenMP disabled in src/common/locallaplacian.c:
[dev_pixelpipe] took 1.951 secs (3.594 CPU) processed `local contrast' on CPU, blended on CPU [export]
with OpenMP enabled but SSE2 codepath disabled in src/common/locallaplacian.c:
[dev_pixelpipe] took 0.937 secs (40.674 CPU) processed `local contrast' on CPU, blended on CPU [export]
and with both OpenMP and SSE2 disabled:
[dev_pixelpipe] took 3.172 secs (4.753 CPU) processed `local contrast' on CPU, blended on CPU [export]

The big culprit is gauss_reduce_sse2, but it turns out that OpenMP was configured suboptimally as well.

ralfbrown:
By using gauss_reduce() instead of gauss_reduce_sse2() and switching from dynamic to static scheduling for OpenMP, the local contrast iop with local laplacians is sped up by a factor of 25 with 64 threads when not using OpenCL.  With 8 threads, wall-time speedup is still a factor of >4 compared to before these changes.

The problem with gauss_reduce_sse2 under OpenMP is that the individual work units to be assigned to threads are just too small.

Dynamic scheduling was the wrong choice here because all iterations of the loops being parallelized take the same amount of time, modulo cache misses.  So the overhead of going back to get another work unit is quite substantial but there are no savings from better load balancing.

Fixes #4788

### DIFF
--- a/src/common/locallaplacian.c
+++ b/src/common/locallaplacian.c
@@ -612,7 +612,7 @@ void local_laplacian_internal(
     output[l] = dt_alloc_align(64, sizeof(float)*dl(w,l)*dl(h,l));
 
   // create gauss pyramid of padded input, write coarse directly to output
-#if defined(__SSE2__)
+#if defined(__SSE2__) && !defined(_OPENMP)
   if(use_sse2)
   {
     for(int l=1;l<last_level;l++)
@@ -651,7 +651,7 @@ void local_laplacian_internal(
 
     // create gaussian pyramids
     for(int l=1;l<=last_level;l++)
-#if defined(__SSE2__)
+#if defined(__SSE2__) && !defined(_OPENMP)
       if(use_sse2)
         gauss_reduce_sse2(buf[k][l-1], buf[k][l], dl(w,l-1), dl(h,l-1));
       else
--- a/src/common/locallaplacian.c
+++ b/src/common/locallaplacian.c
@@ -264,7 +264,7 @@ static inline float *ll_pad_input(
 #pragma omp parallel for default(none) \
     dt_omp_firstprivate(ht, input, max_supp, out, wd, stride) \
     shared(wd2, ht2) \
-    schedule(dynamic) \
+    schedule(static) \
     collapse(2)
 #endif // fill regular pixels:
     for(int j=0;j<ht;j++) for(int i=0;i<wd;i++)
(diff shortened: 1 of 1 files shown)


TurboGit:
Sounds promising !

The CI checks are failing because:

```
/build/darktable/src/common/locallaplacian.c:129:20: error: unused function 'gauss_reduce_sse2' [-Werror,-Wunused-function]
static inline void gauss_reduce_sse2(
                   ^
1 error generated.
```

Can you fix ?

ralfbrown:
Hmm, it had compiled without errors for me, but the fix was easy enough.

TurboGit:
@ralfbrown : sure but the CI compiles with GCC & clang and possibly different versions than yours.

MStraeten:
OSX with current Xcode & macports compiles fine

TurboGit:
Looks good to me now. Thanks.

PR #4790, opened 2020-04-21
Use dt_collection_get_query_no_group() to get selected images

### LINKED ISSUE #3154: File operations act on less files than confirmation dialog states
**Describe the bug**
When, e.g., moving a group of files to the trash, the confirmation dialog asks to confirm the trashing the whole group, but deletes only the representative image of the group.

**To Reproduce**
1. Import a directory with, i.e., both CR2 and JPG images with the same filename stub
2. Ensure that images are grouped
3. Select one group in the lighttable
4. Note that the hint in the top center states "2 images of X in current collection are selected"
5. From the "Selected image[s]" operations, select "trash"
6. Accept the confirmation dialog "do you really want to send 2 selected images to trash?"
7. Note that only 1 image disappeared from the collection

Similar behavior present for other operations, e.g., copy.

**Expected behavior**
Both files of the group to be trashed just as the confirmation dialog asks

**Screenshots**
Sorry, don't have any screenshots right now.

 - OS: Fedora 30
 - Version latest master

**Additional context**
I looked through the source for this and found the corresponding queries where the confirmation dialog just selects `main.selected_images`, and the actual control job gets a different selection passed (mainly selecting the top image of the groups, there's another query already created skipping this step). I can attempt a PR since I use file management a lot on grouped images, but I'm quite new to the code base, and side effects to changes here are a little hard to gauge…

scorpi11:
File operations from the "selected images" module worked only on the first
image of a group when grouping was enabled.

Fixes #3154

### DIFF
--- a/src/common/collection.c
+++ b/src/common/collection.c
@@ -900,7 +900,7 @@ uint32_t dt_collection_get_selected_count(const dt_collection_t *collection)
 GList *dt_collection_get(const dt_collection_t *collection, int limit, gboolean selected)
 {
   GList *list = NULL;
-  const gchar *query = dt_collection_get_query(collection);
+  const gchar *query = dt_collection_get_query_no_group(collection);
   if(query)
   {
     sqlite3_stmt *stmt = NULL;


TurboGit:
Thanks, works on my side.

PR #4799, opened 2020-04-23
Partial translation of the manual into pt_BR

jpellegrini:
Chapters 1, 2, 6, 9 of the manual have been translated Chapter 7 is mising one single string; other chapters are partially translated. Based on previous initial work by Guilherme Brondani Torri.

There is also an update to the pt_BR translation of darktable itself.


TurboGit:
Thanks !

PR #4800, opened 2020-04-23
filmic RGB v4 update : new color science

aurelienpierre:
* add a new desaturation method based on spectral purity of the light emission, independent of XYZ space, and a basic gamut mapping in pipeline RGB space following the same strategy,
* change the default saturation strategy : 0% and 100% luminance bounds are forced to 0% saturation, but the midtones are resaturated instead.
* add an highlights reconstruction method in wavelet space, bounded to the scene white value, to bloom highlights or recover details with 3 different strategies.
* hide the middle-grey settings by default, and set them to 18.45% instead. A box needs to be checked in "options" tab to get them back. These have confused too many users and overlap with exposure setting anyway.
* make the display gamma auto-adjustment optional (checkbox in "options" tab).
* allow 2 different curvatures for the filmic curve toe and shoulder : hard (4th order polynomial), with more tonal compression, and soft (3rd order polynomial), with less tonal compression.

NOTES:
* v3 algos (dt 3.0-3.0.2) are still available as an option.

WARNINGS: 
* older presets are **NOT** imported at this time,
* OpenCL kernels are **NOT** available at this time for the v4 color science and highlights recovery.

DEMO (no additional highlight reconstruction or color adjustments):
filmic v3 (2019):
Shoot Minh-Ly Toronto-0034-_DSC0242_12

filmic v4 (2020):
Shoot Minh-Ly Toronto-0034-_DSC0242_13

filmic v4 without highlights reconstruction:
6Q1B0045_02

filmic v4 with highlights reconstruction:
6Q1B0045_01

filmic v4 without highlights reconstruction:
DSC_0552_06

filmic v4 with highlights reconstruction:
DSC_0552_07

### DIFF
--- a/src/iop/filmicrgb.c
+++ b/src/iop/filmicrgb.c
@@ -105,7 +105,8 @@ typedef enum dt_iop_filmicrgb_methods_type_t
   DT_FILMIC_METHOD_NONE = 0,
   DT_FILMIC_METHOD_MAX_RGB = 1,
   DT_FILMIC_METHOD_LUMINANCE = 2,
-  DT_FILMIC_METHOD_POWER_NORM = 3
+  DT_FILMIC_METHOD_POWER_NORM = 3,
+  DT_FILMIC_METHOD_EUCLIDEAN_NORM = 4
 } dt_iop_filmicrgb_methods_type_t;
 
 
@@ -344,6 +345,15 @@ static inline float clamp_simd(const float x)
 
 
 #ifdef _OPENMP
+#pragma omp declare simd
+#endif
+static inline float sqf(const float x)
+{
+  return x * x;
+}
+
+
+#ifdef _OPENMP
 #pragma omp declare simd aligned(pixel:16)
 #endif
 static inline float pixel_rgb_norm_power(const float pixel[4])
@@ -392,6 +402,9 @@ static inline float get_pixel_norm(const float pixel[4], const dt_iop_filmicrgb_
 
     case(DT_FILMIC_METHOD_POWER_NORM):
       return pixel_rgb_norm_power(pixel);
+
+    case(DT_FILMIC_METHOD_EUCLIDEAN_NORM):
+      return sqrtf(sqf(pixel[0]) + sqf(pixel[1]) + sqf(pixel[2]));
 
     default:
       return (work_profile) ? dt_ioppr_get_rgb_matrix_luminance(pixel,
@@ -494,14 +507,6 @@ static inline float fminabsf(const float a, const float b)
   const float abs_a = fabsf(a);
   const float abs_b = fabsf(b);
   return (abs_a < abs_b) ? a : b;
-}
-
-#ifdef _OPENMP
-#pragma omp declare simd
-#endif
-static inline float sqf(const float x)
-{
-  return x*x;
 }
 
 
@@ -2710,6 +2715,7 @@ void gui_init(dt_iop_module_t *self)
   dt_bauhaus_combobox_add(g->preserve_color, _("max RGB"));
   dt_bauhaus_combobox_add(g->preserve_color, _("luminance Y"));
   dt_bauhaus_combobox_add(g->preserve_color, _("RGB power norm"));
+  dt_bauhaus_combobox_add(g->preserve_color, _("RGB euclidean norm"));
   gtk_widget_set_tooltip_text(g->preserve_color, _("ensure the original color are preserved.\n"
                                                    "may reinforce chromatic aberrations and chroma noise,\n"
                                                    "so ensure they are properly corrected elsewhere.\n"));
--- a/src/iop/filmicrgb.c
+++ b/src/iop/filmicrgb.c
@@ -2327,10 +2327,7 @@ void init(dt_iop_module_t *module)
 {
   module->params = calloc(1, sizeof(dt_iop_filmicrgb_params_t));
   module->default_params = calloc(1, sizeof(dt_iop_filmicrgb_params_t));
-
-  // Auto-apply filmic if basecurve is not
-  module->default_enabled = !dt_conf_get_bool("plugins/darkroom/basecurve/auto_apply");
-
+  module->default_enabled = 0;
   module->params_size = sizeof(dt_iop_filmicrgb_params_t);
   module->gui_data = NULL;
 
--- a/src/iop/filmicrgb.c
+++ b/src/iop/filmicrgb.c
@@ -97,7 +97,7 @@ DT_MODULE_INTROSPECTION(2, dt_iop_filmicrgb_params_t)
                       "split-ivs-in-unroller", "variable-expansion-in-unroller", \
                       "split-loops", "ivopts", "predictive-commoning",\
                       "tree-loop-linear", "loop-block", "loop-strip-mine", \
-                      "finite-math-only", "fp-contract=fast", "fast-math")
+                      "finite-math-only", "fp-contract=fast", "fast-math", "no-math-errno")
 #endif
 
 typedef enum dt_iop_filmicrgb_methods_type_t
@@ -122,6 +122,13 @@ typedef enum dt_iop_filmicrgb_colorscience_type_t
   DT_FILMIC_COLORSCIENCE_V1 = 0,
   DT_FILMIC_COLORSCIENCE_V2 = 1,
 } dt_iop_filmicrgb_colorscience_type_t;
+
+
+typedef enum dt_iop_filmicrgb_reconstruction_type_t
+{
+  DT_FILMIC_RECONSTRUCT_RGB = 0,
+  DT_FILMIC_RECONSTRUCT_RATIOS = 1,
+} dt_iop_filmicrgb_reconstruction_type_t;
 
 
 typedef struct dt_iop_filmic_rgb_spline_t
@@ -894,7 +901,9 @@ static int get_scales(const dt_iop_roi_t *roi_in, const dt_dev_pixelpipe_iop_t *
 #pragma omp declare simd aligned(in, mask, reconstructed:64) uniform(ch, variant, piece, roi_in, roi_out)
 #endif
 static inline gint reconstruct_highlights(const float *const restrict in, const float *const restrict mask,
-                                          float *const restrict reconstructed, const int variant, const size_t ch,
+                                          float *const restrict reconstructed,
+                                          const dt_iop_filmicrgb_reconstruction_type_t variant,
+                                          const size_t ch,
                                           const dt_iop_filmicrgb_data_t *const data, dt_dev_pixelpipe_iop_t *piece,
                                           const dt_iop_roi_t *const roi_in, const dt_iop_roi_t *const roi_out)
 {
@@ -976,9 +985,9 @@ static inline gint reconstruct_highlights(const float *const restrict in, const 
 
     // Compute wavelets high-frequency scales and save the maximum of texture over the RGB channels
     // Note : HF_RGB = detail - LF, HF_grey = max(HF_RGB)
-    if(variant == 0)
+    if(variant == DT_FILMIC_RECONSTRUCT_RGB)
       wavelets_detail_level_RGB(detail, LF, HF_RGB, HF_grey, roi_out->width, roi_out->height, ch);
-    else if(variant == 1)
+    else if(variant == DT_FILMIC_RECONSTRUCT_RATIOS)
       wavelets_detail_level_ratios(detail, LF, HF_RGB, HF_grey, roi_out->width, roi_out->height, ch);
 
     // interpolate/blur/inpaint (same thing) the RGB high-frequency to fill holes
@@ -988,10 +997,10 @@ static inline gint reconstruct_highlights(const float *const restrict in, const 
                                bound_top, bound_bot);
 
     // Reconstruct clipped parts
-    if(variant == 0)
+    if(variant == DT_FILMIC_RECONSTRUCT_RGB)
       wavelets_reconstruct_RGB(HF_RGB, LF, HF_grey, mask, reconstructed, roi_out->width, roi_out->height, ch,
                                gamma, gamma_comp, beta, beta_comp, delta, s, scales);
-    else if(variant == 1)
+    else if(variant == DT_FILMIC_RECONSTRUCT_RATIOS)
       wavelets_reconstruct_ratios(HF_RGB, LF, HF_grey, mask, reconstructed, roi_out->width, roi_out->height, ch,
                                gamma, gamma_comp, beta, beta_comp, delta, s, scales);
   }
@@ -1313,10 +1322,6 @@ void process(dt_iop_module_t *self, dt_dev_pixelpipe_iop_t *piece, const void *c
 
   float *restrict in = (float *)ivoid;
   float *const restrict out = (float *)ovoid;
-
-  const int variant = data->preserve_color;
-  const dt_iop_filmic_rgb_spline_t spline = (dt_iop_filmic_rgb_spline_t)data->spline;
-
   float *const restrict mask =  dt_alloc_sse_ps(roi_out->width * roi_out->height);
 
   // build a mask of clipped pixels
@@ -1330,7 +1335,7 @@ void process(dt_iop_module_t *self, dt_dev_pixelpipe_iop_t *piece, const void *c
 
     if(g->show_mask)
     {
-      display_mask(mask, out, roi_out->width, roi_out->height, 4);
+      display_mask(mask, out, roi_out->width, roi_out->height, ch);
       dt_free_align(mask);
       return;
     }
@@ -1340,7 +1345,7 @@ void process(dt_iop_module_t *self, dt_dev_pixelpipe_iop_t *piece, const void *c
 
   if(recover_highlights && mask && reconstructed)
   {
-    const gint success_1 = reconstruct_highlights(in, mask, reconstructed, 0, ch, data, piece, roi_in, roi_out);
+    const gint success_1 = reconstruct_highlights(in, mask, reconstructed, DT_FILMIC_RECONSTRUCT_RGB, ch, data, piece, roi_in, roi_out);
     gint success_2 = TRUE;
 
     if(data->high_quality_reconstruction > 0 && success_1)
@@ -1354,7 +1359,7 @@ void process(dt_iop_module_t *self, dt_dev_pixelpipe_iop_t *piece, const void *c
         for(int i = 0; i < data->high_quality_reconstruction; i++)
         {
           compute_ratios(reconstructed, norms, ratios, work_profile, DT_FILMIC_METHOD_EUCLIDEAN_NORM, roi_out->width, roi_out->height, ch);
-          success_2 = success_2 && reconstruct_highlights(ratios, mask, reconstructed, 1, ch, data, piece, roi_in, roi_out);
+          success_2 = success_2 && reconstruct_highlights(ratios, mask, reconstructed, DT_FILMIC_RECONSTRUCT_RATIOS, ch, data, piece, roi_in, roi_out);
           restore_ratios(reconstructed, norms, roi_out->width, roi_out->height, ch);
         }
       }
@@ -1368,21 +1373,21 @@ void process(dt_iop_module_t *self, dt_dev_pixelpipe_iop_t *piece, const void *c
 
   if(mask) dt_free_align(mask);
 
-  if(variant == DT_FILMIC_METHOD_NONE)
+  if(data->preserve_color == DT_FILMIC_METHOD_NONE)
   {
     // no chroma preservation
     if(data->version == DT_FILMIC_COLORSCIENCE_V1)
-      filmic_split_v1(in, out, work_profile, data, spline, roi_out->width, roi_in->height, ch);
+      filmic_split_v1(in, out, work_profile, data, data->spline, roi_out->width, roi_in->height, ch);
     else if(data->version == DT_FILMIC_COLORSCIENCE_V2)
-      filmic_split_v2(in, out, work_profile, data, spline, roi_out->width, roi_in->height, ch);
+      filmic_split_v2(in, out, work_profile, data, data->spline, roi_out->width, roi_in->height, ch);
   }
   else
   {
     // chroma preservation
     if(data->version == DT_FILMIC_COLORSCIENCE_V1)
-      filmic_chroma_v1(in, out, work_profile, data, spline, variant, roi_out->width, roi_out->height, ch);
+      filmic_chroma_v1(in, out, work_profile, data, data->spline, data->preserve_color, roi_out->width, roi_out->height, ch);
     else if(data->version == DT_FILMIC_COLORSCIENCE_V2)
-      filmic_chroma_v2(in, out, work_profile, data, spline, variant, roi_out->width, roi_out->height, ch);
+      filmic_chroma_v2(in, out, work_profile, data, data->spline, data->preserve_color, roi_out->width, roi_out->height, ch);
   }
 
   if(reconstructed) dt_free_align(reconstructed);
@@ -2221,8 +2226,7 @@ void commit_params(dt_iop_module_t *self, dt_iop_params_t *p1, dt_dev_pixelpipe_
   d->high_quality_reconstruction = p->high_quality_reconstruction;
 
   // TODO: write OpenCL for v2
-  if(p->version == DT_FILMIC_COLORSCIENCE_V2)
-    piece->process_cl_ready = FALSE;
+  piece->process_cl_ready = FALSE;
 
 
   // compute the curves and their LUT
@@ -2323,7 +2327,10 @@ void init(dt_iop_module_t *module)
 {
   module->params = calloc(1, sizeof(dt_iop_filmicrgb_params_t));
   module->default_params = calloc(1, sizeof(dt_iop_filmicrgb_params_t));
-  module->default_enabled = 0;
+
+  // Auto-apply filmic if basecurve is not
+  module->default_enabled = !dt_conf_get_bool("plugins/darkroom/basecurve/auto_apply");
+
   module->params_size = sizeof(dt_iop_filmicrgb_params_t);
   module->gui_data = NULL;
 
@@ -2331,28 +2338,28 @@ void init(dt_iop_module_t *module)
     = (dt_iop_filmicrgb_params_t){
                                  .grey_point_source   = 18.45,  // source grey
                                  .black_point_source  = -10.55f,// source black
-                                 .white_point_source  = 3.45f,  // source white
-                                 .reconstruct_threshold = 0.0f,
+                                 .white_point_source  = 5.45f,  // source white
+                                 .reconstruct_threshold = -1.0f,
                                  .reconstruct_feather   = 3.0f,
                                  .reconstruct_bloom_vs_details = 100.0f,
-                                 .reconstruct_grey_vs_color    = 0.0f,
-                                 .reconstruct_structure_vs_texture = 50.0f,
+                                 .reconstruct_grey_vs_color    = 100.0f,
+                                 .reconstruct_structure_vs_texture = 0.0f,
                                  .security_factor     = 0.0f,
                                  .grey_point_target   = 18.45f, // target grey
                                  .black_point_target  = 0.0,    // target black
                                  .white_point_target  = 100.0,  // target white
-                                 .output_power        = 5.98f,  // target power (~ gamma)
-                                 .latitude            = 40.0f,  // intent latitude
-                                 .contrast            = 1.30f,  // intent contrast
-                                 .saturation          = 0.0f,   // intent saturation
-                                 .balance             = 12.0f,  // balance shadows/highlights
+                                 .output_power        = 4.0f,   // target power (~ gamma)
+                                 .latitude            = 33.0f,  // intent latitude
+                                 .contrast            = 1.50f,  // intent contrast
+                                 .saturation          = 5.0f,   // intent saturation
+                                 .balance             = 0.0f,  // balance shadows/highlights
                                  .preserve_color      = DT_FILMIC_METHOD_POWER_NORM, // run the saturated variant
                                  .shadows             = DT_FILMIC_CURVE_POLY_4,
                                  .highlights          = DT_FILMIC_CURVE_POLY_4,
                                  .version             = DT_FILMIC_COLORSCIENCE_V2,
-                                 .auto_hardness     = TRUE,
+                                 .auto_hardness       = TRUE,
                                  .custom_grey         = FALSE,
-                                 .high_quality_reconstruction = FALSE
+                                 .high_quality_reconstruction = 1
                               };
   memcpy(module->params, &tmp, sizeof(dt_iop_filmicrgb_params_t));
   memcpy(module->default_params, &tmp, sizeof(dt_iop_filmicrgb_params_t));
@@ -2525,7 +2532,7 @@ void gui_init(dt_iop_module_t *self)
 {
   self->gui_data = malloc(sizeof(dt_iop_filmicrgb_gui_data_t));
   dt_iop_filmicrgb_gui_data_t *g = (dt_iop_filmicrgb_gui_data_t *)self->gui_data;
-  dt_iop_filmicrgb_params_t *p = (dt_iop_filmicrgb_params_t *)self->params;
+  dt_iop_filmicrgb_params_t *p = (dt_iop_filmicrgb_params_t *)self->default_params;
 
   g->show_mask = FALSE;
 
--- a/src/iop/filmicrgb.c
+++ b/src/iop/filmicrgb.c
@@ -48,6 +48,7 @@
 #include <time.h>
 
 
+#define NORM_MIN 1.52587890625e-05f // norm can't be < to 2^(-16)
 
 #define DT_GUI_CURVE_EDITOR_INSET DT_PIXEL_APPLY_DPI(1)
 
@@ -290,7 +291,7 @@ int legacy_params(dt_iop_module_t *self, const void *const old_params, const int
     n->preserve_color = o->preserve_color;
     n->shadows = DT_FILMIC_CURVE_POLY_4;
     n->highlights = DT_FILMIC_CURVE_POLY_3;
-    n->reconstruct_threshold = 3.0f; // for old edits, this ensures clipping threshold >> white level, so it's a no-op
+    n->reconstruct_threshold = 6.0f; // for old edits, this ensures clipping threshold >> white level, so it's a no-op
     n->reconstruct_bloom_vs_details = d->reconstruct_bloom_vs_details;
     n->reconstruct_grey_vs_color = d->reconstruct_grey_vs_color;
     n->reconstruct_structure_vs_texture = d->reconstruct_structure_vs_texture;
@@ -298,7 +299,7 @@ int legacy_params(dt_iop_module_t *self, const void *const old_params, const int
     n->version = DT_FILMIC_COLORSCIENCE_V1;
     n->auto_hardness = TRUE;
     n->custom_grey = TRUE;
-    n->high_quality_reconstruction = FALSE;
+    n->high_quality_reconstruction = 0;
     return 0;
   }
   return 1;
@@ -545,7 +546,7 @@ static inline gint mask_clipped_pixels(const float *const restrict in,
     mask[k / ch] = weight;
 
     // at x = 4, the sigmoid produces opacity = 5.882 %.
-    // any x > 4 will produce neglictible changes over the image,
+    // any x > 4 will produce negligible changes over the image,
     // especially since we have reduced visual sensitivity in highlights.
     // so we discard pixels for argument > 4. for they are not worth computing.
     clipped += (4.f > argument);
@@ -557,7 +558,7 @@ static inline gint mask_clipped_pixels(const float *const restrict in,
 
 
 // B spline filter
-#define fsize 5
+#define FSIZE 5
 
 
 #ifdef _OPENMP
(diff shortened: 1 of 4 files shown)


GLLM:
Looks promising on the samples @aurelienpierre !
Can't wait to test it :)
Thanks for the work !

TurboGit:
Haven't you inverted the first two pictures?

aurelienpierre:
@TurboGit no, why ?

TurboGit:
Because I find the highlight better recovered on the first than on the second. No?

aurelienpierre:
@TurboGit the highlight skin patch on the first has a magenta to yellow gradient which is telling that some individual channel clipping is going on. The second one has less texture, but a more natural color and less magenta fringing.

TurboGit:
Ok, I see. Thanks for the explanations.

TurboGit:
Error reported by CI:

```
/build/darktable/src/tests/unittests/iop/test_filmicrgb.c:375:11: error: implicit declaration of function ‘filmic_desaturate’; did you mean ‘filmic_desaturate_v2’? [-Werror=implicit-function-declaration]
  375 |           filmic_desaturate(p[0], sigma_toe, sigma_shoulder, saturation);
      |           ^~~~~~~~~~~~~~~~~
      |           filmic_desaturate_v2
```

GrahamByrnes:
I'll just applaud a couple of the goals: setting the default saturation back to zero, and changing the curvature of the toe... I was almost always using the tonecurve as a post-hoc fix. Are you planning to make the choice of 3rd vs 4th order separately selectable for the toe and shoulder?

aurelienpierre:
> Error reported by CI:

@Turbogit I don't understand what the CI did. If you search my code, there is no `filmic_desaturate` function in the file, only `filmic_desaturate_v1` or `filmic_desaturate_v2`.

> Are you planning to make the choice of 3rd vs 4th order separately selectable for the toe and shoulder?

It's already there:
Capture d’écran du 2020-04-23 21-18-55

parafin:
@aurelienpierre, you need to fix or disable the unit test.

aurelienpierre:
ah, I forget these. Thanks @parafin

aurelienpierre:
The commit #0275667 will break previous parameters structure compared to c8b8fa2.

src/iop/filmicrgb.c
```diff
@@ -1611,13 +2735,81 @@ void gui_init(dt_iop_module_t *self)
...
-                                                 "you should never touch that unless you know what you are doing."));
-  g_signal_connect(G_OBJECT(g->output_power), "value-changed", G_CALLBACK(output_power_callback), self);
+
+  // Color science
+  g->version = dt_bauhaus_combobox_new(self);
+  dt_bauhaus_widget_set_label(g->version, NULL, _("color science"));
+  dt_bauhaus_combobox_add(g->version, _("v3 (2019)"));
+  dt_bauhaus_combobox_add(g->version, _("v4 (2020)"));
+  gtk_widget_set_tooltip_text(g->version, _("v3 is darktable 3.0 desaturation method, same as color balance.\n"
+                                            "v4 is a newer desaturation method, based on spectral purity of light."));
```
burrima, line 2745: "spectral purity of light" - marketing buzzword? :-) I'm not expert in photo processing, so could be that it is a well-known term in this field?

aurelienpierre: No marketing. A white spectrum is a spectrum where every frequency has an equal weight. A pure spectrum is… well not a spectrum anymore, but almost a single frequency (kind of a laser mono-wavelength). So the idea of the v4 is to sharpen or even the shape of the light spectrum, making it more pure or more white.

src/iop/filmicrgb.c
```diff
@@ -434,13 +478,753 @@ static inline float linear_saturation(const float x, const float luminance, cons
...
+   * 1. The image coverage of each B spline filter of size `fsize` is `2^(level) * (fsize - 1) / 2 + 1` pixels
+   * 2. The coarsest level filter at full resolution should cover `1/fsize` of the largest image dimension.
+   * 3. The coarsest level filter at current zoom level should cover `scale/fsize` of the largest image dimension.
+   *
+   * So we compute the level that solves 1. subject to 3. Of course, integer rounding doesn't make that 1:1 accurate.
+   */
+  const float scale = roi_in->scale / piece->iscale;
+  const size_t size = MAX(piece->buf_in.height * piece->iscale, piece->buf_in.width * piece->iscale);
+  const int scales = floorf(log2f((2.0f * size * scale / ((fsize - 1) * fsize)) - 1.0f));
+  return CLAMP(scales, 1, MAX_NUM_SCALES);
```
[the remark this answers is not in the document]
aurelienpierre: CLAMP is an all-purpose pre-processor macro that does whatever. clamp_simd is a specialty function clamping floats between 0 and 1 and declared with OpenMP pragmas for inlining and vectorization in parallel/SIMD loops. So I use clamp_simd in pixel loops where I need performance, and CLAMP for scalar computations of indices.

src/iop/filmicrgb.c
```diff
@@ -434,13 +478,753 @@ static inline float linear_saturation(const float x, const float luminance, cons
...
+    mask[k / ch] = weight;
+
+    // at x = 4, the sigmoid produces opacity = 5.882 %.
+    // any x > 4 will produce neglictible changes over the image,
+    // especially since we have reduced visual sensitivity in highlights.
+    // so we discard pixels for argument > 4. for they are not worth computing.
+    clipped += (4.f > argument);
+  }
+
+  // If clipped area is < 9 pixels, recovery is not worth the computational cost, so skip it.
```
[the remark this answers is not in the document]
aurelienpierre: 9 pixels defines a square 3×3 patch of immediate neighbours. If you have fewer clipped pixels than that, it's likely that they are barely clipped, and on only one channel, so any downsampling interpolation algo at export should, as a side effect of the low-pass filtering it does, fill that hole. Bottom line, we don't need to unroll fancy wavelets for so little.

src/iop/filmicrgb.c
```diff
@@ -434,13 +478,753 @@ static inline float linear_saturation(const float x, const float luminance, cons
 }
 
 
+#ifdef _OPENMP
+#pragma omp declare simd
+#endif
+static inline float fmaxabsf(const float a, const float b)
```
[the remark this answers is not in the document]
aurelienpierre: By experience, there is little chance to reuse exactly that function in the future, with the exact same optimization. Code sharing is nice, but also makes people lazy when it comes to low-level optimization. All in all, you still need to rewrite them to take into account data locality, vector vs. scalar variants, vector alignment and such.

trougnouf:
I was able to easily improve [a picture](https://commons.wikimedia.org/wiki/File:%C3%89tang_de_l%27Ermite,_le_long_du_Sentier_du_Vuylbeek_(DSCF7007).jpg) which had severe artifacts on v3. Keeping v4, thank you for your great work!

src/iop/filmicrgb.c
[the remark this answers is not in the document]
burrima: Well, the function parameters clearly state that it does not take vectors. But yes, C has no function overloading and no Templates and your functions are so small that they don't cost much to compile.

src/iop/filmicrgb.c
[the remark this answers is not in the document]
burrima: Makes sense to me.

src/iop/filmicrgb.c
burrima: Thanks for your explanation. Already thought that you are not using the term for marketing :-) What market - DT is free!

burrima:
Dumb question that popped into my mind this morning: Is there a reason why the highlight reconstruction is now integrated into filmic? Would it not make sense to locate this code in the highlight reconstruction module - so that it can be used even when filmic is not used?

johnny-bit:
Highlight reconstruction module happens before demosaicing and works differently than highlight reconstruction in filmic (which is late stage in pipe). Non-demosaiced images are weird.

aurelienpierre:
The primary idea of the filmic's highlights reconstruction is actually not to reconstruct, but to ensure a spatial blending between the pixels that will be clipped and those not clipped **at the output of filmic** (not necessarily clipped by the sensor). 

Indeed, the filmic curve does the 1D blending of pixels based on their intensity (slowly degrading to white instead of an hard transition), but it has been shown that it's not enough to really blend the transition (especially with saturated blue skies).

So, the first design goal of filmic's highlights was something like a highlights blooming/blurring, to take care of the 2D blending of pixels based on their locality. But then, I couldn't decide how to drive the radius of blurring for the blooming filter. So I said, crap, let's do it multiscale and merge, then no need to decide on **one** radius. 

And since I was already there, knee-deep in wavelets, the color inpainting and texture reconstruction was only a couple of additions more in the loop. No big deal. And that, kids, is how ~~I met~~… over-engineered that stuff.

GrahamByrnes:
> A white spectrum is a spectrum where every frequency has an equal weight. A pure spectrum is… well not a spectrum anymore, but almost a single frequency (kind of a laser mono-wavelength).

Is it possible you could have another try at this? I'm familiar with the notions of white vs pink noise from audio processing, but it seems far more complicated in tri-stimulus land. Ok, a pure spectrum is an approximate frequency-domain delta function. This however would have little relation to the other interpretations of "white", ie

- a spectrum of light from a black-body radiator with an appropriate colour temperature, or some continuum approximation to that; or

- a set of discrete or continuous frequencies which result in the same perception of light temperature via the human eye.

So are you saying that non-linearities in either transmission (atmospheric effects?) or perception (well recognised multi-modal receptor filters) lead to different effects if generated in continuous vs discrete spectra?

parafin:
I guess "pure spectrum" is monochromatic light, and white light is an opposite of that - all relevant frequencies are present.

aurelienpierre:
> So are you saying that non-linearities in either transmission (atmospheric effects?) or perception (well recognised multi-modal receptor filters) lead to different effects if generated in continuous vs discrete spectra?

Perception is more than ever out of the equation. The last remaining bit (saturation as the distance between RGB channels and Y) is gone. 

But I don't get the part on transmission non-linearities. What is done here is simply using spectral purity correction as a gamut-mapping strategy, so we sharpen or even the light spectrum. Of course, it's not exact since we are using a tri-stimulus, but the principle holds.

GrahamByrnes:
> using spectral purity correction as a gamut-mapping strategy
So you are using a measure of the spectrum width (ie variance, dispersion?) as a measure? So when you wrote "pure spectrum" you were using this to mean "narrower spectrum," without getting to the point where the light becomes monochromatic, and hence not at all white?

aurelienpierre:
You could see that as a variance minimization yes, the variance being taken from the norm. But there is no hope to make the RGB signal purely monochromatic anyway, since no RGB LED is monochromatic either. 

The equation is as follow:
1. compute the RGB norm, as per v3
2. compute the RGB ratios, such that ratios = RGB / norm, as per v3. We therefore rewrite the light emission as a chromatic modulation around a metric of the light energy,
3. if ratios -> 1, the light spectrum is close to achromatic, if ratios -> 0, the light spectrum is close to monochromatic, so we correct the ratios like this : `ratios += (1 - ratios) * saturation(norm)`. The saturation is user-param in [0; 2] and forced to 0 for both ends of the dynamic range.

GrahamByrnes:
Ohhh... so you've a notion of "monochromatic white" which doesn't reflect the physics, but does capure degeneration (variance -> zero) to "pure white" and "pure black" at the upper and lower extremes of the colour space? 

Ok, that makes sense.
Thanks.

EC1000:
> * hide the middle-grey settings by default, and set them to 18.45% instead. A box needs to be checked in "options" tab to get them back. These have confused too many users and overlap with exposure setting anyway.

> hide the middle-grey settings by default, and set them to 18.45% instead. A box needs to be checked in "options" tab to get them back. These have confused too many users and overlap with exposure setting anyway.

[this post is to better understand the reasons to use one setting or the other, not at all to criticize the new design choice (but thanks to let it as an option for now!), if this discussion is more appropriate on a forum like pixl.us or darktable.fr tell me!]

I understand some of the arguments to remove this setting:
1. may be confusing to some users (who are a little too much attached to "the right number"? ^^)
2. the math and code are similar to Exposure module. 
Is it the exact same thing or is there a difference (if I just look at the result -- and in the end it is the only thing that count ! :D -- I don't find much difference, but just out of curiosity I would be interested to know if there is technical difference/limit cases)?
3. Other reason? 

I tend to use the middle-grey setting in the filmic module for the following reasons, but if there are good technical reasons to use Exposure module instead, I would test that more (it took me at least a few months to totally change from base curve to filmic, and now I am **very** happy with the results, so no reason not to change habits! :)): 

1. One less module to activate / I don't need to go back and forth between two modules.
2. If I use the Exposure module, once I have set some value >100, I can't just deactivate filmic to see how the linear image look like (I find that sometimes usefull) without having blown highlights (so I need to deactivate two modules instead of one).
3. Changing middle-grey in filmic automatically adjusts the white/black settings, whereas if I do it with Exposure, I have to change them manually.
4. Masking between Exposure (over 100) and filmic is not possible (and when it will be, since it's out of the usual scale, you will probably have to enter a value with keybord instead of the slider/mouse wheel)
5. Last argument is mainly vocabulary so not really a good one :D, but I still find strange to have to use a higher global exposure (in module) for an image that was not (physically) underexposed! I do understand that it's just an amplification, as I said it's just the name! But for that same reason, a new user would probably find his image dark if he just activate filmic without knowing it must be used with Exposure (I know, Fine Manual...).

EC1000:
> 3\. Other reason?

I answer my own question after reading a recent post [here](https://forums.darktable.fr/showthread.php?tid=4886&pid=41469#pid41469) (in French), I hope this can be of interest to others: 

3. Having a middle-grey already centered at the beginning of filmic will help to have a more stable curve (easier to control), 
4. The input color profile module (between Exposure and Filmic) is more precise for medium tones, so it will benefit of the Exposure module settings. 

PS : also for those who read French, some more detailed explainations on [this article](https://darktable.fr/2020/04/filmique-et-la-malediction-du-nombre/).

Thanks Aurélien for these explainations.

aurelienpierre:
In English, I also posted that : https://discuss.pixls.us/t/there-are-no-underexposed-digital-pictures/17916

MStraeten:
i tried to build this pr for windows and encountered some errors dealing with unspecified variables
see log:
[mingw64_filmicrgb_errors.txt](https://github.com/darktable-org/darktable/files/4598515/mingw64_filmicrgb_errors.txt)

TurboGit:
Note that you'll want to rebase as the color-picker support has been changed deeply. Easier to handle now. See #4978 & #5009.

aurelienpierre:
This is ready for review @TurboGit.

I gave up on OpenCL for this PR, it is for now entirely disabled, and I will submit another PR for OpenCL update (+1600/-300 is enough for one PR I think ;-) )

New defaults have been set and the module will be auto-enabled if basecurve is not auto-applied. The default filmic curve expects exposure to raised by 1 to 2 EV, depending on manufacturers, and is intended to match OOC Jpegs more or less.

The new saturation/gamut-sort-of-mapping is now set to 10% to resaturate mildly the midtones while enforcing 0% saturation at 0% et 100% luminance display-referred.

I have tested it in conjunction with #4940, I'm confident to be able to provide a reasonable default look/pipeline default based on exposure/filmic/channel mixer 2/color balance for the lazy photographer.

groutr:
@aurelienpierre, from a user perspective, I think it would make more sense to have the "auto adjust hardness" an eyedropper button in the "look" tab instead of a checkbox. This would be similar to the eyedropper buttons in the "scene" tab that adjust a parameter based on the image.
Having auto-adjust checked, then changing hardness seems to override the auto-adjust. Then toggling the checkbox (on -> off -> on) basically acts as a reset on the hardness.

MStraeten:
pulled again for a windows build, but  errors from https://github.com/darktable-org/darktable/pull/4800#issuecomment-625767207 are still present

aurelienpierre:
> @aurelienpierre, from a user perspective, I think it would make more sense to have the "auto adjust hardness" an eyedropper button in the "look" tab instead of a checkbox. This would be similar to the eyedropper buttons in the "scene" tab that adjust a parameter based on the image.
> Having auto-adjust checked, then changing hardness seems to override the auto-adjust. Then toggling the checkbox (on -> off -> on) basically acts as a reset on the hardness.

"Auto-adjust hardness" can't be a color picker because it has nothing to do with the image. It only adjusts the parameter formerly known as "display target power function" so the grey stays on the identity line. Therefore, hardness (= power function) is adjusted from white and black relative exposures.

> pulled again for a windows build, but errors from #4800 (comment) are still present

It's an original error. The first error actually shows that the variables are specified in the enclosing parallel.

MStraeten:
now Windows is pleased - it builds fine

aurelienpierre:
\o/ Thanks @MStraeten

TurboGit:
I should look at this soon now !

src/iop/filmicrgb.c
```diff
@@ -1293,12 +2443,26 @@ static gboolean dt_iop_tonecurve_draw(GtkWidget *widget, cairo_t *crf, gpointer
...
+      cairo_line_to(cr, x * width, height * (1.0 - y));
+    }
+  }
+  else if(p->version == DT_FILMIC_COLORSCIENCE_V2)
+  {
+    cairo_move_to(cr, 0, height * (1.0 - filmic_desaturate_v2(0.0f, sigma_toe, sigma_shoulder, saturation)));
+    for(int k = 1; k < 256; k++)
+    {
+      const float x = k / 255.0;
+      float y = filmic_desaturate_v2(x, sigma_toe, sigma_shoulder, saturation);
```
TurboGit, line 2463: const

src/iop/filmicrgb.c
```diff
@@ -1293,12 +2443,26 @@ static gboolean dt_iop_tonecurve_draw(GtkWidget *widget, cairo_t *crf, gpointer
...
+  if(p->version == DT_FILMIC_COLORSCIENCE_V1)
   {
-    const float x = k / 255.0;
-    float y = filmic_desaturate(x, sigma_toe, sigma_shoulder, saturation);
-    cairo_line_to(cr, x * width, height * (1.0 - y));
+    cairo_move_to(cr, 0, height * (1.0 - filmic_desaturate_v1(0.0f, sigma_toe, sigma_shoulder, saturation)));
+    for(int k = 1; k < 256; k++)
+    {
+      const float x = k / 255.0;
+      float y = filmic_desaturate_v1(x, sigma_toe, sigma_shoulder, saturation);
```
TurboGit, line 2453: const

src/iop/filmicrgb.c
```diff
@@ -1105,19 +2184,29 @@ void commit_params(dt_iop_module_t *self, dt_iop_params_t *p1, dt_dev_pixelpipe_
   dt_iop_filmicrgb_params_t *p = (dt_iop_filmicrgb_params_t *)p1;
   dt_iop_filmicrgb_data_t *d = (dt_iop_filmicrgb_data_t *)piece->data;
 
-  d->preserve_color = p->preserve_color;
+  // source and display greys
+  float grey_source, grey_display;
```
TurboGit, line 2188: = 0.0f for both.

src/iop/filmicrgb.c
```diff
@@ -954,8 +1916,101 @@ static void preserve_color_callback(GtkWidget *combo, dt_iop_module_t *self)
...
+}
+
+
+#define ORDER_4 5
+#define ORDER_3 4
+
+
 inline static void dt_iop_filmic_rgb_compute_spline(const dt_iop_filmicrgb_params_t *const p, struct dt_iop_filmic_rgb_spline_t *const spline)
 {
+  float grey_display;
```
TurboGit, line 2001: = 0.0f;

src/iop/filmicrgb.c
```diff
@@ -954,8 +1916,101 @@ static void preserve_color_callback(GtkWidget *combo, dt_iop_module_t *self)
...
+
+static void custom_grey_callback(GtkWidget *widget, dt_iop_module_t *self)
+{
+  if(darktable.gui->reset) return;
+  dt_iop_filmicrgb_params_t *p = (dt_iop_filmicrgb_params_t *)self->params;
+  dt_iop_filmicrgb_gui_data_t *g = (dt_iop_filmicrgb_gui_data_t *)self->gui_data;
+
+  p->custom_grey = gtk_toggle_button_get_active(GTK_TOGGLE_BUTTON(widget));
+
+  const int reset = darktable.gui->reset;
```
TurboGit, line 1946: should be:
`++darktable.gui->reset;`

src/iop/filmicrgb.c
```diff
@@ -954,8 +1916,101 @@ static void preserve_color_callback(GtkWidget *combo, dt_iop_module_t *self)
...
+  darktable.gui->reset = 1;
+  gtk_widget_set_visible(g->grey_point_source, p->custom_grey);
+  gtk_widget_set_visible(g->grey_point_target, p->custom_grey);
+  darktable.gui->reset = reset;
```
TurboGit, line 1950: and
`--darktable.gui->reset;`

src/iop/filmicrgb.c
```diff
@@ -954,8 +1916,101 @@ static void preserve_color_callback(GtkWidget *combo, dt_iop_module_t *self)
...
+  dt_iop_filmicrgb_params_t *p = (dt_iop_filmicrgb_params_t *)self->params;
+  p->auto_hardness = gtk_toggle_button_get_active(GTK_TOGGLE_BUTTON(widget));
+
+  if(p->auto_hardness)
+  {
+    dt_iop_filmicrgb_gui_data_t *g = (dt_iop_filmicrgb_gui_data_t *)self->gui_data;
+
+    p->output_power =  logf(p->grey_point_target / 100.0f) / logf(-p->black_point_source / (p->white_point_source - p->black_point_source));
+
+    const int reset = darktable.gui->reset;
```
TurboGit, line 1969: see above

src/iop/filmicrgb.c
```diff
@@ -423,117 +493,905 @@ static inline float linear_saturation(const float x, const float luminance, cons
...
+  for(size_t k = 0; k < height * width * ch; k += ch)
+  {
+    const float *const restrict pix_in = in + k;
+    float *const restrict pix_out = out + k;
+
+    float norm = get_pixel_norm(pix_in, variant, work_profile);
+    norm = (norm < 1.52587890625e-05f) ? 1.52587890625e-05f : norm; // norm can't be < to 2^(-16)
+
+    // Save the ratios
+    float DT_ALIGNED_PIXEL ratios[4];
```
TurboGit, line 1192: = { 0.0f };

src/iop/filmicrgb.c
```diff
@@ -196,62 +247,64 @@ int default_colorspace(dt_iop_module_t *self, dt_dev_pixelpipe_t *pipe, dt_dev_p
   return iop_cs_rgb;
 }
 
-void init_presets(dt_iop_module_so_t *self)
```
TurboGit, line 199: You have removed all presets? Is that expected?

aurelienpierre: Yes, I'm really not sure it's a good idea to have presets in there. People expect them to be one-size-fits-all so it's only frustration.

src/iop/filmicrgb.c
```diff
@@ -423,117 +493,905 @@ static inline float linear_saturation(const float x, const float luminance, cons
...
+        #endif
+        for(size_t jj = 0; jj < fsize; ++jj)
+          for(size_t c = 0; c < 3; ++c)
+          {
+            int index_x = mult * (jj - (fsize - 1) / 2) + j;
+            index_x = (index_x < bound_left)  ? bound_left  :
+                      (index_x > bound_right) ? bound_right :
+                                                index_x     ;
+
+            static const float DT_ALIGNED_ARRAY filter[fsize] = { 1.0f / 16.0f, 4.0f / 16.0f, 6.0f / 16.0f, 4.0f / 16.0f, 1.0f / 16.0f };
```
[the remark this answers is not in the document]
aurelienpierre: I had it out of the loop before, but it turned out to be slower since it needs to be initialized from the outside for each thread. Since it's static const, I guess OpenMP parallel loops can cache the array in the code and avoid some memory flip-flops at runtime.

aurelienpierre:
@TurboGit the delta E is not normal, the original code has not changed. Could you retry with removing `"no-math-errno"` in the `optimize` pragma below header ? It's the only explanation I can think of.

src/iop/filmicrgb.c
```diff
@@ -423,117 +493,905 @@ static inline float linear_saturation(const float x, const float luminance, cons
...
+                                  const size_t width, const size_t height, const size_t ch)
+{
+  #ifdef _OPENMP
+  #pragma omp parallel for simd default(none) \
+    dt_omp_firstprivate(ch, width, height, norms, ratios, in, work_profile, variant) \
+    schedule(simd:static) aligned(norms, ratios, in:64)
+  #endif
+  for(size_t k = 0; k < height * width * ch; k += ch)
+  {
+    float norm = get_pixel_norm(in + k, variant, work_profile);
```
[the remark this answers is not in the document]
aurelienpierre: Why clamp ? We only need to ensure the lower bound and that function is not vectorizable.

src/iop/filmicrgb.c
aurelienpierre: Oh, that got lost during rebase

src/iop/filmicrgb.c
TurboGit: Your take, but I use them as base work. I think people are used to them now.

TurboGit:
> the delta E is not normal, the original code has not changed. Could you retry with removing `"no-math-errno"` in the `optimize` pragma below header ?

That's the first test I did. Still same issue without this option.

src/iop/filmicrgb.c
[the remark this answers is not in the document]
TurboGit: So maybe a macro to share the code for all values ?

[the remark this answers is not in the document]
parafin: When it was outside, was it declared as static const too? Because IMHO it shouldn't matter where static const declaration is put, it may as well be a global variable - it's allocated in the same space and not initialized at runtime, but at compile time.

src/iop/filmicrgb.c
```diff
@@ -423,117 +493,905 @@ static inline float linear_saturation(const float x, const float luminance, cons
...
+      return;
+    }
+  }
 
-      // Get the desaturation value based on the log value
-      const float desaturation = filmic_desaturate(norm, data->sigma_toe, data->sigma_shoulder, data->saturation);
+  float *const restrict reconstructed = dt_alloc_sse_ps(roi_out->width * roi_out->height * ch);
 
-      for(int c = 0; c < 3; c++) ratios[c] *= norm;
+  if(recover_highlights && mask && reconstructed)
```
TurboGit, line 1346: could the diff I'm seeing on the integration test be there ? shouldn't this be activated only if using new v4 algo ?

In legacy_param() we have:

`    n->reconstruct_threshold = 3.0f; // for old edits, this ensures clipping threshold >> white level, so it's a no-op`

Is the no-op really the case ?

aurelienpierre: There is no option to disable recovery, only you can use crazy high threshold for the highlight masking, so it is practically disabled. That could indeed be the cause, in that case this value needs to be increased (although it's actually a good thing, anyway, that highlights are a bit blurred where they clip).

src/iop/filmicrgb.c
aurelienpierre: There is a large dependency to camera, for presets accuracy/efficiency, so I would rather encourage users to build their own.

src/iop/filmicrgb.c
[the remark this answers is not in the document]
aurelienpierre: Yes, it was static const both cases, but I had to declare is first_private anyway for OpenMP, so I suspect it is the cause. Using a macro seems fine.

src/iop/filmicrgb.c
```diff
@@ -423,117 +493,905 @@ static inline float linear_saturation(const float x, const float luminance, cons
...
+    clipped += (4.f > argument);
+  }
+
+  // If clipped area is < 9 pixels, recovery is not worth the computational cost, so skip it.
+  return (clipped > 9);
+}
+
+
+// B spline filter
+#define fsize 5
```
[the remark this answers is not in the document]
aurelienpierre: done

src/iop/filmicrgb.c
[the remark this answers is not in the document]
aurelienpierre: It doesn't look like we can define arrays in macros

src/iop/filmicrgb.c
```diff
@@ -423,117 +493,905 @@ static inline float linear_saturation(const float x, const float luminance, cons
...
+  {
+    const float alpha = mask[k / ch];
+
+    // cache RGB wavelets scales just to be sure the compiler doesn't reload them
+    const float DT_ALIGNED_ARRAY HF_c[4] = { HF[k], HF[k + 1], HF[k + 2], HF[k + 3] };
+    const float DT_ALIGNED_ARRAY LF_c[4] = { LF[k], LF[k + 1], LF[k + 2], LF[k + 3] };
+
+    // synthesize the max of all RGB channels texture as a flat texture term for the whole pixel
+    // this is useful if only 1 or 2 channels are clipped, so we transfer the valid/sharpest texture on the other channels
+    float grey_texture = gamma * texture[k / ch];
```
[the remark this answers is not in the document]
aurelienpierre: done

src/iop/filmicrgb.c
```diff
@@ -423,117 +493,905 @@ static inline float linear_saturation(const float x, const float luminance, cons
...
+    // this is smoother than grey_texture and will fill holes smoothly in details layers if grey_texture ~= 0.f
+    float grey_details = gamma_comp * fmaxabsf(fmaxabsf(HF_c[0], HF_c[1]), HF_c[2]);
+
+    // synthesize both terms with weighting
+    // when beta_comp ~= 1.0, we force the reconstruction to be achromatic, which may help with gamut issues or magenta highlights.
+    const float grey_HF = beta_comp * (grey_details + grey_texture);
+
+    // synthesize the min of all low-frequency RGB channels as a flat structure term for the whole pixel
+    // when beta_comp ~= 1.0, we force the reconstruction to be achromatic, which may help with gamut issues or magenta highlights.
+    float grey_residual = beta_comp * fminf(fminf(LF_c[0], LF_c[1]), LF_c[2]);
```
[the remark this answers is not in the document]
aurelienpierre: done

src/iop/filmicrgb.c
```diff
@@ -423,117 +493,905 @@ static inline float linear_saturation(const float x, const float luminance, cons
...
+    const float DT_ALIGNED_ARRAY HF_c[4] = { HF[k], HF[k + 1], HF[k + 2], HF[k + 3] };
+    const float DT_ALIGNED_ARRAY LF_c[4] = { LF[k], LF[k + 1], LF[k + 2], LF[k + 3] };
+
+    // synthesize the max of all RGB channels texture as a flat texture term for the whole pixel
+    // this is useful if only 1 or 2 channels are clipped, so we transfer the valid/sharpest texture on the other channels
+    float grey_texture = gamma * texture[k / ch];
+
+    // synthesize the max of all interpolated/inpainted RGB channels as a flat details term for the whole pixel
+    // this is smoother than grey_texture and will fill holes smoothly in details layers if grey_texture ~= 0.f
+    float grey_details = gamma_comp * fmaxabsf(fmaxabsf(HF_c[0], HF_c[1]), HF_c[2]);
```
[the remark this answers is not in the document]
aurelienpierre: done

src/iop/filmicrgb.c
```diff
@@ -423,117 +493,905 @@ static inline float linear_saturation(const float x, const float luminance, cons
...
+
+    // synthesize both terms with weighting
+    // when beta_comp ~= 1.0, we force the reconstruction to be achromatic, which may help with gamut issues or magenta highlights.
+    const float grey_HF = beta_comp * (grey_details + grey_texture);
+
+    // synthesize the min of all low-frequency RGB channels as a flat structure term for the whole pixel
+    // when beta_comp ~= 1.0, we force the reconstruction to be achromatic, which may help with gamut issues or magenta highlights.
+    float grey_residual = beta_comp * fmaxf(fmaxf(LF_c[0], LF_c[1]), LF_c[2]);
```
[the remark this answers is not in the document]
aurelienpierre: done

src/iop/filmicrgb.c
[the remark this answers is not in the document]
aurelienpierre: macro done, and const added.

src/iop/filmicrgb.c
aurelienpierre: can't, `y` is changed later

src/iop/filmicrgb.c
aurelienpierre: done

src/iop/filmicrgb.c
aurelienpierre: initialized with 0.18 and 0.46 instead, but same result.

src/iop/filmicrgb.c
aurelienpierre: done

src/iop/filmicrgb.c
aurelienpierre: done

src/iop/filmicrgb.c
aurelienpierre: done

src/iop/filmicrgb.c
aurelienpierre: done

src/iop/filmicrgb.c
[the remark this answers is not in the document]
parafin: Have you tried making it shared instead of first_private then? I see that first_private is considered a better choice, though I'm not too sure why since I don't know OpenMP well enough.

TurboGit:
@aurelienpierre : I'll push my FilmicRGB integration test.

src/iop/filmicrgb.c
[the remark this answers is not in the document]
aurelienpierre: Shared is really bad: it means only one thread can access the variable at one time, so every thread needs to wait for its turn to lock it. It's what you want to avoid at all cost, since it usually destroys the whole point of using multithreading. The goal here is to make the filter 100% private in threads.

aurelienpierre:
@TurboGit changes pushed

src/iop/filmicrgb.c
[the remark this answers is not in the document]
parafin: So it's impossible with OpenMP to access global memory directly, even constant? OK then.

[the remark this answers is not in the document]
aurelienpierre: I wouldn't say I fully understand what OpenMP does here, but it seems indeed that its threads don't care about global memory.

TurboGit:
@aurelienpierre : I tested it and found it functional. No issue trying most sliders.

I have tested again the regression tests and got similar small dE. The diff image compared to the expected output is somewhat "worrisome". It seems that some random pixels get changed, here is a crop area with red pixels meaning diff:

image

The full diff image is:

diff-ok

Maybe you'll have an idea but this diff seems to rule out the threshold we were talking about as possible culprit.

aurelienpierre:
Weird. The only change I could see responsible for this is in `filmic_spline()` function, where I removed the vectorized masking that turned out to be slower than the naive version. Anyway, if it is random, it sounds like floating point arithmetic issues, nothing systematic, and with dE < 0.8, I think we should be fine.

groutr:
> > @aurelienpierre, from a user perspective, I think it would make more sense to have the "auto adjust hardness" an eyedropper button in the "look" tab instead of a checkbox. This would be similar to the eyedropper buttons in the "scene" tab that adjust a parameter based on the image.
> > Having auto-adjust checked, then changing hardness seems to override the auto-adjust. Then toggling the checkbox (on -> off -> on) basically acts as a reset on the hardness.
> 
> "Auto-adjust hardness" can't be a color picker because it has nothing to do with the image. It only adjusts the parameter formerly known as "display target power function" so the grey stays on the identity line. Therefore, hardness (= power function) is adjusted from white and black relative exposures.

perhaps if the "auto-adjust hardness" checkbox disabled the hardness slider, or made it disappear altogether similar to how "use custom middle-grey" checkbox makes the middle-grey slider appear.
What was confusing to me was that I was able to adjust hardness even though it was supposed to be auto adjusted, then the only way get back to auto-adjusted hardness was to toggle the checkbox.

rkowalke:
Thanks for your amazing work, @aurelienpierre. I have just finished reworking of 22k photos from the last decade with darktable 3.0 and Filmic V3, which was already a hugh step forward compared to my previous basecurve-based edits. So it was time for me to try out the current dev state of Filmic V4 and I'm really glad that my V3 XMP files could be imported without any problems and almost any photo looks like in dt 3.0 (with V3 option set).

However, I found some smaller issues with V4:

# Blue artifacts
I discovered blue artifacts under V4 in combination with Luminance Y for a photo taken at St. Pauls Cathedral at a light installation (assume modern LED lights). 

I cannot reproduce this neither with V3 color science option +  Luminance Y    nor   with V4 + RGB power norm and otherwise identical settings. 

This image needs correcting of WB, which was done with CAT of your other PR #4940. If I disable channel mixer RGB, there are less artifacts but still present, so I assume they are caused by Filmic V4 but boosted by WB correction.

Just a guess: If you look at the wavelet histograms, the red and green channels seem to be muted at the artifacts locations (compared to V3).

Compare Filmic V3 vs V4 - Luminance Y
Compare Filmic V3 vs  V4 - RGB power norm
Compare histograms

# German translations missing
- Tab "reconstruct" => "Rekonstruieren" or "Wiederherstellen"
  - "highlights clipping threshold" => "Lichter Abschneideschwellwert"
  - "highlights clipping transition" => "Lichter Abschneideübergang"
  - "balance structure/texture" => "Struktur/Textur - Balance"
  - "balance bloom/reconstruct" => "Überstrahlung/Wiederherstellung - Balance"
  - "balance grey/colorful details" => "Graue/Farbenfrohe Details - Balance"
- Tab "look"
  - "hardness" => "Härtegrad"
- Tab "options" => "Optionen"
  - "color science" => "Farbenlehre" or "Farbwissenschaft" or "Entsättigungsmethode" ?
  - "contrast in highlights" => "Kontrast in Lichtern"
  - "contrast in shadows" => "Kontrast in Schatten"
  - "use custom middle-grey values" => "Eigenen Wert für Mittelgrau verwenden"
  - "auto adjust hardness => "Härtegrad automatisch einstellen"
  - "iterations of high-quality reconstruction" => "Iterationen der hochqualitativen Wiederherstellung"

aurelienpierre:
@rkowalke you need to understand that no norm in there is perfect, otherwise there wouldn't be 4 of them. The issue you have here is out-of-gamut blue that might still end-up reasonably in the middle of the luminance range so it would not get the filmic desaturation (or even get resaturated if that's what you asked).

If using channel mixer reboot, you can increase the gamut compression factor. Otherwise, just stay away from the Y norm.

jenshannoschwalm:
@aurelienpierre i didn't test your new stuff until now...

Congrats ... REALLY GREAT STUFF.

I just gave it a hard try, took ~2000 shots in very different situations, colorful Utah landscape, german autumn mist, technical stuff, architectural, people. Switched to your suggested workflow, cleared history and voila. 

Almost all images are at least good, most **very** good. 

Thanks for doing this :-)

chhil:
@aurelienpierre  
Will you be updating the auto filmic style for v4?

GrahamByrnes:
I tried it quickly last night... I was finally happy once I had pushed most of the adjustments to either max or min, viz reconstruction threshold etc. The default colour looked "nice", in a slightly bloomish, hand coloured way.

At the moment it doesn't do much for me... but then I'm a B&W photographer and I try hard to avoid excess texture, fake contrast and so on, so I wanted to find that minimalist option. It will take work to find the options I need.

BTW, at some stage I found that Base Curve had turned itself on again... weird bug?

Issue #4250, opened 2020-02-07
SOME params in xmp human legible and editable for timelapse photography and darktable


AxelG-DE:
**Is your feature request related to a problem? Please describe.**
Basically my todays wish is directly linked with my thread in pixls.us, I started, called [Timelapse photography and darktable](https://discuss.pixls.us/t/timelapse-photography-and-darktable/14579)

My major intention is, to push dt a little bit into a direction, where is a *chance*, dt one day can be compatible with [LRTimelapse](https://lrtimelapse.com)

I do not want anybody to fully make dt adopt to LRT, I just would like to see, that dt comes to a stage, it can have an interface. Please see above discussion, which I hope my intention and motivation becomes clear.

**Describe the solution you'd like**
As I have described in my [post #37](https://discuss.pixls.us/t/timelapse-photography-and-darktable/14579/37), I would like to get (for the beginning) 

- two new modules, which are a modification of the today's existing exposure module and white balance module
- to be called _exposure timelapse_ and _white balance timelapse_ or _exposure TL_, _white balance TL_
- having their major setting (like temp and tint for WB and exposure for EXP) extracted from the darktable:params="base64-ziped-code" and store it in a human legible and script editable value (just like RT does)

e.g.
```
darktable:operation="temperature"
darktable:enabled="1"
darktable:modversion="3"
darktable:params="046c32400000803f8c4dd63f0000807f" (example !)
darktable:temperature="5748"
darktable:tint="1.012"
```
For many reasons (not least backwards compatibility), I think, it is necessary to implement this in separate, new modules and leave the todays existing ones untouched. But this is my non-coding-guy's view

For any background questions, please first consult [Timelapse photography and darktable](https://discuss.pixls.us/t/timelapse-photography-and-darktable/14579) and please also note my disclaimer in the initial post.

Thank you very much in advance!

parafin:
It's closely related to #3215 (or even a duplicate).

AxelG-DE:
@parafin: 
thank you for looking up and connect. 
Somehow I hope it will not be considered to be a duplicate (even I admit they are quite similar, for motivation I donno) and finally just closed, as the discussions around this topic and the above mentioned thread are not that simple :-)

AxelG-DE:
I still hope, this function would come live. Hence I'd stop the bot :)

aster94:
[...]

PR #4804, opened 2020-04-24
[RFC] IOP Introspection completion, (partial) refactoring and janitorial work.

dterrahe:
This is meant as a discussion document, very much a work in progress. I will try to update it here based on input I receive in the comments and on IRC. Hopefully there is interest to proceed with some form of it. If there is a better location to organise a discussion, please let me know that as well. Or if someone else wants to lead/organise it.

# 1. INTRODUCTION

A couple of weeks back there was a short discussion with @houz on IRC about using introspection to enable/enhance export _and import_ of human readable versions of module parameters. The current introspection framework already allows specifying minimum and maximum values (and type etc) for each parameter, but this feature is not widely used and couldn't be guaranteed to be in sync with the actual restrictions that are enforced in the module gui.

This started me thinking about making the definition of dt_iop_params_t leading in the creation of module interfaces (gui and otherwise) and possibly also reducing the complexity/opacity of and duplications in the gui aspect of each iop module. As the wiki points out:

> There is duplicated code in the software that could be factorized and put into libraries. 

After playing around with some ideas and checking their viability, I would now like to see if there is interest in going ahead with this and to agree on the design of the basic components. I am offering to take care of a substantial part of the clean up work across all modules, but if others are willing to take some, I would be happy with that.

In first instance, this pull request is a partial implementation to illustrate the infrastructure ideas and how these could be used in a few modules, as examples. It is just meant to facilitate discussion, the conclusions of which I would use to update this first comment. Actual commits of complete infrastructure work and module updates would all go into separate pull requests (see 5.PROJECT below)

# 2. GOALS

Create complete and consistent introspection information for all iop modules.
Reduce the amount of boiler plate required to define module gui; make remaining code more readable/maintainable.
Continue to fully support the current way of defining the user interface for highly taylored modules; conversion/adoption optional.
Achieve more consistency by providing defaults (that can be fully taylored using existing approaches).

# 3. INFRASTRUCTURE

I have spend a lot of time browsing through the ui code of a lot of the iop modules and collecting data on the features being used in defining their interfaces in gui_init ([slider_new_grep.xlsx](https://github.com/darktable-org/darktable/files/4530229/slider_new_grep.xlsx)). There is obviously a lot of similarity and boilerplate code, but there are significant differences as well, so it is impossible to come up with a one size fits all. What I am aiming for is provide componentised utility functions with sensible defaults to help with the largest common denominator, where all the existing customisation can still be added on top.

## 3.1 Widget Creation

Currently, each gui_init function builds the whole module ui from scratch, without referring to the introspection information that is collected from the dt_iop_params_t. The first step is to populate more introspection comments ($MIN, $MAX, $DEFAULT) and then refer to those values in the widget creation. The changes in basicadj.c in this pull request illustrate this.

### 3.1.1 Sliders

For now, only the creation of sliders is implemented. There's a new call in bauhaus.c that replaces much of the boiler plate and then after that any tailoring can still be done in the usual way. Besides creating the sliders using the defaults from introspection, a generic callback function is installed that reads any changes in the slider back into the param structure. 

### 3.1.2 Add more properties

Some of the settings required for the most common sliders cannot yet be specified in introspection comments. I would therefore suggest adding fields for $SOFT_MIN and $SOFT_MAX and possibly $STEP and $DIGITS. This would no longer fit on one line with the parameter type+name, so maybe a shorthand formatting could be introduced. Something like:
`float x, // -16 -4 0 4 16 .01 3`
Very much open for discussion. Maybe @houz would be willing to implement something in the perl code (mine is very rusty)?
Actually, the step size and number of digits are commonly 1/100th of the max and the equivalent number of digits, so just using the defaults often is sufficient. The format is then basic on number of digits. I've added a + if range is negative and positive and no + if only positive numbers. This is not done everywhere currently, but might be nice to have this consistently?
Most formats are variants of %, EV and degrees, so if we could flag the type (for example by adding $P, $E or $D) then we wouldn't need to pass the (partial) format to the dt_bauhaus_slider_new_from_params_box call.
I'm using the $DESCRIPTION comment as label, or if it isn't provided, just the variable name (changing _ to space). This should probably also be done in the perl code, so that it is constistent everywhere.

### 3.1.3 Comboboxes & Checkboxes

In a way similar to sliders, comboboxes and checkboxes could be set up in a utility function call, with a type specific callback function to read changed parameters back to dt_iop_params_t. For comboboxes introspection needs to be extended to allow descriptive text to be added to each enum definition.

## 3.2 Value-Changed Callbacks

For each type, float (slider), int (combobox) and bool (checkbox) a generic callback function is needed in bauhaus.c. I've already set up a possible implementation of the slider version. This gets connected to each slider in the module, with an extra paramater for the field name. It includes all the standard boilerplate to pick up new values and create a history entry. Many modules don't do anything more complicated than this but currently still need to define a complete set of callback functions for each field.

If any further effects are required (changing other widgets values or enabling/disabling them etc) then control can be passed from within the generic value-changed callback to a 1-per-module overloaded callback, to update fields/bounderies etc.
In filmicrgb, this could look like:
```C
static void value_changed_callback(GtkWidget *w, dt_iop_module_t *self, void *prev)
{
  dt_iop_filmicrgb_params_t *p = (dt_iop_filmicrgb_params_t *)self->params;
  dt_iop_filmicrgb_gui_data_t *g = (dt_iop_filmicrgb_gui_data_t *)self->gui_data;

  if (w = g->security_threshold || w = g->black_point_source || w = g->white_point_source)
  {
    if (w = g->security_threshold)
    {
      float previous = *(float*)prev;
      float ratio = (p->security_factor - previous) / (previous + 100.0f);

      float EVmin = p->black_point_source;
      EVmin = EVmin + ratio * EVmin;

      float EVmax = p->white_point_source;
      EVmax = EVmax + ratio * EVmax;

      p->white_point_source = EVmax;
      p->black_point_source = EVmin;

      dt_bauhaus_slider_set_soft(g->white_point_source, p->white_point_source);
      dt_bauhaus_slider_set_soft(g->black_point_source, p->black_point_source);
    }

    p->output_power =  logf(p->grey_point_target / 100.0f) / logf(-p->black_point_source / (p->white_point_source - p->black_point_source));

    dt_bauhaus_slider_set_soft(g->output_power, p->output_power);
  }
}
```
The call from the generic value-changed callback would be wrapped in
```C
darktable.gui->reset = 1;
...
darktable.gui->reset = reset;
```
and also passes on the previous/old version of the field that was changed. This covers all the situations we see now but with much less duplication.

## 3.3 Field_name enums

The existing introspection framework only refers to fields by their name, which leads to many string lookups. It would be useful to generate an enum that provides a number for each field and that can be used within the module itself to refer to fields more efficiently. The value changed callback functions for example could use that enum. 

## 3.4 Localisation 

Any labels that are defined via introspection (based on field name or comment text) need to be localised, so the introspection_.c files need to be processed by xgettext.

## 3.5 Tooltips

In my example code tooltips are still set up in the old way and are not available via introspection. There might be a way to define them immediately after dt_iop_params_t as an array of tuple pairs "field enum", "string"), so that the generic widget creation functions could attach them.

## 3.6 Multiple Widgets

If most widgets don't require any setup beyond the defaults and introspection-provided values, it would be possible to have their creation entirely driven from the list of fields available via introspection. 

### 3.6.1 Range

For the simplest modules, gui_data could contain an array of widgets (rather than individually named widget fields as now), that can be indexed using the field_name enums, and gui_init could be a simple call to `create_widget_range(1, last_field_enum_value)`. Or if the module contains pages/headings or custom widgets, these could be set up something like this:
```C
GtkWidget *page1, *page2, *page3;
self->widget = page1 = GTK_WIDGET(gtk_box_new(GTK_ORIENTATION_VERTICAL, 0));
create_widget_range(self, f_grey_point_source, f_security_factor);
self->widget = page2 = GTK_WIDGET(gtk_box_new(GTK_ORIENTATION_VERTICAL, 0));
create_widget_range(self, f_latitude, f_preserve_color);
etc
```

### 3.6.2 Default gui_init()

If there is no tailoring required at all, gui_init does not need to be defined at all, causing a fall-back to the default initialisation with just generates widgets for all the fields in the introspection.

Similarly, the init method and gui_update could all have a default implementation that loops over all the fields in introspection and picks up their defaults or loads the values from a param struct into the corresponding widgets. 

# 4. FURTHER IDEAS

## 4.1 Sub-Sections

For modules with a more complicated structure a tailor made gui_init would still be required. But I have played around a bit, in filmicrgb.c with ways that this structure can be specified in the comments of the param_t definition, using the $PAGE keyword (or $SECTION keyword). These could either be translated into dummy "fields" in the introspection structure, or be indicated by flags for each field (new_page, new_section).

If the order of the widgets in the module should be different than in the param structure (which can't be changed without creating a new version) this can be indicated with the $AFTER keyword (which is not completely trivial to implement).

## 4.2 Versions:

instead of creating (copy/pasting) new struct for each type, put type version markers $VERSION in type definition and 
let introspection calc type sizes for each version. (or even generate additional partial typedefs, by adding _v1_t etc).
Introspection can then tell that version -1 is a same ordered subset.
Only works if only adding params, but can start at any point (now. version 3. whatever)
And legacy_params might still need to do some transforming.

# 5. PROJECT

I've tried to phrase this brain storming text a little like a menu of options. Maybe there is no apitite for drastic changes, but on the other hand the maximum space and complexity changes can be achieved by being a little ambitious and that would put the code in a state where future changes are less daunting.
To get the maximum benefit from this project, some changes need to be made to all modules (introspection comment tags) whereas others can be made over time or not at all ("refactoring", cleanup, simplification). So I would suggest the following steps:
- [ ] Infrastructure work
  - [ ] introspection
  - [ ] bauhaus
- [ ] Documentation
  - [ ] A few easier to harder modules as showcase
  - [ ] useless.c
- [ ] Bulk
  - [ ] Split modules across participants
  - [ ] Submit separate PR for each; involve main authors
        Decide whether non-default settings need to be kept case-by-case
  - [ ] Complex modules just introspection first
  - [ ] Complex modules deeper adjustments, possibly requiring further infrastrure work.
        Mark any code that is not brought up to new "standard", so it won't be copied.

# Appendix A. CHALLENGE AREAS

Slider = 100*param, (1-param)*100 etc.
Buttons in module.
Read only widgets.

# Appendix B. QUESTIONS:

- I find the definitions of "hard" and "soft" min/max confusing and I'm not sure if it is fully implemented as intended (especially what happens with a reset).
- why are get_p and get_f unrolled (if(!strcmp(name, "black_point")) return &(p->black_point); etc) rather than looping over introspection_linear?
- Should gui defaults be initialised from params or default_params (or by definition the same)
in gui_update, what's diff between dt_bauhaus_slider_set_soft and dt_bauhaus_slider_set? (can we automate gui_update?)
- For legacy version 1->2->3->..->n garanteed to be 1->n? (some legacy_params go through all legacy upgrades. Some just go directly from each old version to current)

# Appendix C. COMMON BUGS:

# Appendix D. PEOPLE I HOPE ARE INTERESTED
@houz @TurboGit @AlicVB @johnny-bit @elstoc 
Please include anybody who might have strong opinions, interest, knowledge. Thanks!

### DIFF
--- a/src/iop/filmicrgb.c
+++ b/src/iop/filmicrgb.c
@@ -128,18 +128,21 @@ typedef struct dt_iop_filmic_rgb_spline_t
 
 typedef struct dt_iop_filmicrgb_params_t
 {
-  float grey_point_source;
-  float black_point_source;
-  float white_point_source;
-  float security_factor;
-  float grey_point_target;
-  float black_point_target;
-  float white_point_target;
-  float output_power;
-  float latitude;
-  float contrast;
-  float saturation;
-  float balance;
+  // $PAGE: "scene"
+  float grey_point_source;  // $MIN: .1 $MAX: .36 $DEFAULT: 9.225 $DESCRIPTION: "middle grey luminance" $P $C
+  float black_point_source; // $MIN: -14 $MAX: -3 $DEFAULT: -10.55 $DESCRIPTION: "black relative exposure" $E $C $AFTER: white_point_source
+  float white_point_source; // $MIN: 2 $MAX: 8 $DEFAULT: 3.45 $DESCRIPTION: "white relative exposure" $E $C
+  float security_factor;    // $MIN: -50 $MAX: 50 $DEFAULT: 0 $DESCRIPTION: "dynamic range scaling" $P
+  // $PAGE: "display" $AFTER: preserve_color
+  float grey_point_target;  // $MIN: .1 $MAX: 50 $DEFAULT: 18.45 $DESCRIPTION: "target middle grey" $P
+  float black_point_target; // $MIN: 0 $MAX: 100 $DEFAULT: 0 $DESCRIPTION: "target black luminance" $P
+  float white_point_target; // $MIN: 0 $MAX: 100 $DEFAULT: 100 $DESCRIPTION: "target white luminance" $P
+  float output_power;       // $MIN: 1 $MAX: 10 $DEFAULT: 5.98 $DESCRIPTION: "target power transfer function"
+  // $PAGE: "look" $AFTER: security_factor
+  float latitude;           // $MIN: 5 $MAX: 45 $DEFAULT: 45 $P
+  float contrast;           // $MIN: .1 $MAX: 2 $DEFAULT: 1.30
+  float saturation;         // $MIN: -50 $MAX: 50 $DEFAULT: 5 $DESCRIPTION: "extreme luminance saturation" $P
+  float balance;            // $MIN: -50 $MAX: 50 $DEFAULT: 12 $DESCRIPTION: "shadows/highlights balance" $P
   int preserve_color;
 } dt_iop_filmicrgb_params_t;
 
@@ -1431,7 +1434,6 @@ void gui_init(dt_iop_module_t *self)
 {
   self->gui_data = malloc(sizeof(dt_iop_filmicrgb_gui_data_t));
   dt_iop_filmicrgb_gui_data_t *g = (dt_iop_filmicrgb_gui_data_t *)self->gui_data;
-  dt_iop_filmicrgb_params_t *p = (dt_iop_filmicrgb_params_t *)self->params;
 
   self->widget = gtk_box_new(GTK_ORIENTATION_VERTICAL, DT_BAUHAUS_SPACE);
   dt_gui_add_help_link(self->widget, dt_get_help_url(self->op));
@@ -1458,12 +1460,13 @@ void gui_init(dt_iop_module_t *self)
 
   dtgtk_justify_notebook_tabs(g->notebook);
 
+  GtkWidget *savedSelfWidget = self->widget;
+
+  self->widget = page1;
+
   // grey_point_source slider
-  g->grey_point_source = dt_bauhaus_slider_new_with_range(self, 0.1, 36., 0.1, p->grey_point_source, 2);
+  g->grey_point_source = dt_bauhaus_slider_new_from_params_box(self, "grey_point_source", " %%");
   dt_bauhaus_slider_enable_soft_boundaries(g->grey_point_source, 0.0, 100.0);
-  dt_bauhaus_widget_set_label(g->grey_point_source, NULL, _("middle grey luminance"));
-  gtk_box_pack_start(GTK_BOX(page1), g->grey_point_source, FALSE, FALSE, 0);
-  dt_bauhaus_slider_set_format(g->grey_point_source, "%.2f %%");
   gtk_widget_set_tooltip_text(g->grey_point_source, _("adjust to match the average luminance of the image's subject.\n"
                                                       "the value entered here will then be remapped to 18.45%.\n"
                                                       "decrease the value to increase the overall brightness."));
@@ -1473,11 +1476,8 @@ void gui_init(dt_iop_module_t *self)
   g_signal_connect(G_OBJECT(g->grey_point_source), "quad-pressed", G_CALLBACK(dt_iop_color_picker_callback), &g->color_picker);
 
   // White slider
-  g->white_point_source = dt_bauhaus_slider_new_with_range(self, 2.0, 8.0, 0.1, p->white_point_source, 2);
+  g->white_point_source = dt_bauhaus_slider_new_from_params_box(self, "white_point_source", " EV");
   dt_bauhaus_slider_enable_soft_boundaries(g->white_point_source, 0.0, 16.0);
-  dt_bauhaus_widget_set_label(g->white_point_source, NULL, _("white relative exposure"));
-  gtk_box_pack_start(GTK_BOX(page1), g->white_point_source, FALSE, FALSE, 0);
-  dt_bauhaus_slider_set_format(g->white_point_source, _("%+.2f EV"));
   gtk_widget_set_tooltip_text(g->white_point_source, _("number of stops between middle grey and pure white.\n"
                                                        "this is a reading a lightmeter would give you on the scene.\n"
                                                        "adjust so highlights clipping is avoided"));
@@ -1487,11 +1487,8 @@ void gui_init(dt_iop_module_t *self)
   g_signal_connect(G_OBJECT(g->white_point_source), "quad-pressed", G_CALLBACK(dt_iop_color_picker_callback), &g->color_picker);
 
   // Black slider
-  g->black_point_source = dt_bauhaus_slider_new_with_range(self, -14.0, -3.0, 0.1, p->black_point_source, 2);
+  g->black_point_source = dt_bauhaus_slider_new_from_params_box(self, "black_point_source", " EV");
   dt_bauhaus_slider_enable_soft_boundaries(g->black_point_source, -16.0, -0.1);
-  dt_bauhaus_widget_set_label(g->black_point_source, NULL, _("black relative exposure"));
-  gtk_box_pack_start(GTK_BOX(page1), g->black_point_source, FALSE, FALSE, 0);
-  dt_bauhaus_slider_set_format(g->black_point_source, _("%+.2f EV"));
   gtk_widget_set_tooltip_text(g->black_point_source, _("number of stops between middle grey and pure black.\n"
                                                        "this is a reading a lightmeter would give you on the scene.\n"
                                                        "increase to get more contrast.\ndecrease to recover more details in low-lights."));
@@ -1501,11 +1498,8 @@ void gui_init(dt_iop_module_t *self)
   g_signal_connect(G_OBJECT(g->black_point_source), "quad-pressed", G_CALLBACK(dt_iop_color_picker_callback), &g->color_picker);
 
   // Security factor
-  g->security_factor = dt_bauhaus_slider_new_with_range(self, -50., 50., 1.0, p->security_factor, 2);
+  g->security_factor = dt_bauhaus_slider_new_from_params_box(self, "security_factor", " %%");
   dt_bauhaus_slider_enable_soft_boundaries(g->security_factor, -50, 200);
-  dt_bauhaus_widget_set_label(g->security_factor, NULL, _("dynamic range scaling"));
-  gtk_box_pack_start(GTK_BOX(page1), g->security_factor, FALSE, FALSE, 0);
-  dt_bauhaus_slider_set_format(g->security_factor, "%+.2f %%");
   gtk_widget_set_tooltip_text(g->security_factor, _("symmetrically enlarge or shrink the computed dynamic range.\n"
                                                     "useful to give a safety margin to extreme luminances."));
   g_signal_connect(G_OBJECT(g->security_factor), "value-changed", G_CALLBACK(security_threshold_callback), self);
@@ -1525,22 +1519,18 @@ void gui_init(dt_iop_module_t *self)
                                                 "ensure you understand its assumptions before using it."));
   gtk_box_pack_start(GTK_BOX(page1), g->auto_button, FALSE, FALSE, 0);
 
+  self->widget = page2;
 
   // contrast slider
-  g->contrast = dt_bauhaus_slider_new_with_range(self, 1., 2., 0.01, p->contrast, 3);
+  g->contrast = dt_bauhaus_slider_new_from_params_box(self, "contrast", "");
   dt_bauhaus_slider_enable_soft_boundaries(g->contrast, 0.0, 5.0);
-  dt_bauhaus_widget_set_label(g->contrast, NULL, _("contrast"));
-  gtk_box_pack_start(GTK_BOX(page2), g->contrast, FALSE, FALSE, 0);
   gtk_widget_set_tooltip_text(g->contrast, _("slope of the linear part of the curve\n"
                                              "affects mostly the mid-tones"));
   g_signal_connect(G_OBJECT(g->contrast), "value-changed", G_CALLBACK(contrast_callback), self);
 
   // latitude slider
-  g->latitude = dt_bauhaus_slider_new_with_range(self, 5.0, 45.0, 1.0, p->latitude, 2);
+  g->latitude = dt_bauhaus_slider_new_from_params_box(self, "latitude", " %%");
   dt_bauhaus_slider_enable_soft_boundaries(g->latitude, 0.01, 100.0);
-  dt_bauhaus_widget_set_label(g->latitude, NULL, _("latitude"));
-  dt_bauhaus_slider_set_format(g->latitude, "%.2f %%");
-  gtk_box_pack_start(GTK_BOX(page2), g->latitude, FALSE, FALSE, 0);
   gtk_widget_set_tooltip_text(g->latitude, _("width of the linear domain in the middle of the curve,\n"
                                              "in percent of the dynamic range (white exposure - black exposure).\n"
                                              "increase to get more contrast and less desaturation at extreme luminances,\n"
@@ -1549,10 +1539,7 @@ void gui_init(dt_iop_module_t *self)
   g_signal_connect(G_OBJECT(g->latitude), "value-changed", G_CALLBACK(latitude_callback), self);
 
   // balance slider
-  g->balance = dt_bauhaus_slider_new_with_range(self, -50., 50., 1.0, p->balance, 2);
-  dt_bauhaus_widget_set_label(g->balance, NULL, _("shadows/highlights balance"));
-  gtk_box_pack_start(GTK_BOX(page2), g->balance, FALSE, FALSE, 0);
-  dt_bauhaus_slider_set_format(g->balance, "%.2f %%");
+  g->balance = dt_bauhaus_slider_new_from_params_box(self, "balance", " %%");
   gtk_widget_set_tooltip_text(g->balance, _("slides the latitude along the slope\n"
                                             "to give more room to shadows or highlights.\n"
                                             "use it if you need to protect the details\n"
@@ -1560,11 +1547,8 @@ void gui_init(dt_iop_module_t *self)
   g_signal_connect(G_OBJECT(g->balance), "value-changed", G_CALLBACK(balance_callback), self);
 
   // saturation slider
-  g->saturation = dt_bauhaus_slider_new_with_range(self, -50., 50., 0.5, p->saturation, 2);
-  dt_bauhaus_widget_set_label(g->saturation, NULL, _("extreme luminance saturation"));
+  g->saturation = dt_bauhaus_slider_new_from_params_box(self, "saturation", " %%");
   dt_bauhaus_slider_enable_soft_boundaries(g->saturation, -50, 200.0);
-  dt_bauhaus_slider_set_format(g->saturation, "%.2f %%");
-  gtk_box_pack_start(GTK_BOX(page2), g->saturation, FALSE, FALSE, 0);
   gtk_widget_set_tooltip_text(g->saturation, _("desaturates the output of the module\n"
                                                "specifically at extreme luminances.\n"
                                                "increase if shadows and/or highlights are under-saturated."));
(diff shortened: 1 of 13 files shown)


AxelG-DE:
@dterrahe 

tbh currently I merely read your **1) Introduction** but I will go thru, as soon I have fixed other things.

From my issue #4250 (and if you would take time to dive into the links in there) you can see, at least from my end, there is hell of an interest.

johnny-bit:
I am very interested in this. I have some code lingering around with me playing with `temperature.c` and some changes.
I'd love to see mentioned by houz thingie about using introspection for xmp<->readable format. I know many would love it :)

I have couple of very small things to note based of on my experience with `temperature.c`:
1. there are sliders that aren't connected at all to module params (although I think storing rgb coeffs in params and not temp&tint might not be ideal ;))
2. there are sliders that can dependent on each other (dunno if any other module uses it). eg moving one slider moves two others and back. 

for other things my personal suggestions:
1. I wouldn't define "paging" in an introspection-related thingie. It's pure presentation stuff.
2. the hard/soft limits are a bit visual stuff too, but should be generally present, look at for example vignetting module - recently @jenshannoschwalm increased it's scale slider to 200% and all was fine, but that scale could easily be soft limit at 200 (or 250%) and have hard limit at 500 or even 1000% where effect is almost non-existent ;) but in normal sliding around for generic user 200 (or 250) would be perfectly fine.
3. some combo/sliders are (or can be) very dynamic. For example in my `temperature.c` work, one slider will allow smaller or larger range of values. and one combo there can have anything from 4 to 54 dynamically added presets :)
4. in some modules (again `temperature.c`) many if not most controlls are visual sugar, so how best to controll what goes where and how it's up for you to decide, but it's just FYI :)

overal - huge :+1: from me!

dterrahe:
@AxelG-DE 
> From my issue #4250 (and if you would take time to dive into the links in there) you can see, at least from my end, there is hell of an interest.

Yes, I know! :-) I hope that the rest of the devs are supportive of this effort.

dterrahe:
> I'd love to see mentioned by houz thingie about using introspection for xmp<->readable format. I know many would love it :)

Having the introspection information set up correctly is a necessary condition. But not sufficient for efficient data validation, since it might be out of sync with the actual limits enforced by the UI. That's why this is an attempt to have both use the same authoritative source (the introspection tags) so that any discrepancies become obviously visible in the UI. And also to have the largest possible benefit from the time invested.

> I have couple of very small things to note based of on my experience with `temperature.c`:

Yes, there are many more complicated situations and those cannot be easily reduced. Rather than restrict what people can do with the UI for their modules, what I'm trying to achieve is that if you only need something simple, this can help make it even simpler. If you want to have bells and whistles, then you'll be responsible for keeping it all working as well.
Maybe there's even a need for a "validate" function independent of the UI.

> 1. I wouldn't define "paging" in an introspection-related thingie. 

Maybe. But it would allow more powerful "no-ui-programming-needed" module building. Thinking further, people could build simple UIs outside of dt (not saying there would be demand for that, but going by the "if you build it, they will come" principle :-) )

> 2. the hard/soft limits are a bit visual stuff too, but should be generally present

Soft limits are used in _many_ modules (look at the spreadsheet I included). I find them very confusing and to be honest I'm not convinced they are implemented as intended. For input validation, the _soft_ limits are more important, because those are actually enforced in the sliders (even with manual input). This is not necessarily what the documentation leads one to expect...

> 3. some combo/sliders are (or can be) very dynamic. For example in my `temperature.c` work, one slider will allow smaller or larger range of values. and one combo there can have anything from 4 to 54 dynamically added presets :)

Sounds like another case for a "validate" function. Introspection is static, so could not possibly deal with dynamic dropdowns.

> 4. in some modules (again `temperature.c`) many if not most controlls are visual sugar, so how best to controll what goes where and how it's up for you to decide, but it's just FYI :)

Yes, this will not take away any options; just make it easier to do easy things without messing up.

ralfbrown:
Another vote for readable XMPs.  The current sidecar files with their binary blobs comply with the letter of the XML spec, but not the spirit.

I know there are folks who want to be able to read and modify settings in the XMPs for things like timelapse series with varying exposures.  Back when I still used Bibble, my export script looked at the sidecar and computed the effective focal length taking into account the crop applied to that image.

There's another benefit from having textual key/value pairs in the sidecar instead of a binary dump of the internal structure -- adding another parameter doesn't require an updated, backwards-incompatible version if the iop's behavior with the default value of the new parameter is the same as before.

houz:
First of all, I love that you are looking into this!

Now, a few comments. I already wrote this on IRC, but for future reference I will repeat it here, too.

- Your code shouldn't go into `bauhaus.{c,h}`. Instead create a new file.
- dt doesn't use checkboxen, instead it uses no/yes comboboxen.
- I agree that the soft/hard min/max stuff has some serious issues. But those are orthogonal to what you are trying to solve here, so let's ignore them in this PR.
- dt's introspection is just describing the binary params blob. Anything related to its visual representation or how to build a GUI is out of its scope. So there shouldn't be fields for tooltips, soft min/max, pseudo section or page entries or anything like that.
- Similarly, generating introspection data for old params versions is not something I would like. For two reasons:
    - out of scope
    - version bumps can do much more than just adding a new field with some default value
- I don't like generating enums for field names. The intended users of introspection can't make use of them, so the only place where it might be used is the iop itself. But that already knows how its params are layed out, so why would it use introspection in the first place? Maybe when building the GUI, but that is a one time thing where a few string comparisons don't matter.
- Introspection is already a hierarchical structure, so in 3.6.1 you don't need to pass pointers into the flat list but can request widgets for complete sub-structs.

Please don't understand these negative comments as a sign that I don't like your idea. I just think you either need to add a second data structure that amends the introspection with GUI metadata (and find a way to keep the two in sync), or you need to limit your scope to the parts that can be done with the already existing introspection data and have the user write some code when he wants to use extra features, like soft min/max.

Finally, I have been asked about human readable params many times, I have promised them many times and I have code for that (currently to go back and forth between params and JSON, but XML would be easy, too). I will definitely finish that and put it into dt, but I need more time. :cry: Just like broken bauhaus sliders, this is not part of this PR though, so let's ignore that topic.

dterrahe:
> * Your code shouldn't go into `bauhaus.{c,h}`. Instead create a new file.

/src/develop/imageop_gui.c/h?

> * dt doesn't use checkboxen, instead it uses no/yes comboboxen.

Indeed very few, but hotpixels.c and denoiseprofile.c have some. Is there a reason why they shouldn't be used?

> * I agree that the soft/hard min/max stuff has some serious issues. But those are orthogonal to what you are trying to solve here, so let's ignore them in this PR.

I'll put this in a separate PR. Looks like it could use a few small fixes and a better named function to set soft limits and then I'll be able to more cleanly transition into the envisioned end state for this PR. I don't think all modules need two passes; could be done in one go; I'll describe in other PR once done.

> * dt's introspection is just describing the binary params blob. Anything related to its visual representation or how to build a GUI is out of its scope. So there shouldn't be fields for tooltips, soft min/max, pseudo section or page entries or anything like that.

I'm absolutely fine with that. Having the intended scope clarified like this is valuable! (but hard to document in such a way that the next novice who wants to understand the structure can easily find it...)

>   * version bumps can do much more than just adding a new field with some default value

Fine with me too. What I was trying to find an approach for is the many modules that have very very simple structures, where the gui is just a direct translation to allow to input the required fields and where, with increased functionality, fields only get added and it is fine if the slider/dropdown goes to the bottom. In just those cases, we could do away with all the boilerplate and loop over introspection for initialisation, gui, gui-callbacks, updates and legacy params.

Any more complex module would just override all of this and do it themselves.

Even then, I would like to revisit legacy params, but that is the next project...

> * I don't like generating enums for field names. The intended users of introspection can't make use of them, so the only place where it might be used is the iop itself. But that already knows how its params are layed out, so why would it use introspection in the first place? Maybe when building the GUI, but that is a one time thing where a few string comparisons don't matter.

I now call my dt_bauhaus_slider_new_from_params_box with a field name string. This means only runtime type checking; not ideal, but ok. That string field name gets passed to generic_slider_callback at each slider move where get_p needs to look it up. Still only UI so speed not of essence, but maybe I can do some pointer arithmetic while connecting the callback.

> * Introspection is already a hierarchical structure, so in 3.6.1 you don't need to pass pointers into the flat list but can request widgets for complete sub-structs.

Not sure I've understood the hierarchical structure. Can you point to a module where that is used in an obvious way?

Also, introspection does not seem to pick up $DESCRIPTION comments for enum values yet. If I had those, I could automatically set up dropdowns. Introspection has the string, but it is just the enum name at the moment. Would this be a trivial change for you, or shall I try my hand at the perl bit?

> Please don't understand these negative comments as a sign that I don't like your idea. 

Quite the contrary! I'm grateful for your time and support.

> I just think you either need to add a second data structure that amends the introspection with GUI metadata (and find a way to keep the two in sync), or you need to limit your scope to the parts that can be done with the already existing introspection data and have the user write some code when he wants to use extra features, like soft min/max.

I don't have any ideas on how to achieve that yet. I'll start with the more modest targets first and then see if it is worth it to push standardization a little further.

> Finally, I have been asked about human readable params many times, I have promised them many times and I have code for that (currently to go back and forth between params and JSON, but XML would be easy, too). I will definitely finish that and put it into dt, but I need more time. 😢 Just like broken bauhaus sliders, this is not part of this PR though, so let's ignore that topic.

Hopefully by that time I'll have provided you with more populated introspection data, so you'll be able to do useful input validation. I was thinking of adding a paragraph 4.3. Validate() input validation function (for cases where valid field ranges are not independent)

houz:
>> dt doesn't use checkboxen, instead it uses no/yes comboboxen.

> Indeed very few, but hotpixels.c and denoiseprofile.c have some. Is there a reason why they shouldn't be used?

They are breaking the look and feel of dt where everything is a line of text (comboboxen) or a line of text with a line underneath (sliders). You also can't quickly change a checkbox with the mouse wheel.

The places that use checkboxen should be converted to dropdowns.

> it is fine if the slider/dropdown goes to the bottom.

Legacy params HAS to turn the old params into something that gives identical results as the old ones. So sliders may ONLY go to the bottom if that makes the iop output the old result.

> I now call my dt_bauhaus_slider_new_from_params_box with a field name string.

Is that code inside the iop or is it in your new .c file?

> Can you point to a module where that is used in an obvious way?

For example liquify.c

dterrahe:
> > Indeed very few, but hotpixels.c and denoiseprofile.c have some. Is there a reason why they shouldn't be used?

> The places that use checkboxen should be converted to dropdowns.

I'll see if these are trivial to convert when I have a look at them. If not, I'll file a "bug" report with the owners.

> > it is fine if the slider/dropdown goes to the bottom.
> 
> Legacy params HAS to turn the old params into something that gives identical results as the old ones. So sliders may ONLY go to the bottom if that makes the iop output the old result.

What I meant was that IF it doesn't matter where new sliders appear in the gui, and IF the default (as per introspection) is what achieves the same result as the previous simpler version (this is what a lot of legacy_param function set) then some of this could use default fallback implementations.

But for now I'm not looking at (legacy) versions. Once I'm done with the gui bit, I'll see if I can put another discussion PR together to see if there's an appetite to make that whole area a bit more robust and uniform. Again, there are some largest common denominators around, that could be simplified a lot, but the more complicated situations will just do what they are doing now.

> > I now call my dt_bauhaus_slider_new_from_params_box with a field name string.

> Is that code inside the iop or is it in your new .c file?

In the iop. For example in filmicrgb.c > gui_init:
g->grey_point_source = dt_bauhaus_slider_new_from_params_box(self, "grey_point_source", " %%");

> > Can you point to a module where that is used in an obvious way?
> 
> For example liquify.c

OK, liquify is one of those complex modules that I'm not even going to look at :-) Scary.

Also, it doesn't have any sliders or comboboxes it seems.

dterrahe:
@houz 
I added a default implementation for the module::init function that tries to pick the $DEFAULTS from introspection.

Two questions:
- I only now (belatedly) realised that there is no Default in dt_introspection_type_enum_t. Maybe you have already added it along with the descriptions in dt_introspection_type_enum_tuple_t. This should probably just be a numerical value rather than an enum identifier (or whatever those are called).
- Many modules initialise default_params twice; once in init() and once in reload_defaults(). Do I understand correctly that the second is only needed if the defaults are image dependent (for example resolution)? In that case, they could just only overwrite those values instead of completely reinitialising default_params. And is either one of those functions responsible for copying default_params into params? They often both do, but in those cases where init() is there but reload_defaults() isn't, I assume params gets reset somewhere else? I'd like to clean up this situation and make it consistent everywhere. Too much duplication hides what is required/used... Agreed?

houz:
> In the iop. For example in filmicrgb.c > gui_init:
g->grey_point_source = dt_bauhaus_slider_new_from_params_box(self, "grey_point_source", " %%");

I see. But why would the iop use the introspection wrapper in the first place? It could just use the current bauhaus API directly.

> Maybe you have already added it along with the descriptions in dt_introspection_type_enum_tuple_t.

Yes, I did.

dterrahe:
> I see. But why would the iop use the introspection wrapper in the first place? It could just use the current bauhaus API directly.

I'm not sure I understand your question. What do you mean by introspection wrapper? This little project is to refactor some of the boilerplate in setting up the gui which will also allow to consistently enforce constraints specified in the introspection comments. To do this, I need to be able to pass a reference to a field definition, not just the value of a field. At the moment the introspection framework only refers to fields via strings, so no compile time verification.

Maybe you could show how you would do it?

> Yes, I did.

Thanks!

dterrahe:
@TurboGit @AlicVB @aurelienpierre anybody else that I should include; at this point I would be grateful for any feedback on the kind of changes I'm proposing here. Have a look at the aggregate changes to basicadj.c and filmicrgb.c for an impression of the amount of boilerplate code reduction (hopefully resulting in more readable and maintainable code).

src/iop/filmicrgb.c
```diff
@@ -1525,110 +1265,132 @@ void gui_init(dt_iop_module_t *self)
...
+    }
+
+    p->output_power =  logf(p->grey_point_target / 100.0f) / logf(-p->black_point_source / (p->white_point_source - p->black_point_source));
+
+    dt_bauhaus_slider_set_soft(g->output_power, p->output_power);
+  
+    darktable.gui->reset = reset;
+  }
+
+  gtk_widget_queue_draw(self->widget); // Is this needed?
```
aurelienpierre, line 1392: this is needed only if the drawing area needs a redraw, aka if the curve params have changed (white, grey, black, latitude, contrast, etc.)

aurelienpierre:
That's surely nice. The GUI/GTK vs. pixel loops code ratio was really crazy, this PR is more than welcome (but will conflict with #4800 :-/). I need to think a bit more about what would be needed.

src/iop/filmicrgb.c
```diff
@@ -128,19 +129,19 @@ typedef struct dt_iop_filmic_rgb_spline_t
...
+  float security_factor;    // $MIN: -50 $MAX: 200 $DEFAULT: 0 $DESCRIPTION: "dynamic range scaling"
+  float grey_point_target;  // $MIN: .1 $MAX: 50 $DEFAULT: 18.45 $DESCRIPTION: "target middle grey"
+  float black_point_target; // $MIN: 0 $MAX: 100 $DEFAULT: 0 $DESCRIPTION: "target black luminance"
+  float white_point_target; // $MIN: 0 $MAX: 100 $DEFAULT: 100 $DESCRIPTION: "target white luminance"
+  float output_power;       // $MIN: 1 $MAX: 10 $DEFAULT: 5.98 $DESCRIPTION: "target power transfer function"
+  float latitude;           // $MIN: 0.01 $MAX: 100 $DEFAULT: 45
+  float contrast;           // $MIN: 0 $MAX: 5 $DEFAULT: 1.30
+  float saturation;         // $MIN: -50 $MAX: 200 $DEFAULT: 5 $DESCRIPTION: "extreme luminance saturation"
+  float balance;            // $MIN: -50 $MAX: 50 $DEFAULT: 12 $DESCRIPTION: "shadows/highlights balance"
+  int preserve_color;       // $DESCRIPTION: "preserve chrominance" $DEFAULT: 3
```
aurelienpierre, line 144: it would be nice if we could use the `enum` type mnemonic instead of the number in there.

dterrahe:
> but will conflict with #4800 :-/

Yes, in hindsight FilmicRGB was an unfortunate random pick to use as illustration. But that doesn't matter; the point is that I will go over _all_ the other modules as well, so a few changes here don't phase me :-)

The first infrastructure commit won't include any module changes. I will submit the modules one by one, after the infrastructure changes in imageop(_gui) have been merged.

Btw the commits in this discussion PR are a mess and also include the slider_set_soft_range addition that was already merge. I'm having trouble rebasing (probably due to the addition of another submodule) so I won't clean up here but only when I submit final PR.

src/iop/filmicrgb.c
dterrahe: Well, in this particular case, preserve_color isn't even an enum. If it were, then the combobox will automatically be filled with the enum values (houz is working on making human readable text from $DESCRIPTION comments available). Not sure how easy using mnemonics in the $DEFAULT comment would be in that case; this is perl code, so I'm doing houz the favor of not touching it :-)

TurboGit:
@dterrahe : Really nice work ! I just did a first pass on the code, like indeed the fact that it reduces the code a lot and less code to write is obviously less bug written too!

There is one point that I don't see or understand in the code. In many IOP moving a slider will change some other sliders. How is that done with this generic code ?

Sorry if the question has already been raised, I did not read all comments yet.

dterrahe:
That can now be done via the new gui_changed method. For an example look at the new code towards the bottom of filmicrgb.c

src/iop/filmicrgb.c
houz: Thank you. ;-)
I just pushed fb46a038203f69d4509c360d32cc0de73773a1c4 which adds (among many other improvements) descriptions to the enum field. You can also just use enum mnemonics in `$DEFAULT` as the value is just copied over verbatim to the C/C++ file. You can have a look at the generated files `build/src/iop/introspection_*` while playing with those magic comments in the iop, maybe that makes it more clear.

PR #4806, opened 2020-04-24
Fix hypersensitive textsize.

### LINKED ISSUE #4769: CSS `font-size` parameter hyper-sensitive upgrading from 3.0.0 to 3.0.1/3.0.2
**Describe the bug**

When trying to adjust the basic font size of the interface, editing `/usr/share/darktable/themes/darktable.css`, changing the value from the default 1em to 0.9em or 1.1em creates very big changes in font size (see the screenshots). It worked perfectly in 3.0.0.  (**corrected, it was 3.0.1 in the first version**)

**To Reproduce**

1. Edit  /usr/share/darktable/themes/darktable.css and around line 160 change the value of `font-size:` from `1em` to something like `0.9em` or `1.1em`.
2. Launch `darktable`

**Expected behavior**

A change of 10% of that parameter should be reflected as a change of (around) 10% of the font size in the main interface. 
It worked that way on 3.0.0

**Screenshots**

Here I made two screenshots with two values of `font-size:`

 - Darktable Version: 3.0.2, installed by PPA https://launchpad.net/~ubuntuhandbook1/+archive/ubuntu/darktable
 - OS: Ubuntu 18.04
 - No OpenCL
 - Internal Intel graphic card

**Additional context**

I can use a workaround changing the DPI of the monitor (although it shouldn't be needed; in 3.0.1 the automatic detection worked fine.)

BTW, thank you again for this great SW

Rmano:
That was removed in 18404309fa0356b4245a8d843bff6babefa2cf34
Fixes #4769

### DIFF
--- a/data/themes/darktable.css
+++ b/data/themes/darktable.css
@@ -223,6 +223,7 @@ eventbox *,
 box,
 box *
 {
+  font-size: 1em;
   background-color: transparent;
 }
 


Nilvus:
Good catch.

That was not "removed". I removed it as it's a duplicate to the * tag as the font-size is the same and default font size is good for me and all other feedback I had. I've just forgot (because never do that and need do that) that if changing * tag font-size, all UI parts are impacted without this one... 

So this line is only useful if someone like you want to change font-size for darktable UI (just as a reminder, 1em seems that the font size is the OS default font size, like for most of the apps you could use).

Said that, for the few users who want to change is indeed better to put back this line.

@TurboGit: you could merge that and close #4769 issue. I will add comments later (with my preferences CSS work) as a reminder for developers like me (and users who want to tweak it).

TurboGit:
Ok, merging now then. Thanks.

TurboGit:
I didn't notice while merging that this was against 3.0.x branch. Please never create PR for any other branches than **master**. Thanks.

Rmano:
> I didn't notice while merging that this was against 3.0.x branch. Please never create PR for any other branches than **master**. Thanks.

Sorry. I didn't know it. Will do in the future, tell me if I can do anything now (should be easy to rebase).

TurboGit:
No, nothing to do. Hopefully I would have merged manually into darktable-3.0.x so no harm. But if this would not have been selected for the branch I would have to revert on the branch. That would have been embarrassing.

Nilvus:
@TurboGit: not sure here but I think it's possible we will have a 3.0.3. So could you just confirm that this PR will be in next minor release (3.0.3)? I only see a 3.2 milestone here. I've just help Luc on darktable.fr forum (he help another user to change font) and the user had the problem (I propose him this fix) on 3.0.1/3.0.2 he use. So needed to be on next minor release.

TurboGit:
@Nilvus : Sure I have merged this into darktable-3.0.x branch on Apr 24. Will be part of 3.0.3 if there is such version.

PR #4807, opened 2020-04-24
Updates to pt_BR manual

jpellegrini:
Several more parts of chapters 3 and 10 translated; chapter 5 now completely translated.

Also, a couple of fuzzies in the darktable source were also translated (updates from these days).


TurboGit:
Thanks !

PR #4809, opened 2020-04-25
Locallaplacian speedup 2

### LINKED ISSUE #4808: locallaplacians.c - scalar and SSE2 code produce different results
**Describe the bug**

gauss_reduce() and gauss_reduce_sse2() produce different output, so results will vary depending on how darktable was compiled.
The cause is use of different kernels in the two code paths.  gauss_reduce() uses 1 5 8 5 1, while gauss_reduce_sse2() uses 1 4 6 4 1.

 - Darktable Version: git master (as of commit 75abcb81c41632f5ab0173b60baa8369116fde66)

ralfbrown:
By rearranging the order of computation, I was able to make the gauss_reduce_sse2 function in src/common/locallaplacian.c parallelizable.  Each thread can now be given 1/t of the entire 2d array instead of 1/t of a single row.  Being able to use vector instructions yields a 20+% speedup compared to the scalar version used in the current Git master.

|		| Git master		| New code		|
| ---		| ---			| ---			|
| No OpenMP	|3.695 secs (3.829 CPU)	|2.447 secs (2.674 CPU)	|
|2 threads	|1.940 secs (3.421 CPU)	|1.477 secs (2.476 CPU)	|
|8 threads	|0.627 secs (4.034 CPU)	|0.451 secs (2.691 CPU)	|
|32 threads	|0.302 secs (6.694 CPU)	|0.262 secs (4.072 CPU)	|
|64 threads	|0.373 secs (7.803 CPU)	|0.280 secs (5.905 CPU)	|

In the process, I found that the scalar and SSE versions were using different kernels.  The third commit in this PR adapts the scalar version to use the SSE version's kernel, since I figure the vast majority of users will have been running that code.

Scalar and SSE2 versions now produce the same output modulo floating-point roundoff causing some of the pixel values to differ by 1 (of a range of 65536) in 16-bit output.

Fixes #4808.

### DIFF
--- a/src/common/locallaplacian.c
+++ b/src/common/locallaplacian.c
@@ -212,7 +212,7 @@ static inline void gauss_reduce_sse2(
     // prime the vertical axis
     const __m128 kernel = _mm_setr_ps(1.f, 4.f, 6.f, 4.f);
     __m128 left = convolve14641_vert(base,wd);
-    for(int col=0; col<cw-1; col+=2)
+    for(int col=0; col<cw-3; col+=2)
     {
       // convolve the next four pixel wide vertical slice
       base += 4;
@@ -231,7 +231,7 @@ static inline void gauss_reduce_sse2(
       base += 4;
       float right = base[0] + 4*(base[wd]+base[3*wd]) + 6*base[2*wd] + base[4*wd];
       __m128 conv = _mm_mul_ps(left,kernel);
-      out[cw-2] = (conv[0] + conv[1] + conv[2] + conv[3] + right) / 256.f;
+      out[cw-3] = (conv[0] + conv[1] + conv[2] + conv[3] + right) / 256.f;
     }
   }
   ll_fill_boundary1(coarse, cw, ch);
--- a/src/common/locallaplacian.c
+++ b/src/common/locallaplacian.c
@@ -29,6 +29,11 @@
 #include <xmmintrin.h>
 #endif
 
+// the maximum number of levels for the gaussian pyramid
+#define max_levels 30
+// the number of segments for the piecewise linear interpolation
+#define num_gamma 6
+
 //#define DEBUG_DUMP
 
 // downsample width/height to given level
@@ -241,8 +246,7 @@ static inline void gauss_reduce(
   const int cw = (wd-1)/2+1, ch = (ht-1)/2+1;
 
   // this is the scalar (non-simd) code:
-  const float w[5] = { 1.f, 4.f, 6.f, 4.f, 1.f };
-  const float scale = 1.f/256.f;
+  const float w[5] = { 1.f/16.f, 4.f/16.f, 6.f/16.f, 4.f/16.f, 1.f/16.f };
   memset(coarse, 0, sizeof(float)*cw*ch);
   // direct 5x5 stencil only on required pixels:
 #ifdef _OPENMP
@@ -257,7 +261,6 @@ static inline void gauss_reduce(
       for(int jj=-2;jj<=2;jj++)
         for(int ii=-2;ii<=2;ii++)
           coarse[j*cw+i] += input[(2*j+jj)*wd+2*i+ii] * w[ii+2] * w[jj+2];
-      coarse[j*cw+i] *= scale;
     }
   ll_fill_boundary1(coarse, cw, ch);
 }
@@ -565,8 +568,6 @@ void local_laplacian_internal(
     const int use_sse2,         // flag whether to use SSE version
     local_laplacian_boundary_t *b)
 {
-#define max_levels 30
-#define num_gamma 6
   // don't divide by 2 more often than we can:
   const int num_levels = MIN(max_levels, 31-__builtin_clz(MIN(wd,ht)));
   int last_level = num_levels-1;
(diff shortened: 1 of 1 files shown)


johnny-bit:
In #4154 @burrima added unit testing capacities. Cases like this are perfect use-case for those. Can you try and also create unit tests for locallaplacian code?

ralfbrown:
I'll take a look  (from scanning that PR, it seems I'll have to learn both Cmake and cmocka to do so).  I'm also not sure how one would make unit tests for something whose results depend on the entire input image....

For cases like this, it may be better to have a system test.  Several representative images that a test script runs through darktable-cli and compares against the "gold" result.  For the comparison that I referred to above, what I did was export to full-resolution 16-bit TIFFs using both the pre- and post-change versions of dt (did it manually though), and then used ImageMagick to compare the resulting images.

~~~
  convert before.tif after.tif -metric pae -compare -format "%[distortion]\n" info:
~~~
should give a result of either 0 (all pixels the same) or 1.52959e-05 (1/65536, differences due to roundoff).  If there were differences, you can count the number of differing pixels with
~~~
  convert before.tif after.tif -metric ae -compare -format "%[distortion]\n" info:
~~~

Speaking of testing, can someone tell me why Travis CI fails the PR?  All I get when I try to see the details is a blank page.

TurboGit:
@ralfbrown : I have also made an integration testsuite see PR #4818, it should be easy to add a test.

TurboGit:
Oh, I have already a test for local contrast in fact, let me test this PR against it...

TurboGit:
Result is not good:

```
$ ./run.sh 0002-local-contrast
Test 0002-local-contrast
      Image mire1.cr2
      CPU & GPU version differs
      Max  dE         11
      Mean dE         0
      Count above max 88474
  FAILS: image visually changed

Total test 1
Errors     1
```
Here is the resulting diff image:

diff

src/common/locallaplacian.c
```diff
@@ -135,83 +157,40 @@ static inline void gauss_reduce_sse2(
...
-  {
-    // horizontal pass, convolve with 1 4 6 4 1 kernel and decimate
-    for(;rowj<=2*j+2;rowj++)
-    {
-      float *const row = ringbuf + (rowj % 5)*stride;
-      const float *const in = input + rowj*wd;
 #ifdef _OPENMP
 #pragma omp parallel for default(none) \
-      dt_omp_firstprivate(cw, in, row) \
+      dt_omp_firstprivate(cw, ch, input)   \
```
TurboGit, line 162: to compile with recent GCC you need to add `wd` and `coarse` in the firstprivate list.

ralfbrown:
Which versions of the code did you compare for that difference image?  Keep in mind that my first-round speedup changed from gauss_reduce_sse2 to gauss_reduce, which as I later discovered is using a different kernel.  This PR changes things to use the same kernel that was in use before my first PR.

Since gauss_expand has factors of 4 and 6 in the computation, I think the 1 4 6 4 1 kernel is in fact the correct one, not the 1 5 8 5 1 that gauss_reduce() had been using.

TurboGit:
> Which versions of the code did you compare for that difference image?

The version from some (~20) minutes ago.

ralfbrown:
What version for the reference image?  The correct comparison is against the code prior to my first PR for local contrast (#4789), because that PR inadvertently changed the results due to the inconsistency between the then-existing scalar and SSE code.

TurboGit:
BTW, there is certainly an issue of non initialized value as I get different results on each run of the testsuite.

TurboGit:
> What version for the reference image?

The reference image has been created just before merging this issue with up-to-date code from master.

ralfbrown:
But up-to-date code from master includes my previous PR, which breaks things.... (by using the 1 5 8 5 1 kernel in the scalar code, instead of the 1 4 6 4 1 kernel used by both SSE and CL code)  This PR fixes that by making every code path use the 1 4 6 4 1 kernel, but that in itself changes the output if you only compare with and without this single PR.

I'll look for uninitialized values.

ralfbrown:
Found it -- off-by-one error.  Multiple consecutive exports give identical results now.

TurboGit:
Ok, I had indeed created the reference from your previous commit. I just recreated a proper reference image, and now my test gives on current master:

```
Test 0002-local-contrast
      Image mire1.cr2
      CPU & GPU version differs
      Max  dE         11
      Mean dE         0
      Count above max 88469
  FAILS: image visually changed
```

With this PR it is a bit better but not perfect:

```
Test 0002-local-contrast
      Image mire1.cr2
      CPU & GPU version differs
      Max  dE         3
      Mean dE         0
      Count above max 69
  FAILS: image visually changed


Total test 1
Errors     1
```

Here is the diff and the good news is that this time the diff is stable so no more random pixels.

diff

Can you please see if the remaining errors can be fixed?

Thanks.

ralfbrown:
Unable to reproduce using TIFF output.

Your use of JPEG output is creating compression artifacts that amplify differences to the point that the minimum possible difference results in an artifact exceeding the dE threshold.  I recognized that test image, so I went and downloaded the .CR2 for Canon 5D Mark III from imaging-resource (hard to find a CR2, since they don't link superceded cameras from their index).

Using
~~~
darktable-cli --hq true --apply-custom-presets false E5D3hSLI000800NR0.CR2{,.xmp} output.tif
~~~
ImageMagick with
~~~
convert expected.tif output.tif -metric pae -compare info:
~~~
reports a maximal pixel difference of 1.5259e-05, which is 1/65536, the minimal representable difference in a 16 bit per channel image.  Change all of the .tif above to .jpg, and the PAE jumps to 0.0235294.

TurboGit:
I fully agree that jpeg is not a good choice. I've used that because I found some sample Python code to read image to be passed to OpenCV to compute Delta-E. I'm sure we can use PNG or TIF.

TurboGit:
All good now ! I have migrated the testsuite to use PNG and the diff compared to the reference version is Delta-E 1. Will merge now.

PR #4818, opened 2020-04-25
po/integration-tests

TurboGit:
First attempt to create an integration testsuite. The goal is for it to be very simple to update and will allow us to get more confidence about some deep internal changes regarding pixel transformation.

Currently two tests only just to get started:

```
$ ./run.sh 
Test 0001-exposure
      Image mire1.cr2
      Max  dE         0
      Mean dE         0
      Count above max 0
  OK

Test 0002-local-contrast
      Image mire1.cr2
      Max  dE         0
      Mean dE         0
      Count above max 0
  OK


Total test 2
Errors     0
```


TurboGit:
Note that the max Delta-E allowed is 2. This is a constant into the python script and the "Count above max" is the number of pixels above this max.

parafin:
Why not put it into existing src/tests?

TurboGit:
I have added the computation of the difference image in case of error.

TurboGit:
> Why not put it into existing src/tests?

I didn't want to mess with unit tests which are different beast, but this can be moved here it making more sense. And yes, you are probably right.

TurboGit:
As a next test I will see if I can validate the change in #4775.

TurboGit:
At least we may need to have two directories one for unit test and one for my integration tests otherwise we won't be able to easily see what is used by one or another.

parafin:
new unittests are already in a separate directory, so we can add e.g. src/tests/integration

TurboGit:
In this new version I have moved the tests.

Also, I do check that the CPU & GPU version are the same or not and display a message + compute the diff image. Could help later, at the moment I do not except to have the identical image anyway.

I have also experienced a diff when the image is exported from cli or from darktable. This looks like some tiling issue but I have seen error around 10 or 12 dE !!! At this stage again this PR is not about tracking those issues, it may come later. We I want to have in place is a reference to check that new code do not break existing output. So the expected.jpg must now be generated with the cli. README.txt has been update accordingly.

johnny-bit:
You create `expected.jpg`, so I lossy format... wouldn't lossless format be desired in integrations test? On one hand that would be tons of additional data but on the other I would be better than relying on lossy nature of jpegs being consistent, no?

TurboGit:
Fully agreed. I'll use another format.

TurboGit:
New version pushed which is using PNG. I expect now to create a test every time an IOP is modified and not yet covered by the testsuite.

TurboGit:
Current output is:

```
$ ./run.sh 
Test 0001-exposure
      Image mire1.cr2
      CPU & GPU version differs
      Max  dE         0.0000
  OK

Test 0002-local-contrast
      Image mire1.cr2
      CPU & GPU version differs
      Max  dE         1.0077
  OK

Test 0003-denoise-bilateral
      Image mire1.cr2
      CPU & GPU version differs
      Max  dE         0.4928
  OK

Test 0004-masks
      Image mire1.cr2
      CPU & GPU version differs
      Max  dE         0.0000
  OK

Test 0005-orientation-minus-90
      Image mire1.cr2
      CPU & GPU version differs
      Max  dE         0.0000
  OK

Test 0006-orientation-plus-90
      Image mire1.cr2
      CPU & GPU version differs
      Max  dE         0.0000
  OK

Test 0007-orientation-180
      Image mire1.cr2
      CPU & GPU version differs
      Max  dE         0.0000
  OK

Test 0008-flip-h
      Image mire1.cr2
      CPU & GPU version differs
      Max  dE         0.0000
  OK

Test 0009-flip-v
      Image mire1.cr2
      CPU & GPU version differs
      Max  dE         0.0000
  OK

Test 0010-flip-hv
      Image mire1.cr2
      CPU & GPU version differs
      Max  dE         0.0000
  OK


Total test 10
Errors     0
```

TurboGit:
If there is no more comments, objections I'll merge this during the week-end. We will then be able to add more tests to ensure that touching an IOP does not break current edit beyond a limit of dE > 2.

johnny-bit:
I mentioned this (indirectly) in #5054 - this could be good idea also for showing why something is/isn't blacklisted re-opencl drivers blacklisting.

PR #4821, opened 2020-04-26
preferences: add save button

elstoc:
adds a save button to the action bar at the top of the preferences
dialog. Now, clicking 'X' or pressing ESC will exit without saving changes.
Clicking on 'save' will exit saving changes.

### DIFF
--- a/src/gui/preferences.c
+++ b/src/gui/preferences.c
@@ -394,7 +394,7 @@ void dt_gui_preferences_show()
   GtkWindow *win = GTK_WINDOW(dt_ui_main_window(darktable.gui->ui));
   _preferences_dialog = gtk_dialog_new_with_buttons(_("darktable preferences"), win,
                                                     GTK_DIALOG_DESTROY_WITH_PARENT | GTK_DIALOG_MODAL | GTK_DIALOG_USE_HEADER_BAR,
-                                                    NULL, NULL);
+                                                    _("save"), GTK_RESPONSE_ACCEPT, NULL);
   gtk_window_set_default_size(GTK_WINDOW(_preferences_dialog), DT_PIXEL_APPLY_DPI(1100), DT_PIXEL_APPLY_DPI(700));
 #ifdef GDK_WINDOWING_QUARTZ
   dt_osx_disallow_fullscreen(_preferences_dialog);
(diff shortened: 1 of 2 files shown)


TurboGit:
But I'm not sure ESC does not save. Have you tested? I remember that this was a request. The settings are changed each time a change is made in the preference.

elstoc:
You're right @TurboGit. I did test but I was testing on the new longstring settings which don't work the same as normal gtk_entry boxes. If I test on the others it works as you described (X and ESC do the same as Save).

elstoc:
On the plus side, looking at the latest commit of the CSS from @Nilvus it appears we can safely remove the hard-coded header/action bar without changing the appearance. So perhaps I'll close this PR and open a new one just to do that. Leave 'save without changing' for another day.

Nilvus:
> On the plus side, looking at the latest commit of the CSS from @Nilvus it appears we can safely remove the hard-coded header/action bar without changing the appearance. So perhaps I'll close this PR and open a new one just to do that. Leave 'save without changing' for another day.

Euh, I've just posted my PR included your save button margin/padding...

elstoc:
Don't worry @Nilvus - leave it in. I've worked out how to achieve the original intent of this PR.

elstoc:
Latest commit should now allow user to save changes by hitting the 'save' button or discard changes by hitting 'X' or 'ESC'.

elstoc:
N.B. this will only affect the auto-built preferences tabs. As discussed on the previous PR, the general, shortcuts and presets tabs have their own separate behaviour.

Nilvus:
For me, it's good that way.

TurboGit:
Changing the language is always saved on my side after ESC or clicking the X. Is that expected?

elstoc:
Yes, as I said above, general, shortcuts and preferences all work differently so I would expect that.

elstoc:
^presets

TurboGit:
Sorry I missed that. Not good to have different behavior but we can probably improve that later. So I'll merge that now. Thanks.

elstoc:
Yes I do  agree, but having the theme changes apply straight away is useful and getting the same behaviour with shortcuts and presets might be quite hard, if it's even possible.  So perhaps we could do it with the language change but then it would be internally inconsistent on that tab.

Anyway thanks for merging - I'll go and look at adjusting #4424 now.

elstoc:
Hate to keep fiddling with this but I've been using it for a bit and without a close button my mouse just automatically goes to the 'X' without thinking and I can't get used to hitting save. Does anyone have any objection if I change it to this instead...

Screenshot_2020-04-26_21-25-02

PR #4822, opened 2020-04-26
Update CSS from new preferences rework by @elstoc

Nilvus:
CSS work from preferences rework by @elstoc. This integrate work on both PR #4747 and #4821 .
This also include minor cleaning and add comments.
I've also integrate .accels_window_list button in same one header for shortcuts and presets to keep consistent UI (so it improves a little bit header bar on accels window also.

Below a screenshot of some preferences categories:
general

lighttable

shortcuts

storage

### DIFF
--- a/data/themes/darktable.css
+++ b/data/themes/darktable.css
@@ -1328,6 +1328,11 @@ scale contents trough slider:hover
   margin: 0px; /* reset default dialog margin to all items. And needed to have an hover and selected effect on all width of categories in sidebar */
 }
 
+#preferences_notebook decoration  /* allow to resize dialog window */
+{
+  margin: 1px;
+}
+
 /* set headerbar buttons */
 #preferences_notebook headerbar button
 {
--- a/data/themes/darktable.css
+++ b/data/themes/darktable.css
@@ -158,7 +158,7 @@
   background-color: @bg_color;
   background-image: none;
   font-family: sans-serif;
-  font-size: 1em;
+  font-size: 1em;     /* you can adjust text font size here ; need to have font-size set to 1em; on line 226 to have the rest of the UI ok */
   color: @fg_color;
   text-shadow:none;
   box-shadow:none;
@@ -223,7 +223,7 @@ eventbox *,
 box,
 box *
 {
-  font-size: 1em;
+  font-size: 1em;    /* it's recommanded to avoid changing this setting to be sure the UI stay ok if changing font size on line 161. By default, duplicate line 161. */
   background-color: transparent;
 }
 
@@ -333,8 +333,8 @@ button
 #iop-plugin-ui button,
 #lib-plugin-ui button
 {
-  margin: 2pt;
-  padding: 2pt;
+  margin: 2px;
+  padding: 2px;
 }
 
 /* Buttons in toolbars in lighttable and darkroom */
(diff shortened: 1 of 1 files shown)


Nilvus:
@TurboGit: so needs to wait what will be validate about headerbar since last post on #4821... before merging that one.

elstoc:
@Nilvus looks nice to me. I've tested it against my latest PR.

elstoc:
Actually, just one comment - with your CSS it seems I can't resize the dialog any more, but I can with the old CSS.

Nilvus:
> Actually, just one comment - with your CSS it seems I can't resize the dialog any more, but I can with the old CSS.

I've seen that preferences dialog window can't be resized but think that was your PR and it don't bother me. Just test and indeed, resize (only horizontally, not possible vertically on your PR) it's possible.

Will check why. By now, I don't see why CSS could do that...

elstoc:
If I change the following to 1px I can resize again
```
#preferences_notebook *
{
  margin: 0px; /* reset default dialog margin to all items. And needed to have an hover and selected effect on all width of categories in sidebar */
}
```

elstoc:
I think the window uses one of the margins as the drag handle and setting it to zero removes the handle.

Nilvus:
> If I change the following to 1px I can resize again
> 
> ```
> #preferences_notebook *
> {
>   margin: 0px; /* reset default dialog margin to all items. And needed to have an hover and selected effect on all width of categories in sidebar */
> }
> ```

That was I was suspected by seeing that by default vertically was not possible. But changing that will break other things. I will change that the good way. Thanks.

Nilvus:
> I think the window uses one of the margins as the drag handle and setting it to zero removes the handle.

That's totally that.

elstoc:
Yes I wasn't suggesting that as the solution - it messes everything else up!

Nilvus:
> Yes I wasn't suggesting that as the solution - it messes everything else up!

Was easy, that's the decoration tag just after #preferences_notebook dialog that set resizing possible or not by setting to 0 or 1px margin. I push the fix commit in a minute.

elstoc:
Also, while the presets dialog looks fine initially as soon as you expand any of them the columns all change position and scroll bars appear. You do get the same behaviour in current master but imo it's worse now.  It doesn't bother me - I don't use this tab - just thought I'd mention it.

elstoc:
> Was easy, that's the decoration tag just after #preferences_notebook dialog that set resizing possible or not by setting to 0 or 1px margin.

Works for me

Nilvus:
> Also, while the presets dialog looks fine initially as soon as you expand any of them the columns all change position and scroll bars appear. You do get the same behaviour in current master but imo it's worse now. It doesn't bother me - I don't use this tab - just thought I'd mention it.

I know that but can't do anything. It's worse just because headerbar (margins/paddings) are better. We can't have both and scrollbar is there for that. It's only presets, I think most users don't use presets 99% of the time. Personally I find easier to manage my presets in darkroom than on preferences window.

elstoc:
I agree. As I say, not a problem for me, just wanted to make sure you were aware.

Nilvus:
> I agree. As I say, not a problem for me, just wanted to make sure you were aware.

I just see a strange thing on that presets columns. Last column auto is by default quite large and a bit larger than others. I've check all CSS settings are the same. The hover is also not affected same way on this column. Don't understand why by now. Could you just check if this column is hardcoded differently than others. I will check again if I don't miss something on CSS.

elstoc:
I agree. Good to merge.

TurboGit:
Very nice work to me! Thanks.

PR #4825, opened 2020-04-26
preferences: amend buttons on header bar

elstoc:
remove the 'X' (close) button from the preferences header bar and amend to
display two buttons: 'save changes' and 'exit without saving'

Screenshot_2020-04-26_21-25-02

### DIFF
--- a/src/gui/preferences.c
+++ b/src/gui/preferences.c
@@ -394,13 +394,17 @@ void dt_gui_preferences_show()
   GtkWindow *win = GTK_WINDOW(dt_ui_main_window(darktable.gui->ui));
   _preferences_dialog = gtk_dialog_new_with_buttons(_("darktable preferences"), win,
                                                     GTK_DIALOG_DESTROY_WITH_PARENT | GTK_DIALOG_MODAL | GTK_DIALOG_USE_HEADER_BAR,
-                                                    _("save"), GTK_RESPONSE_ACCEPT, NULL);
+                                                    _("save changes"), GTK_RESPONSE_ACCEPT, _("exit without saving"), GTK_RESPONSE_CLOSE,NULL);
   gtk_window_set_default_size(GTK_WINDOW(_preferences_dialog), DT_PIXEL_APPLY_DPI(1100), DT_PIXEL_APPLY_DPI(700));
 #ifdef GDK_WINDOWING_QUARTZ
   dt_osx_disallow_fullscreen(_preferences_dialog);
 #endif
   gtk_window_set_position(GTK_WINDOW(_preferences_dialog), GTK_WIN_POS_CENTER_ON_PARENT);
   gtk_widget_set_name(_preferences_dialog, "preferences_notebook");
+
+  //remove close button from header bar
+  GtkWidget *headerbar = gtk_dialog_get_header_bar(GTK_DIALOG(_preferences_dialog));
+  gtk_header_bar_set_show_close_button(GTK_HEADER_BAR(headerbar), FALSE);
 
   //grab the content area of the dialog
   GtkWidget *content = gtk_dialog_get_content_area(GTK_DIALOG(_preferences_dialog));


elstoc:
I would have done it this way the first time but I thought the 'X' was hard-coded. IMO the 'X' plus 'save' buttons looked and felt wrong in use and this feels more natural.

TurboGit:
Look good to me ! Thanks.

chhil:
@elstoc 
 A suggestion:
Save changes currently closes the dialog.
From a usability perspective, would be be possible to have a save, save and exit.
Save would just save the changes and let you make more changes. I generally go tab by tab to make changes and would prefer saving the changes upto the point and continue. Finally, I can save and exit.

elstoc:
I'm not sure that there's much need for that workflow (really, how often are you likely to save changes to one tab then exit without saving on another?) I suspect, after initial setup, most people will just go into preferences to change a single setting and exit, in which case two buttons are sufficient. Though I can see the argument for the 'save changes' button to say 'save and exit' wording instead.

chhil:
In the other IDE (eclipse) that I use I prefer the options. Not too fussed up about DT.

image

Nilvus:
I think this has been merged too quickly. I find the text too long on those buttons and the order is not usual. And now many preferences doesn't have any buttons (see Gnome preferences for example and all preferences are applied when changed only, no need to click on a button.

My point would be to have only close one and all preferences saved when changed.

Said that if we want buttons anyway, I agree with @chhil to have an 'apply and close' (or 'save and close' and 'cancel' (or 'close' only). That will be shorter and took less place.

elstoc:
> the better way would be to simply have all preferences changed when the user changed them

Not so easy with the textview widgets (since IMO they don't quite have a change event) but we can have all the settings apply when the window is closed with X or ESC (which is what the original PR #4747 did). I don't have an issue with this. I removed the bottom close button on request. I added the ability to cancel without changes on request. I was just trying to avoid pressing X cancelling changes - since it's the first thing my mouse automatically goes to when I want to close the window and I was cancelling changes accidentally too often.

If we can come to a consensus on what is required I'm happy to change it. My only caveat is that I don't want 'X' cancelling changes.  That's also the reason for the ordering, since with a window the user automatically mouses to the top-right and having this ('default') user action save seemed the most logical. If the buttons were on the bottom-right I would agree that OK/Cancel in that order seems reasonable and expected.

elstoc:
IMO the least-worst compromise is to remove the hard-coded action bar, moving the (empty) action area back to the bottom (I've tested and the window style still looks good) and have both 'X' and 'ESC' save changes and close the window

Nilvus:
> If the buttons were on the bottom-right I would agree that OK/Cancel in that order seems reasonable and expected.

Short text is better, for that I'm ok

> IMO the least-worst compromise is to remove the hard-coded action header bar, moving the (empty) action area back to the bottom (I've tested and the window style still looks good) and have both 'X' and 'ESC' save changes and close the window
...or put buttons back on the bottom-right

Could you post screenshots of that both proposals to have better idea? You know that I like to see how is the UI visually.

elstoc:
So this was why we removed the bottom button bar - because it messes up the styling of the sidebar.

Screenshot_2020-04-27_12-35-34

And the alternative (no buttons) has the advantage that, certainly for me, it increases the amount of space dedicated to the window content.

Screenshot_2020-04-27_12-38-13

In both cases the header bar is generated by my Window Manager and with this option I could if I wanted (a lot of people do) remove it entirely.

Nilvus:
Indeed, the first one is not wanted as it mess up the sidebar.
For the second one, the problem is that also makes this window by default not consistent with other dialog windows (see import image and folder one for example), or simply on preferences window, click on export/import buttons in presets/shortcuts categories.

Maybe another option would be also add (in general category) an option to hide/show the headerbar on preferences window... Just an idea. We could also do the same way as import window:
- on the left of the headerbar: cancel/exit button
- on the right one: save/apply button

elstoc:
Another option is to change how the preferences window is styled so that we can put the buttons back on the bottom. I'm not a fan of preferences for the preferences window layout. Note that on the import window the cancel is on the left-hand-side.

mbaumgae:
From a usability point of view, the buttons in the header bar are very unusual. I would strongly vote for buttoms at the bottom of the dialog, as users are used to dialog windows with buttons at the bottom. 
All other dialog windows in darktable are showing the usual buttons at the bottom as @Nilvus already mentioned.

elstoc:
Yes, but @Nilvus is also talking about preferences windows in other apps to make the opposing argument (that there should be no buttons at all). I'm willing to bet that Gnome (and others as well) also removed the buttons because they preferred to get the style looking nice rather than for any logical/functional reason.

If we're worried about Ok/Cancel buttons at the bottom interfering with those on the shortcuts/presets tabs (which are also at the bottom), those could easily be moved to the top of the respective tabs. And there are other ways to visually indicate which tab is selected than by background colour alone.

elstoc:
It _might_ be possible to integrate the Ok/Cancel buttons into each of the individual tabs. Maybe.

elstoc:
For example (and bear in mind I just threw this together)

Screenshot_2020-04-27_14-47-13

Nilvus:
> From a usability point of view, the buttons in the header bar are very unusual. I would strongly vote for buttons at the bottom of the dialog, as users are used to dialog windows with buttons at the bottom.
> All other dialog windows in darktable are showing the usual buttons at the bottom as @Nilvus already mentioned.

I've never said. Read again and just check import image or folder (on lighttable) or even import/export buttons on shortcuts/preferences categories in preferences. Buttons are on the headerbar. It's not so unusual (read just after why) but I think it's good there. 

And as @elstoc said, I have proposed the opposite arguments like Gnome preferences with no buttons at all. But we could also see than on default UI guidelines in Gnome, buttons are put in the headerbar. See Nautilus for example or even Gnome Contacts, Gnome Calendar, etc, they all have menus/buttons in the headerbar. And that's also good as using the headerbar is a good way to use empty space. And finally I prefer that way than at the bottom. My personal choice would be no button at alls but as @elstoc said: more work to do.

Nilvus:
> It _might_ be possible to integrate the Ok/Cancel buttons into each of the individual tabs. Maybe. If we did that it wouldn't upset the style, because we wouldn't be using the action area.

Hmm, not sure about that. What the behavior behind that? Ok save all categories or only the current one? First thing not a good one, second also.

Finally on what we have, I really think there's only 2 good possibilities:
- no button at all but need to do more things as you know and already answer here
- buttons on the headerbar with same way we can see on import folder or image in lighttable, so with cancel on left of the headerbar and ok on the right.

parafin:
Gnome design code doesn't apply to dt, so you don't have to make the same choices as Gnome - they might be crappy;)

Nilvus:
> Gnome design code doesn't apply to dt, so you don't have to make the same choices as Gnome - they might be crappy;)

That's just examples based on Gtk, nothing else... and your comment doesn't help...

parafin:
These examples tend to be a little bit one-sided. GUI design isn't limited by used toolkit, so why consider only Gtk apps?

I don't think buttons in the header is a way to go - no other dt dialog works like that. Internal consistency is much more important than looking like current Gtk fashion.

Nilvus:
> These examples tend to be a little bit one-sided. GUI design isn't limited by used toolkit, so why consider only Gtk apps?

Why consider Gtk apps? Maybe because I don't have many apps and they're based on Gtk...  I will not install apps I don't need just to check something else. Again, just examples... and I also answer to other possibilities.

Firefox and Thunderbird I also use as examples are not Gnome apps on what I know.

> I don't think buttons in the header is a way to go - no other dt dialog works like that. Internal consistency is much more important than looking like current Gtk fashion.

Really? Did you read what I wrote? Open darktable and click on import module on the left then click on image... and folder...

To help a screenshot:
Capture-20200427185525-1125x700

So please, propose ideas, how you would see that (by thinking consistency) instead of reading what you think as oriented things where they're not (examples that I are just examples with what I have on my system, nothing more) and don't forgot to see other arguments and examples I have post here.

The only idea here is to propose all ideas with examples, good things and bad ones of each idea. As I said, I'm for no buttons at all, except a close button on the window and settings applied when change. But I'm not against other ideas if they're:
- consistent (to be precise that's a problem already on dialog windows. Main ones like import ones have buttons on headerbard but for example option button on export module have buttons on the bottom right
- keep the UI of the actual preferences window and the sidebar
- think of user UX
- think of both users types: those who prefer using keyboard and those prefer using mouse (and by the way others who use both).

elstoc:
> Open darktable and click on import module on the left then click on image... and folder...

That's not what it looks like for me. Are you sure your OS isn't putting those buttons in the header bar?

Screenshot_2020-04-27_18-31-05

elstoc:
GtkFileChooser dialogs are a bad choice for comparison since, afaik, they are almost entirely constructed by your WM and different for everyone.

Nilvus:
I've always had those buttons there so. And as I said, buttons on dialog window of export options ar not on the headerbar. And as I only have skills on CSS, how can I know that it's not hardcoded that way? I've also work on this CSS part before (align buttons and how they were displayed on headerbar and nobody answer me that was not hardcoded, even after saying that three times here before I post my screenshot. So how can these buttons can move from bottom to headerbar? Even if it's my OS (how I can know that as it is only on those windows), that means for me that how is coded on darktable is possible on these windows, because some dialog windows don't move similar buttons on headerbar.

Sorry but I'm not a diviner and just answer on what I have on my side... So it's not about who is right or not, it's about be clear on things we talk about... I would have like having questions and not critics.. Would be more constructive  than some last posts...

Nilvus:
> GtkFileChooser dialogs are a bad choice for comparison since, afaik, they are almost entirely constructed by your WM and different for everyone.

Indeed, with last unpleasant discussion caused by that here, I totally agree. That's introduced some misunderstanding.

If we change that, that would need to change all import/export dialog windows so to have consistent UI and avoid that.

After reading your answer about the fact your not against having only X if it's not a way to cancel, I totally agree. X should not be the way to cancel but only close window.
Having only that would be the best way but remains textview problem. But that's a big problem if it's not possible to apply change directly after setting the content of textview.

So if that last way is not possible, we have to find a visual way to put buttons at the bottom by keeping the sidebar effect.

By now, the only way I see to do that is to have those buttons on a box in the whole length (so yes below sidebar also) and use borders/margins probably to find a compromise...

Maybe propose a PR with WIP label could be a way to make some tests.

elstoc:
> with last unpleasant discussion

I don't see anything unpleasant written by anyone else @Nilvus - the word crappy was hedged with the word 'might' and it was not directed at you but at the possibly questionable decisions of Gnome developers.

> Even if it's my OS (how I can know that as it is only on those windows)

If you're concerned with unifying darktable look/feel it might be worth running another desktop environment/window manager alongside gnome so that you don't get tripped up by this stuff, since look/feel is pretty much what separates the various flavours of Linux desktop.

> Maybe propose a PR with WIP label could be a way to make some tests

You can try https://github.com/elstoc/darktable/tree/prefbuttons2 if you want to experiment. I'll issue a PR when we can agree what should be in it.

elstoc:
oh and also

> X should not be the way to cancel but only close window.

I don't have an issue with 'X' cancelling if there are OK/Cancel buttons at the bottom. What I was saying was that with 'save' next to it, it looked wrong and my brain basically ignored the save button when I went to close the window and went immediately to the X. With both of the buttons as text and the default on the right, that satisfied my brain.

elstoc:
No buttons at all and X/ESC always saving does at least have the advantage that all of the tabs work the same, and you don't have special behaviour for presets/shortcuts/general. Certainly for shortcuts 'exit without saving' is probably not possible (or too difficult to be worth bothering with) and I kind of like the immediate behaviour on themes (the ability to tweak CSS/theme and immediately see the difference).

Nilvus:
> > with last unpleasant discussion
> 
> I don't see anything unpleasant written by anyone else @Nilvus - the word crappy was hedged with the word 'might' and it was not directed at you but at the possibly questionable decisions of Gnome developers.

My english is not perfect so maybe it could by I made some errors like that. But  I think it's better to also propose ideas than just go on a PR, than interpret something (my suppose Gnome oriented way) and only write critics. But close that as it's all about misunderstanding.

> > Even if it's my OS (how I can know that as it is only on those windows)
> 
> If you're concerned with unifying darktable look/feel it might be worth running another desktop environment/window manager alongside gnome so that you don't get tripped up by this stuff, since look/feel is pretty much what separates the various flavours of Linux desktop.

You're right but I will not do it anyway, that's the first time I have had and, by the way, this misunderstanding here show that some part are not done to have same way/feel. So some things to make more consistent. And the most important reason I do not want to install another environment I'm a minimalist so like to have only essential thing and nothing more.

> You can try https://github.com/elstoc/darktable/tree/prefbuttons2 if you want to experiment. I'll issue a PR when we can agree what should be in it.

Thanks, will check.

Nilvus:
I agree that having text beside an X button is not a good thing so this could be forgotten.

elstoc:
> You're right but I will not do it anyway

Up to you but imo this is equivalent to testing web pages on multiple browsers - especially with Linux where everyone has their own personal favourite desktop environment / window manager and look/feel varies a great deal. Personally the enforced look/feel of gnome is one of the reasons I don't use it. They want to move functionality into the window header, but in most cases I don't even want to _see_ the window header - I want it completely out of the way.

> That's exactly why I think having no buttons at all except X/Esc is the better way. But is it possible to do that for all settings?

That's what my original PR did (before the save button) and I put a screenshot of it above - here it is again...

Screenshot_2020-04-27_12-38-13

In this screenshot the header bar is entirely under the control of the window manager so I could use 'X' to close or I could use 'ESC' if, for example I chose not to have a header bar display at all.

In this case, there's not really much difference between applying immediately (which is what the general/shortcuts/presets tabs do) and applying when you close the window (the rest of the prefs). Either way you have no ability to cancel your change other than by changing it again before you exit.

Nilvus:
@elstoc: when I was working on some websites (few years ago) of course, having all browsers installed was a minimalist way to work good. But here, I only work on darktable CSS UI and so just a few days 3 to 5 times per year since more than a year now.

As I now see the preferences on your prefbuttons 2 branch, I agree that not having the headerbar originally saved on your preferences rework PR, it's better to not have it. And if with that no more buttons and all settings saved and applied automatically (even textviews field), it's probably the better one. So I admit that my mistake error. Testing and discussion are always the better way.

Said that, I test your prefbuttons2 and if buttons, that could go to something like that (but better without buttons):

Capture-20200427223145-1139x800

That's just a draft that's need some minor tweaks but a good idea. And having the possibility (now or later) to have those buttons could then be a way to add a reset button as I see on an app I forgot to check (and not Gnome UI way) who have reset, cancel and ok button at the bottom right.

But adding a reset button would need a popup alert, etc.

So now, this need to confirm if we keep no buttons (would be my first choice) or cancel ok buttons at the bottom right as my screenshot here).

elstoc:
I don't think there's a way to get the OK/Cancel buttons at the bottom as an entirely separate bar without it looking ugly, though as I illustrated above they could be added to each individual tab (with exactly the same functionality) and retain the whole-height separator between the sidebar and the tab. From a user point of view they would see the Ok/Cancel in the same position on every tab so I don't think it would make a user think it was only related to the specific individual tab

However, the best and most consistent way is no buttons and no ability to cancel as in my last screenshot (i.e. back to the functionality of my first PR but without the hard-coded header bar). @TurboGit are you happy with this (i.e. revert back to user being unable to exit without saving at all)? I'd like my next PR on this subject to be my last!

Nilvus:
> I don't think there's a way to get the OK/Cancel buttons at the bottom as an entirely separate bar without it looking ugly, 

On without looking ugly I agree, that's why my first choice remains on no buttons at all. My proposal was to just a try to have a less looking ugly if we have them. So I'm ok to forget that solution.

> though as I illustrated above they could be added to each individual tab (with exactly the same functionality) and retain the whole-height separator between the sidebar and the tab. From a user point of view they would see the Ok/Cancel in the same position on every tab so I don't think it would make a user think it was only related to the specific individual tab

Not sure for that but maybe...

> However, the best and most consistent way is no buttons and no ability to cancel as in my last screenshot (i.e. back to the functionality of my first PR but without the hard-coded header bar). @TurboGit are you happy with this (i.e. revert back to user being unable to exit without saving at all)? I'd like my next PR on this subject to be my last!

I now totally agree with that. Best way. And after all not being able to exit without saving is consistent with some settings already (theme, shortcuts, presets...) and so that just mean preferences settings "save" setting on last remaining state by the user when closing. Simple, no buttons to deal with. Great.

chhil:
Its preference related but not buttons related.

The cpu/gpu/memory one is the widest and when that tab is clicked, the window of its preferences moves right. No other tab shows this behavior.
I have added a ruler to show the pixel shift.

image

image

elstoc:
> Its preference related but not buttons related

Not an issue for me but if you want it looked at please can you raise an issue.

chhil:
> Not an issue for me but if you want it looked at please can you raise an issue.

https://github.com/darktable-org/darktable/issues/4836

Nilvus:
@elstoc: and what about the question of buttons in preferences window? Do you plan to propose a PR soon about that? If yes, I could wait to propose a new PR for CSS where I could add the fix for the issue by @chhil with the changes in CSS needed by your work on those buttons.

elstoc:
I was waiting for @TurboGit to feed back since if I go back to no buttons I'm reverting a change he suggested.

TurboGit:
Ok, I'm not fully ok, but having a partial "no save" is not really appealing. So, go ahead and revert those buttons. If we want to come back with "quit without saving" we will want this to be valid for all items on the dialog.

That being said, do we still want to allow `ESC` key to quit the dialog ? This often mean close without saving.

elstoc:
> That being said, do we still want to allow ESC key to quit the dialog ?

I would rather leave that behaviour. Some people don't have their window manager display  decorations so for those people it's the only way out. For me, my 'X' button is really tiny (as you can see in my screenshots) so it's a nice easy way to exit.

elstoc:
#4851

PR #4827, opened 2020-04-27
Laplacian cleanup

ralfbrown:
Combine SSE and scalar alternative functions such that common code only occurs once.  Why would we ever *not* want to run the faster vectorized code if the target supports it?

### DIFF
--- a/src/common/locallaplacian.c
+++ b/src/common/locallaplacian.c
@@ -188,8 +188,7 @@ static inline __m128 convolve14641_vert(const float *in, const int wd)
 }
 #endif
 
-#if defined(__SSE2__)
-static inline void gauss_reduce_sse2(
+static inline void gauss_reduce(
     const float *const input, // fine input buffer
     float *const coarse,      // coarse scale, blurred input buf
     const int wd,             // fine res
@@ -198,6 +197,8 @@ static inline void gauss_reduce_sse2(
   // blur, store only coarse res
   const int cw = (wd-1)/2+1, ch = (ht-1)/2+1;
 
+#if defined(__SSE2__)
+  // this is the vectorized code
 #ifdef _OPENMP
   // DON'T parallelize the very smallest levels of the pyramid, as the threading overhead
   // is greater than the time needed to do it sequentially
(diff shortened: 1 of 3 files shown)


TurboGit:
I'm wondering if the best would not be to use clone-targets. Look at the toneequal.c implementation, you'll see the use of __DT_CLONE_TARGETS__ in declaration. This is defined in common/darktable.h as:

`#define __DT_CLONE_TARGETS__ __attribute__((target_clones("default", "sse2", "sse3", "sse4.1", "sse4.2", "popcnt", "avx", "avx2", "avx512f", "fma4")))`

The idea is that only the main std CPU version is implemented and we let the compiler create the sse, avx... versions which will be called depending on the support on the target hardware.

parafin:
> Why would we ever _not_ want to run the faster vectorized code if the target supports it?

To validate both codepaths against each other. It's not used right now I think, but let's not remove this possibility.

parafin:
> I'm wondering if the best would not be to use clone-targets. Look at the toneequal.c implementation, you'll see the use of **DT_CLONE_TARGETS** in declaration. This is defined in common/darktable.h as:
> 
> `#define __DT_CLONE_TARGETS__ __attribute__((target_clones("default", "sse2", "sse3", "sse4.1", "sse4.2", "popcnt", "avx", "avx2", "avx512f", "fma4")))`
> 
> The idea is that only the main std CPU version is implemented and we let the compiler create the sse, avx... versions which will be called depending on the support on the target hardware.

I wouldn't just throw away hand-optimized code in favour of compiler-optimized without some benchmarking first. And even then one has to consider that different compilers are used by users, so results might be unpredictable in real life.

parafin:
Overall I'm don't think this PR should be merged. I'm not sure if it's the case now for all iops, but probably process() shouldn't use SSE intrinsics and #ifdef __SSE2__.

aurelienpierre:
I agree with @parafin, having SSE and plain C intricated like that and branching internally doesn't make the code easier to manage and risks degrading performance in subtle ways. Please keep the SSE and C paths separated, so the branching happens at an high level.

TurboGit:
Since we three devs agree about the bad idea to interleave SSE and standard code I'll close this PR.

Issue #4813, opened 2020-04-25
Tagging module request : more consistency within tag creation


GwenaelQ:
Hello,

**Context** 
Pipe caracter is used to create hierarchical tags

**Issue**
Creating a tag with a pipe (that is to say a "path") is allowed within entry zone.
The same is not allowed with  the "create tag" command of dictionary view  !

**Expected behavior**
Both ways should be equivalent and allow the creation of a hierarchical tag

Thank you for listening. Keep the good work.

DT 3.0.2 on windows 8.1

phweyland:
> Creating a tag with a pipe (that is to say a "path") is allowed within entry zone.
> The same is not allowed with the "create tag" command of dictionary view !

"entry zone" defines a full tag (with its full path).
"create tag" creates a tag on top of an existing one.
They cannot be equivalent.

This said, I agree that "create tag" could include a subpath (on top of the current one) and then allow the pipe character.

phweyland:
Done

sjjh:
@phweyland  this issue can be closed, can't it? Might you? :)

phweyland:
> this issue can be closed, can't it? Might you? :)

yes, agreed.

Issue #4812, opened 2020-04-25
Tagging module : creating tag with "text entry" improvement request


GwenaelQ:
Hello,

Please allow user to create tag by the mean of  the "text entry" **even if no image is selected**.
This is the fastest way of creating one and as it is it is quite frustrating !

Thank you for listening

DT 3.0.2 on windows  8.1

phweyland:
I agree this frustrating.

phweyland:
Done

TurboGit:
Closing now.

Issue #4810, opened 2020-04-25
Tagging module : "attached tags view" minor issue


GwenaelQ:
Hello there

First of all thank you for the work. DT is evolving rapidly in many aspects. Among these I'm particularly sensitive to the work done on tagging and metadata.

**Context**

"attached tags view" displays the attached tags of selected images (if no image mouse hovered) **or** those of the hovered image. [Source : [DT manual](https://darktable.gitlab.io/doc/en/tagging.html)]

**The issue:**
Indicators are displayed at the left  of the tags when an image is hovered.

**Expected behavior**
No indicator should be displayed when hovered image tags are displayed. If any it should be the check-mark since the tag is attached to all (=1) hovered images.

Furthemore, **no indicator at all** when hovering an image would make things **much more clear** about if the display is of selected or hovered image(s)

**Platform**
 - Darktable 3.0.2 on Windows 8.1

Thank you for listening

phweyland:
> Indicators are displayed at the left of the tags when an image is hovered.

I cannot reproduce this.

phweyland:
I've removed the indicator in attached view when hovered image or unique selected one.

sjjh:
@phweyland can this get closed as well? Might you? :)

phweyland:
yes, this one can be closed too.

PR #4828, opened 2020-04-27
tag fix - improvement

phweyland:
- allows pipe character in create tag
- allows entry tag(s) creation without image selected
- makes visible and selects the newly created tag in dictionary view
- removes selection indicator in attached view when only one image (selected or hovered)
- fixes wrong indicators when attach tag without selected image
- enum for selection states

Covers some of @GwenaelQ issues / suggestions: #4810, #4812, #4813.

### DIFF
--- a/src/common/tags.c
+++ b/src/common/tags.c
@@ -510,44 +510,36 @@ void dt_tag_set_tags(const GList *tags, const GList *img, const gboolean ignore_
 
 void dt_tag_attach_string_list(const gchar *tags, const GList *img, const gboolean undo_on, const gboolean group_on)
 {
-  if(!img) return;
   // tags may not exist yet
   // undo only undoes the tags attachments. it doesn't remove created tags.
   gchar **tokens = g_strsplit(tags, ",", 0);
   if(tokens)
   {
+    // tag(s) creation
+    GList *tagl = NULL;
+    gchar **entry = tokens;
+    while(*entry)
+    {
+      char *e = g_strstrip(*entry);
+      if(*e)
+      {
+        guint tagid = 0;
+        dt_tag_new(e, &tagid);
+        tagl = g_list_prepend(tagl, GINT_TO_POINTER(tagid));
+      }
+      entry++;
+    }
+
+    // attach newly created tags
     GList *imgs = g_list_copy((GList *)img);
     if(imgs)
     {
       GList *undo = NULL;
       if(group_on) dt_grouping_add_grouped_images(&imgs);
       if(undo_on) dt_undo_start_group(darktable.undo, DT_UNDO_TAGS);
-      GList *tagl = NULL;
-
-      gchar **entry = tokens;
-      while(*entry)
-      {
-        // remove leading and trailing spaces
-        char *e = *entry + strlen(*entry) - 1;
-        while(*e == ' ' && e > *entry)
-        {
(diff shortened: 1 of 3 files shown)


src/common/tags.c
```diff
@@ -510,44 +510,44 @@ void dt_tag_set_tags(const GList *tags, const GList *img, const gboolean ignore_
...
   // undo only undoes the tags attachments. it doesn't remove created tags.
   gchar **tokens = g_strsplit(tags, ",", 0);
   if(tokens)
   {
+    // tag(s) creation
+    GList *tagl = NULL;
+    gchar **entry = tokens;
+    while(*entry)
+    {
+      // remove leading and trailing spaces
```
TurboGit, line 523: Maybe use g_strstrip()

src/libs/tagging.c
```diff
@@ -221,7 +221,64 @@ static void _sort_dictionary_list(dt_lib_module_t *self, gboolean force)
...
+  return found;
+}
+
+// make the tag visible on view
+static void _show_tag_on_view(GtkTreeView *view, const char *tagname)
+{
+  if(tagname)
+  {
+    char *lt = g_strdup(tagname);
+    // remove leading and trailing spaces
```
TurboGit, line 257: likewise g_strstrip()

src/libs/tagging.c
```diff
@@ -506,6 +567,7 @@ static gboolean _find_tag_iter_tagid(GtkTreeModel *model, GtkTreeIter *iter, gin
   return FALSE;
 }
 
+#if 0
```
TurboGit, line 570: Either the code should be removed or some comments added to why it is commented out.

TurboGit: I'm in favor of removing, it is in Git anyway.

src/common/tags.c
phweyland: Great !

src/libs/tagging.c
phweyland: >  it is in Git anyway.

True. I'll remove it.

phweyland:
Review comments integrated. Thanks.

TurboGit:
>     * makes visible and selects the newly created tag in dictionary view

Can you explain this one ? Where should I look for my testing ?

phweyland:
entering aa|bb|cc you should get something like :

image

Entering aa|bb|dd, zz|aa|zz should give something like (open the last tag entered):

image

Should work in list view as well but not in suggestion view (as already attached tags are not shown).
Should work also with create tag and rename path (context menu).

TurboGit:
I don't get the selected text on my side:

image

TurboGit:
I have tested when the selected image as the corresponding tag or not.

phweyland:
> I have tested when the selected image as the corresponding tag or not.

The view you show is a filtered one.(selection has gone for a while).
What do you see just after enter ?

TurboGit:
> What do you see just after enter ?

It is selected as shown on your screenshot. Perfect then.

> * makes visible and selects the newly created tag in dictionary view

I read this as two behaviors "make visible created tag in dictionary view" and "select the newly created tag in dictionary", my bad why reading the simple sentence when one can read something overly complex. Anyway, now I see it all good.... Merging.

Thanks.

PR #4831, opened 2020-04-27
Fix the filters support for rejected images.

### LINKED ISSUE #4823: reject image doesn't go to rejected image filter anymore
**Describe the bug**
If I reject image with "all except rejected images" filter activated on lighttable, the image should disappear when rejected but stay on lighttable view. If I change filter mode on top left to rejected images only, rejected images doesn't appear anymore.

I've tested that on my personal config but also with a default config folder with only a set of imported images: same issue.

**Expected behavior**
Apply as before rejected image in regards of activated filter mode. So having all rejected images visible on all images and rejected only ones only.

darktable master ; Ubuntu 20.04

@TurboGit or @AlicVB: could you confirm that? Could you be one of your latest PR/commit about reject toggle or lighttable work?

TurboGit:
Fixes #4823.

### DIFF
--- a/src/common/collection.c
+++ b/src/common/collection.c
@@ -267,7 +267,7 @@ int dt_collection_update(const dt_collection_t *collection)
        && collection->params.sort_second_order == DT_COLLECTION_SORT_COLOR))
      && (collection->params.query_flags & COLLECTION_QUERY_USE_SORT))
   {
-    selq_pre = dt_util_dstrcat(selq_pre, "SELECT DISTINCT mi.id FROM (SELECT * FROM main.images WHERE ");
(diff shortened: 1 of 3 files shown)


Nilvus:
A first step is done: if I'm on all image except rejected filter, when I reject an image, the image now disappear. But If I go to rejected filter mode, the rejected images are not there. If I then go on all filter mode, I see the image rejected with the red cross.

TurboGit:
New version should be working now.

Nilvus:
All good now. #4823 fix now. Thanks for this fix.

TurboGit:
Great! I'll merge then especially since I have fixed most filters which were broken!

PR #4833, opened 2020-04-27
fix export private tag

### LINKED ISSUE #4826: "private" status is not respected for parent tags
**Describe the bug**
Parent tags that have a different "private" status compare to their child tag are incorrectly exported / not exported. 

**To Reproduce**
Set two hierarchical tags like "tag1" and "tag1|tag2". Apply "tag1|tag2" to an image. 
1. If you set "tag1" as private and "tag2" as public (=private option unticked), the exported image will have both "tag1" and "tag2": "tag1" is exported even it is set as private. 
2. If you set "tag1" as public and "tag2" as private, the exported image will have no tag, even though "tag1" is set as public.  

(in both cases, let unticked "private tags" and "omit hierarchy" in metadata configuration in export module)

**Expected behavior**
I would expect the "private" status to be respected both for child and parent tags. With the examples above, it would mean having in the exported file: 
1. only "tag2"
2. only "tag1"

I supposed the algorithm first check if the (child) tag is private or not, and then export the tag and its parent tags. I think the condition should be check for parent tags separetly. 

**Platform**
 - Darktable Version: 3.0.1 (I have not had the possibility to upgrade to 3.0.2 yet; I did search for similar issue to check if this behavior was already change but I did not find any)
 - OS: Linux Mint 19.1

phweyland:
fixes #4826

### DIFF
--- a/src/common/tags.c
+++ b/src/common/tags.c
@@ -696,41 +696,27 @@ uint32_t dt_tag_get_attached(const gint imgid, GList **result, const gboolean ig
   return count;
 }
 
-uint32_t dt_tag_get_attached_export(const gint imgid, GList **result)
+static uint32_t _tag_get_attached_export(const gint imgid, GList **result)
 {
+  if(!(imgid > 0)) return 0;
+
   sqlite3_stmt *stmt;
   dt_set_darktable_tags();
-  if(imgid > 0)
-  {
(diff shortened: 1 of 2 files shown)


phweyland:
I've simplified a bit as dt_tag_get_list_export() works only for a specific image.

TurboGit:
Looks ok to me. Thanks.

PR #4835, opened 2020-04-28
imageio: avif: Add support for libavif >= 0.7.0

cryptomilk:
Add support for libavif >= 0.7.0


parafin:
What happes with libavif 0.7.0 without this patch?

Did libavif change ABI version of the library at this release (it's a number after libavif.so.)?

I'm concerned about what happens when dt is built against older libavif and then libavif is updated.

cryptomilk:
libavif changed the API and bumped the SO version number from libavif.so.2 to libavif.so.3 for 0.7.0.

```
magrathea:~ # rpm -qa | grep libavif[0-9]
libavif3-0.7.1-1.1.x86_64
libavif2-0.6.2-1.1.x86_64
```

parafin:
OK then

TurboGit:
Looks good to me, thanks !

Issue #3008, opened 2019-09-26
Image scale issue on HiDPI monitor [Windows 10]


Kelvin-Wu:
Hi All. I use darktable on my Win10 PC with 4K monitor. 
Since the resolution is very high, I changed the "size of text, apps and other items" to 200%. (I use chinese version.)
![dpi](https://user-images.githubusercontent.com/55777209/65655350-7ebe9180-e04e-11e9-811d-35439d92ccc4.jpeg)

When I use scale 100% function in datktable, the image size look like 200% not 100%.
![dt](https://user-images.githubusercontent.com/55777209/65655354-85e59f80-e04e-11e9-8bad-39f21cd059e6.jpeg)

But the scale function is work correct on other apps like Lightroom, Rawtherapee, and digiKam.
They show the 100% resolution of image.

**Lightroom**
![lr](https://user-images.githubusercontent.com/55777209/65655357-8bdb8080-e04e-11e9-8846-6d19a52546dc.jpeg)

**Rawtherapee**
![rt](https://user-images.githubusercontent.com/55777209/65655362-9433bb80-e04e-11e9-9e75-a593b89a74c8.jpeg)

**digiKam**
![digi](https://user-images.githubusercontent.com/55777209/65655365-985fd900-e04e-11e9-92eb-b7ab7986eab3.jpeg)

aurelienpierre:
Yes, it's not just on Windows. The 100 % zoom in darkroom is actually a 200 % view if you set the DPI scaling equal to 200 %. Same here, on 4K, to get 1:1 pixel scaling, I actually need to use the 50 % zoom level.

Kelvin-Wu:
I think this problem should be fixed. Or....can we have modified shortcut can make quick switch between 50% and fit screen?

sboukortt:
I agree that it should probably be fixed, as getting 1:1 mapping:

1. is possibly the main use-case for selecting “100%”;
2. is otherwise impossible with certain scaling factors (200% is the “easy case”).

snaggen:
It sounds like the devs need to use gdk_window_get_scale_factor or similar, to get the scaling factor and compensate for it.

https://wiki.gnome.org/HowDoI/HiDpi

Kelvin-Wu:
Hope it can be fixed one day...

jenshannoschwalm:
I think this is fixed, at least for osx and Linux. There is a remaining issue with a secondary monitor non-hdi though.

[...]

PR #4839, opened 2020-04-28
darkroom - navigation : fix wrong scaling factors

aurelienpierre:
This is a WIP attempt to work on the scaling error that happens on High DPI screens (https://github.com/darktable-org/darktable/issues/3008) on both Linux and Windows, as shown in #4816 screenshots. It might also have a connection with #4702 

When using a PPD scaling factor of 2, the zoom scale of the darkroom is twice as big, so the 100% zoom factor is actually 1:2 (200%) and Cairo surface interpolation is used.

To set-up the PPD scaling factor, you use your display environement global GUI scaling or force `screen_ppd_overwrite=2` in `darktablerc`. To test proper scaling, use a square image which input height is around the physical pixel height of your monitor, and see if it fits entirely the viewport.

This is a first attempt that solves part of the issue (the zooming factor displayed in the navigation is properly scaled) but inputting a specific zoom level fails and zooming in the picture with mouse wheel only zooms out. So I need help with that, because the scaling handling across libs and views is quite intricated.

### DIFF
--- a/src/libs/navigation.c
+++ b/src/libs/navigation.c
@@ -200,8 +200,8 @@ static gboolean _lib_navigation_draw_callback(GtkWidget *widget, cairo_t *crf, g
     int closeup = dt_control_get_dev_closeup();
     float zoom_x = dt_control_get_dev_zoom_x();
     float zoom_y = dt_control_get_dev_zoom_y();
-    const float min_scale = dt_dev_get_zoom_scale(dev, DT_ZOOM_FIT, 1<<closeup, 0);
-    const float cur_scale = dt_dev_get_zoom_scale(dev, zoom, 1<<closeup, 0);
+    const float min_scale = dt_dev_get_zoom_scale(dev, DT_ZOOM_FIT, 1<<closeup, 0) * darktable.gui->ppd;
+    const float cur_scale = dt_dev_get_zoom_scale(dev, zoom, 1<<closeup, 0) * darktable.gui->ppd;
     // avoid numerical instability for small resolutions:
     double h, w;
     if(cur_scale > min_scale)
@@ -443,8 +443,8 @@ static void _zoom_preset_change(uint64_t val)
   }
   else if(val == 4u)
   {
-    scale = 0.5f;
-    zoom = DT_ZOOM_FREE;
+    scale = 0.5f * dt_dev_get_zoom_scale(dev, DT_ZOOM_1, 1.0, 0);
+    zoom = DT_ZOOM_1;
   }
   else if(val == 5u)
   {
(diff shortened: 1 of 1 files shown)


TurboGit:
@aurelienpierre : Thanks for starting this and giving a way to reproduce. I'll try to allocate some brain-cycle to this in the coming days.

GrahamByrnes:
In #4702, the killer failing for me was the zooming: it didn't refuse to zoom in, but it first zoomed out to a roughly preview-sized view, then zoomed in again. I've tried again with only the bits of code in develop.c and pixelpipe_hb.c, I've tried moving the code in and out of the mutex... and it's the same.

The delay was about a second for each click of zoom... it would drive anyone mad.

I've no idea if this is what you are experiencing, but I agree that scaling information is intricate!

aurelienpierre:
> In #4702, the killer failing for me was the zooming: it didn't refuse to zoom in, but it first zoomed out to a roughly preview-sized view, then zoomed in again. I've tried again with only the bits of code in develop.c and pixelpipe_hb.c, I've tried moving the code in and out of the mutex... and it's the same.

Same issue here.

GrahamByrnes:
After more fiddling and random changes, it seems my jittering zoom issues were coming from dt_dev_get_zoom_scale(), which makes sense given the commonality.

Anyway, it seems that if you remove ` /= preview_downsampling` from the definition of `ps` then add a scaling after the switch, it behaves.

It might give you something to try with your PPD zoom, if you haven't already been there.

Here's the whole function:

`float dt_dev_get_zoom_scale(dt_develop_t *dev, dt_dev_zoom_t zoom, int closeup_factor, int preview)
{
  float zoom_scale;

  const float w = preview ? dev->preview_pipe->processed_width : dev->pipe->processed_width;
  const float h = preview ? dev->preview_pipe->processed_height : dev->pipe->processed_height;
  const float ps = dev->pipe->backbuf_width
                       ? dev->pipe->processed_width / (float)dev->preview_pipe->processed_width
                       : dev->preview_pipe->iscale;

  switch(zoom)
  {
    case DT_ZOOM_FIT:
      zoom_scale = fminf(dev->width / w, dev->height / h);
      break;
    case DT_ZOOM_FILL:
      zoom_scale = fmaxf(dev->width / w, dev->height / h);
      break;
    case DT_ZOOM_1:
      zoom_scale = closeup_factor;
      if(preview) zoom_scale *= ps;
      break;
    default: // DT_ZOOM_FREE
      zoom_scale = dt_control_get_dev_zoom_scale();
      if(preview) zoom_scale *= ps;
      break;
  }
  if (preview) zoom_scale /= dev->preview_downsampling;                                                  
  return zoom_scale;
}`

TurboGit:
This is now fixed, so closing.

PR #4844, opened 2020-04-28
lighttable : ensure redraw after empty window

### LINKED ISSUE #4834: lighttable not updated when empty
**Describe the bug**

lighttable not updated when empty

**To Reproduce**
1. Select a collection
2. Select any filter in the top-left of the lighttable containing no images (5 stars for example)
3. As expected the lighttable is empty
4. With the mouse scroll change the filter value there will be no image displayed
5. Set a filter where you know there is some iimages (nothing displayed)
6. Click on the lighttable view, it forces a refresh and the thumbs are now displayed

What is strange is that the lighttable is refreshed correctly when changing some thumbs are displayed while change filter value.

**Expected behavior**

The lightttable is always refreshed.

**Screenshots**

 - Darktable Version: master
 - OS: Linux
 - 
 - 

**Additional context**

This is most probably due to latest lt rewrite from @AlicVB

AlicVB:
this fix #4834

### DIFF
--- a/src/dtgtk/thumbtable.c
+++ b/src/dtgtk/thumbtable.c
@@ -1228,6 +1228,7 @@ static void _dt_collection_changed_callback(gpointer instance, dt_collection_cha
       }
       l = g_list_next(l);
     }
+    dt_control_queue_redraw_center();
   }
   else
   {


TurboGit:
Working good ! Thanks.

PR #4845, opened 2020-04-28
crop and rotate : allow float/real cropping format ratios at input

aurelienpierre:
1. treat 5:4 or 4:5 the same way for consistency since, the portrait/landscape orientation doesn't depend on format ratio,
2. always reduce fractions using the greatest common divider, so inputting 6:4 is equivalent to 3:2,
3. allow real format ratios, so inputting 1.5 is equivalent to 0.6667 or 3:2 or 2:3, by finding the closest rational fraction,
4. search if the custom input ratio matches a default one and update the label accordingly.
5. unify the format ratio combobox update with a single path.

### DIFF
--- a/src/iop/clipping.c
+++ b/src/iop/clipping.c
@@ -1629,7 +1629,7 @@ static void aspect_presets_changed(GtkWidget *combo, dt_iop_module_t *self)
         // some sanity check
         if(dd == 0)
         {
-          dt_control_log(_("invalid ratio format. it should be greater than one"));
+          dt_control_log(_("invalid ratio format. it should be non zero"));
           dt_bauhaus_combobox_set(combo, 0);
           return;
         }
--- a/src/iop/clipping.c
+++ b/src/iop/clipping.c
@@ -1594,7 +1594,7 @@ static void aspect_presets_changed(GtkWidget *combo, dt_iop_module_t *self)
   dt_iop_clipping_gui_data_t *g = (dt_iop_clipping_gui_data_t *)self->gui_data;
   dt_iop_clipping_params_t *p = (dt_iop_clipping_params_t *)self->params;
   int which = dt_bauhaus_combobox_get(combo);
-  int d = p->ratio_d, n = p->ratio_n;
+  int d = abs(p->ratio_d), n = p->ratio_n;
   const char *text = dt_bauhaus_combobox_get_text(combo);
   if(which < 0)
   {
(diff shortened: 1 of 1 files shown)


src/iop/clipping.c
```diff
@@ -1798,22 +1924,6 @@ void gui_update(struct dt_iop_module_t *self)
...
-  {
-    const dt_iop_clipping_aspect_t *aspect = iter->data;
-    if((aspect->d == d) && (aspect->n == n))
-    {
-      act = i;
-      break;
-    }
-    i++;
-    iter = g_list_next(iter);
-  }
```
aurelienpierre, line 1816: this is now gone in `aspect_presets_changed`

src/iop/clipping.c
```diff
@@ -1828,20 +1938,7 @@ void gui_update(struct dt_iop_module_t *self)
...
-  if(act == -1)
-  {
-    char str[128];
-    snprintf(str, sizeof(str), "%d:%d", abs(p->ratio_d), p->ratio_n);
-    dt_bauhaus_combobox_set_text(g->aspect_presets, str);
-  }
-  if(dt_bauhaus_combobox_get(g->aspect_presets) == act)
-    aspect_presets_changed(g->aspect_presets, self);
-  else
-    dt_bauhaus_combobox_set(g->aspect_presets, act);
```
aurelienpierre, line 1844: this is now gone in `aspect_presets_changed`, and the case where `dt_bauhaus_combobox_get(g->aspect_presets) == act` is now a no-op.

src/iop/clipping.c
```diff
@@ -1615,9 +1616,101 @@ static void aspect_presets_changed(GtkWidget *combo, dt_iop_module_t *self)
...
+      {
+        // find the closest fraction from the input ratio
+        float rr = atof(text);
+        int dd = ceilf(rr);
+        int nn = floor(rr);
+
+        // some sanity check
+        if(dd == 0)
+        {
+          dt_control_log(_("invalid ratio format. it should be greater than one"));
```
parafin, line 1632: greater than zero? or just "positive"

parafin: Also it doesn't look like you're checking for negative values either.

aurelienpierre: ooops, good catch.

aurelienpierre: the only blocking case is if ==0, negative values are not an issue.

aurelienpierre: edit : negative values make it joyfully crash, but that wasn't handled in the previous code either.

TurboGit:
Works for me. Thanks!

PR #4846, opened 2020-04-28
Use new slider_set_soft_range instead of confusing enable_soft_boundaries

dterrahe:
As discussed on IRC, added dt_bauhaus_slider_set_soft_range which should be less confusing than the currently used dt_bauhaus_slider_enable_soft_boundaries.

Also removed some behaviour that I thought might be unexpected/unwanted. Now setting soft limits doesn't move hard limits or the current value of the slider. The hard limits will be used to guard the values allowed by introspection, so it should be very clear (with set_hard_min/max) if they are manipulated.

I have cursorily checked that nothing depends on the previous behavior (there are not that many calls to set soft limits after the initial widget definitions) but am not giving guarantees because this code is confusing and the naming may have misled people.

Filmicrgb now uses the new function, but I intend to implement it accross all iops when I go over them to add introspection, because that will get rid of a lot of the dt_bauhaus_slider_new_with_range calls anyway.

I'm not completely happy with the name "dt_bauhaus_slider_set_soft", or it's use, either. Its purpose isn't to set soft limits, but to set the position value and move the active limits if the value to be set doesn't fit within that range. The _normal_ dt_bauhaus_slider_set assumes that the current value is within the active limits. The code usually calls dt_bauhaus_slider_set if there are no soft limits. Which means adding soft limits would require carefully vetting the code to see if _soft needs to be added anywhere.

### DIFF
--- a/src/bauhaus/bauhaus.c
+++ b/src/bauhaus/bauhaus.c
@@ -752,7 +752,8 @@ void dt_bauhaus_slider_set_hard_max(GtkWidget* widget, float val)
   d->max = MIN(d->max, d->hard_max);
   d->soft_max = MIN(d->soft_max, d->hard_max);
   if(rawval < d->hard_min) dt_bauhaus_slider_set_hard_min(widget,val);
-  if(pos > val) {
+  if(pos > val) 
+  {
     dt_bauhaus_slider_set_soft(widget,val);
   }
   else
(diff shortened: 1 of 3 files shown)


TurboGit:
Ok, works for me. Thanks.

AxelG-DE:
I did not do a bisect yet [I am not really familiar], but I am afraid, this one caused a regression (but I can be wrong)

I have a Logitech MX Master S2, where the mousewheel can spin freely. I use this function often...

Now, when I go to WB e.g. and spin the mouse wheel, what happens is, the slider stands still and only when I stop the wheel, the slider "jumps" to a value might be represented by the passed spins (while I was able to see the changes thru spinning) in the past (until yesterday)...

PR #4851, opened 2020-04-28
preferences: remove header bar and buttons

elstoc:
removes the hard-coded header bar and buttons from the preferences dialog.


Nilvus:
Thanks @elstoc: better for me that way.

TurboGit:
Thanks !

PR #4855, opened 2020-04-29
Fix reading non-existing new timestamps from old sidecar

jenshannoschwalm:
In `read_xmp_timestamps` we look for the new timestamps and write values into the database.

While importing an old sidecar without those keys available we dont't find them, the comma
removing is bad as there might be none and the syntax for sqlite updating is wrong.

This ends in all_ok being FALSE and import fails to read the sidecar in a correct way, in fact it
loses the whoöe history.

How to reproduce? Import an image with an old sidecar (before introducing the new timestamps)
on the console.

### DIFF
--- a/src/common/exif.cc
+++ b/src/common/exif.cc
@@ -3290,8 +3290,12 @@ gboolean read_xmp_timestamps(Exiv2::XmpData &xmpData, const int imgid)
     if((pos = xmpData.findKey(Exiv2::XmpKey(xmpkey))) != xmpData.end())
     {
       snprintf(tmp, sizeof(tmp), " %s = %ld,", timestamps[i], pos->toLong());
-      g_strlcat(values, tmp, sizeof(values));
     }
+    else
+    {
+      snprintf(tmp, sizeof(tmp), " %s = 0,", timestamps[i]);
+    }
+      g_strlcat(values, tmp, sizeof(values));
   }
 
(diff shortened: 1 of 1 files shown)


src/common/exif.cc
```diff
@@ -3290,8 +3290,12 @@ gboolean read_xmp_timestamps(Exiv2::XmpData &xmpData, const int imgid)
...
-      g_strlcat(values, tmp, sizeof(values));
     }
+    else
+    {
+      snprintf(tmp, sizeof(tmp), " %s = 0,", timestamps[i]);
+    }
+      g_strlcat(values, tmp, sizeof(values));
   }
 
   values[strlen(values) - 1] = '\0'; /* remove last comma */
```
TurboGit, line 3301: I suppose a condition just about this line would fix the issue and be clearer:

```
if(strlen(values) > 0)
{
   values[strlen(values) - 1] = '\0'; /* remove last comma */

   ...
}
```

And then we run the SQL stmt only if needed.

jenshannoschwalm:
Yes, running the sql conditionally probably would be ok but don't we want something specific to be set?

Anyway, this is really high priority as it would break darkroom development while reimporting or re-reading exif data.

Please commit as you like, i am at work and won't have access to my system until tonight lately ...

TurboGit:
I have pushed a fix, indeed urgent as it broke dt and there was a write out of bound. Do not hesitate to review to ensure all is fine now.

Issue #4849, opened 2020-04-28
File extension in new lighttable overlaps with thumbnail


Mark-64:
In the new lighttable, with the following overlay modes
- permanent overlays
- permanent extended overlays
- permanent overlays extended on mouse hover

the file extension indicator in the upper left of the thumbnails overlaps with the image (it goes under).
This happens only for certain zoom levels and it also depends on the image aspect ratio.
See the example below:

![Cattura_w5](https://user-images.githubusercontent.com/43290988/80514042-41bdd300-8980-11ea-91e5-dfe3a961481a.JPG)

Note that the "altered" and "group" icons never overlap, they are scaled correctly.
Perhaps it could be enough to align the file extension indicator to the very top of the thumbnail, just like the icons.

Tested with DT 3.1 git master on Windows 10 and Ubuntu 19.10, HiDPI display

Nilvus:
@Mark-64: yes, I've seen that when working on CSS. It's just indeed on overlays modes you precise only. That don't bother me so I don't post an issue. Said that, it can't be change on CSS. From what I see, the only way would be to change Gtk code and move the extension name the box/widget above, so on same line as group/audio showed on the top right. And that's why altered/group icons are never overlapped.

And from what I see on how thumbnails are made with Gtk Inspector tool, that could move other things so not sure it will be easily done.

@AlicVB: what do you think about that?

AlicVB:
yes, that's "by design". the new overlays mode just show that more obviously. Previously, if you try with a square image, for example,you'll see that.

Mark-64:
In thumbnail.c, line 918, changing top margin for the extension label from 
`gtk_widget_set_margin_top(thumb->w_ext, 0.045 * width);`
to
`gtk_widget_set_margin_top(thumb->w_ext, 0.5 * r1);`
the label will be aligned with the altered / group / audio icons and it will never be cut.

![Screenshot from 2020-04-28 22-18-12](https://user-images.githubusercontent.com/43290988/80534074-f2d36600-899e-11ea-8e93-8464da57d8bf.png)

@AlicVB @Nilvus, it doesn't look bad to me, what do you think ?

Nilvus:
That's just breaks also the goal of this new overlay mode: use more space for the image. See margins that adds on left, right and bottom in regards of your first screenshot with the "issue".

AlicVB:
[...]

PR #4863, opened 2020-04-29
fix file extension label in thumbnails

Mark-64:
Tentative fix for #4849


TurboGit:
Looks perfect and properly aligned with the group and altered icon now. Thanks.

PR #4864, opened 2020-04-29
culling & preview rewrite

### LINKED ISSUE #4515: Culling mode: reject image goes on last selected ones not last shown ones
**Describe the bug**
Since lighttable rework, culling mode have a different behavior. Now, rejected an image (with all except rejected ones filter) make the selection come back to first selected ones and not last ones seen (except rejected one of course)

**To Reproduce**
1. Having a set of images to see on culling mode (at least 5 ones to be sure understanding the behavior)
2. Select first one and go to culling mode with 2 images seen each time
3. Go for example to image 4 and reject the image. The image disappear
4. See that selection come back to images selected (here image 1 and 2) instead of images 2 and 3 like before. So need to go back to last shown images manually to continue selection (quite annoying).

**Expected behavior**
As before, don't reset to selected ones on culling mode but just last shown images, by respecting of course filter selection.
@AlicVB: Of course, as I know you will rewrite culling mode, this is for your to-do list! One more thing!

### LINKED ISSUE #4882: Culling mode: easy way to get to 100% zoom with mouse

The culling mode is great to compare several pictures. It's the perfect tool to select the right shot for macrophotography where I often have tens of pictures of the same thing and only a few sharp enough. But for this use-case, culling is only really relevant with a 100% zoom. For now, I have two minor user-experience glitches:

* The middle-click shortcut in darkroom to get to 100% is great, but not available in culling mode. AFAIK, the middle-click is not mapped to any action on culling mode, so it would make sense to get the same behavior as in darkroom.

* The zoom level is not displayed anywhere, so if I control-scroll to get to 100%, I don't get a visual feedback when I reach 100%. Usually zooming is a bit slow and therefore I can't distinguish "I did control-scroll and nothing happened because my computer is being slow" and "I reached 100%". But I have to admin that I didn't think of any good place to display the zoom level properly (displaying it on top of the image would be distracting, displaying it elsewhere would be misleading since the zoom level can be per image). Perhaps displaying the zoom level of the picture where the mouse is in the message area on top of the screen, below "X images out of Y in current collection are selected" would make sense.

* Same middle-click behavior in culling mode as in darkroom

* Perhaps, display the zoom level somewhere

**Alternatives**

There are `<Darktable>/views/lighttable/preview zoom 100%` and `<Darktable>/views/lighttable/preview zoom fit` keyboard shortcuts that allow getting to 100% and back to "fit" zoom with the keyboard. Almost as good, but not as convenient as the middle-click behavior to me.

AlicVB:
this is the follow up of the filemanager rewrite, but for culling and preview

It should globally have the same behavior and ui than before, except the overlays which are maybe a little bit less beautiful now, but far more usable while keeping almost the same disposition : 
- semi-transarent background
- size adapted to the thumbnail
- disappear after a delay, and not depending on the mouse position, which doesn't works well for culling.

Considering the amount of change, I expect a lot of small glitches and issues, so feel free to test, review, and report any problems.

After that, just one big piece is remaining for the lt rewrite : the "images_to_act_on" stuff... I hope it will be easier ;)

Fix #4515 Fix #4882


AlicVB:
Arrgh... test build failed with this message : 
```
/build/darktable/src/dtgtk/thumbnail.c:1161:49: error: use of undeclared identifier 'r1'
  gtk_widget_set_margin_top(thumb->w_ext, 0.5 * r1);
                                                ^
1 error generated.
```

Although what I have in the latest code at line 1161 (checked here too) is : 
`gtk_widget_set_margin_top(thumb->w_ext, 0.045 * width);`

Does someone know what I have done wrong ?

elstoc:
That line was in a commit very recently: c055cf04

src/views/view.c
```diff
@@ -1357,7 +1354,7 @@ int dt_view_image_expose(dt_view_image_expose_t *vals)
      && dt_view_lighttable_get_layout(darktable.view_manager) == DT_LIGHTTABLE_LAYOUT_CULLING)
   {
     // in culling surrounded images are the ones shown in main view
-    surrounded = dt_view_lighttable_culling_is_image_visible(darktable.view_manager, imgid);
+    // surrounded = dt_view_lighttable_culling_is_image_visible(darktable.view_manager, imgid);
```
TurboGit, line 1357: to be removed

src/gui/gtk.c
```diff
@@ -1964,7 +1955,7 @@ void dt_ui_panel_show(dt_ui_t *ui, const dt_ui_panel_t p, gboolean show, gboolea
   else if(p == DT_UI_PANEL_RIGHT)
     gtk_widget_queue_draw(darktable.gui->widgets.right_border);
 
-  dt_view_lighttable_force_expose_all(darktable.view_manager);
+  // dt_view_lighttable_force_expose_all(darktable.view_manager);
```
TurboGit, line 1958: to be removed

src/gui/gtk.c
```diff
@@ -364,7 +364,7 @@ static gboolean _panels_controls_accel_callback(GtkAccelGroup *accel_group, GObj
   gtk_widget_set_visible(GTK_WIDGET(darktable.gui->widgets.top_border), visible);
   gtk_widget_set_visible(GTK_WIDGET(darktable.gui->widgets.bottom_border), visible);
 
-  dt_view_lighttable_force_expose_all(darktable.view_manager);
+  // dt_view_lighttable_force_expose_all(darktable.view_manager);
```
TurboGit, line 367: to be removed

src/views/view.c
```diff
@@ -1724,7 +1721,7 @@ int dt_view_image_expose(dt_view_image_expose_t *vals)
       cairo_save(cr);
       cairo_new_path(cr);
     }
-    else if(buf_ok && dt_view_lighttable_culling_is_image_visible(darktable.view_manager, imgid))
+    else if(buf_ok) // && dt_view_lighttable_culling_is_image_visible(darktable.view_manager, imgid))
```
TurboGit, line 1724: comment to be removed

elstoc:
I've just been browsing through the shortcuts tab on the preferences dialog and I notice that there are some lighttable-related shortcuts under modules>lighttable (culling and zoom settings) and modules>preferences (mostly overlay settings).

Unless I knew about them beforehand I wouldn't think of looking under 'modules' for these options. Would it be possible to move them into the views>lighttable section so they're all in a more obvious location?

AlicVB:
@elstoc : the main problem is that they are in different places because they are driven by different part of dt. For example in your case, the shortcuts are handled by the global_toolbox lib which is present in almost all views (that's ok, because you have either filemanager either filmstrip in all views) and not by lighttable view.
But overall I'm ok with you that the current situation is confusing, I'm not sure how to solve this, and anyway this is worst a new FR as this concern far more situations than culling/preview

elstoc:
@AlicVB thanks. I agree that it's not easy to resolve (part of the reason I raised it here rather than submitting a PR myself - in case you had some ideas). I'll have a think on it for a bit and see if I can come up with something.

AlicVB:
@TurboGit : got it, thanks. It seems that git automatic rebasing was wrong in fact :(
@elstoc : hence the proposition of a Feature request first, to launch the discussion

TurboGit:
Is that still wip ?

AlicVB:
@TurboGit : it's ready for testing / review, and works afaict but as I expect some forgotten cases to fix, so I have tag this as WIP, maybe it's not the right tag...
I've just fixed your comments

TurboGit:
Thanks! I'll do some more testing on my side, but I'm pretty sure some corner cases will only be detected by having more field testing, so I'll merge if I see nothing wrong. Thanks for all the nice work you've made on this.

TurboGit:
I cannot rate or reject an image in the culling view. Clicking on any star or X does nothing. It works ok from the lighttable.

Nilvus:
@TurboGit: could you wait to merge. I will test that this evening and see especially if my issue #4515 is fixed (@AlicVB: maybe you test/fix that on this PR).

I will also see what things I will have to done on CSS this week-end.

TurboGit:
@Nilvus : Sure, I'll wait for you go before merging. I haven't seen another issue than the rating/reject on culling mode at the moment.

AlicVB:
@TurboGit : it was "just" a refresh problem. Should be fixed now
@Nilvus : yes it should fix the issue, for culling and preview (same code now)

Nilvus:
> I cannot rate or reject an image in the culling view. Clicking on any star or X does nothing. It works ok from the lighttable.

Strange, reject works on culling mode but indeed not stars. On preview mode, most of the times, reject or stars doesn't work but don't know if moving to an image or another, reject image works (not all times). Not find a reproducible way to know when it really works all times but most of the times that doesn't work.

Other thing but I don't know if it could be improved: I really don't find the new way zoom works and culling/preview. The image has some annoying "flash" between each zoom (I use mouse wheel). It's more sensible than before (but also much much faster).

Last thing: the new block infos on hover effect is great (will need some tweak to remain visible and take less place on the image) and first thing is why letting so much empty space on the right of image infos and stars? I don't see anything on CSS for that (no margins/padding or min-width set that could justifying this empty space. So is it harcoded and could you reduce to have only the block content related and only have the possibility to tweak space and margin/padding with CSS?

Other than that, I confirm that my #4515 issue is fixed. I never see lighttable and culling/preview mode so fast. That's an awesome mode and will help selecting images much faster. Many many thanks to you for that work.

Nilvus:
The "flash" I talk about is also when launching for first times culling mode on a new set of images. Maybe the way darktable update cache image to show them. When launching a second time the same images on culling mode (no "flash"). I'm not sure to be clear here...

Nilvus:
Two more things: 

- first a good one: I confirm that last commit fix the reject/stars issue. So great now.
- but another issue: ctrl+z doesn't seems to work on preview mode. To test, just reject an image then press Ctrl+z doesn't work, the image is still rejected. Exit preview mode, Ctrl+z works on lighttable (it works too on culling mode).

AlicVB:
@Nilvus : I see what you mean. I think that I need to implement some sort of double buffering here. I'll have a look, that shouldn't be too hard.
For the Ctrl-Z, I'll have a look too.
Thanks for testing !

Nilvus:
> @Nilvus : I see what you mean. I think that I need to implement some sort of double buffering here. I'll have a look, that shouldn't be too hard.
> For the Ctrl-Z, I'll have a look too.
> Thanks for testing !

For Ctrl+z, I've made four or 5 tests and doesn't work but don't know why, I try to have Ctrl+Z works on preview mode (but after using it on culling mode).

AlicVB:
@Nilvus : this should fix the flickering and the ctrl-z
@phweyland : in https://github.com/darktable-org/darktable/pull/4864/commits/b679222523618eb5832d3f40b31890891d85abf8 i've moved the info_changed callback to thumbnail struct instead of thumbtable. This simplify a little the code, but above all is more generic as thumbnail can belong to culling or preview too. As it's your code, can you have a look ?

AlicVB:
@Nilvus : for the block overlays : 
- the size is now minimized as much as possible
- the timeout is now configurable with an hidden option in darktablerc : `plugins/lighttable/overlay_timeout` default duration is 3 s. (was 2s. before)

phweyland:
> @phweyland : in [b679222](https://github.com/darktable-org/darktable/commit/b679222523618eb5832d3f40b31890891d85abf8) i've moved the info_changed callback to thumbnail struct instead of thumbtable. This simplify a little the code, but above all is more generic as thumbnail can belong to culling or preview too. As it's your code, can you have a look ?

If I understand properly all thumbnails will individually answer this signal (instead of the thumbnail table). I don't see any issue.

Nilvus:
@AlicVB: no more flickering, great. Seems a little bit slower but maybe it's only my system (it remains faster than previous culling. Ctrl+z is now ok.

overlay block is also better, will just need some CSS tweak (but I will work on that).

The main remaining and annoying issue is the zooming part. Would it be possible to have a way to keep the zooming on the part of the image wanted? Unzoom is worse, not fluid and the image doesn't stay on same visible part.

AlicVB:
> The main remaining and annoying issue is the zooming part. Would it be possible to have a way to keep the zooming on the part of the image wanted? Unzoom is worse, not fluid and the image doesn't stay on same visible part.

zooming around pointer is done for preview, but it can't happen for culling due to the multiple images. That said, I'll try to zoom images around their visible center, I think it should better...

Nilvus:
> zooming around pointer is done for preview, but it can't happen for culling due to the multiple images. That said, I'll try to zoom images around their visible center, I think it should better...

Yes, this should at least be much better.

Nilvus:
@AlicVB: I don't have a really powerful computer (and no OpenCL). I have my images generated until -m 6 setting with darktable-generate-cache so darktable normally doesn't need to calculte/render the image. It just have to render the generated image (I don't talk about 100% one because I don't generate on 7 level). What I don't understand (it was also the case with first culling mode) it's why I regularly having some CPU overload when zooming or changing image (not all times but quite regularly. When I have that CPU overload work, I have to wait to could change anything on culling mode (zooming or panning or even change images displayed). darktable seems to calculte lot of things and so became usable when CPU load diminish. And sometimes, it even stuck temporarly my system (after viewing a set of images.

Is it normal and just my CPU limits or is something that could be improved?

My computer is a Macbook pro (under Linux, Ubuntu 20.04), late 2013, with i5-4258U and an Iris Graphics 5100. I also have 8go RAM and Swap 4,5Go. My settings on darktable are already optimized (with some help from Aurélien).

AlicVB:
So... I should have fixed all reported issues (thanks @Nilvus for the tests)
I've also added a zooming level info in the overlays.

@Nilvus : for your perf questions, I see 2 possible "problems" : 
- cache size : not all cached images are stored in RAM, the quantity depend of some settings. That would explain the "not always"
- specific "big images" cache : to avoid any delay when panning, dt load in memory the full image at the corresponding zoom (even the "hidden" parts) so it can very quickly be big images and due to mipmap level mechanism, as soon as the zoomed image is greater than 4096x2560 you need level 7 (and you need to load full image if greater than 5120x3200) Note that there's also a specific cache size for "big images"
- rescaling : once the mipmap loaded, we need to rescale the image to the exact wanted size. If this is the problem, culling should be way faster now, as we do than one time and not at each redraw as before.

Nilvus:
@AlicVB: thanks for your feedback about my perfs questions. I doubt it's cache size, I set it to 8000Mb on dt prefs and when I have overload (before, during and after), max RAM size used is about 5Gb (on 8Gb RAM) and no more than 900Mb of Swap ; on a total of 4,5Gb).

For some big zooming, yes setting à cache with 7 level should be could (I've tried to save space to not use them) but I have same overload even on less 40-50% of zoom max (from 5264x3904 max size). Will generate 7 level images this night. I will see tomorrow.

But maybe it's only the limits of computer...

Another (minor) issue: on preview mode, the hover effect could only be disable (to see the whole image without overlay) when moving the mouse cursor to the very bottom of the darktable window. Before, it was on the about 10-20% top of the image to have the overlays. Could it be something between that ? I would see, ideally, something like:
- 50% top of the preview view: display the overlay block
- 50% bottom, don't have the overlay (so hover effect).

Nilvus:
Other than that, center view for zooming and whole other things are really great and this new culling release is much better than the first one. Having also the zoom level on the block infos is a really great (and well made) idea. Thanks to you for that (again) awesome work. And thanks to @moy for his good ideas.

TurboGit:
>  Could it be something between that ? I would see, ideally, something like:
> 
>     * 50% top of the preview view: display the overlay block
> 
>     * 50% bottom, don't have the overlay (so hover effect).

Agreed, that's what I had implemented for the full preview in previous version.

TurboGit:
@AlicVB : I have issue with the pan & zoom.

The zooming seems ok but I cannot pan the image anymore. Maybe a refresh issue as when I pan nothing happens but when I zoom again it seems that the center area has changed somehow. It is hard to describe precisely as the behavior seems heratic on my side.

Also, I cannot zoom image independently has done in previous version.

AlicVB:
For the "new" block overlay, fact was that the position of the mouse to determine the visibility doesn't works well in culling as you constently move the mouse between image, and mouse position doesn't means anything in this case.
So I've implemented what is done in the other soft I use : hide the overlay after a delay without mouse movement. The delay is configurable and if the mouse is inside the overlay block, it stay always visible.
Of course we can use different behavior for culling than preview, but I'm reluctant to do that, as this will break consistency between overlay modes (think that this new overlay mode can be used in filemanager too)

For the zoom and pan, I'll test, I may have broke something with latest commit, as it was working before... That was in culling and preview ?

TurboGit:
@AlicVB: yes in culling and preview.

Nilvus:
@AlicVB: about the overlay mode to display block infos or not, maybe 2 seconds was better than 3. But said that, I think an option to hide/show the block overlay without this delay could be better. That's why we suggest a way depending on where the mouse cursor is, to hide/show that. Especially when having only one image displayed.

AlicVB:
@TurboGit : should be fixed now, that was due to my very latest changes. Sorry...

For the overlay : here are the constraints I've took in account for the design : 
- should work as the other overlays, and be used in preview, culling, filemanager, ... in a consistent and reproducible way
- should not depend of mouse position because it only work for preview. In culling for example, if you have to images one above the other and you move the mouse from bottom to top, overlay visibility is completely unexpected (there's some bug report for that) It's even worse for filemanager/zoomable
- should not be permanently shown, even if mouse is over the image
- should have a design close to the previous one

I'm absolutely not sure to have found the good design, thought, so I'm open to any suggestion... (as usual)
If I understand well your use cases, maybe changing the default timeout and hiding the overlay at first until mouse move ? (that would avoid to see overlay when going trough images with keyboard)

TurboGit:
I have tested this again. The pan&zoom works now for me. For the overlay I understand the constrains and what I propose (if @Nilvus is OK too) is to merge this as-is now that it is ok and see what others would propose for the overlay after some more use in real conditions.

How does that sound ?

moy:
I did a bit of testing, and got a segfault, backtrace attached:

[darktable_bt_QFHIJ0.txt](https://github.com/darktable-org/darktable/files/4567886/darktable_bt_QFHIJ0.txt)

(Un)fortunately, I couldn't reproduce (I cleared my config and cache directories in the meantime, perhaps this fixed it).

I do like the new UI. A few remarks:

* At first, I didn't understand the "mini" mention in the overlay. "zoom fit", or just "fit" is probably a more common wording to say the same thing.

* I obviously like the new "middle-click to zoom" feature, but a minor nit: it would be even better if the zoom was made around the point clicked instead of being done at the center of the image. Actually, the same remark would apply to the Control-wheel zooming.

* When zooming in, the lack of progress indicator gives surprising results. For example, I zoomed on one of the images, which didn't have a full-scale version cached while the other ones had, and I saw the other image zooming but not the one on which I had my mouse (=> "hey, why are you zooming the wrong image?")

* If I resize the dt window, the 100% zoom is no longer a 100% zoom (this was present before the rewrite). The image gets zoomed out as I rescale the window. Not really a problem, since I never resize the dt window though ;-)

* I got more segfault playing with the orientation (turn left/turn right) buttons in the "selected image" module of lighttable while in culling mode. Not completely reproducible either, but I could trigger it twice:

[darktable_bt_P7HZJ0.txt](https://github.com/darktable-org/darktable/files/4567911/darktable_bt_P7HZJ0.txt)
[darktable_bt_I7QVJ0.txt](https://github.com/darktable-org/darktable/files/4567912/darktable_bt_I7QVJ0.txt)

Thanks again for the work on this feature!

AlicVB:
>     * At first, I didn't understand the "mini" mention in the overlay. "zoom fit", or just "fit" is probably a more common wording to say the same thing.

you're certainly right, and this will use the same terminology than in other dt parts

>     * I obviously like the new "middle-click to zoom" feature, but a minor nit: it would be even better if the zoom was made around the point clicked instead of being done at the center of the image. Actually, the same remark would apply to the Control-wheel zooming.

It does zoom around pointer if there's only 1 image. If you have more than 1 it's tricky : 
- the mouse can be anywhere in the screen, even outside images areas
- even if you do center around pointer only if the mouse is hovering an image, we have to apply this zooming center to all images (if want to compare images, you need to see the same place) which can be really complex and unpredictable if images have different ratios...

>     * When zooming in, the lack of progress indicator gives surprising results. For example, I zoomed on one of the images, which didn't have a full-scale version cached while the other ones had, and I saw the other image zooming but not the one on which I had my mouse (=> "hey, why are you zooming the wrong image?")

You're right here too. I'll try to propose something, but after this PR merging if possible (I've a few ideas)

>     * If I resize the dt window, the 100% zoom is no longer a 100% zoom (this was present before the rewrite). The image gets zoomed out as I rescale the window. Not really a problem, since I never resize the dt window though ;-)

I'm not sure it's worst the pain to add complexity to handle this corner case. If needed, exiting-reentering culling, or even changing images should solve it.

>     * I got more segfault playing with the orientation (turn left/turn right) buttons in the "selected image" module of lighttable while in culling mode. Not completely reproducible either, but I could trigger it twice:

I'll try to reproduce and fix this.

Thanks for all your tests and remarks.

AlicVB:
@moy : I've not reproduced your crash, but thanks to your backtraces, I think I've managed to fix them. Can you have a second try ? Thanks !
Oh and I've replaced mini by fit too...

Nilvus:
> It does zoom around pointer if there's only 1 image. If you have more than 1 it's tricky :
> 
>     * the mouse can be anywhere in the screen, even outside images areas
> 
>     * even if you do center around pointer only if the mouse is hovering an image, we have to apply this zooming center to all images (if want to compare images, you need to see the same place) which can be really complex and unpredictable if images have different ratios...

Just an idea on that: would it be possible to use position of the mouse cursor (or click on an image) related to the border around the image. So if moving the mouse cursor on an image (or click), the zoom is made with that position and, for other images, zoom is made related to the position of the cursor to the border. So that apply the zoom on same position on all images.
And if zooming made with the cursor outside images, just use actual behavior: center of the image.

Just ideas, I of course don't if and how it could possible to implement that. And that also could be another PR.

AlicVB:
@Nilvus : this is possible, but I think that will fail if images have different ratios. For using a extreme case, Imagine 2 portraits images and a pano. The 2 portraits will appear on top, side to side and the pano on bottom, taking almost all the width...
Anyway, I agree with you, I prefer to not add too much things now, so this PR can be merged (and to avoid @TurboGit to do multiple reviews, he will have enough work like that !)

Nilvus:
@AlicVB: I never think of that problem. But I just use culling to compare images and which one I will reject and which one I will keep, so I never compare images with different ratios...

Just a thing about CSS part: I try to adjust with CSS padding and margins with overlay_block and I have 2 things I can do (or not correctly):
- the first one probably can't be done on CSS by now is to adjust the position on the overlays_block and move it a little more to the top (about 1/3 or maybe 50% of the empty side between top of the block and the top image border. Is it harcoded? Having a way to move it with css would be great.

- second thing is how to set margins inside the overlay block: there's a problem actually on the bottom side. For top, left and right, I only have to set padding on bottom_infos_label. But for the bottom, I need to set all icons that could appear on the bottom with same settings. It's not a problem to do that on CSS as you know. So the question is simply: wouldn't it be better to put them on a block? I just thing about users who will want to tweak them? If it would cause some other problems, forget that demand and I will set them together on CSS.

AlicVB:
@Nilvus : I've just pushed a new commit to implement the first point, with the same workaround as for image margins.
For the second point, it should be possible but will add lot of complexity in the code, so I'm tempted to let it like that if you don't mind...

Nilvus:
@AlicVB: just perfect for first point. Thanks.
About the second, that's why I precise at the end to forget it if it would cause problems. So adding a lot of complexity in the code is of course not wanted. So forget it.

For me, as it is, this PR could be merged. Great work again!

Nilvus:
@AlicVB: finally, there's a problem with point 2. I can of course set padding or margins but there's no visible. Indeed, background is applied on #thumb_bottom and icons (group, altered, fit...) are outside this widget but they are visible inside. So adding padding or margins can't change the visible padding at the bottom of these icons. So this padding stay fixed and can't be changed. I don't even understand these icons could be inside the background overlay block but this background can't be adjusted by padding.

So, if you see another idea (than adding a lot of complexity to the code), I'm on.
Anyway, by now I just set padding on left and right to improve it and it's good with default top and bottom ones (even I think it would be just more balanced with 1 or 2 px more on top and bottom ; but will not add on top if I can't at bottom one).

Nilvus:
Just 2 things discussed yesterday as reminders before merging: to be discussed more probably after merging this PR:

- The zooming mode and what I've beginning to be discussed about cursor place (and limits).
- Overlay block on culling and preview modes: having a possibility to hide/show overlay mode on demand. Moving cursor mode also reset the 3 second delays, so a little move and you go to 3 seconds more...

AlicVB:
@Nilvus : great !
For the css problem n°2, It'll tricky to allow css access to that, because in fact buttons are not inside the `#thumb_bottom` widget, but over.
What I can propose to you is to set it once for all to a correct and balancing value in the code.

If I understand well, you should be able to set the top/left/right margins with css, this is the bottom one which is problematic, right ?

If so, and as you compile from source, you can very easily play with line `1193` of `src/dtgtk/thumbnail.c`
the line is currently : `gtk_widget_set_size_request(thumb->w_bottom, CLAMP(w, 25 * r1, width), 6.5 * r1 + h);`
the height of `#thumb_bottom` (which define the bottom margin) is currently `6.5 * r1 + h` where `h` is the height of the text and each button is `3.0 * r1` (including their margins)
Just try to change the `6.5` value and compile again to see the change.

I can also do it for you if you want, but if you manage to do it alone, you'll be able to tweak it exactly to your needs ;)

Nilvus:
@AlicVB: I understand that would be tricky to allow css on that. I've made some tests and I think the best value would be 6.75 instead of 6.5 on line 1193. Could you update that or I add it on my next CSS PR? As you want. It's not a big change of course.

@TurboGit: I'm now working on updating french translation. I know you make it but I see one thing to fix and will update new ones from this PR. So If ok for you, you could forget working on that this time.

AlicVB:
@Nilvus : as you want... maybe include that to your css work, so it'll be consistent and correctly authored...

Nilvus:
> @Nilvus : as you want... maybe include that to your css work, so it'll be consistent and correctly authored...

Ok.

Last thing, it's not totally related to this PR but as you add on overlay modes a new line, I think it could be added there. Could you just add a CSS specific tag to the first line of overlay modes (popover when clicking on star icon). I think this line should be different than the modes lines. So the idea is to have the possibility to easily tweak it. bold font at least.

See highlighted one below:

TurboGit:
> So If ok for you, you could forget working on that this time.

Great, go ahead I'm short on time at the moment.

Nilvus:
@AlicVB: I just note that on culling mode the hover effect (overlay block infos, border...) disappear immediately when the cursor is outside any images and appear only when moving cursor hover an image. But it's not the case on preview mode as the hover mode disappear (except delay option of course) only if moving outside darktable window. So it's not consistent. I know this is a tricky part but just to be sure, wouldn't it be simply possible to have same behavior on both modes. That would be sufficient to solve what we already discussed about that I think.

If I miss something on what you already said, just excuse me for that!

AlicVB:
@Nilvus : good catch. In fact it works  in the same way, it's just that the hover effect in linked to mouse move in the thumbnail area (the image and the borders, which are transparent for culling and preview) and in preview the thumbnail takes the whole screen... I'll have a look on how to fix that.
And I'll fix your other point too, it should be trivial

Nilvus:
@AlicVB: just to take some news but of course take all time you need.

AlicVB:
@Nilvus : done !
- hover block overlays appears now when the mouse is over the image part only. This was a little bit harder than expected, and I fear that it will gives you headhaches for rebasing your css changes :)
- and I've named the overlays popup label so you should be able to use `#overlays_label`

TurboGit:
@AlicVB  : This does not apply to current master. Can you fix? Thanks.

TurboGit:
Fix needed to resolve conflicts:

```
diff --git a/src/views/lighttable.c b/src/views/lighttable.c
index 76dbb3757..983b4fe3b 100644
--- a/src/views/lighttable.c
+++ b/src/views/lighttable.c
@@ -645,7 +645,7 @@ int key_pressed(dt_view_t *self, guint key, guint state)
 static gboolean zoom_in_callback(GtkAccelGroup *accel_group, GObject *acceleratable, guint keyval,
                                            GdkModifierType modifier, gpointer data)
 {
-  int zoom = get_zoom();
+  int zoom = dt_view_lighttable_get_zoom(darktable.view_manager);
 
   zoom--;
   if(zoom < 1) zoom = 1;
@@ -657,7 +657,7 @@ static gboolean zoom_in_callback(GtkAccelGroup *accel_group, GObject *accelerata
 static gboolean zoom_out_callback(GtkAccelGroup *accel_group, GObject *acceleratable, guint keyval,
                                            GdkModifierType modifier, gpointer data)
 {
-  int zoom = get_zoom();
+  int zoom = dt_view_lighttable_get_zoom(darktable.view_manager);
 
   zoom++;
   if(zoom > 2 * DT_LIGHTTABLE_MAX_ZOOM) zoom = 2 * DT_LIGHTTABLE_MAX_ZOOM;
```

TurboGit:
@AlicVB : I found one issue, the stars on the full-preview or culling are often displayed even when the overlay block is not. This is more easy to reproduce with full-preview.

Also, I think we need a bit more padding on the left overlay. But that's probably for after.

Nilvus:
> @AlicVB : I found one issue, the stars on the full-preview or culling are often displayed even when the overlay block is not. This is more easy to reproduce with full-preview.
> 
> Also, I think we need a bit more padding on the left overlay. But that's probably for after.

Maybe it's just CSS tweaks needed for both ones. On last commit and as @AlicVB says, there's some headaches for me to adjust CSS from what I made on my side. Will see that this week-end (probably tonight and tomorrow).

Nilvus:
I'm now near to finish CSS tweaks about last commit. Was faster and simpler than expected.
Padding issue was just not setting (and was something already fixed on my first CSS work here ; so fix). 

@AlicVB: stars issue reported by @TurboGit is not a CSS one. Stars are only hidden if the mouse is not moved (outside or inside the image) after the 3 seconds delay. So when mouse cursor is outside the image with an image with some stars applied, stars remains visible and disappear after 3 seconds without moving the mouse. They appear if you move the cursor (outside or inside the image). Probably a forgotten things to integrate on last commit you made.

Nilvus:
@AlicVB and @TurboGit: good news, star issue is fixed. So CSS part ok now and this PR is ok for me.

So, if it's ok for both you, I think this PR could be merged.

I will propose my PR when this one will be merged... and after some things I want to do on CSS:
- see if I can a way to simplify these new lighttable/culling/preview CSS parts (more things are tweakable and changed so we have now some a huge and heavy part here)
- fix/improve visibility some colors/hover effects on thumbnails icons especially.

Just as a first thing about my CSS work, the star overlay popup on lighttable improved (related to last commit of this PR):

TurboGit:
Good news! So the stars display was a CSS issue !

Your proposal for the overly menu is nice.

I was also wondering if we could have some kind of indicator on the main star icon to display the current zoom level ? I don't have a good proposal... A number inside the star would be hard to read... A number after? Any other idea ? This will help seeing the different levels while zooming and so selecting the proper options for each level.

AlicVB:
@TurboGit : sorry, I'm quite overbooked this days. But that's great to see that @Nilvus is solving all that alone :)
I'll try to find the time tomorrow to rebase.

For your proposal about the maybe we can drop the star and replace by square with the number inside... (will be less meaningful, thought)

TurboGit:
@AlicVB : No problem, I can merge and fix conflicts as I have proposed above. For the star we can do that later. So, let me retest a bit today and if all goes well I'll merge.

TurboGit:
@AlicVB : Just merged this! Thanks to all the work.

@Nilvus : You can go ahead with your CSS work.

Nilvus:
> @AlicVB : No problem, I can merge and fix conflicts as I have proposed above. For the star we can do that later. So, let me retest a bit today and if all goes well I'll merge.

@TurboGit: I can't compile master with this PR merged. An error on lighttable.c. Seems you forgot to fix conflicts.

TurboGit:
Just saw that and now fixed. Fact is that I did resolved the conflicts but forgot to add the changes in my merge commit :(

moy:
Sorry for the late reply. I can't reproduce my crashes in master, so hopefully this is indeed fixed.

AlicVB:
@TurboGit : many thanks to you for finishing this !

PR #4874, opened 2020-04-30
preferences: add lua tab to new layout

### LINKED ISSUE #4868: Lua tab disappeared in the new options dialog
In the new options dialog with vertical tabs, the "lua options" is no longer present (was last tab in the old layout).
Lua scripts work and Options->Shortcuts-> lua is there, but "lua option" just disappeared.

Cattura30

DT 3.1 git master on Windows 10

elstoc:
Resolves #4868

### DIFF
--- a/src/lua/preferences.c
+++ b/src/lua/preferences.c
@@ -256,7 +256,7 @@ static int write_pref(lua_State *L)
 
 static void response_callback_enum(GtkDialog *dialog, gint response_id, pref_element *cur_elt)
 {
-  if(response_id == GTK_RESPONSE_ACCEPT)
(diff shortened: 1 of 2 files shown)


elstoc:
Don't merge this yet. I've just spotted another change.

elstoc:
I've checked that this compiles and @Mark-64 kindly tested that the new option appears but I've not been able to test that it still works correctly, as I don't use Lua. I'm fairly sure it should be ok though.

Mark-64:
I also did some test that lua options work and they do, at least with the scripts I use

elstoc:
Cool thanks again @Mark-64

TurboGit:
So this seems all good ! Thanks to both of you for restoring this.

PR #4877, opened 2020-04-30
grouping fix

### LINKED ISSUE #4840: Group indicator "G" is not reset when current group is diminished to a single image
openSUSE Tumbleweed 20200425
darktable 3.1.0~git1346.532efbf17-2885.1.x86_64
OpenCL on 
i7 12-core 36GB

Select an image in lightroom
duplicate the image, now a group exists and the original image is marked "G"
delete the original image in light table, the "G" indicator moves to the duplicate image, only remaining image, but there is no longer a group.

The "G" indicator should disappear.

phweyland:
grouping fix
- duplicate selected
- remove/delete selected (duplicate or real image)

cleanup unused code in control_jobs.c

fixes #4840

### DIFF
--- a/src/common/grouping.c
+++ b/src/common/grouping.c
@@ -82,6 +82,8 @@ int dt_grouping_remove_from_group(const int image_id)
     wimg->group_id = image_id;
     dt_image_cache_write_release(darktable.image_cache, wimg, DT_IMAGE_CACHE_SAFE);
     imgs = g_list_prepend(imgs, GINT_TO_POINTER(image_id));
+    // refresh also the group leader which may be alone now
+    imgs = g_list_prepend(imgs, GINT_TO_POINTER(img_group_id));
   }
   dt_control_signal_raise(darktable.signals, DT_SIGNAL_IMAGE_INFO_CHANGED, imgs);
 
--- a/src/common/image.c
+++ b/src/common/image.c
@@ -886,12 +886,14 @@ int32_t dt_image_duplicate_with_version(const int32_t imgid, const int32_t newve
 
     g_free(filename);
 
+    const dt_image_t *img = dt_image_cache_get(darktable.image_cache, imgid, 'r');
+    const int grpid = img->group_id;
+    dt_image_cache_read_release(darktable.image_cache, img);
     if(darktable.gui && darktable.gui->grouping)
     {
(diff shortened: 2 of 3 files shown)


src/common/grouping.c
```diff
@@ -82,6 +82,7 @@ int dt_grouping_remove_from_group(const int image_id)
     wimg->group_id = image_id;
     dt_image_cache_write_release(darktable.image_cache, wimg, DT_IMAGE_CACHE_SAFE);
     imgs = g_list_prepend(imgs, GINT_TO_POINTER(image_id));
+    imgs = g_list_prepend(imgs, GINT_TO_POINTER(img_group_id));
```
TurboGit, line 85: Can you explain this one ? I don't understand the use!

Especially since the signal is only caught in thumtable and calls `_dt_image_info_changed_callback` where we iterate over all values comparing to the imgid.

phweyland: One part of the bug is when one removes the second image image of a group of two. The main image remains with the group symbol. 
Then this line forces the refresh of thumbnail info for the remaining image.

> specially since the signal is only caught in thumtable and calls `_dt_image_info_changed_callback` where we iterate over all values comparing to the imgid.

yes, but only on the provided list (and not all the displayed images in lt).

TurboGit: Ok, maybe a small comment to clarify this in the code ?

TurboGit: And maybe while we are at it only register the img_group_id if different than the image_id:

`if(image_id != img_group_id) imgs = g_list_prepend(...);`

Nilvus:
@phweyland: group icons on thumbnails are now ok and all seems to work for what I test on my side. Thanks.

src/common/grouping.c
phweyland: > And maybe while we are at it only register the img_group_id if different than the image_id:
> 
> `if(image_id != img_group_id) imgs = g_list_prepend(...);`

That's already the case, line 50...  :)


> Ok, maybe a small comment to clarify this in the code ?

I'll do it.

phweyland: Done

TurboGit:
All good then. Thanks.

PR #4881, opened 2020-05-01
give kbd focus to centre image when clicking module header

### LINKED ISSUE #4875: <primary>plus and <primary>minus no longer zoom image
openSUSE Tumbleweed 20200428
darktable-3.1.0~git1383.0e7305a3d-2912.1.x86_64
OpenCL on
i7 12-core 36GB

I can no longer zoom image using <ctrl><+> or <ctrl><->
(gtk_accel_path "<Darktable>/global/zoom in" "<Primary>plus")
(gtk_accel_path "<Darktable>/global/zoom out" "<Primary>minus")

scroll wheel and two finger drag on mouse pad still work

elstoc:
Modified resolution to #4717, that doesn't prevent global keyboard
shortcuts from working, by giving keyboard focus to the centre view instead
of to the the module being expanded.

Resolves #4875

### DIFF
--- a/src/develop/imageop.c
+++ b/src/develop/imageop.c
@@ -1752,9 +1752,7 @@ void dt_iop_request_focus(dt_iop_module_t *module)
     gtk_widget_set_state_flags(dt_iop_gui_get_pluginui(module), GTK_STATE_FLAG_SELECTED, TRUE);
 
     //used to take focus away from module search text input box when module selected
-    gtk_widget_set_can_focus(dt_iop_gui_get_pluginui(module), TRUE);
-    gtk_widget_grab_focus(dt_iop_gui_get_pluginui(module));
-    gtk_widget_set_can_focus(dt_iop_gui_get_pluginui(module), FALSE);
+    gtk_widget_grab_focus(dt_ui_center(darktable.gui->ui));
 
     if(module->operation_tags_filter()) dt_dev_invalidate_from_gui(darktable.develop);
 
--- a/src/develop/imageop.c
+++ b/src/develop/imageop.c
@@ -557,7 +557,7 @@ static void dt_iop_gui_delete_callback(GtkButton *button, dt_iop_module_t *modul
 
   // we set the focus to the other instance
   dt_iop_gui_set_expanded(next, TRUE, FALSE);
-  gtk_widget_grab_focus(next->expander);
+  dt_iop_request_focus(next);
 
   const int reset = darktable.gui->reset;
   darktable.gui->reset = 1;
@@ -1751,9 +1751,6 @@ void dt_iop_request_focus(dt_iop_module_t *module)
   {
     gtk_widget_set_state_flags(dt_iop_gui_get_pluginui(module), GTK_STATE_FLAG_SELECTED, TRUE);
 
-    //used to take focus away from module search text input box when module selected
-    gtk_widget_grab_focus(dt_ui_center(darktable.gui->ui));
-
     if(module->operation_tags_filter()) dt_dev_invalidate_from_gui(darktable.develop);
 
     dt_accel_connect_locals_iop(module);
@@ -1905,6 +1902,9 @@ static gboolean _iop_plugin_header_button_press(GtkWidget *w, GdkEventButton *e,
 
       const gboolean collapse_others = !dt_conf_get_bool("darkroom/ui/single_module") != !(e->state & GDK_SHIFT_MASK);
       dt_iop_gui_set_expanded(module, !module->expanded, collapse_others);
+
+      //used to take focus away from module search text input box when module selected
+      gtk_widget_grab_focus(dt_ui_center(darktable.gui->ui));
 
       return TRUE;
     }


elstoc:
There is still a general issue with these global shortcuts in that they don't work the same as other keyboard shortcuts and rely on the centre image having focus. So if you move a slider, for example, keyboard focus is given to that slider and the darkroom view's key_pressed function isn't run unless you click on the image to give focus back.  

Global shortcuts might be better if they were coded independently as 'view' shortcuts in each of the individual views.

elstoc:
While, with #4891, this will no longer be an issue for zoom, it does still affect preview shortcuts so is still required.

ptilopteri:
@elstoc if #4891 is supposed to solve #4875, and it has been published, it is not solved for me.

elstoc:
It's supposed to resolve the only issue I could find for #4875 which was the one I reported in the comments. Have you tried remapping as I suggested? Obviously if your issue remains we can leave that open.  What do you mean by 'and it has been published'? Do you mean you've tested from my PR?

ptilopteri:
I only test master.  Do not know how to test *your* PR.

changed keyboardrc 
(gtk_accel_path "<Darktable>/global/zoom in" "<Primary>plus")
to
(gtk_accel_path "<Darktable>/global/zoom in" "<Primary><shift>plus")

and
(gtk_accel_path "<Darktable>/global/zoom out" "<Primary>minus")
to
(gtk_accel_path "<Darktable>/global/zoom out" "<Primary><shift>minus")

and <ctrl><shift>plus or minus also do not work.

but they do appear in the "HELP" window as global keys

elstoc:
> I only test master.

Then you won't be able to test my fix has worked until it's been merged.

> changed keyboardrc

Please you try changing it in the shortcuts tab on the preferences window instead of by hacking your keyboardrc file. That's the only way to be sure the mapping is correct for your keyboard layout For example, if you have to use the SHIFT key to get either plus or minus on your keyboard, then the shortcuts you have set will not work. I have a UK keyboard layout and I need to hold SHIFT in order to get a "+", so the correct shortcut for me is actually SHIFT+equals. For me the MINUS shortcut works fine because I don't need to use SHIFT to get it.

For me, if I set the shortcuts through the settings window they work fine, but only if I click on the image (in the darkroom) first. With the fixes in this PR and the other mentioned above, I don't need to click the image any more and they work all the time.

ptilopteri:
ok, opened an image, made top panel visible and selected the preference icon
changed global "zoom in" to <ctrl><shift><+> and
"zoom out" to <ctrl><shift><->
<ctrl><shift> and plus or minus neither work

changed back to previous default <ctrl><+> and <ctrl><->
and they do not work.
tks

ptilopteri:
darktable-3.1.0~git1389.0487939fc

ptilopteri:
darktable-3.1.0~git1390.cc18fbf01-2918.1
does not work either.

elstoc:
If neither of them work even after you've mouse-clicked on the image in the darkroom to give it focus then I don't know what the issue is - I am unable to reproduce it. You could try re-mapping to an alphanumeric character and see if that works but apart from that I don't know.

ptilopteri:
@elstoc and today it works, ???

ptilopteri:
@elstoc <quote>
Please you try changing it in the shortcuts tab on the preferences window instead of by hacking your keyboardrc file. That's the only way to be sure the mapping is correct for your keyboard layout For example, if you have to use the SHIFT key to get either plus or minus on your keyboard, then the shortcuts you have set will not work. I have a UK keyboard layout and I need to hold SHIFT in order to get a "+", so the correct shortcut for me is actually SHIFT+equals. For me the MINUS shortcut works fine because I don't need to use SHIFT to get it.
</quote>

of what use are the "keyboardrc" and "keyboardrc_default" if darktable does not properly interpret    them?  To my memory it has always worked in the past.

elstoc:
> @elstoc and today it works, ???

You're saying it works ok today? Great.

> of what use are the "keyboardrc" and "keyboardrc_default" if darktable does not properly interpret them?

keyboardrc (I don't know about keyboardrc_default) is correct and consistent if it's created by/from darktable, but if you modify it manually it's harder to be sure it properly represents the shortcuts you're typing. So I asked you to use the preferences dialog to be sure that the setting of the shortcut accurately matched the keyboard combination you used. I've now submitted a third PR (#4896) which should mean darktable now handles <plus> and <minus> correctly regardless of the modifiers required to access them, so the default settings should now work 'out of the box'.

ptilopteri:
@elstoc tks, currently working correctly/as expected for me.
do you want me to close #4875, or you can?
tks

elstoc:
@ptilopteri: Up to you, either you can close it yourself or this PR will close it, if it is merged.

elstoc:
It's worse than I thought. The previous resolution to #4717 seems to be giving keyboard focus to (a random) one of the child widgets in the clicked module, so this fix will also resolve that.

elstoc:
@TurboGit I think this is worth considering for 3.2 as it means that, after clicking the module header, the arrow shortcuts work to pan within the zoomed view in the darkroom instead of changing a random widget in the module. It also means that the preview shortcut should work correctly. Both of these are issues with my previous fix to #4717.

TurboGit:
I had a look and I don't like the resolution for this problem. It seems wrong to me to get focus to center image in the dt_iop_request_focus() routine to me. This routine has been designed to get focus for the iop. Maybe there is another way to fix ? Can you comment on this.

elstoc:
It would be great if there were a way for a widget to lose focus without giving it to something else but I can't find a way. The previous solution gave focus to the module header widget itself, which then passed focus to a random widget, so that was no better. As I've mentioned on a previous issue/PR, the shortcuts for showing the preview and moving around the image don't work unless the centre image has focus so I still believe this is the least-worst resolution, as it ensures that focus is _not_ given to the things that shouldn't have it (i.e. random widget or search modules box).

Perhaps if we instead moved this to the callback that runs when you click on the module header?

TurboGit:
> Perhaps if we instead moved this to the callback that runs when you click on the module header?

It would probably better. Note that I'm not against this PR just that the focus setting set to center image in a routine that is designed to give the focus to the widget itself seems misleading and could possible introduce other misunderstanding in the long run.

elstoc:
I agree. It's the wrong place for it. I'll move it to the header button press.

elstoc:
I notice that there's another call to ``gtk_widget_grab_focus(next->expander)`` in ``dt_iop_gui_delete_callback`` which could cause similar issues. I think this should probably be calling ``dt_iop_request_focus`` instead. Do you mind if I change that here as well or should I submit a separate PR?

TurboGit:
Here would be fine.

elstoc:
Done

TurboGit:
Fine with me now. Thanks.

PR #4886, opened 2020-05-01
Max thumbtable overlays size for HiDPI

Mark-64:
The purpose of this PR is to make the new thumbtable look nicer for HiDPI screens, also on Windows.
Currently, for HiDPi screens, the size of overlay icons and text is scaled up to a maximum value which is too big and make the overlays look bigger than other GUI icons. See the example below:

Cattura32

For LoDPI screens the scaling is correct and the overlays can grow equal to other icons, but not bigger.

The reason is that other GUI icons are inside buttons which always have an inner border. See buttons.c line 104
`float f_border = ((flags & CPF_DO_NOT_USE_BORDER) ? 4.0 : 6.0);`
Same in togglebuttons.c line 138.
Even for the main GUI buttons where CPF_DO_NOT_USE_BORDER is TRUE, there is a fixed inner margin in pixels which does not scale up with DPI.
This is one of the quirks of the Darktable GUI which would require rewrite, but changing that is far from being easy, as it affects tens of GUI elements distributed in the entire codebase.

Therefore this ad-hoc fix for thumbnail.c, where the max size `DT_PIXEL_APPLY_DPI(20.0)` is split in two parts `DT_PIXEL_APPLY_DPI(16.0) + 4.0` to follow the same logic of the other GUI elements.

The change doesn't affect LoDPI screens, while for HiDPI the effect is this 

Cattura31

### DIFF
--- a/src/dtgtk/thumbnail.c
+++ b/src/dtgtk/thumbnail.c
@@ -21,6 +21,7 @@
 #include "bauhaus/bauhaus.h"
 #include "common/collection.h"
 #include "common/debug.h"
+#include "common/focus.h"
 #include "common/focus_peaking.h"
 #include "common/grouping.h"
 #include "common/image_cache.h"
@@ -33,6 +34,8 @@
 #include "dtgtk/thumbnail_btn.h"
 #include "gui/drag_and_drop.h"
 #include "views/view.h"
+
+static void _thumb_resize_overlays(dt_thumbnail_t *thumb);
 
 static void _set_flag(GtkWidget *w, GtkStateFlags flag, gboolean over)
 {
@@ -62,14 +65,11 @@ static void _thumb_update_extended_infos_line(dt_thumbnail_t *thumb)
   vp->imgid = thumb->imgid;
   vp->sequence = 0;
 
-  gchar *msg = dt_variables_expand(vp, pattern, TRUE);
+  if(thumb->info_line) g_free(thumb->info_line);
+  thumb->info_line = dt_variables_expand(vp, pattern, TRUE);
 
   dt_variables_params_destroy(vp);
 
-  // we change the label
-  g_snprintf(thumb->info_line, sizeof(thumb->info_line), "%s", msg);
-
-  g_free(msg);
   g_free(pattern);
 }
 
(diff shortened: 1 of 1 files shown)


TurboGit:
I'll wait for the big culling rewrite to have a look at that. It may create conflicts.

Nilvus:
I agree, margins should be from CSS only, and of course having same margins to set to extension name and group/altered/audio... icons. Having that on CSS would guarantee to have same display regardless of the user/screen/dpi.

So feel free to give me more work ;-)

Mark-64:
@AlicVB 
Margin of buttons is controlled in two places 
`static gboolean _button_draw()` in button.c
`static gboolean _togglebutton_draw()` in togglebutton.c
In both there is this line `float f_border = ((flags & CPF_DO_NOT_USE_BORDER) ? 4.0 : 6.0);`
So when flag `CPF_DO_NOT_USE_BORDER` is false margin is 6 pixels and when is true is 4 pixels (shouldn't this be zero !?). Both numbers are non DPI aware.
Athough like you I didn't find where the size of buttons is defined, I've done many tests on low and high DPI systems and I got evidence that what is DPI scaled is not the total size of the button, but rather the icon size = total size - 2 * inner margin.

Now, with the new thumbtable you introduced a new class of buttons in thumbnail_btn.c, where the inner margin is set in % of the total size.

```    
if(flags & CPF_DO_NOT_USE_BORDER)
    {
      DTGTK_THUMBNAIL_BTN(widget)->icon(cr, 0, 0, allocation.width, allocation.height, flags, bg_color);
    }
    else
    {
      DTGTK_THUMBNAIL_BTN(widget)->icon(cr, 0.125 * allocation.width, 0.125 * allocation.height,
                                        0.75 * allocation.width, 0.75 * allocation.height, flags, bg_color);
```
So when `CPF_DO_NOT_USE_BORDER` is true, margin is zero. This creates the discrepancy.

IMO, you logic is correct and it should be applied to all buttons in order to have a truly resolution independent GUI.
Problem is, if you try to touch the two hardcoded margins of 4 and 6 pixels, the entire GUI is screwed.
I tried all possible combinations and I realized there is no simple solution. A large rewrite of the button logic is needed (and BTW my confidence with the DT codebase is not yet at the level required to undertake such a task).
Plus, the size of several Cairo icons in paint.c is adjusted ad-hoc and it should be redefined.

For example, if I add the `DT_PIXEL_APPLY_DPI` to the margins, as you propose, the size of many icons becomes way too small for HiDPI and the discrepancy with thumbnail icons become even bigger.

Cattura34
Cattura33

I suggest to accept this change for now, in sight of a better fix.
It has no adverse effects, as it only slightly changes the maximum size of the thumbnail icons for HiDPI.

Nilvus:
@Mark-64: I think the better fix is to set that with CSS. Some buttons you displayed have already margins or padding set for them.

I suggest to wait a little. I will see what's already possible here and what is missing to set them on CSS.

Mark-64:
Will propose a new PR based on culling rewrite

Nilvus:
> Will propose a new PR based on culling rewrite

For your new PR, I would suggest you to also test your work with my actual draft PR as I change some things about thumbnails icons also. See #4959

PR #4891, opened 2020-05-01
keyboard shortcuts: global zoom and module search

elstoc:
Global zoom (in/out) previously wasn't working when a slider or combobox had keyboard
focus (as they always do when they are changed).

These have been changed to be normal per-view shorcuts with the same
defaults and so are no longer dependent on keyboard focus. For darkroom
this also has the effect of placing all zoom shortcuts in the same
section in preferences.

Module search has been changed to also be consistent with other per-view
shortcuts (though it wasn't exhibiting the same issues).

### DIFF
--- a/src/control/control.c
+++ b/src/control/control.c
@@ -669,13 +669,6 @@ int dt_control_key_pressed_override(guint key, guint state)
     darktable.develop->darkroom_skip_mouse_events = TRUE;
     return 1;
   }
-  // set focus to the search module text box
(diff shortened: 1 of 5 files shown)


elstoc:
It would be nice to do this with the preview shortcut too but I'm not sure it's possible without changing the behaviour to a more simple on/off shortcut (so that we don't have to handle the key_released behaviour as well).

TurboGit:
Looks good to me. Thanks!

PR #4896, opened 2020-05-02
keyboard shortcuts: correct handling of shift modifier

### LINKED ISSUE #4894: shift modifier for non-alphanumeric keyboard shortcuts
Different keyboard layouts require different key combinations to access non-alphanumeric characters. For example, on my UK keyboard, the plus sign requires a shift modifier to access it (``shift+equals``), the asterisk character is ``shift+8``. Some default keyboard shortcuts have non-alphanumeric characters hard-coded into them (for example global zoom hard-codes ``<primary>+plus`` and ``<primary>+minus``).

However, whenever the shift modifier is required to access a non-alphanumeric character, its shortcut is displayed and interpreted incorrectly in darktable. In the global zoom example above, darktable's hard-coded default of ``<primary>+plus`` does not work for me and needs to be remapped in the shortcuts window, in which it then appears as ``Shift+Ctrl++``.

Gtk actually provides a mechanism to mask out the shift modifier in these circumstances (https://developer.gnome.org/gdk3/stable/gdk3-Keyboard-Handling.html#gdk-keymap-translate-keyboard-state). Darktable should be modified to use this.

elstoc:
Where the shift modifier is required in order to access a
non-alphanumeric character, shift should not be treated as a modifier
key. For example, when using <plus> for global zoom on a UK keyboard
this incorrectly appears in the preferences window as ``<shift>+<plus>`` and
the default value of ``<primary>+<plus>`` does not function correctly.

Replace all references to ``event->state`` with call to ``dt_gui_translated_key_state``
to ignore this already-consumed modifier for keyboard shortcuts.

Resolves #4894

### DIFF
--- a/src/gui/gtk.c
+++ b/src/gui/gtk.c
@@ -1027,19 +1027,33 @@ static gboolean window_configure(GtkWidget *da, GdkEvent *event, gpointer user_d
   return FALSE;
 }
 
+guint dt_gui_translated_key_state(GdkEventKey *event)
+{
+  if (gdk_keyval_to_lower(event->keyval) == gdk_keyval_to_upper(event->keyval) )
+  {
+    //not an alphabetic character
+    //find any modifiers consumed to produce keyval
+    guint consumed;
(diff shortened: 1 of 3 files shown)


TurboGit:
Works for me! Thanks.

Issue #4869, opened 2020-04-30
DT 3.0.2 - Crop and Rotate module can get into a non-functional state


pahl34:
Version: darktable v3.0.2
OS: Windows 10

It can happen that the **crop and rotate** module gets into a state where it is not properly usable any more. In this case, neither the direct cropping in the image using the control overlay (main tab), nor the margins tab will work. Also, neither the control overlay is displayed in this state, nor is the selected guide (e.g. "rules of thirds", which normally enables dashed lines being shown as an overlay to the image).
The only thing that still works is the angle slider. After using it to set the rotation to another value, the cropping also is enabled again.

Disabling and enabling the module completely does not resolve the problem. Collapsing and then extending the module also usually does not resolve it either - however it was observed that seemingly after longer waiting times in between the collapsing and extending, this sometimes resolves the issue.

Unfortunately, even after playing around for half an hour or so, I could not figure out a way to safely reproduce this behavior. In fact, for me it was only reproducible on one specific group of images (a RAW file and its duplicates). For other files I could not reproduce it, using the same steps, so I cannot rule out that some specific edit I did on the problematic image might have caused this.

**Side note:** I noticed that in general if another duplicate of the same image is loaded while the crop and rotate module is already extended in the right panel, then the controls will not appear after loading of the image has finished, but it is always required to collapse and then extend the crop and rotate module. This might be a separate issue on its own (not sure whether this is 'as designed' behavior) but it might also be related to the issue described here.

Due to failing to find the exact steps to reproduce, I can only try to describe what I did as precisely as possible and hope that somebody with more background knowledge (e.g. knowing the code behind this module) will be able to figure it out.

1. RAW file with two duplicates (thus, there exist images A (original), B and C (duplicates)) - various edits (on all three) had already been done in an earlier session. Among other edits, image A had a change of roation done using the crop and rotate module.
2. Created a third 'virgin' duplicate D.
3. Deactivated the Base Curve on D
4. Applied the style "Filmic Natural (dt 3.0+)" to D, downloaded from [https://dtstyle.net/](https://dtstyle.net/).
5. In lighttable, copied (only) the crop and rotate setting mentioned in 1. from A to D:
5.1 Select A
5.2 In the history stack module, "Copy" --> Deselect all --> Select only 'crop and rotate'
5.3 Select D
5.4 History stack module, "Paste" --> Ok
6. Double click D to load it inot darkroom.
7. The error was present - crop and rotate would not work (as described above).

[...]

PR #4901, opened 2020-05-03
Fixing module in focus when changing image while in darkroom (#4869)

jenshannoschwalm:
We can only have one module "in focus", this state is important as it makes sure
the module's specific expose code is taken. Examples are crop&rotate, vignette.

If we change the image in darkroom by keypresses (space or bs) or via the filmstrip
this module-in-focus state is currently lost.

Why? There is an attempt to gain focus again but this is done while blocking signals
and `dt_iop_request_focus` does not work this way.


TurboGit:
Works for me. Thanks!

PR #4906, opened 2020-05-03
German translation updates

rabauke:
I added a few missing translations, and cleaned-up some others. 

Due to `intltool-update` the diff is unreasonably large and actual changes are difficult do grasp. Following https://stackoverflow.com/a/11291001, I added to `.git/info/attributes`

    *.po    diff=po

and to `.git/config`

    [diff "po"]
    textconv=msgcat --no-location --no-wrap --sort-output

which allows me to inspect the actual changes via `git diff master my_translation_de`.


TurboGit:
Thanks !

PR #4908, opened 2020-05-04
Do not mention "semi-log" scales in the manual

jpellegrini:
Commit c6b446f by Philippe Weyland on 19 Jun 2019 removed the semi-log scales from the tone curve. This patch changes the manual in order to reflect that.

It also changes "double logarithmic" to "logarithmic" in the base curve section of the manual (since there is no semi-log in any of darktable graphs, the word "double" could perhaps be confusing to
users.


TurboGit:
Thanks !

PR #4914, opened 2020-05-04
po/fix selected

TurboGit:


AlicVB:
@TurboGit : well, not sure it's the right fix, as the idea with my thumbtable rewrite is to keep the selection intact even when changing view.
If I understand well @aurelienpierre issue, we just need to ensure that `plugins/lighttable/recentcollect/pos0` is correctly set when entering darkroom (and not just on exiting)

I'll try to have a look tomorrow if you want

TurboGit:
@AlicVB : I'm not sure, Aurélien did talk about the image not being selected and I double checked that when changing the picture via the filmstrip in the darkroom and print module and quitting dt with this patch when you restart the last viewed (selected) image is the one selected on the lighttable.

Maybe this is not the right way to fix that... But in that case I'm not sure what behavior should be implemented? I'll be glad to let you have a look at this if you have some time today!

TurboGit:
@AlicVB : see also #4920 

Looks like many people are expecting this behavior. What is the current design/behavior that you implemented ? I'm not against it, just trying to understand the way selection is kept.

AlicVB:
The idea is to never screw user selection in his back. The user should do voluntary (un)selection actions. This is not a problem in filemanager.
Now the change is for modes with filmstrip, especially darkroom.
- I've replaced the "click to select" by alt-click so the user can change the current active image without loosing it's selection (ctrl-click and shift-click stay unmodified)
- For the "image to act on" part, the idea is in comment in the code of `dt_view_get_images_to _act_on` (active image are the image(s) shown in center view) : 
```
/** Here's how it works
   *
   *             mouse over| x | x | x |   |   |
   *     mouse inside table| x | x |   |   |   |
   * mouse inside selection| x |   |   |   |   |
   *          active images| ? | ? | x |   | x |
   *                       |   |   |   |   |   |
   *                       | S | O | O | S | A |
   *  S = selection ; O = mouseover ; A = active images
   *  the mouse can be outside thumbtable in case of filmstrip + mouse in center widget
   **/
```
Now, reading the bug report, there's some problems with entering/leaving view. That shouldn't change the selection.
As said, I'll have a look, but I just need to find some free time. My job drives me crazy this days ;)

AlicVB:
That said, I think we should add another point like what is done currently for preview and culling : 
if we have only one image selected and this is the image opened in darkroom, the selection should follow the active image, because it's not a "real" selection in term of "carefully selected list of images" 
As it's already how works culling and preview (with and without the rewrite) it shouldn't be too hard to implement.

TurboGit:
Closing this one, better fix propose by @AlicVB in #4924

PR #4915, opened 2020-05-04
snapshot: stop indicator triangle disappearing when zoomed

### LINKED ISSUE #4900: snapshot triangle indicator doesn't appear when zoomed
When activating a snapshot in the snapshot module, a new triangle indicator was recently added to make it clear which side of the image was the snapshot and which the current edit.

While this works fine when the image is zoomed out:

Screenshot_2020-05-02_22-08-40

It doesn't appear when zoomed in:

Screenshot_2020-05-02_22-09-06

If I then zoom out while the zoomed-in snapshot is still enabled, it pops back into view...

Screenshot_2020-05-02_22-13-03

elstoc:
stops the snapshot indicator triangle disappearing off the edge of the
screen when zoomed in

Resolves #4900


TurboGit:
Thanks for fixing that !

PR #4918, opened 2020-05-05
Updates to pt_BR

jpellegrini:
We're getting close to a fully translated manual! (But I'll still need to review the translation later)


TurboGit:
Thanks !

PR #4920, opened 2020-05-05
keyboard shortcuts: minor reorg

elstoc:
A number of keyboard shortcuts are hidden away in the preferences section under "modules>module name" but it is either not obvious to the user that they are "modules" or they generally would sit more logically under either "views" or "global". This PR addresses that issue and hopefully makes those keyboard shortcuts more logical and discoverable.

The following lighttable shortcuts (for zoom and culling) are currently hidden away in the modules>lighttable shortcuts section. This change moves all of them to the lighttable view section.

- The zoom in/out shortcuts were duplicates of existing zoom in/out shortcuts and have been removed. Existing shortcuts have default keys of ``CTRL+`` and ``CTRL-``. Removed shortcuts had default keys of ``ALT-2`` and ``ALT-3``.
- The zoom max/min default shortcuts have been changed to use similar keys (``ALT+``/``ALT-``) as the existing zoom shortcuts. These replace previous shortcuts of ``ALT-1`` and ``ALT-4``.
- Three existing culling mode shortcuts have been modified so that they
now display in the lighttable view shortcut section.

The shortcut "hide navigation thumbnail" has been moved from "modules>navigation" to the darkroom view.

The shortcut "hide histogram" (as well as some other recently added histogram shortcuts) has been moved from "modules>histogram" to the darkroom and tethering views.

The contents of "modules/preferences" have all been moved to the global shortcut section. All thumbnail overlay options have been placed under "global/thumbnail overlays".


AlicVB:
I'll have a look, but I would prefer if we can postponed this after the culling/preview rewrite to avoid too many conflicts. Lot of change have happen in this area !
(sorry for the delay, I'm quite short in time atm)

elstoc:
That's fine - it's mostly just tidying up so that shortcuts are a bit more discoverable. I have a couple of other shortcut movements I want to make using the functionality in this PR (darktable/tethering - hide histogram and darktable - hide navigation thumbnail) so I might go ahead and include them here as well if you want me to wait before merging.

FYI I had also planned to implement something similar to move the 'modules>preferences' section into global (which contains options for the lighttable overlays). I might also include that here as well.

elstoc:
This is ready for review now

TurboGit:
This needs to be rebased. Thanks!

src/views/lighttable.c
```diff
@@ -780,6 +780,20 @@ static gboolean zoom_out_callback(GtkAccelGroup *accel_group, GObject *accelerat
...
+                                           GdkModifierType modifier, gpointer data)
+{
+  dt_view_lighttable_set_zoom(darktable.view_manager, 1);
+  return TRUE;
+}
+
+static gboolean zoom_min_callback(GtkAccelGroup *accel_group, GObject *acceleratable, guint keyval,
+                                           GdkModifierType modifier, gpointer data)
+{
+  dt_view_lighttable_set_zoom(darktable.view_manager, 2 * DT_LIGHTTABLE_MAX_ZOOM);
```
TurboGit, line 793: The max zoom is 25, why 2 times it ? The control at the bottom of the lighttable show max at 25.

And indeed ALT+ - set 50 thumbs per row.

elstoc: I agree. That wasn't introduced by this change but I'm happy to fix it.

TurboGit:
All good now. Thanks.

TurboGit:
@elstoc : just found an issue with this, see : #5280 
Can you have a look ? Thanks.

Issue #4177, opened 2020-01-23
Change documentation format from docbook to markdown


paperdigits:
**Is your feature request related to a problem? Please describe.**

Contributing to the documentation is difficult, the documentation of the documentation is not good. We're behind on the docs release for 3.0. DocBook is difficult and we're not even using DocBook, but our own custom DTD. We have dependencies for this tooling. The docs are hosted on gitlab currently which is confusing as well.

**Describe the solution you'd like**

I will convert all the docs to markdown, so they can be built with the rest of DT.org website.

**Alternatives**

Keep doing what we're doing?

**Additional context**

Spoke with @aurelienpierre  about this in IRC

I'd like to gain consensus from the core team members before I do the work, as this will be a large undertaking.

Nilvus:
That will probably be a huge work but I like the idea !

paperdigits:
I'm willing to do the work, no problem.

AlicVB:
Great !
Will this apply to lua doc too ? iirc, lua doc is generated automatically, but I've no idea how it's done !

Anyway, thanks and good luck for all the work !

aurelienpierre:
@upegelow, since you have maintained the usermanual for the longest time, are you ok with this ?

The goal would be to simply have one markdown file per manual page, laid out in sub-directories, then we thought about using Pandoc to compile the collection of .md as a HTML/PDF/ePub file ([example](https://pandoc.org/epub.html)), then update the website by simply sending the HTML on the server storage.

The benefit is markdown files can be edited directly from the Github UI in an "almost WYSIWYG" fashion. Then, it's very little admin overhead to update all targets.

TurboGit:
I'm all for it as the current build of the documentation is far from trivial and this certainly delay the release of the documentation.

@upegelow : what's your feeling about this? You are probably the only one to be able to build it.

upegelow:
I don't know markdown but I have never been a big fan of docbook (to say the least). It always sucked in generating the HTML version of the manual. Right now even I cannot build it as it requires saxon which is no longer properly detected. Probably this can be fixed but I had no time yet.

The only big strength of docbook for me has been its simple design. You can easily check the structural correctness with xmllint. But maybe markdown has the same advantage then it's fine.

Concerning file structure of the manual. That's just historical. No need to stick to it.

TurboGit:
[...]

PR #4921, opened 2020-05-06
Migrate the docbook content to dita xml.

paperdigits:
First part of migrating to markdown is porting to a neutral format, dita xml. We can then transform the dita xml into markdown. Beginning of #4177


aurelienpierre:
@paperdigits awesome ! Could you provide a minimal set of commands/dependencies to build it ? Thanks !

paperdigits:
@aurelienpierre do the following:

1. Install a recent version of openjdk
2. Download and unzip the dita open toolkit: https://github.com/dita-ot/dita-ot/releases/tag/3.4.1
3. `/path/to/dita-open-toolkit/bin/dita -f pdf2 -i /path/to/darktable/doc/usermanual/darktable-user-guide.ditamap`

Or you can use this docker file: https://github.com/dita-community/dita-ot-project-docker

paperdigits:
@TurboGit any thoughts?

TurboGit:
I need to find time to look at this. The format seems cleaner and less convoluted. That's already a good point.

paperdigits:
@TurboGit Yes, the markup of DITA is more minimalist when compared to docbook. I use DITA at work, so I know it quite well. A few other benefits are: document structure is separate from content, publishing pipeline is plugin based and batteries included (only requires a recent java version), should be able to keep using the same translation workflow, though the actual translation segments will likely be effected, can easily generate PDF and HTML with the HTML possibly being able to be included on the main website instead of a separate site.

TurboGit:
>     3\. /path/to/dita-open-toolkit/bin/dita -f pdf2 -i /path/to/darktable/doc/usermanual/darktable-user-guide.ditamap

The command is actually:

`/path/to/dita-open-toolkit/bin/dita -f pdf2 -i /path/to/darktable/doc/usermanual/darktable_user_manual.ditamap`

Right ?

I have a pdf now. Fact is it looks really nice except for the pictures where most of them are too big and are partly rendered outside the PDF pages. For example p.14 the overview,  p.20 the lighttable picture, p.21 the thumbs is over the text and partly outside the page... Is that an issue on my side ?

And I still hope you'll come up with a proper workflow to put in place for the translators. This is still not clear to me.

paperdigits:
Yes, you are correct, underscores not dashes. 

For the time being, they can use the same workflow with dita, as it is XML.

paperdigits:
> For example p.14 the overview, p.20 the lighttable picture, p.21 the thumbs is over the text and partly outside the page... Is that an issue on my side ?

No, a bit of refinement is still needed, either in a dita plugin or in the source itself.

TurboGit:
> For the time being, they can use the same workflow with dita, as it is XML.

Which is ? Is creating a POT file out of dita XML already supported ? What about the current translated .po files ? We need to find a proper way to upgrade to this new XML for this to be mostly transparent to translators. I have never looked at this part in darktable, but I suppose we may have to update some scripts or CMakeList.txt.... How is the current POT file generated ? Did you had a look at this part ?

parafin:
Is Markdown still a target format? Does it make sense to adapt translators workflow to dita just for it to get updated again after transition is complete?

paperdigits:
I think the only semi-seamless way for translators would be to generate a translation memory for the docbook sources, then import the dita XML sources. There shouldn't be a lot of segment change from docbook to dita,  but there will be some.

PO is brutal because it is a 1:1 mapping of source -> PO file. If we only use PO, I think a lot of stuff will have to be retranslated.

I'm looking at what FOSS tools will support a translation memory.

paperdigits:
Markdown is a target, so far as I'm aware, though I'd be just as happy to keep it in dita, personally. 

Markdown's translation workflow is not as well defined as XML, and in the mean time, there are incoming updates to the documentation, which create a larger delta between the labor intensive conversion and the existing docbook docs. So in dita people can continue to update the docs while I organize the files a bit better and define the markdown workflow.

paperdigits:
I am going to start migrating the content from the conflicts into DITA.

paperdigits:
I've ported over all the changes that have been merged into the docbook since I started.

paperdigits:
I've updated all the PO files from the dita sources. You can generate the pot file using this command:

`xml2po -o po/darktable_user_manual.pot darktable_user_manual.ditamap $(find . -name "*.dita")`

For now, the rest of the translation workflow can stay the same. 

I need to try and update the cmake file, but I don't really know how. I'll ask for help in IRC.

I've also generated translation memory files which were made from the UI strings and the previous dockbook translated PO files. I've included them in the repo.

@TurboGit  after the cmake file is updated, if this isn't good enough, I give up :P

paperdigits:
@aurelienpierre @TurboGit can anyone help with the cmake? It is a bit out of my league.

paperdigits:
The cmake no longer works, it searches for XML files. We need to search for `.ditamap` and `.dita` files. The PO files can be updated using the command above, it just needs to be put in the cmake.

elstoc:
I'd like to start looking at making some changes to the user manual, but have been waiting for this PR to be merged. Much as I hate to say it, if nobody is able/willing to help @paperdigits with the cmake stuff to get this over the line, this PR is going to be the thing that _stops_ the manual being updated rather than the thing that makes it easier.

I know people are busy on the release but it would be good to at least know whether this PR has a future.

TurboGit:
@elstoc : You help on this is welcome. I have no time to look at this at the moment and no expertise on CMake.

But note that the blocking point is that we still don't have a clear idea as to what we will be able to propose to the translators. I don't want to move the documentation and break all translations because there is no easy way to let people work on them.

So before doing anything I will advise again to step back and propose a solution that will smoothly move us from current documentation format to whatever else without breaking any current feature. This means:
- proper screenshot handling (see my comment about that)
- simple way to create HTML pages
- properly generate the top index referencing all the manual in supported languages
- toolset for translators to work on the translation of the manual

Maybe I have forgotten some points?

elstoc:
I'm happy to provide what assistance I can but I also don't know cmake so I can't help with that. My impression was that @paperdigits had a way forward for translation but that the cmake work was the blocker at the moment but perhaps I'm mistaken.

PR #4924, opened 2020-05-06
darkroom : keep selection in sync if unique

### LINKED ISSUE #4916: selection of current image is lost between lighttable and darkroom
Selection has lost consistency between lighttable and darkroom.
To reproduce: 
Select an image in lighttable. 
Enter darkroom mode ("D").
Select other image with single click (I haven't checked for double click, maybe too).
Return to lighttable with "L".

The selection returns to the image, selected in lighttable, not in darkroom.  

That is a dataloss issue - the actions are applied to the wrong image (user is sure the selected one is the last one selected).

Darktable from git master of 2020.05.05.
Linux Mint 19.3 Mate x64.
OpenCL is active,  Radeon RX580.

AlicVB:
this fix #4916

### DIFF
--- a/src/views/darkroom.c
+++ b/src/views/darkroom.c
@@ -23,13 +23,14 @@
 #include "common/debug.h"
 #include "common/file_location.h"
 #include "common/focus_peaking.h"
+#include "common/history.h"
 #include "common/image_cache.h"
 #include "common/imageio.h"
 #include "common/imageio_module.h"
+#include "common/selection.h"
 #include "common/styles.h"
 #include "common/tags.h"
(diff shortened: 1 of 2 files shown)


TurboGit:
Don't we need the same code for the print module ?

AlicVB:
@TurboGit : I don't really know the print module as my printer in b&w only with very low resolution... but I think it makes sens. I'll take care

AlicVB:
@TurboGit : done. I've also fixed an inconsistency between print and other view in the last commit can you double check ?

TurboGit:
Works for me, thanks !

PR #4925, opened 2020-05-06
Introduce dt_color_picker_new minimalist color picker setup

dterrahe:
This introduces a new call dt_color_picker_new which wraps setting up the color_picker structure and linking it to a slider or creating a button and boxing it. Since it passes the color_picker as a parameter in the callback, it can support an unlimited number of pickers per module without needing to maintain a color_picker object in the gui data or having to define an enum to refer to each of the pickers; the picker widget is directly checked in the new overloaded color_picker_apply method. The get-set and update calls are no longer necessary, even for multi-picker.

This is backward compatible with the existing approach, so I have only transitioned a few modules to demonstrate how it works and will do the others at the same time as rolling out #4804 across all modules.

I am also looking at refactoring out the commonalities of colorzones, rgbcurve and blending double-color-pick buttons.

After everything has transitioned and unless there is a desire to maintain compatibility with out-of-tree modules, I'd like to go over color_picker_proxy once again and simplify a lot of complexity that doesn't seem like it would be necessary anymore. I will likely be asking questions on IRC to understand what the original purpose of some of it was. For example, why are the picker point and area stored in dt_iop_module_t, rather than directly in the color picker or even globally? It might require too many changes throughout the modules to change that now to be worth the hassle, but at least I want to make sure we don't have the intention to support more than one color picker at the same time?

BTW negadoctor::_iop_color_picker_update shows how hard it is to keep all that duplicate boiler plate code in sync :-)

@TurboGit @edgardoh

### DIFF
--- a/src/develop/imageop.c
+++ b/src/develop/imageop.c
@@ -256,6 +256,8 @@ int dt_iop_load_module_so(void *m, const char *libname, const char *op)
   if(!g_module_symbol(module->module, "gui_init", (gpointer) & (module->gui_init))) module->gui_init = NULL;
   if(!g_module_symbol(module->module, "gui_update", (gpointer) & (module->gui_update)))
     module->gui_update = NULL;
+  if(!g_module_symbol(module->module, "color_picker_apply", (gpointer) & (module->color_picker_apply)))
+    module->color_picker_apply = NULL;
   if(!g_module_symbol(module->module, "gui_cleanup", (gpointer) & (module->gui_cleanup)))
     module->gui_cleanup = default_gui_cleanup;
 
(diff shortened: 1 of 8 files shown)


TurboGit:
If this really works in all cases, I'm all for it as indeed it simplifies the code. Will test.

dterrahe:
> If this really works in all cases, I'm all for it as indeed it simplifies the code.

I believe that eventually the dt_iop_color_picker_t can be made much simpler; containing just a link to the module, the widget, the current point/area, and the callback function. And maybe the control and shift key status when the button was clicked. I haven't seen anything yet that really justifies the separate treatment for module and blender pickers. But once that distinction is made, there's a lot of stuff that needs to be tracked...

This initial PR just puts a small layer over the existing approach that will allow replacing all the multi-widget color pickers with single instance ones without disturbing anything. Once that is done and nothing refers to multi-instance anymore, I can simplify everything and clean up blenders and at the same time make sure that indeed nothing breaks. That would then just touch color_picker_proxy and blend_gui.

There is only one case, besides the ctrl/shift stuff in colorzones and rgbcurve and blend, where the get_set callback does anything else besides just saving the picker number; in rgblevels.c. Worst case that could be dealt with by attaching a second callback to the colorpicker buttons. But to be honest, I'm not sure bt_select_region is working 100% correctly now, so I would like to study that a bit more anyway.

TurboGit:
I did a first test tonight, all seems to be working ok ! I'll have another look tomorrow.

TurboGit:
Ok, this works for me. Really nice clean-up @dterrahe  !

PR #4927, opened 2020-05-06
Make module in focus slightly outstanding

jenshannoschwalm:
Especially when working with more than one module in expanded state we might want
to know by vision what module currently 'has the focus'.

This pr adds this feature as a UI improvement. It does not use any non-css coloring, only makes
the on/off button slightly outstanding.
Bildschirmfoto von 2020-05-06 18-28-33

Modules in focus will be shown like this 
Bildschirmfoto von 2020-05-06 18-34-15
 
Bildschirmfoto von 2020-05-06 18-28-37

Bildschirmfoto von 2020-05-06 18-28-41

module not in focus will not change.

### DIFF
--- a/src/develop/imageop.c
+++ b/src/develop/imageop.c
@@ -1998,12 +1998,12 @@ GtkWidget *dt_iop_gui_get_expander(dt_iop_module_t *module)
   /* add enabled button */
   if(module->enabled && module->default_enabled && module->hide_enable_button)
   {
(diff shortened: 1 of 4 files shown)


TurboGit:
I like that, will test.

TurboGit:
Thanks !

aurelienpierre:
Would it be possible to have something more discreet for the "enabled + in-focus" state ?

jenshannoschwalm:
Sure. A thinner extra circle?

TurboGit:
I agree with @aurelienpierre , after using it more and more, it stands out a bit too much.

Maybe an **half circle**, thinner at bottom to have a kind of raise-effect ?

PR #4928, opened 2020-05-06
Correct problem with click in color picker and mouse move causing slider to jump right

dterrahe:
A click+move ("drag") on color picker would cause the slider to move to max position.

Also, if there was no color picker, clicking right of the slider would cause it to move to max as well. This PR changes that behavior, but it is possible that some actually prefer it (even if it is less correct).

There is a small area along the color picker that is not ignored in dt_bauhaus_slider_motion_notify because I am not subtracting INNER_PADDING. This is intentional, because otherwise it is impossible to drag the slider all the way to the right and see it take on the max value. It would only actually show the max when releasing the mouse, which could be very frustrating. With the small overlap, moving it slowly over the right edge will allow the slider to attain max, without it being liable to jump right on an unintentional move while clicking in the middle of the color picker.

This (hopefully) solves a bug highlighted on IRC by @AxelG


AxelG-DE:
Tested
Works for me.
THANKS

Issue #4909, opened 2020-05-04
LUA: erratic behaviour of sliders after commit 36d459f


Mark-64:
After commit 36d459f2698695b9245e55cbc516698ab1eb3074, the behaviour of slua slider has become unpredictable, with soft_min and soft_max limits do not respect values set in the code, but are erratically set.
To try to reproduce, take the following script, which defines a simple LT module with one test slider.
[slider.lua.txt](https://github.com/darktable-org/darktable/files/4574804/slider.lua.txt)

Run the script several times, playing with the slider.

Before the mentioned commit, at DT start the result is consistently the following, as expected
![Cattura2](https://user-images.githubusercontent.com/43290988/80969282-a6ae7880-8e19-11ea-8ddc-7e746f256a83.JPG)

after the commit, sometimes I get this
![Cattura](https://user-images.githubusercontent.com/43290988/80969357-be85fc80-8e19-11ea-86d6-d8306b326bb2.JPG)

DT git master on Windows 10.

Mark-64:
I did some investigation and it's pretty weird. I might be wrong, need some help here, but it could be a sort of race condition.
In lua/widget/slider.c, the code
```
int dt_lua_init_widget_slider(lua_State* L)
{
  dt_lua_init_widget_type(L,&slider_type,lua_slider,DT_BAUHAUS_WIDGET_TYPE);

  lua_pushcfunction(L, tostring_member);
  dt_lua_gtk_wrap(L);
  dt_lua_type_setmetafield(L, lua_slider, "__tostring");
  lua_pushcfunction(L,digits_member);
  dt_lua_gtk_wrap(L);
  dt_lua_type_register(L, lua_slider, "digits");
  lua_pushcfunction(L,step_member);
  dt_lua_gtk_wrap(L);
  dt_lua_type_register(L, lua_slider, "step");
  lua_pushcfunction(L,hard_min_member);
  dt_lua_gtk_wrap(L);
  dt_lua_type_register(L, lua_slider, "hard_min");
  lua_pushcfunction(L,hard_max_member);
  dt_lua_gtk_wrap(L);
  dt_lua_type_register(L, lua_slider, "hard_max");
  lua_pushcfunction(L,soft_min_member);
  dt_lua_gtk_wrap(L);
  dt_lua_type_register(L, lua_slider, "soft_min");
  lua_pushcfunction(L,soft_max_member);
  dt_lua_gtk_wrap(L);
[...]

PR #4930, opened 2020-05-07
fix initialization of lua sliders

Mark-64:
Fix for #4909 
Change in default initialization values for lua sliders, as suggested by @dterrahe (thanks !)


TurboGit:
I cannot test that, but code looks ok to me. Thanks.

Issue #4929, opened 2020-05-06
problem updating lighttable


LViatour:
@alicvb

**Describe the bug**

when creating an HDR in the light table, the HDR is not visible. You have to change the display order to see it in the lighttable

**To Reproduce**

1. select multiple raw
2. clic to create HDR
3. the thumbnail of the DNG created is not visible in the light table

**Expected behavior**
Normally the new DNG HDR should appear in front of the raw files selected for HDR

**Platform (please complete the following information):**
-3.1.0~git1412.d999cc8a7
-Ubuntu 20.04 
-open CL yes
-Nvidia drive 440.64
-->

Mark-64:
It seems similar to #4889, I will have a look.

jenshannoschwalm:
I think creating the hdr file just "creates" it. This does not imply to **import** it ...

Mark-64:
Just forcing a refresh in thumbtable, for example changing the visualization order or group/ungroup, makes the DNG appear, so definitely the HDR is imported, only the view is not refreshed, like in #4889

Mark-64:
PR created

Mark-64:
Fixed, via b26f24717c63c0271151346d73973e171daecee5, please close

PR #4932, opened 2020-05-07
force thumbtable refresh when creating HDR

Mark-64:
Fix for #4929, similar to #4904


Mark-64:
Travis CI failed, reason being the use of an uninitialized variable in case HDR creation is aborted.
This is easy to fix, however in that case DT also segfaults, even without this PR.
I will have to investigate more

Mark-64:
The segfault is a more general issue, not related to this part of the code, I will open a specific bug report

Mark-64:
By the way, I could not replicate the segfault. It happens whe you lauch whatever job queue, for example exporting several images and then you click twice on the progress bar abort "X".
But it happens only with the Debug build, never with the Release.
It doesn't seem to be a real issue and in any case nothig related to this PR

TurboGit:
Looks good to me. Thanks!

PR #4939, opened 2020-05-07
Extend a color picker area by picking up a corner

dterrahe:
When selecting an area for a color picker, the first mouse click always sets one corner of the area. So after releasing the button, it is not possible to just extend the existing selection a little bit; you have to go back to the same starting point again.

This change allows to extend an existing selection by clicking sufficiently close to one of its corners (probably where the button was just released) and continue to have the area extend to the existing opposite corner (probably where the last selection started).

### DIFF
--- a/src/views/darkroom.c
+++ b/src/views/darkroom.c
@@ -572,13 +572,27 @@ void expose(
     const float *point = dev->gui_module->color_picker_point;
     if(darktable.lib->proxy.colorpicker.size)
     {
-      cairo_rectangle(cri, box[0] * wd, box[1] * ht, (box[2] - box[0]) * wd, (box[3] - box[1]) * ht);
-      cairo_stroke(cri);
       cairo_translate(cri, 1.0 / zoom_scale, 1.0 / zoom_scale);
-      cairo_set_source_rgb(cri, .8, .8, .8);
-      cairo_rectangle(cri, box[0] * wd + 1.0 / zoom_scale, box[1] * ht,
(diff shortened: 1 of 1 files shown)


TurboGit:
That's really nice ! Could we have a visual indicator when we are in the area with the resize can be done ? At least this will help advertising this feature.

dterrahe:
Don't forget that none of this will be remembered in the module data; go out and back into darkroom and you have to start from scratch. So I don't feel there is too much point advertising a nice little almost free feature that then doesn't live up to the new (higher) expectations. Also, you can't drag the edges, only corner points (like it used to be before, but unlike crop/rotate). Unless there is functionality in crop/rotate that can be directly borrowed for this, I wouldn't really want to over engineer, to be honest. Also because I have no idea how to go about adding indicators in the color picker code.

And new users would find this anyway, just because they would expect it to work and try :-) Existing users can read the changelog ;-)

Unless you are saying we should beef up color pickers and make them rotatable, different shapes etc. ? That would be a nice, bigger, future, project. Maybe an integration with the masking code?

TurboGit:
> Unless you are saying we should beef up color pickers and make them rotatable, different shapes etc. ? That would be a nice, bigger, future, project. Maybe an integration with the masking code?

No, no rotation or anything fancy. I just wanted have a discoverable GUI hint about this feature. Maybe a hint message at least ?

dterrahe:
The size of the little square doesn't necessarily exactly match the zone of sensitivity because one is defined in pixels and the other, in button_pressed, in relative size. But at least there is an indication that the corner can be dragged.

TurboGit:
Well, I like that, at least those handles are really a hint. Let's merge this to get more feedback.

PR #4940, opened 2020-05-07
Reboot channelmixer

aurelienpierre:
This started as a rewrite of channel mixer because the previous one clips RGB at 100%, which causes problems with scene-linear workflow. It also allows to normalize channels to 1 so brightness is (sort-of) preserved (sort-of, because we don't compensate for Helmholtz-Kohlrauch effect). 

I don't think there is a clean way to upgrade the current channel mixer while preserving decent performance. Also, the HSL variant makes zero sense, since the channel mixer is a matrix dot product that already contains a hue rotation as part of the linear map/vector base change it performs.

So, let's start clean.

## What does it do ?

Same as before: R / G / B mixing. But not only. 

Colorfulness and brightness adjustments are backported from the new color science in #4800, so  come from the same norm vs. ratio demodulation, without having to use shitty HSL.

Capture d’écran du 2020-05-07 20-08-19

But since we are already in LMS, why not use it for what is was originally designed : chromatic adaptation transform (CAT), aka white balance adjustment ? The current white balance adjustment in dt is very primitive, inaccurate and rather messy.

So I added an `illuminant.h` library that :
1. computes accurately the correlated color temperature from the white point chromaticity, 
2. computes accurately the white point chromaticity from the color temperature. 
3. contains the whole list of CIE standard illuminants (fluorescent, incandescent, DEL, daylight and black body):

It clips the negative pixels after white balancing as a brute-force gamut-mapping to deal with imaginary colors (especially in blue).

## How does it works ?

This channel mixer works in LMS cone space, a special kind of RGB adjusted for homogeneous chromatic adaptation. That makes the balancing more even, while still preserving the scene-referred connection, since we go from pipeline RGB to XYZ to LMS only by matrice dot products.

The white balance is initialized with the WB coefficients read from the RAW image. Those can be retrieved anytime using "illuminant" -> "compute from camera…". If the WB coefficients lead to a chromaticity close to the planckian locus (< 0.1% error), the UI switches in daylight mode:
Capture d’écran du 2020-05-20 16-36-03

If not, the UI switches to the custom mode, where users are presented the exact x and y chromaticity coordinates for emergency damage-control:
Capture d’écran du 2020-05-20 16-34-37

Here is a comparison between classic dt white balance using camera default (left), and the Bradford transform in the CAT tab of channel mixer, also using camera default (right):
Capture d’écran du 2020-05-07 20-06-32

Notice that you need to set the old white balance to "camera neutral" if you want to use the Bradford CAT instead (don't disable the old white balance, we still need the default D65 coefficients).

I have also added the ability to perform gamut mapping and to clip negative RGB to sanitize files shot under blue LED and weird illuminants. These situations happen a lot to events and concert photographers.

Before gamut mapping:
LV0_5175_06

After:
LV0_5175_05

There is also a "machine-learning" WB auto-adjustment that doesn't try to make the whole image grey when you have legitimate colours.

Current dt detected WB with the spot tool (on the whole image):
IMG_9568_03

"machine-learning" auto-detected WB (on the whole image):
IMG_9568_04

WB as shot (detected by the camera):
IMG_9568_05

## More screenshots

Collection of standard CIE illuminants:
Capture d’écran du 2020-05-20 16-37-30

The color patch shows the color of the current illuminant, the CCT on the left is an approximation of the color temperature for any illuminant (but keep in mind it's meaningful only if the illuminant is close to daylight).

## WIP

What remains to be done:
* ~~auto-adjust white balance~~
* turn the 3×3 params settings in R / G / B tab into a 3×2 rotation and scaling parameter as an alternative GUI.
* OpenCL.


johnny-bit:
This is awesome!

couple of my notes:
1. RE camera neutral WB - would it make more sense to rip out "generic" part of that calculation from temperature.c and have it for use by both - naive `temperature.c` and `channelmixerrgb`
2. UI stuff - you've focused screenshots on temp adjustment and that's great! but... I really get tripped by multi-op modules :/ I know it's powerfull and all + code is the same for ops (haven't looked at it yet, sorry!) and makes sense to keep it together but really - would it be a bad idea to suggest to split tunctionalities over the lines of: CAT02 transform, RGB mixing, that 3 things you haven't mentioned ;)
3. in R/G/B tabs there's still thing that really bothers me in "old" channel mixer - separate r/g/b target tabs (in case of old one - dropdowns). My idea to "fix" old one is to put in destination: "RGB"/"HLS"/"Gray" and have all necessary sliders visible and sectioned... probably even with "demo" of results but that's a bit much ;)

aurelienpierre:
>     1. RE camera neutral WB - would it make more sense to rip out "generic" part of that calculation from temperature.c and have it for use by both - naive `temperature.c` and `channelmixerrgb`

It could. The temperature computation is in an `.h` lib, so you can directly call the function from elsewhere. 

> would it be a bad idea to suggest to split tunctionalities over the lines of: CAT02 transform, RGB mixing, that 3 things you haven't mentioned ;)

1. CAT and RGB mixing, or if we call RGB mixing by what it is, aka correcting the camera RGB primaries, are 2 parts of the same thing : chromatic adjustment. The first corrects the white balance, the second evens the input profile/calibration.
2. They all work in LMS space, converted from Lab to pipeline RGB to XYZ to LMS, so splitting the features would duplicate the conversions(ss). The whole thing runs under 70-80 ms on CPU at 4K as it is.
3. The same argument could be made with multiplying modules.
4. Feature-wise, it's very basic, very versatile, very powerful adjustments that can fix bad lighting like no other module.

johnny-bit:
OK, so... can it be made so that WB module can be disabled, or would it be better to have it enabled but forcefully set to D65?

Regarding spliting functionalites into modules - was just a suggestion but no problem in not splitting it ;)

and last part (not addressed) still stands - merge some tabs (rgb at least) and show 9 sliders at once instead of 3sliders3tabs ;)

Other that this - I need to test it, but after work :)

rawfiner:
I like a lot this new module :-)
I think there should be an option in the CAT tab to disable this correction: the module could be used with several instances with masks for brightness changes, and we don't want to adjust WB in that case.
Would also be nice to have slider with soft bounds (at least for brightness and gray), sometimes, going until 1 or -1 is not enough.

aurelienpierre:
The CAT can be "disabled" (rather, neutralized) if on uses the illuminant "pipeline default" (first in the list). This is equivalent to mapping D65 to D65, so its a no-op.

> and last part (not addressed) still stands - merge some tabs (rgb at least) and show 9 sliders at once instead of 3sliders3tabs ;)

Not sure if 9 sliders at once is better. But I plan anyway to offer a second UI with 3 color wheels and one scaling factor, since the whole 3×3 mixing process can be re-written as a [polar decomposition using Euler angles.](https://en.wikipedia.org/wiki/Polar_decomposition)

johnny-bit:
3 wheels would be fine too :)

9 sliders are simply better representing (IMO) what's going to happen and seem more concise and actually reduce mental load because R/G/B aren't separate items but actual part of whole so the decomposition and bringing them back together is (for me) a bit nicer still... Additionally gimp uses that format now and it is honestly easier to use that way (less switching).

Mark-64:
I vote for color wheels, they make the UI more compact vertically and with less need to use the scrollbars, which helps the usability.
I actually would like color wheels also in Color Balance, which has a lot of sliders...

aurelienpierre:
>  I actually would like color wheels also in Color Balance, which has a lot of sliders...

That's actually a very different thing. Color balance is not an RGB space rotation at all, there is no connection with Euler angles and vector base. The color wheels I have in mind for channel mixer only use a rotation to define vector base switch, so that's a 1D control with color feedback. The color wheels you have in mind for color balance are 2D and use radius + angle to define saturation + hue. They come from video because the color grading guys use hardware wheels like this : 
image
But in photography, using 2D color wheels with a pointer is a lot less accurate than 3 sliders you control in 1D.

aurelienpierre:
Latest commit adds CAT16 in addition of Bradford and offers the option to disable CAT at all.

When using CAT16, the whole* module works in CIECAM16 LMS space.
When using Bradford, the whole* module works in Bradford LMS space.
When not using any CAT, the whole* module works in XYZ scaled with Y.

* Including R, G, B, grey mixing and brightness / colorfulness.

Capture d’écran du 2020-05-08 14-30-48

Mark-64:
> color wheels with a pointer is a lot less accurate than 3 sliders

Well, color wheels can be accurate too. In Capture One you have this
Cattura4
In each wheel, the angle controls hue, the distance from center (or left curved slider) controls saturation and the right curved slider controls the factor.
I have no idea if Gtk allows the creation of such fancy widgets, but just to say.

aurelienpierre:
@Mark-64 I suspect Capture One limits the available range of saturation to 25% or something, in order to keep the whole thing manageable. It will always be less accurate anyway. More compact and more sexy, but less accurate. If you want accuracy, you need GUI real estate, there is no secret.

MStraeten:
latest commits seemed to introduce compile errors on osx:
```
[ 50%] Building C object src/iop/CMakeFiles/channelmixerrgb.dir/introspection_channelmixerrgb.c.o
In file included from /Users/martinstraeten/src/darktable_debug/build/src/iop/introspection_channelmixerrgb.c:104:
In file included from /Users/martinstraeten/src/darktable_debug/src/iop/channelmixerrgb.c:28:
/Users/martinstraeten/src/darktable_debug/src/common/illuminants.h:300:26: error:
      initializer element is not a compile-time constant
  static const float p = powf(1.088932f / 0.818155f, 0.0834f);
                         ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
1 error generated.
```

Dannny1:
Maybe off-topic idea: but recently ability to manipulate primaries was added to ART channel mixer:
https://bitbucket.org/agriggio/art/commits/a12af235be880a104e0c4b049d990d30121b32d6
Do you think something like this could be included to darktable in future?

aurelienpierre:
@Dannny1 Manipulating primaries is exactly the topic of this PR. It's a channel mixer. You already have (a flawed) one in darktable, this is only a revamping of the feature.

Dannny1:
Thank you. I see now, it's just different way.

aurelienpierre:
Latest commit adds a gamut compression option.

The reason is the chromatic adaptation can push RGB values way out of visible spectrum, as shown here:

image

Example from Luc Viatour (original, corrected WB):
LV0_4036_07

Corrected WB + slight gamut mapping
LV0_4036_08

Corrected WB + heavy gamut mapping
LV0_4036_09

MStraeten:
found an error when compiling on osx:
```
In file included from /Users/martinstraeten/src/darktable_debug/build/src/iop/introspection_channelmixerrgb.c:110:
In file included from /Users/martinstraeten/src/darktable_debug/src/iop/channelmixerrgb.c:28:
/Users/martinstraeten/src/darktable_debug/src/common/illuminants.h:405:25: error:
      variable 'LUT_samples' must have explicitly specified data sharing
      attributes
  for(size_t i = 0; i < LUT_samples; i++)
                        ^~~~~~~~~~~
/Users/martinstraeten/src/darktable_debug/src/common/illuminants.h:401:39: note:
      explicit data sharing attribute requested here
                                      ^
1 error generated.
```

aurelienpierre:
Thanks @MStraeten. Should be fixed now.

MStraeten:
new errors:
```
[ 51%] Building C object src/iop/CMakeFiles/channelmixerrgb.dir/introspection_channelmixerrgb.c.o
In file included from /Users/martinstraeten/src/darktable_debug/build/src/iop/introspection_channelmixerrgb.c:110:
In file included from /Users/martinstraeten/src/darktable_debug/src/iop/channelmixerrgb.c:28:
/Users/martinstraeten/src/darktable_debug/src/common/illuminants.h:415:3: error:
      function definition is not allowed here
  {
  ^
/Users/martinstraeten/src/darktable_debug/src/common/illuminants.h:423:60: error:
      assigning to 'struct pair' from incompatible type 'int'
  ...reduction(pairmin:struct pair:omp_out=pair_min(omp_out,omp_in))    \
                                          ^~~~~~~~~~~~~~~~~~~~~~~~~
/Users/martinstraeten/src/darktable_debug/src/common/illuminants.h:431:68: error:
      incorrect reduction identifier, expected one of '+', '-', '*', '&', '|',
      '^', '&&', '||', 'min' or 'max' or declare reduction for type
      'struct pair'
  ...y, T_min, T_range, LUT_samples) reduction(pairmin:min_radius)\
                                               ^~~~~~~
/Users/martinstraeten/src/darktable_debug/src/common/illuminants.h:450:16: error:
      assigning to 'struct pair' from incompatible type 'int'
    min_radius = pair_min(min_radius, radius_tmp);
               ^ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
4 errors generated.
make[2]: *** [src/iop/CMakeFiles/channelmixerrgb.dir/introspection_channelmixerrgb.c.o] Error 1
make[1]: *** [src/iop/CMakeFiles/channelmixerrgb.dir/all] Error 2
make: *** [all] Error 2
```

aurelienpierre:
Latest commit should fix Clang error and introduces a novel white-balance auto-detection ("illuminant" -> "auto-detect from image content…". This one is a custom-made weighting of pixels chromaticity depending on whether they are on an edge (grey-edge assumption) or on a surface (grey-world assumption) and on their luminance (achromatic white assumption). Please test it and report, it seems to behave quite well in my tests, but it's new.

MStraeten:
thanks, compilation was successful and auto detection is also working - maybe it make sense to enable a selection as it is done in filmicrgb or black/white relative exposure to restrict it to the relevant parts of a scene.
i found an issue if the image is cropped - the result of auto-detection is a black frame
sample:
https://mega.nz/file/vXoC2KDR#kYo0kjSBHCyrTC8NetWA95qxYHIoGkVORtakfPB8Bm0
[20200119-IMG_7608.CR2.xmp.txt](https://github.com/darktable-org/darktable/files/4640457/20200119-IMG_7608.CR2.xmp.txt)

aurelienpierre:
Thanks @MStraeten, as usual.

Regarding the color-picker kind of WB selection, it's not really adaptable here. The WB selection you are used to makes an average over the selected area, and call that white. Here, what we do is borderline machine-learning, with an internal edge-detection, so restricting the area might give worse results. But I could backport the basic averaging over area for a manual selection.

I have reproduced the issue with framing, it's super weird, I don't really understand.

src/common/polar_decomposition.h
```diff
@@ -0,0 +1,380 @@
...
+        for it = 1:nit
+            v = IL*v;
+            v(1) = v(1)/D(1,1);v(2) = v(2)/D(2,2);v(3:4) = ID*v(3:4)/DD(1);
+            v = v'*IL;v = v';
+            nv = realsqrt(v(1)*v(1)+v(2)*v(2)+v(3)*v(3)+v(4)*v(4));
+            v = v/nv(1);
+        end
+    end
+    end
+    v(p) = v;
```
johnny-bit, line 379: Everything goes nice, suddenly... Pythons! Bunch of them! :stuck_out_tongue_closed_eyes:

aurelienpierre: WIP. It's actually Matlab code.

johnny-bit: Ah shit... i haven't noticed lack of ':' and misaligned `end` ;)

aurelienpierre:
Latest commit changes the chromaticity controls in white balance/CAT custom illuminant mode. 

Instead of the direct x, y coordinates (which make sense physically, but are tricky to set-up for users due to the heavy perceptual non-linearity of colors), the params are converted on-the-fly, back and forth, from x, y to the Lch space derivated from CIE Luv 1976. 

This ensures the R, G, and B primaries span over exactly a third of the hue angle range. It is a bit different from the Lch space derivated from CIE Lab 1976. The real parameters sent to the processing are still x, y radiometrically linear chromaticity.

Please check that the back-and-forth computations behave (no gap or shift).

MStraeten:
latest commits caused a compilation error on osx:
```
[ 53%] Building C object src/iop/CMakeFiles/channelmixerrgb.dir/introspection_channelmixerrgb.c.o
In file included from /Users/martinstraeten/src/darktable_debug/build/src/iop/introspection_channelmixerrgb.c:111:
In file included from /Users/martinstraeten/src/darktable_debug/src/iop/channelmixerrgb.c:28:
/Users/martinstraeten/src/darktable_debug/src/common/illuminants.h:319:26: error: initializer element is
      not a compile-time constant
  static const float p = powf(1.088932f / 0.818155f, 0.0834f);
                         ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
1 error generated.
make[2]: *** [src/iop/CMakeFiles/channelmixerrgb.dir/introspection_channelmixerrgb.c.o] Error 1
make[1]: *** [src/iop/CMakeFiles/channelmixerrgb.dir/all] Error 2
```

aurelienpierre:
Damit, I keep forgetting that Clang doesn't like static variables computed from something.

eyedear:
i am getting  compile error in windows 10 here is the output. i am new if there is anything not complete please let me know 

$ cmake --build .
[  1%] Updating version string (git checkout)
Version string: 3.1.0+2055~g7306941b8-dirty

[ 60%] Generating introspection_channelmixerrgb.c
Scanning dependencies of target channelmixerrgb
[ 60%] Building C object src/iop/CMakeFiles/channelmixerrgb.dir/introspection_channelmixerrgb.c.obj
In file included from C:/msys64/home/Dell/darktable/src/control/control.h:21,
                 from C:/msys64/home/Dell/darktable/src/bauhaus/bauhaus.h:22,
                 from C:/msys64/home/Dell/darktable/src/iop/channelmixerrgb.c:22,
                 from C:/msys64/home/Dell/darktable/build/src/iop/introspection_channelmixerrgb.c:151:
C:/msys64/home/Dell/darktable/src/iop/channelmixerrgb.c: In function 'auto_detect_WB':
C:/msys64/home/Dell/darktable/src/iop/channelmixerrgb.c:476:48: error: expected ')' before '(' token
  476 |   dt_omp_firstprivate(width, height, ch, temp, stdout, chroma_mean, luma_mean, chroma_std, luma_std, flat_chroma_mean, flat_chroma_var) \
      |                                                ^~~~~~
C:/msys64/home/Dell/darktable/src/common/darktable.h:89:49: note: in definition of macro 'dt_omp_firstprivate'
   89 | #  define dt_omp_firstprivate(...) firstprivate(__VA_ARGS__)
      |                                                 ^~~~~~~~~~~
C:/msys64/home/Dell/darktable/src/common/darktable.h:89:48: note: to match this '('
   89 | #  define dt_omp_firstprivate(...) firstprivate(__VA_ARGS__)
      |                                                ^
C:/msys64/home/Dell/darktable/src/iop/channelmixerrgb.c:476:3: note: in expansion of macro 'dt_omp_firstprivate'
  476 |   dt_omp_firstprivate(width, height, ch, temp, stdout, chroma_mean, luma_mean, chroma_std, luma_std, flat_chroma_mean, flat_chroma_var) \
      |   ^~~~~~~~~~~~~~~~~~~
In file included from C:/msys64/home/Dell/darktable/build/src/iop/introspection_channelmixerrgb.c:151:
C:/msys64/home/Dell/darktable/src/iop/channelmixerrgb.c:504:39: error: 'luma_mean' not specified in enclosing 'parallel'
  504 |       const float weight_luma = expf(-sqf(central_average[2] - luma_mean) / luma_std);
      |                                       ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
C:/msys64/home/Dell/darktable/src/iop/channelmixerrgb.c:475:9: error: enclosing 'parallel'
  475 | #pragma omp parallel for simd default(none) reduction(+:XYZ_surface) reduction(+:norm_surface)\
      |         ^~~
C:/msys64/home/Dell/darktable/src/iop/channelmixerrgb.c:504:33: error: 'luma_std' not specified in enclosing 'parallel'
  504 |       const float weight_luma = expf(-sqf(central_average[2] - luma_mean) / luma_std);
      |                                 ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
C:/msys64/home/Dell/darktable/src/iop/channelmixerrgb.c:475:9: error: enclosing 'parallel'
  475 | #pragma omp parallel for simd default(none) reduction(+:XYZ_surface) reduction(+:norm_surface)\
      |         ^~~
C:/msys64/home/Dell/darktable/src/iop/channelmixerrgb.c:505:40: error: 'flat_chroma_mean' not specified in enclosing 'parallel'
  505 |       const float weight_chroma = expf(-(sqf(central_average[0]) + sqf(central_average[1]) - flat_chroma_mean) / flat_chroma_var);
      |                                        ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
C:/msys64/home/Dell/darktable/src/iop/channelmixerrgb.c:475:9: error: enclosing 'parallel'
  475 | #pragma omp parallel for simd default(none) reduction(+:XYZ_surface) reduction(+:norm_surface)\
      |         ^~~
C:/msys64/home/Dell/darktable/src/iop/channelmixerrgb.c:505:35: error: 'flat_chroma_var' not specified in enclosing 'parallel'
  505 |       const float weight_chroma = expf(-(sqf(central_average[0]) + sqf(central_average[1]) - flat_chroma_mean) / flat_chroma_var);
      |                                   ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
C:/msys64/home/Dell/darktable/src/iop/channelmixerrgb.c:475:9: error: enclosing 'parallel'
  475 | #pragma omp parallel for simd default(none) reduction(+:XYZ_surface) reduction(+:norm_surface)\
      |         ^~~
make[2]: *** [src/iop/CMakeFiles/channelmixerrgb.dir/build.make:88: src/iop/CMakeFiles/channelmixerrgb.dir/introspection_channelmixerrgb.c.obj] Error 1
make[1]: *** [CMakeFiles/Makefile2:3897: src/iop/CMakeFiles/channelmixerrgb.dir/all] Error 2
make: *** [Makefile:172: all] Error 2

TurboGit:
Won't make it for 3.2, code feature freeze is in 10 days.

hatsunearu:
any chance more work can be put in to make it into 3.2?

TurboGit:
@hatsunearu : That's not me to decide but certainly Aurélien has already many things todo for 3.2 (FilmicRGB v4 is to one) and pushing this in 3.2 without much testing isn't going to help darktable. We all love new things but I do want something stable and polished for a release. And after all if it miss 3.2 it will be for 3.4 at then end of year.

aurelienpierre:
I would love to bring it in 3.2, but I'm afraid it wouldn't be reasonable.

MStraeten:
i was able to reproduce the build errors on win10 issued by eyedear (https://github.com/darktable-org/darktable/pull/4940#issuecomment-639246980)

aurelienpierre:
@MStraeten Could you try to insert a space on line 476, between `reduction(+:norm_surface)` and the end of line `\` ?

MStraeten:
didn't help:
```
[ 60%] Building C object src/iop/CMakeFiles/channelmixerrgb.dir/introspection_channelmixerrgb.c.obj
In file included from C:/msys64/home/Me/darktable/src/control/control.h:21,
                 from C:/msys64/home/Me/darktable/src/bauhaus/bauhaus.h:22,
                 from C:/msys64/home/Me/darktable/src/iop/channelmixerrgb.c:22,
                 from C:/msys64/home/Me/darktable/build/src/iop/introspection_channelmixerrgb.c:151:
C:/msys64/home/Me/darktable/src/iop/channelmixerrgb.c: In function 'auto_detect_WB':
C:/msys64/home/Me/darktable/src/iop/channelmixerrgb.c:477:48: error: expected ')' before '(' token
  477 |   dt_omp_firstprivate(width, height, ch, temp, stdout, chroma_mean, luma_mean, chroma_std, luma_std, flat_chroma_mean, flat_chroma_var) \
      |                                                ^~~~~~
C:/msys64/home/Me/darktable/src/common/darktable.h:89:49: note: in definition of macro 'dt_omp_firstprivate'
   89 | #  define dt_omp_firstprivate(...) firstprivate(__VA_ARGS__)
      |                                                 ^~~~~~~~~~~
C:/msys64/home/Me/darktable/src/common/darktable.h:89:48: note: to match this '('
   89 | #  define dt_omp_firstprivate(...) firstprivate(__VA_ARGS__)
      |                                                ^
C:/msys64/home/Me/darktable/src/iop/channelmixerrgb.c:477:3: note: in expansion of macro 'dt_omp_firstprivate'
  477 |   dt_omp_firstprivate(width, height, ch, temp, stdout, chroma_mean, luma_mean, chroma_std, luma_std, flat_chroma_mean, flat_chroma_var) \
      |   ^~~~~~~~~~~~~~~~~~~
In file included from C:/msys64/home/Me/darktable/build/src/iop/introspection_channelmixerrgb.c:151:
C:/msys64/home/Me/darktable/src/iop/channelmixerrgb.c:505:39: error: 'luma_mean' not specified in enclosing 'parallel'
  505 |       const float weight_luma = expf(-sqf(central_average[2] - luma_mean) / luma_std);
      |                                       ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
C:/msys64/home/Me/darktable/src/iop/channelmixerrgb.c:476:9: error: enclosing 'parallel'
  476 | #pragma omp parallel for simd default(none) reduction(+:XYZ_surface) reduction(+:norm_surface) \
      |         ^~~
C:/msys64/home/Me/darktable/src/iop/channelmixerrgb.c:505:33: error: 'luma_std' not specified in enclosing 'parallel'
  505 |       const float weight_luma = expf(-sqf(central_average[2] - luma_mean) / luma_std);
      |                                 ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
C:/msys64/home/Me/darktable/src/iop/channelmixerrgb.c:476:9: error: enclosing 'parallel'
  476 | #pragma omp parallel for simd default(none) reduction(+:XYZ_surface) reduction(+:norm_surface) \
      |         ^~~
C:/msys64/home/Me/darktable/src/iop/channelmixerrgb.c:506:40: error: 'flat_chroma_mean' not specified in enclosing 'parallel'
  506 |       const float weight_chroma = expf(-(sqf(central_average[0]) + sqf(central_average[1]) - flat_chroma_mean) / flat_chroma_var);
      |                                        ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
C:/msys64/home/Me/darktable/src/iop/channelmixerrgb.c:476:9: error: enclosing 'parallel'
  476 | #pragma omp parallel for simd default(none) reduction(+:XYZ_surface) reduction(+:norm_surface) \
      |         ^~~
C:/msys64/home/Me/darktable/src/iop/channelmixerrgb.c:506:35: error: 'flat_chroma_var' not specified in enclosing 'parallel'
  506 |       const float weight_chroma = expf(-(sqf(central_average[0]) + sqf(central_average[1]) - flat_chroma_mean) / flat_chroma_var);
      |                                   ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
C:/msys64/home/Me/darktable/src/iop/channelmixerrgb.c:476:9: error: enclosing 'parallel'
  476 | #pragma omp parallel for simd default(none) reduction(+:XYZ_surface) reduction(+:norm_surface) \
      |         ^~~
make[2]: *** [src/iop/CMakeFiles/channelmixerrgb.dir/build.make:88: src/iop/CMakeFiles/channelmixerrgb.dir/introspection_channelmixerrgb.c.obj] Fehler 1
make[1]: *** [CMakeFiles/Makefile2:3897: src/iop/CMakeFiles/channelmixerrgb.dir/all] Fehler 2
make: *** [Makefile:172: all] Fehler 2

```

aurelienpierre:
oooh it's the `stdout` I use to debug.

MStraeten:
ok on windows10

jade-nl:
I've been playing around with this module for a bit now and I really like how this one is turning out.

I came here to ask you about something specific which was already asked by @johnny-bit but you did not address that question (yet):

Like johnny-bit I'm also curious if the white balance can be automatically forced to D65 if anything but *XYZ (none)* is chosen in the CAT tab. Or a *workflow* option in the preferences (like the scene-referred / display-referred / none workflow option) that takes care of this.

I can really see myself not using the current white balance module in my future workflow any longer and using this module instead to do WB related stuff.

Anyway, thanks for your continuous work on improving darktable. Much appreciated!

rkowalke:
Compilation of this PR fails with current master. 

```
src/iop/channelmixerrgb.c:1906:3: Fehler: Implizite Deklaration der Funktion »dtgtk_justify_notebook_tabs« [-Werror=implicit-function-declaration]
 1906 |   dtgtk_justify_notebook_tabs(g->notebook);
      |   ^~~~~~~~~~~~~~~~~~~~~~~~~~~
```

The reason seems to be that called _dtgtk_justify_notebook_tabs_ was removed with commit fee9a065b744bea665d4aa15cb5bfeb8d0668607 by @dterrahe

dterrahe:
Since this PR was first submitted in draft substantial changes were made to the codebase.  To get it to compile again, replace the notebook setup code in gui_init with this:
```
  g->notebook = GTK_NOTEBOOK(gtk_notebook_new());
  GtkWidget *page0 = dt_ui_notebook_page(g->notebook, _("CAT"), NULL);
  GtkWidget *page1 = dt_ui_notebook_page(g->notebook, _("R"), NULL);
  GtkWidget *page2 = dt_ui_notebook_page(g->notebook, _("G"), NULL);
  GtkWidget *page3 = dt_ui_notebook_page(g->notebook, _("B"), NULL);
  GtkWidget *page4 = dt_ui_notebook_page(g->notebook, _("colorfulness"), NULL);
  GtkWidget *page5 = dt_ui_notebook_page(g->notebook, _("brightness"), NULL);
  GtkWidget *page6 = dt_ui_notebook_page(g->notebook, _("grey"), NULL);
  gtk_widget_show_all(GTK_WIDGET(g->notebook));
  gtk_box_pack_start(GTK_BOX(self->widget), GTK_WIDGET(g->notebook), FALSE, FALSE, 0);
```
Also do a search&replace for "self->dt->" -> "darktable."

Ideally, the gui code should be converted to the introspection based approach introduced in #5335. @aurelienpierre let me know if you want help with that.

rkowalke:
This gui_init code is already in place a few lines above. In my local working copy I deleted the errornous line and replaced "self->dt" as recommended. At least this solved the compilation error.

aurelienpierre:
I will fix that in the next days

aurelienpierre:
Should be fixed now.

aurelienpierre:
Algo tested for 5 months in production, code cleaned, commits squashed, I believe this is ready for review. I also added a color-picker for manual WB selection, as well as 2 modes for the AI to detect illuminant in image.

There is still a glitch when the module is first loaded : it should be inited with "illuminant = as shot in camera", and it does not. You need to reset the module to see it properly, or to go at the first step in history. Same, when history is compressed, I don't get why, but the defaults are not properly loaded (but it seems the history compression triggers `reload_defaults()` as well as the module reset, so I don't get why it behaves differently).

matt-maguire:
> There is still a glitch when the module is first loaded : it should be inited with "illuminant = as shot in camera"

Sorry, I don't get this remark at all. The only place you set the illuminant to `DT_ILLUMINANT_CAMERA` is in `reload_defaults()`, but immediately after that you call `check_if_close_to_daylight()` which will always set the illuminant to `DT_ILLUMINANT_BB` or `DT_ILLUMINANT_D` if the error is small, or in all other cases it will set it to `DT_ILLUMINANT_CUSTOM` and set `adaptation=DT_ADAPTATION_CAT16`. 

Therefore it is impossible for the illuminant to ever get initialised to DT_ILLUMINANT_CAMERA. Or maybe I have misunderstood something?

I didn't understand the remarks about module reset or history compression, as these seem to behave as per the logic above.

aurelienpierre:
Whenever I open a new RAW edit or compress an old edit at step 0, the illuminant gets initialized in UI to `DT_ILLUMINANT_DETECT_EDGES` which makes zero sense, since it is only a GUI helper to trigger a detection. But, similarly, since a couple of weeks, the usual white balance module is not applied by default or RAW images, I need to turn it on manually every time. So maybe there is something going on with the `reload_defaults` at pipeline level.

matt-maguire:
ok, I can't reproduce, but then I just did a clean build just today, and I only pulled down PR4940, so maybe that's why I can't see what you describe.

dterrahe:
> There is still a glitch when the module is first loaded : it should be inited with "illuminant = as shot in camera", and it does not. You need to reset the module to see it properly, or to go at the first step in history. Same, when history is compressed, I don't get why, but the defaults are not properly loaded (but it seems the history compression triggers `reload_defaults()` as well as the module reset, so I don't get why it behaves differently).

I'm not sure if this is the cause of your problems here, but reload_defaults is called immediately after gui_init, but _before_ the gui widget is added to the panel, so it has not been shown_all yet. When it _is_ added afterwards, and all the individual widgets get realised, some of their properties get reinitialised. This causes reload_defaults to behave differently the first time from any later times which causes confusion, complicates testing and requires workarounds.

I'm working on a PR to fix this, but since it is a fundamental change, I'm being careful about it. Will finish as soon as #6648 is done.

TurboGit:
@aurelienpierre : the color picker on the CAT tab to select the white balance seems ineffective.

TurboGit (changes requested):
The color picker should be fixed.

The file polar_decomposition.c not used, not compilable and with MathLab code should be removed.

TurboGit:
I still cannot make the color picker working, it seems to work the first time and then nothing happens (this is when merged on master).

Also changing multiple times from "`detect from image edges`" and "`detect from image surface`" makes dt corrupt output (see screenshot bellow) and sometime crash dt.

image

dterrahe:
Is this going in for 3.4? Can I help with converting to introspection based params/gui?

TurboGit:
Ok, the color picker bug seems to be the same as described in #6360.

The issue is that colorout module set the work-profile to NULL as no module after should care. But the pipe is not restarted from start when changing the area. So no work-profile and we exit early as in color_picker_apply we have and early exit in this case:

`  if(work_profile == NULL) return;`

So not a bug from this module, but a quite big regression that I cannot see how to fix up to know :(

TurboGit:
> Is this going in for 3.4? Can I help with converting to introspection based params/gui?

Yes please go ahead. We'd want this new module in.

dterrahe:
> > Is this going in for 3.4? Can I help with converting to introspection based params/gui?
> 
> Yes please go ahead. We'd want this new module in.

Do I clone from auelien's fork and submit a PR to him, or submit to main branch after this PR is merged?

TurboGit:
I think I have the color-picker fix : #6765

TurboGit:
@dterrahe :
> Do I clone from auelien's fork and submit a PR to him, or submit to main branch after this PR is merged?

If it works for you a PR when merge will be ok by me.

dterrahe:
> If it works for you a PR when merge will be ok by me.

It may take me a few days, so if this lands just before the freeze, I might not be in time. Introspectizing isn't _really_ a new feature and shouldn't lead to new translatable strings, so hopefully you'll be OK with that...

TurboGit:
@dterrahe : ok so to be safe maybe better starting now from a clone. The code should not change much now.

aurelienpierre:
> So not a bug from this module, but a quite big regression that I cannot see how to fix up to know :(

fix in 88adadf. There is really no reason for colorout to nullify the pointer, colorin will overwrite it if the profile changes and since the profile is added to the `dev->allprofiles` `GList`, it will properly cleaned once the pipe exits to avoid memory leaks.

TurboGit:
@aurelienpierre : the crash/corruption on this module is because you do GUI stuff in `process()`. This is not possible. You need to cache the values somehow in process() is needed and do the update in `gui_update()`.

src/iop/channelmixerrgb.c
```diff
@@ -0,0 +1,2468 @@
...
+      g->auto_detect_illuminant = FALSE;
+
+      --darktable.gui->reset;
+
+      dt_control_log(_("auto-detection of white balance completed"));
+
+      dt_dev_add_history_item(darktable.develop, self, TRUE);
+      return;
+    }
+  }
```
TurboGit, line 815: All the GUI stuff in this section should be done in gui_update().

TurboGit (changes requested):
I think I found the crash, see comment.

src/iop/channelmixerrgb.c
aurelienpierre: Are you sure `gui_update()` is triggered after `process()` exits ? `auto_detect_WB(in, g->wb_model, roi_in->width, roi_in->height, ch, RGB_to_XYZ, XYZ);` writes the illuminant found in `XYZ`, but then we need to commit that to the parameters and update temperature and so on, then finally the GUI.

TurboGit: I'm not sure about update indeed, but I'm sure no Gtk code should be used in process() which is run in a different thread.

TurboGit: The best solution I suppose is to remove those two entries from the combo and add buttons that act like color-picker to do the detection.

aurelienpierre: But that won't change a thing if we put the detection in buttons, we will still need to run an algo on the pixels, wait for the processing to finish, get the result then output the value to GUI and update the widgets accordingly. Such things are done already in ashift, exposure and tone equalizer, although they use private `dt_pthread_mutex_unlock(&g->lock);`.

aurelienpierre: Fixed

TurboGit:
@aurelienpierre :

I have many error on the console:

```
[dt_ioppr_set_pipe_output_profile_info] unsupported output profile 2 , it will be replaced with sRGB
[dt_ioppr_set_pipe_output_profile_info] unsupported output profile 2 , it will be replaced with sRGB
[dt_ioppr_set_pipe_output_profile_info] unsupported output profile 2 , it will be replaced with sRGB
[dt_ioppr_set_pipe_output_profile_info] unsupported output profile 2 , it will be replaced with sRGB
[dt_ioppr_set_pipe_output_profile_info] unsupported output profile 8 , it will be replaced with sRGB
[dt_ioppr_set_pipe_output_profile_info] unsupported output profile 8 , it will be replaced with sRGB
[dt_ioppr_set_pipe_output_profile_info] unsupported output profile 8 , it will be replaced with sRGB
[dt_ioppr_set_pipe_output_profile_info] unsupported output profile 8 , it will be replaced with sRGB
[dt_ioppr_set_pipe_output_profile_info] unsupported output profile 8 , it will be replaced with sRGB
[dt_ioppr_set_pipe_output_profile_info] unsupported output profile 8 , it will be replaced with sRGB
[dt_ioppr_set_pipe_output_profile_info] unsupported output profile 8 , it will be replaced with sRGB
[dt_ioppr_set_pipe_output_profile_info] unsupported output profile 8 , it will be replaced with sRGB
```
Any idea ?

TurboGit:
The issue is not due to this PR, so I'll merge now and I'll open an issue for the output profile. Probably due to your previous commit.

TurboGit:
This is now merged. You'll find you polar_decomposition.h on the branch reboot-channelmixer on main dt repository.

aurelienpierre:
Thanks !

TurboGit:
@aurelienpierre : can you propose an entry for the RELEASE_NOTES.md for this new module ? TIA.

Issue #4857, opened 2020-04-29
Add ability to use lens information in the replacement  $(variable) that are used.


chhil:
Currently, I can't get the lens information available in overlays that are used.
Would like a replacement variable available for it.

![image](https://user-images.githubusercontent.com/1119439/80591909-244e3f00-8a3c-11ea-873b-44d6ba2bb1b8.png)

aurelienpierre:
fixed by #4945

PR #4945, opened 2020-05-08
Adding lens info to variables

Mark-64:
This will add a new variable ${LENS} to the ones that is possible to display in the darkroom image info footer toolbar

Cattura3


Mark-64:
Also close #4857

GLLM:
Many thanks for this :)

TurboGit:
Thanks !

PR #4947, opened 2020-05-08
CSS improvement for culling and preview rewrite

Nilvus:
Here it is, CSS improvements for the culling and preview rewrite from @AlicVB.

This fix star icons remains visible when moving mouse cursor outside the image on preview mode (and culling one).
Fix also paddings (thumbnail.c is to adjust bottom padding of overlay block infos).
Fix also images infos not seen on hover extended mode.

I also made a little cleanup for no more used lines and some minor other tweaks.

This also improve popover overlays mode when clicking on star icon.

This PR is over so could be merged if no issue is seen.

Just take in account that some things remains to improve and especially how icons are rendered on thumbnails (group, altered, ext name and even stars). Colors and hover effect are not good as they are but that will be for another PR as I will now work on optimize CSS code before. It's now quite messy with all new things added and some things become quite hard to tweak without having to change different places and sometimes separated ones. Will try to clean that and optimize all things I can. For example, in reality it's not possible to change all text size in one place. line 161 doesn't change all texts. Will take time.

### DIFF
--- a/data/themes/darktable.css
+++ b/data/themes/darktable.css
@@ -133,11 +133,9 @@
 @define-color thumbnail_bg50_color @grey_100;
 @define-color thumbnail_infos_color @grey_70; /* buttons and metadata text color on thumbnails */
 @define-color thumbnail_selected_bg_color @grey_65;
-@define-color thumbnail_hover_bg_color @grey_90;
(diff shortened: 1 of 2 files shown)


TurboGit:
Works fine on my side! Thanks.

PR #4949, opened 2020-05-08
Update french translation

Nilvus:
@TurboGit: Last translations updated and some older ones fixed. Especially some space missing before ':'. No space in english but one needed in french!

### DIFF
--- a/po/fr.po
+++ b/po/fr.po
@@ -2,7 +2,7 @@ msgid ""
 msgstr ""
 "Project-Id-Version: darktable 2.6\n"
 "Report-Msgid-Bugs-To: \n"
-"POT-Creation-Date: 2020-05-03 17:29+0200\n"
+"POT-Creation-Date: 2020-05-08 23:56+0200\n"
 "PO-Revision-Date: 2020-05-08 13:13+0200\n"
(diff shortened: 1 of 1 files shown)


Nilvus:
@TurboGit: just add missing lines I just see. When I launch intltool-update fr, I always have an error about a missing file 'darktable.desktop.in.h' in build/data folder. When I check on it, I see a darktable.desktop.in with lines in fr.po to translate, so I copy it and rename it darktable.desktop.in.h. After that intltool-update fr works. But that doesn't seems to be the good way.

Why I always have this file missing error? And what's the correct way to update po files by including this file as I find no files in code source with that name (with .h extension)?

Nilvus:
@EdgarLux: thanks for your help. I have add the error 3 times before but indeed now when redo intltoot-update (without darktable.desktop.in.h file), it works as you say. So last commit should be the good one...
@TurboGit: if ok for you, I think this could be merged.

Nilvus:
@TurboGit: and just see a wrong translation (fixed). That's too much commit for just french translation, so feel free to merge them at one if you want.

TurboGit:
Merged, thanks !

PR #4953, opened 2020-05-08
Minor fix on lighttable.c

EdgarLux:
Fixing error **"msgtr is not a valid C string format, unlike msgid"** when translating.


TurboGit:
Thanks

PR #4956, opened 2020-05-08
Tone equalizer color pickers fixes

moy:
The tone equalizer uses color pickers in a particular way compared to
others: the user never actually selects an area to pick, the whole
image is always used. As a consequence, the fact that the picker
widget is active doesn't mean "the picker is active and an area can be
selected on screen", but rather "the value of the slider was set
automatically".

The previous behavior was close to this, but broken in some
corner-cases:

* If one selected a value manually, and then clicked the button, then
  the button was toggled to active, but the behavior was to reset the
  value to 0. The visual feedback was the opposite of the behavior.

* When the slider was set manually, the picker widget was kept active,
  while the value wasn't the one set automatically anymore.

Fix this by marking the widget inactive whenever the value is set by
something other than the color picker.

While we're there, fix a trivial typo in a comment.

### DIFF
--- a/src/iop/toneequal.c
+++ b/src/iop/toneequal.c
@@ -1888,7 +1888,7 @@ static void auto_adjust_exposure_boost(GtkWidget *quad, gpointer user_data)
 
   if(p->exposure_boost != 0.0f)
   {
-    // Reset the contrast boost and do nothing
+    // Reset the exposure boost and do nothing
     p->exposure_boost = 0.0f;
     const int reset = darktable.gui->reset;
     darktable.gui->reset = 1;
--- a/src/iop/toneequal.c
+++ b/src/iop/toneequal.c
@@ -1853,6 +1853,7 @@ static void contrast_boost_callback(GtkWidget *slider, gpointer user_data)
 
   // Unlock the colour picker so we can display our own custom cursor
   dt_iop_color_picker_reset(self, TRUE);
+  dt_bauhaus_widget_set_quad_active(slider, FALSE);
 }
 
 static void exposure_boost_callback(GtkWidget *slider, gpointer user_data)
@@ -1867,6 +1868,7 @@ static void exposure_boost_callback(GtkWidget *slider, gpointer user_data)
 
   // Unlock the colour picker so we can display our own custom cursor
   dt_iop_color_picker_reset(self, TRUE);
+  dt_bauhaus_widget_set_quad_active(slider, FALSE);
 }
 
 static void auto_adjust_exposure_boost(GtkWidget *quad, gpointer user_data)
@@ -1897,6 +1899,7 @@ static void auto_adjust_exposure_boost(GtkWidget *quad, gpointer user_data)
 
     invalidate_luminance_cache(self);
     dt_dev_add_history_item(darktable.develop, self, TRUE);
+    dt_bauhaus_widget_set_quad_active(quad, FALSE);
     return;
   }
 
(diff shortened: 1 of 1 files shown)


TurboGit:
This change the behavior and I'm not sure it is expected. The current picker are two states, you activate it, it reads the value on the image, you click a second tome to deactivate it and apply the read value.

Also, note that the current work on simplifying the color pickers done by @dterrahe would probably help here. All in all, I would prefer not merging this now except if Aurélien confirms that the current behavior is not expected and is buggy.

moy:
> This change the behavior and I'm not sure it is expected. The current picker are two states, you activate it, it reads the value on the image, you click a second tome to deactivate it and apply the read value.

This is the behavior of most pickers, probably all except the ones in the tone equalizer. The tone equalizer uses the same icons as the picker, but the behavior is not the one of the picker. 
Actually, the only places where the code uses "gui/color_picker_proxy.h" is to avoid bad interaction between the blending picker and the specific cursor of the tone equalizer, but there's no real picker in the module (hence the other PR #4960 you've merged).

The tone equalizer's picker buttons trigger an action instantly, without letting the user select a piece of the image nor needing a deactivation. Try the current behavior if you don't see what I mean.

> Also, note that the current work on simplifying the color pickers done by @dterrahe would probably help here.

I wasn't aware of this, but I suspect they are unrelated, since the ton equalizer only uses the buttons, not the underlying color picker code.

Obviously, Aurélien's feedback is welcome, though!

aurelienpierre:
Yep, it's better this way. Thanks Matthieu !

TurboGit:
Merging then ! Thanks to both of you.

dterrahe:
Perhaps using a different icon would also help to reduce confusion somewhat? There is a tooltip, so that already helps, but it might not get read... dtgtk_cairo_paint_bulb maybe?

moy:
I did think of changing the icon, but I'm not 100% sure it would actually be better. Having a picker icon for not-a-picker behavior is confusing, but at least it carries the idea that the slider will be auto-set from the image.

aurelienpierre:
I used the color picker because it means "get that value from the image", even though there is no picking. I thought about using a magic wand but I'm not sure it would be legible enough.

todd-prior:
Maybe consider it an "Auto" setting and use a checkbox or small text auto that could be highlighted...un-highlighted if un-selected or modified manually??

moy:
IMHO a checkbox would be even more confusing: I'd expect an "auto" checkbox to be saved in the history stack and adapted when other modules change. Here, it's really a one-time action.

todd-prior:
Are there any tools in DT like that?? Ie an auto mode that updates??? Maybe call it sample as that is what it does sample and calc initial settings





Sent from Outlook Mobile

moy:
> Are there any tools in DT like that?? Ie an auto mode that updates??? Maybe call it sample as that is what it does sample and calc initial settings

For example the "automatic" mode of the exposure module does something more or less related to what we're talking about here, it computes the histogram and finds an exposure correction. But playing a bit with it, it seems the module reads the original image, not the image at this step of the pipeline (changing anything before the module in the pipe doesn't change the exposure compensation).

The closest to what we're talking about is the "auto" button in levels and basic adjustments. It behaves essentially like the pickers in tone equalizer, but uses the text "auto" instead of an icon. The drawback is that the buttons are larger, hence would hardly fit at the right hand side of the slider.

todd-prior:
I think that is what I was getting at…for example choosing spot WB will use the entire image automatically and it does a pretty good job. I have years of old jpgs with WB issues…problem is that you can’t put spot WB in a style to apply…it will apply the settings at the time you create the style so it doesn’t perform an updated spot WB on application. I think a way to do this with some controls in DT would be useful…other times not I suppose..

PR #4960, opened 2020-05-09
toneequal: remove dead code

moy:
One duplicate #include, and two unused fields.

The extra code didn't really harm, but confused me when reading the code, so let's clean it up.


TurboGit:
Thanks !

PR #4963, opened 2020-05-09
Disable CMake packaging for debug builds

Mark-64:
This will disable fixup_bundle() for Debug type of build.
fixup_bundle() is only needed for redistributable builds, normally of Release type, and causes a very long building time on Windows, which makes the debug painful.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -512,9 +512,10 @@ add_subdirectory(tools)
 # This contains fixup_bundle
 # And adding a separate subderectory as a last one will make sure
 # that fixup_bundle will run _after_ all files has been installed
-add_subdirectory(packaging)
-
-include(cmake/darktable-packaging.cmake)
+if("${CMAKE_BUILD_TYPE}" MATCHES "Release")
+    add_subdirectory(packaging)
+    include(cmake/darktable-packaging.cmake)
+endif("${CMAKE_BUILD_TYPE}" MATCHES "Release")
 
 # uninstall target
 configure_file(


parafin:
I wonder who uses that at all? Official Windows packaging?

Mark-64:
Official Windows packaging will use `-DCMAKE_BUILD_TYPE=Release` anyway, as well as people following the Windows build instruction.
This is only relevant for people like me who do some debugging under Windows.

parafin:
I mean do non-Windows developers use these features?

Mark-64:
I leave others to comment, but in case I could restrict the if statement to Windows OS only

GrahamByrnes:
@Mark-64 this seems like excellent news... so as a naive windows-builder: what build_type can I use once this is implemented? This means running from the darktable.exe in the ~/darktable/build folder ?

Mark-64:
@GrahamByrnes If you mean running from the .exe generated in the build folder without installing, no, it doesn't work. You still have to install DT somewhere (meaning you have to build the target "install", specifying `-DCMAKE_INSTALL_PREFIX=<your-install-folder>`. 
At least I couldn't manage to run DT in place without installing, but this is not a problem if you use an IDE, because the the IDE will not overwrite up-to-date files and it's very quick.

The PR is useful if you use a debugger and therefore you have to build with 
`-DCMAKE_BUILD_TYPE=Debug`. In that case the PR will avoid that CMake bundle utilities are run at the end of the build, which takes from 3 to 5 minutes at every build and it's not needed.

CMake bundle utilities are needed if you want to create a redistributable package with `--target package`, but in that case you will build with `-DCMAKE_BUILD_TYPE=Release`

Mark-64:
No love for this simple PR ? ;-)

TurboGit:
Fact is that I have next to zero expertise on CMake and frankly don't like CMake at all :) So I would prefer if @parafin would give the go on this or any other dev with strong knowledge on CMake.

parafin:
I'm not sure why this functionality was added in the first place, so we should take a look at git log and ask initial committer what's that code for. Also it's a question for current Linux and Windows packagers if they use it (macOS packaging doesn't). As for an argument that it doesn't concern packagers since they use Release build type - I for one don't for macOS builds, I use RelWithDebInfo. Which is BTW better option for debugging too (Debug is too slow).

So I can't give a go-ahead for this one.

Mark-64:
Found a better option, which is adding -DBUILD_MSYS2_INSTALL=ON to CMake command.
Closing the PR, thank you for looking into this

PR #4967, opened 2020-05-09
Overlay max size 2

Mark-64:
I repropose this PR, rebased for the new culling rewrite.
It does not affect LoDPI screens but makes the maximum size of icons and text of thumbtable overlays slightly smaller and harmonized with the rest of the GUI.
Please see #4886 for screenshots

### DIFF
--- a/src/dtgtk/thumbnail.c
+++ b/src/dtgtk/thumbnail.c
@@ -1130,7 +1130,7 @@ static void _thumb_resize_overlays(dt_thumbnail_t *thumb)
     // stars + reject having a width of 2 * r1 and spaced by r1 => 18 * r1
     // colorlabels => 3 * r1 + space r1
     // inner margins are 0.045 * width
-    const float r1 = fminf((DT_PIXEL_APPLY_DPI(16.0f) + 4.0f) / 2.0f, 0.91 * width / 22.0f);
+    const float r1 = fminf((DT_PIXEL_APPLY_DPI(16.0f) + 4.0f) / 2.0f, 0.91f * width / 22.0f);
 
     // file extension
     gtk_widget_set_margin_top(thumb->w_ext, 0.5 * r1);
--- a/src/dtgtk/thumbnail.c
+++ b/src/dtgtk/thumbnail.c
@@ -1130,7 +1130,7 @@ static void _thumb_resize_overlays(dt_thumbnail_t *thumb)
     // stars + reject having a width of 2 * r1 and spaced by r1 => 18 * r1
     // colorlabels => 3 * r1 + space r1
     // inner margins are 0.045 * width
-    const float r1 = fminf(DT_PIXEL_APPLY_DPI(20.0f) / 2.0f, 0.91 * width / 22.0f);
+    const float r1 = fminf((DT_PIXEL_APPLY_DPI(16.0f) + 4.0f) / 2.0f, 0.91 * width / 22.0f);
 
     // file extension
     gtk_widget_set_margin_top(thumb->w_ext, 0.5 * r1);
@@ -1201,7 +1201,7 @@ static void _thumb_resize_overlays(dt_thumbnail_t *thumb)
     // we need to squeeze 5 stars + 1 reject + 1 colorlabels symbols on a thumbnail width
     // all icons having a width of 3.0 * r1 => 21 * r1
     // we want r1 spaces at extremities, after reject, before colorlables => 4 * r1
-    const float r1 = fminf(DT_PIXEL_APPLY_DPI(20.0f) / 2.0f, width / 25.0f);
+    const float r1 = fminf((DT_PIXEL_APPLY_DPI(16.0f) + 4.0f)/ 2.0f, width / 25.0f);
 
     // file extension
     gtk_widget_set_margin_top(thumb->w_ext, 0.5 * r1);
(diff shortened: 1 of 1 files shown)


TurboGit:
This works for me! I'll wait for @AlicVB review.

Mark-64:
I will try at some point to rewrite the geometry of buttons, but my confidence with the DT codebase is not yet at the level needed.
Plus it would be time consuming anyway, because as I said many of the cairo icons need rescale/redraw, as there are ad-hoc adjustments.
But this is a change very easy to revert, so in the meantime why not doing it?

Mark-64:
sorry I did close inadvertently

Mark-64:
I studied a bit the current button design, looking for a better solution for this PR. Here is what I have found:
 
Icon buttons are handled in button.c, togglebutton.c and recently in thumbnail_btn.c.
In the first two, inner margin (which determines the icon size) is managed with this statement 
`float f_border = ((flags & CPF_DO_NOT_USE_BORDER) ? 4.0 : 6.0);`
Inner margin is hardcoded to 4px when `CPF_DO_NOT_USE_BORDER` is true and 6px otherwise.
This is non DPI aware and make the GUI scale poorly for HiDPI displays.
In thumbnail_btn.c instead, inner margin is set to 0 when `CPF_DO_NOT_USE_BORDER` is true and 12.5% of the widget allocation otherwise. A different strategy that generates inconsistent icon size DPI scaling.

Now, both strategies look unsatisfactory to me, as they both ignore button padding style property in CSS and contain hardcoded, theme independent assumptions.

Therefore a possible solution would be simply to make the icon inside the button to behave like a child widget. This could be done in two ways:
1. simply forcing in the c code the icon drawing geometry to honor the button css padding
2. defining a real drawing area child widget in the button and draw the icon with respect to that. A bit more complex, but also more flexible as we would have another themable item in css

If we do this, keeping the flag `CPF_DO_NOT_USE_BORDER` makes little sense, as you could set all margins in CSS. Maybe we could keep it as a quick way to bypass css margins and use all the available allocation ?

Unfortunately all the above will require some significant reworking of CSS margins and padding, but I think it's necessary to get a more consistent GUI experience on all screens.

One last remark: all the above does not apply to buttons with label, which are handled differently and look already fully css themable to me. However there are branches of code in button.c and togglebutton.c which are supposed to deal with text buttons, but are no longer used (and most likely do not work anyway). We could remove those and simplify the code.

@TurboGit , @AlicVB , @Nilvus : your opinion ?

Nilvus:
@Mark-64: having hardcoded padding or margins is at the opposite of the darktable 3 Gui, so I'm on to remove them and let those tweaked only with CSS. That would be more consistent too. I just had like to see that changed before my actual complete CSS rework that it's almost done but that's the game!

Just to add a point for that: the CPF_DO_NOT_USE_BORDER has been removed for a widget (an arrow) I pointed on my PR and that @elstoc remove to fix a glitch on his #5014 PR so that's the way to go I think.

@TurboGit: if @Mark-64 make that thing, maybe it will be better to wait to merge my PR. I could after his work, test and update my CSS rework and put only one commit to adapt @Mark-64 work and last @elstoc work on toast message. We have time for that and the more cleaner the CSS will be, the more updates will be easier later. What do you think of that?

AlicVB:
@Mark-64 : Sounds good for me. Some question or remark (in order to include that to your implementation if it's possible) : 
- Are css padding/margin DPI aware ? If not, we may use "false" css padding/margin like what I've done for image margins in `thumbnails.c` `_event_image_draw fct`
- Do we prefer fixed paddings or percentage ones. If we opt for the second, we'll need to use "false" paddings anyway, as gtk doesn't seems to manage css percentages
- We may need to adapt a little bit thumbnails icons placement code, as there alignment are dependent of the "fixed" padding (I can do that if you want)
- If we opt for "false" padding, it may be simpler to use one unique widget. Otherwise, 2 widgets may be better... (but maybe more difficult to handle in the current code design)

Anyway, I like the idea. Don't hesitate to ask for help if needed.

@Nilvus : for postponing your PR, I'm not sure it's needed. You'll have to do the adaptation work anyway... but maybe I've missed something.

Mark-64:
Thank you for supporting the idea, I will start the implementation and keep you posted.
@AlicVB Thanks for the suggestions, I had already noticed your technique with margin%. Luckily I can test on both low and high DPI devices, so we'll see if that is needed. My plan is to start with the simpler one-widget approach and if that is too limiting, add the second.
Agree that postponing @Nilvus PR is not necessary. I will need to tweak CSS a bit just to test that everything works but eventually i don't think to include CSS in my PR, so we don't have conflicts.

Nilvus:
Ok, that was just if some padding or margins will need to be adjusted that I was proposing to postpone. Hope that this work will not need to much tweaks so.

It would be good anyway if you could list all icons that your work will affect (and especially for those who visual aspect/place will change on the UI, so I could then improve CSS if needed).

Mark-64:
I have a draft and it seems to work. BTW, it scales worderfully in HiDPI.
However, on top of the CSS rework, the side effect of removing hardcoded margins is that I need to adjust/rescale some of the cairo icons.
In fact, some of them draw outside the area they are supposed to. With hardcoded margins the problem is not visible, but now we can really have zero margin, if we want to, and those icons get clipped.
Take the "styles" icon, for example, before and after adjustment (I had to do myself this little utility to test)
image
the original version draws well beyond the icon allocation (the red lines).
This will keep me busy for a while

dterrahe:
@Mark-64 Just leaving the remark here (not worth filing a bug report as not that big a deal) that for color picker buttons (not quad) it is not possible to see the active state while hovering over the button. I.e. while the mouse is on the button, you can click away and the view will stay the same. Only when you move the mouse away can you see whether you left the button on or off.

I have no idea where this functionality lives (even though I went searching for it a bit) but probably you understand the code much better then I do!

Mark-64:
Continue in #5074 
closing here

PR #4974, opened 2020-05-10
preferences: add global font size setting

### LINKED ISSUE #3207: Add a font size preference
Too many users have asked for that and it's really too simple to achieve to keep ignoring it…

We need a global preference to setup the UI font size in pt units that will override both OS and theme CSS settings, instead of asking users to find their darktablerc file and edit it manually (assuming they know which line to edit).

You need to set it to `GtkWidget *main_window = dt_ui_main_window(darktable.gui->ui)` and then (I think) propagate it to `pango_layout_set_font_description(layout, darktable.bauhaus->pango_font_desc);`

elstoc:
add a new option to the general tab of the settings dialog to set the
font size for darktable

Resolves #3207.

### DIFF
--- a/src/gui/gtk.c
+++ b/src/gui/gtk.c
@@ -2631,9 +2631,11 @@ void dt_gui_load_theme(const char *theme)
   {
     //font name can only use period as decimal separator
     //but printf format strings use comma for some locales, so replace comma with period
-    const gchar *font_size = dt_util_dstrcat(NULL, _("%.1f"), dt_conf_get_float("font_size"));
-    const gchar *font_name = dt_util_dstrcat(NULL, _("Sans %s"), dt_util_str_replace(font_size, ",", ".")); 
+    gchar *font_size = dt_util_dstrcat(NULL, _("%.1f"), dt_conf_get_float("font_size"));
+    gchar *font_name = dt_util_dstrcat(NULL, _("Sans %s"), dt_util_str_replace(font_size, ",", ".")); 
     g_object_set(gtk_settings_get_default(), "gtk-font-name", font_name, NULL);
+    g_free(font_size);
+    g_free(font_name);
   }
 
   char path[PATH_MAX] = { 0 }, datadir[PATH_MAX] = { 0 }, configdir[PATH_MAX] = { 0 }, usercsspath[PATH_MAX] = { 0 };
--- a/src/gui/gtk.c
+++ b/src/gui/gtk.c
@@ -2627,10 +2627,13 @@ void dt_gui_load_theme(const char *theme)
   //set font size
   if(dt_conf_get_bool("use_system_font"))
     gtk_settings_reset_property(gtk_settings_get_default(), "gtk-font-name");
-  else {
-    gchar *font_name = dt_util_dstrcat(NULL, _("Sans %.1f"), dt_conf_get_float("font_size"));
+  else 
+  {
+    //font name can only use period as decimal separator
+    //but printf format strings use comma for some locales, so replace comma with period
+    const gchar *font_size = dt_util_dstrcat(NULL, _("%.1f"), dt_conf_get_float("font_size"));
+    const gchar *font_name = dt_util_dstrcat(NULL, _("Sans %s"), dt_util_str_replace(font_size, ",", ".")); 
     g_object_set(gtk_settings_get_default(), "gtk-font-name", font_name, NULL);
-    g_free(font_name);
   }
 
   char path[PATH_MAX] = { 0 }, datadir[PATH_MAX] = { 0 }, configdir[PATH_MAX] = { 0 }, usercsspath[PATH_MAX] = { 0 };
--- a/src/gui/gtk.c
+++ b/src/gui/gtk.c
@@ -2621,9 +2621,13 @@ void dt_gui_add_help_link(GtkWidget *widget, const char *link)
 // load a CSS theme
 void dt_gui_load_theme(const char *theme)
 {
+  if(!dt_conf_key_exists("use_system_font"))
+    dt_conf_set_bool("use_system_font", TRUE);
+
   //set font size
-  if(dt_conf_get_float("font_size") != 0.0f)
-  {
+  if(dt_conf_get_bool("use_system_font"))
+    gtk_settings_reset_property(gtk_settings_get_default(), "gtk-font-name");
+  else {
     gchar *font_name = dt_util_dstrcat(NULL, _("Sans %f"), dt_conf_get_float("font_size"));
     g_object_set(gtk_settings_get_default(), "gtk-font-name", font_name, NULL);
     g_free(font_name);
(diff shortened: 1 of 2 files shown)


Nilvus:
Just tested and it's unusable on my side. Tested 2 times by just pressing on + button. Each time, that make my system stuck. After some dozens of seconds, my system crash (first completely and so I need to restart my system ; second time only my Gnome Session crash and so I need to login back).

By the way, using this field is not, I think the best thing. Having to 0.0 system default then that means than clicking on + button would set to 1px. Not wanted. I think, we should separate those settings by adding a checkbox to set to default system size that hide beside the setting used for font size.
So when the system checkbox is checked, setting font size is grey and can't be changed. Unselect and then you can change setting and I would set to 12px or 13px default setting. That's force the user to start with recommended font size and limit users to make bad choices.

elstoc:
Yes that was a rather naive first attempt. Latest commit implements your suggestions and adds a more reasonable range to the font sizes.

Nilvus:
It's better but remains unusable. Uncheck system checkbox works and I 12.0 applied immediately. But field and button doesn't work:

- Changing manually font-size (for example from 12.0 to 14.0) doesn't work, even after pressing enter. Nothing change.
-Clicking on + still (and always reproducible) crash my gnome-shell session (after few seconds where the cursor become stuck and system unresponsive. I've never had that issue with darktable before.

parafin:
Maybe it's better to apply this setting only after dt restart.

elstoc:
> But field and button doesn't work

I can't reproduce your crash - it works fine for me.

Nilvus:
But I have it. Don't know what I can say or do here...

elstoc:
I don't know. Difficult to investigate if I can't reproduce it. Anyone else able to test and gets the issue?

elstoc:
@Nilvus I've pushed a minor change in the hope that it might fix your issue (though I'm not all that hopeful!). Please could you retest.

Nilvus:
@elstoc: that fix the crash but changing font size has a strange behavior now. Default font size (12) is ok but after that, whatever other size I write, the size is saved on field after pressing enter (if I only write 14, after pressing enter, 14.0 is written) but nothing change. The UI remains on default font size. Then if I click on + or - buttons, that all time reduce a lot (too much) font size. This work like darktable reduce it's font size to the minimum it can, each time I press - or +.

See below, how the UI is rendered after pressing + or -:

elstoc:
Sorry I can't reproduce that issue either. + and - both work for me as does typing a value directly

elstoc:
I've pushed another try!

Nilvus:
Same issue, your last commit change nothing. Sorry.

TurboGit:
@elstoc : I'm wondering if this is not a coma vs dot issue. As @Nilvus I'm having trouble to set the font size and I found some pattern with different decimal number. As for Nilvus we are using coma as separator in France. What is your separator ?

elstoc:
@TurboGit I hadn't considered that. I'm in the UK so using a dot.

GrahamByrnes:
Ah! My dt is set up with English as the language, but to enter a float via preferences, I have to use a comma. Which is one reason I went to a choice of string values. So the language setting is not entirely consistent...

(BTW, if you are in a coma you are in hospital with a tube in your trachea... but please don't check my use of double consonants in French :-) )

Nilvus:
Didn't think of that but that's the issue. I've tested by set UK format in Gnome preferences and with a dot on darktable, that's works absolutely great. Good catch @TurboGit.

elstoc:
Problem is that gtk stores and sets the font as a string (e.g. "Sans 10") so if you want to include decimals you need to use something locale-specific, and this PR needs decimals. So the question is how can you write a printf format string that is locale-specific when it comes to the decimal separator?

TurboGit:
Can't we just solve this by removing any decimal parts ? Do we need to have this precision for specifying a font ?

elstoc:
For me I really do want that level of control - I want something between 8 and 9 (8 is too small, 9 is too big). If I lose that level of precision I'm probably back to setting the dpi and I really do prefer to set the font. Seems like there must be a solution for this.

Nilvus:
> For me I really do want that level of control - I want something between 8 and 9 (8 is too small, 9 is too big). If I lose that level of precision I'm probably back to setting the dpi and I really do prefer to set the font. Seems like there must be a solution for this.

So, there's probably a way to set locale before and be sure same thing happens for users who use comma than one who use dot. Just an idea even if I don't know how you can do that.

TurboGit:
On there is code somewhere on darktable to handle this gracefully. This is using a scanf IIRC. Can't check right now.

elstoc:
I've pushed an attempt at a fix, though I'm only able to test that it still works for me (and not for the comma separator) as I can't work out how to easily change the decimal separator on my system. @Nilvus please can you test?

src/gui/gtk.c
```diff
@@ -2627,8 +2628,18 @@ void dt_gui_load_theme(const char *theme)
   //set font size
   if(dt_conf_get_bool("use_system_font"))
     gtk_settings_reset_property(gtk_settings_get_default(), "gtk-font-name");
-  else {
-    gchar *font_name = dt_util_dstrcat(NULL, _("Sans %.1f"), dt_conf_get_float("font_size"));
+  else 
+  {
+    //font name needs to include a comma or period as the decimal separator
+    //depending on locale setting
+    const gchar *font_size = dt_util_dstrcat(NULL, _("%.1f"), dt_conf_get_float("font_size"));
```
TurboGit, line 2635: Sorry I was't clear, this line create a float number with a comma or dot depending on the locale. In France this is generating a comma. And CSS only supports a . so the proper fix is to always do:

`      font_name = dt_util_dstrcat(NULL, _("Sans %s"), dt_util_str_replace(font_size, ",", "."));`

That is changing comma to dot. You can do that conditionally if you prefer but you must ensure that the font_name is always using a dot.

elstoc: Thanks @TurboGit I was just coming to that realisation (hard without testing myself). I'll try again!

elstoc:
Ok I've worked out how to test it myself now and confirmed that this works with commas and periods.

Nilvus:
@elstoc: that works great now. Thanks. All fixed and font size is smooth and clean. And I agree: having possibility to use first decimal is a good idea. Good work.

src/gui/gtk.c
```diff
@@ -2621,6 +2621,21 @@ void dt_gui_add_help_link(GtkWidget *widget, const char *link)
...
+  if(dt_conf_get_bool("use_system_font"))
+    gtk_settings_reset_property(gtk_settings_get_default(), "gtk-font-name");
+  else 
+  {
+    //font name can only use period as decimal separator
+    //but printf format strings use comma for some locales, so replace comma with period
+    const gchar *font_size = dt_util_dstrcat(NULL, _("%.1f"), dt_conf_get_float("font_size"));
+    const gchar *font_name = dt_util_dstrcat(NULL, _("Sans %s"), dt_util_str_replace(font_size, ",", ".")); 
+    g_object_set(gtk_settings_get_default(), "gtk-font-name", font_name, NULL);
+  }
```
TurboGit, line 2637: No consts above and you need:
```
g_free(font_size);
g_free(font_name);
```

And we are all set!

elstoc:
Ready to go

TurboGit:
Ok, working for me.

PR #4978, opened 2020-05-10
Change most iop modules with color pickers to dt_color_picker_new

dterrahe:
This should have no user visible or behaviour changes. Testing would be much appreciated!

There are a few remaining modules that have some additional features so I'm putting those in separate commits so that they could be reverted separately if needed. Those are:
basicadj
colorchecker
colorzones
rgb curve
rgb levels
and
tone curve

As mentioned in #4956, tone equaliser uses color picker buttons but not the same underlying functionality. Thanks to @moy for resolving its confusing behavior.


TurboGit:
I'm impressed ! Very well done. Nice clean-up.

PR #4980, opened 2020-05-10
generalize the use of dt_view_image(s)_to_act_on

### LINKED ISSUE #4292: Collapsed groups in lighttable mode possible regression
**Describe the bug**
It appears that some of the group functions are not handled properly in V3.

In lighttable mode when all groups are collapsed using G at top right, several functions including 'trash', 'move', 'copy', 'remove', 'duplicate', 'export' are only applied to the first member of the group rather than all images in the group.

Tagging appears to work ok, so attaching (or detaching) a tag to/from collapsed group affects all members. 

**To Reproduce**

- Use G at top right of lighttable view to 'collapse grouped images' 
- Select a group (which, for example, has 8 members) and click 'trash'
- A popup appears, saying 'do you really want to send 8 selected images to trash'.
- Click 'yes'.
- Only the first image is deleted leaving 7 not sent to trash. 
- This occurs every time.

Equivalent behaviour observed for remove, copy, move, duplicate, export.

**Expected behavior**

I would expect any of these functions applied to a collapsed group to behave the same way as if the group was expanded, all members were selected, then the function was applied. V2.6.x behaved  like that.

 - Darktable Version: 3.0.0
 - OS: Ubuntu 18.04

### LINKED ISSUE #4669: Style shortcuts in darkroom apply the style to the wrong image
This issue is similar to https://github.com/darktable-org/darktable/issues/4576 but for styles. It only happens when styles are applied via a shortcut, not when using the styles menu in darkroom.

Applying a style via shortcut will affect the selection instead of the current image. Chaging the current image with SPACE, BACKSPACE or clicks to the filmstrip doesn't change the selection. Furthermore, the thumbnails will not update so the change to the images is invisible until the images are revisited in darkroom.

To reproduce:

1. Create a style if you don't have any
2. Go to settings / shortcuts / global / styles and assign a shortcut to a style
3. Select an image in lighttable
4. Switch to darkroom ('D')
5. Switch to next image ('SPACE')
6. Press the style shortcut

Seemingly nothing happens but pressing BACKSPACE shows that the previous image was stealth-changed.

Expected: the current image changes unless user somehow specifically indicates a different intent.

 - Darktable Version: 3.1.0 + current master 3b7a2dd4
 - OS: Ubuntu 18.04
 - 
 -

AlicVB:
As said in some post here, that's the last part of my big lighttable (and a little more) rewrite.
fix #4292, fix #4669

### Why
The idea is to have an unique function to determine on which image we should act on and to use it everywhere, so the user experience is consistent.
There's 2 functions to do that : 
- `dt_view_get_images_to_act_on` (with an argument to say if we want "hidden because grouped" images or not) => we get a list of images
- `dt_view_get_image_to_act_on` => we get the first image to act on

### How they work
I just make a copy/paste of the comment in the code.
Note that active image(s) are the one(s) shown on the center view, except for lighttable filemanager/zoomable where there's none.
- `dt_view_get_images_to_act_on`
```
/** Here's how it works
   *
   *             mouse over| x | x | x |   |   |
   *     mouse inside table| x | x |   |   |   |
   * mouse inside selection| x |   |   |   |   |
   *          active images| ? | ? | x |   | x |
   *                       |   |   |   |   |   |
   *                       | S | O | O | S | A |
   *  S = selection ; O = mouseover ; A = active images
   *  the mouse can be outside thumbtable in case of filmstrip + mouse in center widget
   *
   *  if only_visible is FALSE, then it will add also not visible images because of grouping
   **/
```
- `dt_view_get_image_to_act_on`
```
/** Here's how it works -- same as for list, except we don't care about mouse inside selection or table
   *
   *             mouse over| x |   |   |
   *          active images| ? |   | x |
   *                       |   |   |   |
   *                       | O | S | A |
   *  First image of ...
   *  S = selection ; O = mouseover ; A = active images
   **/
```

### All images or only the visible ones ?
Here's the current (with this PR) state of different actions (I've try to keep the same behavior as before)
```
dt_control_gpx_apply_job_create     => visible
dt_control_time_offset_job_create   => visible
dt_control_copy_images              => all
dt_control_delete_images            => all
dt_control_duplicate_images         => visible
dt_control_flip_images              => visible
dt_control_merge_hdr                => visible
dt_control_refresh_exif             => all
dt_control_remove_images            => all
dt_control_reset_local_copy_images  => all
dt_control_set_local_copy_images    => all
dt_control_write_sidecar_files      => all
export                              => visible
LUA act_on_cb                       => all
metadata paste/clear                => all
history compress/delete/paste       => visible
lt styles create/apply              => visible
lt metadata                         => all
lt ratings/colorlabels              => all
```
Please stand up if you see some error, omission, or whatever...

### Other changes
- This is used too for views entering (darkroom, maps, print, slideshow)
- In the metadata lib, I've removed the "save on exit" because it was completely buggy even without my changes : the text is applied to the image hovered when you quit the lib... not what is expected at all !

### Finally : 
It would be great if we have lot of testers for this, as this is a delicate area.

### DIFF
--- a/src/common/image.c
+++ b/src/common/image.c
@@ -441,7 +441,7 @@ void dt_image_set_location(const int32_t imgid, const dt_image_geoloc_t *geoloc,
 {
   GList *imgs = NULL;
   if(imgid == -1)
-    imgs = dt_view_get_images_to_act_on();
+    imgs = dt_view_get_images_to_act_on(TRUE);
   else
     imgs = g_list_append(imgs, GINT_TO_POINTER(imgid));
   dt_image_set_locations(imgs, geoloc, undo_on, group_on);
@@ -2104,7 +2104,7 @@ void dt_image_synch_xmp(const int selected)
   }
   else
   {
-    GList *imgs = dt_view_get_images_to_act_on();
+    GList *imgs = dt_view_get_images_to_act_on(FALSE);
     dt_image_synch_xmps(imgs);
     g_list_free(imgs);
   }
--- a/src/common/metadata.c
+++ b/src/common/metadata.c
@@ -523,7 +523,7 @@ void dt_metadata_set(const int imgid, const char *key, const char *value, const 
   {
     GList *imgs = NULL;
     if(imgid == -1)
-      imgs = dt_view_get_images_to_act_on();
+      imgs = dt_view_get_images_to_act_on(TRUE);
     else
       imgs = g_list_append(imgs, GINT_TO_POINTER(imgid));
     if(imgs)
@@ -620,7 +620,7 @@ void dt_metadata_set_list(const int imgid, GList *key_value, const gboolean undo
   {
     GList *imgs = NULL;
     if(imgid == -1)
-      imgs = dt_view_get_images_to_act_on();
+      imgs = dt_view_get_images_to_act_on(TRUE);
     else
       imgs = g_list_append(imgs, GINT_TO_POINTER(imgid));
     if(imgs)
@@ -667,7 +667,7 @@ void dt_metadata_clear(const int imgid, const gboolean undo_on, const gboolean g
   {
     GList *imgs = NULL;
     if(imgid == -1)
-      imgs = dt_view_get_images_to_act_on();
+      imgs = dt_view_get_images_to_act_on(TRUE);
     else
       imgs = g_list_append(imgs, GINT_TO_POINTER(imgid));
     if(imgs)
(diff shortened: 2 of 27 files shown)


TurboGit:
I did some tests and it seems to be working ok to me.

I found a little glitch not related to this. So I'll merge this to ensure more field testing and will create a new issue about the little glitch.

TurboGit:
See #4987

PR #4983, opened 2020-05-11
export metadata fix - allow to export multiple values for the same category

### LINKED ISSUE #4964: metadata export: export all values of $(CATEGORYn(category)) when there are several tags at level n
The variable `$(CATEGORYn(category))` gives some great possibilities to use tags in other fields using "metadata export" in export module (examples in the [manual](https://darktable.gitlab.io/doc/en/export_selected.html)). 

When several tags (at the same level) are defined under a category, only one value is exported from `$(CATEGORYn(category))`. 
For example, if you have an image with tags "People|Bob" and "People|John", and define a formula `Xmp.iptcExt.PersonInImage=$(CATEGORY0(People))`, the exported file will only have "John" in the `Xmp.iptcExt.PersonInImage` field but not "Bob".  

I think it would make sense to export all tags of level `n` when several tags are defined.

There are probably different technical cases depending on the type of the target field: 
- For the example above, `Xmp.iptcExt.PersonInImage` is a XmpBag type, like tags when they are exported in `dc:subject`, so it could follow the same structure: 
```
<dc:subject> 
    <rdf:Bag> 
        <rdf:li>Bob</rdf:li> 
        <rdf:li>John</rdf:li>
    </rdf:Bag> 
</dc:subject>
<iptcExt:PersonInImage> 
    <rdf:Bag> 
        <rdf:li>Bob</rdf:li> 
        <rdf:li>John</rdf:li>
    </rdf:Bag> 
</iptcExt:PersonInImage>
```
- When the target field is not a list but a text, it raises the question of how it should be done (maybe something like `TextField=tag1, tag2, tag3` if it is possible)

**Alternatives**

**Additional context**
darktable 3.0.2

phweyland:
fixes #4964 

When an image is tagged with several people the list of people can be exported into a single tag.
If the tag is an xmpBag or an xmpSeq the values are added separately.
If not, an xmpText or a string for example, the values are added as a single string with different values comma separated.

@TurboGit, the exiv2 tags list is now a permanent list. Please tell me if there is better way to implement this.

### DIFF
--- a/src/common/exif.cc
+++ b/src/common/exif.cc
@@ -76,6 +76,9 @@ extern "C" {
 
 #define DT_XMP_EXIF_VERSION 4
 
+// persistent list of exiv2 tags. set up in dt_init()
+static GList *exiv2_taglist = NULL;
+
 static const char *_get_exiv2_type(const int type)
 {
   switch(type)
@@ -154,11 +157,12 @@ static void _get_xmp_tags(const char *prefix, GList **taglist)
   }
 }
 
-GList *dt_exif_get_exiv2_taglist()
+void dt_exif_set_exiv2_taglist()
 {
+  if(exiv2_taglist) return;
+
   Exiv2::XmpParser::initialize();
   ::atexit(Exiv2::XmpParser::terminate);
-  GList *taglist = NULL;
 
   try
   {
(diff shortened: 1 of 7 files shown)


src/common/exif.cc
```diff
@@ -154,11 +154,14 @@ static void _get_xmp_tags(const char *prefix, GList **taglist)
   }
 }
 
-GList *dt_exif_get_exiv2_taglist()
+static GList *taglist;
```
TurboGit, line 157: You're missing a = NULL there otherwise we will be getting random memory value and possible crash.

TurboGit: I also think that such static persistent storage variable should be placed at the start of the file with comments about the persistency and the fact that it is initialize early in dt_init()

phweyland: I had read that static and global variable were always initialized to 0 or NULL...
But you are right, it's more secure with an explicit value.
I'll do

phweyland:
done

TurboGit:
Looks good to me. Thanks!

PR #4632, opened 2020-04-07
WB Presets for Pentax K-1 firmware ver. 1.52

johnny-bit:
Sample set contributed by Thomas_Gulmark via pixls.us

PR #4995, opened 2020-05-11
Remove old WB Presets for Pentax K-1

gribouilleuse:
After another PR (https://github.com/darktable-org/darktable/pull/4632) there are two presets for Pentax K-1. This PR is to remove old presets.


TurboGit:
Good catch !

PR #4996, opened 2020-05-11
Add a ImageMagick loader backend (compatible v6 and v7)

mazhe:
Hi,

Sometime having support for the formats allowed by the graphicmagick loader support is nice, but I thought direct imagemagick support would be nice on platform other than debian.

Would you consider including this backend? I've been using it for some time without issues and tought it could be used in the wild.

Obviously, it's a bad idea to try to enable both magick loaders which have a lot of similar symbol graphicmagick being a fork of imagemagick...

Have a nice day,


TurboGit:
Can you clarify why one would want to keep v6 ? Why not just use this new version and then remove the old graphicmagic loader?

mazhe:
Personnaly, I would not need it -- I began to write the loader at a time where I would occasionaly have a system with v6 but that is not the case now. I was submitting a compatible patch for completeness in case others would need it.

Maybe I could add a commit to remove v6 support, but if needed it would be in history?

TurboGit:
@mazhe : Yes please add a commit to remove GraphicMagic. Just to be sure v7 supports all format supported by v6, right ?

mazhe:
Did you mean remove ImageMagick V6, not GraphicMagic?

No issues on the v7 format support front, despite the big version jump, they did not rewrite everything and support is equal to v6

TurboGit:
@mazhe : Maybe I'm confused. You patch is for ImageMagic v6 and v7. That's fine.

I'm just proposing to remove the olf GraphicMagic support which stop at v4. Please clarifies if this is not correct I don't have a clear view on the difference between GraphicMagic and ImageMagic.

TurboGit:
> Did you mean remove ImageMagick V6, not GraphicMagic?

No, my idea was to remove GraphicMagic loader.

mazhe:
Ok, without going too much in the (sometime messy) history:
- imagemagick was the original project
- in 2001, a project graphicmagick was created due to differences about new APIs (https://en.wikipedia.org/wiki/ImageMagick#Related_software)
- graphicmagick was picked often as the default "magick" package by debian and debian-deritative distributions
- to my knowledge, both projects are sill maintained

My aim was not to advocate one over the other -- of course removing graphicmagick support would not affect me, but that could be an issue for others?

TurboGit:
I see, no clear idea either. Let's add support for ImageMagic (your PR) but do not activate it by default in the CMake. So it would be there and could be added for people recompiling dt for now. At some point in the future we will switch the defaults for GraphicMagic and ImageMagic. How does that sound?

mazhe:
I'll commit v6 support removal (with still a check at cmake-time) after a rebuild just to be sure everything runs fine :)

DefineOptions.cmake
```diff
@@ -11,6 +11,7 @@ option(USE_UNITY "Use libunity to report progress in the launcher" OFF)
 option(USE_OPENMP "Use openmp threading support." ON)
 option(USE_OPENCL "Use OpenCL support." ON)
 option(USE_GRAPHICSMAGICK "Use GraphicsMagick library for image import." ON)
+option(USE_IMAGEMAGICK "Use ImageMagick library for image import." ON)
```
TurboGit, line 14: Please set this to OFF for now as discussed (or again something not clear on my side:). For now devs needed it will activate this option.

mazhe:
Right! This is fixed (note that when graphicmagick loader is selected, imagemagick won't be build as they would conflict)

TurboGit:
Looks ok to me now. Thanks.

PR #4997, opened 2020-05-11
crop & rotate: fix cursor setting issues

elstoc:
fixes issues where the crop & rotate cursor was not being set correctly,
for example:

   - where the cursor is changed by another application (alt-tabbing
     away from darktable)
   - where the cursor is changed by hovering over a sidebar or the
     top/bottom panels


TurboGit:
I'm not able to reproduce the issue, but the change seems good to me. Thanks.

PR #4998, opened 2020-05-11
ensure module focus indicator updated when module gains/loses focus

### LINKED ISSUE #4982: crop & rotate shows as having focus when changes committed
#4927 introduced a change to make the currently in-focus module have a different appearance. 

When you double-click on the image to commit changes in the crop & rotate module, the module loses focus (calls ``dt_iop_request_focus(NULL)``). However, the indicator still suggests that the module has focus.

Ideally all calls to ``dt_iop_request_focus`` should ensure that the indicator is set/unset appropriately.

elstoc:
Resolves #4982


TurboGit:
Works for me !

PR #4999, opened 2020-05-11
culling fixes

### LINKED ISSUE #4965: Selection broken (?) in culling
**Describe the bug**

Viewing images in culling does not seem to allow a way to select an image.

Clicking on the image does not select it... hovering the mouse seems to, but if I try to go to darkroom, that is lost and I get "no image selected". same applies for the thumbnails: hovering selects, but clicking does not.

I can double-click to go to dr, but there are also issues with eg deleting or removing: if an image is selected in file mode, then I go to culling and the selected image is off screen, it is still that image that is selected even if I click on one on screen. Particularly dangerous when the tool-tip says "1 image selected."

There is also the issue of returning from dr and finding that a different image is selected (possibly related to #4911 & #4957), and also the return of an old issue: deleting the last image on a film strip when viewing 1 image / screen leaves no way to get back to images.

**Expected behavior**

Clicking on an image in culling view should select that image. Deleting an image should not leave the screen blank while there are still images in the fim strip.

dt3.1.0+1481 Win 10 release 190318-1202

### LINKED ISSUE #4987: culling / selection inconsistencies
**Describe the bug**
Selections between Ctrl-X and X not consistent.

**To Reproduce**

1. Go to lighttable
2. Select two images say A & B
3. Over on a third one, say C
4. Do a `Ctrl+X`, both A & B are displayed, OK
5. Do not move mouse, stay over C
6. Do a X, then C and following pictures are displayed, OK
7. Still not moving, do again a Ctrl+X, and we get the same images displayed as in step 6

**Expected behavior**

In step 7 I would have expected to have A & B displayed as in step 4.

 - Darktable Version: master
 - OS: Linux

**Additional context**
- Not related to #4980

### LINKED ISSUE #4991: Delopement version DT crash in thumbtable.c
After pasting history on a thumbnail in filmstrip, I realized it was in culling mide, when I changed it to filemanager DT crashed.
Other things I noticed, the pasting wasn't always working.

this is darktable 3.1.0+1512~gba0fbe0d4 reporting an exception:

-------------------

Error occurred on Monday, May 11, 2020 at 15:53:25.

darktable.exe caused an Access Violation at location 0000000063705B87 in module libdarktable.dll Reading from location 0000000000000000.
```
AddrPC           Params
0000000063705B87 000000002F926D30 000000002F887E80 00000000FFFFFFFF  libdarktable.dll!_filemanager_ensure_rowid_visibility  
[F:/msys64/home/chill/darktable/src/dtgtk/thumbtable.c @ 2047]
  2045:   if(rowid < 1) return FALSE;
  2046:   // get first and last fully visible thumbnails
> 2047:   dt_thumbnail_t *first = (dt_thumbnail_t *)g_list_first(table->list)->data;
  2048:   const int pos = MIN(g_list_length(table->list) - 1, table->thumbs_per_row * (table->rows - 1) - 1);
  2049:   dt_thumbnail_t *last = (dt_thumbnail_t *)g_list_nth_data(table->list, pos);
0000000063705070 000000003EADB990 0000000068DD2ACF 000000000061E0F8  libdarktable.dll! ??   [F:/msys64/home/chill/darktable/src/dtgtk/thumbtable.c @ 1679]
  1677:       // we need to ensure they are visible and to mark them with some css effect
  1678:       const int lastid = GPOINTER_TO_INT(g_slist_last(darktable.view_manager->active_images)->data);
> 1679:       dt_thumbtable_ensure_imgid_visibility(table, lastid);
  1680: 
  1681:       GSList *l = darktable.view_manager->active_images;
00000000637063CC 00000000109A8BA0 0000000000000002 000000000061E040  libdarktable.dll! ??   [F:/msys64/home/chill/darktable/src/dtgtk/thumbtable.c @ 862]
   860:   // but we don't really want to draw something, this is just to know when the widget is really ready
   861:   dt_thumbtable_t *table = (dt_thumbtable_t *)user_data;
>  862:   dt_thumbtable_full_redraw(table, FALSE);
   863:   return FALSE; // let's propagate this event
   864: }
```

AlicVB:
fix various glitch in culling mode (see commits msg). Some of them were already present before the rewrite
fix #4991 fix #4987 fix #4965

### DIFF
--- a/src/dtgtk/thumbtable.c
+++ b/src/dtgtk/thumbtable.c
@@ -2045,6 +2045,7 @@ void dt_thumbtable_update_accels_connection(dt_thumbtable_t *table, const int vi
 static gboolean _filemanager_ensure_rowid_visibility(dt_thumbtable_t *table, const int rowid)
 {
   if(rowid < 1) return FALSE;
+  if(!table->list || g_list_length(table->list) == 0) return FALSE;
   // get first and last fully visible thumbnails
   dt_thumbnail_t *first = (dt_thumbnail_t *)g_list_first(table->list)->data;
   const int pos = MIN(g_list_length(table->list) - 1, table->thumbs_per_row * (table->rows - 1) - 1);
@@ -2069,6 +2070,7 @@ static gboolean _filemanager_ensure_rowid_visibility(dt_thumbtable_t *table, con
 static gboolean _zoomable_ensure_rowid_visibility(dt_thumbtable_t *table, const int rowid)
 {
   if(rowid < 1) return FALSE;
+  if(!table->list || g_list_length(table->list) == 0) return FALSE;
 
   int minrowid = 0;
   int maxrowid = 0;
--- a/src/dtgtk/thumbtable.c
+++ b/src/dtgtk/thumbtable.c
@@ -1237,13 +1237,15 @@ static void _dt_collection_changed_callback(gpointer instance, dt_collection_cha
     }
 
     // get the new rowid of the new offset image
+    const int nrow = _thumb_get_rowid(newid);
+    const gboolean offset_changed = (nrow != table->offset);
     table->offset_imgid = newid;
-    table->offset = _thumb_get_rowid(newid);
-    dt_conf_set_int("plugins/lighttable/recentcollect/pos0", table->offset);
+    table->offset = nrow;
+    if(offset_changed) dt_conf_set_int("plugins/lighttable/recentcollect/pos0", table->offset);
 
     dt_thumbtable_full_redraw(table, TRUE);
 
-    dt_view_lighttable_change_offset(darktable.view_manager, FALSE, newid);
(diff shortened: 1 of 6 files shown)


src/dtgtk/thumbtable.c
```diff
@@ -2043,6 +2045,7 @@ void dt_thumbtable_update_accels_connection(dt_thumbtable_t *table, const int vi
 static gboolean _filemanager_ensure_rowid_visibility(dt_thumbtable_t *table, const int rowid)
 {
   if(rowid < 1) return FALSE;
+  if(!table->list || g_list_length(table->list) == 0) return FALSE;
```
TurboGit, line 2048: The first tests is superfluous as g_list_length (as most -all?- g_ routines are coping with NULL). But I'll merge as-is, that was just a remark.

TurboGit:
Looks good to me. Thanks.

PR #5001, opened 2020-05-11
slideshow : fix start point if grouped images

AlicVB:
currently if we have grouped images before the starting point, the start image is not the right one...


TurboGit:
Good catch! This works for me. Thanks.

Issue #3740, opened 2019-12-13
free the top bar in darkroom, replace it with a menu


aurelienpierre:
Right now, the top bar in darkroom contains the same filtering options as the lighttable, which are only useful to fetch images for the filmstrip at the bottom.

As many people, having a 16:9 screen, I try to free a maximum amount of vertical space so the filmstrip is almost always collapsed, as the header panel, and I don't remember once using the filtering options in darkroom (if I need to see the other pictures in the collection, it will be the collection set in lighttable, I don't need to change it in darkroom).

The problem is, when using drawn masks, I never remember which of Ctrl/Shitf/Alt controls opacity, size, orientation or feathering, and these hints appear in my collapsed header. This is awful ergonomy TBH. Also, trying to increase the feathering size of a path with mouse wheel often ends up with dragging it away or increasing its size.

I was using Krita the other day, and they deal with that pretty cleverly:
![Capture d’écran du 2019-12-13 17-06-32](https://user-images.githubusercontent.com/2779157/70814037-f3da7100-1dca-11ea-833a-3bdeab49f4bc.png)

What if we ditch the filtering options in darkroom top bar and replace them with sliders for masks opacity, size, feathering and orientation (for the liquify tool), possibly mask locking options for size or position ?

ptilopteri:
I support this.  I try to utilize all the darkroom screen space for editing.  All except the right panel are collapsed/hidden except for rare occasions.

ArJo45:
This is a very good idea. Left and right thumbs up.

junkyardsparkle:
I like it. There have been several issues reported lately involving the questionable placement of the opacity information, and gaining vertical screen space is always nice.

AlicVB:
For my 2 cents, we should rethink a little more tools placements for darkroom top and bottom panels at least. We have several tools more or less useful for darkroom, but they are placed in different parts of the ui, requiring to let many panels open.
For my part, I would like to have a unique (bottom or top, whatever) bar with that inside (not ordered)  : 
- clipping, etc... buttons
- prefs button
- ?? new masks tool ??
- hint box

For the other tools : 
- the info line can be put in the right panel...
- as said, grouping, sorting, etc... are not really useful in darkroom
- styles button :  why not adding it in history lib, that would centralize styles actions

TurboGit:
[...]

PR #5003, opened 2020-05-11
add new function dt_toast_log for accels and opacity

elstoc:
adds a new function dt_toast_log that replaces dt_control_log for
messages that need to display briefly (e.g. dynamic accelerators)

show mask opacity using new function when amended with CTRL+Scroll as
well as when changing sliders and comboboxes with keyboard shortcuts

### DIFF
--- a/src/control/control.c
+++ b/src/control/control.c
@@ -57,6 +57,11 @@ void dt_control_init(dt_control_t *s)
   s->log_message_timeout_id = 0;
   dt_pthread_mutex_init(&(s->log_mutex), NULL);
 
+  s->toast_pos = s->toast_ack = 0;
+  s->toast_busy = 0;
+  s->toast_message_timeout_id = 0;
+  dt_pthread_mutex_init(&(s->toast_mutex), NULL);
+
   pthread_cond_init(&s->cond, NULL);
   dt_pthread_mutex_init(&s->cond_mutex, NULL);
   dt_pthread_mutex_init(&s->queue_mutex, NULL);
@@ -173,6 +178,7 @@ void dt_control_cleanup(dt_control_t *s)
   dt_pthread_mutex_destroy(&s->queue_mutex);
   dt_pthread_mutex_destroy(&s->cond_mutex);
   dt_pthread_mutex_destroy(&s->log_mutex);
+  dt_pthread_mutex_destroy(&s->toast_mutex);
   dt_pthread_mutex_destroy(&s->res_mutex);
   dt_pthread_mutex_destroy(&s->run_mutex);
   dt_pthread_mutex_destroy(&s->progress_system.mutex);
@@ -381,6 +387,16 @@ static gboolean _dt_ctl_log_message_timeout_callback(gpointer data)
   return FALSE;
 }
 
+static gboolean _dt_ctl_toast_message_timeout_callback(gpointer data)
+{
+  dt_pthread_mutex_lock(&darktable.control->toast_mutex);
+  if(darktable.control->toast_ack != darktable.control->toast_pos)
+    darktable.control->toast_ack = (darktable.control->toast_ack + 1) % DT_CTL_TOAST_SIZE;
+  darktable.control->toast_message_timeout_id = 0;
+  dt_pthread_mutex_unlock(&darktable.control->toast_mutex);
+  dt_control_toast_redraw();
+  return FALSE;
+}
 
 void dt_control_button_pressed(double x, double y, double pressure, int which, int type, uint32_t state)
 {
@@ -411,6 +427,22 @@ void dt_control_button_pressed(double x, double y, double pressure, int which, i
     }
   dt_pthread_mutex_unlock(&darktable.control->log_mutex);
 
+  // ack toast message:
+  dt_pthread_mutex_lock(&darktable.control->toast_mutex);
+  if(darktable.control->toast_ack != darktable.control->toast_pos)
+    if(which == 1 /*&& x > xc - 10 && x < xc + 10*/ && y > yc - 10 && y < yc + 10)
+    {
+      if(darktable.control->toast_message_timeout_id)
+      {
+        g_source_remove(darktable.control->toast_message_timeout_id);
+        darktable.control->toast_message_timeout_id = 0;
+      }
+      darktable.control->toast_ack = (darktable.control->toast_ack + 1) % DT_CTL_TOAST_SIZE;
+      dt_pthread_mutex_unlock(&darktable.control->toast_mutex);
+      return;
+    }
+  dt_pthread_mutex_unlock(&darktable.control->toast_mutex);
+
   if(!dt_view_manager_button_pressed(darktable.view_manager, x, y, pressure, which, type, state))
     if(type == GDK_2BUTTON_PRESS && which == 1) dt_ctl_switch_mode();
 }
@@ -418,6 +450,7 @@ void dt_control_button_pressed(double x, double y, double pressure, int which, i
 static gboolean _redraw_center(gpointer user_data)
 {
   dt_control_log_redraw();
+  dt_control_toast_redraw();
   return FALSE; // don't call this again
 }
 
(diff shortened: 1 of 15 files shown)


TurboGit:
I don't think we should go further on those toast messages. They are getting annoying over the image being worked on.

I think we should sit down and design a proper way to display information to users. The discussion was already started some time ago, don't have the ref right now.

Fact is on darktable everything can be collapsed, so it is hard to find a stable area on the interface where we are sure the message will be seen. I don't have a proposal but I feel that a toast message should be displayed on last resort if the message area is not visible for example.

TurboGit:
Part of the discussion was here #3740.

elstoc:
> it is hard to find a stable area on the interface

I agree that with everything collapsed there is no good answer to this but for a keyboard/mouse action to have potentially very little noticeable effect with no feedback is a shame and this one has lead to incidents being raised. I like the idea of only popping this up if the top bar is closed and am happy to make that change if you think it's better (and would merge it). Then it would work similarly to the dynamic accelerator toasts (which only show if the module is not visible). In the other issue I also proposed the possibility of allowing the user to set the position of the hinter message similar to how they can reposition the info string.

Of course we can always remove these toast messages in the future if a better solution is found. However I fear that the above issue might just stagnate because nobody can find a good answer for when everything is collapsed.

elstoc:
For now, though, perhaps the least-worst option might be to move the toast to the top of the image (slightly more out of the way for most people?) and style it so it's a bit less imposing - get rid of the grey background and just have it as text on the image - similar to the "working..." text that appears when the image is being redrawn.

TurboGit:
Not sure, I fear we will go back and forth on this so I'd like to have a better view. Please open an issue for this to be discussed and define a clear goal.

We cannot reduce the timeout as when exporting, printing... the message displayed is long and we need some time to read it.

What we could do right now is:
- define a new routine dt_toast_log()
- this routine could have the same implementation as dt_control_log() with a smaller timeoout.
- use dt_toast_log() everywhere a short message (dynamic key, controlling opacity...) is displayed

Then later when we will be clear on which part and/or how we want this to be displayed we will have a single routine to rewrite.

One idea, but to be discussed don't rush on that... We could write those small toast message at the top of the mouse cursor ?

How does that sound like a first approach ?

elstoc:
> What we could do right now is...

Sounds like a reasonable approach. Are you suggesting I do that _now_ and then we discuss a longer-term solution on the back of an issue? Or just as a first idea for a longer-term solution?

elstoc:
> We could write those small toast message at the top of the mouse cursor

That's good for this scenario (when changing mask opacity) but I'm not sure for dynamic accels, where the mouse pointer might not be on the image.

TurboGit:
> > What we could do right now is...
> 
> Sounds like a reasonable approach. Are you suggesting I do that _now_ and then we discuss a longer-term solution on the back of an issue? Or just as a first idea for a longer-term solution?

I would say now. This should is easy, will help to identify all point where small toast messages are displayed and hence will help later to examine current situation.

elstoc:
> I would say now.

Ok will do. I'll make it look like my example above to start with, to keep it completely out of the way of the log toast. I'll do a separate PR and then amend this one to use it after if that's ok with you.

TurboGit:
Maybe not with the transparency, it is too hard to read I think.

elstoc:
> Maybe not with the transparency, it is too hard to read I think

Fine if you like - that's just CSS and can be changed later. I can increase the opacity to make it easier to read (it's not entirely transparent) or just leave it as solid grey.

elstoc:
I see you've tagged this as "feature: redesign". I'm not clear if you'd prefer me to include the new dt_toast_log() in this PR or another separate one?

TurboGit:
Well let's include this here.

elstoc:
Ok this is updated now. I've left the box slightly transparent but not so much as before, but happy to change it back if you prefer.

src/control/control.c
```diff
@@ -57,6 +57,11 @@ void dt_control_init(dt_control_t *s)
   s->log_message_timeout_id = 0;
   dt_pthread_mutex_init(&(s->log_mutex), NULL);
 
+  s->toast_pos = s->toast_ack = 0;
```
TurboGit, line 60: You've done a completely different implementation. I was just expecting a new routine, the log_mutex could be reused, no? I don't expect a toast message and a log at the same time anyway? Or maybe this is necessary for a reason I don't see?

TurboGit: Well, I'm not sure I want something :) I'm discovering this as we exchange. My idea was to have a minimal set of changes to ensure we could change this later (after discussion) more easily. My initial idea was just to define a dt_loast_lot() using almost the same code but placing the output at a different position and with a smaller timeout. But again I have no strong opinion on all this at this time.

src/develop/masks/masks.c
```diff
@@ -2409,6 +2411,8 @@ void dt_masks_form_change_opacity(dt_masks_form_t *form, int parentid, int up)
     {
       const float opacity = CLAMP(fpt->opacity + amount, 0.05f, 1.0f);
       fpt->opacity = opacity;
+      const int opacitypercent = opacity * 100;
+      dt_toast_log(_("Opacity: %d%%"), opacitypercent);
```
TurboGit, line 2415: No capital letter on current style.

src/develop/masks/masks.c
```diff
@@ -1727,6 +1727,8 @@ int dt_masks_events_mouse_scrolled(struct dt_iop_module_t *module, double x, dou
 
       opacity = CLAMP(opacity + amount, 0.05f, 1.0f);
       dt_conf_set_float("plugins/darkroom/masks/opacity", opacity);
+      const int opacitypercent = opacity * 100;
+      dt_toast_log(_("Opacity: %d%%"), opacitypercent);
```
TurboGit, line 1731: No capital letter on current style.

src/control/control.c
elstoc: The way dt_control_log works I don't think you could easily reposition the output without a separate implementation (or without repositioning the control log too), because the functionality that sets the log text and the functionality that draws it on the screen are separated. I've very much discovered after the fact that what I did might have been a good idea, since I can see the accel toast text potentially overwriting some important error message and I think we do want error text staying on screen.

TurboGit: Ok, let's keep this implementation. I have tested it and it is anyway far better than the previous toast as less visible. Please fix the case and I'll merge. Thanks again for all the work!

elstoc: All done and no problem - I'm enjoying myself - I don't get to do nearly enough coding in my day job any more.

src/views/lighttable.c
```diff
@@ -1184,6 +1184,8 @@ void gui_init(dt_view_t *self)
   gtk_overlay_add_overlay(GTK_OVERLAY(dt_ui_center_base(darktable.gui->ui)), lib->preview->widget);
   gtk_overlay_reorder_overlay(GTK_OVERLAY(dt_ui_center_base(darktable.gui->ui)),
                               gtk_widget_get_parent(dt_ui_log_msg(darktable.gui->ui)), -1);
+  gtk_overlay_reorder_overlay(GTK_OVERLAY(dt_ui_center_base(darktable.gui->ui)),
```
TurboGit, line 1187: Can we really have toast message on the lighttable ?

Same question for the map below ?

src/dtgtk/thumbtable.c
```diff
@@ -1793,6 +1793,8 @@ void dt_thumbtable_set_parent(dt_thumbtable_t *table, GtkWidget *new_parent, dt_
       {
         gtk_overlay_reorder_overlay(GTK_OVERLAY(dt_ui_center_base(darktable.gui->ui)),
                                     gtk_widget_get_parent(dt_ui_log_msg(darktable.gui->ui)), -1);
+        gtk_overlay_reorder_overlay(GTK_OVERLAY(dt_ui_center_base(darktable.gui->ui)),
```
TurboGit, line 1796: As for map and lighttable we probably don't need this one.

src/views/lighttable.c
elstoc: Not right now but if someone decides to add dynamic accels to lib module sliders or comboboxes... Up to you I can remove it if you like.

src/dtgtk/thumbtable.c
elstoc: Similarly it's mostly there to future-proof (and because I basically did a straight duplication of the existing log messages) but I'll remove it if you want.

Nilvus:
I've tested this PR and why not. It's a start. About CSS part, this will need for me to add a commit after this one is merged to merge those new lines after where the log lines are now on the new CSS. But I also think the color we have to be tweaked for each time.

On dark theme, I think this is good:

#toast-msg
{
   margin-top: 2px;
   padding-bottom: 8px;
   background-color: rgba(40,40,40,0.7);
}

a little margin on top of the message and a padding bottom to center the text.

TurboGit:
Ok, let's merge this now, looks ok to me. CSS part will be reworked in @Nilvus  CSS PR.

PR #5004, opened 2020-05-11
variables list udpate - gtkentry & manual

phweyland:
update gtkentry (default_path_compl_list) and the manual (export selected plugin) as well with the current list of variables


TurboGit:
Thanks !

PR #5008, opened 2020-05-12
Add missing steps to setup guide

Lucki:
While following the guide I've spotted two missing parts:

* When an installation of Darktable already exists `$CAMLIBS` and `$IOLIBS` will point in the windows installation eg `C:\Program Files\Darktable\lib\libgphoto2\2.5.23`. They should be pointing into the Msys2 terminal.
* `cmake` won't be found if the path `/mingw64/bin` is not in `$PATH`


TurboGit:
Sounds correct to me. Thanks.

PR #5009, opened 2020-05-12
Color picker changes to remaining modules and blend

dterrahe:
Converted the color pickers in the remaining modules and the blending parametric mask to the more compact dt_color_picker_new approach.

There is also a change in behaviour for some "set values" pickers; previously ctrl or shift modifiers had to be pressed while selecting the picker button to enable creation of positive or negative curves. Now, the modifier needs to be held while selecting the area (or at least while releasing the mouse button). Together with the area extendable corners commit from a few days back, this allows changing the curve type without clicking the button (off and back on) again. This also turned out easier to implement (no need to remember modifiers) but please let me know if this is an unwelcome change.

There are some iop modules, basicadj, rgblevels and colorchecker, that do not (fully) use the color_picker proxy framework. I do not completely understand the reason for this yet or whether they could be easily converted. But the result is that their buttons do not (yet) get switched off when one of the standard color checkers gets enabled. Similarly the colorpicker (left hand side) module.

Now that all users of colorpickerproxy have been converted to a minimalist use of the framework the next step is to strip out any of the complexity (differentiation between module and blend for example) that is no longer required. At the same time I'll try to address the few corner cases where pickers do not get disabled by enabling another one or switching modules. I would have preferred there to be one global place, rather than one for each module, where color picker results were stored, but changing that now would mean too many changes and opportunity for breakage.

### DIFF
--- a/src/develop/blend.h
+++ b/src/develop/blend.h
@@ -403,8 +403,6 @@ typedef struct dt_iop_gui_blend_data_t
   GtkWidget *lower_polarity;
   GtkWidget *colorpicker;
   GtkWidget *colorpicker_set_values;
-  dt_iop_color_picker_t color_picker;
-  int picker_set_upper_lower;
   GtkWidget *showmask;
   GtkWidget *suppress;
   void (*scale_print[8])(float value, char *string, int n);
@@ -479,6 +477,8 @@ void dt_iop_gui_update_blendif(dt_iop_module_t *module);
 void dt_iop_gui_update_masks(dt_iop_module_t *module);
 void dt_iop_gui_cleanup_blending(dt_iop_module_t *module);
 void dt_iop_gui_blending_lose_focus(dt_iop_module_t *module);
+
+gboolean blend_color_picker_apply(struct dt_iop_module_t *module, dt_dev_pixelpipe_iop_t *piece);
 
 /** routine to translate from mode id to sequence in option list */
 int dt_iop_gui_blending_mode_seq(dt_iop_gui_blend_data_t *bd, int mode);
--- a/src/develop/blend_gui.c
+++ b/src/develop/blend_gui.c
@@ -54,14 +54,6 @@ typedef enum _iop_gui_blendif_channel_t
   ch_max = 4
 } _iop_gui_blendif_channel_t;
 
-typedef enum dt_gui_blendif_pickcolor_type_t
-{
-  DT_BLENDIF_PICK_NONE = 0,
-  DT_BLENDIF_PICK_COLORPICK = 1,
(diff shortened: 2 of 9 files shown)


TurboGit:
> Now, the modifier needs to be held while selecting the area (or at least while releasing the mouse button). Together with the area extendable corners commit from a few days back, this allows changing the curve type without clicking the button (off and back on) again. This also turned out easier to implement (no need to remember modifiers) but please let me know if this is an unwelcome change.

This seems better this way. Will test soon.

TurboGit:
Looks pretty good ! Two issues :
- In **rgb levels** the picker for the highlight does nothing
- In **rgb levels** & **basic adjustment**, using the "picker" to draw an area for auto applying settings based on the area works, but this picker is never deactivated. It should be when selecting another picker. Note that it is properly deactivated when dragging any sliders.

dterrahe:
Thanks for turning around your tests so quickly!

I fixed the regressions you pointed out, but there's still an existing bug in rgblevels where if the auto apply "picker" is switched off by pressing on one of the real pickers, it isn't really off. Pressing the auto apply picker after that _really_ switches it off and pressing it once more finally turns it on.

Like I tried to indicate above; I think the correct solution is to really turn these fake pickers into real color pickers, so they benefit from integration into the framework and any improvements like the dragable corners. But since I don't understand what's going on in the code yet, I didn't want to mess with it. So this "fix" is just a band-aid by installing an extra callback.

dterrahe:
Made another small change in white balance module to avoid explicitly using color_picker structure (and having to initialise it). Now emitting signal from hidden color picker widget that is created using standard dt_color_picker_new and owns its own color picker structure.

TurboGit:
Ok, to me this is ok to go. I did not find more issues. Again a very nice clean-up.

By merging we'll have more field testing really needed in such deep changes.

dterrahe:
Many thanks!

Fully agreed on the deep testing needed; this should really be like-for-like and apart from the positive/negative curve (ctrl/shift) change I've tried to be careful not to change any behavior but it is easy to cause little breakages (like the highlight in rgb levels which I wish I could say I left in there to see how well you test; you passed!! ;-) )

Also the reason why I would like the original authors of rgb levels & basic adjustment (or heavy users) to be involved in integrating it more closely in the color picker framework, because I don't even know exactly what to test. Same for color look up; I've got the impression that it may not work correctly for me anyway, so...

Any area where breakage may be hard to catch is color space handling (mainly in blend). So I will be looking closely at that while simplifying colorpickerproxy. Removing fields there that are not properly referred to in all cases might unearth some errors (either introduced by me or existing).

PR #5022, opened 2020-05-13
More reliable export size in dt_imageio_export_with_flags

### LINKED ISSUE #4979: Could not get original size in export after rotation
Darktable should be able to export a picture in the EXACT original size of the rawfile, even when the orientation (portrait/landscape) is rotated.

I took a series of pictures for panorama-stitching with hugin. All pictures are take with same camera and lens as raw (Olympus E-M1 Mk.II). All raw show the size w=5240 h=3912.
The horizontal pictures arround are taken in orientation=portrait and are automaticly rotated by darktable at import. Vertical pictures from ground and sky are marked as orientation=landscape by the camera.

For the workflow with hugin I need all the pictures from the series in the same size and same orientation to be accepted as from the same lens. 
I rotated the ground- and sky-pictures manually from landscape to portrait.

On export (as 16 bit tif) I could not get the original size 5240/3912 and I got different sizes for manually rotated pictures and automaticly rotated pictures.
One or two pixels got lost in width and/or hight, always different between the 2 groups.
I tried some options for size in the export-module but always got different sizes.

Hugin don't accept this as "from one lens".  I had to switch OFF all rotation (auto and manual) in darktable and do the rotation after export with imagemagick to get unique size and rotation for the following workflow.

**Platform:**
Darktable 3.0.2
Ubuntu 18.04 LTS

jenshannoschwalm:
Fixes #4979, for discussions about problems related to this see #3808 #3666 #3646 #3757

This pr tries to make the export output size more predictable.

1. After doing the overflow checks via `dt_dev_distort_backtransform_plus` we should
correct both sizes if any overflow was detected. This will potentially crop more but
the size does not depend on rotation any more.

2. If we use format_params max_width/height we are more pessimistic about the output
size (likely we will decrease dx and dy after the backtransform test) and thus increase
a given value here.

@upegelow @TurboGit you both worked on this lately

### DIFF
--- a/src/common/imageio.c
+++ b/src/common/imageio.c
@@ -746,8 +746,8 @@ int dt_imageio_export_with_flags(const uint32_t imgid, const char *filename,
             ? FALSE
             : high_quality;
 
-  const int width = format_params->max_width;
-  const int height = format_params->max_height;
+  const int width = format_params->max_width > 0 ? format_params->max_width + 1 : 0;
+  const int height = format_params->max_height > 0 ? format_params->max_height + 1 : 0;
 
   const float max_scale = ( upscale && ( width > 0 || height > 0 )) ? 100.0 : 1.0;
 
@@ -762,11 +762,15 @@ int dt_imageio_export_with_flags(const uint32_t imgid, const char *filename,
 
   if(dt_dev_distort_backtransform_plus(&dev, &pipe, 0.f, DT_DEV_TRANSFORM_DIR_ALL, origin, 1))
   {
-    processed_width = scale * pipe.processed_width + 0.5f;
-    processed_height = scale * pipe.processed_height + 0.5f;
+    processed_width = scale * pipe.processed_width + 0.8f;
+    processed_height = scale * pipe.processed_height + 0.8f;
 
-    if(ceilf(processed_width / scale) + origin[0] > pipe.iwidth) processed_width--;
-    if(ceilf(processed_height / scale) + origin[1] > pipe.iheight) processed_height--;
+    if((ceilf(processed_width / scale) + origin[0] > pipe.iwidth) ||
+       (ceilf(processed_height / scale) + origin[1] > pipe.iheight)) 
+    {
+      processed_width--;
+      processed_height--;
+    }
   }
   else
   {


TurboGit:
Indeed, better this way. I tested different rotation and each time I got the very same width x height. And the bigger dimension was really the one I put in the export module. So more  precise and more predictable. Let me merge this, nice fix really, thanks.

TurboGit:
@jenshannoschwalm  : Sadly this produces regression on the regression tests!

```
$ cd src/tests/integration
$ darktable-cli --width 2048 --height 2048 --hq true images/mire1.cr2 0001-exposure/exposure.xmp output.png
$ exiv2 output.png | grep -i "Image size"
Image size      : 2049 x 1365
```

Do you think this can be improved ?

Reverting : c12e3d31f9cac87a3ce2726c185ff5a0a6c3fe60 and d881b854ffaea135520da69ae6c0f705da559d7c

I get:

```
$ exiv2 output_01.png | grep -i "Image size"
Image size      : 2048 x 1364
```

jenshannoschwalm:
Ooh. I'll have to work on this asap.

jenshannoschwalm:
Had a first look at it, this is really not working the right way, i already understand what's wrong and how it can (hopefully finally) be fixed

jenshannoschwalm:
Dear Pascal @TurboGit. My problem right now is i don't know what to fix "exactly". If we change the algo to have a more predictable / correct size we might and often will have a different output. I would suggest these rules for any image with a size of dx * dy.

1. If we don't specify any output width or height the exported image will have a size of dx-1*dy-1, this probably means less rounding errors than dx*dy. Do you agree? - Although i couldn't detect any quality difference

2. If we specify either width or height the larger dimension (or the one specified) will be exactly as the specified size

3. Do we want to have any special care for cropped images? There were issues like: i cropped to 1024x800 and got an exported image 1023x799.

Let me know what you think please - or any other one reading this.

parafin:
> 1. If we don't specify any output width or height the exported image will have a size of dx-1_dy-1, this probably means less rounding errors than dx_dy. Do you agree? - Although i couldn't detect any quality difference

I don't follow why non-matching size is better.

jenshannoschwalm:
> I don't follow why non-matching size is better.

I am not sure about this either. From reading code & discussions in #3808 it's clear we have to do a check for the size and in some cases the check indicates there would be out-of-bounds access.

The pr then crops the image by one, (option 1) thus staying at a scale of 1.0 - why is this done so? Because of avoiding any blur by rescaling? I thought thats's the reason -> so dx-1*dy-1.

Option 2 would mean to stay with the size but choose a slightly different scale to avoid out-of-bounds access.

TurboGit:
For option one I think one would expect to have dx^dy but more importantly the exact same size if the image is flipped or rotated -90 / +90. So if we can have dx-1^dy-1 in all cases I'm ok.

For option 3, it looks like one more important point as if you crop the image at a specific size you really expect the size to be respected.

parafin:
I think having dx-1*dy-1 for first point and exact crop for third is a contradiction.
Where exactly out-of-bounds access happens? So some iop requires 1 pixel border to function?

parafin:
Scaling IMHO is not a good option either.

jenshannoschwalm:
> Where exactly out-of-bounds access happens?

I don't know. @upegolow has worked on this. At least in the commits i couldn't find a hint about it.

>  So some iop requires 1 pixel border to function?

Seems so.

So what to do?

For sure rotation should not change any exported size/ratio. Thats an easy one already fixed.

What about an export config flag - maybe hidden - "enforce-exact-image-size"?
If true we would scale for the desired exact size, if false accept a cropping.

Unfortunately we can **not** safely use the crop data in the image struct for a border check, accesswise yet but we might read data having no valid content.

parafin:
Another solution would be to allocate a bigger buffer, then out-of-bounds access won't crash. But it may produce image artefacts I guess...

TurboGit:
@jenshannoschwalm : I have created a dedicated issue to discuss this #5228 (better than filling a merged PR).

PR #5023, opened 2020-05-13
remove dt_view_image_expose and co

AlicVB:
Now that thumbnails rewrite is over, we can drop this unreadable and unmaintainable function and all its derivative :)

If someone can test tethering, that would be great, as my camera isn't completely compatible with the capture mode...


TurboGit:
I will test the tethering view, probably tomorrow.

src/views/print.c
```diff
@@ -226,7 +233,24 @@ static void expose_print_page(dt_view_t *self, cairo_t *cr, int32_t width, int32
   cairo_rectangle (cr, ax, ay, awidth, aheight);
   cairo_fill (cr);
 
-  dt_view_image_only_expose(prt->image_id, cr, iwidth, iheight, ix, iy);
+  cairo_surface_t *surf;
```
TurboGit, line 236: You need = NULL here otherwise dt crash entering print module.

src/views/tethering.c
```diff
@@ -197,17 +204,26 @@ static void _expose_tethered_mode(dt_view_t *self, cairo_t *cr, int32_t width, i
...
-    params.image_over = &(lib->image_over);
-    params.imgid = lib->image_id;
-    params.cr = cr;
-    params.width = width - (MARGIN * 2.0f);
-    params.height = height - (MARGIN * 2.0f);
-    params.px = pointerx;
-    params.py = pointery;
-    params.zoom = 1;
-    dt_view_image_expose(&params);
+    cairo_surface_t *surf;
```
TurboGit, line 207: Not tested yet, but probably as for print module. You need = NULL there.

TurboGit:
Apart from that, no regression but the tethering module is not working (don't know since how long).

TurboGit:
@AlicVB : let me fix that, I have tested it and built a fix for tethering mode which is broken since lt rewrite.

PR #5024, opened 2020-05-13
map : don't show all on exit

### LINKED ISSUE #5020: Image seeing overlay block infos in print view when coming from map one
**Describe the bug**
Seeing an image on print mode is shown in hover overlay block infos like in preview or culling mode regardless of overlay mode choosen... if you come from map view. @AlicVB 

**Expected behavior**
Print view is not a place where overlay mode is needed and wanted as we want to see the image we will print only. So no overlay mode could be seen like when you come from lighttable or darkroom view.

 - Darktable Version: Master (a100547dff0a22b4f453452018be1e7232670db1)

AlicVB:
fix #5020


TurboGit:
Thanks !

PR #5025, opened 2020-05-13
darkroom: unify icon sizes

elstoc:
amend the size of the icons on the bottom-right of the darkroom so that
they match those on the bottom-left and top-right.


TurboGit:
Good catch ! Thanks.

PR #5028, opened 2020-05-13
modulegroups: temporarily toggle group icon off while searching

elstoc:
when using module search, temporarily toggle the current module group icon off (since we are no longer viewing modules in that group)

if the search text is cleared, toggle the previously-selected group icon back on and redisplay that group's modules

if the user re-selects the previously-selected module group while searching, instead of toggling it to 'show all', just re-show the contents of that group and toggle the icon back on

selecting a new group while searching will toggle that group and show its modules (as currently)


TurboGit:
Just to clarify clicking on different group icons and testing the search entry I git two icons selected.

elstoc:
Thanks @TurboGit I'd forgotten this functionality existed. Should be fine now.

TurboGit:
No problem, all working good now. Thanks.

PR #5031, opened 2020-05-13
prevent dynamic shortcuts giving keyboard focus to sliders

elstoc:
prevents dynamic slider shortcuts from giving keyboard focus to the
slider being changed. This is not consistent with any other shortcuts
and also gives keyboard focus to the slider when the module itself is collapsed.


TurboGit:
Looks ok to me. Thanks.

PR #5041, opened 2020-05-14
iop module gui focus shortcuts

elstoc:
adds a new "focus module" shortcut to all iop modules to toggle gui
focus and allows control-click on the module enable button to also
toggle gui focus for the module (since this icon is now used to show
focus- as well as enabled-state for modules)

this is probably also a reasonable alternative solution for #4608 and
gives the user the choice of showing crop & rotate gridlines (by
assigning focus) or not prior to executing another module shortcut

for crop & rotate the focus toggle also works as an alternative 'commit'
button/shortcut

### DIFF
--- a/src/develop/imageop.c
+++ b/src/develop/imageop.c
@@ -1060,13 +1060,16 @@ static void dt_iop_gui_multiinstance_callback(GtkButton *button, GdkEventButton 
   dtgtk_button_set_active(DTGTK_BUTTON(button), FALSE);
 }
 
-static void dt_iop_gui_off_callback(GtkToggleButton *togglebutton, gpointer user_data)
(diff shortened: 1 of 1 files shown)


elstoc:
Possibly resolves #3481, though this would require two shortcuts rather than one

TurboGit:
Ok, let's go with that. We can then move guides without deploying the module on all modules having them like vignette, crop&rotate, graduiatednd...

elstoc:
Thanks @TurboGit. I'm happy to close #4608 based on that, but I didn't raise the issue which inspired it so I'm not sure. What do you think?

TurboGit:
Yes I'll close #4608 for now.

PR #5046, opened 2020-05-14
last phase of the colorpicker refactor

dterrahe:
Removed blender specific portions from color_picker_proxy.
Allow DT_COLOR_PICKER_POINT_AREA for quad pickers as well.
Call dt_iop_color_picker_reset from imageop.c to enforce switching off pickers at focus change.
Setting the name of a toggle button widget to "keep-active" means it will not get switched off when any of the other widgets change. Used in color zones, rgb curve and blend.Made color picker structure private to color_picker_proxy; passing the active color picker widget to apply function for easy switch statement. This made the patch big again, sorry.
Do not call apply when picked spot/area is unchanged; this avoids duplicate processing when the first call to apply triggers a recalc and second signal (alternative, hopefully more generic/robust, approach to skip_apply)


TurboGit:
Ok, works for me!

Issue #5021, opened 2020-05-13
Fujifilm X-T4 white balance fine-tuned


JonHrovath:
Here is a fine-tuned white balance preset list for a Fujifilm X-T4.  Hope it helps.

[darktable-whitebalance-20200512.tar.gz](https://github.com/darktable-org/darktable/files/4619480/darktable-whitebalance-20200512.tar.gz)

johnny-bit:
Hey @TurboGit, please assign it to me, I'll do this soon :)

JonHrovath:
Just wanted to say thanks to the devs for getting these changes in so fast.  I just pulled down master and the WB presets all look good.  I don't see any Denoise Profiles yet for the X-T4, so I'm hoping to get time next week to go through the process and get the images taken at each ISO step.  If I find even more time, I'll try and brush up on my C skills and get into some bug fixing...

PR #5048, opened 2020-05-14
fix wb finetune warnings and possible error condition with fuji data

johnny-bit:
This one little thing was found when doing stuff for #5021


TurboGit:
Thanks !

PR #5050, opened 2020-05-14
fix module toggle button

### LINKED ISSUE #5047: On/Off button for modules out of sync
darktable 3.1.0+1599~g7600fb842
OS: Manjaro latest
openCL on
Nvidia GTX-1660.

Take any module, for instance _white balance_.
At start /by default/ it is ON.
Toggle the ON/OFF indicator. Now the indicator goes to OFF, but the module is _not_ turned off.
Hoover the mouse above the indicator, and help text says "white balance is switched on" (but it is not).
Now toggle the ON/OFF indicator. The indicator goes to ON, but the module now turns OFF.
The indicator and module are now out of sync.

The same out-of-sync problem seems to exist for every module.

elstoc:
fix reversed on/off behaviour for module toggle button

Resolves #5047

### DIFF
--- a/src/develop/imageop.c
+++ b/src/develop/imageop.c
@@ -1060,16 +1060,26 @@ static void dt_iop_gui_multiinstance_callback(GtkButton *button, GdkEventButton 
   dtgtk_button_set_active(DTGTK_BUTTON(button), FALSE);
 }
 
-static gboolean dt_iop_gui_off_callback(GtkWidget *w, GdkEventButton *e, gpointer user_data)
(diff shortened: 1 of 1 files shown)


TurboGit:
Still an issue there. Open a RAW file, white balance is activated by default but if you hover over the ON/OFF button it says OFF which is wrong. Clicking on the button afterward does not confuse the module, so this is only the initial state which is not correct.

elstoc:
Ok I think (hope) that's got it now.

TurboGit:
Looks ok now !

PR #5053, opened 2020-05-14
culling : another batch of small fix

### LINKED ISSUE #5035: culling mode no refresh based on filter
**Describe the bug**
The culling display is not refreshed based on filter.

**To Reproduce**

1. Go to lighttable
2. Select 5 images
3. Set filter to "all but rejected"
4. Click on Ctrl+X to display all 5 images
5. Reject one, the image is still displayed

**Expected behavior**
As in previous version I expect the rejected image to be removed from the display. I'm using this view to triage multiple almost identical image to keep only one or two.

 - Darktable Version: master
 - OS: Debian Linux

**Additional context**

AlicVB:
- this fix the update of images list when collection change (broken in one of my last commits) fix #5035
- this add a warning when one enter in culling dynamic without selection
- this fix culling offset when starting dt in culling mode and the filmstrip as been scrolled

### DIFF
--- a/src/dtgtk/culling.c
+++ b/src/dtgtk/culling.c
@@ -785,6 +785,7 @@ void dt_culling_init(dt_culling_t *table, int offset)
 
   // get first id
   sqlite3_stmt *stmt;
+  gchar *query = NULL;
   int first_id = -1;
 
   if(offset > 0)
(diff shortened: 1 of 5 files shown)


TurboGit:
All good to me, thanks!

PR #5062, opened 2020-05-15
Less offensive 'module-in-focus' icon

jenshannoschwalm:
This modifies #4927 as some of us - including me - see the first attempt to be a bit too much standing out after some time using it.

### DIFF
--- a/src/dtgtk/paint.c
+++ b/src/dtgtk/paint.c
@@ -274,7 +274,7 @@ void dtgtk_cairo_paint_switch(cairo_t *cr, gint x, gint y, gint w, gint h, gint 
   if(flags & CPF_FOCUS)
   {
     cairo_set_line_width(cr, 0.05);
-    cairo_arc(cr, 0.5, 0.5, 0.60, 0, 2 * M_PI);
+    cairo_arc(cr, 0.5, 0.5, 0.60, (-50 * 3.145 / 180), (230 * 3.145 / 180));
     cairo_stroke(cr);
   }
   cairo_set_line_width(cr, 0.1);
--- a/src/dtgtk/paint.c
+++ b/src/dtgtk/paint.c
@@ -270,19 +270,18 @@ void dtgtk_cairo_paint_switch(cairo_t *cr, gint x, gint y, gint w, gint h, gint 
   const gint s = w < h ? w : h;
   cairo_translate(cr, x + (w / 2.0) - (s / 2.0), y + (h / 2.0) - (s / 2.0));
   cairo_scale(cr, s, s);
-
-  cairo_set_line_width(cr, 0.1);
   cairo_set_line_cap(cr, CAIRO_LINE_CAP_ROUND);
   if(flags & CPF_FOCUS)
   {
-    cairo_arc(cr, 0.5, 0.5, 0.66, 0, 2 * M_PI);
+    cairo_set_line_width(cr, 0.05);
+    cairo_arc(cr, 0.5, 0.5, 0.60, 0, 2 * M_PI);
     cairo_stroke(cr);
   }
+  cairo_set_line_width(cr, 0.1);
   cairo_arc(cr, 0.5, 0.5, 0.46, (-50 * 3.145 / 180), (230 * 3.145 / 180));
   cairo_move_to(cr, 0.5, 0.0);
   cairo_line_to(cr, 0.5, 0.5);
   cairo_stroke(cr);
-
   if(flags & CPF_ACTIVE) // If active add some diffuse light
   {
     cairo_arc(cr, 0.5, 0.5, 0.45, 0.0, 2*M_PI);
@@ -298,13 +297,13 @@ void dtgtk_cairo_paint_switch_on(cairo_t *cr, gint x, gint y, gint w, gint h, gi
   const gint s = w < h ? w : h;
   cairo_translate(cr, x + (w / 2.0) - (s / 2.0), y + (h / 2.0) - (s / 2.0));
   cairo_scale(cr, s, s);
-
-  cairo_set_line_width(cr, 0.1);
   if(flags & CPF_FOCUS)
   {
-    cairo_arc(cr, 0.5, 0.5, 0.7, 0, 2 * M_PI);
+    cairo_set_line_width(cr, 0.05);
+    cairo_arc(cr, 0.5, 0.5, 0.65, 0, 2 * M_PI);
     cairo_stroke(cr);
   }
+  cairo_set_line_width(cr, 0.1);
   cairo_arc(cr, 0.5, 0.5, 0.35, 0, 2 * M_PI);
   cairo_fill(cr);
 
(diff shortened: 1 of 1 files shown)


src/dtgtk/paint.c
```diff
@@ -270,19 +270,18 @@ void dtgtk_cairo_paint_switch(cairo_t *cr, gint x, gint y, gint w, gint h, gint
...
   cairo_translate(cr, x + (w / 2.0) - (s / 2.0), y + (h / 2.0) - (s / 2.0));
   cairo_scale(cr, s, s);
-
-  cairo_set_line_width(cr, 0.1);
   cairo_set_line_cap(cr, CAIRO_LINE_CAP_ROUND);
   if(flags & CPF_FOCUS)
   {
-    cairo_arc(cr, 0.5, 0.5, 0.66, 0, 2 * M_PI);
+    cairo_set_line_width(cr, 0.05);
+    cairo_arc(cr, 0.5, 0.5, 0.60, (-50 * 3.145 / 180), (230 * 3.145 / 180));
```
TurboGit, line 277: I would even go with:
`cairo_arc(cr, 0.5, 0.5, 0.60, 0, (180 * 3.145 / 180));`

src/dtgtk/paint.c
```diff
@@ -298,13 +297,13 @@ void dtgtk_cairo_paint_switch_on(cairo_t *cr, gint x, gint y, gint w, gint h, gi
   const gint s = w < h ? w : h;
   cairo_translate(cr, x + (w / 2.0) - (s / 2.0), y + (h / 2.0) - (s / 2.0));
   cairo_scale(cr, s, s);
-
-  cairo_set_line_width(cr, 0.1);
   if(flags & CPF_FOCUS)
   {
-    cairo_arc(cr, 0.5, 0.5, 0.7, 0, 2 * M_PI);
+    cairo_set_line_width(cr, 0.05);
```
TurboGit, line 302: That's perfect for always active module.

src/dtgtk/paint.c
jenshannoschwalm: I don't care that much. Personally i don't like half circles and here tried to keep the icon's shape.

TurboGit:
I feel this is better anyway. Let's merge to have more feedback.

elstoc:
Admittedly I have set a smaller font size on my machine but I have to get pretty close to the screen to see the difference between in and out of focus - I think this might have gone a bit too far the other way.

Nilvus:
I agree to @elstoc. It's an interesting idea but not sure it's not useful. But if you let that, I think that need some more ideas to avoid making too much PR. Merged to quickly as I've just had time to test it. I will think about that and try to find a more intersting idea. The first was indeed quite "offensive" as you said, and now not so visible.

elstoc:
I do like the idea of limiting it to more of an 'underline' but the focus indicator is definitely useful, especially if you have multiple modules open.

Nilvus:
I've made some tests with CSS (not on focus mode but on classic mode as set focus on module header is not possible by now with CSS). I don't have the "Yes, that's it" idea by now but I share some ideas I've had anyway on my mind. Maybe those ones will gave you other ideas:

First ones with the downside that they will need at least color to be tweaked by CSS to adjust it on all themes:
- Simply have a left border on the module header on focus, like categories in preferences window (exact same border and color)
- Same idea but with icon color background (absolutely not possible by now with CSS) like I made those last days on file chooser dialog windows (as @elstoc says: quite gnome-ish ;-)!).
- Or add a quite "bottom-border" on the whole module header. The best I've found is by using box-shadow with CSS. See below an example:


And last idea (don't need a CSS tweak):
- made the focus highlight not around the icon but inside. For example, a little hole for always enable button and a sort of little pin (colored) for other buttons. But maybe tricky to do.

Just some thoughts. Hope I will have (or one of you) the "Yes, that's definitely it" idea.

elstoc:
Just to be clear when I said 'underline' I meant the curved semicircular underline that was implemented in this PR. Just needs to be slightly more prominent (while being slightly less prominent than it was before!)

Nilvus:
> Just to be clear when I said 'underline' I meant the curved semicircular underline that was implemented in this PR. Just needs to be slightly more prominent (while being slightly less prominent than it was before!)

My idea is not about your comment, just other ideas. I'm not sure that making focus on icon is a so good idea.

AlicVB:
My 2 cents : why not changing the title color (eventually in bold) to mimic the "blank" of the on/off icon ?
That would be very simple by adding a specific class like 'dt_module_focus' to the main module widget. This would have the advantage of being pure css, so if a theme want to change it to make it more/less visible, it's always possible.

Other than that, I like @nilvus idea of vertical left border too. (and this can be pure css too)

Nilvus:
Good idea @AlicVB. And we can make both together: changing title color and left border (needs to be possible to tweak on CSS when focus mode. Not possibly actually. I think it would finally be better as more visible and consistent than the rest of the UI and the actual focus icon. I think it would be difficult to have a focus icon not "too much standing out" and not too less visible as it is actually.

@jenshannoschwalm, @elstoc and @TurboGit: what do you think about that?

TurboGit:
Looks good to me, but current situation is good to me too :) So I'm ok for the change anyway.

jenshannoschwalm:
I don't have a strong argument for a specific proposal. Personally i still think a modified icon is the 'best' solution as it mimicks the icon changing depending on on/off ...

I tried to make the current status just a bit 'stronger', i don't think the whole line should stand out.

elstoc:
> Personally i still think a modified icon is the 'best' solution as it mimicks the icon changing depending on on/off

I agree, and the fact that the icon changed was the reason I added the control-click behaviour to the toggle button in #5041.

Nilvus:
As you want.

elstoc:
I've had a play around with it. I like the following:
```
  if(flags & CPF_FOCUS)
  {
    cairo_set_line_width(cr, 0.1);
    cairo_arc(cr, 0.5, 0.5, 0.65, M_PI * 0.2, M_PI * 0.8);
    cairo_stroke(cr);
  }
```

TurboGit:
@Nilvus : Don't give up this fast, let's see what your proposal looks like.

Nilvus:
> @Nilvus : Don't give up this fast, let's see what your proposal looks like.

No give up, this is not useful for me as I use the only one module (so the one in focus) way. I can't do my proposal with CSS so if you want to propose a PR or anything else, I'm on. I'm always on for improving UI.

elstoc:
How about if we make use of the existing 'diffuse light' in the middle of the button and have it only apply when the module is focused? The following examples are on (focused), off (unfocused), on(unfocused), off (focused) in that order.

Screenshot_2020-05-16_19-40-20 
Screenshot_2020-05-16_19-40-59

AlicVB:
@elstoc : why not... but it would anyway be really good to expose the focus status to css by adding some specific class to the module widget. If you want I can have a look...

elstoc:
Should just be a case of adding a couple of lines to dt_iop_request_focus where it's currently requesting the widget redraw.

elstoc:
I've raised #5091 with my 'diffuse light' suggestion, if for no other reason than so we can continue this discussion on an open PR.

PR #5067, opened 2020-05-15
add lib for illuminants back and forth computations

aurelienpierre:
This is the `illuminants.h` lib detached from #4940 because @johnny-bit needs it in #4360.

It adds a collection of standards CIE illuminants, and more accurate way to convert illuminant chromaticity to temperature and tint. 

Currently, the white balance module only uses RGB white-balance coeffs internally, and computes correlated color temperature and tint on-the-fly, for GUI purposes. The temperature moves the chromaticity along the Planckian locus, and the tint moves it along the y axis (tint varies here from 0.9 to 1.1 in GUI user params):
image
This is bad because both controls are not orthogonal, so the tint editing does not happen at constant temperature, which makes both settings inter-dependent in weird ways. Also, a part of the cyan region cannot be reached by these controls.

So we introduce a different way of computing the tint, as the distance from the planckian locus in the orthogonal direction (tint varies here from -0.05 to 0.05 in program params):
image
This makes both settings fully independent, but also allows to reach new color regions in the blue/cyan part.

Then we remove the need for bisection and ugly approximations in finding the correlated color temperature, since we use the analytical expression of the Planckian locus and its normal vector at all time. 

Here is the relative error in the xy -> tint -> xy mapping with proposed method:
image

And the corresponding xy uncertainties bars:
image

For the xy -> temperature -> xy mapping, the relative error is below 0.004%:
image

### DIFF
--- a/src/common/illuminants.h
+++ b/src/common/illuminants.h
@@ -1,0 +1,428 @@
+/*
+    This file is part of darktable,
+    Copyright (C) 2020 darktable developers.
+
+    darktable is free software: you can redistribute it and/or modify
+    it under the terms of the GNU General Public License as published by
+    the Free Software Foundation, either version 3 of the License, or
+    (at your option) any later version.
+
+    darktable is distributed in the hope that it will be useful,
+    but WITHOUT ANY WARRANTY; without even the implied warranty of
+    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+    GNU General Public License for more details.
+
+    You should have received a copy of the GNU General Public License
+    along with darktable.  If not, see <http://www.gnu.org/licenses/>.
+*/
+
+#include "common/colorspaces_inline_conversions.h"
+
+/* Standard CIE illuminants */
+typedef enum dt_illuminant_t
+{
+  DT_ILLUMINANT_PIPE   = 0,    // darktable pipeline D50
+  DT_ILLUMINANT_A      = 1,    // incandescent bulb
+  DT_ILLUMINANT_D      = 2,    // daylight
+  DT_ILLUMINANT_E      = 3,    // equi-energy (x = y)
+  DT_ILLUMINANT_F      = 4,    // fluorescent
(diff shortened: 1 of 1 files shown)


src/common/illuminants.h
```diff
@@ -0,0 +1,428 @@
...
+  // Get an sRGB preview of current illuminant
+  float XYZ[4];
+  illuminant_xy_to_XYZ(x, y, XYZ);
+
+  // Fixme : convert to RGB display space instead of sRGB but first the display profile should be global in dt,
+  // not confined to colorout where it gets created/destroyed all the time.
+  dt_XYZ_to_Rec709_D65(XYZ, RGB);
+
+  // Handle gamut clipping
+  float max_RGB = fmaxf(fmaxf(RGB[0], RGB[1]), RGB[2]);
```
TurboGit, line 195: const

src/common/illuminants.h
```diff
@@ -0,0 +1,428 @@
...
+
+    // Current temperature in the lookup range
+    const float T = T_min +  step * T_range;
+
+    // Current x, y chromaticity
+    float x_bb, y_bb;
+    CCT_to_xy_blackbody(T, &x_bb, &y_bb);
+
+    // Compute distance between current planckian chromaticity and input
+    float radius_tmp = hypotf((x_bb - x), (y_bb - y));
```
TurboGit, line 418: const

src/common/illuminants.h
aurelienpierre: done

src/common/illuminants.h
aurelienpierre: done

johnny-bit:
Cool, now I have to go back to coding :smile:

TurboGit:
Sure, we try to keep you active :)

johnny-bit:
If #4360 lands I'll integrate this. Initial tests are a bit rough around the edges for full on switch to blackbody + tint..

PR #5068, opened 2020-05-15
thumbs : add tooltip for groups

AlicVB:
"group" icon on thumbnails show tooltip like : 

------------------------------------------
3 grouped images

**current (leader)**
IMG_005896.NEF
IMG_005478.NEF

### DIFF
--- a/src/dtgtk/thumbnail.c
+++ b/src/dtgtk/thumbnail.c
@@ -74,6 +74,67 @@ static void _thumb_update_extended_infos_line(dt_thumbnail_t *thumb)
   g_free(pattern);
 }
 
+static void _image_update_group_tooltip(dt_thumbnail_t *thumb)
+{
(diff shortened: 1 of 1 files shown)


TurboGit:
Nice !

TurboGit:
Just a minor change needed, when grouping files that are clone you get the same name into the tooltip. So to distinguish them I suppose we want to add the version of the clone. Thoughts ?

AlicVB:
oh yes, I'll have a look

AlicVB:
@TurboGit : done + small optimizations

Note that to avoid to complex computations and loops, I've done a very rough versionning test : if version is not null, it is shown.
Although there might be cases where there's only 1 image version but its version number is not null...

Let's think indicating version numbers in this rare cases is not too harmful :)

TurboGit:
Works for me. Thanks.

Issue #5058, opened 2020-05-15
Sort order is misaligned.


jade-nl:
**Describe the bug**
In lighttable view, using the *sort by* option the chosen order and the order shown are out of sync/scrambled.

**To Reproduce**
1. Colour code a few random shots
2. Select *sort by -> **color label***  (nothing happens)
3. Select *sort by -> **title*** (images are sorted by color label)

**Expected behavior**
Sort order to be as advertised

**Platform (please complete the following information):**
* darktable version: Development 3.1.0+1604
* OS               : Linux - kernel 5.4.0-0.bpo.4-amd64
* Distro           : Debian Buster (rel. 10.4)

jade-nl:
Just did some more testing and the *sort by -> **rating*** is also misaligned.

1. Rate a few random shots
2. Select *sort by -> **rating*** (nothing happens)
3. Select *sort by -> **full path*** (images are sorted by rating)

AlicVB:
@jpverrue : I think you forgot to add your awesome new sort orders to the filter combobox (in lib/tools/filters.c)

jpverrue:
Oh, yes! I completely forgot them here. I'm correcting it right now!

jpverrue:
Yes, the filters are offset by 4 rows (4, like the 4 new timestamps...). However I think that it is not in libs/tools/filter.c that there is a problem, but in src/common/collection.c. In the dt_collection_update_query() function maybe. At the moment, I don't see anything...

AlicVB:
Well, in filter.c, when one change the combobox, we pass the combobox selected index to collection.c and there, this index is used as a sorting enum value. So the list in the combobox filter need to match the list of the enum. Or we need to do some mapping in filter.c before we pass the values to collection...
(Sorry to not be more precise, I haven't dt code under this eyes just now)

jpverrue:
You're right, it was indeed in filter.c. I did the correction in a local branch. Should I do a PR before pushing or the other way around: push then do the PR? (I'm very wrong with Git!

jade-nl:
Noticed the merge.
Tested it:
All fixed!

Thanks @jpverrue  for the speedy action.

PR #5069, opened 2020-05-15
Sort-order-correction

jpverrue:
This RP corrects issue #5058
I transformed the initialization of the combobox which was only done with a label, by creating a table containing the id from the enum dt_collection_sort_t and the label.
The init is now done from this table as well the passing parameter in the _lib_filter_sort_combobox_changed() function.


TurboGit:
Ok, works for me. The stars filter is somehow broken as in the DESC order the rejected are listed first. But this is for me, probably caused by my rework of the way rejected is encoded.

Note that you still have merges into your branch :) I'll squash everything anyway. Thanks for the quick fix.

Issue #4944, opened 2020-05-08
Selecting IOP color picker doesn't deactivate color picker module


dterrahe:
Only one color picker can be active at the same time. If a color picker is selected in an IOP module, all the other color pickers in the same or other modules are deactivated. However, if a color picker is active in the separate color picker module (left side of darkroom), it _stays_ active (and the "add" button is _not_ greyed out) when a color picker is selected in an iop model (right side).

1. In color picker module on left side, click the color drop icon.
2. It becomes active and the "add" button is no longer greyed out.
3. Selecting a point or area in the image in the center changes the color and rgb values shown in the module.
4. Now select any color picker on the right side (for example in filmic rgb, the picker linked to middle grey luminance).
5. Both the color pickers on the right and on the left look active (and the "add" button can still be clicked)
6. Selecting a point or area in the image in the center only impacts the iop picker on the right.
7. Clicking the picker icon in the color picker module on the left _deactivates_ it (and also deactivates, correctly, the color picker on the right)
8. To really activate the color picker on the left, it has to be clicked again.

Putting this as a reminder to look into when I'll be working on further implementation of #4925 

Recent master a4c447d6cfc272d69cce73de206659645af15bfe (on debian).

jenshannoschwalm:
Taking this as a reminder; shouldn't we also make sure to deactivate the pickers if any mask is **added**?

dterrahe:
Probably. I'm planning to go over the color picker code and hopefully dramatically simplify it after #4925. Expecting that it will be easier to get rid of any problem corner cases afterwards/during. So if you are OK with that, I'd say don't spend too much time on fixing issues with the "old" code now. But please do add reminders like these here, so I know what to carefully test.

dterrahe:
Another existing issue that I'll try to address in the next stage but will be carried over in the initial "porting" of the borders/framing module:

1. Select either one of the color pickers in the framing module. This is a point picker.
2. Shift-click module filmic rgb (both modules visible)
3. Would have expected color picker in framing to be deactivated, but it is not (at least; the button is still bold).
4. Click one of the color pickers in filmic rgb. These are all area pickers.
5. Now switch back to framing module. The color picker is still active but previously selected point is not shown.
6. Click+drag in the picture area. The color picker is now an area selector, incorrectly (or at least; unexpectedly)
7. Switch picker off and back on again.
8 Click+drag in the picture area. The color picker is a point selector again.

dterrahe:
[...]

PR #5072, opened 2020-05-15
last two small color picker changes.

dterrahe:
Really.

The colorpicker module is not an iop module so it was patched in via the colorout module. Also it's lifetime is different so it owning both the colorpicker and a link to the iop module would blow up.

My solution is to have a link to a null module be interpreted as colorout by color_picker_proxy. Not the prettiest solution, but better than what was previously going on since now at least the color picker button correctly gets switched off.

Also removing background color from button pickers so they look same as quad pickers. Still behave slightly differently when mouse is hovering.


dterrahe:
Forgot to mention: this closes #4944

@jenshannoschwalm would you please test if the previous colorpicker changes address your "applying masks" concerns? If not, please file a concrete bug report or steps to reproduce in #4944 and I'll take a look.

TurboGit:
Works for me, thanks !

PR #5074, opened 2020-05-15
GUI: reworking of buttons

Mark-64:
This is a reworking of the geometry of most (but not all) icons buttons in DT GUI. You can find the history in #4967.
Here below I do not refer to buttons with label. Those are created in DT with the call `gtk_button_new_with_label ()` directly in Gtk and therefore they fully respect the Gtk/CSS geometry rules already.
The PR tries to fix the following problems in DT icon buttons:

- Hardcoded "inner margin" (aka padding in Gtk/CSS) fixed in pixels (no DPI scaling), controlled by a flag in C `CPF_DO_NOT_USE_MARGIN` set at button creation. However flag true did not correspond to zero padding
- Button geometry misreading/ignoring margin and padding CSS properties
- Clipping of button border in some cases

Cattura2

- Some icons not respecting their clip area, also leading to non homogeneous size

With the PR, the buttons and icons will fully respect the rules of Gtk/CSS geometry, with all properties set in CSS style sheet. They will do so for all components (top/bottom/left/right).

Cattura1

This will hopefully lead to a better scaling of the GUI for HiDPI screens.
I have implemented this without the use of extra drawing area widgets inside the buttons, although the result should be exactly the same.

For the reviewers:

**MAIN CHANGES**
The main changes are inside button.c and togglebutton.c. I implemented the Gtk/CSS geometry logic "manually" (with no extra widget). Also eliminated a branch of code supposed to manage buttons with icon+text, which is not used in DT and was not working anyway.

**ACCESSORY CHANGES**
Several changes in paint.c

- rescaled some cairo icons drawing outside the viewport
- fixed some graphics glitches
- redisegned the "display2" icon (red lines = viewport)

Screenshot from 2020-05-16 06-17-48

**MINOR CHANGES**
I removed the no longer used `CPF_DO_NOT_USE_MARGIN` from all the calls. Buttons are everywhere, so I had to touch a lot of files.... I could have left it there, but it would have been misleading for people reading the code.
Please note that `CPF_DO_NOT_USE_MARGIN` is still used 
- for some special bauhaus.c widgets
- for thumbtable overlay buttons

**STILL TO DO**

I didn't check all the cairo icons, so there could be other glitches I didn't detect. They can be fixed at later time.

@Nilvus CSS must be reworked, sorry about that... I believe you should find it easier to tweak the geometry of buttons now, because it will fully respect the CSS statements. I you find problems, just let me know. 
For your convenience, I attach here a darktable.css file, which I tweaked for testing the PR, although I am not a CSS expert by any mean, so it's not optimized. The format is before your recent reworking.
[darktable.css.txt](https://github.com/darktable-org/darktable/files/4637889/darktable.css.txt)

@AlicVB I didn't touch buttons in the new thumbtable. Those buttons are supposed to be "dynamic", meaning their size depends on the zoom level and you use CSS statements with a different logic.
That's fine, but still there is the problem how to "sync" the size (or to better say, the maximum size) of buttons with the others in the GUI, for visual consistency.
Now the maximum size of thumbtable buttons is hardcoded, but this cannot work, because the size of the other buttons is set in CSS. Maybe we could set the maximum size of by querying the size of a main GUI button, like the help button, what do you say?

### DIFF
--- a/src/develop/blend_gui.c
+++ b/src/develop/blend_gui.c
@@ -1652,15 +1652,15 @@ void dt_iop_gui_init_blendif(GtkBox *blendw, dt_iop_module_t *module)
 
     bd->colorpicker_set_values = dt_color_picker_new(module, DT_COLOR_PICKER_AREA, header);
     dtgtk_togglebutton_set_paint(DTGTK_TOGGLEBUTTON(bd->colorpicker_set_values),
-                               dtgtk_cairo_paint_colorpicker_set_values, CPF_STYLE_FLAT | CPF_DO_NOT_USE_BORDER, NULL);
+                               dtgtk_cairo_paint_colorpicker_set_values, CPF_STYLE_FLAT, NULL);
     gtk_widget_set_tooltip_text(bd->colorpicker_set_values, _("set the range based on an area from the image\n"
                                                               "drag to use the input image\n"
                                                               "ctrl+drag to use the output image"));
 
-    GtkWidget *res = dtgtk_button_new(dtgtk_cairo_paint_reset, CPF_STYLE_FLAT | CPF_DO_NOT_USE_BORDER, NULL);
+    GtkWidget *res = dtgtk_button_new(dtgtk_cairo_paint_reset, CPF_STYLE_FLAT, NULL);
     gtk_widget_set_tooltip_text(res, _("reset blend mask settings"));
 
-    GtkWidget *inv = dtgtk_button_new(dtgtk_cairo_paint_invert, CPF_STYLE_FLAT | CPF_DO_NOT_USE_BORDER, NULL);
+    GtkWidget *inv = dtgtk_button_new(dtgtk_cairo_paint_invert, CPF_STYLE_FLAT, NULL);
     gtk_widget_set_tooltip_text(inv, _("invert all channel's polarities"));
 
     gtk_box_pack_end(GTK_BOX(header), GTK_WIDGET(res), FALSE, FALSE, 0);
@@ -1670,11 +1670,11 @@ void dt_iop_gui_init_blendif(GtkBox *blendw, dt_iop_module_t *module)
     bd->upper_slider = DTGTK_GRADIENT_SLIDER_MULTIVALUE(dtgtk_gradient_slider_multivalue_new(4));
 
     bd->lower_polarity
-        = dtgtk_togglebutton_new(dtgtk_cairo_paint_plusminus, CPF_STYLE_FLAT | CPF_DO_NOT_USE_BORDER | CPF_BG_TRANSPARENT | CPF_IGNORE_FG_STATE, NULL);
+        = dtgtk_togglebutton_new(dtgtk_cairo_paint_plusminus, CPF_STYLE_FLAT | CPF_BG_TRANSPARENT | CPF_IGNORE_FG_STATE, NULL);
     gtk_widget_set_tooltip_text(bd->lower_polarity, _("toggle polarity. best seen by enabling 'display mask'"));
 
     bd->upper_polarity
-        = dtgtk_togglebutton_new(dtgtk_cairo_paint_plusminus, CPF_STYLE_FLAT | CPF_DO_NOT_USE_BORDER | CPF_BG_TRANSPARENT | CPF_IGNORE_FG_STATE, NULL);
+        = dtgtk_togglebutton_new(dtgtk_cairo_paint_plusminus, CPF_STYLE_FLAT | CPF_BG_TRANSPARENT | CPF_IGNORE_FG_STATE, NULL);
     gtk_widget_set_tooltip_text(bd->upper_polarity, _("toggle polarity. best seen by enabling 'display mask'"));
 
     gtk_box_pack_start(GTK_BOX(upslider), GTK_WIDGET(bd->upper_slider), TRUE, TRUE, 0);
@@ -1836,7 +1836,7 @@ void dt_iop_gui_init_masks(GtkBox *blendw, dt_iop_module_t *module)
     gtk_box_pack_start(GTK_BOX(hbox), bd->masks_combo, TRUE, TRUE, 0);
 
     bd->masks_edit
-        = dtgtk_togglebutton_new(dtgtk_cairo_paint_masks_eye, CPF_STYLE_FLAT | CPF_DO_NOT_USE_BORDER, NULL);
+        = dtgtk_togglebutton_new(dtgtk_cairo_paint_masks_eye, CPF_STYLE_FLAT, NULL);
     g_signal_connect(G_OBJECT(bd->masks_edit), "button-press-event", G_CALLBACK(_blendop_masks_show_and_edit),
                      module);
     gtk_widget_set_tooltip_text(bd->masks_edit, _("show and edit mask elements"));
@@ -1844,7 +1844,7 @@ void dt_iop_gui_init_masks(GtkBox *blendw, dt_iop_module_t *module)
     gtk_box_pack_start(GTK_BOX(hbox), bd->masks_edit, FALSE, FALSE, 0);
 
     bd->masks_polarity
-        = dtgtk_togglebutton_new(dtgtk_cairo_paint_plusminus, CPF_STYLE_FLAT | CPF_DO_NOT_USE_BORDER | CPF_BG_TRANSPARENT | CPF_IGNORE_FG_STATE, NULL);
+        = dtgtk_togglebutton_new(dtgtk_cairo_paint_plusminus, CPF_STYLE_FLAT | CPF_BG_TRANSPARENT | CPF_IGNORE_FG_STATE, NULL);
     gtk_widget_set_tooltip_text(bd->masks_polarity, _("toggle polarity of drawn mask"));
     g_signal_connect(G_OBJECT(bd->masks_polarity), "toggled", G_CALLBACK(_blendop_masks_polarity_callback),
                      module);
@@ -1853,7 +1853,7 @@ void dt_iop_gui_init_masks(GtkBox *blendw, dt_iop_module_t *module)
 
     bd->masks_type[0] = DT_MASKS_GRADIENT;
     bd->masks_shapes[0]
-        = dtgtk_togglebutton_new(dtgtk_cairo_paint_masks_gradient, CPF_STYLE_FLAT | CPF_DO_NOT_USE_BORDER, NULL);
+        = dtgtk_togglebutton_new(dtgtk_cairo_paint_masks_gradient, CPF_STYLE_FLAT, NULL);
     g_signal_connect(G_OBJECT(bd->masks_shapes[0]), "button-press-event",
                      G_CALLBACK(_blendop_masks_add_shape), module);
     gtk_widget_set_tooltip_text(bd->masks_shapes[0], _("add gradient"));
@@ -1862,7 +1862,7 @@ void dt_iop_gui_init_masks(GtkBox *blendw, dt_iop_module_t *module)
 
     bd->masks_type[1] = DT_MASKS_PATH;
     bd->masks_shapes[1]
-        = dtgtk_togglebutton_new(dtgtk_cairo_paint_masks_path, CPF_STYLE_FLAT | CPF_DO_NOT_USE_BORDER, NULL);
+        = dtgtk_togglebutton_new(dtgtk_cairo_paint_masks_path, CPF_STYLE_FLAT, NULL);
     g_signal_connect(G_OBJECT(bd->masks_shapes[1]), "button-press-event",
                      G_CALLBACK(_blendop_masks_add_shape), module);
     gtk_widget_set_tooltip_text(bd->masks_shapes[1], _("add path"));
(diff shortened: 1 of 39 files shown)


Nilvus:
@TurboGit: when this will be ready, please wait that I made a visual review of this one before merging. Thanks.

TurboGit:
@Nilvus : Sure, I'll wait for your go.

Mark-64:
> Maybe we could set the maximum size of by querying the size of a main GUI button

@AlicVB I have implemented this and it work pretty well, it keep thumbtable icons size in sync on all OSs.
I can commit that here, or keep it for another PR.

Mark-64:
@Nilvus 
As I feel guilty to give you more work, I'm looking into CSS refactoring, and it's actually pretty easy at a first sight. Take the top panel, right group of icons for example:
image
they are formatted here:
```
/* Set toolbars buttons in lighttable and darkroom */
#header-toolbar #dt-toggle-button,
#header-toolbar #dt-button,
#footer-toolbar #dt-toggle-button,
#footer-toolbar #dt-button
{
  padding: 0.9em;
  min-height: 0.9em; /* align toolbox button height on comboboxe's one */
  min-width: 0.9em;
  border: 0;
  font-size: .7em;
}
```
With gtk inspector, it's easy to see that (in my Ubuntu system) the specified font-size corresponds to 10.26px. All the em units for this group of widgets are relative to that. Margin is not specified, therefore is inherited and it is found to be 0.
The width of the button will be min-width + 2 x padding = 2.7 x font-size = 2.7 x 10.26 = 28px. 
Heigth will be set automatically by the Gtk fill property.
With the current code, CPF_DO_NOT_USE_BORDER is set to false for those buttons and there is a hardcoded inner margin of 4px, meaning the icon size is 20px.
To replicate this with the new code, you do the following:
set min-width to the wanted size of the content (icon): 20px/10.26px = 1.95em
set padding to the previous value of the inner margin 4px: 4px/10.26px = 0.4em
min-height is unimportant and can be omitted.
This is complaent to Gtk/CSS rules and CSS becomes
```
/* Set toolbars buttons in lighttable and darkroom */
#header-toolbar #dt-toggle-button,
#header-toolbar #dt-button,
#footer-toolbar #dt-toggle-button,
#footer-toolbar #dt-button
{
  padding: 0.4em;
  min-width: 1.95em;
  border: 0;
  font-size: .7em;
}
```
The result matches the current one with pixel accuracy on Ubuntu, but it will also scale well in Windows and other systems because now everything is specified in em units.
I offer myself to do the translation and then you can make a final review.
Is that OK ?

aurelienpierre:
> Some icons not respecting their clip area, also leading to non homogeneous size

Yes, that's called optical margin. https://en.wikipedia.org/wiki/Optical_margin_alignment Please do not mess with that.

For example, if you put a circle of radius r and a square of width/height 2r side by side, the circle will look smaller because only a couple of pixel come close to the margins.

So, last year, I spent some time aligning icons on the optical margins, which indeed doesn't respect strictly the bounding box, but looks better.

aurelienpierre:
> As I feel guilty to give you more work, I'm looking into CSS refactoring, and it's actually pretty easy at a first sight. Take the top panel, right group of icons for example:

On this example, the star needs to be enlarged and raised, the parameters wheel needs to be reduced, and the question mark needs to be reduced and lowered. It should be obvious that the line is wonky.

Mark-64:
> On this example, the star needs to be enlarged and raised, the parameters wheel needs to be reduced, and the question mark needs to be reduced and lowered. It should be obvious that the line is wonky.

Aurélien,
this is the current master
image

and this is the present PR
image

If there are defects, they were already there.

I see you point on optical margin, but if you write an icon paint function 
`dtgtk_cairo_paint_something(cairo_t *cr, gint x, gint y, gint w, gint h, gint flags, void *data)`
that actually paints outside the bounding box (w x h), it might get clipped if you remove the hardcoded fixed margin in pixels (which creates the HiDPI scaling issues), unless you remember to put an adequate padding in CSS.
Maybe we should manage the "density" of the icons inside the painting functions by reducing the ones that look bigger and not by growing the others outside the box.
A couple of constants in each function would do.
That would be HiDPI safe, because is a proportional scaling.

I have no problem to revert all changes in the icon design in this PR and propose them with separate PR, although most of them are fixed of obvious mistakes.

johnny-bit:
@Mark-64 - it's not about the icons being physically on the line, it's about them being visually on the line and as it stands now - they are physically on the line but visually waaaay off in both master and in PR.

I agree with problems regarding bounding or drawing outside bounding box. Solution here (IMO) would be to have for some (if not most) icons 2 (or more) functions - actual draw inside fixed space to fill up to the bounds and other functions to draw icon using shape with proper relation to each other, meaning - proper visual center, proper similar weight etc.

I mean - just look at the icon line and the star immediatelly jumps out as not visually centered, plus the leftmost circle and properties gears are heavier than star and question mark... and question mark itself is lanky/tall compared to others making it's visual weight different than the others So generally the line is unbalanced.

Mark-64:
@aurelienpierre @johnny-bit @Nilvus 
Let's try to solve one problem at a time. Perhaps I put different things in the same PR, my bad.
Forget for a moment the design of icons, I will revert the changes and possibly put in other PR.
There we will discuss about visual alignment / proportions.
The problem that I really want to solve with this PR is a about hardcoded padding/margins that do no respect Gtk/CSS rules and creates bad HiDPI scaling, especially on Windows.
So I will close this and propose a slimmer scope in a new PR

Nilvus:
@Mark-64: I've quickly see the result yesterday but will not have time until this week-end (and few time on) so If you have time on your next PR about that to also adjust CSS (at least main things), it's ok for me. And you have Windows as I've understand but not me. So you could make more tests than me. After that, I could/will of course do a review and improve CSS if needed. I'm waiting now for the new PR.

Nilvus:
Last thing forgotten: indeed, we don't have to care if some things are not technically aligned on background. The most important thing is the visual result. So I agree with @aurelienpierre, we need to be careful about optical margins and some icons. That's could sometimes need specific padding/margins for only some icons, regardless of icons beside.

Nilvus:
> @aurelienpierre @johnny-bit @Nilvus
> Let's try to solve one problem at a time. Perhaps I put different things in the same PR, my bad.
> Forget for a moment the design of icons, I will revert the changes and possibly put in other PR.
> There we will discuss about visual alignment / proportions.
> The problem that I really want to solve with this PR is a about hardcoded padding/margins that do no respect Gtk/CSS rules and creates bad HiDPI scaling, especially on Windows.
> So I will close this and propose a slimmer scope in a new PR

@Mark-64: just an add after reading more discussion about icons here. Most of the icons are (at least in lodpi mode) correctly aligned. I've not check precisely all but probably only those star/help/prefs... icons and icons on thumbnails could be improved and have some optical margins. Most of the other so only need your first goal: respect Gtk/CSS rules to have consistent and good display on both lodpi and Hidpi displays.

So said that, as you propose, having a PR for only that first goal is indeed probably better. Then I suggest that you create an issue for optical margins problems and remaining icons that needs to improve alignment. And see after what can be set with CSS or will need at first some specific Gtk code to set them in CSS. That will avoid too much code work to set them before working on them. What's do you think of that way?

johnny-bit:
For icons themselves, can you please take a look at @houz's [show_icons_in_paint_c.pl](https://github.com/darktable-org/darktable/blob/master/tools/show_icons_in_paint_c.pl)? That should draw icons onto PNG where one could visually inspect them.

Mark-64:
@Nilvus 
Agree that most of the icons look already aligned, with few exceptions like the group/help/overlay/preference as @aurelienpierre pointed out, that need some more optical alignment and balancing. That's not a big problem.
You did a fantastic job with css and DT looks nice in LoDPI and with Linux in general.
With my 4k screen under Ubuntu I get DPI=96 and PPD=2, so DPI factor is 1 and everything looks ok.
But when I switch to Windows I get DPI=166, DPI factor jumps to 1.75 and icons are no longer proportioned, nor aligned, due to the misreading of CSS rules.
But there's more: there are two possible values of hardcoded margin in pixels, controlled by a flag "CPF_DO_NOT_USE_BORDER" which is individually set at button creation.
In some cases icons belonging to the same group (one example is the bottom panel in darkroom), use different values of the flag, I guess for optical alignment and size balancing. But you can imagine what remains of the balancing when you leave Ubuntu or Fedora. 
Now to fix this problem it's not enough to enforce the CSS rules: you need to touch the icons, or otherwise you would have to tune CSS for each one of them, which would be super tedious, if even possible.
Sure we need optical aligment and therefore there must be some overbooking in the icon canvas to allow less dense icons to grow beyond the bounding box without clipping. The problem is complicated by the fact that Gtk+ does not support CSS measures in percent. I'm still investigating how to solve this.
Plus there are other small problems: clipping, painting on top of borders, button that look square even if declared with rounded corners and so on.
So I would say you work unfortunately is not based on solid foundation.
My plan is proceed step by step, like follows:

1. Enforce CSS rules in buttons (this is the easiest part)
2. Manage icon canvas overbooking for optical alignment
3. Fixes for borders and corners not properly drawn
4. Icon fixes and optical balancing
5. CSS reworking
6. Code cleanup (there tens of buttons creation call that need to be modified)

Obviously CSS reworking makes sense only when problems in the code and icons are fixed, or we risk to do it twice.
The problem is that the GUI will look equally good or slightly better only after step 5 for the Linux users (most of the developers), and will look actually worse before. It will be a massive difference for Windows users however.

To be honest I almost regret already to have opened this can of worms.

@johnny-bit Thank you for linking this, it could be very useful.

Nilvus:
> @Nilvus
> Agree that most of the icons look already aligned, with few exceptions like the group/help/overlay/preference as @aurelienpierre pointed out, that need some more optical alignment and balancing. That's not a big problem.

And there's more, as I point this on my #5132 PR. There's some hardcoded things that avoid having a background correctly applied (border settings are ignored) on toggle buttons (here group and help but that's for all toggle buttons ; and there's quite a lot of toggle buttons).

> You did a fantastic job with css and DT looks nice in LoDPI and with Linux in general.
> With my 4k screen under Ubuntu I get DPI=96 and PPD=2, so DPI factor is 1 and everything looks ok.
> But when I switch to Windows I get DPI=166, DPI factor jumps to 1.75 and icons are no longer proportioned, nor aligned, due to the misreading of CSS rules.

Thanks! About Windows, I don't use it (and don't want) so I can't see what's going on this side.

> But there's more: there are two possible values of hardcoded margin in pixels, controlled by a flag "CPF_DO_NOT_USE_BORDER" which is individually set at button creation.

I see that, and this should be changed (as I said on my first point) to allow use all applied CSS rules. But on what I test on some ones to remove that, removing that flag not really work as it's worse after that (some CSS rules applied doesn't apply anymore after that). So there's probably something else to do in button.c and togglebutton.c files. But that's beyond my skills here.

> Now to fix this problem it's not enough to enforce the CSS rules: you need to touch the icons, or otherwise you would have to tune CSS for each one of them, which would be super tedious, if even possible.
> Sure we need optical aligment and therefore there must be some overbooking in the icon canvas to allow less dense icons to grow beyond the bounding box without clipping. The problem is complicated by the fact that Gtk+ does not support CSS measures in percent. I'm still investigating how to solve this.

That's probably tricky yes. But Gtk+ could use CSS measures in percent (but not directly). You could do that by using margins for some things. See last CSS and you will see some margins with comment explain that's will align that this way.

> Plus there are other small problems: clipping, painting on top of borders, button that look square even if declared with rounded corners and so on.

Those problems is about hardcoded remaining things with what I describe above. For example, change CPF_STYLE_FLAG by letting only CPF_BG_TRANSPARENT and there's no more clipping. Other way I found to see what's going wrong there was to test the UI by removing all hover buttons states and I see on all toggle buttons (who don't have border rules applied, that I still have an hover effect, not set by CSS but Gtk code).
Button look square is also same problem: border-radius is shown applied by Gtk Inspector Tool but not visibly applied as it's overwritten by hardcoded sets (what I also describe on top of this post.

> So I would say you work unfortunately is not based on solid foundation.

I agree and only for some icons, and especially toggle ones (and a remaining hardcoding background on just one treeview ("more modules" one in bottom right of darkroom). But if that could be changed on Gtk code correctly, actual CSS could be applied directly (without any changes or just a few minor ones).

> My plan is proceed step by step, like follows:
> 
>     1. Enforce CSS rules in buttons (this is the easiest part)
> 
>     2. Manage icon canvas overbooking for optical alignment
> 
>     3. Fixes for borders and corners not properly drawn
> 
>     4. Icon fixes and optical balancing
> 
>     5. CSS reworking
> 
>     6. Code cleanup (there tens of buttons creation call that need to be modified)

That seems a really good plan.

> Obviously CSS reworking makes sense only when problems in the code and icons are fixed, or we risk to do it twice.

Of course!

> The problem is that the GUI will look equally good or slightly better only after step 5 for the Linux users (most of the developers), and will look actually worse before. It will be a massive difference for Windows users however.
 
> To be honest I almost regret already to have opened this can of worms.

I don't know how this is bad on Windows but I don't see Windows users complains about that. So, that's only adjust and fine-tuned the UI. It's not the most important thing to do even if it would be good to fix that and have an even better and beautiful UI. And I have some new hover effects (some I think would be better and more beautiful) that I would like to do but can't do due to what we're talking here about buttons (especially toggle ones). I can do it for some but that would made the UI even less consistent than it is actually. So, that could only be possible if steps 1 to 4 of your plan is done.

It's up to you to do that (or not). The only thing I would like to be solve quickly is the "more modules" treeview where background color hover effect is not applied because overwritten by Gtk code (absolutely not consistent to all other treeviews where it's correctly applied. For icons/buttons, that's a quite harder work. And we have time as the 3.2 will be out in 7 months only.

PR #5076, opened 2020-05-15
set opacity without zooming during path mask creation

elstoc:
resolves issue when pressing ctrl+scroll before placing a path mask,
whereby the opacity was being changed *and* the image was also being
zoomed.


TurboGit:
Can't reproduce this issue! For me using Ctrl+Scroll while a path mash has been initiated works properly by changing the opacity. Can you give a precise scenario ?

elstoc:
No problem - this is a bit of a corner case that I discovered when looking at the opacity toast message. You need to select a path and ctrl+scroll before placing a node onto the image. The same action for any of the other shapes works fine. 

Presumably this is because, for the other shapes, a default shape is shown under the cursor before it's placed on the image, meaning that the scroll is handled by one of the ``dt_<shape>_events_mouse_scrolled`` functions.

TurboGit:
Stupid me ! I tested with the bush tool and not the path tool :( I see the issue and confirm that your fix works perfect. Thanks.

PR #5080, opened 2020-05-16
change gui->reset bracketing to ++/-- for brevity

dterrahe:
just another small attempt to improve code readability/maintainability and brevity (saving one line and variable each time). using prefix notation to make it easier to spot and check matching on/off switches (search for "gui->reset;" including the semi-colon)

but it is an invasive change (touches 35 files, mostly search+replace) that doesn't provide any advantage _now_ so definitely not high priority

### DIFF
--- a/src/common/imageio_module.c
+++ b/src/common/imageio_module.c
@@ -171,9 +171,9 @@ static int dt_imageio_load_modules_format(dt_imageio_t *iio)
       continue;
     }
     module->gui_data = NULL;
-    if(darktable.gui) darktable.gui->reset = 1;
+    if(darktable.gui) ++darktable.gui->reset;
     module->gui_init(module);
-    if(darktable.gui) darktable.gui->reset = 0;
+    if(darktable.gui) --darktable.gui->reset;
     if(module->widget) g_object_ref(module->widget);
     g_free(libname);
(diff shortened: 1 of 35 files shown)


TurboGit:
This looks good to me. I'm wondering if we want not even to be more clean by declaring two inline routines in gtk.h:

```
   void stop_gui_reset(void) { ++darktable.gui->reset }
   void release_gui_reset(void) { --darktable.gui->reset }
```

And use them ? We can do that later and I have no strong feeling. Maybe the check

`  if(darktable.gui->reset) ..`

Would be less evident to understand, so we may also want to have:

`   gboolean is_gui_released (void) { ... }`

How does this sound to you ?

dterrahe:
I'm happy with whichever approach you prefer. I think the increasing/decreasing counter is a fairly well known pattern and using it explicitly has the advantage that it is obvious, without looking it up in the header file, that two locks and one release do not lead to a full release. And I like the symmetry in the ++/-- that also stand out well and give a bracketing effect. But that is all just aesthetics... :-)

TurboGit:
No strong opinion. This version is better than previous. My idea was to have a single point of implementation of the increase/decrease and cope with possible other implementations. I always find better to use routine for better abstraction. But let's merge this and we can always improve later.

PR #5082, opened 2020-05-16
blendop masks: allow continuous creation with ctrl+click

### LINKED ISSUE #5078: Make tools for drawn masks toggle-able as in the retouch module

Drawn masks often requiere more than on brush stroke (or circle, elipse or path). It would be nice if these tools could be toggled on and off so that they remain active until they are turned off. When making use of pressure sensitivity when working with a tablet it happens that you lose contact easily. Which, in the current mode, means that you first have to activate the tool again which disrupts the flow of drawing. It's also an issue when using the mouse when you want to draw more than a single part mask. Keeping one tool active until it's deselected or another tool is selected seems to be pretty standard for all kind of graphic programs (eg. GIMP, inkscape etc.) so people are used to this behaviour.

The retouch module shows my prefered behaviour for the drawing tools. If possible please copy the behaviour of the drawing tools from the retouch module to drawn masks in general.

elstoc:
adds ability to draw continuous masks (similar to how retouch and spot
removal modules work) by control-clicking on the drawn mask creation
buttons.

Resolves #5078

### DIFF
--- a/src/develop/blend_gui.c
+++ b/src/develop/blend_gui.c
@@ -1002,8 +1002,7 @@ static void _blendop_blendif_invert(GtkButton *button, dt_iop_module_t *module)
   dt_dev_add_history_item(darktable.develop, module, TRUE);
 }
 
-
-static int _blendop_masks_add_shape(GtkWidget *widget, GdkEventButton *event, dt_iop_module_t *self)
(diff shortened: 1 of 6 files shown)


TurboGit:
Works for me!

PR #5086, opened 2020-05-16
Fix/cmake run dt from build dir

codingdave:
This is my proposal for being able to run darktable without installing it, meaning out of binary tree runs. 

Basically it is expected from CMake projects to have a working program in the build directory. Until now the user had to build and install darktable. This PR eliminates the install requirement because the application runs from the binary tree, wherever that is. This makes dt more compatible to CMake capable IDEs like e.g. Visual Studio Code. 

There is just a minimum change to the CMake structure. Basically the idea is instead of using absolute pathes to generate relative pathes from CMake that can be resolved at runtime from both, the installed location and the build-tree. The relative path then is expanded at runtime in C.

I was removing redundancies and created functions for the common part. 

**FIXED:** ~~One change that I am not so sure about happened: Before my PR the function dt_loc_init_tmp_dir returns 1 in case of an error, whereas all the other dt_loc_init_xxx_dir functions were exiting with `exit(EXIT_FAILURE);`. Since I put all that code in one function no 1 is returned any longer for the temporary directory, but a exit is triggered like for any other directory. I think consistency wins but I might have overlooked something here.~~

**FIXED:** ~~Also whats strange for me is that I get `A component of file_name does not name an existing file or file_name points to an empty string:` messages for existing files/directories. I am not sure what causes this (realpath returns ENOENT).~~

I'm happy about feedback.
Kind regards 

------------

Possible startup calls:
* ./darktable
* ./some/path/to/darktable
* ~/some/path/to/darktable
* ~$(username)/some/path/to/darktable
* darktable (from anywhere on the file system while starting darktable from PATH)
* OsX: having the bundle installed and clicking the dt icon
* OsX: having the bundle open (mounted) and starting from the mount position


codingdave:
I have no idea why this fails on travis. I cannot reproduce failing tests locally running ci-script.sh. Any suggesions?

dterrahe:
This definitely looks like something I want. At the moment, on debian unstable, I get:

A component of file_name does not name an existing file or file_name points to an empty string: /opt/darktable.test/bin/../share/darktable
A component of file_name does not name an existing file or file_name points to an empty string: /opt/darktable.test/bin/../lib/darktable
A component of file_name does not name an existing file or file_name points to an empty string: /opt/darktable.test/bin/../share/darktable/locale
directory for darktable.localedir has not been set.

Also, it looks like a lot was installed in the install target directory while I build the "all" target. Maybe I'm confused.

codingdave:
@dterrahe 
Do you get the message `A component of file_name does not name an existing file or file_name points to an empty string` for non-existing files or directories only? I get this also for my existing database. No idea why. 
What seems odd is that `darktable.localedir` wasnt set. Can you please check with -d dev, you should see all currently set/found directories. **Please be aware that you need to have the CWD set to the one from your darktable executable.** 

> Also, it looks like a lot was installed in the install target directory while I build the "all" target. Maybe I'm confused.
This should not have happened due to my PR. Can you please check if the vanilla dt behaves the same? If not, can you please provide the tree of all files that are installed already?

dterrahe:
> What seems odd is that `darktable.localedir` wasnt set. Can you please check with -d dev, you should see all currently set/found directories. **Please be aware that you need to have the CWD set to the one from your darktable executable.**

OK, the problem was with having to start from the binary directory. Then it starts correctly and I get:
```
/opt/darktable.test/bin$ ./darktable --configdir ~/.config/darktable.test -d dev
A component of file_name does not name an existing file or file_name points to an empty string: ./../share/darktable
0.002972 darktable.datadir: /opt/darktable.test/share/darktable
A component of file_name does not name an existing file or file_name points to an empty string: ./../lib/darktable
0.003037 darktable.plugindir: /opt/darktable.test/lib/darktable
A component of file_name does not name an existing file or file_name points to an empty string: ./../share/darktable/locale
0.003057 darktable.localedir: /opt/darktable.test/share/darktable/locale
0.003072 darktable.tmpdir: /tmp
0.003101 darktable.configdir: /home/diederik/.config/darktable.test
0.003121 darktable.cachedir: /home/diederik/.cache/darktable

```

 
> > Also, it looks like a lot was installed in the install target directory while I build the "all" target. Maybe I'm confused.

> This should not have happened due to my PR. Can you please check if the vanilla dt behaves the same? If not, can you please provide the tree of all files that are installed already?

It looks like with your patch, the build directory is almost empty, so everything is built directly into the target directory. I hadn't understood that this was the intention. 

And somehow, after running it once from the shell in its own directory, it now runs within the vscode debugger as well. Not sure what changed the second time :-/

parafin:
That doesn't sound right, building shouldn't write to install directory.

codingdave:
Right, this shouldnt be. I have the issue that I want to detect if I am at install time or build time. Looks like my logic needs improvements. 

**BEGIN-This is my mistake.**
Whats happening currently is that I detect whether CMAKE_INSTALL_PREFIX is set and if it is not set I will use the build directory (CMAKE_CURRENT_BINARY_DIR) as CMAKE_INSTALL_PREFIX. 

Then I populate some variables that are used all over the place:

DARKTABLE_LIBDIR    
DARKTABLE_BINDIR    
DARKTABLE_DATADIR   
DARKTABLE_SHAREDIR  
DARKTABLE_LOCALEDIR 
**END-This is my mistake.**

These variables are used to create their relative version

DARKTABLE_RELATIVE_BINDIR    
DARKTABLE_RELATIVE_LIBDIR    
DARKTABLE_RELATIVE_DATADIR   
DARKTABLE_RELATIVE_SHAREDIR  
DARKTABLE_RELATIVE_LOCALEDIR 

and the relative version is being used in the config.cmake.h to define the runtime relative pathes:
#define DARKTABLE_BINDIR    "./bin"
#define DARKTABLE_LOCALEDIR "../share/darktable/locale"
#define DARKTABLE_LIBDIR    "../lib/darktable"
#define DARKTABLE_DATADIR   "../share/darktable"
#define DARKTABLE_SHAREDIR  "../share/darktable"

So if CMAKE_INSTALL_PREFIX is set files are created outside the build directory. And yes, this should not happen.

coffeeandspaceships:
I think this issue may be in the code base rather than your PR, but it seems strange to have any of the explicit declarations of path that include "/share/" when there is a  DARKTABLE_SHAREDIR and at least one place where some XDG support relies on it. However there do seem to be other places in the code where "share" is appended to a path explicitly again. Maybe it's just not consistent in the 3.0.2 codebase?

codingdave:
I think the cmake project files could be improved in various ways. My goal was to facilitate build tree runs and at the same time touch as few as possible. I see it as an iterative process and every iteration fixes some item. I did not tend to clean up all hardcoded directory structures and 'share' induced. 
Looking at the goal of this PR is there anything you suggest me doing @coffeeandspaceships?

coffeeandspaceships:
I completely agree with your point that any movement towards better consistency is good.

I think that's a different goal to "creating a portable build" which I think is the goal in userland here for you and @dterrahe ?

BUT I think you cannot hit your userland goal without also improving the consistency generally, so everybody wins. To me it's worth splitting it into two PRs - first the CMake improvements you're suggesting, and then the second one is to create that portable build, for at least one platform (are you both on Linux?)

codingdave:
I'm not sure I want to create a portable build, but it might be very similar (I guess portable build e.g. also store the config files in the same folder structure and might have other changes as well).

What do you think needs to get improved in terms of consistency?

I'm running Linux (Manjaro xfce).

src/common/grealpath.h
```diff
@@ -20,7 +23,35 @@ static inline gchar *g_realpath(const char *path)
 #define PATH_MAX 4096
 #endif
   char buffer[PATH_MAX] = { 0 };
-  if(realpath(path, buffer))
+  char* res = realpath(path, buffer);
+  
+  if ( errno == EACCES ) {
+    printf("Read or search permission was denied for a component of file_name: %s\n", path);
+    exit(EXIT_FAILURE);
```
houz, line 30: NOOOOO! Why would you call `exit`? If you really think that error handling is required then imitate what glib does and use GError to communicate that to the user.

CMakeLists.txt
```diff
@@ -162,7 +174,7 @@ function(generate_version_gen_c)
 
   add_custom_target(
     create_version_gen ALL
-    COMMAND ${CMAKE_SOURCE_DIR}/tools/create_version_c.sh ${CMAKE_BINARY_DIR}/src/version_gen.c ${_VERSION}
+    COMMAND ${CMAKE_SOURCE_DIR}/tools/create_version_c.sh ${DARKTABLE_BINDIR}/version_gen.c ${_VERSION}
```
houz, line 177: Source files don't belong in the `bin/` dir.

src/common/grealpath.h
codingdave: We're talking about functionality that is called on the initialization of darktable. This happens before the UI starts up. How can we react instead of prompting the reason to the terminal?
If I don't exit here the program will either die a few lines later because it tries to load .so files at invalid pathes or it will look weird because it does not find optional resources. Either way the error should be shown before the UI initializes.

Do you have a better proposal?

CMakeLists.txt
codingdave: It's a generated file, so not real source code. It's also not installed but placed in the build tree, where intermediate files belong.

coffeeandspaceships:
It feels like you've ended up reimplementing a lot of "install" logic completely, which will be hard to maintain, and may probably break some other existing special cases. It seems more like this is a special case of "install" and the existing logic just needs to be extended rather than reinvented.

I guess the problem is the "special case" might be so special it's hard to switch on with a define; maybe this should be made into a different make target completely?

I think this is where my first comment comes into it... you've found a few general improvements which will make doing anything else to CMake easier so maybe get those merged first in a seperate PR and the rest of it will be easier to unpick. You could cherry-pick those across into a new branch.

CMakeLists.txt
codingdave: btw, I did not change the logic to how it was before. The only change is that the old `src` is now `bin`, which was necessary to have a structure in the build-tree like in the install-tree.

old: CMAKE_BINARY_DIR/src: xxx/darktable.git/build/src,
new: DARKTABLE_BINDIR: xxx/darktable.git/build/bin

codingdave:
@coffeeandspaceships you `feel` that, ok. But please look at the code and tell me where I `reinvent the wheel`. 

The only data I copy is THEME_FILES, PIXMAP_FILES, DARKROOM_PLUGIN_PIXMAPS, OTHER_PIXMAPS,  ${RAWSPEED_DATADIR}/cameras.xml ${RAWSPEED_DATADIR}/showcameras.xsl, which are resources that need to lie next to the executable. The rest is done by tuning the build-directory structure. 

> and may probably break some other existing special cases
Which ones? please be specific.

codingdave:
Hi @TurboGit, this PR has finished 7 days ago. How to continue? 

2 Items are open for discussion and I would like to have an answer from you: 
* @houz suggests not to `exit` on detected error cases, but I think this delays an error case and I like to be explicit (better exit with a message that describes the error then fail later without a real pointer to the reason).
* I am not sure why realpath returns ENOENT (`A component of file_name does not name an existing file or file_name points to an empty string`) for existing files. I can remove the message because it seems wrong (the file/directory exists and is not a symlink and dt works fine even with ENOENT returned).

parafin:
ENOENT is probably not returned from realpath. Generally system calls like that work the following way:

- if there's no error, errno is not modified and some valid return value is returned (e.g. not NULL in this case)
- in case of error some special invalid value is returned (here it's NULL) and errno is set to error code

So you either have to reset errno to zero before calling realpath, or don't look at errno if realpath didn't return NULL.

codingdave:
Thanks for the info @parafin. I will test if that's related to realpath.

codingdave:
@parafin I've read that the preferred way is to ignore errno if there is no error. So I modified the code to only read that out on error (if the result is NULL). I can confirm that the strange behavior is gone. Thank you!

codingdave:
@TurboGit it seems to work fine for some users (IRC) and I have resolved the open issues I had pointed to in the PR initial comment. Please let me know whats required to merge this PR.

Kind regards

src/common/file_location.c
```diff
@@ -137,58 +157,122 @@ char *dt_loc_find_install_dir(const char *suffix, const char *searchname)
...
+  DIR* dir = opendir(directory);
+  if (dir) {
+    dt_print(DT_DEBUG_DEV, "%s: %s\n", text, directory);
+    closedir(dir);
+  } 
+  else 
+  {
+    if ( errno == EACCES ) 
+    {
+      printf("Permission denied.\n");
```
TurboGit, line 205: All errors should be sent to stderr -> fprintf(stderr, ...)

And message should not use capital-letter (sorry that's the current convention) and last but not least the messages should be made translatable _(" ... ");.

src/common/file_location.c
```diff
@@ -137,58 +157,122 @@ char *dt_loc_find_install_dir(const char *suffix, const char *searchname)
...
+    } 
+    else if ( errno == ENOENT ) 
+    {
+      printf("%s %s does not exist.\n", text, directory);
```
TurboGit, line 209: Likewise for all printf below.

src/common/grealpath.h
```diff
@@ -20,10 +23,48 @@ static inline gchar *g_realpath(const char *path)
...
+  {
     return g_strdup(buffer);
-  else
-    return NULL;
+  }
+  else 
+  {
+    if ( errno == EACCES ) 
+    {
+      printf("Read or search permission was denied for a component of file_name: %s\n", path);
```
TurboGit, line 36: Likewise, printf() should be fprintf on stderr. All printf below too. + translatable.

src/common/file_location.c
```diff
@@ -137,58 +157,122 @@ char *dt_loc_find_install_dir(const char *suffix, const char *searchname)
...
+    } 
+    else if ( errno == EBADF ) 
+    {
+      printf("fd is not a valid file descriptor opened for reading.\n");
```
TurboGit, line 213: But wait aren't all those strings already available and translated when using strerror() ?

src/common/file_location.c
```diff
@@ -137,58 +157,122 @@ char *dt_loc_find_install_dir(const char *suffix, const char *searchname)
...
+    else if ( errno == ENOTDIR ) 
+    {
+      printf("Name is not a directory.\n");
+    } 
+    else 
+    {
+      printf("opendir() failed for some other reason.\n");
+    }
+
+    if(exit_on_error)
```
TurboGit, line 240: If this is the controversial exit() I am fine with it. For all errors above there is no way to have dt running properly. Better exit sooner than crashing later.

src/common/file_location.c
parafin: Do we really want to translate error messages printed on the console?

TurboGit: @parafin : I was sure that we already did that, but in fact no, no error messages on the console are translated and this is even better indeed. @codingdave : forget about the translation part, @parafin is correct we don't want that. Thanks.

src/common/file_location.c
codingdave: @TurboGit I wasnt aware of strerror(). I can remove all my messages and replace them with strerror() if thats ok for you. The less code the better. I just dont quite understand `And message should not use capital-letter`, does that mean I have to lowercase every character in my messages or just not the first character of the message?

codingdave: what about "NAME_MAX" and "PATH_MAX"?

TurboGit: Should be kept in capital letters. But I suspect all this will be remove if using strerror()

TurboGit: > does that mean I have to lowercase every character in my messages or just not the first character of the message

Yes, look at the current code, all strings are lowercase. This is controversial and would be in favor of changing it, but that's the style at the moment and we want to keep that consistent.

codingdave:
I have
* changed `printf` to `fprintf(stderr, "...")`
* removed uppercase letters
Thanks for the feedback.

codingdave:
@TurboGit 
Just for fun, this code:
`  fprintf(stderr, "EACCES: %s\n", strerror(EACCES));
  fprintf(stderr, "EINVAL: %s\n", strerror(EINVAL));
  fprintf(stderr, "EIO: %s\n", strerror(EIO));
  fprintf(stderr, "ELOOP: %s\n", strerror(ELOOP));
  fprintf(stderr, "ENAMETOOLONG: %s\n", strerror(ENAMETOOLONG));
  fprintf(stderr, "ENOENT: %s\n", strerror(ENOENT));
  fprintf(stderr, "ENOTDIR: %s\n", strerror(ENOTDIR));
  fprintf(stderr, "ENOMEM: %s\n", strerror(ENOMEM));
  fprintf(stderr, "EBADF: %s\n", strerror(EBADF));
  fprintf(stderr, "EMFILE: %s\n", strerror(EMFILE));
  fprintf(stderr, "ENFILE: %s\n", strerror(ENFILE));`
produces this output:
`EACCES: Permission denied
EINVAL: Invalid argument
EIO: Input/output error
ELOOP: Too many levels of symbolic links
ENAMETOOLONG: File name too long
ENOENT: No such file or directory
ENOTDIR: Not a directory
ENOMEM: Cannot allocate memory
EBADF: Bad file descriptor
EMFILE: Too many open files
ENFILE: Too many open files in system
`
So using strerror I get different messages with uppercase content. Which way to go?

TurboGit:
> So using strerror I get different messages with uppercase content. Which way to go?

That's not a dt message per say, so that's fine to have capital letter in message coming from other part of the systems, especially on the console where this is only/mostly for devs.

codingdave:
@TurboGit all requested changes have been applied. Though I see "changes requested". Is there some button I dont see?

johnny-bit:
@codingdave scroll up, at the top on the right side there'll be list of people who left comments on code. click re-request review near @TurboGit name :) should be fine.

src/common/darktable.c
```diff
@@ -756,9 +756,27 @@ int dt_init(int argc, char *argv[], const gboolean init_gui, const gboolean load
 #ifdef _OPENMP
   omp_set_num_threads(darktable.num_openmp_threads);
 #endif
-  dt_loc_init_datadir(datadir_from_command);
-  dt_loc_init_plugindir(moduledir_from_command);
-  dt_loc_init_localedir(localedir_from_command);
+
+  // get the application directory from the first argument containing the full path to the executable
```
parafin, line 760: This is actually incorrect - argv[0] contains full path only if you start the program by using the full path. E.g. if you start it with just darktable, "darktable" is exactly what you get in argv[0].

CMakeLists.txt
```diff
@@ -44,6 +44,18 @@ include(DefineOptions.cmake)
...
+set(DARKTABLE_DATADIR   "${CMAKE_BINARY_DIR}/share/darktable") # darktablerc, themes, icons, ... Cannot use ${CMAKE_BINDIR}/data because not everything resides there. Like icons (pixmaps)
+set(DARKTABLE_SHAREDIR  "${DARKTABLE_DATADIR}")
+set(DARKTABLE_LOCALEDIR "${CMAKE_BINARY_DIR}/share/locale")
+
+set(DARKTABLE_RELATIVE_BINDIR    "./${CMAKE_INSTALL_BINDIR}")
+set(DARKTABLE_RELATIVE_LIBDIR    "../${CMAKE_INSTALL_LIBDIR}/darktable")
+set(DARKTABLE_RELATIVE_DATADIR   "../${CMAKE_INSTALL_DATAROOTDIR}/darktable")
+set(DARKTABLE_RELATIVE_SHAREDIR  "${DARKTABLE_RELATIVE_DATADIR}")
+set(DARKTABLE_RELATIVE_LOCALEDIR "../share/locale")
+
```
parafin, line 58: In all likelihood this approach breaks macOS bundle, since directory structure is different in it compared to installed version - binary is MacOS/darktable, while everything else is in Resources/.

data/CMakeLists.txt
```diff
@@ -118,7 +119,7 @@ if(APPLE)
   # OS X doesn't have aplay, but afplay should do
   set(DEFCONFIG_AUDIOPLAYER "afplay")
 endif(APPLE)
-configure_file( ${CMAKE_CURRENT_SOURCE_DIR}/darktableconfig.dtd ${CMAKE_CURRENT_BINARY_DIR}/darktableconfig.dtd COPYONLY )
+configure_file(${CMAKE_CURRENT_SOURCE_DIR}/darktableconfig.dtd ${DARKTABLE_DATADIR}/darktableconfig.dtd COPYONLY)
```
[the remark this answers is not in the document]
codingdave: The idea is to express explicitly where the files are going to reside independent of the context: 1. CMAKE_CURRENT_BINARY_DIR is not a constant and such changes on context, (meaning the same variable has a different value in a different CMakeLists.txt)
2. I changed the value of CMAKE_CURRENT_BINARY_DIR by modifying `add_subdirectory(src_directory)` to `add_subdirectory(src_directory $New_binary_directory_added_with_this_pr)`, meaning even in the same CMakeLists.txt the variable has a different value now than it had before.

> Doesn't CMAKE_CURRENT_BINARY_DIR equal to DARKTABLE_DATADIR here now?
Yes, that is correct and you might argue the change was not necessary. I did this at some point in time to reduce the complexity of the complete task. You can see on the amount of commits and the date of the first one that the PR was not easy and I had to strip it into very small parts to keep the control (I think I had 4 failed attempts before).

CMakeLists.txt
```diff
@@ -487,7 +499,7 @@ endif(BUILD_TESTING)
 
 # setup some theme specific variables
 set(THEME hicolor)
-set(THEME_DIRECTORY "${CMAKE_INSTALL_DATAROOTDIR}/icons/${THEME}")
+set(THEME_DIRECTORY "${DARKTABLE_RELATIVE_DATADIR}/../icons/${THEME}")
```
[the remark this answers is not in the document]
codingdave: > I think that's the only usage of DARKTABLE_RELATIVE_DATADIR in cmake files. 

correct

> It looks like you intended DARKTABLE_RELATIVE only for usage in config.h

I'm a big fan of explicit and understandable code. I agree that it does not look nice. But what is the alternative?

`set(THEME_DIRECTORY "${DARKTABLE_RELATIVE_DATADIR}/../icons/${THEME}")`
=> `set(THEME_DIRECTORY "../${CMAKE_INSTALL_DATAROOTDIR}/darktable/../icons/${THEME}")`
=>  `set(THEME_DIRECTORY "../${CMAKE_INSTALL_DATAROOTDIR}/icons/${THEME}")`

And honestly, I prefer the current version over the expanded ones

CMakeLists.txt
```diff
@@ -44,6 +44,18 @@ include(DefineOptions.cmake)
 include(GNUInstallDirs)
 include(FeatureSummary)
 
+set(DARKTABLE_LIBDIR    "${CMAKE_BINARY_DIR}/lib/darktable")
+set(DARKTABLE_BINDIR    "${CMAKE_BINARY_DIR}/bin")
+set(DARKTABLE_DATADIR   "${CMAKE_BINARY_DIR}/share/darktable") # darktablerc, themes, icons, ... Cannot use ${CMAKE_BINDIR}/data because not everything resides there. Like icons (pixmaps)
+set(DARKTABLE_SHAREDIR  "${DARKTABLE_DATADIR}")
+set(DARKTABLE_LOCALEDIR "${CMAKE_BINARY_DIR}/share/locale")
+
+set(DARKTABLE_RELATIVE_BINDIR    "./${CMAKE_INSTALL_BINDIR}")
```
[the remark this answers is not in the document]
codingdave: good catch. Done.

CMakeLists.txt
```diff
@@ -44,6 +44,18 @@ include(DefineOptions.cmake)
...
+set(DARKTABLE_RELATIVE_LIBDIR    "../${CMAKE_INSTALL_LIBDIR}/darktable")
+set(DARKTABLE_RELATIVE_DATADIR   "../${CMAKE_INSTALL_DATAROOTDIR}/darktable")
+set(DARKTABLE_RELATIVE_SHAREDIR  "${DARKTABLE_RELATIVE_DATADIR}")
+set(DARKTABLE_RELATIVE_LOCALEDIR "../share/locale")
```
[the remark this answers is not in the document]
codingdave: I believe so. Thanks.

src/common/darktable.c
codingdave: Indeed, the comment was wrong (fixed).

CMakeLists.txt
```diff
@@ -44,6 +44,18 @@ include(DefineOptions.cmake)
...
 
+set(DARKTABLE_LIBDIR    "${CMAKE_BINARY_DIR}/lib/darktable")
+set(DARKTABLE_BINDIR    "${CMAKE_BINARY_DIR}/bin")
+set(DARKTABLE_DATADIR   "${CMAKE_BINARY_DIR}/share/darktable") # darktablerc, themes, icons, ... Cannot use ${CMAKE_BINDIR}/data because not everything resides there. Like icons (pixmaps)
+set(DARKTABLE_SHAREDIR  "${DARKTABLE_DATADIR}")
+set(DARKTABLE_LOCALEDIR "${CMAKE_BINARY_DIR}/share/locale")
+
+set(DARKTABLE_RELATIVE_BINDIR    "./${CMAKE_INSTALL_BINDIR}")
+set(DARKTABLE_RELATIVE_LIBDIR    "../${CMAKE_INSTALL_LIBDIR}/darktable")
+set(DARKTABLE_RELATIVE_DATADIR   "../${CMAKE_INSTALL_DATAROOTDIR}/darktable")
```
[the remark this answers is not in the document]
codingdave: Is it even possible? Do you have any idea how to tackle that @parafin?

CMakeLists.txt
codingdave: @parafin  Hmm, I have no Mac. Can you test that?

CMakeLists.txt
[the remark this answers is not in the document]
parafin: I see 2 options:

1. add a check that CMAKE_INSTALL_BINDIR doesn't contain / or \ and fail with fatal error message if it does
2. generate correct relative paths by checking how deep is CMAKE_INSTALL_BINDIR (how many directory separators it contains)

CMakeLists.txt
parafin: I will when I have time, I can't promise when.

CMakeLists.txt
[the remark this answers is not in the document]
parafin: It seems to me that THEME_DIRECTORY isn't used anywhere. THEME is used only in data/pixmaps/CMakeLists.txt where directory is constructed from DARKTABLE_DATADIR now. So I don't understand what you are doing here. Just remove this variable I guess.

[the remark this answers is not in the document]
codingdave: fixed

CMakeLists.txt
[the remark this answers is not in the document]
codingdave: `DARKTABLE_RELATIVE_BINDIR` does not exist any longer and `CMAKE_INSTALL_BINDIR` is not used any more. It it still needed? If so, can you give me some hypothetical pathes that my PR should be able to work with?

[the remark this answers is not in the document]
parafin: Erhm, what are relative paths relative to? You use CMAKE_INSTALL_BINDIR implicitly here. I already gave you a hypothetical path in first comment.

I'm fine with just erroring out. But until macOS case is tested/fixed it's not obvious if this approach going to work at all.

src/common/darktable.c
parafin: If this comment is wrong, then how does it work when darktable is started with just "darktable"? Have you tested this (most common BTW) case? Because you have removed hard-coded absolute paths from the code.

codingdave: On my linux box I have tested 

- ./darktable
- ./some/path/to/darktable
- ~/some/path/to/darktable
- ~username/some/path/to/darktable

and that all works fine. Now you point me on a usecase I did not test yet 

- darktable (while having the location in the PATH)
And this does not work as in this case application_directory resolves to NULL. I will have to fix that using /proc/self/exe on linux and GetModuleHandle on windows.

codingdave: @parafin https://stackoverflow.com/a/1024937/1399018 this pretty nicely covers the current state and the upcoming problems. Any thoughts on this?

parafin: I would say "good luck";) For macOS bundle it is solved in dt_osx_get_bundle_res_path.

codingdave: Why reinventing the wheel when there are libraries that take over that task. I decided for [whereami](https://github.com/gpakosz/whereami).

codingdave: Should be solved now. I have tested various versions and it works for now. No Idea yet about other Os. But the lib states:

    Supported platforms:

    Windows
    Linux
    Mac
    iOS
    Android
    QNX Neutrino
    FreeBSD
    NetBSD
    DragonFly BSD
    SunOS


I also don't know if the license adds any burden to our project. There is a MIT and a WTFPLv2  license available. @TurboGit from the license side the library works for us, right?

CMakeLists.txt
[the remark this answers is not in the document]
codingdave: I need a complete picture to rule out misunderstandings. Let me start an example based on your suggestion:
* /games/bin
* /lib/darktable/{libdarktable.so, plugins/, views}
* /share/{appdata, applications, darktable,  doc, locale, man}
So in this example using `..` would bring us to `/games`, not `/` and such the path resolution would fail and you suggest to count the directoriy levels from `CMAKE_INSTALL_BINDIR` to adjust for that - got it. So in the example above we would need `../..`.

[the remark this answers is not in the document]
parafin: Yes, you are correct.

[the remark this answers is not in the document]
codingdave: I have recreated the path resolution using no hardcoded pathes. Everything is resolved from system specific cmake variables. I have left debug messages (turned off) in the code and will remove them when the code is tested. It should now cover all possible valid Posix use cases.

CMakeLists.txt
codingdave: I will resolve that because its not on my side anymore.

codingdave:
@parafin how about the new approach? Do you see any issues?

parafin:
Didn't have time to test yet. Maybe next week.

MStraeten:
some findings on osx:
starting darktable from build directory: `GSETTINGS_SCHEMA_DIR=/opt/local/share/glib-2.0/schemas/ XDG_DATA_DIRS=/opt/local/share ./bin/darktable`
1. the language setting is ignored
2. after `sudo make install` darktable start fails with:

```
dt_gui_load_theme: error parsing /usr/local/share/themes/darktable.css: <broken file>:1:0Failed to import: Error opening file /usr/local/share/themes/darktable.css: No such file or directory
ERROR: can't init develop system, aborting.
```

3. after packaging a dmg starting darktable.app (`/Volumes/darktable/darktable.app/Contents/MacOS/darktable`) fails:
```
path lookup '/Volumes/darktable/darktable.app/Contents/MacOS//Volumes/darktable/darktable.app/Contents/Resources/share/darktable' fails with: 'No such file or directory'
```

codingdave:
> 
> 
> some findings on osx:
> starting darktable from build directory: `GSETTINGS_SCHEMA_DIR=/opt/local/share/glib-2.0/schemas/ XDG_DATA_DIRS=/opt/local/share ./bin/darktable`
> 
>     1. the language setting is ignored
> 
>     2. after `sudo make install` darktable start fails with:
> 
> 
> ```
> dt_gui_load_theme: error parsing /usr/local/share/themes/darktable.css: <broken file>:1:0Failed to import: Error opening file /usr/local/share/themes/darktable.css: No such file or directory
> ERROR: can't init develop system, aborting.
> ```
> 
>     1. after packaging a dmg starting darktable.app (`/Volumes/darktable/darktable.app/Contents/MacOS/darktable`) fails:
> 
> 
> ```
> path lookup '/Volumes/darktable/darktable.app/Contents/MacOS//Volumes/darktable/darktable.app/Contents/Resources/share/darktable' fails with: 'No such file or directory'
> ```

Hi @MStraeten,
thanks for testing. 
What are your build parameters? Did you modify CMAKE_INSTALL_PREFIX? If so, what did you enter?
Can you please start darktable with `-d dev` and paste the directories that are listed?

MStraeten:
cmake setting according to osx build instructions:
```
cmake .. \
 -DCMAKE_OSX_DEPLOYMENT_TARGET=10.13\
 -DCMAKE_CXX_FLAGS=-stdlib=libc++ \
 -DOpenMP_C_INCLUDE_DIR=/opt/local/include/libomp \
 -DOpenMP_CXX_INCLUDE_DIR=/opt/local/include/libomp \
 -DCMAKE_LIBRARY_PATH=/opt/local/lib/libomp \
 -DBINARY_PACKAGE_BUILD=ON \
 -DRAWSPEED_ENABLE_LTO=ON \
 -DBUILD_CURVE_TOOLS=ON \
 -DBUILD_NOISE_TOOLS=ON \
 -DCMAKE_BUILD_TYPE=RelWithDebInfo
```
starting from build/bin
```
martinstraeten@Martins-MBP build % GSETTINGS_SCHEMA_DIR=/opt/local/share/glib-2.0/schemas/ XDG_DATA_DIRS=/opt/local/share ./bin/darktable -d dev

(process:54924): GLib-GObject-CRITICAL **: 08:46:19.289: g_object_set: assertion 'G_IS_OBJECT (object)' failed
0,016176 application_directory: /Users/martinstraeten/src/darktable_debug/build/bin
0,016310 darktable.plugindir: /Users/martinstraeten/src/darktable_debug/build/lib
0,016425 darktable.tmpdir: /private/var/folders/d7/8trmr7s56xng0013rpzm2ynr0000gn/T
0,016473 darktable.configdir: /Users/martinstraeten/.config/darktable
0,016516 darktable.cachedir: /Users/martinstraeten/.cache/darktable
```
starting after sudo make install:
```
martinstraeten@Martins-MBP build % GSETTINGS_SCHEMA_DIR=/opt/local/share/glib-2.0/schemas/ XDG_DATA_DIRS=/opt/local/share darktable -d dev

(process:56416): GLib-GObject-CRITICAL **: 08:53:17.094: g_object_set: assertion 'G_IS_OBJECT (object)' failed
0,008498 application_directory: /usr/local/bin
0,008651 darktable.plugindir: /usr/local/lib
0,008776 darktable.tmpdir: /private/var/folders/d7/8trmr7s56xng0013rpzm2ynr0000gn/T
0,008831 darktable.configdir: /Users/martinstraeten/.config/darktable
0,008880 darktable.cachedir: /Users/martinstraeten/.cache/darktable
[dt_pthread_create] info: bumping pthread's stacksize from 524288 to 2097152
<...>
[dt_pthread_create] info: bumping pthread's stacksize from 524288 to 2097152

(darktable:56416): GLib-GObject-WARNING **: 08:53:17.218: invalid cast from 'GtkMenuBar' to 'GtkWindow'

(darktable:56416): Gtk-CRITICAL **: 08:53:17.218: gtk_window_add_accel_group: assertion 'GTK_IS_WINDOW (window)' failed
dt_gui_load_theme: error parsing /usr/local/share/themes/darktable.css: <broken file>:1:0Failed to import: Error opening file /usr/local/share/themes/darktable.css: No such file or directory
ERROR: can't init develop system, aborting.
```
starting from mounted package:
```martinstraeten@Martins-MBP build % 
/Volumes/darktable/darktable.app/Contents/MacOS/darktable -d dev

(process:54948): GLib-GObject-CRITICAL **: 08:48:04.939: g_object_set: assertion 'G_IS_OBJECT (object)' failed
0,063712 application_directory: /Volumes/darktable/darktable.app/Contents/MacOS
path lookup '/Volumes/darktable/darktable.app/Contents/MacOS//Volumes/darktable/darktable.app/Contents/Resources/share/darktable' fails with: 'No such file or directory'
```
starting from package in /Applications:
```
martinstraeten@Martins-MBP build % /Applications/darktable.app/Contents/MacOS/darktable -d dev

(process:54956): GLib-GObject-CRITICAL **: 08:49:14.750: g_object_set: assertion 'G_IS_OBJECT (object)' failed
0,009710 application_directory: /Applications/darktable.app/Contents/MacOS
0,010419 darktable.plugindir: /Applications/darktable.app/Contents/MacOS/Applications/darktable.app/Contents/Resources/lib/darktable
0,010687 darktable.tmpdir: /private/var/folders/d7/8trmr7s56xng0013rpzm2ynr0000gn/T
0,010744 darktable.configdir: /Users/martinstraeten/.config/darktable
0,010784 darktable.cachedir: /Users/martinstraeten/.cache/darktable
[dt_pthread_create] info: bumping pthread's stacksize from 524288 to 2097152
<....>
[dt_pthread_create] info: bumping pthread's stacksize from 524288 to 2097152

(darktable:54956): GLib-GObject-WARNING **: 08:49:14.881: invalid cast from 'GtkMenuBar' to 'GtkWindow'

(darktable:54956): Gtk-CRITICAL **: 08:49:14.881: gtk_window_add_accel_group: assertion 'GTK_IS_WINDOW (window)' failed
dt_gui_load_theme: error parsing /Applications/darktable.app/Contents/MacOS/Applications/darktable.app/Contents/Resources/share/darktable/themes/darktable.css: <broken file>:1:0Failed to import: Error opening file /Applications/darktable.app/Contents/MacOS/Applications/darktable.app/Contents/Resources/share/darktable/themes/darktable.css: No such file or directory
ERROR: can't init develop system, aborting.
```

codingdave:
@MStraeten can you please check if the file `darktable.css` is where it is being looked for? 
Any if you add it, does it work?

> 1. the language setting is ignored
I have no clue why this should be related to this PR. Are you sure it works fine without these changes?

MStraeten:
darktable.css is in /Volumes/darktable/darktable.app/Contents/Resources/**share/darktable/themes**/darktable.css
instead of <...>**/share/themes**/darktable.css
directory_structure
symlinking the themes directory just removes the darktable.css error message:
```
martinstraeten@Martins-MBP build % sudo ln -s /usr/local/share/darktable/themes /usr/local/share
martinstraeten@Martins-MBP build % GSETTINGS_SCHEMA_DIR=/opt/local/share/glib-2.0/schemas/ XDG_DATA_DIRS=/opt/local/share darktable -d dev

(process:23810): GLib-GObject-CRITICAL **: 09:30:49.533: g_object_set: assertion 'G_IS_OBJECT (object)' failed
0,008687 application_directory: /usr/local/bin
0,008831 darktable.plugindir: /usr/local/lib
0,008960 darktable.tmpdir: /private/var/folders/d7/8trmr7s56xng0013rpzm2ynr0000gn/T
0,009015 darktable.configdir: /Users/martinstraeten/.config/darktable
0,009056 darktable.cachedir: /Users/martinstraeten/.cache/darktable
[dt_pthread_create] info: bumping pthread's stacksize from 524288 to 2097152
[dt_pthread_create] info: bumping pthread's stacksize from 524288 to 2097152
[dt_pthread_create] info: bumping pthread's stacksize from 524288 to 2097152
[dt_pthread_create] info: bumping pthread's stacksize from 524288 to 2097152
[dt_pthread_create] info: bumping pthread's stacksize from 524288 to 2097152
[dt_pthread_create] info: bumping pthread's stacksize from 524288 to 2097152
[dt_pthread_create] info: bumping pthread's stacksize from 524288 to 2097152
[dt_pthread_create] info: bumping pthread's stacksize from 524288 to 2097152
[dt_pthread_create] info: bumping pthread's stacksize from 524288 to 2097152
[dt_pthread_create] info: bumping pthread's stacksize from 524288 to 2097152
[dt_pthread_create] info: bumping pthread's stacksize from 524288 to 2097152
[dt_pthread_create] info: bumping pthread's stacksize from 524288 to 2097152

(darktable:23810): GLib-GObject-WARNING **: 09:30:49.655: invalid cast from 'GtkMenuBar' to 'GtkWindow'

(darktable:23810): Gtk-CRITICAL **: 09:30:49.655: gtk_window_add_accel_group: assertion 'GTK_IS_WINDOW (window)' failed
ERROR: can't init develop system, aborting.
```

the share content in Resources/share directory are identical in app-packages built with and without the pr

MStraeten:
here a diff between two packages with and without the pr 
[diff.txt](https://github.com/darktable-org/darktable/files/4735332/diff.txt)

codingdave:
Hi @MStraeten 
> cmake setting according to osx build instructions:

Can you please provide a link?

Can you also please rebuild again with the latest patch and report -d dev again? Previously on Mac the debug output for darktable.datadir and darktable.localedir was missing.

MStraeten:
osx build instructions: https://github.com/darktable-org/darktable/blob/master/packaging/macosx/BUILD.txt

MStraeten:
darktable -d dev:
```
0,008433 application_directory: /Users/martinstraeten/src/darktable_debug/build/bin
0,008544 darktable.datadir: /Users/martinstraeten/src/darktable_debug/build/share
0,008611 darktable.plugindir: /Users/martinstraeten/src/darktable_debug/build/lib
0,008672 darktable.localedir: /Users/martinstraeten/src/darktable_debug/build/share/locale
0,008755 darktable.tmpdir: /private/var/folders/d7/8trmr7s56xng0013rpzm2ynr0000gn/T
0,008898 darktable.configdir: /Users/martinstraeten/.config/darktable
0,008943 darktable.cachedir: /Users/martinstraeten/.cache/darktable
```
gui is opened without correct language setting. trying to access preference dialog results in a crash (even after deleting ./config/darktable)

```
Process 45757 stopped
* thread #1, queue = 'com.apple.main-thread', stop reason = EXC_BREAKPOINT (code=EXC_I386_BPT, subcode=0x0)
    frame #0: 0x00000001005a1e8f libglib-2.0.0.dylib`_g_log_abort + 22
libglib-2.0.0.dylib`_g_log_abort:
->  0x1005a1e8f <+22>: addq   $0x8, %rsp
    0x1005a1e93 <+26>: popq   %rbx
    0x1005a1e94 <+27>: popq   %rbp
    0x1005a1e95 <+28>: retq
Target 0: (darktable) stopped.
```

codingdave:
@MStraeten 
since you are using Mac Ports, do you have the necessary env vars set?
https://trac.macports.org/wiki/GNOME
I have fixed an issue. Would be glad if you can give this another try.

MStraeten:
ok, now the language setting is ok when starting from ./bin:
```
build % GSETTINGS_SCHEMA_DIR=/opt/local/share/glib-2.0/schemas/ XDG_DATA_DIRS=/opt/local/share ./bin/darktable -d dev
0.000706 application_directory: /Users/martinstraeten/src/darktable_debug/build/bin
0.000822 darktable.datadir: /Users/martinstraeten/src/darktable_debug/build/share
0.000884 darktable.plugindir: /Users/martinstraeten/src/darktable_debug/build/lib
0.000936 darktable.localedir: /Users/martinstraeten/src/darktable_debug/build/share/locale
0.000989 darktable.configdir: /Users/martinstraeten/.config/darktable
0.001030 darktable.cachedir: /Users/martinstraeten/.cache/darktable
0.001077 darktable.sharedir: /Users/martinstraeten/src/darktable_debug/build/share
0.001147 darktable.tmpdir: /private/var/folders/d7/8trmr7s56xng0013rpzm2ynr0000gn/T
0.001180 new_xdg_data_dirs: /Users/martinstraeten/src/darktable_debug/build/share:/opt/local/share
```
starting from app package (built via `cd ~/src/darktable/packaging/macosx; ./make-app-bundle` ) is ok:
`build % /Volumes/darktable/darktable.app/Contents/MacOS/darktable -d dev`


after `sudo make install` starting darktable fails:
```
build % GSETTINGS_SCHEMA_DIR=/opt/local/share/glib-2.0/schemas/ XDG_DATA_DIRS=/opt/local/share darktable -d dev
0.000794 application_directory: /usr/local/bin
0.000923 darktable.datadir: /usr/local/share
0.000984 darktable.plugindir: /usr/local/lib
0.001082 darktable.localedir: /usr/local/share/locale
0.001153 darktable.configdir: /Users/martinstraeten/.config/darktable
0.001216 darktable.cachedir: /Users/martinstraeten/.cache/darktable
0.001277 darktable.sharedir: /usr/local/share
0.001357 darktable.tmpdir: /private/var/folders/d7/8trmr7s56xng0013rpzm2ynr0000gn/T
0.001394 new_xdg_data_dirs: /usr/local/share:/opt/local/share
...
ERROR: can't init develop system, aborting.
```
darktable -d opencl indicates that also openCl can’t be activated:
```
0.101667 [opencl_init] options for OpenCL compiler: -w  -DUNKNOWN=1 -I/usr/local/share/kernels
0.101705 [opencl_fopen_stat] could not open file `/usr/local/share/kernels/color_conversion.cl'!
0.101708 [opencl_md5sums] could not open file `/usr/local/share/kernels/color_conversion.cl'!
0.101716 [opencl_fopen_stat] could not open file `/usr/local/share/kernels/colorspaces.cl'!
0.101718 [opencl_md5sums] could not open file `/usr/local/share/kernels/colorspaces.cl'!
0.101725 [opencl_fopen_stat] could not open file `/usr/local/share/kernels/colorspace.cl'!
0.101727 [opencl_md5sums] could not open file `/usr/local/share/kernels/colorspace.cl'!
0.101734 [opencl_fopen_stat] could not open file `/usr/local/share/kernels/common.h'!
0.101735 [opencl_md5sums] could not open file `/usr/local/share/kernels/common.h'!
0.101742 [opencl_init] could not open `/usr/local/share/kernels/programs.conf'!
0.101755 [opencl_init] no suitable devices found.
0.101763 [opencl_init] FINALLY: opencl is NOT AVAILABLE on this system.
0.101766 [opencl_init] initial status of opencl enabled flag is OFF.
```

codingdave:
Hi @MStraeten,

I am not able to build the app package. I hope the issue is gone, though.

MStraeten:
running from build dir: OpenCL kernels can’t be built:
```
 build % ./bin/cltest/darktable-cltest
0.033828 [opencl_init] opencl related configuration options:
0.033850 [opencl_init]
0.033852 [opencl_init] opencl: 1
0.033854 [opencl_init] opencl_scheduling_profile: 'default'
0.033856 [opencl_init] opencl_library: ''
<...>
0.033877 [opencl_init]
0.034554 [opencl_init] found opencl runtime library '/System/Library/Frameworks/OpenCL.framework/Versions/Current/OpenCL'
0.034593 [opencl_init] opencl library '/System/Library/Frameworks/OpenCL.framework/Versions/Current/OpenCL' found on your system and loaded
0.034599 [opencl_init] found 1 platform
0.050161 [opencl_init] found 3 devices
0.050194 [opencl_init] discarding CPU device 0 `Intel(R) Core(TM) i9-9880H CPU @ 2.30GHz'.
0.050213 [opencl_init] device 1 `Intel(R) UHD Graphics 630' supports image sizes of 16384 x 16384
0.050215 [opencl_init] device 1 `Intel(R) UHD Graphics 630' allows GPU memory allocations of up to 384MB
[opencl_init] device 1: Intel(R) UHD Graphics 630
     GLOBAL_MEM_SIZE:          1536MB
     MAX_WORK_GROUP_SIZE:      256
     MAX_WORK_ITEM_DIMENSIONS: 3
     MAX_WORK_ITEM_SIZES:      [ 256 256 256 ]
     DRIVER_VERSION:           1.2(May 26 2020 20:51:24)
     DEVICE_VERSION:           OpenCL 1.2
0.052068 [opencl_init] options for OpenCL compiler: -w  -DUNKNOWN=1 -I/Users/martinstraeten/src/darktable_debug/build/bin/share/darktable/kernels
0.052092 [opencl_fopen_stat] could not open file `/Users/martinstraeten/src/darktable_debug/build/bin/share/darktable/kernels/color_conversion.cl'!
0.052095 [opencl_md5sums] could not open file `/Users/martinstraeten/src/darktable_debug/build/bin/share/darktable/kernels/color_conversion.cl'!
0.052101 [opencl_fopen_stat] could not open file `/Users/martinstraeten/src/darktable_debug/build/bin/share/darktable/kernels/colorspaces.cl'!
0.052103 [opencl_md5sums] could not open file `/Users/martinstraeten/src/darktable_debug/build/bin/share/darktable/kernels/colorspaces.cl'!
0.052108 [opencl_fopen_stat] could not open file `/Users/martinstraeten/src/darktable_debug/build/bin/share/darktable/kernels/colorspace.cl'!
0.052109 [opencl_md5sums] could not open file `/Users/martinstraeten/src/darktable_debug/build/bin/share/darktable/kernels/colorspace.cl'!
0.052114 [opencl_fopen_stat] could not open file `/Users/martinstraeten/src/darktable_debug/build/bin/share/darktable/kernels/common.h'!
0.052116 [opencl_md5sums] could not open file `/Users/martinstraeten/src/darktable_debug/build/bin/share/darktable/kernels/common.h'!
0.052140 [opencl_init] could not open `/Users/martinstraeten/src/darktable_debug/build/bin/share/darktable/kernels/programs.conf'! 
```

after `sudo make install `now everything is fine (including OpenCL)

app-package:
`GSETTINGS_SCHEMA_DIR=/opt/local/share/glib-2.0/schemas/ XDG_DATA_DIRS=/opt/local/share /Volumes/darktable/darktable.app/Contents/MacOS/darktable  -d perf -d opencl`
seems to be fine too

codingdave:
That's fine for me. I wanted that darktable runs from the build directory.
I did not want to change all the other executables.

Then I assume my PR is working fine for Mac now. Thanks for testing.

Martin Straeten <notifications@github.com> schrieb am Sa., 13. Juni 2020,
09:30:

MStraeten:
the opencl support is missing, when running from build directory - even for darktable itself.

codingdave:
Ah, that is less optional then. Thanks.

Martin Straeten <notifications@github.com> schrieb am Sa., 13. Juni 2020,
09:45:

codingdave:
Hi @MStraeten,

thanks for testing again. I have 
* Changed other binaries to also appear in the bin directory, so the PR should make all binaries run from build dir now
* checked your observation and don't think it is correct. I have OpenCL support for both, the bin and the install tree. However, I noticed that I had to restart dt 4 times to get all kernels compiled for the install tree (buggy!), and 10 times for the bin tree (also buggy). 

I guess what you have noticed was solely for darktable-cltest (because it was in a further subdirectory the relative path resolution did not work but this is fixed with my latest CMake changes). You might have seen that darktable does not have OpenCL support because you run into the OpenCL Kernel compilation issue I mention above. To verify: Please call `darktable -d opencl`, see where it fails, try again, and see if it fails at the same kernel.

If there is anything not working, please report.

Kind regards
David

MStraeten:
i tested it, it's fine now - darktable-cltest also needed several attempts to build the kernels.
Everything works as expected. Thanks for it.

codingdave:
@TurboGit @parafin, @MStraeten and I have verified the behavior on a Mac and everything seems to work now. Looking forward to the CR.

Kind regards
David

TurboGit:
@codingdave : thanks for all the work. I'll handle the review after 3.2 release. Ping me if I forget :)

MStraeten:
a windows build in mingw environment (following https://github.com/darktable-org/darktable/blob/master/packaging/windows/BUILD.txt) fails:

```
[ 20%] Building CXX object bin/CMakeFiles/lib_darktable.dir/common/exif.cc.obj
cc1plus.exe: warning: command-line option '-Wold-style-declaration' is valid for C/ObjC but not for C++
[ 20%] Building C object bin/CMakeFiles/lib_darktable.dir/common/file_location.c.obj
C:/msys64/home/Me/darktable/src/common/file_location.c: In function 'dt_loc_init_generic':
C:/msys64/home/Me/darktable/src/common/file_location.c:149:12: error: implicit declaration of function 'g_realpath' [-Werror=implicit-function-declaration]
  149 |   result = g_realpath(path);
      |            ^~~~~~~~~~
C:/msys64/home/Me/darktable/src/common/file_location.c:149:10: error: assignment to 'gchar *' {aka 'char *'} from 'int' makes pointer from integer without a cast [-Werror=int-conversion]
  149 |   result = g_realpath(path);
      |          ^
C:/msys64/home/Me/darktable/src/common/file_location.c: In function 'dt_check_opendir':
C:/msys64/home/Me/darktable/src/common/file_location.c:233:3: error: unknown type name 'DIR'
  233 |   DIR* dir = opendir(directory);
      |   ^~~
C:/msys64/home/Me/darktable/src/common/file_location.c:233:14: error: implicit declaration of function 'opendir'; did you mean 'open'? [-Werror=implicit-function-declaration]
  233 |   DIR* dir = opendir(directory);
      |              ^~~~~~~
      |              open
C:/msys64/home/Me/darktable/src/common/file_location.c:233:14: error: initialization of 'int *' from 'int' makes pointer from integer without a cast [-Werror=int-conversion]
C:/msys64/home/Me/darktable/src/common/file_location.c:236:5: error: implicit declaration of function 'closedir'; did you mean 'close'? [-Werror=implicit-function-declaration]
  236 |     closedir(dir);
      |     ^~~~~~~~
      |     close
cc1.exe: all warnings being treated as errors
make[2]: *** [bin/CMakeFiles/lib_darktable.dir/build.make:362: bin/CMakeFiles/lib_darktable.dir/common/file_location.c.obj] Fehler 1
make[1]: *** [CMakeFiles/Makefile2:2277: bin/CMakeFiles/lib_darktable.dir/all] Fehler 2
make: *** [Makefile:172: all] Fehler 2

```
without this pr everything is fine

codingdave:
Thanks @MStraeten,

anything else?

MStraeten:
builds and runs fine on windows10 now thanks

MStraeten:
issue: denoise profiled doesn't use camera profiles when started from build dir ./bin/darktable.

codingdave:
Thanks, I added some more resources and fixed some pathes:
* lua and luarc (added)
* noiseprofile (added)
* watermarks (added)
* darktable.appdata.xml (changed location)

codingdave:
> There are some not documented assumptions that doesn't necessary hold on all systems (e.g. macOS app b

Hi @parafin, can we resolve this?

codingdave:
> 
> 
> Some comments.
> 
> I have not yet tested. I'm no expert on CMake so I would trust you for most of the changes and will test it on my side to see if all goes well.

Hi @TurboGit the earlier we merge that, the easier it will get to fix possible issues before 3.4. Build systems are a b**ch, so please consider merging this early.

parafin:
> > There are some not documented assumptions that doesn't necessary hold on all systems (e.g. macOS app b
> 
> Hi @parafin, can we resolve this? You are very unclear with that statement. Please point the individual items out, s.t. we can look into them and discuss their validity. TY

I'm planning to look at this PR, yes.

src/common/file_location.c
```diff
@@ -109,93 +171,115 @@ char *dt_loc_find_install_dir(const char *suffix, const char *searchname)
...
+{
+#if defined(__APPLE__)
+  char *directory = dt_loc_find_install_dir(DARKTABLE_DATADIR);
+  darktable.datadir = dt_loc_init_generic(datadir, application_directory, directory ? directory : DARKTABLE_DATADIR);
   g_free(directory);
 #else
-  darktable.localedir = dt_loc_init_generic(localedir, DARKTABLE_LOCALEDIR);
+  darktable.datadir = dt_loc_init_generic(datadir, application_directory, DARKTABLE_DATADIR);
 #endif
+  dt_check_opendir("darktable.datadir", darktable.datadir);
```
parafin, line 258: I get an error:
```
$ /Volumes/darktable/darktable.app/Contents/MacOS/darktable -d dev
0.001000 application_directory: /Volumes/darktable/darktable.app/Contents/MacOS
path lookup '/Volumes/darktable/darktable.app/Contents/MacOS//Volumes/darktable/darktable.app/Contents/Resources/../share/darktable' fails with: 'No such file or directory'
```

data/pixmaps/CMakeLists.txt
```diff
@@ -1,18 +1,28 @@
...
 #
 # Install fixed sized icons
 #
 set(FIXED_SIZES "16x16" "22x22" "24x24" "32x32" "48x48" "64x64" "256x256" "scalable")
+set(THEME hicolor)
 foreach(pixmap_size ${FIXED_SIZES})
 	# install apps icons
 	FILE(GLOB PIXMAP_FILES ${CMAKE_CURRENT_SOURCE_DIR}/${pixmap_size}/*.png ${CMAKE_CURRENT_SOURCE_DIR}/${pixmap_size}/*.svg)
-    install(FILES ${PIXMAP_FILES} DESTINATION ${CMAKE_INSTALL_DATAROOTDIR}/icons/${THEME}/${pixmap_size}/apps COMPONENT DTApplication)
+	install(FILES ${PIXMAP_FILES} DESTINATION ${DARKTABLE_DATADIR}/icons/${THEME}/${pixmap_size}/apps COMPONENT DTApplication)
```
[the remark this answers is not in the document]
codingdave: Thank you. 

I can confirm, that /share/icons files are missing with my patch. Diffing `ls -R` output I can see that this is the only difference apart from many localizations that seem to be not part of 3.2 anymore (af, ca, cs, da, el, fi, gl, hu, ja, nb, nl, pt_PT, ro, sk, sq, sv, th, uk, zh_CN, zh_TW)

src/common/file_location.c
codingdave: I was not yet able to build the mac image because of the scripts python 2 state being incompatible to a standard macs python 3 installation. So I had to patch gtk-mac-bundler to use python2 instead of python 3 like so:
```
head ~/.local/bin/gtk-mac-bundler
#!/usr/bin/env python2
```

I can confirm this issue when trying to launch the mac image (/Volumes/darktable/darktable.app/Contents/Resources/share/darktable) or the installed application coming from the mac image (/Applications/darktable.app/Contents/MacOS/darktable).
Since the binary tree and install dir version work fine I guess the error happens when creating of the mac image.

codingdave: The bundle path handling shall be implemented and working now

codingdave:
@parafin @TurboGit anything left open? One month has passed. Please update me on necessary changes, if any. Thank you.

codingdave:
@parafin @TurboGit anything left open? One month has passed. Please update me on necessary changes, if any. Thank you.

TurboGit:
@codingdave : can you rebase and fix conflicts ? If @parafin has no more comments I'll merge. If you agree I'd like to have a single commit here to make a revert easier if needed. So can you also squash everything? TIA.

parafin:
I will try to find time to test this PR today or tomorrow.

src/common/file_location.c
```diff
@@ -82,120 +114,172 @@ gchar *dt_loc_get_home_dir(const gchar *user)
...
+        // <bundleroot>/Contents/Resources/lib
+        // <bundleroot>/Contents/Resources/share
+        // so the relative path from the binary directory to the other directories differs to the non-bundle version by
+        // ../etc -> ../Resources/etc, 
+        // ../lib -> ../Resources/lib, 
+        // ../share -> ../Resources/share, 
+        // So we have to modify the relative default value
+
+        // +2: removes the two dots '..'
+        g_snprintf(complete_path, sizeof(complete_path), "%s/../Resources%s", application_directory, default_value + 2);
```
parafin, line 157: Not very pretty solution since there is an assumption about how default_value is structured. E.g. it won't work if CMAKE_BINARY_DIR is something like "games/bin". The problem is (as I already said) that you define all paths relative to executable location, while it doesn't really matter. What you need is paths relative to prefix (e.g. /usr/local), then you either:
1. use value passed from command line if any;
2. if it's a Mac bundle, use dt_osx_get_bundle_res_path() as prefix (you now ignore what path it actually returns);
3. deduce prefix from executable location.

parafin (changes requested):
- Some cleanup is required (unneeded whitespace changes, debug output).

- Also I'm getting a crash when hovering over any image when using this PR, I'm pretty sure it's unrelated, but please rebase this PR on master branch.

- macOS bundle now works, but I'm not sure the solution you came up is correct way to do it (see review comments). If it's too hard to fix, I'm fine with merging the current version, since app bundle building requires quite fixed environment anyway, and it works as is.

codingdave:
Thanks @parafin for the review. I hope I have time soon to reply to that.
@TurboGit I understand that you do not want all 81 commits at that time. It is easier to find a regression later if the commits are small. Squashing everything into one big commit makes this commit very hard to debug later on. Also reverting a MR should be the same for git as reverting a single commit or do you have other experiences?
I propose that I will rebase this PR into some smaller chunks after all feedback has been integrated.

TurboGit:
> I propose that I will rebase this PR into some smaller chunks after all feedback has been integrated.

This would work for me if cleaning up your topic branch is something you can handle then perfect.

src/external/CMakeLists.txt
```diff
@@ -37,7 +40,14 @@ set(CMAKE_CXX_INCLUDE_WHAT_YOU_USE_SAVE "${CMAKE_CXX_INCLUDE_WHAT_YOU_USE}")
...
 unset(CMAKE_CXX_INCLUDE_WHAT_YOU_USE)
 
-add_subdirectory(rawspeed)
+add_subdirectory(rawspeed ${DARKTABLE_LIBDIR}/rawspeed)
+# Copying rawspeed resources to where dt expects them
+add_custom_target("deploy" ALL)
+SET(RAWSPEED_DATADIR "${CMAKE_CURRENT_SOURCE_DIR}/rawspeed/data")
+add_custom_command(TARGET "deploy" PRE_BUILD
+  COMMAND mkdir -p ${DARKTABLE_DATADIR}/rawspeed/
+  COMMAND cp ${RAWSPEED_DATADIR}/cameras.xml ${RAWSPEED_DATADIR}/showcameras.xsl ${DARKTABLE_DATADIR}/rawspeed/
```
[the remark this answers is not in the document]
codingdave: good point!

src/generate-cache/main.c
```diff
@@ -30,9 +30,10 @@
 
 #include "common/darktable.h"    // for darktable, darktable_t, dt_cleanup, etc
 #include "common/database.h"     // for dt_database_get
-#include "common/debug.h"        // for DT_DEBUG_SQLITE3_PREPARE_V2
-#include "common/mipmap_cache.h" // for dt_mipmap_size_t, etc
+#include "common/debug.h"        // for DT_DEBUG_SQLITE3_PREPARE_V
```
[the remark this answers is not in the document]
codingdave: eagle eyes.

src/common/file_location.c
codingdave: I agree that this is not very pretty but compromises have to be made. I hope the implemented compromises are the best choice, though.

Whether my approach to set up everything relative to the bin directory (in which the executable darktable lives) can be debated. Changing the relative resolution technique to another one in a second step should be pretty easy because the big work in this PR is to enable relative resolution at all. We can happily discuss the ups and downs later on.

I don't understand why it won't work when CMAKE_BINARY_DIR is /games/bin (did you miss the leading slash?). It was explicitly designed to work with this location, that is why I came to the conclusion that I have to specify everything relative to the binary dir.

What I can see as a next version if necessary is #3. I detect the base dir of all darktable directories by the executable and then carry around ${BINDIR}/.. instead of saying ${BASEDIR}=${BINDIR}/.. and then using BASEDIR instead. However I really have to dig into that again to understand the complexity and how hard it is to change since I left these kind of thoughts end of May, when I thought it is finished.

codingdave (commented):
all requested changes have been made - afaik

CMakeLists.txt
```diff
@@ -45,6 +45,25 @@ include(GNUInstallDirs)
...
 # https://medium.com/@alasher/colored-c-compiler-output-with-ninja-clang-gcc-10bfe7f2b949
+
+include(CMakePrintHelpers)
+
+# we transform the variables to CMAKE_PATH(slash based)
+file(TO_CMAKE_PATH "${CMAKE_INSTALL_FULL_BINDIR}" CMAKE_INSTALL_BINDIR_CMAKE_PATH)
+file(TO_CMAKE_PATH "${CMAKE_INSTALL_FULL_LIBDIR}" CMAKE_INSTALL_LIBDIR_CMAKE_PATH)
+file(TO_CMAKE_PATH "${CMAKE_INSTALL_FULL_DATADIR}" CMAKE_INSTALL_DATADIR_CMAKE_PATH)
+file(TO_CMAKE_PATH "${CMAKE_INSTALL_FULL_LOCALEDIR}" CMAKE_INSTALL_LOCALEDIR_CMAKE_PATH)
+get_filename_component(CMAKE_INSTALL_BINDIR_ABS "${CMAKE_INSTALL_FULL_BINDIR}" REALPATH BASE_DIR "${CMAKE_INSTALL_PREFIX}")
```
parafin, line 56: Why this is needed? CMAKE_INSTALL_FULL_BINDIR should already be absolute. Do you want to resolve symlinks here? Is this because whereami library returns path with resolved symlinks? Also doesn't it make CMAKE_INSTALL_BINDIR_CMAKE_PATH variable unneeded?

src/common/file_location.c
parafin: Let's say BINDIR is /usr/games/bin and LIBDIR is /usr/lib/. Then DARKTABLE_LIBDIR is ../../lib/darktable. With macOS bundle you will get in this line path like /Applications/darktable.app/Contents/MacOS/../Resources/../lib/darktable or /Applications/darktable.app/Contents/lib/darktable which is incorrect. But it doesn't matter much, since such setup is not supported by bundle creation script anyway.
So I agree that it makes sense to merge this PR as is, and maybe change approach as described in these comments later in a separate PR.

parafin (changes requested):
I couple more minor comments, but otherwise this PR works for me.

src/libs/CMakeLists.txt
```diff
@@ -1,12 +1,12 @@
...
+include_directories("${DARKTABLE_BINDIR}/")
 
 include(manage-symbol-visibility)
 
 add_definitions(-include common/module_api.h)
 add_definitions(-include libs/lib_api.h)
 
 add_custom_command(
-  DEPENDS ${CMAKE_SOURCE_DIR}/AUTHORS ${CMAKE_SOURCE_DIR}/tools/authors_h.sh
+  DEPENDS ${CMAKE_SOURCE_DIR}/tools/authors_h.sh ${CMAKE_SOURCE_DIR}/AUTHORS
```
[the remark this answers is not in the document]
codingdave: Right. Consistency fix if you read lines below. Extracted into individual PR https://github.com/darktable-org/darktable/pull/6510.

CMakeLists.txt
codingdave: I don't remember that. The discussion in this PR from that time (bf51e038a0d501f0e1c88a6448e31a44cf75b27a) and my comment left in the code both are not very helpful:
`# we transform the variables to CMAKE_PATH(slash based)`

It might be important to transform on windows based systems, but this is just a guess? I would need to build dt on windows again to recall that. It also might be because of whereami. God, better comment of my own would be very helpful now.

CMAKE_INSTALL_BINDIR_CMAKE_PATH is not used, correct. Removed.

codingdave: CMAKE_INSTALL_<somename>_CMAKE_PATH variables removed. On linux no functionality regression noticeable. Might stem from the learning and debugging CMake and path resolution and trying out different solutions - time.

codingdave: Also updated the description to contain a list of possible startup scenarios. 
@parafin does that cover also all OsX scenarios?

CMakeLists.txt
```diff
@@ -44,7 +44,65 @@ include(DefineOptions.cmake)
 include(GNUInstallDirs)
 include(FeatureSummary)
 
-# https://medium.com/@alasher/colored-c-compiler-output-with-ninja-clang-gcc-10bfe7f2b949
```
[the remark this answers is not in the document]
parafin: And maybe without an empty line between it and `if(${FORCE_COLORED_OUTPUT})`:))

This comment corresponds to that if block about colors, and stylistically there shouldn't be an empty line here.

CMakeLists.txt
parafin: Regarding possible options to start dt on macOS - I think you covered them all (though some options are the same).

Now about these lines:

1. Converting using `TO_CMAKE_PATH` matters only on Windows, if I understand what it does correctly. Have you tested the current version of this PR (without these conversions) on Windows?
2. What is the purpose of the remaining `get_filename_component`? `CMAKE_INSTALL_FULL_BINDIR` is already absolute, isn't it? Is so, specifying `BASE_DIR` here makes no sense. The remaining effect of this line is to resolve symbolic links into a real path. That's a tricky scenario when bindir contains symbolic links, hence my question whether you know what whereami lib returns in this case? Maybe for safety we should run returned path through g_realpath ourselves, then we know for sure and this `get_filename_component` indeed makes sense (without `BASE_DIR` though probably). Everything will of course break if symbolic link is introduced in the bindir path after installation, but I suppose that's not a use-case we should care to support here.

parafin (commented):
Clarification needed on bindir handling.

CMakeLists.txt
codingdave: 1. I am going to test this on windows again, soon. The last tests are 6 Months old.
2. I don't know what whereami returns if we are starting dt with a symbolic link. So I tried:
`ln -s ~/unstable/darktable_fix/cmake-run-dt-from-build-dir/bin/ ~/symdir
~/symdir/darktable -d dev
0.000106 application_directory: /home/david/unstable/darktable_fix/cmake-run-dt-from-build-dir/bin
0.000147 darktable.datadir: /home/david/unstable/darktable_fix/cmake-run-dt-from-build-dir/share/darktable
0.000170 darktable.plugindir: /home/david/unstable/darktable_fix/cmake-run-dt-from-build-dir/lib/darktable
0.000190 darktable.localedir: /home/david/unstable/darktable_fix/cmake-run-dt-from-build-dir/share/locale
0.000216 darktable.configdir: /home/david/.config/darktable
0.000232 darktable.cachedir: /home/david/.cache/darktable
0.000250 darktable.sharedir: /home/david/unstable/darktable_fix/cmake-run-dt-from-build-dir/share/darktable
0.000263 darktable.tmpdir: /tmp
0.000278 new_xdg_data_dirs: /home/david/unstable/darktable_fix/cmake-run-dt-from-build-dir/share/darktable:/home/david/.local/share/flatpak/exports/share:/var/lib/flatpak/exports/share:/usr/local/share:/usr/share:/var/lib/snapd/desktop
0.232634 kernel directory: /home/david/unstable/darktable_fix/cmake-run-dt-from-build-dir/share/darktable/kernels`
This also works, when removing the `BASE_DIR` as suggested.
Yes, we should not try to support use cases that don't cover my intention: Let the built binaries be starable without installing.

CMakeLists.txt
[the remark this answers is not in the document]
codingdave: sorry for making so much trouble for nothing :-)

CMakeLists.txt
parafin: OK, so it seems that whereami resolves symlinks too. Then everything is fine.

parafin:
Github shows that some changes are still requested, but I don't see any unresolved comments. I'm not sure how it works or why it happened.

Anyway, I think the only thing left is confirmation that current version of PR works on Windows too.

codingdave:
right @parafin, I also don't understand that. Im installing the windows system right now to verify.

codingdave:
@parafin took some time but finally got it compiled and it runs relatively, as planned. Can you confirm that it works on OSX? I did not test that again. If so I think the PR is complete.

$ ~/darktable/build/bin/darktable.exe -d dev
0.000000 application_directory: C:\msys64\home\david\darktable\build\bin
0.000000 darktable.datadir: C:\msys64\home\david\darktable\build\share\darktable
0.000000 darktable.plugindir: C:\msys64\home\david\darktable\build\lib\darktable
0.000000 darktable.localedir: C:\msys64\home\david\darktable\build\share\locale
0.013990 darktable.configdir: C:\Users\david\AppData\Local\darktable
0.016988 darktable.cachedir: C:\Users\david\AppData\Local\Microsoft\Windows\INetCache\darktable
0.016988 darktable.sharedir: C:\msys64\home\david\darktable\build\share\darktable
0.016988 darktable.tmpdir: C:\msys64\tmp
0.016988 new_xdg_data_dirs: (null)
0.708892 kernel directory: C:\msys64\home\david\darktable\build\share\darktable\kernels
[l10n] error: can't open iso-codes file `C:\msys64\home\david\darktable\build\share\darktable\..\iso-codes\json\iso_639-2.json'
       there won't be nicely translated language names in the preferences.
[dt_pthread_create] info: bumping pthread's stacksize from 0 to 2097152
...
[iop_lens]: could not load lensfun database in `C:\msys64\home\david\darktable\build\share\lensfun\version_2'!
(darktable.exe:26116): GLib-CRITICAL **: 22:09:01.324: g_str_has_prefix: assertion 'str != NULL' failed
...

@TurboGit this PR took so much time and effort I will not rebase this into smaller chunks. So feel free to merge it as you like. Thanks in advance.

TurboGit:
So this is ready for merging! Great work ! I'll test.

TurboGit:
I cannot build:
```

<ROOT_DIR>/darktable/src/external/whereami/src/whereami.c

  Tried extensions .c .C .c++ .cc .cpp .cxx .cu .m .M .mm .h .hh .h++ .hm
  .hpp .hxx .in .txx


CMake Error at src/external/CMakeLists.txt:5 (add_library):
  No SOURCES given to target: whereami
```

And indeed into the directory whereami I have no file:

```
$ ls -l ./src/external/whereami/
total 0
```
Are you missing a file in a commit ?

TurboGit:
A note I have of course deleted build directory:

`$ rm -fr build ; ./build.sh`

codingdave:
Please call
Git submodule init
Git submodule update
Then you should have whereami

I'm not sure the pr works yet for msys.
Good night.

Pascal Obry <notifications@github.com> schrieb am Mi., 14. Okt. 2020, 22:59:

TurboGit:
A new submodule indeed :) The build went fine. What I'll do is squash everything together so we will have a single commit that could be reverted locally to build with MSYS in case of troubles. I'll merge today to ensure more field testing as I know many people are building on Windows. Thanks for the hard work!

codingdave:
Thank you @TurboGit. I have trouble building on windows and such testing the PR. Also I did not change how l10n is resolved on windows. This hardcoded approach should not get broken by my changeset. I agree, more field testing would be helpful here. Also updating the windows build instructions. I can file some suggestions as PR but I might not be the best person to decide what is good and what is bad in the Windows MSYS2 environment with the interactions to MINGW64, MINGW32, CYGWIN, and what else.

Issue #5042, opened 2020-05-14
Changing the image rating in filtered darkroom


esq4:
Hi

**Describe the bug**
In the darkroom with filter, when I change rating of an image, it disappears from filmstrip, but remains editable, and the working focus does not switch to next image on the right in filmstrip.

**To Reproduce**
1. In darkroom to select a filter other than "all"
2. Change current image rating

**Expected behavior**
Focus in filmstrip must switch to next image on the right and it should be editable.

**Platform (please complete the following information):**
Darktable Version: master
OS: Debian 10

**Additional context**
It may be unrelated to this problem, but current image is not highlighted in filmstrip.

Nilvus:
The filter is applied in filmstrip (that's the normal behavior) so depending on setting, the image could disappear but of course remains on darkroom if you don't ask darktable to change (shortcuts and click on images are here for that).

And which master you use (which commit)? The image is highlighted on my side (it should be since my CSS complete rework merged yesterday.

esq4:
I use e26828239 ("prevent dynamic shortcuts giving keyboard focus to sliders", 2020-05-13).

Weird... I checked again: there really is a highlight in filmstrip. We will consider that this problem was successfully solved.

> ... the image could disappear but of course remains on darkroom if you don't ask darktable to change (shortcuts and click on images are here for that).

But in March or earlier it worked automatically. And it was convenient and unambiguous. And now I must not forget to do extra actions to edit the selected image but not a rejected image already.

AlicVB:
I'm the one to blame for that.
That said, I'm not sure that changing the current behavior is a good idea. darkroom is for editing images, so I would find it weird if the image can disappear during an edit session.

If I want to do some culling, there's specific other mode for that...

esq4:
Nothing strange. During editing, you came to the conclusion that the image is unpromising, you reject it and do not want to edit or see it anymore. But continuing to edit the rejected image is strange.

rfeg:
(newbie here, please be tolerant)
Coming from years of lightroom (amateur) usage this one is one of the comments I have: IMO we should be able to have star rating AND reject flag, i.e. if I remove reject I get the rating (as it does in LR). Now when I reject I loose rating (ver 3.0.2 @ OSX 10.15.4)
Rational:
1. you are rating photos and you have 3 good candidates pics and you rate them 4 stars
2. you filter everything below 4 stars to focus
3. you think one has better composition than the others and reject the other two
[...]

PR #5087, opened 2020-05-16
Allow user to choose to keep delete button on last object in duplicate module

### LINKED ISSUE #4950: Add delete to duplicate manager for unduplicated files
It would be nice to be able to delete from the darkroom without having yet another keyboard short-cut I can't remember, or more buttons.

At present, if I create a duplicate image, each has a little "X" to delete inside the duplicate manager. It disappears once only one version is left.

Why not keep it? It would be a very lightweight way to add the delete function.

GrahamByrnes:
A very small PR: I'd like to be able to delete directly from dr, but the delete button disappears from the duplicate module once only a single image remains.
Since some might be nervous about this causing unintended deletions, it's enabled by a choice in the "security" section of prefs. The default is FALSE, ie current behaviour.

Would close #4950.

### DIFF
--- a/data/darktableconfig.xml.in
+++ b/data/darktableconfig.xml.in
@@ -355,11 +355,12 @@
     <default>false</default>
     <shortdescription>ask before removing empty folders</shortdescription>
     <longdescription>always ask the user before removing any empty folder. this can happen after moving or deleting images.</longdescription>
+  </dtconfig>
   <dtconfig prefs="security">
     <name>allow_delete_last_duplicate</name>
     <type>bool</type>
     <default>false</default>
-    <shortdescription>allow all files to be deleted from duplicate module</shortdescription>
+    <shortdescription>allow all files to be deleted via duplicate module</shortdescription>
     <longdescription>leaves delete button on last file in duplicate module, removes if false</longdescription>
   </dtconfig>
   <dtconfig prefs="lighttable">
--- a/data/darktableconfig.xml.in
+++ b/data/darktableconfig.xml.in
@@ -355,6 +355,12 @@
     <default>false</default>
     <shortdescription>ask before removing empty folders</shortdescription>
     <longdescription>always ask the user before removing any empty folder. this can happen after moving or deleting images.</longdescription>
+  <dtconfig prefs="security">
+    <name>allow_delete_last_duplicate</name>
+    <type>bool</type>
+    <default>false</default>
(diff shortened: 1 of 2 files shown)


TurboGit:
But when you delete the last duplicate what is remaining on the darkroom ? You have to switch to another picture either from filmstrip, but what if there is no more picture on the collection ? I kind of remember some corner cases not really easy to support that's why we have put this limitation to stay on the safe side.

GrahamByrnes:
I just reminded my self:

- If there are other images on the filmstrip, the deleted image remains at the centre until you shift, then it is gone.
- If it is the only image, you are obliged to return to the lt, and then you find yourself with the same empty-folder display as if you had used "selected images" to remove it.

So those cases don't bother me more than they would in another view. 
It's really just an option for doing a cull within dr
"I tried that, I tried the other, no it's still crap, let's get rid of it and move on".

I could live without it. I could also live with the option hidden in the undisplayed darktablerc.
See what the crowd says?

TurboGit:
Yes, but that being said the darkrroom handle edits and so clone. Removing the last is really removing the master RAW (or Jpeg) image and I really think this belongs to the lighttable and will avoid the mentioned issues. Frankly I'd prefer no doing that and closing this PR.

Nilvus:
> I just reminded my self:
> 
>     * If there are other images on the filmstrip, the deleted image remains at the centre until you shift, then it is gone.
> 
>     * If it is the only image, you are obliged to return to the lt, and then you find yourself with the same empty-folder display as if you had used "selected images" to remove it.
> 
> 
> So those cases don't bother me more than they would in another view.
> It's really just an option for doing a cull within dr
> "I tried that, I tried the other, no it's still crap, let's get rid of it and move on".
> 
> I could live without it. I could also live with the option hidden in the undisplayed darktablerc.
> See what the crowd says?

Just to add to that, see what's this user says in this issue (#5042) about the fact that when an image gone away from filmstrip mode, he complains that this image still remains on darkroom. Something similar in some way. So not easy to find a good compromise for main users on remaining last duplicate (or removing an image with actions from filmstrip). So not so easy and probably this has to be consistent on both case. If an image disappear by removing it (or change ratings on filmstrip and have some filter to not display it anymore), we have to see same behavior I think on darkroom. And not so easy.

GrahamByrnes:
Ok, I'll close...

PR #5090, opened 2020-05-16
change icon display for 'always-on' and 'always-off' iop module header buttons

elstoc:
Makes the displayed shade of the 'always-on' module closer to that of the 'on' module (to match the display in the history module), though it is still made slightly darker by gtk because the button is insensitive.

Also adds an 'always-off' icon. Examples of all possible states below (ignoring the potential 'focus highlighting' in #5091), in order:

- Off (can be switched on)
- Forced Off (cannot be switched on for this image)
- Forced On (cannot be switched off for this image)
- On (can be switched off)

Screenshot_2020-05-17_12-38-38

### DIFF
--- a/src/develop/imageop.c
+++ b/src/develop/imageop.c
@@ -1172,16 +1172,11 @@ void dt_iop_gui_set_enable_button(dt_iop_module_t *module)
 {
   if(module->off)
   {
+    gtk_toggle_button_set_active(GTK_TOGGLE_BUTTON(module->off), module->enabled);
     if(module->hide_enable_button)
-    {
-      gtk_toggle_button_set_active(GTK_TOGGLE_BUTTON(module->off), FALSE);
       gtk_widget_set_sensitive(GTK_WIDGET(module->off), FALSE);
-    }
     else
-    {
-      gtk_toggle_button_set_active(GTK_TOGGLE_BUTTON(module->off), module->enabled);
       gtk_widget_set_sensitive(GTK_WIDGET(module->off), TRUE);
-    }
   }
 }
 
--- a/src/develop/imageop.c
+++ b/src/develop/imageop.c
@@ -2012,10 +2012,15 @@ GtkWidget *dt_iop_gui_get_expander(dt_iop_module_t *module)
   gtk_widget_set_name(GTK_WIDGET(hw[IOP_MODULE_PRESETS]), "module-preset-button");
 
   /* add enabled button */
-  if(module->enabled && module->default_enabled && module->hide_enable_button)
+  if(module->default_enabled && module->hide_enable_button)
   {
     hw[IOP_MODULE_SWITCH] = dtgtk_togglebutton_new(dtgtk_cairo_paint_switch_on, CPF_STYLE_FLAT | CPF_BG_TRANSPARENT | CPF_DO_NOT_USE_BORDER, module);
     gtk_widget_set_name(GTK_WIDGET(hw[IOP_MODULE_SWITCH]), "module-always-enabled-button");
+  }
+  else if(!module->default_enabled && module->hide_enable_button)
+  {
+    hw[IOP_MODULE_SWITCH] = dtgtk_togglebutton_new(dtgtk_cairo_paint_switch_off, CPF_STYLE_FLAT | CPF_BG_TRANSPARENT | CPF_DO_NOT_USE_BORDER, module);
+    gtk_widget_set_name(GTK_WIDGET(hw[IOP_MODULE_SWITCH]), "module-always-disabled-button");
   }
   else
   {
--- a/src/dtgtk/paint.c
+++ b/src/dtgtk/paint.c
@@ -313,6 +313,19 @@ void dtgtk_cairo_paint_switch_on(cairo_t *cr, gint x, gint y, gint w, gint h, gi
   cairo_identity_matrix(cr);
 }
 
+void dtgtk_cairo_paint_switch_off(cairo_t *cr, gint x, gint y, gint w, gint h, gint flags, void *data)
+{
+  const gint s = w < h ? w : h;
+  cairo_translate(cr, x + (w / 2.0) - (s / 2.0), y + (h / 2.0) - (s / 2.0));
+  cairo_scale(cr, s, s);
+  cairo_set_line_width(cr, 0.1);
+
+  cairo_arc(cr, 0.5, 0.5, 0.50, 0, 2 * M_PI);
+  cairo_stroke(cr);
+
+  cairo_identity_matrix(cr);
+}
+
 void dtgtk_cairo_paint_switch_deprecated(cairo_t *cr, gint x, gint y, gint w, gint h, gint flags, void *data)
 {
   const gint s = w < h ? w : h;
--- a/src/dtgtk/paint.h
+++ b/src/dtgtk/paint.h
@@ -67,6 +67,8 @@ void dtgtk_cairo_paint_flip(cairo_t *cr, gint x, gint y, gint w, gint h, gint fl
 void dtgtk_cairo_paint_switch(cairo_t *cr, gint x, gint y, gint w, gint h, gint flags, void *data);
 /** Paint an always-on switch icon */
 void dtgtk_cairo_paint_switch_on(cairo_t *cr, gint x, gint y, gint w, gint h, gint flags, void *data);
+/** Paint an always-off switch icon */
+void dtgtk_cairo_paint_switch_off(cairo_t *cr, gint x, gint y, gint w, gint h, gint flags, void *data);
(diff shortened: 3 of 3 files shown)


jenshannoschwalm:
imho the old icon was better. Something always on should need no further amplification.

elstoc:
First of all I should have mentioned that the code right now is inconsistent: ``dt_iop_gui_set_enable_button`` runs     ``gtk_toggle_button_set_active(GTK_TOGGLE_BUTTON(module->off), FALSE)`` but ``dt_iop_gui_get_expander`` has ``  gtk_toggle_button_set_active(GTK_TOGGLE_BUTTON(hw[IOP_MODULE_SWITCH]), module->enabled)``

Second - the icon image already tells you whether you can interact with a module but the colour should tell you whether it is currently switched on to be consistent with the rest of the UI. The module panel is currently inconsistent with the history module, which displays 'always on' and 'manually switched on' in the same shade:

Screenshot_2020-05-17_09-01-11
Screenshot_2020-05-17_09-01-38

parafin:
I would prefer always-on icon be as visible (bright) as simple on icon. I often look over the list of iops to get a sense of what's currently in the pipe (yes, I know there's "active" group).

elstoc:
For comparison, if you don't want to compile, here's what it looks like after this PR:

Screenshot_2020-05-17_10-22-01

Still not quite as bright as the on/off icon in its on state (because it's darkened by gtk when making it insensitive) but better, I think.

parafin:
indeed it's better

elstoc:
I wonder also if it might be better to have an always-off icon as well as an always-on icon, perhaps:

Screenshot_2020-05-17_12-13-34

As opposed to the current:

Screenshot_2020-05-17_12-15-02

Nilvus:
An always-off doesn't make sense for me but I never use them, so as you want. There's no really always-off module, only modules that's applies only some specific (rare) cases. So, I would let them as they are. But hope the module will switch to on and off icons on the specific cases they will be used. We also have to think of potential future modules...

It's ok for other ones, your last proposal of always on module is ok. Your first one was not ok. Always on icon should be less brighter than on icon. But of course it's a good idea to have it brighter than off one.

elstoc:
> There's no really always-off module, only modules that's applies only some specific (rare) cases

Agreed, same is mostly true of the always-on modules (except input and output profiles). Most of them are forced on or off based on whether it's a raw file or a jpeg.

Nilvus:
> > There's no really always-off module, only modules that's applies only some specific (rare) cases
> 
> Agreed, same is mostly true of the always-on modules (except input and output profiles). Most of them are forced on or off based on whether it's a raw file or a jpeg.

I disagree for always-on modules as they are more used than "always-off" one. You forgot demosaic (it's not used in jpg but darktable is at first a raw). For me, we can't compare "always-off" and "always-on". First are rarely used, second are most of the times used. Quite a big difference.

elstoc:
> You forgot demosaic (it's not used in jpg but darktable is at first a raw).

I used it in one of my examples. I still think we're confusing matters by using the terms 'always off' and 'always on'. Demosaic is 'always on' for a raw and 'always off' for a jpeg - I only use the terms because that's how the names are defined in the code.

Better to say 'forced on for this image' and 'forced off for this image'. There are no modules that are truly 'always off' and only two that are truly 'always on' and displayed in the left-hand-side of the screen as interactive iop modules (input/output profile).

So the icon examples in the OP (repeated below) are, in order 

- Off (can be switched on)
- Forced Off (cannot be switched on for this image)
- Forced On (cannot be switched off for this image)
- On (can be switched off)

Screenshot_2020-05-17_12-38-38

TurboGit:
For me the always on/off buttons should be very easy to distinguish between other standard buttons. So I prefer the current version.

elstoc:
If we're just worried about the color/shade, that can be tweaked later with CSS (I've made sure that the two buttons have different CSS names). I was just (initially) concerned with making sure gtk considered the 'enabled/disabled' modules as toggled/not-toggled respectively.

parafin:
This GUI element serves 2 purposes:

1. it's a button to enable/disable the module
2. it's an indicator of whether module is enabled or disabled

For first purpose it should be distinguishable if button is (inter-)active or not. But it shouldn't hurt the second purpose. So I don't like current design - it looks like always-on module is actually off.

Nilvus:
I agree with @parafin arguments here. But maybe the always-on icon remains a little bit too bright. So I would suggest just a compromise by reducing just a little bit the icon. By of course, let this icon brighter than actual current master.

elstoc:
I haven't changed the brightness of the icon, just forced it to 'toggled-on' state in gtk. The colour can be changed in CSS.

Nilvus:
Ok, so I will see that after this is merge and @Mark-64 work on icons.

TurboGit:
Ok, so let's go for it. Thanks all !

todd-prior:
Likely its been discussed but really to scan the list the icons are nice but I would bold the text of any module currently impacting the image as well...the button would further indicate a force or un-forced status...visually then no mistake......

Nilvus:
@todd-prior: bold text is not good seeing (tested and confirm by another developer) but I've added a more white text on focus modules recently. It's less visible on grey theme by now and will check if I can improve it without breaking anything else. Contrast is a little bit harder on grey theme.

But if you want bold text, you can set it with CSS for your personal user one, so directly on preferences. For focus module, just change settings of this:
```
.dt_module_focus #iop-panel-label /* set focus mode */
{
  color: @plugin_focus_color;
}
```

Just replace color line by `font-weight: bold;`

todd-prior:
Thanks for the tip…its fine for me….we highlight the buttons of the active modules vs inactive so why not the text or have them in a background (there is likely a way to tweak the css. I would like the active text to look as it does when you select it ie to have the filled background…to me that jumps out and then I would go bold when selected…)…I think there must be a solution to do it better…esp given the focus on accessibility in software design…not everyone can squint to decipher the buttons.. personally I think there is still a better way to make this completely obvious….the lack of contrast makes the buttons notoriously difficult to see at a glance….although I essentially know what modules fall in to what class so again its np but you have made me think about going away from the grey theme back to a darker one as I think the buttons are white there and so have more contrast…

Nilvus:
Please make efforts on writing, and aerate text. It will be easier to read it and understand it. I just need 3 times to read and be sure to understand you correctly (english is not my native language).

> Thanks for the tip…its fine for me….we highlight the buttons of the active modules vs inactive so why not the text or have them in a background (there is likely a way to tweak the css).

Because we do it. You visibly don't read correctly my answer and I said that I will check how to improve that. But that will not be bold text.

> I would like the active text to look as it does when you select it ie to have the filled background…to me that jumps out and then I would go bold when selected…

On which selected text and filled background? I don't see any filled background there. An example would help and be more precise... Bold will not be a solution. It's not better, it's weird.

> I think there must be a solution to do it better…

Even when it's better, it's always possible to do better...

> esp given the focus on accessibility in software design…not everyone can squint to decipher the buttons.. 

Which focus given on accessibility? The UI is now there since more than a year and I don't see anyone (except you now, so one year after) complain about that for a button and label just changed few days ago and on a development version.
And of course, yes, we search to make people squint to decipher the buttons. That's my specific goal: make people blind... Such an usefulness (and quite accusative one) sentence. Please be more constructive.
So helping you with my last answer is the way to receive such answers... Good to learn!

> the lack of contrast makes the buttons notoriously difficult to see at a glance….

Which buttons? There's a lot on darktable? Do you talk on those buttons? Again, more constructive and precise things will help to improve the UI.

> although I essentially know what modules fall in to what class so again its np but you have made me think about going away from the grey theme back to a darker one as I think the buttons are white there and so have more contrast… 

And now, you interpret wrongly what I said. About grey theme, it's just harder to set good contrast between all elements, not that it's not possible. It's not theme or anything else but just range of colors. A black theme is a the opposite of the color range regardless of white, grey theme is near the middle... 
And grey theme have a lot of benefits to correctly process an image and avoid some optics illusions side-effects that we could have on a 

Last thing, if you reply by mail, think that answer are less readable on Github...

Nilvus:
@todd-prior: just a little tweak on CSS. That's seems for me quite contrasted and not need to "squint to decipher":

image

Nilvus:
Another proposal (near the icon focus highlight):

todd-prior:
That looks nicer to me





Sent from Outlook Mobile

todd-prior:
Your way too serious. Sorry for the brevity I was likely on my phone.. I will supply some examples to clarify. Honestly the comments were simplyeant my opinion which is worthless...





Sent from Outlook Mobile

todd-prior:
This looks really nice to me but again it just my eyes...for me that looks really nice





Sent from Outlook Mobile

PR #5091, opened 2020-05-16
another attempt at the 'module-in-focus' button

elstoc:
Following on from further discussions in #5062, this is my attempt at a 'module-in-focus' button, which indicates focus by slightly shading the on/off button when a module is in focus.

on (focused)
off (unfocused)
Screenshot_2020-05-16_19-40-20 
on(unfocused)
off (focused)
Screenshot_2020-05-16_19-40-59

I've not changed the icons for the modules that can't be toggled on/off. I think those are fine as they are.

### DIFF
--- a/src/dtgtk/paint.c
+++ b/src/dtgtk/paint.c
@@ -271,22 +271,16 @@ void dtgtk_cairo_paint_switch(cairo_t *cr, gint x, gint y, gint w, gint h, gint 
   cairo_translate(cr, x + (w / 2.0) - (s / 2.0), y + (h / 2.0) - (s / 2.0));
   cairo_scale(cr, s, s);
   cairo_set_line_cap(cr, CAIRO_LINE_CAP_ROUND);
-  if(flags & CPF_FOCUS)
-  {
-    cairo_set_line_width(cr, 0.05);
-    cairo_arc(cr, 0.5, 0.5, 0.60, 0, M_PI);
-    cairo_stroke(cr);
-  }
   cairo_set_line_width(cr, 0.1);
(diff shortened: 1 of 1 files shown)


TurboGit:
@Nilvus : Let me know how you fell about this.

AxelG-DE:
I wasn't asked, but looking after  this, since the in-focus obviously has canged :)

As current one from #5062 overshoot a bit for me, too (on laptop invisible, even on my 27" hard to see), I like this idea.

elstoc:
As @AlicVB mentioned in #5062 there's nothing stopping us from making another change so that the CSS can be aware of the focus and enabled status of the modules as well, which would be useful even if it's not immediately used.

AlicVB:
@elstoc : yes, that would be good. If you want an example, you can look here : https://github.com/darktable-org/darktable/blob/7e9029ee4b29216933e7110a79df712a8a7e0233/src/dtgtk/thumbtable.c#L93
and I prefer this proposal for the icons :)

rawfiner:
I like this way more than previous proposals :-)

Nilvus:
Yes, much better than previous ones. @jenshannoschwalm: hope it's ok for you.

jenshannoschwalm:
Yes, absolutely fine :-)

TurboGit:
I like this too !

AlicVB:
@elstoc : sorry to annoy you with that. Do you plan to add the css class, or I can do it if you want...

elstoc:
I'm happy if you want to do it - it definitely sounds useful. I nearly started looking at it for this one but then @TurboGit merged :-)

AlicVB:
@elstoc : see PR #5105

elstoc:
Looks good. I wouldn't have known about ``dt_iop_gui_get_pluginui``

PR #5095, opened 2020-05-17
Lua - change size of side and bottom panels.

### LINKED ISSUE #4174: [lua] implement panel show/hide/resize
User may want to work with different panel configuration for the same view. Example : usual work in darkroom with only right panel, but expand left panel + bottom toolbar sometimes...
In fact it's impossible to handle all use cases with shortcuts directly inside dt

It would be great to add panels configuration with lua. In my mind I would say we "just" need : 
- panel visibility (read/write)
- panel size (read/write)

So user can make their own script to handle every configurations they want and link them to a shortcut.

Sadly I've absolutely no time atm to develop that... So it will either have to wait or if someone want to take it over :)

wpferguson:
Added darktable.gui.panel_get_size and darktable.gui.panel_set_size functions to change the size of the side and bottom panels.

Updated Lua API documentation with new functions.

Updated README.md Lua section.

Fixes #4174.


TurboGit:
Looks good to me. Thanks.

PR #5096, opened 2020-05-17
Exif Exposure info should not include the 's' in tooltip and status line

jenshannoschwalm:
As described above for consistency, also i remember there was an issue,
can't find that atm.


TurboGit:
Not sure about this. All numbers I have are 1 1/5 1/640 ... so a 's' looks just fine. Would be nice to have the actual issue.

TurboGit:
I see, it is for consistency with overlay on lighttable. So I'll merge, consistence is very important. Thanks.

PR #5097, opened 2020-05-17
Create widgets based on introspection data

dterrahe:
This is the first step from #4804: adding the new calls to create sliders and comboboxes from introspection data.

This still needs @houz to find time to finish the perl work to pick up combobox item names from introspection comments in the enum definitions, but if this is merged I can at least start working on all the iop modules and switch their gui_inits to using the new calls. Those should obviously only be merged after all that infrastructure work is completed.

None of the changes in this PR touches any active code apart from setting up the new callbacks and could be reversed out again later for any reason.

### DIFF
--- a/src/develop/imageop_gui.c
+++ b/src/develop/imageop_gui.c
@@ -1,0 +1,219 @@
+/*
+    This file is part of darktable,
+    Copyright (C) 2009-2020 darktable developers.
+
+    darktable is free software: you can redistribute it and/or modify
+    it under the terms of the GNU General Public License as published by
+    the Free Software Foundation, either version 3 of the License, or
+    (at your option) any later version.
+
+    darktable is distributed in the hope that it will be useful,
+    but WITHOUT ANY WARRANTY; without even the implied warranty of
+    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+    GNU General Public License for more details.
+
+    You should have received a copy of the GNU General Public License
+    along with darktable.  If not, see <http://www.gnu.org/licenses/>.
+*/
+
+#include "develop/imageop_gui.h"
+#include "develop/imageop.h"
+#include "bauhaus/bauhaus.h"
+#include "gui/color_picker_proxy.h"
+
+#ifdef GDK_WINDOWING_QUARTZ
+#include "osx/osx.h"
+#endif
+
+#include <assert.h>
+#include <gmodule.h>
+#include <math.h>
+#include <stdlib.h>
+#include <string.h>
+#include <strings.h>
+#if defined(__SSE__)
+#include <xmmintrin.h>
+#endif
+#include <time.h>
+
+typedef struct dt_module_param_t
+{
+  dt_iop_module_t *module;
+  int param_offset;
+} dt_module_param_t;
+
+static void generic_slider_callback(GtkWidget *slider, dt_module_param_t *data)
+{
+  if(darktable.gui->reset) return;
+
+  dt_iop_module_t *self = data->module;
+  dt_iop_params_t *p = (dt_iop_params_t *)self->params;
+  float *field = (float*)(p + data->param_offset);
+
+  float previous = *field;
+  *field = dt_bauhaus_slider_get(slider);
+
+  if (*field != previous)  // ignore unchanged (second call for button release)
+  {
+    if (self->gui_changed) self->gui_changed(self, slider, &previous);
+
+    dt_iop_color_picker_reset(self, TRUE);
+
+    dt_dev_add_history_item(darktable.develop, self, TRUE);
+  }
+}
+
+static void generic_combobox_callback(GtkWidget *combobox, dt_module_param_t *data)
+{
+  if(darktable.gui->reset) return;
+
+  dt_iop_module_t *self = data->module;
+  dt_iop_params_t *p = (dt_iop_params_t *)self->params;
+  int *field = (int*)(p + data->param_offset);
+
+  int previous = *field;
+  *field = dt_bauhaus_combobox_get(combobox);
+
+  if (*field != previous) // ignore unchanged (second call for button release)
+  {
+    if (self->gui_changed) self->gui_changed(self, combobox, &previous);
+
+    dt_iop_color_picker_reset(self, TRUE);
+
+    dt_dev_add_history_item(darktable.develop, self, TRUE);
+  }
+}
+
+GtkWidget *dt_bauhaus_slider_new_from_params_box(dt_iop_module_t *self, const char *param)
+{
+  dt_iop_params_t *p = (dt_iop_params_t *)self->params;
+  dt_introspection_field_t *f = self->so->get_f(param);
+
+  GtkWidget *slider = NULL;
+  gchar *str;
+
+  if (f && f->header.type == DT_INTROSPECTION_TYPE_FLOAT)
+  {
+    float min = f->Float.Min;
+    float max = f->Float.Max;
+    float defval = *(float*)self->so->get_p(p, param);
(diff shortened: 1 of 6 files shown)


dterrahe:
@TurboGit @houz Would you be OK with merging this now so that I can start work on separate branches/PRs for each of the modules that would have a dependency on this? Or is there a better way to organise this on GitHub, i.e. to cook multiple different tracks with dependencies at the same time?

Also, I could also work on one large PR for the modules, with each module going into a separate commit. Then those could be cherry-picked into master instead of all having to wait on the last most complex ones that would take long to review/discuss. But I think that would get more chaotic and harder to track sign-offs from the main contributors for each module.

TurboGit:
I'll review that soon.

dterrahe:
Sorry, i didn't mean to hurry you. I just thought my introductory comment could be read as this still being wip/not ready to merge and I did not mean it like that :-)

src/develop/imageop_gui.c
```diff
@@ -0,0 +1,219 @@
...
+    dt_module_param_t *module_param = (dt_module_param_t *)g_malloc(sizeof(dt_module_param_t));
+    module_param->module = self;
+    module_param->param_offset = self->so->get_p(p, param) - p;
+    g_signal_connect_data(G_OBJECT(combobox), "value-changed", G_CALLBACK(generic_combobox_callback), module_param, (GClosureNotify)g_free, 0);
+
+    if (f->header.type == DT_INTROSPECTION_TYPE_ENUM)
+    {
+     for(dt_introspection_type_enum_tuple_t *iter = f->Enum.values; iter->name; iter++)
+      {
+//        dt_bauhaus_combobox_add(combobox, gettext(iter->description));
```
TurboGit, line 199: Is that expected that no entry are added into the combobox ? If ok and code for later please add a comment otherwise uncomment this line.

src/develop/imageop_gui.c
```diff
@@ -0,0 +1,219 @@
...
+
+    float top = fmaxf(fabsf(min),fabsf(max));
+    if (top>=100)
+    {
+      step = 1.f;
+    }
+    else
+    {
+      step = top / 100;
+      float log10step = log10f(step);
```
TurboGit, line 111: const

src/develop/imageop_gui.c
```diff
@@ -0,0 +1,219 @@
...
+      float fdigits = floorf(log10step+.1);
```
TurboGit, line 112: const

src/develop/imageop_gui.c
```diff
@@ -0,0 +1,219 @@
...
+{
+  dt_iop_params_t *p = (dt_iop_params_t *)self->params;
+  dt_introspection_field_t *f = self->so->get_f(param);
+
+  GtkWidget *slider = NULL;
+  gchar *str;
+
+  if (f && f->header.type == DT_INTROSPECTION_TYPE_FLOAT)
+  {
+    float min = f->Float.Min;
```
TurboGit, line 97: const

src/develop/imageop_gui.c
```diff
@@ -0,0 +1,219 @@
...
+    float max = f->Float.Max;
```
TurboGit, line 98: const

src/develop/imageop_gui.c
```diff
@@ -0,0 +1,219 @@
...
+    float defval = *(float*)self->so->get_p(p, param);
+    int digits = 2;
+    float step;
+
+    float top = fmaxf(fabsf(min),fabsf(max));
```
TurboGit, line 103: const

src/develop/imageop_gui.c
```diff
@@ -0,0 +1,219 @@
...
+  dt_introspection_field_t *f = self->so->get_f(param);
+
+  GtkWidget *slider = NULL;
+  gchar *str;
+
+  if (f && f->header.type == DT_INTROSPECTION_TYPE_FLOAT)
+  {
+    float min = f->Float.Min;
+    float max = f->Float.Max;
+    float defval = *(float*)self->so->get_p(p, param);
```
TurboGit, line 99: const

src/develop/imageop_gui.c
```diff
@@ -0,0 +1,219 @@
...
+    int digits = 2;
+    float step;
```
TurboGit, line 101: = 0;

src/develop/imageop_gui.c
```diff
@@ -0,0 +1,219 @@
...
+  return slider;
+}
+
+GtkWidget *dt_bauhaus_combobox_new_from_params_box(dt_iop_module_t *self, const char *param)
+{
+  dt_iop_params_t *p = (dt_iop_params_t *)self->params;
+  dt_introspection_field_t *f = self->so->get_f(param);
+
+  GtkWidget *combobox = NULL;
+  gchar *str;
```
TurboGit, line 171: = NULL;

src/develop/imageop_gui.c
```diff
@@ -0,0 +1,219 @@
...
+  {
+    combobox = dt_bauhaus_combobox_new(self);
+
+    if (*f->header.description)
+    {
+      dt_bauhaus_widget_set_label(combobox, NULL, gettext(f->header.description));
+    }
+    else
+    {
+      str = dt_util_str_replace(f->header.field_name, "_", " ");
```
TurboGit, line 183: The fact that the introspection filed is transformed to form the name of the combo by removing the _ should be documented somewhere.

dterrahe: My plan is to significantly update "useless.c" once I've submitted (and received feedback on) a number of the non-trivial modules. That might be a good place to document this sort of thing would you agree? Happy to improve documentation elsewhere as well.

Also, as I now commented in the code, I'd prefer this transformation to happen in the introspection code, so it would also be available to all other users of that data. I've mentioned this to @houz on IRC but I'm not sure if he agreed :-) It would save quite a lot of trivial $DESCRIPTION comments.

src/develop/imageop_gui.c
dterrahe: added brief comment in the code saying @houz is working on this

dterrahe:
I'm now wondering; the only string I've marked as translatable is only programmer facing, i.e. it shouldn't ever show up in shipping code, when all widgets have been tested at least once, and a user wouldn't be able to do anything with it anyway. Should I rmove the _() brackets from that? Should the file be added to POTFiles.in anyway, just in case of future additions?

src/develop/imageop_gui.c
TurboGit: Ok, so I better let @houz review and give the go for this.

dterrahe:
@houz @TurboGit I've started implementing the changes in the individual modules and uploaded them as a draft pull request in #5201, split in one commit per module. I'd love to get your reaction.

Obviously, I can make any changes following from @houz review here, or from his implementation of enum introspection in perl. I'm also interested in suggestions for a better name than "dt_bauhaus_slider_new_from_params_box". Insofar as they are not simple search&replaces, they are probably significantly less work than the careful review of all removed code that I've done so far

This also lead to some additional improvements in bauhaus, especially around "callback" transformations, now called curve (this was only used in perspective correction) and multiplication factor/offsets for value display, in order for the actual range of the slider to match the valid range specified in introspection.

One open issue on my side is default values for arrays and enum defaults using their identifiers rather than numbers. If these are implemented in perl then the default init() function could be expanded to cover most cases and reload_defaults would only be required for cases where the defaults are actually image dependent. Otherwise i'm thinking of in most cases having default_init() cover the scalar part of the initialisation of default_params and finishing it off with the non-scalar and non-constant bits in reload_defaults(). What do you think?

PR #5098, opened 2020-05-17
Correct focus peaking overlay, fixes #5059

### LINKED ISSUE #5059: Focus Peaking highlights in lighttable misaligned
**Describe the bug**
In lighttable view, turning on Focus Peaking (Ctrl-Shift F) displays focus regions of images but they are not aligned with the displayed image. It looks as though the highlighted focus points come from the un-edited image, not the displayed version of the image. So cropping, rotation, and lens correction seem to displace the focus points from their correct position.

Note that focus highlights in the darkroom view seem to be ok.

**To Reproduce**
1. Open a folder of edited photos in lighttable
2. Turn on focus peaking (Ctrl-Shift F)
3. Observe the alignment of the focus highlights with respect to the image features.

**Expected behavior**
I would expect that the focus highlights would be at least approximately aligned with the underlying image. They are not.

**Screenshots**
Focus-Peaking-Error

 - Darktable Version: 3.1.0~git1604.1f74b4108
 - OS: Kubuntu 19.10
 - OpenCL activated: yes
 - Nvidia GTX 960M

jenshannoschwalm:
The focus peaking algo works best if we pass the whole dataset to be processed.

While scaling this is slower but gives a) stable overlay results and b) a correct
positioning.

@aurelienpierre correct?

Probably a fix #5059

### DIFF
--- a/src/views/view.c
+++ b/src/views/view.c
@@ -935,7 +935,7 @@ int dt_view_image_get_surface(int imgid, int width, int height, cairo_surface_t 
   const int img_height = buf_ht * scale;
   *surface = cairo_image_surface_create(CAIRO_FORMAT_RGB24, img_width, img_height);
 
-  // we transfer cahed image on a cairo_surface (with colorspace transform if needed)
+  // we transfer cached image on a cairo_surface (with colorspace transform if needed)
   cairo_surface_t *tmp_surface = NULL;
   uint8_t *rgbbuf = (uint8_t *)calloc(buf_wd * buf_ht * 4, sizeof(uint8_t));
   if(rgbbuf)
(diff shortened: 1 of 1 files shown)


aurelienpierre:
> @aurelienpierre correct?

Maybe ? It actually depends on what details scale you want to highlight, since it's only a scope. That extra accuracy can also result in more noise sensitivity, and users might want sharpness as 1:1 display pixel, so the sharpness highlighting follows display zoom-in/zoom-out closely.

TurboGit:
This looks good and is still quite fast (tested on lighttable with many thumbs and on darkrroom).

@aurelienpierre : have tested this ?

PR #5101, opened 2020-05-17
filter fix : make sort combobox work after intermediate sort table insert

phweyland:
- filter fix : make sort combobox work after intermediate sort table insert (#5069)
- selection fix - seg fault


TurboGit:
Indeed ! Looks good, thanks for fixing that.

AxelG-DE:
is it possible, this one causes a regression?
The sort-combobox is for me only visible in the 1* selection but not in 2* or higher!

PR #5103, opened 2020-05-17
again some culling fix

### LINKED ISSUE #5075: Rejected images in culling mode not updated on lighttable behind
**Describe the bug**
When rejecting an image in culling mode (with all except rejected mode set) and go back to lighttable filemanager view, rejected images are still.

Go to rejected filter mode and see new rejected images there. Go back to all except rejected mode, rejected images are no more there (so update is made).

This probably means a some refresh missing between culling mode and lighttable.

**Expected behavior**
Having rejected images correctly updated on lighttable regardless of filters sets when rejected them or set anything else on culling mode.

Tested on commit 6e4adf93f83e8f420557c266ede54ca7a0c19484.

@AlicVB: another one for you!

AlicVB:
- After an intensive session of debugging with @Nilvus we finally found the cause of his weird issue. So this fix #5075. finding that was hard, but the fix is simple...
- This fix also an issue when you reject images in culling with some selection set and no filmstrip shown.
- Last, some cleanup of culling unused code

### DIFF
--- a/src/dtgtk/culling.c
+++ b/src/dtgtk/culling.c
@@ -18,15 +18,10 @@
 /** a class to manage a collection of zoomable thumbnails for culling or full preview.  */
 #include "dtgtk/culling.h"
 #include "common/collection.h"
-#include "common/colorlabels.h"
 #include "common/debug.h"
-#include "common/history.h"
-#include "common/ratings.h"
 #include "common/selection.h"
 #include "control/control.h"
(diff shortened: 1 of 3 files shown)


TurboGit:
Works for me. Thanks.

Nilvus:
Well done @AlicVB. Just tested on master with that fix and all is ok now. Reject/apply stars on culling mode is perfect now and the weird issue solved.

PR #5105, opened 2020-05-17
focus iop : add css class 'dt_module_focus' for later use

AlicVB:
for example you can now do something like : 
```
.dt_module_focus #iop-panel_label
{
  font-weight: bold;
}
```
(yes, it doesn't look so good, but it's just an example)

The class is applied to the top widget of the iop, which include the header and the iop content.

### DIFF
--- a/src/develop/imageop.c
+++ b/src/develop/imageop.c
@@ -1069,7 +1069,7 @@ static gboolean dt_iop_gui_off_button_press(GtkWidget *w, GdkEventButton *e, gpo
   }
   return FALSE;
 }
-  
+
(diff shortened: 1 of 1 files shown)


Nilvus:
Will test what's possible with that and see if something else could be good to add (or not). Please wait before merging, I will see that tonight. Thanks for that @AlicVB

Nilvus:
@AlicVB: it's ok for me and what I see to do is just make title a little brighter with CSS on focus, as the toggle icon focus state is. So ok for me.

TurboGit:
So merging ! Thanks.

PR #5106, opened 2020-05-17
channel mixer: Fix incorrect clipping of only red channel in HSL mix

### LINKED ISSUE #2986: Channel Mixer "red" slider values below 0 in hue/saturation/lightness don't affect image

mtvoid:
In the HSL mix calculations, missing enclosing parentheses were leading to clipping being applied only on the red channel component, instead of the sum of all the RGB components. Fixes #2986

### DIFF
--- a/data/kernels/extended.cl
+++ b/data/kernels/extended.cl
@@ -156,9 +156,9 @@ channelmixer (read_only image2d_t in, write_only image2d_t out, const int width,
 
   float4 pixel = read_imagef(in, sampleri, (int2)(x, y));
 
-  float hmix = clamp(pixel.x * red[CHANNEL_HUE], 0.0f, 1.0f) + pixel.y * green[CHANNEL_HUE] + pixel.z * blue[CHANNEL_HUE];
-  float smix = clamp(pixel.x * red[CHANNEL_SATURATION], 0.0f, 1.0f) + pixel.y * green[CHANNEL_SATURATION] + pixel.z * blue[CHANNEL_SATURATION];
-  float lmix = clamp(pixel.x * red[CHANNEL_LIGHTNESS], 0.0f, 1.0f) + pixel.y * green[CHANNEL_LIGHTNESS] + pixel.z * blue[CHANNEL_LIGHTNESS];
+  float hmix = clamp(pixel.x * red[CHANNEL_HUE] + pixel.y * green[CHANNEL_HUE] + pixel.z * blue[CHANNEL_HUE], 0.0f, 1.0f);
+  float smix = clamp(pixel.x * red[CHANNEL_SATURATION] + pixel.y * green[CHANNEL_SATURATION] + pixel.z * blue[CHANNEL_SATURATION], 0.0f, 1.0f);
+  float lmix = clamp(pixel.x * red[CHANNEL_LIGHTNESS] + pixel.y * green[CHANNEL_LIGHTNESS] + pixel.z * blue[CHANNEL_LIGHTNESS], 0.0f, 1.0f);
 
   if( hmix != 0.0f || smix != 0.0f || lmix != 0.0f )
   {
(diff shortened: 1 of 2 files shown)


TurboGit:
Really strange, maybe a wanted behavior I think it has already been discussed with @aurelienpierre.

TurboGit:
And this will certainly breaks current edit, so if we want to "fix" that, we will need to handle the previous "wrong" computation.

mtvoid:
Of course, this whole thing becomes redundant if it is replaced by the new channel mixer module as proposed in #4940, in which case, that would be the ultimate "fix". But it is buggy as it is in the present form, as it makes no sense for the red channel to be treated differently and also potentially generating negative HSL output values.

johnny-bit:
The #4940 is whole another beast that's way better than 2 or 3 other tools built into one ;)

however I still would suggest being able to use "old" channel mixer and fixing it (after all, filmic rgb essentially kills the need to have base curve and zone system, yet both are present and usable ;))

So - if you can somehow keep old edit on old version with incorrect HLS and once they get to new version give them correct calculation then it'll be great!

aurelienpierre:
Well, nice one ! That's a nasty bug indeed. And the same happens in the OpenCL code.

I think this is too major to invoke compatibility. It's not the intended behaviour, and degrades the feature in a way that can't be alleviated. I'm all for a fix.

aurelienpierre:
Just to give additional context, the RGB -> HSL needs RGB values normalized between [0 ; 1]. The coordinates transform produces negative saturation if at least one channel is greater than 1. So there is no way the actual code could work in general.

TurboGit:
@mtvoid : for the upward compatibility we need a new integer into the param for the actual version of the code to use. V1 is current, V2 will be the one with fixed clip. In V1 we still want to use the old buggy code (sad, but otherwise edits having use this module and having had so non-clipped values will have a different rendering of the image).

TurboGit:
@mtvoid : Let me know if all this is clear to you. You will find such migration in many other commits.

mtvoid:
@TurboGit I'll have a look into implementing backward compatibility, if that is necessary. This is the first time I'm actually looking more closely into the darktable code, so I hope you'll be patient!

TurboGit:
No problem, a not too complex example can be seen here 100e82db950424575c8d706f9546f9b88d8f9fd5.

TurboGit:
@mtvoid  : Any news on this one ?

mtvoid:
> @mtvoid : Any news on this one ?

#4687 supersedes this PR, and with its merging into master, I'm closing this pull request.

PR #5111, opened 2020-05-18
fast_mexp2() fix

ralfbrown:
In looking over the nonlocal-means denoising code (I'll be trying to speed it up), I noticed a broken implementation of code to do fast floating-point exponentiation by directly manipulating the bit representation of the fp number.  It converts the integer values back into a floating-point value to do the manipulation, resulting in a loss of eight bits of precision....

As of this writing, I haven't changed over any code to use the fixed version, but did add an integration test to check how much difference a corrected implementation will make.  (A note on integration-test documentation: it would be good to note which Python packages one needs, as I had to go searching to figure out that "cv2" is installed with "pip3 install opencv-python".  It's still throwing an error about something in numpy instead of computing visual differences.)


aurelienpierre:
Could you check how the standard `powf` or `exp2f` functions behave ? I have found examples in the past where they are actually faster than the integer optimisation trick. CPU abilities have changed quite a lot since these tricks were invented.

rawfiner:
Hi!
I would be happy to talk with you about nonlocal means speed-up (either by email or IRC or on an issue or whatever) :-)
(I have already tried some things to speed up the CPU implementation of nlmeans in denoise profiled, but did not get any significant speedup yet)

I will test the fix of this PR on some pictures with various settings to see if I see a visible difference (which would mean we would need to add a new version of the params for backward compatibility)

Thanks for looking into this :-)

ralfbrown:
@aurelienpierre : I just did a quick check, and changing fast_mexp2f(x) to { return exp2f(-x); } doesn't change the timings (within the normal jitter), but I don't know just how much of the total time is spent there.  So it would make sense to ditch the integer trickery for recent x86 (might still be worthwhile for ARM or other architectures).

@rawfiner : What prompted me to look at nlmeans is that it scales poorly -- past 10-12 threads, wall time actually starts increasing again (though not as dramatically as it did  for local laplacians before my rewrite -- 64 threads takes about 20% longer than 12).  I think with a pair of maximally 37x37 arrays instead of the current single 1xwidth buffer,  it will be possible to apply the same inversion trick I did on l.l.,  so that the OpenMP parallelization can happen on the outermost loop.

rawfiner:
To check timings of the exp, have you used 1 thread only? The timing overheads due to thread communication may be hiding the difference of timing of the exp.

I agree that it scales poorly. I tried to restructure the image to lower the cache miss ratio (there seems to be quite a lot of LLC misses when the number of threads is high) for the slide updates, but it did not change anything. I guess doing parallelization on the outer loop like you suggest is a good idea, if it is possible from a memory use point of view.

TurboGit:
Is this test added here pass ok on your side, I have:

```
Test 0013-denoise-nlmeans
      Image mire1.cr2
      CPU & GPU version differs
      Max  dE         7.1082
      Mean dE         0.0030
      Count above max 74
  FAILS: image visually changed
```

Is the current 12 tests are passing on your side?

BTW, I have the very same result with and with your patch.

ralfbrown:
@rawfiner: single-threaded 14.306s with the original fast_mexp2f(), 14.382 and 14.286s with exp2f(-x).  Still well within timing jitter.

@TurboGit: as mentioned in the original message, I'm currently unable to run deltae due to dependency issues, and this patch does not yet do anything but move code around.  I made the expected.png using git master as of last night, so I don't know why you're getting a different output.

BTW, I'm not suprised that there's a difference between CPU and GPU versions.  Since making the PR, I've taken a close look at all three code paths, and the scalar/SSE2 differ from CL in how they handle the left and right boundaries of the RoI, and that's on top of potentially accumulating rounding errors as the CPU code paths sweep across the image.  BTW, kernels/nlmeans.c has yet another copy of the same borken fast_mexp2f function....

ralfbrown:
Quick update.

@rawfiner: There is in fact a big difference between fast_mexp2f() and exp2f(-x).  I hadn't run the final system-installation step when recompiling, just ran directly out of the build directory, not realizing that the shared objects were being loaded from the system location.  So I was getting identical results because I was running identical code....   Based on what I measured in my replacement code last night, the numbers would be 14.3s with fast_mexp2f and approximately 24s with exp2f.

@TurboGit: it occurred to me why we're getting different output images - we were running different numbers of threads.  So the image was being split into different numbers of horizontal slices, and rounding errors accumulate towards the bottom of each slice.

I now have scalar code that beats the existing scalar code for speed, though it was a bit of a journey.  I'll issue a PR after I implement the SSE version.

v1 was a thing of beauty: 24x speedup with 32 threads, and actually used less memory for images more than 1400 pixels wide.  Unfortunately, it took 55s (would be 45s with fast_mexp2f) single-threaded versus 25.5s for the existing code, so it didn't actually beat the existing code until way more threads than the point where the old code tops out.  That's because it was doing 4x the computation since it couldn't do incremental updates in the vertical direction (would have required enormous amounts of memory).  But 4x the computation in 1.75x the time does show how memory-bound the iop is, since my v1 code has about a 9KB working set per thread, so it can basically run entirely in L1 cache.

My current version runs in 22.5s single-threaded (vs 25.5 for git master and 32.5s using exp2f) and scales to 2.1s at 16 threads.  I think I'm hitting memory bandwidth limits at that point, because 17 threads jumps to 2.5s and 18 to 2.8, where it stays all the way up to 32 threads.

ralfbrown:
>I think I'm hitting memory bandwidth limits at that point, 

Yup, definitely memory bandwidth.  I noticed that dt_get_num_threads() always returns the number of processors, not the number of  threads requested with -t, so some of the instability I was seeing in timing between different numbers of threads was simply down to the image being split into 64 slices that got assigned to N threads.  I did a bit of tweaking to avoid such inefficiences and had a nice surprise....

By using the slice size closest to 20 rows which gives a number of chunks that is an exact multiple of the number of requested threads, I get best performance at 28 threads of 0.996 seconds!  Despite doing more  computation (23.4s single threaded instead of 22.5).  Giving up two thirds of the single-threaded speed gain versus git master with 12-row slices improves best performance to 0.916 second with 32 threads and single-threaded 24.4s.  What is hapening here is that each output pixel references up to nine rows above and nine rows below, so most (with 20-row chunks) or all of the slice has shared reads with two other threads, which means it's much less likely to get a cache miss.

ralfbrown:
> @TurboGit: as mentioned in the original message, I'm currently unable to run deltae due to dependency issues, and this patch does not yet do anything but move code around. I made the expected.png using git master as of last night, so I don't know why you're getting a different output.

Well, I now have all the dependencies installed, but module "colour" has no delta_E method!  This is v0.1.5, the version I get both from my distro's repo as python3-colour and via "pip3 install --user -I colour".

The complete set of methods in "colour" 0.1.5:
```
>>> import colour
>>> dir(colour)
['COLOR_NAME_TO_RGB', 'C_HEX', 'C_HSL', 'C_RGB', 'Color', 'FLOAT_ERROR', 'HEX', 'HSL', 'HSL_equivalence', 'LONG_HEX_COLOR', 'RGB', 'RGB_TO_COLOR_NAMES', 'RGB_color_picker', 'RGB_equivalence', 'SHORT_HEX_COLOR', '__builtins__', '__cached__', '__doc__', '__file__', '__loader__', '__name__', '__package__', '__spec__', '_hue2rgb', 'color_scale', 'hash_or_str', 'hashlib', 'hex2hsl', 'hex2rgb', 'hex2web', 'hsl2hex', 'hsl2rgb', 'hsl2web', 'make_color_factory', 'print_function', 're', 'rgb2hex', 'rgb2hsl', 'rgb2web', 'sys', 'web2hex', 'web2hsl', 'web2rgb', 'with_statement']
>>> 
```

(Python 3.7.6; maybe 3.8 has a newer version of colour available?)

hatsunearu:
Old news but 
https://colour.readthedocs.io/en/latest/generated/colour.delta_E.html?highlight=delta

https://colour.readthedocs.io/en/latest/generated/colour.difference.delta_E_CIE2000.html?highlight=delta

Hope that helps? This works on Lab space though (not sure if that's how it's supposed to be)

TurboGit:
Manually merged. Test is OK. Thanks.

PR #5113, opened 2020-05-18
Store slider default independent of current soft min/max

### LINKED ISSUE #5029: Double click to reset slider is inconsistent in filmic rgb
**Describe the bug**

Filmic rgb doesn't appear to reset values correctly.

**To Reproduce**
1. Open a photo in darktable
2. Activate filmic rgb
3. double click on any slider. When I double click on the middle grey luminance slider, the value "resets" to 3.41%. When I click on the "Reset parameters" button at the top of the module, the value resets to 9.23%.
The white relative luminance resets to 3.29EV from 3.45EV
Black relative exposure resets to -10.23EV from -10.55EV
dynamic range scaling resets to -30% from 0%

The only sliders that reset correctly are shadows/highlights balance and the sliders on the display tab. This is reproducible 100% of the time on my machine with any photo.
No other modules that I had active seem to have this issue.

**Expected behavior**

I expected the double click to not change the default value of the module (ie, it should remain at the same value that the global reset button uses).

 - Darktable Version:  3.1.0+1584~gf91b84212
 - OS: archlinux

dterrahe:
Default slider value was stored as fraction of current slider range, assuming that soft min/max would not change. I explicitly broke that assumption in filmicrgb with #4846, causing #5029.

This change fixes that problem by storing the original default value and doing a full set_soft (including scaling) conversion when restoring to defaults.

Closes #5029

### DIFF
--- a/src/bauhaus/bauhaus.c
+++ b/src/bauhaus/bauhaus.c
@@ -805,8 +805,7 @@ void dt_bauhaus_slider_set_default(GtkWidget *widget, float def)
 {
   dt_bauhaus_widget_t *w = DT_BAUHAUS_WIDGET(widget);
   dt_bauhaus_slider_data_t *d = &w->data.slider;
-  float val = d->callback(widget, def, DT_BAUHAUS_SET);
-  d->defpos = (val - d->min) / (d->max - d->min);
+  d->defpos = def;
 }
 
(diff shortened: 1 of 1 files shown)


TurboGit:
Thanks !

Issue #5070, opened 2020-05-15
In preferences, "int" is "uint"?


GrahamByrnes:
During recent debugging, I attempted to include a setting in preferences with type int, and no min or max.
When I tried to use it, I discovered that the int combo boxes do not allow negative entries.

I'm not sure how recent this is: there are a couple of int values in `darktableconfig.xml.in` which have negative defaults (for selecting ICC profiles), eg `plugins/lighttable/export/icctype` but they aren't shown in the current prefs dialogue. There are very few int settings that have the min="0"  option set, whereas not many of the parameters without a min would be able to accept a negative value. Presumably it far precedes @elstoc 's titanic efforts on the prefs dialogue.

So if int was made genuinely int, there would need to be an edit of the `darktableconfig.xml.in` file. Probably fairly easy: include min="0" everywhere and then remove a couple.

Would it be worth reconfiguring to allow future use? Or too much of  corner case?

elstoc:
Seems it's actually a gint but minimum and maximum are hard-coded (to 0 and G_MAXINT). Look at ``tools/generate_prefs.xsl``.

GrahamByrnes:
Ok, thanks. Too hard then.

elstoc:
Why is that too hard? Just change ``generate_prefs`` to take the minimum and maximum from ``darktableconfig.xml.in``

GrahamByrnes:
Oh, maybe I misunderstood the spin button comment. It wasn't obvious to me looking in generate_prefs... 
I'll re-open and think some more, even if it's for a very small number of cases.

GrahamByrnes:
So then the real question: is it more painful for most developers to have to put min="0" in their new preference options, than is offered to the couple of people who want to use values <0?

elstoc:
Could you make it default to min = 0 (i.e. current functionality) and just override if the developer has explicitly added a min value to their preference option? Similarly for max.

GrahamByrnes:
Probably. Maybe next week... I need to get away from this screen!

PR #5114, opened 2020-05-18
Allow int and int64 to be <0 when given an explicit min

GrahamByrnes:
This allows `int` and `int64` values to be set <0 in preferences. Currently they are limited to >=0 by definition of the range of the spin_buttons.
To preserve current behaviour, the minimum is still set to zero unless an explicit `min="-X"` is supplied with your choice of negative bound -X.

I bumped into this while debugging, it may prove useful to someone in future. Currently there are only a couple of negative int values set in darktablerc and neither of them is visible in prefs, but it seems a very minimal change to create more flexibility.

Discussed in #5070 and thanks to @elstoc for persuading me that xml style sheets are not so hard and suggesting the opt-in strategy :-)


GrahamByrnes:
Lovely.

PR #5116, opened 2020-05-18
set dialog windows position to location near mouse

### LINKED ISSUE #5110: DB Maintenance dialog missing
**Describe the bug**
On close, darktable detects maintenance needed but does not show dialog, so it freezes (never closes, waiting for user input

**To Reproduce**
Force database maintenance needed, open darktable in debug mode and quit darktable.

**Expected behavior**
darktable minimizes but doesn't close, and debug session suggests it is waiting for user input.

**Screenshots**

 - Darktable Version: 3.1.0~git1670.92a
 - OS: ubuntu 20.04
 -  no
 -  no

**Additional context**

johnny-bit:
Fixes #5110

Previously dialog creation didn't use any hinting which might've placed dialog in unexpected place (even I had to look for dialog **somwewhere** on my 2 screens).

This change gives gtk hint to create window possibly near mouse cursor (while keeping whole window visible).

### DIFF
--- a/src/gui/gtk.c
+++ b/src/gui/gtk.c
@@ -2606,6 +2606,8 @@ gboolean dt_gui_show_standalone_yes_no_dialog(const char *title, const char *mar
   gtk_window_set_title(GTK_WINDOW(window), title);
   g_signal_connect(window, "destroy", G_CALLBACK(gtk_main_quit), NULL);
 
+  gtk_window_set_position (GTK_WINDOW(window), GTK_WIN_POS_MOUSE);
+
   GtkWidget *vbox = gtk_box_new(GTK_ORIENTATION_VERTICAL, 0);
   gtk_container_add(GTK_CONTAINER(window), vbox);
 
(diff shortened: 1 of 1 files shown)


TurboGit:
Do we have a way to force it at the center of the darktable UI ? Like the preference Window or the copy/paste history dialog ? It will be more consistent.

johnny-bit:
problem is - those dialogs are created when main window is visible. the yes/no dialogs here need to be shown before main window (in case of db upgrade), or before/after any db operations (db maintenance/backup stuff).

Obviously showing it near mouse almost guarantees that window will be noticed by user (barring some weird issue @gadolf66 has in #5110 where window doesn't happen AT ALL.

With regards to center - sure, the possibility to center is there, however main window is at the time of showing dialog either not yet created or being destroyed so best bet for center would be screen center and that can be done too.

I'd personally opt for "near mouse", what's your opinion?

TurboGit:
I see, yes I agree with this analysis. I missed that this was for the dialog before the UI is displayed. Let's go with that. Thanks.

johnny-bit:
LOL, I was just working on something even better :smile:

johnny-bit:
#5124 is way better and more complete :)

Issue #4218, opened 2020-02-02
Window minimized before quitting on macOS


Sangeppato:
Hello,
I'm on macOS 10.15.2 using Darktable 3.0.0.

My issue is quite "minor" but I still want to report it:
When quitting/closing Darktable, the window gets minimized to the dock and then it finally quits, but the expected behaviour is that there shouldn't be any minimization, it should just quit right away.

johnny-bit:
This is (I believe) fixed in 3.4

PR #5124, opened 2020-05-19
show dialog windows in predictable places - at mouse cursor or center

johnny-bit:
we want to show dialog windows in very predictable manner, so in case of multi-screen action or multi-window action, dialogs blocking input won't cause harm to users.

with this change dialogs are shown at mouse cursor if dt gui isn't yet initialised (or is already closed) or at dt window center on top

additionally based on @parafin suggestion - the main window isn't minimalised on closing. I invite anybody to come up with nice closing message though...

This improves solution from #5116 and should be generally better fir for #5110

### DIFF
--- a/src/gui/gtk.c
+++ b/src/gui/gtk.c
@@ -887,11 +887,8 @@ void dt_gui_gtk_set_source_rgba(cairo_t *cr, dt_gui_color_t color, float opacity
 
 void dt_gui_gtk_quit()
 {
-  //GtkWindow *win = GTK_WINDOW(dt_ui_main_window(darktable.gui->ui));
-
-  // Write out windows dimension before miminizing
+  // Write out windows dimension
   dt_gui_gtk_write_config();
-  //gtk_window_iconify(win);
 
   GtkWidget *widget;
   widget = darktable.gui->widgets.left_border;
@@ -2633,11 +2630,14 @@ gboolean dt_gui_show_standalone_yes_no_dialog(const char *title, const char *mar
   gtk_window_set_title(GTK_WINDOW(window), title);
   g_signal_connect(window, "destroy", G_CALLBACK(gtk_main_quit), NULL);
 
-  if(darktable.gui) {
+  if(darktable.gui)
+  {
     GtkWindow *win = GTK_WINDOW(dt_ui_main_window(darktable.gui->ui));
     gtk_window_set_transient_for(GTK_WINDOW(window), win);
     gtk_window_set_position(GTK_WINDOW(window), GTK_WIN_POS_CENTER_ON_PARENT);
-  } else {
+  }
+  else
+  {
     gtk_window_set_position(GTK_WINDOW(window), GTK_WIN_POS_MOUSE);
   }
 
@@ -2689,11 +2689,14 @@ char *dt_gui_show_standalone_string_dialog(const char *title, const char *markup
   gtk_window_set_title(GTK_WINDOW(window), title);
   g_signal_connect(window, "destroy", G_CALLBACK(gtk_main_quit), NULL);
 
-  if(darktable.gui) {
+  if(darktable.gui) 
+  {
     GtkWindow *win = GTK_WINDOW(dt_ui_main_window(darktable.gui->ui));
     gtk_window_set_transient_for(GTK_WINDOW(window), win);
     gtk_window_set_position(GTK_WINDOW(window), GTK_WIN_POS_CENTER_ON_PARENT);
-  } else {
+  }
+  else 
+  {
     gtk_window_set_position(GTK_WINDOW(window), GTK_WIN_POS_MOUSE);
   }
 
--- a/src/gui/gtk.c
+++ b/src/gui/gtk.c
@@ -890,10 +890,12 @@ void dt_gui_gtk_quit()
   GtkWidget *win = dt_ui_main_window(darktable.gui->ui);
   GtkStyleContext *context = gtk_widget_get_style_context(win);
   gtk_style_context_add_class(context, "dt_gui_quit");
+  gtk_window_set_title(GTK_WINDOW(win), _("closing darktable..."));
+
+  // don't minimize or hide main window, it'll be hiden by gtk
 
   // Write out windows dimension
   dt_gui_gtk_write_config();
-  gtk_widget_hide(GTK_WIDGET(win));
 
   GtkWidget *widget;
   widget = darktable.gui->widgets.left_border;
--- a/src/gui/gtk.c
+++ b/src/gui/gtk.c
@@ -887,6 +887,10 @@ void dt_gui_gtk_set_source_rgba(cairo_t *cr, dt_gui_color_t color, float opacity
 
 void dt_gui_gtk_quit()
 {
+  GtkWidget *win = dt_ui_main_window(darktable.gui->ui);
+  GtkStyleContext *context = gtk_widget_get_style_context(win);
+  gtk_style_context_add_class(context, "dt_gui_quit");
+
   // Write out windows dimension
   dt_gui_gtk_write_config();
 
--- a/src/gui/gtk.c
+++ b/src/gui/gtk.c
@@ -893,6 +893,7 @@ void dt_gui_gtk_quit()
 
   // Write out windows dimension
   dt_gui_gtk_write_config();
+  gtk_widget_hide(GTK_WIDGET(win));
 
   GtkWidget *widget;
   widget = darktable.gui->widgets.left_border;
@@ -2638,7 +2639,14 @@ gboolean dt_gui_show_standalone_yes_no_dialog(const char *title, const char *mar
   {
     GtkWindow *win = GTK_WINDOW(dt_ui_main_window(darktable.gui->ui));
     gtk_window_set_transient_for(GTK_WINDOW(window), win);
-    gtk_window_set_position(GTK_WINDOW(window), GTK_WIN_POS_CENTER_ON_PARENT);
+    if(gtk_widget_get_visible(GTK_WIDGET(win)))
+    {
+      gtk_window_set_position(GTK_WINDOW(window), GTK_WIN_POS_CENTER_ON_PARENT);
+    }
+    else
+    {
+      gtk_window_set_position(GTK_WINDOW(window), GTK_WIN_POS_MOUSE);
+    }
   }
   else
   {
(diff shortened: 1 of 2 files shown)


src/gui/gtk.c
```diff
@@ -2606,7 +2606,13 @@ gboolean dt_gui_show_standalone_yes_no_dialog(const char *title, const char *mar
   gtk_window_set_title(GTK_WINDOW(window), title);
   g_signal_connect(window, "destroy", G_CALLBACK(gtk_main_quit), NULL);
 
-  gtk_window_set_position (GTK_WINDOW(window), GTK_WIN_POS_MOUSE);
+  if(darktable.gui) {
```
TurboGit, line 2609: convention is:
```
if(...)
{
}
else
{
}
```

src/gui/gtk.c
```diff
@@ -2656,7 +2662,13 @@ char *dt_gui_show_standalone_string_dialog(const char *title, const char *markup
...
```
TurboGit, line 2665: likewise

src/gui/gtk.c
```diff
@@ -866,11 +866,11 @@ void dt_gui_gtk_set_source_rgba(cairo_t *cr, dt_gui_color_t color, float opacity
 
 void dt_gui_gtk_quit()
 {
-  GtkWindow *win = GTK_WINDOW(dt_ui_main_window(darktable.gui->ui));
+  //GtkWindow *win = GTK_WINDOW(dt_ui_main_window(darktable.gui->ui));
```
TurboGit, line 869: Please remove commented out code. This is anyway in the Git history.

src/gui/gtk.c
```diff
@@ -866,11 +866,11 @@ void dt_gui_gtk_set_source_rgba(cairo_t *cr, dt_gui_color_t color, float opacity
...
 
   // Write out windows dimension before miminizing
   dt_gui_gtk_write_config();
-  gtk_window_iconify(win);
+  //gtk_window_iconify(win);
```
TurboGit, line 873: likewise

TurboGit:
> I invite anybody to come up with nice closing message though...

Yes, this is really needed now. When you ask to quit (Ctrl+Q) there is no visual indication that the shutdown is going on. We may want indeed to have a message or maybe a very dull CSS version to indicate the shutdown ? @Nilvus : what you think on this?

johnny-bit:
The added class is `dt_gui_quit`... Hope this allows for nice "fix" ;)

Nilvus:
> We may want indeed to have a message or maybe a very dull CSS version to indicate the shutdown ? @Nilvus : what you think on this?

Why not but is good for me without dialog close window.
But if we add that, we just have to remain it short and simple. I've test this PR but seems the closing dialog is not created by now.

For the text, that could be something like "See you soon with new photos. Bye" for example. Or "Hungry for photos. Waiting for new ones. Bye."

TurboGit:
There is no closing dialog at the moment. My idea was to wash out the dt GUI while closing by using the new class added in this PR. Just to show that the GUI is going to be closed.

johnny-bit:
Honestly I have no idea how to best show closing state without being... hmmm...

Anyway - I thought if it's possible via css to mute colours of interface as for it to look "disabled"... And if there's a way - blur whole interface? However muted interface would be IMO quite enough...

Otherwise, having message would mean... hmmm... having hidden custom view such as `knight` and on it have closing message? If so - I could create separate PR for that but honestly "closing messages" would be very hard design-wise for me ;)

Nilvus:
@johnny-bit: the best way for me would be: nothing (as it is actually).

johnny-bit:
I've been wondering about this PR and further steps or whether to close it off completely.

as @parafin said in #5110 minimizing window on close isn't the best way to handle closing state of application, that's why this PR does away with minimizing window. Having window open on close allows for popping up dialog windows on top of main window that look VERY good. However if we don't have dialog open, having window open for period of closing and cleaning up is a bit... weird. Most aplications I use that have long closing time do either nothing (simply all input/clicking on stuff is ignored) or do the fancy. Fancy things possible:
1. since I added class to parent window on closing, muting colors is possible, but it requires writting bunch of CSS IMHO
2. I could create custom view that could be switched to on closing and have things there like closing messages etc. Interesting thing to pursue, however I wonder how long usually darktable closes to be considered "long" since messages there could be tips&trick or other fun stuff
3. ??? any ideas?

Or simply drop it and "#5116 is good enough"?

Nilvus:
> I've been wondering about this PR and further steps or whether to close it off completely.
> 
> as @parafin said in #5110 minimizing window on close isn't the best way to handle closing state of application, that's why this PR does away with minimizing window. Having window open on close allows for popping up dialog windows on top of main window that look VERY good. However if we don't have dialog open, having window open for period of closing and cleaning up is a bit... weird. Most aplications I use that have long closing time do either nothing (simply all input/clicking on stuff is ignored) or do the fancy. Fancy things possible:
> 
>     1. since I added class to parent window on closing, muting colors is possible, but it requires writting bunch of CSS IMHO

Bad idea for me. And which colors? And for what? Just few seconds...

> Or simply drop it and "#5116 is good enough"?

For me just the best way. And for 2 simple reasons:
- As yourself said: "Most aplications I use that have long closing time do either nothing (simply all input/clicking on stuff is ignored)". And I agree with that. On apps I use, I know none of them having fancy things when closing, except just that indeed input/clicking is ignored (and that's the only good way).
- It's most of the times, just few seconds to close. So really no need to spent more time on that. There's a lot of better things to do on darktable.

johnny-bit:
OK, so just to be in the clear here. 2 options, nothing fancy:
1. Don't minimize window on close, just disable/ignore clicks (this would allow to popup exit dialogs on closing dt at center of dt window making them very predictable)
2. Close this PR and let dialogs open at the cursor location. No window lingering.

Nilvus:
> OK, so just to be in the clear here. 2 options, nothing fancy:
> 
>     1. Don't minimize window on close, just disable/ignore clicks (this would allow to popup exit dialogs on closing dt at center of dt window making them very predictable)
> 
>     2. Close this PR and let dialogs open at the cursor location. No window lingering.

Not sure to understand that part but if that PR improve your first merged PR about places of dialog windows, why close it? For me, the only thing is that there's no need to add something when closing darktable (as about all apps, it's good as it is) but taking care of dialog windows places is a good thing. And if it's about maintenance dialog windows, if displayed when opening/closing, just (when needed) be sure that no click on the UI is possible this time.

johnny-bit:
Hey @TurboGit - If you don't mind, this PR can be merged as-is (or closed).

Currently it works in very nice way (imo):
1. if there's no UI window, the dialog appears at the mouse location. 
2. if there's ui window, dialog appears in the middle of window on same screen as window.
3. if darktable is closing, all clicks and events are disabled regardless whether dialog appears or not. this covers @Nilvus's worry that user could click something wrong.

I tried playing a bit with possibilities and upon quit action there's not much possible to do, I couldn't even get switching to my custom empty view happen on quit.  With added class it should possible to mute colors, I tried playing with css and either I don't know much or there's something funny going on and having rule
```css
.dt_gui_quit * {
color: @disabled_fg_color;
}
```
doesn't work... Also gtk inspector worked weird for me since I haven't seen class change on quit despite it working. maybe on quit disables changes to windows? dunno. So ultimatelly either merge or close, nothing to add here ;)

Nilvus:
@johnny-bit: with CSS you can change colors but not really "mute" them. Or to be more precise, changing color to disabled color would only set all colors to disabled default color but a user could always click on it. Anyway, I do not really worry about user try to click, it's just a possibility. As I said, I find how darktable close by now good.

hgustafsson:
The way darktable closes now on OSX (by minimizing the window to the Dock) is very non-standard for mac apps (and the minimization animation is quite obtrusive). See also issue #4218 and PR #5262

Nilvus:
On any OS, minimizing is not a good way.

johnny-bit:
Oh my... I tried and `hide` behaviour is waaay nicer than `iconify`. I tested it in most recent commit and it is VERY nice indeed.

So - one more question: let window 'linger' for a bit, so some dialogs could be shown or hide immediately? I'd rather let it linger for a bit.

parafin:
I think there should be some way to tell when darktable finished exiting (meaning all data has been committed to disk and also that darktable can be restarted at that point). Not hiding a window is IMHO the easiest way to achieve this.

hgustafsson:
@parafin - On mac this is usually visualized by the icon disappearing from the dock (or the dot beneath the app icon disappears if the app is pinned to the dock). It is standard behaviour that a mac app can still be running while not having any windows open and this is shown by keeping it on the dock.

When closing a window I would like to clear up that screen real estate as soon as possible so that I can continue working. The expected behaviour of clicking a window close button is for the window to disappear, and going against that expectation just makes it very disruptive for the user.

If you feel some visual cue is needed for when committing data to disk upon quitting, a smaller, unobtrusive window with a progress / loading indication or animation might be a good compromise.

parafin:
On Mac yeah, and I think it's the same with GNOME, but other systems might not have such indication.

johnny-bit:
OK, I think I struct here some middle ground and it's probably least intrusive:

1. window title is changed to "closing darktable..." (needs translators to translate that ;) )
2. all clicking/input events for main window are blocked
3. window hiding is postponed till most events that may ask for user input are finished (so dialogs can popup in middle of screen)
4. window is hiden since it's BEAUTIFUL and quick transition on most uis (minification is meh on gnome and really obtrusive on mac)
5. cleanup continues till darktable is finally closed.

This allows for best of all worlds - main window lingers for a bit, but title cleary indicates that it's closing and then it's closed nicelly :)

What do you all think?

hgustafsson:
Experience on OSX for commit d5083ab:

- When clicking the close window button (the red x in title bar) it stays pressed until clean up is finished (with window still open and no window title change). Then both the window and the app on the dock closes simultaneously.  
To summarize, it looks like darktable hangs for a while before the window and app closes.

- When pressing Cmd + Q (i.e. quitting the app without closing the window first), the window title does change to "closing darktable..." before both window and app closes simultaneously after clean up.

Is there some way I can help in creating a small (dialog) window showing the closing message instead while hiding the main window? (I'm new to the darktable code.)

johnny-bit:
Damn OSX and it's showstopper features ;( I'll try to cope up with yet another way...

TurboGit:
I'll reschedule this for 3.4.

johnny-bit:
There's no good way to have nice closing experience on macosx & have window up for long AND not introduce splashscreen (IMO). I opened up FR #5782 for it.

In limited scope of this PR, the current functionality is as good as I can get it to be. nothing more to add/remove IMO, so I'm open to further comments :) and/or merging it :)

PR #5130, opened 2020-05-19
darkroom : handle prev/next accels if image not in current list

### LINKED ISSUE #5042: Changing the image rating in filtered darkroom

AlicVB:
this can happen for example with filtered collection if you change rating inside darkroom

### DIFF
--- a/src/views/darkroom.c
+++ b/src/views/darkroom.c
@@ -1026,10 +1026,16 @@ static void dt_dev_jump_image(dt_develop_t *dev, int diff, gboolean by_key)
     new_offset = sqlite3_column_int(stmt, 0);
     new_id = sqlite3_column_int(stmt, 1);
   }
+  else
+  {
+    // if we are here, that means that the current is not anymore in the list
+    // in this case, let's use the current offset image
+    new_id = dt_ui_thumbtable(darktable.gui->ui)->offset_imgid;
+  }
   g_free(query);
   sqlite3_finalize(stmt);
 
-  if(new_id < 0) return;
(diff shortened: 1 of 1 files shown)


TurboGit:
This does not fix the issue on my side. When I have a filter of >= 3 stars for example and I rate in darkroom using the filmstrip the current image with one star it diseapers from the filmstrip but stays current in darkroom. This is the behavior we are seeing in current master. Am I missing something ?

AlicVB:
No, you're right, this doesn't fix exactly the issue (I just remove the tag)
In fact, as you can read in the issue #5042 I don't think it's a good thing to allow the currently edited image to disappear in darkroom because of some lighttable actions.

What it's fixed here : if the image doesn't exist in the filmstrip anymore, make the previous/next accels work by switching to the filmstrip offset (this work even if filmstrip is hidden)

What I propose is to merge this, as anyway this add just a safety guard and to continue the discussion in the mentioned issue...

TurboGit:
@AlicVB : I see, thanks for the clarification. I'm merging this then. Thanks.

PR #5132, opened 2020-05-19
Fix treeview check buttons and improve focus module title

### LINKED ISSUE #5118: [lighttable] some oddity for history stack stack, copy...
**Describe the bug**

When trying to copy history stack with the function in lighttable "copy..." and potentially set 'select none' the tickboxes are totally invisible at the moment.

@Nilvus I learned today, this _might_ be one for you :)

**Expected behavior**

either tick-boxes are visible or clicking the name also activates the relevant entry

BTW: maybe need another BR, but the "always-on", actually do not need to be shown in this dialog at all. As they are redundant, I always have to untick 'em....

**Screenshots**

former view
Screenshot_20200518_204330

current view all ticks on:
Screenshot_20200518_204552

And last but not least, finally the oddity (all ticks off)
Screenshot_20200518_204610b

Nilvus:
Fix #5118

@AxelG-DE: if you want to test.

### DIFF
--- a/data/themes/darktable-elegant-dark.css
+++ b/data/themes/darktable-elegant-dark.css
@@ -38,7 +38,7 @@
 @define-color plugin_bg_color @grey_35;
 @define-color plugin_fg_color @grey_85;
 @define-color section_label @grey_75;
-@define-color plugin_label_color @grey_75;
+@define-color plugin_label_color @grey_70;
 
 /* Modules controls (sliders and comboboxes) */
 @define-color bauhaus_fg @grey_80;
@@ -46,6 +46,7 @@
 @define-color bauhaus_indicator_border @grey_35;
 @define-color bauhaus_fill @grey_55;
 @define-color bauhaus_bg @grey_30;
+@define-color bauhaus_bg_hover @grey_65;
 @define-color bauhaus_fg_hover @grey_90;
 @define-color bauhaus_fg_selected @grey_85;
 @define-color bauhaus_fg_disabled alpha(@bauhaus_fg, 0.4);
--- a/data/themes/darktable-elegant-grey.css
+++ b/data/themes/darktable-elegant-grey.css
@@ -48,7 +48,8 @@
 @define-color bauhaus_indicator_border @grey_50;
 @define-color bauhaus_fill @grey_70;
 @define-color bauhaus_bg @grey_45;
-@define-color bauhaus_fg_hover @grey_90;
+@define-color bauhaus_bg_hover @grey_65;
+@define-color bauhaus_fg_hover @grey_95;
 @define-color bauhaus_fg_selected @grey_85;
 @define-color bauhaus_fg_disabled alpha(@bauhaus_fg, 0.5);
 @define-color bauhaus_fg_insensitive alpha(@bauhaus_fg, 0.5);
--- a/data/themes/darktable-elegant-grey.css
+++ b/data/themes/darktable-elegant-grey.css
@@ -41,6 +41,7 @@
 @define-color plugin_fg_color @grey_90;
 @define-color section_label @grey_80;
 @define-color plugin_label_color @grey_80;
+@define-color plugin_focus_color @grey_90;
 
 /* Modules controls (sliders and comboboxes) */
 @define-color bauhaus_fg @grey_90;
--- a/data/themes/darktable.css
+++ b/data/themes/darktable.css
@@ -88,11 +88,12 @@
 @define-color plugin_label_color @grey_60;
 
 /* Modules controls (sliders and comboboxes) */
+@define-color bauhaus_bg @grey_15;
 @define-color bauhaus_fg @grey_70;
 @define-color bauhaus_border shade(@plugin_bg_color, 0.5);
 @define-color bauhaus_indicator_border @grey_20;
 @define-color bauhaus_fill @grey_40;
-@define-color bauhaus_bg @grey_15;
+@define-color bauhaus_bg_hover @grey_55;
 @define-color bauhaus_fg_hover @grey_90;
 @define-color bauhaus_fg_selected @grey_80;
 @define-color bauhaus_fg_disabled alpha(@bauhaus_fg, 0.4);
@@ -294,7 +295,8 @@ entry selection
 /* Checkbox */
 checkbutton check,
 #lib-plugin-ui checkbutton check,
-#iop-plugin-ui checkbutton check
+#iop-plugin-ui checkbutton check,
+treeview check
 {
   background-color: @bauhaus_bg;
   color: @bauhaus_fg;
@@ -1128,7 +1130,6 @@ filechooser row:hover,
   - Tree view (lists and tables) -
   --------------------------------*/
 /*** Set treeview header on all windows, including preferences and accels window ***/
-
 treeview,
 filechooser widget treeview
 {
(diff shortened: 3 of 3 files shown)


Nilvus:
2nd commit improve focus module since #5105 PR add. Make simply module title brighter when module is on focus.

Nilvus:
@TurboGit: wait before merging this one. I will update something as I see an hover effect on toggle icon in modules not wanted. Will check that.

elstoc:
I was happy enough with just the icon, but this looks fine too. I did see that hover effect when I was testing the icon - I assumed it was intentional but agree it does look odd.

AxelG-DE:
I'll test by tomorrow. Now is ZZZZzzzzz-time :)

AxelG-DE:
> @AxelG-DE: if you want to test.

Tested, looks good to me, thank you!

Nilvus:
@TurboGit: after deepening on hover buttons effect, work will be longer and harder than initially thought. I will even see if I will not change/improve all hover effects (buttons and combobox mainly). So, I suggest finally to merge (if ok) this PR without change on buttons by now.
I will probably propose another PR for hover buttons and combox later.

Just a thing, for that future PR, I see that some buttons have some remaining hardcoded things on GTK. On what @elstoc done for arrow enlarged, I understand that all "CPF_" things on a Gtk widget could be removed. Hope it's ok, I see some on lib.c and imageop.c about those buttons (so that overwrite some CSS work). I will test on my side by removing them.

Nilvus:
It seems than removing the CPF_FLAT..., CPF_DO_NOT_USE_BORDER is not enough. When compile, by removing all them on imageop.c, I can't compile anymore darktable and have some message errors. If someone could help to remove correctly all hardcoded things about all buttons in modules labels (so on all views) to set them all correctly with CSS only.

elstoc:
> On what @elstoc done for arrow enlarged, I understand that all "CPF_" things on a Gtk widget could be removed.

iirc all I did was to make the initial button creation and click activity consistent. I don't know what impact removing "CPF_" settings would have elsewhere (and didn't need to know for that fix). I certainly wouldn't assume you can remove those settings anywhere else.

elstoc:
On further investigation I just wanted to confirm that when we're talking about the 'hover effect' we both mean the same - that when you hover over the button, the foreground colour goes lighter and background colour goes darker If so, that's general button behaviour in darktable and probably should be left alone (consistent).

Nilvus:
> On further investigation I just wanted to confirm that when we're talking about the 'hover effect' we both mean the same - that when you hover over the button, the foreground colour goes lighter and background colour goes darker If so, that's general button behaviour in darktable and probably should be left alone (consistent).

We of course talk about the same thing. But, that's not totally consistent. In reality toggle on/off/always-on/off button is not totally consistent. To make that button not odd on hover mode, a border-radius is set but I don't see it on this button (only). And I haven't change anything on hover button on my complete CSS rework and yesterday, I see that some lines are probably redundant. And it could probably be simplify. And that give me some idea. Anyway, I have check button.c and I begin to understand some things. A flag (CPF_...) need to remain, one at least.

elstoc:
> In reality toggle on/off/always-on/off button is not totally consistent.

Agreed - it only appears on the buttons that can be toggled because the 'always on' / 'always off' buttons are deactivated (insensitive). I'm not sure that's a bad thing.

Nilvus:
> Agreed - it only appears on the buttons that can be toggled because the 'always on' / 'always off' buttons are deactivated (insensitive). I'm not sure that's a bad thing.

That's not consistent. And border-radius is applied (setting is shown in Gtk Inspector Tool) but overwritten. Take for example group, star, help and prefs icon, you have 2 with border-radius visible and 2 without. And it's better with border-radius I think. Unfortunately, I've find quite easy by searching in files with CSS tags and find where on code buttons are drawn. I made some tests by adding/removing some CPF_FLAG, CPF_BG_TRANSPARENT. Nothing works (I can change some state but border-radius are not set for toggle buttons. So probably something to work on on togglebutton.c but that's beyond my skills.

I think that has been changed, hover background and border are set by Gtk code and only some things could be overwritten. In reality that's not border-radius that is not applied, it's border at all. It's impossible to set borders for example on all toggle buttons, we can't also not set hover background or you will have a glitch issue, by the hover hardcoded background display a really short time when hovering the button.

And by the way, that's inconsistent with the idea to set all in CSS and not have hardcoded settings that could be set on CSS. I will need help as I don't know how to change that.
@Mark-64: as you spent time recently on group/star/help and prefs icons, and 2 of them are toggle ones, maybe you know how to fix that? (or @TurboGit, @AlicVB, @aurelienpierre?).

elstoc:
> That's not consistent

I suppose it depends what you're trying to be consistent with. If the intention is to use the changing shades to indicate that something can be pressed/toggled, then all things that can be pressed are handled consistently. Always on/off modules don't have a button that can be pressed/toggled.

Nilvus:
> > That's not consistent
> 
> I suppose it depends what you're trying to be consistent with. If the intention is to use the changing shades to indicate that something can be pressed/toggled, then all things that can be pressed are handled consistently. Always on/off modules don't have a button that can be pressed/toggled, just an icon to indicate their state.

Yes, of course. Said that, that doesn't change the fact that some things remains hardcoded. And you yourself say that hover on/off/always icons on module header is odd.

Nilvus:
Seems that there another thing with hardcoded things: "more modules" treeview in darkroom. The hover applied is only color. Not consistent with all other treeviews (and they all share same lines on CSS), so I'm sure the background is here overwritten. The part applied for color has a background color not applied here. And I've check and a long forgotten one as it was already there on 3.0.0.

So will need Gtk code help for this one too.

elstoc:
> And you yourself say that hover on/off/always icons on module header is odd.

I'm not so sure now based on the behaviour of the other buttons

Nilvus:
@TurboGit: no more WIP for me. If someone could help to remove hardcoded settings describe above, I could then propose an update PR.

phweyland:
>  If someone could help to remove hardcoded settings describe above

I've created some trees and check buttons. However I don't see what you mean by hardcoded...
I can think about two points.
- tag tree and list view, I think I've used a treeview option to keep the check button small (I can find it out if needed)
- import import  & metadata, I've a special css line to keep check button height smaller or equal to entry box height.

Nilvus:
@phweyland: thanks for reply but there's some (sort) of hardcoded things for those parts I cited. But for treeview, it's only on "more modules" part in bottom right of darkroom. All other works great and applied completely CSS settings.  Import metadata is also good and nothing to adjust there.

To resume, where I can't set some CSS (background and border only on hover mode) it's about treeview in more modules buttons, and all toggle buttons only (but for what I would like to test and update, it's only for group and help buttons in top right of main view and left state icon of module header (on/off/always icon).

Mark-64:
@Nilvus yes, the background of toggle buttons is hardcoded.
this is well know to me and fixing that is part of my reworking of buttons. It was actually present in #5074. 
This is because the background is manually drawn with a cairo rectangle. the oter effect is that togglebuttons with no border always look square, even if you set a border radius in CSS
I will repropose the fix in a later PR

Nilvus:
@TurboGit: this PR is over. I plan no more CSS updates by now and not until @Mark-64 work on icons will be over. So this could be merged.

TurboGit:
Ok, merging. Thanks.

jenshannoschwalm:
Pretty late comment: VERY good indeed!

Issue #5018, opened 2020-05-12
[FR] Static and ordered collections


phweyland:
**Is your feature request related to a problem? Please describe.**

Today it's possible to move around images in a given collection (custom order). But the images order is global. Then if you to shuffle another collection with the same images you loose the first order.

Furthermore the stack of recent collection is limited and one is going to loose the collections quickly.

Tagging covers a part of the need (static collection) but is not image position aware.

So in fact there is no way to keep in darktable a collection of images properly ordered (wedding, ...). This is an issue which has been discussed for a long time. See [here](https://discuss.pixls.us/t/virtual-album-collection-in-darktable/18139), [here](https://discuss.pixls.us/t/the-obvious-darktable-how-to-drag-and-drop-move-an-image-within-a-folder/6358), ...

**Describe the solution you'd like**

I would like to solve this and I open this FR to get comments, directions, suggestions ...

The solution I can see would be more or less this one.

A new module is called "collection", or "album", or anything more appropriate. This module  exposes the following actions:
- create collection
- delete collection
- add selected images to one of the collections
- remove selected images from current collection.

The collections are visible from collect module. As such, the user can select one of them and order the images as needed. 
The image (and the xmp) is not aware of collections (each image can belongs to several collections).

Challenge.
Modify the current global image order to turn it specific to each collection. 
From the image order standpoint the current collection, and potentially the 10 recent ones (?), are seen as collections and have each its own order.

Some doubts.
Outside the new static collections, do we need to keep a custom order ?
Does it exists any necessity to have a global order for all images of the database ? Which use case ?

What do you think ?

GrahamByrnes:
I think it's an excellent idea. I have some tag-collections which are already getting a bit too big, and I'm (very) far from a commercial photographer in volume. Of course I could go back and start adding secondary tags, but that is potentially a lot of retrospective work with no immediate pay-off... and life is short. 
Being able to pull the images that I'm excited about _now_ into a new structure would seem to fix this.

spaceChRiS:
Oh, wow, dreams becoming true ... Some random thoughts:
* Calling it album seams appropriate as collection is already used.
* Shall the albums are represented in a tree structure as well or only in a flat hierarchy, wedding-x-y-2020|album ..|presentation etc. vs wedding-x-y-2020-album and ..-presentation? Tree would be more compatible with tagging.
[...]

PR #5134, opened 2020-05-19
Static and ordered album or "images order" aware tags

phweyland:
Attempt to solve #5018 
Main changes
- add album attribute to tag
- add album filter entry in collect module
- add image order selection to album/tag info (highest 16 bits of flags)
- add position to tagged_images
- remove caption & description fields from images table
- position update on tag attachment
- position change on drag & drop for albums
- legacy params

For the user:
An album is a tag which has the "album" attribute (as private or category).

image

A new entry in collect module allows to filter the albums of the db. Unlike tag filter, the user can only see the images of that album, whatever its place in the hierarchy.

image

When the first collect filter is "album", any order change is saved with that specific album, including the custom sort image order.

If the first collect filter is not an album, the order changes are applied on the current collection, just as before.

### DIFF
--- a/src/common/collection.h
+++ b/src/common/collection.h
@@ -87,7 +87,6 @@ typedef enum dt_collection_properties_t
   DT_COLLECTION_PROP_FILMROLL = 0,
   DT_COLLECTION_PROP_FOLDERS,
   DT_COLLECTION_PROP_FILENAME,
-  DT_COLLECTION_PROP_ALBUM,
 
   DT_COLLECTION_PROP_CAMERA,
   DT_COLLECTION_PROP_LENS,
@@ -114,7 +113,7 @@ typedef enum dt_collection_properties_t
   DT_COLLECTION_PROP_HISTORY,
   DT_COLLECTION_PROP_MODULE,
   DT_COLLECTION_PROP_ORDER,
- 
+
   DT_COLLECTION_PROP_LAST
 } dt_collection_properties_t;
 
@@ -165,7 +164,7 @@ typedef struct dt_collection_t
   gchar *query, *query_no_group;
   gchar **where_ext;
   unsigned int count, count_no_group;
-  unsigned int album_id;
+  unsigned int tagid;
   dt_collection_params_t params;
   dt_collection_params_t store;
 } dt_collection_t;
@@ -208,8 +207,8 @@ void dt_collection_set_query_flags(const dt_collection_t *collection, uint32_t f
 
 /** set the film_id of collection */
 void dt_collection_set_film_id(const dt_collection_t *collection, const uint32_t film_id);
-/** set the album_id of collection */
-void dt_collection_set_album_id(dt_collection_t *collection, const uint32_t album_id);
+/** set the tagid of collection */
+void dt_collection_set_tag_id(dt_collection_t *collection, const uint32_t tagid);
 /** set the star level for filter */
 void dt_collection_set_rating(const dt_collection_t *collection, uint32_t rating);
 /** get the star level for filter. The value returned starts on 0 **/
@@ -263,9 +262,9 @@ void dt_collection_split_operator_number(const gchar *input, char **number1, cha
 void dt_collection_split_operator_datetime(const gchar *input, char **number1, char **number2, char **op);
 void dt_collection_split_operator_exposure(const gchar *input, char **number1, char **number2, char **op);
 
-int64_t dt_collection_get_image_position(const int32_t image_id, const int32_t album_id);
+int64_t dt_collection_get_image_position(const int32_t image_id, const int32_t tagid);
 void dt_collection_shift_image_positions(const unsigned int length, const int64_t image_position,
-                                         const int32_t album_id);
+                                         const int32_t tagid);
 
 /* move images with drag and drop */
 void dt_collection_move_before(const int32_t image_id, GList * selected_images);
--- a/src/common/collection.h
+++ b/src/common/collection.h
@@ -79,12 +79,15 @@ typedef enum dt_collection_sort_t
   DT_COLLECTION_SORT_SHUFFLE
 } dt_collection_sort_t;
 
+#define DT_COLLECTION_ORDER_FLAG 0x8000
+
 /* NOTE: any reordeing in this module require a legacy_preset entry in src/libs/collect.c */
 typedef enum dt_collection_properties_t
 {
   DT_COLLECTION_PROP_FILMROLL = 0,
   DT_COLLECTION_PROP_FOLDERS,
   DT_COLLECTION_PROP_FILENAME,
+  DT_COLLECTION_PROP_ALBUM,
 
   DT_COLLECTION_PROP_CAMERA,
   DT_COLLECTION_PROP_LENS,
@@ -162,6 +165,7 @@ typedef struct dt_collection_t
   gchar *query, *query_no_group;
   gchar **where_ext;
   unsigned int count, count_no_group;
+  unsigned int album_id;
   dt_collection_params_t params;
   dt_collection_params_t store;
 } dt_collection_t;
@@ -203,7 +207,9 @@ uint32_t dt_collection_get_query_flags(const dt_collection_t *collection);
 void dt_collection_set_query_flags(const dt_collection_t *collection, uint32_t flags);
 
 /** set the film_id of collection */
-void dt_collection_set_film_id(const dt_collection_t *collection, uint32_t film_id);
+void dt_collection_set_film_id(const dt_collection_t *collection, const uint32_t film_id);
+/** set the album_id of collection */
+void dt_collection_set_album_id(dt_collection_t *collection, const uint32_t album_id);
 /** set the star level for filter */
 void dt_collection_set_rating(const dt_collection_t *collection, uint32_t rating);
 /** get the star level for filter. The value returned starts on 0 **/
@@ -257,8 +263,9 @@ void dt_collection_split_operator_number(const gchar *input, char **number1, cha
 void dt_collection_split_operator_datetime(const gchar *input, char **number1, char **number2, char **op);
 void dt_collection_split_operator_exposure(const gchar *input, char **number1, char **number2, char **op);
 
-int64_t dt_collection_get_image_position(const int32_t image_id);
-void dt_collection_shift_image_positions(const unsigned int length, const int64_t image_position);
+int64_t dt_collection_get_image_position(const int32_t image_id, const int32_t album_id);
+void dt_collection_shift_image_positions(const unsigned int length, const int64_t image_position,
+                                         const int32_t album_id);
 
 /* move images with drag and drop */
 void dt_collection_move_before(const int32_t image_id, GList * selected_images);
--- a/src/common/database.c
+++ b/src/common/database.c
@@ -44,7 +44,7 @@
 
 // whenever _create_*_schema() gets changed you HAVE to bump this version and add an update path to
 // _upgrade_*_schema_step()!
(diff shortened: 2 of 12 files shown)


src/common/collection.h
```diff
@@ -76,15 +76,18 @@ typedef enum dt_collection_sort_t
   DT_COLLECTION_SORT_TITLE,
   DT_COLLECTION_SORT_DESCRIPTION,
   DT_COLLECTION_SORT_ASPECT_RATIO,
-  DT_COLLECTION_SORT_SHUFFLE
+  DT_COLLECTION_SORT_SHUFFLE,
```
TurboGit, line 79: this is unwanted

src/common/database.c
```diff
@@ -1611,6 +1611,56 @@ static int _upgrade_library_schema_step(dt_database_t *db, int version)
...
+
+    TRY_EXEC("INSERT INTO main.i SELECT id, group_id, film_id, width, height, filename, maker, model,"
+             " lens, exposure, aperture, iso, focal_length, focus_distance, datetime_taken, flags,"
+             " output_width, output_height, crop, raw_parameters, raw_denoise_threshold,"
+             " raw_auto_bright_threshold, raw_black, raw_maximum, license, sha1sum,"
+             " orientation, histogram, lightmap, longitude, latitude, altitude, color_matrix, colorspace, version,"
+             " max_version, write_timestamp, history_end, position, aspect_ratio, exposure_bias,"
+             " import_timestamp, change_timestamp, export_timestamp, print_timestamp "
+             "FROM main.images",
+             "[init] can't populate table i\n");
```
TurboGit, line 1655: when doing that the index are discarded. You need to recreate them:
- image_position_index   (note that there is no s on this one)
- images_filename_index
- images_film_id_index
- images_group_id_index

src/common/image.c
```diff
@@ -865,8 +865,13 @@ int32_t dt_image_duplicate_with_version(const int32_t imgid, const int32_t newve
     sqlite3_finalize(stmt);
 
     DT_DEBUG_SQLITE3_PREPARE_V2(dt_database_get(darktable.db),
-                                "INSERT INTO main.tagged_images (imgid, tagid)"
-                                "  SELECT ?1, tagid FROM main.tagged_images WHERE imgid = ?2",
+                                "INSERT INTO main.tagged_images (imgid, tagid, position)"
+                                "  SELECT ?1, tagid, "
+                                "        (SELECT (IFNULL(MAX(position),0) & (4294967295 << 32))"
```
TurboGit, line 870: 4294967295 << 32  ???? Isn't this expected to be "2 << 32" ?

The above is a 64 bits integer where the 32 first bits (high order) are set to 1. Maybe expected, but I prefer to ask, not clear reading the code.

src/common/image.c
```diff
@@ -1210,9 +1215,9 @@ static uint32_t dt_image_import_internal(const int32_t film_id, const char *file
   //insert a v0 record (which may be updated later if no v0 xmp exists)
   DT_DEBUG_SQLITE3_PREPARE_V2
     (dt_database_get(darktable.db),
-     "INSERT INTO main.images (id, film_id, filename, caption, description, license, sha1sum, flags, version, "
+     "INSERT INTO main.images (id, film_id, filename, license, sha1sum, flags, version, "
      "                         max_version, history_end, position, import_timestamp)"
-     " SELECT NULL, ?1, ?2, '', '', '', '', ?3, 0, 0, 0, (IFNULL(MAX(position),0) & (4294967295 << 32))  + (1 << 32), ?4 "
+     " SELECT NULL, ?1, ?2, '', '', ?3, 0, 0, 0, (IFNULL(MAX(position),0) & (4294967295 << 32))  + (1 << 32), ?4 "
```
TurboGit, line 1220: likewise ?

src/common/image.c
```diff
@@ -1811,9 +1816,12 @@ int32_t dt_image_copy_rename(const int32_t imgid, const int32_t filmid, const gc
         sqlite3_step(stmt);
         sqlite3_finalize(stmt);
         DT_DEBUG_SQLITE3_PREPARE_V2(dt_database_get(darktable.db),
-                                    "INSERT INTO main.tagged_images (imgid, tagid)"
-                                    " SELECT ?1, tagid"
-                                    " FROM main.tagged_images"
+                                    "INSERT INTO main.tagged_images (imgid, tagid, position)"
+                                    " SELECT ?1, tagid, "
+                                    "        (SELECT (IFNULL(MAX(position),0) & (4294967295 << 32))"
```
TurboGit, line 1821: likewise.

src/common/tags.c
```diff
@@ -63,7 +63,13 @@ static gchar *_get_tb_added_tag_string_values(const int img, GList *before, GLis
   {
     if(!g_list_find(b, a->data))
     {
-      tag_list = dt_util_dstrcat(tag_list, "(%d,%d),", GPOINTER_TO_INT(img), GPOINTER_TO_INT(a->data));
+      tag_list = dt_util_dstrcat(tag_list,
+                                 "(%d,%d,"
+                                 "  (SELECT (IFNULL(MAX(position),0) & (4294967295 << 32)) + (1 << 32)"
```
TurboGit, line 68: likewise

phweyland:
>  This could really be done with a standard tag, right? The point here is to make more visible a way to have images grouped in album and make this a first class citizen. Right ?
> 
> Now the question is do we really want this added complexity where one could use a hierarchical tag structure

Good questions... This is more a question of usage.

I started with the idea of album concept but ended up by having the sort property on all tags (just because it is easier and faster to apply it everywhere).
So, in collect module, having tag as the first filter, we could have the "album" effect for any tag with the same code.
The single drawback I see is that I find tag filter in collect module a bit heavy and slow. Having the album filter can make it a bit faster.

It would be interesting to get users' opinion here !

A more technical aspect:
Another possibility I thought about is to remove position from images tables and use only tagged_images table position. Having a per default tag for all images (like darktable|something) we can mirror the current behavior of the normal sorting. I haven't gone there because the position sorting on images table should stay faster than on tagged_images table (far less records in the first table) but I haven't run any measure.

src/common/collection.h
phweyland: right !

src/common/database.c
phweyland: Right.

phweyland:
Updated with fixes related to your comments.

I'll remove all album attributes keeping order on tags on the next push. This should simplify the code. If not convenient we'll be able to come back to this commit.

src/common/image.c
```diff
@@ -1811,9 +1818,13 @@ int32_t dt_image_copy_rename(const int32_t imgid, const int32_t filmid, const gc
         sqlite3_step(stmt);
         sqlite3_finalize(stmt);
         DT_DEBUG_SQLITE3_PREPARE_V2(dt_database_get(darktable.db),
-                                    "INSERT INTO main.tagged_images (imgid, tagid)"
-                                    " SELECT ?1, tagid"
-                                    " FROM main.tagged_images"
+                                    "INSERT INTO main.tagged_images (imgid, tagid, position)"
+                                    " SELECT ?1, tagid, "
+                                    // 4294967295 << 32 = 0xFFFFFFFF00000000
```
TurboGit, line 1823: Looks like SQlite3 supports 0xFFFFFFFF00000000 directly have you tried to replace:
`
(4294967295 << 32)`
by
`0xFFFFFFFF00000000`

Maybe clearer to the reader.

phweyland:
In that commit I've removed all references to album.
Then tags are now order aware when in first place in collect filter.

src/common/image.c
phweyland: I'd read somewhere (probably not in the right place) that we could not do that. I haven't given a try....
Done.

phweyland:
0xFFFFFFFF00000000 ! :)

phweyland:
minor formatting plus one fix.

TurboGit:
>     * add position to tagged_images

This part is not clear, why do we want to have the position of a tag ? This makes the tagged_images more than just a unique relation between an imgid and a tagid. So in turns this means that we can have the same tag at different position ? Again, this looks quite complex to me and could certainly hinder the maintenance!

phweyland:
> This part is not clear, why do we want to have the position of a tag ?

It is not adding position to a tag.

This adding a position for the the couple (tagid, imagid).
Then for each tagid, images cab be ordered.

phweyland:
> This makes the tagged_images more than just a unique relation between an imgid and a tagid.

No, this doesn't change anything to the current role of tagged_images.

Just adding the position attribute to (tagid, imgid). This position is used in collect module instead of the images table position when the user has selected a tag.
So, as we have the global ordering today, this change adds the possibility (with custom sort order) order manually the images for each tag.

To resume:
If the first collect filter is anything else than tag the images order is as before (no change).
When the first collect filter is tag,  any change on the order selection is kept along the tagid (in flags of tags table). And if the sort order is set as custom sort, the user can drag and drop the images in a different order. This order is kept in tagged_images position.

TurboGit:
I see, I reread the corresponding issue. This is clearer to me now. I'll review and test again in the coming days with this in mind. Far more challenging that what I thought initially !

TurboGit:
Ok, tested and working on my side. Thanks Philippe and sorry for all my questions, I have hard time to follow closely everything.

phweyland:
> and sorry for all my questions

Pascal, you do a tremendous job ! That's absolutely normal you ask questions. For sure you have to understand what's happening. And if it is not clear, most of the time that's because the presentation or the explanations are not good enough.

Sometimes I have the impression that dt changes are going a bit fast :).

Thanks for the questions !

Nilvus:
@phweyland: if you can, could you read this issue on french forum: https://forums.darktable.fr/showthread.php?tid=4971&pid=42180#pid42180
A user describe an issue with film roll with clones inside. And the master he use is the last one from same day this PR is merged and it's the only thing related to database as he talks about an update.

phweyland:
> if you can, could you read this issue on french forum: https://forums.darktable.fr/showthread.php?tid=4971&pid=42180#pid42180

I've posted a question there but I don't see my message... not sure it has gone there ...
I've some duplicates in my database and I've not reproduced yet the issue.

spaceChRiS:
> Pascal, you do a tremendous job !

Fully agree here. And @phweyland, you too! And all the other contributors. What a magnificent project this is. Since this PR is a bit my fault (k, not really a fault), a big thank you to both of you for implementing and testing! This was extremely fast, I cannot say how thankful I am. This is incredible :smile:.

Unfortunately, I cannot test as darktable no longer builds on my ubuntu 16.04. I am working on this, but it will take a couple of weeks :expressionless:.

I wonder if allowing the sorting for all tags will mess up the data base and produce a lot of unnecessary entries? May it therefore not be better to restrict this feature to tags that have the album property, as in your original implementation?

phweyland:
> I wonder if allowing the sorting for all tags will mess up the data base and produce a lot of unnecessary entries?

phweyland:
> I wonder if allowing the sorting for all tags will mess up the data base and produce a lot of unnecessary entries?

I started this way but it adds a lot of specific cases and is slower especially acting on db. Applying it on all tags keeps it simple. Furthermore the tag order only intervenes when tag is the first entry of collect module. Otherwise everything works as before. 
I'm looking forward to your feedback on this!

jerrac:
I tried installing the latest stable version from the OBS repo, then the latest snapshot version from the OBS snapshot repo, then I finally compiled the master branch and installed it. Every time I edit a tag, there is no "album" checkbox.

So, how does the feature this pull request is about actually work?

PR #5135, opened 2020-05-19
German translation updates

rabauke:
mainly translation of key bindings shown in the accelerators overview in lighttable view


TurboGit:
Thanks !

PR #5137, opened 2020-05-20
Make sure color picker gets applied after enabling

### LINKED ISSUE #5125: Regression in WB module after Commit 62670ed
**Describe the bug**

In the past (at least the recent weeks) WB spot was like this:
- pick a spot, WB will be adjusted
- for what ever reason, one would like to check "camera" or others again, so click something else and get adjustments
- subsequently wanna go **back to spot**
-- the former frame is displayed and the value is set, as in the first spot picking = perfect

What has changed?
- pick up spot, works as normal
- click "camera" for instance and get the related WB values
- chose spot again
-- -> the former spot is framed = correct
-- -> the **WB setting** remains **unchanged** = **incorrect**
-- (reframing and it reacts again)

**To Reproduce**

see above

**Expected behavior**
as it was before: when going back to spot and the frame is shown, WB should be adjusted to that frame.

**Bisect**

I bescted this and got:
```
62670ed9ca4b24f6ad0f42adc49231a739e0b6e1 is the first bad commit
commit 62670ed9ca4b24f6ad0f42adc49231a739e0b6e1
Author: Diederik ter Rahe <dterrahe@yahoo.com>
Date:   Thu May 14 11:43:08 2020 +0100

    reduce complexity of color_picker_proxy by not special casing blender pickers

 src/develop/blend_gui.c      |  24 ++--
 src/develop/imageop.c        |   2 -
 src/develop/imageop.h        |   8 +-
 src/gui/color_picker_proxy.c | 314 +++++++++++++------------------------------
 src/gui/color_picker_proxy.h |  68 +---------
 src/iop/basicadj.c           |   3 -
 src/iop/colorzones.c         |  41 +++---
 src/iop/rgbcurve.c           |  25 ++--
 src/iop/toneequal.c          |   2 +-
 9 files changed, 142 insertions(+), 345 deletions(-)
```
@dterrahe could you pls. have a look into it, even I cannot understand, how this could be responsible for it.

 - Darktable Version:  3.1.0+1700~g3c3ea3119
 - OS:  Gentoo

**Additional context**
 - Can you reproduce with another Darktable version? **see bisect**
 - Are the steps above reproduce with a fresh edit (removing history)? **yes**
-- gcc 9.0.3 (during bisecting, by chance 8.0.3 was set, but cross checked after switching back to 9.0.3)
-- CMAKE_BUILD_TYPE="Release" 
-- CFLAGS="-march=native -O2 -mtune=native -pipe"
 - Is the issue still present using an empty/new config-dir > **Yes**

dterrahe:
Fixes #5125 

62670ed9ca4b24f6ad0f42adc49231a739e0b6e1 changed color picker behavior to only trigger a color_picker_apply callback if the selected region changed. However, an apply should always be triggered immediately after enabling the color picker, otherwise for example switching to white balance/spot doesn't have any effect because the previously selected region is re-enabled.

Many thanks to @AxelG-DE for the bug report.

### DIFF
--- a/src/gui/color_picker_proxy.c
+++ b/src/gui/color_picker_proxy.c
@@ -188,12 +188,14 @@ static gboolean _iop_color_picker_callback_button_press(GtkWidget *button, GdkEv
       float box[4];
       _iop_get_area(self, box);
       dt_lib_colorpicker_set_box_area(darktable.lib, box);
+      self->pick_pos[0] = NAN; // trigger difference on first apply
     }
     else
     {
       float pos[2];
       _iop_get_point(self, pos);
       dt_lib_colorpicker_set_point(darktable.lib, pos[0], pos[1]);
+      self->pick_box[0] = NAN; // trigger difference on first apply
     }
 
     module->dev->preview_status = DT_DEV_PIXELPIPE_DIRTY;


TurboGit:
Works for me! Thanks.

AxelG-DE:
> Many thanks to @AxelG-DE for the bug report.

You are very welcome!

Works like a charm again. Thank you!

PR #5140, opened 2020-05-20
Fix: Appveyor: implement known lensfun version workaround per build.txt

coffeeandspaceships:
I have added scripting to implement the workaround described in the Windows [build instructions](https://github.com/darktable-org/darktable/blob/master/packaging/windows/BUILD.txt). 

- Changes are together in one section of the script to help clarity
- These can be removed again when dt is integrated with lensfun-next, but the timescales for this are unknown, and dependent on upstream
- There is a manual download of the MSYS64 package for lensfun 0.3.2-4 with an explicit URL. I think this risk is acceptable because if MSYS64 change the repo and remove that version, it's very likely the repo changes will break the build in other places anyway.
- The MSYS64 package 0.3.2-4 was packaged with the python version at the time, 3.6; MSYS64 python is now at 3.8 so there is a step included to copy the files across.

### DIFF
--- a/.ci/ci-script-windows.sh
+++ b/.ci/ci-script-windows.sh
@@ -82,6 +82,14 @@ cat > "$FONTCONFIG_FILE" <<EOF
 EOF
 
 execute 'Installing base-devel and toolchain'  pacman -S --needed --noconfirm mingw-w64-x86_64-{toolchain,clang,cmake}
-execute 'Installing dependencies' pacman -S --needed --noconfirm  mingw-w64-x86_64-{exiv2,lcms2,lensfun,dbus-glib,openexr,sqlite3,libxslt,libsoup,libwebp,libsecret,lua,graphicsmagick,openjpeg2,gtk3,pugixml,libexif,osm-gps-map,libgphoto2,flickcurl,drmingw,gettext,python3,iso-codes}
+execute 'Installing dependencies (except lensfun)' pacman -S --needed --noconfirm  mingw-w64-x86_64-{exiv2,lcms2,dbus-glib,openexr,sqlite3,libxslt,libsoup,libwebp,libsecret,lua,graphicsmagick,openjpeg2,gtk3,pugixml,libexif,osm-gps-map,libgphoto2,flickcurl,drmingw,gettext,python3,iso-codes}
+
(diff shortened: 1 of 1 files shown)


coffeeandspaceships:
Re "clang build still fails, but for unrelated reasons" - I now have a branch that's looking hopeful for clang - should I cherry-pick into this PR once I've proved it's working?

coffeeandspaceships:
@TurboGit I think this is good to go now. 
- The two code changes have passed the Travis check so my #ifdef guards must be working.
- The other two commits only affect appveyor scripts, so even if they have a problem, it will still be less of a problem than the current master stale file which does not allow a build at all.

If you are not the right reviewer can you please suggest one of the other maintainers/reviewers?

coffeeandspaceships:
image
Evidence that both Appveyor jobs - gcc and clang - completed successfully last night, triggered by the final commit to this PR.

aurelienpierre:
@coffeeandspaceships I know nothing about dev on Windows, but do you know if it would be possible to have a Windows CI producing .exe snapshots of master automatically ?

coffeeandspaceships:
@aurelienpierre Yes. It is working for me already on my repo; I usually merge this branch into branches I am working on, for that reason. I had a release up for a while but it started causing confusion with other people downloading it, so I removed it.

coffeeandspaceships:
I actually came back here to say it would be great to have this in 3.2 please so anyone who forks 3.2 gets a working AppVeyor build, even if the maintainers don't want it.

But in particular 5442f631c41ae383bc26bc917a6306f1af73ec6a is essential for 3.2 because without that commit (currently on master but not in 3.0.2) , NO clang builds under Windows will work, even locally on a Windows dev box.

coffeeandspaceships:
@turbogit With 10 days to go... any chance for 3.2 please? If not let me know blockers so I can sort for 3.4.

TurboGit:
Ok, this seems ok but I'm not a windows user. Will merge, please watch for any regression to be fixed. Thanks.

coffeeandspaceships:
Brilliant, thanks for merging. 

Note I found out yesterday that Windows support in Travis is rapidly catching up - I'll keep an eye on that and at some time in the future we can have CI for Windows via Travis, which means only one tool to worry about. In the meantime I can now use AppVeyor from my fork without needing to branch from master so managing regressions should be a lot easier following this merge, thanks.

PR #5148, opened 2020-05-21
thumbtable: only show border around active image(s)

### LINKED ISSUE #5144: lighttable: flashing borders around images randomly (ish)
When I exit the darkroom back to the lighttable (file manager), the image I've just edited and the last image in the collection both have a slowly flashing white border around them for a few seconds:

Screenshot_2020-05-20_21-41-51

It seems that the last image in the collection only flashes if I've edited it in the current session.

In fact, testing further, if I edit every image in turn starting from the last, each time I re-enter the lightroom all of the previously edited images have the same flashing white border, and when I reach the first image in the collection, every image in the collection has the same white border. 

Screenshot_2020-05-20_21-46-02

The same doesn't happen if I edit the images in a different order.

Does this border have a purpose / intended behaviour?

@AlicVB?

Issue on 3.1.0+1703~g989d68d45 with and without OpenCL.

elstoc:
Resolves #5144

### DIFF
--- a/src/dtgtk/thumbtable.c
+++ b/src/dtgtk/thumbtable.c
@@ -1661,6 +1661,8 @@ void dt_thumbtable_full_redraw(dt_thumbtable_t *table, gboolean force)
       if(tl)
       {
         dt_thumbnail_t *thumb = (dt_thumbnail_t *)tl->data;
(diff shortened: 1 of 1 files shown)


elstoc:
Just pushed a minor amendment - this is only required where the thumb already exists

TurboGit:
Thanks !

PR #5150, opened 2020-05-21
Icon buttons layout, another attempt

Mark-64:
this is another try to make DT GUI scale correctly for all OSs and HiDPI screens, although the improvement is more evident on non-Linux OSs.
I will not repeat all the history here, you can find in #5074 and precedents.

The previous PR was controversial due to me putting two different things:
1. elimination of hardcoded margins in buttons and enforcing CSS geometry rules instead
2. reworking of some icons
No. 2. in particular generated discussion about optical alignment, balancing etc.

Therefore I present here a minimal set of changes for point No.1 which is the core of the whole idea.
Other steps are necessary (including No. 2) in order to get optimal results, but I will propose them separately, if this PR is accepted.

With the PR the icon buttons (nothing is changed about Gtk label buttons) will fully follow the CSS geometry rules (margin, border, padding, etc.). This will allow a better scalability and indeed my tests on Windows HiDPI show already a much better behaviour.

**Note on optical alignment**
Indeed we need optical alignment and although the icons look decently balanced on Linux with the master branch, this is a fragile equilibrium based on hardcoded margins and ad-hoc fixes, which does not survive a change of OS and scren DPI, neither will survive this PR. To get a more solid solution, we will have to touch the icons with future PRs.
Although this PR does not touch the icons, it implements an extra room in order to give the possibility to the cairo painting functions to possibly draw slightly outside the bounding box in order to allow for optical balancing.
This is implemented by putting a Gtk drawing area inside the buttons. The drawing area will not influence the button allocation, but it will make available new CCS properties.
In particular, the CSS margin of the drawing area is used as a percentage of extra room (tricking Gtk, which does not support size in percent, same technique used by @AlicVB in thumbtable).
This way the icon "overbooking" is controllable from CSS.

CSS must be reworked. I attach here a darktable.css file tweaked by me, which you can use for testing. it's not optimized, but it gives you an idea. Inside there are comments on the main changes

[darktable.css.txt](https://github.com/darktable-org/darktable/files/4664404/darktable.css.txt)

**Next steps**
If this PR finds support, I plan to do the following next:

1. Fix for hardcoded background in toggle buttons (see #5132): I could actually add this here as a commit if you agree (EDIT: done)
2. Icon fixes and optical balancing. I will do this in batches for group of icons sitting next to each other
3. CSS reworking (I can only help testing here)
4. Code cleanup (there is a lot of no longer used code after this changes)


AlicVB:
Great ! I'll have a look at the code, but no in the following days...
Just a note, if I understand well your point about optical margins : 
As you use margin in percentage, maybe it would be simpler to just add a parameter in buttons to define the needed room for optical margins in percentage too. Then, when you draw the icon, you just have to subtract them (and ensure it's >0)
If css define margins to less than the needed room, it'll just reduce the optical margin correction.
That way, you may avoid the extra drawing area widget.
Again, that's just word, as I've not dig in your code atm. So forget that if that's not relevant

Mark-64:
> it would be simpler to just add a parameter in buttons to define the needed room for optical margins in percentage 

this is what I did initially and it works in most cases, but not in all.
I put 12.5% extra room each side, but in some cases you need to put almost no room at all, for example the choose folder icon in the export module, and if it is hardcoded you can't change.
It's not a big complication in CSS, just a statement to set the default:

```
/* default extra margin for icons optical alignment */
#button-canvas,
#toggle-button-canvas
{
  margin: 12px; /* not real pixels, this is used as a percent extra space */
}
```
and then just tweak the few cases where a change is needed.

Nilvus:
Good. I will also see that in the following days. There's sun and 3 days off to be outside before.

Nilvus:
> * Fix for hardcoded background in toggle buttons (see #5132): I could actually add this here as a commit if you agree

"more modules" treeview on darkroom need also same fix (if you can). I know it's not an icon but except buttons, that's the only treeview with same problem. And so no hover background effect possible by now.

> * Icon fixes and optical balancing. I will do this in batches for group of icons sitting next to each other
> 
> * CSS reworking (I can only help testing here)

Of course, I'm on for CSS reworking. So, that will be my work (and you could after test it, especially for Windows and Hidpi). Maybe some optical margins could be set with CSS. To see later, when hardcoded things 1st step) will be fixed.

Nilvus:
@Mark-64: this PR has conflict with your last PR #5152 and with that PR square hover buttons come back. So could you update it with your #5152 PR work? And it would also be good to allow setting border size and color on all icons. It's not possible by now, so things remains hardcoded.

Other than that, tested quickly this PR and, with your CSS proposal it's good. Maybe you could add what you set on the CSS file you propose here, in this PR, based on my last CSS work merged on master?

Mark-64:
@Nilvus 

> And it would also be good to allow setting border size and color on all icons. It's not possible by now

Why not ?
image

> Maybe you could add what you set on the CSS file you propose here, in this PR, based on my last CSS work merged on master?

I can try, but there are still quite some tweaks to do, and I am not sure I am good enough at CSS to do them.
I am afraid if we merge with my CSS we will get complains

BTW I had a look to the more module treeview and yes it is coded differently form all the others. While it is obvious it has hardcoded background color, it's not obvious to me how to fix it.

Nilvus:
> Why not ?
> image
Hmm, sorry was not clear. I of course didn't talk about those ones as they already are borders sets. I was just talking on some other that doesn't accept any borders (but that's not the more important thing as it's only to test some tweaks and remove all hardcoded things). Icons where border size and color is not accepted are group, prefs, help... and module-collapse-button or module-enable-button (at least).

> I can try, but there are still quite some tweaks to do, and I am not sure I am good enough at CSS to do them.
> I am afraid if we merge with my CSS it will look like a regression and we will get complains

I will be there of course to improve CSS after (but I could also see which lines you change and work on that... As you want, but your PR is better with CSS you propose than without. And of course, if you add CSS, just add your tweaks about this PR. 
 
> BTW I had a look to the more module treeview and yes it is coded differently form all the others. While it is obvious it has hardcoded background color, it's not obvious to me how to fix it.

I try (sometimes) to better understand code but hard for me and do not want to spent on this. Not really related but I just see an error on this treeview in film.c on line 88. There's a CSS tag for this module set to lib-modulelist but it's not shown on Gtk Inspector tool. On this line, if I change self->widget to d->tree, the tag then appears on Gtk Inspector tool. Unfortunately, that just fix to set that tag, nothing else.

Mark-64:
Now this is ready for review.
@Nilvus 
I put the darktable.css in the PR, as you requested, rebased and further refined.
Eventually, also the structure of css is simplified when coming to buttons.
The extra drawing area widget inside the buttons has proven to be necessary, as for the small buttons in modules the default icon extra room of 12% makes the icon look inconspicuous and needs to be reduced.
To me it looks pretty good, but please have a final check.

Now I undesrtand what you mean about button borders, so you want to get rid of the whole CPF_xxx flags which are hardcoding the style of the buttons.
I fully agree to move everything to css and I am ready to help you doing this, as it would also result in a substantial simplification of the code.
However it's quite a change for css, so let's first push this PR and then we go step by step.

Nilvus:
@Mark-64: Great work, that's clean and with your CSS tweaks, I see nothing to change. All clean and on my system, the UI doesn't even change. Except little minor fixes that I fixed yesterday and no more needed now. So good to be merge for me, to have more feedback on this.

About button borders (and more modules treeview), yes main goal is to have less hardcoded things as possible. But about buttons, no need to remove all things as all things could be set with CSS except border size and color. Hope it will not be too tricky to do that. But of course, you're right to propose to go step by step. Anyway, simplify substantially the code would be a great side-effect.

Mark-64:
@TurboGit Can you have a look ? I also have other GUI optimizations ready to be proposed, but they all need this as a baseline ;-)

TurboGit:
Looks good to me too. Thanks!

PR #5152, opened 2020-05-21
make toggle button background follow CSS

Mark-64:
This will make the toggle button honor css for background settings for color and border radius.
the issue is mentioned in #5132 
Problem was that the background was manually drawn with a cairo rectangle

it's part of my reworking of buttons, but I propose here separately

### DIFF
--- a/src/dtgtk/togglebutton.c
+++ b/src/dtgtk/togglebutton.c
@@ -45,17 +45,9 @@ static gboolean _togglebutton_draw(GtkWidget *widget, cairo_t *cr)
 
   GtkStateFlags state = gtk_widget_get_state_flags(widget);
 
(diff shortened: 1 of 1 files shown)


Nilvus:
Oh yeah, like that. Thanks, will test ASAP. WIll remains now the more modules treeview...

Nilvus:
Tested and it's perfect. Works great and all buttons have same hover effect now. Good to see that. @TurboGit: this is ok for me.

Mark-64:
I will look in the treeview, but do not expect a quick fix, I am not familiar with that part of code

TurboGit:
@Mark-64 : Thanks for the patch. @Nilvus : Thanks for testing and approving.

Merging.

Issue #5123, opened 2020-05-19
Initial size of patches in liquify after changing to small image


GrahamByrnes:
**Describe the bug**

Open a raw image in liquify, something around 24MPx. Create a warp-patch that is around 2/3 or larger than the height or width of your image. Now close that image and open a small tif of say 400x600Px, go into liquify and press the point tool.

I see an initial warp circle that completely covers the image and has inaccessible control points.

It seems that this is just a parameterisation problem: if you go to `develop/masks/circle.c` you find there is a `dt_conf_get_sanitize` which shrinks previously saved parameters from `darktablerc` down to something compatible with the current display. However this is liquify.c, which is a whole world of its own.

I'll probably try to fix this in the next couple of days, but thought it would be good to get my diagnosis confirmed (or rejected).

Windows 10, cmake, dt310+1680, no openCL.

GrahamByrnes:
BTW, this will probably need to be fiwed soon too:

```
float dt_liquify_ui_widths [] =
{
  // value in 1/96 inch (that is: in pixels on a standard 96 dpi screen)
    2.0, // DT_LIQUIFY_UI_WIDTH_THINLINE
    3.0, // DT_LIQUIFY_UI_WIDTH_THICKLINE
    3.0, // DT_LIQUIFY_UI_WIDTH_DOUBLELINE
    9.0, // DT_LIQUIFY_UI_WIDTH_GIZMO
    7.0, // DT_LIQUIFY_UI_WIDTH_GIZMO_SMALL
  100.0, // DT_LIQUIFY_UI_WIDTH_DEFAULT_RADIUS,
   50.0, // DT_LIQUIFY_UI_WIDTH_DEFAULT_STRENGTH,
    4.0  // DT_LIQUIFY_UI_WIDTH_MIN_DRAG
};
```

TurboGit:
To me your analysis is correct!

PR #5154, opened 2020-05-21
Dodgy wip fix of liquify.c stamp-size variation

GrahamByrnes:
As described in #5123, there is a bug in `liquify.c`: the size of the initial stamp is typically ok for a raw file, but too big to be able to access control points whan applied to a smallish tiff.

This partially fixes that: patches in raw and tiff files open at the same size.
However:
- there is a strange thing where clicking inside the stamp on a small file causes a strange graphic with 3 concentric circles. This seems to be related to the use of preview_downsampling<1
- bizarely, jpg images give  a different size stamp than either tiff or raw.
- I used a hack scaling value of 0.005 to get the initial stamp diameter. It may be wrong on different screens.

On the plus side, the scale beween the patch sizes is about 2, rather than >10 as it was before, so it's usable.

Otherwise:
- the first commit is a clean-up related to initialisation and my allergic reaction to spaces before function arguments. 
- this includes the issue of initialising variables `float complex` According to IBM, after C99, if `complex.h` is included, this should be done via `CMPLXF(0.0f, 0.0f)` ... but msys64 does not recognise this. I just set `float complex pt = 0.0f;` and it built. Let's see what Travis thinks.


GrahamByrnes:
Oh, just realised the "strange graphic with 3 concentric circles" is the feather control mode. So in fact it's just that the control gets sticky when the circle is small, and it's easy to accidentally change modes. But it's still kind of broken, because it's bloody hard to change anything: this might be one of those cases where preview_downsampling is an issue, since taking a small circle of a small image and down sampling, there are not so many pixels left to easily click on...

GrahamByrnes:
Could someone tell me where the number of pieces used to tile the original image is stored?

GrahamByrnes:
I decided to kill the re-formatting, there is too much: I'll re-submit that later if there is any enthusiasm.

So the proposed fix is imperfact. I've tested it on 1024x684 and 2048x1368 jpg's; on 24MP & 12MP DNG/RAF raw files, and on a 1440x2000 and a 564x800 tif. The variation in the stamp size is maybe 50-60% and it always fits inside the image.

@TurboGit What do you think, merge as is? Or should I keep trying for the ideal solution? I'm not sure I'm going to find it, and the difference is just the initial stamp size in a rarely used module... ?

TurboGit:
>     * there is a strange thing where clicking inside the stamp on a small file causes a strange graphic with 3 concentric circles. This seems to be related to the use of preview_downsampling<1

This is ok as you found out later.

>     * bizarely, jpg images give  a different size stamp than either tiff or raw.

Yes, just tried for the first time liquify on a jpeg and there is indeed something very strange. Not only the initial size is different but it is hard to pick the control point or to delete the stamp as the visible control points does not seems to be aligned with the mouse cursor. Some offset somewhere.... This is a real bug and may explain the whole issue (different size).

I don't remember code (I've been the one to do final work for the dt integration) but I think we should fall back to standard size computation as done for all masks in dt. It seems quite wrong to try to have a size relative to the current size of the image. Right?

TurboGit:
> * bizarely, jpg images give  a different size stamp than either tiff or raw.

This is not a RAW vs Jpeg vs TIFF issue but a size issue. A radius of 100 on a RAW with a width of 5000 px will looks smaller than a radius of 100 on a Jpeg of width of 2000 px. So not bug here in fact.

Also the issue I reported above was transient, I cannot reproduce now.

So at the end what do we want to fix really ? Initial size does not seems to be an issue after all. I suppose we have the same for standard mask circle, no ? I don't remember that the initial size is adjusted on the image size.

GrahamByrnes:
I can't see a good reason why it should be different.

I'm currently at a stage where the size is _reasonably_ stable across jpg, tif and raw, for different choices of down_sampling: the magic scaling factor is

  ```
  const dt_dev_pixelpipe_t *devpipe = darktable.develop->preview_pipe;
  const float iwd_min = MIN(devpipe->iwidth, devpipe->iheight);
  const float proc_wdht_min = MIN(devpipe->processed_width, devpipe->processed_height);
  const float pr_d = darktable.develop->preview_downsampling;
  const float scale = devpipe->iscale / (pr_d * get_zoom_scale(module->dev));

  const float im_scale = 150.0f * iwd_min * scale / proc_wdht_min;
  
  *radius = dt_conf_get_sanitize_float(CONF_RADIUS, 0.5f * im_scale, 2.0f * im_scale, im_scale);
```
The "sticky control" issue doesn't seem to be such an issue now... except for the feathering, for which I don't understand the controls... need to go back to Bruce's video.

However, it's not stable against screen-size changes, ie viewing dt in partial-screen mode, and presumably it will fail on a HiDPI screen.

The scary part is that everything is different in liquify, so far as I can see. So at the moment I'd have no idea where to start. Suggestions would be welcome.

GrahamByrnes:
In fact the current config could be "good enough", in the sense that the stamps can be chosen to be controllable on any reasonable size central image, ie not less than about 2/3 of full screen on a 2K monitor, and should still be ok on a 4K at full screen.

It would be nice however, to fill in the missing image-screen-size factor.

GrahamByrnes:
I think this might do. The stamp size is now very consistent across different file formats (jpg, tiff, raw), perfectly consistent with changes of preview_downsampling, and usably consistent with window re-sizing.
The last required using a gtk function to obtain the current window dimensions: it seems these aren't reliably written back into darktablerc until dt is closed, but getting fresh ones works.
I also fixed the start-up radius to be exactly the calculated size, since if it's too big, the control points are unavailable; if too small, the control points get sticky. The control (strength) arrow is now also bounded away from the default position of the radius control, ie in the range 0.1<phi/pi<1.9

Fin bref, it seems to work. @TurboGit, when you have time?

TurboGit:
@GrahamByrnes ; Something not working. The size of the stamp and the force vector is not keep.

- Create a stamp
- Adjust the size of the strength
- Create a new stamp -> the new stamp should start with latest size and strength.

GrahamByrnes:
@turbogit That was a choice, to limit the range of saved diameters. Probably an excessive one, in retrospect: I didn't allow any range at all.
I've just re-set to allow sizes from 0.5 -> 2.0 times the calculated value... I think there needs to be an upper limit, since if the control points end up off screen, it can be confusing, or impossible to get them back. And there is the sticky control issue if the stamp is very small. Let me know if you think the range should be smaller or larger or feel free to change it (only content of the last commit).

TurboGit:
Hum, still does not work on my side. Also the strength vector is always reset on the same direction.

GrahamByrnes:
Next try:
a) relaxed constraints to clamp to ( 1/3 calculated radius , 3 * calculated radius)
b) user frustration reduction: if the stored value is outside the above range, the new stored value is 0.75 * requested value + 0.25 clamp value. So if the user asks for a really big change, most of it will be stored. That should ensure that the new stored value won't fall off the sides of the image.
c) strength direction: the same changes should also help with this. Interesting to realize that the "3 o'clock default" for the radius control point, and also the angle that the strength direction avoids, becomes the "12 o'clock default" with vertical images.

TurboGit:
Ok, better now. There is still the orientation of the vector which is not saved. I'll try to fix it on my side, a bit of coding will be nice :) Thanks.

GrahamByrnes:
Great. I'll put up a bunch of style changes I made (suppression of spaces) and you can decide to keep them or not...

TurboGit:
The issue was simple, phi (angle) do not need any tweaking for down-sampling.

PR #5156, opened 2020-05-22
Add Lua control of lighttable image visibility

### LINKED ISSUE #3831: Lua: need control over lighttable position
Currently there is (undocumented) `darktable.gui.libs.lighttable_mode.zoom_level()` that can be used to read/change zoom level, but nothing to move up-down, navigate.

This would be useful to implement go-to functionality, e,g, go to next tagged, labelled, etc photo in the current collection. So ideally there would be an api like `scroll_to(image)` or `ensure_visible(image)`.

There is api to read hovered image or change selection, but it doesn't affect the current scolling position, so cannot be used for navigation.

wpferguson:
Added lua function darktable.gui.views.lighttable.is_image_visible to check if an image is visible in lighttable view.  Added lua function darktable.gui.views.lighttable.set_image_visible to make an image
visible in the lighttable view.  Updated Release Notes and Lua API manual.

Fixes #3831.

### DIFF
--- a/src/dtgtk/thumbtable.c
+++ b/src/dtgtk/thumbtable.c
@@ -2176,6 +2176,64 @@ gboolean dt_thumbtable_ensure_imgid_visibility(dt_thumbtable_t *table, const int
   return FALSE;
 }
 
+static gboolean _filemanager_check_rowid_visibility(dt_thumbtable_t *table, const int rowid)
+{
+  if(rowid < 1) return FALSE;
+  if(!table->list || g_list_length(table->list) == 0) return FALSE;
+  // get first and last fully visible thumbnails
+  dt_thumbnail_t *first = (dt_thumbnail_t *)g_list_first(table->list)->data;
+  const int pos = MIN(g_list_length(table->list) - 1, table->thumbs_per_row * (table->rows - 1) - 1);
+  dt_thumbnail_t *last = (dt_thumbnail_t *)g_list_nth_data(table->list, pos);
+
+  if(first->rowid <= rowid && last->rowid >= rowid) return TRUE;
+  return FALSE;
+}
+static gboolean _zoomable_check_rowid_visibility(dt_thumbtable_t *table, const int rowid)
+{
+  if(rowid < 1) return FALSE;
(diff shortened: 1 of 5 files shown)


src/dtgtk/thumbtable.c
```diff
@@ -2176,6 +2176,64 @@ gboolean dt_thumbtable_ensure_imgid_visibility(dt_thumbtable_t *table, const int
   return FALSE;
 }
 
+static gboolean _filemanager_check_rowid_visibility(dt_thumbtable_t *table, const int rowid)
+{
+  if(rowid < 1) return FALSE;
+  if(!table->list || g_list_length(table->list) == 0) return FALSE;
```
AlicVB, line 2182: maybe you need to check the whole table->widget visibility too. Otherwise it may return TRUE in darkroom, ...
(same for zoomable)

wpferguson: It's wired into the Lua API at darktable.gui.views.lighttable.is_image_visible().  I'll do some further testing and see if I what happens if I try and call it while I'm in darkroom view.

src/dtgtk/thumbtable.c
```diff
@@ -2176,6 +2176,64 @@ gboolean dt_thumbtable_ensure_imgid_visibility(dt_thumbtable_t *table, const int
...
+}
+gboolean dt_thumbtable_check_imgid_visibility(dt_thumbtable_t *table, const int imgid)
+{
+  if(imgid < 1) return FALSE;
+  if(table->mode == DT_THUMBTABLE_MODE_FILEMANAGER)
+    return _filemanager_check_rowid_visibility(table, _thumb_get_rowid(imgid));
+  else if(table->mode == DT_THUMBTABLE_MODE_ZOOM)
+    return _zoomable_check_rowid_visibility(table, _thumb_get_rowid(imgid));
+
+  return FALSE;
```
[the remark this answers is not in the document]
wpferguson: This is only intended to be used in lighttable view, and then only in file manager or zoom table mode.

Issue #5088, opened 2020-05-16
[FR] Add same metadata import options on import camera/card dialog


Nilvus:
**Is your feature request related to a problem? Please describe.**
This feature is from a user of french darktable forums. 
We have now some more choices on import metadata options in import folder one or image one. Those new options are not in import from card or camera one.

**Describe the solution you'd like**
Just add same options on all import windows and so on camera/card one, with same exact possibility: all lines and possibility to check/uncheck lines.

@phweyland: this one probably for you.

phweyland:
> Just add same options on all import windows and so on camera/card one, with same exact possibility: all lines and possibility to check/uncheck lines.
> 
> @phweyland: this one probably for you.

Do we get xmp metadata directly from camera ? What about iptc ?
Of course some exif (and maybe some iptc) which end up into xmp.

I'll look at it.

Nilvus:
> > Just add same options on all import windows and so on camera/card one, with same exact possibility: all lines and possibility to check/uncheck lines.
> > @phweyland: this one probably for you.
> 
> Do we get xmp metadata directly from camera ? What about iptc ?

For that no, I just post what this issue ask for that but indeed didn't even think of that simple thing...
I will precise that to him.

> Of course some exif (and maybe some iptc) which end up into xmp.
> 
> I'll look at it.

Nilvus:
Just an add: maybe it could anyway possible to add metadata like author. And I think this user would like to have the ability to add on import title and notes on images. Like we can add author or rights. And that's a good idea.

phweyland:
> And I think this user would like to have the ability to add on import title and notes on images.

That related to default setting value (and not about importing info from image).
I don't know how import camera/card works. I'll do my best.

phweyland:
Here what we have on import camera:

![image](https://user-images.githubusercontent.com/23012047/82575741-3d3fa100-9b5f-11ea-96ab-bca088b02c38.png)

Here what we have on import image or folder

![image](https://user-images.githubusercontent.com/23012047/82575942-8263d300-9b5f-11ea-9103-c88f93eea1b3.png)

It is not so different. What I see for now.
- we need to align this module to generic metadata. I've missed this one.
- we can add also the check boxes, as on the second images, as some exif data are translated by default to xmp data.


> maybe it could anyway possible to add metadata like author.
This one is already there, right ?

Nilvus:
> It is not so different. What I see for now.
>
[...]

PR #5157, opened 2020-05-22
Same metadata import options on import camera/card as on import image and folder

phweyland:
Solves #5088

factorizes import metadata code and by the same way turns metadata generic for import camera/card too.

image


TurboGit:
Nice code refactoring. Thanks.

PR #5159, opened 2020-05-22
Revert initialisation of color picker min/max from #4931

dterrahe:
Fixes https://discuss.pixls.us/t/color-zones-color-picker-not-working-properly/18321/4


TurboGit:
Ok, merging!

dterrahe:
You are _fast_!! Many thanks.

@AxelG-DE Would you please check if the problems with the +colorpicker you saw yesterday are also resolved with this or if that's a separate problem?

PR #5160, opened 2020-05-22
Grayscale images should write grayscale tiff files.

jenshannoschwalm:
If we dt develop an image as a grayscale - whatever method we choose to do so -
we can decrease file size for exporting as a tiff.

As dt workflow does not have a special grayscale mode we check all pixels for a
minimum difference between the rgb values and if there is none found we assume
this to be a grayscale.

These images have only one color channel and set `TIFFTAG_PHOTOMETRIC` as
`PHOTOMETRIC_MINISBLACK` instead of `PHOTOMETRIC_RGB`.


TurboGit:
Nice ! Works for me and the grey scale TIFF is lot smaller this way. Thanks.

Issue #5162, opened 2020-05-22
no such column: export_timestamp after upgrading 3.1.0 (commit 9ab4c8a9)


rgo:
**Describe the bug**
After upgrading darktable 3.1.0 from commit 7e9029ee4 version to commit b9ab4c8a9, I can load darktable but ligthtable is empty as you can see below

![image](https://user-images.githubusercontent.com/47124/82688630-7eb47700-9c59-11ea-8cc9-da744c75ceb8.png)

Checking the logs you can find this:

```
sqlite3 error: /usr/src/packages/BUILD/src/common/collection.c:995, function _dt_collection_compute_count(), query "SELECT COUNT(DISTINCT mi.id) FROM (SELECT  id, group_id, film_id, filename, datetime_taken,   flags, version,  position, aspect_ratio  FROM main.images AS mi  ) AS mi WHERE   (flags & 256) != 256  AND  (1=1 AND (film_id IN (SELECT id FROM main.film_rolls WHERE folder LIKE '/home/leptom/Pictures/2019/2019-01-06 Regalos Reyes'))) ORDER BY export_timestamp, print_timestamp , filename, version LIMIT ?1, ?2": no such column: export_timestamp
sqlite3 error: /usr/src/packages/BUILD/src/common/collection.c:999, function _dt_collection_compute_count(): no such column: export_timestamp
sqlite3 error: /usr/src/packages/BUILD/src/common/collection.c:1000, function _dt_collection_compute_count(): no such column: export_timestamp
sqlite3 error: /usr/src/packages/BUILD/src/common/collection.c:995, function _dt_collection_compute_count(), query "SELECT COUNT(DISTINCT mi.id) FROM (SELECT  id, group_id, film_id, filename, datetime_taken,   flags, version,  position, aspect_ratio  FROM main.images AS mi  ) AS mi WHERE   (flags & 256) != 256  AND  (1=1 AND (film_id IN (SELECT id FROM main.film_rolls WHERE folder LIKE '/home/leptom/Pictures/2019/2019-01-06 Regalos Reyes'))) ORDER BY export_timestamp, print_timestamp , filename, version LIMIT ?1, ?2": no such column: export_timestamp
sqlite3 error: /usr/src/packages/BUILD/src/common/collection.c:999, function _dt_collection_compute_count(): no such column: export_timestamp
sqlite3 error: /usr/src/packages/BUILD/src/common/collection.c:1000, function _dt_collection_compute_count(): no such column: export_timestamp
[...]

PR #5165, opened 2020-05-22
fix collect issue - images data missing entries

phweyland:
related to #5162
- fixes the collect issues (missing images info) due to #5134 
- doesn't solve unused codes about timestamps.


TurboGit:
Works ! Thanks a lot for this fix.

PR #5168, opened 2020-05-23
Fix lua entry widget password field

wpferguson:
When `is_password` is set to true, the entry widget should act as a password field, obscuring the text.  The value supplied was being applied to `gtk_set_entry_visibility` unchanged, so the text was visible when `is_password` was true and obscured when `is_password` was false.  Negated the boolean value when it was applied to fix it.  Updated the RELEASE_NOTES.md with the fix.


TurboGit:
Looks good to me ! Thanks.

PR #5169, opened 2020-05-23
Non-local means denoising speedup

ralfbrown:
As previously mentioned in #5111, I've been rewriting the non-local means denoising code.  Here is the result, with both scalar and SSE implementations for both the Denoise (non-local means) and Denoise (profiled) iops.

Per-thread memory use the same as before, though I have added a couple of KB of housekeeping data.  Speed is better single-threaded, and much better with large numbers of threads.  Results **will** differ slightly from the previous CPU implementation, but should be closer to the OpenCL code, as the interval between fully recomputing the patch statistics is shorter, allowing less of a buildup of rounding errors.

I have not removed the old implementation yet, merely made it conditionally compilable.  Change the #define USE_NEW_IMPL at the top of src/iop/nlmeans.c or src/iop/denoiseprofile.c, respectively (it is set to the new implementation in this PR).

Timings with the default parameters on the mire1.cr2 image from the integration test suite:
Denoise(profiled)
Threads / old time / new time
1 / 25.735 / 22.532
8 / 3.483 / 3.055
16 / 1.889 / 1.595
24 / 1.556 / 1.118
32 / 1.641 / 0.856

Denoise(non-local means)
Threads / old time / new time
1 / 7.516 / 6.194
8 / 1.435 / 0.854
16 / 1.395 / 0.435
24 / 1.438 / 0.371
32 / 1.437 / 0.238

Timings using the image and .xmp from the discuss.pixls.us thread "darktable speed (in general, and when using two monitors)"
Denoise (non-local means)
Threads / old time / new time
1 / 14.346 / 11.655
4 / 4.174 / 3.124
8 / 2.783 / 1.589
16 / 2.735 / 0.853
24 / 2.806 / 0.608
28 / 2.853 / 0.605
32 / 2.862 / 0.897

Finally, sorry about the duplication of commits -- git refused to push the branch due to "an outdated HEAD", and when I tried its suggestion of pulling from the repo, it gave me a merge of two identical complete branches....


src/iop/denoiseprofile.c
```diff
@@ -40,6 +41,9 @@
 #include <xmmintrin.h>
 #endif
 
+// which version of the non-local means code should be used?  0=old (this file), 1=new (src/common/nlmeans_core.c)
+#define USE_NEW_IMPL 1
```
coffeeandspaceships, line 45: If USE_NEW_IMPL is defined and #HAVE_OPENCL is defined, this creates an introspection error with a clang build, because all 6 calls to the sign() function have been removed from the code path.

If the NEW_IMPLementation is intended to work in a build without OpenCL, this needs a fix... if it's NOT intended to work without OpenCL, then it's probably good news for the build to fail, but maybe there's a way to make it fail gracefully and indicate the root issue?

I put some thought to a fix but I am not sure what is best to defend against the case of someone adding another feature later which calls sign() after all! Is there a pragma that can make the introspection failure into just a warning?

ralfbrown:
USE_NEW_IMPL is a temporary measure until the new code  is validated (it makes it much easier to switch back and forth), so it's fine to wrap sign() in appropriate #ifdefs.   The core loops in the OpenCL paths from both denoiseprofile.c and nlmeans.c should be moved into nlmeans_core.c anyway so that they can share code, both with each other and with the CPU paths (I have a function which implements the formula that uses sign()).  I didn't want to touch that because I don't have OpenCL even though I do compile with it included.

src/iop/denoiseprofile.c
coffeeandspaceships: > I didn't want to touch that because I don't have OpenCL even though I do compile with it included.

You will be pleased to hear that meanwhile I built this branch with clang and OpenCL on Windows (!) and gave this a good go with a dozen noisy photos (Pentax, motorsport, and heavy rain are not a good combination for noise :-) ) with no problems.

TurboGit:
> Finally, sorry about the duplication of commits -- git refused to push the branch due to "an outdated HEAD", and when I tried its suggestion of pulling from the repo, it gave me a merge of two identical complete branches....

I suppose you need to clean up the branch on your side (rebase) and force push (-f) on your repository.

ralfbrown:
Actually, I had done a rebase just before the push which failed.  (git checkout master; git pull upstream master; git checkout nlmeans_speedup; git rebase master; git push)  So, not being a git expert, I'm open to a suggested sequence of commands to effect a cleanup.

ralfbrown:
No, that function was already present.  I guess the error on sign() must have masked the issue.  I've now done a full rebuild with --disable-opencl, so there shouldn't be any further such surprises.

johnny-bit:
hey @ralfbrown - in order to cleanup your branch do this:

```
git fetch -p origin
git fetch upstream
git checkout master
git merge upstream/master
git push 
git checkout nlmeans_speedup
git rebase master
git push --force
```
that last `--force` is needed because it rewrites history.

ralfbrown:
Well, that doesn't seem to have helped....
Should I just kill this branch and PR and start over with one giant commit?

coffeeandspaceships:
Mm using the branch as the commitish in the rebase doesn't seem to help. It should but it doesn't...

BUT if you want to remove the duplicates, you CAN use git rebase -i <oldest commitish you want to review from>

git rebase -i ea6415501fc28bd0db0d1675cab535f8b8120509 should give you the right list and some instructions at the bottom of the file you are given to edit.

On my downstream version of your fork that generated a merge conflict in nlmeans.c... so this plan will only work if you are used to sorting out git merges I guess.

Otherwise, no need for a giant commit in a new branch. Just created a new branch and cherry-pick the commits you want from the old one. You could even squash some of the ones where you had troubleshooting or iteration, like the "third time lucky."

ralfbrown:
Went with three big commits.   See #5192.

PR #5179, opened 2020-05-23
culling overlays

AlicVB:
This allow to use any existing overlays in culling and preview.
overlays are saved separately for culling and preview.

For the interface, the "star" menu has now 2 parts : 1 for thumbnails and 1 for culling or preview. The visibility of the different parts is set on the fly, depending of what is shown in current view.

Of course, to change overlays for preview (or culling), you'll have to first show the top panel with the global toolbox first.

### DIFF
--- a/data/darktableconfig.xml.in
+++ b/data/darktableconfig.xml.in
@@ -686,6 +686,20 @@
     <longdescription/>
   </dtconfig>
   <dtconfig>
+    <name>plugins/lighttable/overlays/culling/0</name>
+    <type>int</type>
+    <default>6</default>
+    <shortdescription>overlays for culling = block</shortdescription>
+    <longdescription/>
+  </dtconfig>
+  <dtconfig>
+    <name>plugins/lighttable/overlays/culling/1</name>
+    <type>int</type>
+    <default>6</default>
+    <shortdescription>overlays for full preview = block</shortdescription>
+    <longdescription/>
+  </dtconfig>
+  <dtconfig>
     <name>plugins/lighttable/timeline/last_zoom</name>
     <type>int</type>
     <default>0</default>
--- a/src/dtgtk/culling.c
+++ b/src/dtgtk/culling.c
@@ -753,6 +753,28 @@ static void _dt_filmstrip_change(gpointer instance, int imgid, gpointer user_dat
   _thumbs_refocus(table);
 }
 
+// get the class name associated with the overlays mode
+static gchar *_thumbs_get_overlays_class(dt_thumbnail_overlay_t over)
+{
+  switch(over)
+  {
+    case DT_THUMBNAIL_OVERLAYS_NONE:
+      return dt_util_dstrcat(NULL, "dt_overlays_none");
+    case DT_THUMBNAIL_OVERLAYS_HOVER_EXTENDED:
+      return dt_util_dstrcat(NULL, "dt_overlays_hover_extended");
+    case DT_THUMBNAIL_OVERLAYS_ALWAYS_NORMAL:
+      return dt_util_dstrcat(NULL, "dt_overlays_always");
+    case DT_THUMBNAIL_OVERLAYS_ALWAYS_EXTENDED:
+      return dt_util_dstrcat(NULL, "dt_overlays_always_extended");
+    case DT_THUMBNAIL_OVERLAYS_MIXED:
+      return dt_util_dstrcat(NULL, "dt_overlays_mixed");
+    case DT_THUMBNAIL_OVERLAYS_HOVER_BLOCK:
+      return dt_util_dstrcat(NULL, "dt_overlays_hover_block");
+    default:
+      return dt_util_dstrcat(NULL, "dt_overlays_hover");
+  }
+}
+
 dt_culling_t *dt_culling_new(dt_culling_mode_t mode)
 {
   dt_culling_t *table = (dt_culling_t *)calloc(1, sizeof(dt_culling_t));
@@ -770,7 +792,12 @@ dt_culling_t *dt_culling_new(dt_culling_mode_t mode)
     gtk_style_context_add_class(context, "dt_preview");
   else
     gtk_style_context_add_class(context, "dt_culling");
-  gtk_style_context_add_class(context, "dt_overlays_hover_block");
+
+  // overlays
+  gchar *otxt = dt_util_dstrcat(NULL, "plugins/lighttable/overlays/culling/%d", table->mode);
+  table->overlays = dt_conf_get_int(otxt);
+  g_free(otxt);
+  gtk_style_context_add_class(context, _thumbs_get_overlays_class(table->overlays));
 
   // set widget signals
   gtk_widget_set_events(table->widget, GDK_EXPOSURE_MASK | GDK_POINTER_MOTION_MASK | GDK_POINTER_MOTION_HINT_MASK
@@ -1075,7 +1102,7 @@ static gboolean _thumbs_recreate_list_at(dt_culling_t *table, const int offset)
     else
     {
       // we create a completly new thumb
-      dt_thumbnail_t *thumb = dt_thumbnail_new(10, 10, nid, nrow, DT_THUMBNAIL_OVERLAYS_HOVER_BLOCK, TRUE);
+      dt_thumbnail_t *thumb = dt_thumbnail_new(10, 10, nid, nrow, table->overlays, TRUE);
       thumb->display_focus = table->focus;
       thumb->sel_mode = DT_THUMBNAIL_SEL_MODE_DISABLED;
       double aspect_ratio = sqlite3_column_double(stmt, 2);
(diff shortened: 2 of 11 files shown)


Nilvus:
I really think it's a bad idea that way. Most of the modes are not suitable for culling and previews modes. Preview and culling mode do not display thumbnails but images. The worse is really to have group/altered... icons on top. It's really not suitable on those modes.
We need to set not all modes but think about which modes and how they could be adjusted on culling/preview ones. I really think that 2 or 3 specific (including the actual block infos one) it's the good way and, at least the actual block infos ones and a permanent one. 

If you test that with images from different ratios, it's actually even worse. Of course, here I don't talk about what could be tweak on CSS.

AlicVB:
oh no... as soon as I find an "easy" solution, you want me to do more work ;)


> Most of the modes are not suitable for culling and previews modes. Preview and culling mode do not display thumbnails but images

Completely agree. That's why hover_block is the default, because it has been designed especially for those mode ! I would even say (as you pointed) that it's the only mode really adapted to those modes.

That said, there's people with different needs (@jpverrue in our case) and that's perfectly fine ! That's why I prefer to propose as much flexibility as I can here. Maybe (certainly) some modes are not adapted for culling/preview. But apart from a bad look (imo) I don't see any drawback in letting user choose them. That's the same if one try to use hover_block in filemanager... Or if one try to use always extended for filmstrip.

Of course, we can develop some new overlays, ... but I'm not sure that's worse the pain : we have an already working overlay mode and some "almost working" other modes...
In fact, to avoid too much problems and infinite discussions about what should be a correct UI, I have develop "immuable" overlays, that looks similar in all situations. Depending on the size, usage, ... some are well adapted, other not, but they all works.

Another point, specifically about what ask @jpverrue  : iirc he want an "always-visible" block in the background (not interfering with the image), so an "always visible" block will not solve the problem.

Nilvus:
> oh no... as soon as I find an "easy" solution, you want me to do more work ;)

I knew you will like that! ;-)

> That said, there's people with different needs (@jpverrue in our case) and that's perfectly fine ! That's why I prefer to propose as much flexibility as I can here.

Of course, it's good to add flexibility but an adapted one.

> Maybe (certainly) some modes are not adapted for culling/preview. But apart from a bad look (imo) I don't see any drawback in letting user choose them. That's the same if one try to use hover_block in filemanager... Or if one try to use always extended for filmstrip.

A bad look is why I'm here: to chase it and try to have the best UI we can do. And some overlays on culling/preview are really bad UI. And I already see complaints/issues if we let some modes like on thumbnails views. I agree that some overlays are not suitable on filmstrip but just for space they took and when you see them on filmstrip, the seems over and completely done, just because on all other views, you have thumbnails, that just size which change.

On culling/preview, you don't have thumbnails (and it's a good thing of course for those modes) and that's make the big difference. So sorry but that needs adjustments to made. And I think some overlays are not suitable.

> Of course, we can develop some new overlays, ... but I'm not sure that's worse the pain : we have an already working overlay mode and some "almost working" other modes...

It's not really about develop some new overlays but keep only those more suitable and make some adjustments.

Actually, there's 2 overlays that don't need tweaks and that are suitable: the first one and the last one. The first one because nothing is shown except the image (really good thing to have) and the last one as you say because you develop it for those modes.

About others modes, I would see 2 choices:

- Or we keep them but we regroup top infos (group/altered/audio... icons) with other infos in bottom) or we do the opposite: put all infos on top of the image.
- Or we just use the 2 overlays I cite above (first one and block infos one) and we add options on it as @jpverrue himself propose: 

> _It would be to add to the thumbnail information a "hamburger" button that would give access to a popup menu in which one could choose the display duration (including permanent), permanent display even if the pointer is no longer on the photo, and even, let's be crazy, the position of the thumbnail (top left/right, bottom left/right), and finally the transparency rate._

> In fact, to avoid too much problems and infinite discussions about what should be a correct UI, I have develop "immuable" overlays, that looks similar in all situations. Depending on the size, usage, ... some are well adapted, other not, but they all works.

And that's a really great choice you mode, except for culling and preview are they're not thumbnails so can't be exactly displayed same way (the actual way really broke those modes) and even if some things could be improved by CSS, I see no real good way with the actual state of this PR.

> Another point, specifically about what ask @jpverrue : iirc he want an "always-visible" block in the background (not interfering with the image), so an "always visible" block will not solve the problem.

One of the 2 options I propose above will please him. Personally, I'm more for choice 2. @jpverrue: your view on that is more than welcome.

AlicVB:
As a third "cheap" (in term of code) change, what about keeping the "always normal' as it is without the top icons.
Pro : 
- show essential infos permanently outside the image
- same look a in other modes
- very small code change (in fact maybe it will just be a matter of css <= this is my revenge ;))

Con : 
- the right and left icons are aligned on the whole area and not the image itself (but fixing that would require to develop a completely new mode, as this implies to move all the icons into another container)

Nilvus:
> As a third "cheap" (in term of code) change, what about keeping the "always normal' as it is without the top icons.
> Pro :
> 
>     * show essential infos permanently outside the image
> 
>     * same look a in other modes
> 
>     * very small code change (in fact maybe it will just be a matter of css <= this is my revenge ;))
> 
> 
> Con :
> 
>     * the right and left icons are aligned on the whole area and not the image itself (but fixing that would require to develop a completely new mode, as this implies to move all the icons into another container)

It's ok for me that way. For infos remaining, I could propose a specific CSS tweak to improve display

AlicVB:
yes, I just forgot that point... I'll have a look.
I mark this PR as WIP until I'll finish this (and limit the culling overlays modes)

AlicVB:
@Nilvus : I've made the following changes : 
- limit overlays to : none, always_normal, always_extended and block
- redesign a little the bottom icons positions, so that we don't have reject cross and colorlabel to far away from the stars if the thumbnail are big (this is the case for culling/preview, but is also useful for filemanager imo)
- for the top icon, they can be hidden for culling/preview with some css.
- integrate the overlays block timeout (read&write) inside the star menu
- store timeout value per overlays size mode : you can have different timeout for culling and preview and thumbnail size 2, ...
- allow to set timeout to -1 for permanent visibility (if the cursor is on the image)

if you can share your opinion on this new draft... Thanks !

Nilvus:
@AlicVB: just tested a little by now and much better this way. I do not test actually CSS tweaks but it's good to see you adjust bottom icons positions and overlays remaining (best choices I think).

On timeout, why not simply add "during" instead of the point and next duration? What I see would be:
"overlays block on mouse hover during <spinbutton widget> (s)"

I would suggest you to add tooltip on spinbutton widget to explain what you write  about timeout (-1 for permanent visibility and possibility to set timeout when letting the block visible when hovering the image and that moving the mouse reset the duration.

Last thing: wouldn't be possible to not show duration block on lighttable and filmstrip popover menu overlays instead of letting them grey? Most users will not understand why this is grey and will probably even ask to activate it...

I will test CSS on that ASAP (but sun planned for next days and a long week-end so be patient). Maybe on evenings...

AlicVB:
@Nilvus : I trust you on the English wording ;) I'll change it soon.

> Last thing: wouldn't be possible to not show duration block on lighttable and filmstrip popover menu overlays instead of letting them grey? Most users will not understand why this is grey and will probably even ask to activate it...

Not sure to understand... You want that we hide the duration block instead of disabling it when the select overlay is not block ? That would be strange imo because we will have in this case a not finished sentence "overlays block on mouse hover during (s)" without anything after...
Or maybe you want to hide completely the block overlay for filemanager... but this can be useful for big thumbnails sizes...

AlicVB:
So I've change the block sentence as said earlier.
For the tooltip, I've set the text to : 
```
duration before the block overlay is hidden after each mouse movement on the image
set -1 to never hide the overlay
```
or
`timeout only available for block overlay`
I'm absolutely not a great English speaker, so don't hesitate to propose other wording if needed

Nilvus:
Tooltips are good for me. An issue I have on the timeout. It only appears on thumbnails overlays (no issue on culling overlays). So, on lighttable view (filemanager), choose block overlays mode and then change timeout (for example, 1s), then check that the timeout remains to default 3s. Go back to popover menu by clicking on star icon and see that timeout have been reset to 3s.

Nilvus:
@AlicVB: except last cited issue about timeout in lighttable, it would have been better to have bottom infos width aligned to image and not thumb_main, if we set background, but it would probably be tricky. If easy to do, it would be good to have that. But if quite tricky, I think I've found a way to render that good with CSS this way (hover effect):

image

Nilvus:
Forget bottom infos width idea I cited. I should have find something else to do with that.

AlicVB:
> Forget bottom infos width idea I cited. I should have find something else.

that's a good news, because it would have been tricky indeed ;)

I'll have a look at the bug with filemanager

AlicVB:
@Nilvus : I fixed the last issue you mentioned (I completely forgot it)
@TurboGit : should be ready for review, thanks :)

Nilvus:
So all good for me. Thanks for this work again.

TurboGit:
Need a little rebase and conflict resolution.

AlicVB:
> Need a little rebase and conflict resolution.

Oops, didn't notice that. That should be ok now, sorry...

TurboGit:
Works for me! Thanks.

PR #5189, opened 2020-05-24
colorin: handle (blue) LED light gracefully

rawfiner:
This is work in progress, as I have not coded the opencl code, nor the sse code (currently, the sse paths calls the same function as the standard path, but you have to disable opencl)

Blue LEDs are a nightmare with current colorin: they typically produce negative values.
The reason is that the standard color matrix have negative coefficients, and when B is much higher than G and R, this results in an output with negative G and R, that can also have a negative luminance.
While it will be possible to get back in gamut blues that did not result in negative values with @aurelienpierre 's new channel mixer, negative values are still a problem.

In this PR, I propose a new way to apply the standard color matrix in a way that guarantees the absence of negative number in the output, and that lets the user control color regularization (color desaturation) and luminance regularization (that fixes gradients that may look unnatural, typically at the borders of a zone lit by blue led).
The idea is to constitute 2 outputs, one good in terms of color, the other one good in terms of luminance, and mix both.
For each of these, the user can control a threshold, which defines a percentage p: to constitute an output for a channel, the algorithm will make the matrix product in 2 parts: a part with the positive coefficients, and a part with the negative ones, and will make sure that the negative part is not higher than p% of the positive part (with p = 100%, this is equivalent to clipping negative values).
The algorithm uses desaturation (with desaturation formula from filmic v4 with maxRGB norm) to guarantee that for the output "good in color", and uses clipping for the output "good in luminance".

To use: in colorin, switch gamut mapping to "custom".
Use color regularization to control colors (the smaller it is, the more saturated will be the blue leds)
Use luminance regularization to control luminance (increase it if there are some gradient reversals in the image)

Note that for natural images (without blue leds), the 0.7 thresholds proposed rarely affect the image, that's because for each pixel the negative part is not higher than 30% of the positive part.
For this reason, we may consider to set this on by default.

Example (raw file here: https://drive.google.com/file/d/1HddLn5PCoHvXi21H_e8rWHfIFK71i2Tq/view?usp=sharing )
Original:
Capture d’écran du 2020-05-24 17-47-08

Enabling custom regularization with default parameters:
Capture d’écran du 2020-05-24 17-47-15

Lowering color regularisation, we get more saturated blues, closer to the original:
Capture d’écran du 2020-05-24 17-47-26

Lowering luminance regularisation too much (which give luminance closer to what colorin gives by default), we get gradient issues (look at the camera at the bottom right):
Capture d’écran du 2020-05-24 17-47-49

### DIFF
--- a/src/iop/colorin.c
+++ b/src/iop/colorin.c
@@ -1112,8 +1112,13 @@ static void process_cmatrix(struct dt_iop_module_t *self, dt_dev_pixelpipe_iop_t
 {
   const dt_iop_colorin_data_t *const d = (dt_iop_colorin_data_t *)piece->data;
   const int blue_mapping = d->blue_mapping && dt_image_is_matrix_correction_supported(&piece->pipe->image);
+  const int regularized = (d->custom_regularization == 1);
 
-  if(!blue_mapping && d->nonlinearlut == 0)
+  if(regularized)
+  {
+    process_cmatrix_fastpath_regularized(self, piece, ivoid, ovoid, roi_in, roi_out);
+  }
+  else if(!blue_mapping && d->nonlinearlut == 0)
   {
     process_cmatrix_fastpath(self, piece, ivoid, ovoid, roi_in, roi_out);
   }
@@ -1415,8 +1420,14 @@ static void process_sse2_cmatrix_fastpath(struct dt_iop_module_t *self, dt_dev_p
 {
   const dt_iop_colorin_data_t *const d = (dt_iop_colorin_data_t *)piece->data;
   const int clipping = (d->nrgb != NULL);
+  const int regularized = (d->custom_regularization == 1);
 
-  if(!clipping)
+  if(regularized)
+  {
+    //TODO provide sse2 version
+    process_cmatrix_fastpath_regularized(self, piece, ivoid, ovoid, roi_in, roi_out);
+  }
+  else if(!clipping)
   {
     process_sse2_cmatrix_fastpath_simple(self, piece, ivoid, ovoid, roi_in, roi_out);
   }
@@ -1504,8 +1515,14 @@ static void process_sse2_cmatrix(struct dt_iop_module_t *self, dt_dev_pixelpipe_
 {
   const dt_iop_colorin_data_t *const d = (dt_iop_colorin_data_t *)piece->data;
   const int blue_mapping = d->blue_mapping && dt_image_is_matrix_correction_supported(&piece->pipe->image);
+  const int regularized = (d->custom_regularization == 1);
 
-  if(!blue_mapping && d->nonlinearlut == 0)
+  if(regularized)
+  {
+    //TODO provide sse2 version
+    process_cmatrix_fastpath_regularized(self, piece, ivoid, ovoid, roi_in, roi_out);
+  }
+  else if(!blue_mapping && d->nonlinearlut == 0)
   {
     process_sse2_cmatrix_fastpath(self, piece, ivoid, ovoid, roi_in, roi_out);
   }
--- a/src/iop/colorin.c
+++ b/src/iop/colorin.c
@@ -93,7 +93,6 @@ typedef struct dt_iop_colorin_params_t
   dt_iop_color_intent_t intent;
   int normalize;
   float color_regularization;
-  float luminance_regularization;
   int blue_mapping;
   // working color profile
   dt_colorspaces_color_profile_type_t type_work;
@@ -103,7 +102,7 @@ typedef struct dt_iop_colorin_params_t
 typedef struct dt_iop_colorin_gui_data_t
 {
   GtkWidget *profile_combobox, *clipping_combobox, *work_combobox;
-  GtkWidget *color_regularization, *luminance_regularization;
+  GtkWidget *color_regularization;
   GList *image_profiles;
   int n_image_profiles;
 } dt_iop_colorin_gui_data_t;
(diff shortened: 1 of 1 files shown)


aurelienpierre:
Did you plan for the case where the camera input profile is a LUT and not a simple matrice anymore ? Input LUTs have been shown to produce more consistent results for blue lights (https://discuss.pixls.us/t/the-quest-for-good-color-1-spectral-sensitivity-functions-ssfs-and-camera-profiles/18002). That might change the definition of luminance.

I'm a bit suspicious with the results of the gamut mapping, the hue seems to be shifting to purple and saturation is not really the topic of gamut mapping and I'm not sure it should get convolved in it. It's only a matter of shrinking the gamut in the direction of the white point. I will need to think more about that.

rawfiner:
No, it is only for simple matrices. It does not aim at being a perfect solution (it will surely be less good than lut), but to provide the user a way to get something usable when he has to use matrices.

I put it in gamut mapping to avoid making the GUI too complicated, but it is not exactly a gamut mapping operation.

About colors: I don't see hue shifts on all pictures, so I don't exactly know where it comes from.
Either I made a mistake when looking at the code of filmic to "copy" the saturation formula, either it does not behaves well with maxrgb as norm (though it is the only one we can use in this context: luminance and other norms may be negative).
Maybe there is a better solution than desaturation though.

I tried to do the same thing without barely any saturation change (equivalent to put color regularization to 0.01) and use only luminance regularization, and then do gamut mapping with channelmixerrgb, but it requires a huge gamut mapping which desaturates basically all the image, so not ideal either.

rawfiner:
(Which reminds me I need to make the option disapear from the GUI when not using a matrix)

aurelienpierre:
> About colors: I don't see hue shifts on all pictures, so I don't exactly know where it comes from.

It comes from the Abney effect : constant hue lines are not actually lines in the xyY space. Yet we need to preserve radiometric relationships so we need to sort-out something mildly bad in linear, knowing it doesn't fit our perceptual system. For #4940, I use uv space, derivated from xyY.

Maybe start with an euclidean norm. From the property of the dot product, it is the only norm that is preserved by the matrice product, no matter the in/out space, and it will fall close to Y in most cases.

But I don't think RGB spaces are well-suited for this. I mean, the best you can do is simply clip the negatives. For actual mapping, you need to take into account the distance between current pixel and white point, and the hue non-linearity.

rawfiner:
Oh indeed, did not remembered the Abney effect, but that is exactly the issue.

I just pushed several commits:
- in 80f5e155e759af67b5e5fcb4fe48433147665b9e , I tried using u'v' to desaturate (I am not an expert in color spaces, so I hope I did not make a stupid mistake). Though, even if colors are better, I experienced harsh transitions and it seems less reliable to noise.
- in next commits, I revert to some more simple desaturation, but not the same formula as before: I now just add white light to desaturate. Even though the Abney effect is still somewhat visible, it is much less visible than previously. And all color gradients are smooth. I also use maxRGB as a norm now, and that does not requires any luminance regularisation, so GUI is simpler. I personally think this gives a good tradeoff.

src/iop/colorin.c
```diff
@@ -875,6 +875,39 @@ static void process_cmatrix_fastpath_clipping(struct dt_iop_module_t *self, dt_d
   }
 }
 
+static void rgb_to_uvY(float rgb[3], float uvY[3])
+{
+  float X = 2.7689f * rgb[0]
```
aurelienpierre: where do the RGB -> XYZ come from ?

rawfiner: That's wikipedia coefficients for cieRGB to cieXYZ conversion

rawfiner: https://fr.wikipedia.org/wiki/CIE_XYZ#Passage_de_l'espace_CIE_XYZ_%C3%A0_l'espace_CIE_RGB

Was it not the right space to use?

aurelienpierre: But why CIE RGB ? It's either camera RGB or pipeline RGB we need, depending on where you do the mapping, or even XYZ directly. 

XYZ is obtained from camera RGB by a matrice dot product with a matrice produced by least-squares fitting RGB -> XYZ over known calibration samples (a color checker or an IT8 chart), which is essentially the Adobe DNG matrices.

The problem of this method is it is designed for accurate mapping in the central spectrum region, but doesn't care about gamut bounds, so the transvection can result in bad coordinates.

rawfiner: I agree that before colorin, we are in camera rgb, after colorin, either Lab or the RGB working space.

I am in colorin, right after the matrice product.
In colorin, we do a matrice product with camera RGB data, but even if it is written xyz everywhere in the variable names, I believe the output data is RGB, not XYZ:
if I do _xyz[1] /= 4; for all pixels, we see an effect which looks like a scaling on the green channel, not a change of luminance:
see on this image:
Capture d’écran du 2020-05-25 15-18-41

doing _xyz[1] /= 4; for all pixels right after the matrice product (see how the histogramme has not changed for R and B, but has changed for G):
Capture d’écran du 2020-05-25 15-20-59

As I could not take the RGB working space, nor the camera rgb space for the XYZ conversion, I assumed that we were in cieRGB, but maybe that's not the case

aurelienpierre: What you see is not what you get. Even in XYZ… Y is both luminance and "green" stimulus, this is anchored in perception. For practical ends, XYZ is yet another RGB space, only it is a reference one that is gluing spectral input to physiological output. But after the dot product, you should only care about XYZ.

rawfiner: Aaaah ok I understand.
Thanks for taking time to explain that :-)

rawfiner:
Well, as gamut compression from #4940 can now manage negative inputs (thanks Aurélien!), my PR is probably not needed anymore. Its main purpose was to be able to have a solution where channel mixer rgb failed. Better use the new channel mixer to do that now, it will be cleaner. :-)

PR #5191, opened 2020-05-24
Added White Balance Presets for the Fujifilm X-E3

JonHrovath:
A simple change to add the White Balance camera presets for the Fujifilm X-E3.


TurboGit:
Thanks !

PR #5194, opened 2020-05-24
Quoted folder name

jpverrue:
When renaming a folder, if the folder name contains a single quote, the renaming fails. The SQL SELECT returns 'syntax error'.  This is due to the fact that the folder name in the select is itself surrounded by two single quotes, so we need to double those single quotes in the folder name for this to work. This PR therefore produces a name with two single quote for each quote encountered in the folder name, which ensures that the SELECT will work properly.


parafin:
Real problem is that SQL statements shouldn’t be created like that ever, but there’s a lot of such code in dt. Instead prepare and parameter bind must be used. I suggest to rewrite the code in correct way.

TurboGit:
@parafin : fact is that this is almost impossible for the collection support. We have many part of the SQL statement created in different places for the main select, for the where clause depending on the actual collection (and this can include multiple sub-queries for the refinement, include, exclude parts) for the sort part depending on the current filter and for the ordering.

So here I don't see a good solution except maybe rewriting the whole collection support code to also keep track of the final bind information needed with the corresponding index. Not something I would plan as I'm not even sure this path would fulfill all the needs for collection support and if possible it will take a lot of time.

We cannot wait an hypothetical rewrite to fix this.

parafin:
I don't see how it applies to this case.

parafin:
Ah, I see the confusion. I didn't mean rewrite the whole collect code, I meant rewrite this PR.

TurboGit:
> Ah, I see the confusion. I didn't mean rewrite the whole collect code, I meant rewrite this PR.

Agreed, in this PR we can and should use a bind for the parameter.

jpverrue:
@parafin @TurboGit You're both right, and that's what I started doing last night. I wrote something like this:
```
query = "SELECT id, folder FROM main.film_rolls WHERE folder LIKE '?1%'";
DT_DEBUG_SQLITE3_PREPARE_V2(dt_database_get(darktable.db), query, -1, &stmt, NULL);
DT_DEBUG_SQLITE3_BIND_TEXT(stmt, 1, tree_path, -1, SQLITE_STATIC);
```
This code is simpler and more in line with the idea of parameterization that I was talking about. But with this, yesterday, darktable crashes. The error message displayed is 'column index out of range'. I looked for a while to understand and I found a sentence in the sqlite doc that says it's because the 'bind' must be called after a 'step' (https://www.sqlite.org/c3ref/column_blob.html) As I couldn't see how to do it without redoing the whole loop I gave up to write the little piece of code of this PR. But if someone can tell me what was bad in my first code, no problem, I'll change it very gladly, because it's actually much better.

TurboGit:
I suppose you want:

To add % into tree_path.

```
gchar *q_tree_path = NULL;
q_tree_path = dt_util_dstrcat(q_tree_path, "%s%%", tree_path);

query = "SELECT id, folder FROM main.film_rolls WHERE folder LIKE ?1";
DT_DEBUG_SQLITE3_PREPARE_V2(dt_database_get(darktable.db), query, -1, &stmt, NULL)
DT_DEBUG_SQLITE3_BIND_TEXT(stmt, 1, tree_path, -1, SQLITE_STATIC);

g_free(q_tree_path);
```

TurboGit:
> it's because the 'bind' must be called after a 'step'

Didn't check the doc but this seems very wrong and this is not something done in whole dt code, so I suppose it would have been detected before :)

jpverrue:
I thought it was a good idea (putting the '%' in the filename.). I really believed it, but no, dartktable always crashes. Here's the output:
```
q_tree_path=/home/jpv/test-darktable-new-ts/Photos/2019%
query=SELECT id, folder FROM main.film_rolls WHERE folder LIKE ?1
after1
after2
after3
munmap_chunk(): invalid pointer
(core dumped)LANG=$1 ${testDir}/bin/darktable --configdir ${testDir}/config
```
Note that the `column index out of range` error message no longer appears. This is a first progress.
And this is the code portion. I took exactly what you proposed (except the bind where you left the old variable ;-))
```
     gchar *q_tree_path = NULL;
      q_tree_path = dt_util_dstrcat(q_tree_path, "%s%%", tree_path);
/*!!*/printf("q_tree_path=%s\n", q_tree_path);
      query = "SELECT id, folder FROM main.film_rolls WHERE folder LIKE ?1";
/*!!*/printf("query=%s\n", query);
      DT_DEBUG_SQLITE3_PREPARE_V2(dt_database_get(darktable.db), query, -1, &stmt, NULL);
      DT_DEBUG_SQLITE3_BIND_TEXT(stmt, 1, q_tree_path, -1, SQLITE_STATIC);
/*!!*/printf("after1\n");
      g_free(q_tree_path);
/*!!*/printf("after2\n");
      q_tree_path = NULL;
/*!!*/printf("after3\n");
      g_free(query);
/*!!*/printf("after4\n");
      query = NULL;
/*!!*/printf("after5\n");

```
So the crash occurs when you leak the memory space of the query.
I think we're gonna have to take out the debugger. :-(

TurboGit:
Can you send me the collect.c file ? Or push this branch... I'd like to have a look, really strange indeed.

jpverrue:
I tried again by resetting the single quotes around the ?1 parameter. Are they necessary? This caused the error message to reappear, but it doesn't prevent darktable from crashing again.
The output : 
```
q_tree_path=/home/jpv/test-darktable-new-ts/Photos/2019%
query=SELECT id, folder FROM main.film_rolls WHERE folder LIKE '?1'
12,303221 [sql] /home/jpv/dev-c/darktable-dev/src/libs/collect.c:385, function view_popup_menu_onSearchFilmroll(): prepare "SELECT id, folder FROM main.film_rolls WHERE folder LIKE '?1'"
after0
sqlite3 error: /home/jpv/dev-c/darktable-dev/src/libs/collect.c:387, function view_popup_menu_onSearchFilmroll(): column index out of range
after1
after2
after3
munmap_chunk(): invalid pointer
30670 Abandon                 (core dumped)LANG=$1 ${testDir}/bin/darktable -d sql --configdir ${testDir}/config
```
between after0 and after1, this is the macro BIND_TEXT

TurboGit:
Maybe you need to replace SQLITE_STATIC by SQLITE_TRANSIENT ?

jpverrue:
Pushed...

jpverrue:
Exactly the same result with SQLITE_TRANSIENT as with SQLITE_STATIC.
At the same time, a rereading of the definition of the two macros shows that the choice of STATIC is the right one.
(https://www.sqlite.org/c3ref/c_static.html)

src/libs/collect.c
```diff
@@ -377,10 +377,23 @@ static void view_popup_menu_onSearchFilmroll(GtkWidget *menuitem, gpointer userd
...
+/*!!*/printf("query=%s\n", query);
       DT_DEBUG_SQLITE3_PREPARE_V2(dt_database_get(darktable.db), query, -1, &stmt, NULL);
+/*!!*/printf("after0\n");
+      DT_DEBUG_SQLITE3_BIND_TEXT(stmt, 1, q_tree_path, -1, SQLITE_STATIC);
+/*!!*/printf("after1\n");
+      g_free(q_tree_path);
+/*!!*/printf("after2\n");
+      q_tree_path = NULL;
+/*!!*/printf("after3\n");
       g_free(query);
```
TurboGit, line 393: you free query which is a static string :)

jpverrue:
Yesss !
Sorry about the free(). It was there before and was justified since there was memory allocation by dt_util_dstrcat(). I forgot to remove it when I changed the call to a static string assignment. Grrr!
I also had to change the SQLITE_STATIC to SQLITE_TRANSIENT, since I free q_tree_path right after the bind.

I just pushed a functional version.

TurboGit:
Works for me now. Thanks.

PR #5196, opened 2020-05-25
Added checks to ensure that the view is lighttable and the mode

### LINKED ISSUE #3831: Lua: need control over lighttable position

wpferguson:
Added lua function darktable.gui.views.lighttable.is_image_visible to check if an image is visible in lighttable view. Added lua function darktable.gui.views.lighttable.set_image_visible to make an image visible in the lighttable view. Updated Release Notes and Lua API manual.

Fixes #3831.

Was #5156.

Had to resubmit due to me screwing up my tree and having to delete the original branch.

@AlicVB  added checks to ensure that the view is lighttable and the mode is either file manager or zoomable.

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -296,6 +296,13 @@ making a backup is strongly advised.
 - fixed is_password field of entry widget to work according to the API manual, so 
   now when it is set to true the field is hidden.
 
+- Added function darktable.gui.views.lighttable.is_image_visible to check if an image 
+  is visible in lighttable view.
+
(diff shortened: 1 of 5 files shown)


TurboGit:
Thanks !

PR #5199, opened 2020-05-25
don't allow passing NULL to g_hash_table_lookup in history.c

### LINKED ISSUE #5190: Segfault while exporting
**Describe the bug**

After last rebuild of git HEAD Darktable crashes for me very often. It seems to happen after I press ^e to export image while in darkroom view. The file which I was processing doesn't get exported at all. It happens every 3-4 photos. Attaching backtrace log.

**To Reproduce**
1. Open a few images
2. Switch to darkroom mode
3. Press ^e to export

 - Darktable Version: 593f29a92b05d19284e517b4e7035e51b3a9b6b6 or 1deb21385f0ee97c57af37b9b4586d287c095658
 - OS: Gentoo Linux
 - OpenCL activated
 - AMD Fury + Mesa 20.1.0_rc3 + ROCm 3.3.0

**Additional context**
 - Can you reproduce with another Darktable version: Tried two recent git versions.
 - Can you reproduce with a RAW or Jpeg or both: Tried only on RAW
 - Are the steps above reproduce with a fresh edit (removing history): It happens on freshly imported images
 - Did you compile Darktable yourself? Yes, gcc-9.3.0
 - Build log: [darktable_bt_OJQ8K0.txt.gz](https://github.com/darktable-org/darktable/files/4674062/darktable_bt_OJQ8K0.txt.gz)

johnny-bit:
In very rare cases, `operation` might be null. In those cases,
darktable would bail since `g_str_hash` doesn't like null keys.

fixes #5190

I wondered about forcing `NOT NULL` constraint on `operation` column in `history` table, however `dt_dev_write_history_item` actualy inserts null there (and tries to immediatelly update it)

### DIFF
--- a/src/develop/imageop.c
+++ b/src/develop/imageop.c
@@ -2281,8 +2281,13 @@ gchar *dt_iop_get_localized_name(const gchar *op)
       } while((iop = g_list_next(iop)) != NULL);
     }
   }
-
-  return (gchar *)g_hash_table_lookup(module_names, op);
+  if(op != NULL)
+  {
+    return (gchar *)g_hash_table_lookup(module_names, op);
+  }
+  else {
+    return _("ERROR");
+  }
 }
 
 void dt_iop_so_gui_set_state(dt_iop_module_so_t *module, dt_iop_module_state_t state)


TurboGit:
Works for me.

PR #5201, opened 2020-05-25
Adding introspection data and refactoring gui for IOPs

dterrahe:
I'd like to get feedback on the progress I've made on #4804 while waiting for @houz to finalize his work on introspection for enums in perl.

This includes the infrastructure changes that I separately submitted in #5097, but while waiting for those to be accepted I've been working my way through the first 18 modules. It is probably easier in first instance to see the impact looking at the modules directly rather than at the infrastructure files.

I have split this draft PR into separate commits for each file, so everybody can have a look at the impact on their favorite module, but also so that it would be easier to revert individual modules if problems arise after merging. For the final submission I would be happy to split out each file in a separate PR or in any grouping of modules that would be most convenient.

### DIFF
--- a/src/bauhaus/bauhaus.c
+++ b/src/bauhaus/bauhaus.c
@@ -2218,6 +2218,19 @@ float dt_bauhaus_slider_get_step(GtkWidget *widget)
   return d->step;
 }
 
+void dt_bauhaus_slider_set_feedback(GtkWidget *widget, int feedback)
+{
+  dt_bauhaus_widget_t *w = (dt_bauhaus_widget_t *)DT_BAUHAUS_WIDGET(widget);
+
+  if(w->type != DT_BAUHAUS_SLIDER) return;
+
+  dt_bauhaus_slider_data_t *d = &w->data.slider;
+
+  d->fill_feedback = feedback;
+
+  gtk_widget_queue_draw(widget);
+}
+
 void dt_bauhaus_slider_reset(GtkWidget *widget)
 {
   dt_bauhaus_widget_t *w = (dt_bauhaus_widget_t *)DT_BAUHAUS_WIDGET(widget);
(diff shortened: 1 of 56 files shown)


johnny-bit:
When you have time, can you look @ #4360 ? I've added there 3 useful (imo) functions in bauhaus :) whould I need to change anything in bauhaus on my side if yours is merged first? The changes to bauhaus you've done are quite substantial.

dterrahe:
I think dt_bauhaus_slider_get_default will indeed need to be changed; since #5113 the defpos value gets stored pre-transformation (because otherwise it gets messed up by any change to soft min/max). So I believe you should not call "callback" on it anymore.

I was thinking of adding set_feedback as well, because I don't want to add separate slider_new_from_params calls for feedback yes/no. This would then be used in the colorize module.
Don't know if get_feedback is ever useful...

As an aside, I've changed the name of set_callback (which i didn't think was very descriptive since there are all kinds of callbacks all over the place) to set_curve.

More importantly, the curve transformation is now directly applied at the slider level instead of at the input level. This allows reversed transformations, where otherwise min and max would have to be swapped. Also, it means changing the curve/callback doesn't mess up all settings (soft/hard min/max default) that had been set up before. And this is important in the context of my wider project, because some of those other values are now set up from introspection data, i.e. they have to remain constant/can't/shouldn't be fixed up later.

dterrahe:
This is now ready for review, based on earlier discussions in #4804 and #5097, which I'll close.

As a reminder, the goals of this PR (and the next one for the remaining modules if they are not all merged at the same time) is to provide complete introspection data for parameter min/max ranges and defaults and to ensure those are consistent with the values enforced in the gui, while at the same time improving code quality and readability and reducing code size by removing boilerplate.

This PR includes both the required infrastructure (with some additional changes since #5097) and the first batch of modules, up to filmicrgb. I've split the commits per module, with my later rename and init changes committed as fixups. I'm going to squash those during final rebase, to one per module, but wanted to keep the additional detail for now (in case I introduced errors in the last round, which quickly touched all modules again).

Realistically, I think it will be hard to review this on a change-by-change basis; as always with refactorings, because things move all over the place. I would suggest looking at the new code and testing old and new versions side-by-side for each module to see if there are any noticeable changes in behaviour (there shouldn't be any apart from minor bugfixes) or default values.

If it would be easier to split this into multiple PRs I'm happy to do that (for example to separate the larger more complex modules). Otherwise individual modules could be cherry-picked from this one as and when they get reviewed/signed off. Please let me know what works best from experience.

PR #5203, opened 2020-05-25
Add integration test for 'denoise (profiled)' using wavelets in Y0U0V0

ralfbrown:
expected.png generated using git master as of a few minutes ago.  This PR permits regression checking of a speedup for which I will be opening a PR soon.


ralfbrown:
Note that #5204 may moot the original image in this PR, as the existing code is numerically unstable (demonstration code in that PR).

TurboGit:
This is a bit worrisome if my test framework is not stable across computers:

```
Test 0014-denoiseprofile-wavelets
      Image mire1.cr2
      CPU & GPU version differs
      Max  dE         7.3851
      Mean dE         0.0289
      Count above max 1473
  FAILS: image visually changed
```

The number of diffs is not huge (1473 pixels) but yet the check is only done for CPU version of the code.

TurboGit:
One question, you have generated it using the darktable-cli as said in the readme ? This is important as the script avoids any tiling.

TurboGit:
The diff I have on my side (does not look like a tiling issue at all):

diff

ralfbrown:
Yeah, that doesn't look good.  I generated the expected.png exactly as described in the README, by running run.sh and then renaming output.png.  How many of those pixels differ by more than the minimum representable change, though?  They seem to be concentrated in the darker pixels, where a difference of one digital unit (e.g. value 11 vs 12) makes for a much larger delta-E.

I just rebuilt from master to verify that I didn't screw up before (e.g. I may have left out the system-install step and pulled in a modified plugin .so) and used run.sh to generate a new output.png.  "compare -metric ae -verbose" says there are no differing pixels between what's in the PR and what I just generated, though the actual files themselves differ marginally in size.

One known portability difficulty your script will face is that some of the iops (such as nlmeans) will inherently generate different results when run with different numbers of threads due to different splits putting accumulated rounding errors in different places.  For tests of those iops, you'll want to fix the number of threads to ensure reproducibility (I would suggest 4 threads, as nearly everybody nowaways should have a machine capable of that many).

TurboGit:
@ralfbrown : Good idea, I'll look at fixing the number of threads.

TurboGit:
BTW, how many threads used on your side ?

ralfbrown:
Since your script isn't limiting the number, all 64.  Wavelets isn't one of the thread-dependent iops, though.  I just re-ran with four threads and get pixel-identical output.

But I had the thought of "is there anything in the processing chain that relies on system libraries which might behave differently from system to system?"  LCMS2 came to mind, and turning the darktablerc export/force_lcms2 setting from FALSE to TRUE changed the output!  The diff.png looks just like what you posted above.

aurelienpierre:
> LCMS2 came to mind, and turning the darktablerc export/force_lcms2 setting from FALSE to TRUE changed the output! The diff.png looks just like what you posted above.

Of course. LCMS2 does gamut mapping at the output, while the internal dt method does only gamut clipping. 

I would recommend to output directly in Lab (it's an hidden option somewhere) and as float32 (pfm or exr) before running any kind of comparison.

TurboGit:
@aurelienpierre :  I'm not sure we want to use Lab here. We really want to see what is exported (that's the whole point of the integration tests) as standard image. Of course LCMS does change the output I just forgot that. i had already fixed the memory to ensure that no tiling happens, we will fix the number of threads to avoid any accumulated rounded errors. Finally let's force LCMS for the tests, we will be able to check that the different intents settings are indeed properly set by dt.

TurboGit:
@ralfbrown : I tried to run without lcms2 on my side but the test still fails. Can you push the new output.png on your side ? I'll test if it pass on mine. Thanks.

aurelienpierre:
But if you compute a delta E, anyway, you pass through Lab. If your output is Lab already, the delta E is directly `sqrt( (L_ref - L_test)^2 + (a_ref - a_test)^2 + (b_ref - b_test)^2)`, and you avoid any extra step to tamper with it.

TurboGit:
@aurelienpierre : Sure but what the point to check that dt is consistent if a bug is introduced in lcms2 and give different output ? Testing the image through lcms2 seems a good point to me as this is the only way to get good color rendering. No ?

TurboGit:
Note also that I do not compute Delta-E I let OpenCV do that for me :) And it probably does not support Lab output.

aurelienpierre:
Ok then, if LCMS2 is part of the test coverage, but it might make bug spotting more difficult. Upstream stuff should better be tested upstream or separately.

ralfbrown:
Do you want me to push an expected.png generated with LCMS turned **on**?  Because the new one with it still turned off is identical to what's already in this PR.

TurboGit:
@ralfbrown : no no need if it is the same... but you wrote:

> LCMS2 came to mind, and turning the darktablerc export/force_lcms2 setting from FALSE to TRUE changed the output! The diff.png looks just like what you posted above.

So I'm confused now!

ralfbrown:
@TurboGit 
I recompiled with git master, ran run.sh 0014-denoiseprofile-wavelet, and got an output identical to what's in this PR.  Then I went and edited my ~/.config/darktable/darktablerc to change
`plugins/lighttable/export/force_lcms2=FALSE
`
to
`plugins/lighttable/export/force_lcms2=TRUE
`
and got a diff.png that looks like what you posted, from the same executable.

So my question was whether the integration test should use the expected.png produced with LCMS turned on or LCMS turned off.  My personal opinion is that the tests should by default force LCMS **off**, as it is a component outside of darktable's control.  Individual tests that specifically rely on the behavior with LCMS should of course turn it on.

TurboGit:
>  My personal opinion is that the tests should by default force LCMS off,

Ok, that's also @aurelienpierre 's preference so I'll join you on this one.

I'll update the run.sh to not use lcms, but before going further a bug on the export size should be fixed. I have reported here #5022. When this is fixed I'll regenerate all the expected output without lcms2 and with a max thread of 4.

parafin:
Question - how do you intend to limit number of threads?

ralfbrown:
There's a commandline option ( -t N ), which is also how I've been varying the number of threads to give performance results under differing parallelism in my other recent PRs.  The OpenMP spec has an environment variable for that purpose, but dt overrides it, so you have to use -t.

parafin:
Didn't know about this option. Looks what's needed indeed.

TurboGit:
@ralfbrown : I have just pushed a new version of the test driver that fix the number of threads and disable lcms2.

Note that for the test to pass it is needed to revert c12e3d3 and d881b85 (see discussion in #5228)

ralfbrown:
Pulled current master and reverted those two commits.  I still can't run the delta-E computation (as there is no delta_E function in my version of the colour module), but running `compare -metric ae` shows all of the tests to produce pixel-identical results except for 0003-denoise-bilateral, which differs by the minimum possible amount in four pixels.

The big download of all those reference images prompted the thought that it might be worthwhile moving them into a submodule so that those who merely want to compile git-master rather than do any development don't have to download an extra gigabyte in the git-ball (I know, the images are nowhere near that, **now**).

TurboGit:
@ralfbrown : To have delta_e available you need a recent version of the colour-science python module. There was one in Debian: python3-colour but this package was not recent enough for me, so I did:

`$ pip3 install colour-science`

To get latest version with delta_e support. All those information are in the top of my deltae script, let me know if it needs updating somehow.

parafin:
> The big download of all those reference images prompted the thought that it might be worthwhile moving them into a submodule so that those who merely want to compile git-master rather than do any development don't have to download an extra gigabyte in the git-ball (I know, the images are nowhere near that, **now**).

But there are several submodules already, which are required for building, so usual procedure downloads all of them. Having an exception for one submodule isn't convenient at all, I don't see how it can be done.

coffeeandspaceships:
Guys, would it help if I transposed the documentation @TurboGit already included in the integration test source tree over to the Developer Wiki somewhere? I can then also add all the hints and tips on how to use from the original PR which created the tests, and add the further info from this PR. There's very little info on testing in the old Wiki either.

TurboGit:
@coffeeandspaceships: sure go ahead, a wiki entry will be nice!

coffeeandspaceships:
First attempt here: https://github.com/darktable-org/darktable/wiki/Dev-guide---darktable-testing-tools

To keep THIS issue on topic - ff there are specific changes you want, please raise an issue and tag me in it (or assign it to me) - I am happy to create more pages if you nominate specific topics, and grow on what @aurelienpierre has started.

Re this issue, I will catch up my build environment with master, revert those two commits, and see if I can get colour-science working for me.

TurboGit:
Manually merged. Closing. Thanks.

TurboGit:
@coffeeandspaceships : the wiki page looks correct to me. thanks.

PR #5204, opened 2020-05-26
Wavelets denoising speedup

ralfbrown:
Some small speedups from optimization, and a larger relative speedup at higher thread counts by parallelizing a formerly single-threaded computation.  Note that this code generates a slightly different result than the old single-threaded version.  I have verified that the difference is due to numerical instability in the old code and that the new version is in fact more accurate (the fourth commit adds some code to demonstrate; see the full commit comment, and feel free to leave it out of any merge).

Timings with default parameters in Denoise(profiled) on the integration-test image:
Threads / Old / New (delta%)
1 / 11.712 / 11.342 (-3.1%)
2 / 6.108 / 5.814 (-4.8%)
4 / 3.306 / 3.068 (-7.2%)
8 / 1.944 / 1.691 (-13.0%)
16 / 1.181 / 0.969 (-18%)
32 / 0.759 / 0.556 (-26%)
64 / 0.611 / 0.371 (-39%)

### DIFF
--- a/src/iop/denoiseprofile.c
+++ b/src/iop/denoiseprofile.c
@@ -40,6 +40,9 @@
 #include <xmmintrin.h>
 #endif
 
+//check for regression
+//#define VALIDATE_SUMY2
+
 #define REDUCESIZE 64
 #define NUM_BUCKETS 4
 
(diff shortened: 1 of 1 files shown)


rawfiner:
Thanks for looking into the performance optimisation of this too!

About the difference due to numerical instability: if they induce "big" changes in the result for a set of parameters, we will need to add a new parameter to choose between legacy and new version. Though, from my first tests, the changes do not seem noticeable, so we probably won't need to do that.

About performance, preliminary tests are overall good on my side:
with 4 threads:
old | new
11.284 | 10.937
with 2 threads:
20.826 | 20.553
with 8 threads:
8.391 | 7.715

ralfbrown:
On the integration-test image, the reported Peak Absolute Error was the smallest increment representable in an 8bit-per-channel file (though that can yield a delta_E > 2 in darker pixels).

TurboGit:
The result the test image is:

```
Test 0013-denoiseprofile-wavelets
      Image mire1.cr2
      CPU & GPU version differ by 60158 pixels
      Max  dE         2.1908
      Mean dE         0.0014
      Count above max 8
  FAILS: image visually changed
         (11441 pixels changed)
```

On 8 pixels above dE 2.0, so good to me. I'll integrate and update the reference image.

TurboGit:
Manually merged.

PR #5206, opened 2020-05-26
Icons optical balance

Mark-64:
Now that #5150 is merged and geometry of icon buttons can be controlled from CSS, I am hereby proposing some further icons optimizations, with the purpose of removing the residual ad hoc hardcoded assumptions.
For this I have slightly resized / shifted some icons, fixed some drawing mistakes and also redesigned the "second display" that was looking ugly to me.
I took care of optical balancing and alignment, with the help of a tool which I made on purpose to assess the optical density through a blur.
This also lead to some simplification in CSS.
the result look pretty good to me, but there is a bit of subjectivity here, so feel free to give feedback

### DIFF
--- a/data/themes/darktable.css
+++ b/data/themes/darktable.css
@@ -394,13 +394,6 @@ overshoot.right
 }
 
 /* Set toolbars buttons in lighttable and darkroom */
-/* note that some button used to have different hardcoded paddings */
-/* overlay color changer */
-/* quick favorites */
-/* quick styles */
-/* display2 */
-/* they now look unbalanced */
-/* TODO: adjust icons individually */
 #header-toolbar #dt-toggle-button,
 #header-toolbar #dt-button,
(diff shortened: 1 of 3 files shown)


Nilvus:
Good and seeing that, I'm happy that I do not have time yesterday and this day to adapt my recent home work on CSS on top of your last merged PR. And good to see that simplify and the CSS and Cairo code. Will test that ASAP.

I remember some days ago @TurboGit said that with @AlicVB work on lighttable/culling and preview modes and my work on CSS, he finds darktable have a pro UI now. We could add your work on those icons also for that! I like that!

Mark-64:
> darktable have a pro UI now

Agree, and on Windows the improvement is even more visible.

By the way, I'm getting closer to a fix for the "more modules" treeview.
I know what to do, I have just to find the right way to implement it.

TurboGit:
Works for me.

PR #5208, opened 2020-05-26
keep in sync max size of thumbtable icons

Mark-64:
With #5150, all geometry properties of icon button has been moved to CSS.
Thumbtable overlays are independently sized, because they are designed to be dynamic.
Their size is dependent on the zoom level, up to a maximum. However, the maximum size is hardcoded to `DT_PIXEL_APPLY_DPI(20.0f)`.
This makes the size of overlay icons in Windows to grow much beyond the size of the corresponding icons (group, star, etc) in the top and bottom toolbars, and the GUI doesn't look nice.

With this PR, the real allocated size of the icons in the DT top / bottom toolbars is used to set the maximum size of the thumbtable icons.
The two are kept in sync on all OSs and also respond in sync to a change of CSS

Cattura2

### DIFF
--- a/src/dtgtk/thumbnail.c
+++ b/src/dtgtk/thumbnail.c
@@ -1201,6 +1201,9 @@ static void _thumb_resize_overlays(dt_thumbnail_t *thumb)
   PangoAttribute *attr;
   int width = 0;
   int height = 0;
+
+  int max_size = darktable.gui->icon_size;
+
   if(thumb->over != DT_THUMBNAIL_OVERLAYS_HOVER_BLOCK)
   {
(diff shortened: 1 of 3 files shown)


Nilvus:
Visually, good for me. I'm now waiting to see your 3 PR about icons and treeview merged to propose some CSS tweaks on top of those PR.

TurboGit:
Works for me. Merging, @Nilvus : you'll be able to work on the CSS part.

Nilvus:
> Works for me. Merging, @Nilvus : you'll be able to work on the CSS part.

Oh yes! Thanks for merging all those PR (and @Mark-64 of course for doing them).

PR #5210, opened 2020-05-26
More modules treeview

Mark-64:
This is an attempt to fix the behavior of the "more modules" treeview not following CSS style.
Please test it, especially @Nilvus, as i am not 100% sure it works as it should.
Feedback is welcome


Nilvus:
It works as it should on my side with that PR and without changing CSS. So issue solved for me. Hover effect is same as elsewhere. So good to be merged for me!

I'm surprise by what was needed to change for that. Thanks!

Mark-64:
> I'm surprise by what was needed to change for that

I tried at first with simpler solutions with only few changes:
1) using `field_hover_bg` in the color lookup in `_lib_modulelist_populate_callback()`. It works, but it's not really following CSS selector for the treeview, it's just another specific color which happens to be the right one in this version of CSS
2) trying to retrieve background-color CSS property for the active status, but it doesn't work. It always returns rgb(38,38,38) from darktable.css:1171 (`@field_bg`) I don't know why.

That's why I decided to change approach and instead use the selection object associated with the treeview. It requires some effort to get it sync-ed with the status of modules, but then it follows CSS naturally.

PR #5215, opened 2020-05-27
Improvement of the metadata_view window

jpverrue:
This PR proposes an improvement of the metadata_view module window.
The window is now scrollable and resizable with Ctrl+scroll


TurboGit:
Looks good and working for me.

PR #5218, opened 2020-05-28
better balance of icons in more modules treeview

Mark-64:
Just a small resize of the icons in "more modules" treeview, that were looking unbalanced as a consequence of my recent PR on optical alignment.
@Nilvus this will not influence your work on css.


TurboGit:
Works for me, thanks!

PR #5221, opened 2020-05-29
Fix data race in bilateral filter

ralfbrown:
While looking at the bilateral filter for potential speedup (it's used by eight different iops), I discovered that output varied from run to run with identical settings using the same executable.

I tracked it down to an unprotected update of a shared memory location in dt_bilateral_splat.  This causes some updates to be lost, increasing with the number of threads (it seems to require at least ten threads to manifest).  The worst-affected iop seems to be global tonemapping, with up to 5% of pixels changed in 16 bit per channel output when running 32 threads.

The included patch makes that shared update atomic.  The compiler complained about putting an 'atomic' directive inside a 'simd' directive, which is why the latter is commented out.  The patch does slow down the bilateral filter, but fortunately the iops using it aren't super-slow.  I also think we can regain the speed with an SSE code path, which is currently not implemented.


ralfbrown:
bilat-race

ralfbrown:
Never mind, I have a better solution that improves scalability and removes the data race at the same time.
Stay tuned for another PR.

ralfbrown:
See #5222.

PR #5229, opened 2020-05-29
Style changes for liquify.c ... many, many spaces.

GrahamByrnes:
At a certain point all the spaces in liquify.c drove me mad... blame it on confinement.
So if management wants to remove them, here is de-spaced version...


TurboGit:
Thanks !

PR #5231, opened 2020-05-29
add count of changed pixels to integration test

ralfbrown:
As it might be useful to know the total number of changed pixels as well as the number of visually-changed pixels (especially for those of use who can't properly run the deltae script), I've added capturing of that information from `compare` to the `run.sh` script.


TurboGit:
Manually merged. Closing.

PR #5232, opened 2020-05-30
Correct export size again.

### LINKED ISSUE #5228: export size not correct
**Describe the bug**
An attempt to fix this issue has been made in #5022. But this has introduced a regression.

**To Reproduce**
$ cd src/tests/integration
$ darktable-cli --width 2048 --height 2048 --hq true images/mire1.cr2 0001-exposure/exposure.xmp output.png
$ exiv2 output.png | grep -i "Image size"
Image size      : 2049 x 1365

**Expected behavior**

Reverting : c12e3d3 and d881b85

The export size is correct and 2048 x 1364

$ exiv2 output.png | grep -i "Image size"
Image size      : 2048 x 1364

 - Darktable Version: master
 - OS: GNU/Linux

**Additional context**

We should either revert commits c12e3d3 and d881b85 or have a fix for 3.2.

jenshannoschwalm:
To get what this about read #5228 and #5022

Hopefully final fixing for those issues. Done:
1. Corrected algo
2. Added docs about the rules
3. Added debugging output via -d imageio

What this about, (cited from the source)

The pipeline might have out-of-bounds problems at the right and lower borders leading to
artefacts or mem access errors if ignored. (#3646)
It's very difficult to prepare the pipeline avoiding this **and** not introducing artefacts.
But we can test for that situation and if there is an out-of-bounds problem we
have basically two options:
a) reduce the output image size by one for width & height.
b) change the scale. In theory this marginally reduces quality.

These are the rules for export:
1. If we have the **full image** (defined by dt_image_t width, height and crops) we look for upscale.
   If this is off use a), if on use b)
2. If we have defined format_params->max_width or/and height we use b)
3. Thumbnails are defined as in 2 so use b)
4. Cropped images are detected and use b)
5. Upscaled images use b)
6. Rotating by +-90° does not change the output size.
7. Never generate images larger than requested.

fixes #5228

### DIFF
--- a/src/common/imageio.c
+++ b/src/common/imageio.c
@@ -746,19 +746,48 @@ int dt_imageio_export_with_flags(const uint32_t imgid, const char *filename,
             ? FALSE
             : high_quality;
 
+  /* The pipeline might have out-of-bounds problems at the right and lower borders leading to
+     artefacts or mem access errors if ignored. (#3646)
+     It's very difficult to prepare the pipeline avoiding this **and** not introducing artefacts.
+     But we can test for that situation and if there is an out-of-bounds problem we
+     have basically two options:
+     a) reduce the output image size by one for width & height.
+     b) change the scale. In theory this marginally reduces quality.
+
+     These are the rules for export:
+     1. If we have the **full image** (defined by dt_image_t width, height and crops) we look for upscale.
+        If this is off use a), if on use b)
+     2. If we have defined format_params->max_width or/and height we use b)
+     3. Thumbnails are defined as in 2 so use b)
+     4. Cropped images are detected and use b)
+     5. Upscaled images use b)
+     6. Rotating by +-90° does not change the output size.
+     7. Never generate images larger than requested.         
+  */
+
+  const gboolean iscropped =
+    ((pipe.processed_width < (wd - img->crop_x - img->crop_width)) ||
+     (pipe.processed_height < (ht - img->crop_y - img->crop_height)));
+
+  const gboolean exact_size = (
+      iscropped ||
+      upscale ||
+      (format_params->max_width != 0) ||
+      (format_params->max_height != 0) ||
+      thumbnail_export);
 
   int width = format_params->max_width > 0 ? format_params->max_width : 0;
   int height = format_params->max_height > 0 ? format_params->max_height : 0;
-  if((!thumbnail_export) && (width>0))
-    width++;
-  if((!thumbnail_export) && (height>0))
-    height++;
+
+  if(iscropped && !thumbnail_export)
+  {
+    width = pipe.processed_width;
+    height = pipe.processed_height;
+  }
   
   const float max_scale = ( upscale && ( width > 0 || height > 0 )) ? 100.0 : 1.0;
-
-  const double scalex = width > 0 ? fminf(width / (double)pipe.processed_width, max_scale) : max_scale;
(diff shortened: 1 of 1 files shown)


TurboGit:
Still some issues, sorry....

```
/opt/darktable/bin/darktable-cli --width 2048 --height 2048 --hq true images/mire1.cr2 0009-flip-v/flip-v.xmp output.png ; exiv2 output.png | grep "Image size"
Image size      : 3908 x 2599
```

So for some images the final image size does not take into account the requested width and height.

src/common/imageio.c
```diff
@@ -746,19 +746,48 @@ int dt_imageio_export_with_flags(const uint32_t imgid, const char *filename,
...
+        If this is off use a), if on use b)
+     2. If we have defined format_params->max_width or/and height we use b)
+     3. Thumbnails are defined as in 2 so use b)
+     4. Cropped images are detected and use b)
+     5. Upscaled images use b)
+     6. Rotating by +-90° does not change the output size.
+     7. Never generate images larger than requested.         
+  */
+
+  const gboolean iscropped =
```
TurboGit, line 768: Shouldn't this be:
```

  const gboolean iscropped =
    ((pipe.processed_width < img->crop_width) ||
     (pipe.processed_height < img->crop_height));
```
crop_x - crop_width looks suspicious to me.

Let me check on all integration tests.

TurboGit: I confirm, with this change all the integration tests are passing fine, the export size is the same for all tests. So this looks good.

jenshannoschwalm:
Unfortunately your suggestion is not OK. The crop_width and height are misleading, they represent the border size from raw data. Thus I took raw size minus the according borders.

What probably wrong ist the test for cropping in line 782. It should include test for width and height both are 0.

Unfortunately im at work and phone, I'll have to test & fix later. Or you do and just use my pr and modify.

TurboGit:
> The crop_width and height are misleading, they represent the border size from raw data. Thus I took raw size minus the according borders.

Oh ! Really misleading naming indeed :(

> What probably wrong ist the test for cropping in line 782. It should include test for width and height both are 0.

Ok, will test this.

> Unfortunately im at work and phone, I'll have to test & fix later. Or you do and just use my pr and modify.

Ok too, if the above works I'll commit everything with updated expected png for the tests.

Thanks for the quick feedback.

TurboGit:
Lot better and I have decided to push this. The only issue is on the 3 flip tests in the integration testsuite where the actual width is 1362 where for all others we have 1364:

```
$ /opt/darktable/bin/darktable-cli --width 2048 --height 2048 --hq false images/mire1.cr2 0009-flip-v/flip-v.xmp output.png --core -d imageio; exiv2 output.png | grep "Image size"
1.605116 [dt_imageio_export] pipe 3908x2599, range 2048x2048 --> exact 1, upscale 0, scale 0.523797339, size 2048x1362
[export_job] exported to `output.png'
Image size      : 2048 x 1362
```
 
I was not able to see what could be done in imageio.c, I'm leaning into a possible issue in flip IOP. Any idea ? If you could have a look at this it would be great.

jenshannoschwalm:
I will...

jenshannoschwalm:
No, it's not the flip. It's clip&rotate, the parameters passed do a clipping to a smaller pipeline and so the export is smaller.

jenshannoschwalm:
Aah. It's indeed the clipping.

Issue #5227, opened 2020-05-29
Color Balance module: soft range for offset sliders


Mark-64:
In color balance all factor and saturation sliders have the same soft range (0 to 25% for saturation and -50% to 50% for factor).
However the offset part (shadows adjustment) is way too sensitive.
In most practical cases an adjustment of +/- 1% of factor has already a dramatic affect on the image.
You are forced to fiddle with fine right click adjustment all the time, which slows down the workflow.
Shadows naturally require low touch (that's why log scale is often used in audio / video manipulation.
Fact is that soft range can be increased on the gui, but it cannot be decreased.

My proposal is to set by default a narrower soft range for offset sliders. It doesn't make sense to have a range which is almost never used for 99% of its span.

By the way I could make a PR just to test the idea

aurelienpierre:
The problem is the behaviour changes if you are in slope/offset/power or lift/gamma/gain mode. The second behaves well, only the first one is super sensitive. I agree that current state is not ok for slope/offset/power, but managing the UI consistency between both modet is a problem for each I haven't found an elegant solution.

Mark-64:
this is solved in #5533

PR #5234, opened 2020-05-30
Color Balance: set different soft ranges for shadows midtones highlights

Mark-64:
With the aim to improve usability of Color Balance module and in reply to my own #5227 I propose this PR, where the soft range are differentiated per channel, instead of being all equal.

**lift / offset (shadows)**
factor: -5% -> +5% (instead of -50% -> +50%)
saturation: 0% -> 5% (instead of 0% -> 25%)
R/G/B: -0.05 -> +0.05 (instead of -0.5 -> +0.5)

**gamma / power (midtones)**
factor: -20% -> +20% (instead of -50% -> +50%)
saturation: 0% -> 20% (instead of 0% -> 25%)
R/G/B: -0.2 -> +0.2 (instead of -0.5 -> +0.5)

**gamma / power (highlights)**
factor: -50% -> +50% (instead of -50% -> +50%, unchanged)
saturation: 0% -> 50% (instead of 0% -> 25%)
R/G/B: -0.5 -> +0.5 (instead of -0.5 -> +0.5, unchanged)

The rationale is to have a span of control in each channel which perceptually produces a comparable change on the image and increase accuracy of setting, especially for shadows.
Yes the lift/gamma/gain method is less sensitive and needs less this change, but overall the usability looks pretty good to me with all methods.
Of course this is just a default soft range, you can always push it further if more adjustment power is needed.


dterrahe:
If this gets merged first, I'll update 3ba47c7278d1b8f0d206a3778779df7117ca6055 with the changes.

Nilvus:
I agree with those changes, except midtones factor that I would let it unchanged (or setting -40%/+40%)

Mark-64:
We can tune the numbers, I used a logic shadows < midtones < highlights: the weaker the signal, the smaller the correction capability need to be. 
I used a shadows -> x4 -> midtones and midtones -> x2.5 highlights. for sure there are more scientific ways to do it.
With this I can adjust the shadows sliders concentrating on the effect on the image, rather that trying to move the mouse one pixel more or less.

dterrahe:
So #5335 was merged first.. @Mark-64 please let me know if you want advice on how to fix the resulting merge conflicts.

Mark-64:
Messed up with rebase, closing here.
I will open another PR

dterrahe:
Why not force push? We've all been there!

Mark-64:
I still have to lean egit in eclipse... :-(

PR #5236, opened 2020-05-30
consistent ux for drawn masks and add tooltips

elstoc:
currently retouch and spot removal modules allow continuous mask
creation (with click) and single mask creation (with ctrl+click)

blendop masks allow single mask creation (with click) and continuous
mask creation (with ctrl+click)

this PR amends the retouch and spot removal modules to match the
behaviour of blendop masks and adds tooltips to inform the user of the
functionality.


TurboGit:
I had though about this when you create the previous PR. Fact is that in retouch it was thought to be very common to add multiple stroke whereas for others it was less frequent.

I like the consistency in this PR (consistency is an import point).

What other think about this ? @Nilvus you might want to comment as this is a very import user interaction ?

elstoc:
I did ask around on IRC before I raised this and there was some general agreement (2 users who commented) that this would be a good move from a consistency point of view. I do agree that for retouch/spot continuous add is the better default (and specifically why I created the blendop version inconsistently in the first place) but on reflection I think UI consistency is more important, and the shortcut for continuous mask creation is now at least more obvious.

Nilvus:
I'm not sure to understand what blendop means but if this PR is just about having same shortcuts for same actions, it's good for me. Seems more consistent indeed as many classic users will probably try same shortcut on same actions regardless of how the module works.

TurboGit:
So I revise my point of view and will go with this change for consistency.

Issue #5052, opened 2020-05-14
IOP Modules: Which Group?


elstoc:
In testing a recent PR I ran the script ``tools/iop-layout.sh`` to change the module group display order. However, I noticed that quite a few of the modules seemed to move to different groups after I ran the script, so there is a discrepancy between the shell script and the default groups that modules are assigned to in darktable 3.1.

Here's a list of the mismatches. I've highlighted in bold what I think might be the 'correct' grouping (though a few of these are very subjective). Where I think both might be incorrect I've put what I think is correct in brackets. In addition there's one (color transfer) that I can't find in darktable at all and two that aren't mentioned in the shell script (negadoctor/tonequal).

TBH I'm not all that bothered really which groups they should be in (we could just adjust the script to match current darktable gui grouping) but it would be nice to be consistent.

Module | Tools Group | GUI Group
 -- | -- | --
graduatednd | Basic | **Effect**
colorin | **Basic** | Color
lens | Basic | **Correct**
colorout | **Basic** | Color
ashift (perspective corr) | Basic | **Correct**
rotatepixels | Basic | **Correct**
scalepixels | Basic | **Correct**
tonemap | Basic | **Tone**
profile_gamma (Unbreak Input Profile) | Basic | **Color**
filmic | Basic | **Tone**
filmicrgb | Basic | **Tone**
bloom | Tone | **Effect**
colisa (Contrast/Brightness/Saturation) | Tone | **Basic**
atrous (Contrast Equalizer) | **Tone** | Correct
shadhi (Shadows Highlights) | **Tone** | Basic
colormapping | **Color** | Effect
colortransfer | **Color** | ???
colorize | Color | **Effect**
lowlight | Color | **Effect**
splittoning | Color | **Effect**
colorreconstruct | Correct | **Basic**
highlights (recovery) | Correct | **Basic**
liquify | **Effect** | Correct
retouch | Effect | **Correct**
sharpen | Effect | **Correct**
spots | Effect | **Correct**
negadoctor | ? | Basic
tonequal | ? | Basic (**Tone**)

Thoughts?

TurboGit:
BTW, we probably want to add new modules like negadoctor and filmicrgb into the iop_layout*.py scripts.

johnny-bit:
Hi! I've checked the list, thought about it and I recommend following:

|module|desired group|comment
|--|--|--|
graduatednd | Effect
colorin | Basic | it's required there imo
lens | Correct | it's correction :)
[...]

PR #5240, opened 2020-05-30
add missing modules to iop-layout*.sh files

johnny-bit:
adds missing modules to match where they are in gui

refs #5052

This does not change the group they are by default.


TurboGit:
Thanks !

PR #5243, opened 2020-05-30
add histogram modes accel

### LINKED ISSUE #5239: Add shortcut to toggle histogram modes
I find myself switching the histogram between histogram and waveform quite often, mainly because I'm still getting used to the waveform view. Right now it is not possible to assign a keyboard shortcut to the 'change histogram mode' button and I think it would be nice to have one. 

It doesn't need to have a default shortcut assigned, but I would like the possibility to assign one myself.

johnny-bit:
add possibility to change/cycle histogram mode/type via accel

now it'll be possible to:
- cycle between histogram modes/types
- cycle between histogram type
- cycle between waveform type

fixes #5239

### DIFF
--- a/src/libs/histogram.c
+++ b/src/libs/histogram.c
@@ -753,6 +753,106 @@ static gboolean _lib_histogram_collapse_callback(GtkAccelGroup *accel_group,
   return TRUE;
 }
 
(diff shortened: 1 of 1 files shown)


TurboGit:
Thanks !

Egocentrix:
That was quick, awesome!

PR #5244, opened 2020-05-30
Histogram cursor fix

johnny-bit:
histogram uses `GTK_HAND1` cursor to indicate possibility of dragging

this however makes it hard to use buttons because hand obstructs buttons and their icons.

with this PR cursor is set to pointer over histogram buttons.

Also I've noticed wrong indentation in `darktableconfig.xml.in` so this PR fixes that too.


TurboGit:
Another nice little details, like this. Thanks.

PR #5246, opened 2020-05-30
fix duplicate module not being enabled on duplication

### LINKED ISSUE #4796: New created instance of the module in "active modules" tab not shown because deactivated by default
**Describe the bug**

If you reorganize the modules in the "active modules" tab and create a new instance there, new instance is not displayed because it is disabled by default. 

You need to go to the tab where the new instance was created and enable it there to be displayed in the "active modules" tab. 

**To Reproduce**

1. Go to "active modules" tab.
2. Select an active module and click on "new instance" from the menu.  New instance of the module is not shown in the "active modules" tab and there is no indicator that a new instance was created.
3. Go to the tab where this module is normally located. There you will find a new instance that is shown as disabled. 
4. Activate it.
5. Go back to "active modules" tab. New instance of the module is now shown there.

**Expected behavior**

If you create a new instance, this implies that you are done with old one and want to continue working with new one. Accordingly, it would make sense to make the new instance active by default.

**Screenshots**

 - Darktable Version: master (3.1.0~git1282.615016be4)
 - OS: Kubuntu 18.04
 - OpenCL activated
 - GeForce GTX 1070, 

**Additional context**

johnny-bit:
gui reset is no longer needed since toast for hidden controlls are
moved to accel handling

fixes #4796 by essentially reverting fc92a41e28e140dc42bf951d674d4194468af1f7

### DIFF
--- a/src/develop/imageop.c
+++ b/src/develop/imageop.c
@@ -814,7 +814,6 @@ dt_iop_module_t *dt_iop_gui_duplicate(dt_iop_module_t *base, gboolean copy_param
   /* initialize gui if iop have one defined */
   if(!dt_iop_is_hidden(module))
   {
-    ++darktable.gui->reset;
     module->gui_init(module);
     dt_iop_reload_defaults(module); // some modules like profiled denoise update the gui in reload_defaults
     if(copy_params)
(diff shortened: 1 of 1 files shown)


TurboGit:
Works for me. Thanks.

Issue #5226, opened 2020-05-29
automatic cropping in perspective correction not ok when preview downsampling is enabled


MStraeten:
If preview downsampling (#4702) is enabled, the cropping frame for automatic cropping is also scaled - so a manual tweaking is not possible

**To Reproduce**
process a mage that really needs perspective correction (for minimal perspective distortion it's not that noticeable)
1. set downsampling to '1/4'
2. apply automatic perspective correction (eg.vertical correction)
3. activate automatic cropping "original format" or "largest area"
--> 
![perspective-correction](https://user-images.githubusercontent.com/39386816/83257975-bcfbea00-a1b5-11ea-8070-d0d5c0dd47b3.jpg)

**Platform (please complete the following information):**
- Darktable Version: darktable 3.1.0+1806~g70e3a7a78 (plain current master)
- OS: OSX
- OpenCL no dependency

GrahamByrnes:
I'm a couple of versions behind atm, I'll need to do some cloning.

However, with +1796 it works for me, provided I follow the slightly strange logic of crop & rotate:
-do automatic correction
-do automatic cropping (and so I then saw the frame shrunken onto the maximum-area crop, with the cropped off mess still visible;
-click on the module name (not the button) to remove focus.

The third step expands the crop-frame to full-frame. This doesn't work for you?

MStraeten:
the cropping itself is done indeed, but i'm not able to adjust the cropping because the frame doesnt cover the area to be cropped

GrahamByrnes:
It works for me, but there is a reason I created 4 options: try a less severe reduction, or turn it off.

MStraeten:
it should work for any user and platform ... if not then this behaviour needs to be documented as a feature ;)

TurboGit:
@GrahamByrnes : I can reproduce. The identified edges are ok but the cropped area and grids are completely off.

GrahamByrnes:
I'll have a look, but I suspect this is more an issue of how the "automatic" part works... I did write at one stage that reducing the resolution may not be ideal when edges were being traced.
So it may take some time, since I have no idea how auto-perspective works.

For testing, I fed in a completely normal photo and the proposed "correction" was extreme... the algorithm was picking up the lines in a woman's hair, and trying the straigthen it. It did this differently at 1/1 and 1/4, but both were absurd. In @MStraeten 's case, it seems the guess about which part of the image is the more interesting is wrong...

[...]

Issue #5225, opened 2020-05-29
automatic perspective correction not ok when preview downsampling is enabled


MStraeten:
If preview downsampling (https://github.com/darktable-org/darktable/pull/4702) is used the automatic detection gives different results in dependence of the downsampling factor. Just 1/1 gives useable results.

**To Reproduce**
process a mage that really needs perspective correction (for minimal perspective distortion it's not that noticeable)
1. set downsampling to 'original'
2. apply automatic perspective correction (eg.vertical correction)
3. set downsampling to '1/4'
4. apply automatic perspective correction (eg.vertical correction)
--> parameters differs (especially noticeable for rotation)

**Platform (please complete the following information):**
 - Darktable Version: darktable 3.1.0+1806~g70e3a7a78 (plain current master)
 - OS: OSX
 - OpenCL no dependency

**Additional context**

GrahamByrnes:
This doesn't surprise me so much: automatic correction implies finding edges, which will not be so easy if the preview is down-sampled.
So I'd suggest that in this case you'd be better off setting preview_downsampling to 1 (and restart... I'm working on that).

MStraeten:
the identified edges seems to be the same or at lease similar (same colorchecker shot as in #5226) the rotation is about 8 degrees instead of 0.07 
(sample: https://mega.nz/file/nCAXzI4Y#IikcBEwmKgJgkYgl0CDGES88qwa9h9cOlSaL6jwF1zE)

TurboGit:
@GrahamByrnes : I can reproduce. The identified edges are ok but the cropped area and grids are completely off.

GrahamByrnes:
Oh... I see. this is a different effect, yes.

Just the frame being shrunk by the down-sampling ratio. That should be fixable.

MStraeten:
here two very simple shots to demonstrate:
20200530-IMG_8649.CR2 & .xmp
https://mega.nz/file/mX5A1aJQ#WC0_NW1AOYN-TQhlB__JbvFFDhEqpRxSyUnLqQ9Zy9g
https://mega.nz/file/uboE2QKA#Iwm9SmJtdtAdvirc_zsTkyiE9BJXCWy8eP7MWMSoQwg
20200530-IMG_8653.CR2 & .xmp
https://mega.nz/file/mKp0BQ5Y#cAM0Ytp1SQf1WW8rxMaoDOYttpLIvfHFnGwcUhX1rqQ
[...]

PR #5249, opened 2020-05-30
applies preview_downsampling correcting to crop in ashift.c

GrahamByrnes:
This appears to fix issues (not: #5225) #5226 
Finally, my early diagnosis was correct (undersized bounding box wasn't allowing most of the image to be considered), but it took me a long time to find the "obvious" cure.

I still don't think this works as well with the lower resolution: if you rattle it around a few times with the different transformation options it eventually gets close, but at 1/1 it jumps there right away.
Or, there may still be errors... but it seems ok.

### DIFF
--- a/src/iop/ashift.c
+++ b/src/iop/ashift.c
@@ -973,7 +973,6 @@ void distort_mask(struct dt_iop_module_t *self, struct dt_dev_pixelpipe_iop_t *p
   const float cx = roi_out->scale * fullwidth * data->cl;
   const float cy = roi_out->scale * fullheight * data->ct;
 
-
 #ifdef _OPENMP
 #pragma omp parallel for default(none) \
   dt_omp_firstprivate(cx, cy, in, out, roi_in, roi_out) \
@@ -1056,6 +1055,7 @@ void modify_roi_out(struct dt_iop_module_t *self, struct dt_dev_pixelpipe_iop_t 
       yM = MAX(yM, pout[1]);
     }
   }
+
   float width = xM - xm + 1;
   float height = yM - ym + 1;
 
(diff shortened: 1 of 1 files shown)


GrahamByrnes:
PS: tested it on a more "natural" setting, ie  a building with rectangular windows, and it works just fine. A colour-checker in a black box is really a challenge!

MStraeten:
auto cropping is ok now, but the rotation issue is remaining (even with a much simpler image)

GrahamByrnes:
@MStraeten : yes, working on #5225, but completely lost just now.

GrahamByrnes:
I've found the scaling constant that fixes the transformation :-)

The bad news is that in fixing that, there are a bunch of UI things gone to...

Anyway, best not to look at this until I find out how to bring it all together.

TurboGit:
@GrahamByrnes : Good progress !

GrahamByrnes:
Current status: scaling `g_buf->scale` by prev_downsample enables the simplex algorithm to fit a correct transform... the parameters are with about 2% and it looks right. The cropping works.

What doesn't work is the display of edges on the screen: they appear to reflect rotations, but not shift or shear. However, this *seems* to relate only to the display, and not to the transforms. Given a photo with very little rotation and a *lot* of angle, the final image picks up the angle just as you'd like. 

The crop box and its manual adjustment works. The "bounding box" for picking out important lines sort of works, but the circle over here picks up a line over there. In this case it's not simple affine transforms (I thought I was on the way for a moment there), there is a homography command that isn't working properly. Thing is, it's called 11 times, and presumably the calls responsible for shifting images up & down the pipe are correct, and the call that works in tandem with simplex() to calculate the fit quality is correct (since the results are pretty good). But it's the same function...

Ok, back to it.

PS ashift works remarkably well btw, much better than I would have expected.

GrahamByrnes:
I think this works now. As you'd expect from the comment about rotations being correctly mapped but not lens angles, it was an issue of scaling the lines before they were transformed: rotations are invariant wrt vector length :-)

Last night there was a complaint related to OpenMP not opening. I made a change, it was still there, I changed it back, it wasn't. In any case, there is a small change in `process_cl` which I haven't been able to test.

With @MStraeten window image, when p/d is set to 1/4 there is a warning of insufficient structure. However using ctrl-click to get more lines does the trick

20200530-IMG_8649

TurboGit:
Finally works for me too. Great work! I have tested with all down-sampling values and without down-sampling. No issue, merging.

GrahamByrnes:
Quick review! Thanks.

PR #5223, opened 2020-05-29
Move buttons border to css

Mark-64:
In the logic of removing hardcoded gui choices and delegate those to css, with this PR we move to css the control of icon buttons border.
The modification in the code is easy, however there is the problem that in lib and iop modules there are currently both flat (i.e. with no border) and normal (i.e. with border) buttons and they can't be distinguished in css, unless setting extra widget names. So I defined the name "non-flat" for those buttons.
I adjusted css to keep the current aspect and doing so I also fixed remaining glitches from my previous PR.

@Nilvus plase have a look. I leave the choice to you whether to merge this first or rather wait for your modifications (in that case I will rebase, no pb)

PR #5253, opened 2020-05-31
Some more icons optical balance and fixes

Mark-64:
Some small changes and fixes to improve optical balance of icons, especially for the ones in blending box.
It does not need #5223, but of course it benefits from the fixes in there


TurboGit:
Even better triangles on the parametric blend UI. Thanks!

PR #5257, opened 2020-05-31
make sure copy_history module buttons get updated

### LINKED ISSUE #5188: "paste" and "paste all" remain grayed out, not active
openSUSE Tumbleweed 20200520
darktable-3.1.0~git1749.593f29a92-3096.1.x86_64
OpenCL on 
i7 12-core 36GB

after copying any or all of the history stack, "paste" and "paste all" do not indicate active state, remain light gray but mouse clicking on either performs expect action as do the hot keys

johnny-bit:
if history gets copied outside of copy_history module the module wouldn't know.

thanks to this change, module has greater chance of knowing.

fixes #5188

### DIFF
--- a/src/common/history.c
+++ b/src/common/history.c
@@ -1506,6 +1506,7 @@ gboolean dt_history_copy(int imgid)
   if(dt_dev_is_current_image(darktable.develop, imgid)) dt_dev_write_history(darktable.develop);
   return TRUE;
 }
+
 gboolean dt_history_copy_parts(int imgid)
 {
   if(dt_history_copy(imgid))
@@ -1517,6 +1518,7 @@ gboolean dt_history_copy_parts(int imgid)
   else
     return FALSE;
 }
+
 gboolean dt_history_paste_on_list(GList *list, gboolean undo)
 {
   if(darktable.view_manager->copy_paste.copied_imageid <= 0) return FALSE;
--- a/src/libs/copy_history.c
+++ b/src/libs/copy_history.c
@@ -40,7 +40,6 @@ DT_MODULE(1)
 
 typedef struct dt_lib_copy_history_t
 {
-  int32_t imageid;
   GtkWidget *pastemode;
   GtkButton *paste, *paste_parts;
   GtkWidget *copy_button, *delete_button, *load_button, *write_button;
@@ -241,17 +240,32 @@ static void paste_parts_button_clicked(GtkWidget *widget, gpointer user_data)
   }
 }
 
+static void _update(dt_lib_module_t *self)
+{
+  dt_lib_copy_history_t *d = (dt_lib_copy_history_t *)self->data;
+
+  const gboolean has_paste = darktable.view_manager->copy_paste.copied_imageid > 0;
+
+  gtk_widget_set_sensitive(GTK_WIDGET(d->paste), has_paste);
+  gtk_widget_set_sensitive(GTK_WIDGET(d->paste_parts), has_paste);
+}
+
 static void pastemode_combobox_changed(GtkWidget *widget, gpointer user_data)
 {
-  int mode = dt_bauhaus_combobox_get(widget);
+  const int mode = dt_bauhaus_combobox_get(widget);
   dt_conf_set_int("plugins/lighttable/copy_history/pastemode", mode);
+
+  _update((dt_lib_module_t *)user_data);
 }
 
 void gui_reset(dt_lib_module_t *self)
 {
-  dt_lib_copy_history_t *d = (dt_lib_copy_history_t *)self->data;
-  d->imageid = -1;
-  gtk_widget_set_sensitive(GTK_WIDGET(d->paste), FALSE);
+  _update(self);
+}
+
+static void _mouse_over_image_callback(gpointer instance, dt_lib_module_t *self)
+{
+  _update(self);
 }
 
 int position()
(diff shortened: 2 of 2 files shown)


TurboGit:
> if history gets copied outside of copy_history module the module wouldn't know.

You mean copied how ? With a key accel ?

johnny-bit:
Exactly. With alic's code you can copy from every where tumbtable is (ego
filmstrip)

wt., 2 cze 2020, 19:46 użytkownik Pascal Obry <notifications@github.com>
napisał:

TurboGit:
> Exactly. With alic's code you can copy from every where tumbtable is (ego filmstrip)

Ok, but updating by catching a signal when over a picture is not clean. You probably achieve what you want with dt_control_queue_redraw();.

johnny-bit:
hmmm...

I tried with:
- `DT_SIGNAL_CONTROL_REDRAW_ALL` - no luck.
- `DT_SIGNAL_CONTROL_REDRAW_CENTER` - no luck
- `DT_SIGNAL_SELECTION_CHANGED` - good result, status changes when another image is clicked!

would this be clear enough? I could also add check to see if current != pasted from in case single image is selected, so user can't even try to paste on image copied from :) (it wouldn't work anyway)

TurboGit:
Are you sure that dt_control_queue_redraw();. does not call gui_post_export() ? I think that the metadata module does update the GUI using this hook.

johnny-bit:
metadata module uses `DT_SIGNAL_MOUSE_OVER_IMAGE_CHANGE` (i know, I copied code from there ;)) This selection change signal is imo a bit better. metadata needs mouse over image though.

TurboGit:
> metadata module uses `DT_SIGNAL_MOUSE_OVER_IMAGE_CHANGE`

Because it needs to change the content of the entries each time a new image is under the mouse cursor. But here you just need to refresh the module button (status) on some action. And I suppose using gui_post_expose() (you need to add it as not defined in the copy_history lib at the moment) would work in your case.

johnny-bit:
Tested and works. Interestingly `gui_post_expose` is called with same frequency as `DT_SIGNAL_MOUSE_OVER_IMAGE_CHANGE`. That would mean selection change would fire less frequently. additionally I could use `dt_view_get_images_to_act_on` with updates to make sure paste/paste parts are only enabled in case it makes sense to call them (have paste & selected image != source)

What do you think?

TurboGit:
Works for me as-is.

> What do you think?

This improvement would be welcome indeed. No need to paste over the source image.

johnny-bit:
OK, I'll work on it and push once it's working nicely :)

TurboGit:
Testing more it seems that this slow down the thumbs display. Starting dt the thumbs are slow to be displayed and each time you move up or down in lt it is slow to refresh.

TurboGit:
Hum... Maybe not due to this patch...

johnny-bit:
TBH I think using selection changed signal would be better for it, since it's called way less often and history-related things only work on selected items. This way it couldn't mess with slowing down thumbnails ;)

So what do you think? Leave as-is (do updates on `gui_post_expose`) or rather go with `DT_SIGNAL_SELECTION_CHANGED` since it'll be called less often yet more predictable?

I'll obviously update code to only enable/disable buttons to make them work only in meaningful situations.

TurboGit:
> So what do you think? Leave as-is (do updates on `gui_post_expose`) or rather go with `DT_SIGNAL_SELECTION_CHANGED` since it'll be called less often yet more predictable?

I see, let's try the signal `DT_SIGNAL_SELECTION_CHANGED` as it would lead to less update.

TurboGit:
Looks better and better. Yet a minor issue, you can copy and fly over with the mouse to paste on any given image without selecting it. Do you think this case can be covered ? We need to enable paste when not over the image where copy happened.

johnny-bit:
The only clear way to do that (IMO) then is to use `DT_SIGNAL_MOUSE_OVER_IMAGE_CHANGE` :)
But since `_update` does a bit of soul searchin with checking active image etc, why not add nice little timeout, so that it ignores you when you fly over but updates once you hover over image?

I'll try it and see if I like it.

TurboGit:
> But since `_update` does a bit of soul searchin with checking active image etc, why not add nice little timeout, so that it ignores you when you fly over but updates once you hover over image?

Very little one I suppose it would work.

johnny-bit:
wow, now this feels nice :) It really gives nice visual indication of what's currently possible in copy_history module :)

timeout is nicely clamped and is based of average delay in develop mode, so probably should be fine :) max delay is 0.250s so still neglible and start delay is 0.125 so "long enough to give breathing time and short enough not to be noticeable". If someone has beefy rig and average delay gets low, it's clamped at 0.01s, so enough of a delay imo.

What do you think?

TurboGit:
That's perfect this way! It feels like darktable is alive :) Very nice. I'd like @Nilvus feedback on this as it is really a somewhat deep GUI change.

BTW, I'm sure something identical can be done for the module "selected images" and "export selected". If you have some spare cycles :)

johnny-bit:
Huh, I started working on styles one, but no problem in doing rest 😅

czw., 4 cze 2020, 23:06 użytkownik Pascal Obry <notifications@github.com>
napisał:

Nilvus:
Oh yes, really nice effect. That's also opens new possibilities when copy/paste on a lot of images. Thanks, I like that.

So indeed, implement that on other parts cited would be good.

johnny-bit:
I need to a small code correction to match the better way done in #5323 and it'll be ready to roll

johnny-bit:
This is now ready for review!

TurboGit:
All good now. Thanks!

PR #5259, opened 2020-05-31
Fix for markers clipping in gradient sliders

Mark-64:
Fix for incorrect position of triangle markers in gradient sliders, causing clipping
before
image
after
image
It will benefit from #5253, which includes more accurate triangles.


TurboGit:
Works for me, thanks!

PR #5262, opened 2020-05-31
Hides instead of minimizes window when quitting (#4218)

hgustafsson:
Small fix for OSX where the expected behavior on quitting is that the window hides instead of minimizes while cleaning up. (The default minimization animation on OSX is quite obtrusive.) 

Needs testing on Linux and Windows, and, depending on results, this pull request might need to be encapsulated in `#ifdef __APPLE__`

### DIFF
--- a/src/gui/gtk.c
+++ b/src/gui/gtk.c
@@ -870,7 +870,7 @@ void dt_gui_gtk_quit()
 
   // Write out windows dimension before miminizing
   dt_gui_gtk_write_config();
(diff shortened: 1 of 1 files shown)


parafin:
This is covered by #5124 I think.

hgustafsson:
Indeed the window closes as expected on #5124 instead of minimizing.
Thanks four pointing it out!

TurboGit:
So closing.

PR #5264, opened 2020-05-31
fix unnecessary "moving to XDG" nag in borderline cases

### LINKED ISSUE #5139: never ending notice: :[init] moving database into new XDG directory structure"
openSUSE Tumbleweed 20200517
darktable-3.1.0~git1707.1bd0d448b-3072.1.x86_64
OpenCL on 
i7 12-core 36GB

When opening dt from cl, get constant notice:
  [init] moving database into new XDG directory structure

Shouldn't this action complete and not be a never ending notice?

johnny-bit:
in case library was specified on command line - don't bother even calling move routine.

only mention moving to XDG if moving is possible.

fixes  #5139

### DIFF
--- a/src/common/database.c
+++ b/src/common/database.c
@@ -2438,8 +2438,11 @@ dt_database_t *dt_database_init(const char *alternative, const gboolean load_dat
   sqlite3_initialize();
 
 start:
-  /* migrate default database location to new default */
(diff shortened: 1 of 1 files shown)


TurboGit:
Manually merged. Thanks.

PR #5268, opened 2020-06-01
polish thumbnails margins

aurelienpierre:
Essentially, we ensure consistent margins on every side, to make it look cleaner.

before :
Capture d’écran de 2020-06-01 02-27-47

after:
Capture d’écran de 2020-06-01 02-25-33

Notice I have removed the vertical extension name because, until Gtk adds support for `line_height` property, there is no clean way to reduce the vertical letter spacing (in 3.0, it was done manually on Pango/Cairo offset drawing). Also, I'm not sure it's actually good (in ARW, the W is always going to be much larger than the other letters, and having an uneven column is really ugly).

@Nilvus @AlicVB

### DIFF
--- a/src/dtgtk/thumbnail.c
+++ b/src/dtgtk/thumbnail.c
@@ -230,6 +230,20 @@ static void _thumb_retrieve_margins(dt_thumbnail_t *thumb)
   gtk_style_context_get_margin(context, state, thumb->img_margin);
 }
 
+static void _thumb_write_extension(dt_thumbnail_t *thumb)
+{
+  // fill the file extension label
+  const char *ext = thumb->filename + strlen(thumb->filename);
+  gchar *ext2 = NULL;
+  while(ext > thumb->filename && *ext != '.') ext--;
+  ext++;
+  gchar *uext = dt_view_extend_modes_str(ext, thumb->is_hdr, thumb->is_bw);
+  ext2 = dt_util_dstrcat(ext2, "%s", uext);
+  gtk_label_set_text(GTK_LABEL(thumb->w_ext), ext2);
+  g_free(uext);
+  g_free(ext2);
+}
+
 static gboolean _event_image_draw(GtkWidget *widget, cairo_t *cr, gpointer user_data)
 {
   if(!user_data) return TRUE;
@@ -452,25 +466,6 @@ static gboolean _event_image_draw(GtkWidget *widget, cairo_t *cr, gpointer user_
     // for overlay block, we need to resize it
     if(thumb->over == DT_THUMBNAIL_OVERLAYS_HOVER_BLOCK) _thumb_resize_overlays(thumb);
 
-    // now that we know image ratio, we can fill the extension label
-    const char *ext = thumb->filename + strlen(thumb->filename);
-    gchar *ext2 = NULL;
-    while(ext > thumb->filename && *ext != '.') ext--;
-    ext++;
-    gchar *uext = dt_view_extend_modes_str(ext, thumb->is_hdr, thumb->is_bw);
-
-    if(thumb->img_width < thumb->img_height)
-    {
-      // vertical disposition
-      for(int i = 0; i < strlen(uext); i++) ext2 = dt_util_dstrcat(ext2, "%.1s\n", &uext[i]);
-    }
-    else
-      ext2 = dt_util_dstrcat(ext2, "%s", uext);
-
-    gtk_label_set_text(GTK_LABEL(thumb->w_ext), ext2);
-    g_free(uext);
-    g_free(ext2);
-
     // and we can also set the zooming level if needed
     if(thumb->zoomable && thumb->over == DT_THUMBNAIL_OVERLAYS_HOVER_BLOCK)
     {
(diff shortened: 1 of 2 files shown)


Nilvus:
> Notice I have removed the vertical extension name because, until Gtk adds support for `line_height` property, there is no clean way to reduce the vertical letter spacing (in 3.0, it was done manually on Pango/Cairo offset drawing). Also, I'm not sure it's actually good (in ARW, the W is always going to be much larger than the other letters, and having an uneven column is really ugly).

I'm not a fan of vertical extension name so it's better for me to let extension name same way on all images. About the W, as it's the width of the letter that cause that, I don't see anything else to do by now. So good for me.

aurelienpierre:
> Or maybe try to use condensed font instead of regular one here. Did you test that? So good for me.

No, the problem is it should happen in C (detect which letter it is, then stretch or expand the letters), there are no such microtype features in Gtk CSS. Switching to condensed versions on-demand will also change the line width of letters, so it will be weird anyway.

eyedear:
how about using a monospace font such as Roboto Monospace. would that works for the vertical extension. I made a mock up I think it looks nice. but the leading might need to change a bit 
thumb

aurelienpierre:
First of all, what is the problem that vertical extensions try to solve ?

That was introduced in 2.6 I think, because at that time, the extension was in the image box and didn't have its own line, so in portrait mode, the image was overlapping the extension string.

This is now fixed and extension + icons have a dedicated line in the thumbnail header, so they are guaranteed to never be hidden by the image.

Thus there is no problem to solve anymore, and I believe the display should be as systematic as possible (every thumbnail should look the same as much as possible). Having some extensions vertical and some horizontal makes visual scanning more difficult.

Finally, using monospace fonts doesn't solve the line height still, and the vertical extensions still look dented.

src/dtgtk/thumbnail.c
```diff
@@ -1377,12 +1388,14 @@ void dt_thumbnail_resize(dt_thumbnail_t *thumb, int width, int height, gboolean
...
+
   // file extension
-  gtk_widget_set_margin_start(thumb->w_ext, 0.045 * width);
+  gtk_widget_set_margin_start(thumb->w_ext, 0.03 * width);
+  gtk_widget_set_margin_top(thumb->w_ext, 0.03 * width);
 
   // retrieves the size of the main icons in the top panel, thumbtable overlays shall not exceed that
   int max_size = darktable.gui->icon_size;
-  const int fsize = fminf(max_size, .09 * width);
+  const int fsize = fminf(max_size, 0.94 * width / 11.0f);
```
AlicVB, line 1399: just out of curiosity : do you find all those values by try and guess or they comes for some clever computations ?

aurelienpierre: The margins are set to 3% on each side, so we get 94% of the full width for the inner content. In this width:
```
    // we need to squeeze 5 stars + 1 reject + 1 colorlabels symbols on a thumbnail width
    // stars + reject having a width of 2 * r1 and spaced by r1 => 18 * r1
    // colorlabels => 3 * r1 + space r1
    // inner margins are 0.02 * width
```
so we have 22 × r1 to squeeze in 94% of the width, each icon having a diameter of 2 × r1, hence 11 diameters. But for some reason, the font size was not using the icon radius as a base height, so this only sync font size on actual icon diameter.

aurelienpierre:
what about displaying extensions only when overlays are visible too ?

src/dtgtk/thumbnail.c
```diff
@@ -459,13 +459,15 @@ static gboolean _event_image_draw(GtkWidget *widget, cairo_t *cr, gpointer user_
...
     {
       // vertical disposition
       for(int i = 0; i < strlen(uext); i++) ext2 = dt_util_dstrcat(ext2, "%.1s\n", &uext[i]);
     }
     else
-      ext2 = dt_util_dstrcat(ext2, "%s", uext);
+    */
+    ext2 = dt_util_dstrcat(ext2, "%s", uext);
 
     gtk_label_set_text(GTK_LABEL(thumb->w_ext), ext2);
```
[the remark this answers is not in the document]
aurelienpierre: done

aurelienpierre:
I have an issue with the latest commit @AlicVB. When starting the application, extensions don't get displayed even if they should. Once we change the overlays mode, they come back as expected. I wonder where I forgot to init extensions display.

AlicVB:
hmm... reading your code, what I would have done : 
- in _thumb_update_icon : always show the extension widget
- in css : use something like
```
#thumb_ext { color: transparent; }
.dt_overlays_hover #thumb_main:hover #thumb_ext,
.dt_overlays_hover_extended #thumb_main:hover #thumb_ext,
.dt_overlays_always #thumb_ext,
 {color: ???}
```
I can't try just now, but I'll try to find some minutes to test that later in the evening if you want

aurelienpierre:
Ok I found some interesting : if `_thumb_write_extension(thumb);` (aka extraction extension from image) is not in `_event_image_draw()` (where it was before), the string is not properly initialized all the time.

AlicVB:
@aurelienpierre : that's because you need to put _thumb_write_extension a little later in the code : after the widget creation...
I'm now testing the css stuff as said. I come back soon :)

AlicVB:
@aurelienpierre : if I've not mess with git, you can have a look at the last commit from https://github.com/AlicVB/darktable/tree/polish_thumbs_2 I've fixed the extension visibility from your great work.
If that's what you want, maybe the simpler is that you copy-paste the commit content here... it's not a problem !

aurelienpierre:
Ok thanks a lot !

Nilvus:
It seems I will not have work on CSS here. Thanks for the work.

TurboGit:
All good! Thanks.

Nilvus:
Was too slow. Just tested this on my side. An hover effect missing for extension to be consistent for others and will probably be better to integrate those CSS lines in same existing ones to be sure all remains consistent. Will propose my CSS PR in the next days (rain coming so will have more time).

PR #5270, opened 2020-06-01
purge_from_cache.sh: make sure alias is expanded

jade-nl:
The current purge_from_cache.sh script does not expand the parsed alias set in common.sh and fails.

This small addition makes sure that bash expands the alias when doing command substitution.

fixes: https://github.com/darktable-org/darktable/issues/5271


AxelG-DE:
Tested.
works for me

Nilvus:
Works for me too and fix my #5271 issue. Thanks.

TurboGit:
Thanks !

Issue #3005, opened 2019-09-25
Vertical waveform display


spaceChRiS:
For some pictures the waveform display is extremely useful. I understand why cinematographers only rely on a horizontal waveform view, the typical aspect ratios tend to be landscape oriented (this may change with more and more smartphone videgraphy ;) ). However, in photography, it would totally make sense to not only have a horizontal but a vertical waveform display as well. Maybe even an arbitrary direction would make sense, but as pictures tend to be rectangular, maybe a vertical waveform is already enough.

I am not sure if it would make sense to change the orientation of the containing “box”, but probably it would. But as this would mess up the darkroom view a lot, this has to be discussed a bit more.

spaceChRiS:
Hm, I still consider this useful, despite the bot having another opinion.

parafin:
#4149 is somewhat related and mentions this problem in the comments.

dtorop:
@spaceChRiS I'm finally noticing this. I've wondered about this as well. Are you suggesting that the widget becomes vertical, or (and?) that the x/y axes on waveform can flip, such that brightness data is horizontal and spatial data is vertical?

It would be fairly trivial to make the waveform have an option to flip axes. The only question would be where to hide the extra button to do this -- unless this should be automatic for vertical images.

Of course unless the histogram widget changed shape, when axes were flipped there'd suddenly be a plethora of luminance data on the x-axis and not much spatial data on the y-axis. Regarding actually making the widget be vertical, see https://github.com/darktable-org/darktable/issues/4149#issuecomment-616407567:

> The preview on the left and the histogram on the right should have the same size to keep the GUI balanced.

I suppose it would be possible to change the orientation of the preview widget as well for vertical images, but this seems like a significant change.

johnny-bit:
@dtorop - recently @elstoc made a PR which allows histogram to change size. Houz's comment is just "should" not "has to" ;)

elstoc:
Yes, ctrl+scroll should resize the histogram vertically now

dtorop:
Just catching up on recent work! I appreciate the ability to vertically resize the histogram. I added a note (https://github.com/darktable-org/darktable/pull/5276#discussion_r437840580) as it may be easy to just resize the widget without worrying about the underlying buffer -- but I haven't tested how this looks.

[...]

PR #5276, opened 2020-06-01
adjustable histogram height

### LINKED ISSUE #5237: FR: Make histogram resizeable (and moveable?)
While I would never want to get rid of the histogram entirely, it does take up a fair bit of vertical screen real-estate in the darkroom, especially when also using some of the more complex iop modules with blendop masks. It would be great to have the ability to reduce the height of the histogram and/or be able to move it to the left-hand side panel.

elstoc:
allows height of the histogram module to be adjusted with ctrl+scroll

Resolves #5237

### DIFF
--- a/src/develop/develop.c
+++ b/src/develop/develop.c
@@ -141,9 +141,8 @@ void dt_dev_init(dt_develop_t *dev, int32_t gui_attached)
     // (regardless of ppd). Try to get enough detail for a (default)
     // 350px panel, possibly 2x that on hidpi.
     dev->histogram_waveform_width = darktable.mipmap_cache->max_width[DT_MIPMAP_F]/2;
-    // Hardcoded hack: histogram widget will probably be either 175 or
-    // 350 pixels high depending on hidpi
-    dev->histogram_waveform_height = 175;
+    const float histheight = dt_conf_get_int("histogram_height") * 1.0f;
+    dev->histogram_waveform_height = DT_PIXEL_APPLY_DPI(histheight);
     // making the stride work for cairo muddles UI and underlying
     // data, and mipmap widths should already reasonable, but better
     // to be safe, and the histogram is for the sake of UI, after all
--- a/data/darktableconfig.xml.in
+++ b/data/darktableconfig.xml.in
@@ -2058,7 +2058,7 @@
     <shortdescription>method to use for getting the display profile</shortdescription>
     <longdescription>this option allows to force a specific means of getting the current display profile. this is useful when one alternative gives wrong results</longdescription>
   </dtconfig>
-  <dtconfig prefs="misc" section="interface">
+  <dtconfig>
     <name>histogram_height</name>
     <type min="100" max="200">int</type>
     <default>175</default>
--- a/data/darktableconfig.xml.in
+++ b/data/darktableconfig.xml.in
@@ -2058,6 +2058,13 @@
     <shortdescription>method to use for getting the display profile</shortdescription>
     <longdescription>this option allows to force a specific means of getting the current display profile. this is useful when one alternative gives wrong results</longdescription>
   </dtconfig>
+  <dtconfig prefs="misc" section="interface">
+    <name>histogram_height</name>
+    <type min="100" max="200">int</type>
+    <default>175</default>
+    <shortdescription>height of the histogram module</shortdescription>
+    <longdescription>height of the histogram module in the darkroom and tethering views (needs a restart)</longdescription>
+  </dtconfig>
   <dtconfig>
     <name>metadata/resolution</name>
     <type>int</type>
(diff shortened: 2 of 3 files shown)


dterrahe:
Is ctrl-scroll already used in histogram? Otherwise that could be used to adjust height, like in other modules? Or/and grabbing border?

Nilvus:
> Is ctrl-scroll already used in histogram? Otherwise that could be used to adjust height, like in other modules? Or/and grabbing border?

Ctrl+scroll is used to adjust exposure directly on histogram but maybe we can't change the shortcut and, indeed, if possible update the height with Ctrl+scroll. The better would probably be to grab border, like panels, and let shortcut as it is.

dterrahe:
> Ctrl+scroll is used to adjust exposure directly on histogram

Is there a difference between scroll and ctrl+scroll? Because just scroll also adjusts exposure, so maybe ctrl currently simply doesn't change behavior and it could be made to adjust height. For consistency, ctrl-scroll in side-bar = height adjustment would be nice to support anyway. There is no real border between histogram and group buttons, so it might not be very obvious/discoverable if you have to grab the unmarked edge (mouse cursor would change, so that would be a hint). Obviously ctrl-scroll is not the most discoverable either :-)

I just worried that changing a percentage in the settings dialog and then having to restart to see the effect might be an inconvenient way to set a size that you have to judge by eye to get just-right. Or you might want to change it back and forth, for example if you are working in a huge module like color balance.

elstoc:
ctrl+scroll is doable - it's not currently used for anything else. I wouldn't know how to do a grab resize so I'd need some help (or someone to take over the PR).

elstoc:
I've pushed a new commit that allows the histogram to be resized with ctrl+scroll. However, I get a segfault when I use this on the waveform mode and then exit the darkroom.

Any ideas would be gratefully received!

elstoc:
Ok that was fun. I _think_ I've fixed the issue now.

TurboGit:
I think we can have that for 3.2 as it does nothing if we do not use the resize, so safe !

TurboGit:
Works for me! Thanks.

src/develop/develop.c
```diff
@@ -141,9 +141,8 @@ void dt_dev_init(dt_develop_t *dev, int32_t gui_attached)
     // (regardless of ppd). Try to get enough detail for a (default)
     // 350px panel, possibly 2x that on hidpi.
     dev->histogram_waveform_width = darktable.mipmap_cache->max_width[DT_MIPMAP_F]/2;
-    // Hardcoded hack: histogram widget will probably be either 175 or
-    // 350 pixels high depending on hidpi
-    dev->histogram_waveform_height = 175;
+    const float histheight = dt_conf_get_int("histogram_height") * 1.0f;
+    dev->histogram_waveform_height = DT_PIXEL_APPLY_DPI(histheight);
```
dtorop, line 145: I don't think you actually need to change the `histogram_waveform_height` to change the histogram widget size. Cairo draws the histogram buffer (`dev->histogram_waveform`) scaled to the width/height of the histogram widget, and should look OK even if it's not in a 1:1 (or 1:2 for hidpi) ratio to the widget.

It may be worth only adjusting the widget and not adjusting the buffer -- or having a separate config for the buffer size. This is because the memory use and processing time of calculating the waveform increase in proportion to `dev->histogram_waveform_height` even though the image quality doesn't improve that much.

dterrahe:
ctrl-scroll works in the opposite way to everywhere else. In other modules, scrolling down moves the lower border down, so the module gets bigger. Here, scrolling down reduces the height of the histogram.

src/develop/develop.c
elstoc: I think you're probably right here, but the previous height was 175 and its max size is limited to 200. I suspect most people will _reduce_ the height (or that's the impression I get from screenshots people have posted on other issues) which will reduce the processing time and memory use. Please note that a subsequent PR removed the use of DPI_PIXEL_APPLY_DPI, so the waveform height is now limited to between 100 and 200.

Feel free to change it if you end up submitting a PR for #3005.

elstoc:
> In other modules, scrolling down moves the lower border down

Do you have an example - I'm not sure I've seen this behaviour anywhere else? I thought up for bigger and down for smaller seemed logical but I can see the argument the other way too.

dterrahe:
I absolutely agree with you, but consistency trumps rationality. ;-)

In darkroom, look at the tagging or image information modules. There are some more in lighttable.

elstoc:
I agree with you about consistency and if I'd known about that behaviour I would have made it consistent. Thanks for pointing that out. Learn something new every day :-)

elstoc:
@dterrahe: #5376

PR #5277, opened 2020-06-01
imageio: fix thumbs export.

### LINKED ISSUE #5269: Severe display issue on some portrait images (generated cache issue from darktable)
**Describe the bug**
Portrait images could sometimes be shown with pixel offset display. This result in strange image. See 2 examples below:

**To Reproduce**
A way to reproduce (but not reproducible all times) is to display images in culling mode (I use always 2 images at the same time) and after displaying landscapes images, showing new portrait images could result on this artefact.

But, I can reproduce near all the times this issue by rotate to the left a portrait image in lighttable (so with left rotate button in selected images module on the right panel.

This issue has been discussed with @AlicVB and @aurelienpierre yesterday and could be related to  a Gdk issue when rotating the image. What's strange is I never reproduce the issue if rotating on the right.

By discussing with @AlicVB, we find that's probably be a cache generating issue has all portrait images displayed this way have cache images with the issue.

**Expected behavior**
Always have images displayed correctly on portrait ratio.

**Screenshots**
issue2
issue1

And a generated jpg picked up on cache (mipmap 3 here but could be picked up same way on all generated mipmaps cache):

 - Darktable Version: one of the last master
- No OpenCL compatible card so CPU work only here.

TurboGit:
Fixes #5269.

### DIFF
--- a/src/common/imageio.c
+++ b/src/common/imageio.c
@@ -770,11 +770,11 @@ int dt_imageio_export_with_flags(const uint32_t imgid, const char *filename,
      (pipe.processed_height < (ht - img->crop_y - img->crop_height)));
 
   const gboolean exact_size = (
-      iscropped ||
-      upscale ||
-      (format_params->max_width != 0) ||
-      (format_params->max_height != 0) ||
-      thumbnail_export);
+    !thumbnail_export
+    && (iscropped
+        || upscale
+        || (format_params->max_width != 0)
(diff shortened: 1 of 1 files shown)


jenshannoschwalm:
I thought about this for some time and likely found the really stupid problem i introduced, I found it not so easy to explain so i openen a seperate pr and added explainig comments there. Hope this get's to an end ...

PR #5278, opened 2020-06-01
Bilateral filter speedup

ralfbrown:
By precomputing as much as possible, we can remove 48 conditionals and 20 multiplications per pixel from the innermost loop.  This produces a quite noticeable speedup in all of the iops which use the bilateral filter (-25% runtime on the faster ones such as lowpass, monochrome, and local contrast; same absolute but percentage-wise smaller gains on the slower ones like global tonemapping and shadows/highlights).

Requires #5222 (whose commits you see included below as the first three), but posted now so I don't forget about it.


ralfbrown:
Superceded by #5998.

PR #5279, opened 2020-06-01
altered fix - when rotation on an image without history don't take it  for auto-apply.

### LINKED ISSUE #5274: image not altered when rotated
This may be related to #5269, but I'm not sure it's the same issue...

**Describe the bug**
With fresh raw files and "use embedded jpeg" pref, when you just rotate the image to the left or right, this doesn't refresh the thumbnails.

After some test and code reading, this doesn't comes from the lighttable rewrite for once ! If you launch dt with `-d cache`, you can see that the thumbnail of the rotated image if recreated... from the embedded jpeg !
That's due to the fact that the image is not considered as altered : `dt_image_altered` return `FALSE` if the image has just been rotated...
And the code in `mipmap_cache.c` in the function `_init_8` use this check to determine if we can load the image from the embedded jpeg or if we need to load it from scratch.

I remember that there has been some recent change on this area, so I think this comes from here. Maybe we should have 2 check : one for the user perspective and one for internal use ?

**To Reproduce**
1. open a collection with "fresh" images
2. select one landscape image
3. click on the "rotate to the left" button (always in lighttable)
=> see that the thumbnail doesn't change and doesn't have the "altered" icon either. Even if you change the thumb size.

**Expected behavior**
the thumbnail should be rotated

 - Darktable Version: master at 3.1.0+1861~g0750cbd1a
 - OS: ubuntu 20.04
 - OpenCL : same with or without

phweyland:
fixes #5274 
When lighttable uses embedded thumbnail and there is no auto applied presets, the imported images have no history. 
Applying a rotation was then considered as an auto applied operation (as applied on the first run) and the image not set as altered. That is what the fix is about.

Remain some potential improvements.
- in all cases, image rotation applies development (and may look different). If embedded thumbnail is active, rotation should be done without development (keeping the original embedded pixels).
- reset rotation doesn't really reset, but reapplies the initial position. The image is and remains altered.
- there is no undo/redo on rotation operations (there is no possibility to be back unaltered).

### DIFF
--- a/src/develop/develop.c
+++ b/src/develop/develop.c
@@ -70,7 +70,7 @@ void dt_dev_init(dt_develop_t *dev, int32_t gui_attached)
 
   dt_image_init(&dev->image_storage);
   dev->image_status = dev->preview_status = dev->preview2_status = DT_DEV_PIXELPIPE_DIRTY;
-  dev->history_updating = dev->image_force_reload = dev->image_loading = dev->preview_loading = FALSE; 
+  dev->history_updating = dev->image_force_reload = dev->image_loading = dev->preview_loading = FALSE;
   dev->preview2_loading = dev->preview_input_changed = dev->preview2_input_changed = FALSE;
   dev->image_invalid_cnt = 0;
   dev->pipe = dev->preview_pipe = dev->preview2_pipe = NULL;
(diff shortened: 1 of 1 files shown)


TurboGit:
Works for me.  Will merge this.

Do you want me to have a look at the undo for the rotation ? Or are you working on it ?

phweyland:
> Works for me. Will merge this.
> 
> Do you want me to have a look at the undo for the rotation ? Or are you working on it ?

I haven't started anything yet. Feel free to go. But I've some spare time. I can do it if you are too busy.

TurboGit:
I'll work on the **undo**, I need to do a bit of coding to keep in touch with the dt code :)

Issue #2077, opened 2019-02-09
(maybe) deprecate bauhaus controls


aurelienpierre:
The bauhaus widgets are custom-made GTK widgets living in `src/bauhaus`. They provide the comboboxes and sliders for the modules, with additional sweeteners like the fine-tuning mouse adjustement for the sliders.

The thing is the code is a bit messy, and essentially duplicates native GTK widgets, like regular comboboxes and scales. bauhaus comboboxes actually bring nothing more than the regular GTK ones. But they also don't fully implement the range of events native GTK widgets have (hover, active, selected), and can't be styled the same way through CSS (text size, colors, margins, etc.)

This is becoming a burden to maintain, because they need extra care to match visually the GTK comboboxes, so it's twice the work for the same features. 

I can see the added benefit of the sliders over pure GTK scales, but I would ditch the comboboxes without second thought.

PR #5282, opened 2020-06-02
Gradient sliders, further improvement

Mark-64:
Recently I proposed #5259 to avoid clipping in gradient sliders.
However I still had clipping in Windows 10 HiDPI, the reason being hardcoded geometry assumptions.
With this PR, the gradient sliders are further improved, by:

- drawing the widget in a fully proportional, DPI-safe way
- moving to CSS the geometry control (min-height, margin, border, padding)
- increasing a bit the size of the markers with respect to the gradient bar, they were too small and difficult to grab, IMO

The result seems pretty good to me:

Ubuntu
immagine

Windows
immagine

### DIFF
--- a/data/themes/darktable.css
+++ b/data/themes/darktable.css
@@ -550,7 +550,9 @@ overshoot.right
 #gradient-slider
 {
   min-height: 2.0em;
-  margin: 0.5em;
+  margin: 0;
+  border: 0;
+  padding: 1px 0.5em 0.3em 0.5em;
 }
 
 /* Entries options in lighttable */
--- a/data/themes/darktable.css
+++ b/data/themes/darktable.css
@@ -549,7 +549,7 @@ overshoot.right
 /* gradient sliders */
 #gradient-slider
 {
-  min-height: 2.0em;
+  min-height: 1.7em;
   margin: 0;
   border: 0;
   padding: 0.15em 0.6em 0.15em 0.6em;
--- a/data/themes/darktable.css
+++ b/data/themes/darktable.css
@@ -546,6 +546,13 @@ overshoot.right
   margin: 12px; /* not real pixels, this is used as a percent extra space */
 }
 
+/* gradient sliders */
+#gradient-slider
+{
+  min-height: 2.0em;
+  margin: 0.5em;
+}
+
 /* Entries options in lighttable */
 #left #lib-plugin-ui entry /* this will avoid having too much height on entry items in filter collections modules */
 {
--- a/data/themes/darktable.css
+++ b/data/themes/darktable.css
@@ -552,7 +552,7 @@ overshoot.right
   min-height: 1.7em;
   margin: 0;
   border: 0;
-  padding: 0.15em 0.6em 0.15em 0.6em;
+  padding: 0.25em 0.6em;
 }
 
 /* Entries options in lighttable */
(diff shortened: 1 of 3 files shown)


TurboGit:
Looks ok to me. @Nilvus : Is that ok to you ?

Nilvus:
Didn't test last commit by now but I found the arrow now too big. Something intermediate with last small ones would be better. And the hover effect is not consistent (but that was already the case before this PR. When hovering an arrow then moving to another, the hover effet (enlighten a little the arrow) doesn't update or after some time). Could this be improved (it's not CSS thing here)?

Mark-64:
> I found the arrows now too big. Something intermediate with last small ones would be better. 

I have allocated min-height in css as follows: 
30% upper arrows
40% gradient bar
30% lower arrows
If you want to reduce the size of arrows together with the bar it is enough to reduce min-height in css, but if you want to reduce the arrows with respect to the bars, then I have to change the code. Which one ?

> And the hover effect is not consistent (but that was already the case before this PR. When hovering an arrow then moving to another, the hover effet (enlighten a little the arrow) doesn't update or after some time). Could this be improved (it's not CSS thing here)?

The problem is that the Gtk+ custom widget implementation was half-baked,  some events simply do not work, like the enter-notify-event, which is essential for the hover effect.
So for now I moved only the geometry to css, but I am working to make the widget fully themable, which requires quite some changes.

Nilvus:
@Mark-64: Indeed, that's good to tweak min-height. I found 1.7em better than 2em. Could you test on your side and especially with HiDPi.

Another thing I forgot. When launching darktable from Terminal, when I open parametric mask and so your work, I have some lines like that one displayed:

`(darktable:760622): Gdk-CRITICAL **: 11:52:47.776: gdk_window_get_scale_factor: assertion 'GDK_IS_WINDOW (window)' failed`

I tested on master and no lines like that one. Better to fix that too I think.

Nilvus:
> > And the hover effect is not consistent (but that was already the case before this PR. When hovering an arrow then moving to another, the hover effet (enlighten a little the arrow) doesn't update or after some time). Could this be improved (it's not CSS thing here)?
> 
> The problem is that the Gtk+ custom widget implementation was half-baked, some events simply do not work, like the enter-notify-event, which is essential for the hover effect.
> So for now I moved only the geometry to css, but I am working to make the widget fully themable, which requires quite some changes.

About that, rewrite bauhaus widget would be good but seems tricky on what I discussed with @aurelienpierre (he posted long ago an issue about that: #2077). It would indeed be good to tweak the more with CSS. And the idea would of course to relate that with other issues like #5284.

But bauhaus rework should be now for 3.4 if changed, as I said on #5284 issue.

Nilvus:
I made some more tests on your CSS part of this PR.

I would finally suggest to update it that way:
```
#gradient-slider
{
  min-height: 1.7em;
  margin: 0;
  border: 0;
  padding: 0.25em 0.6em;
}
```

On padding/margin ones, if you set same ones on top and bottom and same on left and right, no need to write the four padding/margin, only 2 is better (first for top and bottom, second for left and right).

Here, I upgrade to 0.25em instead of 0.15em top and bottom padding. You will see that this update aerate the sliders and arrow are better suitable/visible, and with a better size with the 1.7em.

Feel free to update your PR with that. No problem for that. Or I will integrate that on my future CSS PR.

Mark-64:
Converted to draft to check the issue and possibly to move to a fully themable widget

Mark-64:
Ready for review. 
The hover effect still doesn't work, but I would leave that for another PR

TurboGit:
Ok works for me. Thanks!

PR #5283, opened 2020-06-02
don't use a special layout for color assessment mode - 

### LINKED ISSUE #5200: Don't save a special view for color assessment mode
When toggling on/off the ISO 12646 mode (color assessment mode) in darkroom, the collapsed state of GUI panels but also their width (since we can dynamically resize them) is kept in memory, for each view mode (default darkroom or darkroom in color assessment mode).

As a result, when toggling the color mode, the whole GUI moves and shrinks/dilate. This behaviour is confusing for several users who would prefer to keep the whole GUI as-is (or, at least, keep the current width of the sidebars), and only add or remove the white frame.

The current proposal is to make color assessment only a sub-option of darkroom and not a whole view state. Panels should be preserved and managed apart.

aurelienpierre:
use generic darkroom layout

Fix #5200

### DIFF
--- a/src/views/darkroom.c
+++ b/src/views/darkroom.c
@@ -194,7 +194,7 @@ static dt_darkroom_layout_t _lib_darkroom_get_layout(dt_view_t *self)
 {
   dt_develop_t *dev = (dt_develop_t *)self->data;
   if(dev->iso_12646.enabled)
-    return DT_DARKROOM_LAYOUT_COLOR_ASSESMENT;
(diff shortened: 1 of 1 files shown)


elstoc:
Surely you don't need the if statement at all now?

aurelienpierre:
We may want to keep the if in case someone wants to extend it in the future (keeping in mind our non-existing dev docs).

TurboGit:
Works for me. Thanks!

PR #5288, opened 2020-06-02
Corrected scaling for image export

### LINKED ISSUE #5269: Severe display issue on some portrait images (generated cache issue from darktable)

jenshannoschwalm:
There are some severe issues related to correct export, see issue #5269.
Also @TurboGit suggested a pr #5277.
> In case width is initialize with pipe.processed_width and height with pipe.processed_height, the sale will be close to 1 but very little less. (max_scale is 1.0).
So the rounding errors could come from this. What do you think ?

I now fear staring at source made me somewhat blind or simply too much work the last week.
It seems to be a very stupid bug ... 

If there is an out-of-range situation it's safe to decrease the output size. But if we want to keep the size we must not decrease the scale but must **increase** it to stretch the image into correct output.

If we decrease we might end up in what has been shown by @Nilvus (we shear by one pixel per line) or by @tomaszg7 (right&bottom line problems)

Please test carefully, hopefully Fixes #5269

### DIFF
--- a/src/common/imageio.c
+++ b/src/common/imageio.c
@@ -752,7 +752,7 @@ int dt_imageio_export_with_flags(const uint32_t imgid, const char *filename,
      But we can test for that situation and if there is an out-of-bounds problem we
      have basically two options:
      a) reduce the output image size by one for width & height.
-     b) change the scale. In theory this marginally reduces quality.
+     b) increase the scale while keeping the output size. In theory this marginally reduces quality.
 
      These are the rules for export:
      1. If we have the **full image** (defined by dt_image_t width, height and crops) we look for upscale.
@@ -786,9 +786,11 @@ int dt_imageio_export_with_flags(const uint32_t imgid, const char *filename,
   }
 
   const double max_scale = ( upscale && ( width > 0 || height > 0 )) ? 100.0 : 1.0;
+
   const double scalex = width > 0 ? fmin((double)width / (double)pipe.processed_width, max_scale) : max_scale;
   const double scaley = height > 0 ? fmin((double)height / (double)pipe.processed_height, max_scale) : max_scale;
   double scale = fmin(scalex, scaley);
+  double corrscale = 1.0f;
 
   int processed_width = 0;
   int processed_height = 0;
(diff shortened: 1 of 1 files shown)


TurboGit:
Tested on my side, no issue found.

@Nilvus : Can you also test on your side ?

Nilvus:
> Tested on my side, no issue found.
> 
> @Nilvus : Can you also test on your side ?

Yes, I just see all your comments on my issue and this PR. I suppose you will close your PR about that @TurboGit. I will compile this PR and remove my broken jpg files from caches before testing. Will come back soon.

Nilvus:
@jenshannoschwalm and @TurboGit: I have some good news too. I tested a lot of vertical images I had and nearly all images that were affected by the issue and try to rotate them to the left and right. I also export one of the images affected I posted on the issue and size is precisely size set, the image is as it should. So no more issues found with that PR. Seems to be solved. Thanks for the work. Hope time will confirm the issue is behind us.

jenshannoschwalm:
Thanks for the good news. At least this was a definetve bug so chances are not bad ...

TurboGit:
Ok, I'll merge this at this point. We really need more field testing.

There is still an issue where

TurboGit:
Manually merged.

jenshannoschwalm:
> There is still an issue where all the flip have size 2048 x 1362 whereas all others are 2048 x 1364.

Ok, i'll stay on track report

jenshannoschwalm:
I guess you want to commit yourself?

TurboGit:
I confirm :( Don't know how this happened! Fixing. Thanks for looking into this. At this stage the integration testsuite is clean.

jenshannoschwalm:
I see you working so hard on dt. Maybe lack of sleep? 😊

PR #5297, opened 2020-06-03
preferences: ensure that spin button values are saved

### LINKED ISSUE #5294: Spin-button field in prefs don't get saved immediatelly on prefs-close
**Describe the bug**

When entering value in spin button inputs (eg lighttable -> number of folder levels to show in lists) value isn't saved when clicking "x" or exiting via "esc" key

**To Reproduce**

Obviously value should be saved :wink: 

**Screenshots**

unfortunatelly performing screenshots is problematic as "changed" or some other event is fired before screenshot is completed.

 - Darktable Version: 3.1.0+1703 (others up to current master are affected)
 - OS: Gentoo Linux
 
**Additional context**

 - Can you reproduce with another Darktable version? Yes, pretty much since new prefs window. With old prefs window it seems that clicking "close" button allowed spinbutton to trigger changed event. clicking on "X" however still produces same result
 - Did you compile Darktable yourself? If so which compiler was used, with what options? Yes, typical build.sh
 - Is the issue still present using an empty/new config-dir? yes :)

This stems from discussion in #5293 and I hope @elstoc'll be able to chime in here :)

elstoc:
move keyboard focus to the preferences dialog in order to ensure that
spin button values are saved.

Resolves #5294


TurboGit:
Thanks !

PR #5299, opened 2020-06-03
Update french translation (and POTFILES.in)

Nilvus:
Update french translation and adds missing files in POTFILES.in to update translations.

I update standard color matrix (and enhanced) but this remains in english in the UI . I don't understand why this is not update. All seems good from colorspaces.c file to POTFILES.in or fr.po but no update of the text in french UI. This remains the only not translated string.


TurboGit:
Ok merged and I have fixed the issue with the color matrix being not translated. Thanks.

Issue #5300, opened 2020-06-03
Glitch issue on vertical images in overlay_hover and overlay_hover_extended modes


Nilvus:
**Describe the bug**

See below (extension name issue and no space on the bottom for stars icons) when hovering images:

hover mode:
![2020-06-03_21-42](https://user-images.githubusercontent.com/45535283/83682011-95f14e00-a5e3-11ea-9218-07108c81ffd5.png)

hover_extended mode:
![2020-06-03_21-47](https://user-images.githubusercontent.com/45535283/83682223-ec5e8c80-a5e3-11ea-88e3-ee2d25ed891d.png)

On both modes hover overlay and hover extended overlay. This issue probably appear from #5268. @aurelienpierre or @AlicVB: this one probably for one of you.

**Expected behavior**
Extension image should not be overwritten by image and star should remains below the image not hover it.

Nilvus:
Same would be if audio/group ones. So need to be sure top and bottom margins are the same on all overlays mode, even if nothing shown.

AlicVB:
> Extension image should not be overwritten by image

That's an "historical" reason : the extension was part of the background. I don't care too much about it's position, so we can make it work like all other icons... @aurelienpierre : any thought ?



> star should remains below the image not hover it.

Nope, that's "by design" : I've added the "always_*" mode which reserve the space need for the overlays. In hover_* mode, we use the maximum space for the thumbnail and we draw overlays over the thumbnail on mouse over (note that there was a semi-transparent background at some point, but I remove it because it was too ugly for lot of people)

 

> So need to be sure top and bottom margins are at least added when hovering or better let them set.

Not sure to like the idea either : that would means a "jumpy" gui when the mouse move between thumbs (not to speak that it would require some rescaling of the thumbnail at each mouse_over_id change and that's exactly what I've tried to avoid with this rewrite)

elstoc:
> Nope, that's "by design"

Yes don't change that! I'm using CSS tweaks to make the 'hover' mode overlays look like that all the time. I want maximum image size and to be able to see star/colour ratings for all my images (like in 3.0.x).

Nilvus:
Ok, I understand about stars, except that they can't be over the image on classic hover mode (first image above) just because the stars will be quickly not visible. But indeed, you're right for the second image and hover_extended mode. My bad here!

So remains just 2 things:

- Top margin should remain the same to be sure all icons and ext name are not hidden by the image or override the image in all mode and hover effect here.
[...]

PR #5303, opened 2020-06-03
New set of CSS tweaks

Nilvus:
- Fix delete folder dialog window when removing all images of a folder. As I don't know Gtk code, I just found the Gtk tweak in film.c. It works but the content is not adjusted (a better fix is probably possible). Anyway, the result is much better than it was. See below:

before:

after:

- Improve treeviews and especially hover and selected effects.
- Improve module label title when on focus mode. I've tried to find a compromise on color for both disabled and enabled modules. This tweak is more suitable on enable module than disable one but can't set a different setting regardless of module status. To improve that, would probably need some Gtk need. See below:

- Make stars on all states consistent in all themes (that was not the case!)
- Make all checkbox consistent and so displayed same way on all the UI
- Integrate CSS tweaks from #5268 with related icons/infos ones to be sure all remains consistent in the future.
- Avoid to apply border-radius and margins on modules-tabs and blending-tabs and have better hover effect on module groups tabs and mask tabs.
- Minor other tweaks
- And set CSS for #5179 PR by @AlicVB (so better to merge it before my PR).

### DIFF
--- a/data/themes/darktable.css
+++ b/data/themes/darktable.css
@@ -133,16 +133,16 @@
 @define-color lighttable_bg_font_color @grey_75;
 
 /* Lighttable and film-strip */
-@define-color thumbnail_font_color @grey_30;
+@define-color thumbnail_font_color @grey_25;
 @define-color thumbnail_bg_color @grey_40;
-@define-color thumbnail_star_bg_color @grey_45;
+@define-color thumbnail_star_bg_color @grey_50;
 @define-color thumbnail_fg_color @grey_60;
 @define-color thumbnail_bg50_color @grey_100;
 @define-color thumbnail_infos_color @grey_20;
-@define-color thumbnail_selected_bg_color @grey_50;
-@define-color thumbnail_hover_bg_color @grey_85;
+@define-color thumbnail_selected_bg_color @grey_45;
+@define-color thumbnail_hover_bg_color @grey_80;
 @define-color filmstrip_bg_color @darkroom_bg_color;
-@define-color thumbnail_localcopy_color @grey_75;
+@define-color thumbnail_localcopy_color @grey_70;
 
 /* Brushes */
 @define-color brush_cursor alpha(white, .9);
@@ -286,6 +286,7 @@ textview
 {
   border-radius: 3px;
 }
+
 /* Default text fields and text boxes */
 entry,
 textview
@@ -1880,9 +1881,19 @@ radiobutton label
   padding: 0.2em;
 }
 
+radiobutton label
+{
+  margin: 2px;
+}
+
 radiobutton radio
 {
   padding: 0.25em;
+}
+
+spinbutton>button
+{
+  margin: 2px 1px;
 }
 
   /*** for following settings line, remember which popover mode name and related line on popover menu
@@ -1903,12 +1914,12 @@ radiobutton radio
 #thumb_image
 {
   border: none;
-  margin: 3px; /* not real pixels, but percentage of the thumb size */
+  margin: 2px; /* not real pixels, but percentage of the thumb size */
 }
 
 .dt_thumbnails_2 #thumb_image /* update consistently margins if big thumbnails (less images per row) are shown */
 {
-  margin: 3px;
+  margin: 1.5px;
 }
 
 #thumb_main:active #thumb_image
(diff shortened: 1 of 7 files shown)


Nilvus:
Forget to rebase... Will fix that.

elstoc:
> Improve module label title when on focus mode

I'm not sure I like the focus effect, judging by the screenshots. It looks blurred to me. And on the disabled module I think it's a bit too prominent.

Nilvus:
> > Improve module label title when on focus mode
> 
> I'm not sure I like the focus effect, judging by the screenshots. It looks blurred to me (maybe that's just the images though?). And on the disabled module I think it's a bit too prominent.

That what I describe on disabled module but I can't add something on label without applying on any focus module regardless of status. And the highlight is as much as I can to the icon highlight you set and the better compromise I found regardless of disabled problem. But the better thing is to test and not only see on screenshots. I'm not totally satisfied but I can't do more without Gtk work to better set it regardless of module status.

Nilvus:
A less little focus label text on last commit.

TurboGit:
@Nilvus : as @elstoc said the selected module name is also quite blurry on my side and hard to read. The previous version was better to me.

data/themes/darktable.css
```diff
@@ -594,7 +612,8 @@ overshoot.right
 
 .dt_module_focus #iop-panel-label /* set focus mode */
 {
-  color: @plugin_focus_color;
+  color: shade(@plugin_label_color, 1.1);
+  text-shadow: 0px 0px 2px shade(@plugin_label_color, 1.05);
```
TurboGit, line 616: I suppose this is the culprit :)

Nilvus: @elstoc and @TurboGit: I've tried on this one to set the label as possible near to the new highlight focus icon but as said on first post, not totally convinced by the result. So you're probably right to not let that. Anyway, the previous version lack of contrast. So could you test with that change before I merge it:

```
.dt_module_focus #iop-panel-label/* set focus mode */
{
  color: shade(@plugin_label_color, 1.2);
}
```
Waiting for your feedback.

Nilvus:
PR put back to draft for label controversial choice and to find a possibility to CSS to fix #5300

data/themes/darktable.css
TurboGit: Looks good to me.

TurboGit:
As you are at it... The lisibility of the selected line in the overlay pref is hard to read:

Capture d’écran de 2020-06-04 23-26-39

This is true in all three themes a bit less problematic for the grey one though.

Nilvus:
Last commit adds more things:
- fix popover hover effect (so issue posted just above by @TurboGit)
- remove collapse-buttons fix as it is fixed by #5314 PR
- set CSS to improve display since #5311 PR
- adjust module label focus display as discussed here with @elstoc and @TurboGit 
- adjust checked and hover buttons effect to have better consistency between themes

And fix #5300 and brings solution discussed with @AlicVB and @elstoc
- propose a new way to make bottom infos visible in hover overlays mode when images are vertical or square (will test to adapt on other overlays ; or not). _I will make more tests so this could be improved/changed._
- hide top infos (icons and extension name) in both hover and hover extended modes to avoid not wanted overriding things and let space to the image (so avoid to add a not wanted top margin on those modes).

Nilvus:
@TurboGit: HELP, big problems here. it's a mess here, how I can I change that?

Nilvus:
Pfff, finally succeed to but is it possible to remove all not wanted history here? The PR become hard to read.

Nilvus:
Last commits improve all overlays modes on all themes. Need to be tested on all modes and especially those ones:

- hover and hover extended in file manager
- always and always extended in culling and preview

I will now make a final review before putting that PR as ready for review.

Nilvus:
@TurboGit or @aurelienpierre: ready for review

Nilvus:
Don't merge, just forgot to adjust something (a little one). Will update that this evening.

Nilvus:
Ok, all good now.

Nilvus:
> Code is good. Testing…

Just made a little cleanup on last commit.

aurelienpierre:
Thanks !

PR #5296, opened 2020-06-03
bauhaus sliders: Thinner baseline and triangle

elstoc:
Resolves #5284.

Before

Screenshot_2020-06-03_17-05-14

After

Screenshot_2020-06-03_16-47-20

Issue #5304, opened 2020-06-03
Map view: polish localisation results list for 3.2


Nilvus:
**Describe the bug**
As we are on polish the UI for the 3.2, it remains a thing I just see and will probably need some Gtk things to then tweak CSS, on map view.

Just search a place on top right module and see:

![2020-06-03_23-27](https://user-images.githubusercontent.com/45535283/83691475-862d3600-a5f2-11ea-9fb9-8b15f946f68c.png)

The list is just awful. Results needs to be aerate between them, the left arrows are strange to be there and that way. So we need to have a polished list and space between results like on most dropdown (for example filter collections).

PR #5309, opened 2020-06-04
make bauhaus sliders slightly more compact

aurelienpierre:
before:
Capture d’écran du 2020-06-04 17-04-40

after:
Capture d’écran du 2020-06-04 17-02-33

following #5296

### DIFF
--- a/src/bauhaus/bauhaus.c
+++ b/src/bauhaus/bauhaus.c
@@ -494,7 +494,7 @@ static void dt_bh_class_init(DtBauhausWidgetClass *class)
 void dt_bauhaus_load_theme()
 {
   darktable.bauhaus->line_space = 1.5;
-  darktable.bauhaus->line_height = 10;
+  darktable.bauhaus->line_height = 9;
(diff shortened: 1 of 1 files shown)


elstoc:
You beat me to it. I was just playing around with those settings. The screenshots are very big! Yet still the difference seems very subtle. I'm not sure I can see much between the before and the after.

Note that these values are also used for comboboxes so it might be better to screenshot something like lens correction (which contains both) for comparison.

elstoc:
> The problem is you don't have a lot of room for adjustment

Agreed. Looks good to me.

TurboGit:
Looks good to me, on the export dialog I gain one line and a half. Not bad at all. I have not checked but I suppose the "color balance" should gain quite some space with all the sliders. Thanks!

Nilvus:
@elstoc and @aurelienpierre: good work for both of you. Will remain now to polish the issue some little thing and especially places result list on map view. See #5304

elstoc:
Yay. Now with this plus a smaller histogram the color balance module finally fits on my screen without scrolling!

PR #5315, opened 2020-06-05
refactor shutter speed formatting code

hatsunearu:
the shutter speed formatting code is repeated in various parts of the code, and in one point incorrectly repeated. 

I put the shutter speed formatting code in variables.c, and it is used everywhere that requires shutter speed formatting.

I also fixed the capital F in the metadata aperture printing.

### DIFF
--- a/src/common/variables.h
+++ b/src/common/variables.h
@@ -61,8 +61,6 @@ char *dt_variables_expand(dt_variables_params_t *params, gchar *source, gboolean
 /** reset sequence number */
 void dt_variables_reset_sequence(dt_variables_params_t *params);
 
-char *dt_format_exposure(float exposuretime);
-
 // modelines: These editor modelines have been set for all relevant files by tools/update_modelines.sh
 // vim: shiftwidth=2 expandtab tabstop=2 cindent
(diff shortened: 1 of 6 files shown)


src/common/variables.h
```diff
@@ -61,6 +61,8 @@ char *dt_variables_expand(dt_variables_params_t *params, gchar *source, gboolean
 /** reset sequence number */
 void dt_variables_reset_sequence(dt_variables_params_t *params);
 
+char *dt_format_exposure(float exposuretime);
```
TurboGit, line 64: As you see the convention is to use dt_variables_... in variables.h.

I also feel that such common routine should not be declared here. Maybe in utility.h ?

`dt_util_format_exposure(const float exposuretime);`

Or maybe another proposal ?

hatsunearu:
i agree. i will work on putting this in util. i was actually looking for a better location to put it in, haha

hatsunearu:
@TurboGit looks like everything checks out, let me know if there are other issues, i'll be going to sleep lol

TurboGit:
Looks good to me now. Thanks.

PR #5316, opened 2020-06-05
bauhaus controls: pass keypress event to view if not handled

elstoc:
if bauhaus controls do not handle a key press event and it is not
handled by an accelerator, finally pass the keypress to the view to
handle

ensures that even while a bauhaus control has focus the darkroom
preview shortcut still functions

also changes the preview shortcut to grab focus to the centre view so
it works when pressed twice in succession

### DIFF
--- a/src/bauhaus/bauhaus.c
+++ b/src/bauhaus/bauhaus.c
@@ -2005,7 +2005,7 @@ static gboolean dt_bauhaus_slider_key_press(GtkWidget *widget, GdkEventKey *even
     delta = -d->scale / 5.0f;
   }
 
-  if(!handled) return FALSE;
+  if(!handled) return dt_control_key_pressed(gdk_keyval_to_lower(event->keyval), dt_gui_translated_key_state(event));
 
   return dt_bauhaus_slider_add_delta_internal(widget, delta, event->state);
 }
@@ -2067,7 +2067,7 @@ static gboolean dt_bauhaus_combobox_key_press(GtkWidget *widget, GdkEventKey *ev
       dt_bauhaus_combobox_set(widget, new_pos);
     return TRUE;
   }
-  return FALSE;
+  return dt_control_key_pressed(gdk_keyval_to_lower(event->keyval), dt_gui_translated_key_state(event));
 }
 
 static gboolean dt_bauhaus_combobox_button_press(GtkWidget *widget, GdkEventButton *event, gpointer user_data)
--- a/src/views/darkroom.c
+++ b/src/views/darkroom.c
@@ -3491,6 +3491,7 @@ int key_pressed(dt_view_t *self, guint key, guint state)
       // we quit the active iop if any
       lib->full_preview_last_module = darktable.develop->gui_module;
       dt_iop_request_focus(NULL);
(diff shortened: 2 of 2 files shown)


elstoc:
This, along with my earlier commits, should mean that the preview and sticky preview shortcuts now work fairly consistently in the darkroom and lighttable views.

TurboGit:
Can you summarize what it fixes ? I can't find a scenario where a key was not handled and where it is now. Maybe I'm missing something.

elstoc:
If you change a bauhaus slider (click on it or scroll over it), the slider grabs keyboard focus. After that point, normal keyboard accelerators work fine but the preview accelerator does not, because nothing runs the key_pressed function in the darkroom view. Similarly for comboboxes.

TurboGit:
You mean for example in lt select and grad the quality slider in export module and then do a Alt+W to enable full preview ?

TurboGit:
I'm asking because this case does not work with or without the patch.

elstoc:
I think in the lighttable you need to be hovering the image for it to work anyway, so this fix doesn't make a difference. Try it in the darkroom.

TurboGit:
In darkroom with or without your patch I can click on Alt-1 to have 100% zoom after dragging sliders. So again I don't see a difference :) I have tried many things and cannot find a case where something not possible is possible now. A precise scenario would help I suppose.

elstoc:
Zoom close-up (Alt-1) is a regular keyboard shortcut and isn't built into the key_pressed function of darkroom.c so I would expect that to work with or without the patch. 

If you drag a slider (on, say, exposure) and then hit the shortcut for the full screen preview (W), it will only work with the patch. It may be that the preview is the only thing that's affected - the other keys (left,right,up,down) will be handled by the bauhaus handler.

TurboGit:
The `W` (full preview in darkroom) does not work for me when grabbing the exposure slider for example. I did test this explicitly, no luck. If I click on the center image and press `W` then the display is fine. Back to any slider and pressing `W` does not work. Now what could be the cause of the discrepancies between us ???

TurboGit:
Maybe a DE issue ? I'm using GNOME Shell 3.36.2 on GNU/Debian.

elstoc:
XFCE on ArchLinux. Bleeding edge distro so it'll have the latest version of everything. What you describe (clicking the centre image) is what I always used to have to do without this patch.

elstoc:
I haven't #included ``control.h`` in ``bauhaus.c``. Could that be it (though why it works for me I don't know)?

Mark-64:
Form me in Ubuntu 20.04 the `W` only works once in darkroom. Then the center image lose focus and I have to click on the image to make it work again.

elstoc:
@Mark-64 that's the second thing fixed by this patch.

elstoc:
I'm working in the dark a little here but I've tried something else if anyone can test.

elstoc:
@johnny-bit reports that this works now

johnny-bit:
after recent change the issue is fixed, i can press "W" whilst keeping mosuse button pressed/dragging slider even!

TurboGit:
Works as expected now ! Thanks.

PR #5319, opened 2020-06-05
Fixes float clipping limits that exceeded 1, avoids lost pixels with preview_downsampling<1.

### LINKED ISSUE #5306: down sampling crop&rotate issue
**Describe the bug**
When down sampling is activated (more visible with 1/4) the crop area cannot be set on top-left corner.

**To Reproduce**
0. Make sure downsampling is set to 1/4 (restart dt)
1. Go to darkroom
2. Activate crop&rotate
3. Try moving the top-left corner
4. See error, cannot move close

**Expected behavior**
It should be possible to set the crop area to contain the whole image. 

**Screenshots**

image

 - Darktable Version: master
 - OS: GNU/Debian

GrahamByrnes:
Partially resolves #5306.
A float variable clip_w was taking values slightly >1, leading to a 6016px wide image being opened as 6029px. It was then a somewhat random process if this would be exported with or without one column, viz 6015px.

This became evident with profile-downsampling =1/4 because of the greater error scaling. The error appears to have been the result of subtracting coordinates and dividing by the image width, which went from 335 to 335.75px wide. I'm not quite sure about this... I've included some fminf and fmaxf to keep everything safely inside [0,1]. None of this relates to operations on the image pixels, so it should have no appreciable effect on speed.

The issue of the asymmetric borders remains.

### DIFF
--- a/src/iop/clipping.c
+++ b/src/iop/clipping.c
@@ -639,20 +639,17 @@ static int _iop_clipping_set_max_clip(struct dt_iop_module_t *self)
   if(!dt_dev_distort_transform_plus(self->dev, self->dev->preview_pipe, self->iop_order, DT_DEV_TRANSFORM_DIR_FORW_EXCL, points, 4))
     return 0;
 
-  g->clip_max_x = points[0] / self->dev->preview_pipe->backbuf_width;
-  g->clip_max_y = points[1] / self->dev->preview_pipe->backbuf_height;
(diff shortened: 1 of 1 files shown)


TurboGit:
Good to me ! Thanks for all your time on this very difficult issue.

GrahamByrnes:
Did you find it also helped the clipping issue?

TurboGit:
I did not see a difference for the clipping issue on my side.

Issue #5129, opened 2020-05-19
expand single darkroom module preference when changing group


elstoc:
If you have the "expand single darkroom module at a time" preference checked this preference is not correctly observed when you change module groups.

If, for example, I perform the following steps

1. Expand and enable the 'exposure' module in the Basic group
2. Expand and enable the 'tone curve' module in the Tone group
3. Select the Active Modules group

Now, in the Active Modules group, both the 'exposure' and the 'tone curve' modules are expanded. That's not "expanding a single darkroom module at a time"! If I want to go back to only one module being expanded I have to collapse one of the expanded modules and then re-expand it. 

Using variations on this approach, it's fairly easy to get multiple modules to display in any of the groups. Similar issues are found if you click twice on a group to view all modules or use the search functionality.

I hesitate to just "fix" the issue, since it does look like it was intended by _someone_: ``imageop.c`` has

```
      if(m != module && dt_iop_shown_in_group(m, current_group))
      {
        all_other_closed = all_other_closed && !m->expanded;
        dt_iop_gui_set_single_expanded(m, FALSE);
      }
```

So the question is, what is the expected (or desired) behaviour here? Either it should be 'only one module expanded, regardless of group' or it should be 'one module may be expanded in each group'. But currently it's not quite one or the other.

My preference is for the first option - each time you expand a module, all other modules should collapse, in every group. It's a simple fix and it's behaviour that's easily understood and consistent with the preference description.

For the second option (one module per group), we would have to treat each group independently (probably regardless of the expansion preference setting) and have some way to record, for each group, which module instance(s) are expanded (in a multi-instance scenario just recording the module so name won't do), possibly also independently recording the scroll position in each (See #3085) and gracefully handling changes to the history stack. Then each time the user moves to a new group, reload the last state.

johnny-bit:
That particular code is member of "using shift click to expand more than one module despite single module expand being set"... and that leads to host of problems. you can view git-blame for that section of code, maybe history from it would shade some light on the issue.

so... that should be very complex handling on all corner cases :/ and seems like 2nd option would be the handling needed including respecting shift-click on active/favourite modules list...

elstoc:
Well essentially shift-click just temporarily reverses the behaviour set in the preference. So if you've checked the preference (expand single module) then shift-click will expand a module without collapsing the others. If you've unchecked the preference then shift-click will expand the clicked module and collapse the others (in that group).

[...]

PR #5322, opened 2020-06-05
regroup modules in 3 tabs : technical, grading and effects 

### LINKED ISSUE #5052: IOP Modules: Which Group?

aurelienpierre:
Favorite and active modules are still there:
Capture d’écran du 2020-06-05 21-55-55

* Fix #5052 and is based on top of #5314 
* Disables user customised module groups since it's probably not needed anymore.
* "technical" contains every module bounded by some physical reality: lens, sensor, in/out color spaces and various signal reconstructions (highlights, noise, etc.)
* "grading" contains every module to be used artistically and doing color and tones corrective and creative work (including chromatic adaptation / white balance)
* "effects" contain the weird kids (local contrast, sharpness, high/low-freq, retouch, liquify, etc.)
* the "technical" tab icon is supposed to be a lens making light rays converge. It's half a fail, feel free to improve.


TurboGit:
The new grouping is not a good idea to me. The result is that now each group contains lot more modules and it makes even harder to find the proper one. It also now delicate to have two open at the same time if they are separated by many other modules.

All in all, I'm not fond of this. Let see what others think about this.

elstoc:
Depends how big your actual working set of modules is I suppose and how well those modules spread across the available groups. I like the principle - I've been using darktable for a few years now and I still find myself hunting around among the groups for modules (and as such now resort to the search more often than not.

Would be useful to have a complete grouped list of modules on the OP, perhaps as an attachment, to save hunting through the 75 files that have been changed.

aurelienpierre:
The problem is : how do you intend to split modules in non-arbitrary groups ? Current groups are non-sense : most color module also work on tones, basic means absolutely nothing and mixes core technicalities will all-purpose stuff, corrections and effects are really debatable.

I'm all for small chunks, but not when the meaning gets lots. I don't even use groups anymore since there is the module search bars.

aurelienpierre:
> Would be useful to have a complete grouped list of modules on the OP, perhaps as an attachment, to save hunting through the 76 files that have been changed.

Technical (raw black/white point is hidden since it's the only one that doesn't fit on my screen):
Capture d’écran du 2020-06-05 23-45-46

Grading:
Capture d’écran du 2020-06-05 23-47-40

Effects:
Capture d’écran du 2020-06-05 23-48-24

TurboGit:
Another point to go in your direction Aurélien is that not all modules are made visible anyway. I have many hidden in the version I'm using as photographer. My comment was after having checked on my dev version where all modules are activated. Anyway, this needs more thoughts.

aurelienpierre:
Sure. We also have 3 different denoising modules, 2 of them being mostly useless now, given the performance of the profiled one.

elstoc:
Only problem I see is that my most used modules almost all sit in the technical group, with only a very few in the others. How about instead of that first group being 'favourites' it was intstead named 'custom' and you could **move** modules into it - so that the module was either in its default group **or** in the custom group but not both. That way putting items into the custom group would also have the effect of reducing the size of the other groups.

That would also have the side-effect of partially resolving #5129.

aurelienpierre:
> How about instead of that first group being 'favourites' it was intstead named 'custom' and you could **move** modules into it - so that the module was either in its default group **or** in the custom group but not both. That way putting items into the custom group would also have the effect of reducing the size of the other groups.

By moving, you mean drag and drop ? It's slightly more difficult to code, but doesn't change the feature at all, since favourite modules are already defined by users. Not sure if removing favourite modules from their original tab is a good idea.

Again, I get the visual design goal of balancing tabs, but let's not forget the meaning of the settings and module, in which many users are lost. A recurring demand of users is to know whether some particular module needs to be set with technical constraints in mind, or can be set freely. As an educator, this order allows me to answer that with : "whatever in grading can be used freely" but "be careful about technical modules". Because both worlds cohabitate in darktable, and they constraints are really different.

elstoc:
> I get the visual design goal of balancing tabs

It's not just about balancing tabs but about avoiding having to scroll to modules all the time.

MStraeten:
In my opinion the reordered groups are very useable from a workflow prospective. You’re usually operating from left to right
1. remove and push around pixels
2. get a start point for color grading 
3. bring out the colors 
4. tweak the appearance 
5. add additional stuff non existant in the raw. 

But 5 groups are too much, so it makes sense to combine 1+2 or 4+5. it can be discussed where some modules should be located.
A user will tidy up the lists soon or later by removing stuff never being used. 
Maybe there can be a different layouts defined that can be set like using tools/iop-layout.sh ...

AlicVB:
Well, in my personal case, I'm not sure about removing the favorite tab (whatever its name) : 
Here's my setup : mostly used modules in favorite : crop&rotate, filmic, color balance, exposure, channel mixer, tone equalizer, and I use only them 90% of the time.

But I've many more modules in other tabs for time to time use... and for those (rarely used) I never manage to remember in which tabs they are located... So why not reducing the number of tabs here.

I've tried to use the search box too (you seems to use it a lot @aurelienpierre ) but never manage to remember the exact name of the modules (french-english mix and stupid brain)

That said, imho, the best solution would be a more versatile approach, where we propose a nice set of disposition presets (tabs and content) but where we allow user to change them if he want. For the UI, the simplest will be to have it in a separate window. I've some idea for that, and I can have a try, if you want...

elstoc:
> I'm not sure about removing the favorite tab

I think every one is agreed (?) that the favourite tab (whatever we call it) has to remain if we're going to have fixed module groups.

> That said, imho, the best solution would be a more versatile approach, where we propose a nice set of disposition presets (tabs and content) but where we allow user to change them if he want.

I think that might be best, given that we'll probably never entirely agree on the "correct" placement of modules in the groups or even what groups there should be - there are so many that are subjective and different approaches to categorising them.

It would be great if the user could also choose from a list of the available darktable icons for their custom groups too. In order that new releases of darktable not provide new 'hidden' modules, a final group that contains 'ungrouped modules' might be an idea. Perhaps a new tab in the preferences screen could replace the 'more modules' module entirely?

Nilvus:
Favorites tab of course needs to be kept as it is by now (and the name is good). I rarely use tabs so I don't care about the choice but I found those new tabs more suitable and logical than actual modes. About scrolling, I think the best is to hide old modules on those tabs (they could always been shown (or not) with more modules. And that's also the purpose of modules presets in more modules: save modules we want on tabs (and possibility to save different settings).

By the way, new icons are gorgeous and much better than actual ones.

Mark-64:
IMO, whatever taxonomy of modules will always be controversial. Just one example:
3D LUT is "technical"? Yes and no, LUT are used since ages for color grading and creative purposes, so it could be equally well classifies in "grading".
We need to leave the user freedom to change the default grouping.

aurelienpierre:
LUT don't go in grading because there is no grading happening there. The LUT is precomputed against a reference and interpolated. It's an on/off module, you don't get to grade anything in that.

MStraeten:
Maybe it makes sense to group from a users prospective - a LUT has two main use cases: 
1. bring a flat image into a ‚editable‘ one - even more used in video editing than in photo editing since we are dealing with raw files instead of log profile jpgs
2. apply look via a LUT late in the pipe ( for this usecase it’s better after filmic in the pipe )

while first might be in a technical group (it brings the source image into a useable form) the second is better in a grading group since it it used to select and apply a look.

aurelienpierre:
"Technical" group means modules bounded to some physical/technical reality therefore constraints.

I don't care if it's for creative or corrective reasons. Notice "grading" modules are also both creative and corrective. It's not about that. Pretty much every module can be used both with creative or corrective intents, so that distinction makes no sense, and is indeed too relative and arbitrary to matter.

A LUT defines is a discretized mapping between 2 color spaces. To create it, you need to calibrate a sample against a reference, but to apply it, you need to check that the input color space of the LUT is the one expected by the particular LUT you are using (which was the color space used by the sample when building the LUT).

So, any LUT needs some careful checking because if it's not applied in the right colorspace, it won't produce the expected results. You are not free to use them as you please. Therefore -> technical.

Also, "grading" conveys the idea that modules can be graded, aka adjusted in intensity to match an expected intent. LUT don't have users parameters (well, the color checker one does, but the whole LUT maker program associated with this module cares only about chart reference, so it's borderline), it is generaly full-on or full-off. 

Grading is a synonym of "timing" in analog darkroom, because color development was achieved with color filters on the enlarger, and you could dial up or down the exposure time for each color filter, which is 1:1 equivalent to the color balance.

rawfiner:
I personally agree with this separation into technical grading and effect.
For technical icon, here is what I got (done on gimp on low resolution, just to give an idea)
technical
My idea was to illustrate chromatic aberration, and path of light through lens until the sensor

aurelienpierre:
@rawfiner That's too much color details for such a small icon though. I like the sensor, but, again… not sure about the amount of details at real size.

MStraeten:
found a linker error in modulegroups on osx:
```
[ 47%] Building C object src/libs/CMakeFiles/modulegroups.dir/modulegroups.c.o
[ 47%] Linking C shared module libmodulegroups.so
Undefined symbols for architecture x86_64:
  "_dtgtk_cairo_paint_modulegroup_grading", referenced from:
      _gui_init in modulegroups.c.o
  "_dtgtk_cairo_paint_modulegroup_technical", referenced from:
      _gui_init in modulegroups.c.o
ld: symbol(s) not found for architecture x86_64
clang: error: linker command failed with exit code 1 (use -v to see invocation)
```

aurelienpierre:
For some reason, the `paint.c` file is not in the commits. I will fix that.

aurelienpierre:
Should be fixed now.

So, did we reach a decision here ?

elstoc:
I think we decided that nobody can agree what the best solution is

johnny-bit:
I personally agree with this separation - makes most sense. Needs also touch manual since manual is divided in module groups, but that can be done post-merge of this (if this gets merged) as anoter PR or as a part of @paperdigits work.

elstoc:
My issue still remains the sheer volume of useful modules in that first tab compared to the others, not from a visual balancing point-of-view but just from fitting everything on the screen without scrolling - especially given that the modules are displayed in pipe order - so unless you know whereabouts in the pipe something is there's a lot of searching in a big list. 

Is there a reasonable logical way to split that first tab in two?

aurelienpierre:
@elstoc keep in mind most of the modules in the first tab are hidden by default:

* of the 4 denoising modules, only one is useful (profiled)
* invert should be deprecated for negadoctor at some point
* rotate & scale pixels never appear except for very old DSLR with rectangular pixels
* raw black/white point is useless outside of debugging, and hidden for users.

That's 6 modules going out of the view already.

elstoc:
I understand but I think I'd still have a group with 16 modules visible which makes finding stuff more problematic. I don't use most of the modules in the others so I think I'd end up with perhaps 10 in each.

Regarding my suggestion of being able to have modules in the favourites tab be removed from their default group we could possibly have four options instead of three for each module (hidden, in default group, in favourites, in both)

aurelienpierre:
> Regarding my suggestion of being able to have modules in the favourites tab be removed from their default group we could possibly have four options instead of three for each module (hidden, in default group, in favourites, in both).

The way I see it is that would be annoying more than anything. Tabs are supposed to represent thematic categories, therefore mean something. Having modules switching tabs is misleading in this regard.

Also adding more tabs means replacing scrolling with clicking. I personally don't see any benefit in there. There is already the module search bar to go directly to the module you want.

SoupyGit:
In a recent darktable post at Pixls outlining new features, there was a quote, "Discussions and proposals are also in progress to improve the management of modules in the darkroom (reorganization of tabs, improvement of presets, etc.)." I replied with some ideas, and was instructed to post them in this thread.

My 2 cents, for what its worth. A module currently can be adjusted in 3 different areas: Active Modules, Favourites, and Groups. It feels a little messy to me, and could perhaps be simplified to just Active Modules. 'Active Modules' would now no longer be only those active, but those adjusted at some point by the user. (eg. If you add Channel Mixer, then turn it off, it stays in module list, incase you want to turn it on again. To remove it from modules list, user can click the little squares and click 'delete'.) No longer showing only active modules, it could potentially be renamed 'Modules' or 'Pipeline'. All the other modules could be found via search, or via the 'More Modules' section underneath, and it is here where they could be grouped by category - much like how an OS start button has a list of applications, grouped by category). The other thing not best for user experience is having large scrolling bars in the modules list. To improve this, all the other current tabs (like Favourites and Groups) could be removed, replaced by an Adjustments/Properties tab and a Masking tab. So the Pipeline tab (formerly Active Modules) shows only the power button, module name and icons. Adjustments would be made in the Adjustments tab, and masks in the Masking tab. To me, that would be much simpler and neater. Favourites would need to be easily accessible, perhaps with a little button next to the 'reset parameters' and 'presets' icons. Clicking this would load all your favourite modules in Pipeline tab, with power button of each turned off by default.

There was pushback by one user on Pixls against my ideas because they liked working left to right through the Group tabs. Of course, if you are planning to change those tabs anyway, this will affect those kind of users. However, I don't see it as a major issue. With groups, instead of moving left to right through the tabs as now, users like that could simply move top to bottom through the more modules list (grouped by category, so it is not just one long list like it is now). It would only take 2 clicks (more modules, module) - which is the same it currently takes (module group, module) - and only slightly more mouse movement.

The above is idea 1. Below is idea 2, which is how categories might be grouped. This is not as big a deal to me personally as idea 1, as I know which modules I like to use in which order, and just type their name into search bar, but it may or may not be helpful.

*Colour Management (monitor profile, input color profile, output color profile)
*Sensor (Raw black/white point, Demosaic, Hot Pixel, Chromatic Aberration, Defringe)
*Transform (Orientation, Lens Profile, Perspective Correction, Crop and Rotate, Framing)
*Colour (White Balance, Channel Mixer, Color Balance, Color Zones, Color Correction, Color Look Up Table, Monochrome, Color Mapping, Split-Toning, Vibrance)
*Toning (Exposure, Midtones Shadows Highlights, Fimic, RGB Levels, RGB Curves, Tone Curve, Base Curve, Local Contrast, Basic Adjustments, Zone System, Tone Equaliser)
*Detail (Contrast Equaliser, Sharpen, High pass, Low pass, Soften, Denoise, Retouch, Grain)

I have not included every module, but this gives a basic idea. (They should be listed alphabetically, OR in pipe order, which I haven't done). Of course some modules are tricky to place because they could belong in multiple groups - but I don't think perfection of grouping is required. Just something that is easy to understand and use. Of course some of these modules will need to be enabled in Pipeline by default.

todd-prior:
If I could do it this way and it would not be for everyone I would only have 2 tabs active and favourites...I would call the other modules one of 2 ways search or for me something like the help sheet that pops up for shortcut keys would be awesome you could then group your modules into some of the categories suggested above or even allow custom ones created by the user  or alphabetized whatever but basically you would just hit a hot key and the list would pop up as it does to explain the hot keys but the modules would be selectable from that page...so then just two tabs for me....active and favourites ...you could even populate favourites manually as now and/or set a threshold for number of times used and have frequently used modules pop in...but that may be way overboard....

todd-prior:
I wonder about having the icons present a drop down list rather than open and entire panel. So basically a selection from technical or whatever categories are established would open a drop list and you select a module to use but don't have it open a whole panel of modules as it does now...Work from one panel and add the modules as you use them. Then you are not force to scroll way too much up and down past all the modules that you don't use in a given category...this combined with search for selection would stream line moving from module to module and the functions at the module level for copying and renaming activating etc would not need to change....Not sure I if I am explaining this correctly

TurboGit:
closing, new iop layout support has landed.

PR #5323, opened 2020-06-05
Dynamicly updating buttons on selected images module

johnny-bit:
Working of off code and inspiration from #5257 (and @TurboGit ) this is making buttons in selected images module very dynamic.

Although I do have a question: How to best get `DT_IMAGE_LOCAL_COPY` flag out of active images? the code I tried causes nasty busy spin when there's a group in collection.

Otherwise this works quite nicelly.

### DIFF
--- a/src/libs/image.c
+++ b/src/libs/image.c
@@ -54,6 +54,7 @@ typedef struct dt_lib_image_t
       *ratings_flag, *colors_flag, *metadata_flag, *geotags_flag, *tags_flag;
   GtkWidget *page1; // saved here for lua extensions
   int imageid;
+  guint timeout_handle;
 } dt_lib_image_t;
 
 typedef enum dt_lib_metadata_id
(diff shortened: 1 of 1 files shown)


TurboGit:
> Although I do have a question: How to best get `DT_IMAGE_LOCAL_COPY` flag out of active images? the code I tried causes nasty busy spin when there's a group in collection.

You need to lock the image and look at the flags. But this should not be done for a collection as it is probably too CPU intensive. I would simplify by activating the local copy or resync if there is at leas one image selected.

johnny-bit:
Done and matches what module requests of job control :)

Next on the menu: selection module :)

TurboGit:
On small glitch, the [group] button should be activated only if there is at least 2 selected images. For the [ungroup] button that's good as-is, we can un-group a single picture.

Apart from that all good! Again it feels as if dt is really alive.

@Nilvus as for #5257, I'd like you take on this one, Thanks.

Nilvus:
Will test this one this week-end.

johnny-bit:
> [group] button should be activated only if there is at least 2 selected images

Done! :)

TurboGit:
Looks good to me. Thanks! I'll wait @Nilvus review before merging.

Nilvus:
Good for me and the group button add is a good idea. Each day actually the UI is more and more polish (and much more "sexy"). Thanks @johnny-bit.

TurboGit:
Merging then. Thanks!

PR #5325, opened 2020-06-06
watermark: enable font substitution when watermark text is empty

hatsunearu:
the watermark module doesn't replace the font and watermark text substitutions unless the watermark text box is populated. i have removed that check.
this allows the watermark fonts to be used even when the watermark text is empty, and when the watermark text is empty, the text `$(WATERMARK_TEXT)` is substituted as an empty string instead of being unchanged

### DIFF
--- a/src/iop/watermark.c
+++ b/src/iop/watermark.c
@@ -430,14 +430,16 @@ static gchar *_watermark_get_svgdoc(dt_iop_module_t *self, dt_iop_watermark_data
     gchar buffer[1024];
 
     // substitute $(WATERMARK_TEXT)
-    g_strlcpy(buffer, data->text, sizeof(buffer));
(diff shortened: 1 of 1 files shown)


TurboGit:
> the watermark module doesn't replace the font and watermark text substitutions unless the watermark text box is populated.

Not a good move to me. The idea was that the $(WATERMARK_TEXT) was displayed to be able to place it, as documentation to write it's own .svg and moreover to have a feedback that something was happening.

What would be good is to keep $(WATERMARK_TEXT) visible when entry is empty and still allow the font to take effect.

hatsunearu:
I agree, the ergonomics is better in that case. I'll get a fix in later today.

TurboGit:
Looks good now. Thanks.

PR #5330, opened 2020-06-06
thumbnails : align icons on css margins

AlicVB:
This is the following of @aurelienpierre PR #5268 : we ensure to align icons like the image, which margins are defined in css.

### DIFF
--- a/src/dtgtk/thumbnail.c
+++ b/src/dtgtk/thumbnail.c
@@ -992,6 +992,7 @@ GtkWidget *dt_thumbnail_create_widget(dt_thumbnail_t *thumb)
     gtk_widget_show(thumb->w_image);
     gtk_overlay_add_overlay(GTK_OVERLAY(thumb->w_image_box), thumb->w_image);
     gtk_overlay_add_overlay(GTK_OVERLAY(thumb->w_main), thumb->w_image_box);
+    _thumb_retrieve_margins(thumb);
 
     // determine the overlays parents
     GtkWidget *overlays_parent = thumb->w_main;
@@ -1224,10 +1225,10 @@ static void _thumb_resize_overlays(dt_thumbnail_t *thumb)
   {
     gtk_widget_get_size_request(thumb->w_main, &width, &height);
 
-    const int margin_l = thumb->img_margin->left * width / 100;
-    const int margin_t = thumb->img_margin->top * height / 100;
-    const int margin_r = thumb->img_margin->right * width / 100;
-    const int margin_b = thumb->img_margin->bottom * height / 100;
+    const int margin_l = MAX(0, thumb->img_margin->left * width / 100);
+    const int margin_t = MAX(0, thumb->img_margin->top * height / 100);
+    const int margin_r = MAX(0, thumb->img_margin->right * width / 100);
+    const int margin_b = MAX(0, thumb->img_margin->bottom * height / 100);
 
     // we need to squeeze 5 stars + 1 reject + 1 colorlabels symbols on a thumbnail width
     // stars + reject having a width of 2 * r1 and spaced by r1 => 18 * r1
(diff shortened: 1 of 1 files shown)


Nilvus:
So, this will allow to let shown top icons without overriding? 

Another thing: I posted on compiled/develop channel in matrix a little glitch about timeout spinbutton in popover menu. The plus/minus buttons needs to be reduced. Height is higher than entry field. Seems it's not possible to do it correctly with CSS. Maybe that could be integrate to this PR if you can just adjust them (like all other spinbutton where you will see plus/minus ones aligned in height to entry field).

AlicVB:
> So, this will allow to let shown top icons without overriding?

nope, this is just a very little change : currently we align icons in thumbnail with hardcoded `0.03 * width` value. This correspond to the default css margins, but they can't change, so better use real image margins than hardcoded ones



> The plus/minus buttons needs to be reduced

yes, seen that. I'll have a look, but maybe for another PR, it'll simplify @TurboGit work if PR are smalls and don't mix multiple issues, I think

BTW : do you see my (late) reply in #5300 ?

Nilvus:
Ok, understand.

src/dtgtk/thumbnail.c
```diff
@@ -1224,17 +1225,22 @@ static void _thumb_resize_overlays(dt_thumbnail_t *thumb)
...
+    // inner margins are defined in css (margin_* values)
 
     // retrieves the size of the main icons in the top panel, thumbtable overlays shall not exceed that
-    const float r1 = fminf(max_size / 2.0f, 0.94 * width / 22.0f);
+    const float r1 = fminf(max_size / 2.0f, (width - margin_l - margin_r) / 22.0f);
 
     // file extension
-    gtk_widget_set_margin_top(thumb->w_ext, 0.03 * width);
-    gtk_widget_set_margin_start(thumb->w_ext, 0.03 * width);
+    gtk_widget_set_margin_top(thumb->w_ext, margin_t);
```
aurelienpierre, line 1242: it's not the same thing though. all margins were defined to 3% of the width, now it's 3% of the width for horizontal margins or 3% of the height for vertical ones. They can be different, so you will generally not be able to get an homogenous margin in thumbnail.

AlicVB: well homogeneous margins are the default in css for this reason. Now if user decide to use smaller/greater margins, I think we need to follow that settings for icons too.
Of course if an user decide to have different margins for left and top, he'll get not homogeneous margins, but that's what he has decide !

Nilvus: I agree with @AlicVB : If we set default same margins on CSS, I'm also more to not hardcode them. The user who want to tweak that do what he want and if he complain about not homogenous margins, it's his problem. Most important thing is that we, developers team, deliver a default homogenous margins.

Nilvus:
I just made a thing on my side in thumbnail.c: reduce reject, stars and colorlabels height. I find them too big by now. But reducing them made bottom right now too height. I'm not sure if your PR here will allow to adjust that only with CSS.

In thumbnail.c, they all set to 3.0 x 3.0. I reduce them to 2.5 x 2.5. Could you test on your side and see if margin change from that setting need other Gtk change. Not sure and not made real tests on that. And what do you think about that change?

Nilvus:
Oh, and feel free to integrate this change on this PR, that will be easy for me as I will need to rebase my PR (and I'm really bad on doing that).

AlicVB:
@Nilvus : I've tested. I agree for smaller stars. But the bottom margin in this case stay unchanged : it's always the bottom margin of the image, so I would let that unchanged and just reduce the icon size...
No problem for that here.

TurboGit:
I'm a bit lost with all the UI PR: #5303 , #5314 and this one. What is built on top of what ? Do we need all of them ? Some have add comments but no more actions, one is still Draft...

@aurelienpierre , @Nilvus : you are the two devs... let me know. thanks.

Nilvus:
@TurboGit: they are for me different things but complete each other. My PR is draft just because it's WIP by now and maybe some update would be made for this PR and needs CSS tweaks. We are discussed that on Matrix (french develop and compile channel).

About #5314 from @aurelienpierre: I'm just waiting the update for pen mask icon and position of color pickers and it should be good after that for this one.

AlicVB:
ok, I stop here. Promised

TurboGit:
All good and my side too. Thanks!

PR #5327, opened 2020-06-06
make select module buttons react to selection status

johnny-bit:
selecting filmroll or unaltered wouldn't raise selection changed signal

this commit fixes that

also adds very nice reactions to selection changes in select module :)

PR #5331, opened 2020-06-06
dynamic enable/disable of export button

johnny-bit:
also adds very nice reactions to images to act on change in export module :)

as "requested" by @TurboGit :)

Follows similar pattern as #5327 #5323 and #5257

### DIFF
--- a/src/libs/export.c
+++ b/src/libs/export.c
@@ -57,6 +57,7 @@ typedef struct dt_lib_export_t
   GtkWidget *export_masks;
   GtkWidget *metadata_button;
   char *metadata_export;
+  guint timeout_handle;
 } dt_lib_export_t;
 
 char *dt_lib_export_metadata_configuration_dialog(char *list, const gboolean ondisk);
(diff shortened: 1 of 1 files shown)


src/libs/export.c
```diff
@@ -83,6 +84,66 @@ uint32_t container(dt_lib_module_t *self)
...
+{
+  dt_lib_export_t *d = (dt_lib_export_t *)self->data;
+
+  GList *imgs = dt_view_get_images_to_act_on(TRUE);
+  const guint act_on_cnt = g_list_length(imgs);
+  g_list_free(imgs);
+
+  char *format_name = dt_conf_get_string(CONFIG_PREFIX "format_name");
+  char *storage_name = dt_conf_get_string(CONFIG_PREFIX "storage_name");
+  const int format_index = dt_imageio_get_index_of_format(dt_imageio_get_format_by_name(format_name));
```
TurboGit, line 97: Can we really have no format / storage selected ? Maybe first run, right ?

johnny-bit:
The same check is on export button 😄 i guess it is possible: not loaded
format /storage so we need to take that into account. Also - maybe some lua
script mess etc... No reason not to have that check IMO 😄

sob., 6 cze 2020, 19:35 użytkownik Pascal Obry <notifications@github.com>
napisał:

TurboGit:
Ok, makes sense. Thanks.

PR #5332, opened 2020-06-06
Make gradient sliders full themable custom widget

Mark-64:
After #5259 and #5282, with this PR the reworking of gradient sliders is complete.
The work is mostly under the hood, now gradient slider is a fully css themeable custom widget with all bells and whistles. In detail:
- make the Gtk custom widget sub-classing `GtkDrawingArea`, to reduce at minimum the class overrides and get a slimmer and more robust implementation
- move the boilerplate Gtk custom widget stuff to macro
- make the signals work, also fixing `blend_gui.c`
- fully implement css properties, including frame, background, status
- define two style classes for the widget (instead of simple widget names) and corresponding css changes
- cleanup and simplify code

This work could be the basis for a future reworking of bauhaus custom widgets


Nilvus:
Wow, so more things to test and check with CSS. Will all that PR for the UI, that's become hard to follow but it's just great to see that. Will test ASAP.

Mark-64:
Added public function to optionally give a name to the widget at creation time.
Applied to `blend_gui.c` and `relight.c`.
I believe we should use this technique for other widgets too, possibly giving a name to each individual instance, so to have the possibility of finer control in css.

Nilvus:
@Mark-64: all good for me and on what I see, nothing to work on with CSS.

Mark-64:
I guess I messed up again trying with rebase....
There was a conflict after the last PR by @Nilvus and I made the mistake to use GitHub to solve it, because I thought I had finished. But there was yet another bug to fix...
The result should be ok, though

TurboGit:
Looks ok, and works for me. Thanks.

PR #5335, opened 2020-06-06
Introspection data completion and gui refactoring for all IOPs

dterrahe:
This is now complete and ready for testing and merging. I'm replacing #5201 with this one, which has tidied up commits, tested with git rebase --exec=cmake --build to make sure it can be bisected. If there turns out to be a problem with an individual module, the specific commit can be reverted while I'm fixing it.

This work is based on the discussions in #4804 and #5097 and the additional work to expand the introspection framework kindly done by @houz .

The goal of this PR is to provide more or less complete introspection data for all IOP modules, which can form the basis for human readable export and import of module_params structures. Because the parameter limits that were previously only known to the actual sliders and dropdowns in the GUI can now be accessed by a file import module, validity of those parameters can be checked. @AxelG-DE

To make sure that the information that is added via comment tags in the params structure definition is in sync with what is actually used in the gui, new slider and dropdown creation methods are provided that link the two together. Similarly, a default parameter initialisation routine picks up the defaults from introspection and thereby also enforces consistency between a module reset and individual widget reset. And widget changes are written back to the module->params struct by a default callback function, making it unnecessary to create a boiler plate routine for each and every widget.

This leads to a significant reduction in code size and hopefully to an as significant increase in readability, maintainability and quality.

Two modules have not been dealt with yet:
- temperature, because @johnny-bit has a PR waiting that adds many improvements and it would be better to apply this framework to it after that has landed.
- negadoctor, because it extensively uses sliders linked to array elements, which dt_bauhaus_slider_from_params does not (yet) support. I'd like to get @aurelienpierre 's opinion on how to best proceed.

I assume this will go into 3.4; it is a massive patch that touches all modules. On the other hand, if it landed after 3.2, it might complicate backporting changes later. Basically, a lot of testing is required. Most of the work was very trivial boilerplate removal and if there's a problem with a complicated module, that can be backed out individually, since there is no requirement to use the new framework (except in the case of ashift.c which depends on the new curve approach, which in turn was required to support sliders with a negative multiplication factor).

Things to look out for:
- slider stepsize and number of digits changes; in cases where I changed them because they seemed excessively small/large I can easily put them back if preferred.
- incorrect soft and hard min/max ranges. Hopefully I haven't overlooked any in my own checks.
- parameters that don't immediately get updated. While culling unnecessary boilerplate it is possible that I have overlooked tricky special cases, although I have tried to capture all of these via the new gui_changed callback.
- Any dynamic module gui changes (un-/hiding sliders, tabs etc)

### DIFF
--- a/src/develop/imageop.c
+++ b/src/develop/imageop.c
@@ -2154,6 +2154,7 @@ GtkWidget *dt_iop_gui_get_expander(dt_iop_module_t *module)
   gtk_box_pack_start(GTK_BOX(iopw), module->widget, TRUE, TRUE, 0);
   dt_iop_gui_init_blending(iopw, module);
   gtk_widget_set_name(module->widget, "iop-plugin-ui-main");
+  dt_gui_add_help_link(module->widget, dt_get_help_url(module->op));
   gtk_widget_hide(iopw);
 
   module->expander = expander;
--- a/src/develop/imageop_gui.c
+++ b/src/develop/imageop_gui.c
@@ -184,6 +184,7 @@ GtkWidget *dt_bauhaus_slider_from_params(dt_iop_module_t *self, const char *para
     g_free(str);
   }
 
+  if(!self->widget) self->widget = gtk_box_new(GTK_ORIENTATION_VERTICAL, DT_BAUHAUS_SPACE);
   gtk_box_pack_start(GTK_BOX(self->widget), slider, FALSE, FALSE, 0);
 
   return slider;
@@ -233,6 +234,7 @@ GtkWidget *dt_bauhaus_combobox_from_params(dt_iop_module_t *self, const char *pa
     g_free(str);
   }
 
+  if(!self->widget) self->widget = gtk_box_new(GTK_ORIENTATION_VERTICAL, DT_BAUHAUS_SPACE);
   gtk_box_pack_start(GTK_BOX(self->widget), combobox, FALSE, FALSE, 0);
 
   return combobox;
--- a/src/develop/imageop_math.h
+++ b/src/develop/imageop_math.h
@@ -299,9 +299,9 @@ static inline float xoshiro256ss(uint64_t state[4])
 
 typedef enum dt_noise_distribution_t
 {
-  DT_NOISE_UNIFORM = 0,
-  DT_NOISE_GAUSSIAN = 1,
-  DT_NOISE_POISSONIAN = 2
+  DT_NOISE_UNIFORM = 0,   // $DESCRIPTION: "uniform"
+  DT_NOISE_GAUSSIAN = 1,  // $DESCRIPTION: "gaussian"
+  DT_NOISE_POISSONIAN = 2 // $DESCRIPTION: "poissonian"
 } dt_noise_distribution_t;
 
 
--- a/src/iop/ashift.c
+++ b/src/iop/ashift.c
@@ -4691,9 +4691,6 @@ void gui_init(struct dt_iop_module_t *self)
   g->lastx = g->lasty = -1.0f;
   g->crop_cx = g->crop_cy = 1.0f;
 
-  self->widget = gtk_box_new(GTK_ORIENTATION_VERTICAL, DT_BAUHAUS_SPACE);
-  dt_gui_add_help_link(self->widget, dt_get_help_url(self->op));
-
   g->rotation = dt_bauhaus_slider_from_params(self, "rotation");
   dt_bauhaus_slider_set_format(g->rotation, "%.2f°");
   dt_bauhaus_slider_set_soft_range(g->rotation, -ROTATION_RANGE, ROTATION_RANGE);
--- a/src/iop/atrous.c
+++ b/src/iop/atrous.c
@@ -1860,10 +1860,8 @@ void gui_init(struct dt_iop_module_t *self)
   c->timeout_handle = 0;
   c->x_move = -1;
   c->mouse_radius = 1.0 / BANDS;
+
   self->widget = gtk_box_new(GTK_ORIENTATION_VERTICAL, DT_BAUHAUS_SPACE);
-  dt_gui_add_help_link(self->widget, dt_get_help_url(self->op));
-  GtkWidget *vbox = gtk_box_new(GTK_ORIENTATION_VERTICAL, 0);
-  gtk_box_pack_start(GTK_BOX(self->widget), vbox, FALSE, FALSE, 0);
 
   c->channel_tabs = GTK_NOTEBOOK(gtk_notebook_new());
 
@@ -1878,7 +1876,7 @@ void gui_init(struct dt_iop_module_t *self)
   gtk_widget_show_all(GTK_WIDGET(gtk_notebook_get_nth_page(c->channel_tabs, c->channel)));
   gtk_notebook_set_current_page(GTK_NOTEBOOK(c->channel_tabs), c->channel);
 
-  gtk_box_pack_start(GTK_BOX(vbox), GTK_WIDGET(c->channel_tabs), FALSE, FALSE, 0);
+  gtk_box_pack_start(GTK_BOX(self->widget), GTK_WIDGET(c->channel_tabs), FALSE, FALSE, 0);
 
   g_signal_connect(G_OBJECT(c->channel_tabs), "switch_page", G_CALLBACK(tab_switch), self);
 
@@ -1886,7 +1884,7 @@ void gui_init(struct dt_iop_module_t *self)
 
   // graph
   c->area = GTK_DRAWING_AREA(dtgtk_drawing_area_new_with_aspect_ratio(0.75));
-  gtk_box_pack_start(GTK_BOX(vbox), GTK_WIDGET(c->area), TRUE, TRUE, 0);
+  gtk_box_pack_start(GTK_BOX(self->widget), GTK_WIDGET(c->area), TRUE, TRUE, 0);
 
   gtk_widget_add_events(GTK_WIDGET(c->area), GDK_POINTER_MOTION_MASK | GDK_POINTER_MOTION_HINT_MASK
                                              | GDK_BUTTON_PRESS_MASK | GDK_BUTTON_RELEASE_MASK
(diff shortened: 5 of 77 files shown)


aurelienpierre:
There is one problem in the process here: if sliders defaults are set from `module->default_params` and these are set at runtime in `reload_defaults()` will the slider default still be constants defined at compilation time ?

dterrahe:
You are correct that this could lead to a discrepancy between the defaults that are loaded by the module "reset parameters" button and by double-clicking on the slider itself to reset just the slider to its default, which would have been set up when the slider was created in gui_init.

This is not a change from what is already happening now in the old situation. So it is possible that I have not fixed all cases. I have however tried to move all constant defaults (i.e. not dependent on currently loaded image) to the init() function (in many cases just letting dt_iop_default_init do the job). Whenever constants were in an unnecessary reload_defaults implementation the module and slider defaults could end up not in sync. Sometimes the slider default was not based on the params passed to gui_init but hardcoded in the slider_new call (to fix such discrepancies).

The correct solution for image dependent defaults is to update the slider default directly if it is changed in reload_defaults (using dt_bauhaus_slider_set_default).

dterrahe:
BTW I'll be working on useless.c to expand it to more of a tutorial format rather than a template. 

Some sections obviously still always need to be included, but some sections are only required for specific situations so I think it would be useful to explain those situations and give simple examples, but not expect them to just be copied over. Existing modules are also being used as examples and copied from, which is why I spend time cleaning those up.

AxelG-DE:
As @dterrahe asked me for testing, I hope it is o.k. I put my findings here, one by one, means I will edit this reply continuosly. If that is not what is wanted, please advice me.

**Testing method:**
git fetch upstream pull/5335/head:introspection_test
git checkout introspection_test
build
start one session with 
/opt/darktable/bin/darktable --cachedir /tmp/intro/ --configdir ~/.config/darktable_introspection

and dt 3.02 with
darktable --cachedir /tmp/302/ --configdir ~/.config/darktable_3.0.2/

two separate folders with same pics, for the sake of separate xmps

**EXPOSURE**
t.b.c.

dterrahe:
@AxelG-DE That's fine, thank you!

dterrahe:
@aurelienpierre I've rebased on the new filmicRGB (twice). Please would you have a look to check that
1. everything still works as before (there shouldn't be any changes)
2. you are happy with the way I have restructured the code (especially in gui_init of course and to some extend reload_defaults). If you are, I think this gives me a lot to go on in filling out useless.c with more concrete guidance on how to approach certain gui aspects.

@houz It looks like there might be a small issue with enum introspection where two params fields have the same enum type (dt_iop_filmicrgb_curve_type_t in filmicrgb). The second one has the correct enum entries count, but the values pointer is NULL, I've worked around this for filmicrgb but would be good to fix.

aurelienpierre:
I will, but I have a shitload of stuff to finish for 3.2 first.

dterrahe:
Thanks @aurelienpierre, fully understood.

To be honest, I wanted to give @TurboGit at least the option of including this in 3.2, which is why I've been working hard to get it ready since I heard about the freeze. I'm afraid if this is merged straight after release of 3.2 it will create a lot of fun trying to backport fixes from master (given my experience with the rebase for filmicrgb).

@AxelG-DE has kindly been testing and as hoped/expected has found few regressions (none?) as this is mostly a like-for-like replacement/cleanup.

Possibly this could still be tested side-by-side a frozen 3.2 branch and a decision to merge could be made halfway through the freeze, since:
- no functionality changes
- no string changes

Leaving that totally up to @TurboGit and others, of course, but I'm committed to be very responsive if/when any problems are found. The color_picker changes can illustrate that, which was a similar exercise though smaller in size. And again, because this is split into individual commits per module, reverting part of it would be very easy. I could even now take out filmicrgb and other large ones if that is preferable, but of course those are the same that would likely cause the biggest backporting challenges.

johnny-bit:
It's good as it helps to remove boilerplate/repeat stuff and ues nice introspection.

My question would be... would it be possible to write unit tests for modules given the one for filmic as an example and via unit testing show no functional change? (integration tests too). That could very well lend itself for testing and futureproofing ;)

dterrahe:
> It's good as it helps to remove boilerplate/repeat stuff and ues nice introspection.

Also, it ensures introspection data is provided and consistent so it can be used for other purposes; this is why the whole project was started in the first place :-)

> My question would be... would it be possible to write unit tests for modules given the one for filmic as an example and via unit testing show no functional change? (integration tests too). That could very well lend itself for testing and futureproofing ;)

I have no idea what is involved here, but if this is not a mega project in itself (there are quite a few modules to go over, again) I would definitely be willing to look at it. Could you provide a few pointers?

johnny-bit:
I think https://github.com/darktable-org/darktable/tree/master/src/tests/unittests would be enough since it's just unit tests. need to check examples... I think I might write one for temperature.c since I kinda dug deep into it...

dterrahe:
If anybody has ideas how to write unittests for the gui part of modules...

Or how to print/export the widget hierarchy as it is displayed by GTK Inspector. That shows no detail information for bauhaus widgets, so would still not be very usefull for regression testing. Could bauhaus somehow provide this information, or would that require a full implementation of Gtk interfaces (and would that be undesirable for some reason)?

AxelG-DE:
> @AxelG-DE has kindly been testing and as hoped/expected has found few regressions (none?)

up to now still none!
I keep on testing unless you advice me differently. Soon I will have been thru my most freq. used modules, then I may appreciate a priority list from @dterrahe :-)

dterrahe:
> I keep on testing unless you advice me differently.

I am much appreciating it @AxelG-DE . @TurboGit hasn't really indicated if he is considering including it in 3.2. Hopefully your testing gives him confidence that wouldn't create any problems. But maybe you've done enough for now to achieve that goal. So please go ahead and give the boring testing of #5335 a break and see if you can find any bugs/curiosities in new functionality that _will_ make it into 3.2. Again, thanks a _lot_ for all the work you've done so far!

I've actually managed to find a very simple clean way to support arbitrary order in dropdowns, so I've now put "Lab-independend channels" in Tone Curve back in the position where it was before. 

This change isn't in this PR yet, because I'm waiting for relight.c to be un-conflicted before rebasing filmicrgb and exposure again.

johnny-bit:
I've done several small tests and I haven't noticed any drawbacks. "Everything just works" :)

TurboGit:
The plan just discussed with @dterrahe is to merge this in August just after 3.2 is released.

A plan is just a plan and I cannot take the decision alone. The plan I have proposed to @dterrahe is based on previous experiences with module order where I had to redo a good part of the pipe order because a huge bug found just before the release. Too much pressure.

But if we have at least two devs ready to test and fixes all issues found before August 10, then I'm ok to merge this today. What about this deal ?

johnny-bit:
I could help support this (although it's mostly @dterrahe + @houz's magic perl ;) ) I'd be all for merge since @AxelG-DE done very in-depth testing and couldn't find single difference in before/after behaviour (and knowing chats from him he has keen eye for any problems) so it shouldn't have any downsides. Plus merging in "now" gives time up to release for additional testing done by those who "ride on master" and not bother testing PRs :) And since it's a big change, mergin this after release would ban any fixes to be easily backproted...

AxelG-DE:
Thanks for the flowers

I'll test, what is needed. Just I don't count as   dev :-)

dterrahe:
@AxelG-DE The official term I believe is "contributor"

TurboGit:
@dterrahe : do you think you can fix the conflicts today ? if not let me know when you can do that. Thanks.

dterrahe:
Working on filmicrgb. Being careful not to break or miss anything and want to do a little bit of testing....

Hopefully an hour or so.

TurboGit:
Perfect !

dterrahe:
Sorry, a little bit longer.

I learned the hard way that if you have a couple of fixup! commits, you should "git rebase -i" those first against the current base before "git rebase -i upstream/master"-ing them.

It is now happily compiling again, but I want to do a few more tests.

dterrahe:
It looks like there may have been more subtle interactions between the crop&rotate speedup patch merged earlier and the fixes I made to the cropping sliders. It is probably not worse than it was before and nobody used the sliders anyway, but I need to look into fixing it again.

Other than that, and some weird new introspection issue that cropped up with dt_noise_distribution_t which means I'll have to convert that to dt_bauhaus_combobox_from_params at some later time, I think this is ready for more testing.

dterrahe:
Oops; last couple of fixes were not picked up by the rebase. Getting them in now.

dterrahe:
OK, Travis says it builds; time to ship!

Nilvus:
Milestone is actually for 3.4. If planning to be landed for 3.2, this need to change ;-)

dterrahe:
On the android app the close button is very unfortunately placed...

TurboGit:
Testing...

TurboGit:
All seems to work fine. Tested lot of iop and they responded as expected.

I have one concern about the translation. After building I have something like 50 fuzzy and 25 non translated strings. After looking at some occurrences it seems that all parameters (even if not having a GUI widget) gets recorded for translation.

Take for example the ashift four parameters cl, cr, ct, cb. There is no description for them and I suppose that in this case the strings in introspection_ashift.c should not have _(" ") around:

```
  .Float = {
    { DT_INTROSPECTION_TYPE_FLOAT, (char*)"float", (char*)"cr", (char*)"cr", N_((char*)"cr"), sizeof(((dt_iop_ashift_params_t*)NULL)->cr), G_STRUCT_OFFSET(dt_iop_ashift_params_t, cr), NULL },
    /*Min*/ -G_MAXFLOAT, /*Max*/ G_MAXFLOAT, /*Default*/ 1.0
  }
```

So maybe that's just a matter of adapting the Perl script ?

dterrahe:
You are right, there are a few more strings that get labeled as translatable. Not all of them; only those that have a $DEFAULT set. In those cases if the widget label is identical to the variable name, there is no need for the last bit of boiler plate like:
float exposure; // $MIN: -18.0 $MAX: 18.0 $DEFAULT: 0.0 $DESCRIPTION: "exposure"

I could exclude anything three characters or shorter from being marked translateable? That would mean some, like "mix", would need to be explicitly marked translateable if they are not already somewhere else (which might be surprising) and there would still be some (but I suspect few) falls positives.

I'll give that a try. Also open to other suggestions of course.

MStraeten:
builds fine on osx, runs fine (at least with the iops i regulary use)

TurboGit:
> I could exclude anything three characters or shorter from being marked translateable?

Why not make translatable only if there is a specific $DESCRIPTION ? (And so force a DESCRIPTION for any field that need to be presented in GUI).

PR #5336, opened 2020-06-06
Mods to allow real-time changes in preview_downsampling

GrahamByrnes:
I always thought the value of preview_downsampling should be selectable without a re-start.

This PR attempts to do that:

- The code from the initialisation section in `develop.c` has been moved to a less intrusive spot whare it is called only when gui_attached is true;
- There are three mods to darkroom.c, to capture` DT_SIGNAL_PREFERENCES_CHANGE`, write them to a float in `dt_view_t` and finally to write those changes into `dev->preview_downsampling`;
- There is the variable added to `views/view.h`

I'm somewhat out of my depth here. In particular, there is a note about keeping view.h and view_api.h in synch... but they don't seem particularly in synch to my unschooled eye. So I added it at the end: advice is welcome.

The overall thinking was that most of the time there would be no need for this, so there should be limited extra calculation: if there has been no change to preferences.c, everything stops at an initial test for a null pointer. Otoh, if you change prefs every time you open dt, there will be no salvation...

### DIFF
--- a/src/develop/develop.c
+++ b/src/develop/develop.c
@@ -104,6 +104,13 @@ void dt_dev_init(dt_develop_t *dev, int32_t gui_attached)
   else if(g_strcmp0(histogram_type, "logarithmic") == 0)
     dev->histogram_type = DT_DEV_HISTOGRAM_LOGARITHMIC;
   g_free(histogram_type);
+  gchar *preview_downsample = dt_conf_get_string("preview_downsampling");
+  dev->preview_downsampling =
+    (g_strcmp0(preview_downsample, "original") == 0) ? 1.0f
+    : (g_strcmp0(preview_downsample, "to 1/2")==0) ? 0.5f
+    : (g_strcmp0(preview_downsample, "to 1/3")==0) ? 1/3.0f
+    : 0.25f;
+  g_free(preview_downsample);
   dev->forms = NULL;
   dev->form_visible = NULL;
   dev->form_gui = NULL;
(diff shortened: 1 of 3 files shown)


GrahamByrnes:
In fact there are 2 things I'd like advice on, both relate to the smallest change :-)

- synching `view.h` and `view_api.h`, as mentioned above;
- initialising the dt_view_t structure that is used by `expose()` in `darkroom.c`, where I haven't been able to find any `default_params` feature as used for eg `dt_iop_module_t`.

TurboGit:
>     * initialising the dt_view_t structure that is used by `expose()` in `darkroom.c`, where I haven't been able to find any `default_params` feature as used for eg `dt_iop_module_t`.

I'm not sure about the question.

dt_view_t is initialized when reading all compiled views (.so files). This struct only contains some variables (size of window for example) and the set of hook initialized from the .so.

Each views (darkroom, lighttable, map...etc) are them selves initializing the data field in the dt_view_t struct as internal/private state data. This is done in the views's init() hook. For darkroom we have:

```
void init(dt_view_t *self)
{
  self->data = malloc(sizeof(dt_develop_t));
```
 
Does this clarify the situation ?

GrahamByrnes:
My thinking was to include `preview_sampling_update` in `dt_view_t` since it's really related to the current viewing of the images, and isn't specific to any one. In that case, I should add an initialisation of the new variable into `dt_view_load_module`.

On the other hand, as you say darkroom.c loads most of its data into `self->data` which is `dt_develop_t`, so possibly I could store the update in that... but I don't understand the volatility of `self->data`: if I change preview_downsampling then go in & out of darkroom several times, will the updated value be lost between images, as it somehow is in develop.c...

Anyway, these are just my current thoughts... if you can point out anything obviously wrong, that would help. Otherwise I'll experiment.

TurboGit:
> but I don't understand the volatility of self->data: if I change preview_downsampling then go in & out of darkroom several times, will the updated value be lost between images, as it somehow is in develop.c...

self-data is initialized **once** when _initializing_ darktable and clean-up when quitting darktable.

TurboGit:
@GrahamByrnes : this is intended for 3.4 (or later), right ?

GrahamByrnes:
Probably yes. I think it works but I haven't done exhaustive testing...

TurboGit:
@GrahamByrnes : This one also needs your attention as there is conflicts.

GrahamByrnes:
I'll close this and make a new version

PR #5340, opened 2020-06-07
thumb : allow to set tooltip visibility per size/mode

AlicVB:
this can be set in the overlays popup menu.
By default, tooltips are on except for "big" thumbnails and culling/preview

### DIFF
--- a/data/darktableconfig.xml.in
+++ b/data/darktableconfig.xml.in
@@ -712,6 +712,83 @@
     <longdescription/>
   </dtconfig>
   <dtconfig>
+    <name>plugins/lighttable/tooltips/0/0</name>
+    <type>bool</type>
+    <default>true</default>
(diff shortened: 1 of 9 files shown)


TurboGit:
Something buggy here, I had permanent overlay for size 1 and 2, now
- set lt to size 1
- open menu for overlay and set "show tooltip"

The lt now is set with overlay on mouse over and not permanent. This can be reproduced all the time.

AlicVB:
damn it... I think I've fixed that already... I'll have a look

AlicVB:
ok, so I've mess up with git again... There was 2 commits of this PR but the second one has been dropped I don't know where... I've finally recreated it from scratch.
Indeed without this one the PR broke everything... sorry !

TurboGit:
Works for me now. Thanks!

Issue #5341, opened 2020-06-07
[bisected] slider update incorrect at the max value (dragged to the right or left)


AxelG-DE:
**Describe the bug**

tbh is kinda comfort issue.
I found it while testing for PR #5335  (means that is not my standard habit in editing :smile: )

**To Reproduce**

**Expected behavior**

Same as e.g. in 3.0.2, where the value gets correctly set to max, no matter how fast you "throw" the mouse to the right

**bisect**

I think I did not mess anything thru bisecting but this result is not clear to me

```
5dc81c456abb0ab91c76fb6874b31e6b1772f70a is the first bad commit
commit 5dc81c456abb0ab91c76fb6874b31e6b1772f70a
Author: hanno@schwalm-bremen.de <hanno@schwalm-bremen.de>
Date:   Wed May 6 18:21:00 2020 +0200

    Make module in focus slightly outstanding
    
    Especially when working with more than one module in expanded state we might want
    to know by vision what module currently 'has the focus'.
    
    This pr adds this feature as a UI improvement.

 src/develop/imageop.c    |  4 ++--
 src/dtgtk/paint.c        | 12 +++++++++++-
 src/dtgtk/paint.h        |  3 ++-
 src/dtgtk/togglebutton.c | 16 ++++++++++------
 4 files changed, 25 insertions(+), 10 deletions(-)
```
@jenshannoschwalm could you anyway pls have a look? 

**Platform (please complete the following information):**
* darktable version : 3.1.0+1967~ga3679c0e9
* OS                : Linux - kernel 5.6.15-gentoo
* Distro            : Gentoo Base System release 2.6
* Processor         : Intel(R) Core(TM) i9-9900K CPU @ 3.60GHz
* Memory            : 32 GB (4 x 8 GB) + 5GB Swap
* Graphics card0    : GeForce GTX 1060 6GB
* Graphics card1    : GeForce RTX 2070 SUPER
* Graphics driver   : nvidia-drivers-440.82-r3
* OpenCL installed  : Yes (opencl-headers-2020.03.13-r1)
* OpenCL activated  : Yes
* Xorg              : xorg-server-1.20.7
* Desktop           : KDE 5
* GTK+              : gtk+-3.24.16
* gcc               : x86_64-pc-linux-gnu-9.3.0
* cflags            : CFLAGS="-march=native-O2-mtune=native-pipe"
* CMAKE_BUILD_TYPE  : "Release"

dterrahe:
[...]

PR #5345, opened 2020-06-08
Continue a slider drag even if it ends up in quad area

dterrahe:
I believe this solves #5341

This problem was introduced by 56ff796b0f900add27bb6b41d02cf2c5e7eec5af which made the quad area insensitive to drags.


TurboGit:
Works for me. Thanks!

PR #5346, opened 2020-06-08
activating module via shortcut adds to active group

elstoc:
when activating a module via a keyboard shortcut with the active module group shown, the activated module is shown in the active group

### DIFF
--- a/src/develop/develop.c
+++ b/src/develop/develop.c
@@ -2155,6 +2155,12 @@ void dt_dev_modulegroups_switch(dt_develop_t *dev, dt_iop_module_t *module)
     dev->proxy.modulegroups.switch_group(dev->proxy.modulegroups.module, module);
 }
 
+void dt_dev_modulegroups_update_visibility(dt_develop_t *dev)
+{
+  if(dev->proxy.modulegroups.module && dev->proxy.modulegroups.switch_group && dev->first_load == FALSE)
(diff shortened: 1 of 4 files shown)


elstoc:
Issue raised via https://discuss.pixls.us/t/active-module-list-not-correctly-updated/18644

elstoc:
I've just realised this solution also places the user into the active group if they're not already in it, which is not good.

elstoc:
ok I just fixed the first. I'll look into the second now!

elstoc:
Ok that should do it. Also now removes from active modules group when you deactivate a module.

TurboGit:
Works now but with a bad side effect. On the active module it was possible to deactivate and reactivate a module. This is an nice and fast way to check what a module is doing in an edit. Now when you deactivate a module it is immediately removed from active module.

elstoc:
Ok. I thought that might be a desired effect. Simple enough to change back.

elstoc:
Fixed now

TurboGit:
Not yet :) Add two modules say exposure and levels.
Go to active modules groups.
1. unselect levels - OK
2. unselect exposure - OK

All good both are there un-activated.

3. select back levels

And see that exposure has been removed from the group.

elstoc:
Well what fun.

elstoc:
I almost gave up on this one, but I _think_ I've finally fixed it now.

TurboGit:
Looks all good to me now. Thanks for the delicate work !

PR #5347, opened 2020-06-08
variables : add $(LABELS_ICONS) to use colorlabels in patterns

AlicVB:
only for info patterns, not for export due to markup contents

### DIFF
--- a/src/common/variables.c
+++ b/src/common/variables.c
@@ -165,6 +165,7 @@ static inline gboolean has_prefix(char **str, const char *prefix)
 static char *get_base_value(dt_variables_params_t *params, char **variable)
 {
   char *result = NULL;
+  gboolean escape = TRUE;
 
   struct tm exif_tm = params->data->have_exif_tm ? params->data->exif_tm : params->data->time;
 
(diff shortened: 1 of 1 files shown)


johnny-bit:
Would changing this for export pattern to color name be better? that way info's there just not supper pretty. Eg:
`<span foreground=\"#ee0000\">⚫ </span>` -> `(red ⚫)`?

that way if somebody uses `$(LABELS_ICONS)` instead of `$(LABELS)` the text won't be empty?

or maybe in case of export fall `$(LABELS_ICONS)` to `$(LABELS)`?
 (meaning: turn `else if(has_prefix(variable, "LABELS"))` into `else if(has_prefix(variable, "LABELS") || has_prefix(variable, "LABELS_ICONS"))` since previous condition should handle case of infos jobcode?

AlicVB:
@johnny-bit : good idea... done.
So now in case we use $(LABELS_ICONS) in export pattern, we fall back to $(LABELS)

src/common/variables.c
```diff
@@ -371,7 +372,46 @@ static char *get_base_value(dt_variables_params_t *params, char **variable)
...
-  else if(has_prefix(variable, "LABELS"))
+  else if(has_prefix(variable, "LABELS_ICONS") && g_strcmp0(params->jobcode, "infos") == 0)
+  {
+    escape = FALSE;
+    GList *res = dt_metadata_get(params->imgid, "Xmp.darktable.colorlabels", NULL);
+    res = g_list_first(res);
+    gchar *txt = NULL;
+    if(res != NULL)
+    {
+      GList *labels = NULL;
```
TurboGit, line 383: can be removed as not used.

src/common/variables.c
```diff
@@ -371,7 +372,46 @@ static char *get_base_value(dt_variables_params_t *params, char **variable)
...
+        {
+          txt = dt_util_dstrcat(txt, "<span foreground=\"#0000ee\">⚫ </span>");
+        }
+        else if(g_strcmp0(lb, "purple") == 0)
+        {
+          txt = dt_util_dstrcat(txt, "<span foreground=\"#ee00ee\">⚫ </span>");
+        }
+      } while((res = g_list_next(res)) != NULL);
+      result = g_strdup(txt);
+      g_list_free(labels);
```
TurboGit, line 409: likewise

src/common/variables.c
AlicVB: you're right...

src/common/variables.c
AlicVB: done

TurboGit:
All good. Thanks.

GLLM:
Hi !
I've tested this in Darkroom infoline ( in dt 3.1.0~git2270.f42aeca9a on Linux Kubuntu 20.04) and I see this
image
i.e. the dots are not in color.
Is this expected ?
Thanks
GLLM

johnny-bit:
Huh, interesting - I can confirm @GLLM's finding! `$(LABELS_ICONS)` doesn't work at all for me :/ @GLLM - can you open an issue for it?

GLLM:
Sure, will do it right away ! :)

PR #5350, opened 2020-06-08
ellipsize buttons in snapshot and history stack modules

elstoc:
allows LHS panel of darkroom to be made smaller where modules with long
names are present in history and snapshot modules


TurboGit:
Works for me! Thanks.

PR #5352, opened 2020-06-08
ellipsize created shapes label in mask manager

elstoc:


elstoc:
I didn't try it on an image with shapes so I didn't spot that. Sounds reasonable.

elstoc:
Actually I can't see the number of shapes. Are we looking at the same place?

TurboGit:
@elstoc : mask manager is fully broken anyway :( See #5359.

elstoc:
Ah. Well I've made the change to ellipsize in the middle instead if you want me to push it. I don't spend much time in mask manager so it's not a big deal for me. I was just going through the modules to find which ones cause the header buttons to disappear off the side of the screen when you shrink the panel.

TurboGit:
I see, this is only for the label, not the shapes name inside the mask manager. All good then.

PR #5355, opened 2020-06-08
handle multiple styles in styles module (action-wise) and dynamically update buttons

### LINKED ISSUE #3705: Delete styles
Description of the problem.
It is related to the list of styles that appears on the light table, there are several buttons to interact with the list that exists, one of them is to delete, which only allows you to do it one by one and in the case of wanting to delete several styles at once, it is not possible.

That I would like to meet.

I would like to be able to choose several styles at once, be they 2 or 35 styles and be able to eliminate them at once, without having to do it tediously one at a time.

Additional context

It would not be necessary to add more buttons, it would only be necessary to be able to choose several styles in the container box by pressing the "shift" key, and in the case of having several styles selected, the option of applying the style to the photograph would be canceled.

Thanks

johnny-bit:
have you ever wanted to apply multiple styles at once?
or maybe delete multiple styles?
or export multiple stypes?

now you can!

fixes #3705 (and couple wishes)

### DIFF
--- a/src/libs/styles.c
+++ b/src/libs/styles.c
@@ -420,7 +420,7 @@ void gui_init(dt_lib_module_t *self)
   gtk_tree_view_column_pack_start(col, renderer, TRUE);
   gtk_tree_view_column_add_attribute(col, renderer, "text", DT_STYLES_COL_NAME);
 
-  gtk_tree_selection_set_mode(gtk_tree_view_get_selection(GTK_TREE_VIEW(d->tree)), GTK_SELECTION_SINGLE);
+  gtk_tree_selection_set_mode(gtk_tree_view_get_selection(GTK_TREE_VIEW(d->tree)), GTK_SELECTION_MULTIPLE);
   gtk_tree_view_set_model(GTK_TREE_VIEW(d->tree), GTK_TREE_MODEL(treestore));
   g_object_unref(treestore);
 
--- a/src/libs/styles.c
+++ b/src/libs/styles.c
@@ -222,20 +222,13 @@ static void _styles_row_activated_callback(GtkTreeView *view, GtkTreePath *path,
   g_list_free(list);
 }
 
-static void apply_clicked(GtkWidget *w, gpointer user_data)
+// get list of style names from selection
+// free returned list with g_list_free_full(list, g_free)
+GList* _get_selected_style_names(GList* selected_styles, GtkTreeModel *model)
 {
-  dt_lib_styles_t *d = (dt_lib_styles_t *)user_data;
-  GtkTreeSelection *selection = gtk_tree_view_get_selection(GTK_TREE_VIEW(d->tree));
-
-  if(gtk_tree_selection_count_selected_rows(selection) == 0) return;
-
   GtkTreeIter iter;
-  GtkTreeModel *model= gtk_tree_view_get_model(d->tree);
-
-  GList *selected_styles = gtk_tree_selection_get_selected_rows(selection, &model);
   GList *style_names = NULL;
-  GList *style = NULL;
-  for (style = selected_styles; style != NULL; style = style->next)
+  for (GList *style = selected_styles; style != NULL; style = style->next)
   {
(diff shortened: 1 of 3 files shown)


src/libs/styles.c
```diff
@@ -232,93 +269,142 @@ static void create_clicked(GtkWidget *w, gpointer user_data)
...
+    for (style = style_names; style != NULL; style = style->next)
     {
-      dt_styles_delete_by_name(name);
-      _gui_styles_update_view(d);
+      dt_styles_delete_by_name((char*)style->data);
     }
+    _gui_styles_update_view(d);
   }
+
+  g_list_free(style_names);
```
TurboGit, line 357: g_list_free_ful()

src/libs/styles.c
```diff
@@ -232,93 +269,142 @@ static void create_clicked(GtkWidget *w, gpointer user_data)
...
-  dt_lib_styles_t *d = (dt_lib_styles_t *)user_data;
-  const char *name = get_style_name(d);
+  GtkTreeIter iter;
+  GtkTreeModel *model= gtk_tree_view_get_model(d->tree);
 
-  if(name)
+  GList *selected_styles = gtk_tree_selection_get_selected_rows(selection, &model);
+  GList *style_names = NULL;
+  GList *style = NULL;
+  for (style = selected_styles; style != NULL; style = style->next)
```
TurboGit, line 315: for(GList *style =

src/libs/styles.c
```diff
@@ -219,6 +222,40 @@ static void _styles_row_activated_callback(GtkTreeView *view, GtkTreePath *path,
...
+
+  if(gtk_tree_selection_count_selected_rows(selection) == 0) return;
+
+  GtkTreeIter iter;
+  GtkTreeModel *model= gtk_tree_view_get_model(d->tree);
+
+  GList *selected_styles = gtk_tree_selection_get_selected_rows(selection, &model);
+  GList *style_names = NULL;
+  GList *style = NULL;
+  for (style = selected_styles; style != NULL; style = style->next)
```
TurboGit, line 238: for(GList *style = ...

src/libs/styles.c
```diff
@@ -232,93 +269,142 @@ static void create_clicked(GtkWidget *w, gpointer user_data)
...
```
TurboGit, line 374: for(GList *style = ...

TurboGit: BTW, this very same loop is repeated in three places, probably better to put this code into a routine.

GList *_get_selected_styles(...).

You'll probably have to not   g_value_unset(&value) inside this routine and so use a g_list_free_full() for the styles_names list. See below.

src/libs/styles.c
```diff
@@ -232,93 +269,142 @@ static void create_clicked(GtkWidget *w, gpointer user_data)
...
-      g_free(filedir);
+      dt_styles_save_to_file((char*)style->data, filedir, FALSE);
+      dt_control_log(_("style %s was successfully saved"), (char*)style->data);
     }
-    g_free(name);
-    gtk_widget_destroy(filechooser);
+    g_free(filedir);
   }
+  gtk_widget_destroy(filechooser);
+  g_list_free(style_names);
```
TurboGit, line 407: here: g_list_free_full()

TurboGit:
I have scheduled this for 3.2, seems safe enough, sounds fine to you ?

johnny-bit:
Yes! I'll do it soon! I'm fiddling with tagging module and same funzies for enabling/disabling buttons (it's the last one I promise! i'm not touching geo one!)

src/libs/styles.c
johnny-bit: the g_value_unset is needed since the list isn't of GValue items but of gchar* items :)

johnny-bit:
Done!

Additionally I've added some log messages so users know that apply went well ;) I found out I rather liked the fact that i can see apply, even if effects on thumbnails are negligible :)

PR #5362, opened 2020-06-09
po/fix masks

TurboGit:


aurelienpierre:
Works here.

jenshannoschwalm:
I just saw the open issue.

Do you really think this fix is the best one?
1. Doesn't this mean a mask visualizer might be shown even if the module is not "active"? So you might be confused ...

2. Should we better show after test for the existing condition **or** the mask manager being active?

TurboGit:
@jenshannoschwalm : ok on both count ! the mask manager is always enabled so the check need to be refined.

jenshannoschwalm:
mask manager widget extended?

TurboGit:
@jenshannoschwalm : new version should be better. ok for you ?

jenshannoschwalm:
Looks good for me.

PR #5367, opened 2020-06-09
dynamic tagging module buttons

johnny-bit:
Cutsey little dynamically updating tagging module buttons.

Plus little bonus: due to postponed update on mouse over less cpu cycles done enumerating image tags (as if that matters ;) )


TurboGit:
All good to me. Thanks!

PR #5368, opened 2020-06-09
bauhaus: ellipsize slider names and give combobox extra space

elstoc:
prevents slider names colliding with values when panes are made narrower

also adds a bit of extra space between label and combobox when panes are
made narrower

### DIFF
--- a/src/bauhaus/bauhaus.c
+++ b/src/bauhaus/bauhaus.c
@@ -100,13 +100,13 @@ static GdkRGBA * default_color_assign()
 }
 
 
-static int show_pango_text(dt_bauhaus_widget_t *w, GtkStyleContext *context, cairo_t *cr, char *text, float x_pos, float y_pos, float max_width, gboolean right_aligned)
+static int show_pango_text(dt_bauhaus_widget_t *w, GtkStyleContext *context, cairo_t *cr, char *text, float x_pos, float y_pos, float max_width, gboolean right_aligned, gboolean calc_only)
 {
   PangoLayout *layout = pango_cairo_create_layout(cr);
 
   if(max_width > 0)
   {
-    pango_layout_set_ellipsize(layout, PANGO_ELLIPSIZE_MIDDLE);
+    pango_layout_set_ellipsize(layout, PANGO_ELLIPSIZE_END);
     pango_layout_set_width(layout, (int)(PANGO_SCALE * max_width + 0.5f));
   }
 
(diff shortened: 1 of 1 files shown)


elstoc:
filmicrgb options panel is a good one to check on as it exhibits both issues

elstoc:
It would be nice to do something similar with comboboxes and their labels but sharing the space between them and deciding which to ellipsize is not so easy. Currently the box is ellipsized but the label is not, which seems the wrong way round. Could we split them 50:50?

elstoc:
Thinking about it more, we could calculate the size of the label, giving it a maximum allocation of, say, 60% of the total width, and then allocate the rest to the combobox.

elstoc:
@TurboGit I notice you've put this as WIP. Does that mean you'd like me to try out my suggestion above?

TurboGit:
> @TurboGit I notice you've put this as WIP. Does that mean you'd like me to try out my suggestion above?

That was my idea but if you prefer to have this merged first and work on the idea on another PR it is ok to me. Just let me know.

elstoc:
No that's fine - I'm experimenting now. I'm also considering changing so that it ellipsizes the end rather than the middle, which would match how most of the buttons work.

But the position of the ellipsis is hard-coded into show_pango_text so it's either one or the other. Personally I think end makes more sense for most things.

elstoc:
Ok I've pushed another attempt. It's really subjective so needs some feedback. I'm not sure there's an easy way to make comboboxes function consistently unless you could calculate widths of both the label and the box before drawing them and then only divide up the space if the total exceeded the available width.

Suggestions welcome.

TurboGit:
Working good on my side! Another nice improvement. Thanks.

dterrahe:
Not a new change, but since I'm now looking at this...

Shouldn't the pango_font_description_copy in show_pango_text have a corresponding pango_font_description_free? This looks like a memory leak to me.

https://github.com/elstoc/darktable/blob/38140b087c461e990743cda9a5dc266231ce9a0f/src/bauhaus/bauhaus.c#L122

Or is there any drawback to not making a copy at all (the font isn't manipulated in any way) and simply calling

  pango_layout_set_font_description(layout, darktable.bauhaus->pango_font_desc);

directly? I've compiled this and it doesn't crash, but not submitting a PR since I don't want to pretend I fully understand ownership rules...

elstoc:
It's not freed as part of ``g_object_unref(layout)``? I admit I don't understand it all either.

dterrahe:
The pango docs don't say that https://developer.gnome.org/pango/stable/pango-Layout-Objects.html#pango-layout-set-font-description transfers ownership of the description structure and https://developer.gnome.org/pango/stable/pango-Fonts.html#pango-font-description-copy explicitly says it should be freed after use. Everywhere else where this is used in darktable there is a call to free and I've now confirmed that this code is indeed leaking memory (slowly) and submitted a tiny PR to fix it.

This fits nicely in the standard programming pattern of noticing memory leaks within hours of a major release. Sorry @TurboGit. It is a tiny leak only. Drip, really...

PR #5369, opened 2020-06-09
Allows more precise aspect ratios in crop & rotate

GrahamByrnes:
This attempts to fix a problem where crop & rotate doesn't allow the original aspect ratio when selecting "original image", if using preview_downsampling < 1 (and possibly otherwise).

It now calculates the aspect ratio from the dimensions of the original image: hence eg a 6016x4000 image is not forcibly reduced to 6016x3996.

### DIFF
--- a/src/iop/clipping.c
+++ b/src/iop/clipping.c
@@ -1352,13 +1352,28 @@ void cleanup_pipe(struct dt_iop_module_t *self, dt_dev_pixelpipe_t *pipe, dt_dev
   piece->data = NULL;
 }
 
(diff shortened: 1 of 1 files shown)


jenshannoschwalm:
Yeah. This seems right. I am also working on this ...

GrahamByrnes:
If you can find why the window is offset, I'll send flowers :-D

GrahamByrnes:
Great. May have a fix for the offset image border in a few minutes...

TurboGit:
@GrahamByrnes : Is that true ! A fix for this infamous bug!!!! That sounds great.

PR #5372, opened 2020-06-10
Improving suport for sensor size and ratio

jenshannoschwalm:
There have been several issues related to exact sensor size and image ratio, also the
available information in "image information" was misleading.

Done here:

1. the dt_image_t struct also has p_width & p_height, both are not kept in the database
   but are filled when developing via rawspeed and rawprepare.
   They hold the size of **used** data for the sensor... 
2. Introduced a new signal `DT_SIGNAL_METADATA_UPDATE` that notes the "image information"
   display about changed data. It's sent if a modified crop has been detected via rawspeed
   or rawprepare for now. I looked through the other signals, `DT_SIGNAL_IMAGE_INFO_CHANGED` or `DT_SIGNAL_TAG_CHANGED` were concidered but did not seem to be well suited.
3. Metadata_view looks for any crop on sensor data and if there is, it shows current size
   and original like "4000 (4016)"
4. Also in metadata_view, as we don't calculate verified_size right now because of performance reasons it's not shown with numbers but as non-existing.
5. In image.c we have `double dt_image_get_sensor_ratio(const struct dt_image_t *img)`
   that can be used to get the exact ratio of raw data.
   Atm this is not further used because of freezing date coming soon ...

@GrahamByrnes might this help?

### DIFF
--- a/src/common/image.c
+++ b/src/common/image.c
@@ -665,6 +665,24 @@ void dt_image_flip(const int32_t imgid, const int32_t cw)
                  dt_history_snapshot_undo_pop, dt_history_snapshot_undo_lt_history_data_free);
 }
 
+/* About the image size ratio 
+   It has been calculated from the exif data width&height, this is not exact as we crop
+   the sensor data in most cases for raws.
+   This is managed by
+   rawspeed - knowing about default crops
+   rawprepare - modify the defaults
+
+   The database does **not** hold the cropped width & height so we fill the data
+   when starting to develop.
+*/
+double dt_image_get_sensor_ratio(const struct dt_image_t *img)
+{
+  if(img->p_height >0)
+    return (double)img->p_width / (double)img->p_height;
+
+  return (double)img->width / (double)img->height;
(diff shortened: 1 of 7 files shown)


GrahamByrnes:
@jenshannoschwalm this looks cool even if I'm not sure how I would use it right away.

I've been thinking more about @aurelienpierre 's suggestion of storing image size as fraction pairs: I'm not sure that this is the right solution, but something that allows the original image size to be reconstructed from any of its downsampled, or eventually upsampled versions. So if we are working with an image that is currently 335x222, to know that should re-scale to 1442x887 or even 9000x6011 for upsized printing... to avoid the loss of information as we pass between different image sizes. 

Probably float is enough: 1/(2^27) in the mantissa should be enough to deal with 1/(sensor linear dimension) for some time to come :-) It would just require some thinking about a protocol for extracting the integer dimensions (eg ceil, floor) to minimise the abuse of interpolation.

That though is a longer term project...

PR #5376, opened 2020-06-10
histogram: reverse resize scroll direction

elstoc:
amend the histogram resize so that scrolling up shrinks and scrolling
down grows (for better consistency with other modules e.g. collect
images and image information)


TurboGit:
Indeed, good catch. Merging this fix. Thanks.

PR #5384, opened 2020-06-10
New css fixes

Nilvus:
- Should improve visibility on text infos in thumbnails (for all themes). @TurboGit: hope it will be good for you. Just consider that I've kept the idea to have better visibility when hovering than without but have good one without hovering anyway. Those changes needs to update accordingly background gradient when hovering thumbnails.
- Improve disabled buttons since new dynamic buttons added by @johnny-bit 
- Fix too big color labels icons in bottom left in lighttable view
- Fix too big buttons in duplicate module in darkroom
- Fix height on add button in picker module
- Hide top icons in thumbnails on duplicate module as it's not readable and ugly with them
- Set back to -1px finally top margin of extension name as -3px was not a good idea I propose to @alicvb.
- And finally, set visually new show tooltip on overlay popover menu

### DIFF
--- a/data/themes/darktable-elegant-dark.css
+++ b/data/themes/darktable-elegant-dark.css
@@ -71,7 +71,8 @@
 @define-color field_hover_fg @grey_20;
 
 /* Tooltips and contextual helpers */
-@define-color tooltip_bg_color @grey_20;
+@define-color tooltip_bg_color @grey_25;
+@define-color tooltip_fg_color @grey_75;
 @define-color log_fg_color @grey_90;
 
 /* Views */
--- a/data/themes/darktable-elegant-dark.css
+++ b/data/themes/darktable-elegant-dark.css
@@ -83,11 +83,12 @@
 @define-color lighttable_bg_font_color @grey_85;
 
 /* Lighttable and film-strip */
-@define-color thumbnail_font_color @grey_35;
 @define-color thumbnail_bg_color @grey_55;
-@define-color thumbnail_infos_color @grey_30;
+@define-color thumbnail_star_bg_color @grey_40;
+@define-color thumbnail_fg_color @grey_60;
 @define-color thumbnail_selected_bg_color @grey_65;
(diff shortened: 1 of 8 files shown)


Nilvus:
> Clearly an overall improvement. There is one point where I think we can do better. In many tooltip we have lot of text for explanation. I find the reading still difficult and I'm just wondering if we want to keep the condensed fonts there ? For reading a long text I think a non condensed font and maybe a bit bigger will help. This won't change really the UI as the tooltip are only transient.
> 
> What do you think ?

Better font and color yes, it's a good idea. But not increase size, even a little, that will make some tooltips take too much space, even if it's for a short time.

Last commit update that. Your idea make me find a bug in CSS code (color line on tooltip/menu/combobox pack not applied as color is not set for a main widget like those. So the color set for tooltip text was not the one applied. This commit fix that and by using Roboto Regular font, tooltip label are much better on all themes now.

One thing I'm not sure is about the background_jobevent_box. I see it nowhere on the UI and on what I see on Gtk code, I think it's not really set. The gtk_widget_set_name is not right after a widget. So probably never used. I remove them on next commit and made a little cleanup (2 eventbox lines written 2 times)

Nilvus:
@TurboGit: ready to review and merged if it's all good. Could you just confirm me that I'm right with last commit and that you could confirm written that way on backgroundjobs.c file, the background_job_eventbox name set is not used, so could be removed safely?

If not, please confirm if doing a git revert on my side then a git push will remove that commit.

TurboGit:
@Nilvus :background_job_eventbox is used and works well. In CSS set the background color to red and see that the box (bottom left) when you export has a red background. Remove the definition in the CSS for `#background_job_eventbox,` and the background is not used anymore.

So this was used.

Otherwise the tooltip are now far more readable. Perfect to me!

Nilvus:
@TurboGit: so commit revert back and little cleanup. Should be good now to review again.

TurboGit:
Works for me, all seems good now. Thanks.

PR #5388, opened 2020-06-10
filmicv4 : add a slight amount of gaussian noise in highlights

aurelienpierre:
helps reducing the "synthetic effect" of bloomed highlights

without: (master, 10 iterations of high quality reconstruction)
DSC_0552_14

with: (this PR)
DSC_0552_19

with: (this PR, 1 iteration of high quality reconstruction)
DSC_0552_20

### DIFF
--- a/src/develop/imageop_math.h
+++ b/src/develop/imageop_math.h
@@ -224,12 +224,100 @@ static inline int FCxtrans(const int row, const int col, const dt_iop_roi_t *con
   return xtrans[irow % 6][icol % 6];
 }
 
+
+#ifdef _OPENMP
+#pragma omp declare simd
+#endif
 static inline int fcol(const int row, const int col, const uint32_t filters, const uint8_t (*const xtrans)[6])
 {
   if(filters == 9)
     return FCxtrans(row, col, NULL, xtrans);
   else
     return FC(row, col, filters);
+}
+
+
+#ifdef _OPENMP
+#pragma omp declare simd
+#endif
+static inline uint64_t splitmix64(const uint64_t seed)
+{
+  // fast random number generator
+  // reference : http://prng.di.unimi.it/splitmix64.c
+  uint64_t result = (seed ^ (seed >> 30)) * 0xBF58476D1CE4E5B9;
+  result = (result ^ (result >> 27)) * 0x94D049BB133111EB;
+  return result ^ (result >> 31);
+}
+
+
+#ifdef _OPENMP
+#pragma omp declare simd aligned(state:64)
+#endif
+static inline void xoshiro256_init(uint64_t seed, uint64_t state[4])
+{
+  // Init the xoshiro256 random generator
+  uint64_t tmp = splitmix64(seed);
+  state[0] = (uint32_t)tmp;
+  state[1] = (uint32_t)(tmp >> 32);
+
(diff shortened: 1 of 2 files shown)


aurelienpierre:
hold on, we may need Poisson noise too.

elstoc:
I notice that the gaussian looks much more 'harsh' (greater range) than the others using the same value in the noise slider. Is this intentional? It might be better if the noise parameter was more consistent in its visual effect.

TurboGit:
I have lot of noise when using this last version even with default setting, here is a screenshot when changing the reconstruct option to show better the noise:

image

TurboGit:
Depending on the settings there is also lot of noise appearing in the shadows.

aurelienpierre:
> Depending on the settings there is also lot of noise appearing in the shadows.

Can you check the highlight mask ? It means the highlight threshold is too low, so most of the image gets reconstructed.

> I have lot of noise when using this last version even with default setting, here is a screenshot when changing the reconstruct option to show better the noise:

What noise level makes it acceptable ?

> I notice that the gaussian looks much more 'harsh' (greater range) than the others using the same value in the noise slider. Is this intentional? It might be better if the noise parameter was more consistent in its visual effect.

That's an effect of the stats distribution of noise, but I seem to have a problem with the poissonian noise anyway.

TurboGit:
All this with the default using "scene referred" mode. 

The highlight mask looks ok:

image

The image:

image

The settings (all default, not touched):

image

image

The level that makes this acceptable is around -.040 IL:

image

aurelienpierre:
Latest commit should fix most of your issues @elstoc and @TurboGit (if not all).

elstoc:
Thanks will test soon.

TurboGit:
Better but still now 100% clean:

image

As soon as you move a bit the threshold to -1.5 IL the noise is back & strong:

image

aurelienpierre:
In that case, @TurboGit you need to allow more iterations of high-quality reconstruction. The chroma noise will then be turned into luma noise that the reconstruction will be able to use to fake texture.

TurboGit:
> In that case, @TurboGit you need to allow more iterations of high-quality reconstruction. The chroma noise will then be turned into luma noise that the reconstruction will be able to use to fake texture.

Ok, indeed it smooth-out the noise in this case. I bet we''ll get lot of questions on this.

TurboGit:
The "add noise in highlights" slider does not behave very well, a mouse scroll does increment of 1 but it is generally for too much and indeed you've set 2 decimals to this slider. We probably want an increment of 0.1 ? Even 0.5 is too much I think.

aurelienpierre:
I have updated the defaults params so the noise level 0.5 instead of 1, I hope it will make it more reliable.

TurboGit:
> I have updated the defaults params so the noise level is set to 0.5 instead of 1, I hope it will make it more reliable.

My very limited experience with this module tell me that 0.2 should be even better as a default (closer to previous behavior).

TurboGit:
Looks good now! Let me play with it tomorrow a bit and we will probably be able to merge.

TurboGit:
Looks better to me now. The default values are more conservative and won't raise too many questions now. Let's have this merged to gain more field testing. Thanks!

PR #5395, opened 2020-06-10
Retouch module: use gradient slider

Mark-64:
With this PR the "preview single scale" in retouch module will use the new gradient slider library widget instead of a custom unique widget. 300 lines of code less to maintain and a more consistent gui.

Before
Screenshot from 2020-06-10 22-33-46

After
Screenshot from 2020-06-10 22-34-53

With the purpose I added some new cool features to the gradient slider widget (proportional scaling), which in turn requires a small change in the calls from other two modules.

### DIFF
--- a/src/develop/blend_gui.c
+++ b/src/develop/blend_gui.c
@@ -774,8 +774,8 @@ static void _blendop_blendif_update_tab(dt_iop_module_t *module, const int tab)
   dt_pthread_mutex_lock(&data->lock);
   for(int k = 0; k < 4; k++)
   {
-    dtgtk_gradient_slider_multivalue_set_value(data->lower_slider, iparameters[k], k);
-    dtgtk_gradient_slider_multivalue_set_value(data->upper_slider, oparameters[k], k);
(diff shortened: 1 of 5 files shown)


TurboGit:
Sounds like a nice clean-up! Will check.

TurboGit:
One issue I found. I have created 6 scales, and when clicking on the auto-adjust levels I sometime have only 2 markers:

image

And on some three as expected:

image

Mark-64:
Strange, never happened to me, will check

Mark-64:
Found the issue. 
I restored an internal clamp function that does some more sophisticated checks than I thought. It handles corner cases that it's not worth to move to the widget.
Also gradient slider didn't check for out of range inputs through the `dtgtk_gradient_slider_set_value()` functions.

TurboGit:
Works for me now! Thanks for this nice clean-up!

dterrahe:
dtgtk_gradient_slider_set_value now has an extra parameter to determine if a value-change signal should be emitted. In other bauhaus controls, this is determined by the value of darktable.gui->reset. Is there a good reason to not be consistent with that approach?

TurboGit:
> Is there a good reason to not be consistent with that approach?

No idea :)

Issue #5209, opened 2020-05-26
Noise profile for the Olympus Tough TG-6


sboukortt:
The fit for very high ISO settings (≥ 6400, the camera calls them “extended”) looks a bit strange for red and blue. Is it something that I can do much about?

[darktable-noiseprofile-20200526.tar.gz](https://github.com/darktable-org/darktable/files/4684679/darktable-noiseprofile-20200526.tar.gz)

hellbuster:
I tested the profile and it seems good. I didn't test with extended ISO images, but I could if needed.

hellbuster:
I took the liberty of opening a Pull Request #5397 with these profiles.

TurboGit:
Merged.

PR #5397, opened 2020-06-11
Olympus TG-6 noise profiles

pitbuster:
Adding profiles from #5209

I tested the profiles and they seemed fine, but I have no experience with the graphs generated by the tool to check if there is something suspicious.


TurboGit:
Looks good to me. Thanks.

PR #5400, opened 2020-06-11
enable hover effect for directory icon

Mark-64:
Just a small change in the directory icon which did not change color when hovering

Untitled


TurboGit:
Good, one less hard setting of rgba in the code! Thanks.

BTW, I feel that the directly icon is a bit too big, maybe one pixel less in all directions would be better.

PR #5405, opened 2020-06-11
cached images_to_act_on list

### LINKED ISSUE #5390: FR - make dt_view_get_images_to_act_on cached/static and have images_to_act_on stats

`dt_view_get_images_to_act_on` is basis for many actions and as such is called to properly calculate list of images to act on by given task. However if we're doing multiple tasks with images list it would be good if that list could be calculated just once.

In some cases (like in PR #5387) all is needed to know If there's something to act on, in other (eg PRs #5257) a number of items is needed.

I'd like to have 2 new methods: `dt_view_get_images_to_act_on_count` returning numbers of images to act on and `dt_view_has_images_to_act_on` simply stating if there's anything to do at all.

In case of thigs like in mentioned PRs (simple module updates) and possibly subseqent actions on images, it would be great if `dt_view_get_images_to_act_on` was calculated once, and refreshed only if change in them occured.

**Alternatives**

Currently collections provide `dt_collection_get_count` and `dt_collection_get_count_no_group` with stats recomputed only on collection change. Selection is just count, so it's not a big deal.

**Additional context**

This concern arose from looking at #5381 and my recent PRs which take advantage of images to act on.

//Edit: @AlicVB - you might help here? I don't want to mess with this and don't really know what's the best way to lower possible number of updates.

AlicVB:
This fix #5390 
This keep in cache the list of images to act on.
I've introduce a parameter to force the recompute of the list for actions on images, this may be useless, but I'm not enough confident to take the risk of applying actions to the wrong images because the list is not up to date...

### DIFF
--- a/src/common/tags.c
+++ b/src/common/tags.c
@@ -457,15 +457,19 @@ gboolean dt_tag_attach_images(const guint tagid, const GList *img, const gboolea
 gboolean dt_tag_attach(const guint tagid, const gint imgid, const gboolean undo_on, const gboolean group_on)
 {
   GList *imgs = NULL;
+  gboolean res;
   if(imgid == -1)
-    imgs = dt_view_get_images_to_act_on(!group_on);
+  {
+    imgs = dt_view_get_images_to_act_on(!group_on, TRUE);
+    res = dt_tag_attach_images(tagid, imgs, undo_on);
+  }
   else
   {
     if(dt_is_tag_attached(tagid, imgid)) return FALSE;
     imgs = g_list_append(imgs, GINT_TO_POINTER(imgid));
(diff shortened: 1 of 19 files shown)


johnny-bit:
Tested and helps very much! Thank!

src/common/tags.c
```diff
@@ -457,15 +457,19 @@ gboolean dt_tag_attach_images(const guint tagid, const GList *img, const gboolea
 gboolean dt_tag_attach(const guint tagid, const gint imgid, const gboolean undo_on, const gboolean group_on)
 {
   GList *imgs = NULL;
+  gboolean res;
```
TurboGit, line 460: = FALSE;

PR #5212, opened 2020-05-27
Fixing camera handling (libgphoto2) via a background thread

jenshannoschwalm:
There are some issues around (#2142 #4917 #4253) related to libgphoto2 camera handling.

What is basically wrong or not working in a good way?

1. scanning for cameras was done in two ways. a) as a background job once at startup and b)
 via g_idle_add later on, thus possibly blocking the gui thread.

2. detection of devices to be removed from the current list and the gui import interface **only**
tested for the model, **not** for the port. This left unavailable cameras in the list (easily when using
two sd cards).

3. updating cameras in gui was not protected from camera updating

How have these problems been (hopefully) fixed?

1. Updating the camera list is done in a background thread (bt).

2. Every 500ms the bt checks for the connected cameras, if any changes are found it
sends a signal for the gui interface to be updated.

3. Updating the camera gui interface is protected via the dt_camctl_t mutex.

4. The rescan button is not necessary any longer so it's gone.

I tested this with cards and cameras i have myself. This all works fine.

Unfortunately i have no camera with tethering so i could **not** test that case at all.

Please review and test!

Likely
fixes #2142
fixes #4917
fixes #4253

PR #5407, opened 2020-06-11
Removing gphoto devices in a backthread if not available any more

### LINKED ISSUE #4917: not noticing that a camera was disconnected, darktable lists it twice - and stalls when the user tries to use the first instance of the camera
**The issue:**
In the lighttable view, if a camera is already detected, the cable is removed, reinserted and I click on "scan for devices", it will be listed a second time. and a third, fourth times, as many times as I repeat this.
Also, if I try to use the "tethered shooting" buttin from the detected-and-removed device, darktable stalls (does not refresh the GUI at all).

**To Reproduce:**
1. Go to the lighttable view
2. Plug in a camera. I did it with a Nikon D3100 (which has limited USB interfacing, but can do tethered shooting)
3. Click on "scan for devices", and the camera will show up
4. Disconnect the cable
5. Connect USB cable again
6. Click on "scan for devices", and the camera will show up again (will be listed twice)
7. Try to use the button "tethered shooting' from the first listing (from when the camera was detected for the first time)
8. Go to the right panel and try to open camera settings. darktable will stall and not refresh the GUI, probably because it is trying to connect to the camera using the USB id that is not active anymore.

**Expected behavior:**
I would expect darktable to notice that the camera is not connected anymore, so this would probably solve both problems: it would not stall trying to access the camera through a USB device that is not active anymore, and would also not list the same camera twice.

**Platform:**
 - Darktable Version: recent git checkout  (3.1.0+1405~gec942f967-dirty), compiled few days ago
 - OS: Debian GNU/Linux
 - No OpenCL
 - Graphics card: Radeon R9 380X (not relevant to this issue)

jenshannoschwalm:
For further discussions see the old pr #5212

As automatic detection of cameras in a backthread can lead to situations where
dt fights with another app about access to a device we still have to click on
the "scan for devices button".

But automatic removal from the list and it's gui representation is safe to be done
in a background thread.

Finally fixes #4917

Intended for 3.4

### DIFF
--- a/src/common/camera_control.c
+++ b/src/common/camera_control.c
@@ -123,9 +123,13 @@ static void dt_camctl_camera_destroy(dt_camera_t *cam);
 
 /** Wrapper to asynchronously look for cameras */
 static int _detect_cameras_callback(dt_job_t *job);
+/** Wrapper to asynchronously look for disconnected cameras */
+static int _remove_cameras_callback(dt_job_t *job);
 
 /* Starts a job to update the camera list */
 void dt_camctl_background_detect_cameras();
+/* Starts a background job (running as long as the control system is present) removing disconnected cameras */
+void dt_camctl_background_remove_cameras();
 
 static int logid = 0;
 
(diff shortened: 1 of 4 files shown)


johnny-bit:
Can I ask if possible to show to user in scan for devices that some detected devices are unavailalbe because they are mounted? Currently it just shows up in log if user has `-d camctl` option in cmd line active. Making it a dialog box would be great!

As for "fighting for control in auto-detection scenario" - that could be worked too: add "eject" button that would unmount auto-mounted device :) but that's probably too much hassle.

jenshannoschwalm:
> Can I ask if possible to show to user in scan for devices that some detected devices are unavailalbe because they are mounted?

I will look into this for 3.4 in another pr to avoid interference. It's a bit tricky to debug this stuff as there so much different hardware involved ... I would prefer to have this pr to be merged pretty soon and afetr there are no negative feedbacks go into the more precise camera detection stuff.

jenshannoschwalm:
Closing as there is a better one

PR #5408, opened 2020-06-11
improve french translation on filmic

Nilvus:
Improve some text and fix some errors.

@TurboGit: I just can't add translations for new luarc installer so it remains in english but it's probably it's not same language (no add on po files after added data/luarc on POTFILES.in on my side and do an intltool-update).

### DIFF
--- a/po/fr.po
+++ b/po/fr.po
@@ -2,16 +2,16 @@ msgid ""
 msgstr ""
 "Project-Id-Version: darktable 2.6\n"
 "Report-Msgid-Bugs-To: \n"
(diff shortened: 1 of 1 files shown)


TurboGit:
> @TurboGit: I just can't add translations for new luarc installer so it remains in english but it's probably it's not same language (no add on po files after added data/luarc on POTFILES.in on my side and do an intltool-update).

Right, we have no way to translate Lua modules.

TurboGit:
Thanks a lot for cleaning up this. Some really bad mistakes from myself :)

Issue #5404, opened 2020-06-11
Image asymmetry in crop & rotate


GrahamByrnes:
**Images in crop & rotate are slightly off set to the lower right**

When you open crop & rotate to crop an image, you see a thin black border at the upper & left edges only. This is very fine if preview_downsampling=1.0, but quite noticeable at 0.25

**Expected behavior**
Symmetric image with uniform black border (I guess).

This originally relates to #5306  opened by @TurboGit  which has given birth to two PR's to refine the aspect ratio. They removed the bit of grey in between the image and the black border, but neither fixed the asymmetry.

This morning I localised the problem to this code in iop/clipping.c at L2523 (+/-):

  ```
 cairo_set_fill_rule(cr, CAIRO_FILL_RULE_EVEN_ODD);
    cairo_rectangle(cr, g->clip_max_x * wd - upleftborder, g->clip_max_y * ht - upleftborder, 
                    g->clip_max_w * wd + lowrightborder, g->clip_max_h * ht + lowrightborder);  /* *** */
    cairo_rectangle(cr, g->clip_x * wd, g->clip_y * ht, g->clip_w * wd, g->clip_h * ht);
    cairo_fill(cr);
```
In the original code, both `upleftborder` and `lowrightborder` are `1.0f`.

Strangeness: if both are equal, ie to 1.0 or 2.0, the images slides to the lower right, and there is a border of 1 or 2 pixels at the upper left only. But if `upleftborder=1.0f` and `lowrightborder=2.0f`, there is a nice even border.

So can anyone enlighten me as to what is going on here, rather than me doing an uninformed hack? I'm suspecting something about matrix indices mapping into cairo, but I don't understand well enough.

jenshannoschwalm:
I think the intention was to draw an rectangle with 2 pixels width. So if you start drawing it ad x-1/y-1 you should increase width & height by 2 ???

GrahamByrnes:
Looking at the code, it appears to try for 1px all the way around, ie -1 on the top & left, +1 at the bottom & right. But that's not what happens. If I set -2 & +2, it becomes -4 & zero. And it scales with prev_downsample, so it's not really the screen resolution...

I've just proposed a fix in PR #5410... but it's a bit of a blind hack :-(

jenshannoschwalm:
The linewidth of the rect is modified by `cairo_scale(cr, zoom_scale, zoom_scale);`

> it appears to try for 1px all the way around

[...]

PR #5410, opened 2020-06-11
Fixes offset image border in clipping.c

GrahamByrnes:
See #5404 
This fixes it, although I really don't understand why the border needs to be 2 x wider on one side than the other.. I added in a scaling by preview_downsampling and now it looks the same across the different values.

The image appears to always be completely present.

There is at some screen magnifications a slight white stripe that appears in between the black border and the image, but it disappears when you zoom a little... it's the problem of doing things at the 1-pixel level, I guess.


TurboGit:
Great detective work and this misleading/wrong display effect is now fixed even at 1/4 downsampling! Thanks a lot for tracking this down all this time.

PR #5411, opened 2020-06-11
Updated PL usermanual

lkarcz:
Note: I'm only in part familiar with git yet so I may have made a couple of messy / unnecessary commits for this PR. However, final doc/usermanual/po/pl.po ...seems... to be OK, PDF is being generated (honestly I have no idea how pl.po alone - 2nd file - appeared, and what to do with it). Git mages, please help :)


TurboGit:
Ok, I'll merge manually only the user manual. You seems to have added a .po at the root directory maybe after using a `git add .` ?

PR #5412, opened 2020-06-11
avoid to call DT_SIGNAL_TAG_CHANGED in loop

### LINKED ISSUE #5381: Performance regression using image selections.
**Describe the bug**
I suspect a severe performance regression with image selections in current master

I didn't do measurements so far and have no data to compare against but since 3.0 selecting a large number of images and handling them seems to be much slower.

**To Reproduce**
1. Select a larger number of images, let's say 1000 somehow. This already seems to be much slower than it used to be.
2. Now delete history of the selected images. This takes "ages" now and the machine seems to be locked. Just a sidenote, after such an operation the database always wants to be maintained.

I worked on history compression and removing a while ago and never stumbled across this. (Here i gave the new filmic mode a try and what it offers for a fresh import.) Until 3.0.x it maybe took a few secs to remove history of 1500 images out of 6000 on my machine here, now it took more than 15 minutes. 

A first guess would be thumbnail redrawing is triggered by a signal **much** too often, database vs drawing interfering. @AlicVB @TurboGit  - any ideas about this? There have been many things done here with the new lighttable code, i didn't follow them in detail ... 

**Expected behavior**
Being fast and responsive.

current dt master 09.06.2020
fefora 32
Intel 7700, 32GBram
nvidia 5GB graphics, cl activated or not doesn't matter.

AlicVB:
this fix #5381 

@phweyland : can you have a look, please, as you have done lot of work in this area recently ?

### DIFF
--- a/src/common/tags.c
+++ b/src/common/tags.c
@@ -473,12 +473,6 @@ gboolean dt_tag_attach(const guint tagid, const gint imgid, const gboolean undo_
   return res;
 }
 
-void dt_tag_attach_from_gui(const guint tagid, const gint imgid, const gboolean undo_on, const gboolean group_on)
-{
-  if(dt_tag_attach(tagid, imgid, undo_on, group_on))
-    dt_control_signal_raise(darktable.signals, DT_SIGNAL_TAG_CHANGED);
-}
-
 void dt_tag_set_tags(const GList *tags, const GList *img, const gboolean ignore_dt_tags,
                      const gboolean clear_on, const gboolean undo_on)
 {
@@ -579,14 +573,6 @@ void dt_tag_detach(const guint tagid, const gint imgid, const gboolean undo_on, 
 
   dt_tag_detach_images(tagid, imgs, undo_on);
   if(imgid != -1) g_list_free(imgs);
-}
-
-
-void dt_tag_detach_from_gui(const guint tagid, const gint imgid, const gboolean undo_on, const gboolean group_on)
-{
-  dt_tag_detach(tagid, imgid, undo_on, group_on);
-
-  dt_control_signal_raise(darktable.signals, DT_SIGNAL_TAG_CHANGED);
 }
 
 void dt_tag_detach_by_string(const char *name, const gint imgid, const gboolean undo_on, const gboolean group_on)
--- a/src/common/tags.c
+++ b/src/common/tags.c
@@ -344,27 +344,32 @@ gboolean dt_tag_exists(const char *name, guint *tagid)
   return FALSE;
 }
 
-static void _tag_add_tags_to_list(GList **list, const GList *tags)
+static gboolean _tag_add_tags_to_list(GList **list, const GList *tags)
 {
   const GList *t = tags;
+  gboolean res = FALSE;
   while(t)
   {
     if(!g_list_find(*list, t->data))
     {
       *list = g_list_prepend(*list, t->data);
+      res = TRUE;
     }
     t = g_list_next(t);
   }
+  return res;
 }
 
-static void _tag_remove_tags_from_list(GList **list, const GList *tags)
+static gboolean _tag_remove_tags_from_list(GList **list, const GList *tags)
 {
   const GList *t = tags;
+  const int nb_ini = g_list_length(*list);
   while(t)
   {
     *list = g_list_remove(*list, t->data);
     t = g_list_next(t);
   }
+  return (g_list_length(*list) != nb_ini);
 }
 
 typedef enum dt_tag_type_t
@@ -384,10 +389,11 @@ typedef enum dt_tag_actions_t
 
 static GList *_tag_get_tags(const gint imgid, const dt_tag_type_t type);
 
-static void _tag_execute(const GList *tags, const GList *imgs,
-                         GList **undo, const gboolean undo_on, const gint action)
+static gboolean _tag_execute(const GList *tags, const GList *imgs, GList **undo, const gboolean undo_on,
+                             const gint action)
 {
   const GList *images = imgs;
+  gboolean res = FALSE;
   while(images)
   {
     const int image_id = GPOINTER_TO_INT(images->data);
@@ -398,23 +404,26 @@ static void _tag_execute(const GList *tags, const GList *imgs,
     {
       case DT_TA_ATTACH:
         undotags->after = g_list_copy(undotags->before);
-        _tag_add_tags_to_list(&undotags->after, tags);
+        if(_tag_add_tags_to_list(&undotags->after, tags)) res = TRUE;
         break;
       case DT_TA_DETACH:
         undotags->after = g_list_copy(undotags->before);
-        _tag_remove_tags_from_list(&undotags->after, tags);
+        if(_tag_remove_tags_from_list(&undotags->after, tags)) res = TRUE;
         break;
       case DT_TA_SET:
         undotags->after = g_list_copy((GList *)tags);
         // preserve dt tags
         GList *dttags = _tag_get_tags(image_id, DT_TAG_TYPE_DT);
         if(dttags) undotags->after = g_list_concat(undotags->after, dttags);
+        res = TRUE;
         break;
       case DT_TA_SET_ALL:
         undotags->after = g_list_copy((GList *)tags);
+        res = TRUE;
         break;
       default:
         undotags->after = g_list_copy(undotags->before);
+        res = FALSE;
         break;
     }
     _pop_undo_execute(image_id, undotags->before, undotags->after);
@@ -424,6 +433,7 @@ static void _tag_execute(const GList *tags, const GList *imgs,
       _undo_tags_free(undotags);
     images = g_list_next(images);
   }
+  return res;
 }
 
 gboolean dt_tag_attach_images(const guint tagid, const GList *img, const gboolean undo_on)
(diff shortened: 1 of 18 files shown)


src/develop/develop.c
```diff
@@ -1016,6 +1016,7 @@ void dt_dev_add_history_item(dt_develop_t *dev, dt_iop_module_t *module, gboolea
   {
     /* signal that history has changed */
     dt_control_signal_raise(darktable.signals, DT_SIGNAL_DEVELOP_HISTORY_CHANGE);
+    dt_control_signal_raise(darktable.signals, DT_SIGNAL_TAG_CHANGED);
```
TurboGit, line 1019: I find this awkward adding DT_SIGNAL_TAG_CHANGED in dt_dev_add_history_item() which is a develop call and as nothing to do with tagging.

johnny-bit: the problem is that dt_dev_add_history_item already adds tag: `darktable|changed` and since changing `dt_tag_attach_from_gui` (that sends signal on it's own) to `dt_tag_attach` (which doesn't) it is needed to have this signal. This handling is better since it also checks if we have gui... it could be further aguamented to check if `dt_tag_attach` attached tag (otherwise no tag change happened) and not raise the signal ;)

AlicVB: yes, I haven't check if the signal raising what necessary or not, I've just added them outside the loops when needed. All other case do the exact same things.
Now I agree that there's room for optimization here : not raising the signal if no tag is really added, for example

jenshannoschwalm: It's really a bit strange to find the tag setting and raising here. But otherwise it's the lowest code level section we can be sure something has changed and so we are on the safe side.

AlicVB:
added : only raise the signal if tags have really been added

src/control/jobs/control_jobs.c
```diff
@@ -1143,18 +1143,20 @@ static int32_t dt_control_local_copy_images_job_run(dt_job_t *job)
...
   while(t && dt_control_job_get_state(job) != DT_JOB_STATE_CANCELLED)
   {
     const int imgid = GPOINTER_TO_INT(t->data);
     if(is_copy)
     {
-      if (dt_image_local_copy_set(imgid) == 0)
-        dt_tag_attach_from_gui(tagid, imgid, FALSE, FALSE);
+      if(dt_image_local_copy_set(imgid) == 0)
+      {
+        if(dt_tag_attach(tagid, imgid, FALSE, FALSE)) tag_change = TRUE;
```
johnny-bit, line 1154: `dt_tag_attach` already returns bool so wouldn't this be ok to change to:
```c
tag_change = dt_tag_attach(tagid, imgid, FALSE, FALSE) || tag_change;
```?

File: src/control/jobs/control_jobs.c [hunk 3]
```diff
@@ -1278,7 +1283,7 @@ static int32_t dt_control_export_job_run(dt_job_t *job)
     // remove 'changed' tag from image
     dt_tag_detach(tagid, imgid, FALSE, FALSE);
     // make sure the 'exported' tag is set on the image
-    dt_tag_attach_from_gui(etagid, imgid, FALSE, FALSE);
+    if(dt_tag_attach(etagid, imgid, FALSE, FALSE)) tag_change = TRUE;
```
johnny-bit, line 1286: `dt_tag_attach` already returns bool so wouldn't this be ok to change to:
```c
tag_change = dt_tag_attach(tagid, imgid, FALSE, FALSE) || tag_change;
```?

File: src/develop/lightroom.c [hunk 4]
```diff
@@ -856,12 +857,13 @@ static void _lrop(const dt_develop_t *dev, const xmlDocPtr doc, const int imgid,
         guint tagid = 0;
         if(!dt_tag_exists((char *)cvalue, &tagid)) dt_tag_new((char *)cvalue, &tagid);
 
-        dt_tag_attach_from_gui(tagid, imgid, FALSE, FALSE);
+        if(dt_tag_attach(tagid, imgid, FALSE, FALSE)) tag_change = TRUE;
```
johnny-bit, line 860: `dt_tag_attach` already returns bool so wouldn't this be ok to change to:
```c
tag_change = dt_tag_attach(tagid, imgid, FALSE, FALSE) || tag_change;
```?

File: src/control/jobs/control_jobs.c (hunk 2 above)
Comment (AlicVB) [reply src/control/jobs/control_jobs.c:1154]: yes this is similar.
Thought I think the previous version is more readable and doesn't change the code size.
Is there any gain using OR and not IF ?

**Comment by phweyland** [comment]
> I find this awkward adding DT_SIGNAL_TAG_CHANGED in dt_dev_add_history_item() which is a develop call and as nothing to do with tagging

Agreed. Tag signal should stay as close as possible from the tag logic if possible, at least to keep the maintenance not much difficult.
As it was designed (before any change I've made), that's up to the caller to know if it wants a signal or not, and when, (that's why we have dt_tag_attach_from_gui() and dt_tag_attach()).

So the first possible action is not signal if the tag is already attached (or detached). This is a tag internal action. Pascal has already made some work in that direction. But that doesn't really answer the following point.
If I understand the issue raised by @jenshannoschwalm the concern is about bulk actions where the same operation is repeated thousands of times. 
Either we are able to offer routines which does all the work in one call, like dt_tag_attach_string_list() for example, and then raises the signal afterwards. Some new routines have been written in that spirit for metadata copy.

Or, probably what Hubert is trying to do, is to track if there was actually at least one tag change within the thousands of actions, and at the end, decide to raise the signal or not.

Seen the number files impacted, the answer is probably not simple.
Is the diagnostic really solid ? It takes for me few seconds for removing 2000 images ...

File: src/control/jobs/control_jobs.c (hunk 2 above)
Comment (johnny-bit) [reply src/control/jobs/control_jobs.c:1154]: Essentially miniscule. for me "OR" version is more readable. in theory the "OR" version should be tiny bit more performant, but since this isn't pipe processing or anything like that the difference would be at most jump+cache loss due to incorrect prediction.

**Comment by AlicVB** [comment]
> Is the diagnostic really solid ? It takes for me few seconds for removing 2000 images ...

let's start by this : yes it is, at least for "discard history" action. We have done timings and we get a HUGE gain if we don't raise the signal, or if we raise it once for all at the end of the bulk action

> As it was designed (before any change I've made), that's up to the caller to know if it wants a signal or not, and when, (that's why we have dt_tag_attach_from_gui() and dt_tag_attach()).

That's the same now, except that the caller need to raise the signal directly.

I've tried to keep the raising inside tag logic, but the needed code change where HUGE : 
if we trace our case : 
- we start in `common/history.c`  `dt_history_delete_on_list`
- which call `dt_history_delete_on_image_ext` for each image
- which call `dt_tag_detach_by_string` 2 times
- which raise the signal

so basically, `dt_tag_detach_by_string` should no if the signal should be raised or not (with boolean params or by duplication fct)
and `dt_history_delete_on_image_ext` should know it too

considering the amount of time those 2 fct are called by other parent fct (which may have parent fct too) this lead to LOT of code change ! (I've tried)

**Comment by phweyland** [comment]
> We have done timings and we get a HUGE gain if we don't raise the signal,

Right, then.

In some way you externalize the signal action. For the tag prospective, it is a simplification, this is positive.
Another point of attention the undo action. But if the original action is a bulk one (and grouped for undo standpoint), the undo/redo action is also grouped (only one signal). And most of your changes are related to not undo-able actions.

... fine for me then.

**Comment by AlicVB** [comment]
OK, great !
Thanks for your review. Can I abuse of your time with my last commit (and I stop here, promise !)

This last commit basically return if tags have been changed or not for most tagging actions. This allow other parts of the code to avoid doing actions if nothing has happen because the tags where already present (for example)

Thanks

File: src/common/tags.c [hunk 5]
```diff
@@ -398,23 +404,26 @@ static void _tag_execute(const GList *tags, const GList *imgs,
...
         if(dttags) undotags->after = g_list_concat(undotags->after, dttags);
+        res = TRUE;
         break;
       case DT_TA_SET_ALL:
         undotags->after = g_list_copy((GList *)tags);
+        res = TRUE;
         break;
       default:
         undotags->after = g_list_copy(undotags->before);
+        res = TRUE;
```
phweyland, line 426: after list = before list means there is no change. So no need to signal here.

BTW, comparing the two lists (before / after) is another way to know if there is a change or not. But I don't know if that's more efficient, more effective...

phweyland:
Great changes. Apart the small case above, this should improve the overall tags behavior.

src/common/tags.c
AlicVB: of course... done

jenshannoschwalm:
All good for me here.

TurboGit:
Well code looks good but quite extensive as it touch lot of files/routines. I'll check again tomorrow and will merge if all goes well, we need more field testing as I'm sure we will forget one scenario in the test.

johnny-bit:
Checking with my little #5433 - the difference is tremendous (in signal countnts) but none in actual behaviour of UI. all the modules work, the behaviours I tested work, just there's A LOT LESS tags signals raised. And IMO parameter-less tags like tag-changed one are the safest to limit since nothing that depends on their params  :)

just imagine... for applying style to 180 images and then clearing history dt-tag-changed is raised:
before this PR: 720 times
after this PR: 2 times.

just imagine how less processing is done...

I'll try to test all possible touched parts but so far I haven't managed to see ANY change except occasional speed-ups.

TurboGit:
@johnny-bit : Thanks a lot for the testing, makes me less nervous for the release!

TurboGit:
I have found one issue but not related to this PR, see #5447.

So I'll merge this now. Thanks all for the delicate work!

PR #5414, opened 2020-06-11
better initialisation of history stack

### LINKED ISSUE #5396: History stack strangeness
Using the new 'scene-referred' defaults, I open a virgin image and the following are the contents of my history stack. For ease I'll refer to the modules by their number in the stack:

Screenshot_2020-06-10_21-51-38

The following is my active modules list:

Screenshot_2020-06-10_21-51-57

If I click any item from 1-8 in the list, my active modules list excludes sharpen but otherwise does not change and nor does the preview image. If I click item 9, the sharpen module appears and if I click item 10, again, nothing changes.

I'd say that the history should have the entirely mandatory modules first (1, 3, 5, 7, 8), followed by the 'always applied' (2, 10) then the workflow modules (4, 6) and finally sharpen (9). Also, given that highlight reconstruction is 'always applied' that should probably appear as well (in the 'always applied' list.

Then clicking through 0 (original) and the first five (mandatory) modules should leave the image unchanged (as if you'd selected the fifth mandatory module) and all the others should progressively apply the modules in turn. 

3.1.0+2023~geb85207f2 on Arch Linux

elstoc:
when clearing the history stack or loading a new image

 - any module which is forced on and cannot be switched off appears first in the history stack
 - any module that is on by default but can be switched off appears next in the history stack
 - finally any modules that are enabled by presets or as a result of a preference setting (workflow or sharpen) are added

the final set of modules can be disabled and removed from the history stack but, as previously, the first two sets will be automatically re-added if they are removed

also fixes bug where highlight reconstruction was not present in the history stack

finally changes the scene-referred exposure defaults to a preset to ensure that they only apply to the first module instance and can be easily reapplied if necessary (by choosing the preset)

Resolves #5396, #5416

### DIFF
--- a/src/iop/exposure.c
+++ b/src/iop/exposure.c
@@ -646,7 +646,6 @@ void reload_defaults(dt_iop_module_t *module)
       // For scene-referred workflow, since filmic doesn't brighten as base curve does,
       // we need an initial exposure boost. This might be too much in some cases but…
       tmp.exposure = 1.0f;
-      module->default_enabled = TRUE;
     }
   }
 
--- a/src/iop/exposure.c
+++ b/src/iop/exposure.c
@@ -276,22 +276,21 @@ void init_presets (dt_iop_module_so_t *self)
                              sizeof(dt_iop_exposure_params_t), 1);
 
 
-  if(strcmp(dt_conf_get_string("plugins/darkroom/workflow"), "scene-referred") == 0)
-  {
-    // For scene-referred workflow, since filmic doesn't brighten as base curve does,
-    // we need an initial exposure boost. This might be too much in some cases but…
-    dt_gui_presets_add_generic(_("scene-referred default"), self->op, self->version(),
-                               &(dt_iop_exposure_params_t){.mode = EXPOSURE_MODE_MANUAL,
-                                                           .black = 0.0f,
-                                                           .exposure = 1.0f,
-                                                           .deflicker_percentile = 50.0f,
-                                                           .deflicker_target_level = -4.0f,
-                                                           .compensate_exposure_bias = TRUE},
-                               sizeof(dt_iop_exposure_params_t), 1);
+  // For scene-referred workflow, since filmic doesn't brighten as base curve does,
+  // we need an initial exposure boost. This might be too much in some cases but…
+  dt_gui_presets_add_generic(_("scene-referred default"), self->op, self->version(),
+                             &(dt_iop_exposure_params_t){.mode = EXPOSURE_MODE_MANUAL,
+                                                         .black = 0.0f,
+                                                         .exposure = 1.0f,
+                                                         .deflicker_percentile = 50.0f,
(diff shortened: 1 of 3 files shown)


elstoc:
For some reason the flip module appears last in the stack and I can't figure out why. By no means is this a showstopper though because this PR fixes all of the issues in 5396 so IMO it's still better than master even if not 100% consistent.

elstoc:
Thinking on this again, the exposure autosetting really has to be managed via a preset, to ensure that subsequent instances don't also start with the same defaults. A second instance of exposure should never need the +1EV or exposure bias compensation. And if we're going to manage exposure with a preset (and sharpen and basecurve are already managed with a preset) it makes sense to do the same with filmicrgb.

TurboGit:
For the flip module it is special cased in different part of the code. For example in dt_image_get_orientation (common/images.c). Maybe that's why it behave differently.

elstoc:
I don't think it's an issue now I've moved everything else to presets. Also related to the second issue, I don't really understand what the code in database.c is designed to do. Maybe that has something to do with the basecurve auto-apply issues.

src/iop/exposure.c
```diff
@@ -274,6 +274,24 @@ void init_presets (dt_iop_module_so_t *self)
                                                          .deflicker_target_level = -4.0f,
                                                          .compensate_exposure_bias = FALSE},
                              sizeof(dt_iop_exposure_params_t), 1);
+
+
+  if(strcmp(dt_conf_get_string("plugins/darkroom/workflow"), "scene-referred") == 0)
```
TurboGit, line 279: I won't make this preset creating conditional this will avoid one case where we need to restart dt to swicth from scene-referred to another choice. Yes I know there is other case like that for basecurve for example.

TurboGit: Of course if we do that we need to auto-apply only for the scene-referred workflow and we are back to the basecure case anyway :(

elstoc:
Yes I agree - I did think about that when I did it. I've pushed that change

elstoc:
I'm still not convinced that any of the modules that may be switched off should be auto-enabled via the module->default_enabled flag due to the problems I've mentioned in my issue (namely that if one can disable the module one should also be able to remove it from the stack) but given how close we are to feature freeze I didn't think it was worth opening that can of worms yet. IMO that flag should really be limited to just those modules that are fully mandatory.

TurboGit:
> but given how close we are to feature freeze I didn't think it was worth opening that can of worms yet. IMO that flag should really be limited to just those modules that are fully mandatory.

Agreed on both counts !

elstoc:
Perhaps I'll raise another FR to discuss for 3.4

TurboGit:
> Perhaps I'll raise another FR to discuss for 3.4

Would be perfect !

TurboGit:
Thanks for working on this. Better initialization indeed.

elstoc:
FR as discussed is #5434.

PR #5417, opened 2020-06-12
Scripts installer not always maximized

### LINKED ISSUE #5401: Lua scripts installer remains (partially) open on darktable start. should not be
**Describe the bug**
Just open darktable and see that new lua scripts installer is partially open. module header bar is on hide/close state but not the content that is shown. Clicking on header bar hide the content. After that, close darktable and open it and see that this module remains partially open (header bar on close state but content shown).

**Expected behavior**

Just have the content hide on darktable start and when the collapse button is in hide/close state (as all other modules) and be sure darktable start again with last state not this default not wanted one.

@wpferguson: this one probably for you.

wpferguson:
Set the module to be opened automatically the first time only and thereafter respect the users setting.  Added a text block with instructions that displays the first time you run the script.  Made the script translatable.

There is still an issue with the collapse arrow not pointing correctly, but that is an API problem that should be addressed in a separate issue.

Fixes #5401

### DIFF
--- a/data/luarc
+++ b/data/luarc
@@ -26,6 +26,15 @@ local _scripts_install = {}
 _scripts_install.dt = require 'darktable'
 
 
+_scripts_install.dt.preferences.register(
+  "_scripts_install",
+  "dont_show",
+  "bool",
+  _scripts_install.dt.gettext.gettext("lua scripts installer dont show again"),
+  _scripts_install.dt.gettext.gettext("do not show scripts_installer if lua scripts are not installed"),
+  false
+)
+
 if not _scripts_install.dt.preferences.read("_scripts_install", "dont_show", "bool") then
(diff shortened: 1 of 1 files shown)


TurboGit:
So, I suppose that after you said "don't show again"  the only way to reactivate the lua installer is to edit the corresponding settings in ~/.config/darktable/darktablerc, right ? I'm wondering if we could have a dt preference to actually reset this ? Maybe just the same key with the 3 options "install script", "ask later", "don't show againi" will work ?

wpferguson:
This thought had crossed my mind too, but then I got distracted fixing other things.  "dont show again" is the only option that you can't recover from unless you edit the darktablerc file, so I'll just register that preference so that the user can reset it if they change their mind.

TurboGit:
Working perfectly now ! Thanks.

PR #5421, opened 2020-06-12
change schedule(dynamic) to schedule(static) in parallelized for loops

ralfbrown:
Yields a substantial speedup at higher thread counts for chromatic aberration, X-trans demosaic, and highlight reconstruction with Lch and "color" methods, by reducing thread-management overhead on loops with a known count of iterations where each iteration takes the same amount of time.  (See also #4789)

### DIFF
--- a/src/iop/highlights.c
+++ b/src/iop/highlights.c
@@ -546,15 +546,15 @@ static void process_lch_bayer(dt_iop_module_t *self, dt_dev_pixelpipe_iop_t *pie
 #ifdef _OPENMP
 #pragma omp parallel for default(none) \
   dt_omp_firstprivate(clip, filters, ivoid, ovoid, roi_out) \
-  schedule(static)
+  schedule(static) collapse(2)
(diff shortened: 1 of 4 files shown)


src/iop/highlights.c
```diff
@@ -546,7 +546,7 @@ static void process_lch_bayer(dt_iop_module_t *self, dt_dev_pixelpipe_iop_t *pie
 #ifdef _OPENMP
 #pragma omp parallel for default(none) \
   dt_omp_firstprivate(clip, filters, ivoid, ovoid, roi_out) \
-  schedule(dynamic)
+  schedule(static)
 #endif
   for(int j = 0; j < roi_out->height; j++)
   {
```
aurelienpierre, line 552: that would be better if the loops could be perfectly nested too, so we could pass `collapse(2)`

ralfbrown: Done.

PR #5422, opened 2020-06-12
don't treat 'Tab' as an escape character in geotagging

### LINKED ISSUE #4969: Input field for offset in geotagging panel is reset
**Describe the bug**
When I try to change the offset when geotagging my images the offset I choose is not accepted by the gui. I also started darktable with `darktable --configdir /tmp/darktable-config/` to ensure that there is no problem with my config.

**To Reproduce**
1. Start darktable
2. go the the lighttable
3. Open the "geotagging" panel
4. change the offset, e. g. to +01:00:00

**Expected behavior**
The valid offset should be accepted and not be reset by the gui.

 - Darktable Version: 3.0.2.
 - OS: Archlinux, Linux  5.6.10-arch1-1

**Additional context**
 - Is the issue still present using an empty/new config-dir: **yes**

johnny-bit:
Dunno reasoning for original handling nor "solution" but "works" :)

fixes #4969

### DIFF
--- a/src/libs/geotagging.c
+++ b/src/libs/geotagging.c
@@ -192,7 +192,6 @@ static gboolean _lib_geotagging_offset_key_press(GtkWidget *entry, GdkEventKey *
   switch(event->keyval)
   {
     case GDK_KEY_Escape:
-    case GDK_KEY_Tab:
     {
       // reset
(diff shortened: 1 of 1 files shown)


TurboGit:
This comes back from 2012 almost initial commit of this module.

TurboGit:
Looks good to me also. Thanks !

PR #5424, opened 2020-06-12
More precise description of added noise in filmicrgb.c

jpellegrini:
@aurelienpierre  --  since it can be uniform, gaussian or poissonian, then it should probably not be called "gaussian" in the description.


aurelienpierre:
right, that's a left-over. Thanks !

PR #5426, opened 2020-06-12
Use /usr/bin/env to invoke bash

ayang64:
This commit adds /usr/bin/env to the magic number line of shells scripts
to increase portability.

One of the uses of the /usr/bin/env is that it can invoke a command
using the user's current environment by searching their ${PATH}.  Also,
since /usr/bin/env is part of many standard distributions (FreeBSD, most
if not all Linux distros, and OSX), it is more likely to be found at a
fixed path than bash.

This allows us to run bash scripts regardless of the shell's location on
the filesystem -- which varies between operating systems and
distributions.

This updates the build-doc.sh, build.sh, and .githooks/pre-commit
scripts.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 
 set -e
 
@@ -180,6 +180,12 @@ num_cpu()
 			fi
 		fi
 		;;
+
+	FreeBSD)
+		export CMAKE_MAKE_PROGRAM=/usr/local/bin/gmake
+		ncpu=$(/sbin/sysctl -n kern.smp.cpus 2>/dev/null)
+		;;
+
 	Darwin)
 		ncpu=$(/usr/sbin/sysctl -n machdep.cpu.core_count 2>/dev/null)
 		;;
@@ -250,6 +256,8 @@ if [ $PRINT_HELP -ne 0 ] ; then
 fi
 
 CMAKE_MORE_OPTIONS=""
+export CMAKE_MAKE_PROGRAM=/usr/local/bin/gmake
+
 for i in $FEATURES; do
 	eval cmake_boolean_option USE_$i \$FEAT_$i
 done
--- a/.githooks/pre-commit
+++ b/.githooks/pre-commit
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 
 # Hash using its key as a search Regex, and its value as associated
 # error message.
--- a/build-doc.sh
+++ b/build-doc.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 #
 # This script is a quick hack to make it easier to build the darktable documetation
 # in PDF, html-multi and html-single formats, for the user manual and LUA API.


rabauke:
Some time ago there was a similar pr, see #4408. After some discussion, it was agreed that `#!/bin/bash` shall be replaced by `#!/bin/sh` rather than `#!/usr/bin/env bash`. Using the `env` tool is actually an abuse as it was created for some rather different purpose. Furthermore, using `env` merely shifts portability issues rather than solving them. 

If one aims for portability, it is much better to restrict shell code to POSIX shell features only and to use `#!/bin/sh`.

ayang64:
thank you for the feedback.

just curious: how is this an abuse?  i didn't understand this by reading the original PR.

johnny-bit:
I looked it up at that time - it's complex. Basically env isn't guaranteed to work (or be where you want it to be), there's no guarantee to have bash at all, there's no guarantee how passing/aliasing would work etc.

Ultimately using `/bin/sh` is closest thing to *portable* since it relies on posix shell standards. Never going beyond recommended sh (it's powerfull enough as it is for most uses) guarantees that it'll work on all weirdo-sh regardless whether it's ash,dash,bash,ksh,zsh or whatever as long as they provide posix-compatible sh.

build.sh
```diff
@@ -250,6 +256,8 @@ if [ $PRINT_HELP -ne 0 ] ; then
 fi
 
 CMAKE_MORE_OPTIONS=""
+export CMAKE_MAKE_PROGRAM=/usr/local/bin/gmake
+
```
johnny-bit, line 260: there's no need for this if above case worked good enough.

rabauke:
@ayang64 

> just curious: how is this an abuse? i didn't understand this by reading the original PR.

I called it an abuse because the primary  purpose of `env` is to run a program in a _modified_ environment. `env bash`, however, runs the bash in an unmodified environment but guarantees (as a mere side effect of `env`) that `bash` is found in the file system rather than as a shell built-in. See also https://www.gnu.org/software/coreutils/env

ayang64:
> @ayang64
> 
> > just curious: how is this an abuse? i didn't understand this by reading the original PR.
> 
> ... that `bash` is found in the file system rather than as a shell built-in. See also https://www.gnu.org/software/coreutils/env

IMHO, there is nothing wrong with using env to improve portability.  But I'm also willing to concede that better minds have already thought about this.

I feel like that bit that you quoted from the GNU manual confirms using for this purpose is appropriate .  `env` may not be guaranteed to be in the same place but in practice, it is there is almost every available modern unix variant.

I'll leave my thought that right now the scripts are incompatible with FreeBSD.  Using `env` will make the scripts widely portable albeit imperfect in some cases which I can't think of at the moment.

parafin:
I don't think @rabauke ever explained what's wrong with env...

coffeeandspaceships:
> I'll leave my thought that right now the scripts are incompatible with FreeBSD. 

Conforming to #!/bin/sh per @rabauke's suggestion would also fix FreeBSD compatibility and still does not require env.

Regarding the thoughts on bash v POSIX sh, note that Ubuntu, for instance, as a major player, now encourages use of `dash` and `zsh`. This shows that relying on bash is not safe in practice.

If this has been discussed a couple of times is it worth raising a "feature request" - whatever you guys call the non-functional equivalent of that... to declare POSIX sh as a coding standard and we can then do some linting etc for 3.4?

Issue #4797, opened 2020-04-23
[crop&rotate] extraneous pixelpipe runs


ralfbrown:
**Describe the bug**

Darktable will often run the pixelpipe when clicking in the on-screen crop box when crop&rotate is active.  It is the actual click that triggers the pixelpipe, not any change in the crop box (in fact, even a right click will trigger it).  I was alerted to this extraneous computation by my PC's fans spinning up.

**To Reproduce**

1. Run darktable -d perf --disable-opencl
2. Select a photo which ideally has a number of computation-heavy iops after crop&rotate enabled (e.g. denoise (nonlocal means), contrast equalizer, local contrast (laplacians))
3. Open crop & rotate
4. click in the crop box and observe the console output
5. not every click triggers a major recomputation, but the following sequence reliably does:
   5a. resize the crop box by dragging an edge or corner
   5b. move the mouse pointer outside the crop box
   5c. move the mouse pointer back into the crop box
   5d. click and release (left or right click, doesn't matter)

**Expected behavior**

Darktable should not be running the pixelpipe, as nothing is being changed.

**Screenshots**

Some times, the click produces only a trivial amount of work, which I refer to a as a good run.  Other times, there is a full recomputation from crop&rotate onward, which I refer to as a bad run.  The below extracts are the outputs resulting from a mouse click, until quiescence.

Sample log from a good run:

~~~~
43.161804 [dev] took 0.000 secs (0.000 CPU) to load the image.
43.162947 [dev_process_image] pixel pipeline processing took 0.001 secs (0.024 CPU)
43.162971 [dev_pixelpipe] took 0.001 secs (0.024 CPU) processed `display encoding' on CPU, blended on CPU [preview]
image colorspace transform RGB-->RGB took 0.003 secs (0.062 CPU) [final histogram]
final histogram took 0.002 secs (0.008 CPU)
final histogram waveform took 0.004 secs (0.117 CPU)
43.172534 [dev_process_preview] pixel pipeline processing took 0.011 secs (0.213 CPU)
~~~~

Sample log from a bad run:

~~~~
63.177706 [dev] took 0.000 secs (0.000 CPU) to load the image.
63.195588 [dev_pixelpipe] took 0.001 secs (0.015 CPU) processed `crop and rotate' on CPU, blended on CPU [preview]
63.196733 [dev_pixelpipe] took 0.001 secs (0.026 CPU) processed `input color profile' on CPU, blended on CPU [preview]
[...]

PR #5430, opened 2020-06-12
filmicrgb: disable highlight reconstruction when using ashift.

### LINKED ISSUE #5423: performance regression with filmic / perspective correction
**Describe the bug**
When FilmicRGB is activated there is a huge performance lost when trying to select the crop with "original format". It is so slow that moving the area is almost impossible.

**To Reproduce**
1. Go to darkroom
2. Activate FilmicRGB (let default settings) 
3. Activate "perspective correction" module
4. Fix perspective
5. Select "original format" in "automatic cropping"
6. Try moving the area

**Expected behavior**
Fast and responsive GUI even when FilmicRGB is activated.

 - Darktable Version: master
 - OS: GNU/Debian

TurboGit:
Also when using clipping module. This is to have a bearable speed
when cropping with original format.

Fixes #5423.

### DIFF
--- a/src/iop/filmicrgb.c
+++ b/src/iop/filmicrgb.c
@@ -1448,9 +1448,18 @@ void process(dt_iop_module_t *self, dt_dev_pixelpipe_iop_t *piece, const void *c
     }
   }
 
+  // the highlight reconstruction is slow and impair a lot the use of perspective correction module
+  // for example when trying to change the crop area. When  ashif & clipping module are the active
+  // module in the GUI we do disable the highlight reconstruction, not really needed while cropping
+  // anyway.
+  const gboolean run_fast = darktable.develop
+    && darktable.develop->gui_module
+    && (strcmp(darktable.develop->gui_module->op, "ashift") == 0
+        || strcmp(darktable.develop->gui_module->op, "clipping") == 0);
+
   float *const restrict reconstructed = dt_alloc_sse_ps(roi_out->width * roi_out->height * ch);
 
-  if(recover_highlights && mask && reconstructed)
+  if(!run_fast && recover_highlights && mask && reconstructed)
   {
     float *const restrict inpainted =  dt_alloc_sse_ps(roi_out->width * roi_out->height * ch);
     inpaint_noise(in, mask, inpainted, data->noise_level, data->reconstruct_threshold, data->noise_distribution,


TurboGit:
I think it is the first time we check for other modules in another module. Not sure if this will work in all cases, at least it definitely fix the perf issue. Even if OpenCL makes this perf issue less important it will not fix the issue for people not having OpenCL.

Thoughts ?

ralfbrown:
Note that the clipping GUI sometimes triggers a recomputation unnecessarily, so this PR will help when that happens.  See #4797.

johnny-bit:
DAMN! If this is allowed then couple combos could be "fixed" eg: tone eq + zoom + denoise (profiled) ;)

parafin:
This looks like a hack. Either enable highlights reconstruction only on export or when GUI focus is currently in the filimic iop, or introduce some special "fast" pipe mode, which will allow to skip heavy processing steps (like denoise).

TurboGit:
First, yes I agree that this is a hack but so close of 3.2 freeze it may be hard to introduce a more generic implementation. We really need something to workaround the slowdown that makes some feature unusable.

> Either enable highlights reconstruction only on export or when GUI focus is currently in the filimic iop

You then mean inverting the test (checking for filmic and then enable slow code) ? Not sure it is a good idea as the picture may look different when working in all IOP except filmic. My idea was that this was not so important (recovery, image quality in general) when cropping.

>   or introduce some special "fast" pipe mode, which will allow to skip heavy processing steps (like denoise).

Can you clarify ? Who will be in charge of setting the fast pipe mode ? This may be for 3.4.

MStraeten:
the way it's implemented is better than just enabling this on export or if filmic iop is in focus. Now it’s just disabled when perspective correction iop is in focus which is ok in my opinion. 
For denoise profiled it might make sense to disable this in zoomed views outside the denoise iop. Maybe this behaviour can be controlled by a preference setting since at least when using drawn masks each iop might be executed in zoomed views ...

aurelienpierre:
A bit too hacky for my taste, it will trigger a tree of special handling pathes and possible side effects.

TurboGit:
@aurelienpierre : what would happen if highlight reconstruction is not called when in ashift module ? Note that as sson as you move module or collapse the ashit module the full re-computation will happen. So yes hacky, but I really don't see what could go wrong in term of rendering.

GrahamByrnes:
> You then mean inverting the test (checking for filmic and then enable slow code) ? Not sure it is a good idea as the picture may look different when working in all IOP except filmic. My idea was that this was not so important (recovery, image quality in general) when cropping.

This is peculiar: unless the default of filmic RGB is a no-op (and I actually think it should be for reconstruction), then necessarily the image will look different if filmic RGB is used.

In any case, this raises a question I was ruminating on this morning: suppose I was to construct a "Special B&W path". It would ideally set white balance to avoid clipping anything not already clipped, it would skip past channel mixer and implement something that used un-clipped channels to fill in for clipped ones at the extremes of exposure, and it would do some sort of norm-dependent exposure adjustment.

That would make sense only if there is the possibility of a "special use" pipe: you use this module, then these are unavailable. This is already present in some hacky way with base-curve vs Filmic, tone equaliser vs tone curve. It would be a step towards the graphical network processing logic, but could be implemented with an internal check list: "if this module is on, these get locked off".

This is too much to digest for 3.2, obviously.

MStraeten:
the user isn’t loosing relevant image informations while tweaking perspective control since the information needed there are not affected by highlight reconstruction in filmic due to their position in the pixel pipe. I doubt it would even be noticed more than more responsive UI ...  
i doubt a separate pipe could be a solution for performance tuning - the most performance consuming iops are retouch, tone equalizer, denoise profiled and filmic rgb. While editing only retouch or denoise profiled might be temporarily disabled to increase responsivity since it doesn’t make sense to disable tone or color corrections ...

parafin:
> > or introduce some special "fast" pipe mode, which will allow to skip heavy processing steps (like denoise).
> 
> Can you clarify ? Who will be in charge of setting the fast pipe mode ? This may be for 3.4.

Good question. It could be some user-settable switch for example. And it could be activated when something interactive is being done which triggers a lot of redraws, e.g. dragging of a slider until user lets go of it.

jenshannoschwalm:
> suppose I was to construct a "Special B&W path"

Let's come back to this after 3.2, there are at least two of us :-)

TurboGit:
See for a more generic solution in #5441.

TurboGit:
I'm closing this one as I think #5441 is the way to go and finally not too much work and invasive so it should be find its way to 3.2.

PR #5432, opened 2020-06-12
Fix histogram button tool tips

### LINKED ISSUE #5431: Histogram tool tips don't toggle
**Histogram: tool tips wrong for green & blue channels**

The histogram has three colour buttons on the image, which allow to either show or hide data from red, green or blue channels. Clicking on the button acts as a toggle.

For the red, the tool tip also toggles, between "hide red channel" and "show red channel".
For the green & blue channels, the tool tip is always "hide XXX channel". The behaviour is correct, the tool tip is 50% wrong.

**To Reproduce**
Open any coloured image in darkroom, switch to regular histogram.

 - Darktable Version: 3.1.0+2115 current master
 - OS: Win 10
 - No OpenCL
 - Nvidea K620

GrahamByrnes:
This should fix #5431


src/libs/histogram.c
```diff
@@ -584,13 +584,15 @@ static gboolean _lib_histogram_motion_notify_callback(GtkWidget *widget, GdkEven
     else if(x > d->green_x && x < d->green_x + d->button_w && y > d->button_y && y < d->button_y + d->button_h)
     {
       d->highlight = DT_LIB_HISTOGRAM_HIGHLIGHT_GREEN;
-      gtk_widget_set_tooltip_text(widget, d->red ? _("click to hide green channel")
+      dt_control_change_cursor(GDK_LEFT_PTR);
```
TurboGit, line 587: if we add dt_control_change_cursor() here we probably want to add it for the red case too.

But is has been removed in 31539181912256535963031360d864be38955f99 2 days ago. I tested this and it was ok without those call to change the cursor. The fix is probably to not reintroduce them.

GrahamByrnes:
I think something got screwed up in the "resolve conflicts" step: the proposition was to replace two words: red -> green, red -> blue. Pretty sure these are just a ctr-v error.

Somehow the `dt_control_change_cursor` got mixed in as I was coming from an old master. Will try to fix

src/libs/histogram.c
GrahamByrnes: Ok, didn't trigger the "resolve conflicts" flag this time, so I think it should be ok.
Dramas 'cos I couldn't download from github :-(

GrahamByrnes:
> We probably want to remove the dt_control_change_cursor(GDK_LEFT_PTR) calls.

LOL... I just did :-)

PR #5433, opened 2020-06-12
add signal debugging options

johnny-bit:
adds possibilit to run `-d signal` to check whether some signals aren't called TOO frequently ;)
this might come in handy when handling some debugging/performance and also when deciding what signal to use when... generally very much developer centric option.

also adds possible backtrace printing for signals (which might not be so useful)
signal backtracing uses GNU extension and is very much hidded when building + enabling it WILL emit a warning. It should also require (or be disabled) in case the compiler doesn't have GNU extensions.

this is build on legs of my comment in #5381 - https://github.com/darktable-org/darktable/issues/5381#issuecomment-643196839

pinging @AlicVB & @jenshannoschwalm (and others) - may it be useful ;)


src/control/signal.c
```diff
@@ -322,15 +349,18 @@ void dt_control_signal_raise(const dt_control_signal_t *ctlsig, dt_signal_t sign
...
                                gpointer user_data)
 {
+  dt_print(DT_DEBUG_SIGNAL, "[signal] connect: %s\n", _signal_description[signal].name);
+  _print_trace("connect");
   g_signal_connect(G_OBJECT(ctlsig->sink), _signal_description[signal].name, G_CALLBACK(cb), user_data);
 }
 
 void dt_control_signal_disconnect(const struct dt_control_signal_t *ctlsig, GCallback cb, gpointer user_data)
 {
+  dt_print(DT_DEBUG_SIGNAL, "[signal] disconnect\n");
```
AlicVB, line 362: not sure of this one : as it doesn't show the signal name, it will be useless imho and just spam the terminal

johnny-bit:
Apart from possible counting of register/unregister it's a bit spamy,
however with enabled signal tracing compile option you know grow where that
call came. It's a bit of shame that we can't really get what signal got
disconnected, but such is life.

pt., 12 cze 2020, 19:15 użytkownik AlicVB <notifications@github.com>
napisał:

jenshannoschwalm:
having sort of signal-debgging aid would indeed be **very** good.

But i think this is just a debugging tool only to be included in a debug version.

Also, if there would be such a signal tracing tool we would
1. need info about who/what function & thread is signal raising
2. signal received by whome, again possibly function & thread
3. probably a list of signals to be traced

I did such stuff in real-time applications and if this info wants to be useful is must be **very** fast to analyse the signal flow as it is without the tracer. Many more things to keep in mind.

I really think we have to agree on "what we we need" and "how could this be done" first ...

johnny-bit:
I'm with Hanno on this one too... buuuuut...

With very little code ingerention as I have here it gives very ballpark info on what's going on.
If compilling with signal_trace enabled this gives info what called signal raise, what registered and what disconnected signal. obviously that's a bit... "less than ideal"

I hoped that https://wiki.gnome.org/Projects/GTK/Inspector has signal tracking but it's kinda not there. As signals used by darktable are gtk signals we'd need either use tools from gtk or do very nasty code changes (At least I have no clear ideas).

As for this PR - it just does extremally basic info as evidenced by code. I have no real good idea what to do better with it. I also technically could log signal raised params and that's probably it.

johnny-bit:
I checked how signal debugging works in GTK/GLib and at current state it's "basically useless". The way presented in this PR is very basic one, although gets things done :wink: 

@jenshannoschwalm - I'm very open to discussing how to do better, especially list of signals to be traced. I honestly have no idea except that weird `_print_trace` on how to get without deeper code changes "who raised"/"who received"

jenshannoschwalm:
Yes, after reading your comment i had a deep look int gtk signals. Indeed they seem not to be intended for what we might be thinking of.

Probably in `dt_control_signal_connect` when calling `g_signal_connect(instance, detailed_signal, c_handler, data)` we currently pass "pure data" like self, this could be replaced by a struct holding "pure data" plus "raising information"

The signal callback would have to read the correct data and could also write info about raiser/sender (maybe also sending time) to log.

Possible, but is it worth it? Maybe define `dt_control_info_signal_connect` so we could replace step-by-step?

johnny-bit:
Just so we're on the same page here and I didn't misunderstood you... The steps would be to:
1. Introduce new method, `dt_control_info_signal_connect` which would be identical to `dt_control_signal_connect`, but would also contain signal registrar info, eg `__func__`, `__FILE__`, `__LINE__`, so we know who/what registers signal
2. maybe create structure, eg `dt_signal_info_t` which would contain such info
3. add `dt_control_info_signal_raise` to which also would contain raiser info
4... and that would be the most involved one... add some debug printing func bolierplate to signal handlers, which would allow to print signal handler info along with signal raiser info.
5. maybe actually have some additional debug param which as you mentioned, would allow tracing single or selected signals (or all)

Does this sound right?

jenshannoschwalm:
It sounds right :-)

About the passed struct, i read the g_signal docs and am not sure right now: Can we be sure we do the callback for every signal raised? If true, the structs can be malloc-ed, otherwise we would need further tricks,

parafin:
Having a callback shouldn't matter, you must use destructor instead, see as example https://github.com/darktable-org/darktable/blob/d8c9e482ba0ec5823ecead4ace72637ae2484e7d/src/control/signal.c#L57

Or I missed something...

johnny-bit:
Yeah, destructors would be indeed needed. And nice macros to better handle all registration/invocations :/

johnny-bit:
Interesting - GObject signals are connected to signal with callback + user_data, but disconnected only by callback + user_data. Which means that if you have the same callback + user_data for several signals it's enough to use 1 disconnect and it will disconnect from all. If you want to disconnect from signle signal it's not really possible.

Please test how you like current output (without introducing `dt_signal_info_t` and further changes in handlers + adding destructors)

parafin:
> Interesting - GObject signals are connected to signal with callback + user_data, but disconnected only by callback + user_data. Which means that if you have the same callback + user_data for several signals it's enough to use 1 disconnect and it will disconnect from all. If you want to disconnect from signle signal it's not really possible.

Well, it's possible in glib, just isn't implemented in darktable's code.

johnny-bit:
> Well, it's possible in glib, just isn't implemented in darktable's code.

Yeah, you're right `g_signal_handlers_disconnect_matched` can have `G_SIGNAL_MATCH_ID` in order to match signal. That being said it's not necessary :)

Now - regarding how it standands - is the current output in good direction? Introducing mentioned info structure is a bit complex but doable (I need more input on that).

Regarding "which signal to analyse" - would it be better to have it as compite-time option or runtime one?

More input please :)

johnny-bit:
I've rebased PR on current master. If you have time - please check how you like it and how can I improve it further?

jenshannoschwalm:
> is the current output in good direction?

Yes, this will help and give the most relevant information already.

>  would it be better to have it as compite-time option or runtime one?

I would suggest runtime.

> which signal to analyse?

All signals traced is very noisy. So maybe a list like
`-d signal DT_SIGNAL_ACTIVE_IMAGES_CHANGE,DT_SIGNAL_DEVELOP_MIPMAP_UPDATED` ?

It might also be beneficial to have counters for specific signal raised and callback_executed displayed.

johnny-bit:
Most recent version with small changes :wink: 

I introduced new commandline options:
1. `--d-signal-act` can limit which action is reported, works same as `-d` so in order to avoid `disconnect` spam, one can specify `--d-signal-act raise --d-signal-act connect`. `--d-signal-act all` is available as a way to enable debug of raise/connect/disconnect calls
2. `--d-signal` which can limit signals being debugged (except in case of disconnect which has no signal id info). accepts 'ALL' as a signal to allow debugging all signals. works as a flag, so specifying 2 signals would be `--d-signal DT_SIGNAL_ACTIVE_IMAGES_CHANGE --d-signal DT_SIGNAL_DEVELOP_MIPMAP_UPDATED`

additionally specifying ONLY `-d signal` without supporting `--d-signal *` results in silence about signals because I didn't want to assume `all` actions and `ALL` signals to be of interest.

So - this is ready for review :)

src/CMakeLists.txt
```diff
@@ -777,6 +777,13 @@ else(USE_GAME)
   message(STATUS "Game: no")
 endif(USE_GAME)
 
+if(USE_SIGNAL_TRACE)
```
TurboGit, line 780: Why not enable support for this unconditionaly ? As for all debug option we just need "-d <name>" to enable it. Maybe there is something special on this one?

johnny-bit: It's very noisy tracing option that prints call backtrace, which would be only useful in very marginal situations like "from where the hell this signal raise/connect/disconnect came???" If enabled it slows down things by a bit due to checking for backtrace. This tracing makes sense imo only for devs or very very narrow use case and also requires GCC because it uses gnu extensions. So it IMO should be behind an option and hidden by default :)

TurboGit: Sure it is very noisy like "-d sql" my point is not to make it always on but to enable it only at runtime when needed with "-d signal" and so I don't see the point of having this enabled/disabled at config time in cmake.

johnny-bit: how about now? if you compiled darktable with gcc, it defines that it's possible to print trace and you get `--d-signal-act print-trace` option to print trace. it's additionally conditional so I think it's low impact.

TurboGit: Looks better to me! Thanks.

dterrahe:
Quick spot: if you define an explicit gui_cleanup, you'll also have to perform the function that default_gui_cleanup otherwise would perform, namely to free self->gui_data ...

johnny-bit:
Thanks! Fixed in  #6163 :)

PR #5440, opened 2020-06-12
make user css work on windows

Mark-64:
This PR will make user css tweak in preferences to work in Windows.
css `@import()` statement doesn't like backslashes or drive letters.
Need to replace both.

### DIFF
--- a/src/gui/gtk.c
+++ b/src/gui/gtk.c
@@ -2772,25 +2772,14 @@ void dt_gui_load_theme(const char *theme)
     char *c2 = usercsspath;
 
 #ifdef _WIN32
-    // for Windows, we need to remove the drive letter and replace '\' with '/'
+    // for Windows, we need to remove the drive letter and the colon, if present, and replace '\' with '/'
     c1 = strchr(path, ':');
     c1 = (c1 == NULL ? path : c1 + 1);
     c2 = strchr(usercsspath, ':');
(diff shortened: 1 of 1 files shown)


src/gui/gtk.c
```diff
@@ -2767,8 +2767,34 @@ void dt_gui_load_theme(const char *theme)
...
+    char *c2 = usercsspath;
+
+#ifdef _WIN32
+    // for Windows, we need to remove the drive letter and replace '\' with '/'
+    c1 = strchr(path, ':');
+    c1 = (c1 == NULL ? path : c1 + 1);
+    c2 = strchr(usercsspath, ':');
+    c2 = (c2 == NULL ? usercsspath : c2 + 1);
+
+    char *c = c1;
```
TurboGit, line 2781: I suppose this is equivalent to:
c1 = g_strdelimit (c1, "\\", "/");

src/gui/gtk.c
```diff
@@ -2767,8 +2767,34 @@ void dt_gui_load_theme(const char *theme)
...
+    while(*c != '\0')
+    {
+      if(*c == '\\') *c = '/';
+      c++;
+    }
+
+    c = c2;
```
TurboGit, line 2788: likewise.

Mark-64:
done, thanks for the suggestion, I didn't know that utility

PR #5441, opened 2020-06-12
po/fast pipe

TurboGit:
This is a new version to improve speed of perspective correction when using filmicrgb.

The patchset seems large but in fact most of the changes are changing:
`   pipe->type == CONST`
to
`   (pipe->type & CONST) == CONST`

There is a new IOP flags and a setting of DT_DEV_PIXELPIPE_FAST in pixelpipe_hb.c and the checking in filmicrgb.c.


TurboGit:
BTW, what is done in filmicrgb could probably also be done in some heavy modules to ensure better perfs when full accuracy is not needed.

ralfbrown:
Don't forget that just optimizing the heavy modules overall will also improve responsiveness.  And my pending PR #5192 for speeding up denoise(nonlocal means) already has a provision to be run even faster by evaluating only half as many patches when full precision is not required (thumbnail and preview).  fastpipe mode can easily turn that on as well.

TurboGit:
@parafin : I suppose this is what you had in mind when commenting my other PR ?

TurboGit:
A comparison on FilmicRGB for preview with and without the fast pipe:

`92,735164 [dev_pixelpipe] took 0,008 secs (0,033 CPU) processed `filmique rvb' on CPU, blended on CPU [preview/fast]`

`95,093831 [dev_pixelpipe] took 0,125 secs (0,597 CPU) processed `filmique rvb' on CPU, blended on CPU [preview]`

aurelienpierre:
Ahahah I was about to propose exactly that on the other PR when I saw you closed it.

I also think we should record, in the dev pipeline structure, which module captured the flag. For example `dev->pipe->fast_pixelpipe_requestor`. 

That way, on the first run, the modules requires the flag for a fast diagnostic pipeline (in this case, edges detection) and sets itself in `dev->pipe->fast_pixelpipe_requestor`. On the next run, diagnostic should have been completed, so if `dev->pipe->fast_pixelpipe_requestor == self` and `DT_DEV_PIXELPIPE_FAST`, that same module can disconnect itself and reset `DT_DEV_PIXELPIPE_FAST` in a clean way.

jenshannoschwalm:
> I also think we should record, in the dev pipeline structure, which module captured the flag. For example (dt_iop_module_t) dev->pipe->fast_pixelpipe_requestor.

Is it impossible or **not desired** to have more than just one module forcing the fast path? Or would it be better to have a counter instead of a boolean flag?

> Also, we could perhaps have a GUI button to run the fast path upon user request for users having lightweight computers

This would have to be switched off for export or mipmap cache?

aurelienpierre:
> Is it impossible or not desired to have more than just one module forcing the fast path? Or would it be better to have a counter instead of a boolean flag?

Not desired. That's a special case of handling that could trigger a shitload of problems if it's not tracked carefully. We need to be super sure that the fast mode will be disabled as soon as it's not needed anymore, and all affected caches are invalidated, or the cure will be worse than the disease. 

Remember every module caches its output, and a hash produced from user params of previous modules is produced to check if the cache is still valid or not (as in: check if user params have changed before in the pipe, and recompute caches if so). Since the fast path doesn't affect user params, the on/off state of the flag will go unnoticed from the cache perspective. So we can imagine cases where the perspective distortion will change, but caches are not invalidated, so the preview will be all garbled.

TurboGit:
I suppose that any tweaking on the fast pipe should come after 3.2 anyway. Let's keep this as simple as possible as the only goal at the moment is:
- fix the filmic rgb slowness in perspective correction
- try to come with an as generic as possible implementation on which we can build better circuitry later

TurboGit:
@rawfiner : do you think this new fast pipe mode could be used in the denoising modules ?

There is probably different strategies:
- makes the denoising a nop (copying input to output)
- use a set of params that are known to be fast (less iterations, smaller search area...)

aurelienpierre:
> I suppose that any tweaking on the fast pipe should come after 3.2 anyway. Let's keep this as simple as possible as the only goal at the moment is:

Sure. But what do you think about tracking the module that requires the fast path ?

TurboGit:
> Sure. But what do you think about tracking the module that requires the fast path ?

For now the module requiring the fast-path is `darktable.develop->gui_module` (the currently active module).

elstoc:
> For now the module requiring the fast-path is darktable.develop->gui_module (the currently active module).

Don't forget that changing sliders and comboboxes via keyboard shortcuts does not currently set gui_module.

TurboGit:
> Don't forget that changing sliders and comboboxes via keyboard shortcuts does not currently set gui_module.

So we will maybe have to refine this at some point. But for the current PR the gui_module is always set as we cannot change perspective correction via keyboard :)

Question, change changing sliders or combo via keyboard is there a variable set somewhere recording the module being activated or not at all?

elstoc:
> as we cannot change perspective correction via keyboard

You can change the parameter sliders/comboboxes without gui focus but presumably you're referring to the auto fit  and get structure buttons.

> change changing sliders or combo via keyboard is there a variable set somewhere recording the module being activated or not at all?

I don't understand the question. The accelerators just change the slider/combobox and emit the 'value-changed' signal. So the module is automatically activated (I assume) by the same code that activates it when you change the slider manually. I'm not 100% sure what code does that.

ralfbrown:
> Remember every module caches its output, and a hash produced from user params .... Since the fast path doesn't affect user params, the on/off state of the flag will go unnoticed from the cache perspective.

So the state of the fast path flag needs to be included in the hash.

TurboGit:
> So the state of the fast path flag needs to be included in the hash.

Any pointer where this should be happening ?

rawfiner:
I am not sure I understand the problem correctly.
Could you detail more?
Is the goal to be able to make changes in modules in the middle of the pipe more responsive by having "lightweight" version of all modules after them?

About denoising modules: you could do a nop, yes, or use downscaling-upscaling to make it faster (less accurate than full denoising, but better than nop in terms of preview)

parafin:
> @parafin : I suppose this is what you had in mind when commenting my other PR ?
This is what I had in mind, yes. But caching issue indeed needs to be checked carefully. dt_dev_pixelpipe_cache_hash function computes the hash, so it should be modified, but we also have mipmap cache of the whole output, so fast version shouldn't get in that one I guess.

TurboGit:
@parafin : thanks for the feedback. My latest commit answer you first point. I'm not sure to get the mipmap cache issue where fast version should be ignored.

TurboGit:
@rawfiner : the fast pipe mode when set says to all module "please if possible just do a minimal set of action to ensure fast response". The idea is to make some interactive modules like changing crop in "perspective correction" usable. See the initial report #5423, the default filmicrgb setting makes moving the crop area impossible. This is because "perspective correction" come before filmicrgb in the pipe. So anyway, my question does not stand at the moment as all the noise reduction modules are very early in the pipe, so not an issue for the case at hand in fact.

Just to be complete, here is how it works:
- Every module can set the flag IOP_FLAGS_ALLOW_FAST_PIPE
- When this module is active (used) the pipe is set in fast mode
- Every other module is then allowed to do a fast job instead of a full and possible slow one
- When switching to another module or collapsing the module the fast-mode is deactivated and the normal "slow" pipe is then active again.

Hope this clarifies the goal.

MStraeten:
denoise profiled should be taken into account since panning and zooming in modules later in the pipe results in a massive slowdown due to recalculation of denoise. So it might make sense to disable it or use a simplified algo when in zoomed views while tweaking other modules ...

rawfiner:
@TurboGit much clearer thanks :-)
@MStraeten The problem is, being zoomed in is the only way to have a preview that is completely accurate for denoising, and we should keep such accuracy. Though I agree that it is not very convenient for tweaking other modules. Best thing in my opinion would be to have faster denoising algorithms (I have ongoing work on that, and ralfbrown's optimisations will help a lot too), but I don't know up to which point it will be possible. Maybe denoising should be performed first on the area seen at screen, and then be computing the full denoise image in the background to have that in cache at some point? (I don't know how feasable that is)

johnny-bit:
> Maybe denoising should be performed first on the area seen at screen, and then be computing the full denoise image in the background to have that in cache at some point

that would be awesome!

ralfbrown:
Fast pipe will be great for moving the sliders in ashift and clipping (and likely other places none of us has thought of yet), but it occurs to me: why is moving/resizing an overlay on top of the image causing any recomputation **at all**?  AFAICT, what gets passed down the pixel pipe doesn't change with varying crops until the final selected crop is applied when the iop is closed or otherwise loses focus, unless you're zoomed in and you pan the viewport.

TurboGit:
@ralfbrown : I have asked myself the same question but I suppose this is because in crop module you can also rotate and so need to recompute everything.

ralfbrown:
Sure, if you change the rotation, you need to recompute.  But if you're just adjusting the crop box, that isn't actually necessary, as shown in my PR.  Another one coming momentarily with a 90% fix for ashift (someone with a better knowledge of the GUI coding will need to look at it).

TurboGit:
Just rebased and ready to be merged now.

TurboGit:
Let's bite the bullet...

aurelienpierre:
I don't know if the project of adding denoise to fast pipe has been completed yet, but having denoise disabled might lead to discrepancies in automatic features like RGB min/max or edges detection.

All in all, I think the fast pipe mode should only apply for modules applied **after** the module that required the fast pipe, to ensure the consistency of the input for that module.

TurboGit:
> I don't know if the project of adding denoise to fast pipe has been completed yet,

No. Only **filmicrgb** and **grain** have a special fast pipe.

And note that this is not a generic pipe used all the time. It is only activated by the `clipping` and `perspective correction` modules. (see setting of IOP_FLAGS_ALLOW_FAST_PIPE in iop modules).

aurelienpierre:
Ok great !

PR #5445, opened 2020-06-13
Better icon themes in Windows HiDPI

Mark-64:
With this PR the icon themes will look better on windows HiDPI, as margins are moved from pixels to em units.
See after (bloom) and before (lowpass)
Cattura7
Needs a small change in darktable.css for margins of the always enabled button, in order to align the svg icon.
@Nilvus Please have a look.

### DIFF
--- a/data/themes/darktable-icons.css
+++ b/data/themes/darktable-icons.css
@@ -18,22 +18,88 @@
 
 @import url("darktable.css");
 
-#iop-panel-icon-sharpen
+#iop-panel-icon-ashift,
+#iop-panel-icon-atrous,
+#iop-panel-icon-basecurve,
+#iop-panel-icon-basicadj,
+#iop-panel-icon-bilat,
+#iop-panel-icon-bilateral,
+#iop-panel-icon-bloom,
+#iop-panel-icon-borders,
+#iop-panel-icon-cacorrect,
+#iop-panel-icon-channelmixer,
+#iop-panel-icon-clahe,
+#iop-panel-icon-clipping,
+#iop-panel-icon-colisa,
+#iop-panel-icon-colorbalance,
+#iop-panel-icon-colorchecker,
+#iop-panel-icon-colorcontrast,
+#iop-panel-icon-colorcorrection,
+#iop-panel-icon-colorin,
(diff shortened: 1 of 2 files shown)


TurboGit:
Nice ! Waiting for @Nilvus review.

data/themes/darktable-icons.css
```diff
@@ -23,511 +23,579 @@
     background-image: url('../pixmaps/plugins/darkroom/sharpen.svg');
     background-size: contain;
     background-repeat: no-repeat;
-    padding-left: 18px;
-    margin: 3px 0px;
+    margin-top: 0.13em;
```
Nilvus, line 26: We need to be careful about optical margins. 0.14em is better I think here. I find the icons a little (optically) too up regardless of toggle buttons beside.

data/themes/darktable.css
```diff
@@ -562,6 +562,8 @@ overshoot.right
 {
   color: @plugin_label_color;
   border: 0;
+  padding: 0.06em;
+  margin: 0 2px 0 1px;
```
Nilvus, line 566: Not needed. Just remove those lines for me. I've made many tests and I see that I can move the icon, reduce it but not increase the size. And that's the issue: the always icon maximum size. Probably something on Gtk or Cairo code?

Nilvus:
Just an add: I just remember that the always icon size is maybe related to recent focus icon change. If you make focus on an icon, you will see that this time, the module icon became smaller than always icon. So, no possibility to have all time always icon and module icon on same size/alignment. Or that would need to change how the focus is made. Maybe enlighten the always icon (plain circle inside) when focus is made would be better and consistent to other icons? @jenshannoschwalm and @elstoc: your view on that as you work (if my memory is good) on focus icons?

jenshannoschwalm:
> Maybe enlighten the always icon (plain circle inside) when focus is made would be better and consistent to other icons?

Yeah, sound good for me too. Or have the outer circle thicker for on?

elstoc:
I don't have a problem with that.

Mark-64:
Changes requested are implemented.

TurboGit:
All good, thanks for implementation and review. Merging.

PR #5446, opened 2020-06-13
let gradient_slider follow gui->reset when sending value-changed signals

dterrahe:
Since I didn't get a response to https://github.com/darktable-org/darktable/pull/5395#issuecomment-643102351 I prepared a quick PR to show what would be the consistent approach to determining whether the gradient slider should fire off a value-changed signal.

If there is a reason to not follow this pattern (that both bauhaus combobox and slider use) then please explain.

I believe this PR shouldn't change the behavior, except in relight::gui->update where the extended dtgtk_gradient_slider_set_value was called with a TRUE flag. Was this intentional and needed? If not, it is the kind of inconsistency that using the global gui_reset flag makes easier to avoid...

### DIFF
--- a/src/develop/blend_gui.c
+++ b/src/develop/blend_gui.c
@@ -774,8 +774,8 @@ static void _blendop_blendif_update_tab(dt_iop_module_t *module, const int tab)
   dt_pthread_mutex_lock(&data->lock);
   for(int k = 0; k < 4; k++)
   {
-    dtgtk_gradient_slider_multivalue_set_value(data->lower_slider, iparameters[k], k, TRUE);
-    dtgtk_gradient_slider_multivalue_set_value(data->upper_slider, oparameters[k], k, TRUE);
+    dtgtk_gradient_slider_multivalue_set_value(data->lower_slider, iparameters[k], k);
+    dtgtk_gradient_slider_multivalue_set_value(data->upper_slider, oparameters[k], k);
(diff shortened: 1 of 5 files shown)


TurboGit:
@dterrahe : Sorry I missed your question... I've now made a very helpful answer :)

dterrahe:
:-) No worries, I know you are a _very_ busy man these days!

@Mark-64 what do you think; will gui->reset work for you? If so, could you please test this PR, since I'm not very familiar with the relight and retouch modules, so don't know what are the corner cases that maybe you were working around.

Mark-64:
@dterrahe Sorry I din't see your other comment
I will check.

Mark-64:
Checked and it's OK.
Before I reworked the gradient slider, the signal was raised unconditionally and it was impossible to use those functions in callbacks. So I introduced that additional parameter. I didn't know of the gui->reset mechanism.
Sorry about that, I am still rather new to the darktable code base

dterrahe:
No worries at all! I'm relatively new too but starting to know some parts of the code base rather well and gui->reset is one of them. I'm glad you are jumping in and perfectly happy to help you learn where I can. Your cleanup is nice and useful; we need more of those!

@TurboGit would you please merge this small PR. It will safe me having to resolve one merge conflict with #5335 ...

TurboGit:
Works on my side, thanks!

dterrahe:
Thank you!

PR #5449, opened 2020-06-13
allow all modules that can be disabled to be removed from history

### LINKED ISSUE #5434: FR: More Consistent Automatic Modules
This Feature Request is to discuss how automatic modules are handled within darktable and how those modules interact with the history stack. The intention is to agree how to handle the different types of automatic modules in a consistent way. It follows on from some of the issues mentioned in #5396

**Forced On Modules**

Certain modules are mandatory and cannot be switched off. For example, for raw files the following are included in this category:

- raw black/white point
- demosaic
- input color profile
- output color profile
- display encoding

These modules can be identified by the fact that ``(module->default_enabled && module->hide_enable_button)`` is TRUE.

**Enabled By Default Modules**

Some modules are highly recommended for the production of a reasonable image and so are always switched on by default. While these modules can be switched off they cannot be removed from the history stack. This is odd and leads to some of the issues mentioned in #5396. For example:

- white balance
- highlight reconstruction
- orientation

Such modules can be identified by the fact that ``(module->default_enabled && !module->hide_enable_button)`` is TRUE.

**Modules Enabled via Preset / Preference**

These are modules that are enabled via a preference setting (scene- or display-referred workflow modules or sharpen) or via an automatically-applied user preset. These modules are only auto-applied when editing a new image or clearing the history stack. They may be manually disabled and removed from the history stack entirely.

**So What's the Problem?**

I've partially resolved some of the issues in #5414. As part of that PR, the history stack initialisation procedure was fixed such that the above categories of module are applied to the history stack in order (with the Forced On modules at the bottom). So this discussion starts with the assumption that you've loaded a fresh image with the history stack applied in that order.

_1. Selecting Auto-Generated Items on the History Stack_

The normal behaviour expected of the history stack is that if you click on a number in the stack, all of the modules referenced below it are applied (switched on) and all of the modules above it are not (switched off). However this is not true of the modules in the first two categories above. For example take the following default history stack:

Screenshot_2020-06-12_16-38-11

If I select any item from 0 to 7 the preview image does not change and the active modules list also does not change. While this makes sense for the 'Forced On' modules (0-5), it really doesn't make sense for the others, since they can be switched off manually. Orientation is a bit of an outlier here as it remains on throughout the whole stack but for some reason (probably because it's auto-set from the image) is added at the end.

_2. Removing Items from the History Stack_

Any module that can be manually switched off should be able to also be removed from the history stack. However, this is not possible for the modules in the second group (Enabled by Default). You can disable them but if you try to remove them from the stack they pop back in (enabled).

**Proposal**

Firstly, it should not be possible to click any of the 'Forced On' modules in the stack, except for either the first or possibly the last one (for the purposes of discarding the stack). They should be forced to the beginning of the stack (even for existing edits if possible) and should not ever move from there. 

Second, the 'Enabled by Default' modules should work the same way as the 'Enabled by Preset/Preference' modules, though by default they should be added to the stack before any preset/preference modules. As with preset/preference modules they should only be forced back onto the stack if the stack is fully discarded (by selecting the last module in the 'forced on' part of the stack and compressing).

elstoc:
modules that are enabled by default (module->default_enabled) but
can be manually disabled (!module->hide_enable_button) can now
be removed from history

when selecting items in the history stack prior to such modules, the
stack and image treat these modules as disabled

does not impact modules that are forced on and cannot be disabled (i.e.
those with module->hide_enable_button set)

Resolves #5434

### DIFF
--- a/src/develop/develop.c
+++ b/src/develop/develop.c
@@ -1402,8 +1402,6 @@ void _dev_insert_module(dt_develop_t *dev, dt_iop_module_t *module, const int im
 
 static gboolean _dev_auto_apply_presets(dt_develop_t *dev)
 {
-  // NOTE: the presets/default iops will be *prepended* into the history.
-
   const int imgid = dev->image_storage.id;
 
   if(imgid <= 0) return FALSE;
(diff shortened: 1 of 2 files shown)


TurboGit:
If this is ok with you I'd like to target this for 3.4.

elstoc:
That's what I expected. It's fine with me.

TurboGit:
Perfect !

aurelienpierre:
tested, it works

elstoc:
My OCD still wants the orientation module much earlier in the stack though

elstoc:
I've now amended this so that default/presets are appended instead of prepended to the history. I haven't found any ill effects through testing it and I can't find any good reason that it should have been done in the first place.

aurelienpierre:
It does not remove all the disabled modules though (for example, chromatic aberrations or local contrast). Is that intentional ?

elstoc:
This PR just _allows_ you to remove modules by turning them off then compressing history. You still have a bit of work to do: disable the module, compress history, select the module before the disabled one and then compress again. But at least this way the disabled modules don't pop back onto the stack, which is what happened before this PR.

PR #5453, opened 2020-06-13
fix duplicate preview in darkroom

### LINKED ISSUE #4638: duplicate manager - inconsistent processing

aurelienpierre:
synchronize Cairo drawing on the main darkroom window to:

1. get the same image resolution (so far, duplicate preview were too blurry for 1:1 comparison)
2. support color assessement mode
3. ensure main and duplicate preview are the same size

~NOT WORKING SO FAR (I need help):~
~When zoomed-in, the duplicate preview surface is not shifted correctly. Can't figure out where the mistake is. Faulty lines are:~
```
const float dx = -zoom_x * img_wd - (img_wd - 2. * tb) * nz / 2. + (img_wd - 2. * tb) / 2.; 
const float dy = -zoom_y * img_ht - (img_ht - 2. * tb) * nz / 2. + (img_ht - 2. * tb) / 2.;
```
~some scaling is not right.~

Fixed thanks to @AlicVB

Fixes #4638

### DIFF
--- a/src/libs/duplicate.c
+++ b/src/libs/duplicate.c
@@ -207,12 +207,24 @@ void gui_post_expose(dt_lib_module_t *self, cairo_t *cri, int32_t width, int32_t
   dt_lib_duplicate_t *d = (dt_lib_duplicate_t *)self->data;
   if (d->imgid == 0) return;
   dt_develop_t *dev = darktable.develop;
-  if(!dev->pipe->output_backbuf || dev->image_status != DT_DEV_PIXELPIPE_VALID) return;
+  if(!dev->preview_pipe->backbuf || dev->preview_status != DT_DEV_PIXELPIPE_VALID) return;
 
   // use the same resolution as main previem image to avoid blur
-  float img_wd = dev->pipe->output_backbuf_width / darktable.gui->ppd;
-  float img_ht = dev->pipe->output_backbuf_height / darktable.gui->ppd;
+  float img_wd = dev->preview_pipe->backbuf_width;
+  float img_ht = dev->preview_pipe->backbuf_height;
   const int32_t tb = darktable.develop->border_size;
+
+  // we rescale the sizes to the screen size
+  if (img_ht * (width - 2 * tb) > img_wd * (height - 2 * tb))
+  {
+    img_wd = img_wd*(height - 2 * tb)/img_ht;
+    img_ht = (height - 2 * tb);
+  }
+  else
+  {
+    img_ht = img_ht*(width - 2 * tb)/img_wd;
+    img_wd = (width - 2 * tb);
+  }
 
   // Get the resizing from borders - only to check validity of mipmap cache size
   float zoom_ratio = 1.f;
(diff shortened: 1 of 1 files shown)


elstoc:
N.B. this will resolve #4638 if merged.

AlicVB:
> @AlicVB do you have an idea ?

yes, buying an hidpi screen ;)

I'll have a look (at the code, not the screens) :)

aurelienpierre:
Ready for review and merging

elstoc:
Doesn't quite do it for me I'm afraid - the two images still look different and theres some subtle pixel-shifting going on and brightness/contrast changes. Here's an example where both images are identical duplicates of each other:

Zoomed in (current image)
1zoomed_before

Zoomed in (duplicate selected)
2zoomed_after

Zoomed out (current image)
3_fullscreen_before

Zoomed out (duplicate selected)
4_fullscreen_after

aurelienpierre:
brightness/contrast changes come with Cairo interpolation, I'm afraid I won't be able to do anything here. What happens is the original preview as computed 1:1 by dt, while the duplicate is choosen from the closest mipmap cache size and interpolated to final size by Cairo.

Pixel shifting is hard to solve, I thought I got rid of it since I don't see it on my screen. Basically it's a matter of ensuring all sizes computations are made the same and rounded exactly the same too which, given the number of intermediate steps involved, is a PITA.

aurelienpierre:
Also you can test various interpolation methods for Cairo by setting in darktablerc : `ui/cairo_filter=best/good/fast`

elstoc:
Thanks @aurelienpierre. ``fast`` is certainly by far the worst. Not only does the whole image shift but bits of it seems to warp and it just seems really odd. Changing it to ``good`` or ``best`` does fix that to a large extent, though there's still the issue that the entire image shifts by a few pixels and it's still not as sharp as the normal darkroom view.

That combined with contrast and brightness differences means it won't be good for testing subtle changes between edits (I would mostly want to use this to assess differences between filmic versions). So I guess I might be back to good old snapshots for that.

AlicVB:
about pixelshift etc : imo that will be difficult to avoid as we don't use different path in pipe with different options for the "regular" image in darkroom and for the duplicate preview.
Maybe we can change that to get exactly same outputs, but that will be quite a challenge and quite heavy for cpu !

jenshannoschwalm:
Some thoughts, although i have never used the duplicate manager personally before, so i might be "out-of-fucus"

If we use the written mipmaps, there might be problems resulting from the situation we wrote them. 1. compresses jpeg, dynamic range, added noise, compression artefacts, all would lead to subtle differences, rescaling will make that much worse.
2. If we write mipmaps, are we sure to use a decent jpeg quality? This is taken from exports default...
3. If we write mipmaps the export does **not** use high quality mode if the image is smaller than the pipeline size.
4. pixelshift might be the result of subtle upscaling while writing the mipmap. I can't see this on the screenshots, but is this worse on a regular pattern?

ideas?
1. When preparing for duplicate comparing, would it be better to have a 16bit image to compare?
2. enforce high resolution mode for larger mipmaps >= 4 maybe?

aurelienpierre:
@jenshannoschwalm since we are comparing output to output, both mipmap and original image are 8 bits with reduced dynamic range, so it's no issue on that regard. However, scaling and interpolation discrepancies are an issue, but the only way to fix it would be to bypass all Cairo filters with our own.

jenshannoschwalm:
Aah, thanks for explaining output vs output.

TurboGit:
I still have a little bit of shift (no HiDPI screen on my side) but less important. So this goes into the right directly for me. Thanks.

PR #5454, opened 2020-06-13
filmicrgb: separate highlight reconstruction mask button

elstoc:
for consistency with other modules and to allow it to be given its own @aurelienpierre style tooltip, place the show highlight reconstruction mask button in its own line

### DIFF
--- a/src/iop/filmicrgb.c
+++ b/src/iop/filmicrgb.c
@@ -178,6 +178,7 @@ typedef struct dt_iop_filmicrgb_gui_data_t
   GtkWidget *grey_point_source;
   GtkWidget *black_point_source;
   GtkWidget *reconstruct_threshold, *reconstruct_bloom_vs_details, *reconstruct_grey_vs_color, *reconstruct_structure_vs_texture, *reconstruct_feather;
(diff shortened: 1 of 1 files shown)


aurelienpierre:
Not sure, it will only increase the height of the module to add little meaning.

elstoc:
The problem is there's no way to really know what it is doing without RTFM and the tooltip is misleading because it relates to the second slider, even though the button relates to both of them.

aurelienpierre:
Hmm fair enough.

elstoc:
Only other alternative would be to be able to set a separate tooltip on the slider and quad but looking at ``bauhaus.c`` I don't think that's currently possible.

aurelienpierre:
Clicking on the button if module is disabled doesn't enable the module anymore, though.

elstoc:
Not in this branch but it should do if it's merged (it's still running the same callback function). I can rebase off current master if you like.

elstoc:
Tested after a rebase and pushed again - seems to work fine.

TurboGit:
Works for me, thanks!

PR #5457, opened 2020-06-13
Properly raise DT_SIGNAL_IMAGE_INFO_CHANGED for compressed history.

### LINKED ISSUE #5455: Lighttable thumbnails are not refreshed when history is compressed
1. In lighttable, select an image
3. hover the "changed" icon, note the modules (especially if some are duplicated)
2. click history stack -> compress history
3. hover the "changed" icon, the history stack has not changed in the list.

TurboGit:
Fixes #5455.

### DIFF
--- a/src/common/history.c
+++ b/src/common/history.c
@@ -1049,6 +1049,9 @@ void dt_history_compress_on_image(int32_t imgid)
   }
   dt_unlock_image(imgid);
   dt_history_hash_write_from_history(imgid, DT_HISTORY_HASH_CURRENT);
+
+  GList *imgs = g_list_append(NULL, GINT_TO_POINTER(imgid));
+  dt_control_signal_raise(darktable.signals, DT_SIGNAL_DEVELOP_MIPMAP_UPDATED, imgs);
 }
 
 int dt_history_compress_on_list(GList *imgs)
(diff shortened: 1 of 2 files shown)


jenshannoschwalm:
Yes, this is a fix for the issue somehow. But i really don't think it's a good one.

1. It forces a mipmap recalculation which is only necessary if i selected something in history and compressed so the image might be different.
2. For bulk history not necessary but would generate tons of signals.
3. I didn't test but i would expect the mipmap re-generation start fighting with image mutexes and of cource cpu-cycles.

I think the signal should be in `_lib_history_compress_clicked_callback` instead.

TurboGit:
@jenshannoschwalm : I have followed what is done for copy/paste. If you put this in the callback there is chance that this won't work in other context. For example when calling compress from Lua.


>     3\. I didn't test but i would expect the mipmap re-generation start fighting with image mutexes and of cource cpu-cycles.

No, the MIPMAP_UPDATED signal means that the mipmap has been updated and just update the thumbnail information (altered status). So that's a somewhat light call.

TurboGit:
I see it also makes the image surface dirty and so force a computation.

@AlicVB : what is the way to only update the altered status ?

TurboGit:
Ok, I'm now using DT_SIGNAL_IMAGE_INFO_CHANGED which is more lightweight.

TurboGit:
This latest version be better and answer your concerns about performances @jenshannoschwalm .

jenshannoschwalm:
Perfect

TurboGit:
Thanks for the reviews!

PR #5458, opened 2020-06-13
update images in collection on tags_changed if filtering by tags

### LINKED ISSUE #5447: collection not refreshed when changing tags
**Describe the bug**
Removing a tag when the collection is set to display pictures with this tag does not remove the picture from the collection.

**To Reproduce**
1. Go to lighttable
2. Click on collect, select tags
3. Click on a tag to display a collection of pictures with this tag
4. Select one of the picture and detach the corresponding tag
5. See, the picture is still displayed in the collection

**Expected behavior**
As for color labels, I expect the picture to be removed from the collection.

 - Darktable Version: master
 - OS: GNU/Debian

**Additional context**

johnny-bit:
When collection was filtered by tags, any change in tags wouldn't reflect in collected images. now it does.

this fixes #5447

I took extra steps not to raise too many signals.

### DIFF
--- a/src/libs/collect.c
+++ b/src/libs/collect.c
@@ -2326,13 +2326,36 @@ static void tag_changed(gpointer instance, gpointer self)
 {
   dt_lib_module_t *dm = (dt_lib_module_t *)self;
   dt_lib_collect_t *d = (dt_lib_collect_t *)dm->data;
-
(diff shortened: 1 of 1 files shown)


TurboGit:
Works for me now, thanks !

PR #5459, opened 2020-06-13
Always-on icons should be cairo-painted the same size as others.

jenshannoschwalm:
Related to #5445

Not in focus
Bildschirmfoto von 2020-06-13 23-17-28

In Focus
Bildschirmfoto von 2020-06-13 23-17-22

Still good to distinguish and nothing new ...


aurelienpierre:
Thanks !

PR #5462, opened 2020-06-13
add keyboard shortcut to toggle tooltip visibility

### LINKED ISSUE #5382: Tool Tips are frequently in an awkward and interferring position expecially over the histogram.
Tool Tips are frequently in an awkward and interferring position expecially over the histogram.

Please provide a way to control their display, either by length of time or toggling on/off, ...

openSUSE Tumbleweed 20200609
darktable 3.1.0+2050~g02678f4d9
OpenCL on 
i7 12-core 36GB

elstoc:
allows tooltip visibility to be toggled via a global keyboard shortcut
(defaults to SHIFT-T)

as this functionality requires window compositing to be enabled on X11
systems an error is displayed if compositing is disabled

Resolves #5382

### DIFF
--- a/data/themes/darktable.css
+++ b/data/themes/darktable.css
@@ -817,6 +817,7 @@ popover,
 combobox window,
 dialog combobox window
 {
+  opacity: 1;
   background-color: @tooltip_bg_color;
   outline-style:none;
   border-radius: 4px;
--- a/src/gui/gtk.c
+++ b/src/gui/gtk.c
@@ -25,6 +25,7 @@
 #include "common/file_location.h"
 #include "common/image.h"
 #include "common/image_cache.h"
+#include "bauhaus/bauhaus.h"
 #include "develop/develop.h"
 #include "develop/imageop.h"
 #include "dtgtk/button.h"
@@ -278,6 +279,29 @@ static gboolean view_switch_key_accel_callback(GtkAccelGroup *accel_group, GObje
   return TRUE;
 }
 
+static gboolean toggle_tooltip_visibility(GtkAccelGroup *accel_group, GObject *acceleratable,
+                                               guint keyval, GdkModifierType modifier, gpointer data)
+{
+  if(gdk_screen_is_composited(gdk_screen_get_default()))
+  {
+    gboolean tooltip_hidden = !dt_conf_get_bool("ui/hide_tooltips");
+    dt_conf_set_bool("ui/hide_tooltips", tooltip_hidden);
+    if(tooltip_hidden)
+      dt_toast_log(_("tooltips off"));
+    else
+      dt_toast_log(_("tooltips on"));
+  }
+  else
+  {
+    dt_conf_set_bool("ui/hide_tooltips", FALSE);
+    dt_control_log(_("tooltip visibility can only be toggled if compositing is enabled in your window manager"));
+  }
+
+  dt_gui_load_theme(dt_conf_get_string("ui_last/theme"));
+  dt_bauhaus_load_theme();
+
+  return TRUE;
+}
 static gboolean _focuspeaking_switch_key_accel_callback(GtkAccelGroup *accel_group, GObject *acceleratable,
                                                guint keyval, GdkModifierType modifier, gpointer data)
 {
@@ -1377,6 +1401,12 @@ int dt_gui_gtk_init(dt_gui_gtk_t *gui)
 
   // accels window
   dt_accel_register_global(NC_("accel", "show accels window"), GDK_KEY_h, 0);
+
+  // View-switch
+  dt_accel_register_global(NC_("accel", "toggle tooltip visibility"), GDK_KEY_T, GDK_SHIFT_MASK);
+
+  dt_accel_connect_global("toggle tooltip visibility",
+                          g_cclosure_new(G_CALLBACK(toggle_tooltip_visibility), NULL, NULL));
 
   darktable.gui->reset = 0;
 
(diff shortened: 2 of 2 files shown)


johnny-bit:
You could listen to preferences change signal and do same thing as in key event making it no longer restart required...

AxelG-DE:
>same thing as in key event making it no longer restart required...

As general theme changes also work right away...?

Nilvus:
Add a default keyboard would be a good thing. On what I test, I would suggest 'Maj+t' as it's the only not used keyboard shortcut related to letter 't' available. Could you add that default shortcut to your PR. And yes, it would be good to not have to restart when using checkbox on preferences like with shortcut.

elstoc:
> You could listen to preferences change signal

No need for that - I could run the theme loader just before that signal is emitted. The downside is that when changing themes/CSS/fonts on the general tab it would mean the theme was reloaded a second time. Also I wasn't sure if the preference was really desirable once the shortcut worked. Finally I wanted to keep the code small in the hopes that it might be considered for 3.2 (since it's really a pretty tiny change).

> On what I test, I would suggest 'Maj+t' as it's the only not used keyboard shortcut related to letter 't' available.

Excuse my lack of knowledge but what is 'Maj'? Do you mean ALT or CTRL?

Nilvus:
> Excuse my lack of knowledge but what is 'Maj'? Do you mean ALT or CTRL?

Oups, sorry. Maj is french word for Shift!

elstoc:
Ah ok. I don't have an issue with shift+t as a default. Doesn't seem to conflict with anything else. The question is whether it needs a preference as well. Given that the panel toggles don't have one probably the answer is no and I should just remove this.

elstoc:
As noted by @parafin in #5382, this only works if window compositing is enabled, but the alternative approach suggested by @TurboGit (wrap all tooltip creation into a specific dt routine and actually register the tooltip when a pref is set to TRUE) might not be worth the effort for a small feature. 

So the options as I see it are 

- close this PR
- make it so that it only actually does anything if compositing is enabled
- merge as is and accept that it won't work with compositing off

ptilopteri:
I vote for "make it so ..."

elstoc:
So do I. I would find it really useful to be able to switch on tooltips at will. Also I would assume (without any data to back this up) that people not using a compositing window manager are probably in the minority.

parafin:
> Also I would assume (without any data to back this up) that people not using a compositing window manager are probably in the minority.

You are probably right. But for the sake of more data it would be interesting to know if it works on Windows and macOS. Most likely it does, but who knows how GTK is implemented there.

elstoc:
Documentation for ``gtk_widget_set_opacity`` suggests it should be fine on Windows:

> For toplevel widgets this depends on the capabilities of the windowing system. On X11 this has any effect only on X screens with a compositing manager running. See gtk_widget_is_composited(). On Windows it should work always, although setting a window’s opacity after the window has been shown causes it to flicker once on Windows.

Nilvus:
> Ah ok. I don't have an issue with shift+t as a default. Doesn't seem to conflict with anything else. The question is whether it needs a preference as well. Given that the panel toggles don't have one probably the answer is no and I should just remove this.

Shortcut is more suitable for that as we could, sometimes, need to check a tooltip for rarely used modules or new ones. and going to preferences, then miscellanious, check or uncheck and restart instead of just pressing a shortcut...

So I would be to remove that option but maybe some will want it... and only keep shortcut.

About window compositing, I also think this would be a really rare case, so your option 2 (make it so) it's the best. And I would also say that release 3.2 is just in 2 months, so that let us 2 months of field testing and we could if needed revert back and think of the other option suggested by @TurboGit if needed.

If this is merge, I think that little transparency I add on my PR is not really useful. I'm quite mixed in reality about that and your PR is a better option. Of course, something I add to improve PR (margin suggested by @johnny-bit or little tweaks I made) is better but hide/show with shortcut is better than a little transparency.

elstoc:
TBH I really only added the preference when I thought the shortcut wasn't going to work and then was too lazy to remove it again. I'm going to remove the preference and add the test for compositing to this PR, so that it's complete in case it gets marked for inclusion in 3.2.

Nilvus:
> TBH I really only added the preference when I thought the shortcut wasn't going to work and then was too lazy to remove it again. I'm going to remove the preference and add the test for compositing to this PR, so that it's complete in case it gets marked for inclusion in 3.2.

I think it could be for 3.2. Feature freeze is only tomorrow and it's a minor add. Anyway, I said to remove preference but maybe some would want it on preferences (it's as you want).

elstoc:
Updated and ready to go

elstoc:
This is confirmed working on macOS (Catalina)

TurboGit:
Sorry I'm a bit late but the preference is necessary to me. In fact if one need to deactivate the tooltip to never see them again it means that it is now required to add a key shortcut just for that. A preference would be quite handy in such case.

TurboGit:
I see that you have added a `Shift-T` shortcut. I would even go with `Ctrl-Shift-T` to avoid using a simple shortcut for a possible rare action.

TurboGit:
Arg... Ctrl-Shit-T is already used for toggling top bar :(

Nilvus:
> Arg...` Ctrl-Shift-T` is already used for toggling top bar :( Can we use `Ctrl-Shift-Alt-T` ? Maybe too much :)

Too much for me. I suggest Shift+t because it was the only one with 2 or 3 keys with T available.

TurboGit:
> Too much yes. I suggest Shift+t because it was the only one with 2 or 3 keys with T available.

So maybe reintroduce a preference for people reusing `Shift-T` for other action ?

Nilvus:
> > Too much yes. I suggest Shift+t because it was the only one with 2 or 3 keys with T available.
> 
> So maybe reintroduce a preference for people reusing `Shift-T` for other action ?

Not sure is needed. As all shortcuts, it can be change on shortcuts category in preferences.

TurboGit:
> Not sure is needed. As all shortcuts, it can be change on shortcuts category in preferences.

Ok, let's go with that then, it works fine on my side. There is not even a "blip", the change is done fast, not blinking. That will be an improvement for people not wanting tooltips.

dterrahe:
Scenario; I'm a novice user looking (or half-remembering) the shortcut to switch off the top bar. I accidentally press Shift-T before finding CTRL-SHIFT-T.  I overlook the brief toast that notifies me tooltips are now off. From now on, I'm not seeing any tooltips anywhere, but I don't even know I was supposed to, and I just decide dt is very unhelpful and user unfriendly...

I have not good suggestions to avoid this scenario; I'm just worried about it :-)

Also, you want to switch tooltips off because you can't see values/sliders behind them, right? And you are using opacity to regulate this anyway; would 50% opacity be a useful setting? Could the shortcut cycle between 100%/75%/50%/0%? I'm just spit-balling here; feel free to ignore if stupid.

Nilvus:
@elstoc: just see that keyboard shortcut do not work on preferences window.

@dterrahe: I find hard to say dt is very unhelpful and user unfriendly just for a shortcut. And you precise you have see notification. Friendly way is to have the possibility to revert back something like that by pressing same shortcut. And that the way this work.
And no, it's not only for those who don't really see sliders, it's also for those who don't need tooltips, so they can be disabled permanently but by letting possibility to show them again (for example if new feature/module or rarely use ones).
Opacity is here only to quickly show or hide tooltips. And if you check my PR #5438 you will see that I try to use different opacity but that makes them quite less readable. Even with 80% of opacity it's not always good, so 50% will be really not usable.

dterrahe:
@Nilvus Yep, I agree not seeing the notification and not thinking to search for an option to turn something back on that you didn't even know you wanted is pretty stupid. That's why I called them a "novice" user. Novice is a kinder word for stupid. I've been in that position myself, so I know what I'm talking about :-)

And apologies that my little thought on opacity was stupid and badly researched as well.

elstoc:
>  just see that keyboard shortcut do not work on preferences window.

I'm not sure any keyboard shortcuts work when the preferences window is open do they?

Nilvus:
> > just see that keyboard shortcut do not work on preferences window.
> 
> I'm not sure any keyboard shortcuts work when the preferences window is open do they?

I don't know and I even don't which one can be used when preferences window open. So that's not really an issue, we could do without.

@dterrahe: no problem.

PR #5463, opened 2020-06-14
eliminate redundant pixelpipe recomputation while adjusting crop box

### LINKED ISSUE #4797: [crop&rotate] extraneous pixelpipe runs

ralfbrown:
I believe that this fixes the full pixelpipe runs when merely moving or resizing the crop box in the clipping iop.  Seems to behave correctly, even when zoomed in so that only part of the crop box is visible, but others should beat on this a bit first to see if there are any corner cases which need to be handled.

See also the discussion in #5441.
Fixes #4797.

### DIFF
--- a/src/iop/clipping.c
+++ b/src/iop/clipping.c
@@ -3143,17 +3143,21 @@ int mouse_moved(struct dt_iop_module_t *self, double x, double y, double pressur
         dt_dev_pixelpipe_iop_t *piece = dt_dev_distort_get_iop_pipe(self->dev, self->dev->preview_pipe, self);
         if(piece)
         {
-          p->cx = points[0] / (float)piece->buf_out.width;
-          p->cy = points[1] / (float)piece->buf_out.height;
-          p->cw = copysignf(points[2] / (float)piece->buf_out.width, p->cw);
-          p->ch = copysignf(points[3] / (float)piece->buf_out.height, p->ch);
+          // only update the sliders, not the dt_iop_clipping_params_t structure, so that the call to
+          // dt_control_queue_redraw_center below doesn't go rerun the pixelpipe because it thinks that
+          // the image has changed when it actually hasn't, yet.  The actual clipping parameters get set
+          // from the sliders when the iop loses focus, at which time the final selected crop is applied.
+          float p_cx = points[0] / (float)piece->buf_out.width;
+          float p_cy = points[1] / (float)piece->buf_out.height;
+          float p_cw = copysignf(points[2] / (float)piece->buf_out.width, p->cw);
+          float p_ch = copysignf(points[3] / (float)piece->buf_out.height, p->ch);
 
           ++darktable.gui->reset;
 
-          dt_bauhaus_slider_set(g->cx, p->cx*100);
-          dt_bauhaus_slider_set(g->cy, p->cy*100);
-          dt_bauhaus_slider_set(g->cw, 100-p->cw*100);
-          dt_bauhaus_slider_set(g->ch, 100-p->ch*100);
+          dt_bauhaus_slider_set(g->cx, p_cx*100);
+          dt_bauhaus_slider_set(g->cy, p_cy*100);
+          dt_bauhaus_slider_set(g->cw, 100-p_cw*100);
+          dt_bauhaus_slider_set(g->ch, 100-p_ch*100);
 
           --darktable.gui->reset;
         }


TurboGit:
Working on my side, thanks !

dterrahe:
I don't think this has any effect given this code in commit_params:

  if(gui_has_focus(self))
  {
    d->cx = 0.0f;
    d->cy = 0.0f;
    d->cw = 1.0f;
    d->ch = 1.0f;
  }
  else
  {
    d->cx = p->cx;
    d->cy = p->cy;
    d->cw = fabsf(p->cw);
    d->ch = fabsf(p->ch);
  }

Are you sure there was a problem to solve in crop&rotate?

dterrahe:
BTW I'm not saying there are _no_ issues in crop&rotate; I see #4797 as well, but I'm not sure it's improved with this patch and my previous comment would indicate why.

ralfbrown:
@dterrahe by not updating p->cx etc in mouse_moved, the pixelpipe hash doesn't change anymore, so the current cached processing can be reused instead of calling all  of the downstream modules to reprocess the image (because the p->cx etc variables are part of the hash even though commit_params says to display the full image).  I was actually expecting to have to figure out some other place to update those variables in the *p* structure, but it just worked....

dterrahe:
So the full p is included in the hash? Not just d? Maybe that is something that should be reviewed then because that does not sound right to me (without _any_ understanding of pipes and hashes).

I tried to reproduce the improvement you saw in master vs this my branche, but I saw similar effects from #4797 in both cases (with "working" showing when moving and clicking inside). Maybe there's a more detailed reproduction (do I need to zoom in?) where there _is_ a significant difference?

I'm sorry if this sounds like I'm just being pedantic; I'm happy if you managed to get an improvement for a corner case, especially with such a small workaround, but since crop+rotate is already such a complicated mess, I wouldn't want anything to slip in that isn't genuinely useful as I'm sure nobody would ever be brave enough to remove it again.

So I'm just trying to "beat on this a bit", like you requested..

ralfbrown:
No special conditions.

Have you run with -d perf?  Even with the patch, the pixelpipe *does* get called on the initial button-down event (which causes the "working..." to be displayed), but the "full" pipe does nothing since the desired final result is cached, and the "preview" pipe only runs "display encoding" and "final histogram" (it's arguable that its doing so is a bug).  Without the patch, you get a full reprocessing from crop&rotate onward and periodic further calls while changing the crop box.

dterrahe:
I've now fixed the sliders in #5335 (and cleaned up the code a bit) to not expect xywh to be updated in params; they are behaving completely sensible again.

When the sliders are moved directly, the pixelpipe will still do a full run. This is useful/necessary (I think?) when the slider is moved via a shortcut while the module does not have focus (i.e. when the cropped picture is shown).

I've tried avoiding reprocessing when the module _is_ focused. Since the sliders are created by dt_bauhaus_slider_from_params, they automatically write their values to params. I've tried putting them back again, and not calling commit_box, but that doesn't seem to help. Maybe because the slider callback also calls dt_dev_add_history_item.

So the only solution then would be to drop dt_bauhaus_slider_from_params and write fully custom callbacks again. Not sure if that's worth it, since nobody seems to use the sliders anyway, except via shortcuts.

But also, I'd like to understand better why a change in params should invalidate the pixelpipe. That does not make immediate sense to me and could impact more modules. Was it maybe done out of an abundance of caution or because of some corner case elsewhere? I am fully ignorant here, so a "yes, this was discussed and is absolutely necessary" would be a satisfactory answer for me...

PR #5464, opened 2020-06-14
ashift: eliminate most pixelpipe runs when adjusting cropping

ralfbrown:
By keeping shadow copies of the current crop box in the dt_iop_ashift_gui_data_t structure instead of immediately updating the values in the dt_iop_ashift_params_t structure, we can avoid most pixelpipe runs while adjusting the crop box (the pixelpipe runs once when cropping is initially turned on and once for the first mouse move **after** a button release and subsequent re-press).

See also the discussion in #5441.

Unfortunately, the value committed when the iop loses focus is not the crop box at the time of the most recent button release, but of the button press which started the drag operation.  My best guess at the moment is that the undo system is behind this -- I update the crop into self->params in button_released(), but the next call to commit_params() or gui_update() has the old values in self->params as of the prior button_pressed()....  That would happen if the undo system is making a fresh copy of self->params and passing that to button_released, but then continuing to use the original elsewhere.


TurboGit:
@ralfbrown : Indeed a bit faster even if not as fast as with my fast-pipe PR.

I also see that the final crop is not the latest from the GUI.

TurboGit:
In  the mouse_moved() routine you want to update the shadow copies :+1: 

```
 if (g->adjust_crop)
  {
    g->cl = ...
   g->cr = ...
```

ralfbrown:
> In the mouse_moved() routine you want to update the shadow copies :
> 
> ```
>  if (g->adjust_crop)
>   {
>     g->cl = ...
>     g->cr = ...
> ```

The body of that *if* calls crop_adjust, which updates those variables, so adding them explicitly as suggested made no difference.  In looking at that if, however, it occurs to me that dt_dev_add_history_item probably needs to see the updated values in self->params, so I'll try temporarily updating p and then restoring it around that call.

> Indeed a bit faster even if not as fast as with my fast-pipe PR.

Yes, this is not a full solution because it still reruns the pixelpipe unnecessarily on  the initial button-down event (but you should be able to drag the mouse as fast as you want once that initial run completes).  And even if we come up with a change such that the pipe is never run while adjusting the crop, fastpipe mode is still highly desireable for other operations in ashift and clipping.

TurboGit:
Ok, working on my side now.

TurboGit:
And code looks good. I'll merge after #5335 is in as this is a quite big change and don't want to break more @dterrahe rebasing and conflicts resolution/

ralfbrown:
Discovered that the very first mouse_moved event the first time the crop box gets dragged after the module goes into focus caused the image to be redrawn without the crop box, yielding an annoying visual glitch when the next mouse_moved event results in a correct redraw.  That's been fixed in the most recent commit.

ralfbrown:
I believe I now have a full solution.  No more pixelpipe runs (other than the histogram on the preview pipe) when resizing the crop box, even after an intermediate button up.

ralfbrown:
And now no more histogram updates while dragging the crop box, either.  Just one when the mouse button is released and the new crop box is added to the undo history.

TurboGit:
@ralfbrown : Now that #5335 is merged there is conflicts. I was expecting that, sorry... Can you rebase and fix the conflicts ? Thanks.

dterrahe:
github doesn't show me the conflicts, so I'll go into git manually, after picking up the new master, and see if I can help with this.

dterrahe:
OK, as expected, this is mostly about widgets not having individual callbacks anymore after #5335 so some of the code you added in there needs to move to gui_changed (and can be deduplicated). Things like:
#ifdef ASHIFT_DEBUG
  model_probe(self, p, g->lastfit);
#endif
  do_crop(self, p);
  commit_crop_box(p,g);

I'm not sure how I could prepare a merge conflict resolution for you and upload to github. Maybe I can dump it somewhere else?

dterrahe:
This looks slightly tricky (in cropmode_callback):

  swap_shadow_crop_box(p,g);	//temporarily update real crop box
  swap_shadow_crop_box(p,g);	//restore p

The default introspection callback unconditionally executes dt_dev_add_history_item after returning from gui_changed. May need to either create a fully custom combobox for cropmode again or connect a second callback, that runs after the default one, to just call swap_shadow_crop_box

I have the same question as under #5463 though; should a changed params invalidate the pixelpipe in the first place? Because if that was really a band aid, then this is a band aid on a band aid. And really sorry again if I'm just repeating a stupid question.

dterrahe:
A merged version here https://github.com/dterrahe/darktable/tree/ashift-ppipe-merge

Seems to work smoothly for me with only a few stutters and without spinning up the fan. But you'll be a much better tester because you actually know what you are looking for.

TurboGit:
@dterrahe : Seems to work on my side. Fact is that this is so fast now that I don't think I'll merge po/fast-pipe now. Maybe you want to create a PR out of your merge work to ease testing by others?

dterrahe:
It may be better for @ralfbrown to pick up my changes and merge them in back here, after confirming he is happy with the behavior. So we don't end up with discussions in two places? I'm still happy to take responsibility for everything I messed up during the merge! ;-)

Also I'd still like to get clarity on whether changes to params immediately should invalidate the pipeline, because we now have two modules with ugly workarounds (my merge _is_ a hack) and there may be more that could benefit from a workaround. Which suggests maybe the design should be looked at? With too many workarounds all over the place, fixes only become harder.

There's also some overlap I guess with the "focused module functionality" discussion that I've only partially followed on IRC.

ralfbrown:
@dterrahe Since I'm new to dt code, I don't know why dt_dev_add_history_item forces a pipe invalidation and redisplay, though I do know that there are a lot of places where it's called that rely on at least the redisplay.  As for checking whether the merge is successfully implementing my changes, run from a terminal with -d perf.  After opening ashift, turning on "original format" crop and waiting for output to stop, when you click the mouse in the crop box and then drag it around, you should get **no** further output, and on release you should get just a couple of lines indicating "display color" and "final histogram" were run for the preview pipe.  Next click and drag should behave exactly the same.

@TurboGit It's still worth having the fastpipe mode for when you adjust the rotation/shift/skew sliders.

Both:  I'll take a look at the merged code tonight (early AM European time), but I'm still new enough to git that I need to ask: how do I get just the ashift-ppipe-merge branch without cloning dterrahe's entire fork?

TurboGit:
> Also I'd still like to get clarity on whether changes to params immediately should invalidate the pipeline,

This is certainly not a case where I'm expert. But first we need to decide what immediately means in this context :) For perspective correction and crop&rotate we have seen that it is indeed better to defer the pipe invalidation when the final cropping is done and validated.

But if you change the exposure slider (dragging it and not releasing it) I suppose you still want the pipe to be recomputed as this can have an effect on all blending masks and other module modify or based on the luminosity of each pixels. Like displaying the clipping pixels or reconstructing the highlight areas. So not sure what to do here.

That was the idea of my fast-pipe implementation. A module cropping does not care about the actual pixel values, so a fast pipe was a solution (workaround?).

At the end I have no clear idea to what should be done in which cases...

TurboGit:
> @TurboGit It's still worth having the fastpipe mode for when you adjust the rotation/shift/skew sliders.

Right ! So maybe it still be a good candidate for 3.2.

TurboGit:
@ralfbrown : to test...

Add a new remote for dterrahe clode:
 
`$ git remote add dterrahe https://github.com/dterrahe/darktable.git`

Pull current sources from main repository and create a local branch (say tmp) to test the change:

```
$ git checkout master
$ git pull
$ git branch -b tmp master
```

And finally merge dterrahe branch into your `tmp` branch:

`$ git merge dterrahe/ashift-ppipe-merge `

Then build and run as usual.

dterrahe:
> ... you still want the pipe to be recomputed as this can have an effect on all blending masks and other module modify or based on the luminosity of each pixels. ...

> That was the idea of my fast-pipe implementation. A module cropping does not care about the actual pixel values, so a fast pipe was a solution (workaround?).

Absolutely, but those changes are all passed to the pixelpipe itself in commit_params, so if the _data_t structure were hashed and a change detected, it should definitely invalidate the pipe.

But what we were seeing in clipping and perspective, I believe, is that already there _was_ a mechanism to not pass changes to commit_params until an action was completed. But even so, because the current status was stored in _params_t, the pipe gets invalidated. So that's why  @ralfbrown had to introduce _another_ layer of buffering to prevent _params_t from being affected "too early". And that doesn't sit well with #5335 which links widgets directly to params; hence needing hacks.

I don't understand where the hash is calculated (so if it could "easily" be changed from hashing params_t to data_t) and I also don't understand at all how your fastpipe mode works its magic, so I'm just the one asking the dumb questions here...

dterrahe:
@TurboGit wrote:
> @ralfbrown : to test...
> 
>      ...many lines of git wizardry...

In cases like this I sometimes fall back to good old copy/paste off the View File on github. Is that really bad for me to admit? Feels easier to double-check than making sure those git spells don't blow up the neighborhood (and I love git btw).

TurboGit:
> Feels easier to double-check than making sure those git spells don't blow up the neighborhood (and I love git btw).

So looks like I love Git more than you do :) :) I'm enjoying a lot the power of this tool.

ralfbrown:
AFAICT, the only times we want a deferred parameter update is when manipulating an overlay that doesn't actually change the underlying image being displayed.  Which right now is limited to the crop boxes in ashift/clipping, the keystoning markers in clipping, and toggling the alignment grid in ashift.  Whether a complete invalidation of all three pixelpipes is appropriate for the non-deferred cases is a separate question.

@dterrahe You're not the only one to use copy/paste for quick checks, but that doesn't work very well in this case because I'll need to pull in the new introspection framework.

IMO, the current implementation of dt_dev_add_history_item is a very poor design, because it does three unrelated actions, only one of which is implied by its name:

1. optionally enable the module if it isn't already (nearly all calls have this option set)

2. store the current parameters in the undo history

3. invalidate the pipe and request a redraw

I would suggest splitting off at least (3) by removing the dt_dev_invalidate_all and dt_control_queue_redraw_center calls from dt_dev_add_history_item and adding copies of those calls after every invocation of dt_dev_add_history_item.  Afterwards, we can then go through and figure out which of those invalidate/redraw calls are **actually** needed.  But that's potentially a huge amount of work since it will touch every iop for which the add_history hasn't already been moved entirely into the new introspection.

dterrahe:
> > Feels easier to double-check than making sure those git spells don't blow up the neighborhood (and I love git btw).
> 
> So looks like I love Git more than you do :) :) I'm enjoying a lot the power of this tool.

Me too!! But I wield it with considerable trepidation!

If it doesn't want to do what you ask of it, it tends to take revenge in a gruesome manner. As in; pieces everywhere.

dterrahe:
> @dterrahe You're not the only one to use copy/paste for quick checks, but that doesn't work very well in this case because I'll need to pull in the new introspection framework.

What I mean is; rebase on current master and "resolve" rebase conflicts by pasting the de-conflicted version on top of it. I'm not guaranteeing it will work...

ralfbrown:
@dterrahe OK, I have run your merged ashift through its paces.  With the exception of one visual glitch, everything works perfectly.  The glitch is clearly a missed initialization of the shadow crop:
  1. open ashift and ensure that "automatic cropping" is set to "off"
  2. put in enough rotation/shift/shear to produce substantial black wedges on the margins
  3. switch automatic cropping to "original format"
  4. the display will momentarily show the correct crop, then immediately redraw with the full viewport inside the crop.  The true crop **is** correctly stored, as it will be applied if you now close ashift, and the crop box will jump to the correct dimensions on the first mouse_moved event after button press.

The incorrect display happens any time you switch automatic cropping from "off" to "original format", but not when switching from "largest area" to "original format" (even though the latter two will typically produce somewhat different crops).

I'll see whether I can track down the root cause, but I may have to spend some time familiarizing myself with the new setup first.

ralfbrown:
@dterrahe In fiddling around more, I was able to generate the same visual glitch for every from/to change of the automatic cropping mode.  It looks to me like restore_shadow_crop_box_callback might be called without a prior call to gui_changed() when the user changes the setting for cropmode.  That will then get the variables for the real and shadow crop boxes out of sync, because swap_shadow_crop_box calls **must** be paired.

dterrahe:
Just to let you know that I'm working on this and that I know what the issues is:
when you select a combobox element by clicking on it, the callback gets called twice, once for button-down and once for button-up. The second call gets filtered out before gui_changed (because the value didn't change anymore), but restore_shadow_crop_box_callback receives both calls. You could try if it works correctly when you change the combobox with your mouse scroll-wheel because that seems to only send one message.

To solve this, I think I may need to set a flag in gui_changed to determine if swap_shadow_crop_box needs to be called in restore_shadow_crop_box_callback. This hack doesn't get any prettier...

dterrahe:
The version I just pushed to my branch doesn't show the glitch for me anymore.

I also completed the default values in introspection and removed the init function that is now no longer needed. Please check that reset parameters is working correctly for you.

dterrahe:
OK, the latest version I think is much less hacky or actually clean. I've moved the call to add_history to the dedicated cropmode callback and bracketed it with swap_shadow_crop_box calls there.

The "normal" handling of the combobox (picking up p->combobox from the widget) is still done in the default introspection handler and gui_changed is called, but nothing done there, and then add_history on the way out. I don't know if that is a problem? (since the real crop_box has not been swapped in?)

If you want to avoid the default handler and call to gui_changed completely, you can disconnect the handler and read the widget value all in cropmod_callback. I've done that in an intermediate version  8d7a68c729bcf00daaa8aafda6f2f645f0b763ec but I didn't notice any difference. It is a little more hacky in the sense that disconnecting depends on the internal implementation of combobox_from_params, which is why I added the error message in case that is ever changed and this breaks.

You'll want to squash my changes/just copy-paste the final version before submitting.

ralfbrown:
@dterrahe Running perfectly.  Ready to merge.

I seem to have made a bit of a mess of the commit history again....  It's actually based on master as of a few minutes ago, so everything but the last commit above should already be in master.  @TurboGit should definitely heed the suggestion to squash merge or copy/paste.

TurboGit:
Indeed, commits were a huge mess :) I have squashed all together and merged manually. Thanks for the hard work! Closing now.

PR #5467, opened 2020-06-14
introduce more responsible maintenance

johnny-bit:
in some cases maintenance/vacuuming/analzye might fail.
in some cases maintenance might just do nothing.

either way previously maintenance was "done deal", with this change all reporting is as honest as possible, making sure no case is missed.

there's also one tiny sqlite oddity that affects vacuum - in case where need to vacuum arised bevause of large move on db that was within this session, vacuum will be no-op. On next session if condition is still present, it'll work just fine.

this should cover all possible edge-cases for db maintenance and avoid any unnecessary bug reports.

there's no logic change and no option added, just a bit more error checking and logging so I hope this'll go into 3.2 :)

### DIFF
--- a/src/common/database.c
+++ b/src/common/database.c
@@ -2907,20 +2907,29 @@ void _dt_database_maintenance(const struct dt_database_t *db)
   {
     dt_print(DT_DEBUG_SQL, "[db maintenance] maintenance deemed unnecesary, performing only analyze.\n");
     sqlite3_exec(db->handle, "ANALYZE data", NULL, NULL, &err);
+    ERRCHECK
     sqlite3_exec(db->handle, "ANALYZE main", NULL, NULL, &err);
+    ERRCHECK
     sqlite3_exec(db->handle, "ANALYZE", NULL, NULL, &err);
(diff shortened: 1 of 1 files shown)


src/common/database.c
```diff
@@ -2875,13 +2875,71 @@ gboolean dt_database_get_lock_acquired(const dt_database_t *db)
...
+  {
+    val = sqlite3_column_int(stmt, 0);
+  }
+  sqlite3_finalize(stmt);
+  g_free(query);
+
+  return val;
+}
+
+#define ERRCHECK {if (err!=NULL) {dt_print(DT_DEBUG_SQL, "[db maintenance] maintenance error: '%s'\n",err); sqlite3_free(err); err=NULL;}}
```
TurboGit, line 2894: This is not used and should be removed.

johnny-bit: Instead those should be actually used :)

johnny-bit:
Done!

TurboGit:
Perfect ! Thanks.

Issue #5063, opened 2020-05-15
[FR] new pluggable input (not for drawing!) system for darktable


johnny-bit:
This sat on my inbox and back burner for far too long :/

On the legs of some comments left in #3722 it became apparent that current "input system" darktable has is a bit "hacky" and requires loads of boilerplate code + isn't very extensible as shown by extent of work needed for MIDI support.

After a bit of discussion there was another one on IRC to get the feel of what needs to be done in what way. It became apparent that the extent of work is rather large and will span months of work.

Currently darktable has: vimkeys, lua integration callbacks, key accels and dbus bindings, all of them being separate. Unification is needed in a pluggable and generic way. Suggested approach to coming up with solution was to sorta follow vimkeys approach of registering every control and then having access to it exposed for all known inputs, however multi-modules etc come a bit in way and do mess with vimkeys as of now. the job for input system would be to handle registering and unregistering of "action targets" (eg sliders, combos etc) and "input modules" (think keyboard shortcuts, midi, vimkeys etc) and handling communications between them.

From modules & control side, input system must handle:
1. multiple instances
2. getting & setting of all possible values/controlls
3. shortcuts discoverability (eg tooltips with shortcut info are easy example)
3. "same" shortcut for separate dt mode (eg tethering and lighttable are guaranteed to never receive same input, so could technically use same shortcuts)

From input modules side, input system must handle:
1. plugging and unplugging of input module
2. conflicts between input modules (eg vimkeys vs keyboard)
3. back communication to device (in MIDI world there are sliders that can be moved mechanically, so if user moves slider using mouse, an attached slider on midi device should move by itself)

From iop/contributor's side input system must guarantee:
1. less boilerplate code (no more signal connections needed to get benefits of shortcuts, so any known slider, gets to have shortcuts, no need for any additional coding)
2. respect set limits (prefferably soft limits), so that input system is guaranteed to never let slider go "up to eleven" if soft-limit is 10. hard limits are to be inputed manually via bauhaus input in right click over slider
3. be completely transparent code-wise from IOP standpoint (similarly to IOP introspection)

In order to create generic input system couple things must be kept in mind:

1. Various inputs, which can be different than what one would expect, so possible "input module" may not seem like input at all, eg:
- Keyboard
- Mouse (with extra keys)
- Pressure sensitive touchpad
- Remote
- Game controller (joysticks, pads, etc)
- MIDI Device (those are VERY varied)
- Pedals
- Lua subsystem
- vimkeys
[...]

PR #5468, opened 2020-06-14
edited fields focus fix

### LINKED ISSUE #5460: Issue on entering metadata or tag info.
**Describe the bug**
Mouse hovering on images makes lose the focus on metadata and tag entries. The result is always unexpected when one is writing ...

**To Reproduce**
Select an entry in metadata, start writing something.
If the mouse moves and hovers an image the written text disappears and is lost.
Worse, if the mouse move happens unexpectedly and one continues to write, dt executes the action corresponding to the entered shortcuts (instead of writing the text).

**Expected behavior**
As long as another image (EDIT or another field) is not selected (clicked on), there is no reason for losing the current text, and to have the risk to break something.

**Additional context**
darktable 3.1.0+2137~g3b19bdfa2
OS debian 10

phweyland:
edited fields focus fix
- don't grab focus on hovering thumbnails
- ~~metadata clear entry when focus is lost~~
- metadata remove redraw widget callback (cairo not involved here)
- ~~tag clear entry when focus is lost~~
- treat tag entry escape

Fixes #5460 

Pinging @AlicVB for potential side effect on thumbtable. On my side I haven't seen any, but your vision on this is surely better than mine.

### DIFF
--- a/src/libs/metadata.c
+++ b/src/libs/metadata.c
@@ -39,6 +39,7 @@ typedef struct dt_lib_metadata_t
 {
   int imgsel;
   GtkTextView *textview[DT_METADATA_NUMBER];
+  gulong lost_focus_handler[DT_METADATA_NUMBER];
   GtkWidget *swindow[DT_METADATA_NUMBER];
   GList *metadata_list[DT_METADATA_NUMBER];
   GtkGrid *metadata_grid;
@@ -137,15 +138,10 @@ static void _fill_text_view(const uint32_t i, const uint32_t count, dt_lib_modul
   _text_set_italic(d->textview[i], multi);
 }
 
-static void _update(dt_lib_module_t *self, gboolean early_bark_out)
+static void _update(dt_lib_module_t *self)
 {
-  //   early_bark_out = FALSE; // FIXME: when barking out early we don't update on ctrl-a/ctrl-shift-a. but
-  //   otherwise it's impossible to edit text
   dt_lib_metadata_t *d = (dt_lib_metadata_t *)self->data;
-  const int imgsel = dt_control_get_mouse_over_id();
-  if(early_bark_out && imgsel == d->imgsel) return;
-
-  d->imgsel = imgsel;
+  d->imgsel = dt_control_get_mouse_over_id();
 
   GList *metadata[DT_METADATA_NUMBER];
   uint32_t metadata_count[DT_METADATA_NUMBER];
@@ -214,35 +210,30 @@ static void _update(dt_lib_module_t *self, gboolean early_bark_out)
 static gboolean _postponed_update(gpointer data)
 {
   // timeout handle clearing is handled by update code
-  _update((dt_lib_module_t *)data, FALSE);
+  _update((dt_lib_module_t *)data);
 
   return FALSE;
 }
 
 static void _image_selection_changed_callback(gpointer instance, dt_lib_module_t *self)
 {
-  _update(self, FALSE);
+  _update(self);
 }
 
 static void _collection_updated_callback(gpointer instance, dt_collection_change_t query_change, gpointer imgs,
                                         int next, dt_lib_module_t *self)
 {
-  _update(self, FALSE);
-}
-
-static gboolean _draw(GtkWidget *widget, cairo_t *cr, dt_lib_module_t *self)
-{
-  if(!dt_control_running()) return FALSE;
-  _update(self, TRUE);
-  return FALSE;
+  _update(self);
 }
 
 static void _clear_button_clicked(GtkButton *button, dt_lib_module_t *self)
 {
+  dt_lib_metadata_t *d = (dt_lib_metadata_t *)self->data;
+  d->editing = FALSE;
   GList *imgs = dt_view_get_images_to_act_on(FALSE, TRUE);
   dt_metadata_clear(imgs, TRUE);
   dt_image_synch_xmps(imgs);
-  _update(self, FALSE);
+  _update(self);
 }
 
 static void _append_kv(GList **l, const gchar *key, const gchar *value)
@@ -254,7 +245,6 @@ static void _append_kv(GList **l, const gchar *key, const gchar *value)
 static void _write_metadata(dt_lib_module_t *self)
 {
   dt_lib_metadata_t *d = (dt_lib_metadata_t *)self->data;
-
   d->editing = FALSE;
 
   gchar *metadata[DT_METADATA_NUMBER];
@@ -280,7 +270,7 @@ static void _write_metadata(dt_lib_module_t *self)
   dt_control_signal_raise(darktable.signals, DT_SIGNAL_METADATA_CHANGED, DT_METADATA_SIGNAL_NEW_VALUE);
 
   dt_image_synch_xmps(imgs);
-  _update(self, FALSE);
+  _update(self);
 }
 
 static void _apply_button_clicked(GtkButton *button, dt_lib_module_t *self)
@@ -315,7 +305,7 @@ static gboolean _key_pressed(GtkWidget *textview, GdkEventKey *event, dt_lib_mod
         event->keyval = GDK_KEY_Tab;
         break;
       case GDK_KEY_Escape:
-        _update(self, FALSE);
+        _update(self);
         gtk_window_set_focus(GTK_WINDOW(dt_ui_main_window(darktable.gui->ui)), NULL);
         d->editing = FALSE;
         break;
@@ -347,9 +337,16 @@ static gboolean _got_focus(GtkWidget *textview, dt_lib_module_t *self)
   return FALSE;
 }
 
+static gboolean _lost_focus(GtkWidget *textview, GdkEventFocus *event, dt_lib_module_t *self)
+{
+  dt_lib_metadata_t *d = (dt_lib_metadata_t *)self->data;
+  d->editing = FALSE;
+  return FALSE;
+}
+
 void gui_reset(dt_lib_module_t *self)
 {
-  _update(self, FALSE);
+  _update(self);
 }
 
 int position()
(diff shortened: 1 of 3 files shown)


AlicVB:
I've not found any drawback either. I admit that I don't exactly remember why the focus in the inside view was required. It was for accels, but they seems to work well like that...

So that seems fine for me !

phweyland:
> So that seems fine for me !

Thanks a lot for checking this.

TurboGit:
IIRC @elstoc changed for focus request to fix key shortcut in some cases. So I'd like his view on this change if possible before merging. @elstoc is that ok with you ?

elstoc:
I haven't tested but I don't think this should be a big issue. The only fixes I have added were:

- give focus to centre screen when clicking on an iop module header (so darkroom only) in order to take focus away from the search box
- pass keyboard shortcuts to key_pressed view handler if not already handled by a bauhaus widget (slider/combobox) that currently has keyboard focus

This change will have the impact that it stops the lighttable keyboard shortcuts (moving image, viewing the preview) from working (unless you click on a thumbnail in the centre view to reassign focus), but that seems reasonable if you've chosen to edit a metadata field. Gtk entry fields seem to stubbornly keep keyboard focus unless you explicitly assign it to something else.

Perhaps we could add something to this so that when you expand/collapse lib views this also delivers focus back to the centre view (dt_ui_center(darktable.gui->uk)) - similar to my first fix above.

Longer term, we really should get rid of the key_pressed handlers in the views (at least for the preview window) and replace them with standard keyboard shortcuts, since they currently require a few hacks to ensure they keep working reliably.

src/libs/metadata.c
```diff
@@ -214,35 +210,30 @@ static void _update(dt_lib_module_t *self, gboolean early_bark_out)
...
 
 static void _collection_updated_callback(gpointer instance, dt_collection_change_t query_change, gpointer imgs,
                                         int next, dt_lib_module_t *self)
 {
-  _update(self, FALSE);
-}
-
-static gboolean _draw(GtkWidget *widget, cairo_t *cr, dt_lib_module_t *self)
-{
-  if(!dt_control_running()) return FALSE;
```
johnny-bit, line 235: this still should IMO have this bit. it's basically saying "if darktable is closing, do nothing on this event"

johnny-bit:
For some of the stuff @elstoc's saying #5063 would be optimal "solution"...

I've tested this PR and it seems like it doesn't affect anything negatively. so it's all good for me except that small dt_running conditional.

src/libs/metadata.c
phweyland: Good to do nothing if exiting dt !
But _draw() is called thousands of times and I haven't still figured out what it does in metadata on normal time...
The "draw" event is related to cairo stuff. In metadata module I don't see anything related to that. It's possible to have had some role in the past. But currently I don't see what it does. Do you have any idea ?

phweyland:
>  but that seems reasonable if you've chosen to edit a metadata field. Gtk entry fields seem to stubbornly keep keyboard focus unless you explicitly assign it to something else.

To edit any field grab focus is required otherwise shortcut are captured and one cannot even write "bloody hell". So when a field is edited it seems normal that shortcuts are disabled (except the edition ones). But the focus is released all shortcuts should work properly.

Hoovering is handy to get on the fly information (about image, tooltips, ...) but should NEVER steal the focus (even in darkroom).

> This change will have the impact that it stops the lighttable keyboard shortcuts (moving image, viewing the preview) from working

Could you pinpoint the code which could fail without having the focus (I don't know these features moving image, viewing the preview ...) ?
I'll check they still work.
dt_view_get_image_to_act_on() is always aware of the hoovered image even while editing (to check that start edit a metadata field (with that PR), hover other images, you can see image information and tagging updated correctly). 
When focus is released, shortcuts are enabled again and the cases you mention should still work.

src/libs/metadata.c
johnny-bit: Since it simply called _update with early `bark_out` (lol) then it's most likelly that it was just overzealous update mechanism. i _think_ for metadata it would be enough to just use collection/selection/mouse_over signals and it would be "enough".

if i remember correctly dt_signal_raise doesn't even raise signals if closing darktable so it avoids unnecessary handling.

phweyland:
> (dt_ui_center(darktable.gui->uk))

For my understanding, do you know what is the difference with (dt_ui_main_window(darktable.gui->ui)), which is commonly used inside modules, in term of focus behavior ?

phweyland:
> For some of the stuff @elstoc's saying #5063 would be optimal "solution"...

Ooops, this one is a big stuff !
However, the discussion here is only about focus, and I'm sure that with any other input system, focus still must be handled properly.

src/libs/metadata.c
phweyland: I agree with what was executed :) but I just don't figure out why the "draw" event was triggered ...

elstoc:
> Could you pinpoint the code which could fail without having the focus

In my experience if any other widget has keyboard focus, the key_pressed function in the darkroom/lighttable view code does not get run. But with a gtkentry widget I think that grabs all of the keyboard input anyway. There was an issue with the module search in the darkroom that, because it never gave up keyboard focus, none of the keyboard shortcuts worked unless you first clicked on the centre image (hence my fix was to give keyboard focus to the centre image). I agree that hovering should never grab the keyboard focus, my issue was that something at some point needs to make the metadata module _lose_ focus in order for the shortcuts to start working again. If you've done this then I'm sure it will work fine, or clicking in the lighttable centre panel will also work.

> For my understanding, do you know what is the difference with (dt_ui_main_window(darktable.gui->ui)), which is commonly used inside modules, in term of focus behavior ?

I assume that the above gives focus to the entire window, which then probably passes that focus down to the first one of its children that can accept keyboard focus, whereas gtk_widget_grab_focus(dt_ui_center(darktable.gui->uk)) specifically hands the focus to the centre panel. I think the gtk documentation suggests its preferable to assign focus to a specific widget as it won't be entirely predictable where the focus will get assigned if you give it to the whole window.

phweyland:
>  There was an issue with the module search in the darkroom that, because it never gave up keyboard focus, none of the keyboard shortcuts worked unless you first clicked on the centre image

I see, that is what one has always done, not very handy, agreed. But that's the role of the escape key, isn't it ? While escape key was already implemented on metadata, I've just added it on tagging where it was missing. Maybe it is also missing in this darkroom search field. I think that escape key is handier than clicking anywhere on the centre image.

> gtk_widget_grab_focus(dt_ui_center(darktable.gui->uk)) specifically hands the focus to the centre panel. I think the gtk documentation suggests its preferable to assign focus to a specific widget as it won't be entirely predictable where the focus will get assigned if you give it to the whole window

It's important to note that `grab focus` is more than `set focus` to a widget. It gives the ability to that widget to get all the entered keys (as we need for editing).

elstoc:
> Maybe it is also missing in this darkroom search field

I think the difference is that with the darkroom search field, once your search is successful the next thing you will do is click on a module header or adjust a bauhaus widget, which is how you indicate you're done with the search box. Hard to see why you'd want to search and then not follow it up with another module action (for which the tab key is fine, if you know to use it).

phweyland:
> > Maybe it is also missing in this darkroom search field
> 
> I think the difference is that with the darkroom search field, once your search is successful the next thing you will do is click on a module header or adjust a bauhaus widget, which is how you indicate you're done with the search box. Hard to see why you'd want to search and then not follow it up with another module action (for which the tab key is fine, if you know to use it).

... when I search a module, I end up clicking on the one I want. This makes the search field lose the focus, before adjusting whatever slider, right ? Then shortcuts should work properly for the slider. Is that still correct ? (Note I do not see any slider before selecting and opening a module).

TAB does make the search field lose the focus, but not escape key. This one is really missing.

src/libs/metadata.c
johnny-bit: the _draw is triggered every time widget for full iop has to be redrawn (so quite often). you could try and comment out
```c
g_signal_connect(self->widget, "draw", G_CALLBACK(_draw), self);
```
in `gui_init` and thus avoid unnecessary calls. then see what (if anything) changes, making sure that the other signals I've mentioned are handled.

elstoc:
> ... when I search a module, I end up clicking on the one I want. This makes the search field lose the focus, before adjusting whatever slider, right ? Then shortcuts should work properly for the slider. Is that still correct ?

So yes it does that, but it didn't used to if you just expanded the module (to look at the parameters maybe) without adjusting a bauhaus widget. It works like that now because I changed so that when you click on the module header the callback function causes the centre panel to grab focus. Prior to that fix the search box retained focus until you clicked or adjusted a bauhaus widget. 

> (Note I do not see any slider before selecting and opening a module)

You might see a slider without opening a module if that module was already open (in another module group).

I agree with you that escape to clear the box and lose focus would be useful.

src/libs/metadata.c
phweyland: >  you could try and comment out

This is already done, isn't it ? :)

phweyland:
>  It works like that now because I changed so that when you click on the module header the callback function causes the centre panel to grab focus.

Why not to give the focus the module on which the user has clicked on ? (that would remove the focus from search field as well). But if that works, that's already good.

I think we are all aligned, aren't we ?

elstoc:
> Why not to give the focus the module on which the user has clicked on ?

That was my first attempt but because the header can't do anything with the keyboard focus it handed it down to the first available (random) bauhaus widget meaning that scroll then changed random sliders/comboboxes depending on the module.

> I think we are all aligned, aren't we ?

Yes we got off on a tangent there didn't we. I'm fine with this change. We can always tweak later if there are any minor issues.

phweyland:
Pascal, is ready for your review when you can.

TurboGit:
I was not sure this was ready for review and was sidetracked by many other things. I should be able to come back to this tomorrow.

TurboGit:
I found a quite big issue.

- select a picture
- enter "toto" into title metadata field
- click on [apply]

The metadata is lost from the field and not applied to the picture.

phweyland:
>     * click on [apply]
> 
> 
> The metadata is lost from the field and not applied to the picture.

Arghhh! Quite possible. I look at it.

phweyland:
I've removed the field clearing on lost of focus. Should work better now. Sorry for that.

I'm not completely satisfied with ergonomics but that's anterior to this fix and can be done in another PR: there is no visual difference between fields already saved and fields newly entered and not yet saved (those which will be cleared if one presses escape).

PR #5469, opened 2020-06-14
expose DPI and GUI scaling factors in prefs

aurelienpierre:
try to redraw the whole GUI when pref window is deleted, but only some widgets are redrawn.

Capture d’écran du 2020-06-14 18-17-06

### DIFF
--- a/src/gui/preferences.c
+++ b/src/gui/preferences.c
@@ -379,7 +379,7 @@ static void init_tab_general(GtkWidget *stack)
 
 
   //font size selector
-  if(dt_conf_get_bool("use_system_font") < 5.0f || dt_conf_get_float("font_size") > 20.0f)
+  if(dt_conf_get_float("font_size") < 5.0f || dt_conf_get_float("font_size") > 20.0f)
     dt_conf_set_float("font_size", 12.0f);
 
   label = gtk_label_new(_("font size in points"));
--- a/src/gui/preferences.c
+++ b/src/gui/preferences.c
@@ -203,6 +203,20 @@ static void font_size_changed_callback(GtkWidget *widget, gpointer user_data)
   dt_bauhaus_load_theme();
 }
 
+static void gui_scaling_changed_callback(GtkWidget *widget, gpointer user_data)
+{
+  dt_conf_set_float("screen_ppd_overwrite", gtk_spin_button_get_value(GTK_SPIN_BUTTON(widget)));
+  dt_configure_ppd_dpi(darktable.gui);
+  dt_bauhaus_load_theme();
+}
+
+static void dpi_scaling_changed_callback(GtkWidget *widget, gpointer user_data)
+{
+  dt_conf_set_float("screen_dpi_overwrite", gtk_spin_button_get_value(GTK_SPIN_BUTTON(widget)));
+  dt_configure_ppd_dpi(darktable.gui);
+  dt_bauhaus_load_theme();
+}
+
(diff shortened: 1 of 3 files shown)


src/gui/preferences.c
```diff
@@ -379,6 +393,33 @@ static void init_tab_general(GtkWidget *stack)
   gtk_spin_button_set_value(GTK_SPIN_BUTTON(fontsize), dt_conf_get_float("font_size"));
   g_signal_connect(G_OBJECT(fontsize), "value_changed", G_CALLBACK(font_size_changed_callback), 0);
 
+  GtkWidget *screen_ppd_overwrite = gtk_spin_button_new_with_range(-1.0f, 8.0f, 0.2f);
+  label = gtk_label_new(_("images DPI scaling factor"));
```
TurboGit, line 397: images -> thumbs

Which will be less ambiguous I think.

TurboGit:
@aurelienpierre : are you targeting this for 3.2 ?

aurelienpierre:
> @aurelienpierre : are you targeting this for 3.2 ?

Yes, it's long due and has been requested many many times.

src/gui/preferences.c
aurelienpierre: done, I have also reworked the other label to make the difference more obvious

TurboGit:
Hum... I set GUI DPI to 96 then GUI scaling factor to 1.2, leaving the entry the 1.2 was changed in 8.0. Exited the preferences and dt was not responding, my computer started to set ventilator very fast I had to kill dt but it was quite difficult as it was eating most of my CPU ! 

I would say not ready for integration :)

parafin:
I wonder if non-integer PPD (1.2) can cause troubles, or does it make sense at all.

coffeeandspaceships:
Below is from `darktable.c`. ppd is used in a few other places where float seems ok; but  I noticed this one carries out a multiply without a cast to int. That will round 1.2 down to 1.0 again if the compiler (and flags applied to compiler...) support implicit conversation, I think? 
```    // blow up the PNG. Ugly, but at least it has the correct size afterwards :-/
    const int width = DT_PIXEL_APPLY_DPI(png_width) * darktable.gui->ppd,
              height = DT_PIXEL_APPLY_DPI(png_height) * darktable.gui->ppd;
    const int stride = cairo_format_stride_for_width(CAIRO_FORMAT_ARGB32, width);
```

parafin:
Your analysis is incorrect:

- first of all there is no "if the compiler supports" - all compilers follow the language standard and behave the same, at least in this regard
- DT_PIXEL_APPLY_DPI multiplies int by double, result value is double (int is converted to double)
- conversion back to int happens after calculation is done (multiplication of double by double)

aurelienpierre:
> Hum... I set GUI DPI to 96 then GUI scaling factor to 1.2, leaving the entry the 1.2 was changed in 8.0. Exited the preferences and dt was not responding, my computer started to set ventilator very fast I had to kill dt but it was quite difficult as it was eating most of my CPU !

I tried to be clever and have new sizes loaded and GUI redrawn when exiting the pref window. First of all, not everything is redrawn(although redrawing main window should redraw all children too…), but it could be the source of the problems. 

Maybe just ask user to restart ?

TurboGit:
Ok, does not crash my DE now :) Which is good... Merging!

TurboGit:
Thanks !

PR #5470, opened 2020-06-14
thumbs : ensure overlays sizes after change

### LINKED ISSUE #5461: Switching from block infos to other overlays add a not wanted glitch
**Describe the bug**
Some not wanted (not CSS on what I check) margin is added when switching form overlay_block_infos to other overlays modes

**To Reproduce**
1. Go to 'popover overlay menu'
2. Click on 'overlays block on mouse hover'
3. See that this mode is shown correctly
4. Go back to 'popover overlay menu' and select another overlays (except permanent overlays where no issue shown).

See the issue:

@AlicVB: as a simple darktable restart set back those overlays modes to their wanted display, it's probably more a refresh issue than CSS issue. Probably since your last work on margins. It was good to switch without restart after my last CSS work on those modes.

AlicVB:
this fix #5461


Nilvus:
Tested and just all good, issue fixed. Thanks for that quick fix.

PR #5477, opened 2020-06-15
forget margin-top offset as it doesn't works if using WM without compositing

Nilvus:
Related to #5438 and revert a part back as it doesn't work if no compositing. That just remove margin-top and so transparent background issue. @elstoc and @parafin: as you have way to quickly test a WM without compositing, could you please test?
It's always better to test if possible before than after merging. Thanks.

### DIFF
--- a/data/themes/darktable.css
+++ b/data/themes/darktable.css
@@ -811,12 +811,13 @@ context-menu,
 menu,
 menuitem > menu, /* sub-menu */
 menuitem arrow,
-tooltip label,
+tooltip,
 popover,
 #background_job_eventbox, /* background of export progress bar */
 combobox window,
 dialog combobox window
 {
+  opacity: 1;  /* this is needed for tooltip on/off feature with Shift+t shortcut */
   background-color: @tooltip_bg_color;
   border-radius: 4px;
   border: 0;
@@ -828,21 +829,12 @@ dialog combobox window
 
 tooltip
 {
-  opacity: 1;  /* this is needed for tooltip on/off feature with Shift+t shortcut */
-}
-
-tooltip *,
-tooltip.background
-{
-  background-color: transparent;
+  border: 1px solid shade(@tooltip_bg_color, 1.5);
 }
 
 tooltip label
 {
-  margin-top: 3px;
-  padding: 0.4em;
   color: @tooltip_fg_color;
-  border: 1px solid shade(@tooltip_bg_color, 1.5);
 }
 
 combobox label
--- a/data/themes/darktable.css
+++ b/data/themes/darktable.css
@@ -830,6 +830,7 @@ dialog combobox window
 tooltip
 {
   border: 1px solid shade(@tooltip_bg_color, 1.5);
+  border-radius: 0;
 }
 
 tooltip label


elstoc:
Looks fine. The only thing you lose without compositing enabled is the rounded corners, which get given a black background (barely noticeable in the screenshot below). 

Screenshot_2020-06-15_13-13-04

But this is probably an issue elsewhere as well, and not worth spending time on for those few that have compositing off. To be honest without compositing there are worse oddities with other gtk applications (for example the polari IRC client has a massive black bar around the window).

Nilvus:
@elstoc: thanks for your quick test. It's better to see that. Not having round border only when not have compositing is really insignificant. Even if there's probably few people using compositing off, if we could have something good seeing for them, it's better.

So, @TurboGit: good to merge.

parafin:
Where else is this an issue in dt? As far as I understand tooltips are separate windows, while all other theming effects in dt deal with rendering in one single window. I think this is still ugly. If you want to have some effect just for people with compositing (I don't argue that it's a majority), add a detection for it like it was done in #5462.

elstoc:
Ok you're right I didn't do my research. Maybe it's not an issue elsewhere.

In terms of detecting compositing I suppose we could have a separate css file that contains compositing-specific effects but it would require gtk code to identify and load it.

parafin:
Does CSS in GTK support something dynamic like calc()? So that C code would define some variable/property that can be checked later in CSS theme.

elstoc:
Not that I'm aware of, but I'm far from an expert.

Nilvus:
I really don't understand the issue here. Rounded corners is not a new thing, nor on that PR, nor on previous one. Border-radius is there since a long time and nobody complains about that. So, if it's ugly now, it was before. Anyway, you could remove the border-radius line on your side and you will see that nothing has changed, as here the line is not even applied. So having it or not for WM without compositing, that will be the same. And we have border-radius on all buttons and combobox and lot of other things (toast message, log message, etc...). It's a minor effect and good for users with compositing...

parafin:
Rounded corners require transparency, which does work for everything except tooltips (see black corners on the screenshot). I've explained why this is happening.

Nilvus:
Thanks, didn't notice black corners. I understand why this is happening. It's just that border-radius is not related to that PR. It's there since months.

Nilvus:
As @elstoc pointed that XFCE allow to use compositing or not, I have just tested that on XFCE without compositing. Having those really little black corners doesn't annoy me and I don't find that ugly (even if it's quite less beautiful than with rounded corners). Tested also without border-radius and it's also good for me.

So I can remove it for tooltip only. Would like to have other feedback on that. @TurboGit or @elstoc: what do you think on that?

parafin:
Right, indeed corners weren't part of these 2 PRs, didn't notice that. I suggest removing them for now and to think about adding compositing-only enhancements the correct way later. This whole situation with GTK tooltips is quite sad, so don't get me wrong - I'm all for improving it, even if not for everybody.

elstoc:
@Nilvus I'm not bothered either way, though rounded corners on tooltips is probably not an important enough feature that we really need to keep it. You're right that this has been present for a while - probably nobody thought to test it on a non-compositing window manager and if anyone noticed it they didn't care enough to raise an issue.

Nilvus:
> Right, indeed corners weren't part of these 2 PRs, didn't notice that. I suggest removing them for now and to think about adding compositing-only enhancements the correct way later. This whole situation with GTK tooltips is quite sad, so don't get me wrong - I'm all for improving it, even if not for everybody (but let's not degrade the experience for others).

Nobody want to degrade the experience for others and we all want same thing: improving it. Sometimes it's just difficult to not degrade experience for some users by improving it for others. Anyway, on tooltip, border-radius doesn't really make difference when compositing is on, if we have it or not. So, I will push a commit to remove it that for tooltip only.

On what I see (quickly) on XFCE without compositing, I don't think we really need to add compositing-only enhancements as all we have actually seems to be displayed correctly. And of course, it's not necessary to add code for some minor thing like rounded corners in tooltip. We can do without them (and I now understand why I see only square corners on tooltips in many Gtk based apps I use).

parafin:
I was having in mind the ability to offset position of the tooltip, which is removed with this PR. But to be honest having tooltip far from mouse pointer looks strange to me. So indeed less code is better;)

aurelienpierre:
I think you could leave tooltips styling to the system, and use whatever properties defined by the Gtk theme. It will be the best tradeoff between ensuring universal support and degrading overall UI/appearance.

elstoc:
> I think you could leave tooltips styling to the system, and use whatever properties defined by the Gtk theme

I basically agree with the proviso that the opacity needs to be defined in darktable.css (as 1) in order for the tooltip toggle to work.

Nilvus:
I remove border-radius for tooltip. I tested that with and without compositing in XFCE and as usual on Gnome (so with compositing) and it's all good for tooltip. So I think now we are good and we could safely merge that.

@aurelienpierre and @elstoc: you lost me there as I'm not sure you mean changing (or not) something here. I assume it's just a comment as it works without any transparency set (nor border-radius nor background transparency now).

I didn't thought that helping just to improve tooltips (which were on start good, except border-radius for WM without compositing), I will go in such can of worms ;-) Anyway, I learn something about compositing or not.

For me, this could be merged now.

aurelienpierre:
What I meant is remove internal darktable style for tootlip (except font-related stuff) completely, so the default Gtk style will be used, and we can bet it will properly deal with compositing and such.

Nilvus:
> What I meant is remove internal darktable style for tootlip (except font-related stuff) completely, so the default Gtk style will be used, and we can bet it will properly deal with compositing and such.

Ok, but it deal properly now with compositing and such, so no need to remove what is remaining.

TurboGit:
@Nilvus : Is that ready for testing / merging ?

Nilvus:
> @Nilvus : Is that ready for testing / merging ?

Yes, I'm waiting for that!

TurboGit:
Ok, working on my side. Thanks.

PR #5480, opened 2020-06-15
variable $(SEQUENCEn), n added to control the number of digits

phweyland:
Small variables enhancement.
variable $(SEQUENCEn), optional 'n' added to control the number of digits.
The default value (4) is kept.

EDIT The $(SEQUENCE) variable can be used in conjunction with custom order tag collection. Then the order of exported images can be kept inside the exported filenames. But 4, if conservative, is a bit too much for most of the collections.

### DIFF
--- a/doc/usermanual/lighttable/panels/export.xml
+++ b/doc/usermanual/lighttable/panels/export.xml
@@ -141,10 +141,10 @@
             </row>
             <row>
               <entry>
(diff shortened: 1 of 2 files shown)


johnny-bit:
Very good idea👍

TurboGit:
Quite safe, let's add this to 3.2. thanks!

Issue #5435, opened 2020-06-12
Thumbnail overlays → Hidden after dt restart


timur-davletshin:
1. Linux, 64 bit, latest git
2. Click "click to change type of overlays shown on thumbnails" (star icon in right-top corner) and choose, let's say, "permanent overlays" — overlays shown as expected.
3. Restart dt, notice that overlays are no longer visible. Selected menu option remains active.

AlicVB:
I can't reproduce of course, otherwise this feature would be useless...
Can you share more things to reproduce the issue ? In which mode are you in lighttable (filemanager, zoomable lighttable, culling, full preview) ? Exact version of dt, ...
Thanks

timur-davletshin:
file manager mode

timur-davletshin:
AlicVB, I tried with new config and I find that having header like mine (see below) is crucial (click that triangle till you get header with hidden buttons)
![header](https://user-images.githubusercontent.com/65483426/84530118-44377a80-aceb-11ea-95b8-9cd36b20324d.png)

AlicVB:
ah ah indeed, that was a crucial info.
I can reproduce now. I'll have a look

AlicVB:
@Mark-64 : I've found the culprit, it's from a part of your code... and I have no idea how to solve it : 
in fact, if we don't show the top toolbar (with the buttons), `darktable.gui->icon_size = 0` and as we rely on this to get the font and icon size for thumbnails...

That's due to the fact that the icon_size is computed once toolbar buttons are realized, which is never the case if the toolbar is hidden...

As potentially, we can start dt with every panels closed, I don't see how we can get this value...

Can you have a look ? Thanks

Nilvus:
@timur-davletshin: could you try on last master from this day to see if you can reproduce?

Nilvus:
@Mark-64: could you check that (maybe you miss @AlicVB question)? Your help would be appreciated here. Thanks.

timur-davletshin:
@Nilvus still reproducible in darktable 3.1.0~git2164.a64940338

Nilvus:
> @Nilvus still reproducible in darktable 3.1.0~git2164.a64940338

Yes, I delete my comment finally (I have not understand your PR at first as I don't reproduce it because understood that needed header bar and collapse border arrows). In fact, that's not that, it's reproducible with only header shown but header toolbar (with prefs, help, group icons...) is hidden. And then I have reproduce it.

[...]

PR #5481, opened 2020-06-15
Thumbtable icons size

Mark-64:
This is an attempt to solve #5435 
However, it doesn't work when starting DT directly in culling mode.
In that case it works only forcing a redraw, although the callback setting the size is executed.
I do not really understand how culling mode works, I need @AlicVB help here.
The other thumbtable modes seem to work, but some more tests are needed

### DIFF
--- a/data/themes/darktable.css
+++ b/data/themes/darktable.css
@@ -1841,6 +1841,11 @@ Details :
   background-color: @lighttable_bg_color;
 }
 
+#thumbtable_filemanager
+{
+min-height: 2em;
+}
+
 #thumb_back
 {
   background-color: @thumbnail_bg_color;
--- a/data/themes/darktable.css
+++ b/data/themes/darktable.css
@@ -1841,9 +1841,13 @@ Details :
   background-color: @lighttable_bg_color;
 }
 
-#thumbtable_filemanager
+/* here we set the maximum size of GUI elements in thumbtable view to be in synch with the rest of the GUI */
+#thumbtable_filemanager,
+#thumbtable_filmstrip,
+#thumbtable_zoom,
+#culling
 {
-min-height: 2em;
+min-height: 1.47em;
 }
 
 #thumb_back
--- a/src/dtgtk/culling.c
+++ b/src/dtgtk/culling.c
@@ -825,6 +825,8 @@ dt_culling_t *dt_culling_new(dt_culling_mode_t mode)
   g_signal_connect(G_OBJECT(table->widget), "button-press-event", G_CALLBACK(_event_button_press), table);
   g_signal_connect(G_OBJECT(table->widget), "motion-notify-event", G_CALLBACK(_event_motion_notify), table);
   g_signal_connect(G_OBJECT(table->widget), "button-release-event", G_CALLBACK(_event_button_release), table);
+  // set the size of icons and other elements
+  g_signal_connect(G_OBJECT(table->widget), "map", G_CALLBACK(dt_thumbtable_icons_register_size), table);
 
   // we register globals signals
   dt_control_signal_connect(darktable.signals, DT_SIGNAL_MOUSE_OVER_IMAGE_CHANGE,
--- a/src/dtgtk/thumbtable.c
+++ b/src/dtgtk/thumbtable.c
@@ -1530,7 +1530,7 @@ static void _event_dnd_end(GtkWidget *widget, GdkDragContext *context, gpointer 
   gtk_style_context_remove_class(tablecontext, "dt_thumbtable_reorder");
 }
 
-void dt_thumbtable_icons_register_size(GtkWidget *widget, gpointer user_data)
+void dt_thumbtable_icons_register_size(GtkWidget *widget, GdkRectangle *allocation, gpointer user_data)
 {
   GtkStateFlags state = gtk_widget_get_state_flags(widget);
   GtkStyleContext *context = gtk_widget_get_style_context(widget);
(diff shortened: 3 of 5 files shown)


src/dtgtk/culling.c
```diff
@@ -825,6 +825,8 @@ dt_culling_t *dt_culling_new(dt_culling_mode_t mode)
   g_signal_connect(G_OBJECT(table->widget), "button-press-event", G_CALLBACK(_event_button_press), table);
   g_signal_connect(G_OBJECT(table->widget), "motion-notify-event", G_CALLBACK(_event_motion_notify), table);
   g_signal_connect(G_OBJECT(table->widget), "button-release-event", G_CALLBACK(_event_button_release), table);
+  // set the size of icons and other elements
+  g_signal_connect(G_OBJECT(table->widget), "map", G_CALLBACK(dt_thumbtable_icons_register_size), table);
```
AlicVB, line 829: this works only by accident : I've forgotten to remove the` #include "dtgtk/thumbtable.h"` (to be fix)
culling and thumbtable should stay independent...

data/themes/darktable.css
```diff
@@ -1841,6 +1841,15 @@ Details :
   background-color: @lighttable_bg_color;
 }
 
+/* here we set the maximum size of GUI elements in thumbtable view to be in synch with the rest of the GUI */
+#thumbtable_filemanager,
+#thumbtable_filmstrip,
+#thumbtable_zoom,
+#culling
+{
+min-height: 1.47em;
```
AlicVB, line 1850: I'm not fan of abusing css names like that... But maybe that's the only solution...

Nilvus: Me too. Hope a better solution could be possible.

AlicVB:
@Mark-64 : I'm not sure that's the right way to follow : apart the comment made in the review, we loose completely the initial goal of your precedent work : uniformizing the icon size between thumbnails icons and toolbar ones.
I'm sure we can do something with font-size. I'll do some tests and report back.

Anyway, thanks for your work on that !

AlicVB:
@Mark-64 : I've done some empiric tests, and it seems that currently, `max_size = 1.133333 * darktable.bauhaus->font_height` I've done many tests with different font sizes.
So using that, we should be fairly close to toolbar icon sizes.
but of course, we'll loose the uniformity if one change buttons margins/paddings

Mark-64:
`we loose completely the initial goal of your precedent work : uniformizing the icon size between thumbnails icons and toolbar ones.`
Well, not really, we just move to the css theme the responsibility to keep them in synch, meaning if someone changes the toolbar icons in the theme, he/she has to remember to change another property (whatever it is) accordingly, to maintain uniformity.
The problem is that there are visualization modes where there is no widget to make reference to. That's why I thought to refer to the thumbtable ancestor. I will study a bit more if it is possible to avoid the race condition.

`max_size = 1.133333 * darktable.bauhaus->font_height `
I have to check if this work in Windows too, with pango win32 backend

Mark-64:
By the way, connecting to size-allocate event instead of map seems to work also in culling, can you test ?
After all, the size-allocate event of the parent should always happen before the child.

Mark-64:
@AlicVB 
`max_size = 1.133333 * darktable.bauhaus->font_height`

there is no font_height field, can you clarify ?

AlicVB:
> there is no font_height field, can you clarify ?

oops... my bad, it's `line_height` sorry for the error...

Mark-64:
`I have to check if this work in Windows too, with pango win32 backend`
It works, but it's not perfect.
In my case, I need 1.2 factor in Ubuntu to have a perfect match, but 1.23 in Windows.
Measured on the "group" icon (by the way the "altered" icon is a bit smaller, we should fix that).

AlicVB:
hmmm... you're right, seems that it's not so simple...



> Also, I don't see how this approach is serving the purpose to synch the toolbars and the thumbtable, unless we use bauhaus to set the toolbar icons as well

because initially, I thought that the toolbar icon size was dependent of the toolbar height, which seems to be fixed by labels height...
but there's some other stuff involved here : if you change the whole interface font-size, the toolbar buttons ratio is not always square, so the icon size is not always correlated to the toolbar height.

Maybe we should find how we determine the icon size for toolbar buttons and try to mimic that...

AlicVB:
> Maybe we should find how we determine the icon size for toolbar buttons and try to mimic that...

... and I don't find it. Seems that it's gtk who decide the width... Question : do we want that ? don't we prefer square buttons in order to keep the corresponding size between font-size and icon-size ? in that case we may fall back to more predictable icon-sizes...
I'll test that in the evening

Mark-64:
Gtk size allocation is a nightmare, let's not even think to mess with that.
I believe the cleanest way is to ficticiously use a css property of some gui element that we can set manually in em units to match the desired size. That way all the element are expressed in em units and they undergo the same transformation rules (OS dependent) so the proportion is kept.
But, hey, in case its too complex we can live with your solution, the difference is barely visible.

Nilvus:
> Well, not really, we just move to the css theme the responsibility to keep them in synch, meaning if someone changes the toolbar icons in the theme, he/she has to remember to change another property (whatever it is) accordingly, to maintain uniformity.

That's a problem for me. It's better to have all related thing together or it will be a mess to remember that something needs to be adjusted if changing another setting (or at worst, they should be together on code if possible). We can't say on CSS for example, if changing this, think about adjusting that way setting on line xxx (as line could be changed anytime and will be hard to not forget updating comment).

But in result, I would just like to see icons as they actually are. I'm ok if we use CSS but if only it's possible to relate things together, so same settings for related/similar things as much as possible. Yes, it add probably difficulty but that too avoid others in the future.

AlicVB:
@Mark-64 : I propose another solution, maybe not optimal, but simpler imho : 
- keep the code we actually have in master
- add a fallback `if (max_size < 2) max_size = ???*darktable.bauhaus->line_height`

that way we keep the same size of icons between toolbar and thumbnails if the toolbar is visible and in the other case, we still have "almost" the same size and more important this size stay in synch to font-size.
The only case where we'll have little glitch (thumbnails with slightly different icon sizes) is if we start dt without the toolbar and show it after. But well... that's acceptable, imho

Mark-64:
Ok, I will implement that, but in another PR. I close here

PR #5490, opened 2020-06-16
thumb : fix colorlabels size

### LINKED ISSUE #5484: Issue on colorlabels: size change after setting it when changing the view
**Describe the bug**
Set a colorlabel on a thumbnail, the size is ok, just move the view to hide related thumbnail. Go back and see how big colorlabels become (and so margins not respected).

Another way to see that is using culling or preview mode: set a colorlabel or more and then go on that view to others images (be sure that image set is no more visible) and go back to those colors.

@AlicVB: sorry for that but probably another one for you. Could this be related to your last PR about margins in overlays? I don't remember that was there before.

AlicVB:
this fix #5484


TurboGit:
Thanks !

PR #5494, opened 2020-06-16
Reverted changes to introspection translation

dterrahe:
...in #5335 at request of @houz on IRC and added fieldname fallback for labels to _from_params functions instead.

### DIFF
--- a/src/develop/imageop_gui.c
+++ b/src/develop/imageop_gui.c
@@ -172,7 +172,18 @@ GtkWidget *dt_bauhaus_slider_from_params(dt_iop_module_t *self, const char *para
       g_signal_connect(G_OBJECT(slider), "value-changed", G_CALLBACK(generic_slider_int_callback), p + f->header.offset);
     }
 
-    dt_bauhaus_widget_set_label(slider, NULL, gettext(f->header.description));
+    if (*f->header.description)
+    {
+      dt_bauhaus_widget_set_label(slider, NULL, NC_("introspection description", f->header.description));
+    }
+    else
+    {
+      str = dt_util_str_replace(f->header.field_name, "_", " ");
+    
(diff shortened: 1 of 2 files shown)


TurboGit:
This is not working for translation. First I have applied this commit. Build fine and no issue to run. I go into perspective correction and I have all translated. Fine.

I go into po and update the french translation and see 13 strings. Strange they are in ashift module due to recent changes.

I start poedit and translate them, they are all in the "introspection description" context.

I then rebuild and then see that the ashift corresponding strings are all in English now. There is no way to translate them.

My thinking is that those strings are set for translaltion in the "introspection description" but does not get checked into this context for display.

TurboGit:
BTW, what was the issue with previous implementation that we want to fix here ?

dterrahe:
The slider/combo labels that were previously marked in the module itself are now picked up by introspection and marked for translation there with the "introspection description". I was using NC_ in the _from_params functions to translate them with context, but that can't work. I've now corrected that to  g_dpgettext2() but it still doesn't see translations. 

If I run intltool_update fr (after all the build/source/iop/introspection_*.c files have been regenerated) I get 222 fuzzy translations. I'm not familiar with the translation process; does fuzzy mean that g_dpgettext2 will fall back to a non-contextual translation automatically, or do the "existing" translations (i.e. the ones without the introspection context) need to be confirmed first? I assumed the latter and that this is why I don't pick up any translations.

dterrahe:
> BTW, what was the issue with previous implementation that we want to fix here ?

@houz requested that translations marked via introspection have the "introspection description" context set.

TurboGit:
Did I really closed it ????? That was not intentional sorry.

TurboGit:
I cannot reopen a PR, can you create a new one.

dterrahe:
NP. Reopened as #5498

PR #5496, opened 2020-06-16
tone curve: reverse order first two enums in colorspace dropdown

dterrahe:
This could be a helper function for bauhaus combobox that looks up the correct position of any enum, but out of order enums currently only used in two places (other in demosaic).

### DIFF
--- a/src/bauhaus/bauhaus.c
+++ b/src/bauhaus/bauhaus.c
@@ -1215,6 +1215,24 @@ gboolean dt_bauhaus_combobox_set_from_text(GtkWidget *widget, const char *text)
   return FALSE;
 }
 
+gboolean dt_bauhaus_combobox_set_from_value(GtkWidget *widget, int value)
+{
+  dt_bauhaus_widget_t *w = DT_BAUHAUS_WIDGET(widget);
+  if(w->type != DT_BAUHAUS_COMBOBOX) return FALSE;
+  dt_bauhaus_combobox_data_t *d = &w->data.combobox;
+  int i = 0;
+  for(GList *iter = d->entries; iter; iter = g_list_next(iter), i++)
+  {
+    const dt_bauhaus_combobox_entry_t *entry = (dt_bauhaus_combobox_entry_t *)iter->data;
+    if(GPOINTER_TO_INT(entry->data) == value)
+    {
+      dt_bauhaus_combobox_set(widget, i);
+      return TRUE;
+    }
+  }
+  return FALSE;
+}
+
 int dt_bauhaus_combobox_get(GtkWidget *widget)
 {
   dt_bauhaus_widget_t *w = DT_BAUHAUS_WIDGET(widget);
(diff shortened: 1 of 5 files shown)


dterrahe:
Actually, let me look tomorrow at fixing this in a better way using g_list_index

TurboGit:
Ok, ping me when better fix is ready. Thanks.

dterrahe:
I'm working on fixing this the "correct" way, by adding a new call dt_bauhaus_combobox_set_from_value to bauhaus.c
But for this to work correctly, I should track down and solve a bug in enum introspection because combobox_from_params based on enums should no longer be combined with manually added combobox entries (or dt_bauhaus_combobox_add_full has to be used). Also, I'd need to test that is not happening anywhere else besides the case in filmicrgb that I am aware of.

So maybe merge the current one line fix first, before someone trips over the issue (not that it is a critical one) and I'll try to land the correct fix in the next few days (after fixing ashift)

dterrahe:
Sorry for going back and forth on this; here is dt_bauhaus_combobox_set_from_value , with the temporary workaround in filmicrgb improved so it works with this as well, until I track down the enum introspection issue.

dterrahe:
@TurboGit This is now ready and merge-unconflicted.

If you think the bauhaus are inappropriate post-freeze, please just cherry-pick c84f8440d0e9496f73a299fcc84136c0c6c05731

TurboGit:
@dterrahe : can you explain what was wrong ? I'm trying to understand the issue we add, but using c84f844 on tonecurve I see not difference that current behavior. I may be looking at the wrong thing...

dterrahe:
enums can now define their values in any order they choose and that will be the order displayed in the combobox. So tone curve has the order it had previously in the gui:
```
  DT_S_SCALE_AUTOMATIC = 1,
  DT_S_SCALE_MANUAL = 0,   
  DT_S_SCALE_AUTOMATIC_XYZ = 2,
  DT_S_SCALE_AUTOMATIC_RGB = 3, 

```
and the default handler will fill p->tonecurve_autoscale_ab with a 1 if you select the first element. 

But if you leave and return to the image, gui_update would use that 1 to set the active element in the combobox, which would be DT_S_SCALE_MANUAL = 0

Just a visual glitch; the wrong value won't get used or saved, it only gets saved when it is changed, but obviously it looks wrong and confusing.

The non standard order is only used here and in demosaic (which also has an adjustment for it in gui_update  ` dt_bauhaus_combobox_set(g->demosaic_method_xtrans, p->demosaicing_method & ~DEMOSAIC_XTRANS);` ) but in theory every user of combobox_from_params _could_ use set_from_value in gui_update to be robust under reordering. Not that I expect there to be much requirement for this feature. Maybe if options get dropped/replaced in enums and we want to be able to remove them completely without renumbering the remaining values. Or if a new option is introduced and we want to make it the default/top item.

TurboGit:
Ok, I see, I have just pushed c84f844. The remaining of this PR will be merged after 3.2 to be safe. Thanks for the detailed explanations.

dterrahe:
Perfect. I'll close this now and submit a new one for 3.4 once I've also sorted out the introspection issue. Thanks!

PR #5498, opened 2020-06-16
Reverted changes to introspection translation

dterrahe:
...in #5335 at request of @houz on IRC and added fieldname fallback for labels to _from_params functions instead.

Reopened after #5494 accidentally closed.

### DIFF
--- a/src/develop/imageop_gui.c
+++ b/src/develop/imageop_gui.c
@@ -172,7 +172,20 @@ GtkWidget *dt_bauhaus_slider_from_params(dt_iop_module_t *self, const char *para
       g_signal_connect(G_OBJECT(slider), "value-changed", G_CALLBACK(generic_slider_int_callback), p + f->header.offset);
     }
 
-    dt_bauhaus_widget_set_label(slider, NULL, gettext(f->header.description));
+    if (*f->header.description)
+    {
+      // we do not want to support a context as it break all translations see #5498
+      // dt_bauhaus_widget_set_label(slider, NULL, g_dpgettext2(NULL, "introspection description", f->header.description));
+      dt_bauhaus_widget_set_label(slider, NULL, gettext(f->header.description));
+    }
+    else
+    {
+      str = dt_util_str_replace(f->header.field_name, "_", " ");
+    
+      dt_bauhaus_widget_set_label(slider, NULL, _(str));
+
+      g_free(str);
+    }
   }
   else
   {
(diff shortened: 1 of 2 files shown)


TurboGit:
I continue discussion started in #5494 here...

> I'm not familiar with the translation process; does fuzzy mean that g_dpgettext2 will fall back to a non-contextual translation automatically, or do the "existing" translations

It may means that now all translated strings from introspection are correctly set with the introspection context but it also means that all strings must be translated again :(

Indeed if you force a new context on all current strings they have to be translated again. That's certainly not a good news for translators.

Let me check if it is the case here...

dterrahe:
Is there a tool/command that approves all fuzzy translations (I'm guessing this means intltool recognizes almost everything is the same except context?) in one go without needing to approve one by one?

TurboGit:
I can confirm that all strings from the introspection are now translatable but the bad news as I had anticipated is that those strings must all be translated again. A no go to me.

>  in one go without needing to approve one by one?

Yes, you need to have a look one by one. A fuzzy translation may be completely wrong. In my case only the introspection ones are to be validated again so I could use an automatic button if one was available but for other languages not up-to-date since some time you really have to look at them one by one.

I suppose the solution for 3.2 is:
- keep current version which is ok
- amend the current PR to not use the introspection context

I think 2 is probably the way to go and we could maybe find a better alternative after 3.2.

What do you think ?

BTW, I don't see the point of having a special context for all introspection strings.

dterrahe:
I agree that for now 2 is best path.

I'm not sure what the benefit of an introspection context could be. For now, there are zero other users of the $DESCRIPTION tag and I don't see a benefit during the current exercise, but houz may well have concrete plans that require it.

I _would_ see a $CONTEXT tag (or a $DESCRIPTION: "context::actual description" format) being needed sometime in future.

So for now, shall I remove the context but maybe leave the current lines as a comment/todo, like this:
```
//  $result = "NC_(\"$GETTEXT_CONTEXT\", $result)" if($string ne "");
  $result = "N_($result)" if($string ne "");

```
and 
```
//     dt_bauhaus_widget_set_label(slider, NULL, g_dpgettext2(NULL, "introspection description", f->header.description));
    dt_bauhaus_widget_set_label(slider, NULL, gettext(f->header.description));

```

TurboGit:
Yes, you can let the old version in comment but please add a why this is in comment otherwise this will just be noise and will be removed in code review by someone at some point.

// we do not want to support a context as it break all translation see #5498 

Likewise in Perl code. Thanks.

dterrahe:
done

TurboGit:
Ok, better. I still have some fuzzy but far less and I don't see how we could fix that. So I'll merge this now. Thanks.

Issue #5348, opened 2020-06-08
Latest master (3.1.0+2001~gd337f2b04) segfaults (data.db related)


jade-nl:
**Describe the bug**
After pulling (and later cloning) the latest master it does not start but segfaults and creates a backtrace.

**To Reproduce**
1. Upgrade to latest
2. start darktable
3. Say yes to db update
4. segfault

After that no db update questions, but it keeps segfaulting when trying to start.

**Expected behavior**
darktable starts up as expected.

**Platform (please complete the following information):**
* OpenCL installed : Yes (nvidia-opencl-common 440.82-1~bpo10+1)
* OpenCL activated : Yes
* Graphics card    : GeForce RTX 2060 SUPER7979 MiB
* Graphics driver  : nvidia (440.82)
* OS               : Linux (kernel 5.4.0-0.bpo.4-amd64)
* Distro           : Debian Buster (rel. 10.4)

**Additional context**
* Self compiled : Yes.
* Compilers     : gcc 8.3.0 - clang 10.0.1 - LLVM 10.0.1
* Cmake         : 3.13.4
* Gtk           : 3.24.5-1
* Glib          : 2.58.3-2
* Sqlite        : 3.27.2
* Optimization  : None

I've tried a few things and it seems that the data.db file is what is causing the problem. If I remove that one darktable starts normally.

Restoring the created backup (data.db-pre-3.1.0) doesn't help.

BT: [darktable_bt_P87DL0.txt](https://github.com/darktable-org/darktable/files/4747138/darktable_bt_P87DL0.txt)

The db file causing the segfault: [data.db.zip](https://github.com/darktable-org/darktable/files/4747142/data.db.zip)

Although the solution seems "simple" I decide to create a bug report and include the bt and db file. This upgrade problem/issue might not only limited to me.

jade-nl:
Problem solved itself: Very latest master (3.1.0+2003~g2ee22dac3) works with the old data.db.

Closing this due to user impatience......

coffeeandspaceships:
I suspect https://github.com/darktable-org/darktable/commit/4f7ba47e58a8ec3ed5442b5ce88aaa4e934a1b26 fixed it. The debug statement(s) in question were added about 4 years ago so they were certainly in your previous affected build.

PR #5501, opened 2020-06-16
Thumbtable icon size 3

### LINKED ISSUE #5435: Thumbnail overlays → Hidden after dt restart

Mark-64:
This PR is to close #5435. See previous discussions in #5348 
Bonus: small size increase of "altered" icon to match "grouping", they are side to side in thumbnails and even a small difference is visible


Nilvus:
Just tested and it's all good too for me. I agree, more simple than previous PR, so much better. Size difference with or without toolbar could not be seen for users. I didn't notice it without measuring them precisely (or compare like @AlicVB), so no need to do better there. Good work guys, previous PR was not something I liked but I love this way.

TurboGit:
Thanks !

Issue #5499, opened 2020-06-16
Little things broken with #5335


Nilvus:
@dterrahe: 2 minor issues shown, probably related to your recent PR #5335.

- On contrast module, detail line on both modes have the default pointer (white point below slider) not aligned on the left. By default, should be aligned on the left (to 0%) as other ones.
![2020-06-16_19-30](https://user-images.githubusercontent.com/45535283/84807639-d69b8f00-b007-11ea-9d74-513c6d30aae0.png)

- On filmic rvb, go to look tab. See contrast slider. Before it was set with 3 digits after point, so for example 1.360 instead of 1.36. It's better for that setting as before. But more of that (maybe related), it was possible to increase/decrease by 0.01 only with mouse wheel. Now it increase/decrease by 0.05. Not suitable for that setting. And if using mouse wheel with Ctrl shortcut, it's only possible to increase by 0.01 but not to decrease.

**Expected behavior**

Just as before, default pointer on detail in contrast module aligned on the left of the slider
And for filmic rvb, possibility to increase/decrease by 0.01 directly with mouse wheel and so having 3 digits after point instead of 2.

dterrahe:
Thanks @Nilvus for the detailed comments. It is very easy to tailor stepsizes and number of digits, so I can add these no sweat. Have you gone over all modules you care most about or shall I wait and include these in a slightly bigger patch?

There definitely other cases where the granularity changed. Sometimes I intentionally didn't replicate the original settings because I wanted confirmation that they were really sensible, rather than just copied through without thought from a different setting. The number of digits sometimes looked crazy for example (for the order of magnitude of the parameter).

Nilvus:
> Thanks @Nilvus for the detailed comments. It is very easy to tailor stepsizes and number of digits, so I can add these no sweat. Have you gone over all modules you care most about or shall I wait and include these in a slightly bigger patch?

I have quickly gone over modules I care about the most and don't see by now anything else.

> There definitely other cases where the granularity changed. Sometimes I intentionally didn't replicate the original settings because I wanted confirmation that they were really sensible, rather than just copied through without thought from a different setting. The number of digits sometimes looked crazy for example (for the order of magnitude of the parameter).

Probably. For contrast I rarely use the third digit but that could sometimes help on this slider.

**Another thing:** on local contrast, detail string seems to not be translated anymore. I use french translation and we lost (as you can see on screenshot) accent on the e of the string. I update and check fr.po and no more detail string alone on it.

[...]

PR #5505, opened 2020-06-16
Correct the slider origin when factor and or offset used

dterrahe:
This impacts:
local contrast
color balance
clipping
split-toning

fixes first issue in #5499


TurboGit:
Working on my side. Thanks.

Issue #5502, opened 2020-06-16
add. exposure instances should have default bias=off, EV=0 currently it is EV=1


AxelG-DE:
**Describe the bug**

We got a little regression through the introspection for exposure

**To Reproduce**

**Expected behavior**

any additional instances of exposure should be bias=off, EV=0 

**Bisect**

```
169bbb36908be2f623bbd0763231fc6e9aba1a1c is the first bad commit
commit 169bbb36908be2f623bbd0763231fc6e9aba1a1c
Author: Diederik ter Rahe <dterrahe@yahoo.com>
Date:   Mon Jun 1 10:01:48 2020 +0100

    [exposure.c] introspection, refactoring, janitorial

 src/iop/exposure.c | 282 +++++++++++++++++++----------------------------------
 1 file changed, 100 insertions(+), 182 deletions(-)
```

So @dterrahe I suppose this one is for you :-)

elstoc:
Reverts changes made in #5414

elstoc:
@TurboGit I think similar changes have been reverted in filmicrgb as well. Might be worth spot-checking the whole of that PR

TurboGit:
@elstoc : yes I was expecting glitches like this, the review could not detect such cases. field testing will help detecting and fixing all those issues hopefully. @dterrahe I'm letting you looking at this one. Thanks.

dterrahe:
@elstoc I believe in the exposure module the whole reload_defaults can now be removed, since the code that was dependent on the loaded image and the "scene-referred" setting was already removed pre-merge of #5335 and the defaults are automatically picked up from introspection:
```
  float exposure; // $MIN: -18.0 $MAX: 18.0 $DEFAULT: 0.0
  int compensate_exposure_bias; // $DEFAULT: FALSE
```
Let me know if I'm missing something.

elstoc:
I'll take your word for it. If you've removed it and it works with introspection then I'm happy.

dterrahe:
> I'll take your word for it. If you've removed it and it works with introspection then I'm happy.

Don't do that!! Test!

;-)

elstoc:
Done - feedback on the PR

TurboGit:
Fixed, closing.

PR #5506, opened 2020-06-16
Small fixes for breakage/step sizes caused by introspection merge #5335

dterrahe:
Addresses, hopefully fixes, #5499 and #5502.

### DIFF
--- a/src/iop/bilat.c
+++ b/src/iop/bilat.c
@@ -410,6 +410,7 @@ void gui_init(dt_iop_module_t *self)
   gtk_widget_set_tooltip_text(g->mode, _("the filter used for local contrast enhancement. bilateral is faster but can lead to artifacts around edges for extreme settings."));
 
   g->detail = dt_bauhaus_slider_from_params(self, "detail");
+  dt_bauhaus_slider_set_step(g->detail, 0.01);
   dt_bauhaus_slider_set_factor(g->detail, 100);
   dt_bauhaus_slider_set_offset(g->detail, 100);
   dt_bauhaus_slider_set_format(g->detail, "%.0f%%");
@@ -431,6 +432,7 @@ void gui_init(dt_iop_module_t *self)
 
   g->highlights = dt_bauhaus_slider_from_params(self, "sigma_r");
   dt_bauhaus_slider_set_hard_max(g->highlights, 2.0);
+  dt_bauhaus_slider_set_step(g->highlights, 0.01);
   dt_bauhaus_widget_set_label(g->highlights, NULL, _("highlights"));
   dt_bauhaus_slider_set_factor(g->highlights, 100);
   dt_bauhaus_slider_set_format(g->highlights, "%.0f%%");
@@ -438,6 +440,7 @@ void gui_init(dt_iop_module_t *self)
 
   g->shadows = dt_bauhaus_slider_from_params(self, "sigma_s");
   dt_bauhaus_slider_set_hard_max(g->shadows, 2.0);
+  dt_bauhaus_slider_set_step(g->shadows, 0.01);
   dt_bauhaus_widget_set_label(g->shadows, NULL, _("shadows"));
   dt_bauhaus_slider_set_factor(g->shadows, 100);
   dt_bauhaus_slider_set_format(g->shadows, "%.0f%%");
--- a/src/iop/colorbalance.c
+++ b/src/iop/colorbalance.c
@@ -2078,12 +2078,14 @@ void gui_init(dt_iop_module_t *self)
 
   g->saturation = dt_bauhaus_slider_from_params(self, "saturation");
   dt_bauhaus_slider_set_soft_range(g->saturation, 0.5f, 1.5f);
+  dt_bauhaus_slider_set_digits(g->saturation, 4);
   dt_bauhaus_slider_set_factor(g->saturation, 100.0f);
   dt_bauhaus_slider_set_format(g->saturation, "%.2f %%");
   gtk_widget_set_tooltip_text(g->saturation, _("saturation correction before the color balance"));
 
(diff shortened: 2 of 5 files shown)


elstoc:
You're still missing a change from #5414. The filmicrgb module should always have ``default_enabled = FALSE`` regardless of workflow as it now gets automatically inserted in ``develop/develop.c``.

Exposure module looks fine now.

elstoc:
All good now thanks

TurboGit:
Seems to have fixed the issue I reported.

Now I'm wondering if on the long run we should not do that to have a uniform settings of all sliders:
- Have a scroll handled to be 1/20 of the MIN .. MAX range
- Have a shift-scroll handled to be 1/10 of MIN .. MAX range
- Have a ctrl-scroll handled to be 1/100 of MIN .. MAX range

All this being the default and we could add a step in introspection that would set the default scroll step and have shift & ctrl modifier do a *2 and /10 respectively of the base step.

How does that sound ?

dterrahe:
I was thinking of trying something like this with #5335 but I thought it would just be too controversial, because it would change the "feel" that everyone got used to with the current settings and we would just end up putting overrides everywhere. I did create a table of the default step sizes for all sliders (see #4804) and based my 1/100 of MAX ( -MAX .. MAX ranges just tended to have the same stepsize as 0 .. MAX ranges) on those investigations. And then I put overrides in where I thought the existing deviating values were somewhat sensible.

If you _are_ willing to experiment with this kind of change, then I'd advice not now :-) This can wait till 3.4 because lot less impact. And we should have a discussion before to agree the values. I think normal scroll would end up closer to 1/100 or 1/50 and shift *5 and ctrl /5 or something like that.

If I had spend more time on polishing, I would have linked set_step, set_digits and set_format in such a way that if you change one, the other will change to at least be sufficient to represent it. So setting step to .005 would set digits to 3 and format to "%.3 %%" (set_format would just set the extension and this would combine internally with digits). Of course this would break compatibility with current so have unintended side effects and needs many eyes to correct again. There would be complaining for days! ;-)

TurboGit:
@dterrahe : I understand the issue with habits, maybe indeed something delicate to change. But anyway, this idea was indeed only for after 3.2 if at all.

hgmarty:
A short note to participate a little, just to try not to forget a few users.

I use keyboard shortcuts for my part. With recent updates, the steps are higher. For example, exposure, a click increases or decreases by 0.1 which is a pretty rough setting. You have to finish with the mouse on the slider to have something more precise. You lose a little bit the speed of development in this case. This also applies to dynamic settings.

These shortcuts that I didn't apprehend at the beginning are powerful. It would be a shame not to have them in the best conditions.

It is of course difficult to satisfy all uses, I understand. But if you could take this into account in further development.

On the other hand, a small idea would make it possible to satisfy the different types of uses. That would be a multiplying factor to put in the preferences. For example, a factor of 0.5 will decrease by 2 the basic step, thus allowing an adjustment by using the keyboard shortcuts or the mouse by the dynamic mode.

Thanks

elstoc:
@hgmarty #5570

hgmarty:
lol
Small newbie I am... I don't very well understand github...
So it arrives good news thx @elstoc

elstoc:
Well let's wait and see if it gets merged though - it's just a proposal at the moment.

dterrahe:
At the same time, I restored previous values in #5572...

Now it is up to @TurboGit to decide.

Although if I were to try to read his mind, then #5570 might feel too much like a new feature, after the freeze.  With translateable strings even! Hence why I tried to keep my change as small as possible and still hopefully not disappoint anyone by causing a deterioration in their experience.

I keep repeating myself though; all of these aspects keep touching each other in that not even step size defaults or number of displayed digits can really be fixed without taking a step back and looking really hard at the whole input layer and coming up with an approach that addresses everything with one consistent overall approach. Patching little bits here and there just prevents us from coming up with a really flexible system that everyone can set up to their liking. So we end up with this weighing of compromises that work well for one (mouse) and not so well for another (shortcut keys, mouse wheel) and not at all for the next thing (midi).

TurboGit:
@dterrahe : I have merged #5570 as it seemed very safe to me. The string freeze will happen end of month. Currently we are in a "lazzy" feature freeze mode where only small & safe additions are considered. I'm doing that because of the introspection have shuffled many part of the code in part and also because some features are really going to polish the current new features added for 3.2. We will "hard-feature freeze" at the same time as string freeze (so end of month), and from there only bugfixes should go in.

And again there might be one exception for the OpenCL implementation of FilmicV4 worked on by Aurélien if ready not too close to the final release. This is somehow a bug or feature depending on how you view things. The added speed for the OpenCL implementation can really be seen as a bug fix. Also with the integration test we will get more confident about the quality of the result.

We usually have a bit more than 3 months to prepare the release and here that's a bit more than 2 months. So I'm trying some compromise :)

dterrahe:
All agreed. What I noticed during the introspection refactoring was that there are a lot of interactions between step/digits/formats and min/max range (also soft) that I don't think have been really considered carefully. So I definitely would like to revisit that part for 3.4, as per your comment here https://github.com/darktable-org/darktable/pull/5506#issuecomment-645190579

One of the aspects I'm thinking of is that if the stepsize, say, is 0.05 and format is %.2f and digits is 2, then a fine step won't have any effect. Whereas if digits is 3, it all looks the same, but now scrolling with a dynamic mouse shortcut will slowly (in two steps), increase from 0.05 -> 0.06 etc. So we need to decide whether those "hidden digits" add value, so we indeed keep a fine step as 1/10 -> 0.005, or whether we would round it up to the nearest "visible" small step, i.e. 0.01. Which would be harder to implement etc etc. Loads of room for discussion.

TurboGit:
> Which would be harder to implement etc etc. Loads of room for discussion.

Agreed to revisit this, yes.

PR #5507, opened 2020-06-16
hsl base setting

rabauke:
Adds an additional neutral preset to the "color zones" module. This preset has a neutral effect, it merely adds a set of nodes. Users may move these nodes up and down with active "edit by area" option to mimic an HSL slider known from various image processing tools.

### DIFF
--- a/src/iop/colorzones.c
+++ b/src/iop/colorzones.c
@@ -703,6 +703,25 @@ void init_presets(dt_iop_module_so_t *self)
   }
   dt_gui_presets_add_generic(_("black & white film"), self->op, version, &p, sizeof(p), 1);
 
+  // neutral preset with just a set of nodes uniformly distributed along the hue axis
+#define DT_IOP_COLORZONES_BANDS_HSL 8
+  p.channel = DT_IOP_COLORZONES_h;
+  for(int k = 0; k < DT_IOP_COLORZONES_BANDS_HSL; k++)
+  {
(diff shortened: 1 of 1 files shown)


TurboGit:
Why not ! Thanks.

src/iop/colorzones.c
```diff
@@ -703,6 +703,25 @@ void init_presets(dt_iop_module_so_t *self)
...
+    p.curve[DT_IOP_COLORZONES_h][k].x = (float)k / DT_IOP_COLORZONES_BANDS_HSL;
+    p.curve[DT_IOP_COLORZONES_h][k].y = 0.5f;
+  }
+  for(int c = 0; c < 3; c++)
+  {
+    p.curve_num_nodes[c] = DT_IOP_COLORZONES_BANDS_HSL;
+    p.curve_type[c] = MONOTONE_HERMITE;
+  }
+  dt_gui_presets_add_generic(_("HSL base setting"), self->op, version, &p, sizeof(p), 1);
+
```
johnny-bit, line 724: Shouldn't `DT_IOP_COLORZONES_BANDS_HSL` be undefined after adding preset? (sorry I missed this before merging of this PR)

TurboGit: I had the same thought, but I thought it was not that important. Anyway undef is ok.

johnny-bit: I just added #5521 I hope it's OK...

PR #5508, opened 2020-06-17
Eliminate unnecessary reprocessing for more responsive GUI

ralfbrown:
In a number of locations, existing code forces a complete reprocessing of both the full (center view) and preview pipes.  For retouch and blending, one can eliminate the reprocessing of the preview pipe when switching between different views (masks, wavelet scales, etc.) because switching views doesn't change the output of the preview pipe.  For clipping, colorchecker, and tone equalizer, the calls to dt_dev_reprocess_all can be eliminated entirely, as existing calls to dt_control_queue_redraw_center are entirely adequate for redrawing the screen.

Less reprocessing makes for faster screen updates and hence a more responsive GUI.  I've checked all of the affected modules and see no differences in behavior.

### DIFF
--- a/src/develop/blend_gui.c
+++ b/src/develop/blend_gui.c
@@ -871,7 +871,7 @@ static void _blendop_blendif_showmask_clicked(GtkWidget *button, GdkEventButton 
     if(module->off) gtk_toggle_button_set_active(GTK_TOGGLE_BUTTON(module->off), TRUE);
 
     dt_iop_request_focus(module);
-    dt_dev_reprocess_all(module->dev);
+    dt_dev_reprocess_center(module->dev);
   }
 }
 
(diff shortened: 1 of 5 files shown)


ralfbrown:
It's a mouse-focus issue of some kind -- if you click into the main image or the border around it between activating and deactivating the color picker, it is correctly removed.  That doesn't affect the color picker for blending, and forcing a refresh of the center image fixes the problem for colorchecker.  I haven't been able to track down where that box is drawn to see whether it's possible to do without the forced refresh.

I was also able to do a few quick experiments to determine how necessary the dt_dev_invalidate_all inside dt_dev_add_history_item really is.  Removing it entirely causes massive breakage, but replacing it by dt_dev_invalidate, which mostly invalidates just the "full" pixelpipe has most things working on a quick check (the two obvious problems were that the navigation image doesn't always get updated immediately, and things which require the preview pipe to have run first, such a "get structure" in ashift, don't work properly).  But the overall processing goes down considerably, so I would like to propose a PR to be implemented this coming weekend:  I'll go through the code and add the dt_dev_invalidate_all/dt_control_queue_redraw_center calls after all instances of dt_dev_add_history_item and remove them from the body of dt_dev_add_history_item.  That will be a noop operationally, but will then permit a more leisurely replacement/removal of the invalidate_all calls on a case-by-case basis in further PRs.  However, I'll only spend the hours if this is something which will be accepted for merging (it will be big, touching every iop and changing about 300 instances of add_history).

TurboGit:
Works for me now! Thanks.

PR #5510, opened 2020-06-17
Don't raise DT_SIGNAL_METADATA_UPDATE in dt_imageio_open

### LINKED ISSUE #5504: Duplicate manager seems to be broken
**Describe the bug**
Duplicate manager seems to be broken since ~2days

**To Reproduce**
Open an image with duplicates in darkroom, removing a duplicate or adding is possible, selecting another makes via double-left-click makes dt hang. Also just pressing left button doesn't show the comparing image any longer. 

**Expected behavior**
As before until ~ 2days.

**Screenshots**

**Platform**
 Darktable Version: 3.1.0 - 2263, Fedora 32, nvidia graphics, opencl does not matter.
 

**Additional context**
Have been testing this 2 days ago to understand the pixel shifting stuff and this worked fine.

jenshannoschwalm:
We can't raise that signal within dt_imageio_open. For sure this broke
duplicate, it might also be related to others.

Will have to find a safe way in 3.4

Fixes #5504

### DIFF
--- a/src/common/imageio.c
+++ b/src/common/imageio.c
@@ -1145,8 +1145,6 @@ dt_imageio_retval_t dt_imageio_open(dt_image_t *img,               // non-const 
 
   img->p_width = img->width - img->crop_x - img->crop_width;
   img->p_height = img->height - img->crop_y - img->crop_height;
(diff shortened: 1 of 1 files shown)


TurboGit:
Thanks!

PR #5515, opened 2020-06-17
thumbnail mouse hover fix

### LINKED ISSUE #5511: new lighttable keep highlighted thumb in one case
**Describe the bug**
Keep a selection when hovering above empty spaces in lighttable and this happens only the first time a collection is opened.

**To Reproduce**
1. Go to lighttable
2. Select a collection that is not too long and so let some empty space into the lighttable thumbtable
3. From collect module move over a picture and hover down in the lighttable until the mouse is above an empty space (no thumb)
4. See error, the last thumb hovered before going to empty space is still highlighted
5. Move back over a thumb
6. Go down to empty space the last thumb is not highlighted

So this issue happens only the first time.

**Expected behavior**
Thumbs should never be highlighted when hovering on empty space.

**Screenshots**

 - Darktable Version: master
 - OS: GNU/Debian

**Additional context**

AlicVB:
this fix #5511 
work around to gdk bug on leave events. I haven't noticed any drawback so far...

### DIFF
--- a/src/dtgtk/culling.c
+++ b/src/dtgtk/culling.c
@@ -21,7 +21,6 @@
 #include "common/debug.h"
 #include "common/selection.h"
 #include "control/control.h"
-#include "dtgtk/thumbtable.h"
 #include "gui/gtk.h"
 #include "views/view.h"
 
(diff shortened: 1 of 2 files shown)


TurboGit:
Working on my side! Thanks for the quick fix.

PR #5516, opened 2020-06-17
don't run iops which don't contribute to mask display

ralfbrown:
Unlike color-channel displays in blendop, the display of the blend mask was still running all of the iops
between the one with GUI focus and gamma, even though gamma then discarded their output.  This patch
eliminates that redundant computation in the full pixelpipe for a much smoother screen update as blending
parameters are modified while the mask display is active.  A separate proposed PR will eliminate the
unnecessary running of the preview pipe as part of much more extensive updates.

### DIFF
--- a/src/develop/pixelpipe_hb.c
+++ b/src/develop/pixelpipe_hb.c
@@ -1344,11 +1344,13 @@ static int dt_dev_pixelpipe_process_rec(dt_dev_pixelpipe_t *pipe, dt_develop_t *
 
     dt_pixelpipe_flow_t pixelpipe_flow = (PIXELPIPE_FLOW_NONE | PIXELPIPE_FLOW_HISTOGRAM_NONE);
 
-    // special case: user requests to see channel data in the parametric mask of a module. In that case
-    // we skip all modules manipulating pixel content and only process image distorting modules. Finally
-    // "gamma" is responsible to display channel data accordingly.
-    if(strcmp(module->op, "gamma") && (pipe->mask_display & DT_DEV_PIXELPIPE_DISPLAY_ANY) && !(module->operation_tags() & IOP_TAG_DISTORT) &&
-      (in_bpp == out_bpp) && !memcmp(&roi_in, roi_out, sizeof(struct dt_iop_roi_t)))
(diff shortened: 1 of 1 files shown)


MStraeten:
this pr affects the display selection in colorzones (right mask icon): instead of a grayscaled image it's turned into black/white:
mask_colorzones

ralfbrown:
Congratulations, you found a bug that was masked (pun intended) by the bug I fixed in the original commit :-).  Since the modules were being run even though their output was being ignored, the various colorspace conversions were still being done.  When I eliminated those runs, the recorded colorspace was still being updated even though the conversion wasn't running, so gamma was getting Lab in (0,100) while expecting RGB in (0,1), and then clamping the results to (0,1).

aurelienpierre:
(what about the OpenCL path ?)

ralfbrown:
Both patches in this PR are in a common code path.  Is there a problem with OpenCL?

MStraeten:
is fine also with OpenCL

TurboGit:
I don't think this will make it for 3.4 at this stage.

ralfbrown:
Is there some problem?  I fixed the reported bug.

johnny-bit:
I've tested - the bug reported by @MStraeten is indeed gone plus IMO the mask preview now works super fast, especially if previewing/modifying parametric mask on module that's early in the pipe...

@ralfbrown - It seems like travis hung on checking your PR, maybe rebase on top of master would kick it back into working?

@TurboGit - i think this PR is perfectly fine to merge, I've been hacking at it for last ~3h and testing without a single hitch both in opencl or no opencl. And it indeed seems to speedup mask display as advertised :)

TurboGit:
> Is there some problem? I fixed the reported bug.

No problem, I didn't notice you've fixed the issue and since I knew you were very short in time until end of year I removed it from 3.4 target. But since it is fixed, I'll re-target for 3.4.

Sorry, I have sometime hard time to follow everything.

TurboGit:
Merged, thanks!

ralfbrown:
> @TurboGit> Sorry, I have sometime hard time to follow everything.

No worries, it's a big job with lots of moving parts....

PR #5519, opened 2020-06-17
Do not popup combobox when empty

dterrahe:
Comboboxes are used as label in several places so that a button can be placed in their quad area. When clicked, those labels would show a 2 * INNER_PADDING black line underneath.

filmicrgb; tab1/auto tune levels, tab2/display mask
color balance; auto optimizers
unbreak input profile; auto tune levels
tone equaliser; display mask


TurboGit:
Thanks !

PR #5520, opened 2020-06-17
Improve Sqlite3 debug infos in dt code

johnny-bit:
The idea came when working on db integrity checks as I messed part of code that didn't have DT_DEBUG* wraps that would otherwise avoid problem.

This changes absolutely nothing in terms of logic, speed or execution, just adds more debugging output in more places essentially limiting use of plain `sqlite3_exec` to opening/closing transaction and db upgrades (which i'm very hesitant to touch)


johnny-bit:
Rebased and checked if works with recent master. No problems introduced :)

PR #5522, opened 2020-06-17
retouch speedup: bypass further iops when displaying wavelet scale

ralfbrown:
The existing code was passing the single wavelet scales through processing by all following modules in the
pixel pipe, which not only slowed things down, it also resulted in different displays depending on the
downstream processing (especially brightness adjustments from exposure, basecurve, filmic, etc.).  Denoising
would also hide details.  This patch adds a new channel-display type that the pixel pipe will treat like the color
channel display from blendop, and skip all non-distorting iops between retouch and gamma.

It may be desireable to tweak the default brightness up a little now that neither basecurve nor filmic is
being applied to the wavelet scale.


TurboGit:
@ralfbrown : If 3.2 will have a ful redesign GUI (and some other things of course) it looks like 3.4 will be improving the speed with all your PR on performance. Really nice to see this happening. Thanks.

TurboGit:
Thanks !

PR #5525, opened 2020-06-17
overlays block : force uniform margins

AlicVB:
@Nilvus : if you can have a look...


TurboGit:
Not part of this PR but I found that when zooming in/out in the lightttable I have many errors like

`(darktable:27850): Gtk-WARNING **: 09:11:50.455: GtkDarktableThumbnailBtn 0x561cdc8fd640 attempted to adjust its size allocation from -96,0 112x15 to 0,0 16x16. adjust_size_allocation must keep allocation inside original bounds`

And when this happens the starts seems squeezed on the right.

image

Will open another ticket.

PR #5529, opened 2020-06-17
make show/hide iop module shortcut more consistent

elstoc:
**The Problem**
Currently, if an iop module search is active and a user attempts to expand/collapse a module using a keyboard shortcut the behaviour is inconsistent and depends on the group that the user was in _before they initiated the search_.

For example, _the filmic rgb module is in the tone group_. Assign a keyboard shortcut to show/hide the filmic module (in my case SHIFT+F) and then attempt the following actions:

1. (a) Select the tone group. Search for the exposure module using the module search. Select the exposure module. Press SHIFT+F to show the filmic module. **Nothing Happens**. (b) Note that if you instead searched for, say "rgb" in this scenario, filmic would be in the search result list and the shortcut would just toggle it without cancelling the search.

2. (a) Clear the search and start again. Select the basic group. Search for the exposure module using the module search. Select the exposure module. Press SHIFT+F to show the filmic module. **The search box is cleared, darktable switches to the tone group and expands the filmic module (if it's not already expanded)**. (b) The same occurs if filmic is shown in the search results (e.g. when searching for "rgb").

The above results should be identical but depend on what module group you start in.

**The Fix**
This PR addresses this by making the behaviour consistent, when using a keyboard shortcut. If a module search is in progress, the search will be cancelled, the user will be switched to the appropriate module group and the module expanded or collapsed depending on its current state. If the module is already present in the group that was selected before the search was initiated (favourites, active), the user will be switched back to that group.

This should not change functionality where the search box or module groups are hidden.

### DIFF
--- a/src/develop/imageop.c
+++ b/src/develop/imageop.c
@@ -2253,9 +2253,14 @@ static gboolean show_module_callback(GtkAccelGroup *accel_group, GObject *accele
   }
 
   const uint32_t current_group = dt_dev_modulegroups_get(module->dev);
+
   if(!dt_iop_shown_in_group(module, current_group))
   {
     dt_dev_modulegroups_switch(darktable.develop, module);
+  }
+  else
+  {
+    dt_dev_modulegroups_set(darktable.develop, current_group);
   }
 
   dt_iop_gui_set_expanded(module, !module->expanded, dt_conf_get_bool("darkroom/ui/single_module"));
--- a/src/libs/modulegroups.c
+++ b/src/libs/modulegroups.c
@@ -579,13 +579,10 @@ static gboolean _lib_modulegroups_set_gui_thread(gpointer user_data)
 
   dt_lib_modulegroups_t *d = (dt_lib_modulegroups_t *)params->self->data;
 
-  const int group = _iop_get_group_order(params->group, params->group);
-
-  /* set current group or just update iop visibility*/
-  if(d->current != group && params->group < DT_MODULEGROUP_SIZE && GTK_IS_TOGGLE_BUTTON(d->buttons[params->group]))
+  /* set current group and update visibility */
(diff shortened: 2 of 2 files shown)


elstoc:
A couple of iterations and it's all much simpler now. Should be ready to go.

TurboGit:
I think I understand the description and I tried the PR but does not see what is described happening;
- when I have no search box active, ALT+F (activate filmicrgb on my side), the module is activated or deactivated but there is no selection of the group where this module is.
- when I have a search active, I have the same behavior as above.

Maybe I'm not understanding or confused by this PR ?

elstoc:
Ah I think I see the issue. This is about the "show module" shortcut functionality. Are you testing with the "enable module" shortcut?

TurboGit:
> Are you testing with the "enable module" shortcut?

Yes. Rereading your description I see that you were talking about show/hide. Anyway, good I made this mistake as I found a similar issue!

But even with the show/hide action I have cannot see the switch to the correct group. I selected the effect-group, started to enter a char in the search entry, did ATL-F and nothing happens (really nothing it does not move to the proper group).

elstoc:
Nothing will happen while you're in the search box as the gtkentry widget swallows all of your keyboard input. You need to click on a module header or the centre image view.

TurboGit:
@elstoc : Indeed, so show/hide is ok. The issue is only in enable.

elstoc:
Enable has never shown/expanded the module as well, just toggled whether it's enabled or not. I'm not trying to fix that and IMO that's not an issue.

At least that's how I would prefer it works.

TurboGit:
> At least that's how I would prefer it works. I quite often want to toggle a module while interacting with another (e.g. denoise profiled).

Makes sense indeed! That was just a procedural error on my side during first testing. So it seems that this can go in as all good then.

PR #5530, opened 2020-06-17
refactoring: eliminate duplicated boilerplate in lighttable modules

ralfbrown:
The same code was used in six separate modules to set up callbacks for a delayed update after a mouse move.  This PR replaces that boilerplate with calls to a pair of shared functions.

### DIFF
--- a/src/libs/copy_history.c
+++ b/src/libs/copy_history.c
@@ -45,7 +45,6 @@ typedef struct dt_lib_copy_history_t
   GtkWidget *copy_button, *delete_button, *load_button, *write_button;
   GtkWidget *copy_parts_button;
   GtkButton *compress_button;
-  guint timeout_handle;
 } dt_lib_copy_history_t;
 
 const char *name(dt_lib_module_t *self)
@@ -66,6 +65,7 @@ uint32_t container(dt_lib_module_t *self)
 
 static void _update(dt_lib_module_t *self)
 {
+  dt_lib_cancel_postponed_update(self);
   dt_lib_copy_history_t *d = (dt_lib_copy_history_t *)self->data;
 
   GList *imgs = dt_view_get_images_to_act_on(TRUE, FALSE);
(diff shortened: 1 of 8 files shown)


dterrahe:
Just a thought, but this still requires a bit of boiler-plate callbacks and signal registration for each module. Have you considered doing a more ambitious refactor where the _update functions get elevated to the status of the gui_init etc set and they all get called (in a delayed manner when appropriate) when the image selection/collection changes?

johnny-bit:
Some modules, like metadata display need immediate action on mouse over changes since those just read and display data, others like import module doesn't need to update at all and I think only those 6 do require some play with delayed update to limit unnecessary cycles... tbh in their case delayed update wouldn't be bad at all for all actions :wink:

If it would be good to do then I do agree with @dterrahe - make the copypasta I've done into deeper integration and in case no update is needed _update could be noop or null and it'll be totally fine :)

dterrahe:
The mechanism for defining virtual functions is that on library load in dt_lib_load_module if a definition isn't found, the redirection in dt_lib_module_t is set to null (or a default implementation, but that would not be needed here) and it simply wouldn't be called. So you never need to define a noop function. And if both the immediate and delayed action are common patterns, you could pass a bool to the function indicating whether this is a deferred call and you'd call it twice for each module.

I was only looking at the current pattern and spit-balled an alternative approach (because I myself also love people who say "hey nice job you did there, why don't you do one 10x bigger!) I'm not saying I have an opinion on whether it would be really much better/useful, but I always like to have plumbing hidden from view so that in future it won't get copied around too much by others who, for lack of background, have no better option than to treat it as magic incantations. There is no benefit _now_ obviously. I just recognise in @ralfbrown a kindred soul who likes things neat and tidy ;-)

OK, I'll shut up now.

TurboGit:
@ralfbrown : This one is ready too for review/integration, right?

ralfbrown:
Yes, it's ready.

TurboGit (approved):
Works for me, thanks!

PR #5533, opened 2020-06-18
Color Balance: set different soft ranges for shadows midtones highlights

Mark-64:
Same as #5234, rebased

With the aim to improve usability of Color Balance module and in reply to my own #5227 I propose this PR, where the soft range are differentiated per channel, instead of being all equal.

lift / offset (shadows)
factor: -5% -> +5% (instead of -50% -> +50%)
saturation: 0% -> 5% (instead of 0% -> 25%)
R/G/B: -0.05 -> +0.05 (instead of -0.5 -> +0.5)

gamma / power (midtones)
factor: -20% -> +20% (instead of -50% -> +50%)
saturation: 0% -> 20% (instead of 0% -> 25%)
R/G/B: -0.2 -> +0.2 (instead of -0.5 -> +0.5)

gamma / power (highlights)
factor: -50% -> +50% (instead of -50% -> +50%, unchanged)
saturation: 0% -> 50% (instead of 0% -> 25%)
R/G/B: -0.5 -> +0.5 (instead of -0.5 -> +0.5, unchanged)

The rationale is to have a span of control in each channel which perceptually produces a comparable change on the image and increase accuracy of setting, especially for shadows.
Yes the lift/gamma/gain method is less sensitive and needs less this change, but overall the usability looks pretty good to me with all methods.
Of course this is just a default soft range, you can always push it further if more adjustment power is needed.

@dterrahe please have a look if I rebased correctly on you work

EDIT: maybe soft ranges can be differentiated for algorithm too: lift/gamma/gain can have wider range vs slope/offset/power

### DIFF
--- a/src/iop/colorbalance.c
+++ b/src/iop/colorbalance.c
@@ -2179,7 +2179,7 @@ void gui_init(dt_iop_module_t *self)
 
   static const char *lift_messages[] = { N_("factor of lift"), N_("lift") };
   (void)lift_messages;
-  ADD_FACTOR(lift)
+  ADD_FACTOR(lift, 5.0f)
 
   g->hue_lift = dt_color_picker_new(self, DT_COLOR_PICKER_AREA, 
                 dt_bauhaus_slider_new_with_range_and_feedback(self, 0.0f, 360.0f, 1.0f, 0.0f, 2, 0));
@@ -2190,7 +2190,7 @@ void gui_init(dt_iop_module_t *self)
   gtk_box_pack_start(GTK_BOX(self->widget), g->hue_lift, TRUE, TRUE, 0);
   g_signal_connect(G_OBJECT(g->hue_lift), "value-changed", G_CALLBACK(hue_lift_callback), self);
 
-  g->sat_lift = dt_bauhaus_slider_new_with_range_and_feedback(self, 0.0f, 25.0f, 0.5f, 0.0f, 2, 0);
+  g->sat_lift = dt_bauhaus_slider_new_with_range_and_feedback(self, 0.0f, 5.0f, 0.05f, 0.0f, 2, 0);
   dt_bauhaus_slider_set_format(g->sat_lift, "%.2f %%");
   dt_bauhaus_slider_enable_soft_boundaries(g->sat_lift, 0.0f, 100.0f);
   dt_bauhaus_widget_set_label(g->sat_lift, NULL, _("saturation"));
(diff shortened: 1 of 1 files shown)


dterrahe:
The bits you touched weren't impacted by my work so all looks fine.

There should really be $MIN $MAX introspection tags defined on 
float lift[CHANNEL_SIZE], gamma[CHANNEL_SIZE], gain[CHANNEL_SIZE]; // $DEFAULT: 1.0
but since it isn't used in widget creation anyway they would not be enforced in the code.

If I were to (re)write the gui part of this today, I'd probably put GtkWidget pointers in dt_iop_colorbalance_gui_data_t in arrays, to match dt_iop_colorbalance_params_t. That way far fewer special case call-back functions would be required and a lot could be done in loops. But it would require careful untangling of current code.

aurelienpierre:
> lift / offset (shadows)
factor: -5% -> +5% (instead of -50% -> +50%)
saturation: 0% -> 5% (instead of 0% -> 25%)
R/G/B: -0.05 -> +0.05 (instead of -0.5 -> +0.5)

OK.

> gamma / power (midtones)
factor: -20% -> +20% (instead of -50% -> +50%)
saturation: 0% -> 20% (instead of 0% -> 25%)
R/G/B: -0.2 -> +0.2 (instead of -0.5 -> +0.5)

factor and RGB should stay ± 50%.  Ok for saturation.

> gamma / power (highlights)
factor: -50% -> +50% (instead of -50% -> +50%, unchanged)
saturation: 0% -> 50% (instead of 0% -> 25%)
R/G/B: -0.5 -> +0.5 (instead of -0.5 -> +0.5, unchanged)

Not sure about saturation. Did you find any picture where values as high as 50% were needed ?

Mark-64:
As requested, changed to

lift / offset (shadows)
factor: -5% -> +5% (instead of -50% -> +50%)
saturation: 0% -> 5% (instead of 0% -> 25%)
R/G/B: -0.05 -> +0.05 (instead of -0.5 -> +0.5)

gamma / power (midtones)
factor: -50% -> +50% (unchanged)
saturation: 0% -> 20% (instead of 0% -> 25%)
R/G/B: -0.5 -> +0.5 (unchanged)

gamma / power (highlights)
factor: -50% -> +50% (unchanged)
saturation: 0% -> 25% (unchanged)
R/G/B: -0.5 -> +0.5 (unchanged)

TurboGit:
Should we plan this for 3.2 ?

aurelienpierre:
It shouldn't hurt 3.2, it's trivial.

TurboGit:
> It shouldn't hurt 3.2, it's trivial.

I have seen many trivial things breaking in very unexpected ways :) A program seems to have far more resource than us to create big surprise :)

TurboGit:
That being said, tested and working fine. Code is clear, let's move this to 3.2. Thanks.

PR #5536, opened 2020-06-18
assign focus to center view when clicking on thumbtable

### LINKED ISSUE #5535: Lighttable (file manager) mode → Scrolling by keys
Linux x64, master snapshot:

It looks that scrolling by up and down keys was broken recently (no more than couple days). To reproduce it just click on random image in file manager mode and try using up/down keys to scroll. It does nothing.

elstoc:
Resolves #5535

### DIFF
--- a/src/dtgtk/thumbtable.c
+++ b/src/dtgtk/thumbtable.c
@@ -962,7 +962,7 @@ static gboolean _event_button_press(GtkWidget *widget, GdkEventButton *event, gp
   if(event->button == 1 && event->type == GDK_BUTTON_PRESS)
   {
     // make sure any edition field loses the focus
-    gtk_window_set_focus(GTK_WINDOW(dt_ui_main_window(darktable.gui->ui)), NULL);
+    gtk_widget_grab_focus(dt_ui_center(darktable.gui->ui));
   }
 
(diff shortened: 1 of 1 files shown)


aurelienpierre:
Thanks !

elstoc:
Do you mean ``// make sure any edition field loses the focus``? This comment is still appropriate.

phweyland:
> Do you mean `// make sure any edition field loses the focus`? This comment is still appropriate.

Yes, it is not false. But the original code was already doing that, wasn't it ?
Your change directs keys to this specific window widget (and the scrolling bars).
BTW your change is twofold. set_focus -> grab_focus and dt_ui_main_window(darktable.gui->ui) -> dt_ui_center(darktable.gui->ui). 
I think that is the second one which solves #5535. I'm unsure about the first one, but that should be easy to verify.

elstoc:
Possibly so, though there are quite a few that do the same thing and my understanding (based on the docs) is that set_focus is only available for the window as a whole (gtk_window_set_focus) and not for individual widgets. Indeed the gtk documentation for gtk_window_set_focus states "To set the focus to a particular widget in the toplevel, it is usually more convenient to use gtk_widget_grab_focus() instead of this function" suggesting they are equivalent in functionality

phweyland:
Great. Thanks for the clarification. My initial comment seems to me still valid, but obviously it's an infinitesimal detail.

PR #5537, opened 2020-06-18
assign focus to center view when clicking lib header

elstoc:
ensures that any gtkentry field loses focus and allows keyboard
shortcuts that depend on the view's key_pressed handler function
correctly (e.g. the navigation shortcuts in the lighttable)


elstoc:
Similar to the fix made in the darkroom view to make the module search lose focus. This resolves the issue, for example, when you're editing a metadata field and then collapse the metadata module, that the arrow keyboard shortcuts to navigate in the lighttable don't work (unless you first click an image).

aurelienpierre:
Thanks !

PR #5539, opened 2020-06-18
Updates to pt_BR

jpellegrini:


aurelienpierre:
Please rebase your branch on master. Something like `git rebase upstream/master` should do. That will remove all the merging commits that we don't want in master's history.

jpellegrini:
> Please rebase your branch on master. Something like `git rebase upstream/master` should do. That will remove all the merging commits that we don't want in master's history.

I'm sorry! Fixed it now.

aurelienpierre:
You don't need to merge branches like that at all. You can only use `git fetch upstream master` then rebase.

jpellegrini:
> You don't need to merge branches like that at all. You can only use `git fetch upstream master` then rebase.

That's what I always do (I have a remote branch set to the darktable main repo), except I do pull (which does merge) instead of fetch. But it makes sense to only fetch -- I'll start doing that.

(I remember when git and other distributed version control systems were in their infancy. git was absolutely not the one I liked most, because its usage was too confusing. But anyway, it became the standard...)

aurelienpierre:
You still have one merging commit ;-)

TurboGit:
I'll squash and merge from there.

PR #5544, opened 2020-06-18
attempt at making tone EQ more robust and responsive

### LINKED ISSUE #3714: FIXME: message for tone equalizer
playing around with tone equalizer and profiled denoise resulted in following message after zoom in to 100%:
FIXME: module `toneequal' changed the roi from 3342 x 3012 @ 2727 / 2185 to 1624 x 1294 | 3586 / 3044 but doesn't have distort_mask() implemented!

**To Reproduce**
Steps to reproduce the behavior - i'm trying fo find minimum reproducible scenario.

**Screenshots**

 - OS: this is darktable 3.0.0rc2+76~gf6026a1fc-dirty

compile options:
  bit depth is 64 bit
  normal build
  OpenMP support enabled
  OpenCL support enabled
  Lua support enabled, API version 5.0.2
  Colord support disabled
  gPhoto2 support enabled
  OpenEXR support enabled

### LINKED ISSUE #5145: [Tone Equalizer] crash after ad path mask misbehavior in soften
darktable 3.1.0+1703~g989d68d45

This backtrace was generated when I was applying a mask using on the soften module using "add path".
As I was dropping points for the mask at one point (see image) the image which was fit to screen started zooming in, its as if the mask was trying to help to put points on the image by zooming and moving the image around.

Image that automatically zoomed in and moved around making it difficult to put points.
image

Then DT crashed.

this is darktable 3.1.0+1703~g989d68d45 reporting an exception:

-------------------

Error occurred on Thursday, May 21, 2020 at 09:08:46.
```

darktable.exe caused an Access Violation at location 0000000065F45539 in module libtoneequal.dll Reading from location 0000000060A01000.

AddrPC           Params
0000000065F45539 000000000645FBF0 0000000064942F28 0000000000000000  libtoneequal.dll!apply_exposure._omp_fn.0  [F:/msys64/home/chill/darktable/src/iop/toneequal.c @ 807]
   805:       for(size_t c = 0; c < ch; ++c)
   806:         out[(i * out_width + j) * ch + c] = (c == 3) ? in[((i + offset_y) * in_width + (j + offset_x)) * ch + 3]  // pass-through alpha mask
>  807:                                                      : in[((i + offset_y) * in_width + (j + offset_x)) * ch + c] *
   808:                                                        correction[(i + offset_y) * in_width + (j + offset_x)];
   809: }
000000000019EAA0 00000000284A6D40 0000000000000000 0000000000000000  libgomp-1.dll!omp_in_final
0000000064945092 000000002522E0E0 0000000000000000 00000000284A6D40  libwinpthread-1.dll!pthread_create_wrapper
00007FFC60B3B04A 00007FFC60B906D0 00000000284A6D40 0000000000000000  msvcrt.dll!_beginthreadex
00007FFC60B3B11C 0000000000000000 0000000000000000 0000000000000000  msvcrt.dll!_endthreadex
00007FFC61F47BD4 0000000000000000 0000000000000000 0000000000000000  KERNEL32.DLL!BaseThreadInitThunk
00007FFC629ACE51 0000000000000000 0000000000000000 0000000000000000  ntdll.dll!RtlUserThreadStart
```

On restarting this is what it opened up into, soften module.
You can see the history on the side if its helpful

image

### LINKED ISSUE #5216: Can't zoom to 100% when Tone Equalizer is on - Mac OS X
When Tone Equalizer is on, I can't zoom to 100%, darktable just keeps trying to load the image. Sometimes it works, but only after an abnormally long time. I'm using Mac OS X, with darktable 3.1.0+1266˜g737735db9.

### LINKED ISSUE #6605: Tone Equalizer with eigf causes weird artifacts when zoomed in and sometimes crashes when zooming out
**Describe the bug**

When using the new eigf masking option of the tone equalizer, everything works fine when completely zoomed out. While zooming in (seems to be at zoom >30%) the image looks completely streaky, like one giant artifact, however with retained colors (see screenshot). Zooming back out or deactivating tone equalizer while zoomed in reverts to the normal image, however sometimes (rarely) darktable segfaults upon zooming out or deactivating tone equalizer.
Switching to the mask view in tone equalizer while being zoomed into the picture shows a normal mask.

Exporting the file appears to work fine, the exported JPG looks normal.

**To Reproduce**
1. Open fresh file
2. activate tone equalizer with eigf (I do not even have to change any settings)
3. Zoom into the image (appears to start if zoomed to more than 30%)
4. Image is one big artifact
5. zooming back out or deactivating tone equalizer sometimes (not always) segfaults with error message:
```
munmap_chunk(): invalid pointer
[1]    161243 abort (core dumped)  darktable
```
6. Switching on the mask view while zoomed in shows the proper mask

**Expected behavior**
No artifacts and no crashes...

**Screenshots**
Zoomed out:
Zoomed out
Zoomed in:
Zoomed in with artifacts

 - Darktable Version: 3.3.0~git1181.7cc1296fd
 - OS: Pop.OS 20.04 (Kernel 5.4.0-7642-generic)
 - OpenCL activated or not, makes no difference
 - Intel i5-6200U, Intel HD Graphics 520

**Additional context**
I tried this on multiple olympus raw files, it always happened.
Also happens on JPG files

aurelienpierre:
This tries to fix #5216 fix #5145 and fix #3714 and fix #6605, e.g. weird segfaults and major slowdowns when zooming in picture while Tone EQ is used.

## How was it before ?

When zoomed-in, the module enlarged the region of interest (ROI) to compute the guided filter. This ensured that, no matter the zoom size, the guided filter will *look* the same in the darkroom, especially around preview edges. At export time, it did not change anything. But that came at a price : the computed image could be close to the full resolution image, when zoomed at 100% with a blur size of 50%, resulting in slow-downs.

## How is it now ?

Only the visible ROI is computed. Padding is removed. The local contrast will slightly change depending on zoom scale and on whether the image details are on the preview ROI edge or center. At export time, it will make no difference.

## Your job

is to tell me whether the preview inaccuracy is worth the speed-up and robustness increase.

Remember that, now as before, setting the global contrast (no matter which module is used) should be made fully zoomed-out (image filling the view or less) to avoid perceptual discrepancies linked to Stevens/Bartleson-Breneman/Hunt effects and central fovea sensitivity while assessing the image contrast and brightness.

### DIFF
--- a/src/iop/toneequal.c
+++ b/src/iop/toneequal.c
@@ -872,7 +872,7 @@ schedule(static) aligned(luminance, out, in:64) collapse(3)
     for(size_t j = 0; j < out_width; ++j)
       for(size_t c = 0; c < ch; ++c)
       {
-        // normalize the mask intensity between -8 EV and 0 EV for clarity, 
+        // normalize the mask intensity between -8 EV and 0 EV for clarity,
         // and add a "gamma" 2.0 for better legibility in shadows
         const float intensity = sqrtf(fminf(fmaxf(luminance[(i + offset_y) * in_width  + (j + offset_x)] - 0.00390625f, 0.f) / 0.99609375f, 1.f));
         out[(i * out_width + j) * ch + c] = (c == 3) ? in[((i + offset_y) * in_width + (j + offset_x)) * ch + 3]
@@ -1096,6 +1096,7 @@ void modify_roi_in(struct dt_iop_module_t *self, struct dt_dev_pixelpipe_iop_t *
   const int radius = (int)((diameter - 1.0f) / ( 2.0f));
   d->radius = radius;
 
+  /*
   // Enlarge the preview roi with padding if needed
   if(self->dev->gui_attached && sanity_check(self))
   {
@@ -1110,6 +1111,7 @@ void modify_roi_in(struct dt_iop_module_t *self, struct dt_dev_pixelpipe_iop_t *
     roi_in->width = roir - roi_in->x;
     roi_in->height = roib - roi_in->y;
   }
+  */
 }
 
 


TurboGit:
Some comments, the speed is amazingly faster. I did not measured it, but at 100% zoom it feels 2x or 3x faster. I never had a crash so I cannot comment. The preview accuracy does not seems important to me, it probably depends on the mask.

Do you have any guidance to reproduce the most inaccurate mask ?

aurelienpierre:
It's really difficult to spot. You need an image with a bright spot in the middle of a dark area. Set the tone EQ to darken bright parts and brighten dark parts. Then zoom in/out and pan, and compare if the brightness is consistent.

Example:
Capture d’écran du 2020-06-18 16-49-38
Capture d’écran du 2020-06-18 16-48-56
Capture d’écran du 2020-06-18 16-48-21

The over-exposed snow spot on the bottom edge is not consistent between scales, but the background holds.

rawfiner:
I did not test yet, but looking at the code it seems that you forgot to update the tooltip text to remove the warning about the possible slowdowns when used with denoising ;-)
Anyway, this will make dt more responsive, actually waaay more responsive when using denoising so I think it is a good thing :-)

aurelienpierre:
Yeah I don't care about tooltips for now X-)

TurboGit:
I suppose I should add a tone equalizer integration test to ensure all is good with this change.

MStraeten:
cudo to this: in 100% view processing speed is about 4 to 5 times faster ...

aurelienpierre:
Don't get excited about this. It's basically harming accuracy and introducing more user-misunderstandings about scale-variance of preview, by computing only the visible part of the image.

MStraeten:
after doing some retouch stuff in 100% view with active denoise profiled an toneequal (i wasnt patient enough to do this without the latest performance tuning stuff) found it quite reasonable. Maybe this can be controlled by a preset setting: Speed at a price of accuracy or accuracy at a price on responsiveness. Unfortunately there's no free lunch around ...

ralfbrown:
I'd say this option is a good candidate for fastpipe mode.

I've been thinking about how to maximize UI responsiveness, and one of the big things I've come up with is having fastpipe mode not be a matter of which iop is open, but of whether the user is actively changing module settings.  For example, engage fastpipe mode whenever another setting change occurs before the pixelpipe run initiated by the previous one has completed (or within 250-500ms even if otherwise idle?).  Once in fastpipe mode, leave it and issue another regular run when there have not been any further setting changes for 250-500ms, or 100ms after the last pixpipe run ends (whichever is later).  Users won't be able to notice minor discrepancies introduced by fastpipe mode if the image changes again in a fraction of a second.

aurelienpierre:
hmmm let's not get carried away with fastpipe and keep things stupid simple. Introducing special cases is bad. Pipeline should be as generic as possible. The reason is, beyond responsiveness, image processing algos rely on assumptions on their input you need to ensure to guaranty the validity of the output. If you start flip-flopping things around like a crazy computer scientist, you may void important assumptions without knowing it and turn your software into a non-deterministic nightmare.

Remember we already have problems dealing with scale-consistency of image output for many local filters. Let's not harm that even more.

aurelienpierre:
> I suppose I should add a tone equalizer integration test to ensure all is good with this change.

This change will not affect exports anyway. Padding happened only if `self->dev->gui_attached == TRUE`. All the rest is coordinates shifting to account for the padded input while writing the smaller output.

aurelienpierre:
Side note : with EIGF in tone EQ, now typical blur sizes are between 1-5%, so the padding is less important.

PR #5546, opened 2020-06-18
modulegroups: ESC key cancels search

elstoc:
while in the module search box, pressing the ESC key clears the search
text and gives focus back to the centre image


aurelienpierre:
Oh yeah, thanks a lot. This one was making me scream everytime.

PR #5548, opened 2020-06-18
Align always-on icon

### LINKED ISSUE #5543: always on button size, again
I thought that after #5445 and cc0bc1d the alignment of icons/labels in module headers was solved, but apparently it's not

icons

no_icons
 
current master, Windows 10

Nilvus:
fix #5543 

Was just a forgotten update since last cairo rewrite of this icon.

### DIFF
--- a/data/themes/darktable-icons.css
+++ b/data/themes/darktable-icons.css
@@ -100,6 +100,7 @@
     background-size: contain;
     background-repeat: no-repeat;
     margin-top: 0.14em;
+    margin-left: 0.30em;
     min-height: 1.1em;
     min-width: 1.1em;
 }
--- a/data/themes/darktable.css
+++ b/data/themes/darktable.css
@@ -548,6 +548,7 @@ overshoot.right
 #module-instance-button,
 #module-enable-button,
 #module-collapse-button,
+#module-always-enabled-button,
 #module-always-disabled-button
 {
   /* background-color: transparent; this doesn't work */
@@ -556,12 +557,6 @@ overshoot.right
   padding: 0.06em;
   border: 0;
   margin: 0 1px;
-}
-
-#module-always-enabled-button /* button on always activated module */
-{
-  color: @plugin_label_color;
-  border: 0;
 }
 
 #module-header #button-canvas
--- a/data/themes/darktable.css
+++ b/data/themes/darktable.css
@@ -605,7 +605,7 @@ overshoot.right
   background-color: @bg_color;
   color: @plugin_label_color;
   border:0;
-  padding: 0 2px 2px 0.5em;
+  padding: 0 2px 2px 0.45em;
   font-weight: normal;
   font-family: sans-serif;
   font-size: 1.1em;


Nilvus:
Just see that would be good to also have same space between state icon/module icon than between module icon/label text.

aurelienpierre:
Maybe a bit more space at the left of icons ?
Capture d’écran du 2020-06-18 20-09-54

Nilvus:
> Maybe a bit more space at the left of icons ?

Yes, that was just my comment at nearly same time as yours. I see that you will also need to add new channel reboot mixer on darktable-icons.css file to add at least default round icon.

Nilvus:
@Mark-64: could you test this on Windows and HiDPI.

For all, I test that with 2 different font size to be sure space remains nearly the same. Font-size used : 11 px and 16px.

Nilvus:
So ready for review now.

TurboGit:
I never use the icons theme, for testing this PR I had activated it. The vertical alignment is fixed, but horizontal alignment seems broken (was also before, not a regression):

image

Maybe fixing this as part of the PR would be good!

TurboGit:
I don't have an HiDPI sceen, maybe that explains the difference with Aurélien's screenshot ?

Nilvus:
@TurboGit: I don't have that issue on my side. Be sure I tested horizontal and vertical alignment carefully yesterday by many screenshots taken and zoomed precisely on Gimp to check, and all is good. So probably something else on your side.

The only thing remaining on my side is that some module icons are just 1 pixel less in height than always on. But space is in the top not in bottom like your screenshots. And it's just because some svg icons do not have exactly same size. Some modules icons are at the opposite, 2 or 3 pixels more in height (like contrast equalizer or white balance), so adjusting that pixel space on top will make higher icons more icons. But I only show that when zooming screenshot on Gimp. When I just see darktable view, this pixel space on top of the icons is hard to see. As you can see what I have:

Nilvus:
> I don't have an HiDPI sceen, maybe that explains the difference with Aurélien's screenshot ?

As I don't have HiDPI screen too, I don't think that explains the difference.

Nilvus:
I just understand why you not misalignement. You have a small font size. I've just tested with 10pt font size and I see your alignment issue. I'm not sure it will be possible to fix with CSS as it works great if you use 11pt or 12pt or even 16pt (that was tests I use).

I never use those icons and I hate them...

TurboGit:
Ok, as I don't use icons either I'll merge this and if a fix is possible will do that in another PR. Thanks for the investigation.

TurboGit:
Oh ! That was the last scheduled for 3.2 :) The milestone is now 100% achieved. Of course will certainly have some more fix under the final release !

Nilvus:
Starting from 10.8 pt on preferences window font size make all icons correctly aligned on my side...

I've tested to change margin-top in darktable-icons.css. When increase it a little (0.17 or 0.18em), alignment is good on 10 pt but became broken on higher sizes. For me, that confirm that CSS would not be the good way (or not alone) to be sure alignment is preserved regarding font-size. Or would need to change em to px but that would cause here other issues (especially I think between HiDPI and LoDPI screens.

So I agree with you: better to merge that and we will see if one day someone complain about that alignment (or if someone want to spent time and rework Gtk/cairo code.

PR #5531, opened 2020-06-18
xmp update fix

phweyland:
- clean up metadata and tags before updating from xmp (triggered by crawler)
- read dt colors if xmp.xmp.label (from external app) not present instead of overwriting them
- some signal raise moved from common to libs (dead locks in _exif_decode_xmp_data)
- ~~some memory leak fixes (g_list_free) in tagging.c~~ fix lists with grouped images instead

fixes #5486

PR #5549, opened 2020-06-18
dt_view_get_images_to_act_on() returns a const list

phweyland:
force to consider dt_view_get_images_to_act_on() returns a const list

EDIT : As discussed in #5531. One of these PR will need to solve rebase conflicts.


TurboGit:
Thanks! Will avoid any more issue with g_free_list()!

PR #5556, opened 2020-06-19
refactoring: eliminate duplicate boilerplate in iops

ralfbrown:
Replace multiple occurrences of boilerplate to create a delayed call to dt_dev_add_history_item (to avoid flooding the pixelpipes with work) with calls to a new function encapsulating that boilerplate.

This is a companion PR to #5530, but the two can be merged independently.

### DIFF
--- a/src/develop/imageop.c
+++ b/src/develop/imageop.c
@@ -2746,6 +2746,38 @@ int dt_iop_count_instances(dt_iop_module_so_t *module)
   return inst_count;
 }
 
+static gboolean _postponed_history_update(gpointer data)
+{
+  dt_iop_module_t *self = (dt_iop_module_t*)data;
+  dt_dev_add_history_item(darktable.develop, self, TRUE);
+  self->timeout_handle = 0;
+  return FALSE; //cancel the timer
+}
+
+/** queue a delayed call of the add_history function after user interaction, to capture parameter updates (but not */
+/** too often). */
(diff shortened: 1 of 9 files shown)


src/develop/imageop.c
```diff
@@ -2746,6 +2746,48 @@ int dt_iop_count_instances(dt_iop_module_so_t *module)
...
+      // let the existing event happen as scheduled
+      return;
+    }
+  }
+  // adaptively set the timeout to 150% of the average time the past several pixelpipe runs took, clamped
+  //   to keep updates from appearing to be too sluggish (though early iops such as rawdenoise may have
+  //   multiple very slow iops following them, leading to >1000ms processing times)
+  const int delay = CLAMP(darktable.develop->average_delay * 3 / 2, 10, 1200);
+  module->timeout_handle = g_timeout_add(delay, _postponed_history_update, module);
+}
```
rabauke, line 2780: This function is never called with `extend_prior == TRUE`. Thus, this function can be simplified, removing the 2nd argument . Even if one considers the 2nd argument as necessary, this function should be rewritten for clarity. In particular the early `return` in the `else` branch obfuscates what goes on here.

ralfbrown: Changed.  The `extend_prior` parameter is for future-proofing, if some other iop wants to use this mechanism that way.

PR #5562, opened 2020-06-19
cleanup unnecessary libsoup stuff

### LINKED ISSUE #5541: doc/README.md: Invalid description of libsoup dependency
**Describe the bug**

_doc/README.md_ says Darktable needs libsoup in order to provide geotagging support. However, comments in _src/CMakeFiles.txt_ suggest libsoup is only required by the Facebook exporter - and geotagging seems to work just fine without it.

**To Reproduce**

1. Compare the contents of _doc/README.md_ with those of _src/CMakeFiles.txt_, and/or
2. Build Darktable with osm-gps-map but without libsoup and try using geotagging.

johnny-bit:
cleans up wording and building stuff with libsoup. since we don't use http_server after removal of facebook export, there's no need to have anything with libsoup in code. bringing i back will be just
pulling from history

on the other hand libsoup is direct dependency of osm-gps-map but in that case we either have osm-gps-map or not. no need to depend on libsoup directly

fixes #5541


houz:
I'd prefer if the http server was kept, it helps with OAuth madness. We might need it again some day.

johnny-bit:
Actually @houz suggests to keep http_server around. will do just PR for wording fixup.

PR #5566, opened 2020-06-19
overlay : show actual thumb size in menu

### LINKED ISSUE #5565: hard to select the right group size in lighttable

I found it hard to find the proper group sizes for the thumbs in the lighttable.

In the following picture we have the actual size (here size 1):

image

But we do not know where we are in the actual group.

What I'd like is to display in parentheses after the size (above 1) the actual thumb side on the lighttable. From there it is possible to see the sizes where it makes sense to have only information when flying over or permanent and actually setup the groups in the preference.

Today the only solution is to do that by trial errors and is cumbersome.

AlicVB:
this fix #5565

### DIFF
--- a/src/libs/tools/global_toolbox.c
+++ b/src/libs/tools/global_toolbox.c
@@ -218,8 +218,9 @@ static void _overlays_show_popup(dt_lib_module_t *self)
   if(thumbs_state)
   {
     // we write the label with the size categorie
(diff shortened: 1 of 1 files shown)


TurboGit:
Great! Working for me and I have been able to find the proper groups size that works best for me more easily with this. Thanks for the quick implementation @AlicVB .

Issue #5558, opened 2020-06-19
Crop & Rotate: Bottom border moves when resizing


elstoc:
When resizing a crop by dragging on any of the top handles (top, top-left, top-right) the bottom border of the image also moves up fractionally. This doesn't happen when dragging on the left, right, bottom, bottom-left or bottom-right handles. Edit: this only happens when I have an aspect ratio set (not when using freehand).

Also, perhaps related (?), very occasionally the crop and rotate module seems to get into a state where merely moving the mouse over the box (without holding any buttons down) causes the crop box to slowly move upwards. The set of circumstances required to reproduce this are seemingly random though.

3.1.0+2325~gd10cb25c3 on ArchLinux with and without openCL

elstoc:
Magic crop-box movement when moving the mouse is illustrated in the attached video. It seems that once I've found the right crop size it happens consistently. Could it be some sort of rounding error?

[issue5558.zip](https://github.com/darktable-org/darktable/files/4805173/issue5558.zip)

AxelG-DE:
I reproduce, see Screencast, attached
[dt_spooky-Magic_20200619-2020-06-19_17.00.37.mp4.zip](https://github.com/darktable-org/darktable/files/4805260/dt_spooky-Magic_20200619-2020-06-19_17.00.37.mp4.zip)

AxelG-DE:
I bisected the bottom-line move up thingie...
```
aac258ae28db0b8b9bc14a29d206186b7133be96 is the first bad commit
commit aac258ae28db0b8b9bc14a29d206186b7133be96
Author: GrahamByrnes <58031430+GrahamByrnes@users.noreply.github.com>
Date:   Tue Jun 9 22:33:39 2020 +0200

    Allows more precise aspect ratios in crop & rotate

 src/iop/clipping.c | 33 +++++++++++++++++++++++++--------
 1 file changed, 25 insertions(+), 8 deletions(-)
```

@GrahamByrnes could you do us a favour to also look into this? :)

GrahamByrnes:
Ooh, look at that... lol. Ok, will go have a look.
Can I just say in advance "What a fcckn nightmare" to save me saying it later?
Ta.

elstoc:
[...]

PR #5569, opened 2020-06-19
Fix the moving lower border in clipping.c

GrahamByrnes:
See #5558.

I'm not sure if it fixes the freaky box-that-crawls effect, but it does seem to fix the lower border that moves.


elstoc:
Looks like that's resolved both issues for me. Thanks @GrahamByrnes.

TurboGit:
Indeed, never saw this issue as I never resize by dragging the side. I only use the corners. Good to see this fixed for the next major release.

AxelG-DE:
@TurboGit it was also visible when you dragged the corners, but really strong only if you go forth and back. one time you just thought "ooops I did a mistake" :)

@GrahamByrnes solves the issue with bottom-move up as well, the magic one, was not reliably reproducible. Thanks for fast reaction

PR #5570, opened 2020-06-19
shortcut to change "keyboard shortcut slider adjustment precision"

elstoc:
Allows user to choose to use normal, fine or coarse adjustments when
amending iop sliders with an increase, decrease or dynamic keyboard
shortcut. Scale can be selected via, of course, a keyboard shortcut.

Uses the same slider adjustment factors that are used for the CTRL and
SHIFT modifiers when adjusting sliders directly.

### DIFF
--- a/data/darktableconfig.xml.in
+++ b/data/darktableconfig.xml.in
@@ -35,6 +35,12 @@
     <longdescription>after applying the above rules, apply the shortcut based on its position in the pixelpipe</longdescription>
   </dtconfig>
   <dtconfig>
+    <name>accel/slider_precision</name>
+    <type>int</type>
+    <default>0</default>
+    <shortdescription>keyboard shortcut slider precision</shortdescription>
+  </dtconfig>
+  <dtconfig>
     <name>bauhaus/scale</name>
     <type>float</type>
     <default>1.4</default>
--- a/src/gui/accelerators.c
+++ b/src/gui/accelerators.c
@@ -669,6 +669,22 @@ void dt_accel_widget_toast(GtkWidget *widget)
 
 }
 
+float dt_accel_get_slider_scale_multiplier()
+{
+  const int slider_precision = dt_conf_get_int("accel/slider_precision");
+  
+  if(slider_precision == DT_IOP_PRECISION_COARSE)
+  {
+    return dt_conf_get_float("darkroom/ui/scale_rough_step_multiplier");
+  }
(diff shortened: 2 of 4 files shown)


elstoc:
oops I forgot the increase/decrease shortcuts!

dterrahe:
> oops I forgot the increase/decrease shortcuts!

Was just about to point that out! ;-)

elstoc:
All done now.

TurboGit:
Seems safe, I put that in 3.2 code looks good. I need to test it.

dterrahe:
Maybe this calls for a little refactoring of dt_bauhaus_slider_add_delta_internal so that it can be used here (or maybe in dt_bauhaus_slider_get_step) so any future users (sorry to mention midi again) would also automatically get them and any future enhancements would only need to be done in one place?

Release technically I was more worried about the additional translations needed. But I suppose 3.2 still hasn't been cut yet...

elstoc:
> Maybe this calls for a little refactoring of dt_bauhaus_slider_add_delta_internal so that it can be used here (or maybe in dt_bauhaus_slider_get_step)

I agree, but perhaps that _would_ be a little more risky for 3.2

elstoc:
I was struggling to find a short description that would still be accurate. I wonder if it would be better as "keyboard shortcut slider precision" rather than "keyboard shortcut slider scale factor"?

elstoc:
and perhaps coarse/fine/normal rather than rough?

elstoc:
Changed as above to call it slider precision with options normal/fine/coarse

TurboGit:
Working for me and new wording is better. Thanks.

dterrahe:
I think this will be a much appreciated addition!

Definitely for 3.4; other ways we could make similar functionality accessible:
- when a slider is selected (by clicking on it), cursor keys (left/right/up/down) all move the slider by step size. We could differentiate with up/down doing normal step, left/right doing small/fine step and page up/down doing coarse step. 
- technically not too complicated, the same keys could work with dynamic shortcuts. So if E assigned to dynamically change exposure, we could allow E+Up/E+Down to move by a step, E+Right/E+Left by a fine step and E+PgUp/E+PgDown by a course step, up and down.

But we are really getting in #5063 territory so if i would require a very intrusive effort/hack to implement all this, then I'd rather address it as part of a larger overhaul.

PR #5573, opened 2020-06-19
preferences: change DPI tooltip

elstoc:
Minor change to reflect the fact that the GUI controls and text DPI
preference is an integer and not a float


TurboGit:
Thanks!

Issue #5489, opened 2020-06-16
Segfault with current master on opening/revisiting a previous edit


codingdave:
**Describe the bug**
Darktable crashes with a previously edited image. I have isolated the image and when I try to import that image + xmp into a fresh darktable environment the crash is reproducible.

**To Reproduce**
1. Import the provided NEF + XMP
2. You will get a Segfault

**Expected behavior**
No Segfault, opening the raw like every other raw file.

**Platform (please complete the following information):**
- c167ee8dc9752e70bd42caf24f9f5f3d5901646b
 - OS: Manjaro
 - OpenGL: Yes
 - nvidia-440x

https://drive.google.com/drive/folders/1q46UEVCoGewbWSL1RPf9uPs3ljHIla8p?usp=sharing

codingdave:
dt 3.02 (Windows) opens the file without error.

TurboGit:
@codingdave : well this image has been opened with a dev version of darktable as there is the hash values in the XMP file.

```
   darktable:history_basic_hash="9d9db76e8845ca28f95dcc0fdc4e62e8"
   darktable:history_auto_hash="16ddf8b7620566315c8735f78ccf069f"
   darktable:history_current_hash="16ddf8b7620566315c8735f78ccf069f">
```

So at this stage hard to tell what is wrong. Possible a picture opened with a buggy dt ?

I'm not sure when it has been opened with 3.0.2 but sure it that is the case the xmp may be corrupted.

MStraeten:
seems to be similar to https://github.com/darktable-org/darktable/issues/5351.  Maybe a full cleanup oft the build environment can help (it did in my case ...)

phweyland:
not sure that can help. unning with debug I get this:

```
Thread 48 "worker res 0" received signal SIGSEGV, Segmentation fault.
[Switching to Thread 0x7fff5aa8d700 (LWP 23100)]
process_sse2_cmatrix_fastpath_simple._omp_fn.7 ()
    at /home/philippe/git/dt/src/iop/colorin.c:1161
1161        __m128 input = _mm_load_ps(in);
```

johnny-bit:
the mm_load_ps is for loading input pixels into sse buffers so... alignment issues? similar to #3209?

phweyland:
an additional detail. when I've opened the image with xmp the first time I got the seg fault but the image was imported. then, even removing the xmp file, dt still crashes (same log) opening this image.

codingdave:
[...]

PR #5582, opened 2020-06-20
iop_order: sanity check to avoid crash when reloading boggus xmp.

TurboGit:
For #5489.


jenshannoschwalm:
Looks just perfect for me.

TurboGit:
Thanks! Merging. We could certainly add more checks there if needed later.

Issue #5581, opened 2020-06-20
bauhaus sliders: fine precision adjustments sometimes do nothing


elstoc:
On some sliders, the 'fine' precision adjustment permitted by CTRL+scroll sometimes does nothing meaning that fine scale adjustment of a slider is not always possible.

For example, on filmic rgb, the 'add noise in highlights' slider has a default step of 0.05 and the fine-scale multiplier is 0.1 * default. This appears to _sometimes_ be insufficient for the slider to move at all meaning that 0.05 is now the smallest scale on which that slider may be changed. The slider will sometimes move and sometimes not depending on its starting position.

I have seen this on other sliders as well. For example, the threshold slider on filmic rgb exhibits the same behaviour.

There probably needs to be something that checks the minimum possible step and always moves the slider by at least that much. This will probably have to be replicated in ``dt_bauhaus_slider_add_delta_internal`` and the changes made in #5570. Perhaps this might be the opportunity to refactor the code as suggested by @dterrahe.

dterrahe:
As per https://github.com/darktable-org/darktable/pull/5506#issuecomment-646991085

The number of digits set for a slider is used for rounding and to set the format, but the latter can be set to something else afterwards, for example  to add a unit or %. If both are in sync, for example 2 and "0.01" (%.2f), it means that any change smaller than the accuracy shown is ignored, because rounded away. So if the step size is not at least 10x the smallest digit, you can't actually do a smaller/fine step. But if digits is set to 3 while the format is "0.01" then the 3rd digit still exists but is hidden, so now you can slowly scroll until the change becomes visible.

Of course the multiplication factor used for "fine" adjustments is user configurable, so the relationship required between digits and format is not fixed.

One answer is to just do away with the rounding. The only case where it would be required is where only whole numbers are allowed (digits=0) for example when the slider is linked to an integer. Maybe we can forego the rounding in all other cases which would solve this problem. But I'm sure some people won't like that when their slider shows "0.67" it could really be anything between "0.665" and "0.675" :-) Can't make everyone happy.

junkyardsparkle:
> But I'm sure some people won't like that when their slider shows "0.67" it could really be anything between "0.665" and "0.675" :-) Can't make everyone happy.

It seems like this has already been the case in some places IIRC. I personally wouldn't mind seeing the displayed precision increased to better expose actual values (this seems more "bauhaus" to me), but maybe nobody else feels that way...

PR #5583, opened 2020-06-20
Don't round slider to digits unless digits is zero

dterrahe:
Just to test and experiment with the approach suggested in https://github.com/darktable-org/darktable/issues/5581#issuecomment-647029190

This might cause problems if any module is expecting their parameters to be rounded to 0.01, 0.02 etc.

Also fixes an issue where first slider in denoise (non-local means) is stuck because step size too small...


elstoc:
It works (at least it prevents slider values becoming stuck) though does feel a bit odd, certainly when using the keys, that it sometimes takes a few key presses for anything to change.

PR #5587, opened 2020-06-21
darktableconfig.xml.in: fix XML and include ui/cairo_filter

dtorop:
Update darktableconfig.xml.in to include an entry for `ui/cairo_filter`. Also, convert ">"'s to entities to make valid XML.

There's still a slew of other conf variables which either aren't in darktableconfig.xml.in or which don't have descriptions.


TurboGit:
Thanks !

PR #5588, opened 2020-06-21
implement fastpipe mode for grain iop

ralfbrown:
The grain iop normally runs a downsampling filter on zoomed-out views, which makes it ~20 times
slower due to averaging 21 points.  This patch makes it bypass the filter in fastpipe mode.

There appears to be a bug in the downsampling filter -- toggling the perspective correction iop open
and closed shows that the only difference in the added grain is a diagonal shift of a few pixels, as
if the grain were printed on a transparent sheet of plastic which gets shifted slightly between
normal and fastpipe modes.


TurboGit:
Very safe to me. Thanks!

PR #5591, opened 2020-06-21
Implement modifier detection in a portable way

### LINKED ISSUE #5443: Auto + or Neg Selections in Color Zones and RGB Tone Curve dont seem to be working...
Expected behaviour...ctrl or shift click on the second picker (range picker) will auto generate a curve in the positive or negative direction based on the selection or initially the whole image....doesn't seem to be working ...not in Color Zones or RGB Tone curve...

It appears to be selecting the range or hue or saturation value but it is not producing the curve based on that range and ctrl or shift click selection...

Running Windows 10 2004
DT version   3.1.0+2031~g4a1331495
-->
image
image

dterrahe:
Detecting if modifiers (CTRL/SHIFT) were pressed without having access to an event did not work under Windows. Now implemented in a portable way, that needs to be tested on Mac.

This impacts color pickers in color zones and rgb curve and in parametric blending mask.

Closes #5443

### DIFF
--- a/src/gui/color_picker_proxy.c
+++ b/src/gui/color_picker_proxy.c
@@ -157,10 +157,8 @@ static gboolean _iop_color_picker_callback_button_press(GtkWidget *button, GdkEv
   // set module active if not yet the case
   if(module->off) gtk_toggle_button_set_active(GTK_TOGGLE_BUTTON(module->off), TRUE);
 
-  const uint32_t state = e != NULL
-                        ? e->state
-                        : gdk_keymap_get_modifier_state(gdk_keymap_get_for_display(gdk_display_get_default()));
-  gboolean ctrl_key_pressed = (state & gtk_accelerator_get_default_mod_mask()) == GDK_CONTROL_MASK;
+  const GdkModifierType state = e != NULL ? e->state : dt_key_modifier_state();
+  const gboolean ctrl_key_pressed = (state == GDK_CONTROL_MASK);
   dt_iop_color_picker_kind_t kind = self->kind;
 
   if (module->picker != self || (ctrl_key_pressed && kind == DT_COLOR_PICKER_POINT_AREA))
--- a/src/iop/colorzones.c
+++ b/src/iop/colorzones.c
@@ -2229,12 +2229,11 @@ void color_picker_apply(dt_iop_module_t *self, GtkWidget *picker, dt_dev_pixelpi
       curve[k].y = d->curve[ch_curve][k].y;
     }
 
-    guint state = gdk_keymap_get_modifier_state(gdk_keymap_get_for_display(gdk_display_get_default()));
-    GdkModifierType modifiers = gtk_accelerator_get_default_mod_mask();
-    int picker_set_upper_lower;
-    if((state & modifiers) == GDK_CONTROL_MASK) // flat=0, lower=-1, upper=1
+    const GdkModifierType state = dt_key_modifier_state();
+    int picker_set_upper_lower; // flat=0, lower=-1, upper=1
+    if(state == GDK_CONTROL_MASK)
       picker_set_upper_lower = 1;
-    else if((state & modifiers) == GDK_SHIFT_MASK)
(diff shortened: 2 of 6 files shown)


src/gui/color_picker_proxy.h
```diff
@@ -55,6 +55,8 @@ void dt_iop_color_picker_cleanup();
 /* link color picker to widget */
 GtkWidget *dt_color_picker_new(dt_iop_module_t *module, dt_iop_color_picker_kind_t kind, GtkWidget *w);
 
+guint dt_get_modifier_state();
```
TurboGit, line 58: I suppose this should be better in `gui/accelerators.h`, no?

And maybe add a bit of comment for the spec.

src/iop/colorzones.c
```diff
@@ -2229,12 +2229,11 @@ void color_picker_apply(dt_iop_module_t *self, GtkWidget *picker, dt_dev_pixelpi
       curve[k].y = d->curve[ch_curve][k].y;
     }
 
-    guint state = gdk_keymap_get_modifier_state(gdk_keymap_get_for_display(gdk_display_get_default()));
-    GdkModifierType modifiers = gtk_accelerator_get_default_mod_mask();
-    int picker_set_upper_lower;
-    if((state & modifiers) == GDK_CONTROL_MASK) // flat=0, lower=-1, upper=1
+    guint state = dt_get_modifier_state();
```
TurboGit, line 2232: const

src/iop/rgbcurve.c
```diff
@@ -508,12 +508,11 @@ void color_picker_apply(dt_iop_module_t *self, GtkWidget *picker, dt_dev_pixelpi
       p->curve_nodes[ch][k].y = d->curve_nodes[ch][k].y;
     }
 
-    guint state = gdk_keymap_get_modifier_state(gdk_keymap_get_for_display(gdk_display_get_default()));
-    GdkModifierType modifiers = gtk_accelerator_get_default_mod_mask();
-    int picker_set_upper_lower;
-    if((state & modifiers) == GDK_CONTROL_MASK) // flat=0, lower=-1, upper=1
+    guint state = dt_get_modifier_state();
```
TurboGit, line 511: const

src/gui/color_picker_proxy.c
```diff
@@ -157,10 +165,8 @@ static gboolean _iop_color_picker_callback_button_press(GtkWidget *button, GdkEv
   // set module active if not yet the case
   if(module->off) gtk_toggle_button_set_active(GTK_TOGGLE_BUTTON(module->off), TRUE);
 
-  const uint32_t state = e != NULL
-                        ? e->state
-                        : gdk_keymap_get_modifier_state(gdk_keymap_get_for_display(gdk_display_get_default()));
-  gboolean ctrl_key_pressed = (state & gtk_accelerator_get_default_mod_mask()) == GDK_CONTROL_MASK;
+  const uint32_t state = e != NULL ? e->state : dt_get_modifier_state();
+  gboolean ctrl_key_pressed = (state == GDK_CONTROL_MASK);
```
TurboGit, line 169: looks like it may be set to const?

src/gui/color_picker_proxy.h
dterrahe: Not really; nothing to do with shortcut keys...

Maybe gtk.c? Just because it already is a collection of functions with complicated gtk dependencies.

Just get_modifier_state? key_modifier_state? dt_key_modifier_state?

TurboGit: Ok for gtk.h/c and let's go with `dt_key_modifier_state`. (a public routine should always start with dt_).

dterrahe:
I made the changes you requested and specified the GdkModifierType.

Just wondering; when I'm making changes per your request like here, do you prefer to see them as individual commits or everything squashed into one commit like I've done now, so you can easily merge a clean one commit change?

TurboGit:
Everything squashed is ok by me.

TurboGit:
Tested (see #5443), working... So merging.

PR #5592, opened 2020-06-21
prefer favorite group when showing/hiding a module with a keyboard shortcut

elstoc:
when showing/hiding a module using a keyboard shortcut, if the module doesn't exist in the currently-displayed group but does exist in the favorites group, switch to favorites rather than the module's default group


TurboGit:
@elstoc : I suppose this could go in 3.2 ?

elstoc:
It's pretty low impact. I agree.

TurboGit:
So, merging. Thanks for the quick feedback.

PR #5599, opened 2020-06-21
Fix active image in filmstrip

### LINKED ISSUE #5598: lossing current edited picture in darkrrom when flying over the filmstrip
With the end of the lock-down here in France I had restarted doing some pictures and so edited some pictures instead of testing all PR :) Fact is that this put you in another perspective... And I found an annoying feature, even a regression compared to previous version.

**Describe the bug**
Lossing current edited picture in darkrrom when flying over the filmstrip.

**To Reproduce**
1. Go to darkrroom
2. See that the current edited picture is highlighted in the filmstrip, fine.
3. Move to filmstrip, the highlighted picture is the one under the mouse, fine too.
4. But the currently edited picture is not highlighted.

**Expected behavior**
We really need to see the current edited picture as this is the only way to copy some edit from another picture and pasting into the current edit.

And no, sometimes it is difficult to see the current edit when the filmstrip contains a series of similar pictures (burst shooting or picture taken with different speed or aperture).

How this could be done ?
1. Either always keep current edited picture highlighted
2. Have some specific color in around the thumb
3. Make it highlighted but with a different color
4. Instead of a plain grey background have some stripes on it

I'm leaning toward 3, but maybe a better solution could be found.

 - Darktable Version: master
 - OS: GNU/Debian

Nilvus:
Fix #5598

I also made a simple cleanup to link some same related colors.

@AlicVB and @TurboGit: I think for that, a simple border-top is better than all borders as before. Hope you will like that. In case, I tested border-bottom and it's quite less visible as less space to have good contrast. Good to be tested with and without bottom headerbar (better without but good for me with).

### DIFF
--- a/data/themes/darktable-elegant-dark.css
+++ b/data/themes/darktable-elegant-dark.css
@@ -78,9 +78,9 @@
 /* Views */
 @define-color darkroom_bg_color @grey_60;
 @define-color darkroom_preview_bg_color @grey_50;
-@define-color print_bg_color @grey_60;
-@define-color lighttable_bg_color @print_bg_color;
-@define-color lighttable_preview_bg_color @grey_50;
+@define-color print_bg_color @darkroom_bg_color;
+@define-color lighttable_bg_color @darkroom_bg_color;
+@define-color lighttable_preview_bg_color @darkroom_preview_bg_color;
 @define-color lighttable_bg_font_color @grey_85;
 
 /* Lighttable and film-strip */
--- a/data/themes/darktable-elegant-grey.css
+++ b/data/themes/darktable-elegant-grey.css
@@ -81,9 +81,9 @@
 /* Views */
 @define-color darkroom_bg_color @grey_75;
 @define-color darkroom_preview_bg_color @grey_65;
-@define-color print_bg_color @grey_75;
-@define-color lighttable_bg_color @print_bg_color;
-@define-color lighttable_preview_bg_color @grey_65;
+@define-color print_bg_color @darkroom_bg_color;
+@define-color lighttable_bg_color @darkroom_bg_color;
+@define-color lighttable_preview_bg_color @darkroom_preview_bg_color;
 @define-color lighttable_bg_font_color @grey_95;
 
 /* Lighttable and film-strip */
--- a/data/themes/darktable.css
+++ b/data/themes/darktable.css
@@ -127,9 +127,9 @@
 /* Views */
 @define-color darkroom_bg_color @grey_45;
 @define-color darkroom_preview_bg_color @grey_35;
-@define-color print_bg_color @grey_45;
-@define-color lighttable_bg_color @print_bg_color;
-@define-color lighttable_preview_bg_color @grey_35;
+@define-color print_bg_color @darkroom_bg_color;
+@define-color lighttable_bg_color @darkroom_bg_color;
+@define-color lighttable_preview_bg_color @darkroom_preview_bg_color;
 @define-color lighttable_bg_font_color @grey_75;
 
 /* Lighttable and film-strip */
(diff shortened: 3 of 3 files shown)


TurboGit:
Tested, on my side it is hard to see which thumb is selected. I have a brownish bar, thin and close to the color of the info bar just above:

 
image

TurboGit:
Forgot to say that yes, it is ok without the bottom headerbar.

TurboGit:
I tried with:
`  border-top: 3px solid @thumbnail_hover_fg_color;
`

It looks fine on all 3 themes for me. What do you think ?

TurboGit:
You may also want to test this:

```
/* Set top margin on active image in filmstrip */
#thumb_main:active #thumb_back
{
    border-top: 3px solid @thumbnail_hover_fg_color;
    background:
        repeating-linear-gradient(
            45deg,
            @thumbnail_bg_color,
            @thumbnail_bg_color 5px,
            @thumbnail_selected_bg_color 5px,
            @thumbnail_selected_bg_color 10px
        );
}
```

image

TurboGit:
I find the stripes not too intrusive and makes visually clear what is the current selection.

Nilvus:
> I tried with:
> ` border-top: 3px solid @thumbnail_hover_fg_color;`
> 
> It looks fine on all 3 themes for me. What do you think ?

Do we have same screens (or is it our eyes) ?
I've find hard to read that my proposal is hard to see with bottom headerbar because I see them easily, even in your screenshot without clicking on it (of course less than without that bar) but I've tested before your try above and it's really less visible on my side (with and without), especially in grey theme.

Last thing is that I think it's better to have border-top visible always when selected so also when hovering the image. And thumbnail_hover_fg_color can't be visible in all themes when hovering the image.

Not sure about the stripes that way but that's an idea to go further and test. I will see that better ASAP.

Nilvus:
So, thanks for your ideas.

TurboGit:
> Do we have same screens (or is it our eyes) ?

Maybe my eyes :) Even if I do not have glasses all the time yet :) Also on the bottom of the darkroom I have the filmstrip but with quite small thumbs. I want to filstrip to be able to copy/paste and  move around. My thumbs are probably around 150x150 px.

AlicVB:
So,
- I agree with @TurboGit that the current proposal doesn't well if the bottom toolbar is shown.
- `border-top: 3px solid @thumbnail_hover_fg_color;` is a little better but fail on mouse over...
- The stripes are a good idea, but we can't see the current status of the image : selected, hovered, etc... and hovered overlays are strange...

to resume : I think we can find something better !
I wonder if drawing an arrow like that wouldn't be better...
Capture d’écran_2020-06-22_18-41-09
Note that this will require some code change, but nothing too hard...

Nilvus:
> So,
> 
>     * I agree with @TurboGit that the current proposal doesn't works well if the bottom toolbar is shown.

That's just what I said on my first post, but that was the best compromise I found on my side by now (with what I can with CSS on borders).

>     * `border-top: 3px solid @thumbnail_hover_fg_color;` is a little better but fail on mouse over...

I'm starting to worrying if both of you says that this one is a little better. I have tested it before @border_color and for me it's just worse on grey theme on all states (hover, just active...) and not visible on hover effect in all themes. As my proposal is visible on all states and even hover effect. The only downside is it's just a little less visible when bottom headerbar visible.

>     * The stripes are a good idea, but we can't see the current status of the image : selected, hovered, etc... and hovered overlays are strange...

Exactly, a good idea but we can't confirm for that reason.

> to resume : I think we can find something better !

Of course, but not with CSS only (all borders around is not good too). The only better thing I found with CSS is use yellow color (or some orange-yellow one) on border-top. That was great on all modes. The only issue is that it's too distracting, so I forget that idea.

> I wonder if drawing an arrow like that wouldn't be a solution...
> Capture d’écran_2020-06-22_18-41-09
> Note that this will require some code change, but nothing too hard...

That would just great. And with that arrow, I really think that my proposal with arrow added (seems you use my proposal for color and border size : 2.5px or 3px is a minor difference event if I find 2.5px a little more balanced). I don't know how you could do the arrow but maybe, my actual PR could be merge if using my CSS settings and just add the arrow?

TurboGit:
I agree the little triangle down proposed by @AlicVB  seems the best option at the moment to me.

TurboGit:
Ok, let's go then. Thanks to both of you !

PR #5590, opened 2020-06-21
Blending tabs icons: enable hover effect

Mark-64:
This PR will enable mouse hover effect for three of the icons in blending tabs:

- parametric
- drawn and parametric
- raster

Example
Screenshot from 2020-06-21 08-07-25
Screenshot from 2020-06-21 08-07-26

PR #5603, opened 2020-06-22
drawn mask and add brush icons

Mark-64:
Following discussions in #5590 and #5314, with this PR I try to improve the "drawn mask", "drawn and parametric mask" and "add brush" icons.
Here are the screenshots:
tabs
tools
Everything is totally subjective here, so please be patient and give feedback.

### DIFF
--- a/src/dtgtk/paint.c
+++ b/src/dtgtk/paint.c
@@ -483,7 +483,7 @@ void dtgtk_cairo_paint_masks_eye(cairo_t *cr, gint x, gint y, gint w, gint h, gi
 
 void dtgtk_cairo_paint_masks_circle(cairo_t *cr, gint x, gint y, gint w, gint h, gint flags, void *data)
 {
-  PREAMBLE(1.1, 0, 0)
+  PREAMBLE(1.10, 0, 0)
 
   cairo_arc(cr, 0.5, 0.5, 0.4, 0, 6.2832);
   cairo_stroke(cr);
@@ -493,7 +493,7 @@ void dtgtk_cairo_paint_masks_circle(cairo_t *cr, gint x, gint y, gint w, gint h,
 
 void dtgtk_cairo_paint_masks_ellipse(cairo_t *cr, gint x, gint y, gint w, gint h, gint flags, void *data)
 {
-  PREAMBLE(1.1, 0, -0.025)
+  PREAMBLE(1.15, 0, 0)
 
   cairo_save(cr);
   cairo_translate(cr, 0.1465, 0);
@@ -507,7 +507,7 @@ void dtgtk_cairo_paint_masks_ellipse(cairo_t *cr, gint x, gint y, gint w, gint h
 
 void dtgtk_cairo_paint_masks_gradient(cairo_t *cr, gint x, gint y, gint w, gint h, gint flags, void *data)
 {
-  PREAMBLE(1, 0, 0)
+  PREAMBLE(1, -0.05, -0.05)
 
   cairo_rectangle(cr, 0.1, 0.1, 0.9, 0.9);
   cairo_stroke_preserve(cr);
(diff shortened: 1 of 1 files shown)


elstoc:
Really like the brush icon. Being really pedantic the brush head seems a bit big in relation to the body (though when drawn smaller that may not be so much of an issue) and the brush and pen icons should probably be pointing the same way for consistency - probably it's the pen icon that needs to change.

Apart from that though, looks nice.

Mark-64:
Direction of the pen is kind of forced by the "drawn and parametric" icon. The pen design is by @aurelienpierre and it was pointing in the other direction, I had to flip it.
Yes, the brush head is quite big, initially I did it smaller but it was looking tiny on an icon of few pixel. Maybe I enlaged too much now.
Les't wait other feedback.

elstoc:
Agreed - you'd need to flip all three (drawn, parametric, drawn and parametric) icons to make them consistent.

TurboGit:
For me the proposal looks good as-is. The first are "modes" to chose the drawn and parametric+drawn that's not a tool really. And the brush in this direction is fine, it really standout and make it clear that this is a tool not a mode as in opposite direction.

Let's not change  this until we have feedback from Nilvus too at least.

elstoc:
> For me the proposal looks good as-is

Agreed. My comments are nit picking really. It's very good as it is now.

aurelienpierre:
For me the brush icon looks more like a mop due to proportions, but no big deal.

johnny-bit:
Absolutelly 100% perfect IMO :D now it looks cute! an re-mop lookalike: ALL brush icons look like mops since it is just like Mark said - problematic to express brush tip with handful of pixels :)

that being said - maaaybe it's a tad bit too big compared to others and needs a bit of optical ballance, but otherwise it's perfectly cromulent.

elstoc:
Actually one thought from testing is that on the "drawn+parametric" icon the pen icon looks a little fat when it's drawn small. Perhaps since it's smaller than the 'drawn' icon it might be better off a bit thinner and without the middle line?

Mark-64:
Just for discussion, I did not commit those changes yet:
Drawn and parametric: pen not shortened and without middle line
immagine

Brush: slimmer, shorter and hopefully less mop-ish
immagine
But consider that this is a HiDPI screenshot

elstoc:
Looks better in my opinion (from the HiDPI screenshot). I wonder whether we need that centre line at all (on the drawn mask icon)? I think it looks cleaner (and would be more consistent) without.

johnny-bit:
Imo middle line is needed for bigger ver giving it a feel of pencil :) (bikesheding sorry!)

elstoc:
It's icons. It's all bikeshedding really ;-). All of the proposals so far are better than the current icons. Perhaps try the smaller one with the middle line retained - I just suspect it might not be visible on lower DPI screens.

Nilvus:
Great work @Mark-64: those new icons are much better than actual ones.
About direction, there's no need to have same direction on all. Only of course drawn and drawn+parametric masks needs to be consistent (same purpose) but brush is a tool in a mode, so really good as-is. Inverting direction make no sens I think and having different directions is also a way to think of right and left-handed. So I agree with you.

About middle pen line, I don't care but of course this should be the same on both mask icon(with and without) even if I prefer with middle line. Consistency between drawn and drawn+parametric mask remains the main think to care of. Pen in last mask is fatter yes but just because width seems the same as drawn one and length is shorter than in drawn mask.

I agree that brush head is too big and enlarged probably too much. But that seems better on your last screenshot.

Mark-64:
So, making a average of the feedback received, here is the tuned version.
This time I am in Ubuntu, non HiDPI.
Screenshot from 2020-06-22 21-43-37
Screenshot from 2020-06-22 21-44-00

elstoc:
Icons look good. Pen is much less 'stubby' now. Is it possible to stop the circle of the parametric mask showing through the pen icon (more noticeable on a lighter theme)?

TurboGit:
Thanks all !

PR #5605, opened 2020-06-22
thumbs : fix key up if first line not full

### LINKED ISSUE #5595: Lighttable: PgUp fails to display topmost row of thumbs
**Describe the bug**

In File Manager mode, if the lighttable thumbs grid starts with an image such that the number of prior images is not a multiple of the width (i.e. there is an incomplete first line of thumbs), it is not possible to use the PageUp key to reach that initial partial line of thumbs.  Once you have reached the first full line of thumbs, PageUp has no further effect.

**To Reproduce**
1. Open lighttable in File Manager mode, select a width of 4
2. Go to a folder/collection with at least 20 images
3. Scroll down to and select image number 20
4. Switch to darkroom and right back to lighttable
5. Note that image 20 is now the top left image on the lighttable
6. Press PageUp three times.  If the lighttable is showing about three rows, it should have scrolled twice and done nothing the third time.
7. select the top-left image and note that it is number 4
8. press Home (or Left Arrow) to display the first three thumbs

**Expected behavior**

The PageUp key should be able to scroll the display all the way to the top.

**Screenshots**

 - Darktable Version: master 3.1.0-2319, but has been around for months
 - OS: Mageia Linux 7.1
 - no OpenCL
 - GCC 8.4

**Additional context**

AlicVB:
this fix #5595

### DIFF
--- a/src/dtgtk/thumbtable.c
+++ b/src/dtgtk/thumbtable.c
@@ -2123,9 +2123,9 @@ void dt_thumbtable_update_accels_connection(dt_thumbtable_t *table, const int vi
   }
 }
 
-static gboolean _filemanager_ensure_rowid_visibility(dt_thumbtable_t *table, const int rowid)
+static gboolean _filemanager_ensure_rowid_visibility(dt_thumbtable_t *table, int rowid)
 {
-  if(rowid < 1) return FALSE;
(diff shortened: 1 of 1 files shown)


TurboGit:
Works for me. Thanks!

PR #5608, opened 2020-06-22
remove unclickable spacing on module header and replace with CSS

elstoc:
this change removes space at the bottom of lib and iop module headers
that could not be clicked (or where clicking it either caused focus to
be transferred without the module being expanded or for nothing to
happen)

this additional space was previously corrected-for in CSS so this change
also adjusts the CSS (since that correction is no longer required)

### DIFF
--- a/data/themes/darktable.css
+++ b/data/themes/darktable.css
@@ -595,7 +595,7 @@ overshoot.right
 /* Set modules name header */
 #module-header
 {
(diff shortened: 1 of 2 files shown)


elstoc:
The extra space at the bottom of the header bar can more easily be seen using gtk inspector:

Screenshot_2020-06-22_23-37-49

TurboGit:
Works for me! One more use of CSS, that's great. Thanks.

PR #5612, opened 2020-06-23
splittoning: invert gradient display in balance.

### LINKED ISSUE #5606: gradient in split-toning balance slider seems inverted
**Describe the bug**
Gradient in split-toning balance slider seems inverted 

**To Reproduce**
1. Go to darkroom
2. Open split-toning, let default
3. Move balance slider to the right (more red) the image becomes more green

**Expected behavior**
The balance slider should be more intuitive, I'm pretty sure this is a regression.

**Screenshots**

 - Darktable Version: master
 - OS: GNU/Debian
 - 
 - 

**Additional context**

TurboGit:
Make sure it goes from shadows to highlight so the slider does
change the color in the right direction.

Fixes #5606.

### DIFF
--- a/src/iop/splittoning.c
+++ b/src/iop/splittoning.c
@@ -305,12 +305,12 @@ static inline void update_balance_slider_colors(
   if(shadow_hue != -1)
   {
     hsl2rgb(rgb, shadow_hue, 1.0, 0.5);
-    dt_bauhaus_slider_set_stop(slider, 0.0, rgb[0], rgb[1], rgb[2]);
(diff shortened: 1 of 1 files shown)


TurboGit:
@pitbuster : @dterrahe : is that looking good to you?

The 1.0 for shadows & 0.0 for highlight is because the slider is having an offset of -100.

dterrahe:
@TurboGit I hope #5615 is a cleaner solution although since this at the moment is the only place that broke, the impact of both fixes would be the same.

TurboGit:
Closing as better fix put in place. Thanks @dterrahe !

PR #5614, opened 2020-06-23
tone equalizer: hide mask when losing focus

elstoc:
hides the exposure mask in the tone equalizer when the module loses focus


TurboGit:
Works for me. Thanks!

PR #5616, opened 2020-06-23
switch off color pickers in rgb curve and color zone when dt_iop_color_picker_reset called from blender

### LINKED ISSUE #5607: rgb curve - color picker not desiabled when selecting any blending mode
**Describe the bug**
In rgb curve - color picker not desiabled when selecting any blending mode

**To Reproduce**
1. Go to darkroom
2. Open rgb curve module
3. Select color picker standard or the +/- on top
4. Click on any blending mode, say drawn mask
5. Picker is not disabled
6. Select pencil
7. Click on image and it still change the color picker location

**Expected behavior**
The color picker must be disabled when a blending mode is selected.

 - Darktable Version master
 - OS: GNU/Debian

**Additional context**

Most probably due to the recent introspection changes. We probably want to double check other modules. "rgb levels" seems to work properly.

dterrahe:
specifically fixes issue with mask drawing in blender.

this cancels existing functionality to keep the left color picker active when module settings are changed, to directly be able to see the impact. this could possibly be reintroduced as part of further clean-up in blender.

fixes #5607

### DIFF
--- a/src/develop/blend_gui.c
+++ b/src/develop/blend_gui.c
@@ -372,7 +372,7 @@ static void _blendop_masks_mode_callback(const unsigned int mask_mode, dt_iop_gu
   else if(data->blendif_inited)
   {
     /* switch off color picker */
-    dt_iop_color_picker_reset(data->module, TRUE);
+    dt_iop_color_picker_reset(data->module, FALSE);
 
     gtk_widget_hide(GTK_WIDGET(data->blendif_box));
   }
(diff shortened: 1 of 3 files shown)


TurboGit:
In colorzone & rgbcurve you don't want to remove the picker area/bar when adjusting the curve.

dterrahe:
Makes sense; I'll keep it so that certain pickers normally _won't_ get switched off, if marked with "keep-active", but blender will switch those off _anyway_.

dterrahe:
I _believe_ this fully restores previous behavior.

TurboGit:
Still not working for me:
- open colorzone
- select area picker (second one)
- draw an area on the image
- this highlight the corresponding region on the display

image

- try to move a control point in the line
- see that the area is removed

TurboGit:
Same issue with "rgb curve". But it works properly on "tone curve".

dterrahe:
Is this a regression though? If I understand correctly what you are describing, this is the same in 3.0.2. And it makes sense, I think, because the second color picker _sets_ the curve, so if you want to override/adjust that, you are rejecting the color picker and it should be switched off.

Only the first/left color picker can stay active, because it is just for display purposes.

Let me know if I'm misunderstanding please.

TurboGit:
@dterrahe : I'm probably too tired... You're right, this has always been the case. The area with Ctrl+Click on the first picker is kept. My bad and sorry for the noise! Merging.

PR #5623, opened 2020-06-23
[bilat] reread sliders on mode change and change mode on hidden slider move

### LINKED ISSUE #5620: local contrast: error when changing mode
If I open and enable the local contrast module, its mode defaults to 'local laplacian filter':

Screenshot_2020-06-23_16-22-47

If I change the mode to 'bilateral grid', all looks fine:

Screenshot_2020-06-23_16-22-56

However, if I then change back to 'local laplacian filter' I get the following:

Screenshot_2020-06-23_16-23-06

3.1.0+2354~g2272cc4aa on ArchLinux. Issue occurs both with and without OpenCL.

dterrahe:
In "local contrast" two sets of sliders are linked to the same fields and the mode determines which one is used. When changing modes, the values of the now activated sliders need to be reloaded into the field.

Previously those sliders would be reinitialised when unhidden, but this is not really necessary, so they now retain their values when hidden. They are just set to default values when a new image is loaded. So switching between modes doesn't wipe out settings.

Also, when hidden sliders are moved (using (dynamic) shortcuts, their values would override a field that currently has a different meaning leading to "interesting" results. Now, such a move leads to a mode change that makes those sliders visible. Especially when the module is hidden, this dual slider+mode change might still be somewhat unexpected.

Fixes #5620

### DIFF
--- a/src/iop/bilat.c
+++ b/src/iop/bilat.c
@@ -358,17 +358,35 @@ void gui_changed(dt_iop_module_t *self, GtkWidget *w, void *previous)
 {
   dt_iop_bilat_gui_data_t *g = (dt_iop_bilat_gui_data_t *)self->gui_data;
   dt_iop_bilat_params_t *p = (dt_iop_bilat_params_t *)self->params;
-  if(w == g->mode)
+  if(w == g->highlights || w == g->shadows || w == g->midtone)
+  {
+    dt_bauhaus_combobox_set(g->mode, s_mode_local_laplacian);
+  }
+  else if(w == g->range || w == g->spatial)
+  {
+    dt_bauhaus_combobox_set(g->mode, s_mode_bilateral);
+  }
+  else if(w == g->mode)
+  {
+    if(p->mode == s_mode_local_laplacian)
+    {
+      p->sigma_r = dt_bauhaus_slider_get(g->highlights);
+      p->sigma_s = dt_bauhaus_slider_get(g->shadows);
+    }
+    else
+    {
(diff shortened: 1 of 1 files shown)


dterrahe:
@elstoc this is funky stuff; please test to an inch of its life. Especially; changing modes back and forth, using shortcuts to move sliders, resetting the module or individual sliders to their defaults (and then changing mode), moving between images (and then changing modes).

Is the behavior I have now implemented at least somewhat intuitive? The side-effects of using the same field in two different meanings, depending on mode, cannot be completely undone, but at least I'm hoping this makes them a little less surprising.

elstoc:
The bug appears fixed, though I'm not sure I'm qualified to comment on the rest. I mostly only use local contrast with the default settings and some minor tweaks, so unless it does something really funky I won't know whether it's working correctly or not.

I only found the bug in the first place with random clicking ;-)

elstoc:
I've tried all the shortcut/slider combinations I could think of and I can't break it. It seems reasonable to me for it to switch mode if using a shortcut to change a currently-non-applicable slider. So all good as far as I'm concerned.

TurboGit:
Works for me. This can be seen as an enhancement or bugfix, so I'm merge into 3.2.

TurboGit:
Thanks !

dterrahe:
Thanks!

> Works for me. This can be seen as an enhancement or bugfix, so I'm merge into 3.2.

Yeah, definitely a bug fix for a regression which could only be fixed with an enhancement ;-)

PR #5624, opened 2020-06-23
remove extra space at top of modules

elstoc:
where a lib or iop module is split into sections, reduce the amount of
space added at the top of the first section.

impacts splittoning, watermark and export modules

### DIFF
--- a/data/themes/darktable.css
+++ b/data/themes/darktable.css
@@ -633,7 +633,8 @@ overshoot.right
   margin: 12px 0 6px 0;
 }
 
-#iop-plugin-ui #section_label.section_label_top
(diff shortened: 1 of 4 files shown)


johnny-bit:
Can you add comment/documentation somewhere about it? I use sections in my temperature.c PR and I'd probably miss this stuff :/

elstoc:
> Can you add comment/documentation somewhere about it?

Where would you like it added? This is something I introduced when I split the filmic reconstruct tab into sections and it looked silly with a big gap at the top, so I've just reapplied it to some more modules. Just add the class section_label_top to the top section label and it will get a reduced top margin.

PR #5626, opened 2020-06-24
Filmstrip active triangle

AlicVB:
as discussed, this PR adds a little triangle to show more obviously the current active image(s).
At the same time, this fix 2 little issues : 
- fix active images list which is lost in certain cases with culling or preview (when starting in culling mode for example)
- fix the top border in culling

### DIFF
--- a/data/themes/darktable.css
+++ b/data/themes/darktable.css
@@ -1948,6 +1948,15 @@ spinbutton>button
 {
   border-top: 2.5px solid @border_color;
 }
+/* Set top cursor on active image in filmstrip */
+#thumb_cursor
+{
+  color: transparent;
+}
+#thumb_main:active #thumb_cursor
+{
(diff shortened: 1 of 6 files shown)


Nilvus:
Even before testing, I already like to see that. Will test now!

data/themes/darktable.css
```diff
@@ -1948,6 +1949,15 @@ spinbutton>button
 {
   border-top: 2.5px solid @border_color;
```
Nilvus, line 1950: With triangle, border top is maybe too big. I suggest to try with 1.5px (or 2px max) only. That's actually a little too big with the triangle now.

AlicVB: you're right of course... done !

TurboGit:
Perfect to me too! Thanks.

PR #5628, opened 2020-06-24
Optional export to grayscale tiff files

### LINKED ISSUE #5621: Mono over-ride
**Dumb printer refuses to load mono tif**

My printer refuses to load a tif which isn't in RGB mode, even if set to "Print in Black and White"

**A manual over-ride on the mono-mode tif output**

It shouldn't be so, but that seems the simplest solution...

@jenshannoschwalm I think this is your work that I'm asking you to desecrate?

jenshannoschwalm:
Some printers/drivers dislike the recently introduced smaller monochrome
tiff files so this should be optional as requested.

I think this should be an option right inside the tiff export widget and not a global
easy-to-miss option or even a hidden one.

Fixes #5621

### DIFF
--- a/src/imageio/format/tiff.c
+++ b/src/imageio/format/tiff.c
@@ -45,6 +45,7 @@ typedef struct dt_imageio_tiff_t
   int bpp;
   int compress;
   int compresslevel;
+  int shortfile;
   TIFF *handle;
 } dt_imageio_tiff_t;
 
@@ -53,6 +54,7 @@ typedef struct dt_imageio_tiff_gui_t
   GtkWidget *bpp;
   GtkWidget *compress;
   GtkWidget *compresslevel;
+  GtkWidget *shortfiles;
 } dt_imageio_tiff_gui_t;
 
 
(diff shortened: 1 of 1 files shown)


GrahamByrnes:
Impeccable!
I would have just thrown in a "if(dt_conf... ==rgb)..." :-)

jenshannoschwalm:
@TurboGit could you have a look. I would concider this as a bugfix.

TurboGit:
@jenshannoschwalm : can you remind me what this option does ? I have tested a B&W image (using monochrome module). When I export is rgb or greyscale the image is the same size and looks the same. Is that supposed to do something meaningful only for true B&W RAW ?

GrahamByrnes:
It was for me, really :-)
It should look the same: it's just the mono file duplicated across the three channels of rgb, so that my print software can read it.
Possibly only of interest to Windows users.

jenshannoschwalm:
Exactly :-)

TurboGit:
> It should look the same: it's just the mono file duplicated across the three channels of rgb, so that my print software can read it.

That's basically what I understood, but then why the TIFF is the same size when we have three channel with the same value instead of the a single one with the grey value ? I was expecting the later to be smaller.

When I export my color picture I have a file of 14Mb TIFF. Then I add the monochrome module and export with the new setting set to RGB, file is 5Mb and when I set it to GREYSCALE it is also 5Mb. So to me this option seems to do nothing, again expect I'm still not understanding the implementation.

GrahamByrnes:
I see 57MB vs 38MB...

jenshannoschwalm:
Please note that the file size reduction varies a lot depending on compression and the algo behind. Effect is best seen with uncompressed tiffs of course. I tested again on a number of 50MP images with compressed 16bit output and the file size reduction is something like 25% for most cases. Also if you add some tint it won't write single channel.

One more, when implementing the mono output i made sure to write all 3 channels for multi-page images. This would be the case if you export masks. So that might be the reason for your same-size issue @TurboGit

TurboGit:
Ok, probably a procedural error on my side while testing.

Exported size
- RAW as-is : 14Mb
- RAW with monochrome module activated in RGB mode : 14Mb
- RAW with monochrome module activated in GREYSCALE mode : 4.5Mb.

So all good! Thanks.

PR #5630, opened 2020-06-24
more compressed color picker module

elstoc:
reduce the size of the picker area and the space between comboboxes

allow the lhs panel to be shrunk further at the expense of the picker area
not always being perfectly square

when panel is expanded past the minimum width, space is shared evenly
between the picker area and the comboboxes

### DIFF
--- a/data/themes/darktable.css
+++ b/data/themes/darktable.css
@@ -1424,8 +1424,15 @@ cell:selected
 /* Module in darkroom left panel */
 #color-picker-area
 {
-  min-height: 9em;
-  min-width: 9em;
+  min-height: 7em;
+  min-width: 7em;
+}
+.picker-module combobox,
+.picker-module combobox button
+{
+  padding-top: 0;
+  padding-bottom: 0;
+  border: 0;
 }
 
 #live-sample


data/themes/darktable.css
```diff
@@ -1424,8 +1424,15 @@ cell:selected
 /* Module in darkroom left panel */
 #color-picker-area
 {
-  min-height: 9em;
-  min-width: 9em;
+  min-height: 7em;
+  min-width: 7em;
```
Nilvus, line 1428: Set to 8em on those 2 lines.

src/libs/colorpicker.c
```diff
@@ -558,7 +558,7 @@ void gui_init(dt_lib_module_t *self)
   data->color_patch = gtk_drawing_area_new();
   g_signal_connect(G_OBJECT(data->color_patch), "draw", G_CALLBACK(main_draw_callback), data);
   gtk_box_pack_start(GTK_BOX(picker_area), data->color_patch, TRUE, TRUE, 0);
-  gtk_box_pack_start(GTK_BOX(output_row), picker_area, FALSE, FALSE, 0);
+  gtk_box_pack_start(GTK_BOX(output_row), picker_area, TRUE, TRUE, 0);
```
Nilvus, line 561: Cancel those changes as it's for me better on master (so without that) but with CSS tweaks and 8em instead of 7em

Nilvus:
With my changes requested, the idea of compressed this module remains with the good thing that color picker come back to be square.

elstoc:
The problem with having a minimum width that large is that it imposes a minimum width for the entire side panel, which is why I initiated this change. If you try to reduce the LHS panel width to the point where this box can no longer shrink any more, the whole panel UI moves behind the image

Screenshot_2020-06-25_11-23-30

I'm all for a nice square box but I'm not sure it should be at the expense of UI problems like the one above.

src/libs/colorpicker.c
elstoc: I'm fine with this change. This was more personal preference.

data/themes/darktable.css
elstoc: I'm not so sure about this one. See my comments below

elstoc:
An alternative might be to move the RGB values to below the picker area, then the size of the box could be significantly reduced while keeping it square. It does seem (to me) very large.

Nilvus:
> The problem with having a minimum width that large is that it imposes a minimum width for the entire side panel, which is why I initiated this change. If you try to reduce the LHS panel width to the point where this box can no longer shrink any more, the whole panel UI moves behind the image

That is not a specific color picker module. You could see for the snapshot button if you reduce a lot, for the history, etc. And that's normal, a minimum width is needed to correctly show text. So that's not something to argue for.  And that's why we have possibility to hide/show panels. So breaking the UI of color picker for that is not a good thing for me.

> I'm all for a nice square box but I'm not sure it should be at the expense of UI problems like the one above. Personally I'd be more happy with 5 or 6em but I thought I'd selected a good compromise with 7.

So for that, remove changes on colorpicker.c you made as having a fixed size is better. With your changes, if you increase width panel, the width of color picker area also increase. And it's awful and not wanted.

I think 8em is better but without colorpicker.c, but just tested 7.5em and it's a better compromise maybe for what you want and what I think.

Nilvus:
> An alternative might be to move the RGB values to below the picker area, then the size of the box could be significantly reduced while keeping it square. It does seem (to me) very large.
> 
> Something like this:
> 
> Screenshot_2020-06-25_12-01-22

No way for me. For consistency and purpose of those infos, that's a bad idea. And same as before, reducing panel will have bad effect you describe on a lot of modules. So that's not the good way to deal with that. Or the user hide the panel if he want to reduce it a lot as minimum space needed to read infos on a module or we add horizontal scrollbar if reducing too much the panel (and all needed modules ; some already have that).

elstoc:
> You could see for the snapshot button if you reduce a lot, for the history, etc. And that's normal, a minimum width is needed to correctly show text.

For most modules the minimum width is quite a lot smaller than that for the colorpicker and actually, most buttons/sliders etc. ellipsize text to allow the widths to be reduced further. There are only a couple of modules left now that impose a large minimum width on the panel: tagging (which I don't use so am not bothered about) and export selected (for which the minimum is imposed because of the buttons on the max size spinners). 

Generally most modules on the left hand side of the darkroom view can be horizontally shrunk quite a bit without impacting their functionality, which means more space for the image itself.

Nilvus:
So, to resume, we are agree to remove colorpicker.c changes.

And about CSS lines, I would be more for 7.5em than 8em as a compromise. Hope it would be ok for you to have that compromise. And if ok, push those changes so that this can be merged.

elstoc:
One last proposal, which allows us to keep the box square and keep it smaller, is to not force it to take up the full height of the comboboxes, like so:

Screenshot_2020-06-25_13-52-55

Nilvus:
@elstoc: just remove colorpicker.c file changes as I said and if ok for you set 7.5em (or at worst keep 7em you set) and it will be all good. Not necessary to go further there. If it's not totally square it's not a problem, and better to let space on bottom like your last proposal. Keep it simple!

elstoc:
> Keep it simple!

What I've just proposed was simple (5 lines of code). What was wrong with it?

Nilvus:
> > Keep it simple!
> 
> What I've just proposed was simple (5 lines of code). What was wrong with it? We could even keep the large size by default if you like (I've reduced it probably more than you'd like - to 6.5em above) - it just means that if a user wants to reduce the size they can do so while keeping it square.

Did your read my comment. The answer is on it. Space below color picker area on your last proposal is not good and so I prefer a not totally square color picker area. And as I said, just remove colorpicker.c changes and keep 7em on lines (or better for me, set them to 7.5em). Even if I don't care fo number lines of code changing (I just care of how it visually rendered), it's simpler to remove 1 change and keep (or set a little adjustment on 2 existing changes) than adding 5 lines...

And remember that you were agree to remove colorpicker.c changes and I'm ok if you keep 7em... So no need to spent time on this. That's also simpler to not spend to much times on such changes.

TurboGit:
Thanks to both of you!

PR #5632, opened 2020-06-25
Move timestamps to bottom of image information

### LINKED ISSUE #5609: Image Information Display
**Describe the bug**
When the import, change, export, and print timestamps got added to the database they were inserted into the metadata.c code where it seemed to make the most sense, according to the code comments.  However, the data displays in the order that it's declared, so the timestamp information is displayed while image information such as exposure, ISO, timestamp, height, width... are pushed out of the viewport.

image_info

**To Reproduce**
1. Open darktable
2. Hover over an image

 - Darktable Version: git master
 

**Additional context**
This is simple enough to fix, but I thought that I would open an issue in case anyone else had ideas about the information displayed and the order.

My plan is to move the timestamps to the bottom, unless someone has a better idea of where to place them.

wpferguson:
Moved import, changed, export, and print timestamps to the bottom of the image information display.  Fixes #5609.

### DIFF
--- a/src/libs/metadata_view.c
+++ b/src/libs/metadata_view.c
@@ -51,10 +51,6 @@ enum
   md_internal_version,
   md_internal_fullpath,
   md_internal_local_copy,
-  md_internal_import_timestamp,
-  md_internal_change_timestamp,
-  md_internal_export_timestamp,
-  md_internal_print_timestamp,
 #if SHOW_FLAGS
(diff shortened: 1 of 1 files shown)


wpferguson:
I left this open in case some screaming started when 3.2.1 was released.  Since I haven't seen/heard any screaming, I'll close this.

PR #5634, opened 2020-06-25
Some cleaning up of blending gui

### LINKED ISSUE #5656: parametric mask: "set range" color picker doesn't keep gradient slider area markers
If I use the "set range" color picker in a parametric mask, it also shows the applicable range of the selected area on the input/output gradient sliders, which is very useful.

Screenshot_2020-06-28_19-15-00

However, if I then manually move one of the sliders, this range is no longer shown

Screenshot_2020-06-28_19-15-25

The only way to get it back is to re-select the same area with the second color picker. But it's very useful to still keep a visual representation of the selected area in order to assist with any adjustments.

I suspect this is probably more of a feature request since I see the same behaviour on 3.0.2

dterrahe:
Including using new dt_bauhaus_combobox_set_from_value and generic slider/combobx callback functions

For 3.4, but I've also fixed some help links. @TurboGit I'm happy to put those in a separate PR for 3.2 if you want.

Edit: Closes #5656

### DIFF
--- a/src/bauhaus/bauhaus.c
+++ b/src/bauhaus/bauhaus.c
@@ -1233,6 +1233,24 @@ gboolean dt_bauhaus_combobox_set_from_text(GtkWidget *widget, const char *text)
   return FALSE;
 }
 
+gboolean dt_bauhaus_combobox_set_from_value(GtkWidget *widget, int value)
+{
+  dt_bauhaus_widget_t *w = DT_BAUHAUS_WIDGET(widget);
+  if(w->type != DT_BAUHAUS_COMBOBOX) return FALSE;
+  dt_bauhaus_combobox_data_t *d = &w->data.combobox;
+  int i = 0;
+  for(GList *iter = d->entries; iter; iter = g_list_next(iter), i++)
+  {
+    const dt_bauhaus_combobox_entry_t *entry = (dt_bauhaus_combobox_entry_t *)iter->data;
+    if(GPOINTER_TO_INT(entry->data) == value)
+    {
+      dt_bauhaus_combobox_set(widget, i);
+      return TRUE;
+    }
+  }
+  return FALSE;
+}
+
 int dt_bauhaus_combobox_get(GtkWidget *widget)
 {
   dt_bauhaus_widget_t *w = DT_BAUHAUS_WIDGET(widget);
--- a/src/develop/blend_gui.c
+++ b/src/develop/blend_gui.c
@@ -2483,7 +2483,7 @@ void dt_iop_gui_init_blending(GtkWidget *iopw, dt_iop_module_t *module)
                      G_CALLBACK(_blendop_masks_invert_callback), bd);
 
     bd->masks_feathering_guide_combo = dt_bauhaus_combobox_new(module);
-    dt_bauhaus_widget_set_label(bd->masks_feathering_guide_combo, _("feathering guide"), _("feathering guide"));
+    dt_bauhaus_widget_set_label(bd->masks_feathering_guide_combo, _("blend"), _("feathering guide"));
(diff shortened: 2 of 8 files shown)


dterrahe:
Regarding help links; are they checked as part of the release process? It looks like the intention was for all of them to be moved to usermanual_url.c? Maybe dt_get_help_url should be merged into dt_gui_add_help_link as part of a total cleanup which feels too big for 3.2.

Looking at it for the first time now, I'm wondering why short tags (like "ratings") are translated in usermanual_url.c (to "star_ratings_and_color_labels.html#star_ratings_and_color_labels") rather than online. Wouldn't it be easier to continue to support people running old versions if they come in via say https://darktable.gitlab.io/doc/3.2/en/helptag/ratings rather than having to maintain a very specific long url (that isn't versioned) for a few years? Not an area I know much about and this was probably extensively discussed before. But if not, should I raise an RFC issue? Or is this already part of the whole documentation effort that I've not been paying particular attention to (sorry!)

TurboGit:
> "star_ratings_and_color_labels.html#star_ratings_and_color_labels"

This is not really an URL, but the page where the description is and a specific ref in the table. We try to keep the html pages and the refs from version to version. So it should be fine.

TurboGit:
And yes please, better have a separate PR for  the user manual help links in a separated PR.

dterrahe:
Do not switch off blending color pickers when blending multi-value sliders are moved. Allows to continue seeing picked range while adjusting mask boundaries.

Fixes #5656

src/develop/blend_gui.c
```diff
@@ -54,14 +55,79 @@ typedef enum _iop_gui_blendif_channel_t
   ch_max = 4
 } _iop_gui_blendif_channel_t;
 
+const dt_develop_name_value_t dt_develop_blend_mode_names[] = {
```
TurboGit, line 58: current dt style is to have "= {" below, can you fix this here and below for consistency.

TurboGit:
Apart from my style comment it looks fine, a nice clean-up!

src/develop/blend_gui.c
dterrahe: I was trying to indicate a difference between null terminated and un-terminated lists ;-)

For the null terminated lists of modes, all elements can be freely moved around, and not having any "special" rows that have = { at the beginning would make the changelog cleaner. But there's really no useful purpose for this, so reverted everything to standard style as requested. Thanks!

src/bauhaus/bauhaus.c
```diff
@@ -1216,6 +1216,24 @@ gboolean dt_bauhaus_combobox_set_from_text(GtkWidget *widget, const char *text)
   return FALSE;
 }
 
+gboolean dt_bauhaus_combobox_set_from_value(GtkWidget *widget, int value)
```
TurboGit, line 1219: Just a comment, do you think it would be clearer to name this routine:
dt_bauhaus_combobox_set_from_index() for me value could be the text actually. Index make it clear that we are looking for the Nth value. What do you think?

dterrahe: Hm... that's not what it does though? Setting from the number of the item is what the normal `dt_bauhaus_combobox_set` does. This finds the enum value (in case the combobox is setting an enum). So could be sparse and/or out-of-order. It could be called `dt_bauhaus_combobox_set_from_data`, but the fact that the enum value is internally stored in a field called _data_ to me feels like an implementation detail.

TurboGit: I see, sorry I read wrongly the code :) All good then.

TurboGit:
There is a conflict to be solved and it is ready for merging.

dterrahe:
Yep; totally expected: there was bound to be a merge conflict with #5691 depending on the order they got merged.

Fixed!

TurboGit:
Thanks !

PR #5635, opened 2020-06-25
Improve progress bar

### LINKED ISSUE #5631: progress bar label padding is awkward, needs CSS fix
**Describe the bug**
The progress bar looks visually awkward. It could use some padding, particular on the left side.

**To Reproduce**
1. Go to lighttable view
2. Select a few images, preferably ones which will be slow to process, so that the progress bar will stay visible
3. Type control-e to start export
4. During export, a progress bar should appear in the bottom left of the window.

**Expected behavior**
The margins/padding of the label should be more pleasing.

**Screenshots**
image

Left/bottom of screenshot are flush with left/bottom of darktable window. The scrollbar-looking thing at the bottom is actually the progress bar.

- Darktable Version: master 851d72a1922a57c2d0b53574e62ee4f3fc582b98
- OS: Debian bullseye/sid
- OpenCL activated: yes

**Additional context**
- I've tried this with the export and print progress bar. Presumably they are all fairly awkward.
- I'm not sure why there is a vertical line in the cancel "X" in the screenshot.

Nilvus:
Fix #5631 and more as it improves progress bar to make that more discrete and integrate as before. See #5631 for discussion about that.

With that PR, progress bar become that:

Example on grey theme but same contrast on other themes.

### DIFF
--- a/data/themes/darktable.css
+++ b/data/themes/darktable.css
@@ -809,7 +809,6 @@ menuitem > menu, /* sub-menu */
 menuitem arrow,
 tooltip,
 popover,
-#background_job_eventbox, /* background of export progress bar */
 combobox window,
 dialog combobox window
 {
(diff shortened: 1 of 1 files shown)


TurboGit:
Thanks !

PR #5637, opened 2020-06-25
clipping: fix original image aspect ratio

### LINKED ISSUE #5629: crop & rotate: "original image" setting not retained
If i choose "original image" in the crop & rotate aspect drop-down, this is replaced with a ratio and the "original image" setting is not retained. For example in the following I have simply enabled crop & rotate, selected "original image" and committed my changes:

Screenshot_2020-06-24_20-03-31

elstoc:
ensures that the "original image" aspect ratio is not immediately
converted to a numeric ratio on committing settings

Resolves #5629

### DIFF
--- a/src/iop/clipping.c
+++ b/src/iop/clipping.c
@@ -1370,17 +1370,17 @@ static float _ratio_get_aspect(dt_iop_module_t *self, GtkWidget *combo)
 
   //retrieve full image dimensions to calculate aspect ratio if "original image" specified
   const char *text = dt_bauhaus_combobox_get_text(combo);
-  if(text && !g_strcmp0(text,"original image"))
+  if(text && !g_strcmp0(text,_("original image")))
   {
     int proc_iwd = 0, proc_iht = 0;
     dt_dev_get_processed_size(darktable.develop, &proc_iwd, &proc_iht);
 
     if(!(proc_iwd > 0 && proc_iht > 0)) return 0.0f;
 
-    p->ratio_d = proc_iwd;
-    p->ratio_n = proc_iht;
-    if(proc_iwd >= proc_iht) return (float)proc_iwd / (float)proc_iht;
-    else  return (float)proc_iht / (float)proc_iwd;
+    if((p->ratio_d > 0 && proc_iwd > proc_iht) || (p->ratio_d < 0 && proc_iwd < proc_iht))
+      return (float)proc_iwd / (float)proc_iht;
+    else
+      return (float)proc_iht / (float)proc_iwd;
   }
 
   // we want to know the size of the actual buffer


elstoc:
Now fixed to handle portrait orientation

TurboGit:
I have tried hard to reproduce this issue, but I cannot ! In no circumstance I was able to get a wrong ratio in the aspect box. The text stays "original image" for me. So, this may already fixed or there is a very specific context to reproduce ?

elstoc:
For me this is consistently reproduceable in master (3.1.0+2374~gda0d15763) - I've tried on images from two different cameras. Go into the darkroom, discard history, select crop&rotate, choose 'original image' and then double-click on the image to commit. 

The issue (as far as I could tell) was that the 'original' doesn't store the actual crop dimensions but just stores p_ratio_d and p_ratio_n as 1 and 0 so when the code tried to find a match to the combobox using the full pixel height/width it couldn't find one so constructed the text itself.

elstoc:
Are you testing on an English version or a translated one? It's just that the code in question is checking the combobox value against an untranslated string meaning that entire if statement might be being ignored. Commenting out that entire if statement also seems to resolve the bug.

dterrahe:
I'm definitely seeing #5629 as well.

elstoc:
> I'm definitely seeing #5629 as well.

Thanks @dterrahe. Are you able to test if my PR resolves it?

dterrahe:
Your PR resolves the obvious case I'm seeing, that only happened when language set to English. Which means that your code is not being used anyway if in a translated language. Which, again, means that if it is actually needed, translated languages will still be broken, even if maybe in a less obvious way.

I'm wondering why that code is there anyway and have to admit I don't fully understand what's going on. But my guess is that this is to deal with presets, where the actual ratio depends on the final image the preset is applied to. And that this is why the sizes are reread, rather than using the stored proportions. If this is indeed the case, then when applying a preset (or maybe this happens in other cases too?) in French, "something" won't work.

Your PR may resolve the case for English, but in French it would still be broken. Just adding translation brackets _(...) may be all that is needed to make this fully correct...

elstoc:
> Your PR may resolve the case for English, but in French it would still be broken. Just adding translation brackets _(...) may be all that is needed to make this fully correct...

I think you're probably right. This code was introduced in #5369 so @GrahamByrnes might have an opinion on this fix.

GrahamByrnes:
Yes, I reproduce #5629
In fact I don't think I ever committed changes when testing. Oops.
But like @dterrahe I never really understood how ratios were being saved (if I'm not being inaccurately rude).
I'll go look...

elstoc:
I think we just need to make the "original image" text translatable and then the if statement will work for all languages

GrahamByrnes:
> I think we just need to make the "original image" text translatable and then the if statement will work for all languages

Ohhh... yes. D'oh.

elstoc:
Now fixed so that the "original image" string is translatable.

TurboGit:
Indeed I can reproduce with the English version... So maybe the proper fix is to force everyone to use the French UI since it is not buggy :)

I'll check the patch soon otherwise!

TurboGit:
Working on the English & French version with this patch. Good detective work! Thanks.

PR #5646, opened 2020-06-27
fix erroneous API usage

rabauke:
Fixes the functions `dt_accel_connect_manual` and `dt_accel_disconnect_list`, which use the GTK functions `g_slist_prepend` and `g_slist_delete_link` not correctly. More precisely, the return value is not utilized as designed by the GTK library.

(I have no idea why the current code works. It should not according to [this documentation](https://developer.gnome.org/glib/stable/glib-Singly-Linked-Lists.html#g-slist-prepend) and [this documentation](https://developer.gnome.org/glib/stable/glib-Singly-Linked-Lists.html#g-slist-delete-link).

### DIFF
--- a/src/gui/accelerators.c
+++ b/src/gui/accelerators.c
@@ -534,14 +534,14 @@ void dt_accel_connect_lua(const gchar *path, GClosure *closure)
   gtk_accel_group_connect_by_path(darktable.control->accelerators, accel_path, closure);
 }
 
-void dt_accel_connect_manual(GSList *list, const gchar *full_path, GClosure *closure)
+void dt_accel_connect_manual(GSList **list_ptr, const gchar *full_path, GClosure *closure)
 {
   gchar accel_path[256];
   dt_accel_path_manual(accel_path, sizeof(accel_path), full_path);
   dt_accel_t *accel = _lookup_accel(accel_path);
   accel->closure = closure;
   gtk_accel_group_connect_by_path(darktable.control->accelerators, accel_path, closure);
-  list = g_slist_prepend(list, accel);
+  *list_ptr = g_slist_prepend(*list_ptr, accel);
 }
 
 static gboolean _press_button_callback(GtkAccelGroup *accel_group, GObject *acceleratable, guint keyval,
@@ -922,14 +922,16 @@ void dt_accel_connect_locals_iop(dt_iop_module_t *module)
   module->local_closures_connected = TRUE;
 }
 
-void dt_accel_disconnect_list(GSList *list)
+void dt_accel_disconnect_list(GSList **list_ptr)
 {
+  GSList *list = *list_ptr;
   while(list)
   {
     dt_accel_t *accel = (dt_accel_t *)list->data;
     if(accel) gtk_accel_group_disconnect(darktable.control->accelerators, accel->closure);
     list = g_slist_delete_link(list, list);
   }
+  *list_ptr = NULL;
 }
 
 void dt_accel_disconnect_locals_iop(dt_iop_module_t *module)
--- a/src/views/darkroom.c
+++ b/src/views/darkroom.c
@@ -716,7 +716,7 @@ int try_enter(dt_view_t *self)
 
 static void dt_dev_cleanup_module_accels(dt_iop_module_t *module)
 {
-  dt_accel_disconnect_list(module->accel_closures);
+  dt_accel_disconnect_list(&module->accel_closures);
   dt_accel_cleanup_locals_iop(module);
 }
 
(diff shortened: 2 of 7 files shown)


src/gui/accelerators.c
```diff
@@ -534,14 +534,15 @@ void dt_accel_connect_lua(const gchar *path, GClosure *closure)
...
 
-void dt_accel_connect_manual(GSList *list, const gchar *full_path, GClosure *closure)
+void dt_accel_connect_manual(GSList **list_ptr, const gchar *full_path, GClosure *closure)
 {
   gchar accel_path[256];
   dt_accel_path_manual(accel_path, sizeof(accel_path), full_path);
   dt_accel_t *accel = _lookup_accel(accel_path);
   accel->closure = closure;
   gtk_accel_group_connect_by_path(darktable.control->accelerators, accel_path, closure);
-  list = g_slist_prepend(list, accel);
```
rabauke, line 544: The result is assigned to a local variable, which is never read afterwards.

elstoc:
Why the thumbtable accels still work is a mystery to me as well. I suspect that gtk probably just doesn't have an issue with multiple similar accels being present and probably darktable just finds the first one it can and ignores any duplicates. In the case of the thumbtable function I agree that  ``dt_accel_disconnect_list`` probably doesn't do anything but only because ``table->accel_closures`` isn't pointing to the head of the list.

To fix that, every call to ``dt_accel_connect_manual`` should return the pointer to the head of the list and should be called with ``table->accel_closures = dt_accel_connect_manual...``.

However, I'm not sure that the changes you've proposed to ``dt_accel_disconnect_list`` are technically necessary since your proposal is functionally equivalent to what went before (it clears the list and sets the appropriate pointer to NULL - so if you just added ``table->accel_closures = NULL`` to the thumbtable function you'd be fine and consistent with the other calls). 

This is probably fine for a code tidy-up but I don't see why the current code shouldn't work.

src/gui/accelerators.c
elstoc: but now this function doesn't prepend the accel to the list any more

src/views/darkroom.c
```diff
@@ -885,7 +885,7 @@ static void dt_dev_change_image(dt_develop_t *dev, const uint32_t imgid)
       dev->iop = g_list_remove_link(dev->iop, g_list_nth(dev->iop, i));
 
       // we cleanup the module
-      dt_accel_disconnect_list(module->accel_closures);
+      dt_accel_disconnect_list(&module->accel_closures);
       dt_accel_cleanup_locals_iop(module);
       module->accel_closures = NULL;
```
elstoc, line 890: possibly don't need ``module->accel_closures = NULL`` any more

rabauke:
@elstoc 

> However, I'm not sure that the changes you've proposed to `dt_accel_disconnect_list` are technically necessary since your proposal is functionally equivalent to what went before (it clears the list and sets the appropriate pointer to NULL - so if you just added `table->accel_closures = NULL` to the thumbtable function you'd be fine and consistent with the other calls).
> 
> This is probably fine for a code tidy-up but I don't see why the current code shouldn't work.

Regarding `dt_accel_disconnect_list`, I do not claim that the function itself has a bug. It has just a poor design. You are right, it is more clean-up.

With the current design, the `GList` pointer that is passed to `dt_accel_disconnect_list` points to an invalid object when the function returns. In the following, the caller must ensure that the pointer is set `NULL`, which indicates a empty list, or it must be ensured that the pointer is never dereferenced. A good api design would not put such a responsibility on the caller side because this is a common source of programming errors.

elstoc:
Agreed I think the changes to ``dt_accel_disconnect_list`` make sense. Just that since we're coming up on the release, I wonder whether they're worth making for 3.2 - the fixes definitely are needed though. Up to @TurboGit I suppose.

TurboGit:
I'm testing that and plan to add this to 3.2 indeed, such a bad usage may certainly lead to unexpected behavior. Fixes can go to 3.2 up to last minute depending on the severity of course!

src/gui/accelerators.c
```diff
@@ -534,14 +534,15 @@ void dt_accel_connect_lua(const gchar *path, GClosure *closure)
...
+  GSList *list = *list_ptr;
+  *list_ptr = g_slist_prepend(list, accel);
```
TurboGit, line 545: The two lines above are fully equivalent to:

`*list_ptr = g_slist_prepend(*list_ptr, accel);`

rabauke: I introduced the variable `list` just for explicitness and clarity. It is not really needed indeed. I will revise my pull request following your suggestion.

TurboGit:
Ok, thanks !

PR #5648, opened 2020-06-27
clipping: retain orientation when changing aspect ratio

elstoc:
when choosing a new aspect ratio from the drop-down, don't change back
to landscape if user has previously selected a portrait orientation


TurboGit:
Works for me! Thanks.

PR #5650, opened 2020-06-28
Added basic support for Sony Alpha 9M2 (ILCE-9M2)

nixwieweg:
Values are the same as ILCE-9 but are taken from exiv2 after DNG conversion.


LebedevRI:
Thank you for the contribution!

Issue #5655, opened 2020-06-28
parametric masks: "set range" color picker changes module parameters


elstoc:
In the following example, I have discarded history and set the exposure slider in the exposure module to 0. I have switched on parametric masks

![Screenshot_2020-06-28_14-43-13](https://user-images.githubusercontent.com/9555491/85949271-b9db5180-b94d-11ea-948e-a5b261caca2e.png)

I click the 'set range from area' color picker and choose an area. I can't take a screenshot of this because as soon as darktable loses focus (because I've switched to the screenshot app) the following happens:

![Screenshot_2020-06-28_14-46-29](https://user-images.githubusercontent.com/9555491/85949375-438b1f00-b94e-11ea-96d5-5849de157776.png)

Somehow the color picker has caused the exposure to be changed to 2.03EV and the black level to be changed as well.

Similar issues in #5654 with the other color picker.

Issue on 3.1.0+2376~g40f2f94e3 on ArchLinux with and without openCL.

jade-nl:
Confirmed.

Not as severe as shown by elstoc, but this should happen at all.

**Platform**
* darktable        : 3.1.0+2383~gc9ca6258d
* OpenCL installed : Yes (nvidia-opencl-common 440.82-1~bpo10+1)
* OpenCL activated : Yes
* Graphics card    : GeForce RTX 2060 SUPER7979 MiB
* Graphics driver  : nvidia (440.82)
* OS               : Linux (kernel 5.4.0-0.bpo.4-amd64)
* Distro           : Debian Buster (rel. 10.4)

**Additional context**
* Self compiled : Yes.
* Compilers     : gcc 8.3.0 - clang 10.0.1 - LLVM 10.0.1
* Cmake         : 3.13.4
* Gtk           : 3.24.5-1
* Glib          : 2.58.3-2
* Sqlite        : 3.27.2
* Optimization  : None

TurboGit:
I cannot reproduce, any step by step reproducer ? Maybe this needs a very specific set of actions ?

elstoc:
It's pretty consistent for me. I've tried a few images. After selecting a range area, as soon as you alt-tab away from darktable or try to adjust the right-hand panel width the exposure and black level sliders change.

elstoc:
The exposure settings look just like if I'd used the color picker on the main body of the exposure module. I suspect darktable is getting confused about which color picker is which.

TurboGit:
[...]

Issue #5654, opened 2020-06-28
parametric mask: issues with color pickers and narrow side panel


elstoc:
When the right hand side panel is expanded fully (so that all controls are visible) the parametric mask color pickers seem to work pretty well.

However, when the panel is narrower, such that some of the mask's channels (e.g. 'g R G B H S L') are not visible there are a number of strange side-effects. The issues seem to start just before the 'B' channel becomes hidden (when shrinking the panel width).

In the following example, I've just discarded history using the scene-referred defaults and then selected the parametric mask button:

![Screenshot_2020-06-28_14-31-44](https://user-images.githubusercontent.com/9555491/85949025-1fc6d980-b94c-11ea-9cce-85973bc677ec.png)

I select the left-hand-side color picker in 'point mode' and the following happens:

![Screenshot_2020-06-28_14-32-44](https://user-images.githubusercontent.com/9555491/85949045-438a1f80-b94c-11ea-9d96-a32233543ced.png)

Obviously this has done some very funky things to the image itself. However, the less obvious feature is that it's also subtly changed the display of the module itself.

This issue is more obvious if I compress the side panel still further to fully hide the blue channel selector. Before:

![Screenshot_2020-06-28_14-34-21](https://user-images.githubusercontent.com/9555491/85949078-84823400-b94c-11ea-8457-062effbbea8b.png)

After selecting the color picker:

![Screenshot_2020-06-28_14-35-07](https://user-images.githubusercontent.com/9555491/85949088-9532aa00-b94c-11ea-898a-3097bea53e23.png)

Now the Blue channel is suddenly visible. Selecting the picker should not do this.

I suspect these issues are related, hence raising them together in one issue.

Tested on 3.1.0+2376~g40f2f94e3 with ArchLinux. Both issues occur regardless of whether openCL is enabled.

jade-nl:
Confirmed and not limited to the above actions.

I can see this happening when:

* right side panel is wide enough for all text/options to be seen,
* use (left) colour picker to pick an area,
* resize the right hand panel (smaller/bigger doesn't matter) **OR** move cursor outside full screen (to second monitor in my case).

I cannot provide a screenshot (going outside the darktable area creates the same distortion/effect)

**Platform**

* darktable : 3.1.0+2383~gc9ca6258d
* OpenCL installed : Yes (nvidia-opencl-common 440.82-1~bpo10+1)
* OpenCL activated : Yes
[...]

PR #5657, opened 2020-06-28
po/fix exposure

TurboGit:


elstoc:
Yep that resolves #5655 for me. Thanks @TurboGit.

elstoc:
I've worked out what the other issue is now (with the module layout changing when toggling a picker). It's because the display on the input/output sliders needs to change to accommodate an additional number (in brackets) after the input/output labels - is that the average of the area or value of the point selected? So ultimately, I suppose that's not an issue then, and this PR should close both #5654 and #5655.

TurboGit:
@elstoc : thanks for feedback, now merged.

TurboGit:
BTW, can you double check that #5654 is fixed ?

elstoc:
I can no longer reproduce the main issue in #5654 so it can be closed (I'll close it now)

dterrahe:
While looking at the code that was fixed here, I was wondering if this code ever does anything useful.
https://github.com/TurboGit/darktable/blob/e265124f1cd4060857bb3ba20f2c73236bad8486/src/iop/exposure.c#L839-L850
It looks at the color picker while drawing, but the picker only sends updated numbers when the region selection is completed and by that time it also sends a signal that picks up the results. So basically, I'm wondering if there is a need for both the callback color_picker_apply _and_ the code in draw.

The code in draw _also_ calls exposure_set_black, but this could move to the callback.

PR #5661, opened 2020-06-29
Notebook and section label changes for narrow panels

dterrahe:
This makes notebook tabs and section labels in iop modules more compressible. When shrinking the right side panel, more of the module will remain visible. In the end, the blending icons will not fit anymore and a horizontal scrollbar will appear.  It might make sense to increase the minimum width of the panel a little to avoid this scenario. Alternatively, maybe it is possible to disable the scrollbar, which presumably would make gtk enforce the minimum width required.

Obviously things get less readable as everything gets scrunched together. This is especially the case with the blending mask slider value labels. To read those one will simply need a wider panel. But at least with these changes the panel no longer jumps around when enabling or disabling the color pickers.

There is also a bit of refactoring and cleanup around notebooks in general.

In blend_gui.c (the calls to) _blendif_print_digits_... are a bit weird, but I didn't want to touch that now as it may well have been finely tuned to behave as it does.

dtgtk_justify_notebook_tabs is no longer needed/used. Should I remove it?

### DIFF
--- a/src/develop/blend_gui.c
+++ b/src/develop/blend_gui.c
@@ -300,27 +300,27 @@ static inline int _blendif_print_digits_ab(float value)
 
 static void _blendif_scale_print_L(float value, char *string, int n)
 {
-  snprintf(string, n, "%-4.*f", _blendif_print_digits_default(value), value * 100.0f);
+  snprintf(string, n, "%-5.*f", _blendif_print_digits_default(value), value * 100.0f);
 }
 
 static void _blendif_scale_print_ab(float value, char *string, int n)
 {
-  snprintf(string, n, "%-4.*f", _blendif_print_digits_ab(value * 256.0f - 128.0f), value * 256.0f - 128.0f);
+  snprintf(string, n, "%-5.*f", _blendif_print_digits_ab(value * 256.0f - 128.0f), value * 256.0f - 128.0f);
 }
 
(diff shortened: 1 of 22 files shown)


dterrahe:
Resolves the issue mentioned here https://github.com/darktable-org/darktable/pull/5657#issuecomment-650811752
Coincidentally I had been working on this at the same time.

dterrahe:
@TurboGit a bit surprised to see you marking this for 3.2
There are changes all over the place, but local and easily testable.
Maybe the layout changes could be controversial (people might have got very used to how the R/G/B switches clustered together on the left side) but this way they are better behaved, so...

I had to remove the spaces padding the translatable strings. Wouldn't that break the work of the translators this late in the cycle? Maybe in most cases the shorter versions, just "R" instead of "  R  " already exist and it is not an issue at all.

TurboGit:
Clearly a mistake, I certainly wanted this for 3.4 and probably clicked on the wrong button indeed.

aurelienpierre:
Could you provide a before/after screenshot @dterrahe ?

dterrahe:
image image

image image

aurelienpierre:
> dtgtk_justify_notebook_tabs is no longer needed/used. Should I remove it?

Let's wait a bit, then probably.

PR #5665, opened 2020-06-29
nlmeans: add module parameter descriptions for translation

rabauke:
Working on the German translation I found this issue. (Strength was not translated).


dterrahe:
`g->strength = dt_bauhaus_slider_from_params(self, N_("strength"));` would work also, with less duplication/chance of typos.

I'll go through and add these everywhere where needed. PR later today.

TurboGit:
Better fix put in place. Closing this one.

PR #5667, opened 2020-06-29
duplicate: regenerate preview surface if preview dimensions have changed

### LINKED ISSUE #5664: duplicate manager not updated when image size changes
If I edit a duplicated image with both the left and right panel expanded and then click on the duplicate image in the duplicate module, that image displays correctly:

Screenshot_2020-06-29_22-02-50

If I collapse the right hand panel and click the duplicate image again, it retains its former size:

Screenshot_2020-06-29_22-03-06

I would expect the second attempt to show the duplicate with an updated size.

3.1.0+2392~gf21a98b12 on ArchLinux. Issue occurs both with and without openCL.

elstoc:
Resolves #5664

### DIFF
--- a/src/libs/duplicate.c
+++ b/src/libs/duplicate.c
@@ -44,6 +44,8 @@ typedef struct dt_lib_duplicate_t
   gboolean busy;
   int cur_final_width;
   int cur_final_height;
+  int32_t preview_width;
+  int32_t preview_height;
   gboolean allow_zoom;
 
   cairo_surface_t *preview_surf;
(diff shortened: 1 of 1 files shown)


src/libs/duplicate.c
```diff
@@ -249,8 +251,12 @@ void gui_post_expose(dt_lib_module_t *self, cairo_t *cri, int32_t width, int32_t
 
   // if not cached, load or reload a mipmap
   int res = 0;
-  if(d->preview_id != d->imgid || d->preview_zoom != nz * zoom_ratio || !d->preview_surf)
+  if(d->preview_id != d->imgid || d->preview_zoom != nz * zoom_ratio || !d->preview_surf 
+      || d->preview_width != width || d->preview_height != height)
```
TurboGit, line 255: you are comparing d->preview_width but this is never set before.

The two new fields preview_width & preview_height must be initialized to 0 before. Or maybe this is already the case ? g_malloc0() ?

elstoc: I can do this, but as it stands, the first time this function is run for an image, d->preview_surf will be NULL. So the additional statements in the 'if' won't get run and the values will be set to the input height and width.

Where should it be initialised? In gui_init()?

elstoc: Can't hurt so I've changed to initialise in gui_init

src/libs/duplicate.c
```diff
@@ -487,6 +493,8 @@ void gui_init(dt_lib_module_t *self)
   d->imgid = 0;
   d->preview_surf = NULL;
   d->preview_zoom = 1.0;
+  d->preview_width = 0;
+  d->preview_height = 0;
```
TurboGit, line 497: Perfect! That's exactly where it must be initialized.

TurboGit:
Quite safe, touch a single lib. Let's go for 3.2. Thanks.

PR #5668, opened 2020-06-30
Better mixing of Shadows & Highlights fixing negative L* values increasing in brightness

### LINKED ISSUE #5625: Shadows and Highlights with negative values on shadow unexpected result
When pulling the "shadows" slider into the negative the shadows become bricht suddenly on some values as if some kind of overflow condition occurs.
(See this short video clip for demonstration https://youtu.be/Sq2mscTcugI )

1. Reduce brightnes of image in exposure so parts of the image are below the black level
2. go to "shadow and highlights" and slowly pull the slider to the left while observing the shadows in the image

**Expected behavior**
The shadows shold stay dark and more parts of the image should become dark or black

**Screenshots**
Video on https://youtu.be/Sq2mscTcugI

**Platform:**
 - Darktable 3.1.0+2136~g9a8da0352
 - OS: debian 10.4
 - Tested with OpenCL and without OpenCL. Result is the same

hatsunearu:
Negative L* values are invalid to begin with, and right now when there is a negative mask and a negative input value, they multiply together to form a positive value, which caused weird behavior, causing things that are meant to be darker to become brighter. Fixes #5625.

Not sure if this is the best way to do it, but since math on negative L* values are invalid to begin with, I'd say we just use this method to be "less wrong".

edit: that space in line 468 is due to clang-format.

### DIFF
--- a/data/kernels/gaussian.cl
+++ b/data/kernels/gaussian.cl
@@ -282,7 +282,7 @@ overlay(const float4 in_a, const float4 in_b, const float opacity, const float t
     float optrans = chunk * transform;
     opacity2 -= 1.0f;
 
-    a.x = la * (1.0f - optrans) + (la > halfmax ? lmax - (lmax - doublemax * (la - halfmax)) * (lmax-lb) : doublemax * la * lb) * optrans;
+    a.x = la * (1.0f - optrans) + (la > halfmax ? lmax - (lmax - doublemax * (la - halfmax)) * (lmax-lb) : doublemax * sign(la) * fabs(la * lb)) * optrans;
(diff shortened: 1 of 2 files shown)


TurboGit:
One thing about this is that changing process means that old edits are not preserved. I'm not against a better algorithm (didn't check it BTW) but we really want to keep previous edit intact meaning that we need to keep old algorithm for old edits and introduce a new one. This means adding a version field in params struct and an entry into legacy_params() to migrate from previous version.

TurboGit (changes requested):
support legacy params.

TurboGit:
Any news on this ?

hatsunearu:
Hi sorry, I'll take a look at this again once I get some more time...

PR #5672, opened 2020-06-30
duplicate: slightly improve pixel shifting

elstoc:
slight changes to when rounding is performed improves but does not
completely eliminate pixel shifting when using the duplicate module

### DIFF
--- a/src/libs/duplicate.c
+++ b/src/libs/duplicate.c
@@ -257,7 +257,7 @@ void gui_post_expose(dt_lib_module_t *self, cairo_t *cri, int32_t width, int32_t
     d->preview_width = width;
     d->preview_height = height;
 
-    res = dt_view_image_get_surface(d->imgid, img_wd * nz, img_ht * nz, &d->preview_surf, TRUE);
+    res = dt_view_image_get_surface(d->imgid, ceilf(img_wd * nz), ceilf(img_ht * nz), &d->preview_surf, TRUE);
 
     if(!res)
     {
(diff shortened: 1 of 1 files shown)


elstoc:
I've reverted the changes to ``view.c`` as they were causing a black line to display at the bottom of the image. Interestingly, if I make the following change, the pixel _shifting_ is eliminated entirely when viewing the full image, though when zoomed the image appears incorrectly offset - I'm not 100% sure how the zoomed image is generated so don't know how to fix this:

```
  float img_wd = dev->pipe->output_backbuf_width;  //dev->preview_pipe->backbuf_width;
  float img_ht = dev->pipe->output_backbuf_height; //dev->preview_pipe->backbuf_height;
```
I don't really know what the difference is between these figures but it might provide a clue how to completely eliminate the shifting.

AlicVB:
I've done some test with your last finding.
I'm starting to wonder if the problem comes really from duplicate...
For a given image (tested with tif in entry to avoid any raw sizing problem) at "zoom to fit" size :
```
initial width = 5472
initial height = 3648
```
=> ratio = 1.5
(same values when exporting)

```
dev->preview_pipe->backbuf_width = 675
dev->preview_pipe->backbuf_height = 450
```
=> ratio = 1.5

```
dev->pipe->output_backbuf_width = 1732
dev->pipe->output_backbuf_height = 1155
```
=> ratio = 1.4995671 (height should be 1154 to get 1.5 ratio)
(same values when exporting with width = 1732 and height = 0, resampling quality doesn't change anything)

So I suspect a rounding problem somewhere in the main pipe when resizing is active (no iop activated except mandatory input/output color profiles)

elstoc:
See also #5701 for issues when the duplicates are differently-cropped. Most of my findings have been from playing around with the code and I'm still not really sure how it all works. I'm just hoping that some of my results might provide some inspiration!

The biggest issue for me (which I can't find anything to resolve) is the strange effect when zoomed in, whereby rows and/or columns of pixels are entirely removed from the image. See the zoomed-in screenshots in #5664.

AlicVB:
So, I've proposed a PR to fix #5701 which is a completely different (and way more simple) issue. See #5713 

For the issue here (which is certainly quite the same as your screenshot in #5453 : 

I really think that the problem comes from the main pipe size computation. If I'm right, this is not the duplicate preview which is shift, but the "normal" darkroom image : 
In my previous test, with an exactly 1.5 image ratio in entry, and with absolutely no iop active, we should get an 1.5 image ratio at the end. But somehow, this is not always the case when output size is not the same...

Sadly, this will be way more difficult to fix, and touch a very delicate area...

TurboGit:
I think this is OBE, right? Should this be closed ?

elstoc:
This only really had only very minor improvements (which I only did by playing around with rounding in different ways) so I don't mind closing it. The general issues are still there with the duplicate module but, as @AlicVB says there are no nice solutions.

PR #5676, opened 2020-06-30
po/fix xmp import

TurboGit:


jenshannoschwalm:
I tested this pr and i think it's good. Also fixes a grouping issue with duplicates ...

TurboGit:
Great! I have merged this after adding some more comments to ensure we won't introduce back this issue.

PR #5678, opened 2020-06-30
image: duplicate version caching bugfix

### LINKED ISSUE #5674: duplicate of an image can have version number of 0
**Describe the bug**
It is possible to duplicate an image such that the version of the duplicate is 0.

**To Reproduce**
1. Select a freshly imported image
2. Go to darkroom view
3. Expand "duplicate manager"
4. Click "create a duplicate with same history stack"
5. Expand "image information" -- verify that you are now editing a duplicate with version 1
6. Click "create a duplicate with same history stack"
6. You are now editing a third duplicate. In "image information" the version is 0.

**Expected behavior**
The version of the second duplicate should be 2.

**Screenshots**
image

**Platform:**
 - Darktable Version: master aef614f0e3caa8bfff91559b7a50df24e4e7357d
 - OS: Debian bullseye/sid, Gnome 3.36.3
 - OpenCL: yes

**Additional context**
 - Have reproduced with a CR2 and a JPEG
 - Have reproduced with an empty/new config-dir
 - This bug has been present at least for the past few weeks

dtorop:
Set duplicate image version in the DB before the cache entry for the image is first allocated. Otherwise, image version in the cache will not reflect the version in the DB.

Also, the change timestamp and group ID for the original image were being written.  Instead, update these for the duplicate.

Fixes #5674.

### DIFF
--- a/src/common/image.c
+++ b/src/common/image.c
@@ -905,15 +905,8 @@ int32_t _image_duplicate_with_version(const int32_t imgid, const int32_t newvers
     sqlite3_step(stmt);
     sqlite3_finalize(stmt);
 
-    // make sure that the duplicate doesn't have some magic darktable| tags
-    if(dt_tag_detach_by_string("darktable|changed", newid, FALSE, FALSE)
-       || dt_tag_detach_by_string("darktable|exported", newid, FALSE, FALSE))
-      dt_control_signal_raise(darktable.signals, DT_SIGNAL_TAG_CHANGED);
-
-    /* unset change timestamp */
-    dt_image_cache_unset_change_timestamp(darktable.image_cache, imgid);
-
     // set version of new entry and max_version of all involved duplicates (with same film_id and filename)
+    // this needs to happen before we do anything with the image cache, as version isn't updated through the cache
     const int32_t version = (newversion != -1) ? newversion : max_version + 1;
     max_version = (newversion != -1) ? MAX(max_version, newversion) : max_version + 1;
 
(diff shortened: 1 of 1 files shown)


src/common/image.c
```diff
@@ -935,9 +928,18 @@ int32_t dt_image_duplicate_with_version(const int32_t imgid, const int32_t newve
 
     g_free(filename);
 
-    const dt_image_t *img = dt_image_cache_get(darktable.image_cache, imgid, 'r');
+    const dt_image_t *img = dt_image_cache_get(darktable.image_cache, newid, 'r');
```
TurboGit, line 931: This looks wrong to me. You are setting the group-id to be the one of the new image (the duplicate itself). The group-id of the artifact should be the group of the main image.

dtorop: That makes sense. I'll fix/re-test...

TurboGit: Note that the issue may already be fixed with my latest work on duplicate.

dtorop: I'm just looking at this. This PR fixes something different, that the calls to dt_tag_detach_by_string() and dt_image_cache_unset_change_timestamp() will call dt_image_cache_unset_change_timestamp() which load the version # in from the DB. So it's important to set the version # before it gets cached. (Or an alternative would be to make the version # writeable via the cache.)

Also dt_image_cache_unset_change_timestamp() should be working on the new image, even if I was wrong about the group ID.

dtorop:
I fixed https://github.com/darktable-org/darktable/pull/5678#discussion_r448252127 and rebased onto 08c1dcd756079487df79b300893dda7963ebc1f6. This PR should still be necessary to fix #5674, which I can otherwise replicate in current git master.

TurboGit:
Ok, works for me. Thanks.

PR #5683, opened 2020-07-01
iop layouts

AlicVB:
here is a first draft proposal for a way to let users set how they want to organize modules in darkroom.
this is another approach of PR #5322 

### why ?
because everyone as different workflow and needs
Even the same user doesn't need the same modules and groups  for black&white, portrait, architecture, ...

### Features
This extend a little the presets capacities to propose a gui to tweak : 
- the groups they wants.
- the icons and names of each groups
- the modules that should be present in each groups

other than that
- order of modules stay in the pipe order (that's an evidence, but I prefer to remind it)
- active group is mandatory, only the next ones can be choose
- a module can belong to as many groups as we want

### state
currently, this more a proof of concept than anything else : 
- the gui is horrible (but functional)
- there's lot of glitch and not functional things
- there's certainly lot of bugs

### shipping presets
Actually, default presets are just some random things which are not useful on purpose : they are only here to show its possible
So we need to define carefully as great set of presets and the default one. They can be based on actual modulelist presets and on great PR #5322 proposal. I'm not sure I'm the best one to define them... maybe @aurelienpierre ?

### upgrade
It's actually possible and quite simple (code wise) to use previously user-defined layout for creating a new preset on the fly so user won't see any change here.
But this would mean that we have to keep "ad vitam" unused module properties... 

### to be done
- removing of modulelist lib, which becomes useless... 1 line saved :)
- reuse nice icons of #5322 
- define nice default presets

### Questions : 
- do we want something like that ?
- do we prefer to "drop" previously user-defined layouts and "drop" some iop properties too ? or the contrary ?


AlicVB:
and some screenshots...
the "new" modulegroup : module
the preset manager : presets
the groups manager : groups

As said, gui are horrible (but functional), but I don't want to waste time on that point before we decide or not to go further.

Nilvus:
> As said, gui are horrible (but functional), but I don't want to waste time on that point before we decide or not to go further.

I'm asking me how that was the first thing I notice :laughing:

I really like @aurelienpierre proposal but why not having some possibilities to tweak a little more. But we already have groups presets so if that is just a way to manage existing groups presets, it's ok. But no more for that. I'm for simplify module groups like in #5322. And the most important thing, and an answer to your last question: iop properties should be the first thing to take in account on group and order (as it is actually).

What I would see about those parts is the #5322 PR from @aurelienpierre and a thing to mix (and so replace) presets groups and more modules in a module editor to just edit/set modules user show on existing groups (those set in #5322).

AlicVB:
Well, in fact, it's "just" a more powerfull "more module" lib, but integrated as presets for modulegroup lib (the line with the group icons and the search box) so
- it remove the "more module" lib
- it remove the need of favorite group (as you can choose what you want to put in each group you define)
- about PR #5322 , the "vanilla" groups and all the "more modules" presets, the idea is to provide them as presets. And yes, why not proposing @aurelienpierre layout as the default preset.

> What I would see about those parts is the #5322 PR from @aurelienpierre and a thing to mix (and so replace) presets groups and more modules in a module editor to just edit/set modules user show on existing groups (those set in #5322).

not sure to understand well, but I think it's exactly that, with the possibility to define the number of groups in addition.

elstoc:
I agree with this. Different people are always going to want different groupings and we'll never come up with a standard that everyone likes. I think this should be done in conjunction with #5322 (to define the default groups).

I don't see a reason not to have favourites as a default group as well though (but not a special one, so the user can delete it). So the default would be the three groups defined in #5322 plus favourites. If a user already has favourites set up, these are automatically added to that group.

One other thing to consider is keyboard shortcuts. Currently if a user chooses to expand/collapse a module using a keyboard shortcut it's expanded in one of three places, in order: (1) The current group, if present (2) The favourite group, if present (3) The module's default group. If a module can be in multiple groups, how we'll need to decide which group to expand the module in. A simple rule like start with the left-most group and move to the right until you get a match would probably do it.

aurelienpierre:
I have the feeling that we are making things more complicated for no benefit. I see that as a dictionary where you give users the power to use a custom letter order… The point is to have a robust index to find stuff, not to interface an extra relational database in the mix. You want that to be reliable, robust and stupid simple. If someone wants Z right after K, well, deal with it.

If users know the name of the module they are looking for, they will likely look it up directly in the search form. If they don't, they need some register/table of contents for the features of the soft. #5322 proposes a thematic table of contents, consistent because modules should be used in the GUI tab order for minimal overhead, and this order happens to split technical from artistic modules. 

Letting them play with the register will not make it better if they need guidance through the GUI. They will only get more options to shoot themselves.

As usual, my only question will be "what problem are we trying to solve here ?". Giving users more options only means we failed to find the best, and there are not hundreds of rational and consistent workflows because technics already set some choices for you.

AlicVB:
@aurelienpierre : 

> what problem are we trying to solve here ?

Giving "power users" the choice to reorganize the layout as they want.
See in #5322 how many different opinion you get from what we can call "power user". Each ones as different need. Not because they are doing wrong things (well... maybe...) but because there's 2 steps involved here : 
- the "technical part" for correcting as you say in one of your post post. It's not the problem, as the technics "set choice for us" here.
- the "artistic" part for creating (which can use many "technical" modules too) It's the more hungry in term of number of modules and depends from everyone needs, tastes, etc... even if there's still some rules to respect. And more difficult, it even varies for the same user depending of the image subject, the delays, etc...


To resume, I think there's 2 approach here : 
- classifying by order of use, which is needed for the technical part
- classifying by usage, which is needed once you have done the first part

In real life with real tools, they are antagonist and of course the only possible one.
But here they can be combined once you allow module to belong to multiple groups.


Speaking for me, (and I admit that I'm not a representative user, so forget that point if I'm alone) I would make
1. a first technical group with filmic, exposure, crop/rotate, lens, denoise 
2. a second technical group with whitebalance tone equalizer color balance tone curve ...
3. a group with b&w modules 3dluts, channel mixer, colorbalance, colorzone
4. a group for portrait specific retouch liguify equalizer
5. a group for landscape specific lens, ashift, haze removal, ...
6. a group for "finalize" frame, watermark

and presets like "portraits with 1,2,3,4 landscape with 1,2,5 etc... which can activated by a simple shortcut as they are standard presets

Speaking of the search box : I really think it can be great, but as many other, I just fail to remember what to type... maybe it's because I mixed french and english interface, certainly it's my age...

TurboGit:
> As usual, my only question will be "what problem are we trying to solve here ?". Giving users more options only means we failed to find the best,

Maybe because there is no best option and that many people have different views on what the workflow and tool organization should be.

On my side I see this has a good project. We had so many discussion at how to properly present the modules and we have seen that there is not a single point of view. The many proposals are probably all valid and this PR will make everyone happy. I'm really all for moving this effort ahead!

elstoc:
> Giving users more options only means we failed to find the best

As has been said, the fact that there is significant difference of opinion about how to define "the best" should tell you everything. There is no 'best' as too many of the modules are subjective. Even in #5322 the third group is basically a grab-bag of "things I couldn't categorise".

aurelienpierre:
>     1. a first technical group with filmic, exposure, crop/rotate, lens, denoise
> 
>     2. a second technical group with whitebalance tone equalizer color balance tone curve ...
> 
>     3. a group with b&w modules 3dluts, channel mixer, colorbalance, colorzone
> 
>     4. a group for portrait specific retouch liguify equalizer
> 
>     5. a group for landscape specific lens, ashift, haze removal, ...
> 
>     6. a group for "finalize" frame, watermark

But that's a can of worms, since many modules will fit several groups, like lens is just basic technical module, retouch or equalizer can be used in many situations and genres, so this deals with arbitrary categories by adding more arbitrary categories.

> There is no 'best' as too many of the modules are subjective. Even in #5322 the third group is basically a grab-bag of "things I couldn't categorise".

No, they are not. There are the "reverting shit done by the camera", "grade pixels based on their input intensity" and "grade pixels based on their locality". Modules might seem subjective if you don't know their transfer function, what problem they solve and what assumption they rely on. 

> Maybe because there is no best option and that many people have different views on what the workflow and tool organization should be.

The workflow should be driven by the pipeline. The fact that people have different views here is the problem, it means that there is some image processing education to be made before even speaking of UI design. Enforcing a GUI that naturally encourages a rational workflow by drawing a sane path to follow is key here. 

There is a flow chart or a block diagram going on in the pipeline, which is a stack of filters. A power user might want to see that. A regular user might want a GUI that hides the technicality of the blocks while still keeping the natural path they define. 

I believe the GUI should be based on the pipeline, especially since the pipeline is re-orderable but we don't have a nodal editor or a real flow chart to graphically display said pipeline. We, devs, know that flow chart and are able to derivate a consistent workflow from that knowledge, based on image processing knowledge.

Now the question is how to translate that knowledge into a GUI so users can come to the same workflow without having to learn C, pass a masters degree in maths, and dedicating 2 years of their life to make sense out of equations.

For now, inside tabs, modules are ordered as the pipe is, but tabs themselves don't mean nothing pipe-wise.

AlicVB:
with last commit, we now have more useful presets : 
- the default one is copied from @aurelienpierre work in #5322 (without the icons... to be done after)
- another one with the actual disposition

note that shown modules are the ones defined for a new fresh install of dt. I think that we should review them too

AlicVB:
@aurelienpierre : 
I really think that all what you say can take place in this PR : we just need to set some nice presets. And as I say, you're probably the better one for that (that's why I've put your #5322 proposal as default preset here)

johnny-bit:
As a dude who loves much configuration possible and done several VERY configureable apps and got schooled in UI/UX I just need to express: while I love this as a general solution, there HAS to be very good and very generic (like Aur's) preset that will undoubtly fit 80-90% of workflows and everything needs to be done in assumption that most users won't change it or will have hard time switching/adjusting. And it absolutelly has to have "form follows function, communicates usage and guides for better workflow" :/

sergei18500:
Hello,

The text below is automatically translated from french by Deepl.
I do not guarantee the translation of the spirit of the FRENCH text.
I read the discussions using the same process and I am not sure I assimilate everything.

*****
ubuntu 20.04 dt master / windows dt 3.0 .2 ( dt learning since about 1 year )
the only other post-processing software I've used is lightzone (no longer maintained) and I know a little bit about GIMP.
*****

I take the liberty to bring another vision, that of a simple lambda user, and which joins for part what has been written previously ( # 5052-5314-5322 ) and in particular :

### What problem are we trying to solve here?

### What is the objective? 

As I understand it, this debate is looking for 'a compromise between ergonomics and functionality' and that it has a place in the future development of darktable that you will create.

### What problem are you trying to solve here, but more importantly, for which users?

The 2019 survey of darktable users is certainly not a reflection of reality.
(you just have to look at the profiles of the subscribers like on darktable.fr and their comments, I think it's the same on the other forums)

I'm thinking more along the lines of this typology:
The real beginner in post-processing who wants something free to try and learn! ( the only alternatives to the paid ones are darktable and rawtherapee )
More experienced users who want to leave Adobe!
The unconditional, advanced users of darktable for several years.
I'm not talking about the developers who by their deep knowledge of darktable's insides have a 'polluted' vision.

I'm neither a developer nor a professional of photography and image, I was looking for a user-friendly, powerful tool (and darktable is on this second point), and which allowed me to discover the software simply, to progress progressively in my learning of the possibilities of each module.
I have 2 flaws (62 years old and a basic technical training) that I still try to understand and that's why I turned to darktable.

I watched many videos of AurelienP on the technical explanation of some modules, it is not always within the reach of the average user. 
I am a member of a small photo club, there is a doctor, a delivery man, craftsmen, salesmen, teachers (literary or technical), a chemist, retired people from various sectors etc.. I don't see myself selling their darktable 'sell' on the basis of technical modules and primary or secondary corrections!!!, knowing that even in your discussions the borderline doesn't clearly exist. 

### An opinion which is only mine, therefore is worth 1 ..........:

In the darkroom since that's the subject.
1) I find very practical the first icon (display all modules / display active modules )
2) I find the second icon ( show favorites ) very interesting. You can put there the 'x modules' you use all the time or those you want to test, and it's very easy to adapt to your personal evolution of darktable use. ( When I first started working on darktable, I only worked like this )
3) Concerning the third icon (basic modules group), the name is very very ambiguous, and can be misleading for a beginner.

### Remark :
the 3 groups evoked by AurelienP 
	technical modules: all technical issues related to lenses (sensor, lenses and color spaces) such as profiling, range and tone mapping, reconstruction of damaged parts, perspective and lens correction, etc.
    	grading: everything concerning primary (corrective) and secondary (creative) subjective corrections of colours and tones,
  	(Special) effects: retouching, liquefaction, flowering, sharpness and others.

### These notions, although fundamental and right, only have real meaning for initiates.

Me what I would like, for example, and in that order:

This group of modules allows you to process the exposure and tones, ( Exposure / Tone )
_tone curve,rvb film,rvb levels,rvb curve,color balance,tone equalizer,exposure,levels,basic curve,basic adjustments,gradient filter,global tone mapping, etc._
This one allows to treat the sharpness ( Sharpness )
_lens correction,local contrast,sharpness enhancement,noise reduction (avg. non-local),high-pass filter,low-pass filter,haze suppression,noise reduction (bilateral),noise reduction (profile),RAW noise reduction, etc._
This one the modification of the colors ( Color )
_monochrome,colorize,color correction,channel mixer,color table,white balance,color reconstruction,3D lute,doctor nega,color zones,vibrance,color contrast,color mapping,chromatic caberrations, etc._
This one manages the retouching ( Retouch ) 
_retouching, spot correction, crop and rotate, perspective correction_
This one allows 'special effects' ( Effect )
_liquefy,watermark,decorative frame,homogenization,vignetting,partial turn,Orton effect (soften),grain,low light,back light,fill light,fringe removal_
This groups together the functional modules ( Functional )
_output color profile,input color profile,orientation,de-mapping,highlight reconstruction,black/white point RAW,input profile correction,pixel scaling,pixel rotation,hot pixels,invert_
( bubble info could detail the group in more or less detail )

### This proposal does not at all have the rigour of the AurelienP vision, will shock some people and does not go in the direction of reducing the groups of modules.
8 icons (all modules / active modules) ( shortcut ) ( + 6 groups of modules )

### On the other hand, it allows:
- to "quickly catch" the neophyte who discovers darktable. It allows him to have a more 'concrete' approach in relation to the visual of his starting photo (and what he has to work on) and the group where he will find the module to process this photo.
- doesn't prevent the curious to progressively deepen their knowledge about what each module does (help, tutorial, youtube ...).
- limits the number of modules per group.
- maintains highly adaptive customization via the 'shortcut' icon
### and..:
- keeps all the options and settings we have in the 'more modules' option. ( note : I have never used the workshops or groups of modules that are there )
- no change to the processing pipeline ( users can still voluntarily change the order as currently possible in v3 )

### That's why the 'completely configurable' proposal as presented by AlicVB would suit me perfectly. ( from what I understood its UI would be simple )

By default there will be the current module groups so that the one who starts has a base.
After each one following his learning, his methodology, his own workflow, structure his groups of modules add, modify, delete.

The must would be that this part of configuration is in a separate file. It could be exchanged between users.

### THIS IS JUST A SHARING WITH NO OTHER PRETENTION THAN TO GIVE AN OPINION ON AN ALWAYS IMPORTANT PART WHICH IS THE DESIGN, THE EASE OF ACCESS FOR THE USER AND THE VULGARIZATION.


Last comment, it would seem obvious to developers and advanced users that some modules are 'obsolete' and remain for some compatibility with photo processing related to 2.x....
Couldn't these modules have another font, or a * in front of the name. In the tooltip a comment ' use xxxx') ?
This could, why not, be an option in the global preferences (display 'obsolete' modules or not) ?

**Thank you for reading to the end.

Thank you for your development work.**

johnny-bit:
Quite honestly that elaborate description is very nice indeed...

aurelienpierre:
> the 3 groups evoked by AurelienP
technical modules: all technical issues related to lenses (sensor, lenses and color spaces) such as profiling, range and tone mapping, reconstruction of damaged parts, perspective and lens correction, etc.
grading: everything concerning primary (corrective) and secondary (creative) subjective corrections of colours and tones,
(Special) effects: retouching, liquefaction, flowering, sharpness and others.

> These notions, although fundamental and right, only have real meaning for initiates.

And the goal of the graphical user interface is to provide that meaning in a more intuitive way. A GUI can also be used to educate users, guide, present, and explain.

I don't agree with your splitting, it simply makes no sense from a retouching point of view.

> I don't see myself selling their darktable 'sell' on the basis of technical modules and primary or secondary corrections!!!, knowing that even in your discussions the borderline doesn't clearly exist.

The borderline does exist. Read any cinematographer book. Primary grading neutralizes color, secondary brings the artistic intent. People need to learn.

If you lack experience, you need to trust people who do that everyday.

johnny-bit:
> The borderline does exist. Read any cinematographer book. Primary grading neutralizes color, secondary brings the artistic intent. People need to learn.

Hey, that's a great idea! Is there any way we could link to free resources of increasing technicalities on how to do processing the cinematography way? On yt's lightroom presenters there's just too many slider-pushers and hardly any interesting stuff, while learning about primary, secondary and effect stuff helps tremendously in speeding up process for better final result.

Or even better - can we include that in manual as to guide processing? I mean, once a person learns primary&secondary grading it should be far easier for them to achieve things like constant look for set of photos that need constant look. Or establish their own "unique" look while allowing them to shoot variety of scenes in variety of lighting conditions and variety of lenses combo.

sergei18500:
Hello,
I take the liberty to bring another vision, that of a simple lambda user. 
Sorry to have polluted this forum with my ignorance and incompetence.
Message received 5 / 5.

aurelienpierre:
> Hey, that's a great idea! Is there any way we could link to free resources of increasing technicalities on how to do processing the cinematography way?

@johnny-bit I have put some resources in there : https://github.com/darktable-org/darktable/wiki/Developer's-guide

> Hello,
I take the liberty to bring another vision, that of a simple lambda user.
Sorry to have polluted this forum with my ignorance and incompetence.
Message received 5 / 5.

There are 2 ways to do GUI : 
1. indulge users habits, even when you know they will backfire, and look at them self-harming and sending emails for support,
2. take GUI as an opportunity to teach the best practices while giving access to the software features.

In medicine or civil engineering, bad practices kill people and best practices avoid as many deaths as possible. But image processing kills nobody so everyone is entitled to think he is right. That's a populist lie. We have color theory and all the experience from the guys at Kodak, VFX houses, printing labs and professional color-graders, to show us the path of least resistance for image editing. 

That path relies on doing the right operation at the right place in the pipe. But that triggers the need to set-up the right operation at the right time in the workflow, or you end up with circular-editing. So people need to understand that pipeline and workflow are 2 different things, and while the workflow is dictated by the pipeline, they are not 1:1 equivalent because the pipeline doesn't care for the workflow. 

So I'm very sorry if this doesn't sound like participative democracy, but here is the cold truth:
1. physics comes first in the pipe because it needs light emissions encoded as untampered light emissions to properly unroll optical models to fix lenses distortions, blur, noise, etc.
2. primary grading (white balance, color casts, gamut mapping from garbage sensor RGB to visible locus) comes second because any edit should begin with a neutral and plausible image in order to be able to use presets or batch process for the next part…
3. secondary grading (creative color shifts) comes third because, once the image**s** have been all properly neutralized to the same state, it's a piece of cake to apply presets or copy/paste history and the results should be both consistent and reliable, from picture to picture,
4. display encoding (tone/gamut mapping, color conversion, black point compensation) comes last because exports can be made to many different outputs (paper prints, SDR and HDR displays) from a single master, so the master look should be fully independent from any assumption on the output medium.

Let us please use the GUI to make that flowchart feel natural to the user so he is naturally compelled to follow it (while still being 100% able to practice self-harm if he absolutely wants to).

AlicVB:
ok guys,

I've finally found some time to work on this, so this PR is now ready to test/review.

**the big points :**
- the groups and their contents are driven by presets
- there's 2 build-in presets : "default" with @aurelienpierre  proposal from #5322 and "3.0 layout" with current groups/modules
- at first startup with this, a "previous layout" preset is created with the exact groups content set by the user
- there's a (nicer) gui to add/remove presets/groups/modules
- based on the name of groups, there's a list of recommended modules to add (not already in place for "default" preset, see further)
- "more modules" lib is removed as not needed anymore

**I've postponed "on purpose" some refinements :**
- the list of "recommended" modules for each group in "default" mode (I've already "stolen" @aurelienpierre  work from #5322, so I prefer to stop here and let that to him)
- the nice groups icons from #5322 (same reason)
- this PR will be able to handle my "old" PR #3648 about the "basic group". everything is in place...

Thanks for your tests !

johnny-bit:
WOW, this looks AWESOME! It just needs some polish(ing)

- The menu in darkroom for accessing presets of favourite modules is now pointless? But I do believe that some people used that for their workflow and favourite modules have special place in users' heart. With changes in preset menu in modules you went with removing "favourite" as a special meaning. I wonder how to somehow get the preset menu of "favourite" modules back int that quick presets menu (or if scrapping that completelly wouldn't caouse users outcry)
- the "X" in dialog simply discards changes even if i did some rather extensive edit. asking whether to save changes or dismiss would be good imo
- with close to "save" there could be "reset"  or something button that would dismiss my current messing around and reset them to last known.
- I don't know if it's a bug or intended but for me hovering over tabs didn't bring up tooltip with the group name.

All in all it's IMO best of all worlds: bringing in "sane" defaults suggested by Aurelien, leaving possibility to have "normal looking tabs" for those accustomed to previous layout and super powerful combo for setting up own personal "flow" with tabs!

Thumbs up for this one!

TurboGit:
Yep really nice! For the current preset (maybe that's part of point one in your postpone section) there is some missing modules like the "RGB curve".

TurboGit:
>     * I wonder how to somehow get the preset menu of "favourite" modules back int that quick presets menu (or if scrapping that completelly wouldn't caouse users outcry)

Since AlicVB was able to create a new layout out of the default preset I'm sure we can recover all user's presets and create new layout out of them.

johnny-bit:
> Since AlicVB was able to create a new layout out of the default preset I'm sure we can recover all user's presets and create new layout out of them.

I meant more in future-proof way. Currently if a module in your favourites has presets, you can easily access them in bottom-left corner as a "quick preset". I even used that couple of times and it's a good idea to have. However in this PR "favourite modules" looses all meaning, so there should be a way for the user to add presets of favourite modules to quick presets menu.

elstoc:
Most of the time I stick to the "favourites" and "active" groups, using the search box for less commonly used modules (though the ability to  browse by group is still useful). One thing that would be really nice would be to have the ability to combine favourite and active modules in a single group so that I don't have to flip between them if I add a less-frequently-used module to my image.

aurelienpierre:
> * the list of "recommended" modules for each group in "default" mode (I've already "stolen" @aurelienpierre  work from #5322, so I prefer to stop here and let that to him)
> 
> * the nice groups icons from #5322 (same reason)

Don't worry about that, just cherry-pick the commits you need or copy-paste useful code, I don't really care who gets credit at the end. I just care that the work gets done.

AlicVB:
update : 
- @johnny-bit : add group name tooltips
- @johnny-bit : I've "inversed" the save logic : save is automatic when leaving editor window or switching to another preset. I've added a reset button in place of the save one to retrieve the last saved state

still to be done : 
- @johnny-bit : I've completely forgot the "quick preset". Effectively, we need to change how it works...
- @aurelienpierre : thanks for the share. I'll try to cherry-pick without screwing my branch... I always tend to do weird things with git ;)

and questions for @TurboGit : 
- You say that we miss some modules in "3.0 layout" preset but I can't reproduce. I've done the list by starting dt master with empty .config dir, add an image without xmp, list all visible modules in different tabs...
- And when I've done the tests, I've seen that filmic rgb is not visible by default (normal, default is display refer) but when switching to scene refer, we don't make it visible. I wonder if we can't do something here...

TurboGit:
>     * You say that we miss some modules in "3.0 layout" preset but I can't reproduce. I've done the list by starting dt master with empty .config dir, add an image without xmp, list all visible modules in different tabs...

I retested, maybe on my side but if I select the "3/0 layout" I don't have "RGB curve" listed. I do have "RGB levels" and "FilmicRGB".

Note that my week-end is very busy, so do not expect timely response on my side.

AlicVB:
@TurboGit : in current master, it seems that RGB curve is not visible by default, but RGB levels is... that's why I don't add it to "3.0 layout" preset.
That said, it would be great to add more "sane" default preset once everything is in place.

Enjoy your week-end !

Nilvus:
I really like that PR with last commits. Thanks for this great work. The preset manager UI is much better (some CSS minor refinements to do but if you want, I can work on that when this will be merged).

**Two things:**
- I totally agree with @johnny-bit and @elstoc about favorites tab. I really think that this tab needs to be kept by default on default presets. For example, I really would love to keep my favorites tab but with @aurelienpierre layout. Actually, I need to make a duplicate preset to add favorites tab and add manually all my favorites modules. Just annoying to do that (and I also think at all users who will/could post issues about that. So, just an idea, why not keep favorites tab on default presets if it's detected that this tab have some modules set? That's also kept some legacy that deals with this great improvement.
- On preset manager, it's sometimes long time to find a module, and I don't understand the order of them. That needs to be improved and ordered alphabetically to find modules faster. And why not also add a search field in the top of the module list?

Nilvus:
About favorites tab, another option (an idea I just had): adding an option on preset manager to add existing favorites tab modules on a preset with just one click. So for example, for what I want, that would need just 2 clicks:
- duplicate default layout
- click on a button to automatically add favorites tab and modules set on it.

AlicVB:
Simple things first : 
- yes, it certainly need some css refinement, maybe once this is merged ?
- in the "add module" list, modules are in pipe order, but maybe it would be better to put them in alphabetical order here, as pipe order doesn't mean anything in this context. For the search box, maybe... but that would add quite complexity in the code. Maybe postpone that once everything else will be in place

Now for the favorite tab : 
in fact "favorite" tab doesn't means anything now. It's just a customizable group like the others.
Now to solve your problem, what I can propose is to add 2 presets when upgrading from a previous version : 
- "previous layout" : to reflect exactly the group that where set before (including "favorite" group)
- "previous layout updated" (better name ?) : to keep all modules that were set before but with the "new" 3 groups layouts + favorite group. What do you think ?

Nilvus:
> Simple things first :
> 
>     * yes, it certainly need some css refinement, maybe once this is merged ?

Exactly my suggestion. And of course, as usual, I will propose my PR for that after this PR will be merged.
 
>     * in the "add module" list, modules are in pipe order, but maybe it would be better to put them in alphabetical order here, as pipe order doesn't mean anything in this context. For the search box, maybe... but that would add quite complexity in the code. Maybe postpone that once everything else will be in place

Search box was just a suggestion. Having the list set alphabetically is more important and enough.

> Now for the favorite tab :
> in fact "favorite" tab doesn't means anything now. It's just a customizable group like the others.
> Now to solve your problem, what I can propose is to add 2 presets when upgrading from a previous version :
> 
>     * "previous layout" : to reflect exactly the group that where set before (including "favorite" group)
> 
>     * "previous layout updated" (better name ?) : to keep all modules that were set before but with the "new" 3 groups layouts + favorite group. What do you think ?

I'm ok with your proposal.

AlicVB:
So, first, thanks to all for your test and great ideas.
I *think* I've done with all the proposal here : 
- merge @aurelienpierre works from #5322 : groups and new icons (I didn't manage to cherry-pick partially, so I just copy-paste :(
- change how "quick presets" by adding a way to choose exactly which presets to show in the list (old favorite presets are retrieve on upgrade)
- on upgrade, create a "previous layout updated" preset, based on old settings (favorite and module visibility) but with @aurelienpierre 3 groups layout
- fix small bugs

@elstoc : I've not take in account your proposal of favorite + active merge, this will need way more thinking, and I'm not sure it's a good idea, considering that active group as very special meaning...

elstoc:
> I've not take in account your proposal of favorite + active merge, this will need way more thinking, and I'm not sure it's a good idea, considering that active group as very special meaning

That's fine. This PR is quite big already and we want to avoid scope-creep. If you wanted to retain the specific meaning of the active set, we could maybe add the option to _also_ include active modules in the favourites (or other user-defined) group instead. Anyway I agree this is one to consider later on if at all.

AlicVB:
so... can someone help me with the build error raised by travis ? The log are truncated and the only error I've found is filmicrgb test that failed... but I've not changed anything in filmicrgb apart its group (which should influence the result of course)
Thanks !

TurboGit:
@AlicVB : don't know what the problem is... Maybe a transient bug as indeed there is an issue with the filmic unit testing and this cannot come from this PR I suppose! Let's see if new push fixes this.

src/iop/filmicrgb.c
```diff
@@ -309,7 +309,7 @@ const char *name()
 
 int default_group()
 {
-  return IOP_GROUP_TONE;
+  return IOP_GROUP_TONE | IOP_GROUP_TECHNICAL;
```
TurboGit, line 312: Could this be the actual problem? This is default_group() and maybe it expect a single group there. Maybe worth double checking?

src/libs/modulegroups.c
```diff
@@ -673,15 +719,1164 @@ static void _lib_modulegroups_switch_group(dt_lib_module_t *self, dt_iop_module_
...
+static gchar *_preset_retrieve_old_layout()
+{
+  gchar *ret = NULL;
+
+  // layout with "old" 5 groups
+  for(int i = 0; i < 6; i++)
+  {
+    // group name and icon
+    if(i == 0)
+      ret = dt_util_dstrcat(ret, "ꬹ1ꬹfavorites|favorites|");
```
TurboGit, line 780: The encoding with the strange character `ꬹ` needs some clarification :)

src/iop/filmicrgb.c
AlicVB: yes, you're right, this comes from here I think : https://github.com/darktable-org/darktable/blob/43533544083436996cfa31066b53593660f146b1/src/tests/unittests/iop/test_filmicrgb.c#L68
Effectively, each iop now belong to 2 groups : the "old" one for the 5 tabs layout and the "new" one for the 3 tabs layout. But in the unit test, we test only for "old" group...

I've no idea how the tests works... Do you think it's enough to just replace `IOP_GROUP_TONE` by `IOP_GROUP_TONE | IOP_GROUP_TECHNICAL` in the test ? or do we need to take in account something more ?

TurboGit: I suppose so, yes.

src/libs/modulegroups.c
AlicVB: yes, completely forgot to document that... (same for quick presets) will do it

src/iop/filmicrgb.c
AlicVB: done

src/libs/modulegroups.c
AlicVB: add comment in the code to explain the syntax. Shortly, the "strange" ꬹ char is to separate groups (=tabs) and | char is to separate modules... Note that I don't mind using other char as separators ;)

Nilvus:
Just tested and it's really great. Good work. Just a thing, previous layout updated term is not clear. I would suggest something like this:
=> **new layout with favourites** or **recommended layout with favourites**

And so for previous layout, why not something like:
=> **3.0 layout with favourites** or **legacy layout with favourites**

TurboGit:
We could also have presets for use in display-referred & scene-referred mode. In scene-referred preset (or maybe all) we probably want to add "RGB curve" at this stage.

AlicVB:
@Nilvus : I fully agree that the names can be better :) the only problems with what you propose is that current "previous..." layout not only retrieve old favorites, but also takes in account the modules the user has chosen to show/hide...
maybe something like : "**my previous config.**" and "**my previous config. with 3 groups**"... or something else...
btw, I wonder if "**legacy layout**" would be better than "**3.0 layout**" for built-in presets ?

@TurboGit : sure, we need to carefully choose the built-in presets content. Currently, I've just done copy-paste from how master works and from #5322 ... Note that presets content will be really simple to change once this PR is merged and I would prefer that, as I don't feel comfortable to do that part ;)

Nilvus:
> @Nilvus : I fully agree that the names can be better :) the only problems with what you propose is that current "previous..." layout not only retrieve old favorites, but also takes in account the modules the user has chosen to show/hide...
> maybe something like : "**my previous config.**" and "**my previous config. with 3 groups**"... or something else...
> btw, I wonder if "**legacy layout**" would be better than "**3.0 layout**" for built-in presets ?

So, why not somethign like:

- my previous config
- my previous config with new layout

AlicVB:
@Nilvus : I take it... thanks :)

AlicVB:
@TurboGit : I just rebased to master to resolve little conflict. I think this is ready to review when you have time... The remaining parts (change of default visible modules, more built-in presets) can occurs after this PR is merged.

Once this in place, I will rework on the "basics widgets" PR #3648

TurboGit:
One issue:
- I have duplicated the default layout
- I have added into the grading group "rgb curve"
- I have then added the "color zones"

And I got a single module into grading "color zones", all others are lost:

image

Also all other presets seems now broken, for example grading in default does not have module:

image

Also note that in the "default" and "legacy layout" window the top left group (techninal or base) area is a bit shorter on the right.

Also, I think we need to at least import the users presets in the PR otherwise we are going to have all devs testing master loose their presets.

TurboGit (changes requested):
Strange bug when duplicating a preset. Maybe because when duplicating a preset we do not properly duplicate all list of module?

TurboGit:
Also a little padding on the right of the group icon to separate it more from the group label would be nice.

AlicVB:
So, great thanks @TurboGit for your test !
Everything should be fixed now : 
- the "Strange bug when duplicating a preset" (I was a little ashamed of that bug, I was sure to have test that lot of time... except after last commit...)
- margins & paddings
- retrieval of user defined modulelist presets

data/themes/darktable.css
```diff
@@ -1258,6 +1257,106 @@ filechooser row:hover,
...
+  margin: 0 0 0 1em;
+}
+#modulegroups-groups-box
+{
+  border-top: 1px solid @plugin_label_color;
+}
+#modulegroups-presets-list
+{
+  padding-top: 1em;
+}
```
Nilvus, line 1324: Remove that setting to have menu aligned on top like in preferences. More consistent and better seeing I think.

data/themes/darktable.css
```diff
@@ -1258,6 +1257,106 @@ filechooser row:hover,
...
+/*--------------------------------
+  - module layouts manager dialog -
+  --------------------------------*/
+
+#modulegroups_editor
+{
+  color: @fg_color;
+  background-color: @plugin_bg_color;
+  padding-left: 0.5em;
+}
```
Nilvus, line 1269: I suggest to add an empty line after between CSS parts. It's help reading CSS by aerate it I think.

data/themes/darktable.css
```diff
@@ -1258,6 +1257,106 @@ filechooser row:hover,
...
+  margin: 0.5em;
+}
+#modulegroups-preset-name
+{
+  font-size: 1.1em;
+}
+#modulegroups-ro
+{
+  font-style: italic;
+}
```
Nilvus, line 1337: I would add a margin setting on this read only text:

`margin: 0.4em 0;`

data/themes/darktable.css
```diff
@@ -1258,6 +1257,106 @@ filechooser row:hover,
...
+#modulegroups-groupbox
+{
+  padding: 0 0.5em 0 0.5em;
+}
```
Nilvus, line 1341: And there, add that setting (add space between separator and groups module:

`margin-top: 0.75em;`

TurboGit:
>     * retrieval of user defined modulelist presets

Still no luck with that ! Any idea about what could be wrong?

data/themes/darktable.css
```diff
@@ -1258,6 +1257,106 @@ filechooser row:hover,
...
+#modulegroups-iop-header,
+#modulegroups-header-center,
+#modulegroups-preset,
+#modulegroups-preset-activated
+{
+  background-color: @bg_color;
+  color: @plugin_label_color;
+  padding: 0.5em 0.5em 0.5em 0.5em;
+  margin: 0 0 0.2em 0;
```
Nilvus, line 1278: Not consistent with preferences. It would be better to make them consistent and add them on existing preferences settings (around lines 1160 in preferences) would be better and avoid multiply lines on CSS.

data/themes/darktable.css
```diff
@@ -1258,6 +1257,106 @@ filechooser row:hover,
...
+}
+#modulegroups-reset
+{
+  padding: 0.5em;
+  margin: 0.5em;
+}
+#modulegroups-preset-name
+{
+  font-size: 1.1em;
+}
```
Nilvus, line 1333: I suggest to aerate and add:

`margin-top: 0.4em;
`

Nilvus (changes requested):
Some first CSS improvements suggested there. If you prefer that I propose a PR with those suggested changes, feel free to ask it.

AlicVB:
> Still no luck with that ! Any idea about what could be wrong?

hmm... have you try with a "clean" darktablerc from master ? previous presets are retrieved in the same way than old layouts : only if `plugins/darkroom/modulegroups_preset` key doesn't exist in darktablerc (which means that we came from a "pre-this-pr" version)

data/themes/darktable.css
AlicVB: I've tried to do that in last commit... but not all can be shared as widgets structure is not the same. I hope it's ok for you !

TurboGit:
@AlicVB : But the presets are not in darktablerc, the user's presets are real presets stored in data.db:

`select * from presets where operation="modulelist";`

AlicVB:
@TurboGit : yes, that's where I get them, `plugins/darkroom/modulegroups_preset` check is just to trigger the retrieval (like for importing old layouts, we don't want to do the retrieval multiple times).
The other way would be to check the database, import the presets and remove the old ones...

Nilvus:
Thanks @AlicVB to integrate my proposals. 3 things remaining:

- There's no hover state existing in left menu (presets names) and would be good to have it (like in preferences). In Gtk inspector tool, when hovering preset name, state remains in backdrop and don't show hover state. So it's not CSS but probably some Gtk code missing I think
- + button need some top margin to aerate but no specific name. I would suggest a little margin like 0.5em.
- And last one: default height of the window needs to be increase a little to avoid seeing scrollbar (since my CSS proposals, we have one) on default preset.

AlicVB:
@Nilvus : thanks for your suggestions : all them should be in place now : 
- I have set the window to the same size size as preference
- I have added the "hover" effect
- I have added the top margin for '+' buttons (0.4em to mimic the margins between group names and modules list)
- plus, I have increased the entry width for changing preset name, I have limited the width of the preset sidebar (with ellipsizing of the preset long names)

As usual, don't hesitate to suggest other improvement...

Nilvus:
@AlicVB: thanks, great work. Just 2 things remaining:

- maybe width of preset name entry is now too big. Something between previous width and new one would be better
- text when hovering preset name on the left is white as in preferences is grey. Would be better to have it grey like in preferences for better visibility

AlicVB:
@Nilvus : here you are... at least I am in progress, I pass from 3 remaining things to 2... waiting for the last one :-P

src/develop/develop.c
```diff
@@ -1417,7 +1417,7 @@ static gboolean _dev_auto_apply_presets(dt_develop_t *dev)
            "        AND ?10 BETWEEN focal_length_min AND focal_length_max"
            "        AND (format = 0 OR format&?11!=0)"
            "        AND operation NOT IN"
-           "            ('ioporder', 'modulelist', 'metadata', 'export', 'tagging', 'collect', '%s'))"
```
TurboGit, line 1420: Not sure we want to remove modulelist from here since while testing the modulelist entries where still in my data.db.

src/libs/modulegroups.c
```diff
@@ -23,6 +23,8 @@
 #include "control/conf.h"
 #include "control/control.h"
 #include "develop/develop.h"
+#include "dtgtk/button.h"
+#include "dtgtk/icon.h"
 #include "gui/gtk.h"
 #include "libs/lib.h"
 #include "libs/lib_api.h"
```
TurboGit, line 30: Shouldn't this module have version upgraded to 2 ?

src/libs/modulegroups.c
```diff
@@ -671,17 +717,1295 @@ static void _lib_modulegroups_switch_group(dt_lib_module_t *self, dt_iop_module_
...
+
+  gchar *tx2 = NULL;
+  tx2 = dt_util_dstrcat(tx2, "ꬹ1ꬹ%s|%s||%s", _("base"), "basic",
+                        "basecurve|toneequal|clipping|flip|exposure|demosaic|highlights|temperature|filmicrgb");
+  tx2 = dt_util_dstrcat(tx2, "ꬹ%s|%s||%s", _("tone"), "tone", "rgblevels|bilat");
+  tx2 = dt_util_dstrcat(tx2, "ꬹ%s|%s||%s", _("color"), "color", "colorbalance|colorin");
+  tx2 = dt_util_dstrcat(tx2, "ꬹ%s|%s||%s", _("correct"), "correct", "sharpen|hazeremoval|lens|denoiseprofile");
+  dt_lib_presets_add(_("legacy layout"), self->plugin_name, self->version(), tx2, strlen(tx2), TRUE);
+
+  // if needed, we add a new preset, based on last user config
```
TurboGit, line 984: Ok, this is certainly why it is not working for me. modulegroups_preset is not currently used in master but was added when I tested this PR so now the import does not happens.

What I would propose is:
- always call the migration routine
- check for modulelist as done now with writeprotect=0
- when a user's modulelist is migrated I would remove it from data.db

To make this easier the call dt_conf_key_exists("plugins/darkroom/modulegroups_preset") could be replaced by a call checking if there is some user's modulelist in data.presets.

src/develop/develop.c
TurboGit: Read below, but indeed if we remove the modulelist as proposed below we can remove it from here.

TurboGit (changes requested):
Some strengthening of the migration code needed. TIA.

src/libs/modulegroups.c
TurboGit: Maybe even better, if we bump the module version we can probably do the migration of old presets into the legacy_params() code. This will be done once and follows the dt "way" of doing migration.

src/libs/modulegroups.c
AlicVB: We can, but as presets where no implemented in previous version, I'm not sure it's needed...

src/libs/modulegroups.c
TurboGit: Well, but not sure it is possible to use legacy_params() as the modulist module is removed.

AlicVB: Indeed, as there's no pre-existing presets for modulegroups,  legacy_params is not called... That said, I've implemented your previous proposal and it works well !

AlicVB:
@TurboGit : I've fix your last comments, but...
- I've not bumped the module version as it's not useful imho (but I can do it if you prefer)
- I remove all old modulelist presets (not just users ones) after retrieving them. Cleaner imho as it doesn't left unused entries in the database

TurboGit:
Great, I'll test again...

TurboGit:
Ok, this is now working on my side. I will merge this now, no need to postpone this more. We can rework/update the groups content to add more module. I think it is a great time to do that for 3.4. Maybe we should keep the default and legacy as done here as they used to exists and add some more. My current thinking is to add a preset for basecurve based workflow and one for filmic based one. And from there we can let users added whatever they like.

AlicVB:
thanks a lot !
For the built-in presets I agree with you. Currently I've done the minimal as I don't feel comfortable enough to go further...
@aurelienpierre : if you see this, maybe you can share you thought ?

What would be great imo : 
- filmic specific preset (same for basecurve)
- rework old modulelist "workspace*" and "subsets*" presets

junkyardsparkle:
Wow, interesting change! I tried to approach this without having closely followed this PR, as a typical user would. The new preset menu icon on the groups header didn't immediately jump out at me as the replacement for the old "more modules...", but I did see it eventually. One comment on the configuration window: when using the "+"  next to "module groups", it wasn't immediately obvious that anything had happened, due to the new group being added out-of-sight to the right of the existing ones. Would it be possible for this action to automatically scroll the newly added group into view?

I also noticed that this window is not a child of the main darktable window in the same way as the general configuration one; when using Alt+Tab to switch back to this browser window, it remained on top, obscuring the page (using XFCE). Other than that, my naive usage attempt went well. :-)

parafin:
> I also noticed that this window is not a child of the main darktable window in the same way as the general configuration one; when using Alt+Tab to switch back to this browser window, it remained on top, obscuring the page (using XFCE). Other than that, my naive usage attempt went well. :-)

Yeah, both "manage modules layouts" and "manage quick presets" windows are non-modal top-level windows, which doesn't seem right to me. You can interact with the main window while these windows are open, e.g. switch to lighttable. Probably not desired.

AlicVB:
@parafin @junkyardsparkle : should be fixed in PR #6461 
@junkyardsparkle : about the scroll when adding new group, I'll have a look...

thanks both for your tests and comments

TurboGit:
@AlicVB : I have just created #6497 for documenting this work. Do you think you'll have time to work on that for 3.4. Let me know if I can help.

PR #5685, opened 2020-07-01
hinter : ellipsize the text

### LINKED ISSUE #5649: Unexpected/annoying window resize when interactive tool description is too long
Hi.

**Describe the bug**

When the "on hover" description of an interactive tool is too long (for example the Italian text of shapes for the spot removal tool) it may not fit in the top panel; when the mouse goes over the tool and the description text is shown, the Darktable window end up changing its width to make room for the whole text, making interaction very annoying as the content that is being edited will shift too and move from under the mouse cursor.

I originally experienced this with a 1280x1024 screen and the Italian translation on Darktable 2.4.3 on Debian, and I reported it in https://redmine.darktable.org/issues/12199

I also experienced the same issue on a 1920x1080 screen with Darktable 3.0.2 on Debian.

**To Reproduce**

One way to reproduce the behavior is to use the Italian locale and:

1. Open Darktable and maximize the window
2. Use the spot removal tool and add one circle
3. Go with the mouse over the circle, the description text will show on top and cause the window to resize, moving the circle from below the mouse pointer possibly loosing the focus on the circle.
4. Because of the maximization status the window will shrink again when the mouse moves and this resize dance is very annoying.

Another way to reproduce is with the English locale and a very narrow width:

1. Open Darktable
2. Use the spot removal tool and add one circle
3. Collapse the right panel
4. Resize the window horizontally to make it narrower (from about 700px width the problem can be reproduced)
5. Go with the mouse over the circle, the description text will show on top and cause the window to resize, moving the circle from below the mouse pointer possibly loosing the focus on the circle.

**Expected behavior**

I would expect the window to never change its size automatically.

A temporary workaround is to use the default text launching Darktable with `LANG=C darktable`, or to collapse the top panel to completely prevent descriptions to show up.

**Screenshots**

The old video still shows what happens:
https://redmine.darktable.org/attachments/download/3947/Darktable-long-tooltips-change-window-width.mp4

 - Darktable Version: 3.0.2
 - OS: Debian Linux
 - Desktop environment: both gnome-flashback and normal gnome-shell

**Additional context**

Also reported in https://redmine.darktable.org/issues/12199 for Darktable 2.4.3

It was suggested to fix this at the Italian translation level, but this may happen for other translation too, and besides that having a windows that changes size depending on its content is rather unexpected anyway.

Thanks,
Antonio

AlicVB:
this fix #5649

### DIFF
--- a/src/libs/tools/hinter.c
+++ b/src/libs/tools/hinter.c
@@ -72,6 +72,7 @@ void gui_init(dt_lib_module_t *self)
 
   self->widget = gtk_event_box_new();
   d->label = gtk_label_new("");
+  gtk_label_set_ellipsize(GTK_LABEL(d->label), PANGO_ELLIPSIZE_END);
   gtk_container_add(GTK_CONTAINER(self->widget), d->label);
 
   darktable.control->proxy.hinter.module = self;


TurboGit:
Thanks !

PR #5692, opened 2020-07-02
po/fix paste parts

TurboGit:


TurboGit:
@AlicVB : I did not found a better fix, was quite tricky to investigate. At least if there is no better fix this could go in 3.2 as the corresponding issue is a show stopper for the release.

TurboGit:
> what about the closing button of the selection widget?

It returns the cancel values as far as I know. Note that this check was there before and has always been working.

jenshannoschwalm:
When I tried to find the problem for this issue  it also happened when pressing that.

TurboGit:
You mean when quitting using the top-right cross in the dialog's title bar ?

TurboGit:
Ok now we check for != OK which ensure that the paste is done only when OK is received. Should be safer anyway.

TurboGit:
Thanks for the review @AlicVB .

PR #5695, opened 2020-07-02
Add tooltip info when tone equalizer is selected (like other modules)

### LINKED ISSUE #5669: FR: Have a way to disable modules from running on OpenCL
Considering issues like #3756, I feel that a global "turn off all OpenCL" is not too fine grained enough. I'm wondering if it is possible to blacklist certain problematic modules from running on OpenCL. 

A button in the preset menu that has a check mark that lets you blacklist a module from running on OpenCL, falling back to the CPU implementation. This could also be used to communicate that there is no OpenCL implementation available.

It could also be done on a global module basis, not a module instance basis, by putting this info in the settings panel.

### LINKED ISSUE #5686: tone equalizer → add keyboard/mouse hints
Add keyboard/mouse hints (just like other modules do) on top of the image for tone equalizer module.

hatsunearu:
Fixes #5686. 

image

### DIFF
--- a/src/iop/toneequal.c
+++ b/src/iop/toneequal.c
@@ -1825,6 +1825,11 @@ static void switch_cursors(struct dt_iop_module_t *self)
     // if pipe is clean and idle and cursor is on preview,
     // hide GTK cursor because we display our custom one
     dt_control_change_cursor(GDK_BLANK_CURSOR);
+    dt_control_hinter_message(darktable.control,
+                  "scroll over image to change tone exposure by 0.25 EV\n"
+                  "shift+scroll to change by 1 EV, and ctrl+scroll to change by 0.1 EV\n"
+                  "A+scroll to zoom");
+
     dt_control_queue_redraw_center();
   }
   else if(!g->cursor_valid)
@@ -1844,6 +1849,7 @@ static void switch_cursors(struct dt_iop_module_t *self)
     GdkCursor *const cursor = gdk_cursor_new_from_name(gdk_display_get_default(), "default");
     gdk_window_set_cursor(gtk_widget_get_window(widget), cursor);
     g_object_unref(cursor);
+
   }
 }
 
--- a/src/iop/toneequal.c
+++ b/src/iop/toneequal.c
@@ -1825,10 +1825,9 @@ static void switch_cursors(struct dt_iop_module_t *self)
     // if pipe is clean and idle and cursor is on preview,
     // hide GTK cursor because we display our custom one
     dt_control_change_cursor(GDK_BLANK_CURSOR);
-    dt_control_hinter_message(darktable.control,
-                              "scroll over image to change tone exposure\n"
-                              "shift+scroll to change in large steps\n"
-                              "ctrl+scroll to change in small steps");
+    dt_control_hinter_message(darktable.control, "scroll over image to change tone exposure\n"
+                                                 "shift+scroll to change in large steps\n"
+                                                 "ctrl+scroll to change in small steps");
 
     dt_control_queue_redraw_center();
   }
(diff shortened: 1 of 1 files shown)


src/iop/toneequal.c
```diff
@@ -1825,6 +1825,11 @@ static void switch_cursors(struct dt_iop_module_t *self)
     // if pipe is clean and idle and cursor is on preview,
     // hide GTK cursor because we display our custom one
     dt_control_change_cursor(GDK_BLANK_CURSOR);
+    dt_control_hinter_message(darktable.control,
+                  "scroll over image to change tone exposure by 0.25 EV\n"
+                  "shift+scroll to change by 1 EV, and ctrl+scroll to change by 0.1 EV\n"
+                  "A+scroll to zoom");
```
elstoc, line 1831: You can't hard-code "A+scroll" here. "A" is just the default key and can be changed in preferences.

src/iop/toneequal.c
```diff
@@ -2278,6 +2284,14 @@ void gui_focus(struct dt_iop_module_t *self, gboolean in)
...
     dt_bauhaus_widget_set_quad_active(GTK_WIDGET(g->show_luminance_mask), FALSE);
     dt_dev_reprocess_center(self->dev);
+    dt_control_hinter_message(darktable.control, "");
+  }
+  else
+  {
+    dt_control_hinter_message(darktable.control,
+                              "scroll over image to change tone exposure by 0.25 EV\n"
+                              "shift+scroll to change by 1 EV, and ctrl+scroll to change by 0.1 EV\n"
+                              "A+scroll to zoom");
```
elstoc, line 2294: similarly here

hatsunearu: Yikes. What should the text be? Should I try to read the preferences?

elstoc: TBH I'd just remove this line. The same shortcut is also used to pan/zoom while editing masks and there probably isn't enough space to add it to the mask hinter messages as well. I think for some languages even the current mask hinter message doesn't completely fit in the available space.

elstoc: If you really wanted to look it up I think it's stored in ``darktable.control->accels.darkroom_skip_mouse_events``

hatsunearu: Alright, I just removed the last line.

src/iop/toneequal.c
```diff
@@ -1825,6 +1825,10 @@ static void switch_cursors(struct dt_iop_module_t *self)
     // if pipe is clean and idle and cursor is on preview,
     // hide GTK cursor because we display our custom one
     dt_control_change_cursor(GDK_BLANK_CURSOR);
+    dt_control_hinter_message(darktable.control,
+                              "scroll over image to change tone exposure by 0.25 EV\n"
```
TurboGit, line 1829: The numbers should not be there as this could change anytime.

TurboGit: I would rewrite this to something like that:

```
scroll over image to change tone exposure
use shift+scroll  for larger steps
use ctrl+scroll for smaller steps.
```

src/iop/toneequal.c
```diff
@@ -2278,6 +2282,13 @@ void gui_focus(struct dt_iop_module_t *self, gboolean in)
     g->mask_display = FALSE;
     dt_bauhaus_widget_set_quad_active(GTK_WIDGET(g->show_luminance_mask), FALSE);
     dt_dev_reprocess_center(self->dev);
+    dt_control_hinter_message(darktable.control, "");
+  }
+  else
+  {
+    dt_control_hinter_message(darktable.control,
+                              "scroll over image to change tone exposure by 0.25 EV\n"
```
TurboGit, line 2290: likewise

AlicVB:
When you are at it, I think it would be great to also update the accel window (H shortcut)
For that, you "just" have to implement `mouse_actions` function. You can have an example here : https://github.com/darktable-org/darktable/blob/dca4c828b051c79bdb49869b8edf28993476833f/src/iop/clipping.c#L3500

This should be quite trivial, as you already have listed all the actions with their text... just ask if you need help !

hatsunearu:
Hmm, I'm guessing I don't have to implement the A+scroll in that thing (since AFAIK that's a global keybind?)

hatsunearu:
image

src/iop/toneequal.c
hatsunearu: this is done.

TurboGit:
Thanks !

PR #5709, opened 2020-07-06
Simplify tiff compress GUI, default level to 6

### LINKED ISSUE #6040: Risk of misconfigured float TIFF compression
**Describe the bug**
There is a high probability the users are not getting the best compression possible when exporting 32-bit float TIFFs because of the default configuration ("deflate with predictor"). 

I suspect most users will just choose "32-bit float" when desired, but without checking and changing to the "deflate with predictor (float)", and the app also doesn't do any validation for them.

**To Reproduce**
1. Go to 'export selected' in lighttable
2. Click on 'file format'
3. Scroll down to 'TIFF'
4. Select '32-bit float' under 'bit depth'
5. Leave the compression combobox at default 'deflate with predictor'
6. Click on 'export'

**Expected behavior**
The float files are exported using integer predictor (in the default configuration) resulting in far from optimal compression: 235MB from a test 24MP NEF. (Exporting w/ no predictor results in 240MB. Uncompressed is 276MB.) 

The same image exported using the correct float predictor is 199MB.

The end user should not be able to end up with a nonsensical/suboptimal export configuration.

 - Darktable Version: 3.2.1 (and earlier)
 - OS: Windows 10
 
**Additional context**
One possible suggestion to address this is by e.g. https://github.com/darktable-org/darktable/pull/5709

kmilos:
IMHO there's really no need to complicate the choice and confuse the end user, the correct predictor shall be chosen depending on sample bit depth anyway. Ideally, the 'predictor' should be a separate checkbox, since in theory it is independent of the actual compression used (works for deflate, LZW, zstd,... though the latter two might not be available in libtiff always).

Also changed the default level to 6, which seems to be the zlib (& libtiff) default.

One way to fix #6040

### DIFF
--- a/src/imageio/format/tiff.c
+++ b/src/imageio/format/tiff.c
@@ -129,7 +129,7 @@ int write_image(dt_imageio_module_data_t *d_tmp, const char *filename, const voi
     TIFFSetField(tif, TIFFTAG_PREDICTOR, (uint16_t)PREDICTOR_NONE);
     TIFFSetField(tif, TIFFTAG_ZIPQUALITY, (uint16_t)d->compresslevel);
   }
-  else if(d->compress >= 2)
+  else if(d->compress == 2)
   {
     TIFFSetField(tif, TIFFTAG_COMPRESSION, (uint16_t)COMPRESSION_ADOBE_DEFLATE);
     if(d->bpp == 32)
@@ -390,7 +390,7 @@ int write_image(dt_imageio_module_data_t *d_tmp, const char *filename, const voi
           TIFFSetField(tif, TIFFTAG_PREDICTOR, (uint16_t)PREDICTOR_NONE);
           TIFFSetField(tif, TIFFTAG_ZIPQUALITY, (uint16_t)d->compresslevel);
         }
-        else if(d->compress >= 2)
+        else if(d->compress == 2)
         {
           TIFFSetField(tif, TIFFTAG_COMPRESSION, (uint16_t)COMPRESSION_ADOBE_DEFLATE);
           if(d->bpp == 32)
@@ -586,7 +586,7 @@ void *legacy_params(dt_imageio_module_format_t *self, const void *const old_para
     g_strlcpy(n->global.style, o->style, sizeof(o->style));
     n->global.style_append = FALSE;
     n->bpp = o->bpp;
-    n->compress = o->compress;
+    n->compress = o->compress == 3 ? 2 : o->compress;  // drop redundant float case
     n->compresslevel = 6;
     n->handle = o->handle;
     *new_size = self->params_size(self);
@@ -615,7 +615,7 @@ void *legacy_params(dt_imageio_module_format_t *self, const void *const old_para
     g_strlcpy(n->global.style, o->style, sizeof(o->style));
     n->global.style_append = o->style_append;
     n->bpp = o->bpp;
-    n->compress = o->compress;
+    n->compress = o->compress == 3 ? 2 : o->compress;  // drop redundant float case
     n->compresslevel = 6;
     n->handle = o->handle;
     *new_size = self->params_size(self);
@@ -634,7 +634,12 @@ void *get_params(dt_imageio_module_format_t *self)
     d->bpp = 32;
   else
     d->bpp = 8;
+
+  // Drop redundant float case from existing config
+  // TODO: Move to legacy eventually
   d->compress = dt_conf_get_int("plugins/imageio/format/tiff/compress");
+  if(d->compress == 3)
+    d->compress = 2;
 
   // TIFF compression level might actually be zero, handle this
   if(!dt_conf_key_exists("plugins/imageio/format/tiff/compresslevel"))
@@ -754,7 +759,11 @@ void gui_init(dt_imageio_module_format_t *self)
 
   const int bpp = dt_conf_get_int("plugins/imageio/format/tiff/bpp");
 
-  const int compress = dt_conf_get_int("plugins/imageio/format/tiff/compress");
+  // Drop redundant float case from existing config
+  // TODO: Move to legacy eventually
+  int compress = dt_conf_get_int("plugins/imageio/format/tiff/compress");
+  if(compress == 3)
+    compress = 2;
 
   // TIFF compression level might actually be zero!
   int compresslevel = 6;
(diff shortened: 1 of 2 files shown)


kmilos:
If ever refactored, the suggested TIFF export GUI should be something like:

- float checkbox Y/N
- bit depth dropdown: 8 or 16 for integer / 16 (new) or 32 for float
- compress checkbox Y/N (+tooltip saying DEFLATE is used)
- predict checkbox Y/N
- compress level slider

TurboGit:
The 3 convert to 2 and 2 convert to 3 seems wrong to me. I don't see the point.

kmilos:
Et, let me check - there should be no more enum 3 whatsoever, that was the whole point.

kmilos:
> The 3 convert to 2 and 2 convert to 3

Sorry, but I can't see where I make any conversion from 2 to 3, can I kindly ask you point it out in the code review?

src/imageio/format/tiff.c
```diff
@@ -646,15 +635,20 @@ void *get_params(dt_imageio_module_format_t *self)
     d->bpp = 32;
   else
     d->bpp = 8;
+
+  // Drop redundant float case from existing config
+  // TODO: Move to legacy eventually
   d->compress = dt_conf_get_int("plugins/imageio/format/tiff/compress");
+  if(d->compress == 3)
```
TurboGit, line 642: if 3 then 2 here.

src/imageio/format/tiff.c
```diff
@@ -766,10 +760,14 @@ void gui_init(dt_imageio_module_format_t *self)
 
   const int bpp = dt_conf_get_int("plugins/imageio/format/tiff/bpp");
 
-  const int compress = dt_conf_get_int("plugins/imageio/format/tiff/compress");
+  // Drop redundant float case from existing config
+  // TODO: Move to legacy eventually
+  int compress = dt_conf_get_int("plugins/imageio/format/tiff/compress");
+  if(compress == 3)
```
TurboGit, line 766: if 3 then 2 here.

src/imageio/format/tiff.c
```diff
@@ -130,12 +130,6 @@ int write_image(dt_imageio_module_data_t *d_tmp, const char *filename, const voi
     TIFFSetField(tif, TIFFTAG_ZIPQUALITY, (uint16_t)d->compresslevel);
   }
   else if(d->compress == 2)
-  {
-    TIFFSetField(tif, TIFFTAG_COMPRESSION, (uint16_t)COMPRESSION_ADOBE_DEFLATE);
-    TIFFSetField(tif, TIFFTAG_PREDICTOR, (uint16_t)PREDICTOR_HORIZONTAL);
-    TIFFSetField(tif, TIFFTAG_ZIPQUALITY, (uint16_t)d->compresslevel);
-  }
-  else if(d->compress == 3)
```
TurboGit, line 138: case 3 removed.

src/imageio/format/tiff.c
```diff
@@ -796,12 +794,11 @@ void gui_init(dt_imageio_module_format_t *self)
   dt_bauhaus_combobox_add(gui->compress, _("uncompressed"));
   dt_bauhaus_combobox_add(gui->compress, _("deflate"));
   dt_bauhaus_combobox_add(gui->compress, _("deflate with predictor"));
-  dt_bauhaus_combobox_add(gui->compress, _("deflate with predictor (float)"));
```
TurboGit, line 799: this case has been removed, but I don't see any explanation.

TurboGit (commented):
So to me there is some part not clear at least to me and no explanations to motivate the changes.

kmilos:
As for the motivation: as mentioned in the opening comment, having both options 2 "deflate with predictor" and 3 "deflate with predictor (float)" is redundant and confusing to the end user, who shouldn't be burdened with these details unnecessarily. Moreover, there is also a bug in the current implementation as it can lead to an illegal TIFF writer configuration (32-bit float + integer predictor of option 2).

This patch removes redundant option 3 and rolls the all the functionality into option 2 w/ automatic choice of predictor.

The only question I have is whether to also write the change to user's config XML once detected?

kmilos:
One way to fix #6040

src/imageio/format/tiff.c
kmilos: This case is redundant and confusing to the end user, there's really no need for it, and the functionallity has been merged into case 2; moreover, the current implementation can lead to an illegal TIFF writer configuration described in #6040

TurboGit:
I'm revisiting this as there is a slow down in the PR at the moment... I'm not yet sure to understand all the changes here, I know almost nothing on TIFF compression. Can you give me a quick summary of:
- the internal changes for std TIFF
- the internal changes for float TIFF
- the change in the compression (what is supported, what is not)
- the GUI change done to express all this.

At this point I don't care about the actual coding part, only the goals and motivations. All this in the hope it will help me decide about integration or not. TIA.

kmilos:
Sure, let me try to recap.

Currently, there are two drop boxes in the TIFF export GUI with the options of "8-bit"/"16-bit"/"32-bit (float)" and "uncompressed"/"deflate"/"deflate with predictor"/"deflate with predictor (float)". As there is no cross-validation between them, all the combos in the matrix below are possible to set:

| | 8-bit | 16-bit | 32-bit (float) |
| --- | --- | --- | --- |
| uncompressed | x | x | x |
| deflate | x | x | x |
| deflate with predictor* | x | x | o |
| deflate with predictor (float) | ^ | ^ | x |

This, however, can result in a corrupt TIFF file being written in the case marked with "o", as described in bug https://github.com/darktable-org/darktable/issues/6040. (The "^" cases are currently and luckily correctly handled when "deflate with predictor (float)" is selected but the actual TIFF predictor tag is converted and set to the integer case). 

So, if you, as you say, are not familiar with the complexity of TIFF compression configuration possibilities, how do you expect your users to be and expect them to always make the correct choice when they are presented with possibilities for invalid configurations "o" and "^"? I eventually had to look into the source code because I was confused with the plethora of options...

This PR is about reducing the complexity of choice from the UX and passing the configuration burden to the implementation, thus reducing the actions on the user to just this minimal natural decision process: "Do you want compression or not? If yes, do you want the additional prediction step to potentially improve the ratio? I'll do the correct export (and select the appropriate predictor for you if requested) whatever pixel format you've set earlier." The table then becomes an unambiguous: 

| | 8-bit | 16-bit | 32-bit (float) |
| --- | --- | --- | --- |
| uncompressed | x | x | x |
| deflate | x | x | x |
| deflate with predictor** | x | x | x |

with the configuration details hidden in the implementation.

Hope the motivation is a bit clearer now, but please do let me know if I should provide any more info.

\* current meaning: integer predictor
** new meaning: don't care, just use the correct predictor for the pixel format

TurboGit:
@kmilos : lot clearer now and I fully agree with you. Sorry I didn't get the goal in previous comments :(

kmilos:
Great, thanks. Sorry I wasn't clear enough from the start!

Issue #5391, opened 2020-06-10
Disable auto-apply any of base curve presets


esq4:
Hi

It may be right to make it possible switch off auto-apply any of base curve presets. 
For now, this needs to be done manually.

esq4:
Yes, it valid but not obvious.

At least indicate in the tooltip for 'auto-apply per camera basecurve presets' that the setting is valid only when selected 'display-referred' for 'auto-apply pixel workflow defaults'.

PR #5711, opened 2020-07-06
fix #5627 (for "Auto applied WB preset is not used for thumbnail #5391")

### LINKED ISSUE #5627: Possible solution for "Auto applied WB preset is not used for thumbnail #5391"
Yes, it valid but not obvious.

At least mention in the tooltip for 'auto-apply per camera basecurve presets' that the setting is valid only when selected 'display-referred' for 'auto-apply pixel workflow defaults'.

_Originally posted by @esq4 in https://github.com/darktable-org/darktable/issues/5391#issuecomment-642761969_

esq4:


esq4:
Yes of course. Exactly "display-referred".
_I'm sorry, I got lost in my thoughts. :(_

esq4:
Do I need to make new pull request?

esq4:
I'm sorry twice, I built my thoughts and (possibly) updated this request.

TurboGit:
Looks ok now, thanks!

PR #5713, opened 2020-07-06
duplicate : fix offset if ratio is different

### LINKED ISSUE #5701: duplicate module: size/position issues when images have different crops
With a fresh install and after deleting my cache, I'm using the duplicate module (for the first time in this session) with two duplicate images that have different crops. The normal darkroom preview:

Screenshot_2020-07-04_22-03-03

Clicking on the duplicate image:

Screenshot_2020-07-04_22-03-20

I had a similar issue before (#5664) that only seemed to occur when I showed then hid the right hand side panel.

The deciding factor here seems to be that the two images have a different crop. If I start with the image that has the larger crop loaded in the darkroom it works fine but not the other way around. I would expect both images to fill the available width or height.

AlicVB:
this fix #5701

note that when image ratio are different, the duplicate preview is shown at "zoom to fit" zoom level.

### DIFF
--- a/src/libs/duplicate.c
+++ b/src/libs/duplicate.c
@@ -212,8 +212,20 @@ void gui_post_expose(dt_lib_module_t *self, cairo_t *cri, int32_t width, int32_t
   if(!dev->preview_pipe->backbuf || dev->preview_status != DT_DEV_PIXELPIPE_VALID) return;
 
   // use the same resolution as main previem image to avoid blur
-  float img_wd = dev->preview_pipe->backbuf_width;
-  float img_ht = dev->preview_pipe->backbuf_height;
+  float img_wd, img_ht;
+  if(d->allow_zoom)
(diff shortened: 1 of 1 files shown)


AlicVB:
@TurboGit : I think it's safe enough to go in 3.2, as it only change the "corner" case where ratio is different between the current image and the duplicate, without modifying other cases.
That said I prefer letting you decide what is the best...

elstoc:
Works well when fully zoomed out. It still does some slightly odd stuff when you zoom even slightly on images with different crops but it's hard to be certain what the correct behaviour _should be_ in that scenario.

TurboGit:
Ok, looking at the code it is indeed low impact. It fixes an issue, not really a bug, but it will make dt look better when comparing duplicates. So I'll merge this to 3.2. Thanks for the work.

Issue #4298, opened 2020-02-15
histogram in tethering mode not displayed


TurboGit:
**Describe the bug**

No histogram displayed in tethering view.

**To Reproduce**

1. Go to tethering view
2. capture a picture, it is properly displayed in center area
3. the histogram is not updated

**Expected behavior**

Histogram should be displayed.

**Platform (please complete the following information):**
 - Darktable Version: 2.6 & 3.0 & current master
 - OS: Linux Debian
 - 
 - 

**Additional context**

Was broken quite some time ago as it does not work on 2.6.

TurboGit:
@dtorop : Is that something you could have a look at as you've been working on the histogram recently?

dtorop:
@TurboGit I'll take a look at it. It may take a week or so. But sounds like if it's been broken since v2.6, that it can wait that long...

TurboGit:
@dtorop : Sure nothing urgent ! Thanks for looking at it.

johnny-bit:
I think this issue also has wrong scope ;) it's not "CLI" related...

TurboGit:
@dtorop : Any news ? Do you think you could have a look before feature freeze in 10 days ?

TurboGit:
@johnny-bit : Right, changed from CLI to UI of course !

dtorop:
@TurboGit I'll look at it this weekend! I don't know if I actually have a camera here which can tether, but at the worst I can look at the code.

dtorop:
@johnny-bit 

> Whenever I opened tethering view to test #5371 i got:
> 
> ```
> [image_cache_allocate] failed to open image 4294967295 from database: no more rows available
> ```

I just saw this in darkroom view. See #5398 for how to reproduce it.

dtorop:
Presumably the "failed to open" message is from here: https://github.com/darktable-org/darktable/blob/f6842cc0d100eb44079394566ab07558259fd9c6/src/common/image_cache.c#L153-L154

TurboGit:
> Huh... that thing with signal not getting thought or something might be on some track here. Whenever I opened tethering view to test #5371 i got:
> 
> ```
> [image_cache_allocate] failed to open image 4294967295 from database: no more rows available
> ```

This seems to happen when no image are selected on the lighttable when moving to tethering mode. This another issue independent of the histogram.

TurboGit:
[...]

PR #5714, opened 2020-07-07
Tether histogram fix

dtorop:
See #4298.

In the case of the darkroom view, things should generally work as before, except that the processing code occurs in lib/histogram.c rather than pixelpipe_hb.c, and the resulting data is stored in dt_lib_histogram_t rather than dt_develop_t.

For tethering, lib/histogram.c will detect when the center view is updated/changed, and then uses its own pixelpipe to create a histogram for the current image. During live view, views/tethering.c has a bit of extra code to send the live view image to be processed.

This PR ended up being a substantial intervention. It moves a bunch of the final histogram processing/storage and control flow from develop & pixelpipe into lib/histogram.c. This is necessary so that live view can request processing of 8-bit preview data incoming from the camera without going through the pixelpipe. This also helps make develop/pixelpipe_hb.c less verbose. If other scope types, with their own processing code, are added to darktable, it makes sense to have them in lib/histogram.c (or common/histogram.c) rather than being thrown into pixelpipe_hb.c.

I haven't been able to test this code yet in tethering view, due to lacking a camera which can tether. I'd be happy to make fixes as possible depending on the test results.

The commit history is cleaned up, but still contains false starts, in particular one which relied on 8-bit thumbnails from the pixelpipe to create a histogram outside of darkroom view.

Note that the histogram code in master has code for "Constraining the area if the colorpicker is active in area mode". This appears to never run. It could be possible/interesting to fix this. Also, the waveform histogram continues not to mind the histogram color profile. Fixing these things could be interesting, though they're beyond the scope of this PR. There are a few misc. bugfixes and cleanups throughout this PR, though.

It goes without saying that such an intrusive PR (which also changes a frozen string) shouldn't be considered for v3.2.


TurboGit:
@dtorop : Thanks a lot for all the nice work on this. I'll have a look at this when 3.2 is out.

TurboGit:
I will at least test this in tether mode soon to tell you if it works there too.

johnny-bit:
Wow, Looks great!

for tethering, I don't know if it's dt limits or gphoto2 but in my case live view is ~15fps.

dtorop:
@johnny-bit Does that mean you see a histogram? That is fabulous news if so!

> I will at least test this in tether mode soon to tell you if it works there too.

@TurboGit I'd be most curious for results of tests. I can do fixes a bit faster now, when all the code is still in my head, but am happy to look at it post-3.2 as well.

> for tethering, I don't know if it's dt limits or gphoto2 but in my case live view is ~15fps.

That's a hardcoded limit. See 1c0073cd34cec2c6541055473d5607983ccd7a65, from 2012. It's possible that eight years later, many computers can handle a higher frame rate...

dtorop:
> @dtorop : Thanks a lot for all the nice work on this. I'll have a look at this when 3.2 is out.

@TurboGit And you're welcome. I somehow thought it'd be a one line fix, but unless I missed something... It does seem good to give bits of the code a periodic look-through and overhaul, at least.

src/libs/lib.h
```diff
@@ -64,6 +64,16 @@ typedef struct dt_lib_t
...
 
+    /** Histogram processing hooks */
+    struct
+    {
+      struct dt_lib_module_t *module;
+      void (*process)(struct dt_lib_module_t *self, const void *const input,
+                      int width, int height, int stride, gboolean is_8bit);
+      // FIXME: should this be a function or just a boolean which histogram lib keeps updated?
+      // FIXME: should this be a darktable-level value, set by lib/histogram.c and noticed by iops?
+      gboolean is_linear;
```
dtorop, line 75: I've some worry about this -- read/write isn't protected by any sort of mutex. I think the worst case would be that an iop would be redrawing its histogram when a keyboard accelerator changed the histogram scale, and the iop's histogram might redraw with the wrong scale. But I don't have a deep understanding of multi-threading in darktable and all the risks/safeties needed.

pitbuster:
> That's a hardcoded limit. See 1c0073c, from 2012. It's possible that eight years later, many computers can handle a higher frame rate...

Maybe it makes sense to have this limit as a parameter, so people with faster computer can tweak to their liking

TurboGit:
Just tested and it works but with some issue. The waveform is ok, but when I select the standard histogram I have on the console:

`[dt_ioppr_get_work_profile_type] can't find colorin iop`

Also, if there is no selection on the lighttablle and you go to Tethering view, dt freeze (need a Ctrl+C) and on the console I see:

`[image_cache_allocate] failed to open image 4294967295 from database: no more rows available`

Anyway, nice progress !

dtorop:
> `[dt_ioppr_get_work_profile_type] can't find colorin iop`
> 
> Also, if there is no selection on the lighttablle and you go to Tethering view, dt freeze (need a Ctrl+C)

@TurboGit Thank you for testing. I tried to fix both of these in bd9c5e6a0664e4bfefafabfdd2cdc59b10c51f3e.

> Maybe it makes sense to have this limit as a parameter, so people with faster computer can tweak to their liking

@pitbuster I added a conf variable in 6e49f73e6a0be34b6334358b9814f4016c1fa293, plugins/capture/camera/live_view_fps, not exposed via the GUI.

dtorop:
I'll get back to this once v3.2 is out. This is also a gateway PR for a bunch of other histogram work, as it changes so much code structure and it's not worth doing much else until the code gets stable again.

I may have a better line now on borrowing cameras which can tether.

TurboGit:
@dtorop : can you send a quick status update on this one ? That's the kind of change I'd like to merge sooner than later. TIA.

dtorop:
@TurboGit : Thanks for checking in. I borrowed a Canon DSLR to test this. Initial results:

- histogram displays after a tethered image is captured, and the histogram looks right -- but I need to verify this
- histogram displays in live view and is responsive
- histogram appears to show only lightness but not color data in live view
- darktable crashes when histogram is switched to waveform
- histogram doesn't update when switching between images in filmstrip, and switching can cause a crash
- it's possible to cause a crash simply by switching into tether view in certain cases

There's two pieces of good news: That there's backtraces for all the crashes, and that this code isn't in v3.2.

I'll see what I can do to fix things up.

dtorop:
And also, yes, I agree that this would be better to merge sooner than later. Best case, between this PR and #5532 it could be possible to remove a bunch of lines of code from pixelpipe_hb.c and leave the code there easier to work on.

dtorop:
This is coming along. The current version should show a histogram in tether view, including in live view. The intent is that it is also stable. (I've actually been testing it using a tether and live view capable camera.)

I got rid of code which calculated tether view histograms by creating a local pixelpipe. Instead, there's a hack where creating a Cairo surface for tether's center view generates a histogram. This means that tether view histograms are now always calculated from 8-bit data. In most cases this should be good enough for judging exposure and getting a rough sense of the image. Some quantization is visible, especially for waveform histograms, and overexposed data will always clip at 100%. On the other hand, there are fewer signal handlers and things don't regularly crash. (The prior code failed when iops changed behavior when a GUI was attached, and in a few other cases.)

Unfortunately, the regular histogram in tether view is slightly different than the histogram for the same image in darkroom view (though the waveform histogram does match closely). There may be a bug in here -- I can look at this more. It may simply be a side effect of too many conversions of 8-bit data and some conversions using different algorithms (lcms vs. internal darktable), or that quantization changes binning and the histogram scale.

As a side effect of recent changes, the live view is now converted to display colorspace before drawing, which should make it more accurate (assuming that the incoming live view is sRGB).

I'm curious for thoughts on this code and any testing.

TurboGit:
I'll test & review this in the coming days. Thanks for all the work and the feedback.

dtorop:
@TurboGit : Great! I have one more idea -- I'll try it out in a moment and write if there's any more changes.

dtorop:
Indeed -- it looks like I can fix up the profile situation a bit. So if possible hold off testing/reviewing. I'll hope to have more changes later today.

dtorop:
More changes pushed to this branch:
- all scopes (not just regular histogram) respect the "histogram profile" (previously waveform and RGB parade were in display profile)
- use imageio_jpeg rather than GdkPixbufLoader to read live view image

This lets the code make fewer intermediary conversions in tether and live view, and generally simplifies code.

In order to not crash darktable, if the histogram profile is work or export, histograms in tether mode silently bypass the colorspace conversion. (In darkroom view these profiles can be determined from the pixelpipe.) Hence the histogram will quite possibly be in sRGB in these cases.

TurboGit:
@dtorop : Perfect ! I'll try to have a look today or tomorrow. If all goes well I'll merge to have more field testing.

dtorop:
@TurboGit : That's great. I'll be quite interested on how it looks.

Indeed field testing will matter. So far I've only tested this with one camera (EOS 5D Mark III), it's possible that there would be quirks with other camera setups.

The hope is that the changes will leave the histogram as-good-or-better for the main use case (darkroom view). And that they'll be useful for tether and live view -- a good enough histogram to judge exposure, even if not as precise as what comes through in darkroom view.

I do have one more idea to try: using the export mechanism to get the tethered image data rather than (as now) pulling it off of what is drawn in the center view. The advantage would be clearer/shorter code and having to do one less colorspace transform. But I don't want to keep making changes to put off testing/review!

src/libs/live_view.c
```diff
@@ -479,6 +479,24 @@ void view_enter(struct dt_lib_module_t *self,struct dt_view_t *old_view,struct d
...
+{
+  dt_lib_live_view_t *lib = self->data;
+
+  // there's no cdode to automatically restart live view when entering
+  // the view, and besides the user may not want to jump right back
+  // into leave view if they've been out of tethering view doing other
+  // things
+  if(gtk_toggle_button_get_active(GTK_TOGGLE_BUTTON(lib->live_view)) == TRUE)
+  {
+#if 0
```
TurboGit, line 492: I have a fix for this one.

src/libs/lib.h
TurboGit: It is not protected but cannot be broken as a gboolean is read/written in an atomic operation anyway. So I suppose this is ok.

src/libs/histogram.c
```diff
@@ -96,14 +128,228 @@ int position()
...
-  dt_control_queue_redraw_widget(self->widget);
+  dt_dev_histogram_collection_params_t histogram_params = { 0 };
+  const dt_iop_colorspace_type_t cst = iop_cs_rgb;
+  dt_dev_histogram_stats_t histogram_stats = { .bins_count = HISTOGRAM_BINS, .ch = 4, .pixels = 0 };
+  uint32_t histogram_max[4] = { 0 };
+  dt_histogram_roi_t histogram_roi = { .width = width, .height = height,
+                                      .crop_x = 0, .crop_y = 0, .crop_width = 0, .crop_height = 0 };
+
+  // Constraining the area if the colorpicker is active in area mode
+  // FIXME: if we do allow roi, then dt_lib_histogram_process() will need to take roi as a param, and this work of determining roi would happen in pixelpipe -- or the color_picker_box/point would move from dev->gui_module to darktable.lib->proxy.histogram
```
TurboGit, line 141: I suppose this is the part that breaks the histogram constrained to the current color picker area.

src/libs/live_view.c
dtorop: That's great. I didn't look at this hard -- I just sketched out code and when it crashed left it off.

src/views/view.c
```diff
@@ -1068,6 +1069,39 @@ int dt_view_image_get_surface(int imgid, int width, int height, cairo_surface_t
     cairo_destroy(cr);
   }
 
+  // hack: in tether mode, use mipmap as adequate data to create a histogram
```
dtorop, line 1072: I just tried using export code in place of this. Results seem better, and there's no need to use noise to fix quantization error. I'll work on this a moment more and then add one more commit.

src/libs/live_view.c
TurboGit: Fixed in b3f830c43fce0abe4e0fcd14dcdf9c4f20c55020, you can reactivate this part.

src/common/camera_control.c
```diff
@@ -485,6 +489,7 @@ static void *dt_camctl_camera_get_live_view(void *data)
 
   int frames = 0;
   double capture_time = dt_get_wtime();
+  int fps = dt_conf_get_int("plugins/capture/camera/live_view_fps");
```
TurboGit, line 492: can probably be a const.

dtorop:
> I have done quite some testing and got no crash. I have fixed the issue with leaving when livre-view is active (see comment in code).

That is great.

> There is only one missing part, the histogram restricted to the current color-picker area is not supported.

This has been broken for a while, I think -- even before this PR. I will look into making this work.

dtorop:
I'm adding one more commit now -- it gets rid of the hack of using 8-bit mipmaps in the tether view histogram. This should mean that the tether view histogram will match the one in darkroom view. Exception is if the histogram profile is set to "work" or "export".

TurboGit:
> This has been broken for a while, I think -- even before this PR. I will look into making this work.

It works for me on master but of course only for the standard histogram as there is no point having the waveform using a ROI since it represents the image width on the X axis.

src/libs/live_view.c
dtorop: > Fixed in b3f830c, you can reactivate this part.

Just put up a commit to do that. I won't rebase this branch on top of b3f830c, to keep this PR cleaner.

dtorop:
> It works for me on master but of course only for the standard histogram as there is no point having the waveform using a ROI since it represents the image width on the X axis.

Ah! I didn't even test it, because the code looked broken. I'll work on this.

I'm curious what a ROI for a waveform histogram would look like. I guess simply a waveform of a smaller portion of that image. Something to look into, but not for this PR.

dtorop:
> There is only one missing part, the histogram restricted to the current color-picker area is not supported.

This should be fixed in the most recent commit.

TurboGit:
All good to me now. I'll merge this now. Thanks for restoring histogram in the tether view and reworking a good part of the histogram code!

dtorop:
> All good to me now. I'll merge this now. Thanks for restoring histogram in the tether view and reworking a good part of the histogram code!

@TurboGit : Thank you for testing/merging. Definitely let me know if there are bugs which come up around this.

This should open things up for some other in progress histogram/scope work.

PR #5721, opened 2020-07-08
Restore slider digits where lost in introspection changes

### LINKED ISSUE #5707: basic adjustments: black level correction
**Describe the bug**

'black level correction' in 'basic adjustments' works with very coarse 0.01 steps

**To Reproduce**

1. Go to 'basic adjustment'
2. Move slide of 'black level correction'

**Expected behavior**

The same fine adjustment (steps of 0.0001) of the 'exposure' module

 - Darktable Version: 3.1.0+2422
 - OS: macOS Catalina 15.5.0

dterrahe:
Closes #5707

Hopefully this resolves any serious regressions people experience; nothing more jumps out at me in [slider_params.xlsx](https://github.com/darktable-org/darktable/files/4890154/slider_params.xlsx)

This should still be revisited as part of a larger input layer review: 
- format&digits should be automatically linked in bauhaus
- rounding to digits might only be needed for integers
- stepsize could be a fixed proportion (1/100th?) of the (current) min/max range

But since this will lead to changes people won't like, there needs to be a way to configure granularly (via lua?) which might be relatively easy/free to achieve as part of a new input layer approach.

All way past 3.2, hence this quick fix for the release.

### DIFF
--- a/src/iop/clipping.c
+++ b/src/iop/clipping.c
@@ -2347,22 +2347,26 @@ void gui_init(struct dt_iop_module_t *self)
   GtkWidget *page2 = self->widget = GTK_WIDGET(gtk_box_new(GTK_ORIENTATION_VERTICAL, 0));
 
   g->cx = dt_bauhaus_slider_from_params(self, "cx");
+  dt_bauhaus_slider_set_digits(g->cx, 4);
   dt_bauhaus_slider_set_factor(g->cx, 100.0);
   dt_bauhaus_slider_set_format(g->cx, "%0.f %%");
   gtk_widget_set_tooltip_text(g->cx, _("the left margin cannot overlap with the right margin"));
 
   g->cw = dt_bauhaus_slider_from_params(self, "cw");
+  dt_bauhaus_slider_set_digits(g->cw, 4);
   dt_bauhaus_slider_set_factor(g->cw, -100.0);
   dt_bauhaus_slider_set_offset(g->cw, 100.0);
   dt_bauhaus_slider_set_format(g->cw, "%0.f %%");
   gtk_widget_set_tooltip_text(g->cw, _("the right margin cannot overlap with the left margin"));
 
   g->cy = dt_bauhaus_slider_from_params(self, "cy");
+  dt_bauhaus_slider_set_digits(g->cy, 4);
   dt_bauhaus_slider_set_factor(g->cy, 100.0);
   dt_bauhaus_slider_set_format(g->cy, "%0.f %%");
   gtk_widget_set_tooltip_text(g->cy, _("the top margin cannot overlap with the bottom margin"));
 
   g->ch = dt_bauhaus_slider_from_params(self, "ch");
+  dt_bauhaus_slider_set_digits(g->ch, 4);
   dt_bauhaus_slider_set_factor(g->ch, -100.0);
   dt_bauhaus_slider_set_offset(g->ch, 100.0);
   dt_bauhaus_slider_set_format(g->ch, "%0.f %%");
--- a/src/iop/basicadj.c
+++ b/src/iop/basicadj.c
@@ -621,6 +621,7 @@ void gui_init(struct dt_iop_module_t *self)
   g->sl_black_point = dt_bauhaus_slider_from_params(self, "black_point");
   dt_bauhaus_slider_set_soft_range(g->sl_black_point, -0.1, 0.1);
   dt_bauhaus_slider_set_step(g->sl_black_point, .001);
+  dt_bauhaus_slider_set_digits(g->sl_black_point, 4);
   gtk_widget_set_tooltip_text(g->sl_black_point, _("adjust the black level to unclip negative RGB values.\n"
                                                     "you should never use it to add more density in blacks!\n"
                                                     "if poorly set, it will clip near-black colors out of gamut\n"
@@ -679,6 +680,7 @@ void gui_init(struct dt_iop_module_t *self)
   gtk_box_pack_start(GTK_BOX(self->widget), autolevels_box, TRUE, TRUE, 0);
 
   g->sl_clip = dt_bauhaus_slider_from_params(self, N_("clip"));
+  dt_bauhaus_slider_set_digits(g->sl_clip, 3);
   gtk_widget_set_tooltip_text(g->sl_clip, _("adjusts clipping value for auto exposure calculation"));
 
   // add signal handler for preview pipe finish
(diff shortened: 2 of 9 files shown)


TurboGit:
This is for 3.2 right? I'm not sure to understand your last sentence (I'm not native English).

dterrahe:
Yes, sorry; I meant the other meanderings about a long term solution/approach are for _after_ 3.2. Therefore the short-term fixes in this PR should go into 3.2 to not cause regressions.

src/iop/clipping.c
```diff
@@ -2347,22 +2347,26 @@ void gui_init(struct dt_iop_module_t *self)
   GtkWidget *page2 = self->widget = GTK_WIDGET(gtk_box_new(GTK_ORIENTATION_VERTICAL, 0));
 
   g->cx = dt_bauhaus_slider_from_params(self, "cx");
+  dt_bauhaus_slider_set_digits(g->cx, 4);
   dt_bauhaus_slider_set_factor(g->cx, 100.0);
   dt_bauhaus_slider_set_format(g->cx, "%0.f %%");
```
TurboGit, line 2352: Don't we need %0.2 here ? Otherwise the values are displayed as plain integer in the interface. Likewise below.

TurboGit: In fact we need only "%0.1"

dterrahe: Don't _need_ to. There are many places where there are more digits than are visible. I didn't intend to "clean" all of those up, as it may not be a bad thing anyway. As i said, i think we could get rid of rounding to nearest "digit" completely, except for integers.

TurboGit: Ok, let's merge this then. I understand that there is probably many cases like this. I think it makes sense to at least fix this one for the release. I'll issue a PR.

dterrahe: To be clear, I didn't consider this a bug and that's why I didn't "fix" it. Rather than showing a lot of digits in the format, the user should just look at the image to see if the cropping is correct.

If you _do_ for some reason want to enforce "consistency" between the number of digits and the format, the cases where they don't match are fairly easy to spot in the spreadsheet I attached. The fixes are all trivial. Again, I didn't do that because I didn't consider it a good idea, but it is up to you.

The number of digits itself (not the format) _is_ important, because the slider is rounding to them, so if not enough digits are set, the user is unable (even invisibly) to get more precision. That's why I submitted this PR to fix such regressions.

TurboGit: > If you _do_ for some reason want to enforce "consistency" between the number of digits and the format,

That's not about consistency. What I want is also a fix, on the crop&rotate if you ctrl-scroll on any slider in margin tab, you have no feedback about the change in the value. You need to do 10 changes to see the value go from 0% to 1% for example. I consider this a bug. And this is a regression (due to this PR) as previously the scroll or ctrl-scroll was both changing the value by step of 1%, so always having a visual feedback.

Also, if you edit the slider and enter .5 the value displayed is still 0%. Really a bug to me.

Hope this clarify my thinking.

dterrahe:
This list may be helpful.

```
module		label		step	digits	factor	format
watermark	opacity		1	2	1	%.f%%
watermark	scale		1	2	1	%.f%%
dither		damping		1	3	1	%.0fdB
velvia		strength	1	2	1	%.0f%%
soften		size		1	2	1	%.0f%%
soften		saturation	1	2	1	%.0f%%
soften		mix		1	2	1	%.0f%%
grain		strength	1	2	1	%.0f%%
grain		midtones bias	1	2	1	%.0f%%
highpass	sharpness	1	2	1	%.0f%%
highpass	contrast boost	1	2	1	%.0f%%
relight		width		0.5	2	1	%.1f
levels		black		0.1	2	1	%.1f%%
levels		gray		0.1	2	1	%.1f%%
levels		white		0.1	2	1	%.1f%%
filmic		latitude	0.05	3	1	%.2f EV
colorzones	mix		10	2	1	%.01f%%
bloom		size		1	2	1	%.0f%%
bloom		threshold	1	2	1	%.0f%%
bloom		strength	1	2	1	%.0f%%
vibrance	vibrance	1	2	1	%.0f%%
graduatednd	hardness	1	2	1	%.0f%%
ashift		focal length	1	2	1	%.0fmm
ashift		lens dependence	1	2	1	%.0f%%
exposure	exposure	0.02	3	1	%.2f EV
tonemap		spatial extent	1	2	1	%.0f%%

```

Only some of those are actual regressions vs 3.0.2.

What I intended to do for 3.4 was to explicitly link digits and format in all cases. Now that only works for the standard format ("%.1f"; setting number of digits also changes the format) but if there needs to be an extension, it needs to be set (again) with a full format ("%.1f%%", "%.2f EV") and it will be overwritten if the number of digits is changed. If instead going forward the format were set with ..._set_format("%%") it could just be concatenated.

Actually, I wanted to do this in `dt_bauhaus_slider_get_text` with

```
if(d->min < 0)
return g_strdup_printf("%+.*f%s", d->digits, dt_bauhaus_slider_get_val(w), d->format);
else
return g_strdup_printf("%.*f%s", d->digits, dt_bauhaus_slider_get_val(w), d->format);
```

The question is, should the number of digits automatically be adjusted if the step size is set/changed. dt_bauhause_slider_from_params does this, but it doesn't take "small" steps (holding CTRL) into account (and is overwritten if an explicit format with extension is set).

_Maybe_ it would be nice to dynamically adjust `digits`. So for example by default it could be %.0f if min-max=0-100 and step is 1. But then if the user moves a small step (ctrl-scroll) to 0.1 (or enters it directly) an extra digits would be added (and kept until moving to a different picture/reentering darkroom/resetting the slider). Or if a picture is loaded that has a value 0.123 set, digits would be adjusted to 3.

I _think_ this could work well (also when users set a custom multiplication factor for ctrl-scroll). Except for values that are calculated by formula and that would have infinite digits...

dterrahe:
To be clear; this would be a larger refactor again, touching quite a few bits in bauhaus and needing changes in all modules in settings formats/steps/digits. As part of that I would also like to make soft-min/-max user configurable (both by UI and via lua). And this should be based on a revised input layer exposing this functionality, so quite a few dependencies here. I'll of course write all of this up for discussion/commenting before I start.

TurboGit:
Understood, so maybe we should live with the "issue" I reported until 3.4 is out and close #5724 ?

dterrahe:
Either way. I have no objection to #5724. I was basically just explaining why I hadn't made that change myself. (So that you didn't think I was just lazy ;-) )

But now that the PR is there, you might as well merge it if you think it looks ok.

PR #5723, opened 2020-07-08
bugfix #5712, add gui_reset to liquify modules

### LINKED ISSUE #5712: liquefier module freezes
Hello,
I wanted to see what the liquefier module was. ( out of curiosity )

I have the following problem:

**activate the module
select: line tool ( for example )
draw no lines
do not deselect the tool
do: initialize the parameters, from the module
In this case, the selected tool does not reset and the tool cannot be changed.
you have to leave darktable.**

Ubuntu 20.04 dt 3.1.0 git 2422

I did the same test under W10 and dt 3.0.2 and I have the same behavior !!!

jenshannoschwalm:
There is a bug report #5712 describing the issue.

I think the module only misses the gui_reset callback, so this
pr adds this. I think it's safe for this bugfix to be for 3.2

Fixes #5712

### DIFF
--- a/src/iop/liquify.c
+++ b/src/iop/liquify.c
@@ -3589,6 +3589,16 @@ void gui_init(dt_iop_module_t *module)
                                                               "ctrl+click: linear, grow, and shrink");
 }
 
+void gui_reset(dt_iop_module_t *self)
+{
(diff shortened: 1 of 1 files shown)


TurboGit:
Looks good, thanks !

PR #5724, opened 2020-07-08
clipping: display more digits.

TurboGit:
This makes changing the value with shift-scroll visible.


TurboGit:
Not the final solution see discussion in #5721 but cannot hurt for 3.2.

PR #5726, opened 2020-07-09
Temperature wandering fix

### LINKED ISSUE #5725: white balance: temperature/tint adjustment interlock
**Describe the bug**
For certain temperature/tint combinations in the white balance module, adjusting temperature will alter the tint slider. Successive temperature adjustments can cause tint to wander.

**To Reproduce**
1. Open an image in darkroom view
2. Set tint to 1
3. Set temperature to 4000K
4. Tint will jump to 0.983, temperature to 3901K
5. Scroll temperature slider to bring temperature up by units of 10
6. When temperature passes 3960K, temperature will jump to 4072K, tint to 1.001
7. Control-click on temperature slider and move the fine-control adjustment back and forth across the 39xxK -> 40xxK gap
8. The tint will progressively increase/decrease

**Expected behavior**
Tint should remain invariant when temperature is altered.

- Darktable Version: 536df44ad9215c45d222cd40f65e80f009a4f9d9
- OS: Debian bullseye/sid, Gnome 3.36.3
- OpenCL activated: yes
- Intel(R) Core(TM) i7-7820HQ CPU @ 2.90GHz

**Additional context**
- 42284087d0db3e4521d43a628c8f78b44c963806 added a call to set the preset to "user modified" in both tint_callback() and temp_callback() in iop/temperature.c. This results in a call to gui_update_from_coeffs() which recalculates temp/tint via temperature_to_XYZ() which may have a discontinuity at 4000K. A solution may be to not call gui_update_from_coeffs() in apply_preset() when the new preset is "user modified".
- I can reproduce this either a raw file or a JPEG
- I can reproduce this with a fresh edit (no history)
- I can reproduce this with empty/new config dir

dtorop:
This is an attempt to fix #5725.

The first commit is the least intrusive, and fixes all except in the case of switch from then back to "user modified" preset. The second combines two callbacks and a helper function into a single function. The third commit is a more full fix, which also keeps temperature/tint stable when switching from/to "user modified".

### DIFF
--- a/src/iop/temperature.c
+++ b/src/iop/temperature.c
@@ -82,6 +82,7 @@ typedef struct dt_iop_temperature_gui_data_t
   int preset_num[50];
   double daylight_wb[4];
   double mod_coeff[4];
+  double mod_temp, mod_tint;
   double XYZ_to_CAM[4][3], CAM_to_XYZ[3][4];
 } dt_iop_temperature_gui_data_t;
 
@@ -780,6 +781,8 @@ void gui_update(struct dt_iop_module_t *self)
 
   gui_sliders_update(self);
   for(int k = 0; k < 4; k++) g->mod_coeff[k] = p->coeffs[k];
+  g->mod_temp = TempK;
+  g->mod_tint = tint;
 
   dt_bauhaus_combobox_clear(g->presets);
   dt_bauhaus_combobox_add(g->presets, C_("white balance", "camera"));
@@ -1190,6 +1193,8 @@ static void temp_tint_callback(GtkWidget *slider, gpointer user_data)
   coeffs[3] /= coeffs[1];
   coeffs[1] = 1.0;
   for(int c = 0; c < 4; c++) p->coeffs[c] = g->mod_coeff[c] = coeffs[c];
+  g->mod_temp = TempK;
+  g->mod_tint = tint;
 
   ++darktable.gui->reset;
   dt_bauhaus_slider_set(g->scale_r, p->coeffs[0]);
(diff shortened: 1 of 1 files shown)


dtorop:
Thinking this over, I'm not sure if 13bc219 makes sense. It's to keep the temperature/tint from shifting slightly (but consistently) when switching from "user modified" to another preset than back to "user modified" when temperature is near 4000K. This will happen regardless when a user switches to a different image then back to the same image, as temp/tint aren't stored, only the RGB coeffs.

It seems to me that a0c871e does make sense -- it fixes a UI glitch which can make white balance hard to use when temperature is near 4000K. But 13bc219 is just there to paper over that the iop really only cares about RGB coeffs, and is translating temp/tint into these using the best model that it can. The real fix might be to improve the model/translation to handle the split in reference at 4000K.

TurboGit:
Let's review that for 3.4.

dtorop:
> Let's review that for 3.4.

Sounds fair. It's an unlikely bug to stumble onto.

johnny-bit:
Actually it's a side-effect of temperature calculation. in my #4360 i also noticed (with Aurelien's help) that tint calculation is wrong as it's not along normal to planckian locus. That and temp+ tint <-> coeffs calculation is nearly always unstable due to possible many solutions.

I'll try to find time and bring my PRs up to speed on current dt ver and address most issues.

dtorop:
> Actually it's a side-effect of temperature calculation ... temp+ tint <-> coeffs calculation is nearly always unstable due to possible many solutions

@johnny-bit Indeed, the instability is the root of this. Up until 42284087d0db3e4521d43a628c8f78b44c963806, if a user set color temperature near 4000K, the pretty-close RGB coefficients would be calculated, but temp/tint would remain (at least until the user reloaded the image) as the user chose. Hence the instability wasn't that visible.

This is compounded by implementing the "user modified" preset, as now on every change to the temperature or tint sliders, RGB coefficients are calculated, then the temperature/tint sliders are recalculated based on the coefficients. Normally, this would result in pretty much the same temp/tint, so all was well. The catch is that near 4000K, the recalculation isn't quite what the user has chosen. And if using fine-control adjustment on temperature, the numerous recalculations end up making tint skew significantly.

(I stumbled across this and thought that there was a ghost altering the tint slider -- and eventually it skewed so far that the image had a substantial color cast.)

The quick fix seems to be just to not recalculate temp/tint from coefficients once the initial switch to "user modified" preset occurs (a0c871ed9f687d3e85f3d8a55cb419f6c99740ac).

I've been following #4360 with interest -- both because it looks great, and because all the issues/questions raised along the way are pretty fascinating. If it can fix #5725, that is great.

TurboGit:
Fixed in #4360.

PR #5728, opened 2020-07-09
Make branding consistent as per faq

kmilos:


aurelienpierre:
The folders names and desktop shortcuts should be preserved as-is for compatibility. Also, for OS shortcut, we abide by Freedesktop standards which is to capitalize apps names.

kmilos:
Fair enough, reverted the entries in question now. However, I discovered some inconsistency in the process, there is both small caps and capitalized path in `data/darktableconfig.xml.in` (see the very last commit)

TurboGit (approved):
Looks ok to me, thanks!

Issue #5735, opened 2020-07-11
[darkroom] filmstrip jumps to first pic when changing rating-view


AxelG-DE:
**Describe the bug**

When changing the rating view e.g. between 3* back to 1*, the filmstrip jumps to first pic in collection.

**To Reproduce**
1. Go to 'darkroom' with a collection having approx. 50 pics in 3* and many more in 1*. place your fokus somewhere in the middle of 3* selection
2. Change from 3* down to 1*
3. See error

**Expected behavior**
Same than lighttable, the focus shall stay centred from where the user left one rating to another (if in the new collection is no such pic, stay very near to that position in the collection)

**Platform (please complete the following information):**
* darktable version : 3.1.0+2448~g4b77a5410
* OS                : Linux - kernel 5.7.7-gentoo
* Distro            : Gentoo Base System release 2.6
* Processor         : Intel(R) Core(TM) i9-9900K CPU @ 3.60GHz
* Memory            : 32 GB (4 x 8 GB) + 5GB Swap
* Graphics card0    : GeForce GTX 1060 6GB
* Graphics card1    : GeForce RTX 2070 SUPER
* Graphics driver   : nvidia-drivers-440.100-r1
* OpenCL installed  : Yes (opencl-headers-2020.03.13-r1)
* OpenCL activated  : Yes
* Xorg              : xorg-server-1.20.8
* Desktop           : KDE 5
* GTK+              : gtk+-3.24.20
* gcc               : x86_64-pc-linux-gnu-9.3.0
* cflags            : CFLAGS="-march=native-O2-mtune=native-pipe"
* CMAKE_BUILD_TYPE  : "Release"

AxelG-DE:
@AlicVB may I ping you here? :slightly_smiling_face:

AlicVB:
@AxelG-DE : this is more a feature request than an issue, as it's not implemented now (in filemanager too) currently, when you refilter the collection, the offset is keep only if the image is still in the new collection, otherwise, it's reseted to beginning.

Implementing the case you want would require quite a lot of work and I'm not even sure there's good way to do it as the collection may very well have changed completely... Need more thought !

I've just fixed a small glitch in filmstrip where in this case, the filmstrip is moved to the right with no centered image.

Nilvus:
[...]

PR #5745, opened 2020-07-13
filmstrip : ensure rowid is always valid

AlicVB:
this avoid to have filmstrip on the right with no centered images in cases like #5735


TurboGit:
Looks good, thanks!

PR #5751, opened 2020-07-14
lut3d: add check to prevent buffer overrun and ensure string zero termination

rabauke:
Fixes a bug that causes buffer overruns when reading ill-formed lut files. Some minor clean-up regarding pointer arithmetics.


TurboGit:
Looks ok to me, @phweyland you may want to double check and approve this change. TIA.

phweyland:
Always the opportunity to learn a bit more !
Looks fine for me. Thanks.

TurboGit:
Thanks to both of you !

PR #5755, opened 2020-07-15
allow shift or control modifiers to scroll slider even with sidebar_scroll_default

### LINKED ISSUE #4899: Ctrl+scroll on module cursors
Hi, 
When the miscellaneous pref "mouse wheel scrolls modules side panel by default" is not ticked you may use your mouse wheel to interact with some modules cursors just scrolling, Shift scrolling and Ctrl scrolling over a cursor while the pointer is hovering it. 
And that's very fine!

Now, for some reason you may prefer having this pref option ticked so that your mouse wheel interact only with the side panel and then you have to Ctrl + Alt + mouse wheel over the cursors to adjust them the way you want. This is well documented, that's OK.

Right, but what about the fine tuning you had with Ctrl + scroll before? 

"Ctrl + Alt + mouse wheel" replaces "mouse wheel", but it seems there's no combination to replace "Ctrl + mouse wheel" for fine tuning, such as for instance "Ctrl + Shift + mouse wheel" (or even Ctrl+Alt+Shift...).

Would it be possible to implement something in that way?

### LINKED ISSUE #5740: X and Y values for Watermark are too coarse
Previously Watermark supported values with 3 decimal places and this made it possible to locate the watermark (I'm using Simple Text) accurately.  E.g. putting your name in the corner of an image.
Recently it has changed and only 2 dec places are now used.  This makes it impossible to put the text exactly where you want.  Right-clicking the X or Y values allows 3 dec places to be entered, but the value is then rounded to 2 places.
Please could we return to the previous functionality.

Master 3.1.0+2440~g673f25287-dirty  2 dec places
Master 3.1.0+2017~ga6c18e1b7-dirty (around 9 June)  3 dec places

Ubuntu 18.04

### LINKED ISSUE #5833: Gradient sliders don't respect "scrolls modules side panel by default"
Gradient sliders in iop modules change if the mouse is over them and the mouse wheel is scrolled, independent of whether the setting "miscellaneous/mouse wheel scrolls modules side panel by default" is active or not. This also applies to the multi-value gradient sliders in the parametric blending settings. As a result, when scrolling through the list of modules, if any module with these types of sliders is open, scrolling will get stuck there and values will get changed.

Expected behaviour; when the option is selected, scrolling should be ignored by the slider and the side panel should continue to scroll. If the slider needs to be moved, this should be achieved by pressing ctrl-alt, as it is with all other kinds of widgets.

This is on debian and windows. Current master.

dterrahe:
When "mouse wheel scrolls modules side panel by default" is enabled, only CTRL-ALT+scroll was moving sliders. With this change, scroll with any modifiers is passed through and CTRL-ALT is treated as if no modifier. This allows all three speeds (precise-normal-coarse) to be used.

Just ALT+scroll will currently also work as a "normal" scroll, but if this is given a special meaning in future, then ctrl-alt will still be available to mean "normal"/no-modifier.

fixes #4899 and hence also fixes #5740

29/07: now also fixes #5833

### DIFF
--- a/src/bauhaus/bauhaus.c
+++ b/src/bauhaus/bauhaus.c
@@ -2043,7 +2043,9 @@ static gboolean dt_bauhaus_slider_scroll(GtkWidget *widget, GdkEventScroll *even
   gdouble delta_y;
   dt_bauhaus_widget_t *w = (dt_bauhaus_widget_t *)widget;
   if(w->type != DT_BAUHAUS_SLIDER) return FALSE;
-  if(((event->state & gtk_accelerator_get_default_mod_mask()) == darktable.gui->sidebar_scroll_mask) != dt_conf_get_bool("darkroom/ui/sidebar_scroll_default")) return FALSE;
+
+  if(dt_gui_ignore_scroll(event)) return FALSE;
+
   gtk_widget_grab_focus(widget);
 
   if(dt_gui_get_scroll_delta(event, &delta_y))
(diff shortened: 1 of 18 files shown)


dterrahe:
closes #5833

dterrahe:
Fixed merge conflict

TurboGit (approved):
Works for me ! Thanks!

PR #5756, opened 2020-07-16
Let minimum width to image infos when reducing panel width

Nilvus:
If we reduce left panel width, image infos (right column) of related module became quickly hidden. Horizontal scrollbar appear too late. This little tweak fix that and allow the right column of this module to keep minimum width for the infos and so let horizontal scrollbar more quickly if we reduce panel width.

Only side-effect is that also change how text in right column is aligned. It's now centered instead of left aligned. But CSS alignment setting is not supported with Gtk. Having those infos aligned centered is good for me.

Of course, it's a sort of bugfix and simple one, so for 3.2.

### DIFF
--- a/data/themes/darktable.css
+++ b/data/themes/darktable.css
@@ -1325,6 +1325,13 @@ progressbar progress
   border-radius: 4px;
 }
 
+/* Be sure to have minimum width allowed for infos on images infos module if reducing panel width */
+#brightbg label
+{
+  min-width: 80px;
+  padding-left: 2px;
+  padding-bottom: 1px;
+}
 
 /* Checkbutton check state */
 #lib-plugin-ui checkbutton check,


TurboGit:
To me the centered text is not ok! It makes the module hard to read.

Nilvus:
> To me the centered text is not ok! It makes the module hard to read.

I don't find so but ok, to considerate. I don't have solution with CSS and do not understand why just adding a minimum width change that. Considering only CSS, this has no sense, so probably something to check on Gtk code. But as you know, for that, I need someone to check.

Nilvus:
> > To me the centered text is not ok! It makes the module hard to read.
> 
> I don't find so but ok, to considerate. I don't have solution with CSS and do not understand why just adding a minimum width change that. Considering only CSS, this has no sense, so probably something to check on Gtk code. But as you know, for that, I need someone to check.

Align on the left could be hardcoded on Gtk code as it can't be with CSS.

TurboGit:
@Nilvus : I'll see what can be done on C side.

Nilvus:
> @Nilvus : I'll see what can be done on C side.

Thanks. I just check and find that is it on src/libs/metadata_view.c. On what I find (and understand on this page: https://developer.gnome.org/gtk3/stable/GtkWidget.html#GtkAlign), the alignment seems to be applied on the left (lines 826 et 827) but that's not applied correctly, or it miss something. I can't be further there.

Nilvus:
> > @Nilvus : I'll see what can be done on C side.
> 
> Thanks. I just check and find that is it on src/libs/metadata_view.c. On what I find (and understand) on this page: https://developer.gnome.org/gtk3/stable/GtkWidget.html#GtkAlign, the alignment seems to be applied on the left (lines 826 et 827) but that's not applied correctly, or it miss something. I can't be further there.

Nilvus:
Last thing: I also just add some padding (little ones) to aerate a little bit all those infos.

TurboGit:
See #5758 for a patch on top of yours which makes the alignment to the left.

TurboGit:
The issue was the event-box which are breaking the alignment somehow.

Nilvus:
Ok, so close this PR.

PR #5757, opened 2020-07-16
This patch almost disable any hightlight reconstruction by default.

TurboGit:
The goal is to make the module faster by default and let user activate
the reconstruction only when necessary.


Nilvus:
Just to support that by default!

jenshannoschwalm:
1. HUGE improvement performancewise
2. I think it does not hurt too much
3. Also helps a lot with thumbnail generation performance

PR #5758, opened 2020-07-16
min width

TurboGit:


Nilvus:
Thanks but seems a bugfix and with no real risk (and in the idea of improving UI) so why not set it to 3.2?

Nilvus:
Just tested, and my CSS tweak doesn't have any effect now. So, that would need to add a tag to label of right column. Before, you had widget called brightbg and label on child. Now all labels are on same level and so those who brightbg tag are those on left column and we want min-width on right column. As brightbg tag is not used, could you apply it to label on right column and so remove label on #brightbg line in CSS?

TurboGit:
@Nilvus : the name of the widget is still there (#brightbg) but now it is directly on the label, so I suppose:

```
#brightbg label
{
```

Should be changed to just:

```
#brightbg
{
```

Or maybe there is something I do not understand/see ?

Nilvus:
Oups, my bad, just didn't check correctly. You're right. Anyway, if you change that, min-width and padding is now added and so ok,  but text is no more aligned left and come back centered... Still something wrong on your code.

TurboGit:
Indeed, you're right meaning my fix is not a fix but just a clean-up... I still don't understand this alignment issue then :(

TurboGit:
@Nilvus : Should be working now. Can you double check all is fine ?

Nilvus:
> @Nilvus : Should be working now. Can you double check all is fine ?

All is fine. Thanks, good work. And good to see that for 3.2.

PR #5762, opened 2020-07-17
filmic v4: add OpenCL kernel

aurelienpierre:
* improve CPU path SIMD code sharing, libify common vector operations
* add CPU path SIMD variant for noise generation and libify it
* optimize CPU wavelet reconstruction
* add OpenCL kernel for tonemapping only (so far), highlight reconstruction missing (for now)

NOTE: you may need to start darktable twice after building this so the OpenCL kernels are properly refreshed.


MStraeten:
I found a current build with filmic v4 opencl to be consuming at least double memory in comparison to a build without this pr. Editing a whole bunch of images can cause darktable to consume several gigabytes of memory. I need to do some further test on my osx machine to be able to quantify it ...

aurelienpierre:
Integration test returns (this PR against master):
```
sh run.sh 0037-filmic-reconstruction/
Test 0037-filmic-reconstruction
      Image mire1.cr2
      CPU & GPU version differ by 44089 pixels
      Max  dE         2.4523
      Mean dE         0.0025
      Count above max 5
  FAILS: image visually changed
         (31946 pixels changed)


Total test 1
Errors     1
```

src/develop/noise_generator.h
```diff
@@ -0,0 +1,253 @@
+/*
+   This file is part of darktable,
+   Copyright (C) 2020 - Aurélien Pierre.
```
TurboGit, line 3: you are reintroducing "deprecated" copyright.

data/kernels/noise_generator.h
```diff
@@ -0,0 +1,155 @@
+/*
+   This file is part of darktable,
+   Copyright (C) 2020 - Aurélien Pierre.
```
TurboGit, line 3: you are reintroducing "deprecated" copyright.

data/kernels/filmic.cl
```diff
@@ -1,6 +1,6 @@
 /*
     This file is part of darktable,
-    copyright (c) 2018 Aurélien Pierre.
+    copyright (c) 2018-2020 Aurélien Pierre.
```
TurboGit, line 3: I suppose all those .cl files got missed when rewritten.

TurboGit (changes requested):
I fail to understand the need for changing the tethering view here. Can you clarify ?

aurelienpierre:
@TurboGit tethering view used the noise generator introduced for filmic v4. The CPU logic of the noise generator was not possible with GPU and also used double precision, which is really slow if supported at all on GPU. So I rewrote the noise generator for single precision and synchronized the CPU and GPU logic to get consistent results. The change in tethering is only to update the noise API.

src/develop/noise_generator.h
aurelienpierre: Wasn't the decision to reintroduce individual copyright ? I don't mind much either way.

TurboGit: No real decision was made, Ulrich was supposed to revert this part and have proper copyright notice. At this stage I suppose we want to be consistent.

TurboGit:
@aurelienpierre : Thanks for the explanation. I'll review this tomorrow anyway. Too late for me now, and this is more a bug fix that any sort of new feature so no problem to merge tomorrow or later. Thanks for the hard work you put on this for the 3.4 release.

src/develop/noise_generator.h
aurelienpierre: No problem.

elstoc:
Just had a play with this and it really makes all the difference with highlight reconstruction. Thanks for all the work on this @aurelienpierre.

aurelienpierre:
> Just had a play with this and it really makes all the difference with highlight reconstruction. Thanks for all the work on this @aurelienpierre.

Cool, because that one was an ordeal. Lots of unexpected problems, much more work than anticipated.

TurboGit (approved):
Works for me, it is day and night in speed when compared to the CPU version and makes the noise recovery usable with other IOP. Until then I disabled the recovery while working on the dev and enabled it back only at the end when needed. Thanks a lot, the code looks good to me. So merging now for more field testing before the release.

MStraeten:
There seems to be memory issues. stepping through several images in a folder by space and backspace and doing some edits increases the consumed memory over 5.5 GB (i got darktable consuming max 2 GB without openCL enabled )

TurboGit:
@MStraeten : I have fixed some memory leaks, can you try current master ?

MStraeten:
While editing it seems to be better, but exporting with high quality resampling active of 85 images blows the used memory over 14 GB. After exporting this memory usage doesn’t shrink ...

MStraeten:
For comparison: with option —disable-opencl the memory usage increases and decreases while processing the images (about 500MB to 3GB)

TurboGit:
@MStraeten : Ok found some other memory leaks in the OpenCL code. Can you double check current master ? TIA.

MStraeten:
did a test, it’s fine now

PR #5770, opened 2020-07-19
add history changes tooltip

dterrahe:
Uses introspection to show changes between revisions in the history module in darkroom.

Edit:
- also includes blending changes (and some refactoring of parametric blending gui code)
- tooltip gets generated dynamically, so does not have performance impact with every change
- basic formatting using tabs in a GtkTextView within the tooltip. This could be changed to a GtkTreeView with drilldown into curves/other arrays, but that would require a pop-up dialog and so would be part of a major revision at some point.

image

### DIFF
--- a/data/themes/darktable.css
+++ b/data/themes/darktable.css
@@ -174,11 +174,6 @@
   min-height: 0;
 }
 
-.monospace
-{
-  font-family: monospace;
-}
-
 /* Be sure now than all others UI widgets doesn't inherit font size already set and have no background color on start */
 #lib-plugin-ui,
 #iop-plugin-ui,
@@ -1417,7 +1412,8 @@ cell:selected
 
 /* Set some specific fonts */
 #live-sample-data,
-#usercss_box textview
+#usercss_box textview,
+#hist-tooltip
 {
   font-family: monospace;
 }
--- a/data/themes/darktable.css
+++ b/data/themes/darktable.css
@@ -174,6 +174,11 @@
   min-height: 0;
 }
 
+.monospace
+{
+  font-family: monospace;
+}
+
 /* Be sure now than all others UI widgets doesn't inherit font size already set and have no background color on start */
 #lib-plugin-ui,
 #iop-plugin-ui,
--- a/src/develop/blend.h
+++ b/src/develop/blend.h
@@ -366,7 +366,8 @@ extern const dt_develop_name_value_t dt_develop_mask_mode_names[];
 extern const dt_develop_name_value_t dt_develop_combine_masks_names[];
 extern const dt_develop_name_value_t dt_develop_feathering_guide_names[];
 extern const dt_develop_name_value_t dt_develop_invert_mask_names[];
-
+extern const dt_develop_name_value_t dt_develop_blendif_names_rgb[];
+extern const dt_develop_name_value_t dt_develop_blendif_names_lab[];
 
 #define DEVELOP_MASKS_NB_SHAPES 5
 
--- a/src/develop/blend.h
+++ b/src/develop/blend.h
@@ -354,18 +354,28 @@ typedef struct dt_iop_gui_blendif_colorstop_t
   GdkRGBA color;
 } dt_iop_gui_blendif_colorstop_t;
 
-typedef struct dt_iop_gui_blendif_slider_t
+typedef struct dt_iop_gui_blendif_channel_t
 {
   char *label;
   char *tooltip;
-  float increments;
+  float increment;
   int numberstops;
   const dt_iop_gui_blendif_colorstop_t *colorstops;
-  int channels[2];
+  dt_develop_blendif_channels_t param_channels[2];
   dt_dev_pixelpipe_display_mask_t display_channel;
   void (*scale_print)(float value, char *string, int n);
   int (*altdisplay)(GtkWidget *, dt_iop_module_t *, int);
-} dt_iop_gui_blendif_slider_t;
+  char *name;
+} dt_iop_gui_blendif_channel_t;
+
+typedef struct dt_iop_gui_blendif_filter_t
+{
+  GtkDarktableGradientSlider *slider;
+  GtkLabel *head;
+  GtkLabel *label[4];
+  GtkLabel *picker_label;
+  GtkWidget *polarity;
+} dt_iop_gui_blendif_filter_t;
 
 typedef struct dt_iop_blend_name_value_t
 {
@@ -390,10 +400,13 @@ typedef struct dt_iop_gui_blend_data_t
   int masks_support;
   int masks_inited;
   int raster_inited;
+
   dt_iop_colorspace_type_t csp;
   dt_iop_module_t *module;
+
   GList *masks_modes;
   GList *masks_modes_toggles;
+
   GtkWidget *iopw;
   GtkBox *top_box;
   GtkBox *bottom_box;
@@ -401,22 +414,13 @@ typedef struct dt_iop_gui_blend_data_t
   GtkBox *blendif_box;
   GtkBox *masks_box;
   GtkBox *raster_box;
-  GtkDarktableGradientSlider *upper_slider;
-  GtkDarktableGradientSlider *lower_slider;
-  GtkLabel *upper_head;
-  GtkLabel *lower_head;
-  GtkLabel *upper_label[8];
-  GtkLabel *lower_label[8];
-  GtkLabel *upper_picker_label;
-  GtkLabel *lower_picker_label;
+
   GtkWidget *selected_mask_mode;
-  GtkWidget *upper_polarity;
-  GtkWidget *lower_polarity;
   GtkWidget *colorpicker;
   GtkWidget *colorpicker_set_values;
+  dt_iop_gui_blendif_filter_t filter[2];
   GtkWidget *showmask;
   GtkWidget *suppress;
-  void (*scale_print[8])(float value, char *string, int n);
   GtkWidget *masks_combine_combo;
   GtkWidget *blend_modes_combo;
   GtkWidget *masks_invert_combo;
@@ -426,19 +430,13 @@ typedef struct dt_iop_gui_blend_data_t
   GtkWidget *blur_radius_slider;
   GtkWidget *contrast_slider;
   GtkWidget *brightness_slider;
+
+  const dt_iop_gui_blendif_channel_t *channel;
   int tab;
-  int channels[8][2];
   int altmode[8][2];
-  int (*altdisplay[8])(GtkWidget *, dt_iop_module_t *, int);
-  dt_dev_pixelpipe_display_mask_t display_channel[8][2];
   dt_dev_pixelpipe_display_mask_t save_for_leave;
   int timeout_handle;
   GtkNotebook *channel_tabs;
-  int numberstops[8];
-  const dt_iop_gui_blendif_colorstop_t *colorstops[8];
-  float increments[8];
-
-  const dt_iop_gui_blendif_slider_t *inout;
 
   GtkWidget *masks_combo;
   GtkWidget *masks_shapes[DEVELOP_MASKS_NB_SHAPES];
@@ -454,7 +452,6 @@ typedef struct dt_iop_gui_blend_data_t
   int control_button_pressed;
   dt_pthread_mutex_t lock;
 } dt_iop_gui_blend_data_t;
-
 
 
 /** global init of blendops */
--- a/src/develop/blend.h
+++ b/src/develop/blend.h
@@ -354,6 +354,18 @@ typedef struct dt_iop_gui_blendif_colorstop_t
   GdkRGBA color;
 } dt_iop_gui_blendif_colorstop_t;
 
+typedef struct dt_iop_gui_blendif_slider_t
+{
+  char *label;
+  char *tooltip;
+  float increments;
+  int numberstops;
+  const dt_iop_gui_blendif_colorstop_t *colorstops;
+  int channels[2];
+  dt_dev_pixelpipe_display_mask_t display_channel;
+  void (*scale_print)(float value, char *string, int n);
+  int (*altdisplay)(GtkWidget *, dt_iop_module_t *, int);
+} dt_iop_gui_blendif_slider_t;
 
 typedef struct dt_iop_blend_name_value_t
 {
@@ -366,8 +378,6 @@ extern const dt_develop_name_value_t dt_develop_mask_mode_names[];
 extern const dt_develop_name_value_t dt_develop_combine_masks_names[];
 extern const dt_develop_name_value_t dt_develop_feathering_guide_names[];
 extern const dt_develop_name_value_t dt_develop_invert_mask_names[];
-extern const dt_develop_name_value_t dt_develop_blendif_names_rgb[];
-extern const dt_develop_name_value_t dt_develop_blendif_names_lab[];
 
 #define DEVELOP_MASKS_NB_SHAPES 5
 
@@ -427,6 +437,8 @@ typedef struct dt_iop_gui_blend_data_t
   int numberstops[8];
   const dt_iop_gui_blendif_colorstop_t *colorstops[8];
   float increments[8];
+
+  const dt_iop_gui_blendif_slider_t *inout;
 
   GtkWidget *masks_combo;
   GtkWidget *masks_shapes[DEVELOP_MASKS_NB_SHAPES];
--- a/src/develop/blend_gui.c
+++ b/src/develop/blend_gui.c
@@ -121,44 +121,6 @@ const dt_develop_name_value_t dt_develop_invert_mask_names[]
         { N_("on"), DEVELOP_COMBINE_INV },
         { "", 0 } };
 
-const dt_develop_name_value_t dt_develop_blendif_names_rgb[]
-    = { { N_("GRAY in"), DEVELOP_BLENDIF_GRAY_in },
-        { N_("RED in"), DEVELOP_BLENDIF_RED_in },
-        { N_("GREEN in"), DEVELOP_BLENDIF_GREEN_in },
-        { N_("BLUE in"), DEVELOP_BLENDIF_BLUE_in },
-
-        { N_("GRAY out"), DEVELOP_BLENDIF_GRAY_out },
-        { N_("RED out"), DEVELOP_BLENDIF_RED_out },
-        { N_("GREEN out"), DEVELOP_BLENDIF_GREEN_out },
-        { N_("BLUE out"), DEVELOP_BLENDIF_BLUE_out },
-
-        { N_("H in"), DEVELOP_BLENDIF_H_in },
-        { N_("S in"), DEVELOP_BLENDIF_S_in },
-        { N_("l in"), DEVELOP_BLENDIF_l_in },
-
-        { N_("H out"), DEVELOP_BLENDIF_H_out },
-        { N_("S out"), DEVELOP_BLENDIF_S_out },
-        { N_("l out"), DEVELOP_BLENDIF_l_out },
-
-        { "", 0 } };
-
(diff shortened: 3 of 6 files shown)


johnny-bit:
Well, this is interesting... very interesting...

dterrahe:
Maybe this is more of a solution ("use introspection for something") looking for a problem than a really useful feature. But I do think that more could be done in the history module than just show all modules where _something_ changed. So maybe this :-)

And maybe (right)clicking on a module in history should activate that module in the right panel? Could be messy if it is not in the currently visible list.

And then one step further, combining these two, the module widgets that were changed by this history item could have a red border around them (like you see in web forms for _input required_). But that would get messy quick and should only be implemented if deemed really useful, not a see-what-I-can-do feature.

johnny-bit:
While I haven't yet tested it i think it's actually useful! I honestly wonder "da fuq I adjusted here" a lot, so if it works how I think it works (LOL) then it'd be useful for me :wink:

> And maybe (right)clicking on a module in history should activate that module in the right panel? Could be messy if it is not in the currently visible list.

That might be a bit problematic, right-click is now ingrained in user's mind that it opens up pop-up menu... and there could be a lot of stuff in history-entry pop-up menu, but let's not go there just yet ;)

dterrahe:
Well, I look forward to your feedback _after_ testing ;)

dterrahe:
Just a random screenshot:
Screenshot_20200720_091030

TurboGit:
Yep ! Really nice indeed.

dterrahe:
Interestingly, this gets badly tripped up by liquify, because the introspection structure does not get correctly initialised for all the enums in its multi-nested params.

dterrahe:
made enum values lookup more robust so the introspection bug in liquify (that hopefully @houz will be able to fix) doesn't trip it up anymore.

enum dt_iop_colormapping_flags_t is really a bag of flags, so if its value is two of those flags combined, a simple history change lookup won't find it. Adding those combinations as separate values explicitly, for now, allows the lookup to succeed, If this were more common, the right approach would be to implement a bitwise lookup of values and represent changes as +/-VALUE. Like:
`enum_field: +FIRST_FLAG -THIRD_FLAG`
means bit 1 switch from 0 to 1, bit 3 switch from 1 to 0 and other bits unchanged. Assuming the enum if defined as:
```
FIRST_FLAG = 1 << 0,
SECOND_FLAG = 1 << 1,
THIRD_FLAG = 1 << 2, etc
```

Also set a limit to a maximum of 4 array element changes to display. Adding a ... if there are more than that. To prevent slowdown and tooltip becoming too large. 2-dimensional arrays can still get large, obviously.

dterrahe:
Added changes in blending.

This is more or less complete at this point.

@johnny-bit would love your feedback/suggestions after testing!

johnny-bit:
THIS IS AWESOME

MUST HAVE

SHIP IT!

dterrahe:
image

unsolvable, so it will never be perfect:
- there's no formatting information in introspection
- some of the changed data may not be used. For example change to separate curves, change all three, then change back to combined curve; will continue to show the changes in all the curves.

This now gets regenerated for every history item every time a change is made. It would be more efficient to show in a pop-up when clicked on or something like that. Would like a consensus before starting work on that.

johnny-bit:
I tested this thing lots and I love it! However... In some weird cases, like curves, nodes etc where numbers are nowhere to be found in module changed but are displayed in tooltip it is a bit weirding me out sometimes (despite that I think I'm a dev :wink: ). In cases of modules like liquefy, *curves, colorzones etc the tooltip feels more like a debugging info than history changes tooltip. Similarly with drawn blending it's weird AF ;)

That being said - for almost all other modules this is totally awesome!

I'm thinking a bit about ux for this, especially for special cases like curves. In those cases It could be as simple as "curve changed", "curve changed, increased number of points", "adjusted 1 liqufy node", "added drawn mask - gradient" etc and then build up some pop-up window with details that could be accessed via right-click (prefferably right-click menu). That pop up window at first could then have "debug info" history changes, but having that would allow one to build special-case handling of nodes, curves etc if needed. 

Overal - this change is IMO very very positive!

dterrahe:
@johnny-bit thanks for the great feedback and suggestions!

I completely agree with the main drawback you mention, i.e. handling of curve changes (or anything with flexible number of nodes). The problem really is that I have no semantic information in introspection. So it'll show an array with 200 slots and an integer somewhere that says only 3 of those are used (but maybe previously there were 7, so there's still garbage data for the remaining 4). There is no indication linking that integer to that array. The only place that is aware of this link is the module gui code. To make a more sensible interpretation of the data, there are three options:
- the history module could do a lot of special casing for individual modules (brrr),
- introspection could be extended with some semantic data (but how much? it is not just array lengths, but also flags that cause whole blocks of data to be switched on/off)
or
- the modules could provide dedicated comparison functions for old/new parameters to be called by the history module (doable, but much more invasive and maybe overkill; one can also visually spot changes by switching back and forth in history anyway).

The exception is blending, which doesn't have introspection data anyway, so the code in history does directly interpret the meaning of the data and could be extended to do more of that.

What _could_ be done relatively easily for arrays/curves, is still limit the number of displayed changes to 4, but when > 4 it could show "array_name: n changes", rather than showing the first 4 changes + an ellipsis. This might make the result less confusing and would still give you a clear idea where to look (even though all the changes could be in unused garbage data/deleted nodes). Do you think this would be an improvement?

Also, the earlier introspection changes added many (translatable) descriptions for fields that already have a label in the gui. Curves etc don't have those, so usually the only thing that can be shown is the variable name. If this feature gets merged and is popular, then more $DESCRIPTIONs (and translations) should be added to make the tooltip more readable.

My feeling for this functionality is that, no matter how tempting, it should not be over-engineered/get too large. It is really _intended_ to just be a debugging feature, so it's not such a bad thing that it looks a little under-processed...

johnny-bit:
> It is really intended to just be a debugging feature, so it's not such a bad thing that it looks a little under-processed...

What you _intend_ means little in regards to how it can be used and ho boy, this is useful as one hell of a feature ;)

> What could be done [..] is limit the number of displayed changes [...] Do you think this would be an improvement?

Yes!

> There is no indication linking that integer to that array. The only place that is aware of this link is the module gui code. To make a more sensible interpretation of the data, there are three options:
> * the history module could do a lot of special casing for individual modules (brrr),

HELL NO!

> * introspection could be extended with some semantic data (but how much? it is not just array lengths, but also flags that cause whole blocks of data to be switched on/off)

That's just asking for a trouble in introspection ;)

> * the modules could provide dedicated comparison functions for old/new parameters to be called by the history module (doable, but much more invasive and maybe overkill; [..]).

That's the sanest way IMO. Module could have code to handle both tooltip part and possible popup window part (if it ever gets to that stage), so module in question could interpret and provide sensible changes display.

Ultimately this is very useful feature in terms of debugging as well as in terms of usability. Having modules providing comparison functions would be great as there wouldn't need to be so much special handling and the excercise could be left for module author on how to show differences.

Additionally:

>one can also visually spot changes by switching back and forth in history anyway

I'd LOVE some auto-snapshot functionality to visually inspect differences between history point without need for snapshot module and having it be able to be panned and zoomed and maybe even highlight areas with most changes? But that's way differentthan comparing param changes in history ;)

dterrahe:
> That's the sanest way IMO. Module could have code to handle both tooltip part and possible popup window part (if it ever gets to that stage), so module in question could interpret and provide sensible changes display.
> 
> Ultimately this is very useful feature in terms of debugging as well as in terms of usability. Having modules providing comparison functions would be great as there wouldn't need to be so much special handling and the excercise could be left for module author on how to show differences.

The worry would be that this wouldn't be the highest priority for module writers _and_ that they could have very different approaches to how to show differences so it could become inconsistent across modules and badly maintained/incomplete. Having useful basic default functionality in history with only some hinting provided by complex modules (whatever that would look like) could be best approach. But this would need to be carefully designed.

One way would be to have two _modes_ for a module, the _normal_ value and the _difference_ value (read-only). This could work for sliders as well as curves. Comboboxes could be shown in red if changed and have old value in tooltip. Just spitballing. But mainly saying that creating a dedicated pop-up for complex modules _just_ to show history changes would in my opinion be too much to ask of devs (having got a feel from the introspection project how much they enjoy gui code in general anyway...)

> > one can also visually spot changes by switching back and forth in history anyway
> 
> I'd LOVE some auto-snapshot functionality to visually inspect differences between history point without need for snapshot module and having it be able to be panned and zoomed and maybe even highlight areas with most changes? But that's way differentthan comparing param changes in history ;)

What I mean is, if you've already seen, for example, in the history tooltip for titem 13 that, there was a change in a curve, then you can open the module with the curve and repeatedly click item 13 and item 12 and see the module content change. Not perfect, but maybe better than spending many many hours development/maintenance time and added complexity/bugs to get something slightly better. Unless you can make a very convincing use case, of course ;-)

Maybe some usability improvements could be implemented to make this strategy more efficient (for example keeping masks visible in a module when switching between history items or adding a shortcut to quickly repeatedly toggle between current and previous items or whatever.)

dterrahe:
@johnny-bit a small change for you to have a look at...

As an aside; one reason we want to show changes to individual array element, is that some modules put separate sliders together in an array. It would be useful to split some of those out into separate floats that can have their own $DESCRIPTION. Since the data format doesn't change, this wouldn't need a new module version. One of those modules is your beloved temperature :-)

TurboGit:
Sounds nice but hard to read. I think we could have columns to display the changes:

```
name-control1      from   ->   to
name2              from   ->   to
```

dterrahe:
Still working on a few ideas for this one.  Planning to use  “query-tooltip” and https://developer.gnome.org/gtk3/stable/GtkTooltip.html#gtk-tooltip-set-custom since it looks like that will allow me to show a grid in a tooltip _and_ to only do the analysis work when a tooltip is actually requested.

Also trying to show the changes for parametric blending more accurately, which needs some more refactoring in blend_gui.

TurboGit:
Looks very nice! One point to rework is probably the way drawn masks are displayed in the tooltip. I see:

`mask_id    0  -> 14567888`

This is not informative to users. I would maybe just display which kind of masks are displayed:

`drawn mask   <empty>     circle, brush`

(<empty> above is just to say that the previous state was without mask, this area is the empty string).

What do you think ?

dterrahe:
There are many things that can be improved (I'm thinking of identifying inserted and deleted nodes in curves, for example) and I think the way forward will be to transition to an expandable treeview (in a dialog) that allows drilling down into details like masks and curves.

For now, this is more like debugging; it gives hints but you'll not find all the details. It will help to know where to look though.
So in that sense, I think "mask_id 0 -> 14567888" (or "mask_id x -> y") gives a useful hint. "drawn mask <empty> -> circle, brush" is slightly more useful, but is "drawn mask circle -> circle"? Having this merged will (hopefully) encourage people to add more $DESCRIPTIONs and provide feedback that will help guide the next phase. Personally, I'd rather get that going, at this point, rather than adding more minor polish that will later be discarded again.

If you don't think in its current state it is useful or polished enough, I would understand of course.

johnny-bit:
From my UX course on "complex things in UI": it's better to have "things changed: codespeak mumbe" rather than have nothing. In case of problems, users have "codespeak mumble" to pass to support and have "something" to hold on instead of [blank]

I've tested it and I'm really impressed. The "polish" can come later it's avesome as-is :)

TurboGit:
I don't agree here! We are speaking on UI which is for end-users. If we need to debug the masks then the information should be displayed on the console.

I understand that there is many room for improvement here, but I doubt having mask_id with a magic number will be useful and this certainly makes the dt UI look far too "geeky" to my taste. If having mask information is not yet ready, then I would just drop this part for now as the rest of the PR is really nice.

dterrahe:
If I leave off the mask_id completely, and if that was the _only_ change, then no tooltip will appear.

How about this; 
if mask_id 0->x: "a drawn mask was added"
mask_id x->y: "the drawn mask was changed"
mask_id y->0: "the drawn mask was removed"

johnny-bit:
> How about this;
> if mask_id 0->x: "a drawn mask was added"
> mask_id x->y: "the drawn mask was changed"
> mask_id y->0: "the drawn mask was removed"

THIS is IMO best possible outcome for now :) and is "less geeky" :)

dterrahe:
> I don't agree here! We are speaking on UI which is for end-users. If we need to debug the masks then the information should be displayed on the console.

BTW what I meant here is a user debugging their own actions, because they don't remember what they did.

TurboGit:
> If I leave off the mask_id completely, and if that was the _only_ change, then no tooltip will appear.
> 
> How about this;
> if mask_id 0->x: "a drawn mask was added"
> mask_id x->y: "the drawn mask was changed"
> mask_id y->0: "the drawn mask was removed"

Sounds good to me too!

dterrahe:
I will also translate the blendif bitbucket into polarity indicators. The average user might otherwise struggle to interpret the hex values... ;-)

dterrahe:
Turns out changes to an existing drawn mask do not give a new mask_id, so the message "the drawn mask was changed" doesn't normally show up. But at least no useless cryptic numbers anymore.

Plus polarity indicators.

Let me know if there's anything more I can do.

johnny-bit:
Just tested - this is AMAZING! Generally any polish can now come from modules (eg better names/enums?)

dterrahe:
I'm still experimenting a bit with monospaced fonts to make things more readable.

@TurboGit please give me maybe another day to see if I can make it slightly prettier.

Afterwards, indeed modules can improve things.

TurboGit:
@dterrahe : Sure, I'll check anyway right now because I'm curious :)

Nilvus:
> If I leave off the mask_id completely, and if that was the _only_ change, then no tooltip will appear.
> 
> How about this;
> if mask_id 0->x: "a drawn mask was added"
> mask_id x->y: "the drawn mask was changed"
> mask_id y->0: "the drawn mask was removed"

I discover this PR and not tested yet (will do that) and seems really awesome. I agree with @TurboGit, tooltips needs to be clear for an end-user (and not a geeky one!) so much better like this. For more detailed infos, maybe another PR could be added later to improve and be more precise and keep things user-friendly.

dterrahe:
switched to monospaced and calculating tab stops to make it line up better. Hopefully this is more pleasing to everybody (even if monospaced is somewhat ugly)

image

The only problem I have now that I need to find the pango size of the font used. Currently it works fine with the system font, but font size > 10 messes everything up.

dterrahe:
Done. Please test.

Nilvus:
Just made a quick test and some feedback:
- there's an annoying glitch if you move a little the cursor on same line. See blinking at the right of the tooltip. Width of the tooltip seems to reduce a little before taking back on good size.
- there's a not wanted text on input color profile module (need a more user-friendly one):

- I don't like the way tooltip is displayed (too big border) but I could work on it later to improve that.

data/themes/darktable.css
```diff
@@ -174,6 +174,11 @@
   min-height: 0;
 }
 
+.monospace
+{
+  font-family: monospace;
```
Nilvus, line 179: Please add that new CSS setting on block starting line 1418 please (/* Set some specific fonts */ part).
I would also suggest to add a more specific tag (like hist-tooltip for example).

Nilvus (changes requested):

johnny-bit:
> there's a not wanted text on input color profile module (need a more user-friendly one):

That's just side-effect of lack of descriptions etc (IMO). No need to work on that in this PR, but can be used as a nudge to get modules to get better history tootlip (input color profile in this case) :) Not a blocker IMO ;)

> I don't like the way tooltip is displayed (too big border) but I could work on it later to improve that.
:heart: 

> there's an annoying glitch if you move a little the cursor on same line

I wonder if that's a GTK glitch, I haven't experienced that in my testing...

dterrahe:
Learned more about tooltip widgets than I ever wanted to know. Ask me anything!

dterrahe:
@Nilvus I assume thumbs-up mean the tooltip changes work for you and you no longer see the blinking either?

@TurboGit once this is merged, I'm hoping others will be sufficiently encouraged, by seeing raw enum names etc in the history changes of their favorite modules, to add @DESCRIPTION tags to the structs (params) and enums that still lack them. That should address @Nilvus's other critique. @Nilvus himself can then also change the styling to his liking. For me personally, a somewhat wider border is helpful to separate the tooltip from the module names in the history list itself, but I appreciate there are aesthetic and consistency considerations.

TurboGit:
At this stage I see no reason to delay this more. The output looks good, indeed there is enums and non translated strings here and there but this is a job to be done in each iop. Thanks a lot for this nice enhancement.

Can you provide an entry for the release-notes and some paragraph for the documentation ?

Nilvus:
@dterrahe: to be honest, I haven't tested it. Thumbs up was just for your quick fix and I was sure fix was good. And I test it on master, it's all good!

About aesthetic of tooltip, I will see ASAP if I can propose something else. But it's a good start.

Nilvus:
Thanks for your work.

chhil:
Take channel mixer as an example in the history.

This one is where the change was made.
image

This one is selecting the channel mixer later to see what the history means. (only viewing it). See the green missing.

image

For whatever reason the tooltip has changed.

Whats possibly missing specifically in this case is the destination channel that was used, in my case it was grey.
Would the changing destination to see the grey channel trigger a diff with different values?
image


Thank you for adding this feature, its very helpful.

dterrahe:
The tooltip only shows changes compared to the previous version. So #24 says it was created because there are differences with #20. Given that those differences don't show up in the values it prints, I can only assume that they are rounding errors. Maybe caused by the values being recalculated when changing the channel.

This is internal to the channel mixer module, so if you think this is undesirable behavior (a new history item getting created) you'd have to file a bug report against it.

Introspection doesn't know the _meaning_ of the index into the array of values the module stores. So they can't be interpreted and labelled _lightness_ or _saturation_ or whatever. This is a bit similar to my NB1 remark here https://github.com/darktable-org/darktable/pull/6075#issuecomment-680942466 that introspection can't know which portion of an array is used. Introspection is meant to be an internally focused tool and it is unlikely that features will be added just to help presentation.

Nilvus:
> Learned more about tooltip widgets than I ever wanted to know. Ask me anything!

@dterrahe: need helps here to work on CSS. I can't find which CSS tag to change border color/size, padding inside, etc. I've tried #show-tooltip, #hist-tooltip and #query-tooltip (and .query-tooltip). Nothing works. And as it's impossible to move the cursor hover a tooltip, I can't use Gtk Inspector Tool for that. Could you help me find (or add) tag to test/update that?

dterrahe:
So... the tooltip itself that I get handed in the "query-tooltip" callback, isn't a widget, so I can't give it a name that you can refer to in css. 

I can put a widget _inside_ whatever the tooltip is. And I can give that widget a name, #hist-tooltip , so you can set padding on that (border doesn't seem to do anything). But since default padding is 0, you can only increase that way...

There is a "tooltip" label, but that would then refer to _all_ tooltips, so I'm guessing changing the border/setting on that is undesirable. And anyway, it doesn't seem to have any effect.

Between the tooltip and my widget, there is, if I understand correctly, a box. If I mess with padding/border of that one, I get the strange effect that my widget gets cut off slightly (on the right side).

```
tooltip *
{
  padding: 0px;
  border: 0px;
}
```

image

What I think looks fine/good under all themes, because I don't mind the border/think it is useful, is just:

```
#hist-tooltip
{
  background-color: @tooltip_bg_color;
}

```

Nilvus:
> What I think looks fine/good under all themes, because I don't mind the border/think it is useful, is just:
> 
> ```
> #hist-tooltip
> {
>   background-color: @tooltip_bg_color;
> }
> ```

Thanks for you complete and precise answer. I totally agree on that proposal. It's looks great on all themes. No need to do something else.

As it's your proposal (and a simple and really good one), I let you post a PR to propose that. Thanks.

PR #5771, opened 2020-07-19
po/fix user aspect

TurboGit:


elstoc:
This is a very finicky module. It pretty much always seems to be one bug out one bug in! It no longer hangs as far as I can tell, but there are still some odd things going on. 

If I key a ratio of 1.29, commit it (double-click) it looks ok:

Screenshot_2020-07-19_19-18-52

I then click the 'change aspect' button and it changes as follows:

Screenshot_2020-07-19_19-19-17

Change aspect and commit one more time and all the ratios go negative:

Screenshot_2020-07-19_19-19-36

From here on in, change aspect has no effect.

elstoc:
I was asked to test this - seems to have been merged before I finished

aurelienpierre:
Argh sorry, I was requested for review, so I thought that was finished work, and only saw the issue after.

elstoc:
@MStraeten has also raised another similar issue against the linked bug

MStraeten:
not a new issue, but a comment to the original. the ugly part (darktable hangs) is removed, but the negative values and ability to rotate after leaving the module and enter it again is remaining.

PR #5775, opened 2020-07-20
darkroom: better visual indication of drag-drop operations

### LINKED ISSUE #5772: FR: Better visual indication and control when changing module order
Usually a user will only have a small subset of available modules displayed in the right hand panel.

If a user chooses to reorder the modules by dragging and dropping, it is not clear how that module will be placed in relation to the modules that are **not** displayed, either during or after the drag-and-drop operation.

For example, take the following random subset of modules (where all modules are shown in the right hand panel)

Screenshot_2020-07-19_20-11-14

I amend my favourites group to show alternate modules from this list

Screenshot_2020-07-19_20-12-52

Now, say I want to move fill light so that it sits between color zones and velvia, there are two choices - I can either place it immediately before color zones (after vibrance) or immediately after velvia (before vibrance) in the pipe. Currently darktable makes that choice for me. However, it should be possible for the user to choose between these options and for darktable to visually indicate which one will be done when the module is dropped.

The best suggestion I can think of is that, while dragging, the user is presented with two empty 'placeholder' slots into which they may drop the module. The placeholders appear when the user drags over a module header (they appear before that header). They move if the user drags over another module header and they disappear if the user drops the module or drags over the body of another module or outside of the right hand panel. The module can only be dropped into one of these placeholders.

elstoc:
when dragging and dropping modules in the darkroom, show the gap into which the dragged module will be inserted and draw a line around the module after|before which the insert will be made.

For example, the following shows that the exposure module will be placed immediately after the color mapping module in the pipe

Screenshot_2020-07-20_21-13-20

Resolves #5772.

### DIFF
--- a/data/themes/darktable.css
+++ b/data/themes/darktable.css
@@ -1779,6 +1779,17 @@ scale contents trough slider:hover
   background-color: @scroll_bar_active;
 }
 
+.iop_drop_before
+{
+   border: 2px solid @plugin_bg_color;
+   border-bottom: 2.8em solid @plugin_bg_color;
+}
+
+.iop_drop_after
+{
+   border: 2px solid @plugin_bg_color;
+   border-top: 2.8em solid @plugin_bg_color;
+}
 
 /*** --- Thumbtable css part ---
 
--- a/src/views/darkroom.c
+++ b/src/views/darkroom.c
@@ -2500,7 +2500,7 @@ static void _on_drag_begin(GtkWidget *widget, GdkDragContext *context, gpointer 
 
       GdkPixbuf *pixbuf = gdk_pixbuf_get_from_window(window, allocation_w.x, allocation_w.y,
                                                      allocation_w.width, allocation_w.height);
-      gtk_drag_set_icon_pixbuf(context, pixbuf, allocation_w.width / 2, 0);
+      gtk_drag_set_icon_pixbuf(context, pixbuf, allocation_w.width / 2, allocation_w.height / 2);
     }
   }
 }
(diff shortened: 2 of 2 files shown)


elstoc:
I've made a couple of minor tweaks and apart from agreeing the CSS look/feel I think this is done.

data/themes/darktable.css
```diff
@@ -1779,6 +1779,17 @@ scale contents trough slider:hover
...
+{
+   border: 2px solid black;
+   border-bottom: 3.5em solid black;
+}
+
+.iop_drop_after
+{
+   border: 2px solid black;
+   border-top: 3.5em solid black;
+}
```
Nilvus, line 1792: 3.5em is really too big and make that not fine-tuned. 2.6em is for me the better much better choice  as it's just little more bigger than module height. 2.8em is the higher acceptable value here for me. On both lines of course.

For all lines, color black is not suitable on all themes. I just suggest to replace them by @plugin_bg_color that is exactly done for such a new thing.

Nilvus, line 1792: 3.5em is really too big and make that not fine-tuned. 2.6em is for me the better much better choice  as it's just little more bigger than module height. 2.8em is the higher acceptable value here for me. On both lines of course.

For all lines, color black is not suitable on all themes. I just suggest to replace them by @plugin_bg_color that is exactly done for such a new thing.

Nilvus:
Made some mistakes and can't find how to change a review from to request review...

data/themes/darktable.css
elstoc: > 2.6em is for me the better much better choice...2.8em is the higher acceptable value here for me.

Fine with me. I prefer 2.8 personally (or perhaps even a little bigger - see screenshot below).

> I just suggest to replace them by @plugin_bg_color

The borders serve two purposes here. Firstly to make a space to indicate where the dragged module will be dropped but secondly to highlight the destination module to which the dragged module is being "pinned" (i.e. to show which module it will be placed immediately before or after). Because either module may or may not be expanded, the border was the only way I could think of to indicate this. 

I could perhaps make the remainder of the border a bit larger. So, for example the following indicates that the exposure module will be dropped immediately before channel mixer (here the bottom border is 2.8em and the remaining borders are 4px):

Screenshot_2020-07-20_20-54-58

Open to suggestions of other ways to indicate this

Nilvus: I understand all purposes but that's not good the way you propose and I don't understand the problem. With 2.8em and plugin_bg_color, on all themes, I see without any problem where I drag and drop a module (start point and end point). The way you propose with black and 3.5em make that just really too big and not adapted to the theme color, especially in grey and dark ones of course (for black color) and so plugin_bg_color just adapt that. On darker theme and so default one, plugin_bg_color is just a color near to black.

On screenshot you post, I don't see any problems, we see module moved and yes, the idea for me is to not have too much space around (quite awful and not consistent with the fine-tuned UI we made for 3.2). We see that we move module without needed of so much space and with 2.8, you still see a border around (a thin one and it's just sufficient for me). I find even 3em too big.

elstoc: Here it is with 2.8em and 2px (exposure will be placed after color mapping in the pipe)

Screenshot_2020-07-20_21-13-20

Ok I can live with 2px. I agree 4px is too much

Nilvus: Thanks. All good for me and works great on my side.

elstoc: Excellent. Thanks @Nilvus

Nilvus: Oh, even if it's not the subject here, I name you in the future 3.2 article for darktable.org/blog (and so translation versions), so if you want to check/read: https://github.com/darktable-org/dtorg/pull/111

PR #5777, opened 2020-07-20
clipping: fix clipping using float values.

### LINKED ISSUE #5774: Crop & Rotate: Float / Ratio differences
If I key 129:100 into the Crop & Rotate module, the aspect is displayed as follows:

Screenshot_2020-07-19_21-12-24

However, if I key 1.29 (the same thing) this results in a much different ratio:

Screenshot_2020-07-19_21-13-14

I would expect both of these actions to result in the aspect ratio displayed in the first screenshot

TurboGit:
The float value is now immune to the binary base representation. We
do parse the string and at the same time handle both , and . as
separator.

Fixes #5774.

### DIFF
--- a/src/iop/clipping.c
+++ b/src/iop/clipping.c
@@ -1617,6 +1617,37 @@ void reload_defaults(dt_iop_module_t *self)
   memcpy(self->params, self->default_params, sizeof(dt_iop_clipping_params_t));
 }
 
+static void _float_to_fract(const char *num, int *n, int *d)
+{
+  char tnum[100];
+  gboolean sep_found = FALSE;
+  char *p = (char *)num;
+  int k = 0;
+
+  *d = 1;
+
+  while(*p)
+  {
+    if(sep_found) *d *= 10;
+
+    // look for decimal sep
+    if((*p == ',') || (*p == '.'))
+    {
+      sep_found = TRUE;
+    }
+    else
+    {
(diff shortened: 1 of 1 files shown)


elstoc:
Thanks @TurboGit. Works for me.

elstoc:
Minor thing - for some reason it won't let me key "1" but it will let me key "1.0"

MStraeten:
works, for 3.4 maybe a input format check can make sense: 1,2,3 or 1.5,6:5.6 are accepted and processed without errors - so quite stable and foolproof ;)

elstoc:
Agreed, and perhaps throw an error if you try to enter non-numeric characters! Not important for this fix - it's the user's problem if they decide to key 1.a (which by the way gets evaluated to 10:1)

TurboGit:
> Minor thing - for some reason it won't let me key "1" but it will let me key "1.0"

I had a look at this and for 1 the value-changed callback `aspect_presets_changed` is not called ! Looks like 1 is always the current value (whatever it means) for this combo and so 1 change nothing !!!???? Anyway, I'll merge this PR for the release as the issue with 1 is different.

TurboGit:
@elstoc : feel free to create an issue for the 1 being ignored, this is probably a Gtk or bauhaus widget issue.

elstoc:
Done. #5784

PR #5781, opened 2020-07-21
imageio: Add support for libavif >= 0.8.0

cryptomilk:
Add support for libavif >= 0.8.0

### DIFF
--- a/src/imageio/format/avif.c
+++ b/src/imageio/format/avif.c
@@ -297,6 +297,62 @@ int write_image(struct dt_imageio_module_data_t *data,
   avifImageRGBToYUV(image, &rgb);
 
   if (imgid > 0) {
+    bool use_icc = false;
+
+#if AVIF_VERSION >= 800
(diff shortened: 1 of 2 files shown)


cryptomilk:
With libavif 0.8.0 they changed the API so this just adopts the code to compile with the new API nothing else. It might be worth considering it as a bugfix for darktable 3.2 or you would only be able to compile with libavif 0.7.x and not any newer versions.

src/imageio/format/avif.c
```diff
@@ -402,14 +458,16 @@ int write_image(struct dt_imageio_module_data_t *data,
         break;
     }
 
+    if (nclx.colourPrimaries != AVIF_NCLX_COLOUR_PRIMARIES_UNKNOWN) {
```
TurboGit, line 461: Would be nice to fix the style on this file. The { and } in dt are always on their own line. I haven't noticed that before.
So we should have:

```
if()
{
}
else
{
}
```

Likewise for switch

```
switch()
{
}
```
I will merge this for 3.2 as it seems safe. The reformatting can come later.

PR #5785, opened 2020-07-21
Demosaic: Remove unused parameter on Amaze and delete some dead code.

pitbuster:


parafin:
I think the idea behind keeping that dead code is to keep it as close to the original as possible to ease potential update. The question is whether it's still something we care about. Given the need to keep backwards compatibility, I don't think it's going to happen, so I guess this PR can be merged.

PR #5798, opened 2020-07-24
create a GUI button for focus-peaking

### LINKED ISSUE #5748: Add toggle button for focus peaking
Currently the focus peaking mode is accessible with ctrl+shift+f by default. I have read some stories from people who have accidentally enabled it and don't know what has happened, and I think the lack of a visual cue that a certain mode has been enabled would help a lot.

The point of this issue is not to complain and it certainly shouldn't be high priority, it is not a problem for me personally, but to discuss whether this would improve darktable's usability.

My suggestion for a position is in the top bar, next to the overlay/group/settings icons, because those controls affect images in both the filmstrip and the lighttable. The bottom bar controls only affect the selected image (afaik) so putting a button there would be confusing I think.

aurelienpierre:
* fix #5748
* read and write focus-peaking state in `darktablerc` only when app starts and closes, not during runtime
* use thread mutex locks to read/write `darktable.gui` states consistently

Darkroom:
Screenshot_20200724_160209
Lighttable:
Screenshot_20200724_160017

### DIFF
--- a/src/dtgtk/paint.h
+++ b/src/dtgtk/paint.h
@@ -179,6 +179,8 @@ void dtgtk_cairo_paint_zoom(cairo_t *cr, gint x, gint y, gint w, gint h, gint fl
 void dtgtk_cairo_paint_multiinstance(cairo_t *cr, gint x, gint y, gint w, gint h, gint flags, void *data);
 /** paint a simple grid icon */
 void dtgtk_cairo_paint_grid(cairo_t *cr, gint x, gint y, gint w, gint h, gint flags, void *data);
+/** paint focus peaking icon **/
+void dtgtk_cairo_paint_focus_peaking(cairo_t *cr, gint x, gint y, gint w, gint h, gint flags, void *data);
 
 /** paint active modulegroup icon */
 void dtgtk_cairo_paint_modulegroup_active(cairo_t *cr, gint x, gint y, gint w, gint h, gint flags, void *data);
--- a/src/dtgtk/paint.c
+++ b/src/dtgtk/paint.c
@@ -928,6 +928,58 @@ void dtgtk_cairo_paint_grid(cairo_t *cr, gint x, gint y, gint w, gint h, gint fl
   FINISH
 }
 
+void dtgtk_cairo_paint_focus_peaking(cairo_t *cr, gint x, gint y, gint w, gint h, gint flags, void *data)
+{
+  PREAMBLE(1, 0, 0)
+
+  // stigmometer aka split focusing screen aka Dodin's prism
+  cairo_arc(cr, 0.5, 0.5, 0.2, 0, 2. * M_PI);
+  cairo_move_to(cr, 0.3, 0.5);
+  cairo_line_to(cr, 0.7, 0.5);
+  cairo_stroke(cr);
+
+  // corners
+  const double center = 0.5;
+  const double width = 1.;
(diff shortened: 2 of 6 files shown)


src/dtgtk/paint.h
```diff
@@ -179,6 +179,8 @@ void dtgtk_cairo_paint_zoom(cairo_t *cr, gint x, gint y, gint w, gint h, gint fl
 void dtgtk_cairo_paint_multiinstance(cairo_t *cr, gint x, gint y, gint w, gint h, gint flags, void *data);
 /** paint a simple grid icon */
 void dtgtk_cairo_paint_grid(cairo_t *cr, gint x, gint y, gint w, gint h, gint flags, void *data);
+/** paint focus peaking icon **/
```
johnny-bit, line 182: in order for `show_icons_in_paint_c.pl` tool to work, the comment needs to be in form of:
```c
/** TEXT TEXT TEXT */
```
or should I fix the pattern matching in `show_icons_in_paint_c.pl` :wink:

TurboGit:
I find the icon a bit small. Not well balanced with others. @Nilvus , what's your view on this?

Egocentrix:
The icon is something I drew as an example in the issue thread (#5748), I don't think it should actually be used here because it's a) not actually symmetrical and b) not square.
I'll have time to try to make a proper icon on Monday, but probably someone with more knowledge of the darktable icon style can do a better job.

Nilvus:
Not sure on this one. I find it well-balanced for what is and if considering that the icon represent a camera with a focus point. But, indeed, it's not balanced at all regardless of other icons. But for that, the only way would be to have a square icon with a bigger circle inside. Not sure it will look good.

Anyway, the icon could probably be higher at least to be better. I agree that's too small actually.

Egocentrix:
mockup
I made a quick drawing of what a square icon would look like, just comparison.

I also tried it with a larger circle, but then it starts to look more like a button than a viewfinder to me.

aurelienpierre:
Using a square icon will refer to medium format cameras (6×6 cm) which don't even have a stigmometer. 

I agree the 3:2 icon is unbalanced, but it doesn't look like what it represent anymore if you make it square. Typical case of "making it pretty make it loose its meaning". 

I don't know. Real focusing screens look like this : 
image

Nilvus:
That indeed confirm square icon is worse. Maybe make it 4:3 size, that could be a better compromise.

TurboGit:
So what should we do about this ? Should we go as-is ? Test a 4:3 icon ? I would be in favor of the later...

aurelienpierre:
To change the ratio, in `paint.c`, replace the 2./3. fraction here by 3./4. : `const double height = width * 2. / 3.;` (line 944)

I'm currently busy on filmic v4 OpenCL so this is low priority for me. Feel free to merge and change the ratio later.

TurboGit:
> I'm currently busy on filmic v4 OpenCL so this is low priority for me. Feel free to merge and change the ratio later.

I'm ok with that. Thanks Aurélien, will do.

TurboGit:
Manually merged and button handling fixed avoiding lot of:
`(darktable:383750): Gtk-CRITICAL **: 14:38:09.227: gtk_box_pack: assertion '_gtk_widget_get_parent (child) == NULL' failed
`

aurelienpierre:
Thanks @TurboGit !

PR #5800, opened 2020-07-24
Cannot select an area with the parametric mask color picker

hlecleme:
It seems that the selection of an area (through ctrl+clicking) by the first color picker of the parametric mask blending section is not possible anymore.

This PR suggests a fix for this.

### DIFF
--- a/src/gui/color_picker_proxy.c
+++ b/src/gui/color_picker_proxy.c
@@ -157,7 +157,8 @@ static gboolean _iop_color_picker_callback_button_press(GtkWidget *button, GdkEv
   // set module active if not yet the case
   if(module->off) gtk_toggle_button_set_active(GTK_TOGGLE_BUTTON(module->off), TRUE);
 
-  const GdkModifierType state = e != NULL ? e->state : dt_key_modifier_state();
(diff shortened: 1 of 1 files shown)


dterrahe:
This looks good thank!

@TurboGit can this still go into 3.2? It's a clear trivial bug fix for a regression.

And sorry for messing this up in #5591; no bug reports for almost a month, but I doubt you were the first one to notice, so apologies to all the others as well :-)

dterrahe:
This bug also impacts color zones, rgb curve and tone curve under linux. It worked fine on windows; this cross platform developing/debugging/testing is fun!

MStraeten:
this bug isn't present on osx.

dterrahe:
The bug is triggered on linux by having num or capslock switched on. Capslock _does_ actually trigger it on windows as well :-/

darktable developers have no need for capslock, so...

hlecleme:
Thanks @dterrahe for better explaining it. I confirm that it was observed on Linux with num lock switched on ;-)

TurboGit:
I'll have a look at this early next week.

Issue #5826, opened 2020-07-27
metadata: title not updated


TurboGit:
**Describe the bug**
The title (and possibly other metadata) are not properly updated from XMP.

**To Reproduce**
1. On lighttable select one image
2. Add title "mytitle"
3. Make sure the option to check for updated XMP on disk is activated
3. Close darktable'
4. With one editor change "mytitle" to "mytitleNEW" for example
5. Start darktable
6. When asked to update XMP do so, select the picture and update dt library from XMP (left button)
7. See error : the title has been removed from image

**Expected behavior**
The title should have been updated

**Platform (please complete the following information):**
 - Darktable Version: master
 - OS: GNU/Debian
 
**Additional context**
Maybe related to #5722.

TurboGit:
@phweyland : Can you have a look ? I have made the reproducer as simple as possible.

TurboGit:
Just tested, the same issue arise for all metadata, not only title.

TurboGit:
We have a race condition somewhere and this may be the issue. Another test I did on 225 pictures.
- Select all pictures
- Assign tag `tagtoreplace`
- Quit dt
- On the command line mass change the tag:
`sed -i -s 's/tagtoreplace/tagtoreplaceNEW/g' *.xmp`
- Start dt, when asked select all pictures and update db from XMP

Result is that some images have the `tagtoreplaceNEW` and some others `tagtoreplace`.

phweyland:
> @phweyland : Can you have a look ? I have made the reproducer as simple as possible.

For sure

phweyland:
Hi Pascal, the two scenarios above work for me with the current master.
I've made the test with a fresh db, imported some images and run your scenarios.
No change in the settings except "look for updated xmp files on startup".
...

phweyland:
I've repeated the second scenario with 367 images both for metadata and tags (metatoreplace and tagtoreplace) and everything is still fine.

TurboGit:
Good news... But worrisome as I can consistently reproduce the first scenario on my side.

After having removed all XMP and imported my collection I cannot reproduce the second scenario. So let's forget about this one for now, but again it may be triggered by some context issue...

TurboGit:
@phweyland : of course you need the option to always write an XMP sidecar to be activated. Is that the case on your side?

TurboGit:
@phweyland : I'm really worried now... I cannot find the code to read the metadata (title, published...). And this simple scenario show the issue:

[...]

PR #5827, opened 2020-07-27
init metadata flags at start time

phweyland:
Related to #5826
I haven't reproduced any of the issues, but it makes sense not to rely on UI to initialize the metadata flags EDIT and to do it before calling crawler.
Does that solve the main concern ?


TurboGit:
Thanks, working on my side. I double checked that before this commit I have the flags set to 0 when removed and now they are properly set to 4. I cannot tell why I have this behavior on my side, but I must say that I'm now confident that this PR fixes a possible disastrous issue! Thanks a lot Philippe.

PR #5836, opened 2020-07-29
Use @RvRijsselts workaround to fix broken locallaplacian on AMD

matze:
Don't want to take credit but AFAICT a Darktable 3.2 release is imminent …


TurboGit:
Note that I **trust you** for having tested this. I don't have an AMD card. I have tested this patch on my side against all integration tests. No issue, as expected it won't impact other GPU. So I'll merge for the coming release.

matze:
> Note that I trust you for having tested this.

Thanks! On AMD the change looks like this (top with the mis-compiled `switch`, bottom using `select`):

locallaplacian-issue

I will keep an eye on any weird behavior but the way it is now makes the local contrast module usable again.

Issue #5837, opened 2020-07-30
WB Presets for Sony ILCE-7RM4


matt-maguire:
[darktable-whitebalance-20200730.tar.gz](https://github.com/darktable-org/darktable/files/4999310/darktable-whitebalance-20200730.tar.gz)

johnny-bit:
Looks great! Associated PR didn't autoclose this issue, would you mind closing it manually, please?

matt-maguire:
ILCE-7RM4 WB preset changes accepted and merged.

PR #5838, opened 2020-07-30
[Issue #5837] Sony ILCE-7RM4 WB Presets FW 1.10

matt-maguire:
Extracted WB preset info out of Sony ILCE-7RM4 FW Ver 1.10, and inserted it into wb_presets.c. The associated .tar.gz file is attached to Issue #5837 , if you want to double-check anything.


TurboGit:
Thanks !

PR #5846, opened 2020-07-31
demosaicing option: changing outdated tooltip

EdgarLux:
After preferences window redesign, 'pixel interpolation' option is no longer below 'demosaicing for zoomed out darkroom mode' option. Tooltip changed.


TurboGit:
This will have to wait for next release as it breaks the translation (we are on string freeze).

EdgarLux:
> This will have to wait for next release as it breaks the translation (we are on string freeze).

Okay, I understand. Thank you!

TurboGit:
Thanks !

PR #5848, opened 2020-07-31
Updates to pt_BR

jpellegrini:
A few more...


TurboGit:
Sources are now cut, this will be for 3.4 some hours too late sorry.

jpellegrini:
> Sources are now cut, this will be for 3.4 some hours too late sorry.

Even for changes to the manual only? If that's so, no problem - I must have misunderstood it.

TurboGit:
Oh that's user's manual, no problem then. Not clear from subject line :)

PR #5849, opened 2020-08-01
Add Nikon Z 50 white balance presets (fine)

vernonrj:
fine white balance presets for all 12 presets supported by the Nikon Z 50.

Following the redmine wiki page [for how to generate White Balance Presets](https://redmine.darktable.org/projects/darktable/wiki/White_balance_presets), I ran `extract_wb.py *.NEF` on the set of pictures I had taken at different white balance settings. I then copied the resulting struct definitions directly into `wb_presets.c`.

(note: I also ran `extract_wb_from_images.sh *.NEF` later to generate a tarball with the white balance details from each image. I have included that here for completeness: [darktable-whitebalance-20200730.tar.gz](https://github.com/darktable-org/darktable/files/5009503/darktable-whitebalance-20200730.tar.gz))


johnny-bit:
This looks great!

PR #5859, opened 2020-08-02
Updating pt_BR translation

jpellegrini:
Not the user manual, this is for the darktable strings (yup, I know it won't be in for the next release - not sure why this I didn't see one)


TurboGit:
Thanks !

PR #5860, opened 2020-08-02
lighttable: fix arrow navigation in file manager view

### LINKED ISSUE #5858: lighttable: column position lost when navigating with arrow keys
In the lighttable file manager view, click any image that is **not** in column 1 to give it focus. Say, an image in column 2.

Press the up arrow key until you reach the top row. Each time you move up, the highlighted image changes and the view scrolls as needed (all as expected). In this example, when the top row is reached, the image in row 1 column 2 is highlighted.

Press the up arrow one last time and _no image_ is highlighted.

After this point, pressing the down key causes focus to be given to row 2 column 1. If instead you press the right arrow key, focus is given to row 1 column 2.

It looks like on the final press of the up arrow key the fist image becomes the one selected (row 1 column 1) but there is no visual indication of this.

I would expect that on the final press of the up arrow the screen should remain unchanged (row 1 column 2 should remain highlighted). This is what happens when the bottom row of the lighttable is reached.

Issue found on current master and 3.2.x branch.

elstoc:
Ensure that when navigating up with the arrow key in lighttable file manager view, the cursor stops at the top (or when
navigating up would cause the cursor to move to an empty slot in the grid).

Resolves #5858

### DIFF
--- a/src/dtgtk/thumbtable.c
+++ b/src/dtgtk/thumbtable.c
@@ -2320,7 +2320,7 @@ static gboolean _filemanager_key_move(dt_thumbtable_t *table, dt_thumbtable_move
     newrowid = baserowid - 1;
   else if(move == DT_THUMBTABLE_MOVE_RIGHT && baserowid < maxrowid)
     newrowid = baserowid + 1;
-  else if(move == DT_THUMBTABLE_MOVE_UP && baserowid >= 2)
+  else if(move == DT_THUMBTABLE_MOVE_UP && baserowid - table->thumbs_per_row >= 1)
     newrowid = baserowid - table->thumbs_per_row;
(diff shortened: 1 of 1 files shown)


elstoc:
Probably is only a complete fix when combined with #5861, since there may still be issues with up/down keys when lighttable reorders to leave some of the first slots empty.

elstoc:
@TurboGit please can we have a final communication to the dev mailing list on the final plan for 3.2. There are some contradictory messages going about on IRC.

TurboGit:
The plan is still the same : release panned on Aug 10th. I have just updated the source tarball.

I'm not on IRC and don't know who spread information about the release but no one contacted me to talk about the release so the plan is still has described in my latest message on the dev mailing list. Hope this clarifies the situation.

TurboGit:
@elstoc : message sent to dev mailing list.

elstoc:
Thanks

Issue #5722, opened 2020-07-08
Updating a lot of XML sidecars freezes Darktable


elman22:
**Describe the bug**
I use 2 computers to edit my photos. I use syncthing to synchronize photo folders. Each Darktable installation updates a lot of xmp files (even thou I don't edit all those files, but that is different issue I guess). So when I sync xmp to the other Darktable installation and start it, I'm prompted about updated xml (sometimes about 10000 of them). Then I select all xmp and click "update database from selected xmp files". Darktable then freezes. Not sure if after everything is processed or before, but some xmp files on disk have current timestamp. Also not sure why, since I want to update database, not xmp files. Maybe some racing condition. Or maybe it is actually something doing for so long, and progress indicator would be nice. When I kill Darktable and start it again, I'm again prompted with updated xmp files. When I repeat all steps, database is updated and I'm able to close the window.

**To Reproduce**
1. Import about 10000 images to first Darktable installation
2. Move all files including xmp to second Darktable installation (same version of Darktable)
3. Edit some files in second installation, maybe set time offset to all of them
4. Move all files to first installation and overwrite files
5. Start first Darktable
6. Select all xmp files and click update database from selected xmp files"

**Expected behavior**
Window content should be updated (with empty contents) and I should be able to close it.

**Platform (please complete the following information):**
 - Darktable Version: 3.0.2
 - OS: Manjaro Linux
 - OpenCL activated
 - nVidia proprietary 440.xx

Mark-64:
Indeed, as wrote in #5697, in current git master this use case seems broken as DT sees all .xmp as modified.
This didn't happen in DT 3.0.2

TurboGit:
Should be good to be fixed in 3.2 indeed. Now the reproducer is not ok to me. Please can you find a simple way to reproduce without the 10000 pictures. You can't expect any dev to handle such reproducer. Can you reproduce with one or two XMP ? If so can you describe the steps to reproduce ? Thanks.

elman22:
OK, I will try to find simpler way to reproduce.

TurboGit:
Very much appreciated, thanks!

ralfbrown:
I suspect the problem is that all of the work is being done in a button callback, specifically _reload_button_clicked in src/control/crawler.c, which calls dt_history_load_and_apply for each and every selected XMP before returning control to GTK.  Similarly, _overwrite_button_clicked just below it calls dt_image_write_sidecar_file for every selected XMP before returning control to GTK.
[...]

PR #5869, opened 2020-08-04
crawler fix - update write timestamp in db if required but do not write back the xmp file

### LINKED ISSUE #5847: All XMP are rewriting when "update database from selected xmp files"
I am editing the same shots on Windows or Linux computers. When I migrate my collection from Linux to Windows (or vice vers) all XMPs are overwritten (I use the "check xmp on start" option and click "update database from selected xmp files" in the crawler dialog).
But this only changes the values "import_timestamp" and "change_timestamp". And this values for each individual snapshot are always the same.
For example:
on Linux
```
   darktable:import_timestamp="1594311247"
   darktable:change_timestamp="1594311413"
```
on Windows
```
   darktable:import_timestamp="1594233276"
   darktable:change_timestamp="1594233508"
```

Darktable Version: master
OS: Debian 10 and Windows 7

phweyland:
attempt to fix #5847 

If xmp timestamp is newer than db write timestamp, and the user requests "update database from selected xmp files", db is updated with xmp data (as today) but:

- db write time stamp is updated with the xmp file timestamp
 - the xmp file is not updated (as now db data = xmp data) even if "write sidecar file for each image" is set.

A g_free() is also added, which should avoid a memory leak. Would that help #5722 ?

EDIT: as pinpointed by @esq4 fixes also the wrong update of import_timestamp in dt_image_cache_write_release()

### DIFF
--- a/src/common/history.c
+++ b/src/common/history.c
@@ -173,7 +173,9 @@ int dt_history_load_and_apply(const int imgid, gchar *filename, int history_only
     /* if current image in develop reload history */
     if(dt_dev_is_current_image(darktable.develop, imgid)) dt_dev_reload_history_items(darktable.develop);
 
-    dt_image_cache_write_release(darktable.image_cache, img, DT_IMAGE_CACHE_SAFE);
+    dt_image_cache_write_release(darktable.image_cache, img,
+    // ugly but if not history_only => called from crawler - do not write the xmp
+                                 history_only ? DT_IMAGE_CACHE_SAFE : DT_IMAGE_CACHE_RELAXED);
     dt_mipmap_cache_remove(darktable.mipmap_cache, imgid);
     dt_image_reset_final_size(imgid);
   }
--- a/src/common/image_cache.c
+++ b/src/common/image_cache.c
@@ -251,7 +251,7 @@ void dt_image_cache_write_release(dt_image_cache_t *cache, dt_image_t *img, dt_i
       "crop = ?15, orientation = ?16, raw_parameters = ?17, group_id = ?18, longitude = ?19, "
       "latitude = ?20, altitude = ?21, color_matrix = ?22, colorspace = ?23, raw_black = ?24, "
       "raw_maximum = ?25, aspect_ratio = ROUND(?26,1), exposure_bias = ?27, "
-      "change_timestamp = ?28, change_timestamp = ?29, export_timestamp = ?30, print_timestamp = ?31 "
+      "import_timestamp = ?28, change_timestamp = ?29, export_timestamp = ?30, print_timestamp = ?31 "
       "WHERE id = ?32",
       -1, &stmt, NULL);
   DT_DEBUG_SQLITE3_BIND_INT(stmt, 1, img->width);
--- a/src/control/crawler.c
+++ b/src/control/crawler.c
@@ -23,6 +23,7 @@
 #include <string.h>
 
 #include "common/darktable.h"
+#include "common/debug.h"
 #include "common/database.h"
 #include "common/history.h"
 #include "common/image.h"
@@ -42,6 +43,7 @@ typedef enum dt_control_crawler_cols_t
   DT_CONTROL_CRAWLER_COL_XMP_PATH,
   DT_CONTROL_CRAWLER_COL_TS_XMP,
   DT_CONTROL_CRAWLER_COL_TS_DB,
+  DT_CONTROL_CRAWLER_COL_TS,    // new timestamp to db
   DT_CONTROL_CRAWLER_NUM_COLS
 } dt_control_crawler_cols_t;
 
@@ -265,16 +267,33 @@ static void _reload_button_clicked(GtkButton *button, gpointer user_data)
   {
     gboolean selected;
     int id;
-    gchar *xmp_path;
-    gtk_tree_model_get(gui->model, &iter, DT_CONTROL_CRAWLER_COL_SELECTED, &selected,
-                       DT_CONTROL_CRAWLER_COL_ID, &id, DT_CONTROL_CRAWLER_COL_XMP_PATH, &xmp_path, -1);
+    gchar *xmp_path = NULL;
+    time_t timestamp;
+    gtk_tree_model_get(gui->model, &iter,
+                       DT_CONTROL_CRAWLER_COL_SELECTED, &selected,
+                       DT_CONTROL_CRAWLER_COL_ID, &id,
+                       DT_CONTROL_CRAWLER_COL_XMP_PATH, &xmp_path,
+                       DT_CONTROL_CRAWLER_COL_TS, &timestamp,
+                       -1);
     if(selected)
     {
+      // align db write timestamp on xmp file timestamp
+      sqlite3_stmt *stmt;
+      DT_DEBUG_SQLITE3_PREPARE_V2
+        (dt_database_get(darktable.db),
+         "UPDATE main.images SET write_timestamp = ?2 WHERE id = ?1",
+         -1, &stmt, NULL);
+      DT_DEBUG_SQLITE3_BIND_INT(stmt, 1, id);
+      DT_DEBUG_SQLITE3_BIND_INT(stmt, 2, timestamp);
+      sqlite3_step(stmt);
+      sqlite3_finalize(stmt);
+
       dt_history_load_and_apply(id, xmp_path, 0);
       valid = gtk_list_store_remove(GTK_LIST_STORE(gui->model), &iter);
     }
     else
       valid = gtk_tree_model_iter_next(gui->model, &iter);
+    g_free(xmp_path);
   }
   // we also want to disable the "select all" thing
   _clear_select_all(gui);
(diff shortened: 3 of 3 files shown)


phweyland:
Pascal, I think there are 2 fixes you could add directly to 3.2:
- g_free() in crawler.c
- import_timestamp in image_cache.c

TurboGit:
It is too late for 3.2, the tarball is already created (two times already). And we need to let time to packagers of the WIndows and MacOS version. This is really annoying issue but not critical. I'll add a note on the release notes about this possible issue.

Can you help me describe the more precisely in which context it happens and how to avoid it?

Thanks.

phweyland:
> Can you help me describe the more precisely in which context it happens and how to avoid it?

#5722 & #5847 are related to the usage of a stock of images + xmp from two different systems and/or databases with "update database from selected xmp files" (that means that both "write sidecar file for each image" and "check xmp on start" are set).

In that case, the current system sees the xmp timestamps left by the other system and show up the corresponding list of images. If the user  click on "update database from selected xmp files" dt updates the current database but also the xmp files, changing again the xmp timestamps. So each time the user switches to the other system and/or database, dt triggers again the change detection.

Workaround ? Not really but the effect can be limited. Before "update database from selected xmp files" only select the modified images. Or re-import manually these modified images or folders. That will limit the issue to those modified folders or images (instead of all images).

This PR aligns the timestamps of the current database with the current xmp timestamps and removes the re-write of the xmp files (which is not needed as the database has just been aligned on them). So the detection does not occur again on both system and/or database as long as no change is made by the user.

The bug on dt_image_cache_write_release() has no workaround. The import timestamp can be wrong (overwritten by change timestamp).

esq4:
> In that case, the current system sees the xmp timestamps left by the other system and show up the corresponding list of images. If the user click on "update database from selected xmp files" dt updates the current database but also the xmp files, changing again the xmp timestamps. So each time the user switches to the other system and/or database, dt triggers again the change detection.

Exactly!

> Workaround ? Not really but the effect can be limited. Before "update database from selected xmp files" only select the modified images. Or re-import manually these modified images or folders. That will limit the issue to those modified folders or images (instead of all images).

That is, I must have to manually do the work that should (and used to be) done automatically. I must remember what I edited and try to ignore the rest.

phweyland:
> That is, I must have to manually do the work that should (and used to be) done automatically. I must remember what I edited and try to ignore the rest. Regression.

My comments there are related to 3.2 for which the PR is not applicable anymore.

Now it would be great if you could build the current master with that PR and tell us if that solves your issue or not.
And if not what are the remaining issues.
Thanks

esq4:
I built the current master with your PR. It solves my issue but there is a couple of flaws.
1. Import timestamp and change timestamp are different in different db (I have no other timestamps). Was it intended?
2. No timestamps are displayed on Windows (they are present in the db)

timestaps in windows 7

phweyland:
> I built the current master with your PR. It solves my issue but there is a couple of flaws.

Thanks a lot for testing.

Could you share the values of timestamps you in db and xmp file for that image ?

phweyland:
>     2\. No timestamps are displayed on Windows (but they are present in the db)

Does that means they are displayed on Linux ?

esq4:
> Could you share the values of timestamps you have in db and xmp file for that image ?

When I changed the image in Linux, the change_timestamp values in XMP and db became equal 1596659390.

After transferring collection to Windows, the change_timestamp value was -1 in the db.

When I changed the image in Windows the change_timestamp values in XMP and db became equal 1596732261.



> Does that means they are displayed on Linux ?

Yes of course.

phweyland:
>     1. Import timestamp and change timestamp are different in different db (I have no other timestamps). Was it intended?

Import and change are two different events so that seems normal to have them different.

>     2\. No timestamps are displayed on Windows (but they are present in the db)

> After transferring collection to Windows, the change_timestamp value was -1 in the db.

Seems these issues are specific to windows. I don't reproduce them on linux.

Replaying this my finding is the following: timestamps are updated in db, but in cache. Then when the cache is written, timestamps (all) are lost.
If I succeed in fixing this I'll issue another PR.

phweyland:
@TurboGit Pascal, I'm fine with PR. You can review it when you want.

esq4:
> > ```
> > 1. Import timestamp and change timestamp are different in different db (I have no other timestamps). Was it intended?
> > ```
> 
> 
> Import and change are two different events so that seems normal to have them different.

No, not like that. I mean:
import timestamps are different in different db;
change timestamps are different in different db.

Was it intended?

phweyland:
> > > ```
> > > 1. Import timestamp and change timestamp are different in different db (I have no other timestamps). Was it intended?
> > > ```
> > 
> > 
> > Import and change are two different events so that seems normal to have them different.
> 
> No, not like that. I mean:
> import timestamps are different in different db (on Linux and Windows in my case);
> change timestamps are different in different db (on Linux and Windows in my case).
> 
> Was it intended?

Sorry for the misunderstanding.
As explained in #5879, the timestamps are not copied (imported) for the time being. This explains (at least partly) what you see.
This said, import_timestamp is not copied intentionally and I haven't changed this.
So import_timestamp remains specific to each database.

esq4:
> As explained in #5879, the timestamps are not copied (imported) for the time being.

"For the time being" is good.

> This said, import_timestamp is not copied intentionally and I haven't changed this.
> So import_timestamp remains specific to each database.

I do not understand the reason for this decision and I think it is not correct. Let me explain.
I suppose we can assume that the db is an image of all changes made to the collection, a kind of hash. Since the collection can change on different computers or systems, the hashes must be the same. Otherwise, the meaning of such a convenient Darktable function is blurred.

esq4:
> @TurboGit Pascal, I'm fine with PR. You can review it when you want.

It did not help.

ptilopteri:
@esq4 
> > This said, import_timestamp is not copied intentionally and I haven't changed this.
> > So import_timestamp remains specific to each database.
> 
> I do not understand the reason for this decision and I think it is not correct. Let me explain.
> I suppose we can assume that the db is an image of all changes made to the collection, a kind of hash. Since the collection can change on different computers or systems, the hashes must be the same. Otherwise, the meaning of such a convenient Darktable function is blurred.
> 
> In other words, it doesn't matter to me when I imported the image on this particular computer. It is important for me when I imported a picture from the camera.

I would not expect two different library db's on different computers to have the save information unless subject image was just imported and had the same xmp as the "other" computer/database.

You are LOOKING for trouble trying to work images on different computers with different databases.  Better workflow having images on shared drive and main computer with library.  Other computer(s) utilize "darktable --library :memory:" and write xmp files on.  When on main computer, reimport images to update database, or set dt to look for changed xmp files and have them saved to your library.

esq4:
> You are LOOKING for trouble trying to work images on different computers with different databases. Better workflow having images on shared drive and main computer with library. Other computer(s) utilize "darktable --library :memory:" and write xmp files on. When on main computer, reimport images to update database, or set dt to look for changed xmp files and have them saved to your library.

I'm not LOOKING for problems, I just want to work FAST and EFFICIENTLY. A process that requires additional tweaks, actions and time could not be the best.

I will explain. I use tags to quickly find images. And I want to do this on all my computers.
Right now I just need to connect the disk and update the db at the start of dt. Now, thanks to you, this operation takes seconds.

According to the methodology you proposed, as a simple user, I first need to make special settings for launching dt (i. e. at first study and understand the documentation, and most importantly, then INVENT this method). And after that, every time I have to import the ENTIRE collection. But in collection, for example, I already have more than 10 thousand images. It's even scary to think about the waiting time.

esq4:
> I would not expect two different library db's on different computers to have the save information unless subject image was just imported and had the same xmp as the "other" computer/database.

If it is forbidden to transfer "import timestamp" from one db to XMP and to another db (although I still do not understand the reason for this prohibition), then maybe you will introduce a new parameter to display the time of import from the camera, for example, "from camera timestamp"?

phweyland:
> I would not expect two different library db's on different computers to have the save information unless subject image was just imported and had the same xmp as the "other" computer/database.

I can agree this is not the basic usage of dt.
But this function (xmp detection) should work properly even in that case (it seems that various people (try to) work this way).

I don't use xmp, except manually when I have a specific transfer to do. I prefer counting on a solid backup of the database.
If I had to work (alone) on several computers (same OS) I would share the db. If I had to use several computers with different OS, I would implement a kind of switch updating film_rolls table accordingly (If I remember well digikam has some mechanism which turn this quite easy). It is what I do manually when I need to run some tests on windows.

ptilopteri:
> According to the methodology you proposed, as a simple user, I first need to make special settings for launching dt (i. e. at first study and understand the documentation, and most importantly, then INVENT this method). 

No, you do not, just start dt as:  "darktable --library :memory:"
really not that difficult.

> And after that, every time I have to import the ENTIRE collection. But in collection, for example, I already have more than 10 thousand images. It's even scary to think about the waiting time.

guess you failed to read what I proposed.  I follow this and it is fairly fast and not difficult.  But you continue to struggle and find problems and then complain when there is a workflow that just works.

I will bother you no more!

esq4:
@phweyland 
> If I had to use several computers with different OS, I would implement a kind of switch updating film_rolls table accordingly (If I remember well digikam has some mechanism which turns this quite easy). It is what I do manually when I need to run some tests on windows.

Please write in more detail how to implement this. What if it can be implemented in the dt? ;)

https://github.com/darktable-org/darktable/pull/5879
This didn't resolve the issue in Windows.

But still
> If it is forbidden to transfer "import timestamp" from one db to XMP and to another db (although I still do not understand the reason for this prohibition; _explain it, please_), _then maybe you will introduce a new parameter_ to display the time of import from the camera, for example, "from camera timestamp"?

phweyland:
> This didn't resolve the issue in Windows.

I am not surprised. That's probably another bug which would need some debug on windows itself.



> But still

As already said please open a specific issue for this.
On my side I won't do more before these 2 PR are merged (if accepted of course).

esq4:
@phweyland Well. Thank you.

TurboGit:
@phweyland : can you do a quick review of #5896 ?

PR #5879, opened 2020-08-06
fix read_xmp_timestamps() - write to cache instead of db

phweyland:
in `dt_exif_xmp_read()`, image cache being write protected,  `read_xmp_timestamps() `writes the timestamps to cache instead of db (where they are lost with `dt_image_cache_write_release()`).

I've left import_timestamp not updated as commented in the code. 

Fixes some issue discussed in #5847 and #5869 and pinpointed by @esq4 .

### DIFF
--- a/src/common/exif.cc
+++ b/src/common/exif.cc
@@ -288,7 +288,7 @@ public:
 }
 
 static void _exif_import_tags(dt_image_t *img, Exiv2::XmpData::iterator &pos);
-static gboolean read_xmp_timestamps(Exiv2::XmpData &xmpData, const int imgid);
+static void read_xmp_timestamps(Exiv2::XmpData &xmpData, dt_image_t *img);
 
 // this array should contain all XmpBag and XmpSeq keys used by dt
 const char *dt_xmp_keys[]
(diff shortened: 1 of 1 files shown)


phweyland:
I quote a comment from @esq4 on #5869 because this thread is more appropriate:
  
> > This said, import_timestamp is not copied intentionally and I haven't changed this.
> > So import_timestamp remains specific to each database.
> 
> I do not understand the reason for this decision and I think it is not correct. Let me explain.
> I suppose we can assume that the db is an image of all changes made to the collection, a kind of hash. Since the collection can change on different computers or systems, the hashes must be the same. Otherwise, the meaning of such a convenient Darktable function is blurred.
> 
> In other words, it doesn't matter to me when I imported the image on this particular computer. It is important for me when I imported a picture from the camera.

I can agree with what you say but I think that has been done like that to make sure that import_timestamp reflects the last import operation (and not the value in xmp). 
Then doing this would be more a behavior change than bug fixing.

Feel free to open a specific issue for this if you think it is worth to. And ping the author @jpverrue for more input.

PR #5880, opened 2020-08-06
Update Hungarian translation

BathoryPeter:


BathoryPeter:
Uhh, there is some weird error in the po, the last ~500 row is the duplication of lines starting from 9090 :S
Closing this and create a new pr…

PR #5881, opened 2020-08-07
imageio: Fix possible compile error in avif.c

cryptomilk:
Missing include for `bool`.


TurboGit:
In darktable we use gboolean and not C++ bool so the proper fix may be to change bool to gbooelan in avif.c.

cryptomilk:
This is a C99 bool not C++. I guess we do C99 right?

parafin:
That's not the point. The fact that you technically can use bool, doesn't mean that it matches dt's code convention. E.g. if you need to pass boolean value between different darktable parts, you must choose some specific type, which is @TurboGit said is gboolean.

cryptomilk:
Moved to gboolean ...

PR #5882, opened 2020-08-07
Update Hungarian translation

BathoryPeter:


BathoryPeter:
@TurboGit I just noticed that Hungarian translation is missing from the new release :(
Deadline for translations was August 8th and I sent this PR one day before. I spent all my free time for two weeks to be ready before the release, so I'm really sad now.

TurboGit:
I understand your frustration, I'm the culprit as I forgot to count the days needed to build the binaries for the final release. We always count around one full week to let maintainers of the Windows & MacOS version the time to build and possibly fix OS specific issues. So again I was wrong when I told about August 8th and the time constraint was quite high on this release that I forgot about this, not counting that last week I was away from home with little connection.

All your work will be in 3.4 just some months away.

BathoryPeter:
We will have a lot of frustrated user in Hungary. I'm sure many of them don't speak English at all, so completely removing Hungarian language for a couple of month is a huge loss. Even worse than simply leave the translation at around 80% completeness.
Additionally 3.2 will likely be packaged in Ubuntu 20.10. For half a year and for a wider user base.

I would like to ask you to consider a new point release.

PR #5891, opened 2020-08-08
Adding developers demosaic algo showing photosite values

jenshannoschwalm:
While working on fixing chromatic aberration issues i needed a dt-internal way to
find out the raw values from the photosites represented as rgb colors.

No opencl implementation done as this is not intended for user photo developing but
for analysing color details so not performance critical.


aurelienpierre:
Works, thanks !

PR #5928, opened 2020-08-11
export: replace max size spin buttons with entry widgets

### LINKED ISSUE #5366: Pixel size on output can be disturbed by mouse wheel
I noticed my max size on outputting in lighttable (the max pixel dimensions) had got altered.  The mouse wheel lets me move up and down the lighttable sections, as expected, and it doesn't allow values to be changed, as expected, EXCEPT for these two pixel values. So it's quite easy to accidentally change them as you flick around the right hand side of the screen.  I suppose someone will say you can use the scroll bar, but that's rather fiddly compared to just mouse wheeling  anywhere on the right.

dt master 3.1.0+2017~ga6c18e1b7-dirty    built today.

elstoc:
spin buttons don't really make much sense for pixel values

removal of the spin buttons means that the right hand panel may be
reduced in size without the panel UI shifting and means that the values
are not inadvertently disturbed when scrolling the panel

Resolves #5366

### DIFF
--- a/src/libs/export.c
+++ b/src/libs/export.c
@@ -39,6 +39,7 @@
 #include <gdk/gdkkeysyms.h>
 #include <gtk/gtk.h>
 #include <stdlib.h>
+#include <ctype.h>
 
 DT_MODULE(7)
 
@@ -47,9 +48,10 @@ DT_MODULE(7)
 
 typedef struct dt_lib_export_t
 {
-  GtkSpinButton *width, *height;
+  GtkWidget *width, *height;
   GtkWidget *storage, *format;
   int format_lut[128];
+  uint32_t max_allowed_width , max_allowed_height;
   GtkWidget *upscale, *profile, *intent, *style, *style_mode;
   GtkButton *export_button;
   GtkWidget *storage_extra_container, *format_extra_container;
@@ -220,13 +222,22 @@ static void export_button_clicked(GtkWidget *widget, dt_lib_export_t *d)
   g_free(icc_filename);
 }
 
+void _set_dimensions(dt_lib_export_t *d, int max_width, int max_height)
+{
+  gchar *max_width_char = g_strdup_printf("%d", max_width);
+  gchar *max_height_char = g_strdup_printf("%d", max_height);
+  gtk_entry_set_text(GTK_ENTRY(d->width), max_width_char);
+  gtk_entry_set_text(GTK_ENTRY(d->height), max_height_char);
+  g_free(max_width_char);
+  g_free(max_height_char);
+}
+
 void gui_reset(dt_lib_module_t *self)
 {
   // make sure we don't do anything useless:
   if(!dt_control_running()) return;
   dt_lib_export_t *d = (dt_lib_export_t *)self->data;
-  gtk_spin_button_set_value(d->width, dt_conf_get_int(CONFIG_PREFIX "width"));
-  gtk_spin_button_set_value(d->height, dt_conf_get_int(CONFIG_PREFIX "height"));
+  _set_dimensions(d, dt_conf_get_int(CONFIG_PREFIX "width"), dt_conf_get_int(CONFIG_PREFIX "height"));
 
   // Set storage
   gchar *storage_name = dt_conf_get_string(CONFIG_PREFIX "storage_name");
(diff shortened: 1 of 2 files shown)


dterrahe:
Have you considered using a flow box? This could be an alternative approach (but I'm not objecting to yours either!)
```
  GtkBox *hbox; // TODO: move declaration to next use. Why was there a double hbox?
  GtkFlowBox *hbox1 = GTK_FLOW_BOX(gtk_flow_box_new());
  gtk_widget_set_name(GTK_WIDGET(hbox1), "export-max-size");
  label = gtk_label_new(_("max size"));
  gtk_label_set_ellipsize(GTK_LABEL(label), PANGO_ELLIPSIZE_END);
  gtk_label_set_xalign(GTK_LABEL(label), 0.0);
  gtk_container_add(GTK_CONTAINER(hbox1), label);
  gtk_container_add(GTK_CONTAINER(hbox1), GTK_WIDGET(d->width));
  label = gtk_label_new(_("x"));
  gtk_label_set_xalign(GTK_LABEL(label), 1.0);
  gtk_container_add(GTK_CONTAINER(hbox1), label);
  gtk_container_add(GTK_CONTAINER(hbox1), GTK_WIDGET(d->height));
  gtk_flow_box_set_min_children_per_line(hbox1, 2);
  gtk_box_pack_start(GTK_BOX(self->widget), GTK_WIDGET(hbox1), FALSE, TRUE, 0);
```
I like the consistency of being able to change values with the scroll wheel. I don't think it would be hard to make these controls respect "mouse wheel scrolls modules side panel by default" (using dt_gui_ignore_scroll to be introduced in #5755)

TurboGit:
And what about adding a little button at the end of the line with pre-defined values for the sizes:
- max
- HD
- Full-HB
- ..

Where for example selecting max from the combo would fill the entries with 0 x 0.

elstoc:
> I don't think it would be hard to make these controls respect "mouse wheel scrolls modules side panel by default"

I agree. Really my main motivation was to remove the minimum panel width that these controls imposed.

> Have you considered using a flow box?

Not really. I'm not sure how this would apply?

Of course one alternative would be to place the max width spin button on top of the max height spin button (rather than have them side-by-side).  This was really way more code than I wanted to write for something so simple (it really should have been possible to just hide the buttons in a GtkSpinButton control but apparently they're not really GtkButtons).

> And what about adding a little button...

Maybe. Let's agree what to do with the boxes first because I spent a while on this and I'd like to only change it once more!

dterrahe:
> > Have you considered using a flow box?

> Not really. I'm not sure how this would apply?

My code snippet (to replace the existing setup in gui_init) makes the second size box flow to the next line when they don't both fit, so that also avoids imposing a minimum panel size.

To save you having to compile it; this is what it looks like with a wide panel:
image
and this with a narrow one:
image
There's a bit in the middle where it is somewhat messy :-)
image

Even if you don't want to use this here, it might be useful later on somewhere else in your quest for narrowest panels ;-)

Sort of independently, the current code puts an hbox1 inside an hbox before putting it in self->widget. I think that's redundant (unless I'm unaware of an intended side-effect). Might as well clean that up.

elstoc:
That might work. I'm not sure how the flow stuff works but if you were to put the 'max size' label and first spinbutton in one hbox and the 'x' and the second spinbutton in another, you might be able to get them to flow as a block? Then have the label at the start and spin button at the end of each box (and with the spin buttons having fixed widths)? Sounds like you're some way to a reasonable alternative. Do you want to try as a 'competing' PR? I'm reluctant to re-code this one unless we're agreed on a reasonable alternative.

TurboGit:
I'm not fond of the flow in this case, cutting the entry on multiple lines is not clean.

elstoc:
Unless you did it permanently and called it 'max height' and 'max' width' as two clearly separate spin buttons?

johnny-bit:
/me votes for single-entry with button for size presets (and maaaybe ability to do calc on values eg you input `512*2` and it corrects it to `1024`)

:wink:

dterrahe:
> Sounds like you're some way to a reasonable alternative.

To be honest, I've been playing around with a few different layerings of hbox and flow boxes and it doesn't seem possible to always make it do the right thing (unless, maybe, I start playing around with GtkSizeGroups, but I'm not willing to sink that low).

Also, I don't want to antagonise Pascal!! (no matter how much I disagree with him! ;-) )

Except on the point of "adding a little button"; that sounds like he wants to invite a crop&rotate disaster into this PR. My suggestion would be to separate that out into a new PR (that can then be easily reverted!)

elstoc:
I don't know. I think the only issue with an extra button is the possibility that it increases the width again. I suspect maybe presets are the better solution.

dterrahe:
Presets would affect all other settings in the module, which doesn't really make sense.

Hm... “populate-popup”... just kidding!!

dterrahe:
In case it isn't clear; I have no objections against proceeding with this PR as is (although I haven't had time to test it).

An extra button for default sizes could be a separate PR.

elstoc:
@TurboGit how would you like me to proceed with this one?

TurboGit:
I would remove the spin buttons, do not use a flow box. Not looked at this recently but I think this is exactly what is in this PR.

From there that's up to you if you want to add the buttons with selectable common aspect ratio in this PR or in another or not at all :) Let me know.

elstoc:
> I would remove the spin buttons, do not use a flow box. Not looked at this recently but I think this is exactly what is in this PR.

Yes that's what's in this PR. I'm not bothered personally about common aspect ratio box (I mostly use either 0x0 or 1920x1920) so let's leave that for another PR (if at all).

src/libs/export.c
```diff
@@ -480,10 +504,39 @@ static void profile_changed(GtkWidget *widget, dt_lib_export_t *d)
...
+  _validate_dimensions(d);
+  dt_conf_set_int(CONFIG_PREFIX "height", atoi(gtk_entry_get_text(GTK_ENTRY(d->height))));
+}
+
+static void insert_text_handler(GtkEditable *entry, char *text, int length, gpointer position, gpointer user_data)
+{
+  int i, count=0;
+  gchar *result = g_new (gchar, length);
+
+  for (i=0; i < length; i++) {
```
TurboGit, line 526: I'll fix that, but please not that { should be on its own line.

TurboGit:
Thanks !

src/libs/export.c
elstoc: I agree. I didn't see that. I think I 'borrowed' some of this code from elsewhere.

PR #5930, opened 2020-08-11
adds ccache option to CMake by lefticus:

codingdave:
CMake can be set up to optionally use compiler caches to speed up the compilation. This PR is adding support for cached builds.

See https://github.com/lefticus/cpp_starter_project/blob/master/cmake/Cache.cmake for the original source

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -22,6 +22,9 @@ else()
   project(darktable VERSION 0 LANGUAGES CXX C)
 endif()
 
+# enable cache system
+include(Cache)
+
 # Allow forcing the C/CPP compiler that is actually used during the compilation
 # to something other than what is used by the cmake run. This is useful when
 # the compiler for some reason breaks the initial cmake checks but works fine
--- a/cmake/modules/Cache.cmake
+++ b/cmake/modules/Cache.cmake
@@ -1,0 +1,29 @@
+option(ENABLE_CACHE "Enable cache if available" ON)
+if(NOT ENABLE_CACHE)
+  return()
+endif()
+
+set(CACHE_OPTION
+    "ccache"
+    CACHE STRING "Compiler cache to be used")
+set(CACHE_OPTION_VALUES "ccache" "sccache")
+set_property(CACHE CACHE_OPTION PROPERTY STRINGS ${CACHE_OPTION_VALUES})
+list(
+  FIND
+  CACHE_OPTION_VALUES
+  ${CACHE_OPTION}
+  CACHE_OPTION_INDEX)
+
+if(${CACHE_OPTION_INDEX} EQUAL -1)
+  message(
+    STATUS
+      "Using custom compiler cache system: '${CACHE_OPTION}', explicitly supported entries are ${CACHE_OPTION_VALUES}")
+endif()
+
+find_program(CACHE_BINARY ${CACHE_OPTION})
+if(CACHE_BINARY)
+  message(STATUS "${CACHE_OPTION} found and enabled")
+  set(CMAKE_CXX_COMPILER_LAUNCHER ${CACHE_BINARY})
+else()
+  message(WARNING "${CACHE_OPTION} is enabled but was not found. Not using it")
+endif()


parafin:
I don't think it's useful. ccache can be activated just by setting CC and CXX, no need to add more cmake code anywhere.

codingdave:
Well, there are reasons for it:
* It works fine
* It is clean and transparent
* It is isolated and does not have any negative impact on other code
* It is caching the users choice
* It turns on faster compilation for people that have installed ccache but forgot about it
* CC and CXX works but is not elegant, can be forgotten, and is more a hack than a supported feature

parafin:
Then let me rephrase - I don't think adding a new supported feature is desired. More code and options - more difficult to support. Setting CC and CXX is very easy and then responsibility is moved to user instead of darktable project. I am against adding more cmake code with no clear advantage.

cryptomilk:
@codingdave the right code for adding ccache support would be:

```
option(WITH_CCACHE "Build with ccache" OFF)
if (WITH_CCACHE)
    find_program(CCACHE_BINARY ccache)
    if (CCACHE_BINARY)
        set_property(GLOBAL PROPERTY RULE_LAUNCH_COMPILE ${CCACHE_BINARY})
        set_property(GLOBAL PROPERTY RULE_LAUNCH_LINK ${CCACHE_BINARY})
    endif()
endif()
```

a) You should not turn it on by default.
b) You also want to use it for linking

parafin:
As far as I know ccache doesn't do linking, so it's pointless to use as linker.

cryptomilk:
`man ccache`

```
called for link │ The compiler was called for linking, not compiling.
```

parafin:
It doesn't mean it does anything (speed things up) in this mode. See https://ccache.dev/ - "Other types of compilations (multi-file compilation, linking, etc) will silently fall back to running the real compiler."

codingdave:
> @codingdave the right code for adding ccache support would be:
> 
> ```
> option(WITH_CCACHE "Build with ccache" OFF)
> if (WITH_CCACHE)
>     find_program(CCACHE_BINARY ccache)
>     if (CCACHE_BINARY)
>         set_property(GLOBAL PROPERTY RULE_LAUNCH_COMPILE ${CCACHE_BINARY})
>         set_property(GLOBAL PROPERTY RULE_LAUNCH_LINK ${CCACHE_BINARY})
>     endif()
> endif()
> ```
> 
> a) You should not turn it on by default.
> b) You also want to use it for linking

Hi @cryptomilk, 

thanks for the feedback. Regarding
A: For me having this turned ON makes sense - I mean we even build the docs per default. But honestly I leave that decision to the project owner
B: The discussion above shows that this is questionable. Im fine with any decision here as well.

Also I like your nice small and compact code for turning on ccache. However this PR also works for sccache. So the functionality slightly differs. 

For all questions 
1. Do we want to have this in DT
2. Do we want this to be on per default
3. Do we want to turn on linker support
I am fine with whatever the PO decides.

@TurboGit is it fair to call you PO? If not, can you please forward the decision to someone? Thank you.

parafin:
1. This PR doesn't set CMAKE_C_COMPILER_LAUNCHER, so only C++ sources are compiled with ccache, which are the minority of darktable's source code.
2. I would be very concerned if ccache affected linking. It's quite easy to test by timing the build with and without using ccache for linking. On the other hand using it for linking too doesn't hurt much. But RULE_LAUNCH_COMPILE and RULE_LAUNCH_LINK are wrong properties to use for this anyway, and there is no CMAKE_LINKER_LAUNCHER variable it seems, so using ccache for linking in cmake is also hacky.
3. If WITH_CCACHE is OFF by default, then you're just switching `-DCMAKE_C_COMPILER_LAUNCHER=ccache -DCMAKE_CXX_COMPILER_LAUNCHER=ccache` cmake option for `-DWITH_CCACHE=ON` by adding more cmake code. I see this as totally pointless exercise.

codingdave:
I would have been happy to have that inside dt, cause I often fail to turn that on correctly using environment variables and I don't know why it fails in that scenario. 
Until I find out if this is a wanted feature or doomed to die I will not spend further work on it. 

> If WITH_CCACHE is OFF by default, then you're just switching -DCMAKE_C_COMPILER_LAUNCHER=ccache -DCMAKE_CXX_COMPILER_LAUNCHER=ccache cmake option for -DWITH_CCACHE=ON by adding more cmake code. I see this as totally pointless exercise.

`CMAKE_C_COMPILER_LAUNCHER` and `DCMAKE_CXX_COMPILER_LAUNCHER` are not cached. So the user needs to re-enter that everytime he wants to use it. `WITH_CCACHE` on the other hand is cached, so it is pretty convenient.

parafin:
Do you often re-run cmake without cleaning the build directory? If so, why, what is the use-case?..

codingdave:
If I am testing/developing a branch, yes. Also you could have a build-directory for each branch, like I have with many things (configs, PREFIX, ...).

parafin:
Maybe I'm missing something, but why re-run the cmake? Why not just do make? Though make sometimes re-runs cmake itself, I wonder if `CMAKE_C_COMPILER_LAUNCHER` gets reset in this case too...

OK, caching might be a real argument for this feature, but I'm not convinced that it's needed. Someone else should add their opinion and decide.

parafin:
You can add CMAKE_C_COMPILER_LAUNCHER and CMAKE_CXX_COMPILER_LAUNCHER to your environment, then you don't have to remember to set it in each cmake call.

codingdave:
```cat ~/bin/darktable_standard_build.sh 
#!/bin/bash
git fetch --all
env CMAKE_MORE_OPTIONS="-DRAWSPEED_ENABLE_LTO=ON -DBUILD_CURVE_TOOLS=ON -DBUILD_NOISE_TOOLS=ON" CMAKE_C_COMPILER_LAUNCHER=ccache CMAKE_CXX_COMPILER_LAUNCHER=ccache

./build.sh \
    --build-dir build \
    --prefix ~/unstable/darktable_$(git rev-parse --abbrev-ref HEAD) \
    --jobs 4 --install
```

```
CMAKE_C_COMPILER_LAUNCHER=ccache

parafin:
Well, that's not how you set environment variables. CMAKE_MORE_OPTIONS won't work either. You have to use "export" command. I suggest you read a little bit about how environment variables work in UNIX and shell. Specifically they are not global, but inherited from parent to child process, and env is such, since it's not a built-in shell command.

codingdave:
I can confirm that fixing my envs is working fine on linux.

PR #5938, opened 2020-08-12
Remove unused code in buttons

Mark-64:
Removing some dead branch of code in button.c and togglebutton.c, which only manage icon buttons, while label button are created with regular gtk calls.

### DIFF
--- a/src/dtgtk/button.c
+++ b/src/dtgtk/button.c
@@ -55,19 +55,6 @@ static gboolean _button_draw(GtkWidget *widget, cairo_t *cr)
   else
     flags &= ~CPF_PRELIGHT;
 
-  /* create pango text settings if label exists */
-  PangoLayout *layout = NULL;
-  int pw = 0, ph = 0;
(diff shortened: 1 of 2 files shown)


TurboGit (commented):
> while label button are created with regular gtk calls.

Today maybe, but I'm not sure removing support for label in dtgtk buttons is a good thing. The dtgtk code is generic API to be used freely by dt. So I don't think removing this is a good idea.

Mark-64:
dtgtk buttons changed deeply in 3.2, in orderd to make them compatible with css and HiDPI / Windows.
Most of these changes have been done by me, but I never touched the part of the code which is meant to manage labels.
I believe 99% that code is no longer functional and I can't even fix it as it is not used anywhere and I don't know how it is supposed to work.
To me it would be cleaner to remove it and add it again if and when it is needed.
But you decide.

TurboGit:
Understood, so let's merge this then. Thanks for the work and explanations.

PR #5940, opened 2020-08-12
Restore separate saturation soft limits in color balance

### LINKED ISSUE #5939: PR #5691 breaks soft limits for saturation in Color Balance
DT 3.3 in Ubuntu 20.04
After #5691, shadows/midtones/highlights saturation sliders in Color Balance have all a soft max of 5%, while they should be 5% / 20% / 25%

dterrahe:
fixes #5939


TurboGit:
Working on my side, thanks!

PR #5691, opened 2020-07-02
Transition negadoctor, colorbalance and retouch to introspection framework

dterrahe:
with corresponding cleanup
add basic support for array fields to dt_bauhaus_slider_from_params

also add option for notebook tabs to colorbalance to make full module fit on screen

PR #5944, opened 2020-08-13
Fix stack corruption when creating slider from long variable name

dterrahe:
Only caught on osx and causes crash there. Reported against #5691


dterrahe:
Also incorporated a few helpful comments on IRC

parafin:
looks good

TurboGit:
Thanks !

MStraeten:
thanks for the fix - it's fine now

PR #5951, opened 2020-08-13
darkroom: only show module header when moving modules

elstoc:
when dragging and dropping modules that are expanded, this prevents the entire module ui being shown during the drag operation (which is cumbersome for modules with a large ui).

So...

Screenshot_2020-08-13_19-50-23

Rather than...

Screenshot_2020-08-13_19-52-58


TurboGit:
All good and indeed better this way!

PR #5957, opened 2020-08-14
preferences: display toast message if restart is required

### LINKED ISSUE #5241: Prominent message needed when a change in preferences requires dt to be restarted
I have the three items in preferences/processing about auto applying basecurve and sharpening unticked.  However when I import a new raw and go into darktable, these modules are On (surely they should be Off).  And to be sure, just before importing the new image, I ensure they are Off in the current image.

This is Master -   dt 3.1.0+1837~gee49c4778-dirty    built today 30/5/2020.  Ubuntu 18.04

elstoc:
if any settings are changed that require a restart, a toast message is displayed when exiting the preferences dialog

n.b. this works from the 'changed' signal (and its variants) so if a user changes a setting and then changes it back again the toast will still be displayed, but better safe than sorry.

Resolves #5241


elstoc:
There are probably other preferences that require a restart in order to take effect. I just did a search for the word 'restart' so I'm sure I haven't found them all.

PR #5959, opened 2020-08-15
Fix scripts installer not starting

### LINKED ISSUE #5885: lua scripts installer does not start
**Describe the bug**
When darktable 3.2.x is started it should display the _lua scripts installer_ module if the lua scripts are not installed, but it does not.

**To Reproduce**

1. Ensure lua scripts are not installed.
2. Start darktable 3.2.x
3. No lua scripts installer module appears in the lower left panel

**Expected behavior**
lua scripts installer should be present

 - Darktable Version: 3.2.x+
 - OS: all

**Additional context**
When I did PR #5857 I checked to ensure I fixed #5840.  I did not check to see if I introduced a regression.  I misread the API and ended up coding a test that would never pass, so the lua scripts installer will not start.  As it turns out, this may have been a fortunate error.  While troubleshooting this problem I discovered #5884.  If a lot of people had installed the scripts and had any action that would have started darktable in darkroom mode (i.e. front end to gimp, double click in file manager, etc.) then darktable would have hung on start.

#5884 must be fixed before this is fixed.  I will also feel more comfortable after trying this in a couple of release candidates, especially on windows.

wpferguson:
Added check to make sure we are in lighttable mode  If not we add an event listener to catch the view-changed event and then install the GUI module when we are back in lighttable.  Fixes #5885

### DIFF
--- a/data/luarc
+++ b/data/luarc
@@ -23,11 +23,14 @@ ipairs = function(t)
 
 local _scripts_install = {}
 
+_scripts_install.module_installed = false
+_scripts_install.event_registered = false
+
 _scripts_install.dt = require 'darktable'
 
 -- check for gui so that we don't hang darktable-cli
 
-if _scripts_install.dt.configuration.has_gui and _scripts_install.dt.gui.current_view() == "lighttable" then
+if _scripts_install.dt.configuration.has_gui  then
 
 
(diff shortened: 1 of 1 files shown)


TurboGit:
Thanks !

PR #5965, opened 2020-08-15
tether: properly initialize job name from conf

dtorop:
Complete the partial changeover from plugins/capture/jobcode to plugins/session/jobcode in 38b982b7725cc205a9746369779fcfa791890fbe.

This makes behavior consistent with the manual (default capture job is "capture job" rather than an empty string).


TurboGit:
Sounds good, thanks!

PR #5969, opened 2020-08-15
preferences: improved theme tweaks workflow

### LINKED ISSUE #5966: CSS theme tweaks in preferences need better workflow
**Describe the bug**

1. Changes to the tweaks text area are lost if preferences dialog is closed without pressing the save button
2. It's not obvious that one has to enable "modify selected theme" checkbox

**Expected behavior**

1. Either confirmation dialog should appear when trying to close preferences without saving the tweaks, or some very visible warning should be displayed near the text area
2. "modify selected theme" should be enabled by default or/and auto-switched on when saving theme tweaks or/and very visible notice should be displayed near the text area when checkbox is off

 - Darktable Version: 3.2.1
 - OS: macOS

elstoc:
changed label on css theme tweaks save button to read 'save and apply'
and amended its functionality so that it also automatically checks the
'modify selected theme...' checkbox before reloading the theme

this means that the save button always applies the theme immediately and
prevents issues where the user forgets to check the box

to prevent users inadvertently closing preferences without saving theme
tweaks, automatically save any changes on close (in line with the rest
of the screen). Changes will not be automatically applied (because they
are saved regardless of whether any edits were made and the user flag
may be unset)

theme tweaks check box is on by default

Resolves #5966.

### DIFF
--- a/data/darktableconfig.xml.in
+++ b/data/darktableconfig.xml.in
@@ -1,6 +1,12 @@
 <?xml version="1.0"?>
 <!DOCTYPE dtconfiglist SYSTEM "darktableconfig.dtd">
 <dtconfiglist>
+  <dtconfig>
+    <name>themes/usercss</name>
+    <type>bool</type>
+    <default>true</default>
+    <shortdescription>modify theme with user tweaks</shortdescription>
+  </dtconfig>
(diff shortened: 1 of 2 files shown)


TurboGit:
Works for me, thanks !

PR #5974, opened 2020-08-15
Non-local means denoising speedup  (v3)

ralfbrown:
Rebased and cleaned-up #5192.

@TurboGit: when checking the integration tests, I found that current master (**without** this PR) shows a max dE of 17.86 for 0020-denoise-nlmeans, with the difference pixels in the blown-out highlights and immediately adjacent, and max dE of 19.9745 for 0013-denoiseprofile-wavelets (again on the highlights).

ETA: Now switches denoise(profiled) to the new nlmeans code as well.


TurboGit:
@ralfbrown : indeed there is a recent regression on master which makes most tests failing :( I'm trying to find the culprit...

TurboGit:
The culprit is a00f9d30d1e4b3b1841fab4439065cb2d0fcb130

TurboGit:
Tested on my side (with #5978):

Before this PR with a new expected.png for test 0023:

```
Test 0020-denoise-nlmeans
      Image mire1.cr2
      CPU & GPU version differ by 985899 pixels
      Max  dE         0.0000
  OK

Test 0023-denoiseprofile-nlmeans
      Image mire1.cr2
      CPU & GPU version differ by 613936 pixels
      Max  dE         0.0000
  OK

```

Now compiling with this PR:

```
Test 0020-denoise-nlmeans
      Image mire1.cr2
      CPU & GPU version differ by 985918 pixels
      Max  dE         1.5742
  OK

Test 0023-denoiseprofile-nlmeans
      Image mire1.cr2
      CPU & GPU version differ by 613936 pixels
      Max  dE         0.0000
  OK
```

So test 0020 get some diff but 0023 does not. Are you sure the XMP is using the nlmeans algorithm?

TurboGit:
@ralfbrown : the regression on master is now fixed.

ralfbrown:
Test 0023 **is** using nlmeans, but as I mentioned above, the relevant code is not being changed by this PR, hence zero difference.  If I had included that commit, you would have seen max-dE of 3.5, average of 0.25, and 1.8 million changed pixels....

TurboGit:
@ralfbrown : Indeed, I read your comment and then forgot about this after fixing master... Sorry for the noise!

TurboGit:
@ralfbrown : This one ready now, right ?

ralfbrown:
Yes, it's ready.

TurboGit:
Thanks a lot, working fine on my side.

rawfiner:
I have approximately a 4x speedup with 4 threads on my computer, which is awesome!
Thanks a lot @ralfbrown for this work!

src/iop/denoiseprofile.c
```diff
@@ -41,10 +42,15 @@
 #include <xmmintrin.h>
 #endif
 
+// which version of the non-local means code should be used?  0=old (this file), 1=new (src/common/nlmeans_core.c)
+#define USE_NEW_IMPL_CL 0
```
dterrahe, line 46: @ralfbrown was trying to figure out if there was any dedup opportunities between `nlmeans_denoiseprofile_cl` and `nlmeans_denoise_cl` as they look fairly similar. Then came across this define and wondered if the plan still was to switch to the new code? Not completely clear from the discussion in the PR at the time. Maybe one of the two paths (and then unused code) could be cleaned up?

ralfbrown: Ideally we would use the new path, but I have no way of testing it (no working OpenCL here, video card is too old) so I never activated it.  If someone can verify that the new version gives the same results, we can switch to it and remove the old version.

PR #5978, opened 2020-08-16
Revert part of a00f9d30.

TurboGit:
This makes all integration tests to pass.

### DIFF
--- a/src/develop/pixelpipe_hb.c
+++ b/src/develop/pixelpipe_hb.c
@@ -353,7 +353,7 @@ void dt_dev_pixelpipe_synch_all(dt_dev_pixelpipe_t *pipe, dt_develop_t *dev)
   {
     dt_dev_pixelpipe_iop_t *piece = (dt_dev_pixelpipe_iop_t *)nodes->data;
     piece->hash = 0;
-    piece->enabled = (piece->module->default_enabled && piece->module->hide_enable_button);
+    piece->enabled = piece->module->default_enabled;
     dt_iop_commit_params(piece->module, piece->module->default_params, piece->module->default_blendop_params,
                          pipe, piece);
     nodes = g_list_next(nodes);


TurboGit:
@elstoc : Can you have a look at this one ? It fixes the integration tests which were really broken after a00f9d30.

Also, please make sure you run the integration tests when dealing with any develop part of code. Let me know if you need help running those tests. TIA.

elstoc:
Unfortunately the functionality introduced within #5449 does not work correctly with this fix added. The idea is that if I select a step prior to (for example) white balance in the history stack and compress history then (a) The module no longer shows in the stack (PASS); (b) The module shows as disabled on the right-hand-panel (PASS); and (c) The preview image shows the effect of disabling the module (FAIL).

So it looks like the UI thinks the module is disabled when in fact it is not.

Regarding the integration tests, I was vaguely aware of them but I didn't know how to run them. I see someone has added some documentation to the wiki now though (https://github.com/darktable-org/darktable/wiki/Dev-guide---darktable-testing-tools) so I'll consult that and let you know if I have any questions.

I'm not sure how to proceed with this one though given that it fails my testing.

TurboGit:
@elstoc : This new feature is certainly complex and may have some huge drawback (copy history ? in style ? support for module order ?). I don't think this should have been merge this fast... What about reverting this for now until we have a clear idea and more testing about all the points above ? Note that we are only at 3 months or so of next code freeze !

elstoc:
I'm ok with that if you think it's best. It was only a 'nice to have' from my point of view and I can live without it. It's just a pain to have to remember re-disable things when going back to earlier points in the stack. Please can you reopen the bug and the PR when you revert.

I'll probably need some support if it gets much more complex though.

TurboGit:
Ok, I'll revert and will reopen the bug but I cannot reopen a PR when merged.

TurboGit:
Revert done and issue reopened. Closing this PR.

PR #5985, opened 2020-08-16
imageop: replace module rename popup with GtkPopover

elstoc:
current method of showing the module rename popup is heavily dependent on the user's Window Manager responding in a consistent and well understood manner (which they often don't)

rename popup is therefore replaced with a GtkPopover widget which operates in a much more consistent fashion

Screenshot_2020-08-16_23-48-30

### DIFF
--- a/src/develop/imageop.c
+++ b/src/develop/imageop.c
@@ -1021,36 +1021,18 @@ static gboolean _rename_module_key_press(GtkWidget *entry, GdkEventKey *event, d
 
 static void _iop_gui_rename_module(dt_iop_module_t *module)
 {
-  const int bs = DT_PIXEL_APPLY_DPI(12);
-  gint px = 0, py = 0;
-
   dt_iop_gui_rename_module_t *d = (dt_iop_gui_rename_module_t *)calloc(1, sizeof(dt_iop_gui_rename_module_t));
   d->module = module;
-  GtkWidget *window = dt_ui_main_window(darktable.gui->ui);
 
   GList *childs = gtk_container_get_children(GTK_CONTAINER(module->header));
   GtkWidget *label = g_list_nth_data(childs, IOP_MODULE_LABEL);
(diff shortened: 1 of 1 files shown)


TurboGit:
Does work for me, I tried renaming the "coloriser" module on the right and the entry was full left over the image:

image

TurboGit:
I  kind of remember that the placement of the entry was not that easy to get properly on all platforms. A tricky one :)

elstoc:
Yes indeed. It works consistently for me. Hmm headscratcher.

elstoc:
With virtually no confidence I've pushed an alternative that 'also works' for me. Please can you try it.

elstoc:
Ok I've really found a much better alternative now (GtkPopover). Please can you retest as this solution should no longer be WM dependent.

dterrahe:
My version, in #6006, was slightly late, as I found when I went to rebase against master. :-)

There's no need for a pop-up window there, but I guess its a matter of taste whether that's preferable.

src/develop/imageop.c
```diff
@@ -1021,49 +1021,29 @@ static gboolean _rename_module_key_press(GtkWidget *entry, GdkEventKey *event, d
...
 static void _iop_gui_rename_module(dt_iop_module_t *module)
 {
-  const int bs = DT_PIXEL_APPLY_DPI(12);
-  gint px = 0, py = 0;
-
   dt_iop_gui_rename_module_t *d = (dt_iop_gui_rename_module_t *)calloc(1, sizeof(dt_iop_gui_rename_module_t));
   d->module = module;
-  GtkWidget *window = dt_ui_main_window(darktable.gui->ui);
 
   GList *childs = gtk_container_get_children(GTK_CONTAINER(module->header));
```
dterrahe, line 1027: childs _does_ need to be g_list_free'd (and might as well be renamed children then)

PR #5990, opened 2020-08-16
Fix scripts installer not running when locale is not english

wpferguson:
changed check from name, which is translated, to id, which is not translated.


TurboGit:
Ah probably my issue... Let me check...

PR #5996, opened 2020-08-17
filmic : improve graph view for user education

aurelienpierre:
Some users still have a hard time to understand what filmic does, is, or how it compares to base curve.

This PR aims at providing more graph views presenting information in different ways. Cycling through the views is achieved by clicking on the graph (left click : forward, right click : backward, double click : reset).

## Default view : look only

It's the old one but with legends added.
Screenshot_20200817_023947
The scene EVs in abscissa are meant to say that the bounds of the graph are actually scalable, which is a major difference compared to classic tone curves (bounded to 0-100% by design). What we do is actually scaling this look to whatever dynamic range we have in input.

## Look + mapping (linear)
Screenshot_20200817_023958
This is actually the base curve that the whole module, by stacking the log mapping first, then the look, then the power function. It is meant to be compared to the base curve module, so abscissa ends at 100% but the true position of the white point is still noted inside parentheses.

## Look + mapping (log)
Screenshot_20200817_024006
Same as the previous one, but rescaled with a log transform which is the same as the base curve log view. This helps looking at the lowlights.

## Dynamic range mapping
Screenshot_20200817_024016
This is the innovation here, inspired by the graphs I drew in my videos to explain the concept of tone mapping. It uses the zone system formalism to show how the dynamic range is remapped (1D to 1D, instead of the usual 2D curve which, if you ask me, adds more confusion than anything), especially which EV will be dilated and which ones will be compressed. It also tries to show that the grey point is our fulcrum/pivot here, so it is the only value that will travel through filmic unaffected.

### DIFF
--- a/src/iop/filmicrgb.c
+++ b/src/iop/filmicrgb.c
@@ -143,6 +143,16 @@ typedef struct dt_iop_filmic_rgb_spline_t
 } dt_iop_filmic_rgb_spline_t;
 
 
+typedef enum dt_iop_filmic_rgb_gui_mode_t
+{
+  DT_FILMIC_GUI_LOOK = 0,          // default GUI, showing only the contrast curve in a log/gamma space
+  DT_FILMIC_GUI_BASECURVE = 1,     // basecurve-like GUI, showing the contrast and brightness curves, in lin/lin space
+  DT_FILMIC_GUI_BASECURVE_LOG = 2, // same as previous, but log-scaled
+  DT_FILMIC_GUI_RANGES = 3,        // zone-system-like GUI, showing the range to range mapping
+  DT_FILMIC_GUI_LAST
+} dt_iop_filmic_rgb_gui_mode_t;
+
+
 typedef struct dt_iop_filmicrgb_params_t
 {
   float grey_point_source;  // $MIN: 0 $MAX: 100 $DEFAULT: 18.45 $DESCRIPTION: "middle grey luminance"
@@ -203,6 +213,30 @@ typedef struct dt_iop_filmicrgb_gui_data_t
   GtkDrawingArea *area;
   struct dt_iop_filmic_rgb_spline_t spline DT_ALIGNED_ARRAY;
   gint show_mask;
+  dt_iop_filmic_rgb_gui_mode_t gui_mode;
+    
+  // Cache Pango and Cairo stuff for the equalizer drawing
+  float line_height;
+  float sign_width;
+  float zero_width;
+  float graph_width;
+  float graph_height;
+  float gradient_left_limit;
+  float gradient_right_limit;
+  float gradient_top_limit;
+  float gradient_width;
+  float legend_top_limit;
+  float x_label;
+  int inset;
+  int inner_padding;
+
+  GtkAllocation allocation;
+  cairo_surface_t *cst;
+  cairo_t *cr;
+  PangoLayout *layout;
+  PangoRectangle ink;
+  PangoFontDescription *desc;
+  GtkStyleContext *context;
 } dt_iop_filmicrgb_gui_data_t;
 
 typedef struct dt_iop_filmicrgb_data_t
@@ -515,6 +549,15 @@ static inline float log_tonemapping_v1(const float x, const float grey, const fl
 static inline float log_tonemapping_v2(const float x, const float grey, const float black, const float dynamic_range)
 {
   return clamp_simd((log2f(x / grey) - black) / dynamic_range);
+}
+
+#ifdef _OPENMP
+#pragma omp declare simd uniform(grey, black, dynamic_range)
+#endif
+static inline float exp_tonemapping_v2(const float x, const float grey, const float black, const float dynamic_range)
+{
+  // inverse of log_tonemapping
+  return grey * exp2f(dynamic_range * x + black);
 }
 
 
@@ -2046,6 +2089,7 @@ void gui_update(dt_iop_module_t *self)
   dt_iop_color_picker_reset(self, TRUE);
 
   g->show_mask = FALSE;
+  g->gui_mode = DT_FILMIC_GUI_LOOK;
 
   self->color_picker_box[0] = self->color_picker_box[1] = .25f;
   self->color_picker_box[2] = self->color_picker_box[3] = .50f;
@@ -2144,6 +2188,15 @@ void gui_reset(dt_iop_module_t *self)
   dt_iop_color_picker_reset(self, TRUE);
 }
 
+#define LOGBASE 20.f
+
+static inline void dt_cairo_draw_arrow(cairo_t *cr, double origin_x, double origin_y, double destination_x, double destination_y)
+{
+  cairo_move_to(cr, origin_x, origin_y);
+  cairo_line_to(cr, destination_x, destination_y);
+  cairo_stroke(cr);
+}
+
 static gboolean dt_iop_tonecurve_draw(GtkWidget *widget, cairo_t *crf, gpointer user_data)
 {
   dt_iop_module_t *self = (dt_iop_module_t *)user_data;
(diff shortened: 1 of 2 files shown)


codingdave:
Hi @aurelienpierre 
I like the views very much and think they can remove confusion because different concepts are visualized and gaps in understanding can be closed with this PR.

> "white -> (985)"/"the true position of the white point"

I dont understand this completely and expect this to raise questions from the users as well.

> Cycling through the views is achieved by clicking on the graph (left click : forward, right click : backward, double click : reset).

For me the graph and dynamic range mapping are read only views and as a user I dont expect them to be clickable or worse - if I click on them I expect a different behavior than to rotate through the views. I don't know the gtk best practises but isnt there an expander or a dropdown menu that can be used to cycle through the items?

Also I don't think there is a view-A-logically-follows-view-B concept and such the users might sometimes want to go from e.g. "look +mapping (lin)" to "dynamic range mapping" without stopping at any of the other 2 views. Why not giving the user the explicit choice here?

Whenever the label/dropdown menu has changed there should be a different tooltip explaining the view. Is this possible with gtk?

aurelienpierre:
> > "white -> (985)"/"the true position of the white point"
> I dont understand this completely and expect this to raise questions from the users as well.

The value of the white point is 985% in this example. The graph ends at 100% on the right. The white point cannot be represented on the graph, so we simply display its value for information.

> For me the graph and dynamic range mapping are read only views and as a user I dont expect them to be clickable or worse - if I click on them I expect a different behavior than to rotate through the views. I don't know the gtk best practises but isnt there an expander or a dropdown menu that can be used to cycle through the items?

I kind of agree, the problem is adding another menu or element will take even more screen real estate. I already had to increase the vertical dimension of the graph to fit all the labels and legend, I know notebooks users always complain that modules are too high, I would prefer to avoid yet another GUI thing.

codingdave:
> The value of the white point is 985% in this example. The graph ends at 100% on the right. The white point cannot be represented on the graph, so we simply display its value for information.
Ah, see, the unit is missing (%) ;-)

aurelienpierre:
> Ah, see, the unit is missing (%) ;-)

Well, it's the same unit as the whole axis :-P

parafin:
> > Ah, see, the unit is missing (%) ;-)
> 
> Well, it's the same unit as the whole axis :-P

I would still add the unit, because I was confused too. Also maybe make it clear that graph continues to the right? Like adding an arrow or smth...

kofa73:
I thought I understood the mapping, but now I'm getting confused.
What does it mean that camera 985% is mapped to 100% in display? Does it mean that a value almost 10x above the camera white would appear as white on the screen, in other words, that the brightest point (100% camera) in the image will still be dark on the screen?
Also, I assume the 'camera' axis actually represents camera + adjustments in exposure module [plus any other transformations coming before filmic, including tone EQ] (= input of the mapping), right?

aurelienpierre:
> What does it mean that camera 985% is mapped to 100% in display? Does it mean that a value almost 10x above the camera white would appear as white on the screen, in other words, that the brightest point (100% camera) in the image will still be dark on the screen?

1. camera records between 0-100% of camera scale (by definition). Everything above 100% is clipped.
2. pipeline and display expect middle grey at 18% because it's the center of our visual dynamic range (sort of). But in generally, middle grey is not recorded by the camera at 18% of its range, but much lower. To compensate for that, we perform an exposure compensation (linear correction) so camera middle grey gets matched to 18% and camera peak intensity gets matched to 985%. It doesn't matter since nothing is clipped yet, we are still in the pipe, we play with numbers.
3. filmic shuts up and remap that display 985% camera to display 100% because that's what it was designed for.

> Also, I assume the 'camera' axis actually represents camera + adjustments in exposure module [plus any other transformations coming before filmic, including tone EQ] (= input of the mapping), right?

The camera axis is the camera space, aka values scaled as recorded. Then, of course, the whole stack of operations gets projected onto that space.

Scene-referred means that 100% means nothing anymore, since we don't clip. Where the display-referred workflow pivots around 100% and clip above, scene-referred pivots around 18% and let displays deal with highlights depending on their dynamic range.

ptilopteri:
@kofa73 
<What does it mean that camera 985% is mapped to 100% in display? Does it mean that a value almost 10x above the camera white would appear as white on the screen, in other words, that the brightest point (100% camera) in the image will still be dark on the screen?>

I may be wrong but I believe "985" refers to 98.5%

kofa73:
Thanks for the explanation. It makes sense (and matches my previous understanding and experience). However:

- I'm not sure other users are not going to get confused like I did. Maybe the 'camera' axis shouldn't be called 'camera', but rather 'scene'?
- Since '100%' kind of means/suggests 'full', 'complete' and 'maximum' intuitively (plus 'camera records between 0-100% of camera scale (by definition). Everything above 100% is clipped'), I'd think many users will think that clipping occurs on charts that refer to values 'above camera 100%' .
- The fact that anything above the user-selected white point is clipped (or is subject to highlight reconstruction), is easy to understand. I find the log views (EVs) here much more clear than (truncated) linear percents (especially when those percents don't really mean anything, are just a number relative to the '18%' value of 'mid-grey') [if I understand correctly].

@ptilopteri : if I understood Aurélien correctly, it means 985% (a brightness with a linear measure 985/18=54.7 times the value representing 'mid-grey'). It's not clipped; the curve will map it to 100% display brightness.

ptilopteri:
@kofa73 
tks, you have increased my limited understanding.

aurelienpierre:
> I'm not sure other users are not going to get confused like I did. Maybe the 'camera' axis shouldn't be called 'camera', but rather 'scene'?

The scene has no 100%, it is not bounded. 100% has to be relative to some bounded space, and so far, the custom has been choosen to map the bounds of the camera to the bounds of the display, without going back to scene-referred scaling.

If we want to have a view consistent with the base curve, for comparison, we have to have a graph with a 100% bound. BTW, that camera axis is exactly the same as the base curve. So, after decades of manipulating axes with no labels, and numbers with no units, it's no surprise that users don't get a sense of what is going on.

> Since '100%' kind of means/suggests 'full', 'complete' and 'maximum' intuitively (plus 'camera records between 0-100% of camera scale (by definition). Everything above 100% is clipped'), I'd think many users will think that clipping occurs on charts that refer to values 'above camera 100%' .

But that's exactly the point. "See that 100% ? in display-referred, it's clipped. In scene-referred, we don't care, it stays". The need is to show that this 100% has lost its meaning with a view that most users know and think they understand.

kofa73:
I think if the summary of what you've explained here makes it into the manual, and the screens have a direct link to the corresponding link in the manual (and/or one of your videos), you'll achieve your purpose. Thanks for all your work of coding and educating!

juhavlintula:
I think the last view is brilliant. It gives a good understanding how your input tonal range is mapped to output.

But I have also difficulties in understanding the second and the third view as the x-axis is cut at 100%. Just as an idea, what if  the graph is drawn to to full extent, but the >100% would be darker/more dim so that the part up to 100% is kind of "highlighted".

jakubfi:
I would love the view button to react also on mouse wheel if possible.

aurelienpierre:
Some cosmetic changes:
Screenshot_20200817_221152

Screenshot_20200817_221207

> But I have also difficulties in understanding the second and the third view as the x-axis is cut at 100%. Just as an idea, what if the graph is drawn to to full extent, but the >100% would be darker/more dim so that the part up to 100% is kind of "highlighted".

If I do that, the scaling will cram black, latitude and middle grey to the left, to the point where you won't see anything but a simple log curve. Just imagine the full graph above, condensed 10 times on the left.

 > I would love the view button to react also on mouse wheel if possible.

React how ? To cycle through views ?

jakubfi:
> React how ? To cycle through views ?

Correct.

juhavlintula:
I didn't think to the end. I didn't think that you cannot extend the graph to the right and make it wider.

cigno5:
I love that feature! Would be great!...

If I may give a suggestion:
> > For me the graph and dynamic range mapping are read only views and as a user I dont expect them to be clickable or worse - if I click on them I expect a different behavior than to rotate through the views. I don't know the gtk best practises but isnt there an expander or a dropdown menu that can be used to cycle through the items?
> 
> I kind of agree, the problem is adding another menu or element will take even more screen real estate. I already had to increase the vertical dimension of the graph to fit all the labels and legend, I know notebooks users always complain that modules are too high, I would prefer to avoid yet another GUI thing.

What about using the same trick used on the histogram? On the mouse over a little top right semi-transparent bar appears with the 4 view mode. You can easily switch the one you like the most and it doesn't take more space. Moreover is still consistent with another element in the UI!

aurelienpierre:
> What about using the same trick used on the histogram? On the mouse over a little top right semi-transparent bar appears with the 4 view mode. You can easily switch the one you like the most and it doesn't take more space. Moreover is still consistent with another element in the UI!

That's possible, even though it equates creating a whole Gtk widget from Cairo drawing… again. Much more work.

ostueker:
> > For me the graph and dynamic range mapping are read only views and as a user I dont expect them to be clickable or worse - if I click on them I expect a different behavior than to rotate through the views. I don't know the gtk best practises but isnt there an expander or a dropdown menu that can be used to cycle through the items?
> 
> I kind of agree, the problem is adding another menu or element will take even more screen real estate. I already had to increase the vertical dimension of the graph to fit all the labels and legend, I know notebooks users always complain that modules are too high, I would prefer to avoid yet another GUI thing.

How about having a small icon (maybe based on  ) right next to the name of the view?

This icon would indicate that one can cycle through the options. The name itself can still remain clickable and a tool-tip could also indicate that one can cycle through different views.   
From @aurelienpierre's first set of screenshots it seems like there is sufficient space on the right side of the name.

kofa73:
The orange dot represents mid-grey, right? (18%, or 0 EV etc.) And it's always mapped to mid-grey, 18% display luminance.
On the charts, why does it move up and down, when black and white point are adjusted? Why doesn't it always appear at the same position (e.g. at vertical middle)? (I understands it moves left and right as the black and white points are adjusted, as the width of the chart is fixed in pixels, but the range it represents changes, so mid-grey appears always where the white and black relative offsets dictate).

rkowalke:
> The orange dot represents mid-grey, right? (18%, or 0 EV etc.) And it's always mapped to mid-grey, 18% display luminance.
> On the charts, why does it move up and down, when black and white point are adjusted? Why doesn't it always appear at the same position (e.g. at vertical middle)? 

The position of the orange dot is (by default) calculated such that it stays at the diagonal. So changing the dynamic range will result in movements of both horizontal and vertical coordinates.

src/iop/filmicrgb.c
```diff
@@ -2144,132 +2188,775 @@ void gui_reset(dt_iop_module_t *self)
...
+      pango_font_description_set_style(g->desc, backup);
+      pango_layout_set_font_description(g->layout, g->desc);
     }
 
-    cairo_line_to(cr, x * width, height * (1.0 - y));
-    cairo_stroke(cr);
-    cairo_move_to(cr, x * width, height * (1.0 - y));
+    // mark the y axis legend
+    set_color(g->cr, darktable.bauhaus->graph_fg);
+    snprintf(text, sizeof(text), _("%% display"));
```
TurboGit, line 2624: strncpy() lighter than snprintf in such case.

src/iop/filmicrgb.c
```diff
@@ -2144,132 +2188,775 @@ void gui_reset(dt_iop_module_t *self)
...
+    pango_layout_get_pixel_extents(g->layout, &g->ink, NULL);
+    cairo_move_to(g->cr, -2. * g->inset - g->zero_width - g->ink.x,
+                        -g->line_height -g->inset - 0.5 * g->ink.height - g->ink.y);
+    pango_cairo_show_layout(g->cr, g->layout);
+    cairo_stroke(g->cr);
+
+    // mark the x axis legend
+    set_color(g->cr, darktable.bauhaus->graph_fg);
+    if(g->gui_mode == DT_FILMIC_GUI_LOOK) 
+      snprintf(text, sizeof(text), _("EV scene"));
```
TurboGit, line 2635: strncpy()

src/iop/filmicrgb.c
```diff
@@ -2144,132 +2188,775 @@ void gui_reset(dt_iop_module_t *self)
...
+    else if(g->gui_mode == DT_FILMIC_GUI_BASECURVE || g->gui_mode == DT_FILMIC_GUI_BASECURVE_LOG) 
+      snprintf(text, sizeof(text), _("%% camera"));
```
TurboGit, line 2637: strncpy()

src/iop/filmicrgb.c
```diff
@@ -2144,132 +2188,775 @@ void gui_reset(dt_iop_module_t *self)
...
+    // for uint10 output, white should be set to 400%, so anything above 100% increases DR
+    const float display_DR = 8.f + log2f(p->white_point_target / 100.f);
+
+    // FIXME : if darktable becomes HDR-10bits compatible (for output), this needs to be updated
+    const float y_display = g->allocation.height / 3.f;
+    const float y_scene = 2. * g->allocation.height / 3.f;
+
+    // labels
+    set_color(g->cr, darktable.bauhaus->graph_fg);
+    snprintf(text, sizeof(text), _("display"));
```
TurboGit, line 2662: strncpy()

src/iop/filmicrgb.c
```diff
@@ -2144,132 +2188,775 @@ void gui_reset(dt_iop_module_t *self)
...
+    pango_layout_set_text(g->layout, text, -1);
+    pango_layout_get_pixel_extents(g->layout, &g->ink, NULL);
+    cairo_move_to(g->cr, 0., y_display - 0.5 * g->ink.height - g->ink.y);
+    pango_cairo_show_layout(g->cr, g->layout);
+    cairo_stroke(g->cr);
+    const float display_label_width = g->ink.width;
+
+    set_color(g->cr, darktable.bauhaus->graph_fg);
+    snprintf(text, sizeof(text), _("scene"));
```
TurboGit, line 2671: strncpy()

src/iop/filmicrgb.c
```diff
@@ -2144,132 +2188,775 @@ void gui_reset(dt_iop_module_t *self)
...
+    const float column_left = fmaxf(display_label_width, scene_label_width) + g->inset;
+    const float column_right = g->allocation.width - column_left - darktable.bauhaus->quad_width;
+
+    // compute dynamic ranges left and right to middle grey
+    const float display_HL_EV = -log2f(p->grey_point_target / p->white_point_target);       // compared to white EV
+    const float display_LL_EV = display_DR - display_HL_EV;  // compared to black EV
+    const float scene_HL_EV = p->white_point_source;                        // compared to white EV
+    const float scene_LL_EV = -p->black_point_source;                       // compared to black EV
+
+    fprintf(stdout, "%f, %f \n", display_HL_EV, display_LL_EV);
```
TurboGit, line 2694: debug code ?

src/iop/filmicrgb.c
```diff
@@ -2144,132 +2188,775 @@ void gui_reset(dt_iop_module_t *self)
...
+    const float display_top = y_display - g->line_height / 2;
+    const float display_bottom = display_top + g->line_height;
+
+    const float scene_top = y_scene - g->line_height / 2;
+    const float scene_bottom = scene_top + g->line_height;
+
+    // show EV zones for display - zones are aligned on 0% and 100%
+    cairo_set_line_width(g->cr, DT_PIXEL_APPLY_DPI(1.));
+
+    fprintf(stdout, "%i \n", (int)display_DR);
```
TurboGit, line 2721: debug code ?

src/iop/filmicrgb.c
```diff
@@ -2144,132 +2188,775 @@ void gui_reset(dt_iop_module_t *self)
...
+    snprintf(text, sizeof(text), "%+.1f", p->white_point_source);
+    pango_layout_set_text(g->layout, text, -1);
+    pango_layout_get_pixel_extents(g->layout, &g->ink, NULL);
+    cairo_move_to(g->cr, scene_white_x - 0.5 * g->ink.width - g->ink.x,
+                         scene_bottom + 2. * g->inset + 0. * g->ink.height + g->ink.y);
+    pango_cairo_show_layout(g->cr, g->layout);
+    cairo_stroke(g->cr);
+
+    // axis legend
+    snprintf(text, sizeof(text), _("(EV)"));
```
TurboGit, line 2860: strncpy()

src/iop/filmicrgb.c
```diff
@@ -2144,132 +2188,775 @@ void gui_reset(dt_iop_module_t *self)
...
+    snprintf(text, sizeof(text), "%.0f", p->white_point_target);
+    pango_layout_set_text(g->layout, text, -1);
+    pango_layout_get_pixel_extents(g->layout, &g->ink, NULL);
+    cairo_move_to(g->cr, display_white_x - 0.5 * g->ink.width - g->ink.x,
+                         display_top - 4. * g->inset - g->ink.height - g->ink.y);
+    pango_cairo_show_layout(g->cr, g->layout);
+    cairo_stroke(g->cr);
+
+    // axis legend
+    snprintf(text, sizeof(text), _("(%%)"));
```
TurboGit, line 2896: strncpy()

src/iop/filmicrgb.c
```diff
@@ -2144,132 +2188,775 @@ void gui_reset(dt_iop_module_t *self)
...
+
+  char text[256];
+
+  // reduce a bit the font size
+  const gint font_size = pango_font_description_get_size(g->desc);
+  pango_font_description_set_size(g->desc, 0.95 * font_size);
+  pango_layout_set_font_description(g->layout, g->desc);
+
+  // Get the text line height for spacing
+  snprintf(text, sizeof(text), "X");
```
TurboGit, line 2247: strncpy();

src/iop/filmicrgb.c
```diff
@@ -2144,132 +2188,775 @@ void gui_reset(dt_iop_module_t *self)
...
+  pango_layout_set_text(g->layout, text, -1);
+  pango_layout_get_pixel_extents(g->layout, &g->ink, NULL);
+  g->line_height = g->ink.height;
+
+  // Get the width of a minus sign for legend labels spacing
+  snprintf(text, sizeof(text), "-");
```
TurboGit, line 2253: strncpy();

src/iop/filmicrgb.c
```diff
@@ -2144,132 +2188,775 @@ void gui_reset(dt_iop_module_t *self)
...
+  pango_layout_set_text(g->layout, text, -1);
+  pango_layout_get_pixel_extents(g->layout, &g->ink, NULL);
+  g->sign_width = g->ink.width / 2.0;
+
+  // Get the width of a zero for legend labels spacing
+  snprintf(text, sizeof(text), "0");
```
TurboGit, line 2259: strncpy();

src/iop/filmicrgb.c
```diff
@@ -2144,132 +2188,775 @@ void gui_reset(dt_iop_module_t *self)
...
+  // set the graph as the origin of the coordinates
+  cairo_translate(g->cr, margin_left, margin_top);
+
+  cairo_set_line_cap(g->cr, CAIRO_LINE_CAP_ROUND);
+
+  // write the graph legend at GUI default size
+  pango_font_description_set_size(g->desc, font_size);
+  pango_layout_set_font_description(g->layout, g->desc);
+  if(g->gui_mode == DT_FILMIC_GUI_LOOK) 
+    snprintf(text, sizeof(text), _("look only"));
```
TurboGit, line 2296: strncpy();

Here and three more instances just below.

TurboGit (commented):
Some random comments after a quick review.

aurelienpierre:
> On the charts, why does it move up and down, when black and white point are adjusted? Why doesn't it always appear at the same position (e.g. at vertical middle)? (I understands it moves left and right as the black and white points are adjusted, as the width of the chart is fixed in pixels, but the range it represents changes, so mid-grey appears always where the white and black relative offsets dictate).

It moves along the identity line only in look view (which is a log-gamma graph) if you set "auto-adjust hardness" in options. The "auto-adjust hardness" is only a GUI feature that compensate the hardness so it stays on the identity line, because that gives an easier S curve to control and prevent clipping of the curve. That's also because grey is forever our pivot, so it should exit at the same value it entered.

@TurboGit Noted, thanks. On @elstoc's request, I made the views saved in darktablerc for next picture, added a label-free variant to spare GUI real-estate, and I'm working on adding buttons to cycle between view (same idea as histogram). I should be done tomorrow.

rabauke:
I wonder if the labels "look + mappling (lin)" and "look + mappling (log)" should be replaced by "base curve (lin)" and  "base curve (log)". In my view, the current labels require a very deep understanding how filmic rgb works. Furthermore, the current labels may be wrongly understood as if two things were shown, the look and the mapping. (Aren't there, two graphs shown?)

Therefore, I would like to suggest to change the mentioned labels as indicated and possibly also to remove the saturation graph from the two "look + mappling" views. The saturation graph lies mainly outside the axis ranges, i.e., it is not vivisible, if "middle tones saturation" is > 0%.

PR #6012, opened 2020-08-18
draw accurate boxes for color picker live samples

Mark-64:
There is a graphic glitch in the drawing of color picker live sample boxes with current master.
It is fixed with this PR
before: the left side of the box is a double line
Screenshot from 2020-08-18 15-38-03

after
Screenshot from 2020-08-18 15-38-50

Also, the box is drawn immediately when pressing button "add", while in master it is drawn only after clicking somewhere else in the central image, giving the impression of no feedback to the user.


TurboGit:
Works for me, thanks!

PR #6013, opened 2020-08-18
introduce a way to have single button to discard history

### LINKED ISSUE #5972: FR: darkroom - single button to discard history
Currently in the darkroom it is possible to discard history by selecting the 'original' entry and compressing the history stack.

It would be nice to have a one-button way to do this (it's perhaps not always obvious to new users).

I would suggest either a separate button immediately below 'compress history stack' or to use the 'reset parameters' button on the module header. I'd say the latter seems more consistent with the rest of the UI and has the advantage of not taking any extra screen space. Also a button beneath 'compress' could be pressed by accident.

Also now that selecting 'original image' has the effect of showing the image without white balance (due to #5449) it might be less jarring than the previous method.

johnny-bit:
fixes #5972 by using module reset button in history module

tested and works fine, also supports undo and all that fun :)

### DIFF
--- a/src/libs/history.c
+++ b/src/libs/history.c
@@ -838,7 +838,7 @@ void gui_reset(dt_lib_module_t *self)
 
     GtkWidget *dialog = gtk_message_dialog_new(
         GTK_WINDOW(win), GTK_DIALOG_DESTROY_WITH_PARENT, GTK_MESSAGE_QUESTION, GTK_BUTTONS_YES_NO,
-        _("do you really want to reset history of current image?"));
+        _("do you really want to clear history of current image?"));
 #ifdef GDK_WINDOWING_QUARTZ
     dt_osx_disallow_fullscreen(dialog);
 #endif
@@ -855,12 +855,12 @@ void gui_reset(dt_lib_module_t *self)
                           dt_ioppr_iop_order_copy_deep(darktable.develop->iop_order_list));
 
 
-  dt_history_delete_on_image_ext(imgid, FALSE);
+    dt_history_delete_on_image_ext(imgid, FALSE);
 
-  dt_control_signal_raise(darktable.signals, DT_SIGNAL_DEVELOP_HISTORY_CHANGE);
-  dt_dev_modulegroups_set(darktable.develop, dt_dev_modulegroups_get(darktable.develop));
+    dt_control_signal_raise(darktable.signals, DT_SIGNAL_DEVELOP_HISTORY_CHANGE);
+    dt_dev_modulegroups_set(darktable.develop, dt_dev_modulegroups_get(darktable.develop));
 
-  dt_control_queue_redraw_center();
+    dt_control_queue_redraw_center();
   }
 }
 
--- a/src/libs/history.c
+++ b/src/libs/history.c
@@ -825,6 +825,45 @@ static void _lib_history_create_style_button_clicked_callback(GtkWidget *widget,
   }
 }
 
+void gui_reset(dt_lib_module_t *self)
+{
+  const int32_t imgid = darktable.develop->image_storage.id;
+  if(!imgid) return;
+
+  gint res = GTK_RESPONSE_YES;
+
+  if(dt_conf_get_bool("ask_before_delete"))
+  {
+    const GtkWidget *win = dt_ui_main_window(darktable.gui->ui);
+
+    GtkWidget *dialog = gtk_message_dialog_new(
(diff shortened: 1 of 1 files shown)


src/libs/history.c
```diff
@@ -825,6 +825,45 @@ static void _lib_history_create_style_button_clicked_callback(GtkWidget *widget,
...
+
+  gint res = GTK_RESPONSE_YES;
+
+  if(dt_conf_get_bool("ask_before_delete"))
+  {
+    const GtkWidget *win = dt_ui_main_window(darktable.gui->ui);
+
+    GtkWidget *dialog = gtk_message_dialog_new(
+        GTK_WINDOW(win), GTK_DIALOG_DESTROY_WITH_PARENT, GTK_MESSAGE_QUESTION, GTK_BUTTONS_YES_NO,
+        _("do you really want to reset history of current image?"));
```
elstoc, line 841: ``reset->clear`` to match the lighttable confirmation message

elstoc:
Works for me apart from a minor change to the message for consistency. Thanks @johnny-bit

johnny-bit:
The undo here is a bit weird thingie worth investingating once we iron out other more pressing bugs. somehow for me undo "works" by pressing ctrl-z twice. same for compress history.

elstoc:
Ah yes indeed: undo twice does work

TurboGit:
A little glitch with the undo:
1. start dt
2. select image (with some non default history) and go to darkroom
3. click on reset, say yes
4. Do undo : Ctrl+Z no matter how many time the history is lost

If between set 2 and 3 you select another place in history and then the last history item before reset it works. This means that we are probably missing some init somewhere.

Note that this is not related to this PR as the very same issue can be observed when using the compress button. Let me know if you're ok to fix this and if you prefer this PR merged first and the fix done in a separate PR.

elstoc:
@TurboGit I want to add another PR to separate the preference for 'prompt when discarding' and 'prompt when deleting from disk' into two (it's really odd that two such different actions are covered by the same pref). So I'd prefer to merge this and raise a separate issue/PR for the undo.

TurboGit:
Understood ! Thanks.

src/libs/history.c
```diff
@@ -825,6 +825,45 @@ static void _lib_history_create_style_button_clicked_callback(GtkWidget *widget,
...
+    const GtkWidget *win = dt_ui_main_window(darktable.gui->ui);
+
+    GtkWidget *dialog = gtk_message_dialog_new(
+        GTK_WINDOW(win), GTK_DIALOG_DESTROY_WITH_PARENT, GTK_MESSAGE_QUESTION, GTK_BUTTONS_YES_NO,
+        _("do you really want to clear history of current image?"));
+#ifdef GDK_WINDOWING_QUARTZ
+    dt_osx_disallow_fullscreen(dialog);
+#endif
+
+    gtk_window_set_title(GTK_WINDOW(dialog), _("delete images' history?"));
```
TurboGit, line 846: Just saw this now... Shouldn't this be :
`"delete image's history?"`

As we are speaking of a single image.

johnny-bit: darn. i missed that one. Not worth doing PR for such small change, so if you can please change to `"delete image's history"` :) I can do PR for couple other spelling mistakes that I did, so - whatever you prefer!

elstoc: I just submitted a PR to separate the prefs. I'll include it in that.

elstoc: #6016

TurboGit: Fixed! Sorry didn't see your message @elstoc !

TurboGit: And no worry, Git will handle that without trouble!

elstoc: That's ok it's in my PR as well so there aren't any conflicts

MStraeten:
introduced a build error on osx:
```
Scanning dependencies of target history
[ 50%] Building C object src/libs/CMakeFiles/history.dir/history.c.o
/Users/martinstraeten/src/darktable_debug/src/libs/history.c:843:5: error: implicit declaration of function
      'dt_osx_disallow_fullscreen' is invalid in C99 [-Werror,-Wimplicit-function-declaration]
    dt_osx_disallow_fullscreen(dialog);
    ^
```

johnny-bit:
Plz check if #6017 (forgot osx-specific include)

TurboGit:
@johnny-bit : history undo/redo fixed in #6018.

PR #6016, opened 2020-08-19
preferences: separate options for delete and discard

elstoc:
create separate options for delete/trash and discard history

minor code changes to use consistent terminology (delete -> discard)


TurboGit (approved):
Looks ok to me. Thanks!

PR #6017, opened 2020-08-19
fix osx compilation error caused by #6013

johnny-bit:
This should fix the osx compilation issue from #6013 raised by @MStraeten :)


MStraeten:
Build was successful. Thanks

TurboGit (approved):
Thanks !

PR #6018, opened 2020-08-19
fix history undo/redo in darkroom

TurboGit:


elstoc:
Works fine for the history stack 'reset parameters' button but the other way of doing it (click original and compress) still seems to need an extra 'undo'. I think somehow it's splitting the rebuilding of a 'default' stack into (1) Build the default_enabled modules; then (2) Add any modules enabled by workflow/presets.

I don't know if this is expected but I suspect not.

TurboGit:
> but the other way of doing it (click original and compress) still seems to need an extra 'undo'

In my latest version it needs only 2 undo: one for restoring history and one for replacing the history end.

elstoc:
Yes that looks better. All good for me.

TurboGit:
Thanks for testing !

TurboGit:
Manually rebased and merged.

PR #6021, opened 2020-08-19
rename iop modules inline

dterrahe:
to rename a multi-instance module, add a GtkEntry to the module header in the panel and resize it to its current content dynamically.

no longer requires a popup or popover window.


TurboGit:
No string feeling ! The pop-over works for me and this one too. Since this one seems to please more people and closer to previous implementation... let's merge this. Thanks all for testing, trying and discussing this GUI issue.

Issue #5900, opened 2020-08-09
Running the darktable-cli without locking the database


ziadloo:
**Is your feature request related to a problem? Please describe.**
I'm interested in running the `darktable-cli` program multiple times and concurrently in order to increase the throughput of the export process. It's just that once run, it locks the database and there's no way to run more than one instance of the program. This is the error thrown:

```
[init] the database lock file contains a pid that seems to be alive in your system: 159599
[init] database is locked, probably another process is already using it
ERROR: can't acquire database lock, aborting.
```

**Describe the solution you'd like**
I don't know what is written in that database but it would be nice to have an option in the `darktable-cli` program to open the database in read-only mode so multiple instances of the program could be run.

**Alternatives**
Or if possible, running the `darktable-cli` program without accessing the database at all.
One completely different approach could be adding the option to darktable application to work on multiple images while exporting (in order to improve the export performance).

**Additional context**
In general, the export process is too slow for me. I'm looking for a way to improve it.

parafin:
Use --apply-custom-presets false option as documented in the man page.

ziadloo:
Thanks @parafin but that parameter has no effect in this matter. I'm still getting the error:

```
[init] the database lock file contains a pid that seems to be alive in your system: 184504
[init] database is locked, probably another process is already using it
ERROR: can't acquire database lock, aborting.
```

Even though it seems this option should have worked but it does not. Not sure if this is a bug or that option is not intended for this purpose.

parafin:
Most likely you're using too old darktable version (which you never mentioned, even though it's a required information for bug report). Try with 3.2.1 which is to be released today.

ziadloo:
Thanks for the info, @parafin . I'm using Darktable 3.0.2. I'll update this ticket once I get to test 3.2.1. Thanks again.

ziadloo:
@parafin I managed to install the latest version:

```
$ darktable --version
this is darktable 3.2.1
copyright (c) 2009-2020 johannes hanika
darktable-dev@lists.darktable.org

compile options:
  bit depth is 64 bit
  normal build
  SSE2 optimized codepath enabled
  OpenMP support enabled
  OpenCL support enabled
  Lua support enabled, API version 6.0.0
  Colord support enabled
  gPhoto2 support enabled
  GraphicsMagick support enabled
  ImageMagick support disabled
  OpenEXR support enabled
```

And I'm calling darktable-cli like this twice and at the same time:

```
[...]

PR #6024, opened 2020-08-19
add db paths to sql debugging output

johnny-bit:
This is code I used to track and try to replicate issue in #5900 to no avail. Might be useful in very rare situations.


TurboGit (approved):
Looks good to me indeed. Thanks.

PR #6058, opened 2020-08-22
histogram: clear scopes when leave live view

dtorop:
Behavior up until now has been that if:

1) enter tether view with a new session (no image selected)
2) enter live view
3) leave live view

The live view image disappears in the center view, but the histogram from that live view image remains.

This PR blacks out the histogram as well in this case.

Also, it removes an obsolete comment about histograms not matching.


TurboGit:
Works for me! Thanks.

PR #6066, opened 2020-08-23
make "work profile" work in tether view histogram

dtorop:
Previously if the histogram profile was "work profile" the data was displayed in the export profile.

This fix also led to realizing that in tether view the code can just export the image straight to the histogram profile. This saves a colorspace in conversion in most cases, and always saves some unneeded allocation/copying.

To make all this work I had to add a new routine `dt_colorspaces_get_work_profile()` to colorspace.c and add a bunch of const's to make the colorspaces and iop_profile code agree with each other.

This is a follow-up to #5714, to fix loose ends.

### DIFF
--- a/src/common/colorspaces.c
+++ b/src/common/colorspaces.c
@@ -825,6 +825,56 @@ static cmsHPROFILE dt_colorspaces_create_linear_infrared_profile(void)
   return profile;
 }
 
+const dt_colorspaces_color_profile_t *dt_colorspaces_get_work_profile(const int imgid)
+{
+  // find the colorin module -- the pointer stays valid until darktable shuts down
+  static dt_iop_module_so_t *colorin = NULL;
+  if(colorin == NULL)
+  {
+    GList *modules = g_list_first(darktable.iop);
+    while(modules)
+    {
+      dt_iop_module_so_t *module = (dt_iop_module_so_t *)(modules->data);
+      if(!strcmp(module->op, "colorin"))
+      {
+        colorin = module;
+        break;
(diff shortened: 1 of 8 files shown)


dtorop:
This should now work. I left out a conditional due to over-using git rebase to make legible commits. I rebased again to put the conditional back in and fix a few other careless bugs.

TurboGit (approved):
Nice use of introspection! Works for me, thanks.

TurboGit:
@dtorop : I have just fixed leaving the live-view mode for my Nikin D850 camera, can you double check all is good on your side?

dtorop:
> I have just fixed leaving the live-view mode for my Nikin D850 camera, can you double check all is good on your side?

@TurboGit : Yes, after 1842a451ae36a41ed08acb5d51dc80dbbc9bb086 not only do things still work, but you have fixed the behavior of the Canon 5D Mark III which I'm using for testing. Previously the mirror wouldn't go back down until after darktable shutdown. Now it goes back down when leaving live view mode.

dtorop:
> Nice use of introspection!

We've got Tobias Ellinghaus to credit for it: 52ce8f83aa3798293339082fd547079e7bf088fe, then c0f84c27c475fdfbfde606f498a2f3cb83758be4. I just cribbed the code from dt_colorspaces_get_output_profile() with slight modification.

> Works for me, thanks.

Thank you for reviewing/merging. I think I see how to make "work profile" work -- in some sense -- for live view as well. A small detail, but I'll try to get that done.

PR #6070, opened 2020-08-23
theme module name edit entry background color

dterrahe:
Make new inline editing of (multi-instance) module names (introduced in #6021) more obvious with a different background color.

image

@Nilvus I've named the entry widget #iop-panel-label so that it picks up the same padding etc settings of the module name itself, but this also, I assume, blocks
```
entry:active
{
  color: @field_active_fg;
  background-color: @field_active_bg;
}
```
If there is a better/more correct/consistent approach to doing this, I'm happy to amend this PR. However it currently achieves the hoped-for look in all three official themes.

### DIFF
--- a/data/themes/darktable.css
+++ b/data/themes/darktable.css
@@ -598,6 +598,11 @@ overshoot.right
   padding: 0.5em 0.5em 0.5em 0.5em;
 }
 
+#module-header entry 
+{
+  background-color: @field_active_bg;
+}
+
 /* Labels in modules */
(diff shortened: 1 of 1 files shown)


TurboGit:
Looks good, waiting for @Nilvus OK for this one.

Nilvus (approved):
Good catch and much better like this. And good place on CSS file. Thanks!

TurboGit:
Thanks to both of you!

PR #6071, opened 2020-08-23
culling-preview zoom fixes

### LINKED ISSUE #5941: [3.2.1] Lighttable culling no longer retains zoom when advancing to next image
Under 3.1.x in culling if you zoomed in on a pair of images (fixed zoom) it would match the zoom on both. If you then scrolled to compare another pair of images then the zoom level (and offset of zoom) was retained. I found this particularly useful if I was interested in looking at a burst of 5+ images and comparing the detail in a small area (e.g. with animals in motion to spot where focus on the eye was good out of a sequence).

In 3.2.1 if I have a pair showing and zoom in then it zooms in the pair, but if I then scroll on to the next then although it retains the zoom on the one out of the pair still showing, on the newly shown one the zoom is reset to "fit".

I suspect this is a bug rather than intended!

(I'm running under Mint linux using the deb from http://download.opensuse.org/repositories/graphics:/darktable/xUbuntu_20.04/ )

### LINKED ISSUE #5986: Lighttable - Zoomlevel in overlay while culling is incorrect.
**Describe the bug**

When using ctrl+scroll to zoom in to an image in the culling mode of lighttable, the zoom level in the overlay on top of the image is wrong. The percentage is totally unrelated to, for example, the 100% when middle clicking in the darkroom view.

**To Reproduce**

ctrl+scroll an image while culling, look at zoom percentage in the overlay.

**Expected behavior**

Match the percentage with the actual zoom level of the image.

 - Darktable Version: 3.2.1
 - OS: Linux Mint 19.4

AlicVB:
Changes summary : 
- culling and preview now keep the zoom values for new images when showing next-previous image. Note that zoom values are reset when exit-reentering this modes. fix #5941 
- in culling and preview the 100% zoom was completely off (a stupid leftover during the rewrite) fix #5986

### DIFF
--- a/src/dtgtk/culling.c
+++ b/src/dtgtk/culling.c
@@ -869,6 +869,18 @@ void dt_culling_init(dt_culling_t *table, int offset)
   table->navigate_inside_selection = FALSE;
   table->selection_sync = FALSE;
 
+  // reset remaining zooming values if any
+  GList *l = table->list;
+  while(l)
+  {
+    dt_thumbnail_t *thumb = (dt_thumbnail_t *)l->data;
+    thumb->zoom = 1.0f;
+    thumb->zoomx = 0.0;
+    thumb->zoomy = 0.0;
+    thumb->img_surf_dirty = TRUE;
+    l = g_list_next(l);
+  }
+
   const gboolean culling_dynamic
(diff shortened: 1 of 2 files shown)


TurboGit:
Works for me, thanks!

PR #6076, opened 2020-08-23
filmic : fix silent bug, optimize defaults

aurelienpierre:
* fix a silent bug in the spline initialization : the output power was not applied on white and black display, which didn't show since users were advised to let white to 100% and black to 0% (so 0^x = 0, and 1^x = 1), but triggers unintuitive behaviour otherwise,
* based on the new graph UI, optimize the default parameters to maximize use of output dynamic range and take 8 bits sRGB quantization error as display black value (0.016 %).

### DIFF
--- a/src/iop/filmicrgb.c
+++ b/src/iop/filmicrgb.c
@@ -19,25 +19,25 @@
 #include "config.h"
 #endif
 #include "bauhaus/bauhaus.h"
-#include "common/image.h"
-#include "common/iop_profile.h"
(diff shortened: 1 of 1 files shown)


TurboGit:
There is a lot of reformatting making it hard to review the essential part. This looks ok to me finally but it was painful to review and I may have missed something... Anyway merging. Thanks.

aurelienpierre:
Ah, sorry, I switched IDE and this one is doing clever stuff in the background.

PR #6079, opened 2020-08-23
export: scaling improvement

esq4:
In addition to specifying the exact max.size for export, adds an alternative scaling factor.

Parameters are mutually exclusive. When one changes, the other is cleared and not counted in the export.

In addition, each parameter can be reset by clicking the MMB.

### DIFF
--- a/src/common/imageio_module.c
+++ b/src/common/imageio_module.c
@@ -437,6 +437,47 @@ void dt_imageio_insert_storage(dt_imageio_module_storage_t *storage)
   DT_DEBUG_CONTROL_SIGNAL_RAISE(darktable.signals, DT_SIGNAL_IMAGEIO_STORAGE_CHANGE);
 }
 
+gchar *dt_imageio_resizing_factor_get_and_parsing(double *num, double *denum)
+{
+  double _num, _denum;
+  gchar *scale_str = dt_conf_get_string("plugins/lighttable/export/resizing_factor");
+
+  char sep[4] = "";
+  snprintf( sep, 4, "%g", (double) 3/2);
+  int i = -1;
+  while(scale_str[++i])
+  {
+      if ((scale_str[i] == '.') || (scale_str[i] == ',')) scale_str[i] = sep[1];
+  }
+
+  gchar *pdiv = strchr(scale_str, '/');
+
+  if (pdiv == NULL)
+  {
+    _num = atof(scale_str);
+    _denum = 1;
+  }
+  else if (pdiv-scale_str == 0)
+  {
+    _num = 1;
+    _denum = atof(pdiv + 1);
+}
+  else
+{
+    _num = atof(scale_str);
+    _denum = atof(pdiv+1);
+  }
+
+  if (_num == 0.0) _num = 1.0;
+  if (_denum == 0.0) _denum = 1.0;
+
+  *num = _num;
+  *denum = _denum;
+
+  dt_conf_set_string("plugins/lighttable/export/resizing_factor", scale_str);
+  return scale_str;
+}
+
 // modelines: These editor modelines have been set for all relevant files by tools/update_modelines.sh
 // vim: shiftwidth=2 expandtab tabstop=2 cindent
 // kate: tab-indents: off; indent-width 2; replace-tabs on; indent-mode cstyle; remove-trailing-spaces modified;
--- a/src/common/imageio_module.h
+++ b/src/common/imageio_module.h
@@ -231,6 +231,13 @@ int dt_imageio_get_index_of_storage(dt_imageio_module_storage_t *storage);
 /* add a module into the known module list */
 void dt_imageio_insert_storage(dt_imageio_module_storage_t *storage);
 
+// This function returns value of string which stored in the
+// "plugins/lighttable/export/resizing_factor" parameter of the configuration file
+// and its "num" and "denum" fraction's elements to calculate the scaling factor
+// and improve the readability of the displayed string itself in the "scale" field
+// of the settings export.
+gchar *dt_imageio_resizing_factor_get_and_parsing(double *num, double *denum);
+
 // modelines: These editor modelines have been set for all relevant files by tools/update_modelines.sh
 // vim: shiftwidth=2 expandtab tabstop=2 cindent
 // kate: tab-indents: off; indent-width 2; replace-tabs on; indent-mode cstyle; remove-trailing-spaces modified;
--- a/src/libs/export.c
+++ b/src/libs/export.c
@@ -42,6 +42,8 @@
 #include <stdlib.h>
 #include <ctype.h>
 
+#include <glib.h>
+
 DT_MODULE(7)
 
 #define EXPORT_MAX_IMAGE_SIZE UINT16_MAX
@@ -53,6 +55,8 @@ typedef struct dt_lib_export_t
   GtkBox *print_size;
   GtkWidget *unit_label;
   GtkWidget *width, *height;
+  GtkWidget *scale, *size_in_px;
+  GtkBox *hbox1, *hbox2, *hbox_scale;
   GtkWidget *storage, *format;
   int format_lut[128];
   uint32_t max_allowed_width , max_allowed_height;
@@ -70,7 +74,8 @@ typedef enum dt_dimensions_type_t
 {
   DT_DIMENSIONS_PIXELS = 0, // set dimensions exactly in pixels
   DT_DIMENSIONS_CM     = 1, // set dimensions from physical size in centimeters * DPI
-  DT_DIMENSIONS_INCH   = 2  // set dimensions from physical size in inch
+  DT_DIMENSIONS_INCH   = 2,  // set dimensions from physical size in inch
+  DT_DIMENSIONS_SCALE   = 3  // set dimensions by scale
 } dt_dimensions_type_t;
 
 char *dt_lib_export_metadata_configuration_dialog(char *list, const gboolean ondisk);
@@ -120,6 +125,8 @@ static inline uint32_t print2pixels(dt_lib_export_t *self, const float value)
       return cm2pixels(self, value);
     case(DT_DIMENSIONS_INCH):
       return inch2pixels(self, value);
+    case(DT_DIMENSIONS_SCALE):
+      ;
   }
 
   // should never run this
@@ -137,6 +144,8 @@ static inline float pixels2print(dt_lib_export_t *self, const uint32_t pix)
       return pixels2cm(self, pix);
     case(DT_DIMENSIONS_INCH):
       return pixels2inch(self, pix);
+    case(DT_DIMENSIONS_SCALE):
+      ;
   }
 
   // should never run this
@@ -192,6 +201,73 @@ static void _collection_updated_callback(gpointer instance, dt_collection_change
 static void _mouse_over_image_callback(gpointer instance, dt_lib_module_t *self)
 {
   dt_lib_queue_postponed_update(self, _update);
+}
+
+bool _is_int(double value)
+{
+    return (value == (int)value);
+}
+
+static void _scale_optim()
+{
+  double num=1, denum=1;
+  dt_imageio_resizing_factor_get_and_parsing(&num, &denum);
+  gchar *scale_str = dt_conf_get_string(CONFIG_PREFIX "resizing_factor");
+  gchar _str[6] = "";
+
+  gchar *pdiv = strchr(scale_str, '/');
+
+  gchar scale_buf[64] = "";
+  if (pdiv == NULL)
+  {
+    if (_is_int(num) && num > 0.0)
+    {
+      sprintf(_str, "%d", (int) num);
+      g_strlcat(scale_buf, _str, sizeof(scale_buf));
+    }
+    else
+    {
+      g_strlcat(scale_buf, scale_str, sizeof(scale_buf));
+    }
+  }
+  else if (pdiv-scale_str == 0)
+  {
+    if (_is_int(denum) && denum > 0.0)
+    {
+      sprintf(_str, "%d", (int) denum);
+      g_strlcat(scale_buf, _str, sizeof(scale_buf));
+    }
+    else
+    {
+      g_strlcat(scale_buf, "1/", sizeof(scale_buf));
+      g_strlcat(scale_buf, pdiv+1, sizeof(scale_buf));
+    }
+  }
+  else
+  {
+    if (_is_int(num) && num > 0.0)
+    {
+      sprintf(_str, "%d", (int) num);
+      g_strlcat(scale_buf, _str, sizeof(scale_buf));
+    }
+    else
+    {
+      g_strlcat(scale_buf, scale_str, sizeof(scale_buf));
+    }
+    g_strlcat(scale_buf, "/", sizeof(scale_buf));
+    if (_is_int(denum) && denum > 0.0)
+    {
+      sprintf(_str, "%d", (int) denum);
+      g_strlcat(scale_buf, _str, sizeof(scale_buf));
+    }
+    else
+    {
+      g_strlcat(scale_buf, pdiv+1, sizeof(scale_buf));
+    }
+  }
+  dt_conf_set_string(CONFIG_PREFIX "resizing_factor", scale_buf);
+
+  free(scale_str);
 }
 
 static void _export_button_clicked(GtkWidget *widget, dt_lib_export_t *d)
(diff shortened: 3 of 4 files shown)


TurboGit:
I'm not sure about this, let's see if there is supporters for adding this. It seems like cluttering the export for little. What I don't like is that the export size and scaling are exclusive and yet present both on the GUI and clearing each other... Maybe just me.

esq4:
> What I don't like is that the export size and scaling are exclusive and yet present both on the GUI and clearing each other

I'll think about it.

esq4:
If such a variant UI?

image

image

image

The last settings are saving.

johnny-bit:
IMO this is good idea UI-wise now. Seems OK.

For additional points you could throw in dt_calculator_solve in there :) (for both scalling and width/height) or that could be separate PR :)

esq4:
> For additional points you could throw in dt_calculator_solve in there

I was thinking about using a calculator. But this seems redundant to me. And besides that, I tried to make "foolproof".

src/common/imageio_module.c
```diff
@@ -437,6 +437,89 @@ void dt_imageio_insert_storage(dt_imageio_module_storage_t *storage)
...
+
+
+void dt_imageio_scale_optim()
+{
+  double x=1, y=1;
+  dt_imageio_scale_splitup(&x, &y);
+  gchar *scale_str = dt_conf_get_string("plugins/lighttable/export/resizing_factor");
+
+  char sep[4] = "";
+  snprintf( sep, 4, "%g", (double) 3/2);
```
TurboGit, line 482: not space after (

src/common/imageio_module.c
```diff
@@ -437,6 +437,89 @@ void dt_imageio_insert_storage(dt_imageio_module_storage_t *storage)
...
+    if (y == 1.0) {
+      g_strlcat(scale_buf, "1", sizeof(scale_buf));
+    } else {
+      g_strlcat(scale_buf, pdiv+1, sizeof(scale_buf));
+    }
+  }
+  dt_conf_set_string("plugins/lighttable/export/resizing_factor", scale_buf);
+
+  free(scale_str);
+
```
TurboGit, line 520: superfluous empty line.

src/common/imageio_module.c
```diff
@@ -437,6 +437,89 @@ void dt_imageio_insert_storage(dt_imageio_module_storage_t *storage)
...
+  if (_y == 0.0) _y = 1.0;
+
+  *x = _x;
+  *y = _y;
+
+  free(scale_str);
+}
+
+
+void dt_imageio_scale_optim()
```
TurboGit, line 475: can  you add some comments, not sure I understand all this business:

src/libs/export.c
```diff
@@ -192,11 +202,15 @@ static void export_button_clicked(GtkWidget *widget, dt_lib_export_t *d)
...
 
+  gtk_entry_set_text (GTK_ENTRY(d->scale), dt_conf_get_string("plugins/lighttable/export/resizing_factor"));
+
   GList *list = g_list_copy((GList *)dt_view_get_images_to_act_on(TRUE, TRUE));
   dt_control_export(list, max_width, max_height, format_index, storage_index, high_quality, upscale, export_masks,
                     style, style_append, icc_type, icc_filename, icc_intent, d->metadata_export);
 
   g_free(icc_filename);
+
+  dt_imageio_scale_optim();
```
TurboGit, line 213: Can't we have the cale optim routine internal to export.c ? I don't like having that exported as routine of imageio.

src/libs/export.c
```diff
@@ -213,6 +227,135 @@ void _set_dimensions(dt_lib_export_t *d, int max_width, int max_height)
...
+static void scale_mdlclick(GtkEntry *spin, GdkEventButton *event, dt_lib_export_t *d)
+{
+  if (event->button == 2){
+    size_changed_reset_scale(GTK_WIDGET(spin), d);
+  } else {
+    scale_changed(spin);
+  }
+}
+
+static void widht_mdlclick(GtkEntry *spin, GdkEventButton *event, gpointer user_data)
```
TurboGit, line 313: widht -> width

Also current dt style is to have all static routines to start with an underscore. I know there is some missing....

src/libs/export.c
```diff
@@ -213,6 +227,135 @@ void _set_dimensions(dt_lib_export_t *d, int max_width, int max_height)
...
+    g_signal_handler_block(GTK_ENTRY(spin), d->height_chg2_handler_id);
+    gtk_entry_set_text(GTK_ENTRY(spin), "0");
+    g_signal_handler_unblock(GTK_ENTRY(spin), d->height_chg1_handler_id);
+    g_signal_handler_unblock(GTK_ENTRY(spin), d->height_chg2_handler_id);
+    dt_conf_set_int(CONFIG_PREFIX "height", atoi(gtk_entry_get_text(GTK_ENTRY(spin))));
+  } else {
+    height_changed(GTK_EDITABLE(spin), user_data);
+  }
+}
+static void doom_changed(GtkWidget *spin, gpointer user_data)
```
TurboGit, line 342: doom???

TurboGit (changes requested):
Ok, lot of changes for a first review.

src/libs/export.c
esq4: I'm sorry, it was a splinter of mental anguish :-(

src/common/imageio_module.c
esq4: This is where the string is cleared for better readability: from "1./2.0" is made "1/2".

src/common/imageio_module.h
```diff
@@ -231,6 +231,8 @@ int dt_imageio_get_index_of_storage(dt_imageio_module_storage_t *storage);
 /* add a module into the known module list */
 void dt_imageio_insert_storage(dt_imageio_module_storage_t *storage);
 
+void dt_imageio_scale_splitup(double *x, double *y);
```
TurboGit, line 234: either this should have the scale string as parameter (and not use the conf string) and/or a comment should be added to clarify the semantic as as-is it is really not clear what this is about.

src/libs/export.c
esq4: Yes, it has become possible. Did.

src/common/imageio_module.h
esq4: `The function from the string stored in the "plugins/lighttable/export/resizing_factor" parameter of the configuration file generates the x and y fraction's elements to calculate the scaling factor and improve the readability of the string itself displayed in the "scale" field of the export settings.`

Does this comment clarify anything?

TurboGit: Yes, I still don't like this global routine using an hidden prefs. I would prefer having a more generic routine taking a string as parameter and returnin x & y.

TurboGit:
BTW, please rebase on top of current master and do not merge master into your topic branch. TIA.

src/libs/export.c
```diff
@@ -122,6 +132,81 @@ static void _mouse_over_image_callback(gpointer instance, dt_lib_module_t *self)
...
 
+bool _is_int(double value)
+{
+    return (value == (int)value);
+}
+
+static void _scale_optim()
+{
+  double a=1, b=1;
+  dt_imageio_scale_splitup(&a, &b);
```
TurboGit, line 143: So my proposal is more something like:

```
gchar *scale_str = dt_conf_get_string(CONFIG_PREFIX "resizing_factor");
dt_imageio_scale_splitup(scale_str, &a, &b);
```

And a, b could probably be renamed num, denum (for numerator denominator).

src/common/imageio_module.h
TurboGit: You did not change this? Can you explain why? I mean I'd like to understand if there is good arguments to keep it the way it is.

src/common/imageio_module.h
```diff
@@ -231,6 +231,12 @@ int dt_imageio_get_index_of_storage(dt_imageio_module_storage_t *storage);
 /* add a module into the known module list */
 void dt_imageio_insert_storage(dt_imageio_module_storage_t *storage);
 
+// This function generates the a and b fraction's elements to calculate the scaling factor
```
TurboGit, line 234: you probably also to fix a & b in comment.

TurboGit:
Apart from my latest comments I tested it and it is working fine on my side. Will be a nice addition.

src/common/imageio_module.h
esq4: Yes, of course I will fix it. I'm sorry.

src/libs/export.c
```diff
@@ -213,6 +227,135 @@ void _set_dimensions(dt_lib_export_t *d, int max_width, int max_height)
   g_free(max_height_char);
 }
 
+static void scale_changed(GtkEntry *spin)
+{
+  const char *validSign = ",.0123456789";
+  const gchar *value = "";
+  value =  gtk_entry_get_text(spin);
+
+  int len = sizeof(value);
```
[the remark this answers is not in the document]
TurboGit: This is marked as resolved, but it has not been change to const, I suppose this is possible?

src/common/imageio_module.h
```diff
@@ -231,6 +231,12 @@ int dt_imageio_get_index_of_storage(dt_imageio_module_storage_t *storage);
 /* add a module into the known module list */
 void dt_imageio_insert_storage(dt_imageio_module_storage_t *storage);
 
+// This function generates the "num" and "denum" fraction's elements to calculate the scaling factor
+// and improve the readability of the string itself displayed in the "scale" field of the
+// export settings from the string stored in the "plugins/lighttable/export/resizing_factor"
+// parameter of the configuration file.
+gchar *dt_imageio_resizing_factor_splitup(double *num, double *denum);
```
TurboGit, line 238: Again, I asked to have the scale string as parameter here. I still think this is not ok, using an hidden pref as parameter for a public API sounds wrong.

TurboGit (changes requested):
Last but not least, on #6336 there is a proposal to add a way to select physical units like (inch or centimeter) and a DPI to get a final size in pixels. I suppose this PR will need to be reworked a bit to fit on this new module layout. 

We have a combo names unit with entries : pixel, cm, in

I propose that we add a new entry name : factor

In this mode the label "max size" would be renamed "scaling" with the single entry as done here.

So maybe, just wait a bit for the other PR to be merged as I have asked for some modifications.

src/common/imageio_module.h
esq4: This function is quite specific and works with a single configuration parameter.
The current implementation made it possible to exclude an extra line of accessing the configuration from the code.
Your option returns it. But of course I can implement it.

esq4:
> So maybe, just wait a bit for the other PR to be merged as I have asked for some modifications.

Well I will wait.

TurboGit:
@esq4 : the print-export PR is merged so you can rework this PR on top of master. I would add a new entry in the combobox for support of scaling (with have pixel, cm & inch). And note that this must be only a GUI thing and we want only to have the width & height to be set properly. TIA.

esq4:
OK, wil do.

TurboGit:
Perfect thanks! So I'll wait before merging  #6479 to also document the new scaling support.

esq4:
I offer to do so

image

image

image

image

image

TurboGit:
I don't understand why the scale is not one of the options in the combo where we already have cm, inch, pixel? Having another combo just makes this gui more complex than necessary.

esq4:
> I don't understand why the scale is not one of the options in the combo where we already have cm, inch, pixel? Having another combo just makes this gui more complex than necessary.

"Scale" is a multiplier. It's logical to separate it from units of measurement. In addition, when switching the combo "unit" by scrolling, the sizes are recalculated between in, cm and px. And it seemed convenient to me. Switching to scale (unintentional) in this sequence is confusing.

TurboGit:
I don't see this this was.

The module is an interface to compute the final size in pixels and only the width in pixels and height in pixels are actually needed by this module. Now we have many ways to specify those values:

- enter directly in pixel
- enter in CM and in this case the size in pixels are displayed but not editable
- enter in INCH and in this case the size in pixels are displayed but not editable

I'm proposing to add a new option:

- enter a scale from the size of the picture and in this case the size in pixels should be displayed but not editable.

This way the UI stay not too saturated of controls.

esq4:
How do you suggest calculating the size in pixels when several images of different sizes are selected? I do not know how.

TurboGit:
> And what is the point of displaying the size in pixels when the file is being prepared for printing?

I don't get the question. Why not ? It gives an indication of the actual number of pixels for the exported image as this is the export module (and not the print module) and at the end you have a picture on disk with a number of pixels in width and height.

TurboGit:
> How do you suggest calculating the size in pixels when several images of different sizes are selected? I do not know how.

Indeed, that's a good point. We need to take this into account.

TurboGit:
> Therefore, it is logically correct to separate these methods in UI.

Maybe in the code, but I still don't agree in the GUI. We may want to have an entry scale in the current combo as I proposed and maybe in this case hide the non editable size in pixels. Then in scale mode we want to record the scale but we probably want to do lot of tweaking afterward as IIRC in imageio code which does the export knows only about max pixels sizes.

esq4:
Maybe so?

image

image

image

image

TurboGit:
Yes, looks better to me but it looks like you have removed the non editable size in pixels for cm & inch. This is very informative, I don't see why you've removed this. Also even if not used later for other pictures it may make sense to also keep this for scale display. I have really not strong opinion for this last part...

esq4:
I find it superfluous to display a non-editable size in pixels for several reasons:
- when preparing for printing, only the paper size and printer resolution are important. Maximum size in pixels does not specify anything;
- if the user wants to see the maximum size in pixels, then it is enough to select the appropriate item in the combo;
- with all this, it simply takes up space, saturating of the UI of the module.

There is a thought to display this information in a tooltip, but most likely it will be work for the sake of work and is not in demand by anyone.

TurboGit:
@esq4 : The display of pixel sizes was done by Aurélien some days ago. I won't remove that except if Aurélien approves this change. @aurelienpierre : Can you cast your vote on this one ? Or anyone else BTW...

esq4:
@TurboGit : Yes, it's better to know his opinion.

aurelienpierre:
The actual pixels count is the result of the computation. I just like this kind of double-checking display, both for quick reference and to ensure everything is alright. 

Also, from an user education point of view, it makes very clear that the pixel count is the only thing that actually matters for the export, and DPI × size is merely an interface that produces the pixel count indirectly. Indeed, you will find many users wondering at what DPI they should export for web, or that kind of things that reeks of misunderstanding, so I never let an occasion to use the GUI to teach (because, let's be honest, nobody reads the docs).

esq4:
OK, the approach is clear.

esq4:
image

image

TurboGit:
Some issues:
- The equal pixel displayed in print size (cm or inch) is not refreshed when changing DPI.
- Going from scale to pixels does not refresh the entries (scaling factor is still displayed)
- Select pixel 0 x 0 export OK, then scale 1/2 OK, then select CM and then pixels (just to have the entries selected show 0x0), export and you still have the scale 1/2
- Last but not least, each time I export I have on the console:

```
[lcms2] error 9: Wrong output color space on transform
unsupported output profile `ISO Coated' has been replaced by sRGB!
[lcms2] error 9: Wrong output color space on transform
unsupported output profile `ISO Coated' has been replaced by sRGB!
[lcms2] error 9: Wrong output color space on transform
unsupported output profile `ISO Coated' has been replaced by sRGB!
[lcms2] error 9: Wrong output color space on transform
unsupported output profile `ISO Coated' has been replaced by sRGB!
[export_job] exported to `/home/obry/Images/Photos/export/tests/img_19_10.jpg'
```

Please be sure to test all possible combinations. Thanks.

TurboGit:
Hum, looks like last issue is not due to this PR!

esq4:
> Some issues:
> ...

Thanks, fixed. I have not found others yet.

TurboGit:
Still some issues:
- select pixels 1920x1920 export OK
- select scale (let 1) as proposed => export image is 1920 WRONG
- enter 1 in scale (already there but editing seems to reset)
- export OK - size is max
- back to pixels, the size has been reset to 0 x 0 (would be nice to keep last entered 1920x1920).

TurboGit (changes requested):
Still some issues.

esq4:
> * enter 1 in scale (already there but editing seems to reset)

Yes, it is reset only when editing. When switching, I do not change, since it is very easy to do it through negligence.
I'll think about how to solve this.

esq4:
> > * enter 1 in scale (already there but editing seems to reset)
> 
> Yes, it is reset only when editing. When switching, I do not change, since it is very easy to do it through negligence.
> I'll think about how to solve this.

Removed reseting values when switching. All settins are saving.
Added resetting of dimensions by MMB.

src/libs/export.c
```diff
@@ -902,18 +1135,23 @@ void gui_init(dt_lib_module_t *self)
...
+  dt_bauhaus_combobox_add(d->dimensions_type, _("in cm (for print)"));
+  dt_bauhaus_combobox_add(d->dimensions_type, _("in inchs (for print)"));
+  dt_bauhaus_combobox_add(d->dimensions_type, _("by scale (for file)"));
   dt_bauhaus_combobox_set(d->dimensions_type, dt_conf_get_int(CONFIG_PREFIX "dimensions_type"));
 
   d->print_width = gtk_entry_new();
-  gtk_widget_set_tooltip_text(d->print_width, _("maximum output width\nset to 0 for no scaling"));
+  gtk_widget_set_tooltip_text(d->print_width, _("maximum output width limit.\n"
+                                                "click middle mouse button sets to 0\n"
+                                                " for cancel the constraint"));
```
TurboGit, line 1149: should be "to cancel..." but I would just drop this line. I would rephrase the line above:

"click middle mouse button to reset to 0"

src/libs/export.c
```diff
@@ -902,18 +1135,23 @@ void gui_init(dt_lib_module_t *self)
...
   gtk_entry_set_width_chars(GTK_ENTRY(d->print_width), 5);
   d->print_height = gtk_entry_new();
-  gtk_widget_set_tooltip_text(d->print_height, _("maximum output height\nset to 0 for no scaling"));
+  gtk_widget_set_tooltip_text(d->print_height, _("maximum output height limit.\n"
+                                                 "click middle mouse button sets to 0\n"
```
TurboGit, line 1153: see above

src/libs/export.c
```diff
@@ -927,45 +1165,70 @@ void gui_init(dt_lib_module_t *self)
   dt_gui_key_accel_block_on_focus_connect(d->print_dpi);
 
   d->width = gtk_entry_new();
-  gtk_widget_set_tooltip_text(d->width, _("maximum output width\nset to 0 for no scaling"));
+  gtk_widget_set_tooltip_text(d->width, _("maximum output width limit.\n"
+                                          "click middle mouse button sets to 0\n"
+                                          " for cancel the constraint"));
```
TurboGit, line 1170: see above

src/libs/export.c
```diff
@@ -927,45 +1165,70 @@ void gui_init(dt_lib_module_t *self)
...
   gtk_entry_set_width_chars(GTK_ENTRY(d->width), 5);
   d->height = gtk_entry_new();
-  gtk_widget_set_tooltip_text(d->height, _("maximum output height\nset to 0 for no scaling"));
+  gtk_widget_set_tooltip_text(d->height, _("maximum output height limit.\n"
+                                           "click middle mouse button sets to 0\n"
```
TurboGit, line 1174: see above

src/libs/export.c
```diff
@@ -927,45 +1165,70 @@ void gui_init(dt_lib_module_t *self)
...
+  gtk_box_pack_start(d->hbox1, d->width, TRUE, TRUE, 0);
+  gtk_box_pack_start(d->hbox1, gtk_label_new(_("x")), FALSE, FALSE, 0);
+  gtk_box_pack_start(d->hbox1, d->height, TRUE, TRUE, 0);
+  gtk_box_pack_start(d->hbox1, gtk_label_new(_("px")), FALSE, FALSE, 0);
+
+  d->scale = gtk_entry_new();
+  gtk_entry_set_text (GTK_ENTRY(d->scale), dt_conf_get_string(CONFIG_PREFIX "resizing_factor"));
+  gtk_widget_set_tooltip_text(d->scale, _("it can be an integer, decimal fraction or simple fraction.\n"
+                                          "zero or empty values are equal to 1.\n"
+                                          "click the middle mouse button sets the value to 1."));
```
TurboGit, line 1203: "click the middle mouse button to reset to 1."

TurboGit (changes requested):
All working fine on my side, just some comments about the English text in tooltip.

src/libs/export.c
esq4: I tried to explain the meaning of this operation. Think don't need it?

TurboGit: Yes I think it is not needed and not proper English.

TurboGit (approved):
Looks good to me now. I'll merge now to get more field testing but on my side it is working properly now. Thanks a lot !

PR #5213, opened 2020-05-27
Contrast equalizer speedup

ralfbrown:
Optimized an innermost-loop function on the SSE codepath for >10% speedup.  Verified to produce pixel-identical output on mire1.cr2.

Using "clarity" preset on the integration-test image:
Threads  / Old / New (delta%)
1 / 6.343 / 5.398 (-14.8%)
2 / 3.275 / 2.757 (-15.8%)
4 / 1.705 / 1.493 (-12.4%)
8 / 0.886 / 0.763 (-13.8%)
16 / 0.494 / 0.433 (-12%)

PR #6080, opened 2020-08-23
Guided filter speedup

ralfbrown:
The guided filter is used by the Haze Removal iop and by blender when the mask is feathered.  The existing guided filter implementation runs 17 separate monochrome box mean filters.  By adding a four-channel version of the code, this can be reduced to four 4-channel plus one monochrome box filter runs, improving memory locality and giving the compiler some opportunities for vectorization.  This yields a 3x speedup.

I tried implementing SSE code, both for the box filter and for the variance/covariance computations, but it wasn't any faster than what the compiler produced from the scalar code....

Results verified to be pixel-identical for the two integration tests the PR adds (numbered 0025 and 0026, because 0024 is the number in the still-to-be-merged #5213.

Benchmark results, Haze Removal:
Threads   Old   New
1	4.911	1.746
2	2.871	0.998
4	1.547	0.566
6	1.041	0.408
8	0.826	0.333
12	0.586	0.255
16	0.462	0.214
24	0.352	0.185
32	0.351	0.188

Benchmark results, Exposure+Blender with large feathering radius
Threads  Old   New
1	15.487	4.176
2	9.145	2.514
4	4.642	1.329
6	3.763	1.126
8	3.766	1.110
12	2.263	0.762
16	2.260	0.741
24	2.277	0.719
32	2.265	0.727
(for comparison, turning off feathering results in 0.329 seconds for 1 thread and 0.069 for 32)
The reason that this second benchmark tops out at twelve threads is because that's the total number of tiles used due to the large radius, and tiles are processed in parallel rather than running parallel code on one tile at a time.


TurboGit (approved):
Nice job! Thanks a lot for this new speed-up!

dterrahe:
Hi guys,

I have to admit I didn't look at this before it got merged and just noticed coincidentally that it reintroduced the explicit init and cleanup implementations in hazeremoval.c  that were removed in the large introspection update (because they can just as well be handled by the default implementations). One of those cases where git is not as smart as we would hope, I'm afraid...

No harm done, but I know @ralfbrown is just as dedicated to code size and redundancy reduction as I am ;-)

I'm working on another cleanup/refactoring patch so I'll just remove it again in that context if that's ok with you?

Just something to watch out for though with other patches that have been maturing for a while; rebasing will sometimes mess this up (I know, from more than enough experience).

Cheers!

ralfbrown:
Sure, go ahead and remove the redundant code.  This was indeed something I started months ago, but only got around to finishing last weekend.

PR #6093, opened 2020-08-25
presets rework - easier removal, no undeletable presets

### LINKED ISSUE #5275: [Basecureve] Bug : new preset created cannot be deleted.
darktable-3.1.0+1862~g0d825a4ce-win64.exe

Enabled basecurve and applied the "panasonic like preset"
Use the same one to create a preset XYZ and checked the boxes to auto apply and show preset for matching images.

Now I applied the XYZ preset to another image for some reason it did not auto apply.

Now I am trying to delete the preset XYZ.
When I look at the presets I see both "panasonic like" and "XYZ" bold.
image

When I try to delete the preset (note 2 of them were active) I get the dialog to delete panasonic one but I want to delete xyz.
image

I select yest for deleting but it doesn't delete the panasonic one either. I am stuck with the 2 enabled presets.

johnny-bit:
this essentially fixes #5275 by ~not allowing users to create presets identical as non-writable ones.~ preffering to remove user-created ones instead of internal ones, so deletion and edit is triggered only for non-write protected presets

adds possibility to delete preset from edit dialog in prefs

cleanly removes preset accels.

### DIFF
--- a/src/gui/preferences.c
+++ b/src/gui/preferences.c
@@ -1492,7 +1492,7 @@ static gboolean tree_key_press_presets(GtkWidget *widget, GdkEventKey *event, gp
     gint rowid;
     gchar *name;
     GdkPixbuf *editable;
-    gtk_tree_model_get(model, &iter, P_ROWID_COLUMN, &rowid, P_NAME_COLUMN, &name, 
+    gtk_tree_model_get(model, &iter, P_ROWID_COLUMN, &rowid, P_NAME_COLUMN, &name,
                        P_EDITABLE_COLUMN, &editable, -1);
     if(editable == NULL)
     {
--- a/src/gui/preferences.c
+++ b/src/gui/preferences.c
@@ -1492,11 +1492,22 @@ static gboolean tree_key_press_presets(GtkWidget *widget, GdkEventKey *event, gp
     gint rowid;
     gchar *name;
     GdkPixbuf *editable;
-    gtk_tree_model_get(model, &iter, P_ROWID_COLUMN, &rowid, P_NAME_COLUMN, &name, P_EDITABLE_COLUMN,
-                       &editable, -1);
+    gtk_tree_model_get(model, &iter, P_ROWID_COLUMN, &rowid, P_NAME_COLUMN, &name, 
+                       P_EDITABLE_COLUMN, &editable, -1);
     if(editable == NULL)
     {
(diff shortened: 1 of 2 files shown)


parafin:
> this essentially fixes #5275 by not allowing users to create presets identical as non-writable ones.

This is probably a bad idea, since it was an intended feature. Use-case is to be able to auto-apply builtin presets, which I think you broke here. The system isn't completely flexible, since you can only have one auto-apply rule for given preset, but this PR is making the situation worse.

johnny-bit:
> This is probably a bad idea, since it was an intended feature

thanks for explaination!

> but this PR is making the situation worse.

Not any more :D Now the delete/edit menuitems prefer the user-created ones in case user created is copied from internal preset.
Is it good idea?

parafin:
> > but this PR is making the situation worse.
> 
> Not any more :D Now the delete/edit menu items prefer the user-created ones in case user created is copied from internal preset.
> Is it good idea?

It wasn't so before? Then definitely yes:)

johnny-bit:
> It wasn't so before? Then definitely yes:)

as evidenced by #5275 wasn't :)

TurboGit (approved):
Works for me, thanks!

MStraeten:
compile error on osx:
```
[ 31%] Building C object src/CMakeFiles/lib_darktable.dir/gui/preferences.c.o
/Users/martinstraeten/src/darktable_debug/src/gui/preferences.c:2064:34: error: incompatible pointer types passing 'GtkDialog *'
      (aka 'struct _GtkDialog *') to parameter of type 'GtkWidget *' (aka 'struct _GtkWidget *')
      [-Werror,-Wincompatible-pointer-types]
      dt_osx_disallow_fullscreen(dialog);
                                 ^~~~~~
/Users/martinstraeten/src/darktable_debug/src/osx/osx.h:29:44: note: passing argument to parameter 'widget' here
void dt_osx_disallow_fullscreen(GtkWidget *widget);
                                           ^
1 error generated.
```

johnny-bit:
Oh for frog's snacks... I'll fix that in a bit. Shame that we can't have travis autobuilds for mac, would help a lot with osx oddities and dedicated codepaths...

johnny-bit:
Can you please test #6161 please? that one should work.

MStraeten:
ok, builds fine now

Issue #5806, opened 2020-07-26
WB presets for the NIKON 1 V2 camera.


nkbj:
I have extracted WB preset data for the NIKON 1 V2 camera and I hope you will consider including them in a future release.

Best regards,
Niels Kristian Bech Jensen

[darktable-whitebalance-20200726.tar.gz](https://github.com/darktable-org/darktable/files/4977202/darktable-whitebalance-20200726.tar.gz)

PR #6094, opened 2020-08-26
[Issue #5806] Adding Nikon 1 V2 WB Presets

matt-maguire:
Fixes darktable-org/darktable#5806 
Added whitebalance presets for Nikon 1 V2 camera into wb_presets.c based on the inputs provided by the originator of ticket #5806


TurboGit:
Thanks !

PR #6095, opened 2020-08-26
live view histogram improvements

dtorop:
Do colorspace conversion of live view images "by hand" in tether view, rather than relying on histogram process code. This allows for "work profile" as the histogram profile to do something closer to the right thing, saves memory, and eliminates stderr warnings from dt_ioppr_get_histogram_profile_type() when histogram profile is "work" or "export".

Add some noise to 8-bit JPEG live view image before doing conversion to histogram colorspace. Otherwise it can be badly quantized (spiky histogram, banded waveform). Most visible if histogram profile is Rec.709, but noticeable in most histogram profiles.

Add a const in histogram process code.

Fix another comment typo in live view code.

### DIFF
--- a/src/libs/histogram.c
+++ b/src/libs/histogram.c
@@ -329,7 +329,7 @@ static void dt_lib_histogram_process(struct dt_lib_module_t *self, const float *
 
     if(out_profile_type != DT_COLORSPACE_NONE)
     {
-      const dt_iop_order_iccprofile_info_t *profile_info_to =
+      const dt_iop_order_iccprofile_info_t *const profile_info_to =
         dt_ioppr_add_profile_info_to_list(dev, out_profile_type, out_profile_filename, DT_INTENT_PERCEPTUAL);
       img_display = dt_alloc_align(64, width * height * 4 * sizeof(float));
       if(!img_display) return;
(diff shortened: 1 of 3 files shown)


TurboGit (approved):
Even better ! The histogram in live-view is still more on the right than the final captured pictures but it has a similar form. This looks good enough for a live-view I suppose.

dtorop:
> The histogram in live-view is still more on the right than the final captured pictures but it has a similar form. This looks good enough for a live-view I suppose.

On the Canon I see the live-view histogram more to the left, but also with a similar form. I'm imagining that whatever in-camera processing happens to generate the live view takes some shortcuts (even more so than dt's preview/thumbnail pixelpipe) so the live view and capture images will never quite match.

I realize I should delete some now-extraneous comments in  dt_lib_histogram_process() in histogram.c. Perhaps I can append that commit to this PR?

TurboGit:
No this is not possible when a PR is merged (so closed).

You can still push on the same branch and create a new PR out of the same but updated branch.

dtorop:
> You can still push on the same branch and create a new PR out of the same but updated branch.

OK, did that in #6107. Apologies for the extra noise in the incoming pull requests.

PR #6099, opened 2020-08-26
imageop: request focus to module applying presets.

### LINKED ISSUE #6090: [Dev 3.3.0-xxx Vignette] The vignette default mask/circles persist across modules and images
Vignette masks present when a preset from a another module is applied.

Edit an image, select and enable vignette module.
The vignette circle masks get displayed.
The vignette module is still in the expanded state.

Now scroll to a different module and apply the preset from the 3 horizontal lines. The preset gets applied but the vignette modules circular mask is still present. It's as if the active module did not change.

This gets worse, now go to lightable and select a different image collection and double click an image from it. This image opens in darkroom with the vignette circle mask on it.

I have, expand single module at a time enabled for darkroom preferences.

this is darktable 3.3.0+337~g53e333a11 (windows)
compile options:
  bit depth is 64 bit
  normal build
  OpenMP support enabled
  OpenCL support enabled
  Lua support enabled, API version 6.1.0
  Colord support disabled
  gPhoto2 support enabled
  ImageMagick support disabled
  OpenEXR support enabled

TurboGit:
Fixes #6090.

### DIFF
--- a/src/develop/imageop.c
+++ b/src/develop/imageop.c
@@ -1755,6 +1755,9 @@ static void dt_iop_gui_reset_callback(GtkButton *button, dt_iop_module_t *module
   /* update ui to default params*/
   dt_iop_gui_update(module);
 
+  /* and give focus to the module*/
+  dt_iop_request_focus(module);
+
   dt_dev_add_history_item(module->dev, module, TRUE);
 
   if(dt_conf_get_bool("accel/prefer_expanded") || dt_conf_get_bool("accel/prefer_enabled") || dt_conf_get_bool("accel/prefer_unmasked"))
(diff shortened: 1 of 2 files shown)


TurboGit:
@elstoc : you may want to review/comment on this PR.

elstoc:
It seems perfectly reasonable to me that if you click somewhere in the module (including the presets menu) the module gets focus but it's not consistent with any of the other controls in the header. I wonder whether it would make more sense to give focus when *applying/selecting* a preset rather than when clicking the menu?

TurboGit:
@elstoc : Indeed, probably better...

TurboGit:
New version which enable the module only if a preset is applied. Still fixes the issue and cleaner indeed.

TurboGit:
I just found out that we have the same issue if one click on the reset of another module. The vignette/crop&rotate control keep displayed but the module is not active anymore. I'll also fix this case here.

TurboGit:
Manually merged.

PR #6108, opened 2020-08-26
Fix missing include

matze:
Otherwise dt_colorspaces_color_profile_type_t is not found. Probably introduced with f1cd7.


TurboGit (approved):
Thanks!

PR #6109, opened 2020-08-27
Add validation check and fix combobox logic when typing aspect ratio of "1" into crop module

### LINKED ISSUE #5784: Crop & Rotate: Aspect ratio of "1" not accepted
If I right-click and key an aspect ratio of "2", the crop & rotate module correctly interprets this as "2:1". Similarly with other integers 3-9.

However, if I key "1" nothing happens. "1.0" is correctly interpreted as "1:1".

matt-maguire:
This PR has two commits to resolve #5784 (I split across two commits, in case you don't want to accept the combobox change)
1. When someone enters in crop module an aspect ratio with keyboard, make sure the text they type is really a number, and doesn't contain junk text characters. (this issue was mentioned in PR #5777 but wasn't addresed).

2. The reason you can't enter "1" as a crop aspect ratio is that there are more than two items in the dropdown list that match a prefix of "1". All other digits will match only 0, 1 or 2 items in the dropdown list.
  If a bauhaus combo box is flagged as "editable", it doesn't make sense to me why you would only accept custom input values where the custom value only matches a small number of items. Therefore, I propose to remove this check, and if a bauhaus combo box is editable, and the users enters something not on the downdown list, then this custom value should be accepted, otherwised what does it mean for a combobox to be "editable". Perhaps one of the other devs knows the background to this, which is why I submit this change in its own commit.

### DIFF
--- a/src/bauhaus/bauhaus.c
+++ b/src/bauhaus/bauhaus.c
@@ -1560,20 +1560,16 @@ static void dt_bauhaus_widget_accept(dt_bauhaus_widget_t *w)
         i++;
         g_free(text_cmp);
       }
-      // if list is short (2 entries could be: typed something similar, and one similar)
-      if(k < 3)
+      // didn't find it, but had only one matching choice?
+      if(k == 1 && match)
+        dt_bauhaus_combobox_set(widget, kk);
+      else if(d->editable)
       {
-        // didn't find it, but had only one matching choice?
-        if(k == 1 && match)
-          dt_bauhaus_combobox_set(widget, kk);
-        else if(d->editable)
-        {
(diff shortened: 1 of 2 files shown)


matt-maguire:
Thinking a bit more, thought I better include a check to make sure the user doesn't enter more than one decimal point (without that check, any subsequent decimal points would just be silently ignored).

johnny-bit:
Tried it and for me it seems to work perfectly for crop&rotate usecase. I couldn't break it there :smile:

only couple other modules use editable combo: framing, lens correction and pdf export and those seem to work too so it's :+1: from me :)

TurboGit:
Works for me, thanks!

Issue #5968, opened 2020-08-15
with display filtered to "All except rejected", rejecting an image in darkroom does not advance to next image


ptilopteri:
openSUSE Tumbleweed 20200813
NVIDIA GF106 [GeForce GTS 450], 390.138
darktable-3.3.0~git152.6930e52c6
OpenCL off

When I have the current display filtered to show "all except rejected" and work a group of images in darkroom and find the current image unacceptable, tagging it as "rejected" always advanced to the next image in the group.  Currently this is not happening.

Previous and I know not when it changed, when rejecting an image it would disappear from the display and the next image in the goup would appear.  A behaviour which I liked and desire.

junkyardsparkle:
IIRC this changed when the new lighttable was added. I also found the previous behavior to be convenient. When moving forward through images in darkroom view, I don't find it to be a big deal to hit the spacebar to advance after rejecting, however - if I happen to want to go *backwards* from a still-displayed image which has just been rejected, using the backspace key, the progress is to the next image in a forward direction, exactly as when using spacebar. I find that behavior particularly annoying/confusing.

AlicVB:
This is done by design and as already been discussed before (I don't remember the issue number)

@junkyardsparkle : not sure to understand your last point : you say that when you hit previous key, it show the next image ?

ptilopteri:
So when working in darkroom mode with "all except rejected" and one rejects an image, they have no knowledge that the rejection succeeded because the image is still there, no screen flash, no longer the image disappears (which should happen).

I guess I missed the "discussion" and it surely was not in this context.  And when the rejection is applied in lighttable mode, the image disappears but the selection is not advanced.

this "selection" action, lack of selection, is not intuitive and not what one would expect despite "prior discussion".

junkyardsparkle:
> you say that when you hit previous key, it show the next image ?

Yes, in the specific situation described in this issue: when still viewing an image in darkroom view after it has been rejected, when the "view" filter is set to "all except rejected". It works normally otherwise. This happens when using XFCE desktop, so it isn't one of those weird fluxbox bugs. ;-)

The behavior when rejecting the last image in a film roll is that you end up on the *first* image when using the shortcut for either "next" or "previous", which is also disorienting, but IIRC that also happened immediately upon rejecting the last image in older versions, so not really a new issue.

ptilopteri:
[...]

PR #6118, opened 2020-08-27
fix filmstrip glitches in darkroom

AlicVB:
- fix previous image when active image is removed from collection (fix issue raised by @junkyardsparkle in #5968 discussion)
- fix filmstrip rewind when last image is removed from collection (same)


TurboGit (approved):
Work for me.

junkyardsparkle:
Thanks! The new last-image behavior makes this an over-all improvement from the older versions for me; hopefully nobody will complain about that being changed...

PR #6053, opened 2020-08-22
zoomable fixes

AlicVB:
there's some glitch in zoomable lightable after the rewrite (and that's expected). Here's a bunch of fix to enhance that mode, based on recent issues : 
- fix number of image per row to 13 and not to max zooming value.
- don't change layout when reentering zoomable. This fix #5887 and #5914 
- fix alignment of rows > 0 after a redraw
- ensure image visibility when zooming in/out

PR #6126, opened 2020-08-28
zoomable : fix offset on filter change

### LINKED ISSUE #6061: Zoomable mode changes position after some file actions (paste/discard history etc.)
@AlicVB If you can check this out please. This is the only bug/glitch that remained, as Zoomable mode is now flawlessly working on my end.

However, under certain conditions like paste settings or discarding history, sometimes it moves the selected image away from the view.

In this video note that after copy/pasting the settings (with ctrl-c and ctrl-v) from one image to another, the bottom row of images disappeared. The same can happen with discard history as well.

[Bottom row disappeared](http://cvrkutanje.net/disappear.webm)

This bug is much more annoying. Here, after I discard the history, the entire row/image changes position. These two videos are similar, but involving different images.

[glitch 1](http://cvrkutanje.net/glitch.webm)

[glitch 2](http://cvrkutanje.net/glitch2.webm)

**Steps to reproduce:**
1. Open zoomable mode, show more images at once
2. Try to discard history or paste settings to images
3. Sometimes the image you pasted on will change place 
OR
4. the bottom row of images will disappear.
 
**Additional info:**
- Happens in zoomable mode only.
- Arch linux, DT GIT version.

AlicVB:
bug reported in #6053 
Chance that it fix #6061 too. To be confirmed as I've not managed to reproduce the bug

### DIFF
--- a/src/dtgtk/thumbtable.c
+++ b/src/dtgtk/thumbtable.c
@@ -1670,6 +1670,9 @@ void dt_thumbtable_full_redraw(dt_thumbtable_t *table, gboolean force)
       posx = dt_conf_get_int("lighttable/zoomable/last_pos_x");
       posy = dt_conf_get_int("lighttable/zoomable/last_pos_y");
       offset = dt_conf_get_int("lighttable/zoomable/last_offset");
+      // ensure that the overall layout doesn't change
+      // (i.e. we don't get empty spaces in the very first row)
+      offset = (offset - 1) / table->thumbs_per_row * table->thumbs_per_row + 1;
       table->thumbs_area.x = posx;
       table->thumbs_area.y = posy;
     }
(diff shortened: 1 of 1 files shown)


TurboGit:
Looks good, thanks!

maboleth:
> bug reported in #6053
> Chance that it fix #6061 too. To be confirmed as I've not managed to reproduce the bug

> @maboleth : can you check and confirm that PR #6126 fix your issue ? Thanks

@AlicVB 
I checked and it's not fixed. Still going strog.

Please check my next screen video. In it, I do these things:

[glitchy glitchy grid](http://cvrkutanje.net/glitch-zoom.webm)

- I'm giving 1 and 5 ratings to the same image.
- I zoom one level, try to rate - grid moves (always going to the bottom right).
- I do the above on every zoom level - the bigger the image becomes, the worse behaviour (rated image is getting lost from the screen).

Note: The grid moves every time I rate the image - per level. If I don't change the zoom level, I can rate as many images as I want without any bad consequences.

PR #6128, opened 2020-08-28
lightable : ensure selection background is updated in any cases

### LINKED ISSUE #6086: lighttable - selecting using keyboard - "enter" does not fully work when in zoom mode - pressing W
**Describe the bug**

When browsing lighttable - if we enter in full screen mode or zoom mode by pressing "W" then use arrow keys to navigate to another image and press enter to select it - the selection does not stick when we move the mouse.

**To Reproduce**

**Expected behavior**

the image should be selected - even when we move the mouse it should remain selected. 

**Screenshots**

The image is soft selected - we can see it but as soon as we move the mouse (and most of the cases we must) the image becomes deselected. The feature used to work in the past.
will enclose video

 - Darktable Version:  3.2.1
 - OS:  Pop os 20.04
 -  yes
 -  Nvidia RTX 2070

**Additional context**

https://www.dropbox.com/s/mjmd9xk02sgwj2m/2020-08-24%2016-52-57.mkv?dl=0

AlicVB:
this fix #6086

### DIFF
--- a/src/dtgtk/thumbtable.c
+++ b/src/dtgtk/thumbtable.c
@@ -1775,6 +1775,9 @@ void dt_thumbtable_full_redraw(dt_thumbtable_t *table, gboolean force)
       dt_control_signal_raise(darktable.signals, DT_SIGNAL_ACTIVE_IMAGES_CHANGE);
     }
 
+    // if we force the redraw, we ensure selection is updated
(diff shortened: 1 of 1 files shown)


TurboGit:
Looks good, thanks!

PR #6133, opened 2020-08-29
tone equalizer: bypass further iops when displaying luminance mask

ralfbrown:
@aurelienpierre I saw your post on pixls.us about bypassing further iops when the tone equalizer's luminance mask is being displayed.  Since pending PR #5522 already does the same thing for Retouch, it was very easy to come up with the corresponding patch for Tone Equalizer.


aurelienpierre:
Great, thanks @ralfbrown ! Code looks ok, feature is much needed indeed, need to test.

src/iop/toneequal.c
```diff
@@ -1017,7 +1017,10 @@ void toneeq_process(struct dt_iop_module_t *self, dt_dev_pixelpipe_iop_t *piece,
   if(self->dev->gui_attached && (piece->pipe->type & DT_DEV_PIXELPIPE_FULL) == DT_DEV_PIXELPIPE_FULL)
   {
     if(g->mask_display)
+    {
       display_luminance_mask(in, luminance, out, roi_in, roi_out, ch);
+      piece->pipe->mask_display = DT_DEV_PIXELPIPE_DISPLAY_PASSTHRU;
+    }
     else
       apply_toneequalizer(in, luminance, out, roi_in, roi_out, ch, d);
```
aurelienpierre, line 1025: Don't you need to reset `piece->pipe->mask_display` in the `else` branch ?

ralfbrown: Not necessary, it gets reset by the pixelpipe code.  The mask display properly disappears when toneequal is turned off or loses focus, or the mask button is clicked again.  Neither colorzones nor retouch explicitly reset mask_display.

MStraeten:
mask refresh issue: 
prerequisite: filmic active with default settings
1. tone equalizer mask is displayed, with default settings -> fine greyscaled mask
2. disabling filmic turns the mask black (just disable by clicking the toggle
3. switch mask off and on again gives the expected greyscale mask  
activating filmic in step 2 switches off mask display

ralfbrown:
Not just filmic, it looks like toggling the active state of *any* iop after toneequal in the pipeline causes a black display, provided you use the "power button".  If you open the other iop instead, toneequal loses GUI focus and runs its gui cleanup function, causing the mask display to be toggled off.  So something isn't being updated properly somewhere along the pipe when an iop's state changes without it gaining focus.  Interestingly, if you toggle the other iop's state twice (putting it back where it was), the proper mask display reappears -- probably pulled out of the pixelpipe cache.

BTW, I also tried implementing Aurelien's suggestion of resetting piece->pipe->mask_display in the else clause, and that just makes it impossible for iops *before* toneequal to display their masks.

My guess is that the pixelpipe cache will need some reworking to take into account when an iop is displaying its mask, because I can get related behavior by turning on mask display in retouch (given a gray image), toggling filmic off and back on (yielding a color image most likely pulled from the cache, but with the mask button in retouch still active).  Then clicking the retouch mask button doesn't change the image, but a second click turns on the mask again and reverts to gray.

The value of piece->pipe->mask_display may need to be included by the hash function used by the pixelpipe cache.

TurboGit:
@ralfbrown : can you fix the conflict ? probably due to the merging of the speed-up for retouch. TIA.

ralfbrown:
That was an easy one - two different feature branches added the same line, just differing by whitespace.  Git isn't smart enough to figure that out for itself.

TurboGit:
Rebased & merged manually. Thanks.

PR #6129, opened 2020-08-28
export: disable insert-text handler when programatically inserting

elstoc:
Resolves #6127

Issue #6127, opened 2020-08-28
[lighttable] Heap-overflow on darktable startup


cryptomilk:
```
=================================================================
==18761==ERROR: AddressSanitizer: heap-buffer-overflow on address 0x602000e88f34 at pc 0x7f4ad2762f41 bp 0x7ffc0d3f2920 sp 0x7ffc0d3f20d0
READ of size 5 at 0x602000e88f34 thread T0
    #0 0x7f4ad2762f40  (/usr/lib64/libasan.so.6+0x3ff40)
    #1 0x7f4ad17100b3 in g_strdup (/usr/lib64/libglib-2.0.so.0+0x740b3)
    #2 0x7f4ad0a397fc  (/usr/lib64/libgobject-2.0.so.0+0x3a7fc)
    #3 0x7f4ad0a2b54b in g_signal_emit_valist (/usr/lib64/libgobject-2.0.so.0+0x2c54b)
    #4 0x7f4ad0a2c0be in g_signal_emit_by_name (/usr/lib64/libgobject-2.0.so.0+0x2d0be)
    #5 0x7f4ad10b1f34  (/usr/lib64/libgtk-3.so.0+0x1baf34)
    #6 0x7f4a7520e16d in insert_text_handler /home/abuild/rpmbuild/BUILD/darktable-3.3.0~git389.d04857316/src/libs/export.c:511
    #7 0x7f4ad0a12fc1 in g_closure_invoke (/usr/lib64/libgobject-2.0.so.0+0x13fc1)
    #8 0x7f4ad0a25692  (/usr/lib64/libgobject-2.0.so.0+0x26692)
    #9 0x7f4ad0a2bd38 in g_signal_emit_valist (/usr/lib64/libgobject-2.0.so.0+0x2cd38)
    #10 0x7f4ad0a2c0be in g_signal_emit_by_name (/usr/lib64/libgobject-2.0.so.0+0x2d0be)
    #11 0x7f4ad10b1f34  (/usr/lib64/libgtk-3.so.0+0x1baf34)
    #12 0x7f4ad10b9fcf in gtk_entry_set_text (/usr/lib64/libgtk-3.so.0+0x1c2fcf)
    #13 0x7f4a7520ef25 in _set_dimensions /home/abuild/rpmbuild/BUILD/darktable-3.3.0~git389.d04857316/src/libs/export.c:204
[...]

PR #6136, opened 2020-08-29
export: try to be defensive about potential problems in insert-text

johnny-bit:
additionally to #6129, this uses try_new0 so that in case of any problem the insert-text handler bail early and prevents insertion of  text in fields

the #6129 needs to be merged to fix the problem mentioned in #6127. This pr just adds a "sanity check" for insert text and lenght (try_new0 will return NULL in all problem cases including when length is 0)


TurboGit:
Looks ok to me!

PR #6142, opened 2020-08-29
culling : avoid flickering when switching between images

### LINKED ISSUE #5916: [3.2.1] Lighttable: moving among images causes annoying blinks/flickering
In lighttable with both
- culling, zoom 1, fixed zoom
- file manager view, zoom level 1;

while using mouse scroll wheel, to switch among images, you get a kind of flash effect during the transition. I mean, when you change image, for a fraction of second you see one of the images stretched on the whole area, I mean, one of the images is shown outside its natural borders. You can easily see this if you are moving among photos that have the same aspect ratio.
To better understand please see this video and make sure to select 1080P 60 FPS Youtube option.
https://youtu.be/CHbfP2I83sg

Another variant of this is when you also have your mouse pointer over the image, in addition to the previously described flash, you will get also a flash from the background colour appearing while switching among photos
To better understand please see this video and make sure to select 1080P 60 FPS Youtube option.
https://youtu.be/F0VxA66yk2U

So in both scenarios, when you move among various images you get a very annoying flashing effect. If you cannot easily manage to reproduce the effect, trying scrolling the mouse wheel faster.
Please note that I already let the system create all previews of photos in the active collection.

 - Darktable Version:3.2.1
 - OS: Fedora 32
 - It happens with OpenCL and without OpenCL. I have the feeling that with OpenCL it happens more often but I am not sure
 - Radeon RX480. When using OpenCL [1], the driver is amdgpu-pro-20.10-1048554, when not using OpenCL the driver is amdgpu included in kernel 5.7.11-200.fc32.x86_64

**Additional context**
I don't remember it was happening with 3.0.x
Both

[1]: Started with command `$ OPENCL_VENDOR_PATH=/home/caterpillar/amdgpu-pro-20.10-1048554-rhel-8.1/etc/OpenCL/vendors/  LD_LIBRARY_PATH=/home/caterpillar/amdgpu-pro-20.10-1048554-rhel-8.1/opt/amdgpu-pro/lib64/ darktable`

AlicVB:
this fix #5916

### DIFF
--- a/src/dtgtk/culling.c
+++ b/src/dtgtk/culling.c
@@ -1452,12 +1452,16 @@ void dt_culling_full_redraw(dt_culling_t *table, gboolean force)
   float old_z = 1.0;
   float old_zx = 0.0;
   float old_zy = 0.0;
+  int old_margin_x = 0;
+  int old_margin_y = 0;
   if(g_list_length(table->list) > 0)
   {
     dt_thumbnail_t *thumb = (dt_thumbnail_t *)g_list_nth_data(table->list, 0);
     old_z = thumb->zoom;
     old_zx = thumb->zoomx;
     old_zy = thumb->zoomy;
+    old_margin_x = gtk_widget_get_margin_start(thumb->w_image_box);
+    old_margin_y = gtk_widget_get_margin_top(thumb->w_image_box);
   }
   // we recreate the list of images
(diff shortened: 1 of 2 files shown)


TurboGit:
Works for me, thanks!

Germano0:
@TurboGit can you also test file manager view? At the moment I cannot do that

PR #6143, opened 2020-08-29
zoomable : avoid grid move on collection change

### LINKED ISSUE #6061: Zoomable mode changes position after some file actions (paste/discard history etc.)

AlicVB:
this fix #6061
this was due to a missing update of zoomable positions after each zoom change

### DIFF
--- a/src/dtgtk/thumbtable.c
+++ b/src/dtgtk/thumbtable.c
@@ -781,6 +781,15 @@ static void _zoomable_zoom(dt_thumbtable_t *table, int oldzoom, int newzoom)
   changed += _thumbs_remove_unneeded(table);
   if(changed > 0) _pos_compute_area(table);
 
+  // we update all the values
+  dt_thumbnail_t *first = (dt_thumbnail_t *)g_list_first(table->list)->data;
+  table->offset = first->rowid;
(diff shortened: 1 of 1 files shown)


TurboGit:
Looks good, thanks!

PR #6144, opened 2020-08-29
Guided filter scalability improvement

ralfbrown:
Flipped around the parallelization (parallel-process tiles one at a time instead of single-thread processing of multiple tiles in parallel), which improves the scalability when the filter radius is large and thus there are few tiles.  This also reduces memory requirements because only one tile is active at a time instead of multiple tiles.

Merged two of the four-channel and the remaining single-channel box filter into a 9-channel filter, which got a noticeable speedup despite no longer nicely matching up to vector instructions.  Memory bandwidth is still the limitation at high thread counts.

Timings on mire1:
Exposure+Blender feathering (test 0025)
Thr  / Before / After
1 / 4.176 / 3.501
2 / 2.514 / 1.920
4 / 1.329 / 1.072
6 / 1.126 / 0.791
8 / 1.110 / 0.636
12 / 0.762 / 0.524
16 / 0.741 / 0.461
24 / 0.719 / 0.398
32 / 0.727 / 0.386

Haze Removal
Thr / Before / After
1 / 1.746 / 1.529
2 / 0.998 / 0.877
4 / 0.566 / 0.504
6 / 0.408 / 0.364
8 / 0.333 / 0.316
12 / 0.255 / 0.229
16 / 0.214 / 0.202
24 / 0.185 / 0.176
32 / 0.188 / 0.187
(The guided filter accounts for less than half the runtime of the haze removal.)

### DIFF
--- a/src/common/guided_filter.c
+++ b/src/common/guided_filter.c
@@ -213,6 +213,78 @@ static inline void box_mean_1d_4ch(int N, const float *x, float *y, size_t strid
   }
 }
 
+// calculate the one-dimensional moving average over a window of size 2*w+1 independently for each of eight channels
+// input array x has stride 8, output array y has stride stride_y
+static inline void box_mean_1d_8ch(int N, const float *x, float *y, size_t stride_y, int w)
+{
+  float n_box = 0.f, m[8] = { 0.f, 0.f, 0.f, 0.f, 0.f, 0.f, 0.f, 0.f };
+  float c[8] = { 0.f, 0.f, 0.f, 0.f, 0.f, 0.f, 0.f, 0.f };
+  if(N > 2 * w)
+  {
+    for(int i = 0, i_end = w + 1; i < i_end; i++)
+    {
+      SIMD_FOR (int k = 0; k < 8; k++)
+        m[k] = Kahan_sum(m[k], &c[k], x[8*i+k]);
+      n_box++;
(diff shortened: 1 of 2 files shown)


src/common/guided_filter.c
```diff
@@ -55,6 +55,18 @@
 # define _mm_prefetch(where,hint)
 #endif
 
+#ifndef dt_omp_shared
+#ifdef _OPENMP
+#if defined(__clang__) || __GNUC__ > 8
+# define dt_omp_shared(var, ...)  shared(var, __VA_ARGS__)
```
TurboGit, line 61: I suppose this should be:
`# define dt_omp_shared(...)  shared(__VA_ARGS__)`

To allow for a single argument to be passed.

ralfbrown: You're right, though interestingly enough GCC9 is fine with that definition and a single arg.  I had just copied the definition I used in common/nlmeans_core.c; now that it's used in two places, it should probably be refactored into a header file (the obvious place being right after the definition of dt_omp_firstprivate in common/darktable.h).

TurboGit:
Works for me! Thanks.

src/common/guided_filter.c
TurboGit: I'll try to do that indeed.

PR #6145, opened 2020-08-29
thumbtable : redefine mousover after collection change

### LINKED ISSUE #5983: Lighttable - Filtering for "All except rejected" and rejecting an image by pressing R doesn't select the newly display image
**Describe the bug**
In the lighttable view, while in culling mode, filtering for "All except rejected" and then rejecting an image by pressing `R` correctly hides the reject image and shows the next image in the filmstrip. But it doesn't select that next image.

Pressing `R` again un-rejects the previous.

When moving the mouse after pressing `R`, the newly displayed image is selected.

This is very annoying  when rejecting images in a burst shot where you have 10 images or so but only 1 worth keeping.

**To Reproduce**

Consider images A, B, C

1. Open the lighttable
1. Set the view to `All except rejected`
1. Select image **A**
1. Press the `R` key to reject it
1. image **A** is rejected and hidden. The image **B** is displayed but not selected.
1. Press the `R` key
1. The image **A** is un-rejected and shown in the filmstrip. Image **B** is now selected.
1. Press the `R` key
1. Image **B** is rejected and hidden, image **C** is displayed but not selected.
1. Press the `R` key
1. image **B** is un-rejected and shown in the filmstrip. Image **C** is now selected.

**Expected behavior**

1. Open the lighttable
1. Set the view to `All except rejected`
1. Select image **A**
1. Press the `R` key to reject it
1. image **A** is rejected and hidden. The image **B** is displayed **and selected.**
1. Press the `R` key
1. Image **B** is rejected and hidden, image **C** is displayed **and selected**

 - Darktable Version: 3.2.1 (from Flatpack)
 - OS: Linux Mint 19.3 (Tricia)
 - OpenCL not available

AlicVB:
this fix #5983

### DIFF
--- a/src/dtgtk/thumbtable.c
+++ b/src/dtgtk/thumbtable.c
@@ -1230,6 +1230,7 @@ static void _dt_collection_changed_callback(gpointer instance, dt_collection_cha
   dt_thumbtable_t *table = (dt_thumbtable_t *)user_data;
   if(query_change == DT_COLLECTION_CHANGE_RELOAD)
   {
+    int old_hover = dt_control_get_mouse_over_id();
     /** Here's how it works
      *
      *          list of change|   | x | x | x | x |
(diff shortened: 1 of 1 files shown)


TurboGit:
@AlicVB : on my side after R is pressed the next image is not selected.

AlicVB:
@TurboGit : nope, I don't think it's good to change selection here. I've just updated correctly the mouseover : 
Previously, if the mouseover image is remove from the collection after changing its rating, no image is hovered anymore, even the image under mouse.
Now, if the mouseover image is remove from the collection, the first untouched image after is hovered.

This allow the workflow @jobec describe in #5983 : you put your mouse on image A, then multiple R press will reject images B, C, D,... without having to move the mouse.
I've taken "selected" as "hovered" in his description...

If we want to change the selection, it end with very unexpected things : for example, imagine you have selected images ABCD that would mean that you get images ABCE selected after rejected image D ? Not good imho...

TurboGit:
@AlicVB : understood, this was not clear to me reading #5983. So merging, thanks.

PR #6154, opened 2020-08-30
set history tooltip widget background to match tooltip border

dterrahe:
As discussed here https://github.com/darktable-org/darktable/pull/5770#issuecomment-683318582

@Nilvus please confirm if you agree with the location where I put the overrides and the small renaming of the tag (to align with the other #history- tags and to avoid confusion with histogram related tags).

### DIFF
--- a/data/themes/darktable.css
+++ b/data/themes/darktable.css
@@ -1390,6 +1390,12 @@ cell:selected
   font-size: 0.5em;
 }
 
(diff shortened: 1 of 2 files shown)


Nilvus (approved):
All good for me. Thanks.

TurboGit:
Thanks!

PR #6157, opened 2020-08-31
[Issue #6124] "Update Preset" operation sometimes fails silently in e…

### LINKED ISSUE #6124: Export preset not updating with Style
**Describe the bug**
I've created a Style preset which I want to include in an Export preset.
But the existing Export preset will not update.

**To Reproduce**
I'm on Windoze 10, running 3.2.1
I have an Export preset called Instagram which epxorts to a set folder, at a set resolution and file format.
I decided I wanted to add a Style preset to this Export preset.
I add the Style.
I click on the hamburger and select Update preset Instagram`, and click Yes on the confirmation.
Select any other Export preset.
Then select Instagram Export preset.
Style has not been saved to the preset.

Apologies if this has been raised elsewhere.

matt-maguire:
Fixes #6124 
When doing an update preset, sometime the operation will fail silently.
The SQL UPDATE command searches based on preset name only, but this can return multiple rows if the same preset name is used in multiple modules. It then tries to force the "operation" field to be the same on all those rows, but this causes a primary key violation, and so the update operation fails.

Solution is to use the operation field as part of the selection criteria, thereby only updating a single row relevant to the specific module in which the presets are being updated.

### DIFF
--- a/src/libs/lib.c
+++ b/src/libs/lib.c
@@ -294,7 +294,7 @@ static void menuitem_update_preset(GtkMenuItem *menuitem, dt_lib_module_info_t *
     // commit all the module fields
     sqlite3_stmt *stmt;
     DT_DEBUG_SQLITE3_PREPARE_V2(dt_database_get(darktable.db),
-                                "UPDATE data.presets SET operation=?1, op_version=?2, op_params=?3 WHERE name=?4",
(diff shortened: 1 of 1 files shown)


TurboGit (approved):
Looks correct indeed, thanks!

PR #6158, opened 2020-08-31
Bilateral fix

### LINKED ISSUE #6065: Denoise bilateral filter error
Windows 10
darktable 3.2.1
OpenCL actived
multiple GPUs
(GTX 1080TI 11GB & GTX 1070 8GB)
NVIDIA DRIVER Version 411.66

1. Open an image in darkroom
2. Activate denoise (bilateral filter) module
3. Error: strange aleatory strips over image.
denoise bilateral filter01

ralfbrown:
Partially fixes #6065 
Eliminates the visual errors, but playing around with the sliders in denoise(bilateral) eventually causes Linux to spontaneously reboot both before and after this patch.  Crashes occurred while increasing the color-blurring sliders, which is ironic because a larger value produces a coarser grid and hence a bilateral lattice with fewer points.


MStraeten:
Doesn’t fix on osx reproduceable. Sometimes it fixes denoise (bilateral filter) but after moving sliders the issue is back. At least it doesn’t crash osx ;)

TurboGit (approved):
Looks ok even if this does not fixes all issues. Thanks.

PR #6163, opened 2020-08-31
temperature.c: avoid memleak in gui_cleanup

johnny-bit:
Fixes a memleak detected by @dterrahe in #5433 

for completness i checked and indeed all other IOPs that declare their own gui_cleanup do free gui_data :)


TurboGit:
Perfect! Thanks.

PR #6170, opened 2020-09-01
zoomable : fix nb of row computation

### LINKED ISSUE #6160: Zoomable - thumbnails of the bottom row flickering
@AlicVB 
I hope this is the last glitch in Zoomable that I discover. :D 

It happens when trying to rate images, the bottom row flickers (actually, thumbnails are centering themselves to the grid). 

The pre-condition for this to happen is to have thumbnails (the bigger the better) slightly or partly covered by the screen - so they are not shown in their full fit.

Like this:
flicker

Steps:
- Zoomable mode showing many images at once (however, bigger thumbnails display it faster). Try several zoom levels.
- Be sure to select the thumbnails from the middle, not from the beginning.
- Have the bottom row of thumbnails partly covered by the screen.
- Try to rate any image on the screen.
- Notice flickering/centering of the bottom row.

Latest GIT master ver.

AlicVB:
this fix #6160 by avoiding to remove and reload the last line

### DIFF
--- a/src/dtgtk/thumbtable.c
+++ b/src/dtgtk/thumbtable.c
@@ -379,7 +379,7 @@ static gboolean _compute_sizes(dt_thumbtable_t *table, gboolean force)
       table->view_width = allocation.width;
       table->view_height = allocation.height;
       table->thumb_size = table->view_width / npr;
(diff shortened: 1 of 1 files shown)


TurboGit:
Thanks !

PR #6181, opened 2020-09-02
selection : fix range selection if collection has changed

### LINKED ISSUE #5853: Ranged selection can be started from an image not in the current collection.
Following a discussion on IRC I decided to submit this as an issue.

**Describe the bug**
Ranged selection (shift+click) can be started from an image that is not in the current collection (but still selected). This can cause unintentionally selecting (and modifying) hundreds of images, which is not user friendly.

**To Reproduce**
1. Open a collection in lighttable
2. Select an image
3. Remove this image from the current collection (e.g. rejecting or removing a tag), or switch to a different collection that does not contain this image.
4. Shift+click on an image. All the images up to (or from) this image are now selected.

**Expected behavior**
Only the image clicked in step 4 is selected, and can be used as the start of a new ranged (shift+click) selection.

**Platform**
 - Darktable Version: can reproduce in official 3.0.2 and self-compiled 3.2.0 builds.
 - OS: Manjaro Gnome

**Other comments**
I guess this issue can also be formulated in a broader sense: Images stay selected when no longer in the current collection. This can be intended behaviour, but I think this is not intuitive.

On IRC, the following possible solutions were proposed.
1. When using shift+click, check if the starting point is still in the current collection.
2. When an image is removed from the collection (or the active collection is changed), reset the selection.
3. When an image is removed from the collection, set the next image in line as selected. On switching collections, reset the selection or select the first image (up for discussion).

For me, the most intuitive solution would be: Select the next image when an image is removed. On switching collections, reset the selection (unless the old image is also in the new collection).

AlicVB:
this fix #5853

the problem was when the last selected image was no more in current collection. In this case, we use either the last image in selection either the first image in collection as "starting" point.

### DIFF
--- a/src/common/selection.c
+++ b/src/common/selection.c
@@ -282,12 +282,12 @@ void dt_selection_select_range(dt_selection_t *selection, uint32_t imgid)
 {
   gchar *fullq = NULL;
 
-  if(!selection->collection || selection->last_single_id == -1) return;
+  if(!selection->collection) return;
 
   /* get start and end rows for range selection */
   sqlite3_stmt *stmt;
   int rc = 0;
-  uint32_t sr = -1, er = -1;
+  int sr = -1, er = -1;
   DT_DEBUG_SQLITE3_PREPARE_V2(dt_database_get(darktable.db), dt_collection_get_query_no_group(selection->collection),
                               -1, &stmt, NULL);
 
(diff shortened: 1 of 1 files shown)


TurboGit:
Works for me! A somewhat non trivial use, good to see this fixed. Thanks!

Issue #5863, opened 2020-08-03
darktable 3.3.0: cannot set white balance to 4000K, it results in 3901K


bd23843:
git rev. cb17b3bfa91538d8ee4e5f9721312915cd3f066a on ubuntu 20.04, opencl on/off makes no difference.

There is a white balance range around 4000k that cannot be set in this version. No matter if I use the slider or enter the value numerically, it is not possible to get 4000k, darktable changes value to 3901K . The tint is also automagically modified.

There is no message on terminal indicating a problem.

johnny-bit:
That's not a problem, that's just "a feature" of how "temp" is calculated. you see: everything above 4000K is "daylight" reference, while anything below is technically "blackbody" reference. however there's an overlap at roughly 3900-4000K where coeficients calculated with daylight reference may match those with blackbody reference and vice versa. You still have the same range of creative freedom, nothing really changes, the actual values that mean anything are r/g/b coefficients and temp/tint is just for your convinience. sort of "visual sugar"

As for the "fix" it is planned to be included soon enough, see #5726 which fixes this issue

johnny-bit:
fixed in #5726, not a problem with modern color calibration workflow...

PR #6183, opened 2020-09-02
[temperature.c] introspection, refactoring, janitorial

### LINKED ISSUE #6165: [regression] #5125 came back --> Color picker in white balance loses content if switcing to other mode like camera
**Describe the bug**

@johnny-bit did the wonderful re-design of the white balance module, however a regression happened as #5125 came back

**To Reproduce**
pls see #5125

**Expected behavior**

as after #5137

**bisect**
```
There are only 'skip'ped commits left to test.
The first bad commit could be any of:
1443fff87122fe59e6302b9c67bf82161fa45d45
c5569b4a6d8ab5a349f8967ed40613e2bb965d34
b8d6720dd41bf2944e2c547ddcdff58024ef2227
e1372161b8aabf61d7d2cd323c67951a24934e69
2eeb206f79517121d823b3bc19c176d667adaef7
cad6f517b7df1a427c5250412f68821c01cf67cb
1b15f29408a9904387e124a3c36b4e6f4f3ab430
2481e8e073910e19555da56abd01e142b180606e
1f2ee768f744455718ba2aaf07fbcd8cbdaf300a
f144985ad890832fa01a48d6f95a2ed1194c19c5
833765d117d56dd1cff74d714ee1ecf1c5683a65
c236a81f3193dc27d26d45001ec6c5f158840887
13c78d93f9e8bd8912ca7d1d54314bd7fc6f36e6
f14bb1e945e3ce116ee5ecc0d7e9ee3e04ca3ed9
1b184750cd18918722c04be4778a694e2c8a7e73
f1132944c06f6d60ec0d576a1272bb0b9dbc2024
44784516104c46654107c1f9731186350f69cce1
6041e68c9849b5a368e87da7077b11c0242c6db6
3d98abc754c976dad43014c3ff3952ea080320e9
454c64cb6b7a52d3925bc19c350ae0090f0bd9ae
9726c902cbdf1f3699a299398bfa195ab5ff92c8
f67b72424c618a6e665bb3cac2e8c8f01d2cd599
442f5df8ff7ad1d5e42f89f4c6f8603ea08f2042
ebfe23fb7282e83e5a21619da434b730f5b71a25
3dc02d15b528a185b71fbd16b4720546aae02037
5ffa9483ece446dc081724af75bfdbeec35119c6
a4f5087ffa7800e656da50947b193b00e31395b7
d984f1566388334b9e0ce1de898396cdad016554
235c8744378dc102e27adc997b6b97eca6f023e7
6bb8a1a451c09f608d922bb23caa611c2abd6097
fecacbac367a52347b8136dd4533ca7ef978aa40
```
Keine binäre Suche mehr möglich! --> no more binary search possible (all did not build and had to be skipped and are from johnny_bit, sorry for that)

 - Darktable Version: 
 - OS: 
 - 
 - 
* darktable version : 3.3.0+215~gfecacbac3
* OS                : Linux - kernel 5.7.15-gentoo
* Distro            : Gentoo Base System release 2.7
* Processor         : Intel(R) Core(TM) i9-9900K CPU @ 3.60GHz
* Memory            : 32 GB (4 x 8 GB) + 5GB Swap
* Graphics card0    : GeForce GTX 1060 6GB
* Graphics card1    : GeForce RTX 2070 SUPER
* Graphics driver   : nvidia-drivers-440.100-r2
* OpenCL installed  : Yes (opencl-headers-2020.03.13-r1)
* OpenCL activated  : Yes
* Xorg              : xorg-server-1.20.8-r1
* Desktop           : KDE 5
* GTK+              : gtk+-3.24.22
* gcc               : x86_64-pc-linux-gnu-9.3.0
* cflags            : CFLAGS="-march=native-O2-mtune=native-pipe"
* CMAKE_BUILD_TYPE  : "Release"

@dterrahe maybe you also have an idea?

dterrahe:
fixes #6165

@johnny-bit please test thoroughly; I did a fair bit of culling so probably broke some things. That needed to be broken so they can be fixed. The whole preset generating/using code could use a finer comb still to see if it has more potential array overruns.

I don't have any raws that show a finetuning slider or trigger the cygm view. Would you know where I could download some for testing?

Introspection is not dynamic, so the history tooltip will always interpret the 4 coeffs as rgbg2.

I'm working on a separate cleanup patch which (hopefully) resolves some issues with init/reload_defaults/gui_update separation. That will allow moving a few more things to reload_defaults so they are only done once for each image. This patch already moves the preset generation. Please test if this causes any issues.

### DIFF
--- a/src/iop/temperature.c
+++ b/src/iop/temperature.c
@@ -34,6 +34,7 @@
 #include "control/control.h"
 #include "control/conf.h"
 #include "develop/develop.h"
+#include "develop/imageop_gui.h"
 #include "develop/imageop_math.h"
 #include "develop/tiling.h"
 #include "dtgtk/expander.h"
@@ -65,14 +66,14 @@ DT_MODULE_INTROSPECTION(3, dt_iop_temperature_params_t)
 #define DT_IOP_TEMP_USER 2
 #define DT_IOP_TEMP_D65 3
 
-//storing the last picked color (if any)
-static float old[4] = { 0.0f, 0.0f, 0.0f, 0.0f };
-
 static void gui_sliders_update(struct dt_iop_module_t *self);
 
 typedef struct dt_iop_temperature_params_t
 {
-  float coeffs[4];
+  float red;    // $MIN: 0.0 $MAX: 8.0
+  float green;  // $MIN: 0.0 $MAX: 8.0
+  float blue;   // $MIN: 0.0 $MAX: 8.0
+  float g2;     // $MIN: 0.0 $MAX: 8.0 $DESCRIPTION: "emerald"
 } dt_iop_temperature_params_t;
 
 typedef struct dt_iop_temperature_gui_data_t
@@ -80,9 +81,6 @@ typedef struct dt_iop_temperature_gui_data_t
   GtkWidget *scale_k, *scale_tint, *coeff_widgets, *scale_r, *scale_g, *scale_b, *scale_g2;
   GtkWidget *presets;
   GtkWidget *finetune;
-  GtkWidget *box_enabled;
-  GtkWidget *label_disabled;
-  GtkWidget *stack;
   GtkWidget *buttonbar;
   GtkWidget *colorpicker;
   GtkWidget *btn_asshot; //As Shot
@@ -91,9 +89,10 @@ typedef struct dt_iop_temperature_gui_data_t
   GtkWidget *coeffs_expander;
   GtkWidget *coeffs_toggle;
   int preset_cnt;
-  int preset_num[50];
+  int preset_num[54];
   double daylight_wb[4];
   double mod_coeff[4];
+  float mod_temp, mod_tint;
   double XYZ_to_CAM[4][3], CAM_to_XYZ[3][4];
   int colored_sliders;
   int blackbody_is_confusing;
@@ -134,14 +133,24 @@ int legacy_params(dt_iop_module_t *self, const void *const old_params, const int
     dt_iop_temperature_params_v2_t *o = (dt_iop_temperature_params_v2_t *)old_params;
     dt_iop_temperature_params_t *n = (dt_iop_temperature_params_t *)new_params;
 
-    n->coeffs[0] = o->coeffs[0];
-    n->coeffs[1] = o->coeffs[1];
-    n->coeffs[2] = o->coeffs[2];
-    n->coeffs[3] = NAN;
+    n->red = o->coeffs[0];
+    n->green = o->coeffs[1];
+    n->blue = o->coeffs[2];
+    n->g2 = NAN;
 
     return 0;
   }
   return 1;
+}
+
+static inline void _temp_params_from_array(dt_iop_temperature_params_t *p, const double a[4])
+{
+  p->red = a[0]; p->green = a[1]; p->blue = a[2]; p->g2 = a[3];
+}
+
+static inline void _temp_array_from_params(double a[4], const dt_iop_temperature_params_t *p)
+{
+  a[0] = p->red; a[1] = p->green; a[2] = p->blue; a[3] = p->g2;
 }
 
 static int ignore_missing_wb(dt_image_t *img)
@@ -198,39 +207,11 @@ int default_colorspace(dt_iop_module_t *self, dt_dev_pixelpipe_t *pipe, dt_dev_p
   return iop_cs_RAW;
 }
 
-static gboolean _set_preset_camera(GtkAccelGroup *accel_group, GObject *acceleratable, guint keyval,
-                                   GdkModifierType modifier, gpointer data)
-{
-  dt_iop_module_t *self = data;
-  dt_iop_temperature_gui_data_t *g = self->gui_data;
-  dt_bauhaus_combobox_set(g->presets, DT_IOP_TEMP_AS_SHOT);
-  return TRUE;
-}
-
-static gboolean _set_preset_camera_neutral(GtkAccelGroup *accel_group, GObject *acceleratable, guint keyval,
-                                           GdkModifierType modifier, gpointer data)
-{
-  dt_iop_module_t *self = data;
-  dt_iop_temperature_gui_data_t *g = self->gui_data;
-  dt_bauhaus_combobox_set(g->presets, DT_IOP_TEMP_D65);
-  return TRUE;
-}
-
 static gboolean _set_preset_spot(GtkAccelGroup *accel_group, GObject *acceleratable, guint keyval,
-                                 GdkModifierType modifier, gpointer data)
+                                 GdkModifierType modifier, dt_iop_module_t *self)
 {
-  dt_iop_module_t *self = data;
   dt_iop_temperature_gui_data_t *g = self->gui_data;
   dt_bauhaus_combobox_set(g->presets, DT_IOP_TEMP_SPOT);
-  return TRUE;
-}
-
-static gboolean _set_preset_user(GtkAccelGroup *accel_group, GObject *acceleratable, guint keyval,
-                                 GdkModifierType modifier, gpointer data)
-{
-  dt_iop_module_t *self = data;
-  dt_iop_temperature_gui_data_t *g = self->gui_data;
-  dt_bauhaus_combobox_set(g->presets, DT_IOP_TEMP_USER);
   return TRUE;
 }
 
@@ -241,12 +222,13 @@ void init_key_accels(dt_iop_module_so_t *self)
   dt_accel_register_slider_iop(self, FALSE, NC_("accel", "red"));
   dt_accel_register_slider_iop(self, FALSE, NC_("accel", "green"));
   dt_accel_register_slider_iop(self, FALSE, NC_("accel", "blue"));
+  dt_accel_register_slider_iop(self, FALSE, NC_("accel", "emerald"));
   dt_accel_register_combobox_iop(self, FALSE, NC_("accel", "presets"));
 
-  dt_accel_register_iop(self, TRUE, NC_("accel", "preset/as shot"), 0, 0);
-  dt_accel_register_iop(self, TRUE, NC_("accel", "preset/camera reference"), 0, 0);
-  dt_accel_register_iop(self, TRUE, NC_("accel", "preset/from image area"), 0, 0);
-  dt_accel_register_iop(self, TRUE, NC_("accel", "preset/user modified"), 0, 0);
+  dt_accel_register_iop(self, FALSE, NC_("accel", "preset/as shot"), 0, 0);
+  dt_accel_register_iop(self, FALSE, NC_("accel", "preset/camera reference"), 0, 0);
+  dt_accel_register_iop(self, FALSE, NC_("accel", "preset/from image area"), 0, 0);
+  dt_accel_register_iop(self, FALSE, NC_("accel", "preset/user modified"), 0, 0);
 }
 
 void connect_key_accels(dt_iop_module_t *self)
@@ -258,22 +240,15 @@ void connect_key_accels(dt_iop_module_t *self)
   dt_accel_connect_slider_iop(self, "red", GTK_WIDGET(g->scale_r));
   dt_accel_connect_slider_iop(self, "green", GTK_WIDGET(g->scale_g));
   dt_accel_connect_slider_iop(self, "blue", GTK_WIDGET(g->scale_b));
-  dt_accel_connect_slider_iop(self, "green2", GTK_WIDGET(g->scale_g2));
+  dt_accel_connect_slider_iop(self, "emerald", GTK_WIDGET(g->scale_g2));
   dt_accel_connect_combobox_iop(self, "presets", GTK_WIDGET(g->presets));
 
-  GClosure *closure;
-
-  closure = g_cclosure_new(G_CALLBACK(_set_preset_camera), (gpointer)self, NULL);
-  dt_accel_connect_iop(self, "preset/as shot", closure);
-
-  closure = g_cclosure_new(G_CALLBACK(_set_preset_camera_neutral), (gpointer)self, NULL);
-  dt_accel_connect_iop(self, "preset/camera reference", closure);
-
-  closure = g_cclosure_new(G_CALLBACK(_set_preset_spot), (gpointer)self, NULL);
+  GClosure *closure = closure = g_cclosure_new(G_CALLBACK(_set_preset_spot), (gpointer)self, NULL);
   dt_accel_connect_iop(self, "preset/from image area", closure);
 
-  closure = g_cclosure_new(G_CALLBACK(_set_preset_user), (gpointer)self, NULL);
-  dt_accel_connect_iop(self, "preset/user modified", closure);
+  dt_accel_connect_button_iop(self, "preset/as shot", GTK_WIDGET(g->btn_asshot));
+  dt_accel_connect_button_iop(self, "preset/camera reference", GTK_WIDGET(g->btn_d65));
+  dt_accel_connect_button_iop(self, "preset/user modified", GTK_WIDGET(g->btn_user));
 }
 
 /*
@@ -402,7 +377,7 @@ static cmsCIEXYZ temperature_tint_to_XYZ(double TempK, double tint)
 }
 
 // binary search inversion
-static void XYZ_to_temperature(cmsCIEXYZ XYZ, double *TempK, double *tint)
+static void XYZ_to_temperature(cmsCIEXYZ XYZ, float *TempK, float *tint)
 {
   double maxtemp = DT_IOP_HIGHEST_TEMPERATURE, mintemp = DT_IOP_LOWEST_TEMPERATURE;
   cmsCIEXYZ _xyz;
@@ -460,12 +435,13 @@ static void temp2mul(dt_iop_module_t *self, double TempK, double tint, double mu
   xyz2mul(self, xyz, mul);
 }
 
-static cmsCIEXYZ mul2xyz(dt_iop_module_t *self, const float coeffs[4])
+static cmsCIEXYZ mul2xyz(dt_iop_module_t *self, const dt_iop_temperature_params_t *p)
 {
   dt_iop_temperature_gui_data_t *g = (dt_iop_temperature_gui_data_t *)self->gui_data;
 
   double CAM[4];
-  for(int k = 0; k < 4; k++) CAM[k] = coeffs[k] > 0.0f ? 1.0 / coeffs[k] : 0.0f;
+  _temp_array_from_params(CAM, p);
+  for(int k = 0; k < 4; k++) CAM[k] = CAM[k] > 0.0f ? 1.0 / CAM[k] : 0.0f;
 
   double XYZ[3];
   for(int k = 0; k < 3; k++)
@@ -480,9 +456,9 @@ static cmsCIEXYZ mul2xyz(dt_iop_module_t *self, const float coeffs[4])
   return (cmsCIEXYZ){ XYZ[0], XYZ[1], XYZ[2] };
 }
 
-static void mul2temp(dt_iop_module_t *self, float coeffs[4], double *TempK, double *tint)
+static void mul2temp(dt_iop_module_t *self, dt_iop_temperature_params_t *p, float *TempK, float *tint)
 {
-  XYZ_to_temperature(mul2xyz(self, coeffs), TempK, tint);
+  XYZ_to_temperature(mul2xyz(self, p), TempK, tint);
 }
 
 /*
@@ -786,7 +762,10 @@ void commit_params(struct dt_iop_module_t *self, dt_iop_params_t *p1, dt_dev_pix
     return;
   }
 
-  for(int k = 0; k < 4; k++) d->coeffs[k] = p->coeffs[k];
+  d->coeffs[0] = p->red;
+  d->coeffs[1] = p->green;
+  d->coeffs[2] = p->blue;
+  d->coeffs[3] = p->g2;
 
   // 4Bayer images not implemented in OpenCL yet
   if(self->dev->image_storage.flags & DT_IMAGE_4BAYER) piece->process_cl_ready = 0;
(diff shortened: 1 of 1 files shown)


src/iop/temperature.c
```diff
@@ -91,9 +89,8 @@ typedef struct dt_iop_temperature_gui_data_t
   GtkWidget *coeffs_expander;
   GtkWidget *coeffs_toggle;
   int preset_cnt;
-  int preset_num[50];
+  int preset_num[54];
   double daylight_wb[4];
-  double mod_coeff[4];
```
johnny-bit, line 96: removing this broke "user modified" functionality...

johnny-bit (changes requested):
There's some stuff in testing...

### channel coeffs
when you have expanded channel coeffs expander and you open up image in darkroom here's what you see:
Zrzut ekranu z 2020-09-02 19-58-05
but if you cycle it to proper color mode it starts to look ok:
Zrzut ekranu z 2020-09-02 19-58-31

probably missing clear stops somewhere.

### user modified
I don't really know in which part of code, but this breaks "user modified" functionality. Here's how it should work:
1. user sets custom params
2. user checks preset, say... "camera reference"
3. user wants to get back to previous settings via "user modifed".

check how #4676 done it :)

### finetuning

well - it works ;) I don't really know why [54] instead of [50] (but then it could be dynamic for all I care ;) 

for testing: get 1 raw from Canon EOS 77D and 1 raw from Nikon D750 from raw.pixls.us should work :)

I haven't seen CYGM sensor raw at all (and so I couldn't really test it too) neither have i seen RGBE one :/

For disabled WB get raw from leica monochrome - that too seem to work just fine

dterrahe:
Thanks for the thorough testing @johnny-bit !

I completely misunderstood _user modified_ (and thought the extra copy was for the added accuracy of doubles). Some of the [4] arrays are doubles, whereas params is float, so the copy function i introduced doesn't work everywhere.

I'll look at the slider stops initialisation tomorrow. Probably to do with the reload_defaults issues i mentioned, so night temporarily have to go into gui_update.

I tested with a b&w Leica file. I'll get some canon and nikon files (normally testing on sony). Now I'm curious who's using cygm...

johnny-bit:
check 
```
<LebedevRI> Canon PowerShot G1, Nikon E5700 :)
```

:)

dterrahe:
Slightly broken with the E5700 file; I'll fix all issues tomorrow.

johnny-bit:
If you'd like to, you can check what @dtorop did in non merged PR #5726 to alleviate issues like #5863 or #5725 ;) (but for me those issues aren't problem, it's all just visual sugar)

dterrahe:
> If you'd like to, you can check what @dtorop did in non merged PR #5726 to alleviate issues like #5863 or #5725 ;) 

@dtorop if you like, I can incorporate those fixes in my changes here? They make sense and it would save you the trouble of rebasing.

I'm wondering if I can change the doubles to floats? in params only floats are saved so it seems unnecessary to double the accuracy for temporary storage. Some of the conversion functions might benefit from internally calculating using doubles? 

> (but for me those issues aren't problem, it's all just visual sugar)

That is funny when the same can be said of your whole monster patch that took months of work! ;-)
 Or was that all sucralose?

johnny-bit:
> I'm wondering if I can change the doubles to floats?

I think so...

> Some of the conversion functions might benefit from internally calculating using doubles?

maaaaybe... but highly doubt it...

> That is funny when the same can be said of your whole monster patch that took months of work! ;-)

Well.. it IS visual sugar :)

dterrahe:
OK, second attempt. Restructured a lot more to achieve hopefully the same effect with a simpler structure. I sincerely do hope johnny-bit doesn't find as many problems this time!

Incorporated the changes in @dtorop's patch. Agree with him that this is not complete solution, as there isn't one; the temp/tint won't wander while moving them, but going out and coming back in, they may have completedly changed due to the inherent instability of the solution. People will still be confused, but nothing to be done about that...

I hope the calls to set_feedback for the sliders are as intended now, i.e. no "feedback" in either of the color modes. The existing code was a bit confusing/contradictory so please let me know if I got it wrong. Fixes first issue in https://github.com/darktable-org/darktable/pull/6183#pullrequestreview-481058339

I noticed an existing "bug" that sometimes a camera preset with little finetuning is recognised as  another preset with a large adjustment. Or not at all (and showing as user modified. I'm guessing this is due to instability in the fine tune calculation? Is it strange that sometimes no preset is found at all? I have treated that code as black box for now; not tried to understand it as my brain is already hurting from the gui part :-)

The 4th slider, in CYGM mode, does not get colored at all. Are the others making sense in that case? If only rgb is supported, maybe the coloring should be disabled for other cases?

dterrahe:
Thinking out loud here, but with the button bar above the dropdown, it feels a little to me as if the dropdown provides extra options for just the camera reference button (the rightmost one). The dropdown label "setting" feels very far away to the left, disconnected.
image

I'm not saying that reversing the order looks better, but maybe slightly less confusing?
image

I have no strong opinions! (was just playing around)

Other question; at the moment, user modified gets updated in gui_update, which means every time you go to a different point in history as well. Is that optimal? Maybe, if it wasn't changed manually, it should refer to when the picture was openend? (i.e. reload_defaults)

elstoc:
@dterrahe I think the buttons look better at the bottom.

johnny-bit:
> I hope the calls to set_feedback for the sliders are as intended now, i.e. no "feedback" in either of the color modes. The existing code was a bit confusing/contradictory so please let me know if I got it wrong. Fixes first issue in #6183 (review)

the `feedback` is a fill param of the slider triangle. so feedback=0 means fill=0. This could be expressed better ;)

> I noticed an existing "bug" that sometimes a camera preset with little finetuning is recognised as another preset with a large adjustment.

Sorta user fault sorta preset fault. All presets are in `wb_coeffs.c`. In case camera has presets with finetuning and finetuning ranges overlap there's no sane way to know you meant "shadow with -1 finetine" instead of "cloudy with +9" if r/g/b coeffs are the same. Additionally some finetuning presets are half-assed, meaning users created -6,0,+6 and "have fun interpolating" :/ I too didn't touch interpolation. But I did touch finetuning calculation, because previously finetuning was hardcoded -9..+9 when some presets were available only in -3..+3 (this also allows nikon owners to have -12..+12 finetuning :smile: )

> The 4th slider, in CYGM mode, does not get colored at all. Are the others making sense in that case? If only rgb is supported, maybe the coloring should be disabled for other cases?

No problem there. The temp/tint sliders should be colored if enabled, but for cygm sliders colors make no sense because both rgb sliders coloring options are pure rgb magic

> Thinking out loud here, but with the button bar above the dropdown, it feels a little to me as if the dropdown provides extra options for just the camera reference button (the rightmost one). The dropdown label "setting" feels very far away to the left, disconnected.

This is weird as IMO butons are additional layer of gui sugar ;) plus I bet those buttons still look out of place in the bottom when there's no finetuning ;) Initially (as you can see in my PR) I made them in a grid next to temp/tint sliders which for some made them look as if they related to temp/tint and not to wb setting/preset. I've got suggestion to make them into a line like toolbar and toolbars are usually top-aligned to area they relate to so that's where I put them. Plus I specifically made setting to show/hide button bar ;)
No strong feelings but I prefer top aligned. Plus "setting" combo looks mighty fine in elstoc-proof mode where rightside panel is squeezed so tight it leaks juices ;)
I think that perhaps Moy had good idea to use thin line to separate those. I wonder if there's such element available to use...

> Other question; at the moment, user modified gets updated in gui_update, which means every time you go to a different point in history as well. Is that optimal? Maybe, if it wasn't changed manually, it should refer to when the picture was openend? (i.e. reload_defaults)

Dunno rly. Let's check the logic and make sure it follows logic ok? If I understood Hanno's intentions it was like this:
1. image opens up. user modified is "nothing"
2. user manually adjusts stuff - that's new user modified
3. user switches to named preset or color picker (however many timest that can be), user modified is NOT modified
4. user switches to user modifed, receives settings as set in point 2 (eg. last known user-manipulated value)

the less code gets executed the better so if you can make it follow that logic, then It's fine by me :)

> I sincerely do hope johnny-bit doesn't find as many problems this time!

What problems? I found literally 2 small problems! That's nothing. We'll see tomorrow if I can find anything ;)

dterrahe:
Changes:
User modified only gets initialised the first gui_update, not afterwards, for example when switching through history
rgb sliders only get colored _when they are rgb_
buttons cannot be switched off by clicking on them (what would that mean?) only by changing to another setting. Accept colorpicker
accelerators for the presets now use  dt_accel_connect_button_iop. It is shorter, and the tooltip shows the shortcut.
This doesn't work for colorpicker (yet) because it responds to button-press-event (because it wants the event modifier mask). Now that I've implemented dt_key_modifier_state to catch if ctrl is pressed, that could probably be changed. But that would require a lot of (cross-platform) testing, so not something I want to do in this context.

Two issues I've noticed with accelerators; somehow, after switching into darkroom, they don't work until moving to another image. No idea why yet. I'd like to simplify accelerators anyway, so they don't get enabled only depending on if a module is in the active subset. This annoys the hell out of me every time I start from a clean profile (with _subset: default modules_ because seemingly randomly some shortcuts don't work even though the module might be enabled).

Second: the white balance shortcuts for the 4 basic settings are called "preset/as shot"/"preset/camera reference" etc. The combobox is called "setting" (even though the tooltip calls it preset). I think setting causes less confusion with module presets? 

@elstoc you're the shortcut man; please check that I didn't break anything (except for the issue I mentioned above, which might be broken for all buttons?)

elstoc:
> @elstoc you're the shortcut man; please check that I didn't break anything (except for the issue I mentioned above, which might be broken for all buttons?)

All of the register function calls for the buttons pass in TRUE as the second parameter. This means that the accel is local to the module and only initialised when the module has gui focus. I think these should all be set to FALSE.

dterrahe:
> All of the register function calls for the buttons pass in TRUE as the second parameter. This means that the accel is local to the module and only initialised when the module has gui focus. I think these should all be set to FALSE.

Two-out-of-three prarmeters right! Not bad!

OK, fixed and working. Thank you!

src/iop/temperature.c
```diff
@@ -198,55 +207,28 @@ int default_colorspace(dt_iop_module_t *self, dt_dev_pixelpipe_t *pipe, dt_dev_p
...
-
 void init_key_accels(dt_iop_module_so_t *self)
 {
   dt_accel_register_slider_iop(self, FALSE, NC_("accel", "tint"));
   dt_accel_register_slider_iop(self, FALSE, NC_("accel", "temperature"));
   dt_accel_register_slider_iop(self, FALSE, NC_("accel", "red"));
   dt_accel_register_slider_iop(self, FALSE, NC_("accel", "green"));
   dt_accel_register_slider_iop(self, FALSE, NC_("accel", "blue"));
+  dt_accel_register_slider_iop(self, FALSE, NC_("accel", "emerald"));
   dt_accel_register_combobox_iop(self, FALSE, NC_("accel", "presets"));
```
johnny-bit, line 226: since combo is now labeled "settings" can we switch this up too? i forgot about that in mine ;)

src/iop/temperature.c
```diff
@@ -258,22 +240,15 @@ void connect_key_accels(dt_iop_module_t *self)
   dt_accel_connect_slider_iop(self, "red", GTK_WIDGET(g->scale_r));
   dt_accel_connect_slider_iop(self, "green", GTK_WIDGET(g->scale_g));
   dt_accel_connect_slider_iop(self, "blue", GTK_WIDGET(g->scale_b));
-  dt_accel_connect_slider_iop(self, "green2", GTK_WIDGET(g->scale_g2));
+  dt_accel_connect_slider_iop(self, "emerald", GTK_WIDGET(g->scale_g2));
   dt_accel_connect_combobox_iop(self, "presets", GTK_WIDGET(g->presets));
```
johnny-bit, line 244: if we could change "presets" to "settings" to match combo label...

johnny-bit (commented):
So I tried testing it and:

user modified still has a bit of a problem, here's how to replicate:
1. open up image, mess around with params so that it's user modifed.
2. switch to other image
3. switch back to image from point 1.
4. set wb setting to whatever, eg "as shot"
5. set to "user modified" to get your user modified back... see that it's not that working as it should (doesn't change params at all)

The rest seem fine...

dterrahe:
Hm, interesting; user modified now stores the first item in the history stack, which might be the default, rather than the active history item. I'll try tomorrow to overwrite only if module first entered _or_ if the setting was user modified. That would still mean that clicking on a user modified setting in history would overwrite it and switching back to the latest history item would not restore its user modified setting if it is for example as shot. I'm not sure if that would be more useful/predictable than the old behavior where user modified would _always_ be updated in gui_update, even if that means taking settings from for example as shot. I'll try think of anything better.

dterrahe:
@johnny-bit hope user modified now does what you want. It gets updated when you manually move the sliders _or_ when previous user modified settings are reloaded (gui_update; this is either when you switch pictures or _history_ within the same picture).

It gets initialised only the first time you enter darkroom (from whatever settings that picture has; might be from as shot or whatever). So if you switch from one image with user modified to another image with, say, as shot, you can retrieve the previous image's user modified settings, because they won't have been overwritten. This might not be very useful, but at that point there is nothing more useful to reinitialise user modified with (except for the as shot settings; let me know if that's what you prefer). One case where this could be bad is if you move between rgb and cygm pictures. Still, not much harm done and I think it should be kindof obvious to the user what happened.

Also consistently renamed preset(s) -> setting(s). In shortcuts everything is now grouped under _settings_; the 4 individual options and the previous/next from the dynamic accelerator.

src/iop/temperature.c
```diff
@@ -223,12 +223,12 @@ void init_key_accels(dt_iop_module_so_t *self)
...
+  dt_accel_register_combobox_iop(self, FALSE, NC_("accel", "settings"));
 
-  dt_accel_register_iop(self, FALSE, NC_("accel", "preset/as shot"), 0, 0);
-  dt_accel_register_iop(self, FALSE, NC_("accel", "preset/camera reference"), 0, 0);
-  dt_accel_register_iop(self, FALSE, NC_("accel", "preset/from image area"), 0, 0);
-  dt_accel_register_iop(self, FALSE, NC_("accel", "preset/user modified"), 0, 0);
+  dt_accel_register_iop(self, FALSE, NC_("accel", "settings/as shot"), 0, 0);
+  dt_accel_register_iop(self, FALSE, NC_("accel", "settings/from image area"), 0, 0);
+  dt_accel_register_iop(self, FALSE, NC_("accel", "settings/user modified"), 0, 0);
+  dt_accel_register_iop(self, FALSE, NC_("accel", "settings/camera reference"), 0, 0);
```
johnny-bit, line 231: Technicalllyyyy... one might be noisy about it but grouping direct access to settings with shortcuts to settings might look weird. When you open up shortcuts it's like:

settings:
- as shot
- camera reference
- dynamic
- from image area
- next
- previous
- user modifed

It might be absolute nitpicking but separating that into
setting -> [shortcut to predefined setting]
settings -> [combo box handling]
would make this 100% awesome ;)

johnny-bit (approved):
This is all love :) it's all great!

src/iop/temperature.c
dterrahe: I actually think this is great and as it should be; all six of them are about choosing a setting in the drop-down...
Having two groups that actually manipulate the same thing, as before, looked wrong to me.

But if I haven't convinced you, I'll gladly change it back; this is all about making you happy!! After all the work you put in, I don't want you to feel the end result isn't _exactly_ what you wanted.

johnny-bit: > I actually think this is great and as it should be

Then it should be that way! all 6 in one place :)

So this is perfect as is!

src/iop/temperature.c
dterrahe: Ugh. This made me look through the shortcuts dialog and notice that the automatically generated shortcuts for "preset"s sometimes generate spurious sub-sections when there is a / in the preset name. For example under channel mixer preset b w  artifacts boost.

If I ever get round to working on https://github.com/darktable-org/darktable/issues/5063 (all the clean-up I'm doing now is just ground work to not later on have to deal with all the exceptions) getting rid of the /-based accelerator organisation is high on the list of secondary objectives.

dterrahe:
Time for @AxelG-DE to start hammering on this and see if he can still break #6165

dtorop:
> Incorporated the changes in @dtorop's patch. Agree with him that this is not complete solution, as there isn't one; the temp/tint won't wander while moving them, but going out and coming back in, they may have completedly changed due to the inherent instability of the solution. People will still be confused, but nothing to be done about that...

Glad if #5726 changes make it in via this PR. And excited to see all the recent work on temperature iop. I'm only now reading through this PR, a few days late.

Fixing #5863 could be "sugar", for all that @johnny-bit writes. Or simply: the temperature/tint sliders are a leaky abstraction over the RGB sliders.

#5725 still seems to me a legit -- if obscure -- bug, in that it's possible to get to a place where by moving around the temperature slider across the 39xx->4000K gap, the tint slider gets unexpectedly driven to disconcerting places.

AxelG-DE:
> Time for @AxelG-DE to start hammering on this and see if he can still break #6165

Strange, I did not get the notification email, found this just by browsing thru the PRs for leisure...

I will test now ( I wish I could be able to write well designed skripts to inject not-jet-merged PRs. Manual is kinda pain if you don't do it often). Just as a preperation, I tested current status in current master and found, when you use the "setting" instead of "button" it sporadically did what I expected... We will see...

dterrahe:
@johnny-bit at some point mentioned a git alias setting he uses to pull a pr with a simple command. I can't find it now. The github extension to vscode also has this functionality.

AxelG-DE:
I have tested:
- Using buttons (all combinations I thought of in the 'early' morning): **worked**
- Using setting drop-down: **worked**
- Using sliders and mousewheel (allso checked step-size or increments) and then jkump between the diff buttons: **worked**
- defined an accel for temp and tint and then jumped between the different buttons and setting drop-down: **worked**

Anything else I could do?

> Strange, I did not get the notification email

So before someone thinks I am ignorant or whatever, please also ping me on IRC

> @johnny-bit at some point mentioned a git alias setting he uses to pull a pr with a simple command. I can't find it now. The github extension to vscode also has this functionality.

I dunno vscode but I then will has Hubert on IRC about that.

dterrahe:
If you use the "user defined" settings, set something, move between pictures or out and back into darkroom and see if the setting still behaves as you expect it to.
"worked" means; you don't see the flaw with colorpicker anymore that was in #6165 ?

AxelG-DE:
> If you use the "user defined" settings, set something, move between pictures or out and back into darkroom and see if the setting still behaves as you expect it to.
Thanks for guiding, unfortunately on this way I found a flaw... 
What I did:
- change temp to 20.000 and tint to 0.5 (gives me a very reddish something, easy to notice), than:
-- changed between the different buttons and settings: **worked** (see next item for definition)
-- pressed 'space' and 'backspace': **worked** (means color cast is still there and also shown in thumpnail)
-- used mouse to click other images on filmstrip (and back): **worked**
-- reset to camera before I switch to other pic, switch back to test pic and chose "user modified": **worked**
-- set to area before I switch to other pic, switch back to test pic and chose "user modified": **failed** (means the user defined setting is lost and set to WB-area)
-- above behaves similar when you exit to lighttable and back
-- set "cloud" before I switch to other pic, switch back to test pic and chose "user modified": _**failed**_ (means the user defined setting is lost and set to): **worked**

--> _t.b.c. if needed_

> "worked" means; you don't see the flaw with colorpicker anymore that was in #6165 ?
Correct!
Sorry for leaving interpretation room for you

dterrahe:
Thanks again @AxelG-DE !

The scenario you describe, where you use colorpicker and then lose the user defined setting, is indeed not something I had specifically considered, so it is not supported.

Quick thoughts about this; when leaving the darkroom, the area used by the colorpicker is also lost. So if the current values (coming from the "old" colorpicker) were not stored in user modified,  it would not be possible to retrieve them after switching to any of the other modes. So I think the current behavior is probably the best we can have (besides being a natural consequence of the implementation). Being able to save both user defined _and_ colorpicker separately would mean that both would actually need to be stored in the database, which would be a whole new level of complexity. So I'm hoping you are ok with the current behavior.

But many thanks for pointing out a shortcoming that we/I were not even that conscious of because we were thinking backwards from the implementation and expecting it...

If there is nothing else then this should be ready for @TurboGit to consider merging to close the bug.

AxelG-DE:
> Quick thoughts about this; when leaving the darkroom, the area used by the colorpicker is also lost

I can live with that, however, did you see, when I set to "cloudy", it works....

dtorop:
> I wish I could be able to write well designed skripts to inject not-jet-merged PRs. Manual is kinda pain if you don't do it often

Off topic, but "magit forge" looks appealing, for Emacs users. See https://magit.vc/manual/forge/Branching.html#Branching: 

`forge-branch-pullreq: This command creates and configures a new branch from a pull-request, creating and configuring a new remote if necessary.`

johnny-bit:
> Fixing #5863 could be "sugar", for all that @johnny-bit writes. Or simply: the temperature/tint sliders are a leaky abstraction over the RGB sliders.

because it is :D but it's "cool" anyway.

> #5725 still seems to me a legit -- if obscure -- bug, in that it's possible to get to a place where by moving around the temperature slider across the 39xx->4000K gap, the tint slider gets unexpectedly driven to disconcerting places.

I've worked a bit on that with @aurelienpierre but due to lack of time (on my side) and lack of math knowledge (on my side) i haven't worked it in my UX pr for temp. Problem is, there's a disjoin it temp/tint calculation due to disjoin of planckian locus calculation formula right around meeting point of blackbody and daylight-referenced temp. PRoblem is if we would have stable formula driven from blackbody it wouldn't match anything above 4000 K and if we drove everything with daylight reference, then there would be "nothing meaningfull" below 4000 K. So technically since formulas jump around 3900-4000 K then tint has to jump too :/ IMO the jump is not big enough to be concerned about since all that matter are rgb coeffs ;) and also current tint is NOT perpendicular to planckian locus which adds another level of fun. (btw: having tint moving along normal to plackian locus may create additional "vague points" since planckian locus is (on some representations) U-shaped and if we use that U-shaped representation and it's formula to calculate the normal then inside U shape normals will intersect...

dtorop:
Regarding 3900-4000K problem: if the base math/physics for <3900K and >4000K are OK, maybe we simply need a LUT -- or well-crafted function -- for the problem range which has a stable behavior? Math/algorithmic sugar rather than UI sugar...

I haven't looked at this closely, though...

But yes, all this is a side topic to this PR, and the other more key temperature iop changes.

TurboGit:
Works for me, thanks !

PR #6190, opened 2020-09-03
Revert "append default/preset modules to history instead of prepending"

### LINKED ISSUE #6186: Default modules applied on top of copied history stack
**Describe the bug**
In dt 3.3.0+199~gc8474d8f9, when I copy a history stack on an untouched image I get my default modules applied on top of the copied stack. If I touch the image before, so that the default modules are applied and I paste (append mode) the history stack on the image I get the correct behavior. Also in the official release (3.2.1) the default modules are not applied on top.

**To Reproduce**
1. Create a setting for a module (e.g. Exposure EV = +1.0) and auto-apply to your images
2. Edit an image and change the exposure to EV = 0.0
3. Copy-paste your history stack to another image that is untouched
4. Open the image and see that the current EV = +1.0

**Expected behavior**
I expect that the default modules would be applied first and the history stack copied on top and not the other way round and how 3.2.1 works.

**Screenshots**
Can provide if needed.

 - Darktable Version: 3.3.0+199~gc8474d8f9
 - OS: Win10-64

**Additional context**
 - Cannot be reproduced with another version 3.2.1
 - Problem exists only with fresh edits (no history stack)
 - dt 3.3 was compiled on Mingw-64 with gcc10.2 with standard build file

[Corrected typo in step 4]

elstoc:
This reverts commit 8561dbc5d3ae00722efc2bee92d2eaa97af13904.

Resolves #6186

### DIFF
--- a/src/develop/develop.c
+++ b/src/develop/develop.c
@@ -1354,6 +1354,8 @@ void _dev_insert_module(dt_develop_t *dev, dt_iop_module_t *module, const int im
 
 static gboolean _dev_auto_apply_presets(dt_develop_t *dev)
 {
+  // NOTE: the presets/default iops will be *prepended* into the history.
+
   const int imgid = dev->image_storage.id;
 
   if(imgid <= 0) return FALSE;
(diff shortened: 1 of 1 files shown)


elstoc:
This reverts the second commit from #5449, the first of which has already been reverted due to issues with the integration tests.

TBH I might abandon this PR entirely - I can't work out why the integration tests fail and I'm not sure it's important enough to spend much further time on.

TurboGit:
This seems correct indeed, and I had tested this scenario but on my side I cannot reproduce the issue describe in #6186 with current master. I tried with OP scenario and with your scenario @elstoc, in both cases I had the modified exposure on top of the list and not the auto-applied one (either from scene-referred or auto preset). Really strange !

But anyway, I suspected such possible issue and reverting this part seems better to me so I'll merge. Thanks.

PR #6203, opened 2020-09-04
develop: fix scene-referred default modules.

### LINKED ISSUE #6200: filmicrgb: issues with scene-referred default preset
#6097 attempted to stop the change in workflow requiring a restart.

However replacing the auto-apply of the filmicrgb with a default preset does not achieve the same as the original intent for this module.

A simple demonstration with scene-referred workflow (after a restart to simplify the scenario)

1. Discard the history stack of an image (with no in-camera exposure bias set) and load it in the darkroom
2. Click the 'preset' menu on filmicrgb (it does not show the "scene-referred default" in bold so it is not selected)
3. Manually select the "scene-referred default" preset or click 'reset parameters'
4. Examine the history stack: the latest entry is filmicrgb and it shows that the hardness has changed from its default

If you attempt reset parameters on an image with an in-camera exposure bias the white and black relative exposures also change. If you select the 'default' preset on the same image only the hardness changes. The former (what happens when you reset parameters) is how it should work by default.

When I moved exposure to a using a preset for scene-referred workflow I intentionally did not do the same with filmicrgb because the default settings for this module include adjustments based on the exposure bias from the image's Exif data. As far as I'm aware, this cannot be reflected in a preset.

TurboGit:
We cannot use a preset for filmicrgb which is actually dynamically
computing the parameters based on image expose and the in camera
exposuse bias.

Fixes #6200.

### DIFF
--- a/src/develop/develop.c
+++ b/src/develop/develop.c
@@ -1377,6 +1377,25 @@ static gboolean _dev_auto_apply_presets(dt_develop_t *dev)
   const gboolean is_display_referred = strcmp(workflow, "display-referred") == 0;
   g_free(workflow);
 
+  //  Add scene-referred workflow
+  //  Note that we cannot use the a preset for FilmicRGB as the default values are
+  //  dynamically computed depending on the actual exposure compensation
+  //  (see reload_default routine in filmicrgb.c)
+  if(dt_image_is_matrix_correction_supported(image) && is_scene_referred)
+  {
+    for(GList *modules = dev->iop; modules; modules = g_list_next(modules))
+    {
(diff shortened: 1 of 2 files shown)


elstoc:
Does the workflow setting still work without a restart (not sure what part of your PR fixed that)?

TurboGit:
Yes, should still work without a restart.

elstoc:
Ok I'll give it a quick test

elstoc:
Looks good to me

TurboGit:
Thanks for testing !

PR #6205, opened 2020-09-04
database.c: skip maintenance/optimization on :memory: dbs

johnny-bit:
This simply skips db maintenance and optimization in case either library or data (or both) are `:memory:` as any vacuum/optimize on them is pointless and in case where one of dbs is memory, no point in messing around with trying.

### DIFF
--- a/src/common/database.c
+++ b/src/common/database.c
@@ -2996,9 +2996,14 @@ gboolean _ask_for_maintenance(const gboolean has_gui, const gboolean closing_tim
     return shall_perform_maintenance;
 }
 
+static inline gboolean _is_mem_db(const struct dt_database_t *db)
+{
+  return !g_strcmp0(db->dbfilename_data, ":memory:") || !g_strcmp0(db->dbfilename_library, ":memory:");
+}
+
 void dt_database_maybe_maintenance(const struct dt_database_t *db, const gboolean has_gui, const gboolean closing_time)
 {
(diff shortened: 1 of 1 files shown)


src/common/database.c
```diff
@@ -2998,6 +2998,9 @@ gboolean _ask_for_maintenance(const gboolean has_gui, const gboolean closing_tim
 
 void dt_database_maybe_maintenance(const struct dt_database_t *db, const gboolean has_gui, const gboolean closing_time)
 {
+  if(!g_strcmp0(db->dbfilename_data, ":memory:") || !g_strcmp0(db->dbfilename_library, ":memory:"))
```
pitbuster, line 3001: Since the new code is the same for both `dt_database_maybe_maintenance` and `dt_database_optimize`, I would make a unique function to be called in both places.

pitbuster (changes requested):

src/common/database.c
johnny-bit: It's an `if`statement. I could change it to `if(is_db_memory(db)) return;` would that be better?

pitbuster: Exactly!

pitbuster (approved):

pitbuster:
@johnny-bit Thanks!

TurboGit (approved):
Thanks !

Issue #4380, opened 2020-02-27
Hello. I would like to have a darktable to ask me about unmounting a camera if it's not free, and unmount it if I agree.


KasheyV:
Hello. I would like to have a darktable to ask me about unmounting a camera if it's not free, and unmount it if I agree. 

Like it is in Entangle. Freeing me from checking why darktable tethering does not work be perfect.

Regards.

coffeeandspaceships:
I guess this depends on #5407 first?

jenshannoschwalm:
@johnny-bit closing this as done?

johnny-bit:
Yeah, I think so.

@KasheyV - This is somewhat fixed: darktable thanks to @jenshannoschwalm now shows in ui information that camera might be not free and instructs user to unmount it from other application first. It doesn't do "forced-retake" like entangle can since that MIGHT lead to instabilities. The solution is now part of master and will be part of 3.6 release

I'm closing this issue for now. If you have comments/suggestions feel free to comment here (and I'll re-open the issue) or open a new issue :)

PR #6208, opened 2020-09-05
Handle gphoto device in background thread

jenshannoschwalm:
Updating the list of devices to be used by import or tethering is currently done in a job triggered by a button press. Devices currently used by other apps are not visualised.

I already did some approaches on this topic (see my now closed prs #5407 #5212 and find out about related issues)

What is done here and how?

1) For cam control we maintain two lists, a) holds structs of available devices (this has not changed), b) holds structs of devices that are available but can not be used because of other programs locking it

2) Keeping track of the list is done in a background thread raising a signal if any of the lists has changed. /adding or removing of any gphoto device)

The GUI is some changes too.

1) the 'scan for devices' button has gone, it is not necessary any more and we have less clobbering.

2) The visual representation of devices than can be used has not changed

3) Devices currently not available are displayed but don't have an import button or alike.

What is missing?

>  @johnny-bit: i have a question though - is it possible to have fix for #4380 in it too? I browsed entangle code for that and they detect blocked camera and try to eject it (essentially giving controll back to gphoto for further detection) upon user confirmation.

Did not implement this yet. Handling both lists is a starting point but i have no camera supporting tethering so i honestly can't test properly.

### DIFF
--- a/src/common/camera_control.c
+++ b/src/common/camera_control.c
@@ -123,7 +123,7 @@ static void _dispatch_camera_property_value_changed(const dt_camctl_t *c, const 
 static void dt_camctl_camera_destroy(dt_camera_t *cam);
 
 /** Wrapper to asynchronously look for cameras */
-static int _detect_cameras_callback(dt_job_t *job);
+static int _update_cameras_callback(dt_job_t *job);
 
 /* Starts a job to update the camera list */
 void dt_camctl_background_detect_cameras();
@@ -586,7 +586,7 @@ static void _camctl_unlock(const dt_camctl_t *c)
 
 void dt_camctl_background_detect_cameras()
 {
-  dt_job_t *job = dt_control_job_create(&_detect_cameras_callback, "detect connected cameras");
+  dt_job_t *job = dt_control_job_create(&_update_cameras_callback, "update connected cameras");
   if(job)
   {
     dt_control_job_set_params(job, NULL, NULL);
@@ -704,9 +704,11 @@ static gint _compare_camera_by_port(gconstpointer a, gconstpointer b)
   return g_strcmp0(ca->port, cb->port);
 }
 
-static void dt_camctl_detect_cameras(const dt_camctl_t *c)
+static void dt_camctl_update_cameras(const dt_camctl_t *c)
 {
   dt_camctl_t *camctl = (dt_camctl_t *)c;
+  if(!camctl) return;
+
   dt_pthread_mutex_lock(&camctl->lock);
   gboolean changed_camera = FALSE;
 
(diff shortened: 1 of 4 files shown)


johnny-bit:
I'll try to test this soon, as I have fully tethering-capable camera :)

as for #4380 - The "eject" of camera is weird thing when I looked at entangle code, it's like 3-4 different attempts at removing camera and then mounting it (or I haven't digged into code enough). I think the "enough" part would be to inform user that there are cameras that darktable could use but currently can't since they are "mounted by the system or used by other apps" and then just tell user to unmount them using their system and/or stop using other soft that took over the camera ;) So if you have displayed list of not available devices I think having a tooltip about it would be enough :)

src/common/camera_control.c
```diff
@@ -824,8 +841,67 @@ static void dt_camctl_detect_cameras(const dt_camctl_t *c)
     } while(citem && (citem = g_list_next(citem)) != NULL);
   }
 
+  /* check c->locked_cameras in available_cameras */
+  if(c->locked_cameras && g_list_length(c->locked_cameras) > 0)
```
johnny-bit, line 845: i think `g_list_length > 0` is pointless if `c->locked_cameras` is not null, since non-null list is always at least 1 item long, right?

src/common/camera_control.c
```diff
@@ -824,8 +841,67 @@ static void dt_camctl_detect_cameras(const dt_camctl_t *c)
...
+        changed_camera = TRUE;
+      }
+    } while(c_lock_item && (c_lock_item = g_list_next(c_lock_item)) != NULL);
+  }
+ 
   gp_list_unref(available_cameras);
 
+  /* check c->cameras in locked_cameras */
+  if((c->cameras && g_list_length(c->cameras) > 0) &&
+     (c->locked_cameras && g_list_length(c->locked_cameras) > 0))
```
johnny-bit, line 878: i think `g_list_length > 0` is pointless if `c->cameras` is not null, since non-null list is always at least 1 item long, right?

src/common/camera_control.c
jenshannoschwalm: ok

src/common/camera_control.c
jenshannoschwalm: ok

TurboGit:
Looks promising, I'll try to have a look soon.

TurboGit:
Working great ! I have removed the USB cable in capture more or switch off the camera and we're back to lighttable. I have switch on the camera and it was properly listed as available. Switch off and on again... Everything things seems to works well! That's a nice job.

I'll merged now, I have fixed the conflict which was simple and obvious but please double check :)

We probably need some work on the manual.

Issue #6195, opened 2020-09-04
inaccurate dropping of position on map


paolobenve:
When dropping a photo or a group of photos on the map, the resulting position has become inaccurate in 3.2.1 version. Before, you could drop the marker and it was put in the exact position where you drop it. Now the position remains somewhat displaced.

The same inaccuracy is observable when dragging an already positioned marker.

paolobenve:
I think it's a regression in 3.2.1

PR #6213, opened 2020-09-05
remove map center action on filmstrip click

### LINKED ISSUE #6045: Map: Can not change location of an image because map view jumps to old location
**Describe the bug**
It is not possible to set the GPS location of an image in _map_ view if the image already has a location somewhere else on the map assigned. The problem is that as soon as one clicks on the image to start a drag&drop the map view jumps to the old location of the image.

**To Reproduce**
1. Go to _map_ view
2. Click on an image and use drag&drop to assign a location (lets call it A) to it
3. Move the map to somewhere else, A should be out of view now
4.  Click on the previous image and use drag&drop to assign a new location (lets call it B) to it

Step 4 will not work in dt 3.2.1. As soon as one clicks on the image, the map view jumps to location A, so location B is out of reach and can therefore not be assigned.

**Expected behavior**
In step 4 the view should not jump to location A if the click is part of a drag&drop.

 - Darktable Version: 3.2.1
 - OS: Windows 10
 - OpenCL activated

phweyland:
fixes #6045 EDIT and #6195 

- when a the user wants to place a filmstrip image on the map, he must click on it. So the map center action on filmstrip click is counter intuitive and leads to the issue #6045 
- note that F5 key redraws the map centering the images (F5 triggers collection change, I don't know how ...)

### DIFF
--- a/src/dtgtk/thumbtable.c
+++ b/src/dtgtk/thumbtable.c
@@ -968,7 +968,8 @@ static gboolean _event_enter_notify(GtkWidget *widget, GdkEventCrossing *event, 
 static gboolean _event_button_press(GtkWidget *widget, GdkEventButton *event, gpointer user_data)
 {
   dt_thumbtable_t *table = (dt_thumbtable_t *)user_data;
-
+  dt_view_manager_t *vm = darktable.view_manager;
+  dt_view_t *view = vm->current_view;
   const int id = dt_control_get_mouse_over_id();
 
   if(id > 0 && event->button == 1
@@ -978,7 +979,7 @@ static gboolean _event_button_press(GtkWidget *widget, GdkEventButton *event, gp
     dt_view_manager_switch(darktable.view_manager, "darkroom");
   }
   else if(id > 0 && event->button == 1 && table->mode == DT_THUMBTABLE_MODE_FILMSTRIP
-          && event->type == GDK_BUTTON_PRESS
+          && event->type == GDK_BUTTON_PRESS && strcmp(view->module_name, "map")
           && (event->state & (GDK_SHIFT_MASK | GDK_CONTROL_MASK | GDK_MOD1_MASK)) == 0)
   {
     DT_DEBUG_CONTROL_SIGNAL_RAISE(darktable.signals, DT_SIGNAL_VIEWMANAGER_THUMBTABLE_ACTIVATE, id);
@@ -1039,6 +1040,22 @@ static gboolean _event_motion_notify(GtkWidget *widget, GdkEventMotion *event, g
 static gboolean _event_button_release(GtkWidget *widget, GdkEventButton *event, gpointer user_data)
 {
   dt_thumbtable_t *table = (dt_thumbtable_t *)user_data;
+
+  if(table->dragging == FALSE)
+  {
+    // on map view consider click release instead of press
+    dt_view_manager_t *vm = darktable.view_manager;
+    dt_view_t *view = vm->current_view;
+    const int id = dt_control_get_mouse_over_id();
+    if(id > 0 && event->button == 1 && table->mode == DT_THUMBTABLE_MODE_FILMSTRIP
+            && event->type == GDK_BUTTON_RELEASE && !strcmp(view->module_name, "map")
+            && (event->state & (GDK_SHIFT_MASK | GDK_CONTROL_MASK | GDK_MOD1_MASK)) == 0)
+    {
+      DT_DEBUG_CONTROL_SIGNAL_RAISE(darktable.signals, DT_SIGNAL_VIEWMANAGER_THUMBTABLE_ACTIVATE, id);
+      return TRUE;
+    }
+  }
+
   if(table->mode != DT_THUMBTABLE_MODE_ZOOM) return FALSE;
 
   table->dragging = FALSE;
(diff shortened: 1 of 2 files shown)


phweyland:
Thinking a bit more...
There may be a need to center the map on a given image (there may be a lot on the map).
If the click (GDK_BUTTON_PRESS) is not convenient, we could have another action to center the map on the selected image(s).
Which action would be the best ?

phweyland:
> If the click (GDK_BUTTON_PRESS) is not convenient, we could have another action to center the map on the selected image(s).
> Which action would be the best ?

@AlicVB would you have an advice about this ?

phweyland:
The user may want to:
- center the map on the selection (done when one moves from lighttable to map)
- center the map on the filtered collection (done when one presses F5)
Related to that there is the configuration point "only draw images on map that are currently collected and filtered". If not all images in db located on that map are drawn too.

We could have all these 3 commands on map settings. What do you think ?

TurboGit:
> fixes #6045
> 
>     * when a the user wants to place a filmstrip image on the map, he must click on it. So the map center action on filmstrip click is counter intuitive and leads to the issue #6045

At the same time that's a good way to see the location of the image on the map!

F5 is not the same as it try to have all images on the filmstrip on the map.

What about centering the map on the image clicked on button-release only? That means that on button-press no move and we can then drag & drop it on another location.

phweyland:
> What about centering the map on the image clicked on button-release only?

Yes, that's what I was going to try.

Nilvus:
Is this PR could also fix the wrong place where photo is pinned? Recently, photos are place on the right of the cursor as before there where placed right under the cursor. See this issue: #6195

phweyland:
> Is this PR could also fix the wrong place where photo is pinned?

I'll look at it too. Do we know when (or which commit) this has got broken ?

Nilvus:
> > Is this PR could also fix the wrong place where photo is pinned?
> 
> I'll look at it too. Do we know when (or which commit) this has got broken ?

I don't know but I think that appears on spring. And only commit I found on that period that could be related (but maybe not) is https://github.com/darktable-org/darktable/commit/ec7d533cc4af371ae6e955027b514997b1488c7c

Sorry but I can't test that by now more.

Nilvus:
Just a thing: when launching darktable from that PR, I see some errors on Terminal:

```
(darktable:188427): OsmGpsMap-WARNING **: 12:44:51.996: Error getting missing tile

(darktable:188427): OsmGpsMap-WARNING **: 12:44:51.996: Error getting missing tile

(darktable:188427): OsmGpsMap-WARNING **: 12:44:52.041: Error getting missing tile

(darktable:188427): OsmGpsMap-WARNING **: 12:44:52.041: Error getting missing tile

(darktable:188427): OsmGpsMap-WARNING **: 12:44:52.117: Error getting missing tile
```

phweyland:
For some reason the pin had been moved to the center of the thumbnail.
It's now back to the bottom left corner. That fixes #6195.

 

>  I see some errors on Terminal:

Yes, this message seems to be related to zoom + action. Shouldn't bother the user however.

phweyland:
@TurboGit, Pascal, that's ok for review.

TurboGit:
Works for me! As it touch the new lighttable code I'd like @AlicVB review on this.

Nilvus:
@phweyland: indeed, pinning on map is fixed. Thanks.

Nilvus:
#6195 is not linked, so will not be closed when this PR will be merged. Maybe add "fix" to first post before number issue could link the issue.

johnny-bit:
This is great and seems to resolve both mapping problems! (on the other hand osm-gps-map widget is in dire need of new release :/)

TurboGit:
I have linked #6195 .

phweyland:
> on the other hand osm-gps-map widget is in dire need of new release

Could you tell a bit more about this ?

Nilvus:
> > on the other hand osm-gps-map widget is in dire need of new release
> 
> Could you tell a bit more about this ?

On what I understand, I think @johnny-bit refer of actual state of osm-gps-map and that this library hasn't been update since more than a year (and many issues waiting). See there: https://github.com/nzjrs/osm-gps-map

phweyland:
> See there: https://github.com/nzjrs/osm-gps-map

The last tag is 1.1.0. The current version on my system is 1.1.0-7.
Last, the dt CMakeLists file already takes in account 1.1.0 differences.

Maybe there are available functionalities that dt doesn't use yet ?

johnny-bit:
@phweyland last "official" release of osm-gps-map was in 2015. There is long standing issue https://github.com/nzjrs/osm-gps-map/issues/40 that requests a new release. Since 2015 release there were 8 PRs merged, especially https://github.com/nzjrs/osm-gps-map/pull/48

Also there's an outstanding https://github.com/nzjrs/osm-gps-map/pull/63 that deals with Python 3 support (and as you know python2 is getting booted out of distros) so this needs to get merged AND have new release.

That's why I mention osm-gps-map needs a new release :)

AlicVB (approved):
Ideally, thumbtable should work independently of the view, and each "specific" action should be handled in the view itself, but in that case, that would require lot more stuff (adding new signals, ...) and certainly not worth the pain...
So let's live with that ! Thanks for the work !

TurboGit:
Thanks !

PR #6215, opened 2020-09-05
po/collect combo

TurboGit:
Need some GUI work, the combo with all collect values are not readable because the combo window has the size of the widget.

Any GTK expect knowing how to fix that?

### DIFF
--- a/src/develop/blend_gui.c
+++ b/src/develop/blend_gui.c
@@ -1961,7 +1961,7 @@ static GtkWidget *_combobox_new_from_list(dt_iop_module_t *module, const gchar *
   dt_bauhaus_widget_set_label(combo, _("blend"), label);
   gtk_widget_set_tooltip_text(combo, tooltip);
   for(; *list->name; list++)
-    dt_bauhaus_combobox_add_full(combo, _(list->name), DT_BAUHAUS_COMBOBOX_ALIGN_RIGHT,
+    dt_bauhaus_combobox_add_full(combo, list->name, DT_BAUHAUS_COMBOBOX_ALIGN_RIGHT,
                                  GUINT_TO_POINTER(list->value), NULL, TRUE);
 
   return combo;
--- a/data/themes/darktable.css
+++ b/data/themes/darktable.css
@@ -1738,6 +1738,16 @@ treeview#delete-dialog:selected
   color: @plugin_bg;
 }
 
+#bauhaus-combobox *
+{
+    color: shade(@bauhaus_fg, 0.8);
+}
+
+#bauhaus-combobox *:selected
+{
+    color: shade(@bauhaus_fg_selected, 1.2);
+}
+
 /* Notebooks states */
 notebook tab:hover
 {
--- a/data/themes/darktable.css
+++ b/data/themes/darktable.css
@@ -580,6 +580,12 @@ overshoot.right
   margin: 8px 0px;
 }
 
+#left #lib-plugin-ui #lib-collect-entry
+{
+  margin-top: 0;
+  margin-bottom: 4px;
+}
+
 #left #lib-plugin-ui scrolledwindow entry /* this will avoid having too much height on entry items in clone manager without touching other parts of the UI */
 {
   padding: 0px;
--- a/src/bauhaus/bauhaus.c
+++ b/src/bauhaus/bauhaus.c
@@ -1024,6 +1024,7 @@ void dt_bauhaus_combobox_from_widget(dt_bauhaus_widget_t* w,dt_iop_module_t *sel
   d->defpos = 0;
   d->active = -1;
   d->editable = 0;
+  d->scale = 1;
   d->entries_ellipsis = PANGO_ELLIPSIZE_END;
   memset(d->text, 0, sizeof(d->text));
 
@@ -1103,6 +1104,14 @@ int dt_bauhaus_combobox_get_editable(GtkWidget *widget)
   if(w->type != DT_BAUHAUS_COMBOBOX) return 0;
   dt_bauhaus_combobox_data_t *d = &w->data.combobox;
   return d->editable;
+}
+
+void dt_bauhaus_combobox_set_popup_scale(GtkWidget *widget, const int scale)
+{
+  dt_bauhaus_widget_t *w = DT_BAUHAUS_WIDGET(widget);
+  if(w->type != DT_BAUHAUS_COMBOBOX) return;
+  dt_bauhaus_combobox_data_t *d = &w->data.combobox;
+  d->scale = scale;
 }
 
 void dt_bauhaus_combobox_remove_at(GtkWidget *widget, int pos)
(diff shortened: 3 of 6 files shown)


TurboGit:
Ok, new version pushed which have a correct popup display. Now this certainly needs some CSS tweak as the entries in the popup have too much spaces between them. More compact it would be lot better. @Nilvus : can you have a look ?

TurboGit:
Here is how it looks now:

image

elstoc:
I'm not sure CSS will do it. It's a bauhaus popup and not a normal widget so Gtk can't see any deeper than the window in which the text is rendered.

Nilvus:
> I'm not sure CSS will do it. It's a bauhaus popup and not a normal widget so Gtk can't see any deeper than the window in which the text is rendered.

Exactly, I had already check that on same one for masks and, unfortunately, nothing can be changed with CSS on those menus.

Anyway, I suggest you to check how it is coded on masks. Space is better by default. See below:

Nilvus:
Last thing, if you find how to make category bold with Gtk, that would be better (on masks menu too)

elstoc:
Yeh I don't know. I've looked at the code and I can't work out why it's so different here.

TurboGit:
Ok, I'll look at blend modes, thanks for the hint.

elstoc:
Ok so with much investigation I've worked out what the issue is here. The height of the entries in the popup are determined by the height allocated to the combobox widget itself. Because the combobox is packed into a horizontal box its height is determined by the highest thing in that box which is, in this case, the GtkEntry that contains the search field.

I've pushed a branch that fixes this but it's just a suggestion: https://github.com/elstoc/darktable/tree/test6215

Nilvus:
@elstoc: same menu in masks doesn't need any CSS add to have better space. So I think it's more about a GTK only fix.

Nilvus:
And it would be better to have menus consistent between them, so avoid having a specific tweak on one.

elstoc:
> @elstoc: same menu in masks doesn't need any CSS add to have better space. So I think it's more about a GTK only fix.

That's because it doesn't have any other elements in the same horizontal box dictating its allocated height - it's a straight bauhaus combobox on a line by itself (as are most of the ones in the gui). Yes it could possibly be fixed by some GTK alternative widget packing mechanism but I haven't found it yet.

> avoid having a specific tweak on one

This is a fairly unusual combination of widgets all on a line together (bauhaus combobox + entry + drop-down menu) so it might  not be possible to avoid a 'specific tweak'. In fact the drop-down width has already needed some extra tweaking for a similar reason. So it's a question of whether the 'specific tweak' should be a Gtk one or a CSS one. The CSS one is probably simpler (though I don't know how to quantify whether it's 'better').

Anyway I'm happy if you can suggest something better - I was happy just to have found the cause.

Nilvus:
@elstoc: I don't really know how those menus are build and so thought that they were quite the same. But if it's not, so maybe your CSS tweak is a good thing. I don't see anything else possible with CSS and don't know how Gtk works to propose something better.

TurboGit:
@elstoc : Thanks, this looks more like a work around. I have just pushed a new version that use the line_width which is already available and computed base on current font size. This seems like a better fix for the bahaus widget handling.

elstoc:
In terms of the combobox popup that definitely looks like a better fix. 

However, we're still left with the fact that the combobox itself doesn't line up with the entry beside it because it's been expanded to fill the available box height. This might still need some box-packing shenanigans to resolve:

Screenshot_2020-09-06_15-00-24

TurboGit:
@elstoc : Yes, I've seen that and I suppose this can be worked with some CSS. I'll try to have a look.

TurboGit:
@elstoc : In fact you CSS fix does fix this issue, so I'll integrate it.

TurboGit:
@Nilvus : new version with bold text for section labels.

@all : ready for testing again.

elstoc:
All good for me now. 

One thing I've noticed (and bear in mind this is common to all bauhaus comboboxes and not really anything to do with this change) is that the currently-selected option has a hover effect and an ever so slightly different color to the rest of the entries. It might be good if we could highlight the currently-selected option in a more obvious manner (because the popup often hides the combobox itself) and extend the hover effect to all entries (so it's absolutely clear what you're selecting when you click).

TurboGit:
@elstoc : That's indeed an issue that is already on master. But I think I'll try to fix that too as part of this PR. There is no urgency for this one, so let's make it even better :)

johnny-bit:
I've tested and it looks and feels great!

TurboGit:
@elstoc : done, color should be more visible and distinguishable.

johnny-bit:
This is even better! :shipit: !

elstoc:
Yeah that's looking really good now. Much improved.

Nilvus:
Really better that way. Good work. The only downside (and a big one for me) is that now selected collection remains on the right in dropdown menu when selection is made.

On actual master:


With that PR:

elstoc:
I thought there was something off about that combobox (didn't notice it so much because my panel is quite narrow). This makes sense for most bauhaus comboboxes since it's usually the label that sits on the left of the module and the combobox sits on the right. But yeah - looks wrong here.

TurboGit:
@elstoc @Nilvus : I don't think we have a way to change that. Labels on bauhaus comboboxes are aligned on the right and section on the left. The only possible solution would be to "invert" the section (aligned on right) and labels (aligned on left) with a new bauhaus option. If you think this is worth the effort I can work on that...

TurboGit:
Oh, and for me it is not looking that bad :)

elstoc:
It's less the effort and more how much do we want to add to the generic bauhaus functions to cater for this one specific combobox? 

I mean it would be better if we could fix it but it sounds like a pain.

TurboGit:
Ok, found a way to add support for selected text alignment. New version should be ok now.

elstoc:
Yes I think we have a winner there. Looks great.

src/develop/blend_gui.c
```diff
@@ -1961,7 +1961,7 @@ static GtkWidget *_combobox_new_from_list(dt_iop_module_t *module, const gchar *
   dt_bauhaus_widget_set_label(combo, _("blend"), label);
   gtk_widget_set_tooltip_text(combo, tooltip);
   for(; *list->name; list++)
-    dt_bauhaus_combobox_add_full(combo, _(list->name), DT_BAUHAUS_COMBOBOX_ALIGN_RIGHT,
```
dterrahe, line 1964: This one was correct and necessary I'm afraid. A few drop downs in blending are now not being translated.

TurboGit: I don't understand how is this supposed to work, maybe I'm wrong ! If you revert just this the text is translated again?

dterrahe: Yes.

_() is an alias for gettext. It serves two purposes; the scanner looks for a quoted text between the brackets and if it finds it, will add it to the translation list. At runtime, this will simply translate whatever parameter it is fed; that doesn't have to be a literal.

TurboGit: Stupid me, got confused then :( Sorry let me know if you want me to revert this part or if you already have done that.

dterrahe: if you wouldn't mind reverting? it leaves a cleaner  record and I'm in the middle of a large rebase (that's how I came across it) that I don't want to mess up (too  much). Thank you!

TurboGit: Done, thanks for catching!

PR #6217, opened 2020-09-05
exif: fix reading iop-order.

### LINKED ISSUE #6207: Iop Order Issue w/ 2x Profiled Denoise (after 3.2.1 upgrade)
**Describe the bug**
Some pictures that I edited in a previous version of darktable are now clobbered in 3.2.1.  I have not looked at all photos, but all of the ones I looked at closely (in darkroom) have two instances of the denoise module.  The images are now just lines or grids of totally random colors.  If I open the file and select the first instance of denoise in the history stack, then it shows the image fine, but when I select the second instance, it is back to a garbage image.  I have tried to remove the images and re-import them (with xmp), but this did not solve the issue.  I'm not certain what version of darktable I originally edited this image in.  Furthermore, I do not see the second instance of the denoise module shown in darkroom (in applied modules on the right panel).

**To Reproduce**
Load image with the accompanying xmp file.

**Expected behavior**
Image is the same as when I first edited it.

**Screenshots**
Notice that there is no second "denoise" shown on the right side panel (activated modules):
screenshot

When I select the first denoise instance, it seems fine:
first_denoise

But selecting the second makes it bad again:
second_denoise

 - Darktable Version: 3.2.1
 - OS: Ubuntu 18.04.5 LTS
 - OpenCL activated: Yes (problem persists when disabled too)
 - GeForce GTX 650 Ti BOOST/PCIe/SSE2 with "nvidia-driver-450 (open source)".  I've tried a few others without success.

**Additional context**
Darktable is installed from http://download.opensuse.org/repositories/graphics:/darktable/xUbuntu_18.04/
I've attached an example xmp file.  XMP file upload appear to not be supported for github upload, so I've added a ".txt" extension to the xmp file.
[IMG_6793.CR2.xmp.txt](https://github.com/darktable-org/darktable/files/5177513/IMG_6793.CR2.xmp.txt)

TurboGit:
This fix a convoluted case. What is happening here is probably that
forcing xmp write with a version 3.x without editing the image makes
the iop_order_version to 0. A case that was not properly covered as
it should not happen when an image is edited in darkroom.

Fixes a serious bug in the handling of legacy xmp.

Fixes #6207.

### DIFF
--- a/src/common/exif.cc
+++ b/src/common/exif.cc
@@ -2775,7 +2775,9 @@ int dt_exif_xmp_read(dt_image_t *img, const char *filename, const int history_on
 
       if((pos = xmpData.findKey(Exiv2::XmpKey("Xmp.darktable.iop_order_version"))) != xmpData.end())
       {
-        iop_order_version = pos->toLong() == 2 ? DT_IOP_ORDER_LEGACY : DT_IOP_ORDER_V30;
+        //  All iop-order version before 3 are legacy one. Starting with version 3 we have the first
+        //  attempts to propose the final v3 iop-order.
+        iop_order_version = pos->toLong() < 3 ? DT_IOP_ORDER_LEGACY : DT_IOP_ORDER_V30;
         iop_order_list = dt_ioppr_get_iop_order_list_version(iop_order_version);
       }
       else


jenshannoschwalm:
Good here, thanks for fixing!

Spectre5:
Thanks for looking deeper into this!  Much appreciated.  Funny,part of the reason I forced writing the xmp files with 3.0.2 was that I was hoping it would "upgrade" all the xmp files to the latest logic since I knew there had been a lot of growing pains with the iop operations at the start of DT 3.  So I thought doing that was my way of making all my old edits "future proof"!

TurboGit:
@Spectre5 : One fix if you don't want to wait for 3.4 you may want to set the iop-order version to 2 in the XMP. But make sure you have good backup first :)

TurboGit:
Merging!

Spectre5:
Thanks @TurboGit!  I tested a handful of xmp files and changing iop-version to 2 worked on all of them.  I do have back-ups, but I'll just start with a new library for 3.2.1 for new edits and then bring in the old ones after 3.4 is released.

PR #6224, opened 2020-09-06
hide iop header buttons until mouse over

dterrahe:
In the darkroom right hand panel with module headers, the column of three buttons (instance/reset/preset) takes up a large amount of real estate and makes module names hard to read when the panel is narrow. These buttons convey no information (as opposed to the left most button). Thiis PR hides them until the mouse is over the module header and then shows them to be clicked on.

In the screenshot the mouse is over local contrast.
image

### DIFF
--- a/src/develop/imageop.c
+++ b/src/develop/imageop.c
@@ -601,46 +601,19 @@ void dt_iop_init_pipe(struct dt_iop_module_t *module, struct dt_dev_pixelpipe_t 
   dt_iop_commit_params(module, module->default_params, module->default_blendop_params, pipe, piece);
 }
 
-static gboolean _show_hide_header_buttons(GtkWidget *header, GdkEventCrossing *event, gboolean show_buttons, gboolean always_hide)
-{
-  // check if Entry widget for module name edit exists
-  if(gtk_container_get_focus_child(GTK_CONTAINER(header))) return TRUE;
-
-  if(event && (darktable.develop->darkroom_skip_mouse_events ||
-     event->detail == GDK_NOTIFY_INFERIOR ||
-     event->mode != GDK_CROSSING_NORMAL)) return TRUE;
-
-  GList *children = gtk_container_get_children(GTK_CONTAINER(header));
-
-  GList *button;
-  for(button = g_list_last(children);
-      button && GTK_IS_BUTTON(button->data);
-      button = g_list_previous(button))
-  {
-    gtk_widget_set_visible(GTK_WIDGET(button->data), show_buttons);
-    gtk_widget_set_opacity(GTK_WIDGET(button->data), 1.0);
-  }
-  if(GTK_IS_DRAWING_AREA(button->data)) // (de)activate width trigger widget
-    gtk_widget_set_visible(GTK_WIDGET(button->data), !show_buttons && !always_hide);
-
-  g_list_free(children);
-
-  return TRUE;
-}
-
 static gboolean _header_motion_notify_show_callback(GtkWidget *eventbox, GdkEventCrossing *event, GtkWidget *header)
 {
-  return _show_hide_header_buttons(header, event, TRUE, FALSE);
+  return dt_iop_show_hide_header_buttons(header, event, TRUE, FALSE);
 }
 
 static gboolean _header_motion_notify_hide_callback(GtkWidget *eventbox, GdkEventCrossing *event, GtkWidget *header)
 {
-  return _show_hide_header_buttons(header, event, FALSE, FALSE);
+  return dt_iop_show_hide_header_buttons(header, event, FALSE, FALSE);
 }
 
 static gboolean _header_menu_deactivate_callback(GtkMenuShell *menushell, GtkWidget *header)
 {
-  return _show_hide_header_buttons(header, NULL, FALSE, FALSE);
+  return dt_iop_show_hide_header_buttons(header, NULL, FALSE, FALSE);
 }
 
 static void dt_iop_gui_delete_callback(GtkButton *button, dt_iop_module_t *module)
@@ -1053,7 +1026,7 @@ static gboolean _rename_module_key_press(GtkWidget *entry, GdkEventKey *event, d
   {
     g_signal_handlers_disconnect_by_func(entry, G_CALLBACK(_rename_module_key_press), module);
     gtk_widget_destroy(entry);
-    _show_hide_header_buttons(module->header, NULL, TRUE, FALSE); // after removing entry
+    dt_iop_show_hide_header_buttons(module->header, NULL, TRUE, FALSE); // after removing entry
     dt_iop_gui_update_header(module);
 
     return TRUE;
@@ -1098,7 +1071,7 @@ static void _iop_gui_rename_module(dt_iop_module_t *module)
   g_signal_connect(entry, "style-updated", G_CALLBACK(_rename_module_resize), module);
   g_signal_connect(entry, "changed", G_CALLBACK(_rename_module_resize), module);
 
-  _show_hide_header_buttons(module->header, NULL, FALSE, TRUE); // before adding entry
+  dt_iop_show_hide_header_buttons(module->header, NULL, FALSE, TRUE); // before adding entry
   gtk_box_pack_start(GTK_BOX(module->header), entry, TRUE, TRUE, 0);
   gtk_widget_show(entry);
   gtk_widget_grab_focus(entry);
@@ -2118,23 +2091,28 @@ static void header_size_callback(GtkWidget *widget, GdkRectangle *allocation, Gt
 
   GList *children = gtk_container_get_children(GTK_CONTAINER(header));
 
-  int num_buttons = IOP_MODULE_LAST - IOP_MODULE_LABEL - 1;
+  const gint panel_trigger_width = 250;
 
   GList *button = g_list_first(children);
-
   GtkRequisition button_size;
+  gtk_widget_show(GTK_WIDGET(button->data));
   gtk_widget_get_preferred_size(GTK_WIDGET(button->data), &button_size, NULL);
+
+  int num_buttons = 0;
+  for(button = g_list_last(children);
+      button && GTK_IS_BUTTON(button->data);
+      button = g_list_previous(button)) num_buttons++;
 
   gboolean hide_all = FALSE;
   int num_to_unhide = (allocation->width - 2) / button_size.width;
   double opacity = num_to_unhide > 0 ? 1.0 : (double) allocation->width / button_size.width;
   gboolean all_opaque = FALSE;
 
-  if(!g_strcmp0(config, "off"))
+  if(!g_strcmp0(config, "always"))
   {
     num_to_unhide = num_buttons;
   }
-  else if(!g_strcmp0(config, "on"))
+  else if(!g_strcmp0(config, "active"))
   {
     hide_all = TRUE;
   }
@@ -2149,7 +2127,7 @@ static void header_size_callback(GtkWidget *widget, GdkRectangle *allocation, Gt
     if(allocation->width == 1) hide_all = TRUE;
   }
   else
-  {
+  { // these all (un)hide all buttons at the same time
     if(num_to_unhide < num_buttons) num_to_unhide = 0;
     if(allocation->width == 1) hide_all = TRUE;
 
@@ -2159,15 +2137,19 @@ static void header_size_callback(GtkWidget *widget, GdkRectangle *allocation, Gt
     }
     else
     {
-      // "fit"
-      opacity = 1.0;
+      opacity = 1.0; // "fit"
+
+      GdkRectangle total_alloc;
+      gtk_widget_get_allocation(header, &total_alloc);
 
       if(!g_strcmp0(config, "auto"))
       {
-        GdkRectangle total_alloc;
-        gtk_widget_get_allocation(header, &total_alloc);
-
-        if(total_alloc.width < 250) hide_all = TRUE;
+        if(total_alloc.width < panel_trigger_width) hide_all = TRUE;
+      }
+      else if(!g_strcmp0(config, "fade"))
+      {
+        opacity = (total_alloc.width - panel_trigger_width) / 100.;
+        all_opaque = TRUE;
       }
     }
   }
@@ -2199,7 +2181,53 @@ static void header_size_callback(GtkWidget *widget, GdkRectangle *allocation, Gt
 
   GtkAllocation header_allocation;
   gtk_widget_get_allocation(header, &header_allocation);
-  gtk_widget_size_allocate(header, &header_allocation);
+  if(header_allocation.width > 1) gtk_widget_size_allocate(header, &header_allocation);
+}
+
+gboolean dt_iop_show_hide_header_buttons(GtkWidget *header, GdkEventCrossing *event, gboolean show_buttons, gboolean always_hide)
+{
+  // check if Entry widget for module name edit exists
+  if(gtk_container_get_focus_child(GTK_CONTAINER(header))) return TRUE;
+
+  if(event && (darktable.develop->darkroom_skip_mouse_events ||
+     event->detail == GDK_NOTIFY_INFERIOR ||
+     event->mode != GDK_CROSSING_NORMAL)) return TRUE;
+
+  gchar *config = dt_conf_get_string("darkroom/ui/hide_header_buttons");
+  gboolean dynamic = g_strcmp0(config, "always");
+  g_free(config);
+
+  GList *children = gtk_container_get_children(GTK_CONTAINER(header));
+
+  GList *button;
+  for(button = g_list_last(children);
+      button && GTK_IS_BUTTON(button->data);
+      button = g_list_previous(button))
+  {
+    gtk_widget_set_visible(GTK_WIDGET(button->data), show_buttons);
+    gtk_widget_set_opacity(GTK_WIDGET(button->data), 1.0);
+  }
+  if(GTK_IS_DRAWING_AREA(button->data)) // (de)activate width trigger widget
+    gtk_widget_set_visible(GTK_WIDGET(button->data), !show_buttons && !always_hide);
+  else
+  {
+    if(dynamic)
+    {
+      GtkWidget *space = gtk_drawing_area_new();
+      gtk_box_pack_end(GTK_BOX(header), space, TRUE, TRUE, 0);
+      g_signal_connect(G_OBJECT(space), "size-allocate", G_CALLBACK(header_size_callback), header);
+    }
+  }
+
+  g_list_free(children);
+
+  if(!show_buttons && !always_hide)
+  {
+    GdkRectangle fake_allocation = {.width = UINT16_MAX};
+    header_size_callback(NULL, &fake_allocation, header);
+  }
+
+  return TRUE;
 }
 
 GtkWidget *dt_iop_gui_get_expander(dt_iop_module_t *module)
@@ -2222,9 +2250,15 @@ GtkWidget *dt_iop_gui_get_expander(dt_iop_module_t *module)
 
   /* setup the header box */
   g_signal_connect(G_OBJECT(header_evb), "button-press-event", G_CALLBACK(_iop_plugin_header_button_press), module);
+  gtk_widget_add_events(header_evb, GDK_POINTER_MOTION_MASK);
+  g_signal_connect(G_OBJECT(header_evb), "enter-notify-event", G_CALLBACK(_header_motion_notify_show_callback), header);
+  g_signal_connect(G_OBJECT(header_evb), "leave-notify-event", G_CALLBACK(_header_motion_notify_hide_callback), header);
 
   /* connect mouse button callbacks for focus and presets */
   g_signal_connect(G_OBJECT(body_evb), "button-press-event", G_CALLBACK(_iop_plugin_body_button_press), module);
+  gtk_widget_add_events(body_evb, GDK_POINTER_MOTION_MASK);
+  g_signal_connect(G_OBJECT(body_evb), "enter-notify-event", G_CALLBACK(_header_motion_notify_show_callback), header);
+  g_signal_connect(G_OBJECT(body_evb), "leave-notify-event", G_CALLBACK(_header_motion_notify_hide_callback), header);
 
   /*
    * initialize the header widgets
@@ -2293,22 +2327,7 @@ GtkWidget *dt_iop_gui_get_expander(dt_iop_module_t *module)
   for(int i = IOP_MODULE_LAST - 1; i > IOP_MODULE_LABEL; i--)
     if(hw[i]) gtk_box_pack_end(GTK_BOX(header), hw[i], FALSE, FALSE, 0);
 
-  gchar *config = dt_conf_get_string("darkroom/ui/hide_header_buttons");
-  if(g_strcmp0(config, "off"))
-  {
-    GtkWidget *space = gtk_drawing_area_new();
-    gtk_box_pack_end(GTK_BOX(header), space, TRUE, TRUE, 0);
-    g_signal_connect(G_OBJECT(space), "size-allocate", G_CALLBACK(header_size_callback), header);
-
-    gtk_widget_add_events(header_evb, GDK_POINTER_MOTION_MASK);
-    g_signal_connect(G_OBJECT(header_evb), "enter-notify-event", G_CALLBACK(_header_motion_notify_show_callback), header);
-    g_signal_connect(G_OBJECT(header_evb), "leave-notify-event", G_CALLBACK(_header_motion_notify_hide_callback), header);
-
-    gtk_widget_add_events(body_evb, GDK_POINTER_MOTION_MASK);
-    g_signal_connect(G_OBJECT(body_evb), "enter-notify-event", G_CALLBACK(_header_motion_notify_show_callback), header);
-    g_signal_connect(G_OBJECT(body_evb), "leave-notify-event", G_CALLBACK(_header_motion_notify_hide_callback), header);
-  }
-  g_free(config);
+  dt_iop_show_hide_header_buttons(module->header, NULL, FALSE, FALSE);
 
   dt_gui_add_help_link(header, "interacting.html");
 
--- a/src/develop/imageop.c
+++ b/src/develop/imageop.c
@@ -601,6 +601,48 @@ void dt_iop_init_pipe(struct dt_iop_module_t *module, struct dt_dev_pixelpipe_t 
   dt_iop_commit_params(module, module->default_params, module->default_blendop_params, pipe, piece);
 }
 
+static gboolean _show_hide_header_buttons(GtkWidget *header, GdkEventCrossing *event, gboolean show_buttons, gboolean always_hide)
+{
+  // check if Entry widget for module name edit exists
+  if(gtk_container_get_focus_child(GTK_CONTAINER(header))) return TRUE;
+
+  if(event && (darktable.develop->darkroom_skip_mouse_events ||
+     event->detail == GDK_NOTIFY_INFERIOR ||
+     event->mode != GDK_CROSSING_NORMAL)) return TRUE;
+
+  GList *children = gtk_container_get_children(GTK_CONTAINER(header));
+
+  GList *button;
+  for(button = g_list_last(children);
+      button && GTK_IS_BUTTON(button->data);
+      button = g_list_previous(button))
+  {
+    gtk_widget_set_visible(GTK_WIDGET(button->data), show_buttons);
+    gtk_widget_set_opacity(GTK_WIDGET(button->data), 1.0);
+  }
+  if(GTK_IS_DRAWING_AREA(button->data)) // (de)activate width trigger widget
+    gtk_widget_set_visible(GTK_WIDGET(button->data), !show_buttons && !always_hide);
+
+  g_list_free(children);
+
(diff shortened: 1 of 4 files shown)


johnny-bit:
This is IMO good idea (and seems to not break anything in my preliminary testing).

Question: would it be possible (could be another PR once this is merged) to have leftmost buttons present when header width is "big enough"? (please ignore if not possible or too complex, this imo looks great as-is!)

johnny-bit:
OK, I've tested it a bit more and I have some ideas to share:
1. headers should be consistent across UI, so if you're hiding for IOPs, hide for everything too :+1:
2. when using buttons menus, the buttons disappear leaving no connection that this is the one module that we've clicked the menu button for. this is a bit :-1: interaction, I do wonder if that could be changed so that if there's a menu active, then the header buttons are visible
3. maybe a bit unnecessary, but i'd also show buttons if module is open (or being operated on)
4. if having buttons visible if the header is big enough is not possible in general, I'd introduce an option (default true) to enable or disable hiding buttons

That's all from me, and overal huge :+1:

dterrahe:
Not sure which buttons you mean?  The ones to the left of the label (module state & icon) are always there (depending on theme). 

The leftmost of the three on the right (multi-instance one)? If that was always there but the two to the right of it jumped in and out of view, it would jump right and left...

In any case, doing something based on available space (this was my first idea) is tricky, because it would have to respond to resize events and renaming of modules. And in the end I decided it would probably not even look better. I now quite like this; it is much cleaner/peaceful. Except when the  text jumps around when you move the mouse over a narrow panel, but hey, if that's what you want.

johnny-bit:
OK, I'll explain:
Re.1. - Just the ones on the left. So just hide leftmost buttons for all libs/iops creating visual consistency
Re.2. - What I mean - if I open multi-instance menu for one module, all 3 buttons should still be visible, till menu closes (same goes for presets menu). If that's possible it would be great!
Re.3. - I hope that one's clear (eg keep those 3 buttons visible not only on header hover but also when module is expanded)
Re.4. - So, since responding to width is not something necessary (and as you mentioned it looks more peaceful that way) then if you could just introduce an option to have button hiding enabled/disabled (default enabled) that would be great :)

Hope I made myself clear :)

dterrahe:
> 1. headers should be consistent across UI, so if you're hiding for IOPs, hide for everything too 👍

Of course that was my intention, after waiting to see if this got shot down. Because I like consistency too. A lot! But now that I think about it... the other panels all feel a little different. They have fewer buttons and their titles are shorter (1 or 2 exceptions) and there are fewer modules, so it is easier to know where you are. But it can be done of course. I think it needs separate implementations for each case, but haven't really looked at how much can be directly reused. The iop header was very easy because of some changes I made for the inline module name editing a few weeks back. Could port those over to the other panels as well.

> 2. when using buttons menus, the buttons disappear leaving no connection that this is the one module that we've clicked the menu button for. this is a bit 👎 interaction, I do wonder if that could be changed so that if there's a menu active, then the header buttons are visible

Yeah, that is annoying (and maybe not trivial to fix) but it can also be helpful to be able to read the full module name that you are working on. If the whole name is already visible because the panel is wide enough anyway, then it is only annoying :-/

> 3. maybe a bit unnecessary, but i'd also show buttons if module is open (or being operated on)

I'd say no, because you definitely want to be able to read the full title then...

> 4. if having buttons visible if the header is big enough is not possible in general, I'd introduce an option (default true) to enable or disable hiding buttons

I was expecting that to be a requirement, because people don't like change, but now I actually think this is better enough that people should just get used to it!! (just kidding of course. but I do think we should be careful of filling the settings dialog with switches that we think have only one correct setting. It confuses the new users and there's already so much that confuses them and that they need to figure out and try. So let's see what the consensus is, if any).

dterrahe:
Ah, I seem to always be one question behind with my answer. This gets confusing.

> Re.2. - What I mean - if I open multi-instance menu for one module, all 3 buttons should still be visible, till menu closes (same goes for presets menu). If that's possible it would be great!

The issue with this is that currently this functionality is very self-contained; if the mouse is _over this module_ then show things _in this module_. To keep the buttons visible, I think I need to keep track of whether the mouse is in a menu that was called up from this header or whether that is already closed or... And I don't know if I can even monitor all those conditions (might need many more event handlers). I'll experiment a bit but definitely won't promise.

elstoc:
> when using buttons menus, the buttons disappear leaving no connection that this is the one module that we've clicked the menu button for. this is a bit -1 interaction

Don't forget that for the presets menu you can also right-click anywhere on the module to get it to appear (or use a keyboard shortcut) so I don't think this is a big deal really as long as the menu appears in the right place.

> if having buttons visible if the header is big enough is not possible in general, I'd introduce an option (default true) to enable or disable hiding buttons

I don't think this needs a preference setting. It's either better (in which case let's do it) or it's not. I think it is better.

elstoc:
Now that I'm testing it the change to ellipsization looks a bit odd when hovering. Might it look better if the buttons were just overlaid on top of the module name?

dterrahe:
> Now that I'm testing it the change to ellipsization looks a bit odd when hovering. Might it look better if the buttons were just overlaid on top of the module name?

It might be, but as far as I can tell, overlaying is not a thing in GTK :-( I've wished for that many times.

Remember that the thing that's odd is that you want those tiny scrunched up panels! ;-) (and me)

dterrahe:
So by checking if the header is left "normally" or via a grab event (such as a pop-up) I can prevent the buttons from disappearing.
```
  if(darktable.develop->darkroom_skip_mouse_events || event->detail == GDK_NOTIFY_INFERIOR ||
     event->mode != GDK_CROSSING_NORMAL) return TRUE;

```
The problem now becomes that the popup "owns" the buttons and needs to be modified to make them go away again after closing. Otherwise they'll just hang around until another mouse up-and-over erases them again.

So I'll need to dig into pop-up code. If I remember correctly this is not standard gtk, so might not have events defined. My fear now is that it would be doable to erase the buttons if an action is selected in one of the pop-ups (because that would execute some command on the module anyway). But if the pop-up is closed with escape, the module would receive no action. So now the pop-up needs to undertake action itself. Unless there's anything that can be hooked into, this might require changes in the pop-up code.

elstoc:
> Remember that the thing that's odd is that you want those tiny scrunched up panels

I really don't scrunch them up that much. Just enough

dterrahe:
The GtkMenu _does_ provide the necessary callbacks, so 2. is probably going to work (until it causes some other issue). Finishing this tomorrow.

dterrahe:
Done now. 

Needed an extra xalign 0.0 on the label to prevent the left side from jumping around a bit when it gets ellipsized. Now it is quieter; just the end jumping as needed. Hope this satisfies @elstoc 

also slightly fixed the location where the popup menus appear; please check you agree this is subtly better.

was going to implement for lib modules, but then realised that the buttons there actually convey information by being dimmed when for example presets are not available. That would then tell people not to expect right-click within the module to work. Of course, right-click within the module doesn't work anyway at the moment, because it hasn't been implemented presumably, but I'm sure somebody will address that inconsistency...

which made me rethink a thought I had before; it would be useful if the multiinstance icon gave some feedback. Different icons for multiinstance supported and "there exist multiple instances of this module". if/when this gets implemented it should no longer be hidden but be permanently visible. it could then either be moved to the left of the label, or all the way to the right, because otherwise it would jump back&forth when the other two buttons get (un)hidden.

parafin:
> which made me rethink a thought I had before; it would be useful if the multiinstance icon gave some feedback. Different icons for multiinstance supported and "there exist multiple instances of this module". if/when this gets implemented it should no longer be hidden but be permanently visible. it could then either be moved to the left of the label, or all the way to the right, because otherwise it would jump back&forth when the other two buttons get (un)hidden.

Asking out of the interest - how useful is knowing that more than one instance already exists? And also I think one cares if multi-instance is available only when one wants to duplicate an iop, it's not as relevant as enabled status to be shown all the time IMHO.

elstoc:
I agree with @parafin: there's no need to flag that multiple instances of a module are present.

dterrahe:
fair enough. I was thinking of the very far fetched situation where two instances of a module exist and one of them has been manually moved to a completely different part of the pipe, so they are no longer visible side by side. Now if you notice that something is off and you want to adjust it, you could end up fiddling with the wrong instance. How often will this happen? Probably never. But _theoretically_ ...

dterrahe:
other than that, have I satisfied all wishes for this PR? have you tried and succeeded to break it?

what's the feeling; is a configuration switch necessary?

elstoc:
I'm happy enough without a configuration switch. My only comment so far is that the menu buttons should be hidden when renaming modules, regardless of whether the mouse is over the header.

dterrahe:
> I'm happy enough without a configuration switch. My only comment so far is that the menu buttons should be hidden when renaming modules, regardless of whether the mouse is over the header.

Maybe. But then one _has_ to finish the rename (with enter or click outside) to get the button back to, for example, move the module.  So I could see that as being equally annoying as having to move the mouse a little to have the button disappear. I think the current situation is maybe more predictable (even in the instances where it is annoying; predictable annoyance is better than unpredictable where the "solution" needs to be thought of).

All in all not worth the effort and complexity to "solve". But I'm willing to bow to an opposing consensus of course.

johnny-bit:
I'd still introduce a switch since we have even different themes based on whether somebody wants to display icons or not and displaying actionable buttons is IMO bigger deal than IOP icons ;)

While I agree that non-iop header buttons convey a bit more info than iop ones (because iop ones are always on) they perform exact same things as IOP ones on conceptual level and lack of consistency of visibility creates a bit of imbalance in UI IMO. And their visual status is just mouse hover away :D and IMO you need their status info only if you intend to use it :)

Menus location change is imo great idea! I like it very much!

For module renaming i think annoyance that's consistent with normal action is better if solution is too complex for little gain or if there's possible side-effects and unpredictability.

elstoc:
Fair enough but now that you mention clicking a menu while renaming I note that there are some things that you can click outside the box that keep the rename entry active. For example clicking (or ctrl+clicking) on the on/off button or dragging-and-dropping the module. 

Anyway those are beyond the scope of this PR. I think generally though I would prefer it if you _were_ forced to dismiss the name change before doing anything else (which is how the old mechanism worked because it opened a modal window).

dterrahe:
> I'd still introduce a switch since we have even different themes based on whether somebody wants to display icons or not and displaying actionable buttons is IMO bigger deal than IOP icons ;)

I disagree :-) Let's see what the consensus is. (with a veto from @TurboGit , as always ;-) )
To me, the actionable buttons are just there for that; to take action. And then, they are still there. The icons are a visual flourish that some people assign added value to and some people don't. I don't think anybody would _like_ looking at a long column of buttons. But maybe the "jumping" is aggravating enough that it needs to be disableable. I think dt users have thicker skin than that (by necessity) but...

> And their visual status is just mouse hover away :D and IMO you need their status info only if you intend to use it :)

My counterargument was a right-click in the middle of the module to call up the menu that might not exist. Anyway, for now I'm not planning on implementing this unless @TurboGit forces me ;-)

dterrahe:
> I think generally though I would prefer it if you _were_ forced to dismiss the name change before doing anything else

Why? It is not _that_ high risk an operation. worst case you'll have to retype the name again if you messed it up. And I'm sure some people like saving themselves from the extra enter press. Why go out of your way to torture them?

elstoc:
> Why? It is not that high risk an operation

Agreed. It's just personal preference and I'm not enormously bothered either way but clicking away from the entry should always dismiss it (to be consistent), wherever you click, and I suppose that _might_ be easier if you could make it a two-click operation.

TurboGit:
I'm not fond of this one as it makes using the buttons harder.

For example, when the mouse is over the central picture try to reset one module. It is just hard to place the mouse over the button in first try. I always have to adjust left/right or up/down to actually place the mouse over and be able do the reset action.

elstoc:
I've wondered why the reset button is sat in between two menus. It would make more sense for it to be separate (either to the right of the menu buttons or next to the on/off button. But then presumably you'd have the same issue with the other buttons?

TurboGit:
@elstoc : yes this was just an example, but since you don't see the button before  hand you need to adjust when close to where you think it should have been displayed. So that's for all buttons. The mouse movement is really precise and set to reach the target area by your brain, but since the target area is not visible that's far more difficult!

dterrahe:
OK, I sense one vote (the golden vote) for a configuration option. ;-)

There could be a series of options.

0. buttons only visible when mouse over module header (as before)
1. If the scenario you describe happens most with the active module, then maybe the  commit I just pushed, "Show hidden iop header buttons when pointer in body", already helps a little bit. There is a much larger area to hit with the mouse, before you get to the button. Let me know what you think (but it is probably not enough for you).
2. Keep buttons for focused module always visible. This would be on top of 1. and 0. (so if more than 1 module is expanded, you also benefit from 1. This would require hiding buttons when the module loses focus and it probably will be affected by the whole module-focus conundrum that @elstroc always seems to be fighting. Alternatively, this might work surprisingly well :-)
3. Auto mode. Only hide buttons if the panel width is  below a certain limit. Combined with 1. or 2. or those could be separate options (3a/3b). If this were the default, it could make the feature nicely discoverable. And the narrow-panel-crowd is probably the largest customer base anyway.
4. For the discoverability to work really well, 3 should work instantly, rather than requiring leaving and reentering darkroom. For that, all modules would need to be reset on-the-fly, because otherwise you'd only see the effect after swiping over all headers. An additional callback for resizing the panel is needed. If we are doing this anyway, we could also figure out the maximum width of any header name and hide the buttons if less than, say, 75% of that is visible (the longest name will still be recognised if ellipsed a little). Another callback for font changes and also for editing module names.

All-in-all this would make this feature much more complex than I initially envisioned and maybe it won't be popular enough to justify that. In that case, @TurboGit would you be willing to merge this with just a hidden configuration option to switch on 1.?

Nilvus:
I agree with @TurboGit here and I think it's better, by default, to let buttons visible. So I'm more for an option in preferences to set if buttons are always visible or only visible on mouse hover.

sergei18500:
Personally I find that this PR proposed by @dterrahe brings simplification. ( especially for small screens )
I am in favor of the last comment of @Nilvus 
Buttons visible by default with an option in :
preference :
darkroom :
visible header button: yes/no
(only for the 3 buttons on the right)

TurboGit:
Sure  feel that an option will be necessary here.

dterrahe:
Definitely. I'm going to experiment a little with some of the options and see if I'll make it an on/off switch or one with several choices.

dterrahe:
@TurboGit @elstoc I'd like some feedback on this experiment with overlay/opacity to see if it is worth pursuing further and fixing at least two issues it currently has.
This keeps buttons visible at all times, but as overlay, so long module names go underneath, and 50% transparent, so the names are readable.
image
As before, a mouseover over the module will restore full opacity to make it clear button is active and can be clicked (mouse is over local contrast in screenshot).

question is if this gives @TurboGit  enough of a target to aim his mouse at and if at the same time the text is sufficiently readable to make this a solution for narrow panels.

I haven't spent time perfecting this, so at least two issues:
- when editing the name, the buttons have to disappear otherwise the edit area can't be seen.
- even at full opacity the buttons are too light. They should be as bright as an "enabled" module button (white). But obviously this should be done with styles, so the active/inactive states need to get their own names. opacity can be set in styles. Maybe in mouseover the button background should be fully opaque (so the text behind it can't be seen at all anymore/doesn't distract).

with this solution there is no "problem" with jumping ellipsizing labels on mouse over (not that I thought that was a big issue).

what do you think; is this (after polishing) a possible one-size-fits-all solution? if not, it could be one of a few options in settings, if it isn't too hard to make this reconfigurable on-the-fly.

elstoc:
Not a fan of this one I'm afraid. Buttons overlapping with text just looks like a glitch to me.

Nilvus:
I agree with @elstoc

johnny-bit:
I'm not a fan either. While I have no problem with icons looking as if they were disabled, my probelm is when text overlays on top of icons. i mean - the "shadows and highlights" looks like "shadows and high̸̒̀̈́̚͜l̶̪̫̜̗͗̎͒ì̶̧͍̱̓ǵ̸̞̑̅h̵͈̒͝ţ̸̩̹̬̐̏s̷̥͕̓͠͝ͅ"

however that could really work if the text didn't overlap on top of icons.

dterrahe:
> Buttons overlapping with text just looks like a glitch to me.

Fair enough. That's the reaction I was expecting. And definitely the active buttons need their non-transparent background to fix readability. I think at that point, with the right polish, it would be obvious that it is intentional and that it would make long names (much) more readable than ellipsiziing. But it'll still be a matter of taste...

dterrahe:
> however that could really work if the text didn't overlap on top of icons.

then there's no point, because there would be no extra space. This experiment is an attempt to explore path to not 100% of the time use space that could be more usefully showing longer module names, while at the same time dealing with Pascal's requirement that he needs to know where the button will be so he can aim his mouse.

So I'll ditch this (not unhappily; the implementation also gets messy quick) but would be happy to run with any ideas anybody else comes up with.

TurboGit:
I'm no fan of this either.

Nilvus:
Not a fan on last proposal. @dterrahe: why make this so complicated? I really think that just a switch in preferences to activate your first proposal (buttons appears on mouse hover) for those who want that and let by default as it is actually in master, is the simplest and better way to go.

dterrahe:
> why make this so complicated?

@Nilvus am I correct in reading from your reactions above that you will just keep this feature switched off (keep everything as is now)? I don't think you'll have to worry about @TurboGit merging something that takes that option away from you. In that case, I understand that you are not very interested in making the feature work as well as possible. Please feel free to ignore the discussion here.

I'd like to continue gathering ideas from the people that _do_ see the benefit so that maybe I can improve it beyond my initial attempt. Maybe it can be good enough that people who do not have years of getting used to the current way might prefer it. As usual I'm following the @johnny-bit strategy; propose something that is so clearly wrong that people _have_ to respond and tell you what the correct way of doing it is. But so far, in this particular case, it is not working.

elstoc:
I think your first proposal was fine and, to make the reset button easier to hit it would be good if the order were changed to put that button on the far right of the headers (and maybe even make just that button always visible). Personally I don't use the menus as often as I use reset. If you wanted something ultra-configurable you could have the ability to control both the order and default visibility of those three icons :)

johnny-bit:
> As usual I'm following the @johnny-bit strategy; propose something that is so clearly wrong that people have to respond and tell you what the correct way of doing it is.

Oh shit, they discovered me! Damn!

But since I am interested in this discussion, here's quick sum up of what I think about hiding header buttons:
- It's "good", but needs to hide both - for IOPs as well as libs
- It's "good", but needs to be able to be disabled (or enabled if disabled by default) for those who want/don't want it.
- The "transparent" one looks like buttons are disabled and only enabled after hover + that overlaping text is icky so if going "transparent route" you'd have to have non-transparent text background so in case text overlaps buttons, the icons don't affect text readability. But this isn't bad per se...
- I do agree with @elstoc's comment on reset button order, but not neccesarly on "always visible" part - if I want buttons hidden then may them be hidden always! However reordering reset buton comes with downside of unfamiliar ui. Usually menu is furthest right item and any action is to the left of it, which means that having button on far right would mess up order. This isn't bad, but would need a bit of thought. For some people preset menu is useless, for some it's like the most important part of modules. For some multi-instance fans multi instance menu is far more important and having default right click action on module be preset menu is weird for them... but i think reset is universally important for all.

dterrahe:
> * It's "good", but needs to hide both - for IOPs as well as libs

I think libs are different enough already; two buttons vs three, generally shorter names that don't quickly become unreadable  ("recently used collections" proving the rule) and there being fewer of them, in fixed positions. So I don't think they have the same problem I'm trying to fix (long lists of modules with unreadable names so it is impossible to find the correct ones. And I don't feel a different behavior would be problematic. So in first instance, I would not implement this for libs, but if there is an outcry/flood of FR bugs, of course it would not be hard to add there too.

> * It's "good", but needs to be able to be disabled (or enabled if disabled by default) for those who want/don't want it.

We all agree on this. My remaining question is whether there would be a few distinct variants rather than on/off. One of those variants could be "auto", i.e. automatically hide buttons when panel is "narrow" (by some definition to be determined).

> * The "transparent" one looks like buttons are disabled and only enabled after hover + that overlaping text is icky so if going "transparent route" you'd have to have non-transparent text background so in case text overlaps buttons, the icons don't affect text readability. But this isn't bad per se...

This was a now discarded experiment. Feedback was overwhelmingly negative. But there was feedback!

> * I do agree with @elstoc's comment on reset button order, but not neccesarly on "always visible" part - if I want buttons hidden then may them be hidden always! However reordering reset buton comes with downside of unfamiliar ui. Usually menu is furthest right item and any action is to the left of it, which means that having button on far right would mess up order. This isn't bad, but would need a bit of thought. For some people preset menu is useless, for some it's like the most important part of modules. For some multi-instance fans multi instance menu is far more important and having default right click action on module be preset menu is weird for them... but i think reset is universally important for all.

This is probably a completely separate/independent issue (that I'm not particularly bothered about). But I'm willing to provide provocative suggestions! If you are going to make this completely configurable, why not have an option to move reset next to the on/off switch on the left? And why not change the icon to indicate whether all values are at default or not.

johnny-bit:
> I think libs are different enough already

While I agree totally with you on difference between IOPs/LIBs, my standpoint is purely visual one: when buttons for iops are hidden but buttons for lib aren't, then it creates a bit of visual imbance.

> My remaining question is whether there would be a few distinct variants rather than on/off.

I think it got lost with outrage for transparent buttons ;) IMO - few variants are ok to do. What are your current proposed variants?

> But I'm willing to provide provocative suggestions! If you are going to make this completely configurable, why not have an option to move reset next to the on/off switch on the left? And why not change the icon to indicate whether all values are at default or not.

Yeah, needs separate issue :) Otherwise the discussion will overwhelm main theme here: hideable buttons and it's variants :)

Nilvus:
@dterrahe: you were not correct about me. I'm interested in how that feature could come. And make what you want possible is good but we need to keep known behavior for some users (some want to keep some habits). And that's also what @johnny-bit said in a previous comment: https://github.com/darktable-org/darktable/pull/6224#issuecomment-688223096

Anyway, I hadn't understand that you want some other ideas. So, some ideas:
- let actual behavior by default (for legacy like darktable theme instead of added elegant darker one in the new UI last year)
- Add a possibility to hide icons and just show them on mouse hover (your first proposal) on preferences
- On preferences, add (with checkbox?) possibility to choose which icons we want to keep (preset, reset, instance) for those who just use one or two of those ones. And of course, with the options to show icons with legacy behavior or mouse over.

That could be added in preferences setting in same way we have overlays in star icon: select which icons display (preset, reset, instance) and then have a checkbox to activate mouse hover.

dterrahe:
Restructured the whole thing.

Added a few options; off is the default. There should be no changes then, except a few fixes in sub-menu location and hiding buttons while editing module name.

If "off" is the setting when switching to darkroom, none of the required callbacks are installed, so switching to any other option will require reentering the darkroom (not a full restart).
Once the callbacks are installed, switching between options in the settings will work immediately, but existing headers will not get refreshed until they are moused over or, easier, the panel is slightly resized; then they will all immediately show according to the new option.

"auto" could be suitable as an alternative default. Everything will be unchanged until the panel width is reduced. I like it, because it makes the functionality very discoverable. The impact of the other options is seen best while resizing the panel. Not all of them are intended very seriously but I see no harm in keeping all of them. Some additional options could even be considered (for example "dim" could be combined with other options).

If anybody implements functionality to optionally permanently hide specific buttons, the num_buttons calculation needs to be made a little more robust. Assuming that the buttons are not created at all, rather than hidden, the rest of the code might simply work. If they _are_ in the header box, but hidden, then this code might need substantial changes.

dterrahe:
image
"glide" mode.
But you have to see it in movement.

src/develop/imageop.c
```diff
@@ -2063,6 +2112,96 @@ static gboolean _iop_plugin_header_button_press(GtkWidget *w, GdkEventButton *e,
...
+
+  GList *button = g_list_first(children);
+
+  GtkRequisition button_size;
+  gtk_widget_get_preferred_size(GTK_WIDGET(button->data), &button_size, NULL);
+
+  bool hide_all = FALSE;
+  int num_to_unhide = (allocation->width - 2) / button_size.width;
+  double opacity = num_to_unhide > 0 ? 1.0 : (double) allocation->width / button_size.width;
+  bool all_opaque = FALSE;
```
johnny-bit, line 2131: Please change `bool` to `gboolean`.

johnny-bit (changes requested):
When I try to compile i see the following report:

```
/home/johnny/darktable_repos/darktable/src/develop/imageop.c: In function ‘header_size_callback’:
/home/johnny/darktable_repos/darktable/src/develop/imageop.c:2128:3: error: unknown type name ‘bool’
 2128 |   bool hide_all = FALSE;
      |   ^~~~
/home/johnny/darktable_repos/darktable/src/develop/imageop.c:2131:3: error: unknown type name ‘bool’
 2131 |   bool all_opaque = FALSE;
      |   ^~~~
gmake[2]: *** [src/CMakeFiles/lib_darktable.dir/build.make:1239: src/CMakeFiles/lib_darktable.dir/develop/imageop.c.o] Błąd 1
```

it's because you use bool (which isn't standard to C I think) so probably better to use gboolean.

dterrahe:
Ah, betrayed by my c++ history again. Thanks for catching. Strange that the automatic checks didn't flag this.

Easily fixed; gboolean should work. Will push later.

dterrahe:
> When I try to compile

I assume you managed to compile after that fix? Any first impressions?

johnny-bit:
Ah, i forgot about that ;) sorry

The first impression is "this is awesome!"

really, the "glide" mode is freaking perfection IMO.

this is now triple :+1: from me and total :shipit: (shipit) :)

tiny nudge from me would be to react to settings moddified signal and re-draw headers on it. that's basically it.

dterrahe:
> tiny nudge from me would be to react to settings moddified signal and re-draw headers on it. that's basically it.

tiny nudges that kill my morning.

OK, this works immediately upon settings change now. Some of the (less-fired) signals are now always connected, but the helper widget and its size-allocate event only gets set up if the mode is not "always" visible.

Renamed the option and the modes; off -> always (visible), on -> active. Seems like the more logical way round for a new user.
always visible is still the default, of course, but I've added another mode, "fade" which I think is a good candidate to become default.

Please hammer on this and see if you can find the tiny tiny ways in which it can temporarily break. Then agree with me that those don't matter :-) Or give me a simple solution that doesn't require connecting to all the remaining signals.

@TurboGit I think this may be ready to merge, but please wait for green light from @johnny-bit at least.

johnny-bit:
Oh my, I've became an authority on "greenlighting" thingies? Surely you jest.

Anyway - yeah, haven't managed to break it in visible ways. All options are perfect whe way they are, there's nothing I saw immediatelly to be changed... Yep, ready to merge imo ;)

(i won't mention the lib modules buttons, that might be for another pr)

dterrahe:
> (i won't mention the lib modules buttons, that might be for another pr)

I still disagree violently with you on that point! ;-)

dterrahe:
@TurboGit Just to give an indication of what is happening here/you should test:

if mode = "always", then everything should be as before, with a few small fixes:
- positioning of popup menus for multi-instance and presets is now aligned with the bottom-right of the button, rather than pointer (random) or header in general.
- when editing module name, the inline Entry now goes all the way to the end; the buttons are temporarily hidden.

If the setting has always been in "always" mode (in the same darkroom session) then only the enter/leave signals for each header are connected, but those should have no effect.

If another setting is/has been active, then the header gets an additional widget injected which sends a signal that indicates how much space is available for buttons and this is then responded to depending on the mode. This deals correctly with panel width, module name and font sizes changes. Of course if mode is "always" this still shows all buttons all the time.

TurboGit:
I'm not sure we have any options to fit everybody needs :) 8 possibilities for displaying 3 buttons :)

I certainly very like the "dim" option.

All this looks good to me and working fine on my side.

Please can you provide a description of this preference for the user's manual ?

TurboGit:
And thanks a lot for this nice addition!

dterrahe:
> I'm not sure we have any options to fit everybody needs :) 8 possibilities for displaying 3 buttons :)

Yeah, I worry about that too! It is very much a taste thing. But also, some are dynamic (look good _while_ you are changing panel width) others are better static, especially if you hate different opacities.

> I certainly very like the "dim" option.

dim+auto is missing. Do you want it? Only couple of lines.

Point is, the infrastructure was the major work. Would be a pity if people didn't use it because it doesn't do _exactly_ what they want. I was thinking of having a drop down with major categories (off, auto, fit, individual) and then buttons to add "dynamic" and "dim". But some combinations just don't make sense and it would be harder to explain than just give a few nice options and let people try them out. Anyway... glad you like it.

> Please can you provide a description of this preference for the user's manual ?

@elstoc I've seen that there's a _lot_ of work going on, but don't know where. I could post here, but I want to get an idea of the current style and whether I just need to describe the option in settings or whether this should be mentioned/described in the context of resizeable panels? Same for colorbalance layouts.

TurboGit:
> I've seen that there's a lot of work going on, but don't know where. 

Yes I'm doing lot of user manual update at the moment. And @Nilvus has redo the whole preference section. For the preference at hand you want to edit:
`
doc/usermanual/preferences/darkroom_options.xml`

PR #6225, opened 2020-09-06
filmicrgb: remove line breaks in comments

rabauke:
Line breaks in comments break auto-generated code in `introspection_filmicrgb.c`, which leads to untranslated ui strings. 

I think, we should bracket the definition of `dt_iop_*_params_t` types in
```
// clang-format off
...
// clang-format on
```
in all modules to prevent accidental changes of introspection code in source comments.


TurboGit:
Works for me, thanks !

PR #6230, opened 2020-09-07
Support for reading 16-bit (half) float TIFFs

kmilos:
Partly addresses https://redmine.darktable.org/issues/8380 (which seems didn't make it over to github...)

### DIFF
--- a/src/common/imageio_tiff.c
+++ b/src/common/imageio_tiff.c
@@ -46,6 +46,45 @@ typedef struct tiff_t
 } tiff_t;
 
 
+static inline float _half_to_float(uint16_t h)
+{
+  /* see https://en.wikipedia.org/wiki/Half-precision_floating-point_format#Exponent_encoding
+     and https://en.wikipedia.org/wiki/Single-precision_floating-point_format#Exponent_encoding */
+
+  /* TODO: use intrinsics when possible */
+
+  uint32_t f;
+  uint32_t s = (h >> 15) & 0x1;
+  uint32_t e = (h >> 10) & 0x1f;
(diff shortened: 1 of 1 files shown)


kmilos:
A test image can be generated froma  32-bit float TIFF using Imagemagick:

`magick convert test_float.tif -depth 16 -define quantum:format=floating-point [-compress zip -define [tiff:predictor=3]] test_half.tif`

TurboGit:
Looks good to me. Just to be sure, the current master does read those TIF but using lib ImageMagick if dt is compiled with it.

```
[tiff_open] error: Not a supported tiff image format.[rawspeed] (test_half.tif) virtual std::unique_ptr<rawspeed::RawDecoder> rawspeed::RawParser::getDecoder(const rawspeed::CameraMetaData*), line 91: No decoder found. Sorry.
[GraphicsMagick_open] image `test_half.tif' loading
```

This patch makes the loading native and does not require ImageMagic. Correct?

kmilos:
3.2.1 wouldn't read those for me on Windows (skullhead icon in lighttable, error message when double clinking), and I don't know if the Win build comes w/ ImageMagick or not... Sorry I haven't tried master first...

So yes, this makes it "native" (for the read only), just like 32-bit float, relying only on libtiff (+zlib for compressed variants).

The are alternatives like relying on similar code in OpenEXR or rawspeed, but they seem to be exposed through a C++ interface only, and I didn't feel it was necessary to rewrite the entire TIFF reader...

TurboGit:
Ok, thanks for the details. That match my understanding and since this works let's merge. Thanks.

PR #6231, opened 2020-09-07
po/lr import

TurboGit:


src/common/exif.cc
```diff
@@ -452,9 +452,9 @@ static bool _exif_decode_xmp_data(dt_image_t *img, Exiv2::XmpData &xmpData, int
       dt_image_set_xmp_rating(img, stars);
     }
 
-    if(!exif_read) dt_colorlabels_remove_labels(img->id);
     if(FIND_XMP_TAG("Xmp.xmp.Label"))
     {
+      if(!exif_read) dt_colorlabels_remove_labels(img->id);
```
TurboGit, line 457: @phweyland : can you review this part ? see commit log for the why of this change. TIA.

TurboGit:
@phweyland : just in case you missed this one!

phweyland:
Sorry, I had commented but ... not submitted the review

src/common/exif.cc
TurboGit: But in the case at hand. You have a RAW + Lr XMP. You import the picture and the labels coded in Lr XMP are not imported. Any idea why and where to look ?

phweyland: Would you have such a lightroom xmp to share ? I could have a look.

TurboGit: Yes, there is one in the referenced issue:
https://drive.google.com/file/d/1pfoerxv3awnEpe9CvHSJVqPkJvp98PCv/view?usp=sharing

Note that you'll need the first commit on the PR as the XMP is from Camera Raw and not Lightroom. But that's a separate issue.

phweyland: Without "write sidecar file for each image" that works.
So the first phase of import creates an xmp, then import lightroom set the colorlabel, then the nef.xmp is read again, clearing the label set by lightroom.
I continue ...

phweyland: dt_image_write_sidecar_file() is called 4 times for a single image ...

But  in _image_import_internal() we have the sequence:
-  `dt_lightroom_import()`
- `_image_read_duplicates() -> dt_exif_xmp_read -> _exif_decode_xmp_data ->` clear labels
- ` dt_image_synch_all_xmp()` but it is too late to save the lightroom label

So one working solution is to call dt_image_write_sidecar_file() once more just after dt_lightroom_import().

phweyland: my current traces for reference (with the fix):
```
import image (dt_exif_read)
dt_exif_read (_exif_decode_xmp_data) 18 true
import image (dt_exif_xmp_read)
import image (dt_lightroom_import)
dt_lightroom_import 18 dev (nil) (dt_get_lightroom_xmp)
_lrop colorlabel 18 dev (nil)
dt_image_write_sidecar_file (dt_exif_xmp_write) 18
dt_exif_xmp_write (_exif_xmp_read_data) 18
_exif_xmp_read_data 18
dt_image_write_sidecar_file (dt_exif_xmp_write) 18
dt_exif_xmp_write (_exif_xmp_read_data) 18
_exif_xmp_read_data 18
dt_lightroom_import (dt_colorlabels_set_label) 18
dt_image_write_sidecar_file (dt_exif_xmp_write) 18
dt_exif_xmp_write (_exif_xmp_read_data) 18
_exif_xmp_read_data 18
_image_read_duplicates (dt_exif_xmp_read) 18
dt_exif_xmp_read (_exif_decode_xmp_data) 18 false
_exif_decode_xmp_data - clear labels 18
dt_image_write_sidecar_file (dt_exif_xmp_write) 18
dt_exif_xmp_write (_exif_xmp_read_data) 18
_exif_xmp_read_data 18
dt_lightroom_import 18 dev 0x55b837d2ad20 (dt_get_lightroom_xmp)
_lrop colorlabel 18 dev 0x55b837d2ad20
dt_image_write_sidecar_file (dt_exif_xmp_write) 18
dt_exif_xmp_write (_exif_xmp_read_data) 18
_exif_xmp_read_data 18
```

phweyland: > So one working solution is to call dt_image_write_sidecar_file() once more just after dt_lightroom_import().

Another solution would  be not to call _image_read_duplicate() if a lightroom xmp has been found.
Cleaner but just a bit more change.

TurboGit: Good to have a fix and since this is during import and happens once it is not so bad!

TurboGit: > Another solution would be not to call _image_read_duplicate() if a lightroom xmp has been found.

I'm not sure this will make the code easier to read as it is a special case.

phweyland: makes sense

TurboGit: How do you want to proceed? I can merge my two first commits only from this PR and you can propose a PR with your fix. Would that work for you?

phweyland: That's fine for me, but I would not sad if you made that change too !  :)
I'll do it. Thanks

TurboGit: Thanks! This is now merged.

PR #6232, opened 2020-09-07
don't register lightroom and exifEX to exiv2 at init time

### LINKED ISSUE #6222: xmp "lr" property not recognized calling exiv2
Not critical but long lasting issue.

In export metadata config the "Xmp.lr.hierarchicalSubject" tag should be listed in: 
image

The code which extracts the list of known tags is there:
https://github.com/darktable-org/darktable/blob/4f01984998bc9c0a042a5205d0b769501eb44603/src/common/exif.cc#L149

I've opened an issue on [Exiv2](https://github.com/Exiv2/exiv2/issues/1206) but that's work fine on their side, whatever the exiv2 version. That's why I'm opening the issue on dt side now.
Thanks for the support of Robin @clanmills.

phweyland:
fixes #6222

A big thanks to Robin @clanmills  for finding out this (and 4 biers please !)

### DIFF
--- a/src/common/exif.cc
+++ b/src/common/exif.cc
@@ -161,9 +161,6 @@ void dt_exif_set_exiv2_taglist()
 {
   if(exiv2_taglist) return;
 
-  Exiv2::XmpParser::initialize();
-  ::atexit(Exiv2::XmpParser::terminate);
-
   try
   {
     const Exiv2::GroupInfo *groupList = Exiv2::ExifTags::groupList();
@@ -4104,8 +4101,13 @@ void dt_exif_init()
   Exiv2::XmpParser::initialize();
   // this has to stay with the old url (namespace already propagated outside dt)
   Exiv2::XmpProperties::registerNs("http://darktable.sf.net/", "darktable");
-  Exiv2::XmpProperties::registerNs("http://ns.adobe.com/lightroom/1.0/", "lr");
-  Exiv2::XmpProperties::registerNs("http://cipa.jp/exif/1.0/", "exifEX");
+  // check is Exiv2 version already knows these prefixes
+  const Exiv2::XmpPropertyInfo *pl = Exiv2::XmpProperties::propertyList("lr");
+  if(!pl)
+    Exiv2::XmpProperties::registerNs("http://ns.adobe.com/lightroom/1.0/", "lr");
+  pl = Exiv2::XmpProperties::propertyList("exifEX");
+  if(!pl)
+    Exiv2::XmpProperties::registerNs("http://cipa.jp/exif/1.0/", "exifEX");
 }
 
 void dt_exif_cleanup()


parafin:
Does this fix depend on specific exiv2 version?

clanmills:
No!  These prefix/URI namespaces are already registered in Exiv2 and have been for a long time.  Removing that code from darktable makes no difference to which version of Exiv2 is in use.

I've documented what I intend to investigate in Exiv2 concerning this issue:  https://github.com/Exiv2/exiv2/issues/1206#issuecomment-688488556

phweyland:
I've just upgraded from Exiv2 0.25. The issue was exactly the same. 
Currently dt requires Exiv2 0.24. 
I had a look at the [change log file](https://github.com/Exiv2/exiv2/commit/88ccfae682e602b4eeabeb34dd7a4eec2546b54d) and haven't found any reference to the introduction of lightroom and exifex, just a reference to a Xmp.lr.hierarchicalSubject bug for the 0.25 release, which sounds like lightroom was already known in 0.24.

phweyland:
As it is not clear when this did happen, just avoid to register the prefixes if they are already known by exiv2.

clanmills:
I believe that's correct.  I think Andreas added "lr" in 2011 and (for reasons I don't know), the declaration was modified in 2015.  'exifEX' was totally new in 2015.

This morning, I built and tested Exiv2::XmpProperties::propertyList("lr"); with exiv2 v0.25.  The current behaviour and v0.25 are the same. I think it has always been like this and simply not noticed!  

Concerning your call to `Exiv2::XmpProperties::registerNs("http://darktable.sf.net/", "darktable");`.  This call adds the namespace, however it has no properties.  If you want to add "properties" such as Xmp.darktable.this and Xmp.darktable.that, we will need to prepare a patch with xmpDarktableInfo in exiv2.  I've discussed this in my report here:  https://github.com/Exiv2/exiv2/issues/1206#issuecomment-688488556

src/common/exif.cc
```diff
@@ -4104,8 +4101,13 @@ void dt_exif_init()
   Exiv2::XmpParser::initialize();
   // this has to stay with the old url (namespace already propagated outside dt)
   Exiv2::XmpProperties::registerNs("http://darktable.sf.net/", "darktable");
-  Exiv2::XmpProperties::registerNs("http://ns.adobe.com/lightroom/1.0/", "lr");
-  Exiv2::XmpProperties::registerNs("http://cipa.jp/exif/1.0/", "exifEX");
+  // check is Exiv2 version already knows these prefixes
```
TurboGit, line 4104: is -> if

src/common/exif.cc
```diff
@@ -161,9 +161,6 @@ void dt_exif_set_exiv2_taglist()
 {
   if(exiv2_taglist) return;
 
-  Exiv2::XmpParser::initialize();
```
TurboGit, line 164: Are those removals really wanted? I'm don't see the relation with the fix below... Just in case I prefer asking.

phweyland: Yes, the exiv2 initialization is already done in dt_exif_init() which is called before in any case.

phweyland:
> Concerning your call to `Exiv2::XmpProperties::registerNs("http://darktable.sf.net/", "darktable");`. This call adds the namespace, however it has no properties. If you want to add "properties" such as Xmp.darktable.this and Xmp.darktable.that, we will need to prepare a patch with xmpDarktableInfo in exiv2

If we wanted to list darktable tags to make them tweak-able by the user for image exportation that would be the way to go.
So far I have not seen any request in that direction but it's good to know.

TurboGit:
Thanks !

PR #6237, opened 2020-09-08
po/fix style edit

TurboGit:


RawConvert:
Style edit panel checked in Master 3.3.0+584~g5424fa2e3-dirty  and looks ok.
White Balance no longer duplicated.
Black/White Point module no longer appears - I assume this is deliberate.

Before:
style11

After:
style11-test

PR #6238, opened 2020-09-08
colorin: properly reset all default parameters.

### LINKED ISSUE #5291: Non-RAW files don't get their embedded ICC profile initialized when history is deleted/compressed
1. Open a JPG, TIFF, PNG, etc. file
2. Either: 
   * In darkroom, click on the 0th step in history, then compress it,
   * In lighttable, click on "discard history"
3. See the input profile set as a matrice instead of embedded ICC. Sometimes, the setting says "embedded ICC" but the image is too bright, which means the actual profile used in the pipe is linear (so, the matrice) and the actual embedded profile is not properly initialized.
4. Reset the input profile module, now the right profile (embedded ICC) is used.

## What happens ?

Input profile, as white balance, initializes its defaults at runtime, in `reload_defaults()` by reading the image storage (`module->dev->image_storage`) and extracting relevant tags (color profile or white balance coefficients). But input profile actually reads the image buffer, extracts the embedded profile, and **writes** the profile in the image storage too, meaning it is not initialized before it runs. And this is the core issue of the problem (since you may notice white balance doesn't suffer from the same issue).

When history is manipulated, it uses only the database info. No image storage is passed to history management so the defaults cannot be initialized with the relevant settings and fallback to safety defaults.

## Solution 

Option 1:
History handling functions should also care about image storage and give modules the opportunity to properly read them to init sane defaults.

Option 2:
Embedded profile extraction should be done prior to pipeline startup, so the image container is complete and consistent before launching the pipeline and history handling (and not write it in colorin `reload_defauts`). That could be when opening the image or initializing the cache container.

TurboGit:
The default profile type must be reset to default if no other
profile is found.

Fixes #5291.

### DIFF
--- a/src/iop/colorin.c
+++ b/src/iop/colorin.c
@@ -1946,6 +1946,8 @@ void reload_defaults(dt_iop_module_t *module)
     d->type = DT_COLORSPACE_SRGB;
   else if(!isnan(module->dev->image_storage.d65_color_matrix[0]))
     d->type = DT_COLORSPACE_EMBEDDED_MATRIX;
+  else
+    d->type = DT_COLORSPACE_ENHANCED_MATRIX;
 
   dt_image_cache_write_release(darktable.image_cache, img, DT_IMAGE_CACHE_RELAXED);
 


jenshannoschwalm:
Oh man, I was staring at that code and couldn't find the obvious this morning :-(( absolutely correct.

PR #6239, opened 2020-09-08
snapshots: remove superfluous function call

### LINKED ISSUE #6235: toggle-last-snapshot accel has significant latency
**Describe the bug**
When using snapshots, I often set them to overlay the entire central view, then click on the button(s) for a fast A/B comparison. Clicking works very fast, and the speed is especially helpful when the differences are subtle, or when making a large number of comparisons while looking at different areas of the image each time. When I finally created a keyboard shortcut for the recently added "toggle last snapshot" accel, I found that it doesn't work as quickly as the button clicks, having a lag of perhaps 150ms on my (not terribly slow) hardware. It doesn't make the feature *useless*, but it does make it less useful.

**To Reproduce**
Bind a keyboard shortcut to this accel, then compare the latency to using a click on the buttons.

**Expected behavior**
The "hoped for" behavior would be a fast toggle, as similar to the clicking behavior as possible.

 - Darktable Version: 3.2.1 and current master
 - OS: Linux

**Additional context**
My guess, looking at `snapshots.c`, is that this is due to `gtk_widget_activate` being called *before* `_lib_snapshots_toggled_callback` in `_lib_snapshots_toggle_last`; I don't know what the correct fix for this would be, but I think there are some smart people here with experience in balancing widget responsiveness vs. preview responsiveness who can probably figure it out. :-)

elstoc:
Resolves #6235

### DIFF
--- a/src/libs/snapshots.c
+++ b/src/libs/snapshots.c
@@ -534,11 +534,9 @@ static gboolean _lib_snapshots_toggle_last(GtkAccelGroup *accel_group, GObject *
 {
   dt_lib_module_t *self = (dt_lib_module_t *)data;
   dt_lib_snapshots_t *d = (dt_lib_snapshots_t *)self->data;
-  if(d->num_snapshots)
-  {
-    gtk_widget_activate(d->snapshot[0].button);
-    _lib_snapshots_toggled_callback((GtkToggleButton *)(d->snapshot[0].button), data);
-  }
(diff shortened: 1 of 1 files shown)


elstoc:
Latest change appears to actually resolve #6235

TurboGit:
Looks ok to me.

PR #6240, opened 2020-09-08
Clean init/gui/reload/update separation

dterrahe:
Another set of patches to clean up copied around code, mostly in the iop gui sections.

The main aim of this set of commits (there are some extraneous fixes and cleanups as well) was to implement a clean set of responsibilities for the module init(), gui_init(), reload_params() and gui_update() methods. The pattern that could often be seen is that similar/identical code is copied into several of these methods, "just to make sure it works", rather than just to the one that should be correctly used for it. Partly this is due to implementation quirks. Once this is merged, I'm aiming to resolve the remaining issue that reload_params sometimes gets called when the gui is not fully initialised yet. And partly it is probably due to progressive "fixes" being implemented due to code working in one situation (for example switching into darkroom), but not another (changing images) . Due to this confusion, testing can also be complicated. I'm currently updating useless.c to provide a more narrative introduction to how this framework is intended to be used.

I recommend reviewing (if not testing) these commits separately, because I've tried to keep each dedicated mostly to one type of change. They should all apply cleanly as well, which will help bisecting if necessary. I'd rather not submit them as separate PRs because that just means I can't make any assumptions about dependencies between them and it'd be an ongoing rebasing nightmare.

One separate "feature" is the restructuring of rawprepare_params_t so that it now shows up cleaner in the history changes popup that was merged recently. This required supporting ushort in slider_from_params.

The button and label refactoring enforce consistent ellipsizing, which helps narrow panels in lighttable. Some other small fixes here and there round off that work.

### DIFF
--- a/src/iop/useless.c
+++ b/src/iop/useless.c
@@ -265,8 +265,7 @@ void gui_update(dt_iop_module_t *self)
 void gui_init(dt_iop_module_t *self)
 {
   // init the slider (more sophisticated layouts are possible with gtk tables and boxes):
-  self->gui_data = malloc(sizeof(dt_iop_useless_gui_data_t));
-  dt_iop_useless_gui_data_t *g = (dt_iop_useless_gui_data_t *)self->gui_data;
+  dt_iop_useless_gui_data_t *g = IOP_GUI_ALLOC(useless);
 
   self->widget = gtk_box_new(GTK_ORIENTATION_VERTICAL, DT_BAUHAUS_SPACE);
 
@@ -284,8 +283,8 @@ void gui_init(dt_iop_module_t *self)
 void gui_cleanup(dt_iop_module_t *self)
 {
   // nothing else necessary, gtk will clean up the slider.
-  free(self->gui_data);
-  self->gui_data = NULL;
+
+  IOP_GUI_FREE;
 }
 
 /** additional, optional callbacks to capture darkroom center events. */
--- a/src/iop/useless.c
+++ b/src/iop/useless.c
@@ -21,46 +21,86 @@
 // our includes go first:
 #include "bauhaus/bauhaus.h"
 #include "develop/imageop.h"
+#include "develop/imageop_gui.h"
+#include "gui/color_picker_proxy.h"
 #include "gui/gtk.h"
 #include "iop/iop_api.h"
 
 #include <gtk/gtk.h>
 #include <stdlib.h>
 
-// this is the version of the modules parameters,
-// and includes version information about compile-time dt
-DT_MODULE_INTROSPECTION(2, dt_iop_useless_params_t)
+// This is an example implementation of an image operation module that does nothing useful.
+// It demonstrates how the different functions work together. To build your own module,
+// take all of the functions that are mandatory, stripping them of comments.
+// Then add only the optional functions that are required to implement the functionality
+// you need. Don't copy default implementations (hint: if you don't need to change or add
+// anything, you probably don't need the copy). Make sure you choose descriptive names
+// for your fields and variables. The ones given here are just examples; rename them.
+//
+// To have your module compile and appear in darkroom, add it to CMakeLists.txt, with
+//  add_iop(useless "useless.c")
+// and to iop_order.c, in the initialisation of legacy_order & v30_order with:
+//  { {XX.0f }, "useless", 0},
+
+// This is the version of the module's parameters,
+// and includes version information about compile-time dt.
+// The first released version should be 1.
+DT_MODULE_INTROSPECTION(3, dt_iop_useless_params_t)
 
 // TODO: some build system to support dt-less compilation and translation!
 
+// Enums used in params_t can have $DESCRIPTIONs that will be used to
+// automatically populate a combobox with dt_bauhaus_combobox_from_params.
+// They are also used in the history changes tooltip.
+// Combobox options will be presented in the same order as defined here.
+// These numbers must not be changed when a new version is introduced.
+typedef enum dt_iop_useless_type_t
+{
+  DT_USELESS_NONE = 0,     // $DESCRIPTION: "no"
+  DT_USELESS_FIRST = 1,    // $DESCRIPTION: "first option"
+  DT_USELESS_SECOND = 2,   // $DESCRIPTION: "second one"
+} dt_iop_useless_type_t;
+
 typedef struct dt_iop_useless_params_t
 {
-  // these are stored in db.
-  // make sure everything is in here does not
-  // depend on temporary memory (pointers etc)
-  // stored in self->params and self->default_params
-  // also, since this is stored in db, you should keep changes to this struct
-  // to a minimum. if you have to change this struct, it will break
-  // users data bases, and you should increment the version
-  // of DT_MODULE(VERSION) above!
-  int checker_scale;
-  float factor;
+  // The parameters defined here fully record the state of the module and are stored
+  // (as a serialized binary blob) into the db.
+  // Make sure everything is in here does not depend on temporary memory (pointers etc).
+  // This struct defines the layout of self->params and self->default_params.
+  // You should keep changes to this struct to a minimum.
+  // If you have to change this struct, it will break
+  // user data bases, and you have to increment the version
+  // of DT_MODULE_INTROSPECTION(VERSION) above and provide a legacy_params upgrade path!
+  //
+  // Tags in the comments get picked up by the introspection framework and are
+  // used in gui_init to set range and labels (for widgets and history)
+  // and value checks before commit_params.
+  // If no explicit init() is specified, the default implementation uses $DEFAULT tags
+  // to initialise self->default_params, which is then used in gui_init to set widget defaults.
+  //
+  // These field names are just examples; chose meaningful ones! For performance reasons, align
+  // to 4 byte bounderies (use gboolean, not bool).
+  int checker_scale; // $MIN: 0 $MAX: 10 $DEFAULT: 1 $DESCRIPTION: "size"
+  float factor;      // $MIN: -5.0 $MAX: 5.0 $DEFAULT: 0
+  gboolean check;    // $DESCRIPTION: "checkbox option"
+  dt_iop_useless_type_t method; // $DEFAULT: DT_USELESS_SECOND $DESCRIPTION: "parameter choices"
 } dt_iop_useless_params_t;
 
 typedef struct dt_iop_useless_gui_data_t
 {
-  // whatever you need to make your gui happy.
-  // stored in self->gui_data
-  GtkWidget *scale, *factor; // this is needed by gui_update
+  // Whatever you need to make your gui happy and provide access to widgets between gui_init, gui_update etc.
+  // Stored in self->gui_data while in darkroom.
+  // To permanently store per-user gui configuration settings, you could use dt_conf_set/_get.
+  GtkWidget *scale, *factor, *check, *method, *extra; // this is needed by gui_update
 } dt_iop_useless_gui_data_t;
 
 typedef struct dt_iop_useless_global_data_t
 {
-  // this is optionally stored in self->global_data
+  // This is optionally stored in self->global_data
   // and can be used to alloc globally needed stuff
   // which is needed in gui mode and during processing.
 
-  // we don't need it for this example (as for most dt plugins)
+  // We don't need it for this example (as for most dt plugins).
 } dt_iop_useless_global_data_t;
 
 // this returns a translatable name
@@ -87,18 +127,57 @@ int default_colorspace(dt_iop_module_t *self, dt_dev_pixelpipe_t *pipe, dt_dev_p
   return iop_cs_Lab;
 }
 
+// Whenever new fields are added to (or removed from) dt_iop_..._params_t or when their meaning
+// changes, a translation from the old to the new version must be added here.
+// A verbatim copy of the old struct definition should be copied into the routine with a _v?_t ending.
+// Since this will get very little future testing (because few developers still have very
+// old versions lying around) existing code should be changed as little as possible, if at all.
+//
+// Upgrading from an older version than the previous one should always go through all in between versions
+// (unless there was a bug) so that the end result will always be the same.
+//
+// Be careful with changes to structs that are included in _params_t
+//
+// Individually copy each existing field that is still in the new version. This is robust even if reordered.
+// If only new fields were added at the end, one call can be used:
+//   memcpy(n, o, sizeof *o);
+//
+// Hardcode the default values for new fields that were added, rather than referring to default_params;
+// in future, the field may not exist anymore or the default may change. The best default for a new version
+// to replicate a previous version might not be the optimal default for a fresh image.
+//
+// FIXME: the calling logic needs to be improved to call upgrades from consecutive version in sequence.
 int legacy_params(dt_iop_module_t *self, const void *const old_params, const int old_version,
                   void *new_params, const int new_version)
 {
+  typedef dt_iop_useless_params_t dt_iop_useless_params_v3_t; // always create copy of current so code below doesn't need to be touched
+
+  typedef struct dt_iop_useless_params_v2_t
+  {
+    int checker_scale;
+    float factor;
+  } dt_iop_useless_params_v2_t;
+
+  if(old_version == 2 && new_version == 3)
+  {
+    dt_iop_useless_params_v2_t *o = (dt_iop_useless_params_v2_t *)old_params;
+    dt_iop_useless_params_v3_t *n = (dt_iop_useless_params_v3_t *)new_params;
+
+    memcpy(n, o, sizeof *o);
+    n->check = FALSE;
+    n->method = DT_USELESS_SECOND;
+    return 0;
+  }
+
+  typedef struct dt_iop_useless_params_v1_t
+  {
+    int checker_scale;
+  } dt_iop_useless_params_v1_t;
+
   if(old_version == 1 && new_version == 2)
   {
-    typedef struct dt_iop_useless_params_v1_t
-    {
-      int checker_scale;
-    } dt_iop_useless_params_v1_t;
-
     dt_iop_useless_params_v1_t *o = (dt_iop_useless_params_v1_t *)old_params;
-    dt_iop_useless_params_t *n = (dt_iop_useless_params_t *)new_params;
+    dt_iop_useless_params_v2_t *n = (dt_iop_useless_params_v2_t *)new_params;
 
     n->checker_scale = o->checker_scale;
     n->factor = 0.0;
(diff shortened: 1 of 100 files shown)


johnny-bit:
The less code, the better. The less copypasta boilerplate the better. The more is handled in "generic" way the better :)

TurboGit:
@johnny-bit : Fully agree, now let's ensure all this is equivalent code :)

dterrahe:
> now let's ensure all this is equivalent code :)

... or better... ;-)

dterrahe:
@elstoc I've updated useless.c with a bit of narrative on how some of the new infrastructure to define the iop guis hangs together. this is purely developer targeted, so no need for too much polish or translations, but if at some point you could give it a read an let me know where I'm completely failing to clarify things, that would be much appreciated. assuming you are interested in it yourself...

many thanks in advance!

elstoc:
Ok. I might not be able to resist copy-editing as well.

src/iop/useless.c
```diff
@@ -21,18 +21,37 @@
...
 #include "gui/gtk.h"
 #include "iop/iop_api.h"
 
 #include <gtk/gtk.h>
 #include <stdlib.h>
 
+// This is an example implementation of an image operation module that does nothing usefull.
+// It demonstrates how the different functions work together. To build your own module,
+// take only the required functions, stripping them of comments, and add the ones to implement
+// the functionality you need. Don't copy default implementations (hint: if you don't need to change
```
elstoc, line 34:
```suggestion
// take only the required functions, stripping them of comments, and add new functions to implement
// any additional functionality. Don't copy default implementations (hint: if you don't need to change
```

src/iop/useless.c
```diff
@@ -21,18 +21,37 @@
...
+// anything, you probably don't need the copy)
+
 // this is the version of the modules parameters,
```
elstoc, line 37: (not your comment I know but) should be "module parameters"

dterrahe:
if you want to, absolutely, that would be great! I'm not much of a writer myself (get too impatient) but if you want to make it a joint effort we could do this iteratively or however. definitely change whatever you want (as long as it isn't incorrect), to improve understanding and usefulness. and obviously if you need anything explained further, just ask here.

one doubt I had is if it can be too large (so that people don't read it at all anymore). I guess one of the functions would be that when reviewing patches we all (@TurboGit especially) can point here when some functionality is used incorrectly.

src/iop/useless.c
```diff
@@ -21,18 +21,37 @@
...
 // and includes version information about compile-time dt
-DT_MODULE_INTROSPECTION(2, dt_iop_useless_params_t)
+DT_MODULE_INTROSPECTION(3, dt_iop_useless_params_t)
 
 // TODO: some build system to support dt-less compilation and translation!
 
+// enums used in params_t can have $DESCRIPTIONs that will be used to
+// automatically populate a combobox with dt_bauhaus_combobox_from_params.
+// They are also used in the history changes tooltip.
+// The order here will be used in the combobox.
```
elstoc, line 46:
```suggestion
// Combobox options will be presented in the same order as defined here 
```

src/iop/useless.c
```diff
@@ -21,18 +21,37 @@
...
+// The numbers must not be changed when a new version is introduced.
```
elstoc, line 47: **These** numbers

src/iop/useless.c
```diff
@@ -41,17 +60,24 @@ typedef struct dt_iop_useless_params_t
   // stored in self->params and self->default_params
   // also, since this is stored in db, you should keep changes to this struct
   // to a minimum. if you have to change this struct, it will break
-  // users data bases, and you should increment the version
-  // of DT_MODULE(VERSION) above!
-  int checker_scale;
-  float factor;
+  // users data bases, and you have to increment the version
+  // of DT_MODULE(VERSION) above and provide a legacy_params upgrade path!
+  // Tags in the comments get picked up by the introspection framework and are
```
elstoc, line 65: The above 9 lines need a bit more sentence structure. I'm not sure where a sentence ends and the next begins. Here's what I think it says (with a couple of corrections thrown in): 

"The variables in this struct are stored in the database. Make sure none of these variables depend on temporary memory (pointers etc.). This struct is stored in `self->params` and `self->default_params`. Please keep changes to this struct to a minimum. In order to provide backward-compatibility with old edits, each time the struct changes you must increment the version of DT_MODULE(VERSION) above and provide a method to upgrade old parameters in legacy_params."

src/iop/useless.c
```diff
@@ -41,17 +60,24 @@ typedef struct dt_iop_useless_params_t
...
+  // used to set range and default for widgets in gui_init, labels (for widgets and history)
+  // and value checks before commit_params.
+  // If no explicit init() is specified, the default implementation uses $DEFAULT tags
+  // to initialise self->default_params, which is then used in gui_init to set widget defaults.
+  int checker_scale; // $MIN: 0 $MAX: 10 $DEFAULT: 1 $DESCRIPTION: "size"
+  float factor;      // $MIN: -5.0 $MAX: 5.0 $DEFAULT: 0
+  gboolean check;    // $DESCRIPTION: "checkbox option"
+  dt_iop_useless_type_t method; // $DEFAULT: DT_USELESS_SECOND $DESCRIPTION: "parameter choices"
```
elstoc, line 73: You might want to explicitly state that the above are examples as it's not clear from the names or descriptions (like it is clear for the combobox typedef above)

src/iop/useless.c
```diff
@@ -41,17 +60,24 @@ typedef struct dt_iop_useless_params_t
...
 } dt_iop_useless_params_t;
 
 typedef struct dt_iop_useless_gui_data_t
 {
   // whatever you need to make your gui happy.
   // stored in self->gui_data
-  GtkWidget *scale, *factor; // this is needed by gui_update
+  GtkWidget *scale, *factor, check, method, extra; // this is needed by gui_update
```
elstoc, line 80: Similarly be clear that these are examples.

src/iop/useless.c
```diff
@@ -87,18 +113,57 @@ int default_colorspace(dt_iop_module_t *self, dt_dev_pixelpipe_t *pipe, dt_dev_p
   return iop_cs_Lab;
 }
 
+// whenever new fields are added to (or removed from) dt_iop_..._params_t or when their meaning
+// changes, a translation from the old to the new version is added here.
```
elstoc, line 117: *must be* added here

src/iop/useless.c
```diff
@@ -87,18 +113,57 @@ int default_colorspace(dt_iop_module_t *self, dt_dev_pixelpipe_t *pipe, dt_dev_p
...
+// A verbatim copy of the old struct definition should be copied into the routine with a _v?_t ending.
+// Since this will get very little future testing (because few developers still have very
+// old versions lying around) existing code should be changed as little as possible, if at all.
+//
+// upgrading from an older version than the previous one should always go through all in between versions
+// (unless there was a bug) so that the end result will always be the same.
```
elstoc, line 123: An upgrade path must be provided between any previous version of the module parameters and the current version. For example if the current version is 5 but the old edit is on version 2, the upgrade path must take the parameters cleanly from v2->v3->v4->v5

src/iop/useless.c
```diff
@@ -87,18 +113,57 @@ int default_colorspace(dt_iop_module_t *self, dt_dev_pixelpipe_t *pipe, dt_dev_p
...
+// Be careful with changes to structs that are included in _params_t
+//
+// individually copy each existing field that is still in the new version. This is robust even if reordered.
+// If only new fields were added at the end, one call can be used:
+//   memcpy(n, o, sizeof(o));
+//
+// hardcode the default values for new fields that were added, rather than referring to default_params;
+// in future, the field may not exist anymore or the default may change. the best default for a new version
+// to replicate a previous version might not be the optimal default for a fresh image.
+//
```
elstoc, line 134: Again clarify that the following is an example and comment on what it's doing.

src/iop/useless.c
```diff
@@ -184,32 +249,22 @@ void process(struct dt_iop_module_t *self, dt_dev_pixelpipe_iop_t *piece, const
...
+  // in the comments of the params_t struct definition.
+  // an explicit implementation of init is only required if not all fields are
+  // fully supported by dt_iop_default_init, for example arrays with non-identical values.
+  // in that case, dt_iop_default_init can be called first followed by additional initialisation.
+  // the values in params will not be used and default_params can be overwritten by
+  // reload_params on a per-picture basis.
+  dt_iop_default_init(module);
+
+  // our module is enabled by default:
+  module->default_enabled = 1;
```
elstoc, line 267: Might also be worth referring to `module->hide_enable_button` as well?

src/iop/useless.c
```diff
@@ -219,6 +274,9 @@ void init_global(dt_iop_module_so_t *module)
 
 void cleanup(dt_iop_module_t *module)
 {
+  // releases any memory allocated in init(module)
+  // implement explicitly if the module allocates additional memory besides (default_)params.
```
elstoc, line 278: implement **this function** explicitly...

src/iop/useless.c
```diff
@@ -232,60 +290,185 @@ void cleanup_global(dt_iop_module_so_t *module)
...
   dt_dev_add_history_item(darktable.develop, self, TRUE);
+  dt_control_queue_redraw_widget(self->widget);
 }
 
-/** gui callbacks, these are needed. */
+/** gui setup and update, these are needed. */
 void gui_update(dt_iop_module_t *self)
 {
-  // let gui slider match current parameters:
+  // this gets called when switching to darkroom, with each picture change or when
```
elstoc, line 355: There are a couple of places where you say 'picture'. Pedantic I know but 'image' is probably the preferred term.

src/iop/useless.c
```diff
@@ -232,60 +290,185 @@ void cleanup_global(dt_iop_module_so_t *module)
...
+  // a different history item is selected.
+  // here all the widgets need to be set to the current values in param.
+  // in contrast, if individual widgets are changed, the widget callback is called or,
+  // for introspection mapped widgets (_from_params) gui_changed.
```
elstoc, line 359: I'm not 100% sure what you're saying here

src/iop/useless.c
```diff
@@ -232,60 +290,185 @@ void cleanup_global(dt_iop_module_so_t *module)
...
+  // whenever the widget is changed. additionally, gui_changed is called, if it exists,
+  // so that any other required changes, to dependent fields or to gui widgets, can be made.
+  // whenever self->params changes (switching images or history) the widget values have to
+  // be updated in gui_update.
+  // any widgets that are _not_ directly linked to a field need to have a custom callback
+  // function set up to respond to the "value-changed" signal.
+  // do not set the value of widgets or configure them depending on field values; this should
+  // be done in gui_update (or gui_changed or individual widget callbacks)
+  // if any default values for (slider) widgets or options (in comboboxes) depend on the
+  // type of picture, then the widgets have to be updated in reload_params.
```
elstoc, line 422: This is too long for me and needs some structure. Some of the comments here might work better with a worked example.

elstoc (commented):
I probably got a bit lazy towards the end so I might come back with more stuff later and I'm still not 100% sure how it all _should_ hang together (I'm learning bits of it as I investigate bugs etc). Generally though it's explaining complex stuff so some sentence and paragraph structure is needed, if only to give the reader a bit of a breather. There are points where it's not immediately obvious where sentences begin and end. This (useless.c) is definitely very useful to have but I think it might be easier to provide a more structured explanation of some of the more complex stuff in a separate wiki article with this as a worked example to read alongside.

src/iop/useless.c
```diff
@@ -232,60 +290,185 @@ void cleanup_global(dt_iop_module_so_t *module)
...
 }
 
 void gui_cleanup(dt_iop_module_t *self)
 {
-  // nothing else necessary, gtk will clean up the slider.
-  free(self->gui_data);
-  self->gui_data = NULL;
+  // this only needs to be provided if gui_init allocates any memory or resources besides
+  // self->widget and gui_data_t. in that case IOP_GUI_FREE needs to be called to free the
+  // latter. the widget will be destroyed automatically.
```
elstoc, line 469: The term "the latter" in this sentence refers to "gui_data_t" but I'm not sure this was intentional.

src/iop/useless.c
dterrahe: I think that's not exactly what I meant to say. Maybe this:
"to build your own module, take all of the functions that are mandatory, stripping them of comments. then add only the the optional functions that are required to implement the functionality you need. don't copy default implementations (hint: if you don't need to change anything, you probably don't need the copy)

src/iop/useless.c
dterrahe: or maybe even "module's parameters"?

src/iop/useless.c
dterrahe: that applies to everything in the whole module though...
"This is an example implementation..."
so if this was going to be repeated here, maybe it could make clear that applies to everything in the rest of the module as well, rather then repeating it everywhere which could become tiresome? I'll take suggestions for a phrase!

src/iop/useless.c
dterrahe: see above. we are not aiming for the very lowest common denominator here; unless I'm wrong, anybody that has any chance of understanding any of this will get that these are examples? from a writer's perspective you are absolutely right of course (and if you give me concrete sentence to insert I'll do it!) but my brain just can't think of anything to add without feeling silly ;-)

src/iop/useless.c
dterrahe: OK, so this bit is somewhat tentative because the full functionality is not yet there in dt_dev_read_history_ext. So _at the moment_ legacy_params needs to provide 1->5, 2->5, 3->5 and 4->5. This leads to the need to amend earlier version upgrades when a new one is added. Some modules already internally implement 1->5 by calling 1->4 and 4->5; this logic should be implemented centrally (which, I think aurelien pointed out already happens for presets or something). In future, once dt_dev_read_history_ext is fixed, only 1->2, 2->3, 3->4 etc need to be provided. At that point it is "guaranteed" to not make a difference if a file is upgraded in one step from 1->5 or, say, from 1->3 and 3->5.
My version doesn't clearly explain this at all. Your example is clear, but the first sentence...
But anyway, this has to be more clearly marked as not operational yet. On my todo list...

src/iop/useless.c
dterrahe: well, it's not doing anything useful, as promised ;-)

upgrading the example to do something mildly interesting/sensible would be a bigger job.

I probably made it less sensible by adding some random fields just to demonstrate the two version upgrade process...

src/iop/useless.c
elstoc: That sounds fine. Add some capital letters at the start of sentences though please!

src/iop/useless.c
elstoc: That works too

src/iop/useless.c
dterrahe: could do, but this isn't intended to be exhaustive. just one non-default setting to justify not falling back completely to the default init (be not providing our own init at all).

src/iop/useless.c
elstoc: Thing is some parts of the example module are a general template (should be used in any module) and some are part of the module-specific functionality (the example). It's not always entirely clear which is which, though mostly it is implied by context. Perhaps if you used the same naming convention as you do for the struct itself and call the variables useless_scale, useless_factor etc?

src/iop/useless.c
dterrahe: thanks! I was vaguely aware I wasn't being consistent. searched-and-replaced!

src/iop/useless.c
elstoc: As above, possibly well chosen variable names would make it clear.

src/iop/useless.c
dterrahe: yes, that's what is meant. if you can think of a way to make that less ambiguous, I'll take it!

src/iop/useless.c
elstoc: Ok I didn't realise that (legacy params is one of those bits I'm hazy on).

src/iop/useless.c
elstoc: Doesn't gui_cleanup also need to free the additional data/resource as well though (besides gui_data_t)? Isn't that the point of this function?

src/iop/useless.c
elstoc: ok

src/iop/useless.c
dterrahe: This is probably the most crucial section as I expect gui_init is where people would look for information most, so I'm branching to all the other functions from here. I've broken it into paragraphs and moved one into the example section.

please try to read again!

src/iop/useless.c
dterrahe: Ah, but IOP_GUI_FREE can't do that...

Basically what I'm saying is, the default function (if an explicit one isn't provided here) takes care of gui_data_t (and widget gets destroyed anyway). If you _override_ the default, you have to do whatever you have to do, and _also_ call IOP_GUI_FREE to clean up gui_data_t.

src/iop/useless.c
elstoc: It's better like that. Still need to consistently start sentences with capital letters though. It really does help with readability (I try to make sure sentences don't start with the word 'darktable' or I can't sleep at night).

I still feel like I need to understand it all better in the first place to comment too much though.

src/iop/useless.c
dterrahe: I'm not sure I'm allowed to! @TurboGit what is the rule around capitals in _d_arktable source precisely? or is there no rule?

src/iop/useless.c
dterrahe: I think I forgot to commit a comment here. I think I suggested:
"This struct defines the layout of self->params and self->default_params."

src/iop/useless.c
dterrahe: I don't _disagree_ with you, and I see how it takes a _little bit_ of effort to deduce from context, but if someone can't manage that, they have no hope of getting any of the rest right and definitely should not start by implementing a module from scratch. It may be a kindness to scare them off at an early stage!

(I'm only half joking here)

Adding useless_ to everything might suggest that the naming convention is to add the module name in front of every field, which I personally would not want to encourage.

src/iop/useless.c
elstoc: That sounds good. Say that.

src/iop/useless.c
elstoc: That makes more sense

src/iop/useless.c
elstoc: > useless_ to everything might suggest that the naming convention

I did think of that. I'm a long way from being able to implement a module from scratch so maybe I should be scared off. How about if you prefixed them all with some_ instead of having them be semi-meaningful. So `some_int` `some_float` `some_gboolean`.

src/iop/useless.c
elstoc: Capital letters in comment blocks should be fine. There are some already - just not consistently. I'm not suggesting adding any to the actual code but you're basically giving big chunks of prose here so proper sentences (_d_arktable notwitstanding) shouldn't be an issue.

All lower case is fine if it's a one-liner.

dterrahe:
Now actually compiles and loads...

TurboGit (approved):
Looks good!

TurboGit (commented):
Some comments on useless.c

src/iop/useless.c
```diff
@@ -21,46 +21,85 @@
...
+//
+// To have your module compile and appear in darkroom, add it to CMakeLists.txt, with
+//  add_iop(useless "useless.c")
+// and to iop_order.c, in the initialisation of v30_order with:
+//  { {79.0f }, "useless", 0},
+
+// this is the version of the module's parameters,
 // and includes version information about compile-time dt
-DT_MODULE_INTROSPECTION(2, dt_iop_useless_params_t)
+DT_MODULE_INTROSPECTION(3, dt_iop_useless_params_t)
```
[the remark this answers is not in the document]
dterrahe: If you think that's necessary? It seems pretty obvious to me, also from the legacy_params list of previous versions. I feel a little silly writing it down :-)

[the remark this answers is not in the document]
TurboGit: Well I won't make this blocking of course but I always find better to word things even if obvious for us :)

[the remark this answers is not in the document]
dterrahe: OK, I added it. Now you'll have to judge by something else if someone submitting a completely new module has any common sense... ;-)

TurboGit:
I have tested quite a big and all this seems working fine to me. I'll do a second pass tomorrow and if goes well I'll merge as we don't want to delay this too much as it touches quite a lot of code !

Thanks for cleaning lot of part of dt, will really help maintenance.

dterrahe:
Thanks @TurboGit 
I am well aware that with all this clean-up work half the time is spent in testing so I very much appreciate your cooperation.
My thinking/hope is that the benefits come in four parts;
- simpler code is easier to read and check; less time spend double-checking code that works
- when amending, better understanding and fewer side effects reduce regressions
- cleaner structure is easier to follow, hence introducing fewer bugs in corner cases
- working from a cleaner base means that suboptimal code doesn't get copied everywhere
So if it all works out, hopefully you will also win back some of the time you spent in future...

TurboGit:
No issue found again today. Let's merge. Thanks!

PR #6253, opened 2020-09-09
over-exposure warning : add meaningfull modes

aurelienpierre:
Currently, the over-exposure preview, in darkroom, shows any of the 3 RGB channels that is outside of the valid range defined by the upper and lower thresholds. This information is not very helpful since that kind of clipping can come from a mix of luminance clipping (so, **true** overexposure) and gamut clipping (that is, oversaturation or lack of proper gamut-mapping). Yet, users take it very seriously and get mistaken. I have seen people on YouTube reducing exposure until oversaturated colors got back in gamut, just because of that faulty and misleading preview.

### Changes :
* rename overexposure "clipping" because that what this feature is about.
* change the black threshold for an EV (log) control, that should be more meaningful:
Screenshot_20200909_223436
The tooltip gives the typical black clipping/emission values:
Screenshot_20200909_223536
* add a "luminance only" mode, that shows the clipping over luminance, which means tone mapping or exposure settings have been ill-set
* add a "saturation only" mode, that shows the over-saturated colors, imposible to represent in the target color space, which means gamut mapping or saturation settings have been ill-set,
* rename current mode "any RGB channel", because that's what it does,
* add a "full gamut" mode, that shows the combination of the 3 previous options. This should be consistent with the gamut alert tool (same result), but doesn't rely on LittleCMS2 lib and has an OpenCL path, meaning it's 2 to 6 times faster,
* optimize the code for SIMD vectorization,
* change defaults settings to sRGB black clipping for black (-12.68 EV) and almost peak-display luminance (99.99%) so users stop overreacting to valid values (previous defaults were 0.1% and 98%).

Full gamut preview against Adobe RGB (this PR – 34 ms with GPU/OpenCL, 64 ms on CPU/OpenMP):
Screenshot_20200909_224142
Gamut alert against Adobe RGB (using LittleCMS2 from colorout – 174 ms):
Screenshot_20200909_224154
Same settings but with desaturation in color balance (mostly just the luminance clipping):
Screenshot_20200909_225800

### How to test:

* in luminance mode, ensure you get the same alert for a color image and its fully desaturated version (using saturation in color balance),
* in saturation mode, ensure a full desaturation (in color balance) removes any alert you had for the RGB file,
* in full gamut mode, ensure the over-exposed (red) area perfectly overlaps the gamut alert, when softproofing profile is the same as the histogram profile.

### DIFF
--- a/src/iop/overexposed.c
+++ b/src/iop/overexposed.c
@@ -42,7 +42,7 @@ typedef enum dt_iop_overexposed_colorscheme_t
   DT_IOP_OVEREXPOSED_PURPLEGREEN = 2
 } dt_iop_overexposed_colorscheme_t;
 
-static const float dt_iop_overexposed_colors[][2][4]
+static const float DT_ALIGNED_ARRAY dt_iop_overexposed_colors[][2][4]
     = { {
           { 0.0f, 0.0f, 0.0f, 1.0f }, // black
           { 1.0f, 1.0f, 1.0f, 1.0f }  // white
@@ -148,7 +148,7 @@ static void _transform_image_colorspace(dt_iop_module_t *self, const float *cons
       = dt_ioppr_add_profile_info_to_list(self->dev, darktable.color_profiles->display_type,
                                           darktable.color_profiles->display_filename, INTENT_PERCEPTUAL);
   const dt_iop_order_iccprofile_info_t *const profile_info_to
-      = dt_ioppr_add_profile_info_to_list(self->dev, histogram_type, histogram_filename, INTENT_PERCEPTUAL);
+      = dt_ioppr_add_profile_info_to_list(self->dev, histogram_type, histogram_filename, INTENT_RELATIVE_COLORIMETRIC);
 
   if(profile_info_from && profile_info_to)
     dt_ioppr_transform_image_colorspace_rgb(img_in, img_out, roi_in->width, roi_in->height, profile_info_from,
@@ -162,7 +162,8 @@ void process(struct dt_iop_module_t *self, dt_dev_pixelpipe_iop_t *piece, const 
 {
   dt_develop_t *dev = self->dev;
 
-  const int ch = piece->colors;
+  const int ch = 4;
+  assert(piece->colors == ch);
 
   float *const img_tmp = dt_alloc_align(64, ch * roi_out->width * roi_out->height * sizeof(float));
   if(img_tmp == NULL)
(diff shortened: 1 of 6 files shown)


elstoc:
I've tried this, resetting both sliders to default. I don't get any black clipping indicators until I change the lower threshold to approx -9.6 EV even when setting the filmic black slider such that most of my image is black.

When using the 'full gamut' mode would it be a good idea to use more colors to show what is being indicated (gamut/overexposure)?

Now that you've changed it to a 'clipping' indicator instead of 'overexposed' should we change the default keyboard shortcut (O)?

aurelienpierre:
> I've tried this, resetting both sliders to default. I don't get any black clipping indicators until I change the lower threshold to approx -9.6 EV even when setting the filmic black slider such that most of my image is black.

That only means filmic does its job. The new default is to remap display black to -12.68 EV out of filmic. Testing with filmic on is probably not a good idea. But that also tells me that users are so used to false positives in that alert, that they find it strange if a piece of soft in the pipe actually prevents clipping efficiently. 

> When using the 'full gamut' mode would it be a good idea to use more colors to show what is being indicated (gamut/clipping)?

Gamut clipping is the whole package: over/under-exposure + oversaturation. The thing is these 2 will often overlap, so trying to differentiate them is doomed to fail in general. That's why you get the saturation only/luminance only, to try to make more sense out of that mess. Full gamut alert is pretty much useless, unless you want a quick preview to ensure everything is fine in a glance, but it's too much information at once to be usable and lead to a solution.

> Now that you've changed it to a 'clipping' indicator instead of 'overexposed' should we change the default keyboard shortcut (O)?

Not sure.

elstoc:
> Testing with filmic on is probably not a good idea. But that also tells me that users are so used to false positives in that alert, that they find it strange if a piece of soft in the pipe actually prevents clipping efficiently.

Maybe I'm misunderstanding what clipping means then because as far as I can tell I can make half my image black (0,0,0) and still have no clipping indicator shown (because it's not _below_ zero?). Basically you're saying that with filmic this will never show any area as clipped in which case I'm not sure there's much call for it at all is there? Surely as a minimum it should show which areas of the image are pure white and which areas of the image are pure black?

aurelienpierre:
Clipping is what will happen, when saved to JPEG or sent to display, with RGB values outside the valid range. 

The valid range starts where the `float -> uint` rounding errors are greater than the signal, e.g. for values lower than `log2( EOTF(0.5 / (2^bit_depth - 1)) )`, and ends at 100%, which is peak display luminance, whatever the medium is.

Having a flat area where details are lost doesn't necessarily means that area is clipped, aka out of range. It only means that local contrast was lost there, perhaps because of a thresholding, perhaps because of an excessive tonal compression. But whatever happened there, the black value is still valid and can be represented on your output medium. Only your gradient got disturbed.

TurboGit:
>     * add a "full gamut" mode, that shows the combination of the 3 previous options. This should be consistent with the gamut alert tool (same result), but doesn't rely on LittleCMS2 lib and has an OpenCL path, meaning it's 2 to 6 times faster,

Then should we keep the gammut check based on LittleCMS2 ?

aurelienpierre:
> Then should we keep the gammut check based on LittleCMS2 ?

Ultimately, I hope to get rid of that lib because it is painfully slow. But for now, it is a well-tested and widely used lib that can be used as a reference to test internal color affairs. I wouldn't rush into removing it.

TurboGit:
You also need to update the defaults for lower/upper in data/darktableconfig.xml.in.

It would also be good to set lower with the new default for previous installation has otherwise by default we have a full blue display. Not sure how to do so properly though.

aurelienpierre:
I will just enforce the lower value to -4 EV maximum (6.25 %).

TurboGit:
I would have expected the upper and lower default in darktableconfig.xml.in to match the default set on the sliders, no?

aurelienpierre:
Ah, stupid, I forget the upper…

TurboGit:
> Ultimately, I hope to get rid of that lib because it is painfully slow. But for now, it is a well-tested and widely used lib that can be used as a reference to test internal color affairs. I wouldn't rush into removing it.

And forgot actually hat the LittleCMS based gammut clipping works also with the intent which is not the case of this PR.

src/iop/overexposed.c
```diff
@@ -148,7 +148,7 @@ static void _transform_image_colorspace(dt_iop_module_t *self, const float *cons
       = dt_ioppr_add_profile_info_to_list(self->dev, darktable.color_profiles->display_type,
                                           darktable.color_profiles->display_filename, INTENT_PERCEPTUAL);
   const dt_iop_order_iccprofile_info_t *const profile_info_to
-      = dt_ioppr_add_profile_info_to_list(self->dev, histogram_type, histogram_filename, INTENT_PERCEPTUAL);
+      = dt_ioppr_add_profile_info_to_list(self->dev, histogram_type, histogram_filename, INTENT_RELATIVE_COLORIMETRIC);
```
TurboGit, line 151: Why changing the intent here?

src/views/darkroom.c
```diff
@@ -2203,33 +2213,56 @@ void gui_init(dt_view_t *self)
     gtk_container_add(GTK_CONTAINER(dev->overexposed.floating_window), vbox);
 
     /** let's fill the encapsulating widgets */
+    /* preview mode */
+    GtkWidget *mode = dt_bauhaus_combobox_new(NULL);
+    dt_bauhaus_widget_set_label(mode, NULL, _("clipping preview mode"));
+    dt_bauhaus_combobox_add(mode, _("full gamut"));
+    dt_bauhaus_combobox_add(mode, _("any RGB channel"));
+    dt_bauhaus_combobox_add(mode, _("luminance only"));
+    dt_bauhaus_combobox_add(mode, _("saturation only"));
```
TurboGit, line 2222: As "any RGB channel" is what was done before, maybe we should we keep this as the default value instead of "full gamut" which is mostly equivalent to the other button (gammut check) ?

TurboGit (commented):
Some comments...

src/iop/overexposed.c
aurelienpierre: Perceptual is black magic and poorly suited for gamut checking. For media profiles, it uses a vendor-dependent LUT to do the gamut mapping, along with some S curve to perform the black point adaptation. But for standard RGB spaces (without that LUT defined), what they do is undefined.

Relative colorimetric is more predictable, since it desaturates out-of-gamut colors at same luminance, until they reach the target gamut. Since we are here doing diagnostic over clipping, relative colorimetric is just more reliable.

src/views/darkroom.c
aurelienpierre: Any RGB channel is really misleading as a default and is the reason why people do bad things now. It only shows oversaturation for whites, that is no actual over/under-exposure, but doesn't show oversaturation for valid luminances either. Full gamut has been chosen as a default because it is the combination of the 3 other modes, so it's a more complete and robust estimation of what's wrong in the picture.

aurelienpierre:
> And forgot actually hat the LittleCMS based gammut clipping works also with the intent which is not the case of this PR.

To be honest, intents are shit. Perceptual is 100% vendor-based, relative colorimetric desaturates at same luminances but doesn't use a hue-linear desaturation, and the others are useless to photography. Once I tackle the gamut mapping in hue-linear ways, it will do nothing more than we already to internally.

TurboGit:
Ok, thanks for the explanations.

TurboGit:
No more comment, thanks!

PR #6254, opened 2020-09-09
darkroom: allow better translations

rabauke:
Punctuation marks need to be wrapped into gettext macros too to allow accurate translations.

### DIFF
--- a/src/views/darkroom.c
+++ b/src/views/darkroom.c
@@ -421,12 +421,12 @@ void expose(
     if(dev->image_invalid_cnt)
     {
       fontsize = DT_PIXEL_APPLY_DPI(20);
-      load_txt = dt_util_dstrcat(NULL, "%s `%s', %s\n\n%s\n%s",
-          _("darktable could not load"),
-          dev->image_storage.filename,
-          _("switching to lighttable now."),
-          _("please check the image (use exiv2 or exiftool) for corrupted data. if the image seems to"),
-          _("be intact, please consider opening an issue at https://github.com/darktable-org/darktable.") );
+      load_txt = dt_util_dstrcat(
+          NULL,
+          _("darktable could not load `%s', switching to lighttable now.\n\n"
+            "please check the image (use exiv2 or exiftool) for corrupted data. if the image seems to\n"
+            "be intact, please consider opening an issue at https://github.com/darktable-org/darktable."),
+          dev->image_storage.filename);
       if(dev->image_invalid_cnt > 400)
       {
         dev->image_invalid_cnt = 0;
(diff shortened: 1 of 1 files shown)


src/views/darkroom.c
```diff
@@ -421,12 +421,11 @@ void expose(
...
     {
       fontsize = DT_PIXEL_APPLY_DPI(20);
-      load_txt = dt_util_dstrcat(NULL, "%s `%s', %s\n\n%s\n%s",
-          _("darktable could not load"),
-          dev->image_storage.filename,
+      load_txt = dt_util_dstrcat(
+          NULL, NC_("darkroom", "%s `%s', %s\n\n%s"), _("darktable could not load"), dev->image_storage.filename,
           _("switching to lighttable now."),
-          _("please check the image (use exiv2 or exiftool) for corrupted data. if the image seems to"),
-          _("be intact, please consider opening an issue at https://github.com/darktable-org/darktable.") );
```
rabauke, line 429: There is no good reason to split this string into two translatable string. It appears, as it is just because of code formatting.

johnny-bit: I think the whole string should be something like:
```
"darktable could not load `%s', switching to lighttable now.\n\n"
"please check the image (use exiv2 or exiftool) for corrupted data. if the image seems to\n"
"be intact, please consider opening an issue at https://github.com/darktable-org/darktable."
```
and only format parameter should be `dev->image_storage.filename`.

that way the whole string is translatable and allows for better grammar in languages that would need different construction than "could not load" + "file"

src/views/darkroom.c
```diff
@@ -436,7 +435,7 @@ void expose(
     else
     {
       fontsize = DT_PIXEL_APPLY_DPI(14);
-      load_txt = dt_util_dstrcat(NULL, "%s `%s' ...", _("loading"), dev->image_storage.filename);
+      load_txt = dt_util_dstrcat(NULL, NC_("darkroom", "%s `%s' ..."), _("loading"), dev->image_storage.filename);
```
johnny-bit, line 438: similarly here the string could be just "loading `%s' ..."

johnny-bit (commented):
Looks great, I think it's mostly problem with how strings were thought about. I even wonder if changing `dt_util_dstrcat` to `g_strdup_printf` would be another good idea..

src/views/darkroom.c
TurboGit: Agreed !

src/views/darkroom.c
TurboGit: Agreed too !

rabauke:
I revised my pull request taking into account the reviews.

TurboGit (approved):
Thanks !

Issue #6258, opened 2020-09-10
Errors when building on MacOS Mojave


matt-maguire:
When following the build instructions for MacOSX:
https://github.com/darktable-org/darktable/blob/master/packaging/macosx/BUILD.txt

an empty "build" directory is created inside the clone of the darktable repository, and from within that empty build directory, the following command is run:
`cmake .. -DCMAKE_OSX_DEPLOYMENT_TARGET=10.7 -DCMAKE_CXX_FLAGS=-stdlib=libc++ -DOpenMP_C_INCLUDE_DIR=/opt/local/include/libomp -DOpenMP_CXX_INCLUDE_DIR=/opt/local/include/libomp -DCMAKE_LIBRARY_PATH=/opt/local/lib/libomp -DBINARY_PACKAGE_BUILD=ON -DRAWSPEED_ENABLE_LTO=ON -DBUILD_CURVE_TOOLS=ON -DBUILD_NOISE_TOOLS=ON`

If we then start building wit hthe "make" command, the build aborts halfway through with the error message:
```
Scanning dependencies of target filter
[ 50%] Building C object src/libs/CMakeFiles/filter.dir/tools/filter.c.o
/Users/mmaguire/src/darktable-dev/src/libs/tools/filter.c:205:68: error: implicit conversion from enumeration type 'const dt_collection_filter_t'
      (aka 'const enum dt_collection_filter_t') to different enumeration type 'dt_collection_sort_t' (aka 'enum dt_collection_sort_t') [-Werror,-Wenum-conversion]
  gtk_combo_box_set_active(GTK_COMBO_BOX(widget),_filter_get_items(sort));
                                                 ~~~~~~~~~~~~~~~~~ ^~~~
/Users/mmaguire/src/darktable-dev/src/libs/tools/filter.c:418:72: error: implicit conversion from enumeration type 'const dt_collection_filter_t'
      (aka 'const enum dt_collection_filter_t') to different enumeration type 'dt_collection_sort_t' (aka 'enum dt_collection_sort_t') [-Werror,-Wenum-conversion]
    gtk_combo_box_set_active(GTK_COMBO_BOX(d->sort), _filter_get_items(sort));
                                                     ~~~~~~~~~~~~~~~~~ ^~~~
/Users/mmaguire/src/darktable-dev/src/libs/tools/filter.c:437:72: error: implicit conversion from enumeration type 'const dt_collection_filter_t'
[...]

PR #6261, opened 2020-09-10
fix collection sort type

phweyland:
fix for #6258


TurboGit (approved):
Obviously correct, thanks!

Issue #5892, opened 2020-08-08
No timestamps are displayed on Windows


esq4:
No timestamps are displayed on Windows (but they are present in the db)

![timestaps in windows 7](https://user-images.githubusercontent.com/15331117/89547938-96e76b80-d830-11ea-8128-82c450c95109.png)

Darktable Version: master
OS: Windows 7 Pro

parafin:
That fix broke darktable on macOS for me, it crashes with the following backtrace:
```
* thread #1, queue = 'com.apple.main-thread', stop reason = EXC_BAD_ACCESS (code=1, address=0x0)
  * frame #0: 0x00000001005e84bf libglib-2.0.0.dylib`g_utf8_validate + 15
    frame #1: 0x000000010f161156 libmetadata_view.so`_mouse_over_image_callback [inlined] _metadata_update_value(label=0x0000000106d3cd90, value=0x0000000000000000) at metadata_view.c:208:24 [opt]
    frame #2: 0x000000010f161145 libmetadata_view.so`_mouse_over_image_callback at metadata_view.c:297 [opt]
    frame #3: 0x000000010f160ba4 libmetadata_view.so`_mouse_over_image_callback(instance=<unavailable>, user_data=<unavailable>) at metadata_view.c:774 [opt]
    frame #4: 0x00000001012244af libgobject-2.0.0.dylib`g_closure_invoke + 192
    frame #5: 0x000000010123792b libgobject-2.0.0.dylib`signal_emit_unlocked_R + 2315
    frame #6: 0x0000000101236f17 libgobject-2.0.0.dylib`g_signal_emitv + 485
    frame #7: 0x00000001001afbf0 libdarktable.dylib`_signal_raise(user_data=0x000000010c1eaa60) at signal.c:222:3 [opt]
    frame #8: 0x00000001005bfcc6 libglib-2.0.0.dylib`g_main_context_invoke_full + 68
    frame #9: 0x00000001001afff5 libdarktable.dylib`dt_control_signal_raise(ctlsig=0x00000001059689f0, signal=DT_SIGNAL_MOUSE_OVER_IMAGE_CHANGE) at signal.c:333:5 [opt]
    frame #10: 0x0000000100241191 libdarktable.dylib`_event_main_motion(widget=<unavailable>, event=<unavailable>, user_data=0x000000010f48c320) at thumbnail.c:646:55 [opt]
[...]

PR #6262, opened 2020-09-10
fix crashes darktable on macOS on #5901

esq4:
Could this be the solution for #5892?

### DIFF
--- a/src/libs/metadata_view.c
+++ b/src/libs/metadata_view.c
@@ -294,7 +294,11 @@ static void _metadata_view_update_values(dt_lib_module_t *self)
       char datetime[200];
       // just %c is too long and includes a time zone that we don't know from exif
       strftime(datetime, sizeof(datetime), "%a %x %X", localtime(&img->import_timestamp));
+#ifdef __APPLE__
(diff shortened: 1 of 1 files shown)


parafin:
There are 2 questions here:

1. Does #5892 original issue exist on macOS - I will check.
2. Can crash from #5901 happen on platforms other than macOS - I don't know, but checking for errors is a good idea even if errors seem to never happen.

So let's have someone else's opinion here too.

johnny-bit:
I looked at the issue here at length today and I wanted to call dibs, alas I'm 24 minutes late :/

If you don't mind I'd like to propose an alternate solution that should be universal, not depend on ifdefs (root of all evil) and not leaking memory like crazy ;)

Plz give me ~1h

parafin:
I've checked and timestamps on macOS look strange - I see some hieroglyphs in there, but it's not empty like on Windows (originally). So better fix is indeed welcome here.

johnny-bit:
@parafin - please check #6263 With it we've got a fighting chance to understand why there are hieroglyphs on macos and it should cover windows usecase (windows strftime returns non-utf8 compatible string that can be converted)

TurboGit:
Closing as better fix put in place.

PR #6263, opened 2020-09-10
[metadata_view]: correctly handlne invalid utf in strftime

### LINKED ISSUE #5892: No timestamps are displayed on Windows

johnny-bit:
This is imo the correct way to fix #5892 

This is an alternative to #6262 

Selling points: no memory leaks, error handling :) if string is utf-8 compatible there's no burned cpu cycles :)

### DIFF
--- a/src/libs/metadata_view.c
+++ b/src/libs/metadata_view.c
@@ -542,9 +542,7 @@ static void _metadata_view_update_values(dt_lib_module_t *self)
       tt_exif.tm_mon--;
       tt_exif.tm_isdst = -1;
       mktime(&tt_exif);
-      // just %c is too long and includes a time zone that we don't know from exif
-      strftime(datetime, sizeof(datetime), "%a %x %X", &tt_exif);
-      _metadata_update_value(d->metadata[md_exif_datetime], g_locale_to_utf8(datetime,-1,NULL,NULL,NULL));
+      _metadata_update_timestamp(d->metadata[md_exif_datetime], &tt_exif);
     }
     else
       _metadata_update_value(d->metadata[md_exif_datetime], img->exif_datetime_taken);
--- a/src/libs/metadata_view.c
+++ b/src/libs/metadata_view.c
@@ -222,6 +222,39 @@ static void _metadata_update_value_end(GtkLabel *label, const char *value)
   gtk_widget_set_tooltip_text(GTK_WIDGET(label), str);
 }
 
+static void _metadata_update_timestamp(GtkLabel *label, const time_t *value)
+{
+  char datetime[200];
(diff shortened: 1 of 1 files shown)


parafin:
You didn't get all of them, so it still crashes for me. Check line 547.

parafin:
As for why I get hieroglyphs - it seems to be one of those wonderful locale problems on macOS - if I set LC_ALL=C, then everything looks good (in English). So let's ignore that for now, maybe I'll get around investigating it sometime (or better yet someone else).

johnny-bit:
I'll fix it in a bit (line 547) - as for hieroglyphs it still should in case of utf8 conv problems complain or if there are no complaints, those hieroglyphs are valid utf-8 and strftime may have invalid locale indeed...

johnny-bit:
Fixed. Can you please check if it doesn't crash now? should cover all timestamps now :)

parafin:
Doesn't build now:
```
/Users/parafin/src/darktable/src/libs/metadata_view.c:545:65: error: incompatible pointer types passing 'struct tm *' to parameter of type 'const time_t *' (aka 'const long *') [-Werror,-Wincompatible-pointer-types]
      _metadata_update_timestamp(d->metadata[md_exif_datetime], &tt_exif);
                                                                ^~~~~~~~
/Users/parafin/src/darktable/src/libs/metadata_view.c:225:71: note: passing argument to parameter 'value' here
static void _metadata_update_timestamp(GtkLabel *label, const time_t *value)
                                                                      ^
/Users/parafin/src/darktable/src/libs/metadata_view.c:540:12: error: unused variable 'datetime' [-Werror,-Wunused-variable]
      char datetime[200];
           ^
2 errors generated.

```

johnny-bit:
Arght. Wrong branch built + rush :D Now should be good + properly tested :)

parafin:
Works now. I've also found a workaround for hieroglyphs. It seems setlocale on macOS is kind of (or totally) broken.

parafin:
BTW that workaround fixes the crash also, so this PR isn't required:)) But probably a good idea nonetheless.

johnny-bit:
At least it fixes memleaks (really, the mem would leak like crazy) + technically it only does utf8 conversion when absolutely needed so that's another plus and it even shows when conversion has problems :D

Anyway - Ready for review and merging :)

TurboGit:
It probably also fixes #4088.

TurboGit (approved):
All good to me! Thanks.

PR #6264, opened 2020-09-10
[darkroom] remove memleak in styles menu generation

johnny-bit:
calling g_strconcat in a loop leaks mem + causes multiple allocations

the code was changed to basically act as in-place g_strconcat since we know all elements before call.

Selling points: no memleak, single allocation per menuitem :D


TurboGit (approved):
Thanks !

PR #6267, opened 2020-09-11
fix string incorrectly marked for translation in rawprepare

dterrahe:
single line fix as requested by @TurboGit 

corrects mistake in #6240


TurboGit (approved):
Thanks !

PR #6269, opened 2020-09-11
implement clickable label in colorbalance

dterrahe:
inspired by @johnny-bit's clickable label in temperature.

mostly for fun, with a small improvement in very narrow slider label printing, but might be useful for some.

image

### DIFF
--- a/data/darktableconfig.xml.in
+++ b/data/darktableconfig.xml.in
@@ -2385,6 +2385,19 @@
     <shortdescription>arange buttons as bar below temp/tint sliders</shortdescription>
     <longdescription>if enabled, the buttons in white balance module will be aranged below temp/tint sliders, otherwise they'll be shown in a grid to the right of temp/tint sliders</longdescription>
   </dtconfig>
+  <dtconfig prefs="darkroom" section="interface">
+    <name>plugins/darkroom/colorbalance/layout</name>
+    <type>
+      <enum>
+        <option>box</option>
+        <option>notebook</option>
+        <option>grid</option>
+      </enum>
+    </type>
+    <default>box</default>
+    <shortdescription>colorbalance slider block layout</shortdescription>
+    <longdescription>choose how to organise the slider blocks for lift, gamma and gain:\nbox - all sliders are shown in one long list (with headers),\nnotebook - use tabs to switch between the blocks of sliders,\ngrid - the blocks of sliders are shown next to eachother (in narrow columns)</longdescription>
+  </dtconfig>
 
   @DARKTABLECONFIG_IOP_ENTRIES@
 
--- a/src/bauhaus/bauhaus.c
+++ b/src/bauhaus/bauhaus.c
@@ -1959,9 +1959,9 @@ static gboolean dt_bauhaus_draw(GtkWidget *widget, cairo_t *crf, gpointer user_d
       }
       // label on top of marker:
       set_color(cr, *text_color);
-      show_pango_text(w, context, cr, w->label, 0, 0,
-                      width - darktable.bauhaus->quad_width - INNER_PADDING*2.0 - value_width,
-                      FALSE, FALSE, PANGO_ELLIPSIZE_END, FALSE);
+      float label_width = width - darktable.bauhaus->quad_width - INNER_PADDING * 2.0 - value_width;
+      if(label_width > 0)
+        show_pango_text(w, context, cr, w->label, 0, 0, label_width, FALSE, FALSE, PANGO_ELLIPSIZE_END, FALSE);
     }
     break;
     default:
--- a/src/iop/colorbalance.c
+++ b/src/iop/colorbalance.c
@@ -107,6 +107,9 @@ typedef struct dt_iop_colorbalance_gui_data_t
 {
   dt_pthread_mutex_t lock;
   GtkWidget *master_box;
+  GtkWidget *main_label;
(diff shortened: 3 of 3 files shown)


johnny-bit:
"Color balance is tall module"
"Nah, with this PR it's T H I C C"

dterrahe:
Well, there's also a notebook mode, but some have moral objections to that, because they need to see all three blocks at the same time.
image

The grid mode is quite compressible, with the little fix to bauhaus, so that even in narrow width, the values are still readable. Anyone who uses scroll wheel to adjust sliders should be able to use it without problems in this mode.
image

Better than having to scroll the whole module up & down if you have a low resolution screen and want to see both histogram & filmstrip...

dt is all about flexibility!

dterrahe:
To be clear, the default is of course as before, one long list. This PR gives a total of 3 modes:
- box (vertical list as before; default)
- notebook
- grid

Unless you accidentally click on the first label, you won't ever be bothered by this change. If you do, it reconfigures on the fly. As does the setting. No restart needed.

TurboGit (changes requested):
The clickable label does not respect the current editing mode.

Select TSL only. Click on label and the R, G, B sliders will be displayed.

dterrahe:
Oops. Sorry for not noticing that. I'll fix.

I was wondering if the mode should also be saved so that it carries over between sessions? Shall i add this? There's a combo in the module so it does not need to be added to the settings dialog.

johnny-bit:
> I was wondering if the mode should also be saved so that it carries over between sessions?
IMO Yes, most modules work that way

> Shall i add this?

IMO Yes :)

dterrahe:
Both fixed and thanks for catching this!

One of the remaining gtk things I don't understand; why it sometimes requires a gtk_widget_show_all for widget changes (which obliterates the individual visible settings). If there's an alternative to show_all for these cases, I definitely want to know!

johnny-bit:
there's some no_show_all property that cancels visibility on call to show_all (like in white balance for emerald slider)... I think it's some kind of misguided api design in v1 making it's way to the end of times because nobody thought "somebody in the future might need a sane and robust way of showing/hiding selected widgets dynamically" :) Other than that I guess it's the question to gtk people :)

elstoc:
> why it sometimes requires a gtk_widget_show_all for widget changes

I'm not entirely sure what you're trying to achieve (so I might be wrong) but would `gtk_widget_queue_draw(g->main_box)` work?

dterrahe:
> > why it sometimes requires a gtk_widget_show_all for widget changes
> 
> I'm not entirely sure what you're trying to achieve (so I might be wrong) but would `gtk_widget_queue_draw(g->main_box)` work?

Yes, I thought it should too, but tried and it didn't. Calling it on `new_container` (the newly created and added widget) doesn't work either.
I tried both `gtk_widget_realize` and `gtk_widget_map` (even though the manual sort of warns against them) and got a crash. So far `show_all `is the only one that works reliably, but then you need to rehide any sub-widgets that were invisible before.

dterrahe:
> there's some no_show_all property that cancels visibility on call to show_all (like in white balance for emerald slider)... I think it's some kind of misguided api design in v1 making it's way to the end of times because nobody thought "somebody in the future might need a sane and robust way of showing/hiding selected widgets dynamically" :) Other than that I guess it's the question to gtk people :)

Yes, `no_show_all` prevents `show_all` from propagating all the way down the tree. But that might result in some of the leaves not receiving any `show`s at all, and therefore not showing up ever. So doing a `show_all` first, then setting `no_show`s and then setting visible should make it work. And then there's the question of whether your widget is already connected to the top-level tree or not. In gui_init we are just building the module tree, but it gets connected later...

What I'm doing here is _disconnecting_ widgets subtrees that were working perfectly before and then _reconnecting_ them below a different, newly created, widget. What mostly seems to work is calling `show` (instead of `show_all`) on the new container (box/grid/notebook). Then the sliders show up (because they were already `show`n before). But besides the transplanted lists of sliders, there are _also_ some new labels. So those new labels won't show up until they are explicitly `show`n. `gtk_widget_show` calls everywhere!

For now, the blunt force `show_all` and re`hide` is simpler and works. I just don't think it is very elegant.

TurboGit (approved):
Looks good and works for me. Thanks.

TurboGit:
A little paragraph and possibly some screenshot for the user's manual would be nice.

elstoc:
A bit late to the party but I've just played with it a bit now it's been merged. I think some visual indicator on the labels (perhaps the expander arrow as used in whitebalance) might prevent "accidental label click" bug reports (perhaps it could change direction when you change mode). Also the preference options might make more sense as "list/tabs/columns" rather than "box/notebook/grid".

dterrahe:
Yeah, those started out as internal flags and then adding them to darktableconfig.xml.in exposed them externally...

This is more similar to the slider coloring switch in white balance than to the expander section. The "scene illuminant temp" label has no further indication that clicking on it will have any effect. I think in this case, the effect is so dramatic that it will not escape notice and hopefully people will still remember the last thing they did was accidentally click somewhere and then try if clicking again will reverse the effect. And otherwise the option is there in settings. And there is a tooltip. And there is no lost functionality.

aurelienpierre:
The lift/gamma/gain tabs names are inaccurate if used in slope/offset/power mode.

PR #6271, opened 2020-09-11
Trying to fix #6246

### LINKED ISSUE #6246: Coredump, double free detected in tcache 2, master d6338aaf2
openSUSE Tumbleweed 20200907
NVIDIA GF106 [GeForce GTS 450], 390.138
darktable-3.3.0~git589.d63318aaf2
OpenCL loaded but not available

imported ~500 images, applying tag to about half selected
get coredump:
    [xmp_import] creating tag: Second Half
    free(): double free detected in tcache 2
    Aborted (core dumped)

coredump attached
err, too large, see:
  http://wahoo.no-ip.org/~paka/core.darktable.1000.a651b386468741b9b24b66be5d32f8a4.1168.1599620702000000000000.lz4.

jenshannoschwalm:
Referring to late discussions in issue #6246 

I tried **very** hard to get trigger a crash with dt master cd9c486 on fedora32, intel G2288, lots of ram and a nvidia driver using opencl or not. Compiled fresh with all defaults, **lua enabled but without scripts included or in use**, so this is an open question.

 I removed a whole directory with 700 photos, reimported to be sure there are no thumbnails with pre-existinmg xmp files or without. There were images with faulty data, from different cameras including some fuji images, jpegs, tiffs.

Nothing unexpected happend, i simply couldn't get dt crashing or locking. So i am stuck with a reproducer.

I looked at the whole code demosaic.c there are a few minor complaints from my side,
1) in `reload_defaults` ther is no init for `d->median_thrs`
2) the slider doesn't get a default set.

For both i can't imagine this to be a real problem but just to make sure ...

About @TurboGit remarks referring to `dt_bauhaus_combobox_remove_at()` about also freeing the data, i looked up the combo-box code, yes sure but i think the code here is truely correct.

```
  g->demosaic_method_bayer = dt_bauhaus_combobox_from_params(self, "demosaicing_method");
  for(int i=0;i<6;i++) dt_bauhaus_combobox_remove_at(g->demosaic_method_bayer, 5);
  
  g->demosaic_method_xtrans = dt_bauhaus_combobox_from_params(self, "demosaicing_method");
  for(int i=0;i<5;i++) dt_bauhaus_combobox_remove_at(g->demosaic_method_xtrans, 0);
  ```
We build two full comboboxes having **all** methods and remove those for each box unwanted.

One last point, i don't understand the reason for
```
  dt_accel_register_combobox_iop(self, FALSE, NC_("accel", "method (bayer)"));
  dt_accel_register_combobox_iop(self, FALSE, NC_("accel", "method (xtrans)"));
```
and
```
  dt_accel_connect_combobox_iop(self, "method (bayer)", GTK_WIDGET(g->demosaic_method_bayer));
  dt_accel_connect_combobox_iop(self, "method (xtrans)", GTK_WIDGET(g->demosaic_method_xtrans));
```

as any keyboard shortcuts make no sense. Could this be a problem?

So - i am stuck here. Any further ideas?

### DIFF
--- a/src/iop/demosaic.c
+++ b/src/iop/demosaic.c
@@ -187,8 +187,6 @@ int default_colorspace(dt_iop_module_t *self, dt_dev_pixelpipe_t *pipe, dt_dev_p
 void init_key_accels(dt_iop_module_so_t *self)
 {
   dt_accel_register_slider_iop(self, FALSE, NC_("accel", "edge threshold"));
-  dt_accel_register_combobox_iop(self, FALSE, NC_("accel", "method (bayer)"));
-  dt_accel_register_combobox_iop(self, FALSE, NC_("accel", "method (xtrans)"));
   dt_accel_register_combobox_iop(self, FALSE, NC_("accel", "color smoothing"));
   dt_accel_register_combobox_iop(self, FALSE, NC_("accel", "match greens"));
 }
@@ -198,8 +196,6 @@ void connect_key_accels(dt_iop_module_t *self)
   dt_iop_demosaic_gui_data_t *g = (dt_iop_demosaic_gui_data_t *)self->gui_data;
 
   dt_accel_connect_slider_iop(self, "edge threshold", GTK_WIDGET(g->median_thrs));
-  dt_accel_connect_combobox_iop(self, "method (bayer)", GTK_WIDGET(g->demosaic_method_bayer));
-  dt_accel_connect_combobox_iop(self, "method (xtrans)", GTK_WIDGET(g->demosaic_method_xtrans));
   dt_accel_connect_combobox_iop(self, "color smoothing", GTK_WIDGET(g->color_smoothing));
   dt_accel_connect_combobox_iop(self, "match greens", GTK_WIDGET(g->greeneq));
 }
--- a/src/iop/demosaic.c
+++ b/src/iop/demosaic.c
@@ -5002,7 +5002,7 @@ void gui_update(struct dt_iop_module_t *self)
 
 void reload_defaults(dt_iop_module_t *module)
 {
-  dt_iop_demosaic_params_t *d = module->default_params;
+  dt_iop_demosaic_params_t *d = (dt_iop_demosaic_params_t *)module->default_params;
 
   if(dt_image_is_monochrome(&module->dev->image_storage))
     d->demosaicing_method = DT_IOP_DEMOSAIC_PASSTHROUGH_MONOCHROME;
@@ -5013,6 +5013,7 @@ void reload_defaults(dt_iop_module_t *module)
 
   d->color_smoothing = 0;
   d->green_eq = DT_IOP_GREEN_EQ_NO;
+  d->median_thrs = 0.0f;
 
   module->hide_enable_button = 1;
 
@@ -5022,6 +5023,12 @@ void reload_defaults(dt_iop_module_t *module)
   else
   {
     module->default_enabled = 0;
+  }
+
+  dt_iop_demosaic_gui_data_t *g = (dt_iop_demosaic_gui_data_t *)module->gui_data;
+  if(g)
+  {
+    dt_bauhaus_slider_set_default(g->median_thrs, 0.0f);
   }
 }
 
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by wpferguson** [comment]
I've been working on this all day.  The version that made this happen with I call darktable33, which is just a daily build of git master that I use to test features and play with the latest.  It's an outgrowth of darktable31.  I have a version that I use for when I code up stuff called dttest, which is a pretty clean environment.  I add some debug statements trying to figure this out, and like you couldn't get it to crash.  After hours of trying various things I stumbled across the thought that maybe the darktable33 environment somehow got polluted, whether from carryover from darktable31 or something else.  I moved .config/darktable33 to .config/darktable33.old and started darktable33 with no configuration.  I then set my usual settings and imported the 11,000+ images that I play with.  I couldn't make it crash.  So then I started darktable33 with the .config/darktable33.old and crashed it within a minute.  Used darktable33 with .config/darktable33 on the same data set and couldn't crash it.  So, my conclusion is that this is the result of a somewhat polluted configuration environment.

File: src/iop/demosaic.c [hunk 1]
```diff
@@ -5023,6 +5024,12 @@ void reload_defaults(dt_iop_module_t *module)
   {
     module->default_enabled = 0;
   }
+
+  dt_iop_demosaic_gui_data_t *g = (dt_iop_demosaic_gui_data_t *)module->gui_data;
```
TurboGit, line 5028: gui stuff should not be part of reload_defaults(), it should be in gui_update()

dterrahe:
> 1. in `reload_defaults` ther is no init for `d->median_thrs`

this actually gets set in the default init() function (which gets implicitly used here); it takes the value from the introspection tags in the struct definition:
`  float median_thrs; // $MIN: 0.0 $MAX: 1.0 $DEFAULT: 0.0 $DESCRIPTION: "edge threshold"
`

> 2. the slider doesn't get a default set.

`  g->median_thrs = dt_bauhaus_slider_from_params(self, "median_thrs");
`
uses the default in default_params, which has been set by the above mentioned dt_iop_default_init()

actually, the explicit (re)settings in reload_defaults of
```
  d->color_smoothing = 0;
  d->green_eq = DT_IOP_GREEN_EQ_NO;
```
are superfluous and could/should be removed.

src/iop/demosaic.c
dterrahe: This is not completely correct. reload_defaults gets called both in headless and gui mode and if any settings (for example defaults) should be changed on a per-image basis, this should happen here.

The problem is that upon switching to darkroom, reload_defaults is called with a newly created gui widget that has not yet been added to the right panel. This means it behaves slightly differently from when it later gets called if the image is changed from the filmstrip. Maybe this is because it has not been realized and mapped yet, or maybe a `show` is called when it is added, so that any `hide`s are undone (which is a common per-image action if some widget is not applicable).
Because of this, some of the once-per-image actions that should be in reload_defaults are put in update_gui, so that they actually work.

I'm working on a follow up to #6240 that hopefully succeeds in removing this last issue in init/reload/update separation. But since it will mess with initialisation order, I'm worried about corner cases and will need a lot of testing to cover all of them.

jenshannoschwalm:
Thanks for telling me @dterrahe and @TurboGit

src/iop/demosaic.c
TurboGit: @dterrahe : ok I though that we already did a gui_upadet() after reload_defaults() if a GUI was set. So yes I'm looking forward to your next clean-up, this will certainly clarify the responsibility of each hook.

dterrahe: We _do_, so that's why putting everything in gui_update works. But sometimes the per-image changes are quite substantive (completely rebuilding comboboxes based on long lists of presets and maybe having to query databases etc) so doing them for every single gui_update would just not be efficient and it shouldn't be necessary. So the intuition would be to put that in reload_defaults (in my interpretation this just doesn't have gui_ in the name because it also gets called for non-gui). But if you do that, at the moment some things subtly break (but only on first switching to darkroom). While others work. So following intuition can be frustrating/time-wasting/bug-causing.

I'm hoping I can make the intuition just work. The fact that then some _current_ implementations can maybe also be optimised a bit is very secondary to the goal of making people waste less time in future on trying to understand all this. And fixing bugs.

PR #6278, opened 2020-09-12
styles: properly set the target image id.

### LINKED ISSUE #6275: bug with styles since Friday 09-11-2020
Problem between version darktable_3.3.0 ~ git605.a685d2730_amd64.deb
and Commits on Sep 11, 2020

Here is the problem:

applying a style to an image works but spas a second time. the second time darktable hangs.

In a terminal I have the following message:
[image_cache_allocate] failed to open image 4294967295 from database: no more rows available

I took in apt-cache the DT version of Thursday 10 and there no problem.

DT Master version on Ubuntu 20.04 updated.
Nvidia GeForce GTX 970M graphics card
drive 440,100

TurboGit:
Fixes #6275.

### DIFF
--- a/src/common/styles.c
+++ b/src/common/styles.c
@@ -700,6 +700,8 @@ void dt_styles_apply_style_item(dt_develop_t *dev, dt_style_item_t *style_item, 
   {
     dt_iop_module_t *module = (dt_iop_module_t *)calloc(1, sizeof(dt_iop_module_t));
 
+    module->dev = dev;
+
     if(dt_iop_load_module(module, mod_src->so, dev))
     {
       module = NULL;
@@ -816,6 +818,7 @@ void dt_styles_apply_to_image(const char *name, const gboolean duplicate, const 
     dt_dev_init(dev_dest, FALSE);
 
     dev_dest->iop = dt_iop_load_modules_ext(dev_dest, TRUE);
+    dev_dest->image_storage.id = imgid;
 
     // now let's deal with the iop-order (possibly merging style & target lists)
     GList *iop_list = dt_styles_module_order_list(name);


TurboGit:
@dterrahe : this revert a part of clean-up done by 23dc3a9980. This is needed for styles (and preset as it was documented).

johnny-bit:
I think overal fix would be to prevent calling to reload defaults from preset/styles entirelly...

TurboGit:
I'm not sure, we probably need reload_defaults() but not the part where a different default is used for some specific images (where we need an imgid). So maybe a better fix would be to ensure we have the imgid in dev structure set when applying a style. I can look at this...

TurboGit:
Better fix proposed, probably cleaner. Thoughts?

johnny-bit:
Now it's not a revert :) so title can be canged and if this is the case then indeed it looks like a better fix

TurboGit:
@johnny-bit : the title of the commit is ok :)

TurboGit:
@dterrahe : no more a revert of anything. I've merged.

dterrahe:
I should have pointed out, I guess, that this was a change I made that would need specific testing, sorry.

Historically, default initialisation was spread out between init() and reload_defaults(). In the sense that some fields did not get initialised _at all_ unless reload_defaults was called. So it made sense that reload_defaults was _always_ called after init. Which was a bit silly. My assumption was that this was the only reason why anything to do with styles and presets would call reload defaults without providing an image. In my opinion doing so would be wrong, because the only purpose of reload_defaults is to adjust the defaults depending on the image. If there is no image...

From quickly looking at this, indeed a fix elsewhere was needed, but now all those checks in reload_defaults if there is no image are not necessary anymore, correct?

TurboGit:
> From quickly looking at this, indeed a fix elsewhere was needed, but now all those checks in reload_defaults if there is no image are not necessary anymore, correct?

That's **should be** correct... I'm not yet sure for presets. Fact is that it is not because I have this working that it works in all cases. But in general yes, so we could certainly try to remove all those checks and see what is happening.

dterrahe:
In 23dc3a9 I already removed the call to reload_defaults from init_presets, and (most of?) the checks within reload_defaults to see if there was an image as well...

I'm not sure I understand why styles could be image dependent?

TurboGit:
In short, style are very complex beast to apply properly to an image. It was already done directly on the db but to handle multiple-instances, or instance with the same name on the style and on the target image, etc... that style was moved to a new algorithm. The style is now applied using a dummy dev environment to have styles applied from lighttable working using the same code when applying from darkroom. And this dummy dev initialize the modules list, loading them and so at this stage the reload_default() is called to initialize each module. Hence the need to set the image id on the dummy dev as done in this PR to avoid breakage on the pipe.

dterrahe:
> I'm not sure, we probably need reload_defaults() but not the part where a different default is used for some specific images (where we need an imgid).

So I tried to clean up all those instances and move them into init() so that reload_defaults never needs to be called to just have a valid fully initialised module; init() should be sufficient for that.

This was one of those cases where the code seemed to be doing extra things _just-to-make-sure (TM)_ and where, once this was locked in, other parts of the code started to depend on this illogical behavior.

dterrahe:
> And this dummy dev initialize the modules list, loading them and so at this stage the reload_default() is called to initialize each module.

That (implicit) call should probably not be required and disabled; these are the bits that I'm looking at the clean up the final inconsistencies, but I keep getting distracted (not just with dt). I meant to separate the really hazardous changes in a separate PR but had misjudged (or forgotten) the risk of 23dc3a9

The need for a dummy dev suggests that maybe a bit more refactoring is needed?

There are also parts that are only used in upgrading from old versions of params. Harder to test, because I'd need to feed a database that has old versions _for each module_ (ideally). Is there such a dataset anywhere? My first database doesn't go further back than early 2019, I think.

johnny-bit:
> There are also parts that are only used in upgrading from old versions of params. Harder to test, because I'd need to feed a database that has old versions for each module (ideally). Is there such a dataset anywhere? My first database doesn't go further back than early 2019, I think.

Since old releases are still available building dt 2.2 (or even older) should be "no biggie" :)

dterrahe:
Sorry by the way if I sound like I think this is all easily solved and if I'm messing with stuff that I don't understand. I _don't_ understand all of it (and I don't think anybody does anymore, maybe?) so I'm trying to simplify and fix so it all becomes easier to understand again. I'm sorry it does end up breaking stuff (and usually when I'm away from the computer for a bit) but it is almost impossible to make things better without doing so.

I'm wondering if maybe I should run an _experimental_ branch and ask more people to hammer on that without master immediately being impacted when the test team is expanded. Any thoughts greatly appreciated.

johnny-bit:
Master being hammertime is imo ok-ish. What we do need is "stable" branch where hammered out polished fixes can go and do something like 3.4.10 by oct 2021 ;) But that's hard maintenance cost and Pascal already has a lot on the plate. Testing on master gives chance to catch more bugs than if there were an "unstable" branch, because people would be more affraid to test.

dterrahe:
You mean a LTS branch? Pascal is already maintaining the separate branches for released versions (not for 3.2 I think because it is too short until xmas and 3.4 anyway to release any 3.2 updates).

I do think testing on master is ok _ish_ but many people probably use it as their daily (me kindof included).

TurboGit:
> The need for a dummy dev suggests that maybe a bit more refactoring is needed?

I don't think so. The dummy dev is there to use the same routines from lighttable and darkrroom. This seems correct to me and frankly I don't have an idea to do better or differently even if good part of dt should be rewritten.

But yes, we can run an experimental branch with some earthquake changes :) One could be to run reload_defaults() only if an image id is available for the module as this is needed for demosaic, filmicrgb, colorin and possibly others.

Do you have more in mind?

And I agree with you that there is probably nobody knowing very well this part, and this may be true for a good part of dt code :)

dterrahe:
> Since old releases are still available building dt 2.2 (or even older) should be "no biggie" :)

just trying with 2.0, but as expected I seem to be running into problems with too new library/compiler versions. might all be fixable, or not.

dterrahe:
> Do you have more in mind?

I'm going to take it easy on this one and not rush it. Having a test database of old module params versions will be useful for testing, but also I need to learn more about the styles and presets functionality so I can maybe spot more errors myself.

> earthquake changes

It's more like "I put the engine back together but got some leftover parts; does anybody need those?"

PR #6284, opened 2020-09-13
[Issue #6283] WB Presets Sony ILCE-7RM4 with finetuning

### LINKED ISSUE #6283: WB Presets for Sony ILCE-7RM4 with finetuning
Attached is the WB preset info for Sony ILCE-7RM4 that covers the full range of finetune settings.
[darktable-whitebalance-20200913.tar.gz](https://github.com/darktable-org/darktable/files/5213420/darktable-whitebalance-20200913.tar.gz)

matt-maguire:
Resolves #6283 

Re-did the WB presets for Sony ILCE-7RM4 camera, with all the fine tunings included.

There seem to be some inconsistencies with the presets I previously extracted and submitted. It seems I previously may have had some setting in my camera that was influencing the WB coefficients being written into my raw files. To check, I did a full factory reset of my camera, and the coefficients I get after the reset are consistent with the ones I extracted for this PR.


matt-maguire:
I think I found the reason for the discrepency -- when I first submitted the basic WB presets, I was using my 35mm f1.8 prime lens, but when I took the sample images for this PR, I was using my 24-105 f4 zoom lens. I can reproduce the earlier coefficients by swapping the lens. The coefficients recorded don't seem to be impacted by len correction settings in the camera.

TurboGit:
Ok, thanks for the explanations.

parafin:
Probably the best idea is to take them without attached lens, or at least with a 3rd party one.

matt-maguire:
That's a novel idea, let me try and see how the values compare.

matt-maguire:
Ok, so shooting without a lens is giving values consistent with what I originally got with the 35mm f1.8 prime lens. Let me capture a full set of parameters with no lens, and I'll push out an update to this PR.

PR #6286, opened 2020-09-13
[Issue #6283] WB Presets Sony ILCE-7RM4 with finetuning (no lens)

### LINKED ISSUE #6283: WB Presets for Sony ILCE-7RM4 with finetuning

matt-maguire:
Resolves #6283 

Previously PR #6284 implemented whitebalance presets for Sony ILCE-7RM4 with fine tuning parameters, but it was done with a Sony 24-105mm F4 G-lens attached, and it seems that this causes some lens-specific correction to be applied to the whiteblance coefficients before writing them into the RAW file. 

In order to keep the darktable WB presets for this camera lens-independent, I regenerated the sample images with no lens attached, and this PR is to replace the 24-105mm lens-specific presets with these generic lenless ones.


TurboGit:
Does that means that all our WB coefficients could be plagued by this ?

parafin:
Might be, though it's the first camera I hear which does this.

TurboGit:
Thanks parafin ! A bit worrisome but nothing we can do.

matt-maguire:
I added a remark to the WB preset wiki page that we recommend to make the test shots with no lens attached. It seems the magnitude of this effect on the Sony camera is only worth about one notch of adjustment on the finetune adjustment slider, so it's probably not worth breaking our heads over.

PR #6289, opened 2020-09-13
Bug fix for grid+ histogram scaling with log scale in tonecurve, basecurve

GrahamByrnes:
There is a bug with the log-scale slider in tonecurve & basecurve. If you edge it just off zero to take a value <1, you will see the grid and histogram jump left, instead of right. Look carefully and you'll see that this doesn't effect the curve itself.

It's due to the loglogscale variable being fed into the draw programs: the line scale takes >0 as meaning log, the draw programs take the value as being the base, so it is too small by 1.0.

Fixed by adding some 1.0f's


TurboGit:
Reproduced and this fixed it for me on my side. Thanks!

GrahamByrnes:
Cool.

PR #6291, opened 2020-09-13
Fix the PPC64LE build.

RobbieAB:
I would prefer not to set the NO_WARN_X86_INTRINSICS globally, as I believe it is better that we know explicitly where it is being set. This delightful mess of conditional include logic is required to make the code build in the scenario where it is not set globally.


TurboGit:
Works for me. Thanks.

PR #6303, opened 2020-09-15
map - images grouping to speed up display

phweyland:
Today map view displays each image even if hidden by others. To keep the display time acceptable the user can limit the number of images, 100 is good, 1000 is already a burden.

So this PR to group the images by proximity and display only one image per group. Additionally displays the number of images by group.

image

Still work in progress because:
- when ones moves to map, the map is displayed 5 times. There may be a bit of mess on current signaling to be cleaned up.
- relationship between geo-coordinates and pixels is not stable (in particular when a border of the map appears on the screen)


TurboGit:
Yeah! Thanks @phweyland for polishing the map view. Looks promising.

phweyland:
@TurboGit 
I'm quite happy with the speed (with 30.000 geotagged images / 50.000) despite the fact it remains a lot a useless triggering (maybe due to osm-gps-map internals).
There are still some old issues I haven't been able to fix so far (missing tiles, undo, ...) 

But I think it deserves your review and certainly some more tests.
Thanks

TurboGit:
Looks nice and way better than current implementation.

I've always found the vignette on the map too small to be very useful. I would try to double the size and eventually group them when they are a bit farther to avoid all thumbs to overlap. Not sure how this will render... You may find something better.

TurboGit:
BTW, it looks like your thumbs are bigger than mine on your screenshot above, see mine:

image

phweyland:
> BTW, it looks like your thumbs are bigger than mine on your screenshot above, see mine:

I've made a screenshot with a small window (not full screen).


> I've always found the vignette on the map too small to be very useful.

I agree the size is a bit too small. Currently the thumb size is fixed. It could be possible to make it proportional to screen size ? Or to make it configurable ?

phweyland:
> and eventually group them when they are a bit farther to avoid all thumbs to overlap.

As done the group size is half the thumb size. This allows a maximum overlap of half a thumbnail. We could avoid completely the overlap making the group size equal to thumb size. Or make the ratio configurable ?

TurboGit:
Proportional to screen size (taking into account DPI) seems like the best option to avoid another preference.

phweyland:
> Proportional to screen size (taking into account DPI)

I locate the DPI things. Is the screen size (or factor) already known in dt ? Just to be sure before using something like gdk_window_get_scale_factor().

TurboGit:
I think DT_PIXEL_APPLY_DPI should work in that case.

phweyland:
> I think DT_PIXEL_APPLY_DPI should work in that case.

Probably not enough, because already applied (line 873)...

phweyland:
Maybe a bit too large now ?

phweyland:
BTW white count corresponds to images exactly at the same location, yellow otherwise.

TurboGit:
@phweyland : Large ok but maybe not too much but the numbers are really too small!

It seems that the thumbs have a black border, right? Maybe add a white one to ensure proper visibility on all backgrounds?

TurboGit:
What would be nice is to be able to right+click (or ctrl+click) on a thumbs and be able to go to the corresponding filmstrip in the "collection filter" module. More interactivity with the map would be nice I think.

phweyland:
> Large ok but maybe not too much but the numbers are really too small!

So between 64 and 128 for the thumbs ? What would be the right value ?
About the font size, I've taken the same as in the timeline. Font size = 10. Should 12 the right size ?

> Maybe add a white one to ensure proper visibility on all backgrounds?

I don't really see the benefit here.


> at would be nice is to be able to right+click (or ctrl+click) on a thumbs and be able to go to the corresponding filmstrip in the "collection filter" module. More interactivity with the map would be nice I think.

100% agreed. It was my intention to have another PR to create "location" collections based on the geotag. And one of the way to create a new one can be as you say.

phweyland:
> > Maybe add a white one to ensure proper visibility on all backgrounds?
> 
> I don't really see the benefit here.

We could add a white border between image and black border to highlight a group. And remove the number when the image is alone.

For the next 10 days I may not have access to my keyboard....
There is no hurry, but I think there are a lot of small adjustments still to do, which could be done with other PRs.

TurboGit:
> About the font size, I've taken the same as in the timeline. Font size = 10. Should 12 the right size ?

Yes, maybe 12 will be better, 10 is really too small to be read easily.

> There is no hurry, but I think there are a lot of small adjustments still to do, which could be done with other PRs.

Agreed, let's change the font and I'll review again and merge if all ok. This is already a nice improvement no need to hold it longer.

phweyland:
Font-size on thumbnails set to 12.

TurboGit:
Thanks Philippe ! Looks good now to me.

Issue #5596, opened 2020-06-21
filmic vs base curve preferences → both modules enabled/disabled


timur-davletshin:
Linux, 64 bit, latest master snapshot

1. Make sure that display-referred is enabled in dt preferences → processing
2. Discard history of sample image
3. Open that image, make sure that base curve module is enabled
4. Leave to lighttable mode and change preferences to scene-referred
5. Open same image for editing
6. Click "original" in history tab and press "compress history stack"
7. See both "base curve" and "filmic rgb" enabled in new history.
8. In opposite situation (changing scene referred to display referred) no modules are enabled after clicking "original" and "compress history stack".

Expected behavior: enable modules according selected preferences.

elstoc:
Did you restart after changing the preference (as advised in the tooltip)?

timur-davletshin:
If restart is really required, old logic should be applied before that. But we have "all enabled" or "all disabled".

aurelienpierre:
> If restart is really required, old logic should be applied before that. But we have "all enabled" or "all disabled".

If C was really a programming language, we would discover these kind of errors at compilation time. But here we are…

Question is: what happens if you follow the guidelines ? I don't care how the soft should fail if user doesn't comply with guidelines, at that point the daktable's pipeline is 10 years old and has been hot-plugged with masks, multi-instances and pipe reordering in ways it was not designed for, so, yeah, brittle stuff is going on and nobody understands the pipeline anymore.

timur-davletshin:
> If C was really a programming language, we would discover these kind of errors at compilation time. But here we are…

Brian Kernighan is still alive...

ralfbrown:
That reminds me of something which is surprising behavior more than a bug (it's actually doing exactly what it's been told): if you choose scene-referred workflow and leave "auto-apply per-camera basecurve" set, you get both filmic and basecurve enabled on initial open if dt has a basecurve for the specific camera model that took the picture.

matt-maguire:
Fixed in #6304

TurboGit:
Closing as fixed.

PR #6304, opened 2020-09-15
[Issue #6287] Fix scene-referred workflow for Digikam XMP

### LINKED ISSUE #6287: Basecurve and Sharpen applied although "scene referred" workflow is activated
**Describe the bug**
The modules "basecurve" and "sharpen" are applied to a freshly imported RAW file (here Nikon), even though "scene-referred" workflow is activated (see screenshot "settings") and auto-apply basecurve and auto-apply sharpen is set inactive. No presets exist that would auto-apply these modules.
It seems as if this issue depends on a sidecar being available when importing the RAW first time into digikam. 

**To Reproduce**
I narrowed this down to the "conclusion" that the issue seems to occur as soon as a sidecar was written by Digikam for the RAW file we want to import into digikam, e.g. by applying a rating, adding tags or similar.

- if the RAW file was **not** e.g. rated in Digikam prior to import in darktable (and thus the RAW file has **no side-car file** written by Digikam at the time of import), everything works fine: only the expected modules are auto-applied in darktable. See screenshot "notRatedInDigikamBeforeImportToDarktable.png"
- if the RAW file was e.g. rated in Digikam (and thus **has a side-car file** written by Digikam at the time of import), then besides filmic, also basecurve and sharpen are applied, exposure is missing in this case. See screenshot "ratedInDigikamBeforeImportToDarktable.png"

**Expected behavior**
The desired modules should be applied during import independent from a side-car file being there or not.

**Screenshots**
notRatedInDigikamBeforeImportToDarktable.png
grafik
ratedInDigikamBeforeImportToDarktable.png
grafik
settings.png
grafik

**Additional information**
When choosing "discard history" in lighttable for the incorrect processed image, and then changing back to darktable again, the settings are applied correctly. 

 - Darktable Version: 3.2.1 (current, official installer)
 - Digikam Version: 7.0.0 (current, official installer)
 - Windows 10 x64 pro (1909)

matt-maguire:
Fixes #6287 

When digikam generates an XMP file for a raw image, and then that image+XMP file is imported into darktable, then if darktable
has scene-referred workflow set in the preferences, the scene-referred default presets are not applied. Instead, the display-referred presets get applied, except for basecurve.

This PR is to make a check whether the xmp file has ever been imported into darktable before, and if has not, then don't assume the XMP is a legacy darktable XMP. Rather, a new XMP that hasn't been been imported to darktable before should be treated as a fresh image with no legacy presets.


TurboGit:
Sounds good to me. Thanks.

matt-maguire:
Fixes #5596

PR #6311, opened 2020-09-16
[l10n] try to set lang to match selected by user, based on parafin's 428b3f3

johnny-bit:
Tis is just continuation of @parafin's work that left a `TODO`. I've tested this and managed to get Portugese date abbreviations on system that has preferred PL:EN prefered but PT installed :)

### DIFF
--- a/src/common/l10n.c
+++ b/src/common/l10n.c
@@ -36,67 +36,58 @@
 
 static gchar* _dt_full_locale_name(const char *locale)
 {
-  // TODO: check if this works on Windows
-  // this gets languages in order of preference of user
-  // usually with full extent of charset preference
-  // also - it does include "C" lang.
-  const gchar * const * locales = g_get_language_names();
-  int i = 0,j = 0;
-  while(locales[i])
+#ifdef __APPLE__
+  return dt_osx_full_locale_name(ui_lang);
+#endif
+#ifdef __linux__
+  gchar *output = NULL;
+  GError *error = NULL;
+  // This would likelly work on linux and macos
+  if(!g_spawn_command_line_sync("locale -a", &output, NULL, NULL, &error))
   {
-    gchar **variants = g_get_locale_variants(locales[i]);
-    while(variants[j])
+    if(error)
     {
-      if(g_str_has_prefix(variants[j], locale))
+      fprintf(stderr, "couldn't check locale: '%s'\n", error->message);
+      g_error_free(error);
+    }
+  }
+  else
+  {
+    if(output)
+    {
+      gchar **locales = g_strsplit (output, "\n", -1);
+      g_free(output);
+      int j = 0;
+      while(locales[j])
       {
-        // return first found variant - this is most likelly the best one
-        gchar *ret=g_strdup(variants[j]);
-        g_strfreev(variants);
-        return ret;
+        if(g_str_has_prefix(locales[j], locale))
+        {
+          // return first found variant - this is most likelly the best one
+          gchar *ret=g_strdup(locales[j]);
+          g_strfreev(locales);
+          return ret;
+        }
+        j++;
       }
-      j++;
     }
-    g_strfreev(variants);
-    i++;
   }
-  // If we're here, this means user chose locale that isn't present in prefered
-  // languages/locales for user, but maaaybe the locale is installed
-  j=0;
-  gchar **variants = g_get_locale_variants(locale);
-  while(variants[j])
-  {
-    if(g_str_has_prefix(variants[j], locale))
-    {
-      gchar *ret=g_strdup(variants[j]);
-      g_strfreev(variants);
-      return ret;
-    }
-    j++;
-  }
-  g_strfreev(variants);
-  // locale isn't installed, have fun :D
   return NULL;
+#else
+  // TODO: check a way to do above on windows
+  return NULL;
+#endif
 }
 
 static void set_locale(const char *ui_lang, const char *old_env)
 {
   if(ui_lang && *ui_lang)
   {
-#ifdef __APPLE__
-    char* full_locale = dt_osx_full_locale_name(ui_lang);
-    if(full_locale)
-    {
-      g_setenv("LANG", full_locale, TRUE);
-      free(full_locale);
-    }
-#else
     gchar *full_locale = _dt_full_locale_name(ui_lang);
     if(full_locale)
     {
       g_setenv("LANG", full_locale, TRUE);
       g_free(full_locale);
     }
-#endif
     g_setenv("LANGUAGE", ui_lang, TRUE);
     gtk_disable_setlocale();
   }
--- a/src/common/l10n.c
+++ b/src/common/l10n.c
@@ -42,7 +42,6 @@ static gchar* _dt_full_locale_name(const char *locale)
 #ifdef __linux__
   gchar *output = NULL;
   GError *error = NULL;
-  // This would likelly work on linux and macos
   if(!g_spawn_command_line_sync("locale -a", &output, NULL, NULL, &error))
   {
     if(error)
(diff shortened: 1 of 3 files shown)


parafin:
Out of the interest, which value of LANG do you get, and what does `locale -a` show on your system?

johnny-bit:
locale -a:
```
$ locale -a
C
C.utf8
en_US.utf8
pl_PL.utf8
POSIX
pt_PT.utf8
```
checking with system env set to just pl_PL.utf8:
```
pl_PL.utf8:
  pl_PL.utf8
  pl_PL
  pl.utf8
  pl
pl_PL:
  pl_PL
  pl
pl.utf8:
  pl.utf8
  pl
pl:
  pl
C:
  C
```

asking code for `en_US` gets me:
```
locales en_US: 
  en_US.utf8
  en_US
  en.utf8
  en
```

You can check how it behaves by dropping this piece of code right before `return 1` in `darktable.c` line 579 and start `darktable --version`
```c
        const gchar * const * locales = g_get_language_names();
        int i=0;
        printf("locales avail: \n");
        while(locales[i])
        {
          printf("%s:\n", locales[i]);
          int j=0;
          gchar **variants = g_get_locale_variants(locales[i]);
          while(variants[j])
          {
            printf("  %s\n", variants[j++]);
          }
          g_strfreev(variants);
          i++;
        }
        int j=0;
        printf("locales en_US: \n");
          gchar **variants = g_get_locale_variants("en_US.utf8");
          while(variants[j])
          {
            printf("  %s\n", variants[j++]);
          }
          g_strfreev(variants);
          j=0;
        printf("locales fr: \n");
          variants = g_get_locale_variants("fr");
          while(variants[j])
          {
            printf("  %s\n", variants[j++]);
          }
          g_strfreev(variants);
          j=0;
        printf("locales pt_PT: \n");
          variants = g_get_locale_variants("pt_PT");
          while(variants[j])
          {
            printf("  %s\n", variants[j++]);
          }
          g_strfreev(variants);
```

The problem I have is i'm not 100% sure about 2 things: will this work in windows and does it check existence of locale files or not (I honestly have bunch of locales).

johnny-bit:
I'll try to work on this a bit more as it pretty much now depends on user locale preference setting being correct (eg if user has 3 langiages set in order of precedence this will correctly set the language in dt among those 3. if not - problem). Also I think i went overboard with `g_get_locale_variants` because it is just string op, not full locale check :/

johnny-bit:
@parafin: after a lot of digging... "the only sane way" seems to be a call to `locale -a` and scrape it's output for proper locale name with proper definition. there's no library call for that nor any alternative I'm affraid. so the order of things would be to check locale preferences if those match what we need, failing that - call locale -a. and that's ONLY on linux i think.

Are you ok with me pursuing this direction or just "not worth it"?

parafin:
"locale -a" will also work on macOS (so my macOS-specific code can be removed then), but obviously not for Windows. I tried to avoid calling another program from dt, but aside from Windows-incompatibility it might not be a bad idea after all. Though I don't think any other part of dt does it. I don't have any strong feelings either way, so decide for yourself.

johnny-bit:
there are 3 other places calling external apps :) so this would be 4th :)

parafin:
I'll check your code on macOS. What are 3 other places that call external apps I wonder?

johnny-bit:
> What are 3 other places that call external apps I wonder?

1. spawning audio player in lighttable
2. spawning turboprint in print module
3. spawning xdg-email in email export

src/common/l10n.c
```diff
@@ -34,20 +34,60 @@
...
 
+static gchar* _dt_full_locale_name(const char *locale)
+{
+#ifdef __APPLE__
+  return dt_osx_full_locale_name(ui_lang);
+#endif
+#ifdef __linux__
+  gchar *output = NULL;
+  GError *error = NULL;
+  // This would likelly work on linux and macos
```
parafin, line 45: likelly -> likely, but it does work on macOS, and I assume on most Linux installation too, so just remove this comment I think.

johnny-bit: OK, done :+1:

parafin:
I've tested it on macOS, and, as expected, it works as expected;) So please apply the following patch too:
```
diff --git a/src/common/l10n.c b/src/common/l10n.c
index 29c3c0f98..51b5962b8 100644
--- a/src/common/l10n.c
+++ b/src/common/l10n.c
@@ -36,10 +36,7 @@
 
 static gchar* _dt_full_locale_name(const char *locale)
 {
-#ifdef __APPLE__
-  return dt_osx_full_locale_name(ui_lang);
-#endif
-#ifdef __linux__
+#if defined(__linux__) || defined(__APPLE__)
   gchar *output = NULL;
   GError *error = NULL;
   // This would likelly work on linux and macos
diff --git a/src/osx/osx.h b/src/osx/osx.h
index 225285601..acf850cc6 100644
--- a/src/osx/osx.h
+++ b/src/osx/osx.h
@@ -31,7 +31,6 @@ gboolean dt_osx_file_trash(const char *filename, GError **error);
 char* dt_osx_get_bundle_res_path();
 void dt_osx_prepare_environment();
 void dt_osx_focus_window();
-char* dt_osx_full_locale_name(const char* locale);
 
 #ifdef __cplusplus
 }
diff --git a/src/osx/osx.mm b/src/osx/osx.mm
index e66cdddc4..a629b9a0a 100644
--- a/src/osx/osx.mm
+++ b/src/osx/osx.mm
@@ -236,39 +236,6 @@ void dt_osx_focus_window()
   [NSApp activateIgnoringOtherApps:YES];
 }
 
-char* dt_osx_full_locale_name(const char* locale)
-{
-  @autoreleasepool
-  {
-    if(!strcmp(locale, "C"))
-    {
-      // C isn't in the list of locales
-      return strdup("C");
-    }
-    char *saved_locale = strdup(setlocale(LC_ALL, NULL));
-    NSString* locale_str = [NSString stringWithUTF8String: locale];
-    NSArray<NSString*>* locales = [NSLocale availableLocaleIdentifiers];
-    for(NSString* item in locales)
-    {
-      NSLocale* locale_ns = [NSLocale localeWithLocaleIdentifier: item];
-      NSString* locale_c = [NSString stringWithFormat: @"%@_%@", [locale_ns languageCode], [locale_ns countryCode]];
-      if([locale_c hasPrefix: locale_str])
-      {
-        const char* locale_utf8 = [locale_c UTF8String];
-        if(setlocale(LC_ALL, locale_utf8))
-        {
-          setlocale(LC_ALL, saved_locale);
-          free(saved_locale);
-          return strdup(locale_utf8);
-        }
-      }
-    }
-    // haven't found any matching locale
-    free(saved_locale);
-    return NULL;
-  }
-}
-
 // modelines: These editor modelines have been set for all relevant files by tools/update_modelines.sh
 // vim: shiftwidth=2 expandtab tabstop=2 cindent
 // kate: tab-indents: off; indent-width 2; replace-tabs on; indent-mode cstyle; remove-trailing-spaces modified;
```

johnny-bit:
> I've tested it on macOS, and, as expected, it works as expected;) So please apply the following patch too

Done!

parafin:
So to me it looks like ready for merge.

johnny-bit:
@TurboGit - this is ready :)

src/common/l10n.c
```diff
@@ -34,20 +34,56 @@
 #include <windows.h>
 #endif
 
+static gchar* _dt_full_locale_name(const char *locale)
+{
+#if defined(__linux__) || defined(__APPLE__)
+  gchar *output = NULL;
+  GError *error = NULL;
+  if(!g_spawn_command_line_sync("locale -a", &output, NULL, NULL, &error))
```
TurboGit, line 42: Just curious, don't we have an API for querying the locale ?

TurboGit: Ok, forgot about this I see it has been discussed in the comments.

TurboGit:
Works for me. Thanks!

PR #6322, opened 2020-09-18
image infos -> image information

elstoc:
change "infos" to "information" in the darktable preferences window
(informations is not a word)

also minor change to tooltip text


TurboGit (approved):
Thanks !

PR #6324, opened 2020-09-18
Improve thumbnails contrast in all themes

### LINKED ISSUE #6197: slide colors too similar between selected and unselected photos
I'm referring to ligthttable view, but to filmstrip in the bottom in darktable and map views too.

Darktable uses three different grays for unselected, selected and hovered photo slides.

I consider that the three grays are too close one another, and they should be more differentiated, I usually I struggle to distinguish them.

Well, actually the _hovered_ gray is quite recognizable, while the selected/unselected ones are really very similar.

Nilvus:
Fix #6197 and so improve thumbnails contrast in all themes. Thanks to @paolobenve help.


TurboGit:
Works for me, all details count ! Thanks.

PR #6326, opened 2020-09-18
make sure xmp metadata are imported when write_sidecar_files is checked

### LINKED ISSUE #6295: Lost title when removing image and reimporting from sidecar
I had to remove and reimport a directory, and lost all titles on the files. I don't know exactly what metadata is lost. Title is gone. Tags seem to be kept, number of stars too

To reproduce:
Remove an image from the collection
Import it again
Title is gone (from the XMP too)

This occurs on DT 3.2.1 archlinux. 

I didn't try downgrading DT yet to see when this appeared.

phweyland:
fixes #6295 
when write_sidecar_files is set:
- options are not shown win import dialog
- xmp metadata are always imported


TurboGit:
This is good to me and will avoid many issues I think. Thanks for working on this.

PR #6332, opened 2020-09-19
dynamic header buttons optimisation and manual description

dterrahe:
Is this too much detail? Should I add newline/paragraph markers for the options/optiongroups?
I don't know how to test what this looks like locally.


dterrahe:
Optimised the implementation of "active" and "dim" modes. These are now no longer dependent on the extra widgets for variable width. Those get dynamically installed and removed when switching modes now. This means there's only one codepath for "always" now, rather than it being dependent on whether another mode had been previously active.

TurboGit:
Thanks !

PR #6334, opened 2020-09-19
ellipse lib module names at end instead of middle

dterrahe:
this should be different from iop module names, where multi-instance names are only distinguishable by the ending. In general though, PANGO_ELLIPSIZE_END is better readable, because it only cuts off one word instead of two.

See, @johnny-bit? No real reason to hide right-side buttons.

image

image

image

### DIFF
--- a/src/develop/develop.c
+++ b/src/develop/develop.c
@@ -2437,7 +2437,7 @@ gchar *dt_history_item_get_name_html(const struct dt_iop_module_t *module)
   if(!module->multi_name[0] || strcmp(module->multi_name, "0") == 0)
     label = g_strdup_printf("%s", module->name());
   else
-    label = g_strdup_printf("%s <span size=\"smaller\">%s</span>", module->name(), module->multi_name);
+    label = g_markup_printf_escaped("%s <span size=\"smaller\">%s</span>", module->name(), module->multi_name);
   return label;
 }
 
--- a/src/develop/imageop.c
+++ b/src/develop/imageop.c
@@ -1243,7 +1243,7 @@ static void _iop_panel_label(GtkWidget *lab, dt_iop_module_t *module)
   gchar *label = dt_history_item_get_name_html(module);
   gchar *tooltip = g_strdup(module->description());
   gtk_label_set_markup(GTK_LABEL(lab), label);
-  gtk_label_set_ellipsize(GTK_LABEL(lab), PANGO_ELLIPSIZE_MIDDLE);
+  gtk_label_set_ellipsize(GTK_LABEL(lab), !module->multi_name[0] ? PANGO_ELLIPSIZE_END: PANGO_ELLIPSIZE_MIDDLE);
   g_object_set(G_OBJECT(lab), "xalign", 0.0, (gchar *)0);
   gtk_widget_set_tooltip_text(lab, tooltip);
   g_free(label);
--- a/src/libs/lib.c
+++ b/src/libs/lib.c
@@ -1056,7 +1056,8 @@ GtkWidget *dt_lib_gui_get_expander(dt_lib_module_t *module)
   hw[DT_MODULE_LABEL] = gtk_label_new("");
   gtk_label_set_markup(GTK_LABEL(hw[DT_MODULE_LABEL]), module->name(module));
   gtk_widget_set_tooltip_text(hw[DT_MODULE_LABEL], module->name(module));
-  gtk_label_set_ellipsize(GTK_LABEL(hw[DT_MODULE_LABEL]), PANGO_ELLIPSIZE_MIDDLE);
+  gtk_label_set_ellipsize(GTK_LABEL(hw[DT_MODULE_LABEL]), PANGO_ELLIPSIZE_END);
+  g_object_set(G_OBJECT(hw[DT_MODULE_LABEL]), "xalign", 0.0, (gchar *)0);
   gtk_widget_set_name(hw[DT_MODULE_LABEL], "lib-panel-label");
 
   /* add reset button if module has implementation */


dterrahe:
Actually, why not _only_ middle-ellipsize iop module names when there's actually a good reason you want to see the end, namely because it has been edited or there are multiple instances:
image

In _all_ other cases, which are way more common, end-ellipsizing is preferable.

What do you think @elstoc @Nilvus ? any objections from an aesthetic  perspective?

elstoc:
I think if you're going to do that it would be better to split the label in two and ellipsize only the module name and not the duplicate name. So your examples might read "denoise (bilater... 2". Then it would look consistent.

johnny-bit:
I'm with @elstoc on this one.

Also I'd like to repeat that I didn't said that hiding lib buttons is required to enhance label visibility, but only for visual consistency with iop modules 😋

dterrahe:
> you're going to do that it would be better to split the label in two and ellipsize only the module name and not the duplicate name.

I hadn't really noticed that the original/first module in multi-instance actually gets a name "0" that then isn't shown. Using that (i.e. with even simpler code) I get this:
image

That looks good to me. Your suggestion of not ellipsizing the instance name, besides being more complicated to implement, would also make the panel not shrinkable anymore if someone enters a really long name...

> Also I'd like to repeat that I didn't said that hiding lib buttons is required to enhance label visibility, but only for visual consistency with iop modules 😋

I know!!! But I'm much more bothered by the arrow button on the left-hand side that is different. Or the visual indication that reset and/or presets are not supported by the module. And why only two buttons? Can we please put in a third with some weak excuse? And there isn't a module icon when I select a theme like darktable-icons?? Somebody should fix that _with the highest urgency_!! 😋😋

Meanwhile I can spend time on implementing functionality that for 5/19 buttons _cannot_ be triggered (the labels are too short anyway) and for the remaining only get triggered in extreme cases (I recently got the impression even @elstoc isn't _that_ extreme) and even then only add _very_ little benefit. ("export ..." -> "export selected"; oh _now_ I can find the module I want, which is always bottom right anyway).

***dterrahe** calming down again

Nilvus:
> I think if you're going to do that it would be better to split the label in two and ellipsize only the module name and not the duplicate name. So your examples might read "denoise (bilater... 2". Then it would look consistent.

I totally agree with @elstoc. Nothing more to add there.

Nilvus:
> That looks good to me. Your suggestion of not ellipsizing the instance name, besides being more complicated to implement, would also make the panel not shrinkable anymore if someone enters a really long name...

Entering a long name in duplicate instances is user responsibility. I could be wrong but I don't think we have to take care of that. We need to take care of what we propose to the user, not what he will do with that. Or, at worst, maybe ellipsize duplicate name separately after a number of characters (something like 10 or 12 at least). 
 
> > Also I'd like to repeat that I didn't said that hiding lib buttons is required to enhance label visibility, but only for visual consistency with iop modules yum
> 
> I know!!! But I'm much more bothered by the arrow button on the left-hand side that is different. Or the visual indication that reset and/or presets are not supported by the module. And why only two buttons? Can we please put in a third with some weak excuse? And there isn't a module icon when I select a theme like darktable-icons?? Somebody should fix that _with the highest urgency_!! yumyum

You know that I'm also caring a lot about consistency but I think that lib and iop modules are not really the same. We already have differences between them, on the left and on right buttons. So I think that not having same behavior there on lib and iop remain consistent as it's already the case. So I agree with you @dterrahe. I think we can let lib modules as they are actually.

elstoc:
> That looks good to me. Your suggestion of not ellipsizing the instance name, besides being more complicated to implement, would also make the panel not shrinkable anymore if someone enters a really long name...

As @Nilvus says I don't think that's an issue. Always ellipsizing the module name at the end seems the more consistent behaviour to me

dterrahe:
@Nilvus @elstoc it is not exactly clear to me what your proposed solution is so unfortunately I can't try to implement it. Please give concrete examples with the desired result. If you suggest a tailor made ellipsizing algorithm just for this particular case, then I'm worried that this would involve adding a separate label widget to the header just for the multi instance name and capturing (several?) signals to make it behave as you want. Unfortunately gtk doesn't offer a plugin system for such algorithms. If you have a good idea on how to implement this, I'm looking forward to your PR. In the mean time, I consider the current behavior a bug and that this PR fixes that bug. It doesn't result in perfect behavior for the specific case of multiple instances, but it also isn't worse. In fact, it is unchanged. When your solution is finished it can then be merged to replace my (temporary) fix.

@turbogit before and after snapshots, first for the first commit
image image

and then for the second.
image image

I think that both changes lead to improved readability for the common case, even in this extreme example, and no worse readability/no change for the less common.

Nilvus:
The idea is simple: ellipsize only at the end of module name and not by including duplicate instance name set by user. 
About making a PR, not possible for me. My only skills in coding for darktable are on CSS.

elstoc:
The idea is that the module label is split into two: `[module-operation][duplicate-name]` rather than the current single label with just different text sizes to distinguish between its component parts. `[module-operation]` is always ellipsized at the end; `[duplicate-name]` is never ellipsized and so takes priority. Yes that would lead to a minimum widget size but that's under the user's control.

So, for example, if I created a named instance of the contrast equalizer module it might be ellipsized as follows:

```
contrast equalizer some long name
contrast equali... some long name
contrast eq... some long name
cont... some long name
```

dterrahe:
> The idea is simple: ellipsize only at the end of module name and not by including duplicate instance name set by user.

You'd get the behavior that you currently see _while_ editing the name (try it; imagine the background color difference isn't there); the duplicate-name expands, the module name shrinks to just "..." and then the duplicate name causes the panel to be uncompressable and scroll out of view to the right.
image

Now, as you say, this is under the user's control, and if they edit the name while the panel is narrow, they may notice the impact. But if they have created a few very long duplicate-names and _then_ try to shrink the panel, they will find it doesn't work. And it won't be obvious from the _..._, because the user doesn't know that only part of the label was shrinking. Worse, the offending label might not even be in view. I'd find that undesirable behavior and I would expect a few bug reports.

But again, if you want to implement it, please go ahead and submit. If I'm wrong and everybody likes it better, then that's fine. I'm just not going to spend time and effort, and add complexity to the code, to achieve a result that I think is inferior.

In the mean time, would you argue that this PR makes the current situation worse?

> There's no perfect solution and even in your original proposal might ellipsize to `contrast...ng name`.

You are absolutely right and then it would be obvious that to see the whole name, you'd have to increase the width of the panel. This is the situation we have now. That's why I think there will be people who would increase the panel width just to edit names and later when they try to decrease it again run into the situation I described above.

elstoc:
You make some good points and I don't really disagree with any of them. Personally, while I dislike ellipsizing in the middle, overall I probably prefer the consistency of ellipsizing everything the same which is where my suggestion came from.

But at the end of the day it's a minor look-and-feel thing, I don't feel particularly strongly either way and you've made some strong arguments against my suggestion: sometimes I'm wrong :)

Nilvus:
@dterrahe: you make some good points indeed there. Nothing more to say there.

dterrahe:
@TurboGit This seemed an appropriate place for a small bug fix that will help people who want to put "B&W" in a module instance name.

This is ready for your review; a consensus was reached in the discussion above.

TurboGit:
Works for me, all this is better sane defaults in the different context. Thanks.

Issue #5947, opened 2020-08-13
Lighttable previews rendered at low resolution on 4K screen


bastibe:
**Describe the bug**
In 3.2.1, previews in the lighttable render at a non-native resolution, as can be seen in the screenshot below.

This applies to all preview sizes, including the full-screen sticky preview, and makes judging focus very difficult. The previous release of Darktable (3.0.1) did render at the correct resolution.

The issue appears only in the lighttable, darkroom rendering is unaffected.

**To Reproduce**
Import images into Darktable, look at them in the lighttable.  
On a 4K screen.

**Expected behavior**
Image previews should render in the screen's native resolution, just like the rest of the UI, and in the darktable.

**Screenshots**

![Pixelated Lighttable](https://user-images.githubusercontent.com/649181/90115684-b6acf100-dd54-11ea-9fd0-1091934a0e07.png)

**Platform (please complete the following information):**
 - Darktable Version: 3.2.1
 - OS: Ubuntu 20.04
 - OpenCL active (disabling does not fix the issue)
 - AMD Vega M with RocM 2.9

cryptomilk:
What did you set `settings -> lighttable -> high quality thumbs processing form size` to?

bastibe:
I tried several, including "always" and "never", but it does not seem to make a difference. I restarted Darktable in between changes to "high quality thumb processing from size" as well.

bastibe:
The issue also persists in a pristine copy of Darktable right from the OBS latest release debian repo with my previous `~/.config/darktable` removed.

Here is at least one more user affected by this: https://discuss.pixls.us/t/lighttable-previews-pixelated-on-highdpi-screen/19690/3

cryptomilk:
Hmm, I don't have that issue.

bastibe:
Is there anything I can do to help to debug this issue?

A-Herold:
Had the same problem, but rolled back to 3.0.2 due to #5945. I experienced another related problem. For the same screen settings (3 images in row on a  a MacBook Pro Retina 13’’, 2560 x 1600) now cached previews of size 5 are required, before it was 4.

konig12:
I am seeing the same issue. I have high quality thumbs processing set to "Always". I have not touched this setting in months. The affect applies to images imported before and after the upgrade.

I'm happy to help with any debugging or diagnostics that would be helpful as well.

Darktable 3.2.1
Arch Linux with i3 window manager
Nvidia GTX 1050Ti
4k display

[...]

PR #6335, opened 2020-09-19
Make the cairo rendering quality option visible

jenshannoschwalm:
There are issues around like #5947, the cairo rendering quality is at least
one part of the problem. I think we should make the existing hidden option
visible.

### DIFF
--- a/data/darktableconfig.xml.in
+++ b/data/darktableconfig.xml.in
@@ -112,19 +112,6 @@
     <shortdescription>maximum width of the side panels in pixels</shortdescription>
     <longdescription>(needs a restart)</longdescription>
   </dtconfig>
-  <dtconfig>
-    <name>ui/cairo_filter</name>
-    <type>
-      <enum>
-	<option>fast</option>
-	<option>good</option>
-	<option>best</option>
-      </enum>
-    </type>
-    <default>fast</default>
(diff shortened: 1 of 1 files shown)


parafin:
I really feel that amount of options in preferences being added to dt is excessive. This is especially the kind of stuff that should work out-of-the-box without any fiddling in settings. Exposing it as combobox isn't a solution IMHO. Not a good one at least.

jenshannoschwalm:
> I really feel that amount of options in preferences being added to dt is excessive. This is especially the kind of stuff that should work out-of-the-box without any fiddling in settings

I absolutely agree here. Personally i think there are already too many options. 

Let's just have this dirty option available while debugging #5947 and revert after fully understanding whether this does really matter.

rgo:
> > I really feel that amount of options in preferences being added to dt is excessive. This is especially the kind of stuff that should work out-of-the-box without any fiddling in settings
> 
> I absolutely agree here. Personally i think there are already too many options.
> 
> Let's just have this dirty option available while debugging #5947 and revert after fully understanding whether this does really matter.

Maybe, it's a good a idea to hide the option in darktablerc without exposing it in preferences.
At the end, it's only for debugging purposes.

jenshannoschwalm:
ok, i drop this. Will have to find another way to discuss results on other peoples machines...

PR #6336, opened 2020-09-19
export lib: add ability to define export size from print size × resolution

aurelienpierre:
Long asked feature. Export size can now be specified in pixels, inches and centimeters.
Screenshot_20200919_173800
Screenshot_20200919_173746
@elstoc I have a problem with the `insert_text_handler()` function, it doesn't really work with floating point input, especially when we try to erase numbers, the cursor gets repositionned at the beginning. I don't really get what happens there, could you have a look ? Thanks !

TODO: 
* maybe write the print DPI in the `Exif.Image.XResolution` metadata tag of the exported file. (Anybody knows how ?)
* write those print dimensions in params for presets ? (not sure if it's needed, this is only an helper UI to write the actual pixel sizes, and these are the only ones that matter).

### DIFF
--- a/data/darktableconfig.xml.in
+++ b/data/darktableconfig.xml.in
@@ -1758,6 +1758,34 @@
     <longdescription>pixel interpolator used in rotation and lens correction (bilinear, bicubic, lanczos2, lanczos3).</longdescription>
   </dtconfig>
   <dtconfig>
+    <name>plugins/lighttable/export/dimensions_type</name>
+    <type>int</type>
+    <default>0</default>
+    <shortdescription>unit of the printing size</shortdescription>
+    <longdescription>unit in which to input the image size.</longdescription>
+  </dtconfig>
+  <dtconfig>
+    <name>plugins/lighttable/export/print_dpi</name>
+    <type>int</type>
+    <default>300</default>
+    <shortdescription>DPI</shortdescription>
+    <longdescription>print resolusion in DPI.</longdescription>
+  </dtconfig>
+  <dtconfig>
+    <name>plugins/lighttable/export/print_width</name>
+    <type>float</type>
+    <default>0</default>
+    <shortdescription>width of the exported image</shortdescription>
+    <longdescription>width of the exported image, or 0 if no scaling should be done.</longdescription>
+  </dtconfig>
+  <dtconfig>
+    <name>plugins/lighttable/export/print_height</name>
+    <type>float</type>
+    <default>0</default>
+    <shortdescription>height of the exported image</shortdescription>
+    <longdescription>height of the exported image, or 0 if no scaling should be done.</longdescription>
+  </dtconfig>
+  <dtconfig>
     <name>plugins/lighttable/export/width</name>
     <type>int</type>
     <default>0</default>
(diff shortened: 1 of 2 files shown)


TurboGit:
The control of the entry is strange. For example, I select cm I double click on the first entry, I want to enter 21 and in fact the entry contains 210. The 0 is added automatically and only if I type the dot (or coma) I have 21.0. This is the only entry working this way in dt!

TurboGit (changes requested):
Entry control strange behavior.

TurboGit:
Note also that we have two competing PR (see #6079) on the export module. I suppose this one will be in first and the other will have to be reworked based in this one.

aurelienpierre:
> The control of the entry is strange. For example, I select cm I double click on the first entry, I want to enter 21 and in fact the entry contains 210. The 0 is added automatically and only if I type the dot (or coma) I have 21.0. This is the only entry working this way in dt!

Yes, there is a problem with the text entry and numbers, that is hidden with integers (in the px entries).

This issue comes from the fact that the entry is text, but the validation layer forces conversion to float then back to text with 3 decimals. As long as you input numbers, the entry is treated as integers and the numbers are inserted before the 0 if any. Only inputting a . or , will force-set the numbers position.

I'm not sure what the best option would be. Text entries are not really suited for that work, but the spin buttons we had earlier seem too ugly. 

> Note also that we have two competing PR (see #6079) on the export module. I suppose this one will be in first and the other will have to be reworked based in this one.

I'm also in process of removing all the direct read-write to config file that makes a mess of this module. I discovered that, with this PR, current px sizes are not properly saved to config. But the incessant read-write to config make it impossible to track. So I have 3 functions, one that reads config at init time and save params to a struct, one that sanitizes the values, and one that writes the config at cleanup time, and everything else reads or writes in the params. That would also make the presets blobs encoding and decoding much easier… (notice that resetting the module is broken currently).

elstoc:
>  but the spin buttons we had earlier seem too ugly

Agreed, though the primary reason for their removal was the large minimum width the +/- buttons imposed on the right panel. Keeping the spin buttons with _three_ entries will be worse for that. Unfortunately GTK doesn't give us many options here :(

TurboGit:
> I'm also in process of removing all the direct read-write to config file that makes a mess of this module. I discovered that, with this PR, current px sizes are not properly saved to config. But the incessant read-write to config make it impossible to track.

Beware that those configs are used because using for example Ctrl+E from darkrroom does not give access to the export libs parameters.

TurboGit:
> This issue comes from the fact that the entry is text, but the validation layer forces conversion

Maybe just remove the validation for now. We can validate when leaving the entry.

TurboGit:
@aurelienpierre : any update on this, there is another PR waiting to be worked on top of this one and I want to avoid squashing too many things at the last minute (we are close to 1 month from code freeze) :)

aurelienpierre:
@TurboGit this lib is a clusterfuck of bad things, it turns out to be much more work than expected so I'm not sure I will be able to finish it before code freeze. Don't wait for me :-)

TurboGit:
Closing this PR, I'm working on a new one on top of this to fix the remaining issues.

PR #6338, opened 2020-09-19
fixed offset/power/slope headers in color balance

dterrahe:
As requested here https://github.com/darktable-org/darktable/pull/6269#issuecomment-695325239 by @aurelienpierre 

Also fixed tooltips, but not dynamically.


TurboGit:
Thanks !

PR #6340, opened 2020-09-20
Clamp mask values to [0,1]. Fix #6315

### LINKED ISSUE #6315: Strange blowouts when using Colour balance + non-opaque parametric mask together with filmic.
Originally reported in:
https://discuss.pixls.us/t/is-this-a-3-2-darktable-bug/20247/12

When using colour balance with a parametric mask, and the blend opacity is around 80% or so, and the mask contrast is reduced to -.65 or so, and filmicrgb is enabled (with v3.0 module ordering), then it is possible to get these weird blowout effects appearing on your screen if you fiddle with the feathering slider. If I turn off filmic, I can't get these artifacts to appear, nor if I set the colourbalance blend mode opacity to 100%.

**To Reproduce**
open the following image + XMP:
https://drive.google.com/drive/folders/1gRgCkQQWghDxfqPZOqTyZeBtJrLGW4xl?usp=sharing

or check out the post in pixls.us (link given above).

**Expected behavior**
It should not be possible to produce these sorts of blowouts with these settings.

**Screenshots**
refer to https://discuss.pixls.us/t/is-this-a-3-2-darktable-bug/20247/12

 - Darktable Version: 3.3.0+766~g82963f4f0
 - OS: Manjaro Linux 20.0.3
 - OpenCL in on (but makes no difference if I turn it off) 
 - NVIDIA TU117M [GeForce GTX 1650 Mobile / Max-Q] driver: nvidia v: 440.100

**Additional context**
Original poster reported this on darktable 3.2.1 on WIndows.

pitbuster:
Combinations of negative values on mask contrast and blend opacity may push mask values outside of [0,1], which may produce artifacts when doing the actual blending.

This should fix #6315

### DIFF
--- a/data/kernels/blendop.cl
+++ b/data/kernels/blendop.cl
@@ -1022,7 +1022,8 @@ blendop_mask_tone_curve(__read_only image2d_t mask_in, __write_only image2d_t ma
     scaled_opacity = (scaled_opacity + brightness) / (1.f + brightness);
     scaled_opacity = fmax(scaled_opacity, -1.f);
   }
-  opacity = ((scaled_opacity * e / (1.f + (e - 1.f) * fabs(scaled_opacity))) / 2.f + 0.5f) * gopacity;
+  opacity = clamp(
+      ((scaled_opacity * e / (1.f + (e - 1.f) * fabs(scaled_opacity))) / 2.f + 0.5f) * gopacity, 0.f, 1.f);
   write_imagef(mask_out, (int2)(x, y), opacity);
 }
 
--- a/src/develop/blend.c
+++ b/src/develop/blend.c
@@ -2840,7 +2840,8 @@ void dt_develop_blend_process(struct dt_iop_module_t *self, struct dt_dev_pixelp
           x = (x + brightness) / (1.f + brightness);
           x = fmaxf(x, -1.f);
         }
-        mask[k] = ((x * e / (1.f + (e - 1.f) * fabsf(x))) / 2.f + 0.5f) * opacity;
+        mask[k] = clamp_range_f(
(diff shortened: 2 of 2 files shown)


TurboGit:
@parafin @aurelienpierre : This looks correct, a mask should always have values between 0 and 1.  Ok to merge ?

parafin:
I'm not sure about how blending works in such details. I mean if it's expected that computed value can be outside of [0:1] boundaries here, then it makes sense to clamp indeed.

TurboGit:
@parafin : that's where I have question indeed, the blending using "normal" normal mode does not clamp but we have "normal bounded" which clamp the values. As I have reported in the issue using "normal bounded" does fix the issue. That's why I have a little doubt about clamping in normal mode... But at the same time I've hard time to understand why an opacity mask could have values outside of [0 , 1 ]. I've never working on the blending part of dt!

pitbuster:
Note that this does not clamp the image values, it clamps the values of the opacity mask (om) used in the formula: out = in * (1-om) + proc * om (where out is the output, in the input and proc the value obtained in the current module). Values of in and proc can be still outside [0,1], but om will not.

TurboGit:
@pitbuster : Yes I understand (at least I think I do) that clamping opacity mask to [ 0 , 1 ] looks ok and that the blending mode applies to the final pixel values. But since I'm not really expert on this part of dt I wanted to have confirmation and be sure we have explored possible side effects...

As I said this patch looks correct to me !

TurboGit:
Ok, more thinking and still think all this is good. All integration tests are passing. So I'll merge. Thanks a lot!

PR #6350, opened 2020-09-22
Overhaul color picker module

### LINKED ISSUE #6282: color picker module automatically changes to "area" mode
Open and expand the color picker module. Defaults to my last selected option "point":

Screenshot_2020-09-12_22-43-10

Click on the color picker icon. Option changes to "area":

Screenshot_2020-09-12_22-44-24

Same thing happens if I disable the picker, change it back to "point" and then re-enable the picker.

The only way to get to point mode is to enable the picker and then change it in the drop-down while the picker is enabled.

Also (possibly unrelated) it used to be possible to change the active color picker in an IOP module from point to area mode using this module (and the documentation states that it should) but this seems to no longer be possible.

dterrahe:
As promised here https://github.com/darktable-org/darktable/issues/6282#issuecomment-692161038

image

Removed broken functionality and streamlined module.
Improved the tooltip.
Works better with narrow panels.

Fixes #6282

### DIFF
--- a/src/libs/colorpicker.c
+++ b/src/libs/colorpicker.c
@@ -224,11 +224,11 @@ static gboolean _sample_tooltip_callback(GtkWidget *widget, gint x, gint y, gboo
 
   gchar **sample_parts = g_malloc0_n(12, sizeof(char*));
 
-  sample_parts[3] = g_strdup_printf("%22s(0x%02X%02X%02X)\n%s:", " ",
+  sample_parts[3] = g_strdup_printf("%22s(0x%02X%02X%02X)\n<b>%14s</b>", " ",
                                     (int)round(sample->rgb.red   * 255.f),
                                     (int)round(sample->rgb.green * 255.f),
                                     (int)round(sample->rgb.blue  * 255.f), _("RGB"));
-  sample_parts[7] = g_strdup_printf("\n%s:", _("Lab"));
+  sample_parts[7] = g_strdup_printf("\n<b>%14s</b>", _("Lab"));
 
   for(int i = 0; i < 3; i++)
   {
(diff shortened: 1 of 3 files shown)


src/libs/colorpicker.c
```diff
@@ -466,194 +452,165 @@ static void _set_sample_area(dt_lib_module_t *self, float size)
...
+  dt_bauhaus_widget_set_label(data->color_mode_selector, NULL, _("model"));
+  dt_bauhaus_combobox_add(data->color_mode_selector, _("RGB"));
+  dt_bauhaus_combobox_add(data->color_mode_selector, _("Lab"));
+  dt_bauhaus_combobox_set(data->color_mode_selector, dt_conf_get_int("ui_last/colorpicker_model"));
+  dt_bauhaus_combobox_set_entries_ellipsis(data->color_mode_selector, PANGO_ELLIPSIZE_NONE);
+  g_signal_connect(G_OBJECT(data->color_mode_selector), "value-changed", G_CALLBACK(_color_mode_changed), self);
+  gtk_widget_set_valign(data->color_mode_selector, GTK_ALIGN_END);
+  gtk_box_pack_start(GTK_BOX(picker_row), data->color_mode_selector, TRUE, TRUE, 0);
+
+  gtk_box_pack_start(GTK_BOX(picker_row), gtk_label_new("  "), FALSE, FALSE, 0);
```
TurboGit, line 552: This should be done in CSS if needed.

TurboGit (changes requested):

TurboGit:
The line with statistic & model is too crowded and in some language (here French) the label are ellipsised.

image

I think the statistic & model should be on 2 lines or the label "model" removed as Lab RGB is clear on this widget?

dterrahe:
Mostly I added the label, because without it, the leftmost combobox kindof floats in the air, because it is right aligned; "mean" doesn't start at the left border. I'll experiment a little with other layouts (in French). Removing "model" might not help, because it won't change the split allocated to the left combobox, I think.

dterrahe:
I'm just seeing #6352 pop up which overlaps significantly with this PR.

Let me know what you want to do with that. If you have the intention of merging it, it would make sense to do that first and wait with this one until I have rebased it incorporating the needed changes, rather than forcing msdm to deal with that.

Nilvus:
> Mostly I added the label, because without it, the leftmost combobox kindof floats in the air, because it is right aligned; "mean" doesn't start at the left border. I'll experiment a little with other layouts (in French). Removing "model" might not help, because it won't change the split allocated to the left combobox, I think.

Why not simply move picker under that line and then add a zone button. That would aerate text there and have a visual way to understand both possibilities: point and zone picker. I have understand that actually zone could be set with ctrl+click (tooltip missing) but could be simpler to have two buttons and so a more aerate UI, by remaining a better one than actual one.

src/libs/colorpicker.c
dterrahe: removed

dterrahe:
Removed the combobox labels, so they don't get ellipsized anymore.
Moved names to end of line in tooltip, which causes less problems with non-utf8 languages (Russian now looking fine).
Added button tooltip.

image

@Nilvus We don't have separate icons for point and area pickers and using ctrl is more consistent with elsewhere, but technically your suggestion would be possible. Especially after removing the labels, I don't feel aerating the text is necessary. Rather the opposite; I added the labels because in English the dialog is quite sparse.

TurboGit:
@dterrahe : I have already reviewed this PR, so #6352 will have to be reworked on top of this. It is rare to see such conflicts, but that's good news as it means quite a lot of changes in dt :)

TurboGit:
Ok, works for me. Really nice and functional. I really like the large color patch, really handy to visually evaluate the color. All goof now, merging, we'll have more field testing. Thanks.

PR #6352, opened 2020-09-22
Add LCh view mode to color picker

msdm:
This PR adds LCh support to the color picker module as a view mode.


TurboGit:
Now that #6350 has been merged you need to rebase and rework this PR. TIA!

dterrahe:
Is what is happening here useful outside of just the colorpicker left-panel module? This Lch transformation is intended/will be just for presentational purposes, right? Could it not simply be done in the colorpicker module itself, in _update_sample_label and _sample_tooltip_callback, rather than having to carry it forward through several function interfaces? Maybe I am mistaken that it is not computationally identical? The dt_Lab_2_LCH doesn't look heavy enough to require so much buffering.

msdm:
> Could it not simply be done in the colorpicker module itself, in _update_sample_label and _sample_tooltip_callback, rather than having to carry it forward through several function interfaces?
I'm absolutely OK with your suggestion. I just implemented it this way to keep color space calculations separated from the presentation logic. Though, it seems there is not a separation between such logics in the source code of darktable.

TurboGit:
Ok, so closing.

PR #6355, opened 2020-09-23
add lch presentation option to color picker module

dterrahe:
This seems to get similar results to #6352 but in a more localised way.

The tooltip is getting rather large with this addition...

### DIFF
--- a/src/libs/colorpicker.c
+++ b/src/libs/colorpicker.c
@@ -145,6 +145,7 @@ static void _update_sample_label(dt_colorpicker_sample_t *sample)
 
   // Setting the output label
   char text[128] = { 0 };
+  float alt[3] = { 0 };
 
   switch(model)
   {
@@ -159,6 +160,18 @@ static void _update_sample_label(dt_colorpicker_sample_t *sample)
     case 1:
       // Lab
       snprintf(text, sizeof(text), "%6.02f %6.02f %6.02f", CLAMP(lab[0], .0f, 100.0f), lab[1], lab[2]);
+      break;
+
+    case 2:
+      // LCh
+      dt_Lab_2_LCH(lab, alt);
+      snprintf(text, sizeof(text), "%6.02f %6.02f %6.02f", CLAMP(alt[0], .0f, 100.0f), alt[1], alt[2] * 360);
+      break;
+
+    case 3:
+      // HSL
+      dt_RGB_2_HSL(rgb, alt);
+      snprintf(text, sizeof(text), "%6.02f %6.02f %6.02f", alt[0] * 360, alt[1] * 100, alt[2] * 100);
       break;
   }
 
@@ -224,11 +237,11 @@ static gboolean _sample_tooltip_callback(GtkWidget *widget, gint x, gint y, gboo
 
   gchar **sample_parts = g_malloc0_n(12, sizeof(char*));
 
-  sample_parts[3] = g_strdup_printf("%22s(0x%02X%02X%02X)\n<b>%14s</b>", " ",
+  sample_parts[3] = g_strdup_printf("%22s(0x%02X%02X%02X)\n<big><b>%14s</b></big>", " ",
                                     (int)round(sample->rgb.red   * 255.f),
                                     (int)round(sample->rgb.green * 255.f),
                                     (int)round(sample->rgb.blue  * 255.f), _("RGB"));
-  sample_parts[7] = g_strdup_printf("\n<b>%14s</b>", _("Lab"));
+  sample_parts[7] = g_strdup_printf("\n<big><b>%14s</b></big>", _("Lab"));
 
   for(int i = 0; i < 3; i++)
   {
@@ -241,9 +254,9 @@ static gboolean _sample_tooltip_callback(GtkWidget *widget, gint x, gint y, gboo
                                       (int)round(CLAMP(rgb[1], 0.f, 1.f) * 255.f),
                                       (int)round(CLAMP(rgb[2], 0.f, 1.f) * 255.f), " ");
 
-    sample_parts[i + 4] = g_strdup_printf("<span foreground='red'>%6d</span>  "
-                                          "<span foreground='green'>%6d</span>  "
-                                          "<span foreground='blue'>%6d</span>  %s",
+    sample_parts[i + 4] = g_strdup_printf("<span foreground='#FF7F7F'>%6d</span>  "
+                                          "<span foreground='#7FFF7F'>%6d</span>  "
+                                          "<span foreground='#7F7FFF'>%6d</span>  %s",
                                           (int)round(rgb[0] * 255.f),
                                           (int)round(rgb[1] * 255.f),
                                           (int)round(rgb[2] * 255.f), _(name[i]));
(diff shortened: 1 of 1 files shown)


msdm:
I haven't tried this PR but if it gives similar results to mine (#6352), I think this one should be merged as it is simpler with fewer changes in the code. I was also planning to add HSL support and it would be nice if you could do that part as well.

dterrahe:
OK, I added HSL as well as a presentation option in the combobox and converted using dt_RGB_2_HSL. I have no idea if this is the right thing to do and if any scaling/clamping would be required.

However, I have two reservations here:
- I don't think it makes sense to add all possible transformations to the tooltip as well, since it would get huge. Maybe LCh shouldn't really be there either.
- Is this conceptually correct? i.e. does the min/max/mean after these transformation really correspond to a transformation of the rgb or lab statistics at the end, or should this be done at the pixel level? I have no opinion, just asking. If this is just for fun/comparability/indication of scale, then "accuracy" might not be that relevant. Doing at pixel level of course would require significantly more work (and possibly cause slowdown) so in that case the addition would need some justification.

msdm:
Thanks a lot

TurboGit:
>     * I don't think it makes sense to add all possible transformations to the tooltip as well, since it would get huge. Maybe LCh shouldn't really be there either.

Agreed, let's remove this from tooltip.

src/libs/colorpicker.c
```diff
@@ -160,6 +161,18 @@ static void _update_sample_label(dt_colorpicker_sample_t *sample)
       // Lab
       snprintf(text, sizeof(text), "%6.02f %6.02f %6.02f", CLAMP(lab[0], .0f, 100.0f), lab[1], lab[2]);
       break;
+
+    case 2:
+      // LCh
+      dt_Lab_2_LCH(lab, alt);
+      snprintf(text, sizeof(text), "%.02f %.02f %.02f", CLAMP(alt[0], .0f, 100.0f), alt[1], alt[2] * 360);
```
TurboGit, line 168: all format should be : %2.02f as LCh values can be > 10. otherwise the number does not align well.

src/libs/colorpicker.c
```diff
@@ -164,8 +164,14 @@ static void _update_sample_label(dt_colorpicker_sample_t *sample)
...
-      dt_Lab_2_LCH(lab, lch);
-      snprintf(text, sizeof(text), "%.02f %.02f %.02f", CLAMP(lch[0], .0f, 100.0f), lch[1], lch[2] * 360);
+      dt_Lab_2_LCH(lab, alt);
+      snprintf(text, sizeof(text), "%.02f %.02f %.02f", CLAMP(alt[0], .0f, 100.0f), alt[1], alt[2] * 360);
+      break;
+
+    case 3:
+      // HSL
+      dt_RGB_2_HSL(rgb, alt);
+      snprintf(text, sizeof(text), "%.02f %.02f %.02f", alt[0], alt[1], alt[2]);
```
msdm, line 174: The "H" channel (alt[0]) of HSL should be multiplied to 360 just like the "h" channel of LCh.

dterrahe:
Do we like a slightly bigger font for the RGB/Lab headers in the tooltip?

TurboGit:
All this looks good except the fact that the numbers are moving around because they are not using the same length in all color mode. Add three live sample and change the RVB, Lab, LCh... 

I would use everywhere: "%6.02f %6.02f %6.02f"

And probably also adjust the RGB format to be "%6d %6d %6d".

Last, the color for the RVB numbers makes the numbers hard to read. I would use standard white, better lisibility than "fun" effect to me. I understand that this has not been changed here, but maybe better to do that as were are at it.

TurboGit:
> Do we like a slightly bigger font for the RGB/Lab headers in the tooltip?

Looks ok to me.

dterrahe:
> I would use everywhere: "%6.02f %6.02f %6.02f"

Yes, definitely! Much calmer.

> And probably also adjust the RGB format to be "%6d %6d %6d".

Not sure; I see RGB as a double option, because it also makes the color patch larger, in case you don't care about the numbers.
If you really want to have all the numbers the same size, then I'd want to add another option "Patch", that simply doesn't print a number at all.

> Last, the color for the RVB numbers makes the numbers hard to read. I would use standard white, better lisibility than "fun" effect to me. I understand that this has not been changed here, but maybe better to do that as were are at it.

You are right (and I was wondering why you didn't bring that up before!) I have now changed these to brighter color number (#FF7F7F rather than just 'red'). I think this works quite well. Could go even brighter if needed. The only drawback now is that these colors this way are not themable, but they work well with all current themes imho.

I won't cry if you make me remove them after all, but I think it is fun and _almost_ useful this way.

TurboGit:
> and I was wondering why you didn't bring that up before!

I thought it would work ok, but after using it I found it not ok. So just changing my mind. I'll check the new colors on all themes and will come back.

TurboGit:
> then I'd want to add another option "Patch",

Not sure about this and I would still prefer to avoid jumping numbers and patch resized when going from RGB to Lab which are the two most commons options I'm using. But let's see what others have to say about that as it is really a matter of preference.

dterrahe:
Fair enough! You know I rely on you to catch my experiments that don't work. Means I don't have to self-censure (too much).

dterrahe:
> > then I'd want to add another option "Patch",
> 
> Not sure about this

I just tried it. Works quite well, especially with narrow panels when the text is unreadable anyway. And there's still the tooltip.

"None" is a better name. "Color Only" would be too long (for narrow panels). And there should be at least one character (to indicate where to aim for the tooltip). How about a hamburger? (☰) It might give the impression there is something clickable.

Bullseye? ◎ (U+25CE)

I like it. Can I have it? Please?

Completely separately, I had the idea of automatically adding a sample when clicking (or releasing after drawing an area) with CTRL held. It turns out this is somewhat problematic, because a continuous series of updates is sent to the function linked to darktable.lib->proxy.colorpicker.update_panel so many samples would be added. I know the + button can be linked to an accelerator, but that would permanently cost a key (and besides nobody does that before they decide they'd like it). This would need a bit of work, so I'm not sure if it is worth it.

dterrahe:
image

johnny-bit:
bullseye looks great!

TurboGit:
> This would need a bit of work, so I'm not sure if it is worth it.

I would say no, to me this is not needed and would pollute the color picker lib when trying multiple time to select the proper area.

I'll merge this as-is. I'm not sure "none" is really useful (at least not to me) as the point of the color picker is to get the actual values of the pointed area. But maybe this will be useful for others and it does not make the GUI more complex. So let's this in.

One thing I've just seen is that there is the text on the tooltip "hover to highlight sample on canvas" but there is nothing happening. This is not a regression due to this PR, so can be fixed in another PR. I suspect a regression due to the pipe rewrite/optimization done recently.

TurboGit:
Thanks for this nice revamp of the color picker. Since 3.0 I think we have reworked almost every GUI part of dt.

dterrahe:
> I'll merge this as-is. I'm not sure "none" is really useful (at least not to me) as the point of the color picker is to get the actual values of the pointed area. But maybe this will be useful for others and it does not make the GUI more complex. So let's this in.

This would basically just be an easy access to the tooltip which has all the details (more than just the one statistic). The alternative would be to automatically switch to the bullseye when the text becomes unreadably small, but that would be more complicated...

PR #6367, opened 2020-09-23
colorpicker: display live sample location even if display sample off.

TurboGit:
Only by overring the patch sample we now display the area on the
picture.


TurboGit:
@dterrahe : kind of usability fix to me, no need to display all samples to see where one specific sample is on the picture.

PR #6368, opened 2020-09-24
make whole color picker sample row show location

dterrahe:
#6367 is great and made getting this to work worth it.

Adding the event box was trivial. Figuring out that capturing GDK_NOTIFY_INFERIOR was necessary, otherwise moving over the color patch would just briefly flash the location indication, was a _pain_. Why is nothing in gtk easy?

;-)

Now the selected area will be highlighted _while_ hovering over the delete button!


TurboGit:
Works for me !

PR #6375, opened 2020-09-24
resize notebook tabs using natural width of ellipsized labels

dterrahe:
When notebook tab labels are ellipsized, they are all treated as having size 0 and are allocated equal width. This is suboptimal when using narrow panels; long labels start to ellipsize while short labels still have lots of space. This especially shows when very short and very long labels are mixed in the same notebook. (#4940)
One can attempt to solve the problem by setting a minimum size for the longest labels, but if done too aggressively this will prevent the notebook (and panel) from shrinking.

This PR uses the natural width of the labels to allocate width to them. The result is that they will be more evenly, later, or sometimes not at all, ellipsized.

image
image

With this commit and the switch to dt_ui_notebook_page as described here https://github.com/darktable-org/darktable/pull/4940#issuecomment-692861672 , the new channel mixer looks like this (at the point where it previously would prevent the panel from being further narrowed):
image
(it needs some of the checkbox buttons to be ellipsized too to be able to shrink further).

Other benificiaries:
image image image image

This is basically a bug fix/workaround for gtk. Initially I thought this was how ellipsized notebooks were supposed to work (I think other containers get resized/allocated this way) but I couldn't find the right options. Getting this fixed upstream looked like too daunting a path to me (especially with the focus on gtk4). If anybody thinks I should try that, please let me know.

### DIFF
--- a/src/gui/gtk.c
+++ b/src/gui/gtk.c
@@ -3005,15 +3005,60 @@ GdkModifierType dt_key_modifier_state()
   return state & gtk_accelerator_get_default_mod_mask();
 }
 
+static void notebook_size_callback(GtkNotebook *notebook, GdkRectangle *allocation, gpointer *data)
+{
+  int n = gtk_notebook_get_n_pages(notebook);
+
+  GtkRequestedSize *sizes = g_malloc_n(n, sizeof(GtkRequestedSize));
+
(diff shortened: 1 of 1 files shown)


TurboGit (approved):
Thanks !

PR #6383, opened 2020-09-26
[signal] allow to ignore print-trace if not available

johnny-bit:
Previously if darktable was compilled with print trace disabled, specifying print-trace on commandline debug options would fail. with this change it's ignored with meaningfull message to user.


TurboGit:
Looks good to me, thanks !

PR #6387, opened 2020-09-27
[colormapping] only clear src/target on clusters change

### LINKED ISSUE #6328: Color Mapping Module broken...
Running updated Windows 10. Responding to a query on facebook it was noted that after acquiring the source image and then the target image in the module that activating either of the sliders simple clears the color mapping information. This is noted for me on the current 3.2.1....will check later to see if this is still an issue on dev build but likely.....

johnny-bit:
During introspection change, the clusters reset got a bit overreaching, which caused cluster reset on any gui change. Now cluster reset is done only on cluster number change.

This fixes #6328


TurboGit:
Works for me! Thanks!

PR #6389, opened 2020-09-27
[masks] make module window height adjustable, same as other modules

### LINKED ISSUE #4838: the window of the mask manager extensible
I work with a lot of masks & the window of the mask manager is static. Is it possible to make it extensible as the left & right panels, il will be more comfortable & quicker. Thank you

johnny-bit:
Similarly to many other modules (code is basically a copypasta from other module) allow masks to have adjustable height

This fixes #4838

### DIFF
--- a/src/libs/masks.c
+++ b/src/libs/masks.c
@@ -1572,6 +1572,25 @@ static void _lib_masks_selection_change(dt_lib_module_t *self, int selectid, int
   lm->gui_reset = 0;
 }
 
+static gboolean _mouse_scroll_view(GtkWidget *treeview, GdkEventScroll *event, dt_lib_module_t *self)
+{
+  if (event->state & GDK_CONTROL_MASK)
+  {
+    dt_lib_masks_t *d = (dt_lib_masks_t *)self->data;
+    const gint increment = DT_PIXEL_APPLY_DPI(10.0);
+    const gint min_height = DT_PIXEL_APPLY_DPI(100.0);
+    const gint max_height = DT_PIXEL_APPLY_DPI(1000.0);
+    gint width, height;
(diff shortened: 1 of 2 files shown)


src/libs/masks.c
```diff
@@ -1572,6 +1572,25 @@ static void _lib_masks_selection_change(dt_lib_module_t *self, int selectid, int
...
+    dt_lib_masks_t *d = (dt_lib_masks_t *)self->data;
+    const gint increment = DT_PIXEL_APPLY_DPI(10.0);
+    const gint min_height = DT_PIXEL_APPLY_DPI(100.0);
+    const gint max_height = DT_PIXEL_APPLY_DPI(1000.0);
+    gint width, height;
+    gtk_widget_get_size_request (GTK_WIDGET(d->scroll_window), &width, &height);
+    height = height + increment * event->delta_y;
+    height = (height < min_height) ? min_height : (height > max_height) ? max_height : height;
+    gtk_widget_set_size_request(GTK_WIDGET(d->scroll_window), -1, (gint)height);
+    dt_conf_set_int("plugins/darkroom/masks/heightview", (gint)height);
```
TurboGit, line 1588: Don't we want to add this into darktableconfig.xml.in with a default value ? All conf values should be in darktableconfig.xml.in.

johnny-bit: I don't know - other libs with adjustable heights don't seem to use `darktableconfig.xml.in`... added just in case :)

johnny-bit:
Addressed comments made, should be fine :)

TurboGit (approved):
Good to me now, thanks !

dterrahe:
Only just noticed this; I've been working on this (and the other remaining libs modules) as well. Will probably submit tomorrow early, after rebasing off this and config changes.

PR #6390, opened 2020-09-27
Improve appearance of combobox menu

amousset:
Avoid horizontal space between menu border and the items to make the general appearance cleaner.

Before:
menua2
menua1

After:
menu2
menu3

I'm not sure if this is the cleanest way to achieve it in CSS though.


TurboGit:
Looks ok as this is what is done by default in GNOME applications. I'd still like to have @Nilvus feedback on this. Thansk.

Nilvus (approved):
Indeed, it's much better that way. And CSS is just great. All details count, thanks @amousset

TurboGit:
Thanks!

PR #6398, opened 2020-09-28
[presets] add option to hide built-in presets of iops

### LINKED ISSUE #1994: Capability to disable/hide hard-coded presets
Hard-coded presets cannot be removed. Lately, a lot of such presets have been added to the master branch that make the presets list of some iops lengthy, specially when someone has also a bunch of personal presets. I personally prefer my own presets and almost never use those hard-coded ones.

Add capability to hide/disable hard-coded presets. I think the presets section of preferences is a good place for this feature, so that a user can selectively hide the presets (s)he doesn't use.

johnny-bit:
Adds capability to hide/disable hard-coded presets. The option is shown in darkroom section since it hides only presets for darkroom modules :)

This fixes #1994

@msdm - I hope this would work for you :)

### DIFF
--- a/data/darktableconfig.xml.in
+++ b/data/darktableconfig.xml.in
@@ -1027,6 +1027,13 @@
     <longdescription>draw borders around grouped images when grouping is turned off and the mouse hovers over one of the images of the group</longdescription>
   </dtconfig>
   <dtconfig prefs="darkroom">
+    <name>plugins/darkroom/default_presets_first</name>
+    <type>bool</type>
+    <default>true</default>
+    <shortdescription>sort built-in presets first</shortdescription>
+    <longdescription>whether to show built-in presets first before user's presets in presets menu.</longdescription>
+  </dtconfig>
+  <dtconfig prefs="darkroom">
     <name>plugins/darkroom/hide_default_presets</name>
     <type>bool</type>
     <default>false</default>
--- a/data/darktableconfig.xml.in
+++ b/data/darktableconfig.xml.in
@@ -1026,6 +1026,13 @@
     <shortdescription>draw borders around grouped images</shortdescription>
     <longdescription>draw borders around grouped images when grouping is turned off and the mouse hovers over one of the images of the group</longdescription>
   </dtconfig>
+  <dtconfig prefs="darkroom">
+    <name>plugins/darkroom/hide_default_presets</name>
+    <type>bool</type>
+    <default>false</default>
+    <shortdescription>hide built-in presets</shortdescription>
+    <longdescription>hides built-in presets of modules in both presets and favourites menu.</longdescription>
+  </dtconfig>
 
   <dtconfig prefs="import" section="session">
     <name>session/base_directory_pattern</name>
--- a/data/themes/darktable.css
+++ b/data/themes/darktable.css
@@ -873,6 +873,13 @@ menuitem check
   margin: 0 0.5em;
 }
 
+menu separator 
+{
+  margin: 0.6em 2em;
+  padding-bottom: 1px;
+  background-color: @selected_bg_color;
+}
+
 /*----------------------
   - GTK Notebooks tabs -
   ----------------------*/
--- a/src/gui/presets.c
+++ b/src/gui/presets.c
@@ -795,6 +795,7 @@ void dt_gui_favorite_presets_menu_show()
   darktable.gui->presets_popup_menu = GTK_MENU(gtk_menu_new());
   menu = darktable.gui->presets_popup_menu;
   gboolean presets = FALSE; /* TRUE if we have at least one menu entry */
+  const gboolean hide_default = dt_conf_get_bool("plugins/darkroom/hide_default_presets");
 
   GList *modules = darktable.develop->iop;
   if(modules)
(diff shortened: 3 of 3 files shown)


TurboGit:
Well I'm not really convinced by this... Another option for this? What's wrong we default presets? Anyway let's see what others think about this.

johnny-bit:
Well, the request in #1994 is a request, @msdm and I thik @junkyardsparkle seemed interested in it... I think this option could be hidden so it doesn't spam the preferences win... And the only thing "wrong" with default presets expressed in #1994 was the list is long/too long and makes it hard to reach user's presets.

TurboGit:
> the list is long/too long and makes it hard to reach user's presets.

So what about add the user's presets first ? Also it could be nice to have a separator between user and system presets and also before the actions like (store new preset, favorite...). What do you think?

johnny-bit:
Separators would be OK imo, sorting user presets firs... Dunno really, because it's subjective - user's presets are now close to actions with them and more connected... Would have to be an option IMO :/ but obviously both sorting and hiding could be hidden options. I wouldn't mind having separators in menu though...

TurboGit:
Ok, let's try separators between user & system presets and before the actions at the bottom. I think this would be a nice usability fix and will continue our GUI revamp.

johnny-bit:
I found out something interesting: separators are hidden :) this necessitates css change if you want to show separators :)

junkyardsparkle:
Some visual separation seems like a good idea in any case, especially for the actions (the way the presets run right into those has always been a little bit weird). Sorting is complicated because the lists don't always "drop down", depending on the screen position. The only other thing I can think of would be allowing for nested submenus, which some users might find useful for organization in general... but I have no idea how difficult GTK3 would make this.

johnny-bit:
Submenus? It's possible and not "too bad" but how to reliably organize it is totally different problem.

as for separators - @Nilvus - if you have time, can you look into css setup for separator? currently it's disabled/hidden everywhere, but a small item with padding wouldn't be too bad, would it?

junkyardsparkle:
If there's a strong push against separators for whatever reason, then a subtle distinction such as a slightly lighter background color might be enough (or better than nothing, at least).

TurboGit:
In fact separators are implemented but not visible, to have them:

```
menu separator 
{
background: @selected_bg_color;
padding: 1px;
}
```

@Nilvus : can you validate this CSS snippet ? TIA.

johnny-bit:
tbh after playing with this - I'm not so fond of separators in form of light line in menu... or a very visible line... i tried following and it imo creates nice visual separation without going overboard:

```css
/* Separator */
menu separator 
{
padding: 1.5px 0px;
}
```

junkyardsparkle:
This is my preference, after trying a few things in elegant-dark theme:

menu separator 
{
  padding: 0.1em;
  background-color: @bauhaus_indicator_border;
}

I'm assuming that this color is defined in an equally subtle way in the other themes, but haven't checked... and I guess it makes more sense for it to be scaled to text rather than hardcoded pixels? Anyway, thanks for pointing out that the separators were actually there! :-)

TurboGit:
Thanks for all contributions on the CSS, @Nilvus will have the last word :)

Nilvus:
I don't see any separators with CSS added between users and built-in presets. Something missing in code?

I have tested all proposals and the best one is the last one by @junkyardsparkle I think. I just improve it as I prefer really thin separators.

Here's my proposal:

```
menu separator 
{
  margin-left: 2em;
  margin-right: 2em;
  margin-top: 0.6em;
  margin-bottom: 0.3em;
  padding-top: 0.05em;
  background-color: @bauhaus_indicator_border;
}
```

And what it's look like (with the issue about no separator between my user preset "GX8..." and built-in ones):

Nilvus:
> EDIT: I found `@bauhaus_indicator_border` to be visible in the other themes, but very subtle in some cases. :-)

Too subtle. Or need, to include in shade() and 1.4 for darker themes. Better if possible to have one ok in all themes. And selected is near bauhaus indicator color.

TurboGit:
> I don't see any separators with CSS added between users and built-in presets. Something missing in code?

It is not there, I will certainly work on that after this PR except if done by @johnny-bit .

@johnny-bit : can you rework this PR with @Nilvus proposal based on @junkyardsparkle one?

Nilvus:
@johnny-bit: just a little update on CSS to have better vertical alignement. And allow to simplify the code:

```
menu separator 
{
  margin: 0.6em 2em;
  padding-bottom: 1px;
  background-color: @selected_bg_color;
}
```

johnny-bit:
Done. If it's to everybody's linking? :)

TurboGit:
That's really really nice... Yes a very nice polishing of this preset menu. At the same time we also get the separator in the multi-instance menu. Thanks to all involved in this PR!

junkyardsparkle:
Looks great, thanks for the work!

msdm:
@johnny-bit, sorry for the late response, and thanks a lot for implementing this. Actually I had a different approach in mind to have finer control over built-in presets. It would be handy to be able to hide individual presets from the presets group in the preferences (maybe by adding a column of show/hide checkbox before the name column). This is because I do use some of the built-in presets in regular-basis.

johnny-bit:
Recent change from @AlicVB introduced very finegrained selection of presets in favourite modules presets menu, which alows one to build that list from scratch. Showing/hiding built-in presets in module presets menu would require additional sql magic I wasn't (and still am not) comfortable introducing due to how built-in preset are handled in db (they are re-created on every start afaik)

junkyardsparkle:
I get around this by simply making user presets from the few built-ins that I sometimes use. Maybe there could be added a quick and simple method for doing this.

PR #6400, opened 2020-09-28
Update or remove links pointing to the redmine wiki in the README

amousset:


README.md
```diff
@@ -194,8 +194,8 @@ then install it. This can usually be done in your package manager (not the appli
 customarily provided by default in your distribution) or on the internet with a search engine. 
 You may need to install a package manager first (like Synaptic on Debian/Ubuntu, or DNF Dragora on Fedora/RHEL).
 
-This process might be tedious but you only need to do it once. See this 
-[outdated page on building darktable 2.6](https://redmine.darktable.org/projects/darktable/wiki/Building_darktable_26)
+This process might be tedious but you only need to do it once. See this (partly outdated)
```
TurboGit, line 197: I'm not ok with the "partly outdated" here. I agree that it is, but this should not be part of the README.md we could add a notes on the Wiki itself maybe?

TurboGit (changes requested):
No need to state that the wiki is outdated.

amousset:
Indeed. BTW I also started actually updating the page (general build instructions and instructions for Fedora).

TurboGit (approved):
Thanks !

PR #6403, opened 2020-09-29
[conf] get default conf val in case normal/overriden is garbage

### LINKED ISSUE #6139: Please sanity check configuration values on input
I recently had a Debian bug [report] where the user somehow had an empty height setting for the histogram, and was mystified that their histogram didn't work. In order to avoid such problems (and bug reports) it would be nice if obviously nonsensical configuration values could be rejected. One idea that came up on IRC was clamping the numerical values to the ranges specified in the default `darktablerc`.

[report]: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=969137

johnny-bit:
This PR, along with conf changes in PR #6364 allow for sanity checking and clamping of conf values. This means it ultimately fixes  #6139

Now code depending on numerical values and value clamping can safely get sane values (either default, clamped or both), just needs to start using it, but that's for authors and other PRs if it's of any benefit to them :smile:

### DIFF
--- a/src/control/conf.c
+++ b/src/control/conf.c
@@ -72,6 +72,11 @@ fin:
   return str;
 }
 
+static inline char *dt_conf_get_default_var(const char *name)
+{
+  return (char *)g_hash_table_lookup(darktable.conf->defaults, name);
+}
+
 /* set the value only if it hasn't been overridden from commandline
  * return 1 if key/value is still the one passed on commandline. */
 static int dt_conf_set_if_not_overridden(const char *name, char *str)
@@ -125,7 +130,28 @@ int dt_conf_get_int(const char *name)
 {
   const char *str = dt_conf_get_var(name);
   float new_value = dt_calculator_solve(1, str);
-  if(isnan(new_value)) new_value = 0.0;
+  if(isnan(new_value))
+  {
+    //we've got garbage, check default
+    const char *def_val = dt_conf_get_default_var(name);
+    if(def_val)
+    {
+      new_value = dt_calculator_solve(1, def_val);
+      if(isnan(new_value))
+        new_value = 0.0;
+      else
+      {
+        char *fix_badval = g_strdup(def_val);
+        if(dt_conf_set_if_not_overridden(name, fix_badval))
+          g_free(fix_badval);
+      }
+    }
+    else
+    {
+      new_value = 0.0;
+    }
+  }
+
   int val;
   if(new_value > 0)
     val = new_value + 0.5;
@@ -138,7 +164,28 @@ int64_t dt_conf_get_int64(const char *name)
 {
   const char *str = dt_conf_get_var(name);
   float new_value = dt_calculator_solve(1, str);
-  if(isnan(new_value)) new_value = 0.0;
+  if(isnan(new_value))
+  {
+    //we've got garbage, check default
+    const char *def_val = dt_conf_get_default_var(name);
+    if(def_val)
+    {
+      new_value = dt_calculator_solve(1, def_val);
+      if(isnan(new_value))
+        new_value = 0.0;
+      else
+      {
+        char *fix_badval = g_strdup(def_val);
+        if(dt_conf_set_if_not_overridden(name, fix_badval))
+          g_free(fix_badval);
+      }
+    }
+    else
+    {
+      new_value = 0.0;
+    }
+  }
+
   int64_t val;
   if(new_value > 0)
     val = new_value + 0.5;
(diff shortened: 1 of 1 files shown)


src/control/conf.c
```diff
@@ -125,7 +130,28 @@ int dt_conf_get_int(const char *name)
 {
   const char *str = dt_conf_get_var(name);
   float new_value = dt_calculator_solve(1, str);
-  if(isnan(new_value)) new_value = 0.0;
+  if(isnan(new_value))
```
TurboGit, line 133: If we have no garbage and there is a min/max set in conf, can we clamp the value to ensure it is still in the right range? This would add another layer of safety.

johnny-bit: I did that already in #6364 by introducing `dt_conf_get_and_sanitize_*` set of functions. This puts a burden of clamping on code calling and adds very little overhead IMO. Otherwise conf.c would have to also load + parse xml with min/max/default which could be a bit over the top imo... I mean... It's possible, but not really needed in many cases, and cases that do need clamped values ca use `dt_conf_get_and_sanitize_*` class of calls :)

johnny-bit: TL;DR my explaination: me to dum to do dat ;(

TurboGit: Indeed, we don't have code generated for min/max in preferences.

johnny-bit:
TBH - I think this is probably as good as it gets with what we have. Next prs might be to either use `dt_conf_get_and_sanitize_*` or to add hard xml dependency...

TurboGit (approved):
Ok, so looks ok to me.

src/control/conf.c
dterrahe: > Indeed, we don't have code generated for min/max in preferences.

Haven't looked at all yet how all of this works, but rather than live reading the xml at run-time, could some validation code be generated that fixes up all values (clamping plus defaulting for garbage) in conf.c at load time? Once the whole darktablerc is read, it stays in memory so won't need to be revalidated again.

This doesn't yet cover the case where the code would want to use the min/max to do clamping on values that are generated some other way (i.e. not read from the config). It would be a bit unfortunate to have to maintain min/max both in the xml _and_ in the code (since they could easily get out of sync). Maybe again a statically generated lookup table, the code for which would be generated at compile time?

johnny-bit: another `generate_prefs.xsl` only for default config?  huh... that might be super painful...

dterrahe: Isn't there already something like that in case there is no darktablerc at all yet? I believe a new one, containing all the default values, is auto generated, presumably based on the xml, no? Don't know how any of this works. "xsl"! Brrr.

johnny-bit: the "default" mechanism relies on existence of "darktablerc" in darktable install datadir :) if that's missing then... it might not play nice :)

dterrahe: Ah. And that does not get autogenerated from the xml?

Generating validation code should be somewhat simpler than generating the preference dialog, no? But it has been a very long time since I looked at xsl. And that was from a long distance and between my fingers...

dterrahe: What would be needed is an xsl that generates just a long list of calls to 
dt_conf_get_and_sanitize_,  filling in the min/max values from darktableconfig.xml. This code would then be called from conf.c at startup. 
Actually, it would be better to extend "sanitize" so that it would fill in the default value, instead of the max, when there is no value at all.
With that, there would be no need to sprinkle calls to dt_conf_get_and_sanitize_ all around the code (and there would only be _one_ authoritative definition of the min/max values.

johnny-bit: That would work for predefined conf vals that are defined in darktableconfig.xml.in... some are dynamic etc. But sorta yeah, agree on generating some sort of safe conf

TurboGit: I'm working on a xsl to generate min/max/default and corresponding API.

johnny-bit: I have some ideas but I do trust you to come up with  something better :)

dterrahe: > I'm working on a xsl to generate min/max/default and corresponding API.

Please make it so it doesn't require changes everywhere to benefit from it (i.e. the default "get" functions should receive valid/cleaned responses. Load time would be the correct time to validate the values in darktablerc once and for all, imho.

But would be great to be able to access min, max, default separately from value via an api afterwards.

TurboGit: WIP here : #6412

PR #6404, opened 2020-09-29
[prefs] remove tiny memleak from generate_prefs

johnny-bit:
`g_strdup` needs to be freed after use, but IMO it's not even necessary here so changed to use const strings.


TurboGit (approved):
Looks ok, thanks!

PR #6406, opened 2020-09-29
[prefs] visually highlight changed prefs

### LINKED ISSUE #6176: no way to recognize changed preferences
An issue I have or may have, specially when upgrading to a new version, is that I have no way to know what preferences are not at the default value.

Many apps are used to format bold or italic (probably another way could be better) the preferences which have been changed.

I think this would be useful with shortcuts, too. They are organized in a tree-way, if a leaf isn't at the default, all the tree would be bold/italic/ecc. Also because of the _reset_ function: without a way to know what isn't already reset, that operation will produce fear in the user, fear to loose some shortcut he/she has set.

johnny-bit:
I haven't had the best idea here, but hey - at least it works ;)

I'm open to recommendations on making this better!

Fixes #6176 :)

### DIFF
--- a/src/control/conf.c
+++ b/src/control/conf.c
@@ -377,6 +377,19 @@ int dt_conf_key_exists(const char *key)
   return res;
 }
 
+gboolean dt_conf_is_default(const char *key)
+{
+  const char *def_val = dt_conf_get_default_var(key);
+  if(!def_val)
+    return TRUE; // well if default doesn't know about it, it's default
+
+  const char* cur_val = dt_conf_get_var(key);
+  if(cur_val && (!g_ascii_strcasecmp(def_val, "true") || !g_ascii_strcasecmp(def_val, "false")))
+    return !g_ascii_strcasecmp(def_val, cur_val); // booleans must be compared case insensitive :/
+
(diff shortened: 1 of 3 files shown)


rgo:
Hi, could you add an span or something with a class(instead of the b tag) ? And then add the style using the css. 
Thanks for your time!

TurboGit (changes requested):
Yep, better to use the CSS here. I would try setting a specific widget name for changed label and add a corresponding CSS rule.

src/control/conf.c
```diff
@@ -377,6 +377,19 @@ int dt_conf_key_exists(const char *key)
   return res;
 }
 
+gboolean dt_conf_is_default(const char *key)
```
TurboGit, line 380: This would be rewritten using dt_confgen_get() from my po/conf_gen branch.

johnny-bit:
Yeah, thought so too... Will check this after your PR lands.

TurboGit:
PR is merged now.

johnny-bit:
I think I'll rework this PR totally now :)

johnny-bit:
I'm closing this one in favour of  #6429 - hope new PR makes more sense :)

PR #6408, opened 2020-09-29
make height of remaining libs modules configurable/refactor

dterrahe:
Allow setting a maximum on the widget height (using ctrl-scroll) of all the remaining lib modules in lighttable and darkroom that didn't have this feature yet;
history
duplicates
colorpicker
snapshots
styles
recent collections
and refactored metadata and masks (which had coincidentally just a few days back received this feature from @johnny-bit)

The functionality here is slightly different from elsewhere in that there is no hard minimum size; if you want to, you can force the list to be one line high. However if you allow it to be higher than the hardcoded "minimum" size, it will only use the space required for the current number of entries. Also, the height, and step size, when using the mouse wheel to scroll on the list itself, not the scrollbar, will always be a multiple of the line/entry height.

There are also some other changes and fixes to some of the modules. The layout of duplicates has changed; I have moved the add duplicate/virgin buttons to the bottom to match better with the layout of the history and snapshots modules. Some theming had to be adjusted to make sure items/lines are of fixed height (especially multiline labels as in metadata). @nilvus please comment on the changes.

If the changes here are liked, I should be able to extend the support to also cover GtkTextView and refactor all the other modules as well.

There is a remaining issue with updating the widgets after a resize. This is most obvious with the metadata module if there are some images with multiline metadata tags. While mouse moving over them, the size of the module will always be one step behind. I've tried adding all combinations of gtk_widget_queue_resize/draw/allocate to _scroll_wrap_resize, but so far haven't found a solution. I'd be very happy for someone to point out to me what I'm overlooking...

### DIFF
--- a/src/gui/gtk.c
+++ b/src/gui/gtk.c
@@ -3093,7 +3093,7 @@ static gint _get_container_row_heigth(GtkWidget *w)
   return height;
 }
 
-static void _scroll_wrap_resize(GtkWidget *w, GtkAllocation *allocation, const char *config_str)
+static gboolean _scroll_wrap_resize(GtkWidget *w, void *cr, const char *config_str)
 {
   GtkWidget *sw = gtk_widget_get_parent(w);
   if(GTK_IS_VIEWPORT(sw)) sw = gtk_widget_get_parent(sw);
@@ -3131,6 +3131,8 @@ static void _scroll_wrap_resize(GtkWidget *w, GtkAllocation *allocation, const c
   gint value = gtk_adjustment_get_value(adj);
   value -= value % increment;
   gtk_adjustment_set_value(adj, value);
+
+  return FALSE;
 }
 
 static gboolean _scroll_wrap_scroll(GtkScrolledWindow *sw, GdkEventScroll *event, const char *config_str)
@@ -3168,7 +3170,7 @@ GtkWidget *dt_ui_scroll_wrap(GtkWidget *w, gint min_size, char *config_str)
   gtk_scrolled_window_set_policy(GTK_SCROLLED_WINDOW(sw), GTK_POLICY_NEVER, GTK_POLICY_AUTOMATIC);
   gtk_scrolled_window_set_min_content_height(GTK_SCROLLED_WINDOW(sw), - DT_PIXEL_APPLY_DPI(min_size));
   g_signal_connect(G_OBJECT(sw), "scroll-event", G_CALLBACK(_scroll_wrap_scroll), config_str);
-  g_signal_connect(G_OBJECT(w), "size-allocate", G_CALLBACK(_scroll_wrap_resize), config_str);
+  g_signal_connect(G_OBJECT(w), "draw", G_CALLBACK(_scroll_wrap_resize), config_str);
   gtk_container_add(GTK_CONTAINER(sw), w);
 
   return sw;
--- a/src/gui/gtk.c
+++ b/src/gui/gtk.c
@@ -3066,6 +3066,114 @@ GtkWidget *dt_ui_notebook_page(GtkNotebook *notebook, const char *text, const ch
   return page;
 }
 
+static gint _get_container_row_heigth(GtkWidget *w)
+{
+  gint height = DT_PIXEL_APPLY_DPI(10);
+
+  if(GTK_IS_TREE_VIEW(w))
+  {
+    gint cell_height = 0;
+    gtk_tree_view_column_cell_get_size(gtk_tree_view_get_column(GTK_TREE_VIEW(w), 0),
+                                       NULL, NULL, NULL, NULL, &cell_height);
+    GValue separation = { G_TYPE_INT };
+    gtk_widget_style_get_property(w, "vertical-separator", &separation);
+
+    if(cell_height > 0) height = cell_height + g_value_get_int(&separation);
+  }
+  else
+  {
+    GList *children = gtk_container_get_children(GTK_CONTAINER(w));
+    if(children)
+    {
+      height = gtk_widget_get_allocated_height(GTK_WIDGET(children->data));
+      g_list_free(children);
+    }
+  }
+
+  return height;
+}
+
+static void _scroll_wrap_resize(GtkWidget *w, GtkAllocation *allocation, const char *config_str)
+{
+  GtkWidget *sw = gtk_widget_get_parent(w);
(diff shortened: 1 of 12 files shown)


src/gui/gtk.c
```diff
@@ -3066,6 +3066,110 @@ GtkWidget *dt_ui_notebook_page(GtkNotebook *notebook, const char *text, const ch
...
+
+  gint increment = _get_container_row_heigth(w);
+
+  gint height = dt_conf_get_int(config_str);
+
+  gint max_height = DT_PIXEL_APPLY_DPI(1000);
+
+  height = (height < 1) ? 1 : (height > max_height) ? max_height : height;
+
+  dt_conf_set_int(config_str, height);
```
johnny-bit, line 3109: After my recent fun with dt conf you might replace this whole bit with `dt_conf_get_and_sanitize_int()`

johnny-bit (commented):
I haven't tested nor did anything other than to suggest to use my recently introduced stuff :)

src/gui/gtk.c
dterrahe: We are almost 100% in sync!

Nilvus:
@dterrahe: a good direction there and I love seeing ctrl-scroll extended.

About duplicates modules, 3 things: 
- first, I think it's better with that PR, especially having label text taking all size and so delete icon on top right.
- but I find strange now to have number centered alone that way. I would suggest to put it on top left, on same line as delete cross
- some margins missing: on the right of thumbnails and on top of new buttons

For last one, here my CSS proposal to fix it (to integrate around line 580 in darktable.css, between buttons and entry settings):

```
#left #lib-plugin-ui .duplicate-ui entry 
{
  margin-left: 0.5em;
}

#left .duplicate-ui .text-button
{
  margin-top: 1em;
}
```

Feel free of course to integrate it on your PR.

Nilvus:
Last thing: about 2nd issue, if you put number on top left, it should of course be vertically aligned to label left, so have same left margin as entry proposed on my CSS (if ok for you to integrate it).

dterrahe:
Thanks for your review @Nilvus 
I put the number on the same line as the delete button as requested, but I find it slightly more pleasing looking centered; it is the same layout as top level windows, so the close button on the right also feels very natural this way. Have a look at the screen shot below and tell me what you think.
image

I've also reduced the 1em top margin for the buttons you suggested to 4px; now it is more consistent with the history and snapshot buttons (which are 2px margin and 2px padding).

Lastly, I've pushed a change to make scrolling with "mouse wheel scrolls modules" enabled behave the same as with the other scrolled-windows modules; if there is no scroll bar, the whole panel is scrolled. In the previous iteration of this PR, the mouse scroll would always be captured and not move on.
Ideally, in cases where there is a scrollbar visible, I would like the scrolling to be captured within the module until the end of the viewport is reached, and then I would like the panel to continue scrolling. But I don't know how to achieve this effect as it isn't what the scrollbars do by default. So I'll leave this behavior as before and elsewhere.

src/gui/gtk.c
dterrahe: > After my recent fun with dt conf you might replace this whole bit with `dt_conf_get_and_sanitize_int()`

After the recent conf changes, what do you suggest I do here? Add 
`    <type min="1" max="1000">int</type>
`
to each of the entries in the xml and just do a simple dt_conf_get_int here?

Nilvus:
> Thanks for your review @Nilvus
> I put the number on the same line as the delete button as requested, but I find it slightly more pleasing looking centered; it is the same layout as top level windows, so the close button on the right also feels very natural this way. Have a look at the screen shot below and tell me what you think.

Great for me that way. Centered on same line is much better than alone.

> I've also reduced the 1em top margin for the buttons you suggested to 4px; now it is more consistent with the history and snapshot buttons (which are 2px margin and 2px padding).

Didn't check that. I found that less aerate but indeed more consistent with other modules. So good for me. Thanks for your work.

src/gui/gtk.c
johnny-bit: Later this weekend I'll propose a PR that will allow just that :)

johnny-bit: My PR is done now :) also I'd recommend more strict "min". `1` seems a little low for module window height :)

dterrahe: It gets rounded up to full row/element heights (depending on font etc). So 1 is intentional; there will always be one full row. I don't see a need to force it to be higher than that, if someone want a 1 row scroll window, why stop them?

TurboGit:
When I open the style module in lighttable (closed when starting) I get this:

Capture d’écran de 2020-10-05 08-26-24

This gets properly redrawn as soon as I fly over the module with the mouse. So probably a missing redraw signal.

TurboGit:
Same issue with history & mask module when entering darkrrom:

image

TurboGit (changes requested):
Many modules does not get properly displayed until some refresh signal is received.

dterrahe:
> Many modules does not get properly displayed until some refresh signal is received.

Thanks @TurboGit for looking at this.

I mentioned this problem in the submission comment to this PR. I have tried all kinds of gtk_widget_queue_resize/draw etc, on the widget, the scrolledwindow and parent, and even tried to force an immediate recalc/redraw with gtk_widget_size_allocate etc but nothing makes the window update itself. It is not a problem with the underlying box, grid or view; that gives the correct new size, so the size_request is putting in the required size. It is just the scrolledwindow not redrawing; the next redraw "fixes" it, but always a step behind. I'm even wondering if updates get temporarily disabled or postponed, because usually a size_request triggers an immediate repaint. If anybody has any other suggestions of what could fix this I would be very grateful. Who knows most about this? @AlicVB ?

dterrahe:
After having tried everything, I did one more experiment, which I hadn't attempted before because it seemed obviously incorrect; the resize routine for the scrolled window now connects to the "draw" signal for the wrapped widget, rather than "size-allocate". Miraculously, now it all seems to work as intended.

@TurboGit Please have another look and let me know if this works for you and if you have any reservations. The draw signal must get called many more times than the size-allocate signal, but to my honest surprise I don't see any malfunctions.

TurboGit:
> I mentioned this problem in the submission comment to this PR.

Right and I read it when the PR landed here and forgot :( Anyway good to see that you've found a solution to this problem. I'll have  a look.

TurboGit (approved):
Works for me. The "draw" signal is ok by me but I'm not expert on Gtk+ trick :) A nice addition to dt. Thanks.

PR #6412, opened 2020-09-30
po/conf gen

TurboGit:


johnny-bit:
IMO looks good!

dterrahe:
Indeed looks fine.

From reading the code, it looks like this will fix (i.e. writes back the default) an invalid value the first time it gets used. If a value _doesn't_ get used on a particular run, it doesn't get fixed, correct? The wrong value just gets written back to darktablerc when the program closes.

"invalid" in this case just means NaN. There's no checking against min/max yet. So that would have to be done in code, every time. Which would be (needlessly) expensive for values that get requested on every mouse scroll for example.

I still think the better approach would be to validate (check+correct) all values at start-up and then _know_ that they are all valid when they are requested. The only assumption (which I think is fine) is that we don't ourselves write back a NaN or out of bounds value during the session.

TurboGit:
@dterrahe : 100% agreed, this is still WIP, next commit will validate when reading conf.

johnny-bit:
@dterrahe - I think invalid value can be kept in darktablerc for all I care, the only tnigh when that value needs to be valid is when reading (and possibli writting :P )

Validating on init is imo "meh" can be, can be not.

I thought about different method - instead of 3 lookups (def, min, max) have a struct with those, so there's only 1 lookup needed. Then both reading and writting could get nicelly clamped for any pre-defined values. And those that aren't predefined could stil use get_and_sanintize or set_sanitized methods to have bounds checking in code.

TurboGit:
@johnny-bit : Yes a struct would be better. Will do that. Again this is WIP and not even really tested :)

TurboGit:
Still WIP but this version is getting better !

TurboGit:
No more WIP, I think this is good to go. If ok with you let me know and @johnny-bit will be able to rework #6403 on top of this.

src/control/conf.c
```diff
@@ -259,10 +264,70 @@ gchar *dt_conf_get_string(const char *name)
...
+  {
+    case DT_INT:
+    {
+      float v = dt_calculator_solve(1, value);
+
+      const int min = item->min ? (int)dt_calculator_solve(1, item->min) : INT_MIN;
+      const int max = item->max ? (int)dt_calculator_solve(1, item->max) : INT_MAX;
+      // if garbadge, use default
+      const int val = isnan(v) ? dt_confgen_get_int(name, DT_DEFAULT) : (int)v;
+      l_sprintf(result, sizeof(result), "%d", CLAMP(val, min, max));
```
johnny-bit, line 286: I can't compile dt with this change, here's the error:
```
/home/johnny/darktable_repos/darktable/src/control/conf.c: In function ‘_sanitize_confgen’:
/home/johnny/darktable_repos/darktable/src/control/conf.c:286:7: error: implicit declaration of function ‘l_sprintf’; did you mean ‘g_sprintf’? [-Werror=implicit-function-declaration]
  286 |       l_sprintf(result, sizeof(result), "%d", CLAMP(val, min, max));
      |       ^~~~~~~~~
      |       g_sprintf
```

TurboGit: Should be fixed, a copy&paste error, this is intended to be used only for lua code and I suppose you don't have lua activated on your side.

johnny-bit:
Yeah, it's fine now and compiles.

I do have some notes though :) (you can not care about those for this PR and if they'll bother me, I'll try to fix them anyway in different pr, but just listing them for sake of remembering)

1. generated `_insert_type` for enum types is quite a bit loooooong, i'd either skip or just say type = string for enums
2. `dt_conf_init` reading default from darktable_datadir/darktablerc is imo "busy spin" on conf creation because all default values are in code anyway ;) so the conditional to load defaults could be reworked and the loop `for(i=0; i<2; i++)` could be just single run over user's config and in case there is no user's config- create it :)

i think that's all :)

TurboGit:
Ok point 1 is done. Indeed there is nothing we can do for enums at this stage.

TurboGit:
For 2, I'll look, but indeed this loop makes not sense now to me.

TurboGit:
Point 2 handled now, the reading is now lot easier to understand.

johnny-bit:
Yeah, seems good and understandable, however (this is my nitpicking, but hey ;) ) we're still reading the file and relying on presence of datadir/darktablerc, when it's not really needed since every read will do passthrough to defaults if table is completely empty (eg when user's darktablerc is present but it's empty file). IMO it would be totally fine to just create empty darktablerc file on config init or even just skip reading file totally in case it's not present.

Anyway - now it's completelly fine and my nitpick above is just a nitpick ;)

TurboGit:
>  IMO it would be totally fine to just create empty darktablerc file on config init

Just to be sure about what you are proposing:
- only read user darktablerc
- if it is not there no problem we keep the table empty
- when a specific key is read, if present in the table ok otherwise we get the default
- when leaving dt we record in user darktablerc only the config in table, so only the values that are non default

Is that what you're proposing ? I suppose this would work indeed.

dterrahe:
Not arguing against or for this, but I think at the moment one of the purposes of the rc (both darktable and keyboard) is to list all the options that are available for editing. If they are not all in there, the user might not have another easy way of figuring what can be configured (and how).

johnny-bit:
> Just to be sure about what you are proposing:

mostly correct

I'd just suggest ininting ->table with defaults (->x_default) and not read any files (and since those are in-code defaults, no sanity check is needed for them) and then writting on close darktablerc to user dir

That should also cover @dterrahe's comment

TurboGit:
> I'd just suggest ininting ->table with defaults (->x_default) and not read any files (and since those are in-code defaults, no sanity check is needed for them) and then writting on close darktablerc to user dir

But in x_default there is missing conf so I don't see the point especially since we do check for default (dt_conf_get_var)  if a variable is not found. So basically my proposal (doing nothing) is probably equivalent or am I missing something?

johnny-bit:
in datadir/darktablerc we have all the defaults as generated during compile time, so same as x_default. that's why it's not necessary to read the file.

problem is - @dterrahe rightfully points that ->table only gets updated with default value if anything is read from config, that would mean that configs not read/modified during single run of darktable would be missing form user's darktablerc.

so imo if user's darktablerc is missing, just copy all content from ->x_default to ->table so that user's darktablerc is filled with values.

(that would also prooobably mean that then having datadir/darktablerc would be pointless ;) )

TurboGit:
> in datadir/darktablerc we have all the defaults as generated during compile time, so same as x_default. that's why it's not necessary to read the file.

In fact not since I have removed enum. So we may want to handle properly enum and store the default value first.

And I'll see indeed if we can avoid reafing datadir/darktablerc, I see your point now I think.

TurboGit:
Ok, so what about this last version ?

johnny-bit:
IMO perfect!

TurboGit:
Ok, merged then!

PR #6422, opened 2020-10-01
updating italian translations

### LINKED ISSUE #6319: italian translation issue
in the preferences - general, the phrase

"modify selected theme with CSS tweaks below"

is incorrectly translated as 

"Modifica il tema selezionato con le chiavi CSS qui sotto"

Tweaks means adjustments, and cannot be translated into italian as "chiavi", but with "modifiche", "ritocchi" or something similar:

"**modifica il tema selezionato con i ritocchi del CSS qui sotto**"

Please note the upper to lower case change at the beginning, too.

dennybiasiolli:
Close #6319


TurboGit (approved):
Thanks !

PR #6424, opened 2020-10-01
fix error when hovering over blending slider while in Entry

dterrahe:
Entry widgets (for example the module search or rename edit fields) disable accelerators. The blending code would do the same _again_ (and throw the error below) when the mouse enters it and grab the focus afterwards. Reversing those two actions causes whoever had focus before to reenable accelerators so they can be happily disabled again here.

(darktable:295164): Gtk-CRITICAL **: 13:45:28.580: _gtk_accel_group_detach: assertion 'g_slist_find (accel_group->priv->acceleratables, object) != NULL' failed


TurboGit:
Looks good to me, thanks!

PR #6425, opened 2020-10-01
Simplified approach for `screen_ppd_overwrite` and `ui/cairo_filter`

jenshannoschwalm:
The PPD (pixels-per-dot) value is read by default from OS setting resulting in values
of either 1.0f or 2.0f for normal / high-resolution monitors. Or you could specify it via
the general->preferences.

The tooltip there suggested to increase the value for better visual output regarding thumbs & previews,
or decrease it for better performance. In fact, the first suggestion is not true any more since
the thumbnail size is calculated correctly. The second suggestion (decreasing) indeed leads to
increased performance but also lead to blurring in gtk/bauhaus elements. (To test this on a normal
monitor, set it to 0.5f, go to darkroom and look at the iops)

Another option was the hidden `ui/cairo_filter` option. It was introduced a while ago for more 'crisp'
display of thumbs & previews when in hires mode, this is not necessary any more.

**But** both above ways would be fine for slow computers to have faster thumb&preview rendering.

What has been done in this pr?

1) Both config options have been removed and were replaced by a drop down menu selecting between
   system(default), standard monitor(ppd==1) and hires monitor(ppd==2). For both specified monitor
   types we can select between normal vs. slow computer mode.
   "default"
   "standard monitor"
   "standard monitor, (slow)"
   "high resolution monitor"
   "high resolution monitor, (slow)"

2) As we want the thumbs to be displayed with different ppd than the GUI ppd we add `ppd_thb` to
   `dt_gui_gtk_t`, this is used in all thumbs drawing code recently introduced when fixing the hires
   impaired output.

3) When selecting one of non-slow modes, ppd_thb is the same as ppd and rendering is cairo_good.
4) If we use one of the slow modes we select cairo_fast and a smaller ppd_thb.
   This results in an overall > 4-fold better performance for the slow modes.

### DIFF
--- a/src/gui/gtk.c
+++ b/src/gui/gtk.c
@@ -1550,51 +1550,36 @@ void dt_gui_gtk_run(dt_gui_gtk_t *gui)
   dt_cleanup();
 }
 
+// refactored function to read current ppd, because gtk for osx has been unreliable
+// we use the specific function here. Anyway, if nothing meaningful is found we default back to 1.0
+double dt_get_system_gui_ppd(GtkWidget *widget)
+{
+  double res = 0.0f;
+#ifndef GDK_WINDOWING_QUARTZ
+  res = gtk_widget_get_scale_factor(widget);
+#else
+  res = dt_osx_get_ppd();
+#endif
+  if((res < 1.0f) || (res > 4.0f))
+  {
+    dt_print(DT_DEBUG_CONTROL, "[dt_get_system_gui_ppd] can't detect system ppd\n");
+    return 1.0f;
+  }
+  dt_print(DT_DEBUG_CONTROL, "[dt_get_system_gui_ppd] system ppd is %f\n", res);
+  return res;
+}
+
 void dt_configure_ppd_dpi(dt_gui_gtk_t *gui)
 {
   GtkWidget *widget = gui->ui->main_window;
 
-  gui->ppd = gui->ppd_thb = 1.0f;
+  gui->ppd = gui->ppd_thb = dt_get_system_gui_ppd(widget);
   gui->filter_image = CAIRO_FILTER_GOOD;
-
-  // check if in HiDPI mode
-#if (CAIRO_VERSION >= CAIRO_VERSION_ENCODE(1, 13, 1))
-  int monitor = dt_conf_get_int("ui/monitor");
-  if(monitor >= 0)
+  if(dt_conf_get_bool("ui/performance"))
   {
-    if(monitor == 2)
-    {
-      gui->ppd_thb = 0.7f;
+      gui->ppd_thb = 0.7f * gui->ppd;
       gui->filter_image = CAIRO_FILTER_FAST;
-    }
-    else if(monitor == 3)
-    {
-      gui->ppd_thb = gui->ppd = 2.0f;
-    }
-    else if(monitor == 4)
-    {
-      gui->ppd = 2.0f;
-      gui->ppd_thb = 1.4f;
-      gui->filter_image = CAIRO_FILTER_FAST;
-    }
   }
-  else
-  {
- #ifndef GDK_WINDOWING_QUARTZ
-    gui->ppd = gui->ppd_thb = gtk_widget_get_scale_factor(widget);
- #else
-    gui->ppd = gui->ppd_thb = dt_osx_get_ppd();
- #endif
-
-    if(gui->ppd < 0.0f)
-    {
-      gui->ppd = gui->ppd_thb = 1.0f;
-      dt_print(DT_DEBUG_CONTROL, "[HiDPI] can't detect screen settings, switching off\n");
-    }
-    else
-      dt_print(DT_DEBUG_CONTROL, "[HiDPI] setting ppd to %f\n", gui->ppd);
-  }
-#endif
   // get the screen resolution
   float screen_dpi_overwrite = dt_conf_get_float("screen_dpi_overwrite");
   if(screen_dpi_overwrite > 0.0)
--- a/src/gui/gtk.c
+++ b/src/gui/gtk.c
@@ -1554,31 +1554,45 @@ void dt_configure_ppd_dpi(dt_gui_gtk_t *gui)
 {
   GtkWidget *widget = gui->ui->main_window;
 
+  gui->ppd = gui->ppd_thb = 1.0f;
+  gui->filter_image = CAIRO_FILTER_GOOD;
+
   // check if in HiDPI mode
 #if (CAIRO_VERSION >= CAIRO_VERSION_ENCODE(1, 13, 1))
-  float screen_ppd_overwrite = dt_conf_get_float("screen_ppd_overwrite");
-  if(screen_ppd_overwrite > 0.0)
+  int monitor = dt_conf_get_int("ui/monitor");
+  if(monitor >= 0)
   {
-    gui->ppd = screen_ppd_overwrite;
-    dt_print(DT_DEBUG_CONTROL, "[HiDPI] setting ppd to %f as specified in the configuration file\n", screen_ppd_overwrite);
+    if(monitor == 2)
+    {
+      gui->ppd_thb = 0.5f;
+      gui->filter_image = CAIRO_FILTER_FAST;
+    }
+    else if(monitor == 3)
+    {
+      gui->ppd_thb = gui->ppd = 2.0f;
+    }
+    else if(monitor == 4)
+    {
+      gui->ppd = 2.0f;
+      gui->filter_image = CAIRO_FILTER_FAST;
+    }
   }
   else
   {
-#ifndef GDK_WINDOWING_QUARTZ
-    gui->ppd = gtk_widget_get_scale_factor(widget);
-#else
-    gui->ppd = dt_osx_get_ppd();
-#endif
-    if(gui->ppd < 0.0)
+ #ifndef GDK_WINDOWING_QUARTZ
+    gui->ppd = gui->ppd_thb = gtk_widget_get_scale_factor(widget);
+ #else
+    gui->ppd = gui->ppd_thb = dt_osx_get_ppd();
+ #endif
+
+    if(gui->ppd < 0.0f)
     {
-      gui->ppd = 1.0;
+      gui->ppd = gui->ppd_thb = 1.0f;
       dt_print(DT_DEBUG_CONTROL, "[HiDPI] can't detect screen settings, switching off\n");
     }
     else
       dt_print(DT_DEBUG_CONTROL, "[HiDPI] setting ppd to %f\n", gui->ppd);
   }
-#else
-  gui->ppd = 1.0;
 #endif
   // get the screen resolution
   float screen_dpi_overwrite = dt_conf_get_float("screen_dpi_overwrite");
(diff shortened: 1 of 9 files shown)


TurboGit:
I very much like the description. Looks like a nice usability improvement and code clean-up at the same time. Will test this soon.

src/gui/preferences.c
```diff
@@ -392,6 +400,28 @@ static void init_tab_general(GtkWidget *dialog, GtkWidget *stack, dt_gui_themetw
...
+  widget = gtk_combo_box_text_new();
+  labelev = gtk_event_box_new();
+  gtk_widget_add_events(labelev, GDK_BUTTON_PRESS_MASK);
+  gtk_container_add(GTK_CONTAINER(labelev), label);
+  gtk_grid_attach(GTK_GRID(grid), labelev, 0, line++, 1, 1);
+  gtk_grid_attach_next_to(GTK_GRID(grid), widget, labelev, GTK_POS_RIGHT, 1, 1);
+
+  gtk_combo_box_text_append_text(GTK_COMBO_BOX_TEXT(widget), "default");
+  gtk_combo_box_text_append_text(GTK_COMBO_BOX_TEXT(widget), "standard monitor");
+  gtk_combo_box_text_append_text(GTK_COMBO_BOX_TEXT(widget), "standard monitor, (slow)");
```
rgo, line 414: I'd say "slow computers" or "fast(er) ".
"slow" literal is used in other places to say that it's slower with that option enabled (e.g. demosaic modes).

rgo:
Indeed great usability addition 👍🏻

jenshannoschwalm:
There is something i'd like to discuss before thinking of merge.

In src/gui/gtk.c some sections seem to be very old, we test for cairo >v.1.13 which is >6ys old. Also the gtk functions testing for ppd like `gtk_widget_get_scale_factor` are in use for many years now.

The menu would be much simpler - just normal vs performance mode - if we could take that for granted.

johnny-bit:
I think the gtk_widget_get_scale_factor should be used, since it's relatively old by now. And I think we could bump min cairo version if that's of benefit to us, since many distros are on 1.16

So - oldest LTS ubuntu (xenial) supports cairo 1.14
bionic- 1.15.10
focal - 1.16
(same applies to debian)

redhat-based current distros (starting from fedora 30) and opensuse support at least 1.15

So - I'd be for bumping lib requirements, since all current distros should support at least cairo 1.14 (or ever 1.15)

TurboGit:
Yes, Debian/stable is using 1.16 and this distro is not reputed to be bleeding edge :)

jenshannoschwalm:
OK. Agreed. @parafin, any problems expected for osx?

johnny-bit:
I think there shouldn't be? https://ports.macports.org/port/cairo/ <- says 1.16... But I'd really wait for @parafin's greenlight on it.

TurboGit:
We also need to check the Windows build. I have no idea what is used there.

jenshannoschwalm:
```
#ifndef GDK_WINDOWING_QUARTZ
    gui->ppd = gtk_widget_get_scale_factor(widget);
#else
    gui->ppd = dt_osx_get_ppd();
#endif`
```

Windows is ok, it's just the osx special call.

src/gui/preferences.c
parafin: +1

parafin:
What is the question? Bumping the lib dependencies? Just check macports to see what they have.

jenshannoschwalm:
> What is the question?

Do we still need to use `dt_osx_get_ppd()` or can we also use the generic `gtk_widget_get_scale_factor(widget)`?

parafin:
I will have to test it, but why change this?

jenshannoschwalm:
Just simpler code?

parafin:
If you look at git blame you will see that simpler code was already attempted once, but ended up as it's usual with GTK: https://github.com/darktable-org/darktable/commit/f1f1dd873df786b4a67cff58a5138501718f21af

I would rather we not touch this, I don't want to spend time on this any more than absolutely needed.

jenshannoschwalm:
Thanks! I have to learn using git blame, sorry for taking your time.

src/gui/preferences.c
jenshannoschwalm: ok. Fixed by introducing just a single button

jenshannoschwalm:
Ok, discussing this has lead to further simplified preferences.

1) The system ppd reading has been refactored in `double dt_get_system_gui_ppd(GtkWidget *widget)`. It returns the currently used ppd (either by gtk code or the osx specific), if nothing sensible is found it defaults back to 1.0
2) In preferences there is now just a single button switching performance mode on/off, in performance mode we use the smaller thumbs and cairo_fast
3) As this is now a single bool not depending on any gtk stuff running we can also integrate this `ui/performance` setting into `dt_configure_performance()`
4) As we always use the system ppd we don't have to restart dt when switching performance mode on/off

jenshannoschwalm:
Ok, seems good for me now.

src/gui/gtk.c
```diff
@@ -1550,36 +1550,36 @@ void dt_gui_gtk_run(dt_gui_gtk_t *gui)
   dt_cleanup();
 }
 
+// refactored function to read current ppd, because gtk for osx has been unreliable
+// we use the specific function here. Anyway, if nothing meaningful is found we default back to 1.0
+double dt_get_system_gui_ppd(GtkWidget *widget)
+{
+  double res = 0.0f;
+#ifndef GDK_WINDOWING_QUARTZ
```
TurboGit, line 1558: We have a huge number of #ifdef for this and 4 #ifndef. I'm proposing to use #ifdef here, easier to read to me.

jenshannoschwalm: right

jenshannoschwalm:
done

src/gui/gtk.c
```diff
@@ -1550,36 +1550,36 @@ void dt_gui_gtk_run(dt_gui_gtk_t *gui)
...
-#if (CAIRO_VERSION >= CAIRO_VERSION_ENCODE(1, 13, 1))
-  float screen_ppd_overwrite = dt_conf_get_float("screen_ppd_overwrite");
-  if(screen_ppd_overwrite > 0.0)
+  gui->ppd = gui->ppd_thb = dt_get_system_gui_ppd(widget);
+  gui->filter_image = CAIRO_FILTER_GOOD;
+  if(dt_conf_get_bool("ui/performance"))
   {
-    gui->ppd = screen_ppd_overwrite;
-    dt_print(DT_DEBUG_CONTROL, "[HiDPI] setting ppd to %f as specified in the configuration file\n", screen_ppd_overwrite);
+      gui->ppd_thb = 0.7f * gui->ppd;
```
TurboGit, line 1580: If I'm reading this correctly it is equivalent to:

`gui->ppd_thb *= 0.7ff;`

Now the question is what is 0.7f ? Maybe using a named-number would make this clearer.

src/views/darkroom.c
```diff
@@ -4392,34 +4392,10 @@ static void _second_window_configure_ppd_dpi(dt_develop_t *dev)
...
-    }
-    else
-      dt_print(DT_DEBUG_CONTROL, "[HiDPI] setting ppd to %f\n", dev->second_window.ppd);
-  }
-#else
-  dev->second_window.ppd = 1.0;
-#endif
+  dev->second_window.ppd = dev->second_window.ppd_thb = dt_get_system_gui_ppd(widget);
+  if(dt_conf_get_bool("ui/performance"))
+    dev->second_window.ppd_thb = 0.7f * dev->second_window.ppd;
```
TurboGit, line 4397: likewise for 0.7f and equivalent to:

`dev->second_window.ppd_thb *= 0.7f;`

src/gui/gtk.c
jenshannoschwalm: Right. Any naming suggestion?

I wrote it like that to make the dependency on ppd more obvious...

src/gui/gtk.h
```diff
@@ -26,6 +26,8 @@
 
 #define DT_GUI_IOP_MODULE_CONTROL_SPACING 0
 
+#define DT_GUI_THUMBSIZE_REDUCE 0.7f
```
TurboGit, line 29: Ok, better...Can you just tell me why 0.7f, how this was found ? Testing ? Just curious...

jenshannoschwalm: You can of course use any number less than 1 :-)
0.7 leads to thumbs with ~ half number of pixels with a perf gain of 2.
0.5 was too bad visually ...

TurboGit: Thanks ! 'll test...

TurboGit (approved):
Looks good to me! Thanks.

aurelienpierre:
That makes no sense to me.

The purpose of the PPD setting is to ensure previews and thumbnails are computed 1:1 pixel, in which case we don't need fancy cairo filter to interpolate.

But if you downscale previews at 70% of final display size, for the sake of performance, using `CAIRO_FILTER_FAST` (aka nearest neighbour, aka no interpolation), you will end up with aliasing artifacts. So you need a slower cairo filters for anti-aliasing, and will still get a blurry output.

I really wouldn't mess with pixmaps sizes. Just let them 1:1. All the more if the sizes discrepancies have been fixed and we can now get a true 1:1 scaling.

jenshannoschwalm:
I can't look at the code, just using phone. But there is no downscaling to 70%. And don't we have fixed thumbnails and preview sizes?

jenshannoschwalm:
The downscaling 0.7 just selects a smaller thumb group.

parafin:
I don't think dt uses thumbnail sizes to match exactly the dimensions on the screen - there are fixed set of sizes and the closest one is selected. So 1:1 scaling for thumbs might happen only due to coincidence.

LebedevRI:
While i don't have an opinion on the details of the change, i'm going to revert this for now:
with default config, load some image, zoom in 1600% percent, and see:
image
Compare that with:
image

See the problem? Why's there vignetting? Why it's smoothed?

parafin:
Hmm... Why darkroom was affected at all? I thought this change should have only touched lighttable and filmstrip.

LebedevRI:
```
commit 1716749e1d492975741d7c1fac43240bc2da1194 (HEAD -> master, upstream/master, origin/master, origin/HEAD)
Author: Roman Lebedev <lebedev.ri@gmail.com>
Date:   Tue Oct 27 22:52:33 2020 +0300

    Revert "Merge pull request #6425 from jenshannoschwalm/pref_ppd_new"
    
    This reverts commit 5aacde6b83b76436d8a1b0b6681bb36585eaefc0, reversing
    changes made to 79a0d3bd3dd07b4274e1e586d915635cba245bac.
    
    See https://github.com/darktable-org/darktable/pull/6425#issuecomment-717493303

commit 6085d5b551d33c9919d532f5f8fb5da4f6eeddfb
Author: Roman Lebedev <lebedev.ri@gmail.com>
Date:   Tue Oct 27 22:57:39 2020 +0300

    Revert "print: fix display of the image on center area in performance mode."
    
    This reverts commit e23f5401912c4d11e512aaab75d28330332d1400.

```

TurboGit:
I'm not happy with that. Reverting is not the solution, the solution is fixing and we are at work to have this ready for the code freeze. Fact is that now it is reverted we won't have field testing at all, which is very bad late in the game. At least before reverting we must discuss with the developer of the code. I did, and @jenshannoschwalm was aware and will be working on a fix.

TurboGit:
@LebedevRI : I have just pushed a fix for your issue. Please let us know if there is still issues with the rendering.

We have one pending issue, in performance mode the zoomed image in the culling mode cannot be fully panned.

LebedevRI:
Reverting is *always* The solution, because then the broken stuff is gone, and there is no pressure to fix it,
only implied pressure to reland after fixing. In fact, trigger-happy reverting early is the only correct solution.
The longer broken code stays in, the harder it is to revert it.

jenshannoschwalm:
I code is crap reverting is the solution, right. 

But you ignore the problem of blurred thumbs resulting from incorrect use of the ppd and that was a tricky bug that had to be fixed in some way.

LebedevRI:
> I code is crap reverting is the solution, right.

It's not that someone particular writes crap code. *Everyone* writes crap code.
It just takes a few rewrites for it to be reasonable.



> But you ignore the problem of blurred thumbs resulting from incorrect use of the ppd and that was a tricky bug that had to be fixed in some way.

Without having much background, i guess you are talking about the fact that this PR was actually fixing some problem?
That's good, yes.

TurboGit:
> Reverting is _always_ The solution

When close to the release yes. But again this PR is complex and touching many places and we have found regressions, fixing them one by one. This was one and I fixed in minutes, not that bad. We have still one known issue and if we can't fix it before the release we may revert indeed. But again for such complex change we need field testing, we don't have the resource for testing anything. 

@LebedevRI  : BTW, the fact that you found about this is issue is an example of why we don't want to revert early. If I had reverted because of the know issue still to be fixed you'll not have found it. And not many people are using 1600% zoom level so this issue could still be in 3.4.

And last, reverting should always be discussed with the original developer in my point of view.

PR #6426, opened 2020-10-02
[conf] don't truncate strings

johnny-bit:
A problem was detected by @AxelG-DE and @jade-nl - conf values, especially strings were truncated to 100 chars max

This PR prevents that and also preffers to use dynamic printf (and possibly fixes possible problem with possible bool values having problems since in xml they are repreented lower case, but in conf code they are written upper case)

### DIFF
--- a/src/control/conf.c
+++ b/src/control/conf.c
@@ -270,8 +270,7 @@ static char *_sanitize_confgen(const char *name, const char *value)
 
   if(!item) return g_strdup(value);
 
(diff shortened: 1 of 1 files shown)


TurboGit:
Obviously good and fixing my mess :) Thanks.

jade-nl:
Providing quick solutions, as always!

Thanks!

AxelG-DE:
Faster than I could open an issue even :-)

THANKS!

PR #6428, opened 2020-10-02
po/conf validate enum

TurboGit:


TurboGit:
Another data validation in case of garbage in darktablerc.

PR #6429, opened 2020-10-02
[prefs] highlight changed preferences

### LINKED ISSUE #6176: no way to recognize changed preferences

johnny-bit:
Fixes #6176 :)

This is completely reworked #6406 - I created separate PR because it would be totally different than original.

I had idea about making it more dynamic, but it was IMO unnecessary so this ver is pretty basic.

### DIFF
--- a/data/themes/darktable.css
+++ b/data/themes/darktable.css
@@ -1142,6 +1142,12 @@ dialog scrollbar
   margin: 1px 0;
 }
 
+#preferences_notebook label.preference_non_default
+{
+  font-weight: bolder;
+  font-style: italic;
+}
+
 #preferences_notebook combobox,
 #preferences_notebook spinbutton
 {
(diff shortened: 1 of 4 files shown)


TurboGit:
Looks promising, I'll test that early tomorrow now :) Thanks!

TurboGit:
> I had idea about making it more dynamic, but it was IMO unnecessary so this ver is pretty basic.

Would be nice to have it dynamic. But this is already really nice! Merging.

PR #6431, opened 2020-10-03
Fix crash caused by lua widget garbage collector

### LINKED ISSUE #6417: a8f47e409 continuously hanging
a8f47e409 continuously hanging

running under gdb, outputs:
  http://wahoo.no-ip.org/~paka/dt.a8f47e409.gdb.txt
  http://wahoo.no-ip.org/~paka/dt.a8f47e409.1.gdb.txt
  http://wahoo.no-ip.org/~paka/dt.a8f47e409.2.gdb.txt
  http://wahoo.no-ip.org/~paka/dt.a8f47e409.3.gdb.txt

openSUSE Tumbleweed 20200920
NVIDIA GF106 [GeForce GTS 450], 390.138
darktable-3.3.0~git877.f8b51737c
OpenCL loaded but not available
i7 12-core 36GB

wpferguson:
Removed code that tried to destroy a widget since widget_gc should be called after the widget has been destroyed.  Added an error message to the console when an attempt to garbage collect a widget that hasn't been destroyed.

Fixes #6417

### DIFF
--- a/src/lua/widget/widget.c
+++ b/src/lua/widget/widget.c
@@ -1,7 +1,6 @@
 /*
    This file is part of darktable,
-   copyright (c) 2015 Jeremy Rosen
-   copyright (c) 2015 tobias ellinghaus
+   Copyright (C) 2015-2020 darktable developers.
 
    darktable is free software: you can redistribute it and/or modify
    it under the terms of the GNU General Public License as published by
@@ -44,7 +43,7 @@ dt_lua_widget_type_t widget_type = {
 
 static void cleanup_widget_sub(lua_State *L,dt_lua_widget_type_t*widget_type,lua_widget widget);
 static void cleanup_widget_sub(lua_State *L,dt_lua_widget_type_t*widget_type,lua_widget widget) {
-  if(widget_type->parent) 
+  if(widget_type->parent)
     cleanup_widget_sub(L,widget_type->parent,widget);
   if(widget_type->gui_cleanup) {
     widget_type->gui_cleanup(L,widget);
@@ -53,9 +52,9 @@ static void cleanup_widget_sub(lua_State *L,dt_lua_widget_type_t*widget_type,lua
 
 static void init_widget_sub(lua_State *L,dt_lua_widget_type_t*widget_type);
 static void init_widget_sub(lua_State *L,dt_lua_widget_type_t*widget_type) {
-  if(widget_type->parent) 
+  if(widget_type->parent)
     init_widget_sub(L,widget_type->parent);
-  if(widget_type->gui_init) 
+  if(widget_type->gui_init)
     widget_type->gui_init(L);
 }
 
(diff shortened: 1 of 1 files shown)


TurboGit:
I would certainly remove all the commented out code. This is kept in Git anyway.

TurboGit:
I see reading the issue #6417 that this is only a workaround and not the proper fix. I'm not sure what to do there, my proposal would be to merge this just before 3.4 sources are cut if we don't have a better fix. IIUC this will result in memory leak in the lua code which is far better than crashing/hanging !

wpferguson:
My original thought was the change was a workaround because I wasn't sure why the garbage collector was trying to garbage collect the widget when it shouldn't have been.  Then I discovered that widgets that were in a table didn't get garbage collected by mistake.  After that discovery my thought is that the garbage collector (erroneously) believes the widget is no longer referenced and is just doing what it's supposed to and trying to recover the memory.

Now that I have an idea of why the garbage collector gets triggered my thought is that the piece of code causing the crash should never have been reached, so I replaced it with an error message.  If I changed it to make it work as originally intended, I would have sent the destroy signal to the widget, then returned.  The widget would have been destroyed, then the garbage collector called again and it would have just returned because the widget no longer existed.  The side effect of this is that when you went to the exporter and selected enfuseAdvanced the widgets would have been missing, if the gui handled the destruction of the widgets gracefully.

So, there is no memory leak because the widget isn't supposed to be destroyed.  

I've included the debugging version of src/lua/widget/widget.c that I used.  Replace src/lua/widget/widget.c with this, compile and start darktable in a terminal.  Enable enfuseAdvanced and you'll see all the widgets created.  Close darktable and you'll see all the widgets destroyed, then garbage collected.  If you want to see the crash, enable enfuseAdvanced, image_stack, and geoToolbox, then scroll back and forth through the lighttable until it happens.

[widget.c.zip](https://github.com/darktable-org/darktable/files/5321943/widget.c.zip)

There are one or more issues with widget garbage collection.  I, and others, have managed to trigger the "trying to destroy a parented widget" branch and now we have "trying to garbage collect a widget that hasn't been destroyed".  There's a way to code around both of these using lua (turn off garbage collection while creating large numbers of widgets for the first, and wrapping the widgets in a table for the second). The issue might actually be lua garbage collection in general, and we're just seeing it with the widgets.  Or it could be that garbage collection works differently than we understand.  Either way it needs investigating.  I'll open an issue for this.

In summary, this change will fix the crashes with no side effects that I can forsee and lua garbage collection needs investigating.

TurboGit:
The copyright notice change is not expected.

wpferguson:
@TurboGit thanks for removing the trailing spaces from src/lua/widget/widget.c.  It forced me to look at the log and I noticed commits 19c564582 and b5effdf58 which were supposed to fix a memory leak, but instead created a double free.  Reverted the changes and tested.  No crashes any more.

Note to self:  Check git log for recent changes :-).

wpferguson:
> The copyright notice change is not expected.

Copied the file from 3.0.0. After I pushed I saw the copyright change, so got the current one and fixed it.

TurboGit:
Thanks ! Glad by trailing white spaces has helped :) You have reintroduced some BTW, but I have squashed all together and removed again the white spaces. Please double check that the change is properly integrated.

wpferguson:
@TurboGit perfect, thanks.

PR #6433, opened 2020-10-03
preferences: make the changed preferences with a little tick character.

TurboGit:
After using the text in italic/bold I find it a bit disturbing. The bold looks a bit like the section labels.

I find this proposal nice and less disturbing.

### DIFF
--- a/data/themes/darktable.css
+++ b/data/themes/darktable.css
@@ -1149,12 +1149,6 @@ dialog scrollbar
   margin: 1px 0;
 }
 
-#preferences_notebook label.preference_non_default
-{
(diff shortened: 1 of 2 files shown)


johnny-bit:
I think you can do that with css... Also css would allow dynamic stuff I'm working on.

TurboGit:
Perfect, closing this then.

johnny-bit:
Question - how best to indicate changed preference? @Nilvus ?

Nilvus:
No idea actually but yes your update was in good direction but too disturbing. I will think about that ASAP.

PR #6434, opened 2020-10-03
cleanup minimal cairo version

jenshannoschwalm:
We had some compile time checks for very old cairo versions.
These are simply superfluous as the gtk check for 3.22 implies cairo to be at last 1.4.


TurboGit (approved):
Thanks !

PR #6437, opened 2020-10-03
Improve display of combobox windows

amousset:
* align window with the button
* avoid moving the current choice when dropdown windows opens
* align button to the left in preferences

combo2
combo3
combo1

### DIFF
--- a/data/themes/darktable.css
+++ b/data/themes/darktable.css
@@ -846,11 +846,11 @@ dialog combobox window
   outline-style: none;
   min-height: 0.4em;
   min-width: 0.4em;
-  padding: 2px 0;
+  padding: 0.2em 0;
 }
 
 menu menuitem {
-  padding: 0 2px;
+  padding: 0 0.2em;
 }
 
 tooltip
@@ -1089,7 +1089,7 @@ dialog scrollbar
   margin: 0px; /* reset default dialog margin to all items. And needed to have an hover and selected effect on all width of categories in sidebar */
 }
 
-#preferences_notebook dialog > decoration  /* allow to resize preferences dialog window */
+#preferences_notebook decoration  /* allow to resize preferences dialog window */
 {
   margin: 1px;
 }
@@ -1171,7 +1171,8 @@ dialog scrollbar
 
 #preferences_notebook button
 {
-  padding: 0 2px;
+  margin-left: 1px;
+  padding: 0 1px;
 }
 
 /* Set buttons and search field on presets and shortcuts control buttons */
--- a/data/themes/darktable.css
+++ b/data/themes/darktable.css
@@ -389,13 +389,25 @@ overshoot.right
   background-color: transparent;
   outline-style:none;
   border-radius: 2px;
-  padding: 0.4em;
   color: @section_label;
+}
+
+#view_dropdown menu {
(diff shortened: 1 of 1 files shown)


data/themes/darktable.css
```diff
@@ -870,9 +886,9 @@ menuitem check
...
 
-/* Header and footer combobox */
-#header-toolbar combobox,
-#footer-toolbar combobox
+/* Header and footer labels */
+#header-toolbar label,
+#footer-toolbar label
 {
   margin: 0 0.5em;
 }
```
Nilvus, line 894: Do not remove combobox here, add it between toolbar and label for both header and footer line. That will avoid having that setting applied for some label we have outside a combobox.

data/themes/darktable.css
```diff
@@ -1073,7 +1089,7 @@ dialog scrollbar
   margin: 0px; /* reset default dialog margin to all items. And needed to have an hover and selected effect on all width of categories in sidebar */
 }
 
-#preferences_notebook decoration  /* allow to resize preferences dialog window */
+#preferences_notebook dialog > decoration  /* allow to resize preferences dialog window */
```
Nilvus, line 1092: A change really not useful as it works without it.

data/themes/darktable.css
```diff
@@ -825,16 +841,16 @@ dialog combobox window
...
   background-color: @tooltip_bg_color;
-  border-radius: 0.2em;
+  border-radius: 2px;
   border: 0;
   outline-style: none;
   min-height: 0.4em;
   min-width: 0.4em;
-  padding: 0.2em 0;
+  padding: 2px 0;
 }
```
Nilvus, line 850: Why changing them? Having related sizes is better here and 0.2em works great with your changes. So I suggest to put them back as that don't really change your PR.

data/themes/darktable.css
```diff
@@ -825,16 +841,16 @@ dialog combobox window
...
 
 menu menuitem {
-  padding: 0 0.2em;
+  padding: 0 2px;
```
Nilvus, line 853: Same here as above.

Nilvus (changes requested):
Great work. Combobox are better. Thanks for that work but just some minor comments and things to change on your PR.

data/themes/darktable.css
Nilvus: Of course it works with it too.

Nilvus:
>     * align button to the left in preferences

Last thing, buttons as they are actually was wanted. But testing your proposal make me think that something between actual state and your proposal would be better. Like just 1px between entry and buttons and between buttons.

data/themes/darktable.css
```diff
@@ -1148,9 +1164,13 @@ dialog scrollbar
...
 
+#preferences_notebook combobox menuitem
+{
+  padding: 0.2em;
+}
+
 #preferences_notebook button
 {
-  margin: 0 2px;
   padding: 0 2px;
```
Nilvus, line 1174: Just made some test and I think it would be better with margin like that:
`
margin-left: 1px;`

Nilvus (changes requested):

TurboGit:
Thanks !

PR #6439, opened 2020-10-03
Use same font as website for the logo

amousset:
How would you feel about using the same font for the logo as on https://www.darktable.org/ ? I feel like it could help maintaining a consistent appearance and look a bit more polished.

exo
(only a quick-and-dirty draft for now)


Nilvus:
Visually, why not. But this will need that all users have font installed. And that's a bad thing to add such specific font for that reason. Different users will have different rendering depending on fonts installed on their system. So not safe.

AlicVB:
one solution could be to use an image for 'darktable', to avoid any font problems. (to be checked if text color stay the same in all themes)

amousset:
This approach is indeed not practical, but it could probably be an svg with the text converted to paths.

Nilvus:
An svg would indeed be better.

TurboGit:
Agreed an SVG would better !

src/common/utility.c
```diff
@@ -377,17 +377,11 @@ dt_logo_season_t dt_util_get_logo_season(void)
   return DT_LOGO_SEASON_NONE;
 }
 
-cairo_surface_t *dt_util_get_logo(const float size)
+static cairo_surface_t *dt_util_get_svg_img(gchar *logo, const float size)
```
TurboGit, line 380: static routine should not start with dt_ but just _ (a coding convention).

TurboGit (changes requested):
A minor comment on style apart from that it looks good and work fine.

src/common/utility.c
amousset: :+1:

TurboGit:
All fine now, and thanks for this nice little detail for darktable visual identity !

PR #6444, opened 2020-10-04
Exposure-independent guided filtering for tone-equalizer

rawfiner:
### Short summary for @aurelienpierre and @hanatos that followed parts of the development of this filter:
I kept the original idea of dividing variance by the pixel value squared, and of using gaussian blurs. I do a weighted averaging of "a" and "b" instead of an unweighted averaging, which allows averaging a and b without the bright halos we had when doing an unweighted. This also gives much better result than if averaging of a and b is removed.

## Full description:
To illustrate my point, I will use an image from Gustavo Adolfo available here: https://discuss.pixls.us/t/scene-revisited/18694
Here is the image mask for toneequalizer without any smoothing:
Capture d’écran du 2020-10-04 14-59-22

### Guided filter problems
The guided filter main issue I had is that it is not exposure independent.
In toneequalizer, we can see that the highlights are much less smoothed than shadows:
Capture d’écran du 2020-10-04 14-59-41

Clouds are much less smoothed than the towns.

Also, if we change the mask exposure compensation, the blurring we get is completely different:
Capture d’écran du 2020-10-04 15-02-11

### Why this happens
Guided filters use the local variance to determine whether to blur a lot or not.
But variance is exposure-dependent: add 1EV and variance will be multiplied by 4.

### Proposed modification of the filter to solve it
Divide variance by the pixel value, squared.
This ratio will remain the same whatever the change of exposure.

### Cool, but...
After determining for each pixel how strong the blur will be at this point (let's call this blur strength to make explanation easier, for people who know guided filter, this is "a" and "b" coefficients), the guided filter does a local average of the blur strength.
But the local average gave bright halos (see on the middle-left):
Capture d’écran du 2020-10-04 15-11-36

### Proposed modification of the filter to solve that too
In Anisotropic guided filtering (https://ieeexplore.ieee.org/document/8844987/), the authors propose to use a weighted average of the blur strength.
I took this idea, even though I don't use the same weight at all.
I weight the blur by 1/pixel_value^2. This weight is completely empirical, but removes very well these halos (at the expense of a less smoothing near very strong edges).
Capture d’écran du 2020-10-04 17-27-45

### Changes in toneeq GUI
In GUI, I put EIGF as default, and changed accordingly default values of smoothing diameter and feathering. These default values should work fine in most cases.
Feathering of EIGF is adapted under the hood depending on smoothing diameter, this allows the user to increase the smoothing diameter with changing feathering and see directly the effect on its image with a very similar blur amount (without this, increasing smoothing diameter would give an image less and less smoothed. Note that this adaptation is imperfect, but better than nothing).

### DIFF
--- a/src/common/eigf.h
+++ b/src/common/eigf.h
@@ -60,116 +60,117 @@
  * just like what is done in fast_guided_filter.h
 **/
 
-static inline void exposure_independent_guided_filter(const float *const restrict guide, // I
+/* computes average and variance of guide and mask, and put them in out.
+ * out has 4 channels:
+ * - average of guide
+ * - variance of guide
+ * - average of mask
+ * - covariance of mask and guide. */
+static inline void eigf_variance_analysis(const float *const restrict guide, // I
                                     const float *const restrict mask, //p
-                                    float *const restrict ab,
+                                    float *const restrict out,
                                     const size_t width, const size_t height,
-                                    const float sigma, const float feathering)
+                                    const float sigma)
 {
   // We also use gaussian blurs instead of the square blurs of the guided filter
   const size_t Ndim = width * height;
-  float *const restrict blurred_guide = dt_alloc_sse_ps(Ndim);
-  // guide_x_guide = (guide - blurred_guide)^2
-  float *const restrict guide_x_guide = dt_alloc_sse_ps(Ndim);
-  // guide_variance = blur(guide_x_guide)
-  float *const restrict guide_variance = dt_alloc_sse_ps(Ndim);
-  float *const restrict guide_x_mask = dt_alloc_sse_ps(Ndim);
-  // guide_mask_covariance = blur(guide_x_mask)
-  float *const restrict guide_mask_covariance = dt_alloc_sse_ps(Ndim);
-  float *const restrict blurred_mask = dt_alloc_sse_ps(Ndim);
+  float *const restrict in = dt_alloc_sse_ps(Ndim * 4);
 
   float ming = 10000000.0f;
   float maxg = 0.0f;
   float minm = 10000000.0f;
   float maxm = 0.0f;
+  float ming2 = 10000000.0f;
+  float maxg2 = 0.0f;
+  float minmg = 10000000.0f;
+  float maxmg = 0.0f;
 #ifdef _OPENMP
 #pragma omp parallel for simd default(none) \
-dt_omp_firstprivate(guide, mask, Ndim) \
-  schedule(simd:static) aligned(guide, mask:64) \
-  reduction(max:maxg, maxm)\
-  reduction(min:ming, minm)
+dt_omp_firstprivate(guide, mask, in, Ndim) \
+  schedule(simd:static) aligned(guide, mask, in:64) \
+  reduction(max:maxg, maxm, maxg2, maxmg)\
+  reduction(min:ming, minm, ming2, minmg)
 #endif
   for(size_t k = 0; k < Ndim; k++)
   {
     const float pixelg = guide[k];
     const float pixelm = mask[k];
+    const float pixelg2 = pixelg * pixelg;
+    const float pixelmg = pixelm * pixelg;
+    in[k * 4] = pixelg;
+    in[k * 4 + 1] = pixelg2;
+    in[k * 4 + 2] = pixelm;
+    in[k * 4 + 3] = pixelmg;
     if(pixelg < ming) ming = pixelg;
     if(pixelg > maxg) maxg = pixelg;
     if(pixelm < minm) minm = pixelm;
     if(pixelm > maxm) maxm = pixelm;
+    if(pixelg2 < ming2) ming2 = pixelg2;
+    if(pixelg2 > maxg2) maxg2 = pixelg2;
+    if(pixelmg < minmg) minmg = pixelmg;
+    if(pixelmg > maxmg) maxmg = pixelmg;
   }
 
-  dt_gaussian_t *g = dt_gaussian_init(width, height, 1, &maxg, &ming, sigma, 0);
+  float max[4] = {maxg, maxg2, maxm, maxmg};
+  float min[4] = {ming, ming2, minm, minmg};
+  dt_gaussian_t *g = dt_gaussian_init(width, height, 4, max, min, sigma, 0);
   if(!g) return;
-  dt_gaussian_blur(g, guide, blurred_guide);
-  dt_gaussian_free(g);
-
-  g = dt_gaussian_init(width, height, 1, &maxm, &minm, sigma, 0);
-  if(!g) return;
-  dt_gaussian_blur(g, mask, blurred_mask);
-  dt_gaussian_free(g);
-
-  float mingg = 10000000.0f;
-  float maxgg = 0.0f;
-  float mingm = 10000000.0f;
-  float maxgm = 0.0f;
-#ifdef _OPENMP
-#pragma omp parallel for simd default(none) \
-  dt_omp_firstprivate(guide, mask, blurred_guide, blurred_mask, guide_x_guide, guide_x_mask, Ndim) \
-  schedule(simd:static) aligned(guide, mask, blurred_guide, blurred_mask, guide_x_guide, guide_x_mask:64) \
-  reduction(max:maxgg, maxgm)\
-  reduction(min:mingg, mingm)
-#endif
-  for(size_t k = 0; k < Ndim; k++)
-  {
-    const float deviation = guide[k] - blurred_guide[k];
-    const float squared_deviation = deviation * deviation;
-    guide_x_guide[k] = squared_deviation;
-    if(squared_deviation < mingg) mingg = squared_deviation;
-    if(squared_deviation > maxgg) maxgg = squared_deviation;
-    const float cov = (guide[k] - blurred_guide[k]) * (mask[k] - blurred_mask[k]);
-    guide_x_mask[k] = cov;
-    if(cov < mingm) mingm = cov;
-    if(cov > maxgm) maxgm = cov;
-  }
-
-  g = dt_gaussian_init(width, height, 1, &maxgg, &mingg, sigma, 0);
-  if(!g) return;
-  dt_gaussian_blur(g, guide_x_guide, guide_variance);
-  dt_gaussian_free(g);
-
-  g = dt_gaussian_init(width, height, 1, &maxgm, &mingm, sigma, 0);
-  if(!g) return;
-  dt_gaussian_blur(g, guide_x_mask, guide_mask_covariance);
+  dt_gaussian_blur_4c(g, in, out);
   dt_gaussian_free(g);
 
 #ifdef _OPENMP
 #pragma omp parallel for simd default(none) \
-  dt_omp_firstprivate(guide, mask, blurred_guide, guide_variance, blurred_mask, guide_mask_covariance, ab, Ndim, feathering) \
-  schedule(simd:static) aligned(guide, mask, blurred_guide, guide_variance, blurred_mask, guide_mask_covariance, ab:64)
+dt_omp_firstprivate(out, Ndim) \
+  schedule(simd:static) aligned(out:64)
 #endif
   for(size_t k = 0; k < Ndim; k++)
   {
-    const float normg = fmaxf(blurred_guide[k] * guide[k], 1E-6);
-    const float normm = fmaxf(blurred_mask[k] * mask[k], 1E-6);
-    const float normalized_var_guide = guide_variance[k] / normg;
-    const float normalized_covar = guide_mask_covariance[k] / sqrtf(normg * normm);
-    ab[2 * k] = normalized_covar / (normalized_var_guide + feathering);
-    ab[2 * k + 1] = blurred_mask[k] - ab[2 * k] * blurred_guide[k];
+    out[4 * k + 1] -= out[4 * k] * out[4 * k];
+    out[4 * k + 3] -= out[4 * k] * out[4 * k + 2];
   }
 
-  dt_free_align(blurred_guide);
-  dt_free_align(guide_x_guide);
-  dt_free_align(guide_variance);
-  dt_free_align(guide_x_mask);
-  dt_free_align(guide_mask_covariance);
-  dt_free_align(blurred_mask);
+  dt_free_align(in);
+}
+
+void eigf_blending(float *const restrict image, const float *const restrict mask,
+                  const float *const restrict av, const size_t Ndim,
+                  const dt_iop_guided_filter_blending_t filter,
+                  const float feathering)
+{
+#ifdef _OPENMP
+#pragma omp parallel for simd default(none) \
+  dt_omp_firstprivate(image, mask, av, Ndim, feathering, filter) \
+  schedule(simd:static) aligned(image, mask, av:64)
+#endif
+  for(size_t k = 0; k < Ndim; k++)
+  {
+    const float avg_g = av[k * 4];
+    const float avg_m = av[k * 4 + 2];
+    const float var_g = av[k * 4 + 1];
+    const float covar_mg = av[k * 4 + 3];
+    const float norm_g = fmaxf(avg_g * image[k], 1E-6);
+    const float norm_m = fmaxf(avg_m * mask[k], 1E-6);
+    const float normalized_var_guide = var_g / norm_g;
+    const float normalized_covar = covar_mg / sqrtf(norm_g * norm_m);
+    const float a = normalized_covar / (normalized_var_guide + feathering);
+    const float b = avg_m - a * avg_g;
+    if(filter == DT_GF_BLENDING_LINEAR)
+    {
+      image[k] = fmaxf(image[k] * a + b, MIN_FLOAT);
+    }
+    else
+    {
+      // filter == DT_GF_BLENDING_GEOMEAN
+      image[k] *= fmaxf(image[k] * a + b, MIN_FLOAT);
+      image[k] = sqrtf(image[k]);
+    }
+  }
 }
 
 __DT_CLONE_TARGETS__
 static inline void fast_eigf_surface_blur(float *const restrict image,
                                       const size_t width, const size_t height,
-                                      const int radius, float feathering, const int iterations,
+                                      const float sigma, float feathering, const int iterations,
                                       const dt_iop_guided_filter_blending_t filter, const float scale,
                                       const float quantization, const float quantize_min, const float quantize_max)
 {
@@ -178,8 +179,8 @@ static inline void fast_eigf_surface_blur(float *const restrict image,
 
   // A down-scaling of 4 seems empirically safe and consistent no matter the image zoom level
   // see reference paper above for proof.
-  const float scaling = 1.0f;
-  const float ds_sigma = fmaxf((float)radius / scaling, 1.0f);
+  const float scaling = fmaxf(fminf(sigma, 4.0f), 1.0f);
+  const float ds_sigma = fmaxf(sigma / scaling, 1.0f);
 
   const size_t ds_height = height / scaling;
   const size_t ds_width = width / scaling;
@@ -187,48 +188,46 @@ static inline void fast_eigf_surface_blur(float *const restrict image,
   const size_t num_elem_ds = ds_width * ds_height;
   const size_t num_elem = width * height;
 
+  float *const restrict mask = dt_alloc_sse_ps(dt_round_size_sse(num_elem));
   float *const restrict ds_image = dt_alloc_sse_ps(dt_round_size_sse(num_elem_ds));
   float *const restrict ds_mask = dt_alloc_sse_ps(dt_round_size_sse(num_elem_ds));
-  float *const restrict ds_ab = dt_alloc_sse_ps(dt_round_size_sse(num_elem_ds * 2));
-  float *const restrict ab = dt_alloc_sse_ps(dt_round_size_sse(num_elem * 2));
+  // average - variance arrays: store the guide and mask averages and variances
+  float *const restrict ds_av = dt_alloc_sse_ps(dt_round_size_sse(num_elem_ds * 4));
+  float *const restrict av = dt_alloc_sse_ps(dt_round_size_sse(num_elem * 4));
 
-  if(!ds_image || !ds_mask || !ds_ab || !ab)
+  if(!ds_image || !ds_mask || !ds_av || !av)
   {
     dt_control_log(_("fast exposure independent guided filter failed to allocate memory, check your RAM settings"));
     goto clean;
   }
 
-  // Downsample the image for speed-up
-  interpolate_bilinear(image, width, height, ds_image, ds_width, ds_height, 1);
-
-  // empirical formula to have consistent smoothing when increasing the radius
-  const float adapted_feathering = feathering * radius * sqrt(radius) / 40.0f;
   // Iterations of filter models the diffusion, sort of
   for(int i = 0; i < iterations; i++)
   {
     // (Re)build the mask from the quantized image to help guiding
-    quantize(ds_image, ds_mask, ds_width * ds_height, quantization, quantize_min, quantize_max);
-    exposure_independent_guided_filter(ds_mask, ds_image, ds_ab, ds_width, ds_height, ds_sigma, adapted_feathering);
-
+    quantize(image, mask, width * height, quantization, quantize_min, quantize_max);
+    // Downsample the image for speed-up
+    interpolate_bilinear(image, width, height, ds_image, ds_width, ds_height, 1);
+    interpolate_bilinear(mask, width, height, ds_mask, ds_width, ds_height, 1);
+    eigf_variance_analysis(ds_mask, ds_image, ds_av, ds_width, ds_height, ds_sigma);
+    // Upsample the variances and averages
+    interpolate_bilinear(ds_av, ds_width, ds_height, av, width, height, 4);
     if(i != iterations - 1)
     {
       // Process the intermediate filtered image
-      apply_linear_blending(ds_image, ds_ab, num_elem_ds);
+      eigf_blending(image, mask, av, num_elem, DT_GF_BLENDING_LINEAR, feathering);
+    }
+    else
+    {
+      // Finally, blend the guided image
+      eigf_blending(image, mask, av, num_elem, filter, feathering);
     }
   }
 
-  // Upsample the blending parameters a and b
-  interpolate_bilinear(ds_ab, ds_width, ds_height, ab, width, height, 2);
-
-  // Finally, blend the guided image
-  if(filter == DT_GF_BLENDING_LINEAR)
-    apply_linear_blending(image, ab, num_elem);
-  else if(filter == DT_GF_BLENDING_GEOMEAN)
-    apply_linear_blending_w_geomean(image, ab, num_elem);
-
 clean:
-  if(ab) dt_free_align(ab);
-  if(ds_ab) dt_free_align(ds_ab);
+  if(av) dt_free_align(av);
+  if(ds_av) dt_free_align(ds_av);
   if(ds_mask) dt_free_align(ds_mask);
   if(ds_image) dt_free_align(ds_image);
+  if(mask) dt_free_align(mask);
 }
--- a/src/common/eigf.h
+++ b/src/common/eigf.h
@@ -132,6 +132,61 @@ dt_omp_firstprivate(out, Ndim) \
   dt_free_align(in);
 }
 
+// same function as above, but specialized for the case where guide == mask
+// for increased performance
+static inline void eigf_variance_analysis_no_mask(const float *const restrict guide, // I
+                                    float *const restrict out,
+                                    const size_t width, const size_t height,
+                                    const float sigma)
+{
+  // We also use gaussian blurs instead of the square blurs of the guided filter
+  const size_t Ndim = width * height;
+  float *const restrict in = dt_alloc_sse_ps(Ndim * 2);
+
+  float ming = 10000000.0f;
+  float maxg = 0.0f;
+  float ming2 = 10000000.0f;
+  float maxg2 = 0.0f;
+#ifdef _OPENMP
+#pragma omp parallel for simd default(none) \
+dt_omp_firstprivate(guide, in, Ndim) \
+  schedule(simd:static) aligned(guide, in:64) \
+  reduction(max:maxg, maxg2)\
+  reduction(min:ming, ming2)
+#endif
+  for(size_t k = 0; k < Ndim; k++)
+  {
+    const float pixelg = guide[k];
+    const float pixelg2 = pixelg * pixelg;
+    in[2 * k] = pixelg;
+    in[2 * k + 1] = pixelg2;
+    if(pixelg < ming) ming = pixelg;
+    if(pixelg > maxg) maxg = pixelg;
+    if(pixelg2 < ming2) ming2 = pixelg2;
+    if(pixelg2 > maxg2) maxg2 = pixelg2;
+  }
+
+  float max[2] = {maxg, maxg2};
+  float min[2] = {ming, ming2};
+  dt_gaussian_t *g = dt_gaussian_init(width, height, 2, max, min, sigma, 0);
+  if(!g) return;
+  dt_gaussian_blur(g, in, out);
+  dt_gaussian_free(g);
+
+#ifdef _OPENMP
+#pragma omp parallel for simd default(none) \
+dt_omp_firstprivate(out, Ndim) \
+  schedule(simd:static) aligned(out:64)
+#endif
+  for(size_t k = 0; k < Ndim; k++)
+  {
+    const float avg = out[2 * k];
+    out[2 * k + 1] -= avg * avg;
+  }
+
+  dt_free_align(in);
+}
+
 void eigf_blending(float *const restrict image, const float *const restrict mask,
                   const float *const restrict av, const size_t Ndim,
                   const dt_iop_guided_filter_blending_t filter,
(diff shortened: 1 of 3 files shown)


src/common/eigf.h
```diff
@@ -0,0 +1,302 @@
...
+  // guide_mask_covariance = blur(guide_x_mask)
+  float *const restrict guide_mask_covariance = dt_alloc_sse_ps(Ndim);
+  float *const restrict blurred_mask = dt_alloc_sse_ps(Ndim);
+  float *const restrict a = dt_alloc_sse_ps(Ndim);
+  float *const restrict b = dt_alloc_sse_ps(Ndim);
+  // weight to compute the weighted blur of a and b
+  float *const restrict weights = dt_alloc_sse_ps(Ndim);
+  float *const restrict blurred_a = dt_alloc_sse_ps(Ndim);
+  float *const restrict blurred_b = dt_alloc_sse_ps(Ndim);
+  float *const restrict blurred_weights = dt_alloc_sse_ps(Ndim);
```
aurelienpierre, line 96: Can't you reduce the number of memory buffers ? In fast_guided_filter.h, I use :
```
  const size_t Ndim = width * height;
  const size_t Ndimch = width * height * 4;

  float *const restrict temp = dt_alloc_sse_ps(Ndimch); // array of structs { { mean_I, mean_p, corr_I, corr_Ip } }
  float *const restrict guide_x_mask = dt_alloc_sse_ps(Ndim);
  float *const restrict guide_x_guide = dt_alloc_sse_ps(Ndim);
```
and output an `{a, b}` vector that gets blurred in a single loop × 2 channels. Everything else is computed in-loop, with the blur merged in there.

src/common/eigf.h
```diff
@@ -0,0 +1,302 @@
...
+  {
+    const float pixelg = guide[k];
+    const float pixelm = mask[k];
+    if(pixelg < ming) ming = pixelg;
+    if(pixelg > maxg) maxg = pixelg;
+    if(pixelm < minm) minm = pixelm;
+    if(pixelm > maxm) maxm = pixelm;
+  }
+
+  dt_gaussian_t *g = dt_gaussian_init(width, height, 1, &maxg, &ming, sigma, 0);
```
aurelienpierre, line 119: So you blur the guide and the image first, doesn't that create halos around edges ?

src/common/eigf.h
rawfiner: Actually, to use gaussian blur I need either a 1 channel buffer, or a 4 channels buffer. That's why I have a and b in separate buffers

src/common/eigf.h
rawfiner: The first blurs are here to compute the esperance, which is in particular used to compute variance (you have these blurs also in the original guided filter, that's your tmp[0] and tmp[1] in your implementation)

src/common/eigf.h
```diff
@@ -0,0 +1,302 @@
...
+#ifdef _OPENMP
+#pragma omp parallel for simd default(none) \
+  dt_omp_firstprivate(guide, mask, blurred_guide, blurred_mask, guide_x_guide, guide_x_mask, Ndim) \
+  schedule(simd:static) aligned(guide, mask, blurred_guide, blurred_mask, guide_x_guide, guide_x_mask:64) \
+  reduction(max:maxgg, maxgm)\
+  reduction(min:mingg, mingm)
+#endif
+  for(size_t k = 0; k < Ndim; k++)
+  {
+    const float deviation = guide[k] - blurred_guide[k];
```
aurelienpierre, line 142: here, if you had one array of vectors `{ guide, mask, guide_x_guide, guide_x_mask }`, you could blur them horizontally all in that loop, with SIMD, then compute the vertical blur in another loop, where the covariance and (a, b) computation can happen after the convolution. That would save a lot of I∕O, take advantage of data alignment and avoid using the `dt_gaussian_xx` lib which is quite slow IIRC.

rawfiner: Ok, I will use box averages instead of gaussian blurs and do that

src/common/eigf.h
```diff
@@ -0,0 +1,302 @@
...
+  reduction(min:mina, minb, minw)
+#endif
+  for(size_t k = 0; k < Ndim; k++)
+  {
+    const float pixelg = fmaxf(guide[k], 1E-6);
+    const float pixelm = fmaxf(mask[k], 1E-6);
+    const float normalized_var_guide = guide_variance[k] / (pixelg * pixelg);
+    const float normalized_covar = guide_mask_covariance[k] / (pixelg * pixelm);
+    // empirical weight
+    float w = 1.f / pixelg;
```
aurelienpierre, line 183: I wonder if `w = 1.f / (blurred_mask[k] * blurred_guide[k])` would not work better.

rawfiner: here is a comparison
using `w = 1.f / (blurred_mask[k] * blurred_guide[k])`
Capture d’écran du 2020-10-05 16-18-29
using `w = 1.f / (mask[k] * guide[k])`
Capture d’écran du 2020-10-05 16-19-25

The problem with `w = 1.f / (blurred_mask[k] * blurred_guide[k])` is that `blurred_guide` (and `blurred_mask`) is a very bad approximation of guide in the dark area which are close to bright areas

rawfiner:
This push:
- adds a box-blur based EIGF. Implementation aims to choose between this and gaussian-blur based EIGF, so I modified guided filter code to have box-blur based EIGF (switching between the 2 is controled by a macro, the idea was to keep maximal performance for the proof of concept to compare correctly with gaussian-blur based EIGF for performance). Implementation follows the same steps as gaussian-blur based EIGF, except the fact that a and b are blurred twice at the end, otherwise one could see too "harsh" boundaries between blurred and sharp areas due to box blurs. In GUI, I temporarily removed the possibility to select the "standard" guided filter, just while we decide which filter to keep.
- change a bit the formulas used for the weight and the normalized variance: new formulas keep exposure independence of course, but are less sensitive to noise

On my computer, performance is better with gaussian-blur based EIGF, especially for large radius:
| smoothing diameter | box-blur based EIGF | gaussian-blur based EIGF |
|--------------------|---------------------|--------------------------|
| 5%                 | 2s                  | 0.8s                     |
| 25%                | 6.9s                | 0.8s                     |

Now we need to decide between box-blur and gaussian-blur, and then I will clean up the PR, removing the implementation we decide to drop

src/common/eigf.h
```diff
@@ -0,0 +1,301 @@
...
+#pragma omp parallel for simd default(none) \
+  dt_omp_firstprivate(guide, mask, blurred_guide, guide_variance, blurred_mask, guide_mask_covariance, a, b, weights, Ndim, feathering) \
+  schedule(simd:static) aligned(guide, mask, blurred_guide, guide_variance, blurred_mask, guide_mask_covariance, a, b, weights:64) \
+  reduction(max:maxa, maxb, maxw)\
+  reduction(min:mina, minb, minw)
+#endif
+  for(size_t k = 0; k < Ndim; k++)
+  {
+    // allow max 2EV below average for normalization (avoid noise issues)
+    const float pixelg = fmaxf(guide[k], fmaxf(blurred_guide[k] * 0.25f, 1E-6));
```
rawfiner, line 179: Here, we can see the change that give better noise tolerance:
we normalize using the maximum between the pixel value and the local average - 2EV.
In other word, we consider any pixel which is below local average - 2EV to be exactly 2EV below average.
As both the pixel value and 0.25 * average are changed the same way when changing exposure, the filter remains exposure-invariant

aurelienpierre:
Nice, will test this week

src/common/eigf.h
aurelienpierre: ok, the example is self-explanatory. Thanks.

aurelienpierre:
Here are my runtimes (4K preview, Intel Xeon @ 4.1 GHz × 8 ):
* radius 25 %:
  * box-blur EIGF : 0.290–0.516 s
  * gaussian blur EIGF : 0.219–0.242 s 
* radius 5% : 
  * box-blur : 0.289–0.313 s
  * gaussian-blur EIGF : 0.231–0.251 s

So it seems, indeed, that the gaussian approach is better. It also gives a subjectively better masking, with better separations in shadows. 

However, I would still blur `guide, mask, guide_x_mask, guide_x_guide` as a 4 channels image, in one shot.

rawfiner:
Actually, we can indeed use a 4 channels gaussian blur for that, but it is a trade-of between performance and accuracy.
Doing only 1channel blurs, I can compute variance doing E[X-E[X]] instead of using E[X^2]-E[X]^2 (which allows to use the 4 channels blur, but can give catastrophic cancellation)

In our case, we don't need that much accuracy to compute variance: catastrophic cancellation would appear in areas with high average and low variance, and we want to blur them anyway so it should not be a problem if variance gets equal to 0 in such cases.

So I will try to use the 4 channel gaussian to see if we can get some performance improvement, good idea

aurelienpierre:
@rawfiner please fine my latest version, probably more accurate to the anisotropic GF than the previous one I sent you today (with correct weights normalization). Blur radii are usually good between 1-5 % and feathering values lie between 50-100.

```
diff --git a/src/common/eigf.h b/src/common/eigf.h
index da7c1a62f..5b802ec9e 100644
--- a/src/common/eigf.h
+++ b/src/common/eigf.h
@@ -166,32 +166,67 @@ dt_omp_firstprivate(guide, mask, Ndim) \
   float maxa = 0.0f;
   float maxb = 0.0f;
   float maxw = 0.0f;
+
+  // The by-the-book gaussian filter is normalized such that the sum of the kernel elements == 1.
+  // but here we want to defer the normalization for later and have the weight == 1.f in center of the window
+  // so we multiply the image by norm before the gaussian filter to get the weights in [0; 1]
+  const float norm = sigma * sqrtf(2.f * M_PI);
+
 #ifdef _OPENMP
-  dt_omp_firstprivate(guide, mask, blurred_guide, guide_variance, blurred_mask, guide_mask_covariance, a, b, weights, Ndim, feathering) \
-  schedule(simd:static) aligned(guide, mask, blurred_guide, guide_variance, blurred_mask, guide_mask_covariance, a, b, weights:64) \
-  reduction(max:maxa, maxb, maxw)\
-  reduction(min:mina, minb, minw)
+  dt_omp_firstprivate(guide, mask, blurred_guide, guide_variance, blurred_mask, guide_mask_covariance, weights, Ndim, feathering, norm) \
+  schedule(simd:static) \
+  aligned(guide, mask, blurred_guide, guide_variance, blurred_mask, guide_mask_covariance, weights:64) \
+  reduction(max:maxw)\
+  reduction(min:minw)
 #endif
   for(size_t k = 0; k < Ndim; k++)
   {
+    // normalize variance and covariance, aka make them exposure-invariant
     // allow max 2EV below average for normalization (avoid noise issues)
-    const float pixelg = fmaxf(guide[k], fmaxf(blurred_guide[k] * 0.25f, 1E-6));
-    const float pixelm = fmaxf(mask[k], fmaxf(blurred_mask[k] * 0.25f, 1E-6));
-    const float normalized_var_guide = guide_variance[k] / (pixelg * pixelg);
-    const float normalized_covar = guide_mask_covariance[k] / (pixelg * pixelm);
-    // empirical weight. allow 5EV below average here.
-    float w = 1.f / (fmaxf(guide[k], fmaxf(blurred_guide[k] * 0.03f, 1E-6)));
-    w *= w;
-    a[k] = w * normalized_covar / (normalized_var_guide + feathering);
-    b[k] = w * blurred_mask[k] - a[k] * blurred_guide[k];
+    const float pixelg = fmaxf(sqrtf(blurred_guide[k] * guide[k]), 1e-6f);
+    const float pixelm = fmaxf(sqrtf(blurred_mask[k] * mask[k]), 1e-6f);
+    guide_variance[k] /= (pixelg * pixelg);
+    guide_mask_covariance[k] /= (pixelg * pixelm);
+
+    const float epsilon = guide_mask_covariance[k]; // ??
+    const float alpha = 128.f; // anisotropy coeff
+    const float w = norm * epsilon / (powf(guide_variance[k], alpha) + epsilon);
+
     weights[k] = w;
+    if(w < minw) minw = w;
+    if(w > maxw) maxw = w;
+  }
+
+  const float sigma_2 = sigma / 4.f;
+
+  g = dt_gaussian_init(width, height, 1, &maxw, &minw, sigma_2, 0);
+  if(!g) return;
+  dt_gaussian_blur(g, weights, blurred_weights);
+  dt_gaussian_free(g);
+
+  const float normalize_2 = sigma * sqrtf(2.f * M_PI);
+
+#ifdef _OPENMP
+#pragma omp parallel for simd default(none) \
+  dt_omp_firstprivate(blurred_guide, guide_variance, blurred_mask, guide_mask_covariance, a, b, blurred_weights, Ndim, feathering, normalize_2) \
+  schedule(simd:static) aligned(blurred_guide, guide_variance, blurred_mask, guide_mask_covariance, a, b, blurred_weights:64) \
+  reduction(max:maxa, maxb)\
+  reduction(min:mina, minb)
+#endif
+  for(size_t k = 0; k < Ndim; k++)
+  {
+    const float w = blurred_weights[k];
+    a[k] = w * guide_mask_covariance[k] / (guide_variance[k] + feathering);
+    b[k] = w * blurred_mask[k] - a[k] * blurred_guide[k];
+
+    // same as before, denormalize the gaussian filter ahead because we defer normalization
+    blurred_weights[k] *= normalize_2;
+
     if(a[k] < mina) mina = a[k];
     if(b[k] < minb) minb = b[k];
-    if(w < minw) minw = w;
     if(a[k] > maxa) maxa = a[k];
     if(b[k] > maxb) maxb = b[k];
-    if(w > maxw) maxw = w;
   }
 
   g = dt_gaussian_init(width, height, 1, &maxa, &mina, sigma, 0);
@@ -204,21 +239,22 @@ dt_omp_firstprivate(guide, mask, Ndim) \
   dt_gaussian_blur(g, b, blurred_b);
   dt_gaussian_free(g);
 
-  g = dt_gaussian_init(width, height, 1, &maxw, &minw, sigma, 0);
+  // Compute the sum of the mask values over the window
+  float *const restrict normalize = dt_alloc_sse_ps(Ndim);
+  g = dt_gaussian_init(width, height, 1, &maxb, &minb, sigma, 0);
   if(!g) return;
-  dt_gaussian_blur(g, weights, blurred_weights);
+  dt_gaussian_blur(g, blurred_weights, normalize);
   dt_gaussian_free(g);
 
 #ifdef _OPENMP
-  dt_omp_firstprivate(ab, blurred_a, blurred_b, blurred_weights, Ndim) \
-  schedule(simd:static) aligned(ab, blurred_a, blurred_b, blurred_weights:64)
+  dt_omp_firstprivate(ab, blurred_a, blurred_b, blurred_weights, Ndim, normalize) \
+  schedule(simd:static) aligned(ab, blurred_a, blurred_b, blurred_weights, normalize:64)
 #endif
   for(size_t k = 0; k < Ndim; k++)
   {
-    const float normalize = blurred_weights[k];
-    ab[2 * k] = blurred_a[k] / normalize;
-    ab[2 * k + 1] = blurred_b[k] / normalize;
+    ab[2 * k] = blurred_a[k] / normalize[k];
+    ab[2 * k + 1] = blurred_b[k] / normalize[k];
   }
 
   dt_free_align(blurred_guide);
@@ -233,6 +269,7 @@ dt_omp_firstprivate(guide, mask, Ndim) \
   dt_free_align(blurred_a);
   dt_free_align(blurred_b);
   dt_free_align(blurred_weights);
+  dt_free_align(normalize);
 }
 
 __DT_CLONE_TARGETS__
diff --git a/src/iop/toneequal.c b/src/iop/toneequal.c
index 6b7ffb381..d165ab060 100644
--- a/src/iop/toneequal.c
+++ b/src/iop/toneequal.c
@@ -3128,7 +3128,7 @@ void gui_init(struct dt_iop_module_t *self)
                                              "preview much slower if denoise profiled is used."));
 
   g->feathering = dt_bauhaus_slider_from_params(self, "feathering");
-  dt_bauhaus_slider_set_soft_range(g->feathering, 1.0, 50.0);
+  dt_bauhaus_slider_set_soft_range(g->feathering, 0.001, 50.0);
   dt_bauhaus_slider_set_step(g->feathering, 0.2);
   gtk_widget_set_tooltip_text(g->feathering, _("precision of the feathering :\n"
                                                "higher values force the mask to follow edges more closely\n"
```

rawfiner:
Thank you @aurelienpierre 
I did a quick test, it behaves very strangely for very high feathering (I tried with 10000, we should see the original image, but I get an image with much higher contrast instead)
It blurs much better, but there are way more halos (both bright halos and dark halos) as well, including with feathering=100 (for example on the image from here: https://discuss.pixls.us/t/yet-another-sunset-over-the-lake/18739 ). I will test again tomorrow when I have more time

matt-maguire:
> I'm really in the dark regarding that epsilon.

It looks similar to the epsilon in this paper:
http://mmlab.ie.cuhk.edu.hk/pdf/pami12Guided%20Image%20Filtering.pdf

but I don't get in (24) above why the numerator is multiplied by epsilon (I don't have the paper you are quoting it from).

aurelienpierre:
> It looks similar to the epsilon in this paper:
http://mmlab.ie.cuhk.edu.hk/pdf/pami12Guided%20Image%20Filtering.pdf

The epsilon in this paper is the variance threshold, dubbed "feathering" in the tone EQ GUI, that part is clear ;-) The other paper I'm referring to calls it gamma.

aurelienpierre:
@rawfiner v0.2 with normalization and sigmas corrected, downsampling reduced and the epsilon param adjusted to the average guide variance over the whole image (+ luminance display gets a brightening for better reading). Try blending radii between 0.5% and 2%, and feathering between 20-50.

```
diff --git a/src/common/eigf.h b/src/common/eigf.h
index da7c1a62f..ff4a34183 100644
--- a/src/common/eigf.h
+++ b/src/common/eigf.h
@@ -21,6 +21,60 @@
 #include "common/fast_guided_filter.h"
 #include "common/gaussian.h"
 
+static inline void normalize_kernel(const float *const restrict in, float *const restrict weight,
+                                    const size_t width, const size_t height,
+                                    const int radius)
+{
+  // Get the sum of an implicit random kernel and output it as a normalization coefficient
+
+  const size_t Ndim = width * height;
+  float *const restrict temp = dt_alloc_sse_ps(Ndim);
+
+  // Convolve box average along columns
+#ifdef _OPENMP
+#pragma omp parallel for default(none) \
+  dt_omp_firstprivate(in, temp, width, height, radius) \
+  schedule(simd:static) collapse(2)
+#endif
+  for(size_t j = 0; j < width; j++)
+  {
+    for(size_t i = 0; i < height; i++)
+    {
+      const size_t begin_convol = (i < radius) ? 0 : i - radius;
+      size_t end_convol = i + radius;
+      end_convol = (end_convol < height) ? end_convol : height - 1;
+      const size_t index = (i * width + j);
+
+      float w = 0.0f;
+      for(size_t c = begin_convol; c <= end_convol; c++) w += in[(c * width + j)];
+      temp[index] = w;
+    }
+  }
+
+  // Convolve box average along rows and output result
+#ifdef _OPENMP
+#pragma omp parallel for default(none) \
+  dt_omp_firstprivate(in, temp, width, height, radius, weight) \
+  schedule(simd:static) collapse(2)
+#endif
+  for(size_t i = 0; i < height; i++)
+  {
+    for(size_t j = 0; j < width; j++)
+    {
+      const size_t begin_convol = (j < radius) ? 0 : j - radius;
+      size_t end_convol = j + radius;
+      end_convol = (end_convol < width) ? end_convol : width - 1;
+      const size_t stride = i * width;
+      const size_t index = (stride + j);
+      float w  = 0.0f;
+      for(size_t c = begin_convol; c <= end_convol; c++) w += temp[(stride + c)];
+      weight[index] = w / w;
+    }
+  }
+
+  if(temp != NULL) dt_free_align(temp);
+}
+
 /***
  * DOCUMENTATION
  *
@@ -166,32 +220,62 @@ dt_omp_firstprivate(guide, mask, Ndim) \
   float maxa = 0.0f;
   float maxb = 0.0f;
   float maxw = 0.0f;
+
+  // The by-the-book gaussian filter is normalized such that the sum of the kernel elements == 1.
+  // but here we want to defer the normalization for later and have the weight == 1.f in center of the window
+  // so we multiply the image by norm before the gaussian filter to get the weights in [0; 1]
+  const float norm = sigma * sqrtf(2.f * M_PI);
+
 #ifdef _OPENMP
-  dt_omp_firstprivate(guide, mask, blurred_guide, guide_variance, blurred_mask, guide_mask_covariance, a, b, weights, Ndim, feathering) \
-  schedule(simd:static) aligned(guide, mask, blurred_guide, guide_variance, blurred_mask, guide_mask_covariance, a, b, weights:64) \
-  reduction(max:maxa, maxb, maxw)\
-  reduction(min:mina, minb, minw)
+  dt_omp_firstprivate(guide, mask, blurred_guide, guide_variance, blurred_mask, guide_mask_covariance, weights, Ndim, feathering, norm) \
+  schedule(simd:static) \
+  aligned(guide, mask, blurred_guide, guide_variance, blurred_mask, guide_mask_covariance, weights:64) \
+  reduction(max:maxw)\
+  reduction(min:minw)
 #endif
   for(size_t k = 0; k < Ndim; k++)
   {
+    // normalize variance and covariance, aka make them exposure-invariant
     // allow max 2EV below average for normalization (avoid noise issues)
     const float pixelg = fmaxf(guide[k], fmaxf(blurred_guide[k] * 0.25f, 1E-6));
     const float pixelm = fmaxf(mask[k], fmaxf(blurred_mask[k] * 0.25f, 1E-6));
-    const float normalized_var_guide = guide_variance[k] / (pixelg * pixelg);
-    const float normalized_covar = guide_mask_covariance[k] / (pixelg * pixelm);
+    guide_variance[k] /= (pixelg * pixelg);
+    guide_mask_covariance[k] /= (pixelg * pixelm);
+
+    const float epsilon = 1.f; // ??
+    const float alpha = 8.0f; // anisotropy coeff
+    const float w = norm * epsilon / (powf(guide_variance[k], alpha) + epsilon);
+
+    weights[k] = w;
+    if(w < minw) minw = w;
+    if(w > maxw) maxw = w;
+  }
+
+  g = dt_gaussian_init(width, height, 1, &maxw, &minw, sigma / 4.f, 0);
+  if(!g) return;
+  dt_gaussian_blur(g, weights, blurred_weights);
+  dt_gaussian_free(g);
+
+
+#ifdef _OPENMP
+#pragma omp parallel for simd default(none) \
+  dt_omp_firstprivate(blurred_guide, guide_variance, blurred_mask, guide_mask_covariance, a, b, blurred_weights, Ndim, feathering) \
+  schedule(simd:static) aligned(blurred_guide, guide_variance, blurred_mask, guide_mask_covariance, a, b, blurred_weights:64) \
+  reduction(max:maxa, maxb)\
+  reduction(min:mina, minb)
+#endif
+  for(size_t k = 0; k < Ndim; k++)
+  {
     // empirical weight. allow 5EV below average here.
-    float w = 1.f / (fmaxf(guide[k], fmaxf(blurred_guide[k] * 0.03f, 1E-6)));
-    w *= w;
-    a[k] = w * normalized_covar / (normalized_var_guide + feathering);
+    //const float epsilon = fmaxf(guide[k], fmaxf(blurred_guide[k] * 0.03f, 1E-6));
+    const float w = blurred_weights[k];
+    a[k] = w * guide_mask_covariance[k] / (guide_variance[k] + feathering);
     b[k] = w * blurred_mask[k] - a[k] * blurred_guide[k];
-    weights[k] = w;
     if(a[k] < mina) mina = a[k];
     if(b[k] < minb) minb = b[k];
-    if(w < minw) minw = w;
     if(a[k] > maxa) maxa = a[k];
     if(b[k] > maxb) maxb = b[k];
-    if(w > maxw) maxw = w;
   }
 
   g = dt_gaussian_init(width, height, 1, &maxa, &mina, sigma, 0);
@@ -204,21 +288,22 @@ dt_omp_firstprivate(guide, mask, Ndim) \
   dt_gaussian_blur(g, b, blurred_b);
   dt_gaussian_free(g);
 
-  g = dt_gaussian_init(width, height, 1, &maxw, &minw, sigma, 0);
+  // Compute the sum of the mask values over the window
+  float *const restrict normalize = dt_alloc_sse_ps(Ndim);
+  g = dt_gaussian_init(width, height, 1, &maxb, &minb, sigma, 0);
   if(!g) return;
-  dt_gaussian_blur(g, weights, blurred_weights);
+  dt_gaussian_blur(g, blurred_weights, normalize);
   dt_gaussian_free(g);
 
 #ifdef _OPENMP
-  dt_omp_firstprivate(ab, blurred_a, blurred_b, blurred_weights, Ndim) \
-  schedule(simd:static) aligned(ab, blurred_a, blurred_b, blurred_weights:64)
+  dt_omp_firstprivate(ab, blurred_a, blurred_b, blurred_weights, Ndim, normalize) \
+  schedule(simd:static) aligned(ab, blurred_a, blurred_b, blurred_weights, normalize:64)
 #endif
   for(size_t k = 0; k < Ndim; k++)
   {
-    const float normalize = blurred_weights[k];
-    ab[2 * k] = blurred_a[k] / normalize;
-    ab[2 * k + 1] = blurred_b[k] / normalize;
+    ab[2 * k] = blurred_a[k] / normalize[k];
+    ab[2 * k + 1] = blurred_b[k] / normalize[k];
   }
 
   dt_free_align(blurred_guide);
@@ -233,6 +318,7 @@ dt_omp_firstprivate(guide, mask, Ndim) \
   dt_free_align(blurred_a);
   dt_free_align(blurred_b);
   dt_free_align(blurred_weights);
+  dt_free_align(normalize);
 }
 
 __DT_CLONE_TARGETS__
diff --git a/src/iop/toneequal.c b/src/iop/toneequal.c
index 6b7ffb381..d165ab060 100644
--- a/src/iop/toneequal.c
+++ b/src/iop/toneequal.c
@@ -3128,7 +3128,7 @@ void gui_init(struct dt_iop_module_t *self)
                                              "preview much slower if denoise profiled is used."));
 
   g->feathering = dt_bauhaus_slider_from_params(self, "feathering");
-  dt_bauhaus_slider_set_soft_range(g->feathering, 1.0, 50.0);
+  dt_bauhaus_slider_set_soft_range(g->feathering, 0.001, 50.0);
   dt_bauhaus_slider_set_step(g->feathering, 0.2);
   gtk_widget_set_tooltip_text(g->feathering, _("precision of the feathering :\n"
                                                "higher values force the mask to follow edges more closely\n"
```

rawfiner:
After some iterations with @aurelienpierre , we decided to remove the last blur that blurs a and b.
This blur improves the smoothing, but creates halos, which are hard to remove. Doing a weighted blur removes the halos, but create large areas without smoothing.
Results are finally better without this blur.

However, the "downscaling - process - upscale - finish" strategy had to be adapted a little, as computing a and b on downscaled image lead to artefacts.
So I compute all variance/covariance/averages on the downscaled image (with only one 4-channels gaussian blur), I upscale these, and compute a and b on the full resolution image from them.

Performance on my computer:
without downscaling-upscaling (removing the calls to interpolate_bilinear) : 1.8s
with downscaling-upscaling, with scaling=4 : 0.9s

For this version, the time is spent in the following way:
2% for downscaling
1% for quantize (quantization is disabled)
11% for variance covariance and averages computation
56% for upscaling
30% for blending

Capture d’écran du 2020-10-17 23-06-56

rawfiner:
Last push just removes box-blur eigf, to compare more easily performance between eigf and original guided filter.
On same image as before, export takes between 0.8s and 3.9s depending on smoothing diameter value (takes 1.1s for smoothing diameter = 5%) with the original guided filter.

aurelienpierre:
No anisotropic weighting on this version ?

src/common/eigf.h
```diff
@@ -0,0 +1,229 @@
...
+  schedule(simd:static) aligned(image, av:64)
+#endif
+  for(size_t k = 0; k < Ndim; k++)
+  {
+    const float avg_g = av[k * 4];
+    const float avg_m = av[k * 4 + 2];
+    const float var_g = av[k * 4 + 1];
+    const float covar_mg = av[k * 4 + 3];
+    const float norm_g = fmaxf(avg_g * image[k], 1E-6);
+    const float norm_m = fmaxf(avg_m * image[k], 1E-6);
```
aurelienpierre, line 151: shouldn't this be `avg_m * mask[k]` ?

rawfiner: mask is computed on downscaled image for performance optimisation, so does not exist at full resolution, that's why I use image[k]

rawfiner:
> No anisotropic weighting on this version ?

When testing your last version, I noticed I got best results by using a very small sigma_weight (if you remember, I divided the value by 100), which is equivalent to disabling any blur depending on sigma_weight. So, I removed all the blurs of sigma_weight, which removed anisotropic weighted blur. But actually anisotropic weighting is not needed anymore as a and b are no more blurred.

src/common/eigf.h
rawfiner: But we can quantize on full resolution and downscale the mask if you prefer

aurelienpierre:
> So, I removed all the blurs of sigma_weight, which removed anisotropic weighted blur. But actually anisotropic weighting is not needed anymore as a and b are no more blurred.

Hmm the purpose of the anisotropic weight is to stick better to edges and get rid of the haloing at all, I fear that the haloing we got here earlier was connected to the downsampling, which is not used in the reference paper anyway. Setting sigma_weight ~= 0 disabled a, b and w blurs altogether, so it's hard to tell which one was responsible for the halos.

src/common/eigf.h
aurelienpierre: I think it would make sense to use the mask value here, just from a theoretical point of view.

aurelienpierre:
@rawfiner  here is my proposal with the anisotropic mask back on. It gives better blurring inside surfaces for large radii, without creating halos around edges AFAIK.
[eigf.tar.gz](https://github.com/darktable-org/darktable/files/5398122/eigf.tar.gz)

(no optimization so far)

aurelienpierre:
Also, not having a and b blurred behaves bad at high ISO (dark squares for the lowest intensity pixels).

rawfiner:
@aurelienpierre 
For noisy image, what I see is noise pixels that are not blurred, but no dark squares, do you have an example image where you have such?

Your patch is very similar to my proposition: you don't blur a and b. Your patch just multiplies each pixel of the output by a number close to 1 (your weights). It gives nearly identical results to what I had (including for noisy images). (and there is no anisotropic blurring, I guess you just forgot to put the blur on a and b at the end.

As for which blurring between a, b, and w gives problem, I am pretty sure it is blurring b that gives halos.
I will try to blur a weighted by w *before* computing b.

rawfiner:
Actually, even blurring a alone (with anisotropic weighting) reintroduce halos and less visually uniform blur:
Capture d’écran du 2020-10-19 08-44-19

src/common/eigf.h
rawfiner: I agree, I fixed that in my last push. The change in performance seems very small :-)

rawfiner:
I did a performance analysis, and figured out that toneequalizer spent as much time in compute_luminance_mask and in compute_correction.
41% of the time spent in compute_luminance_mask
40% of the time spent in compute_correction
18% of the time spent in apply_exposure

Looking at compute_correction code, it is particularly suited to use a lookup-table: inputs are bounded, their range is not wide, and we don't need a tremendous amount of precision.
Last commit makes the function use a lookup table, with a resolution of 10000 points per EV.

This gives approximately a 4x speedup to the function, and a 1.5x speedup to toneequalizer

aurelienpierre:
Regarding LUT, the alpha version of the tone EQ actually used one, but it turns out to be slower on some CPUs (including mine). I guess it's a matter of I/O bandwidth vs. clock speed. For GPU it is definitely slower.

rawfiner:
@aurelienpierre could you do a test again on your CPU?
I tested on my 2 computers (one i7 6th generation, and one (tiny) atom processor with 4 cores), I get a 4x speedup on both.
To test, in toneequalizer, I replace the code calling compute_correction by:
```
    clock_t begin = clock();
    compute_correction(luminance, correction, d->factors, d->smoothing, num_elem);
    clock_t end = clock();
    printf("compute corr: %lf\n", (double)(end - begin)/CLOCKS_PER_SEC);
```
and I compare printed times using the 2 different codes of compute_correction

Thank you :-)

rawfiner:
I duplicated compress shadows highlight presets, to have for each of them one preset with eigf and one with gf.
The presets with gf use the "flaw" that gf blurs more shadows than highlight in order to give better detail preservation in the shadows. eigf would perform better at preserving details in highlight than gf though, so I think it is good to let the user choose depending on their image.

rawfiner:
Added another performance optimisation: use different code path when quantization is disabled, to use the fact that guide == mask to get a speedup.
When using a high number of iterations (for instance 5), this gives a 1.5x speedup on my CPU. For 1 iteration only, not a lot of change.

rawfiner:
And I have nothing more on my todolist for this PR, so you can do another review please @aurelienpierre :-)

rawfiner:
I just realized that gaussian blur API authorized 2-channels gaussian blurs, so I simplified the code for the code path where mask == guide (I was using two 1-channel gaussian blur previously)

rawfiner:
Sorry, I couldn't resist at investigating where we could gain again some performance (I'll stop with this one).
By merging functions compute_correction and apply_exposure, we get an additionnal 1.2x speedup, thanks to the fact that the loop is executed only once.

All in all, execution time at export on my test images went from ~1.8s to ~0.38s, hence a 4.7x speedup :-)

aurelienpierre:
I can confirm that the LUT path is much faster. With your latest commit, at 5% radius, I'm down to 0.065-0.095 s instead of 0.230-0.250 s. Good job !

But the LUT could be stored in `d->LUT` and computed in `commit_params()` to spare a recomputation when the params have not changed between 2 runs.

aurelienpierre:
On another note, since the blurring of a and b has been removed, the feathering slider lower bound should be updated to something like 0.1 or so.

rawfiner:
Feathering lower bound is already 0.01 with right-clic, but I can change the soft bound to be 0.1 instead of 1

aurelienpierre:
That would be great. Just say so if you want me to move the LUT computation to `commits_params()`.

rawfiner:
I'll try to do it tomorrow morning before going to work, if I don't have the time I'll tell you, thank you Aurélien!

aurelienpierre:
Ok perfect. I will do the final review tomorrow then. Thanks to you, it's really great work and great added value for the tone EQ.

rawfiner:
Thank you Aurélien :-)

rawfiner:
So I had the time to change feathering soft bound and move the computation of correction LUT in commit params :-)

aurelienpierre:
Awesome, thanks !

rawfiner:
Thank you for the code reviews and all the help in designing and testing the filter!

PR #6445, opened 2020-10-04
[conf] hardening of dt_conf_get_* for no fun and no profit

johnny-bit:
This is a bit of code for #6408 so taht @dterrahe (and others) can get full advantage of predefined min/max values in darktableconfig.xml and never worry about "polluting" their code with calls to "sanitize" or "clamp" or anything like that.

After this PR it would be beneficial to actually check darktableconfig.xml to have min/max in more fields, just because it's safer that way and also it would be great if devs can check their code to see if they can use all apis presented with recent changes to conf.[ch]


src/control/conf.c
```diff
@@ -556,13 +589,37 @@ int dt_confgen_get_int(const char *name, dt_confgen_value_kind_t kind)
 {
   const char *str = dt_confgen_get(name, kind);
```
TurboGit, line 590: if kind does not exist str is NULL...

src/control/conf.c
```diff
@@ -556,13 +589,37 @@ int dt_confgen_get_int(const char *name, dt_confgen_value_kind_t kind)
...
   const float value = dt_calculator_solve(1, str);
```
TurboGit, line 591: do you confirm that this will return isnan() when str is NULL?

TurboGit: Maybe a better option is to explicitly check that the value exists, at least the intention will be clear when reading code.

So I would suggest calling dt_confgen_value_exists() and take proper action.

johnny-bit: `dt_calculator_solve` first lines is:

```c
if(formula == NULL || *formula == '\0') return NAN;
```

so I think It's fine to feed null into it. Do you still recomend to check with `dt_confgen_value_exists`?

TurboGit: Yes otherwise one will have to look at the implementation of dt_calculator_solve() to be sure code is ok, just reading this on this PR doesn't make this clear. So to me a clear and explicit check will ease maintenance.

johnny-bit: Done!

src/control/conf.c
```diff
@@ -574,8 +671,40 @@ gboolean dt_confgen_get_bool(const char *name, dt_confgen_value_kind_t kind)
...
+      break;
+    case DT_MAX:
+      return FLT_MAX;
+      break;
+    default:
+      return 0.0f;
+      break;
+    }
+  }
   const char *str = dt_confgen_get(name, kind);
```
TurboGit, line 690: At this stage we know it does not exist... Why retrieve str and call solve?

all returns:
`return isnan(value) ? -FLT_MAX : (value > 0 ? value + 0.5f : value - 0.5f);`
can just be:
`return -FLT_MAX;`

Just put an else here! Or am I missing something ?

src/control/conf.c
```diff
@@ -554,15 +589,77 @@ const char *dt_confgen_get(const char *name, dt_confgen_value_kind_t kind)
...
+      break;
+    case DT_MAX:
+      return INT64_MAX;
+      break;
+    default:
+      return 0;
+      break;
+    }
+  }
   const char *str = dt_confgen_get(name, kind);
```
TurboGit, line 646: see below

src/control/conf.c
TurboGit: Sorry!!!! I hadn't see the !  :)

TurboGit: So the code is perfect... Read too fast!

TurboGit (approved):
Thanks !

johnny-bit:
Yeah, sorry! int and int64 versions are correct due to possible rounding. float version is totally incorrect, please remove that +/- 0.5

src/control/conf.c
johnny-bit: Noooo! it's not perfect! float is wrong!!!

plz change to:

```c
float dt_confgen_get_float(const char *name, dt_confgen_value_kind_t kind)
{
  if(!dt_confgen_value_exists(name, kind))
  {
    //early bail
    switch (kind)
    {
    case DT_MIN:
      return -FLT_MAX;
      break;
    case DT_MAX:
      return FLT_MAX;
      break;
    default:
      break;
    }
    return 0.0f;
  }
  
  const char *str = dt_confgen_get(name, kind);

  //if str is NULL or empty, dt_calculator_solve will return NAN
  const float value = dt_calculator_solve(1, str);

  switch (kind)
  {
  case DT_MIN:
    // to anyone askig FLT_MIN is superclose to 0, not furthest value from 0 possible in float
    return isnan(value) ? -FLT_MAX : value;
    break;
  case DT_MAX:
    return isnan(value) ? FLT_MAX : value;
    break;
  default:
    break;
  }
  return isnan(value) ? 0.0f : value;
}
```

TurboGit: Indeed, missed that...

PR #6446, opened 2020-10-04
Updates to pt_BR

jpellegrini:


TurboGit:
That was fast :) I have just fixed some strings so you'll have a update the translation.

PR #6447, opened 2020-10-04
Improving monochrome images workflow

jenshannoschwalm:
Many of us enjoy monochrome images, only a few have a real mono camera.

So we develop in darkroom taking color-raws but have to remember what images were already meant to be developed as monochrome. Unfortunately there is no exif2 tag for this to be checked.

I tested many images shot with some sort of bw preset from different cameras and different brands, for all raw images tested the preview was grayscale. So i suggest to check for a "suggested" monochrome image workflow via testing the preview while reading the exif data, it is optionally switched on in preferences.

 1. The `ui/detect_mono_exif` config bool decides, whether the image preview is checked to be
   a monochrome while reading the exif data either while importing or when re-reading.
   This is optional as it slows down both.
2. In `dt_image_flags_t` we have three new bits defined.
   DT_IMAGE_MONOCHROME_PREVIEW will be set if there was a monochrome preview detected
   DT_IMAGE_MONOCHROME_BAYER is set if the passthrough monochrome demosaicer is used
   DT_IMAGE_MONOCHROME_WORKFLOW is right now just declared.
     - it will be used by modules that are mono-workflow-aware
     - it will be set by the user for a specific image like one of the color tags

  `int dt_image_monochrome_flags(const dt_image_t *img);` returns bit mask defining a monochrome image,
  `gboolean dt_image_use_monochrome_workflow(const dt_image_t *img);` is available for modules to test a monochrome workflow. It true if there is a monochrome image and DT_IMAGE_MONOCHROME_WORKFLOW is set.

3. In imageio we have two new functions
   `gboolean dt_imageio_has_mono_preview(const char *filename);`
   `void dt_imageio_update_monochrome_workflow_tag(int32_t id, int mask);` replacing the old bw_tag

4. In `dt_exif_read` we now look for true monochromes and also for previews if asked for in preferences.

5. The B&W info in lighttable uses all monochrome flags now.

6. In metadataview we also show the monochrome flag.

7. If the demosaicer uses one of the passthrough_monochrome modes the DT_IMAGE_MONOCHROME_BAYER flag should be set.

8. Also tests are done for true monochrome cameras.

9. If **any** of these 3 tests (preview, passthrough, true mono) it true the image gets a tag set `darktabe|mode|monochrome` so we can search for monochromes.

10. cacorrect and lens modules don't want to take corrections for passthrough_monochromes anyway, this is also checked there.

11. To allow modules to take care of special paths in a monochrome workflow there are some test functions already there

**Summary:** So far this pr allows us to check and search for images that are either true monochrome or intended to be developed in a monochrome workflow.

**ToDo:**
1. If we get along with this sort of workflow support i would add a switchable "activate workflow button" as we already have the color "flags" to enable the workflow in modules
2. Some modules might modify the processing path is a monochrome image is developed

### DIFF
--- a/data/darktableconfig.xml.in
+++ b/data/darktableconfig.xml.in
@@ -2325,7 +2325,14 @@
     <shortdescription>auto-apply sharpen</shortdescription>
     <longdescription>this added sharpen is not recommended on cameras without a low-pass filter. you should disable this option if you use one of those more recent cameras or sharpen your images with other means. (need a restart)</longdescription>
   </dtconfig>
-  <dtconfig>
+  <dtconfig prefs="processing">
+    <name>ui/detect_mono_exif</name>
+    <type>bool</type>
+    <default>false</default>
+    <shortdescription>detect monochrome previews</shortdescription>
+    <longdescription>many monochrome images can be identified via exif and preview data. beware: this slows down imports and reading exif data</longdescription>
+ </dtconfig>
+ <dtconfig>
     <name>screen_dpi_overwrite</name>
     <type>float</type>
     <default>-1.0</default>
--- a/src/common/exif.cc
+++ b/src/common/exif.cc
@@ -1223,8 +1223,10 @@ static bool _exif_decode_exif_data(dt_image_t *img, Exiv2::ExifData &exifData)
       dt_imageio_set_hdr_tag(img);
 
     if(is_monochrome)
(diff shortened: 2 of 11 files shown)


TurboGit (commented):
Looks ok to me, I'll do some tests.

TurboGit:
Works for me, thanks !

PR #6450, opened 2020-10-04
Improve appearance of credits in about window

amousset:
* Change credits background (also avoids superimposed text during credit animation)
* Properly split credits by section in AUTHORS file

credits


TurboGit:
Manually merged with many other fixes (bash, translations...). Thanks!

PR #6451, opened 2020-10-04
[imageio_tiff] faster half to float conversion

kmilos:
Improvement on https://github.com/darktable-org/darktable/pull/6230 with less branches and no loop

Taken from public domain https://gist.github.com/rygorous/2156668


TurboGit (approved):
Look ok, thanks !

PR #6453, opened 2020-10-05
clipping: adjust keystone control area size depending on the downsamp…

### LINKED ISSUE #6219: Crop - keystone correction: correction guide hard to adjust
On Ubuntu 20.04 with KDE, self-compiled cd8e1df2.

When **'reduce resolution of preview image'** (**preview_downsampling**) is set to **1/4**, I find the guides for adjusting keystone in the crop module very hard to adjust; there is **no problem when the setting is left at 'original'**.
I have to very accurately target the control point until the cursor switches into cross-hairs (from the 4-way arrows). I do not recall this behaviour with previous versions.

TurboGit:
…ling.

Fixes #6219.


GrahamByrnes:
A pleasure to use :-)

TurboGit:
Thanks! Merging then.

PR #6456, opened 2020-10-05
slightly improve squeezed fonts in colorpicker and blender

### LINKED ISSUE #6407: global color picker: tiny font
When reducing the width of the left panel the R/G/B labels become very small, even though the gap between them remains large -- so there's plenty of space for larger numbers.

Screenshot_2020-09-29_20-13-06

Please make the numbers bigger.

dterrahe:
fixes #6407

rather than reducing font size, when there is not enough space for all three values in the color picker module, this attempts to reduce font width only. This works for fonts that have (semi-) condensed versions installed on the system. This doesn't include the Monospace font on mine. The NK57 Monospace font works well. I can be downloaded here:
https://www.fontspring.com/fonts/typodermic/nk57-monospace
https://www.fontspring.com/fonts/typodermic/nk57-monospace-semi-condensed
https://www.fontspring.com/fonts/typodermic/nk57-monospace-condensed

and then after installation the following css tweak will activate it:
```
#live-sample-data
{
  font-family: NK57 Monospace;
}
```
Monospace:
image
NK57 Monospace:
image
Wider panel:
image

The other place where a lot of numbers get squeezed with a narrow panel is above the parametric blending gradient slider. Those numbers get ellipsized quite quickly and are then basically unreadable. Monospaced fonts make this worse, because they make everything wider, but with normal fonts the numbers all move around when one of them changes.

I've tried to improve this here by using an overlay widget, which allows assigning each number a fixed proportion of the width (so they don't jump around). When the panel is too narrow, they start to overlap and become unreadable. There is really no one-size fits all here, so unless we want to provide a lot of configuration options, letting the user specify their preferred fonts type (for example condensed) and size gives the most flexibility. The #blend-data name is now set for all number labels to make this easier. Their vertical alignment is set alternatingly to top and bottom, so by increasing the min-height, reducing the font-size, or a combination, they can be made to not clash when the panel width is reduced. I am using 
```
#blend-data
{
  font-size: 10px;
  min-height: 22px;
}
```
which looks like this:
image
without css tweaks it would look like this:
image
which isn't pretty, but previously it would have look something like this:
image
which is just as useless.

I think the changes here will allow more flexibility to finetune the display to personal preferences.

### DIFF
--- a/data/themes/darktable.css
+++ b/data/themes/darktable.css
@@ -522,10 +522,15 @@ overshoot.right
   margin-left: 2px
 }
 
-/* Adjust add button in picker module */
+/* Adjust add and remove buttons in picker module */
 .picker-module button
 {
-  padding: 0 3px;
+  padding: 0px;
+}
+
+.picker-module label
+{
+  padding-right: 3px;
 }
 
 /* Set history (darkroom) and recent collections (lighttable) modules, that make use of a
@@ -1675,7 +1680,7 @@ cell:selected
 
 #live-sample
 {
-  padding: 2px;
+  padding: 0px;
   min-height: 1em;
   min-width: 3em;
 }
--- a/src/develop/blend_gui.c
+++ b/src/develop/blend_gui.c
@@ -1579,19 +1579,28 @@ void dt_iop_gui_init_blendif(GtkBox *blendw, dt_iop_module_t *module)
       gtk_widget_set_tooltip_text(sl->polarity, _("toggle polarity. best seen by enabling 'display mask'"));
       gtk_box_pack_end(GTK_BOX(slider_box), GTK_WIDGET(sl->polarity), FALSE, FALSE, 0);
 
-      GtkWidget *label_box = gtk_box_new(GTK_ORIENTATION_HORIZONTAL, 0);
+      GtkWidget *label_box = gtk_grid_new();
+      gtk_grid_set_column_homogeneous(GTK_GRID(label_box), TRUE);
 
       sl->head = GTK_LABEL(dt_ui_label_new(in_out ? _("output") : _("input")));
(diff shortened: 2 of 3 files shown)


parafin:
There is no "Monospace" font, it's just an alias, that can be adjusted using fontconfig config files. Depending on what fonts are installed in the system, "Monospace" font might change.

data/themes/darktable.css
```diff
@@ -521,7 +521,7 @@ overshoot.right
 /* Adjust add button in picker module */
 .picker-module button
 {
-  padding: 0 3px;
+  padding: 0 0px;
```
Nilvus, line 524: It's worse when panel is not reduce too much. And having 3px as actually in master works with your PR AND when width is ok to see all information correctly.

Nilvus (changes requested):
I really think that caring about reduced panels is ok but not too far. It's normal to have some infos not so visible if we reduce the panel but it's better to take care of that and keeping a good rendering when panel is wider and show all infos. So, i think keeping some margin/padding is necessary.

data/themes/darktable.css
dterrahe: I didn't remove the left-right padding on the button to save space, but because it changes the shape from square to rectangular, which I felt didn't look as nice. I'd prefer to add the padding to the label with the numbers instead, if that is alright with you?

When the panel is wide, the padding has little effect because the label is centered in the available space, so there's additional empty space to the left and right anyway. So this really only impacts narrow panels, but if you think it's better, I don't mind.

Nilvus: On label it's indeed better. I didn't thought about button shape. Good catch.

dterrahe: Done! Thank you.

TurboGit (approved):
Works for me. Thanks!

PR #6461, opened 2020-10-05
iop_layouts : fix dialog creation as child of main window

AlicVB:
same for quick presets

@TurboGit : at the same time, in french translation I think that there's a typo : "disposition originelle" is better than "disposition originel"...


TurboGit (approved):
Thanks !

PR #6467, opened 2020-10-07
Fix typo in export.c causing init glitch

### LINKED ISSUE #6464: Second pixel output size is lost, becomes zero
dt 3.3.0+584~g5424fa2e3-dirty    Ubuntu 18.04

I've been noticing for a while that sometimes the second of the two output pixel dimensions in lighttable is zero, rather than the value I used previously.  The first value does not suffer in this way.

I'm able to reproduce this as follows -
1. Invoke dt via terminal window - $HOME/DT-Mast-8Sep20/bin/darktable --cachedir $HOME/.cache/darktable-8sep --configdir $HOME/.config/darktable-8sep
2. See second value is zero.  Set it to 3600.
3. Close dt by clicking the "x" top right.
4. Invoke dt as (1)
5. See zero again.

junkyardsparkle:
Fixes #6464

### DIFF
--- a/src/libs/export.c
+++ b/src/libs/export.c
@@ -383,7 +383,7 @@ static void _validate_dimensions(dt_lib_export_t *d)
   uint32_t height = atoi(gtk_entry_get_text(GTK_ENTRY(d->height)));
   if(width > d->max_allowed_width || height > d->max_allowed_height)
   {
-    width = width > d->max_allowed_width ? dt_conf_get_int(CONFIG_PREFIX "width"): width;
+    width = width > d->max_allowed_width ? dt_conf_get_int(CONFIG_PREFIX "width") : width;
     height = height > d->max_allowed_height ? dt_conf_get_int(CONFIG_PREFIX "height") : height;
     _set_dimensions(d, width, height);
   }


src/libs/export.c
```diff
@@ -383,7 +383,7 @@ static void _validate_dimensions(dt_lib_export_t *d)
   uint32_t height = atoi(gtk_entry_get_text(GTK_ENTRY(d->height)));
   if(width > d->max_allowed_width || height > d->max_allowed_height)
   {
-    width = width > d->max_allowed_width ? dt_conf_get_int(CONFIG_PREFIX "width"): width;
+    width = width > d->max_allowed_width ? dt_conf_get_int(CONFIG_PREFIX "width") : width;
```
TurboGit, line 386: Is a single space before `:` really changing something ?

junkyardsparkle: It seems to correct the behavior in #6464 for me... go figure. :-)

junkyardsparkle: Ok, I just tried a few more quit/start cycles, and in one case still had the reset-to-zero problem happen, so... maybe that wasn't the cause? If that's the case, sorry for the noise...

TurboGit: I have never seen that and I've been working quite a lot on the export module recently (see print-size PR). Anyway, that's a reformatting, so I'll merge :)

junkyardsparkle: Ok, but it closed the issue, which should probably still be open... sorry! I'm really curious about what the actual cause is, but I need to go to sleep before I make any more dumb assumptions.

RawConvert:
The fix seems it was addressing the width, but it is the height (2nd value) that is playing up, surely?

junkyardsparkle:
@RawConvert Yes, I was deluded. :-/

It was the only odd thing I found, and after fixing it the issue vanished... until later. I didn't realize it was that kind of intermittent bug, maybe a race condition somewhere.

Issue #6363, opened 2020-09-23
feature Import from camera : allow import of a single folder


Coqorico:
**Is your feature request related to a problem? Please describe.**

I have a Canon camera which allows the creation of folders on the CF or SD card. When I import from the card in darktable using the import from a camera feature, all the pictures on the card are scanned and displayed in a single list. I have to wait for the scan of the whole card to be finished before I can import any picture.
When there are multiple folders to import, using different job names, this requires the scan of the whole card each time.

**Describe the solution you'd like**

It would be great to:
1 (shall) - Allow selecting and browsing a single folder of the camera or card, then scan only the selected folder to reduce the time required before import is possible.
2 (should) - Allow importing a whole folder without waiting for the card or the selected folder to be scanned. A preview of the content of each folder content could be provided by scanning and showing the first picture only.
3 (nice to have) - Allow importing the whole folder currently browsed in a click (on a dedicated button), without using ctrl-A then import.

**Alternatives**

When importing from a local disk folder, allow to copy / move raw files in a folder created by darktable, same as import from camera.
If this is made possible, the content of the card could be copied in a temporary local folder before it is imported with darktable. This would let darktable create folders and rename files and avoid the use of a third-party software to handle that before pictures are imported in darktable.

**Additional context**

Nilvus:
This one could be related to my old FR: #2762

Coqorico:
Hello Nilvus,
Indeed, it could be related and packaged together. Our issues are complementary as they relate to the ergonomics and features of the import from camera functionality.

parafin:
Just use card-reader or/and some other software (file manager, rapid photo downloader) instead of darktable's import from device functionality. Really, only import from folder should be used in darktable IMHO.

Coqorico:
I used import from folder, but it requires to use another tool to import the raw on the hard drive in order to generate the folders and rename the imported files.
I didn't know rapid photo dowldader, I'm having a look and it seems perfect, thanks for the advice! This seems to fulfill my needs.

Nilvus:
> Really, only import from folder should be used in darktable IMHO.

Why not use it? Import from SD card works great. We just talk about how to improve it.

parafin:
[...]

Issue #3942, opened 2019-12-30
Build warning in master for libavif


chhil:
This is probably related to the new AVIF support.

Building clean master get the following warning.

CMake Warning at src/CMakeLists.txt:324 (find_package):
  By not providing "Findlibavif.cmake" in CMAKE_MODULE_PATH this project has
  asked CMake to find a package configuration file provided by "libavif", but
  CMake did not find one.

  Could not find a package configuration file provided by "libavif"
  (requested version 0.5.2) with any of the following names:

    libavifConfig.cmake
    libavif-config.cmake

  Add the installation prefix of "libavif" to CMAKE_PREFIX_PATH or set
  "libavif_DIR" to a directory containing one of the above files.  If
  "libavif" provides a separate development package or SDK, be sure it has
  been installed.

TurboGit:
Should be fixed.

chhil:
Fixed via https://github.com/darktable-org/darktable/pull/3947

Issue #3924, opened 2019-12-29
Importing not working in DT 3.0


muenznej:
Hello,
After the update from 2.6 to 3.0 i cannot import images via cable from my Olympus E-M 10 Mark II. The images are "imported" but corrupted. The focal length is f/0.0 and no preview is available. I only have raws to import.They can be previewed with the "import images from camera" gui. 
![image](https://user-images.githubusercontent.com/26899110/71556737-545ce580-2a3c-11ea-8972-d0434719be21.png)
![dt3_error_imprt](https://user-images.githubusercontent.com/26899110/71556739-62ab0180-2a3c-11ea-9763-c2b4834053ba.PNG)
![image](https://user-images.githubusercontent.com/26899110/71556769-bd445d80-2a3c-11ea-80d9-263f7a55a8e0.png)

The files are not readable by other tools after importing.  The files are not corrupted on the sdcard.

Thanks for the wonderful DT and hopefully for a quick solution :)

**Platform (please complete the following information):**
 - Darktable Version: [3.00] 
 - OS: [Windows 8.1]
 - OpenCL activated
 - HD5500 and geforce 940m

MStraeten:
What happens if you import the files after copying themn manually to your disc? Just to reduce the process chain, where errors might occur

muenznej:
importing via the direct camera connection is the only problem. I copied all files to a folder manually and imported that folder which worked out fine.

Lucki:
I can confirm this issue. While direct wire importing worked in Darktable 2.6 it doesn't work anymore in version 3.0. Manually copying the pictures to the hard drive and importing them as a folder still works normally.

In the import dialog Darktable is able to see all pictures. After pressing import Darktable processes all of the selected pictures but every single one is corrupted and displayed as a skull and don't have any correct meta data. Closing and reopening Darktable automatically removes all images but one from the set.
All files and the corresponding `*.xmp` files are created on the hard drive and the file sizes _looks_ normal. However I couldn't verify if they're actually corrupted.

* Darktable 3.0.0
* Windows 10
* Nikon D3500 through [WinUSB](https://www.darktable.org/about/faq/#faq-tethering)
* Radeon Vega 8 and GeForce GT 710

Lucki:
A bit more information gathered through testing:
* The problem was introduced between [2.6.2 and 2.6.3](https://github.com/darktable-org/darktable/compare/release-2.6.2...release-2.6.3)
[...]

PR #6474, opened 2020-10-07
 Try to fix sdcard imports

jenshannoschwalm:
At least sdcard imports on windows machines seem to be pretty unreliable.

Related Issues #3924 #6363 and probably others hinting to something specific for windows build setups.
So checking in src/common/camera_control.c for `dt_camctl_import`

1) In the data copy code we use `g_open` and `write`. Reading the docs for g_open() at
   https://developer.gnome.org/glib/stable/glib-File-Utilities.html#g-open
   suggests there might be problems when combined with `write` depending on the build
   process.
   Unfortunately i have no idea about the windows build stuff so i can't check that
   but using the much shorter code using `g_file_set_contents` seems to be a safer option.

   The other code sections where @houz changed open to g_open for good reasons (commit ad7350b157f2458f9a34a076045a53bc19fbb353) are ok  because we later use `gp_file_new_from_fd` which seems not to be affected by the gphoto glitches.

2) Remembering the sometimes wrong data passed after `gp_file_new` (see #5333 for details)
   that could explain the "empty / wrong" data written to files reported in #3924

3) The `active_camera->port` can be tested by a simple strncmp for being a disc/sdcard.
   In that case we don't want to use the gphoto2 reading but prefer to use some sort
   of file-copy (including `dt_exif_get_datetime_taken`), as we have a proper filename
   available we can read the data via `g_file_get_contents` and write to the target
   file as described above.

   Hopefully fixing #3942

### DIFF
--- a/src/common/camera_control.c
+++ b/src/common/camera_control.c
@@ -1009,9 +1009,7 @@ void dt_camctl_import(const dt_camctl_t *c, const dt_camera_t *cam, GList *image
 {
   _camctl_lock(c, cam);
   const gboolean sdcard = !strncmp(c->active_camera->port, "disk:", 5);
-  if(sdcard)
-    dt_print(DT_DEBUG_CAMCTL, "[dt_camctl_import] started an import session from `%s`\n", c->active_camera->port);
-
+    
   GList *ifile = g_list_first(images);
   if(ifile) do
   {
--- a/src/common/camera_control.c
+++ b/src/common/camera_control.c
@@ -1007,96 +1007,83 @@ static gboolean _camera_initialize(const dt_camctl_t *c, dt_camera_t *cam)
 
 void dt_camctl_import(const dt_camctl_t *c, const dt_camera_t *cam, GList *images)
 {
-  // dt_camctl_t *camctl=(dt_camctl_t *)c;
   _camctl_lock(c, cam);
+  const gboolean sdcard = !strncmp(c->active_camera->port, "disk:", 5);
+  if(sdcard)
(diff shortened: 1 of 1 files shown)


AlicVB:
>     3\. The `active_camera->port` can be tested by a simple strncmp for being a disc/sdcard.
>         In that case we don't want to use the gphoto2 reading but prefer to use some sort
>         of file-copy (including `dt_exif_get_datetime_taken`), as we have a proper filename
>         available we can read the data via `g_file_get_contents` and write to the target
>         file as described above.

That means that if we use this with an sdcard mounted on the filesystem, we don't use gphoto anymore, except for the starting check ? That would be great, as I'm really more confident in "classic" file transfert :)
I even wonder... Do you think it would be feasible to allow this without gphoto at all ?

jenshannoschwalm:
> That means that if we use this with an sdcard mounted on the filesystem, we don't use gphoto anymore   ....

We still use it for more atm like for fetching the previews and taking care of connected devices. But that is all handled in gphoto correctly. But - yes - the file is copied with "normal" file-copy code.

> Do you think it would be feasible to allow this without gphoto at all?

Yes :-) But that would mean much more code to be changed. ATM i only hope to fix the open issue and let import do what it should do :-) 

There has been the discussion in #6363 about a one-in-all solution. Personally i am not sure about that idea, there might be so many user options to handle, many personal workflow filtering, i simply haven't thought about what would be needed.

TurboGit (approved):
Works for me! Hopefully this will fix issues on Windows as I had never had issues on GNU/Linux importing from sdcard. Thanks.

PR #6480, opened 2020-10-08
map : don't move map view on undo/redo

phweyland:
When one uses undo/redo on a resized or moved map view, undo/redo centers again the map view on collection, which is not user friendly (lost of the current focus).

This PR avoids this inconvenience.

Additionally
- remove unused declaration
- fix standard internal routine naming
- set thumb overlay back to a low value (seems to have been lost in the related commit ([5012b60](https://github.com/darktable-org/darktable/pull/6303/commits/5012b60a6a4732e7f83b3ce4de7bf2e845b211b7)))

### DIFF
--- a/src/views/map.c
+++ b/src/views/map.c
@@ -38,13 +38,6 @@
 
 DT_MODULE(1)
 
-typedef struct dt_undo_geotag_t
-{
-  int imgid;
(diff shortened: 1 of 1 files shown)


TurboGit (approved):
Works for me, thanks !

PR #6489, opened 2020-10-09
po/module groups

TurboGit:


TurboGit:
I have tried to come-up with three presets for the new module groups designed by @AlicVB. I have removed the default and legacy presets in the hope to start clear with dt 3.4. A preset will never fit everybody needs and is a base for working out our own presets which is easy using new module-groups GUI.

The three presets proposed are:
- all modules
- scene-referred
- display-referred

The later are using new module instead of "legacy one". That is I have favored the curvergb over the tonecure, etc...

I have tried to keep the presets simple and this is based on my experience on using dt since years. There is many modules that I don't see a benefit for new comers, I'm not saying that those are not useful but just that you do not need to do 99% of the development work, so better not having them by default.

That being said all this is very delicate and will certainly generate lot of discussions... why are highly welcomed :)

So let's start... What do you think ?

src/libs/modulegroups.c
```diff
@@ -907,6 +907,7 @@ static void _preset_retrieve_old_presets(dt_lib_module_t *self)
 
     gchar *tx = _preset_retrieve_old_layout(list, fav);
     dt_lib_presets_add(pname, self->plugin_name, self->version(), tx, strlen(tx), FALSE);
+    g_free(tx);
```
AlicVB, line 910: for some reason, freeing the string was crashing dt at some point. That's why I've not done that. But apparently this is no more the case :) maybe the problem was elsewhere (or something as changed in presets code)

src/libs/modulegroups.c
```diff
@@ -975,32 +976,67 @@ static GList *_preset_from_string(gchar *txt)
...
 void init_presets(dt_lib_module_t *self)
 {
+  // all modules
   gchar *tx = NULL;
-  tx = dt_util_dstrcat(
-      tx, "ꬹ1ꬹ%s|%s||%s", _("technical"), "technical",
-      "colorin|hazeremoval|filmicrgb|clipping|flip|lens|exposure|denoiseprofile|demosaic|highlights");
-  tx = dt_util_dstrcat(tx, "ꬹ%s|%s||%s", _("grading"), "grading", "rgblevels|colorbalance|toneequal|temperature");
-  tx = dt_util_dstrcat(tx, "ꬹ%s|%s||%s", _("effects"), "effect", "sharpen|bilat");
-  dt_lib_presets_add(_("default"), self->plugin_name, self->version(), tx, strlen(tx), TRUE);
```
AlicVB, line 984: The "default" preset is use in many place in the code as a fallback in case no valid preset is selected (first start, deletion of current preset, etc...) so we have to decide which is the new "default" preset and change the code in that places too.

AlicVB (changes requested):
For the code, we just need to ensure the fallback preset, otherwise no problem. For the presets content, I'll post my suggestions in a different message

AlicVB:
That work is really needed ! Thanks you very much for taking care of it !
As said earlier, you are certainly way more qualified than me here ;)

Some quick questions...
1. So you seems to have abandoned the "3 groups" layout... Is there any reason for that (I quite like @aurelienpierre reasoning which is behind it, in #5322 for memories )
2. I won't have put so many modules in "default" presets, but you have way more experience with what need new comers, So I won't argue too much here...
3. Is there any reason why chanel mixer is not present in "default" presets ? (I use it quite often)
4. For what I've read in different materials around (esp. @aurelienpierre comments and articles) I'm not sure it's wise to advertise monochrome, velvia, color correction, ... esp. in scene referred preset
5. maybe we should propose a "minimal" preset too, this would be handy as a start base or for a combination of lightweight groups + search

again, that's just questions. Don't take it as criticism :)

jenshannoschwalm:
> Is there any reason why chanel mixer is not present in "default" presets ? (I use it quite often)

Me too. Or is there still hope for #4940 :-)

> maybe we should propose a "minimal" preset too ...

Why not **call** it "fresh user tools" or alike?

TurboGit:
> Me too. Or is there still hope for #4940 :-)

Exactly, I expect the new channel mixer to be in 3.4 and part of the presets.

TurboGit:
> So you seems to have abandoned the "3 groups" layout...

No, forgot about this. Let me revisit this proposal and propose a preset based on it.

TurboGit:
> I won't have put so many modules in "default" presets, but you have way more experience with what need new comers, So I won't argue too much here...

Maybe we need a "default" preset ? Mine are real preset to be usable in 90% of dev.

TurboGit:
> maybe we should propose a "minimal" preset too, this would be handy as a start base or for a combination of lightweight groups + search

Agreed, we still need to define what should be in minimal. My needs are high and so I need many tools, what could be useful for beginners, that's the real question and I'm not sure we will have an easy consensus :)

TurboGit:
@AlicVB : thanks for the feedback, no problem and I don't take this as criticisms, as I said and anticipated this will raise many comments and we will have hard time to accommodate for every needs so discussions are really welcomed.

src/libs/modulegroups.c
TurboGit: Maybe we should default to "scene referred" preset ? Maybe using a #define to specify the default preset. Do you have a list of all places in code where we reference it ?

TurboGit:
@AlicVB : default preset support now restored. The new default is the 3 tabs based on Aurélien's proposal in #5322.

TurboGit:
At this stage we have 4 presets:
- default (3 tabs based on Aurélien's proposal in #5322)
- all modules
- display referred
- scene referred

I'd like have no more than 4 dt presets. Seems good enough, but we can improve the definition of them.

TurboGit:
> Why not **call** it "fresh user tools" or alike?

The presets display-referred or scene-referred are really for starters. Those have not a lot of modules. Do you really think we need as even more trimmed version for beginners ? I fear to go too far...

jenshannoschwalm:
> The presets display-referred or scene-referred are really for starters

Yes. True. My comment was meant as a suggestion for starters like "Look here and see how you get along with these", it should be made clear on first sight.

TurboGit:
@jenshannoschwalm : thanks for the feedback...let me add one of the point I'd like all of us to agree on...

Currently when one user is lost with dt the answer is:

"don't use all modules, add some in the favorites (exposure, white balance, crop&rotate...) play with them and add more modules when you are fluent with those."

One of the goal here is to be able to say for dt 3.4:

"don't use all modules, as a beginner start with preset "bla bla bla" and when you are fluent with those modules just create a new preset based on this one and add some more modules".

My question is what "bla bla bla" preset should look like ? I'm currently ok with "display referred" one as it is easier to grasp than "scene referred one" and because the current default workflow is display-referred. But I'm really curious to see if this is a common agreement or just some fantasy on my side :)

jenshannoschwalm:
I guess @aurelienpierre might have a different opinion on the suggested workflow though.

I can agree with your suggestion. But: have a look on the number of beginners issues. What you propose is also some sort of Mini-FAQ to be read **before** getting lost. But where and how to present as a reminder? A long and extensive tooltip on the module-section buttons? Adding a "beginner" checkbox defaulting to true disabling some of the fancy things?

TurboGit:
> But: have a look on the number of beginners issues. What you propose is also some sort of Mini-FAQ to be read **before** getting lost. But where and how to present as a reminder?

Not sure, maybe of the documentation that no one reads :) Or possibly making the "beginner" presets the default ? All is possible, as at this time I'm trying to devise the best option for beginners and standard users.

Nilvus:
Just a thing there: we need to kept previous layout set by user before iop layout feature and especially favorites things. So maybe had a preset with legacy and new default one with favorites feature (or a button to just add favorites feature. See my 2 comments about that on @AlicVB PR: https://github.com/darktable-org/darktable/pull/5683#issuecomment-695781807

aurelienpierre:
I don't understand the difference between this feature and the workspace presets we have.

TurboGit:
@aurelienpierre : this PR only add some more default presets. The workspace presets where done on the modulelist module which has been removed and superseded by the modulegroup.

aurelienpierre:
Ok so maybe it's a good idea to port the workspaces concept (landscape/lowlight/portrait/etc.) to this new method.

TurboGit:
@aurelienpierre : I'm not convinced this is needed now. The new modulegroup can be design as needed by anyone. Group order and content. So to me we should just have some simple defaults plus the all module preset and let users create whatever preset they want.

AlicVB:
@Nilvus : except if I have missed something, this PR doesn't remove the "old layout retrieval". That means, user will still have the 2 "my previous layout ..." presets when upgrading from older versions.
The question is more which other "built-in" preset we want.

Nilvus:
> @Nilvus : except if I have missed something, this PR doesn't remove the "old layout retrieval". That means, user will still have the 2 "my previous layout ..." presets when upgrading from older versions.
> The question is more which other "built-in" preset we want.

I didn't test it so if they're still there, it's good.

TurboGit:
I have now added a minimal preset. The scene referred has also been stripped of some "toxic" modules based on Aurélien's list. At this stage I think this PR is ok and will merge it later today if there is no more comment. We can still improve it but having more field testing will help refine if needed.

PR #6496, opened 2020-10-10
Monochrome workflow supporting presets

jenshannoschwalm:
As we now have detection of monochrome images the next obvious step is to
make presets to be aware of monochrome images.

On the gui side: whenever we make/modify a preset we had flags for normal, raw and hdr
images. **Added** are flags for `monochrome` and `color` so as an example you can create specific presets for monochrome raw and can auto apply them.

In `gui/presets.h` you find the bitmask extended
```
...
  FOR_NOT_MONO = 1 << 3,
  FOR_NOT_COLOR = 1 << 4
} dt_gui_presets_format_flag_t;

```
Please note the new bits are used as excluded, the gui side takes care of this.
Implementing it this way means no hassle with existing presets.


jenshannoschwalm:
BTW looks like this
Bildschirmfoto von 2020-10-10 15-51-26

TurboGit (approved):
Works for me, thanks !

Can you propose a change for the manual ? I'll update the screen shot. TIA.

PR #6501, opened 2020-10-11
Update preferences manual for performance mode and monochrome previews

jenshannoschwalm:
1. Image and description should reflect the new `performance mode`

2. Add docs about `detect monochrome previews`


TurboGit (changes requested):
Please ensure the comment in the CSS Window is in English ! 

Should be:

/* Enter CSS theme tweaks here */

Thanks.

jenshannoschwalm:
Aah, i missed that. It's probably a bug as dt preferences interface language is set to english.

TurboGit:
Not a bug really. This gets written the first time dt is started into the user's CSS and after that is is not translated as it is only the content of the file which is displayed.

jenshannoschwalm:
uploaded.

TurboGit (approved):
Thanks !

PR #6510, opened 2020-10-12
no functional change. Makes CMake more consistent

codingdave:
This PR does not change any effect. It only cleans up a tiny bit:

Usually the order is AUTHORS authors_h.sh, just this line is the other way around. Fixing the order makes it more quickly visible that nothing different happens then on the other lines.


TurboGit (approved):
Looks ok to me !

PR #6520, opened 2020-10-13
po/safe history copy

TurboGit:


jenshannoschwalm:
1. What about "white balance" - many issues about it :-)
2. Not sure about lens correction although very hardware related.

Otherwise -- cool!

parafin:
Also probably flip and colorin, definitely rotatepixels and scalepixels

maboleth:
Thumbs up! Step in the right direction IMO.

junkyardsparkle:
I can personally vouch for at least one workflow where it's assumed that white balance and colorin are included in a "copy/paste all" operation. Maybe I'm just weird, but it might be good to get input from a larger group of users about those... remember that unlike styles. copy/paste is something people are quite likely to be using in an *ad hoc* manner for a group of closely related photos. The situation with rawprepare and auto-ISO seems like something people are likely to not think about in this situation (I'll admit that I didn't, previously), but if adjustments have been made to colorin or WB, it seems like not including them might be an unexpected change in behavior.

jenshannoschwalm:
>  but if adjustments have been made to colorin or WB, it seems like not including them might be an unexpected change in behavior.

I would say i would not lead to unexpected results :-) Unfortunately modules and image structs have no "hardware dongle" to allow checking for "copy&paste on same hardware".

junkyardsparkle:
> I would say i would not lead to unexpected results :-)

Well, I know for a fact that it would for *me*, so I guess that it might for at least some other people. I'm just suggesting that before including too many other modules in this purge, it might be good to take a quick survey at pixls.us or somewhere about actual usage of "copy/paste all". Remember that it's a feature that some people have been using for a loooooong time. :-)

TurboGit:
@jenshannoschwalm : lens correction module has been implemented to always get the information from the image (camera model & lens) even when copying. So to me it should be kept in the copy/paste.

TurboGit:
I have added scalepixel, rotatepixel, whitebalance & flip. The good news is that with this change I'll be able to use more copy/paste that I do not use to avoid those issues, most of the time I do a "copy parts" and then paste.

@parafin : Not clear to me why colorin should be removed ?

parafin:
I was thinking about copying from JPEG to RAW.

TurboGit:
I see ! I would keep this away from the equation for now, copying from JPEG to RAW or reverse is a corner case that most of the time doesn't work well anyway.

jenshannoschwalm:
> lens correction module has been implemented to always get the information from the image

Yeah. Have you tried to auto-apply a preset to the lens module  for camera A. Then copy&paste to camera B. Surprise ... but maybe this should be fixed elsewhere ...

TurboGit:
> Yeah. Have you tried to auto-apply a preset to the lens module for camera A. Then copy&paste to camera B. Surprise ... but maybe this should be fixed elsewhere ...

No, just did on two pictures now and saw nothing wrong but if this is causing trouble in many cases then I'm also ok to remove it from "copy all".

TurboGit:
I have now added lens, let's have "copy all" on the safe side when possible to make it more usable.

parafin:
To me the correct behaviour of "copy all" would be not to include any auto-added iops (with original params that is). That was how it worked before these auto-applied modules got included in the history stack, so I think it was the intended behaviour of this feature. Implementing it now though seems quite hard technically. Approach from this PR seems like a good compromise. List of the iops seems OK too - these are the technical modules correcting the picture to get correct data and not related to further processing.

TurboGit:
Thanks all for the valuable comments. I'll merge and see how this flies!

PR #6521, opened 2020-10-13
Fix popover padding

amousset:
before
to
afetr

### DIFF
--- a/data/themes/darktable.css
+++ b/data/themes/darktable.css
@@ -871,7 +871,13 @@ dialog combobox window
   padding: 0.2em 0;
 }
 
-menu menuitem {
+popover
+{
+  padding: 0.4em;
+}
+
+menu menuitem
+{
   padding: 0 0.2em;
 }
 


data/themes/darktable.css
```diff
@@ -871,7 +870,16 @@ dialog combobox window
...
 }
 
-menu menuitem {
+popover
+{
+  background-color: @tooltip_bg_color;
+  border-radius: 2px;
+  border: 0;
+  padding: 0.4em;
+}
```
Nilvus, line 879: Please put back popover in line 860. That has been done that way to avoid some glitches in some desktop environment. Just add a specific popover part has you done for only padding specific size. I remember anyway that this padding was good quite recently and I'm quite surprise to see that it has been broken. Anyway, thanks for fixing that.

Nilvus (changes requested):

data/themes/darktable.css
amousset: Thanks for your feedback.

Nilvus (approved):

TurboGit (approved):
Thanks, this has bothered me for some time now :) Good to see this fixed!

Issue #6487, opened 2020-10-09
Windows Build Issue


chhil:
I set up the build environment from scratch.

Couple of things that were different

SAXON :
The requirement to download it and place it and set classpath, uses saxon9he.jar but thats no longer available and I used SaxonHE10-2J.zip , extracted the files correctly and updated classpath to use saxon-he-10.2.jar.

Older version of lensfun is no longer available so I went with what got installed and did the lensfun update.

-----
While trying to build I got a 
```
-- The following OPTIONAL packages have not been found:

 * Saxon


```
Then later I got the error

```
[  6%] Checking validity of data/darktableconfig.xml
[  6%] Built target validate_darktableconfig_xml
[  6%] Built target generate_conf
[  6%] Built target generate_preferences
[  6%] Built target validate-cameras.xml
[ 18%] Built target rawspeed
[ 18%] Building C object src/CMakeFiles/lib_darktable.dir/common/camera_control.c.obj
C:/msys64/home/chill/darktable/src/common/camera_control.c: In function 'dt_camctl_import':
C:/msys64/home/chill/darktable/src/common/camera_control.c:1044:73: error: passing argument 3 of 'gp_file_get_data_and_size' from incompatible pointer type [-Werror=incompatible-pointer-types]
 1044 |       if((res = gp_file_get_data_and_size(camfile, (const char**)&data, &size)) < GP_OK)
      |                                                                         ^~~~~
      |                                                                         |
      |                                                                         size_t * {aka long long unsigned int *}
In file included from C:/msys64/mingw64/include/gphoto2/gphoto2.h:50,
                 from C:/msys64/home/chill/darktable/src/common/camera_control.h:30,
                 from C:/msys64/home/chill/darktable/src/common/camera_control.c:21:
C:/msys64/mingw64/include/gphoto2/gphoto2-file.h:136:30: note: expected 'long unsigned int *' but argument is of type 'size_t *' {aka 'long long unsigned int *'}
  136 |           unsigned long int *size);
      |           ~~~~~~~~~~~~~~~~~~~^~~~
cc1.exe: all warnings being treated as errors
make[2]: *** [src/CMakeFiles/lib_darktable.dir/build.make:1831: src/CMakeFiles/lib_darktable.dir/common/camera_control.c.obj] Error 1
make[1]: *** [CMakeFiles/Makefile2:2331: src/CMakeFiles/lib_darktable.dir/all] Error 2
make: *** [Makefile:172: all] Error 2
```

[...]

Issue #6459, opened 2020-10-05
Build on Windows fails 3.3.0+996


juhavlintula:
**Describe the bug**
Build on Windows fails
```
...
Scanning dependencies of target generate_authors_h
[ 54%] Generating authors.h for about dialog.
%1 is not a valid Win32 application
make[2]: *** [src/libs/CMakeFiles/generate_authors_h.dir/build.make:85: src/libs/tools/darktable_authors.h] Error 1
make[1]: *** [CMakeFiles/Makefile2:3520: src/libs/CMakeFiles/generate_authors_h.dir/all] Error 2
make: *** [Makefile:172: all] Error 2
```

**To Reproduce**
1. `make --build .`
2. See error

**Platform (please complete the following information):**
 - Darktable Version: 3.3.0+996
 - OS: Windows10

**Additional context**
I removed build directory and recreated it, but it didn't help.

chhil:
Reproducible
Version string: 3.3.0+997~gf6465132a
Updating 7340b7b55..24a8060de

Probably related to changes made on Oct 5 : ad1760b715161b26a7bfd746ba697f324b4c7b74 or 7f84b15744b06cfc7b94ae770d484722b49ed0e7

TurboGit:
Can one of you try this patch:

```
diff --git a/src/libs/CMakeLists.txt b/src/libs/CMakeLists.txt
index 9565d07c6..b696899fa 100644
--- a/src/libs/CMakeLists.txt
+++ b/src/libs/CMakeLists.txt
@@ -9,7 +9,7 @@ add_custom_command(
   DEPENDS ${CMAKE_SOURCE_DIR}/AUTHORS ${CMAKE_SOURCE_DIR}/tools/authors_h.sh
   OUTPUT ${CMAKE_CURRENT_BINARY_DIR}/tools/darktable_authors.h
   COMMAND ${CMAKE_COMMAND} -E make_directory ${CMAKE_CURRENT_BINARY_DIR}/tools/
-  COMMAND ${CMAKE_COMMAND} -E env LANG=C LC_ALL=C ${CMAKE_SOURCE_DIR}/tools/authors_h.sh ${CMAKE_SOURCE_DIR}/AUTHORS ${CMAKE_CURRENT_BINARY_DIR}/tools/darktable_authors.h
[...]

PR #6532, opened 2020-10-15
Fix/windows msys build

codingdave:
Having the latest MSYS2 environment and Pacman updated without this change CMake does fail when trying to invoke a bash shell to call authors_h.sh. This is a workaround for compiling locally from @dterrahe. However, he thinks this might be not correct upstream. 
@dterrahe is it because we cannot guarantee that /usr/bin/bash is correct for all systems? What about removing `bash` and let shebang handle finding the correct interpreter?

### DIFF
--- a/src/libs/CMakeLists.txt
+++ b/src/libs/CMakeLists.txt
@@ -9,7 +9,7 @@ add_custom_command(
   DEPENDS ${CMAKE_SOURCE_DIR}/tools/authors_h.sh ${CMAKE_SOURCE_DIR}/AUTHORS
   OUTPUT ${CMAKE_CURRENT_BINARY_DIR}/tools/darktable_authors.h
   COMMAND ${CMAKE_COMMAND} -E make_directory ${CMAKE_CURRENT_BINARY_DIR}/tools/
-  COMMAND ${CMAKE_COMMAND} -E env LANG=C LC_ALL=C bash ${CMAKE_SOURCE_DIR}/tools/authors_h.sh ${CMAKE_SOURCE_DIR}/AUTHORS ${CMAKE_CURRENT_BINARY_DIR}/tools/darktable_authors.h
+  COMMAND ${CMAKE_COMMAND} -E env LANG=C LC_ALL=C /usr/bin/bash ${CMAKE_SOURCE_DIR}/tools/authors_h.sh ${CMAKE_SOURCE_DIR}/AUTHORS ${CMAKE_CURRENT_BINARY_DIR}/tools/darktable_authors.h
   COMMENT "Generating authors.h for about dialog."
 )
 
--- a/src/libs/CMakeLists.txt
+++ b/src/libs/CMakeLists.txt
@@ -9,7 +9,7 @@ add_custom_command(
   DEPENDS ${CMAKE_SOURCE_DIR}/tools/authors_h.sh ${CMAKE_SOURCE_DIR}/AUTHORS
   OUTPUT ${CMAKE_CURRENT_BINARY_DIR}/tools/darktable_authors.h
   COMMAND ${CMAKE_COMMAND} -E make_directory ${CMAKE_CURRENT_BINARY_DIR}/tools/
-  COMMAND ${CMAKE_COMMAND} -E env LANG=C LC_ALL=C /usr/bin/bash ${CMAKE_SOURCE_DIR}/tools/authors_h.sh ${CMAKE_SOURCE_DIR}/AUTHORS ${CMAKE_CURRENT_BINARY_DIR}/tools/darktable_authors.h
+  COMMAND ${CMAKE_COMMAND} -E env LANG=C LC_ALL=C ${CMAKE_SOURCE_DIR}/tools/authors_h.sh ${CMAKE_SOURCE_DIR}/AUTHORS ${CMAKE_CURRENT_BINARY_DIR}/tools/darktable_authors.h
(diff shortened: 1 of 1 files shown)


dterrahe:
It's just that I don't know how this is _supposed_ to be resolved. I got it to function working backwards from the error message and putting an ugly hack in that works for _me_ but it might inflict pain on others...

btw this seems to be going a bit in a loop: 89f0da9

parafin:
You need to set user name and email correctly, in your commits it's foobar ATM. At least a nickname should be provided, something to identify an author.

What is the error you're getting? Are you sure the error is bash not being found, instead of bash not finding the author.sh script?

parafin:
What does env command output for you? Specifically what is the content of PATH variable?

I think:

1. bash must be provided in the command, since on Windows shebang doesn't work (at least in all cases).
2. full path to bash is probably not needed, since env command should look in PATH, and I suspect it contains /bin and /usr/bin.
3. most likely bash can't find author.sh script, maybe pathname is not in format bash expects (e.g. it can't handle disk name and expects normal UNIX path starting with /, so it treats the passed argument as relative path).

So IMHO this going in circles misses the point.

chhil:
```

chill@DESKTOP-MKCA8HN MINGW64 ~
$ echo $PATH
/mingw64/bin:/mingw64/bin:/usr/local/bin:/usr/bin:/bin:/c/Windows/System32:/c/Windows:/c/Windows/System32/Wbem:/c/Windows/System32/WindowsPowerShell/v1.0/:/usr/bin/site_perl:/usr/bin/vendor_perl:/usr/bin/core_perl:/C/jdk1.8.0_152/bin:/home/chill/java/fop

chill@DESKTOP-MKCA8HN MINGW64 ~
$ which bash
/usr/bin/bash

```

When I put /usr/bin/bash , it builds. 
You can look at the conversation here  https://github.com/darktable-org/darktable/issues/6487

chhil (commented):
There was a reason why bash was added. It fixed a build problem on windows.
The problem with bash as is created a new issue and was being worked around by the /usr/bin/bash but that workaround does not seem to be universal.
You can find details here.

https://github.com/darktable-org/darktable/issues/6459#issuecomment-706577534

phrrk:
I too did a clean install of MSYS2 to try to get to the bottom of this.  The bizarre thing with this is that whilst having bash there fails, as described above, replacing with sh works fine and on MSYS2 /bin/sh /usr/bin/sh /bin/bash /usr/bin/bash are all the same thing!

$ sh --version
GNU bash, version 4.4.23(2)-release (x86_64-pc-msys)
Copyright (C) 2016 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>

This is free software; you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.

We could just introduce if(WIN32) block to get around this issue if linux/MacOS don't want to/cannot use sh.

codingdave:
thanks @dterrahe,  @chhil for linking #6459 
thanks @phrrk for linking #6487
thanks @parafin to mention my fake username. I did not intend to develop on windows and now I'm doing it :-( I hope force-pushing wasnt bad in this scenario to fix my username.

building 7a637e6f6 (clean master) with VERBOSE=1 make
> [ 49%] Generating authors.h for about dialog.
> cd /C/msys64/home/david/darktable/build/src/libs && /C/msys64/mingw64/bin/cmake.exe -E make_directory C:/msys64/home/david/darktable/build/src/libs/tools/
> cd /C/msys64/home/david/darktable/build/src/libs && /C/msys64/mingw64/bin/cmake.exe -E env LANG=C LC_ALL=C bash C:/msys64/home/david/darktable/tools/authors_h.sh C:/msys64/home/david/darktable/AUTHORS C:/msys64/home/david/darktable/build/src/libs/tools/darktable_authors.h
> /bin/bash: C:/msys64/home/david/darktable/tools/authors_h.sh: No such file or directory
> make[2]: *** [src/libs/CMakeFiles/generate_authors_h.dir/build.make:85: src/libs/tools/darktable_authors.h] Error 127
> make[2]: Leaving directory '/home/david/darktable/build'
> make[1]: *** [CMakeFiles/Makefile2:3522: src/libs/CMakeFiles/generate_authors_h.dir/all] Error 2
> make[1]: Leaving directory '/home/david/darktable/build'
> make: *** [Makefile:172: all] Error 2

However, calling that (copy,paste,execute) works:
> $ LANG=C LC_ALL=C bash C:/msys64/home/david/darktable/tools/authors_h.sh C:/msys64/home/david/darktable/AUTHORS C:/msys64/home/david/darktable/build/src/libs/tools/darktable_authors.h

my `PATH` is `PATH=/usr/lib/ccache/bin:/mingw64/bin:/usr/local/bin:/usr/bin:/bin:/opt/bin:/c/Windows/System32:/c/Windows:/c/Windows/System32/Wbem:/c/Windows/System32/WindowsPowerShell/v1.0/:/usr/bin/site_perl:/usr/bin/vendor_perl:/usr/bin/core_perl:/C/Program Files/Java/jre1.8.0_261/bin:/home/david/java/fop`

codingdave:
@phrrk I also did a clean MSYS2 install on Win10. For me it works. So we might end up having to see where our specs differ?

codingdave:
Just finished reading https://github.com/darktable-org/darktable/issues/6487. I run into the same issue. I would close this PR and continue on the existing thread.

PR #6533, opened 2020-10-15
[Issue #2451] Styles dialog updates

### LINKED ISSUE #2451: Darkroom: improve style management
The situation: I have edited a photo and now I want to save the history stack as a style.
I click on 'create a new style from history stack' and see the 'create new style' window. Here I have the following issues:

* The list of history steps to be included is incomplete. Some checkboxes are active and some are not. To use all history steps I have to click on many checkboxes.
* Since I cannot overwrite an existing style I have to keep version numbers or naming conventions in mind. When I accidentally enter a style name that already does exist, the saving process ends with an error message.
I have to click on 'create a new style from history stack' again and have to reclick many checkboxes. I choose another name and hope that it is unique now.

I like to have the following features:
* When the 'create new style' window opens I want all checkboxes to be activated. If there is a reason why not all are active, there should be a 'select all' button.
* When I try to save a style with a name that already does exist, I want a dialog saying 'This style name already exists. Do you want to overwrite?' and two buttons 'Yes' and 'No'. When I click 'Yes', the old style is overwritten. When I click 'No' I am back in the 'create new style' window with the previously entered data (names and checkboxes).

darkelectron:
1. This implements/fixes #2451 an outstanding feature mentioned in the referenced issue for the overwrite dialog on style name existence.

2. Makes sure the overwrite dialog in both __create new style__ and __edit  presets__ do not continue to overwrite when the ESC  key is pressed.

3. Added a placeholder for the filter entry of styles in lighttable.

### DIFF
--- a/src/gui/presets.c
+++ b/src/gui/presets.c
@@ -289,8 +289,11 @@ static void edit_preset_response(GtkDialog *dialog, gint response_id, dt_gui_pre
         gint dlg_ret = gtk_dialog_run(GTK_DIALOG(dlg_overwrite));
         gtk_widget_destroy(dlg_overwrite);
 
-        // if result is BUTTON_NO exit without destroy dialog, to permit other name
-        if(dlg_ret == GTK_RESPONSE_NO) return;
+        // if result is BUTTON_NO or ESCAPE keypress exit without destroying dialog, to permit other name
+        if(dlg_ret != GTK_RESPONSE_YES) 
+        {
+          return;
+        }
       }
       else
       {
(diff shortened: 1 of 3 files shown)


TurboGit (approved):
Works for me, thanks !

junkyardsparkle:
@darkelectron Style export could probably use some help, too... can't test right now, but IIRC, in the case of a filename collision, darktable reports that the style was "successfully saved" but doesn't actually overwrite (or otherwise write) the file...

darkelectron:
I think you mean import... or rather it also happens on import as well. Will definitely look into it.

junkyardsparkle:
> I think you mean import... or rather it also happens on import as well.

Thanks! I did mean export, not import, as to a "stylename.dtstyle"  XML file in a user-selected directory. It's something I forgot about until this PR reminded me.

PR #6538, opened 2020-10-15
Adjusting case for true -> TRUE, false -> FALSE

RobbieAB:
the gboolean type uses TRUE/FALSE, and this is what is used everywhere
else in the C code. The lower case versions, for some reason, were
causing build to fail on ppc64le, so it seems reasonable to switch them
for consistency.

Deeper investigation has identified that the absence of Lua on an x86_64 build would trigger a build failure on x86_64. I am not currently building ppc64le with Lua support.

### DIFF
--- a/src/control/jobs/control_jobs.c
+++ b/src/control/jobs/control_jobs.c
@@ -603,7 +603,7 @@ static int32_t dt_control_monochrome_images_job_run(dt_job_t *job)
     if(imgid >= 0)
     {
       dt_image_t *img = NULL;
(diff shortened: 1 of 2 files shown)


Nilvus:
Thanks. That fixes issue I just had when trying to build darktable since commit 569b31911e09bb1005aeab35b4b8b0cb50f4c5a4 from @jenshannoschwalm.

TurboGit (approved):
Indeed! Thanks!

PR #6542, opened 2020-10-15
po/monochrome undo

TurboGit:


jenshannoschwalm:
Thanks, also teaches me how to handle such things :-)
Will do testing tonight...

BTW there is one thing that really worries me, you remenber the message we observed related to a imgid = 0xFFFFFFFF represented as decimal. In this pr you declare the imgid as uint32_t. Isn't this simply wrong? In image.h the id is defined signed as int32_t and the cache reading functions test id for being <0 as an error condition. Don't we **have** to use this either always as int32_t and can test for <0 or should we always declare as uint32_t and initialize as 0xFFFFFFFF and test for that as an error condition? (I know - slightly off-topic).

TurboGit:
>  Isn't this simply wrong?

It is wrong! I meant int32_t let me fix...

TurboGit:
All fixed now.

jenshannoschwalm:
I checked carefully 2 days ago for these inconsistencies and i think there were definetely more. 

Otherwise this looks good for me.

PR #6546, opened 2020-10-16
[githooks] add wherami to hooks

johnny-bit:
Totally trivial: add wherami to pre-commit hook. That should help in future :)


TurboGit (approved):
Thanks !

PR #6549, opened 2020-10-16
Making imgid signed all over for consistency

jenshannoschwalm:
While debugging the monochrome workflow there was an error detected because imgid was used as an unsigned int32_t. For consistency and maintaining fixed all over.

Some might be a real problem, some are just cleanup.


TurboGit (approved):
Looks good, thanks for this clean-up !

PR #6560, opened 2020-10-17
iop_layouts : fix the list when we show all modules

### LINKED ISSUE #6555: Module list shown when no group is selected does not depend on the preset
With darktable 3.2.1, when unselecting the active group (such that no group is selected), the modules that are visible are those that are active or favorite. It looks like it is somewhat different with revision 722508b1cfd9c8f30cbab499481bf64e9152f3f9 of master: with all presets the same module list is shown. Is this the expected (new) behavior?

**To Reproduce**
1. Start `darktable --conf plugins/darkroom/workflow=none` with an empty config directory
2. Import a raw image and open it with darkroom
3. Select the "Scene referred" preset (of the module groups)
4. Click on the "base" tab and see that "filmic rgb" is present in the list of modules (but not enabled)
5. Click on the "show only active modules tab": it shows all the enabled modules
6. Click on the "show only active modules tab" again: it becomes unselected and it shows the "base curve" module but not the "filmic rgb" module.

**Expected behavior**
Only the enabled modules and the modules selected by the preset should be shown: e.g. "base curve" should not be visible, and "filmic rgb" should be.

**Platform**
 - Darktable Version: 722508b1cfd9c8f30cbab499481bf64e9152f3f9, local build with CMAKE_BUILD_TYPE=RelWithDebInfo, gcc 7.5.0
 - OS: Ubuntu 18.04
 - Same behavior with and without OpenCL

**First analysis**
In function `_lib_modulegroups_update_iop_visibility` of `libs/modulegroups.c` with `d->current == DT_MODULEGROUP_NONE`, the `module->so->state` is used to determine whether to show the module. The `module->so->state` is initialized from the `plugins/darkroom/%s/visible` and `plugins/darkroom/%s/favorite` configuration parameters in function `dt_iop_load_module_by_so` of `develop/imageop.c` and it seems that no calls to `dt_iop_gui_set_state` or `dt_iop_so_gui_set_state` to update the module state are done when selecting a new module group preset. The "base curve" is shown because it is "DEFAULT_VISIBLE" in `src/iop/CMakeLists.txt`.

AlicVB:
this fix #6555

### DIFF
--- a/src/libs/modulegroups.c
+++ b/src/libs/modulegroups.c
@@ -551,8 +551,7 @@ static void _lib_modulegroups_update_iop_visibility(dt_lib_module_t *self)
         case DT_MODULEGROUP_NONE:
         {
           /* show all except hidden ones */
-          if((module->so->state != dt_iop_state_HIDDEN || module->enabled)
-             && (!(module->flags() & IOP_FLAGS_DEPRECATED)))
+          if(_lib_modulegroups_test_visible(self, module->op) || module->enabled)
           {
             if(w) gtk_widget_show(w);
           }


TurboGit (approved):
Works for me, thanks!

PR #6576, opened 2020-10-19
usermanual : iop layouts

AlicVB:
Documentation for the new iop layout presets/manager

Keep in mind that I'm not really good at writing English, so proofreading is needed here ;)


TurboGit:
@AlicVB : I'll review, I'm not terrible in English either, but you may know that there is an ongoing project to rework the whole dt documentation. So at the end it does not matter too much :)

AlicVB:
Ok, I've also completely reordered the modules part to present them like the default preset.
Not too much to proofread, but maybe some double-check needed, to verify I've not missed or inverted something ;)

When doing this work, I've found that basic adjustments isn't described... @TurboGit : should I open an issue for that ?

AlicVB:
a note for @aurelienpierre and any other proofreader : 
the description of the three groups are copy-paste from PR #5322 . Maybe we should enhance them a bit, especially for the third one ?

TurboGit:
> When doing this work, I've found that basic adjustments isn't described... @TurboGit : should I open an issue for that ?

Yes please !

doc/usermanual/darkroom/modules/effects/group_effects.xml
```diff
@@ -1,35 +1,44 @@
-<sect2 status="final" id="effect_group">
+<sect2 status="final" id="effects_group">
```
TurboGit, line 1: Don't we need to update the integrated help (when clicking on ?) as links have changed? Probably true for many other changes here I think.

AlicVB: oh... forgot that... But yes, there should be many links to change (at least all the iops...) I'll have a look !

AlicVB: hmm... so it seems I have *tons* of links to change in `usermanual_url.c ` (basically each iop link as changed)
problem : once they are changed, they will point to invalid urls until we release the manual... should I do this anyway ? or should we postponed it ?

AlicVB: done. and in the same time I've fixed lot of invalid urls or missing ones. Surprising that no user raise issues... Maybe a sign that this feature is underutilized ?

TurboGit (approved):
Good to me, thanks !

PR #6578, opened 2020-10-19
Travis revamp

codingdave:
My playground trying to help @johnny-bit


TurboGit:
@codingdave : Looks like this has introduced an issue in the build and some rules are failing, see:

https://travis-ci.org/github/darktable-org/darktable/builds/737751910

codingdave:
@TurboGit still wip, there is also revamp 2, both don't work yet. Trying to help out @johnny-bit. We will ping you when we're done. Thanks.

PR #6582, opened 2020-10-20
Workaround for lensfun no longer required on windows

phrrk:
The MSYS repository has reverted lensfun  to version 0.3.2 from broken 0.3.95 version and hence workaround is no longer required


TurboGit (approved):
Looks ok. Thanks.

PR #6584, opened 2020-10-20
do not treat enabling dynamic shortcuts as special

dterrahe:
dynamic shortcuts were implemented using a different approach from all other shortcuts. A separate list is maintained and scanned every time an unmapped key is pressed. I'm not sure why this approach was chosen, but I changed this, ripping out all the dedicated data structures and just using a normal shortcut with callback function to enable the dynamic mode. This mode is then disabled, as before, when a key release is received.

The testing I have done shows no change in behavior at all, but less code and no special case so easier to maintain/extend.

@elstoc can I ask you to hammer on this a bit, since you know shortcuts. If this works for you, it will make the other cleanups I'm working on much easier. Thanks!

(As a reward, I'll give you "processing modules" and "utility modules" as top level names in shortcut settings, without changes to keyboardrc)

### DIFF
--- a/src/control/control.c
+++ b/src/control/control.c
@@ -191,11 +191,6 @@ void dt_control_cleanup(dt_control_t *s)
   {
     g_slist_free_full(s->accelerator_list, g_free);
   }
-  if(s->dynamic_accelerator_list)
-  {
-    g_slist_free(s->dynamic_accelerator_valid);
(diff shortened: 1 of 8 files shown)


elstoc:
Seems to work well for me and much cleaner code. No idea why dynamic accels were built like this.

TurboGit (approved):
Looks good, thanks!

Issue #6586, opened 2020-10-20
redesign copy and paste workflow


parafin:
Split from #6564:

I find current copy-paste system unintuitive. I think there should be 3 buttons instead of current 4:

1. copy (without dialog)
2. filter (dialog where you select the iops to transfer)
3. paste (without dialog)

The current situation where iop selection is bundled with copy and paste operations results in complexity and non-transparent behaviour without adding anything.

Now, filter button in the proposed scheme is where all the magic happens. It should display its status (all selected, only safe, custom selection), and we can have presets for it. By default it should be only safe modules (whatever the list might be), and reset action should reset it to that state. Not sure if its state should be saved across dt restarts, probably not, presets should be used instead. I think it will fix all use-cases without the need for yet another preference option (really, you gotta stop adding them).

Answering to @TurboGit comment:
> > I think there should be 3 buttons instead of current 4:
> 
> Not sure, one of my workflow is:
> 
> * copy all
> * select some pictures
> * paste parts
> * select some other pictures
> * paste parts (not the same iop than above)

It will change to:

* copy
* select some pictures
* filter
* paste
* select some other pictures
* filter
* paste

You can also paste the same iops several times (to different selections) without going through the selection dialog.

> 
> The other is:
> 
> * copy parts
> * select some pictures
> * paste all

It will change to:

* copy
* select some pictures
* filter
* paste

Notice how it's not different at all from the first use-case, meaning it's now simpler to understand. And this use-case is exactly the problem I'm having with current design - what does "paste all" pastes here? Only iops selected in "copy parts" action? But then if I do "paste parts" I can then click "select all" in appeared dialog and paste more iops than "paste all" does. What does "all" mean then? At the very least it's a poor choice of words.

elstoc:
I still think the copy and select/filter should be combined - it makes it easier when using it with shortcuts outside of the lighttable. You would have two forms of copy (with or without the dialog) and then only a single form of paste that copies whatever was selected. You can customise the copy-without-dialog using a default set of "all" or "safe" within the lighttable module. Similar to how you select overwrite or append modes.

parafin:
Well, we can have shortcut that calls both copy and filter, why not. No need to change anything in the design to do that IMHO. And yes, selected filter should be persistent across view changes.

elstoc:
I think that's the more common use case so it should be simple. Is it really a common use case to paste different subsets of modules from a single source to different destinations (which feels like the only time you'd need to re-filter)?

[...]

PR #6587, opened 2020-10-20
po/copy dialog

TurboGit:


elstoc:
Now it feels like the "copy..." and "paste..." buttons need renaming though. Perhaps they could be "copy selected" and "paste selected"

TurboGit:
Why? Don't the ... Is not clear enough? This is a common GUI indication that a dialog will be used, no?

elstoc:
Is it? ok

parafin:
It is indeed a common thing, but having buttons that differ only by these dots is still probably not the best idea. Maybe "copy partially"? "copy selected" sounds like it's referring to selected images (after all when you press, you haven't selected any iops, so they are not "selected" at that point).

elstoc:
"copy/paste parts"? 

"copy partially" doesn't read well.

johnny-bit:
While having pre-selected "safe" modules in dialog is ok, it's very funny to see "copy ... / copy " buttons next to each other (similar to paste)

(i still think wb + lens should be considered "safe" but thats for other PR)
and now I think about something else - you know you can enter copy... dialog, then press cancel, yet it still marks image as copy source and preselected modules are selected for copying... form me more and more @parafin's #6586 makes loads of sense...

junkyardsparkle:
My thought with the naming was that "copy..." would be ok if "copy" was renamed to indicate that it was a "quick" action with no intervening dialog. In any case, it seems like distinguishing the action names clearly would be good not only for buttons, but for clarity in places like keyboard shortcut assignment and maybe even within the code.

TurboGit:
@elstoc : I'll go with "copy/past parts".

PR #6591, opened 2020-10-21
rename shortcuts categories for modules  in settings

dterrahe:
The current separation of shortcuts into 'modules' and 'image operations' isn't very clear (since image operations are also modules). This has been resolved in the dtdocs version of the user manual by instead naming them "processing modules" and "utility modules".

This PR adopts that naming convention in the shortcuts screen in settings.

There is no impact on keyboardrc, which will keep using the old category names, so existing keyboardrc files will not be broken.

Translations are also not impacted, except of course the two additional terms need new translations.


elstoc:
Works fine for me. Thanks @dterrahe

TurboGit:
Work for me! Thanks.

PR #6592, opened 2020-10-21
Quick presets fix + docs

AlicVB:
Some work on the "quick" presets : 
- fix : only show modules with presets in the manager (and in alphabetical order)
- fix : remember the state of presets in the manager
- docs : slightly change the usermanual to reflect the change from 3.2

### DIFF
--- a/doc/usermanual/darkroom/panels/bottom_panel.xml
+++ b/doc/usermanual/darkroom/panels/bottom_panel.xml
@@ -25,8 +25,13 @@
     <para>
       Clicking the
       <inlinegraphic fileref="&icon_module_presets;"  scalefit="1" width="2%" align="center"/>
-      icon opens a combobox that gives you quick access to your favorite module's presets. Click
+      icon opens a menu that gives you quick access to your favorite presets. Click
       on the preset name to apply it to the image.
+    </para>
+
+    <para>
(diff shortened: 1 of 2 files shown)


TurboGit:
I'm wondering if the module name should not be displayed in the preset menu. When you have preset from different module it is not clear which module the preset will use/change.

AlicVB:
yes, I wonder that too... I just fear that it was too long, but in fact, due to the position of the icon, we have at least all the center widget width for us. I'll do a try...

AlicVB:
@TurboGit : done (+ order the presets like the pipe)
This was in fact a necessity because of multi-instance : now we now on which instance we apply the preset !

TurboGit:
Looks way better this way! Thanks @AlicVB!

junkyardsparkle:
Nice! Maybe I can stop prefacing my preset names with the module name now. :-)

I'm finding more use for this feature now that I have builit-in presets disabled in the module menus... it's convenient to just enable the few built-ins I want to use here, and now those show module names, too. I wonder if it would be worth having checkboxes in the manager for the modules themselves, to quickly enable/disable all of that modules presets?

Mark-64:
Why not grouping per module ? The list can become very long with one level only

PR #6594, opened 2020-10-21
common:imageio: Add support for libavif >= 0.8.2

cryptomilk:
This gets us a bit deeper into #ifdef hell. We could also require libavif 0.8.2 and remove all of that. The questions when is the right time to require a newer libavif version :-)

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -327,7 +327,7 @@ if(USE_WEBP)
 endif(USE_WEBP)
 
 if (USE_AVIF)
-    find_package(libavif 0.7.2 CONFIG)
+    find_package(libavif 0.8.2 CONFIG)
     if (TARGET avif)
         list(APPEND LIBS avif)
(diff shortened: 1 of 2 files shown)


cryptomilk:
@TurboGit would you be fine with requiring libavif 0.8.2? Then I would throw out all the #ifdef stuff.

TurboGit:
I need more data since I don't use/build with libavif. I would say yes if libavif 0.8.2 is now a release that we find in some platforms (for the record 0.8.2 has landed into GNU/Debian sid). If on all platforms we need to build this library from sources I think it is fine to require 0.8.2 as the minimal version.

cryptomilk:
Great, I've updated openSUSE to libavif 0.8.2 and also Fedora. Then I will change the patch to require that version!

TurboGit:
Sounds good !

cryptomilk:
Ready to go :-)

TurboGit:
Cannot test myself, but looks good. Thanks.

PR #6598, opened 2020-10-21
fix preset popup accel

### LINKED ISSUE #6573: How are userdefined shortcuts supposed to work?
How are userdefined shortcuts supposed to work?

I have setup the following
image

The idea is one preset for export applies a style which removes frames.

So I created a shortcut Ctrl+M to show me the menu to select any preset.
I created a shortcut specific to the no frame preset so that I can export somehow using the preset.

I press the shortcuts in darkroom (when I am on an image that I want to export) and lightable (by selecting an image) but not seeing the preset change or ask me for a selection. 

Pretty sure I don't understand how to invoke the shortcuts.

Also can we have the ability to have more than 2 keystrokes for shortcut? My initial idea was to use CTRL+E + <another letter> for export related shortcuts, but it takes the first 3.

elstoc:
Resolves #6573

### DIFF
--- a/src/develop/imageop.c
+++ b/src/develop/imageop.c
@@ -1933,25 +1933,20 @@ static void _preset_popup_position(GtkMenu *menu, gint *x, gint *y, gboolean *pu
 }
 #endif
 
-static void popup_callback(GtkButton *button, GdkEventButton *event, dt_iop_module_t *module)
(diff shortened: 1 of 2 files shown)


elstoc:
sorry just spotted the issue was raised in the lighttable. Let me fix that too.

TurboGit:
Looks good, thanks!

PR #4131, opened 2020-01-17
Code cleaning

Mark-64:
Windows specific code cleaning:
- removing a function no longer needed
- creating a DT_DEBUG_WIN debug option, for future use

PR #6606, opened 2020-10-22
using glib function to trash files in windows

Mark-64:
Continues from #4131 (rebased)

For me using the glib function to trash files in windows is working fine, there are no confirmation dialogs, so we can drop the special function used before and simplify the code.
But if someone else tests the PR, that would be nice.


Mark-64:
Did more tests, in reality it depends on setting in windows the confirmation of trashing files.
This is not desired, I believe.
So no, it doesn't always work.
closing

PR #6613, opened 2020-10-23
Map: add locations feature. Locations are visible in geotag collections.

phweyland:
As suggested [here](https://github.com/darktable-org/darktable/pull/6303#issuecomment-696995980) this PR brings the possibility to create locations (longitude, latitude, radius) on map.
These locations are visible in geotagging collection.
The user can jump from a given location to the corresponding lighttable collection.
Note: as locations can overlap the sum of collected images can be higher than the the total number of images.

[Fix](https://github.com/darktable-org/darktable/commit/24d9cf4accd16250849db312827b9a4994c86377) also a small bug on the new collection list when some metadata are hidden.

image

EDIT. I've avoided to create a new table in the db and used the tags synonyms to store the locations coordinates (see line 1771 in tags.c about how the information are extracted in a query). If found not appropriate I can create a specific table (with some code around :)) instead.


Nilvus:
I like the idea but I would not implement it that way:
- It would be better to link that to _find location_ module, by marking location with pointer as you add it with that PR but also by searching it. So, improving _find location_ by adding that to it would be better than adding another module like you do on that PR.
- Also, It would be better to have a specific filter (with lot of locations, I'm worried about not seeing directly tagged and not tagged images (I like the fact that actually they are just together to quickly find images I do not geotag. So having a specific table would probably be better for me.

phweyland:
>     * It would be better to link that to _find location_ module, by marking location with pointer as you add it with that PR but also by searching it. So, improving _find location_ by adding that to it would be better than adding another module like you do on that PR.

Find location is related to map provider. The main purpose is to navigate on the map, not to find related images.
I don't see how to integrate map with user location, used to group images. If I'm mistaken, please develop a bit more.


>     * Also, It would be better to have a specific filter (with lot of locations, I'm worried about not seeing directly tagged and not tagged images (I like the fact that actually they are just together to quickly find images I do not geotag. So having a specific table would probably be better for me.

An high number of locations can be indeed an issue with a flat representation. I've already thought about it: as integrated with tag, it should not be a lot of work to have a hierarchy of locations (both in the locations module and in collect module). In that case all locations would be descendent of "tagged" category (by principle).
You may have missed them but the category "tagged" and "not tagged" are still there as usual. The unique difference is that "tagged" group only geotagged images which are outside of all locations.
Having a specific table would not change anything from that point of view (just replicate a lot of mechanisms which exist in tags).
Instead a specific table could just provide better performance, but I'm even not sure (as there is no indexing). As performance standpoint I only see some cache mechanism??

Nilvus:
> > ```
> > * It would be better to link that to _find location_ module, by marking location with pointer as you add it with that PR but also by searching it. So, improving _find location_ by adding that to it would be better than adding another module like you do on that PR.
> > ```
> 
> Find location is related to map provider. The main purpose is to navigate on the map, not to find related images.
> I don't see how to integrate map with user locations, used to group images. If I'm mistaken, please develop a bit more.

I know that. What I mean is to add locations in _find location_ module to allow search a location, then add it and name it as your PR allows in same module. A process could be:

1) Search a location to point on the map like we already do in _find location_ module
2) Add it to locations
3) Name/rename location for filter collection

or

1) directly create a new location (like that PR)

Hope, I'm more clear here. Another way to summarize my idea: _find location_ module and _locations_ module are really near and could work together, so I think both could be in same module, each one for what it does actually AND add the possibility to link a location found in _find location_ like actually in master and it as a location like with that PR.

> > ```
> > * Also, It would be better to have a specific filter (with lot of locations, I'm worried about not seeing directly tagged and not tagged images (I like the fact that actually they are just together to quickly find images I do not geotag. So having a specific table would probably be better for me.
> > ```
> 
> A high number of locations can be indeed an issue with a flat representation. I've already thought about it: as integrated with tag, it should not be a lot of work to have a hierarchy of locations (both in the locations module and in collect module). In that case all locations would be descendent of "tagged" category (by principle).
> You may have missed them but the categories "tagged" and "not tagged" are still there as usual. The unique difference is that "tagged" group only geotagged images which are outside of all locations.

I didn't missed them, just that I found that easier to have tagged and not tagged following. So, having locations as a descendent of "tagged" category would be much better.

> Having a specific table would not change anything from that point of view (replication of a lot of mechanisms which exist in tags).
> Instead a specific table could just provide better performance, but I'm even not sure (as there is no indexing). As performance standpoint I only see some cache mechanism??
> EDIT: I may have not been clear on this last point. A specific table would allow to have 3 specific columns for longitude, latitude and radius (avoiding to have to split the synonyms column in three parts). I don't see any other difference.

If locations go to a descendent of "tagged" category, locations could stay in geotagging filter, so you could forget the idea to separate them from geotagging filter.

phweyland:
> _find location_ module and _locations_ module are really near and could work together,

Got it. Yes, they are close. In fact that works exactly as you describe.
1. centre the map on the localisation you want (by find location or by moving the map)
2. click on new location and manage the newly created location. 

The two steps are mandatory because the user can use one without the other.
The name "find location" is not accurate for both features.
So I don't see any roadblock to put all features in a same module but I cannot say we should do it (there were similar discussions about grouping or not features in iop :) ).

phweyland:
> If locations go to a descendent of "tagged" category, locations could stay in geotagging filter, so you could forget the idea to separate them from geotagging filter.

Working on it.

Nilvus:
> > _find location_ module and _locations_ module are really near and could work together,
> 
> Got it. Yes, they are close. In fact that works exactly as you describe.
> 
>     1. centre the map on the localisation you want (by find location or by moving the map)
> 
>     2. click on new location and manage the newly created location.
> 
> 
> The two steps are mandatory because the user can use one without the other.
> The name "find location" is not accurate for both features.
> So I don't see any roadblock to put all features in a same module but I cannot say we should do it (there were similar discussions about grouping or not features in iop :) ).

We could put them in a same module, and change module name. I don't see any problems there. _locations_ would be suitable for both modules and just split them in 2 parts: with tabs or on same view (like color balance for example).
Or at worst, put _locations_ module just below _find location_ one, not after _map settings_.

phweyland:
Currently I'm creating "locations" with longitude, latitude and radius. Coming to collection, the query retrieves the images whose (longitude, latitude) matches the "locations". Performances wise that may not be the more scalable process ...

An alternative would be to tag (in the sense of tagging) with the "location" name each image which matches the "location" at each "location" creation/update. On the other hand whenever the geotag of an image is changed, its potential "locations" are updated.
As a result, considering that all (hierarchical) "locations are below the tag category "map", the corresponding collections would be found in tag collection, under the "map" root. The geotagging collection would come back the original one.

Hoping to be clear enough, any thought ?

TurboGit:
@phweyland : Nice new feature ! And yes I think it would be better to have a new table/column in the database.

phweyland:
image

Remains to synchronize with imported image (exif/xmp) but can be tested as is if somebody wanted to.

phweyland:
Update on import images with geotags done.

phweyland:
some fixes on the treeview.

phweyland:
synchro with Ctrl-Z/Ctr-Y on geotags

phweyland:
@turbogit, I'm short on detected issues. You can review when you can. Thanks.

Nilvus:
Great work, much better. Thanks @phweyland.

TurboGit:
I'm starting slowly to test:
- when clicking "hide location" I would deselect the currently selected location.
- when a location is selected "new location" will create a sub-location, maybe updating the button label in this case would help.

I would prefer to have the location in a proper table with lat/long/radius, this would make it possible to "easily" have SQL request to select all images into the area (bounding box).

phweyland:
> * when clicking "hide location" I would deselect the currently selected location.
> * when a location is selected "new location" will create a sub-location, maybe updating the button label in this case would help.

Good ideas. I'll do.




> I would prefer to have the location in a proper table with lat/long/radius, this would make it possible to "easily" have SQL request to select all images into the area (bounding box).

What do you mean by easily ? For the developer ? If yes, I can package back the query as it was few commits before. This turn the integration in others queries easy.
Otherwise how would you see the design ? The options I can see now:
- to have an extension to the tags table: location name in tags table and geotag data in the extension.
- to have a geotags table: location name and geotag data in this table. We need another images_tagged table to link the images.
- any other idea ?
In both cases I can think off, we need to rewrite a tens of basic routines (new, remove, attach, detach, get data, set data, exists, rename, get list,...

Note 1. Map view seems to be built in a different way than lighttable or iops. Instead of a common set of routines in src/common, it uses a set of proxy routines hold by view.h Is there a specific reason for that ? As a consequence, map_location plugin cannot expose services for map ...

TurboGit:
What I had in mind is to be closer to what a collection is:
- new table geolocation (id, name, lat, long, radius)
- new field in images (geo_id) PRIMARY_KEY of geolocation.id
- a new collection geo-location (which will be able to combine with other collection as usual)

And we don't need a new  images_tagged kind to link the images since the relelation between image and geo-location is 1:1.

And yes it probably needs more work than your proposal. But not much I would say since we have only a new field which is already removed when a row in images table is removed.

You already do all the rest, when you place an image on the map inside a location you do add the corresponding tag, here you just need to set images.geo_id with the proper value. When you move an image outside of a location you remove the tag, here you just need to set (clear) images.geo_id to NULL. Let me know if I forgot something....

How does that sound ?

phweyland:
> And we don't need a new images_tagged kind to link the images since the relation between image and geo-location is 1:1.

The relationship is not 1:1. An image can be in several locations (circles which overlap). And a locations usually have several images. Really as tags.

On my first version, locations were not linked to images (not tagged). The collect query had to retrieve images per location based on the fit between image's coordinates and locations definition. Quite demanding when locations cover 1000s of images.

On the second version, relationship between images and locations is updated on any change of locations and/or image's coordinates. The collect query works for locations exactly as for tags.

Moreover, at the beginning I was thinking about locations which could have different shapes (circle, rectangle, ...) and even be combination of shapes, for example to gather images along a path. Not sure there were any demand on this. So I haven't done anything for this, but the extension will remain possible. 

I share the concern not having a simple table (and a simple query). But I was concerned also for duplication of code.
Probably dedicated tables will facilitate future improvement.
I'll make a try.

quovadit:
I would love to have this feature, and it would be great, if I could draw a concrete polygon (not just a rectangle) in order to define my shooting-areas

TurboGit:
> The relationship is not 1:1. An image can be in several locations

Missed that indeed!

TurboGit:
> Probably dedicated tables will facilitate future improvement.

Yes, certainly my thinking too.

phweyland:
> > * when clicking "hide location" I would deselect the currently selected location.
> > * when a location is selected "new location" will create a sub-location, maybe updating the button label in this case would help.
> 
> Good ideas. I'll do.

Done. plus some treeview control cleaning.

phweyland:
changes requested are done.

quovadit:
This feature is awesome, thanks a lot!

Just one question: How do I select all geo-tagged photos?
- Clicking on 'tagged' shows only those images which do not belong to a drawn location
- Ctrl-Click shows only photos which do belong to a drawn location
- Shift-Click should then show all geo-tagged photos in my understanding, but it does the same as Ctrl-Click

TurboGit:
One issue:
- create a location & place it on the map
- click on the locations list in an empty line
- no more selection in this locations module
- the [hide location] menu is gray out -> fine
- BUT the location circle on the map is still visible (it should he hidden)

TurboGit:
I would add a tooltip over the button to select circle or square area.

TurboGit (changes requested):
My first review...

We also need to add common/map_locations.c (even if in this one I did not spot any translated strings at the moment) and lib/map_locations.c into po/POFILES.in.

There is no undo/redo for the locations, but I suppose this can come later as this is certainly not a blocking issue.

I find all this a really nice new feature. It makes the map module a first class citizen and better integrated with the way collections are related to the new location support. Love it !

src/common/map_locations.c
```diff
@@ -0,0 +1,398 @@
...
+*/
+
+#include "common/geo.h"
+#include "common/map_locations.h"
+#include "common/darktable.h"
+#include "common/debug.h"
+#include "common/tags.h"
+
+// root for location geotagging
+const char *location_tag = "darktable|loctag";
```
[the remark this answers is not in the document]
phweyland: ok. done

src/common/database.c
```diff
@@ -1854,6 +1854,15 @@ static int _upgrade_data_schema_step(dt_database_t *db, int version)
 
     new_version = 6;
   }
+  else if(version == 6)
+  {
+    TRY_EXEC("CREATE TABLE data.locations "
+             "(tagid INTEGER PRIMARY KEY, type INTEGER, longitude REAL, latitude REAL, "
```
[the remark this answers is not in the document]
phweyland: ok

src/common/database.c
```diff
@@ -2033,6 +2042,10 @@ static void _create_data_schema(dt_database_t *db)
                NULL, NULL, NULL);
   sqlite3_exec(db->handle, "CREATE UNIQUE INDEX data.presets_idx ON presets (name, operation, op_version)",
                NULL, NULL, NULL);
+  ////////////////////////////// (map) locations
+  sqlite3_exec(db->handle, "CREATE TABLE data.locations (tagid INTEGER PRIMARY KEY, "
```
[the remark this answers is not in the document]
phweyland: ok

phweyland:
Thanks for testing!

@TurboGit 
I agree with all your comments.
About undo/redo. The geotag side (change on image) (should) works as usual. Remains the location side. There is no undo/redo on tags but this may not be an issue. But yes, can be dobne later one.

@quovadit 
ctrl or shift click (% or |%) work on a location, but not on "tagged" key. I'll try to find a solution.

phweyland:
> @quovadit
> ctrl or shift click (% or |%) work on a location, but not on "tagged" key. I'll try to find a solution.

In fact tagged|% does work. tagged% doesn't work...

phweyland:
I think I've answered all the topics, including the @quovadit one.
I've also added the deletion of groups of locations.
It was already possible to create some levels like France|Paris, France|Marseille where France is not a location but only a group.
To remove it, just replace France by an empty string.
Thanks

src/common/database.c
```diff
@@ -1854,6 +1854,15 @@ static int _upgrade_data_schema_step(dt_database_t *db, int version)
 
     new_version = 6;
   }
+  else if(version == 6)
+  {
+    TRY_EXEC("CREATE TABLE data.locations "
+             "(locid INTEGER PRIMARY KEY, type INTEGER, longitude REAL, latitude REAL, "
```
TurboGit, line 1860: Sorry I see that I was wrong, the id here is really the tagid to which it corresponds, right ?

If I detach a tag the picture is till in the location, the location still count this image as in (as shown by the count around parenthesis). So maybe something to improve here ?

TurboGit: To make this clear we need to write:

`CREATE TABLE data.locations (tagid , locid INTEGER PRIMARY KEY, type INTEGER, longitude REAL, latitude REAL, delta1 REAL, delta2 REAL, FOREIGN KEY(tagid) REFERENCES tags(id));`

src/common/database.c
```diff
@@ -2033,6 +2042,10 @@ static void _create_data_schema(dt_database_t *db)
                NULL, NULL, NULL);
   sqlite3_exec(db->handle, "CREATE UNIQUE INDEX data.presets_idx ON presets (name, operation, op_version)",
                NULL, NULL, NULL);
+  ////////////////////////////// (map) locations
+  sqlite3_exec(db->handle, "CREATE TABLE data.locations (locid INTEGER PRIMARY KEY, "
```
TurboGit, line 2046: same as above

src/common/map_locations.h
```diff
@@ -0,0 +1,92 @@
...
+  guint id;
+  gchar *tag;
+  guint count;
+} dt_map_location_t;
+
+// create a new location
+const guint dt_map_location_new(const char *const name);
+
+// remove a location
+void dt_map_location_delete(const guint tagid);
```
TurboGit, line 50: As a clean-up I would also rename all tagid in the parameters to locid. I understand that now the relation between both but it reads better here to talk about location id and not tag id.

TurboGit (changes requested):
second review... and sorry for having missed the tagid & locid relation before.

src/common/map_locations.h
phweyland: Ok, so to resume, on db side, back to tagid, and all location parameter usage of locid instead of tagid, right ?

TurboGit: Right, that's my suggestion. Is that ok with you ?

src/common/database.c
phweyland: > If I detach a tag the picture is till in the location. 

Just to be sure: you detach the tag in tagging plugin, correct ?

phweyland: Good catch!

This case should be avoided. The relationship is not user dependent but location data dependent.
If the user removes the tag, he screws the location mechanisms.
In the same way, it would be wrong to attach the location tag to an image outside the location area.
Unless it's intentional....
Not sure about we should do here...

phweyland: Additional note. As soon the location changes, the user action will be lost (as the relationship is calculated again).

phweyland: I see 2 possibilities.
1. forbid to attach/detach locations from tagging. This seems to be the more logical (as the relationship is calculated).
2. "signal" the change to location plugin, which in turn recalculates the location's images and undoes the user action. This may give inconsistent result between map and collect module (if collect module is triggered first).

Thought ?

phweyland: In fact "darktable" should not be attachable/detachable by a direct user action, as they are all darktable mechanism dependent.
Do you agree with that ?

phweyland:
changes for tagid/locid done.
I've let the issue with tagging for the time being, waiting for you standpoint.
On my side I would say that darktable tags should not be manipulated by the user. In any case he can expect unexpected consequences. This can be seen as a drawback or an advantage (the possibility to fix manually a dt issue)...

src/common/map_locations.h
phweyland: yes

phweyland:
I experience some disagreements when I move images on the map:
- if no active location, the map recentre on the collection: the map should stay as it is.
- if location, I cannot move an image inside the location. Something like ctrl-click to give priority to image (over location)
Another one is zooming the map when a location is active. As I have to scroll outside the location, the centre of the map is lost. A zoom on the map when location active should keep the current centre (centre of location).
I'll do this.

phweyland:
Above changes done.

phweyland:
Question.
Currently we have, click to select, double click to edit, button to hide.
You may find more practical (more natural) to have:
- click to select (show)/unselect (hide) as a toggle button, and the right button to edit (if selected).
What do you think ?

src/common/database.c
TurboGit: Yes, I would tend to agree that internal dt tags should not be manipulated in general. I'm not sure to see all the consequences if we forbid that but I'm ok to do that for 3.4. It cannot make dt unusable anyway just maybe some annoyances...

TurboGit:
>     * click to select (show)/unselect (hide) as a toggle button, and the right button to edit (if selected).

And removing the [hide location] button altogether, that's what you propose ?

src/views/map.c
```diff
@@ -938,6 +1231,65 @@ static gboolean _view_map_motion_notify_callback(GtkWidget *widget, GdkEventMoti
...
+  {
+    OsmGpsMapPoint *p = osm_gps_map_get_event_location(lib->map, (GdkEventButton *)event);
+    float lat, lon;
+    osm_gps_map_point_get_degrees(p, &lat, &lon);
+    if(dt_map_location_included(lon, lat, &lib->loc.data))
+    {
+      if(lib->loc.data.shape == MAP_LOCATION_SHAPE_RECTANGLE &&
+        (event->state & GDK_SHIFT_MASK))
+      {
+        if(event->direction == GDK_SCROLL_UP)
```
TurboGit, line 1247: this must be inverted for consistency with mask.

src/views/map.c
```diff
@@ -938,6 +1231,65 @@ static gboolean _view_map_motion_notify_callback(GtkWidget *widget, GdkEventMoti
...
+          lib->loc.data.delta1 *= 1.1;
+        else
+          lib->loc.data.delta1 /= 1.1;
+      }
+      else if(lib->loc.data.shape == MAP_LOCATION_SHAPE_RECTANGLE &&
+             (event->state & GDK_CONTROL_MASK))
+      {
+        if(event->direction == GDK_SCROLL_UP)
```
TurboGit, line 1255: likewise

src/views/map.c
```diff
@@ -938,6 +1231,65 @@ static gboolean _view_map_motion_notify_callback(GtkWidget *widget, GdkEventMoti
...
+          lib->loc.data.delta2 *= 1.1;
+        else
+          lib->loc.data.delta2 /= 1.1;
+      }
+      else
+      {
+        if(event->direction == GDK_SCROLL_UP)
```
TurboGit, line 1262: likewise

TurboGit (changes requested):
ensure growth/shrink of forms on the map is consistent with mask forms.

phweyland:
> EDIT: I would use toggle to show/hide location. double-click to edit and remove [hide locations].

I haven't found the way to get double-click without the click...

src/views/map.c
phweyland: OK

TurboGit:
> I haven't found the way to get double-click without the click...

So right-click will do.

phweyland:
last commit with those changes:
- invert scroll direction for location sizing
- click on selection works in a toggle way, driving the show / hide location
- now the expand button works independently from row selection (was not the case before)
- remove the hide button
- disable attach/detach actions in attached_view of tagging plugin. This doesn't prevent other darktable tag change through other actions like copy metadata tags, ...

phweyland:
Plus
- ctrl-shift to edit (to avoid double effect with click)
- "edit" entry in sub menu

TurboGit:
@phweyland : You missed the comment here maybe about the foreign key : https://github.com/darktable-org/darktable/pull/6613#discussion_r515520852

phweyland:
>  You missed the comment here maybe about the foreign key : [#6613 (comment)](https://github.com/darktable-org/darktable/pull/6613#discussion_r515520852)

Sorry. I've missed that. Far better indeed.

phweyland:
DB issue fixed.

I've two ideas wip.
- try to stabilize the group of images when map is moved. The current regrouping gives a bad effect. However when zooming the map in or out the regrouping is normal as the density changes.
- add a "show others" / "hide others" button to show / hide the other locations present on the current portion of map. I miss that to better control locations which are closed each other.
These two pieces are independent and can be done in another PR (I would prefer).

src/common/database.c
TurboGit, line 1860: `PRIMARY KEY` should be removed from here.

src/common/database.c
```diff
@@ -2033,6 +2042,11 @@ static void _create_data_schema(dt_database_t *db)
                NULL, NULL, NULL);
   sqlite3_exec(db->handle, "CREATE UNIQUE INDEX data.presets_idx ON presets (name, operation, op_version)",
                NULL, NULL, NULL);
+  ////////////////////////////// (map) locations
+  sqlite3_exec(db->handle, "CREATE TABLE data.locations (tagid INTEGER PRIMARY KEY, "
```
TurboGit, line 2046: likewise

TurboGit:
>     * These two pieces are independent and can be done in another PR (I would prefer).

Agreed. Let me test, fix the SQL statements and merge if all is well.

TurboGit:
There is a new regression, the circle location are not kept and seems to always be the location where they were created.
- create a new location
- move it
- change size
- unselect it
- now select it again, and see it does not corresponds to last location/size

TurboGit (changes requested):
Two more fix needed:
- remove primary key from SQL
- fix location of shapes

src/common/database.c
phweyland: I would agree on this.
Foreign key maintains the relationship with the master table.
Primary key guaranties the unicity of the key.

image

src/common/database.c
phweyland: idem

phweyland:
>     * now select it again, and see it does not corresponds to last location/size

Cannot reproduce this. Please consider that selecting the location again centres the map on that location (you may have moved) and resizes the map to fit with the location size.

src/common/database.c
phweyland: Sorry for the tipo. The primary key should stay.

TurboGit: Ok, you're right !

TurboGit (approved):
All good as now I cannot reproduce the issue with the location/size of the circles.

PR #6615, opened 2020-10-23
preferences: make the changed preferences with a little tick character.

TurboGit:


TurboGit:
@Nilvus : this is my proposal to highlight the non default prefs.

TurboGit:
It is not dynamic... maybe we can do better with some CSS rules ?

TurboGit:
I found previous PR and discussion #6433.

TurboGit:
@johnny-bit : are you still looking at this ? Maybe this PR is just a good compromise for now until we have a better solution ?

Nilvus:
I will test that ASAP.

johnny-bit:
I hoped to have guidance from @Nilvus on what to do and do it with css... I'm no fan of changing labels if we can avoid it. It should be possible to do checkmarks in css too but IMO checkmarks look worse than bold/italic label but I have no idea for something both nice looking and easy to code.

elstoc:
I'm also not a big fan of the bold/italic. I do like the idea of visually indicating changed prefs but I'm not sure a tick does it for me either (since some people might read other meaning into it - like "this is a good setting - well done"!). Would an asterisk be too inconspicuous?

TurboGit:
Yes asterisk is fine to me.

elstoc:
Could also change the 'double-click to reset' tooltip to say 'this configuration is at its default value' or something

Nilvus:
Asterisk is not fine for me. Asterisk means (at least in french but If I'm not wrong also in english) required setting/field... So, not a good idea.

Nilvus:
> Could also change the 'double-click to reset' tooltip to say 'this configuration is at its default value' or something

Like that but I would suggest: "double-click to reset this as it's default value".

elstoc:
> Like that but I would suggest: "double-click to reset this as it's default value".

I'm not sure what you mean by this. I'm suggesting a tooltip for when it's already at the default value, so double-click won't do anything.

Nilvus:
> > Like that but I would suggest: "double-click to reset this as it's default value".
> 
> I'm not sure what you mean by this. I'm suggesting a tooltip for when it's already at the default value, so double-click won't do anything.

I just understand the opposite for what you mean. So, just forget that.

elstoc:
> Asterisk is not fine for me. Asterisk means (at least in french but If I'm not wrong also in english) required setting/field

Yes I suppose it could be interpreted that way. That's the problem with using 'standard' characters - they're almost all loaded with some other implied meaning. Other possibilities: dagger †, double-dagger ‡

TurboGit:
I have a new version, I have used an asterisk for now, but we can decide the best character. The most important change in this version is that the character gets updated as soon as the value is changed. So we dynamically have an indication of a setting being default or not.

elstoc:
Works for me.

johnny-bit:
Due to family problems I can't work on my version...  but I still dislike that it's string shenenigans instead of css looking at a code. The problem mentioned by @elstoc is because the font isn't monospaced so the space padding for character doesn't work.

I thought about using additional element on the row and display on the element icon/character and show/hide that element via css class.

I like how the indicator is dynamic tho.

elstoc:
If we have a different css class for those that are at their default value, then something along the following lines would probably work (probably needs some tweaking though):
```
#preferences_notebook label.preference_default
{  border-left: 10px solid @plugin_bg_color;  }

#preferences_notebook label.preference_non_default
{  border-left: 10px solid @bg_color; }
```
gives...
Screenshot_2020-10-24_14-21-48

johnny-bit:
I like elstoc suggestion.

sob., 24 paź 2020, 15:23 użytkownik Chris Elston <notifications@github.com>
napisał:

TurboGit:
Sorry didn't see the message I was working on a new UI:

The star is now on it's own column and well visible. I find this good and probably good enough for now.

image

TurboGit:
@elstoc : Agreed I didn't like messing with the label, see new version, lot cleaner and makes the non-default well visible and can be tweaked by CSS.

TurboGit:
The character used is still an asterisk but we can use any UTF-8 character instead.

TurboGit:
Now seeing @elstoc proposal I still prefer my last proposal.

elstoc:
> Now seeing @elstoc proposal I still prefer my last proposal

I agree. Perhaps a bullet character or icon would be best.

elstoc:
Also, give the asterisk/bullet label a tooltip "this setting has been modified" so people don't raise "what does this bullet mean" queries

TurboGit:
@elstoc  :tooltip added.

TurboGit:
I have set a bullet instead of the asterisk, more visible.

elstoc:
Works fine for me. Thanks @TurboGit - definitely an improvement over the bold/italic text

TurboGit:
Let's merge at this point then!

PR #6631, opened 2020-10-25
Add LLVM 11 as supported version

amousset:
Fedora 33 uses LLVM 11.

This change appears to change the following message:

```
-- The following OPTIONAL packages have been found:

 * LLVM (required version >= 11)
```

while it should be 3.9, I don't understand why for now.


TurboGit (approved):
Looks ok, thanks !

PR #6640, opened 2020-10-26
thumbnail: record expose_again source ID and cancel timeout on destroy

### LINKED ISSUE #6620: Segmentation fault when scrolling down in lighttable
**Describe the bug**
Darktable crashes with segmentation fault when scrolling fast in the lighttable.

**To Reproduce**
1. Load a film roll with enough images so that you must scroll further down to see all images
2. Scrolling down _fast_ causes darktable to crash.

Note: Darktable does not crash when scrolling slowly enough - either by mouse wheel or pressing down arrow and wait until all thumbnails appear before scrolling again. Otherwise I can edit image in darkroom, export and all other functions all seem to work unaffected.

**Expected behavior**

Darktable should display thumbnails of the film roll as I scroll down the film roll.

**Screenshots**

Recording of crash scene: <https://youtu.be/hFJJpYedRf8>

Backtrace when directly run in gdb:
```
Thread 1 "darktable" received signal SIGSEGV, Segmentation fault.
0x00007ffff7d8919a in _thumb_expose_again (user_data=0x555559394450)
    at ../src/dtgtk/thumbnail.c:199
199     ../src/dtgtk/thumbnail.c: 没有那个文件或目录.
(gdb) bt
#0  0x00007ffff7d8919a in _thumb_expose_again (user_data=0x555559394450)
    at ../src/dtgtk/thumbnail.c:199
#1  0x00007ffff7826b94 in  () at /usr/lib/libglib-2.0.so.0
#2  0x00007ffff78274bd in g_main_context_dispatch () at /usr/lib/libglib-2.0.so.0
#3  0x00007ffff7829418 in  () at /usr/lib/libglib-2.0.so.0
#4  0x00007ffff782a3a3 in g_main_loop_run () at /usr/lib/libglib-2.0.so.0
#5  0x00007ffff725e5c5 in gtk_main () at /usr/lib/libgtk-3.so.0
#6  0x00007ffff7dc151d in dt_gui_gtk_run (gui=<optimized out>) at ../src/gui/gtk.c:1473
#7  0x00005555555550a0 in main (argc=<optimized out>, argv=<optimized out>) at ../src/main.c:93
```

Backtrace file produced by crash handler: 
[darktable_bt_X6W0S0.txt](https://github.com/darktable-org/darktable/files/5432279/darktable_bt_X6W0S0.txt)

 - Darktable Version: 3.2.1
 - OS: Linux - kernel 5.8.2
 - OpenCL: Yes
 - Graphics: Nvidia 1050Ti, driver version 450.66

**Additional context**

- Reproducible with previous versions?

This problem is not present in 3.0.2.

I built darktable myself. Darktable was originally compiled with various optimization, but I managed to reproduce this issue without any compiler flags. The backtrace is produced from `RelWithDebInfo` with no extra CFLAGS LDFLAGS compiled by gcc 9.

 - Is the issue still present using an empty/new config-dir?

Yes

cthbleachbit:
This resolves a race condition where a thumbnail widget may be destroyed before the expose_again timeout is triggered, causing a use-after-free problem. In particular, this commit fixes #6620.

### DIFF
--- a/src/dtgtk/thumbnail.c
+++ b/src/dtgtk/thumbnail.c
@@ -198,8 +198,8 @@ static void _image_get_infos(dt_thumbnail_t *thumb)
 static gboolean _thumb_expose_again(gpointer user_data)
 {
   dt_thumbnail_t *thumb = (dt_thumbnail_t *)user_data;
-  gpointer w_image = thumb->w_image;
   if(!thumb) return FALSE;
+  gpointer w_image = thumb->w_image;
   if(!w_image || !GTK_IS_WIDGET(w_image)) return FALSE;
 
   thumb->expose_again_timeout_id = 0;
--- a/src/dtgtk/thumbnail.c
+++ b/src/dtgtk/thumbnail.c
@@ -422,7 +422,7 @@ static gboolean _event_image_draw(GtkWidget *widget, cairo_t *cr, gpointer user_
       {
         // if the image is missing, we reload it again
         if(!thumb->expose_again_timeout_id)
-          thumb->expose_again_timeout_id = g_timeout_add(10, _thumb_expose_again, thumb);
+          thumb->expose_again_timeout_id = g_timeout_add(250, _thumb_expose_again, thumb);
 
         // we still draw the thumb to avoid flickering
         _thumb_draw_image(thumb, cr);
--- a/src/dtgtk/thumbnail.c
+++ b/src/dtgtk/thumbnail.c
@@ -197,10 +197,13 @@ static void _image_get_infos(dt_thumbnail_t *thumb)
 
 static gboolean _thumb_expose_again(gpointer user_data)
 {
-  if(!user_data || !GTK_IS_WIDGET(user_data)) return FALSE;
+  dt_thumbnail_t *thumb = (dt_thumbnail_t *)user_data;
+  gpointer w_image = thumb->w_image;
+  if(!thumb) return FALSE;
+  if(!w_image || !GTK_IS_WIDGET(w_image)) return FALSE;
 
-  GtkWidget *widget = (GtkWidget *)user_data;
-  gtk_widget_queue_draw(widget);
+  thumb->expose_again_timeout_id = 0;
+  gtk_widget_queue_draw((GtkWidget *)w_image);
   return FALSE;
 }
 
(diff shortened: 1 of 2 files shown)


src/dtgtk/thumbnail.c
```diff
@@ -197,10 +197,13 @@ static void _image_get_infos(dt_thumbnail_t *thumb)
 
 static gboolean _thumb_expose_again(gpointer user_data)
 {
-  if(!user_data || !GTK_IS_WIDGET(user_data)) return FALSE;
+  dt_thumbnail_t *thumb = (dt_thumbnail_t *)user_data;
+  gpointer w_image = thumb->w_image;
+  if(!thumb) return FALSE;
```
TurboGit, line 202: this test is too late as just above you use the pointer to access w_image.

I would keep:
`if(!user_data)) return FALSE;`
as the first line on this routine.

cthbleachbit: You are right. A fix is incoming.

src/dtgtk/thumbnail.c
```diff
@@ -418,7 +421,9 @@ static gboolean _event_image_draw(GtkWidget *widget, cairo_t *cr, gpointer user_
       if(res)
       {
         // if the image is missing, we reload it again
-        g_timeout_add(250, _thumb_expose_again, widget);
+        if(!thumb->expose_again_timeout_id)
+          thumb->expose_again_timeout_id = g_timeout_add(10, _thumb_expose_again, thumb);
```
TurboGit, line 425: The 10ms for the timeout seems a bit low for a thumb draw which is time consuming operation. If it was not ready then we certainly need some more time before trying again. Any explanation as to why to set it to 10 instead of 250 ?

cthbleachbit: My understanding here is that the draw signal handler would be called repeatedly (in `expose_again`) until the thumbnail is actually drawn. The previous implementation could have added multiple timeouts for `expose_again` scheduled at the same time depending on user input (i.e. when a user triggers a draw event). Now the rule "at any time, a thumbnail can only have at most 1 expose_again scheduled" is enforced, so I think it is safe to check more frequently if the drawing is complete. Keep using 250 wouldn't cause any errors though.

TurboGit: @cthbleachbit : sure I understand that, but calling more time without actually doing something is a waste of CPU time :) I would certainly restore to 250 except if there is a good reason. So please restore the value and if we have issue with that (too slow redrawing or whatever) let's re-discuss this. Works for you ?

cthbleachbit: I admit that 10ms is probably too short to account for all I/O and stuff. However, the exact reason I changed it from 250 to 10 is that it took perceivably longer for the thumbnails to show up than before. I'll restore 250 for now, and see if there are some other confounding variables in there.

TurboGit (approved):
Looks good to me now, thanks!

PR #6653, opened 2020-10-27
generate-prefs: fix spacing of miscellaneous tab

elstoc:
Makes misc tab consistent with other tabs

Before 

Screenshot_2020-10-27_12-20-28

After

Screenshot_2020-10-27_12-20-54


TurboGit:
Thanks !

PR #6664, opened 2020-10-28
po/performance mode

TurboGit:


parafin:
Looks correct, but one thing to check is how PPD affects this code. With PPD == 2, 100% zoom is 1:1 real pixels or 1:1 dots sized 2x2 pixels? I mean FAST should be used only when scale is more than 1:1 real pixels.

jenshannoschwalm:
Can't test but i think that for PPD==2 100% zoom we will have 1:1 real pixels for thumbs, not sure about darkroom though. Otherwise yes. BTW for current linux cairo FAST is the same as NEAREST.

parafin:
Another question would be is how should 100% zoom work on HighDPI. Probably it should be 1:1 real pixels.

TurboGit:
@parafin : you have a point here, we want to disable any cairo filter when above 1:1 and so taking into account the ppd.

New patch set pushed which works for ppd == 1 and ppd = 2 (tested).

TurboGit:
For HiDPI screen I suppose my last patch does cover this case too. I can't test as I don't have an HiDPI screen.

jenshannoschwalm:
I you are working under gnome there is the system setting available doing exactly this, switch to system ppd==2

TurboGit:
> I you are working under gnome there is the system setting available doing exactly this, switch to system ppd==2

Where is this setting ? Is that the scale 200% in the screen preferences ?

matt-maguire:
Just below the navigation inage at the top left of the darkroom view, there is a little dropdown menu where you can set the scale factor for the preview image -- setting this to 200% provkes the sluggish behaviour.

jenshannoschwalm:
Yes.

TurboGit:
Perfect! Tested and all seems ok for HiDPI. I'll merge. Thanks.

PR #6673, opened 2020-10-29
Final part of init_gui/reload/update cleanup

dterrahe:
This finishes off #6240 

The biggest change is to make sure the iop gui->widget is added to the side panel _before_ reload_defaults is called. This means reload_defaults can always count on all widgets having been shown/realised (by a call to show_all). So the behavior is now predictable; no longer are workarounds required to make things that work after an image change also work the 1st time. So I've been able to remove some duplicated calls that were just put in "to make sure".

gui_init is now more light weight. It tries to avoid showing/realising widgets if not necessary; the show_all when the complete widget is added to the panel takes care of that. This makes #6648 more efficient.

There are also other random fixes/cleanups that I came across while I did this work.

The commits are grouped by theme, but I haven't tested them separately for subtly dependencies, so I wouldn't recommend cherry-picking.

### DIFF
--- a/src/bauhaus/bauhaus.c
+++ b/src/bauhaus/bauhaus.c
@@ -583,7 +583,7 @@ void dt_bauhaus_init()
   darktable.bauhaus->key_val = NULL;
   memset(darktable.bauhaus->key_history, 0, sizeof(darktable.bauhaus->key_history));
 
-  darktable.bauhaus->skip_accel = 0;
+  darktable.bauhaus->skip_accel = 1;
 
   // this easily gets keyboard input:
   // darktable.bauhaus->popup_window = gtk_window_new(GTK_WINDOW_TOPLEVEL);
--- a/src/develop/blend_gui.c
+++ b/src/develop/blend_gui.c
@@ -2442,9 +2442,6 @@ void dt_iop_gui_init_blending(GtkWidget *iopw, dt_iop_module_t *module)
     gtk_widget_set_name(GTK_WIDGET(iopw), "blending-wrapper");
 
     bd->blend_inited = 1;
-    gtk_widget_queue_draw(GTK_WIDGET(iopw));
-    dt_iop_gui_update_blending(module);
-
   }
 }
 
--- a/src/develop/develop.c
+++ b/src/develop/develop.c
@@ -1077,15 +1077,14 @@ void dt_dev_reload_history_items(dt_develop_t *dev)
     {
       if(!dt_iop_is_hidden(module) && !module->expander)
       {
-        ++darktable.gui->reset;
-        module->gui_init(module);
-        dt_iop_reload_defaults(module);
-        --darktable.gui->reset;
+        dt_iop_gui_init(module);
 
         /* add module to right panel */
         GtkWidget *expander = dt_iop_gui_get_expander(module);
         dt_ui_container_add_widget(darktable.gui->ui, DT_UI_CONTAINER_PANEL_RIGHT_CENTER, expander);
         dt_iop_gui_set_expanded(module, TRUE, FALSE);
+
+        dt_iop_reload_defaults(module);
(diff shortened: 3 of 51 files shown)


dterrahe:
There are still a few widgets that get realised/shown in gui_init, but I brought it down from 113 to 34. The remaining ones fall mostly in 3 categories;
- notebook tabs. As per gtk documentation, a tab element needs to be shown in order for it to be selectable, so in cases where the tab selection gets restored to a saved state, the embedded box has to be shown first.
- colorswatches
- text-buttons
I haven't further investigation why these last two get realised during creation or whether it is avoidable.

I used this code to track down these cases:
```
void _print_visible_children(GtkWidget *widget, gpointer module)
{
  if(gtk_widget_get_visible(widget))
  {
    char *path = gtk_widget_path_to_string(gtk_widget_get_path(widget));
    fprintf(stderr, "%s: %s\n", ((dt_iop_module_t *)module)->op, path);
    g_free(path);
  }

  if(GTK_IS_CONTAINER(widget))
    gtk_container_foreach(GTK_CONTAINER(widget), _print_visible_children, module);
}

void dt_iop_gui_init(dt_iop_module_t *module)
{
  ++darktable.gui->reset;
  if(module->gui_init) module->gui_init(module);
  --darktable.gui->reset;

  _print_visible_children(module->widget, module);
}
```
Also, the clipping module does a lot of initialisation in gui_init that could possibly be postponed, or skipped during acceleration creation (after #6648 get merged).

dterrahe:
@TurboGit resolved merge conflicts with #6648

This could improve startup speed because the gui_init/cleanup to register shortcuts is now more lean (fewer widgets are "shown"/realised). Also, if bauhaus set_label is called outside of gui_init (with a changed name), it doesn't try to connect the widget to a non-existing accelerator anymore. This would otherwise happen in colorbalance, filmicrgb, lens and negadoctor.

A few small memory leaks were fixed on the way.

I noticed that the per-module "fusion" slider used to have a non-standard translation in "accel" that was removed as a result of #6648. Therefore I implemented part of the suggestion I made there to support Q_ style accelerator contexts in accelerators. Since this (for now) is a one-off, I used the expression `NC_("accel","fusion") ? "accel|fusion" : ""` to mark for translation with the correct context. If needed, this should be more correctly done by defining a macro `NQ_(String) (String)` that gets processed by xgettext by adding `--keyword=NQ,1g` . I haven't tested this yet, but if we get this set up, we could actually directly add `--keyword=dt_bauhaus_widget_set_label,3g` and not require any N_ translation marking for that function call at all anymore! For introspection to support this kind of context marking fully, the perl code should use NQ_ instead of N_ and any users of introspection (_from_params, history diff) should use Q_ instead of _.

TurboGit:
Is this expected to be plan for 3.4 or should it be in a later minor release ?

dterrahe:
I think it is ready, but if you get squeezed for testing time, I don't think there's anything crucial in here. Some minor bug fixes/memory leaks and (probably small) performance improvements, apart from code cleanup/structure/clarity.

The only regression in #6648 that might need fixing is the translation of the fusion (blending) accelerator. That now shows up as "fusion d'exposition" in the shortcut settings; once that hits the translators, I presume it will be more work later on putting the "accel" translation back. So you could consider cherry picking 5adb238 if you don't want to merge the whole PR.

TurboGit:
Thanks for the feedback, I'll see if I have enough time for reviewing otherwise I'll pickup the commit to fix translation.

TurboGit:
Ok, works for me. No issue found. Let's go for 3.4.

dterrahe:
Thank you! For _all_ your work.

I saw your 15bef3b removing some gui_update stumps. I left those in place intentionally, because there's no default gui_update and sometimes it gets called unconditionally so NULL crashes.

I'll submit a quick PR to add a check before calling, which is the correct way.

PR #6679, opened 2020-10-29
filmic GUI : add latitude bounds in dashed line

aurelienpierre:
Screenshot_20201029_223615-1

Minor tweak to filmic GUI, I'm not sure it adds benefit, and I'm not set on that. Merge if you like, ditch if you don't.


elstoc:
Better without I think

matt-maguire:
You think dashes make it look too busy? maybe use a different color instead of dashes? worry is what happens with different themes

aurelienpierre:
It's not just dashes, it's 2 additional lines that may overlay with EV lines.

aurelienpierre:
To be clear, this shows the range that is sent to display almost as-is (little tone compression), inside the dashed lines, apart from the range that gets compressed a lot.

Issue #6568, opened 2020-10-18
Ignoring "allow upscaling" (?)


esq4:
If "allow upscaling"  set to “yes” darktable not performing an upscaling step in case the user defined maximum width and height exceed the original dimensions. The exported image's dimensions not exceeding the original ones (after or without cropping). 

Is this how it should be and something I did not understand?

TurboGit:
No, it is not supposed to work this way. The image should have at the end the size request by user even if it is originally smaller.

TurboGit:
And this works on my side ! Can you double check ?

esq4:
I checked it again. Addition: one of the dimensions must be 0.

esq4:
Found where, but did not figure out how to fix it.
```
  if(dt_dev_distort_backtransform_plus(&dev, &pipe, 0.f, DT_DEV_TRANSFORM_DIR_ALL, origin, 1))
  {
    if((width == 0) && exact_size)
      width = pipe.processed_width;
    if((height == 0) && exact_size)
      height = pipe.processed_height;

>     scale = fmin(width >  0 ? fmin((double)width / (double)pipe.processed_width, max_scale) : max_scale,
>                  height > 0 ? fmin((double)height / (double)pipe.processed_height, max_scale) : max_scale);

```

TurboGit:
That's not a bug then. A 0 means 'image size' so '0 x 100000' even with upscale = TRUE won't create a image having a width of 100000.

esq4:
I understood the math. I don't understand the logic of the UI.
For example, I want to limit the width of an image. So what difference does it make if it is larger or smaller than the original? Why is it enough to set one size for cropping, and both are necessary for enlargement? Not unambiguous ...
Maybe it makes sense to make these settings the same for both cases?

esq4:
It may be better to give the combination "larger_maximum_size x 0" the same meaning as "smaller_maximum_size x 0" and not to force the user to do unnecessary work?

TurboGit:
>  Why is it enough to set one size for cropping, and both are necessary for enlargement? Not unambiguous ...

I get your point and agree, if we can do better let's go for it.

parafin:
Right now 0 basically mean "original picture size" if I understand correctly the logic. It can be re-defined to mean "other export dimension * picture aspect ratio", then it will work for enlargement too without breaking no-upscale case. 0x0 though still has to be defined as a special case of "original picture size".

esq4:
```
[...]

PR #6681, opened 2020-10-30
Ignoring "allow upscaling" #6568 try fix

esq4:


esq4:
try fix #6568

jenshannoschwalm:
I don't think - just from memory - that simply removing that part is correct. We have to make sure that the thumbnails are fine.

esq4:
This code actually limits the maximum scale to 1. Perhaps this is useful for thumbnails. I propose to limit this code to such a condition:

```
  if(dt_dev_distort_backtransform_plus(&dev, &pipe, 0.f, DT_DEV_TRANSFORM_DIR_ALL, origin, 1))
  {
    if((width == 0) && thumbnail_export)
      width = pipe.processed_width;
    if((height == 0) && thumbnail_export)
      height = pipe.processed_height;

    scale = fmin(width >  0 ? fmin((double)width / (double)pipe.processed_width, max_scale) : max_scale,
                 height > 0 ? fmin((double)height / (double)pipe.processed_height, max_scale) : max_scale);

```

TurboGit:
Just a thinking... When a single value is 0 would it works to just set the other value to any very large value ?

TurboGit:
Just to follow up about my thinking. All is working fine when both values (width & height) are not 0. So if one is 0 and the other has a specific value the actual crop will be based on the given value. Fine. So if in the export module and only of of both values are 0 we force the other to a very large number it should work as expected regarding the upscaling setting, no?

esq4:
If, for example, width is not very large and height = 0, then we choose the smaller of "not_large_scale_width" and max_scale. If the height is very large, then we again choose the smaller of "scale_width" and (max_scale or "very_large_scale_width"). The result will always be "not_very_large_scale_width".

If in this example the width would be equal to a large value, then we will get "large_scale_width", which can be bad for thumbnails. And the suggested limitation of scale to 1 for thumbnails will be useful here.

TurboGit:
Not sure to understand, I am proposing to do that in the export module it should not impact the thumbnails.

esq4:
I'm talking about the same thing.


`  if(dt_dev_distort_backtransform_plus(&dev, &pipe, 0.f, DT_DEV_TRANSFORM_DIR_ALL, origin, 1))`
`  {`
`    if((width == 0) && ` **thumbnail_export**`)`
`      width = pipe.processed_width;`
`    if((height == 0) && `**thumbnail_export**`)`
`      height = pipe.processed_height;`

esq4:
> If, for example, width is not very large and height = 0, then we choose the smaller of "not_large_scale_width" and max_scale. If the height is very large, then we again choose the smaller of "scale_width" and (max_scale or "very_large_scale_width"). The result will always be "not_very_large_scale_width".
> 
> If in this example the width would be equal to a large value, then we will get "large_scale_width", which can be bad for thumbnails. And the suggested limitation of scale to 1 for thumbnails will be useful here.

Here I parsed the work of this line:
```
    scale = fmin(width >  0 ? fmin((double)width / (double)pipe.processed_width, max_scale) : max_scale,
                 height > 0 ? fmin((double)height / (double)pipe.processed_height, max_scale) : max_scale);
```

TurboGit:
@esq4 : No, I'm speaking of doing that in export.c look at line 279 in function _export_button_clicked.

We get:
```
  const uint32_t max_width = dt_conf_get_int(CONFIG_PREFIX "width");
  const uint32_t max_height = dt_conf_get_int(CONFIG_PREFIX "height");
```

And we pass those values into:

```
  dt_control_export(list, max_width, max_height, format_index, storage_index, high_quality, upscale, export_masks,
                    style, style_append, icc_type, icc_filename, icc_intent, d->metadata_export);
```

Since all is working as expected when both values are not 0 and when only one is zero it means don't care the other will be the constraint...; I'm wondering if passing to dt_control_export a large value instead of 0 will just work.

TurboGit:
And I'm proposing to do that only in scaling mode of course.

esq4:
> I'm wondering if passing to dt_control_export a large value instead of 0 will just work.

Yes, it works within the signed int.

esq4:
> And I'm proposing to do that only in scaling mode of course.

If you are writing about the changes I propose in imageio.c, then scaling mode performs an upscaling without them. The problem is with size mode.
Or I didn't understand you. :(

TurboGit:
> The problem is with size mode.

Right, I meant in size mode, sorry for the confusion.

TurboGit:
> Yes, it works within the signed int.

Can you clarify ? I'm not sure to understand this... "signed int" ?

esq4:
> Can you clarify ? I'm not sure to understand this... "signed int" ?

To about 2147483647.

TurboGit:
> To about 2147483647.

You mean that my proposal of setting the width to some high value works up to 2147483647, right?

That sounds perfect as I don't think we need to go to this image size at all ! Maybe we can use some insane value like 500000 as width or height ?

esq4:
I understand what you lead. I started with this option, but discarded it with all my might. 
After all, this is not beautiful: to correct the logic of work by substitution of data, and in a different place.

TurboGit:
But, doing it this way, we are sure that this does not breaks the thumbs generation.

And again the exact_size was added to ensure that the expected output size was most (if not all) the time equal to what the user has put in the export width & height field. We had report about a diff of one pixels which was bad. So at the end I'm really reluctant to change this especially so close of a release. Have you looked at the commit I pointed about the work leading to this code ?

jenshannoschwalm:
I think the suggestion of inserting an insane high value should be safe.

It is mostly important *not* to have thumbs increased in size as when reading them back again that might lead to out of memory access.

TurboGit:
BTW, @esq4 have you passed all the integration tests ? IIRC some were failings due to the one pixel off when exporting after rotation for example.

esq4:
> BTW, @esq4 have you passed all the integration tests ? IIRC some were failings due to the one pixel off when exporting after rotation for example.

Here are the results:
```
a@debian:~/integration$ ./run.sh
Test 0000-nop
      Image mire1.cr2
      CPU & GPU version differ by 24225 pixels
      Max  dE         0.0000
  OK

Test 0001-exposure
      Image mire1.cr2
      CPU & GPU version differ by 25699 pixels
      Max  dE         0.0000
  OK

Test 0002-local-contrast
      Image mire1.cr2
      CPU & GPU version differ by 1.75081e+06 pixels
      Max  dE         0.0000
  OK

Test 0003-denoise-bilateral
      Image mire1.cr2
      CPU & GPU version differ by 29464 pixels
      Max  dE         1.6894
  OK

Test 0004-masks
      Image mire1.cr2
      CPU & GPU version differ by 23188 pixels
      Max  dE         0.0000
  OK

Test 0005-orientation-minus-90
      Image mire1.cr2
      CPU & GPU version differ by 24216 pixels
      Max  dE         0.0000
  OK

Test 0006-orientation-plus-90
      Image mire1.cr2
      CPU & GPU version differ by 23974 pixels
      Max  dE         0.0000
  OK

Test 0007-orientation-180
      Image mire1.cr2
      CPU & GPU version differ by 23987 pixels
      Max  dE         0.0000
  OK

Test 0008-flip-h
      Image mire1.cr2
      CPU & GPU version differ by 23121 pixels
      Max  dE         0.0000
  OK

Test 0009-flip-v
      Image mire1.cr2
      CPU & GPU version differ by 23126 pixels
      Max  dE         0.0000
  OK

Test 0010-flip-hv
      Image mire1.cr2
      CPU & GPU version differ by 23354 pixels
      Max  dE         0.0000
  OK

Test 0011-gdnd-90
      Image mire1.cr2
      CPU & GPU version differ by 22626 pixels
      Max  dE         0.0000
  OK

Test 0012-gdnd-45
      Image mire1.cr2
      CPU & GPU version differ by 22818 pixels
      Max  dE         0.0000
  OK

Test 0013-denoiseprofile-wavelets
      Image mire1.cr2
      CPU & GPU version differ by 60172 pixels
      Max  dE         1.0369
  OK

Test 0014-filmic-rgb
      Image mire1.cr2
      CPU & GPU version differ by 29437 pixels
      Max  dE         1.1610
  OK

Test 0015-shadhi-bilateral
      Image mire1.cr2
      CPU & GPU version differ by 31738 pixels
      Max  dE         1.0228
  OK

Test 0016-lowpass-bilateral
      Image mire1.cr2
      CPU & GPU version differ by 34164 pixels
      Max  dE         6.5180
      Mean dE         0.0001
      Count above max 29
  FAILS: image visually changed
         (662 pixels changed)

Test 0017-monochrome
      Image mire1.cr2
      CPU & GPU version differ by 7490 pixels
      Max  dE         0.8737
  OK

Test 0018-perspective-corr
      Image mire1.cr2
      CPU & GPU version differ by 35804 pixels
      Max  dE         0.8924
  OK

Test 0019-color-mapping
      Image mire1.cr2
      CPU & GPU version differ by 48050 pixels
      Max  dE         7.9731
      Mean dE         0.0026
      Count above max 236
  FAILS: image visually changed
         (13124 pixels changed)

Test 0020-denoise-nlmeans
      Image mire1.cr2
      CPU & GPU version differ by 985913 pixels
      Max  dE         1.5742
  OK

Test 0021-retouch-wavelets
      Image mire1.cr2
      CPU & GPU version differ by 32487 pixels
      Max  dE         1.2316
  OK

Test 0022-color-zones
      Image mire1.cr2
      CPU & GPU version differ by 34568 pixels
      Max  dE         1.0534
  OK

Test 0023-channelmixer
      Image mire1.cr2
      CPU & GPU version differ by 34623 pixels
      Max  dE         1.4635
  OK

Test 0024-contrast-equalizer
      Image mire1.cr2
      CPU & GPU version differ by 783999 pixels
      Max  dE         1.0133
  OK

Test 0025-exposure-guided-filter
      Image mire1.cr2
      CPU & GPU version differ by 25637 pixels
      Max  dE         0.0000
  OK

Test 0026-haze-removal
      Image mire1.cr2
      CPU & GPU version differ by 550659 pixels
      Max  dE         1.7114
  OK

Test 0027-denoiseprofile-nlmeans
      Image mire1.cr2
      CPU & GPU version differ by 613983 pixels
      Max  dE         1.1315
  OK


Total test 28
Errors     2
```

TurboGit (changes requested):
> I think the suggestion of inserting an insane high value should be safe.

We can also just insert a value 100 times bigger than the other limit, I don't think having an image with an aspect ration above (or even near) 1:100 makes any sense. So it should be safe.

TurboGit:
Ok, see 80ff46c3ef352a8a237afca7116508f42cc0508a.

PR #6685, opened 2020-10-30
zero out node past end of curve on delete

dterrahe:
When shortening a curve, the leftover rubbish shows up in the history diff and incorrectly causes identical presets not to be recognised. For example, in _basecurve_ apply the first preset. The preset menu will indicate in bold that it is active. Now "accidentally" add a node by clicking and dragging on the curve. Correct the mistake by right-clicking on the new node. Result is identical to the preset but is no longer matched.

### DIFF
--- a/src/iop/basecurve.c
+++ b/src/iop/basecurve.c
@@ -1938,6 +1938,7 @@ static gboolean dt_iop_basecurve_button_press(GtkWidget *widget, GdkEventButton 
       basecurve[k].x = basecurve[k + 1].x;
       basecurve[k].y = basecurve[k + 1].y;
     }
+    basecurve[nodes - 1].x = basecurve[nodes - 1].y = 0;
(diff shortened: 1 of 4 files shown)


parafin:
I suspect this won't fix already existing history stacks and presets? But it's probably doesn't matter much.

dterrahe:
> I suspect this won't fix already existing history stacks and presets? But it probably doesn't matter much.

That's correct.

If more metadata was added to introspection so that it was known which field indicates the number of items that are used in an array, then we could do scrubbing on read (I suggested this elsewhere too for other reasons). But currently only the module itself understands these relationships and doing this in commit_params would be incorrect (it shouldn't modify params?) and too late? Also inefficient as would be done each time.

I'd categorise this as _mostly_ cosmetic and an easy clean but not worth too much extra effort.

TurboGit (approved):
Looks good to me ! Thanks.

PR #6687, opened 2020-10-30
ctrl+click iop reset button to apply auto-presets

### LINKED ISSUE #6682: user-defined default presets
Auto-apply presets are great but it would be really useful to have them automatically reapply if resetting a module's parameters or going back in the history stack.

As a suggested solution for this, for both iop and lib modules, allow the user to define a preset named "default" which, if present, will be applied by default for new images or reapplied automatically when the user clicks on the 'reset parameters' button.

elstoc:
`Ctrl+click` on an iop module reset button will reapply any automatic presets to that module. If no automatic presets are found, the module parameters will be reset to their default values (i.e. same as left-clicking on the reset button).

Resolves #6682

### DIFF
--- a/src/gui/presets.c
+++ b/src/gui/presets.c
@@ -718,9 +718,8 @@ static void menuitem_new_preset(GtkMenuItem *menuitem, dt_iop_module_t *module)
   edit_preset(_("new preset"), module);
 }
 
-static void menuitem_pick_preset(GtkMenuItem *menuitem, dt_iop_module_t *module)
+static void apply_preset(const gchar* name, dt_iop_module_t *module)
 {
-  gchar *name = g_object_get_data(G_OBJECT(menuitem), "dt-preset-name");
   sqlite3_stmt *stmt;
   DT_DEBUG_SQLITE3_PREPARE_V2
     (dt_database_get(darktable.db),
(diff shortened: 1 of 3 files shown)


dterrahe:
Great additional functionality.

Looking at the additional work required to achieve this makes me wonder if there's room for a little refactoring here. I always worry when a block of mostly/partly duplicate code appears, especially in a different file, that future changes/fixes won't get applied here. Then it could start to diverge from what happens originally at import, or get buggy.

elstoc:
> makes me wonder if there's room for a little refactoring here

I would tend to agree with you and I did consider it, but the techniques for applying the presets in the two sections of code where this appears are so different that I'm not sure I know the best way to go about it. The other place this exists performs a direct insert into the history and sits alongside other bits of code that auto-enable modules.

So I would prefer to merge this for 3.4 (assuming it's not too late) since this change is fairly low risk, and consider refactor for the next release.

TurboGit:
> So I would prefer to merge this for 3.4 (assuming it's not too late) since this change is fairly low risk, and consider refactor for the next release.

I'm always uneasy about this, but I'm ok to merge in 3.4 just be sure to open an issue for the refactoring otherwise it won't happen I fear :)

src/gui/presets.c
```diff
@@ -777,6 +776,82 @@ static void menuitem_pick_preset(GtkMenuItem *menuitem, dt_iop_module_t *module)
...
+{
+  gchar *name = g_object_get_data(G_OBJECT(menuitem), "dt-preset-name");
+  apply_preset(name, module);
+}
+
+gboolean dt_gui_presets_autoapply_for_module(dt_iop_module_t *module)
+{
+  dt_image_t *image = &module->dev->image_storage;
+
+  gchar *workflow = dt_conf_get_string("plugins/darkroom/workflow");
```
TurboGit, line 789: should be freed at some point.

TurboGit (changes requested):
A memory leak to fix.

elstoc:
> A memory leak to fix.

Fixed

>  just be sure to open an issue for the refactoring

Will do

TurboGit (approved):
Works for me now, thanks !

Issue #3085, opened 2019-10-06
GUI enhancement: keep scroll position per module group


junkyardsparkle:
**Is your feature request related to a problem? Please describe.**
When using the module group buttons in darkroom to switch between different groups, adjusting the scrolling for optimal display of a module interface will replace a previous such adjustment in another group, requiring constant re-adjustment when frequently switching between two modules in different groups.

**Describe the solution you'd like**
If the scroll position could be retained *per group*, the behavior would be more similar to "tabs", which these buttons more or less perform the role of. Using this method of switching between modules in different groups is IMHO less tedious and haphazard than doing large amounts of scrolling within a group, particularly on moderate-size displays.

**Describe alternatives you've considered**
I have tried different approaches involving the favorites group and the active modules group, but find myself increasingly favoring switching between groups when using the newer 2.7 GUI.

**Additional context**
Also adding the ability to bind keyboard shortcuts to the group button actions would enhance this enhancement. :-)

junkyardsparkle:
Bump. This is still a valid feature request as far as I'm concerned.

junkyardsparkle:
This is somewhat less relevant now with the expanded group configuration ability coming in 3.4; I'm leaving it open pending discussion in #5129 about module and group dynamics.

Nilvus:
@junkyardsparkle: since #5129 issue you link this one to, had been fixed and closed, what about this one? I think actual master has better behavior now. Do you think this issue need something to be solved? If yes, could you precise it as you said in your last comment that was less relevant now (in october)? If no, maybe we could close this issue.

PR #6688, opened 2020-10-31
switch notebook page to show accelerated widget

dterrahe:
When a bauhaus widget is changed using shortcuts/dynamic mouse scroll, a toast shows up with the new value if the widget isn't visible. However, if the module is expanded, but the widget is on a notebook tab that is not currently shown, it is still considered "visible" so the toast isn't shown. This PR switches (in most/all?) cases the notebook to show the changing widget.

Widgets are still treated as "visible" when they are scrolled out of sight or covered.

### DIFF
--- a/src/gui/accelerators.c
+++ b/src/gui/accelerators.c
@@ -641,7 +641,14 @@ void dt_accel_widget_toast(GtkWidget *widget)
 
     g_free(text);
   }
-
+  else
+  {
+    GtkWidget *box = gtk_widget_get_parent(widget);
+    GtkWidget *notebook = gtk_widget_get_parent(box);
+    if(GTK_IS_NOTEBOOK(notebook))
+      gtk_notebook_set_current_page(GTK_NOTEBOOK(notebook),
+                                    gtk_notebook_page_num(GTK_NOTEBOOK(notebook), box));
+  }
 }
 
 float dt_accel_get_slider_scale_multiplier()


elstoc:
Personally, now that it's fairly inconspicuous, I would prefer to just show the toast message always. Even if you have the module open it can sometimes take a few seconds to visually locate the slider that you're updating.

dterrahe:
Could do that or I could flip this PR around and expand the visibility check to special case notebooks and show the toast when the widget is on another page. Maybe I should run a survey here. Does github support? (if not, why?!)

The more I deal with notebooks, the more they seem like a patched implementation bogged down by legacy compatibility requirements. But I still like the look of them :-)

dterrahe:
> Maybe I should run a survey here.

BTW the thought behind implementing it the current way was that if you change one parameter on a notebook tab, you might be interested to see/check the others on that tab as well. We don't have shortcuts to change tab pages (I'm thinking about it, but it is less trivial) so this could be an alternative way to achieve the same thing. The side effect for some might be the main effect.

elstoc:
I don't have an issue with this change. If the module is visible it makes sense to switch tabs. I just now look for the toast message whenever I use a shortcut and do a double-take when I don't see it, which just interrupts my flow a little.

dterrahe:
_I_ would be supportive of that change too. It would make (some form of) this PR less needed, but I think they're both useful.

Always showing the toast when using shortcuts might be a more visually invasive change, so I wanted to avoid the discussion by not including it in this PR, but I suspect the venn of people who use shortcuts but who hate toasts is very narrow. Let's resist the urge to require an extra config option for this one!

In my original implementation of value-change toasts in bauhaus, the visibility check was important, because otherwise the toasts would _always_ show, even when the slider was moved directly by mouse. But since quite quickly it only got called from accelerators, it may have been a misjudgment to carry over the visibility check.

TurboGit:
I agree with @elstoc, the dynamic shortcut should not change tab, the whole point is really to change a value in a module without having to show this module at all and we want to stay on the current selected module.

I have a dynamic shortcut on exposure and sometime when working on another module I do find that the exposure need to be adjusted (like when I'm on filmicrgb) and certainly do not want the tab to be switched at this point.

elstoc:
> I agree with @elstoc, the dynamic shortcut should not change tab

That wasn't what I was saying. If a control is being adjusted on a module that's currently expanded I don't mind automatically switching tabs within that module. Having said that I can see the use case for perhaps tweaking filmic's black level while modifying the settings in another filmic tab, in which case maybe this is not desirable. @TurboGit I wonder from your response if you're perhaps misunderstanding the use of the word tab here, since you refer to exposure which doesn't have a tabbed interface?

TurboGit:
@elstoc : Indeed I understood that "tab" was referring to the "module group tab" :) So my comment can be discarded, sorry.

elstoc:
@dterrahe: Do you want to include the always-show-toast change here as well, since it seems fairly uncontroversial? I could raise a separate PR but it's the same function as you're already amending.

dterrahe:
> @dterrahe: Do you want to include the always-show-toast change here as well, since it seems fairly uncontroversial?

Happy to do so, but you've just made _this_ change more controversial, so maybe you don't want to tie them together? ;-) I'll wait for a consensus to settle.

> Having said that I can see the use case for perhaps tweaking filmic's black level while modifying the settings in another filmic tab, in which case maybe this is not desirable.

I really didn't think a config option was needed here, but you've almost convinced me otherwise. Do you think there's a group of users who use shortcuts _specifically_ because this allows them to adjust controls on two different notebook pages at the same time, one by direct slider manipulation, the other via (dynamic) shortcuts? I definitely would be screwing over that user, unintentionally (and with a heavy heart, because it is such a smart way to (ab)use the toys you're given).

elstoc:
I could take or leave this change and I'd rather avoid another darkroom config option (we're going to have to create a new tab soon if we're not careful). If we think people might find it useful not to change tabs maybe we should drop this one and replace it with the always-show-toast change instead.

dterrahe:
1. show toast if invisible _or_ on different notebook page. (<-this is just an uncontroversial bug fix)
2. show toast if invisible. switch notebook page to show. (this PR current state)
3. always show toast, don't switch notebook page
4. always show toast _and_ switch notebook page
5. config switch between 1 and 2
6 config switch between 3 and 4

(ok, this is just always show toast or not & switch no/yes/config)

dterrahe:
> If we think people might find it useful not to change tabs maybe we should drop this one and replace it with the always-show-toast change instead.

This PR is not _just_ to see the change somewhere. There probably are people who use shortcuts but have the module open so they can monitor the change there. They would probably want to see it in context of the correct page.

Plus, as I described above, this implicitly gives you a shortcut to change notebook tabs, which I think is independently useful.

elstoc:
I'm fine with option 3 or 4. I think we should just choose one and introduce a config option if enough people object.

dterrahe:
OK, let's see what @TurboGit thinks.

I should have pointed out that I'm happy for this to go into 3.6 (or 3.4.2)
@elstoc you might feel differently about always showing toasts, so still worth submitting that. It is just the removal of a condition. Or the whole _if_, because I suspect the darktable.gui->reset test isn't required either, but obviously that would require more careful testing (for "corner cases" if there are any).

elstoc:
Ok that's fine I do have a PR ready to go :)

junkyardsparkle:
> This PR is not _just_ to see the change somewhere. There probably are people who use shortcuts but have the module open so they can monitor the change there. They would probably want to see it in context of the correct page.

Just to present a data point, my use of the dynamic shortcuts (so far) is weighted very heavily towards making an exposure adjustment while "physically" engaging the controls of a different module (eg. filmic RGB) without being pulled away from that activity. Thus, I would probably want to disable this feature, if I'm understanding it correctly.

> Plus, as I described above, this implicitly gives you a shortcut to change notebook tabs, which I think is independently useful.

That does seem useful. For various reasons, I don't really think of the module groups as orthodox "tabs", but that's another disctussion, started already in #5129, #3085...

elstoc:
> Just to present a data point, my use of the dynamic shortcuts (so far) is weighted very heavily towards making an exposure adjustment while "physically" engaging the controls of a different module (eg. filmic RGB) without being pulled away from that activity. Thus, I would probably want to disable this feature, if I'm understanding it correctly.

I'm not sure you are understanding it correctly. This is how I primarily use dynamic shortcuts as well. I think this change is more that if you were adjusting, say, the black slider of filmic RGB with a shortcut while on the look tab, the filmic module would switch to the scene tab automatically to show the slider being adjusted. Adjusting exposure while in the filmic module should continue to work the same as it does today - with #6690 the toast message will appear even if the exposure module is also visible and nothing will change on the filmic module.

> For various reasons, I don't really think of the module groups as orthodox "tabs",

I think you are under the same misunderstanding as @TurboGit was - this is about tabbed interfaces within individual module and not about module groups.

junkyardsparkle:
> I think you are under the same misunderstanding as @TurboGit was - this is about tabbed interfaces within individual modules and not about module groups.

Bingo! Well, now that it's been spelled out that explicitly, hopefully I'll be the last person to get confused about it. Thanks!

TurboGit:
> This PR switches (in most/all?) cases the notebook to show the changing widget.

Doesn't work for me, or I still miss the point of this PR :+1: 

- I have added a dynamic shortcut for "light shadows" in tone equalizer (simple tab)
- Using it when the module is collapse I see the toast
- Using it when the module is expanded and on the simple tab, I see the toast
- Using it when the module is expanded and I'm on the masque tab, I see the toast **but the simple tab is not made visible**

I think this last point should work, no?

dterrahe:
#6690 broke this one. I didn't think it was uncontroversial enough to go into 3.4, so didn't revise it yet. Just removing the else in line 644 should do it (i.e. making the addition unconditional).

TurboGit:
Ok, so let's just postpone this for later or never depending on more feedback. No need to rush with that in 3.4.

dterrahe:
Yeah, this is no longer a bug fix, now that the toast is always shown.

dterrahe:
Closing this as it is not/no longer a desired feature.

PR #6690, opened 2020-10-31
accelerators: always show toast message when adjusting controls

elstoc:
As discussed in #6688, this changes the keyboard accelerators so that slider and combobox adjustments always show a toast message regardless of whether the module being adjusted is visible.

### DIFF
--- a/src/gui/accelerators.c
+++ b/src/gui/accelerators.c
@@ -596,16 +596,13 @@ void dt_accel_widget_toast(GtkWidget *widget)
 {
   dt_bauhaus_widget_t *w = (dt_bauhaus_widget_t *)DT_BAUHAUS_WIDGET(widget);
 
-  if(!gtk_widget_is_visible(GTK_WIDGET(w)) && !darktable.gui->reset)
+  if(!darktable.gui->reset)
   {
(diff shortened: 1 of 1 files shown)


src/gui/accelerators.c
```diff
@@ -617,26 +614,23 @@ void dt_accel_widget_toast(GtkWidget *widget)
...
+        dt_toast_log(_("%s: %s"), w->label, text);
+    }
+    else
+    { //label is empty
+      if(w->module && w->module->multi_name[0] != '\0')
+        dt_toast_log(_("%s %s / %s"), w->module->name(), w->module->multi_name, text);
+      else if(w->module)
+        dt_toast_log(_("%s / %s"), w->module->name(), text);
+      else
+        dt_toast_log(_("%s"), text);
```
TurboGit, line 633: we don't need the `_(...)` here as there is nothing to translate Above it is different as there is the `/` or `:` that could be changed or spaced differently.

TurboGit (changes requested):
A minor translation to remove.

elstoc:
done

TurboGit:
Thanks.

PR #6691, opened 2020-10-31
[Issue #6668] Film roll sorted descending

### LINKED ISSUE #6668: Wrong Film roll sorting order
**Describe the bug**
The sorting order of film rolls does not match the sorting order of folders if 'recent to older' is set.

**To Reproduce**
1. preferences: sort film rolls by: folder
2. preferences: sort collection recent to older: true
3. Collect by folders --> 2020 on top of the list (OK)
4. Collect by film roll --> 2020 on bottom of the list (NOK)

**Expected behavior**
If 'sort by folder' is selected, the sorting order of film rolls should match the sorting order of folders.
So if 'recent to older' is set, the newest film rolls (2020...) should be on top.

**Screenshots**
sort-filmrolls

**Platform:**
Darktable Version: 3.2.1
OS: Linux Mint 20 Cinnamon
OpenCL activated
NVIDIA Driver Version: 450.80.02

quovadit:
This is a suggestion how to fix #6668 
As this is my first contribution, I hope I did not make any mistakes...

### DIFF
--- a/src/libs/collect.c
+++ b/src/libs/collect.c
@@ -1723,9 +1723,13 @@ static void list_view(dt_lib_collect_rule_t *dr)
         {
           gchar *order_by = NULL;
           if(strcmp(dt_conf_get_string("plugins/collect/filmroll_sort"), "id") == 0)
+            order_by = g_strdup("ORDER BY film_rolls_id");
+          if(strcmp(dt_conf_get_string("plugins/collect/filmroll_sort"), "id descending") == 0)
             order_by = g_strdup("ORDER BY film_rolls_id DESC");
-          else
+          else if(strcmp(dt_conf_get_string("plugins/collect/filmroll_sort"), "folder") == 0)
             order_by = g_strdup("ORDER BY folder");
+          else if(strcmp(dt_conf_get_string("plugins/collect/filmroll_sort"), "folder descending") == 0)
+            order_by = g_strdup("ORDER BY folder DESC");
 
           // filmroll
           g_snprintf(query, sizeof(query),
--- a/src/libs/collect.c
+++ b/src/libs/collect.c
@@ -1723,13 +1723,9 @@ static void list_view(dt_lib_collect_rule_t *dr)
         {
           gchar *order_by = NULL;
           if(strcmp(dt_conf_get_string("plugins/collect/filmroll_sort"), "id") == 0)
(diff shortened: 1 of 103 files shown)


src/libs/collect.c
```diff
@@ -1723,9 +1723,13 @@ static void list_view(dt_lib_collect_rule_t *dr)
         {
           gchar *order_by = NULL;
           if(strcmp(dt_conf_get_string("plugins/collect/filmroll_sort"), "id") == 0)
```
TurboGit, line 1725: I see that some code was already there, but let's get this properly. The string returned by dt_conf_get_string must be freed.

So, please take the value before this line and free when not needed anymore.

But, note that I'm not sure this is useful. For example the "id descending" will always list the older collection first. Meaning that new collection will always be listed at the bottom of the list. I have more than 1000 collections and I don't see why I would hurt myself using this setting. And note also that this is a pref, meaning you set it once basically.

So all in all, I'm not ok to merge this as-is. Maybe others have different opinions, we'll see...

quovadit: @TurboGit : Thanks for your input, as I said: this is my first contribution, and I am happy about any help.

I don't want to have the most recent collections at the bottom of the list neither.
But unfortunately 'id' does not work for me, because I often import older shootings at a later point, so the list would be shuffled.
And the setting 'sort film rolls by folder' does show the most recent collections at the bottom, although I've also checked 'sort collection recent to older'.
That's my problem, and that's what by bug #6668 is all about.

The suggestion of this pull request is to add an option, which does exactly what I am missing: sort film-rolls by folder descending.

Just adding this one option would lead to the following list:
- id
- folder
- folder descending

As 'id' is also descending at the moment, this would be misleading in my opinion.
That's why I split it up:
- id
- id descending (default!)
- folder
- folder descending

You are absolutely right, that id (ascending) is quite useless, but in my opinion it makes the list more clear.
And of course id descending is set as default

But I am happy about concrete alternatives to solve my issue.

One more thing: I don't understand your point 'free the string'. I just added else clauses and did not change the logic. Could you tell me what exactly I have to change?

elstoc:
As I mentioned in the linked issue, my opinion is that it would be better to include a reverse order option in the collect module itself, so that it can be used for all filter criteria, though I haven't investigated the practicality of this.

quovadit:
see second try: #6709

PR #6692, opened 2020-10-31
denoiseprofile: add preset wavelets chroma only

rawfiner:
Following the discussion here: https://discuss.pixls.us/t/remove-just-chroma-noise-in-darktable/19855/27

Adds a way to set denoiseprofile to wavelets in Y0U0V0 with all the Y0 curve set to 0, to remove only chroma noise.
To have best color accuracy, I set preserve shadows to 0 (which basically disable the non linear part of the variance stabilisation transform, hence all the algorithm is performed on linear data: wavelets perform well enough even on non-stabilized data) and bias set to 0.
strength is set to 3 to remove well all chroma noise whatever the noise amount in the image.

### DIFF
--- a/src/iop/denoiseprofile.c
+++ b/src/iop/denoiseprofile.c
@@ -589,6 +589,37 @@ int legacy_params(dt_iop_module_t *self, const void *const old_params, const int
   return 1;
 }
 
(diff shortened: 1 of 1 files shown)


TurboGit (approved):
Thanks ! Works great for me.

PR #6693, opened 2020-10-31
add draw multiple shapes to liquify

dterrahe:
holding ctrl while pushing one of the shape (point/line/curve) buttons (or using the separate accelerators) now adds more than one shape, like in retouch and spot removal.

Also fixes a small existing bug (not a regression from #6648) where pushing the reset parameters button _while_ drawing a shape caused a hang.

### DIFF
--- a/src/iop/liquify.c
+++ b/src/iop/liquify.c
@@ -281,6 +281,8 @@ typedef struct
   GtkLabel *label;
   GtkToggleButton *btn_point_tool, *btn_line_tool, *btn_curve_tool, *btn_node_tool;
 
+  gboolean creation_continuous;
+  gboolean just_started;
 } dt_iop_liquify_gui_data_t;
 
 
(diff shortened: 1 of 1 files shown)


src/iop/liquify.c
```diff
@@ -3158,6 +3160,30 @@ int button_pressed(struct dt_iop_module_t *module,
...
+  float radius = 0.0f, r = 1.0f, phi = 0.0f;
+  get_stamp_params(module, &radius, &r, &phi);
+  //  start a new path
+  g->temp = alloc_move_to(module, pt);
+  g->temp->warp.radius = pt + radius;
+  g->temp->warp.strength = pt + r * cexp(phi * I);
+  g->status |= DT_LIQUIFY_STATUS_PREVIEW;
+  g->status |= DT_LIQUIFY_STATUS_NEW;
+
+  g->just_started = TRUE;
```
TurboGit, line 3180: I support this is the issue I'm experiencing. Place a point, then place a curve with at least 3 points. You'll see the mess on the screen :)

TurboGit (changes requested):
issue with the bezier curve.

src/iop/liquify.c
dterrahe: I may not be seeing what you are seeing. Would you please post a screenshot and/or exact steps so I can reproduce?

Placing a point and then a curve with >3 nodes works fine here. Either with the normal or ctrl/multiple version.

TurboGit: Ok, it is not that easy to reproduce, but take the image used for the integration test.

1. clear the history
2. select liquify
3. use the non continuous mode (the continuous has not been tested)
4. place a single point in the center of the perceptional scale (white circle on the right)
5. place a curve 3 points
6. reset the module with the button
7. repeatgit pull https://github.com/dterrahe/darktable.git init_gui_accel 4
8. repeat 5


image

TurboGit:
Ok, I checked with current master and was also able to reproduce. So that's not a regression, just me trying hard to break things :) So I'll merge this for now.

src/iop/liquify.c
dterrahe: Could this have something to do with crap from old points being left behind in the params when deleting or resetting? Similar to #6685 for curves.

I'm putting on my todo list to have a look at this. Management of the point list isn't as obvious in liquify, so I'll need to analyse and do careful testing. Similar issues probably in retouch and spot.

PR #6703, opened 2020-11-01
watermark iop parallelization

ralfbrown:
This PR parallizes the final blending  of the image and watermark for about a 1/3 decrease in wall time with 32 threads.


TurboGit:
For the record before/after:

```
$ ./run.sh 0032-watermark/
Test 0032-watermark
      Image mire1.cr2
      CPU & GPU version differ by 23936 pixels
      Max  dE         0.0000
  OK


Total test 1
Errors     0
```

PR #6709, opened 2020-11-01
[Issue #6668] Film roll by folder sorted depending on pref-setting

### LINKED ISSUE #6668: Wrong Film roll sorting order

quovadit:
Second suggestion how to fix #6668
As folders and dates are already sorted depending on the checkbox 'sort collection recent to older', it makes sense to also include film-rolls.

### DIFF
--- a/data/darktableconfig.xml.in
+++ b/data/darktableconfig.xml.in
@@ -544,8 +544,8 @@
     <name>plugins/collect/descending</name>
     <type>bool</type>
     <default>true</default>
-    <shortdescription>sort collection recent to older</shortdescription>
-    <longdescription>changes the default collections-list order for folders, times and dates to run from recent to older</longdescription>
+    <shortdescription>sort collection in descending order</shortdescription>
+    <longdescription>changes the collections-list order for folders, film-rolls by folder, times and dates to descending</longdescription>
   </dtconfig>
   <dtconfig>
     <name>ui_last/colorpicker_mean</name>
--- a/src/libs/collect.c
+++ b/src/libs/collect.c
@@ -1725,7 +1725,12 @@ static void list_view(dt_lib_collect_rule_t *dr)
           if(strcmp(dt_conf_get_string("plugins/collect/filmroll_sort"), "id") == 0)
             order_by = g_strdup("ORDER BY film_rolls_id DESC");
           else
-            order_by = g_strdup("ORDER BY folder");
+          {
+            if(dt_conf_get_bool("plugins/collect/descending"))
+              order_by = g_strdup("ORDER BY folder DESC");
+            else
+              order_by = g_strdup("ORDER BY folder");
+          }
 
           // filmroll
           g_snprintf(query, sizeof(query),


elstoc:
I'm not sure whether 'recent to older' is the right term for this preference setting when applied to folders or film rolls. I think "sort collection in descending order" is the better name, and the tooltip should be updated to reflect that it applies to film rolls as well. These settings are in ``data/darktableconfig.xml.in``

quovadit:
@elstoc : good point, I updated name and tooltip of the setting

TurboGit:
If this will probably work, let's discuss another solution that will be more generic and will work for all collections.

1. remove the preference
2. add a button in "collect" module to sort the list/tree in order or reverse order
3. use this new setting to sort the collections

This will mean that we would be able to easily change the sort from the interface (the pref is not really usable) easily and it will work for all kind of collections (ISO, aperture, filmroll, date/time, maker, lens...).

How does that sound ? If it doesn't make it into 3.4 it can be added into a later minor release, no urgency on this I would say.

elstoc:
> How does that sound ?

Sounds like a good plan to me. Anything that moves stuff like this out of prefs and into the individual modules gets my vote.

quovadit:
For me the prefs is fine, because it's a one-time-decision, and I don't see the usecase of sorting iso or aperture descending.

If there would be a button in the collect-module, it would be important that the decision is remembered as well

elstoc:
> it would be important that the decision is remembered as well

Yes it can still be stored to the same preference setting, just set within the module instead of within the preference dialog.

quovadit:
I think it's more important, that the UI which I am using every day is clean, and a sorting button makes it more complicated.
For me the preferences are a dialog I use once, there I decide how I work and if I am using foldernames with a date, I chose the options 'sort by folder' and 'descending true'.
I really don't see a usecase, where I want to switch the sorting order frequently, and I still don't see the need to sort anything else than folder, film-roll and dates.

TurboGit:
> For me the prefs is fine, because it's a one-time-decision

No it is not, many time I'll add the need to sort the folder to find them more easily. When you have lot of them it is easier than going through a very long list. Likewise for date, ISO or whatever. A pref is indeed a one time setting, and here it is not one time in my point of view.

TurboGit:
> I think it's more important, that the UI which I am using every day is clean, and a sorting button makes it more complicated.

Well, we are just speaking of a button with a little arrow as in the sorting of the lighttable thumbs on the top bar.

quovadit:
I absolutely disagree.
There are already at least two very similar arrows in the collect module, another one leads to mistake and also needs space.
sort-button-in-collect-module

Another point: it makes sense to sort folder, film-roll by folder and dates descending, to have the most recent items on top in a very long list.
But: I don't think it's an expected behaviour, that as a consequence the creator, ISO or title are also sorted descending.

And I definitely don't want to change the sort order back and forth all the time.

I think my change in preferences is very subtle, helps me a lot, and I can't imagine it could bother anybody.
A UI-change has always the potential to disturb users, and I still don't see the advantage.

TurboGit:
> And I definitely don't want to change the sort order back and forth all the time.

dt is created for other people than just you.

> I think my change in preferences is very subtle, helps me a lot,

Again dt is not meant to be used only by you.

With such thinking I would have probably done less than half the work I've done on dt !

elstoc:
A tick box at the bottom of the pane would be less visually intrusive to the existing interface.

quovadit:
I am really sorry if I created the impression, that I want to implement a feature just for me!
Of course every feature must work for all dt users!
That's exactly why I would prefer a subtle preferences-change instead of a UI-change.
But as already said: it's just a suggestion, I am sure the experienced dt-developers will decide which solution is best for the majority of dt-users.

parafin:
I would suggest a ban on adding new preferences or hard limit on their number, which can't be exceeded (so you have to remove some other options before adding a new one). Adding a new preference is an easy way out of making decisions, which is a wrong thing to do. For this particular feature - sorting order should be a part of the module that displays the list, I've never seen such setting hidden away in preferences, it just doesn't make sense.

quovadit:
see #6797

PR #6715, opened 2020-11-01
don't force square aspect on graph in tone equaliser

### LINKED ISSUE #6662: Tone Eq UI....lots of blank space at the bottom of the module between sliders and mask button bar....
Current Dev version has excessive empty space in the Tone eq module that requires substantial scrolling to get to the mask buttons to adjust opacity or blend mode....Likely a simple fix..

This occurs on WIndows 10 with a recent build I think 1285??  I can update with detail if this is not obvious to others or is not showing on Linux builds....

dterrahe:
The minimum height of the tone equaliser module is determined by the 9 sliders on the simple tab. This PR makes the graph on the advanced tab always use that height, rather than leave a large area unused when it matched the height to the width of a narrow panel. Conversely, if the panel is very wide, the 1-1 aspect ratio forced the graph to be very high as well, causing unused space to appear on the 1st and 3rd tab.

image)image

fixes #6662

Edit: now also restoring the page it was on last time and small improvement to dragging of curve nodes.

### DIFF
--- a/data/darktableconfig.xml.in
+++ b/data/darktableconfig.xml.in
@@ -2588,4 +2588,11 @@
     <shortdescription>height of the styles list</shortdescription>
     <longdescription>maximum height the styles list in lighttable will grow to before scrolling</longdescription>
   </dtconfig>
+  <dtconfig>
+    <name>plugins/darkroom/toneequal/gui_page</name>
+    <type>int</type>
+    <default>0</default>
+    <shortdescription>active tab in tone equaliser module</shortdescription>
+    <longdescription>which of simple, advanced or masking tabs will show at startup</longdescription>
+  </dtconfig>
 </dtconfiglist>
--- a/src/iop/toneequal.c
+++ b/src/iop/toneequal.c
@@ -3197,6 +3197,10 @@ void gui_init(struct dt_iop_module_t *self)
   // start building top level widget
   self->widget = gtk_box_new(GTK_ORIENTATION_VERTICAL, 0);
 
+  int active_page = dt_conf_get_int("plugins/darkroom/toneequal/gui_page");
+  gtk_widget_show(gtk_notebook_get_nth_page(g->notebook, active_page));
+  gtk_notebook_set_current_page(g->notebook, active_page);
+
   g_signal_connect(G_OBJECT(g->notebook), "button-press-event", G_CALLBACK(notebook_button_press), self);
   gtk_box_pack_start(GTK_BOX(self->widget), GTK_WIDGET(g->notebook), FALSE, FALSE, 0);
 
@@ -3224,6 +3228,8 @@ void gui_cleanup(struct dt_iop_module_t *self)
 {
   dt_iop_toneequalizer_gui_data_t *g = (dt_iop_toneequalizer_gui_data_t *)self->gui_data;
   self->request_color_pick = DT_REQUEST_COLORPICK_OFF;
+
+  dt_conf_set_int("plugins/darkroom/toneequal/gui_page", gtk_notebook_get_current_page (g->notebook));
 
   DT_DEBUG_CONTROL_SIGNAL_DISCONNECT(darktable.signals, G_CALLBACK(_develop_ui_pipe_finished_callback), self);
   DT_DEBUG_CONTROL_SIGNAL_DISCONNECT(darktable.signals, G_CALLBACK(_develop_ui_pipe_started_callback), self);
--- a/src/iop/toneequal.c
+++ b/src/iop/toneequal.c
@@ -2754,7 +2754,17 @@ static gboolean area_leave_notify(GtkWidget *widget, GdkEventCrossing *event, gp
   if(!self->enabled) return 0;
 
   dt_iop_toneequalizer_gui_data_t *g = (dt_iop_toneequalizer_gui_data_t *)self->gui_data;
+  dt_iop_toneequalizer_params_t *p = (dt_iop_toneequalizer_params_t *)self->params;
 
+  if(g->area_dragging)
+  {
+    // cursor left area : force commit to avoid glitches
+    ++darktable.gui->reset;
+    update_exposure_sliders(g, p);
+    --darktable.gui->reset;
+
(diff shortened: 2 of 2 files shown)


elstoc:
It's a nice idea but perhaps let it grow just a bit bigger than the height dictated by the simple tab (perhaps +20%) or add the ctrl+scroll resize functionality like in the lib modules. Some people like to make their panels wider to get finer control over those sorts of curves. 

Also bear in mind that the "issue" (or feature if you like) of module's height increasing proportionally to its width also occurs in quite a few other modules (base curve, tone curve, rgb curve, filmic, levels, rgb levels, color mapping, contrast equalizer etc. etc.). The only difference here is that the other tabs being less filled with controls make the gap look a bit more strange.

dterrahe:
> Some people like to make their panels wider to get finer control over those sorts of curves.

There are nine (9!) sliders on the "simple" tab that are happy to accept their right clicks.

> The only difference here is that the other tabs being less filled with controls make the gap look a bit more strange.

Exactly. And that there is a natural minimum height for the graph here that imho is sufficient.

dterrahe:
As an update; I'm trying to fix the sensitivity at the top/bottom of the graph so that a node drag doesn't immediately end when the mouse passes the threshold, which makes it difficult to exactly position a node at the extreme (or pick it up from there). We had a similar issue with slider drags a while back. (@AxelG-DE will remember)

Since the sizing of the graph used to incorrectly refer to the module widget allocation (self->widget) rather than the drawingarea (widget) some of the graph size adjustments are a bit confused/confusing. So I'm (again) spending more time on this than intended.

dterrahe:
Previously, when dragging nodes up or down, when hitting the edge of the graph the drag would be cancelled, whether the active node was at the edge or not. If you naively think that you need to click _on_ the node to be able to drag it, this makes it very difficult to drag it exactly to the edge or to pick it up from there.

Now, it keeps dragging until the mouse reaches the end of the drawingarea widget. As a consequence it is possible to keep dragging a node that is already at the end a little more, which causes the nodes next to it to continue to move down. Moving back up will then not restore the previous shape. This would already happen if you started the downward drag somewhere above the active node, so presumably this is intentional.

@aurelienpierre since this is yours, can you please confirm you are happy with the (slightly) changed behavior?

I was thinking that maybe the drag should stop (not cancel) when the _node_ (not the mouse) reaches the lower or upper end of the range. If the mouse movement was then reversed, the node should be picked up again from the point where it was left, so that moving down and back up to the same spot for the mouse would also return the node to its starting position. But this may not be worth pursuing, because in the mean time another (lower) node could have reached the boundary first and the curve would have started to deform. Reversing this too would be hard/not worth it.

AxelG-DE:
> As an update; I'm trying to fix the sensitivity at the top/bottom of the graph so that a node drag doesn't immediately end when the mouse passes the threshold, which makes it difficult to exactly position a node at the extreme (or pick it up from there). We had a similar issue with slider drags a while back. (@AxelG-DE will remember)

I do remember, we have had the issue, where clicking the color picker made the slider jump to the max. position. Is that, what you meant? 

Anyhow, if you need a "test-engineer" I volunteer (just a bit less responsive than last time, as I am not in vacation atm)

Actually I do have a little bit opposite phenomenon in contrast equalizer, sometimes it gets difficult to go to the top border as it feels a bit "rubberbanded" following the mouse, but this is OT here, likely.

dterrahe:
> I do remember, we have had the issue, where clicking the color picker made the slider jump to the max. position. Is that, what you meant?

that was somewhat related, but specifically I meant that if you dragged a slider too quickly to the right (continuing to drag into the color picker area) the slider wouldn't continue all the way to the end; it would get dropped slightly _before_ the end, because the mouse went in one step over the end position. To get the slider exactly to max was very finicky (you had to move the mouse real slow).

> Actually I do have a little bit opposite phenomenon in contrast equalizer, sometimes it gets difficult to go to the top border as it feels a bit "rubberbanded" following the mouse, but this is OT here, likely.

Contrast equalizer? That seems to be quite predictable for me; only the one node directly following the mouse. Tone Equalizer does some fancy interpolation/curve fitting so many nodes move at the same time and interact.

AxelG-DE:
> but specifically I meant that if you dragged a slider too quickly to the right (continuing to drag into the color picker area) the slider wouldn't continue all the way to the end; it would get dropped slightly _before_ the end, because... 

I remember now :slight_smile:

aurelienpierre:
> @aurelienpierre since this is yours, can you please confirm you are happy with the (slightly) changed behavior?

I don't mind.

dterrahe:
> I don't mind.

@TurboGit after Aurelien's glowing endorsement, I think this is ready. It is partly a "bugfix" for mouse dragging behavior at end of graph, partly new feature, saving the active page, and partly an aesthetic improvement for unnecessary empty space on the other tabs. Could go equally into 3.4.2.

elstoc:
> Could go equally into 3.4.2

I assume you mean 3.4.1?

dterrahe:
> > Could go equally into 3.4.2
> 
> I assume you mean 3.4.1?

Pascal normally reserves 3.4.1 for urgent bugfixes only. Heck, we might never even release 3.4.0 ;-)

(maybe that's not a good joke, I don't know; might help to unjinx things though...)

elstoc:
> Pascal normally reserves 3.4.1

That was an aberration for 3.2.1 because an urgent fix was identified immediately after the release was tagged. The .1 release is the first bugfix release which, barring emergencies is usually released some time between Jan and March.

TurboGit:
@elstoc that was an aberration for 3.2.1 but indeed I do keep the first .1 release for urgent and critical fixes. If 3.4.0 is pretty unusable for any reasons we need a very safe fix in .1 and no other possible commit introducing yet some other bad side effects. In other word I really like have .1 to be very conservative and safe.

TurboGit:
Let's try to squash this in 3.4, I'll review asap.

dterrahe:
Thank you!

TurboGit:
Looks ok to me, and nice idea to keep the active tab and restore it ! Thanks a lot.

PR #6718, opened 2020-11-02
guard call to gui_update against missing implementation

dterrahe:
Prevents crash if an iop module is missing a gui_update implementation (15bef3b)

### DIFF
--- a/src/develop/imageop.c
+++ b/src/develop/imageop.c
@@ -1895,7 +1895,7 @@ void dt_iop_gui_update(dt_iop_module_t *module)
     ++darktable.gui->reset;
     if(!dt_iop_is_hidden(module))
     {
-      if(module->params) module->gui_update(module);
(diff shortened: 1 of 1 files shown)


dterrahe:
@TurboGit you might want to merge this quickly (it is trivial/obviously correct, I hope) because currently master crashes which stops people from testing #6673 ;-)

TurboGit (approved):
Thanks !

TurboGit:
Strange that I didn't get any crash on my side...

dterrahe:
It would only crash when switching to darkroom. Maybe not even then if both those two modules are somehow disabled in your config.

dterrahe:
Anyway, the end result is now cleaner. Hopefully it will reliably crash in testing if ever somebody decides to directly call all module->gui_updates, rather than using dt_iop_gui_update

PR #6719, opened 2020-11-02
mask-manager: fix cleanup unused shapes

elstoc:
now cleans up shapes that are *only* used by the mask manager module instead of those that are *not* used by the mask manager module


TurboGit (approved):
Indeed, a bad typo and really the compiler should complain here I suppose :(

PR #6720, opened 2020-11-02
iop: reset parameters tooltip to reference Ctrl+click functionality

elstoc:


TurboGit (approved):
Ok, will just set ctrl all lower case for consistency.

PR #6723, opened 2020-11-02
exif: convert from APEX values

kmilos:
See http://www.cipa.jp/std/documents/download_e.html?DC-008-Translation-2019-E Annex C

### DIFF
--- a/src/common/exif.cc
+++ b/src/common/exif.cc
@@ -771,7 +771,7 @@ static bool _exif_decode_exif_data(dt_image_t *img, Exiv2::ExifData &exifData)
     else if(FIND_EXIF_TAG("Exif.Photo.ShutterSpeedValue"))
     {
       // uf_strlcpy_to_utf8(uf->conf->shutterText, max_name, pos, exifData);
-      img->exif_exposure = 1.0 / pos->toFloat();
+      img->exif_exposure = exp2f(-1.0f * pos->toFloat());  // convert from APEX value
     }
 
     // Read exposure bias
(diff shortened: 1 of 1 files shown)


aurelienpierre:
Could you develop a bit on the why ?

kmilos:
Bien sur - I think this is simply a corner case bug: `ExposureTime` and `ShutterSpeedValue` are not interchangeable without conversion, and the same applies to `FNumber` and `ApertureValue`.

From the tag definitions in the Exif spec linked above:

No. | Tag | Description
--- | --- | --- 
33434 (829A.H) | [ExposureTime](https://www.awaresystems.be/imaging/tiff/tifftags/privateifd/exif/exposuretime.html) | Exposure time, given in seconds (sec). 
37377 (9201.H) | [ShutterSpeedValue](https://www.awaresystems.be/imaging/tiff/tifftags/privateifd/exif/shutterspeedvalue.html) | Shutter speed. The unit is the APEX (Additive System of Photographic Exposure) setting (see Annex C). 

and similar for the other pair. Annex C gives the conversion formulae.

This is also implemented in exiv2 as part of the [tag custom printing functions](https://github.com/Exiv2/exiv2/blob/3e78cac3f4115d5d7003db0db6bf5dcb166e37dd/src/tags_int.cpp#L811); the tag values are stored in APEX, but [printed](https://github.com/Exiv2/exiv2/blob/3e78cac3f4115d5d7003db0db6bf5dcb166e37dd/src/tags_int.cpp#L2509) uniformly as seconds using the appropriate APEX [conversion](https://github.com/Exiv2/exiv2/blob/3e78cac3f4115d5d7003db0db6bf5dcb166e37dd/src/tags_int.cpp#L2201) (and similar for F#).

It just hasn't popped up yet because it seems most camera vendors ship `ExposureTime` and `FNumber` so the other case is rarely hit.

TurboGit (approved):
Correct ! Thanks.

PR #6728, opened 2020-11-02
re-enable fast-math for Release build and libmvec for GCC

aurelienpierre:
This enables https://sourceware.org/glibc/wiki/libmvec.

`-fast-math` was removed in d5afea7 hoping to fix weird bugs. But it turns out that it completely disables auto-vectorization too, as it shows here : https://godbolt.org/z/fjxYqK vs. https://godbolt.org/z/xc7zs3.

Notice that this will make some modules slower than before. I will fix them in the next weeks.


johnny-bit:
Interesting that this breaks filmic unittests ;)

aurelienpierre:
The test broken is a division of a sum of powers, and I suspect this has to do with denormals flushing close to zero.

johnny-bit:
Is this PR made obsolete by #8923?

TurboGit:
Yes, OBE to me. Closing.

PR #6735, opened 2020-11-03
ppd corrected dragging distance in culling/preview mode

jenshannoschwalm:
While dragging image content in preview or culling mode the amount of shift
in both directions wants to be corrected for ppd_thb for a more pleasant user interface.


TurboGit:
Works for me ! Thanks.

PR #6738, opened 2020-11-03
Introducing the rcd demosaicer (2nd attempt)

jenshannoschwalm:
The rcd demosaicer developed by Luis Sanz Rodríguez is publically available at
https://github.com/LuisSR/RCD-Demosaicing and is licenced under the GNU GPL version 3.

I came across this demosaicer via the rt issue tracker https://github.com/Beep6581/RawTherapee/issues/5908 and have been using it in dt for a month or so.

The results are pretty good, quality en-par with Amaze, in some situations amaze is better, in others it's rcd having better results.

The biggest pro argument is less color overshoot, the penalty is minimally increased blur.

The algo has been directly taken from the original source, adopted to dt style and performance tuned via omp.

Right now the performance is just about that of amaze. (the first attempt had a number of issues leading to sometimes wrong colors and access on out-of-buffer crashes in the border interpolation code.)

The border interpolation used the dcraw algo in the original code, i chose the code used by rt as it has less artefacts.

### DIFF
--- a/src/iop/demosaic.c
+++ b/src/iop/demosaic.c
@@ -2696,8 +2696,26 @@ static void rcd_demosaic(float *out, const float *cfa, dt_iop_roi_t *const roi_o
     {
       const int indx = row * width + col;
       const float cfai = cfa[indx];
-      const float V_Stat = fmaxf(epssq, -18.f * cfai * (cfa[indx - w1] + cfa[indx + w1] + 2.f * (cfa[indx - w2] + cfa[indx + w2]) - cfa[indx - w3] - cfa[indx + w3]) - 2.f * cfai * (cfa[indx - w4] + cfa[indx + w4] - 19.f * cfai) - cfa[indx - w1] * (70.f * cfa[indx + w1] + 12.f * cfa[indx - w2] - 24.f * cfa[indx + w2] + 38.f * cfa[indx - w3] - 16.f * cfa[indx + w3] - 12.f * cfa[indx - w4] + 6.f * cfa[indx + w4] - 46.f * cfa[indx - w1]) + cfa[indx + w1] * (24.f * cfa[indx - w2] - 12.f * cfa[indx + w2] + 16.f * cfa[indx - w3] - 38.f * cfa[indx + w3] - 6.f * cfa[indx - w4] + 12.f * cfa[indx + w4] + 46.f * cfa[indx + w1]) + cfa[indx - w2] * (14.f * cfa[indx + w2] - 12.f * cfa[indx + w3] - 2.f * cfa[indx - w4] + 2.f * cfa[indx + w4] + 11.f * cfa[indx - w2]) + cfa[indx + w2] * (-12.f * cfa[indx - w3] + 2.f * (cfa[indx - w4] - cfa[indx + w4]) + 11.f * cfa[indx + w2]) + cfa[indx - w3] * (2.f * cfa[indx + w3] - 6.f * cfa[indx - w4] + 10.f * cfa[indx - w3]) + cfa[indx + w3] * (-6.f * cfa[indx + w4] + 10.f * cfa[indx + w3]) + cfa[indx - w4] * cfa[indx - w4] + cfa[indx + w4] * cfa[indx + w4]);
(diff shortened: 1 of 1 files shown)


src/iop/demosaic.c
```diff
@@ -2422,6 +2426,452 @@ static void passthrough_monochrome(float *out, const float *const in, dt_iop_roi
...
... - 1] * cfa[indx + w4 + 4] + 46.f * cfa[indx - w1 - 1] * cfa[indx - w1 - 1] + 24.f * cfa[indx + w1 + 1] * cfa[indx - w2 - 2] - 12.f * cfa[indx + w1 + 1] * cfa[indx + w2 + 2] + 16.f * cfa[indx + w1 + 1] * cfa[indx - w3 - 3] - 38.f * cfa[indx + w1 + 1] * cfa[indx + w3 + 3] - 6.f * cfa[indx + w1 + 1] * cfa[indx - w4 - 4] + 12.f * cfa[indx + w1 + 1] * cfa[indx + w4 + 4] + 46.f * cfa[indx + w1 + 1] * cfa[indx + w1 + 1] + 14.f * cfa[indx - w2 - 2] * cfa[indx + w2 + 2] - 12.f * cfa[indx - w2 - 2] * cfa[indx + w3 + 3] - 2.f * cfa[indx - w2 - 2] * cfa[indx - w4 - 4] + 2.f * cfa[indx - w2 - 2] * cfa[indx + w4 + 4] + 11.f * cfa[indx - w2 - 2] * cfa[indx - w2 - 2] - 12.f * cfa[indx + w2 + 2] * cfa[indx - w3 - 3] + 2 * cfa[indx + w2 + 2] * cfa[indx - w4 - 4] - 2.f * cfa[indx + w2 + 2] * cfa[indx + w4 + 4] + 11.f * cfa[indx + w2 + 2] * cfa[indx + w2 + 2] + 2.f * cfa[indx - w3 - 3] * cfa[indx + w3 + 3] - 6.f * cfa[indx - w3 - 3] * cfa[indx - w4 - 4] + 10.f * cfa[indx - w3 - 3] * cfa[indx - w3 - 3] - 6.f * cfa[indx + w3 + 3] * cfa[indx + w4 + 4] + 10.f * cfa[indx + w3 + 3] * cfa[indx + w3 + 3] + 1.f * cfa[indx - w4 - 4] * cfa[indx - w4 - 4] + 1.f * cfa[indx + w4 + 4] * cfa[indx + w4 + 4], epssq );
```
aurelienpierre, line 2773: any chance this could be formatted in a shape that makes some filter appear ? That's hardly legible.

jenshannoschwalm: You mean like keeping line length in sensible order plus some formatting? Sure...

jenshannoschwalm: should be better now

aurelienpierre: I meant a way where the ordering of coefficients shows some logic. Something like : 
https://github.com/darktable-org/darktable/blob/bc2763e464277cd618be4b954dca40e65fd92113/src/iop/filmicrgb.c#L2092-L2097

But I understand there is no paper that explains what happens in the algo ?

jenshannoschwalm: Unfortunately not. Some discussion about this at https://discuss.pixls.us/t/diagonal-interpolation-correction-artifacts-with-amaze-demosaicing/3338/94

jenshannoschwalm:
Using tiles now for much lower mem footprint.

jenshannoschwalm:
I think it's ready to go now ...

jakubfi:
It seems that te new demosaicer does "color smoothing" according to how it was set for other demosaicers, but doesn't have that slider in the UI.

jenshannoschwalm:
It indeed does some smoothing but that is not after applying the maths but is used to flatten the reconstruction leading to less overshoots. There is some explaining in https://github.com/LuisSR/RCD-Demosaicing/blob/master/README.md

jakubfi:
What I mean is that the results of this demosaicer are dependant on a parameter set for another demosaicer. There should be either no dependency, or explicit slider for "color smootihing" in RCD UI.

jenshannoschwalm:
> What I mean is that the results of this demosaicer are dependant on a parameter set for another demosaicer

Aah, but that is not so. There is no parameter taken from any slider. There is no dependency ...

jakubfi:
My test was:
Load an image, select amaze demozaicer, set color smoothing to 0, switch to RCD:

Switch to amaze, set color smoothing to 5, switch back to RCD:

jenshannoschwalm:
Aah, thanks to pointing to this, in fact the smoothing is post-processed and i didn't check this :-( Anyway it wouldn't hurt to have this also available ...

jakubfi:
Yep, this seems to be a good idea. I, for one, would love to have color smoothing available for RCD too. :-)
BTW: Thanks for adding RCD, I find it better than AMaZE in basically all cases. AMaZE tends to produce impulse-noise-like artifacts sometimes. Not sure if this is supposed to be like this, or is it a bug in the implementation.

jenshannoschwalm:
Feedback welcome. I also prefer this over AMaZE in almost all cases...

jakubfi:
Now I have to ask. :-) Why do you think there is no use for color smoothing with RCD? From my observations RCD benefits from it the _exact_ same way other demosaicers do. Color artifacts for AMaZE and RCD are basically the same and color smoothing helps to deal with them in both cases.

TurboGit:
@jenshannoschwalm : can this wait for 3.4.x ? we had a first shake of dt code today, another is planned for the blending PR, if we can relax a bit and have this integrated later it will help :) but if you really like this to be in 3.4 just say so and we'll try to squeeze it in.

jenshannoschwalm:
@jakubfi ok, switched this on, maybe some people find it useful and it doesn't hurt

jenshannoschwalm:
@TurboGit I fully understand time limits :-) So feel free to look into this when you find some time to do so.

The positive side of this addition is: it affects only one file and only one module :-)

There have been more late commits than i desired making pr reading a bit of a mess. Would it help to make a new pr with just one commit?

aurelienpierre:
The change is not massive, but that would be much easier to review if some paper showed the equations we try to resolve. Instead, we will need extensive tests to assess the quality of the algo.

jenshannoschwalm:
> The change is not massive, but that would be much easier to review if some paper showed the equations ...

Right. I think there is not enough time to do such testing with all the other stuff coming in dt master for 3.4.

I will a) ask the original author and some people of the rt team about this and b) come back with hopefully more info / test scenarios.

PR #6741, opened 2020-11-03
Add darktable.gui->ppd_thb and preview_downsampling correction to focus display.

### LINKED ISSUE #6737: CTRL+W  (preview with focus detection) seems to be offset top left for windows build
When I do a ctrl+w in lighttable the red focus boxes appear top left. Which clearly is not the focus area. I wonder if the placement of these boxes has got affected with some of the dpi changes or image display resolution changes.

image

On todays master build for windows.

Same image ctrl+w
On windows 
image

On linux
image

jenshannoschwalm:
Fixes #6737

### DIFF
--- a/src/common/focus.h
+++ b/src/common/focus.h
@@ -294,7 +294,7 @@ static void dt_focus_draw_clusters(cairo_t *cr, int width, int height, int imgid
     if(ht * scale <= height) fy = 0;
   }
 
-  cairo_translate(cr, -wd / 2.0f + fx / scale, -ht / 2.0f + fy / scale);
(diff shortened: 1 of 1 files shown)


TurboGit:
That's not fixing the issue to me when the preview down-sampling is set to something different that original. Just set it to 1/4 and you'll see that the focus point are all grouped into the top-left corner.

TurboGit (changes requested):
Looks like there is still a missing scaling somewhere.

jenshannoschwalm:
Ah, i didn't check the preview downsampling at all, i never use it.

jenshannoschwalm:
Yeah, but that was related to preview_downscaling ...

jenshannoschwalm:
Should be fixed

chhil:
Tested changes on windows and linux, the boxes appear same size and location. I consider this fixed.

TurboGit (approved):
Works for me, thanks!

PR #6742, opened 2020-11-03
Map: better group of images based on cluster calculation (dbscan)

phweyland:
The effect is really good as long as the number of images on the map don't exceed 2000.
You can shift the map, the images groups remain stable.
In map settings one can adjust the main parameters of dbscan (epsilon and min images)

But, as the calculation time is f(n^2), for 30.000 I wait 6 or 7 seconds to get the clusters calculated.

Not usable as is, unfortunately, but I post the PR hoping I'll get some suggestions to improve the performance, at least from the UI standpoint.

### DIFF
--- a/src/libs/map_settings.c
+++ b/src/libs/map_settings.c
@@ -15,6 +15,7 @@
     You should have received a copy of the GNU General Public License
     along with darktable.  If not, see <http://www.gnu.org/licenses/>.
 */
+#include "bauhaus/bauhaus.h"
 #include "common/collection.h"
 #include "common/darktable.h"
 #include "common/debug.h"
@@ -51,6 +52,7 @@ uint32_t container(dt_lib_module_t *self)
 typedef struct dt_lib_map_settings_t
 {
   GtkWidget *show_osd_checkbutton, *filtered_images_checkbutton, *map_source_dropdown;
+  GtkWidget *epsilon_factor, *min_images;
 } dt_lib_map_settings_t;
 
 int position()
@@ -91,6 +93,26 @@ static void _map_source_changed(GtkWidget *widget, gpointer data)
     map_source = g_value_get_int(&value);
     g_value_unset(&value);
     dt_view_map_set_map_source(darktable.view_manager, map_source);
+  }
+}
+
+static void _epsilon_factor_callback(GtkWidget *slider, gpointer data)
+{
+  const float epsilon = dt_bauhaus_slider_get(slider);
+  dt_conf_set_int("plugins/map/epsilon_factor", epsilon);
+  if(darktable.view_manager->proxy.map.view)
+  {
+    darktable.view_manager->proxy.map.redraw(darktable.view_manager->proxy.map.view);
+  }
+}
+
+static void _min_images_callback(GtkWidget *slider, gpointer data)
+{
+  const int min_images = dt_bauhaus_slider_get(slider);
+  dt_conf_set_int("plugins/map/min_images_per_group", min_images);
+  if(darktable.view_manager->proxy.map.view)
+  {
+    darktable.view_manager->proxy.map.redraw(darktable.view_manager->proxy.map.view);
   }
 }
 
@@ -153,6 +175,19 @@ void gui_init(dt_lib_module_t *self)
   g_signal_connect(G_OBJECT(d->map_source_dropdown), "changed", G_CALLBACK(_map_source_changed), NULL);
 
   gtk_box_pack_start(GTK_BOX(self->widget), GTK_WIDGET(hbox), TRUE, TRUE, 0);
+
+  d->epsilon_factor = dt_bauhaus_slider_new_with_range(NULL, 10., 100., 5.,
+                            dt_conf_get_int("plugins/map/epsilon_factor"), 0);
+  gtk_widget_set_tooltip_text(d->epsilon_factor, _("modify the spatial size of an images group on the map"));
+  dt_bauhaus_widget_set_label(d->epsilon_factor, NULL, N_("group size factor"));
+  g_signal_connect(G_OBJECT(d->epsilon_factor), "value-changed", G_CALLBACK(_epsilon_factor_callback), self);
+  gtk_box_pack_start(GTK_BOX(self->widget), GTK_WIDGET(d->epsilon_factor), TRUE, TRUE, 0);
+  d->min_images = dt_bauhaus_slider_new_with_range(NULL, 1.0, 10.0, 1.0,
+                            dt_conf_get_int("plugins/map/min_images_per_group"), 0);
+  gtk_widget_set_tooltip_text(d->min_images, _("minimum of images to make a group"));
+  dt_bauhaus_widget_set_label(d->min_images, NULL, N_("min images per group"));
+  g_signal_connect(G_OBJECT(d->min_images), "value-changed", G_CALLBACK(_min_images_callback), self);
+  gtk_box_pack_start(GTK_BOX(self->widget), GTK_WIDGET(d->min_images), TRUE, TRUE, 0);
 
   g_object_unref(model);
   g_free(map_source);
--- a/src/views/map.c
+++ b/src/views/map.c
@@ -54,6 +54,8 @@ typedef struct dt_map_t
   gboolean filter_images_drawn;
   int max_images_drawn;
   float lat0, lat1, lon0, lon1;
+  int time_out;
+  int timeout_event_source;
   struct
   {
     dt_location_draw_t main;
@@ -68,8 +70,6 @@ typedef struct dt_map_image_t
   gint imgid;
   double latitude;
   double longitude;
-  double cum_lat;
-  double cum_lon;
   int group;
   int group_count;
   gboolean group_same_loc;
@@ -78,9 +78,21 @@ typedef struct dt_map_image_t
   gint width, height;
 } dt_map_image_t;
 
+#define UNCLASSIFIED -1
+#define NOISE -2
+
+#define CORE_POINT 1
+#define NOT_CORE_POINT 0
+
+typedef struct geo_position_t
+{
+  double x, y;
+  int cluster_id;
+  int imgid;
+} geo_position_t;
+
 static const int thumb_size = 128, thumb_border = 2, image_pin_size = 13, place_pin_size = 72;
 static const int cross_size = 16, max_size = 1024;
-static const float thumb_overlap = 1.2f;
 static const uint32_t thumb_frame_color = 0x000000aa;
 static const uint32_t thumb_frame_sel_color = 0xffffffee;
 static const uint32_t pin_outer_color = 0x0000aaaa;
@@ -138,7 +150,9 @@ static gboolean _view_map_dnd_failed_callback(GtkWidget *widget, GdkDragContext 
 static void _view_map_dnd_remove_callback(GtkWidget *widget, GdkDragContext *context, gint x, gint y,
                                           GtkSelectionData *selection_data, guint target_type, guint time,
                                           gpointer data);
-
+// find the images clusters on the map
+static void _dbscan(geo_position_t *points, unsigned int num_points, double epsilon,
+             unsigned int minpts);
 static gboolean _view_map_prefs_changed(dt_map_t *lib);
 static void _view_map_build_main_query(dt_map_t *lib);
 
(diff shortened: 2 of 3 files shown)


pitbuster:
If can sort your points by raster order (from bottom to top and then left to right for same y) you may improve the epsilon neighbors calculation like this (sorry for the pseudo-code, but on any question feel free to ask)

```
min_y <- min possible y where an epsilon_neighbor could be located
max_y <- max possible y where an epsilon_neighbor could be located

for y in [min_y, max_y]
    min_x <- min possible x for an epsilon_neighbor at height y
    max_x <- max possible x for an epsilon_neighbor at height y
    look for points in the interval [min_x, max_x] using binary search
    now test that the reduced list of points are actually epsilon_neighbors
```

phweyland:
Good idea. I've thought we could reduce the calculation time dividing the numbers of images into several pieces.
Your suggestion makes sense. I'll make a try on this.

phweyland:
With all other parameters equal, filtering the longitude divides the response time by two. This already very good but not enough.
And as is, doesn't work with both longitude and latitude.
That's probably where the usage of a tree of points can help. I'll look further on this.

pitbuster:
You need to do the binary searches to filter in the other coordinate. I've implemented this kind of geometric query a few times before, and generally there is no need to complicate things with a tree if you do that.

phweyland:
> You need to do the binary searches to filter in the other coordinate.

Sorry, I don't understand this. What do you mean by "binary search" ? Any clarification would be welcome.Thanks.

pitbuster:
I mean this https://en.wikipedia.org/wiki/Binary_search_algorithm

The idea is that you can search the first point _greater or equal_ to (xmin, y) and the last point _less or equal_ to (xmax,y) in logarithmic time, thanks to the points being sorted.

phweyland:
Got it. Thanks

phweyland:
But, the binary search for latitude cannot work on a segment of points(longitude, latitude) sorted by longitude then by latitude, as latitude are not sorted on that segment... 
To be sure to get all points within latitude bounds I always needs to scan each point of the segment.
Am I mistaken ?

pitbuster:
I am not sure if I am getting your question, but you need to perform several binary searches, as the window you need is defined by several intervals in the raster order (as shown in the diagram below)

image

phweyland:
I get a list of points (longitude, latitude) ordered by longitude then (for each longitude) by latitude.
Scanning the list by ordered longitude, following principle you have suggested, I've succeeded to decrease the time by two (eliminating from the potential neighbours those further than epsilon).

The remaining question is about latitude. As latitudes are sorted for each longitude value, on the former neighbourhood list of points (by longitude), latitudes are actually not ordered at all. Then this is not adequate for a binary search, instead I have to scan each point of that list.

A good news however, the accuracy of the haversine distance is not necessary to show groups of images on the map. Then using only the test of distance on longitude and latitude (no square nor sinus calculation) I'm now below 1.5 for 30,000 images, which may be more acceptable.

pitbuster:
Each pair of brackets in the diagram above represents an interval in raster order (meaning points that are greater than a lower_bound and lower than an upper bound in **that** order), since all of those points have the same longitude. In other words, for a fixed longitude, you can get the neighbors and the you just iterate on the possible longitudes (which is what I was trying to convey in the first pseudo-code).

phweyland:
> Each pair of brackets in the diagram above represents an interval in raster order (meaning points that are greater than a lower_bound and lower than an upper bound in **that** order), since all of those points have the same longitude. In other words, for a fixed longitude, you can get the neighbors and the you just iterate on the possible longitudes (which is what I was trying to convey in the first pseudo-code).

I think I understand now.
Once we've got a list of points with longitude inside an interval, we order this list by latitude and then execute the binary search.
It's possible I could save a bit more time ...

phweyland:
> It's possible I could save a bit more time ...

As it is actually done (without distance routine), it  won't because we have to go through the full list to order it. Currently I just select the points which are in latitude interval and that's done. Ordering the list will take more time.

phweyland:
@turbogit, thanks to @pitbuster's help I think we are not too bad now.
The algo is an approximation but it's enough to provide meaningful images clusters.
The max time I've now for 32,000 images is around 1.2 sec.
I've reshaped a bit the execution of the map display, not to stuck the UI and let the progressive update going on. 
I hope this will be fine for you.
Thanks

TurboGit:
Thanks, I'll review that soon.

phweyland:
I've rebased but I still get this error:
image

Any direction to solve this ?

pitbuster:
@phweyland I realized was always assuming the coordinates are integer (which is required for my approach to work), and I think a lot of the misunderstanding come from there :(. While it can be extended to floating point coordintes, now I think trying a tree structure may be interesting.

phweyland:
First travis error fixed.
Now I get this:
```
E: Failed to fetch https://packagecloud.io/github/git-lfs/ubuntu/dists/focal/InRelease  402  Payment Required [IP: 54.193.34.251 443]
E: The repository 'https://packagecloud.io/github/git-lfs/ubuntu focal InRelease' is no longer signed.
Error: The operation was canceled.

```

phweyland:
> now I think trying a tree structure may be interesting.

that's another way to explore indeed.

I would let this merge first. Then I may investigate a bit more about dbscan on a tree, hoping the time to build  it is normally compensated by the rest of algorithm. This should be the case when one increases a lot the epsilon value.

phweyland:
Rebase fails on this, at least I don't succeed in doing it well.
I've copied the changes manually, hopping I haven't forgotten anything.

I've also made some tests with a 2dtree. While it seems the initial tree build is quite fast, the resulting dbscan is still 3 times slower than simple dbscan version. But as I'm discovering this algo I may be wrong in diverse parts ...
I haven't integrated it.

HansBull:
spatilite extension has a `Distance()` function that might be used to generate clusters based on a minimum distance according to the zoom level. But this would add one more dependency and require a change of the `latitude` and `longitude` columns in the db.

TurboGit:
@phweyland : Is that ready for review ? I've lost track...

phweyland:
> @phweyland : Is that ready for review ? I've lost track...

From my standpoint, yes, it is.

Probably not yet the best and the most efficient way to provide clusters of images, but already better than the current one.

If in the future we find a better way to group images (as kd-tree or spatialite (R-tree)), it should be possible to replace this one transparently, at least from user point of view.

src/libs/map_settings.c
```diff
@@ -154,6 +176,24 @@ void gui_init(dt_lib_module_t *self)
 
   gtk_box_pack_start(GTK_BOX(self->widget), GTK_WIDGET(hbox), TRUE, TRUE, 0);
 
+  if(!dt_conf_key_exists("plugins/map/epsilonfactor"))
```
TurboGit, line 179: I suppose here we better have an entry in darktableconfig.xml.in. The type float should have the default value 0.25 and most importantly a min / max to ensure proper recovery when a value is corrupt. This is recent work, see for example ce34dc336895d937aaa7e4afe6f070e3e65e1895. And then you don't need to check for existence for this key.

src/libs/map_settings.c
```diff
@@ -154,6 +176,24 @@ void gui_init(dt_lib_module_t *self)
...
+    dt_conf_set_float("plugins/map/epsilonfactor", 0.25);
+  d->epsilon_factor = dt_bauhaus_slider_new_with_range(NULL, 0.1, 1.0, 0.05,
+                            dt_conf_get_float("plugins/map/epsilonfactor"), 2);
+  gtk_widget_set_tooltip_text(d->epsilon_factor, _("modify the physical size of an images group on the map"));
+  dt_bauhaus_widget_set_label(d->epsilon_factor, NULL, N_("group size factor"));
+  g_signal_connect(G_OBJECT(d->epsilon_factor), "value-changed", G_CALLBACK(_epsilon_factor_callback), self);
+  gtk_box_pack_start(GTK_BOX(self->widget), GTK_WIDGET(d->epsilon_factor), TRUE, TRUE, 0);
+
+  if(!dt_conf_key_exists("plugins/map/minimages"))
```
TurboGit, line 188: likewise.

src/views/map.c
```diff
@@ -2177,6 +2279,145 @@ GSList *mouse_actions(const dt_view_t *self)
...
+}
+
+static void _dbscan_spread(unsigned int index)
+{
+  db.spreads->num_members = 0;
+  _get_epsilon_neighbours(db.spreads, index);
+
+  if(db.spreads->num_members >= db.minpts)
+  {
+    geo_position_t *d;
```
TurboGit, line 2349: remove here...

src/views/map.c
```diff
@@ -2177,6 +2279,145 @@ GSList *mouse_actions(const dt_view_t *self)
...
+    for(int i = 0; i < db.spreads->num_members; i++)
+    {
+      d = &db.points[db.spreads->index[i]];
```
TurboGit, line 2352: geo_position_t *d = ...

TurboGit (changes requested):
Some changes, just reviewed not yet tested.

src/libs/map_settings.c
phweyland: Is darktableconfig.xml.in the only file to set up (including for min max) ?

phweyland: no, but I don't find it ... I remember I've already got difficulty yo find it ...

src/views/map.c
phweyland: ok

src/libs/map_settings.c
TurboGit: It is here : `data/darktableconfig.xml.in`

phweyland: I've run the build.sh script and I got it.
Is there a place (a way) to define the number of digits and the increment in config (it is .001 currently) ?
image

TurboGit: Well it looks like you are using the generated for from the template. The file to modify is committed and so not generated.

No there is no way to describe the increment. This must be done in dt.

phweyland: > No there is no way to describe the increment. This must be done in dt.

Where ? in gui/preferences.c ?
I remember I had to code something to wrap folder configuration for lut3d but I've forgotten where.

TurboGit: For preferences I don't think we have a way to set the increment.

phweyland: Thanks. I've moved to integer to avoid the problem.

phweyland:
> spatilite extension has a `Distance()`

Thanks a lot for this interesting reference.
At first glance spatialite has no function to calculate cluster directly on the db, but I'll investigate a bit more to be sure.
It would be interesting to have this function directly against the db. But geotags are closely tight to images (and their optimizations), it could be tricky to change the structure.

HansBull:
> At first glance spatialite has no function to calculate cluster directly on the db, but I'll investigate a bit more to be sure.

This might give some hint: https://groups.google.com/g/spatialite-users/c/LIuCaAeAYok

> It would be interesting to have this function directly against the db. But geotags are closely tight to images (and their optimizations), it could be tricky to change the structure.

After creating the point geometry column, populate it with the values from the existing existing `longitude` and `latitude` columns (which could also stay, after all). Goal should be to get rid of filling the list from line 1038 to 1049 in maps.c (if you have 30k images, this means 30k times calls to `calloc` and `g_slist_prepend` just to iterate over the list in a second step to find groups). Other thing, the db currently does not seem to have a index for `longitude` and `latitude`.

TurboGit:
Works for me now. And sorry I wasn't clear the preferences do not need to get an entry into the preference dialog. I will remove the section specification. In this case you have sliders in the lib module to change them. There is a huge number of preference that have no gui entry in the dialog, they are just documented there nad permit now to set the default and have corruption recovery.

Anyway, merging that.

phweyland:
> I will remove the section specification. In this case you have sliders in the lib module to change them. There is a huge number of preference that have no gui entry in the dialog, they are just documented there nad permit now to set the default and have corruption recovery.

Perfect. Thanks.


> Goal should be to get rid of filling the list

You are right about the goal, but the lines you are talking about are related to images of groups, which normally are below 20 ...
The images list scanning is made at the bottom of the file, and the number of malloc() is only 3.


For the reference this algo has the following limitations (therefore there is some room for improvement :)):
- calculation made on the projection map (no haversine)
- no distance calculation neither, only latitudes and longitudes comparison (no power operations)
- epsilon has not the same weight for longitude and latitude, and its weight varies with latitude.

PR #6743, opened 2020-11-03
fix colorbalance sliders unhiding

### LINKED ISSUE #6739: color balance displays both sets of sliders
Every time I switch from lighttable to darkroom, color balance displays **both** sets of sliders even if only
HSL is selected (see screenshot).
After clicking on the color control sliders pull down menu, it starts working again as expected: _i.e._ the correct set of sliders is displayed.

However, after going back to lighttable and then darkroom again, the same issue reappears.

This issue started happening only few days ago.

dt version: 3.3.0~git1347.9bba79445
os: linux

ss

dterrahe:
fixes #6739


TurboGit (approved):
Works for me, thanks !

PR #6745, opened 2020-11-04
[CI/CD] Github Action Fixes - nightly, job dependencies and IRC notification

johnny-bit:
This PR fixes mis-copied windows nightly builds (now shouldn't fail) and also:

- Makes sure that windows/osx depend on linux builds, so that we don't waste win/osx resources while knowing linux build had failed
- ~run integration tests to confirm that dt is indeed running after compile [on linux, will see how it goes for other OSs]~
- On successfull push it adds irc notification. might be cool.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -7,8 +7,135 @@ on:
     branches: [ master ]
 
 jobs:
+
+  Linux:
+    name: Linux.${{ matrix.os.code }}.${{ matrix.compiler.compiler }}.${{ matrix.target }}.${{ matrix.btype }}
+    runs-on: ${{ matrix.os.label }}
+    strategy:
+      fail-fast: true
+      matrix:
+        os: 
+          - { label: ubuntu-20.04, code: focal }
+          #- { label: ubuntu-18.04, code: bionic }
+        compiler:
+          #- { compiler: GNU8,   CC: gcc-8,    CXX: g++-8,      packages: gcc-8 g++-8 }
+          - { compiler: GNU9,   CC: gcc-9,    CXX: g++-9,      packages: gcc-9 g++-9 }
+          - { compiler: GNU10,  CC: gcc-10,   CXX: g++-10,     packages: gcc-10 g++-10 }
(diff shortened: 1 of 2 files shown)


johnny-bit:
Well... Damn...
For some reason there's a problem in integration tests. They are obviously expensive time-wise, so need to be ran in very few selected cases (eg release maybe triggered manually or weekly...)

Another set of problems:
- LLVM10 RelWithDebInfo build: 2 errors, just pixel problems on filmicrgb and color mapping
- LLVM10 Release build: 4 errors. segfault in denoise/denoiseprogiled nlmeans x2 and those 2 errors above.
- GCC 9/10 full builds: hang in nlmeans
- GCC no-features builds: no problem ;)

Annd i'd love if integration test driver (run.sh) could have either "fast-fail" mode where first NOK/ dt-cli segfault would result in full test stoppage or fail-return where run with detected errors would have non-success exit.

Disabling integration tests for now.

TurboGit:
> Annd i'd love if integration test driver (run.sh) could have either "fast-fail" mode where first NOK/ dt-cli segfault would result in full test stoppage

I can do that indeed. I can propose also adding a single test using the most common modules that could be run in the CI. Would that help ?

TurboGit:
@johnny-bit : the --fast-fail option is implemented.

johnny-bit:
@TurboGit - Thanks! 

> I can propose also adding a single test using the most common modules that could be run in the CI. Would that help ?
It certainly would imo. NO-OP test is ok for "it wurkz" assumption but "most common modules" could be interesting in terms of process testing...

> --fast-fail option is implemented.
That's wonderful!

Thank you!

TurboGit:
@johnny-bit : I have added a all-in-one test using many module, you can at least run this test on the CI.

The modules activated in this test are:

- sharpen
- filmicrgb
- exposure
- flip
- rgbcurve
- clipping
- bilat
- colorbalance
- liquify
- retouch
- lens
- denoiseprofile
- borders
- watermark

johnny-bit:
Great!

I'm not ready to do PR nor have a time so just a quickie if you don't mind:

here:
https://github.com/darktable-org/darktable/blob/3f3ef161b430a4293676ccb27c7eef20c9c8c684/.github/workflows/ci.yml#L163-L169

uncomment first if (usermanual) and comment out 2nd if (false) and add specific test to run in line 169 and it should work fine :)

Issue #6707, opened 2020-11-01
'do high quality processing for slideshow' tooltip


kofa73:
The tooltip says 'same option as for export, but applies to slideshow', which is quite unhelpful. I was unable to find a related 'export' setting in preferences; in the options of the export module on the lighttable, there's a 'high quality resampling', is that the one this is referring to?
The documentation for 'do high quality processing for slideshow' says 'Controls how images are processed for the slideshow view. It this option is enabled, the image will first be processed in full resolution, and downscaled at the very end. This can result in better quality sometimes, but will always be slower (default on).' 'high quality resampling' is not documented at https://darktable.gitlab.io/doc/en/export_selected.html; the new manual (https://elstoc.github.io/dtdocs/module-reference/utility-modules/lighttable/export-selected/) explains the setting as 'Set this to ‘yes’ to perform high quality resampling on the image. The image will be processed in full resolution and only downscaled at the very end. This can sometimes result in better quality, but will always be slower.'

kofa73:
Config option no longer available

PR #6746, opened 2020-11-04
Remove `plugins/slideshow/high_quality` preferences

jenshannoschwalm:
This pr is a suggestion and a step to reduce the overwhelming size of the preferences.

I think instead of the extra `plugins/slideshow/high_quality` we could be using the
`ui/performance` flag too.

Related #6707


TurboGit:
Works for me. Thanks!

PR #6747, opened 2020-11-04
map: add view all locations button

phweyland:
Gives the possibility to show/hide all the locations which are visible on the map.

Plus a small change on select location: when this location is already on the map, do not move the map.

### DIFF
--- a/src/common/database.c
+++ b/src/common/database.c
@@ -45,7 +45,7 @@
 // whenever _create_*_schema() gets changed you HAVE to bump this version and add an update path to
 // _upgrade_*_schema_step()!
 #define CURRENT_DATABASE_VERSION_LIBRARY 30
-#define CURRENT_DATABASE_VERSION_DATA     7
+#define CURRENT_DATABASE_VERSION_DATA     8
 
 typedef struct dt_database_t
 {
@@ -1863,6 +1863,13 @@ static int _upgrade_data_schema_step(dt_database_t *db, int version)
 
     new_version = 7;
   }
+  else if(version == 7)
+  {
+    TRY_EXEC("ALTER TABLE data.locations ADD COLUMN ratio FLOAT DEFAULT 1",
+             "[init] can't add column `ratio' column to locations table\n");
+
+    new_version = 8;
+  }
   else
     new_version = version; // should be the fallback so that calling code sees that we are in an infinite loop
 
@@ -2044,9 +2051,8 @@ static void _create_data_schema(dt_database_t *db)
                NULL, NULL, NULL);
   ////////////////////////////// (map) locations
   sqlite3_exec(db->handle, "CREATE TABLE data.locations (tagid INTEGER PRIMARY KEY, "
-               "type INTEGER, longitude REAL, latitude REAL, delta1 REAL, delta2 REAL, "
+               "type INTEGER, longitude REAL, latitude REAL, delta1 REAL, delta2 REAL, ratio FLOAT, "
                "FOREIGN KEY(tagid) REFERENCES tags(id))", NULL, NULL, NULL);
-
 }
 
 // create the in-memory tables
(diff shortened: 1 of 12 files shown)


phweyland:
@TurboGit I'm done for this one.

I've doubled the line size of main location to highlight it amongst the others when present. I'm not sure it's enough however.
We could also use another color. Let me know.
Thanks

TurboGit:
A little issue:
- create some locations
- click [show all]
- select one location : OK it is highlighted
- zoom-in the map with a single mouse scroll
- select another location : NOK - new one is highlighted but the previous is not visible anymore

Also I would certainly make the highlighted location more visible, maybe with an even thicker line ?

TurboGit (changes requested):
Minor issue found. I'll review code later.

phweyland:
Good catch ! I thought about this case but, .... I did forget it.
Thanks

phweyland:
Idea. Clicking on another location (not highlighted) on the map could select it and the right record in the location list as well.
Would you find that handy ?

TurboGit:
> Would you find that handy ?

Yes, would be nice to have this. In this mode (when area are displayed on map) you cannot move the picture, right ? Otherwise it will be quite difficult to get right I suppose.

phweyland:
> In this mode (when area are displayed on map) you cannot move the picture, right ?

When an image is inside the main location, ctrl-click allows to take it and move it.
At the end all is question of order in which we decide to treat the events to give expected behavior. 
But you are right, we have to check at the end if that works without too much headache.

phweyland:
Done.

TurboGit:
Some issues with location, maybe not related to this PR, but probably worth fixing it now as part of the PR.

First image, we start from here, tags are ok:

image

We move the face on the right to a new location, but it receive the tag **p1** even though it is not inside the circle:

image

phweyland:
Interesting indeed.

The same angle doesn't give the same pixel size in longitude and latitude.
Said another way, a constant angle should be displayed as a (vertical) ellipse.
Or, to keep a circle on the map, the test of inclusion should be a horizontal ellipse.
That the way I've just tried but that a bit complicated and doesn't improve the response time (more calculation).
Of course the form of ellipse depends on latitude.
The simpler way would be to draw an ellipse (at Dieppe the ratio is 0.65 ...) instead of a circle and to keep the queries on circle detection.
On the other hand he advantage to draw a circle and detect on an ellipse is that we are not far to let the user transform the ellipse vertically or horizontally as we do for the rectangle...
What are your thoughts ?

TurboGit:
I think that for the user the simpler and more friendly solution is to display a circle. An ellipse with a different ratio depending on the area on the map will certainly raises questions if not frustrations.

So let's go with display a circle and detect on an ellipse. TIA.

phweyland:
Hoping this works in Dieppe area now ... :)

TurboGit:
Well Dieppe is the only place where it needs working anyway :)

phweyland:
Around the cap Horn it still works, but north of Groenland it is not as accurate as in Dieppe...
The ratio latitude/longitude is close to 1 around the equator, and increases moving to the poles until 5 or 6 ...
And it seems not to be totally symmetric.

TurboGit:
Looks like a new issue has been introduced, now I cannot see any tag added when moving a picture into a location in the map.

phweyland:
I guess you are talking about existing locations. The data in the database are not exactly compatible.
Normally, if you move (shift) a bit the location, its data and its images are recalculated. Then the detection should work again. Could you check ?
At time of db migration, osmgpsmap is not yet initialized so we cannot convert properly the data.
Should I create a new version of location plugin to do this ?

TurboGit:
> I guess you are talking about existing locations. The data in the database are not exactly compatible.

Indeed, so all good. No need for a new db version as we are talking about dev versions. Redoing the locations is ok at this stage.

TurboGit (commented):
Let's go... Looks ok to me now. Thanks !

PR #6756, opened 2020-11-05
update toneeq documentation

rawfiner:
Update the toneequalizer documentation with these new changes: eigf mode, new presets, and the problem of slowdown
when used with denoising algorithm is fixed

Note that I did not manage to build the documentation with the build script unfortunately, even if cmake does not give any error

### DIFF
--- a/doc/usermanual/darkroom/modules/grading/toneequalizer.xml
+++ b/doc/usermanual/darkroom/modules/grading/toneequalizer.xml
@@ -186,6 +186,31 @@
     </sect5>
 
     <sect5>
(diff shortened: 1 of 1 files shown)


TurboGit (approved):

TurboGit:
@rawfiner : the templates download from sourceforge are a pain! Well sourceforge is a pain actually, and the number of download of the xslt templates are limited. The safest way is to call the `build-doc.sh` script with `--jobs 1`.

And sometimes you need to call the script multiple times.

I'm not even sure how sourceforge can even still exists!

PR #6757, opened 2020-11-05
Allow to hide/show the borders controls on the slideshow.

TurboGit:


AlicVB (approved):
OK for me.
A first, it doesn't seems to works, but finally it does, no idea why... Maybe a second start was needed, or a shortcut change ?

TurboGit:
Thanks !

PR #6762, opened 2020-11-06
imageio_tiff: enable shortfile w/ masks, misc fixes

kmilos:
Fixes a rare potential bug: MASKS_USE_SAME_FORMAT currently hard-codes 3 channels being stored for the mask, but a new "shortmode" was introduced to store grayscale image data as a single channel in 3.2.1; this keeps the image and mask channels in sync.

This PR also fixes a potential bug when MASKS_USE_SAME_FORMAT is not defined for some reason, ensuring the compression predictor matches the hard-coded floating point mask data format instead of the predictor set based on image data format.


TurboGit:
Manually merged into master, queued for 3.4.1.

kmilos:
Merci

PR #6772, opened 2020-11-07
Prevent drawing of selection triangle in zoomable thumbs

### LINKED ISSUE #6637: flag selected thumb (triangle on top) visible on lighttable
**Describe the bug**
flag selected thumb (triangle on top) visible on lighttable

**To Reproduce**
1. Go to lighttable
2. Click X to enter the culling mode
3. Click X to leave the culling mode
4. Fly over the 2 thumbs which where displayed in culling view (to refresh the display)
5. You'll see the little triangle on top

This triangle should only be displayed on the filmstrip.

**Expected behavior**
no triangle visible on lightatble

**Screenshots**

image

 - Darktable Version: master
 - OS:GNU/Debian

jenshannoschwalm:
The "selected thumb" marker should only be shown for filmstrip thumbs.

As we don't have information in `dt_thumbnail_t` who owns the thumb we
use the zoomable gboolean instead.

@AlicVB what do you think about this?

Hopefully fixes #6637

### DIFF
--- a/src/dtgtk/thumbnail.c
+++ b/src/dtgtk/thumbnail.c
@@ -265,6 +265,8 @@ static gboolean _event_cursor_draw(GtkWidget *widget, cairo_t *cr, gpointer user
   if(!user_data || !widget) return TRUE;
   dt_thumbnail_t *thumb = (dt_thumbnail_t *)user_data;
 
+  if(thumb->zoomable) return TRUE;
+
   GtkStateFlags state = gtk_widget_get_state_flags(thumb->w_cursor);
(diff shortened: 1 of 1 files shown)


AlicVB:
hmm... that will not change anything for #6637 imo as thumbnail in lighttable are not zoomable either :(
And anyway, it would be better to fix the problem at its root : if the triangle is shown, that means that main thumbnail widget is flaged as active, which shouldn't happen outside filmstrip...

I'll have a look, and anyway thanks a lot lot for having looked into this !

jenshannoschwalm:
Thanks, couldn't find the root cause ...

PR #6779, opened 2020-11-07
introspectise channelmixer rgb

dterrahe:
Please test especially the last changes that were force pushed before the module got merged, because I had to rebase off those mostly manually and without a clear changelog. I (or git) could easily have broken some of them, despite trying to carefully review.

Also please check behavior of illuminant combobox when switching between jpg and raw images; I assumed the last ("as shot in camera") option needs to be dynamically removed/added (rather than statically in gui_init).

### DIFF
--- a/src/common/chromatic_adaptation.h
+++ b/src/common/chromatic_adaptation.h
@@ -23,11 +23,11 @@
 
 typedef enum dt_adaptation_t
 {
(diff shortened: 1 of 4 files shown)


dterrahe:
Also check out the huge number of shortcuts set up for (almost; I specified sections) free...

TurboGit:
Ok, let's merge this now. It works on my side. Thanks.

PR #6780, opened 2020-11-07
German translation updates

rabauke:
yet another update to (almost) complete the translation


TurboGit (approved):
Thanks !

TurboGit:
@rabauke : note that channelmixerrgb was not yet added into the list of file to be translated. I just did that and you'll see that there is many new strings to translate :)

PR #6782, opened 2020-11-08
rename channel mixer RGB to color calibration

aurelienpierre:
People are starting to get confused about a chromatic adaptation transform (CAT) in the channel mixer.

Well…

Your input color profile is nothing but a channel mixer in which the coefficients are already preset (aka it's a 3×3 matrice).

Your white balance is nothing but a channel mixer in which you only set the red/red, green/green and blue/blue coefficients (aka the diagonal of the 3×3 matrice). What any fancy temperature + tint does is precomputing these 3 coeffs from 2 parameters.

So if you think about it from a broader perspective, channel mixer lets you correct your input profile and white balance, by scaling and rotating the primaries in 3D.

That is color calibration. So I propose to name it that way, because people care more about the label than the action.

### DIFF
--- a/src/iop/channelmixerrgb.c
+++ b/src/iop/channelmixerrgb.c
@@ -124,7 +124,7 @@ typedef struct dt_iop_channelmixer_rbg_data_t
 
 const char *name()
 {
-  return _("channel mixer rgb");
+  return _("color calibration");
 }
 
 int flags()


elstoc:
Given that this has only just been merged, IMO we should rename the file too

aurelienpierre:
No, me and other people have been using the module in prod for several months, changing the name of the file will break their edits.

TurboGit:
I have no strong opinion but `channel mixe`r was a good name to me. Most of the tab are identical to the current channel mixer. So let's people be confused I would say and drop this PR :)

Nilvus:
Channel mixer rgb is good for me. But I like color calibration name. This channel mixer is more than just a channel mixer (related to historical one we have) and indeed it's a channel mixer and white balance one, so color calibration is appropriate.

Just a thing that could also be related: in preset manager added by @AlicVB there's recommended and other modules for each group. By now, channel mixer rgb is not a recommended group. It should be good for 3.4 and not advanced users to add it in a recommended group (probably grading/color ones!)

elstoc:
I'm fine with it being channel mixer but to prevent confusion it might be better to present the gui as a 'traditional' channel mixer - i.e. put the CAT tab at the end and have it default to settings that perform a no-op in terms of white balance (the current settings adjust the image's white balance by default)

aurelienpierre:
CAT can't go at the end since it's the first processing step and defines in what space the whole module will work.

parafin:
How often does default colorspace need changing? Maybe leave CAT as first tab, but select second tab by default?

aurelienpierre:
I use the module for the CAT, mostly, so it's often the only tab I open.

elstoc:
Remember last open tab

aurelienpierre:
That's an ok solution. Can you code it ?

elstoc:
I probably could, yes

aurelienpierre:
> I probably could, yes

That would be awesome.

To all, can you cast your vote with thumbs up/down on the first message ?

elstoc:
Sorry to clarify are we voting on the rename?

aurelienpierre:
On the rename, yes.

TurboGit:
Let's have a poll:

<p><a href="https://api.gh-polls.com/poll/01EPM73Q1964ND67NPVAHZHTJB/keek%20name%20%22channel%20mixer%20rgb%22/vote"></a>
<a href="https://api.gh-polls.com/poll/01EPM73Q1964ND67NPVAHZHTJB/rename%20to%20%22color%20calibration%22/vote"></a></p>

aurelienpierre:
Github is full of surprises

elstoc:
See #6793.

TurboGit:
@aurelienpierre : to create a poll, go here : https://app.gh-polls.com/ generate XML and paste it here.

aurelienpierre:
Ok the motion carries with 16 votes to 3.

PR #6783, opened 2020-11-08
add example distort_transform and distort_backtransform

ralfbrown:
While looking into a performance bottleneck involving the distort_transform, I noticed that useless.c had no mention of either distort_transform or distort_backtransform, though it does briefly mention the related distort_mask.

This PR adds sample implementations of the two functions with explanatory comments to useless.c.


TurboGit:
Thankks !

PR #6785, opened 2020-11-08
refactor: merge duplicated code in liquify.c back into develop.c

ralfbrown:
liquify.c contains a complete copy of dt_dev_distort_{,back}transform_plus as it existed before the preview downscaling was implemented, less the history locking (to permit recursive use).

This PR factors out the common code into new functions in develop.c and removes the duplicate functions from liquify.c.  It also adds an integration test for the Liquify iop.


TurboGit (approved):
Works for me, thanks !

PR #6792, opened 2020-11-08
Showing duplicates must be corrected for ppd_thb

### LINKED ISSUE #6790: Duplicate preview is ill-zoomed
In darkroom, if you have several editing duplicates, you can hold your click on one of them thumbnails, and the preview of the duplicate will overlay the current edit.

Currently, the size of the duplicate doesn't match the size of the current preview, it seems there is a wrong zoom factor and the duplicate is magnified compared to the original.

I'm on High DPI screen, at 4K and pixel factor of 2 (1 px GUI -> 2 px device).

jenshannoschwalm:
Fixes #6790


TurboGit:
Another ppd issue fixed ! Works for me, thanks!

PR #6793, opened 2020-11-08
channelmixerrgb: remember last selected notebook page

elstoc:


src/iop/channelmixerrgb.c
```diff
@@ -1915,6 +1915,9 @@ void gui_init(struct dt_iop_module_t *self)
   NOTEBOOK_PAGE(grey, grey, N_("grey"), N_("grey"), FALSE)
 
   self->widget = GTK_WIDGET(g->notebook);
+  int active_page = dt_conf_get_int("plugins/darkroom/channelmixerrgb/gui_page");
```
TurboGit, line 1918: const

elstoc: ok. I copied this from tone equalizer, so have added a const there as well.

TurboGit:
BTW, I'm not sure this is good. When I use this module (or previous channel mixer BTW), when I change the R channel I do not expect to restart here for next edit. So not really convince this is helpful.

elstoc:
I don't want to start in the CAT tab either, because I mostly use it for monochrome conversion

ptilopteri:
make it a preference within the module

johnny-bit:
I think this is in the same vein as recent toneeq change, where last used tab is remembered. It's imo good practice for multi-tab modules, as it gives good workflow by default for those that are tab-independent (like toneeq and channelmixer). I wouldn't recommend remembering tabs on filmic though ;)

TurboGit:
Ok, I let's go with that and we"ll see how people react. Note that I'm not for a preference as we already have many, or if a preference is to be added it should be applied to ALL modules.

matt-maguire:
Using channel mixer for making monochrome images or for manually adjusting primaries are not common use cases, whereas CAT tab is relevent for all images. That's why I think it should be first and selected by default.

elstoc:
Back to that discussion about the assumptions that the module is making. It should either be a no-op when activated or it should somehow set the white balance module to D65.

rabauke:
I used channel mixer rgb for the 1st time today. I think it's great. But to things a confusing (at least at the beginning)

- The module combines the functionality of a "classic" channel mixer with white balancing.
- The default setting represents not a no-op. (Source should be "same as pipeline (D50)")

elstoc:
@rabauke: exactly.

PR #6782 should partially resolve this so that people aren't expecting a 'traditional' channel mixer. There is still some work needed to integrate this module into a reasonable workflow (namely how to ensure that the WB module is set to D65 when the CAT tab is used).

parafin:
Not all modules have default settings resulting in no-op, so it's not a rule per se. The question is how particular module is usually used, if returning to no-op settings is something that people want to do often, then it makes sense to change defaults I guess.

dterrahe:
> I think this is in the same vein as recent toneeq change, where last used tab is remembered.

I rather belatedly realise now (while chasing down a crash for @AxelG-DE in midi) that this is somewhat problematic for multi-instance (more conceptually rather than implementation). The last gui_cleanup called will set the saved tab. So if different tabs are active in several instances, the saved state might be unpredictable. Could be (I haven't experimented/verified this) that if you create a 2nd instance, change tabs there, then close it, that tab maybe gets saved. Now if you create another instance again, it would use that saved state, rather than whatever tab the instance you duplicate from is on.

Just to mention this and see if anybody has a simple obviously better (and working!) solution. I don't think this is critical; if you _always_ want to use a particular tab that'll end up being the one that gets saved and restored anyway.

PR #6794, opened 2020-11-08
tagging: fix completion issue

### LINKED ISSUE #6753: [DT 3.3.0-XXX] Tag cannot be cleared from the textbox that was used to enter using the "x" button.
To reproduce : 
There are no existing tags in the system (except the ones DT creates internally).
Adding a new tag.
Enter text in text box under the attached button.
Hit enter.
Tag gets created.
The created tag is still present in the user entry text box.
Now try and hit the "x" to remove the tag that was typed for creation. It does not get cleared. However you can backspace and clear it.

image

phweyland:
Fixes  #6753

### DIFF
--- a/src/libs/tagging.c
+++ b/src/libs/tagging.c
@@ -2566,7 +2566,7 @@ static gboolean _completion_match_func(GtkEntryCompletion *completion, const gch
   {
     lastTag = key;
   }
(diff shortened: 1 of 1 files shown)


TurboGit (approved):
Thanks !

PR #6795, opened 2020-11-08
common: Fix DT_FMA macro

cryptomilk:
Build error:
```
    src/common/darktable.h:199:37: error: expected ')' before ';' token
        #define DT_FMA(x, y, z)  x * y + z;
                                         ^
```

### DIFF
--- a/src/common/darktable.h
+++ b/src/common/darktable.h
@@ -196,7 +196,7 @@ static inline int dt_version()
 #ifdef FP_FAST_FMAF
   #define DT_FMA(x, y, z) fmaf(x, y, z)
 #else
-  #define DT_FMA(x, y, z)  x * y + z;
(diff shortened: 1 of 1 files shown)


cryptomilk:
llvm build failed, but I don't really see where. Searching for `error:` doesn't show a compile error. I this ad po4a-translate issue?

parafin:
You need to remove ; at the end of the line also.

cryptomilk:
What the heck, I've removed it. It magically appeared here again, I swear!

TurboGit:
Nice bug fix, hopefully we weren't in a bad condition when using this macro in dt !

jenshannoschwalm:
About your swearing... some smart gedit extension at work?

PR #6797, opened 2020-11-08
[Issue #6668] Define sorting order in collect module instead of prefs

quovadit:
As suggested in #6668 I've moved the definition of the sorting order from preferences to the collect module.

At the moment it uses the same pref-setting as before and therefor only affects folder, film-roll sorted by folder and dates

sort-popup-committed

### DIFF
--- a/data/darktableconfig.xml.in
+++ b/data/darktableconfig.xml.in
@@ -540,13 +540,6 @@
     <shortdescription>sort film rolls by</shortdescription>
     <longdescription>sets the collections-list order for film rolls</longdescription>
   </dtconfig>
-  <dtconfig prefs="lighttable">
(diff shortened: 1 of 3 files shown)


quovadit:
@TurboGit : Is there a chance that this will be merged as is, or should I extend the sorting to all properties (iso, aperture, ...)?

TurboGit:
Not for 3.4 as there is some questions yet after I had a quick look and I don't have more time for 3.4 new feature. This will be for next minor release hopefully so it won't be a wait of one full year.

quovadit:
minor release would be great, thanks!

TurboGit:
This has been supersede by #7530.

PR #6802, opened 2020-11-08
Enable internal Lua fallback

wpferguson:
darktable uses Lua 5.3.  Lua 5.4 is out but not widely adopted by distributions, so far.  darktable's transition to Lua 5.4 is a work in progress currently targeted at darktable 3.6.  Enabling the fallback internal Lua will ensure that Lua is available regardless of what version of Lua (5.3 or 5.4) is installed on your system or what version of Lua darktable is using.


parafin:
Who are the intended beneficiaries of changed defaults? I mean, people who self-compile can probably find this setting themselves, while this change could be a bad surprise for packagers if they didn't explicitly set this setting before...

wpferguson:
> this change could be a bad surprise for packagers if they didn't explicitly set this setting before...

I thought of that as I was writing this.  The intended target is the self-compilers since there have been a couple of issues opened about it.  My thought was packagers only compile once, while self-compilers compile once a day or more. I'm not really sure how the packagers compile, but I would think when they build they have Lua available otherwise Lua support wouldn't be compiled in.  I guess the only surprise would be Lua support is always compiled in unless you specify explicitly disable it, i.e. --disable-lua.

parafin:
The issue is that some distributions are against bundled libraries due to security concerns, so they don't want to ever use internal lua.

TurboGit:
I would prefer having an entry into the RELEASE_NOTES for people compiling dt themselves. @wpferguson : Would that work for you ?

wpferguson:
@parafin I understand and that makes sense.  @TurboGit an entry in the RELEASE_NOTES works for me.  I'll close this, write something up and submit it.

PR #6814, opened 2020-11-09
module order: purge cache one step below

aurelienpierre:
I'm not sure I fully understand how this part of the code works, but I might have find a minor issue. @TurboGit, can you check ?

When moving a module down the pipe, the function finds the "previous" module and insert the current module before this. But then the pipe gets invalidated from the "previous" module, which is one step too late. This could lead to the current module being not flushed from cache properly, and preview updates inconsistencies. So this commit ensures we flush caches from the current module, which is one step before what is done as of today. It is even possible that we must flush even before current module.

Note the `insert_after` works properly, the cache is flushed from under the current module.


TurboGit:
Indeed, this change makes sense to me and you are right the move-up is correct and the equivalent of move-up in move-down is exactly what you propose. So to me this is ok.

TurboGit (approved):
Seems OK to me. Thanks.

PR #6818, opened 2020-11-09
Spanish translation update

EdgarLux:
Updating translation and partially adding new translated strings.

### DIFF
--- a/po/es.po
+++ b/po/es.po
@@ -2,8 +2,8 @@ msgid ""
 msgstr ""
 "Project-Id-Version: darktable 3.4\n"
 "Report-Msgid-Bugs-To: \n"
-"POT-Creation-Date: 2020-09-16 00:36-0500\n"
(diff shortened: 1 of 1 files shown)


TurboGit (approved):
Thanks !

pitbuster:
I noticed that some of the translations don't make any sense (like they are translations from other strings), could it be a problem with github diff rendering or it is a problem with the po file?

EdgarLux:
> I noticed that some of the translations don't make any sense (like they are translations from other strings), could it be a problem with github diff rendering or it is a problem with the po file?

Thanks @pitbuster. I verified it and I didn't find any problem with what you mention. As you may have noticed, there have been many changes, additions and improvements to the darktable code, so translation is still a WIP as well.

For the same reason, it is normal that sometimes some translations change, are missing or need to be improved. If there is a specific one that concerns you, you can let me know, if you prefer, by e-mail.

PR #6823, opened 2020-11-10
add a chroma adaptation workflow

### LINKED ISSUE #6803: channel mixer rgb not neutral
**Describe the bug**

activate the channel mixer rgb changes the image

**To Reproduce**
1. Go to darkroom
2. Active channel mixer rgb
3. See image changes

**Expected behavior**

The module should be a not by default.

 - Darktable Version: master
 - OS: GNU/Linux debian

**Additional context**

Maybe setting CAT adaptation to "none" by default could be a solution ?

aurelienpierre:
fix #6803

Similar in logic to display-referred/scene-referred workflow. Offers 2 worflows : 
* modern
* legacy
Screenshot_20201110_024018

Legacy is the current one. It will set channelmixerrgb defaults to be a simple channel mixer, disabled by default, and will rely on current white balance module to apply the raw WB from EXIF coeffs.

Modern will set channelmixer enabled by default with chromatic adaptation adjusted from raw WB EXIF coeffs and will set the white balance to camera default (D65) only to help demosaicing. Since white balance is handled in channelmixerrgb, the white balance module is stripped from most of its options, only the actual RGB coeff are accessible but toggled off by default.
Screenshot_20201110_023637

### DIFF
--- a/data/darktableconfig.xml.in
+++ b/data/darktableconfig.xml.in
@@ -2380,6 +2380,18 @@
     <shortdescription>auto-apply pixel workflow defaults</shortdescription>
     <longdescription>scene-referred workflow is based on linear modules and will auto-apply filmic and exposure,\ndisplay-referred workflow is based on Lab modules and will auto-apply base curve and the legacy module pipe order.</longdescription>
   </dtconfig>
+  <dtconfig prefs="processing">
+    <name>plugins/darkroom/chromatic-adaptation</name>
+    <type>
+      <enum>
+        <option>modern</option>
+        <option>legacy</option>
+      </enum>
+    </type>
+    <default>legacy</default>
+    <shortdescription>auto-apply chromatic adaptation defaults</shortdescription>
+    <longdescription>legacy performs a basic chromatic adaptation using only the white balance module\nmodern uses a combination of white balance module and color calibration module, with improved color science for chromatic adaptation.</longdescription>
+  </dtconfig>
   <dtconfig prefs="processing" restart="true">
     <name>plugins/darkroom/basecurve/auto_apply_percamera_presets</name>
     <type>bool</type>
--- a/src/develop/develop.c
+++ b/src/develop/develop.c
@@ -1377,6 +1377,10 @@ static gboolean _dev_auto_apply_presets(dt_develop_t *dev)
   gchar *workflow = dt_conf_get_string("plugins/darkroom/workflow");
   const gboolean is_scene_referred = strcmp(workflow, "scene-referred") == 0;
   const gboolean is_display_referred = strcmp(workflow, "display-referred") == 0;
+
+  workflow = dt_conf_get_string("plugins/darkroom/chromatic-adaptation");
+  const gboolean is_modern_chroma = strcmp(workflow, "modern") == 0;
+
   g_free(workflow);
 
   //  Add scene-referred workflow
@@ -1387,16 +1391,18 @@ static gboolean _dev_auto_apply_presets(dt_develop_t *dev)
   const gboolean has_matrix = dt_image_is_matrix_correction_supported(image);
 
   const gboolean auto_apply_filmic = has_matrix && is_scene_referred;
+  const gboolean auto_apply_cat = has_matrix && is_modern_chroma;
   const gboolean auto_apply_sharpen = dt_conf_get_bool("plugins/darkroom/sharpen/auto_apply");
 
-  if(auto_apply_filmic || auto_apply_sharpen)
+  if(auto_apply_filmic || auto_apply_sharpen || auto_apply_cat)
   {
     for(GList *modules = dev->iop; modules; modules = g_list_next(modules))
     {
       dt_iop_module_t *module = (dt_iop_module_t *)modules->data;
 
       if(((auto_apply_filmic && strcmp(module->op, "filmicrgb") == 0)
-          || (auto_apply_sharpen && strcmp(module->op, "sharpen") == 0))
+          || (auto_apply_sharpen && strcmp(module->op, "sharpen") == 0)
+          || (auto_apply_cat && strcmp(module->op, "channelmixerrgb") == 0))
          && !dt_history_check_module_exists(imgid, module->op)
          && !(module->flags() & IOP_FLAGS_NO_HISTORY_STACK))
       {
--- a/src/iop/channelmixerrgb.c
+++ b/src/iop/channelmixerrgb.c
@@ -1762,10 +1762,28 @@ void reload_defaults(dt_iop_module_t *module)
   d->illuminant = module->get_f("illuminant")->Enum.Default;
   d->adaptation = module->get_f("adaptation")->Enum.Default;
 
+  module->default_enabled = FALSE;
+
+  gchar *workflow = dt_conf_get_string("plugins/darkroom/chromatic-adaptation");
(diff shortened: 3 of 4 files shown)


TurboGit:
> Modern will set channelmixer enabled by default with chromatic adaptation adjusted

Aren't you missing a part of change in the commit ? On my side I don't have the color calibration enabled by default.

TurboGit (changes requested):
Double check that color calibration is enabled by default.

elstoc:
Is discarding history the only way to change for a given image

parafin:
> Is discarding history the only way to change for a given image, and how will this affect old edits?

Looking at the code changes to defaults are done in reload_defaults, so they only affect new history stacks - so discarding is the way, and it doesn't affect old edits.

GUI changes in WB module on the other hand I think affect old edits too, which is wrong. That's why on IRC I proposed to also check if parameters match D65 and only then change the UI.

elstoc:
> so discarding is the way, and it doesn't affect old edits

If I have my workflow set to "modern", I open an old edit (using non-D65 white balance) and enable channel mixer rgb without changing anything else what happens?

codingdave:
Small issue: whats modern today can be historic tomorrow. Also it is not descriptive. Are there better words we can use that are self explanatory and stay valid over several years?

aurelienpierre:
Oops, I forgot to commit a file

aurelienpierre:
>  Also these terms are not descriptive. Are there better words we can use that are self explanatory and stay valid over several years?

I can't think of any that would be polite.

aurelienpierre:
@parafin good catch, see latest commit.

SoupyGit:
"Small issue: whats modern today can be historic (legacy) tomorrow.
Also these terms are not descriptive. Are there better words we can use that are self explanatory and stay valid over several years?"

I agree. When module order was introduced, the options were 'legacy' and 'v3.0 (default)'. Could do the same with this?

aurelienpierre:
Not to offend anyone, but I have worked an average of 10h/day on darktable for the past 2 weeks with no week end, so the name of the workflow is right now the least of my concerns.

TurboGit:
Works for me ! Thanks.

parafin:
> > so discarding is the way, and it doesn't affect old edits
> 
> If I have my workflow set to "modern", I open an old edit (using non-D65 white balance) and enable color calibration without changing anything else what happens?

Good catch, I haven't thought about this case. But in my suggestion on IRC I proposed to add a warning messages to color calibration iop too when CAT WB is enabled in it but WB isn't set to D65 in pipe. That would have taken care of this case too IMHO, but it's not implemented in this PR.

elstoc:
No we should do the same thing as for the white balance: disable the controls and make color calibration a pure channel mixer or at a very minimum set it to passthrough by default.

parafin:
Sounds good, but I don't think there is a simple way to implement it. Someone will have to look into saving workflow setting in a history stack (DB/XMP).

elstoc:
Also just disabling color calibration or compressing it away will lead to a very subtle (in many cases) change to white balance that might not be immediately obvious to a user. Disabling white balance has very obvious effects and a user will notice it. The same cannot be said of inadvertently disabling color calibration. And of course white balance will pop back on the stack if it's accidentally compressed away.

elstoc:
> Someone will have to look into saving workflow setting in a history stack (DB/XMP).

I think that's the only robust solution

parafin:
color calibration will pop back too, won't it?

elstoc:
> color calibration will pop back too, won't it?

Nope

parafin:
Hmm, why white balance is different in this regard?

As for disabling an iop - I don't understand the problem. Just because the effect is subtle, doesn't mean we should forbid disabling the module. If user disabled it, it's his right...

parafin:
I think you should describe how everything should behave instead of pointing out particular issues with current approach - it will be more productive. Because I'm not sure you can have a consistent behaviour anyhow.

elstoc:
> If user disabled it, it's his right

Well yes it's everyone's right to do something silly but this won't just be done by users who understand the consequences and are doing it on purpose. It should be sane by default and maybe there should be a "do whatever you like mode".

> I think you should describe how everything should behave instead of pointing out particular issues with current approach

I agree. We should have decided how it should work before merging it. That's the discussion my queries were trying to provoke.

elstoc:
I have suggested a solution to this when we were discussing it on IRC and I'll repeat it here

1. The chroma adaptation workflow is stored against the image and defaults to "old" for old edits. Perhaps it could be a three-way option "old, new, none".
2. It is also stored as a preference. This preference is used to set the image-workflow for new edits.
3. Both the white balance and color calibration modules are dependent on this setting
4. White balance only allows D65 for "new" workflow
5. Color calibration only allows "passthrough" for "old" workflow
6. There is a way to alter the workflow on a per-image basis (as we do for module order)
7. For "new" workflow, color calibration is treated the same as white balance and pops back onto the stack if removed

parafin:
I really don't agree with "only allows", I'd rather there would be a way to change settings, but after confirming that user knows what he is doing.

I have a question for point 6 - what happens when you change the workflow on existing stack - what changes in parameters and in history stack? It seems tricky and not obviously needed feature.

And you didn't cover if disabling color calibration iop is allowed with "new" workflow. If it is, then you still have the problem you raised about it having only a subtle easy-to-miss effect (which I don't think is a problem, but you do).

elstoc:
> I really don't agree with "only allows", I'd rather there would be a way to change settings, but after confirming that user knows what he is doing.

That's what the "none" setting is for, which should be changeable per-image.

> what happens when you change the workflow on existing stack

Depending on the setting either the white balance or color calibration modules ignore user parameters in their algorithms and instead operate based on the workflow setting.

> disabling color calibration iop is allowed with "new" workflow

Yes that's a question. I don't know the correct answer to that one. It's probably not a major problem for me and I'm sure it's not a problem for you but not every user is as involved in the details of how this all works.

Bear in mind I'm just trying to suggest a logical way round this. I don't know the internals of how a lot of this stuff works so I'll have to rely on others to decide whether it's practical.

parafin:
> > what happens when you change the workflow on existing stack
> 
> Depending on the setting either the white balance or color calibration modules ignore user parameters in their algorithms and instead operate based on the workflow setting.

So, if going from old to new:

1. wb switches to D65;
2. color calibration gets enabled, adding a new history item, and CAT WB is enabled.

And when going from new to old:

1. wb switches to "as shot" wb (or other previously chosen setting);
2. color calibration iop disables its CAT WB, but I guess isn't automatically disabled or removed from history stack?

elstoc:
Yes that sounds about right (I think)

parafin:
With old workflow there isn't a color calibration iop in history at all, so you need to add it, at least if it's not present.

elstoc:
Agreed

junkyardsparkle:
Sorry if this is a dumb question (or if already discussed elsewhere), but does colorin need to be considered for this context-switching? There are some cameras with "enhanced" matrices, which are currently applied by default, and it's not clear to me how important the use of a "standard" or otherwise known-good input matrix is for this workflow. In my experience, at least one of the Olympus "enhanced" matrices is questionable, but these obviously need to be kept for legacy reasons. Maybe it's time for a way to disable these, similarly to camera-specific base curves?

Issue #6822, opened 2020-11-09
select RGB scene blending when scene-referred workflow used


TurboGit:
**Is your feature request related to a problem? Please describe.**
select RGB scene blending when scene-referred workflow used

**Describe the solution you'd like**
select RGB scene blending when scene-referred workflow used. this must be done only for new edits.

TurboGit:
See #4688 (to be done when merged).

PR #6825, opened 2020-11-10
blend: select the default mode based on the current workflow.

TurboGit:
We have two possible workflow : display or scene referred.

When display referred is selected the default blend mode is RGB (display)
when using scene referred the default mode is RGB (scene).

This finalize the automatic workflow setting in dt.

Note that this happens only for new edits or when discarding the
history stack of images.

For #6822.

### DIFF
--- a/src/develop/blend.c
+++ b/src/develop/blend.c
@@ -49,6 +49,11 @@ static dt_develop_blend_params_t _default_blendop_params
 
 dt_develop_blend_colorspace_t dt_develop_blend_default_module_blend_colorspace(dt_iop_module_t *module)
 {
+  gchar *workflow = dt_conf_get_string("plugins/darkroom/workflow");
+  const gboolean is_scene_referred = strcmp(workflow, "scene-referred") == 0;
+  const gboolean is_display_referred = strcmp(workflow, "display-referred") == 0;
+  g_free(workflow);
+
   switch(module->blend_colorspace(module, NULL, NULL))
   {
     case iop_cs_RAW:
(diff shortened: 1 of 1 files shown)


TurboGit:
@hlecleme @aurelienpierre : please comment, approve (or not). TIA.

hlecleme:
This is probably not enough to keep the correct blending mode for old edits. The way it works for the migration (`dt_develop_blend_legacy_params` in `blend.c`) is that the blend parameters start with a fresh copy of the default parameters (which are initialized in function `dt_iop_load_default_params` or `dt_iop_load_module_by_so` in `imageop.c`). That initialization function will call the `dt_develop_blend_default_module_blend_colorspace` that could now return the new blending mode.

To make it working you'll need to reset the blend mode (and boost parameters) or better, use the `dt_develop_blend_init_blend_parameters` instead of taking the default blending configuration.

How to test: set scene referred workflow, (restart darktable), load old edit (e.g. the XMP of the test case 0034) and look at the blending mode of the exposure module.

TurboGit:
@hlecleme : any hint to where the call to dt_develop_blend_init_blend_parameters() should be done ?

hlecleme:
Implementation proposal: [patch.gz](https://github.com/darktable-org/darktable/files/5515982/default-blend-mode-from-workflow.patch.gz)

Also, JzCzhz must be scene referred, like HSL must be display referred as they are linked to what is processed by the respective blending modes.

TurboGit:
Thanks for the patch, working great, will commit.

PR #6829, opened 2020-11-10
connect histogram scroll to right exposure instance

### LINKED ISSUE #6806: Mouse wheel turn on new image causes it to pixelate
3.3.0+1440~g44fe67e70 Windows.
Opencl off.

Coppied history then realized I want to edit this fresh. Copy had multiple exposure modules. Compressed history.

Turning mouse wheel on histogram for exposure immediately pixelated.
Notice history has exposure 1, but there is no exposure 1 in modules.

This has happened on a few images now. 

image

dterrahe:
fixes #6806

removed list of exposure module proxies used by histogram to respond to mouse wheel scroll. The list-of-modules approach was orthogonal to the active multi-instance module scoring approach and got broken by #6648.

I've left the find_last_exposure_instance in place, just in case in future different modules are able to respond to exposure change requests, but at the moment it just returns the one instance of the proxy. Also to limit the size of this bug-fix patch.

### DIFF
--- a/src/develop/develop.c
+++ b/src/develop/develop.c
@@ -109,7 +109,7 @@ void dt_dev_init(dt_develop_t *dev, int32_t gui_attached)
   dev->iop_order_version = 0;
   dev->iop_order_list = NULL;
 
-  dev->proxy.exposure = NULL;
+  dev->proxy.exposure.module = NULL;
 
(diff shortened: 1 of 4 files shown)


TurboGit:
Works for me ! Thanks.

PR #6832, opened 2020-11-10
do not ellipse short notebook labels

### LINKED ISSUE #6827: Remove ellipsizing on parametric mask channels
Most parametric mask channels are one or two characters and it doesn't really make sense to ellipsize them when shrinking panels. For example:

Screenshot_2020-11-10_08-31-25

There is already an existing control that allows you to cycle through them with arrow buttons and that should be sufficient

dterrahe:
fixes #6827


TurboGit (approved):
Thanks !

PR #6834, opened 2020-11-10
map: scrolling over an images group scrolls the images of the group

phweyland:
some related changes:
- scrolling over an images group scrolls the images of the group
- fixes dnd of a group of images (only the first was moved since the grouping function)
- dnd of a group shows the next image while the first one is dragged

### DIFF
--- a/src/views/map.c
+++ b/src/views/map.c
@@ -681,8 +681,8 @@ void init(dt_view_t *self)
     }
 
     /* allow drag&drop of images from filmstrip */
-    gtk_drag_dest_set(GTK_WIDGET(lib->map), GTK_DEST_DEFAULT_ALL, target_list_internal, n_targets_internal,
-                      GDK_ACTION_COPY);
+    gtk_drag_dest_set(GTK_WIDGET(lib->map), GTK_DEST_DEFAULT_ALL,
+                      target_list_internal, n_targets_internal, GDK_ACTION_MOVE);
     g_signal_connect(GTK_WIDGET(lib->map), "scroll-event", G_CALLBACK(_view_map_scroll_event), self);
     g_signal_connect(GTK_WIDGET(lib->map), "drag-data-received", G_CALLBACK(_drag_and_drop_received), self);
     g_signal_connect(GTK_WIDGET(lib->map), "changed", G_CALLBACK(_view_map_changed_callback), self);
@@ -1476,7 +1476,8 @@ static gboolean _view_map_motion_notify_callback(GtkWidget *widget, GdkEventMoti
 
       GdkDragContext *context =
         gtk_drag_begin_with_coordinates(GTK_WIDGET(lib->map), targets,
-                                        GDK_ACTION_COPY, 1, (GdkEvent *)e, -1, -1);
+                                        GDK_ACTION_MOVE, 1,
+                                        (GdkEvent *)e, -1, -1);
 
       GdkPixbuf *cross = _draw_cross();
       gtk_drag_set_icon_pixbuf(context, cross, DT_PIXEL_APPLY_DPI(cross_size),
@@ -1559,7 +1560,8 @@ static gboolean _view_map_motion_notify_callback(GtkWidget *widget, GdkEventMoti
                            h - count_height + _thumb_border);
 
       GdkDragContext *context = gtk_drag_begin_with_coordinates(GTK_WIDGET(lib->map), targets,
-                                                                GDK_ACTION_COPY, 1, (GdkEvent *)e, -1, -1);
+                                                                GDK_ACTION_MOVE, 1,
+                                                                (GdkEvent *)e, -1, -1);
 
       gtk_drag_set_icon_pixbuf(context, thumb, 0, h + 2 * _thumb_border);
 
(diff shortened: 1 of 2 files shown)


phweyland:
Good for review.

TurboGit:
Not sure it is related to this PR, but I have set the minimum of image for a group to 5. But only after having dnd 6 images close to each others I had a group created.

>     * fixes dnd of a group of images (only the first was moved since the grouping function)

Still the case for me.

BTW, another enhancement that could be great is to show the locations (circle or square) when moving them. It is hard otherwise to place it at the proper location.

src/views/map.c
```diff
@@ -38,13 +38,22 @@
 
 DT_MODULE(1)
 
+typedef struct geo_position_t
```
TurboGit, line 41: dt_get_position_t

src/views/map.c
```diff
@@ -38,13 +38,22 @@
...
+{
+  double x, y;
+  int cluster_id;
+  int imgid;
+} geo_position_t;
```
TurboGit, line 46: likewise

TurboGit (changes requested):
Code looks ok, just a minor comment.

src/views/map.c
phweyland: Done!

phweyland:
> Not sure it is related to this PR, but I have set the minimum of image for a group to 5. But only after having dnd 6 images close to each others I had a group created.

Hummm. I've already checked this issue but I haven't found yet why. I don't give up.



> > ```
> > * fixes dnd of a group of images (only the first was moved since the grouping function)
> > ```
> 
> 
> Still the case for me.

It works well for me. Do you use shift-click ? Do you release the shift before the left click ?




> BTW, another enhancement that could be great is to show the locations (circle or square) when moving them. It is hard otherwise to place it at the proper location.

The flying pixbuf has no transparency. This seems to be a known issue of cairo pixbuf. So I've let the cross but it is really not ideal.

phweyland:
> The flying pixbuf has no transparency. This seems to be a known issue of cairo pixbuf. So I've let the cross but it is really not ideal.

If you have any idea to solve this let me know. The dnd from filmstrip to map suffers the same issue.

TurboGit:
> It works well for me. Do you use shift-click ? Do you release the shift before the left click ?

Oh I see, you need shift-click and you need to release shift before ! Not really nice :( Could all this be just a shift-click and whatever is done when releasing ?

phweyland:
> > It works well for me. Do you use shift-click ? Do you release the shift before the left click ?
> 
> Oh I see, you need shift-click and you need to release shift before ! Not really nice :( Could all this be just a shift-click and whatever is done when releasing ?

Also an old behaviour. I look at it and tell you.

phweyland:
Just remains the pixbuf transparency.

phweyland:
I've found a way to control transparency differently (I don't know if that would work) but I cannot compile it:

      GdkPixmap *pixmap;
      GdkBitmap *mask;
      gdk_pixbuf_render_pixmap_and_mask(cross, &pixmap, &mask, 1);
      gtk_drag_set_icon_pixmap(context, NULL, pixmap, mask,
                               DT_PIXEL_APPLY_DPI(cross_size),
                               DT_PIXEL_APPLY_DPI(cross_size)); 

GdkPixmap, GdkBitmap, gdk_pixbuf_render_pixmap_and_mask and gtk_drag_set_icon_pixmap are unknown ...
Any idea on this ?

parafin:
I'm not sure what exactly you are trying to achieve here, but you may want to check that your solution works on non-compositing window managers (see #5438 and #5382 for context).

phweyland:
> I'm not sure what exactly you are trying to achieve here, but you may want to check that your solution works on non-compositing window managers (see #5438 and #5382 for context).

Thanks for jumping in.
Currently the icon (pixbuf) on drag and drop doesn't respect the alpha channel. This is not an issue for plain image but it is for an empty circle or rectangle.
I'll look at your links.

phweyland:
@parafin, yes, css was the way to go. Thanks !

phweyland:
Travis is happy. So I am.

TurboGit:
@phweyland : the circle & rectangle when dragged are not adjusted for the ratio so a circle becomes an ellipse.

TurboGit:
Apart from that, the shift-click and min number of image for creating a group is working fine on my side.

phweyland:
> @phweyland : the circle & rectangle when dragged are not adjusted for the ratio so a circle becomes an ellipse.

I've also a slight difference between the original and the dragged one, but it is very small. 
Let me check.

phweyland:
I've forgotten to check the most important, i.e. the Dieppe area, sorry ... :)
Well done. Thanks.

TurboGit:
Ah ah ! Dieppe is THE place to check:) note that I was born there but I live around Paris now... Even if I would prefer not!

TurboGit:
Perfect for me now, thanks !

phweyland:
I've seen some discussions about docs. Where I should update them now ?

TurboGit:
On the same documentation as before until 3.4. After that we will migrate to the new support.

PR #6836, opened 2020-11-10
channelmixerrgb: display an error message when temperature.c does not…

aurelienpierre:
… use camera reference white balance

### DIFF
--- a/src/develop/imageop.c
+++ b/src/develop/imageop.c
@@ -197,6 +197,7 @@ void dt_iop_default_init(dt_iop_module_t *module)
   module->default_params = (dt_iop_params_t *)malloc(param_size);
 
   module->default_enabled = 0;
+  module->has_trouble = FALSE;
   module->gui_data = NULL;
 
   dt_introspection_field_t *i = module->so->get_introspection_linear();
@@ -1217,7 +1218,8 @@ gboolean dt_iop_shown_in_group(dt_iop_module_t *module, uint32_t group)
 static void _iop_panel_label(GtkWidget *lab, dt_iop_module_t *module)
 {
   gtk_widget_set_name(lab, "iop-panel-label");
-  gchar *label = dt_history_item_get_name_html(module);
+  gchar *label = g_strdup_printf("%s%s", (module->has_trouble) ? "⚠ " : "",
+                                         dt_history_item_get_name_html(module));
   gchar *tooltip = g_strdup(module->description());
   gtk_label_set_markup(GTK_LABEL(lab), label);
   gtk_label_set_ellipsize(GTK_LABEL(lab), !module->multi_name[0] ? PANGO_ELLIPSIZE_END: PANGO_ELLIPSIZE_MIDDLE);
@@ -1280,6 +1282,12 @@ void dt_iop_gui_set_enable_button(dt_iop_module_t *module)
 
 void dt_iop_gui_update_header(dt_iop_module_t *module)
 {
+  _iop_gui_update_header(module);
+}
+
+void dt_iop_set_module_in_trouble(dt_iop_module_t *module, const gboolean state)
+{
+  module->has_trouble = state;
   _iop_gui_update_header(module);
 }
 
--- a/src/develop/imageop.c
+++ b/src/develop/imageop.c
@@ -1162,8 +1162,6 @@ static void dt_iop_gui_off_callback(GtkToggleButton *togglebutton, gpointer user
 
       if(dt_conf_get_bool("darkroom/ui/activate_expand") && !module->expanded)
         dt_iop_gui_set_expanded(module, TRUE, dt_conf_get_bool("darkroom/ui/single_module"));
-
-      module->gui_update(module);
     }
     else
     {
--- a/src/develop/imageop.c
+++ b/src/develop/imageop.c
@@ -438,6 +438,7 @@ int dt_iop_load_module_by_so(dt_iop_module_t *module, dt_iop_module_so_t *so, dt
   module->header = NULL;
   module->off = NULL;
   module->hide_enable_button = 0;
+  module->has_trouble = FALSE;
   module->request_color_pick = DT_REQUEST_COLORPICK_OFF;
   module->request_histogram = DT_REQUEST_ONLY_IN_GUI;
   module->histogram_stats.bins_count = 0;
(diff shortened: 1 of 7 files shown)


elstoc:
If we're going to rely on a warning message on the CAT tab to prevent user error, I suppose we might be better off reverting #6793.

TurboGit:
> I suppose we might be better off reverting #6793.

Or revert to CAT tab when an error is reported. Maybe a big annoying, but the user has to take some action anyway.

elstoc:
It might be worth displaying a similar warning on the WB module where D65 isn't selected but CAT tab is enabled (or modern workflow is enabled). On old edits WB remains fully editable with all of its controls, regardless of the workflow setting.

aurelienpierre:
The latest commit introduces a global method fr modules to show a warning sign in header if they require attention:

aurelienpierre:
I wouldn't go as far as nagging users in white balance module too. There are legitimate reasons to not use the default coeffs in WB, for example because the Nikon input matrices tend to overdo the red, so the best option is to save a preset with a slightly reduced R coeff, compared to default D65.

elstoc:
> I wouldn't go as far as nagging users in white balance module too

I was talking about old edits, where the module still displays all of its controls because the user hasn't yet reset it.

aurelienpierre:
Old edits should not use color calibration, there is no need to reset anything. If users choose to enable color calibration, they will see the message and take appropriate measures. Old edits should be status quo.

elstoc:
Ok just trying to cover all bases here. Should the message in color calibration recommend resetting parameters in white balance instead of choosing camera reference, since this actually removes most of the controls from the module.

Note there is an issue for old edits whereby the white balance module shows all of its controls but editing those controls has no effect. The user physically cannot select camera reference. Or perhaps I need to rebuild?

aurelienpierre:
The 2 latest commits simplify things a lot.

1. temperature.c will write a boolean in `dev->proxy.wb_use_D65`, the pipeline structure, if it uses the camera reference coefficients
2. channelmixerrgb.c will write its first instance (in the pipeline order) that does chromatic adaptation in `dev->proxy.chroma_adaptation`, in the pipeline structure.

Then, warning are displayed in white balance and color calibration modules, depending on the pipeline and inner states.

1. if white balance sees a color calibration module has captured the CAT on the pipe, and its coeffs don't match D65, it will show a warning
2. if color calibration does CAT and sees `wb_use_D65 == FALSE`, it will show a warning about white balance,
3. if color calibration does CAT and sees that `dev->proxy.chroma_adaptation` is captured by another instance than itself, it will show a warning about another color calibration module.

In case several color calibration instances are competing to capture `dev->proxy.chroma_adaptation`, the one that has the lowest pipeline order wins. This should be updated as soon as the instances are inverted.

Notice there are some glitches regarding GUI updates, you need to interact with the modules for the labels to be updated. They are supposed to be updated as soon as the pipeline recomputes too, but it doesn't work.

matt-maguire:
So, had the CAT in legacy mode for a new image, temperature is doing the WB. Enable channelmixer-rgb, CAT is bypassed as expected. set CAT to something like Daylight, then get warning about conflict with temperature module.  Go down to temperature, disable it, and compress history stack. Now channelmixer-rgb complains about double CAT. There's only one instance, but the shift in position seems to have confused things. It's a bit hard to recover from this, maybe need to go back to lighttable, discard history, exit and re-enter darktable.

aurelienpierre:
Another issue with this is enabling/disabling the local contrast module switches the method (bilateral vs. local laplacian) too.

src/develop/imageop.c
```diff
@@ -1282,6 +1291,20 @@ void dt_iop_gui_update_header(dt_iop_module_t *module)
...
+  gboolean previous_state = module->has_trouble;
+
+  // we don't set disabled modules in trouble, that would be annoying
+  if(module->enabled)
+    module->has_trouble = state;
+  else
+    module->has_trouble = FALSE;
+
+  if(module->has_trouble != previous_state)
+    _iop_gui_update_header(module);
```
elstoc, line 1305: If you remove the conditionality here it resolves (mostly) the error where white balance module header isn't being updated when moving around the history stack. There still seems to potentially be an issue where you change the history while there is a pipe computation in progress so it's not 100% reliable but it's much better without this if statement.

elstoc: There's a similar error when you discard history with modern workflow, whereby the color calibration header isn't updated until you click on it, but I've not been able to track it down yet. When discarding history the gui_changed function is called once and for some reason it thinks that the current module is not the CAT module: `CAT_instance != current_instance` is `true`

elstoc:
I've made some changes that I think solve most of the issues I've found. Hopefully they're not to contentious. https://github.com/elstoc/darktable/commit/edb65e52f68e00e2c985a1218c8075588af84831

elstoc:
> Another issue with this is enabling/disabling the local contrast module switches the method (bilateral vs. local laplacian) too

What on Earth is doing that?

elstoc:
Ok really I'm gonna stop testing now. One final attempt: https://github.com/elstoc/darktable/commit/39afe426e777a0591f88d4c57d5d9889e5cda2c5.

As an aside: is this the best way to suggest changes to PRs?

src/develop/imageop.c
aurelienpierre: Updating GUI after pipe finished is supposed to be done in `_develop_ui_pipe_finished_callback()` in channelmixerrgb.c, connected to ```DT_DEBUG_CONTROL_SIGNAL_CONNECT(darktable.signals, DT_SIGNAL_DEVELOP_UI_PIPE_FINISHED,
                            G_CALLBACK(_develop_ui_pipe_finished_callback), self);```.
For some reason, that call is useless.

aurelienpierre:
Thanks for patches, I will test soon.

aurelienpierre:
The local contrast bug is still present, and now color calibration gets auto-applied with wrong settings on some older pictures, leading to full black images.

aurelienpierre:
Seems to work as expected now. @TurboGit waiting for your review.

elstoc:
Phew that took some work. Definitely pretty well user-proof now so I think it was worth it.

TurboGit:
I'll review tomorrow.

TurboGit (approved):
Works for me!

PR #6842, opened 2020-11-10
add new blending fields to history diff

dterrahe:
also correct translation marking

### DIFF
--- a/src/develop/blend.h
+++ b/src/develop/blend.h
@@ -451,6 +451,7 @@ typedef struct dt_iop_blend_name_value_t
   int value;
 } dt_develop_name_value_t;
 
+extern const dt_develop_name_value_t dt_develop_blend_colorspace_names[];
 extern const dt_develop_name_value_t dt_develop_blend_mode_names[];
 extern const dt_develop_name_value_t dt_develop_mask_mode_names[];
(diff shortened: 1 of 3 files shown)


TurboGit:
Maybe not directly related but I see that for the flip module we have an entry saying "colorspace  : RGB (scene) -> RGB (display)" I have changed the default in preference. But anyway it seems that we do not want to look at the blend introspection information for module having no blending mode. Worth fixing.

dterrahe:
For modules that _do_ have blending mode, do you want each individual module to show a colorspace change if you change the default in preferences? Otherwise, the changed default should probably be reflected in default_blendop_params. The history diff will then show any changes with regards to _your_ defaults.

TurboGit:
> For modules that _do_ have blending mode, do you want each individual module to show a colorspace change if you change the default in preferences?

Yes, I would say so.

dterrahe:
Now hiding blending diff if the module doesn't support blending.

I know it helps code simplification, but really, should we be allocating (and storing in database and xmp) a blend_params structure if the module doesn't support blending? A bunch of the standard modules store and load (and update from legacy) the default blending params over and over again without using them at all.

TurboGit:
Works for me, thanks !

PR #6845, opened 2020-11-10
Hide the output parametric mask channels by default

hlecleme:
The output parametric masks are sensitive to the output of the module, which means that when the settings of the module are updated and the output changes, it may be needed to adapt the masks again.

In order to simplify the usage of the parametric masking (and prefer the use of the input parametric masks), the output parametric mask controls will be hidden by default except if the blending of the module uses some of the output parametric mask channels.

An additional option is shown in the blending menu to show or hide the output parametric mask controls.

When hiding the output parametric mask controls the settings of those channels will be reset to avoid impacting the mask with hidden information.

@aurelienpierre I hope this suits you ;)

### DIFF
--- a/src/develop/blend.h
+++ b/src/develop/blend.h
@@ -161,7 +161,8 @@ typedef enum dt_develop_blendif_channels_t
   DEVELOP_BLENDIF_SIZE = 16,
 
   DEVELOP_BLENDIF_Lab_MASK = 0x3377,
-  DEVELOP_BLENDIF_RGB_MASK = 0x77FF
+  DEVELOP_BLENDIF_RGB_MASK = 0x77FF,
+  DEVELOP_BLENDIF_OUTPUT_MASK = 0xF0F0
 } dt_develop_blendif_channels_t;
 
 
@@ -443,6 +444,7 @@ typedef struct dt_iop_gui_blendif_filter_t
   GtkLabel *label[4];
   GtkLabel *picker_label;
   GtkWidget *polarity;
+  GtkBox *box;
 } dt_iop_gui_blendif_filter_t;
 
 typedef struct dt_iop_blend_name_value_t
@@ -509,6 +511,7 @@ typedef struct dt_iop_gui_blend_data_t
   dt_dev_pixelpipe_display_mask_t save_for_leave;
   int timeout_handle;
   GtkNotebook *channel_tabs;
+  gboolean output_channels_shown;
 
   GtkWidget *channel_boost_factor_slider;
 
@@ -632,6 +635,7 @@ void dt_iop_gui_update_blendif(dt_iop_module_t *module);
 void dt_iop_gui_update_masks(dt_iop_module_t *module);
 void dt_iop_gui_cleanup_blending(dt_iop_module_t *module);
 void dt_iop_gui_blending_lose_focus(dt_iop_module_t *module);
+void dt_iop_gui_blending_reload_defaults(dt_iop_module_t *module);
 
 gboolean blend_color_picker_apply(dt_iop_module_t *module, GtkWidget *picker, dt_dev_pixelpipe_iop_t *piece);
 
--- a/src/develop/blend_gui.c
+++ b/src/develop/blend_gui.c
@@ -378,6 +378,52 @@ static void _blendif_scale_print_default(float value, float boost_factor, char *
   snprintf(string, n, "%-5.*f", _blendif_print_digits_default(scaled), scaled * 100.0f);
 }
 
+static gboolean _blendif_are_output_channels_used(const dt_develop_blend_params_t *const blend,
+                                                  const dt_develop_blend_colorspace_t cst)
+{
+  const gboolean mask_inclusive = blend->mask_combine & DEVELOP_COMBINE_INCL;
+  const uint32_t mask = cst == DEVELOP_BLEND_CS_LAB ? DEVELOP_BLENDIF_Lab_MASK & DEVELOP_BLENDIF_OUTPUT_MASK
+                                                    : DEVELOP_BLENDIF_RGB_MASK & DEVELOP_BLENDIF_OUTPUT_MASK;
+  const uint32_t active_channels = blend->blendif & mask;
+  const uint32_t inverted_channels = (blend->blendif >> 16) ^ (mask_inclusive ? mask : 0);
+  const uint32_t cancel_channels = inverted_channels & ~blend->blendif & mask;
+  return active_channels || cancel_channels;
+}
+
+static gboolean _blendif_clean_output_channels(dt_iop_module_t *module)
+{
+  const dt_iop_gui_blend_data_t *const bd = (dt_iop_gui_blend_data_t *)module->blend_data;
+  if(!bd || !bd->blendif_support || !bd->blendif_inited) return 0;
+
+  gboolean changed = FALSE;
+  if(!bd->output_channels_shown)
+  {
+    const uint32_t mask = bd->csp == DEVELOP_BLEND_CS_LAB ? DEVELOP_BLENDIF_Lab_MASK & DEVELOP_BLENDIF_OUTPUT_MASK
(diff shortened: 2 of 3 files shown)


AxelG-DE:
I remember @aurelienpierre said, the output mask also is buffered, hence has a performance impact. Hiding that output mask will not eliminate the buffering, I suppose, hence we carry on the performance "loss" for something hidden...

I know...backward compatibility, just in case, ... but I remember, similar things got solved with filmic time ago. (when the eliminated function was used in an old edit, the option popped up, anything else, eliminated)

src/develop/blend_gui.c
```diff
@@ -378,6 +378,52 @@ static void _blendif_scale_print_default(float value, float boost_factor, char *
...
+  const unsigned int mask_inclusive = blend->mask_combine & DEVELOP_COMBINE_INCL;
+  const uint32_t mask = cst == DEVELOP_BLEND_CS_LAB ? DEVELOP_BLENDIF_Lab_MASK & DEVELOP_BLENDIF_OUTPUT_MASK
+                                                    : DEVELOP_BLENDIF_RGB_MASK & DEVELOP_BLENDIF_OUTPUT_MASK;
+  const uint32_t active_channels = blend->blendif & mask;
+  const uint32_t inverted_channels = (blend->blendif >> 16) ^ (mask_inclusive ? mask : 0);
+  const uint32_t cancel_channels = inverted_channels & ~blend->blendif & mask;
+  return active_channels || cancel_channels;
+}
+
+static int _blendif_clean_output_channels(dt_iop_module_t *module)
```
TurboGit, line 393: I know we have plenty of bad coding like this in the current code, but to me here it is better to use a gboolean to make it clear from start that this is a two state value and not some kind of status.

src/develop/blend_gui.c
```diff
@@ -378,6 +378,52 @@ static void _blendif_scale_print_default(float value, float boost_factor, char *
...
+{
+  const dt_iop_gui_blend_data_t *const bd = (dt_iop_gui_blend_data_t *)module->blend_data;
+  if(!bd || !bd->blendif_support || !bd->blendif_inited) return 0;
+
+  int changed = 0;
```
TurboGit, line 398: gboolean changed = FALSE;

src/develop/blend_gui.c
```diff
@@ -378,6 +378,52 @@ static void _blendif_scale_print_default(float value, float boost_factor, char *
...
+  if(!bd->output_channels_shown)
+  {
+    const uint32_t mask = bd->csp == DEVELOP_BLEND_CS_LAB ? DEVELOP_BLENDIF_Lab_MASK & DEVELOP_BLENDIF_OUTPUT_MASK
+                                                          : DEVELOP_BLENDIF_RGB_MASK & DEVELOP_BLENDIF_OUTPUT_MASK;
+    dt_develop_blend_params_t *const d = module->blend_params;
+    const uint32_t need_inversion = d->mask_combine & DEVELOP_COMBINE_INCL ? (mask << 16) : 0;
+    if((d->blendif & need_inversion) != need_inversion || (d->blendif & mask) != 0)
+    {
+      changed = 1;
```
TurboGit, line 407: = TRUE;

src/develop/blend_gui.c
```diff
@@ -378,6 +378,52 @@ static void _blendif_scale_print_default(float value, float boost_factor, char *
...
+      d->blendif = (d->blendif & ~(mask | (mask << 16))) | need_inversion;
+    }
+    for (size_t ch = 0; ch < DEVELOP_BLENDIF_SIZE; ch++)
+    {
+      if ((DEVELOP_BLENDIF_OUTPUT_MASK & (1 << ch)) && (d->blendif_parameters[ch * 4 + 0] != 0.0f
+          || d->blendif_parameters[ch * 4 + 1] != 0.0f || d->blendif_parameters[ch * 4 + 2] != 1.0f
+          || d->blendif_parameters[ch * 4 + 3] != 1.0f))
+      {
+        changed = 1;
```
TurboGit, line 416: = TRUE;

src/develop/blend_gui.c
```diff
@@ -1240,7 +1288,7 @@ gboolean blend_color_picker_apply(dt_iop_module_t *module, GtkWidget *picker, dt
     float picker_mean[8], picker_min[8], picker_max[8];
     float picker_values[4];
 
-    int in_out = (dt_key_modifier_state() == GDK_CONTROL_MASK) ? 1 : 0;
+    int in_out = ((dt_key_modifier_state() == GDK_CONTROL_MASK) && data->output_channels_shown) ? 1 : 0;
```
TurboGit, line 1291: I don't see all the code in this diff, but this may be also a gboolean.

src/develop/blend.h
```diff
@@ -509,6 +511,7 @@ typedef struct dt_iop_gui_blend_data_t
   dt_dev_pixelpipe_display_mask_t save_for_leave;
   int timeout_handle;
   GtkNotebook *channel_tabs;
+  int output_channels_shown;
```
TurboGit, line 514: gboolean

src/develop/blend_gui.c
```diff
@@ -2613,6 +2714,7 @@ void dt_iop_gui_init_blending(GtkWidget *iopw, dt_iop_module_t *module)
     bd->csp = DEVELOP_BLEND_CS_NONE;
     bd->blend_modes_csp = DEVELOP_BLEND_CS_NONE;
     bd->channel_tabs_csp = DEVELOP_BLEND_CS_NONE;
+    bd->output_channels_shown = 0;
```
TurboGit, line 2717: = FALSE;

src/develop/blend_gui.c
```diff
@@ -1385,9 +1433,37 @@ static void _blendif_select_colorspace(GtkMenuItem *menuitem, dt_iop_module_t *m
...
+  }
+}
+
+static void _blendif_hide_output_channels(GtkMenuItem *menuitem, dt_iop_module_t *module)
+{
+  dt_iop_gui_blend_data_t *bd = (dt_iop_gui_blend_data_t *)module->blend_data;
+  if(!bd || !bd->blendif_support || !bd->blendif_inited) return;
+  if(bd->output_channels_shown)
+  {
+    bd->output_channels_shown = 0;
```
TurboGit, line 1453: = FALSE;

src/develop/blend_gui.c
```diff
@@ -1385,9 +1433,37 @@ static void _blendif_select_colorspace(GtkMenuItem *menuitem, dt_iop_module_t *m
   }
 }
 
+static void _blendif_show_output_channels(GtkMenuItem *menuitem, dt_iop_module_t *module)
+{
+  dt_iop_gui_blend_data_t *bd = (dt_iop_gui_blend_data_t *)module->blend_data;
+  if(!bd || !bd->blendif_support || !bd->blendif_inited) return;
+  if(!bd->output_channels_shown)
+  {
+    bd->output_channels_shown = 1;
```
TurboGit, line 1442: = TRUE;

TurboGit (changes requested):
Coding style comments.

TurboGit:
This works for me and I have flagged that for 3.4 as a simplification of the bend UI is most welcomed !

src/develop/blend_gui.c
hlecleme: Done and updated `_blendif_are_output_channels_used` also.

src/develop/blend_gui.c
hlecleme: The `in_out` variable is used to distinguish between the input parametric mask widgets and output parametric mask widgets. It is also used as an index in the array (of two elements) storing those widgets.
Therefore, I prefer to keep it as an `int`.

hlecleme:
@TurboGit Code has been updated. There is a small additional change to reset the `output_channels_shown` flag when switching between pictures in darkroom (as it seems that the “default” modules are re-used).

TurboGit:
Works for me, thanks !

PR #6849, opened 2020-11-11
rework modules set and workflow

aurelienpierre:
1. rewrite the modules set as follows, for better clarity:
  * modules: all, (formerly : all)
  * modules: default, (formerly : default with new layout)
  * workflow: beginner, (formerly : minimal)
  * workflow: scene-referred, (formerly : scene-referred),
  * workflow: display-referred, (formerly : display-referred).
2. move temperature module from grading to technical, since that's what its purpose with new chromatic adaptation becomes
3. remove RGB levels and tonecurves from scene-referred workflow since their GUI implicitely expects display-referred encoding (graph x axis spans between 0 and 100 %).
4. fix a minor typo leading to bad settings (forgotten | ).


TurboGit (approved):
Thanks !

PR #6851, opened 2020-11-11
channelmixerrgb.h: typo fixed

EdgarLux:
There is only one "s" in the name (according to the source and label images).


TurboGit (approved):
Good catch ! Thanks.

PR #6852, opened 2020-11-11
Keep parametric mask settings on colorspace change

### LINKED ISSUE #6843: keep parametric mask setting when changing new blending mode

Keep parametric mask setting when changing new blending mode

When editing a picture I had started setting a parametric mask. I wanted to see if I could do better or at least the difference when using the "RGB (display)" blending mode, back to "RGB (scene)" mode I discovered that my settings were lost.

It would be really nice to keep the setting of widgets in all modes even if only the current one is used.

hlecleme:
When the color space of the blending module is changed, the parametric mask settings would be lost because they need to be reinitialized.

Improve the situation by looking up in the history the last edit of the blending parametric masks with the selected color space and copying them instead of starting with a fresh setting.

When changing the color space of the blending module, a new history item is now forced on top of the history.

Fixes #6843

### DIFF
--- a/src/develop/blend_gui.c
+++ b/src/develop/blend_gui.c
@@ -1357,9 +1357,29 @@ static gboolean _blendif_change_blend_colorspace(dt_iop_module_t *module, dt_dev
   {
     dt_develop_blend_init_blendif_parameters(module->blend_params, cst);
 
+    // look for last history item for this module with the selected blending mode to copy parametric mask settings
+    GList *history = g_list_last(darktable.develop->history);
+    while(history)
+    {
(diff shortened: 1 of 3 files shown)


hlecleme:
@TurboGit If permitted to use the history in this way, I found this option the most predictable and not too heavy to implement. If I have time today, I can try to implement option 2.

TurboGit:
@hlecleme : No problem, this solution is OK by me too, a bit more complex to implement than my proposal but you've done the work, so really no problem especially since it is better this way. Will test.

TurboGit:
Works for me!

PR #6860, opened 2020-11-11
Improve display of parametric mask channels

hlecleme:
The display of the individual parametric mask channels has currently several small issues. To improve the situation this pull request contains the following changes:

- Include the boost factor in the computation of the display value for the channel, e.g. to see if the channel is clipped. This is something that was overlooked in #4688 

- Display the lab parametric channel masks correctly. The values should be converted from RGB to Lab using the same profile as will be used by the pipeline to convert the output of the module (in Lab) to RGB (used for gamma). Before #4688 the Lab masks were not working at all, and it seems that the fix in #4688 is not correct.

- Try to have more similar colors between the three blending implementations where it is meaningful (luminance, chroma and hue).

### DIFF
--- a/src/develop/blends/blendif_lab.c
+++ b/src/develop/blends/blendif_lab.c
@@ -1367,10 +1367,9 @@ static _blend_row_func *_choose_blend_func(const unsigned int blend_mode)
 #endif
 static inline void _display_channel_value(float *const restrict out, const float value, const float mask)
 {
-  // We are in the lab color space, write only the luminance
-  out[0] = value * 100.0f;
-  out[1] = 0.0f;
-  out[2] = 0.0f;
+  out[0] = value;
+  out[1] = value;
+  out[2] = value;
   out[3] = mask;
 }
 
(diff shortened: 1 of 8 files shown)


hlecleme:
@aurelienpierre Could you review this and see if some improvements can be done (although I'm not sure 100% compliance is needed for something that is only there to give some visual feedback)?

The computation of the displayed values is probably not fully correct, as I'm not exactly sure what kind of values the gamma modules takes (as a lot of modules can be skipped when the masks are displayed → perhaps any RGB kind that could be present in the pipe?) and what kind of values it produces (sRGB?).

@TurboGit Except if there are big "color science" mistakes in it, I think it would be worth to include it in 3.4 ;)

Thanks in advance!

hlecleme:
With the latest change, tried to improve the coherence between the colored (hue) sliders of the parametric masking and the display of the individual channels. This makes also the hues in the different blending color spaces have a more similar look.

src/develop/blends/blendif_lab.c
```diff
@@ -122,7 +122,7 @@ static inline void _blendif_lab_a(const float *const restrict pixels, float *con
 {
   for(size_t x = 0, j = 0; x < stride; x++, j += DT_BLENDIF_LAB_CH)
   {
-    mask[x] *= _blendif_compute_factor(pixels[j + 1] / 128.0f, invert_mask, parameters);
+    mask[x] *= _blendif_compute_factor(pixels[j + 1] / 256.0f, invert_mask, parameters);
```
aurelienpierre, line 125: shouldn't that be 255 ?

src/develop/blends/blendif_lab.c
```diff
@@ -1362,93 +1362,128 @@ static _blend_row_func *_choose_blend_func(const unsigned int blend_mode)
 }
 
 
+#ifdef _OPENMP
+#pragma omp declare simd aligned(sRGB, XYZ: 16)
+#endif
+static inline void _linear_sRGB_to_XYZ(const float *const restrict sRGB, float *const restrict XYZ)
```
aurelienpierre, line 1368: linear sRGB is Rec709, there is a function for that in `colorspaces_inline_conversion.h`, unless the white point does not match.

aurelienpierre (approved):
The code makes sense, I need to test. Just 2 minor concerns to address.

src/develop/blends/blendif_lab.c
hlecleme: Not sure, I guess it depends on the conventions. Looking at the code of darktable 3.2.1 we have:
```
      scaled[DEVELOP_BLENDIF_A_in]
          =clamp_range_f((input[1]+128.0f)/256.0f, 0.0f, 1.0f); // a scaled to 0..1
      scaled[DEVELOP_BLENDIF_B_in]
          =clamp_range_f((input[2]+128.0f)/256.0f, 0.0f, 1.0f);                 // b scaled to 0..1
```
It looks like darktable is using the convention of a and b values between -128.0f and 128.0f.

src/develop/blends/blendif_lab.c
hlecleme: Thanks :) Yes the white point matches the ones of the function `dt_XYZ_to_Rec709_D50`, used that one instead.

aurelienpierre:
Ok everything seems to behave as expected, I didn't kind anything suspicious. Thanks !

PR #6862, opened 2020-11-11
fix buffer overflow

rabauke:
darktable crashes due to a buffer overflow when "in cm (for print)" is selected in the export dialog. This pull request fixes this by using `snprintf` and a larger buffer. The buffer size is chosen generously in order not to limit translators unnecessarily. Also revised the wording and enabled translation.


TurboGit (approved):
Thanks ! Wording is also better.

PR #6869, opened 2020-11-12
map: fix drag and drop from and to filmstrip

### LINKED ISSUE #6866: map: d&d on the map does not work
**Describe the bug**
drag and drop from filmstrip to map to geolocalize on image does not work anymore

Probably due to recent work on map view.

**Expected behavior**
Image should be placed on the map and geolocalised.

 - Darktable Version: master
 - OS: GNU/Linux

phweyland:
Fixes #6866 
I've just replaced the thumbtable and thumbnail GDK_ACTION_COPY by GDK_ACTION_MOVE as in map.
I haven't touched the darkroom ones related to modules.
I've checked that move of images works in lighttable (including for tags collections).
For the record I've made this change in map to avoid to have to release SHIFT before left button.
But I think also that GDK_ACTION_MOVE is more appropriate for those actions than GDK_ACTION_COPY.
So should be changed also in darkroom ...

### DIFF
--- a/src/dtgtk/thumbnail.c
+++ b/src/dtgtk/thumbnail.c
@@ -1011,7 +1011,7 @@ GtkWidget *dt_thumbnail_create_widget(dt_thumbnail_t *thumb)
   {
     // this is only here to ensure that mouse-over value is updated correctly
     // all dragging actions take place inside thumbatble.c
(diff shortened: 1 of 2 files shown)


TurboGit:
I have also checked that drag & drop from OS in lighttable properly import the corresponding picture. All seems ok on my side. Thanks.

I don't know for darkrroom, nothing seems broken there.

PR #6871, opened 2020-11-12
map: factor code draw image & location

phweyland:
Use the same code both for still images & locations and for dragged images & locations.
The dragged image got the pin thanks to the new transparency.

### DIFF
--- a/data/themes/darktable.css
+++ b/data/themes/darktable.css
@@ -1546,7 +1546,7 @@ border-bottom: 1px solid @button_border;
   color: @field_hover_fg;
 }
 
-#map_location_drag_icon
+#map_drag_icon
 {
   background: transparent;
 }
--- a/src/views/map.c
+++ b/src/views/map.c
@@ -65,13 +65,13 @@ typedef struct dt_map_t
   float lat0, lat1, lon0, lon1;
   int time_out;
   int timeout_event_source;
+  GtkWidget *drag_icon;
   struct
   {
     dt_location_draw_t main;
     gboolean drag;
     int time_out;
     GList *others;
-    GtkWidget *drag_icon;
   } loc;
 } dt_map_t;
 
@@ -605,7 +605,7 @@ void init(dt_view_t *self)
     lib->thumb_lat_angle = 0.01, lib->thumb_lon_angle = 0.01;
     lib->time_out = 0, lib->timeout_event_source = 0;
     lib->loc.main.id = 0, lib->loc.main.location = NULL, lib->loc.time_out = 0;
-    lib->loc.others = NULL, lib->loc.drag_icon = NULL;
+    lib->loc.others = NULL, lib->drag_icon = NULL;
 
     OsmGpsMapSource_t map_source
         = OSM_GPS_MAP_SOURCE_OPENSTREETMAP; // open street map should be a nice default ...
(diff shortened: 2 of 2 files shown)


phweyland:
Going this way, I'll try to integrate also the drag image from filmstrip.

phweyland:
The doc (gtk_drag_set_icon_widget ()) says that the widget is not destroyed by gtk, but each time it has not destroyed, next dnd, when gtk_widget_destroy() is called the widget is said as invalid. From this I deduce that, in some way, it has been destroyed.
Is there any way to track such a potential memory leak ?

phweyland:
> From this I deduce that, in some way, it has been destroyed.

When the corresponding pixbuf is unref / deleted ?

phweyland:
I've an issue with travis:
Map mode: disabled, please install libosmgpsmap[ line 263](https://github.com/darktable-org/darktable/pull/6871/checks?check_run_id=1391674107#step:4:263)

Do I have to set the condition `#ifdef HAVE_MAP` in thumbtable code or is it an issue of travis settings ?

phweyland:
> Do I have to set the condition `#ifdef HAVE_MAP` in thumbtable code or is it an issue of travis settings ?

Thinking a bit more, I have to do it obviously. Sorry for the noise.

Nilvus:
> Going this way, I'll try to integrate also the drag image from filmstrip.

That would be good

phweyland:
Dnd from filsmstrip. Apart a doubt above about the widget_destroy, this ok for me.

src/dtgtk/thumbtable.c
```diff
@@ -1481,38 +1481,53 @@ static void _event_dnd_begin(GtkWidget *widget, GdkDragContext *context, gpointe
...
-
-    if(buf.buf)
+    if(table->drag_list)
+      dt_view_map_drag_set_icon(darktable.view_manager, context,
+                                GPOINTER_TO_INT(table->drag_list->data),
+                                g_list_length(table->drag_list));
+  }
+  else
+  {
+#endif
```
TurboGit, line 1496: I suppose this can be moved above...

src/dtgtk/thumbtable.c
```diff
@@ -1481,38 +1481,53 @@ static void _event_dnd_begin(GtkWidget *widget, GdkDragContext *context, gpointe
...
 
-    dt_mipmap_cache_release(darktable.mipmap_cache, &buf);
-  }
+        if(source) g_object_unref(source);
+        if(scaled) g_object_unref(scaled);
+      }
 
+      dt_mipmap_cache_release(darktable.mipmap_cache, &buf);
+    }
+#ifdef HAVE_MAP
```
TurboGit, line 1528: and this one removed.

TurboGit (changes requested):
Quick review, I suppose this is for post 3.4 ?

src/dtgtk/thumbtable.c
phweyland: Sorry, I don't get you. Could you tell me a bit more ?
If HAVE_MAP is false, the code must be as before. But if TRUE, the `if{ new code }else{ old code }` must be complete.

TurboGit: In C a simple  block like this is correct:
```
{
   ...
}
```

So I propose to use a single HAVE_MAP:

```
#if HAVE_MAP
if (...)
{
   ...
}
else
#endif
{
   ...
}
```

johnny-bit:
> From this I deduce that, in some way, it has been destroyed.
> Is there any way to track such a potential memory leak ?

GTK is weird in that way. if I understand correctly, you shouldn't call destroy on attached widget, but you should first detach it from container and/or unref it. it's weirdo manual ref counting mem management...

src/dtgtk/thumbtable.c
phweyland: so obvious ! thanks for your patience !

phweyland:
> GTK is weird in that way. if I understand correctly, you shouldn't call destroy on attached widget, but you should first detach it from container and/or unref it. it's weirdo manual ref counting mem management...

I've made also a try with unref(widget) (before or after the unref(pixpuf)) but I always get an error...(unref => not exist)

src/dtgtk/thumbtable.c
phweyland: done

phweyland:
> Quick review, I suppose this is for post 3.4 ?

Yes, if you find it ok (and if you have the time to check it :)).

TurboGit (approved):
Ok, a clean cannot hurt... well it can but this seems good to me. Thanks.

PR #6874, opened 2020-11-12
Spanish translation update

EdgarLux:
Adding more translated strings.

### DIFF
--- a/po/es.po
+++ b/po/es.po
@@ -2,8 +2,8 @@ msgid ""
 msgstr ""
 "Project-Id-Version: darktable 3.4\n"
 "Report-Msgid-Bugs-To: \n"
-"POT-Creation-Date: 2020-11-10 21:13-0600\n"
-"PO-Revision-Date: 2020-11-10 20:55-0600\n"
+"POT-Creation-Date: 2020-11-11 14:01-0600\n"
+"PO-Revision-Date: 2020-11-11 23:40-0600\n"
 "Last-Translator: Edgar Lux <edgarlux@disroot.org>\n"
 "Language-Team: Edgar Lux\n"
 "Language: es_ES\n"
@@ -49,7 +49,7 @@ msgstr "cortar"
 #: ../build/lib/darktable/plugins/introspection_ashift.c:138
 #: ../build/lib/darktable/plugins/introspection_ashift.c:257
 #: ../src/common/collection.c:688 ../src/gui/preferences.c:837
-#: ../src/gui/preferences.c:1916 ../src/gui/presets.c:543
+#: ../src/gui/preferences.c:1916 ../src/gui/presets.c:525
 #: ../src/libs/camera.c:581 ../src/libs/metadata_view.c:124
 msgid "focal length"
 msgstr "longitud focal"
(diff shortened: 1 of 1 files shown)


rgo:
Sorry, it's impossible to comment directly such a big diff file from my tablet.

Line 311: gamut could be translated as "espacio de color" or "espectro" (I prefer first translation). When I looked for monitors I don't remember to read about "gama" referring to gamut.

L2217: I'd translate it as "Mostrar grupos de módulos y/o el buscador" 
By the way,  module groups term is used below, at line 2237, and it's translated differently. 

One doubt, should you capitalize first letter when the literal is starting a sentence or is done by the presentation layer? 

Tomorrow, I'll review the whole diff from my computer and I'll add comments in line. 

Good job @EdgarLux 👌🏻

rgo:
It's late has been merged! 🤦🏻‍♂️

EdgarLux:
Thanks, Rafa. I really appreciate you taking the time to make your comments. I would like to describe some considerations to you: 

> Line 311: gamut could be translated as "espacio de color" or "espectro" (I prefer first translation). When I looked for monitors I don't remember to read about "gama" referring to gamut.
> 
As you have noticed, it is a word that translated into Spanish can have more than one meaning. And in this case, although their relationship is implicit, they are not necessarily the same. The context must be taken into account.

> L2217: I'd translate it as "Mostrar grupos de módulos y/o el buscador"
> By the way, module groups term is used below, at line 2237, and it's translated differently.
> 
It is precisely this that serves as an example, not only of taking into account the context, but also of taking into account other considerations and going a little deeper. For example, [here.](http://udep.edu.pe/castellanoactual/duda-resuelta-uso-del-yo/)

> One doubt, should you capitalize first letter when the literal is starting a sentence or is done by the presentation layer?
> 
Do you use the Spanish UI? Perhaps it would be clearer to see it there than to describe it here in words.

Several translations require more comprehensive documentation and take into account certain considerations, not only of the program but also of the language. This goes beyond taking a few words from one language to another in a few minutes. On the contrary, it can take even longer. And for a review you would expect no less.

> Good job @EdgarLux 👌🏻

Thank you @rgo! It's my pleasure.

rgo:
@EdgarLux 
> > Line 311: gamut could be translated as "espacio de color" or "espectro" (I prefer first translation). When I looked for monitors I don't remember to read about "gama" referring to gamut.
> 
> As you have noticed, it is a word that translated into Spanish can have more than one meaning. And in this case, although their relationship is implicit, they are not necessarily the same. The context must be taken into account.

Maybe I didn't explain correctly. Having it translated as "gama" doesn't seems correct at all (despite being one of the meanings in the dictionary), in photography domain I'd say is used the terms "espacio de color" or "espectro" (I don't know which one is more preferably).

> > L2217: I'd translate it as "Mostrar grupos de módulos y/o el buscador"
> > By the way, module groups term is used below, at line 2237, and it's translated differently.
> 
Understood the point about "y/o", new thing learnt.

Please, take into account the other comment about module groups is being translated in different ways, to standardize the term.

> 
> > One doubt, should you capitalize first letter when the literal is starting a sentence or is done by the presentation layer?
> 
> Do you use the Spanish UI? Perhaps it would be clearer to see it there than to describe it here in words.

I use it from time to time but always come back to the English one (sounds strange because I'm used to the English terms).

In fact, my question was more related about how is it implemented. I'm interested in know if the presentation layer is in charge of capitalising and formatting the terms or the translations should be formatted. 

Regards

PR #6875, opened 2020-11-12
Add shortcuts for blending and group shapes

dterrahe:
Creates a separate "pseudo" module called "blending" in the shortcuts list. Shortcuts defined here will be applicable to the currently  active module (subject to multi-instance prioritisation).

There are sub-groups for tools and shapes (and these have also been introduced in retouch/liquify/spots).

Edit 2: removed shortcuts for colorpickers as those do not get grouped correctly when translated.

Edit: Tone equaliser shortcuts are called "dark highlights", "light shadows", "speculars" etc so it is hard to for example assign keys 1/2/3 etc in the correct order. Now a new section "simple" is created with duplicates of the slider shortcuts called +0 EV, -1 EV, -2 EV etc.

### DIFF
--- a/src/bauhaus/bauhaus.c
+++ b/src/bauhaus/bauhaus.c
@@ -873,7 +873,7 @@ void dt_bauhaus_widget_set_label(GtkWidget *widget, const char *section_orig, co
 
   if(w->module)
   {
-    if(!darktable.bauhaus->skip_accel && (!section_orig || strcmp("blend", section_orig)))
+    if(!darktable.bauhaus->skip_accel)
     {
       gchar *combined_label = section_orig
                             ? g_strdup_printf("%s`%s", section_orig, label_orig)
--- a/src/develop/blend_gui.c
+++ b/src/develop/blend_gui.c
@@ -2760,6 +2760,8 @@ void dt_iop_gui_init_blending(GtkWidget *iopw, dt_iop_module_t *module)
   /* create and add blend mode if module supports it */
   if(module->flags() & IOP_FLAGS_SUPPORTS_BLENDING)
   {
+    --darktable.bauhaus->skip_accel;
+
     module->blend_data = g_malloc0(sizeof(dt_iop_gui_blend_data_t));
     dt_iop_gui_blend_data_t *bd = (dt_iop_gui_blend_data_t *)module->blend_data;
 
@@ -2986,6 +2988,8 @@ void dt_iop_gui_init_blending(GtkWidget *iopw, dt_iop_module_t *module)
     gtk_widget_set_name(GTK_WIDGET(iopw), "blending-wrapper");
 
     bd->blend_inited = 1;
+
+    ++darktable.bauhaus->skip_accel;
   }
 }
 
--- a/src/develop/blend_gui.c
+++ b/src/develop/blend_gui.c
@@ -504,7 +504,7 @@ static void _blendop_masks_mode_callback(const unsigned int mask_mode, dt_iop_gu
     if(data->module->blend_colorspace(data->module, NULL, NULL) == iop_cs_RAW)
     {
       data->module->request_mask_display = DT_DEV_PIXELPIPE_DISPLAY_NONE;
-      dtgtk_button_set_active(DTGTK_BUTTON(data->showmask), FALSE);
+      gtk_toggle_button_set_active(GTK_TOGGLE_BUTTON(data->showmask), FALSE);
       gtk_widget_hide(GTK_WIDGET(data->showmask));
 
       // disable also guided-filters on RAW based color space
@@ -527,7 +527,7 @@ static void _blendop_masks_mode_callback(const unsigned int mask_mode, dt_iop_gu
   else
   {
     data->module->request_mask_display = DT_DEV_PIXELPIPE_DISPLAY_NONE;
-    dtgtk_button_set_active(DTGTK_BUTTON(data->showmask), FALSE);
+    gtk_toggle_button_set_active(GTK_TOGGLE_BUTTON(data->showmask), FALSE);
     data->module->suppress_mask = 0;
     gtk_toggle_button_set_active(GTK_TOGGLE_BUTTON(data->suppress), FALSE);
 
@@ -1046,9 +1046,9 @@ static void _blendop_blendif_showmask_clicked(GtkWidget *button, GdkEventButton 
       module->request_mask_display |= (has_mask_display ? 0 : DT_DEV_PIXELPIPE_DISPLAY_MASK);
 
     if(module->request_mask_display & (DT_DEV_PIXELPIPE_DISPLAY_MASK | DT_DEV_PIXELPIPE_DISPLAY_CHANNEL))
-      dtgtk_button_set_active(DTGTK_BUTTON(button), TRUE);
+      gtk_toggle_button_set_active(GTK_TOGGLE_BUTTON(button), TRUE);
     else
-      dtgtk_button_set_active(DTGTK_BUTTON(button), FALSE);
+      gtk_toggle_button_set_active(GTK_TOGGLE_BUTTON(button), FALSE);
 
 
     if(module->off) gtk_toggle_button_set_active(GTK_TOGGLE_BUTTON(module->off), TRUE);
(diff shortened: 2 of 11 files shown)


dterrahe:
> Color pickers in the "quad" area of sliders/comboboxes can be assigned a shortcut as well

Hm; in French, the "pipette de couleur" is not grouped together with the other slider shortcuts, but an equally named group is created next to it (in preferences/raccoucis). This is because I'm (re)using the bauhaus label field, but that has already been translated. So the shortcut also ends up partially translated in keyboardrc. I'll need to think about a solution for this.

Maybe this finally needs the restructuring of darktable.control->accelerator_list so that it stores a widget pointer and a list of actions and then the colorpicker action can be added to that directly (after looking up the widget) rather than creating a sort-of stand-alone shortcut that happens to end up in the same group. Accelerator_lists should also store the section (which now is forgotten after the call from set_label) so colorpickers attached to sectioned widgets (color balance, calibration, split-toning) also end up in the correct group.

One more use case (together with #3648, vimkeys etc) that feeds into the design decision for accelerator_list_ng.

johnny-bit:
I like the idea! (Will test later tho...)

dterrahe:
No hurry; 3.6 is soon enough for me.

dterrahe:
Hi @johnny-bit have you had time to have a quick look at this? Thoughts?

johnny-bit:
I've think I've tested just a bit and it seemed to work nicely for me. I think it's a bit complex to wrap around your head in prefereces settings.

dterrahe:
Agreed that the whole multi-instance shortcut preferences thing is complex, by its nature. I don't think it can be made simple.

The blending shortcuts just follow the normal multi-instance shortcuts. So the only thing "unintuitive"/"unusual" about them, is that they always apply to an instance of the currently active module, whereas module shortcuts can also work on modules where none of the instances are currently active. To do this for blending, somehow they would have to be module specific too (adding the full set of blending shortcuts to each and every module). I don't think that's useful, especially with the limited size keyboards that most people have ;-)

In the end I think the solution is simple; just get a really gifted writer to explain it in the documentation!

(I think he's following this thread, so we are good!)

More seriously; if somebody _does_ have any ideas on how to make it simple, I'm all ears. If not, well, let's give it to the people who can handle it and not force anybody to use it who doesn't want to.

elstoc:
> Shortcuts defined here will be applicable to the currently active module (subject to multi-instance prioritisation).

Could you explain this further please (disclaimer: I haven't played with this yet). Does this add a blending shortcut section to every module? I would have thought this would be easier if it just applied to the module with gui focus (I think most of the blend shape display and the like does this anyway). So just a single global shortcut that, once you've given a module instance focus will just act on that instance.

elstoc:
hmm I think this would be better if it just worked on the current gui-focused module

johnny-bit:
I played with it a bit. Seems fun! I mostly agree with @elstoc - make order of preffered modules always favour currently expanded module/active module always... Multi-instances are mind-boggers. I *think* the intention might be to quickly enable masking for module, but so far I just used it to have shortcut to enable drawn + paramteric and then single shapes...

elstoc:
It's just the idea of "find out what the current gui instance is and then select a different one" seems a bit odd.

dterrahe:
> hmm I think this would be better if it just worked on the current gui-focused module. It's different with shortcuts that target a specific iop module because they don't care about focus at all.

Makes sense I guess. I'll look if changing it is doable. Challenge is; this now uses the same infrastructure as the rest of the per-module _widgets_ (the shortcuts aren't per-module, but the widgets are). So I'd need to make a distinction between them to be able to handle them differently on focus changes.

I still kind of think that a mode where they are in sync with the module affected by module shortcuts is useful too. Because when I change it to follow-focus, I'm sure I'll get the response "I change the exposure and then add a mask and it ends up in a different module". Configuration option, anyone? * ***hides*** *

dterrahe:
How about approaching it from the other side; what if for _normal_ module shortcuts, the focused instance always gets priority (if there is one) and otherwise the current preference scheme applies. That might also be an improvement for those normal shortcuts, would make the most sense for blending _and_ they are now always in sync.

johnny-bit:
IMO souns good!

elstoc:
Surely the two will never be in sync because they work in different ways. In your proposal, module-based shortcuts (e.g. exposure/exposure" will go "is the current gui module an instance of exposure? if not go search for exposure instances using the rules". The blending shortcuts simply won't work if there's not a module with gui focus, because the current preferences assume you start with knowing what iop you want to interact with.

dterrahe:
The way multi-instance short cuts work is not by looking up which instance to apply them to after they have been pressed, but by setting up the shortcut links in advance to an instance after each change in module state (enable/disable/expand/collapse). Blending shortcuts are treated in the same way, which means they'll point to the last module that was interacted with and the instance of that module as determined by the prefs.

Focus doesn't normally feature in this, _except_ for _local_ shortcuts, which only get enabled when the instance has focus. It turns out the blending _shapes_ were marked as local and therefore get out of sync with the other blending shortcuts. Which may have been part of the reason it got extra confusing for @johnny-bit. I have now removed the local flag (which I'm not fond of anyway; I wonder if it is ever really needed. It definitely is often confusing).

This PR has quite a bit refactoring in it too and grouping of the (non-blending) tools buttons in other modules. I'm wondering if, rather than let it bit-rot, we could merge it first and then get more feedback on preference treatment of blending shortcuts and fix them later if needed. It is probably easier to talk about when it can actually be played with.

Unless there is a consensus that blending shortcuts are not something we'll ever want.

johnny-bit:
I'd vote for merge. I *like* blending shortcuts. I dream of the day of midi consoles & graphic tablets in dt where wiring it up will be dream come true...

dterrahe:
I'm working towards it, step by step...

TurboGit:
Looks like a nice addition indeed! A comment, not sure what should be done:
- I have added a shortcut for adding a circle
- I have selected a module and triggered the shortcut
- I was able to add a circle

So all fine.

But I was not using a blending mode. Maybe we can do something about this, but I'm not sure it is easy or even desirable (side effects?). My idea was:

- if there is no blending mode selected or current mode is uniform 
- when creating a form
- select the drawn mask mode

- if current mode is parametric 
- when creating a form
- select the drawn mask + parametric mode

Again, that's just an idea but not sure it is worth it, let me know.

dterrahe:
For consistency, it might be best to do the same here as was introduced recently for shortcuts for bauhaus widgets; ignore them if the widget is hidden.

I had considered your ideas about automatically switching on a blending mode (but then forgot that there was an issue here that needed to be solved or maybe only concluded that for sliders there isn't a problem anymore) but I think the drawback is that toggling the "add circle" mode back off doesn't get you to the previous state. And if the module isn't expanded, that might not be obvious. You have to _know_ that you have to go in and change the blending mode back.

So I think disabling the hidden shortcut would be safer. Do you agree?

TurboGit:
> So I think disabling the hidden shortcut would be safer. Do you agree?

Yes. That would be a good first step and we can revisit if we can do better later.

dterrahe:
> Yes. That would be a good first step and we can revisit if we can do better later.

There are quite a few other places where activating a hidden widget could trigger a mode change that would make the widget visible. So if we go down that path, we should have a clear idea what the user expectation should be in the end.

johnny-bit:
Yay! Use cases, user stories and scenarios! any time and we'll have SCRUM master :P

Anyway - I *think* that activation/ignoring of shortcut should follow ui intent like so:
1. module has no blending
2. user tries to add circle -> ignored
3. user uses shortcut to activate drawn+parametric blending
4. uses shortcut to add circle -> working

dterrahe:
Agreed

> 4\. uses shortcut to add circle -> working

Yes, otherwise the shortcut would _never_ work? Would mean quite a bit of work for nothing! ;-)

johnny-bit:
interesting problem... For starters I always use drawn+parametric and never just drawn. some other people use drawn... otherwise activating addcricle could activate drawn/drawn+parametric blending mode... dunno...

dterrahe:
Two additions:
- deactivate shortcuts to blending settings that are currently not active. Same approach as with dynamic shortcuts to disabled (by hiding) widgets, except since there is one more layer of box within box in the blending settings, the widget's parent's parent needs to be checked for visibility too.
- make multi-instance shortcuts always point to the focused module, if it is of the same module type. This means blending and other "module" shortcuts will be in sync. Shortcuts are reconnected both when acquiring and when losing focus, because in the latter case the standard shortcut preference rules need to be applied again.

This uncovered a possibly longstanding bug with module focusing that needs more investigation. Liquify seems to not want to let go of focus once it has it. I was totally unable to focus on any other module until a restart, which obviously interfered with this patch, but more generally breaks focusing behavior, which may be why it is perceived as unreliable/usable. I removed the offending call to dt_iop_request_focus in liquify, but this needs a better solution because now it is possible to get into "add shape" mode without liquify having focus, which means the mask isn't being shown.

@TurboGit I think that should be treated as a separate bug, rather than trying to fix it here and creating a hard dependency. I think the focus stealing is worse than the invisible mask while editing, so my preference would be to merge this including the removal of that dt_iop_request_focus and then file a bug that liquify mask editing needs a correct fix.

johnny-bit:
This look and feels awesome!

dterrahe:
@TurboGit any further reservations about this patch? Otherwise I think it is ready (and if I interpret @johnny-bit 's comment correctly, he does so too).

TurboGit:
@dterrahe : No, just need to test it again a bit before merging. On my TODO list for next week or maybe later tonight.

TurboGit (approved):
Ok, works for me. Thanks for this nice addiction!

dterrahe:
Thanks!

And you are right, this whole shortcut thing does feel a bit like an addiction... ;-)

PR #6881, opened 2020-11-13
deprecate rawdenoise.c

aurelienpierre:
this module is vastly superceded by the denoiseprofile one, which has the same wavelets decomposition method with much better settings. I see no reason to keep it, it confuses users.

### DIFF
--- a/src/iop/rawdenoise.c
+++ b/src/iop/rawdenoise.c
@@ -123,7 +123,7 @@ const char *name()
 
 int flags()
 {
-  return IOP_FLAGS_SUPPORTS_BLENDING;
+  return IOP_FLAGS_SUPPORTS_BLENDING | IOP_FLAGS_DEPRECATED;
 }
 
 int default_group()


TurboGit:
Let's vote:

<p><a href="https://api.gh-polls.com/poll/01EQ06HWYZJ98N7QYZQHJ19B17/never%20used%20it%2C%20deprecate%20it/vote"></a>
<a href="https://api.gh-polls.com/poll/01EQ06HWYZJ98N7QYZQHJ19B17/almost%20nerver%20used%20it%2C%20don&#39;t%20care%20baout%20the%20final%20choice/vote"></a>
<a href="https://api.gh-polls.com/poll/01EQ06HWYZJ98N7QYZQHJ19B17/this%20is%20an%20important%20module%20for%20me%2C%20keep%20it%20active/vote"></a></p>

elstoc:
does the fact that it works before demosaic not make a difference?

rawfiner:
Well actually the wavelets are not the same as in denoise profiled. Raw denoise has some pros in my opinion: it is very fast, and before demosaic

rawfiner:
But I almost never use it in practice, I find denoise profiled results superior

TurboGit:
Right, this is a unique module where it is placed in the pipe. I don't think removing it makes sense. Not having it in default module group presets is ok, but at least that's a tool we can still use.

aurelienpierre:
The main problem of that module is at blurs picture a lot, compared to denoise profile.

aurelienpierre:
18 for, 14 against, 14 don't care. Motion denied.

PR #6884, opened 2020-11-13
rename "denoise (non-local means)" to "astrophoto denoise"

aurelienpierre:
The name of the denoising method is not informative for most users and is a duplicate of the non-local means in denoiseprofile.c. As a result, users get prompted with 4 different modules to denoise and are lost.

This module denoise in Lab, which has a cubic root built-in, which makes it better suited to remove Poisson noise that is typical in astrophotography. Hence the rename, which should make it clearer.

### DIFF
--- a/src/iop/nlmeans.c
+++ b/src/iop/nlmeans.c
@@ -85,7 +85,7 @@ typedef struct dt_iop_nlmeans_global_data_t
 
 const char *name()
 {
-  return _("denoise (non-local means)");
+  return _("astrophoto denoise");
 }
 
 int default_colorspace(dt_iop_module_t *self, dt_dev_pixelpipe_t *pipe, dt_dev_pixelpipe_iop_t *piece)
--- a/src/iop/nlmeans.c
+++ b/src/iop/nlmeans.c
@@ -88,6 +88,11 @@ const char *name()
   return _("astrophoto denoise");
 }
 
+const char *aliases()
+{
+  return _("denoise (non-local means)");
+}
+
 int default_colorspace(dt_iop_module_t *self, dt_dev_pixelpipe_t *pipe, dt_dev_pixelpipe_iop_t *piece)
 {
   return iop_cs_Lab;


TurboGit:
Let's vote:

<p><a href="https://api.gh-polls.com/poll/01EQ0711VHKGF698NP677CH50X/new%20name%20seems%20clearer%2C%20please%20rename/vote"></a>
<a href="https://api.gh-polls.com/poll/01EQ0711VHKGF698NP677CH50X/don&#39;t%20care%20about%20the%20name%2C%20don&#39;t%20care%20about%20the%20final%20choice/vote"></a>
<a href="https://api.gh-polls.com/poll/01EQ0711VHKGF698NP677CH50X/please%20keep%20denoise%20as%20first%20work%2C%20seems%20clearer/vote"></a></p>

TurboGit:
Also I found astrophoto too  restrictive, I have never done astophoto but I did use this module many times.

rawfiner:
Also poisson noise is also very present in basically every picture
In denoise profile there is a variance stabilisation transform for that (a power, like the cubic root of Lab)

johnny-bit:
I understand it's "best suited" for astrophoto, but then it's still a denoising module, which will work on any photo... but the name "non-local means" indeed says nothing to "self-described average user".

aurelienpierre:
> Also poisson noise is also very present in basically every picture

Yes but long exposures are something special. For average photography, I don't think this module brings anything more than denoise profile, especially since it works in Lab and has to come late in the pipe.

ralfbrown:
I've found it handy for smoothing the sky (with a parametric mask), so I have a preset specifically for that use case.

aurelienpierre:
Alternatively, instead of "astrophoto denoise", what about "long exposure denoise" ?

rawfiner:
I think it would be maybe misleading as we would expect it to handle correctly the hot pixels that are quite frequent in long exposures

Sorry that I don't have a better name in mind yet... I agree that a more meaningful name would be nice but it is hard to find one which fits well the purpose of the module, especially since its usage overlaps with other denoising modules

matt-maguire:
Calling it "astrophoto denoise" suggests one of its primary use cases, but that doesn't preclude people using it for other purposes. In the documentation we can describe what some of those other purposes may be in the introduction to the reference page for the module.

PR #6885, opened 2020-11-13
rename "denoise (bilateral filter)" to "surface blur"

aurelienpierre:
The bilateral filter is one of the first edge-avoiding blurring filters. Its use as a denoising has been improved through the non-local means. It is a rather coarse way to denoise and it is known to overshoot.

However, as a creative filter, it produces interesting results that look like comics or can, if blended in chrominance, quickly even the color of a surface (say, to remove skin redness).

The purpose of the rename is to discourage users to use it to deblur, since we have better modules, but also advertise what it does rather than what it is (and suggest possible uses).

In the future, an RGB guided filter could be added as another options, since the goal is the same, minus the overshoot.

### DIFF
--- a/src/iop/bilateral.cc
+++ b/src/iop/bilateral.cc
@@ -69,7 +69,7 @@ typedef struct dt_iop_bilateral_data_t
 
 const char *name()
 {
-  return _("denoise (bilateral filter)");
+  return _("surface blur");
 }
 
 int default_group()


TurboGit:
Let's vote:

<p><a href="https://api.gh-polls.com/poll/01EQ072CPGY3N0DDZJEYQ6GNPX/new%20name%20seems%20clearer%2C%20please%20rename/vote"></a>
<a href="https://api.gh-polls.com/poll/01EQ072CPGY3N0DDZJEYQ6GNPX/don&#39;t%20care%20about%20the%20name%2C%20don&#39;t%20care%20about%20the%20final%20choice/vote"></a>
<a href="https://api.gh-polls.com/poll/01EQ072CPGY3N0DDZJEYQ6GNPX/please%20keep%20denoise%20as%20first%20work%2C%20seems%20clearer/vote"></a></p>

parafin:
Won't renaming create a problem for users who use this iop? They will try to find the module they are used to enable in new dt version, but would fail. Then they will have to search for release notes or documentation...

aurelienpierre:
> Won't renaming create a problem for users who use this iop? They will try to find the module they are used to enable in new dt version, but would fail. Then they will have to search for release notes or documentation...

Sure, status quo doesn't require adaptation. But status quo, here, is unnecessarily technical.

aurelienpierre:
@parafin https://github.com/darktable-org/darktable/pull/6894 should address part of your concern by adding the former name into module alias, which can be used to search modules in the search bar.

parafin:
Sounds perfect.

aurelienpierre:
Given the votes and the ability to use aliases for module search, the motion passes.

PR #6889, opened 2020-11-13
modulegroups: set current group when entering darkroom

elstoc:
resolves issue whereby, when exiting darkroom with "active" group selected, reentering darkroom changed the group to "none" (i.e. show all)


TurboGit:
Works for me, thanks!

TurboGit:
@elstoc : this has introduce regression, select any other group than active, quit dt, restart and none of the group are selected. can you have a look?

elstoc:
ok

PR #6891, opened 2020-11-13
Prevent crash when pressing edit shortcut for hidden slider

dterrahe:
Assign a shortcut key to the edit action of an iop slider in a module that is not in your active group. Press that key just after entering darkroom. Watch dt crash.

This change doesn't check if the slider is visible; it just catches the case where it has not been allocated an area yet, which caused dt to crash. If the slider has been shown once, but then hidden, the edit area may pop up in unexpected places (or with unexpected size if the panel has been resized) but at least it doesn't crash (as much?)


dterrahe:
push white space change to retrigger failed checks (caused by broken master)

aurelienpierre:
Thanks !

PR #6892, opened 2020-11-13
modulegroups: fix regression caused by #6889

elstoc:


TurboGit:
Issue fixed, thanks !

PR #6894, opened 2020-11-13
Module aliases

aurelienpierre:
1. add the ability to add aliases (alternative names or keywords) to modules, that can be translated
2. add examples aliases in denoise (bilateral) to address concerns in #6885 
3. allow the module search bar to lookup modules by their aliases too.

### DIFF
--- a/src/develop/imageop.c
+++ b/src/develop/imageop.c
@@ -117,6 +117,11 @@ static char *default_description(struct dt_iop_module_t *self)
   return g_strdup("");
 }
 
+static const char *default_aliases(void)
+{
+  return g_strdup("");
+}
+
 static void default_commit_params(struct dt_iop_module_t *self, dt_iop_params_t *params,
                                   dt_dev_pixelpipe_t *pipe, dt_dev_pixelpipe_iop_t *piece)
 {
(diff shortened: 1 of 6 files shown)


parafin:
So aliases are used only for searching or are they discoverable anywhere else in GUI? I don't think it's needed, but maybe there are good ideas on where to display them (e.g. tooltip).

aurelienpierre:
Aliases are metadata only used to search. Since they are an unformatted dump, I wouldn't display them in GUI.

AlicVB (approved):
Simple and clean !

*maybe* just add a comment to say that multiple name are separated by | (and maybe an entry in useless.c ?)

aurelienpierre:
useless.c will be taken care of after feature freeze. Thanks for the review.

src/develop/imageop.c
```diff
@@ -117,6 +117,11 @@ static char *default_description(struct dt_iop_module_t *self)
   return g_strdup("");
 }
 
+static const char *default_aliases(void)
+{
+  return g_strdup("");
```
johnny-bit, line 122: shouldn't this return just `""` not `g_strdup("")`? `g_strdup` without `g_free` causes memleak...

aurelienpierre: `""` without `g_strdup` causes segfault…

johnny-bit: Huh... OK... No problem then.

TurboGit: I'll have a look, you can't have a const and g_strdup() very suspicious !

TurboGit: The only place where module->aliases() is called already to a g_strdump:

`        g_hash_table_insert(module_aliases, module->op, g_strdup(module->aliases()));`

So I bet you have fixed the issue in two places maybe....

TurboGit: All is well with the g_strdup() in default_aliases() so I have removed it.

rawfiner:
great idea!

PR #6895, opened 2020-11-13
export: add tooltip for intent.

TurboGit:


TurboGit:
@aurelienpierre : can you review ? TIA.

src/libs/export.c
```diff
@@ -1303,6 +1303,32 @@ void gui_init(dt_lib_module_t *self)
   dt_bauhaus_combobox_add(d->intent, _("absolute colorimetric"));
   gtk_box_pack_start(GTK_BOX(self->widget), d->intent, FALSE, TRUE, 0);
 
+  tooltip = g_strdup_printf(_("• perceptual : "
+                              "smoothly moves out-of-gamut colors into gamut,"
```
aurelienpierre, line 1307: perceptual is often a proprietary LUT that depends on the destination space. Usually, it compresses the whole gamut range such that the color transitions are preserved indeed.

src/libs/export.c
```diff
@@ -1303,6 +1303,32 @@ void gui_init(dt_lib_module_t *self)
...
+                              "preserving gradations, but distorts in-gamut colors in the process."
+                              "\n\n"
+
+                              "• relative colorimetric : "
+                              "reproduces colors with as little desaturation as possible."
```
aurelienpierre, line 1312: relative colorimetric keeps luminance while reducing saturation until colors fit in gamut

src/libs/export.c
```diff
@@ -1303,6 +1303,32 @@ void gui_init(dt_lib_module_t *self)
...
+                              " space."
+                              "\n\n"
+
+                              "• saturation : "
+                              "designed to present eye-catching business graphics"
+                              " by preserving the saturation. (not suited for photography)."
+                              "\n\n"
+
+                              "• absolute colorimetric : "
+                              "reproduces colors scaled to the white point of the source color"
```
aurelienpierre, line 1324: I would phrase it "adapt white point of the image to the white point of the destination medium and do nothing else"

src/libs/export.c
TurboGit: Ok, I took this as-is, simpler and less verbose.

src/libs/export.c
TurboGit: Works for me.

src/libs/export.c
TurboGit: Ok.

TurboGit:
@aurelienpierre : new version based on your feedback just pushed.

aurelienpierre (approved):

TurboGit:
Thanks, let's merge.

Issue #6655, opened 2020-10-27
Show number of selected images in map module


quovadit:
**Problem**
When dragging one image from the filmstrip to the map, all geocoordinates of all images are overwritten, because all images were selected.

The reason for this is the new (for me unintuitive) way of selecting images in the filmstrip with Alt+click (#5495).

**Solution**
Show the number of currently selected images in map module as it is already done in lighttable and darkroom.
![nr-of-selected-images](https://user-images.githubusercontent.com/5677563/97324230-e65a9800-1871-11eb-9649-bdf965a93a2e.png)

**Alternatives**
- show the number of images while dragging them to the map
- define behaviour how to select images in preferences (3.2 'Alt+click' vs. 'just click')

phweyland:
>     * show the number of images while dragging them to the map

Done with #6871 !

quovadit:
@phweyland : thanks a lot, that's great to see the number of images while dragging

In addition I would also add the information in the header in advance (see linked pull request), what do you think?

phweyland:
Sounds good !

quovadit:
Done with #6871 and #6655

PR #6897, opened 2020-11-13
[Issue #6655] Number of images in map module

quovadit:
Show number of selected images in map module to solve #6655 

map-nr-images


TurboGit:
Thanks !

PR #6899, opened 2020-11-14
create benchmark test

### LINKED ISSUE #6654: Feature request: Benchmark
No

Add function

**Alternatives**
n/a

**Additional context**

Feature request: Benchmark function

IMHO it makes great sense for darktable and helps users with what hardware brings speed improvements, and which not. It's also much easier to compare the different O/S - and what O/S tweaks work, or tweaks that do not help performance.

Affinity Photo will have a benchmark function in the beta (1.9) version. Image below. 

Screenshot 2020-10-13 11 08 30

ralfbrown:
Python script to run darktable-cli and parse timings (pixelpipe and image load), plus associated
sidecars.  The script runs darktable-cli once with a "null" sidecar to warm up caches, then runs it
multiple times with the benchmarking sidecar and reports the average timing plus a throughput
rating.  The rating is essentially the number of benchmark images which could be exported per hour.
It is an approximately linear measure, which means that a machine with double the rating is twice as
fast for running darktable.

The benchmarking sidecar files enable (nearly) every iop that darktable supports.  The "3.4legacy"
and "3.4legacy-inv" sidecars include iops which were deprecated for darktable v3.4.  "3.4legacy-inv"
includes all iops except Scale Pixels, Rotate Pixels, and 3D CLUT (which would have to depend on
an external file; "3.4legacy" additionally omits "Invert" and "Negadoctor".  Some iops have multiple
instances to exercise multiple algorithms (e.g. heal and wavelets for Retouch, bilateral and laplacian
for Local Contrast).

darktable-bench-3.4.xmp is the default sidecar used by the script, and compared to the "legacy" sidecars
omits the now-deprecated iops such as relight and tonemapping.

Resolves #6654


TurboGit (changes requested):
Some changes here:

- The darktable-bench should be made executable.
- It would be good to use darktable-cli pointed to by DARKTABLE_CLI if set. This was done on the integration test to make it runnable on GitLib-CI.
- A bit of doc about adding new bench and reading the results would be good, I would add here a README.txt as for integration test.

ralfbrown:
Done.  Also made sure that /tmp doesn't get littered with scratch files.

ralfbrown:
Rebased and added a new sidecar that includes colorbal-rgb, censorize, chromatic abberations, and retouch in clone mode; and drops vibrance, defringe, spot removal, and basic adjustments.

TurboGit (approved):
I suppose this should have been merged since a long time... Sorry for such delay !

ralfbrown:
That's OK, I wasn't exactly pushing to get it merged, and the delay means that the default sidecar will be up-to-date with the coming release (unless there's another new module I haven't heard about yet about to be merged :-).

Issue #6663, opened 2020-10-28
Shift-Click not only includes sub-hierarchy of tag but performs 'like%' query


quovadit:
**Describe the bug**
Shift-Click on the parent tag of a tag-hierarchy (a) does not only include the sub-hierarchy (b), but performs a 'like' query, which also includes other tags with the same beginning (c).

a) Place|AT|Wien
b) Place|AT|Wien|Kagran
c) Place|AT|Wiener Neudorf

**To Reproduce**
1. Go to 'collect images' > tag
2. Shift-Click on parent tag 'Place|AT|Wien'
3. The query changes automatically to 'Place|AT|Wien%'
4. Images with tag Place|AT|Wiener Neudorf are also shown

**Expected behavior**
- Shift-Click only includes parent tag + sub-hierarchy (same as left-click + Ctr-click)
- the query could be displayed as 'Place|Wien*'

**Screenshots**
![tag-query](https://user-images.githubusercontent.com/5677563/97438725-7ad20280-1925-11eb-9bc1-f11ba8f07042.png)

**Platform:**
- Darktable Version: 3.3.0+1225~g80cd1b20e
- OS: Linux Mint 20 Cinnamon
- OpenCL activated
- NVIDIA Driver Version: 450.80.02

quovadit:
done with #6903

PR #6903, opened 2020-11-14
[Issue #6663] Fix tag-sub-hierarchy

quovadit:
Fix for #6663: only include items in and under the selected hierarchy without using a wildcard % which also would include similar named items

### DIFF
--- a/src/common/collection.c
+++ b/src/common/collection.c
@@ -1441,6 +1441,7 @@ gchar *dt_collection_get_makermodel(const char *exif_maker, const char *exif_mod
 static gchar *get_query_string(const dt_collection_properties_t property, const gchar *text)
 {
   char *escaped_text = sqlite3_mprintf("%q", text);
+  unsigned int escaped_length = strlen(escaped_text);
   gchar *query = NULL;
 
   switch(property)
@@ -1582,9 +1583,9 @@ static gchar *get_query_string(const dt_collection_properties_t property, const 
     case DT_COLLECTION_PROP_TAG: // tag
     
       /* shift-click adds an asterix * to include items in and under this hierarchy without using a wildcard % which also would include similar named items */
-      if (!strcmp(strrchr(escaped_text, '\0') - 1, "*"))
+      if ((escaped_length > 0) && (escaped_text[escaped_length-1] == '*')) 
       {
-        escaped_text[strlen(escaped_text)-1] = 0;        
+        escaped_text[escaped_length-1] = '\0';
         query = dt_util_dstrcat(query, "(id IN (SELECT imgid FROM main.tagged_images AS a JOIN "
                                      "data.tags AS b ON a.tagid = b.id WHERE name LIKE '%s' OR name LIKE '%s|%%'))",
                               escaped_text, escaped_text);
--- a/src/common/collection.c
+++ b/src/common/collection.c
@@ -1509,7 +1509,7 @@ static gchar *get_query_string(const dt_collection_properties_t property, const 
       {
         const gboolean not_tagged = strcmp(escaped_text, _("not tagged")) == 0;
         const gboolean no_location = strcmp(escaped_text, _("tagged")) == 0;
-        const gboolean all_tagged = strcmp(escaped_text, _("tagged%")) == 0;
+        const gboolean all_tagged = strcmp(escaped_text, _("tagged*")) == 0;
         char *escaped_text2 = g_strstr_len(escaped_text, -1, "|");
         if(not_tagged || all_tagged)
           query = dt_util_dstrcat(query, "(id %s IN (SELECT id AS imgid FROM main.images "
--- a/src/common/collection.c
+++ b/src/common/collection.c
@@ -1580,9 +1580,22 @@ static gchar *get_query_string(const dt_collection_properties_t property, const 
       query = dt_util_dstrcat(query, ")");
       break;
     case DT_COLLECTION_PROP_TAG: // tag
-      query = dt_util_dstrcat(query, "(id IN (SELECT imgid FROM main.tagged_images AS a JOIN "
+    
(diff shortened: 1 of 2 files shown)


src/common/collection.c
```diff
@@ -1580,9 +1580,22 @@ static gchar *get_query_string(const dt_collection_properties_t property, const
       query = dt_util_dstrcat(query, ")");
       break;
     case DT_COLLECTION_PROP_TAG: // tag
-      query = dt_util_dstrcat(query, "(id IN (SELECT imgid FROM main.tagged_images AS a JOIN "
+    
+      /* shift-click adds an asterix * to include items in and under this hierarchy without using a wildcard % which also would include similar named items */
+      if (!strcmp(strrchr(escaped_text, '\0') - 1, "*"))
```
TurboGit, line 1585: if escaped_text is an empty string, -1 will point to the char before the actual string which is wrong and could not even be allocated. Use strchr(... '*') and check that it is the last char strchr == strlen or something like that.

TurboGit (changes requested):
fix possible crash

src/common/collection.c
quovadit: @TurboGit : thanks a lot for the hint, I just pushed a fix

src/libs/collect.c
```diff
@@ -2106,7 +2106,7 @@ static void row_activated_with_event(GtkTreeView *view, GtkTreePath *path, GtkTr
          * hierarchy. */
         else if(event->state & GDK_SHIFT_MASK)
         {
-          gchar *n_text = g_strconcat(text, "%", NULL);
+          gchar *n_text = g_strconcat(text, "*", NULL);
```
TurboGit, line 2109: Ok, let's try to improve that. We have:
- click (no wildchar)
- ctrl-click  `|%` at end
- shift-click `%` at end

You changed the shift-click to use `*` but the tooltip is now wrong.

I propose to keep `%` at end (which is uniform with all others collection kind) and check if we end with `%` and not with `|%` in the code above. Should work too, and will not introduce a new wildchar.

src/common/collection.c
```diff
@@ -1580,9 +1581,22 @@ static gchar *get_query_string(const dt_collection_properties_t property, const
       query = dt_util_dstrcat(query, ")");
       break;
     case DT_COLLECTION_PROP_TAG: // tag
-      query = dt_util_dstrcat(query, "(id IN (SELECT imgid FROM main.tagged_images AS a JOIN "
+    
+      /* shift-click adds an asterix * to include items in and under this hierarchy without using a wildcard % which also would include similar named items */
+      if ((escaped_length > 0) && (escaped_text[escaped_length-1] == '*'))
```
TurboGit, line 1586: check here that we have `%` and not` |%` (see below).

TurboGit (changes requested):
A minor rework proposed to keep consistency. TIA.

src/libs/collect.c
quovadit: It was on purpose to introduce a new character because I don't want to change the current wildcard-behaviour.

If a user wants to get all tags starting with 'birthday' he can now enter birthday% and gets tags like 'birthday mary' and 'birthday john'.

This would not work anymore if the new hierarchy-logic uses the same character % as trigger.

So the tooltip is not wrong, because % still works for the birthday% usecase.

Maybe it should be another character instead of * to make it clear that this is a special-function and no wildcard.

Any suggestions?

TurboGit: I was thinking about that indeed ! So what I propose is to just update the tooltip for this new wildchar and explain its usage.

quovadit: So we are talking about the tooltip "type your query, use `%' as wildcard", right?

I would prefer to keep this as is, because:
- the % wildcard works for all collections, the * function is limited to tags
- the wildcard % is obvious to be entered in the textbox, but the function of the asterisk is not trivial
- so I would not enter * manually in the textbox but only use the shift-click, so a longer tooltip could be confusing

Instead it could be better to promote the ctrl-click and shift-click possibility
- tooltip when hovering a tag in the list
- describe it in the user manual

TurboGit: You can special case the tooltip for the tag and geotag (as this is where you have changed the semantic). See, there is already different tooltip for different collections.

quovadit: yes, that's true, I realized that after clicking 'comment'...

Nevertheless I would prefer to explain the shift-click functionality when hovering the tag in the list, because this would be much easier to understand compared to manually adding an asterisk in the textbox.

But if it's better to just change the existing tooltip for tags I will do it of course

TurboGit: Not sure it is easy to add a tooltip to all entries into the treeview! At least, in my view, we need to add the asterisk into the current tooltip as this is the way we current document the use of the search box. So better do it all at the same place, the tricky part is to come with a short and clear description :)

quovadit:
I tried the tooltip in the treeview, but it was very annoying when navigating through the hierarchy, so I am now convinced, that the extension of the existing tooltip in the textbox is the best solution - as @TurboGit suggested.

I pushed two wordings for tag + geotag

TurboGit:
I have merged it manually. And very sorry when squashing the commit together I have taken the authorship for this work. This is very wrong and I usually properly set the author name when committing. Too many things to do and I just did this wrong this time. Again sorry !

quovadit:
no problem, I am happy that the function is committed!

Nilvus:
Thanks for that @quovadit. Just a thing, your issue #6663 has not been closed (not linked because you add "for" between Fix and issue number).

quovadit:
> Thanks for that @quovadit. Just a thing, your issue #6663 has not been closed (not linked because you add "for" between Fix and issue number).

thanks for the info, I am still learning...

PR #6910, opened 2020-11-14
Mask display is not always updated when defining parametric masks with the color picker

hlecleme:
When using the color picker to define the values of the parametric mask channels while the mask display is active, the channel values are updated correctly, but the displayed mask does not seem (always) to be updated correctly.

This seems to occur when using OpenCL and using a blending color space different than the native color space the module is working with.

Updated the references to not access `module->blend_params` while processing the image (changed them to `piece->blendop_data`) to avoid the race condition that could occur.

### DIFF
--- a/src/develop/pixelpipe_hb.c
+++ b/src/develop/pixelpipe_hb.c
@@ -939,20 +939,18 @@ static void _pixelpipe_pick_primary_colorpicker(dt_develop_t *dev, const float *
 }
 
 // returns 1 if blend process need the module default colorspace
-static int _transform_for_blend(dt_iop_module_t *self, dt_dev_pixelpipe_iop_t *piece, const int cst_in, const int cst_out)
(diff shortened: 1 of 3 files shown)


src/develop/pixelpipe_hb.c
```diff
@@ -939,20 +939,18 @@ static void _pixelpipe_pick_primary_colorpicker(dt_develop_t *dev, const float *
 }
 
 // returns 1 if blend process need the module default colorspace
-static int _transform_for_blend(dt_iop_module_t *self, dt_dev_pixelpipe_iop_t *piece, const int cst_in, const int cst_out)
+static gboolean _transform_for_blend(const dt_iop_module_t *const self, const dt_dev_pixelpipe_iop_t *const piece)
```
TurboGit, line 942: yes, thanks :)

TurboGit:
Looks good to me, thanks!

PR #6911, opened 2020-11-14
Update release notes w.r.t. the blending changes

hlecleme:


RELEASE_NOTES.md
```diff
@@ -156,6 +167,9 @@ making a backup is strongly advised.
 - Fix some issues with lighttable display (thumbnails) when using
   HiDPI monitor.
 
+- Fix display of the masks of the individual channels for the Lab/LCh
```
TurboGit, line 170: Isn't that a fix for a recent 3.4 implementation ? So in other word this was not a problem in 3.2 or am I wrong ?

hlecleme: It wasn't working correctly with master before the merge of #4688, so I thought it was an issue present before. Checked with 3.2.1 and surprise, it is working in that version... and actually the output is quite strange: I would have expected a Lab->RGB conversion between the color balance and gamma modules:
```
24.618706 [dev_pixelpipe] took 0.512 secs (3.244 CPU) processed `demosaic' on CPU, blended on CPU [full]
24.621321 [dev_pixelpipe] took 0.003 secs (0.013 CPU) processed `input color profile' on CPU, blended on CPU [full]
24.638229 [dev_pixelpipe] took 0.017 secs (0.110 CPU) processed `color balance' on CPU, blended on CPU [full]
24.655566 [dev_pixelpipe] took 0.016 secs (0.113 CPU) processed `display encoding' on CPU, blended on CPU [full]
24.656016 [dev_process_image] pixel pipeline processing took 0.589 secs (3.721 CPU)
```
The change adding the conversion and thus causing the regression for the display of the channel mask is commit 5c7d52e005f2cca3f136f6371f76bc614a26b24b.
Therefore, removed that line as it was a regression between 3.2.1 and 3.4 ;)

TurboGit:
All good now, Thanks!

PR #6912, opened 2020-11-14
po/soft deprecated

TurboGit:
Let's poll on this !

<p><a href="https://api.gh-polls.com/poll/01EQ3Y6DW27FH8M6D69713D7Z8/soft%20deprecation%20is%20an%20heresy/vote"></a>
<a href="https://api.gh-polls.com/poll/01EQ3Y6DW27FH8M6D69713D7Z8/thanks%2C%20my%20broken%20modules%20are%20still%20accesible%20if%20I%20like/vote"></a></p>

### DIFF
--- a/src/libs/modulegroups.c
+++ b/src/libs/modulegroups.c
@@ -573,8 +573,14 @@ static void _lib_modulegroups_update_iop_visibility(dt_lib_module_t *self)
 
         default:
         {
+          // show deprecated module in specific group deprecated
+          dt_lib_modulegroups_group_t *gr =
+            (dt_lib_modulegroups_group_t *)g_list_nth_data(d->groups, d->current - 1);
+
           if(_lib_modulegroups_test_internal(self, d->current, module)
-             && (!(module->flags() & IOP_FLAGS_DEPRECATED) || module->enabled))
+            && (!(module->flags() & IOP_FLAGS_DEPRECATED)
+                || module->enabled
+                || !strcmp(gr->name, _("deprecated"))))
           {
             if(w) gtk_widget_show(w);
           }
@@ -1080,6 +1086,12 @@ void init_presets(dt_lib_module_t *self)
                        "|grain|highpass|liquify|lowlight|lowpass|monochrome|retouch|sharpen"
                        "|soften|spots|vignette|watermark");
   dt_lib_presets_add(_("modules: default"), self->plugin_name, self->version(), tx, strlen(tx), TRUE);
+  g_free(tx);
+
+  tx = NULL;
+  tx = dt_util_dstrcat(tx, "ꬹ1ꬹ%s|%s||%s", C_("modulegroup", "deprecated"), "basic",
+                       "zonesystem|invert");
+  dt_lib_presets_add(_("modules: deprecated"), self->plugin_name, self->version(), tx, strlen(tx), TRUE);
   g_free(tx);
 
   // if needed, we add a new preset, based on last user config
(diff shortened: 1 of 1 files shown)


AlicVB:
I quite like that solution : still accessible for urgency but need specific action from user each time.
ideally, i think we can even : 
1. add an explicit message on top of the group, something like : "following modules are deprecated because ....  they will be completely remove for new edits in next release"
2. don't allow to clone this preset in the editor
3. add a message on top of each module panel to propose replacement modules "this module is deprecated, better use ..." (this may be redundant with point 1, but not totally)

@TurboGit : I can take care if you want, someone says to me that you have a "little" amount of work this days ;)

TurboGit:
> @TurboGit : I can take care if you want, someone says to me that you have a "little" amount of work this days ;)

Yes, indeed... I still had to fix a down-sampling issue this morning for circle and ellipse forms and I'm reviewing code to fix memory leaks (I found some in modulegroups while working on this PR). Quite busy indeed, and I'll stop working on dt for today I think so yes please go ahead and build on top of this PR for a clear message about why modules are deprecated. Thanks !

elstoc:
This poll needs a third option "these modules are rubbish but we should give our users some warning before we remove stuff"

AlicVB:
@elstoc : in fact this is exactly the purpose here (+ the addition I'm currently doing)

Note that as before **nothing is removed** your old edits will be unchanged. Deprecation only hide the module when editing image that don't have the module in their history (new edits, ...)
I'm sure you are aware of that, but better say it to avoid misunderstanding ;)

elstoc:
Yes I'm aware that deprecation stops the module being used on new edits, and we should do this for 3.6 for sure. But some people are slow to change or will have to spend time working out a new workflow, so it would be nice to let them use their old workflow for a bit more and warn them that they need to move to a new workflow soon. Currently deprecation means "can't use ever on any new edits", and someone is already coming up with a workaround for this on pixls. Polls on github are great and all but developers are a pretty niche audience imo it would be better not to spring deprecation on people like this. I'll go with the majority on this but just want people to bear it in mind.

AlicVB:
@elstoc : again : the purpose here is to let "just-deprecated" module accessible for one more release, before "removing" them for new edits. They will be accessible in a special modulegroup preset (and thus require user "deliberate" action) with all needed messages so no one can ignore them.

elstoc:
Apologies I had misunderstood - I thought they would still be available via search as well but this just adds a modulegroup preset. This is fine, but some people don't use module groups (I can only assume, since we have the option to switch them off) so I still think it would be better if they were still searchable (albeit with "deprecated" as part of the description).

matt-maguire:
I don't see how including the modules in a new group discourages peope from using them at all -- if I don't see a module in my group (eg. you won't see contrast equalizer in the scene-referred groups), I just search for it, switch it on and use it, and I have no idea if it is "deprecated" or not... Seems to me this "soft deprecation" is useless.

elstoc:
> if I don't see a module in my group (eg. you won't see contrast equalizer in the scene-referred groups), I just search for it

In this PR, the deprecated modules can only be used by choosing the deprecated modules preset (not by searching) in modulegroups which makes it a pain to use them (encouraging users to switch)

> and I have no idea if it is "deprecated" or not

This is what @AlicVB is trying to do with his PR - flag the modules as deprecated

TurboGit:
Note to all, this is a way to always keep deprecated module if needed. Just create a style with the deactivated modules. But please don't tell anyone :)

elstoc:
Too late https://discuss.pixls.us/t/some-modules-deprecated/21303 :)

AlicVB:
Now user will see a message on top of all deprecated modules... So no one can say "I was not aware..." and claim for fix or I don't know what on this modules !

Issue #6913, opened 2020-11-14
git master 49904d5a4 compile error in src/libs/histogram.c


wpferguson:
**Describe the bug**
Pulled latest commits.  Build responded with 

```
/home/bill/src/darktable/darktable/src/libs/histogram.c: In function ‘_lib_histogram_process_waveform’:
/home/bill/src/darktable/darktable/src/libs/histogram.c:226:9: error: ‘wf_linear’ is predetermined ‘shared’ for ‘shared’
   shared(wf_linear) aligned(input, wf_linear:64) \
         ^
/home/bill/src/darktable/darktable/src/libs/histogram.c: In function ‘_lib_histogram_draw_waveform_channel’:
/home/bill/src/darktable/darktable/src/libs/histogram.c:516:9: error: ‘wf_display’ is predetermined ‘shared’ for ‘shared’
   shared(wf_display) aligned(wf_linear, wf_display, primaries_linear:64) \
         ^
/home/bill/src/darktable/darktable/src/libs/histogram.c:535:9: error: ‘wf_8bit’ is predetermined ‘shared’ for ‘shared’
   shared(wf_8bit) aligned(wf_8bit, wf_display:64) \
         ^
lib/darktable/plugins/lighttable/CMakeFiles/histogram.dir/build.make:62: recipe for target 'lib/darktable/plugins/lighttable/CMakeFiles/histogram.dir/histogram.c.o' failed
make[2]: *** [lib/darktable/plugins/lighttable/CMakeFiles/histogram.dir/histogram.c.o] Error 1
CMakeFiles/Makefile2:3136: recipe for target 'lib/darktable/plugins/lighttable/CMakeFiles/histogram.dir/all' failed
make[1]: *** [lib/darktable/plugins/lighttable/CMakeFiles/histogram.dir/all] Error 2
make[1]: *** Waiting for unfinished jobs....

```

**To Reproduce**
1. Download latest git
2. Build
3. See error

**Expected behavior**
Build without error

**Platform (please complete the following information):**
 - Darktable Version: current git
 - OS: Ubuntu 18.04

wpferguson:
According to Stack Overflow declaring the variables as const implies sharing, so they don't need to be explicitly shared.  Removing the const declaration let it compile with error. Putting const back and removing the shared() pragmas works also.  So, one or the other works, but not both.

johnny-bit:
This is same error as one described by @jade-nl on irc.

[...]

PR #6914, opened 2020-11-14
Fix histogram.c compile error

wpferguson:
declaring variables const implies sharing, so explicitly sharing in a pragma causes and error

### DIFF
--- a/src/libs/histogram.c
+++ b/src/libs/histogram.c
@@ -36,6 +36,19 @@
 
 #define HISTOGRAM_BINS 256
 
+#ifndef dt_omp_shared
+#ifdef _OPENMP
+#if defined(__clang__) || __GNUC__ > 8
+# define dt_omp_shared(...)  shared(__VA_ARGS__)
+#else
+  // GCC 8.4 throws string of errors "'x' is predetermined 'shared' for 'shared'" if we explicitly declare
+  //  'const' variables as shared
+# define dt_omp_shared(var, ...)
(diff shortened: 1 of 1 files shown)


wpferguson:
Fixes #6913

ralfbrown:
Yep, just hit the same here with GCC 8.4.  Apparently GCC9+ and Clang require the "shared" declaration, while GCC8 complains abouut it....  I added a version-dependent definition for dt_omp_shared to several files (guided_filter.c, nlmeans_core.c, etc) due to this same issue.  We should probably consolidate those into a header file (darktable.h?).

ralfbrown:
Ouch.  Much cleaner and easier to maintain if you use something like my dt_omp_shared macro to move the conditional definition elsewhere, so you don't need to duplicate all of the other directives.

wpferguson:
@ralfbrown if you are going to consolidate the dt_omp_shared macro, then I'll just close this and let you scoop it up with the other files.

wpferguson:
@ralfbrown Fixed it with the macro.  When you decide to consolidate, include this one in the list.  Thanks for macro.

ralfbrown:
I won't have time until next weekend, so go with the PR as is for now so that GCC8 users can compile, and I'll take care of the refactoring later.

wpferguson:
Sounds good to me,

Thanks

TurboGit (approved):
Thanks !

PR #6915, opened 2020-11-14
Soft deprecated

### LINKED ISSUE #6907: Softer module deprecation
A number of modules have recently been deprecated. While I agree with the goal and I don't use any of these modules (or will now use their better replacement), a number of people in the poll stated that they do still use them.

The normal process in most software is to issue a release that warns of pending deprecation before finally removing them in a later release. This gives people a chance to explore alternatives while still using the modules they are used to in the meantime.

As we now have the ability to flag warning messages in module headers I propose that, for 3.4, we use this mechanism to warn users of pending deprecation and then fully remove them in a future release.

AlicVB:
this close #6912 , this fix #6907 
Here is some refinement on top of @TurboGit PR #6912 

it create a special modulegroups preset with all recently deprecated modules.
So this modules are still accessible until next release when they will disappear from modules groups, except for old edit using them.
That way, it let  some time to users to adapt their workflow accordingly.

Warning messages are show on top of each modules, with a proposal of replacement modules.

@aurelienpierre : can I ask you to review all the deprecated message inside modules ? Most are just copy-paste of some of your comments around... eventually we can even add the reason of deprecation but I'm not sure it's worth the pain...
@TurboGit : thanks for the base implementation

### DIFF
--- a/src/iop/colortransfer.c
+++ b/src/iop/colortransfer.c
@@ -112,6 +112,11 @@ int flags()
   return IOP_FLAGS_DEPRECATED | IOP_FLAGS_ONE_INSTANCE | IOP_FLAGS_PREVIEW_NON_OPENCL;
 }
 
+const char *deprecated_msg()
+{
+  return _("This module is deprecated.");
+}
+
 int default_colorspace(dt_iop_module_t *self, dt_dev_pixelpipe_t *pipe, dt_dev_pixelpipe_iop_t *piece)
 {
   return iop_cs_Lab;
--- a/data/themes/darktable.css
+++ b/data/themes/darktable.css
@@ -1403,7 +1403,8 @@ margin: 0 2px;
   color: @button_hover_fg;
 }
 
-#modulegroups-deprecated-msg
+#modulegroups-deprecated-msg,
+#iop-plugin-deprecated
 {
   background-color: @grey_10;
   padding-top: 0.45em;
--- a/data/themes/darktable.css
+++ b/data/themes/darktable.css
@@ -1403,6 +1403,13 @@ margin: 0 2px;
   color: @button_hover_fg;
 }
 
+#modulegroups-deprecated-msg
+{
+  background-color: @grey_10;
+  padding-top: 0.45em;
+  padding-bottom: 0.45em;
+}
+
 /*---------------------------------------------------------------
   - Set sidebars settings on preferences window and filechooser -
   ---------------------------------------------------------------*/
--- a/src/develop/imageop.c
+++ b/src/develop/imageop.c
@@ -122,6 +122,11 @@ static const char *default_aliases(void)
   return "";
 }
 
+static const char *default_deprecated_msg(void)
+{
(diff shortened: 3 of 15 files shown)


TurboGit (approved):
Thanks, probably the last feature merge for 3.4. Thanks. I'll fix some strings where we don't want to introduce capital letters for consistency and we are all set. Thanks AlicVB for handling this so fast.

src/iop/colortransfer.c
```diff
@@ -112,6 +112,11 @@ int flags()
   return IOP_FLAGS_DEPRECATED | IOP_FLAGS_ONE_INSTANCE | IOP_FLAGS_PREVIEW_NON_OPENCL;
 }
 
+const char *deprecated_msg()
+{
+  return _("This module is deprecated.");
```
ralfbrown, line 117: Use color mapping module instead

(color mapping is basically color transfer v2)

chhil:
Git pull :  e8204990d..c8f2d6cd9  master     -> origin/master

Build 3.3.0+1613~gc8f2d6cd9

Is this warning correct with the modules listed under it?
This was an image that was imported prior to this build but was not edited.

image

AlicVB:
it's a bug... that'll be fixed soon :)

ralfbrown:
In addition to what chhil pointed out, **every** non-deprecated module now has a blank space just below its title bar when expanded.
(3.3.0+1618~g831906b21)

AlicVB:
both bugs should be fixed now in master. Thanks for the report !

src/iop/colortransfer.c
AlicVB: see PR #6929 thanks for the info !

ralfbrown:
fix confirmed for both bugs

chhil:
Just got this.... current master

3.3.0+1652~gf4d6d9920


image

TurboGit:
Expected as you are on the deprecated module group if I see correctly the top of the image.

parafin:
But how it looks is very misleading...

AlicVB:
@chhil : seems like a bug. Lens is not deprecated and you doesn't seems to be on the deprecated preset... You have only 1 group shown... I guess you have set the pref to only show search box. right ? if so, I'm currently fixing this !

chhil:
I had a clean fresh install, and that big on/off button looked odd.

I had got a similar thing for crop too when I pasted the image with a red label, you could check that too. 

I have show both (module groups and search).

AlicVB:
but in that case, why do you have only the "active modules" group ?
you've done your own preset ?

chhil:
I think when I install there is no selection and it only shows active modules.

image

Once I select a group one then there is no way to get that big central active modules.

After selecting "all modules"
image

AlicVB:
So... you've found 2 bugs in one it seems :) I'll take care !

chhil:
Do check @aurelienpierre s new modulegroup pteset fix , may be related.

PR #6916, opened 2020-11-14
Correct channel display of JzCzhz hue

hlecleme:
Tried the magic trick of @aurelienpierre on the conversions in the gamma module, and it failed for the JzCzhz hue... This PR tries to fix those issues:

- Fixes a small error in the computation of JzAzBz to XYZ D65
- Use a direct XYZ D65 to sRGB conversion

There is a still a slight difference for the display of a “neutral” color that is probably due to the use of floats instead of doubles for the JzAzBz to XYZ conversion.

@TurboGit I hope this is the last one on this subject!

### DIFF
--- a/src/common/colorspaces_inline_conversions.h
+++ b/src/common/colorspaces_inline_conversions.h
@@ -803,7 +803,7 @@ static inline void dt_XYZ_2_JzAzBz(const float *const DT_RESTRICT XYZ_D65, float
   for(int i = 0; i < 3; i++)
   {
     LMS[i] = M[i][0] * XYZ[0] + M[i][1] * XYZ[1] + M[i][2] * XYZ[2];
-    LMS[i] = powf(fmax(LMS[i] / 10000.f, 0.0f), n);
+    LMS[i] = powf(fmaxf(LMS[i] / 10000.f, 0.0f), n);
     LMS[i] = powf((c1 + c2 * LMS[i]) / (1.0f + c3 * LMS[i]), p);
   }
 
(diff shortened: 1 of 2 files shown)


src/common/colorspaces_inline_conversions.h
```diff
@@ -890,7 +890,7 @@ static inline void dt_JzAzBz_2_XYZ(const float *const DT_RESTRICT JzAzBz, float
 
   // X'Y'Z -> XYZ_D65
   XYZ_D65[0] = (XYZ[0] + (b - 1.0f) * XYZ[2]) / b;
-  XYZ_D65[1] = (XYZ[1] + (g - 1.0f) * XYZ[0]) / g;
+  XYZ_D65[1] = (XYZ[1] + (g - 1.0f) * XYZ_D65[0]) / g;
```
aurelienpierre, line 893: this does not seem right

hlecleme: Cf. Appendix A of reference document.

Inversion of first step:
```
X' = b*X - (b-1)*Z
Y' = g*Y - (g-1)*X
Z' = Z
```
becomes (starting with `Z`, then `X` and finally `Y`; `XYZ` is {`X'`,`Y'`,`Z'`} and `XYZ_D65` is {`X`,`Y`,`Z`}):
```
Z = Z'
X = (X' + (b-1)*Z)/b = (X' + (b-1)*Z')/b
Y = (Y' + (g-1)*X)/g
```

Also by experimentation it gives a better D65 “white” when setting the chroma Cz (or Az and Bz) to zero.

aurelienpierre: Hmm not sure I understand, but I'm pretty busted.

aurelienpierre: @hlecleme is the OpenCL path updated with that ?

hlecleme: @aurelienpierre  The function `dt_JzAzBz_2_XYZ` is only used by `gamma.c` which doesn't seem to have an OpenCL version → that function/code is not present in the OpenCL kernel files.

aurelienpierre: Ok thanks. Merging then.

PR #6927, opened 2020-11-15
fix disappearing dpi box in export

dterrahe:
Improve behaviour of #6079 for narrow panels. The use of the layout widget caused the dpi field and label to quietly move off the screen to the right (without a scroll bar appearing). This is resolved here by introducing a flow box that pushes the field to the next line if it doesn't fit.

Also hides the row that shows the pixel dimensions when it is not needed.

There is (for me) still an issue with advertised (in tooltip) middle clicking to reset to zero; if there is a number on the clipboard, this gets pasted instead (making reset to zero even more needed, but unavailable).

Previously (even before fixing ellipsizing of "that is equal" label):
image

new:
imageimage

### DIFF
--- a/src/libs/export.c
+++ b/src/libs/export.c
@@ -52,11 +52,11 @@ DT_MODULE(7)
 typedef struct dt_lib_export_t
 {
   GtkWidget *dimensions_type, *print_dpi, *print_height, *print_width;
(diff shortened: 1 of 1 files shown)


TurboGit (approved):
Works for me. Thanks !

PR #6933, opened 2020-11-15
Improve deprecated CSS

Nilvus:
@TurboGit and @AlicVB: change discussed and shown in Matrix.

For others, eventually, the change on darker theme for example:


TurboGit:
Works for me, thanks !

PR #6941, opened 2020-11-16
imageop: prevent module being re-enabled

### LINKED ISSUE #6924: modules on/off not respected when module is on and expanded. Clicking on/off collapses module instead of switching it off.
3.3.0+1613~gc8f2d6cd9

Pasted history from image1 to kosher image2.

Realized the crop needed to be fixed in image2.

Clicked the enabled crop module to expand it.

Then without changing the crop clicked the module on/off to turn it off (the mouse was specifically on the in/off and it was highlighted), but it treats the click as module collapse.

elstoc:
Resolves #6924

### DIFF
--- a/src/develop/imageop.c
+++ b/src/develop/imageop.c
@@ -1182,13 +1182,12 @@ static void dt_iop_gui_off_callback(GtkToggleButton *togglebutton, gpointer user
 
       if(dt_conf_get_bool("darkroom/ui/activate_expand") && !module->expanded)
         dt_iop_gui_set_expanded(module, TRUE, dt_conf_get_bool("darkroom/ui/single_module"));
+
+      dt_dev_add_history_item(module->dev, module, FALSE);
     }
     else
     {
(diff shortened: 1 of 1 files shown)


elstoc:
This definitely fixes the bug and doesn't _appear_ to cause any other effects, but I don't know what this functionality is doing so needs review.

TurboGit:
@elstoc : the operation_tags_filter() was introduced back in 2011 in commit cd24715ef91d0223130cb0342ac6cc9141117b2b.

I'm not sure to see all the consequence of using this mode and as there is no reference to an issue hard to tell.

I have tested this and indeed it fixes the issue. We may want to check all modules using distortion and how they works when using together (lens, perspective correction, crop & rotate, maybe liquify).

elstoc:
@TurboGit I think I've pushed an alternative that's less likely to break things.

TurboGit:
@elstoc : works for me too and indeed looks lot safer. thanks.

PR #6944, opened 2020-11-16
channelmixerrgb: when creating new instances avoid two CAT.

TurboGit:
When an instance of the channelmixerrgb already exists and is
already doing CAT set the new instance adpatation to none.

Will make easier working with multiple instance of channelmixerrgb.

### DIFF
--- a/src/iop/channelmixerrgb.c
+++ b/src/iop/channelmixerrgb.c
@@ -1880,6 +1880,28 @@ void reload_defaults(dt_iop_module_t *module)
   d->illuminant = module->get_f("illuminant")->Enum.Default;
   d->adaptation = module->get_f("adaptation")->Enum.Default;
 
+  // note that if there is already an instance of this module with an
+  // adaptation set we default to RGB (none) in this instance.
+
+  gboolean CAT_already_applied = FALSE;
+
+  GList *iop = darktable.develop->iop;
+  while(iop)
+  {
+    const dt_iop_module_t *m = (dt_iop_module_t *)iop->data;
+    if(module != m && m->enabled && !strcmp(m->op, "channelmixerrgb"))
+    {
+      const dt_iop_channelmixer_rgb_params_t *mp =
+        (dt_iop_channelmixer_rgb_params_t *)m->params;
+      if(!(mp->adaptation == DT_ADAPTATION_RGB || mp->illuminant == DT_ILLUMINANT_PIPE))
+      {
+        // CAT already applied, default to none (bypass)
+        CAT_already_applied = TRUE;
+        break;
+      }
+    }
+    iop = g_list_next(iop);
+  }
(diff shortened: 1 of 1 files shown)


src/iop/channelmixerrgb.c
```diff
@@ -1880,6 +1880,29 @@ void reload_defaults(dt_iop_module_t *module)
...
+
+  GList *iop = darktable.develop->iop;
+  while(iop)
+  {
+    const dt_iop_module_t *m = (dt_iop_module_t *)iop->data;
+    if(module != m && m->enabled && !strcmp(m->op, "channelmixerrgb"))
+    {
+      const dt_iop_channelmixer_rgb_params_t *mp =
+        (dt_iop_channelmixer_rgb_params_t *)m->params;
+      if(mp->adaptation != DT_ADAPTATION_RGB)
```
aurelienpierre, line 1896: if(!(mp->adaptation == DT_ADAPTATION_RGB || mp->illuminant == DT_ILLUMINANT_PIPE))

aurelienpierre (changes requested):

aurelienpierre:
OOooooh so we can fetch one module's parameters from another module like that ?

TurboGit:
> OOooooh so we can fetch one module's parameters from another module like that ?

Yes We Can !

src/iop/channelmixerrgb.c
TurboGit: Done.

dterrahe:
Does this make the "2nd" instance permanently behave different from the "1st"? What about if the ordering is changed? And after saving/reloading/switching images?

TurboGit:
> Does this make the "2nd" instance permanently behave different from the "1st"? What about if the ordering is changed? And after saving/reloading/switching images?

There is no assumption about instance number. Only if one of the active instance is having CAT enabled then the newly created one won't have it. If a single instance exists and does not have CAT the new instance will have CAT.

johnny-bit:
This seems to also allow multiple instances WITH CAT, just that the default for new instances is bypass :)

matt-maguire:
Having multiple instances of channelmixerrgb with CAT is a valid use case (eg. when used together with masks). If one does this, there is a warning provided, however the software shouldn't actively interfere which what the user is trying to do.

chhil:
May want to document this behavior. 

@matt-maguire did mention in another thread that people may want to use masking and still use white balance and in that case they may not want to bypass it.

I am happy with this bypass on second module, already got bit by it.

TurboGit:
Yes all keep possible, yet by default (the expected more common usage) is that only one instance has CAT.

TurboGit:
> I am happy with this bypass on second module, already got bit by it.

Yes, I've done that because I saw your message.

PR #6947, opened 2020-11-16
src:common: Add missing string.h for strerror() in grealpath.h

cryptomilk:
This is later used for printing errors.


TurboGit:
Quite safe I would say :)

TurboGit:
Thanks !

PR #6949, opened 2020-11-16
modulegroup: force display tab reodering options to appear

aurelienpierre:
This started because I accidentally removed the presets allowing to change the layout. Fact is the presets are not read-only, but also they are stored in database. Once removed, they can't be restored.

I'm not sure the layout reordering should be a preset. I think it would be more robust to have it in a function that stays in the program and dynamically re-allocate the tabs. Also, it's confusing to mix layouts and lists of modules in presets, as layouts are views while lists of modules are content.

### DIFF
--- a/src/libs/modulegroups.c
+++ b/src/libs/modulegroups.c
@@ -1109,18 +1109,16 @@ void init_presets(dt_lib_module_t *self)
   g_free(tx);
 
   // if needed, we add a new preset, based on last user config
-  if(!dt_conf_key_exists("plugins/darkroom/modulegroups_preset"))
-  {
-    tx = _preset_retrieve_old_layout(NULL, NULL);
-    dt_lib_presets_add(_("previous config"), self->plugin_name, self->version(), tx, strlen(tx), FALSE);
-    dt_conf_set_string("plugins/darkroom/modulegroups_preset", _("previous layout"));
-    g_free(tx);
+  tx = _preset_retrieve_old_layout(NULL, NULL);
+  dt_lib_presets_add(_("previous config"), self->plugin_name, self->version(), tx, strlen(tx), TRUE);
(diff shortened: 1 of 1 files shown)


TurboGit:
@AlicVB : we need your review here.

src/libs/modulegroups.c
```diff
@@ -1109,18 +1109,16 @@ void init_presets(dt_lib_module_t *self)
   g_free(tx);
 
   // if needed, we add a new preset, based on last user config
-  if(!dt_conf_key_exists("plugins/darkroom/modulegroups_preset"))
-  {
```
AlicVB, line 1113: if you remove that check, that will means we redo the pre-3.4 preset retrieval at each startup, even for fresh install. and more problematic that will create the "previous_config"... presets for fresh install too.
As you make this retrieval-presets read-only in next lines, that should be enough...

AlicVB (changes requested):
Well... I'm really not sure to understand everything in @aurelienpierre msg...
> I accidentally removed the presets allowing to change the layout

Are we speaking of the entry "manage presets..." at the bottom of the the preset menu ? In that case it's really another bug as this is not a "special" preset at all but, as you say, a special function that allow to modules to do specific thing things to edit presets (here we show and use the editor windows) Although it's only used in this lib for instance, it can be used in every other modules if needed.

> Fact is the presets are not read-only

Built-in presets should be read-only, as every other built-in presets, otherwise it's a bug... As in other modules, they are defined in init_presets so even in case of database corruption, they will be recreated on next launch (this allow dev to change them on the fly too, for example to add/remove modules inside.

Anyway, I quite don't get how this change are related to your issue, that's why I say that I don't understand ;) I think we are not speaking of the same things...

PR #6957, opened 2020-11-17
suppress nan/inf history diffs

dterrahe:
history diff for white balance usually shows
`emerald nan -> nan` or `nan -> inf`

this change suppresses such messages but will still show rubbish if they are valid numbers. individual modules have to clean params of values that are not used.

also now translates "EV" in format for blend fulcrum diffs.

### DIFF
--- a/src/libs/history.c
+++ b/src/libs/history.c
@@ -756,7 +756,7 @@ static gchar *_lib_history_change_text(dt_introspection_field_t *field, const ch
     }
     break;
   case DT_INTROSPECTION_TYPE_FLOAT:
-    if(*(float*)o != *(float*)p)
+    if(*(float*)o != *(float*)p && (isfinite(*(float*)o) || isfinite(*(float*)p)))
       return g_strdup_printf("%s\t%.4f\t\u2192\t%.4f", d, *(float*)o, *(float*)p);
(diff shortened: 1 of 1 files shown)


TurboGit (approved):
Looks good to me!

src/libs/history.c
```diff
@@ -882,7 +886,7 @@ static gboolean _changes_tooltip_callback(GtkWidget *widget, gint x, gint y, gbo
     add_blend_history_change_enum(blend_cst, _("colorspace"), dt_develop_blend_colorspace_names);
     add_blend_history_change_enum(mask_mode, _("mask mode"), dt_develop_mask_mode_names);
     add_blend_history_change_enum(blend_mode, _("blend mode"), dt_develop_blend_mode_names);
-    add_blend_history_change(blend_parameter, "%.2f EV", _("blend fulcrum"));
+    add_blend_history_change(blend_parameter, _("%.2f EV"), _("blend fulcrum"));
```
TurboGit, line 889: Yes, nice catch before the string freeze!

TurboGit: It seems it was already translated anyway, so it wont have broken the translation but better have it translated anyway.

PR #6963, opened 2020-11-17
temperature: fix as-shot setting, fixes #6926

### LINKED ISSUE #6926: White balance: Wrong output vs settings combo + green image using as shot button (legacy)
**Describe the bug**
When I open a fresh image in darkroom and look at the white balance module I see that the *white balance settings* are set to *user modified*, clicking on *as shot* icon makes the image green.

This is done while *chromatic adaption* set to *legacy*, on a fresh image and no further editing/resetting done.

**To Reproduce**
1. Set *auto apply chromatic adaption defaults* to *legacy*
2. Open an image in darkroom
3. Expand the white balance module
4. Default is set to *user modified* -> which is incorrect
5. click on *as shot* -> Image turns green

**Expected behaviour**
1. The modules shows that the camera value is taken as default
2. Buttons react "as expected" when clicking them

**Screenshots**
When freshly opened:
screen1

After clicking on *as shot*:
screen2

**Platform**
* darktable        : 3.3.0+1617 ge1d2411d8
* OpenCL installed : Yes (nvidia-opencl-common 450.80.02-1)
* OpenCL activated : Yes
* Graphics card    : GeForce RTX 2060 SUPER 7979 MiB
* Graphics driver  : nvidia (450.80.02)
* OS               : Linux (kernel 5.8.10-1)
* Distro           : Debian Buster (rel. 10.6)

**Additional context**
* Self compiled : Yes.
* Compilers     : gcc 8.3.0 / clang 10.0.1
* Cmake         : 3.13.4
* Gtk           : 3.24.5-1
* Glib          : 2.58.3-2
* Sqlite        : 3.27.2
* Optimization  : None

TurboGit:


TurboGit:
@aurelienpierre : This fix the issue but I'm not sure it is correct in all cases. Can you review ?

aurelienpierre:
@TurboGit The logic is ok, but you need to ensure that the pointer g is not NULL aka we have a GUI.

TurboGit:
I see, but there is lot of `g->...` after those lines ! So I'll rework all this to include all the lines accessing g into the if section and will merge.

TurboGit:
> @TurboGit The logic is ok, but you need to ensure that the pointer g is not NULL aka we have a GUI.

We are in the section where g is not null.

TurboGit:
So let's merge. Thanks for the review.

Issue #6945, opened 2020-11-16
[Histogram] odd colors


AxelG-DE:
**Describe the bug**

In my histogram I have odd colors, but I admit, it might be intentional

**Expected behavior**

as previously

**Screenshots**

NEW (no groups is due to bisecting)
![Screenshot_20201116_150943](https://user-images.githubusercontent.com/49004103/99265215-bb84b380-2821-11eb-8de7-00bf917c793f.png)

previous
![Screenshot_20201116_151157](https://user-images.githubusercontent.com/49004103/99264594-f4705880-2820-11eb-9707-605d383e345a.png)

**Bisect**
bisect gave me:

```
855c230401f0c585b38c3ab7660d37f9d034f13a is the first bad commit
commit 855c230401f0c585b38c3ab7660d37f9d034f13a
Author: Dan Torop <dant@pnym.net>
Date:   Sat Nov 14 11:57:24 2020 -0500

    histogram: draw waveform in display profile (#6135)
    
    * histogram: draw waveform in display profile
    
    Generate waveform and RGB parade in Linear Rec709 RGB. Convert to
    display colorspace before drawing.
    
    Advantages:
    
    - Gets rid of gamma code when processing image to produce a
    waveform. This lets the process happen extremely quickly, in a single
    loop. No longer need code to cache powf() output to speed up this
    transformation. Use darktable's fast colorspace conversion code instead.
    
    - In a wide gamut display, if blue is rendered as (0,0,255) in display
    profile, it may practically invisible. (0,0,255) in Rec.709 is more
    visible.
    
    - The new code allocates 3 times the amount of permanent memory (now
    360 x 175 x (sizeof(float) x 2 + sizeof(uint8_t))) = 553K, vs. 184K in
    the old code. But the old code would allocate approx. 379K to process
    the histogram and 984K to draw a waveform histogram. The new code
    doesn't use any such temporary buffers. There's no malloc needed during
    processing or drawing.
    
    * histogram: use stylesheet to set waveform colors
    
    Set specific values for red, green, and blue in the waveform, rather
    than using the primaries of Rec.709. Switch to drawing in Rec.2020 --
    a reasonable colorspace for contemporary monitors.
    
    * histogram: use stylesheet colors for RGB parade
    
    * histogram: use stylesheet colors for regular histogram

 data/themes/darktable.css |   1 +
 src/bauhaus/bauhaus.c     |   1 +
 src/bauhaus/bauhaus.h     |   2 +-
[...]

PR #6968, opened 2020-11-18
Optimize colorspaces conversions

aurelienpierre:
Since colorspaces conversion in-pipe are done a lot more now that we use custom blending spaces, this ensures they are as fast as possible.

On CPU, this drops typical runtimes from 12-20 ms to 8-12 ms.
On GPU, from 8-12 ms to 4-9 ms.
The histogram conversion (from non-linear RGB to non-linear RGB, with 2 LUTs searches) drops from 22-24 ms on both CPU and GPU to 11 ms on CPU and 15-16 ms on GPU.

I have removed entirely the SSE2 code that was commented, since even after optimization, it was at least twice as slow as the pure C optimized by GCC10. For future reference, colorin and colorout use the same code, it might be good to remove the SSE2 function for them as well. 

I have also factorized the RGB -> XYZ -> RGB conversions by premultiplying the RGB -> XYZ and XYZ -> RGB matrices before the pixel loop. That saves 1 matrix multiplication per pixel.

### DIFF
--- a/data/kernels/colorspace.cl
+++ b/data/kernels/colorspace.cl
@@ -72,13 +72,11 @@ inline float4 lab_f_inv(float4 x)
 inline float4 Lab_to_XYZ(float4 Lab)
 {
   const float4 d50 = (float4)(0.9642f, 1.0f, 0.8249f, 0.0f);
-  float4 f, XYZ;
+  float4 f;
   f.y = (Lab.x + 16.0f)/116.0f;
   f.x = Lab.y/500.0f + f.y;
   f.z = f.y - Lab.z/200.0f;
-  XYZ = d50 * lab_f_inv(f);
-
-  return XYZ;
+  return d50 * lab_f_inv(f);
 }
 
 inline float4 prophotorgb_to_XYZ(float4 rgb)
--- a/data/kernels/colorspaces.cl
+++ b/data/kernels/colorspaces.cl
@@ -29,12 +29,11 @@ colorspaces_transform_lab_to_rgb_matrix(read_only image2d_t in, write_only image
   if(x >= width || y >= height) return;
 
   float4 pixel = read_imagef(in, sampleri, (int2)(x, y));
+  pixel.xyz = Lab_to_XYZ(pixel).xyz;
+  pixel = matrix_product(pixel, profile_info->matrix_out);
 
-  float4 xyz, RGB;
-
-  xyz = Lab_to_XYZ(pixel);
-  RGB = matrix_product(xyz, profile_info->matrix_out);
-  pixel.xyz = apply_trc_out(RGB, profile_info, lut).xyz;
+  if(profile_info->nonlinearlut)
+    pixel = apply_trc_out(pixel, profile_info, lut);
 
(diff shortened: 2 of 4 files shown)


Nilvus:
Seems intersting. Maybe (or not) this could be linked also to histogram work by @dtorop and issue #6945

dtorop:
Indeed interesting. I think this matters for #6135 and will speed up waveform drawing. What I'm working on to fix #6945 is all GUI elements drawn via Cairo in a buffer of uint8_t, so I'm using LittleCMS rather than the nice float code which @aurelienpierre is improving.

TurboGit:
Sounds really promising !

aurelienpierre:
> Indeed interesting. I think this matters for #6135 and will speed up waveform drawing. What I'm working on to fix #6945 is all GUI elements drawn via Cairo in a buffer of uint8_t, so I'm using LittleCMS rather than the nice float code which @aurelienpierre is improving.

Never. Use. LittleCMS. When. You. Can. Avoid. It. :D That shit is slow as hell, it's only good to parse ICC profiles, extract data, and feed them to optimized pixel loops.

On another note, I just noticed that colorin takes about 21-23 ms on my system with CPU, to process exactly what is done in the PR in less than 10 ms, so it really needs a rewrite.

TurboGit:
> so I'm using LittleCMS

Note that LittleCMS is an optional dependency, so you cannot rely on it.

parafin:
> > so I'm using LittleCMS
> 
> Note that LittleCMS is an optional dependency, so you cannot rely on it.

It is not an optional dependency.

TurboGit:
@parafin : Indeed, I was wrong, the use of littlecms can be deactivated for for handling the output color profile only. Sorry for the noise!

dtorop:
I'm glad LittleCMS isn't optional -- for now. To cut down on cmsDoTransform*() use, a uint8_t version of dt_ioppr_transform_image_colorspace*() would make a difference. I don't think that'd be terribly taxing to write, given the current float code. I have this idea that iop_profile.c code only handles perceptual intent, which may be why colorout in particular keeps a path to use LittleCMS.

TurboGit:
Manually merged and queued for 3.4.1. Thanks.

MStraeten:
commit fed438bc8984482a5cc23b0eb94d65e1609e76c5
Author: Aurélien PIERRE <aurelien@aurelienpierre.com>
Date:   Wed Nov 18 14:06:06 2020 +0100

    optimize in-pipe color spaces conversion for OpenCL

    fix the order of matrices multiplication

 data/kernels/colorspace.cl  |  6 ++----
 data/kernels/colorspaces.cl | 30 +++++++++++++++---------------
 src/common/iop_profile.c    | 19 ++++++++++++++++++-

changes the behaviour of Lut3d module:

before
before

after
after

used LUT:
Apple Spectacular Cold

on osx with/without opencl enabled

xmp:
[mire1.cr2.xmp.txt](https://github.com/darktable-org/darktable/files/5724879/mire1.cr2.xmp.txt)

PR #6970, opened 2020-11-18
modulegroup fix : quick way to add modules

AlicVB:
As pointed by some of you, new modulegroups is way more flexible but as a big issue, especially for "old" users workflows : it's now way more difficult to quickly show modules for temporary use.

This PR allow to right-click on the group button to show a menu with all modules to add.
Internally, when adding a module this way, a temporary preset named "last modified layout" is created on the fly.

I know we are actual in feature freeze, but imho, this clearly solve an issue as current state breaks some legacy workflows. Codewise, the change are imo quite reduced, safe and not intrusive at all...

### DIFF
--- a/src/libs/modulegroups.c
+++ b/src/libs/modulegroups.c
@@ -41,6 +41,9 @@ DT_MODULE(1)
 #define DEPRECATED_PRESET_NAME     "modules: deprecated"
 #define T_DEPRECATED_PRESET_NAME _("modules: deprecated")
 
+#define CURRENT_PRESET_NAME "last modified layout"
+#define T_CURRENT_PRESET_NAME _("last modified layout")
+
 // if a preset cannot be loaded or the current preset deleted, this is the fallabck preset
 
 #define PADDING 2
@@ -128,6 +131,8 @@ static void _lib_modulegroups_viewchanged_callback(gpointer instance, dt_view_t 
 
 static void _manage_preset_update_list(dt_lib_module_t *self);
 static void _manage_editor_load(const char *preset, dt_lib_module_t *self);
+
+static void _buttons_update(dt_lib_module_t *self);
 
 const char *name(dt_lib_module_t *self)
 {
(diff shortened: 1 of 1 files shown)


aurelienpierre (approved):
That sounds like a better, however incomplete, way to revert the user-experience regression induced by the layout change. At least now we have a quick way to show modules in UI. Last step will be a quick way to remove them from UI.

Test shows no runtime problem. I will let Pascal check for the Gtk validity and memory leaks ;-)

AlicVB:
@aurelienpierre : add+remove is already done here. But I prefer to postpone that for 3.4.1... I'll do a PR once this one is merged to keep git happy ;)

TurboGit:
I think there is something wrong. Select the default preset "workflow: scene-referred". Right click on basic button the list seems ok with the recommended modules first and the other after. Likewise for all buttons except the last one where this is no section.

TurboGit:
Or maybe just because there is no recommended modules for this group ?

AlicVB:
hmm... yes, if no section is shown, that means that there's no recommended modules to add. Either they are all here already, either no recommended modules are bound to the group. That's already like that in the editor...

What is strange, is that here, with preset "workflow: scene-referred" that doesn't happen, I have recommended modules for all groups... With "modules: default", I have exactly what you mentioned... are you sure you can reproduce with built-in "workflow: scene refereed" preset ?

TurboGit:
Yes, I can reproduce all the time with "workflow: scene-referred" & ""workflow: display-referred". It works as expected for all other presets. For "modules: all" I can reproduce but here it is expected and the same for all groups.

BTH, this new menu should probably be deactivated for the "modules: deprecated" where it makes no sense.

TurboGit:
I confirm, just tested with a memory library and clean darktablerc file and same issue for both scene & display referred presets.

TurboGit:
As we are at it, can we put an 's' to others in line 1433:

`    lb = gtk_label_new(_("others"));`

TurboGit:
> As we are at it, can we put an 's' to others in line 1433:
> 
> ` lb = gtk_label_new(_("others"));`

Or better use a translation taking into account the plural. I'll look at this in a separate PR.

AlicVB:
> Or better use a translation taking into account the plural

is it worse the pain ? singular would mean that *all-1* existing modules have been added to one unique group. Not sure it will happen...

TurboGit:
> is it worse the pain ? singular would mean that _all-1_ existing modules have been added to one unique group. Not sure it will happen...

Not sure but "recommended" in English is a common word and we should in French translate it as "recommandés" which may be wrong in other part of the code. So at the end I have also used "recommended module" (with plural form) as it was not more costly than that.

TurboGit:
@AlicVB : see #6972

AlicVB:
OK, your other bug was a nice one : that was a translation problem !
well, in fact, in the code we used C_("modulegroup", "effect") in some places and _("effect") in some other... everything changed to contextual translation and force pushed !

src/libs/modulegroups.c
```diff
@@ -1433,18 +1397,20 @@ static void _manage_editor_module_add_popup(GtkWidget *widget, gpointer data)
...
 
         // does it belong to recommended modules ?
         GtkWidget *vbc = vb2;
         if(((module->default_group() & IOP_GROUP_BASIC) && g_strcmp0(gr->name, _("base")) == 0)
            || ((module->default_group() & IOP_GROUP_COLOR) && g_strcmp0(gr->name, _("color")) == 0)
            || ((module->default_group() & IOP_GROUP_CORRECT) && g_strcmp0(gr->name, _("correct")) == 0)
            || ((module->default_group() & IOP_GROUP_TONE) && g_strcmp0(gr->name, _("tone")) == 0)
-           || ((module->default_group() & IOP_GROUP_EFFECT) && g_strcmp0(gr->name, _("effect")) == 0)
+           || ((module->default_group() & IOP_GROUP_EFFECT)
+               && g_strcmp0(gr->name, C_("modulegroup", "effect")) == 0)
```
TurboGit, line 1409: Ah that was it ! Now the question is how could this have been working for you :( Some code are mystery to me :)

AlicVB: because I mostly use dt in English...

TurboGit: All clear then !

TurboGit:
Seems good now, I'll wait for the GitLab-CI full report and will merge if all goes well.

We will need a bit of CSS from @Nilvus for this new menu which is missing a border.

AlicVB:
this menu is not really new as it's already used inside the layout manager, but yes some css is certainly needed !

TurboGit:
Thanks !

Nilvus:
> Seems good now, I'll wait for the GitLab-CI full report and will merge if all goes well.
> 
> We will need a bit of CSS from @Nilvus for this new menu which is missing a border.

All other popup menus in darktable never had a border. This menu needs at least a padding. I'm not against adding a border but I think is not necessary with a little padding. But if we add a border there, that would mean for consistency a lot of others popup menu (preset, styles in darkroom, gamut, display, etc) to have same borders.

TurboGit:
@Nilvus : Well there is a border in fact for grey & darker theme, but for dark there is no and so the content of the menu is not delimited. See:

dark:

image

darker:

image

TurboGit:
And BTW, the border is maybe a bit too visible a thinner may be better. And yes, padding (as you've already proposed is needed).

Nilvus:
@TurboGit: strange, but other menus (check presets ones, or styles on bottom bar in darkroom) doesn't have borders. I will check that ASAP.

PR #6670, opened 2020-10-29
Implement styles dialog when exporting style[s].

darkelectron:
This implements a dialog for overwriting existing style(s) when exporting. It uses a similar dialog to copy/move dialog from file managers.

Hey, @junkyardsparkle.

PR #6971, opened 2020-11-18
Implement overwrite dialog when importing style[s].

darkelectron:
This adds a dialog when importing styles, similar to the one when exporting(#6670).

Adds a toast message when unable to read a file on import.

### DIFF
--- a/src/common/styles.c
+++ b/src/common/styles.c
@@ -1505,6 +1505,7 @@ void dt_styles_import_from_file(const char *style_path)
   else
   {
     // Failed to open file, clean up.
+    dt_control_log(_("could not read file `%s'"), style_path);
     g_markup_parse_context_free(parser);
     dt_styles_style_data_free(style, TRUE);
     return;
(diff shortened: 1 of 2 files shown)


TurboGit (changes requested):
If I try to import two styes already imported I get the dialog asking for confirmation. I select the tick to apply my choice to all styles and click overwrite. Still I get the message "style ... already present.", seems like the option to use the choice for all styles is not properly used.

darkelectron:
@TurboGit I can't seem to reproduce on my side.

One scenario I suspect is that the filename may not be the same as the style name. Meaning `style_name.dtstyle` and the name in the xml(`<info><name>stylename<name>...`) file are different. That will produce the "already exist" message, since primarily what I use to check, is the filename instead the name in the xml or both(haven't found a way, yet).

johnny-bit:
styles can use hierarchy searator `|` which isn't really liked character for filenames, so style can have name `B&W|high contrast`, which will be in xml written as `B&amp;W|high contrast`, but will probably have name like `B_W_high_contrast.dtstyle`. Obviously what needs to be unique is name that will land in the DB :)

darkelectron:
@TurboGit could you check if its okay now. Pushed this a while back.

What's happening now is: I get the name from the xml file.

TurboGit:
Ok, works for me now.

TurboGit (approved):

PR #6972, opened 2020-11-18
support plural in list of modules

TurboGit:
And update French translation.


AlicVB (approved):
all good for me. You're right, that was needed for "recommended" too. (I trust you on the po file)
Thanks !

PR #6973, opened 2020-11-18
culling : ensure we don't loose focus

### LINKED ISSUE #6848: Using reject in image preview with filtered view does not fully advance to the next image
**Describe the bug**
When rejecting images from the light table image preview, the next image is not selected. As a result, pressing reject again will un-reject the previous image instead of rejecting the next one. See the steps below for more details.

**To Reproduce**
Start in the light table view with multiple images, all with 1 star
- Set the view to >= 1 star
- Select the first image by clicking on it
- Enter the image preview by pressing and holding W
- Reject the image (press R). Note that the next image appears (expected), but is not selected and does not have a white border (not expected)
- Reject again. Note that the rating does not change (not expected), but that the current image is now selected.
- exit the image preview by releasing W. Note that the first image is now un-rejected.

Note that the mouse must not move while performing these steps.

**Expected behavior**
When performing the reject in the steps above I would expect the next image to be selected in addition to being shown. The end result should be the following workflow:
- Select the first image by clicking on it
- Enter the image preview by holding W
- Reject the first image by pressing R, the second image appears
- Reject the second image by pressing R, the third image appears
- etc

Note that this is the behavior when not using the image preview, but simply hovering the cursor over the first image.

 - Darktable Version: Master (3.3.0). Also reproducible on 3.2.1.
 - OS: Arch linux

**Other info**
- This seems to be related to https://github.com/darktable-org/darktable/issues/6333, though I do not see the issue reported there.

AlicVB:
this fix #6848

### DIFF
--- a/src/dtgtk/culling.c
+++ b/src/dtgtk/culling.c
@@ -1534,6 +1534,26 @@ void dt_culling_full_redraw(dt_culling_t *table, gboolean force)
   // we prefetch next/previous images
   _thumbs_prefetch(table);
 
+  // ensure one of the shown image as the focus (to avoid to keep focus to hidden image)
+  gboolean in_list = FALSE;
+  const int selid = dt_control_get_mouse_over_id();
+  l = table->list;
+  while(l)
+  {
+    dt_thumbnail_t *thumb = (dt_thumbnail_t *)l->data;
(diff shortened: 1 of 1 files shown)


TurboGit (approved):
Works for me, thanks!

PR #6975, opened 2020-11-18
modulegroups : allow to remove from right-click too

AlicVB:
this enhance right-click on modulegroups tabs by allowing to add or *remove* modules.

### DIFF
--- a/data/themes/darktable.css
+++ b/data/themes/darktable.css
@@ -1329,7 +1329,8 @@ margin: 0 2px;
 #modulegroups-header #dt-button,
 #modulegroups-preset #dt-button,
 #modulegroups-preset-activated #dt-button,
-#modulegroups-iop-popup-name
+#modulegroups-iop-popup-name,
+#modulegroups-iop-popup-cb
 {
   border: 0;
 }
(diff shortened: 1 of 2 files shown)


aurelienpierre:
For me, this could go in 3.4 to finish the migration from modulelist.c to modulegroups.c with no loss of usability.

johnny-bit:
Well, It can go into 3.4.1 :) it's imo necessary.

TurboGit:
@AlicVB : This PR needs some work to fix the conflicts. TIA.

AlicVB:
ah ah... I was just fighting with rebase and co when you write the msg... Should be done now ;)

TurboGit:
I don't see that the tick in front of the module name is useful. In fact you already have two sections, as soon as you click on a module in a section it is either added or removed. So anyway we don't see the tick state changing and since we already have two sections I just find this superfluous. Thoughts ?

AlicVB:
Agree, it's not mandatory. A tooltip like "click to remove" or click to add" should be sufficient for the discoverability...

AlicVB:
@TurboGit : done :)

TurboGit:
Perfect now! Merged into master and queued for 3.4.1.

Issue #6936, opened 2020-11-15
DT 3.3 git master on Windows: LUT3D module disappeared


Mark-64:
Last DT 3.3 git master, build on Windows 10 on a fully updated MSYS2.

LUT3D module is no longer there, not evean searching for it
![immagine](https://user-images.githubusercontent.com/43290988/99195576-de15be80-2786-11eb-946e-3b89dece76bd.png)

However in the build log, lut3d.c and lut3dgmic.cpp seem to be compiled normally.

Any clue ?

EDIT: On Ubuntu LUT3D is present

TurboGit:
The build on windows is probably missing the libgmic.

chhil:
I am on the latest build and latest msys, lut3d is available.

Mark-64:
Nope
![immagine](https://user-images.githubusercontent.com/43290988/99232259-fd960100-27f1-11eb-93e1-58800a3dbd56.png)

Also, cmake finds gmic and as I said the source files are compiled (they would not be if gmic is not found)

chhil:
Gmic package is optional, so it would still compile, I wondered if it broke, but doesn't seem to be the case.

TurboGit:
So nothing wrong just a missing optional dependency.

Mark-64:
@TurboGit sorry but for me the issue is not solved. Gmic is there, I even tried to reistall it. Cmake find it.
I build DT on the same system hundreds times amd lut3d was alway there.

I will try to bisect the commits. Can you please reopen the issue ?

TurboGit:
Maybe there is another dependency... Don't remember... @phweyland : can you remind us the needed dependency to have the Lut module ?

phweyland:
If I remember well lut3d module fails to initialize when the run time gmic library is not found. In case of windows, this library should be included in the dt package... 
I've still my former windows alive. I'll try to make a build there again.

phweyland:
Doesn't start well:

```
$ pacman -S mingw-w64-x86_64-gmic
error: target not found: mingw-w64-x86_64-gmic
```

chhil:
![image](https://user-images.githubusercontent.com/1119439/99287137-dd9d2680-285f-11eb-8075-c2ad2e7d1cc9.png)

phweyland:
Yes, my msys is very old, needs to update, key issues,... normal life...

phweyland:
No luck so far. Some warnings and errors I don't master... 

[...]

PR #6976, opened 2020-11-18
lut3d: fix dependencies with GMIC 9.0-3 on windows

phweyland:


phweyland:
Fixes #6936

TurboGit (approved):
Works for me, thanks !

cmake/modules/FindGMIC.cmake
```diff
@@ -41,22 +42,43 @@ if(GMIC_LIBRARY AND GMIC_INCLUDE_DIR)
...
-      if(GMIC_LIBRARY)
-        find_library(FFTW_LIBRARY NAMES fftw3)
+      find_library(FFTW_LIBRARY NAMES fftw3)
+      if(NOT FFTW_LIBRARY)
+        message(STATUS "missing GMIC dependency fftw3")
+      else()
         list(APPEND GMIC_LIBRARY ${FFTW_LIBRARY})
-      ENDIF(GMIC_LIBRARY)
+      endif(NOT FFTW_LIBRARY)
+      # workaround for msys2 gmic 9.0-3. Should be reviewed when gmic 9.3 is available
```
chhil, line 51: Can we fix the version in the comment to be 2.9.0-3 to avoid confusion?
This is what shows up when I query the package manager for gmic.
$ pacman -Ss gmic
mingw32/mingw-w64-i686-gmic 2.9.0-3
    A Full-Featured Open-Source Framework for Image Processing (mingw-w64)
mingw64/mingw-w64-x86_64-gmic 2.9.0-3 [installed]
    A Full-Featured Open-Source Framework for Image Processing (mingw-w64)

cmake/modules/FindGMIC.cmake
```diff
@@ -41,22 +42,43 @@ if(GMIC_LIBRARY AND GMIC_INCLUDE_DIR)
...
+        list(APPEND GMIC_LIBRARY ${OPENCV_VIDEOIO_LIBRARY})
+      endif(NOT OPENCV_CORE_LIBRARY OR NOT OPENCV_VIDEOIO_LIBRARY)
     endif(WIN32)
 
-    # Set the include dir variables and the libraries and let libfind_process do the rest.
-    # NOTE: Singular variables for this library, plural for libraries this lib depends on.
-    set(GMIC_PROCESS_INCLUDES ${GMIC_INCLUDE_DIR})
-    set(GMIC_PROCESS_LIBS ${GMIC_LIBRARY})
-    libfind_process(GMIC)
+    # workaround for msys2 gmic 9.0-3. Should be reviewed when gmic 9.3 is available
```
chhil, line 63: Same as above.

packaging/windows/BUILD.txt
```diff
@@ -19,6 +19,9 @@ A) Native compile using MSYS2:  How to make a darktable windows installer (64 bi
     Install optional libraries and dependencies:
   $ pacman -S mingw-w64-x86_64-gmic
 
+    Note: GMIC. Waiting for GMIC 9.3 you may need to rename libopencv_code4xx and libopencv_videoio4xx
```
chhil, line 22: Fix version.

cmake/modules/FindGMIC.cmake
phweyland: Good catch. I'll fix this.

todd-prior:
Sorry trying to follow all these back and forth on the issues....I built DT today and no 3DLUT module....I renamed the potentially problematic dll and still did not fix the issue. I believe all dependencies are in place ...what is the best way to find the issue , during the build or with a DT debug???

phweyland:
First, if you don't use gmz file there is an option not to use GMIC at build time in dt/DefineOptions,cmake.
Then you can check you have the lut3d module alive.

With USE_GMIC ON...
What does the build say about GMIC ?
What is the version of GMIC ? 
In darktable/bin, do you have libopencv*.dll files ? Which version ?

todd-prior:
Thanks Philippe,

I don’t have access to that machine for a while but your comments are noted and I will follow-up. My main PC was happy with just changing the dll names…..

PR #6977, opened 2020-11-18
modulegroups : no right-click for deprecated group

AlicVB:
@TurboGit : you mentioned that in your original review, but I completely forgot to fix that point... Sorry !

### DIFF
--- a/src/libs/modulegroups.c
+++ b/src/libs/modulegroups.c
@@ -1459,6 +1459,7 @@ static gboolean _manage_direct_popup(GtkWidget *widget, GdkEventButton *event, d
   if(event->type == GDK_BUTTON_PRESS && event->button == 3)
   {
     dt_lib_modulegroups_group_t *gr = (dt_lib_modulegroups_group_t *)g_object_get_data(G_OBJECT(widget), "group");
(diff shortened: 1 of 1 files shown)


Nilvus:
To avoid a PR from me for a little CSS hack on your work, maybe adding to that PR (or a specific commit @TurboGit) to improve padding in popup menu in modulegroups:

```
#modulegroups_iop_title,
#modulegroups-iop-popup-name
{
   padding: 0.2em 0.5em;
}
```

About a border suggested by @TurboGit, I've answered on PR #6970.

Nilvus:
Another thing I just thought and related to modulegroups presets. In all modules the preset order have been improved to first display users preset, then after a separator, display default darktable presets. This is not implemented in modulegroups presets. I consider that as an issue as it's not consistent. Let me know if I should create an issue for that or if that could be easily implemented to correct that issue in 3.4?

TurboGit:
> Let me know if I should create an issue for that or if that could be easily implemented to correct that issue in 3.4?

Yes, please would be nice to indeed do that for consistency.

TurboGit (approved):
Works for me, thanks!

Issue #6695, opened 2020-10-31
FR: Precise control over parametric mask sliders


elstoc:
On most bauhaus sliders in darktable it is possible to right-click on a slider and key a precise value, however this doesn't seem to be possible in the parametric mask controls. Such precise values can be easily set in the shadows by enabling log mode, however nothing similar is available in the highlights. It is possible to use the Ctrl modifier, which is ok as a workaround, but such changes are not shown in the displayed slider value, which continues to output a truncated/rounded integer for any value over 10.

I would like to be able to key precise values for these sliders using the standard bauhaus slider control, and for the displayed values to be shown with a consistent number of decimal places (as per other slider controls). Some change to the UI might be required to achieve this, perhaps similar to some of those discussed in #6407.

elstoc:
right-click alteration would still be very nice.

asdkant:
I just bumped into this problem, again.


What I'm trying to achieve right now is a "split" color calibration, where the split is defined by a parametric mask. I think there's an opportunity to make a smarter GUI to achieve this kind of thing, something that links two instances of the same module through a shared mask that gives the inverted version of that mask to the second module.

Still, more precise control (and an easier GUI to reproduce that precise control somewhere else) would be really nice.

asdkant:
Can confirm this is still not a feature but still something we would like to have.

Issue #6255, opened 2020-09-09
FR: bauhaus slider popup


elstoc:
Where a bauhaus slider is located near the bottom of the screen, the right-click popup is not entirely visible. This is particularly noticeable in the overexposed/clipping indicator setting

![Screenshot_2020-09-09_23-31-58](https://user-images.githubusercontent.com/9555491/92661601-de13c100-f2f4-11ea-8924-e3542b13a8e7.png)

but also occurs in iop modules

![Screenshot_2020-09-09_23-32-39](https://user-images.githubusercontent.com/9555491/92661617-e966ec80-f2f4-11ea-9eb9-5464844f61dc.png)

It would be great if darktable could detect this situation and have the popup appear above the slider instead of below it.

Nilvus:
Bauhaus are something that just need to be completely reworked (and CSS editable). This could be related to #2077

Nilvus:
@elstoc, just as a reminder for another bauhaus issue you posted some months ago (and still there!).

elstoc:
yeah - not one I know how to resolve though.

dterrahe:
The code that was added in #11378 to deal with a similar issue for combos could be put outside the switch to also apply to sliders.
https://github.com/darktable-org/darktable/blob/c31d6a5b2f08ea4f3a6de209d774322c79e7c28d/src/bauhaus/bauhaus.c#L423-L444
But it looks like there are some subtle interactions (in adjusted and saved mouse location) that need to be taken care of and with the testing involved and dealing with feedback, this is not something I care enough about to take on.

PR #6989, opened 2020-11-19
Improve modulegroups GUI

Nilvus:
This improves modulegroups GUI and also add a thin border on all popover menus (so presets, styles menu in darkroom, etc.) for consistency and better visibility as pointed by @TurboGit here: https://github.com/darktable-org/darktable/pull/6970#issuecomment-729970054

@AlicVB: as this is on top of your work, your review is welcomed!

### DIFF
--- a/data/themes/darktable.css
+++ b/data/themes/darktable.css
@@ -826,6 +826,28 @@ dialog .dialog-vbox notebook stack box box checkbutton label  /* fix a margin on
     padding: 2px;
 }
 
+/*--------------------------------------------------------
+  - Bauhaus controls (sliders and comboboxes in modules) -
+  --------------------------------------------------------*/
+
+#bauhaus-popup
+{
+  color: @bauhaus_fg;
+}
+
+#bauhaus-popup:selected
+{
+  color: @bauhaus_fg_selected;
+}
+
+#bauhaus-combobox,
+#bauhaus-slider
+{
+  min-height: 1em;
+  background-color: transparent;
+  color: @bauhaus_fg;
+}
+
 /*----------------------------------------
   - Context menu & tooltips & comboboxes -
   ----------------------------------------*/
@@ -867,8 +889,8 @@ tooltip,
 popover,
 combobox window,
 dialog combobox window,
-#modulegroups-iop-popup,
-#modulegroups-iop-popup viewport
+#bauhaus-popup,
+#modulegroups-iop-popup
 {
   opacity: 1;  /* this is needed for tooltip on/off feature with Shift+t shortcut */
   background-color: @tooltip_bg_color;
@@ -878,6 +900,11 @@ dialog combobox window,
   min-height: 0.4em;
   min-width: 0.4em;
   padding: 0.2em 0;
+}
+
+#modulegroups-iop-popup viewport
+{
+  background-color: @tooltip_bg_color;
 }
 
 popover
@@ -1018,29 +1045,6 @@ notebook tab label /* disabled tabs */
 notebook header /* add space between notebook tabs and options below */
 {
   margin-bottom: 6px;
-}
-
-/*--------------------------------------------------------
-  - Bauhaus controls (sliders and comboboxes in modules) -
-  --------------------------------------------------------*/
-
-#bauhaus-popup
-{
-  background-color: @really_dark_bg_color;
-  color: @bauhaus_fg;
-}
-
-#bauhaus-popup:selected
-{
-  color: @bauhaus_fg_selected;
-}
-
-#bauhaus-combobox,
-#bauhaus-slider
-{
-  min-height: 1em;
-  background-color: transparent;
-  color: @bauhaus_fg;
 }
 
 /*--------------------------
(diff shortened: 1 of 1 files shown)


TurboGit:
Lot better to me ! Yet I think we can improve this. At the moment we have 3 different colors for menus. Let's take example on the dark theme:

The menu on module grroups:

image

The menu at the bottom of the lighttable & darkrroom which darker (softproofing for example):

image

And the combo-boxes in all dt which is even more darker:

image

If we could make this uniform it would be better. What do you think ?

Nilvus:
Good catch, I didn't check that. I will see and make all that consistent with that PR. Thanks for pointing that.

Nilvus:
@TurboGit: just a thing, not really related to that PR. Yesterday, I had received a mail notification on that PR about a Travis CI fail for a MacOS X compile. I've just seen in commits here on Github that compile fails for that same reasons on last commits since your commit 2bdb831887f4c5ec488282a11552e03c7bbc8e5a. Maybe something to check.

TurboGit:
I'll look but I doubt my commit is the culprit as it adds only an if.

Nilvus:
> I'll look but I doubt my commit is the culprit as it adds only an if.

It's just the first I have seen with the red cross of Travis build fails in commit.

Nilvus:
@TurboGit: background color on modulegroups popover was overriden by viewport children which had background-color from * first setting in darktable.css (line 163) so I have add it now to fix different color you see (and this glitch was only on dark and grey themes). This should be correct now.

About darker background you see, as a reminder, all completely dark background are bauhaus widget and we can't set them with CSS. All bauhaus widget and combobox would need a complete rewrite to allow setting background. Setting in CSS to allow add background-color is already possible but doesn't have any effect. The issue is on how bauhaus widget are written and this has already been discussed a long time ago. That would be a really good thing to see them rewritten for a future darktable release. Other bauhaus widget, with so the same black background, are popover displayed to change a setting in a slider with the right-click.

See issues about that: #2077 and #6255 and #6695? They all show that those bauhaus are just some piece of shit and needs to be rewritten. Probably the last thing in the UI to really improve.

Said that, this PR is so ready to merge for me.

TurboGit:
> About darker background you see, as a reminder, all completely dark background are bauhaus widget and we can't set them with CSS. All bauhaus widget and combobox would need a complete rewrite to allow setting background. Setting in CSS to allow add background-color is already possible but doesn't have any effect.

Actually it is possible using the #bauhaus-popup rule in darktable.css. There is no more hard coded color in bauhaus at this point.

Nilvus:
> Actually it is possible using the #bauhaus-popup rule in darktable.css. There is no more hard coded color in bauhaus at this point.

Oh, didn't know that one could allow that. Thanks for pointing that. Last commit update that and make so all three popover menus same color. So, only right-click popover on sliders remains black now but this really can't be changed by now.

So, should be ready to merge now.

data/themes/darktable.css
```diff
@@ -866,12 +889,15 @@ menuitem arrow,
 tooltip,
 popover,
 combobox window,
-dialog combobox window
+dialog combobox window,
+popover #bauhaus-popup,
```
TurboGit, line 893: Shouldn't this be only #bauhaus-popup ? And popover removed ? Otherwise all the standard combos in modules still have a black background, see for example in exposure module:

image

And note that the #bauhaus-popup also control the right-click menu over the sliders. So we can really makes all popup looks alike.

Two other things:

The menu on the module groups seems to have a double border, one around the window enclosing the scrollbar and on inner border around the content:

image

And the menu on bottom of darkrrom, lighttable seems to have a black border on top and bottom:

image

TurboGit (changes requested):
Better but still some room of improvement. TIA.

data/themes/darktable.css
Nilvus: As I said, it's not possible to apply a background color on right-click popover in sliders. I've tried and if I remove popover on that line, the only thing I add is that background-color of sliders change, but not right-click popover background. And that makes just modules awful. So popover is really needed. The only way to add background-color would be to update Gtk code to allow that and have a specific tag for only the right-click popover menu on sliders (and avoid to update sliders background the same way. So not CSS here.

About black border, I have seen that but I don't know where they are. You could see that I add that to same settings and borders than all other menus. And those others menus have a thin clearer border all around. And that is not applied there even if it's set. I agree, some improvement needed but not with CSS here. It's some piece of Gtk code to update to allow border setting to be applied here.

Last thing remaining and yes probably CSS, it's the double border. Will check that.

Nilvus: Forget what I've said just above about right-click in sliders. I made a new test and don't understand what I've done before. It works indeed great without the popover menu. Anyway, applying borders in bauhaus widget is impossible. I see them applied with Gtk inspector tool, so CSS is ok but they are not applied. Something in Gtk code.

Those bauhaus have been improved but they remains awful to work on.

Nilvus:
@TurboGit: last commit resolve your comment, except so borders in bauhaus popover. This needs surely to work on Gtk code and indeed, it would be good to have border I set (so same as other popover) applied to a better UI and a more consistent one. If you have time to check that.

TurboGit:
Lot better indeed now ! I'll merge that and will try to find out for the black top-bottom borders which are not looking good. Thanks @Nilvus !

PR #6996, opened 2020-11-20
interpolation: fix possible access to unaligned memory.

TurboGit:


chhil:
Did not fix the seg fault :(

PR #6997, opened 2020-11-20
[Issue #6987]Fix language detection when several locales are installed

### LINKED ISSUE #6987: Fail to select the appropriate default language
On my computer when I start darktable from /opt, the command g_get_language_names is returning all installed locales but sorted such that the selected one is appearing first ([as described in documentation](https://developer.gnome.org/glib/stable/glib-I18N.html#g-get-language-names)). 
The problem in my case is that it recognized then the wrong default language, and it is then not possible to use this language.

I have already a patch for this which has to be reviewed.

adrien-bellanger:
fix #6987  
It is my first PR for darktable, I think that I had follows the guidelines.

### DIFF
--- a/src/common/l10n.c
+++ b/src/common/l10n.c
@@ -332,6 +332,9 @@ dt_l10n_t *dt_l10n_init(gboolean init_list)
 
     const gchar * const * default_languages = g_get_language_names();
 
+    // For some directories will g_get_language_names retrieve a sorted list of lanquages
+    int current_sys_default_index = INT_MAX;
+
     char localedir[PATH_MAX] = { 0 };
     dt_loc_get_localedir(localedir, sizeof(localedir));
     GDir *dir = g_dir_open(localedir, 0, NULL);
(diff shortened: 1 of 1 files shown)


Nilvus:
Just a thing good to know, if you write on first post something like "fix #number-issue", that will link the issue (you could check that on right menu of the PR here) and so close automatically the issue when the PR will be merged. See that article to better understand how this work (and other possible words to do that): https://docs.github.com/en/free-pro-team@latest/github/managing-your-work-on-github/linking-a-pull-request-to-an-issue

adrien-bellanger:
I don't understand why the macOS check is failing. 
It fails with the following errors, which I think are not related to my changes.
```
/Applications/Xcode_12.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.15.sdk/System/Library/Frameworks/Foundation.framework/Headers/NSObjCRuntime.h:492:1: error: expected unqualified-id
@class NSString, Protocol;
^
```
Has someone an idea?

Nilvus:
This fail is not related to your PR. The same fails is present in all recent commits since yesterday as you can see on commits list: https://github.com/darktable-org/darktable/commits/master (I have asked @TurboGit to check that).

TurboGit:
And I had a look at the failure... But it does not makes sense to me, the failure seems related to MacOS and the reported error is strange... So for now let's forget about this.

@johnny-bit : Any idea about the MacOS failure since some days ?

johnny-bit:
well, nothing changed in osx on our side, that's for sure ;) it maybe due to xcode being not very nice and the debug would have to be a mac-only. so either I can play on CI and figure out what's wrong and/or @parafin or somebody with recent mac can chime in on what's going on ;)

parafin:
It's cmake update it seems - it now forces .mm to be compiled as C++ for some reason, while it should be ObjC++. I will update my environment locally and push a fix.

TurboGit:
Thanks a lot @parafin !

adrien-bellanger:
@parafin has fix the macOS bug yesterday, but I have no idea how I can run the checks one more time. Is it even possible?

parafin:
I've pressed the button to re-run the jobs, let's see how it goes.

parafin:
Seems to work.

adrien-bellanger:
I am not able to reproduce the bug anymore.
I don't think that it is already solved, but has not to be fixed yet.

PR #6922, opened 2020-11-15
histogram: brighten primaries

dtorop:
These colors are closer in saturation to the appearance prior to #6135. This is as suggested by @Nilvus, see https://github.com/darktable-org/darktable/pull/6135#issuecomment-727248993.

Before #6135:

image

With this PR:

image

After #6135:

image

PR #6999, opened 2020-11-21
draw histogram widget using CSS colors in display colorspace

### LINKED ISSUE #6945: [Histogram] odd colors

dtorop:
This walks back some of #6135. The intent is to fix #6945, and generally simplify the code.

This adds CSS colors `graph_red`, `graph_green`, and `graph_blue` to specify to primaries with which to draw the histogram, waveform, and RGB parade. The intent is to choose color which will be legible in a range of display gamuts (and white points), ranging from sRGB to Rec.2020.

The CSS colors replace prior code which used the display profile to calculate these primaries from hard-coded Lab values. These colors would appear pink when overlaid in a screen with a cooler white point (e.g. D65).

Other code from #6135 remains, including the fast calculation/drawing of the waveform histogram thanks to using darktable's fast internal colorspace conversion code.

Here is how the regular histogram looks with this code and the current CSS:

image

This is the waveform histogram:

image

For comparison see the screenshots in #6922. The primaries with this PR are much closer to the "classic" darktable primaries. It should be possible to soften out the display primaries to make them closer to the primaries from that PR.

### DIFF
--- a/data/themes/darktable.css
+++ b/data/themes/darktable.css
@@ -156,6 +156,13 @@
 @define-color graph_grid @grey_05;
 @define-color graph_overlay @grey_50;
 @define-color inset_histogram alpha(@grey_65, 0.50);
+/* Primaries selected for visual legibility across a variety of
+   gamuts, and to combine to pleasing secondaries. The combination of
+   all three must either to either the same number or >= 255 to
+   produce a neutral overlap. */
+@define-color graph_red rgb(237,30,20);
+@define-color graph_green rgb(28,235,26);
+@define-color graph_blue rgb(14,14,233);
 
 /* Reset GTK defaults - Otherwise dt inherits Adwaita default theme dark */
 *
--- a/src/bauhaus/bauhaus.c
+++ b/src/bauhaus/bauhaus.c
@@ -536,6 +536,9 @@ void dt_bauhaus_load_theme()
   gtk_style_context_lookup_color(ctx, "graph_fg_active", &darktable.bauhaus->graph_fg_active);
   gtk_style_context_lookup_color(ctx, "graph_overlay", &darktable.bauhaus->graph_overlay);
   gtk_style_context_lookup_color(ctx, "inset_histogram", &darktable.bauhaus->inset_histogram);
+  gtk_style_context_lookup_color(ctx, "graph_red", &darktable.bauhaus->graph_primaries[0]);
+  gtk_style_context_lookup_color(ctx, "graph_green", &darktable.bauhaus->graph_primaries[1]);
+  gtk_style_context_lookup_color(ctx, "graph_blue", &darktable.bauhaus->graph_primaries[2]);
 
   PangoFontDescription *pfont = 0;
   gtk_style_context_get(ctx, GTK_STATE_FLAG_NORMAL, "font", &pfont, NULL);
--- a/src/bauhaus/bauhaus.h
+++ b/src/bauhaus/bauhaus.h
@@ -232,6 +232,7 @@ typedef struct dt_bauhaus_t
 
   // colors for graphs
   GdkRGBA graph_bg, graph_border, graph_fg, graph_grid, graph_fg_active, graph_overlay, inset_histogram;
+  GdkRGBA graph_primaries[3];
 } dt_bauhaus_t;
 
 #define DT_BAUHAUS_SPACE 0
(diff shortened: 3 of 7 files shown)


Nilvus:
Just tested and really much betters. Colors are just great, not over-saturated like legacy histogram and no odd colors. Neutral colors and with the possibility for the user to adjust more easily colors with CSS, it seems to be the good one now for improving the histogram. Actual histogram needs to be updated with that for 3.4. On my side, white is white and I don't see colors issue. Thanks for that work.

aurelienpierre:
Now this sucks because CSS stylesheets are supposed to always use sRGB color space to define colors.

AxelG-DE:
I donno what happened to my comment... I have tested and looks good on my end with my EIZO CS270 too (laptop still to be tested)

TurboGit:
Clearly a good improvement to me. The histogram looks OK on my laptop and AdobeRGB screen. Thanks !

dtorop:
@TurboGit @AxelG-DE @Nilvus Thank you for testing! Let's hope this settles things for the histogram for v3.4! Thanks for bearing with all the changes.

> CSS stylesheets are supposed to always use sRGB color space to define colors

@aurelienpierre I didn't know this. In this case, as with the rest of darktable, the CSS-specified colors are going straight to Cairo -> GDK -> GTK etc...

PR #7007, opened 2020-11-21
fix crash when right border of frame completely fills viewport

### LINKED ISSUE #6805: DT 3.3.0+1440 : Zoomed in and Retouch crashes DT
Retouched imaged, zoomed in to further retouch image.

Windows.
this is darktable **3.3.0+1440~g44fe67e70** reporting an exception:

-------------------

Error occurred on Monday, November 9, 2020 at 12:50:47.

darktable.exe caused an Access Violation at location 000000006E941AFC in module libborders.dll Writing to location 00000000971CE000.

```
AddrPC           Params
000000006E941AFC 0000000007F72700 00007FF9893D2EF8 0000000000000000  libborders.dll!set_outer_border_sse._omp_fn.0  [C:/msys64/mingw64/lib/gcc/x86_64-w64-mingw32/10.2.0/include/xmmintrin.h @ 980]
   978: _mm_store_ps (float *__P, __m128 __A)
   979: {
>  980:   *(__m128 *)__P = __A;
   981: }
   982: 
00007FF969D8DE30 00000000561730F0 0000000000000000 0000000000000000  libgomp-1.dll!omp_in_final
00007FF9893D4F33 0000000056075460 0000000000000000 00000000561730F0  libwinpthread-1.dll!pthread_create_wrapper
00007FF9AC53B04A 00007FF9AC5906D0 00000000561730F0 0000000000000000  msvcrt.dll!_beginthreadex
00007FF9AC53B11C 0000000000000000 0000000000000000 0000000000000000  msvcrt.dll!_endthreadex
00007FF9AAC87C24 0000000000000000 0000000000000000 0000000000000000  KERNEL32.DLL!BaseThreadInitThunk
00007FF9ACA8CEA1 0000000000000000 0000000000000000 0000000000000000  ntdll.dll!RtlUserThreadStart
```

ralfbrown:
Clamp cp_width to be non-negative.

If a frame has been enabled in borders.c, and is wide enough at the
current zoom level that the right border completely fills the center
display, borders.c:modify_roi_in returns an x offset greater than the
image width, which results in a negative value for the number of
pixels to copy in the blit loop.

May fix #6805, though the crash is in a completely different place (pixelpipe_hb.c instead of borders.c).

### DIFF
--- a/src/develop/pixelpipe_hb.c
+++ b/src/develop/pixelpipe_hb.c
@@ -1220,16 +1220,19 @@ static int dt_dev_pixelpipe_process_rec(dt_dev_pixelpipe_t *pipe, dt_develop_t *
           const int cp_width = MAX(0, MIN(roi_out->width, pipe->iwidth - in_x));
           const int cp_height = MIN(roi_out->height, pipe->iheight - in_y);
 
+          if (cp_width > 0)
+          {
 #ifdef _OPENMP
 #pragma omp parallel for default(none) \
           dt_omp_firstprivate(bpp, cp_height, cp_width, in_x, in_y) \
           shared(pipe, roi_out, roi_in, output) \
           schedule(static)
 #endif
-          for(int j = 0; j < cp_height; j++)
-            memcpy(((char *)*output) + (size_t)bpp * j * roi_out->width,
-                   ((char *)pipe->input) + (size_t)bpp * (in_x + (in_y + j) * pipe->iwidth),
-                   (size_t)bpp * cp_width);
+            for(int j = 0; j < cp_height; j++)
+              memcpy(((char *)*output) + (size_t)bpp * j * roi_out->width,
+                     ((char *)pipe->input) + (size_t)bpp * (in_x + (in_y + j) * pipe->iwidth),
+                     (size_t)bpp * cp_width);
+          }
         }
         else
         {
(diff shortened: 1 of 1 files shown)


TurboGit:
This scenario still crash for me, maybe cp_height needs the same clamping ?

TurboGit:
Even clamping cp_height I can reproduce the crash. So this needs more work I fear.

ralfbrown:
cp_height is only used as the termination condition for the following for loop, so a negative value causes the loop to be skipped. 
Interestingly, when I tried it just now, it crashed immediately on changing to 100% zoom with the entire border outside the viewport, and did so in set_outer_border_sse as reported in #6805.  #7009 fixes that crash (I hadn't properly handled the border size in my previous speedup), but this PR is still needed for the right-border crash in the pixelpipe code.

TurboGit:
@ralfbrown : Understood, I'll wait for the final fix before merging this PR.

ralfbrown:
AFAICT, the final fix is to merge both #7007 (this PR) and #7009 because there are in fact two separate crashing bugs which just happen to trigger under nearly identical circumstances.  While 7009 was primarily intended to optimize speed, it also replaces the problematic set_outer_border_sse.  I now know that that function can crash when the border is not in the viewport because it overruns the roi_out in drawing the border anyway.

TurboGit:
Ok, let's first merge this as a saftly fix for 3.4 and I'll review if #7007 is safe to go in 3.4, it only change borders.c so that's already a good point.

PR #7008, opened 2020-11-21
Update lua code example

wpferguson:
Copied the example and tested it.  darktable.control.execute returns the opposite of coroutine.yield (the previous implementation of the example) so the test needed to be reversed,


TurboGit (approved):
Thanks !

PR #7013, opened 2020-11-22
gui_cleanup before destroying multi-instance widget

dterrahe:
similar to #4096

Tone equaliser and color calibration save notebook page in gui_cleanup. When changing images a multi-instance would already have been destroyed by then.


TurboGit (approved):
Looks ok to me, thanks.

PR #7021, opened 2020-11-22
po/pref sections

TurboGit:
Add sections to darkroom & lighttable preferences tabs for readability. Both page where quite crowded, two sections have been introduced.

### DIFF
--- a/src/gui/preferences.c
+++ b/src/gui/preferences.c
@@ -551,7 +551,7 @@ void dt_gui_preferences_show()
   g_signal_connect(G_OBJECT(_preferences_dialog), "delete-event", G_CALLBACK(preferences_window_deleted), NULL);
 #endif
 
-  gtk_window_set_default_size(GTK_WINDOW(_preferences_dialog), DT_PIXEL_APPLY_DPI(1100), DT_PIXEL_APPLY_DPI(700));
+  gtk_window_set_default_size(GTK_WINDOW(_preferences_dialog), DT_PIXEL_APPLY_DPI(1100), DT_PIXEL_APPLY_DPI(750));
 #ifdef GDK_WINDOWING_QUARTZ
   dt_osx_disallow_fullscreen(_preferences_dialog);
 #endif
@@ -573,7 +573,7 @@ void dt_gui_preferences_show()
   GtkWidget *stack = gtk_stack_new();
   GtkWidget *stacksidebar = gtk_stack_sidebar_new();
   gtk_stack_sidebar_set_stack(GTK_STACK_SIDEBAR(stacksidebar), GTK_STACK(stack));
-  gtk_widget_set_size_request(stack, DT_PIXEL_APPLY_DPI(900), DT_PIXEL_APPLY_DPI(700));
+  gtk_widget_set_size_request(stack, DT_PIXEL_APPLY_DPI(900), DT_PIXEL_APPLY_DPI(750));
   gtk_box_pack_start(GTK_BOX(box), stacksidebar, FALSE, FALSE, 0);
   gtk_box_pack_start(GTK_BOX(box), stack, TRUE, TRUE, 0);
 
--- a/data/darktableconfig.dtd
+++ b/data/darktableconfig.dtd
@@ -2,7 +2,7 @@
 <!ELEMENT dtconfig (name,type,default,capability?,shortdescription?,longdescription?)>
 <!ATTLIST dtconfig
 	prefs (import|lighttable|darkroom|otherviews|processing|security|cpugpu|storage|misc) #IMPLIED
-        section (import|session|interface|other|tags|geoloc|slideshow|cpugpu|database|xmp|accel|general|modules) #IMPLIED
+        section (import|session|interface|other|tags|geoloc|slideshow|cpugpu|database|xmp|accel|general|modules|thumbs) #IMPLIED
         restart (true|false) #IMPLIED
         capability CDATA #IMPLIED
 >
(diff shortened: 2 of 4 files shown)


TurboGit:
This is between an enhancement and a bug fix and makes the prefs more readable, a plus for the 3.4 as very safe (only touching generated code which has been already used and tested). That's why I propose this for 3.4.

elstoc:
The only issue I see is that we're getting close to having scrolling tabs and this PR possibly takes us a bit closer. I see you've increased the default size of the window but, including any window-manager decorations I suppose it's possible that might take the height close to and even over some people's max screen resolution (for people on laptops). An alternative might be to split each of the darkroom/lighttable tabs into two tabs (rather than tabs split into sections). Just some possible alternatives to consider - feel free to disregard.

src/gui/preferences.c
```diff
@@ -551,7 +551,7 @@ void dt_gui_preferences_show()
   g_signal_connect(G_OBJECT(_preferences_dialog), "delete-event", G_CALLBACK(preferences_window_deleted), NULL);
 #endif
 
-  gtk_window_set_default_size(GTK_WINDOW(_preferences_dialog), DT_PIXEL_APPLY_DPI(1100), DT_PIXEL_APPLY_DPI(700));
+  gtk_window_set_default_size(GTK_WINDOW(_preferences_dialog), DT_PIXEL_APPLY_DPI(1100), DT_PIXEL_APPLY_DPI(750));
```
Nilvus, line 554: I suggest to increase to 765

src/gui/preferences.c
```diff
@@ -573,7 +573,7 @@ void dt_gui_preferences_show()
   GtkWidget *stack = gtk_stack_new();
   GtkWidget *stacksidebar = gtk_stack_sidebar_new();
   gtk_stack_sidebar_set_stack(GTK_STACK_SIDEBAR(stacksidebar), GTK_STACK(stack));
-  gtk_widget_set_size_request(stack, DT_PIXEL_APPLY_DPI(900), DT_PIXEL_APPLY_DPI(700));
+  gtk_widget_set_size_request(stack, DT_PIXEL_APPLY_DPI(900), DT_PIXEL_APPLY_DPI(750));
```
Nilvus, line 576: So same here.

Nilvus (commented):
I agree with @elstoc: that's a risk. But on my side, with a grey theme (so Roboto font) with a font size to 11, the height is only 784 (including window bar from my desktop environment), so all good. But If I change font size to 13, I have a scrollbar (for about least than a line) in lighttable tab but not on darkroom one (but it's really on the limit). 

Anyway, it's better with those sections. I remember someone recently in another PR (I think it was @jenshannoschwalm but my memory could be wrong) saying that we are now maybe just too many options. And I agree with that. And at least, we should be careful to not add too many prefs in the future. It's good to have something really configurable but we need to be careful that prefs do not become a nightmare to go through.

Notebook tabs could be another way but I'm quite worried that it could be worse and overload the prefs UI.

As it's only add a really little scrollbar on my side with 13 font size, and probably really few users will use more, increasing as suggested window height should be good by now. And for me, it will be good as it is and better than in actual master.

Nilvus:
It's just a comment and suggestion, by using it more, having a scrollbar is not bad. So, it's as you want, increasing more the window will make it quite big. So actual size in that PR is good and increasing too. It's only for what I test on my side (different sizes and fonts) with 13 font size that scrollbars appears. And probably, many users use 11 or 12 font size.

elstoc:
One option would be to move the settings that only apply to a single module onto the module itself. For example, the color balance and white balance options could be a settings popup in the respective modules. Regarding the height I'm sure I remember someone commenting that they had a 768px high screen and that was why I set it to 700, but I can't find that comment now.

Nilvus:
> One option would be to move the settings that only apply to a single module onto the module itself. For example, the color balance and white balance options could be a settings popup in the respective modules. Regarding the height I'm sure I remember someone commenting that they had a 768px high screen and that was why I set it to 700, but I can't find that comment now.

I like the idea. And we already have "modules" with specific settings in lighttable like metadata editor and export ones. So it would be better. Anyway, as we are on feature freeze, it's not the good time to do so. I suggest by now and for 3.4 to merge that PR as it is (so with 750 height and forget my idea to increase it) and see for 3.4.1 what lines could be moved and how. I already see 2 lines on lighttable that could easily go in star overlays menu and 2 lines in darkroom for respectively white balance and color balance.

Maybe we should open an issue to continue that? @elstoc, @TurboGit: what do you think of all that?

TurboGit:
Ok, I'll update the preference height tomorrow and merge this for 3.4.

We can then rework whatever we want for next release but at this stage more than what is proposed here is not wanted so close of the release.

> Maybe we should open an issue to continue that? @elstoc, @TurboGit: what do you think of all that?

Yes good idea.

elstoc:
Agreed - we definitely don't want to be making those sorts of changes now. I'm happy for this PR to go in as is and we can work on preference reduction later.

Nilvus:
> Ok, I'll update the preference height tomorrow and merge this for 3.4.

I hope you don't talk about increasing height. As I said, forget my suggestion to increase. This PR is good as it is and discussion since my proposal change it. Or we let 750 height or we come back to 700. Personally, I don't care.

> Yes good idea.

So, I will post the issue proposal tomorrow too.

TurboGit:
@Nilvus : Ok, I misunderstood, so no change and merging this PR.

PR #7024, opened 2020-11-22
modules : use deprecated message as tooltip if needed

AlicVB:
for consistency with other modules.


TurboGit (approved):
Yep, better this way!

Issue #6938, opened 2020-11-15
image colorspace problem in master


ptilopteri:
openSUSE Tumbleweed 20201114
NVIDIA GF106 [GeForce GTS 450], 390.138
darktable 3.3.0+1639~ga70022f88
OpenCL on
i7 12-core 36GB
kernel 5.9.1-2-default

when opening jpg files from samsung galaxy 10+ from
control line:
  darktable --library :memory: /home/paka/Pictures/2020/S10+PhoneCamera/

get follwoing as thumbnails are displayed:

[dt_ioppr_transform_image_colorspace_cl] invalid conversion from 2 to 0
[_transform_matrix] invalid conversion from 2 to 0
[dt_ioppr_transform_image_colorspace] invalid conversion from 2 to 0
[dt_ioppr_transform_image_colorspace_cl] invalid conversion from 0 to 2
[_transform_matrix] invalid conversion from 0 to 2

TurboGit:
Can you double check that there is no white-balance module applied?

ptilopteri:
fresh single jpg imported provides:

[dt_ioppr_transform_image_colorspace_cl] invalid conversion from 2 to 0
[_transform_matrix] invalid conversion from 2 to 0
[dt_ioppr_transform_image_colorspace] invalid conversion from 2 to 0
[dt_ioppr_transform_image_colorspace_cl] invalid conversion from 0 to 2
[_transform_matrix] invalid conversion from 0 to 2
[dt_ioppr_transform_image_colorspace] invalid conversion from 0 to 2
[_transform_matrix] invalid conversion from 2 to 0
[dt_ioppr_transform_image_colorspace] invalid conversion from 2 to 0
[_transform_matrix] invalid conversion from 0 to 2
[dt_ioppr_transform_image_colorspace] invalid conversion from 0 to 2


attached xmp file

[IMG_20201115_172246.jpg.xmp.txt]
(https://github.com/darktable-org/darktable/files/5546708/IMG_20201115_172246.jpg.xmp.txt)

appears no white-balance auto applied, and I did not apply one.
tks

aurelienpierre:
fixed by #7029 it seems

ptilopteri:
tks, I no longer see the notices

Issue #6666, opened 2020-10-28
PFM imports as linear but sRGB is shown for colorin module


kmilos:
**To Reproduce**
1. Download PFM and EXR samples at http://www.pauldebevec.com/Research/HDR/PFM/
2. Import the EXR sample and verify the input profile is "embedded matrix" 
3. Import the PFM sample and verify the input profile shown is "sRGB"
4. Re-select/re-activate the "sRGB" input color profile
5. The image tone changes dramatically
6. Select "linear Rec709 RGB" input color profile instead
7. The image is restored to the initially imported state (and matches the EXR)

**Expected behavior**
colorin should show the actual profile that was used after import (i.e. "linear (709)" fallback in the PFM case?)

**Platform (please complete the following information):**
 - darktable version: 3.3.0+1248~g479426b49
 - OS: Windows 10

kmilos:
Still the case after all the gui init accel changes in 3.3.0+1330~gdcc2032de

kmilos:
This was OK btw in 3.2.1, colorin shows linear 709 on import...

aurelienpierre:
Notice that PFM is a technical exchange format, as far as I know, it doesn't store any metadata, so the color profile is not inside it and needs to be set by user.

kmilos:
Oh, no argument there! Not expecting wonders with no metadata around ;)

The bug is that the linear 709 fallback (perfectly valid starting point) is used by the pipe to provide the initial render (set around [here](https://github.com/darktable-org/darktable/blob/024988292eb32d1ef2ea354ac1d9fd51f23e6e55/src/iop/colorin.c#L1600) I presume?), but colorin GUI combobox was out of sync and reflecting sRGB...

Btw, I tested your #7029, seems unrelated to this after all (log is attatched.)

[debug_pfm_import.zip](https://github.com/darktable-org/darktable/files/5592976/debug_pfm_import.zip)

kmilos:
Re the proposed fix (I'll test and feedback in a bit): there's really no better way of keeping the pipe and the GUI in sync? I see quite a few strcmp(ext, ...) around there (not just PFM related), it just doesn't feel "nice"...

aurelienpierre:
No I don't see another way, we will have to test every file extension, one way or another.

kmilos:
Ok, merci

PR #7029, opened 2020-11-23
Fix default params loading

aurelienpierre:
There is an annoying bug which affects TIFF, PNG and JPEG when opening them in darkroom that prevents the embedded color profile from being correctly loaded. As a result, images look to bright because the gamma is not reverted. It disappear when resetting the colorin module. It reappears when discarding the history. It doesn't affect images for which the profile is set by user (even if the setting is merely "embedded profile"). I think @jenshannoschwalm had a shot at it, maybe fixed it for a while, then it reappeared. It's likely linked to #6938.

To help debugging, in a7cdeb2, I make sure the parameters GUI reset uses `reload_defaults` too (`src/develop/imageop.c:1864`). Now that the code path is the same, it feels like a synchronisation issue between `reload_defaults` and `commit_params` at the pipeline level, at init time.

Searching around the `commit_params` calls, I discover that these functions are actually called multiple times per pipeline. In fact, in `src/develop/imageop.c:607`, there is even a `/// FIXME: Commit params is already done in module` from 7fac531 (2013) that says so. At the library level, in `imageop.c`, `commit_params` is called in `init_pipe` general API.

So, 253f122 removes the `commit_params` calls duplicated in the individual IOP `init_pipe`. Then, the `commit_params` are only ever called in the `init_pipe`. But this function has actually only one call in the whole program : when initing the pipe nodes, `src/develop/pixelpipe_hb.c:328`. So 2dfa677 moves the `commit_params` at node init, and adds a `reload_defaults` before to be sure we init modules in the correct order (built defaults, init pipe, commit). So we can remove the hacky call to `reload_defaults` when loading the .so of the IOP, `src/develop/imageop.c:1667`, because it's really not clean to try call it where we possibly have no pipeline inited yet (thus no image to init it upon).

So, now we know that `commit_param` should be called only from one place, when initing pipe nodes, and we are sure that `reload_params` is executed before, and also we know that when nodes are inited, there is an image in the pipeline. So things should work. 

Noooope.

To figure out what happens, I add debugging info in `colorin.c`, that output the profile type at the end of `reload_params` and at the beginning of `commit_params`. And this is what I get when opening a brand new JPEG in darkroom:
```
[colorin] profile type in defaults_params for pipe 0: 9
[colorin] profile type in commit_params for pipe 2: 9
[colorin] profile type in commit_params for pipe 2: 9
[colorin] profile type in commit_params for pipe 2: 11
[colorin] profile type in defaults_params for pipe 0: 9
[colorin] profile type in commit_params for pipe 4: 9
[colorin] profile type in commit_params for pipe 4: 9
[colorin] profile type in commit_params for pipe 4: 11
```
It's an sRGB JPEG with embedded profile, so we expect a type 9 (`DT_COLORSPACE_EMBEDDED_ICC`). Type 11 is `DT_COLORSPACE_STANDARD_MATRIX`, aka the default when `reload_params` has exhausted all its options. We see that `default_params` properly get the correct profile. But the resulting image is too bright, meaning the correct profile is not applied. What I can't explain is why we have 3 calls to `commit_params`, the 2 first using the correct profile, and the third using something else. I don't understand how the parameter is changed, nor where the calls come from.

I need help, and I think we better have the proper profile loaded for 3.4.

### DIFF
--- a/src/iop/colorchecker.c
+++ b/src/iop/colorchecker.c
@@ -834,7 +834,6 @@ void commit_params(struct dt_iop_module_t *self, dt_iop_params_t *p1, dt_dev_pix
 void init_pipe(struct dt_iop_module_t *self, dt_dev_pixelpipe_t *pipe, dt_dev_pixelpipe_iop_t *piece)
 {
   piece->data = malloc(sizeof(dt_iop_colorchecker_data_t));
-  self->commit_params(self, self->default_params, pipe, piece);
 }
 
 void cleanup_pipe(struct dt_iop_module_t *self, dt_dev_pixelpipe_t *pipe, dt_dev_pixelpipe_iop_t *piece)
--- a/src/develop/develop.c
+++ b/src/develop/develop.c
@@ -633,6 +633,18 @@ restart:
     DT_DEBUG_CONTROL_SIGNAL_RAISE(darktable.signals, DT_SIGNAL_DEVELOP_UI_PIPE_FINISHED);
 }
 
+
+static inline void _dt_dev_load_pipeline_defaults(dt_develop_t *dev)
+{
+  GList *modules = g_list_last(dev->iop);
+  while(modules)
+  {
+    dt_iop_module_t *module = (dt_iop_module_t *)(modules->data);
+    dt_iop_reload_defaults(module);
+    modules = g_list_previous(modules);
+  }
+}
+
 // load the raw and get the new image struct, blocking in gui thread
 static inline void _dt_dev_load_raw(dt_develop_t *dev, const uint32_t imgid)
 {
@@ -1684,6 +1696,9 @@ void dt_dev_read_history_ext(dt_develop_t *dev, const int imgid, gboolean no_ima
     // cleanup
     DT_DEBUG_SQLITE3_EXEC(dt_database_get(darktable.db), "DELETE FROM memory.history", NULL, NULL, NULL);
 
+    // make sure all modules default params are loaded to init history
+    _dt_dev_load_pipeline_defaults(dev);
+
     // prepend all default modules to memory.history
     _dev_add_default_modules(dev, imgid);
     const int default_modules = _dev_get_module_nb_records();
@@ -1691,6 +1706,7 @@ void dt_dev_read_history_ext(dt_develop_t *dev, const int imgid, gboolean no_ima
     // maybe add auto-presets to memory.history
     first_run = _dev_auto_apply_presets(dev);
     auto_apply_modules = _dev_get_module_nb_records() - default_modules;
+
     // now merge memory.history into main.history
     _dev_merge_history(dev, imgid);
 
--- a/src/develop/develop.c
+++ b/src/develop/develop.c
@@ -1685,12 +1685,12 @@ void dt_dev_read_history_ext(dt_develop_t *dev, const int imgid, gboolean no_ima
                             dt_ioppr_iop_order_copy_deep(darktable.develop->iop_order_list));
 
   int history_end_current = 0;
-  sqlite3_stmt *stmt;
+  int auto_apply_modules = 0;
+  gboolean first_run = FALSE;
+  gboolean legacy_params = FALSE;
 
   dt_ioppr_set_default_iop_order(dev, imgid);
 
-  int auto_apply_modules = 0;
-  gboolean first_run = FALSE;
   if(!no_image)
   {
     // cleanup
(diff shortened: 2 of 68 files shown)


aurelienpierre:
Ok, I found it. In `pixelpipe_hb.c:363`, 
```
void dt_dev_pixelpipe_synch_all(dt_dev_pixelpipe_t *pipe, dt_develop_t *dev)
{
  dt_pthread_mutex_lock(&pipe->busy_mutex);
  // call reset_params on all pieces first.
  GList *nodes = pipe->nodes;
  while(nodes)
  {
    dt_dev_pixelpipe_iop_t *piece = (dt_dev_pixelpipe_iop_t *)nodes->data;
    piece->hash = 0;
    piece->enabled = piece->module->default_enabled;
    dt_iop_commit_params(piece->module, piece->module->default_params, piece->module->default_blendop_params,
                         pipe, piece);
    nodes = g_list_next(nodes);
  }
  // go through all history items and adjust params
  GList *history = dev->history;
  for(int k = 0; k < dev->history_end && history; k++)
  {
    dt_dev_pixelpipe_synch(pipe, dev, history);
    history = g_list_next(history);
  }
  dt_pthread_mutex_unlock(&pipe->busy_mutex);
  fprintf(stdout, "[pixelpipe] synch all\n");
}
```

If I remove the block :
```
  // go through all history items and adjust params
  GList *history = dev->history;
  for(int k = 0; k < dev->history_end && history; k++)
  {
    dt_dev_pixelpipe_synch(pipe, dev, history);
    history = g_list_next(history);
  }
```
everything works as expected. It seems the history is either improperly inited or should not be read at the first run.

aurelienpierre:
Ok, found the son of a bitch.

In the order, when in image is loaded, it uses:
* `dt_dev_load_image()` from `develop.c:693`, 
  * which calls `dt_dev_read_history()` from `develop.c:1954` 
    * which calls `dt_dev_read_history_ext()` from `develop.c:1663`, 
      * which calls `_dev_add_default_modules()` from `develop.c: 1534` in a non-thread-safe way
        * which calls `_dev_insert_module()` from `develop.c:1354`,
          * which copies `module->default_params` into history DB.

But `dt_dev_load_image()` is called in `enter()` from `darkroom.c` and `dt_dev_read_history()` is called several times. And at that point, since there is no pipe, `module->default_params` may not be initialized.

TurboGit:
I think @dterrahe or @ralfbrown can help here as this is a related to the pixelpipe.

I'll try to have a look later.

kmilos:
Could this also be responsible for #6666 ?

aurelienpierre:
It may be linked to whatever involves initial loading of parameters, defaults and workflows. It's basically an initialization failure.

aurelienpierre:
The culprits are most likely `dt_dev_read_history_ext()`. Whether the with the 3rd argument is set to `TRUE` or `FALSE` does not seem to change the control.

aurelienpierre:
The latest commit shows that, for a thumbnail with an history, when the lighttable relaunch a pipe, the history is deleted and re-initialized before `reload_defaults` is even triggered:

```
[history] history deleted
[history] history reset to defaults
[colorin] profile type in defaults_params: 9
[pixelpipe] synch all modules with defaults_params for pipe 8
[colorin] profile type in commit_params for pipe 8: 9
[pixelpipe] synch all modules with history for pipe 8
[colorin] profile type in commit_params for pipe 8: 11
```

aurelienpierre:
The latest commit ensures the `reload_params` functions are properly called ahead of the history init. Nevertheless, the problem remains:

```
[history] history deleted
[colorin] profile type in defaults_params: 9
[history] module colorin inserted
[history] module colorout inserted
[history] module gamma inserted
[history] module flip inserted
[history] history reset to defaults
[colorin] profile type in defaults_params: 9
[colorin] could not find requested profile `standard color matrix'!
[pixelpipe] pipeline changed for pipe 2
[pixelpipe] synch all modules with defaults_params for pipe 2
[colorin] profile type in commit_params for pipe 2: 9
[pixelpipe] synch all modules with history for pipe 2
[colorin] profile type in commit_params for pipe 2: 11
[colorin] could not find requested profile `standard color matrix'!
```

I'm now blaming `_dev_insert_module()` which actually does the copy from `module->default_params` to SQL history.

TurboGit:
@aurelienpierre  : I'll try to have a look tomorrow, can you share a jpeg with embedded profile to be sure we are looking at the same thing ?

aurelienpierre:
@TurboGit any JPEG works, but here is a sample : 
20200906_Wien_1636_02

aurelienpierre:
So, after checking the whole control, `_dev_insert_module()` seems to write the correct data into a temporary SQL DB in memory : `memory.history`. What is actually commited into the history stack of modules is however a `main.images` DB. The function responsible for merging `memory.history` into `main.images` is `_dev_merge_history()` and that's where things may be convoluted.

I don't understand what happens there.

TurboGit:
Hum... now the real question.... What is not working ? On my side I have opened the image, I have no message on the console, the image looks correct and not too bright, the colorin module does list "profil ICC intégré":

image

Any idea ?

aurelienpierre:
Well, in my case the color profile is loaded wrong until reset by user.

Also, in your case, the reload params function starts by outputing 11 then switches to 9, while it should have been 9 since the beginning.

aurelienpierre:
For the record, the profile issue was on my side, with a colorin preset auto-applying the standard matrix instead of the enhanced one. Since the preset made no difference between raw and JPEGs, it tried to applied it on them too.

Nevertheless, this PR fixes many inconsistencies, superfluous calls and makes the pipe easier to debug.

src/iop/colorchecker.c
```diff
@@ -121,7 +121,7 @@ const char *aliases()
 
 const char *description(struct dt_iop_module_t *self)
 {
-  return dt_iop_set_description(self, _("perform color space corrections and apply looks."),
+  return dt_iop_set_description(self, _("perform color space corrections and apply looks"),
```
TurboGit, line 124: unwanted, we have dot on description on all others... maybe a wrong merge conflict resolution ?

aurelienpierre: yes, it's a wrong rebase

TurboGit:
Thanks !

PR #7033, opened 2020-11-23
darktableconfig: use consistent terminology

elstoc:
Alter terminology used in darktable preferences to match that used in the dtdocs user manual and shortcuts tab:

- "lib" modules are referred to as "utility modules"
- "iop" modules are referred to as "processing modules"
- [top, bottom and side panels are referred to as "panels"]


elstoc:
Really, the "utility" module preferences should be moved out of the lighttable tab entirely, since they apply to other views as well.

TurboGit (approved):
Thanks !

PR #7038, opened 2020-11-23
imageop: prevent always on/off modules being enabled/disabled

elstoc:
if the on/off button is hidden don't allow it to be toggled with a keyboard shortcut


elstoc:
Partial resolution for 7037.

TurboGit (approved):
Looks good, thanks!

PR #7045, opened 2020-11-24
adds wheremi and modifies existing entries to consistently contain th…

codingdave:
…e URL of the projects


TurboGit:
I'll add too the `--recurse-submodules` in the clone command just above this change.

codingdave:
Good idea @TurboGit. I have added that now also in a1c88b1. 4d862d6 combines init and update into one line.

TurboGit (approved):
Perfect, thanks !

PR #7047, opened 2020-11-24
[wb_presets] add Sony ILCE-6600 finetuned

johnny-bit:
New set of WB presets for Sony ILCE-6600 contributed by @PimPolacko

fixes #6983

PR #7048, opened 2020-11-24
[extract_wb] deal with possible sony half-steps

johnny-bit:
Based of of work done in #7047 it seems that newer sony ilce models offer "halfsteps", similarly to nikon (why oh why don't they just use normal number of steps instead of having "precise" halfsteps of exacty 0.5 is beyond me)


TurboGit (approved):
Looks ok, thanks.

Issue #7050, opened 2020-11-24
Improve and simplify preferences content


Nilvus:
**Is your feature request related to a problem? Please describe.**

darktable become more and more configurable on each release. Globally it's a good thing but we need to take care about not overload preferences. They're probably already a little bit. Saying that, it's hard to see which settings are surplus. Maybe all are welcomed and needed.

Said that, by checking settings, some are specific to a module and could be moved. For example, a setting for white balance and a setting for color balance. But maybe more could be moved.

This comes from previous discussion in some PR or issues like  #6706, #7021 or #6704

**Describe the solution you'd like**

We already have some modules, in lighttable, with specific preferences like metadata editor or export ones. The idea would be to moved really specific settings to a module in the module.

That could simplify a little bit preferences window and so help finding remaining settings more quickly and at the same time, having specific settings of a module directly accessible in the module or part of lighttable. 

We just need to be careful to only moved some specific settings. To avoid moving too much settings (that could be the risk), I think only some settings for a specific part of a darktable view could be moved. For example, moving settings for darkroom, it's no but moving a setting for a darkroom module, that could be yes as it's a specific part of the view and not related to the view or many parts of the view. Hope I'm clear here about the limit to fix.

The idea here is to discuss about that and see how that could be moved for next release. That could be for the next minor one 3.4.1 or next major one, depending on choices discussed.

Some settings I think they can be moved:

**- In lighttable tab:**
_delimiters for size categorise_
_overlay txt sidecar over zoomed images_

They can be moved in star overlays menu as they are related to information and display of thumbnail, visually, like all other options in star overlays menu.

**- In darkroom tab:**
_ white balance slider colors_: could be moved in white balance module
_colorbalance slider block layout_: could be moved in color balance module

Those are just ideas and could be changed/completed/updated, etc. It's all about discussing that and find the best way for a better experience. We need to remember some things for that discussion:

- Avoid to have in the future an overloaded preferences window
- Be sure all choices are consistent between them
- Avoid having also too many places for settings (that's the main reason for my suggestion to limit setting to be moved to specific part of a view, cited above here)

@elstoc, @jenshannoschwalm, @TurboGit, @aurelienpierre, @AlicVB, @johnny-bit: your views are welcomed here.

AlicVB:
@Nilvus : first, I completely agree.

Now for the part I know the best : 
- `overlay txt sidecar over zoomed images` is not used anymore... good catch. Just done a PR to remove it.
[...]

PR #7052, opened 2020-11-24
pref : remove unused "show text sidecar"

AlicVB:
as pointed by @Nilvus in #7050


TurboGit (approved):
Good !

wpferguson:
This actually goes with the lua script official/generate_image_txt.lua.  If we are doing away with the option to display it, then I guess we can do away with the script that generates the text to display.  

I don't know if anyone uses this.  I think I'll hold onto the script for awhile and see if this raises enough of a ruckus that it needs to be reverted.  If not, then I'll remove the script in a few months.

AlicVB:
this feature has not disappear in fact, but is replaced by the use of $(SIDECAR_TXT) in overlay content.

wpferguson:
@AlicVB thanks.  I'll undeprecate the script. :)

PR #7053, opened 2020-11-24
 colorin: add default color profile for PFM

aurelienpierre:
fix evil #6666

based upon #7029


jenshannoschwalm:
Reminds me of something i forgot to mention, the new colorin stuff also gives messages for true monochrome images. Do you need mor info?

PR #7055, opened 2020-11-24
doc : remove unused sidecar_txt + improve thumb delimiters + fix and refactor variables

AlicVB:
discussed in #7050


AlicVB:
Last commit wasn't originally meant to be on this branch... but as it's for docs too...

So lats commit changes : 
- ensure the list of available variables is complete (some were missing)
- put the variable list in its own section, as it's used in several places now, not only in export
- change the links accordingly

As usual, don't hesitate to review and propose better english sentences...

Nilvus:
@AlicVB: tooltip of thumb delimiters needs to be improved too. It's really not clear by now. What you write here in doc about that one is much better. Tooltip in preferences should be more practical to how and why use this setting. Something like:

"Set different overlays and css values depending of the size of the thumbnail. Set here thumbnails sizes delimiters, separated by "|", for example "120|400" means 3 categories of thumbnails : 0px->120px, 120px->400px and >400px"

AlicVB:
@Nilvus : done. thanks for the reminder, I completely forget that

TurboGit (approved):
Works for me, thanks!

PR #7059, opened 2020-11-25
colorpicker: convert color patch to display profile

### LINKED ISSUE #2106: Color picker with softproofing displays incorrect color
When activating softproofing with the new "color-space aware" picker, the RGB values displayed seem correct, but the color rendered on screen is incorrect.

**To Reproduce**
1. Select an area with the color picker. Optionally click "add" on the picker panel to remember the area. We get:
picker-normal
2. Choose a histogram&picker profile very different from display profile, eg Linear Rec2020 RGB.
3. Activate softproofing. The color patches change completely:
picker-softproof

**Expected behavior**
Unless softproofing actually changes the image displayed because of out-of-gamut colors, the color patches displayed in the picker panel shouldn't change. What's expected is that the RGB values displayed as numbers should change, but I guess a conversion from softproof color space to display color space is missing when displaying the color patches.

 - OS: Ubuntu
 - Version: master as of today.

Cc: @edgardoh, author of #2069 .

aurelienpierre:
fix #2106

Now the color is consistent with the image no matter what histogram profile is used. Before, it was completely off, the color used the numbers expressed in histogram space to paint in display space.

Notice the tooltip color is not fixed, I don't know where it is coded. Buuuut… it's our first ever color-managed GUI bit.

Hi @moy by the way !

Screenshot_20201125_034434

### DIFF
--- a/src/libs/colorpicker.c
+++ b/src/libs/colorpicker.c
@@ -113,9 +113,11 @@ static inline gboolean _convert_color_space(const GdkRGBA *restrict sample, GdkR
                              display_profile->unbounded_coeffs_out, display_profile->lutsize,
                              display_profile->nonlinearlut);
 
-  color->red = RGB[0];
-  color->green = RGB[1];
-  color->blue = RGB[2];
+  // Sanitize values and ensure gamut-fitting
+  // we reproduce the default behaviour of colorout, which is harsh gamut clipping
+  color->red = CLAMP(RGB[0], 0.f, 1.f);
+  color->green = CLAMP(RGB[1], 0.f, 1.f);
+  color->blue = CLAMP(RGB[2], 0.f, 1.f);
 
   return FALSE;
 }
@@ -184,9 +186,9 @@ static void _update_sample_label(dt_colorpicker_sample_t *sample)
   }
 
   // Setting the output button
-  sample->rgb.red   = CLAMP(rgb[0], 0.f, 1.f);
-  sample->rgb.green = CLAMP(rgb[1], 0.f, 1.f);
-  sample->rgb.blue  = CLAMP(rgb[2], 0.f, 1.f);
+  sample->rgb.red   = rgb[0];
+  sample->rgb.green = rgb[1];
+  sample->rgb.blue  = rgb[2];
 
   // Setting the output label
   char text[128] = { 0 };
--- a/src/libs/colorpicker.c
+++ b/src/libs/colorpicker.c
@@ -19,6 +19,7 @@
 #include "bauhaus/bauhaus.h"
 #include "libs/colorpicker.h"
 #include "common/darktable.h"
+#include "common/iop_profile.h"
 #include "control/conf.h"
 #include "control/control.h"
 #include "develop/develop.h"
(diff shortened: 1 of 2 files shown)


johnny-bit:
Tooltip color is @dterrahe's work i tink, so ping?

dterrahe:
Tooltip colors are set here
https://github.com/darktable-org/darktable/blob/24e7212c1887e5601e6c3dc56da5f52ca2260d93/src/libs/colorpicker.c#L258

so another call to gdk_cairo_set_source_rgba should do it?

btw is the gdk_rgba_copy required? the resulting _color_ seems to be overwritten by _convert_color_space anyway? so just allocating a local `GdkRGBA color;` might work too (and not need freeing so simplifying code. Haven't tried so could be wrong.

aurelienpierre:
@dterrahe so the color is painted through an XML field ? That's kind of brittle, the CSS colors are supposed to be given in sRGB no matter what. At least it is the W3C specification, I'm not sure how close Gtk CSS follows it.

dterrahe:
OK, it _looks_ like gtk treats css color codes the same as gdk color codes (as far as I can tell the colors in the tooltip are identical to the colors drawn in the colorpicker lib). Their CSS is loosely based on the standard, but doesn't seem to want to follow it _too_ closely. Now they definitely _could_ break this in future and therefore you _could_ decide that to be future proof you have to build your tooltip from a set of widgets rather than giving it a few lines of markup. And then _hope_ that when they go about breaking stuff, they don't break widgets in tooltips.

_Or_ you could decide to spend that time when they actually break your colors.

Or _even_ if you were so inclined, you could rip out working code and functionality because you don't like its non-future proofness; I'm hereby giving you permission to "revert" the relevant parts of #6350.

aurelienpierre:
Well it's surely more robust to handle Gdk buffers directly, without an extra layer of possibly broken interface. For now it will do, I think. Otherwise, painting a colored & color-managed rectangle is fairly easy:

```
  GdkRGBA *color = gdk_rgba_copy(&sample->rgb);
  if(_convert_color_space(&sample->rgb, color))
  {
    // function failed, profiles are not set, color will be wrong, exit.
    gdk_rgba_free(color);
    return TRUE;
  }

  gdk_cairo_set_source_rgba(cr, color);
  cairo_rectangle(cr, 0, 0, width, height);
  cairo_fill (cr);

  gdk_rgba_free(color);
```

Something else I noticed, the regular color picker UI gives the RGB figures rounded to 0-255, while the tooltip doesn't round. I think not rounding is indeed better since it makes out-of-gamut pops up in your face.  Shouldn't we synchronize the behaviours ?

@TurboGit do we consider this bug fixing for 3.4 ?

dterrahe:
> Otherwise, painting a colored & color-managed rectangle is fairly easy:

The tricky bit is that you have to put it inside a tooltip. You can give a tooltip either a string (possibly marked up) or a (custom) widget. So now you have to build a widget (probably a grid of them) to paint the colors and also the rest of the text fields in the tooltip. And you have to be a bit careful about memory managing that widget, because it occasionally gets automatically destroyed. None of it is rocket science (although it is finicky) but it is work.

If you want to go that route, I recommend having a look at the history tooltip, which does this, albeit with just one widget.

The inconsistent not-/rounding I think was there historically. I usually try not to break something that possibly was put there intentionally, but I agree it doesn't make sense to me either.

TurboGit:
> @TurboGit do we consider this bug fixing for 3.4 ?

Definitely for me, wrong colors in our loved RAW processor is certainly bad enough for this to be fixed in next release.

TurboGit:
@aurelienpierre : do you want this to be merged now or prefer having the fix for the tooltip in at the same time ?

src/libs/colorpicker.c
```diff
@@ -251,9 +295,35 @@ static gboolean _sample_tooltip_callback(GtkWidget *widget, gint x, gint y, gboo
 
   for(int i = 0; i < 3; i++)
   {
-    const float *rgb = i == 0 ? sample->picked_color_rgb_mean :
```
TurboGit, line 254: keep that, name rgb picked_rgb

src/libs/colorpicker.c
```diff
@@ -251,9 +295,35 @@ static gboolean _sample_tooltip_callback(GtkWidget *widget, gint x, gint y, gboo
...
-                       i == 1 ? sample->picked_color_rgb_min :
-                                sample->picked_color_rgb_max;
+    float rgb[3];
+    float rgb_strings[3];
+
+    if(i == 0)
```
TurboGit, line 301: remove the whole condition and write:

`for(size_t c = 0; c < 3; c++) rgb[c] = rgb_strings[c] = picked_rgb[c];`

A single for loop will make the code lighter and easier to read.

TurboGit (changes requested):
Minor style change.

aurelienpierre:
@TurboGit ok I will rewrite the code to make it lighter. Merging ASAP is good for me.

src/libs/colorpicker.c
aurelienpierre: done

aurelienpierre:
I have done the change and removed the clipping of RGB values in labels.

Now I wonder if having the name of the color space in which the values are expressed wouldn't be a good idea, especially since these spaces can be either linear or have an OETF on top, in which case the middle grey value would be very different.

TurboGit (approved):
Works for me !

TurboGit:
> Now I wonder if having the name of the color space in which the values are expressed wouldn't be a good idea,

Probably, worth recording an enhancement request.

I have a question for you on #7067

dterrahe:
> Now I wonder if having the name of the color space in which the values are expressed wouldn't be a good idea,

A good reason to provide introspection $DESCRIPTIONs for dt_colorspaces_color_profile_type_t in colorspaces.h? Those could then be uses in the colorpicker tooltip (this is where I assume you want it to show?) but would also be used in history diff tooltip.

aurelienpierre:
> A good reason to provide introspection $DESCRIPTIONs for dt_colorspaces_color_profile_type_t in colorspaces.h? Those could then be uses in the colorpicker tooltip (this is where I assume you want it to show?) but would also be used in history diff tooltip.

Yes, maybe. But not just in the tooltip, actually it could go in the RGB/Lab combobox (so "RGB" would become "Adobe RBG" or "sRGB) or somewhere in a label. Otherwise, users have to know the color picker uses histogram profile (which is… not intuitive) and remember which space they set. I feel it's especially important now that we don't clip the RGB values and show out-of-gamut: many users will not consider that the out of gamut are in sRGB space and will freak out, because RGB is absolute RGB for most of them.

PR #7062, opened 2020-11-25
darkoom: hidpi fix for iop drag icon

dtorop:
The icon made from the iop header previously rendered extra large when re-ordering iops on hidpi displays. This is because it was rendered via a GdkPixbuf, which is not aware of device vs. logical pixels. Instead, use Cairo -- which is hidpi aware -- to draw the icon.

This also fixes the mouse being 1/4 rather than 1/2-way along the icon on hipdi.

This PR also makes the gap which appears between iops be slightly smaller -- about the size of an iop header.

**Before**

image

**After**

image

There is now a class `iop_drop_icon` which can be used to style the dragging-iop image.

### DIFF
--- a/data/themes/darktable.css
+++ b/data/themes/darktable.css
@@ -2088,13 +2088,11 @@ scale contents trough slider:hover
 
 .iop_drop_before
 {
-   border: 2px solid @plugin_bg_color;
    border-bottom: 2.4em solid @plugin_bg_color;
 }
 
 .iop_drop_after
 {
-   border: 2px solid @plugin_bg_color;
    border-top: 2.4em solid @plugin_bg_color;
 }
 
--- a/data/themes/darktable.css
+++ b/data/themes/darktable.css
@@ -2080,16 +2080,22 @@ scale contents trough slider:hover
   background-color: @scroll_bar_active;
 }
 
+.iop_drag_icon
+{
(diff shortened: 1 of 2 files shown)


dtorop:
More CI problems with this PR! Not sure what is up. I can rebase slightly and push the branch again if it helps.

johnny-bit:
Yea, CI sometimes has problems... you can either rebase and force-push, push empty commit or ask someone to re-kick the ci :)

dtorop:
It looks like CI got nudged... I added another commit to fix CSS anyhow.

I realize that there's also one more annoyance, that the _get_dnd_dest_module() doesn't always choose the right module, which can make the visual feedback on drag be confusing. That could be another PR, not to make this one too complicated.

dtorop:
For reference here is better code for determining where the iop should drag to: https://github.com/dtorop/darktable/compare/iop-drag-icon-fix...dtorop:iop-drag-order-fix. I'm not totally thrilled with this, as it's a UI hack to deal with diverging cases in the underlying move code. I'll leave it out of this PR, which simply fixes a long-standing cosmetic hidpi bug.

TurboGit (approved):
Works for me, thanks!

dtorop:
> Works for me, thanks!

Great! Thank you for reviewing. Glad to have a tidier hidpi drag image.

PR #7068, opened 2020-11-25
ioporder : apply the right order for non-develop images

AlicVB:
this is a fix for the following "problem" :
(Note that if it's intended, I'll close this PR, as I can live with that and a automatic preset or such...)
- open dt in `display-referred` workflow and use embedded previews
- import some images without xmps
- open one image in darkroom and see order = legacy => normal
- change workflow to `scene-referred`
- open anther image. The order is legacy (should be v3.0 imo)

Now just for the record : 
- set dt on `scene-referred`
- import some images without xmps
- open an image in darkroom : order = v3.0 => ok

So the result is that for all our "old" images (non-edited) imported with the "old" display-reffered workflow we have to change manually the order in darkroom to get the right order we need for the workflow... Not the best imo...

What I've done is to use the auto-apply preset routine : here, we already check for auto-apply order. So if no order is auto-applied, we applied the order corresponding to the wanted workflow.
I think it's safe, because this function is only called for images without any history...

### DIFF
--- a/src/develop/develop.c
+++ b/src/develop/develop.c
@@ -1402,6 +1402,7 @@ static gboolean _dev_auto_apply_presets(dt_develop_t *dev)
   gchar *workflow = dt_conf_get_string("plugins/darkroom/workflow");
   const gboolean is_scene_referred = strcmp(workflow, "scene-referred") == 0;
   const gboolean is_display_referred = strcmp(workflow, "display-referred") == 0;
+  const gboolean is_workflow_none = strcmp(workflow, "none") == 0;
 
   workflow = dt_conf_get_string("plugins/darkroom/chromatic-adaptation");
   const gboolean is_modern_chroma = strcmp(workflow, "modern") == 0;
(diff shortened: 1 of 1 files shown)


src/develop/develop.c
```diff
@@ -1531,6 +1531,18 @@ static gboolean _dev_auto_apply_presets(dt_develop_t *dev)
     g_list_free_full(iop_list, free);
     dt_ioppr_set_default_iop_order(dev, imgid);
   }
+  else if(is_display_referred || is_scene_referred)
```
TurboGit, line 1534: why not just else here

src/develop/develop.c
```diff
@@ -1531,6 +1531,18 @@ static gboolean _dev_auto_apply_presets(dt_develop_t *dev)
...
+  {
+    // we have no auto-apply order, so apply iop order, depending of the worflow
+    GList *iop_list;
+    if(is_display_referred)
```
TurboGit, line 1538: and test if(is_scene_referred) here to use DT_IOP_ORDER_V30 and in all other cases (display referred or none) use the DT_IOP_ORDER_LEGACY order. I mean the iop order is not optional, we need one so even when no workflow is set we can apply it here. Or am I missing something ?

TurboGit (commented):
Some comments and possible actions if you agree.

src/develop/develop.c
AlicVB: oh, I thought that one order was applied in any case actually, so what done is more a correction than anything.
But you're right, we can handle all case here (remove this if above) as a safety check.
Now the question is : what should be the default order if workflow is set to none ? I would say v3.0 as it's more a "power user" value, but...

TurboGit: Agreed, none -> v3.0

AlicVB:
So, know we have in this order : 
1. auto-applied specific preset (set by user)
2. else workflow `scene-referred` or `none` => v3.0
3. else legacy

TurboGit:
Thanks !

PR #7069, opened 2020-11-25
eliminate false sharing in PRNG used  for dithering

ralfbrown:
The dithering and vignetting iops, as well as the OpenCL benchmarking code, use the Tiny Encryption Algorithm as a pseudorandom number generator.  Unfortunately, in each case, **they fail to pad** out the eight-byte TEA state, so that eight threads are contending for a single cache line and thus getting no more work done than a single thread would.

This PR pads out the state to a full cache line and consolidates the three completely separate copies of TEA into a single copy in the new file src/common/tea.h.  It yields an 8x speedup in random dithering and 2x speedup in vignetting at 8, 16, and 32 threads.

Integration tests are provided for the vignetting and dithering iops (both Floyd-Steinberg and random dithering, though FS is not affected by this PR).  expected.png files were produced using current master, 3.3.0+1796-gb1f735daf, and results with this PR were verified to by pixel-identical to those from master.


johnny-bit:
huh, interesting that gcc 9 build with debug config codepath doesn't want to build and error is:
```
bin/libdarktable.so: undefined reference to `encrypt_tea'
bin/libdarktable.so: undefined reference to `tpdf'
bin/libdarktable.so: undefined reference to `free_tea_states'
bin/libdarktable.so: undefined reference to `get_tea_state'
bin/libdarktable.so: undefined reference to `alloc_tea_states'
collect2: error: ld returned 1 exit status
```

parafin:
C is stupid like that, you should add static keyword to inline. In C++ just inline works, but not in C.

parafin:
See https://stackoverflow.com/questions/16245521/c99-inline-function-in-c-file/16245669#16245669

ralfbrown:
Can you tell that I work primarily in C++?  Forgot about the whole multiple instantiations when not inlining thing since that just works in C++.

TurboGit:
@ralfbrown : I have two tests failing with very large error:

```
DARKTABLE_CLI=/opt/darktable/bin/darktable-cli ./run.sh 0039-dithering-random 0040-vignetting
Test 0039-dithering-random
      Image mire1.cr2
      CPU & GPU version differ by 22460 pixels
      Max  dE         118.2754
      Mean dE         36.1391
      Count above max 2466156
  FAILS: image visually changed
         (2.5688e+06 pixels changed)

Test 0040-vignetting
      Image mire1.cr2
      CPU & GPU version differ by 729874 pixels
      Max  dE         6.9983
      Mean dE         0.2255
      Count above max 69393
  FAILS: image visually changed
         (681909 pixels changed)


Total test 2
Errors     2
```

TurboGit:
But the tests are also failing (the same way) with current master:

```
DARKTABLE_CLI=/opt/darktable/bin/darktable-cli ./run.sh 0039-dithering-random 0040-vignetting
Test 0039-dithering-random
      Image mire1.cr2
      CPU & GPU version differ by 22460 pixels
      Max  dE         118.2754
      Mean dE         36.1391
      Count above max 2466156
  FAILS: image visually changed
         (2.5688e+06 pixels changed)

Test 0040-vignetting
      Image mire1.cr2
      CPU & GPU version differ by 729874 pixels
      Max  dE         6.9983
      Mean dE         0.2255
      Count above max 69393
  FAILS: image visually changed
         (681909 pixels changed)


Total test 2
Errors     2
```

So we probably just need to update the expected.png... Let me try that.

HansBull:
Not to hijack the (much appreciated) thread, but would it be possible to compress those pngs with `zopflipng` to shave some 10% off ?

ralfbrown:
> But the tests are also failing (the same way) with current master:

I generated the tests with the version of master from which I branched, i.e. from yesterday.  I think we're being hit by differences in OpenMP again -- the dither and vignette code uses the thread number as part of the seed: `` tea_state[0] = j * height + dt_get_thread_num();``

So we'll not only get different outputs for different numbers of threads, but also if the threads get allocated different numbers.  Perhaps we should simply remove that and seed based on just the row number (j).  That will alter the rendering of old edits, but the added noise is supposed to be random anyway :).

TurboGit:
Ok, I regenerated them and not diff with master and with this PR. I have merged manually.

Thanks a lot @ralfbrown for this tedious but important work!

TurboGit:
> the dither and vignette code uses the thread number as part of the seed:

Which are fixed with the integration test driver and we do only the check for non OpenCL code for now.

ralfbrown:
I pulled and built master with this PR merged, and now I get the exact same numbers for the dithering-random and vignetting tests that you got earlier....
```
this is darktable-cli 3.3.0+1821~g15e50c75e

Test 0044-dithering-random
      Image mire1.cr2
      Max  dE         118.2754
      Mean dE         36.1391
      Count above max 2466156

Test 0045-vignetting
      Image mire1.cr2
      Max  dE         6.9983
      Mean dE         0.2255
      Count above max 69393
```

I'll open a PR to remove the thread number from the seed to gain reproducibility.

TurboGit:
There is something else, no ? The number of threads are fixed and I can run the test any number of time without issue on my side.

ralfbrown:
It's the thread **number**.  We can each be running four threads, but if your OpenMP assigns them in order 0,1,2,3 and mine in order 3,2,1,0, we'll get different results, because the existing code uses those numbers as part of the random seed for each row of the image, not merely to locate a thread-specific buffer.

TurboGit:
Oh I see, me stupid :( Ok we want this to be removed indeed as we cannot guarantee a reproducible output from one computer to another.

PR #7070, opened 2020-11-25
fix buffer overflow in colorcontrast

ralfbrown:
Due to a typo (repeated via copy-and-paste) of "+4" instead of "+3", the colorcontrast code can access one float beyond the image buffer.
This PR fixes that error and adds an integration test for colorcontrast.


TurboGit:
Manually merged. Thanks.

PR #7071, opened 2020-11-25
add integration tests for recently-deprecated iops

ralfbrown:
This PR adds integration tests for the four recently-deprecated iops which didn't already have them.  Since it will soon become difficult to create these tests (when the iops disappear entirely from the user interface some time next year), we should add them now to ensure that even the no-longer-used modules are checked for regressions.

The test numbers have been chosen to avoid clashing with PRs 7069 and 7070.


TurboGit:
Manually merged, thanks.

PR #7072, opened 2020-11-25
don't compute random number for dither when dithering is disabled

ralfbrown:
This is a companion to #7069, but can be merged independently.  Yields a substantial speedup by skipping the work of computing a random number and then multiplying it by zero.


TurboGit (approved):
Ok, looks good. Thanks.

PR #7073, opened 2020-11-26
Improve OpenMP directives

ralfbrown:
This PR adds the "collapse(2)" option to the #pragma omp directive for several parallelized loops which have an inner loop nested perfectly inside.

It also adds three integration tests for demosaic, to test the PPG, Amaze, and VNG4 algorithms with different add-on options.


TurboGit:
Thanks !

PR #7074, opened 2020-11-26
refactor: consolidate declarations of dt_omp_shared

ralfbrown:


johnny-bit:
This CI run actually succeeded, there are only issues with brew on macOS. Re-kicking CI action should help :)

parafin:
Re-clicked. Though it's very stupid that I have the ability only to re-run all jobs, not to retry the failed one and continue the process...

PR #7075, opened 2020-11-26
[performance] rewrite discrete wavelet transform decomposition

ralfbrown:
* Rewrite plain C codepath, ensuring that it is easy to auto-vectorize
* Remove SSE codepath, because the auto-vectorized plain C is now just as fast
* Reorder the processing to make it cache-friendlier, gaining extra speed

Due to the changed order of computation, rounding errors accumulate differently, so there is a minor difference in output (24 pixels changed on 0021-retouch-wavelets, max DE = 1.22).

Timings:

```
Thr	Old	       New
1	1.160	0.571
2	0.748	0.420
4	0.532	0.367
5	0.496	0.353
6	0.479	0.349
8	0.449	0.344
12	0.425	0.339
16	0.413	0.338
32	0.403	0.341
```
(Note that these times are for the entire retouch iop, which is doing a lot more than the DWT decomposition, including a fair amount of sequential work.)

### DIFF
--- a/src/common/dwt.c
+++ b/src/common/dwt.c
@@ -138,8 +138,8 @@ static inline int rowid_to_row(const int rowid, const int height, const int scal
   // process rows 0, 16, 32, ..., then 1, 17, 33, ..., 2, 18, 34, ..., etc.
   if (height <= scale)
     return rowid;
-  int per_pass = ((height + scale - 1) / scale);
-  int long_passes = height % scale;
+  const int per_pass = ((height + scale - 1) / scale);
+  const int long_passes = height % scale;
   // adjust for the fact that we have some passes with one fewer iteration when height is not a multiple of scale
   if (long_passes == 0 || rowid < long_passes * per_pass)
     return (rowid / per_pass) + scale * (rowid % per_pass);
--- a/src/common/dwt.c
+++ b/src/common/dwt.c
@@ -105,126 +105,24 @@ int dt_dwt_first_scale_visible(dwt_params_t *p)
   return _first_scale_visible(p->scales, p->preview_scale);
 }
 
-#define INDEX_WT_IMAGE(index, num_channels, channel) (((index) * (num_channels)) + (channel))
-#define INDEX_WT_IMAGE_SSE(index, num_channels) ((index) * (num_channels))
-
-/* code copied from UFRaw (which originates from dcraw) */
-#if defined(__SSE__)
-static void dwt_hat_transform_sse(float *temp, const float *const base, const int st, const int size, int sc,
-                                  const dwt_params_t *const p)
-{
-  int i;
-  const __m128 hat_mult = _mm_set1_ps(2.f);
-  __m128 valb_1, valb_2, valb_3;
-  sc = (int)(sc * p->preview_scale);
-  if(sc > size) sc = size;
-
-  for(i = 0; i < sc; i++, temp += 4)
-  {
-    valb_1 = _mm_load_ps(&base[INDEX_WT_IMAGE_SSE(st * i, p->ch)]);
-    valb_2 = _mm_load_ps(&base[INDEX_WT_IMAGE_SSE(st * (sc - i), p->ch)]);
-    valb_3 = _mm_load_ps(&base[INDEX_WT_IMAGE_SSE(st * (i + sc), p->ch)]);
-
-    _mm_store_ps(temp, _mm_add_ps(_mm_add_ps(_mm_mul_ps(hat_mult, valb_1), valb_2), valb_3));
-  }
-  for(; i + sc < size; i++, temp += 4)
-  {
-    valb_1 = _mm_load_ps(&base[INDEX_WT_IMAGE_SSE(st * i, p->ch)]);
-    valb_2 = _mm_load_ps(&base[INDEX_WT_IMAGE_SSE(st * (i - sc), p->ch)]);
-    valb_3 = _mm_load_ps(&base[INDEX_WT_IMAGE_SSE(st * (i + sc), p->ch)]);
-
-    _mm_store_ps(temp, _mm_add_ps(_mm_add_ps(_mm_mul_ps(hat_mult, valb_1), valb_2), valb_3));
-  }
-  for(; i < size; i++, temp += 4)
-  {
-    valb_1 = _mm_load_ps(&base[INDEX_WT_IMAGE_SSE(st * i, p->ch)]);
-    valb_2 = _mm_load_ps(&base[INDEX_WT_IMAGE_SSE(st * (i - sc), p->ch)]);
-    valb_3 = _mm_load_ps(&base[INDEX_WT_IMAGE_SSE(st * (2 * size - 2 - (i + sc)), p->ch)]);
-
-    _mm_store_ps(temp, _mm_add_ps(_mm_add_ps(_mm_mul_ps(hat_mult, valb_1), valb_2), valb_3));
-  }
-}
+#ifdef _OPENMP
+#pragma omp declare simd aligned(img,layers : 16)
 #endif
-
-static void dwt_hat_transform(float *temp, const float *const base, const int st, const int size, int sc,
-                              const dwt_params_t *const p)
+void dwt_add_layer(const float *const img, float *const layers, dwt_params_t *const p, const int n_scale)
 {
-#if defined(__SSE__)
-  if(p->ch == 4 && p->use_sse)
-  {
-    dwt_hat_transform_sse(temp, base, st, size, sc, p);
-    return;
-  }
-#endif
-
-  int i, c;
-  const float hat_mult = 2.f;
-  sc = (int)(sc * p->preview_scale);
-  if(sc > size) sc = size;
-
-  for(i = 0; i < sc; i++)
-  {
-    for(c = 0; c < p->ch; c++, temp++)
-    {
-      *temp = hat_mult * base[INDEX_WT_IMAGE(st * i, p->ch, c)] + base[INDEX_WT_IMAGE(st * (sc - i), p->ch, c)]
-              + base[INDEX_WT_IMAGE(st * (i + sc), p->ch, c)];
-    }
-  }
-  for(; i + sc < size; i++)
-  {
-    for(c = 0; c < p->ch; c++, temp++)
-    {
-      *temp = hat_mult * base[INDEX_WT_IMAGE(st * i, p->ch, c)] + base[INDEX_WT_IMAGE(st * (i - sc), p->ch, c)]
-              + base[INDEX_WT_IMAGE(st * (i + sc), p->ch, c)];
-    }
-  }
-  for(; i < size; i++)
-  {
-    for(c = 0; c < p->ch; c++, temp++)
-    {
-      *temp = hat_mult * base[INDEX_WT_IMAGE(st * i, p->ch, c)] + base[INDEX_WT_IMAGE(st * (i - sc), p->ch, c)]
-              + base[INDEX_WT_IMAGE(st * (2 * size - 2 - (i + sc)), p->ch, c)];
-    }
-  }
-}
-
-#if defined(__SSE__)
-static void dwt_add_layer_sse(float *const img, float *layers, dwt_params_t *const p, const int n_scale)
-{
-  const int i_size = p->width * p->height * 4;
-
+  assert(p->ch == 4);
+  const int i_size = 4 * p->width * p->height;
 #ifdef _OPENMP
 #pragma omp parallel for default(none) \
-  dt_omp_firstprivate(img, i_size) \
-  shared(layers) \
-  schedule(static) num_threads(MIN(8,darktable.num_openmp_threads))
+  dt_omp_firstprivate(i_size) \
+  dt_omp_sharedconst(img, layers)   \
+  schedule(static) \
+  num_threads(MIN(6,darktable.num_openmp_threads))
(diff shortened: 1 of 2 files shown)


johnny-bit:
I'm amazed at autovectorization stuff :)

The CI just needs a re-kick :) Otherwise it succeeded :)

dterrahe:
I fully agree with @johnny-bit sentiments; thanks @ralfbrown for all your work on this! It is unsexy invisible background stuff, but appreciated by all! And the "unsexy" bit can be debated.

About the SSE removal; is it policy that SSE and OpenMP are selected based on speed? Basically asking whether OpenMP is available/supported/stable on all platforms. Otherwise there might still be a reason to keep SSE around. Although I'm generally much in favour of cleaning up old cruft.

And another question, since you look to be the resident openmp expert; have you come across openmp gpu offloading and could you explain it to me/us? Basically I'm wondering if this is something that could be used in dt and whether it would be a huge amount of work (possibly duplicated for each supported GPU type) or be (much) less efficient than opencl. It seems the main objective is to avoid having to implement calculations in two completely different frameworks (openmp & cl) which would be a very worthwhile goal for dt.

I'm talking about this 
https://gcc.gnu.org/wiki/Offloading
this
https://hpc-wiki.info/hpc/Building_LLVM/Clang_with_OpenMP_Offloading_to_NVIDIA_GPUs
and this
https://github.com/ROCm-Developer-Tools/aomp

parafin:
Kicked the CI.
Regarding removal of SSE code - clang complains a lot about optimizations that were requested through openmp pragmas but it couldn't perform (look for "warning: loop not vectorized:" messages in build logs). So these speedups should be checked on clang build too before discarding the original SSE code (I suspect reported benchmarks are with GCC build).

AxelG-DE:
> It is unsexy invisible background stuff, but appreciated by all! And the "unsexy" bit can be debated.

I am the performance hunter. To me it is damned sexy work!
@ralfbrown Thanks a lot!

ralfbrown:
> About the SSE removal; is it policy that SSE and OpenMP are selected based on speed? 

You'd have to ask Pascal or Aurelien about that.  @parafin The numbers I reported above were indeed with GCC 8.4, but compare the old SSE code against the new auto-vectorized code.  The SSE code would have had to be rewritten as well, which I could still do if there's demand (won't get around to it until Christmas, but I'm not expecting this to be merged until after the 3.4 release anyway).

> And another question, since you look to be the resident openmp expert; have you come across openmp gpu offloading and could you explain it to me/us?

Not really an expert, have only been using it since I started contributing to dt in April, and am just now getting into the SIMD aspects of it.  I'm aware of the GPU offloading, but no details; it's actually a separate spec from OpenMP parallelization.

parafin:
> > About the SSE removal; is it policy that SSE and OpenMP are selected based on speed?
> 
> You'd have to ask Pascal or Aurelien about that. @parafin The numbers I reported above were indeed with GCC 8.4, but compare the old SSE code against the new auto-vectorized code. The SSE code would have had to be rewritten as well, which I could still do if there's demand (won't get around to it until Christmas, but I'm not expecting this to be merged until after the 3.4 release anyway).

The demand depends on whether this openmp code works as fast with clang. Maybe if you share how you did the benchmark I could try to run it with Apple Clang, then we'll know if SSE code should stay or not.

parafin:
But looking at CI build logs I don't see the warnings I mentioned for dwt.c file, so it's hopeful.

ralfbrown:
in src/tests/integration:

$(CLI) --hq 1 images/mire1.cr2 0021-retouch-wavelets/retouch-wavelets.xmp /tmp/bench.png --core --configdir $(SCRATCH) --library :memory: -d perf -t $(THREADS) |& grep retouch

Replace $CLI by either the master build or the PR build and of course vary $THREADS.  I took the average of two runs in each condition to account for timing jitter.

TurboGit:
I agree that sse2 hand coded function should be avoided. Compiler are doing things that no human would be able to write.

The target-clones that Aurélien had introduced in some routines have more futur as not only it support sse2 but sse3, sse4, avx, avx2, avx512 and we could add more when needed. And again, I bet the compiler will beat human here.

@parafin : now the real question are target-clones supported in clang today ? If not, is there plan for supporting them ?

parafin:
No idea...

johnny-bit:
Clang has (well at least since 7.0) attribure `target` which seems to act same way as GCC's `target_clone` from cursory glance. Might be worth taking a look...

aurelienpierre:
For color conversions (see #6968) the SSE2 path is about twice as slow than the pure C with OpenMP and native `__builtin_assume_aligned(ptr, 16)`. I tried to optimize it manually for a day, using SSE4 dot product intrinsics, but I still couldn't get as fast.

So, when you put together the non-portability, barely legible SSE2 intrinsics (that are actually quite outdated), and the code redundancy that can lead to inconsistencies and mistake, better ditch SSE2 path wherever it is slower than regular C.

aurelienpierre:
> Due to the changed order of computation, rounding errors accumulate differently, so there is a minor difference in output (24 pixels changed on 0021-retouch-wavelets, max DE = 1.22).

There is a `DT_FMA(x, y, z)` macro in `darktable.h` that uses hardware FMA if available, or else defaults to x × y + z, if you want proper rounding when possible.

ralfbrown:
It's an inherent loss of low-order bits from summing multiple values, and which bits get lost depends on the order of summing....
FMA won't help any in this case, since we're just doing (a+a+b+c) and then taking those results and doing (d+d+e+f) over them.

src/common/dwt.c
```diff
@@ -105,268 +105,159 @@ int dt_dwt_first_scale_visible(dwt_params_t *p)
...
-    bl[i] = bl[i] * lpass_mult;
-    bh[i] -= bl[i];
-  }
+  // to make this algorithm as cache-friendly as possible, we want to interleave the actual processing of rows
+  // such that the next iteration processes the row 'scale' pixels below the current one, which will already
+  // be in L2 cache (if not L1) from having been accessed on this iteration so if vscale is 16, we want to
+  // process rows 0, 16, 32, ..., then 1, 17, 33, ..., 2, 18, 34, ..., etc.
+  if (height <= scale)
+    return rowid;
+  int per_pass = ((height + scale - 1) / scale);
```
TurboGit, line 141: const

src/common/dwt.c
```diff
@@ -105,268 +105,159 @@ int dt_dwt_first_scale_visible(dwt_params_t *p)
...
+  int long_passes = height % scale;
```
TurboGit, line 142: const

src/common/dwt.c
```diff
@@ -415,7 +295,7 @@ static void dwt_wavelet_decompose(float *img, dwt_params_t *const p, _dwt_layer_
 
   if(p->merge_from_scale > 0)
   {
-    merged_layers = dt_alloc_align(64, p->width * p->height * p->ch * sizeof(float));
+    merged_layers = dt_alloc_sse_ps(p->width * p->height * p->ch);
```
TurboGit, line 298: likewise

src/common/dwt.c
```diff
@@ -380,33 +271,22 @@ static void dwt_wavelet_decompose(float *img, dwt_params_t *const p, _dwt_layer_
...
-  // setup a temp buffer
-  temp = dt_alloc_align(64, MAX(p->width, p->height) * p->ch * sizeof(float));
-  if(temp == NULL)
-  {
-    printf("not enough memory for wavelet decomposition");
-    goto cleanup;
-  }
-  memset(temp, 0, MAX(p->width, p->height) * p->ch * sizeof(float));
-
+  buffer[1] = dt_alloc_sse_ps(size);
```
TurboGit, line 283: it is a bit strange to see all sse code removed, but using dt_alloc_sse_ps() here.

src/common/dwt.c
```diff
@@ -380,33 +271,22 @@ static void dwt_wavelet_decompose(float *img, dwt_params_t *const p, _dwt_layer_
...
   // buffer to reconstruct the image
-  layers = dt_alloc_align(64, p->width * p->height * p->ch * sizeof(float));
-  if(layers == NULL)
+  layers = dt_alloc_sse_ps(4 * p->width * p->height);
```
TurboGit, line 285: likewise

src/common/dwt.c
```diff
@@ -380,33 +271,22 @@ static void dwt_wavelet_decompose(float *img, dwt_params_t *const p, _dwt_layer_
...
+  // scratch buffer for decomposition
+  temp = dt_alloc_sse_ps(dt_get_num_threads() * 4 * p->width);
```
TurboGit, line 287: likewise

TurboGit (changes requested):
some comments & possibly changes

src/common/dwt.c
ralfbrown: I used dt_alloc_sse_ps because it's specifically for allocating floats (so no need for sizeof(float)), and declares alignment to the compiler.  It is in fact just a wrapper for dt_alloc_align.  We could make a copy called something like dt_alloc_aligned_float to make the intent clearer.  And by hiding the alignment specification inside the function, it becomes a single location to update if we ever need to port to an architecture with cache lines bigger than 64 bytes.

src/common/dwt.c
ralfbrown: Added.

src/common/dwt.c
TurboGit: Yes, I'm for the copy with new name. My point is really about the code clarity even if current code is correct it is misleading to me.

aurelienpierre:
Also, to improve perf you can tag `restrict` the pointers you know will not be aliased in the program (in, out, above, center, bottom, etc.)

johnny-bit:
Regrarding atribute `target_clones`, here's what Roman told me:

```
LebedevRI
hm, target_clones support seems to have never landed
yeah, i'm also surprized (https://reviews.llvm.org/D51650)
what should work is intel spelling cpu_dispatch/cpu_specific
```

src/common/dwt.c
TurboGit: @ralfbrown : Ping ? You may have skipped this one ?

jenshannoschwalm:
Something i am not sure about, does dt use -O3 by default on your system or do you have to set that manually?

johnny-bit:
@jenshannoschwalm - CMake sets those flags. `RelWithDebInfo` is `-O2` while `Release` is `-O3`. You can see it in here:

https://github.com/darktable-org/darktable/blob/7694d8c79557002cc85dee15482ce10b2e6de9ee/src/CMakeLists.txt#L580-L581

jenshannoschwalm:
Aah. Somehow my cmake-gui misses that ...

src/common/dwt.c
TurboGit: @ralfbrown : I've done that and will merge.

parafin:
> But looking at CI build logs I don't see the warnings I mentioned for dwt.c file, so it's hopeful.

So with my older Xcode there are such warnings:
```
src/common/dwt.c:177:9: warning: loop not vectorized: the optimizer was unable to perform the requested transformation; the transformation might be disabled or specified as part of an unsupported
      transformation ordering [-Wpass-failed=transform-warning]
#pragma omp simd aligned(center, above, below, temprow : 16)
        ^
src/common/dwt.c:177:9: warning: loop not vectorized: the optimizer was unable to perform the requested transformation; the transformation might be disabled or specified as part of an unsupported
      transformation ordering [-Wpass-failed=transform-warning]
```
But the resulting code is still almost twice as fast on my MacBook, so indeed removing SSE codepath was the right decision.

erichkeane:
@LebedevRI ended up sending me this link separately.  Note attribute `target` and attribute `target_clones` multiversioning work the same way (using an indirect function aka `ifunc`), but have different ways of interacting with them.  `target_clones` takes 1 function definition and splits it, then dispatches on the split ones.  `target` allows you to type a separate definition for each, then dispatches on all of them.  For example:

`__attribute__((target_clones("sandybridge","ivybridge","default"))) int foo() { return 0; }`

and 

`__attribute__((target("sandybridge"))) int foo() { return 0; }
__attribute__((target("ivybridge"))) int foo() { return 0; }
__attribute__((target("default"))) int foo() { return 0; }`

are functionally identical.  So as a worst-case, you can use `target` with a macro to just blit out all versions of your function.

The advantage that `target` has is that you can give each version its own implementation.  This is typically used when you have a version of the function with inline-assembly custom-tuned to each processor, or that uses specific intrinsics.

the `cpu_specific` and `cpu_dispatch` are an Intel invented version (that predates both, side note, Clang's `target` implementation is based on GCC's as well) that was put into clang a few years back that has some of the advantages of both, with the ad/dis-advantage of having to specify the list of targets in the dispatch function.  This has pros/cons (the con being you have to type it, the pros being you can control which TU the dispatch appears).

If you expect to use multiple compilers on Linux, I obviously suggest `target`, but I'm hopeful `target_clones` will be supported in Clang 12.  (ICC also supports all 3 of these versions).

johnny-bit:
I hope for availability of `target_clones`, especially for binary packaging that would be awesome: "just take best possible autovectorization path" :) (On a side note - i hope that for march=native all that cloning is stripped to just native ;) )

erichkeane:
Unfortunately that is not an optimization that happens yet as far as I know.  I don't believe our optimizer looks through ifuncs.

the `target` with copy/paste WILL do that (see the example above), but you end up having to do the manual duplication.

PR #7076, opened 2020-11-26
pixelpipe_hb: for colorout (global picker) use the default colorspace.

### LINKED ISSUE #7067: regression : color picker error on console
**Describe the bug**
color picker issue error on console, this was fixed and it is back

**To Reproduce**

1. Go to 'darkroom'
2. Click on the color picker and select a point or area on image
3. See error

```
[_transform_matrix] invalid conversion from 1 to -1
[dt_ioppr_transform_image_colorspace] invalid conversion from 1 to -1
[_transform_matrix] invalid conversion from 2 to -1
[dt_ioppr_transform_image_colorspace] invalid conversion from 2 to -1
```

**Expected behavior**
no error on console

 - Darktable Version: master
 - OS: GNU/Linux

TurboGit:
This is used for the transform, and need to be set here as for the
global picker there is no picker inside the module.

Fixes #7067.

### DIFF
--- a/src/develop/pixelpipe_hb.c
+++ b/src/develop/pixelpipe_hb.c
@@ -983,7 +983,16 @@ static gboolean _transform_for_blend(const dt_iop_module_t *const self, const dt
 
 static dt_iop_colorspace_type_t _transform_for_picker(dt_iop_module_t *self)
 {
-  dt_iop_colorspace_type_t picker_cst = dt_iop_color_picker_get_active_cst(self);
+  // Special case the colorout module used for the global colorpicker. In this module
+  // we don't have the default colorspace set as there is no picker in the module itself.
+  // the global colorpicker use the colorout in the pipe as process point for taking the
+  // samples. And in this case we want the colorout default_colorspace().
+
+  const dt_iop_colorspace_type_t picker_cst =
+    !strcmp(self->op, "colorout")
+    ? self->default_colorspace(self, NULL, NULL)
+    : dt_iop_color_picker_get_active_cst(self);
+
   switch(picker_cst)
   {
     case iop_cs_RAW:
--- a/src/develop/pixelpipe_hb.c
+++ b/src/develop/pixelpipe_hb.c
@@ -983,15 +983,8 @@ static gboolean _transform_for_blend(const dt_iop_module_t *const self, const dt
 
 static dt_iop_colorspace_type_t _transform_for_picker(dt_iop_module_t *self)
 {
-  // Special case the colorout module used for the global colorpicker. In this module
-  // we don't have the default colorspace set as there is no picker in the module itself.
-  // the global colorpicker use the colorout in the pipe as process point for taking the
-  // samples. And in this case we want the colorout default_colorspace().
-
   const dt_iop_colorspace_type_t picker_cst =
-    !strcmp(self->op, "colorout")
-    ? self->default_colorspace(self, NULL, NULL)
-    : dt_iop_color_picker_get_active_cst(self);
+    dt_iop_color_picker_get_active_cst(self);
 
   switch(picker_cst)
   {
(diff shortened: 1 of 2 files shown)


src/develop/pixelpipe_hb.c
```diff
@@ -983,7 +983,14 @@ static gboolean _transform_for_blend(const dt_iop_module_t *const self, const dt
 
 static dt_iop_colorspace_type_t _transform_for_picker(dt_iop_module_t *self)
 {
-  dt_iop_colorspace_type_t picker_cst = dt_iop_color_picker_get_active_cst(self);
+  // Special case the colorout module used for the global colorpicker. In this module
```
hlecleme, line 986: Perhaps move this comment to `_request_color_pick`

hlecleme (approved):
Looks good to me.

The comment in `_transform_for_picker` should be moved or updated.

Although the picker in the color look up table module seemed to work, it is much better to make it consistent with what is done for the other modules :+1:

src/develop/pixelpipe_hb.c
TurboGit: Done !

src/iop/colorchecker.c
```diff
@@ -845,8 +846,8 @@ void cleanup_pipe(struct dt_iop_module_t *self, dt_dev_pixelpipe_t *pipe, dt_dev
 void gui_reset(struct dt_iop_module_t *self)
 {
   dt_iop_colorchecker_gui_data_t *g = (dt_iop_colorchecker_gui_data_t *)self->gui_data;
-  self->request_color_pick = DT_REQUEST_COLORPICK_OFF;
   dt_bauhaus_widget_set_quad_active(g->combobox_patch, 0);
```
dterrahe, line 849: No longer needed.

src/iop/colorchecker.c
```diff
@@ -933,25 +931,20 @@ void cleanup_global(dt_iop_module_so_t *module)
...
     self->gui_update(self);
   }
   dt_control_queue_redraw_center();
+}
 
-  if(self->off) gtk_toggle_button_set_active(GTK_TOGGLE_BUTTON(self->off), 1);
+static void _color_picker_callback(GtkWidget *button, gpointer user_data)
+{
+  dt_iop_module_t *self = (dt_iop_module_t *)user_data;
+  if(darktable.gui->reset) return;
```
dterrahe, line 946: This gets called when _iop_color_picker_callback_button_press (in color_picker_proxy.c) activates the togglebutton. It first sets gui->reset however, so this code returns with setting cst. If it _did_ set cst, the result looks wrong.

src/iop/colorchecker.c
```diff
@@ -1362,8 +1351,12 @@ void gui_init(struct dt_iop_module_t *self)
     snprintf(cboxentry, sizeof(cboxentry), _("patch #%d"), k);
     dt_bauhaus_combobox_add(g->combobox_patch, cboxentry);
   }
-  self->request_color_pick = DT_REQUEST_COLORPICK_OFF;
-  dt_bauhaus_widget_set_quad_paint(g->combobox_patch, dtgtk_cairo_paint_colorpicker, CPF_NONE, NULL);
+
+  GtkWidget *hbox = gtk_box_new(GTK_ORIENTATION_HORIZONTAL, DT_BAUHAUS_SPACE);
```
dterrahe, line 1355: These four lines can be simplified to 
`dt_color_picker_new(self, DT_COLOR_PICKER_POINT_AREA, g->combobox_patch);`
with the addition of 
`g_signal_connect(G_OBJECT(g->combobox_patch), "quad-pressed", G_CALLBACK(_color_picker_callback), self);`
if the picker_cst needs to be set, but as mentioned above this currently would cause an issue.

Also, I think the quad picker looks better aligned than the stand-alone color picker. They used to look very similar if I remember correctly, but the toggle button version has had more space added around it which now causes the combo to somewhat misalign.

Also remove g->colorpicker (in dt_iop_colorchecker_gui_data_t) if you make the change.

src/iop/colorchecker.c
```diff
@@ -1394,15 +1387,14 @@ void gui_init(struct dt_iop_module_t *self)
   dt_bauhaus_combobox_add(g->combobox_target, _("relative"));
   dt_bauhaus_combobox_add(g->combobox_target, _("absolute"));
 
-  gtk_box_pack_start(GTK_BOX(self->widget), g->combobox_patch, TRUE, TRUE, 0);
+  gtk_box_pack_start(GTK_BOX(self->widget), hbox, TRUE, TRUE, 0);
```
dterrahe, line 1390: revert if you change to quad-picker above

dterrahe (commented):
I'd be happy to submit a PR with the changes I'm suggesting, but I don't know what to do about the call to dt_iop_color_picker_set_cst

src/iop/colorchecker.c
hlecleme: Sorry, I think I overlooked the `dt_iop_color_picker_set_cst` call... I suppose that the module expects the picked colors in Lab and not Lch so, `dt_iop_color_picker_set_cst(self, iop_cs_Lab)` would be more correct.

As the initialization of the picker_cst is done to the module color space (Lab in this case) when creating the color picker (`dt_color_picker_new` via `_iop_init_picker`), it would be better to remove the call to `dt_iop_color_picker_set_cst`, and so perhaps remove the callback?

TurboGit:
> I'd be happy to submit a PR with the changes I'm suggesting, but I don't know what to do about the call to dt_iop_color_picker_set_cst

Let's go then, appreciated. Note that the dt_iop_color_picker_set_cst call is needed and is part of the fix for this issue. I'll be happy to test the rework on this but we must be very careful. And I don't pretend understand all the implication of all this as it is quite "complex".

dterrahe:
> Note that the dt_iop_color_picker_set_cst call is needed and is part of the fix for this issue. 

It seems that dt_iop_color_picker_set_cst  _NOT_ getting called anymore was part of the fix for this issue. It took me a long while to understand why my simplification misbehaved until I realised that the original change disabled dt_iop_color_picker_set_cst  and I reenabled it.

As @hlecleme says above, the default gets set to iop_cs_Lab and that seems to be correct. Whether it is _always_ correct, I don't know, so yes, more testing wouldn't hurt. But what I'm going to submit I believe is functionally equivalent (without confusing inactive code) to what is there now.

Issue #7037, opened 2020-11-23
Wrongfully disabling always-on modules via shortcuts and possibly more issues


jenshannoschwalm:
**We can enable or disable** any module via a user defined shortcut even if it is always-on or should be off for a specified image. The results are non-predictable, crashes included.

One example has been reported in #6992 (at least the issue of a disabled always-on module got a safety check) 

**To Reproduce**
Go to the preferences->shortcuts section, select the demosaic module and define a shortcut to enable it.
Go to darkroom using a raw image and press as you just defined before. Observe the module-status icon changing. (you can't change that by a mouse-click of course.)

**Expected behavior**
Difficult to describe, there might be many issues around.
1. There might be modules that must be off but are switched on. 
2. We could have switches or sliders hidden on purpose in that module that are not properly checked in commit_parameters so could lead to non-obvious changes or faults.

How to solve? I am not a friend of all these shortcuts and have personally not used any of them so far so i have not looked into or worked on that code at all. Problem (1) and the one reported in #6992 might want a flag in the module header to prohibit any on/off switching. Problem(2) seems very difficult to me.

**Platform (please complete the following information):**
 Current git master

**Additional context**
Valid for all OS, CL enabled or not.

elstoc:
The shortcut issue is fairly simple and is resolved in #7038. There might conceivably be modules that sometimes have an enable button and sometimes do not, so I've left it so that the user can still define a shortcut. It just won't do anything.

elstoc:
I didn't flag the PR as resolving this issue because I wasn't clear if there was something else you wanted to do here in addition

jenshannoschwalm:
Is it also possible to disable a shortcut for non-visible widgets? (For example in demosaic we have the 'edge threshold' slider that is not used in other demosaicers.)

elstoc:
Probably, but that would be more involved and would have to be done on a case-by-case basis.

jenshannoschwalm:
I found very simple ways to trigger such conditions

1. Open a raw in darkroom - Ctrl-C to copy history -- open a jpeg in darkroom -- Ctrl-V pasting rawdenoise and demosaic to the jpegs history.

2. As above also in lightroom

At least the is no crashing any more

PR #7078, opened 2020-11-26
don't accelerate hidden sliders/combos

dterrahe:
In response to https://github.com/darktable-org/darktable/issues/7037#issuecomment-732369258

This will not disable accelerators for widgets that are on a different (not the active) notebook tab. I think this is desirable behavior.

For dynamic accelerators, this just means pressing the key does nothing (it doesn't switch to the "dynamic" mode) so if you then scroll, it will zoom (normal scroll key behavior). Alternatively, the mode can be changed but the actual scroll ignored; in that case the hint (at top of screen) will be misleading. So we'd have to do both; amend the hint (to say widget is hidden/inactive) _and_ ignore, rather than passthrough, the scroll. Then you might want the hint to contain an explanation of _why_ the widget is hidden and _then_ you've created a lot of work...

### DIFF
--- a/src/gui/accelerators.c
+++ b/src/gui/accelerators.c
@@ -627,10 +627,18 @@ float dt_accel_get_slider_scale_multiplier()
   return dt_conf_get_float("darkroom/ui/scale_step_multiplier");
 }
 
+static gboolean _widget_invisible(GtkWidget *w)
+{
+  return (!gtk_widget_get_visible(w) ||
+          !gtk_widget_get_visible(gtk_widget_get_parent(w)));
+}
+
 static gboolean bauhaus_slider_increase_callback(GtkAccelGroup *accel_group, GObject *acceleratable,
                                                  guint keyval, GdkModifierType modifier, gpointer data)
 {
   GtkWidget *slider = GTK_WIDGET(data);
+
+  if(_widget_invisible(slider)) return TRUE;
 
   float value = dt_bauhaus_slider_get(slider);
   float step = dt_bauhaus_slider_get_step(slider);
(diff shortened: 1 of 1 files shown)


johnny-bit:
Does that work with white balance and R/G/B (not emerald) sliders when expander is collapsed?

dterrahe:
Afk.

Probably/maybe not. The "correct" approach there is to expand them when moved via accelerator, because they're just _hidden_, not really _disabled_, imho.

jenshannoschwalm:
Yeah. A sort of defensive solution but a safe one :-) Thanks for looking into this.

TurboGit (changes requested):
This breaks dynamic accels. Just assign E to exposure dynamic accel. Press E and scoll the darkroom is zooming instead of changing the exposure.

dterrahe:
You're right, this needs a more careful approach. Using _get_visible instead of _is_ probably doesn't give the same breakage, but might also not disable some _really_ hidden sliders.

Tbh i didn't think this was needed for 3.4. Changing a value that is hidden and not used isn't _pretty_, but it shouldn't cause major issues and as far as i know this is not a regression.

dterrahe:
Second attempt.

There are two cases (that I can think of) where this would not work:
1. If a widget is hidden by hiding an outer box that contains an inner box that contains the widget. To catch this, the parent's parent could be checked as well, but...
2. if a widget (slider/combo) is directly set as the module's top-level widget. In that case, when the module is closed/not shown in the modulegroup, the widget would not respond to shortcuts because this code would think it was deliberately hidden.

I think both cases are hypothetical, at least I'm not aware of any instances, so I'm not sure if it is necessary to make the code more robust (and slower), by adding the extra parent and by comparing the widget and parent to widget->module->widget before checking if _their_ parent is visible. There might also be side-effects to that that I'm not considering. @TurboGit what do you think?

TurboGit:
> There might also be side-effects to that that I'm not considering. @TurboGit what do you think?

The new implementation seems ok to me. I cannot thinks of a case where this would _actually_ not work. So let's merge this now, I'm confident that this will fill 99.9999% of the needs :) Thanks.

PR #7081, opened 2020-11-26
dithering: remove thread number from random seed

ralfbrown:
... since it makes results unreproducible

As mentioned in #7069.


TurboGit:
Thanks !

PR #7082, opened 2020-11-26
add integration tests

ralfbrown:
These are for modules I expect to be working on in the near future (rawdenoise already started), so that we don't wind up having to renumber tests or forcing @TurboGit to manually merge.

As suggested by @HansBull the expected.png files have been recompressed with zopflipng, making them 10-12% smaller.  It makes sense to do this for all new or updated expected.pngs going forward, but redoing existing images will simply bloat the repo.


TurboGit:
All good to me, thanks !

PR #7084, opened 2020-11-26
[performance] rewrite DWT in rawdenoise

ralfbrown:
Yields a 3x performance gain for 1-2 threads, decreasing to ~10% at 32 threads (memory bandwith limits) for Bayer-filtered sensors.  I haven't touched the X-Trans version yet, as I'll need to set up correctness tests.

Use integration test 0049-rawdenoise from #7082 to verify correctness.

Once #7075 is merged, a good chunk of the code can be moved to src/common/dwt.c which will allow elimination of some code I needed to duplicate while that PR was still pending.

### DIFF
--- a/src/iop/rawdenoise.c
+++ b/src/iop/rawdenoise.c
@@ -218,8 +218,8 @@ static inline int rowid_to_row(const int rowid, const int height, const int scal
   // process rows 0, 16, 32, ..., then 1, 17, 33, ..., 2, 18, 34, ..., etc.
   if (height <= scale)
     return rowid;
-  int per_pass = ((height + scale - 1) / scale);
-  int long_passes = height % scale;
+  const int per_pass = ((height + scale - 1) / scale);
+  const int long_passes = height % scale;
   // adjust for the fact that we have some passes with one fewer iteration when height is not a multiple of scale
   if (long_passes == 0 || rowid < long_passes * per_pass)
     return (rowid / per_pass) + scale * (rowid % per_pass);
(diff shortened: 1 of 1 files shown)


ralfbrown:
@TurboGit Regarding X-Trans sensors, it would be good to have an X-Trans image for the integration test suite (though we wouldn't want to run on it for every test and double the test suite's run time).  That will allow checks of X-Trans demosaicing algorithms as well as the X-Trans version of the raw denoising.

TurboGit:
>  it would be good to have an X-Trans image for the integration test suite

Sure would be nice to have one in the images directory. Do you have one to share ? I don't have an X-Trans camera, but I'm sure we can find a free sample somewhere. Let me know.

TurboGit:
And yes I agree we certainly don't want to run every test with it, but a test for the demosaic is really welcome.

johnny-bit:
@TurboGit - raw.pixls.us are no.1. source for all raws, including XTrans and all are CC0 licenced so can be freely included.

ralfbrown:
Just found something really strange -- the integration test fails with a bad strip at the right edge, but during dev I had been testing with the same image and .xmp and getting pixel-identical results.  The difference: integration test downsamples to 2048px wide, I had been using full-size output.  Will have to track that down....

TurboGit (approved):
Looks good to me, thanks !

TurboGit:
@johnny-bit : I'd like to have an **X-Trans with at least a color checker** in it. I'm not able to find one. I have found that the color checker in the Bayer sample we have in dt is very useful to do some checking when we encounter issues.

PR #7085, opened 2020-11-27
add integration tests (part 2)

ralfbrown:
A second batch of integration tests for modules not yet covered (at least one more such batch to go).

expected.png files generated with 3.3.0+1846~g877ba6e66 and recompressed with zopflipng.


TurboGit:
Lens correction does not pass, no many wrong pixels but yet with some dE > 12.

```
Test 0060-lens-correction
      Image mire1.cr2
      CPU & GPU version differ by 327679 pixels
      Max  dE         12.6827
      Mean dE         0.0323
      Count above max 229
  FAILS: image visually changed
         (290088 pixels changed)
```

ralfbrown:
Will investigate.  Would not be surprised if there are different correction parameters in different versions of lensfun.
I used deliberately "bad" settings to make any differences more obvious (just look at the huge color fringes from the extreme TCA settings).

parafin:
Which version of lensfun library both of you are using? I hope it's not 0.3.95.

TurboGit:
@parafin : Of course not, this version is broken but worth checking indeed, mine is 0.3.2-5.

ralfbrown:
I have 0.3.2-8 (and I think the -N are just distro's patches for both of us).  I'd never run the database update, so tried that; didn't change the output after the update.
In the interest of not delaying this PR, maybe I should just swap out the lens correction for another module that isn't covered yet?  Then we can deal with lensfun reproducibility later (because 290K changed pixels is not "a few").

TurboGit:
@ralfbrown : let's do that yes so we can move forward.

ralfbrown:
Done -- replaced lens correction by rgb curve for 0060.

TurboGit:
Works for me. I have reworked the history to no include and remove the png for the lens correction module. So merged manually. Thanks.

Issue #7086, opened 2020-11-27
Add X-Trans image for the integration test suite


ralfbrown:
>  it would be good to have an X-Trans image for the integration test suite

Sure would be nice to have one in the images directory. Do you have one to share ? I don't have an X-Trans camera, but I'm sure we can find a free sample somewhere. Let me know.

_Originally posted by @TurboGit in https://github.com/darktable-org/darktable/issues/7084#issuecomment-734499401_

Looking through raw.pixls.us as suggested by johnny-bit, I would recommend one of the following if we would like to keep the file size modest:
XQ2: https://raw.pixls.us/data/Fujifilm/XQ2/DSCF4186.RAF (19 MB)
X-T2: https://raw.pixls.us/data/Fujifilm/X-T2/20170114_174341_TFW04727.RAF (22 MB)

The former is a picture of a boat on a lake on a nearly cloudless day, so it's rather lacking in reds and greens.

The latter is a shot of a photographic target with stepped gray scales and color "spectra", but it's **substantially** underexposed.

A couple of the others I downloaded are even less suitable.

I haven't looked through the models with bigger sensors, but if you're OK with a 30-45MB image file, I can see if there's something better for exercising image processing.

ralfbrown:
Found a sample with a ColorChecker in it: https://raw.pixls.us/data/Fujifilm/X-T4/DSCF0267.RAF  (27 MB)

TurboGit:
@ralfbrown : Perfect I was looking for an image with a color checker in it too. The latest is perfect.

Do you want me to integrate it ?

I would place the image into the `integration/images` directory under the name `mire1-xtrans.raf`.

chhil:
If this helps....used my x100v
https://www.dropbox.com/s/noo9tkkhz6rrppq/DSCF0783.RAF?dl=0

ralfbrown:
@chhil That would be good, but the color checker is underexposed (the bright sky fooled the meter).

@TurboGit I'll leave it up to you to decide on the image and integrate it.

chhil:
Ping me if you still need one and I can take another one tomorrow morning (my time).

TurboGit:
X-Trans tests image now integrated. I have taken the first with color checker where the exposition is ok.

PR #7087, opened 2020-11-27
integration tests for X-Trans demosaic

ralfbrown:
Ref #7086


TurboGit:
On my side I have rawdenoise test failing:

```
Test 0068-rawdenoise-xtrans
      Image mire1-xtrans.raf
      CPU & GPU version differ by 18913 pixels
      Max  dE         6.0342
      Mean dE         0.0083
      Count above max 235
  FAILS: image visually changed
         (58014 pixels changed)
```

TurboGit:
And the diff is showing a really bad pattern:

diff

ralfbrown:
That's exactly the pattern that comes out when comparing a 16-threaded run against a single-threaded run.  But I used run.sh to create the expected.png, so we should both be running with four threads.  Hmm....

The code is doing a nearest-neighbor interpolation to generate a full-size image in each of the R/G/B channels, which involves copying each R and B pixel to all eight of its neighbors.  When parallelizing, the first row of each horizontal slice gets a different value, because instead of the lower row overwriting the higher one, the other thread later comes along and overwrites the lower row from the higher one in the adjacent slice.  The DWT then spreads those changes 32 pixels in all four directions.  Bayer doesn't have this issue, because it processes quarter-size single-channel images for R/G1/G2/B.

TurboGit:
You put me on the right track... I have 16 threads indeed and found out that none of our // for are following the max threads setting.

See #7097 for a patch that fixes the issue on my side.

We should probably do that in every // for, do you agree ?

TurboGit:
Could this also explain the issue with lens test ?

ralfbrown:
That's a rather nasty fix.  OpenMP is properly honoring the desired threads setting on my system, so a better solution  than changing **every** // for would be figuring out why yours isn't, and what setting will make it use the desired number of threads.  Do you have any of the OMP_* environment variables set?  (Though on my system, OMP_NUM_THREADS has no effect, because the thread-setting code in dt's init overrides it -- when I start darktable or darktable-cli with "-t 2", I get the message "[dt_init] using 2 threads for openmp parallel sections" and every single parallel for then uses two threads.)

And yes, thread-counts affecting lens correction had occurred to me, since it's doing subpixel stuff.

TurboGit:
Are you sure yours is honoring the max thread? The parameter is only set into OpenMP in the job controls (exporting for example) but it does not seems set in other places.

TurboGit:
No, I'm wrong it is set in darktable.c  :(

TurboGit:
@ralfbrown : Found a single // for where this is needed, I don't see why really :( See my updated PR, any idea ?

ralfbrown:
That's the loop doing the nearest-neighbor interpolation I described above.  On my system, it properly honors the requested number of threads, so I get different outputs depending on which -t N is specified.

I've been looking at the environment variables for OpenMP.  Try setting OMP_THREAD_LIMIT=4 before running the test.  If that gets you a match to my reference image, then my suspicion is that OMP_DYNAMIC is either set or defaults to TRUE, and that dynamic thread allocation in your OpenMP goes all the way up to the thread limit regardless of OMP_NUM_THREADS -- contrary to the wording in the 4.5 spec, where OMP_DYNAMIC is allowed to choose anything between 1 and min(available,limit).  The fix for that would be adding a call omp_set_dynamic(0) in the init.

TurboGit:
> OMP_THREAD_LIMIT=4

This fixes the problem indeed.

But adding omp_set_dynamic() does not solves the issue:

```
diff --git a/src/common/darktable.c b/src/common/darktable.c
index 9c6c75162..165e68309 100644
--- a/src/common/darktable.c
+++ b/src/common/darktable.c
@@ -866,6 +866,7 @@ int dt_init(int argc, char *argv[], const gboolean init_gui, const gboolean load
 
 #ifdef _OPENMP
   omp_set_num_threads(darktable.num_openmp_threads);
+  omp_set_dynamic(0);
 #endif
```

ralfbrown:
Doing more web searches hasn't located any further useful hints (unlike NUM_THREADS, there's no function call equivalent to the environment variable for OMP_THREAD_LIMIT).

Since the integration tests are the only place where it's critical to have the exact number of threads, let me suggest just adding OMP_THREAD_LIMIT=4 to run.sh.

TurboGit:
Yep, will do that.

I have even checked that omp_get_max_threafs() properly return 4 but yet it seems that the loop is using far more threads.

TurboGit:
All good now, merging !

PR #7091, opened 2020-11-27
vibrance: clean up main loop

ralfbrown:
Cleaned up main loop in process() and added simd directive, plus added lots of consts.  GCC 8.4 partially vectorizes it now.

Verified against integration test 0056.


TurboGit (approved):
Thanks !

